Issue #1096, opened 2019-11-29
Add API endpoint for delegation fee estimation


Anviking:
# Context

Users of the API would like to know the cost of joining a stake-pool before pulling the trigger.

In this task we implement the API, leveraging the work of #1094. 

# Decision

Add the following endpoint:
```yaml
/wallets/\{walletId\}/delegations/fees: 
    get: 
      operationId: getDelegationFee
      tags: ["Stake Pools"]
      summary: Estimate Fee
      parameters: 
        - *parametersWalletId
      responses: *responsesGetDelegationFee

x-responsesGetDelegationFee: &responsesGetDelegationFee
  <<: *responsesErr403
  <<: *responsesErr404
  <<: *responsesErr405
  <<: *responsesErr406
  200:
    description: Ok
    schema: *ApiFee
```

# Acceptance Criteria

- The endpoint _must_ be implemented
- JSON roundtrip tests for any added Api-types _should_ exist
- The endpoint _should_ be tested with integration tests

---

# Development 

- #1116 

# QA 

- Added integration test scenarios here: https://github.com/input-output-hk/cardano-wallet/blob/decdad3e649ae9278d47815926d729ee56e25941/lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs#L515-L558

jonathanknowles:
We might want to add integration tests to check that:
- The _eventual_ fee paid for joining a stake pool is similar or identical to the _estimated_ delegation fee (similar to `BYRON_MIGRATE_03`).
- The estimated delegation fee for a non-empty wallet is `> 0` (similar to `BYRON_CALCULATE_01`).
- Attempting to calculate a delegation fee for an empty wallet produces an appropriate error (similar to `BYRON_CALCULATE_02`).
- Attempting to calculate a delegation fee for a non-Shelley wallet produces an appropriate error (similar to `BYRON_CALCULATE_03`).
- Attempting to calculate a delegation fee for a non-existing wallet produces an appropriate error (similar to `BYRON_CALCULATE_04`).

jonathanknowles:
@KtorZ I've added a question regarding the original requirement here: https://jira.iohk.io/browse/WB-32?focusedCommentId=11089&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-11089

[...]

Issue #1094, opened 2019-11-29
Add `Wallet.estimateJoinStakePoolFee`


Anviking:
# Context

We currently have [`selectCoinsForPayment`](https://github.com/input-output-hk/cardano-wallet/blob/870914d3e684191bf5953f6445b0f1dbf32601d2/lib/core/src/Cardano/Wallet.hs#L738-L771), [`estimatePaymentFee`](https://github.com/input-output-hk/cardano-wallet/blob/870914d3e684191bf5953f6445b0f1dbf32601d2/lib/core/src/Cardano/Wallet.hs#L829-L851) and [`selectCoinsForDelegation`](https://github.com/input-output-hk/cardano-wallet/blob/870914d3e684191bf5953f6445b0f1dbf32601d2/lib/core/src/Cardano/Wallet.hs#L773-L799).

We can estimate the fees of a normal transaction before submitting it. With WB-32 we want to estimate the fees for joining a stake-pool.

These functions do not appear to have direct unit tests.

# Decision

Add a `Wallet.estimateJoinStakePoolFee` function to complete the pattern of the three mentioned functions.

# Acceptance Criteria

1. There _must_ exist a function in `Cardano.Wallet` that can estimate the fees required to join a stake-pool

# Development 

- I have a feeling there _might_ be opportunities for refactoring here
- #1116 

# QA 

- Added integration test scenarios here: https://github.com/input-output-hk/cardano-wallet/blob/decdad3e649ae9278d47815926d729ee56e25941/lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs#L515-L558

jonathanknowles:
@Anviking, are you able to expand a little more on the following point:

> If opportunity arises, we may refactor the functions.

 I suspect this is something that could go in the `Development` section (instead of being an acceptance criterion per se). What do you think?

jonathanknowles:
**Regarding the precision of our fee estimation:**

When we added support for migration from Byron wallets (to Jörmungandr wallets), we were able to offer a **_precise_** fee calculation endpoint (`getByronWalletMigrationInfo`). Furthermore, we actually _test_ that the predicted value is precisely the same as the eventual value in our integration tests. (See `BYRON_MIGRATE_03`.)

[...]

PR #1116, opened 2019-12-04
Estimate fees for joining a stake pool

Anviking:
# Issue Number

WB-32: #1094 #1096 (Found it easier to tackle both at once)

# Overview

Also see commit history, but main points:

- [x] Added API endpoint *GET /stake-pools/{stake-pool}/wallets/{wallet}/fee*
- [x] Added `feeBalance :: CoinSelection -> Word64`. I re-use `selectCoinsForDelegation` together with `feeBalance` to implement the `joinStakePoolFee` handler.
- [x] Integration tests:
    - STAKE_POOLS_ESTIMATE_FEE_01 - fee matches eventual cost
    - STAKE_POOLS_ESTIMATE_FEE_02 - empty wallet cannot estimate fee
    - STAKE_POOLS_ESTIMATE_FEE_03 - can't use byron wallets
    - STAKE_POOLS_ESTIMATE_FEE_04 - invalid pool and wallet ids

### DIFF
--- a/lib/core-integration/src/Test/Integration/Framework/TestData.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/TestData.hs
@@ -32,6 +32,10 @@ module Test.Integration.Framework.TestData
     , russianWalletName
     , wildcardsWalletName
 
+    -- * Stake pool ids
+    , nonExistingStakePool1
+    , invalidPoolIds
+
     -- * Helpers
     , addressPoolGapMax
     , addressPoolGapMin
@@ -265,6 +269,23 @@ arabicWalletName = "ثم نفس سقطت وبالتحديد،, جزيرتي با
 wildcardsWalletName :: Text
 wildcardsWalletName = "`~`!@#$%^&*()_+-=<>,./?;':\"\"'{}[]\\|❤️ 💔 💌 💕 💞 \
 \💓 💗 💖 💘 💝 💟 💜 💛 💚 💙0️⃣ 1️⃣ 2️⃣ 3️⃣ 4️⃣ 5️⃣ 6️⃣ 7️⃣ 8️⃣ 9️⃣ 🔟🇺🇸🇷🇺🇸 🇦🇫🇦🇲🇸"
+
+
+--
+-- Stake Pools
+--
+
+nonExistingStakePool1 :: Text
+nonExistingStakePool1 = "008d686a02c6e625b5a59cc9e234f32e5d72987012f9c25c9a6b60ddade197d1"
+
+invalidPoolIds :: [(String, String)]
+invalidPoolIds =
+        [ ("64 chars non-hex", replicate 64 'ś')
+        , ("63 chars hex", replicate 63 '1')
+        , ("65 chars hex", replicate 65 '1')
+        , ("64 chars hex", replicate 64 '1')
+        , ("empty", "")
+        ]
 
 ---
 --- Helpers
--- a/lib/core-integration/src/Test/Integration/Framework/TestData.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/TestData.hs
@@ -32,10 +32,6 @@ module Test.Integration.Framework.TestData
     , russianWalletName
     , wildcardsWalletName
 
-    -- * Stake pool ids
-    , nonExistingStakePool1
-    , invalidPoolIds
-
     -- * Helpers
     , addressPoolGapMax
     , addressPoolGapMin
@@ -270,22 +266,6 @@ wildcardsWalletName :: Text
 wildcardsWalletName = "`~`!@#$%^&*()_+-=<>,./?;':\"\"'{}[]\\|❤️ 💔 💌 💕 💞 \
 \💓 💗 💖 💘 💝 💟 💜 💛 💚 💙0️⃣ 1️⃣ 2️⃣ 3️⃣ 4️⃣ 5️⃣ 6️⃣ 7️⃣ 8️⃣ 9️⃣ 🔟🇺🇸🇷🇺🇸 🇦🇫🇦🇲🇸"
 
-
---
--- Stake Pools
---
-
-nonExistingStakePool1 :: Text
-nonExistingStakePool1 = "008d686a02c6e625b5a59cc9e234f32e5d72987012f9c25c9a6b60ddade197d1"
-
-invalidPoolIds :: [(String, String)]
-invalidPoolIds =
-        [ ("64 chars non-hex", replicate 64 'ś')
-        , ("63 chars hex", replicate 63 '1')
-        , ("65 chars hex", replicate 65 '1')
-        , ("64 chars hex", replicate 64 '1')
-        , ("empty", "")
-        ]
 
 ---
 --- Helpers
--- a/lib/core/src/Cardano/Wallet/Api.hs
+++ b/lib/core/src/Cardano/Wallet/Api.hs
@@ -138,8 +138,8 @@ type CoreApi t =
 type StakePoolApi t =
     ListStakePools
     :<|> JoinStakePool t
-    :<|> JoinStakePoolFee
     :<|> QuitStakePool t
+    :<|> DelegationFee
 
 type CompatibilityApi n =
     DeleteByronWallet
@@ -284,12 +284,11 @@ type JoinStakePool t = "stake-pools"
     :> PutAcccepted '[JSON] (ApiTransaction t)
 
 
--- | https://input-output-hk.github.io/cardano-wallet/api/#operation/joinStakePool
-type JoinStakePoolFee = "stake-pools"
-    :> Capture "stakePoolId" (ApiT PoolId)
-    :> "wallets"
+-- | https://input-output-hk.github.io/cardano-wallet/api/#operation/getDelegationFee
+type DelegationFee = "wallets"
     :> Capture "walletId" (ApiT WalletId)
-    :> "fee"
+    :> "delegations"
+    :> "fees"
     :> Get '[JSON] ApiFee
 
 -- | https://input-output-hk.github.io/cardano-wallet/api/#operation/quitStakePool
--- a/lib/core/src/Cardano/Wallet/Api.hs
+++ b/lib/core/src/Cardano/Wallet/Api.hs
@@ -138,6 +138,7 @@ type CoreApi t =
 type StakePoolApi t =
     ListStakePools
     :<|> JoinStakePool t
+    :<|> JoinStakePoolFee
     :<|> QuitStakePool t
 
 type CompatibilityApi n =
@@ -281,6 +282,15 @@ type JoinStakePool t = "stake-pools"
     :> Capture "walletId" (ApiT WalletId)
     :> ReqBody '[JSON] ApiWalletPassphrase
     :> PutAcccepted '[JSON] (ApiTransaction t)
+
+
+-- | https://input-output-hk.github.io/cardano-wallet/api/#operation/joinStakePool
+type JoinStakePoolFee = "stake-pools"
+    :> Capture "stakePoolId" (ApiT PoolId)
+    :> "wallets"
+    :> Capture "walletId" (ApiT WalletId)
+    :> "fee"
+    :> Get '[JSON] ApiFee
 
 -- | https://input-output-hk.github.io/cardano-wallet/api/#operation/quitStakePool
 type QuitStakePool t = "stake-pools"
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -787,8 +787,8 @@ stakePools
 stakePools ctx spl =
     listPools spl
     :<|> joinStakePool ctx spl
-    :<|> joinStakePoolFee ctx spl
     :<|> quitStakePool ctx
+    :<|> delegationFee ctx
 
 listPools
     :: StakePoolLayer IO
@@ -838,7 +838,7 @@ joinStakePool ctx spl (ApiT pid) (ApiT wid) (ApiWalletPassphrase (ApiT pwd)) = d
   where
     liftE = throwE . ErrJoinStakePoolNoSuchWallet
 
-joinStakePoolFee
+delegationFee
     :: forall ctx s t n k.
         ( DelegationAddress n k
         , Buildable (ErrValidateSelection t)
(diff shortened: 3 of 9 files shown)


lib/core-integration/src/Test/Integration/Framework/TestData.hs
```diff
@@ -270,6 +273,14 @@ wildcardsWalletName :: Text
 wildcardsWalletName = "`~`!@#$%^&*()_+-=<>,./?;':\"\"'{}[]\\|❤️ 💔 💌 💕 💞 \
 \💓 💗 💖 💘 💝 💟 💜 💛 💚 💙0️⃣ 1️⃣ 2️⃣ 3️⃣ 4️⃣ 5️⃣ 6️⃣ 7️⃣ 8️⃣ 9️⃣ 🔟🇺🇸🇷🇺🇸 🇦🇫🇦🇲🇸"
 
+
+--
+-- Stake Pools
+--
+
+stakePoolId1 :: Text
+stakePoolId1 = "4f8d686a02c6e625b5a59cc9e234f32e5d72987012f9c25c9a6b60ddade197d1"
```
KtorZ: **[change required]**

Use `listStakePools` in integration scenarios to find a stake pool id.

**Why?**

This increases the area of "things-that-have-to-be-changed" every time we make a change in the genesis.yaml. Avoid hard-coded ids as much as possible, especially when they can be easily retrieved by other means.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -845,6 +847,27 @@ estimatePaymentFee ctx wid recipients = do
...
+    -> WalletId
+    -> ExceptT ErrEstimateJoinStakePoolFee IO Fee
+estimateJoinStakePoolFee ctx wid = do
+    (wal, _, _) <- withExceptT ErrEstimateJoinStakePoolFeeNoSuchWallet $
+        readWallet @ctx @s @k ctx wid
+    sel <- withExceptT ErrEstimateJoinStakePoolSelectForDelegation $
+        selectCoinsForDelegation @ctx @s @t @k ctx wid
+    let bp = blockchainParameters wal
+    let opts = feeOpts tl (computeCertFee 1) (bp ^. #getFeePolicy)
+    return $ estimateFee opts sel
```
KtorZ: **[change required]**

Use instead `inputBalance`, `changeBalance` and `outputBalance` to compute the _actual fee_ on a given selection.

**Why?**

There's no need to re-estimate fees whereas they can really be known exactly. This has little importance for Jörmungandr but is _more correct_ and will matter for the Haskell nodes.

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
```diff
@@ -404,6 +414,79 @@ spec = do
...
+        fee <- do
+            w <- fixtureWallet ctx
+            (_, r) <- joinStakePoolFee ctx (p ^. #id) w
+            print r
+            return $ either (error . show) (getQuantity . amount) r
+
+        w <- fixtureWalletWith ctx [fee - 1]
+        joinStakePoolFee ctx (p ^. #id) w >>= flip verify
+            [ expectResponseCode HTTP.status403
+            , expectErrorMessage "I'm unable to select enough coins to pay for a"
```
KtorZ: **[change required]**

Use `errMsg403DelegationFee` from `Test.Integration.Framework.TestData`

**Why?**

Avoid magic constants in code.

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
```diff
@@ -404,6 +414,79 @@ spec = do
...
+    it "STAKE_POOLS_ESTIMATE_FEE_01 - \
+        \wallet with [fee - 1] => cannot estimate fee" $ \ctx -> do
+        putStrLn "hello"
+        (_, p:_) <- eventually $
+            unsafeRequest @[ApiStakePool] ctx listStakePoolsEp Empty
+        fee <- do
+            w <- fixtureWallet ctx
+            (_, r) <- joinStakePoolFee ctx (p ^. #id) w
+            print r
+            return $ either (error . show) (getQuantity . amount) r
```
KtorZ: **[suggestion]**

Re-write as: 

```hs
getFromResponse amount <$> joinStakePoolFee ctx (p ^. #id) w
```
**Why?**

The intent is clearer and the overall code is more concise.

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
```diff
@@ -404,6 +414,79 @@ spec = do
...
+            let ep = listTxEp w mempty
+            request @[ApiTransaction n] ctx ep Default Empty >>= flip verify
+                [ expectListItemFieldEqual 0 direction Outgoing
+                , expectListItemFieldEqual 0 status InLedger
+                ]
+
+        -- Verify the wallet is now delegating
+        request @ApiWallet ctx (getWalletEp w) Default Empty >>= flip verify
+            [ expectFieldEqual delegation (Delegating (p ^. #id))
+            ]
```
KtorZ, line 538: **[question]** 
Why joining a stake pool after looking for the fee? Shouldn't this maybe check that the actual fees are somewhat closed to the estimated one? (they should be exact for Jörmungandr though..).

lib/core-integration/src/Test/Integration/Framework/TestData.hs
Anviking: This was intended as a _not-necessarily_ existing pool id… I could change the id and rename it to `nonExistingPoolId`…

Otherwise, fully agree on the "things-that-have-to-be-changed" argument of course.

KtorZ: Ah! If it's a non existing pool id, then, a better name would do!

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
Anviking: This was testing that a `fixtureWalletWith [fee]` can join **for sure**. It found the [edge-case bug](https://input-output-rnd.slack.com/archives/GBT05825V/p1575449724031000) in the coin-selection when the fee-coefficient was enabled.

But you have a point, it is not relevant to fee estimation; I have removed removed it from this pr.

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
```diff
@@ -493,6 +503,52 @@ spec = do
...
+        r <- joinStakePool ctx (p ^. #id) (w, fixturePassphrase)
+        verify r
+            [ expectFieldEqual amount fee
+            ]
+
+    it "STAKE_POOLS_ESTIMATE_FEE_02 - \
+        \empty wallet cannot estimate fee" $ \ctx -> do
+        (_, p:_) <- eventually $
+            unsafeRequest @[ApiStakePool] ctx listStakePoolsEp Empty
+        w <- emptyWallet ctx
```
Anviking, line 533: We are now not testing edge-cases. I think @KtorZ [suggested](https://input-output-rnd.slack.com/archives/GBT05825V/p1575459104040400?thread_ts=1575309301.012300&cid=GBT05825V) that there are existing unit tests for coin selection that should be relevant and better suited for edge-cases. (haven't looked for them myself yet)

lib/core-integration/src/Test/Integration/Framework/TestData.hs
```diff
@@ -270,6 +274,23 @@ wildcardsWalletName :: Text
 wildcardsWalletName = "`~`!@#$%^&*()_+-=<>,./?;':\"\"'{}[]\\|❤️ 💔 💌 💕 💞 \
 \💓 💗 💖 💘 💝 💟 💜 💛 💚 💙0️⃣ 1️⃣ 2️⃣ 3️⃣ 4️⃣ 5️⃣ 6️⃣ 7️⃣ 8️⃣ 9️⃣ 🔟🇺🇸🇷🇺🇸 🇦🇫🇦🇲🇸"
 
+
+--
+-- Stake Pools
+--
+
+nonExistingStakePool1 :: Text
+nonExistingStakePool1 = "008d686a02c6e625b5a59cc9e234f32e5d72987012f9c25c9a6b60ddade197d1"
```
jonathanknowles:
```suggestion
nonExistingStakePool1 =
    "008d686a02c6e625b5a59cc9e234f32e5d72987012f9c25c9a6b60ddade197d1"
```

lib/core-integration/src/Test/Integration/Framework/TestData.hs
```diff
@@ -270,6 +274,23 @@ wildcardsWalletName :: Text
...
+
+invalidPoolIds :: [(String, String)]
+invalidPoolIds =
+        [ ("64 chars non-hex", replicate 64 'ś')
+        , ("63 chars hex", replicate 63 '1')
+        , ("65 chars hex", replicate 65 '1')
+        , ("64 chars hex", replicate 64 '1')
```
jonathanknowles: Question: Is this one (64 hexademical characters) actually invalid? If so, why is it invalid?

(Apologies if I have missed something obvious!)

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -282,6 +283,15 @@ type JoinStakePool t = "stake-pools"
     :> ReqBody '[JSON] ApiWalletPassphrase
     :> PutAcccepted '[JSON] (ApiTransaction t)
 
+
+-- | https://input-output-hk.github.io/cardano-wallet/api/#operation/joinStakePool
```
jonathanknowles: I suspect this one should be `joinStakePoolFee`. So:
```suggestion
-- | https://input-output-hk.github.io/cardano-wallet/api/#operation/joinStakePoolFee
```

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -836,6 +837,27 @@ joinStakePool ctx spl (ApiT pid) (ApiT wid) (ApiWalletPassphrase (ApiT pwd)) = d
...
+        , k ~ ShelleyKey
+        , HardDerivation k
+        , ctx ~ ApiLayer s t k
+        )
+    => ctx
+    -> StakePoolLayer IO
+    -> ApiT PoolId
+    -> ApiT WalletId
+    -> Handler ApiFee
+joinStakePoolFee ctx _spl (ApiT _pid) (ApiT wid) = do
```
jonathanknowles: At first glance, it seems as though the following arguments can be completely omitted:

1. `_spl :: StakePoolLayer IO`
2. `_pid :: PoolId`

The first argument can be omitted by changing the definition of `stakePools` to:
```hs
stakePools ctx spl =
    listPools spl
    :<|> joinStakePool ctx spl
    :<|> joinStakePoolFee ctx
    :<|> quitStakePool ctx
```

The second argument leads me to another question:

Will the fee calculation always be independent of the pool ID? If the answer is yes, then is there any need for the endpoint to include the pool ID?

lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
```diff
@@ -92,6 +93,9 @@ outputBalance = foldl' addTxOut 0 . outputs
 changeBalance :: CoinSelection -> Word64
 changeBalance = foldl' addCoin 0 . change
 
+feeBalance :: CoinSelection -> Word64
+feeBalance sel = inputBalance sel - outputBalance sel - changeBalance sel
```
jonathanknowles, line 97: :+1:

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
```diff
@@ -493,6 +503,52 @@ spec = do
...
 
+    it "STAKE_POOLS_ESTIMATE_FEE_01 - fee matches eventual cost" $ \ctx -> do
+        (_, p:_) <- eventually $
+            unsafeRequest @[ApiStakePool] ctx listStakePoolsEp Empty
+        w <- fixtureWallet ctx
+        fee <- getFromResponse amount <$> joinStakePoolFee ctx (p ^. #id) w
+        r <- joinStakePool ctx (p ^. #id) (w, fixturePassphrase)
+        verify r
+            [ expectFieldEqual amount fee
+            ]
```
jonathanknowles, line 529: **Question**: Is this expectation valid?

I was under the impression that the precision of fee estimation cannot be guaranteed, because of randomization that occurs during coin selection. See: https://github.com/input-output-hk/cardano-wallet/issues/1094#issuecomment-559773456

If there is randomization going on under the hood, then will this occasionally trigger a test failure?

specifications/api/swagger.yaml
```diff
@@ -1311,6 +1316,20 @@ paths:
...
+    get:
+      operationId: joinStakePoolFee
+      tags: ["Stake Pools"]
+      summary: Estimate
+      description: |
+        <p align="right">status: <strong>stable</strong></p>
+
+        Estimate fee for joining a stake pool.
+      parameters:
+        - *parametersStakePoolId
```
jonathanknowles: Why is this parameter required?

See https://github.com/input-output-hk/cardano-wallet/pull/1116/files#diff-610bd6b69383f967d3fad205d6c6d86aR854

jonathanknowles (commented):
Hi @Anviking

Looks mostly good to me! Though I have a few questions (see comments).

My main question:

Why is the `stakePoolId` argument required by the `joinStakePoolFee` endpoint? It seems as though the fee calculation is independent of the particular stake pool being joined. Are we anticipating a future in which the calculation will depend on the target stake pool?

lib/core-integration/src/Test/Integration/Framework/TestData.hs
paweljakubas: @Anviking  why not to use something like this : 
```
let poolIdAbsent = PoolId $ BS.pack $ replicate 32 0
```
it is already used in `"STAKE_POOLS_JOIN_01 - Cannot join non-existant stakepool"`
then you can invoke `toText` on it ...

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
KtorZ: This is true only by "accident". The wallet test UTxO are all very homogeneous (10 coins of the same amount). So, this compensate the random aspect of it. 

Yet, I agree that checking this in a scenario is misleading and somewhat wrong.

lib/core/src/Cardano/Wallet/Api/Server.hs
KtorZ: I've indeed removed the pool id from the endpoint path and from the handler arguments.

KtorZ (approved):
rebased + applied reviewed comments.

PR #1120, opened 2019-12-04
Add memoization to speed up release-lib.nix

hamishmack:
`release-lib.nix` calls `packageSet` many times. This
results in lots of repeated work, both during eval stage
and when the jobs are run.

By memoizing the calls here we get a very significant
speedup of the eval time on hydra.

Long term, someone should look into fixing or replacing
`release-lib.nix` so that it does not call the `packageSet`
argument any more than is necessary.

### DIFF
--- a/nix/stylish-haskell.nix
+++ b/nix/stylish-haskell.nix
@@ -4,18 +4,23 @@ let
   src = fetchFromGitHub {
     owner = "jaspervdj";
     repo = "stylish-haskell";
+    # When updating this use `checkMaterialization` flag below to
+    # update the `stack-sha256` and `materialized` arguments below.
     rev = "9958a5253a9498c29508895450c4ac47542d5f2a";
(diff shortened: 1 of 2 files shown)


hamishmack:
How much faster?  The eval time for this PR was 3m.  The shortest eval time I have seen without it has been over 17m.

rvl:
That is a significant improvement, thanks.

However even a 3 minute evaluation is a long time. Why is the package set taking this long to evaluate?

That slowness is the root cause and it also impacts building things from `default.nix` and starting `nix-shell`.

If it's absolutely necessary to memoize the package set, now is the time to add memoization [`release-lib.nix`](https://github.com/input-output-hk/iohk-nix/blob/master/lib/release-lib.nix).

KtorZ:
@rvl @hamishmack Anything we want to do with this PR? Shall we simply merged it as it is, close it? 

@rvl If it's nothing essential, I'd rather have us working on something else :+1: , if it's a nice-to-have already as it is, let's merge it then and see later about possible improvements.

rvl:
I am not too keen on the methods used here to speed up evaluation time.

Issue #1067, opened 2019-11-25
Benchmark is failing


piotr-iohk:
# Context
 
 

# Test Case

DB Benchmark -> https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/289

# Failure / Counter-example

```
PersistMarshalError "Couldn't parse field `checkpointHeaderHash` from table `checkpoint`. not a valid value: Invalid blockHeader hash: expecting a hex-encoded value that is 32 bytes in length."
db: PersistMarshalError "Couldn't parse field `checkpointHeaderHash` from table `checkpoint`. not a valid value: Invalid blockHeader hash: expecting a hex-encoded value that is 32 bytes in length."
```

---

# Resolution 

- Fix the generation of arbitrary `Hash` which now must be 32 bytes in length ==> #1111 
- Revised a bit bench scenarios and how db transaction are executed ==> #1124 

---

# QA 

- Latest nightly on top of #1124 https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/303

PR #1124, opened 2019-12-05
Core Benchmark Adjustments

KtorZ:
# Issue Number

#1067 

# Overview

1/ Use db transactions "correctly" :s .. I noticed that the benchmark where doing a lot of mapM_ (atomically action) instead of atomitically (mapM_ action). The difference is subtle but the later is much faster (one db transaction) and actually, better reflects what the actual code does in the wallet engine.
2/ I've removed some pointless extreme case like trying to insert 1000 checkpoints. We insert at most k/100 + 10 checkpoints, so that's ~30 checkpoints. So, now, the worst case bench is just about inserting 100 checkpoints. Though, as a counterpart, I've increased the max size of the UTxO to make sure they better reflect what we might face.
3/ I've removed the "NBatch" parameter for transactions. We always insert transaction history as one batch (though the internal db engine might do multiple batch, that's irrelevant for the bench setup).
4/ I've added some heavier transaction write with 255 and 255 max inputs / outputs since this is the actual limit of Jörmungandr (which didn't end up using a soft tx limit in the end).

# Comments

Nightly here: https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/303

### DIFF
--- a/lib/core/test/bench/db/Main.hs
+++ b/lib/core/test/bench/db/Main.hs
@@ -277,23 +277,25 @@ bgroupSeqState db = bgroup "SeqState"
 -- - 100 outputs
 bgroupWriteTxHistory :: DBLayerBench -> Benchmark
 bgroupWriteTxHistory db = bgroup "TxHistory (Write)"
-    --              #NBatch  #BatchSize #NInputs #NOutputs  #SlotRange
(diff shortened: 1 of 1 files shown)


piotr-iohk (approved):
ok... I guess :sweat_smile:

Issue #1115, opened 2019-12-04
Faililing tests integration tests on Windows


piotr-iohk:
# Context
 
#703

# Test Case

Following 21 tests are failing (after the path issue fix -> #1109):
```
Failures:

  test/integration/Cardano/Wallet/Jormungandr/NetworkSpec.hs:135:9:
  1) No backend required, Cardano.Wallet.NetworkSpec, Happy Paths, getAccountState: successfully gets account state
       uncaught exception: IOException of type PermissionDenied
       C:\Users\katar\AppData\Local\Temp\cardano-wallet-jormungandr-dbc61946bb87b0ae\chain\blocks.sqlite-shm: removeDirectoryRecursive:removeContentsRecursive:removePathRecursive:removeContentsRecursive:removePathRecursive:DeleteFile "\\\\?\\C:\\Users\\katar\\AppData\\Local\\Temp\\cardano-wallet-jormungandr-dbc61946bb87b0ae\\chain\\blocks.sqlite-shm": permission denied (Odmowa dostępu.)

  To rerun use: --match "/No backend required/Cardano.Wallet.NetworkSpec/Happy Paths/getAccountState: successfully gets account state/"

  test/integration/Cardano/Wallet/Jormungandr/NetworkSpec.hs:147:9:
  2) No backend required, Cardano.Wallet.NetworkSpec, Happy Paths, get network tip
       uncaught exception: IOException of type PermissionDenied
       C:\Users\katar\AppData\Local\Temp\cardano-wallet-jormungandr-fddede207e198264\chain\blocks.sqlite-shm: removeDirectoryRecursive:removeContentsRecursive:removePathRecursive:removeContentsRecursive:removePathRecursive:DeleteFile "\\\\?\\C:\\Users\\katar\\AppData\\Local\\Temp\\cardano-wallet-jormungandr-fddede207e198264\\chain\\blocks.sqlite-shm": permission denied (Odmowa dostępu.)

  To rerun use: --match "/No backend required/Cardano.Wallet.NetworkSpec/Happy Paths/get network tip/"

  test/integration/Cardano/Wallet/Jormungandr/NetworkSpec.hs:155:9:
  3) No backend required, Cardano.Wallet.NetworkSpec, Happy Paths, get some blocks from the genesis
       uncaught exception: IOException of type PermissionDenied
[...]

PR #1125, opened 2019-12-05
Fixing Windows  failing tests

piotr-iohk:
# Issue Number

#1115

# Overview

- [ ] Fixed `TRANS_EXTERNAL_CREATE_01` and `CLI_VERSION`

# Comments

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/CLI/Miscellaneous.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/CLI/Miscellaneous.hs
@@ -11,7 +11,7 @@ module Test.Integration.Scenario.CLI.Miscellaneous
 import Prelude
 
 import Cardano.Wallet.Version
-    ( gitRevision, showFullVersion, version )
+    ( showVersion, version )
 import Control.Monad
     ( forM_ )
 import System.Command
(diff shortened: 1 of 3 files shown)


lib/core-integration/src/Test/Integration/Scenario/CLI/Miscellaneous.hs
```diff
@@ -32,7 +32,7 @@ spec = do
     it "CLI_VERSION - cardano-wallet shows version" $  do
         (Exit c, Stdout out) <- cardanoWalletCLI @t ["version"]
         let v = L.dropWhileEnd (== '\n') out
-        v `shouldBe` (showFullVersion version gitRevision)
+        v `shouldContain` (showVersion version <> " (git revision: " )
```
KtorZ, line 35: **[question]**
How o.O ? `showFullVersion` is just `showVersion v <> " (git revision: " <> T.unpack r <> ")"` . The last parenthesis doesn't show up on Windows? Or maybe there's some trailing characters :thinking: ?

KtorZ (approved):

lib/core-integration/src/Test/Integration/Scenario/CLI/Miscellaneous.hs
piotr-iohk: Not sure why it is like this but the windows testing bundle expects this version -> `2019.11.18 (git revision: 0000000000000000000000000000000000000000)` and the test is failing because of this. (I checked that the binary `cardano-wallet-jormungandr.exe version` shows the correct version, so I just went after @rvl's suggestion for fixing this test https://github.com/input-output-hk/cardano-wallet/issues/1115#issuecomment-561584001).

rvl: The test exes aren't stamped with the git revision, so it's 000...

rvl (approved):

PR #1129, opened 2019-12-06
Buildkite: Check for modified golden test data after running tests

rvl:
Prevents CI from passing if golden test data is missing.


rvl:
Buildkite fails because `ApiByronWalletBalance.json` is not checked in.

KtorZ (approved):
Awesome.

Issue #1091, opened 2019-11-29
Add a database mapping for registered stake pool owners


rvl:
# Context

In order to provide stake pool metadata, the wallet must track the mapping from `PoolId` to `PoolOwner`.

# Decision

In `Cardano.Pool.DB` add the following database table:

```
PoolOwner sql=pool_owner
    poolOwnerPoolId     W.PoolId     sql=pool_id
    poolOwnerOwner      W.PoolOwner  sql=pool_owner
    -- poolOwnerSlot        W.SlotId     sql=slot
    -- poolOwnerSlotRetired W.SlotId Maybe sql=slot_retired
    deriving Show Generic
```

This is a many-to-many relation by the way.

Add DBLayer functions:

```haskell
putStakePoolOwner :: DBLayer m -> SlotId -> PoolId -> PoolOwner -> m ()
-- deleteStakePoolOwners :: DBLayer m -> SlotId -> PoolId -> m ()
readStakePoolOwners :: DBLayer m -> PoolId -> m [PoolOwner]
```

Note that reading stake pool owners for a list of stake pools will not be very efficient, but let's ignore that until will start caching stake pool metadata in the database.

~Rollback is handled by recording the slots when an owner was registered/deregistered.~

# Acceptance Criteria

1. The stake pool owner mapping must be stored in the DB and be queryable.

---

# Development 

1. Implement per _Decision_ section.
2. Add stake pool owners to the model DB and state machine tests.

# QA 

1. The `putStakePoolOwner` and `readStakePoolOwners` functions are tested together in `Cardano.Pool.DB.Properties`.

rvl:
For the time being we are not concerned about handling stake pool retirement certificates, nor rollback. The mapping may contain stale entries, but the entries which aren't stale will still be correct. [WB-40](https://jira.iohk.io/projects/WB/issues/WB-40)

piotr-iohk:
This will require db migration, correct?

KtorZ:
Probably _not_ because it's a new table. So it'll be part of the initial migration `CREATE TABLE IF NOT EXISTS` that are ran automatically by persistent.

_I think_

piotr-iohk:
lgtm

PR #1130, opened 2019-12-06
Add registered stake pool owners to database schema

rvl:
# Issue Number

Relates to #1091.

# Overview

- Adds database table for stake pool owners
- Adds sqlite queries for adding owners and querying by pool id.
- Adds model/mvar implementation (fwiw)
- Property test combinines the put and read functions.

# Comments

This will cause a database migration, which is adding a new table.

### DIFF
--- a/lib/core/src/Cardano/Pool/DB/Model.hs
+++ b/lib/core/src/Cardano/Pool/DB/Model.hs
@@ -37,6 +37,8 @@ module Cardano.Pool.DB.Model
     , mReadPoolProduction
     , mPutStakeDistribution
     , mReadStakeDistribution
+    , mPutStakePoolOwner
+    , mReadStakePoolOwners
     , mReadSystemSeed
     , mRollbackTo
     , mReadCursor
@@ -45,7 +47,7 @@ module Cardano.Pool.DB.Model
 import Prelude
 
 import Cardano.Wallet.Primitive.Types
-    ( BlockHeader (..), EpochNo (..), PoolId, SlotId (..) )
+    ( BlockHeader (..), EpochNo (..), PoolId, PoolOwner (..), SlotId (..) )
 import Data.Map.Strict
     ( Map )
 import Data.Maybe
@@ -54,6 +56,10 @@ import Data.Ord
     ( Down (..) )
 import Data.Quantity
     ( Quantity (..) )
+import Data.Set
+    ( Set )
+import Data.Text.Class
+    ( toText )
 import Data.Word
     ( Word64 )
 import GHC.Generics
@@ -63,6 +69,7 @@ import System.Random
 
 import qualified Data.List as L
 import qualified Data.Map.Strict as Map
+import qualified Data.Set as Set
 
 {-------------------------------------------------------------------------------
                             Model Database Types
@@ -74,6 +81,9 @@ data PoolDatabase = PoolDatabase
 
     , distributions :: !(Map EpochNo [(PoolId, Quantity "lovelace" Word64)])
     -- ^ Store known stake distributions for epochs
+
+    , owners :: !(Set (PoolId, PoolOwner))
+    -- ^ Mapping between pool ids and owners
 
     , seed :: !SystemSeed
     -- ^ Store an arbitrary random generator seed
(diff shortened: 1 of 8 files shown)


lib/core/src/Cardano/Pool/DB/Model.hs
```diff
@@ -146,6 +153,20 @@ mReadStakeDistribution epoch db@PoolDatabase{distributions} =
     , db
     )
 
+mPutStakePoolOwner :: PoolId -> PoolOwner -> ModelPoolOp ()
+mPutStakePoolOwner poolId poolOwner db@PoolDatabase{owners} =
+    ( Right ()
+    , db { owners = (entry:filter (/= entry) owners) }
```
KtorZ: **[remark]**

Alternatively, could use a `Set` for this.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -163,6 +163,15 @@ newDBLayer logConfig trace fp = do
                 [ StakeDistributionEpoch ==. fromIntegral epoch ]
                 []
 
+        , putStakePoolOwner = \poolId poolOwner -> do
+            deleteWhere [PoolOwnerPoolId ==. poolId, PoolOwnerOwner ==. poolOwner]
+            insert_ (PoolOwner poolId poolOwner)
```
KtorZ: **[suggestion]**

Use [`repsert`](https://hackage.haskell.org/package/persistent-2.10.1/docs/Database-Persist-Class.html#v:repsert). 

**Why?**

Better convey the semantic behind these two "statements", and slightly more concise.

lib/core/test/unit/Cardano/Pool/DB/Arbitrary.hs
```diff
@@ -98,6 +107,14 @@ instance Arbitrary PoolId where
         bytes <- vectorOf 32 (elements ['a'..'z'])
         return $ PoolId $ B8.pack bytes
 
+instance Arbitrary StakePoolOwnersFixture where
+    arbitrary = StakePoolOwnersFixture <$> fmap PoolId bs <*> listOf entry
+      where
+        entry = (,) <$> fmap PoolId bs <*> fmap PoolOwner bs
+        bs = B8.replicate 32 <$> elements ['a'..'e']
```
KtorZ: **[suggestion]**

rename `bs` to `genBs` or `genBytes` 

**Why?**

At first glance, a reader may interpret `bs` as a byte string. Yet, it is indeed a generator here and therefore leads to some confusion. Prefixing with `gen`  better reflect what the let binding captures.

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
```diff
@@ -347,6 +351,32 @@ prop_putStakePutStake DBLayer {..} epoch a b =
...
+prop_stakePoolOwner DBLayer {..} (StakePoolOwnersFixture poolId entries) =
+    monadicIO (setup >> prop)
+  where
+    setup = run $ atomically cleanDB
+    prop = do
+        run . atomically $ mapM_ (uncurry putStakePoolOwner) entries
+        res <- run $ atomically (readStakePoolOwners poolId)
+        monitor $ classify dups "duplicate entries"
+        monitor $ classify (haveMany snd) "multiple owners per pool"
+        monitor $ classify (haveMany fst) "multiple pools per owner"
```
KtorZ, line 367: :+1:

KtorZ (approved):
LGTM. Some minor remarks and tweaks .

lib/core/src/Cardano/Pool/DB/Model.hs
rvl: True, fixed.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
rvl: I considered repsert, but it also needed a Key argument which was meh.
But I've changed it now.

lib/core/test/unit/Cardano/Pool/DB/Arbitrary.hs
rvl: OK fixed.

Issue #1090, opened 2019-11-29
Track stake pool registrations in `applyBlock`


rvl:
# Context

In order to provide stake pool metadata, the wallet must track the mapping from `PoolId` to `PoolOwner`.

# Decision

Add new fields to `Cardano.Pool.Metrics.Block`.

```haskell
poolRegistrations :: Block -> [(PoolId, [PoolOwner])]
-- poolRetirements :: Block -> [PoolId]
```

Modify `Cardano.Pool.Metrics.toSPBlock` to get stake pool registration certificates from the `Jormungandr.Binary.Block` type.

# Acceptance Criteria

1. Stake pool registrations must be in the `Block` (stake pool) type.
2. ~Stake pool retirements must be in the FliteredBlock result.~

---

# Development 

0. Remove stake pool retirements from scope ⇒ WB-40.
1. Implement as described, but wrap registration in a type as is done for DelegationCertificate ⇒ #1131.
2. Keep track of tax parameters as well and display them with the metadata ⇒ #1171 
3. Resolve #1181 ⇒ #1182 

# QA 

- Integration test registering new pools here: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs#L195-L220

- Property tests for the database: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Cardano/Pool/DB/Properties.hs#L139-L148

- Unit testing the registration loop with a mock network layer: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Cardano/Pool/MetricsSpec.hs#L148-L150

- See also #1182 comments

PR #1131, opened 2019-12-06
Include registration certificates in Cardano.Pool.Metrics.Block

rvl:
# Issue Number

Relates to #1090.

# Overview

- Adds registration certificates to `Cardano.Pool.Metrics.Block`.

# Comments

- Testing will be done in #1154.

### DIFF
--- a/lib/core/src/Cardano/Pool/Metrics.hs
+++ b/lib/core/src/Cardano/Pool/Metrics.hs
@@ -55,6 +55,7 @@ import Cardano.Wallet.Primitive.Types
     , EpochLength (..)
     , EpochNo (..)
     , PoolId (..)
+    , PoolRegistrationCertificate (..)
     , SlotId (..)
     , SlotNo (unSlotNo)
     )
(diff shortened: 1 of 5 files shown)


lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -426,7 +429,7 @@ prefilterBlock
 prefilterBlock b u0 = runState $ do
     delegations <- mapMaybeM ourDelegation (b ^. #delegations)
     (transactions, ourU) <- foldM applyTx (mempty, u0) (b ^. #transactions)
-    return (FilteredBlock {delegations, transactions}, ourU)
+    return (FilteredBlock {delegations, poolRegistrations = b ^. #poolRegistrations , transactions}, ourU)
```
KtorZ: I don't think this has its place into the **wallet** primitive model and in particular, `prefilterBlock` which role is to identify pieces that are related to our wallet (unless we allowed to register stake pool through the wallet). 

Plus, this function won't likely be "available" from the stake pool workers context because they have no wallet to filter from. Instead, I suggest to keep track of pool registrations within: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Pool/Metrics.hs#L130

rvl: Ah ok I forgot about the Metrics module. I have updated the description in #1090.

KtorZ (approved):

Issue #899, opened 2019-10-23
Adjust restoration loop to look for known certificate registrations 


KtorZ:
# Context

With #898, certificate registrations will now be part of blocks normally processed by the wallet layer.

# Decision

- Adjust restoration loop and keep track of certificates created by our wallet within already existing checkpoints.

# Acceptance Criteria

1. The wallet engine _must_ keep track of all certificates (modulo checkpoints) discovered on the chain as long as their date (as slot) of activation 
2. The delegation status returned by the API _must_ corresponds to the status maintained by the wallet engine (no certificate == not delegating, or, delegating according to the latest certificate).

---

# Development plan

## Generalization of existing discovery classes

PR: #1075 

- [x] Generalize the definition of the `IsOurs` type class so that it can be applied to more types of entities than Address.
- [x] Add an `IsOurs s ChimericAccount` constraint to the `Wallet` type.

## Adjust restoration loop to look for delegations from wallet's chimeric account

PR: #1050 

- [x] Add implementation of `IsOurs (SeqState n k) ChimericAccount`.
- [x] Adjust the return type of `applyBlock` so that it references any stake delegation that has occurred on behalf of the specified wallet.
- [x] Adjust the definition of `restoreBlocks` so that it correctly processes the registrations returned by `applyBlocks`.
- [x] Fix atomiticity issue when inserting delegation certificates in a distinct transaction than checkpoints. 

# PRs

| Number   | Base      |
| ---      | ---       |
| #1075        | `master`  |
| #1050        | `master`  |

# QA 

PR: #1050 

- [x] Add tests to ensure that `applyBlock` does successfully discover delegations made by the wallet under test, and only that wallet. (`STAKE_POOLS_JOIN_01`)
- [x] Add tests to ensure that a wallet's delegation status is correctly updated in the context of discovered delegations. (`STAKE_POOLS_JOIN_01`)

jonathanknowles:
@KtorZ

Our current definition of `Block` in the wallet layer includes a set of delegations from `ChimericAccount` to `PoolId`:

```hs
data Block = Block
    { header  
        :: !BlockHeader   
    , transactions
        :: ![Tx]  
    , delegations 
        :: ![(ChimericAccount, PoolId)]   
    } deriving (Show, Eq, Ord, Generic)
```

The definition of `applyBlock` has access to a `Wallet s`, which implies an `IsOurs s` constraint, which in turns gives us the `isOurs` function:

```hs
[...]

PR #1132, opened 2019-12-06
atomically write checkpoints, tx history and delegation certs when discovered

KtorZ:
# Issue Number

#899

# Overview

- [x] I have adjusted the restoration loop to atomically write checkpoints, tx history and delegation certs when discovered

# Comments


jonathanknowles (approved):
Looks good to me!

Issue #904, opened 2019-10-23
Properly fetch and assign reward account value when fetching wallets


KtorZ:
# Context

Glue code between #903 and #902 

# Decision

- Effectively implement the reward account reporting by fetching the value from the network layer when fetching or listing a wallet.

# Acceptance Criteria

1. The 'balance.reward' of the wallet resource _must_ reflect the actual value for that reward account as known by the node.

---

# Development Plan

1. Forward `getAccountState` of `JormungandrClient` to `NetworkLayer` (easiest option for now)
2. ~use `getAccountState` and `deriveRewardAccount` from `Api.Server`. Maybe add a `getRewards`-helper  in `Cardano.Wallet`.~ The derivation requires the passphrase. We need to store the reward account in the DB.
3. Re-generate faucet addresses be grouped addresses instead of single
4. Test that rewards accumulate after joining the top stake pool in the integration tests

# PR

| Number   | Base      |
| ---      | ---       |
| #1082 |  `master`  |

# QA 

- Quite extensive integration scenario added here testing that rewards are indeed accumulated in the reward account after a certificate is submitted: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs#L213-L264

- Good to know: when a certificate is submitted at epoch `e`, the delegation is only active at epoch `e+2` and, rewards only starts at epoch `e+3` (for the delegator needs to have "helped" the pool with his stake for at least one epoch to get a reward).

Anviking:
I don't see why this would be completely blocked by #899 so I'm starting it.

piotr-iohk:
lgtm.

PR #1137, opened 2019-12-06
Add an extra scenario to illustrate the delegation "latency"

KtorZ:
# Issue Number

#904 

# Overview

- [x] I have added an extra scenario regarding rewards behavior described below

# Comments

At epoch e: Join stake pool P1
At epoch e+1: Quit stake pool P1
At epoch e+2: Stake is being delegated to P1
At epoch e+3: Reward is earned from the delegation in e+2

### DIFF
--- a/lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
+++ b/lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
@@ -275,9 +275,18 @@ spec = do
             [ expectResponseCode HTTP.status202
             ]
 
+        reward <- eventually $ do
+            r <- request @ApiWallet ctx (getWalletEp w) Default Empty
+            verify r
+                [ expectFieldSatisfy balanceReward (> 0)
+                ]
+            pure $ getFromResponse balanceReward r
+
+        waitForNextEpoch ctx
+
         eventually $ do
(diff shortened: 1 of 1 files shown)


lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
```diff
@@ -252,17 +252,32 @@ spec = do
...
+
+        quitStakePool ctx (p1 ^. #id) (w, fixturePassphrase) >>= flip verify
+            [ expectResponseCode HTTP.status202
+            ]
+
+        eventually $ do
+            request @ApiWallet ctx (getWalletEp w) Default Empty >>= flip verify
+                [ expectFieldSatisfy balanceReward (> 0)
+                ]
+
```
piotr-iohk, line 293: Maybe it would be worth also to `waitForNextEpoch ctx` after the last request/assertion and make sure that the rewards are still the same?

KtorZ: Agree.

KtorZ: See https://github.com/input-output-hk/cardano-wallet/pull/1137/commits/48ca2832d29c21729b76cfaf3b36f4453d3dbfd1

piotr-iohk (approved):
:+1:

PR #1139, opened 2019-12-06
Update stack.yaml (add clock-0.8)

angerman:
clock-0.8 finally lifts a bug in the clock package that prevents it from being cross compiled.


KtorZ (approved):
bors r+

PR #1140, opened 2019-12-06
release pipeline revisions

KtorZ:
# Issue Number

# Overview

- Enforce a stronger validation for the release tag (to avoid wrong tags to be pushed by mistakes)
- Add macos and windows artifacts to the release by pulling them from hydra
- Fix race condition between artifacts and doc publication
- Remove the checksums which have little value in the end. If we want to do it properly, we should add GPG signatures.

# Comments

I'd like to run a "mock" release after that just to check it works correctly before Monday.


piotr-iohk (approved):
:heavy_check_mark:

PR #1143, opened 2019-12-07
Use $STRIP instead of hardcoding `strip`

angerman:
Nix will supply the correct strip command via the STRIP env variable. This will include the proper target prefix if needed.


rvl (approved):
Great, thanks!

The [cardano-wallet-jormungandr-linux64](https://hydra.iohk.io/job/Cardano/cardano-wallet-pr-1143/cardano-wallet-jormungandr-linux64/latest) build works now.

bors r+

Issue #1144, opened 2019-12-09
Transactions with minimal amount leave too many fee 


KtorZ:
# Context
 
 

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | 07c0ff4eb  |
| Platform     | All           |
| Installation | N/A             |

# Steps to Reproduce

1. Have a wallet with exactly the right amount to pay for delegation fee (i.e. enough to pay for a certificate, constant fee and a single input).
2. Make a transaction

## Expected behavior

1. The transaction is accepted and the wallet has a balance of 0

## Actual behavior
 

2. The transaction is rejected by the server and it requires an additional amount which matches the fee for an extra input/output.

---

# Resolution 

- Adjusted fee calculation to better handle edge-case ==> #1145
- Use fee estimator instead of hard-coded fee in the integration tests ==> #1145 

---

# QA 

See comments below.

piotr-iohk:
I see that delegation fee is still shown incorrectly. Looking at our own integration tests fees config:
```
fees: 
  certificate: 42
  coefficient: 1
  constant: 14
```
so, if I understand correctly, the fee for delegation fee should be 42 + 14 = 56. (if paid from one utxo)

Delegation fee endpoint shows = 58
Delegating to stake pool takes fee = 58

piotr-iohk:
The same goes for transactions. 
Tx fee (for single utxo tx) is calculated = 17 (instead of 16).

That is for both, making tx and tx fee.

piotr-iohk:
So... tests do not catch that, Perhaps it would be better to have fee in the integration tests being part of hardcoded TestData rather than calculated by feeEstimator?

Refering to this -> https://github.com/input-output-hk/cardano-wallet/pull/1145/commits/3eee4a23f77ca8e99600c62717b5f8c122b5119d

KtorZ:
@piotr-iohk you're forgetting the `coefficient` part. `1` means that each input or output cost `1`. So, in the case of a standard delegation that requires only one UTxO, it'll typically cost `58` because it requires one input (the UTxO) and one change output! 

Only if your input is **exactly** of the right amount (i.e. 57 coins) will you be able to pay `57`, because no change output is needed. `56` will never be possible however because you always need an input.

KtorZ:
We have integration tests testing for these edge-cases (sorry, I see the QA section on this ticket is empty :| ... ), let me find that back for you.

KtorZ:
So, 

[...]

PR #1145, opened 2019-12-09
Use fee estimator instead of hard-coded fee value for stake pools joining

KtorZ:
# Issue Number

#1144

# Overview

- [x] I have removed the hard-coded `stakePoolFee` and used a proper fee estimator which computes expected fee from a transaction description and values from the genesis block.
- [x] I have added some non-null fees for the "coefficient" part of the fee policy, so that inputs/outputs would also cost something.
- [x] I have fixed fee calculation at the boundaries, especially, when the reduction of change outputs modifies the fees. 
- [x] I have fixed the external tx builder to not use hard-coded fees but instead, use fees from the genesis file

# Comments

~~- The behavior and amounts in most tests are currently wrong, I'll fix in another PR.~~ Done.


piotr-iohk (approved):
:+1:

KtorZ:
nix _"reliable builds"_ failing again with:

> /var/lib/buildkite-agent/.stack/setup-exe-cache/x86_64-linux-nix/Cabal-simple_mPHDZzAJ_2.4.0.1_ghc-8.6.5: error while loading shared libraries: libffi.so.6: cannot open shared object file: No such file or directory

KtorZ:
bors retry

PR #1151, opened 2019-12-09
update compatibility matrix in README

KtorZ:
# Issue Number

N/A

# Overview

- [ ] I have updated compatibility matrix in README

# Comments

Not quite sure about what we should put as a compatible Jörmungandr's version. I've put rc10 initially because there is no any breaking changes announced compared to rc8 apart from the configuration file... But we haven't tested anything with rc9, nor rc9+1 nor rc10, so I should probably revert lower that down to `rc8` even if, in principle, it should work with the subsequent versions.

@piotr-iohk, thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,11 +45,11 @@ See **Installation Instructions** for each available [release](https://github.co
 
 > ### Latest releases
 > 
-> | cardano-wallet                                                                            | Jörmungandr                                                                          |
+> | cardano-wallet                                                                            | Jörmungandr (compatible versions)                                                    |
 > | ---                                                                                       | ---                                                                                  |
+> | [v2019-12-09](https://github.com/input-output-hk/cardano-wallet/releases/tag/v2019-12-09) | [v0.8.0-rc8](https://github.com/input-output-hk/jormungandr/releases/tag/v0.8.0-rc8) |
 > | [v2019-11-18](https://github.com/input-output-hk/cardano-wallet/releases/tag/v2019-11-18) | [v0.7.0](https://github.com/input-output-hk/jormungandr/releases/tag/v0.7.0)         |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,11 +45,11 @@ See **Installation Instructions** for each available [release](https://github.co
...
-> | [v2019-11-18](https://github.com/input-output-hk/cardano-wallet/releases/tag/v2019-11-18) | [v0.7.0](https://github.com/input-output-hk/jormungandr/releases/tag/v0.7.0)         |
-> | [v2019-11-14](https://github.com/input-output-hk/cardano-wallet/releases/tag/v2019-11-14) | [v0.7.0](https://github.com/input-output-hk/jormungandr/releases/tag/v0.7.0)         |
-> | [v2019-11-07](https://github.com/input-output-hk/cardano-wallet/releases/tag/v2019-11-07) | [v0.7.0-rc7](https://github.com/input-output-hk/jormungandr/releases/tag/v0.7.0-rc7) |
+> | cardano-wallet                                                                            | Jörmungandr (compatible versions)                                                      |
+> | ---                                                                                       | ---                                                                                    |
+> | [v2019-12-09](https://github.com/input-output-hk/cardano-wallet/releases/tag/v2019-12-09) | [v0.8.0-rc10](https://github.com/input-output-hk/jormungandr/releases/tag/v0.8.0-rc10) |
```
piotr-iohk: are we sure?

KtorZ: Hence my comment on the PR :/ 
I've reverted to `rc8`, it's a safer assumption. rc10 doesn't mention any _breaking changes_ but we have so-to-speak, only tested with `rc8`.

piotr-iohk: ok, missed the comment. Yeah, better revert to rc8.

piotr-iohk (approved):

README.md
piotr-iohk: We can update if we manage to verify rc10...

Issue #1092, opened 2019-11-29
Update stake pool registrations in the restore loop


rvl:
# Context

In order to provide stake pool metadata, the wallet must track the mapping from `PoolId` to `PoolOwner`.

# Decision

Combine the results of #1090 and #1091.

When processing filtered blocks in the stake pool restore loop, update the `pool_owner` table.

# Acceptance Criteria

1. The `pool_owner` table must reflect the stake pool registrations and retirements received in the jormungandr block fragments. (i.e. the wallet must track stake pool registrations and their owners).

---

# Development 

1. As above ⇒ PR #1154.
2. Unit testing is tricky because the logic is locked inside `monitorStakePools`.

# QA 

There is a property test in `lib/core/test/unit/Cardano/Pool/MetricsSpec.hs`.

The integration test will come with #1065.

piotr-iohk:
lgtm

PR #1154, opened 2019-12-10
Update registration ownership info in monitorStakePools

rvl:
Relates to #1092.

# Overview

This modifies the stake pool network follower loop so that pool registrations are tracked in the database.

Tested with a unit test that mocks the network layer.


KtorZ (approved):

KtorZ (approved):
@rvl I've completed the unit test but will leave the review to you. 
In the end, I am not sure it's buying us much more than what the db properties already buy us in terms of testing.. 

Still, I carried on because this test setup could open some nice possibilities to also test other stuff like the metrics.

KtorZ:
bors try

rvl:
Thanks @KtorZ.

> In the end, I am not sure it's buying us much more than what the db properties already buy us in terms of testing..

I thought the same thing. There is only a small amount of logic in `monitorStakePools` that is not covered by unit tests. But it needs to be unit tested somehow.

I have added an extra assertion for to the test case that stake pool registration certificates are logged.

lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -169,9 +169,9 @@ monitorStakePools tr nl DBLayer{..} = do
         liftIO $ logInfo tr $ "Writing stake-distribution for epoch " <> pretty ep
 
         let registrations = concatMap poolRegistrations blocks
-        unless (null registrations) $
-            liftIO $ logInfo tr $ "Discovered stake pool registrations: "
-                <> pretty registrations
+        liftIO $ forM_ registrations $ \registration ->
+            logInfo tr $ "Discovered stake pool registration: "
+                <> pretty registration
```
jonathanknowles, line 174: :+1:

lib/core/test/unit/Cardano/Pool/MetricsSpec.hs
```diff
@@ -288,6 +312,9 @@ prop_trackRegistrations test = monadicIO $ do
                 pure (0, mempty)
             , networkTip =
                 pure header0
+            , staticBlockchainParameters =
+                (block0, mockBlockchainParameters
+                    { getEpochStability = Quantity 42 })
```
jonathanknowles: Question: where does the number 42 come from, and does it have a special meaning?

Even if it's just an arbitrary number, perhaps adding a named constant will help future readers to know that it's just an arbitrary number? (For example, `arbitraryEpochStability` or `mockEpochStability`.)

jonathanknowles (approved):
I reviewed the following changes:
* https://github.com/input-output-hk/cardano-wallet/pull/1154/commits/0779895d646da44669f19fab366a9fbd33e50a52
* https://github.com/input-output-hk/cardano-wallet/pull/1154/commits/e5fe9263cca9cd87e3fff0a085e2ed6286cc0958
* https://github.com/input-output-hk/cardano-wallet/pull/1154/commits/9851083053268eb38e2caa531225214be52c8c80

LGTM!

Though it'd be great to have a named constant for the magic number `42`. (See: https://github.com/input-output-hk/cardano-wallet/pull/1154/commits/9851083053268eb38e2caa531225214be52c8c80#diff-baca5dbe242634249ae8122255a26866R317)

lib/core/test/unit/Cardano/Pool/MetricsSpec.hs
rvl: OK fixed.

rvl:
bors merge

PR #1156, opened 2019-12-10
Expect 200 for fee-estimation

Anviking:
# Issue Number

#1096 

# Overview

- [x] I have added a `expectResponseCode HTTP.status200` for *STAKE_POOLS_ESTIMATE_FEE_01*
- [x] I tweaked the description of *STAKE_POOLS_ESTIMATE_FEE_04*

# Comments

### DIFF
--- a/lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
+++ b/lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
@@ -548,9 +548,12 @@ spec = do
         (_, p:_) <- eventually $
             unsafeRequest @[ApiStakePool] ctx listStakePoolsEp Empty
         w <- fixtureWallet ctx
-        fee <- getFromResponse amount <$> delegationFee ctx w
-        r <- joinStakePool ctx (p ^. #id) (w, fixturePassphrase)
+        r <- delegationFee ctx w
         verify r
+            [ expectResponseCode HTTP.status200
+            ]
+        let fee = getFromResponse amount r
+        joinStakePool ctx (p ^. #id) (w, fixturePassphrase) >>= flip verify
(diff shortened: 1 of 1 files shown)


lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
```diff
@@ -572,7 +573,7 @@ spec = do
             , expectErrorMessage $ errMsg404NoWallet (w ^. walletId)
             ]
 
-    describe "STAKE_POOLS_ESTIMATE_FEE_04 - wallet ids" $ do
+    describe "STAKE_POOLS_ESTIMATE_FEE_04 - invalid wallet ids" $ do
```
Anviking: So the wallet ids are either for a non-existing wallet, or totally wrong.

Maybe "false wallet ids" would be better (consistent with the `falseWalletIds` name) 🤷‍♂

piotr-iohk:
swagger.yaml says it should be `200` -> https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/getDelegationFee

I think that `200` suits better and is consistent with other endpoints that estimate fee (transactions and migrations).

Anviking:
@piotr-iohk Thanks, good point, actually I was testing the response code on the completely wrong call.

The `delegationFee` is already returning 200. It is `joinStakePool` that returns 202.

I removed the 202 check on `joinStakePool` (it is tested in its own test), and added a 200 check for `delegationFee`.

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
Anviking: Went with "false wallet ids".

KtorZ (approved):
bors r+

PR #1157, opened 2019-12-10
Misc. nix improvements

### LINKED ISSUE #1138: Why fix the compiler *and* the LTS in the stack.yaml?
as lts's are compiler dependent, and hence bring in a compiler and with it all its boot/base packages, why specify the compiler as well as the lts?

rvl:
# Overview

- Remove some unused nix code.
- Use a library function from iohk-nix rather than a local copy.
- Remove compiler version from `stack.yaml` (fixes #1138).
- Really fix `Test.Utils.Paths` this time, so that it also works under nix-shell.


KtorZ (approved):
LGTM for the Haskell part.

hamishmack (approved):
Looks good

Issue #1065, opened 2019-11-25
listStakePools - Include metadata in API responses


rvl:
# Context

Part of [WB-25 Collect stake pool metadata](https://jira.iohk.io/projects/WB/issues/WB-25).

Extend the response of `listStakePools` with a `metadata` matching those from the registry (minus the `id`).
For known stake pools which aren't present in the registry the metadata entry should be `Nothing` / `null`.

# Decision

Add a function to look up metadata by `PooId`. This will first need to find the pool owners of the known stake pools, then use `getStakePoolMeta` to find metadata from the owners.

Adjust `Cardano.Wallet.Api.Server.listPools` to  augment the response with this information.

### Integration tests

Integration tests ought to be configured with an alternative URL, so that they are self-contained. This testing file can be served up with [`Network.Wai.Application.Static`](http://hackage.haskell.org/package/wai-app-static-3.1.6.3/docs/Network-Wai-Application-Static.html#v:staticApp).

The `master.zip` served by the integration tests should correspond with our stake-pools in test setup, but that also covers following cases:
 - data of the stake pool that does not exist in our integration setup
 - missing data for one of the stake pools that we have in the integration setup (for which stake-pool the metadata returned by the request would be null)

The case that `master.zip` is not available at the location should be covered by integration tests.

# Acceptance Criteria

1. Available metadata about stake pools must be included with the API response.
2. Integration tests should not require downloading from external URLs.
3. Conflicting metadata from multiple owners must result in no metadata being presented by cardano-wallet.

---

# Development 

1. Add the logic described aboce to `StakePoolLayer.listStakePools` to keep the web handler clean.
1. Update/add unit tests of listStakePools to include metadata.
1. Allow environment variable override of stake pool registry URL for integration (and other) testing.
1. Integration test fixtures will need stake pool registrations added, if not already done.
1. Update/add integration tests of listStakePools.
1. Print the registry in use at start-up ==> TODO

# QA 

1. The `CARDANO_WALLET_STAKE_POOL_REGISTRY_URL` environment variable can be set for ad-hoc manual testing of stake pool listing.
2. The URL of the registry in use is logged when the metadata is downloaded.
3. There are unit tests added in `lib/core/test/unit/Cardano/Pool/MetricsSpec.hs`. These test `associateMetadata`, which is the guts of the logic for adding metadata. Regrettably, this is not a proper unit test of `Cardano.Pool.Metrics.listStakePools`, but more like testing of the implementation.
[...]

Issue #1158, opened 2019-12-10
Apparent performance could be wrong for brief times (because of multiple sources of truth)


Anviking:
# Context

The `Pool.Metrics` chain-follower ensures a source of stake distributions and pool productions that are consistent with each other (always on the same fork).

The performance calculation is:
Apparent Performance = (n / N) * (S / s)
where
- n is the number of blocks produced by the pool in the last epoch
- N is the number of slots in the epoch
- s is the stake owned by the pool in the last epoch
- S is the total stake delegated to pools in the last epoch

In the implementation however,
https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Pool/Metrics.hs#L239

the tip used to calculate `N` for the current epoch is retrieved separately from the node.

Because it is a different source of data from the DB, it is not guaranteed to be on the same chain as the DB.

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      |  |
| Platform     |           |
| Installation |             |

# Steps to Reproduce

Not tested:

Consider the scenario:
1. Slots per epoch = 20, poolA has 100% stake
2. stake-pool-worker syncs to slot `2.3`
3. node rolls forward to `2.19`
4. user calls `listStakePools` without the worker yet noticing the roll-foward

## Expected behavior

Performance is calculated from consistent sources of data.

Likely we'd expect:
For epoch 2:
s / S = 1
N = 3
n = 3

AP = 100%
## Actual behavior
 

Hypothetical, not tested:

For epoch 2:
s / S = 1
N = 19
n = 3

AP = 16%

(As the chain follower catches up, the inconsistency would disappear)
---

# Resolution 

---

We could calculate the `N` using the tip from the Pool DB instead of the tip from the node's REST api. ==>  #1168

# QA 

- Added some unit test which checks that the tip used for the performance calculation (which happen to be the reported when the pools aren't in sync) is the right one, and not the network's one:   https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Cardano/Pool/MetricsSpec.hs#L153-L155

PR #1159, opened 2019-12-11
listStakePools - Include metadata in API responses

rvl:
Relates to #1065.

# Overview

- Modified Api.Server.listPools and StakePoolLayer listStakePools to include metadata.
- Made it possible to use an environment variable for the registry url.
- Handles the case where different stake pool owners register different metadata by logging and then ignoring all inconsistent data.

# Comments

- This is a draft.


lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -261,6 +266,13 @@ findArchiveFile repoPath = find isRepoFile . Map.keys <$> getEntries
 registryFile :: PoolOwner -> FilePath
 registryFile owner_ = "registry" </> T.unpack (toText owner_) <.> "json"
 
+-- | Returns the Cardano Foundation stake pool registry zipfile URL, or the
+-- @CARDANO_WALLET_STAKE_POOL_REGISTRY_URL@ environment variable if it is set.
+getRegistryZipUrl :: IO String
+getRegistryZipUrl = fromMaybe cardanoFoundationRegistryZip <$> lookupEnv var
+  where
+    var = "CARDANO_WALLET_STAKE_POOL_REGISTRY_URL"
```
KtorZ: :+1:

lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -211,12 +231,20 @@ data ErrListStakePools
...
-     -> Trace IO Text
-     -> StakePoolLayer IO
+    :: DBLayer IO
+    -> NetworkLayer IO t block
+    -> Trace IO StakePoolLayerMsg
+    -> StakePoolLayer IO
 newStakePoolLayer db@DBLayer{..} nl tr = StakePoolLayer
     { listStakePools = do
+        lift $ logTrace tr MsgListStakePoolsBegin
+        stakePools <- rankKnownPools
```
KtorZ: **[remark]** 

You have used to forbidden word (i.e. `rank`) :scream: 
It makes things clearer though :+1:

lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -238,8 +266,25 @@ newStakePoolLayer db@DBLayer{..} nl tr = StakePoolLayer
...
+    findMetadata poolIds = do
+        owners <- atomically $ mapM readStakePoolOwners poolIds
+        -- note: this will become simpler once we cache metadata in the database
+        let owners' = nub $ concat owners
+        url <- getRegistryZipUrl
+        let tr' = contramap (fmap MsgRegistry) tr
+        getStakePoolMetadata tr' url owners' >>= \case
+            Left _ -> do
+                logTrace tr MsgMetadataUnavailable
+                pure $ replicate (length poolIds) Nothing
```
KtorZ, line 304: **[remark]**

Isn't `trace` maybe too low for this kind of error? We are not really expecting this to be unavailable, are we? 
I'd rather go for `Warning` or `Notice` for this one. Though `Warning` is maybe too strong because there is really nothing we can do about it.

rvl: The log level is set in `instance DefineSeverity StakePoolLayerMsg`. I had made it Warning but actually now that you mention it, Notice is more suitable.

KtorZ: Ho! Yes. I missed that sorry, got confused with `logTrace` :+1:

lib/core/test/unit/Cardano/Pool/MetricsSpec.hs
```diff
@@ -485,3 +497,69 @@ arbitraryChunks xs = do
...
+    it "works in sunny day case" $ do
+        let owner = PoolOwner "a"
+        let md = mkMetadata owner "AAAA"
+        let pid = PoolId "1"
+
+        let res = associateMetadata
+                (Map.fromList [(pid, [owner])])
+                [(owner, Just md)]
+
+        res `shouldBe` [(MsgMetadataUsing pid owner md, Just md)]
```
KtorZ, line 525: :+1:

lib/core/test/unit/Cardano/Pool/MetricsSpec.hs
```diff
@@ -485,3 +497,69 @@ arbitraryChunks xs = do
...
+
+    it "missing metadata" $ do
+        let res = associateMetadata
+                (Map.fromList [(PoolId "1", [PoolOwner "a"])])
+                [(PoolOwner "a", Nothing)]
+
+        res `shouldBe` [(MsgMetadataMissing (PoolId "1"), Nothing)]
```
KtorZ, line 532: :+1:

lib/core/test/unit/Cardano/Pool/MetricsSpec.hs
```diff
@@ -485,3 +497,69 @@ arbitraryChunks xs = do
...
+
+        case msg of
+            MsgMetadataUsing pid owner md -> do
+                pid `shouldBe` PoolId "1"
+                [PoolOwner "a", PoolOwner "b"] `shouldContain` [owner]
+                md `shouldSatisfy` sameStakePoolMetadata mda
+                md `shouldSatisfy` sameStakePoolMetadata mdb
+            _ -> fail $ "wrong log message " ++ show msg
+
+        fmap (sameStakePoolMetadata mda) res `shouldBe` Just True
```
KtorZ, line 551: :+1:

KtorZ (approved):

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
```diff
@@ -170,6 +183,30 @@ spec = do
             expectResponseCode @IO HTTP.status405 r
             expectErrorMessage errMsg405 r
 
+    it "STAKE_POOLS_LIST_04 - Discovers new pools when they are registered" $ \ctx -> do
```
KtorZ, line 186: @rvl I added a rather extensive integration test. The setup behind `registerStakePool` is a bit complex because we have to do everything with `jcli` and use accounts to pay for the registration certificate! I got to the end of it and.. nicely caught a small bug in the "zipping logic" from the `listStakePools` endpoint (cf my last commit). 
Please have another round of review / fixes and let's get this merged! 

Some outstanding points remaining after that:

- Maybe do some caching? I expect this endpoint to be quite heavily hammered by clients like Daedalus. The registry won't change very often. Renewing a cache every hour or so should also mitigate the concern. 
- Currently the handler for `joinStakePool` relies on `listStakePool` to get a list of known stake pool... that's kinda lame because `listStakePool` is a very heavy operation and, all stake pools ids can be known by simply making a database call. So we should fix that.
- https://github.com/input-output-hk/cardano-wallet/issues/1158

KtorZ (approved):

KtorZ:
bors try

Issue #1088, opened 2019-11-29
Add `nextEpoch` field to `ApiNetworkInformation`


jonathanknowles:
# Context

In order to support the ability for users to query the API about the _next epoch_, we need to extend the existing `getNetworkInformation` endpoint so that it provides this information.

# Decision

Add a `nextEpoch` field to the existing `ApiNetworkInformation` type, so that the definition becomes:

```hs
data ApiNetworkInformation = ApiNetworkInformation
    { networkTip :: !ApiNetworkTip
    , nextEpoch :: !ApiEpochInfo
    , nodeTip :: !ApiBlockReference   
    , syncProgress :: !(ApiT SyncProgress)
    } deriving (Eq, Generic, Show)
```
The associated swagger definition should also be updated so that the definition becomes:
```yaml
ApiNetworkInformation: &ApiNetworkInformation
  type: object
  required: 
    - sync_progress
    - node_tip
    - network_tip
    - next_epoch
  properties: 
    sync_progress: *networkInformationSyncProgress
    node_tip: *networkInformationNodeTip
    network_tip: *networkInformationNetworkTip
    next_epoch: *networkInformationNextEpoch
```

# Acceptance Criteria

- [x] The `ApiNetworkInformation` value returned by the `getNetworkInformation` endpoint _must_ contain a `nextEpoch` field of type `ApiEpochInfo`.
 
# Development 

- [x] Add wrapper function `calculateNextEpoch` around `Primitive.Types.epochCeiling` that computes the next epoch after the _current time_, based on the current `SlotParameters`.
- [x] Extend the `ApiNetworkInformation` type to include the `nextEpoch` field.
- [x] Extend function `network` in `Api.Server` to call the `calculateNextEpoch` wrapper function.

# QA 

- [x] Adjust integration tests to check that the `getNetworkInformation` endpoint returns the correct information within the `nextEpoch` field.

# PRs

| Number | Base |
| -- | -- |
| #1122 | `master` |

paweljakubas:
Task goals/development plan looking good and clear. Nevertheless, I would give here `2` estimate due to integration test to be added

piotr-iohk:
Additional test -> #1162.
I was also wondering -> https://github.com/input-output-hk/cardano-wallet/pull/1162#discussion_r356563598

Also played around manually on testnet and calculated nextEpoch datetime is correct from observations.

piotr-iohk:
lgtm

PR #1162, opened 2019-12-11
additional test for next epoch num

piotr-iohk:
# Issue Number

#1088

# Overview

- [ ] I have added additional check for the nextEpochNum in the integration test
- [ ] added additional test: Calculated next epoch no is the next epoch no

# Comments

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
@@ -20,13 +20,14 @@ import Control.Monad.IO.Class
 import Data.Time.Clock
     ( getCurrentTime )
 import Test.Hspec
-    ( SpecWith, describe, it )
+    ( SpecWith, describe, it, shouldBe )
 import Test.Integration.Framework.DSL
     ( Context (..)
     , Headers (..)
     , Payload (..)
     , emptyByronWallet
     , emptyWallet
+    , eventually
     , eventually_
     , expectErrorMessage
     , expectEventually'
@@ -51,7 +52,7 @@ import qualified Network.HTTP.Types.Status as HTTP
 spec :: forall t. SpecWith (Context t)
 spec = do
     it "NETWORK - Can query network information" $ \ctx -> do
-        eventually_ $ do
+        r <- eventually $ do
             now <- liftIO getCurrentTime
             r <- request @ApiNetworkInformation ctx networkInfoEp Default Empty
             expectResponseCode @IO HTTP.status200 r
@@ -59,6 +60,13 @@ spec = do
                 [ expectFieldSatisfy nextEpoch ((> now) . epochStartTime)
                 , expectFieldEqual syncProgress Ready
                 ]
+            return r
+
+        let currentEpochNum =
+                getFromResponse (#networkTip . #epochNumber . #getApiT) r
+        let nextEpochNum =
+                getFromResponse (#nextEpoch . #epochNumber . #getApiT) r
+        nextEpochNum `shouldBe` (currentEpochNum + 1)
 
     it "NETWORK_SHELLEY - Wallet has the same tip as network/information" $
         \ctx -> do
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
@@ -43,6 +43,7 @@ import Test.Integration.Framework.DSL
     , state
     , syncProgress
     , verify
+    , waitForNextEpoch
     )
 import Test.Integration.Framework.TestData
     ( errMsg405, getHeaderCases )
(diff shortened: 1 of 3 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
```diff
@@ -51,14 +53,32 @@ import qualified Network.HTTP.Types.Status as HTTP
 spec :: forall t. SpecWith (Context t)
 spec = do
     it "NETWORK - Can query network information" $ \ctx -> do
-        eventually_ $ do
+        r <- eventually $ do
             now <- liftIO getCurrentTime
             r <- request @ApiNetworkInformation ctx networkInfoEp Default Empty
             expectResponseCode @IO HTTP.status200 r
             verify r
                 [ expectFieldSatisfy nextEpoch ((> now) . epochStartTime)
```
piotr-iohk, line 61: I am not sure if there can be more strict to way to check this... (To make sure that `epochStartTime` is really the the exact time new epoch starts)

lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
```diff
@@ -51,14 +53,32 @@ import qualified Network.HTTP.Types.Status as HTTP
...
                 ]
+            return r
+
+        let currentEpochNum =
+                getFromResponse (#networkTip . #epochNumber . #getApiT) r
+        let nextEpochNum =
+                getFromResponse (#nextEpoch . #epochNumber . #getApiT) r
+        nextEpochNum `shouldBe` (currentEpochNum + 1)
+
+
```
jonathanknowles: Suggestion: use just a single line of blank space here.

https://github.com/input-output-hk/cardano-wallet/wiki/Coding-Standards#use-only-a-single-blank-line-between-top-level-definitions

lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
```diff
@@ -51,14 +53,32 @@ import qualified Network.HTTP.Types.Status as HTTP
...
                 [ expectFieldSatisfy nextEpoch ((> now) . epochStartTime)
                 , expectFieldEqual syncProgress Ready
                 ]
+            return r
+
+        let currentEpochNum =
+                getFromResponse (#networkTip . #epochNumber . #getApiT) r
+        let nextEpochNum =
+                getFromResponse (#nextEpoch . #epochNumber . #getApiT) r
+        nextEpochNum `shouldBe` (currentEpochNum + 1)
```
jonathanknowles: :+1: 

We could drop the parentheses here:
```suggestion
        nextEpochNum `shouldBe` currentEpochNum + 1
```

jonathanknowles (approved):
Looks good to me!

I have one suggestion: (which could be the subject of a **further PR**)

**Suggestion:**

We could add a test that:
1. fetches a fresh value **`info1`** of type `ApiNetworkInformation`.
2. gets time value **`t`** from `epochStartTime (nextEpoch info1)`;
3. waits until time **`t`**; (for this, we'll need a function like `waitUntil :: UTCTime -> IO ()`).
4. fetches a fresh value **`info2`**  of type `ApiNetworkInformation`;
5. checks that `epochNo (nextEpoch info2) == epochNo (nextEpoch info1) + 1`; 
6. checks that `epochStartTime (nextEpoch info2) > epochStartTime (nextEpoch info1)`.

**Alternative:**

We could adjust the existing test (that uses `waitForNextEpoch`) to check that we have indeed passed the predicted time **`t`** while waiting.

piotr-iohk:
> We could adjust the existing test (that uses waitForNextEpoch) to check that we have indeed passed the predicted time t while waiting.

Went for that :point_up_2:  for now. Thx!
Also addressed other review remarks.

piotr-iohk:
Made `nextEpochStartTime <= now` instead of `nextEpochStartTime < now` to fix a race condition in test.

bors r+

lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
```diff
@@ -51,14 +53,34 @@ import qualified Network.HTTP.Types.Status as HTTP
...
+        let currentEpochNum =
+                getFromResponse (#networkTip . #epochNumber . #getApiT) r
+        let nextEpochNum =
+                getFromResponse (#nextEpoch . #epochNumber . #getApiT) r
+        nextEpochNum `shouldBe` currentEpochNum + 1
+
+    it "NETWORK - Calculated next epoch is the next epoch" $ \ctx -> do
+        r1 <- request @ApiNetworkInformation ctx networkInfoEp Default Empty
+        let calculatedNextEpoch = getFromResponse (#nextEpoch . #epochNumber) r1
+        let nextEpochStartTime = getFromResponse (#nextEpoch . #epochStartTime) r1
```
jonathanknowles, line 75: Perhaps add the following here:
```hs
        now <- liftIO getCurrentTime
        nextEpochStartTime > now `shouldBe` True
```

lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
```diff
@@ -51,14 +53,34 @@ import qualified Network.HTTP.Types.Status as HTTP
...
+
+        waitForNextEpoch ctx
+        now <- liftIO getCurrentTime
+        nextEpochStartTime <= now `shouldBe` True
+
+        r2 <- request @ApiNetworkInformation ctx networkInfoEp Default Empty
+        let currentEpoch = getFromResponse (#networkTip . #epochNumber) r2
+        currentEpoch `shouldBe` calculatedNextEpoch
```
jonathanknowles, line 83: Perhaps add a check here to verify that `nextEpoch.epochStartTime` has indeed changed from the previous value?

lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
piotr-iohk: I think I haven't added it because it is checked in previous test case.

lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
piotr-iohk: That could be added.

Issue #1136, opened 2019-12-06
Provide implementation of `selectCoins` API function.


jonathanknowles:
# Context

In order to support users that wish to use a hardware wallet, we would like to expose our coin selection functionality as a new API function `selectCoins`. This function will take a list of payments as its input, and produce a coin selection as its output.

# Decision

Provide a complete server implementation of the skeleton endpoint created in #1135.

Ensure that the following error codes are handled correctly:

```yaml
x-responsesSelectCoins: &responsesSelectCoins
  <<: *responsesErr400
  <<: *responsesErr403
  <<: *responsesErr404
  <<: *responsesErr405
  <<: *responsesErr406
  <<: *responsesErr415
```

# Acceptance Criteria

- [ ] The endpoint _must_ produce a valid coin selection, when given a valid input.
- [ ] The endpoint _must_ correctly handle all of the error conditions indicated in the Swagger specification.

---

# Development 

# QA

KtorZ:
No QA nor Development section here, but deducing from other tickets, I've found:

https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs#L1483-L1518

We are missing:

- Negative tests for the headers and path parameters
- Swagger file description to "stable", putting back "In Progress"

PR #1165, opened 2019-12-12
Provide implementation of `selectCoins` API function.

jonathanknowles:
# Issue Number

#1136 

# Overview

This PR:

- [x] Provides an implementation of the `selectCoins` API function.
- [x] Provides an integration test to verify that a singleton payment is included in the output.
- [x] Provides an integration test to verify that multiple payments are included in the output.
- [x] Adds an `UnsignedTx` type to capture the concept of a set of inputs and outputs that are ready to be signed into a real `Tx`.

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -80,6 +80,7 @@ module Cardano.Wallet
     , ErrFetchRewards (..)
 
     -- ** Address
+    , assignChangeAddressesAndCheckpoint
     , listAddresses
 
     -- ** Payment
@@ -984,6 +985,43 @@ signPayment ctx wid argGenChange pwd (CoinSelection ins outs chgs) = db & \DBLay
   where
     db = ctx ^. dbLayer @s @k
     tl = ctx ^. transactionLayer @t @k
+
+-- | Augments the given set of outputs with additional outputs for each of the
+--   given change outputs.
+--
+-- This function associates each change value with a target address belonging
+-- to the given wallet, updates the wallet state to keep track of pending
+-- change addresses, and writes a database checkpoint to record the updated
+-- state.
+--
+-- See 'assignChangeAddresses'.
+--
+assignChangeAddressesAndCheckpoint
+    :: forall ctx s k.
+        ( GenChange s
+        , HasDBLayer s k ctx
+        , IsOwned s k
+        , NFData s
+        , Show s
+        )
+    => ctx
+    -> WalletId
+    -> ArgGenChange s
+    -> [TxOut]
+    -> [Coin]
+    -> ExceptT ErrNoSuchWallet IO [TxOut]
+assignChangeAddressesAndCheckpoint ctx wid argGenChange outputs change =
+    db & \DBLayer{..} ->
+        mapExceptT atomically $ do
+            cp <- withNoSuchWallet wid
+                $ readCheckpoint
+                $ PrimaryKey wid
+            (allOuts, s') <- assignChangeAddresses
+                argGenChange outputs change (getState cp)
+            putCheckpoint (PrimaryKey wid) (updateState s' cp)
+            pure allOuts
+  where
+    db = ctx ^. dbLayer @s @k
 
 signDelegation
     :: forall ctx s t k.
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -224,6 +224,7 @@ import Cardano.Wallet.Primitive.Types
     , TxOut (..)
     , TxStatus (..)
     , UTxOStatistics
+    , UnsignedTx (..)
     , WalletDelegation (..)
     , WalletId (..)
     , WalletMetadata (..)
@@ -1003,10 +1004,7 @@ selectCoinsExternal
     -> WalletId
     -> ArgGenChange s
     -> NonEmpty TxOut
-    -> ExceptT (ErrSelectCoinsExternal e) IO
-        ( NonEmpty (TxIn, TxOut)
-        , NonEmpty TxOut
-        )
+    -> ExceptT (ErrSelectCoinsExternal e) IO UnsignedTx
 selectCoinsExternal ctx wid argGenChange payments = do
     CoinSelection mInputs mPayments mChange <-
         withExceptT ErrSelectCoinsExternalUnableToMakeSelection $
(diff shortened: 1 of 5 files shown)


lib/core/src/Cardano/Wallet.hs
```diff
@@ -985,6 +986,43 @@ signPayment ctx wid argGenChange pwd (CoinSelection ins outs chgs) = db & \DBLay
...
+--   given change outputs.
+--
+-- This function associates each change value with a target address belonging
+-- to the given wallet, updates the wallet state to keep track of pending
+-- change addresses, and writes a database checkpoint to record the updated
+-- state.
+--
+-- See 'assignChangeAddresses'.
+--
+assignChangeAddressesAndCheckpoint
```
KtorZ: **[change]**

Rename this to (for example) `selectCoinsForExternal` and move the logic from the API handler to here.

**Why?** 

Having to define errors type for the Api handlers is a good sign that some logic is in the wrong place. We should keep the API handlers simple and move as much as the wallet logic down to this module. The API layer should only compose functions from the wallet engine. Plus, the name `assignChangeAddressesAndCheckpoint` already suggest that something seems wrong (everytime one is tempted to put `And` in the name of a function is a good sign that this function is probably not operating at the right level of abstraction).

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -622,14 +625,43 @@ coinSelections =
...
+    CoinSelection mInputs mOutputs mChange <-
+        liftHandler $ withWorkerCtx ctx wid liftE $ \wrk ->
+            W.selectCoinsForPayment @_ @s @t wrk wid $
+                coerceCoin <$> (body ^. #payments)
+    mkApiCoinSelection
+        <$> (ensureNonEmptyInputs mInputs)
+        <*> (ensureNonEmptyOutputs =<<
+                liftHandler (withWorkerCtx ctx wid throwE $ \wrk ->
+                    W.assignChangeAddressesAndCheckpoint
+                        @_ @s @k wrk wid () mOutputs mChange))
```
KtorZ: **[suggestion]**

Swap the use of applicative for a monadic approach. 

```hs
let payments = coerceCoin <$> (body ^. #payments)
CoinSelection mInputs mRecipients mChange <- liftHandler 
    $ withWorkerCtx ctx wid liftE 
    $ \wrk -> W.selectCoinsForPayment @_ @s @t wrk wid payments

mOutputs  <- liftHandler 
    $ withWorkerCtx ctx wid throwE 
    $ \wrk -> assignChangeAddressesAndCheckpoint @_ @s @k wrk wid () mRecipients mChange

neInputs  <- ensureNonEmpty ErrSelectCoinsUnableToAssignInputs mInputs
neOutputs <- ensureNonEmpty ErrSelectCoinsUnableToAssignOutputs mOutputs

pure $ mkApiCoinSelection neInputs neOutputs

```

**Why?**

(In this case) It leads to more readable code without changing much of the semantic.

jonathanknowles: Hi @KtorZ. Thanks for this suggestion: I agree this could be more readable.

I've made the following change, in response to your suggestion: https://github.com/input-output-hk/cardano-wallet/pull/1165/commits/c9a6b065b733612b876860ed2b2aea928fd7b549.

lib/core/src/Cardano/Wallet.hs
jonathanknowles: > Rename this to (for example) selectCoinsForExternal and move the logic from the API handler to here.

Fair enough! I've made the following change to implement your suggestion: https://github.com/input-output-hk/cardano-wallet/pull/1165/commits/62b45c411171f5e91ea044f50e48144596faf78b.

Of course, let me know if this is not what you had in mind.

As a result of the above change, the tuple `(NonEmpty (TxIn, TxOut), NonEmpty TxOut)` appears in the signatures of both the following functions:

* `selectCoinsForExternal`
* `mkApiCoinSelection`

To simplify this, I've made a further commit which replaces this tuple with a dedicated type, to capture the concept of an _**unsigned transaction**_: https://github.com/input-output-hk/cardano-wallet/pull/1165/commits/09ce374339f33cb9f711ff7804738313f1bb3c82.

Let me know if this is suitable. If not, I'd be happy to revise it further. Thanks in advance!

lib/core/src/Cardano/Wallet.hs
```diff
@@ -985,6 +988,55 @@ signPayment ctx wid argGenChange pwd (CoinSelection ins outs chgs) = db & \DBLay
...
+        withExceptT ErrSelectCoinsExternalNoSuchWallet $
+            mapExceptT atomically $ do
+                cp <- withNoSuchWallet wid $ readCheckpoint $ PrimaryKey wid
+                (mOutputs, s') <- assignChangeAddresses
+                    argGenChange mPayments mChange $ getState cp
+                putCheckpoint (PrimaryKey wid) (updateState s' cp)
+                pure mOutputs
+    UnsignedTx
+        <$> ensureNonEmpty mInputs  ErrSelectCoinsExternalUnableToAssignInputs
+        <*> ensureNonEmpty mOutputs ErrSelectCoinsExternalUnableToAssignOutputs
```
KtorZ, line 1022: **[remark]** 

I am not fond of having `xxxUnableToAssignxxx` at the value-level here because, these are clearly invariant violation. If you give a non-empty list of output to the coin selection (as we do) then the result MUST have non empty inputs and outputs. Anything else is an invariant violation and ought to be treated as an exceptional case (i.e. with an exception). 

In a sense, this is a consequence of our "insufficient" typing regarding `CoinSelection`. A possible "easy fix" would be to simply move that contract inside the coin selection functions (`random` and `largestFirst`) adding invariant about the shape of the result.

KtorZ (approved):

KtorZ:
(cancelling to make sure we pick up the **current** tagged master for the upcoming release, unfortunately, we can only fetch artifacts from the _latest_ hydra build at the moment, will merge a little later).

lib/core/src/Cardano/Wallet.hs
jonathanknowles: > I am not fond of having `xxxUnableToAssignxxx` at the value-level here because, these are clearly invariant violation. If you give a non-empty list of output to the coin selection (as we do) then the result MUST have non empty inputs and outputs. Anything else is an invariant violation and ought to be treated as an exceptional case (i.e. with an exception).
> 
> In a sense, this is a consequence of our "insufficient" typing regarding `CoinSelection`. A possible "easy fix" would be to simply move that contract inside the coin selection functions (`random` and `largestFirst`) adding invariant about the shape of the result.

I completely agree, and think it would be better to have the "cannot be empty" invariants encoded in the `CoinSelection` type itself.

I'm guessing this would require quite a bit of refactoring. Perhaps we should put this on our list to tidy up during a recovery week?

Issue #1164, opened 2019-12-12
Add caching of stake pool metadata registry


rvl:
# Context

Part of [WB-42](https://jira.iohk.io/browse/WB-42).

The listStakePools endpoint will be polled frequently by clients like Daedalus. It will fetch stake pool metadata from the registry every time. The metadata registry data won't change very often.

Add caching of the stake pool metadata with a TTL of say 1 hour.

# Decision

The best solution would be to load the registry metadata into the database, where it can be joined with the stake pool registration data.

But a quicker interim solution could be to download and cache the registry zip file into the databases directory.

# Acceptance Criteria

1. The stake pool metadata must not be downloaded every single time the API endpoint is called.

---

# Development 

- Cache downloaded archive in a temporary folder and renew it every hours ==> #1166 
- Make sure that the cache can still be downloaded if the temporary cache dir is removed ==> #1174 

# QA 

- Added unit tests using a mock server to serve the archive.
    - 1. Test that the the archive is only fetched once from the server by making two successive calls and monitoring logs for tracing code execution: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Cardano/Pool/MetadataSpec.hs#L97-L105

   - 2. Another test that make a first call, then forces the cache to expire (by manually changing the directory "modification time" to one second after the cache TTL) and monitoring logs for tracing code execution: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Cardano/Pool/MetadataSpec.hs#L107-L117 

   - 3. A third test that tries to download the archive, cache it, remove the cache folder and re-download successfully the archive. See #1174

piotr-iohk:
TC that fails:
 - list pools with metadata -> OK
 - remove the `/tmp/stake-pools-xxx` directory
 - list stake pools again -> Metadata is not fetched, in the log there is:
```
[cardano-wallet:Info:2224] [2019-12-13 15:29:09.14 UTC] Downloading https://github.com/piotr-iohk/incentivized-testnet-stakepool-registry/archive/master.zip -> /tmp/stake-pool-metadata-3fd60f4b16799520/registry.zip: Started download
[cardano-wallet:Notice:2224] [2019-12-13 15:29:09.14 UTC] Stake pool metadata is unavailable
```

piotr-iohk:
Putting back to _in progress_ for now until :point_up_2:  addressed (also such case can be added). Otherwise looks good.

[...]

PR #1166, opened 2019-12-12
Cache the stake pool metadata registry

rvl:
Relates to #1164.

# Overview

Caches the downloaded stake pool metadata registry zip file in a temporary directory.

### DIFF
--- a/lib/core/src/Cardano/Pool/Metadata.hs
+++ b/lib/core/src/Cardano/Pool/Metadata.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 
 -- |
@@ -23,6 +24,7 @@ module Cardano.Pool.Metadata
     , getStakePoolMetadata
     , getRegistryZipUrl
     , cardanoFoundationRegistryZip
+    , cacheTTL
     , FetchError (..)
 
       -- * Logging
@@ -45,7 +47,7 @@ import Cardano.Wallet.Primitive.Types
     ( PoolOwner (..), ShowFmt (..) )
 import Codec.Archive.Zip
 import Control.Exception
-    ( IOException, displayException, tryJust )
+    ( IOException, displayException, try, tryJust )
 import Control.Monad
     ( join, (>=>) )
 import Control.Monad.IO.Class
@@ -70,6 +72,8 @@ import Data.Text
     ( Text )
 import Data.Text.Class
     ( FromText (..), TextDecodingError (..), ToText (..) )
+import Data.Time.Clock
+    ( NominalDiffTime, UTCTime, diffUTCTime, getCurrentTime )
 import Fmt
     ( fmt, (+|), (|+) )
 import GHC.Generics
@@ -78,14 +82,14 @@ import Network.HTTP.Client
     ( HttpException, parseUrlThrow )
 import Network.HTTP.Simple
     ( httpSink )
+import System.Directory
+    ( getModificationTime )
 import System.Environment
     ( lookupEnv )
 import System.FilePath
     ( isPathSeparator, (<.>), (</>) )
 import System.IO
-    ( Handle, hClose, hTell )
-import System.IO.Temp
-    ( withSystemTempFile )
+    ( IOMode (WriteMode), hTell, withFile )
 
 import qualified Data.Aeson as Aeson
 import qualified Data.ByteString as BS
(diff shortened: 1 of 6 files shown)


lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -194,31 +198,59 @@ defaultRecordTypeOptions = Aeson.defaultOptions
...
-    req <- parseUrlThrow url
-    httpSink req $ \_response -> Conduit.mapM_ (BS.hPut zipFile)
-    hTell zipFile
+    checkCached = try (getModificationTime zipFileName) >>= \case
+        Right mtime -> do
+            now <- getCurrentTime
+            pure $ if (diffUTCTime now mtime < cacheTTL)
+                then Just mtime
+                else Nothing
+        Left (_ :: IOException) -> pure Nothing
```
KtorZ: **[question]**

I suppose that `getModificationTime` throws an IOException when the file doesn't exist?

lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -377,3 +411,5 @@ instance ToText RegistryLogMsg where
             fmt $ "Could not parse metadata: "+|e|+""
         MsgExtractFileResult (Just (Right _)) ->
             "Successfully parsed metadata"
+        MsgUsingCached _ mtime ->
+            "Using cached file last modified at " <> T.pack (show mtime)
```
KtorZ, line 433: :+1:

lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -194,31 +198,59 @@ defaultRecordTypeOptions = Aeson.defaultOptions
...
-        fetchStakePoolMetaZip url zipFile >>= \case
-            Right size -> do
-                logTrace tr' $ MsgDownloadComplete size
-                hClose zipFile
-                Right <$> getMetadataFromZip tr' zipFileName poolOwners
-            Left e -> pure $ Left e
+getStakePoolMetadata tr cacheDir url poolOwners = do
+    let zipFileName = cacheDir </> "registry.zip"
+    let tr' = contramap (fmap (RegistryLog url zipFileName)) tr
+    createDirectoryIfMissing False cacheDir
```
KtorZ: **[question]**

Why not also create parent directory (i.e. ~~False~~ -> True) if they don't exist to easily allow creating nested directories? :thinking:

lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -194,31 +198,59 @@ defaultRecordTypeOptions = Aeson.defaultOptions
...
+    fetchStakePoolMetaZipCached tr' url zipFileName >>= \case
+        Right _ -> Right <$> getMetadataFromZip tr' zipFileName poolOwners
+        Left e -> pure $ Left e
+
+cacheTTL :: NominalDiffTime
+cacheTTL = 3600
```
KtorZ: **[suggestion]**

Add a comment about the duration. `NominalDiffTime` interprets `Int` as seconds for conversions through the `Num` instance, but that's not necessarily obvious :sweat_smile:

lib/core/test/unit/Cardano/Pool/MetadataSpec.hs
```diff
@@ -76,26 +75,47 @@ import qualified Data.Text as T
...
 
+        it "Caches the zip" $ \(baseUrl, metadataDir, tr, getMsgs) -> do
+            void $ getStakePoolMetadata tr metadataDir (testUrl baseUrl) presentOwners
+
+            res <- getStakePoolMetadata tr metadataDir (testUrl baseUrl) presentOwners
+            res `shouldBe` Right (map Just presentMetas)
+
+            msgs <- map registryLogMsg <$> getMsgs
+            length (filter isMsgDownloadStarted msgs) `shouldBe` 1
+            length (filter isMsgUsingCached msgs) `shouldBe` 1
```
KtorZ, line 105: **[remark]**

It's kinda cool to be able to test log messages directly.

lib/core/src/Cardano/Pool/Metadata.hs
rvl: Yeah, maybe. I was originally planning to bury this file in the database directory, which should exist. But actually the database is optional, so now it's now a temporary directory. So to resolve this question I now remove the createDirectoryIfMissing altogether.

lib/core/test/unit/Cardano/Pool/MetadataSpec.hs
rvl: Yes, Neil told me this also. The tests can easily find out what your program is doing.

lib/core/test/unit/Cardano/Pool/MetadataSpec.hs
```diff
@@ -76,26 +75,47 @@ import qualified Data.Text as T
...
+
+        it "Refreshes expired cache" $ \(baseUrl, metadataDir, tr, getMsgs) -> do
+            void $ getStakePoolMetadata tr metadataDir (testUrl baseUrl) presentOwners
+
+            old <- makeExpired <$> getCurrentTime
+            forFilesIn metadataDir $ \f -> setModificationTime f old
```
KtorZ: **[suggestion]**

Wouldn't be "simpler" to give the cache TTL to `getStakePoolMetadata` as an argument. We could make use of a config record type such as:

```hs
data MetadataConfig = MetadataConfig
    { cacheArchive :: FilePath
    , cacheTTL :: NominalDiffTime
    , registryURL :: String
    }
```

With a default configuration and an accessor like `getMetadataConfig :: IO MetadataConfig` which takes care of overriding the URL with the right ENV var.

Thoughts :thinking: ?

KtorZ (approved):

lib/core/test/unit/Cardano/Pool/MetadataSpec.hs
rvl: Yes, good suggestion.

rvl:
OK I have addressed the review comments.

KtorZ (approved):

PR #1168, opened 2019-12-12
use pool production tip to compute performance

KtorZ:
# Issue Number

#1158 

# Overview

- [ ] I have used pool production tip to compute performanceinstead of the node tip. The production worker might not have caught up with the network yet and therefore, yield statistics that are slightly off

# Comments

### DIFF
--- a/lib/core/src/Cardano/Pool/Metrics.hs
+++ b/lib/core/src/Cardano/Pool/Metrics.hs
@@ -281,11 +281,12 @@ newStakePoolLayer tr db@DBLayer{..} nl metadataDir = StakePoolLayer
             <$> (Map.fromList <$> readStakeDistribution nodeEpoch)
             <*> readPoolProduction nodeEpoch
 
-        let tip = maximumBy (compare `on` view #slotId)
+        let prodTip = maximumBy (compare `on` view #slotId)
                 $ mconcat (Map.elems prod)
 
         when (Map.null distr || Map.null prod) $ do
-            computeProgress tip >>= throwE . ErrMetricsIsUnsynced
+            liftIO $ logTrace tr $ MsgComputedProgress prodTip nodeTip
+            throwE $ ErrMetricsIsUnsynced $ computeProgress prodTip nodeTip
 
         if nodeEpoch == genesisEpoch
         then do
@@ -293,7 +294,8 @@ newStakePoolLayer tr db@DBLayer{..} nl metadataDir = StakePoolLayer
             combineWith (sortArbitrarily seed) distr (count prod) mempty
 
         else do
-            perfs <- liftIO $ readPoolsPerformances db epochLength (tip ^. #slotId)
+            let sl = prodTip ^. #slotId
+            perfs <- liftIO $ readPoolsPerformances db epochLength sl
             combineWith (pure . sortByPerformance) distr (count prod) perfs
 
     -- For each pool, look up its metadata. If metadata could not be found for a
@@ -334,11 +336,6 @@ newStakePoolLayer tr db@DBLayer{..} nl metadataDir = StakePoolLayer
             Left e ->
                 throwE $ ErrListStakePoolsMetricsInconsistency e
 
-    poolProductionTip :: IO (Maybe BlockHeader)
-    poolProductionTip = atomically $ readPoolProductionCursor 1 >>= \case
-        [x] -> return $ Just x
-        _ -> return Nothing
-
     sortByPerformance :: [StakePool] -> [StakePool]
     sortByPerformance = sortOn (Down . apparentPerformance)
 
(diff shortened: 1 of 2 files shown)


lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -274,20 +276,22 @@ newStakePoolLayer db@DBLayer{..} nl tr = StakePoolLayer
 
         (distr, prod) <- liftIO . atomically $ (,)
             <$> (Map.fromList <$> readStakeDistribution nodeEpoch)
-            <*> (count <$> readPoolProduction nodeEpoch)
+            <*> readPoolProduction nodeEpoch
+
+        let tip = maximumBy (compare `on` view #slotId)
```
Anviking: Why not use `poolProductionCursor` and `case` on the `Maybe`?

lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -274,20 +276,22 @@ newStakePoolLayer db@DBLayer{..} nl tr = StakePoolLayer
...
+                $ mconcat (Map.elems prod)
 
         when (Map.null distr || Map.null prod) $ do
-            computeProgress nodeTip >>= throwE . ErrMetricsIsUnsynced
+            computeProgress tip >>= throwE . ErrMetricsIsUnsynced
```
Anviking: I think here is a case when we actually don't want the `poolProductionTip` 😅 

`computeProgress` compares the `poolProductionTip` with this tip, so I think it will only display 100% now.

Anviking: I see two options:
- Keep it comparing with the node tip
- Compare against slot 0 of the node's epoch (should be when the results are actually first displayed 🤔 )

Maybe best to keep it the same for now.

Anviking (changes requested):

lib/core/src/Cardano/Pool/Metrics.hs
KtorZ: Because side-effects are bad :grimacing: ? 
We already have that information available, so there's no need to use another call for this. 

cf my last commit.

lib/core/src/Cardano/Pool/Metrics.hs
KtorZ: Done :+1:

lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -602,14 +599,13 @@ instance ToText StakePoolLayerMsg where
         MsgRegistry msg -> toText msg
         MsgListStakePoolsBegin -> "Listing stake pools"
         MsgMetadataUnavailable -> "Stake pool metadata is unavailable"
-        MsgComputedProgress (Just dbTip) nodeTip -> mconcat
+        MsgComputedProgress prodTip nodeTip -> mconcat
             [ "The node tip is:\n"
             , pretty nodeTip
             , ",\nbut the last pool production stored in the db"
```
Anviking, line 611: This is not _exactly_ true anymore. `prodTip` is now the greatest slot stored in the DB, where `epoch == nodeTip.epoch`. 

If the DB were to find itself at a greater slotId than the node this distinction would matter… But that shouldn't be possible.

edit: see other comment which is more concerning

lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -274,20 +276,24 @@ newStakePoolLayer db@DBLayer{..} nl tr = StakePoolLayer
 
         (distr, prod) <- liftIO . atomically $ (,)
             <$> (Map.fromList <$> readStakeDistribution nodeEpoch)
-            <*> (count <$> readPoolProduction nodeEpoch)
+            <*> readPoolProduction nodeEpoch
+
+        let prodTip = maximumBy (compare `on` view #slotId)
```
Anviking: From the docs for [`maximumBy`](https://hackage.haskell.org/package/base-4.12.0.0/docs/Data-List.html#v:maximumBy):
> The largest element of a non-empty structure with respect to the given comparison function.

If there are no pool productions in the epoch of the node tip, this will crash. I.e when it should have  thrown `ErrMetricsIsUnsynced`, it will crash instead.

You cannot move this binding to after the `when (Map.null distr || Map.null prod` check, since `computeProgress` relies on the `prodTip`.

I would suggest using `poolProductionTip`. You can put it inside the `atomically` and there shouldn't be any downsides:

```haskell
(distr, prod, dbTip) <- liftIO . atomically $ (,,)
             <$> (Map.fromList <$> readStakeDistribution nodeEpoch)
             <*> readPoolProduction nodeEpoch
             <*> (readPoolProductionCursor 1 >>= toMaybe)
```

Anviking (changes requested):
Looks like `maximumBy` will crash when the follower is un-synced. We should also at least test the PR  manually on the testnet. 

(but not sure if there is a testnet we are compatible with at the very moment)

Anviking:
bors try

lib/core/src/Cardano/Pool/Metrics.hs
KtorZ: Hmmm. Good point!

lib/core/src/Cardano/Pool/Metrics.hs
KtorZ: Well, any epoch stored that would be greater than the nodeTip is irrelevant. That the symptoms of a rollback that hasn't yet happened on the database. In practice, the most recent production tip from the node's epoch is our current state.

KtorZ:
@Anviking 

- Fixed tip calculation (using the db cursor in the end :man_shrugging:)
- Added two basic unit tests (one to test the case where there's no production, indeed, it crashed. And one to test that the calculation works as expected).

Anviking (approved):
lgtm

PR #1169, opened 2019-12-12
use another stake pool with metadata in STAKE_POOL_LIST_04

piotr-iohk:
# Issue Number

#1065

# Overview

- [ ] I thought it will be useful to use the second stake-pool with metadata as well. If anything bad happens, it is more likely to be discovered when there are more pools... I think. (Also added some more comments around the stake pool faucet)

# Comments


KtorZ (approved):
Why not :man_shrugging:

PR #1171, opened 2019-12-12
Add 'cost' and 'margin' to stake pools

KtorZ:
# Issue Number

#1090 

# Overview

- [x] I have added cost and margin to 'PoolRegistrationCertificate'  and adjusted Jörmungandr binary decoders to also parse these.
- [x] I have slightly revised the pool database layer to not store (resp. read) only pool owners, but store (resp. read) registration certificates entirely.
- [x] I have adjusted the logic from `listStakePools` to attach `cost` and `margin` to each pool 
- [x] I have extended the API types
- [x] adjusted the swagger file to include these 

# Comments

```
Cardano.Pool.DB.Sqlite
  Sqlite
    Stake Pool properties
      putPoolRegistration then readPoolRegistration yields expected result
        +++ OK, passed 100 tests.
      rollback of pool Registrations
        +++ OK, passed 100 tests:
        87% rolled back some
        77% owner has many pools
```

### DIFF
--- a/lib/core/src/Cardano/Pool/DB.hs
+++ b/lib/core/src/Cardano/Pool/DB.hs
@@ -86,10 +86,12 @@ data DBLayer m = forall stm. MonadFail stm => DBLayer
         -- This is useful for the @NetworkLayer@ to know how far we have synced.
 
     , putPoolRegistration
-        :: PoolRegistrationCertificate
+        :: EpochNo
+        -> PoolRegistrationCertificate
         -> stm ()
-        -- ^ Add a mapping between stake pools and their certificate. If the
-        -- mapping already exists, this will be a no-op.
+        -- ^ Add a mapping between stake pools and their corresponding
+        -- certificate. If the mapping already exists, data are replaced with
+        -- the latest version.
 
     , readPoolRegistration
         :: PoolId
--- a/lib/core/src/Cardano/Pool/DB.hs
+++ b/lib/core/src/Cardano/Pool/DB.hs
@@ -20,7 +20,12 @@ module Cardano.Pool.DB
 import Prelude
 
 import Cardano.Wallet.Primitive.Types
-    ( BlockHeader, EpochNo (..), PoolId, PoolOwner (..), SlotId (..) )
+    ( BlockHeader
+    , EpochNo (..)
+    , PoolId
+    , PoolRegistrationCertificate
+    , SlotId (..)
+    )
 import Control.Monad.Fail
     ( MonadFail )
 import Control.Monad.Trans.Except
@@ -73,23 +78,23 @@ data DBLayer m = forall stm. MonadFail stm => DBLayer
         -> stm [(PoolId, Quantity "lovelace" Word64)]
 
     , readPoolProductionCursor
-        :: Int -> stm [BlockHeader]
+        :: Int
+        -> stm [BlockHeader]
         -- ^ Read the latest @k@ blockheaders in ascending order. The tip will
         -- be the last element in the list.
         --
         -- This is useful for the @NetworkLayer@ to know how far we have synced.
 
-    , putStakePoolOwner
-        :: PoolId
-        -> PoolOwner
+    , putPoolRegistration
+        :: PoolRegistrationCertificate
         -> stm ()
-        -- ^ Add a mapping between stake pools and owners. If the mapping
-        -- already exists, this will be a no-op.
+        -- ^ Add a mapping between stake pools and their certificate. If the
+        -- mapping already exists, this will be a no-op.
 
-    , readStakePoolOwners
+    , readPoolRegistration
         :: PoolId
-        -> stm [PoolOwner]
-        -- ^ List the owners of a given stake pool.
+        -> stm (Maybe PoolRegistrationCertificate)
+        -- ^ Find a registration certificate associated to a given pool
 
     , readSystemSeed
         :: stm StdGen
@@ -99,7 +104,8 @@ data DBLayer m = forall stm. MonadFail stm => DBLayer
         -- results across requests.
 
     , rollbackTo
-        :: SlotId -> stm ()
+        :: SlotId
+        -> stm ()
         -- ^ Remove all entries of slot ids newer than the argument
 
     , cleanDB
--- a/lib/core/src/Cardano/Pool/DB.hs
+++ b/lib/core/src/Cardano/Pool/DB.hs
@@ -98,6 +98,13 @@ data DBLayer m = forall stm. MonadFail stm => DBLayer
         -> stm (Maybe PoolRegistrationCertificate)
         -- ^ Find a registration certificate associated to a given pool
 
+    , listRegisteredPools
+        :: stm [PoolId]
+        -- ^ List the list of known pools, based on their registration
+        -- certificate. This list doesn't necessarily match the keys of the
+        -- map we would get from 'readPoolProduction' because not all registered
+        -- pools have necessarily produced any block yet!
+
     , readSystemSeed
         :: stm StdGen
         -- ^ Read the seed assigned to this particular database. The seed is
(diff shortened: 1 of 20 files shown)


lib/core/src/Cardano/Pool/DB.hs
```diff
@@ -73,23 +78,23 @@ data DBLayer m = forall stm. MonadFail stm => DBLayer
...
 
-    , putStakePoolOwner
-        :: PoolId
-        -> PoolOwner
+    , putPoolRegistration
+        :: PoolRegistrationCertificate
         -> stm ()
-        -- ^ Add a mapping between stake pools and owners. If the mapping
-        -- already exists, this will be a no-op.
+        -- ^ Add a mapping between stake pools and their certificate. If the
```
paweljakubas: **[suggestion]**

```suggestion
        -- ^ Add a mapping between stake pools and their corresponding certificates. If the
```

lib/core/src/Cardano/Pool/DB/Model.hs
```diff
@@ -85,6 +92,9 @@ data PoolDatabase = PoolDatabase
     , owners :: !(Set (PoolId, PoolOwner))
     -- ^ Mapping between pool ids and owners
 
+    , metadata :: !(Map PoolId (Percentage, Quantity "lovelace" Word64))
+    -- ^ On-chain metadata associated with pools
```
paweljakubas, line 94: :+1:

lib/core/src/Cardano/Pool/DB/Model.hs
```diff
@@ -156,18 +166,35 @@ mReadStakeDistribution epoch db@PoolDatabase{distributions} =
     , db
     )
 
-mPutStakePoolOwner :: PoolId -> PoolOwner -> ModelPoolOp ()
-mPutStakePoolOwner poolId poolOwner db@PoolDatabase{owners} =
+mPutPoolRegistration :: PoolRegistrationCertificate -> ModelPoolOp ()
+mPutPoolRegistration registration db@PoolDatabase{owners,metadata} =
     ( Right ()
-    , db { owners = Set.insert entry owners }
+    , db { owners = Set.union entries owners
```
paweljakubas: :+1:

lib/core/src/Cardano/Pool/DB/Model.hs
```diff
@@ -156,18 +166,35 @@ mReadStakeDistribution epoch db@PoolDatabase{distributions} =
...
+        , poolCost
+        , poolMargin
+        } = registration
+
+mReadPoolRegistration :: PoolId -> ModelPoolOp (Maybe PoolRegistrationCertificate)
+mReadPoolRegistration poolId db@PoolDatabase{owners, metadata} =
+    ( Right $ case Map.lookup poolId metadata of
+        Nothing -> Nothing
+        Just (poolMargin, poolCost) ->
+            let poolOwners = L.sortOn toText
```
paweljakubas: **[minor question]**

So we are sorting here on name? If yes, why not based on margin or cost?

lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -164,15 +171,33 @@ newDBLayer logConfig trace fp = do
...
+            , poolOwners
+            , poolMargin
+            , poolCost
+            } -> do
+            let poolMargin_ = fromIntegral $ fromEnum poolMargin
+            let poolCost_ = getQuantity poolCost
             repsert
+                (PoolMetadataKey poolId)
+                (PoolMetadata poolId poolMargin_ poolCost_)
+            forM_ poolOwners $ \poolOwner -> repsert
```
paweljakubas: **[question]**

`repsert` is used here. so "inserting" many rows at once and (a) if no entry then insert, (b) if entry present update it. Is it ok to have already present `poolId` and new `poolOwner` value? If no then shouldn't be there error? Moreover, don't we afraid of this `repsert` because of performance? So, we are updating many not changed rows every call?

lib/core/src/Cardano/Pool/DB/Sqlite/TH.hs
```diff
@@ -76,4 +76,13 @@ PoolOwner sql=pool_owner
 
     Primary poolOwnerPoolId poolOwnerOwner
     deriving Show Generic
+
+-- Mapping of metadata to pool
+PoolMetadata sql=pool_metadata
```
paweljakubas: **[remark]**

good idea to decouple PoolMetadata from PoolOwner. The former one is expected to have different update rate. And here I am thinking once again about repsert comment from above :thinking:

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
```diff
@@ -351,31 +358,28 @@ prop_putStakePutStake DBLayer {..} epoch a b =
...
 
-prop_stakePoolOwner
+-- | Heavily relies on the fact that PoolId have a entropy that is sufficient
+prop_poolRegistration
     :: DBLayer IO
-    -> StakePoolOwnersFixture
+    -> [PoolRegistrationCertificate]
     -> Property
-prop_stakePoolOwner DBLayer {..} (StakePoolOwnersFixture poolId entries) =
+prop_poolRegistration DBLayer {..} entries =
```
paweljakubas, line 370: **[remark]**

As we have now  PoolMetadata decoupled from PoolOwner it would be nice to introduce properties like : 
1. registering pools and see if we ONLY have pool metadata relating to this
2. registering pools and then we have pool metadata not belonging to the pool we registered - and what happens?
3. registering and deregistring some pools and what about pool metadata consistency (maybe we cannot deregister pool)

lib/jormungandr/src/Cardano/Wallet/Jormungandr/Binary.hs
```diff
@@ -276,7 +279,7 @@ data Fragment
     -- ^ A standard signed transaction
     | StakeDelegation (StakeDelegationType, ChimericAccount, Tx)
     -- ^ A signed transaction with stake pool delegation
-    | PoolRegistration (PoolId, [PoolOwner], Tx)
+    | PoolRegistration (PoolId, [PoolOwner], TaxParameters, Tx)
```
paweljakubas, line 282: :+1:

paweljakubas (approved):
I like very much that PoolMetadata is decoupled from PoolOwner. Left some questions/remarks flying mainly around this idea. And also property tests suggestions to make sure those two tables are consistent as I think they should be

lib/core/src/Cardano/Pool/DB/Model.hs
KtorZ: This is sorting the _owners_ inside **each** pool, not the pool themselves. This is really just to avoid the database (and the model) returning the list of owner in arbitrary order. This way, we can expect successive calls to return the exact same thing.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
KtorZ: Hmmmm... So, there's nothing _wrong_ with replacing the value. However, thinking again about this well, it would probably be better to simply use `insert_` and make sure that certificates are rolled back correctly like the rest (which isn't the case right now, they are really just treated as "static" metadata, hence the repsert).

lib/core/src/Cardano/Pool/DB/Sqlite/TH.hs
KtorZ: They are actually updated at the very same time (when a registration certificate is discover on-chain). The reason they are decoupled is because there's a many:many relationship between pools and owners (one pool can have multiple owners, and an owner multiple pools).

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
KtorZ: We can't retire pool at the moment indeed, we can only register new pools indeed. The property currently test that for any arbitrary list of registration, if you insert them, you can read them back, therefore making sure that the binding between owners and certificate is correctly respected. It covers your second case (where many metadata does not belong to your pool, but still you don't get extra metadata when fetching) and your first case as well (when fetching, you only get your metadata).

piotr-iohk:
Maybe it would be nice to update also integration test `STAKE_POOLS_LIST_01` asserting also for existence of `cost` and `margin`.

Issue #1160, opened 2019-12-11
Make old Haskell node network prototype compile again


Anviking:
# Context

> As a wallet backend developer and a cardano-node project manager
I want to identify gaps and outstanding effort required to integrate the wallet backend with the haskell nodes,
So that I can anticipate risks and start planning the integration.

[WB-26](https://jira.iohk.io/projects/WB/issues/WB-26?filter=allopenissues)

A while back, Matthias wrote a [prototype for integrating](https://github.com/input-output-hk/cardano-wallet/blob/cbd921f5cb8348baaf5ca7956ad4a5a3a5d60124/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs) with the Haskell node. It does not compile with current version of the node.

# Decision

Adapt the code such that it compiles and works. Focus on getting something minimal working. We may extend the prototype later to check things that interest us. (This task _might_ or _might not_ cover the second point of the acceptance criteria in the Jira U/S)

# Acceptance Criteria

- We must be able to follow the Haskell node chain and handle rollbacks
- We should be a able to construct a `Cardano.Wallet.Primitive.Types.BlockHeader` for each block

---

# Development 

# QA

KtorZ:
@Anviking note also that this prototype was doing a "conversion" from stateful to stateless which should not be necessary anymore.

Anviking:
I think I can close this. No QA since nothing is merged.

I'll open a separate issue for a potentially more extensive prototype.

PR #1173, opened 2019-12-13
Haskell Byron Node Prototype

Anviking:
# Issue Number

#1160 

# Overview

- I made @KtorZ's old prototype compile, and simplified it  (or sabotaged depending on your view)
- I start syncing from the middle of the chain
- I fetch 10 blocks and print the transactions in them

# Comments

```
=== SlotNo 416950
=== ByronHash {unByronHash = AbstractHash dd42824af406bc7e1042a6173c46791f2b9ad963ef417cfab8dab4a53ea555cc}


[UnsafeTx {txInputs = TxInUtxo (AbstractHash 0605489384485884c752b1d094d81da4cd22a2014900304e9beab61d4c420142) 1 :| [TxInUtxo (AbstractHash 09ddf887bb9ee8613219df48f80a88a528a0033250b71fcea0c68ebe28dcb6d4) 0], txOutputs = TxOut {txOutAddress = Address {addrRoot = AbstractHash 4d6fb44ea75e70897c8f23eebe304ebb23d554e708b42952311fe69c, addrAttributes = Attributes { data: AddrAttributes {aaVKDerivationPath = Just (HDAddressPayload {getHDAddressPayload = "\191\230\202t'\137A$\213rB\185\219_\212\250\135\142\249\DC1\197\232nw\184*/\184"}), aaNetworkMagic = NetworkMainOrStage} }, addrType = ATVerKey}, txOutValue = Lovelace {getLovelace = 1000116246}} :| [TxOut {txOutAddress = Address {addrRoot = AbstractHash f7ee858829d88d6088ccaafc35a2c9f66169d153f29d55381151a925, addrAttributes = Attributes { data: AddrAttributes {aaVKDerivationPath = Just (HDAddressPayload {getHDAddressPayload = "\130\185\180.\164\&0\SOH\232\231\223\&9\ETB&_~\SOH\183a\r~Lu\"\137\243\179\DC4x"}), aaNetworkMagic = NetworkMainOrStage} }, addrType = ATVerKey}, txOutValue = Lovelace {getLovelace = 169000000}}], txAttributes = Attributes { data: () }},UnsafeTx {txInputs = TxInUtxo (AbstractHash 33eb7b25fb7965e7b1d600c946cdd3915d75ab707007d61ed248b4ac70c275d7) 0 :| [], txOutputs = TxOut {txOutAddress = Address {addrRoot = AbstractHash a97bc7774f05f25c3bedbc0bd4914ed57b6926895dc6fdaa8ab0c4dd, addrAttributes = Attributes { data: AddrAttributes {aaVKDerivationPath = Just (HDAddressPayload {getHDAddressPayload = "\217\253\161\&6d\195)\138q\206N\ESC\139\181\206%u\230CN\CAN\245\151A\173\189\185\EM"}), aaNetworkMagic = NetworkMainOrStage} }, addrType = ATVerKey}, txOutValue = Lovelace {getLovelace = 26988486614}} :| [TxOut {txOutAddress = Address {addrRoot = AbstractHash c0ae29d63eb6772accf49f0d2b975e269ab65eabf3e41a1d2a9f8abd, addrAttributes = Attributes { data: AddrAttributes {aaVKDerivationPath = Just (HDAddressPayload {getHDAddressPayload = "\130\185\180.\164\&0\SOH\152O\201\197\ETB\208\135r\177Q9~N\155\204\153\214\239]\SOH>"}), aaNetworkMagic = NetworkMainOrStage} }, addrType = ATVerKey}, txOutValue = Lovelace {getLovelace = 5000000000}}], txAttributes = Attributes { data: () }},UnsafeTx {txInputs = TxInUtxo (AbstractHash 9052a89b328245613a479b6070ba046018b90334e59cd937fa2173750168c876) 0 :| [], txOutputs = TxOut {txOutAddress = Address {addrRoot = AbstractHash b5afa87128e449b63d4e3ec7e2b3ad35e4226245a583caa2698bcaeb, addrAttributes = Attributes { data: AddrAttributes {aaVKDerivationPath = Just (HDAddressPayload {getHDAddressPayload = "\234\DC4%\204\221d\155c -\DEL~\151\ETXjJ\150k1\141u\145\\\132\208\196%\SYN"}), aaNetworkMagic = NetworkMainOrStage} }, addrType = ATVerKey}, txOutValue = Lovelace {getLovelace = 996885581117}} :| [TxOut {txOutAddress = Address {addrRoot = AbstractHash 68d394771b0cc58a7e3ce8a66c4fcee5474868798aab3b2b1c107d21, addrAttributes = Attributes { data: AddrAttributes {aaVKDerivationPath = Just (HDAddressPayload {getHDAddressPayload = "X=\ETB\144\142$\254\241\193\253A^\USd\145W\161o`^m\154o\EOT\161Q\244\&7"}), aaNetworkMagic = NetworkMainOrStage} }, addrType = ATVerKey}, txOutValue = Lovelace {getLovelace = 154844000}}], txAttributes = Attributes { data: () }}]
```

### DIFF
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
@@ -1,0 +1,523 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+
+{-
+Good to read before / additional resources:
+
+- Module's documentation in `ouroboros-network/typed-protocols/src/Network/TypedProtocols.hs`
+- Data Diffusion and Peer Networking in Shelley (see: https://raw.githubusercontent.com/wiki/input-output-hk/cardano-wallet/data_diffusion_and_peer_networking_in_shelley.pdf)
+    - In particular sections 4.1, 4.2, 4.6 and 4.8
+-}
+
+module Cardano.Wallet.Shelley.Network
+    (
+      -- * Top-Level Interface
+      NetworkLayer (..)
+    , ErrNextBlocks (..)
+    , ChainParameters (..)
+
+    -- * Re-Export
+    , EpochSlots (..)
+    , ProtocolMagicId (..)
+
+      -- * Constructors
+    , newNetworkClient
+    , connectClient
+    , dummyNodeToClientVersion
+
+      -- * Transport Helpers
+    , localSocketAddrInfo
+    , localSocketFilePath
+    ) where
+
+import Prelude
+
+import Cardano.Chain.Slotting
+    ( EpochSlots (..) )
+import Cardano.Crypto
+    ( ProtocolMagicId (..) )
+import Codec.SerialiseTerm
+    ( CodecCBORTerm )
+import Network.Mux.Interface
+    ( AppType (..) )
+import Network.Mux.Types
+    ( MuxError )
+import Network.TypedProtocol.Channel
+    ( Channel )
+import Network.TypedProtocol.Codec
+    ( Codec )
+import Network.TypedProtocol.Codec.Cbor
+    ( DeserialiseFailure )
+import Network.TypedProtocol.Driver
+    ( TraceSendRecv, runPeer )
+import Ouroboros.Consensus.Ledger.Byron
+    ( ByronBlockOrEBB (..)
+    , GenTx
+    , decodeByronBlock
+    , decodeByronGenTx
+    , decodeByronHeaderHash
+    , encodeByronBlock
+    , encodeByronGenTx
+    , encodeByronHeaderHash
+    )
+import Ouroboros.Consensus.Ledger.Byron.Config
+    ( ByronConfig )
+import Ouroboros.Consensus.NodeId
+    ( CoreNodeId (..) )
+import Ouroboros.Network.Block
+    ( decodePoint, encodePoint )
+import Ouroboros.Network.Mux
+    ( OuroborosApplication (..) )
+import Ouroboros.Network.NodeToClient
+    ( NodeToClientProtocols (..)
+    , NodeToClientVersion (..)
+    , NodeToClientVersionData (..)
+    , connectTo
+    , nodeToClientCodecCBORTerm
+    )
+import Ouroboros.Network.Protocol.ChainSync.Client
+    ( ChainSyncClient (..)
+    , ClientStIdle (..)
+    , ClientStIntersect (..)
+    , ClientStNext (..)
+    , chainSyncClientPeer
+    )
+import Ouroboros.Network.Protocol.ChainSync.Codec
+    ( codecChainSync )
+import Ouroboros.Network.Protocol.ChainSync.Type
+    ( ChainSync )
+import Ouroboros.Network.Protocol.Handshake.Version
+    ( DictVersion (..), simpleSingletonVersions )
+import Ouroboros.Network.Protocol.LocalTxSubmission.Client
+    ( LocalTxSubmissionClient (..), localTxSubmissionClientPeer )
+import Ouroboros.Network.Protocol.LocalTxSubmission.Codec
+    ( codecLocalTxSubmission )
+import Ouroboros.Network.Protocol.LocalTxSubmission.Type
+    ( LocalTxSubmission )
+
+import Control.Exception
+    ( catch, throwIO )
+import Control.Monad.Class.MonadST
+    ( MonadST )
+import Control.Monad.Class.MonadSTM
+    ( MonadSTM
+    , TQueue (..)
+    , atomically
+    , newEmptyTMVarM
+    , newTQueue
+    , putTMVar
+    , readTQueue
+    , takeTMVar
+    , writeTQueue
+    )
+import Control.Monad.Class.MonadThrow
+    ( MonadThrow )
+import Control.Monad.Class.MonadTimer
+    ( MonadTimer, threadDelay )
+import Control.Monad.Trans.Except
+    ( ExceptT (..) )
+import Control.Tracer
+    ( Tracer, contramap )
+import Data.ByteString.Lazy
+    ( ByteString )
+import Data.Text
+    ( Text )
+import Data.Void
+    ( Void )
+import Network.Socket
+    ( AddrInfo (..), Family (..), SockAddr (..), SocketType (..) )
+
+import qualified Codec.Serialise as CBOR
+import qualified Network.Socket as Socket
+import qualified Ouroboros.Network.Block as Abstract
+
+--------------------------------------------------------------------------------
+--
+-- Stateless Interface
+--
+-- This is a first attempt at using the ChainSync mini protocol in a stateless
+-- manner. Note that this is a bit unfortunate as the
+
+-- | A stateless interface for dealing with the Haskell 'NetworkClient'
+--
+-- See 'newNetworkClient' and 'connectClient' methods to construct a
+-- 'NetworkLayer':
+--
+--     let nodeId = CoreNodeId 0
+--     let addr = localSocketAddrInfo (localSocketFilePath nodeId)
+--     let params = ChainParameters
+--          { epochSlots = EpochSlots 21600
+--          , protocolMagic = ProtocolMagicId 1097911063
+--          }
+--     (client, network) <- newNetworkClient nullTracer params
+--
+--     -- Launch the stateful connection in a separate thread
+--     void $ forkIO $ connectClient client dummyNodeToClientVersion addr
+--
+--     -- Use the network layer to send requests to the node
+--     response <- runExceptT $ nextBlocks network genesisPoint
+--     point <- networkTip network
+--     response' <- runExceptT $ nextBlocks network point
+--
+data NetworkLayer m = NetworkLayer
+    { nextBlocks :: Point -> ExceptT ErrNextBlocks m [Block]
+    , networkTip :: m Point
+    }
+
+-- | This translates the 'NetworkLayer' interface from the wallet backend.
+data ChainSyncRequest m
+    = ReqNextBlocks Point (Either ErrNextBlocks [Block] -> m ())
+    | ReqNetworkTip (Point -> m ())
+
+-- | What is considered errors in the 'NetworkLayer'
+data ErrNextBlocks
+    = ErrNextBlocksNoIntersection
+    | ErrNextBlocksRollBack
+    deriving (Show)
+
+
+--------------------------------------------------------------------------------
+--
+-- Concrete Types
+
+type Block = ByronBlockOrEBB ByronConfig
+
+type Point = Abstract.Point Block
+
+type Tx = GenTx Block
+
+data ChainParameters = ChainParameters
+    { epochSlots :: EpochSlots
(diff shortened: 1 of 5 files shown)


lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -0,0 +1,467 @@
...
+    NodeToClientProtocols
+        -- ^ Specifies which mini-protocols our client is talking.
+        -- 'NodeToClientProtocols' allows for two mini-protocols:
+        --  - Chain Sync
+        --  - Tx submission
+    m
+        -- ^ Underlying monad we run in
+    ByteString
+        -- ^ Concrete representation for bytes string
+    ()
```
Anviking, line 222: I changed this from `Void` to `()` in hopes of being able to terminate the protocol with `SendMsgDone`. Not sure how effective it was.

I initially thought I could have `[ByronBlock]` here, and retrieve them in the end, but that did not seem like the case. It seems like `()` vs `Void` are the only useful types here.

Anviking: Seems like this return type is lost in `connectTo` which returns `IO ()`. https://github.com/input-output-hk/ouroboros-network/blob/9d8493e6d609f879c9ebc684bc5d830f1f146bb6/ouroboros-network/src/Ouroboros/Network/NodeToClient.hs#L161-L174

KtorZ: Hmmm. The chain follower is never supposed to end, so `Void` should be the de-facto type here.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -0,0 +1,467 @@
...
+    putStrLn "\n"
+    case byronBlockRaw b  of
+        CC.ABOBBlock cb -> do
+            let body = CC.blockBody cb
+            let txPay = CC.bodyTxPayload body
+            let txAuxs = CC.aUnTxPayload txPay
+            let txs = map CC.taTx txAuxs
+            print txs
+        CC.ABOBBoundary _ ->
+            putStrLn "Our friend the Epoch Boundary Block!"
```
KtorZ: Not really my type of friend though.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -0,0 +1,515 @@
...
+convertBlockHeader :: CC.Block -> W.BlockHeader
+convertBlockHeader b = W.BlockHeader
+    { W.slotId =
+        error "2. can't easily get the SlotId"
+            -- The cardano-ledger equivalent is EpochAndSlotCount, i.e epoch
+            -- number and the slot counted from the start of the epoch.
+            --
+            -- But block headers only contain the slot number (SlotNo), counted
+            -- from genesis.
+    , W.blockHeight = error "3. We can't get the blockHeight"
```
Anviking: This is wrong. We [_can_](https://github.com/input-output-hk/ouroboros-network/blob/9d8493e6d609f879c9ebc684bc5d830f1f146bb6/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Block.hs#L198) get it.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -0,0 +1,515 @@
...
+    { W.txId = convertHash $ CC.hash tx
+    , W.resolvedInputs = map convertTxIn $ NE.toList $ CC.txInputs tx
+    , W.outputs = map convertTxOut $ NE.toList $ CC.txOutputs tx
+    }
+  where
+    convertHash = W.Hash . BA.convert
+
+    convertTxIn (CC.TxInUtxo txH idx) = (W.TxIn (convertHash txH) idx, coin)
+       where
+          coin = error "1. We can't get the coin of Byron inputs!"
```
KtorZ: Yeah, I was expecting that one :/

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -0,0 +1,515 @@
...
+
+convertBlockHeader :: CC.Block -> W.BlockHeader
+convertBlockHeader b = W.BlockHeader
+    { W.slotId =
+        error "2. can't easily get the SlotId"
+            -- The cardano-ledger equivalent is EpochAndSlotCount, i.e epoch
+            -- number and the slot counted from the start of the epoch.
+            --
+            -- But block headers only contain the slot number (SlotNo), counted
+            -- from genesis.
```
KtorZ: So this matches more the idea of "flat" slots that we wanted to have right ?

Anviking: Yes. I think this is just good. 

Converting `SlotNo` to `UTCTime` is simpler than from `SlotId`. 

Sometimes we need to be aware of epochs. E.g.
1. To know when  [delegation certificates](https://github.com/input-output-hk/cardano-ledger/blob/dbdb643722e431e4d232345a0eafdc7bdeab7b60/cardano-ledger/src/Cardano/Chain/Delegation/Certificate.hs#L90) take effect.
2. Counting blocks per epoch (if we are still going to do that)

But… we could just use the [`BoundaryBlock`](https://github.com/input-output-hk/cardano-ledger/blob/dbdb643722e431e4d232345a0eafdc7bdeab7b60/cardano-ledger/src/Cardano/Chain/Block/Header.hs#L448-L453) for that, which tell us the new epoch when it starts.

KtorZ: Meh. I'd rather also adopt this approach of keeping track of `SlotNo` instead of `SlotId`. There are trade-offs obviously, especially in the way we display things at the API-level but it's probably worth the short.

lib/shelley/src/Cardano/Wallet/Shelley/Prototype1.hs
```diff
@@ -0,0 +1,108 @@
...
+-- Goal: see what info we can extract from blocks!
+rollForward :: ByronBlock -> IO ()
+rollForward b = do
+    let slot = show (O.unSlotNo . O.byronBlockSlotNo $ b)
+    let hash = show $ O.byronBlockHash b
+    putStrLn "\n\n"
+    putStrLn $ "=== SlotNo " ++ slot
+    putStrLn $ "=== " ++ hash
+    putStrLn "\n"
+    print $ convertBlockHeader b
```
Anviking, line 95: Printing converted `BlockHeader`s as set out to do.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -0,0 +1,457 @@
...
+    ( ChainSync )
+import Ouroboros.Network.Protocol.Handshake.Version
+    ( DictVersion (..), simpleSingletonVersions )
+import Ouroboros.Network.Protocol.LocalTxSubmission.Client
+    ( LocalTxSubmissionClient (..), localTxSubmissionClientPeer )
+import Ouroboros.Network.Protocol.LocalTxSubmission.Codec
+    ( codecLocalTxSubmission )
+import Ouroboros.Network.Protocol.LocalTxSubmission.Type
+    ( LocalTxSubmission )
+
```
paweljakubas, line 106: why space here ?

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -0,0 +1,457 @@
...
+-- $ cd cardano-node
+-- $ ./scripts/mainnet.sh
+--
+-- Then run main:
+runExperiment
+    :: Point ByronBlock
+       -- ^ Point to start fetching blocks from
+    -> Int
+       -- ^ How many blocks to fetch
+    -> (ByronBlock -> IO ())
```
paweljakubas, line 163: why do you have here `ByronBlock` and not `Point ByronBlock` like for roll-backward? maybe it is going to be evident downwards...

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -0,0 +1,457 @@
...
+    :: forall vData. (vData ~ NodeToClientVersionData)
+    => NetworkClient IO
+    -> (vData, CodecCBORTerm Text vData)
+    -> AddrInfo
+    -> IO ()
+connectClient client (vData, vCodec) addr = do
+    let vDict = DictVersion vCodec
+    let versions = simpleSingletonVersions NodeToClientV_1 vData vDict client
+    let connectTo' = connectTo $ NetworkConnectTracers nullTracer nullTracer
+    -- TODO(anviking): what are these traces?
```
paweljakubas, line 248: see here : https://github.com/input-output-hk/cardano-node/tree/master/cardano-node/src/Cardano/Tracing

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -0,0 +1,457 @@
...
+         { epochSlots = EpochSlots 21600
+         , protocolMagic = ProtocolMagicId 764824073
+         }
+
+    let t = nullTracer
+    let client = OuroborosInitiatorApplication $ \pid -> \case
+            ChainSyncWithBlocksPtcl ->
+                chainSyncWithBlocks
+                    pid t params
+                    rollForward
```
paweljakubas, line 184: ok, `rollForward` is not decorated by `Point` as `chainSyncWithBlocks takes it which is defined by you below ;-)

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
Anviking: Hm, right. Mistake. Although I don't think we ever want to merge this PR? (Or am I wrong?)

paweljakubas:
By and large, I think the prototype showed its purpose. It works :-) It would be more reviewable (for me) if we could (of course in the next step) to fit it in our code structure, ie., to function as alternative side by side to Jormungandr. If done that way, then we would see very precisely what is reusable and what needs separate treatment (although it can be seen even from this prototype), where are differences that needs abstraction improvement, etc.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
KtorZ: Because, to roll forward, you need the content of the block to apply! But to roll backward, a reference to the block (the header hash + slot id) is sufficient.

Anviking:
> to fit it in our code structure

> If done that way, then we would see very precisely what is reusable and what needs separate treatment

In my previous two-day-attempt I got the idea that the current `NetworkLayer` should be made more abstract by replacing `nextBlocks` with just a `follow`-function. https://github.com/input-output-hk/cardano-wallet-laboratory/issues/15

I avoided looking into that this time, partly because I've already done it, partly because I interpreted @KtorZ's "We are not trying yet to redesign the NetworkLayer." to hint at it not being relevant.

Merging with the NetworkLayer would more be question of: "what is our solution". It is an interesting topic. But it is also different from asking "what are the problems".

In WB-26 we at least want to clearly know "what are the problems". Actually creating a Haskell `NetworkLayer` sounds tempting , but I don't think its necessary, and it might be better to do when we're actually integrating the work… Or maybe 🤔 Actually, yes, maybe it would be manageable and good… I think I'll give it a shot.

Thanks for the thought, @paweljakubas!!

Anviking:
I think I can close this and potentially open separate ticket for a more extensive prototype.

PR #1174, opened 2019-12-13
Additional unit test for cache archive: handle missing directory

KtorZ:
# Issue Number

#1164 

# Overview

- [x] I have added an extra unit test for metadata cache: cache, remove, re-download
- [x] I have seen my test fail
- [x] I have fixed the issue by making sure that the parent folder exist before downloading the archive

# Comments


piotr-iohk (approved):
:+1:

Issue #1177, opened 2019-12-13
Possible missing db migration between balance-check version and latest one


KtorZ:
# Context

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | 2019.12.13 (git revision: 11f269891d54b870b8768455292d2345e95b6810)  |
| Platform     | All           |
| Installation | Source             |

# Steps to Reproduce

1. Start a wallet server using v2019-11-18, create a wallet, wait for a few block to sync...
2. Quit
3. Start a wallet on the same state dir using v2019-12-09 or v2019-12-13, wait for a first block to sync.

4. (expected) everything is fine.
4. (actual) :boom:  
```
SQLite3 returned ErrorConstraint while attempting to perform step: NOT NULL constraint failed: seq_state_backup.reward_xpub
[cardano-wallet.worker:Error:29] [2019-12-14 16:11:56.57 UTC] 822329c8: Worker has exited unexpectedly: SQLite3 returned ErrorConstraint while attempting to perform step: NOT NULL constraint failed: seq_state_backup.reward_xpub
```

## Expected behavior

In the case of these 2 versions, a DB migration is actually impossible, because it needs access to the xprv to populate the DB.

In the general case going forward:
- it should automatically migrate the wallet databases when possible.
- the stake pools database can be cleared at startup and restored from scratch because there is no wallet-specific data there. 
- if not possible to migrate, it should fail to start with a good error message.
- (perhaps) if the jormungandr version was updated, `cardano-wallet launch` should delete the `chain` directory.
- only forwards migrations (not backwards) need be supported.

## Actual behavior
 
- As above

---

# Resolution 

1. Implement clearing of stake pools database => #1201.
2. Add automatic tests of DB migrations so that we don't release with a migration error => #1190.

---

# QA 

1. Ability to migrate to newer versions should be tested in CI.

piotr-iohk:
Ok. there are automated migration tests in CI for Windows and Linux now.
Note that I couldn't test manually migration between `v2019-11-18` and latest `v2020-01-14` because supported Jormungandr versions (`0.7.0` and `0.8.6` respectively) are not compatible.

The error is:

```
Jan 20 13:43:48.307 Parsing error on Block0
 |-> Structure invalid: Invalid config parameter tag
INFO[cardano-wallet.network:Error:19] [2020-01-20 12:43:49.59 UTC] Child process jormungandr exited with status 4
[...]

PR #1190, opened 2019-12-16
Add database migration test

rvl:
Relates to #1177.

# Overview

Adds a migration test that can be run against different versions of the wallet.

The idea is to use an api client to set up the database on a wallet server running one version. Then start up the wallet server on the current version and use the api client to check the database.

All the possible upgrade paths from previous released versions to the current version _could_ be exhaustively tested. However, in this initial version, only the direct upgrade from each release to latest is tested.

At present the testing actions are quite basic. It first restores a wallet, migrates, then lists the wallets.

The migration tests are run as part of nightly CI.


lib/core/cardano-wallet-core.cabal
```diff
@@ -104,6 +105,7 @@ library
       Cardano.Pool.Metadata
       Cardano.Wallet
       Cardano.Wallet.Api
+      Cardano.Wallet.Api.Client
```
KtorZ: :+1:

KtorZ:
Looks good overall but I have some hard time grasping the actual test thoroughly by simply reading through the code. I am going to poke it a bit locally to get a better understanding of what's happening.

KtorZ:
There seem to be some nix typo / issues with the nightly script: https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/319#91ac35cc-9906-466f-80a9-1baf6141dd76

rvl:
I renamed the file but then didn't update the yaml. It works now: https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/323#cb027612-c778-427b-812d-53212d77d846

nix/migration-tests.nix
```diff
@@ -0,0 +1,105 @@
...
+      allowFail = true; }
+    { rev = "v2019-12-09";
+      sha256 = "1ld415pbbhj8hg14325w2dcwxkkanm3vx09s7cs1viy8lcl69sln";
+      allowFail = true; }
+    { rev = "v2019-12-13";
+      sha256 = "1a2b4iflwwp824b1k9b82jw2q8pqlar6hg96nv03zv55glkgdllm";
+      allowFail = true; }
+    { rev = "v2019-12-16";
+      sha256 = "0y5xf43lrc7gygvxh5ywkglr3d1xcc19dsskm7frl0v6m9yxzni6"; }
+  ];
```
KtorZ, line 49: :thinking: I am not sure we have to go that far in the past. In particular, `v2019-11-18` is certainly not compatible with `v2019-12-09`, isn't it ?

KtorZ:
Hmmm. The migration-test is using the latest version of the wallet backend API client. Therefore, it fails at parsing older versions of the network information resource:

https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/323#cb027612-c778-427b-812d-53212d77d846/6-10306

```
[migration-test:Error:19] [2019-12-19 08:07:56.99 UTC] Request failed: DecodeFailure "Error in $: key \"next_epoch\" not present" (Response {responseStatusCode = Status {statusCode = 200, statusMessage = "OK"}, responseHeaders = fromList [("Transfer-Encoding","chunked"),("Date","Thu, 19 Dec 2019 08:07:56 GMT"),("Server","Warp/3.2.27"),("Content-Type","application/json;charset=utf-8")], responseHttpVersion = HTTP/1.1, responseBody = "{\"network_tip\":{\"epoch_number\":513520,\"slot_number\":9},\"node_tip\":{\"height\":{\"quantity\":0,\"unit\":\"block\"},\"epoch_number\":0,\"slot_number\":0},\"sync_progress\":{\"status\":\"syncing\",\"progress\":{\"quantity\":0,\"unit\":\"percent\"}}}"})
```

This would also be true if we add some fields to the wallet resource for instance (that wouldn't be resolved when listing). So ideally I'd say that the `migration-test` should also be compiled with multiple versions matching the server version.

rvl:
> The migration-test is using the latest version of the wallet backend API client. Therefore, it fails at parsing older versions of the network information resource.

Yes hmm. I want to always use the migration test setup from the latest version however.

I have changed the test to use untyped requests. So this will usually be compatible with the APIs under test.

>  I am not sure we have to go that far in the past

There are currently a lot of revisions listed in the test harness. These are just for illustration. They are effectively ignored through the `allowFail` flag. Before too soon we will need to consider upgrading from versions other than just the previous version.

lib/jormungandr/test/migration/migration-test.hs
```diff
@@ -0,0 +1,236 @@
...
+    mnem <- mnemonicToText @15 . entropyToMnemonic <$> genEntropy :: IO [Text]
+    let postData = object
+            [ "mnemonic_sentence" .= toJSON mnem
+            , "name" .= testWalletName
+            , "passphrase" .= String "0000000000000000000000000000"
+            ]
+    wal <- post (url apiBase "wallets") (toJSON postData)
+    let walId = wal ^. responseBody . key "id" . _String
+    logInfo tr $ "Create wallet with id " <> walId
+    void $ waitForRestore apiBase walId
```
KtorZ: **[question]**

What is the need for this call? It doesn't seem that the state of the wallet would have any influence on the step2 :thinking: ?

nix/launch-migration-test.sh
```diff
@@ -0,0 +1,27 @@
...
+  echo "usage: $0 (run | step1 | step2)"
+  exit 1
+fi
+
+if [ -z "${stateDir:-}" ]; then
+  echo "$0: set the stateDir environment variable before running this script."
+  exit 1
+fi
+
+: "${src:=$(cd `dirname $0`/..; pwd)}"
```
KtorZ, line 23: Slightly magic at first glance. Had to look up the semantic of the colon `:` which actually still feels like a magic trick (though I guess most bash scripts are in a way a condensed collection of magic tricks :upside_down_face: )

Why not simply ```src=$(cd `dirname $0`/..; pwd)``` ? Do we expect `src` to be possibly defined globally ?

KtorZ (approved):

KtorZ:
Started: https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/343

lib/jormungandr/test/migration/migration-test.hs
rvl: There is no specific need, but generally, migrations are more likely to fail if the DB tables have some rows in them. That's why it creates a wallet and waits for it to restore.

When we actually implement some migrations, we should add extra test data, but that would depend on what the migrations are.

nix/launch-migration-test.sh
rvl: The `src` environment variable that can be set externally so that the script can be used both inside and outside of a nix build.

I thought the `:` for variable default values was standard well-known bash awfulness.

KtorZ: I was not referring to the `:` in `src:=`, but to the leading one. First time I see this of my life to be honest :sweat_smile: ... and I've written quite some bash already.

KtorZ:
--> https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/354

rvl:
I have rebased this on master - [buildkite nightly](https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds?branch=rvl%2F1177%2Fdb-migration-test)

KtorZ (commented):
Nice work with this. I am worried with one thing though, I'd expect this to run into: https://github.com/input-output-hk/cardano-wallet/issues/1251 but it doesn't :thinking: ...

rvl:
> I'd expect this to run into: #1251 but it doesn't

Ah yes, it was using the old version for both step1 and step2, doh. I fixed that. It should now hit an error, then ignore it due to allowFail = true.

Issue #1150, opened 2019-12-09
Fee estimation for payments doesn't take fee balancing after selection into account


KtorZ:
# Context
 
 

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | b13ddbc95  |
| Platform     | All           |
| Installation | N/A             |

When estimate fee for a standard payment, we stop the estimation after the coin selection. However, adjusting for fees may have an influence on the final fee amount (if change outputs are removed after balancing!)

# Steps to Reproduce

N/A

## Expected behavior

- Estimation should take fee balancing into account

## Actual behavior
 

- Estimation doesn't.

---

# Resolution 

- [x] Compute fees as the difference of the result of a payment selection (for which the balancing does occur) ==> #1191 

---

# QA 

- Function modified, integration tests still pass :+1:

PR #1191, opened 2019-12-16
compute payment fee as the difference between inputs/outputs

KtorZ:
# Issue Number

#1150 

# Overview

- [x] I have re-used the existing coin selection function and removing the need for an extra 'estimateFee' function in the core engine

# Comments


KtorZ:
bors try

paweljakubas (approved):
everything now is in `selectCoinsForPayment` and no more need for `estimatePaymentFee`
:100:

Issue #1071, opened 2019-11-27
Wallet not deleted properly (corrupted seq.*.sqlite file  -> StatementAlreadyFinalized "BEGIN" in the log)


piotr-iohk:
# Context
 
 

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | `cardano-wallet-jormungandr.exe version => 2019.11.18` |
| Platform     | Windows         |
| Installation |  (Daedalus: https://buildkite.com/input-output-hk/daedalus/builds/9145)         |

Issue encountered (only once now) in Daedlus by @alan-mcnicholas. Steps to reproduce:

1. Restore two wallets
2. Try and delete one -> strange behavior observed here (i.e. Daedalus was stuck on the delete pop-up)
3. Close Daedalus 
4. Try and reimport wallet -> Daedalus says wallet exists

in the log (`node.log-20191126140040`, [logs-2019-11-26T140733.0183Z.zip](https://github.com/input-output-hk/cardano-wallet/files/3896788/logs-2019-11-26T140733.0183Z.zip)) there are 8 subsequent requests:

```
[DELETE] /v2/wallets/43e9ccb4004771b124c32bee39840de067a882f6
```
1st one     => `200`
the other 7 => `404`

After deletion there are also some failures on `[GET] /v2/wallets` i.e.:

```
[GET] /v2/wallets
StatementAlreadyFinalized "BEGIN"

```

Also the corresponding [seq.43e9ccb4004771b124c32bee39840de067a882f6.zip](https://github.com/input-output-hk/cardano-wallet/files/3896789/seq.43e9ccb4004771b124c32bee39840de067a882f6.zip) file has NOT been deleted as a result of the DELETE action

# Steps to Reproduce

There are no repeatable steps to reproduce. The issue was encountered while using Daedalus on Windows. There was also a poor internet connection.

## Expected behavior

Wallet is deleted. There are no further issues.

## Actual behavior
 
The wallet `seq.*.sqlite` file was not deleted, although the DELETE action seems to return `200`.
The file also got broken somehow in the process. I am able to browse it using https://sqlitebrowser.org/, but when I put it inside my own `wallets` dir in my local cardano-wallet I cannot get any wallets listed. Getting:
```
$ cardano-wallet-jormungandr wallet list
Something went wrong
```
In the log there is:
```
[GET] /v2/wallets
StatementAlreadyFinalized "BEGIN"

```

[...]

PR #1192, opened 2019-12-16
make sure to close and destroy the database before removing files from disk

KtorZ:
# Issue Number

#1071 

# Overview

- [x] I have made sure to close and destroy the database before removing files from disk

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
@@ -23,7 +23,7 @@
 
 module Cardano.Wallet.DB.Sqlite
     ( newDBLayer
-    , mkDBFactory
+    , newDBFactory
     , findDatabases
     , withDBLayer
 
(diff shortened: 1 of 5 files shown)


lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -216,33 +218,38 @@ mkDBFactory
...
+                withDBLayer cfg tracerDB (Just $ databaseFile wid) $ \(ctx,db) -> do
+                    putMVar mvar ctx
+                    action db
+            , removeDatabase = \wid -> do
+                let files =
+                        [ databaseFile wid
+                        , databaseFile wid <> "-wal"
+                        , databaseFile wid <> "-shm"
+                        ]
+                readMVar mvar >>= destroyDBLayer
```
KtorZ, line 242: That's the main change here. Make sure the db is closed before we attempt to remove the files from the disk ^.^ ...

paweljakubas (approved):
LGTM!. It would be good to have ultimate proof here, ie., have a test that really checks that, after clearing a database, the files are gone

Issue #1196, opened 2019-12-16
API Specifications NETWORK - Calculated next epoch is the next epoch


KtorZ:
# Context
 
 

# Test Case

https://buildkite.com/input-output-hk/cardano-wallet/builds/4900#d7f15657-d697-4054-96aa-5b98849a9fce

# Failure / Counter-example

N/A

---

# Resolution 

- #1202 

---

# QA

PR #1199, opened 2019-12-17
API layer house-keeping in preparation of byron Yoroi support

KtorZ:
# Issue Number

N/A 

# Overview

- [x] I have re-organized a the API layer and code to be structured more like the swagger file and be more consistent in naming and declarations.
- [x] I have removed _some_ redundant constraints here and there in the API, there's more to do, but will do in another PR
- [x] I have changed the `getNetworkInformation` handler signature to not take a full `ApiLayer` but only a `NetworkLayer` since that's all it needs. 

# Comments

The diff won't be very readable I think :s ... but there's no logical change, just code moved around.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -492,8 +492,7 @@ server byron shelley spl =
 
 postWallet
     :: forall ctx s t k n.
-        ( DelegationAddress n k
-        , s ~ SeqState n k
+        ( s ~ SeqState n k
         , k ~ ShelleyKey
         , ctx ~ ApiLayer s t k
         , HasDBFactory s k ctx
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -15,8 +15,6 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 
-{-# OPTIONS_GHC -fno-warn-redundant-constraints #-}
-
 -- |
 -- Copyright: © 2018-2019 IOHK
 -- License: Apache-2.0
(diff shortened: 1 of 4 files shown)


paweljakubas (approved):
very nice reshuffling and renaming that contributes to better code understanding : 
1. 1-to1 aligning of API types' names to swagger operations
2. renaming of `rndCtx -> byron` and `seqCtx -> shelley` which is more correct and opens a more clear way to add `yoroi`, etc .... and not to just differentiate on sequential vs random

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -416,58 +409,112 @@ ioToListenError hostPreference portOpt e
...
-
-stakePoolServer
-    :: forall ctx s t n k.
-        ( DelegationAddress n k
+-- | A Servant server for our wallet API
+server
+    :: forall t n byron shelley.
+        ( shelley ~ ApiLayer (SeqState n ShelleyKey) t ShelleyKey
+        , byron ~ ApiLayer (RndState 'Mainnet) t ByronKey
+        -- yoroi ~ ApiLayer (SeqState n ByronKey) t ByronKey
```
rvl, line 417: [question] What's the reason for putting these types in the constraints, rather than directly, or as type aliases? e.g.

```haskell
type ShelleyApiLayer (n :: NetworkDiscriminant) t = ApiLayer (SeqState n ShelleyKey) t ShelleyKey
```

rvl (approved):
This will unfortunately merge-conflict with #1190, but it's a worthy refactor.

rvl:
Fixed a ghc warning in Server.hs.

The SQLite tests have gone flaky so this may not succeed.

bors r+

KtorZ:
Another instance of #1196

KtorZ:
bors retry

lib/core/src/Cardano/Wallet/Api/Server.hs
KtorZ: Actually no particular reason. But such type alias may be justified indeed :+1:

Issue #1200, opened 2019-12-17
Rollbacks issue in stake pool worker


KtorZ:
# Context
 
 

Relates to #1181. 

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | v2019-12-16 46391803d2a35aa7608d46988411bed13d98b92c  |
| Platform     | All           |
| Installation | N/A             |

# Steps to Reproduce

1. If the node rolls back to a previous epoch, it may happen that it discards a few delegation certificates. However, the database can only handle rollbacks with an "epoch precision" (it discards data that are strictly or greater to the epoch of rollback). But, we may still discover certificates that are known within the same epoch and, end up in situation like the following:

```
SQLite3 returned ErrorConstraint while attempting to perform step: UNIQUE constraint failed: pool_registration.pool_id
```

## Expected behavior

- Rollbacks doesn't lead the stake pools database in a blocked state. 

## Actual behavior
 

- If this error is encountered, we'll simply keep looping and re-run into this error forever. 

---

# Resolution 

- Modified the database (in a non backward-compatible) way to store slot id instead of mere epoch number allowing for handling rollbacks correctly ==> #1201 

- Handled db unsafe migration for stake pools by re-creating the pool database from scratch if an old database is encountered (forcing a re-sync) ==> #1201 

---

# QA 

- Regression test added to make sure that migrations from previous versions work seamlessly: https://github.com/input-output-hk/cardano-wallet/blob/cf72aea27790efa8dc2f388d4c314ade85b8a0e4/lib/core/test/unit/Cardano/Pool/DB/SqliteSpec.hs#L57-L84

- Modify the rollback property to work with slot id instead of epoch, failed initially, green after the fix: https://github.com/input-output-hk/cardano-wallet/blob/cf72aea27790efa8dc2f388d4c314ade85b8a0e4/lib/core/test/unit/Cardano/Pool/DB/Properties.hs#L383-L414

PR #1201, opened 2019-12-18
Fix pool registration rollback

KtorZ:
# Issue Number

#1200

# Overview

- commit 18852c0d1aadbd9e4c47dadad2043f516aba3a55
    modify rollback property to capture the rollback in-within epoch problem. It now fails

- commit 3a4d4fd317b9a8813613ee5491408285d340e4d2
    store registration with full slot and not only epoch to allow correct rollback!
    
    This is however a DB breaking change which requires migration...
    Yet, there's no way to migrate from an existing database to this new
    schema. Since however this is a preferable solution and, re-syncing
    the pools data is cheap, we'll simply discards any existing database
    that has an invalid format and, resync from scratch :)

- commit f21f4e3bc39606e5be6d8cf7dc2f30057968d192
    add a regression test to show that migration from an old database doesn't work out of the box
    Currently fails with:
    
    ```
    PersistError 
    Database migration: manual intervention required.
    The unsafe actions are prefixed by '***' below:
    CREATE TEMP TABLE \"pool_registration_backup\"(\"pool_id\" VARCHAR NOT NULL,\"slot\" INTEGER NOT NULL,\"margin\" INTEGER NOT NULL,\"cost\" INTEGER NOT NULL, PRIMARY KEY (\"pool_id\"));
    INSERT INTO \"pool_registration_backup\"(\"pool_id\",\"margin\",\"cost\") SELECT \"pool_id\",\"margin\",\"cost\" FROM \"pool_registration\";
    *** DROP TABLE \"pool_registration\";
    CREATE TABLE \"pool_registration\"(\"pool_id\" VARCHAR NOT NULL,\"slot\" INTEGER NOT NULL,\"margin\" INTEGER NOT NULL,\"cost\" INTEGER NOT NULL, PRIMARY KEY (\"pool_id\"));
    INSERT INTO \"pool_registration\" SELECT \"pool_id\",\"slot\",\"margin\",\"cost\" FROM \"pool_registration_backup\";
    DROP TABLE \"pool_registration_backup\";"
    ```

- commit 440e09930daa229264aead7236e46ca5f34bda71 
    handle migration error in stake pool db layer
    As described in the previous commit.

- commit 112927be39e03376707b1b3ed0a9b9d61ce64535
    better assertions on logs message for db migration tests

# Comments


lib/core/test/unit/Cardano/Pool/DB/SqliteSpec.hs
```diff
@@ -21,20 +23,54 @@ import Cardano.BM.Data.Severity
...
 spec :: Spec
-spec = withDB (newMemoryDBLayer testingLogConfig) $ do
+spec = do
+    withDB (newMemoryDBLayer testingLogConfig) $ do
         describe "Sqlite" properties
 
+    describe "Migration Regressions" $ do
+        test_migrationFromv20191216
+
+test_migrationFromv20191216 :: Spec
```
paweljakubas, line 57: :100:

lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -205,6 +208,9 @@ instance ToText DBLog where
         MsgQuery stmt _ -> stmt
         MsgConnStr connStr -> "Using connection string: " <> connStr
         MsgClosing fp -> "Closing database ("+|fromMaybe "in-memory" fp|+")"
+        MsgDatabaseReset ->
```
paweljakubas, line 211: :+1:

paweljakubas (approved):
PR is remedy for 
```
SQLite3 returned ErrorConstraint while attempting to perform step: UNIQUE constraint failed: pool_owner.pool_id, pool_owner.pool_owner
```
problem and more.
1. It makes pool registration dependent on slot id, and not epoch number, which is important for rollbacks - and produced db problem ^^^. It would be enough to solve this problem BUT would require user intervention which was decided to be avoided. So, we have point 2.
2. It solves it potential user engagement in solving manually db migration (without bothering user) by removing db related files and instantiating again db backend. So maybe there are going to be other cases than the one illustrated in point 1 where avoiding manual db intervention is needed and hence is goes beyond solving this particular problem

lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -39,6 +40,8 @@ import Cardano.Pool.DB
     ( DBLayer (..), ErrPointAlreadyExists (..) )
 import Cardano.Wallet.DB.Sqlite.Types
     ( BlockId (..) )
+import Cardano.Wallet.Logging
+    ( logTrace )
```
paweljakubas: **[change]**

```
import Cardano.Wallet.Logging (logTrace, transformTextTrace)
```

lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -39,6 +40,8 @@ import Cardano.Pool.DB
...
 import Cardano.Wallet.Logging
     ( transformTextTrace )
```
paweljakubas: **[change]**

import Cardano.Wallet.Logging (logTrace, transformTextTrace)

KtorZ:
Nix failed on stylish haskell job :man_shrugging: ... re-running manually...

PR #1202, opened 2019-12-18
Fix flaky next epoch test

KtorZ:
# Issue Number

#1196 

# Overview

- commit 88fd188343407f0f35d8c61b064536c6b15453d4

    simplify calculation of next epoch and remove deadcode
    
    The previous code was a bit more convoluted and re-doing
    time comparison and computations that were unnecessary.
    We know what the next epoch should be, it's the tip + 1.

- commit 42f71363016af4030f844fc31919c621d8877162

    remove usage of 'waitForNextEpoch' to check for the next epoch time

    This function uses the NODE tip to check whether we have moved to the next epoch or not.
    However, the next epoch start time is about where the network is at, so we are comparing
    apple and oranges here. Instead, the test now waits for the time indicated by the API and
    query the network tip right away to check whether we have indeed moved to the next epoch

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -207,7 +207,7 @@ import Control.Monad
 import Control.Monad.IO.Class
     ( MonadIO, liftIO )
 import Control.Monad.Trans.Except
-    ( ExceptT, catchE, except, throwE, withExceptT )
+    ( ExceptT, catchE, throwE, withExceptT )
 import Data.Aeson
     ( (.=) )
 import Data.Function
@@ -242,6 +242,8 @@ import Data.Word
     ( Word32, Word64 )
 import Fmt
     ( Buildable, pretty )
+import GHC.Stack
+    ( HasCallStack )
 import Network.HTTP.Media.RenderHeader
     ( renderHeader )
 import Network.HTTP.Types.Header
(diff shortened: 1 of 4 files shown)


lib/core/test/unit/Cardano/Wallet/Primitive/TypesSpec.hs
```diff
@@ -489,110 +487,6 @@ spec = do
...
-        it "epochFloor (utcTimePred (epochStartTime minBound)) == Nothing" $
-            withMaxSuccess 1000 $ property $ \sps ->
-                epochFloor sps (utcTimePred (epochStartTime sps minBound))
-                    === Nothing
-
-        it "epochFloor (utcTimeSucc (epochStartTime maxBound)) == maxBound" $
-            withMaxSuccess 1000 $ property $ \sps ->
-                epochFloor sps (utcTimeSucc (epochStartTime sps maxBound))
-                    === Just maxBound
-
```
KtorZ, line 595: Now testing deadcode.

KtorZ:
bors try

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1082,16 +1088,11 @@ getNetworkInformation (_block0, bp, st) nl = do
...
-    Nothing -> Left $ ErrNextEpochCannotBeCalculated $ Iso8601Time time
-    Just en -> Right $ ApiEpochInfo
-        { epochNumber = ApiT en
-        , epochStartTime = W.epochStartTime sps en
-        }
+    -- Unsafe constructor for the next epoch. Chances to reach the last epoch
+    -- are quite unlikely in this context :)
+    unsafeEpochSucc :: HasCallStack => W.EpochNo -> W.EpochNo
+    unsafeEpochSucc = fromMaybe bomb . W.epochSucc
+      where bomb = error "reached final epoch of the Blockchain!?"
```
paweljakubas, line 1095: **[silly or not question]**

even, when there is hard fork?

paweljakubas (approved):
LGTM

lib/core/src/Cardano/Wallet/Api/Server.hs
KtorZ: Ho yeah. I mean, the only way we reach the last epoch is in a few thousand years from now. Capitalism would have ended humanity well before that.

PR #1203, opened 2019-12-18
Gratuitous fixes in restoration worker (wrong log line + move pruning inside same tx)

KtorZ:
# Issue Number

# Overview

- [x] I have fixed log line in the restoration function, logging the right checkpoint (i.e. the last one!)
- [x] Moved pruning of the database within the same db transaction as the one inserting checkpoints.. just in case..

# Comments

I noticed this looking at some user logs.

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -607,12 +607,7 @@ restoreBlocks ctx wid blocks nodeTip = db & \DBLayer{..} -> do
         liftIO $ logCheckpoint (NE.last cps)
         putCheckpoint (PrimaryKey wid) (NE.last cps)
 
-    -- NOTE
-    -- We prune the database in a separate db transaction since this is not
-    -- strictly required to maintain integrity of the data. It's quite a heavy
-    -- operation, so we may still want to unlock the database to allow some
-    -- other operations to flow in the meantime.
-    mapExceptT atomically $ prune (PrimaryKey wid)
+        prune (PrimaryKey wid)
 
     liftIO $ do
         progress <- walletSyncProgress @ctx ctx (NE.last cps)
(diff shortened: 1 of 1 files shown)


lib/core/src/Cardano/Wallet.hs
```diff
@@ -604,15 +604,10 @@ restoreBlocks ctx wid blocks nodeTip = db & \DBLayer{..} -> do
                 liftIO $ logCheckpoint cp'
                 putCheckpoint (PrimaryKey wid) cp'
 
-        liftIO $ logCheckpoint cp
+        liftIO $ logCheckpoint (NE.last cps)
```
Anviking, line 607: So it would `Creating checkpoint at` the existing checkpoint instead of the new? 🙃 👍

lib/core/src/Cardano/Wallet.hs
```diff
@@ -604,15 +604,10 @@ restoreBlocks ctx wid blocks nodeTip = db & \DBLayer{..} -> do
...
         putCheckpoint (PrimaryKey wid) (NE.last cps)
 
-    -- NOTE
-    -- We prune the database in a separate db transaction since this is not
-    -- strictly required to maintain integrity of the data. It's quite a heavy
-    -- operation, so we may still want to unlock the database to allow some
-    -- other operations to flow in the meantime.
-    mapExceptT atomically $ prune (PrimaryKey wid)
+        prune (PrimaryKey wid)
```
Anviking, line 610: I guess nothing bad can come from changing this?

Anviking (approved):

lib/core/src/Cardano/Wallet.hs
KtorZ: No it wouldn't :) 
The next line (doing the insert) was fine, but the log line was off.

lib/core/src/Cardano/Wallet.hs
KtorZ: Not really, some _slowness_ maybe, although the pruning is quite fast. And now that we have separate db for each wallet, It should be okay.

lib/core/src/Cardano/Wallet.hs
Anviking: Ah, yes, implicitly meant _log_ `"Creating checkpoint at"`.

PR #1204, opened 2019-12-18
Recovery week - cleaning number 1

paweljakubas:
# Issue Number

Recovery week

# Overview

- [x] I have introduced `servantErrorMsg` and use it in tens of examples

When doing servant bump during sync-tolerance PR month ago there was a need to change tens of error messages : 
```
-            , expectErrorMessage "expected [a], encountered String"
+            , expectErrorMessage "parsing [] failed, expected Array, but \
+                                 \encountered String"

-            , expectErrorMessage "expected Text, encountered Number"
+            , expectErrorMessage "parsing Text failed, expected String, but \
+                                 \encountered Number"
```
So this messages were delocalized and independently used in many places. I think it is good to introduce `servantErrorMsg` in one place and reuse it in those tens of places.  

# Comments

### DIFF
--- a/lib/core/src/Network/Wai/Middleware/ServantError.hs
+++ b/lib/core/src/Network/Wai/Middleware/ServantError.hs
@@ -9,6 +9,7 @@
 
 module Network.Wai.Middleware.ServantError
     ( handleRawError
+    , servantErrorMsg
     ) where
 
 import Prelude
@@ -86,3 +87,7 @@ responseBody = \case
         Nothing
     ResponseStream{} ->
         Nothing
+
+servantErrorMsg :: String -> String -> String
+servantErrorMsg expected received =
+    "expected " <> expected <> ", encountered " <> received
--- a/lib/core-integration/src/Test/Integration/Scenario/API/ByronWallets.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/ByronWallets.hs
@@ -967,7 +967,7 @@ spec = do
         r <- request @ApiByronWallet ctx postByronWalletEp Default payload
         verify r
             [ expectResponseCode @IO HTTP.status400
-            , expectErrorMessage "expected [a], encountered String"
+            , expectErrorMessage (servantErrorMsg "[a]" "String")
             ]
 
     it "BYRON_RESTORE_05 - Num as mnemonic_sentence -> fail" $ \ctx -> do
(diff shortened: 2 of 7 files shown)


lib/core/src/Network/Wai/Middleware/ServantError.hs
```diff
@@ -86,3 +87,7 @@ responseBody = \case
         Nothing
     ResponseStream{} ->
         Nothing
+
+servantErrorMsg :: String -> String -> String
+servantErrorMsg expected received =
+    "expected " <> expected <> ", encountered " <> received
```
KtorZ, line 93: **[remark]**

Hey! I've got several things for which I have mixed feelings here ^.^ 

- Such function doesn't feel in the right place. The module provides a Wai middleware for wrapping raw servant error into JSON objects. But it does little care about error messages so to speak. A better home would probably be: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core-integration/src/Test/Integration/Framework/TestData.hs#L49

- This error "expected X, encountered Y" is actually not coming from `servant` but it's coming from `aeson`. See [typeMismatch](https://hackage.haskell.org/package/aeson-1.4.5.0/docs/Data-Aeson-Types.html#v:typeMismatch)

- The signature is very much stringly-typed. If introducing anything like this, I'd suggest to reuse `typeMismatch` from Aeson, and perhaps, provide an extra function in addition to convert a `Parser a -> String` though that feels a bit weird. 

- A better fix would actually be to either:
    - find a way to not assert error messages (though with Aeson is very much stringly typed as well :s)
    - Or, actually define a test helper function to perform these tests! So, instead of 

    ```hs
        it "ApiT AddressPoolGap (not a integer)" $ do
            let msg = "Error in $: expected Integer, encountered floating number\
                    \ 2.5"
            let msg = "Error in $: " <> servantErrorMsg "Integer" "floating number 2.5"
            Aeson.parseEither parseJSON [aesonQQ|
                2.5
            |] `shouldBe` (Left @String @(ApiT AddressPoolGap) msg)
    ```

    one could write:

    ```hs
    testAesonParseError (Proxy (ApiT AddressPoolGap)) (toJSON 2.5) 
    ```

    The _right_ error message could probably be inferred from an `Arbitrary` instance on the given type, and a `ToJSON`. If that turns out to be too complicated, then we could imagine give the expected string type as an argument.

paweljakubas: @KtorZ thanks! yeah, of course it comes from aeson :upside_down_face: I will try to go into `testAesonParseError` direction

paweljakubas:
After second look, I think it is not needed to change at this moment. And introduce additional function that hide just `Aeson.parseEither parseJSON` and `shouldBe`

PR #1205, opened 2019-12-19
Review and clean State Machine file

paweljakubas:
# Issue Number

Recovery week 

# Overview

- [x] I have reviewed the file and removed superfluous type alias
- [x] I have added a case of wallet creation with some tx history along with already present case of empty Tx history

# Comments

### DIFF
--- a/lib/core/test/unit/Cardano/Wallet/DB/StateMachine.hs
+++ b/lib/core/test/unit/Cardano/Wallet/DB/StateMachine.hs
@@ -194,10 +194,6 @@ import qualified Test.StateMachine.Types.Rank2 as Rank2
 -- types.
 type Mock s = Database MWid s MPrivKey
 
--- | Shortcut for wallet type.
--- TODO: Shortcut is now potless. Remove.
-type MWallet s = Wallet s
-
 -- | Mock wallet ID -- simple and easy to read.
 newtype MWid = MWid String
     deriving (Show, Eq, Ord, Generic)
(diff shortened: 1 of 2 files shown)


lib/core/test/unit/Cardano/Wallet/DB/StateMachine.hs
```diff
@@ -350,9 +349,12 @@ runIO db@DBLayer{..} = fmap Resp . go
...
         CleanDB -> do
             Right . Unit <$> cleanDB db
-        -- FIXME Create a wallet with already some tx history
-        CreateWallet wid wal meta -> catchWalletAlreadyExists (const (NewWallet (unMockWid wid))) $
+        CreateWallet wid wal meta ->
+            catchWalletAlreadyExists (const (NewWallet (unMockWid wid))) $
             mapExceptT atomically $ initializeWallet (widPK wid) wal meta mempty
+        CreateWalletWithTxs wid wal meta txs ->
+            catchWalletAlreadyExists (const (NewWallet (unMockWid wid))) $
+            mapExceptT atomically $ initializeWallet (widPK wid) wal meta txs
```
KtorZ, line 351: *[suggestion]* 

This is a more general case of `CreateWallet`. What about simply fixing the generator for `CreateWallet` instead of creating a new action? Certainly, the generator will sometime yield wallet creation with no txs (and we can even tag it to convince ourselves). So, no need for two separate commands here :)

KtorZ (approved):
LGTM, modulo the superfluous command that is simply a generalization of an existing one IMO.

lib/core/test/unit/Cardano/Wallet/DB/StateMachine.hs
paweljakubas: yes, just extended CreateWallet and resigned from CreateWalletWithTxs

PR #1206, opened 2019-12-19
Make api handlers more polymorphic to reduce duplication

KtorZ:
# Issue Number

N/A

# Overview

- [ ] I have made api handlers more polymorphic to reduce duplication

 It was okay-ish to duplicate handlers for Byron random wallets in order
to quickly iterate and develop handler for those. However, a lot of the
logic was duplicated in the process and, with the upcoming support of
Yoroi wallets, we would need even more. However, this commit take
re-locate the discrepency between wallets types to arguments passed to
other more generic functions. This way, we can group the logic for
top-level functions and adjust how wallets are represented when needed.

This enables rather seamless support for Yoroi wallets, pretty much out
of the box.

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -665,7 +665,7 @@ deleteWallet ctx wid = db & \DBLayer{..} -> do
 -- Rather than force all callers of 'readWallet' to wait for fetching the
 -- account balance (via the 'NetworkLayer'), we expose this function for it.
 fetchRewardBalance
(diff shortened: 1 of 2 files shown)


KtorZ:
bors try

paweljakubas (approved):
very nice that getWallet lost `s ~ SeqState` and is reused for byron/shelley and is ready to service yoroi. Also the choice of its return type, ie., `Handler (APiWallet, UTCTime)` enables removing the same functions *withCreationTime

Issue #1153, opened 2019-12-10
Misleading Mnemonic Sentence Error message


craigem:
# Context
 
 

When running `cardano-wallet-jormungandr` from the CLI, if you submit a valid English word that is not in the the list of valid words, your are greeted with this misleading message:

```
$ cardano-wallet-jormungandr wallet create fruity
Please enter a 15–24 word mnemonic sentence: one two three four five six seven eight nine diary twenty coin regret cry thumb glare tribe slice angle pitch boy ankle resource broken
Found invalid (non-English) word: "four".
Please enter a 15–24 word mnemonic sentence:
```

This also happens at the second factor:

```
$ cardano-wallet-jormungandr wallet create fruity
Please enter a 15–24 word mnemonic sentence: diary twenty coin regret cry thumb glare tribe slice angle pitch boy ankle resource broken
Please enter a 9–12 word mnemonic second factor: one two three four five six seven eight nine
Found invalid (non-English) word: "four".
```

This example goes on for many valid English words.

A more accurate error would instead advise that the the selected words are not on the approved word list and perhaps even suggest using `cardano-wallet-jormungandr mnemonic generate` for a valid list.

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | 2019.12.9 (git revision: 0000000000000000000000000000000000000000) |
| Platform     | Linux |
| Installation | GitHub (via IOHK Nix) |

# Steps to Reproduce

1. As per above examples.

## Expected behavior

An error message that accurately describes the problem

## Actual behavior
 

As per above, a misleading error message

---

# Resolution 

Providing an error that accurately directs the user to correct the issue.

---

# PR

| Number   | Base      |
| ---      | ---       |
| #1212        | `master` |

# QA

KtorZ:
Indeed! It might not be obvious that a mnemonic sentence is an encoding made of a composition of words coming from a known pre-established list and isn't simply free-form. 

We could even point to some documentation like https://github.com/bitcoin/bips/blob/master/bip-0039/english.txt

cc @piotr-iohk

PR #1212, opened 2019-12-20
Refine mnemonic error msg

paweljakubas:
# Issue Number

#1153 

# Overview

- [x] I have refined mnemonic error msg in case when the word in missing in dictionary and reviewed the integration tests affected
- [x] I added english.txt dictinary to specifications/mnemonic

# Comments


paweljakubas:
bors try

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -460,7 +460,9 @@ instance
                 <> show (natVal (Proxy :: Proxy mw))
                 <> " words are expected."
             ErrDictionary (ErrInvalidDictionaryWord w) ->
-                "Found invalid (non-English) word: \"" <> B.toList w <> "\"."
+                "Found invalid (ie., not present in https://github.com/bitcoin\
+                \/bips/blob/master/bip-0039/english.txt) word: \"" <> B.toList w
```
KtorZ: **[suggestion]**

What about a bit more prose here?

e.g.

```
 "Found an unknown word not present in the pre-defined dictionary:" <> B.toList w <> ". The full dictionary is available here: ..."
```

Also, I'd suggest to put a link that we are in control of. So, either a fork of the bips repository owned by IOHK or, simply move the list to the wallet backend repository or wiki.

paweljakubas:
bors try

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
paweljakubas: ok, done

paweljakubas:
bors try

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -460,7 +460,10 @@ instance
                 <> show (natVal (Proxy :: Proxy mw))
                 <> " words are expected."
             ErrDictionary (ErrInvalidDictionaryWord w) ->
-                "Found invalid (non-English) word: \"" <> B.toList w <> "\"."
+                "Found an unknown word not present in the pre-defined dictionary\
+                \: \"" <> B.toList w <> "\". The full dictionary is available \
+                \here: https://github.com/input-output-hk/cardano-wallet/tree/m\
+                \aster/specifications/mnemonic/english.txt"
```
KtorZ, line 466: **[remark]** 

Alternatively, I just thought that the dictionary is also available here: https://github.com/input-output-hk/cardano-crypto/blob/master/src/Crypto/Encoding/BIP39/English.hs#L33

KtorZ (approved):

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
paweljakubas: ok, thanks. I will stick to the current solution. This one w control. The one in cardano-crypto can change (for example it can no longer be line number 33)

paweljakubas:
bors retry

Issue #1161, opened 2019-12-11
Inventory Jörmungandr on-chain types & Haskell equivalents


Anviking:
# Context

> As a wallet backend developer and a cardano-node project manager
I want to identify gaps and outstanding effort required to integrate the wallet backend with the haskell nodes,
So that I can anticipate risks and start planning the integration.

[WB-26](https://jira.iohk.io/projects/WB/issues/WB-26?filter=allopenissues)

# Decision

Make an inventory of all the types we retrieve from Jormungandr and look for Haskell equivalents.

# Acceptance Criteria

- An inventory of elements used from Jörmungandr on-chain representation _must_ be clearly established (fee policy? blocks? transactions?)
- Where possible, we should find their equivalent for the Haskell nodes
- The "Haskell Nodes" section of our resources wiki page _may_ be updated with new elements (https://github.com/input-output-hk/cardano-wallet/wiki/Resources#haskell-node)

---

# Development 

# QA

KtorZ:
@Anviking , can you make sure to wrap-up this one today? The content of the draft PR should come here in the ticket instead.

Anviking:
# Integrating cardano-wallet with the new Byron Haskell node

Goal: identify gaps and risks for integrating with the Haskell node.

Structure:
- [Inventory of Jörmungandr on-chain types we use](#Inventory)
- [Discussion](#Discussion)
- [Other, fuzzier difficulties of integration](#other-fuzzier-differences-of-integration)
- [Summary Byron](#Summary-Byron)
- [Extra: A glimpse into Shelley](#extra-A-glimpse-into-shelley)

## Inventory

### Protocol Parameters & Constants

These are the protocol parameters and constants the wallet is currently using:

| Parameter | Description | Exact equivalent in Haskell Byron Node |
| ----- |:-------------| --- |
| genesisBlockHash | Hash of the very first block | ✔ |
| genesisBlockDate | Start time of the chain. | ✔ |
| feePolicy | Policy regarding transaction fee. | ❌ |
| slotLength | Length, in seconds, of a slot. | ✔ |
| epochLength | Number of slots in a single epoch. | ✔ |
| txMaxSize | Maximum size of a transaction (soft or hard limit). | ✔ |
| epochStability | Length of the suffix of the chain considered unstable | ✔ |
| activeSlotCoefficient | In Genesis/Praos, corresponds to the % of active slots (i.e. slots for which someone can be elected as leader). | ❌ |

[...]

PR #1214, opened 2019-12-20
[Do not merge] Inventory of Jörmungandr types & Haskell node equivalents

Anviking:
# Issue Number

#1161 

# Overview

- Inventory
- I managed to restore a wallet on mainnet via the Haskell node. I added a section in the document for difficulties I discovered here. (empty ticket: #1243)

# Comments

- Press "view file" in the diff for a rendered version.
- I put this in a PR such that one could comment on individual lines if desired.

### DIFF
--- a/Inventory.md
+++ b/Inventory.md
@@ -65,6 +65,14 @@ In Jörmungandr a slot can only be inhabited by a single block. On the Haskell s
 
 There is no limit on how far the Jörmungandr nodes can rollback. The Haskell Nodes have a limit: (`k` blocks). Haskell nodes have a clear separation between mutable and immutable parts of the chain.
 
+### Genesis block
+
+The `block0` of Jörmungandr contains the initial UTxO distribution and protocol parameters. The wallet retrieves it from Jörmungandr REST API like other blocks, but the wallet then treats it specially; to even start the wallet server, we first need the `block0`.
+
+The Haskell node's initial protocol parameters and initial UTxO can be derived from the "Genesis Config", which is read from a config file ([example](https://github.com/input-output-hk/cardano-ledger/blob/0ad59ce46b3da8af6f6af97064e7c77cc7c2ac8b/cardano-ledger/mainnet-genesis.json#L1)).
+
+So from the wallet's perspective, Jörmungandr and Haskell are very similar, except for *how* the genesis data is retrieved. Small ❌.
+
 ### Transactions
 
 Transaction inputs in Jörmungandr consist of (txHash, index, coinValue). In the Haskell Byron implementation they are only (txHash, index). ❌
@@ -150,8 +158,6 @@ We cannot implement delegation features. We could
 
 I tried to get a wallet executable working with the Haskell node and discovered new kinds of difficulties.
 
-- Chain parameters are not in the Haskell genesis block. The wallet will need to hard-code or read them from a config file at startup.
-- The genesis block does not have to be treated differently because of above. The Jörmungandr wallet currently treats it completely separate.
 - NetworkLayer and `follow` seem like the wrong abstractions to work with both Jörmungandr and Haskell.
 - If `follow` is removed from core, `monitorStakePools` breaks (or will have be made jormungandr-specific).
 - Both CLI and API will differ. (E.g. different configuration and CLI arguments)
--- a/Inventory.md
+++ b/Inventory.md
@@ -174,5 +174,20 @@ I tried to get a wallet executable working with the Haskell node and discovered 
 
 ## Extra: A glimpse into Shelley
 
-- We need to support changing protocol parameters.
-- Delegation features will likely work differently. E.g through imported ledger-rules or by asking for certain ledger state once per epoch. (See Duncan's messages: https://input-output-rnd.slack.com/archives/C819S481Y/p1576776488005100) The very good thing is that we will be able to ask for the ledger state of any point in the unstable chain. In Jörmungandr we could only retrieve the latest stake-distribution and we didn't know what point it belonged to when we got it. This was very tricky to deal with.
+We need to support changing protocol parameters. Most likely via a function from cardano-ledger, as specified in the [WB-44](https://jira.iohk.io/secure/RapidBoard.jspa?rapidView=46&projectKey=WB&view=planning&selectedIssue=WB-44&issueLimit=100) proposal.
+
+We will use different mechanisms for retrieving data than for the jörmungandr delegation features:
+
+> For protocol params, stake pools etc, you have a choice:
+>
+> Use the ledger rules (via the lib) to maintain the appropriate subset of the ledger state (everything other than the big things like utxo, reward accounts & stake distribution). With this approach you always have locally the full values and they are automatically in-sync with the block you're up to.
+>
+> Don't use the ledger rules, and every time you want to find some bit of the ledger state (other than the wallet utxo), you can use the local state query protocol to get reference to a snapshot of the ledger state as of a recent block, and then run queries against it to get what you want.
+>
+>[...]
+
+>If it were not for the reward accounts I'd say don't even think about 2 and do everything via 1.
+
+—- [Duncan](https://input-output-rnd.slack.com/archives/C819S481Y/p1576777084005800?thread_ts=1576776488.005100&cid=C819S481Y)
+
+In Jörmungandr we could only retrieve the latest stake-distribution and we had to cleverly make sure that it corresponded to the same chain as our local chain. With the `LocalStateQuery` protocol we should be able to ask for the ledger state of any point within `k` to the tip. This is great as we won't have to be concerned about race-conditions.
--- a/Inventory.md
+++ b/Inventory.md
@@ -127,7 +127,7 @@ With a good abstraction we could remove the prevBlockHash field in BlockHeader, 
 
 We can probably get by with magic values to represent the `Origin`-case. ✅
 
-A way to treat points (like ourobouros-network's`Point ByronBlock`) more abstract in the wallet could potentially be needed to avoid sabotaging other abstractions.
+A way to treat points (like ourobouros-network's`Point ByronBlock`) more abstract in the wallet could potentially be needed to avoid polluting other abstractions.
 
 ### SlotId vs SlotNo
 
--- a/Inventory.md
+++ b/Inventory.md
@@ -1,0 +1,129 @@
+# Integrating cardano-wallet with the new Byron Haskell node
+
+Goal: identify gaps and risks for integrating with the Haskell node.
+
+Structure:
+- [Inventory of Jörmungandr on-chain types we use](#Inventory)
+- [Discussion](#Discussion)
+- [Summary Byron](#Summary-Byron)
+- [A glimpse into Shelley](#A-glimpse-into-shelley)
+
+## Inventory
+
+### Protocol Parameters & Constants
+
+These are the protocol parameters and constants the wallet is currently using:
+
+| Parameter | Description | Exact equivalent in Haskell Byron Node |
+| ----- |:-------------| --- |
+| genesisBlockHash | Hash of the very first block | ✔ |
+| genesisBlockDate | Start time of the chain. | ✔ |
+| feePolicy | Policy regarding transaction fee. | ❌ |
+| slotLength | Length, in seconds, of a slot. | ✔ |
+| epochLength | Number of slots in a single epoch. | ✔ |
+| txMaxSize | Maximum size of a transaction (soft or hard limit). | ✔ |
+| epochStability | Length of the suffix of the chain considered unstable | ✔ |
+| activeSlotCoefficient | In Genesis/Praos, corresponds to the % of active slots (i.e. slots for which someone can be elected as leader). | ❌ |
+
+Genesis and slotting information should trivially be the same. 
+
+The fee policies work differently. 
+
+Jörmungandr charges `constant + coeff * (inputs + outputs) + optional_cert_specific_fee`
+
+The Haskell Byron node charges `constant + coeff * tx_size_in_bytes`
+
+some links:
+
+- [Wallet BlockchainParameters](https://github.com/input-output-hk/cardano-wallet/blob/468e0fb14ed3b50ab3d16ba127f24242a41ed126/lib/core/src/Cardano/Wallet/Primitive/Types.hs#L1084-L1102)
+- [cardano-ledger ProtocolParameters](https://github.com/input-output-hk/cardano-ledger/blob/0ad59ce46b3da8af6f6af97064e7c77cc7c2ac8b/cardano-ledger/src/Cardano/Chain/Update/ProtocolParameters.hs#L32-L62)
+- [wallet use of Jörmungandr fee policy](https://github.com/input-output-hk/cardano-wallet/blob/468e0fb14ed3b50ab3d16ba127f24242a41ed126/lib/jormungandr/src/Cardano/Wallet/Jormungandr/Transaction.hs#L105)
+- [doc of Jörmungandr fee policy: format.md](https://github.com/input-output-hk/chain-libs/blob/master/chain-impl-mockchain/doc/format.md)
+- [cardano-ledger TxFeePolicy](https://github.com/input-output-hk/cardano-ledger/blob/0ad59ce46b3da8af6f6af97064e7c77cc7c2ac8b/cardano-ledger/src/Cardano/Chain/Common/TxFeePolicy.hs#L41-L59)
+
+
+### Protocol Parameter Updates
+
+Jörmungandr does not fully support protocol parameter updates. If protocol parameters can change, the wallet needs to be aware.
+
+- Jörmungandr: static (updates not implemented)
+- Haskell Byron: static (updates possible, but not likely)
+- Haskell Shelley: dynamic (updates possible, wallet should handle this)
+
+### Blocks
+
+The Jörmungandr wallet uses the `BlockHeader = (slotId, hash, prevHash)` record to identify
+points on the chain.
+
+The Byron Haskell node only *needs* `(slotNo, hash)` to start giving us blocks, but 
+`prevHash` is available. Small ❌.
+
+Byron `SlotNo` counts slots from genesis in a single number. The current Jörmungandr-compatible wallet `SlotId` counts the epoch AND the slot in that epoch. ❌
+
+### Transactions
+
+Transaction inputs in Jörmungandr consist of (txHash, index, coinValue). In the Haskell Byron implementation they are only (txHash, index). ❌
+
+### Delegation certificates (Shelley)
+
+There are delegation certificates in the Byron Haskell node.
+[haskell cardano-ledger `Certificate`](https://github.com/input-output-hk/cardano-ledger/blob/0ad59ce46b3da8af6f6af97064e7c77cc7c2ac8b/cardano-ledger/src/Cardano/Chain/Delegation/Certificate.hs#L87-L99<Paste>)
+I think they might be there for compatibility and be nothing we should concern ourselves with. (TODO: confirm)
+
+The Jörmungandr wallet looks at delegation certificates to determine if, and to what it is delegating. This is pointless in Byron.
+
+### Pool registrations (Shelley)
+
+The wallet looks at pool registration certificates to collect metadata (pool-owners), for when it needs to list stake pools.
+
+### Block producer (Shelley)
+
+The stake pool metrics worker needs to know the minter of each block. In Byron there are no pools, so the worker would be pointless.
+
+## Discussion
+
+###  FeePolicy
+
+We supported cardano-http-bridge in the past. Existing abstractions should already be able to deal with this difference. It *should* not be that much trouble, but it could be. ⚠️
+
+Making sure related functionality (fee-estimation, coin selection) could maybe be subtly difficult. (@paweljakubas thoughts?)
+
+### Tx inputs
+
+The same goes for the different transaction inputs:
+- We previously had a `Tx` abstraction for supporting inputs with and without `coin`-values. Removing it was conceptually simple but touched a lot of code. Before re-adding it we could take the opportunity to think of alternative solutions. ⚠️
+- The coins in the Jörmungandr tx inputs are exposed in the Wallet API. Something needs to be changed to be able to serve Haskell Byron tx inputs in the API. If we want to continue exposing them in for the Jörmungandr-compatible wallet we need some added abstraction in the wallet API code. ⚠️
+
+### Active slot coeff
+
+While missing in Byron, it will exist in Shelley. We could set it to 100% internally in the wallet. Should be a problem at all. ✅
+
+### Superfluous prevBlockHash in BlockHeader
+
+With a good abstraction, or dropped Jörmungandr support we could remove the prevBlockHash field in BlockHeader, but it does no damage there. ✅
+
+### SlotId vs SlotNo
+
+When slotsPerEpoch cannot change the conversion between SlotId and SlotNo is simple.
+
+But to support protocol parameter updates, we should consider making the core wallet logic deal with SlotNo for the Haskell node.
+
+For Byron, we can assume they are static. ✅
+
+### Delegation features
+
+We cannot implement delegation features. We could
+1. Keep the API the same, and use dummy implementations internally. E.g return an empty list of stake pools. E.g say that wallets have 0 account balance. I think this should work without any hustle.
+2. Make the wallet modular enough to not have to do the above. I am not sure how difficult this would be.
+
+### Summary Byron
+- Deal with absence of `coin` value in transaction inputs (both internally and in the API)
+- Deal with of tx fee policy working differently
+- Absence of delegation features requires dummy implementation or re-designed API for Byron
+- A few small problems with trivial solutions.
+
+### A glimpse into Shelley (TODO)
+
+- We need to support changing protocol parameters.
+- Delegation features will likely work differently.
+- Interesting thread: https://input-output-rnd.slack.com/archives/C819S481Y/p1576776488005100 
--- a/Inventory.md
+++ b/Inventory.md
@@ -5,9 +5,9 @@ Goal: identify gaps and risks for integrating with the Haskell node.
 Structure:
 - [Inventory of Jörmungandr on-chain types we use](#Inventory)
 - [Discussion](#Discussion)
+- [Other, fuzzier difficulties of integration](#other-fuzzier-differences-of-integration)
 - [Summary Byron](#Summary-Byron)
 - [Extra: A glimpse into Shelley](#extra-A-glimpse-into-shelley)
-- [Extra: Other, fuzzier differences](#extra-other-fuzzier-differences)
 
 ## Inventory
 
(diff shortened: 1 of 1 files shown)


Anviking:
Hey, @paweljakubas, this should be somewhat ready to be looked at.

If I manage to make a more extensive prototype, it might influence / validate / invalidate things here. I think I will try to do that *before* potentially looking into things deeper. (But also possible that one can call it kind-of done)

But I'd appreciate to hear what you think.

Anviking:
@paweljakubas want to have a look?

Anviking:
Since my comment from before Christmas I did this:

> I managed to restore a wallet on mainnet via the Haskell node. I added a section in the document for difficulties I discovered here.

paweljakubas:
hey @Anviking  I will look at this today. need to refresh the material and confront every statement with code

Inventory.md
```diff
@@ -0,0 +1,173 @@
...
+
+The stake pool metrics worker needs to know the minter of each block. In Byron there are no pools, so the worker would be pointless.
+
+## Discussion
+
+###  FeePolicy
+
+We supported cardano-http-bridge in the past. Existing abstractions should already be able to deal with this difference. It *should* not be that much trouble, but it could be. ⚠️
+
+Making sure related functionality (fee-estimation, coin selection) could maybe be subtly difficult. (@paweljakubas?)
```
paweljakubas: I do not see problem with coin selection at all - this is kinda independent. for fee estimation we should just take haskell node hard-coded A and B in A+B*x and do what we did when added support for flat fee for delegation certificates, ie., add new function;-)

Inventory.md
```diff
@@ -0,0 +1,173 @@
...
+### Addresses
+
+We can trivially re-encode addresses to work with the wallet core and re-decode them later
+(when checking if they are ours or not).
+
+We might be able to treat addresses more abstractly and avoid this.
+
+### Active slot coeff
+
+While missing in Byron, it will exist in Shelley. We could set it to 100% internally in the wallet. Should be a problem at all. ✅
```
paweljakubas: you meant `should not be a problem at all` , right? ;-)

Inventory.md
```diff
@@ -0,0 +1,173 @@
...
+
+For mainnet Shelley, EBBs do provide an easy way of knowing when a new epoch starts without having to keep track of protocol parameters (and when slots are counted from genesis). This would be an argument for not filtering them out. We will however most likely be maintaining up-to-date protocol parameters anyway, and would be able to tell when we're entering a new epoch through the protocol-params and the current slotNo.
+
+### No rollbacks further than "k"
+
+With the Haskell node we will be able to say that old enough state/checkpoints are immutable and can never be rolled back. We might want to change which checkpoints we store in DB. There is no point in keeping multiple checkpoints in the immutable part of the chain.
+
+### Delegation features
+
+We cannot implement delegation features. We could
```
paweljakubas, line 154: recent Matthias endpoint type magic refactoring could help us here probably

Inventory.md
```diff
@@ -0,0 +1,173 @@
...
+1. Keep the API the same, and use dummy implementations internally. E.g return an empty list of stake pools. E.g say that wallets have 0 account balance. I think this should work without any hustle.
+2. Make the wallet modular enough to not have to do the above. I am not sure how difficult this would be.
+
+## Other, fuzzier difficulties of integration
+
+I tried to get a wallet executable working with the Haskell node and discovered new kinds of difficulties.
+
+- Chain parameters are not in the Haskell genesis block. The wallet will need to hard-code or read them from a config file at startup.
+- The genesis block does not have to be treated differently because of above. The jörmungandr wallet currently treats it completely separate.
+- NetworkLayer and `follow` seem like the wrong abstractions to work with both jörmungandre and Haskell.
```
paweljakubas: why `follow` is wrong? could you give more details on why you think this is wrong abstraction? And maybe sketch of alternative?

Inventory.md
```diff
@@ -0,0 +1,173 @@
...
+- Need to deal with of tx fee policy working differently
+- Absence of delegation features requires dummy implementation or re-designed API for Byron
+- There are a few small problems with trivial immediate solutions.
+- There seem to be several fuzzier differences. For instance, the wallet `NetworkLayer` seem to be the wrong abstraction.
+- Some of the fuzzier difficulties might lack cheap temporary solutions. More careful consideration might be useful there, re-evaluating current architecture while keeping future goals in mind.
+
+## Extra: A glimpse into Shelley
+
+- We need to support changing protocol parameters.
+- Delegation features will likely work differently. E.g through imported ledger-rules or by asking for certain ledger state once per epoch. (See Duncan's messages: https://input-output-rnd.slack.com/archives/C819S481Y/p1576776488005100) The very good thing is that we will be able to ask for the ledger state of any point in the unstable chain. In Jörmungandr we could only retrieve the latest stake-distribution and we didn't know what point it belonged to when we got it. This was very tricky to deal with.
```
paweljakubas: so every time we interact with with node, we will need to ask for protocol parameters. Or haskell node has push mechanism (one can subscribe for change of protocol parameters event) - I am asking as they put huge effort in networking layer so maybe there is a way

paweljakubas (approved):
Great comparison! We should really merge it IMO or persist it somewhere - it will act as blueprint and high level view of perspective of what is ahead.. I was surprised by a number of things, like
1. different way of fee calculation between rust and haskell nodes
2. dynamic nature of protocol updating of haskell node and challenges it poses.
3. differences in representing of blocks between nodes (did not get how haskell node deals with a situation that in a given slot there is not block).

Inventory.md
KtorZ: Agreed. There's no problem, the coin selection or fee balancing does not make any assumption about how fees are calculated. The fee policy is given as an argument and so long as we can implement a function from `CoinSelection -> Fee`, everything's fine.

Inventory.md
```diff
@@ -0,0 +1,173 @@
...
+
+- Jörmungandr: static (updates not implemented)
+- Haskell Byron: static (updates possible, but not likely)
+- Haskell Shelley: dynamic (updates possible, wallet should handle this)
+
+### Blocks
+
+The Jörmungandr wallet uses the `BlockHeader = (slotId, blockHeight, hash, prevHash)` record to identify points on the chain.
+
+The Byron Haskell node only *needs* `Point ByronBlock = Origin | At (slotNo, hash)` to start giving us blocks. The wallet can retrieve `blockHeight` and `prevHash` from block headers. But those two fields are not needed to initiate the chain-sync protocol. Small ❌.
```
KtorZ, line 58: I don't really see a problem here to be honest. If our type can do more, then it can certainly do less. We aren't looking for the minimal implementation possible; integrating with multiple backends necessarily create some overhead.

Inventory.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Blocks
+
+The Jörmungandr wallet uses the `BlockHeader = (slotId, blockHeight, hash, prevHash)` record to identify points on the chain.
+
+The Byron Haskell node only *needs* `Point ByronBlock = Origin | At (slotNo, hash)` to start giving us blocks. The wallet can retrieve `blockHeight` and `prevHash` from block headers. But those two fields are not needed to initiate the chain-sync protocol. Small ❌. 
+
+We might need to convert the genesis point `Origin`, to a `BlockHeader` if we are to persist it in the DB ❌
+
+Byron `SlotNo` counts slots from genesis in a single number. The current Jörmungandr-compatible wallet `SlotId` counts the epoch AND the slot in that epoch. ❌
```
KtorZ: This approach (SlotNo) should actually make the handling of protocol updates easier :+1:

Inventory.md
```diff
@@ -0,0 +1,173 @@
...
+
+The Jörmungandr wallet looks at delegation certificates to determine if, and to what it is delegating. This is pointless in Byron.
+
+### Pool registrations (Shelley)
+
+The wallet looks at pool registration certificates to collect metadata (pool-owners), for when it needs to list stake pools.
+
+### Block producer (Shelley)
+
+The stake pool metrics worker needs to know the minter of each block. In Byron there are no pools, so the worker would be pointless.
```
KtorZ, line 96: Yes, and we've built this whole stake pool separately for a reason :) 
This way, we can rather easily create a server / API which does not implement anything regarding pools (modulo some tweaking in the way the "server" gets started).

Inventory.md
```diff
@@ -0,0 +1,173 @@
...
+
+We supported cardano-http-bridge in the past. Existing abstractions should already be able to deal with this difference. It *should* not be that much trouble, but it could be. ⚠️
+
+Making sure related functionality (fee-estimation, coin selection) could maybe be subtly difficult. (@paweljakubas?)
+
+### Tx inputs
+
+The same goes for the different transaction inputs:
+- We previously had a `Tx` abstraction for supporting inputs with and without `coin`-values. Removing it was conceptually simple but touched a lot of code. Before re-adding it we could take the opportunity to think of alternative solutions. ⚠️
+- The coins in the Jörmungandr tx inputs are exposed in the Wallet API. Something needs to be changed to be able to serve Haskell Byron tx inputs in the API. If we want to continue exposing them in for the Jörmungandr-compatible wallet we need some added abstraction in the wallet API code. ⚠️
```
KtorZ: This is incorrect. What is exposed in the API are inputs that can be resolved by the wallet. This is the case for inputs of outgoing transactions for which are still part of our total UTxO. This manipulation applies equally to Haskell nodes :+1: 

Having said that, I'd be inclined to remove this extra complexity since it can only resolve inputs on pending outgoing transactions. As soon as the transaction is inserted in the ledger, the inputs are consumed and can't be resolved anymore. This is really not a useful feature, I'll chat with Daedalus once again about this.

Inventory.md
```diff
@@ -0,0 +1,173 @@
...
+
+In Jörmungandr a slot can only be inhabited by a single block. On the Haskell side, epoch boundary blocks will share slot with the normal block that comes after it. ❌
+
+There is no limit on how far the Jörmungandr nodes can rollback. The Haskell Nodes have a limit: (`k` blocks). Haskell nodes have a clear separation between mutable and immutable parts of the chain.
+
+### Transactions
+
+Transaction inputs in Jörmungandr consist of (txHash, index, coinValue). In the Haskell Byron implementation they are only (txHash, index). ❌
+
+The addresses we get from the Haskell chain-sync protocol are not byte-strings, but rather in their deserialized form.
```
KtorZ, line 80: I don't get this? Addresses can certainly be serialize if needed, right ?

Inventory.md
```diff
@@ -0,0 +1,173 @@
...
+
+We might be able to treat addresses more abstractly and avoid this.
+
+### Active slot coeff
+
+While missing in Byron, it will exist in Shelley. We could set it to 100% internally in the wallet. Should be a problem at all. ✅
+
+### Superfluous prevBlockHash in BlockHeader
+
+With a good abstraction, or dropped Jörmungandr support we could remove the prevBlockHash field in BlockHeader, but it does no damage there. ✅
```
KtorZ: > dropped Jörmungandr support

This can't happen anytime soon though.

Inventory.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Point ByronBlock vs BlockHeader
+
+We can probably get by with magic values to represent the `Origin`-case. ✅
+
+A way to treat points (like ourobouros-network's`Point ByronBlock`) more abstract in the wallet could potentially be needed to avoid sabotaging other abstractions.
```
KtorZ: Sabotaging is a strong word. Use with care.

Inventory.md
```diff
@@ -0,0 +1,173 @@
...
+
+### No rollbacks further than "k"
+
+With the Haskell node we will be able to say that old enough state/checkpoints are immutable and can never be rolled back. We might want to change which checkpoints we store in DB. There is no point in keeping multiple checkpoints in the immutable part of the chain.
+
+### Delegation features
+
+We cannot implement delegation features. We could
+1. Keep the API the same, and use dummy implementations internally. E.g return an empty list of stake pools. E.g say that wallets have 0 account balance. I think this should work without any hustle.
+2. Make the wallet modular enough to not have to do the above. I am not sure how difficult this would be.
```
KtorZ: The wallet ALREADY IS sufficiently modular. We don't have to create a server with the whole API so this is trivially achieved.

Inventory.md
KtorZ: Saying that something is "the wrong abstractions" is one thing. Explaining *why* is actually much more relevant. 
To me, it seems like a relatively good abstraction to work with both backends (modulo the bits related to stake-pools that were added a bit in a hurry and should probably not be part of the network layer in the first place...)

Inventory.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Other, fuzzier difficulties of integration
+
+I tried to get a wallet executable working with the Haskell node and discovered new kinds of difficulties.
+
+- Chain parameters are not in the Haskell genesis block. The wallet will need to hard-code or read them from a config file at startup.
+- The genesis block does not have to be treated differently because of above. The jörmungandr wallet currently treats it completely separate.
```
KtorZ: Hmmm. No it doesn't treat the genesis block differently apart from using it as a starting point when instantiating a wallet. But in essence, it could be any block.

Inventory.md
KtorZ: @paweljakubas the haskell protocols are mostly stateful (i.e., the node remembers where we are at and only changes on explicit requests). 
So, with the haskell nodes, you can easily say:

- Find an intersection between your chain and my local chain
- And now, give me the state of your protocol parameters from this intersection

With jormungandr, finding this intersection is something that WE do within the implementation of the network layer for Jörmungandr. Plus, by the time we've found the intersection, the node has likely moved forward (or worse, switch chain), so by the time you get the actual data in a second HTTP call, you might receive completely outdated data and have to re-sync / find a new intersection and retry. 

To get the sense of it (and how tricky it is in practice), have a look at https://github.com/input-output-hk/cardano-wallet/blob/master/lib/jormungandr/src/Cardano/Wallet/Jormungandr/Network.hs#L383

Fortunately, stateful protocols should make our life MUCH easier because all that logic is now done by the node itself, and our network layer can just blindly follow the node's instructions.

Anviking: > For protocol params, stake pools etc, you have a choice:
> 1. Use the ledger rules (via the lib) to maintain the appropriate subset of the ledger state (everything other than the big things like utxo, reward accounts & stake distribution). With this approach you always have locally the full values and they are automatically in-sync with the block you're up to.
>
> 2. Don't use the ledger rules, and every time you want to find some bit of the ledger state (other than the wallet utxo), you can use the local state query protocol to get reference to a snapshot of the ledger state as of a recent block, and then run queries against it to get what you want. 
>
> [...]
>
> **If it were not for the reward accounts I'd say don't even think about 2 and do everything via 1**.

— [Duncan](https://input-output-rnd.slack.com/archives/C819S481Y/p1576777084005800?thread_ts=1576776488.005100&cid=C819S481Y)

> so every time we interact with with node, we will need to ask for protocol parameters. 

@paweljakubas so no. We will likely know 1) the protocol params at genesis, and 2) when we apply a block we use a function from cardano-ledger to get updated protocol parameters to store in the DB.

From Rodney's [WB-44](https://jira.iohk.io/secure/RapidBoard.jspa?rapidView=46&projectKey=WB&view=planning&selectedIssue=WB-44&issueLimit=100) (Unless I'm missing something subtle this is the same idea):
```Haskell
updateSystemState :: UpdateSystemState -> Block -> UpdateSystemState
```

> And now, give me the state of your protocol parameters from this intersection

@KtorZ I don't believe so. The `LocalStateQuery` protocol will — according to my understanding — allow us to get the full ledger state in the style of `Point -> m FullLedgerState` (but maybe not exactly like that) for points in the **mutable part of the chain**. `LocalStateQuery` is a different protocol from the `ChainSync` protocol (which we have been prototyping with). `LocalStateQuery` does not have anything to do with finding intersects, whereas `ChainSync` provides a `MsgFindIntersect`.

https://github.com/input-output-hk/ouroboros-network/blob/9d8493e6d609f879c9ebc684bc5d830f1f146bb6/ouroboros-network/src/Ouroboros/Network/Protocol/LocalStateQuery/Type.hs#L64-L66

Anviking: I will see if I can make this clearer in the document.

KtorZ: Hmm. Sweet. So this one would be stateless and allow querying the state from any point in time :thinking: ? Sounds like quite powerful and useful.

Inventory.md
Anviking: It is a difference nonetheless. (I tried to separate the differences themselves from my "review" of them. I could have mentioned that in the beginning of the document.)

I don't see why this in particular would make protocol updates easier to handle (for the wallet).

Inventory.md
Anviking: Yes. Still a difference. 

At a glance it seems wasteful to 1. decodeAddress, 2. encodeAddress, 3. decodeAddress. But it's easy to implement.

We could treat addresses more generically to avoid this.

(I say basically this in the discussion, on L106)

Inventory.md
Anviking: > We aren't looking for the minimal implementation possible; integrating with multiple backends necessarily create some overhead.

We need to convert `BlockHeader` to `Origin | At (slotNo, headerHash)`. What special values should correspond to the `Origin` case? Could this be an implementation detail of the DB? Does this now need to be known inside `NetworkLayer`? If we are converting `SlotId -> SlotNo` we now depend on dynamic protocol parameters (though I guess we are likely to change that).

I think we can get by. Still, I *am* — *slightly* — worried that not being abstract enough now could lead to unnecessary complexity later on.

My despair from adding jörmungandr support vaguely comes to mind:

> To add jormungadnr support we had to
> - come up with a tip that works with all targets (BlockHeader)
> - Modify the *core* Wallet s t state & walletlayer etc to support that tip
> - change NetworkLayer to support that tip (edited)

[src](https://input-output-rnd.slack.com/archives/DBHFW31K7/p1560445363080300)

Inventory.md
Anviking: Looking now, the `Server.start` function seems to assume a specific API. But that must be easy to change indeed 👍 

https://github.com/input-output-hk/cardano-wallet/blob/ed10bfeabac2e184a3d090d875846091249a3288/lib/core/src/Cardano/Wallet/Api/Server.hs#L313-L352

Inventory.md
Anviking: >  apart from using it as a starting point when instantiating a wallet

Exactly. If we didn't have to support jörmungandr, the easiest thing would probably be to let wallets be created completely empty, and apply the genesis block (without special treatment) when we start the restoration.

Inventory.md
Anviking: Yes, it's cool. 

> So this one would be stateless and allow querying the state from any point in time

That is my impression. (Note **mutable part of the chain**, or `within K of the tip` though).

Duncan also said that it is **not complete** and that they want to **work with us to decide what is needed**:

> note that the integration of the local state protocol is not complete. While the pattern of "acquire ref, query, query, release" is set, the exact queries you can do remains to be defined, and we'll want to work with your team to decide what's needed.

Also, I was wondering if we couldn't "jump start" restorations by filtering the full UTxO instead of slow chain sync, but as Duncan pointed out, that doesn't work with sequential wallets, since we don't know which addresses to look for.

Inventory.md
Anviking: Yes 👍 

With a related potential problem, though:

> If `follow` is removed from core, `monitorStakePools` breaks (or will have be made jormungandr-specific).

Will try to elaborate elsewhere.

Inventory.md
KtorZ: It's hard to be more generic than `Bytestring` though :grimacing: ... We _could_ make address a type parameter of the wallet engine but.. trust me, no one wants to go through this.

Inventory.md
KtorZ: No. No. No. Being unable to create empty wallet is a feature BY DESIGN. Because "an empty wallet" with no blocks requires handling the initialization as a special case, which means more logic, and increased chances of introducing bugs. This has nothing to do with Jörmungandr.

Inventory.md
KtorZ: Because, to reliably keep track of `SlotId` as we do right now, we would need to maintain a record of protocol parameters to know about the epoch lengths and when they change. The slotting arithmetic becomes tricky because, the "next slot" needs to lookup its corresponding epoch length and makes computation from there. 

If you consider `SlotNo` only, computing the next or previous slot is straightforward and can be done regardless of the epoch length value.

Inventory.md
KtorZ: - we know if a block is the genesis block (i.e. origin) by comparing its hash with the known genesis hash
- SlotId -> SlotNo is a transformation that needs to happen regardless. So, not really a problem.

 Our `BlockHeader` type is currently a superset of this `Origin | At (slotNo headerHash)` data-type.

Inventory.md
Anviking: > It's hard to be more generic than Bytestring though

Quite the opposite though, right? 😬  With `ByteString` we can do whatever we want with it. `addr` requires us to be explicit with what we can do with it.

> trust me, no one wants to go through this

I'm at least curious / open to the possibility :)

Inventory.md
Anviking: >  Because "an empty wallet" with no blocks requires handling the initialization as a special case

Could you elaborate on what this refers to?

In my quick prototype I changed `initWallet` to take a block header instead of a full block, and I didn't see any problems.

https://github.com/input-output-hk/cardano-wallet/blob/ed10bfeabac2e184a3d090d875846091249a3288/lib/core/src/Cardano/Wallet/Primitive/Model.hs#L187-L197

Inventory.md
Anviking: > The slotting arithmetic becomes tricky because

I don't think we are actually using any slot arithmetic in the wallet. Only when testing. (E.g search for `slotSucc`).

My view was that as long as we have the current `ProtocolParameters`, a new `Block` and a ledger function to produce the new protocol parameters, we're fine.

Conversions between `SlotNo` and `UTCTime` will still require protocol params like  `SlotId`-`UTCTime` does. (but that's not the point here I believe)

Inventory.md
KtorZ: What happened to the transactions in the genesis block :) ? 
So now, we have a wallet that has already applied the given block header, but it has missed all the transactions from the genesis block. How do we make sure that we still get to apply the corresponding block, while still making sure that we don't apply the same block twice ?

Anviking: I assume that is why we should initialise our wallets with `Origin :: Point ByronBlock`!

(Not sure I've actually confirmed this, but I assume `Origin` is the point before genesis. Or that specifying it as a known point will make the chain-sync yield the genesis block.

This is not in my prototype, you're right it's probably not syncing the genesis block.)

Running the second prototype with `Origin` as starting point:
```
rollback to: Origin



=== SlotNo 0
=== ByronHash {unByronHash = AbstractHash 89d9b5a5b8ddc8d7e5a6795e9774d97faf1efea59b2caf7eaf9f8c5b32059df4}


*** Exception: how do we represent the genesis hash?
```
i.e must be as I expected.

KtorZ: > the chain-sync yield the genesis block.

I hardly believe that. The genesis block weights hundred of megabytes. I don't see this being transmitted over the network just like this, but I may be wrong.

Anviking: Actually, you might be right 🤔 👀

Anviking: Seems like you're right 👍  https://github.com/input-output-hk/cardano-ledger/blob/0ad59ce46b3da8af6f6af97064e7c77cc7c2ac8b/cardano-ledger/src/Cardano/Chain/UTxO/GenesisUTxO.hs#L19-L36

Although I wonder if "block" is a correct description of it when it is called `Genesis.Config` in the code 🤔 I guess it works still.

I also tried writing the first few blocks from chain-sync to disk, starting at `Origin`, and they  were all ≈670 bytes.

Inventory.md
Anviking: I was in a hurry too when making the final prototype and writing this section — sorry! (or the U/S was supposed to be already done then)

I couldn't figure out how to implement `networkTip` and `nextBlocks` in a good way with the Haskell node. I will try to provide some reasoning in hindsight:

The `NetworkLayer` (and therefore cardano-wallet-jörmungandr) implements
```haskell
    { nextBlocks
        :: Cursor target
        -> ExceptT ErrGetBlock m (NextBlocksResult target block)
```
which the `follow` function calls repeatedly with different `Cursor target`.

The Haskell chain sync protocol is stateful in the sense that the node knows where we are on the chain and will tell us when to roll forward and when to roll back automatically. A  `nextBlocks` without the `Cursor target` argument (local tip) would work better. But I thought the most straight-forward abstraction for the Haskell node would be:

```haskell
follow 
    :: [point] -- points/BlockHeaders we know about
    -> (RollBackOrForward block -> IO ()) 
    -> IO () -- syncs forever
```
which we could put inside `NetworkLayer. `[src](https://github.com/input-output-hk/cardano-wallet/blob/ed10bfeabac2e184a3d090d875846091249a3288/lib/core/src/Cardano/Wallet/Network.hs#L65-L68), and note this is different from our current `follow` in `Cardano.Wallet.Network`.

It was my intention for this `follow` to also work with `cardano-wallet-jormungandr`. But it does not! [`monitorStakePools`](https://github.com/input-output-hk/cardano-wallet/blob/cc7b767608399279fe1f88f5be2ed84268af3d35/lib/core/src/Cardano/Pool/Metrics.hs#L155-L229) relies on being able to tell the current `Cardano.Wallet.Network.follow` to `RetryImmediately`. This is to avoid *jörmungandr-specific* race-conditions when fetching the current stake-distribution. We can't tell the Haskell node to give us a block it has already sent us. Or even if we can with some trickery, it would just be unnecessary. Which is why I didn't  think the current `monitorStakePools` or `Cardano.Wallet.Network.follow` would be good abstractions, nor `nextBlocks`.

NOTE: I believe the chain-sync protocol might be meant to be used with some `TVar` thing. That might invalidate my reasoning *slightly*. I believe @KtorZ should be able tell easily though.

Inventory.md
Anviking: > type magic refactoring

I think that was the `Api.Link` thing, which I think would only reduce boilerplate in the integration tests. Shouldn't help us here.

Matthias pointed that the wallet already is modular enough though. https://github.com/input-output-hk/cardano-wallet/pull/1214#discussion_r364637707

Inventory.md
Anviking: 🙂 fine.

Inventory.md
Anviking: Thanks for catching. I simply removed the entire bullet-point.

Inventory.md
Anviking: I replaced the section with a slightly modified version of this comment, and updated the summary-section.

Inventory.md
Anviking: *cited Duncan's comment directly in the text and added a link to WB-44*

Anviking:
Thanks for having a look @paweljakubas @KtorZ! I added the inventory as an issue-comment in #1161.

Issue #961, opened 2019-11-05
--logging-config FILE.YAML more info and tests required


piotr-iohk:
| Release | Operating System       | Cause |
| ---     | ---                    | ---   |
| next    | Windows & OSX & Linux) | Code  v Configuration v Environment v Human v Unknown |

# Context
 
As part of #851 there was this option provided for `launch` and `serve` commands:

```
  --logging-config FILE.YAML
                           File to configure the iohk-monitoring framework.
```

The format of the file is somewhat unknown and it should be provided by iohk-monitoring team.
We need to have it documented https://github.com/input-output-hk/cardano-wallet/wiki/Wallet-command-line-interface and/or link to appropriate resource added. Also information about default values needs to be provided.

We should also have some tests illustrating the behavior of `--logging-config`.
 

# Steps to Reproduce

## Actual behavior
 
<A clear and concise description of what you observe instead. If applicable add screenshots to help explain your problem.>

---

# Resolution Plan

# PR

| Number   | Base      |
| ---      | ---       |
| #1253        | `master` |

# QA

paweljakubas:
In the end, we removed logging-config in #1253. wiki was also updated accordingly

PR #1217, opened 2019-12-22
Improve logging documentation of configs

paweljakubas:
# Issue Number

#961 

# Overview

- [x] I have replicated (tried to do it) monitoring config needed to retrieve what happens when just `defaultConfigStdout` from iohk-monitoring-framework is called
- [x] I have replicated (tried to do it) monitoring config needed to retrieve what happens when `initTracer` with Nothing from cli is called
- [ ] I have updated CLI Manual

# Comments

In order to test do : 
```
$ stack test --ta '-m "initTracer"' cardano-wallet-cli:unit
```

### DIFF
--- a/lib/cli/test/unit/Cardano/CLISpec.hs
+++ b/lib/cli/test/unit/Cardano/CLISpec.hs
@@ -476,6 +476,12 @@ spec = do
             requireFilePath fp
             fileConfig <- getCG . fst . snd <$> initTracer (Just fp) Debug
             compareCfgs defaultConfig fileConfig
+        it "default setting" $ do
+            defaultConfig <- getCG . fst . snd <$> initTracer Nothing Debug
+            let fp = "../../specifications/logging/default.yaml"
+            requireFilePath fp
+            fileConfig <- getCG . fst . snd <$> initTracer (Just fp) Debug
+            compareCfgs defaultConfig fileConfig
   where
     backspace :: Text
     backspace = T.singleton (toEnum 127)
--- a/lib/cli/test/unit/Cardano/CLISpec.hs
+++ b/lib/cli/test/unit/Cardano/CLISpec.hs
@@ -11,6 +11,12 @@ module Cardano.CLISpec
 
 import Prelude
 
+import Cardano.BM.Configuration.Model
+    ( Configuration (..) )
+import Cardano.BM.Configuration.Static
+    ( defaultConfigStdout )
+import Cardano.BM.Data.Severity
+    ( Severity (..) )
 import Cardano.CLI
     ( MnemonicSize (..)
     , Port (..)
@@ -24,13 +30,15 @@ import Cardano.CLI
     , cmdWallet
     , hGetLine
     , hGetSensitiveLine
+    , initTracer
+    , requireFilePath
     )
 import Cardano.Wallet.Primitive.AddressDerivation
     ( NetworkDiscriminant (..) )
 import Control.Concurrent
     ( forkFinally )
 import Control.Concurrent.MVar
-    ( newEmptyMVar, putMVar, takeMVar )
+    ( newEmptyMVar, putMVar, readMVar, takeMVar )
 import Control.Monad
     ( mapM_ )
 import Data.Proxy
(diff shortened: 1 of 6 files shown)


specifications/logging/default.yaml
```diff
@@ -0,0 +1,45 @@
...
+  - scKind: StdoutSK
+    scFormat: ScJson
+    scName: json
+
+# if not indicated otherwise, then log output is directed to this:
+defaultScribes:
+  - - StdoutSK
+    - text
+
+# this is probably a bug of parsing, cannot enforce cgBindAddrPrometheus = Nothing
```
paweljakubas, line 20: this is something I cannot go through : 
so when I put `hasPrometheus:` or when I omit it at all to specify this option I end up with: 
```
cgBindAddrPrometheus = Just ("127.0.0.1",12799)
```
When I `hasPrometheus: ["",0]` I end up with
```
cgBindAddrPrometheus = Just ("",0)
```
which is wrong (it should be validated). And it is expected to have `cgBindAddrPrometheus = Nothing`
BTW. it is inconsistent with example configuration specified here : https://github.com/input-output-hk/iohk-monitoring-framework/blob/bea0e079fc32ed316ce352d17d14199a680e3f6c/iohk-monitoring/test/config.yaml
When we put `hasPrometheus: 12799` we end up with parsing error

specifications/logging/default.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+# these backends are initialized:
+setupBackends: [AggregationBK, KatipBK]
+
+# if not indicated otherwise, then messages are passed to these backends:
+defaultBackends:
+  - KatipBK
+
+# more options which can be passed as key-value pairs:
+options:
```
paweljakubas, line 32: This is needed in order to have set : 
```
cgMapSubtrace = fromList [("#messagecounters.monitoring",NoTrace),("#messagecounters.ekgview",NoTrace),("#messagecounters.aggregation",NoTrace),("#messagecounters.graylog",NoTrace),("#messagecounters.katip",NoTrace),("#messagecounters.switchboard",NoTrace)]
```
but it also sets : 
```
cgOptions = fromList [("mapSubtrace",fromList [("#messagecounters.monitoring",Object (fromList [("subtrace",String "NoTrace")])),("#messagecounters.ekgview",Object (fromList [("subtrace",String "NoTrace")])),("#messagecounters.aggregation",Object (fromList [("subtrace",String "NoTrace")])),("#messagecounters.graylog",Object (fromList [("subtrace",String "NoTrace")])),("#messagecounters.katip",Object (fromList [("subtrace",String "NoTrace")])),("#messagecounters.switchboard",Object (fromList [("subtrace",String "NoTrace")]))])],
```
which is maybe as it should be. BUT, it is expected in empty configuration that `cgOptions = fromList []` So it is inconsistent in such a sense, That empty is set like : https://github.com/input-output-hk/iohk-monitoring-framework/blob/master/iohk-monitoring/src/Cardano/BM/Configuration/Model.lhs#L543
and then on top of it is https://github.com/input-output-hk/iohk-monitoring-framework/blob/1ebe57df3a76a8ae4e3b5a31a9b85132f71534fc/iohk-monitoring/src/Cardano/BM/Configuration/Static.lhs#L26
So it behaves differently when yaml setting and in the code.

rvl:
It seems like it will be difficult for us to test and document the entire log config format, and that is really a job for the iohk-monitoring team.

The main use for this `--logging-config` option would probably be as a way to enable logging to file with rotation by log file size.

paweljakubas:
@rvl @KtorZ exactly, we probably should resign from giving user possibility of setting whole logging config, but expose some features like rotation, etc. 

Nevertheless, the above findings may be of interest to iohk-monitoring-framework team

lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -461,9 +469,26 @@ spec = do
             , expectedStdout = "Prompt: ******\ESC[1D \ESC[1D\ESC[1D \ESC[1D**\n"
             , expectedResult = "pata14" :: Text
             }
+    describe "initTracer" $ do
+        it "no settings" $ do
+            defaultConfig <- getCG <$> defaultConfigStdout
+            let fp = "../../specifications/logging/empty.yaml"
```
KtorZ, line 475: **[change required]**

Use [`(</>)`](https://hackage.haskell.org/package/filepath-1.4.2.1/docs/System-FilePath-Posix.html#v:-60--47--62-) from `System.FilePath` instead of hard-coded separators. 

**Why**

Hard-coded path don't play well with cross-compiled code :s, so, always rely on platform-agnostic combinator like `(</>)` to combine paths.

specifications/logging/default.yaml
KtorZ: I'd suggest to open an issue on the iohk-monitoring-framework. Fixing this is not part of our scope, but reporting it is :)

KtorZ:
Hmmm. @paweljakubas , the goal for #961 I believe is simply to get _some degree_ of documentation on our side. Pointing to some existing documentation / examples on the iohk-monitoring-framework would probably be sufficient. 
For testing, I concur @rvl's remark, it's not our scope to check whether the framework works as intended. We already have a few test cases regarding logging and it's quite okay for the wallet backend already.

paweljakubas:
closing in favor of #1253

PR #1218, opened 2019-12-23
Fix cabal v2-build

rvl:
- The dependencies in cabal.project were out of sync with stack.yaml.
- I haven't measured whether we even need to build with `-O2`, but it's certainly not needed when `-fdevelopment` is enabled.


KtorZ (approved):

rvl:
Disk space issues on Buildkite, retrying ...

bors r+

Issue #1208, opened 2019-12-19
Support and discriminate 12 and 15 words in the byron createWallet endpoint


KtorZ:
# Context

Byron Daedalus wallets use 12 words, Byron Yoroi wallets use 15.

# Decision

To keep the API simple and user friendly, we'll use the number of words to discriminate between  Yoroi and Daedalus byron wallets and pick a corresponding API layer.

# Acceptance Criteria

1. One _must_ be able to restore a legacy yoroi wallet using a 15 word mnemonic passphrase via the same api endpoint https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/postByronWallet

2. Other endpoint for legacy wallets _must_ be accessible to the wallet yielded
3. Migration of Yoroi legacy wallet _may_ be postponed to another task.

---

# Development 

- Support for 15-word mnemonic in the API ==> #1209 
- Work-around for the initial blockchain scan in the genesis block ==> #1221 

# QA 

- Added a new faucet to integration scenario pointing to Icarus wallets in the genesis integration file. Faucet can be used in an analog way to Byron wallets (i.e. `fixtureIcarusWallet`). See also:
    - https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core-integration/src/Test/Integration/Scenario/API/ByronWallets.hs#L149
    - https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core-integration/src/Test/Integration/Scenario/API/ByronTransactions.hs#L77

- Added unit tests for `IcarusKey` derivations methods: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivation/IcarusSpec.hs#L78-L86

- Resurrected golden tests comparing known Icarus addresses with addresses obtained through implementation: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivation/IcarusSpec.hs#L63-L76

- Made the `SequentialSpec` unit tests more polymorphic so that we could test with both Icarus and Shelley Key, as well as any possible combination of the accounting style: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/SequentialSpec.hs#L134-L146

[...]

PR #1220, opened 2019-12-23
Use big pool gap for Icarus wallet first scan

KtorZ:
# Issue Number

#1208 

# Overview

- commit 4f07b8d7b75550667ee7b1f37f739ac32a1a1636

    add special integration test for Icarus wallets with high indexes
    Currently fails as expected with:

    ```
    1) API Specifications BYRON_RESTORE_08 - Icarus wallet with high indexes
         expected: 1000000000000
          but got: 0
    ```

- commit 5783c86dcfac063689cc5ed289419da6659288ca 
    scan genesis block with bigger pool gap for Icarus wallets
    This to compensate the lack of transaction history in the ITN.

- [ ] TODO: Maybe add some unit test for the `shrinkPool` function :grimacing: 

# Comments

This is what an existing non-empty wallet looks like with this (not mentioning the fact that addresses are discovered correctly!):

```
(master) $ ls -lh ~/.local/share/cardano-wallet/jormungandr/testnet/wallets
total 13M
-rw-r--r-- 1 ktorz ktorz 1,2M déc.  23 11:47 ica.f2a072b15fa9a2eabae5472344cdcc85fa82931d.sqlite
```

Without this little pool shrinks, the wallet size was ~960MB :man_shrugging:


piotr-iohk (approved):
OK, but have couple of questions/suggestions for tests. (However maybe that's not the scope of this particular task).

Test with other "byron" endpoints:
 - GET
 - LIST (in particular that 12-word and 15-word mnemonic wallets are listed "side-by-side" ordered from oldest to newest.)
 - DELETE
 - list/forget transactions - (btw. not sure if I get the part "the lack of transaction history in the ITN")
 - Migrate/Calculate fee (but I suppose it is out of scope for this task for now)

KtorZ:
@piotr-iohk I've actually extended the byron integration scenarios last week to test every time with Byron and Icarus wallets (e.g. https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core-integration/src/Test/Integration/Scenario/API/ByronWallets.hs#L149). I've done it for relevant endpoints (MIGRATE, CALCULATE, RESTORE ... ) but haven't done LIST :thinking: 

I haven't tried to _mix_ them though (apart from manual testing) I don't see any reason it wouldn't work. Still, the list endpoint may have a surprising behavior. Icarus wallets would be sorted, Byron wallets would be sorted, but they'll be in two groups (So, like `[ica1, ica2, ica3, byr1, byr2, byr3]`). Can be fixed rather easily though.

> not sure if I get the part "the lack of transaction history in the ITN

For the incentivized testnet, we start from an initial snapshot of the mainnet. The snapshot contains the UTxO at the moment it is taken. Now, if we do sequential discovery and, have a pool gap of let's say `20`, the algorithm says that we must first generate the first `20` addresses, as we discover addresses, we have to generate further. 

Now, what if, the first address in the snapshot starts at index 21 :man_shrugging: ... That's quite possible for a user that has made a few transactions and has used all of its first `20` addresses. Therefore, this PR does 2 things:

- Illustrate the case where a wallet would have addresses that are beyond the initial default gap and that wouldn't be discovered by a "standard discovery"
- Cheat a bit when scanning the genesis block to use a gap as big as 10.000, with the assumption that it would be sufficient for most users...

KtorZ:
bors try

PR #1227, opened 2019-12-24
Update manual tests and db migration manual scenario

piotr-iohk:
# Issue Number

# Overview

- [ ] I have updated `SyncProgress` manual test to also check for Yoroi wallets, and check `network information` and `stake-pool list` endpoints while jormungandr is down
- [ ] I have added manual test for DB migration (Seeing this -> #1190, but until it is functional manual scenario is something we need it seems)
- [ ] added test `.bat` scripts to aid ad-hoc/release checklist testing on windows

# Comments


KtorZ (approved):
:tada: good idea.

Issue #1228, opened 2019-12-24
Unhandled Data.Binary.Get error observed in node logs


alan-mcnicholas:
# Context
 
When running the latest version of Daedalus (https://buildkite.com/input-output-hk/daedalus/builds/10138) the following error message appears in the node logs.

```
Dec 24 16:05:27.357 INFO fetching block 8e4d2a343f3dcf9330ad9035b3e8d168e6728904262f2c434a4f8f934ec7b676, peer_address: 52.28.91.178:3000, block: 8e4d2a343f3dcf9330ad9035b3e8d168e6728904262f2c434a4f8f934ec7b676, task: init
cardano-wallet-jormungandr.exe: Data.Binary.Get.runGet at position 0: not enough bytes
CallStack (from HasCallStack):
  error, called at libraries/binary/src/Data/Binary/Get.hs:351:5 in binary-0.8.6.0:Data.Binary.Get
```

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | `2019.12.23` |
| Platform     | Windows 10         |
| Installation | https://buildkite.com/input-output-hk/daedalus/builds/10138          |

# Steps to Reproduce

1. Install version of Daedalus from the link above. Note this error does not always appear however it can be forced by closing the Daedalus client.

## Expected behavior

Any errors should be handled correctly.

## Actual behavior

Unhandled error is output in the node logs however this does not appear to have any adverse effect on the application. 

# Resolution 

- Avoid raw usage of `runGet` in places outside of the 'Jormungandr.Binary'  module. In particular:
    - DaedalusIPC on Windows
    - Jormungandr HTTP Client 'MimeUnrender' 

==> #1229 

---

# QA 

[logs-2019-12-24T134635.0659Z.zip](https://github.com/input-output-hk/cardano-wallet/files/3998484/logs-2019-12-24T134635.0659Z.zip)

piotr-iohk:
lgtm

PR #1229, opened 2019-12-24
Define and use safe 'eitherRunGet' to avoid throwing on binary decoding

KtorZ:
# Issue Number

#1228 

# Overview

- ca2675b014b63e0b7b504c1c0d67e885fb56c4a8
  illustrate non-safety in Jörmungandr API client binary decoders
  
- 8e87584965bdf14a165b0cf0625627f6b18ecd28
  define and export a safe version of 'runGet'
  
- 428d0c8a6ff184ce718caac20f75bfe9f6125d53
  use safe 'eitherRunGet' to decode API types
  
- 36d794baa544302dc998f3a2f3c9fc2b18d0a901
  move 'eitherRunGet' into its own module for re-usability
  
- f6bec23ee15233471bba69487eb56f2aa50c5630
  safely decode IPC messages on Windows using eitherRunGet
  
- 26dd771e89ff402d9cb24490d529e83d50de3b59
  re-generate nix machinery

### DIFF
--- a/lib/core/cardano-wallet-core.cabal
+++ b/lib/core/cardano-wallet-core.cabal
@@ -139,6 +139,7 @@ library
       Cardano.Wallet.Unsafe
       Cardano.Wallet.Version
       Cardano.Wallet.Version.TH
+      Data.Binary.Get.Safe
       Data.Function.Utils
       Data.Time.Text
       Data.Time.Utils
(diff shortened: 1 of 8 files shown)


KtorZ:
bors try

rvl (commented):
Looks like this will quash the error call from `runGet`.

The `runNodeChannel` loop relies on an exception to terminate. On posix the `hGet` function will throw an `IOError` if the channel is closed. It appears that on windows, `hGet` returns an empty string. It may not even throw an exception. I will check that.

PR #1231, opened 2019-12-27
Fix missing test data in windows testing bundle

rvl:
Relates to #1228.

# Overview

- Adds `mock-daedalus.js` to the windows testing bundle so that the DaedalusIPC tests can run.
- Fixes a failing DaedalusIPC test case.
- Makes `mock-daedalus.js` use `child_process.disconnect()` rather than exiting and leaving behind stray processes on windows.

# Comments

[Latest build](https://hydra.iohk.io/job/Cardano/cardano-wallet-pr-1231/cardano-wallet-jormungandr-tests-win64)


KtorZ (approved):
Nice catch

Issue #1232, opened 2019-12-27
Allow restoring 24-word mnemonic in the Byron API 


KtorZ:
# Context

We ought to support both Trezor and Ledger mnemonic restoration through the API. They both typically use 24-word mnemonic. 

# Decision

- Allow 24-word mnemonic in the API
- Provide a way to discriminate between Trezor and Ledger via a query parameter

# Acceptance Criteria

1. One _must_ be able to restore a Hardware Ledger 24-word mnemonic via the API
2. One _must_ be able to restore a Hardware Trezor 24-word mnemonic via the API

---

# Development 

- #1236 

# QA 

- Rather exhaustive list of scenarios exercising the legacy restoration endpoints with various sizes and various styles of wallets: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core-integration/src/Test/Integration/Scenario/API/ByronWallets.hs#L872-L898

- Added a special integration tests restoring a Ledger wallet with funds, since they are a bit special https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core-integration/src/Test/Integration/Scenario/API/ByronWallets.hs#L1258-L1277

- Already existing integration test covering both Icarus and Trezor: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core-integration/src/Test/Integration/Scenario/API/ByronWallets.hs#L1235-L1256

- All other already existing integration tests, testing Icarus (sequential) and Random (random) wallets.

piotr-iohk:
I think there are some cases that would be worth adding:
 - migration/calcuate migration fee of funds for trezor/ledger to shelley
 - listing of interleaving wallets trezor/ledger/icarus/random (it's there for icarus and random)
 - deletion of trezor/ledger
 - listing transactions for trezor/ledger/icarus (only for random)
 - forgeting transactions for trezor/ledger/icarus (only for random)

Maybe something else? Thoughts?

KtorZ:
Yes. Though, I don't think we should _add_ tests per se, but instead modify the existing scenarios so that they run for each type.

PR #1234, opened 2019-12-27
Support Hardware Ledger Auxiliary Seed Generation Method

KtorZ:
# Issue Number

#1232 

# Overview

- 7ea19ff729839bdc9203bd610649f2edb9af7f61
  Add auxiliary seed generation method for Hardware Ledger
  This implement the firmware generation that is happening on Hardware Ledger Nano S and
Nano X. Unfortunately, this process isn't really documented anywhere, and the implementation
has only been tested superficially from a single known mnemonic. I'll try out again with
an actual device as soon as I get my hand on one

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Icarus.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Icarus.hs
@@ -26,6 +26,7 @@ module Cardano.Wallet.Primitive.AddressDerivation.Icarus
 
     -- * Generation and derivation
     , generateKeyFromSeed
+    , generateKeyFromHardwareLedger
     , unsafeGenerateKeyFromSeed
     , minSeedLengthBytes
     ) where
@@ -66,20 +67,34 @@ import Cardano.Wallet.Primitive.AddressDerivation
     )
 import Cardano.Wallet.Primitive.AddressDerivation.Byron
     ( decodeLegacyAddress )
+import Cardano.Wallet.Primitive.Mnemonic
+    ( Mnemonic, mnemonicToText )
 import Cardano.Wallet.Primitive.Types
     ( Address (..), Hash (..), invariant )
+import Control.Arrow
+    ( first, left )
 import Control.DeepSeq
     ( NFData (..) )
 import Control.Monad
     ( (<=<) )
+import Crypto.Error
+    ( eitherCryptoError )
 import Crypto.Hash
     ( hash )
+import Crypto.Hash.Algorithms
+    ( SHA256 (..), SHA512 (..) )
+import Crypto.MAC.HMAC
+    ( HMAC, hmac )
 import Data.Bifunctor
     ( bimap )
+import Data.Bits
+    ( clearBit, setBit, testBit )
 import Data.ByteString
     ( ByteString )
 import Data.Coerce
     ( coerce )
+import Data.Function
+    ( (&) )
 import Data.Maybe
     ( fromMaybe )
 import Data.Proxy
@@ -88,11 +103,17 @@ import Data.Word
     ( Word32 )
 import GHC.Generics
     ( Generic )
+import GHC.Stack
+    ( HasCallStack )
 
 import qualified Cardano.Byron.Codec.Cbor as CBOR
 import qualified Codec.CBOR.Write as CBOR
+import qualified Crypto.ECC.Edwards25519 as Ed25519
+import qualified Crypto.KDF.PBKDF2 as PBKDF2
 import qualified Data.ByteArray as BA
 import qualified Data.ByteString as BS
+import qualified Data.Text as T
+import qualified Data.Text.Encoding as T
 
 -- | A cryptographic key for sequential-scheme address derivation, with
 -- phantom-types to disambiguate key types.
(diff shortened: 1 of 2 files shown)


lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Icarus.hs
```diff
@@ -151,6 +172,84 @@ generateKeyFromSeed
...
+    -- - Set the second highest bit of the last byte
+    --
+    -- As described in:
+    --
+    -- "BIP32-Ed25519 Hierarchical Deterministic Keys over a Non-linear Keyspace".
+    normalizeEd25519 :: ByteString -> ByteString
+    normalizeEd25519 bytes =
+        let
+            (firstByte, rest) = BS.splitAt 1 bytes
+            (rest', lastByte) = BS.splitAt 30 rest
```
rvl: We could do this to remove the need for `BS.index`.
```suggestion
            Just (firstByte, rest) = BS.uncons bytes
            Just (rest', lastByte) = BS.unsnoc rest
```
Then put them back together with `BS.cons` and `BS.snoc` to remove the need for `mconcat` and `singleton`.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Icarus.hs
```diff
@@ -151,6 +172,84 @@ generateKeyFromSeed
...
+            $ mnemonicToText mnemonic
+
+    let cc = hmacSha256 (BS.pack [1] <> seed)
+    let (kL, kR) = first normalizeEd25519 $ hashRepeatedly seed
+    pA <- ed25519ScalarMult kL
+
+    prv <- left show $ xprv $ kL <> kR <> pA <> cc
+    pure $ IcarusKey (xPrvChangePass (mempty :: ByteString) pwd prv)
+  where
+    unsafeFromRight :: (HasCallStack, Show e) => Either e a -> a
```
rvl: I think `e` is always `String` here, because `show` has already been applied.
And I don't think that HasCallStack will help if the top-level function doesn't have that constraint.
You could add a comment that any Left values here would be a result of a bug in this function.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Icarus.hs
```diff
@@ -151,6 +172,84 @@ generateKeyFromSeed
...
+                    & (`clearBit` 7)
+                    & BS.singleton
+                ]
+
+    ed25519ScalarMult :: ByteString -> Either String ByteString
+    ed25519ScalarMult bytes = do
+        scalar <- left show $ eitherCryptoError $ Ed25519.scalarDecodeLong bytes
+        pure $ Ed25519.pointEncode $ Ed25519.toPoint scalar
+
+    pbkdf2HmacSha512 :: ByteString -> ByteString
```
rvl, line 270: Add a comment referencing [BIP39 From mnemonic to seed](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#From_mnemonic_to_seed).

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Icarus.hs
```diff
@@ -151,6 +172,84 @@ generateKeyFromSeed
...
+        (kL, kR) -> (kL, kR)
+
+    -- - Clear the lowest 3 bits of the first byte
+    -- - Clear the highest bit of the last byte
+    -- - Set the second highest bit of the last byte
+    --
+    -- As described in:
+    --
+    -- "BIP32-Ed25519 Hierarchical Deterministic Keys over a Non-linear Keyspace".
+    normalizeEd25519 :: ByteString -> ByteString
```
rvl: It's called buffer "pruning" in [RFC 8032 sec 5.1.5. Key Generation](https://tools.ietf.org/html/rfc8032#section-5.1.5), so it might be better to call it "prune" rather than "normalize".

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Icarus.hs
```diff
@@ -151,6 +172,84 @@ generateKeyFromSeed
     -> IcarusKey 'RootK XPrv
 generateKeyFromSeed = unsafeGenerateKeyFromSeed
 
+-- | Hardware Ledger devices generates keys from mnemonic using a different
+-- approach (different from the rest of Cardano).
```
rvl, line 174: The approach for generating the master key and chain code is called [SLIP-0010](https://github.com/satoshilabs/slips/blob/master/slip-0010.md), right?

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Icarus.hs
```diff
@@ -151,6 +172,84 @@ generateKeyFromSeed
...
+
+    prv <- left show $ xprv $ kL <> kR <> pA <> cc
+    pure $ IcarusKey (xPrvChangePass (mempty :: ByteString) pwd prv)
+  where
+    unsafeFromRight :: (HasCallStack, Show e) => Either e a -> a
+    unsafeFromRight = either (error . show) id
+
+    hashRepeatedly :: ByteString -> (ByteString, ByteString)
+    hashRepeatedly bytes = case BS.splitAt 32 (hmacSha512 bytes) of
+        (kL, kR) | testBit (kL `BS.index` 31) 5 -> hashRepeatedly (kL <> kR)
```
rvl: Perhaps we could clarify the guard condition by putting it into a `where` clause and calling it `isInvalidKey`.

From SLIP-0010:

> To avoid invalid master keys, the algorithm is retried with the intermediate hash as new seed if the key is invalid.
> ....
> For ed25519, the last step always succeeds since every 256-bit number (even 0) is a valid private key.

Given that, what is this testBit for? It excludes keys which are multiples of 32 ... are these known to be weak keys?

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivation/IcarusSpec.hs
```diff
@@ -135,6 +179,30 @@ goldenAddressGeneration test = it title $ do
...
+goldenHardwareLedger encPwd sentence addrs =
+    it title $ do
+        let Right mnemonic = mkMnemonic @24 sentence
+        let rootXPrv = generateKeyFromHardwareLedger mnemonic encPwd
+        let acctXPrv = deriveAccountPrivateKey encPwd rootXPrv minBound
+        let deriveAddr = deriveAddressPrivateKey encPwd acctXPrv UTxOExternal
+
+        forM_ (zip [0..] addrs) $ \(ix, addr) -> do
+            let addrXPrv = deriveAddr (toEnum ix)
+            base58 (paymentAddress @'Mainnet $ publicKey addrXPrv) `shouldBe` addr
```
rvl, line 254: Good - Base58 for golden tests to match the source data.

rvl (commented):
I reviewed this by referencing the specs rather than the Python code.

It seems good, except that I'm not sure why `hashRepeatedly` is required. Does the Ledger also do this?

It would help with understanding if the `generateKeyFromHardwareLedger` function had comments and naming that referenced the BIPs, SLIPs, and RFCs that are used.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Icarus.hs
KtorZ: True :+1:

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Icarus.hs
KtorZ: I kept the same terminology as the Rust guys. But I am happy to follow an RFC :)

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Icarus.hs
KtorZ: Damn. I've been my head through these documents for days, going from one to another and trying to figure out which one applied to the Hardware Ledger that I lost track of them at some point but, I think you're right!

KtorZ:
@rvl I'll update a bit the code and comments to reference the various papers / design documents that are involved. 
Also, note that it was recently confirmed that both Ledger and Trezor may use any arbitrary number of words between 12 and 24 so, as hinted in the past already, discriminating the type of wallets based on the number of mnemonic words is not a good idea. So, as discussed with Darko as well, we'll introduce a small "breaking change" in the wallet creation of the legacy API and have a query or path parameter to specify the type of wallet to restore.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Icarus.hs
KtorZ: To be honest. I have no idea and I couldn't find any satisfactory piece of documentation about this. Only thing I know is that the Ledger implementation seems to do this, for... a mysterious reason.

rvl (approved):
Looks good. If it matches the Ledger implementation then good. It would be nice to eventually know the explanation for `isInvalidKey`.

KtorZ:
I'll add some additional golden tests provided by Darko from its device.

PR #1236, opened 2019-12-30
API hardware wallet restoration

KtorZ:
# Issue Number

#1232 

# Overview

- [x] I have refactored in a non backward-compatible way API endpoints for creation of Byron wallets after discussion with Nikola and Darko from Daedalus. This would be clearer and more extensible than what we started.

- [x] I have added support for: 
    - Hardware Ledger wallets restoration
    - Hardware Trezor wallets restoration

# Comments

NOTE: Integration scenarios have become a mess.. we need to spend some time cleaning that up.

Screenshot from 2019-12-30 10-34-20
Screenshot from 2019-12-30 10-34-38
Screenshot from 2019-12-30 10-34-52


KtorZ:
bors try

KtorZ:
@piotr-iohk I'll rework a bit the integration tests for the Byron wallets in another PR to make sure to tests all 4 types correctly.

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -320,15 +321,18 @@ type DelegationFee = "wallets"
 -------------------------------------------------------------------------------}
 
 type ByronWallets =
-         PostByronWallet
+         PostByronWallet "random" '[12]
+    :<|> PostByronWallet "icarus" '[15]
+    :<|> PostByronWallet "trezor" '[12,15,18,21,24]
+    :<|> PostByronWallet "ledger" '[12,15,18,21,24]
```
piotr-iohk, line 327: Before we assumed only 24-word mnemonic as hardware one. Is it the fact that they both support the whole range [15..24]?

specifications/api/swagger.yaml
```diff
@@ -159,24 +159,6 @@ x-walletMnemonicSentence: &walletMnemonicSentence
...
   example: ["squirrel", "material", "silly", "twice", "direct", "slush", "pistol", "razor", "become", "junk", "kingdom", "flee", "squirrel", "silly", "twice"]
 
-x-byronWalletMnemonicSentence: &byronWalletMnemonicSentence
-  description: |
-    A list of mnemonic words.
-
-    Important:
-      - a 12-word mnemonic sentence will be treated as being a Daedalus Byron wallet.
-      - a 15-word mnemonic sentence will be treated as being a Yoroi Byron wallet.
-      - a 24-word mnemonic sentence will be treated as being a Hardware wallet (Trezor or Ledger).
```
piotr-iohk: Adding separate endpoints for each wallet type adds a lot to readability and clearness of what type of wallet is created! :+1: 

(not sure if it breaks some RESTful API rules cause other Byron endpoints are common for all of the wallets, but it makes sense to me)

specifications/api/swagger.yaml
```diff
@@ -1476,18 +1490,60 @@ paths:
...
+          schema: *ApiByronWalletIcarusPostData
+      responses: *responsesPostByronWallet
+
+  /byron-wallets/trezor:
+    post:
+      operationId: postByronTrezorWallet
+      tags: ["Byron Wallets"]
+      summary: Restore (Trezor)
+      description: |
+        <p align="right">status: <strong>stable</strong></p>
```
piotr-iohk, line 1527: is it stable?

specifications/api/swagger.yaml
```diff
@@ -1476,18 +1490,60 @@ paths:
...
+
+  /byron-wallets/ledger:
     post:
-      operationId: postByronWallet
+      operationId: postByronLedgerWallet
       tags: ["Byron Wallets"]
-      summary: Restore
+      summary: Restore (Ledger)
       description: |
         <p align="right">status: <strong>stable</strong></p>
```
piotr-iohk, line 1541: is it stable?

piotr-iohk (approved):
Approved (few minor comments/questions).

lib/core/src/Cardano/Wallet/Api.hs
darko-mijic: Both devices support only 12, 18 and 24 words.

darko-mijic: The full range does not hurt I guess.

specifications/api/swagger.yaml
KtorZ: Added some integration tests to exercise restoration of Random, Icarus, Trezor and Ledger wallets. Remaining endpoints are still common across Byron wallets in general so, I'd say --> yes it is.

specifications/api/swagger.yaml
KtorZ: Same as the above.

specifications/api/swagger.yaml
KtorZ: It does _not_ (from my PoV at least. We can see the wallet's style as a sub-resource of the "byron-wallet" resource. All endpoints that are common to byron-wallets can be reached directly with no sub-typing necessary. If there were endpoints only available to some of the byron wallets, then they could have been referenced under their corresponding sub-type). 

Note that this works okay only because they have a same resource representation in the end `ApiByronWallet`.

PR #1245, opened 2020-01-02
Include other types of byron wallets in tests

piotr-iohk:
# Issue Number

# Overview

- [ ] I have included random, icarus, trezor and ledger byron wallets in existing tests for byron wallets. No new tests have been added.

# Comments
Those changes increased execution time to 1400 s (24 minutes) for whole suite (I think it was ~ 15min before). So maybe some parts can be revised...

### DIFF
--- a/lib/core-integration/src/Test/Integration/Framework/DSL.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/DSL.hs
@@ -80,9 +80,10 @@ module Test.Integration.Framework.DSL
     -- * Helpers
     , (</>)
     , (!!)
-    , emptyRandomWallet
-    , emptyIcarusWallet
+    , byronWalletMnemonics
+    , emptyByronWallet
     , emptyByronWalletWith
+    , emptyRandomWallet
     , emptyWallet
     , emptyWalletWith
     , getFromResponse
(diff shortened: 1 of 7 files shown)


KtorZ:
@piotr-iohk The thing is, there's _no difference_ between Icarus, Ledger and Trezor wallets once they've been created. The only difference occurs at the moment they are created. Afterwards, they are part of the same API context, and manipulated exactly identically (there's not even a single trait on them that could tell us whether they were originally created via the "trezor" endpoint or via the "icarus" endpoint. 

Hence why so far I've limited the testing to: 

- Testing the create endpoint with various scenarios. 
- Testing interaction between Random wallets and Icarus wallets. 

I understand that this require knowing the implementation so it's not entirely blackbox testing here but, I think It's a fair ~assumption~ reasoning to make to avoid almost doubling the duration of testing for a value that is known to be quite low because of the way things are implemented. 

I'd even argue that not everything should be tested, that's actually one of the main selling point of using Haskell. There are things we can enforce at the type-level to get static guarantees about the code. This is one of them. Types don't lie :no_mouth:

piotr-iohk:
@KtorZ, my approach was black box indeed and I wasn't aware that icarus, trezor and ledger are _the same_ under the hood. (I thought that ledger is somewhat different, no?)

> there's no difference between Icarus, Ledger and Trezor wallets once they've been created.

Would it make sense to limit the tests to `["random", "icarus"]` instead of `["random", "icarus", "trezor", "leger"]` then?

> I'd even argue that not everything should be tested,

What particular areas you think could be dropped?

> Types don't lie

```haskell
>>> read @Word8 "-1000"
24
```
:trollface:

piotr-iohk:
I have also noticed this:
 - using the same 12-word mnemonics I can create random, trezor and ledger wallet that coexist on the same wallet backend.
 - using the same 15-word mnemonics I can create icarus and ledger or trezor and ledger. (Cannot create icarus and trezor)
 - using the same mnemonics (12, 15, 18 or 24) I can create trezor and ledger wallet...

I suppose that is expected?
I suppose it indicates that under the hood random, ledger and (icarus, trezor) are  a bit different (with icarus, trezor being quite the same)?

KtorZ:
@piotr-iohk 

> (I thought that ledger is somewhat different, no?)

The manipulation of wallet restored from a Ledger seed is not different from any other sequential wallet. The way the seed is computed from a mnemonic is however different from Byron random wallets and Icarus (resp. Trezor) wallets. 

> What particular areas you think could be dropped?

What imperatively needs testing is the creation of wallets (i.e. `postWallet`) because these are rather different methods for each type (except between Icarus and Trezor which are basically the same). 
Behind the scene, it requires two different API layers (one using sequential derivation for Icarus, Trezor and Ledger, and one using random derivation for Random), and they only "interact" together when listing wallets. For everything else, the server will simply select the right API layer and execute the handler within this context. 

Thus, so long as we test other endpoints using at least a sequential and a random wallet, we're good (though, they really use the exact same code underneath, so it's really a matter of checking that the right API layer gets selected...). 

 > I have also noticed this:

The wallet id is a hash of the root public key. So, wallet that have the same public keys have the same id (because they're intrinsically the same wallets). 

- _using the same 12-word mnemonics I can create random, trezor and ledger wallet that coexist on the same wallet backend._ ==> Yes, they all use different seed generation methods so, from the same mnemonic, you end up with 3 different root keys, and therefore, 3 different wallets.

- _using the same 15-word mnemonics I can create icarus and ledger or trezor and ledger. (Cannot create icarus and trezor)_ ==> Yes, Icarus and Trezor are basically the same thing. They've been separated at the API-level only for a matter of documentation. But they are both sequential wallets using the same mnemonic-to-root-seed generation method. Ledger uses a different mnemonic-to-root-seed method. So, the same mnemonic gives you a different wallet.

- _using the same mnemonics (12, 15, 18 or 24) I can create trezor and ledger wallet..._ ==> Yep, as highlighted above.

piotr-iohk:
I have changed tests to use only "random" and sequential ("icarus") as suggested.
As a result there is 92 tests more in the suite and exec time is ~3-4 minutes longer than before.

current master:

```
Finished in 1124.5275 seconds
1108 examples
```

this branch:

```
Finished in 1328.6268 seconds
1200 examples,
```

If that is not satisfactory I can reduce some tests.
Alternatively we can close this pr as the expectations mentioned here -> https://github.com/input-output-hk/cardano-wallet/pull/1245#issuecomment-571572440 are satisfied with the tests in current master.
Yet, my goal was to extend tests as stated here -> https://github.com/input-output-hk/cardano-wallet/issues/1232#issuecomment-569904539.

piotr-iohk:
Closing for now in favor of Clean up integration tests changes that are comming: https://github.com/input-output-hk/cardano-wallet/pull/1247, https://github.com/input-output-hk/cardano-wallet/milestone/45.

Issue #1146, opened 2019-12-09
Zero balance reported on the wallet despite receiving funds from faucet


piotr-iohk:
# Context
 
Issue encountered by @alan-mcnicholas on his wallet connected to **nightly** testnet (in Daedalus)
Wallet had some funds from the faucet. After some time (after launching Daedalus again some time later) it was showing zero balance, despite there is transaction in the history.

The same wallet restored on different wallet backend shows balance correctly.

# Steps to Reproduce

No steps to reproduce. Occurred once so far.

## Expected behavior

Balance should be correct.

## Actual behavior
 
Apparently the balance was shown correctly at the beginning, but after launching Daedalus again _next morning_ the balance was 0.

Wallet db: [wallets.tar.gz](https://github.com/input-output-hk/cardano-wallet/files/3939528/wallets.tar.gz)

Logs:  [Uploading Logs.tar.gz…]()

Daedalus:
![daedalus_balance](https://user-images.githubusercontent.com/42900201/70434657-1d229680-1a85-11ea-8cf1-46fa59d59452.png)

Transaction details:
```
{
  "inserted_at": {
    "time": "2019-12-07T13:08:42Z",
    "block": {
      "height": {
        "quantity": 2987,
        "unit": "block"
      },
      "epoch_number": 3,
      "slot_number": 7073
    }
  },
  "status": "in_ledger",
  "amount": {
    "quantity": 10000000000,
    "unit": "lovelace"
  },
  "inputs": [
    {
      "id": "3e55108546d9525296f3c4e45e255781c3b27accee84a3e6ceea5ce972132948",
      "index": 255
    }
  ],
  "direction": "incoming",
  "outputs": [
    {
      "amount": {
        "quantity": 10000000000,
        "unit": "lovelace"
      },
      "address": "addr1ssgjwyddpxx6wq9msl4vxtzy8mfgmy3s53mx3w2ztrwjw3tsmtvncqrcgpy3p0tv8jsy5sghlpz787cxyrqerps8p76887d4rt2rsww8lrk3gz"
    }
  ],
  "depth": {
    "quantity": 8245,
    "unit": "block"
  },
  "id": "20425e2e42f01e0e6aed9c00cf6880776e95b5afc9c06224b9c7e7a04ec1a42d"
}
```
---

# Resolution 

- Make sure that the wallet and its network layer are still in sync when the wallet has rolled back ==> #1246 

---

# QA

alan-mcnicholas:
[...]

PR #1246, opened 2020-01-03
Revisions of the logic handling roll backs 

KtorZ:
# Issue Number

#1146 

# Overview

- f8f7f225caa33366a444d5d6687b233a68495324
  exits 'follow' when node switches chain
  This has several advantages:

1. It reduces some logic duplication and actually, manages rollbacks
   very much like a restart. So, there's now only one way of syncing
   with the chain, regardless of whether this is immediately after
   rolling back or, if this is after a long period of inactivity with
   the wallet shut down.

2. it gives clients more flexibility about what they want to do
   in case of rollbacks.

3. It makes the overall logic for 'follow' a bit easier.

- f52946f348d74e36ef096b78d0703c78e59ca28e
  log actual point of rollback returned by db
  
- c64eb3d4d7657aca832a09f820beeb8d3e2d5fdc
  review wording in comment: before ~> earlier

# Comments

The issue on #1146 is relatively hard to reproduce, so here it's mostly a guess based on [this analysis](https://github.com/input-output-hk/cardano-wallet/issues/1146#issuecomment-569610564).

### DIFF
--- a/lib/core/src/Cardano/Pool/Metrics.hs
+++ b/lib/core/src/Cardano/Pool/Metrics.hs
@@ -163,15 +163,19 @@ monitorStakePools
     -> NetworkLayer IO t Block
     -> DBLayer IO
     -> IO ()
-monitorStakePools tr nl DBLayer{..} = do
+monitorStakePools tr nl db@DBLayer{..} = do
     cursor <- initCursor
     logInfo tr $ mconcat
-        [ "Start monitoring stake pools. Currently at "
+        [ "Monitoring stake pools. Currently at "
         , case cursor of
             [] -> "genesis"
             _  -> pretty (last cursor)
         ]
-    follow nl tr cursor forward backward header
+    follow nl tr cursor forward header >>= \case
+        Nothing    -> pure ()
+        Just point -> do
+            liftIO . atomically $ rollbackTo point
+            monitorStakePools tr nl db
   where
     initCursor :: IO [BlockHeader]
     initCursor = do
@@ -185,13 +189,6 @@ monitorStakePools tr nl DBLayer{..} = do
                     Nothing -> putPoolRegistration sl0 r
                     Just{}  -> pure ()
             readPoolProductionCursor k
-
-    backward
-        :: SlotId
-        -> IO (FollowAction ErrMonitorStakePools)
-    backward point = do
-        liftIO . atomically $ rollbackTo point
-        return Continue
 
     forward
         :: NonEmpty Block
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -631,10 +631,8 @@ rollbackBlocks
     -> ExceptT ErrNoSuchWallet IO ()
 rollbackBlocks ctx wid point = db & \DBLayer{..} -> do
     liftIO $ logInfo tr $ "Try rolling back to " <> pretty point
-    mapExceptT atomically $ rollbackTo (PrimaryKey wid) point
-    -- TODO
-    -- point' <- mapExceptT atomically $ rollbackTo (PrimaryKey wid) point
-    -- liftIO $ logInfo tr $ "Rolled back to " <> pretty point'
+    point' <- mapExceptT atomically $ rollbackTo (PrimaryKey wid) point
+    liftIO $ logInfo tr $ "Rolled back to " <> pretty point'
   where
     db = ctx ^. dbLayer @s @k
     tr = ctx ^. logger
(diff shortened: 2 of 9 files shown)


KtorZ:
bors try

lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -163,15 +163,19 @@ monitorStakePools
...
         , case cursor of
             [] -> "genesis"
             _  -> pretty (last cursor)
         ]
-    follow nl tr cursor forward backward header
+    follow nl tr cursor forward header >>= \case
+        Nothing    -> pure ()
+        Just point -> do
+            liftIO . atomically $ rollbackTo point
+            monitorStakePools tr nl db
```
Anviking, line 178: Like this `atomically $ rollbackTo point` and `initPoolProductionCursor` of the next `monitorStakePools` call are in different DB transactions. 

But… actually, that should impossibly be a problem.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -605,8 +605,11 @@ restoreWallet
 restoreWallet ctx wid = db & \DBLayer{..} -> do
     cps <- liftIO $ atomically $ listCheckpoints (PrimaryKey wid)
     let forward bs h = run $ restoreBlocks @ctx @s @k ctx wid bs h
-    let backward sid = run $ mapExceptT atomically $ rollbackTo (PrimaryKey wid) sid
-    void $ liftIO $ follow nw tr cps forward backward (view #header)
+    liftIO (follow nw tr cps forward (view #header)) >>= \case
+        Nothing -> pure ()
+        Just point -> do
+            rollbackBlocks @ctx @s @k ctx wid point
+            restoreWallet @ctx @s @t @k ctx wid
```
Anviking, line 612: 👌

lib/core/src/Cardano/Wallet.hs
```diff
@@ -615,6 +618,25 @@ restoreWallet ctx wid = db & \DBLayer{..} -> do
...
+        , HasDBLayer s k ctx
+        )
+    => ctx
+    -> WalletId
+    -> SlotId
+    -> ExceptT ErrNoSuchWallet IO ()
+rollbackBlocks ctx wid point = db & \DBLayer{..} -> do
+    liftIO $ logInfo tr $ "Try rolling back to " <> pretty point
+    point' <- mapExceptT atomically $ rollbackTo (PrimaryKey wid) point
+    liftIO $ logInfo tr $ "Rolled back to " <> pretty point'
```
Anviking, line 635: Nice logs!

lib/core/src/Cardano/Wallet/DB.hs
```diff
@@ -225,8 +225,13 @@ data DBLayer m s k = forall stm. (MonadIO stm, MonadFail stm) => DBLayer
     , rollbackTo
         :: PrimaryKey WalletId
         -> SlotId
-        -> ExceptT ErrNoSuchWallet stm ()
+        -> ExceptT ErrNoSuchWallet stm SlotId
         -- ^ Drops all checkpoints and transaction data after the given slot.
+        --
+        -- Returns the actual slot to which the database has rolled back. This
+        -- slot is guaranteed to be smaller or equal to the given point of
```
Anviking: > slot is guaranteed to be smaller or equal

Nit-pick: I'd find "earlier [slot]" or "before [slot]" easier to parse than "smaller [slot]".

lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -335,17 +342,16 @@ follow nl tr cps yield rollback header =
             liftIO $ logInfo tr $ mconcat
                 [ "Applying blocks [", pretty slFst, " ... ", pretty slLst, "]" ]
 
-            yield blocks nodeTip >>= handle cursor' "Failed to roll forward: "
+            yield blocks nodeTip >>= handle cursor'
 
-        Right (RollBackward cursor') -> do
-            let point = cursorSlotId nl cursor'
-            logInfo tr $ "Rolling back to " <> pretty point
-            rollback point >>= handle cursor' "Failed to roll backward: "
```
Anviking, line 343: Alternatively to this PR, I guess we could have made `DB.rollback` and the backward-callback fail when a matching checkpoint don't exist.

Anviking (approved):
I suspect "exit on rollback" isn't the ultimate elegant solution (particulary not when we get to the Haskell node), but seems neat for now 👍 

In hindsight the bug here seems obvious. I wonder what tests we would have needed to catch it.

lgtm!

lib/core/src/Cardano/Wallet.hs
KtorZ: Yeah, these tree lines make it somewhat obvious what (one of) the problem was. Took me long enough though...

PR #1247, opened 2020-01-06
add type-safe module to construct API links

KtorZ:
# Issue Number

#1232

# Overview

- 1f78ab4d0b003712dbbb880e854a8684a5e0f3e1
  add type-safe module to construct API links
  This is mostsly relying on servant machinery ('safeLink') but, it
defines easy-to-use constructors that can play well in integration
tests. The reasons behind this change are plural:

1. In order to correctly re-design the way fixtures are created for the
   various types of byron wallets (to tidy up the existing scenarios and
   make them easier to use), I needed a more holistic way to deal with
   wallet endpoints (shelley or byron) which was generic enough to avoid
   too many duplication in the integration DSL. One of the main
   empediment for this was the way we defined endpoints.

2. The current module interface we use for dealing with endpoints is
   very much stringly typed and, although it is mostly sufficient at the
   call-site, it can be quite error-prone and doesn't prevent bad
   surprises when passing along wrong data.

3. We have the 'Cardano.Wallet.Api.Client' when a strongly-typed client
   is needed, but, if we don't want a strongly typed client (like in the
   integration tests), we have to fallback all the way down a fully
   untyped client. This module now offers a middle-ground where we can
   still statically check at least the URL and methods being used.

4. The way the current module interface works in the DSL duplicates all
   the API path. This new module now fully uses 'Cardano.Wallet.Api' as
   a source of truth, removing a lot of duplications in the paths.

Now, TODO: updates the integration tests to use this approach and clean
up the DSL module from the existing 'xxxEp' helpers.

- 1367da0b07da81af9294a58ed400c5d7653055d6
  replace DSL endpoints with new '.Link' module
  TODO:

- [x] Framework/DSL.hs
- [x] Scenario/API/Addresses.hs
- [x] Scenario/API/ByronTransactions.hs
- [x] Scenario/API/ByronWallets.hs
- [x] Scenario/API/Network.hs
- [x] Scenario/API/Transactions.hs
- [x] Scenario/API/Wallets.hs
- [x] Jormungandr/Scenario/API/StakePools.hs
- [x] Jormungandr/Scenario/API/Transactions.hs
- [x] bench/Latency

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api.hs
+++ b/lib/core/src/Cardano/Wallet/Api.hs
@@ -17,19 +17,52 @@ module Cardano.Wallet.Api
 
       -- * Shelley
     , Wallets
+        , DeleteWallet
+        , GetWallet
+        , ListWallets
+        , PostWallet
+        , PutWallet
+        , PutWalletPassphrase
+        , GetUTxOsStatistics
+
     , Addresses
+        , ListAddresses
+
     , CoinSelections
+        , SelectCoins
+
     , Transactions
+        , CreateTransaction
+        , PostTransactionFee
+        , ListTransactions
+        , DeleteTransaction
+
     , StakePools
+        , ListStakePools
+        , JoinStakePool
+        , QuitStakePool
+        , DelegationFee
 
     -- * Byron
     , ByronWallets
+        , DeleteByronWallet
+        , GetByronWallet
+        , ListByronWallets
+        , PostByronWallet
+
     , ByronTransactions
+        , ListByronTransactions
+        , DeleteByronTransaction
+
     , ByronMigrations
+        , MigrateByronWallet
+        , GetByronWalletMigrationInfo
 
     -- * Miscellaneous
     , Network
+
     , Proxy_
+        , PostExternalTransaction
 
       -- * Api Layer
     , ApiLayer (..)
@@ -49,7 +82,8 @@ import Cardano.BM.Trace
 import Cardano.Wallet
     ( WalletLayer (..) )
 import Cardano.Wallet.Api.Types
-    ( ApiAddress
+    ( AllowedMnemonics
+    , ApiAddress
     , ApiByronWallet
     , ApiByronWalletMigrationInfo
     , ApiCoinSelection
@@ -64,10 +98,12 @@ import Cardano.Wallet.Api.Types
     , ApiWallet
     , ApiWalletPassphrase
     , ByronWalletPostData
+    , ByronWalletStyle (..)
     , Iso8601Time
     , PostExternalTransactionData
     , PostTransactionData
     , PostTransactionFeeData
+    , StyleSymbol
     , WalletPostData
     , WalletPutData
     , WalletPutPassphraseData
@@ -103,8 +139,6 @@ import Data.Text
     ( Text )
 import GHC.Generics
     ( Generic )
-import GHC.TypeLits
-    ( Symbol )
 import Network.HTTP.Media
     ( (//), (/:) )
 import Servant.API
@@ -321,18 +355,18 @@ type DelegationFee = "wallets"
 -------------------------------------------------------------------------------}
 
 type ByronWallets =
-         PostByronWallet "random" '[12]
-    :<|> PostByronWallet "icarus" '[15]
-    :<|> PostByronWallet "trezor" '[12,15,18,21,24]
-    :<|> PostByronWallet "ledger" '[12,15,18,21,24]
+         PostByronWallet 'Random
+    :<|> PostByronWallet 'Icarus
+    :<|> PostByronWallet 'Trezor
+    :<|> PostByronWallet 'Ledger
     :<|> DeleteByronWallet
     :<|> GetByronWallet
     :<|> ListByronWallets
 
 -- | https://input-output-hk.github.io/cardano-wallet/api/#operation/postByronWallet
-type PostByronWallet (style :: Symbol) mw = "byron-wallets"
-    :> style
-    :> ReqBody '[JSON] (ByronWalletPostData mw)
+type PostByronWallet (style :: ByronWalletStyle) = "byron-wallets"
+    :> StyleSymbol style
+    :> ReqBody '[JSON] (ByronWalletPostData (AllowedMnemonics style))
     :> PostCreated '[JSON] ApiByronWallet
 
 -- | https://input-output-hk.github.io/cardano-wallet/api/#operation/deleteByronWallet
--- a/lib/core/src/Cardano/Wallet/Api/Link.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Link.hs
@@ -1,0 +1,502 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE InstanceSigs #-}
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-- |
+-- Copyright: © 2018-2020 IOHK
+-- License: Apache-2.0
+--
+-- Type-safe endpoint accessors for the wallet API. Under normal circumstances,
+-- one would prefer to use 'WalletClient' from 'Cardano.Wallet.Api.Client' and
+-- not to bother with endpoints at all.
+--
+-- Yet, in some cases (like in black-box testing), one could want to purposely
+-- send malformed requests to specific endpoints. Thus, this module facilitates
+-- the construction of valid endpoints that'd be accepted by the server, and for
+-- which, users are free to send all sort of data as payload, valid or invalid.
+--
+-- This module is meant to be used via qualified imports and with
+-- type-applications since all exposed functions are type ambiguous in a
+-- variable @style@ of type 'WalletStyle'.
+--
+-- @import qualified Cardano.Wallet.Api.Link as Link@
+--
+-- For examples:
+--
+-- >>> Link.deleteWallet @'Shelley myWallet
+-- ( "DELETE", "/v2/wallets/2512a00e9653fe49a44a5886202e24d77eeb998f" )
+--
+-- >>> Link.getWallet @('Byron 'Icarus) myWallet
+-- ( "GET", "/v2/byron-wallets/2512a00e9653fe49a44a5886202e24d77eeb998f" )
+
+module Cardano.Wallet.Api.Link
+    ( -- * Wallets
+      deleteWallet
+    , getWallet
+    , listWallets
+    , postWallet
+    , putWallet
+    , putWalletPassphrase
+    , getUTxOsStatistics
+    , getMigrationInfo
+    , migrateWallet
+
+      -- * Addresses
+    , listAddresses
+    , listAddresses'
+
+      -- * CoinSelections
+    , selectCoins
+
+      -- * Transactions
+    , createTransaction
+    , listTransactions
+    , listTransactions'
+    , getTransactionFee
+    , deleteTransaction
+
+      -- * StakePools
+    , listStakePools
+    , joinStakePool
+    , quitStakePool
+    , getDelegationFee
+
+      -- * Network
+    , getNetworkInfo
+
+      -- * Proxy
+    , postExternalTransaction
+    ) where
+
+import Prelude
+
+import Cardano.Wallet.Api
+    ( Api )
+import Cardano.Wallet.Api.Types
+    ( ApiT (..)
+    , ApiTxId (ApiTxId)
+    , ByronWalletStyle (..)
+    , Iso8601Time
+    , WalletStyle (..)
+    )
+import Cardano.Wallet.Primitive.AddressDerivation
+    ( NetworkDiscriminant (..) )
+import Cardano.Wallet.Primitive.Types
+    ( AddressState, Hash, PoolId, SortOrder, WalletId )
+import Data.Function
+    ( (&) )
+import Data.Generics.Internal.VL.Lens
+    ( (^.) )
+import Data.Generics.Product.Typed
+    ( HasType, typed )
+import Data.Proxy
+    ( Proxy (..) )
+import Data.Text
+    ( Text )
+import GHC.TypeLits
+    ( Symbol )
+import Network.HTTP.Types.Method
+    ( Method )
+import Servant.API
+    ( (:>)
+    , Capture
+    , Header'
+    , IsElem
+    , QueryParam
+    , ReflectMethod (..)
+    , ReqBody
+    , Verb
+    )
+import Servant.Links
+    ( HasLink (..), safeLink' )
+import Web.HttpApiData
+    ( ToHttpApiData (..) )
+
+import qualified Cardano.Wallet.Api as Api
+
+--
+-- Wallets
+--
+
+-- NOTE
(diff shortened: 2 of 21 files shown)


lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -262,7 +292,7 @@ data ApiUtxoStatistics = ApiUtxoStatistics
 
 data WalletPostData = WalletPostData
     { addressPoolGap :: !(Maybe (ApiT AddressPoolGap))
-    , mnemonicSentence :: !(ApiMnemonicT '[15,18,21,24] "seed")
+    , mnemonicSentence :: !(ApiMnemonicT (AllowedMnemonics 'Shelley) "seed")
```
jonathanknowles: Hi @KtorZ.

The combination of the generic name `WalletPostData` and the rather specific reference to `Shelley` made me pause to think a bit here.

If this type is specific to Shelley (and shouldn't be generalized), then perhaps we could rename it to `ShelleyWalletPostData` (similarly to `ByronWalletPostData`)?

Or if it's a type that's meant to be generalized, then perhaps the specific style of wallet could be expressed as a type variable on `WalletPostData`?

So something like the following (pseudocode):
```hs
data WalletPostData (style :: WalletStyle) = WalletPostData
    { addressPoolGap :: !(Maybe (ApiT AddressPoolGap))
    , mnemonicSentence :: !(ApiMnemonicT (AllowedMnemonics style) "seed")
    ...
```

Thoughts?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -195,6 +200,31 @@ import qualified Data.ByteString.Lazy as BL
...
+    WalletStyleSymbol 'Icarus  = "icarus"
+    WalletStyleSymbol 'Trezor  = "trezor"
+    WalletStyleSymbol 'Ledger  = "ledger"
+
+type family AllowedMnemonics (style :: WalletStyle) :: [Nat] where
+    AllowedMnemonics 'Shelley = '[15,18,21,24]
+    AllowedMnemonics 'Random  = '[12]
+    AllowedMnemonics 'Icarus  = '[15]
+    AllowedMnemonics 'Trezor  = '[12,15,18,21,24]
+    AllowedMnemonics 'Ledger  = '[12,15,18,21,24]
```
jonathanknowles, line 228: I also find that vertical alignment helps here. :+1: 

Though I might be tempted to write it as (to make it even more obvious at a glance which wallet styles include which lengths):

```hs
type family AllowedMnemonics (style :: WalletStyle) :: [Nat] where
    AllowedMnemonics 'Shelley = '[   15,18,21,24]
    AllowedMnemonics 'Random  = '[12            ]
    AllowedMnemonics 'Icarus  = '[   15         ]
    AllowedMnemonics 'Trezor  = '[12,15,18,21,24]
    AllowedMnemonics 'Ledger  = '[12,15,18,21,24]
```

I've noticed that people have started using vertical alignment style more and more recently. If we think that some vertical alignment can help in general (when used judiciously), then **perhaps we should adjust our coding standard document?**

I would add the proviso that we should still move things onto a new line if the lines become too long as a result of vertical alignment.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -195,6 +200,31 @@ import qualified Data.ByteString.Lazy as BL
...
+    | Icarus
+    | Trezor
+    | Ledger
+
+type family WalletStyleSymbol (style :: WalletStyle) :: Symbol where
+    WalletStyleSymbol 'Shelley = "shelley"
+    WalletStyleSymbol 'Random  = "random"
+    WalletStyleSymbol 'Icarus  = "icarus"
+    WalletStyleSymbol 'Trezor  = "trezor"
+    WalletStyleSymbol 'Ledger  = "ledger"
```
jonathanknowles: I wonder, is there a way to get these conversions "for free", without having to define them manually?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -195,6 +200,31 @@ import qualified Data.ByteString.Lazy as BL
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
 
+{-------------------------------------------------------------------------------
+                               Style of Wallets
```
jonathanknowles:
```suggestion
                               Styles of Wallets
```

lib/core/src/Cardano/Wallet/Api/Link.hs
```diff
@@ -0,0 +1,520 @@
...
+--
+-- - A 'Text' string (the URL)
+-- - An HTTP 'Method' for calling this endpoint (e.g. GET, POST, ...)
+--
+-- This function does not type-check if the given endpoint (via type
+-- application) is not actually part of the wallet API.
+--
+-- Note that the 'MkLink endpoint Text' depends on the type of 'endpoint'.
+--
+-- - For simple endpoints with no path nor query parameters we have:
```
jonathanknowles:
```suggestion
-- - For simple endpoints with no path or query parameters we have:
```

lib/core/src/Cardano/Wallet/Api/Link.hs
```diff
@@ -0,0 +1,520 @@
...
+--
+--   @MkLink endpoint Text ~ Text@ and therefore, we can simply do
+--
+--   >>> endpoint @Api.ListWallets id
+--   ( "GET", "/v2/wallets" )
```
jonathanknowles: Nice!

lib/core/src/Cardano/Wallet/Api/Link.hs
```diff
@@ -0,0 +1,520 @@
...
+    -> a
+    -> a
+discriminate shelley byron =
+    if isShelley
+        then shelley
+        else byron
+  where
+    isShelley = isJust $ sameSymbol
+        (Proxy @(WalletStyleSymbol 'Shelley))
+        (Proxy @(WalletStyleSymbol style))
```
jonathanknowles: It seems this function would continue to compile if we added another `WalletStyle`.

But if that new wallet style is neither Shelly nor Byron, then we'd always end up returning the `byron` value.

Can this function be rewritten so that it pattern matches on all of the cases, and fails with a compiler error if a new wallet style is added in the future?

lib/core/src/Cardano/Wallet/Api/Link.hs
```diff
@@ -0,0 +1,520 @@
...
+--
+-- Proxy
+--
+
+postExternalTransaction
+    :: (Method, Text)
+postExternalTransaction =
+    endpoint @Api.PostExternalTransaction id
+
+
```
jonathanknowles: We should probably delete the second empty line here. ^^

(https://github.com/input-output-hk/cardano-wallet/wiki/Coding-Standards#use-only-a-single-blank-line-between-top-level-definitions)

lib/core/src/Cardano/Wallet/Api/Link.hs
```diff
@@ -0,0 +1,520 @@
...
+    endpoint @Api.GetUTxOsStatistics (wid &)
+  where
+    wid = w ^. typed @(ApiT WalletId)
+
+listWallet
+    :: forall (style :: WalletStyle).
+        ( KnownSymbol (WalletStyleSymbol style)
+        )
+    => (Method, Text)
+listWallet = discriminate @style
```
jonathanknowles: I would suggest renaming this wrapper function to `listWallets` (for consistency).

lib/core/src/Cardano/Wallet/Api/Link.hs
```diff
@@ -0,0 +1,520 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE NumericUnderscores #-}
```
jonathanknowles: I think this extension is not used (at least not in this module).

lib/core/src/Cardano/Wallet/Api/Link.hs
```diff
@@ -0,0 +1,520 @@
...
+{-# LANGUAGE OverloadedLabels #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE UndecidableSuperClasses #-}
```
jonathanknowles: This extension also seems to be unused? (I tried compiling without it, and seemed to compile, but perhaps I'm missing something?)

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -17,19 +17,52 @@ module Cardano.Wallet.Api
...
+
     , ByronMigrations
+        , MigrateByronWallet
+        , GetByronWalletMigrationInfo
 
     -- * Miscellaneous
     , Network
+
     , Proxy_
+        , PostExternalTransaction
```
jonathanknowles, line 65: I can see what you're attempting to do here (with nested indentation).

Though when documentation gets rendered with Haddock, the exported entries will still appear as a linear list, without any nesting.

Perhaps, in that case, it would be nicer to use Haddock to indicate nesting?

jonathanknowles (commented):
Hi @KtorZ 

I think this is a nice change, overall.

I do have some reservations about the API wrapper functions in `lib/core/src/Cardano/Wallet/Api/Link.hs`, as they do seem to introduce some boilerplate. I wonder if there's a way to reduce or eliminate that.

However, I think that this PR is a definite improvement.

I've left a few comments on the PR, would be interested to hear your feedback!

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: The type is specific to Shelley indeed. We haven't specialized the name because pretty much **everything** that is Shelley-specific at the moment isn't specialized. That's an _unspoken convention_ of our codebase as far as I can tell... Types are Shelley-specific unless said differently explicitly. 

Renaming to 'ShelleyWalletPostData' makes sense, although it would induce a lot of other renaming elsewhere about many other types. 

Side-note: having 'Shelley' as part of the same 'WalletStyle' ADT feels kinda wrong (what would it mean to have a `ByronWalletPostData 'Shelley` ? I am still grooming this and I intended to open this PR as a draft, sorry.

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: There are ways. Though we are at the type-level here, so things get a little more complicated. I found that simple type-family to be a nice middleground :)

lib/core/src/Cardano/Wallet/Api/Link.hs
KtorZ: > and fails with a compiler error

You like challenging me, don't you. I'll see what's possible to do here.

lib/core/src/Cardano/Wallet/Api/Link.hs
KtorZ: Whoops ^.^ ... not the first draft, I went through several ways, so there are leftovers.

lib/core/src/Cardano/Wallet/Api.hs
KtorZ: Could be, I haven't spent much time on the Haddock output _yet_. I'll see what we can do here too :thinking:

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: > perhaps we should adjust our coding standard document?

I think we have found a few cases where vertical alignment can make sense, especially when it's not exaggerated. For things like, function signatures, record declarations and most case pattern-match newlines are still the de-facto approach IMO. 

I am not sure whether we should adjust the coding standard however. These are coding standards and, as developers (and reviewer), we should know and tolerate when it is reasonable to deviate. Perhaps we could give some extra examples of good usage of vertical alignment (explaining also _why_ the alignment makes sense in that case) and, be more clear about the intent of the coding standards: they are guidelines, not hard constraints.

lib/core/src/Cardano/Wallet/Api/Link.hs
```diff
@@ -0,0 +1,520 @@
...
+-- Some constraints have been explictly added to make functions to NOT
+-- type-check under certain circumstances. The constraints are so-to-speak
+-- redundant but they help enforcing some guarantees.
+{-# OPTIONS_GHC -fno-warn-redundant-constraints #-}
+
+-- |
+-- Copyright: © 2018-2020 IOHK
+-- License: Apache-2.0
+--
+-- Type-safe endpoint accessors for the wallet API. Under normal circumpstances,
```
piotr-iohk: _circumpstances_ -> _circumstances_

paweljakubas:
```suggestion
-- Type-safe endpoint accessors for the wallet API. Under normal circumstances,
```

lib/core/src/Cardano/Wallet/Api/Link.hs
```diff
@@ -0,0 +1,520 @@
...
+-- one would prefer to use 'WalletClient' from 'Cardano.Wallet.Api.Client' and
+-- not bother with endpoints at all.
```
paweljakubas:
```suggestion
-- not to bother with endpoints at all.
```

lib/core/src/Cardano/Wallet/Api/Link.hs
```diff
@@ -0,0 +1,520 @@
...
+        (Proxy @(WalletStyleSymbol 'Shelley))
+        (Proxy @(WalletStyleSymbol style))
+
+-- | Some endpoints are parameterized via a network discriminant in order to
+-- correctly encode their end type (for example, 'CreateTransaction n'). Yet, in
+-- the context of this module, the network discrimination doesn't matter for it
+-- has no influence on the endpoint's value and/or path parameters.
+--
+-- To ease type signatures, we therefore arbitrarily fix the network to Mainnet.
+type Net = 'Mainnet
```
paweljakubas, line 482: why not also `Testnet?

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: I am also here for explicite saying what type level string looks like. Introducing type level function doing the same probably would only decrease readibility of this part of code.

lib/core/src/Cardano/Wallet/Api/Link.hs
KtorZ: Did you read the comment just above :no_mouth:  ?

paweljakubas (approved):
very nice next level refactoring enabling us to better control multidimensional matrices of values (embracing legal ones, and type outlawing the illegal ones) of mnemonics, strings, endpoints.... I need to get accustomed to this - check those outlawed cases and see myself if they are not going to be compiled ;-)

lib/core/src/Cardano/Wallet/Api/Types.hs
jonathanknowles: I do think some extra clarity in the coding standard would be helpful.

The main reason is that the standard specifically marks this kind of alignment as "BAD".

I couldn't find an example that's exactly the same as the code in your PR, but this example is typical of that given in the standard (with `::` instead of `=`):

```hs
-- BAD
data MyRecord = MyRecord 
    { _myRecordLongNameField :: !String
    , _myRecordShort         :: ![Int]
    }
```

Now I don't think the above is necessarily bad (hence my original comment), but since we are seeing a pattern (whereby people are using this kind of alignment more regularly), wouldn't it be better to be explicit about what we do and don't want, and then update the standard accordingly?

KtorZ:
bors try

Issue #1242, opened 2020-01-02
Error in the log after deleting a wallet


piotr-iohk:
# Context
 
 

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | 2019.12.23 (git revision: 7e7c9111c98330c560de7a2a99fd93ef19ebe228) |
| Platform     | Linux (probably all)         |
| Installation | From source            |

# Steps to Reproduce

**Note**: it is observed for all types of wallets: shelley and all byron (random, yoroi, trezor, ledger)

1. Start wallet
```
cd lib/jormungandr/test/data/jormungandr/test_scripts/
cardano-wallet-jormungandr launch --genesis-block ../block0.bin --node-port 8080 --state-dir ./data_launch -- --secret ../secret.yaml --config ../config.yaml
```
2. create a wallet `cardano-wallet-jormungandr wallet create`
3. delete a wallet `cardano-wallet-jormungandr wallet delete <wallet id>`

## Expected behavior

Wallet is deleted. There is no error in the log.

## Actual behavior
 
Wallet is deleted. In the log there is an error:
```
[cardano-wallet.worker:Error:237] [2020-01-02 10:34:35.47 UTC] 14598538: Worker has exited unexpectedly: SQLite3 returned ErrorMisuse while attempting to perform close: bad parameter or other API misuse

```

---

# Resolution 

---

# QA

jonathanknowles:
I managed to confirm this error with running with commit `c7d9bc481fd9cebe8dfe25e3f7d9842158ec3f06`.

jonathanknowles:
Okay. I think I've managed to locate the root cause of this issue. It appears that in the course of the API call, `destroyDBLayer` is called **twice**:

```hs
[cardano-wallet.api:Info:255] [2020-01-07 05:02:04.59 UTC] [RequestId 1] [DELETE] /v2/wallets/67f6430300c403a0a0f2920c2a7c4cd565118a9d
[cardano-wallet.database:Debug:255] [2020-01-07 05:02:04.59 UTC] Closing database (/home/jsk/.local/share/cardano-wallet/jormungandr/testnet/wallets/seq.67f6430300c403a0a0f2920c2a7c4cd565118a9d.sqlite)
[...]

PR #1248, opened 2020-01-07
Fix database error that occurs during wallet deletion.

jonathanknowles:
# Issue Number

#1242 

# Overview

This PR adjusts the implementation of `API.Server.deleteWallet` so that `Cardano.DB.destroyDBLayer` is only called once during its execution.

This can be verified by inspecting the logs before and after applying this PR.

**Before:**
```hs
[cardano-wallet.api:Info:255] [2020-01-07 05:02:04.59 UTC] [RequestId 1] [DELETE] /v2/wallets/67f6430300c403a0a0f2920c2a7c4cd565118a9d
[cardano-wallet.database:Debug:255] [2020-01-07 05:02:04.59 UTC] Closing database (/home/jsk/.local/share/cardano-wallet/jormungandr/testnet/wallets/seq.67f6430300c403a0a0f2920c2a7c4cd565118a9d.sqlite)
[cardano-wallet.database:Debug:185] [2020-01-07 05:02:04.63 UTC] Closing database (/home/jsk/.local/share/cardano-wallet/jormungandr/testnet/wallets/seq.67f6430300c403a0a0f2920c2a7c4cd565118a9d.sqlite)
[cardano-wallet.worker:Error:185] [2020-01-07 05:02:04.63 UTC] 67f64303: Worker has exited unexpectedly: SQLite3 returned ErrorMisuse while attempting to perform close: bad parameter or other API misuse
[cardano-wallet.api:Info:255] [2020-01-07 05:02:04.63 UTC] [RequestId 1] 204 No Content in 0.043179776s
[cardano-wallet.api:Debug:255] [2020-01-07 05:02:04.63 UTC] [RequestId 1] Completed response to API request
```
**After:**
```hs
[cardano-wallet.api:Info:212] [2020-01-07 05:15:22.31 UTC] [RequestId 1] [DELETE] /v2/wallets/4152d133208bb9f2f3ae8ef76a53086c59709c2b
[cardano-wallet.database:Debug:127] [2020-01-07 05:15:22.31 UTC] Closing database (/home/jsk/.local/share/cardano-wallet/jormungandr/testnet/wallets/seq.4152d133208bb9f2f3ae8ef76a53086c59709c2b.sqlite)
[cardano-wallet.api:Info:212] [2020-01-07 05:15:22.31 UTC] [RequestId 1] 204 No Content in 0.001427646s
[cardano-wallet.api:Debug:212] [2020-01-07 05:15:22.31 UTC] [RequestId 1] Completed response to API request
```

# Comments

The second call to `Cardano.DB.destroyDBLayer` happens implicitly, as a result of calling `liftIO $ Registry.remove re wid`.

### DIFF
--- a/lib/core/src/Cardano/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/DB/Sqlite.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedLabels #-}
 {-# LANGUAGE RankNTypes #-}
@@ -51,7 +52,7 @@ import Control.Exception
 import Control.Monad
     ( mapM_ )
 import Control.Monad.Catch
-    ( Handler (..), MonadCatch (..), handleJust )
+    ( Handler (..), MonadCatch (..), handleIf, handleJust )
 import Control.Monad.Logger
     ( LogLevel (..) )
 import Control.Retry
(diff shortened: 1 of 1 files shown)


KtorZ:
@jonathanknowles there's an inherent race-condition here unfortunately. While one thread is deleting the database file from disk, another thread might be trying to read or write to this very database. This is why `removeDatabase` made sure of destroying the database layer first, preventing any other thread from reading / writing to it. 

See my comment in #1242

jonathanknowles:
> @jonathanknowles there's an inherent race-condition here unfortunately. While one thread is deleting the database file from disk, another thread might be trying to read or write to this very database. This is why `removeDatabase` made sure of destroying the database layer first, preventing any other thread from reading / writing to it.
> 
> See my comment in #1242

I did actually test for this condition (on Linux), by actively deleting the DB files and then continuing to issue operations that affect the DB. These operations continue to succeed without any errors, presumably because Sqlite creates RW file descriptors for the files at the point when the connection is initiated, only releasing those descriptors once the "close" command has been issued. 

So even though we issue a "delete" operation, this simply [`unlink`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/unlink.html)s the files from the `.local/share/cardano-wallet/jormungandr/testnet/wallets/` directory, but the files still remain accessible to the DB while it holds on to its file descriptors.

However, I think it's not a good idea to rely on the behaviour described above.

For Windows, I suspect this behaviour is completely different: from memory, I recall that Windows will prevent a process from deleting a file if that file is open (for reading or writing) by another process.

Therefore, I agree with your assessment that we need to do something else: I've updated this PR to make the `destroyDBLayer` idempotent.

KtorZ (approved):
LGTM

PR #1254, opened 2020-01-08
Buildkite: Only run weeder if the stack build succeeds

rvl:
If the stack build fails somewhere and exits early, there will be uncompiled modules, and weeder will return incorrect results.


KtorZ (approved):
Good :tada:

PR #1255, opened 2020-01-09
Fix LOGGING test on Windows

piotr-iohk:
# Issue Number

#1115 

# Overview

- 71c8f3028cb3012d765327d3ecf9f83826a4df50
  Fix LOGGING test on Windows
  
- c3e9828c89b28bbc106951240a1e021b2fcd225f
  chmod +x for test_scripts

# Comments
Fixes test:
```
  2) CLI Specifications, LOGGING - cardano-wallet serve logging [SERIAL], LOGGING - Serve shuts down logging correctly
       "... a lot of logs here..." does not contain "Logging shutdown"
```
which now pass on Windows and Linux.


lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/CLI/Server.hs
```diff
@@ -153,7 +153,7 @@ spec = do
                 let cmd = Command
                         (commandName @t)
                         ["serve"
-                        , "--database", "/does-not-exist"
+                        , "--database", "/*"
```
KtorZ: :thinking: what was the error on windows :thinking: ? Was the folder simply created ?

rvl:
I think what's happening here is:

1. The directory check of `/does-not-exist` succeeds when it was assumed that it would fail (with a permissions error).

   I tested the following two commands on Windows.

   ```
   C:\Users\win\cw\tests>cardano-wallet-jormungandr.exe serve --database /does-not-exist --node-port 8081 --random-port --log-level DEBUG --genesis-block-hash  3c07030e36bfffe67e2e2ec09e5293d384637cd2f004356ef320f3fe6c52041a
   [cardano-wallet:Info:3] [2020-01-10 05:50:40.99 UTC] Using directory: /does-not-exist
   [cardano-wallet:Info:3] [2020-01-10 05:50:40.99 UTC] Running as v2020.1.7 (git revision: 21ee81799ffb815ddba756bd9d94640b789e969b)


   C:\Users\win\cw\tests>cardano-wallet-jormungandr.exe serve --database /* --node-port 8081 --random-port --log-level DEBUG --genesis-block-hash  3c07030e36bfffe67e2e2ec09e5293d384637cd2f004356ef320f3fe6c52041a
   [cardano-wallet:Info:3] [2020-01-10 05:50:53.69 UTC] Creating directory: /*
   [cardano-wallet:Debug:3] [2020-01-10 05:50:53.69 UTC] Logging shutdown.
   cardano-wallet-jormungandr.exe: \*: CreateDirectory "\\*": invalid argument (The filename, directory name, or volume label syntax is incorrect.)
   ```

   In the first, `C:\does-not-exist` is created.

2. Rather than promptly exiting at an error, the 10 second wait in the test case lapses.

3. `withCreateProcess` cleans up the `cardano-wallet-jormungandr.exe` process ([source](http://hackage.haskell.org/package/process-1.6.5.0/docs/src/System.Process.html#terminateProcess)).

   Windows TerminateProcess does not give cardano-wallet a chance to shut down cleanly (indicated by "Logging shutdown" message).

   From [Win32 docs](https://docs.microsoft.com/en-us/windows/win32/procthread/terminating-a-process)

   > If a process is terminated by TerminateProcess, all threads of the process are terminated immediately with no chance to run additional code. This means that the thread does not execute code in termination handler blocks.

So `/*` or any filename that is invalid on windows (e.g. `/aux`) would be sufficient. I pushed a comment to this PR. (Hopefully no developers are running tests with `sudo`. Otherwise they will actually end up with a directory `/*`. Then to remove this directory: `rm -rf /*` ... oops!)

KtorZ (approved):
Slightly worried about the consequences of creating a "/*" directory but :man_shrugging:

rvl:
`/aux` would achieve the same result as `/*` but looks more innocuous.

Issue #1235, opened 2019-12-30
Allow logs to be filtered by components


rvl:
# Context

[User story WB-31](https://jira.iohk.io/browse/WB-31)

As a technical user using the command-line
I want to be able to increase or decrease logging verbosity of specific components of the wallet server
So that I can debug more easily issues with the software

# Decision

# Acceptance Criteria

- When starting the wallet server (using `start` or `launch` command), it must be possible to control the logging severity of various distinct parts of the server.

---

# Development 

1. Move from `Trace m a` to `Tracer m a`
2. Collect top-level traces in a data type, and configure this at startup.
3. Remove cli `--quiet` and `--verbose`
4. Add cli options for enabling/disabling top-level traces, and setting their minimum severity. ⇒ #1241
5. ~Add cli options for altering verbosity of top-level traces (this is distinct from severity).~ ← really only useful with structured logging output.
8. Adjust cli integration tests accordingly ⇒ #1257
9. Update usage documentation on wiki ⇒ [CLI Wiki Page](https://github.com/input-output-hk/cardano-wallet/wiki/Wallet-command-line-interface#logging-options-for-launch--serve)

# QA 

- Added integration tests for new options ⇒ [`.../CLI/Server.hs`](https://github.com/input-output-hk/cardano-wallet/blob/0bd418e3362c8b041c9c211d1f1e93b7e4f718c9/lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/CLI/Server.hs#L115-L178)

piotr-iohk:
@rvl is this done/qa ready?

rvl:
@piotr-iohk sorry I forgot - I have updated the QA section.

piotr-iohk:
lgtm.

PR #1257, opened 2020-01-10
Add integration test for CLI tracer options

rvl:
Relates to #1235.

# Overview

- 758ab28117e55d4a54c59f53231becd04a40596a
  Add Cardano.Wallet.Logging.stdoutTextTracer
  This comes in handy when trying functions in GHCi.

- fd4069cdc82e06e4a45d064d74b9d531e629beb1
  More logging information at program startup
  
- dd22021afab13ae37808335510b2e5c319d9611e
  tests: Add integration test for per-component tracing

### DIFF
--- a/lib/core/src/Cardano/Wallet/Logging.hs
+++ b/lib/core/src/Cardano/Wallet/Logging.hs
@@ -17,6 +17,7 @@ module Cardano.Wallet.Logging
     , trMessage
     , trMessageText
     , filterTraceSeverity
+    , stdoutTextTracer
     ) where
 
 import Prelude
@@ -39,6 +40,8 @@ import Data.Text
     ( Text )
 import Data.Text.Class
     ( ToText (..) )
+
+import qualified Data.Text.IO as T
 
 -- | Converts a 'Text' trace into any other type of trace that has a 'ToText'
 -- instance.
@@ -82,6 +85,8 @@ trMessageText tr = Tracer $ \arg -> do
        <*> pure (LogMessage msg)
    traceWith tracer lo
 
+-- | Tracer transformer which converts 'Trace m a' to 'Tracer m a' by wrapping
+-- typed log messages into a 'LogObject'.
 trMessage
     :: (MonadIO m, DefinePrivacyAnnotation a, DefineSeverity a)
     => Tracer m (LogObject a)
@@ -116,3 +121,8 @@ filterNonEmpty tr = Tracer $ \arg -> do
   where
     nonEmptyMessage (LogMessage msg) = msg /= mempty
     nonEmptyMessage _ = True
+
+-- | Creates a tracer that prints any 'ToText' log message. This is useful for
+-- debugging functions in the REPL, when you need a 'Tracer' object.
+stdoutTextTracer :: (MonadIO m, ToText a) => Tracer m a
+stdoutTextTracer = Tracer $ liftIO . T.putStrLn . toText
--- a/lib/jormungandr/exe/cardano-wallet-jormungandr.hs
+++ b/lib/jormungandr/exe/cardano-wallet-jormungandr.hs
@@ -27,7 +27,7 @@ import Prelude
 import Cardano.BM.Data.Severity
     ( Severity (..) )
 import Cardano.BM.Trace
-    ( Trace, logInfo )
+    ( Trace, appendName, logDebug, logInfo )
 import Cardano.CLI
     ( Port (..)
     , cli
@@ -62,6 +62,7 @@ import Cardano.Wallet.Api.Server
     ( HostPreference, Listen (..) )
 import Cardano.Wallet.Jormungandr
     ( TracerSeverities
+    , Tracers
     , Tracers' (..)
     , serveWallet
     , setupTracers
@@ -75,12 +76,14 @@ import Cardano.Wallet.Jormungandr.Network
     , JormungandrConfig (..)
     , JormungandrConnParams (..)
     )
+import Cardano.Wallet.Logging
+    ( transformTextTrace )
 import Cardano.Wallet.Primitive.AddressDerivation
     ( NetworkDiscriminant (..) )
 import Cardano.Wallet.Primitive.Types
     ( BlockchainParameters, Hash (..), SyncTolerance )
 import Cardano.Wallet.Version
-    ( gitRevision, showFullVersion, version )
+    ( GitRevision, Version, gitRevision, showFullVersion, version )
 import Control.Applicative
     ( Const (..), optional, (<|>) )
 import Data.List
@@ -89,6 +92,8 @@ import Data.Maybe
     ( fromMaybe )
 import Data.Text
     ( Text )
+import Data.Text.Class
+    ( ToText (..) )
 import Network.Socket
     ( SockAddr )
 import Options.Applicative
@@ -114,6 +119,8 @@ import Options.Applicative
     )
 import Options.Applicative.Types
     ( readerAsk, readerError )
+import System.Environment
+    ( getArgs, getExecutablePath )
 import System.Exit
     ( exitWith )
 import System.FilePath
@@ -156,13 +163,12 @@ main = withUtf8Encoding $ do
         <> cmdVersion
 
 beforeMainLoop
-    :: Trace IO Text
+    :: Trace IO MainLog
     -> SockAddr
     -> Port "node"
     -> BlockchainParameters
     -> IO ()
-beforeMainLoop tr addr _ _ = do
-    logInfo tr $ "Wallet backend server listening on " <> T.pack (show addr)
+beforeMainLoop tr addr _ _ = logInfo tr $ MsgListenAddress addr
 
 {-------------------------------------------------------------------------------
                             Command - 'launch'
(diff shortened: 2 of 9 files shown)


lib/jormungandr/exe/cardano-wallet-jormungandr.hs
```diff
@@ -396,7 +399,7 @@ helperTracingText = unlines $
     [ "Additional tracing options:"
     , ""
     , "  --log-level SEVERITY     Global minimum severity for a message to be logged."
-    , "                           Defaults to \"INFO\" unless otherwise configured."
+    , "                           Defaults to \"DEBUG\" unless otherwise configured."
```
paweljakubas: :+1:

lib/jormungandr/src/Cardano/Wallet/Jormungandr.hs
```diff
@@ -201,7 +206,7 @@ serveWallet
 serveWallet
     Tracers{..} sTolerance databaseDir hostPref listen lj beforeMainLoop = do
     installSignalHandlers (traceWith applicationTracer MsgSigTerm)
-    traceWith applicationTracer MsgStarting
+    traceWith applicationTracer $ MsgStarting lj
```
paweljakubas: `lj` is here the only one name not self-descriptive (ie., in 207 line)

lib/jormungandr/exe/cardano-wallet-jormungandr.hs
```diff
@@ -412,3 +415,39 @@ helperTracingText = unlines $
         "  " ++ padRight maxLength ' ' name ++ "  " ++ desc
       where
         padRight n char str = take n $ str ++ replicate n char
+
+{-------------------------------------------------------------------------------
+                                    Logging
+-------------------------------------------------------------------------------}
+
+data Log
```
paweljakubas: we use `ApplicationLog` in ` lib/jormungandr/src/Cardano/Wallet/Jormungandr.hs` - maybe we could give some prefix to `Log` ?

paweljakubas:
@rvl @KtorZ  general question. We have logs datatypes decentralized thought the code. Why we didn't decide to put them in one place - I see this done in cardano-node for logging. Maybe it is better to have them in one place for maintenance (less expansive ripple effect in the codebase when changes happen) or it follows from how monitoring-framework is constructed? Nevertheless, I like that is is scattered in modules that logging concerns (modules take responsibility of their logging)... - it will be easier to decouple them and make them independent what is the path we are taking

rvl:
I asked about this on slack yesterday because I agree with your concern.

My preference would be to have logging types and ToText instances in the same module as the code which uses them, and the other instances (DefineSeverity, ToObject, etc) in an "orphans" module.

lib/jormungandr/exe/cardano-wallet-jormungandr.hs
rvl: Yeah, I had to discard this to make other integration tests happy.

lib/jormungandr/src/Cardano/Wallet/Jormungandr.hs
rvl: OK fixed - I was trying to stay under 80 columns. :grin:

lib/jormungandr/exe/cardano-wallet-jormungandr.hs
rvl: I changed it to `MainLog`.

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/CLI/Server.hs
```diff
@@ -131,6 +131,51 @@ spec = do
...
+                void $ withBackendProcess nullTracer cmd $ do
+                    threadDelay (5 * oneSecond)
+                hClose hLogs
+                logged <- T.lines <$> TIO.readFile logs
+                let grep str = filter (T.isInfixOf str)
+                let allDebugLogs = grep "Debug" logged
+                let componentDebugLogs = grep "stake-pool-db" allDebugLogs
+                let networkDebugLogs = grep "network" allDebugLogs
+                length componentDebugLogs `shouldNotBe` 0
+                length networkDebugLogs `shouldBe` 0
```
paweljakubas: so we here are serving with log-level=debug. after 5 seconds we close log handle. and then look into logs "greping" Debug ones. And we do not see neither stake-pool-db nor network related logs. What I do not undestand is why this test is titled `Serve debug logs for one component` Don't we want to show that something is present here, rather than absent?

paweljakubas: ah, my bad. there is `shouldNotBe` - yes, that was what I expected to capture!!!! :+1:

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/CLI/Server.hs
```diff
@@ -131,6 +131,51 @@ spec = do
...
+                        ]
+                        (pure ())
+                        (UseHandle hLogs)
+                void $ withBackendProcess nullTracer cmd $ do
+                    threadDelay (5 * oneSecond)
+                hClose hLogs
+                logged <- T.lines <$> TIO.readFile logs
+                let countLogs comp = length $
+                        filter (T.isInfixOf $ "cardano-wallet." <> comp) logged
+                countLogs "network" `shouldBe` 0
```
paweljakubas, line 176: indeed, after trace-network is off I expect to see no network related logs :+1:

paweljakubas (approved):
LGTM! (just slight suggestion to add more tests)

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/CLI/Server.hs
```diff
@@ -131,6 +131,51 @@ spec = do
...
 
+        it "LOGGING - Serve debug logs for one component" $ \ctx -> do
+            withTempFile $ \logs hLogs -> do
+                let cmd = Command
+                        (commandName @t)
+                        ["serve"
+                        , "--node-port", show (ctx ^. typed @(Port "node"))
+                        , "--random-port"
+                        , "--genesis-block-hash", block0H
+                        , "--log-level", "debug"
```
paweljakubas, line 144: this line is not needed according to defaults - that is my expectation. this is nit-picking, but without it we would show that defaults work properly. Maybe it is good idea (in some next PR) to extend this test in such a way that we play with different combinations of general-log level and component log level and see that defaults are as expected. What happens when component log level is Info and general is Debug, or opposite? Also check two component logs with the same, different log levels... What about showing that when Debug is set then we will not have for sure Info related logs?

rvl: I needed to change the default global minimum severity back to Info to prevent debug-level MainLog messages from leaking out.

If the global log level is Info then there will be no Debug-level messages whatsoever. There is a check for that.
It the global log level is Debug then there will be Debug, Info, and higher-severity messages.
I added an assertion for showing that when Info is set then we will have for sure Info related logs.
I think that should cover it.

rvl:
Oops

bors r+

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/CLI/Server.hs
KtorZ: @rvl would you mind making "debug" the default again and adding this very comment as a note in place where we set this default to DEBUG ? This is good to know and have in the source code. I didn't realize this when I changed the global default to be INFO.

lib/core/src/Cardano/Wallet/Logging.hs
```diff
@@ -116,3 +121,8 @@ filterNonEmpty tr = Tracer $ \arg -> do
   where
     nonEmptyMessage (LogMessage msg) = msg /= mempty
     nonEmptyMessage _ = True
+
+-- | Creates a tracer that prints any 'ToText' log message. This is useful for
+-- debugging functions in the REPL, when you need a 'Tracer' object.
+stdoutTextTracer :: (MonadIO m, ToText a) => Tracer m a
+stdoutTextTracer = Tracer $ liftIO . T.putStrLn . toText
```
KtorZ, line 128: :+1:

lib/jormungandr/src/Cardano/Wallet/Jormungandr.hs
```diff
@@ -199,19 +204,25 @@ serveWallet
...
-            either id id <$> race
-                (daedalusIPC daedalusIPCTracer wPort $> ExitSuccess)
-                (serveApp socket)
+    Tracers{..}
+    sTolerance
+    databaseDir
+    hostPref
+    listen
+    backend
+    beforeMainLoop = do
```
KtorZ: That's the kind of moment where I think it's totally fine to not be under 80-lines :grimacing: ...

KtorZ (approved):
LGTM

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/CLI/Server.hs
rvl: I have put that change into #1282. There are are few extra debug severity MainLog messages printed due to this, but I think it's ok.

rvl:
Needed to update the launcher tests too, and update some tracer names.

bors r+

Issue #1250, opened 2020-01-07
Port pool desirability formulas from delegation design document


Anviking:
# Context

The ranking of pools in the ITN are sub-optimal.

We want to adhere to the ranking described in the [Delegation Design Spec](https://hydra.iohk.io/build/1389333/download/1/delegation_design_spec.pdf), ignoring the differences of:
- Jörmungandr Pool Max Tax Cap (nothing we can do about it, most pools don't have it set anyway)
- Leader stake (always 0 in Jörmungandr)

# Decision

Convert relevant formulas in section 6.5 and 6.6 to well-documented and understandable code, maintaining close correspondence and structure with the specs.

If the usefulness of a formula is not clear to the wallet, we may still consider adding it as code for the sake of completeness, learning, and structure.

# Acceptance Criteria

- Relevant formulas in 6.5 and 6.6 _should_ be ported
- The code _must_ be descriptive and well-documented
- The code _should_ follow the same structure as the spec (to make it easy to compare the two)

---

# Development 

- #1259 

# QA 

- Ported fewer formulas than first anticipated. https://github.com/input-output-hk/cardano-wallet/blob/6f09f4bbc58e837547a6ae5c66575e880f060c07/lib/core/src/Cardano/Pool/Ranking.hs#L68
- Currently unused.
- Fairly strict newtypes making sure that we later on don't pass in invalid values. `unsafeMkRatio`, `unsafeMkPositive`, `unsafeMkNonNegative` will crash otherwise. Fairly extensive property tests: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Cardano/Pool/RankingSpec.hs#L75-L121

piotr-iohk:
lgtm

PR #1259, opened 2020-01-10
Calculate pool desirability according to delegation design spec

Anviking:
# Issue Number

#1250 

# Overview

- [x] New completely isolated module implementing ~`nonMyopicMemberRewards`~ `desirability` and related helper functions from the [spec](https://hydra.iohk.io/build/1389333/download/1/delegation_design_spec.pdf).
- [x] Some simple unit/property tests as sanity checks

# Comments

- To be squashed before merging. Leaving a couple un-squashed for now, for history.

⬇️ Now reverted. Coming in a later PR! (Task #1276 )

Work in progress example ranking (without averaging, and with some hard-coded slightly off values)
Skärmavbild 2020-01-10 kl  18 46 39

With basic averaging (not exponential):

Skärmavbild 2020-01-10 kl  21 06 41

A newer screenshot (lots of 1PCT pools at the top now):

Skärmavbild 2020-01-13 kl  15 47 52

### DIFF
--- a/lib/core/src/Cardano/Pool/Ranking.hs
+++ b/lib/core/src/Cardano/Pool/Ranking.hs
@@ -1,0 +1,206 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeOperators #-}
+
+-- | A self contained module for ranking pools according to the delegation design
+-- spec /Design Specification for Delegation and Incentives in Cardano/
+-- [(Kant et al, 2019)](https://hydra.iohk.io/build/1389333/download/1/delegation_design_spec.pdf).
+--
+-- The module currently implements the non-myopic desirability, and might later
+-- support calculating the full non-myopic pool member rewards. The latter being
+-- the recomended way to rank stake-pools (see section 4.3).
+--
+-- The term non-myopic is explained on page 37:
+--
+-- @
+-- It would be short-sighted (“myopic”) for stakeholders to directly use the
+-- reward splitting formulas from Section 6.5. They should instead take the
+-- long-term (“non-myopic”) view. To this end, the system will calculate and
+-- display the “non-myopic” rewards that pool leaders and pool members can
+-- expect, thus supporting stakeholders in their decision whether to create a
+-- pool and to which pool to delegate their stake.
+--
+-- The idea is to first rank all pools by “desirability”, to then assume that
+-- the k most desirable pools will eventually be saturated, whereas all other
+-- pools will lose all their members, then to finally base all reward
+-- calculations on these assumptions.
+-- @
+--
+-- == Relevant identifiers
+--
+-- Epoch Parameters
+--
+-- +------------+-------------+------------------------------------------------+
+-- |identifier  | domain      | description                                    |
+-- +============+=============+================================================+
+-- + R          | @ [0, ∞) @  | total available rewards for the epoch (in ada).|
+-- +------------+-------------+------------------------------------------------+
+-- + a0         | @ [0, ∞) @  | owner-stake influence on pool rewards.         |
+-- +------------+-------------+------------------------------------------------+
+-- + k          | @ [0, ∞) @  | desired number of pools                        |
+-- +------------+-------------+------------------------------------------------+
+-- + z0         | 1/k         | relative stake of a saturated pool             |
+-- +------------+-------------+------------------------------------------------+
+--
+-- Pool's Parameters
+--
+-- +------------+-------------+------------------------------------------------+
+-- |identifier  | domain      | description                                    |
+-- +============+=============+================================================+
+-- | c          | @ [0, ∞)@   | costs                                          |
+-- +------------+-------------+------------------------------------------------+
+-- | f          | @ [0, ∞) @  | rewards                                        |
+-- +------------+-------------+------------------------------------------------+
+-- | m          | @ [0, 1] @  | margin                                         |
+-- +------------+-------------+------------------------------------------------+
+-- | p_apparent | @ [0, 1] @  | apparent performance                           |
+-- +------------+-------------+------------------------------------------------+
+-- | s          | @ [0, ∞) @  | relative stake of the pool leader (i.e pledge) |
+-- +------------+-------------+------------------------------------------------+
+-- | σ          | @ [0, 1] @  | total relative stake of the pool               |
+-- +------------+-------------+------------------------------------------------+
+module Cardano.Pool.Ranking
+    (
+      -- * Formulas
+      desirability
+
+    , saturatedPoolRewards
+    , saturatedPoolSize
+
+      -- * Types
+    , EpochConstants (..)
+    , Pool (..)
+    , RelativeStakeOf (..)
+    , mkRelativeStake
+    , Lovelace (..)
+    , Margin
+    , unsafeMkMargin
+    , getMargin
+    , NonNegative (..)
+    , Positive (..)
+    , unsafeToPositive
+    , unsafeToNonNegative
+    )
+    where
+
+import Prelude
+
+import GHC.Generics
+    ( Generic )
+import GHC.TypeLits
+    ( Symbol )
+
+--------------------------------------------------------------------------------
+-- Formulas from spec
+--------------------------------------------------------------------------------
+
+-- | Non-myopic pool desirability according to section 5.6.1.
+--
+-- Is /not/ affected by oversaturation nor pool stake in general.
+desirability
+    :: EpochConstants
+    -> Pool
+    -> Double
+desirability constants pool
+    | f_saturated <= c = 0
+    | otherwise    = (f_saturated - c) * (1 - m)
+  where
+    f_saturated = saturatedPoolRewards constants pool
+    m = getMargin $ margin pool
+    c = getNonNegative $ getLovelace $ cost pool
+
+-- | Total rewards for a pool if it were saturated.
+--
+-- When a0 = 0 this reduces to just p*R*z0 (tested
+-- by @prop_saturatedPoolRewardsReduces@)
+saturatedPoolRewards :: EpochConstants -> Pool -> Double
+saturatedPoolRewards constants pool =
+    let
+        a0 = getNonNegative $ leaderStakeInfluence constants
+        z0 = unRelativeStake $ saturatedPoolSize constants
+        s = unRelativeStake $ leaderStake pool
+        _R = getNonNegative $ getLovelace $ totalRewards constants
+        p = getNonNegative $ recentAvgPerformance pool
+        -- ^ technically \hat{p} in the spec
+    in
+        (p * _R) / (1 + a0)
+        * (z0 + ((min s z0) * a0))
+
+-- | Determines z0, i.e 1 / k
+saturatedPoolSize :: EpochConstants -> RelativeStakeOf "pool"
+saturatedPoolSize constants =
+    RelativeStake $ 1 / fromIntegral (getPositive $ optimalNumberOfPools constants)
+
+--------------------------------------------------------------------------------
+-- Types
+--------------------------------------------------------------------------------
+
+data EpochConstants = EpochConstants
+    { leaderStakeInfluence :: NonNegative Double
+      -- ^ a_0
+    , optimalNumberOfPools :: Positive Int
+      -- ^ k
+    , totalRewards :: Lovelace
+      -- ^ Total rewards in an epoch. "R" in the spec.
+    } deriving (Show, Eq, Generic)
+
+data Pool = Pool
+    { leaderStake :: RelativeStakeOf "pool leader"
+    , cost :: Lovelace
+    , margin :: Margin
+    , recentAvgPerformance :: NonNegative Double
+      -- ^ An already averaged performance-value
+    } deriving (Show, Eq, Generic)
+
+newtype Lovelace = Lovelace { getLovelace :: NonNegative Double }
+    deriving (Show, Eq)
+    deriving newtype (Ord, Num)
+
+newtype Margin = Margin { getMargin :: Double }
+    deriving (Show, Eq)
+    deriving newtype Ord
+
+unsafeMkMargin :: Double -> Margin
+unsafeMkMargin x
+    | x >= 0 && x <= 1  = Margin x
+    | otherwise         = error $ "unsafeMkMargin: " ++ show x
+                          ++ "not in range [0, 1]"
+
+-- | Stake relative to the total active stake in an epoch
+--
+-- The value
+-- 0.01 :: RelativeStakeOf "pool"
+-- would mean that a pool has a stake that is 1% of the total active stake in
+-- the epoch.
+newtype RelativeStakeOf (tag :: Symbol)
+    = RelativeStake { unRelativeStake :: Double }
+    deriving (Eq, Ord, Show, Generic)
+    deriving newtype (Num, Fractional)
+
+mkRelativeStake :: Lovelace -> EpochConstants -> RelativeStakeOf (tag :: Symbol)
+mkRelativeStake (Lovelace (NonNegative stake)) constants =
+    RelativeStake $ stake / total
+  where
+    total = getNonNegative $ getLovelace $ totalRewards constants
+
+newtype Positive a = Positive { getPositive :: a }
+    deriving (Generic, Eq, Show)
+    deriving newtype (Ord, Num)
+
+unsafeToPositive :: (Ord a, Show a, Num a) => a -> (Positive a)
+unsafeToPositive x
+    | x > 0    = Positive x
+    | otherwise = error $ "unsafeToPositive: " ++ show x ++ " > 0 does not hold"
+
+newtype NonNegative a = NonNegative { getNonNegative :: a }
+    deriving (Generic, Eq, Show)
+    deriving newtype (Ord, Num)
+
+unsafeToNonNegative :: (Ord a, Show a, Num a) => a -> (NonNegative a)
+unsafeToNonNegative x
+    | x >= 0    = NonNegative x
+    | otherwise = error $ "unsafeToNegative: " ++ show x ++ " >= 0 does not hold"
--- a/lib/core/src/Cardano/Pool/Ranking.hs
+++ b/lib/core/src/Cardano/Pool/Ranking.hs
@@ -77,9 +77,9 @@ module Cardano.Pool.Ranking
     , RelativeStakeOf (..)
     , mkRelativeStake
     , Lovelace (..)
-    , Margin
-    , unsafeMkMargin
-    , getMargin
+    , Ratio
+    , unsafeToRatio
+    , getRatio
     , NonNegative (..)
     , Positive (..)
     , unsafeToPositive
@@ -110,7 +110,7 @@ desirability constants pool
     | otherwise    = (f_saturated - c) * (1 - m)
   where
     f_saturated = saturatedPoolRewards constants pool
-    m = getMargin $ margin pool
+    m = getRatio $ margin pool
     c = getNonNegative $ getLovelace $ cost pool
 
 -- | Total rewards for a pool if it were saturated.
@@ -153,7 +153,7 @@ data Pool = Pool
       -- ^ s
     , cost :: Lovelace
       -- ^ c
-    , margin :: Margin
+    , margin :: Ratio
       -- ^ m
     , recentAvgPerformance :: NonNegative Double
       -- ^ \hat{p}, an already averaged performance-value.
@@ -165,14 +165,14 @@ newtype Lovelace = Lovelace { getLovelace :: NonNegative Double }
     deriving (Show, Eq)
     deriving newtype (Ord, Num)
 
-newtype Margin = Margin { getMargin :: Double }
+newtype Ratio = Ratio { getRatio :: Double }
     deriving (Show, Eq)
     deriving newtype Ord
 
-unsafeMkMargin :: Double -> Margin
-unsafeMkMargin x
-    | x >= 0 && x <= 1  = Margin x
-    | otherwise         = error $ "unsafeMkMargin: " ++ show x
+unsafeToRatio :: Double -> Ratio
+unsafeToRatio x
+    | x >= 0 && x <= 1  = Ratio x
+    | otherwise         = error $ "unsafeToRatio: " ++ show x
                           ++ "not in range [0, 1]"
 
 -- | Stake relative to the total active stake in an epoch
--- a/lib/core/src/Cardano/Pool/Ranking.hs
+++ b/lib/core/src/Cardano/Pool/Ranking.hs
@@ -153,9 +153,10 @@ data Pool = Pool
     , margin :: Ratio
       -- ^ m
     , recentAvgPerformance :: NonNegative Double
-      -- ^ \hat{p}, an already averaged performance-value.
+      -- ^ \hat{p}, an already averaged (apparent) performance-value.
       --
-      -- Should mostly be in the range [0, 1], but lucky pools may exceed 1.
+      -- Should mostly be in the range [0, 1]. May be higher than 1 due to
+      -- randomness.
     } deriving (Show, Eq, Generic)
 
 newtype Lovelace = Lovelace { getLovelace :: Word64 }
--- a/lib/core/src/Cardano/Pool/Ranking.hs
+++ b/lib/core/src/Cardano/Pool/Ranking.hs
@@ -133,7 +133,7 @@ saturatedPoolRewards constants pool =
 -- | Determines z0, i.e 1 / k
 saturatedPoolSize :: EpochConstants -> RelativeStakeOf "pool"
 saturatedPoolSize constants =
-    RelativeStake $ 1 / fromIntegral (getPositive $ optimalNumberOfPools constants)
+    RelativeStake $ 1 / fromIntegral (getPositive $ desiredNumberOfPools constants)
 
 --------------------------------------------------------------------------------
 -- Types
@@ -142,7 +142,7 @@ saturatedPoolSize constants =
 data EpochConstants = EpochConstants
     { leaderStakeInfluence :: NonNegative Double
       -- ^ a_0
-    , optimalNumberOfPools :: Positive Int
+    , desiredNumberOfPools :: Positive Int
       -- ^ k
     , totalRewards :: Lovelace
       -- ^ Total rewards in an epoch. "R" in the spec.
@@ -150,10 +150,15 @@ data EpochConstants = EpochConstants
 
 data Pool = Pool
     { leaderStake :: RelativeStakeOf "pool leader"
+      -- ^ s
     , cost :: Lovelace
+      -- ^ c
     , margin :: Margin
+      -- ^ m
     , recentAvgPerformance :: NonNegative Double
-      -- ^ An already averaged performance-value
+      -- ^ \hat{p}, an already averaged performance-value.
+      --
+      -- Should mostly be in the range [0, 1], but lucky pools may exceed 1.
     } deriving (Show, Eq, Generic)
 
 newtype Lovelace = Lovelace { getLovelace :: NonNegative Double }
--- a/lib/core/src/Cardano/Pool/Ranking.hs
+++ b/lib/core/src/Cardano/Pool/Ranking.hs
@@ -2,7 +2,6 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
-{-# LANGUAGE KindSignatures #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeOperators #-}
 
@@ -74,12 +73,11 @@ module Cardano.Pool.Ranking
       -- * Types
     , EpochConstants (..)
     , Pool (..)
-    , RelativeStakeOf (..)
-    , mkRelativeStake
     , Lovelace (..)
     , Ratio
     , unsafeToRatio
     , getRatio
+    , unsafeMkRelativeStake
     , NonNegative (..)
     , Positive (..)
     , unsafeToPositive
@@ -91,8 +89,6 @@ import Prelude
 
 import GHC.Generics
     ( Generic )
-import GHC.TypeLits
-    ( Symbol )
 
 --------------------------------------------------------------------------------
 -- Formulas from spec
@@ -121,8 +117,8 @@ saturatedPoolRewards :: EpochConstants -> Pool -> Double
 saturatedPoolRewards constants pool =
     let
         a0 = getNonNegative $ leaderStakeInfluence constants
-        z0 = unRelativeStake $ saturatedPoolSize constants
-        s = unRelativeStake $ leaderStake pool
+        z0 = getRatio $ saturatedPoolSize constants
+        s = getRatio $ leaderStake pool
         _R = getNonNegative $ getLovelace $ totalRewards constants
         p = getNonNegative $ recentAvgPerformance pool
         -- ^ technically \hat{p} in the spec
@@ -131,9 +127,9 @@ saturatedPoolRewards constants pool =
         * (z0 + ((min s z0) * a0))
 
 -- | Determines z0, i.e 1 / k
-saturatedPoolSize :: EpochConstants -> RelativeStakeOf "pool"
+saturatedPoolSize :: EpochConstants -> Ratio
 saturatedPoolSize constants =
-    RelativeStake $ 1 / fromIntegral (getPositive $ desiredNumberOfPools constants)
+    Ratio $ 1 / fromIntegral (getPositive $ desiredNumberOfPools constants)
 
 --------------------------------------------------------------------------------
 -- Types
@@ -149,7 +145,7 @@ data EpochConstants = EpochConstants
     } deriving (Show, Eq, Generic)
 
 data Pool = Pool
-    { leaderStake :: RelativeStakeOf "pool leader"
+    { leaderStake :: Ratio
       -- ^ s
     , cost :: Lovelace
       -- ^ c
@@ -175,20 +171,9 @@ unsafeToRatio x
     | otherwise         = error $ "unsafeToRatio: " ++ show x
                           ++ "not in range [0, 1]"
 
--- | Stake relative to the total active stake in an epoch
---
--- The value
--- 0.01 :: RelativeStakeOf "pool"
--- would mean that a pool has a stake that is 1% of the total active stake in
--- the epoch.
-newtype RelativeStakeOf (tag :: Symbol)
-    = RelativeStake { unRelativeStake :: Double }
-    deriving (Eq, Ord, Show, Generic)
-    deriving newtype (Num, Fractional)
-
-mkRelativeStake :: Lovelace -> EpochConstants -> RelativeStakeOf (tag :: Symbol)
-mkRelativeStake (Lovelace (NonNegative stake)) constants =
-    RelativeStake $ stake / total
+unsafeMkRelativeStake :: Lovelace -> EpochConstants -> Ratio
+unsafeMkRelativeStake (Lovelace (NonNegative stake)) constants =
+    unsafeToRatio $ stake / total
   where
     total = getNonNegative $ getLovelace $ totalRewards constants
 
--- a/lib/core/src/Cardano/Pool/Ranking.hs
+++ b/lib/core/src/Cardano/Pool/Ranking.hs
@@ -75,13 +75,13 @@ module Cardano.Pool.Ranking
     , Pool (..)
     , Lovelace (..)
     , Ratio
-    , unsafeToRatio
+    , unsafeMkRatio
     , getRatio
     , unsafeMkRelativeStake
     , NonNegative (..)
     , Positive (..)
-    , unsafeToPositive
-    , unsafeToNonNegative
+    , unsafeMkPositive
+    , unsafeMkNonNegative
     )
     where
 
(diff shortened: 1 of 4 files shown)


lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -144,6 +146,7 @@ data StakePool = StakePool
     , stake :: Quantity "lovelace" Word64
     , production :: Quantity "block" Word64
     , apparentPerformance :: Double
+    , desirabilitiy :: Double
```
jonathanknowles:
```suggestion
    , desirability :: Double
```

lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -352,12 +355,44 @@ newStakePoolLayer tr db@DBLayer{..} nl metadataDir = StakePoolLayer
...
+            in  map (\((pool, x), des) -> (pool {desirability = des}, x))
+                . sortOn (Down . snd)
+                . (P.nonMyopicMemberRewards jormungandrEnv toM memberStake)
+                $ xs
+
+    -- | Constant needed for ranking pools
+    jormungandrEnv :: P.Env
+    jormungandrEnv = P.Env
+        { P.leaderStakeInfluence = 0
+        , P.punishmentHarshness = 1 -- TODO: What should this be?
```
Anviking: Need to decide on this.

Skärmavbild 2020-01-13 kl  15 22 04

lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -352,12 +355,44 @@ newStakePoolLayer tr db@DBLayer{..} nl metadataDir = StakePoolLayer
...
+        , P.optimalNumberOfPools = 100
+        , P.totalRewards = 1000000*1000000 -- TODO: What should this be?
```
Anviking: Will determine the influence of fixed `cost` on the ranking.

Seems non-trivial (but perhaps possible) to calculate from jörmungandr's reward constants. If e.g. tx fees can affect it significantly, it might be tricky to take into account.

I suspect it's okay-ish to ship with a hard-coded estimation. But pools with high `cost` will be ranked unfairly.

Anviking: *will try calculating from jörmungandr's config params*

https://github.com/input-output-hk/chain-libs/blob/master/chain-impl-mockchain/src/rewards.rs#L147-L190 🤔

Anviking: Ideally we wouldn't duplicate the calculation. Hopefully we can ask the Haskell Shelley Node for `R` over the LocalStateQuery protocol, when we get to it.

lib/core/src/Cardano/Pool/Performance.hs
```diff
@@ -0,0 +1,278 @@
...
+    } deriving (Show, Eq)
+
+type Cost = Lovelace
+type Margin = Double
+newtype Rewards = Rewards Double
+
+withRewards :: Rewards -> (Double -> Double) -> Rewards
+withRewards (Rewards r) f = Rewards $ f r
+
+type Lovelace = Double
```
KtorZ: Let's not use type aliases for immediate obvious types. Especially with double indirections `type Cost = Lovelace = Double`. 
Type aliases are harming readability more than they're helping. Prefer:

- newtypes if there are properties that need to hold for that type
- record-types to disambiguate arguments passed to functions. e.g.

```hs
poolMemberReward
    :: Rewards -- \hat{f}
    -> Cost -- c
    -> Margin  -- m
    -> RelativeStakeOf "member" -- t
    -> RelativeStakeOf "pool" -- \sigma
    -> Double
poolMemberReward
```

becomes:

```hs
poolMemberReward 
    :: PoolParameters
    -> RelativeStakeOf "member"
    -> RelativeStakeOf "pool"
    -> Double
poolMemberReward PoolParameters{cost,margin,rewards} t σ
```

Having said that, in this context, I'd opt for sticking to the paper's terminology as much as possible to make review and comparison easier. Cf other comment about identifiers.

lib/core/src/Cardano/Pool/Performance.hs
```diff
@@ -0,0 +1,278 @@
...
+    ( sortOn )
+import GHC.TypeLits
+    ( Symbol )
+import Prelude
+
+--------------------------------------------------------------------------------
+-- Types
+--------------------------------------------------------------------------------
+
+data Env = Env
```
KtorZ: Avoid generic names like `Env` or `Config` which gives poor information about what the type really is about (unless meant for a qualified usage). Prefer something like `IncentiveParameters`

lib/core/src/Cardano/Pool/Performance.hs
```diff
@@ -0,0 +1,278 @@
...
+saturatedPoolSize env =
+    RelativeStake $ 1 / fromIntegral (optimalNumberOfPools env)
+
+data Pool = Pool
+    { leaderStake :: RelativeStakeOf "pool leader"
+    , stake :: RelativeStakeOf "pool"
+    , cost :: Lovelace
+    , margin :: Margin
+    , recentAvgPerformance :: Double
+      -- ^ An already averaged performance-value. This should be an exponential
```
KtorZ: ~~should be~~ could be. Be careful, these words have a precise meaning.

lib/core/src/Cardano/Pool/Performance.hs
```diff
@@ -0,0 +1,278 @@
...
+--
+-- The idea is to first rank all pools by “desirability”, to then assume that
+-- the k most desirable pools will eventually be saturated, whereas all other
+-- pools will lose all their members, then to finally base all reward
+-- calculations on these assumptions.
+--
+-- (Kant et al, 2019)
+
+-- @Rank@ and desirability here are temporary concepts that should not be
+-- confused with the ordering of stake-pools in Daedalus.
```
KtorZ: I do not understand this comment, can you clarify :thinking:  ?

lib/core/src/Cardano/Pool/Performance.hs
```diff
@@ -0,0 +1,278 @@
...
+type Rank = Int
+
+-- | How desirable a pool would be if it were saturated.
+--
+-- (6.6.1 Pool Desirability and Ranking)
```
KtorZ: *5.6.1

lib/core/src/Cardano/Pool/Performance.hs
```diff
@@ -0,0 +1,278 @@
...
+saturatedDesirability
```
KtorZ: There's no such thing as "unsaturatedDesirability" so what about calling this "desirability" to better match the spec?

lib/core/src/Cardano/Pool/Performance.hs
```diff
@@ -0,0 +1,278 @@
...
+    :: Env
+    -> Cost
+    -> Margin
+    -> RelativeStakeOf "pool leader"
+    -> Double
+saturatedDesirability env cost' margin' (RelativeStake leaderStake') =
+        if saturatedPoolRewards > cost'
+        then (saturatedPoolRewards - cost') * (1 - margin')
+        else 0
```
KtorZ: Alternatively, 

```hs
    | saturedPoolRewards <= cost' = 0
    | otherwise                   = (saturatedPoolRewards - cost') * (1 - margin')
```

to be closer to the spec notation + guards read better than if-then-else.

lib/core/src/Cardano/Pool/Performance.hs
```diff
@@ -0,0 +1,278 @@
...
+    -> Double
+poolMemberReward
+    (Rewards rew)
+    cost'
+    margin'
+    (RelativeStake memberStake)
+    (RelativeStake totPoolStake) =
+    if rew > cost'
+    then (rew - cost') * (1 - margin') * (memberStake / totPoolStake)
+    else 0
```
KtorZ: The calculation of pool and member rewards is way out of scope for this U/S. The calculation of `R` shouldn't force us to re-implement the whole reward mechanism. Rather, this is something we want to fetch from the node's.

lib/core/src/Cardano/Pool/Performance.hs
```diff
@@ -0,0 +1,278 @@
...
+-- z_0 = 1/k -> size of a saturated pool
+-- σ -> relative stake of pool (includes stake from leader)
+-- s -> relative stake of pool leader
+-- σ’  = min(σ, z_0)
+-- s’ = min(s, z_0)
+-- p_j_expected = n_j / max(N_expected, 1)
+-- \hat{f} -> pool rewards (includes average!)
+-- c -> pool costs in ADA
+-- m : [0, 1]  -> margin
+-- σ : [0,1]
```
KtorZ: What about using these identifiers consistently below instead of ad-hoc names like `saturedPoolRewards` ? Also, after a quick look, it looks like we truly need only a subset of these which we can categorize as follows:

```
Epoch Parameters
================

identifier   | domain / definition      | description
---          | ---                      | ---
R            | [0, ∞)                   | the total available rewards for the epoch (in ada).
a0           | [0, ∞)                   | owner-stake influence on pool rewards.
k            | [[0, ∞)                  | desired number of pools
z0           | 1/k                      | size of a saturated pool


Pool's Parameters
=================

identifier | domain / definition | description
---        | ---                 | ---
c          | [0, ∞)              | pool's costs
f          | [0, ∞)              | pool's rewards
m          | [0, 1]              | pool's margin
p_apparent | [0, 1]              | pool's apparent performance
s          | [0, ∞)              | relative stake of pool leader (i.e. pledge amount / total pool's stake)
σ          | [0, 1]              | relative stake of the pool, the fraction of the absolute stake that the pool controls
```

lib/core/src/Cardano/Pool/Performance.hs
```diff
@@ -0,0 +1,278 @@
...
+-- This formula holds if the pool is amongst the top k rated pools. If it is
+-- not, we should expect no rational player to delegate to it, and accordingly
+-- we will have to replace σnm by s + t in
+nonMyopicStakeOfPool
+    :: Env
+    -> RelativeStakeOf "pool leader"
+    -> Rank
+    -> RelativeStakeOf "pool"
+    -> RelativeStakeOf "pool"
+nonMyopicStakeOfPool env _leaderStake' rank poolStake
```
KtorZ: This is out-of-scope.

lib/core/src/Cardano/Pool/Performance.hs
```diff
@@ -0,0 +1,278 @@
...
+-- carry information.
+--
+-- The order of the list is not preserved.
+nonMyopicMemberRewards
+    :: Env
+    -> (pool -> Pool)
+    -> RelativeStakeOf "member"
+    -> [pool]
+    -> [(pool, Double)]
+nonMyopicMemberRewards env convert memberStake pools =
```
KtorZ: This is out-of-scope.

KtorZ (changes requested):
I am a bit confused. There seems to be a lot of focus on the calculation of rewards, rewards per stakeholder and  whereas.. the point of the user story was to compute the `desirability` in order to rank pools.

This boils down to:

```hs
desirability (R, z0, a0) c m s p_apparent
    | f_satured <= c = 0
    | otherwise      = (f_satured - c) * (1 - m)
  where
    f_satured = p_apparent * optimalReward (R, z0, a0) z0 s

optimalReward (R, z0, a0) σ s = 
    R / (1 + a0) * (σ' + s'*a0*(σ' - s'*(z0-σ')/z0) / z0)
  where
    σ' = min σ z0              
    s' = min s z0
```

where `p_apparent` is already computed by the `Metrics` module. There's the question of `R` which we are definitely not looking to re-compute (unless trivial) but, fetch from the node. 

Please revise and focus on the calculation of `desirability` as highlighted above. I'd recommend sticking with the same terminology as in the specs to make the mapping immediate and avoid potential translation errors. Keep types simple (`Double` are fine here, really). Leave the calculation or fetching of `R` to another PR.

Anviking:
@KtorZ I had to double-check, but there is reason to this madness. 

We *are* supposed to order by non-myopic member rewards, not by "desirability". 

"Desirability" does not take actual performance into account. I found the name confusing. Hence the comment and renaming to `saturatedDesirability`.

Skärmavbild 2020-01-14 kl  14 09 14

I will make this clearer.

KtorZ:
@Anviking No! This is why we have user stories. If the story specifies X, it's not to implement Y. 
If you don't understand why `X` or think that it should be `Y`, then ask. 

We are doing baby steps and are looking to implement the **desirability**, not the non-myopic pool member rewards _(yet)_. So please, stay focus on the task.

Anviking:
> "Desirability" does not take actual performance into account. I found the name confusing. Hence the comment and renaming to saturatedDesirability.

For the record: I was looking at the wrong version of the spec locally :/. The `desirability` in the latest version does take performance into account.

lib/core/src/Cardano/Pool/Performance.hs
Anviking: I was looking at the old spec -> explains it.

lib/core/src/Cardano/Pool/Performance.hs
```diff
@@ -0,0 +1,158 @@
...
+    ( Generic )
+import GHC.TypeLits
+    ( KnownNat, Nat, Symbol, natVal )
+import Prelude
+
+--------------------------------------------------------------------------------
+-- Types
+--------------------------------------------------------------------------------
+
+newtype a >= (b :: Nat) = Bounded { unBounded :: a }
```
Anviking: Will likely remove.

lib/core/test/unit/Cardano/Pool/PerformanceSpec.hs
```diff
@@ -0,0 +1,131 @@
...
+                    allElseEqualProperty
+                        "performance"
+                        (P.recentAvgPerformance)
+                        (\x y -> x{P.recentAvgPerformance = y})
+                        (show . P.unBounded)
+                        (>)
+
+-- | If two pools differ only in one field, and one is more desirable than the
+-- other, how should the field values compare?
+allElseEqualProperty
```
Anviking: Sadly the important `(<)` `(>)` distinction on the call-sites ended up less readable than I'd hoped.

lib/core/src/Cardano/Pool/Performance.hs
Anviking: Under the constraint "the wallet should order pools by non-myopic member rewards" this holds by definition. Under the constraint "the wallet should order pools by desirability" — no.

lib/core/src/Cardano/Pool/Performance.hs
KtorZ: What about using something more like `newtype NonNegative a = NonNegative a` ? A phantom type sounds a bit convoluted.

lib/core/src/Cardano/Pool/Performance.hs
```diff
@@ -0,0 +1,155 @@
...
+    -> Double
+desirability env (Pool (RelativeStake leaderStake') (Lovelace c) (Margin m) (Bounded p))
+    | saturatedPoolRewards <= c = 0
+    | otherwise                 = (saturatedPoolRewards - c) * (1 - m)
+  where
+    -- | Total rewards for a pool if it were saturated.
+    saturatedPoolRewards :: Double
+    saturatedPoolRewards =
+        let
+            a0 = unBounded $ leaderStakeInfluence env
```
Anviking: I am still using the `env` variable name after renaming the type. Should change.

I'm still not sure of the name of this module. But I like importing it as `qualified`. It might affect the name of what is now `RankingConstants`.

Anviking:
```
Progress 1/2: cardano-wallet-core
Cardano.Pool.Ranking
  Stake Pool Ranking
    all else equal...
      higher margin => lower desirability (or equal)
        +++ OK, passed 100 tests (60% equal desirability).
      higher cost => lower desirability (or equal)
        +++ OK, passed 100 tests (56% equal desirability).
      higher performance => higher desirability (or equal)
        +++ OK, passed 100 tests (53% equal desirability).
      higher oversaturation => desirability NOT affected
        +++ OK, passed 100 tests; 52 discarded.
    desiraility
      desirability >= 0
        +++ OK, passed 100 tests.
    saturatedPoolRewards
      a0 == 0 => saturatedPoolRewards == p*R*z0
        +++ OK, passed 100 tests.
      saturatedPoolRewards >= 0
        +++ OK, passed 100 tests.
    mkRelativeStake
      stake > R, R >= 1 => mkRelativeStake ∈ [0,1] (even with conversions)
        +++ OK, passed 100 tests; 106 discarded.

Finished in 0.0117 seconds
8 examples, 0 failures
```

lib/core/test/unit/Cardano/Pool/PerformanceSpec.hs
Anviking: Think I can simplify the structure a bit. Will do tomorrow.

lib/core/src/Cardano/Pool/Ranking.hs
```diff
@@ -0,0 +1,170 @@
...
+    let
+        a0 = unBounded $ leaderStakeInfluence constants
+        z0 = unRelativeStake $ saturatedPoolSize constants
+        s = unRelativeStake $ leaderStake pool
+        _R = getLovelace $ unBounded $ totalRewards constants
+        p = unBounded $ recentAvgPerformance $ pool
+        -- ^ technically \hat{p} in the spec
+    in
+        (p * _R) / (1 + a0)
+        * (z0 + ((min s z0) * a0))
```
KtorZ, line 128: May I suggest to define the "optimalRewards" at the top-level as:

```hs
optimalReward (R, z0, a0) σ s = 
    R / (1 + a0) * (σ' + s'*a0*(σ' - s'*(z0-σ')/z0) / z0)
  where
    σ' = min σ z0              
    s' = min s z0
```

Where the saturated pool reward can be obtained by replacing `σ` by `z0`:

```hs
f_satured = p_apparent * optimalRewards (R, z0, a0) z0 s
```

Anviking: Sounds good, but maybe not necessary now. Causes minor cascading changes.

lib/core/src/Cardano/Pool/Performance.hs
Anviking: Tbh, I really like how `Double >= 0` reads.  But, yes, that seems decent. There are parse issues with `>= 0` and hlint.

So going with 2:

### 1. `Double >= 0`, `Double >= 1`
- Tbh, I do really like how this reads
- TypeOperators, Nats
- Looks like we only need the `>= 0` and `>= 1` cases, so it is more general than we need.
- Causes parse-issues with both stylish-haskell and hlint. Even with `hlint` built on `master`.

### 2. `NonNegative Double`, `Positive Double`
- Similar to 1, but less fancy

### 3. Enforce in each newtype: `Lovelace`, `TotalRewards`, ...
- We'd need *a lot of* new types. Seems bad to me.
- `TotalRewards` are in fact in Lovelace, but would have to be separate, unrelated newtypes.

lib/core/src/Cardano/Pool/Ranking.hs
```diff
@@ -0,0 +1,188 @@
...
+--------------------------------------------------------------------------------
+
+data EpochConstants = EpochConstants
+    { leaderStakeInfluence :: Double >= 0
+      -- ^ a_0
+    , punishmentHarshness :: Double >= 0
+      -- ^ \gamma
+    , optimalNumberOfPools :: Int >= 1
+      -- ^ k
+    , totalRewards :: Lovelace >= 1
```
Anviking: Hm… not sure why I forced R >= 1 instead if 0 🤔

lib/core/src/Cardano/Pool/Metrics.hs
Anviking: Don't think it is there at all in the latest version of the spec. And not needed in this PR anyway. Removed.

lib/core/src/Cardano/Pool/Performance.hs
Anviking: Put it in actual haddock table syntax:

Skärmavbild 2020-01-16 kl  18 07 43

I realised one could argue this table to be superfluous as we have these definitions (mostly I believe) in the code itself, but might be nice to have 🤷‍♂

lib/core/src/Cardano/Pool/Performance.hs
Anviking: >  I'd opt for sticking to the paper's terminology

That was my original intent. 

But I later found myself mixing up the identifiers and making severe mistakes. That's why I started introducing longer names.

It should now be a nice mixture of both, with the records having descriptive names, but the formulas using let-bindings to the short names matching the spec.

lib/core/src/Cardano/Pool/Metrics.hs
Anviking: Not needed in this PR. Will be tackled in #1276

lib/core/test/unit/Cardano/Pool/RankingSpec.hs
```diff
@@ -0,0 +1,226 @@
...
+                (property prop_desirabilityNonNegative)
+
+        describe "saturatedPoolRewards" $ do
+            it "a0 == 0 => saturatedPoolRewards == p*R*z0"
+                (property prop_saturatedPoolRewardsReduces)
+            it "saturatedPoolRewards >= 0"
+                (property prop_saturatedPoolRewardsNonNegative)
+
+        describe "mkRelativeStake" $ do
+            it "stake > R, R >= 1 => mkRelativeStake ∈ [0,1] (even with conversions)"
```
Anviking: I was afraid of rounding errors (not sure if rightfully so).

lib/core/src/Cardano/Pool/Ranking.hs
```diff
@@ -0,0 +1,206 @@
...
+        p = getNonNegative $ recentAvgPerformance pool
+        -- ^ technically \hat{p} in the spec
+    in
+        (p * _R) / (1 + a0)
+        * (z0 + ((min s z0) * a0))
+
+-- | Determines z0, i.e 1 / k
+saturatedPoolSize :: EpochConstants -> RelativeStakeOf "pool"
+saturatedPoolSize constants =
+    RelativeStake $ 1 / fromIntegral (getPositive $ optimalNumberOfPools constants)
```
KtorZ: Why is this of type `RelativeStakeOf` ? I thought the point of these newtypes was to provide insights on the semantic of the various numbers, `σ` stands for the relative stake of the pool and, although they have the same dimension, labelling both as "RelativeStakeOf" feels misleading. 

Aype that captures what the domain actually is would be preferable (which is what types for in the first place, representing a range of possible values): `Ratio 1 0 ` or `Percentage` ?

lib/core/src/Cardano/Pool/Ranking.hs
```diff
@@ -0,0 +1,206 @@
...
+
+--------------------------------------------------------------------------------
+-- Types
+--------------------------------------------------------------------------------
+
+data EpochConstants = EpochConstants
+    { leaderStakeInfluence :: NonNegative Double
+      -- ^ a_0
+    , optimalNumberOfPools :: Positive Int
+      -- ^ k
```
KtorZ, line 143: `k` is referred to as the _"desired number of pools"_, which is different from an _"optimal"_ number. Rather, we chose to fix `k` and then, determine what other parameters should be. But, chosing `k=10` or `k=100` or `k=1000` is rather arbitrary and simply witnesses of how decentralized we want the network to be.

lib/core/src/Cardano/Pool/Ranking.hs
```diff
@@ -0,0 +1,206 @@
...
+    , totalRewards :: Lovelace
+      -- ^ Total rewards in an epoch. "R" in the spec.
+    } deriving (Show, Eq, Generic)
+
+data Pool = Pool
+    { leaderStake :: RelativeStakeOf "pool leader"
```
KtorZ: A comment to link this to `s` is missing. Also, same remark about the choice of type regarding "RelativeStakeOf" applies here too.

```hs
-- s 
```

lib/core/src/Cardano/Pool/Ranking.hs
```diff
@@ -0,0 +1,206 @@
...
+    , cost :: Lovelace
+    , margin :: Margin
```
KtorZ: `margin` and `leaderStake` have so-to-speak the same dimension, so arguably they should have the same type.

lib/core/src/Cardano/Pool/Ranking.hs
```diff
@@ -0,0 +1,206 @@
...
+    , recentAvgPerformance :: NonNegative Double
+      -- ^ An already averaged performance-value
+    } deriving (Show, Eq, Generic)
+
+newtype Lovelace = Lovelace { getLovelace :: NonNegative Double }
```
KtorZ: Why are Lovelace represented as `Double` :thinking: ? These clearly are non-negative integral values?  Also, the extra `Lovelace` wrapping seems superfluous since we don't attach any properties to the fact that they are Lovelace apart from:

- The fact that they are non-negative
- The fact that they are integers

Which can be entirely captured with "NonNegative Integer", or, an existing type: "Natural".

lib/core/src/Cardano/Pool/Ranking.hs
```diff
@@ -0,0 +1,206 @@
...
+    total = getNonNegative $ getLovelace $ totalRewards constants
+
+newtype Positive a = Positive { getPositive :: a }
+    deriving (Generic, Eq, Show)
+    deriving newtype (Ord, Num)
+
+unsafeToPositive :: (Ord a, Show a, Num a) => a -> (Positive a)
+unsafeToPositive x
+    | x >= 0    = Positive x
+    | otherwise = error $ "unsafeToPositive: " ++ show x ++ " > 0 does not hold"
```
KtorZ: Note that `Positive a` in this module is only used for `Positive Int` which can be substituted with `Natural` from the `base` package. The `Positive` data type is therefore redundant.

lib/core/src/Cardano/Pool/Ranking.hs
```diff
@@ -0,0 +1,206 @@
...
+      -- ^ k
+    , totalRewards :: Lovelace
+      -- ^ Total rewards in an epoch. "R" in the spec.
+    } deriving (Show, Eq, Generic)
+
+data Pool = Pool
+    { leaderStake :: RelativeStakeOf "pool leader"
+    , cost :: Lovelace
+    , margin :: Margin
+    , recentAvgPerformance :: NonNegative Double
```
KtorZ, line 155: `NonNegative Double` would suppose that the performance could be greater than 1, which isn't the case. Incidentally `margin`, `leaderStake`  and `recentAvgPerformance` all have the same unit / type which should be equivalent to a fractional number between 0 and 1.

KtorZ (changes requested):
If using dedicated data-types for the various fields we should make sure to use relevant types to capture the right domain of values. 
I'll have a look at the properties and will comment on this bit in a moment.

lib/core/src/Cardano/Pool/Ranking.hs
Anviking: Sorry, I got the condition wrong here. It should be `x > 0`, which is not captured by `Natural`.

This is to prevent division-by-zero if anyone sets `k = 0`.

I should add tests for these `unsafeTo`-constructors if we intend to have them.

lib/core/src/Cardano/Pool/Ranking.hs
Anviking: I introduced `RelativeStakeOf` to protect against mixing up `\sigma` and `s`.

Although, at that time, most functions took only the pool parameters they needed. And it was easy to get the order wrong. Now both `desirability` and `saturatedPoolSize` take `Pool`, and the `RelativeStakeOf tag` tag might be less useful.

lib/core/src/Cardano/Pool/Ranking.hs
Anviking: >  performance could be greater than 1, which isn't the case

I cannot conclude this from the spec.

> Even when a pool’s true performance is 1, its actual rewards might be less than its optimal rewards, because of the randomness in re. Likewise, sometimes a pool’s actual rewards will be **higher than their optimal rewards**, if they are **lucky** in the leader election process.
>
> **These effects will balance each other**, so that a well performing pool will get their optimal rewards on average over multiple epochs.

and from the formulas I don't see how `\bar{p}` would be constrained to `[0,1]`:

Skärmavbild 2020-01-17 kl  11 48 04

I have another concern about clamping the **pre**-averaged performances, especially, with `min 1` , but that is separate to this.

Anviking: If there is a choice to be made whether to restrict to `[0,1]` or not it seems better to me to not make that choice here. Especially if the type of `recentAvgPerformance` is going to error when a value is out of bounds.

KtorZ: Fair enough :+1:

lib/core/test/unit/Cardano/Pool/RankingSpec.hs
```diff
@@ -0,0 +1,254 @@
...
+
+prop_oversaturationDoesNotInfluenceDesirability
+    :: Pool
+    -> EpochConstants
+    -> Property
+prop_oversaturationDoesNotInfluenceDesirability pool constants =
+    forAll arbitrary $ \v1 v2 -> do
+        let (higherStake, lowerStake) = sortTuple (v1, v2)
+        let z0 = unRelativeStake $ saturatedPoolSize constants
+        higherStake > z0 ==>
```
KtorZ: Side note but, this pre-condition should really be `lowerStake > z0`, shouldn't it?

lib/core/test/unit/Cardano/Pool/RankingSpec.hs
```diff
@@ -0,0 +1,254 @@
...
+        _R = R.getNonNegative $ getLovelace $ totalRewards constants
+        p  = R.getNonNegative $ recentAvgPerformance pool
+    in
+        saturatedPoolRewards constants pool === p*_R*z0
+
+prop_oversaturationDoesNotInfluenceDesirability
+    :: Pool
+    -> EpochConstants
+    -> Property
+prop_oversaturationDoesNotInfluenceDesirability pool constants =
```
KtorZ: I am not entirely convinced by what this property tests. Of course, the pool's stake doesn't influence the desirability because it's not even a parameter of the equation!

lib/core/test/unit/Cardano/Pool/RankingSpec.hs
```diff
@@ -0,0 +1,254 @@
...
+        , KnownSymbol field
+        )
+    => EffectOnDesirability
+        -- ^ Expected effect on the desirability
+    -> Pool
+       -- ^ The base pool which is later modified
+    -> EpochConstants
+       -- ^ Constants used
+    -> Property
+allElseEqualProperty expectedEffect pool constants =
```
KtorZ, line 182: Good one :+1:

lib/core/test/unit/Cardano/Pool/RankingSpec.hs
```diff
@@ -0,0 +1,254 @@
...
+deriving via (NonNegative Double) instance (Arbitrary Lovelace)
+deriving via (NonNegative Double) instance (Arbitrary (R.NonNegative Double))
+deriving via (Positive Int) instance (Arbitrary (R.Positive Int))
+
+instance Arbitrary Margin where
+    arbitrary = unsafeMkMargin <$> choose (0, 1)
+    shrink = map unsafeMkMargin
+        . filter (\a -> a >= 0 && a <= 1)
+        . map getMargin
+        . shrink
```
KtorZ: The arbitrary instances for `Margin` and `RelativeStakeOf` are identical which re-enforce the argument that these should really be of the same type.

lib/core/test/unit/Cardano/Pool/RankingSpec.hs
```diff
@@ -0,0 +1,254 @@
...
+        describe "desirability" $ do
+            it "desirability >= 0"
+                $ property prop_desirabilityNonNegative
+
+        describe "saturatedPoolRewards" $ do
+            it "a0 == 0 => saturatedPoolRewards == p*R*z0"
+                $ property prop_saturatedPoolRewardsReduces
+
+            it "saturatedPoolRewards >= 0"
+                $ property prop_saturatedPoolRewardsNonNegative
```
KtorZ, line 98: :+1:

lib/core/test/unit/Cardano/Pool/RankingSpec.hs
Anviking: If `lowerStake > z0` and `higherStake >= lowerStake`, then both pools would be over-saturated.

It's enough to have the *one* pool be oversaturated, for it to be "more oversaturated" than the other one.

(So I guess I'm defining `oversaturation = max 0 (sigma - z0)`)

lib/core/test/unit/Cardano/Pool/RankingSpec.hs
Anviking: I think it's easy to guess that `desirability` should take oversaturation into account. I wanted to make it obvious that it doesn't.

If we ever add `stake` as a field to `Pool` it will not be obvious from the type-signatures alone.

lib/core/test/unit/Cardano/Pool/RankingSpec.hs
Anviking:
```Haskell
        , HasField' field Pool a
        , KnownSymbol field
```

Nice simplification 👍 , suspected that was theoretically possible

lib/core/src/Cardano/Pool/Ranking.hs
```diff
@@ -0,0 +1,198 @@
...
+      -- ^ c
+    , margin :: Ratio
+      -- ^ m
+    , recentAvgPerformance :: NonNegative Double
+      -- ^ \hat{p}, an already averaged performance-value.
+      --
+      -- Should mostly be in the range [0, 1], but lucky pools may exceed 1.
+    } deriving (Show, Eq, Generic)
+
+newtype Lovelace = Lovelace { getLovelace :: Word64 }
```
KtorZ, line 162: Minor note: I find "Natural" conceptually better than `Word64` (in this context at least, but generally speaking) to describe unbounded positive integers. `Word64` is something we would find at boundaries of the system like databases or types in network serializations. Yet, conceptually, the _right_ type to describe positive integers is `Natural`.

lib/core/src/Cardano/Pool/Ranking.hs
```diff
@@ -0,0 +1,198 @@
...
+    deriving newtype Ord
+
+unsafeToRatio :: Double -> Ratio
+unsafeToRatio x
+    | x >= 0 && x <= 1  = Ratio x
+    | otherwise         = error $ "unsafeToRatio: " ++ show x
+                          ++ "not in range [0, 1]"
+
+unsafeMkRelativeStake :: Lovelace -> EpochConstants -> Ratio
+unsafeMkRelativeStake (Lovelace stake) constants =
```
KtorZ: Minor note: I find it inconsistent to have within the same module names like `unsafeToXXX` vs `unsafeMkXXX` while the constructors fulfill the same goal. Having said that, that's probably my OCD being triggered but there's nothing intrinsically wrong with either of those names.

lib/core/test/unit/Cardano/Pool/RankingSpec.hs
KtorZ: But.. the `stake` of the pool doesn't even enter into account in the computation of the desirability. So, the property reduces to: `forall d. d === d ~ True`. This property is a tautology. Let's remove it and feel free to drop a comment in the source module about the effect of over-saturation.

KtorZ (approved):

lib/core/src/Cardano/Pool/Ranking.hs
Anviking: >  I find it inconsistent to have within the same module names like

Yes — I agree, which one do you prefer?

KtorZ: My own preference would be `unsafeMk` but I have no valid rationale to justify this.

Anviking: Done.

Anviking:
@KtorZ FYI, I intended the pools to be printed sorted by desirability on failure, but they were not. I fixed, and had to ~revert some of your style here. effecd981618e675ec09903b827b82fdf06de780

lib/core/test/unit/Cardano/Pool/RankingSpec.hs
Anviking: > But.. the stake of the pool doesn't even enter into account in the computation of the desirability. So, the property reduces to: forall d. d === d ~ True. This property is a tautology. Let's remove it and feel free to drop a comment in the source module about the effect of over-saturation.

I think my response above applies here. Yes it is a tautology, but the moment we add a `stake` field (as you indirectly suggested in a previous comment, or when we implement non-myopic member rewards), it will not longer be.

Anviking: Although, if you insist, let me know…

There is already a comment on `desirability`.

lib/core/test/unit/Cardano/Pool/RankingSpec.hs
```diff
@@ -0,0 +1,284 @@
...
+
+-- TODO: We should ideally not export the NonNegative and Positive constructors,
+-- in which case we wouldn't be able to use DerivingVia.
+deriving via Word64 instance (Arbitrary Lovelace)
+deriving via (NonNegative Double) instance (Arbitrary (R.NonNegative Double))
+deriving via (Positive Int) instance (Arbitrary (R.Positive Int))
+
+instance Arbitrary Ratio where
+    arbitrary = unsafeMkRatio <$> choose (0, 1)
+    shrink _ = [] -- TODO: Implement a shrinker that works.
```
Anviking, line 242: It was the `Ratio` shrinker that was faulty. I removed it, and re-added `genericShrink` for `Pool`.

lib/core/test/unit/Cardano/Pool/RankingSpec.hs
KtorZ: > the moment we add a stake field (as you indirectly suggested in a previous comment, or when we implement non-myopic member rewards), it will not longer be.

Then, we'll add the property at this moment :+1: 
I insist :)

lib/core/src/Cardano/Pool/Ranking.hs
Anviking: I just realised that `unsafeMkRelativeStake` is fatally incorrect. It calculates `stake / totalRewards`, and not `stake / totalStake`.

Issue #1251, opened 2020-01-07
Missing database migration for `active_slot_coeff`


KtorZ:
# Context
 
 

We "recently" added the `active_slot_coeff` blockchain parameters to the checkpoint data. Yet, for database existing before the release, we have missed a migration.. 

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | v2019.12.16+  |
| Platform     | All           |
| Installation | N/A             |

# Steps to Reproduce

1. Create a wallet with a version prior to v2019.12.16
2. Stop the server
3. Start a server running on v2019.12.16+ 

## Expected behavior

1. Wallets can be restored successfully

## Actual behavior
 

1. The worker exits because of a failing migration. 

```
SQLite3 returned ErrorConstraint while attempting to perform step: NOT NULL constraint failed: checkpoint_backup.active_slot_coeff
```

---

# Resolution 

- Run manual migration to migrate databases with default values for missing field ==> #1266 

---

# QA 

- Correct behavior is visible from the logs:

```
[...]
[cardano-wallet.network:Info:22] [2020-01-18 17:01:00.96 UTC] Waiting for Jörmungandr to be ready on tcp/35027
[cardano-wallet.application:Info:22] [2020-01-18 17:01:00.97 UTC] Found existing wallet: c8416497faead398770a1ae14eb31a41683c8f67
[cardano-wallet.wallet-db:Notice:28] [2020-01-18 17:01:00.97 UTC] checkpoint table does not contain required field 'active_slot_coeff'. Adding this field with a default value of 1.0.
[...]
[cardano-wallet.wallet-engine:Info:28] [2020-01-18 17:01:00.99 UTC] c8416497: Applying blocks [1158224.6 ... 1158226.8]
```

How to reproduce the above:

1. Download v2019-12-13 
2. Launch the server using integration config params
3. Create a wallet
4. Turn it off, and launch the server using latest `master`
5. Observe

On subsequent runs, nothing is logged, unless the severity is at least `DEBUG`

```
$ cardano-wallet launch ... --trace-wallet-db debug --log-level debug -- ... 
[...]
[cardano-wallet.wallet-db:Debug:28] [2020-01-18 17:02:06.23 UTC] checkpoint table already contains required field 'active_slot_coeff'.
[cardano-wallet.wallet-db:Debug:28] [2020-01-18 17:02:06.23 UTC] Running database action - Start
[...]

PR #1261, opened 2020-01-13
widen 'isMigrationError' to catch constraint errors

KtorZ:
# Issue Number

#1251 

# Overview

- c68cc47dd148911f7b84ca475a56fb898cb841d7
  widen 'isMigrationError' to catch constraint errors
  On start-up, we run persistent migrations and in case of migration issues, we
execute some particular handling (e.g. the pool db resets itself). Yet, we
failed at catching migration issues coming from constraints introduced later
and that can't be satisfied on already existing data. This commit fixes it.

#### Before

```
[cardano-wallet.network:Info:22] Jörmungandr is ready.
cardano-wallet-jormungandr: SQLite3 returned ErrorConstraint while attempting to perform step: NOT NULL constraint failed: pool_owner_backup.pool_owner_index
(exiting with an error code)
```

#### After

```
[cardano-wallet.network:Info:22] Jörmungandr is ready.
[cardano-wallet.stake-pool-db:Error:22]  Failed to migrate the database: : NOT NULL constraint failed: pool_owner_backup.pool_owner_index
[cardano-wallet.stake-pool-db:Notice:22] Non backward compatible database found. Removing old database and re-creating it from scratch. Ignore the previous error.
[cardano-wallet.stake-pool-db:Notice:22] 1 migrations were applied to the database.
(... and continuing as normal)
```

- 25fa3182322ae508a6397c27000012a02b6c803c
  rename 'IsMigration...' to 'MatchMigration...'
  
- 9d6143765eba02c53137b6ca42606b4970813579
  flatten result after tryJust to a single 'Either'

# Comments

### DIFF
--- a/lib/core/src/Cardano/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/DB/Sqlite.hs
@@ -45,7 +45,7 @@ import Control.Concurrent.MVar
 import Control.Exception
     ( Exception, bracket_, tryJust )
 import Control.Monad
-    ( mapM_ )
+    ( join, mapM_ )
 import Control.Monad.Catch
     ( Handler (..), MonadCatch (..), handleIf, handleJust )
 import Control.Monad.Logger
@@ -191,6 +191,8 @@ startSqliteBackend migrateAll trace fp = do
     migrations <- runQuery (runMigrationQuiet migrateAll)
         & tryJust (matchMigrationError @PersistException)
         & tryJust (matchMigrationError @SqliteException)
+        & fmap join
+        :: IO (Either MigrationError [Text])
     traceWith trace $ MsgMigrations (fmap length migrations)
     let ctx = SqliteContext backend runQuery fp trace
     case migrations of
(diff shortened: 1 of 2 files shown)


lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -195,14 +199,23 @@ startSqliteBackend migrateAll trace fp = do
...
-isMigrationError :: PersistException -> Maybe MigrationError
-isMigrationError e
-    | mark `isInfixOf` msg = Just $ MigrationError $ T.pack msg
-    | otherwise = Nothing
-  where
-    msg = show e
-    mark = "Database migration: manual intervention required."
+class Exception e => IsMigrationError e where
+    -- | Exception predicate for migration errors.
+    isMigrationError :: e -> Maybe MigrationError
```
jonathanknowles: I think the function name `isMigrationError` is a bit misleading, as the prefix `is` strongly implies a Boolean result, when what we actually want is to _match and extract_ an error of a particular type (if it exists).

**Suggestion**: rename this to something like `catchMigrationError` or `matchMigrationError`.

jonathanknowles (approved):
I've tested this going from `v2019-12-13` (f3576be15de4976e753dea212aa5d44c334bd204) to the current `master` (10b5512222e09ef788dcea82628113add4291460), and it does appear to solve the problem of recreating the stake pool DB:

```hs
[cardano-wallet.stake-pool-db:Error:22] [2020-01-14 02:52:08.72 UTC] Failed to migrate the database: : NOT NULL constraint failed: pool_owner_backup.pool_owner_index
[cardano-wallet.stake-pool-db:Notice:22] [2020-01-14 02:52:08.72 UTC] Non backward compatible database found. Removing old database and re-creating it from scratch. Ignore the previous error.
```

I only have a small suggestion regarding naming of the function `isMigrationError`. (See comments.)

PR #1266, opened 2020-01-15
Add a manual DB migration step for `active_slot_coeff`.

jonathanknowles:
# Issue Number

#1251 

# Overview

This PR:

- [x] Adds the ability to perform manual DB migration actions after connecting to a SQLite database.
- [x] Uses this ability to detect whether or not the `active_slot_coeff` field is present in the `checkpoint` table.
- [x] Adds the `active_slot_coeff` field manually, if it was not present already, using a default value that is passed down from the network layer.
- [ ] Figure out the cause of repeated log entries of the form:
```hs
[cardano-wallet.wallet-engine:Info:29] [2020-01-16 09:05:14.78 UTC] e834cdb3: Try rolling back to 0.0
[cardano-wallet.wallet-engine:Info:29] [2020-01-16 09:05:14.78 UTC] e834cdb3: Rolled back to 0.0
```

# Comments

### DIFF
--- a/lib/jormungandr/src/Cardano/Wallet/Jormungandr.hs
+++ b/lib/jormungandr/src/Cardano/Wallet/Jormungandr.hs
@@ -264,7 +264,10 @@ serveWallet
         let (block0, bp) = staticBlockchainParameters nl
         let tracer = contramap MsgDatabaseStartup applicationTracer
         wallets <- maybe (pure []) (Sqlite.findDatabases @k tracer) databaseDir
-        db <- Sqlite.newDBFactory walletDbTracer databaseDir
+        db <- Sqlite.newDBFactory
+            walletDbTracer
+            (getActiveSlotCoefficient bp)
+            databaseDir
         Server.newApiLayer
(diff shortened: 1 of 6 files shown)


lib/jormungandr/src/Cardano/Wallet/Jormungandr.hs
```diff
@@ -266,7 +266,10 @@ serveWallet
         let (block0, bp) = staticBlockchainParameters nl
         let tracer = contramap MsgDatabaseStartup applicationTracer
         wallets <- maybe (pure []) (Sqlite.findDatabases @k tracer) databaseDir
-        db <- Sqlite.newDBFactory walletDbTracer databaseDir
+        db <- Sqlite.newDBFactory
+            walletDbTracer
+            (DefaultFieldValues $ getActiveSlotCoefficient bp)
```
KtorZ, line 269: :+1:

KtorZ (approved):
Nice work. I've slightly revised the logs and the migration creation to avoid duplicated information (table's name, field's name and field's type) and derive that from the actually corresponding Haskell type. 
Also distinguished between migration needed and migration not needed logs so that we can assign them different severities!

PR #1268, opened 2020-01-15
endpoint: force wallet resync

KtorZ:
# Issue Number

#1146 

# Overview

- 92c2aa01289c6274a94fa07d55eac6181e35ecb5
  extend swagger specification with 'forceResync'
  As an extreme measure to force re-sync the wallet

- 206affcab73d87d1e7c0c46c433eb2ea9494993c
  extend API with 'forceResync' operation
  - Added a 'forceResync' link
- Extended the servant Client
- Provide an initial straightforward implementation re-using 'rollbackBlocks'

- 3d9bc77220fa58341cfe803d4720db04d0422e92
  implement 'forceResync' handler properly
  Taking care of concurrent workers possibly acting on the database
at the same time as well as restarting the restoration worker at
the right point

- 4b06845b9fdacccf268302aa443fbf4cfb812366
  add integration test for the 'forceResync' endpoint
  Doesn't pass currently because of the in-memory sqlite implementation.
It works like a charm with the file database though:

```
(Right before making the request, observe the thread id #142)
[cardano-wallet.wallet-engine:Info:142] [2020-01-15 10:38:04.54 UTC] bdb212cc: In sync with the node.

(Upon receiving the request)
[cardano-wallet.api-server:Info:526] [2020-01-15 10:38:04.97 UTC] [RequestId 10] [DELETE] /v2/wallets/bdb212ccd1d556b62a59f4a4a1de0cff29f61ac1/utxos

(Turning off the worker, rolling back, and switching the worker back on)
[cardano-wallet.wallet-engine:Info:142] [2020-01-15 10:38:04.97 UTC] bdb212cc: Worker has exited: main action is over.
[cardano-wallet.wallet-engine:Info:526] [2020-01-15 10:38:04.98 UTC] bdb212cc: Try rolling back to 0.0
[cardano-wallet.wallet-engine:Info:526] [2020-01-15 10:38:05.00 UTC] bdb212cc: Rolled back to 0.0

(Responding to the request)
[cardano-wallet.api-server:Info:526] [2020-01-15 10:38:05.01 UTC] [RequestId 10] 204 No Content in 0.044038928s

(And later, the worker restarted, applying blocks from the start, new thread id #528).
[cardano-wallet.wallet-engine:Info:528] [2020-01-15 10:38:05.03 UTC] bdb212cc: Applying blocks [1136063.1 ... 1144130.1]
[cardano-wallet.wallet-engine:Info:528] [2020-01-15 10:38:05.03 UTC] bdb212cc: Creating checkpoint at feb1b2c2-[1136063.1#1]
[cardano-wallet.wallet-engine:Info:528] [2020-01-15 10:38:05.04 UTC] bdb212cc: Creating checkpoint at 8920cbb4-[1136063.2#2]
[cardano-wallet.wallet-engine:Info:528] [2020-01-15 10:38:05.04 UTC] bdb212cc: Creating checkpoint at 73a21e9f-[1136063.3#3]

[...]

(Eventually, reaches the tip)
[cardano-wallet.wallet-engine:Info:528] [2020-01-15 10:38:05.18 UTC] bdb212cc: MyWallet, created at 2020-01-15 10:35:28.24351865 UTC, not delegating
[cardano-wallet.wallet-engine:Info:528] [2020-01-15 10:38:05.18 UTC] bdb212cc: syncProgress: restored
[cardano-wallet.wallet-engine:Info:528] [2020-01-15 10:38:05.18 UTC] bdb212cc: discovered 0 new transaction(s)
[cardano-wallet.wallet-engine:Info:528] [2020-01-15 10:38:05.18 UTC] bdb212cc: local tip: 8b9aba60-[1144130.1#100]
```

- d196fd1be0d19021ca500e53ba32881765c4ac55
  preserve in-memory database states between calls
  For the in-memory Sqlite database, we do actually preserve the database
after the 'action' is done. This allows for calling 'withDatabase'
several times within the same execution and get back the same database.
The memory is only cleaned up when calling 'removeDatabase', to mimic
the way the file database works!

Without this, code like:

```hs
  Registry.remove re wid
  registerWorker ctx wid
```

would lead to weird behavior where the wallet `wid` which existed when
the worker was stopped didn't exist anymore when register the worker
again. This was because, killing the worker closes the connection
to the database and re-opening the "same" connection on the in-memory
database actually yield a completely different memory representation!

- 423296b5e3a62db258a269027f5672b703f0a09c
  revise 'expectResponseCode' to give more details on failure
  It was kinda hard to understand what was going on from the error message.
This is now slightly better and at least, allow some reasonning

- 5eae1fa032ec1c62935236bd04b903eaccc17829
  re-generate nix machinery
  
- fab6e17d2c88a073a1a883a80491318537a0ca83
  fix database closing error handling
  - Added some additional log messages
- Added an extra exception handler
- Swapped worker removal with database closing
- Reviewed how the in-memory database got cleaned up to avoid race condition

- 9babcc2c202a51d300520f0618d50c3871f476c3
  use on-disk database in integration tests

# Comments

Screenshot from 2020-01-15 16-50-37

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
@@ -1,4 +1,6 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedLabels #-}
 {-# LANGUAGE QuasiQuotes #-}
@@ -13,6 +15,8 @@ module Test.Integration.Scenario.API.Wallets
 
 import Prelude
 
+import Cardano.Wallet.Api.Link
+    ( Discriminate )
 import Cardano.Wallet.Api.Types
     ( AddressAmount (..)
     , ApiCoinSelection
@@ -29,13 +33,20 @@ import Cardano.Wallet.Primitive.Mnemonic
 import Cardano.Wallet.Primitive.Types
     ( SyncProgress (..)
     , WalletDelegation (..)
+    , WalletId
     , walletNameMaxLength
     , walletNameMinLength
     )
 import Control.Monad
     ( forM_ )
+import Data.Aeson
+    ( FromJSON )
 import Data.Generics.Internal.VL.Lens
     ( view, (^.) )
+import Data.Generics.Product.Fields
+    ( HasField' )
+import Data.Generics.Product.Typed
+    ( HasType )
 import Data.List.NonEmpty
     ( NonEmpty ((:|)) )
 import Data.Quantity
@@ -44,6 +55,8 @@ import Data.Text
     ( Text )
 import Data.Text.Class
     ( toText )
+import GHC.Generics
+    ( Generic )
 import Numeric.Natural
     ( Natural )
 import Test.Hspec
@@ -59,8 +72,10 @@ import Test.Integration.Framework.DSL
     , coinSelectionInputs
     , coinSelectionOutputs
     , delegation
+    , emptyIcarusWallet
     , emptyRandomWallet
     , emptyWallet
+    , eventually
     , expectErrorMessage
     , expectEventually
     , expectFieldEqual
@@ -93,6 +108,7 @@ import Test.Integration.Framework.TestData
     , chineseMnemonics18
     , chineseMnemonics9
     , errMsg400ParseError
+    , errMsg403RejectedTip
     , errMsg403WrongPass
     , errMsg404NoEndpoint
     , errMsg404NoWallet
@@ -1779,3 +1795,101 @@ spec = do
         ru <- request @ApiWallet ctx ("GET", endpoint) Default newName
         expectResponseCode @IO HTTP.status404 ru
         expectErrorMessage (errMsg404NoWallet wid) ru
+
+    describe "WALLETS_RESYNC_01" $ do
+        scenarioWalletResync01_happyPath @'Shelley emptyWallet
+        scenarioWalletResync01_happyPath @'Byron emptyRandomWallet
+        scenarioWalletResync01_happyPath @'Byron emptyIcarusWallet
+
+    describe "WALLETS_RESYNC_02" $ do
+        scenarioWalletResync02_notGenesis @'Shelley emptyWallet
+        scenarioWalletResync02_notGenesis @'Byron emptyRandomWallet
+        scenarioWalletResync02_notGenesis @'Byron emptyIcarusWallet
+
+    describe "WALLETS_RESYNC_03" $ do
+        scenarioWalletResync03_invalidPayload @'Shelley emptyWallet
+        scenarioWalletResync03_invalidPayload @'Byron emptyRandomWallet
+        scenarioWalletResync03_invalidPayload @'Byron emptyIcarusWallet
+
+
+-- force resync eventually get us back to the same point
+scenarioWalletResync01_happyPath
+    :: forall style t n wallet.
+        ( n ~ 'Testnet
+        , Discriminate style
+        , HasType (ApiT WalletId) wallet
+        , HasField' "state" wallet (ApiT SyncProgress)
+        , FromJSON wallet
+        , Generic wallet
+        , Show wallet
+        )
+    => (Context t -> IO wallet)
+    -> SpecWith (Context t)
+scenarioWalletResync01_happyPath fixture = it
+  "force resync eventually get us back to the same point" $ \ctx -> do
+    w <- fixture ctx
+
+    -- 1. Wait for wallet to be synced
+    eventually $ do
+        v <- request @wallet ctx (Link.getWallet @style w) Default Empty
+        verify v [ expectFieldSatisfy @IO #state (== (ApiT Ready)) ]
+
+    -- 2. Force a resync
+    let payload = Json [json|{ "epoch_number": 0, "slot_number": 0 }|]
+    r <- request @wallet ctx (Link.forceResyncWallet @style w) Default payload
+    verify r [ expectResponseCode @IO HTTP.status204 ]
+
+    -- 3. The wallet eventually re-sync
+    eventually $ do
+        v <- request @wallet ctx (Link.getWallet @style w) Default Empty
+        verify v [ expectFieldSatisfy @IO #state (== (ApiT Ready)) ]
+
+-- force resync eventually get us back to the same point
+scenarioWalletResync02_notGenesis
+    :: forall style t n wallet.
+        ( n ~ 'Testnet
+        , Discriminate style
+        , HasType (ApiT WalletId) wallet
+        , HasField' "state" wallet (ApiT SyncProgress)
+        , FromJSON wallet
+        , Generic wallet
+        , Show wallet
+        )
+    => (Context t -> IO wallet)
+    -> SpecWith (Context t)
+scenarioWalletResync02_notGenesis fixture = it
+  "given point is not genesis (i.e. (0, 0))" $ \ctx -> do
+    w <- fixture ctx
+
+    -- 1. Force a resync on an invalid point (/= from genesis)
+    let payload = Json [json|{ "epoch_number": 14, "slot_number": 42 }|]
+    r <- request @wallet ctx (Link.forceResyncWallet @style w) Default payload
+    verify r
+        [ expectResponseCode @IO HTTP.status403
+        , expectErrorMessage errMsg403RejectedTip
+        ]
+
+-- force resync eventually get us back to the same point
+scenarioWalletResync03_invalidPayload
+    :: forall style t n wallet.
+        ( n ~ 'Testnet
+        , Discriminate style
+        , HasType (ApiT WalletId) wallet
+        , HasField' "state" wallet (ApiT SyncProgress)
+        , FromJSON wallet
+        , Generic wallet
+        , Show wallet
+        )
+    => (Context t -> IO wallet)
+    -> SpecWith (Context t)
+scenarioWalletResync03_invalidPayload fixture = it
+  "given payload is invalid (camelCase)" $ \ctx -> do
+    w <- fixture ctx
+
+    -- 1. Force a resync using an invalid payload
+    let payload = Json [json|{ "epochNumber": 0, "slot_number": 0 }|]
+    r <- request @wallet ctx (Link.forceResyncWallet @style w) Default payload
+    verify r
+        [ expectResponseCode @IO HTTP.status400
+        , expectErrorMessage "key 'epoch_number' not present"
+        ]
--- a/lib/core/src/Cardano/Wallet/Api.hs
+++ b/lib/core/src/Cardano/Wallet/Api.hs
@@ -24,6 +24,7 @@ module Cardano.Wallet.Api
         , PutWallet
         , PutWalletPassphrase
         , GetUTxOsStatistics
+        , ForceResyncWallet
 
     , Addresses
         , ListAddresses
@@ -49,6 +50,7 @@ module Cardano.Wallet.Api
         , GetByronWallet
         , ListByronWallets
         , PostByronWallet
+        , ForceResyncByronWallet
 
     , ByronTransactions
         , ListByronTransactions
@@ -89,6 +91,7 @@ import Cardano.Wallet.Api.Types
     , ApiCoinSelection
     , ApiFee
     , ApiNetworkInformation
+    , ApiNetworkTip
     , ApiSelectCoinsData
     , ApiStakePool
     , ApiT
@@ -192,6 +195,7 @@ type Wallets =
     :<|> PutWallet
     :<|> PutWalletPassphrase
     :<|> GetUTxOsStatistics
+    :<|> ForceResyncWallet
 
 -- | https://input-output-hk.github.io/cardano-wallet/api/#operation/deleteWallet
 type DeleteWallet = "wallets"
@@ -231,6 +235,13 @@ type GetUTxOsStatistics = "wallets"
     :> "statistics"
     :> "utxos"
     :> Get '[JSON] ApiUtxoStatistics
+
+-- | https://input-output-hk.github.io/cardano-wallet/api/#operation/forceResync
+type ForceResyncWallet = "wallets"
+    :> Capture "walletId" (ApiT WalletId)
+    :> "tip"
+    :> ReqBody '[JSON] ApiNetworkTip
+    :> PutNoContent '[Any] NoContent
 
 {-------------------------------------------------------------------------------
                                   Addresses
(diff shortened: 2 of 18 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
```diff
@@ -1779,3 +1794,39 @@ spec = do
         ru <- request @ApiWallet ctx ("GET", endpoint) Default newName
         expectResponseCode @IO HTTP.status404 ru
         expectErrorMessage (errMsg404NoWallet wid) ru
+
+    describe "WALLETS_RESYNC_01 - \
+        \ force resync eventually get us back to the same point" $ do
+        scenarioWalletResync01 @'Shelley emptyWallet
+        scenarioWalletResync01 @'Byron emptyRandomWallet
+        scenarioWalletResync01 @'Byron emptyIcarusWallet
```
KtorZ: @piotr-iohk this is an example of how I think we could re-structure a bit the integration scenarios. Key points are:

- named function, that can potentially be re-used in other scenarios (imagine for instance 'scenarioWalletRestore' or 'scenarioMakeTransaction' instead of re-writing the same things over and over).

- The scenario itself is generic in terms of wallet type, so that it can easily be replayed with any type of wallet we have.

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
```diff
@@ -1779,3 +1794,39 @@ spec = do
...
+
+    -- 2. Force a resync
+    r <- request @wallet ctx
+        (Link.forceResyncWallet @style w) Default Empty
+    verify r [ expectResponseCode @IO HTTP.status204 ]
+
+    -- 3. The wallet eventually re-sync
+    eventually $ do
+        v <- request @wallet ctx (Link.getWallet @style w) Default Empty
+        verify v [ expectFieldSatisfy @IO #state (== (ApiT Ready)) ]
```
paweljakubas, line 1845: we are fine here to check only `ApiWallet`'s `state` field, right? If there is no point 2 we are obviously satisfying point 3. It would be nice (although I do not know if it is easy) to show in point 2 upon forcing that `state` is `Syncing ...` so something happens as a result of force resync.

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -232,6 +235,12 @@ type GetUTxOsStatistics = "wallets"
     :> "utxos"
     :> Get '[JSON] ApiUtxoStatistics
 
+-- | https://input-output-hk.github.io/cardano-wallet/api/#operation/forceResync
+type ForceResyncWallet = "wallets"
+    :> Capture "walletId" (ApiT WalletId)
+    :> "utxos"
```
paweljakubas: why this? :thinking:

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -383,6 +393,12 @@ type GetByronWallet = "byron-wallets"
 type ListByronWallets = "byron-wallets"
     :> Get '[JSON] [ApiByronWallet]
 
+-- | https://input-output-hk.github.io/cardano-wallet/api/#operation/forceResyncByron
+type ForceResyncByronWallet = "byron-wallets"
+    :> Capture "walletId" (ApiT WalletId)
+    :> "utxos"
```
paweljakubas: the same question as above . why not `forceResync` or just `resync` -  whatever ...

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -877,6 +885,27 @@ getUTxOsStatistics ctx (ApiT wid) = do
...
+    liftIO $ Registry.remove re wid
+    liftHandler (ExceptT safeRollback) `finally` liftIO (registerWorker ctx wid)
+    pure NoContent
+  where
+    re = ctx ^. workerRegistry @s @k
+    tr = ctx ^. logger
+    df = ctx ^. dbFactory @s @k
+    -- NOTE Safe because it happens without any worker running.
+    safeRollback = do
+        let tr' = Registry.transformTrace wid tr
```
paweljakubas, line 908: we will have log as coming from withLogging about request..I wonder if we do not want to say more here. although we will have also additional logs from `rollbackBlocks` below. But maybe removing Registry is worth logging

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -223,17 +223,30 @@ newDBFactory
 newDBFactory tr mDatabaseDir = do
     mvar <- newMVar mempty
     case mDatabaseDir of
+        -- NOTE
+        -- For the in-memory database, we do actually preserve the database
+        -- after the 'action' is done. This allows for calling 'withDatabase'
+        -- several times within the same execution and get back the same
+        -- database. The memory is only cleaned up when calling
+        -- 'removeDatabase', to mimic the way the file database works!
```
paweljakubas, line 231: :100:

paweljakubas (approved):
I left some comments/questions. 

What about **adding the test** that shows that when using force resync any incoming transaction will be erased whereas outgoing transactions will be put back as pending?

lib/core/src/Cardano/Wallet/DB.hs
```diff
@@ -57,20 +57,18 @@ import Data.Quantity
...
 data DBFactory m s k = DBFactory
-    { withDatabase :: WalletId -> (DBLayer m s k -> IO ()) -> IO ()
+    { withDatabase :: forall a. WalletId -> (DBLayer m s k -> IO a) -> IO a
         -- ^ Creates a new or use an existing database, maintaining an open
         -- connection so long as necessary
 
     , removeDatabase :: WalletId -> IO ()
         -- ^ Erase any trace of the database
-    } deriving (Generic)
+    }
```
paweljakubas, line 70:
```
```
not needed after this change

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
piotr-iohk: > It would be nice (although I do not know if it is easy) to show in point 2 upon forcing that state is Syncing ...

that would be great. +1

Also I think that it would be nice to assert for few other stuff (maybe as separate scenarios):
 - balance (total, available, rewards?)
 - delegation status (if it the wallet was delegating is should be delegating still)
 - addresses (if it makes sense ?)
 - transaction history (whatever it should be after resync)

lib/core/src/Cardano/Wallet/Api.hs
paweljakubas: After second thought - `DELETE /wallet/{walletID}/utxos` starts to make sense - remove the wallets utxos, so it implies resyncing... but was expecting that it would be `POST /wallet/{walletID}/resync` ....

specifications/api/swagger.yaml
```diff
@@ -1277,6 +1284,22 @@ paths:
...
+    delete:
+      operationId: forceResync
+      tags: ["Wallets"]
+      summary: Force Resync
+      description: |
+        <p align="right">status: <strong>stable</strong></p>
+
+        Force the wallet to rewind back to the genesis block and sync again with the chain.
+        Any incoming transaction will be erased whereas outgoing transactions will be put back
+        as pending. This action should only be taken as an extreme measure to resolve potential
```
piotr-iohk, line 1316: I don't understand this part. Is it for all the transactions, even if they are in ledger? 

Why I'm asking: when you delete a wallet and create it again you have your transaction history available... My understanding and expectation of the forceResync would be to be an equivalent of delete and create (but without the need of deleting and creating a wallet again), am I wrong? (I bet I am...)

lib/core/src/Cardano/Wallet/Api.hs
KtorZ: you mean why `utxos` specifically or, why an extra sub-level ?

paweljakubas: yeah, utxos...more thoughts on this below

lib/core/src/Cardano/Wallet/Api.hs
KtorZ: Well, it's more to stay close to a REST philosophy and intuition. In REST, endpoints are not "actions" or "commands", instead, they identify **resources**. Operations we can do on resources are defined by the HTTP verbs (create = POST, replace = PUT, read = GET etc ...). 

So, what would be the "resync" resource exactly :thinking: ? It doesn't really fit in the REST philosophy. Now, if we think of what the operation actually does: it removes all the UTxOs and rewind the wallet back to its original genesis block. To stick with the REST approach, I though of two possible ways:

- This one, using `DELETE /wallets/{walletId}/utxos`
- Or, thinking of it in terms of the slot: `PATCH /wallets/{walletId}/tip` with a body that contains a slot id (epoch no + slot no). 

I think the second approach is a little more intuitive, though, it suppose that the `tip` can be set to _anything_. The wallet backend might still enforce that the given tip argument is necessarily `(0, 0)` though. Perhaps that's a cleaner way of approaching this?

cc @piotr-iohk

lib/core/src/Cardano/Wallet/Api/Server.hs
KtorZ: logs look as follows:

```
[cardano-wallet.api-server:Info:526] [2020-01-15 10:38:04.97 UTC] [RequestId 10] [DELETE] /v2/wallets/bdb212ccd1d556b62a59f4a4a1de0cff29f61ac1/utxos
[cardano-wallet.wallet-engine:Info:142] [2020-01-15 10:38:04.97 UTC] bdb212cc: Worker has exited: main action is over.
[cardano-wallet.wallet-engine:Info:526] [2020-01-15 10:38:04.98 UTC] bdb212cc: Try rolling back to 0.0
[cardano-wallet.wallet-engine:Info:526] [2020-01-15 10:38:05.00 UTC] bdb212cc: Rolled back to 0.0
[cardano-wallet.api-server:Info:526] [2020-01-15 10:38:05.01 UTC] [RequestId 10] 204 No Content in 0.044038928s
[cardano-wallet.wallet-engine:Info:528] [2020-01-15 10:38:05.03 UTC] bdb212cc: Applying blocks [1136063.1 ... 1144130.1]
[cardano-wallet.wallet-engine:Info:528] [2020-01-15 10:38:05.03 UTC] bdb212cc: Creating checkpoint at feb1b2c2-[1136063.1#1]
[cardano-wallet.wallet-engine:Info:528] [2020-01-15 10:38:05.04 UTC] bdb212cc: Creating checkpoint at 8920cbb4-[1136063.2#2]
[cardano-wallet.wallet-engine:Info:528] [2020-01-15 10:38:05.04 UTC] bdb212cc: Creating checkpoint at 73a21e9f-[1136063.3#3]
[...]
```

I think it's quite sufficient here :thinking:

specifications/api/swagger.yaml
KtorZ: @piotr-iohk force-resync is not equivalent to deleting and re-creating the wallet. It's actually equivalent to (and is a) rollback to the genesis block. So, after this, some transaction will be put back in pending such that, users can still track them. 
The annoying part with deleting and re-creating your wallet is also that you completely loose track of your wallet history. If you attempted to made a payment, it won't show up unless visible in the ledger. But you may still have transaction floating around in mempools that don't show up in your wallet. Whereas, by using force resync here, you'll still see them.

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: indeed, logs looks good. `Worker has exited...` logs removing of registry. and as expected there are logs from rollbackBlocks

lib/core/src/Cardano/Wallet/Api.hs
paweljakubas: yeah,  resource `state` (instead of utxos) along with DELETE would be the most intuitive (for me). Indeed, it is good to stick to ```resources + VERB``` and hence to REST philosophy

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
KtorZ: I 100% agree with you guys and I actually thought of this. The main problem I have here is .... race-conditions. 

Technically speaking, right after force-resyncing a wallet, your tip should be at genesis. But, by the time you make the request to produce some assertion... your wallet might as well have restored itself :| ... So, it's kinda hard to come up with reliable assertions that still make it possible to assess the behavior, precisely because this endpoint is meant to be used in situations where something went wrong / wasn't supposed to happen. 

So here, the scenario is much more about exercising the endpoint and making sure that, nothing wrong really happen, whereas the real confirmation that it works come from the logging output.

piotr-iohk: I understand that race conditions are the preventing the addition of the "Syncing" assertion? But checking for the balance, delegation, addresses, transaction history should be doable right? (That would also add confidence that nothing wrong happened with the wallet after resync and it is still usable... perhaps making transaction after _resync_ could be also a test that's nice to have )

lib/core/src/Cardano/Wallet/Api.hs
piotr-iohk: We also have `GET /wallets/{walletId}/statistics/utxos` which now I think could be just `GET /wallets/{walletId}/utxos` since we have `DELETE /wallets/{walletId}/utxos`...

KtorZ: Okay, I've opted for the following in the end. 

```
PUT /wallets/{walletId}/tip
```

This requires a `{ "slotNumber": ..., "epochNumber:" ... }` as argument, but will reject anything that is different from the genesis point (0, 0). I think it's a bit clearer / more intuitive.

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
KtorZ: I've added a few tests but, nothing really satisfactory in terms of testing _states_. If you guys have ideas, I'll happily do this in another PR (merging as is).

lib/core/src/Cardano/Wallet/Api.hs
piotr-iohk: Why to expose `{ "slotNumber": ..., "epochNumber:" ... }` since one cannot put anything other than (0,0)?

specifications/api/swagger.yaml
piotr-iohk: > Any incoming transaction will be erased whereas outgoing transactions will be put back as pending.

I think this statement may be somewhat misleading. After reading it I thought that after doing force resync all my incoming txs will be erased and not visible at all and outgoing will become pending. And it will be a finite state. But if they are in the ledger they will be discovered again and visible again. Hence the first comment, cause it was a bit surprising to me.

KtorZ:
bors retry

KtorZ:
CI is stuck with "Check Cabal Reconfigure" for _I am not sure what the reason is_ but I don't think it's remotely related to the PR, so merging all-the-same :|

Issue #1256, opened 2020-01-10
"Something went wrong" when listing stake pools from inside a Docker container


bourkefloyd:
# Context
 
Maybe related to changes here https://github.com/input-output-hk/cardano-wallet/issues/1164

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | 2019.12.23 (git revision: 398565cc1a65553827834520c5a95827a7c48d41) |
| Platform     | Docker FROM ubuntu:18.04    |
| Installation | Source 0.8.5   |

# Steps to Reproduce
Setup jormungandr w/ jci per instructions and run w/ jormungandr cmd line as leader
Setup cardano-wallet per instructions and run w/ serve

NOTE: 8443 is jcli and 8299 is the node
tcp        0      0 0.0.0.0:8443            0.0.0.0:*               LISTEN      14/jormungandr      
tcp        0      0 0.0.0.0:8444            0.0.0.0:*               LISTEN      13/cardano-wallet   
tcp        8      0 0.0.0.0:8299            0.0.0.0:*               LISTEN      14/jormungandr 

1. Note the first request to /v2/stake-pools (CLI or rest api) returns the stakepools without the meta data. 
2. All following requests to /v2/stake-pools (CLI or rest api) fail with 'Something went wrong(base)'

If you delete the `/tmp/stake-pool-metadata-90e545fc4b1cdae6/registry.zip` cache that seems to be failing to deserialize, you can start at step 1 again, and the repo steps will start over.

## Expected behavior

stake-pools list WITH metadata on the second request to /v2/stake-pools (not crash/error)
https://input-output-hk.github.io/cardano-wallet/api/edge/#tag/Stake-Pools

## Actual behavior
 
 'Something went wrong(base)'

---

# Resolution 

- Cleanup incomplete metadata archive when failing to download ==> #1272 
- Adds SSL root certificates to the docker image so that metadata registry downloads work. ==> #1256

---

# QA 

- Added unit tests to capture behavior on broken cache files ==> https://github.com/input-output-hk/cardano-wallet/pull/1272/files#diff-926cc24c9fd58a975e70cd61762a5ddbR141-R160

- Added also docker documentation in wiki ==> https://github.com/input-output-hk/cardano-wallet/wiki/Docker

bourkefloyd:
Just tested with cardano-wallet `v2020-01-07` same issue

rvl:
Hi @chbfiv, thanks for reporting.

[...]

PR #1272, opened 2020-01-16
Fix failed metadata download error handling

rvl:
Relates to #1256.

# Overview

- Add more error handling, logging, clean up.
- Adds SSL root certificates to the docker image so that metadata registry downloads work.


lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -219,8 +229,10 @@ getStakePoolMetadata
 getStakePoolMetadata tr cfg poolOwners = do
     let tr' = contramap (RegistryLog (registryURL cfg) (cacheArchive cfg)) tr
     fetchStakePoolMetaZipCached tr' cfg >>= \case
-        Right f -> Right <$> getMetadataFromZip tr' f poolOwners
-        Left e -> pure $ Left e
+        Right f -> getMetadataFromZip tr' f poolOwners
+        Left e -> do
+            traceWith tr' (MsgDownloadError e)
```
KtorZ, line 234: :+1:

lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -251,26 +263,37 @@ fetchStakePoolMetaZip
...
-    fmap join $ tryJust fileExceptionHandler $
+    res <- flip onException cleanup $ fmap join $ tryJust fileExceptionHandler $
         withFile zipFileName WriteMode $ \zipFile -> tryJust handler $ do
             req <- parseUrlThrow url
             httpSink req $ \_response -> Conduit.mapM_ (BS.hPut zipFile)
             size <- hTell zipFile
             traceWith tr $ MsgDownloadComplete size
             pure zipFileName
+    when (isLeft res) cleanup
+    pure res
```
KtorZ, line 274: It seems redundant to catch exceptions at 3 different places here and, capture some of them at the value-level while others remains interruptions. 
What about using a more "uniform" approach and, get the "unknown exception" as part of the `FetchError` data-type?

KtorZ (approved):

lib/core/src/Cardano/Pool/Metadata.hs
rvl: Yeah I know what you mean. Multiple handlers, apparent mixing of checked and unchecked exceptions, two cleanup paths. I tried to come up with something simpler but -

I want to differentiate between the exception types so that the error messages can describe what it was doing to cause the exception.

As far as I can tell, the only normal exceptions we should get are IOException and HttpException. If we add an "unknown exception" constructor for the Left FetchError, that exception handler should not swallow async exceptions.

So we would still need both `onException cleanup` and `when (isLeft res) cleanup`. The first one for the async exceptions that are not swallowed.

rvl:
The buildkite agent was misconfigured.

bors r+

KtorZ:
"thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 24, kind: Other, message: "Too many open files" }', src/libcore/result.rs:1084:5" from Jörmungandr :thinking:

Issue #1281, opened 2020-01-17
Rolling back stake pools take ages


KtorZ:
# Context
 
 

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | v2020-12-23+  |
| Platform     | All           |
| Installation | N/A             |

# Steps to Reproduce

1. Create a standalone private blockchain (e.g. using the integration config).
2. Turn on the wallet server and wait for a few slots / epochs.
3. Turn the wallet off.
4. Change a setting in the genesis configuration and re-generate the genesis block (effectively creating another blockchain). 
5. Restart the server on the same state directory. 

## Expected behavior

1. In cases of fork, the stake pool monitor can quickly rollback to a known point and keep syncing normally.

## Actual behavior
 

1. The pool monitor rolls back roughly 10 by 10 until it finds a common ancestor. 

```
[cardano-wallet.stake-pool-monitor:Info:362] [2020-01-17 18:09:53.65 UTC] Monitoring stake pools. Currently at 351628d6-[96.3233#64202]
[cardano-wallet.stake-pool-monitor:Info:362] [2020-01-17 18:09:53.89 UTC] Monitoring stake pools. Currently at 04fc30db-[96.3073#64193]
[cardano-wallet.stake-pool-monitor:Info:362] [2020-01-17 18:09:54.66 UTC] Monitoring stake pools. Currently at 70bf9891-[96.2978#64184]
[cardano-wallet.stake-pool-monitor:Info:362] [2020-01-17 18:09:55.17 UTC] Monitoring stake pools. Currently at edbf30a9-[96.2861#64175]
[cardano-wallet.stake-pool-monitor:Info:362] [2020-01-17 18:09:55.68 UTC] Monitoring stake pools. Currently at 105f158c-[96.2785#64166]
[cardano-wallet.stake-pool-monitor:Info:362] [2020-01-17 18:09:56.19 UTC] Monitoring stake pools. Currently at 3c67a40b-[96.2681#64157]
```

---

# Resolution 

Why is this happening? Because:

a) The stake pool monitoring doesn't clean up old checkpoints. 
b) In case where the chain follower is unable to find a suitable intersection, it'll instrument the client to enter in a "recovery" mode, by first, asking it to rollback to the oldest known checkpoint in the current window. If there's no older checkpoint, it recover from genesis.

[...]

PR #1285, opened 2020-01-20
Make stake pools engine rollback faster.

KtorZ:
# Issue Number

#1281 

# Overview

- fe138cca759b973c2b67c0b23cf5982e7d3ff4f3
  Log roll back attempts in pools monitoring
    When rolling back, we currently only log monitoring restarts, which
  can be somewhat confusing when done several times in a row:

  [pools-engine] Monitoring stake pools. Currently at f4260343-[1158372.5#220]
  [pools-engine] Monitoring stake pools. Currently at 20bd31a4-[1158372.1#216]
  [pools-engine] Monitoring stake pools. Currently at 437a7400-[1158371.6#212]
  [pools-engine] Monitoring stake pools. Currently at 5c874264-[1158371.2#208]

  Although this is typical of rollbacks, it's better to mention it
  explicitely to better track what the system is doing.

- 76bf85aa68403760a9e77e2df6de96322a5a1fbc
  increase network layer observation window's length

# Comments

See details in #1281

### DIFF
--- a/lib/core/src/Cardano/Pool/Metrics.hs
+++ b/lib/core/src/Cardano/Pool/Metrics.hs
@@ -177,7 +177,7 @@ monitorStakePools tr nl db@DBLayer{..} = do
                 readPoolRegistration poolId >>= \case
                     Nothing -> putPoolRegistration sl0 r
                     Just{}  -> pure ()
-            readPoolProductionCursor k
+            readPoolProductionCursor (max 100 k)
 
     forward
         :: NonEmpty Block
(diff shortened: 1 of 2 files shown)


KtorZ:
bors try

lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -175,7 +177,7 @@ monitorStakePools tr nl db@DBLayer{..} = do
                 readPoolRegistration poolId >>= \case
                     Nothing -> putPoolRegistration sl0 r
                     Just{}  -> pure ()
-            readPoolProductionCursor k
+            readPoolProductionCursor (max 100 k)
```
paweljakubas, line 180: why 100 here?

paweljakubas: ah, ok. explained below

paweljakubas (approved):
LGTM!

PR #1286, opened 2020-01-20
document --trace-NAME option and extend doc descriptions

KtorZ:
# Issue Number

#1273 

# Overview

- 5e690157e636532fa27de88622aad8f85b7919e5
  document --trace-NAME option and extend doc descriptions
    Now:

  ```
  Usage: cardano-wallet-jormungandr launch [--listen-address HOST]
                                           ([--random-port] | [--port INT])
                                           [--node-port INT] [--state-dir DIR]
                                           [--sync-tolerance DURATION]
                                           [--trace-NAME SEVERITY]
                                           (--genesis-block-hash STRING |
                                           --genesis-block FILE)
                                           [[-- ARGUMENTS...]]
    Launch and monitor a wallet server and its chain producers.

  Available options:
    -h,--help                Show this help text
    --help-tracing           Show help for tracing options
    [...]
    --log-level SEVERITY     Global minimum severity for a message to be logged.
                             Individual tracers severities still need to be
                             configured independently. Defaults to "DEBUG".
    --trace-NAME SEVERITY    Individual component severity for 'NAME'. See
                             --help-tracing for details and available tracers.
    [...]
  ```

  ```
  Additional tracing options:

    --log-level SEVERITY     Global minimum severity for a message to be logged.
                             Defaults to "DEBUG".

    --trace-NAME=off         Disable logging on the given tracer.
    --trace-NAME=SEVERITY    Minimum severity for a message to be logged, or
                             "off" to disable the tracer. Note that component
                             traces still abide by the global log-level. For
                             example, if the global log level is "INFO", then
                             there will be no "DEBUG" messages whatsoever.
                             Defaults to "INFO".

  The possible log levels (lowest to highest) are:
    debug info notice warning error critical alert emergency

  The possible tracers are:
    application    About start-up logic and the server's surroundings.
    api-server     About the HTTP API requests and responses.
    wallet-engine  About background wallet workers events and core wallet engine.
    wallet-db      About database operations of each wallet.
    pools-engine   About the background worker monitoring stake pools and stake pools engine.
    pools-db       About database operations on stake pools.
    network        About networking communications with the node.
    daedalus-ipc   About inter-process communications with Daedalus.
  ```

# Comments

Note: this required the introduction of an artificial option `--trace-NAME` which parsing always fails (of type `Void`). If a user tries to pass in `--trace-NAME`, it'll get an error like:

```
$ cardano-wallet-jormungandr launch --trace-NAME patate --genesis-block block.bin
[...]

Issue #1269, opened 2020-01-16
OWASP risks - Rate-limiter and others


paweljakubas:
# Context

Here I am working on the following assumption : 
**cardano-wallet API will be run locally, listening on localhost, for use by a trusted application (daedalus).**

**If it is going to be part of public API** the following areas need to be addressed : 
(a) obligatory use of HTTPS
Here, remember about adding response header HSTS, ie.,
`("Strict-Transport-Security", "max-age=31536000")`
(b) multi-level use of rate-limiter (on reverse proxy, gateway, and also attached directly to API as will be covered below) BEFORE next security components
(c) authentication (as a measure against DoS attacks and an input for access control)
Here, huge number of options are available (starting from simple ones)
- basic http authentication
- using secure password storage (Scrypt, Argon2, Bcrypt, ...) - see recent recommendation : https://pages.nist.gov/800-63-3/sp800-63b.html (and here, usage of memory-hard hash functions)
- Bearer auth scheme - OAuth2

(d) audit logging (so extend current logging by adding authorization logs). Consider using DB solution for logging (as a way to have in-house analytics on top of it or just try to use elasticsearch or summologic solutions directly on logs) 
(e) access control - differentiate actions AFTER authorization (401 vs 403). Think about Access Control Lists (ACL) and privilege escalation.

Generally speaking the above security mechanisms (ie., security controls of rate-limiting, encryption, authentication and audit logging), when properly applied, can address STRIDE threat (Spoofing, Tampering, Repudation, Information disclosure, Denial of service, Elevation of privilege).

Given the assumption, though, I think it would be beneficial to introduce rate-limiter solution to cardano-wallet and to have a solution that throttle number of calls possible to API (even trusted application can do it unintentionally). If this happens, then adequate response will be emitted.

# Decision

# Acceptance Criteria

---

# Development 

If we are to use rate-limiter then we should probably try to use https://hackage.haskell.org/package/wai-middleware-throttle
and as a test we should use : 
```
for i in {1..100}; do curl http://localhost:8090/v2/wallets --http1.1 ; done
``` 
and depending on number of request set in a given period for some of them we should end up with response header : `429 Too Many Requests` 

`Conclusions` : consider adding rate limiter as sketched above. If the assumption is not valid anymore then all security mechanisms enumerated above needs to be implemented in order to address STRIDE threat.  

# QA

jonathanknowles:
Hi @paweljakubas 

Thanks for writing this up.

[...]

Issue #1264, opened 2020-01-14
OWASP risks : Insufficient Logging & Monitoring


paweljakubas:
# Context

**Risk API10:2019**

Insufficient logging and monitoring, coupled with missing or ineffective integration with incident response, allows attackers to further attack systems, maintain persistence, pivot to more systems to tamper with, extract, or destroy data. Most breach studies demonstrate the time to detect a breach is over 200 days, typically detected by external parties rather than internal processes or monitoring.

# Decision
1. Check what is logged for each API endpoint
2. Check what happens when the instance is long lived when it comes to logs. 

# Acceptance Criteria

---

# Development 

## ad 1

API logging is realized in https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Network/Wai/Middleware/Logging.hs#L80
Basically in every API call we log : request id (every new api call incremented +1), method name and endpoint. Also for the corresponding call response is logged (the request id, http response code and the time it tooks from request to response. Eg.,
```
[cardano-wallet.api-server:Info:44] [2020-01-14 09:48:37.04 UTC] [RequestId 1] [GET] /v2/wallets/135bfb99b9f7a0c702bf8c658cc0d9b1a0d797a2
[cardano-wallet.api-server:Info:44] [2020-01-14 09:48:37.05 UTC] [RequestId 1] 200 OK in 0.009605239s
```
This setting follows from https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Network/Wai/Middleware/Logging.hs#L275-L295
`Public` and `Info` are set for integration testing and cardano-wallet-jormungandr - here we can set it via command line options: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/jormungandr/exe/cardano-wallet-jormungandr.hs#L360). Also, we do care that some values are never shown : https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/Api/Server.hs#L351-L358
Which is fine and covers all values that need to be obfuscated.

Additionally, some actions taken in `Cardano.Wallet` also result in logging that reveals information present reflected in API calls (although as  indirectly and at later time). In details, this modules outputs : 
- https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet.hs#L633-L635
- https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet.hs#L691-L701
[...]

Issue #1263, opened 2020-01-14
OWASP risks : Improper Assets Management


paweljakubas:
# Context

**Risk API9:2019**

APIs tend to expose more endpoints than traditional web applications, making proper and updated documentation highly important. Proper hosts and deployed API versions inventory also play an important role to mitigate issues such as deprecated API versions and exposed debug endpoints.

# Decision

Check if API is up-to-date reflected in documentation 

# Acceptance Criteria

---

# Development 

*Conclusion*: Each endpoint, its corresponding inputs and outputs is documented, and upadted upon development in https://input-output-hk.github.io/cardano-wallet/api/edge/
There are no undocumented and bogus endpoints operative.
# QA

PR #1287, opened 2020-01-20
OWASP api security

paweljakubas:
# Issue Number

#1260  
#1263 
#1264 
#1286 
#1269 

# Overview

- [x] I have added example SQL injection tests
- [x] I have added API security version-controlled document in specifications

# Comments


lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
```diff
@@ -177,6 +177,37 @@ spec = do
             , expectFieldNotEqual passphraseLastUpdate Nothing
             ]
 
+    it "CREATE_WALLET_01 - SQL injection attempt when creating a wallet" $ \ctx -> do
+        let mnemonics = [ "pulp", "ten", "light", "rhythm", "replace"
+                        , "vessel", "slow", "drift", "kingdom", "amazing"
+                        , "negative", "join", "auction", "ugly", "symptom"] :: [Text]
+        let payload = Json [json| {
+                "name": "new wallet\",'',''); DROP TABLE \"wallet\"; --",
```
paweljakubas: why `"new wallet\",'','');` it follows from https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/DB/Sqlite.hs#L585 (so after name there are two more fields to be filled) and the use of `insert_` https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/DB/Sqlite.hs#L415

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
```diff
@@ -177,6 +177,37 @@ spec = do
...
+            , expectFieldEqual (#balance . #getApiT . #total) (Quantity 0)
+            , expectFieldEqual (#balance . #getApiT . #reward) (Quantity 0)
+            , expectEventually ctx (Link.getWallet @'Shelley)
+                (#state . #getApiT) Ready
+            , expectFieldEqual delegation (NotDelegating)
+            , expectFieldEqual walletId
+                "135bfb99b9f7a0c702bf8c658cc0d9b1a0d797a2"
+            , expectFieldNotEqual passphraseLastUpdate Nothing
+            ]
+        rl <- request @[ApiWallet] ctx (Link.listWallets @'Shelley) Default Empty
```
paweljakubas: checking if the table is not dropped although we are sure because of the above : 
```
expectFieldEqual walletName "new wallet\",'',''); DROP TABLE \"wallet\"; --"
```

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
paweljakubas: interesting fact : When I replaced "insert_" with wrongly constructing of SQL text by hand (using T.intersperse) and then rawExecute) I was managed to produce SQL injection, and drop wallet table .... So escaping used in persistent-sqlite and persistent (https://github.com/yesodweb/persistent/blob/53fa7589db2cabe586d506fea12006b759b1ce79/persistent-sqlite/Database/Persist/Sqlite.hs#L483) works properly for this kind of injections (so called `First Order SQL injection attacks`)

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
```diff
@@ -177,6 +177,37 @@ spec = do
             , expectFieldNotEqual passphraseLastUpdate Nothing
             ]
 
+    it "CREATE_WALLET_01 - SQL injection attempt when creating a wallet" $ \ctx -> do
```
piotr-iohk: Maybe, for better recognition, we could tag those tests following the OWASP naming (or ordering) of particular risks (from https://owasp.org/www-project-api-security/). So for the injection it could be:
 - `OWASP_INJECTION_CREATE_WALLET` or
 - `OWASP_API8_CREATE_WALLET`

piotr-iohk: Also, maybe it makes sense to add similar injection tests for some other requests that have string type payload parameters?
 - https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/putWallet
 - https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/putWalletPassphrase
 - https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/postByronRandomWallet (and others, i.e. icarus, trezor, ledger)
 - https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/postTransaction
 - https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/postTransactionFee
 - https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/migrateByronWallet
 - https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/joinStakePool

paweljakubas: will switch to `OWASP_INJECTION_CREATE_WALLET`

paweljakubas: @piotr-iohk when it comes to other injection tests...I played with a number of endpoints last week when researching - was not able to have any malfunction. mainly because we use `persistent`'s `insert_`, `repsert`,... all of them use proper escaping... which is very good. we use in one case, in very controlled, safe way, `rawExecute`

specifications/api/security.md
```diff
@@ -0,0 +1,176 @@
...
+
+If cardano-wallet API part will become part of other component that will be deployed publicly, it would be good to consider creating restrictive user as a consequence of principle of least authority (POLA) - ie., user/processes in the system will be given only those permissions needed to its job, no more, no less. So, if we identify that user does need to DROP/DELETE then we could add something like that upon DB creation:
+```
+CREATE USER restricted_api_user PASSWORD 'password';
+GRANT SELECT, INSERT ON table1, table2 TO restricted_api_user;
+```
+
+### OWASP risk : Improper Assets Management
+
+**Conclusion**: Each endpoint, its corresponding inputs and outputs is documented, and upadted upon development in https://input-output-hk.github.io/cardano-wallet/api/edge/
```
jonathanknowles:
```suggestion
**Conclusion**: Each endpoint, its corresponding inputs and outputs is documented, and updated upon development in https://input-output-hk.github.io/cardano-wallet/api/edge/
```

specifications/api/security.md
```diff
@@ -0,0 +1,176 @@
...
+
+Logging is emitted to stdout and there is no way to change it now being the user of cardano-wallet. Daedalus, using cardano-wallet inside, realizes rotation based logging itself. It is set, user cannot control it. Moreover, it includes also logs from its layer. Some users, though, use cardano-wallet without Daedalus.
+
+**Conclusions** : Logs land to stdout, which has its pros and cons (it can be lost when terminal is set and by definition is less persistant). It would be good to add option for rotation, and embrace possibilities of iohk-monitoring-framework : https://github.com/input-output-hk/iohk-monitoring-framework/blob/bea0e079fc32ed316ce352d17d14199a680e3f6c/iohk-monitoring/src/Cardano/BM/Data/Rotation.lhs#L29
+Maybe it is even good to set this as default - users with problems could then immediately share logs with us.
+
+### OWASP risk : Security Misconfiguration
+
+1. Misconfigured HTTP headers
+Even is API is standalone piece of software, we need to think how it will behave being a part of HTML application. And here, XSS attacks against web browser clients can be realized. In order to avoid it a number of measures can be taken :
```
jonathanknowles:
```suggestion
Even if the API is a standalone piece of software, we need to think how it will behave as part of an HTML application. And here, XSS attacks against web browser clients can be realized. In order to avoid it a number of measures can be taken:
```

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
```diff
@@ -177,6 +177,37 @@ spec = do
             , expectFieldNotEqual passphraseLastUpdate Nothing
             ]
 
+    it "OWASP_INJECTION_CREATE_WALLET_01 - SQL injection when creating a wallet" $ \ctx -> do
+        let mnemonics = [ "pulp", "ten", "light", "rhythm", "replace"
+                        , "vessel", "slow", "drift", "kingdom", "amazing"
+                        , "negative", "join", "auction", "ugly", "symptom"] :: [Text]
```
jonathanknowles: I'm going to be the bad guy here and point out the team coding standard. :stuck_out_tongue: 

In particular:
- We should stick to the line length limit here.
- We should avoid using variable horizontal alignment where indentation could be used instead.

Here we can solve both of these issues by writing:
```hs
    it "OWASP_INJECTION_CREATE_WALLET_01 - \  
        \SQL injection when creating a wallet" $ \ctx -> do   
        let mnemonics =   
                [ "pulp", "ten", "light", "rhythm", "replace" 
                , "vessel", "slow", "drift", "kingdom", "amazing" 
                , "negative", "join", "auction", "ugly", "symptom"
                ] :: [Text]
```

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
```diff
@@ -177,6 +177,37 @@ spec = do
...
+        let payload = Json [json| {
+                "name": "new wallet\",'',''); DROP TABLE \"wallet\"; --",
+                "mnemonic_sentence": #{mnemonics},
+                "passphrase": "12345678910"
+                } |]
+        r <- request @ApiWallet ctx (Link.postWallet @'Shelley) Default payload
+        verify r
+            [ expectResponseCode @IO HTTP.status201
+            , expectFieldEqual walletName "new wallet\",'',''); DROP TABLE \"wallet\"; --"
```
jonathanknowles:
```suggestion
            , expectFieldEqual
                walletName
                "new wallet\",'',''); DROP TABLE \"wallet\"; --"     
```

specifications/api/security.md
paweljakubas: corrected

specifications/api/security.md
paweljakubas: corrected

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
jonathanknowles: @paweljakubas By the way, have you tried using other types of quotation here? Such as single quotes for table names. There are also unicode homoglyphs for `(` and `)`, which may (or may not) be converted silently to their ASCII equivalents during processing.

The following article talks about using Unicode homoglyphs for SQL injection attacks: https://hackernoon.com/%CA%BC-%C5%9B%E2%84%87%E2%84%92%E2%84%87%E2%84%82%CA%88-how-unicode-homoglyphs-will-break-your-custom-sql-injection-sanitizing-functions-1224377f7b51

There's also a homoglyph generator here: https://www.irongeek.com/homoglyph-attack-generator.php

jonathanknowles (changes requested):
Hi @paweljakubas 

Thanks for making this PR!

I've had a chance to look through. I have two main comments:

1. The SQL injection attack test looks great. Have you thought about the possibility of expanding this test to cover attacks with Unicode homoglyphs? There are homoglyphs for many of the punctuation characters such as `"`, `(`, and `)`, which may (or may not be) missed by functions that are supposed to be performing escaping of reserved characters. I have no idea if the libraries we are using handle these cases correctly, so it might be useful to check (perhaps in a subsequent PR).

2. Regarding the document **`specifications/api/security.md`**. I think it's very useful for us to have this document to capture your findings. Though I have a suggestion regarding the **location**: perhaps we could move it to a dedicated `reports` directory? For example: `reports/api/security/2020-01-21.md`. My reason for this suggestion is that this document is a report that describes the current state of the world at a particular time, rather than a specification to be continually updated. What do you think?

Other than these two main points, and some points regarding style (for which I've added individual comments), this PR looks good to me. :+1:

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
paweljakubas: corrected

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
paweljakubas: yes, I tried to use diverse homoglyphs in names. I extended the test and added some. It seems, persistent-sqlite escape is immune to this kind of meddling;-)

paweljakubas:
@jonathanknowles I changed the location to reflect the date of API security audit : 
```
renamed:    specifications/api/security.md -> specifications/api/security/2020-01-21.md
```

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
jonathanknowles: Great, thanks!

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
```diff
@@ -177,6 +177,47 @@ spec = do
             , expectFieldNotEqual passphraseLastUpdate Nothing
             ]
 
+    describe "OWASP_INJECTION_CREATE_WALLET_01 - SQL injection when creating a wallet" $  do
```
jonathanknowles:
```suggestion
    describe "OWASP_INJECTION_CREATE_WALLET_01 - \
        \SQL injection when creating a wallet" $ do
```

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
```diff
@@ -177,6 +177,47 @@ spec = do
...
+                  , "new wallet','\346\949\8466\8455\8450\430\8217',\
+                    \'\346\949\8466\8455\8450\430\8217'); DROP TABLE \"wallet\"; --"
+                  ) ]
+        forM_ matrix $ \(nameIn, nameOut) -> it nameIn $ \ctx -> do
+            let payload = Json [json| {
+                    "name": #{nameIn},
+                    "mnemonic_sentence": #{mnemonics},
+                    "passphrase": "12345678910"
+                    } |]
+            r <- request @ApiWallet ctx (Link.postWallet @'Shelley) Default payload
```
jonathanknowles:
```suggestion
            let postWallet = Link.postWallet @'Shelley
            r <- request @ApiWallet ctx postWallet Default payload
```
or
```suggestion
            r <- request
                @ApiWallet ctx (Link.postWallet @'Shelley) Default payload
```
https://github.com/input-output-hk/cardano-wallet/wiki/Coding-Standards#limit-line-length-to-80-characters

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
```diff
@@ -177,6 +177,47 @@ spec = do
...
+                , expectFieldEqual (#balance . #getApiT . #total) (Quantity 0)
+                , expectFieldEqual (#balance . #getApiT . #reward) (Quantity 0)
+                , expectEventually ctx (Link.getWallet @'Shelley)
+                    (#state . #getApiT) Ready
+                , expectFieldEqual delegation (NotDelegating)
+                , expectFieldEqual walletId
+                    "135bfb99b9f7a0c702bf8c658cc0d9b1a0d797a2"
+                , expectFieldNotEqual passphraseLastUpdate Nothing
+                ]
+            rl <- request @[ApiWallet] ctx (Link.listWallets @'Shelley) Default Empty
```
jonathanknowles:
```suggestion
            let listWallets = Link.listWallets @'Shelley
            rl <- request @[ApiWallet] ctx listWallets Default Empty
```
https://github.com/input-output-hk/cardano-wallet/wiki/Coding-Standards#limit-line-length-to-80-characters

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
```diff
@@ -177,6 +177,47 @@ spec = do
...
+                    "mnemonic_sentence": #{mnemonics},
+                    "passphrase": "12345678910"
+                    } |]
+            r <- request @ApiWallet ctx (Link.postWallet @'Shelley) Default payload
+            verify r
+                [ expectResponseCode @IO HTTP.status201
+                , expectFieldEqual walletName nameOut
+                , expectFieldEqual
+                    (#addressPoolGap . #getApiT . #getAddressPoolGap) 20
+                , expectFieldEqual (#balance . #getApiT . #available) (Quantity 0)
```
jonathanknowles:
```suggestion
                , expectFieldEqual
                    (#balance . #getApiT . #available)
                    (Quantity 0)
```
https://github.com/input-output-hk/cardano-wallet/wiki/Coding-Standards#limit-line-length-to-80-characters

jonathanknowles (changes requested):
Hi @paweljakubas 

Thanks for addressing the previous review comments. :+1: 

I just have one suggestion: instead of placing the document in the `specifications` directory, I would like to propose that we create a dedicated `reports` directory and place it in there (as it seems to be a technical report, rather than a specification to be followed). What do you think?

Aside from this, I've added a few coding standard-related change suggestions, which should hopefully be easy to apply.

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
jonathanknowles: > yes, I tried to use diverse homoglyphs in names. I extended the test and added some. It seems, persistent-sqlite escape is immune to this kind of meddling;-)

Nice!

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
paweljakubas: done

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
paweljakubas: done

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
paweljakubas: done

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
paweljakubas: done

paweljakubas:
did:
```
specifications/api/security/2020-01-21.md -> reports/security/2020-01-21.md
```

jonathanknowles (approved):
LGTM!

paweljakubas:
bors retry

Issue #1276, opened 2020-01-16
Sort stake pools on `desirability`


Anviking:
# Context

WB-48

We implemented a module for calculating the `desirability` of stake pools.

WB-48 asks us to order stake pools by `desirability`. This is a step in the direction on ordering by non-myopic pool member rewards as specified by the delegation design spec.

# Decision

Sort the `listStakePools` endpoint on `desirability` instead of performance.

# Acceptance Criteria

- [x] stake-pools (returned from API) _must_ be sorted by desirability.

---

# Development

- We need to specify `R`, the total rewards, somehow  ==> #1288 
- Should be just a couple lines of code to calculate `desirability` and sort using `Cardano.Pool.Ranking`. ==> #1289 
- Not sure how to test. Maybe nothing. Maybe a simple hard-coded integration test. I don't think duplicating the setup for the desirability-calculation in a test would be preferable.

- [x] Revised calculation of the apparent performance ==> #1280 

- [x] Log constants used to calculate `desirability` (as NOTICE when they look oddly small) ==> #1298 
- [x] Small fixup to integration tests ==> #1299

# QA 

- Pools are now ordered by desirability. The desirability-calculation rely on information derived from reward parameters in `block0`. The wallet will exit if the information somehow would not be there.
- Existing integration integration tests verify that the `desirability` ordering the same as the previous `performance`-ordering. (of course, in general they are not the same)
- Golden tests comparing parts of the reward-calculation to the jormungandr docs https://github.com/input-output-hk/cardano-wallet/blob/a0c66af028316d999be49e2e5055fec2fd8e01db/lib/jormungandr/test/unit/Cardano/Wallet/Jormungandr/RewardsSpec.hs#L74.
- `totalRewards` (a.k.a `R`)  will not exactly match jormungandr. This shouldn't matter. It should have minimal effect on the ranking, assuming pools don't have a truly massive *fixed* costs, or other extreme scenarios.
    - Our  `R` takes minting into account, but not tx fees. If there is no minting on the chain we're connecting to, we will estimate it to be 0, and pools will have equal desirability. [thread](https://input-output-rnd.slack.com/archives/GBT05825V/p1579694549014000?thread_ts=1579694010.013900&cid=GBT05825V) this will not happen on ITN.
    - `R` doesn't take the finite reward-pot into account either https://input-output-rnd.slack.com/archives/GBT05825V/p1579794144028500 🤷‍♂

piotr-iohk:
[...]

PR #1288, opened 2020-01-20
Calculate rewards from genesis parameters

Anviking:
# Issue Number
#1276 

# Overview

- e2c8683848493634e8bcb13930b9e8e6bb6f5f5b
  [WIP] Calculate rewards from block0 params
  I am trying to keep the reward-calculation jörmungandr-specific. I don't
want to extend BlockChainParameters in core, since it would require
DB-migrations (I'm pretty sure). Even so, the parameters /are/
jormungandr-specific, and should not belong in core.

- 79f53435242b07a2f877a3af1c63b6e290fbe05d
  Move to separate module + add Ratio type
  
- 2e2054344a09d22848f95ba9bea10942f0cc3cf5
  add rankingEpochConstants function
  
- b905f2cc028b852857e45beef023c0ca03ad5427
  simplify and fix Jormungandr reward params extraction
  
- 3047095c57765b109979e6f4a67945c800f2d178
  finish wiring up incentive parameters with jormungandr's startup
  
- e15f53a45d8bb0489a08ef7f91fe9021a76055af
  add simple unit test to compare ITN params with calculation
  
- 1c2fcb154b3df043f7ea218e35114e3839fe9cb8
  fixup: typo in Ranking module
  
- 99dd776d5a99f8c066b59898ff2bed4f7353708f
  re-generate nix machinery
  
- 8ec5ef6586f02da59d0dd5e7d408b255c57491b4
  additional test vectors for rewards calculation
  
- 905cb71807adc658d33bad90c991da42961e79ef
  make sure pools are sorted by apparent performance in scenario
  
- 2037c6049f72b2274f38d91915d310395bc9c690
  use total block production instead of latest one
    The rationale is that it is quite confusing as many pools
  have only a few blocks produced per epoch (1 or 2) and sometimes
  even 0 for several epochs.
  This makes it even harder to assess whether a pool is doing okay
  or not. Looking at the total number of blocks produced makes more
  sense.

- 22fff45281556d3996f4d086cf7c0c2aa587439d
  add simple properties to test readTotalProduction

# Comments

- I am trying to keep the reward-calculation jörmungandr-specific. I don't
want to extend BlockChainParameters in core, since it would require
DB-migrations (I'm pretty sure). Even so, the parameters /are/
jormungandr-specific, and should not belong in core.

### DIFF
--- a/lib/jormungandr/src/Cardano/Wallet/Jormungandr.hs
+++ b/lib/jormungandr/src/Cardano/Wallet/Jormungandr.hs
@@ -64,8 +64,6 @@ import Cardano.Launcher
     ( ProcessHasExited (..), installSignalHandlers )
 import Cardano.Pool.Metrics
     ( StakePoolLayer, StakePoolLog, monitorStakePools, newStakePoolLayer )
-import Cardano.Pool.Ranking
-    ( EpochConstants (..), unsafeMkNonNegative, unsafeMkPositive )
 import Cardano.Wallet
     ( WalletLog )
 import Cardano.Wallet.Api
@@ -137,6 +135,8 @@ import Control.Concurrent.Async
     ( race )
 import Control.DeepSeq
     ( NFData )
+import Control.Exception
+    ( throwIO )
 import Control.Monad
     ( void )
 import Control.Tracer
@@ -219,14 +219,14 @@ serveWallet Tracers{..} sTolerance databaseDir hostPref listen backend beforeMai
         Left e -> handleNetworkStartupError e
         Right (cp, nl) -> do
             let nPort = Port $ baseUrlPort $ _restApi cp
-            let (_, bp) = staticBlockchainParameters nl
+            let (block0, bp) = staticBlockchainParameters nl
             let byronTl = newTransactionLayer (getGenesisBlockHash bp)
             let icarusTl = newTransactionLayer (getGenesisBlockHash bp)
             let shelleyTl = newTransactionLayer (getGenesisBlockHash bp)
             let poolDBPath = Pool.defaultFilePath <$> databaseDir
             Pool.withDBLayer stakePoolDbTracer poolDBPath $ \db ->
                 withSystemTempDirectory "stake-pool-metadata" $ \md -> do
-                    poolApi <- stakePoolLayer nl db md
+                    poolApi <- stakePoolLayer nl db block0 md
                     byronApi <- apiLayer "byron" byronTl nl
                     icarusApi <- apiLayer "icarus" icarusTl nl
                     shelleyApi <- apiLayer "shelley" shelleyTl nl
@@ -282,22 +282,18 @@ serveWallet Tracers{..} sTolerance databaseDir hostPref listen backend beforeMai
     stakePoolLayer
         :: NetworkLayer IO t J.Block
         -> Pool.DBLayer IO
+        -> J.Block
         -> FilePath
         -> IO (StakePoolLayer IO)
-    stakePoolLayer nl db metadataDir = do
+    stakePoolLayer nl db block0 metadataDir = do
         void $ forkFinally (monitorStakePools tr nl' db) onExit
+        getEpCst <- maybe (throwIO ErrStartupMissingIncentiveParameters) pure $
+            J.rankingEpochConstants block0
         pure $ newStakePoolLayer tr getEpCst db nl' metadataDir
       where
         nl' = toSPBlock <$> nl
         tr  = contramap (MsgFromWorker mempty) stakePoolEngineTracer
         onExit = defaultWorkerAfter stakePoolEngineTracer
-
-        -- FIXME Extract these constants from the genesis block
-        getEpCst = const $ EpochConstants
-            { leaderStakeInfluence = unsafeMkNonNegative 0
-            , desiredNumberOfPools = unsafeMkPositive 100
-            , totalRewards = Quantity 3452054796000
-            }
 
     handleNetworkStartupError :: ErrStartup -> IO ExitCode
     handleNetworkStartupError err = do
@@ -384,6 +380,8 @@ instance ToText ApplicationLog where
             ErrStartupNodeNotListening -> mempty
                     <> "Waited too long for Jörmungandr to become available. "
                     <> "Giving up!"
+            ErrStartupMissingIncentiveParameters ->
+                "Couldn't find incentive parameters in the genesis block."
         MsgServerStartupError startupErr -> case startupErr of
             ListenErrorHostDoesNotExist host -> mempty
                     <> "Can't listen on "
@@ -418,6 +416,7 @@ exitCodeNetwork = \case
         ErrGetBlockchainParamsNetworkUnreachable _ -> 30
         ErrGetBlockchainParamsGenesisNotFound _ -> 31
         ErrGetBlockchainParamsIncompleteParams _ -> 32
+    ErrStartupMissingIncentiveParameters -> 32
     ErrStartupInvalidGenesisBlock _ -> 33
     ErrStartupInvalidGenesisHash _ -> 34
     ErrStartupCommandExited pe -> case pe of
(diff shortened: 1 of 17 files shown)


lib/jormungandr/src/Cardano/Wallet/Jormungandr/Binary.hs
```diff
@@ -1053,3 +1155,91 @@ poolRegistrationsFromBlock (Block _hdr fragments) = do
...
+-- Rewards
+--
+-- TODO: Maybe move to separate module
+--
+-- Based on the information in:
+-- https://github.com/input-output-hk/chain-libs/blob/master/chain-impl-mockchain/doc/incentives.md
+
+rewardParamsFromBlock0
+    :: Block
+    -> Maybe (RewardParams, TreasuryTax, Maybe (Word64, Word64))
```
Anviking: I think I should have introduced a `Ratio` type instead of `(Word64, Word64)`.

lib/jormungandr/src/Cardano/Wallet/Jormungandr/Binary.hs
```diff
@@ -824,6 +839,63 @@ getConfigParam = label "getConfigParam" $ do
...
 
+    getRewardParams :: Get RewardParams
+    getRewardParams = do
+        tag <- getWord8
+        case tag of
+            1 -> LinearRewards <$> getLinearParams
+            2 -> HalvingRewards <$> getHalvingParams
+            _ -> fail $ "Unknown reward parameters with tag " ++ show tag
+
+    -- NOTE: getHalvingParams and getLinearParams are currently identical.
```
Anviking, line 848: Actually, maybe we should leverage that they are identical. Jormungandr does not, but if jörmungandr differentiates them in the future… that's a problem for… the future.

lib/jormungandr/src/Cardano/Wallet/Jormungandr/Binary.hs
```diff
@@ -824,6 +836,49 @@ getConfigParam = label "getConfigParam" $ do
...
+
+    getRatio :: Get Ratio
+    getRatio = do
+        ratioNum <- getWord64be
+        ratioDen <- getWord64be
+        return $ Ratio ratioNum ratioDen
+
+
+    getRewardLimitByAbsoluteStake :: Get Ratio
+    getRewardLimitByAbsoluteStake = getRatio
```
KtorZ: This function seems superfluous and adds an unnecessary indirection. I'd suggest to simply use `getRatio`.

lib/jormungandr/src/Cardano/Wallet/Jormungandr/Binary.hs
KtorZ: When do you mean, "they" are identical ? What is identical? The shape of the serialized data?

Anviking: Yes.

lib/jormungandr/src/Cardano/Wallet/Jormungandr/Rewards.hs
```diff
@@ -0,0 +1,127 @@
...
+-- Based on the information in:
+-- https://github.com/input-output-hk/chain-libs/blob/master/chain-impl-mockchain/doc/incentives.md
+
+rewardParamsFromBlock0
+    :: Block
+    -> Maybe (RewardFormula, TreasuryTax, Maybe Ratio)
+rewardParamsFromBlock0 block = do
+        rewardParams <- find $ \case
+                ConfigRewardFormula x -> Just x
+                _ -> error "missing config reward"
```
KtorZ: Why use `find` and `Maybe` for safety while using `error` for missing values?

Anviking: Oops, this was while debugging, to have more information than `Maybe`. Ultimately an `Either` would be good, I think, but I am not prioritising that.

KtorZ:
bors try

KtorZ (approved):

KtorZ:
bors retry

lib/jormungandr/src/Cardano/Wallet/Jormungandr.hs
```diff
@@ -282,23 +282,19 @@ serveWallet Tracers{..} sTolerance databaseDir hostPref listen backend beforeMai
     stakePoolLayer
         :: NetworkLayer IO t J.Block
         -> Pool.DBLayer IO
+        -> J.Block
         -> FilePath
         -> IO (StakePoolLayer IO)
-    stakePoolLayer nl db metadataDir = do
+    stakePoolLayer nl db block0 metadataDir = do
         void $ forkFinally (monitorStakePools tr nl' db) onExit
+        getEpCst <- maybe (throwIO ErrStartupMissingIncentiveParameters) pure $
```
Anviking, line 290: 👍

lib/jormungandr/src/Cardano/Wallet/Jormungandr/Rewards.hs
```diff
@@ -0,0 +1,169 @@
...
+  where
+    q :: Double -> Quantity "lovelace" Word64
+    q = Quantity . round
+
+    ep = fromIntegral (unEpochNo epochNo)
+
+    linearAbsorption RewardParams{rFixed,rRatio,rEpochStart,rEpochRate} =
+        if a > c then 0 else c - a
+      where
+        a = r * (n / e)
```
Anviking: I think there should be a `floor` here.

See: https://github.com/input-output-hk/chain-libs/blob/149e01cf61a9eef2660a6944fe0acc73d490e70f/chain-impl-mockchain/doc/incentives.md#L105-L117

example with rate > 1

Anviking: _Looking more into other things first_

KtorZ: You are pointing to the linear formula but the example from your link is about the halving formula ?

KtorZ: Yet, these are nice examples to use as test vectors perhaps.

lib/jormungandr/src/Cardano/Wallet/Jormungandr/Binary.hs
```diff
@@ -1044,12 +1082,57 @@ overrideFeePolicy linearFee@(LinearFee a b _) override =
...
+    rewardLimit   <- mRewardLimit
+    treasuryTax   <- mTreasuryTax
+    rewardFormula <- mRewardFormula
+    pure $ \ep totalStake -> EpochConstants
+        { leaderStakeInfluence =
+            unsafeMkNonNegative 0
+        , desiredNumberOfPools =
+            unsafeMkPositive $ fromIntegral $ maxParticipation poolCapping
+        , totalRewards =
+            rewardsAt (rewardLimit, totalStake) treasuryTax ep rewardFormula
```
Anviking, line 1107: super-nitpick: if the name is rewards**At**, the `epoch` should be the first argument, I think.

KtorZ:
Failure is unrelated. Opening a corresponding ticket.

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
```diff
@@ -843,3 +848,10 @@ joinStakePoolWithFixtureWallet ctx = do
             [ expectFieldEqual delegation (Delegating (p ^. #id))
             ]
     return (w, p)
+
+sortByPerformance
+    :: Functor f
+    => (a, f [ApiStakePool])
+    -> (a, f [ApiStakePool])
+sortByPerformance =
+    second (fmap (sortOn (Down . view #apparentPerformance)))
```
Anviking, line 857: This should not be needed!

The problem is that the total rewards are 0 in the integration tests.

If `R` is not 0, the tests pass without `sortByPerformance`.

This highlights an interesting problem: what if the rewards *are* 0 in production?

Issue #1279, opened 2020-01-17
`persistent` clear foreign tables on automatic migration due to cascading delete


KtorZ:
# Context
 
 

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | All  |
| Platform     | All           |
| Installation | N/A             |

See https://input-output-rnd.slack.com/archives/CGH3J2Q1L/p1579253925019600

# Steps to Reproduce

1. Build master.
2. Start the wallet backend with cardano-wallet-jormungandr launch ... 
3. Create a brand new wallet.
4. Stop the wallet backend.
5. Use sqlite3 to manually observe the contents of the seq_state_address table (with something like select distinct(slot) from seq_state_address;).
6. You should see that it typically has ~11 entries (genesis + 10 checkpoints) entries.
7. Add a new column foo to the checkpoint table definition in lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs, giving it a type of Maybe Word32. (Persistent should be able to make an automatic migration for this case, as we've used Maybe.)
8. Rebuild, and then restart the wallet backend.
9. Observe in the logs that persistent has performed several automatic migrations.
10. Stop the wallet backend.

## Expected behavior

11. Now use sqlite3 to look at the entries of the seq_state_address table. Previous entries should still be there (modulo the checkpoints moving forward) and in particular the entry for the genesis checkpoint should remain!

## Actual behavior
 

11. Now use sqlite3 to look at the entries of the seq_state_address table. All previous entries are gone.

---

# Resolution 

- One possible approach: forcing a manual rollback to genesis whenever there are migrations applied to the database, regardless of the migration. Only the foreign tables are affected, so static metadata and wallet credentials are preserved. This mean that we would force resync of wallets on every updates that have database changes, but that's a fair price to pay for now. ==> #1290 

---

# QA 

See comment of https://github.com/input-output-hk/cardano-wallet/pull/1290#issue-364947415

jonathanknowles:
@KtorZ 

Just as an aside, the persistent manual recommends against the use of automatic migrations in production environments:

> Using automated database migrations is only recommended in development environments. Allowing your application to modify your database schema in a production environment is _very strongly discouraged_.

(source: https://www.yesodweb.com/book/persistent)

KtorZ:
Perhaps that is something they could highlight even more, especially in the haddock documentation :sweat_smile:  ... Yet, it's open-source, so we might as well give them a little PR.

[...]

PR #1290, opened 2020-01-20
Force rollback of tables with foreign keys / cascading deletes on automatic db migrations

KtorZ:
# Issue Number

#1279 

# Overview

- 74c232e1ee94ec9c79152312f7294ebe2da7cc64
  force rollback to genesis after automated migrations
    `persistent` clear foreign tables on automatic migration due to
  cascading delete. One possible approach to solve this: forcing a
  manual rollback to genesis whenever there are migrations applied
  to the database, regardless of the migration. Only the foreign
  tables are affected, so static metadata and wallet credentials
  are preserved. This mean that we would force resync of wallets
  on every updates that have database changes, but that's a fair
  price to pay for now.

- de3825dd82ae2f0f7caf7b80102dccddebd00aa8
  remove foreign keys to 'Wallet' table and cascade deletes manually
  This to prevent 'persistent' from doing it during automated migration! We don't do
this for checkpoints because the cascading updates are truly helpful there to manage
the database state and, can be more easily rolled back at any time (since the data
regarding checkpoints entirely comes from the chain).

# Comments

~~:warning: Currently manually testing db migration from various previous versions.~~

Automated migration passes! And on old version, we have a nice log showing what's going on, including manual migration, and handling of automated ones!

```
 *** Migrating from v2019-12-13 ***

[cardano-wallet.application:Info:22] [2020-01-20 17:58:12.47 UTC] Found existing wallet: 99ef09bdab5f62579f2d6db02ba7df5e88b7e865
[cardano-wallet.wallet-db:Notice:28] [2020-01-20 17:58:12.47 UTC] checkpoint table does not contain required field 'active_slot_coeff'. Adding this field with a default value of 1.0.
[cardano-wallet.wallet-db:Notice:28] [2020-01-20 17:58:12.52 UTC] 48 migrations were applied to the database.
[cardano-wallet.wallet-db:Notice:28] [2020-01-20 17:58:12.52 UTC] Some automated database migrations happened. Forced to rollback to genesis.
[cardano-wallet.main:Info:22] [2020-01-20 17:58:12.53 UTC] Wallet backend server listening on 127.0.0.1:8090
[cardano-wallet.wallet-engine:Info:28] [2020-01-20 17:58:12.53 UTC] 99ef09bd: Applying blocks [1167051.7 ... 1167051.7]
[cardano-wallet.wallet-engine:Info:28] [2020-01-20 17:58:12.53 UTC] 99ef09bd: Creating checkpoint at 3b2894d8-[1167051.7#1]
```

So, to summarize what's happening:

1. An existing wallet database is found. 
2. Before running any automatic migration, the database is inspected and an extra column is found. So we manually alter the table to add a column with a default value.
3. Then, persistent runs its own set of automated migrations on top of the database, running 48 of them.
4. This is then detected and induces immediately a rollback to genesis of all the on-chain data.
5. The wallet starts applying blocks and creating checkpoints, starting with a checkpoint at height `#1` (right after genesis)

:tada:

### DIFF
--- a/lib/core/src/Cardano/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/DB/Sqlite.hs
@@ -76,6 +76,8 @@ import Data.Maybe
     ( fromMaybe )
 import Data.Proxy
     ( Proxy (..) )
+import Data.Quantity
+    ( Quantity (..) )
 import Data.Text
     ( Text )
 import Data.Text.Class
@@ -211,7 +213,7 @@ startSqliteBackend
     -> Migration
     -> Tracer IO DBLog
     -> Maybe FilePath
-    -> IO (Either MigrationError SqliteContext)
+    -> IO (Either MigrationError (Quantity "migration" Int, SqliteContext))
 startSqliteBackend manualMigration migrateAll trace fp = do
     backend <- createSqliteBackend trace fp manualMigration (queryLogFunc trace)
     lock <- newMVar ()
@@ -231,7 +233,8 @@ startSqliteBackend manualMigration migrateAll trace fp = do
         Left e -> do
             destroyDBLayer ctx
             pure $ Left e
-        Right _ -> pure $ Right ctx
+        Right ms -> do
+            pure $ Right (Quantity (length ms), ctx)
 
 class Exception e => MatchMigrationError e where
     -- | Exception predicate for migration errors.
@@ -279,6 +282,7 @@ sqliteConnStr = maybe ":memory:" T.pack
 
 data DBLog
     = MsgMigrations (Either MigrationError Int)
+    | MsgForcedRollback
     | MsgQuery Text Severity
     | MsgRun Bool
     | MsgConnStr Text
@@ -343,6 +347,7 @@ instance DefineSeverity DBLog where
         MsgMigrations (Right 0) -> Debug
         MsgMigrations (Right _) -> Notice
         MsgMigrations (Left _) -> Error
+        MsgForcedRollback{} -> Notice
         MsgQuery _ sev -> sev
         MsgRun _ -> Debug
         MsgConnStr _ -> Debug
@@ -362,6 +367,10 @@ instance ToText DBLog where
             fmt $ ""+||n||+" migrations were applied to the database."
         MsgMigrations (Left err) ->
             "Failed to migrate the database: " <> getMigrationErrorMessage err
+        MsgForcedRollback -> mconcat
+            [ "Some automated database migrations happened. Forced to roll back "
+            , "to genesis."
+            ]
         MsgQuery stmt _ -> stmt
         MsgRun False -> "Running database action - Start"
         MsgRun True -> "Running database action - Finish"
--- a/lib/core/src/Cardano/Pool/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Pool/DB/Sqlite.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedLabels #-}
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE UndecidableInstances #-}
@@ -46,15 +47,20 @@ import Cardano.Wallet.Primitive.Types
     , PoolId
     , PoolRegistrationCertificate (..)
     , SlotId (..)
+    , slotMinBound
     )
 import Control.Exception
     ( bracket, throwIO )
+import Control.Monad
+    ( when )
 import Control.Monad.IO.Class
     ( liftIO )
 import Control.Monad.Trans.Except
     ( ExceptT (..) )
 import Control.Tracer
     ( Tracer, traceWith )
+import Data.Function
+    ( (&) )
 import Data.List
     ( foldl' )
 import Data.Map.Strict
@@ -135,13 +141,15 @@ newDBLayer
        -- ^ Database file location, or Nothing for in-memory database
     -> IO (SqliteContext, DBLayer IO)
 newDBLayer trace fp = do
-    let io = startSqliteBackend
-            (ManualMigration mempty)
-            migrateAll
-            trace
-            fp
-    ctx@SqliteContext{runQuery} <- handlingPersistError trace fp io
-    return (ctx, DBLayer
+    let start = startSqliteBackend (ManualMigration mempty) migrateAll trace fp
+    (nMigrations, ctx) <- handlingPersistError trace fp start
+    let dbLayer = mkDBLayer ctx
+    when (nMigrations > Quantity 0) $ do
+        traceWith trace MsgForcedRollback
+        dbLayer & \DBLayer{..} -> atomically $ rollbackTo slotMinBound
+    pure (ctx, dbLayer)
+  where
+    mkDBLayer SqliteContext{runQuery} = DBLayer
         { putPoolProduction = \point pool -> ExceptT $
             handleConstraint (ErrPointAlreadyExists point) $
                 insert_ (mkPoolProduction pool point)
@@ -223,7 +231,7 @@ newDBLayer trace fp = do
             deleteWhere ([] :: [Filter StakeDistribution])
 
         , atomically = runQuery
-        })
+        }
 
 -- | 'Temporary', catches migration error from previous versions and if any,
 -- _removes_ the database file completely before retrying to start the database.
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
@@ -424,7 +424,15 @@ newDBLayer trace defaultFieldValues fp = do
             selectWallet wid >>= \case
                 Nothing -> pure $ Left $ ErrNoSuchWallet wid
                 Just _  -> Right <$> do
-                    deleteCascadeWhere [WalId ==. wid]
+                    deleteWhere [WalId ==. wid]
+                    deleteWhere [TxMetaWalletId ==. wid]
+                    deleteWhere [PrivateKeyWalletId ==. wid]
+                    deleteWhere [SeqStateWalletId ==. wid]
+                    deleteWhere [SeqStatePendingWalletId ==. wid]
+                    deleteWhere [RndStateWalletId ==. wid]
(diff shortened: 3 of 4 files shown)


KtorZ:
bors try

KtorZ:
bors try

lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -211,7 +213,7 @@ startSqliteBackend
     -> Migration
     -> Tracer IO DBLog
     -> Maybe FilePath
-    -> IO (Either MigrationError SqliteContext)
+    -> IO (Either MigrationError (Quantity "migration" Int, SqliteContext))
```
jonathanknowles, line 216: **Suggestion**: use `migrations` (plural) or `migrationCount` here, to really emphasize that it is a _count_ of the number of migrations applied.

lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -231,7 +233,8 @@ startSqliteBackend manualMigration migrateAll trace fp = do
         Left e -> do
             destroyDBLayer ctx
             pure $ Left e
-        Right _ -> pure $ Right ctx
+        Right ms -> do
+            pure $ Right (Quantity (length ms), ctx)
```
jonathanknowles, line 237: **Question:** this count (`ms`) appears to only include _automatic_ migrations.

Should we not also include any _manual_ migrations in this total count?

So: **_manual migrations_ + _automatic migrations_**?

I'm assuming it's possible for there to be cases where there are manual migrations but no automatic migrations. (This would happen if our manual migrations successfully produce the exact schema that Sqlite is expecting.)

lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -362,6 +367,10 @@ instance ToText DBLog where
             fmt $ ""+||n||+" migrations were applied to the database."
         MsgMigrations (Left err) ->
             "Failed to migrate the database: " <> getMigrationErrorMessage err
+        MsgForcedRollback -> mconcat
+            [ "Some automated database migrations happened. Forced to rollback "
```
jonathanknowles:
```suggestion
            [ "Some automated database migrations happened. Forced to roll back "
```
(When used as a noun, we can certainly use "rollback", but here it is used as a **phrasal verb**, so it should take a space.)

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -992,6 +1000,27 @@ findNearestPoint wid sl =
...
+-------------------------------------------------------------------------------}
+
+-- | @persistent@ handles automatic migrations by creating copies of existing
+-- tables, and re-creating them from scratch. When paired with automatic cascade
+-- deletion, this has dramatic consequences on the wallet data integrity (when a
+-- source table is deleted due do a migration, all foreign tables are also
+-- deleted, but only the rows in the source table are then replaced back!).
+--
+-- To cope with this, when any automated migration is detected, we manually
+-- rollback all rows referencing checkpoints and replace the genesis checkpoint.
```
jonathanknowles:
```suggestion
-- roll back all rows referencing checkpoints and replace the genesis checkpoint.
```
(see earlier comment!)

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -992,6 +1000,27 @@ findNearestPoint wid sl =
...
 
+
+{-------------------------------------------------------------------------------
+                                    Logging
+-------------------------------------------------------------------------------}
+
+-- | @persistent@ handles automatic migrations by creating copies of existing
+-- tables, and re-creating them from scratch. When paired with automatic cascade
+-- deletion, this has dramatic consequences on the wallet data integrity (when a
+-- source table is deleted due do a migration, all foreign tables are also
```
jonathanknowles:
```suggestion
-- source table is deleted due to a migration, all foreign tables are also
```

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -992,6 +1000,27 @@ findNearestPoint wid sl =
...
+-- deleted, but only the rows in the source table are then replaced back!).
```
jonathanknowles:
```suggestion
-- deleted, but only the rows in the source table are replaced!).
```

jonathanknowles (commented):
Hi @KtorZ 

I've had a chance to test your PR, and I was able to successfully start with a wallet that was created with the `v2019-12-13` branch. So far so good! :champagne: 

I do have a couple of questions / concerns though:

1. **The trigger condition for performing a forceful rollback** is that we see a _nonzero_ migration count. However, this count currently only includes _automatic_ migrations. Should we not also include _manual_ migrations in this count? It's conceivable that a successful manual migration will bring the database into a state where no automatic migrations are necessary. But it might be wiser to err on the side of caution and still perform a forceful rollback in this case. (We can always remove this in future, if we're more confident, and have more testing.)

2. **The removal of automatic cascading deletes**. The nice thing about having cascading deletes is that we can rely on the database implementation to ensure referential integrity (no dangling references). Even in the case where Persistent's automatic migration mechanism (erroneously) causes rows to be deleted in tables that reference the checkpoint table, we can still rely on SQLite to preserve referential integrity. However, if we remove this automatic mechanism for ensuring referential integrity, do we not increase our testing burden? Haven't we now made it easier for future code to break referential integrity?

Apart from these concerns, this PR looks good to me. I have a couple of comments regarding Haddock comments and names of types.

lib/core/src/Cardano/DB/Sqlite.hs
KtorZ: I disagree here ^.^
The `Symbol` attached with a Quantity is used as a `unit`. We have several occurrences of this in the source code, like `Quantity "lovelace" ...` or `Quantity "block" ...`, always singular. Same applies here IMO.

lib/core/src/Cardano/DB/Sqlite.hs
KtorZ: Your assumption is correct: there can be manual migrations but no automatic ones. 
Still, the migration count here (and this particular fix) is solely about automatic ones. It comes as a work-around for persistent's behavior of deleting tables combined with the cascading deletion of foreign tables. 
For manual migration, we are in total control of what's being done and, it is very unlikely we would make a manual migration that requires rolling back the chain. If we ever do, then it'd be probably time to revisit this whole migration logic instead of patching up things on top of each other. 
The migration logic has become quite convoluted already.  Manual migrations shouldn't be included because they aren't causing implicit uncontrolled behavior here.

jonathanknowles: @KtorZ wrote:

> For manual migration, we are in total control of what's being done and, it is very unlikely we would make a manual migration that requires rolling back the chain. If we ever do, then it'd be probably time to revisit this whole migration logic instead of patching up things on top of each other.

Fair enough! This is a convincing reason.

lib/core/src/Cardano/DB/Sqlite.hs
paweljakubas: yeah, `Quantity "migration"` is complied with convention adopted in other places

lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -362,6 +367,10 @@ instance ToText DBLog where
             fmt $ ""+||n||+" migrations were applied to the database."
         MsgMigrations (Left err) ->
             "Failed to migrate the database: " <> getMigrationErrorMessage err
+        MsgForcedRollback -> mconcat
+            [ "Some automated database migrations happened. Forced to roll back "
+            , "to genesis."
```
paweljakubas, line 372: **minor suggestion**
```
Some automated database migrations happened. Rollback to the genesis was forced.
```
it is somewhat more natural to me

lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -135,13 +141,15 @@ newDBLayer
...
-            (ManualMigration mempty)
-            migrateAll
-            trace
-            fp
-    ctx@SqliteContext{runQuery} <- handlingPersistError trace fp io
-    return (ctx, DBLayer
+    let start = startSqliteBackend (ManualMigration mempty) migrateAll trace fp
+    (nMigrations, ctx) <- handlingPersistError trace fp start
+    let dbLayer = mkDBLayer ctx
+    when (nMigrations > Quantity 0) $ do
```
paweljakubas, line 147: :+1:

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -992,6 +1000,27 @@ findNearestPoint wid sl =
...
+
+{-------------------------------------------------------------------------------
+                                    Logging
+-------------------------------------------------------------------------------}
+
+-- | @persistent@ handles automatic migrations by creating copies of existing
+-- tables, and re-creating them from scratch. When paired with automatic cascade
+-- deletion, this has dramatic consequences on the wallet data integrity (when a
+-- source table is deleted due to a migration, all foreign tables are also
+-- deleted, but only the rows in the source table are replaced!).
```
paweljakubas, line 1012: :100: for the comment above

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -424,7 +424,15 @@ newDBLayer trace defaultFieldValues mDatabaseFile = do
             selectWallet wid >>= \case
                 Nothing -> pure $ Left $ ErrNoSuchWallet wid
                 Just _  -> Right <$> do
-                    deleteCascadeWhere [WalId ==. wid]
+                    deleteWhere [WalId ==. wid]
```
paweljakubas, line 427: if so then ON DELETE CASCADE in TH.hs are to be deleted

paweljakubas (approved):

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
paweljakubas: and they are

lib/core/src/Cardano/DB/Sqlite.hs
KtorZ: Why `the` however :thinking: ?

paweljakubas: yeah, probably without `the`. so no `to the genesis` but `to genesis`

Issue #1291, opened 2020-01-21
Introduce endpoint scaffolding and needed types to service it


paweljakubas:
# Context

U/S

As an API client, 
I can get blockchain parameters for various epochs,
And I can get blockchain parameters for the latest epoch,
So that I can build better interfaces based on these parameters.

# Decision

We are creating : 
```
paths: 
  /network/parameters/{epoch}:
    get: 
      operationId: getNetworkParameters
      tags: ["Network"]
      summary: Parameters
      parameters: 
        - *parametersEpoch
      responses: *responsesGetNetworkParameters
```

# Acceptance Criteria

1. Create endpoint scaffolding and introduce new types (if necessary to service that)

---

# Development 

# QA 

Tests in #1302

piotr-iohk:
lgtm.

PR #1294, opened 2020-01-21
Blockchain params endpoint - part 1

paweljakubas:
# Issue Number

#1291 

# Overview

- [x] I have added new endpoint scaffolding with needed types
- [x] I have extended swagger
- [x] I have updated unit tests

# Comments


lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -62,6 +62,8 @@ module Cardano.Wallet.Api
 
     -- * Miscellaneous
     , Network
+        , GetNetworkInformation
```
Anviking, line 65: These two imports look misaligned

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -446,6 +466,33 @@ instance FromHttpApiData Iso8601Time where
 instance ToHttpApiData Iso8601Time where
     toUrlPiece = toText
 
+instance FromText ApiEpochNumber where
+    fromText txt = case txt of
+        "latest" -> Right (ApiEpochNumber "latest")
```
Anviking: 🤔 What about defining `ApiEpochNumber` as something like:

```haskell
data ApiEpochNumber = ApiEpochNumberLatest | ApiEpochNumber EpochNumber
```

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -787,6 +800,25 @@ spec = do
...
+                            slotLength (x :: ApiNetworkParameters)
+                    , epochLength =
+                            epochLength (x :: ApiNetworkParameters)
+                    , epochStability =
+                            epochStability (x :: ApiNetworkParameters)
+                    , activeSlotCoefficient =
+                            activeSlotCoefficient (x :: ApiNetworkParameters)
+                    }
+            in
+                x' === x .&&. show x' === show x
```
Anviking, line 851: > "pointless tests to trigger coverage for record accessors"

Hm, ok 👍 

Unrelated to the PR, but do we still care about coverage to this extent? We have removed the badge from the readme at least.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1306,6 +1310,14 @@ getNetworkInformation (_block0, bp, st) nl = do
     unsafeEpochSucc = fromMaybe bomb . W.epochSucc
       where bomb = error "reached final epoch of the Blockchain!?"
 
+getNetworkParameters
+    :: forall t. ()
+    => (Block, BlockchainParameters, SyncTolerance)
+    -> NetworkLayer IO t Block
+    -> ApiEpochNumber
+    -> Handler ApiNetworkParameters
+getNetworkParameters _ _ _ = error "not implemented"
```
Anviking: I think best practice is to `throwError err501` instead of erroring.

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -1171,6 +1203,31 @@ instance Arbitrary ApiNetworkInformation where
...
+instance Arbitrary (Hash "Genesis") where
+    arbitrary = Hash . B8.pack <$> replicateM 32 arbitrary
+
+instance Arbitrary StartTime where
+    arbitrary = StartTime <$> genUniformTime
+
+instance Arbitrary (Quantity "second" NominalDiffTime) where
+    shrink (Quantity 0.0) = []
+    shrink _ = [Quantity 0.0]
+    arbitrary = Quantity . fromInteger <$> choose (0, 100)
```
Anviking: Seems ok, but why so small?

specifications/api/swagger.yaml
```diff
@@ -1732,12 +1800,23 @@ paths:
...
+
+  /network/parameters/{epoch}:
+    get:
+      operationId: getNetworkParameters
+      tags: ["Network"]
+      summary: Parameters
+      parameters:
+        - *parametersEpoch
+      description: |
+        <p align="right">status: <strong>under development</strong></p>
```
Anviking, line 1817: 🆒 👍

Anviking (changes requested):
Looks good! Left a few comments. Most notably I think `ApiEpochNumber` could be less stringly typed.

lib/core/src/Cardano/Wallet/Api/Types.hs
Anviking: It seems like when actually using `ApiEpochNumber`, you'd have to re-parse the `Text` into either the notion of latest or an `EpochNumber` anyway.

Anviking:
CI failing with 
```
Path parameter 'epoch' for 'get' operation in '/network/parameters/{epoch}' was not resolved
```

lib/core/src/Cardano/Wallet/Api.hs
paweljakubas: just followed pattern from the above for other endpoints

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: I would do this if we had not `Handler` but `ExceptT` . When `Handler` then I need to do : 
```
liftHandler $ throwE err501
```
and then we have 
```
No instance for (LiftHandler ServantErr) 
```
So I would rather to stick to error and in replace it in the next PR with implementation

lib/core/src/Cardano/Wallet/Api.hs
Anviking: Ah, I see 👍

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
paweljakubas: ok, increased to 10000

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: ok, but we can use here : 
```
liftIO $ throwIO err501
```

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: I was persuaded and used the propose type. It is more precise and better reflect situation

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
paweljakubas: yeah, `pointless tests` title says it all

paweljakubas:
added also `ApiEpochNumber` to `Api.TypesSpec` and ... found subtle error where `T.decimal` was cycling `Word31` numbers. Adopted proper fix

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: I have a memory of `err501` being used in the wallet, but I don't remember exactly how.

With `liftIO $ throwIO err501` we actually get a 500:
```
http http://localhost:8090/v2/network/parameters/latest
HTTP/1.1 500 Internal Server Error
Content-Type: text/plain; charset=utf-8
Date: Thu, 23 Jan 2020 10:08:45 GMT
Server: Warp/3.2.27
Transfer-Encoding: chunked

Something went wrong
```

Which is the same result as `error "not implemeted"`.

Whereas with `Handler $ throwE err501` we at least get a 501 (but with generic code and message)
```
http http://localhost:8090/v2/network/parameters/latest
HTTP/1.1 501 Not Implemented
Content-Type: application/json;charset=utf-8
Date: Thu, 23 Jan 2020 10:10:33 GMT
Server: Warp/3.2.27
Transfer-Encoding: chunked

{
    "code": "unexpected_error",
    "message": "It looks like something unexpected went wrong. Unfortunately I don't yet know how to handle this type of situation. Here's some information about what happened: "
}
```

paweljakubas: agree. used `Handler $ throwE err501`

Anviking (approved):
lgtm

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -334,6 +341,18 @@ newtype ApiFee = ApiFee
...
+data ApiEpochNumber = ApiEpochNumberLatest | ApiEpochNumber EpochNo
+    deriving (Eq, Generic, Show)
+
+data ApiNetworkParameters = ApiNetworkParameters
+    { genesisBlockHash :: !(ApiT (Hash "Genesis"))
+    , blockchainStartTime :: !(ApiT StartTime)
+    , slotLength :: !(Quantity "second" NominalDiffTime)
+    , epochLength :: !(Quantity "slot" Word32)
+    , epochStability :: !(Quantity "block" Word32)
+    , activeSlotCoefficient :: !(Quantity "percent" Percentage)
```
Anviking, line 353: > Percentage

Though I think you might find in the next PR that `ActiveSlotCoeff` is a `Double` whereas `Percentage` is a `Word, [0, 100]`

Anviking: Ok, U/S says:
> active_slot_coefficient: *percentage

And in `x-percentage` is a `number`. So it should be able to be a floating point.

I imagine some users would like to have the full `33.333333%` instead of just `33%`.

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: in the end used Servant's `throwError err501`

PR #1298, opened 2020-01-22
Log the epoch constants used when ranking stake pools

Anviking:
# Issue Number

#1276 

# Overview

- [x] I added a log for the `Ranking.EpochConstants` when used to sort pools. If the order of pools were to be messed up, this is useful information to have.

# Comments

### DIFF
--- a/lib/core/src/Cardano/Pool/Metrics.hs
+++ b/lib/core/src/Cardano/Pool/Metrics.hs
@@ -563,12 +563,12 @@ instance DefineSeverity StakePoolLog where
         MsgStakePoolRegistration _ -> Info
         MsgRollingBackTo _ -> Info
         MsgUsingRankingEpochConstants ec
-            | (ec ^. #totalRewards) <= Quantity 100
+            | oddlySmallRewards ec
                 -> Notice
             | otherwise
                 -> Debug
         MsgUsingTotalStakeForRanking s
-            | s <= Quantity 100
+            | oddlySmallTotalStake s
                 -> Notice
             | otherwise
                 -> Debug
@@ -609,10 +609,25 @@ instance ToText StakePoolLog where
             "Writing stake-distribution for epoch " <> pretty ep
         MsgStakePoolRegistration pool ->
             "Discovered stake pool registration: " <> pretty pool
-        MsgUsingRankingEpochConstants csts ->
-            "Using ranking epoch constants: " <> pretty csts
-        MsgUsingTotalStakeForRanking s ->
-            "The total stake used to determine epoch constants was: " <> pretty s
+        MsgUsingRankingEpochConstants ec
+            | oddlySmallRewards ec -> mconcat
+                [ "The total rewards for this epoch are oddly small. "
+                , "This may cause stake pool ranking to be unreliable. "
+                , "Epoch constants are: "
+                , pretty ec
+                ]
+            | otherwise ->
+                "Using ranking epoch constants: " <> pretty ec
+        MsgUsingTotalStakeForRanking s
+            | oddlySmallTotalStake s -> mconcat
+                [ "The total stake is oddly small ("
+                , pretty s
+                , "). This may result in unreliable epoch constants."
+                ]
+            | otherwise -> mconcat
+                [ "The total stake used to determine epoch constants was: "
+                , pretty s
+                ]
         MsgRollingBackTo point ->
             "Rolling back to " <> pretty point
         MsgApplyError e -> case e of
@@ -624,3 +639,9 @@ instance ToText StakePoolLog where
                 "Race condition when fetching stake distribution."
             ErrMonitorStakePoolsPoolAlreadyExists{} ->
                 ""
+
+oddlySmallRewards :: EpochConstants -> Bool
+oddlySmallRewards ec = (ec ^. #totalRewards) <= Quantity 100
+
+oddlySmallTotalStake :: Quantity "lovelace" Word64 -> Bool
+oddlySmallTotalStake = (<= Quantity 100)
--- a/lib/core/src/Cardano/Pool/Metrics.hs
+++ b/lib/core/src/Cardano/Pool/Metrics.hs
@@ -288,13 +288,13 @@ newStakePoolLayer tr getEpCst db@DBLayer{..} nl metadataDir = StakePoolLayer
         then do
             seed <- liftIO $ atomically readSystemSeed
             let epCst = getEpCst 0
-            combineWith epCst (sortArbitrarily seed) distr prod mempty
+            combineWith (sortArbitrarily seed) epCst distr prod mempty
 
         else do
             let currentEpoch = prodTip ^. #slotId . #epochNumber
             perfs <- liftIO $ readPoolsPerformances db currentEpoch
             let epCst = getEpCst currentEpoch
-            combineWith epCst (pure . sortByDesirability) distr prod perfs
+            combineWith (pure . sortByDesirability) epCst distr prod perfs
 
     readPoolProductionTip = readPoolProductionCursor 1 <&> \case
         []  -> header block0
@@ -321,13 +321,16 @@ newStakePoolLayer tr getEpCst db@DBLayer{..} nl metadataDir = StakePoolLayer
     (block0, _) = staticBlockchainParameters nl
 
     combineWith
-        :: (Quantity "lovelace" Word64 -> EpochConstants)
-        -> ([(StakePool, [PoolOwner])] -> IO [(StakePool, [PoolOwner])])
+        :: ([(StakePool, [PoolOwner])] -> IO [(StakePool, [PoolOwner])])
+        -> (Quantity "lovelace" Word64 -> EpochConstants)
         -> Map PoolId (Quantity "lovelace" Word64)
         -> Map PoolId (Quantity "block" Word64)
         -> Map PoolId Double
         -> ExceptT ErrListStakePools IO [(StakePool, [PoolOwner])]
-    combineWith epCst sortResults distr prod perfs = do
+    combineWith sortResults mkEpCst distr prod perfs = do
+        liftIO $ do
+            traceWith tr $ MsgUsingTotalStakeForRanking totalStake
+            traceWith tr $ MsgUsingRankingEpochConstants epConstants
         case combineMetrics distr prod perfs of
             Left e ->
                 throwE $ ErrListStakePoolsMetricsInconsistency e
@@ -344,6 +347,8 @@ newStakePoolLayer tr getEpCst db@DBLayer{..} nl metadataDir = StakePoolLayer
         totalStake =
             Quantity $ Map.foldl' (\a (Quantity b) -> a + b) 0 distr
 
+        epConstants = mkEpCst totalStake
+
         mergeRegistration poolId (stake, production, performance) =
             fmap mkStakePool <$> readPoolRegistration poolId
           where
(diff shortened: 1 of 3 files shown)


lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -321,13 +321,14 @@ newStakePoolLayer tr getEpCst db@DBLayer{..} nl metadataDir = StakePoolLayer
     (block0, _) = staticBlockchainParameters nl
 
     combineWith
-        :: (Quantity "lovelace" Word64 -> EpochConstants)
-        -> ([(StakePool, [PoolOwner])] -> IO [(StakePool, [PoolOwner])])
+        :: ([(StakePool, [PoolOwner])] -> IO [(StakePool, [PoolOwner])])
+        -> (Quantity "lovelace" Word64 -> EpochConstants)
```
Anviking, line 325: Diff here and above is just changing the order of the arguments to `combineWith`. Not needed, but think the sorting argument were originally meant to be the first argument, after "With".

lib/core/src/Cardano/Pool/Ranking.hs
```diff
@@ -181,6 +183,16 @@ data EpochConstants = EpochConstants
...
 
+instance Buildable EpochConstants where
+    build ec = blockListF' "" id
+        [ "leaderStakeInfluence (a0) =  "
+            <> build (getNonNegative $ leaderStakeInfluence ec)
+        , "desiredNumberOfPool (k) = "
+            <> build (getPositive $ desiredNumberOfPools ec)
+        , "totalRewards (R) = "
+            <> build (getQuantity $ totalRewards ec)
+        ]
```
Anviking, line 194: Running integration tests (with temporary tweaked R):
```
[cardano-wallet.pools-engine:Info:114] [2020-01-22 11:46:29.32 UTC] Downloading http://localhost:49177/test-integration-registry.zip -> /private/var/folders/h_/mlzym27d5lz_fx3ph2d547700000gn/T/stake-pool-metadata-564df08caee116ef/registry.zip: Using cached file last modified at 2020-01-22 11:45:27.974117167 UTC
[cardano-wallet.api-server:Info:114] [2020-01-22 11:46:29.32 UTC] [RequestId 601] 200 OK in 0.007592s
[cardano-wallet.api-server:Info:114] [2020-01-22 11:46:29.82 UTC] [RequestId 602] [GET] /v2/stake-pools
[cardano-wallet.pools-engine:Info:114] [2020-01-22 11:46:29.83 UTC] Using ranking epoch constants:  leaderStakeInfluence (a0) =  0.0
 desiredNumberOfPool (k) = 10
 totalRewards (R) = 10000

[cardano-wallet.pools-engine:Info:114] [2020-01-22 11:46:29.83 UTC] Downloading http://localhost:49177/test-integration-registry.zip -> /private/var/folders/h_/mlzym27d5lz_fx3ph2d547700000gn/T/stake-pool-metadata-564df08caee116ef/registry.zip: Using cached file last modified at 2020-01-22 11:45:27.974117167 UTC
[cardano-wallet.api-server:Info:114] [2020-01-22 11:46:29.83 UTC] [RequestId 602] 200 OK in 0.008747s
[cardano-wallet.api-server:Info:114] [2020-01-22 11:46:30.33 UTC] [RequestId 603] [GET] /v2/stake-pools
[cardano-wallet.pools-engine:Info:114] [2020-01-22 11:46:30.34 UTC] Using ranking epoch constants:  leaderStakeInfluence (a0) =  0.0
 desiredNumberOfPool (k) = 10
 totalRewards (R) = 10000
```

lib/core/src/Data/Quantity.hs
```diff
@@ -120,6 +122,12 @@ instance FromText b => FromText (Quantity sym b) where
 instance ToText b => ToText (Quantity sym b) where
     toText (Quantity b) = toText b
 
+-- Builds (Quantity "lovelace" Word64) as "42 lovelace"
+instance (KnownSymbol unit, Buildable a) => Buildable (Quantity unit a) where
+    build (Quantity a) = build a <> fmt " " <> build u
+      where
+        u = symbolVal (Proxy :: Proxy unit)
```
Anviking, line 129: Think it is ok to add this general instance.

lib/core/src/Cardano/Pool/Metrics.hs
KtorZ: Side-note: this comment **should** really be in your commit message. Message of commits tell about the why. The commit itself tells the what :+1:

lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -538,6 +541,7 @@ data StakePoolLog
     | MsgStakePoolRegistration PoolRegistrationCertificate
     | MsgRollingBackTo SlotId
     | MsgApplyError ErrMonitorStakePools
+    | MsgUsingRankingEpochConstants Ranking.EpochConstants
```
KtorZ: The qualified import isn't needed. Please refer to: https://github.com/input-output-hk/cardano-wallet/wiki/Coding-Standards#we-use-explicit-imports-by-default-and-favor-qualified-imports-for-ambiguous-functions

lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -555,6 +559,7 @@ instance DefineSeverity StakePoolLog where
         MsgStakeDistribution _ -> Info
         MsgStakePoolRegistration _ -> Info
         MsgRollingBackTo _ -> Info
+        MsgUsingRankingEpochConstants _ -> Info
```
KtorZ: This ought to be `Debug`, no ?

lib/core/src/Cardano/Pool/Ranking.hs
KtorZ: This would look even better with an extra newline before the leaderStakeInfluence, don't you think ?

lib/core/src/Cardano/Pool/Metrics.hs
Anviking: I should make it a habit to review the diffs locally first, where I can easily edit the commits, but the GH-diff is convenient.

lib/core/src/Cardano/Pool/Metrics.hs
Anviking: Hm… you're probably right

lib/core/src/Cardano/Pool/Metrics.hs
KtorZ: You can always easily edit the commits after opening a PR you know :)

KtorZ: I agree that the GH view is convenient. I often do the same.

Anviking: > You can always easily edit the commits after opening a PR you know :)
 
Yes, but on GH you can comment without "breaking the flow" (while going through the entire diff). Moving to your editor, finding the relevant source or commit is a context switch, especially when you haven't made it a habit or have setup your tools in a good way.

lib/core/src/Cardano/Pool/Metrics.hs
Anviking: I wonder if it would be wrong to increase it to `Notice` only when `R == 0` and `Debug` otherwise.

lib/core/src/Cardano/Pool/Metrics.hs
Anviking: The thing is, I _do_ to some extent view `EpochConstants` as ambiguous and requiring context — `Ranking.EpochConstants` — in a similar way to `Map.filter`.

`EpochConstants` are not _all_ epoch-specific constants, but precisely the ones that are needed for `Ranking`. 

(Although,  yes, we actually only have one type called `EpochConstants`.)

Would you still prefer un-qualified?

lib/core/src/Cardano/Pool/Metrics.hs
Anviking: I didn't do that. I think it would be neat, but I want to notice when the stake is 0 too. So I did add two new NOTICE traces: ff3bf9b373778dfd1e9f57c33f61e18d57a46c45 for when `totalStake` and `totalRewards` are 0

lib/core/src/Cardano/Pool/Metrics.hs
KtorZ: I disagree about the ambiguity here. There's nothing ambiguous about `EpochConstants` here, especially since this identifier only exists once in the entire codebase. On the contrary, if you look at the Registry module for instance, most functions are ambiguous there (`insert`, `remove` etc ...). 

If you think `EpochConstants` lack too much context, then perhaps change the name of the type to `RankingEpochConstants` (although I truly think it's overkill here). 

> Would you still prefer un-qualified?

Yes.

lib/core/src/Cardano/Pool/Metrics.hs
KtorZ: Good idea (modulo the comment about where to put the logic regarding the interpretation of logging). 
Still, this one should really be Debug.

Anviking: now debug

lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -555,6 +563,12 @@ instance DefineSeverity StakePoolLog where
         MsgStakeDistribution _ -> Info
         MsgStakePoolRegistration _ -> Info
         MsgRollingBackTo _ -> Info
+        MsgUsingRankingEpochConstants ec
+            | (ec ^. #totalRewards) == Quantity 0
+                -> Notice
```
Anviking, line 572: From the logging [guidelines](https://github.com/input-output-hk/cardano-wallet/wiki/Logging-Guidelines)
> NOTICE -> needs attention; something not progressing properly
> WARNING -> may continue into an error condition if continued

I am not sure this is really a warning. If `R` is zero, there will be no error. The pools will just be ordered arbitrarily/randomly. 

It might be expected that `R` is zero (… can be debated).

It might be fine that the total stake is zero, ignoring the fact that I think no blocks will be produced then. But that's not the concern of this module.

KtorZ: Fair enough.

Anviking:
`Debug` doesn't appear in the terminal output, nor in production logs I think.

This means that these added traces are only appear when they are labeled `Notice`.

A `== 0` condition for Notice would have helped neither for `totalRewards`, nor for `totalStake` to catch the problem of our integration tests having to small stake to have pools sorted properly.

I made the Notice-trigger `< 100 lovelace` instead. This is an arbitrary number. I don't think there is any hard limit we can draw. 

I don't see any cons of treating `< 100 lovelace` arbitrarily as notice, and I hope it could be helpful.

What do you think, @KtorZ?

Now we see:

```
[cardano-wallet.pools-engine:Notice:40] [2020-01-23 13:49:01.26 UTC] The total stake used to determine epoch constants was: 3 lovelace
[cardano-wallet.pools-engine:Notice:40] [2020-01-23 13:49:01.26 UTC] Using ranking epoch constants:
 leaderStakeInfluence (a0) =  0.0
 desiredNumberOfPool (k) = 10
 totalRewards (R) = -0
```
without #1299

lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -592,6 +609,10 @@ instance ToText StakePoolLog where
             "Writing stake-distribution for epoch " <> pretty ep
         MsgStakePoolRegistration pool ->
             "Discovered stake pool registration: " <> pretty pool
+        MsgUsingRankingEpochConstants csts ->
+            "Using ranking epoch constants: " <> pretty csts
+        MsgUsingTotalStakeForRanking s ->
+            "The total stake used to determine epoch constants was: " <> pretty s
```
KtorZ: One last note: this message is displayed at a Notice level when the stake is low. Yet, when reading this message, it's hard to understand _why_ this is a notice and what's wrong with that. What about rephrasing it to something explicit like:

```hs
"The total stake is oddly small (" <> pretty s <> "). This may result in unreliable epoch constants."
```

And, same about the above actually. 

```hs
"The total reward for this is oddly small. This may cause pool's ranking to be unreliable. Epoch constants are:  " <> pretty csts
```

KtorZ (approved):
LGTM, modulo a small revision on the log message.

Issue #1292, opened 2020-01-21
Cannot list wallets when using same state directory but different genesis


piotr-iohk:
# Context

https://github.com/input-output-hk/cardano-wallet/issues/1281#issuecomment-576636085

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | 2020.1.14 (git revision: 5e690157e636532fa27de88622aad8f85b7919e5) |
| Platform     | Linux (all?)        |
| Installation | From source code            |

# Steps to Reproduce

 - Create a standalone private blockchain (e.g. using the integration config).
 - Turn on the wallet server and wait for a few slots / epochs.
 - **Create a wallet**
 - Turn the wallet off.
 - Change a setting in the genesis configuration and re-generate the genesis block (effectively creating another blockchain).
 - Restart the server on the same state directory.

In the log there is:
```
[cardano-wallet.wallet-engine:Error:29] [2020-01-21 11:04:31.30 UTC] Worker has exited unexpectedly: ErrCheckIntegrityDifferentGenesis (Hash {getHash = "\216\163\131S\189W\b]\213+\139U\169Ue\205W\215\158\SI\131}\151ed\ACK\250\186O~\233\&7"}) (Hash {getHash = "<\a\ETX\SO6\191\255\230~..\192\158R\147\211\132c|\210\240\EOT5n\243 \243\254lR\EOT\SUB"})
```
and then:

## Expected behavior

```
$ cardano-wallet-jormungandr wallet list
Ok.
[]
```

## Actual behavior
 
```
$ cardano-wallet-jormungandr wallet list
Something went wrong
```

---

# Resolution 

---

# QA

jonathanknowles:
Confirmed on `v2020-01-21` (d188a5fcf73d26f51391c5aea5030dc8e152741d).

KtorZ:
@jonathanknowles I did some investigation yesterday. The main issue is due to the worker not being unregistered from the registry when it dies.

jonathanknowles:
@KtorZ wrote:
> @jonathanknowles I did some investigation yesterday. The main issue is due to the worker not being unregistered from the registry when it dies.

Yes, I found the same thing in my investigation.

But there's something else that's going on. The current implementation of `newWorker` returns a `Worker` **even if** evaluating the `before` action fails:

[...]

PR #1301, opened 2020-01-23
Simplify worker registration lifecycle.

jonathanknowles:
# Issue Number

#1292 

# Overview

This PR adjusts the worker registration life cycle so that creating and registering a worker is now a single operation, rather than a two-step operation.

Whereas previously callers would need to first call `newWorker` followed by `insert`, they now only need to call `register`.

In addition, we now provide automatic garbage collection of stale worker references from the registry, so that attempting to look up a previously-registered but prematurely-terminated thread will result in `Nothing`.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Registry.hs
+++ b/lib/core/src/Cardano/Wallet/Registry.hs
@@ -58,6 +58,8 @@ import Control.Exception
     , asyncExceptionFromException
     , finally
     )
+import Control.Monad
+    ( void )
 import Control.Monad.IO.Class
     ( MonadIO, liftIO )
 import Control.Tracer
@@ -240,11 +242,10 @@ register registry ctx k (MkWorker before main after acquire) = do
                 }
         registry `insert` worker
         return worker
-    cleanup mvar e = finally
-        (registry `unregister` k)
-        (finally
-            (tryPutMVar mvar Nothing)
-            (after tr e))
+    cleanup mvar e = do
+        registry `unregister` k
+        void $ tryPutMVar mvar Nothing
+        after tr e
 
 {-------------------------------------------------------------------------------
                                     Logging
--- a/lib/core/src/Cardano/Wallet/Registry.hs
+++ b/lib/core/src/Cardano/Wallet/Registry.hs
@@ -13,8 +13,8 @@ module Cardano.Wallet.Registry
     , empty
     , keys
     , lookup
-    , remove
     , register
+    , unregister
 
       -- * Worker
     , Worker
@@ -64,6 +64,8 @@ import Control.Monad.IO.Class
     ( MonadIO, liftIO )
 import Control.Tracer
     ( Tracer, traceWith )
+import Data.Foldable
+    ( traverse_ )
 import Data.Generics.Internal.VL.Lens
     ( (^.) )
 import Data.Generics.Labels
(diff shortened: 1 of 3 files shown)


lib/core/src/Cardano/Wallet/Registry.hs
```diff
@@ -137,6 +136,16 @@ insert
...
 
+-- | Unregister a worker from the registry, but don't cancel the running task.
+--
+unregister
+    :: Ord key
+    => WorkerRegistry key resource
+    -> key
+    -> IO ()
+unregister (WorkerRegistry mvar) k =
+    modifyMVar_ mvar (pure . Map.delete k)
```
KtorZ, line 152: Having both `unregister` and `remove` makes the module API somewhat non-orthogonal. What would you think of revising a bit the interface here by: 

- Keeping only one of the two, likely (`unregister` as it sounds more dual to `register`)
- Having `unregister` returns the actual `Worker key resource` when removed from the registry 
- Let the caller of `unregister` kill the worker thread itself

So, on call sites, instead of:

```hs
    liftHandler $ withWorkerCtx @_ @s @k ctx wid throwE $ \_ -> pure ()
    liftIO $ Registry.remove re wid
```

We would have:

```hs
    liftIO $ Registry.unregister re wid >>= \case
        Nothing  -> liftHandler $ throwE $ ErrNoSuchWallet wid
        Just wrk -> liftIO $ killThread $ workerThread wrk
```

which actually feels way less awkward that the artificial 'withWorkerCtx' call and make the whole thing much more explicit.

lib/core/src/Cardano/Wallet/Registry.hs
```diff
@@ -190,44 +200,51 @@ defaultWorkerAfter tr = traceWith tr . \case
...
-        ( key ~ WorkerKey ctx
+        ( Ord key
+        , key ~ WorkerKey ctx
         , msg ~ WorkerMsg ctx
         , HasLogger (WorkerLog key msg) ctx
         , HasWorkerCtx resource ctx
         )
     => ctx
     -> key
+    -> WorkerRegistry key resource
```
KtorZ: Side-note, all function working with the registry takes the registry as first argument. It'd be more consistent to also do this for this new function.

lib/core/src/Cardano/Wallet/Registry.hs
```diff
@@ -190,44 +200,51 @@ defaultWorkerAfter tr = traceWith tr . \case
...
+                , workerThread = threadId
+                , workerResource = resource
+                }
+        registry `insert` worker
+        return worker
+    cleanup mvar e = finally
+        (registry `unregister` k)
+        (finally
+            (tryPutMVar mvar Nothing)
+            (after tr e))
```
KtorZ: I do not understand the use of both `finally` here :thinking: ? These calls (`unregister` & `trryPutMVar`) are safe / don't throw, do there's no reason to wrap them in `finally` clauses, unless I am missing something?
The `cleanup` could be simply written:

```hs
cleanup mvar e = do
    registry `unregister` k
    tryPutMVar mvar Nothing
    after tr e
```

lib/core/src/Cardano/Wallet/Registry.hs
jonathanknowles: @KtorZ wrote:
> Side-note, all function working with the registry takes the registry as first argument. It'd be more consistent to also do this for this new function.

Fixed in commit af7722992e91af4d8e8ccf5cc8826f8748cc6949.

lib/core/src/Cardano/Wallet/Registry.hs
jonathanknowles: @KtorZ wrote:
> I do not understand the use of both `finally` here thinking ? These calls (`unregister` & `trryPutMVar`) are safe / don't throw, do there's no reason to wrap them in `finally` clauses, unless I am missing something?

Thanks for spotting this. I think this was just a result of my paranoia while testing.

Fixed in commit 3bae4916ac3a26b756e70e7602d09a09f2178818.

lib/core/src/Cardano/Wallet/Registry.hs
jonathanknowles: @KtorZ I like the general idea, though I have one suggestion.

Since the registry is responsible for handling thread creation, perhaps it should also be responsible for thread destruction?

How about the following (names can be changed):
```hs
-- Exported:
unregister
    :: Ord key
    => WorkerRegistry key resource
    -> key
    -> IO ()  
unregister registry k =
    delete registry k >>= traverse_ (killThread . workerThread)

-- Internal operation, not exported:
delete
    :: Ord key
    => WorkerRegistry key resource
    -> key
    -> IO (Maybe (Worker key resource))   
delete (WorkerRegistry mvar) k = do   
    mWorker <- Map.lookup k <$> readMVar mvar 
    modifyMVar_ mvar (pure . Map.delete k)
    pure mWorker 
```

Then, `deleteWallet` can be:
```hs
deleteWallet ctx (ApiT wid) = 
   liftHandler $ withWorkerCtx @_ @s @k ctx wid throwE $ \_ -> do 
      liftIO $ Registry.unregister re wid  
      liftIO $ removeDatabase df wid  
      return NoContent   
```

jonathanknowles: @KtorZ 

I've revised the interface of `Registry` so that `register` and `unregister` are **dual** operations of one another.

* `register` is responsible for:
    1. adding an entry into the registry map.
    2. thread creation.
* `unregister` is responsible for:
    1. removing an entry from the registry map.
    2. thread destruction.

See commit f0d5cbd612832e2757c7f8105d81a71362a958c9.

Anviking (approved):
Lgtm!

> In addition, we now provide automatic garbage collection of stale worker references from the registry, so that attempting to look up a previously-registered but prematurely-terminated thread will result in Nothing.

So if a worker terminates, it will be removed from the in-memory registry?

`listWallets` uses the in-memory Registry keys as source. If a worker *were* to terminate, the wallet would then go missing in Daedalus. Is that the most desirable behaviour? 🤔

jonathanknowles:
> So if a worker terminates, it will be removed from the in-memory registry?

That is correct.

> listWallets uses the in-memory Registry keys as source. If a worker were to terminate, the wallet would then go missing in Daedalus. Is that the most desirable behaviour? :thinking: 

That is a very good point. I'll make a note on issue #1292 about this.

PR #1303, opened 2020-01-24
Byron-Rewrite node's integration

KtorZ:
# Issue Number

# Overview

- efe7fcd18ec71de51705328c72d9c34a22f0193a
  Initial NetworkLayer implementation for Byron
  
- 53b074e1397dc274658cc69e532967d32403ee2b
  rename 'networkTip' to 'currentNodeTip' in NetworkLayer
  This name was misleading. What 'networkTip' actually returns is the tip of the local
chain of the underlying node. The network tip can be known statically from the blockchain
start date.

- 6f2aee621235bcaf38a71c74b3c8f2dc57c4b069
  remove 'findIntersection' from top-level NetworkLayer interface
  And move it only to tests, where it's used

- 0d0a62eefab83d8161356c9cee5f023db9b79ef5
  make 'send' timeout when no response is given in a timely manner
  This avoids blocking potentially indefinitely if the Network Client
somehow dies at the other end and the connection isn't cleaned up
correctly.

- baa4106bbddadcd935e9a58e37bff075c7646037
  Support multiple node connections through the network layer
  
- fbab14ea967402e6f4a2875c5eca4b2a52309504
  add note about retrying in the chain follower
  
- 574c3a5525dcf34806d567cac732ad89207c27a8
  remove unused 'TxWitness' data-type
  
- 67028cd47c51c1572f6614a3d385d12788b7a327
  re-implement base of transaction layer for Byron
  
- fa55babee093765a8c8406411233b0a3bf0b5a56
  remove unused 'dummyAddress' and 'dummyKey'
  
- 1ae275939037664fbe31dc7a53fdb98c415b309b
  disable shelley-specific endpoint in Byron server
  Until the address format is stabilized on the Haskell side, we will keep this one off

- 5bd164a28417d17b89084d6e3e90b9105d2a77fc
  re-generate nix machinery

# Comments

@Anviking So, this is mostly "complete" now. We will need to re-introduce the transaction submission in the legacy API (potentially?) and also, enable / resurrect integration tests and actually all other tests regarding the byron stuff. Will do in some other PR(s). 

I've cleaned up a few things along the way, some remainders of the http-bridge era.


lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -0,0 +1,611 @@
...
+    , getAccountBalance = _getAccountBalance
+    }
+  where
+    _initCursor headers = do
+        let points = genesisPoint : (toPoint <$> headers)
+        queue `send` CmdFindIntersection points >>= \case
+            Left e  -> throw e
+            Right intersection -> pure $ Cursor intersection
+
+    _nextBlocks _ = do
```
Anviking: The `_` would mean:
- `nextBlocks t >> nextBlocks t` doesn't work as you'd expect
- `RetryImmediately :: FollowAction` will silently not work as expected when passed to `follow`

🤷‍♂

lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -106,24 +106,18 @@ data NetworkLayer m target block = NetworkLayer
...
         -- can be used to fetch blocks.
 
     , cursorSlotId
         :: Cursor target -> SlotId
         -- ^ Get the slot corresponding to a cursor.
 
-    , networkTip
-        :: ExceptT ErrNetworkTip m BlockHeader
-        -- ^ Get the current network tip from the chain producer
+    , currentNodeTip
```
Anviking, line 118: Nice, you renamed it 👍 

Wanted to say though:
- We seem to use `currentNodeTip` only for:
    - `GetNetworkInformation`
    -  `monitorStakePools` (in a jörmungandr influenced way)
- In this PR `currentNodeTip` for Byron seem to return the  **current point** of the in-progress chain-sync, right?

> have one network client per wallet (there's one global network layer, but that can't work with the stateful haskell nodes).

What network-layer will you use for the `NetworkInformation`-endpoint?

I wonder if `currentNodeTip` can not be removed or implemented in another way than the chain-sync protocol.

KtorZ: `currentNodeTip` in the byron layer returns the tip of the underlying node. Which is what we want. The cursor tells you where you're at.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
KtorZ: I am not sure to follow you on this ?

KtorZ: Ah, yes. The protocol is stateful now indeed. So using the same cursor twice will have an unexpected behavior. I'll see if we can handle that properly.

lib/core/src/Cardano/Wallet/Network.hs
Anviking: Hm… Ah! You're right.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
KtorZ: Added a note.

KtorZ:
bors try

lib/byron/src/Cardano/Wallet/Byron/Network.hs
Anviking: It feels much better to me now when you put the queue inside the `Cursor` 👍 

One can look at the cursor, see that there is some kind of state there, and not be surprised.

Also, `initCursor` will always create a new node-connection & cursor. This is neat. Had the state been in the NetworkLayer, we could have accidentally messed up an ongoing chain-sync by calling `initCursor` in the middle of it (which would have changed the node's notion of where the wallet's tip). I imagine at least.

Anviking:
bors try

lib/byron/src/Cardano/Wallet/Byron/Compatibility.hs
```diff
@@ -0,0 +1,356 @@
...
+        , headerHash = fromByronHash $ Point.blockPointHash blk
+        -- TODO
+        -- We only use the parentHeaderHash in the
+        -- 'Cardano.Wallet.Network.BlockHeaders' chain follower only required for
+        -- Jörmungandr, this is therefore useless to have in 'normal' BlockHeader
+        --
+        -- Yet, since we also serialize these to the database, this requires
+        -- some non-trivial changes. Not fixing this right now is also a
+        -- possibility.
+        , parentHeaderHash = W.Hash "parentHeaderHash - unused in Byron"
```
Anviking, line 355: I think it would be cleaner if nothing but the DB Layer / glue code knew about `parentHeaderHash`.

(_note:_ very optimally)

Anviking:
bors try

lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -0,0 +1,587 @@
...
+    , initCursor = _initCursor
+    , cursorSlotId = _cursorSlotId
+    , postTx = _postTx
+    , staticBlockchainParameters = _staticBlockchainParameters
+    , stakeDistribution = _stakeDistribution
+    , getAccountBalance = _getAccountBalance
+    }
+  where
+    _initCursor headers = do
+        queue <- atomically newTQueue
```
Anviking, line 204: `follow` now returns on each (non-trivial) rollback. If I understand things correctly, a new connection to the node will be established, and the intersection negotiated each time.

This could be often. Are the connections closed automatically?

It sounds saner to me to not return on rollback if possible.

lib/byron/src/Cardano/Wallet/Byron/Compatibility.hs
```diff
@@ -0,0 +1,356 @@
...
+toSlotNo :: W.SlotId -> SlotNo
+toSlotNo =
+    SlotNo . W.flatSlot byronEpochLength
+
+fromByronBlock :: W.Hash "Genesis" -> ByronBlock -> W.Block
+fromByronBlock genesisHash byronBlk = case byronBlockRaw byronBlk of
+  ABOBBlock blk  ->
+    mkBlock $ fromTxAux <$> unTxPayload (blockTxPayload blk)
+  ABOBBoundary _ ->
+    mkBlock []
```
Anviking, line 274: Does our DB work well with boundary blocks? (Multiple blocks in the same slot I believe)

Anviking (approved):
Looks good to me.

A neat aspect of this PR is that it doesn't have to refactor core abstractions and risk affecting jormungandr. I.e. low-risk to merge now.

But on the other hand, I am concerned that we're piling on un-intrinsic complexity to the Network-code, that will make the Byron-integration more difficult / bug-prone.

To me it seems we should have a  **thin** and *well-tested* (as needed) glue between the wallet and the Haskell ChainSyncProtocol that is neat and simple to understand. This should be a matter of re-evaluating existing abstractions and the boundary between _core_ and _cardano-wallet-jormungandr_.

I am [still](https://github.com/input-output-hk/cardano-wallet/issues/1161#issuecomment-573010700) curious about having something more like:
```Haskell
follow 
    :: [point] -- points/BlockHeaders we know about
    -> (RollBackOrForward block -> IO ()) 
    -> IO () -- syncs forever
```
modulo that re-negotiation of the intersection would be needed with our current sparse checkpoint strategy.

Anyhow, merging… if CI finally allows.
bors r+

Anviking:
I was not able to reproduce locally. I thought this was a flaky test. But it is here failing reliably 🤔

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Issue #1302, opened 2020-01-23
Implement blockchain parameters endpoint


paweljakubas:
# Context

WB-47

As an API client,
I can get blockchain parameters for various epochs,
And I can get blockchain parameters for the latest epoch,
So that I can build better interfaces based on these parameters.

# Decision

Implement already added endpoint realizing the above goal 

# Acceptance Criteria

1. The endpoint's implemention builds on previous PR and satisfies : 
- The API must allow fetching blockchain parameters for any given epoch from a new endpoint
- The keyword "latest" may be used to refer to the latest epoch (from the perspective of the network)
2. All error specified in the task description are implemented: 
```
400 # When the epoch parameter is malformed
404 # When the epoch is valid doesn't exist
405 # When the HTTP method isn't "GET"
406 # When the client doesn't Accept JSON
```
3. The endpoint is thoroughly  tested against expected errors and valid requests. 

---

# Development 

# QA 
See added tests, confront it with requirements and add more if needed

The endpoint works like below which is worth testing more:
```
[pawel@arch ~]$ curl http://localhost:8090/v2/network/parameters/2000000 --http1.1
{"code":"not_such_epoch_no","message":"I couldn't show blockchain parameters for epoch number later than current one. You requested 2000000 epoch. Current one is 1200581. Use smaller epoch than current or 'latest'."}

[pawel@arch ~]$ curl http://localhost:8090/v2/network/parameters/aaaa --http1.1
{"code":"invalid_epoch_no","message":"I couldn't show blockchain parameters for aaaa. It should be either 'latest' or integer from 0 to 2147483647."}

[pawel@arch ~]$ curl http://localhost:8090/v2/network/parameters/latest --http1.1
{"slot_length":{"quantity":2,"unit":"second"},"epoch_stability":{"quantity":5,"unit":"block"},"genesis_block_hash":"f8c0622ea4b768421fea136a6e5a4e3b4c328fc5f16fad75817e40c8a2a56a56","blockchain_start_time":"2019-04-25T14:20:57Z","epoch_length":{"quantity":10,"unit":"slot"},"active_slot_coefficient":{"quantity":100,"unit":"percent"}}

[pawel@arch ~]$ curl http://localhost:8090/v2/network/parameters/200 --http1.1
[...]

PR #1306, opened 2020-01-27
Blockchain params endpoint impl

paweljakubas:
# Issue Number

#1302 

# Overview

- [x] I have refactored ApiEpochNumber to handle wrong param path later
- [x] I have handled err 400 - invalid path param 
- [x] I have handled err 405 and 406
- [x] I have handled epoch does not exist error - 404
- [x] I have added impl of the GET with corresponding tests 
- [x] I have revised added code from the point of reusability

# Comments

Works as in specification to the task in JIRA:
```
[pawel@arch ~]$ curl http://localhost:8090/v2/network/parameters/2000000 --http1.1
{"code":"not_such_epoch_no","message":"I couldn't show blockchain parameters for epoch number later than current one. You requested 2000000 epoch. Current one is 1200581. Use smaller epoch than current or 'latest'."}

[pawel@arch ~]$ curl http://localhost:8090/v2/network/parameters/aaaa --http1.1
{"code":"invalid_epoch_no","message":"I couldn't show blockchain parameters for aaaa. It should be either 'latest' or integer from 0 to 2147483647."}

[pawel@arch ~]$ curl http://localhost:8090/v2/network/parameters/latest --http1.1
{"slot_length":{"quantity":2,"unit":"second"},"epoch_stability":{"quantity":5,"unit":"block"},"genesis_block_hash":"f8c0622ea4b768421fea136a6e5a4e3b4c328fc5f16fad75817e40c8a2a56a56","blockchain_start_time":"2019-04-25T14:20:57Z","epoch_length":{"quantity":10,"unit":"slot"},"active_slot_coefficient":{"quantity":100,"unit":"percent"}}

[pawel@arch ~]$ curl http://localhost:8090/v2/network/parameters/200 --http1.1
{"slot_length":{"quantity":2,"unit":"second"},"epoch_stability":{"quantity":5,"unit":"block"},"genesis_block_hash":"f8c0622ea4b768421fea136a6e5a4e3b4c328fc5f16fad75817e40c8a2a56a56","blockchain_start_time":"2019-04-25T14:20:57Z","epoch_length":{"quantity":10,"unit":"slot"},"active_slot_coefficient":{"quantity":100,"unit":"percent"}}
```


paweljakubas:
bors try

lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
```diff
@@ -152,3 +157,71 @@ spec = do
...
+
+    describe "NETWORK - HTTP headers" $ do
+        forM_ (getHeaderCases HTTP.status200)
+            $ \(title, headers, expectations) -> it title $ \ctx -> do
+                r <- request @ApiNetworkParameters ctx
+                    Link.getNetworkInfo headers Empty
+                verify r expectations
+
+    it "NETWORK - Cannot query blockchain parameters with epoch later \
+       \than current one" $ \ctx -> do
```
Anviking, line 192: 👍

lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
```diff
@@ -152,3 +157,71 @@ spec = do
...
+        r1 <- request @ApiNetworkInformation ctx
+            Link.getNetworkInfo Default Empty
+        let (ApiT (EpochNo currentEpochNo)) =
+                getFromResponse (#nextEpoch . #epochNumber) r1
+
+        let reqEpochNo = T.pack $ show $ currentEpochNo + 10
```
Anviking: Minor suggestion: 

How about calling this variable `futureEpoch` to better capture what we're doing here?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1299,26 +1299,61 @@ getNetworkInformation (_block0, bp, st) nl = do
...
+    sp = toSlotParameters bp
 
     -- Unsafe constructor for the next epoch. Chances to reach the last epoch
     -- are quite unlikely in this context :)
     unsafeEpochSucc :: HasCallStack => W.EpochNo -> W.EpochNo
     unsafeEpochSucc = fromMaybe bomb . W.epochSucc
       where bomb = error "reached final epoch of the Blockchain!?"
 
+toSlotParameters :: BlockchainParameters -> W.SlotParameters
+toSlotParameters bp = W.SlotParameters
```
Anviking: Already exists as https://github.com/input-output-hk/cardano-wallet/blob/3be4db2f27b20ed2e66dfdbaef2e997801b7bd7c/lib/core/src/Cardano/Wallet/Primitive/Types.hs#L1135-L1141

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1299,26 +1299,61 @@ getNetworkInformation (_block0, bp, st) nl = do
...
-    => (Block, BlockchainParameters, SyncTolerance)
-    -> NetworkLayer IO t Block
+    :: (Block, BlockchainParameters, SyncTolerance)
     -> ApiEpochNumber
     -> Handler ApiNetworkParameters
-getNetworkParameters _ _ _ = throwError err501
+getNetworkParameters (_block0, bp, _st) apiEpochNum = do
+    case apiEpochNum of
+        ApiEpochNumberInvalid txt ->
+            liftHandler $ throwE $ ErrGetNetworkParametersInvalidValue txt
```
Anviking, line 1317: Were there any specific problem that caused you to add `ApiEpochNumberInvalid`? 

I would have thought we could get by without it, and that `ErrGetNetworkParametersInvalidValue` (or at least the HTTP code) could be thrown via the `FromJSON` instance.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1299,26 +1299,61 @@ getNetworkInformation (_block0, bp, st) nl = do
...
+        Quantity . mkPercentage
+    resp :: ApiNetworkParameters
+    resp = ApiNetworkParameters
+        (ApiT $ bp ^. #getGenesisBlockHash)
+        (ApiT $ bp ^. #getGenesisBlockDate)
+        (Quantity slotLength)
+        (Quantity $ W.unEpochLength $ bp ^. #getEpochLength)
+        (bp ^. #getEpochStability)
+        ( handlePercentageConv
+            $ round @Double @Word
```
Anviking: I wonder if it is not a shame to round here.

https://github.com/input-output-hk/cardano-wallet/pull/1294#discussion_r370050003

I think I recall Matthias wondering in hindsight why the sync progress (in percentage) didn't have higher precision. ([here](https://input-output-rnd.slack.com/archives/GJY4XC3DM/p1578671245150200?thread_ts=1578668909.148800&cid=GJY4XC3DM))

Maybe we could think about changing the `Percentage`-type if it makes sense?

lib/core/test/data/Cardano/Wallet/Api/ApiEpochNumber.json
```diff
@@ -1,15 +0,0 @@
...
-    "samples": [
-        "1617893896",
-        "258954262",
-        "latest",
-        "latest",
-        "1310519342",
-        "latest",
-        "42220161",
-        "latest",
-        "1542634324",
```
Anviking, line 12: This golden should be *changed* and not *deleted*?

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: Maybe best treated as separate U/S or issue

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: I have in task specification : 
```
400 # When the epoch parameter is malformed
404 # When the epoch is valid doesn't exist
```
which is reasonable. We want either `latest` or epoch number `0` or `225` So if we put here `-1` or `aaaa` then it is invalid value and I want to emit `err400`.

But we can have also legal value but bigger than current epoch number and we want to differentiate this case from the above. And want to have this : 
```
[pawel@arch ~]$ curl http://localhost:8090/v2/network/parameters/2000000 --http1.1
{"code":"not_such_epoch_no","message":"I couldn't show blockchain parameters for epoch number later than current one. You requested 2000000 epoch. Current one is 1196747. Use smaller epoch than current or 'latest'."}
```
So, what I did I moved responsibility from aeson parser to Server and here I am validating this and properly (and in controlled way) handle it

lib/core/test/data/Cardano/Wallet/Api/ApiEpochNumber.json
paweljakubas: I deleted it because I moved validation downstream, as described in the above comment

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: yea, I was thinking about `Percentage` and its validity. The argument in favor of leaving it as it is is .... that it is like that in specification of the task in jira. I am kinda complying with it.

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: ok, thanks - I will definitely use this from Primitive.Types

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: I don't see how the U/S specifies it to be an integer though.

It's specified to be a `percentage`.

Whose quantity-field is a `number` https://github.com/input-output-hk/cardano-wallet/blob/master/specifications/api/swagger.yaml#L120-L134

`number` means _any number_ according to https://swagger.io/docs/specification/data-models/data-types/#numbers

paweljakubas: ok, I thought it refers to https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Data/Quantity.hs#L136 

I can change it to Double - no problem:-) Thought, that maybe we want to have additional validation (for free) stemming from the use of `mkPercentage`

paweljakubas: On the one hand, this field comes from double field of active slot coefficient, on the other when using Percentage (which is bounded enum of Word from 0 to 100 has its value that it is validated for free)...But if this field has differences important on 0.1 level then I am 100% persuaded we need to use double here.. Just do not know the value window and peculiarities of getActiveSlotCoefficient. I see it is 1% after rounding : 
```
[pawel@arch ~]$ curl http://localhost:8090/v2/network/parameters/latest --http1.1
{"slot_length":{"quantity":2,"unit":"second"},"epoch_stability":{"quantity":5,"unit":"block"},"genesis_block_hash":"f8c0622ea4b768421fea136a6e5a4e3b4c328fc5f16fad75817e40c8a2a56a56","blockchain_start_time":"2019-04-25T14:20:57Z","epoch_length":{"quantity":10,"unit":"slot"},"active_slot_coefficient":{"quantity":1,"unit":"percent"}}
```

paweljakubas:
bors try

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: > ok, I thought it refers to

I would think that the reference in the U/S-swagger schema refers to the existing swagger schema rather than the Haskel type. It could be an overly mechanical reading though. I would understand if it was intended / interpreted as integer.

> I can change it to Double - no problem:-) Thought, that maybe we want to have additional validation (for free) stemming from the use of mkPercentage

Note (as said in DM also): I'm not saying that we should *definitely* do this. My comment was more arguing against the "it's specified as integer in the U/S" argument.

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: > what I did I moved responsibility from aeson parser to Server and here I am validating this and properly (and in controlled way) handle it

Ok thanks, want to have a look and understand this a bit more tomorrow

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: after deep thought - will change to `activeSlotCoefficient :: !(Quantity "percent" Double)`

paweljakubas:
bors try

lib/core/test/data/Cardano/Wallet/Api/ApiEpochNumber.json
Anviking: > I moved validation downstream

As I see it, the JSON does the same validation/parsing, but the actual failure (`throwE`/`fail`) is delayed to later.

The `jsonRoundtripAndGolden` ensure the custom json instances… roundtrips… and doesn't change drastically. 

I do not understand why you removed it 🤔

paweljakubas: I wanted to capture it on Server level, use in `ErrGetNetworkParametersInvalidValue` and emit err400. Otherwise it would return earlier - just when json validation and could not (easily - maybe not true) embrace it and treat equally as `ErrGetNetworkParametersNotSuchEpochNumber`

paweljakubas: So I extended : 
```
data ApiEpochNumber =
    ApiEpochNumberLatest | ApiEpochNumber EpochNo | ApiEpochNumberInvalid Text
```
and funneled text that do not satisfy "latest" or is not convertible to `Word31` to `ApiEpochNumberInvalid`. And if so I can on `Server` level handle errors, control error code, and err msg (this of course can be done also as it was done in the previous PR in instances )

Anviking: Yes, but

1) `ApiEpochNumberInvalid Text` seems strange to me. I guessed the problem it is solving is not new to our wallet, but it looked like a new solution.
2) I don't see how the existence of the roundtrip is related to this. I.e. why is the move of error control an argument for removing the roundtrip? It doesn't hurt to just keep it?

I looked in to this a bit.

Previously you used `fail` in `FromText ApiEpochNumber`. I noticed the endpoint would then `err500 "Something bad happened"` if provided an invalid parameter.

I suspect this is due to `fail` throwing an exception:
```Haskell
(fail "no") :: (Either String Int)
*** Exception: no
```

I noticed that existing endpoints ending with path parameters would not be _not found_ when the path parameter would be incorrect:

e.g. 

```bash
http delete http://localhost:8090/v2/wallets/1efcf74dfc539599f6fa3343859c172a325abbc7/transactions/1423856bc91c49e928f6f30f4e8d665d53eb4ab6028bd0ac971809d514c92db
HTTP/1.1 404 Not Found
Content-Type: application/json;charset=utf-8
Date: Tue, 28 Jan 2020 10:33:12 GMT
Server: Warp/3.2.27
Transfer-Encoding: chunked

{
    "code": "not_found",
    "message": "I couldn't find the requested endpoint. If the endpoint contains path parameters, please ensure they are well-formed, otherwise I won't be able to route them correctly."
}
```

Though doing so here would be against the U/S as you said.

However, I tried replacing your `ApiEpochNumberInvalid` with `Left $ TextDecodingError`, and somehow it actually _does_ return a `err400`:

```bash
http http://localhost:8090/v2/network/parameters/late
HTTP/1.1 400 Bad Request
Content-Type: application/json;charset=utf-8
Date: Tue, 28 Jan 2020 10:42:58 GMT
Server: Warp/3.2.27
Transfer-Encoding: chunked

{
    "code": "bad_request",
    "message": ""
}
```

I am not sure why, or from where the `bad_request` or `message` comes from.

🤷‍♂

Anviking: So seems a bit odd to me, but user-facing, it is a really nice experience the way you've done it.

Still, I'm wondering about this:

> I don't see how the existence of the roundtrip is related to this. I.e. why is the move of error control an argument for removing the roundtrip? It doesn't hurt to just keep it?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1313,12 +1308,39 @@ getNetworkInformation (_block0, bp, st) nl = do
...
+    (W.SlotLength slotLength) = bp ^. #getSlotLength
+    resp :: ApiNetworkParameters
+    resp = ApiNetworkParameters
+        (ApiT $ bp ^. #getGenesisBlockHash)
+        (ApiT $ bp ^. #getGenesisBlockDate)
+        (Quantity slotLength)
+        (Quantity $ W.unEpochLength $ bp ^. #getEpochLength)
+        (bp ^. #getEpochStability)
+        (Quantity $ W.unActiveSlotCoefficient
+            $ bp ^. #getActiveSlotCoefficient )
```
Anviking, line 1339: We need to multiply with 100 here!

On our integration setup, the coefficient is 100%, but I'm getting

```json
http http://localhost:8090/v2/network/parameters/latest
HTTP/1.1 200 OK
Content-Type: application/json;charset=utf-8
Date: Tue, 28 Jan 2020 10:41:18 GMT
Server: Warp/3.2.27
Transfer-Encoding: chunked

{
    "active_slot_coefficient": {
        "quantity": 1,
        "unit": "percent"
    },
    "blockchain_start_time": "2019-04-25T14:20:57Z",
    "epoch_length": {
        "quantity": 10,
        "unit": "slot"
    },
    "epoch_stability": {
        "quantity": 5,
        "unit": "block"
    },
    "genesis_block_hash": "f8c0622ea4b768421fea136a6e5a4e3b4c328fc5f16fad75817e40c8a2a56a56",
    "slot_length": {
        "quantity": 2,
        "unit": "second"
    }
}
```

Anviking: It would be neat if we could add expectations for all the values in the successful integration test

lib/core/test/data/Cardano/Wallet/Api/ApiEpochNumber.json
paweljakubas: no, we can resurrect it if you want. I thought, that when we move validation downstream, away from json decoder, then it is not needed to focus on roundtrips

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: ok, 1 is 100%. you are right

paweljakubas: will add also additional test in integration test suite

lib/core/test/data/Cardano/Wallet/Api/ApiEpochNumber.json
paweljakubas: BUT, as you can see we do not need ToJSON and FromJSON instances of ApiEpochNumber anymore (when we went downstream), so why to add them? Just for testing - I thought that maybe it is better not to bloat code base ;-)

Saying that I am proponent of NOT introducing superfluous things.

Hope, I was able to convince you that roundtrip for this type is not needed in the current constallation

piotr-iohk: I'd vote for keeping the tests for:
 - making sure the system still passes those roundtrips after your changes.
 - as an additional guard for any future changes in the area.

(Unless there are now similar tests elsewhere?)

lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
```diff
@@ -152,3 +157,71 @@ spec = do
...
+                verify r expectations
+
+    it "NETWORK - Cannot query blockchain parameters with epoch later \
+       \than current one" $ \ctx -> do
+        r1 <- request @ApiNetworkInformation ctx
+            Link.getNetworkInfo Default Empty
+        let (ApiT (EpochNo currentEpochNo)) =
+                getFromResponse (#nextEpoch . #epochNumber) r1
+
+        let futureEpochNo = T.pack $ show $ currentEpochNo + 10
```
piotr-iohk, line 198: That is good. I'd like to suggest to use boundary value, i.e. (the next neighboring epoch number)

```
let futureEpochNo = T.pack $ show $ currentEpochNo + 1
```
As it is (the boundary value) usually something that is prone to discover potential issue. 

Also we could try other values:
- max possible epoch number = 2147483648 -> should return `404`
- max possible epoch number + 1 = 2147483649 - > expecting to return `400`

Anviking: I think `currentEpochNo + 1` could cause race conditions and flaky tests.

If the current epoch were to change just after we fetched it, then `futureEpochNo` would no longer be the future epoch number.

Maybe, `+ 2` and an explanatory comment in that case.

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: added in https://github.com/input-output-hk/cardano-wallet/pull/1306/commits/bafd7cce8f4a647c7883b9e3716c275b9f7b6287

lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
piotr-iohk: I see there are boundary values tested in the suite, but perhaps it could be organized for better clarity:

positive ["0", "latest",current, current-1] -> expecting 200 and a response
negative404 [current+1, maxPossibleEpochNumber] - > expecting 404
negative400 ["-1", maxPossibleEpochNumber + 1] -> expecting 400

paweljakubas: it is checked here https://github.com/input-output-hk/cardano-wallet/pull/1306/files#diff-c30d095b1b1142f3da80b28597625b7fR169

paweljakubas: @piotr-iohk feel free to address more things in your review/testing PR for this task:-)

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -477,30 +481,18 @@ instance FromText ApiEpochNumber where
...
 
 instance FromHttpApiData ApiEpochNumber where
     parseUrlPiece = first (T.pack . getTextDecodingError) . fromText
 
-instance ToJSON ApiEpochNumber where
-    toJSON = Aeson.String . toText
-
-instance FromJSON ApiEpochNumber where
-    parseJSON =
-        parseJSON >=> eitherToParser . bimap ShowFmt Prelude.id . fromText
```
paweljakubas, line 502: @Anviking @piotr-iohk  This instances are deleted as validation went downstream. In my humble opinion we do not need to resurrect roundtrips for ApiEpochNumber

paweljakubas: BUT, we still have roundtrip for `jsonRoundtripAndGolden $ Proxy @ApiNetworkParameters` which is very fine

paweljakubas: nevertheless, what Johannes proposed (in private communication) is very reasonable, ie. add textual roudtrip for `ApiEpochNumber` -> https://github.com/input-output-hk/cardano-wallet/pull/1306/commits/8d7e850f3d41a7ac784430fed4b267bc703af31f

lib/core/test/data/Cardano/Wallet/Api/ApiEpochNumber.json
paweljakubas: added textual roundtrip https://github.com/input-output-hk/cardano-wallet/pull/1306/commits/8d7e850f3d41a7ac784430fed4b267bc703af31f

Anviking (approved):

PR #1310, opened 2020-01-28
Correct link label

piotr-iohk:
# Issue Number

# Overview

- [ ] The link label was v2019-01-27 instead of v2020-01-27 :man_facepalming: 

# Comments


Anviking (approved):
Ah!

paweljakubas (approved):
hehe, we are already in roaring 2020s Piotr:-)

PR #1313, opened 2020-01-28
NETWORK_PARAMS - refactor tests and add test for maxBound epoch boundary value

piotr-iohk:
# Issue Number

#1302

# Overview
- 010621576e0cd961313e24a4c669d9dc1631092e
  NETWORK_PARAMS - refactor tests and add test for maxBound epoch bonduary value
  
- 4b6288ded9d1a48e63f848a1463365f521b5baa2
  extract verifyEpochNumOK from tests

# Comments

 - refactored/tagged (`NETWORK_PARAMS`) existing tests to comply with https://docs.google.com/spreadsheets/d/1fadIA_j4nCd3FbylPo5J8K9Fy6tixSC2T9V3xHKuUvU/edit#gid=0
- added test for maxBound epoch value to test boundary value

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
@@ -197,33 +197,29 @@ spec = do
                 Link.getNetworkInfo Default Empty
             let (ApiT (EpochNo currentEpochNo)) =
                     getFromResponse (#nextEpoch . #epochNumber) r1
-
             let futureEpochNo = T.pack $ show $ currentEpochNo + 10
-            let endpoint = ( "GET", "v2/network/parameters/"<>futureEpochNo )
-            r2 <- request @ApiNetworkParameters ctx endpoint Default Empty
-
-            expectResponseCode @IO HTTP.status404 r2
-            expectErrorMessage (errMsg404NoEpochNo (T.unpack futureEpochNo)) r2
+            verifyEpochNumWrong ctx
+                    futureEpochNo
+                    HTTP.status404
+                    (errMsg404NoEpochNo (T.unpack futureEpochNo))
 
         it "Epoch max value" $ \ctx -> do
             let maxEpochValue =
                     T.pack $ show $ fromIntegral @Word31 @Int maxBound
-
-            let endpoint =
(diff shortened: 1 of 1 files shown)


paweljakubas:
Why not to extract this part of code in `where` and reuse : 
```
           let endpoint = ( "GET", "v2/network/parameters/"<>arg )
            r <- request @ApiNetworkParameters ctx endpoint Default Empty
            expectResponseCode @IO HTTP.status200 r

            let networkParams = getFromResponse id r
            networkParams `shouldBe` expectedBlockchainParams
```
?

piotr-iohk:
> Why not to extract this part of code...

Done in 4b6288ded9d1a48e63f848a1463365f521b5baa2

lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
```diff
@@ -162,77 +162,92 @@ spec = do
...
+                Link.getNetworkInfo Default Empty
+            let (ApiT (EpochNo currentEpochNo)) =
+                    getFromResponse (#nextEpoch . #epochNumber) r1
+
+            let futureEpochNo = T.pack $ show $ currentEpochNo + 10
+            let endpoint = ( "GET", "v2/network/parameters/"<>futureEpochNo )
+            r2 <- request @ApiNetworkParameters ctx endpoint Default Empty
+
+            expectResponseCode @IO HTTP.status404 r2
+            expectErrorMessage (errMsg404NoEpochNo (T.unpack futureEpochNo)) r2
```
paweljakubas: this one also can be reused with `Epoch max value` and others below

paweljakubas (approved):
LGTM! maybe it would be worth to introduce also `verifyEpochNumWrong ctx epochNo errStatus errMsg`

lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
piotr-iohk: extracted `verifyEpochNumWrong` in 96bf986e189c3c361937ce5eedc91a01d44daff6.
I think nothing else left/makes really sense to extract :)

Issue #1314, opened 2020-01-29
Add shutdown handler for new launcher


rvl:
# Context

Part of [ADP-92](https://jira.iohk.io/browse/ADP-92) (new launcher).

## User story

As an API client (Javascript)
I want a method to cleanly and reliably stop cardano-wallet (and its node)
so that I can exit my application and release resources allocated to it.

# Decision

Implement the cardano-wallet side of the [launcher shutdown process](https://github.com/input-output-hk/cardano-launcher/blob/master/docs/windows-clean-shutdown.md).

DaedalusIPC can stay for the time being, but will be deleted later.

# Acceptance Criteria

 - The shutdown handler must work on both Windows and POSIX.

---

# Development 

- PR  #1315

# QA 

1. Unit tests in `Cardano.StartupSpec`
2. Integration test in `.../Scenario/CLI/Server.hs`.

piotr-iohk:
hm, one of the integration tests fails for me locally. (btw. thanks for handy `PATATE` handler ;) https://github.com/input-output-hk/cardano-wallet/blob/master/lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/CLI/Server.hs#L72)

```
Failures:

  test/integration/Test/Integration/Jormungandr/Scenario/CLI/Server.hs:162:17: 
  1) CLI Specifications, PATATE, LOGGING - cardano-wallet serve logging [SERIAL], LOGGING - Serve debug logs for one component
       not expected: 0

  To rerun use: --match "/CLI Specifications/PATATE/LOGGING - cardano-wallet serve logging [SERIAL]/LOGGING - Serve debug logs for one component/"

Randomized with seed 831114914

Finished in 3.8773 seconds
13 examples, 1 failure
```

KtorZ:
Also fails for me locally for a reason I can't explain, but does not on CI :thinking:

piotr-iohk:
@rvl, @KtorZ:
 - is this failure on local machine (https://github.com/input-output-hk/cardano-wallet/issues/1314#issuecomment-594061933) is something that should be investigated further?

 - also acceptance criteria for ADP-92 is quite extensive:
```
1. I want to start cardano-wallet (and its node)
2. I want to cardano-wallet to use a free TCP port for its server
3. I want to receive events when the wallet backend is started/stopped
4. I want to receive events when the node is started/stopped
5. I want to receive an event when the wallet API is ready
[...]

PR #1315, opened 2020-01-29
Add shutdown handler for new launcher

rvl:
Relates to #1314.

# Overview

This adds a simpler way of ensuring clean shutdown of the wallet on windows (and linux), which doesn't require DaedalusIPC. The mechanism will be used by cardano-launcher.

- Adds the shutdown handler thread.
- Needed to rearrange startup functions around a little bit.
- Unit tests and an integration test.

# Comments

[Hydra jobset](https://hydra.iohk.io/jobset/Cardano/cardano-wallet-pr-1315)


default.nix
```diff
@@ -39,6 +39,8 @@ let
       haskellBuildUtils = iohkLib.haskellBuildUtils.package;
     };
 
+    inherit (haskellPackages.cardano-wallet-byron.components.exes) cardano-wallet-byron;
```
rvl, line 34: This is used by shell.nix in cardano-launcher.

rvl:
bors try

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/CLI/Server.hs
```diff
@@ -237,6 +244,30 @@ spec = do
...
+                        , "--random-port"
+                        , "--genesis-block-hash", block0H
+                        ]
+                        (pure ())
+                        CreatePipe
+                        stream
+
+                withBackendProcessHandle nullTracer cmd $ \(Just hStdin) _ -> do
+                    threadDelay oneSecond
+                    hClose hStdin
```
paweljakubas, line 264: **[question]**

it all comes down to this when shutting down on cmd, right? No way that we have some kind of shutdown without closing stdin handle?

paweljakubas (approved):
I was looking only at Haskell code only trying to understand logic, mainly in Startup and Launcher modules. Looks fine. Very fond of different cases covered in Cardano.StartupSpec. Looks like you nail it. Of course the ultimate verification is upon operation which to be seen.
btw. I see some complaining of weeder : 
```
warning: unknown setting 'sandbox-fallback'
```

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/CLI/Server.hs
rvl: Yes good point - I added a check to make sure that the process exits with status 0.

rvl:
I fixed weeder and rebased.
It works well on Linux with the cardano-launcher code.
On Windows it deadlocks because it basically does this `race (BS.hGet stdin 1000) (threadDelay 1000000)`, which also deadlocks on Windows. I haven't found a workaround yet.

KtorZ (approved):

Issue #1238, opened 2019-12-30
Revise expectations in integration tests


piotr-iohk:
# Context

In the course of adding integration tests we have added a lot of different and specific assertions which can be replaced with more powerful combinators reducing the code.

# Decision

Replace the myriad of assertions functions with more powerful combinators. For instance, we have `expectFieldEqual`, `expectFieldBetween`, `expectFieldNotEqual` etc ... which can all be expressed via `expectFieldSatisfy` ... Also, all of the `expectEventually` ... can be replaced by a combination of other expectation and the `eventually` combinator we introduced recently.

# Acceptance Criteria

---

# Development 

In #1317:

1. Removed bunch of assertion functions which where replaced by only:
 - expectField
 - expectListField
 - expectCliField
 - expectCliListField

All of which take as an argument a function that is supposed to verify the condition we want to check `(a -> Expectation)`. For instance:

```
expectField #migrationCost (`shouldBe` Quantity 0)
expectField id (`shouldSatisfy` (not. null))
```
`greaterThan` and `(.>=)` helper functions added to use with expectations.

2. Removed `expectEventually` and repaced with the combination of `eventually_` and other expectations

In #1339, revise names of new functions.

# QA 
All tests are passing after the changes, no tests where removed or added.

Anviking:
lgtm! 🎉

PR #1317, opened 2020-01-29
Revise test expectations

piotr-iohk:
# Issue Number

#1238

# Overview

- 0a19dd650666ba3ca53b3e76da7d92a117fa0802
  Remove expectations:
  - expectFieldEqual  
  - expectFieldNotEqual
  - expectListItemFieldEqual
  - expectFieldBetween
  - expectListItemFieldBetween

- 074544e3b09d27bb51dd1917e626146988951e3c
  Add expectations:
  - expectCliFieldSatisfy
  - expectCliListItemFieldSatisfy

  and remove:
  - expectCliFieldBetween
  - expectCliFieldEqual
  - expectCliFieldNotEqual
  - expectCliListItemFieldBetween
  - expectCliListItemFieldEqual

- 7cf0c88575b8367c78ac0f9709dba9625544fb57
  Use eventually_ to remove:
  - expectEventually
  - expectEventually'
  - expectEventuallyL

- 2b2c51dc3a4775d88033c6e97d2c4c73f90bf332
  Rename expect.. methods to make them more concise

# Comments

### DIFF
--- a/lib/core-integration/src/Test/Integration/Framework/DSL.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/DSL.hs
@@ -34,6 +34,8 @@ module Test.Integration.Framework.DSL
     , expectCliField
     , expectCliListField
     , expectWalletUTxO
+    , between
+    , greaterThan
     , verify
     , Headers(..)
     , Payload(..)
@@ -46,6 +48,7 @@ module Test.Integration.Framework.DSL
     -- * Helpers
     , (</>)
     , (!!)
+    , (.>=)
     , emptyRandomWallet
     , emptyIcarusWallet
     , emptyByronWalletWith
@@ -217,7 +220,7 @@ import System.Process
     , withCreateProcess
     )
 import Test.Hspec
-    ( HasCallStack, expectationFailure )
+    ( HasCallStack, expectationFailure, Expectation )
 import Test.Hspec.Expectations.Lifted
     ( shouldBe, shouldContain )
 import Test.Integration.Faucet
@@ -301,22 +304,20 @@ expectResponseCode want (got, a) =
 expectField
     :: (HasCallStack, MonadIO m, MonadFail m, Show a)
     => Lens' s a
-    -> (a -> Bool)
+    -> (a -> Expectation)
     -> (HTTP.Status, Either RequestException s)
     -> m ()
 expectField getter predicate (_, res) = case res of
     Left e  -> wantedSuccessButError e
     Right s ->
         let a = view getter s in
-        if predicate a
-            then return ()
-            else fail $ "predicate failed for: " <> show a
+        liftIO $ predicate a
 
 expectListField
     :: (HasCallStack, MonadIO m, MonadFail m, Show a)
     => Int
     -> Lens' s a
-    -> (a -> Bool)
+    -> (a -> Expectation)
     -> (HTTP.Status, Either RequestException [s])
     -> m ()
 expectListField i getter predicate (c, res) = case res of
@@ -380,7 +381,7 @@ expectCliListField
     :: (HasCallStack, MonadIO m, MonadFail m, Show a)
     => Int
     -> Lens' s a
-    -> (a -> Bool)
+    -> (a -> Expectation)
     -> [s]
     -> m ()
 expectCliListField i getter predicate xs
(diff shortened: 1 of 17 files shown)


piotr-iohk:
bors try

lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -530,62 +372,30 @@ expectValidJSON _ str =
...
-                "expected " <> show a <> " >= " <> show aMin
-            a | a > aMax -> fail $
-                "expected " <> show a <> " <= " <> show aMax
-            _ ->
-                return ()
-
-expectCliListItemFieldBetween
-    :: (HasCallStack, MonadIO m, MonadFail m, Show a, Eq a, Ord a)
+expectCliListItemFieldSatisfy
+    :: (MonadIO m, MonadFail m, Show a)
```
Anviking: Let's keep `HasCallStack` such that the failures point to the source location where `expectCliFieldSatisfy` was called, instead of to this definition!

You kept it in the other ones:
```
 src/Test/Integration/Scenario/API/Transactions.hs:195:5:
  1) API Specifications TRANS_CREATE_01 - Single Output Transaction
       user error (predicate failed for: 24)

  To rerun use: --match "/API Specifications/TRANS_CREATE_01 - Single Output Transaction/"

Randomized with seed 1239180884
```

lib/core-integration/src/Test/Integration/Scenario/API/Transactions.hs
```diff
@@ -207,37 +204,38 @@ spec = do
         verify r
             [ expectSuccess
             , expectResponseCode HTTP.status202
-            , expectFieldBetween
-                    (#amount . #getQuantity) (feeMin + amt, feeMax + amt)
-            , expectFieldEqual (#direction . #getApiT) Outgoing
-            , expectFieldEqual (#status . #getApiT) Pending
+            , expectFieldSatisfy (#amount . #getQuantity) (>= feeMin + amt)
+            , expectFieldSatisfy (#amount . #getQuantity) (<= feeMax + amt)
```
Anviking: The direction here is very nice. But unfortunately the failure-descriptions are now unhelpful:
```
Failures:

  src/Test/Integration/Scenario/API/Transactions.hs:195:5:
  1) API Specifications TRANS_CREATE_01 - Single Output Transaction
       user error (predicate failed for: 24)

  To rerun use: --match "/API Specifications/TRANS_CREATE_01 - Single Output Transaction/"
```

I think we need to capture the same `between` we had before while keeping it as composable as you have now, by making the predicate return failure reasons. (Not just `a -> Bool`)

At the call-site, maybe it could look like this:

```haskell
expectFieldSatisfy 
    (#amount . #getQuantity) 
    (between (feeMin + amt, feeMax + amt))
```
or some variant as you prefer

_actually, the above would read as a sentence if expectFieldSatisfy was named expectField or expect_ 🤔

piotr-iohk: That is a fair point, for the current `expectFieldBetween` the failulre message is:

```
  src/Test/Integration/Scenario/API/Transactions.hs:198:5: 
  1) API Specifications TRANS_CREATE_01x - Single Output Transaction
       user error (expected 24 <= 23)

```

So we are missing the expected condition now.

I think the idea of #1238 was, for now, to replace exisitng expactations with `expect...Satisfy` to make DSL lighter.
On top of those changes we may start introducing improvements. (in subsequent PRs)

Good point also on makeing the methods shorter. I have change that in -> 2b2c51dc3a4775d88033c6e97d2c4c73f90bf332
I also like the idea of `between` and having:

```
expectField #amount (between (feeMin, feeMax))
``` 
Which, indeed, looks neat. 

Again, I would postpone it to the subsequent PRs. Thoughts?

Anviking: > Again, I would postpone it to the subsequent PRs. Thoughts?

My concern would be that failing tests will be more difficult to debug before we fix that.

I also suspect it would be easier to go straight from `expectFieldBetween` to `expectField #foo (between (a, b))`, than converting to two `expectFieldSatisfy` checks (`>=`, and `<=`)

Anviking: I think we would only need:
- `equal` - we can re-use `Test.Hspec.Expectations.shouldBe`
- `notEqual` - we can re-use `Test.Hspec.Expectations.shouldNotBe`
- `between :: (a, a) -> a -> IO ()` the only one we need to implement ourselves

piotr-iohk:
bors try

KtorZ (approved):

Issue #1312, opened 2020-01-28
Better report delegation status in the context of delays


paweljakubas:
# Context

U/S

As a wallet client,
I want to reliably know how my delegation changes,
So that I can better inform users about their current and next delegation settings.

A/C

The wallet engine must only report a wallet as being delegated when it actively IS (i.e. after the 2 epochs grace period upon discovering a delegation certificate).

If multiple certificates are discovered within the same epoch, the wallet must report only about the last one.

The 'WalletAPI' model must be modified to match the following swagger definition:

# Decision

We are going to have the following situations: 
1. not delegating and not joined :
```
"delegation": {
    "active": {
      "status": "not_delegating"
      }
}
``` 
2. not delegating but joined and discovered at epoch=13 (and we are at epoch = 13)
```
"delegation": {
    "active": {
      "status": "not_delegating"
      },
     "next": {
      "status": "delegating",
      "target" : "poolId1",
      "changes_at": {
        "epoch_number": 15,
        "epoch_start_time": "2020-01-22T10:18:54.246Z"
      }}
```
3. not delegating, discovered at epoch = 13, and we are at 14:
```
"delegation": {
    "active": {
      "status": "not_delegating"
      },
     "next": {
      "status": "delegating",
      "target" : "poolId1",
      "changes_at": {
        "epoch_number": 15,
        "epoch_start_time": "2020-01-22T10:18:54.246Z"
      }}
```
4. delegating and at epoch >=15:
```
"delegation": {
    "active": {
      "status": "delegating",
      "target" : "poolId1"
      }
      }}
```
5. delegating and joining at epoch = 20 and discovered at epoch=20
```
"delegation": {
    "active": {
      "status": "delegating",
      "target" : "poolId1"
      },
     "next": {
      "status": "delegating",
      "target" : "poolId2",
      "changes_at": {
        "epoch_number": 22,
        "epoch_start_time": "2020-01-23T10:18:54.246Z"
      }}
```
6. delegating, discovered at epoch = 20 and now epoch =21
```
"delegation": {
    "active": {
      "status": "delegating",
      "target" : "poolId1"
      },
     "next": {
      "status": "delegating",
      "target" : "poolId2",
      "changes_at": {
        "epoch_number": 22,
        "epoch_start_time": "2020-01-23T10:18:54.246Z"
      }}
```
7. delegating to another pool (now epoch>=22):
```
"delegation": {
    "active": {
      "status": "delegating",
[...]

PR #1318, opened 2020-01-29
Better delegation status impl - swagger and types

paweljakubas:
# Issue Number

#1312 

# Overview

- [x] I have updated swagger
- [x] I have added needed types and check them

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -282,7 +282,7 @@ data ApiWallet = ApiWallet
     } deriving (Eq, Generic, Show)
 
 newtype ApiDelegationStatus = ApiDelegationStatus
-    { status :: !(ApiT (WalletDelegation (ApiT PoolId)))
+    { status :: ApiT (WalletDelegation (ApiT PoolId))
     } deriving (Eq, Generic, Show)
 
 data ApiWalletDelegationNext = ApiWalletDelegationNext
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -714,15 +714,20 @@ instance ToJSON ApiWalletDelegation where
 
 instance FromJSON ApiWalletDelegationNext where
     parseJSON = withObject "apiWalletDelegationNext" $ \o -> do
+        st <- o .: "status"
         t  <- o .:? "target"
         chAt <- o .: "changesAt"
-        case t of
-            Nothing ->
-                pure $ ApiWalletDelegationNext
-                (ApiDelegationStatus $ ApiT NotDelegating) chAt
-            Just poolId ->
+        case (t, st) of
+            (Just poolId, Aeson.String "delegating") ->
                 pure $ ApiWalletDelegationNext
                 (ApiDelegationStatus $ ApiT $ Delegating poolId) chAt
+            (Nothing, Aeson.String "not_delegating") ->
+                pure $ ApiWalletDelegationNext
+                (ApiDelegationStatus $ ApiT NotDelegating) chAt
+            _ ->
+                fail "Parse error of ApiWalletDelegationNext. Expecting either \
+                     \delegating status and target or non-delegating status \
+                     \without target"
 instance ToJSON ApiWalletDelegationNext where
     toJSON (ApiWalletDelegationNext (ApiDelegationStatus st) chAt) = do
         let a = fromValue $ toJSON st
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -67,6 +67,9 @@ module Cardano.Wallet.Api.Types
     , Iso8601Time (..)
     , ApiEpochNumber (..)
     , ApiNetworkParameters (..)
+    , ApiWalletDelegation (..)
+    , ApiDelegationStatus (..)
+    , ApiWalletDelegationNext (..)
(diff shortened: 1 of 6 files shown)


lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -278,6 +281,7 @@ spec = do
             jsonRoundtripAndGolden $ Proxy @ApiBlockReference
             jsonRoundtripAndGolden $ Proxy @ApiNetworkInformation
             jsonRoundtripAndGolden $ Proxy @ApiNetworkParameters
+            jsonRoundtripAndGolden $ Proxy @ApiWalletDelegation
```
Anviking, line 284: Why not add the two sub-types here for good measure too?… Or maybe unnecessary 🤷‍♂

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -687,6 +707,35 @@ instance FromJSON (ApiT (WalletDelegation (ApiT PoolId))) where
...
     toJSON = genericToJSON walletDelegationOptions . getApiT
 
+instance FromJSON ApiWalletDelegation where
+    parseJSON = genericParseJSON defaultRecordTypeOptions
+instance ToJSON ApiWalletDelegation where
+    toJSON = genericToJSON defaultRecordTypeOptions
+
+instance FromJSON ApiWalletDelegationNext where
+    parseJSON = withObject "apiWalletDelegationNext" $ \o -> do
+        t  <- o .:? "target"
```
Anviking, line 718: [`.:?`](http://hackage.haskell.org/package/aeson-1.4.6.0/docs/Data-Aeson.html#v:.:-63-) returns `Nothing` if the key is not there or if the value is `null`.

Here you're determining the delegation status based off the existence of the `target` field.

It would seem more logical to me to first parse and case split on the `delegation_status`.

If the status is `delegating`, then the `target` field _must_ exist, right?

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
paweljakubas: yeah, it is worth adding them

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: yeah, we have either : 
```
{ status : delegating, target: poolId, changes_at : date }
```
or
```
{ status : not_delegating, changes_at : date }
```
Cannot be anything else

paweljakubas: but still we can be more rigorous in case and check ALSO if status is delegating when target is present

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/1318/commits/d4aca5887172e02285be2e9582cbae5cecc15887

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/1318/commits/d4aca5887172e02285be2e9582cbae5cecc15887

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -275,6 +281,20 @@ data ApiWallet = ApiWallet
...
+    } deriving (Eq, Generic, Show)
+
+data ApiWalletDelegationNext = ApiWalletDelegationNext
+    { status :: ApiDelegationStatus
+    , changesAt :: ApiEpochInfo
+    } deriving (Eq, Generic, Show)
+
+data ApiWalletDelegation = ApiWalletDelegation
+    { active :: ApiDelegationStatus
+    , next :: !(Maybe ApiWalletDelegationNext)
```
Anviking, line 295: So the `next` field can be missing, I take, from your examples in #1312.

I wonder if it will accept null… 🤔 That would depend on the `genericParseJSON defaultRecordTypeOptions` options I guess.

Anviking: It seems that `defaultRecordTypeOptions` will encode the `Nothing` as a missing field, but also accept `null` when decoding.

lib/core/src/Cardano/Wallet/Api/Types.hs
Anviking: I think your use of `.:?` still would accept `{"status": "not_delegating", "target": null}` as valid.

Is this desirable?

Also, are the `Aeson.String` patterns necessary? Ambiguous string type otherwise?

Anviking (approved):
I think your parsers (and maybe some existing json parsers in the wallet) accept `null` where they — arguably — shouldn't. (OpenAPI 3.0 has `nullable: false` per default)

But I wouldn't necessarily worry about it. The `ToJSON` seems to follows the spec which is the important thing in our case.

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: I treat `Aeson.String "delegating/not_delegating"` as additional check here. If other things will appear we make it `fail`. I will merge it as it is and also think about `null` stuff (probably will add examples to show the case) in next PR concluding this U/S

Issue #1316, opened 2020-01-29
Add CLI command for retrieving root private keys from mnemonics


Anviking:
# Context

[ADP-81](https://jira.iohk.io/secure/RapidBoard.jspa?rapidView=106&projectKey=ADP&view=planning&selectedIssue=ADP-81&issueLimit=100)

It is currently possible to retrieve the reward account private key of a (shelley) mnemonic.

We want to make it possible to expose the root private key (and later address keys) given a mnemonic. And not just for shelley.

# Decision

As a meaningful first step in the story, add command(s) for retrieving the root private key.

# Acceptance Criteria

---

- [x] We _must_ be able to retrieve the root xprvs of Byron-, Icarus-, Trezor- and ledger-wallets.
- [x] Keys _should_ be output as hexadecimal strings
- [x] The output keys _should_  be extended keys
- [x]  The keys _may_  be compatible with jcli

# Development 

- [x] Refactoring existing code to make the different key-types more similar #1343
- [x] Allow internal `XPrv` to be converted _to_ and _from_ a 96-byte long representation suitable for usage in CLI, and that matches jcli   #1369
- [x] Add `key root` command #1321

# QA 
- It is possible to extract root keys using the `icarus`, `trezor`, and `ledger` styles. `byron` support was dropped (https://jira.iohk.io/browse/ADP-81?focusedCommentId=13669&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13669)
- Integration tests checking that `unXPrvStripPub`-output is compatible with `jcli` https://github.com/input-output-hk/cardano-wallet/blob/c392526ffdf063290580bc2d108d5ffdfc29bd74/lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/CLI/Keys.hs#L61-L67
- For testing, heavy emphasis was put on making sure _key-to-hex_ conversion _always_ roundtrips.
    - When encoding a private key to hex, the wallet checks that the output will be able to roundtrip back to the same private key again. If not it errors. This happens in the case of Byron keys and encrypted keys. (more relevant for the `key child` command where Byron-derivation (now dropped) would produce non-roundtripping children for roundtripping parents.)
[...]

PR #1321, opened 2020-01-31
Add CLI command for extracting root xprvs

Anviking:
# Issue Number

#1316 

# Overview

- [x] I have added `cardano-wallet-jormungandr key root --type random <mnemonic words>`
- [x] I added unit tests for help-text and actual usage.
- [x] I added a *pending* test making sure keys are compatible with jcli (which would fail)

# Comments

```bash
$ cardano-wallet-jormungandr key --help
Usage: cardano-wallet-jormungandr key COMMAND
  Derive keys from mnemonics.

Available options:

Available commands:
  root                     Extract root xprv as hex (64 bytes private key + 32
                           bytes chain code)
$ cardano-wallet-jormungandr key root --help
Usage: cardano-wallet-jormungandr key root --type KEYTYPE MNEMONIC_WORDS...
  Extract root xprv as hex (64 bytes private key + 32 bytes chain code)

Available options:
  --type KEYTYPE           Any of the following:
                             random (Daedalus, 12 words)
                             icarus (15 words)
                             trezor (12, 15, 18, 21, or 24 words)
                             ledger (12, 15, 18, 21, or 24 words)
$ cardano-wallet-jormungandr key root --type random flock advance execute country leader exotic mix twenty six margin orient meat
68a0f29e6bd5d8af7ffd00a55006afa8af6fbdbded07984ddf7fb1c31c66f7460685e5d1016553fccc9724f5ee95dd8d66facd2ac1bb2f6fcd7fa5e53c97a57f50c592fcd18b67bf3393a16184d009fb25450b2de8079f870222874e804584a8
```

### DIFF
--- a/lib/cli/cardano-wallet-cli.cabal
+++ b/lib/cli/cardano-wallet-cli.cabal
@@ -44,6 +44,7 @@ library
     , fmt
     , http-client
     , iohk-monitoring
+    , memory
     , servant-client
     , servant-client-core
     , text
@@ -69,12 +70,14 @@ test-suite unit
       -Werror
   build-depends:
       base
+    , bytestring
     , cardano-wallet-cli
     , cardano-wallet-core
     , filepath
     , hspec
     , optparse-applicative
     , QuickCheck
+    , silently
     , temporary
     , text
     , text-class
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -1,18 +1,22 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE ApplicativeDo #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
-{-# LANGUAGE KindSignatures #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE QuasiQuotes #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
 
 -- |
 -- Copyright: © 2018-2020 IOHK
@@ -34,6 +38,7 @@ module Cardano.CLI
     , cmdStakePool
     , cmdNetwork
     , cmdVersion
+    , cmdKey
 
     -- * Option & Argument Parsers
     , optionT
@@ -57,6 +62,15 @@ module Cardano.CLI
     , TxId
     , MnemonicSize (..)
     , Port (..)
+    , CliKeyScheme (..)
+    , CliWalletStyle (..)
+
+    , newCliKeyScheme
+    , hexTextToXPrv
+    , xPrvToHexText
+    , hoistKeyScheme
+    , mapKey
+
 
     -- * Logging
     , withLogging
@@ -102,6 +116,7 @@ import Cardano.Wallet.Api.Server
     ( HostPreference, Listen (..) )
 import Cardano.Wallet.Api.Types
     ( AddressAmount
+    , AllowedMnemonics
     , ApiEpochNumber
     , ApiMnemonicT (..)
     , ApiT (..)
@@ -120,13 +135,22 @@ import Cardano.Wallet.Api.Types
 import Cardano.Wallet.Network
     ( ErrNetworkUnavailable (..) )
 import Cardano.Wallet.Primitive.AddressDerivation
-    ( FromMnemonic (..)
+    ( ErrUnXPrvStripPub (..)
+    , ErrXPrvFromStrippedPubXPrv (..)
+    , FromMnemonic (..)
+    , FromMnemonicError (..)
+    , NatVals (..)
     , Passphrase (..)
     , PassphraseMaxLength
     , PassphraseMinLength
+    , SomeMnemonic (..)
     , WalletKey (..)
+    , XPrv
     , deriveRewardAccount
+    , hex
     , unXPrv
+    , unXPrvStripPub
+    , xPrvFromStrippedPubXPrv
     )
 import Cardano.Wallet.Primitive.AddressDiscovery.Sequential
     ( AddressPoolGap, defaultAddressPoolGap )
@@ -143,27 +167,40 @@ import Control.Arrow
 import Control.Exception
     ( bracket, catch )
 import Control.Monad
-    ( join, unless, void, when )
+    ( join, unless, void, when, (>=>) )
 import Control.Tracer
     ( Tracer, traceWith )
 import Data.Aeson
     ( (.:) )
 import Data.Bifunctor
     ( bimap )
+import Data.ByteArray.Encoding
+    ( Base (Base16), convertFromBase )
 import Data.Char
     ( toLower )
+import Data.List
+    ( intercalate )
 import Data.List.Extra
     ( enumerate )
 import Data.List.NonEmpty
     ( NonEmpty (..) )
 import Data.Maybe
     ( fromMaybe )
+import Data.Proxy
+    ( Proxy (..) )
 import Data.String
     ( IsString )
 import Data.Text
     ( Text )
 import Data.Text.Class
-    ( FromText (..), TextDecodingError (..), ToText (..), showT )
+    ( CaseStyle (..)
+    , FromText (..)
+    , TextDecodingError (..)
+    , ToText (..)
+    , fromTextToBoundedEnum
+    , showT
+    , toTextFromBoundedEnum
+    )
 import Data.Text.Read
     ( decimal )
 import Data.Void
@@ -198,6 +235,7 @@ import Options.Applicative
     , footer
     , header
     , help
+    , helpDoc
     , helper
     , hidden
     , info
@@ -213,6 +251,8 @@ import Options.Applicative
     , subparser
     , value
     )
+import Options.Applicative.Help.Pretty
+    ( string, vsep )
 import Options.Applicative.Types
     ( ReadM (..), readerAsk )
 import Servant.Client
@@ -249,6 +289,7 @@ import System.IO
     , hGetEcho
     , hIsTerminalDevice
     , hPutChar
+    , hPutStrLn
     , hSetBuffering
     , hSetEcho
     , stderr
@@ -258,6 +299,8 @@ import System.IO
 
 import qualified Cardano.BM.Configuration.Model as CM
 import qualified Cardano.BM.Data.BackendKind as CM
+import qualified Cardano.Wallet.Api.Types as API
+import qualified Cardano.Wallet.Primitive.AddressDerivation.Icarus as Icarus
 import qualified Cardano.Wallet.Primitive.AddressDerivation.Shelley as Shelley
 import qualified Codec.Binary.Bech32 as Bech32
 import qualified Codec.Binary.Bech32.TH as Bech32
@@ -266,6 +309,7 @@ import qualified Data.Aeson.Encode.Pretty as Aeson
 import qualified Data.Aeson.Types as Aeson
 import qualified Data.Bifunctor as Bi
 import qualified Data.ByteString as BS
+import qualified Data.ByteString.Char8 as B8
 import qualified Data.ByteString.Lazy as BL
 import qualified Data.ByteString.Lazy.Char8 as BL8
 import qualified Data.List.NonEmpty as NE
@@ -296,6 +340,169 @@ runCli :: ParserInfo (IO ()) -> IO ()
 runCli = join . customExecParser preferences
   where
     preferences = prefs showHelpOnEmpty
+
+{-------------------------------------------------------------------------------
+                            Commands - 'HD Derivation'
+-------------------------------------------------------------------------------}
+
+cmdKey :: Mod CommandFields (IO ())
+cmdKey = command "key" $ info (helper <*> cmds) $ mempty
+    <> progDesc "Derive keys from mnemonics."
+  where
+    cmds = subparser $ mempty
+        <> cmdRootKey
+
+
+-- | Record with mnemonic and key derivation funcionality — /without/ any type
+-- parameters related to scheme.
+--
+-- This means that we can have a value for byron, a value for icarus, both with
+-- the same type @CliKeyScheme@.
+--
+-- @CliKeyScheme@ is on the other hand parameterized over @key@ and @m@.
+-- @hoistKeyScheme@ is provided for mapping over @m@. @mapKey@ is provided for
+-- mapping over @key@.
+--
+-- This way we can test mapping @XPrv@s to @Text@ as pure code in
+-- @Either String@, rather than @IO@.
+data CliKeyScheme key m = CliKeyScheme
+    { allowedWordLengths :: [Int]
+    , mnemonicToRootKey :: [Text] -> m key
+    }
+
+-- | Change the underlying monad of a @CliKeyScheme@.
+hoistKeyScheme
+    :: (forall a. m1 a -> m2 a)
+    -> CliKeyScheme key m1
+    -> CliKeyScheme key m2
+hoistKeyScheme eta s = CliKeyScheme
+    { allowedWordLengths = allowedWordLengths s
+    , mnemonicToRootKey = eta . mnemonicToRootKey s
+    }
+
+xPrvToHexText :: XPrv -> Either String Text
+xPrvToHexText = left showErr . fmap (T.pack . B8.unpack . hex) . unXPrvStripPub
+  where
+    showErr ErrCannotRoundtrip =
+        "That private key looks weird. Is it encrypted? Or an old Byron key?"
+
+hexTextToXPrv :: Text -> Either String XPrv
+hexTextToXPrv txt = do
+    bytes <- fromHex $ B8.pack $ T.unpack txt
+    left showErr $ xPrvFromStrippedPubXPrv bytes
+  where
+    showErr (ErrInputLengthMismatch expected actual) = mconcat
+        [ "Expected extended private key to be "
+        , show expected
+        , " bytes but got "
+        , show actual
+        , " bytes."
+        ]
+    showErr (ErrInternalError msg) = mconcat
+        [ "Unexpected crypto error: "
+        , msg
+        ]
+    fromHex = left (const "Invalid hex.")
+        . convertFromBase Base16
+
+-- | Map over the key type of a CliKeyScheme.
+--
+-- Can be used with e.g @xPrvToHexText@.
+mapKey
+    :: Monad m
+    => (key1 -> m key2)
+    -> CliKeyScheme key1 m
+    -> CliKeyScheme key2 m
+mapKey f s = CliKeyScheme
+    { allowedWordLengths = allowedWordLengths s
+    , mnemonicToRootKey = (mnemonicToRootKey s) >=> f
+    }
+
+eitherToIO :: Either String a -> IO a
+eitherToIO (Right a) = return a
+eitherToIO (Left e) = hPutStrLn stderr e >> exitFailure
+
+data CliWalletStyle = Icarus | Ledger | Trezor
+    deriving (Show, Eq, Generic, Bounded, Enum)
+
+instance FromText CliWalletStyle where
+    fromText = fromTextToBoundedEnum SnakeLowerCase
+
+instance ToText CliWalletStyle where
+    toText = toTextFromBoundedEnum SnakeLowerCase
+
+newCliKeyScheme :: CliWalletStyle -> CliKeyScheme XPrv (Either String)
+newCliKeyScheme = \case
+    Icarus ->
+        let
+            proxy = Proxy @'API.Icarus
+        in
+            CliKeyScheme
+                (apiAllowedLengths proxy)
+                (fmap icarusKeyFromSeed . seedFromMnemonic proxy)
+    Trezor ->
+        let
+            proxy = Proxy @'API.Trezor
+        in
+            CliKeyScheme
+                (apiAllowedLengths proxy)
+                (fmap icarusKeyFromSeed . seedFromMnemonic proxy)
+    Ledger ->
+        let
+            proxy = Proxy @'API.Ledger
+        in
+            CliKeyScheme
+                (apiAllowedLengths proxy)
+                (fmap ledgerKeyFromSeed . seedFromMnemonic proxy)
+
+  where
+    seedFromMnemonic
+        :: forall (s :: API.ByronWalletStyle).
+            (FromMnemonic (AllowedMnemonics s))
+        => Proxy s
+        -> [Text]
+        -> Either String SomeMnemonic
+    seedFromMnemonic _ =
+        left getFromMnemonicError . fromMnemonic @(AllowedMnemonics s)
+
+    apiAllowedLengths
+        :: forall (s :: API.ByronWalletStyle). ( NatVals (AllowedMnemonics s))
+        => Proxy s
+        -> [Int]
+    apiAllowedLengths _ =
+         (map fromIntegral (natVals $ Proxy @(AllowedMnemonics s)))
+
+    icarusKeyFromSeed :: SomeMnemonic -> XPrv
+    icarusKeyFromSeed = Icarus.getKey . flip Icarus.generateKeyFromSeed pass
+
+    ledgerKeyFromSeed :: SomeMnemonic -> XPrv
+    ledgerKeyFromSeed = Icarus.getKey
+        . flip Icarus.generateKeyFromHardwareLedger pass
+
+    -- We don't use passwords to encrypt the keys here.
+    pass = mempty
+
+data KeyRootArgs = KeyRootArgs
+    { _walletStyle :: CliWalletStyle
+    , _mnemonicWords :: [Text]
+    }
+
+cmdRootKey :: Mod CommandFields (IO ())
+cmdRootKey =
+    command "root" $ info (helper <*> cmd) $ mempty
+        <> progDesc "Extract root extended private key from a mnemonic sentence."
+  where
+    cmd = fmap exec $
+        KeyRootArgs <$> walletStyleOption <*> mnemonicWordsArgument
+    exec (KeyRootArgs keyType ws) = do
+        xprv <- mnemonicToRootKey scheme ws
+        TIO.putStrLn xprv
+      where
+        scheme :: CliKeyScheme Text IO
+        scheme =
+            hoistKeyScheme eitherToIO
+            . mapKey xPrvToHexText
+            $ newCliKeyScheme keyType
 
 {-------------------------------------------------------------------------------
                             Commands - 'mnemonic'
@@ -1013,6 +1220,49 @@ loggingSeverityOrOffReader = do
         "off" -> pure Nothing
         _ -> Just <$> loggingSeverityReader
 
+-- | [--wallet-style=WALLET_STYLE]
+--
+-- Note that we in the future might replace the type @CliWalletStyle@ with
+-- another type, to include Shelley keys.
+walletStyleOption :: Parser CliWalletStyle
+walletStyleOption = option (eitherReader fromTextS)
+    ( long "wallet-style"
+    <> metavar "WALLET_STYLE"
+    <> helpDoc (Just (vsep typeOptions))
+    )
+  where
+    typeOptions = string <$>
+        [ "Any of the following:"
+        , "  icarus (" ++ allowedWords Icarus ++ ")"
+        , "  trezor (" ++ allowedWords Trezor ++ ")"
+        , "  ledger (" ++ allowedWords Ledger ++ ")"
+        ]
+
+    allowedWords = (++ " words")
+        . formatEnglishEnumeration
+        . map show
+        . allowedWordLengths
+        . newCliKeyScheme
+       where
+          -- >>> formatEnglishEnumeration ["a", "b", "c"]
+          -- "a, b or c"
+          --
+          -- >>> formatEnglishEnumeration ["a", "b"]
+          -- "a or b"
+          --
+          -- >>> formatEnglishEnumeration ["a"]
+          -- "a"
+         formatEnglishEnumeration = formatEnglishEnumerationRev . reverse
+         formatEnglishEnumerationRev [ult, penult]
+            = penult ++ " or " ++ ult
+         formatEnglishEnumerationRev (ult:penult:revBeginning)
+            = intercalate ", " (reverse revBeginning)
+                ++ ", "
+                ++ penult
+                ++ " or "
+                ++ ult
+         formatEnglishEnumerationRev xs = intercalate ", " (reverse xs)
+
 -- | <wallet-id=WALLET_ID>
 walletIdArgument :: Parser WalletId
 walletIdArgument = argumentT $ mempty
@@ -1039,6 +1289,10 @@ transactionSubmitPayloadArgument :: Parser PostExternalTransactionData
 transactionSubmitPayloadArgument = argumentT $ mempty
     <> metavar "BINARY_BLOB"
     <> help "hex-encoded binary blob of externally-signed transaction."
+
+-- | <mnemonic-words=MNEMONIC_WORD...>
+mnemonicWordsArgument :: Parser [Text]
+mnemonicWordsArgument = some (argument str (metavar "MNEMONIC_WORD..."))
 
 -- | Helper for writing an option 'Parser' using a 'FromText' instance.
 optionT :: FromText a => Mod OptionFields a -> Parser a
--- a/lib/cli/test/unit/Cardano/CLISpec.hs
+++ b/lib/cli/test/unit/Cardano/CLISpec.hs
@@ -2,7 +2,9 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 
 module Cardano.CLISpec
@@ -12,11 +14,14 @@ module Cardano.CLISpec
 import Prelude
 
 import Cardano.CLI
-    ( MnemonicSize (..)
+    ( CliKeyScheme (..)
+    , CliWalletStyle (..)
+    , MnemonicSize (..)
     , Port (..)
     , TxId
     , cli
     , cmdAddress
+    , cmdKey
     , cmdMnemonic
     , cmdNetwork
     , cmdStakePool
@@ -24,13 +29,28 @@ import Cardano.CLI
     , cmdWallet
     , hGetLine
     , hGetSensitiveLine
+    , hexTextToXPrv
+    , mapKey
+    , newCliKeyScheme
+    , xPrvToHexText
     )
 import Cardano.Wallet.Primitive.AddressDerivation
-    ( NetworkDiscriminant (..) )
+    ( NetworkDiscriminant (..), XPrv, unXPrv )
+import Cardano.Wallet.Primitive.Mnemonic
+    ( ConsistentEntropy
+    , EntropySize
+    , Mnemonic
+    , entropyToMnemonic
+    , mnemonicToText
+    )
+import Cardano.Wallet.Unsafe
+    ( unsafeMkEntropy )
 import Control.Concurrent
     ( forkFinally )
 import Control.Concurrent.MVar
     ( newEmptyMVar, putMVar, takeMVar )
+import Control.Exception
+    ( SomeException, try )
 import Control.Monad
     ( mapM_ )
 import Data.Proxy
@@ -39,57 +59,93 @@ import Data.Text
     ( Text )
 import Data.Text.Class
     ( FromText (..), TextDecodingError (..), toText )
+import GHC.TypeLits
+    ( natVal )
 import Options.Applicative
     ( ParserResult (..), columns, execParserPure, prefs, renderFailure )
+import System.Exit
+    ( ExitCode (..) )
 import System.FilePath
     ( (</>) )
 import System.IO
-    ( Handle, IOMode (..), hClose, openFile )
+    ( Handle, IOMode (..), hClose, openFile, stderr )
+import System.IO.Silently
+    ( capture_, hCapture_ )
 import System.IO.Temp
     ( withSystemTempDirectory )
 import Test.Hspec
-    ( Spec, describe, it, shouldBe )
+    ( Spec, describe, expectationFailure, it, shouldBe, shouldStartWith )
 import Test.QuickCheck
     ( Arbitrary (..)
+    , Gen
     , Large (..)
+    , Property
     , arbitraryBoundedEnum
     , checkCoverage
     , counterexample
     , cover
+    , expectFailure
+    , forAll
     , genericShrink
+    , oneof
     , property
+    , vectorOf
+    , (.&&.)
     , (===)
     )
 import Test.Text.Roundtrip
     ( textRoundtrip )
 
+import qualified Data.ByteString as BS
 import qualified Data.Text as T
 import qualified Data.Text.IO as TIO
 
 spec :: Spec
 spec = do
+
+    let defaultPrefs = prefs (mempty <> columns 65)
+
+    let parser = cli $ mempty
+            <> cmdMnemonic
+            <> cmdWallet @'Testnet
+            <> cmdTransaction @'Testnet
+            <> cmdAddress @'Testnet
+            <> cmdStakePool @'Testnet
+            <> cmdNetwork @'Testnet
+            <> cmdKey
+
+
+    let shouldStdOut args expected = it (unwords args) $
+            case execParserPure defaultPrefs parser args of
+                Success x -> capture_ x >>= (`shouldBe` expected)
+                CompletionInvoked _ -> expectationFailure
+                    "expected parser to show usage but it offered completion"
+                Failure failure ->
+                    expectationFailure $ "parser failed with: " ++ show failure
+    let expectStdErr args expectation = it (unwords args) $
+            case execParserPure defaultPrefs parser args of
+                Success x ->
+                    hCapture_ [stderr] (try @SomeException x) >>= (expectation)
+                CompletionInvoked _ -> expectationFailure
+                    "expected parser to show usage but it offered completion"
+                Failure failure -> do
+                    let (str, code) = renderFailure failure ""
+                    code `shouldBe` (ExitFailure 1)
+                    expectation str
     describe "Specification / Usage Overview" $ do
-        let parser = cli $ mempty
-                <> cmdMnemonic
-                <> cmdWallet @'Testnet
-                <> cmdTransaction @'Testnet
-                <> cmdAddress @'Testnet
-                <> cmdStakePool @'Testnet
-                <> cmdNetwork @'Testnet
 
-        let defaultPrefs = prefs (mempty <> columns 65)
-
-        let expectationFailure = flip counterexample False
-
+        let expectationFailure' = flip counterexample False
         let shouldShowUsage args expected = it (unwords args) $
                 case execParserPure defaultPrefs parser args of
-                    Success _ -> expectationFailure
+                    Success _ -> expectationFailure'
                         "expected parser to show usage but it has succeeded"
-                    CompletionInvoked _ -> expectationFailure
+                    CompletionInvoked _ -> expectationFailure'
                         "expected parser to show usage but it offered completion"
                     Failure failure -> property $
                         let (usage, _) = renderFailure failure mempty
-                        in counterexample usage $ expected === lines usage
+                            msg = "*** Expected:\n" ++ (unlines expected)
+                                ++ "*** but actual usage is:\n" ++ usage
+                        in counterexample msg $ expected === lines usage
 
         ["--help"] `shouldShowUsage`
             [ "The CLI is a proxy to the wallet server, which is required for"
@@ -111,6 +167,7 @@ spec = do
             , "  address                  Manage addresses."
             , "  stake-pool               Manage stake pools."
             , "  network                  Manage network."
+            , "  key                      Derive keys from mnemonics."
             ]
 
         ["mnemonic", "--help"] `shouldShowUsage`
@@ -145,6 +202,19 @@ spec = do
             , "  -h,--help                Show this help text"
             , ""
             , "!!! Only for the Incentivized Testnet !!!"
+            ]
+
+
+        ["key", "--help"] `shouldShowUsage`
+            [ "Usage:  key COMMAND"
+            , "  Derive keys from mnemonics."
+            , ""
+            , "Available options:"
+            , "  -h,--help                Show this help text"
+            , ""
+            , "Available commands:"
+            , "  root                     Extract root extended private key from"
+            , "                           a mnemonic sentence."
             ]
 
         ["wallet", "--help"] `shouldShowUsage`
@@ -384,9 +454,23 @@ spec = do
             , "  EPOCH_NUMBER             epoch number parameter or 'latest'"
             ]
 
+        ["key", "root", "--help"] `shouldShowUsage`
+            [ "Usage:  key root --wallet-style WALLET_STYLE MNEMONIC_WORD..."
+            , "  Extract root extended private key from a mnemonic sentence."
+            , ""
+            , "Available options:"
+            , "  -h,--help                Show this help text"
+            , "  --wallet-style WALLET_STYLE"
+            , "                           Any of the following:"
+            , "                             icarus (15 words)"
+            , "                             trezor (12, 15, 18, 21 or 24 words)"
+            , "                             ledger (12, 15, 18, 21 or 24 words)"
+            ]
+
     describe "Can perform roundtrip textual encoding & decoding" $ do
         textRoundtrip $ Proxy @(Port "test")
         textRoundtrip $ Proxy @MnemonicSize
+        textRoundtrip $ Proxy @CliWalletStyle
 
     describe "Transaction ID decoding from text" $ do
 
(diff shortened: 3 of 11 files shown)


erikd:
Please give PRs a proper title instead of just accepting the default. With a proper title I can see whan a PR is about without having to follow the link.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -298,6 +319,144 @@ runCli = join . customExecParser preferences
...
+          -- >>> formatEnglishEnumeration ["a", "b", "c"]
+          -- "a, b or c"
+          --
+          -- >>> formatEnglishEnumeration ["a"]
+          -- "a"
+         formatEnglishEnumeration = formatEnglishEnumerationRev . reverse
+         formatEnglishEnumerationRev (ult:penult:revBeginning)
+            = intercalate ", " (reverse revBeginning)
+                ++ ", "
+                ++ penult
```
Anviking: This probably doesn't work correctly for two elements. We only use and test with 1 and >2 elements.

Anviking: Fixed in 3d54525

lib/cli/src/Cardano/CLI.hs
```diff
@@ -298,6 +319,147 @@ runCli = join . customExecParser preferences
...
+            (Byron.getKey . flip Byron.generateKeyFromSeed pass)
+    Icarus ->
+        mkScheme (Proxy @'Icarus)
+            icarusStyleKeyToSeed
+    Trezor ->
+        mkScheme (Proxy @'Trezor)
+            icarusStyleKeyToSeed
+    Ledger ->
+        mkScheme (Proxy @'Ledger)
+            icarusStyleKeyToSeed
```
KtorZ: Trezor and Ledger have different seed generation styles. See: https://input-output-hk.github.io/cardano-wallet/haddock/edge/cardano-wallet-core-2020.1.27/Cardano-Wallet-Primitive-AddressDerivation-Icarus.html#v:generateKeyFromHardwareLedger

lib/cli/src/Cardano/CLI.hs
```diff
@@ -298,6 +319,147 @@ runCli = join . customExecParser preferences
...
+data CLIKeyScheme = CLIKeyScheme
+    { wordsToSeed :: [Text] -> Either String (Passphrase "seed")
+    , allowedWordLengths :: [Int]
+    , seedToRootKey :: Passphrase "seed" -> XPrv
+    }
+
+byronCliKeyScheme :: ByronWalletStyle -> CLIKeyScheme
+byronCliKeyScheme = \case
+    Random -> mkScheme (Proxy @'Random)
+            (Byron.getKey . flip Byron.generateKeyFromSeed pass)
```
KtorZ: For consistency and readability with below, I'd be better to have that as a named function as well :+1:

lib/cli/src/Cardano/CLI.hs
```diff
@@ -298,6 +319,147 @@ runCli = join . customExecParser preferences
...
+-- We need it in this module to get the result of AllowedMnemonics as a value.
+class ManyNatVal (xs :: [Nat]) where
+  manyNatVal :: proxy xs -> [Integer]
+
+instance ManyNatVal '[] where
+  manyNatVal _ = []
+
+instance (KnownNat a, ManyNatVal as) => ManyNatVal (a ': as) where
+  manyNatVal _ =
+    natVal (Proxy :: Proxy a) : manyNatVal (Proxy :: Proxy as)
```
KtorZ: This belongs to `Cardano.Wallet.Api.Types` next to `AllowedMnemonics`, don't you think?

lib/cli/src/Cardano/CLI.hs
```diff
@@ -298,6 +319,147 @@ runCli = join . customExecParser preferences
...
+  where
+    toIO :: Either TextDecodingError ByronWalletStyle -> IO ByronWalletStyle
+    toIO (Right x) =
+        return x
+    toIO (Left (TextDecodingError e)) =
+        hPutStrLn stderr e >> exitFailure
+
+    typeOptions = string <$>
+        [ "Any of the following:"
+        , "  random (Daedalus, " ++ allowedWords Random ++ ")"
```
KtorZ: We should not mention Daedalus here because this wallet scheme is not specific to Daedalus. Exchanges are using it as well. Stick with `random` :+1:

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -242,6 +247,9 @@ data ByronWalletStyle
     | Icarus
     | Trezor
     | Ledger
+    deriving (Show, Generic, Eq, Bounded, Enum)
+    deriving (ToText, FromText)
+            via (BoundedEnum 'KebabLowerCase ByronWalletStyle)
```
KtorZ: This seems like an odd way of writing: 

```hs
instance FromText ByronWalletStyle where
    fromText = fromTextToBoundedEnum SnakeLowerCase

instance ToText ByronWalletStyle where
    toText = toTextFromBoundedEnum SnakeLowerCase
```

In order to save 2 lines, we have to introduce `DerivingVia` + a whole new type `BoundedEnum`. Please don't.
To be clear, I am not advocating against `DerivingVia` but this is not the right place for introducing this because:

a) We really aren't saving much here... The instances are already straightforward.
b) Rest of the code using the casing functions don't use DerivingVia. So, if something like this is to be introduced, then it needs to be done consistently across all call-site where `SnakeLowerCase` are used. Else, we now have 2 different ways of maintaining parts of code that are doing the exact same thing. 
c) This PR isn't the right place for going through b).

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -707,3 +711,20 @@ hex = convertToBase Base16
 -- | Decode a 'ByteString' from base16
 fromHex :: ByteArray bout => ByteString -> Either String bout
 fromHex = convertFromBase Base16
+
+-- | Convert a key as hex-encoded text
+keyToHexText :: XPrv -> Text
+keyToHexText =
```
KtorZ: This is a very misleading name for a function that actually modifies the internals of the `XPrv`! 
Plus, this function is trying to do 2 things at once:

- Stripping the public part
- Converting to hex

We already have a function that convert ByteString to hexadecimal, make use of it! It's declared 10 lines above!
Now, what about writing a function `stripPubKey :: ByteString -> ByteString` and at call-site, compose them all!

```hs
let bytes = hex . stripPubKey . unXPrv $ xprv 
```

Much clearer and honest about what's being done!

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivationSpec.hs
```diff
@@ -211,6 +215,14 @@ spec = do
         it "XPub IcarusKey"
             (property $ prop_roundtripXPub @IcarusKey)
 
+    xdescribe "keyToHexText (not actually compatible with jcli)" $ do
```
KtorZ: :thinking: ?

KtorZ (changes requested):

lib/core/src/Cardano/Wallet/Api/Types.hs
Anviking: > We really aren't saving much here... The instances are already straightforward.

Personally I find:
- It's easier to see that the `instance` declarations belong to `ByronWalletStyle` as a `deriving` statement directly under the type declaration. Multiple instances following the type declaration make it more difficult to see the structure in the module.
- `DerivingVia` guarantees that `ToText` and `FromText` are consistently derived using the same mechanism. I.e both via `BoundedEnum` and both using `KebabLowerCase`.

> then it needs to be done consistently across all call-site where

I thought some diversity / incremental-ity could be fine, but sure.

>  but this is not the right place for introducing this because:

Fair point.

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivationSpec.hs
Anviking: Stripping the public key in `keyToHexText` was done to match our format with that of jcli. It didn't get us all the way  https://input-output-rnd.slack.com/archives/CCPKJ93Q9/p1580396136029600

You probably think we shouldn't include it as a pending test?

lib/cli/src/Cardano/CLI.hs
Anviking: A fleeting concern I had was that one _might_ think that that `random` meant "choose one of the other three schemes at random" 🙈

Anviking: 10c1378dbc9f858455fde1d5bb52c5e0fa69c70e

lib/cli/src/Cardano/CLI.hs
Anviking: 6c5ec2116d98396351514ca03250a6903ae73c3e

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: I agree with both of your first point and again, I am not against using DerivingVia. But, if we use `A`, we do it consistently. Merging this as it is would introduce `B` and keep `A` in all other places. This makes the code less maintainable. The key for maintainability is consistency. If patterns are not obvious, then we will inevitably deviate from existing patterns and soon, there will be a `C` way of doing it.

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivationSpec.hs
KtorZ: But certainly it does work for keys that are rightfully generated (i.e. with the ed25519 bits tweak)? The Arbitrary instance you use in the test have encryption passphrases, so in most case, you're not trying to import a key, but its encrypted version (which likely is an invalid key). 

This test works just fine if you use empty passphrases / unencrypted keys. 

```hs
keyToHexText (not actually compatible with jcli)
    is compatible with jcli (Shelley)
      +++ OK, passed 100 tests.
```

Anviking: 🤨 Hmm…

Anviking: Fixed.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
Anviking: 👍 Went with `unXPrvStripPub` which you mentioned as an alternative in DM

>  compose them all!

`stripPubKey` is a bit too composable for my liking in this case. `stripPubKey . stripPubKey` makes no sense.

Here I don't see any benefit of composability. In fact, it would also be nice to be keep the jcli-compatibility test as similar to the actual usage in the CLI as possible.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -298,6 +322,133 @@ runCli = join . customExecParser preferences
...
+                ++ penult
+                ++ " or "
+                ++ ult
+         formatEnglishEnumerationRev xs = intercalate ", " (reverse xs)
+
+parseSeedWords :: Parser [Text]
+parseSeedWords = some (argument str (metavar "MNEMONIC_WORDS..."))
+
+-- | Lay thine eyes upon the type parameters and see that there are none.
+data CLIKeyScheme = CLIKeyScheme
```
jonathanknowles: Nit-pick suggestion: use PascalCase here, so:

```suggestion
data CliKeyScheme = CliKeyScheme
```

We then achieve consistency with other functions, such as:
```hs
 byronCliKeyScheme :: ByronWalletStyle -> CliKeyScheme
```

lib/cli/src/Cardano/CLI.hs
```diff
@@ -298,6 +322,133 @@ runCli = join . customExecParser preferences
...
+            )
+        => Proxy s
+        -> (Passphrase "seed" -> XPrv)
+        -> CLIKeyScheme
+
+    -- Construct a KeyScheme using the type families of the promoted type
+    -- argument.
+    mkScheme _ = CLIKeyScheme
+            (left getFromMnemonicError . fromMnemonic @(AllowedMnemonics s) @"seed")
+            (map fromIntegral (manyNatVal $ Proxy @(AllowedMnemonics s)))
```
jonathanknowles: Somehow this indent (and others like it) ended up being 8 spaces instead of 4 spaces.

Perhaps reduce this back to 4 spaces?

lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -473,6 +533,44 @@ spec = do
...
+            \abfa9b0d8764fedecea41018ee6dcf51861bc3923e371dc1c14f79f9e6ea\n"
+
+    describe "key derivation (negative tests)" $ do
+        (["key", "root", "--type", "random"] ++ mw15) `expectStdErr`
+            (`shouldBe` "Invalid number of words: 12 words are expected.\n")
+
+        (["key", "root", "--type", "random"]) `expectStdErr`
+            (`shouldStartWith` "Missing: MNEMONIC_WORDS...")
+
+        let shrug = "¯\\_(ツ)_/¯"
```
jonathanknowles, line 586: :laughing:

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivationSpec.hs
```diff
@@ -279,6 +292,23 @@ prop_passphraseHashMalformed
...
+    monitor (counterexample $ "\nkey bytes = " ++ hexXPrv)
+    (code, stdout, stderr) <- run $ jcliKeyFromHex hexXPrv
+    monitor (counterexample $ "\n" ++ show code)
+    monitor (counterexample $ "Stdout: " ++ show stdout)
+    monitor (counterexample $ "Stderr: " ++ show stderr)
+    assert (stderr == "")
+  where
+    jcliKeyFromHex = readProcessWithExitCode
+            "jcli"
+            ["key", "from-bytes", "--type", "ed25519bip32"]
```
jonathanknowles: Another mysterious 8 character indent! :wink:

lib/cli/src/Cardano/CLI.hs
Anviking: done.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -388,43 +388,83 @@ parseSeedWords = some (argument str (metavar "MNEMONIC_WORDS..."))
...
+         (map fromIntegral (manyNatVal $ Proxy @(AllowedMnemonics s)))
 
-    icarusStyleKeyToSeed =
+    icarusKeyFromSeed =
         (Icarus.getKey . flip Icarus.generateKeyFromSeed pass)
 
+    byronKeyFromSeed = (Byron.getKey . flip Byron.generateKeyFromSeed pass)
+
+    hardwareKeyFromMnemonic :: [Text] -> Either String XPrv
+    hardwareKeyFromMnemonic mw = Icarus.getKey <$> case length mw of
```
Anviking: This is now very ugly 🤔 
- It is duplicated with the API.
- It does **not** respect the `AllowedMnemonics`.

Anviking: We'd need something like
https://github.com/input-output-hk/cardano-wallet/blob/09242e8b7a6d9fd367796d2c82853e49bd37de48/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs#L392-L401

Maybe:
- we can return some `forall mw. (Mnemonic mw)` instead of `Passphrase purpose` (in fromMnemonic)
- we can adjust some existing pieces (like `generateKeyFromHardwareLedger`) to work better together somehow
- we can remove the `mw` type parameter on `Mnemonic mw`. I've grown more and more questioning of it at least.

lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -473,6 +533,44 @@ spec = do
...
+        (["key", "root", "--type", "trezor"] ++ mw15) `shouldStdOut`
+            "003a914372e711b910a75b87e98695929b6960bd5380cfd766b572ea844ea14080\
+            \9eb7ad13f798d06ce550a9f6c48dd2151db4593e67dbd2821d75378c7350f1366b\
+            \85e0be9cdec2213af2084d462cc11e85c215e0f003acbeb996567e371502\n"
+        (["key", "root", "--type", "ledger"] ++ mw15) `shouldStdOut`
+            "003a914372e711b910a75b87e98695929b6960bd5380cfd766b572ea844ea14080\
+            \9eb7ad13f798d06ce550a9f6c48dd2151db4593e67dbd2821d75378c7350f1366b\
+            \85e0be9cdec2213af2084d462cc11e85c215e0f003acbeb996567e371502\n"
+        (["key", "root", "--type", "random"] ++ mw12) `shouldStdOut`
+            "f0ca39a0f8e259704de6fe79b61926411515b97ec99c265a55aefb5c2094e25a80\
```
Anviking: I've never confirmed these goldens with anything. It might be neat to guarantee that deriving address keys manually via the CLI match the unused addresses in the API for new wallets.

Would have to be done when we have `key child --path` though.

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivationSpec.hs
Anviking: Not sure, I might be affected by multiline let-bindings inside do-notation where 4-characters results in parse errors 🤷‍♂

lib/cli/src/Cardano/CLI.hs
Anviking: Done in 09242e8b7a6d9fd367796d2c82853e49bd37de48, although somewhat awkwardly (see https://github.com/input-output-hk/cardano-wallet/pull/1321#discussion_r377580950)

lib/cli/test/unit/Cardano/CLISpec.hs
KtorZ: Where do the values come from then?

Anviking: From the currently implemented CLI itself. 

I.e. they are not guaranteed to be correct, but they won't accidentally change.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -246,6 +260,20 @@ type instance AllowedMnemonics 'Ledger    = '[12,15,18,21,24]
...
+-- Needed to get the result of @AllowedMnemonics@ as a value.
+class ManyNatVal (xs :: [Nat]) where
+  manyNatVal :: proxy xs -> [Integer]
+
+instance ManyNatVal '[] where
+  manyNatVal _ = []
+
+instance (KnownNat a, ManyNatVal as) => ManyNatVal (a ': as) where
+  manyNatVal _ =
+    natVal (Proxy :: Proxy a) : manyNatVal (Proxy :: Proxy as)
```
KtorZ: I just remembered that this already exists as `NatVals` in module Cardano.Wallet.Primitive.AddressDerivation

lib/cli/src/Cardano/CLI.hs
Anviking: Resolved with #1343

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivationSpec.hs
```diff
@@ -214,6 +219,14 @@ spec = do
         it "XPub IcarusKey"
             (property $ prop_roundtripXPub @IcarusKey)
 
+    describe "keyToHexText" $ do
```
Anviking: Function was renamed to `unXPrvStripPub`. Should update.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -296,6 +323,156 @@ runCli = join . customExecParser preferences
...
+                ++ penult
+                ++ " or "
+                ++ ult
+         formatEnglishEnumerationRev xs = intercalate ", " (reverse xs)
+
+parseSeedWords :: Parser [Text]
+parseSeedWords = some (argument str (metavar "MNEMONIC_WORDS..."))
+
+-- | Lay thine eyes upon the type parameters and see that there are none.
+data CliKeyScheme = CliKeyScheme
```
Anviking: It might be a good idea to move most of this, as well as `unXPrvStripPub` to a separate module in `cardano-wallet-cli`.

Anviking: _also, that comment could be a bit more helpful_

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -722,3 +728,28 @@ hex = convertToBase Base16
...
+    stripPub xprv' = prv <> chainCode
+      where
+        (prv, rest) = BS.splitAt 64 xprv'
+        (_pub, chainCode) = BS.splitAt 32 rest
+
+-- Create a @XPrv@ from a prv <> cc bytestring.
+--
+-- TODO: I'm not quite sure how this works, despite rountrip test passing.
+-- Sanity check.
+xPrvFromStrippedPubXPrv :: ByteString -> Either String XPrv
```
Anviking: This confused me:
```haskell
> bytes = "(\134\242|I\141L\EM\NUL\128\173\252q\191\172\167>f \218\222\167.\136\DC4\216\191\253r8cD8&I\STX;\185&\177\172E\241\185\241\157\226\r\163+\EM\GS\232-\188\250[E^N\129J\158\STX\134\188!\241\DLEzt\222\199\247U\143\ETB\128,\226Q\"\230\234\"\191\177\250\230\167\n\214X\244z\\"
> f = unXPrvStripPub . (either error Prelude.id . xPrvFromStrippedPubXPrv)
> f (BS.take 2 bytes) == bytes
True
-- However:
> BS.take 2 bytes
"(\134"
> f "(\134" == bytes
False
```
At first it seemed like I was magically able to re-create the XPrv from 2 bytes. But then it seemed like there is some behind the scenes magic related to `BS.take`.

Anviking:
_will leave the commit history as is for the moment_

lib/cli/src/Cardano/CLI.hs
Anviking: - Leaving in the current module for now.
- Comment for `CliKeyScheme` has been updated.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
Anviking: This is interesting also 😅 
```haskell
>>> unXPrv <$> xPrvFromStrippedPubXPrv (BS.take 0 bytes)
fish: 'stack ghci cardano-wallet-core' terminated by signal SIGSEGV (Address boundary error)
```

Anviking: https://github.com/input-output-hk/cardano-crypto/issues/67

KtorZ: @Anviking lazyness maybe? Nothing tells you that you have to reduce your bytestring to WHNF for comparing them, perhaps the underlying implementation looks at length first, which is known to be different, so the Bytestring has to be different :man_shrugging:

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -722,3 +728,28 @@ hex = convertToBase Base16
...
+xPrvFromStrippedPubXPrv x = xprv
+    $ CC.unEncryptedKey
+    $ CC.encryptedCreateDirectWithTweak x pass
+  where
+    pass :: ByteString
+    pass = ""
```
KtorZ, line 787: It'd be better to have this as a function argument and pass an empty passphrase at the call-site! Avoid hard-coding things in functions.

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivationSpec.hs
```diff
@@ -282,6 +299,29 @@ prop_passphraseHashMalformed
 prop_passphraseHashMalformed pwd = monadicIO $ liftIO $ do
     checkPassphrase pwd (Hash mempty) `shouldBe` Left ErrWrongPassphrase
 
+prop_keyToHexTextJcliCompatible
+    :: WalletKey k
+    => Unencrypted (k 'RootK XPrv)
+    -> Property
+prop_keyToHexTextJcliCompatible (Unencrypted k) = monadicIO $ do
```
KtorZ: This is calling `jcli` directly, so it is so-to-speak an integration test. We expect unit test to not require any 3rd party interaction so that they can be run in isolation if needed. Please, move the test :pray:

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -230,6 +236,13 @@ data ByronWalletStyle
     | Icarus
     | Trezor
     | Ledger
+    deriving (Show, Generic, Eq, Bounded, Enum)
+
+instance FromText ByronWalletStyle where
+    fromText = fromTextToBoundedEnum KebabLowerCase
+
+instance ToText ByronWalletStyle where
+    toText = toTextFromBoundedEnum KebabLowerCase
```
KtorZ: Side-note: we use `SnakeLowerCase` everywhere else. It has no influence here because all words are single words but it'd be better to be consistent with casing on other parts..

lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -473,10 +558,132 @@ spec = do
...
+genAllowedMnemonic s = oneof (map genMnemonicOfSize $ allowedWordLengths s)
+
+genMnemonicOfSize :: Int -> Gen [Text]
+genMnemonicOfSize 12 = mnemonicToText <$> genMnemonic @12
+genMnemonicOfSize 15 = mnemonicToText <$> genMnemonic @15
+genMnemonicOfSize 18 = mnemonicToText <$> genMnemonic @18
+genMnemonicOfSize 21 = mnemonicToText <$> genMnemonic @21
+genMnemonicOfSize 24 = mnemonicToText <$> genMnemonic @24
+genMnemonicOfSize n = error $ "when this test was written, "
+    ++ show n ++ " was not a valid length of a mnemonic"
```
KtorZ: style suggestion, use LambdaCase to avoid repeating the function name 6 times :grimacing:

lib/cli/test/unit/Cardano/CLISpec.hs
KtorZ: It seems very wrong that the `trezor` and `ledger` keys are equal.

lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -147,6 +204,19 @@ spec = do
...
+
+        ["key", "--help"] `shouldShowUsage`
+            [ "Usage:  key COMMAND"
+            , "  Derive keys from mnemonics."
+            , ""
+            , "Available options:"
+            , "  -h,--help                Show this help text"
+            , ""
+            , "Available commands:"
+            , "  root                     Extract root xprv as hex (64 bytes"
```
KtorZ: Extract root extended private key from a mnemonic sentence* 

would be better. 

a) Avoid "cryptic" abbreviation in public-facing interface. `xprv` sounds familiar to us, but users external to the project have no idea what this is about. 
b) Don't mention "hex" here, that's simply the encoding format and is relatively irrelevant (the command-line could even support multiple encoding format). 
c) It does say from where / what the key is extracted.

lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -384,9 +454,24 @@ spec = do
             , "  EPOCH_NUMBER             epoch number parameter or 'latest'"
             ]
 
+        ["key", "root", "--help"] `shouldShowUsage`
+            [ "Usage:  key root --type KEYTYPE MNEMONIC_WORDS..."
```
KtorZ: I am going to be slightly pedantic here but `MNEMONIC_WORDS` should be singular. The meta-type is repeated. Also, why is there an underscore between `MNEMONIC` and `WORDS` but there's none between `KEY` and `TYPE`? The underscore reads better, but it should be used consistently across the interface.

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivationSpec.hs
Anviking: Something like `Test.Integration.Jormungandr.Scenario.CLI.Key`… even though this test is not _completely_ related to the `key` command? 🤔

lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -473,10 +558,132 @@ spec = do
...
+genMnemonic
+    :: forall mw ent csz.
+     ( ConsistentEntropy ent mw csz
+     , EntropySize mw ~ ent
+     )
+    => Gen (Mnemonic mw)
+genMnemonic = do
+        let n = fromIntegral (natVal $ Proxy @(EntropySize mw)) `div` 8
+        bytes <- BS.pack <$> vectorOf n arbitrary
+        let ent = either (error . show) id $ mkEntropy @(EntropySize mw) bytes
```
KtorZ: what about `unsafeMkEntropy` ?

lib/cli/src/Cardano/CLI.hs
```diff
@@ -296,6 +331,202 @@ runCli = join . customExecParser preferences
...
+xPrvToHexText :: XPrv -> Text
+xPrvToHexText = T.pack . B8.unpack . hex . unXPrvStripPub
+
+unsafeHexTextToXPrv :: Text -> XPrv
+unsafeHexTextToXPrv t = either error Prelude.id $
+    xPrvFromStrippedPubXPrv
+    . unsafeFromHex
+    . B8.pack
+    . T.unpack
+    $ t
```
KtorZ: What about having that in `Cardano.Wallet.Unsafe` ?

lib/cli/src/Cardano/CLI.hs
```diff
@@ -296,6 +331,202 @@ runCli = join . customExecParser preferences
...
+            CliKeyScheme
+                (allowedLengths proxy)
+                (fmap icarusKeyFromSeed . seedFromMnemonic proxy)
+    Trezor ->
+        let
+            proxy = Proxy @'Trezor
+        in
+            CliKeyScheme
+                (allowedLengths proxy)
+                (fmap hardwareKeyFromSeed . seedFromMnemonic proxy)
```
KtorZ: Trezor /= Ledger. Trezor uses the same key generation method as Icarus. This is incidentally why the function is called generateKeyFromHardware**Ledger** !

lib/cli/src/Cardano/CLI.hs
```diff
@@ -296,6 +331,202 @@ runCli = join . customExecParser preferences
...
+    -- We don't use passwords to encrypt the keys here.
+    pass = mempty
+
+cmdRootKey :: Mod CommandFields (IO ())
+cmdRootKey =
+    command "root" $ info (helper <*> cmd) $ mempty
+        <> progDesc "Extract root xprv as hex\n\
+                    \ (64 bytes private key + 32 bytes chain code)"
+  where
+    cmd = do
```
KtorZ: Why not follow the same pattern as other command-line handlers? All of them have separated the parsing of the cli argument from the execution of the command. 
This may sounds artificial, but it allows for better readability, better code-reuse (as the parsers can easily be re-used) and easier refactoring options (because the code is already decoupled). 

So: 

a) Declare a new data-type to carry the argument. (e.g. `TransactionListArgs`)
b) Declare your parsers with regards to the existing convention:
    - options are postfixed with "Option" (e.g. portOption, timeRangeOption, poolGapOption, paymentOption, randomPortOption, databaseOption, addressStateOption etc. ..) and declared under the relevant section.
    - arguments are postfixed with "Argument" (e.g. walletIdArgument, epochArgument, transactionIdArgument) 
    - Options are arguments are named after they actual name, camelCased, in the command-line to reduce discrepancy and make mapping between command-line arguments and their relevant parser easier. 

In this case, `parseKeyType` should really be called `typeOption` (although, I suggest to rename this in both the command line and here to perhaps something less ambiguous like `--wallet-style` or `--style` ?)

lib/cli/src/Cardano/CLI.hs
```diff
@@ -296,6 +331,202 @@ runCli = join . customExecParser preferences
...
+
+    typeOptions = string <$>
+        [ "Any of the following:"
+        , "  random (" ++ allowedWords Random ++ ")"
+        , "  icarus (" ++ allowedWords Icarus ++ ")"
+        , "  trezor (" ++ allowedWords Trezor ++ ")"
+        , "  ledger (" ++ allowedWords Ledger ++ ")"
+        ]
+
+    allowedWords = (++ " words")
```
KtorZ: I think this is something we should have near the `AllowedWords` type family declaration.
What about:

```hs
prettyAllowedWords
    :: forall style.
        ( NatVals (AllowedMnemonics style)
        )
    => Proxy style
    -> String
prettyAllowedWords _ =
    show $ natVals $ Proxy @(AllowedMnemonics style) -- TODO: pretty formatting of english enumeration
```

lib/cli/src/Cardano/CLI.hs
```diff
@@ -296,6 +331,202 @@ runCli = join . customExecParser preferences
...
+cmdKey = command "key" $ info (helper <*> cmds) $ mempty
+    <> progDesc "Derive keys from mnemonics."
+  where
+    cmds = subparser $ mempty
+        <> cmdRootKey
+
+-- | Parse a key type
+--
+-- Note that we in the future might replace the type @ByronWalletStyle@ with
+-- another type, to include Shelley keys.
```
KtorZ: Actually yes. I don't understand why all the functions and types are limited to "Byron". Although, "Shelley" key are simply generated following the Icarus' generation style, so they are to-so-speak supported but functions should not imply that this is only a byron thing, because it clearly isn't.

lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -473,10 +558,132 @@ spec = do
...
+  where
+    s :: CliKeyScheme XPrv (Either String)
+    s = byronCliKeyScheme style
+
+    propCanRetrieveRootKey :: [Text] -> Property
+    propCanRetrieveRootKey mw = case mnemonicToRootKey s mw of
+        Right _ -> property True
+        Left e -> counterexample
+            (show (length mw) ++ " words, failed with: " ++ e)
+            (property False)
```
KtorZ, line 637: Good properties to have. But out of curiosity, what problems do the "CliKeyScheme" solve? I fail a bit to see why this type is necessary :thinking:  ?

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
Anviking: In that case we'd have:
```haskell
xPrvFromStrippedPubXPrv :: Passphrase "encryption" -> ByteString -> Either ErrXPrvFromStrippedPubXPrv XPrv
```
taking a password, but
```haskell
unXPrvStripPub :: XPrv -> ByteString
```
not taking a password.

But I guess this is fine as the password would already be "embedded" in the `XPrv`.

That said I'm not quite sure these functions should work regarding passwords. The following test fails:

```haskell
prop_strippedPubRoundtrip
    :: Unencrypted (ByronKey 'RootK XPrv)
    -> Passphrase "encryption"
    -> Property
prop_strippedPubRoundtrip (Unencrypted k) pwd = do
    let f = (xPrvFromStrippedPubXPrv pwd) . unXPrvStripPub . (xPrvChangePass empty pwd) . getRawKey
    let raw = unXPrv . getRawKey
    either (error . show) unXPrv (f k) === raw k
  where
    empty = Passphrase ""
```
because the public keys are different (prv <> pub <> cc) 🤔 
```
"e70d773a098fc7f290b335851a2d7bc5f5a887e97f5b9e7691a414ad04ffbea3d4df6900b462e1e0f9ffeda60d42f25e35555b9059f2a9a4c1c7c019d0efbe677ddaa81e31d0b2c1a939221cf94268e2209d09e191f1fe10ec5a4662cdb1e09caa420eff34b0baf7afccf387517ac43cc2a8059204571bd94808d0e5cb917c14"
/=
"e00d773a098fc7f290b335851a2d7bc5f5a887e97f5b9e7691a414ad04ffbe43d4df6900b462e1e0f9ffeda60d42f25e35555b9059f2a9a4c1c7c019d0efbe6773687a4adefbd180e9aa41e0f4a06638aeee9c6bb76a03374b05f8f574462263aa420eff34b0baf7afccf387517ac43cc2a8059204571bd94808d0e5cb917c14"
```

lib/cli/src/Cardano/CLI.hs
Anviking: I'm inclined to keep it here because:
- I want it next to `xPrvToHexText` as a dual.
- I had thoughts about making `mapKey` take a bidirectional map that can fail* `(key1 -> m key2, key2 -> m key1)`. It would not be `unsafe` in that case. (* more on that later)

lib/core/src/Cardano/Wallet/Api/Types.hs
Anviking: It seemed more appropriate for the CLI… which I'll admit might be iffy when the instance itself lives with the Api-code…

Changing to `SnakeLowerCase` since it doesn't matter.

lib/cli/src/Cardano/CLI.hs
Anviking: Ah! I see.

Renaming `hardwareKeyFromSeed` to `ledgerKeyFromSeed` then. (And using `icarusKeyFromSeed` for Trezor.)

lib/cli/test/unit/Cardano/CLISpec.hs
Anviking: The first goal I had with it was to concentrate the logic related to case switching, and demotion of type-level logic to one place.

I think it captures what we're doing neatly. For instance, it is easy to write the property `prop_allowedWordLengthsAllWork` for **all** different key derivation schemes / types.

When it comes to testing I wanted to be sure that a user performing a sequence of actions via the CLI got the expected result — that nothing ended up wrong when the result of one command was used as input of another. I thought about doing a roundtrip in IO using `caputre_` (didn't go well). Instead I opted to parameterise over `key` and `m`. This allows intricate logic to happen in `mapKey`, which can be tested easily as long we're still in `Either String`.

Depending on https://github.com/input-output-hk/cardano-wallet/issues/1353#issuecomment-587370636 it might have to be altered slightly in the future. (maybe all future functionality will be the same regardless of wallet style, not sure)

Anviking: > This allows intricate logic to happen in mapKey

This would favour later having a `keyArgument :: Parser Text` instead of `keyArgument :: Parser XPrv`, which you might be at odds with.

Instinctively I'd feel being able to use, test and roundtrip a `CliKeyScheme Text (Either String)` makes it worth it.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
KtorZ: unXPrv would give you the **encrypted** key, so you would need to decrypt it first. It's not that the passphrase is embedded, it is that the key is encrypted. 
In principles calling `unXPrvStripPub` on a key that is encrypted should fail. Also, what exactly is the use for "xPrvFromStrippedPubXPrv" ?

Anviking: > Also, what exactly is the use for "xPrvFromStrippedPubXPrv" ?

It will be needed to parse 96-byte key as a CLI argument. I.e `key child` in the future.

For now, it is only used for roundtrips.

We could move it to `Cardano.CLI`. I contemplated doing that.

lib/cli/src/Cardano/CLI.hs
Anviking: >  I don't understand why all the functions and types are limited to "Byron". 

I believe you specifically said not to support `Shelley` keys here.

With that in mind, I didn't want to unnecessarily create a new type, since `ByronWalletStyle` already existed.

The parser is now called `walletStyleOption` (not mentioning byron.) Maybe it is only a matter of renaming `byronCliKeyScheme` to say `newCliKeyScheme`?

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
KtorZ: > It will be needed to parse 96-byte key as a CLI argument. I.e key child in the future.

:+1: Ok, then a passphrase makes little sense here, but using `XPrv` feels also misleading because we do actually treat them and manipulate them as unencrypted key. A newtype could capture that nicely... or at the very least a comment / note explaining why the passphrase is forced to be empty. (Still, I'd add an invariant or runtime check that, when trying to xPrvStripXPub, the given key is not 
encrypted and fail loudly if it is).

lib/cli/test/unit/Cardano/CLISpec.hs
Anviking: Indeed 👍 

That said, I'm starting to question how valuable the `unsafe...` family of functions really are. `mkEntropy`,  `EntropySize`, etc.  come from `Cardano.Wallet.Primitive.Mnemonic`, whereas `unsafeMkEntropy` now is the odd one coming from `Cardano.Wallet.Unsafe`.

I guess they make more sense when they're able to `f . g . either (error . show) id .  h`, rather than just `either (error . show) id . f` like in this specific example.

… And I think there's _something_ attractive about having a multitude of various `unsafe...` functions all in the _same_ module. Maybe it's easier to import `unsafeFoo` from `Cardano.Wallet.Unsafe` than remember where the original `foo`-function is defined  when you're writing a test.

lib/cli/test/unit/Cardano/CLISpec.hs
Anviking: Done. `MNEMONIC_WORD...` and `KEY_TYPE`.

(Underscores are already used elsewhere in the CLI)

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivationSpec.hs
Anviking: We need to duplicate the `Arbitrary (Unencrypted key)`  instances to `cardano-wallet-jormungandr:integration`. But we still need to keep them here (`cardano-wallet-core:unit`) for the `unXPrvStripPub` roundtrips.

And we need to duplicate `getMnemonic` yet again because that weird type mismatch stops us from moving it to `test-utils`.

😶 🔫

lib/cli/test/unit/Cardano/CLISpec.hs
KtorZ: Note that the `unsafe` function are only convenience. We (ought to) only use them in test code for either literal values or values that are hand-crafted and known to be correct. The real question is "why is this module part of the library and not the test-suite only". And the main reason for this is just that it's shared across multiple test modules (for what the argument is worth).

lib/cli/src/Cardano/CLI.hs
KtorZ: Meh. Yes but Shelley /= Jormungandr, although Jörmungandr uses Icarus' key style so it's already covered...  to some extend, if one knows that one can use `--icarus` to generate "Jörmungandr keys". 

>  it is only a matter of renaming byronCliKeyScheme to say newCliKeyScheme?

Yes.

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivationSpec.hs
KtorZ: "that weird type mismatch" ?

Anviking: https://input-output-rnd.slack.com/archives/GBT05825V/p1581523877011800

```
/Users/anviking/IOHK/cardano-wallet/lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs:477:47: error:
    • Couldn't match type ‘cardano-wallet-core-2020.1.27:Cardano.Wallet.Primitive.Mnemonic.Mnemonic
                             12’
                     with ‘Cardano.Wallet.Primitive.Mnemonic.Mnemonic mw0’
      NB: ‘Cardano.Wallet.Primitive.Mnemonic.Mnemonic’
            is defined in ‘Cardano.Wallet.Primitive.Mnemonic’
                in package ‘cardano-wallet-core-2020.1.27’
          ‘cardano-wallet-core-2020.1.27:Cardano.Wallet.Primitive.Mnemonic.Mnemonic’
            is defined in ‘Cardano.Wallet.Primitive.Mnemonic’
                in package ‘cardano-wallet-core-2020.1.27’
      Expected type: Gen (Cardano.Wallet.Primitive.Mnemonic.Mnemonic mw0)
        Actual type: Gen
                       (cardano-wallet-core-2020.1.27:Cardano.Wallet.Primitive.Mnemonic.Mnemonic
                          12)
    • In the second argument of ‘(<$>)’, namely ‘genMnemonic @12’
      In the expression: Just . SomeMnemonic <$> genMnemonic @12
      In the expression: (3, Just . SomeMnemonic <$> genMnemonic @12)
    |
477 |                 , (3, Just . SomeMnemonic <$> genMnemonic @12)
    |                                               ^^^^^^^^^^^^^^^

```
when defining `genMnemonic` in the `test-utils` package.

lib/cli/src/Cardano/CLI.hs
Anviking: Done. Now `newCliKeyScheme`.

lib/cli/src/Cardano/CLI.hs
Anviking: Eh, maybe in a separate pr please. We should maybe have a notion of `all` wallet styles also. And might be best to have at the term-level.

lib/cli/src/Cardano/CLI.hs
Anviking: Thanks for the explanations. I was too unfamiliar with everything here and didn't see the structure.

> although, I suggest to rename this in both the command line and here to perhaps something less ambiguous

Yeah,  `key root --wallet-style icarus` is probably clearer than `key root --type icarus` 🤷‍♂

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivationSpec.hs
Anviking: It's moved at least. Duplication of generators is a separate issue we can live with for the moment.

lib/cli/cardano-wallet-cli.cabal
```diff
@@ -69,12 +69,14 @@ test-suite unit
...
   build-depends:
       base
+    , bytestring
     , cardano-wallet-cli
     , cardano-wallet-core
     , filepath
     , hspec
     , optparse-applicative
     , QuickCheck
+    , silently
```
KtorZ, line 80: A left-over unused dependency?

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -722,3 +729,41 @@ hex = convertToBase Base16
...
+unXPrvStripPub :: XPrv -> ByteString
+unXPrvStripPub = stripPub . unXPrv
+  where
+    -- Converts  prv <> pub <> cc
+    -- To        prv <>        cc
+    stripPub :: ByteString -> ByteString
+    stripPub xprv' = prv <> chainCode
+      where
+        (prv, rest) = BS.splitAt 64 xprv'
+        (_pub, chainCode) = BS.splitAt 32 rest
```
KtorZ, line 768: :+1:

lib/cli/test/unit/Cardano/CLISpec.hs
Anviking: > Note that the unsafe function are only convenience.

Well, I didn't (clearly) see the convenience in writing `unsafeFoo` rather than `either (error . show) id foo`, or (the hypothetical) `unsafe foo`.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -722,3 +729,41 @@ hex = convertToBase Base16
...
+data ErrXPrvFromStrippedPubXPrv
+    = ErrInputLengthMismatch Int Int -- ^ Expected, Actual
+    | ErrInternalError String
+    deriving (Eq, Show)
+
+-- Create a @XPrv@ from a prv <> cc bytestring.
+xPrvFromStrippedPubXPrv :: ByteString -> Either ErrXPrvFromStrippedPubXPrv XPrv
+xPrvFromStrippedPubXPrv x =
+    let
+        res = toXPrv $ CC.encryptedCreateDirectWithTweak x pass
```
KtorZ: This won't work for Byron keys though. The "tweak" mention here does actually change a few bits that could have a disastrous effects on legacy keys:

```
	secret_key[0] &= 248;   /* clears the bottom 3 bits */
	secret_key[31] &= 0x1F; /* clears the 3 highest bits */
	secret_key[31] |= 64;   /* set the 2nd highest bit */
```

We should instead go for `wallet_encrypted_initialize` which simply computes the public key from an already tweaked private key. Yet, this is not exported via the FFI on Haskell. So, another option would be to make this function depends on the target version (V1 vs V2) and use the appropriate creation function.

Anviking: > This won't work for Byron keys though. 

What won't work? For _unencrypted_ keys, all the roundtrips are passing. The CLISpec roundtrips include Byron keys.

I think I see how we could throw in `unXPrvStripPub` if the keys are _encrypted_, like you suggested elsewhere.

lib/cli/cardano-wallet-cli.cabal
Anviking: Actually not 😆

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
KtorZ: The issue has nothing to do with encryption here. Keys are a random sequence of bytes. However, some **bits** in the first and last **byte** are tweaked in Icarus and Byron keys for some security reasons, but in a slightly different way.

Looking back into the generation methods (https://gist.github.com/KtorZ/b086754a15718affcc1b5c03c12c078a#byron), the Byron approach isn't exactly the same but should work all-the-same **for a key that has already been generated**. Soooooo. Okay. Nevermind.

Anviking:
_could split off bfe2568 to a separate pr_ 🤔

I made `unXPrvStripPub` fail if it cannot be roundtripped. And started adding properties. Not happy with them yet though.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
Anviking: > Byron approach isn't exactly the same but should work all-the-same for a key that has already been generated.

Are you saying that the difference between bit-tweaking of Byron and Icarus keys shouldn't matter in this case?

If this tweak "clears the 3 highest bits" but Byron keys only have the highest bit cleared… shouldn't that modify the key? It looks like the 2nd highest bit is set, but the 3rd highest bit should differ?

Although roundtrips for non-encrypted Byron-keys haven't failed so far. I would have expected them to fail in that case.

> for a key that has already been generated

Not sure I get this.

Anviking: I tested with some old mnemonics and they seem to work (have roundtripping root keys) though.

_actually, when doing Byron child derivation I'm seeing problems with roundtrips. Haven't looked close yet._

KtorZ (approved):
:+1: (please wait after #1358 before merging, I cherry-picked one of your commit so the two branches will conflict).

PR #1322, opened 2020-01-31
Better delegation reporiting - impl

paweljakubas:
…ns discovered

# Issue Number

#1312 

# Overview

- [x] I have extended DB layer to obtain the latest two delegation discovered (slot id and potential pool id)
- [x] I have added wallet function that retrieves last 2 delegations discovered
- [x] I have added Api.Server logic function that basing on current epoch and last two delegation certificates determine what to report in delegation value of ApiWallet 
- [x] I have adjusted all integration tests affected
- [x] I have added integration tests showing the cases to be covered in task description
- [x] I have updated all things that needs to be fixed for this change : use of new ApiWalletDelegation as delegation type 

# Comments

### DIFF
--- a/lib/core-integration/src/Test/Integration/Framework/DSL.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/DSL.hs
@@ -7,7 +7,6 @@
 {-# LANGUAGE QuasiQuotes #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 
@@ -92,6 +91,7 @@ module Test.Integration.Framework.DSL
     , toQueryString
     , withMethod
     , withPathParam
+    , notDelegatingNotJoining
 
     -- * CLI
     , command
@@ -120,12 +120,14 @@ import Cardano.Wallet.Api.Types
     , ApiAddress
     , ApiByronWallet
     , ApiCoinSelection
+    , ApiDelegationStatus (..)
     , ApiFee
     , ApiNetworkInformation
     , ApiT (..)
     , ApiTransaction
     , ApiUtxoStatistics (..)
     , ApiWallet
+    , ApiWalletDelegation (..)
     , ByronWalletStyle (..)
     , Iso8601Time (..)
     , WalletStyle (..)
@@ -145,6 +147,7 @@ import Cardano.Wallet.Primitive.Types
     , TxOut (..)
     , UTxO (..)
     , UTxOStatistics (..)
+    , WalletDelegation (..)
     , WalletId (..)
     , computeUtxoStatistics
     , log10
@@ -1097,7 +1100,7 @@ toQueryString kvs = if T.null suffix then mempty else "?" <> suffix
 
 infixr 5 </>
 (</>) :: ToHttpApiData a => Text -> a -> Text
-base </> next = mconcat [base, "/", toQueryParam next]
+base </> thenext = mconcat [base, "/", toQueryParam thenext]
 
 ---
 --- CLI
@@ -1305,6 +1308,12 @@ deleteTransactionViaCLI ctx wid tid = cardanoWalletCLI @t $ join
     [ ["transaction", "forget"]
     , ["--port", show (ctx ^. typed @(Port "wallet")), wid, tid]
     ]
+
+notDelegatingNotJoining :: ApiWalletDelegation
+notDelegatingNotJoining =
+    ApiWalletDelegation
+    (ApiDelegationStatus $ ApiT NotDelegating)
+    Nothing
 
 proc' :: FilePath -> [String] -> CreateProcess
 proc' cmd args = (proc cmd args)
--- a/lib/core-integration/src/Test/Integration/Framework/DSL.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/DSL.hs
@@ -92,6 +92,10 @@ module Test.Integration.Framework.DSL
     , withMethod
     , withPathParam
     , notDelegatingNotJoining
+    , notDelegatingJoining
+    , delegating
+    , delegatingQuiting
+    , delegatingJoining
 
     -- * CLI
     , command
@@ -121,6 +125,7 @@ import Cardano.Wallet.Api.Types
     , ApiByronWallet
     , ApiCoinSelection
     , ApiDelegationStatus (..)
+    , ApiEpochInfo (..)
     , ApiFee
     , ApiNetworkInformation
     , ApiT (..)
@@ -128,6 +133,7 @@ import Cardano.Wallet.Api.Types
     , ApiUtxoStatistics (..)
     , ApiWallet
     , ApiWalletDelegation (..)
+    , ApiWalletDelegationNext (..)
     , ByronWalletStyle (..)
     , Iso8601Time (..)
     , WalletStyle (..)
@@ -139,9 +145,11 @@ import Cardano.Wallet.Primitive.Mnemonic
 import Cardano.Wallet.Primitive.Types
     ( Address (..)
     , Coin (..)
+    , EpochNo
     , Hash (..)
     , HistogramBar (..)
     , PoolId (..)
+    , SlotParameters
     , SortOrder (..)
     , TxIn (..)
     , TxOut (..)
@@ -248,6 +256,7 @@ import Web.HttpApiData
     ( ToHttpApiData (..) )
 
 import qualified Cardano.Wallet.Api.Link as Link
+import qualified Cardano.Wallet.Primitive.Types as Types
 import qualified Data.Aeson as Aeson
 import qualified Data.ByteString.Char8 as B8
 import qualified Data.ByteString.Lazy as BL
@@ -1311,9 +1320,64 @@ deleteTransactionViaCLI ctx wid tid = cardanoWalletCLI @t $ join
 
 notDelegatingNotJoining :: ApiWalletDelegation
 notDelegatingNotJoining =
+    ApiWalletDelegation (ApiDelegationStatus $ ApiT NotDelegating) Nothing
+
+notDelegatingJoining
+    :: ApiT PoolId
+    -> EpochNo
+    -> SlotParameters
+    -> ApiWalletDelegation
+notDelegatingJoining poolId epochNo sp =
     ApiWalletDelegation
     (ApiDelegationStatus $ ApiT NotDelegating)
+    (Just $
+        ApiWalletDelegationNext
+        ( ApiDelegationStatus $ ApiT $ Delegating poolId)
+        ( ApiEpochInfo
+            (ApiT $ epochNo + 2) (Types.epochStartTime sp (epochNo + 2))
+        )
+    )
+
+delegating
+    :: ApiT PoolId
+    -> ApiWalletDelegation
+delegating poolId =
+    ApiWalletDelegation
+    (ApiDelegationStatus $ ApiT $ Delegating poolId)
     Nothing
+
+delegatingQuiting
+    :: ApiT PoolId
+    -> EpochNo
+    -> SlotParameters
+    -> ApiWalletDelegation
+delegatingQuiting poolId epochNo sp=
+    ApiWalletDelegation
+    (ApiDelegationStatus $ ApiT $ Delegating poolId)
+    (Just $
+        ApiWalletDelegationNext
+        ( ApiDelegationStatus $ ApiT NotDelegating)
+        ( ApiEpochInfo
+            (ApiT $ epochNo + 2) (Types.epochStartTime sp (epochNo + 2))
+        )
+    )
+
+delegatingJoining
+    :: ApiT PoolId
+    -> ApiT PoolId
+    -> EpochNo
+    -> SlotParameters
+    -> ApiWalletDelegation
+delegatingJoining poolId1 poolId2 epochNo sp =
+    ApiWalletDelegation
+    (ApiDelegationStatus $ ApiT $ Delegating poolId1)
+    (Just $
+        ApiWalletDelegationNext
+        ( ApiDelegationStatus $ ApiT $ Delegating poolId2)
+        ( ApiEpochInfo
+            (ApiT $ epochNo + 2) (Types.epochStartTime sp (epochNo + 2))
+        )
+    )
 
 proc' :: FilePath -> [String] -> CreateProcess
 proc' cmd args = (proc cmd args)
--- a/lib/core-integration/src/Test/Integration/Framework/DSL.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/DSL.hs
@@ -91,11 +91,6 @@ module Test.Integration.Framework.DSL
     , toQueryString
     , withMethod
     , withPathParam
-    , notDelegatingNotJoining
-    , notDelegatingJoining
-    , delegating
-    , delegatingQuiting
-    , delegatingJoining
 
     -- * CLI
     , command
@@ -124,16 +119,12 @@ import Cardano.Wallet.Api.Types
     , ApiAddress
     , ApiByronWallet
     , ApiCoinSelection
-    , ApiDelegationStatus (..)
-    , ApiEpochInfo (..)
     , ApiFee
     , ApiNetworkInformation
     , ApiT (..)
     , ApiTransaction
     , ApiUtxoStatistics (..)
     , ApiWallet
-    , ApiWalletDelegation (..)
-    , ApiWalletDelegationNext (..)
     , ByronWalletStyle (..)
     , Iso8601Time (..)
     , WalletStyle (..)
@@ -145,17 +136,14 @@ import Cardano.Wallet.Primitive.Mnemonic
 import Cardano.Wallet.Primitive.Types
     ( Address (..)
     , Coin (..)
-    , EpochNo
     , Hash (..)
     , HistogramBar (..)
     , PoolId (..)
-    , SlotParameters
     , SortOrder (..)
     , TxIn (..)
     , TxOut (..)
     , UTxO (..)
     , UTxOStatistics (..)
-    , WalletDelegation (..)
     , WalletId (..)
     , computeUtxoStatistics
     , log10
@@ -256,7 +244,6 @@ import Web.HttpApiData
     ( ToHttpApiData (..) )
 
 import qualified Cardano.Wallet.Api.Link as Link
-import qualified Cardano.Wallet.Primitive.Types as Types
 import qualified Data.Aeson as Aeson
 import qualified Data.ByteString.Char8 as B8
 import qualified Data.ByteString.Lazy as BL
@@ -1318,66 +1305,6 @@ deleteTransactionViaCLI ctx wid tid = cardanoWalletCLI @t $ join
     , ["--port", show (ctx ^. typed @(Port "wallet")), wid, tid]
     ]
 
-notDelegatingNotJoining :: ApiWalletDelegation
-notDelegatingNotJoining =
-    ApiWalletDelegation (ApiDelegationStatus $ ApiT NotDelegating) Nothing
-
-notDelegatingJoining
-    :: ApiT PoolId
-    -> EpochNo
-    -> SlotParameters
-    -> ApiWalletDelegation
-notDelegatingJoining poolId epochNo sp =
-    ApiWalletDelegation
-    (ApiDelegationStatus $ ApiT NotDelegating)
-    (Just $
-        ApiWalletDelegationNext
-        ( ApiDelegationStatus $ ApiT $ Delegating poolId)
-        ( ApiEpochInfo
-            (ApiT $ epochNo + 2) (Types.epochStartTime sp (epochNo + 2))
-        )
-    )
-
-delegating
-    :: ApiT PoolId
-    -> ApiWalletDelegation
-delegating poolId =
-    ApiWalletDelegation
-    (ApiDelegationStatus $ ApiT $ Delegating poolId)
-    Nothing
-
-delegatingQuiting
-    :: ApiT PoolId
-    -> EpochNo
-    -> SlotParameters
-    -> ApiWalletDelegation
-delegatingQuiting poolId epochNo sp=
-    ApiWalletDelegation
-    (ApiDelegationStatus $ ApiT $ Delegating poolId)
-    (Just $
-        ApiWalletDelegationNext
-        ( ApiDelegationStatus $ ApiT NotDelegating)
-        ( ApiEpochInfo
-            (ApiT $ epochNo + 2) (Types.epochStartTime sp (epochNo + 2))
-        )
-    )
-
-delegatingJoining
-    :: ApiT PoolId
-    -> ApiT PoolId
-    -> EpochNo
-    -> SlotParameters
-    -> ApiWalletDelegation
-delegatingJoining poolId1 poolId2 epochNo sp =
-    ApiWalletDelegation
-    (ApiDelegationStatus $ ApiT $ Delegating poolId1)
-    (Just $
-        ApiWalletDelegationNext
-        ( ApiDelegationStatus $ ApiT $ Delegating poolId2)
-        ( ApiEpochInfo
-            (ApiT $ epochNo + 2) (Types.epochStartTime sp (epochNo + 2))
-        )
-    )
 
 proc' :: FilePath -> [String] -> CreateProcess
 proc' cmd args = (proc cmd args)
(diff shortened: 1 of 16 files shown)


paweljakubas:
bors try

specifications/api/swagger.yaml
```diff
@@ -723,6 +705,35 @@ definitions:
...
+        type: object
+        <<: *ApiDelegationStatus
+        description: Currently active delegation status.
+        example:
+          status: delegating
+          target: 1423856bc91c49e928f6f30f4e8d665d53eb4ab6028bd0ac971809d514c92db1
+      next:
+        type: object
+        <<: *ApiDelegationStatus
+        description: Next delegation settings about to change. Only present when a status change has to reported.
```
Anviking: > when a status change has to reported

This sentence sounds a bit weird.

to -> been?

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
```diff
@@ -799,6 +823,79 @@ spec = do
...
+            [ expectFieldEqual #delegation notDelegatingNotJoining
+            ]
+
+        -- joining first pool
+        r1 <- joinStakePool ctx (p1 ^. #id) (w, fixturePassphrase)
+        expectResponseCode HTTP.status202 r1
+        eventually $ do
+            let ep = Link.listTransactions @'Shelley w
+            request @[ApiTransaction n] ctx ep Default Empty >>= flip verify
+                [ expectListItemFieldEqual 0 (#direction . #getApiT) Outgoing
```
Anviking, line 815: So we _need_ to check that the tx is `InLedger`, but maybe it is superfluous to check that the tx is outgoing in this test?

lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -1306,6 +1318,67 @@ deleteTransactionViaCLI ctx wid tid = cardanoWalletCLI @t $ join
     , ["--port", show (ctx ^. typed @(Port "wallet")), wid, tid]
     ]
 
+notDelegatingNotJoining :: ApiWalletDelegation
+notDelegatingNotJoining =
+    ApiWalletDelegation (ApiDelegationStatus $ ApiT NotDelegating) Nothing
+
+notDelegatingJoining
```
Anviking: Alternatives: notDelegating**AboutTo**Join, or notDelegating**And**Joining , but feel free to disregard

lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -245,6 +256,7 @@ import Web.HttpApiData
     ( ToHttpApiData (..) )
 
 import qualified Cardano.Wallet.Api.Link as Link
+import qualified Cardano.Wallet.Primitive.Types as Types
```
Anviking: We often tend to import this as `W` I think?

lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -1306,6 +1318,67 @@ deleteTransactionViaCLI ctx wid tid = cardanoWalletCLI @t $ join
...
+notDelegatingNotJoining :: ApiWalletDelegation
+notDelegatingNotJoining =
+    ApiWalletDelegation (ApiDelegationStatus $ ApiT NotDelegating) Nothing
+
+notDelegatingJoining
+    :: ApiT PoolId
+    -> EpochNo
+    -> SlotParameters
+    -> ApiWalletDelegation
+notDelegatingJoining poolId epochNo sp =
```
Anviking: [pure question]
Also should these live in the `DSL` module? Did you consider putting them in like `Api.Server.Types`?

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -1524,6 +1525,13 @@ instance Buildable PoolRegistrationCertificate where
         <> " owned by "
         <> build o
 
+data DelegationDiscovered = DelegationDiscovered
```
Anviking, line 1529: What is this? Documentation comment, please!

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -674,6 +677,101 @@ postShelleyWallet ctx body = do
...
+                        ApiT $ W.Delegating (ApiT poolId))
+                    Nothing
+    [ (W.DelegationDiscovered (W.SlotId epochN1 _) poolIdMaybe1)
+        , (W.DelegationDiscovered (W.SlotId epochN2 _) poolIdMaybe2)] ->
+        if epochN2 > epochN1 then
+            error "delegation discovered later cannot have epoch number \
+                  \older than the older discovered delegation"
+        else case (poolIdMaybe1, poolIdMaybe2) of
+            (Nothing, Nothing) ->
+                error "there should not be two quit delegation in row"
```
Anviking: If we end up here something is seriously wrong, and it would be good to know about. Could we trace the error or respond with a more detailed error message?

I believe the `error msg` will render as `500 Something went wrong.` to the user.

edit: 501 -> 500

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -674,6 +677,101 @@ postShelleyWallet ctx body = do
...
+                        (ApiEpochInfo
+                            (ApiT $ epochN1 + 2)
+                            (W.epochStartTime sp (epochN1 + 2)))
+                    )
+                else
+                    ApiWalletDelegation
+                    (ApiDelegationStatus $ ApiT W.NotDelegating)
+                    Nothing
+    _ ->
+        error "takeDelegationsDiscovered cannot return more than two \
```
Anviking: This `error` seems arise from the DB-function promises to return just 0 to 2 elements. Could this be expressed better via the type-system?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -674,6 +677,101 @@ postShelleyWallet ctx body = do
     wid = WalletId $ digest $ publicKey rootXPrv
     wName = getApiT (body ^. #name)
 
+toDelegationMsg
+    :: W.EpochNo
+    -> W.SlotParameters
+    -> [W.DelegationDiscovered]
+    -> ApiWalletDelegation
+toDelegationMsg currentEpochNo sp dlgs = case dlgs of
```
Anviking: [pure question] Did you consider having this logic in say `Cardano.Wallet`?

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: Same goes for the other two, i.e.
- "there should not be quiting delegation when no prior joining delegation is present"
- "delegation discovered later cannot have epoch number older than the older discovered delegation"

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
```diff
@@ -799,6 +823,79 @@ spec = do
...
+                , expectListItemFieldEqual 0 (#status . #getApiT) InLedger
+                ]
+        (currentEpochNo1, sp1) <- getSlotParams ctx
+        request @ApiWallet ctx (Link.getWallet @'Shelley w) Default Empty >>= flip verify
```
Anviking, line 819: This test whole test is very nice and readable. Perhaps it would be more so if we created a helper
```haskell
let getWallet = request @ApiWallet ctx (Link.getWallet @'Shelley w) Default Empty
```
for re-use and brevity?

What do you think?

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
```diff
@@ -477,33 +497,6 @@ spec = do
         let poolId = toText $ getApiT $ p ^. #id
         expectErrorMessage (errMsg403PoolAlreadyJoined poolId) r
 
-    it "STAKE_POOLS_JOIN_01 - \
```
paweljakubas, line 480: I delete this as it is replicated (and even more) in 
```
it "STAKE_POOLS_JOIN/QUIT - Checking delegation expectations
```
added below

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
```diff
@@ -799,6 +823,79 @@ spec = do
...
+        -- rejoining second pool
+        r2 <- joinStakePool ctx (p2 ^. #id) (w, fixturePassphrase)
+        expectResponseCode HTTP.status202 r2
+        eventually $ do
+            let ep = Link.listTransactions @'Shelley w
+            request @[ApiTransaction n] ctx ep Default Empty >>= flip verify
+                [ expectListItemFieldEqual 0 (#direction . #getApiT) Outgoing
+                , expectListItemFieldEqual 0 (#status . #getApiT) InLedger
+                , expectListItemFieldEqual 1 (#direction . #getApiT) Outgoing
+                , expectListItemFieldEqual 1 (#status . #getApiT) InLedger
```
Anviking, line 837: Maybe clarify why there are two transactions here.

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
```diff
@@ -799,6 +823,79 @@ spec = do
...
+                ]
+        (currentEpochNo3, sp3) <- getSlotParams ctx
+        request @ApiWallet ctx (Link.getWallet @'Shelley w) Default Empty >>= flip verify
+            [ expectFieldEqual #delegation
+                  (delegatingQuiting (p2 ^. #id) currentEpochNo3 sp3)
+            ]
+        eventually $ do
+            request @ApiWallet ctx (Link.getWallet @'Shelley w) Default Empty >>= flip verify
+                [ expectFieldEqual #delegation notDelegatingNotJoining
+                ]
```
Anviking, line 870: Nice test!

paweljakubas:
bors try

Anviking (commented):
Looks good! Nice integration test.

A bit concerned about the uses of `error` in `toDelegationMsg`.

lib/core-integration/src/Test/Integration/Framework/DSL.hs
Anviking: Reasoning: At *first glance* I had trouble correctly reading this as — "currently not delegating, about to join", where `joining` is to be read separately from `notDelegating`.

Short doc-comments would help too.

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
Anviking: In that case, the `HTTP 202` expectations should be moved too, I believe.

paweljakubas:
bors try

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: > If we end up here something is seriously wrong

I should clarify: I'm particularly worried because we have had bugs with the chain follower.

lib/core-integration/src/Test/Integration/Framework/DSL.hs
paweljakubas: indeed, corrected

lib/core-integration/src/Test/Integration/Framework/DSL.hs
paweljakubas: changed the names to better reflect what is going on, added comments and also moved everything from DSL to Types.Api : 
```
89       -- * Delegation status helpers                                                                                                                                                       
90       , notDelegating                                                                                                                                                                      
91       , notDelegatingAboutToJoin                                                                                                                                                           
92       , delegating                                                                                                                                                                         
93       , delegatingAboutToQuit                                                                                                                                                              
94       , delegatingAboutToRejoin                                                                                                                                                            
95       , toApiWalletDelegation
```

lib/core-integration/src/Test/Integration/Framework/DSL.hs
paweljakubas: this is very good idea, did it. and also moved toDelegationMsg there and reused those functions. Thanks to that the function shrunk in size and is better readable

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: I was thinking about it, but in the end think Api.Types is better place as I want to get ApiWalletDelegation values as a result

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: yeah, I knew you would make a complain here :-) Frankly speaking, every place where is error emitted means we have really huge problem with either DB or how it should work (eg. we cannot quit twice, etc). I need to think more about it frankly speaking :thinking:

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: maybe, but I think it is overstretch here. We use DBLayer function that suppose to return **at most** two elements (we have `take 2`), so I have this error here although I think it will "never" happen

lib/core/src/Cardano/Wallet/Primitive/Types.hs
paweljakubas: done

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
paweljakubas: they are corresponding : 
```
expectResponseCode HTTP.status202 rX
```
in the test that replaces this one

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
paweljakubas: yes, this means we discovered the certificate that was messaged as tx to the blockchain

paweljakubas: I could also use eventually with notDelegatingAboutToJoin....but I check it later

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
paweljakubas: yes, for sure it is good idea . But this is the challenge throughout all integration test. Maybe Piotr will tackle that in coming PRs cleaning this modules  . Now I am more focusing on logic and comply with convention presen throught code (in integration scenarios)

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: If we don't like adding proper error handling for "extreme cases", maybe we could `throw` a custom exception and then catch and render as `err501 { message = messageFromException }` somehow.

I.e. 

```haskell
throwInternalError "there should not be two quit delegation in row"
-- instead of 
error "there should not be two quit delegation in row"
```

But that might be a controversial solution.

For listing stake pools, we just have proper error cases for the supposedly impossible. https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/Api/Server.hs#L1985-L1991

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
paweljakubas: After rejoining I want to discover **two** certificates (which were communicated to blockchain as txs).
Likewise, after quitting I am waiting for another one (so want to have **three**)

specifications/api/swagger.yaml
paweljakubas: nice catch . changed to : 
```
Only present when a status change has to be reported.
```

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
Anviking: Fair enough.

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
Anviking: Sorry, meant in a comment. And maybe you could say which his the most recent one? (The 0th one, right?)

Or… perhaps it would just be cluttering…

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: I see. I will add then the error for "the supposedly impossible" also.

paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/1322/commits/9256bc9d417b474de22dd3315369e4327411dc3b

paweljakubas:
bors try

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1105,3 +1117,119 @@ gDecodeAddress decodeShelley text =
     tryBech32 = do
         (_, dp) <- either (const Nothing) Just (Bech32.decodeLenient text)
         dataPartToBytes dp
+
+-- not delegating and not about to join any stake pool
```
Anviking: Why not haddock comments? `-- |` 

Then we see this text in the generated Haddock docs.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1105,3 +1117,119 @@ gDecodeAddress decodeShelley text =
...
+notDelegating :: ApiWalletDelegation
+notDelegating =
+    ApiWalletDelegation (ApiDelegationStatus $ ApiT NotDelegating) Nothing
+
+-- not delegating, joined as it follows from discovered delegation certificates
+-- and effective joining epoch and its start time is specified
+notDelegatingAboutToJoin
+    :: ApiT PoolId
+    -> EpochNo
```
Anviking, line 1133: Small concrete advice: let's just add a `-- ^ Current epoch` to disambiguate.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
Anviking: Nice, how about:
```Haskell
-- | A delegation certificate discovered by the "current wallet".
data DelegationDiscovered = DelegationDiscovered
     { slotId :: !SlotId
      -- ^ Slot corresponding to the block the certificate was discovered.
     , poolId :: !(Maybe PoolId)
      -- ^ Stake pool identifier - in case the certificate pertains to (re)joining
      -- or @Nothing@ if quitting.
```
- Removed "Stores information of" since it seemed superfluous 
- Moved the `(a)` and `(b)` parts to their respective fields
- I added `belonging to the current wallet`. At first I thought "discovered" meant "discovered on chain", but I realise it probably means "discovered on chain and belonging to the current wallet" (similar to AddressDiscovery). I think the extra clarity is worth it. I presume this is _why_ you couldn't re-use the existing `DelegationCertificate` (which also contains the `ChimericAccount`).

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1105,3 +1117,119 @@ gDecodeAddress decodeShelley text =
     tryBech32 = do
         (_, dp) <- either (const Nothing) Just (Bech32.decodeLenient text)
         dataPartToBytes dp
+
+-- not delegating and not about to join any stake pool
+notDelegating :: ApiWalletDelegation
+notDelegating =
+    ApiWalletDelegation (ApiDelegationStatus $ ApiT NotDelegating) Nothing
+
+-- not delegating, joined as it follows from discovered delegation certificates
```
Anviking: > -- not delegating, **joined as it follows from** discovered delegation certificates
>  -- and effective joining epoch and its start time is specified
> notDelegatingAboutToJoin

I'm not sure I'd understand this.

If "it" were to be removed I would understand it better.

specifications/api/swagger.yaml
```diff
@@ -723,6 +705,35 @@ definitions:
...
+        type: object
+        <<: *ApiDelegationStatus
+        description: Currently active delegation status.
+        example:
+          status: delegating
+          target: 1423856bc91c49e928f6f30f4e8d665d53eb4ab6028bd0ac971809d514c92db1
+      next:
+        type: object
+        <<: *ApiDelegationStatus
+        description: Next delegation settings about to change. Only present when a status change has to be reported.
```
Anviking: Maybe
```
Next delegation status that is not yet in effect. Only present when a new delegation certificate has been submitted or discovered.
```
?

- `delegation settings` -> `delegation status` for consistency with ApiDelegationStatus above.
- `has to be` -> `has been`

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1105,3 +1119,125 @@ gDecodeAddress decodeShelley text =
...
+    -> SlotParameters
+    -> [DelegationDiscovered]
+    -> Either ErrDelegationsDiscoveredInconsistency ApiWalletDelegation
+toApiWalletDelegation currentEpochNo sp dlgs = case dlgs of
+    [] ->
+        Right notDelegating
+    [(DelegationDiscovered (SlotId epochN _) poolIdMaybe)] ->
+        case poolIdMaybe of
+            Nothing ->
+                err "there should not be quiting delegation when \
```
Anviking:
```suggestion
                err "there should not be quitting delegation when \
```
Now these could be displayed to the user!

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1831,6 +1845,15 @@ instance LiftHandler ErrMkTx where
                 , "Retrying may work, but something really went wrong..."
                 ]
 
+instance LiftHandler ErrDelegationsDiscoveredInconsistency where
+    handler = \case
+        ErrDelegationsDiscoveredInconsistency errMsg ->
+            apiError err500 InvalidDelegationDiscovery $ mconcat
+                [ "That's embarassing. I couldn't establish proper delegation "
+                , "status. This is the exact issue I have encountered : "
```
```suggestion
                , "status. This is the exact issue I have encountered: "
```

Anviking (commented):
I have left a few comments about English/comments etc. Minor - but I think that kind of stuff still deserve some attention.

I have a more high-level question about 
```Haskell
(ApiT $ epochNo + 2) (W.epochStartTime sp (epochNo + 2))
```
. This means that the delegation "becomes active" at the start of the second next epoch? Doesn't the delegation "become active" at the start of the next epoch, but that the first rewards are received at the start of the second next epoch?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1105,3 +1119,125 @@ gDecodeAddress decodeShelley text =
...
+        , (DelegationDiscovered (SlotId epochN2 _) poolIdMaybe2)] ->
+        if epochN2 > epochN1 then
+            err "delegation discovered later cannot have epoch number \
+                \older than the previously discovered delegation"
+        else case (poolIdMaybe1, poolIdMaybe2) of
+            (Nothing, Nothing) ->
+                err "there should not be two quit delegation certificates\
+                      \ discovered in row"
+            (Just pId, Nothing) ->
+                if currentEpochNo < epochN1 + 2 then
```
Anviking, line 1235: Idea: maybe we could remove the duplication of this `if currentEpochNo < epochN1 + 2` by first doing something like:
```haskell
let activeDelegation = safeLast $ filter (...) dlgs
let upcomingDelegation = safeLast $ filter (...) dlgs
case (activeDelegation, upcomingDelegation) of
...
```

I came to realise that this function is pretty complex and I hadn't checked how it works.

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: change the descriptions - hope they are much clearer now

specifications/api/swagger.yaml
paweljakubas: changed to : 
```
Next delegation status that is not yet in effect. Only present when a new delegation certificate has been discovered.
```
it has to be discovered. submitting does not affect this field

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: yeah, I was contemplating this several times. For me explicit if ... else clauses are the clearest to understand in this situation

paweljakubas:
@Anviking according to ADP-111 specification we have : 
```
The wallet engine must only report a wallet as being delegated when it actively IS (i.e. after the 2 epochs grace period upon discovering a delegation certificate).
```
I included this requirement in #1312 description of the expectations what happens when we discover delegation certificate throughout the time (counting epochs)

paweljakubas:
bors try

Anviking (approved):
Lgtm! (somewhat at a glance)

Modulo:
The exact U/S terminology about the `e + 2` seems weird to me. I'd confirm or note the exact definition in the api docs.

PR #1326, opened 2020-02-05
Check Bech32 human-readable prefix literals at compile time.

### LINKED ISSUE #1325: Remove `unsafeHumanReadablePartFromText` from `bech32` library.
# Context

We'd like to switch to using the Hackage version of the `bech32` library instead of our own copy (in `cardano-wallet/lib/bech32`). In order to do so, we must bring our local copy and the Hackage version into parity with one another.

Our local copy of the library has one extra function not present in the Hackage version: `unsafeHumanReadablePartFromText`

# Analysis

The `unsafeHumanReadablePartFromText` function throws an exception at run time if the specified input cannot be parsed. It's the only function that explicitly throws a run-time exception in a library of otherwise pure functions.

We currently only use this function to parse **_hard-coded string literals_**. Therefore, there is no need to defer the verification to run time. We can remove the function entirely and perform the parsing/checking at compile time instead.

# Decision

- [x] Remove the `unsafeHumanReadablePartFromText` function from our local copy of the `bech32` library.
- [x] Create a module in `cardano-wallet-core:lib` that defines and exports a Bech32 human-readable part (HRP) quasiquoter.
- [x] Use this quasiquoter to replace existing calls to `unsafeHumanReadablePartFromText`.

Example:
```hs
hrp = [Bech32.humanReadablePart|addr|]
```
 
# Acceptance Criteria

- [x] Our local copy of the `bech32` library _must not_ export the `unsafeHumanReadablePartFromText` function.
- [x] We _must_ have a quasiquoter capable of parsing human-readable prefix literals at compile time, so that attempting to use an invalid literal will result in a compile time error.
- [x] All existing human-readable prefix literals _must_ use the quasiquoter.
---

# Development 

As above.

# QA 

- [x] Migrate any coverage of the `unsafeHumanReadablePartFromText` function to the official version of the `bech32` library. (See https://github.com/input-output-hk/bech32/pull/7)

jonathanknowles:
# Issue Number

#1325 

# Overview

This PR:

- [x] Introduces a new quasiquoter `Bech32.humanReadablePart` that makes it possible to write Bech32 human-readable prefix literals that are checked for validity at **compile time**.
- [x] Removes the `unsafeHumanReadablePartFromText` function from our local copy of the `bech32` library.

A sample error produced when attempting to compile with an invalid Bech32 prefix literal:
```hs
/home/jsk/projects/input-output-hk/cardano-wallet/lib/core/src/Cardano/Wallet/Primitive/Types.hs:546:19: error:
    • Exception when trying to run compile-time code:
        HumanReadablePartContainsInvalidChars [CharPosition 0,CharPosition 1,CharPosition 2,CharPosition 3] when processing text "\33253\34382\34255\40845ed25519_pk".
  error, called at src/Codec/Binary/Bech32/TH.hs:47:20 in cardano-wallet-core-2020.1.27-CMaJoUKlZ2DGqxtjoUjXZ8:Codec.Binary.Bech32.TH
      Code: Language.Haskell.TH.Quote.quoteExp
              Bech32.humanReadablePart "\33253\34382\34255\40845ed25519_pk"
    • In the quasi-quotation:
        [Bech32.humanReadablePart|臥虎藏龍ed25519_pk|]
    |
546 | poolOwnerPrefix = [Bech32.humanReadablePart|臥虎藏龍ed25519_pk|]
```

### DIFF
--- a/lib/core/src/Codec/Binary/Bech32/TH.hs
+++ b/lib/core/src/Codec/Binary/Bech32/TH.hs
@@ -1,0 +1,58 @@
+{-# LANGUAGE TemplateHaskell #-}
+
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+-- |
+-- Copyright: © 2018-2020 IOHK
+-- License: Apache-2.0
+--
+-- This module contains Template-Haskell-specific extensions to the Bech32
+-- library.
+
+module Codec.Binary.Bech32.TH
+    ( humanReadablePart
+    ) where
+
+import Prelude
+
+import Codec.Binary.Bech32
+    ( HumanReadablePart, humanReadablePartFromText, humanReadablePartToText )
+import Language.Haskell.TH.Quote
+    ( QuasiQuoter (..) )
+import Language.Haskell.TH.Syntax
+    ( Lift, lift )
+
+import qualified Data.Text as T
+
+-- | A quasiquoter for Bech32 human-readable prefixes.
+--
+-- This allows the construction of prefixes to be checked at compile time.
+--
+humanReadablePart :: QuasiQuoter
+humanReadablePart = QuasiQuoter
+    { quoteExp  = lift . unsafeHumanReadablePartFromText . T.pack
+    , quotePat  = notHandled "patterns"
+    , quoteType = notHandled "types"
+    , quoteDec  = notHandled "declarations"
+    }
+  where
+    notHandled things =
+      error $ things <>
+          " are not handled by the Bech32 humanReadablePart quasiquoter."
+
+unsafeHumanReadablePartFromText :: T.Text -> HumanReadablePart
+unsafeHumanReadablePartFromText t =
+    either throwError id $ humanReadablePartFromText t
+  where
+    throwError e = error $ mconcat
+        [ show e
+        , " when processing text "
+        , show t
+        , "."
+        ]
+
+instance Lift HumanReadablePart where
+    lift t =
+        [|unsafeHumanReadablePartFromText $ T.pack
+            $(lift $ T.unpack $ humanReadablePartToText t)|]
+
--- a/lib/bech32/src/Codec/Binary/Bech32.hs
+++ b/lib/bech32/src/Codec/Binary/Bech32.hs
@@ -32,7 +32,6 @@ module Codec.Binary.Bech32
     , HumanReadablePartError (..)
     , humanReadablePartFromText
     , humanReadablePartToText
-    , unsafeHumanReadablePartFromText
     ) where
 
 import Codec.Binary.Bech32.Internal
--- a/lib/bech32/src/Codec/Binary/Bech32/Internal.hs
+++ b/lib/bech32/src/Codec/Binary/Bech32/Internal.hs
@@ -51,7 +51,6 @@ module Codec.Binary.Bech32.Internal
     , HumanReadablePart
     , HumanReadablePartError (..)
     , humanReadablePartFromText
-    , unsafeHumanReadablePartFromText
     , humanReadablePartToText
     , humanReadablePartToWords
     , humanReadablePartMinLength
@@ -108,8 +107,6 @@ import Data.Text.Class
     ( splitAtLastOccurrence )
 import Data.Word
     ( Word8 )
-import GHC.Stack
-    ( HasCallStack )
 
 import qualified Data.Array as Arr
 import qualified Data.ByteString as BS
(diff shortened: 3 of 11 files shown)


lib/core/src/Codec/Binary/Bech32/TH.hs
```diff
@@ -0,0 +1,58 @@
...
+        [ show e
+        , " when processing text "
+        , show t
+        , "."
+        ]
+
+instance Lift HumanReadablePart where
+    lift t =
+        [|unsafeHumanReadablePartFromText $ T.pack
+            $(lift $ T.unpack $ humanReadablePartToText t)|]
```
paweljakubas, line 57: do we need brackets here?

lib/core/src/Codec/Binary/Bech32/TH.hs
```diff
@@ -0,0 +1,58 @@
...
+    , quotePat  = notHandled "patterns"
+    , quoteType = notHandled "types"
+    , quoteDec  = notHandled "declarations"
+    }
+  where
+    notHandled things =
+      error $ things <>
+          " are not handled by the Bech32 humanReadablePart quasiquoter."
+
+unsafeHumanReadablePartFromText :: T.Text -> HumanReadablePart
```
paweljakubas, line 43: can you add (or maybe recover) some testing to that? In the previous checked-at-run-time we had `unsafeHumanReadablePartFromText` in `Bench32.Internal` and there were persuasive tests in `Bech32Spec`.

paweljakubas:
I understand the move here. To have check at compile time rather than during runtime. I also buy that when we opt for this way of checking human readable part then quasi quotes is a way to implement it. I wonder what exactly is rationale to opt for it? It is so important for us to have hrp proper that in other case we cannot even compile it? We do not want here "flexibility" (or a more ways to testing) and, if make mistake, then block it during run-time?

I am sure @Anviking  will have strong opinion on that which I am very interested to hear:-)

Anviking (approved):
I don't have a strong opinion here 🤷‍♂ 

Seems interesting/cool. The problem of mistyping hrps doesn't seem *that* likely for our uses.

So I'd share Pawel's questioning
> It is so important for us to have hrp proper that in other case we cannot even compile it?

But I don't know of any cons, so seems fine. Though I barely looked at the implementation.

paweljakubas (approved):

lib/core/src/Codec/Binary/Bech32/TH.hs
jonathanknowles: Yes, they do appear to be necessary. The `$()` is a Template Haskell expression.

lib/core/src/Codec/Binary/Bech32/TH.hs
jonathanknowles: @paweljakubas yes, definitely!

IMO these tests should really be part of the testing for the `humanReadablePartFromText` function. (The `unsafeHumanReadablePartFromText` was just a thin wrapper around that.)

Since the goal here is to move away from our internal copy of the `bech32` library and instead rely on the [actual `bech32` library](https://github.com/input-output-hk/bech32), I will make a further PR against that library, adding equivalent test coverage.

jonathanknowles:
@paweljakubas wrote:
> It is so important for us to have hrp proper that in other case we cannot even compile it?

It's not of critical importance, but there were two obvious choices here:

1. Add the `unsafe` function to the [actual `bech32` library](https://github.com/input-output-hk/bech32). Undesirable, since this adds exception-throwing code to what was otherwise a library of pure functions.

2. Remove the `unsafe` function from our copy of the `bech32` library, replacing it with something else. Since all our uses of the `unsafeHumanReadablePartFromText` function were checking hard-coded string literals, it wasn't necessary to defer these checks to run time. Because everything can be checked at compile time, we can do away with the `unsafe` function altogether. (It still remains internally, as a helper, but it's no longer exported.)

The impetus for doing this is that we want to eventually get rid of our local copy of `bech32` and use the [actual `bech32` library](https://github.com/input-output-hk/bech32).

If in future we ever need to parse strings constructed at run time (possibly from user-supplied input), then I would argue for using `humanReadablePartFromText` (which returns `Either`).

jonathanknowles:
@Anviking wrote:

> Seems interesting/cool. The problem of mistyping hrps doesn't seem _that_ likely for our uses.
> So I'd share Pawel's questioning
> > It is so important for us to have hrp proper that in other case we cannot even compile it?

See: https://github.com/input-output-hk/cardano-wallet/pull/1326#issuecomment-582702772

lib/core/src/Codec/Binary/Bech32/TH.hs
jonathanknowles: See here: https://github.com/input-output-hk/bech32/pull/7

KtorZ:
The more we get checked statically, the better. Neat! (Although, maybe a bit excessive for this particular one :sweat_smile:)

Issue #1327, opened 2020-02-05
Stake pools have weirdly high performance


Anviking:
# Context
 
We expect statistics to cause small pools to have high performance with small amount of data.

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | 2020.1.27  |
| Platform     | Mac OS |
| Installation | Internal build / from source |

# Steps to Reproduce

```bash
cardano-wallet-jormungandr stake-pool list | jq "." > response.json
```

Here is my sqlite db:
[stake-pools.sqlite.zip](https://github.com/input-output-hk/cardano-wallet/files/4160937/stake-pools.sqlite.zip)

## Expected behavior
Unclear. Supposedly there should be few pools with a performance of over 100%.

## Actual behavior

There is a top pool with 156 produced blocks, and performance of 325%.

(exact json from another build, but should be similar)
```json
{
    "saturation": 0.010251718287484165,
    "metrics": {
      "controlled_stake": {
        "quantity": 1160630473392,
        "unit": "lovelace"
      },
      "produced_blocks": {
        "quantity": 156,
        "unit": "block"
      }
    },
    "cost": {
      "quantity": 250000000,
      "unit": "lovelace"
    },
    "margin": {
      "quantity": 5,
      "unit": "percent"
    },
    "apparent_performance": 3.140912576100334,
    "metadata": {
      "homepage": "https://www.archpool.io",
      "owner": "ed25519_pk1th5pgvs3cg7qjwq5gw8rmf4jh6fdpckp37flpjuxfzv75gkfvr3syxxcpm",
      "name": "ARCH Staking Pool",
      "ticker": "ARCH1",
      "pledge_address": "addr1s4w7s9pjz8prczfcz3pcu0dxk2lf958zcx8e8uxtseyfn63ze9swxug37z7",
      "description": "Archimedes Staking Pool for ADA Cardano"
    },
    "id": "08b46d4bc5d8110c76dde2cc7a4c9dda0830c4434dedd7fcd32b1a4815c1a4f3",
    "desirability": 102767222060.36679
  }
```

---

# Resolution 

1. Improve logging in `Performance` module.
2. Follow the spec more strictly on the performance calculation

---

# QA

Anviking:
Here's the pool: https://pooltool.io/pool/08b46d4bc5d8110c76dde2cc7a4c9dda0830c4434dedd7fcd32b1a4815c1a4f3

Anviking:
[...]

PR #1329, opened 2020-02-05
Debug high pool performance

Anviking:
# Issue Number

#1327

# Overview

Partly-throwaway PR for investigating pool performance.

# Comments

- Hopefully some aspects of the logging could make it into master at the end.

### DIFF
--- a/lib/core/src/Cardano/Pool/Performance.hs
+++ b/lib/core/src/Cardano/Pool/Performance.hs
@@ -139,14 +139,20 @@ apparentPerformance
         -- ^ Average performance
 apparentPerformance     [] = 0
 apparentPerformance epochs =
-    sum (n <$> epochs) / max 1 (sum (μ <$> epochs))
+    avg $ p <$> epochs
   where
-    n = double . poolProduction
-    μ e = _N * s / _S
+    p e = beta / sigma
       where
+        sigma = s / (max 1 _S) -- TODO: _S cannot be 0? throw?
+        beta = n / (max 1 _N)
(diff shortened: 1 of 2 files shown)


lib/core/src/Cardano/Pool/Performance.hs
```diff
@@ -103,25 +139,43 @@ apparentPerformance
...
-        <*> (Map.fromList <$> readStakeDistribution ep)
+readPoolsPerformances DBLayer{..} currentEpoch tr = do
+    stats <- atomically $ forM historicalEpochs $ \ep -> do
+        prod <- count <$> readPoolProduction ep
+        stake <- Map.fromList <$> readStakeDistribution ep
+        let stats = mkEpochStats prod stake
+
+        -- FOR DEBUGGING
+        let p = PoolId $ unsafeFromHex "08b46d4bc5d8110c76dde2cc7a4c9dda0830c4434dedd7fcd32b1a4815c1a4f3"
+        case Map.lookup p stats of
```
Anviking, line 172: If we have ~1000 pools, that seems too big to just log in its entirety.

Improving logging in master might be tricky. Perhaps we could aggregate and log some statistics of the distributions of pool performance, or which epochs we have stake distributions for 🤷‍♂

paweljakubas: maybe log only outliers? overperformers and underperformers by some margin. and also log newly discovered stake pools and say explicitly what performance they are attributed?

PR #1334, opened 2020-02-07
Suppress automatic migrations for pre-existing wallet databases.

jonathanknowles:
# Issue Number

#1279 (wallet shows balance of **zero** after upgrade from older version)

In particular, see the following comments:
- https://github.com/input-output-hk/cardano-wallet/issues/1279#issuecomment-577213679
- https://github.com/input-output-hk/cardano-wallet/issues/1279#issuecomment-578242895

# Overview

This PR:
- [x] reverts the changes introduced in PR #1290.
- [x] suppresses automatic migrations for **_pre-existing file-based wallet databases_**, so that we rely entirely on the manual migration mechanism.

However, we still allow automatic migrations **_when creating new wallet databases_**. This is necessary, because `persistent` relies on the automatic migration mechanism in order to build new databases from scratch.

# Reproduction steps

1. Start with the `v2020-01-14` build (or earlier) of the wallet backend (`cardano-wallet`).
2. Create a Jörmungandr wallet with a non-zero balance `b`. (For example, create a wallet with the passphrase `vintage poem topic machine hazard cement dune glimpse fix brief account badge mass silly business`.)
3. Stop the wallet backend, but do not delete the database files created by step 2.
4. Start with the `v2020-01-21` build (or later) of `cardano-wallet`.
5. Observe the new balance of the wallet.

## Behaviour before application of this PR

The balance is 0.

## Behaviour after application of this PR

The balance is `b`.

### DIFF
--- a/lib/core/src/Cardano/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/DB/Sqlite.hs
@@ -32,6 +32,9 @@ module Cardano.DB.Sqlite
     , startSqliteBackend
     , unsafeRunQuery
 
+    -- * Automatic migration
+    , AutoMigration (..)
+
     -- * Manual Migration
     , ManualMigration (..)
     , MigrationError (..)
@@ -204,30 +207,59 @@ destroyDBLayer (SqliteContext {getSqlBackend, trace, dbFile}) = do
                            Internal / Database Setup
 -------------------------------------------------------------------------------}
 
+-- | Represents an automatic migration to be performed on a database.
+data AutoMigration
+    = AutoMigrationNone
+      -- ^ Do not perform an automatic migration.
+    | AutoMigrationOnCreate Migration
+      -- ^ Perform the specified automatic migration when creating a new
+      -- database.
+    | AutoMigrationOnCreateOrRestart Migration
+      -- ^ Perform the specified automatic migration when creating a new
+      -- database or when restarting a pre-existing database.
+
+-- | Indicates whether or not a database is already populated with data.
+data DbPopulationStatus
+    = DbIsEmpty
+      -- ^ Indicates that a database is empty: the database is without any
+      -- tables or data.
+    | DbIsPopulated
+      -- ^ Indicates that a database is not empty: the database has at least
+      -- one table.
+
 -- | Opens the SQLite database connection, sets up query logging and timing,
 -- runs schema migrations if necessary.
 startSqliteBackend
     :: ManualMigration
-    -> Migration
+    -> AutoMigration
     -> Tracer IO DBLog
     -> Maybe FilePath
     -> IO (Either MigrationError SqliteContext)
-startSqliteBackend manualMigration migrateAll trace fp = do
-    backend <- createSqliteBackend trace fp manualMigration (queryLogFunc trace)
+startSqliteBackend manualMigration autoMigration trace fp = do
+    (backend, dbPopulationStatus) <-
+        createSqliteBackend trace fp manualMigration (queryLogFunc trace)
     lock <- newMVar ()
     let traceRun = traceWith trace . MsgRun
     let observe :: IO a -> IO a
         observe = bracket_ (traceRun False) (traceRun True)
     let runQuery :: SqlPersistT IO a -> IO a
         runQuery cmd = withMVar lock $ const $ observe $ runSqlConn cmd backend
-    migrations <- runQuery (runMigrationQuiet migrateAll)
-        & tryJust (matchMigrationError @PersistException)
-        & tryJust (matchMigrationError @SqliteException)
-        & fmap join
-        :: IO (Either MigrationError [Text])
-    traceWith trace $ MsgMigrations (fmap length migrations)
+    let requiredAutoMigration = case (autoMigration, dbPopulationStatus) of
+            (AutoMigrationNone               , _            ) -> Nothing
+            (AutoMigrationOnCreateOrRestart m, DbIsEmpty    ) -> Just m
+            (AutoMigrationOnCreateOrRestart m, DbIsPopulated) -> Just m
+            (AutoMigrationOnCreate          m, DbIsEmpty    ) -> Just m
+            (AutoMigrationOnCreate          _, DbIsPopulated) -> Nothing
+    autoMigrationResult <- case requiredAutoMigration of
+        Nothing -> pure $ pure mempty
+        Just am -> runQuery (runMigrationQuiet am)
+            & tryJust (matchMigrationError @PersistException)
+            & tryJust (matchMigrationError @SqliteException)
+            & fmap join
+            :: IO (Either MigrationError [Text])
+    traceWith trace $ MsgMigrations $ fmap length autoMigrationResult
     let ctx = SqliteContext backend runQuery fp trace
-    case migrations of
+    case autoMigrationResult of
         Left e -> do
             destroyDBLayer ctx
             pure $ Left e
@@ -257,18 +289,31 @@ instance MatchMigrationError SqliteException where
 newtype ManualMigration = ManualMigration
     { executeManualMigration :: Sqlite.Connection -> IO () }
 
+-- | Determines whether or not the given database is populated with data.
+determineDbPopulationStatus :: Sqlite.Connection -> IO DbPopulationStatus
+determineDbPopulationStatus conn = do
+    query <- Sqlite.prepare conn "select count (*) from sqlite_master;"
+    count <- Sqlite.step query >> Sqlite.columns query
+    Sqlite.finalize query
+    pure $ case count of
+        [Persist.PersistInt64 0] -> DbIsEmpty
+        [Persist.PersistInt64 _] -> DbIsPopulated
+        _ -> error "Unexpected result when determining DB population status."
+
 createSqliteBackend
     :: Tracer IO DBLog
     -> Maybe FilePath
     -> ManualMigration
     -> LogFunc
-    -> IO SqlBackend
+    -> IO (SqlBackend, DbPopulationStatus)
 createSqliteBackend trace fp migration logFunc = do
     let connStr = sqliteConnStr fp
     traceWith trace $ MsgConnStr connStr
     conn <- Sqlite.open connStr
+    dbPopulationStatus <- determineDbPopulationStatus conn
     executeManualMigration migration conn
-    wrapConnectionInfo (mkSqliteConnectionInfo connStr) conn logFunc
+    backend <- wrapConnectionInfo (mkSqliteConnectionInfo connStr) conn logFunc
+    pure (backend, dbPopulationStatus)
 
 sqliteConnStr :: Maybe FilePath -> Text
 sqliteConnStr = maybe ":memory:" T.pack
--- a/lib/core/src/Cardano/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/DB/Sqlite.hs
@@ -76,8 +76,6 @@ import Data.Maybe
     ( fromMaybe )
 import Data.Proxy
     ( Proxy (..) )
-import Data.Quantity
-    ( Quantity (..) )
 import Data.Text
     ( Text )
 import Data.Text.Class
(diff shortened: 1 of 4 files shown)


piotr-iohk:
@jonathanknowles  I have tried manually testing on the branch, for instance:

`v2020-01-14` -> `jonathanknowles/restrict-auto-migration`

There is NO 0 balance issue, I.e. funds are there on byron faucet wallet after migration.
One thing that is bothering me is that in the log there is `No database migrations were necessary.`:
```
[cardano-wallet.pools-db:Notice:20] [2020-02-10 10:19:51.35 UTC] No database migrations were necessary.
[cardano-wallet.application:Info:20] [2020-02-10 10:19:51.35 UTC] Found existing wallet: aa129a07d1ce083e67597348f1788747a034686e
[cardano-wallet.pools-engine:Info:27] [2020-02-10 10:19:51.35 UTC] Monitoring stake pools. Currently at 754cebd6-[1256396.3#41]
[cardano-wallet.wallet-db:Notice:28] [2020-02-10 10:19:51.35 UTC] No database migrations were necessary.
[cardano-wallet.application:Info:20] [2020-02-10 10:19:51.36 UTC] Found existing wallet: 2d4cc31a4b3116ab86bfe529d30d9c362acd0b44
[cardano-wallet.wallet-db:Notice:31] [2020-02-10 10:19:51.37 UTC] No database migrations were necessary.

```

Whereas when migrating: `v2020-01-14` -> `v2020-01-27` in the log there is:

```
[cardano-wallet.wallet-db:Notice:28] [2020-02-10 10:57:56.94 UTC] 48 migrations were applied to the database.
[cardano-wallet.wallet-db:Notice:28] [2020-02-10 10:57:56.94 UTC] Some automated database migrations happened. Forced to roll back to genesis.
```

lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -206,35 +207,63 @@ destroyDBLayer (SqliteContext {getSqlBackend, trace, dbFile}) = do
...
+-- | Represents an automatic migration to be performed on a database.
+data AutoMigration
+    = AutoMigrationNone
+      -- ^ Do not perform an automatic migration.
+    | AutoMigrationOnCreate Migration
+      -- ^ Perform the specified automatic migration when creating a new
+      -- database.
+    | AutoMigrationOnCreateOrRestart Migration
+      -- ^ Perform the specified automatic migration when creating a new
+      -- database or when restarting a pre-existing database.
```
KtorZ, line 217: What about simply calling this `AutoMigrationAlways` ?

lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -206,35 +207,63 @@ destroyDBLayer (SqliteContext {getSqlBackend, trace, dbFile}) = do
                            Internal / Database Setup
 -------------------------------------------------------------------------------}
 
+-- | Represents an automatic migration to be performed on a database.
+data AutoMigration
+    = AutoMigrationNone
+      -- ^ Do not perform an automatic migration.
```
KtorZ: This case seems unused? I can see the conceptual reason to have it, but in the meantime, why bother?

lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -141,15 +136,13 @@ newDBLayer
...
-    pure (ctx, dbLayer)
-  where
-    mkDBLayer SqliteContext{runQuery} = DBLayer
+    let io = startSqliteBackend
+            (ManualMigration mempty)
+            (AutoMigrationOnCreateOrRestart migrateAll)
+            trace
+            fp
+    ctx@SqliteContext{runQuery} <- handlingPersistError trace fp io
+    return (ctx, DBLayer
```
KtorZ, line 145: Hmmm. This would however re-introduce a bug in the stake pool database. For the same reason as in the wallet database (foreign references and cascade deletion), upon automatic migrations of `pool_registration` table, we would loose all data in the `pool_owner` table, creating an inconsistency. So, I am afraid that we also need an `AutoMigrationOnCreate` here too. Which seemingly makes the `AutoMigrationOnCreate` the only actual used branch for the `AutoMigration` data-type :) ... making the type itself somewhat redundant.

jonathanknowles:
@piotr-iohk wrote:
>  One thing that is bothering me is that in the log there is `No database migrations were necessary.`:

Well spotted! That's the result of a deliberate change: https://github.com/input-output-hk/cardano-wallet/pull/1334/commits/72ce159701e1fc98afe8108b05b0ee16a2fc9a83

**Reasoning**: unless I'm mistaken (please correct me if I am mistaken), items with the `Debug` log severity level are (by default) **_not included_** in user-supplied logs.

But given the difficulty of debugging database migration failures in the field, it seems to me that it would be rather useful to be able to see clear evidence that **_no migrations were applied_** on a particular startup.

What do you think?

lib/core/src/Cardano/DB/Sqlite.hs
jonathanknowles: @KtorZ wrote:
> This case seems unused? I can see the conceptual reason to have it, but in the meantime, why bother?

It has actually been rather useful during testing to be able to temporarily change the migration type to `AutoMigrationNone`. Nevertheless, it's easy to add back in if we ever need it, so I have no objection to removing it.

**Fixed, see:** https://github.com/input-output-hk/cardano-wallet/commit/6f9bddb18f01826162dc924251cf64e90e71391e

lib/core/src/Cardano/DB/Sqlite.hs
jonathanknowles: @KtorZ wrote:

> What about simply calling this `AutoMigrationAlways` ?

Fair enough! I don't have a strong opinion about this. I've changed it to match your suggestion.

**Fixed in:** https://github.com/input-output-hk/cardano-wallet/pull/1334/commits/50978cc5084eb6c892e617ef2fc890925a136fa4

lib/core/src/Cardano/Pool/DB/Sqlite.hs
jonathanknowles: @KtorZ wrote:

> Hmmm. This would however re-introduce a bug in the stake pool database. For the same reason as in the wallet database (foreign references and cascade deletion), upon automatic migrations of `pool_registration` table, we would loose all data in the `pool_owner` table, creating an inconsistency. So, I am afraid that we also need an `AutoMigrationOnCreate` here too. 

Hmm, I'm not convinced by this argument.

My reason for introducing the `AutoMigration` type is so that we can:

- **Change** the current behaviour for **wallet** databases (so that we perform automatic migrations **only** when creating new databases); but
- **Preserve** the current behaviour for **pool** databases, which is to **always** run migrations, even when attempting to load old databases.

Indeed, with this PR applied, when we attempt to automatically migrate an old pool database, we see the following failure, which is identical to what we see on `master`:
```hs
[cardano-wallet.pools-db:Error:20] [2020-02-11 04:43:25.31 UTC] Failed to migrate the database: : NOT NULL constraint failed: pool_owner_backup.pool_owner_index
```
It's the failure of this automatic migration that triggers our desired **_database reset_**:
```hs
[cardano-wallet.pools-db:Notice:20] [2020-02-11 04:43:25.31 UTC] Non backward compatible database found. Removing old database and re-creating it from scratch. Ignore the previous error.
[cardano-wallet.pools-db:Notice:20] [2020-02-11 04:43:25.32 UTC] 5 migrations were applied to the database.
[cardano-wallet.pools-engine:Info:27] [2020-02-11 04:43:25.33 UTC] Monitoring stake pools. Currently at genesis
[cardano-wallet.pools-engine:Info:27] [2020-02-11 04:43:25.38 UTC] Applying blocks [1242320.6 ... 1259707.4]
[cardano-wallet.pools-engine:Info:27] [2020-02-11 04:43:25.38 UTC] Writing stake-distribution for epoch 1259707
```
This works because we have the following code to handle the automatic migration failure:
```hs
handlingPersistError trace fp action = action >>= \case
    Right ctx -> pure ctx
    Left _ -> do
        traceWith trace MsgDatabaseReset
        maybe (pure ()) removeFile fp
        action >>= either throwIO pure
```

However, if we change the pool DB to use the `AutoMigrationOnCreate` strategy, then loading a pre-existing old database will no longer cause an automatic migration failure, and we'll no longer trigger the `handlingPersistError` handler, and the DB will no longer be reset, resulting in the following error on startup:
```hs
[cardano-wallet.pools-db:Notice:23] [2020-02-11 04:46:20.75 UTC] No database migrations were necessary.
[cardano-wallet.pools-engine:Error:28] [2020-02-11 04:46:20.76 UTC] Worker has exited unexpectedly: SQLite3 returned ErrorError while attempting to perform prepare "SELECT \"pool_id\", \"slot\", \"margin\", \"cost\" FROM \"pool_registration\" WHERE (\"pool_id\"=?) LIMIT 1": no such table: pool_registration
```

So it seems to me that in the case of pool databases, preserving the current behaviour is actually what we want, because we're relying on the presence of an auto-migrate failure to trigger our database reset.

KtorZ:
> That's the result of a deliberate change. Unless I'm mistaken (please correct me if I am mistaken), items with the Debug log severity level are (by default) not included in user-supplied logs.

I agree with your statement here. Although, I disagree about putting this log line to `Notice`. The fact that no migration were applied is not something that should catch our attention. It's actually the opposite, hence the 'Debug' level in the past.

jonathanknowles:
@KtorZ wrote:
> I disagree about putting this log line to `Notice`. The fact that no migration were applied is not something that should catch our attention. It's actually the opposite, hence the 'Debug' level in the past.

Fair enough! I've removed this particular change. (Originally change 72ce159701e1fc98afe8108b05b0ee16a2fc9a83)

My motivation here was really to lower the cognitive burden of a future developer who is attempting to debug something that has gone wrong in the field. It could be useful to see _positive evidence_ that the database was loaded _without_ performing migrations.

Of course, it's possible to _deduce_ this by noticing the absence of "n migrations performed" in the log, but it's (arguably) more work to determine this.

On the other hand, I can sympathize with the desire to keep logs as quiet as possible (by default).

lib/core/src/Cardano/Pool/DB/Sqlite.hs
KtorZ: @jonathanknowles I am not arguing against the type so-to-speak, the problem isn't the type, the problem is the logic being applied. 
I agree, in case of migration errors, the pool database will reset as expected. But the issue I am talking about (which causes the 0-balance issue) will occurs on successful migrations!

Remember how persistent migrate the SQLite tables:

- copy the content of the table to a temporary table
- drop the table 
- recreate the new modified table
- copy back the data

During the second step, dropping the table will cause some other tables to loose data because of the foreign references and cascading delete. So even upon a successful migration, we may loose data!

jonathanknowles: @KtorZ wrote:
> But the issue I am talking about (which causes the 0-balance issue) will occurs on successful migrations!

Fair enough. In that case, we can adopt the approach that you suggest, of running automatic migrations _**only when we create a new database**_, for both pools and wallets.

Though for the pool DB, we still need to do a bit of extra work, since disabling automatic migrations for pre-existing DBs causes the following error on startup if we start with an old DB (for example, from last year):
```hs
[cardano-wallet.pools-engine:Error:28] [2020-02-11 04:46:20.76 UTC] Worker has exited unexpectedly: SQLite3 returned ErrorError while attempting to perform prepare "SELECT \"pool_id\", \"slot\", \"margin\", \"cost\" FROM \"pool_registration\" WHERE (\"pool_id\"=?) LIMIT 1": no such table: pool_registration
```

I'll add a patch to fix this.

KtorZ: Also, make sure to execute the automatic db migration tests! 

```
rm -rf state-migration-test*
nix-build nix/migration-tests.nix -o migration-tests
./migration-tests/runall.sh
```

jonathanknowles:
Closing, as the solution provided in PR #1340 supercedes this one.

Issue #1333, opened 2020-02-07
Remove local copies of Bech32 libraries.


jonathanknowles:
# Context

We'd like to switch to using the Hackage version of the [Bech32 libraries](https://github.com/input-output-hk/bech32) instead of our own copy (in `cardano-wallet/lib/bech32`).

# Decision

After bringing the published libraries and our local copies into parity with one another, we will remove our local copies of the libraries, and depend only on the [published versions](https://github.com/input-output-hk/bech32).

# Acceptance Criteria

- [x] The `cardano-wallet` code base _must_ only depend on the published [Bech32 libraries](https://github.com/input-output-hk/bech32).
- [x] The `cardano-wallet` repository _must not_ include local duplicates of the [Bech32 libraries](https://github.com/input-output-hk/bech32).
- [x] All published libraries _must_ be available on Hackage.
- [x] All published libraries _must_ be available on Stackage.

# Rejected Acceptance Criteria

- [x] ~~The `cardano-wallet` repository _may_ include a reference to the [Bech32 repository](https://github.com/input-output-hk/bech32) via a `git` submodule.~~
    **Reasons for rejection:**
    1. Including the published libraries as extra dependencies in `stack.yaml` is a simple solution that works well enough.
    2. We don't currently have a compelling use case for including a submodule. If such a use case appears in future, we can revisit the idea of adding a submodule.

# Development 

- [x] Remove local copy of `bech32` library, and rely instead of the Hackage version.
- [x] Create a new library `bech32-th` to contain Template-Haskell-specific extensions to the `bech32` library. (See https://github.com/input-output-hk/bech32/pull/15)
- [x] Add the `bech32-th` library to Hackage.
- [x] Remove our local copy of `Codec.Binary.Bech32.TH` and instead depend on the Hackage version of `bech32-th`.
- [x] Add the `bech32-th` library to Stackage. (https://github.com/commercialhaskell/stackage/pull/5196).

# QA 

- [x] Ensure that the `bech32-th` library has a suitable test suite. At the very least, we want to check that compile errors happen when they should. (See https://github.com/input-output-hk/bech32/pull/15)
- [x] Ensure that the `bech32` repository CI script runs all test suites (for both `bech32` and `bech32-th`).  (See https://github.com/input-output-hk/bech32/pull/15)

KtorZ:
[...]

PR #1336, opened 2020-02-10
Remove local copy of `bech32` library.

jonathanknowles:
# Issue Number

#1333 

# Overview

This PR:

- [x] Removes our local copy of the `bech32` library.
- [x] Specifies [version `1.0.1` of the library from Hackage](https://hackage.haskell.org/package/bech32-1.0.1) as an extra dependency within `stack.yaml`.
- [x] Updates all the relevant `nix` machinery.


KtorZ:
@jonathanknowles we should also update the `stack.yaml` file to use the "remote" bech32 as an external dep instead of a local package. For this one at least, I think it makes sense to treat it as an external lib. What do you think?

Pros:

- We don't have to rebuild it all the time
- Tests will be executed on a different CI for the bech32 package

Cons:

- Can be counter-intuitive if we make modifications to the submodule without changing the reference in the stack.yaml and expect them to show up straight away in the rest of the code.

jonathanknowles:
> @jonathanknowles we should also update the `stack.yaml` file to use the "remote" bech32 as an external dep instead of a local package. For this one at least, I think it makes sense to treat it as an external lib. What do you think?

I'd definitely be happy with adding it as an  `extra-dep` within `stack.yaml`.

If we do that though, should we remove the git submodule, to keep things simple?

(If later on we need to start working intensively with the `bech32` code base later in such a way that it requires iteration within  `cardano-wallet`, then perhaps it will be worthwhile adding a submodule back in again.)

I was going to suggest just relying on `stackage` to select the latest version for us, but `bech32` still isn't included in any LTS versions. (Though it is starting to appear in the [nightly builds](https://www.stackage.org/package/bech32/snapshots), starting from the end of last week.)

So we could add an entry in `extra-deps` for now, and then remove it later on after `stackage` includes it in an LTS version.

What do you think?

jonathanknowles:
@KtorZ 

I've updated the PR so that it:

- [x] Specifies [version `1.0.1` of the library from Hackage](https://hackage.haskell.org/package/bech32-1.0.1) as an extra dependency within `stack.yaml`.
- [x] Updates all the relevant `nix` machinery.

I've **removed** the `git` submodule for now. If we need it later on, we can add it back in again.

KtorZ (approved):

PR #1337, opened 2020-02-10
Correct the delegation certificates DB retrieval logic and move the logic downstream

paweljakubas:
# Issue Number

#1342 

# Overview

- [x] I have updated DB dlg cert logic, moved responsibility by extending WalletDelegation and adjusted rest of the code 
- [x] I have updated swagger to embrace multiple next's
- [x] I have adjusted WalletDelegation' next change and adjusted rest of the code and everything is passing
- [x] I have refined logic to account for the cases with more than one next 
- [x] I have added integration test with more than one next
- [x] cleaning and better reuse 

# Comments


paweljakubas:
bors try

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -515,9 +515,22 @@ newDBLayer trace defaultFieldValues fp = do
             selectWallet wid >>= \case
                 Nothing -> pure $ Left $ ErrNoSuchWallet wid
                 Just _ -> do
-                    Right . fmap (delegationDiscoveredFromEntity . entityVal)
-                        . take 2
+                    latest <- fmap entityVal . take 1
                         <$> selectList [CertWalletId ==. wid] [Desc CertSlot]
+                    case latest of
+                        [(DelegationCertificate _ (W.SlotId epoch _) _)] ->
```
paweljakubas: here, I should probably rely on current epoch not epochNo from last discovered certificate as it could be the one in the last epochs

KtorZ (changes requested):
We should definitely move the logic that is Api.Types down below. Also, the `WalletDelegation` type needs to be updated to reflect this! Right now, it can't capture the "next" status, hence why I think all the logic happens in the API. But, `WalletDelegation` from `Primitive.Types` should be updated, and then, populated correctly directly when resolved in the `Sqlite` handlers.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1388,7 +1373,13 @@ joinStakePool ctx wid (pid, pools) argGenChange pwd = db & \DBLayer{..} -> do
     walMeta <- mapExceptT atomically $ withExceptT ErrJoinStakePoolNoSuchWallet $
         withNoSuchWallet wid $ readWalletMeta (PrimaryKey wid)
 
-    when (walMeta ^. #delegation == Delegating pid) $
+    let checkAlreadyJoined (WalletDelegation (Delegating pid') Nothing) =
+            pid' == pid
+        checkAlreadyJoined (WalletDelegation _ (Just (WalletDelegationNext (Delegating pid') _))) =
+            pid' == pid
+        checkAlreadyJoined _ = False
```
KtorZ: What about moving this function declaration to the where close and use a lambda case instead?

- Avoid repeating the function's name
- Keep the handler's body neat

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1431,7 +1422,13 @@ quitStakePool ctx wid pid argGenChange pwd = db & \DBLayer{..} -> do
...
         withNoSuchWallet wid $ readWalletMeta (PrimaryKey wid)
 
-    when (walMeta ^. #delegation /= Delegating pid) $
+    let checkIfJoined (WalletDelegation (Delegating pid') Nothing) =
+            pid' /= pid
+        checkIfJoined (WalletDelegation _ (Just (WalletDelegationNext (Delegating pid') _))) =
+            pid' /= pid
+        checkIfJoined _ = True
+
+    when (checkIfJoined (walMeta ^. #delegation)) $
```
KtorZ: Same as for `joinStakePool` :+1: 

Alternatively, these two functions could be moved close to the declaration of the `WalletDelegation` type, as helpers / accessors on the data-type.

paweljakubas:
bors try

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -505,19 +505,18 @@ newDBLayer trace defaultFieldValues fp = do
...
+                Just cp -> do
+                    latestMaybe <- fmap entityVal
+                        <$> selectFirst [CertWalletId ==. wid] [Desc CertSlot]
+                    case latestMaybe of
+                        Just dlg -> do
+                            let (W.SlotId currentEpoch _) = checkpointSlot cp
+                            determineWalletDelegation wid dlg currentEpoch
+                        Nothing ->
+                            retrieveWalletMetadata wid
+                            (delegationFromCerts Nothing Nothing)
```
KtorZ: I am not sure to understand the distinction here? Why is there 2 paths for figuring out what the delegation status is?

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -586,22 +585,95 @@ newDBLayer trace defaultFieldValues fp = do
 
         }
 
-delegationDiscoveredFromEntity
+retrieveWalletMetadata
+    :: W.WalletId
+    -> W.WalletDelegation W.PoolId
+    -> SqlPersistT IO (Maybe W.WalletMetadata)
+retrieveWalletMetadata wid walDel =
```
KtorZ: suggestion: `readWalletMetadata`, for consistency with other methods.

paweljakubas:
bors try

paweljakubas:
bors try

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
paweljakubas: right, moved downstream

paweljakubas:
bors try

paweljakubas:
bors try

lib/core/src/Cardano/Wallet.hs

lib/core/src/Cardano/Wallet.hs

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
```diff
@@ -910,6 +906,123 @@ spec = do
...
+            request @ApiWallet ctx (Link.getWallet @'Shelley w) Default Empty >>= flip verify
+                [ expectField (#tip . #epochNumber . #getApiT)
+                    (`shouldBe` (EpochNo $ currentEpochNo1' + 3))
+                ]
+        request @ApiWallet ctx (Link.getWallet @'Shelley w) Default Empty >>= flip verify
+            [ expectField #delegation
+                (`shouldBe` delegatingAboutToRejoin
+                    (p1 ^. #id) (p3 ^. #id) (EpochNo $ currentEpochNo1'+2) sp1)
+            ]
+
```
paweljakubas, line 961: so here we are going through consecutive epochs (consequence of new `Buildable` instance of `WalletDelegation` in `Primitive.Types`): 
```
[cardano-wallet.wallet-engine:Info:61] [2020-02-13 11:18:04.16 UTC] 2d4cc31a: Faucet Wallet, created at 2020-02-13 11:18:03.948464054 UTC, current wallet delegation: not delegating, awaiting no change
[cardano-wallet.wallet-engine:Info:61] [2020-02-13 11:18:25.41 UTC] 2d4cc31a: Faucet Wallet, created at 2020-02-13 11:18:03.948464054 UTC, current wallet delegation: not delegating, awaiting delegating to 712dd028 which is expected to happen at epoch number 1269534
[cardano-wallet.wallet-engine:Info:61] [2020-02-13 11:18:49.17 UTC] 2d4cc31a: Faucet Wallet, created at 2020-02-13 11:18:03.948464054 UTC, current wallet delegation: not delegating, awaiting first delegating to 712dd028 which is expected to happen at epoch number 1269534 then awaiting delegating to 494b7852 which is expected to happen at epoch number 1269535
[cardano-wallet.wallet-engine:Info:61] [2020-02-13 11:19:09.17 UTC] 2d4cc31a: Faucet Wallet, created at 2020-02-13 11:18:03.948464054 UTC, current wallet delegation: delegating to 712dd028, awaiting delegating to 494b7852 which is expected to happen at epoch number 1269535
```

lib/core/src/Cardano/Wallet/DB/Model.hs
```diff
@@ -291,29 +291,81 @@ mReadWalletMeta wid db@(Database wallets _) =
...
-  where
-    res = maybe mempty mkDelegations $ Map.lookup wid wallets
-    mkDelegations :: WalletDatabase s xprv -> [DelegationDiscovered]
-    mkDelegations WalletDatabase{certificates} =
-        map toDelegationDiscovered $
-        L.sortOn (Down . fst) $
-        Map.toList certificates
-    toDelegationDiscovered (slotId, poolIdM) =
-        DelegationDiscovered slotId poolIdM
+    mkMetadata WalletDatabase{checkpoints,certificates,metadata} =
```
paweljakubas: this stuff needs further refactoring and enhancement. Now it passes every time, but it is because the series of events : two or more delegation certificates inserted within two adjacent epochs (X, X+1) AND checkpoint at (X+1) is so very much improbable. We should introduce this kind of event to state machine to properly test this cases. Then, `mReadWalletDelegations` should fail as it produces different result that Sqlite implementation

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1406,6 +1390,9 @@ joinStakePool ctx wid (pid, pools) argGenChange pwd = db & \DBLayer{..} -> do
     pure (tx, txMeta, txTime)
   where
     db = ctx ^. dbLayer @s @k
+    checkAlreadyJoined = \case
+        (WalletDelegation (Delegating pid') _) -> pid' == pid
+        _ -> False
```
paweljakubas: I am thinking if this is ok. Probably ok as first approximation. I just did not know how we want to treat situations like : we delegated to pId1, then can we quit as it is in the next? not active? Let's assume if yes, then what about situation joined to pId1, not active, then quit, then rejoined pId1. Etc. As I do not have this situations clear I opted for the simplest option here

paweljakubas:
bors try

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -578,22 +572,256 @@ newDBLayer trace defaultFieldValues mDatabaseFile = do
...
+                W.Delegating pool ->
+                    case (toWalletDelegationStatus dlg1, toWalletDelegationStatus dlg2) of
+                        (W.NotDelegating, W.NotDelegating) ->
+                            W.WalletDelegation (W.Delegating pool) []
+                        (W.Delegating pool1, W.NotDelegating) ->
+                            comparePools pool pool1 epoch1
+                        (W.NotDelegating, W.Delegating pool1) ->
+                            comparePools pool pool1 epoch2
+                        (W.Delegating pool1, W.Delegating pool2)
+                            | pool == pool1 && pool == pool2 ->
```
paweljakubas: so here, I just assumed that if we have situation that we delegating to poolId, and then in one epoch X there we a number of certificates with the latest joining to poolId, and in the X+1 epoch the same, then we do not want anything in next, as we are delegating to poolID and it is not going to be changed. This is what I feel should be presented (if we do not outlaw this kind of stuff which was already subject of another comment). Below are similar situations taken into consideration

paweljakubas:
bors try

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
```diff
@@ -390,55 +391,6 @@ spec = do
                         (`shouldBe` reward)
                 ]
 
-    it "STAKE_POOLS_JOIN_01 - I can join another stake-pool after previously \
```
paweljakubas, line 393: removed as it is duplicate of what is in : 
```
it "STAKE_POOLS_JOIN/QUIT - Checking delegation expectations"
```
and we are close to timeout in CI

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
```diff
@@ -512,30 +464,6 @@ spec = do
         let poolId = toText $ getApiT $ p ^. #id
         expectErrorMessage (errMsg403PoolAlreadyJoined poolId) r
 
-    it "STAKE_POOLS_JOIN_01 - \
```
paweljakubas, line 515: removed as it is duplicate of what is in : 
```
it "STAKE_POOLS_JOIN/QUIT - Checking delegation expectations"
```
and we are close to timeout in CI

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs
```diff
@@ -781,32 +709,6 @@ spec = do
             it ("Quit: " ++ title) $ \ctx ->
                 verifyIt ctx Link.quitStakePool headers expectations
 
-    it "STAKE_POOLS_QUIT_01 - Can quit stake pool" $ \ctx -> do
```
paweljakubas, line 784: removed as it is duplicate of what is in : 
```
it "STAKE_POOLS_JOIN/QUIT - Checking delegation expectations"
```
and we are close to timeout in CI

paweljakubas:
bors try

paweljakubas:
bors try

KtorZ:
bors try

KtorZ (approved):

PR #1340, opened 2020-02-12
Disable foreign key constraints during automatic database migration.

jonathanknowles:
# Issue Number

#1279 

# Overview

This PR:

- [x] Adds support for **_enabling_** and **_disabling_** foreign key constraints in SQLite databases.
- [x] Adds a bracket-style function to execute a task in a context where foreign key constraints are _**temporarily disabled**_.
- [x] Temporarily disables foreign key constraints while applying automatic updates generated by `persistent`.

This approach works because it prevents cascading deletes while automatic updates are being applied, _even though_ tables are dropped and recreated in the course of a migration.

# Testing

I have tested this informally (with a database created using the `v2019-12-14` version of `cardano-wallet` and I have also run the state migration tests as follows:

```hs
rm -rf state-migration-test-*
nix-build nix/migration-tests.nixnix -o migration-tests
./migration-tests/runall.sh
```

### DIFF
--- a/lib/core/src/Cardano/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/DB/Sqlite.hs
@@ -233,6 +233,19 @@ startSqliteBackend manualMigration migrateAll trace fp = do
             pure $ Left e
         Right _ -> pure $ Right ctx
 
+-- | Run the given task in a context where foreign key constraints are
+--   /temporarily disabled/, before re-enabling them.
+--
+withForeignKeysDisabled
+    :: Tracer IO DBLog
+    -> Sqlite.Connection
+    -> IO a
+    -> IO a
+withForeignKeysDisabled t c =
+    bracket_
+        (updateForeignKeysSetting t c ForeignKeysDisabled)
+        (updateForeignKeysSetting t c ForeignKeysEnabled)
+
 -- | Specifies whether or not foreign key constraints are enabled, equivalent
 --   to the Sqlite 'foreign_keys' setting.
 --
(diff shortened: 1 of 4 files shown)


lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -213,28 +211,105 @@ startSqliteBackend
...
         Left e -> do
             destroyDBLayer ctx
             pure $ Left e
-        Right ms -> do
-            pure $ Right (Quantity (length ms), ctx)
+        Right _ -> pure $ Right ctx
+
+-- | Run the given task in a context where foreign key constraints are
+--   /temporarily disabled/, before re-enabling them.
+--
```
KtorZ, line 240: Note that what's interesting here is not really the "what", but the "why". The comment is relatively redundant with the function's name, though I see you explained more about the "why" down below in the "ForeignKeySetting" documentation. Fair enough :+1:

lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -213,28 +211,105 @@ startSqliteBackend
...
+
+-- | Update the current value of the Sqlite 'foreign_keys' setting.
+--
+updateForeignKeysSetting
+    :: Tracer IO DBLog
+    -> Sqlite.Connection
+    -> ForeignKeysSetting
+    -> IO ()
+updateForeignKeysSetting trace connection desiredValue = do
+    traceWith trace $ MsgUpdatingForeignKeysSetting desiredValue
```
KtorZ, line 295: :+1:

KtorZ (approved):
Neat.

jonathanknowles:
@KtorZ wrote:
> Neat

Thanks!

PR #1343, opened 2020-02-13
Introduce SomeMnemonic as source of root keys (instead of entropy)

Anviking:
# Issue Number

#1316, preliminary work to unblock #1321 

# Overview

- Add SomeMnemonic as return type of fromMnemonic
- Make e.g. unsafeGenerateKeyFromSeed take a SomeMnemonic instead
entropy. This is more similar to `Icarus.generateKeyFromHardwareLedger`
- Add genMnemonic helper in a shared location

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE InstanceSigs #-}
 {-# LANGUAGE LambdaCase #-}
@@ -11,6 +12,7 @@
 {-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE RoleAnnotations #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
@@ -75,6 +77,7 @@ module Cardano.Wallet.Primitive.AddressDerivation
     , Passphrase(..)
     , PassphraseMinLength(..)
     , PassphraseMaxLength(..)
+    , SomeMnemonic(..)
     , FromMnemonic(..)
     , FromMnemonicError(..)
     , ErrWrongPassphrase(..)
@@ -92,11 +95,10 @@ import Cardano.Wallet.Primitive.Mnemonic
     , DictionaryError (..)
     , EntropyError (..)
     , EntropySize
+    , Mnemonic
     , MnemonicError (..)
     , MnemonicWordsError (..)
-    , entropyToBytes
     , mkMnemonic
-    , mnemonicToEntropy
     )
 import Cardano.Wallet.Primitive.Types
     ( Address (..), ChimericAccount (..), Hash (..) )
@@ -113,7 +115,7 @@ import Crypto.KDF.PBKDF2
 import Crypto.Random.Types
     ( MonadRandom (..) )
 import Data.Bifunctor
-    ( first )
+    ( bimap )
 import Data.ByteArray
     ( ByteArray, ByteArrayAccess, ScrubbedBytes )
 import Data.ByteArray.Encoding
@@ -146,6 +148,10 @@ import GHC.Generics
     ( Generic )
 import GHC.TypeLits
     ( KnownNat, Nat, Symbol, natVal )
+
+import Data.Type.Equality
+import Type.Reflection
+    ( typeOf )
 
 import qualified Basement.Compat.Base as B
 import qualified Data.ByteArray as BA
@@ -386,6 +392,16 @@ instance
 instance ToText (Passphrase purpose) where
     toText (Passphrase bytes) = T.decodeUtf8 $ BA.convert bytes
 
+data SomeMnemonic where
+    SomeMnemonic :: forall mw. KnownNat mw => Mnemonic mw -> SomeMnemonic
+
+deriving instance Show SomeMnemonic
+instance Eq SomeMnemonic where
+    (SomeMnemonic mwa) == (SomeMnemonic mwb) =
+        case typeOf mwa `testEquality` typeOf mwb of
+            Nothing -> False
+            Just Refl -> mwa == mwb
+
 -- | Create a passphrase from a mnemonic sentence. This class enables caller to
 -- parse text list of variable length into mnemonic sentences.
 --
@@ -394,8 +410,8 @@ instance ToText (Passphrase purpose) where
 --
 -- Note that the given 'Nat's **have** to be valid mnemonic sizes, otherwise the
 -- underlying code won't even compile, with not-so-friendly error messages.
-class FromMnemonic (sz :: [Nat]) (purpose :: Symbol) where
-    fromMnemonic :: [Text] -> Either (FromMnemonicError sz) (Passphrase purpose)
+class FromMnemonic (sz :: [Nat]) where
+    fromMnemonic :: [Text] -> Either (FromMnemonicError sz) SomeMnemonic
 
 -- | Error reported from trying to create a passphrase from a given mnemonic
 newtype FromMnemonicError (sz :: [Nat]) =
@@ -407,19 +423,19 @@ instance {-# OVERLAPS #-}
     ( n ~ EntropySize mw
     , csz ~ CheckSumBits n
     , ConsistentEntropy n mw csz
-    , FromMnemonic rest purpose
+    , FromMnemonic rest
     , NatVals rest
     ) =>
-    FromMnemonic (mw ': rest) purpose
+    FromMnemonic (mw ': rest)
   where
     fromMnemonic parts = case parseMW of
         Left err -> left (promote err) parseRest
         Right mw -> Right mw
       where
         parseMW = left (FromMnemonicError . getFromMnemonicError) $ -- coerce
-            fromMnemonic @'[mw] @purpose parts
+            fromMnemonic @'[mw] parts
         parseRest = left (FromMnemonicError . getFromMnemonicError) $ -- coerce
-            fromMnemonic @rest @purpose parts
+            fromMnemonic @rest parts
         promote e e' =
             let
                 sz = fromEnum <$> natVals (Proxy :: Proxy (mw ': rest))
@@ -447,11 +463,10 @@ instance
     , csz ~ CheckSumBits n
     , ConsistentEntropy n mw csz
     ) =>
-    FromMnemonic (mw ': '[]) purpose
+    FromMnemonic (mw ': '[])
   where
     fromMnemonic parts = do
-        m <- first (FromMnemonicError . pretty) (mkMnemonic @mw parts)
-        return $ Passphrase $ entropyToBytes $ mnemonicToEntropy m
+        bimap (FromMnemonicError . pretty) SomeMnemonic (mkMnemonic @mw parts)
       where
         pretty = \case
             ErrMnemonicWords ErrWrongNumberOfWords{} ->
--- a/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
@@ -138,6 +138,8 @@ import Data.Text.Class
     , fromTextToBoundedEnum
     , toTextFromBoundedEnum
     )
+import Data.Type.Equality
+    ( (:~:) (..), testEquality )
 import Data.Typeable
     ( Typeable )
 import Data.Word
@@ -148,8 +150,6 @@ import GHC.Generics
     ( Generic )
 import GHC.TypeLits
     ( KnownNat, Nat, Symbol, natVal )
-
-import Data.Type.Equality
 import Type.Reflection
     ( typeOf )
 
--- a/lib/core/test/bench/db/Main.hs
+++ b/lib/core/test/bench/db/Main.hs
@@ -55,7 +55,7 @@ import Cardano.Wallet.Primitive.AddressDerivation
     ( DelegationAddress (..)
     , Depth (..)
     , NetworkDiscriminant (..)
-    , Passphrase (..)
+    , SomeMnemonic (..)
     , WalletKey (..)
     , XPub
     , xpub
@@ -71,7 +71,7 @@ import Cardano.Wallet.Primitive.AddressDiscovery.Sequential
     , mkSeqState
     )
 import Cardano.Wallet.Primitive.Mnemonic
-    ( EntropySize, entropyToBytes, genEntropy )
+    ( EntropySize, entropyToMnemonic, genEntropy )
 import Cardano.Wallet.Primitive.Model
     ( Wallet, initWallet, unsafeInitWallet )
 import Cardano.Wallet.Primitive.Types
@@ -100,7 +100,7 @@ import Cardano.Wallet.Primitive.Types
     , fromFlatSlot
     )
 import Cardano.Wallet.Unsafe
-    ( unsafeRunExceptT )
+    ( unsafeMkSomeMnemonicFromEntropy, unsafeRunExceptT )
 import Control.DeepSeq
     ( NFData (..), force )
 import Control.Exception
@@ -122,6 +122,8 @@ import Data.ByteString
     ( ByteString )
 import Data.Functor
     ( ($>) )
+import Data.Proxy
+    ( Proxy (..) )
 import Data.Quantity
     ( Quantity (..) )
 import Data.Time.Clock.System
@@ -145,7 +147,6 @@ import System.IO.Unsafe
 import System.Random
     ( mkStdGen, randoms )
 
-import qualified Data.ByteArray as BA
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Char8 as B8
 import qualified Data.Map.Strict as Map
@@ -607,12 +608,15 @@ instance NFData SqliteContext where
 testCp :: WalletBench
 testCp = snd $ initWallet block0 genesisParameters initDummyState
 
+{-# NOINLINE initDummyState #-}
 initDummyState :: SeqState 'Testnet ShelleyKey
 initDummyState =
     mkSeqState (xprv, mempty) defaultAddressPoolGap
   where
-    bytes = entropyToBytes <$> unsafePerformIO $ genEntropy @(EntropySize 15)
-    xprv = generateKeyFromSeed (Passphrase bytes, mempty) mempty
+    mnemonic = unsafePerformIO
+        $ SomeMnemonic . entropyToMnemonic @15
+        <$> genEntropy @(EntropySize 15)
+    xprv = generateKeyFromSeed (mnemonic, Nothing) mempty
 
 testMetadata :: WalletMetadata
 testMetadata = WalletMetadata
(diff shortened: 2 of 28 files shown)


lib/core/test/unit/Cardano/WalletSpec.hs
```diff
@@ -510,13 +513,20 @@ instance Arbitrary (Passphrase purpose) where
     arbitrary =
         Passphrase . BA.convert . BS.pack <$> replicateM 16 arbitrary
 
+
+instance {-# OVERLAPS #-} Arbitrary SomeMnemonic where
+    arbitrary = do
+        bytes <- BS.pack <$> vectorOf 16 arbitrary
```
Anviking: To be removed

lib/core/test/unit/Cardano/Wallet/Primitive/TypesSpec.hs
```diff
@@ -215,8 +214,8 @@ spec = do
 
     describe "Buildable" $ do
         it "WalletId" $ do
-            let seed = Passphrase (BA.convert @ByteString "0000000000000000")
-            let xprv = generateKeyFromSeed (seed, mempty) mempty :: ShelleyKey 'RootK XPrv
+            let mw = either (error . show) id $ SomeMnemonic . entropyToMnemonic @12 <$> mkEntropy "0000000000000000"
```
Anviking: Line looks long

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/SequentialSpec.hs
```diff
@@ -652,3 +647,7 @@ data Key = forall (k :: Depth -> * -> *).
     , AddressPoolTest k
     ) => Key (Proxy k)
 instance Show Key where show (Key proxy) = show (typeRep proxy)
+
+dummyMnemonic :: SomeMnemonic
+dummyMnemonic = either (error . show) id $
+    SomeMnemonic . entropyToMnemonic @12 <$> mkEntropy (BS.pack $ replicate 16 0)
```
Anviking: Originally seeds of different lengths were used.

Also, matching the 12 with the `16` bytes has to be done manually. We could maybe define a helper:

```Haskell
mkEntropyFromByteStringOfSize @16 $ replicate 16 0
```

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/RandomSpec.hs
```diff
@@ -188,16 +189,16 @@ goldenSpecTestnet =
...
     , accIndex :: Word32
     , addrIndex :: Word32
     , expected :: Bool
     } deriving (Show, Eq)
 
 -- An arbitrary mnemonic sentence for the tests
-arbitraryMnemonic :: [Text]
-arbitraryMnemonic =
+arbitraryMnemonic :: SomeMnemonic
+arbitraryMnemonic = either (error . show) id $ fromMnemonic @'[12]
```
Anviking, line 201: Having to explicitly tell (`@'[12]`) which mnemonic lengths are supported, when we are hard-coding the mnemonic, could be considered cumbersome.

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivation/IcarusSpec.hs
```diff
@@ -69,24 +70,24 @@ import qualified Data.Text as T
...
-            seed0 (toEnum 0x80000000) UTxOExternal (toEnum 0x0000000E)
-            "Ae2tdPwUPEZDLWQQEBR1UW7HeXJVaqUnuw8DUFu52TDWCJbxbkCyQYyxckP"
-
-        goldenAddressGeneration $ GoldenAddressGeneration
-            seed0 (toEnum 0x8000000E) UTxOInternal (toEnum 0x0000002A)
-            "Ae2tdPwUPEZFRbyhz3cpfC2CumGzNkFBN2L42rcUc2yjQpEkxDbkPodpMAi"
-
+--        let seed0 = Passphrase "4\175\242L\184\243\191 \169]\171 \207\r\v\233\NUL~&\ETB"
+--
+--        goldenAddressGeneration $ GoldenAddressGeneration
```
Anviking: We cannot use the seed (which is really entropy) anymore.

We could
1. Remove the golden and not replace it
2. Replace it with another golden (from a mnemonic)

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivation/ByronSpec.hs
```diff
@@ -97,11 +98,11 @@ data GenerateKeyFromSeed = GenerateKeyFromSeed
 
 generateTest :: GenerateKeyFromSeed -> Expectation
 generateTest GenerateKeyFromSeed{..} =
-    getKey (generateKeyFromSeed (Passphrase seed) pwd)
+    getKey (generateKeyFromSeed mw pwd)
     `shouldBe`
     getKey rootKey
   where
-    Right (Passphrase seed) = fromMnemonic @'[12] mnem
+    Right mw = fromMnemonic @'[12] mnem
```
Anviking: `either (error . show) id` would provide better failure reasons if this were to ever fail.

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
```diff
@@ -646,8 +647,8 @@ testCpSeq = snd $ initWallet block0 genesisParameters initDummyStateSeq
 initDummyStateSeq :: SeqState 'Testnet ShelleyKey
 initDummyStateSeq = mkSeqState (xprv, mempty) defaultAddressPoolGap
   where
-      bytes = entropyToBytes <$> unsafePerformIO $ genEntropy @(EntropySize 15)
-      xprv = Seq.generateKeyFromSeed (Passphrase bytes, mempty) mempty
+      mw = SomeMnemonic $ unsafePerformIO (generate $ genMnemonic @15)
```
Anviking, line 651: Ok to call a `SomeMnemonic` `mw` as in "mnemonic words"?

We have used the `mw` name in the past, I believe.

Anviking:
bors try

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivation/IcarusSpec.hs
Anviking: I'm just removing it.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -147,6 +149,10 @@ import GHC.Generics
 import GHC.TypeLits
     ( KnownNat, Nat, Symbol, natVal )
 
+import Data.Type.Equality
+import Type.Reflection
+    ( typeOf )
```
KtorZ, line 154: Why are these imports are separated from the rest? 
Also, `Data.Type.Equality` is importing stuff in context implicitly, let's make the import fully explicit for readability.

lib/core/test/bench/db/Main.hs
```diff
@@ -607,12 +606,16 @@ instance NFData SqliteContext where
...
 initDummyState :: SeqState 'Testnet ShelleyKey
 initDummyState =
     mkSeqState (xprv, mempty) defaultAddressPoolGap
   where
-    bytes = entropyToBytes <$> unsafePerformIO $ genEntropy @(EntropySize 15)
-    xprv = generateKeyFromSeed (Passphrase bytes, mempty) mempty
+    mnemonic = unsafePerformIO $
+        SomeMnemonic
+        . entropyToMnemonic @15
+        <$> genEntropy @(EntropySize 15)
```
KtorZ, line 617: Minor style suggestion: 

This is a relatively weird choice of indentation. It doesn't seem to improve much readability. The composition is at the same level as the applicative, whereas it's really the composition that is applied to value inside the IO from `genEntropy`. So, what about grouping things as:

```hs
mnemonic = unsafePerformIO 
    $ SomeMnemonic . entropyToMnemonic @15 
   <$> genEntropy @(EntropySize 15)
```

lib/core/test/bench/db/Main.hs
```diff
@@ -629,12 +632,16 @@ testPk :: PrimaryKey WalletId
...
+    seed = SomeMnemonic $ entropyToMnemonic @15 ent
+    ent = either (error . show) id $ mkEntropy $ BS.replicate 32 0
 
 rewardAccount :: ShelleyKey 'AddressK XPub
-rewardAccount = publicKey $ unsafeGenerateKeyFromSeed (seed, mempty) mempty
-  where seed = Passphrase $ BA.convert $ BS.replicate 32 0
+rewardAccount = publicKey $ unsafeGenerateKeyFromSeed (seed, Nothing) mempty
+  where
+    seed = SomeMnemonic $ entropyToMnemonic @15 ent
+    ent = either (error . show) id $ mkEntropy $ BS.replicate 32 0
```
KtorZ: I'd suggest to move this to `Cardano.Wallet.Unsafe` under `unsafeSomeMnemonic`  with a type parameter to apply.

lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs
```diff
@@ -464,25 +467,25 @@ rootKeysSeq = unsafePerformIO $ generate (vectorOf 10 genRootKeysSeq)
     genRootKeysSeq :: Gen (ShelleyKey 'RootK XPrv)
     genRootKeysSeq = do
         (s, g, e) <- (,,)
-            <$> genPassphrase @"seed" (16, 32)
-            <*> genPassphrase @"generation" (0, 16)
+            <$> (SomeMnemonic <$> genMnemonic @12)
+            <*> (SomeMnemonic <$> genMnemonic @12)
             <*> genPassphrase @"encryption" (0, 16)
-        return $ Seq.generateKeyFromSeed (s, g) e
+        return $ Seq.generateKeyFromSeed (s, Just g) e
```
KtorZ: Instead of forcing all second passphrases to be `Just`, it'd be better to `fmap SomeMnemonic` in the generator 2 lines above and actually generate a `Maybe SomeMnemonic`. This way, the generator would also cover cases where no second factor is provided, like before..

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivation/IcarusSpec.hs
```diff
@@ -69,24 +70,10 @@ import qualified Data.Text as T
...
-            seed0 (toEnum 0x80000000) UTxOExternal (toEnum 0x00000000)
-            "Ae2tdPwUPEZGQVrA6qKreDzdtYxcWMMrpTFYCpFcuJfhJBEfoeiuW4MtaXZ"
-
-        goldenAddressGeneration $ GoldenAddressGeneration
-            seed0 (toEnum 0x80000000) UTxOExternal (toEnum 0x0000000E)
-            "Ae2tdPwUPEZDLWQQEBR1UW7HeXJVaqUnuw8DUFu52TDWCJbxbkCyQYyxckP"
-
-        goldenAddressGeneration $ GoldenAddressGeneration
-            seed0 (toEnum 0x8000000E) UTxOInternal (toEnum 0x0000002A)
-            "Ae2tdPwUPEZFRbyhz3cpfC2CumGzNkFBN2L42rcUc2yjQpEkxDbkPodpMAi"
```
KtorZ, line 84: Hmmmm... Why ? We can't just get rid of golden tests like this :|

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/RandomSpec.hs
KtorZ: Alternatively, use: `unsafeMkMnemonic` from `Cardano.Wallet.Unsafe`

lib/jormungandr/test/unit/Cardano/Wallet/Jormungandr/TransactionSpec.hs
```diff
@@ -529,23 +531,27 @@ goldenTestDelegationCertTx tl keystore pool (accountXPrv, pass) inps outs bytes'
...
-    ( Rnd.unsafeGenerateKeyFromSeed derPath (Passphrase $ BA.convert seed) pwd
+xprvRndFromSeed bytes =
+    ( Rnd.unsafeGenerateKeyFromSeed derPath seed pwd
     , pwd
     )
   where
     pwd = mempty
     derPath = (minBound, minBound)
+    seed = SomeMnemonic $ entropyToMnemonic ent
+    ent = either (error . show) id $ mkEntropy @(EntropySize 12) bytes
```
KtorZ: I've seen this probably 10 times in this PR. I think that's a good case for a `unsafeMkEntropy` with a proper call stack to be defined in our unsafe module.

KtorZ (commented):
Looks good overall, but some minor tweaks to go through before merging :+1:

lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs
Anviking: > fmap SomeMnemonic

There is no arbitrary instance for `Mnemonic`, but I added a `frequency` combinator.

🤔 Could probably have added an `Arbitrary` instance… But I think this is fine too.

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivation/IcarusSpec.hs
Anviking: Re-added. Originally I didn't see how we'd go from entropy to `SomeMnemonic`.

lib/core/test/bench/db/Main.hs
Anviking: I added `unsafeSomeMnemonicFromEntropy`.

Anviking: The export list now looks like:
-    ( unsafeFromHex
-    , unsafeFromHexFile
-    , unsafeDecodeAddress
-    , unsafeDecodeHex
-    , unsafeFromText
-    , unsafeRunExceptT
-    , unsafeXPrv
-    , unsafe**Mk**Mnemonic
-    , unsafe**Mk**Entropy
-    , unsafeSomeMnemonicFromEntropy
-    , unsafeDeserialiseCbor
-    , unsafeBech32DecodeFile
-    , unsafeBech32Decode

I guess `unsafeSomeMnemonicFromEntropy` should be `unsafeMkSomeMnemonicFromEntropy`.

Anviking: Done.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
Anviking: Done.

lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs
```diff
@@ -463,26 +467,28 @@ rootKeysSeq = unsafePerformIO $ generate (vectorOf 10 genRootKeysSeq)
   where
     genRootKeysSeq :: Gen (ShelleyKey 'RootK XPrv)
     genRootKeysSeq = do
-        (s, g, e) <- (,,)
-            <$> genPassphrase @"seed" (16, 32)
-            <*> genPassphrase @"generation" (0, 16)
-            <*> genPassphrase @"encryption" (0, 16)
+        s <- (SomeMnemonic <$> genMnemonic @12)
```
Anviking:
```suggestion
        s <- SomeMnemonic <$> genMnemonic @12
```

lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs
```diff
@@ -463,26 +467,28 @@ rootKeysSeq = unsafePerformIO $ generate (vectorOf 10 genRootKeysSeq)
...
+    mw = SomeMnemonic $ unsafePerformIO $ generate $ genMnemonic @15
 
 arbitraryRewardAccount
     :: ShelleyKey 'AddressK XPub
 arbitraryRewardAccount =
-    publicKey $ unsafeGenerateKeyFromSeed (seed, mempty) mempty
+    publicKey $ unsafeGenerateKeyFromSeed (mw, Nothing) mempty
   where
-    seed = Passphrase $ BA.convert $ BS.replicate 32 0
+    mw = SomeMnemonic $ unsafePerformIO $ generate $ genMnemonic @15
```
Anviking: Should use `unsafeMkSomeMnemonicFromEntropy` here now, to keep the old behaviour.

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivation/ByronSpec.hs
```diff
@@ -97,11 +98,11 @@ data GenerateKeyFromSeed = GenerateKeyFromSeed
 
 generateTest :: GenerateKeyFromSeed -> Expectation
 generateTest GenerateKeyFromSeed{..} =
-    getKey (generateKeyFromSeed (Passphrase seed) pwd)
+    getKey (generateKeyFromSeed mw pwd)
     `shouldBe`
     getKey rootKey
   where
-    Right (Passphrase seed) = fromMnemonic @'[12] mnem
+    mw = either (error . show) id $ fromMnemonic @'[12] mnem
```
Anviking: Could use `unsafeMkMnemonic` here, I believe.

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivationSpec.hs
```diff
@@ -413,3 +416,6 @@ genLegacyAddress range = do
     addrPayload <- BS.pack <$> vectorOf n arbitrary
     let crc = BS.pack [26,1,2,3,4]
     return $ Address (prefix <> addrPayload <> crc)
+
+instance {-# OVERLAPS #-} Arbitrary SomeMnemonic where
```
Anviking:
```suggestion
instance Arbitrary SomeMnemonic where
```

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/SequentialSpec.hs
```diff
@@ -346,8 +346,10 @@ prop_genChangeGap
 prop_genChangeGap g =
     property prop
   where
-    seed = Passphrase (BA.convert @ByteString "0000000000000000")
-    key = Shelley.unsafeGenerateKeyFromSeed (seed, mempty) mempty
+    mw = SomeMnemonic
+        $ entropyToMnemonic @12
+        $ unsafeMkEntropy "0000000000000000"
```
Anviking: `unsafeMkSomeMnemonicFromEntropy`

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/SequentialSpec.hs
```diff
@@ -652,3 +647,7 @@ data Key = forall (k :: Depth -> * -> *).
     , AddressPoolTest k
     ) => Key (Proxy k)
 instance Show Key where show (Key proxy) = show (typeRep proxy)
+
+dummyMnemonic :: SomeMnemonic
+dummyMnemonic =
+    unsafeMkSomeMnemonicFromEntropy (Proxy @12) (BS.pack $ replicate 16 0)
```
Anviking: Replacing both 16- and 32- bytes long entropies in this module with 16, seems iffy.

Adding a `dummyMnemonic :: (many constraints) => Proxy mw -> SomeMnemonic` sounds tempting…

But in-doubt, I will not, I think.

Will remove `dummyMnemonic`.

KtorZ (approved):

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivation/IcarusSpec.hs
KtorZ: Yeah, the mnemonic is basically the entropy + checksum encoded as words. The checksum can be computed from the entropy itself, so entropy and mnemonic are somewhat equivalent,

KtorZ:
(IMO, this is a good case where the last commits don't bring much to the story the PR is telling. Because they're simply correcting stuff introduced earlier in the PR itself. This is a good case for squashing :wink:)

Anviking:
Argh! 😄 

bors r+

Anviking:
> Because they're simply correcting stuff introduced earlier in the PR itself.

But maybe the indirect path from A to B in the PR was interesting… which the author of the PR might be biased when judging 😉

KtorZ:
> maybe the indirect path from A to B in the PR was interesting

Was it ? ( :trollface: )

Issue #1347, opened 2020-02-14
Add tx submission to the byron API


KtorZ:
# Context

We didn't initially add this for Jörmungandr because it is not possible to make transaction to legacy addresses in Jörmungandr (which makes it fairly impractical to construct transactions from legacy wallets which have change addresses that are necessarily legacy addresses). 

Yet, for the Byron integration, we need wallets to support transaction submission in the very same way. So, it'd be nice to get closer to API parity between byron and shelley wallets.

# Decision

- Allow POSTing transaction to `/byron-wallets/.../transactions` similarly to 'standard' wallets.
- On Jörmungandr, this is known to generate invalid transaction.

# Acceptance Criteria

- The API _must_ support creating payment transaction from byron wallets.

---

# Development 

- Implement the API endpoint ==> #1350 

# QA 

- Added integration setup for cardano-node, this does not run in CI _yet_ :
    https://github.com/input-output-hk/cardano-wallet/blob/master/lib/byron/test/integration/Main.hs#L111-L115

- Integration tests cover byron wallet scenarios, byron transaction scenarios, network scenarios as well as transaction scenarios ported from the core-integration scenarios but adjusted to work on the newly introduced transactions endpoint in the Byron API.
   - see: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/byron/test/integration/Test/Integration/Byron/Scenario/API/Transactions.hs#L99-L145

piotr-iohk:
lgtm.

It would be nice then to employ those tests in CI as soon as possible.

Issue #1346, opened 2020-02-14
Complete Local Tx Submission integration


KtorZ:
# Context

When doing the first bits of integration with the Haskell nodes, we left a placeholder for transaction submission: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/byron/src/Cardano/Wallet/Byron/Network.hs#L559

# Decision

- Complete the integration with the local tx submission protocol
- Test this on a private blockchain, if possible.

# Acceptance Criteria

- The network interface _must_ support sending transaction via the local tx submission
- `cardano-wallet` _must_ be integrated with the most recent release of the cardano-node's stack 

---

# Development 

- Integrate with the tx local submission and completed the postTx handler of the network layer ==> #1348 
- Upgrade cardano-node's dependencies to latest release (1.5.0) ==> #1347  #1358 
- See if we can connect the wallet to a private Haskell blockchain ==> #1414 

# QA 

See #1347

piotr-iohk:
lgtm

PR #1348, opened 2020-02-14
LocalTxSubmission real implementation (remove 'localTxSubmissionNull' placeholder)

KtorZ:
# Issue Number

#1346 

# Overview

- 9b89a053bb4f3e40a0f6909d89e3bc594c50579d
  Allow mapping on 'NextBlocksResult's cursor
  This makes the type slightly more flexible and avoid pushing down too
much unrelated context regarding cursors. See Byron.Network adjustments
in the next commit

- c9fba67e393d9f52bc113f4be554dca5534fe3f5
  define compatibility type translation between sealedtx and byron gentx
  
- ba8c30d48e9d982b1991d75e76f8529e287e3ad5
  rename 'NetworkClientCmd' to 'ChainSyncCmd'
  - And move 'Cursor' definition out of the chain sync client
- Also renamed 'queue' to 'chainSyncQ' to make the distinction with
  the upcoming local tx submission queue clearer

- 93865ae0b4f7465e943ade1429c7d00e29be11f0
  wire up 'LocalTxSubmission' protocol in Byron Network

# Comments

Testing will come soon, I first need to enable the transaction submission through the API (see #1347) and start a local blockchain.

### DIFF
--- a/lib/byron/src/Cardano/Wallet/Byron/Network.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron/Network.hs
@@ -40,6 +40,7 @@ import Cardano.Wallet.Byron.Compatibility
     , genesisTip
     , toByronHash
     , toEpochSlots
+    , toGenTx
     , toPoint
     )
 import Cardano.Wallet.Logging
@@ -48,6 +49,7 @@ import Cardano.Wallet.Network
     ( Cursor
     , ErrGetBlock (..)
     , ErrNetworkUnavailable (..)
+    , ErrPostTx (..)
     , NetworkLayer (..)
     , NextBlocksResult (..)
     , mapCursor
@@ -82,7 +84,7 @@ import Control.Monad.Class.MonadTimer
 import Control.Monad.IO.Class
     ( MonadIO )
 import Control.Monad.Trans.Except
-    ( ExceptT (..), withExceptT )
+    ( ExceptT (..), throwE, withExceptT )
 import Control.Tracer
     ( Tracer, contramap )
 import Data.ByteString.Lazy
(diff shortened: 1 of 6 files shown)


lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -274,15 +284,21 @@ newNetworkLayer tr bp addrInfo versionData = NetworkLayer
...
-        (NextBlocksResult (m Byron) ByronBlock -> m ())
+        (NextBlocksResult (Point ByronBlock) ByronBlock -> m ())
     | CmdCurrentNodeTip
         (Tip ByronBlock -> m ())
 
+-- | Sending command to the localTxSubmission client. See also 'ChainSyncCmd'.
+data LocalTxSubmissionCmd (m :: * -> *)
+    = CmdSubmitTx
+        (GenTx ByronBlock)
+        (Maybe String -> m ())
```
Anviking, line 300: `GenTx ByronBlock` is basically a `Tx` corresponding to the `ByronBlock`?

`Maybe String` -> is `Nothing` success and `Just str` an error?

Anviking (approved):
lgtm!

lib/byron/src/Cardano/Wallet/Byron/Network.hs
KtorZ: `Gen` stands for _Generalized_  as far as I understood, and yes, it corresponds to classic transactions from byron blocks although, there are several constructors for a `GenTx`, some of them being used to construct delegation transaction for instance. 

And yes, the way the `LocalTxSubmission` protocol works is by returning a `Maybe err` as a message, where `err` is currently instantiated as a `String`.

PR #1350, opened 2020-02-14
Extend byron API to support making transactions from byron wallets 

KtorZ:
# Issue Number

#1347 

# Overview

- e04b67cf47ec471231497ae3b8864962cc0e6ba6
  re-generate nix machinery
  
- f554ea41c49ecc58922682e6280f51a85039aa70
  extend swagger API specifications byron transactions to support payment
  
- bbe3d0303b3b2fb7c7aeff8bf7dc27ac94f17b33
  implement postTransaction (& fee) for byron endpoints
  
- ae8f84ccbaf96a25e99ef7da8b7e87395598649e
  review fixture functions to also return mnemonic
  And add two helper function to derive addresses from mnemonic sentences

- 3b3045cfeee7ab956cadc03d61efab74f0c3a80b
  extend Link.createTransaction and Link.postTransactionFee for Byron style
  
- 2ec53ee4c596953e73f07bb439f7206e1c099245
  implement basic integration scenario for byron txs
  
- 8a301bed9052eb3052e6cffa3ea6a46e3701217a
  implement protocol magic prefix in Byron transaction signing
  
- 36773806e34412b00d4148212de71f44452b0401
  Fix missing header bytes on byron transactions
  
- e6924c5f0d68abf160cdfa86b1e48ba10d02261e
  implement proper fee estimator for Byron
  
- 45e9927064b8e681d9b8100f1d6ee0e83503b123
  Cover multi-output transactions in cardano-node Byron scenarios
  
- e7511cebce9e050cbe488dc8531b891c3a3ab3b0
  move network discriminant choice up to integration/Main
  
- f1cb42293cbd47077691a268d0254eb8b0bbff55
  add remaining TRANS_CREATE and TRANS_ESTIMATE scenarios for byron w/ cardano-node
  
- dd8859eb6f2398b5a98eb4ac4e8cc570237cebb2
  re-generate nix machinery
  
- 3f4be72ca480a1312672610b9d919416554780a8
  log network magic with the network name
  
- ca1d3348bd862e0c4d8d9833e10ff89589a1d2f0
  fixup missing updates to create and fees endpoints in Latency benchmarks

# Comments

- I am starting to really believe that the `server` declaration should live in the target-specific packages (resp. cardano-wallet-jormungandr and cardano-wallet-byron) since the shape of the API is highly depending on them. Will do in another PR

- This closes the loop of the necessary things needed in the app for testing, now I need to investigate how to setup integration tests for the byron rewritten nodes to test that out!

### DIFF
--- a/lib/byron/src/Cardano/Wallet/Byron.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron.hs
@@ -78,6 +78,7 @@ import Cardano.Wallet.Primitive.AddressDerivation
     ( Depth (..)
     , NetworkDiscriminant (..)
     , NetworkDiscriminantVal
+    , PaymentAddress
     , PersistPrivateKey
     , WalletKey
     , networkDiscriminantVal
@@ -150,6 +151,7 @@ import qualified Network.Wai.Handler.Warp as Warp
 serveWallet
     :: forall (n :: NetworkDiscriminant) t.
         ( NetworkDiscriminantVal n
+        , PaymentAddress n IcarusKey
         , DecodeAddress n
         , EncodeAddress n
         , WorstSizeOf Address n IcarusKey
--- a/lib/byron/src/Cardano/Wallet/Byron.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron.hs
@@ -67,7 +67,7 @@ import Cardano.Wallet.Byron.Network
 import Cardano.Wallet.Byron.Transaction
     ( newTransactionLayer )
 import Cardano.Wallet.Byron.Transaction.Size
-    ( WorstSizeOf )
+    ( MaxSizeOf )
 import Cardano.Wallet.DB.Sqlite
     ( DatabasesStartupLog, DefaultFieldValues (..), PersistState )
 import Cardano.Wallet.Logging
@@ -154,8 +154,8 @@ serveWallet
         , PaymentAddress n IcarusKey
         , DecodeAddress n
         , EncodeAddress n
-        , WorstSizeOf Address n IcarusKey
-        , WorstSizeOf Address n ByronKey
+        , MaxSizeOf Address n IcarusKey
+        , MaxSizeOf Address n ByronKey
         , t ~ IO Byron
         )
     => Tracers IO
--- a/lib/byron/src/Cardano/Wallet/Byron.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron.hs
@@ -61,7 +61,7 @@ import Cardano.Wallet.Api.Server
 import Cardano.Wallet.Api.Types
     ( DecodeAddress, EncodeAddress )
 import Cardano.Wallet.Byron.Compatibility
-    ( Byron, ByronBlock, fromByronBlock )
+    ( Byron, ByronBlock, fromByronBlock, fromNetworkMagic )
 import Cardano.Wallet.Byron.Network
     ( AddrInfo, withNetworkLayer )
 import Cardano.Wallet.Byron.Transaction
@@ -136,7 +136,7 @@ import Network.Wai.Handler.Warp
 import Network.Wai.Middleware.Logging
     ( ApiLog )
 import Ouroboros.Network.NodeToClient
-    ( NodeToClientVersionData )
+    ( NodeToClientVersionData (..) )
 import System.Exit
     ( ExitCode (..) )
 
@@ -202,9 +202,10 @@ serveWallet
     serveApp socket = do
         withNetworkLayer nullTracer bp addrInfo versionData $ \nl -> do
             withNtpClient ntpClientTracer ntpSettings $ \ntpClient -> do
-                byronApi   <- apiLayer (newTransactionLayer @n) nl
-                icarusApi  <- apiLayer (newTransactionLayer @n) nl
-                startServer socket byronApi icarusApi ntpClient $> ExitSuccess
+                let pm = fromNetworkMagic $ networkMagic $ fst versionData
+                randomApi  <- apiLayer (newTransactionLayer @n pm) nl
+                icarusApi  <- apiLayer (newTransactionLayer @n pm) nl
+                startServer socket randomApi icarusApi ntpClient $> ExitSuccess
 
     startServer
         :: Socket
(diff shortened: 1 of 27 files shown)


lib/byron/src/Cardano/Wallet/Byron.hs
```diff
@@ -200,22 +202,23 @@ serveWallet
     serveApp socket = do
         withNetworkLayer nullTracer bp addrInfo versionData $ \nl -> do
             withNtpClient ntpClientTracer ntpSettings $ \ntpClient -> do
-                byronApi   <- apiLayer (newTransactionLayer @n) nl
-                icarusApi  <- apiLayer (newTransactionLayer @n) nl
-                startServer socket byronApi icarusApi ntpClient $> ExitSuccess
+                let pm = fromNetworkMagic $ networkMagic $ fst versionData
```
paweljakubas, line 208: maybe it would be worth to log some info here, `versionData` etc

lib/byron/src/Cardano/Wallet/Byron.hs
```diff
@@ -200,22 +202,23 @@ serveWallet
...
+                startServer socket randomApi icarusApi ntpClient $> ExitSuccess
 
     startServer
         :: Socket
         -> ApiLayer (RndState 'Mainnet) t ByronKey
         -> ApiLayer (SeqState 'Mainnet IcarusKey) t IcarusKey
         -> NtpClient
         -> IO ()
-    startServer socket byron icarus ntp = do
+    startServer socket random icarus ntp = do
```
paweljakubas, line 219: I understand (or I hope to understand) the change here - we have Byron era, and in Byron era a number of things can be supported random, icarus... But is not icarus using random scheme? Any chance to put somewhere documentation about differences between random and icarus and how they fit within Byron - or maybe link to the source describing this?

lib/byron/test/integration/Test/Integration/Byron/Scenario/API/Transactions.hs
```diff
@@ -0,0 +1,145 @@
...
+        , Discriminate style
+        , HasType (ApiT WalletId) wallet
+        , FromJSON wallet
+        , Generic wallet
+        , Show wallet
+        )
+    => (Context t -> IO (wallet, Mnemonic mw))
+    -> (Mnemonic mw -> [Address])
+    -> SpecWith (Context t)
+scenario_TRANS_CREATE_01 fixtureWallet fixtureAddresses =
```
paweljakubas: *[question]*

Can I send from random wallet to icarus wallet and vice versa and expect it to work?

lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -834,6 +849,56 @@ delegationFee
...
+--
+-- To be typically used as:
+--
+-- >>> take 1 (randomAddresses @n)
+-- [addr]
+randomAddresses
+    :: forall (n :: NetworkDiscriminant). (PaymentAddress n ByronKey)
+    => Mnemonic 12
+    -> [Address]
+randomAddresses mw =
```
paweljakubas, line 1004: I am wondering if we should not try to unify *random* and *icarus* functions under one hood. But maybe not, having them standalone maybe incur additional LOC, but it is more clear how they are constructed and hence they are much clearer to understand (for me at least)

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -401,6 +412,13 @@ type ListByronTransactions n = "byron-wallets"
     :> QueryParam "order" (ApiT SortOrder)
     :> Get '[JSON] [ApiTransaction n]
 
+-- | https://input-output-hk.github.io/cardano-wallet/api/#operation/postByronTransactionFee
+type PostByronTransactionFee n = "byron-wallets"
+    :> Capture "walletId" (ApiT WalletId)
+    :> "payment-fees"
```
paweljakubas, line 418: cannot be `txs-fees`, `transactions-fees` or just `fees`?

paweljakubas (approved):
looks fine! (modulo swagger and nix stuff)

lib/byron/src/Cardano/Wallet/Byron.hs
KtorZ: > we have Byron era, and in Byron era a number of things can be supported random, icarus...

Correct :+1: 

> .. But is not icarus using random scheme? 

Incorrect :sweat_smile: 

> Any chance to put somewhere documentation about differences between random and icarus and how they fit within Byron - or maybe link to the source describing this?

Here: https://github.com/input-output-hk/cardano-wallet/wiki/About-Address-Derivation#differences-between-hd-random-and-hd-sequential-wallets

and here: https://github.com/input-output-hk/cardano-wallet/wiki/Wallet-Cryptography-and-Encoding#master-key-generation

KtorZ: (not that I renamed this to indeed avoid confusion between the Byron era which so to speak includes Icarus wallets and Random wallets which are anyway part of the Shelley era  too :upside_down_face: )

lib/byron/test/integration/Test/Integration/Byron/Scenario/API/Transactions.hs
KtorZ: Yes! Note that I have only ported a single scenario here, but I'll add a few more this morning, basically, most scenarios we already have in `API/Transactions`. While doing so, I am abstracting away the bits specific to wallets so that once done, we can simply use these functions directly in both cases (Jormungandr & cardano-node) without duplicating any code.

lib/core-integration/src/Test/Integration/Framework/DSL.hs
KtorZ: > I am wondering if we should not try to unify random and icarus functions under one hood. 

I am wondering the same to be honest. So far, I haven't really found a good justification for doing so however, so I am refraining myself ^.^

lib/byron/src/Cardano/Wallet/Byron.hs
KtorZ: Done.

lib/core/src/Cardano/Wallet/Api.hs
KtorZ: This is to distinguish between delegation-fees and payment-fees. Both are transactions, but they are different kind of transactions.

PR #1265, opened 2020-01-15
Update nixpkgs channel to 19.09

angerman:
# Overview

This updates the nixpkgs version to 19.09 (latest stable release) from 19.03 (unsupported old release).

# Details

haskell.nix (as referenced in this repository) for the purpose of running CI, was using a custom nixpkgs-19.03 branch. It has been the goal for a while to upgrade from nixpkgs-19.03 to nixpkgs-19.09. Ideally to the stock upstream one without custom patches. This PR updates the nixpkgs-19.03 pin to nixpkgs-19.09 and fixes the issues (as encountered below) in haskell.nix. Subsequently this PR also contains a updated haskell.nix pin. The `w64` issue stems from nixpkgs changing the identifier for windows cross from `x86_64-pc-mingw32` to `x86_64-w64-mingw32` from nixpkgs 19.03 to 19.09. This interacts badly with the jormungandr reference for windows, and @rvl will take a look at that within cardano-wallet, as it can not be fixed in haskell.nix.

# Comments

Issues encountered:
- [x] ```error: attribute 'series3' missing, at /nix/store/87zygi9fxvgqkdflsd5hmx947aj3mf1m-nixpkgs/pkgs/misc/emulators/wine/packages.nix:6:11```
     This is due to missing https://github.com/NixOS/nixpkgs/pull/71216

- [x] ```Unknown vendor: w64```
   This seems to be due to jourmungandr using some 18.09git checkout, which doesn't sport the `w64` vendor, but uses `pc` instead.

- [x] ```/nix/store/d7q2kp5saj7xd9wii7cyfis8y3s4d2sj-network-2.6.3.6-setup-x86_64-w64-mingw32/bin/Setup: error while loading shared libraries: libgmp.so.10: cannot open shared object file: No such file or directory```
   Which is due to
   ```
   ldd /nix/store/d7q2kp5saj7xd9wii7cyfis8y3s4d2sj-network-2.6.3.6-setup-x86_64-w64-mingw32/bin/Setup
           linux-vdso.so.1 (0x00007ffeb9369000)
           libm.so.6 => /nix/store/pnd2kl27sag76h23wa5kl95a76n3k9i3-glibc-2.27/lib/libm.so.6 (0x00007f7a4695c000)
           librt.so.1 => /nix/store/pnd2kl27sag76h23wa5kl95a76n3k9i3-glibc-2.27/lib/librt.so.1 (0x00007f7a46952000)
[...]

PR #1351, opened 2020-02-14
use iohk-nix and haskell.nix as overlays

disassembler:
# Issue Number

# Overview

- [ ] I have ...

# Comments


nix/haskell.nix
```diff
@@ -0,0 +1,149 @@
...
+          "ghc" "array" "binary" "bytestring" "containers"
+          "filepath" "ghc-boot" "ghc-compact" "ghc-prim"
+          # "ghci" "haskeline"
+          "hpc"
+          "mtl" "parsec" "text" "transformers"
+          "xhtml"
+          # "stm" "terminfo"
+        ];
+        #reinstallableLibGhc = true;
+      }
```
angerman: This will result in `configFiles` not found. Primarily because for windows we'll eventually end up depending on Win32, and the stackage snapshot does not contain the `Win32` package.

angerman (commented):
There seems to be some old nixpkgs being pulled in. One where we need to use `x86_64-pc-mingw32`. I'm not yet sure where that comes from :-/

hamishmack:
@disassembler. It looks like the [required](https://hydra.iohk.io/job/Cardano/cardano-wallet-pr-1351/required) jobs are working now.  There are still a number of failures, but I am not sure which of these are important.  All the eval time issues have been resolved.

The fixes generally involved:

- Switching from `pkgs` to `pkgs.buildPackages` to fix cross compilation.
- Disabling jobs expected to fail (cross compilation or platform issues).

Using `packagePlatformsOrig` means that it is now possible to disable jobs with `meta.platforms`,  but keep in mind that we must set `meta.platforms` to something like `lib.platforms.all` to get macOS builds (haskell.nix derivations set this by default, but `mkDerivation` does not).

Let us know if there is anything else we need to investigate.

KtorZ (approved):

rvl:
@disassembler @angerman The stack build in CI is failing for the same reason as in #1265.

disassembler:
> @disassembler @angerman The stack build in CI is failing for the same reason as in #1265.

I think it's a matter of getting the cache populated on all the CI machines to handle the mass rebuild as CI is passing in the PR, just not the bors job.

craigem (approved):
bors r+

disassembler:
ah, bors does a FullTest, everything else does a quick test. That explains why it only fails with bors.

rvl:
@disassembler Would you be able to update the [wiki docs for pinning sources](https://github.com/input-output-hk/cardano-wallet/wiki/Building#iohk-nix-pin)?

PR #1358, opened 2020-02-17
Bump cardano-node and its dependencies to 1.5.0, and stack LTS in the meantime

KtorZ:
# Issue Number

#1346 

# Overview

- 843b188bcac961eca9cf2a5a5b2789fe2abe546d
  bump stack LTS to 14.26 & bump cardano-node to 1.5.0
  
- fdfdd3c7097be36cd6bbca5e608b3cf53fa3a831
  fix incorrect golden test file for Byron wallets
  There's actually no reward balance for Byron wallets!

- ddd57de771734c044fb1ae33c71df6d7dd0e3231
  fix parsing error messages due to LTS bump
  
- ad7080e9673aa10f5f1a8eb0d6afd6e3124d08ce
  re-generate nix machinery
  
- 4cd0790572e1a41d7e9f25b2fa90692797bb50a5
  make parsing error for stake pool id more consistent with others
  
- 47f553270ea043538da53c3bc75f5df87c5dc302
  Make API path params capture tests as unit tests
  These are needlessly taking time in the integration tests and above all, are duplicated all over the place
for all endpoints, whereas what we really care about it here is to test the behavior of the API when
capturing path params that are invalid. The same thing should be done for all the headers tests and basically,
any test that doesn't actually reaches the API handlers!

- 15fbe4bd7c34024f89235bf1af5876de4c798aa0
  Make API capture test generic in terms of the API
  
- dd9badf3032e1c2aa83df82ef25341c803e64b5a
  revise ApiEpochNumber malformation error handling
  There's no reason to have a branch in the ApiEpochNumber for malformed epoch number... this should really go
to the Left side of the Either when decoding them from textual representation such that it can be handled
by Servant altogether.

- a420164c352941a4f0c3f56afb874fc4318e72a6
  Capture MethodNotAllowed scenarios in unit tests too
  
- f704afaa129c45df08fdff617fc5144765ea4d6d
  do not spit back wrong dictionary word when parsing mnemonic
  This is against best-practices and makes code simply harder to test

- 2b14f0e5b3611522eafd9be414f641cbd955331a
  BREAKING-CHANGE: revise endpoint names for estimating fees
  Fixes parsers issues with servant-0.17 and, is closer to the REST philosophy

- 3c537fa0f21b50c64ca9dbcfbcd6bd05caca339d
  remove malformed param tests from integration (now done at the unit level)
  
- d3bb2e15054582905927303b0d281fa6d39ebec2
  BREAKING-CHANGE! remove backward-compatible placeholder
  Since we are pushing out some breaking changes, might as well
get rid of this backward compatible helper too!

- f43256733ae72ae80d13dd4b924afb121403d5bd
  Derive malformed header test generically from the API
  
- cfce4c1fd2d6cdd643f2aee9b7f45ca5238a975e
  remove header cases from integration tests (now done at the unit-level)
  
- bd712c6aab40dca45131fbaa364453cda9514628
  breaking change! Move wallet style down body payload
  Instead of being a path parameter, this makes the API more RESTFul.

- 99c6aa96efc5f092315366fe4840f7c81ca7601f
  reduce epoch length and slot duration
  And remove hard-coded blockchain parameters from integration tests

- cee5f4924347ea8e7adb510e484c52b2c8d40dc5
  remove redundant imports & class constraints
  
- cf2a1fb53eb991cc97d3f5bd50dae536bafecfaf
  white-list PUT /stake-pools/*/wallets/{walletId} for method not allowed test
  The router actually pick up this one for /stake-pools/{poolId}/wallets/{walletId} and returns
a valid error

- 7188d9f42e6cf061964277feaec9a57ddf8fc1b3
  use stack LTS-14.25 instead
  We depend on stackage.nix to build the application on CI, and 14.25 is the latest available.
We could also bump stackage.nix and haskell.nix but we are currently using an ad-hoc branch
which includes fixes about windows that we need to preserve.

- 647e3abf62139c56645620746035cc373a9cce58
  Fix nix/regenerate.sh (iohk-nix update) and use it in pipeline
  
- e309b2efbb0c3a4d7b29399c06423e1126e61b41
  re-generate nix machinery
  
- 6328e35951cdab35e98e5fc318a08de794fa2d19
  remove malformed unit tests in Jörmungandr, moved to core/ApiSpec already

# Comments

- I didn't bump to the very latest stack LTS but to the latest one still using GHC 8.6.5 to make this easier. Bumping the compiler version would likely break something in the nix pipeline.. so I'd rather not do it now if not needed...


lib/cli/src/Cardano/CLI.hs
```diff
@@ -1084,11 +1084,9 @@ handleResponse encode res = do
             BL8.putStrLn (encode a)
         Left e -> do
             let msg = case e of
-                    FailureResponse r -> fromMaybe
+                    FailureResponse _ r -> fromMaybe
                         (T.decodeUtf8 $ BL.toStrict $ responseBody r)
                         (decodeError $ responseBody r)
-                    ConnectionError t ->
-                        t
```
KtorZ, line 1092: Before, the `ConnectionError` branch used to hold a `Text`, it now holds a `SomeException`, so the default case `T.pack $ show e` handles it just fine.

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -348,7 +333,7 @@ type JoinStakePool n = "stake-pools"
     :> "wallets"
     :> Capture "walletId" (ApiT WalletId)
     :> ReqBody '[JSON] ApiWalletPassphrase
-    :> PutAcccepted '[JSON] (ApiTransaction n)
```
KtorZ, line 351: While updating this, I also realized why I couldn't import `PutAccepted` in the past despite being present in the doc....

`PutAcccepted` vs `PutAccepted` ... subtle.. :man_facepalming:

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -492,23 +477,6 @@ type PostExternalTransaction = "proxy"
...
-
--- | Any media type
-data Any
-
-instance Accept Any where
-    contentTypes _ = ("*" // "*") :|
-        -- We also 'conveniently' accept JSON format
-        [ "application" // "json"
-        , "application" // "json" /: ("charset", "utf-8")
-        ]
```
KtorZ, line 507: Not needed anymore, `servant-0.17` now has a `NoContentVerb` which doesn't require this awkward work-around for responses with `NoContent`.

lib/core/test/data/Cardano/Wallet/Api/ApiByronWallet.faulty.json
```diff
@@ -0,0 +1,293 @@
+{
```
KtorZ, line 1: Whoops.. shouldn't have committed this.

lib/core/test/data/Cardano/Wallet/Api/ApiByronWallet.json
```diff
@@ -1,330 +1,296 @@
...
+                "progress": {
+                    "quantity": 53,
+                    "unit": "percent"
+                }
             },
             "balance": {
-                "reward": {
-                    "quantity": 18,
-                    "unit": "lovelace"
-                },
```
KtorZ, line 15: Not sure how this slipped through but we had ByronWallet with reward balance in the golden files :thinking:

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -418,7 +412,7 @@ spec = do
             |] `shouldBe` (Left @String @(ApiT AddressPoolGap) msg)
 
         it "ApiT AddressPoolGap (not a integer)" $ do
-            let msg = "Error in $: expected Integer, encountered floating number\
+            let msg = "Error in $: parsing Integer failed, unexpected floating number\
```
piotr-iohk, line 443: Probably few integration tests will have to be updated also...

KtorZ: Yes, I am running them locally at the moment, waiting for them to complete before going through all the cases.

KtorZ:
```
1149 examples, 151 failures, 1 pending
```

Good.

piotr-iohk (approved):

Issue #1355, opened 2020-02-17
Implement restore hardware wallet endpoint


paweljakubas:
# Context

As a cardano-wallet API client,
I want to be able to restore a sequential wallet from an account public key 
so that I can use cardano-wallet to keep track of my wallet 
And I can keep my private keys in another place (e.g. a hardware device)

AC

Given I have a hardware wallet, 
And I export my public key,
Then I can restore to Cardano Wallet 
with public key from hardware wallet.

# Decision

The endpoint is to be implemented. Similar to function to `postShelleyWallet` is to be added, which accepts public account key and wallet name. Additional `mkSeqStateFromAccountXPub` function in AddressDiscovery module needs to be implemented.

# Acceptance Criteria

1.  `postShelleyHardwareWallet` is implemented
2.  `mkSeqStateFromAccountXPub` is implemented
3. corresponding integration tests emulating hardware stored private key are added

---

# Development 

# QA 

1. As the implementation was added the proof of concept integration test was added : 
```
it "WALLETS_CREATE_03 - Restoration from account public key"
```
Here, basically we are creating wallet from root private key, doing some action on it (sending money). then we are deleting it and restoring from account public key. and checking if the state of the wallet (money) is as expected. 

2. Also `mkSeqStateFromAccountXPub` was added with property test - in addition to `mkSeqStateFromRootXPrv` which previously was named `mkSeqState`.

3. CLI support is ONLY for root prv key based POST at this moment. not for account public key version. Maybe it would be good to check if POST wallet works in CLI as previously.

piotr-iohk:
@paweljakubas, I think that few more tests would be handy here on integration level (currently we only have the first one under the tag `WALLETS_CREATE_03`):
 

Tag | Desc | Tests
-- | -- | --
HW_WALLETS_01 | Can create wallet, wallet preserves funds |  
HW_WALLETS_02 | Can create wallet, wallet preserves delegation |  
HW_WALLETS_03 | Cannot do operations requiring private key on the wallet (appropriate error message is displayed) | tx, delegation, update pass
HW_WALLETS_04 | Can manage such wallet the same way as others | delete, update name, utxo, list tx, list addresses, force resync
HW_WALLETS_05 | Wallet is listed and can be GET |  
HW_WALLETS_06 | Test `account_public_key` values in payload |  (maybe this one can be tested lower)



In particular `HW_WALLETS_03` I see that currently it produces response with:
```
Code = 410,
[...]

Issue #1354, opened 2020-02-17
Prepare new endpoint for hardware wallet restoring


paweljakubas:
# Context

As a cardano-wallet API client,
I want to be able to restore a sequential wallet from an account public key 
so that I can use cardano-wallet to keep track of my wallet 
And I can keep my private keys in another place (e.g. a hardware device)

AC

Given I have a hardware wallet, 
And I export my public key,
Then I can restore to Cardano Wallet 
with public key from hardware wallet.

# Decision

We add POST endpoint to service that which accepts `public account key` and `wallet name`. 

# Acceptance Criteria

1. Endpoint with needed errors are added.
2. TypesSpec unit testing is updated
3. Swagger is updated

---

# Development 

# QA 

New type `WalletOrAccountPostData` was added. What is important in `swagger.yaml` it uses 3.0 feature `oneOf` and it would be worth double checking if either `AccountPostData` and `WalletPostData` is displayed as an alternative in POST `wallets`. Moreover, tests in `Api.TypesSpec` where added to check several things

piotr-iohk:
lgtm and closing. (Adding some tests in https://github.com/input-output-hk/cardano-wallet/pull/1400, but will keep #1355 in QA for that)

PR #1360, opened 2020-02-17
Extend wallet restore endpoint to accept also account public key 

paweljakubas:
# Issue Number

#1354 
#1355 

# Overview

- [x] I have updated swagger
- [x] I have added the endpoint and corresponding types
- [x] I have updated Api.TypesSpec
- [x] I have renamed `mkSeqState` to `mkSeqStateFromRootXPrv`
- [x] I have added `mkSeqStateFromAccountXPub` and property test
- [x] I have added impl of postAccountWallet
- [x] I have added integration tests illustrating the case

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -95,7 +95,8 @@ import Cardano.Wallet.Api
     , workerRegistry
     )
 import Cardano.Wallet.Api.Types
-    ( AddressAmount (..)
+    ( AccountPostData (..)
+    , AddressAmount (..)
     , ApiAddress (..)
     , ApiBlockReference (..)
     , ApiByronWallet (..)
@@ -132,6 +133,7 @@ import Cardano.Wallet.Api.Types
     , PostTransactionData
     , PostTransactionFeeData
     , WalletBalance (..)
+    , WalletOrAccountPostData (..)
     , WalletPostData (..)
     , WalletPutData (..)
     , WalletPutPassphraseData (..)
@@ -450,7 +452,7 @@ server byron icarus shelley spl =
     wallets = deleteWallet shelley
         :<|> (fmap fst . getWallet shelley mkShelleyWallet)
         :<|> (fmap fst <$> listWallets shelley mkShelleyWallet)
-        :<|> postShelleyWallet shelley
+        :<|> postWallet shelley
         :<|> putWallet shelley mkShelleyWallet
         :<|> putWalletPassphrase shelley
         :<|> getUTxOsStatistics shelley
@@ -646,6 +648,20 @@ type MkApiWallet ctx s w
     -> Handler w
 
 --------------------- Shelley
+postWallet
+    :: forall ctx s t k n.
+        ( s ~ SeqState n k
+        , k ~ ShelleyKey
+        , ctx ~ ApiLayer s t k
+        , HasDBFactory s k ctx
+        , HasWorkerRegistry s k ctx
+        )
+    => ctx
+    -> WalletOrAccountPostData
+    -> Handler ApiWallet
+postWallet ctx (WalletOrAccountPostData body) = case body of
+    Left body' -> postShelleyWallet ctx body'
+    Right body' -> postAccountWallet ctx body'
 
 postShelleyWallet
     :: forall ctx s t k n.
@@ -675,6 +691,11 @@ postShelleyWallet ctx body = do
     wid = WalletId $ digest $ publicKey rootXPrv
     wName = getApiT (body ^. #name)
 
+postAccountWallet
+    :: ctx
+    -> AccountPostData
+    -> Handler ApiWallet
+postAccountWallet _ctx _body = throwError err501
 
 mkShelleyWallet
     :: forall ctx s t k n.
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -165,7 +165,7 @@ import Cardano.Wallet.Primitive.AddressDiscovery
 import Cardano.Wallet.Primitive.AddressDiscovery.Random
     ( RndState, mkRndState )
 import Cardano.Wallet.Primitive.AddressDiscovery.Sequential
-    ( SeqState (..), defaultAddressPoolGap, mkSeqState )
+    ( SeqState (..), defaultAddressPoolGap, mkSeqStateFromRootXPrv )
 import Cardano.Wallet.Primitive.CoinSelection
     ( CoinSelection (..), changeBalance, feeBalance, inputBalance )
 import Cardano.Wallet.Primitive.Model
@@ -675,7 +675,7 @@ postShelleyWallet
     -> WalletPostData
     -> Handler ApiWallet
 postShelleyWallet ctx body = do
-    let state = mkSeqState (rootXPrv, pwd) g
+    let state = mkSeqStateFromRootXPrv (rootXPrv, pwd) g
     void $ liftHandler $ initWorker @_ @s @k ctx wid
         (\wrk -> W.createWallet  @(WorkerCtx ctx) @s @k wrk wid wName state)
         (\wrk -> W.restoreWallet @(WorkerCtx ctx) @s @t @k wrk wid)
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -96,6 +96,7 @@ import Cardano.Wallet.Api
     )
 import Cardano.Wallet.Api.Types
     ( AccountPostData (..)
+    , AccountPublicKey (..)
     , AddressAmount (..)
     , ApiAddress (..)
     , ApiBlockReference (..)
@@ -165,7 +166,11 @@ import Cardano.Wallet.Primitive.AddressDiscovery
 import Cardano.Wallet.Primitive.AddressDiscovery.Random
     ( RndState, mkRndState )
 import Cardano.Wallet.Primitive.AddressDiscovery.Sequential
-    ( SeqState (..), defaultAddressPoolGap, mkSeqStateFromRootXPrv )
+    ( SeqState (..)
+    , defaultAddressPoolGap
+    , mkSeqStateFromAccountXPub
+    , mkSeqStateFromRootXPrv
+    )
 import Cardano.Wallet.Primitive.CoinSelection
     ( CoinSelection (..), changeBalance, feeBalance, inputBalance )
 import Cardano.Wallet.Primitive.Model
@@ -692,10 +697,27 @@ postShelleyWallet ctx body = do
     wName = getApiT (body ^. #name)
 
 postAccountWallet
-    :: ctx
+    :: forall ctx s t k n.
+        ( s ~ SeqState n k
+        , k ~ ShelleyKey
+        , ctx ~ ApiLayer s t k
+        , HasWorkerRegistry s k ctx
+        )
+    => ctx
     -> AccountPostData
     -> Handler ApiWallet
-postAccountWallet _ctx _body = throwError err501
+postAccountWallet ctx body = do
+    let state = mkSeqStateFromAccountXPub accXPub g
+    void $ liftHandler $ initWorker @_ @s @k ctx wid
+        (\wrk -> W.createWallet  @(WorkerCtx ctx) @s @k wrk wid wName state)
+        (\wrk -> W.restoreWallet @(WorkerCtx ctx) @s @t @k wrk wid)
+    fst <$> getWallet ctx (mkShelleyWallet @_ @s @t @k) (ApiT wid)
+  where
+    g = maybe defaultAddressPoolGap getApiT (body ^. #addressPoolGap)
+    wName = getApiT (body ^. #name)
+    (AccountPublicKey accXPubApiT) =  body ^. #accountPublicKey
+    accXPub = getApiT accXPubApiT
+    wid = WalletId $ digest accXPub
 
 mkShelleyWallet
     :: forall ctx s t k n.
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -80,6 +80,11 @@ module Cardano.Wallet.Api.Types
     , ApiByronWalletMigrationInfo (..)
     , ByronWalletPostData (..)
 
+    -- * API Types (Hardware)
+    , AccountPostData (..)
+    , AccountPublicKey (..)
+    , WalletOrAccountPostData (..)
+
     -- * User-Facing Address Encoding/Decoding
     , EncodeAddress (..)
     , DecodeAddress (..)
@@ -95,17 +100,20 @@ import Prelude
 import Cardano.Pool.Metadata
     ( StakePoolMetadata )
 import Cardano.Wallet.Primitive.AddressDerivation
-    ( FromMnemonic (..)
+    ( Depth (..)
+    , FromMnemonic (..)
     , NetworkDiscriminant (..)
     , Passphrase (..)
     , PassphraseMaxLength (..)
     , PassphraseMinLength (..)
+    , PersistPublicKey (..)
     , SomeMnemonic (..)
+    , XPub
     )
 import Cardano.Wallet.Primitive.AddressDerivation.Byron
     ( decodeLegacyAddress )
 import Cardano.Wallet.Primitive.AddressDerivation.Shelley
-    ( decodeShelleyAddress )
+    ( ShelleyKey (..), decodeShelleyAddress, xpubFromText )
 import Cardano.Wallet.Primitive.AddressDiscovery.Sequential
     ( AddressPoolGap, getAddressPoolGap )
 import Cardano.Wallet.Primitive.Mnemonic
@@ -159,6 +167,7 @@ import Data.Aeson
     , tagSingleConstructors
     , withObject
     , (.:)
+    , (.:?)
     , (.=)
     )
 import Data.Bifunctor
@@ -344,6 +353,20 @@ data ByronWalletPostData mw = ByronWalletPostData
     { mnemonicSentence :: !(ApiMnemonicT mw)
     , name :: !(ApiT WalletName)
     , passphrase :: !(ApiT (Passphrase "encryption"))
+    } deriving (Eq, Generic, Show)
+
+newtype AccountPublicKey = AccountPublicKey
+    { key :: (ApiT (ShelleyKey 'AccountK XPub))
+    } deriving (Eq, Generic, Show)
+
+newtype WalletOrAccountPostData = WalletOrAccountPostData
+    { postData :: Either WalletPostData AccountPostData
+    } deriving (Eq, Generic, Show)
+
+data AccountPostData = AccountPostData
+    { name :: !(ApiT WalletName)
+    , accountPublicKey :: !AccountPublicKey
+    , addressPoolGap :: !(Maybe (ApiT AddressPoolGap))
     } deriving (Eq, Generic, Show)
 
 newtype WalletPutData = WalletPutData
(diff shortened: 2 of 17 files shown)


lib/core/test/data/Cardano/Wallet/Api/HardwareWalletPostData.json
```diff
@@ -0,0 +1,45 @@
+{
+    "seed": -4365653956898380124,
+    "samples": [
+        {
+            "account_public_key": "6562643630396264366339313064643635653532306663393630306235353336353539623864363366653965313164613866633565366531633632626635653965343239343432613732653338613234383764613639393462613532376363376166636337366636323736643363393830366437333063363563646466636261",
```
paweljakubas: I wonder if this account public key generated is realistic :thinking:

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -658,6 +674,22 @@ instance FromJSON WalletPostData where
 instance ToJSON  WalletPostData where
     toJSON = genericToJSON defaultRecordTypeOptions
 
+instance FromJSON AccountPublicKey where
+    parseJSON bytes = do
+        xs <- parseJSON bytes
+        case convertFromBase Base16 $ T.encodeUtf8 xs of
+            Left _ -> fail "wrong account public key format"
+            Right bs ->
+                pure $ AccountPublicKey $ ApiT $ unsafeDeserializeXPub bs
```
paweljakubas: I am not sure if `unsafeDeserializeXPub` is ok here? The same for `serializeXPub` below

Anviking: Will it not throw (i.e "something went wrong" for the user) if provided valid hex, but which is an invalid key?

lib/core/test/data/Cardano/Wallet/Api/HardwareWalletPostData.json
KtorZ: The account public key is 64 bytes in length, so 128 characters hex-encoded.

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: Whola! Yes. Do not use any `unsafe` method in the API when presented user-facing inputs :scream: ! 
The "unsafe" constructors are only tolerated at the boundaries with the database, where we have controlled first what we inserted!

lib/core/test/data/Cardano/Wallet/Api/HardwareWalletPostData.json
paweljakubas: indeed, I corrected the corresponding code

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: it was temporary...now removed unsafe code. thanks for being vigilant here!

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -346,6 +355,20 @@ data ByronWalletPostData mw = ByronWalletPostData
     , passphrase :: !(ApiT (Passphrase "encryption"))
     } deriving (Eq, Generic, Show)
 
+newtype AccountPublicKey = AccountPublicKey
+    { key :: (ApiT (ShelleyKey 'AccountK XPub))
+    } deriving (Eq, Generic, Show)
+
+newtype PostData = PostData
+    { postData :: Either WalletPostData AccountPostData
```
Anviking, line 363: So the plan is to re-use the existing endpoint? Why is this preferable to having two endpoints?

When a wallet is created using just an account public key, will that not require special treatment inside the `cardano-wallet-core`, as it has no private key? Or does existing abstractions handle this already?

And if using the same endpoint, maybe call it `WalletOrAccountPostData` instead of `PostData`? Or `WalletPostData = SomeName WalletMnemonicPostData | SomeName2 WalletAccountPostData` or something 🤷‍♂  `PostData` sounds a bit too generic at least.

paweljakubas: @Anviking yes, we will opt for ONE endpoint here. We talked about this with @KtorZ  and the conclusion was (when thinking about two endpoints): 
```
We are building a REST API, and we have to follow patterns and idioms of REST API. In REST, everything is about resources. Here, we are creating the same resource, so it should be the same endpoint.
```
Nevertheless, `WalletOrAccountPostData` is better than `PostData` - I agree

KtorZ: > does existing abstractions handle this already?

Precisely. Attaching a private key is done as a separate operation, and, operation that requires a private key explicitly fails if there's none (e.g. signing tx)

> maybe call it WalletOrAccountPostData instead of PostData

:+1:

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -659,7 +680,7 @@ postShelleyWallet
     -> WalletPostData
     -> Handler ApiWallet
 postShelleyWallet ctx body = do
-    let state = mkSeqState (rootXPrv, pwd) g
+    let state = mkSeqStateFromRootXPrv (rootXPrv, pwd) g
```
Anviking, line 683: 👍

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Shelley.hs
```diff
@@ -450,8 +451,11 @@ instance PersistPublicKey (ShelleyKey depth) where
     unsafeDeserializeXPub =
         either err ShelleyKey . xpubFromText
       where
-        xpubFromText = xpub <=< fromHex @ByteString
         err _ = error "unsafeDeserializeXPub: unable to deserialize ShelleyKey"
+
+xpubFromText :: ByteString -> Either String XPub
```
Anviking, line 456: But we lose the ability to distinguish between "invalid hex" and "correct hex, invalid data" from the `FromJSON` because of this function?

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
```diff
@@ -571,6 +572,27 @@ mkSeqState (rootXPrv, pwd) g =
...
+        , MkKeyFingerprint k (k 'AddressK XPub)
+        , MkKeyFingerprint k Address
+        )
+    => k 'AccountK XPub
+    -> AddressPoolGap
+    -> SeqState n k
+mkSeqStateFromAccountXPub accXPub g =
+    let
+        rewardXPub =
+            deriveAddressPublicKey accXPub MutableAccount minBound
```
Anviking, line 590: So this matches the reward address for "normal wallets", right? Assuming the accountXPub is the first account.

Here `minBound` is _also_ the first _Soft_ index  https://github.com/input-output-hk/cardano-wallet/blob/b9daac5b448308cd807163028af4b90d71f22352/lib/core/src/Cardano/Wallet.hs#L1167

because of 
```haskell
instance HardDerivation ShelleyKey where
    type AddressIndexDerivationType ShelleyKey = 'Soft
```
(so it should work. _Wasn't obvious to me at first_)

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -1035,6 +1045,22 @@ instance Arbitrary SortOrder where
     arbitrary = arbitraryBoundedEnum
     shrink = genericShrink
 
+instance Arbitrary WalletOrAccountPostData where
+    arbitrary = do
+        let walletPostDataGen = arbitrary :: Gen WalletPostData
+        let accountPostDataGen = arbitrary :: Gen AccountPostData
+        oneof [ WalletOrAccountPostData . Left <$> walletPostDataGen
+              , WalletOrAccountPostData . Right <$> accountPostDataGen ]
```
Anviking, line 1065: Now, there's no shrinking. Might not be _that_ much of a problem…?

But is there any specific reason not to use `genericArbitrary` and `genericShrink` for `WalletOrAccountPostData`, and `AccountPostData` too?

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/SequentialSpec.hs
```diff
@@ -335,16 +339,32 @@ prop_poolEventuallyDiscoverOurs _proxy (g, addr) =
...
     mw = someDummyMnemonic (Proxy @12)
     key = Shelley.unsafeGenerateKeyFromSeed (mw, Nothing) mempty
-    s0 = mkSeqState (key, mempty) g
+    s0 = mkSeqStateFromRootXPrv (key, mempty) g
+    prop =
+        length (fst $ changeAddresses [] s0) === fromEnum g
+
+-- | We can always generate at exactly `gap` change addresses (on the internal
+-- chain) using mkSeqStateFromAccountXPub
+prop_genChangeGapFromAccountXPub
```
Anviking, line 357: Is this really needed?

Is this not more a property of `SeqState` than of `mkSeqStateFromRootXPrv`/`mkSeqStateFromAccountXPub `?

lib/core/src/Cardano/Wallet/Api/Types.hs
Anviking: Btw, I find the use of `Either` (`postData :: Either WalletPostData AccountPostData`) _slightly_ weird because `Left` and `Right` imply success and error to me, and it has monad / functor instances. **But don't mind too much.**

I guess the tricky thing with `WalletOrAccountPostData = C1 WalletPostData | C2 AccountPostData` is naming the constructors. I'm not sure what conventions there are on it, but this is an example from `cardano-ledger`:

https://github.com/input-output-hk/cardano-ledger/blob/master/cardano-ledger/src/Cardano/Chain/Block/Block.hs#L359-L362

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -301,6 +309,8 @@ spec = do
             jsonRoundtripAndGolden $ Proxy @(PostTransactionData 'Testnet)
             jsonRoundtripAndGolden $ Proxy @(PostTransactionFeeData 'Testnet)
             jsonRoundtripAndGolden $ Proxy @WalletPostData
+            jsonRoundtripAndGolden $ Proxy @AccountPostData
+            jsonRoundtripAndGolden $ Proxy @WalletOrAccountPostData
```
Anviking, line 315: 👍

Anviking (approved):
Lgtm.
Maybe most significantly (see comments):
- is `prop_genChangeGapFromAccountXPub` really needed?
- Should we not be able to distinguish between invalid hex encoding and invalid data?

specifications/api/swagger.yaml
```diff
@@ -845,6 +853,20 @@ definitions:
...
+    type: object
+    required:
+      - name
+      - account_public_key
+    properties:
+      name: *walletName
+      account_public_key: *walletAccountPubkey
+      address_pool_gap: *walletAddressPoolGap
+
+  ApiWalletOrAccountPostData: &ApiWalletOrAccountPostData
```
paweljakubas: so I have problem here. how to express `oneof` . This one unfortunately gives : 
```
 1) Cardano.Wallet.Api.Types, verify that every type used with JSON content type in a servant API has compatible ToJSON and ToSchema instances using validateToJSON., WalletOrAccountPostData
       Falsifiable (after 3 tests):
         WalletOrAccountPostData {postData = Right (AccountPostData {name = ApiT {getApiT = WalletName {getWalletName = ":p6&vFta~? Yp\147454n\30864rWs~\"!r#:ms 3+KZUelcepj>n$0F99`$UVGunr\27692}V0xnw3+0CJ_fe/\152955N|)EBVO6Yd[u%,(rSJ:"}}, accountPublicKey = AccountPublicKey {key = ApiT {getApiT = ShelleyKey {getKey = XPub {xpubPublicKey = "Qd\237Bm\207\154\179\ESC>\208I\219\236c|\b\174\141\199w\179\SUB\242\183\RS\219\186\&7\179\&5\194", xpubChaincode = ChainCode "\226c\ENQP\DC3#\FS\"\221\133\150n@[33\154\138xh\133%\187V\181^6,v\220\131l"}}}}, addressPoolGap = Nothing})}
         Validation against the schema fails:
           * property "mnemonic_sentence" is required, but not found in "{\"account_public_key\":\"5164ed426dcf9ab31b3ed049dbec637c08ae8dc777b31af2b71edbba37b335c2e263055013231c22dd85966e405b33339a8a78688525bb56b55e362c76dc836c\",\"name\":\":p6&vFta~? Yp\240\163\191\190n\231\162\144rWs~\\\"!r#:ms 3+KZUelcepj>n$0F99`$UVGunr\230\176\172}V0xnw3+0CJ_fe/\240\165\149\187N|)EBVO6Yd[u%,(rSJ:\"}"
           * property "passphrase" is required, but not found in "{\"account_public_key\":\"5164ed426dcf9ab31b3ed049dbec637c08ae8dc777b31af2b71edbba37b335c2e263055013231c22dd85966e405b33339a8a78688525bb56b55e362c76dc836c\",\"name\":\":p6&vFta~? Yp\240\163\191\190n\231\162\144rWs~\\\"!r#:ms 3+KZUelcepj>n$0F99`$UVGunr\230\176\172}V0xnw3+0CJ_fe/\240\165\149\187N|)EBVO6Yd[u%,(rSJ:\"}"
           * property "account_public_key" is found in JSON value, but it is not mentioned in Swagger schema
         
         JSON value:
         {
             "account_public_key": "5164ed426dcf9ab31b3ed049dbec637c08ae8dc777b31af2b71edbba37b335c2e263055013231c22dd85966e405b33339a8a78688525bb56b55e362c76dc836c",
             "name": ":p6&vFta~? Yp𣿾n碐rWs~\"!r#:ms 3+KZUelcepj>n$0F99`$UVGunr氬}V0xnw3+0CJ_fe/𥕻N|)EBVO6Yd[u%,(rSJ:"
         }
         
         Swagger Schema:
         {
             "required": [
                 "name",
                 "mnemonic_sentence",
                 "passphrase"
             ],

```

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -658,6 +681,39 @@ instance FromJSON WalletPostData where
...
+        val <-
+            (withObject "postData" $
+             \o -> o .:? "passphrase" :: Aeson.Parser (Maybe Text)) obj
+        case val of
+            Nothing -> do
+                xs <- parseJSON obj :: Aeson.Parser AccountPostData
+                pure $ WalletOrAccountPostData $ Right xs
+            Just _ -> do
+                xs <- parseJSON obj :: Aeson.Parser WalletPostData
+                pure $ WalletOrAccountPostData $ Left xs
```
KtorZ, line 710: Tips: parsers are alternatives. So we can simply write:

```hs
parseJSON obj = WalletOrAccountPostData <$>
    (Left <$> parseJSON obj <|> Right <$> parseJSON obj)
```

;)

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
KtorZ: >  Wasn't obvious to me at first

This indeed deserve a proper note as a comment in the source code. All your assumptions are correct, but we need to explain _why_ they are and what's the rationale behind this.

lib/core/src/Cardano/Wallet/Api/Types.hs
Anviking: Could it not potentially result in worse error messages?

If there is a slight mistake in the `Left` branch, will it not continue with the `Right` branch and might produce an error message specific to the right branch?

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
Anviking: Another gotcha I realised was that:
```haskell
deriveAddressPrivateKey _ _ _ minBound
```
infers `minBound` to be the first **soft** index (for ShelleyKey)

but for
```haskell
deriveAccountPrivateKey _ _ _ minBound
```
`minBound` is the first **hardened** index (regardless of `AddressIndexDerivationType`). https://github.com/input-output-hk/cardano-wallet/blob/b9daac5b448308cd807163028af4b90d71f22352/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs#L285

(hence the `0x80000000` index in the integration test)

_type-inference…_ 😅

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: That's pretty much the case now too. If you forget the `passphrase` (or mistype it), we'll just assume an `AccountPostData` :man_shrugging:

We can probably do something more elaborate like we did for `FromMnemonic`. My two cents would actually be that, we should probably collect the error message from both branches and provide both to the user in a readable way.

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -1470,6 +1495,12 @@ instance ToSchema ApiTxId where
 instance ToSchema WalletPostData where
     declareNamedSchema _ = declareSchemaForDefinition "ApiWalletPostData"
 
+instance ToSchema AccountPostData where
+    declareNamedSchema _ = declareSchemaForDefinition "ApiAccountPostData"
+
+instance ToSchema WalletOrAccountPostData where
+    declareNamedSchema _ = declareSchemaForDefinition "ApiWalletOrAccountPostData"
```
paweljakubas: will need probably to do something here, utilize `proxy` in `declareNamedSchema proxy` and basing on that in one case 
```
declareSchemaForDefinition "ApiAccountPostData"
```
and 
```
declareSchemaForDefinition "ApiWalletPostData"
```

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: done

specifications/api/swagger.yaml
paweljakubas: fixed in https://github.com/input-output-hk/cardano-wallet/pull/1360/commits/794224a42c9a7f7e646bdd58a76572f670c1260c

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/SequentialSpec.hs
paweljakubas: I just treat it as some place to check that both `mkSeqStateFromRootXPrv` and `mkSeqStateFromAccountXPub` produce consistent results. The test is not utterly persuasive but gives some value added (in my modest opinion) ;-)

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
paweljakubas: We are going to have either one or other object - I think it is ok to see them as they fail

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
paweljakubas: added the suggested comment

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
paweljakubas: fixed in 794224a

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: hmmm, I have a number of things like that : 
```

1) API Specifications, WALLETS_CREATE_05 - Mnemonics, [] as mnemonic_sentence -> fail
--
  | "Error in $: key \"id\" not present: Response {responseStatus = Status {statusCode = 400, statusMessage = \"Bad Request\"}, responseVersion = HTTP/1.1, responseHeaders = [(\"Transfer-Encoding\",\"chunked\"),(\"Date\",\"Mon, 24 Feb 2020 19:49:06 GMT\"),(\"Server\",\"Warp/3.2.27\"),(\"Content-Type\",\"application/json;charset=utf-8\")], responseBody = \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"Error in $: key 'account_public_key' not present\\\"}\", responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}" does not contain "Invalid number of words: 15, 18, 21 or 24 words are expected."
```
will need to work on error management in this instance

paweljakubas:
bors retry

Issue #1342, opened 2020-02-12
Multiple nexts in delegation reporting


paweljakubas:
# Context

U/S

As a wallet client,
I want to reliably know how my delegation changes irrespective of how often certificate is discovered,
So that I can better inform users about their current and next delegation settings.

A/C

The current implementation handles well cases when the certificate is discovered at most once per 2 epochs. We may want to handle situations where : 

In a given epoch there were more than one certificate discovered
Certificate(s) could be discovered in the adjacent epochs
If so, we may want to take the latest certificate in a given epoch to determine what is next. Moreover, if two (or more) certificates are discovered in adjacent epochs we may want to have two `next` in reporting. 

# Decision

We will define WalletDelegation that accommodates list of nexts

# Acceptance Criteria

1. Logic is in Primitive.Types' WalletDelegation
2. Integration test showing the case is added. 

---

# Development 

- Revise api quit endpoint to use a `*` instead of an explicit pool id ==> TODO

# QA

piotr-iohk:
A "corner case":

1. have a wallet delegating to pool A
2. in epoch X delegate to pool B
```
I have:
delegation active = A
delegation next = Target B, changes_at: X+2
```
  3. in epoch X+1 quit pool A
```
I have:
delegation active = A
delegation next1 = Target B, changes_at: X+2
delegation next2= not_delegating, changes_at: X+3
 - a bit wierd, because delegating to Target B, will already automatically quit my active target A, so why my delegation next2 is not_delegating?
```
  4. in X + 2
```
delegation active = B
delegation next = not_delegating, changes_at: X+3
...
```
 5. in X + 3
delegation active = not_delegating
```
So, I have made a request to quit A and as a result I have quit B...
```


## Expected behavior

Perhaps, when you are delegating to A and have already any next delegation target active, you shouldn't be able to request quit A in a first place...

or

```
DELETE /stake-pools/{stakePoolId}/wallets/{walletId}
```
Should not be taking {stakePoolId} as a parameter not to mislead.

piotr-iohk:
I will move to in progress for now until #1386 is done.

piotr-iohk:
LGTM. Testing addressed in #1386.

PR #1361, opened 2020-02-18
Testing delegation active/next

piotr-iohk:
# Issue Number 
#1342

# Overview

- a57326ac1f71934c222856eda1245ad7110750e5
  Adjust quitStakePool/joinStakePool to not allow joining/quitting only when last pending interferes
  
- cc508344afdfa1811ae506f3448dc26b21c7c1c3
  Nix regenerate
  
- d87907c870947a4def285d68efd8bb172b2a9b6e
  Possible fix for NETWORK_BYRON

# Comments


paweljakubas (approved):
fond of introducing/using `waitForNextEpoch` and `waitAllTxsInLedger`

piotr-iohk:
I let myself adding also:

44d97a0
Forbid quitting SP when not_delegating is already a target in the delegation next

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1417,6 +1417,9 @@ quitStakePool ctx wid argGenChange pwd = db & \DBLayer{..} -> do
     unless (isDelegatingTo anyone (walMeta ^. #delegation)) $
         throwE ErrQuitStakePoolNotDelegating
 
+    when (any statusNotDelegating (walMeta ^. #delegation . #next)) $
+        throwE ErrQuitStakePoolNextNotDelegating
```
KtorZ: The check should be on the most recent pending delegation only. If i quit delegation an epoch E+1, but I join again at epoch E+2, then the API should accept any delegation certificate to replace the most recent one at E+2.

Also, joining likely needs some adjustments. Right now, we do prevent delegation if the wallet is delegating toa particular pool at any stage. But, I could be in a situation where I have:

E: delegate to P
E+1: quit
E+2: -

It'd be totally okay to delegate to P at this stage but the API would prevent it. Again, when making assertions on the delegation status in at the API, I think we should only be looking at the status at E+2.

piotr-iohk: I have adjusted join and quit stake pool as pointed :point_up_2: and updated integration tests.

af12bb8
Adjust quitStakePool/joinStakePool to not allow joining/quitting only when last pending interferes

d70ae21
Adjust integration tests...

I think that integration tests are fairly exhaustive (`STAKE_POOLS_NEXT`), but they are quite long:

```
Finished in 566.3162 seconds
12 examples, 0 failures
```

KtorZ (approved):

piotr-iohk:
So, yeah, buildkite times out:
```
Timed out after 30 minutes.
```
Cause newly added tests add ~9 minutes to execution, which make integration tests to run for slightly more than 30 mins now.

KtorZ:
> Cause newly added tests add ~9 minutes to execution

Awe... Should we consider maybe moving these cases at a unit level? I mean, we aren't much testing the integration at this point, but simply different scenarios of delegation given different starting point. 

We could very much have this at the unit level if we move the conditions for delegating / retiring to separate pure functions (e.g. "canDelegate" "canRetire")

piotr-iohk:
>Awe... Should we consider maybe moving these cases at a unit level? 

If that's possible then why not :)

piotr-iohk:
OK, as discussed leaving in this PR:

d30ce0d
I have reorganized a bit integration tests for "stake pool active/next" groupping them into 5 scenarios... (so basically the same tests as before but re-organized)
2d9e46e
Adjust quitStakePool/joinStakePool to not allow joining/quitting only when last pending interferes

And will open another PR with further integration tests with the intention to move them into unit tests next week...

bors r+

piotr-iohk:
Buildkite still times out, so leaving only:
36d3dcd - Adjust quitStakePool/joinStakePool to not allow joining/quitting only when last pending interferes

Tests in another pr.

bors r+

KtorZ:
Interesting, how does it work locally ?

piotr-iohk:
Hm, Actually exec time is ok but 3 integration tests failed for me (however not related with the change I think):

```
Failures:

  test/integration/Test/Integration/Jormungandr/Scenario/CLI/Server.hs:96:9: 
  1) CLI Specifications, DaedalusIPC [SERIAL], Should reply with the port --random
       uncaught exception: IOException of type NoSuchThing
       node: createProcess: runInteractiveProcess: exec: does not exist (No such file or directory)

  To rerun use: --match "/CLI Specifications/DaedalusIPC [SERIAL]/Should reply with the port --random/"

  test/integration/Test/Integration/Jormungandr/Scenario/CLI/Server.hs:102:9: 
  2) CLI Specifications, DaedalusIPC [SERIAL], Should reply with the port --port
       uncaught exception: IOException of type NoSuchThing
       node: createProcess: runInteractiveProcess: exec: does not exist (No such file or directory)

  To rerun use: --match "/CLI Specifications/DaedalusIPC [SERIAL]/Should reply with the port --port/"

  test/integration/Test/Integration/Jormungandr/Scenario/CLI/Server.hs:109:9: 
  3) CLI Specifications, DaedalusIPC [SERIAL], Regression test for #1036
       uncaught exception: IOException of type NoSuchThing
       node: createProcess: runInteractiveProcess: exec: does not exist (No such file or directory)

  To rerun use: --match "/CLI Specifications/DaedalusIPC [SERIAL]/Regression test for #1036/"

Randomized with seed 1731288685

Finished in 1193.6961 seconds
1149 examples, 3 failures, 1 pending
```

piotr-iohk:
The same tests fail for me locally on master, so I don't know... maybe related to Ubuntu 19.10, to which I just switched...

bors r+

KtorZ:
@piotr-iohk you need to have Node.js installed and available in your $PATH to run these tests. So, it's probably normal if you don't.

nikolaglumac:
@KtorZ any update on when this PR could be merged? This is currently blocking us from releasing next Daedalus version...
Or is there another PR which excludes tests and is perhaps already merged?

piotr-iohk:
now this integration test fails -> `NETWORK_BYRON`, which I cannot reproduce locally... (issued: https://github.com/input-output-hk/cardano-wallet/issues/1368)

bors r+

paweljakubas (approved):

piotr-iohk:
bors try

piotr-iohk:
NETWORK_BYRON failing tests appears to be fixed in d87907c870947a4def285d68efd8bb172b2a9b6e.

bors r+

lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
```diff
@@ -125,11 +126,12 @@ spec = do
...
-            eventually_ $ do
+            -- make sure you're at the beginning of the epoch
+            waitForNextEpoch ctx
+            r <- eventually $ do
                 sync <- getNetworkInfo
-                verify sync
-                    [ expectField (#syncProgress . #getApiT) (`shouldBe` Ready) ]
-            r <- getNetworkInfo
+                expectField (#syncProgress . #getApiT) (`shouldBe` Ready) sync
+                return sync
```
KtorZ, line 134: Ah!

PR #1362, opened 2020-02-19
Remove local `Codec.Binary.Bech32.TH` module.

jonathanknowles:
Use the `bech32-th` library from Hackage, instead.

# Issue Number

#1333 

# Overview

In this PR, we:

- [x] Remove the local copy of the `Codec.Binary.Bech32.TH` module.
- [x] Use the version provided by the `bech32-th` Hackage library instead.
- [x] Update the basic `nix` machinery.
- [x] Update the `nix/haskell-nix-src.json` to include a newer version of Hackage.
    This must include version `1.0.2` of both `bech32` and `bech32-th`.


KtorZ (approved):
bors r+

paweljakubas (approved):

PR #1363, opened 2020-02-20
Create new cardano-wallet-core-arbitrary library & move Cardano.Wallet.Gen there

Anviking:
# Issue Number

#1316 

# Overview

- [x] I moved `Cardano.Wallet.Gen` from `cardano-wallet-core:unit` to a new library such that it can be re-used from other test-suites.

# Comments

- I had trouble placing it in `test-utils` in the past (got a strange type-mismatch error). I now believe this was due to cyclic dependencies. (Cabal hinted at that.) So I created a completely new package.


Anviking:
Oh well…
```
Running Haddock on library 'cardano-wallet-core-arbitrary' for cardano-wallet-core-2020.2.17..
--
  | Cabal-simple_mPHDZzAJ_2.4.0.1_ghc-8.6.5: internal error when calculating
  | transitive package dependencies.
  | Debug info: []
  | error: Command exited with code 1!
```
https://github.com/haskell/cabal/issues/1919
👀

Anviking:
Sadly:
- `cardano-wallet-core:cardano-wallet-core-arbitrary` will not work because of the haddock issue
- `cardano-wallet-arbitrary` (separate package) will not work due to circular package dependency

PR #1369, opened 2020-02-24
Add unXPrvStripPub (& inverse) that matches jcli

Anviking:
# Issue Number

#1316 

# Overview

- [x] Added
```haskell
unXPrvStripPub :: XPrv -> Either ErrUnXPrvStripPub ByteString
xPrvFromStrippedPubXPrv :: ByteString -> Either ErrXPrvFromStrippedPubXPrv XPrv
```

To convert `XPrv` (`prv <> pub <> cc`) to a 96-byte bytestring of the form `prv <> cc`, and back.

Unusually, _both_ functions may fail. `unXPrvStripPub` fails if the resulting byte string cannot be converted back to the _same_ `XPrv` using `xPrvFromStrippedPubXPrv` (i.e. roundtrip)

- [x] Roundtrip properties.
    - `either roundtrips or fails (if xprv is encrypted)`
    - `(xPrvFromStrippedPubXPrv bs) fails if (BS.length bs) /= 96`
- [x] Integration test verifying that this format works with jcli

# Comments

- We need to convert between 96-byte long hex-encoded bytestrings and
XPrvs when implementing the `key root` and `key child` CLI commands.

Example showing that we cannot `unXPrvStripPub` encrypted XPrvs:
```haskell
λ> bytes = "(\134\242|I\141L\EM\NUL\128\173\252q\191\172\167>f \218\222\167.\136\DC4\216\191\253r8cD8&I\STX;\185&\177\172E\241\185\241\157\226\r\163+\EM\GS\232-\188\250[E^N\129J\158\STX\242s\215\149\142\217;\168P\DLE\159jHO\149'\ENQ\ETX!\129\CAN\140\176\221\DEL'*\DC4=<B\226\134\188!\241\DLEzt\222\199\247U\143\ETB\128,\226Q\"\230\234\"\191\177\250\230\167\n\214X\244z\\" :: BS.ByteString

λ> unXPrvStripPub k
Right "(\134\242|I\141L\EM\NUL\128\173\252q\191\172\167>f \218\222\167.\136\DC4\216\191\253r8cD8&I\STX;\185&\177\172E\241\185\241\157\226\r\163+\EM\GS\232-\188\250[E^N\129J\158\STX\134\188!\241\DLEzt\222\199\247U\143\ETB\128,\226Q\"\230\234\"\191\177\250\230\167\n\214X\244z\\"

λ> unXPrvStripPub $ CC.xPrvChangePass (""::BS.ByteString) ("newpass"::BS.ByteString) k
Left ErrNoRoundtripMismatch
```

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
@@ -55,8 +55,14 @@ module Cardano.Wallet.Primitive.AddressDerivation
     , unXPrv
     , xprv
     , xpub
+
+    -- * Helpers
     , hex
     , fromHex
+    , unXPrvStripPub
+    , xPrvFromStrippedPubXPrv
+    , ErrXPrvFromStrippedPubXPrv (..)
+    , ErrUnXPrvStripPub (..)
 
     -- * Network Discrimination
     , NetworkDiscriminant (..)
@@ -107,7 +113,7 @@ import Control.Arrow
 import Control.DeepSeq
     ( NFData )
 import Control.Monad
-    ( unless )
+    ( unless, when )
 import Crypto.Hash
     ( Digest, HashAlgorithm )
 import Crypto.KDF.PBKDF2
@@ -153,6 +159,7 @@ import GHC.TypeLits
 import Type.Reflection
     ( typeOf )
 
+import qualified Cardano.Crypto.Wallet.Encrypted as CC
 import qualified Data.ByteArray as BA
 import qualified Data.ByteString as BS
 import qualified Data.Text as T
@@ -720,3 +727,67 @@ hex = convertToBase Base16
 -- | Decode a 'ByteString' from base16
 fromHex :: ByteArray bout => ByteString -> Either String bout
 fromHex = convertFromBase Base16
+
+data ErrUnXPrvStripPub
+    = ErrCannotRoundtrip
+      -- ^ The resulting bytestring would have been unable to roundtrip using
+      -- @xPrvFromStrippedPubXPrv@. Most likely because the input @XPrv@ was
+      -- encrypted, or because it was an old (Byron) key.
+    deriving (Eq, Show)
+
+-- | Convert a @XPrv@ to a 96-byte long extended private key that does /not/
+-- include the public key.
+--
+-- The format is:
+-- > Extended Private Key (64 bytes) <> ChainCode (32 bytes)
+--
+-- Returns @Left@ if the resulting bytestring fails to roundtrip back to the
+-- original @XPrv@. This can happen:
+-- - If the @XPrv@ was encrypted
+-- - If a DerivationScheme1 (Byron) key was used (that does not conform to the
+-- "tweak")
+unXPrvStripPub :: XPrv -> Either ErrUnXPrvStripPub ByteString
+unXPrvStripPub k = do
+    let res = stripPub . unXPrv $ k
+
+    -- Check that it roundtrips.
+    case (fmap unXPrv . xPrvFromStrippedPubXPrv $ res) of
+        Right bytes
+            | bytes == unXPrv k -> Right res
+            | otherwise         -> Left ErrCannotRoundtrip
+        Left  _                 -> error "unXPrvStripPub: this state cannot be \
+            \reached from a rightfully crafted XPrv"
+  where
+    -- Converts  xprv <> pub <> cc
+    -- To        xprv <>        cc
+    stripPub :: ByteString -> ByteString
+    stripPub xprv' = prv <> chainCode
+      where
+        (prv, rest) = BS.splitAt 64 xprv'
+        (_pub, chainCode) = BS.splitAt 32 rest
+
+data ErrXPrvFromStrippedPubXPrv
+    = ErrInputLengthMismatch Int Int -- ^ Expected, Actual
+    | ErrInternalError String
+    deriving (Eq, Show)
+
+-- | Create a @XPrv@ from a 96-byte long extended private key
+--
+-- The format is:
+--
+-- > Extended Private Key (64 bytes) <> ChainCode (32 bytes)
+xPrvFromStrippedPubXPrv :: ByteString -> Either ErrXPrvFromStrippedPubXPrv XPrv
+xPrvFromStrippedPubXPrv x = do
+        when (BS.length x /= expectedInputLength) $
+            Left $ ErrInputLengthMismatch expectedInputLength (BS.length x)
+        toXPrv $ CC.encryptedCreateDirectWithTweak x pass
+  where
+    pass :: ByteString
+    pass = ""
+
+    expectedInputLength = 96
+
+    -- @xprv@ can fail. But because it is calling @encryptedKey@ internally,
+    -- and we are feeding it the output of @unEncryptedKey@, it really shouldn't.
+    toXPrv :: CC.EncryptedKey -> Either ErrXPrvFromStrippedPubXPrv XPrv
+    toXPrv = left ErrInternalError . xprv . CC.unEncryptedKey
--- a/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
@@ -346,7 +346,7 @@ toChimericAccount =
 -- readability in function signatures.
 newtype Passphrase (purpose :: Symbol) = Passphrase ScrubbedBytes
     deriving stock (Eq, Show)
-    deriving newtype (Semigroup, Monoid, NFData)
+    deriving newtype (Semigroup, Monoid, NFData, ByteArrayAccess)
 
 type role Passphrase phantom
 
--- a/lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivationSpec.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivationSpec.hs
@@ -25,6 +25,7 @@ import Cardano.Wallet.Gen
 import Cardano.Wallet.Primitive.AddressDerivation
     ( Depth (..)
     , DerivationType (..)
+    , ErrUnXPrvStripPub (..)
     , ErrWrongPassphrase (..)
     , FromMnemonic (..)
     , FromMnemonicError (..)
@@ -42,6 +43,8 @@ import Cardano.Wallet.Primitive.AddressDerivation
     , checkPassphrase
     , encryptPassphrase
     , getIndex
+    , unXPrvStripPub
+    , xPrvFromStrippedPubXPrv
     )
 import Cardano.Wallet.Primitive.AddressDerivation.Byron
     ( ByronKey (..) )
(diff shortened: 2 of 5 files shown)


lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -722,3 +729,61 @@ hex = convertToBase Base16
...
+data ErrXPrvFromStrippedPubXPrv
+    = ErrInputLengthMismatch Int Int -- ^ Expected, Actual
+    | ErrInternalError String
+    deriving (Eq, Show)
+
+-- Create a @XPrv@ from a prv <> cc bytestring.
+xPrvFromStrippedPubXPrv :: ByteString -> Either ErrXPrvFromStrippedPubXPrv XPrv
+xPrvFromStrippedPubXPrv x =
+    let
+        res = toXPrv $ CC.encryptedCreateDirectWithTweak x pass
```
Anviking: Related thread https://github.com/input-output-hk/cardano-wallet/pull/1321#discussion_r382525139

With downstream WIP work, I managed to get all Byron root keys to roundtrip, but **not** all child keys. I'm looking at using some constructor _without_ tweak here.

Anviking: The failure seems related to child keys using `DerivationScheme1` more than it seems related to `Random` keys
```
cardano-wallet-jormungandr key root --wallet-style icarus express theme celery coral permit dizzy nothing stadium lemon tube olive brass dutch find mask
68f0cb3d83b5278f0b4c9c4a4ab50e49aef13f348ceafaf8257168fd8f3b894fa47646b6e206864404f3208b7dee1e71cd16096ac9205d9dd5250ae0e963dd79411337bda4d5bc0216462480b809824ffb48f17e08d95ab9f1b91d391e48e66b
cardano-wallet-jormungandr key child --path "0" 68f0cb3d83b5278f0b4c9c4a4ab50e49aef13f348ceafaf8257168fd8f3b894fa47646b6e206864404f3208b7dee1e71cd16096ac9205d9dd5250ae0e963dd79411337bda4d5bc0216462480b809824ffb48f17e08d95ab9f1b91d391e48e66b --wallet-style random | jcli key from-bytes --type ed25519bip32
invalid secret key
  |-> Invalid Secret Key structure
```

Anviking: With c07d738ea5a901b48724ba6d05daa8519ed0f1d7 we are using `CC.encryptedCreateDirectWithoutTweak`, and there is no problem.

stack.yaml
```diff
@@ -34,8 +34,8 @@ extra-deps:
     - persistent-template
 
 # cardano-crypto
-- git: https://github.com/input-output-hk/cardano-crypto
-  commit: 3c5db489c71a4d70ee43f5f9b979fcde3c797f2a
+- git: https://github.com/Anviking/cardano-crypto
+  commit: a6c3bfe12290cf4f6ac47836a1acb3d417da9d96
```
Anviking: https://github.com/input-output-hk/cardano-crypto/compare/master...Anviking:anviking/encryptedCreateDirectWithoutTweak?expand=0

^ didn't turn off stylish-haskell, but that's only a concern when PRing the main repo

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -722,3 +729,62 @@ hex = convertToBase Base16
 -- | Decode a 'ByteString' from base16
 fromHex :: ByteArray bout => ByteString -> Either String bout
 fromHex = convertFromBase Base16
+
+data ErrUnXPrvStripPub
+    = ErrNoRoundtripMismatch
```
Anviking: Seems like a double negative. Maybe just `ErrNoRoundtrip` instead.

Anviking:
bors try

stack.yaml
KtorZ: Alternatively, we could also just _forbid_ the use of keys using the derivation scheme #1 since it is known to have security issues.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -722,3 +729,62 @@ hex = convertToBase Base16
...
+
+-- Create a @XPrv@ from a prv <> cc bytestring.
+xPrvFromStrippedPubXPrv :: ByteString -> Either ErrXPrvFromStrippedPubXPrv XPrv
+xPrvFromStrippedPubXPrv x =
+    let
+        res = toXPrv $ CC.encryptedCreateDirectWithoutTweak prv cc pass
+    in
+        if BS.length x == expectedInputLength
+        then res
+        else Left $ ErrInputLengthMismatch expectedInputLength (BS.length x)
```
KtorZ: Style suggestion, this reads better using a do-notation + `when` 

```hs
when (BS.length x /= expectedInputLength) $
    Left $ ...
```

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -722,3 +729,62 @@ hex = convertToBase Base16
...
+  where
+    (prv, cc) = BS.splitAt 64 x
+    pass :: ByteString
+    pass = ""
+
+    expectedInputLength = 96
```
KtorZ, line 788: Why 96? 

(I know why, but another reader might not).

stack.yaml
Anviking: > could also just forbid the use of keys using the derivation scheme #1 since it is known to have security issues

Yes. I'm not sure if we'd want to do that 🤷‍♂. Taking your approval https://github.com/input-output-hk/cardano-wallet/pull/1321#pullrequestreview-364276474 as this way is fine for now though.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
KtorZ: `ErrNoRoundtrip` feels like something called "Roundtrip" is missing, like you would say "ErrNoPassphrase" for a missing field. 

Perhaps: `ErrCannotRoundtrip` or `ErrRoundtripFailure` ?

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -722,3 +729,62 @@ hex = convertToBase Base16
...
 fromHex :: ByteArray bout => ByteString -> Either String bout
 fromHex = convertFromBase Base16
+
+data ErrUnXPrvStripPub
+    = ErrNoRoundtripMismatch
+      -- ^ The resulting bytestring would have been unable to roundtrip using
+      -- @xPrvFromStrippedPubXPrv@. Most likely because the input @XPrv@ was
+      -- encrypted.
+    | ErrUnexpectedRoundtripError
+      -- ^ Something unexpected went wrong.
```
KtorZ: This second branch is quite dubious. There's nothing unexpected here, it feels more like an invariant violation. This branch is reached because of the check using `xPrvFromStrippedPubXPrv` if the given bytes are not of the right length. However, these bytes are produced by `unXPrvStripPub` itself, so it's more of a implementor error if we ever reach that point and shouldn't be pushed onto the function interface. 

So basically, `Left _ -> Left ErrUnexpectedRoundtripError` should actually really be:

```hs
    Left _ -> error "impossible: this state cannot be reached from a rightfully crafted XPrv". 
```

Or better, change the way the check is written to get rid of this impossible pattern match:

```hs
unXPrvStripPub k = do
    let res = stripPub . unXPrv $ k

    unless (Right res == roundtrip res) $ 
      Left ErrRoundtripFailed

    pure res
  where
    roundtrip = fmap unXPrv . xPrvFromStrippedPubXPrv
```

stack.yaml
KtorZ: > Taking your approval #1321 (review) as this way is fine for now though.

Hmmm. No. These are two separate things. I am approving a different PR, not this one :| ... 

So, let's drop "random" from the possible options in the command-line and simply, not expose it for now. Could you comment on the relevant JIRA ticket about this decision please?

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
Anviking: If `unXPrvStripPub` were to only be used from the CLI, the extra `ErrUnexpectedRoundtripError` would do no damage, and provide _slightly_ neater/controllable output… But sure — invariant might be more sane.

>  Or better, change the way the check is written to get rid of this impossible pattern match:

_When_ the impossible happens, I'd rather see a `error "impossible` than have it disguised as a `ErrRoundtripFailed`. Or conversely: if I see an `ErrRoundtripFailed` I don't want to doubt why it happened.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
Anviking: indeed! 💯

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
Anviking: I elaborated on the format in the doc-comments of `unXPrvStripPub` and `xPrvFromStrippedPubXPrv`, and mentioned "96 bytes" clearly. I won't add a comment here, specifically.

stack.yaml
Anviking: Done. I dropped the changes to cardano-crypto and extended tests. And commented on jira.

I could have added a "expect failure"-test case for roundtripping keys derived using `DerivationScheme1`, but I deemed it not worth it.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
Anviking: Also opting to use `case` instead of `unless` because I want to differentiate between `ErrRoundtripFailed` and `error "impossible` (and might as well `return res` instead of `return ()`, but could be debated)

Anviking:
Thanks for the review, @KtorZ! I believe I've addressed the remarks.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -720,3 +727,67 @@ hex = convertToBase Base16
...
+    -- Check that it roundtrips.
+    case (fmap unXPrv . xPrvFromStrippedPubXPrv $ res) of
+        Right bytes
+            | bytes == unXPrv k -> Right res
+            | otherwise         -> Left ErrCannotRoundtrip
+        Left  _                 -> error "unXPrvStripPub: this state cannot be \
+            \reached from a rightfully crafted XPrv"
+  where
+    -- Converts  xprv <> pub <> cc
+    -- To        xprv <>        cc
```
KtorZ, line 762: minor note here, `xprv` should really be `prv`. The`xprv` is so-to-speak `prv` + `cc`

Anviking: I changed this after reading your notes (was slightly wondering about it before), which use the term:

> 64 bytes: an extended Ed25519 secret key composed of:
>  - 32 bytes: Ed25519 curve scalar from which few bits have been tweaked (see below)
>  - 32 bytes: Ed25519 binary blob used as IV for signing

KtorZ (approved):

Anviking:
Merging as part of #1321 instead.

Issue #1353, opened 2020-02-17
Add `key child --path` command for CLI key derivation


Anviking:
# Context

[ADP-81](https://jira.iohk.io/browse/ADP-81)

The groundwork for CLI key derivation is being established in #1316.

# Decision

Implement command for deriving child keys, according to the example:
```bash
  $ key child --path 44'/1815'/0'/0 5bd164a28...1ae275939
  0d0a62eef...6f2aee621
```

# Acceptance Criteria

- We _must_ be able to derive keys using hard derivation.
- We _must_ support soft derivation on relevant paths.

---

# Development 

- [x] Added `key child` command  #1321 

# QA 

- Integration tests: https://github.com/input-output-hk/cardano-wallet/blob/9eb96075f841565bd7494eb815a44c3c01cf4a29/lib/cli/test/unit/Cardano/CLISpec.hs#L669-L702
- We have `ToText`/`FromText` roundtrips, and goldens for many special cases for `DerivationPath` and `DerivationIndex` in `CLISpec`
- Existing property `"scheme == mapKey (fromHex . toHex) scheme"` has been extended to confirm that schemes derive the same child keys when round-tripped.

KtorZ:
> We may (?) support soft derivation.

We must, on relevant paths.

Anviking:
@KtorZ 

How will we know whether to use DerivationScheme1 or DerivationScheme2?

It seems Byron / Random keys are using derivation scheme 1, but Icarus (and Shelley) is using scheme 2.

Do we have to take a `--type KEY-TYPE` cli option?

https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Icarus.hs#L325

https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Byron.hs#L326

KtorZ:
> It seems Byron / Random keys are using derivation scheme 1, but Icarus (and Shelley) is using scheme 2.

Yes.

> Do we have to take a --type KEY-TYPE cli option?

No, rather a `--byron` flag, if present, use derivation scheme 1, if not, use derivation scheme 2.

piotr-iohk:
lgtm.

PR #1370, opened 2020-02-24
Add cli `key child` command

Anviking:
# Issue Number

#1353 

# Overview

- [x] Preparatory addition of CLI `DerivationIndex` and `DerivationPath` types
- [x] Make `mapKey` xprv to hex transformation bidirectional
- [x] Add `key child --path` command to cli
- [x] Confirming that we can derive an actual wallet address key using `key child` as a manual golden test.

# Comments
- [ ] Not here: Parsing hexadecimal indexes e.g `0x8000073c` (could be neat 🤷‍♂ )

Manual example. Note that `to-pub` is not functionality of `cardano-wallet` — yet.
```bash
$ cardano-wallet-jormungandr key child --path "2147485500/2147485463/0'/0/0" 588102383ed9ecc5c44e1bfa18d1cf8ef19a7cf806a20bb4cbbe4e511666cf48d6fd7bec908e4c6ced5f0c4f0798b1b619d6b61e6110492b5ebb430f570488f074a9fc9a22f0a61b2ab9b1f1a990e3f8dd6fbed4ad474371095c74db3d9c743a | to-pub
eb806fca3bfda11ae14dee52c5ad6564e68e7531c1a1d498a76172898d2841cb3d9078e802011f1580465c80e7040f1e4d8e24f978d23f01c1d2cf18fcf741a7

# Address comes from an actual empty wallet:
$ jcli address info addr1sn4cqm728076zxhpfhh993ddv4jwdrn4x8q6r4yc5ash9zvd9pqukyh6fnkrsrsz4w9jvu8hyre20xm7m78jf4kww4h2n5maytuhtflrax9duz
discrimination: testing
public key: ed25519_pk1awqxlj3mlks34c2daefvttt9vnnguaf3cxsafx98v9egnrfgg89s4y7wfm
group key:  ed25519_pk1ztayempcpcp2hzexwrmjpu48ndldlrey6m882m4f6d7j97t45l3s5h5gl4
```

### DIFF
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -1370,10 +1370,11 @@ walletStyleOption = option (eitherReader fromTextS)
     ( long "wallet-style"
     <> metavar "WALLET_STYLE"
     <> helpDoc (Just (vsep typeOptions))
+    <> value Icarus -- NOTE: Remember to update help text when updating!
     )
   where
     typeOptions = string <$>
-        [ "Any of the following:"
+        [ "Any of the following (default: icarus)"
         , "  icarus (" ++ allowedWords Icarus ++ ")"
         , "  trezor (" ++ allowedWords Trezor ++ ")"
         , "  ledger (" ++ allowedWords Ledger ++ ")"
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -405,7 +405,7 @@ xPrvToTextTransform = (xPrvToHexText, hexTextToXPrv)
 
     hexTextToXPrv :: Text -> Either String XPrv
     hexTextToXPrv txt = do
-        bytes <- fromHex $ B8.pack $ T.unpack txt
+        bytes <- fromHex $ B8.pack $ T.unpack . T.strip $ txt
         left showErr $ xPrvFromStrippedPubXPrv bytes
       where
         showErr (ErrInputLengthMismatch expected actual) = mconcat
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -406,8 +406,7 @@ xPrvToTextTransform = (xPrvToHexText, hexTextToXPrv)
       where
         -- NOTE: This error should never happen from using the CLI.
         showErr ErrCannotRoundtripToSameXPrv =
-            "The private key I'm trying to encode looks wierd. \
-            \Is it encrypted? Or an old Byron key?"
+            "Internal error: Failed to safely encode an extended private key"
 
     hexTextToXPrv :: Text -> Either String XPrv
     hexTextToXPrv txt = do
@@ -422,8 +421,8 @@ xPrvToTextTransform = (xPrvToHexText, hexTextToXPrv)
             , " bytes."
             ]
         showErr (ErrCannotRoundtripToSameBytes) = mconcat
-            [ "Couldn't decode that extended private key while making sure it "
-            , "can be encoded back again. Is it an old Byron key?"
+            [ "That extended private key looks weird. "
+            , "Is it encrypted? Or is it an old Byron key?"
             ]
         fromHex = left (const "Invalid hex.")
             . convertFromBase Base16
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -1455,7 +1455,7 @@ walletStyleOption = option (eitherReader fromTextS)
          formatEnglishEnumerationRev xs = intercalate ", " (reverse xs)
 
 keyArgument :: Parser Text
-keyArgument = T.pack <$> (argument str (metavar "HEX-XPRV"))
+keyArgument = T.pack <$> (argument str (metavar "XPRV"))
 
 pathOption :: Parser DerivationPath
 pathOption = option (eitherReader fromTextS) $
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -490,15 +490,16 @@ instance FromText DerivationIndex where
 
         mkDerivationIndex :: Int -> Either TextDecodingError DerivationIndex
         mkDerivationIndex =
-            indexInv >=> (return . DerivationIndex . toEnum . fromEnum)
+            fmap (DerivationIndex . toEnum . fromEnum) . indexInv
 
-        parseSoftIndex =
-            parseWord >=> softIndexInv >=> mkDerivationIndex
+        parseSoftIndex txt = do
+            num <- parseWord txt >>= softIndexInv
+            mkDerivationIndex num
 
-        parseHardenedIndex =
-            parseWord
-            >=> (return . (+ fromIntegral firstHardenedIndex))
-            >=> mkDerivationIndex
+        parseHardenedIndex txt = do
+            num <- parseWord txt
+            let i = num + fromIntegral firstHardenedIndex
+            mkDerivationIndex i
 
         softIndexInv a = do
             idx <- mkDerivationIndex a
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -64,6 +64,8 @@ module Cardano.CLI
     , Port (..)
     , CliKeyScheme (..)
     , CliWalletStyle (..)
+    , DerivationIndex (..)
+    , DerivationPath (..)
 
     , newCliKeyScheme
     , xPrvToTextTransform
@@ -134,10 +136,13 @@ import Cardano.Wallet.Api.Types
 import Cardano.Wallet.Network
     ( ErrNetworkUnavailable (..) )
 import Cardano.Wallet.Primitive.AddressDerivation
-    ( ErrUnXPrvStripPub (..)
+    ( Depth (..)
+    , DerivationType (..)
+    , ErrUnXPrvStripPub (..)
     , ErrXPrvFromStrippedPubXPrv (..)
     , FromMnemonic (..)
     , FromMnemonicError (..)
+    , Index (..)
     , NatVals (..)
     , Passphrase (..)
     , PassphraseMaxLength
@@ -206,6 +211,8 @@ import Data.Text.Read
     ( decimal )
 import Data.Void
     ( Void )
+import Data.Word
+    ( Word32 )
 import Fmt
     ( Buildable, pretty )
 import GHC.Generics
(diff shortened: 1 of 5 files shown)


lib/cli/src/Cardano/CLI.hs
```diff
@@ -431,6 +447,64 @@ instance FromText CliWalletStyle where
...
+
+        mkDerivationIndex :: Int -> Either TextDecodingError DerivationIndex
+        mkDerivationIndex = inv >=> (return . DerivationIndex . toEnum . fromEnum)
+
+        parseSoftIndex =
+            parseWord >=> mkDerivationIndex
+
+        parseHardenedIndex =
+            parseWord
+            >=> (return . (+ 0x80000000))
```
Anviking: Should probably leverage the existing type
```Haskell
newtype Index (derivationType :: DerivationType) (level :: Depth) = Index
```
to avoid hard-coding `0x80000000` again.

Anviking: However, I'm not convinced that we need as much flexibility as the `Index` type provides.

For now I resorted to defining
```Haskell
firstHardenedIndex :: Word32
firstHardenedIndex = getIndex $ minBound @(Index 'Hardened 'AddressK)
```

lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -455,22 +458,63 @@ spec = do
...
-
     describe "Can perform roundtrip textual encoding & decoding" $ do
         textRoundtrip $ Proxy @(Port "test")
         textRoundtrip $ Proxy @MnemonicSize
         textRoundtrip $ Proxy @CliWalletStyle
+        textRoundtrip $ Proxy @DerivationIndex
+        textRoundtrip $ Proxy @DerivationPath
+
+    describe "DerivationPath/Index fromText goldens" $ do
+        let golden str desc expected = it (show str ++ " " ++ desc) $ fromText str `shouldBe` expected
```
Anviking: A few lines here are too long. Should look over.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -501,9 +594,33 @@ cmdRootKey =
...
+
+    exec (KeyChildArgs path style k0) = do
+        child <- foldM (deriveChildKey scheme) k0 path
+        TIO.putStrLn child
+      where
+        scheme :: CliKeyScheme Text IO
+        scheme =
+            hoistKeyScheme eitherToIO
+            . mapKey xPrvToTextTransform
+            $ newCliKeyScheme style
```
Anviking: Ok, I see 

> No, rather a --byron flag, if present, use derivation scheme 1, if not, use derivation scheme 2.

that we shouldn't take `--wallet-style` as an argument.

`CliKeyScheme` might need to be broken apart then. 🤔

Anviking:
@KtorZ A quick look, particularly related to the `DerivationIndex` would be appreciated! (https://github.com/input-output-hk/cardano-wallet/pull/1370#discussion_r384563730, 0774f7ee77b1a06e8aa0eb818415971b4ad00d15)

lib/cli/src/Cardano/CLI.hs
Anviking: I guess one _could_ strategically use a hard-coded `Icarus` scheme here and leave a todo for potential second U/S. And add three goldens for `key child` such that they don't end up diverging by accident.

Anviking: >  I guess one could strategically use a hard-coded `Icarus`

Done.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -378,44 +386,54 @@ hoistKeyScheme
...
-        ]
-    showErr (ErrInternalError msg) = mconcat
-        [ "Unexpected crypto error: "
-        , msg
-        ]
-    fromHex = left (const "Invalid hex.")
-        . convertFromBase Base16
+    hexTextToXPrv :: Text -> Either String XPrv
+    hexTextToXPrv txt = do
+        bytes <- fromHex $ B8.pack $ T.unpack . T.strip $ txt
```
paweljakubas: probably hlint will shout here because of `T.strip  txt`

lib/cli/src/Cardano/CLI.hs
```diff
@@ -378,44 +386,54 @@ hoistKeyScheme
...
     { allowedWordLengths = allowedWordLengths s
     , mnemonicToRootKey = eta . mnemonicToRootKey s
+    , deriveChildKey = \k i -> eta $ deriveChildKey s k i
     }
 
-xPrvToHexText :: XPrv -> Either String Text
-xPrvToHexText = left showErr . fmap (T.pack . B8.unpack . hex) . unXPrvStripPub
+-- | Pair of functions representing the bidirectional transformation between
+-- a @XPrv@ and its 96-byte hex-encoded form.
+xPrvToTextTransform :: (XPrv -> Either String Text, Text -> Either String XPrv)
```
paweljakubas: can't we just use `FromText` and `ToText` of `XPrv` (or probably some wrapped `XPrv` like we do in `Api.Types` using `ApiT` to avoid orphan instances) instances instead?

lib/cli/src/Cardano/CLI.hs
```diff
@@ -378,44 +386,54 @@ hoistKeyScheme
...
+xPrvToTextTransform = (xPrvToHexText, hexTextToXPrv)
   where
-    showErr ErrCannotRoundtrip =
-        "That private key looks weird. Is it encrypted? Or an old Byron key?"
+    xPrvToHexText :: XPrv -> Either String Text
+    xPrvToHexText = left showErr . fmap (T.pack . B8.unpack . hex) . unXPrvStripPub
+      where
+        showErr ErrCannotRoundtrip =
+            "That private key looks weird. Is it encrypted? Or an old Byron key?"
```
paweljakubas: no chance to say precisely here whether it is old Byron key or the other?

Anviking: Maybe.

The private part of a Byron key will be changed by the tweak. So both the private key and the public key will be different.

An encrypted `DerivationScheme2` key will roundtrip the private key correctly. But it will re-derive a different public key.

**However**, this error message is for the `XPrv -> Hex` conversion. It is only used to encode the resulting child key. It should actually never be called* 🙃 🤔 👀 

*) except if we were to re-introduce derivation using `DerivationScheme1` / Byron.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -431,6 +449,79 @@ instance FromText CliWalletStyle where
...
+        mkDerivationIndex :: Int -> Either TextDecodingError DerivationIndex
+        mkDerivationIndex =
+            indexInv >=> (return . DerivationIndex . toEnum . fromEnum)
+
+        parseSoftIndex =
+            parseWord >=> softIndexInv >=> mkDerivationIndex
+
+        parseHardenedIndex =
+            parseWord
+            >=> (return . (+ fromIntegral firstHardenedIndex))
```
rvl: I think do notation or a fmap would be clearer here.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1263,6 +1399,20 @@ walletStyleOption = option (eitherReader fromTextS)
...
+pathOption :: Parser [DerivationIndex]
+pathOption = option (eitherReader f) $
+    mempty
+    <> long "path"
+    <> metavar "DER-PATH"
+    <> help "Derivation path e.g. \"44'/1815'/0'/0\""
+    <> value ([DerivationIndex 0x80000000])
+  where
+    f :: String -> Either String [DerivationIndex]
+    f s = left getTextDecodingError $ mapM fromText $ T.splitOn "/" (T.pack s)
```
rvl: Is the empty derivation path handled anywhere?

This seems to overlap the functionality of `instance FromText DerivationPath`.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -431,6 +449,79 @@ instance FromText CliWalletStyle where
...
+
+firstHardenedIndex :: Word32
+firstHardenedIndex = getIndex $ minBound @(Index 'Hardened 'AddressK)
+
+instance FromText DerivationIndex where
+    fromText "" = Left $ TextDecodingError
+        "An empty string is not a derivation index!"
+    fromText x = do
+       -- NOTE: T.takeEnd will not throw, but may return "".
+       if T.takeEnd 1 x == "\'"
```
rvl: I put a comment on ADP-81 about this. I don't think using the `'` character will work very well in the most common shells (sh, bash, zsh).

rvl (approved):
Looks good

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1263,6 +1399,20 @@ walletStyleOption = option (eitherReader fromTextS)
...
+keyArgument :: Parser Text
+keyArgument = T.pack <$> (argument str (metavar "HEX-XPRV"))
+
+pathOption :: Parser [DerivationIndex]
+pathOption = option (eitherReader f) $
+    mempty
+    <> long "path"
+    <> metavar "DER-PATH"
+    <> help "Derivation path e.g. \"44'/1815'/0'/0\""
+    <> value ([DerivationIndex 0x80000000])
```
Anviking: I think this default `value` _definitely_ shouldn't be here.

lib/cli/src/Cardano/CLI.hs
Anviking: > This seems to overlap the functionality o

😮 Indeed, thanks!

> Is the empty derivation path handled anywhere?

It is currently handled somewhat implicitly by `T.splitOn "/"` returning `""`, which then results in the `DerivationIndex` failing with _An empty string is not a derivation index!_

Perhaps it could be desirable to allow empty paths (returning the input key), not sure. In-doubt I thought it would probably be a user error, and the CLI erroring would be desirable.

lib/cli/src/Cardano/CLI.hs
Anviking: > can't we just use FromText and ToText of XPrv (or probably some wrapped XPrv like we do in

We _could_ probably have gone that route.

I thought this way played nicely with
```Haskell
data CliKeyScheme key m = CliKeyScheme
    { allowedWordLengths :: [Int]
    , mnemonicToRootKey :: [Text] -> m key
    , deriveChildKey :: key -> DerivationIndex -> m key
    }
```
since it
- Encapsulated more logic in `CliKeyScheme` that could be tested easily. It enables writing the `scheme == mapKey (fromHex . toHex) scheme` property.*
- More easily allows multiple encodings (e.g. if wanted to add additional support for bech32 at some point)

*) Although I'm not clear on exactly what value this brings compared two normal `FromText`/`ToText` roundtrip properties. Might (just) be to ensure the roundtrips are performed with the all XPrvs that are used for the other tests, and not a subset.

Had we supported `Byron`, it would have made sure DerivationScheme1 child keys also round-tripped, without us having to add DerivationScheme1 child-keys to the `Arbitrary XPrv` generator (not an obvious thing to do).

lib/cli/src/Cardano/CLI.hs
Anviking: I think for instance the risk of turning `0'/0'` into `0/0` by accident makes this a valid concern. I changed to `H` as suggested by Matthias.

lib/cli/src/Cardano/CLI.hs
Anviking: I removed some `>=>` in a commit

lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -488,6 +511,40 @@ spec = do
...
+
+        fromTextGolden @DerivationIndex "4294967295H" "should fail" $
+            Left "6442450943 is too high to be a derivation index."
+
+        fromTextGolden @DerivationPath "" "should fail" $
+            Left "An empty string is not a derivation index!"
+
+        fromTextGolden @DerivationPath "44H/0H/0" "" $
+            Right . DerivationPath . map DerivationIndex $
+                [0x80000000 + 44, 0x80000000 + 0, 0]
```
paweljakubas: maybe we can give name to `0x8000000` - it is minBound of soft derivation index, right?

paweljakubas (approved):

lib/cli/test/unit/Cardano/CLISpec.hs
Anviking: It's the minBound of the **hardened** indexes. Can move to a named let-binding just here

(wanted to duplicate the definition at https://github.com/input-output-hk/cardano-wallet/blob/c0479c0258bd6a8e8fd00d5e708a5fbe2e3fcfba/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs#L230 on purpose though)

paweljakubas: *hardened* yes yes

Anviking:
thanks @paweljakubas @rvl !

bors r+

PR #1371, opened 2020-02-24
eventually with description for better identification when action inside eventually fails

piotr-iohk:
# Issue Number

# Overview

- 86ffd6b15f2cc505b4bac790b95e11638ac40c3b
  eventually with description for better identification when action inside eventually fails
  

# Comments

```
        eventually "Wallet balance = 10" $ do
            rb <- request @ApiWallet ctx
                (Link.getWallet @'Shelley wDest) Default Empty
            expectField
                (#balance . #getApiT . #available) (`shouldBe` Quantity 10) rb
```

in case of failure:
```
  src/Test/Integration/Scenario/API/Addresses.hs:153:5: 
  1) API Specifications ADDRESS_LIST_03 - Generates new address pool gap
       Worker has exited: main action is over.
       user error (waited more than 3min for action to eventually resolve. Action: "Wallet balance = 10")
```

### DIFF
--- a/lib/core-integration/src/Test/Integration/Framework/DSL.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/DSL.hs
@@ -78,8 +78,6 @@ module Test.Integration.Framework.DSL
     , utcIso8601ToText
     , eventually
     , eventuallyUsingDelay
-    , eventually_
-    , eventuallyUsingDelay_
     , fixturePassphrase
     , waitForNextEpoch
     , toQueryString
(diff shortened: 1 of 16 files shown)


piotr-iohk:
bors try

piotr-iohk:
bors try

piotr-iohk:
bors try

lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -511,31 +509,22 @@ eventually = eventuallyUsingDelay (500 * ms)
...
+            \ Action: " ++ show desc)
         Right a ->
             return a
   where
     trial = io `catch` \(_ :: SomeException) -> do
         threadDelay delay
         trial
 
-eventually_ :: IO () -> IO ()
-eventually_ = eventuallyUsingDelay_ (500 * ms)
```
KtorZ, line 529: What's the reason for getting rid of this one :thinking: ?

KtorZ (approved):
LGTM

lib/core-integration/src/Test/Integration/Framework/DSL.hs
piotr-iohk: It appears that you can use `eventually` like this:

```
eventually "xyz" $ do
  ...
```
or
```
r <- eventually "xyz" $ do
   ...
```
so I figured `eventually_` is not needed any longer.

PR #1373, opened 2020-02-25
SRE-49 - fix windows wallet tests in hydra

jbgi:
### Issue

[SRE-49](https://jira.iohk.io/browse/SRE-49)

### Summary

Fixes windows tests under wine by:
1. Updating the [wine version to 3.2.1](https://github.com/input-output-hk/haskell.nix/pull/472) and the [windows testWrapper](https://github.com/input-output-hk/haskell.nix/pull/471) so that we can run jormungandr under wine (allows integration tests to run under wine).
2. Fix utf-8 encoding issue in cli specs.
3. Fix errors found when deleting a wallet while its database is currently open.
4. In `DBFactory.removeDatabase`, only delete the file once there are no `withDatabase` actions in progress.

### Comments

- [x] [cardano-wallet-cli.unit](https://hydra.iohk.io/job/Cardano/cardano-wallet-pr-1373/x86_64-w64-mingw32.checks.cardano-wallet-cli.unit.x86_64-linux)
- [x] [cardano-wallet-core.unit](https://hydra.iohk.io/job/Cardano/cardano-wallet-pr-1373/x86_64-w64-mingw32.checks.cardano-wallet-core.unit.x86_64-linux)
- [ ] [cardano-wallet-jormungandr.integration](https://hydra.iohk.io/job/Cardano/cardano-wallet-pr-1373/x86_64-w64-mingw32.checks.cardano-wallet-jormungandr.integration.x86_64-linux)
- [ ] [cardano-wallet-launcher.unit](https://hydra.iohk.io/job/Cardano/cardano-wallet-pr-1373/x86_64-w64-mingw32.checks.cardano-wallet-launcher.unit.x86_64-linux) ⇒ #1425

### DIFF
--- a/lib/cli/test/unit/Cardano/CLISpec.hs
+++ b/lib/cli/test/unit/Cardano/CLISpec.hs
@@ -36,6 +36,8 @@ import Cardano.CLI
     , newCliKeyScheme
     , xPrvToTextTransform
     )
+import Cardano.Startup
+    ( setUtf8EncodingHandles )
 import Cardano.Wallet.Primitive.AddressDerivation
     ( NetworkDiscriminant (..), XPrv, unXPrv )
 import Cardano.Wallet.Primitive.Mnemonic
@@ -45,8 +47,6 @@ import Cardano.Wallet.Primitive.Mnemonic
     , entropyToMnemonic
     , mnemonicToText
     )
-import Cardano.Startup
-    ( setUtf8EncodingHandles )
 import Cardano.Wallet.Unsafe
     ( unsafeMkEntropy )
 import Control.Arrow
--- a/lib/cli/test/unit/Cardano/CLISpec.hs
+++ b/lib/cli/test/unit/Cardano/CLISpec.hs
@@ -45,6 +45,8 @@ import Cardano.Wallet.Primitive.Mnemonic
     , entropyToMnemonic
     , mnemonicToText
     )
+import Cardano.Startup
+    ( setUtf8EncodingHandles )
 import Cardano.Wallet.Unsafe
     ( unsafeMkEntropy )
 import Control.Arrow
@@ -127,14 +129,16 @@ spec = do
             <> cmdKey
 
 
-    let shouldStdOut args expected = it (unwords args) $
+    let shouldStdOut args expected = it (unwords args) $ do
+            setUtf8EncodingHandles
             case execParserPure defaultPrefs parser args of
                 Success x -> capture_ x >>= (`shouldBe` expected)
                 CompletionInvoked _ -> expectationFailure
                     "expected parser to show usage but it offered completion"
                 Failure failure ->
                     expectationFailure $ "parser failed with: " ++ show failure
-    let expectStdErr args expectation = it (unwords args) $
+    let expectStdErr args expectation = it (unwords args) $ do
+            setUtf8EncodingHandles
             case execParserPure defaultPrefs parser args of
                 Success x ->
                     hCapture_ [stderr] (try @SomeException x) >>= (expectation)
--- a/lib/core/src/Cardano/DB/Sqlite/Delete.hs
+++ b/lib/core/src/Cardano/DB/Sqlite/Delete.hs
@@ -1,18 +1,27 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE TupleSections #-}
 
 -- |
 -- Copyright: © 2018-2020 IOHK
 -- License: Apache-2.0
 --
--- A function to delete a SQLite database file, which isn't as straightforward
--- as it sounds.
+-- A function to wait until a suitable time to delete a SQLite database file,
+-- and a function to delete a SQLite database file, which isn't as
+-- straightforward as it sounds.
 
 module Cardano.DB.Sqlite.Delete
-    ( deleteSqliteDatabase
+    ( -- * Removing files with retry
+      deleteSqliteDatabase
     , DeleteSqliteDatabaseLog (..)
+    -- * Ref-counting open databases
+    , RefCount
+    , newRefCount
+    , withRef
+    , waitForFree
     ) where
 
 import Prelude
@@ -21,8 +30,22 @@ import Cardano.BM.Data.Severity
     ( Severity (..) )
 import Cardano.BM.Data.Tracer
     ( DefinePrivacyAnnotation (..), DefineSeverity (..) )
+import Control.Concurrent.MVar
+    ( MVar, modifyMVar, modifyMVar_, newMVar, readMVar )
+import Control.Exception
+    ( bracket_ )
+import Control.Retry
+    ( capDelay, fibonacciBackoff, limitRetriesByCumulativeDelay, retrying )
+import Control.Tracer
+    ( Tracer, traceWith )
 import Data.Aeson
     ( ToJSON )
+import Data.Function
+    ( (&) )
+import Data.Map.Strict
+    ( Map )
+import Data.Maybe
+    ( fromMaybe, isJust )
 import Data.Text.Class
     ( ToText (..) )
 import GHC.Generics
@@ -39,16 +62,17 @@ import Control.Retry
     , recovering
     , retryPolicy
     )
-import Control.Tracer
-    ( Tracer, traceWith )
 import System.IO.Error
     ( isPermissionError )
 #else
-import Control.Tracer
-    ( Tracer )
 #endif
 
+import qualified Data.Map.Strict as Map
 import qualified Data.Text as T
+
+{-------------------------------------------------------------------------------
+                           Removing files with retry
+-------------------------------------------------------------------------------}
 
 -- | Remove a SQLite database file.
 --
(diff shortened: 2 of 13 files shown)


lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -127,17 +128,21 @@ spec = do
...
             case execParserPure defaultPrefs parser args of
                 Success x -> capture_ x >>= (`shouldBe` expected)
                 CompletionInvoked _ -> expectationFailure
                     "expected parser to show usage but it offered completion"
                 Failure failure ->
                     expectationFailure $ "parser failed with: " ++ show failure
-    let expectStdErr args expectation = it (unwords args) $
+    let expectStdErr args expectation = it (unwords args) $ do
+            hSetEncoding IO.stdout utf8
+            hSetEncoding IO.stderr utf8
```
KtorZ: @jbgi you may want to use: https://input-output-hk.github.io/cardano-wallet/haddock/edge/cardano-wallet-launcher-2020.3.3/Cardano-Startup.html#v:withUtf8Encoding (for a cross-platform solution to this). We've ran into this a couple of times already :(

jbgi: ah nice!

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -281,6 +298,36 @@ newDBFactory tr defaultFieldValues mDatabaseDir = do
...
+    -- handle to the file is closed. Unfortunately there are a lot of
+    -- system services that can have a file temporarily opened using a shared
+    -- read-only lock, such as the built in AV and search indexer.
+    --
+    -- We can't really guarantee that these are all off, so what we can do is
+    -- whenever after an rm the file still exists to try again and wait a bit.
+    retryOnPermissionErrors 5 . removePathForcibly
+#else
+    removePathForcibly
+#endif
```
KtorZ: We'd need to check if there's a way to check / make sure that the SQLite has been properly closed before removing the file. In practice, we always:

a. Kill the thread that has the connection open (which, because of bracketting-style resource allocation should free-up the connection at some point)

b. Only remove the database file when we know that the connection has ended. 

On Linux, it seems that either the connection is released fast enough or, that it doesn't matter if the connection is still open when the file is deleted. On Windows, that's a different story indeed. 

One possible "easy way" to work around this I see would be to have an MVar shared between the `withDatabase` and `removeDatabase` such that, `removeDatabase` would only attempt to remove the file once the connection has been terminated on `withDatabase`; so waiting on an MVar before `mapM_ removePathForcibly ...` with a protective time-out. 

cc @rvl

lib/cli/test/unit/Cardano/CLISpec.hs
jbgi: `withUtf8Encoding` did not have any effect on the test failing (presumably due to `withCP65001` logic), had to use `setUtf8EncodingHandles` directly.

jbgi:
`haskell.nix` branch used in this PR is https://github.com/input-output-hk/haskell.nix/tree/hkm/cardano-wallet-fix augmented with https://github.com/input-output-hk/haskell.nix/pull/471 and https://github.com/input-output-hk/haskell.nix/pull/472

nix/sources.json
```diff
@@ -1,26 +1,26 @@
 {
     "haskell.nix": {
-        "branch": "master",
+        "branch": "jbgi/cardano-wallet-fix-windows",
```
rvl: This needs to be master branch, so your PRs on Haskell.nix need to be merged first.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
rvl: @jbgi This is interesting. What references do you have for these claims?

1. "On Windows ... a FileDelete command just marks a file for deletion. The file is really only removed when the last handle to the file is closed."
2. "there are a lot of system services that can have a file temporarily opened using a shared read-only lock, such as the built in AV and search indexer."

For 1, I thought that was more like the behaviour on Linux (although on Linux, the directory entry is immediately removed, but the inode stays). Do you mean that the filename lingers in the directory listing untll it has actually been deleted?

For 2, do these services _really_ cause files to not be deleted promptly? I thought that would break lots of things that expect FileDelete to do the obvious.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -281,6 +296,36 @@ newDBFactory tr defaultFieldValues mDatabaseDir = do
...
+retryOnPermissionErrors :: Int -> IO a -> IO a
+retryOnPermissionErrors maxRetries io = do
+    res <- try io
+    case res of
+        Right a -> return a
+        Left ex | isPermissionError ex && maxRetries > 1 -> do
+                    let retries' = traceShow (ex, maxRetries) (maxRetries - 1)
+                    threadDelay ((maxRetries - retries') * 200)
+                    retryOnPermissionErrors retries' io
+                | otherwise -> traceShow ("OTHER ERROR" :: String) (throw ex)
```
rvl: If we end up using the retry approach: You can log with the iohk-monitoring tracer here, rather than Debug.Trace.traceShow. Also some debug logging on retries would help.

rvl (commented):
It might be better to implement some code using MVars which deletes the sqlite db file after the last connection has closed, as @KtorZ suggests.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
jbgi: main reference: https://github.com/haskell/directory/issues/96
I originally implemented this "fix" in order to try getting a test pass on wine, but in the end, it was not necessary for the test to pass, or at least not enough, though I still think this hack is useful from what I have read.

nix/sources.json
jbgi: agreed, though cardano-wallet master is currently not using haskell.nix master but https://github.com/input-output-hk/haskell.nix/tree/hkm/cardano-wallet-fix

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
jbgi: Yes, though is this falling outside my haskell proficiency atm :(.
There is debug logging on retries at line 323:
```haskell
 let retries' = traceShow (ex, maxRetries) (maxRetries - 1)
```

rvl: OK - I will convert the logging to use Control.Tracer.

nix/sources.json
rvl: cardano-wallet master is on haskell.nix master I'm pretty sure.

jbgi: It is not, see https://input-output-rnd.slack.com/archives/CG386Q1NE/p1582245051365500?thread_ts=1582215136.357100&cid=CG386Q1NE for context. (maybe the issue has been solved on master since then, cc/ @hamishmack)

rvl:
I added a test case and reworked `internalRemovePathForcibly`, but haven't yet added `withDatabase` refcounting to `newDBFactory`.

rvl:
I have updated Haskell.nix to the latest master with @jbgi's two PRs merged, added refcounting to DBFactory, and added unit tests.

rvl (approved):
Remaining two test failures can be fixed in a subsequent PR.

lib/core/src/Cardano/DB/Sqlite/Delete.hs
```diff
@@ -0,0 +1,237 @@
...
+
+{-------------------------------------------------------------------------------
+                          Ref-counting open databases
+-------------------------------------------------------------------------------}
+
+-- | Mutable variable containing reference counts to IDs of type @ix@.
+data RefCount ix = RefCount
+    { _refCount :: MVar (Map ix Int) -- ^ number of references to each index
+    , _takeLock :: MVar () -- ^ lock on incrementing references
+    }
```
KtorZ, line 173: I don't think we do have more than one `withDatabase` call per wallet, so there should be always at most a single connection per database :thinking:. 
If we can do more, we can also do less so I guess it's fine.

KtorZ (approved):
LGTM. Haven't tried locally but looks solid and quite convinced unit tests.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -258,28 +260,33 @@ newDBFactory tr defaultFieldValues mDatabaseDir = do
...
-                        [ databaseFile wid
-                        , databaseFile wid <> "-wal"
-                        , databaseFile wid <> "-shm"
-                        ]
-                mapM_ removePathForcibly files
+                let widp = pretty wid
+                -- try to wait for all 'withDatabase' calls to finish before
+                -- deleting database file.
+                let trWait = contramap (MsgWaitingForDatabase widp) tr
+                waitForFree trWait refs wid $ \inUse -> do
```
Anviking, line 277: Why do we need the reference counting now, but didn't previously?

Only using `deleteSqliteDatabase`, without `waitForFee`, wouldn't achieve the same thing? (which is, I guess, to make sure the file is really gone when `removeDatabase` returns)

Some more wandering questions from trying to understand:

Is `withDatabase` is called once by the restoration worker? Then I guess `waitForFree` makes sure that the wallet is not in the registry. _And_ it looks like `destroyDBLayer` makes sure to close the sqlite connection before it is removed from the registry.

So we could never cancel a sqlite query, but theoretically an API request could acquire a `SqliteContext` when it is open, wait, and when it tries to use it it has been closed?

Anviking (approved):

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
KtorZ: > Why do we need the reference counting now, but didn't previously?

We actually always needed it. It has never worked properly on Windows :)

KtorZ:
failures from bors:

-       1) Cardano.DB.Sqlite.Delete.RefCount waitForFree waits for withRef to finish
-       2) Cardano.Wallet.DB.Sqlite.DBFactory removeDatabase waits for connections to close

I am going to look into these.

rvl:
Oh drat, both test cases have the same problem. I think the `waitForFree` side of `concurrently` needs a small delay so that `withRef` happens first.

lib/core/src/Cardano/DB/Sqlite/Delete.hs
rvl: I thought that might be the case, but am hoping it will be more resilient if it supports more than one `withDatabase` call per wallet.

nix/sources.json
```diff
@@ -5,22 +5,22 @@
...
+        "rev": "0933c58908816bca525fbf8873abdbc3e072a87e",
+        "sha256": "1shgaaqj0j2d5f7967xssp69nn6dv34dfvkvqkiakb99ggycgdyg",
         "type": "tarball",
-        "url": "https://github.com/input-output-hk/haskell.nix/archive/2d1a60c0c7ff268e3cf6eea13b4f24420fc711d2.tar.gz",
+        "url": "https://github.com/input-output-hk/haskell.nix/archive/0933c58908816bca525fbf8873abdbc3e072a87e.tar.gz",
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     },
     "iohk-nix": {
-        "branch": "master",
+        "branch": "nixpkgs-common-channels",
```
jbgi: need https://github.com/input-output-hk/iohk-nix/pull/342 to be merged (or rollback this change).

rvl: Thanks

rvl:
Fixed test failures and rebased.

bors try

KtorZ:
(timed out somehow).

rvl:
Rebased on latest master.

bors try

KtorZ:
What's the matter with this one :thinking: 
I've tried running it locally yesterday and it went fine.

rvl:
It passed when I ran it locally today as well. But this branch consistently fails on CI. I need to look in the log...

rvl:
On the last build, two stake pools integration tests failed. But these were red herrings.

I looked more carefully at the output from `stack test`. and it's the `cardano-wallet-jormungandr` *unit* tests which were still running when it timed out.

Since these exact same tests pass in the non-bors builds, I wonder that whether 30 minutes is not long enough to run the full test suite? Unfortunately the Buildkite logs aren't timestamped, so ...

bors r+

Issue #1386, opened 2020-02-26
Translate integration tests for active/next delegation into unit tests.


piotr-iohk:
# Context

When reviewing and testing https://github.com/input-output-hk/cardano-wallet/issues/1342 we have created 12 integration tests (tagged `STAKE_POOL_NEXT`) that fairly comprehensively test the requirement. They have been quite long to execute adding ~9 minutes of execution to the overal integration test suite exec time and as a result made CI to timeout.

The idea here is to translate all or of some of them into unit tests to cut on test execution time and still have reliable tests.

There is an open PR with the integration tests -> https://github.com/input-output-hk/cardano-wallet/pull/1374. It can be used to translate the tests.

# Decision

Translate `STAKE_POOL_NEXT` tests from https://github.com/input-output-hk/cardano-wallet/pull/1374 into unit tests.

# Acceptance Criteria

---

# Development 

https://github.com/input-output-hk/cardano-wallet/pull/1392

# QA 

Unit tests and properties added which let us get rid of 2/3 of lengthy integration tests saving ~6min of CI execution. See also description: https://github.com/input-output-hk/cardano-wallet/pull/1392

KtorZ:
@piotr-iohk @paweljakubas I've factored out the logic responsible for controlling the delegation transitions as `guardJoin` and `guardQuit`. 

From these two pure functions, we can now easily tests various delegation scenarios given some initial `WalletDelegation` state, and control that the outcome is as expected. I didn't think much through it but I believe there also exists some nice properties to try out. 

See: https://github.com/input-output-hk/cardano-wallet/commit/718f931ef53846a325b816a09169adb927b987be on `KtorZ/1386/unit-tests-delegation-transitions`. Feel free to add anything you want on this branch :+1:

PR #1374, opened 2020-02-25
Next delegation integration tests => next delegation unit tests

piotr-iohk:
# Issue Number

Follow up to #1361 

# Overview

- afeee24084c3c417e1ee285a8ba3196c4e15da55
  I have reorganized a bit integration tests for "stake pool active/next" groupping them into 5 scenarios:
  - STAKE_POOL_NEXT_01 - Can join/re-join another/quit stake pool
  - STAKE_POOL_NEXT_02 - Join 2 sp in the same epoch => delegating to the 
last one in the end
  - STAKE_POOL_NEXT_02 - Join and quit in the same epoch => 
not_delegating in the end
  - STAKE_POOL_NEXT_03 - Join 2 in two subsequent epochs => delegating to 
1st in epoch X + 2 and 2nd in epoch X + 3
  - STAKE_POOL_NEXT_03 - Join and quit in two subsequent epochs => 
delegating in epoch X + 2, not_delegating in epoch X + 3

- 3241e1ff51020fccf07e5c23bd5f260d7b1bf8bc
  Further tests for join/quit active/next

# Comments

Additional integration tests for #1342 after changes made in #1361. The tests are somewhat comprehensive but at the same time they take ~9minutes, which made CI to timeout before...

The idea would be to turn some of them into unit tests...

PR #1392, opened 2020-02-27
Unit tests delegation transition

piotr-iohk:
# Issue Number

#1386

# Overview

- 6902ac2bfae0e85991e27549b072006644e388f9
  expose guardQuit and guardJoin for unit testing
  
- 7c6761b0adbc8796a33f813debb4262af926d514
  guardJoin - check if pool is known at first
  
- 3e0b82430f37edee61e91bbe89fc6fa48bcaf51f
  Integration tests for next/active delegation
  
- f617302842774bc87bcf6559ec68cd4306f11447
  guardJoin, guardQuit properties and unit tests

# Comments

I have cherry-picked integration tests from #1374 on top of 718f931ef53846a325b816a09169adb927b987be. 

Integration tests that I kept.
```
STAKE_POOL_NEXT_01 - Can join/re-join another/quit stake pool
STAKE_POOL_NEXT_02/STAKE_POOLS_QUIT_01 - Cannot quit when active: not_delegating
STAKE_POOL_NEXT_02 - Override join with join in the same epoch => delegating to the last one in the end
STAKE_POOL_NEXT_03 - Join 2 in two subsequent epochs => delegating to 1st in epoch X + 2 and 2nd in epoch X + 3
```

removed, covered by unit tests and properties
```
TO_REMOVE - STAKE_POOL_NEXT_02 - Override join with quit in the same epoch => not_delegating in the end
TO_REMOVE - STAKE_POOL_NEXT_02 - Override quit with join in the same epoch => delegating to the last one in the end
TO_REMOVE - STAKE_POOL_NEXT_03 - Join and quit in two subsequent epochs => delegating in epoch X + 2, not_delegating in epoch X + 3
TO_REMOVE - STAKE_POOL_NEXT_04 - Cannot quit when most recent delegation next: not_delegating
TO_REMOVE - STAKE_POOL_NEXT_04 - Cannot quit when 'next 1': not_delegating, 'next 2': not_delegating
TO_REMOVE - STAKE_POOL_NEXT_05/STAKE_POOLS_JOIN_01 - Cannot join A, when active: A
TO_REMOVE - STAKE_POOL_NEXT_05/STAKE_POOLS_JOIN_01 - Cannot join A, when 'next 1': A and there is no 'next 2'
TO_REMOVE - STAKE_POOL_NEXT_05/STAKE_POOLS_JOIN_01 - Cannot join A, when 'active' : A, 'next 1': B and 'next 2': A
```

### DIFF
--- a/lib/core/test/unit/Cardano/WalletSpec.hs
+++ b/lib/core/test/unit/Cardano/WalletSpec.hs
@@ -66,6 +66,7 @@ import Cardano.Wallet.Primitive.Types
     , Direction (..)
     , EpochNo (..)
     , Hash (..)
+    , PoolId (..)
     , SealedTx (..)
     , SlotId (..)
     , SlotNo (..)
@@ -78,6 +79,9 @@ import Cardano.Wallet.Primitive.Types
     , TxMeta (..)
     , TxOut (..)
     , TxStatus (..)
+    , WalletDelegation (..)
+    , WalletDelegationNext (..)
+    , WalletDelegationStatus (..)
     , WalletId (..)
     , WalletMetadata (..)
     , WalletName (..)
@@ -140,13 +144,18 @@ import Test.QuickCheck
     , arbitrarySizedBoundedIntegral
     , choose
     , elements
+    , label
+    , oneof
     , property
     , scale
     , shrinkIntegral
     , vectorOf
     , withMaxSuccess
+    , (===)
     , (==>)
     )
+import Test.QuickCheck.Arbitrary.Generic
+    ( genericArbitrary, genericShrink )
 import Test.QuickCheck.Monadic
     ( monadicIO )
 import Test.Utils.Time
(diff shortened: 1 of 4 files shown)


piotr-iohk:
bors try

piotr-iohk:
In the end I have left 4 integration tests:
```
STAKE_POOL_NEXT_01 - Can join/re-join another/quit stake pool
STAKE_POOL_NEXT_02/STAKE_POOLS_QUIT_01 - Cannot quit when active: not_delegating
STAKE_POOL_NEXT_02 - Override join with join in the same epoch => delegating to the last one in the end
STAKE_POOL_NEXT_03 - Join 2 in two subsequent epochs => delegating to 1st in epoch X + 2 and 2nd in epoch X + 3
```

The rest should be covered by unit tests and two properties: https://github.com/input-output-hk/cardano-wallet/pull/1392/commits/f617302842774bc87bcf6559ec68cd4306f11447

paweljakubas (approved):
LGTM! I am very fond of moving the checking to unit level. I hope it is going to spark initiative to move also other parts that reside now on integration test

piotr-iohk:
> I hope it is going to spark initiative to move also other parts that reside now on integration test

I think that whenever it is possible and makes sense we should do it. That also requires writing code in a way that it is feasible to test reliably and exhaustively on unit level. I think that ideally integration tests should not be testing many specific cases for particular requirement but just a few e2e scenarios to make sure that integration of all components works fine.

lib/core/test/unit/Cardano/WalletSpec.hs
```diff
@@ -202,10 +211,93 @@ spec = do
...
+        Left W.ErrAlreadyDelegating{} ->
+            label "ErrAlreadyDelegating" (W.guardQuit dlg === Right ())
+
+prop_guardQuitJoin
+    :: [PoolId]
+    -> WalletDelegation
+    -> PoolId
+    -> Property
+prop_guardQuitJoin knownPools dlg pid =
+    pid `elem` knownPools ==> case W.guardQuit dlg of
```
KtorZ, line 294: It can be _"dangerous"_ to use a pre-condition for filtering generated data. It can lead to false-negative where QC will give-up if it has to discard too many cases (and will assume that the pre-condition can't be met). 

A better practice would be to generate the right data in the first place. Here, the property doesn't really have any assertions the `pid` or the list of pools. So, instead of generating them separately, one could generate a `NonEmpty PoolId` and pick a pool id from it (and it's guaranteed to be in the list :grimacing:)!

Issue #1389, opened 2020-02-26
Add `key public` and `key inspect` cli commands


Anviking:
# Context

https://jira.iohk.io/browse/ADP-81

We need a command to derive public keys from private keys.

We also need an `inspect` command that splits the 32-byte chain code  from the 64-byte extended private key. It should also do the corresponding thing when inspecting public keys.

The CLI should also _ultimately_ support soft derivation using public keys. This may need some thought to do properly.

# Decision

Implement quick, simple versions of `key inspect` and `key public`. 

Don't bother with adding support for inspecting public keys. That would be best handled later when we add public key derivation support in another U/S.

# Acceptance Criteria

- We must implement `key public`
- We must implement `key inspect`
- The commands should conform to the usage:
```bash
$ key public 5bd164a28...1ae275939
  10bf8a5a8...fbab14ea9

$ key inspect 5bd164a28...1ae275939
  private-key: ...
  chain-code : ...
```

---

# Development 

- Quick implementation of `key public` #1393
- Quick implementation of `key inspect` #1397

# QA 

- Added one golden test each for `public` and `inspect`.
- Negative scenarios (relating to decoding xprvs) should be covered by the tests for `key child`.

Anviking:
With the current `CliKeyScheme`, it might be as well to implement public key derivation at the same time.

We could also probably be "fine" with a `Text -> m Text` implementation, and a golden test or two. There really shouldn't be any corner cases with converting a `XPub` to hex `Text`. (Unlike the intricacies of going from hex to `XPrv`).

piotr-iohk:
lgtm.

PR #1397, opened 2020-02-27
Add quick `key inspect` cli command

Anviking:
# Issue Number

#1389

# Overview

- [x] I implemented a quick `key inspect` cli command that extracts the `extended private key (64 bytes)`, and `chain code (32 bytes)`, from an `extended private key (96 bytes)`

# Comments

- `key inspect` does currently **not** support inspecting public keys.


paweljakubas (approved):
looks fine

Issue #1331, opened 2020-02-06
0.5% profit margin of pool is rounded to 0%


Anviking:
# Context

The pool margin/fee is a rational number on-chain.

It is converted to a `Percentage`-type here: https://github.com/input-output-hk/cardano-wallet/blob/fcde98eeb8e895db0a770be7cab17995881b9ed2/lib/jormungandr/src/Cardano/Wallet/Jormungandr/Binary.hs#L1082-L1083

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      |  |
| Platform     |           |
| Installation |             |

# Steps to Reproduce

1. List stake pools via CLI or Daedalus
2. Lookup the margin of the `PNT5` pool.

## Expected behavior

1. It should have a `0.5%` profit margin. (Have not confirmed on chain, but sounds very plausible)

## Actual behavior

1. Profit margin is displayed as `0%`.
 

---

# Resolution 

- #1399 Make `Percentage` type more precise

---

# QA 

- Tested manually with Daedalus that PNT5 has `0.5%` margin, and that migration works.
- Golden `ToText`/`ToJSON` tests https://github.com/input-output-hk/cardano-wallet/blob/7482427ff79946124d7142dc4784909c877023d4/lib/core/test/unit/Data/QuantitySpec.hs#L79-L89

piotr-iohk:
lgtm!

PR #1399, opened 2020-02-28
Rework use  of `Percentage` type

Anviking:
# Issue Number

#1331 

# Overview

- [x] Increase precision of `Percentage` from `[0,100]` natural to `Rational`.
- [x] Changed sqlite schema to store the pool margin as `(Word64, Word64)` instead of `Word8`.

# Comments

- We can make the `ToText` instance _always_ have the same number of digits after the decimal separator (e.g. `1.45%`, `2.00%`). But we can't do this for `ToJSON`, as Aeson uses `Scientific` and not `String` to represent numbers.

Skärmavbild 2020-03-05 kl  15 06 51

migration worked, and pool margin looks correct 🎉


Anviking:
bors try

paweljakubas (approved):
LGTM!

paweljakubas:
I wonder if this change will not force some changes in swagger.yaml. we have there *percentage* reference. maybe we should investigate this

Anviking:
> I wonder if this change will not force some changes in swagger.yaml. we have there percentage reference. maybe we should investigate this

IIRC the swagger specifies percentage as a _number_ (floating or integer), and it should be fine. But we'll see

paweljakubas:
right, `stack test --ta '-m "Cardano.Wallet.Api.Types"' cardano-wallet-core:unit` passes?

Anviking:
> right, stack test --ta '-m "Cardano.Wallet.Api.Types"' cardano-wallet-core:unit passes?

Yup.

rvl (commented):
Double doesn't seem like the right type. It has both too much and too little precision, if you get what I mean...

Could we delete the `Percentage` type and replace it with `Rational`, then add a function to format these as a percentage string?

Anviking:
> Could we delete the Percentage type and replace it with Rational, then add a function to format these as a percentage string?

Yes, but we'd basically need a function `format :: Scientific -> Scientific` because we need to give `Aeson` a `Scientific` and not a `String`.

lib/core/src/Data/Quantity.hs
```diff
@@ -134,49 +135,65 @@ instance (KnownSymbol unit, Buildable a) => Buildable (Quantity unit a) where
...
-    toEnum = either (error . ("toEnum: " <>) . show) id . mkPercentage
-
 instance ToText Percentage where
-    toText (Percentage p) = T.pack (show p) <> "%"
+    toText (Percentage p) = T.pack (showS p) <> "%"
+      where
+        showS =
+            formatScientific
+            Fixed
+            (Just percentageNumberOfFractionalDigits)
```
KtorZ, line 174: Please indent this so that it is obvious which identifiers are arguments and which are function names! Having them all at the same level is confusing.

lib/core/src/Data/Quantity.hs
```diff
@@ -134,49 +135,65 @@ instance (KnownSymbol unit, Buildable a) => Buildable (Quantity unit a) where
...
+        in
+            pure . Percentage . round' . toPercent $ r
+  where
+    round' :: Double -> Scientific
+    round' x = (fromRational (toRational i)) / (fromRational (toRational factor))
+      where
+        i :: Int
+        i = round (factor * (toRational x))
+
+        factor = 10 ^ percentageNumberOfFractionalDigits
```
KtorZ: This function has several issues here:

a) It forces `Double` as an argument, which pretty much looses all the benefits of using a type like Scientific or Ratio underneath.

b) It mixes formatting with the construction of the type by already rounding the argument. 

Why not choose `Rational` for the inner type of the `Percentage` type and do the formatting / cast to 2-decimal scientific only at the API boundaries?

Anviking: > which pretty much looses all the benefits of using a type like Scientific or Ratio underneath.

Firstly: What benefits? Any potential errors would be hidden by the rounding?

Secondly: Things get more complicated with `Rationals`.

```haskell
        ActiveSlotCoefficient f = sp ^. #getActiveSlotCoefficient
        remainingBlocks = remainingSlots * f
        progress = bhTip % (bhTip + remainingBlocks)
```
This is essentially `a/(b+c) where, a:Word, b:Word, c:Double`.

What do we do?
- Do the calculation as a `Double` and _then_ convert to `Rational`? Seems iffy to me.
- Round `remainingBlocks`? I guess this will make the sync progress less smooth.

Anviking: > Why not choose Rational for the inner type of the Percentage type

Maybe we could use `Scientific` 🤔 (in `mkPercentage`)

Downside: I think we'd need to add `scientific` as a dependency in a lot of .cabal files.

KtorZ: > Firstly: What benefits? Any potential errors would be hidden by the rounding?

PRECISION! Double are not good at representing rational numbers. Yet, percentage are almost always computed as a ratio of two numbers, hence why `Rational` is a better fit here.

> Secondly: Things get more complicated with Rationals.

This sounds actually much simpler to me... We can safely round `remainingBlocks` to an integer actually, because it's a number of block! What does it mean to have 3.4 remaining blocks anyway? Flooring or Ceiling doesn't really matter here because this is a progress reporting, not something that need more careful thinking. 

> Round remainingBlocks? I guess this will make the sync progress less smooth.

We are talking about 4M blocks for mainnet and ~2.5M for the ITN. Really, a 1 block error margin is fine.

Anviking: > This sounds actually much simpler to me

I'm happy there's an approach you like 🙏 

Still, for the sake of it:

> Double are not good at representing rational numbers. 

Seems like a vague and general argument in these circumstances.

> Yet, percentage are **almost always** computed as a ratio of two numbers, hence why Rational is a better fit here.

In this case we had an instance of `(a/(b+(c/d)))` (basically) as you saw. 

- Using `round` might introduce an error of `2e-6` (on [0,1] sync progress for ITN).*
- We are displaying `Percentage` with a precision of `1e-4` (i.e. 100.00%=1.0000). 
- The error we get from a `Double` must surely be smaller than **even** 1e-14?

The 2e-6 should _probably_** not matter. But 1e-14 should matter even less.

> What does it mean to have 3.4 remaining blocks anyway?

It _might_ be useful to make the sync progress smoother for small values of `f`, in some circumstances. If f=1%, should the progress only start decreasing once we are 100 slots behind the current time? 🤷‍♂ 

*)
According to pooltool.io the block height is ≈273000.
If we use `round` the average error should be 0.5.
The error in syncProgress is then 0.5/273000 = 2e-6

**) But I'm at the same time not convinced that it will not matter under some circumstances

_modulo minor math notation / terminology mistakes. Might be mixing up error and precision a bit_

Anviking:
bors try

lib/core/src/Data/Quantity.hs
```diff
@@ -134,49 +136,63 @@ instance (KnownSymbol unit, Buildable a) => Buildable (Quantity unit a) where
...
-    fromEnum (Percentage p) = fromEnum p
-    toEnum = either (error . ("toEnum: " <>) . show) id . mkPercentage
-
 instance ToText Percentage where
-    toText (Percentage p) = T.pack (show p) <> "%"
+    toText (Percentage p) = T.pack (showS p) <> "%"
+      where
+        showS = formatScientific
+            Fixed
+            (Just percentageNumberOfFractionalDigits)
```
Anviking, line 174: @KtorZ 
> Why not choose Rational for the inner type of the Percentage type and do the formatting / cast to 2-decimal scientific only at the API boundaries

My thinking was that it would be safer to concentrate the rounding to the only constructor of `Percentage` (I removed the `Enum` instance also). I.e not have to remember to round in each `ToText`, `ToJSON` instance, etc.

But I guess I already use `formatScientific`, and cold as just as well move the `round' :: Scientific -> Scientific` function from `mkPercentage` to `toJSON`…

lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -380,16 +382,16 @@ newStakePoolLayer tr block0H getEpCst db@DBLayer{..} nl metadataDir = StakePoolL
     computeProgress
         :: BlockHeader -- ^ ... / denominator
         -> BlockHeader -- ^ numerator /...
-        -> Quantity "percent" Percentage
+        -> Ratio Word64
```
Anviking: I guess I should keep returning a `Quantity "percent" Percentage` here by using `unsafeMkPercentage`

lib/core/test/unit/Cardano/Wallet/Primitive/TypesSpec.hs
```diff
@@ -318,25 +318,34 @@ spec = do
...
-                    , (mockBlockHeader (SlotId 0 5) 3, 60)
-                    , (mockBlockHeader (SlotId 0 6) 3, 60)
-                    , (mockBlockHeader (SlotId 0 7) 4, 66)
-                    , (mockBlockHeader (SlotId 0 8) 5, 83)
-                    , (mockBlockHeader (SlotId 0 9) 5, 100)
-                    , (mockBlockHeader (SlotId 1 0) 5, 100)
+                    [ (mockBlockHeader (SlotId 0 0) 0, 0.0)
+                    , (mockBlockHeader (SlotId 0 1) 1, 0.2)
+                    , (mockBlockHeader (SlotId 0 2) 2, 0.3333)
+                    , (mockBlockHeader (SlotId 0 3) 2, 0.3333)
```
Anviking: Here we see the effect of rounding `remainingBlocks`!

These should be the numbers from the un-rounded `remainingBlocks`:
```
[ (mockBlockHeader (SlotId 0 0) 0, 0.0)
, (mockBlockHeader (SlotId 0 1) 1, 0.1818)
, (mockBlockHeader (SlotId 0 2) 2, 0.3333)
, (mockBlockHeader (SlotId 0 3) 2, 0.3636)
, (mockBlockHeader (SlotId 0 4) 2, 0.40)
, (mockBlockHeader (SlotId 0 5) 3, 0.5455)
, (mockBlockHeader (SlotId 0 6) 3, 0.60)
, (mockBlockHeader (SlotId 0 7) 4, 0.7273)
, (mockBlockHeader (SlotId 0 8) 5, 0.8333)
, (mockBlockHeader (SlotId 0 9) 5, 0.9091)
, (mockBlockHeader (SlotId 1 0) 5, 1.00)
```
Notice 0.3333 and 0.3636 vs 0.3333 and 0.3333

_Although, why was it not 33% and 36% previously then?_ 🤔

lib/core/src/Data/Quantity.hs
KtorZ: > Seems like a vague and general argument in these circumstances.

Not really........

> In computing, floating-point arithmetic (FP) is arithmetic using formulaic representation of real numbers as an approximation to support a trade-off between range and precision. 

source: https://en.wikipedia.org/wiki/Floating-point_arithmetic

I mean, that's simple enough as an explanation. Floating point arithmetic is a vast topic in computer science but the basic idea is easy to grasp: floating numbers are an approximation. 

Now, what I am saying is not that the approximation in this particular case is not okay (as you just outlined), but that if we can avoid this approximation by actually using a type that enables us to represent values as non-approximated, we are better off using that approach. This actually saves us a lot of headache and hours trying to calculate and demonstrate how rounding this particular value is okay. Using a Rational, we only have to worry about the moment we need to convert it to a float and chose a given precision to work with. Instead of carrying around an approximation error where it's not needed.

lib/core/src/Data/Quantity.hs
Anviking: Or… do we _really_ have to?

rvl (approved):
LGTM

lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -365,7 +369,7 @@ newStakePoolLayer tr block0H getEpCst db@DBLayer{..} nl metadataDir = StakePoolL
                         Ranking.desirability epConstants $ Ranking.Pool
                             (unsafeMkRatio 0) -- pool leader pledge
                             poolCost
-                            (unsafeMkRatio $ fromIntegral (getPercentage poolMargin) / 100)
+                            (unsafeMkRatio . fromRational . toRational $ poolMargin)
```
KtorZ: What type has `poolMargin` here? It seems to already be a ratio, that we convert 3 times and back into a ratio :thinking: ?

Anviking: Pool margin is a `Ratio Word64`.

unsafeMkRatio is a `Double -> Cardano.Pool.Ranking.Ratio` 😬

lib/core/src/Data/Quantity.hs
KtorZ: I think you missed the point of my comment here... From the first comment: 

> Why not choose Rational for **the inner type of the Percentage type** [...]

Now, we have a Rational as an argument that we convert to double and convert back to Scientific!

lib/core/src/Cardano/Pool/Metrics.hs
Anviking: Note that this is for the desirability calculation, and not to display the pool margin.

lib/core/src/Data/Quantity.hs
Anviking: Done.

lib/core/test/unit/Cardano/Pool/MetricsSpec.hs
```diff
@@ -429,7 +433,7 @@ instance Arbitrary PoolRegistrationCertificate where
     arbitrary = PoolRegistrationCertificate
         <$> arbitrary
         <*> fmap (L.nub . getNonEmpty) (scale (`mod` 3) arbitrary)
-        <*> fmap toEnum (choose (0, 100))
+        <*> ((1 %) <$> choose (1, 10000))
```
Anviking: I think this instance is fine, particularly since it seems unimportant to the tests in this module.

lib/core/test/unit/Data/QuantitySpec.hs
```diff
@@ -58,17 +55,29 @@ spec = do
...
                     ]
             forM_ tests $ \test -> it (T.unpack test) $ do
                 fromText @Percentage test `shouldBe` Left err
 
+        describe "ToText & ToJSON goldens (checking rounding)" $ do
+            let tests =
+                    [ (unsafeMkPercentage $ 1 % 3, "33.33%", "33.33")
+                    , (unsafeMkPercentage 1, "100.00%", "100")
+                    , (unsafeMkPercentage 0, "0.00%", "0")
+                    ]
```
Anviking, line 89: Added goldens

lib/core/src/Cardano/Pool/Metrics.hs
KtorZ: Why having a `Ratio` type here that has exactly the same purpose as the `Percentage` type :thinking: ?

lib/core/src/Data/Quantity.hs
```diff
@@ -134,49 +136,82 @@ instance (KnownSymbol unit, Buildable a) => Buildable (Quantity unit a) where
...
         Left PercentageOutOfBoundsError
-    | let Percentage sup = maxBound in i > fromIntegral sup =
+    | r > 1 =
         Left PercentageOutOfBoundsError
     | otherwise =
-        pure $ Percentage $ fromIntegral i
+        let
+            toPercent = (* 100)
+        in
+            pure . Percentage . toPercent $ r
```
KtorZ: Why the multiplication here?
Representing percentage as values between 0 and 100 is formatting concern, so it should only happens within the `toText` or `toJSON` instance, it's quite error prone to store a 100x the actual value. 

Or is this to cope with the `Bounded` instance maybe?

lib/core/test/unit/Cardano/Wallet/Primitive/TypesSpec.hs
KtorZ: Indeed, on a chain of length ~5, which already gives satisfying results regarding progress tracking. Don't you think?

Anviking:
@KtorZ Ah! I think I see what you're getting at. Will try using `Percentage` more 👍 (instead of treating it as a Api-like type)

KtorZ:
> (instead of treating it as a Api-like type)

:pray: This is precisely why it's part of `Data.Quantity` and not part of the API types. This is really just a wrapper around `Rational` that are bounded between [0,1] (although one could argue that a percentage could be greater than 1 ... So `Percentage` may not be the best name here)

paweljakubas:
Q: using http://hackage.haskell.org/package/base-4.12.0.0/docs/src/Data.Fixed.html#Centi or even https://hackage.haskell.org/package/fixed-point-0.5.0.1/docs/Data-Fixed-Binary.html is not ok for our purposes here? If yes, why?

Anviking:
> is not ok for our purposes here? If yes, why?

The direction that this PR now took was: make `Percentage` match the precision of on-chain types `Ratio Word64`, and use it everywhere. So we shouldn't want to use `Centi` as internal type of `Percentage`… but as a shortcut for presentation, indeed, maybe! 🤔 

If we want to keep `0` rendering as `0.00%`, we might want to keep the `formatScientific` in `ToText` though. (Makes the percentage values more constant width in the logs)

_actually, not sure it simplifies the instances_

lib/core/test/unit/Data/QuantitySpec.hs
```diff
@@ -98,7 +123,8 @@ instance Arbitrary a => Arbitrary (Quantity u a) where
...
 instance Arbitrary Percentage where
-    shrink p
-        | p == minBound = []
-        | otherwise = [minBound]
-    arbitrary = arbitraryBoundedEnum
+    shrink _ = []
+    -- Note: We are only generating values that pass roundtrips here!
+    --
+    -- 0.01% is the precision we use for presentation.
+    arbitrary = unsafeMkPercentage . (% 10000) <$> choose (0, 10000)
```
Anviking, line 139: With this specific generator the roundtrips pass

lib/jormungandr/src/Cardano/Wallet/Jormungandr/Binary.hs
```diff
@@ -1079,8 +1078,7 @@ overrideFeePolicy linearFee@(LinearFee a b _) override =
 poolRegistrationsFromBlock :: Block -> [W.PoolRegistrationCertificate]
 poolRegistrationsFromBlock (Block _hdr fragments) = do
     PoolRegistration (poolId, owners, taxes, _tx) <- fragments
-    let margin = fromRight maxBound $ mkPercentage @Int $ round $
-            100 * ratio (taxRatio taxes)
+    let margin = unsafeMkPercentage $ taxRatio taxes
```
Anviking: Reverting to using `fromRight maxBound` to be safe.

KtorZ (approved):

Anviking:
A slight con with replacing `Pool.Ranking.Ratio = Ratio Double` with `Percentage = Percentage Rational` is that test failures in `RankingSpec` are more tricky to read.

It was very easy to get overflow using `Ratio Word64` in `RankingSpec` just using `(>)`.

So using `Rational` (`Ratio Integer`) instead. (Could have used `Ratio Natural` also, but that is less convenient 🤷‍♂ )

bors r+

PR #1400, opened 2020-02-28
HW wallets tests

piotr-iohk:
# Issue Number

#1355

# Overview

- c8a33b989375d26d1b0897b92a609d2d091299a9
  Move HW wallets integration tests into separate module
  
- c0dc735d77839d910cd25bd006d3b71523f13c98
  Move waitAllTxsInLedger and getSlotParams to DSL.hs
  
- f2f8f96d4750888700bae945b45f530c3763a151
  HW wallets integration tests
  
- 399e9280513343711b34e554891ee55912406969
  return 403 when operation on HW wallet requires private key
  
- 9197fbd769c69811dc285cfa33a9de1c79032ff5
  add tests for malformed account_public_key and make error message more informative.
  
- 2d096cfb9376eebd9761e17c36d9d350b45b8f62
  add test that the same account and mnemonic wallet can live side-by-side
  
- 4e1eb0ec59929d49be0cdc30a6d35ce4e7f0d883
  small fix to Wallet.hs tests
  
- d45edd7d15ddc745dc25dc058b77999bc78d0c37
  refactor param names in getSlotParams

# Comments
- integration tests `HW_WALLETS` added
```
Finished in 13.3691 seconds
16 examples, 0 failures
```
 - ~still to test -> HW_WALLETS_06 https://github.com/input-output-hk/cardano-wallet/issues/1355#issuecomment-591430246~ -> done in 9197fbd769c69811dc285cfa33a9de1c79032ff5

### DIFF
--- a/lib/core-integration/src/Test/Integration/Framework/DSL.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/DSL.hs
@@ -80,6 +80,7 @@ module Test.Integration.Framework.DSL
     , eventuallyUsingDelay
     , fixturePassphrase
     , waitForNextEpoch
+    , waitAllTxsInLedger
     , toQueryString
     , withMethod
     , withPathParam
@@ -88,6 +89,7 @@ module Test.Integration.Framework.DSL
     , mkEpochInfo
     , notDelegating
     , delegating
+    , getSlotParams
 
     -- * CLI
     , command
@@ -135,17 +137,21 @@ import Cardano.Wallet.Primitive.AddressDerivation
 import Cardano.Wallet.Primitive.Mnemonic
     ( entropyToMnemonic, genEntropy, mnemonicToText )
 import Cardano.Wallet.Primitive.Types
-    ( Address (..)
+    ( ActiveSlotCoefficient (..)
+    , Address (..)
     , Coin (..)
+    , EpochLength (..)
     , EpochNo
     , Hash (..)
     , HistogramBar (..)
     , PoolId (..)
-    , SlotParameters
+    , SlotLength (..)
+    , SlotParameters (..)
     , SortOrder (..)
     , StartTime (..)
     , TxIn (..)
     , TxOut (..)
+    , TxStatus (..)
     , UTxO (..)
     , UTxOStatistics (..)
     , WalletId (..)
@@ -230,7 +236,7 @@ import System.Process
 import Test.Hspec
     ( Expectation, HasCallStack, expectationFailure )
 import Test.Hspec.Expectations.Lifted
-    ( shouldBe, shouldContain )
+    ( shouldBe, shouldContain, shouldSatisfy )
 import Test.Integration.Faucet
     ( nextTxBuilder, nextWallet )
 import Test.Integration.Framework.Request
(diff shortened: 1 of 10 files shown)


lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1808,7 +1808,7 @@ instance LiftHandler ErrCreateWallet where
 instance LiftHandler ErrWithRootKey where
     handler = \case
         ErrWithRootKeyNoRootKey wid ->
-            apiError err404 NoRootKey $ mconcat
+            apiError err410 NoRootKey $ mconcat
```
piotr-iohk: I have updated this to `410` because there was some inconsistency. 
 - trying to send tx or join SP on HW wallet resutled in `410` and this error message
- updating password on HW wallet resulted in `404` and the same error message

Now all operations on HW wallet (requiring private key) result with `410`.
(On the other hand, I was thinking that if this code should be rather `401` or `403`, but maybe `410` fits as well...)

piotr-iohk:
bors try

lib/core/src/Cardano/Wallet/Api/Server.hs
KtorZ: I'd rather keep 401 for anything that is authentication related (if we end up having any form of authentication on the API, which is unlikely). 401 makes only sense in the presence of an `Authorization` header. 

403 would make sense here indeed. I haven't any strong preferences nevertheless 410 somewhat supposed that the root key was at some point available and removed later which doesn't really reflect what's happening here. I think the rationale was that in the past, the scenario were no root keys was found could only occur during the 3-step of the transaction creation: selection/signing/submission. On the second step, a the key is needed to sign a transaction and we do not have database atomicity between these steps, so _in theory_, the wallet could disappear between the "selection" and "signing" steps, resulting in an `NoRootKey` error, and therefore, leading to a 410 Gone. 

Now, returning 410 when updating passphrase doesn't really fit IMO, because the resource (root key) was never there in the first place. This was also the rationale behind 404. 

So, between:

a. Adding more logic to distinguish the cases where the key was there and gone because of a concurrent requests (410) and the case where it was never there to being with (404).

b. A more "generic" error code that fits both cases (403). 

I'd rather go for b. :man_shrugging:

lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -1191,6 +1215,23 @@ for = flip map
 --
 -- Helper for delegation statuses
 --
+getSlotParams
+    :: (Context t)
+    -> IO (EpochNo, SlotParameters)
+getSlotParams ctx = do
+    r2 <- request @ApiNetworkInformation ctx
```
paweljakubas: why `r2` ;-) and not `r1`

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: yep, option b seems reasonable and less work  ;-)

lib/core-integration/src/Test/Integration/Framework/DSL.hs
piotr-iohk: I think that you actually wrote this, I just moved it to different place ;)

lib/core/src/Cardano/Wallet/Api/Server.hs
piotr-iohk: done in 399e928 - return 403 when operation on HW wallet requires private key

lib/core-integration/src/Test/Integration/Framework/DSL.hs
piotr-iohk: done in d45edd7 - refactor param names in getSlotParams

piotr-iohk:
bors try

paweljakubas (approved):
looks fine

PR #1402, opened 2020-03-02
nix: Fix build caching

rvl:
### Issue number

None

### Overview

I noticed in input-output-hk/cardano-launcher#15 that the cache was not being used for nix builds, because source filtering and naming was removed in #1351.

This puts back and improves source filtering, and adds a few source comments to the nix code.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,3 +1,42 @@
+############################################################################
+#
+# Cardano Wallet Nix build
+#
+# Derivation attributes of this file can be build with "nix-build -A ..."
+# Discover attribute names using tab-completion in your shell.
+#
+# Interesting top-level attributes:
+#
+#   - cardano-wallet-jormungandr - cli executable
+#   - cardano-wallet-byron - cli executable
+#   - tests - attrset of test-suite executables
+#     - cardano-wallet-core.unit
+#     - cardano-wallet-jormungandr.integration
+#     - etc (layout is PACKAGE.COMPONENT)
+#   - checks - attrset of test-suite results
+#     - cardano-wallet-core.unit
+#     - cardano-wallet-jormungandr.integration
+#     - etc
+#   - benchmarks - attret of benchmark executables
+#     - cardano-wallet-core.db
+#     - cardano-wallet-jormungandr.latency
+#     - etc
+#   - migration-tests - tests db migrations from previous versions
+#   - dockerImage - tarball of the docker image
+#   - shell - imported by shell.nix
+#   - haskellPackages - a Haskell.nix package set of all packages and their dependencies
+#     - cardano-wallet-core.components.library
+#     - etc (layout is PACKAGE-NAME.components.COMPONENT-TYPE.COMPONENT-NAME)
+#
+# The attributes of this file are imported by the Hydra jobset and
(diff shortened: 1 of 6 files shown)


nix/haskell.nix
```diff
@@ -8,6 +8,7 @@
 , buildPackages
 , config ? {}
 # GHC attribute name
+## fixme: what is this for?
```
rvl: It looks like this variable is unused.

default.nix
```diff
@@ -1,27 +1,25 @@
...
-# Import IOHK common nix lib
-, sourcesOverride ? {}
-# Use pinned Nixpkgs with Haskell.nix overlay
-, pkgs ? import ./nix { inherit system crossSystem config sourcesOverride; }
+# Import pinned Nixpkgs with iohk-nix and Haskell.nix overlays
+, pkgs ? import ./nix/default.nix { inherit system crossSystem config sourcesOverride; }
 # Use this git revision for stamping executables
 , gitrev ? pkgs.commonLib.commitIdFromGitRepoOrZero ./.git
+# Use this to reference local sources rather than the pinned versions (see nix/default.nix)
+, sourcesOverride ? {}
```
rvl, line 48: How do developers use this parameter for their builds? What is the format of the attrset?

nix/haskell.nix
angerman: The `compiler` attribute should be removed. The compiler is supposed to be inferred fro either the stack.yaml or the cabal.project file. *if* we want to override it, we might want to provide `null` here and only use the override value if non-null.

Forcing `ghc-8.6.5` here will inevitably lead to issues if the stack.yaml (or project.cabal) specify a different compiler.

rvl: OK thanks, removed.

default.nix
manveru: The format is the same as the one returned from the `sources.nix`. This is documented nowhere it seems, so I'll attempt to do so here.

The `sourcesOverride` should be an attrset with names being one of the source names in `nix/sources.json`. In cardano-wallet these are at the moment: `haskell.nix` `iohk-nix` `niv` `nixpkgs-19.03`.
The value then can be a simple path to a local checkout of those repos, for example if you want to quickly experiment with some changes without having to make commits and updating the sources.json all the time.

Example usage:

    nix build -f . stackNixRegenerate --arg sourcesOverride '{ nixpkgs = ~/github/nixos/nixpkgs; iohk-nix = ~/github/input-output-hk/iohk-nix; }'

The values don't have to be paths, they can also be simply the result of a fetchTarball, fetchGit, fetchurl, fetchFromGitHub, or whatever else produces paths once evaluated.

manveru (approved):
LGTM

default.nix
jbgi: The documentation has not been copied over indeed, though most other project do document the parameter, including the skeleton project:
https://github.com/input-output-hk/iohk-nix/blob/master/skeleton/default.nix#L12-L18

rvl: Thanks for the info @manveru @jbgi. I have added comments accordingly.

PR #1352, opened 2020-02-17
bump versions to v2020-02-17

KtorZ:
# Issue Number

N/A

# Overview

- [ ] I have bumped versions to v2020-02-17 in preparation of Today's release

# Comments

PR #1405, opened 2020-03-03
Bump package versions to 2020.3.3

Anviking:
# Overview

```bash
find . -name "*.nix" ! -path "*.stack-work*" | xargs sed -i "s/$OLD_VERSION/$NEW_VERSION/"
# or macOS version:
find . -name "*.nix" ! -path "*.stack-work*" | xargs sed -i "" "s/$OLD_VERSION/$NEW_VERSION/"
```

# Comments

### DIFF
--- a/lib/cli/cardano-wallet-cli.cabal
+++ b/lib/cli/cardano-wallet-cli.cabal
@@ -1,5 +1,5 @@
 name:                cardano-wallet-cli
-version:             2020.2.17
+version:             2020.3.3
 synopsis:            Utilities for a building Command-Line Interfaces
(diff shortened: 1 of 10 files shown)


piotr-iohk:
Looks like there should be few more :thinking: (comparing to -> https://github.com/input-output-hk/cardano-wallet/pull/1352/files)

Anviking:
@piotr-iohk yeah, I notice the command does not change nix files. I wonder if the nix-regenerate CI check will give a patch changing all remaining instances of `2020.2.17`… otherwise I guess the command should be tweaked

piotr-iohk (approved):
:+1:

Issue #1398, opened 2020-02-28
Wallet restoration extremely slow on Windows


piotr-iohk:
# Context
 | Information  | -                                                     |
| ---          | ---                                                   |
| Version      | 2020.2.17, master, `2020.2.17 (git revision: ede977b563a548d7f36d99332cfaa0835b6533e5)` |
| Platform     | Windows          |
| Installation | From source / github release           |

cardano-wallet on Windows has always been slower than on Linux. However recently it has been often reported (by @alan-mcnicholas and others) that on Windows platform wallet restoration has been slower than before on ITN or QA jormungandr testnet.

Some tests that I have done on QA testnet:

Creating a wallet on Windows and syncing 0-100%. Jormungandr node synced.
```
cardano-wallet master      / jormungandr 8.12   =  ~12 min
cardano-wallet v2020-02-17 / jormungandr 8.9    =  ~12 min
cardano-wallet v2020-01-27 / jormungandr 8.7    =  ~24 min
```
For comparison, on Linux and the same QA testnet wallet creation and sync 0-100% takes ~1min.

It has been reported that on ITN testnet wallet restoration took sometimes as much as **50mins** on Windows with Linux time = ~1-2 mins.

# Steps to Reproduce

 - On Windows. Connect to ITN or QA testnet.
```
cardano-wallet-jormungandr launch --genesis-block-hash 8e4d2a343f3dcf9330ad9035b3e8d168e6728904262f2c434a4f8f934ec7b676 --node-port 8080 --state-dir ./data_incentivized -- --config node_config_incentivized.yaml
```
- Create a wallet and wait until it is synced 100%

```
curl  -vX POST http://localhost:8090/v2/wallets \
  -H "Content-Type: application/json; charset=utf-8" \
  -d '{
"name": "Wallet1",
"mnemonic_sentence": ["identify", "screen", "lock", "bargain", "inch", "drop", "canyon", "flock", "dry", "zone", "wash", "argue", "system", "glory", "light"],
"passphrase": "Secure Passphrase",
"address_pool_gap": 20
}' --http1.1  |  jq
```
## Expected behavior

Syncing time on both platform should be same/similar.

## Actual behavior
 
Windows is very slow.

---

# Resolution 

---

# QA

rvl:
I have not noticed this because my Windows 10 Home install inside VirtualBox runs very slowly just on its own, even before I start cardano-wallet-jormungandr.

[...]

PR #1407, opened 2020-03-03
Integration Setup Foundation for cardano-node w/ Byron

KtorZ:
# Issue Number

#1347 

# Overview

- 91b6d63e6845b0ac193b93aaa576d4d846aab9d3
  move and rename 'sockAddrPort' to 'unsafePortNumber'
  a. Allow sharing the code outside of the jormungandr package
b. A bit more true about the underlying intent of the function
c. The function had so-to-speak nothing to do with Cardano.Faucet

- 72aad48ffccbf195431c78f767c0f791e7c66c0c
  remove 'nodePort' and 'feePolicy' from 'Context'
  this type is common to various integration setup. However, these constituants are
only used in Jormungandr-specific scenario and do not apply to cardano-node

- beb6032685c132c51a66b5f286ffdb74138877cc
  Remove hard-coded byron parameters from compat functions
  Passing them as arguments instead. Also, currently abusing the
'Testnet' discriminant with integration data. Later, I'll make sure
that serveWallet takes blockchain parameters as arguments instead of
a 'KnownNetwork =>' constraints to allow constructing private blockchain
from arbitrary genesis blocks.

- 9650acec6e362e37123e816bb50e5c8bd22477b5
  remove 'staticBlockchainParameters' from NetworkLayer
  a. Blockchain parameters aren't actually static and comes from genesis + protocol updates.
b. Allow more flexibility for cardano-node in the way the genesis block is passed
c. Remove awkward conversion in the byron compat code for cardano node

- b9bb6d86e91ea4a594729d793df99e04aea4b14d
  Implement 'fromGenesisTxOut' to construct transaction from genesis UTxO
  These transaction are a bit special and by convention, the transaction id is a hash of the genesis address.

- fb193b4146ef54ea7c3a8c7b252442d3aff44905
  Fix wrong checkpoint reference in unstable assertion
    While doing ... something totally different, I noticed that some integration tests, when restoring a wallet were showing something like 1000x "Creating checkpoint at xxxxxxxx-[X.X#X]."

  This caught my attention for a second because:

  a. With the setting I had, it should only create around 10 checkpoints, not a thousand.
  b. This happened not only near the tip, but at any time.

  Inspecting the code responsible for inserting checkpoints, I found this in `Cardano.Wallet#restoreBlocks`

  ```hs
          forM_ (NE.init cps) $ \cp' -> do
              let (Quantity h) = currentTip cp ^. #blockHeight
              when (fromIntegral h `elem` unstable) $ do
                  liftIO $ logCheckpoint cp'
                  putCheckpoint (PrimaryKey wid) cp'
  ```

  Note the `currentTip cp` instead of `currentTip cp'` ! So, the assertion of whether or not a checkpoint should be inserted was based on the current checkpoint from the database, and not on the actual checkpoint. So basically, every time the database checkpoint is considered a stable block, we get to insert ~1000 checkpoints in the database man_facepalming ... That can take some time...

  I am not sure whether this is the cause of the Windows slowness mentioned in #1398 (and why this doesn't really impact Linux then?) but it is definitely wrong.

# Comments

This is only a foundation for setting up the integration tests (still not committed here). I've got the integration setup working and running, though I still need to go through a few fixes to avoid hard-coded some of the genesis parameters and instead, read them from the genesis file.


KtorZ:
bors try

paweljakubas (approved):
Amazed that setting `mainnetGenesis` and `blockchainParameters` plus small changes to Jormungandr.Network and extending argument list in several places was enough to make it work :trophy: 

Q: I see that commit of ouroborous-network is month old in stack.yaml. We are ok with that in the context of this PR ? (I bet bumping it is going to be work for another PR and it will affect considerably byron codebase, but maybe also integration testing layer?). 

Also weeder complaining.

KtorZ:
>  I see that commit of ouroborous-network is month old in stack.yaml. We are ok with that in the context of this PR ?

Yes. Locally, I am running a 1.6 cardano-node and the client code is using the libraries from 1.5. It works fine. I'll bump the dependencies at the end.

KtorZ:
3x in a row the same error. But I can't reproduce locally. This is weird.

PR #1413, opened 2020-03-05
Fix threads synchronization in around all

KtorZ:
# Issue Number

N/A

# Overview

- 62d717d068f29f580371f24ae04a1e03529af2f7
  Fix threads synchronization in around all

  The issue was that, the main thread was not waiting for the 'resource thread' to be done and to
have cleaned up the resource. In our integration setup, the resource actually spawns an
underlying process! So, not waiting for the resource to be cleared means leaving a process
running even after the tests have exited! So, I added another point of synchronization between
the threads to make sure that the main thread would only exits AFTER the resource has been released.

# Comments


KtorZ:
bors try

paweljakubas (approved):
LGTM! Figure helps with comprehending the steps and lifetime of threads engaged

rvl:
Good find. :+1:

PR #1414, opened 2020-03-05
cardano-node OBFT - test integration setup 

KtorZ:
# Issue Number

#1346 

# Overview

- 25fe5815f66de7a909a915dbcc2aa1dd517d4f68
  make byron's 'newNetworkLayer' use bracket style allocation
  So that the underlying connection to the node gets cleaned-up when the network layer is destroyed.
Also, make sure to clean up past clients when creating new cursors

- b2b466064e1d0eaa5e47716faa77d1ad88ae8b9a
  Retry connecting to client when the socket does not exist
  
- fc51990aa95d43d6476e0012adb343cb7e6c6414
  move byron integration scenarios requiring a shelley wallet to Scenario/API/Wallets
  
- 230ac845bcb62c116d511652b7ca7ad04e04e563
  implement 'getCurrentNodeTip' for Byron's network layer
  
- bf65a99611e863175d1b220f2eec077205f3e1bf
  Move faucets declarations to core-integration for sharing
  
- 455aba4264a5cfa8cd72cfc314479fbd339d2118
  add basic integration setup for cardano-node OBFT
  
- 99de24a5697fb9f04038ec536c732a9c5bbde308
  fixup Byron network layer implementation.
  
- b32000c67b71cdc6852898a946893c9c13264fb6
  remove hard-coded blockchain parameters from integration scenarios
  The parameters are actually different for each network, in particular on Byron where the start_time
changes on every restart

- 2cfee3c199188c5f957307ff593bba27e59972a2
  remove unused dependencies and weeds
  
- cd4f04dec3a16c3f514f8f19df5fa810fcec409a
  re-generate nix machinery
  
- a10afb5c5575a19024dc22704565afa1c8e80f70
  remove unavailable commands from cardano-wallet-byron
  These require non-byron wallets, so it doesn't make sense to have them part of the command-line.

- fcaed2b76b2d06179c761bd78d05b90741db3635
  Fix wrong protocol magic for TestNet
  Whoops. Fortunately, we only used MainNet Byron addresses on the ITN :no_mouth:

- 4c970983bf0634dd70b51e91e5ecc814450ce683
  require no magic for cardano-node in the integration setup
  We use the 'mainnet' network target which doesn't require any magic inside addresses and such.

# Comments

```
$ stack test cardano-wallet-byron:integration
.
.
.
.
Finished in 385.0136 seconds
62 examples, 0 failure
```

:warning: There's two scenarios that are _pretty long_ to run in the NETWORK_ group: they need to wait for _the next epoch_, but in OBFT byron, the epoch length is `10*k`. I am not quite sure about setting `k` to values that are _too small_, so I set it to `k=10` at the moment which seems fine (might revisit later) but still leads to an epoch length of 100 slots.


lib/jormungandr/test/integration/Cardano/Faucet.hs
```diff
@@ -1,1483 +0,0 @@
-{-# LANGUAGE DataKinds #-}
-{-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE QuasiQuotes #-}
-{-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TypeApplications #-}
-
-module Cardano.Faucet
```
KtorZ, line 7: Renamed to `Cardano.Wallet.Jormungandr.Faucet` and, move the hard-coded mnemonics to `Test.Integration.Faucet`

lib/core-integration/src/Test/Integration/Scenario/API/ByronTransactions.hs
```diff
@@ -169,28 +167,6 @@ spec = do
...
-        \ Shelley endpoint does not list Byron wallet transactions" $ \ctx -> do
-        w <- emptyRandomWallet ctx
-        let wid = w ^. walletId
-        let ep = ("GET", "v2/wallets/" <> wid <> "/transactions")
-        r <- request @([ApiTransaction n]) ctx ep Default Empty
-        verify r
-            [ expectResponseCode @IO HTTP.status404
-            , expectErrorMessage (errMsg404NoWallet wid)
-            ]
-
```
KtorZ, line 193: Moved to `Scenario/Api/Transactions.hs`

lib/core-integration/src/Test/Integration/Scenario/API/ByronWallets.hs
```diff
@@ -103,29 +86,11 @@ import Test.Integration.Framework.TestData
...
-          where
-            inputBalance = fromIntegral
-                . sum
-                . fmap (view (#amount . #getQuantity))
-                . mapMaybe ApiTypes.source
-                . view #inputs
-            outputBalance = fromIntegral
-                . sum
-                . fmap (view (#amount . #getQuantity))
-                . view #outputs
```
KtorZ, line 127: Moved to `Scenario/Api/Wallets.hs`

lib/core-integration/src/Test/Integration/Scenario/API/ByronWallets.hs
```diff
@@ -175,379 +140,6 @@ spec = do
...
-        expectResponseCode @IO HTTP.status404 r
-        expectErrorMessage (errMsg404NoWallet $ w ^. walletId) r
-
-    it "BYRON_GET_03 - Shelley ep does not show Byron wallet" $ \ctx -> do
-        w <- emptyRandomWallet ctx
-        r <- request @ApiWallet ctx
-            (Link.getWallet @'Shelley w) Default Empty
-        expectResponseCode @IO HTTP.status404 r
-        expectErrorMessage (errMsg404NoWallet $ w ^. walletId) r
-
```
KtorZ, line 550: Moved to `Scenario/Api/Wallets.hs`

lib/core-integration/src/Test/Integration/Scenario/API/ByronWallets.hs
```diff
@@ -593,95 +185,6 @@ spec = do
...
-        w <- emptyWallet ctx
-        r <- request @ApiByronWallet ctx (Link.deleteWallet @'Byron w) Default Empty
-        expectResponseCode @IO HTTP.status404 r
-        expectErrorMessage (errMsg404NoWallet $ w ^. walletId) r
-
-    it "BYRON_DELETE_03 - Shelley ep does not delete Byron wallet" $ \ctx -> do
-        w <- emptyRandomWallet ctx
-        r <- request @ApiByronWallet ctx (Link.deleteWallet @'Shelley w) Default Empty
-        expectResponseCode @IO HTTP.status404 r
-        expectErrorMessage (errMsg404NoWallet $ w ^. walletId) r
```
KtorZ, line 683: Moved to `Scenario/Api/Wallets.hs`

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
```diff
@@ -1471,13 +1498,532 @@ spec = do
...
+        r0 <- request @[ApiTransaction n] ctx
+            (Link.migrateWallet sourceWallet targetWallet )
+            Default
+            (Json [json|{"passphrase": "not-the-right-passphrase"}|])
+        verify r0
+            [ expectResponseCode @IO HTTP.status403
+            , expectErrorMessage errMsg403WrongPass
+            ]
+
+    it "NETWORK_SHELLEY - Wallet has the same tip as network/information" $
```
piotr-iohk, line 1963: This one had to be moved to Wallets.hs? It rather belongs to Network.hs...

KtorZ: Yes / No. I moved it out of `Network` because it actually does make check on a given wallet that has to be a Shelley wallet (which we don't have on the cardano-node integration). Why in `Wallets` then? Because it's still fetching for the wallet's tip and comparing that to the network's tip, so in that sense, both are a good fit (and to be honest, I didn't feel like creating an additional `Scenario/Api/ByronNetwork` because, it's kinda confusing.. the network endpoints are no different between byron and shelley..)

piotr-iohk (approved):
:ok_hand:  
That's really great that we will be able to run our integration tests on `cardano-node OBFT`!
(small comment about NETWORK_SHELLY moved to `Wallets.hs`)

lib/core-integration/src/Test/Integration/Scenario/API/Wallets.hs
piotr-iohk: ok, makes sense. :+1:

lib/byron/test/data/cardano-node/node.config
```diff
@@ -0,0 +1,73 @@
+# vim: set filetype=yaml:
+# -*- mode: yaml -*-
+
+#     _   _           _          ____             __ _
+#    | \ | | ___   __| | ___    / ___|___  _ __  / _(_) __ _
+#    |  \| |/ _ \ / _` |/ _ \  | |   / _ \| '_ \| |_| |/ _` |
+#    | |\  | (_) | (_| |  __/  | |__| (_) | | | |  _| | (_| |
+#    |_| \_|\___/ \__,_|\___|   \____\___/|_| |_|_| |_|\__, |
+#                                                      |___/
```
paweljakubas, line 9: very nice comments:-)

lib/byron/test/integration/Cardano/Wallet/Byron/Config.hs
```diff
@@ -0,0 +1,251 @@
...
+updateStartTime
+    :: Aeson.Object
+    -> IO Aeson.Object
+updateStartTime m = do
+    time <- round @_ @Int <$> getPOSIXTime
+    pure $ HM.insert "startTime" (toJSON time) m
+
+-- | Do something with an a JSON object. Fails if the given JSON value isn't an
+-- object.
+withObject
```
paweljakubas, line 225: wonder if this should not go to Cardano.Wallet.Api.Types and then we would remove aeson dependency?

lib/byron/test/data/cardano-node/node.config
KtorZ: :upside_down_face:

lib/byron/test/integration/Cardano/Wallet/Byron/Config.hs
```diff
@@ -0,0 +1,251 @@
...
+                , nodeToClientCodecCBORTerm
+                )
+              )
+            )
+
+    teardownConfig
+        :: (FilePath, b, c, d)
+        -> IO ()
+    teardownConfig (dir, _, _, _) = do
+        noCleanup <- maybe False (not . null) <$> lookupEnv "NO_CLEANUP"
```
paweljakubas, line 210: :+1:

lib/byron/test/integration/Cardano/Wallet/Byron/Config.hs
KtorZ: Hmmm. Yeah, could be in a `Data.Aeson.Extra` possibly. The function is not unsafe per se if we instantiate the monad `m` to `Maybe` for example. Yet I am not sure it is quite needed anywhere else so we may as well move it if/when needed :thinking: ?

lib/byron/test/data/cardano-node/node.config
KtorZ: One has to find an occupation when re-compiling a new LTS.

lib/byron/test/integration/Main.hs
```diff
@@ -0,0 +1,157 @@
...
+            Network.spec
+
+specWithServer
+    :: Trace IO Text
+    -> SpecWith (Context Byron)
+    -> Spec
+specWithServer tr = aroundAll withContext . after tearDown
+  where
+    withContext :: (Context Byron -> IO ()) -> IO ()
+    withContext action = do
```
paweljakubas, line 106: would be good to reuse somehow : https://github.com/input-output-hk/cardano-wallet/blob/master/lib/jormungandr/test/integration/Main.hs#L164 if possible and maybe put it somewhere in core-integration

lib/byron/test/integration/Main.hs
```diff
@@ -0,0 +1,157 @@
...
+
+    -- | teardown after each test (currently only deleting all wallets)
+    tearDown :: Context t -> IO ()
+    tearDown ctx = do
+        (_, wallets) <- unsafeRequest @[ApiByronWallet] ctx
+            (Link.listWallets @'Byron) Empty
+        forM_ wallets $ \w -> void $ request @Aeson.Value ctx
+            (Link.deleteWallet @'Byron w) Default Empty
+
+-- FIXME: Write fee estimator for Byron nodes.
```
paweljakubas, line 151: so here, we cannot just reuse https://github.com/input-output-hk/cardano-wallet/blob/master/lib/jormungandr/test/integration/Main.hs#L219 because it is going to be different, right?

lib/byron/test/integration/Cardano/Wallet/Byron/Config.hs
```diff
@@ -0,0 +1,251 @@
...
+-- | Spins up a @cardano-node@ in another process.
+--
+-- IMPORTANT: @cardano-node@ must be available on the current path.
+withCardanoNode
+    :: Trace IO Text
+    -- ^ Some trace for logging
+    -> (AddrInfo -> Block -> (BlockchainParameters, NodeVersionData) -> IO a)
+    -- ^ Callback function with a socket description and genesis params
+    -> IO a
+withCardanoNode tr action =
```
paweljakubas, line 101: if `cardano-node` is not on current path are we going to get adequate msg?

lib/byron/test/integration/Cardano/Wallet/Byron/Config.hs
paweljakubas: ah, aeson is needed also for yaml processing..

lib/byron/test/data/cardano-node/node.config
```diff
@@ -0,0 +1,73 @@
...
+
+# these backends are initialized:
+setupBackends:
+  - KatipBK
+
+# here we set up outputs of logging in 'katip':
+setupScribes:
+  - scName: stdout
+    scKind: StdoutSK
+    scFormat: ScText
```
paweljakubas, line 73: still touched by the not-so-user-friendly level of sophistication needed to set logging....

lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -171,7 +187,7 @@ data instance Cursor (m Byron) = Cursor
     (TQueue m (ChainSyncCmd m))
 
 -- | Create an instance of the network layer
-newNetworkLayer
+withNetworkLayer
```
paweljakubas, line 183: definitely better naming, sometimes I think it should be part of our coding style to enforce naming such things like that

lib/byron/test/data/cardano-node/node.config
```diff
@@ -0,0 +1,73 @@
+# vim: set filetype=yaml:
```
paweljakubas, line 1: this one is probably not needed but maybe should stay for vim users;-)

lib/byron/test/data/cardano-node/node.config
```diff
@@ -0,0 +1,73 @@
...
+
+
+#     _   _           _       _          ____
+#    | | | |_ __   __| | __ _| |_ ___   |  _ \ __ _ _ __ __ _ _ __ ___  ___
+#    | | | | '_ \ / _` |/ _` | __/ _ \  | |_) / _` | '__/ _` | '_ ` _ \/ __|
+#    | |_| | |_) | (_| | (_| | ||  __/  |  __/ (_| | | | (_| | | | | | \__ \
+#     \___/| .__/ \__,_|\__,_|\__\___|  |_|   \__,_|_|  \__,_|_| |_| |_|___/
+#          |_|
+
+ApplicationName: cardano-sl
```
paweljakubas, line 28: really this application name?

lib/jormungandr/test/integration/Cardano/Wallet/Jormungandr/Faucet.hs
```diff
@@ -0,0 +1,271 @@
...
+mkTxBuilder
+    :: FeePolicy
+    -> (TxIn, String)
+    -> (Address, Coin)
+    -> IO ByteString
+mkTxBuilder (LinearFee cst coeff _) (TxIn inpTx inpIx, key) (addr, Coin amt) =
+    withSystemTempDirectory "cardano-wallet-jormungandr" $ \d -> do
+        let txFile = d </> "trans.tx"
+        let witnessFile = d </> "witness"
+        let keyFile = d </> "key.prv"
```
paweljakubas, line 84: wonder if we should not have everywhere consistent policy like you used in `teardownConfig` when persisting to filesystem

lib/byron/test/integration/Cardano/Wallet/Byron/Faucet.hs
```diff
@@ -0,0 +1,21 @@
...
+    ( newMVar )
+import Test.Integration.Faucet
+    ( Faucet (..), icaMnemonics, rndMnemonics, seqMnemonics )
+
+initFaucet :: IO Faucet
+initFaucet = Faucet
+    <$> newMVar seqMnemonics
+    <*> newMVar icaMnemonics
+    <*> newMVar rndMnemonics
+    <*> newMVar [] -- FIXME: txBuilder for external transaction.
```
paweljakubas, line 21: here, if I understand differences between cardano-node and jormungadr, we will definitely NOT reuse `mkTxBuilder` from Jormungandr.Faucet but deliver dedicated one. Right?

paweljakubas (approved):
Looks like crucial step to have dedicated integration step cor cardano-node obft. Left some questions which may give rise to better code reuse and also for me to understand better

lib/byron/test/integration/Main.hs
KtorZ: Exactly! To save up some time, I've just provided a dummy implementation for now and will provide a real one (resurrected fromthe http-bridge) in #1350

lib/byron/test/integration/Cardano/Wallet/Byron/Config.hs
KtorZ: Good question :thinking: I'll check that.

lib/byron/test/integration/Main.hs
KtorZ: Yeah, though the annoying part is that, despite the functions having the same names, they have subtle differences and I am quite reluctant in trying to unify them into one common abstraction :/ 
I am afraid of the complexity this would introduce. Maybe worth the try however.

lib/byron/test/data/cardano-node/node.config
KtorZ: :no_mouth:

lib/byron/src/Cardano/Wallet/Byron/Network.hs
KtorZ: Yeah, it really is a recurring pattern for resource allocation.

lib/byron/test/data/cardano-node/node.config
KtorZ: Ah! Good point :thinking: ... I _think_ it does not really matter here since this is our integration configuration, so its a private blockchain. It could be anything we like :) 

(as for where this comes from, I just shamelessly copied it from: https://github.com/input-output-hk/cardano-node/blob/master/benchmarking/cluster3nodes/configuration/log-config-0.yaml)

lib/byron/test/integration/Cardano/Wallet/Byron/Faucet.hs
KtorZ: >  reuse mkTxBuilder from Jormungandr.Faucet but deliver dedicated one. Right?

Precesily. The tx builder in Jormungandr extensively uses jcli and construct transactions that follow Jormungandr's binary format. For cardano-node, it's a different story.

lib/jormungandr/test/integration/Cardano/Wallet/Jormungandr/Faucet.hs
KtorZ: What do you mean :thinking: ? (about the file extensions I picked for the node: node.config, node.key, node.cert, node.topology etc... ?)

lib/byron/test/data/cardano-node/node.config
KtorZ: And the one below is for Emacs users. Because obviously, there are only two text editors out there :no_mouth:

lib/jormungandr/test/integration/Cardano/Wallet/Jormungandr/Faucet.hs
paweljakubas: I meant that for Byron.Integration you used `teardownConfig` which removed tmp files unless `NO_CLEANUP` is set. And wonder if cleaning "policy" should not be the same here in Jormungandr. Or we should also use `NO_CLEANUP` here to have the same cleaning policy of tmp dir.  Just for consistency

KtorZ: It is the same in Jörmungandr actually but you have to compare the same elements (i.e. the configuration generation) :wink: 

--> https://github.com/input-output-hk/cardano-wallet/blob/master/lib/jormungandr/test/integration/Cardano/Wallet/Jormungandr/Launch.hs#L64-L74

In the `Faucet` however, those temporary files are merely used as buffers because it is trickier to provide inputs on stdin when invoking `jcli` from Haskell.

lib/byron/test/integration/Cardano/Wallet/Byron/Config.hs
KtorZ:
```
  1) API Specifications, BYRON_TXS, TRANS_CREATE_01 - Single Output Transaction
       uncaught exception: ProcessHasExited
       ProcessDidNotStart "cardano-patate" cardano-patate: createProcess: runInteractiveProcess: exec: does not exist (No such file or directory)
```

Issue #1394, opened 2020-02-27
Implement network/clock endpoint


paweljakubas:
# Context
NTP is a protocol to allow time synchronization between remote machines. This is particularly essential for Ouroboros is a time-based protocol. In case a local clock is drifting too much compared to the global clock, one would want to re-sync his clock.

# Decision
We are adding `network/clock` endpoint which returns NtpStatus type (Drift, Pending, Unavailable). 

# Acceptance Criteria

1. `network/clock` endpoint along with corresponding types has to be added
2. swagger.yaml has to be updated
3. in the implementation of the endpoint https://github.com/input-output-hk/ouroboros-network/tree/master/ntp-client is used
4. illustrative tests of endpoint usage are added

---

# Development 

# QA

piotr-iohk:
@paweljakubas I'm seeing this error in the log when invoking `network/clock`:

```
[cardano-wallet.ntp-client:Info:30] [2020-03-04 12:30:46.31 UTC] query to ntp client invoked
[cardano-wallet.ntp-client:Alert:10310] [2020-03-04 12:30:46.32 UTC] ntp client experienced error Network.Socket.ByteString.sendManyTo: does not exist (Network is unreachable) when sending packet using IPv6
[cardano-wallet.ntp-client:Alert:10304] [2020-03-04 12:30:46.32 UTC] ntp client experienced error Network.Socket.ByteString.sendManyTo: does not exist (Network is unreachable) when using IPv6 protocol
```

Also, if I'm not mistaken there is only a single integration test added for this requirement (https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core-integration/src/Test/Integration/Scenario/API/Network.hs#L202-L210). 

I'm wondering if there could be some more unit tests (or properties) added for https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Network/Ntp.hs#L132-L143 that would excercise all patterns?
Not sure how it works underneath, but is it possible maybe to get `NtpSyncUnavailable` by tweaking https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Network/Ntp.hs#L44 ?

piotr-iohk:
After discussing with @paweljakubas -> #1410.

rvl:
The base library `ouroboros-network/ntp-client` might need more unit tests added. It is missing error path tests such as `NtpSyncUnavailable` mentioned by @piotr-iohk.

[...]

PR #1415, opened 2020-03-06
stdoutTextTracer: Use line-buffered print statements

rvl:
### Issue number

Found while reviewing #1394.

### Summary

Use `B8.putStrLn`, which seems to be line-buffered, whereas `T.putStrLn` is character-buffered.

So rather than

    Starting ntp client
    quernyt pt oc lnitepn tc liise nttr iigngveorkeedd

we get

    Starting ntp client
    query to ntp client invoked
    ntp client is triggered

### DIFF
--- a/lib/core/src/Cardano/Wallet/Logging.hs
+++ b/lib/core/src/Cardano/Wallet/Logging.hs
@@ -41,7 +41,8 @@ import Data.Text
 import Data.Text.Class
     ( ToText (..) )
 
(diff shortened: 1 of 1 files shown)


Anviking (approved):

KtorZ (approved):
Ho. I thought only the string version `putStrLn :: String -> IO` was character-buffered by default :man_facepalming: ...

PR #1417, opened 2020-03-06
Move remaining integration tests to ApiSpec.hs

piotr-iohk:
# Issue Number

# Overview

- 1cc22892bee08502c98fff359caadec04d99023d
  Move tests to ApiSpec.hs:  - SomeByronWalletPostData  - WalletOrAccountPostData
  
- afcfac20ce584024e94c54b4fe266001676b83cd
  Move tests to ApiSpec.hs:  - WalletPutPassphraseData  - WalletPutData  - ApiWalletPassphrase  - ApiSelectCoinsData  - PostTransactionData  - PostTransactionFeeData  - ApiNetworkTip
  
- cd7d754d30e68d1fb149484cf6bb7152083bf87d
  No cases for PostExternalTransactionData
  
- 5de6d8db0e173474280219982bf606445e8356b7
  Add cases covering malformed JSON as body param and remove outstanding 'MOVE TO' comments

# Comments

PostExternalTransactionData which is using `application/octet-stream`
    - it seems impossible to test in `ApiSpec.hs` as all `400` requests are served by the error:
```
{
  "code": "malformed_tx_payload",
  "message": "I couldn't verify that the payload has the correct binary format. Therefore I couldn't send it to the node. Please check the format and try again."
}
```
which does not come from Servant but from `Server.hs` code -> https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/Api/Server.hs#L1923
There are integration tests testing that -> https://github.com/input-output-hk/cardano-wallet/blob/master/lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/Transactions.hs#L308-L342


lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
```diff
@@ -126,6 +127,25 @@ class Malformed t where
 --
 -- Class instances (PathParam)
 --
+instance Wellformed (PathParam (QueryParam "state" (ApiT AddressState))) where
+    wellformed = PathParam "?state=used"
+
+instance Malformed (PathParam (QueryParam "state" (ApiT AddressState))) where
+    malformed = first PathParam <$>
```
piotr-iohk: @KtorZ, I'm not able to make it work for the AddressState, which is a QueryParam in the form of "?state=used" Added this Malformed instance for `AddressState` but no new test cases are being generated in `ApiSpec.hs`. I have also tried:
```
instance Malformed (PathParam (ApiT AddressState))
```
but still no luck.

KtorZ: QueryParams are not yet captured by the generic tests, cf: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Cardano/Wallet/ApiSpec.hs#L343-L350

piotr-iohk: OK, I have wrapped up this PR without `QueryParams`. All `PathParam` and `BodyParam` cases should be moved and covered in `ApiSpec.hs` though.

KtorZ (approved):

Issue #1416, opened 2020-03-06
Restore sequential wallet from account public key - CLI support


paweljakubas:
# Context

As a cardano-wallet API client, 
I want to be able to restore a sequential wallet from an account public key using CLI

# Decision

# Acceptance Criteria

1. CLI support and corresponding tests are added

---

# Development 

# QA

PR #1421, opened 2020-03-09
CLI support for account public key restoration

paweljakubas:
# Issue Number

#1416 

# Overview

- [x] I have added `cmdRestoreWallet`
- [x] I have extended CLISpec
- [x] I have added illustrative integration test to CLI scenarios

# Comments

### DIFF
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -785,7 +785,7 @@ cmdWallet = command "wallet" $ info (helper <*> cmds) $ mempty
     cmds = subparser $ mempty
         <> cmdWalletList @t
         <> cmdWalletCreate @t
-        <> cmdWalletRestore @t
+        <> cmdWalletCreateFromPublicKey @t
         <> cmdWalletGet @t
         <> cmdWalletUpdate @t
         <> cmdWalletDelete @t
@@ -845,25 +845,31 @@ cmdWalletCreate = command "create" $ info (helper <*> cmd) $ mempty
                 (ApiT wName)
                 (ApiT wPwd)
 
-cmdWalletRestore
+-- | Arguments for 'wallet create from-public-key' command
+data WalletCreateFromPublicKeyArgs = WalletCreateFromPublicKeyArgs
+    { _port :: Port "Wallet"
+    , _name :: WalletName
+    , _gap :: AddressPoolGap
+    , _key :: AccountPublicKey
+    }
+
+cmdWalletCreateFromPublicKey
     :: forall t. (DecodeAddress t, EncodeAddress t)
     => Mod CommandFields (IO ())
-cmdWalletRestore = command "restore" $ info (helper <*> cmd) $ mempty
-    <> progDesc "Restore a wallet using a public account key."
+cmdWalletCreateFromPublicKey =
+    command "create from-public-key" $ info (helper <*> cmd) $ mempty
+    <> progDesc "Create a wallet using a public account key."
   where
-    cmd = fmap exec $ WalletCreateArgs
+    cmd = fmap exec $ WalletCreateFromPublicKeyArgs
         <$> portOption
         <*> walletNameArgument
         <*> poolGapOption
-    exec (WalletCreateArgs wPort wName wGap) = do
-        pubAccKey <- do
-            let prompt = "Please enter a 128 character public account key: "
-            let parser = fromText @AccountPublicKey
-            fst <$> getLine prompt parser
+        <*> accPubKeyArgument
+    exec (WalletCreateFromPublicKeyArgs wPort wName wGap wAccPubKey) =
         runClient wPort Aeson.encodePretty $ postWallet (walletClient @t) $
             WalletOrAccountPostData $ Right $ AccountPostData
                 (ApiT wName)
-                pubAccKey
+                wAccPubKey
                 (Just $ ApiT wGap)
 
 -- | Arguments for 'wallet get' command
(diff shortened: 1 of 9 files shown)


lib/cli/src/Cardano/CLI.hs
```diff
@@ -836,6 +839,27 @@ cmdWalletCreate = command "create" $ info (helper <*> cmd) $ mempty
...
+  where
+    cmd = fmap exec $ WalletCreateArgs
+        <$> portOption
+        <*> walletNameArgument
+        <*> poolGapOption
+    exec (WalletCreateArgs wPort wName wGap) = do
+        pubAccKey <- do
+            let prompt = "Please enter a 128 character public account key: "
+            let parser = fromText @AccountPublicKey
+            fst <$> getLine prompt parser
```
KtorZ: What drives the choice for an interactive prompt here :thinking: ? 
It seems easier / more interoperable (i.e. piping) to have the command-line take the key as an argument here.

lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -243,6 +243,8 @@ spec = do
             , "  list                     List all known wallets."
             , "  create                   Create a new wallet using a sequential"
             , "                           address scheme."
+            , "  restore                  Restore a wallet using a public"
```
KtorZ: `create` and `restore` are actually two synonyms for the wallet :) 

What about creating a sub-command of `create` in the form of 

`create from-mnemonic`  and `create from-public-key`

lib/cli/src/Cardano/CLI.hs
piotr-iohk: Right, I suppose public key does not have to be "hidden" inside interactive prompt since it is public... (unlike mnemonic sentence for that matter)

paweljakubas: persuasive points: done in https://github.com/input-output-hk/cardano-wallet/pull/1421/commits/a2b4df64a33184d951009ff9c34e600da8d6fc3a

lib/cli/test/unit/Cardano/CLISpec.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/1421/commits/a2b4df64a33184d951009ff9c34e600da8d6fc3a and https://github.com/input-output-hk/cardano-wallet/pull/1421/commits/4cf7ba4b9df864e1e7d41504f653e1054a83f440

paweljakubas:
bors try

lib/cli/src/Cardano/CLI.hs
```diff
@@ -804,11 +807,12 @@ data WalletCreateArgs = WalletCreateArgs
...
     }
 
-cmdWalletCreate
+cmdWalletCreateFromMnemonic
     :: forall t. (DecodeAddress t, EncodeAddress t)
     => Mod CommandFields (IO ())
-cmdWalletCreate = command "create" $ info (helper <*> cmd) $ mempty
-    <> progDesc "Create a new wallet using a sequential address scheme."
+cmdWalletCreateFromMnemonic =
+    command "create from-mnemonic" $ info (helper <*> cmd) $ mempty
```
KtorZ: Shouldn't this actually be:

```hs
cmdWalletCreate :: Mod CommandFields (IO ())
cmdWalletCreate = command "create" $ info (helper <*> cmds) $ mempty
    <> progDesc "Create a new wallet."
  where
    cmds = subparser $ mempty
        <> cmdWalletCreateFromMnemonic
        <> cmdWalletCreateFromPublicKey

-- etc ...
```

i.e., having both as sub-commands of a parent command (like it's done for `cmdMnemonic` for instance)

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -625,6 +625,15 @@ data ApiPoolId
     | ApiPoolId PoolId
     deriving (Eq, Generic, Show)
 
+instance FromText AccountPublicKey where
+    fromText txt = case xpubFromText (T.encodeUtf8 txt) of
+        Left _ ->
+            Left $ TextDecodingError $ unwords $
+            [ "AccountPublicKey: unable to deserialize ShelleyKey from json. "
+            , "Expecting hex-encoded string of 128 characters."]
```
KtorZ: The error message shouldn't mention:

- `ShelleyKey` (that's an internal data-type, not really suitable for public interfaces)
- `json` (it's actually not deserializing from JSON here, although we are probably using this instance also in JSON instances). 

I'd suggest something more aligned with other parser from hex-encoded bytestring:

_"Invalid account public key: expecting a hex-encoded value that is 64 bytes in length."_

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -625,6 +625,15 @@ data ApiPoolId
     | ApiPoolId PoolId
     deriving (Eq, Generic, Show)
 
+instance FromText AccountPublicKey where
```
KtorZ: Why is this instance located here :thinking: ? Or perhaps, if `AccountPublicKey` is an API type, what about renaming it to follow the same convention as other API types (either starting with `Api..` or, wrapped in `ApiT`).

lib/cli/src/Cardano/CLI.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/1421/commits/eb3a59a43df8cccfca3f48912e290fc519c62316

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: yes, my bad. I move it directly from FromJSON. corrected : https://github.com/input-output-hk/cardano-wallet/pull/1421/commits/ddec074ddadbddb70b0cfbd8899efa145b9c0dc7

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: changed `AccountPublicKey` in `Api.Types` to `ApiAccountPublicKey` and propagated through code base in https://github.com/input-output-hk/cardano-wallet/pull/1421/commits/ddec074ddadbddb70b0cfbd8899efa145b9c0dc7

paweljakubas:
bors try

paweljakubas:
bors try

paweljakubas:
bors try

paweljakubas:
bors try

paweljakubas:
bors try

KtorZ (approved):
bors r+

PR #1425, opened 2020-03-10
Fix launcher unit tests on windows

rvl:
### Overview

- Fixes a hang in the test cases (happened on Linux and Windows).
- Add integration test for case where parent process sets stdin to `/dev/null`.
- Make `Cardano.Startup.withShutdownHandler` work on Windows without hanging.

### Comment

- [x] [cardano-wallet-launcher.unit](https://hydra.iohk.io/job/Cardano/cardano-wallet-pr-1425/x86_64-w64-mingw32.checks.cardano-wallet-launcher.unit.x86_64-linux)

### DIFF
--- a/lib/launcher/test/unit/Cardano/StartupSpec.hs
+++ b/lib/launcher/test/unit/Cardano/StartupSpec.hs
@@ -106,7 +106,7 @@ spec = describe "withShutdownHandler" $ do
 
 withPipe :: ((Handle, Handle) -> IO a) -> IO a
 withPipe = bracket createPipe closePipe
-    where closePipe (a, b) = hClose a >> hClose b
+    where closePipe (a, b) = hClose b >> hClose a
 
 captureLogging' :: (Tracer IO msg -> IO a) -> IO [msg]
 captureLogging' = fmap fst . captureLogging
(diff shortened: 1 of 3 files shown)


lib/launcher/test/unit/Cardano/StartupSpec.hs
```diff
@@ -32,19 +33,55 @@ import Test.Hspec
...
         it "race stdin" $ do
-            res <- race (BS.hGet stdin 1000) (threadDelay decisecond)
+            pendingOnWindows "deadlocks on windows"
+            res <- race (getChunk stdin) (threadDelay decisecond)
             res `shouldBe` Right ()
 
         it "race pipe" $ withPipe $ \(a, _) -> do
-            res <- race (BS.hGet a 1000) (threadDelay decisecond)
+            pendingOnWindows "deadlocks on windows"
+            res <- race (getChunk a) (threadDelay decisecond)
```
KtorZ, line 84: So we can't simply read chunks from pipes on Windows?

KtorZ (approved):

lib/launcher/test/unit/Cardano/StartupSpec.hs
rvl: Yeah actually we can, it's fine as long as it's wrapped in `forkIO`. I changed the `hWaitForInput` calls back to `hGet`.

rvl:
I fixed the test case.

bors r+

rvl:
Trying again because I think the integration test may be flaky (#1446).

bors r+

rvl:
Trying again because Hydra evaluation was not working yesterday.

bors r+

PR #1429, opened 2020-03-11
Allow serving wallet w/ cardano-node on either testnet or mainnet 

KtorZ:
# Issue Number

N/A (some debts regarding the cardano-wallet-byron command-line)

# Overview

- 839135bc2c6fd7c91fe427e0201c1c71b704fbf9
  remove network discriminant type parameter for cmdNetwork
  The same approach (i.e. using clients based on sub-part of the API,
instead of the global 'Api t' type) could be used for a few other
commands that are actually fairly agnostic to what the network
discriminant is.

- 57ca08ceca6f7e1a475a892f34dbe27b463d3a16
  allow serving cardano-wallet-byron on either testnet or mainnet
  There's no particular requirement as of _now_ about being able to serve custom networks from the
CLI. So, the command-line serve only so-called 'known networks', with pre-defined configuration.
Note that, to things working fine at the type-level, some extra type machinery was needed, mostly
playing around with proxies. The command-line now expects a flag, either '--testnet' or '--mainnet'
to be provided, or fails, e.g.:

    cardano-wallet serve --node-socket node.socket
    Missing: (--mainnet | --testnet)

# Comments

### DIFF
--- a/lib/byron/exe/cardano-wallet-byron.hs
+++ b/lib/byron/exe/cardano-wallet-byron.hs
@@ -57,7 +57,8 @@ import Cardano.Startup
 import Cardano.Wallet.Api.Server
     ( HostPreference, Listen (..) )
 import Cardano.Wallet.Byron
-    ( TracerSeverities
+    ( SomeNetworkDiscriminant (..)
+    , TracerSeverities
     , Tracers
     , Tracers' (..)
     , serveWallet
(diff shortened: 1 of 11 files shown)


paweljakubas (approved):
nice way of extracting network discriminant!

Just small things : nix patch and hlint : 
```
lib/byron/src/Cardano/Wallet/Byron/Compatibility.hs:9:1: Warning: Unused LANGUAGE pragma
--
  | Found:
  | {-# LANGUAGE TypeApplications #-}
  | Perhaps you should remove it.

```

piotr-iohk (approved):

PR #1432, opened 2020-03-12
fix LAUNCH - Restoration workers restart hanging infinitely

piotr-iohk:
# Issue Number

# Overview

- f7df783733bf6bfe6319d03a17435a2f9e329f09
  fix LAUNCH - Restoration workers restart hanging infinitely

# Comments

 Jormungandr needs to have `--config`, otherwise it complains and doesn't start:
```
Mar 12 15:05:53.564 ERRO trusted-peers cannot be empty. to avoid bootstrap use 'skip_bootstrap: true', task: bootstrap
```


KtorZ (approved):
bors r+

rvl (approved):
Nice catch :+1:

PR #1438, opened 2020-03-13
Better response example in api doc for network/clock

piotr-iohk:
# Issue Number

# Overview

- 62128be5b54998fdc014f82d2fc606cc3fb10411
  Better response example in api doc for network/clock
  
- e632629f8a5f3818d4b49689482603185b88f596
  Fix for failng validateEveryToJSON test: "Cardano.Wallet.Api.Types, verify that every type used with JSON content type in a servant API has compatible ToJSON and ToSchema instances using validateToJSON., ApiNetworkClock"

# Comments

| Before | After  |
|--|--|
|Screenshot from 2020-03-13 13-13-39 | Screenshot from 2020-03-13 13-13-57|

### DIFF
--- a/specifications/api/swagger.yaml
+++ b/specifications/api/swagger.yaml
@@ -581,7 +581,6 @@ x-networkInformationNtpStatus: &networkInformationNtpStatus
       properties:
         quantity:
           type: integer
-          minimum: 0
           example: 14
         unit:
           type: string
--- a/specifications/api/swagger.yaml
+++ b/specifications/api/swagger.yaml
@@ -565,10 +565,10 @@ x-networkInformationNtpStatus: &networkInformationNtpStatus
     status:
       type: string
       enum:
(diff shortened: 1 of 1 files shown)


paweljakubas (approved):
:+1:

paweljakubas:
hmm wonder if the changes you adopted will not induce errors in `Cardano.Wallet.Api.Types` :thinking:  probably (swagger vs Types.Api we have)

piotr-iohk:
I think this should fix this -> e632629f8a5f3818d4b49689482603185b88f596.
Nice tests btw. I suppose they are running against the "active examples" in swagger only, so after changed the example it failed. (because the `minimum` was there before)

piotr-iohk:
bors try

piotr-iohk:
Timed out... :(

bors r+

Issue #1442, opened 2020-03-14
Unable to retrieve stake-pool list


PegasusPool:
# Context

Unable to retrieve stake-pool list for the last couple of hours.
v2020-01-27 has been running for months without issues.
 
Tried deleting the entire database used by cardano-wallet, the error still occurs after restarting. Upgrading to v2020-03-11 did not resolve the issue.

# Steps to Reproduce

Start Cardano-Wallet
```
cardano-wallet serve --node-port 3101 --port 3102 --database db --genesis-block-hash 8e4d2a343f3dcf9330ad9035b3e8d168e6728904262f2c434a4f8f934ec7b676
```

Execute
```
cardano-wallet stake-pool list --port 3102
```

## Expected behavior

Return a list of information about stake pools

## Actual behaviour
 
Response:
```
Something is terribly wrong with the metrics I collected. I recorded that some blocks were produced by 607bd76be14c0b852be7f6a2795e95bfdebd290649b7b61af3a81bb2fdb18604 but the node doesn't know about this stake pool!
```

# Possible cause

Related to the recent introduction of pool retirement feature in jormungandr?
Talked to the owner of pool 607bd76be14c0b852be7f6a2795e95bfdebd290649b7b61af3a81bb2fdb18604 and confirmed 
 that this pool was retired in epoch 89.

| Information  | -                                                     |
| ---          | ---                                                   |                                      
| Version      | v2020-03-11 & v2020-01-27						       |
| Platform     | Ubuntu 18.04        		  						   |
| Installation | Binary from Github Release                            |

SebastienGllmt:
Pool retirements were not part of the scope for cardano-wallet Jormungandr support so there is no plan to add this as far as I know.

PegasusPool:
Do you mean that cardano-wallet is not expected to work with jormungandr anymore  and I shouldn't expect the stake-pool list call to ever work again unless it's used with cardano-node?

KtorZ:
@PegasusPool are you able to build cardano-wallet from sources? 
I can't sync with the ITN today it seems. My node is at epoch 1 and progressing VERY slowly. I got a fix which is kinda hard to test with a private blockchain so I was thinking trying that out on the actual ITN directly.

GaborPeto:
@KtorZ I usually just used the binaries but I can surely try to compile. Which branch would you like me to try?

KtorZ:
@GaborPeto thank you.

[...]

PR #1443, opened 2020-03-14
Don't error when finding blocks by pools not in the stake distribution

Anviking:
# Issue Number

#1442 potential solution.

# Overview

```
The erroring was originally a precaution.

While I haven't checked, it must be that jormungandr removes pools in
the stake-distribution... _before_ they are retired?

This patch seems to solve the problem.
```
# Comments

- ⚠️ We should investigate more what is going on ⚠️ 
- [ ] TODO: If we do this, we should remove the now unused `ErrMetricsInconsistency`.

### DIFF
--- a/lib/core/src/Cardano/Pool/Metrics.hs
+++ b/lib/core/src/Cardano/Pool/Metrics.hs
@@ -99,7 +99,7 @@ import Data.List
 import Data.List.NonEmpty
     ( NonEmpty )
 import Data.Map.Merge.Strict
-    ( traverseMissing, zipWithMatched )
+    ( dropMissing, traverseMissing, zipWithMatched )
 import Data.Map.Strict
     ( Map )
 import Data.Ord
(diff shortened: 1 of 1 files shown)


KtorZ:
@Anviking what is going on is that stake pool de-registration have landed into `jcli` and jormungandr's API. So operators are starting to de-register pools. While I thought that would originally only impact the listing view (in the sense that de-registered pools would still show up) I didn't quite realize this would also affect the metrics calculations. 

Thanks for the PR but I think I'll take it even further and actually have correct support for de-registration certificates. ~~We already parse them at the binary level but simply do nothing with them.~~ (edit: actually not quite...) I am still not quite sure how to properly test that from the wallet since the self node we use for testing behaves weirdly when it comes to pool registration. ~~We have fixtures for submitting de-registration certificates~~ (edit: nope, we actually don't). but the pool we may registered via this way simply don't produce.

Anviking:
Superseded by #1444

PR #1444, opened 2020-03-14
Store pool retirement certificates + fix stake distribution fetching & combining.

KtorZ:
# Issue Number

#1442

# Overview

- c74803733f4d84285f768a40c9b12f9c5614491c
  bump versions in README compatibility matrix
  
- ded40cc9f34dfddf79a840a5b50dc997435fe770
  add Jörmungandr binary deserialisers for pool retirement certificates
  
- 42e387d6d73bf796e249ed765072863dacf64448
  track 'PoolRetirementCertificate' in the stake pool worker
  The corresponding database function doesn't exist _yet_, coming in the next commit.
So far, this would simply add the retirement certificates into a new table but, do
nothing with them.

- 61203d457017d5b6bfffea007d2c6a50eef75a37
  implement initial behavior for 'putPoolRetirement' in the database layer
    Any pool that has submitted a retirement certificate will be removed
  from the list, regardless of the retirement time set by the pool.

  Ideally, we would prefer to only discard pool when they are actually
  retired (i.e. based on the time they set).

- 7c8623a1a75a427124846f8057e1e794702e130a
  Fetch stake distribution by epoch instead of latest one
  
- 9d2fd50c1d7d11ddc09772051ee658d2dd8d74ee
  Ignore blocks produced by pools not in the stake distribution.
  The erroring was originally a precaution.

While I haven't checked, it must be that jormungandr removes pools in
the stake-distribution... before they are retired?

This patch seems to solve the problem.

- 7f43e64350284fe0ad75e124bb5a023ce207d5e3
  cleanup ErrMetricsInconsistency now obsolete

- 0858080446eb9f0964a437a4e5f2e729be328e59
  update jormungandr.nix references for 0.8.14
  
# Comments

:warning: turns out that handling de-registration certificates was kinda useless. Problem was that we were trying to merge all-time production with stake distribution of the latest epoch which.. doesn't work.

On the way fixing this, I also implemented fetching the stake distribution based on the relevant epoch number, so this is now done as part of the forward loop and should automatically make the metrics & ranking much more stable. 

Also added code for managing retirement certificates. We still do nothing with that but.. it's there now :man_shrugging: ... when relevant, we can then think about notifying frontend applications about de-registered pools.

### DIFF
--- a/lib/core/src/Cardano/Pool/Metrics.hs
+++ b/lib/core/src/Cardano/Pool/Metrics.hs
@@ -8,6 +8,7 @@
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections #-}
 
 -- | This module can fold over a blockchain to collect metrics about
 -- Stake pools.
@@ -81,7 +82,7 @@ import Cardano.Wallet.Unsafe
 import Control.Arrow
     ( first )
 import Control.Monad
-    ( forM_, when )
+    ( forM, forM_, when )
 import Control.Monad.IO.Class
     ( liftIO )
 import Control.Monad.Trans.Class
@@ -93,7 +94,7 @@ import Control.Tracer
 import Data.Functor
     ( (<&>) )
 import Data.Generics.Internal.VL.Lens
-    ( (^.) )
+    ( view, (^.) )
 import Data.List
     ( nub, nubBy, sortOn, (\\) )
 import Data.List.NonEmpty
@@ -122,6 +123,7 @@ import System.Random
     ( StdGen )
 
 import qualified Cardano.Pool.Ranking as Ranking
+import qualified Data.List.NonEmpty as NE
 import qualified Data.Map.Merge.Strict as Map
 import qualified Data.Map.Strict as Map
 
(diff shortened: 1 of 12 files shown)


lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -212,15 +213,28 @@ newDBLayer trace fp = do
...
+            { poolId
+            , poolRetirementActiveBy
+            } -> do
+            -- Conversion functions treat 'NominalDiffTime' as seconds.
+            let timeSinceBlock0 = round poolRetirementActiveBy
+            insert_ $ PoolRetirement poolId point timeSinceBlock0
+
+        , readPoolRetirement = \poolId -> do
+            fmap (fromPoolRetirement . entityVal)
+                <$> selectFirst [ PoolRetirementPoolId ==. poolId ] []
```
Anviking: (For the part you now split off) There can't be multiple pool retirements, or something sneaky?

lib/core/src/Cardano/Pool/Metrics.hs
```diff
@@ -195,15 +197,19 @@ monitorStakePools tr (block0, Quantity k) nl db@DBLayer{..} = do
         -> BlockHeader
         -> IO (FollowAction ErrMonitorStakePools)
     forward blocks nodeTip = handler $ do
-        (ep, dist) <- withExceptT ErrMonitorStakePoolsNetworkUnavailable $
-            stakeDistribution nl
+        let epochs = NE.nub $ view (#header . #slotId . #epochNumber) <$> blocks
+        distributions <- forM epochs $ \ep -> do
+            liftIO $ traceWith tr $ MsgStakeDistribution ep
+            withExceptT ErrMonitorStakePoolsNetworkUnavailable $
+                (ep,) <$> stakeDistribution nl ep
```
Anviking, line 200: So most likely we'll still make 1 request and 2 at epoch boundaries, without the tip changing — should be just fine.

Anviking (commented):
lgtm but getting
```
[cardano-wallet.pools-engine:Error:26] [2020-03-14 19:38:31.86 UTC] Recoverable error following the chain: ErrUnexpectedNetworkFailure (Link {_segments = ["api","v0","stake","0"], _queryParams = []}) (FailureResponse (Request {requestPath = (BaseUrl {baseUrlScheme = Http, baseUrlHost = "127.0.0.1", baseUrlPort = 61988, baseUrlPath = ""},"/api/v0/stake/0"), requestQueryString = fromList [], requestBody = Nothing, requestAccept = fromList [application/json;charset=utf-8,application/json], requestHeaders = fromList []), requestHttpVersion = HTTP/1.1, requestMethod = "GET"} (Response {responseStatusCode = Status {statusCode = 404, statusMessage = "Not Found"}, responseHeaders = fromList [("content-length","0"),("date","Sat, 14 Mar 2020 19:38:31 GMT")], responseHttpVersion = HTTP/1.1, responseBody = ""}))
```

Anviking:
Ah, I had _installed_ the new jormungandr, but I was still running the old instance through Daedalus 🤦‍♂ will retry

Anviking (approved):
👍

PR #1447, opened 2020-03-17
Update cardano-node to latest master version

rvl:
### Issue Number

None

### Overview

- This updates cardano-node to the revision 55d238cd5e9656b90fd7b101b223f9aca4a4036e. This is one (necessary) PR ahead of the 1.8.0 tag. The windows build of `cardano-node.exe` should now work.
- It also bumps other niv sources to latest versions.
- It adds the `niv` tool to the default nix-shell. And it adds `stack` (same version as used by Buildkite), in case you don't have that globally installed.
- It adds a windows release job for `cardano-wallet-byron`, and adds it to the windows testing bundle.
- The `Network.NTP.Client` library is now on the same git rev as the rest of `ouroboros-network`
- `db-converter` is added as a top-level build attribute, so that Daedalus can import it.

### Comments

[Hydra jobset](https://hydra.iohk.io/jobset/Cardano/cardano-wallet-pr-1447)


craigem:
This looks good to me Rodney. Bump it with another commit and that failing test ought to pass (known issue) and then I'll approve and bors.

A minor nit, personal idiosyncrasy - Nix can do inline comments and often they are more readable and have better context ie:

```
, exe            # executable component for the program, from Haskell.nix
, cardano-node   # node backend
```

I've not done it as a review, as it's not a blocker, just a comment :-)

stack.yaml
```diff
@@ -102,21 +104,24 @@ extra-deps:
...
     - ouroboros-consensus
+    - ouroboros-consensus-byron
     - typed-protocols
-    - typed-protocols-cbor
+    - typed-protocols-examples
+    - ouroboros-network-framework
+    - ouroboros-network-testing
 
 - git: https://github.com/input-output-hk/ouroboros-network
   commit: f12e7c2168fe3bc4e236898a6e0207479cda3302
```
paweljakubas: @rvl  @Anviking  please take notice that we used this trick (repeating ouroboros-network with newer commit) to have implementation of ntp-client. I bet now it can be erased and `ntp-client` can be put in subdirs above. It would be worth checking this

stack.yaml
```diff
@@ -132,5 +139,7 @@ flags:
 # See https://github.com/ndmitchell/weeder/issues/53
 ghc-options: {"$locals": -ddump-to-file -ddump-hi}
 
+allow-newer: true
```
rvl: It would be preferable not to use this. We should be able to avoid it.

craigem: This is pretty wide spread. I found it in the following, non-exhaustive list of repos:

* cardano-node
* daedalus
* cardano-ledger
* cardano-rest
* cardano-explorer
* cardano-base
* cardano-sl
* cardano-db-sync

default.nix
```diff
@@ -73,6 +73,7 @@ let
     inherit (jmPkgs) jormungandr jormungandr-cli;
     inherit (pkgs) cardano-node;
     inherit (haskellPackages.cardano-wallet-core.identifier) version;
+    inherit (haskellPackages.ouroboros-consensus-byron.components.exes) db-converter;
```
rvl, line 78: Could you add a comment explaining why this needs to be here? Otherwise someone might remove it.

nix/haskell.nix
```diff
@@ -38,6 +38,22 @@ let
   pkgSet = haskell.mkStackPkgSet {
     inherit stack-pkgs;
     modules = [
+      # Allow reinstallation of Win32
```
rvl: This does not adequately explain it for me ... why is this needed, what does it do? @cleverca22

cleverca22: https://github.com/input-output-hk/ouroboros-network/blob/master/nix/haskell.nix#L24-L39
it was copied directly from the same file in `ouroboros-network`
all of the packages in that list, are baked into the ghc, and cant be updated, causing `haskell.nix` to silently ignore the version you specify in `stack.yaml`

the default list includes `Win32`, and this will overwrite that default list

stack.yaml
cleverca22: i suspect it isn't needed, and things will still pass if its taken back out

stack.yaml
rvl: Thanks @paweljakubas - I have addressed this in de807ece5459ec002175f292e5de9b0f9770e821.

stack.yaml
craigem: I suspect it is. I'd have a chat with @erikd to confirm.

IIRC we had issues with at least one build where add this line resolved it.

rvl:
I attempted to add cardano-node as a git extra-dep in `stack.yaml`, but this turned into an unsuccessful dependency chase.

Anviking (approved):
I'm not qualified to review most of this, but LGTM. @KtorZ reviewed my additions in #1451 which is included here

nix/haskell.nix
rvl: Thanks, I have added more detail to the comment.

stack.yaml
rvl: I removed it because it's not required.

rvl:
Thanks @Anviking @cleverca22 @craigem @paweljakubas.

bors r+

Issue #1436, opened 2020-03-13
Implement the endpoint for byron restoration from XPrv


paweljakubas:
# Context

I want Daedalus to be able to automatically restore my existing Byron wallets, 
So that I don't need to re-enter my recovery phrase 
And I can enjoy a smooth migration experience to Byron reboot.

# Decision

We have scaffolding present after #1435 , no w we want implementation.

# Acceptance Criteria
1. Implementation of endpoint
2. Add illustrative test that this is correct.

---

# Development 

# QA 

several things were addressed here that should be double-checked and maybe needs further more extensive testing:
1. `swagger.yaml` was corrected in a sense that 64-byte constraint on passphrase hash was lifted. This is because **Scrypt encryption** can be realized with different parameters (for example number of bytes of content) and also salt length. Those factors contribute to different encrypted hash length as shown in golden tests in `AddressDerivationSpec`. What is interesting : using **different** Scrypt encryption parameters and salt combinations give rise to proper `passphraseCheck` and ability to send legal tx that is accepted by cardano-node. So lifting 64-byte restriction we are making our migration capabilities more flexible (in case in different versions of cardano-sl different Scrypt parameters were used).
2. A number of tests mentioned in point 1 were added to AddressDerivationSpec.
3. In integration testing with cardano-node we have demonstrated the following:
(a) we can restore wallet using new endpoint and balance is correct.
(b) we can restore wallet using proper root xprv and wrong encrypted hash and it is expected as recovery depends on root xprv
(c) when restoring with wrong encrypted hash  we can get proper balance but cannot send tx and we get nice error 403 
(d) when restoring wallet with root xprv and proper encrypted hash password we can send tx to cardano-node that is accommodated in block

piotr-iohk:
lgtm, additional test for name update -> https://github.com/input-output-hk/cardano-wallet/pull/1497

Note: Also handy would be a test for updating pass of Xprv restored wallet to add (once pass update is available).

PR #1454, opened 2020-03-18
Byron wallet restoration from xprv endpoint impl

paweljakubas:
# Issue Number

#1436 

# Overview

- [x] I have worked on separating attachPrivateKey into two groups using different encryption
- [x] I have added  implementation of `generateKeyFromMasterKey` in `AddressDerivation.Byron`
- [x] I have added implementation of `postRandomWalletFromXPrv`
- [x] I have added corresponding unit tests
- [ ] I have added illustrative integration tests  (in next PR)

# Comments


lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Byron.hs
```diff
@@ -266,6 +269,21 @@ hdPassphrase masterKey = Passphrase $
...
+
+unsafeGenerateKeyFromMasterKey
+    :: DerivationPath depth
+    -> XPrv
+    -> ByronKey depth XPrv
+unsafeGenerateKeyFromMasterKey derivationPath masterKey = ByronKey
+    { getKey = masterKey
+    , derivationPath
+    , payloadPassphrase = hdPassphrase (toXPub masterKey)
+    }
```
KtorZ, line 285: These functions aren't generating any keys though :grimacing: 
I don't think they are necessary to be honest, the calling code could simply be calling `ByronKey` directly, or, maybe we can simply rename the function to `mkByronKey :: XPrv -> ByronKey 'RootK XPrv`

paweljakubas: yes, fair remarks. `mkByronKey` is better naming

paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/1454/commits/af27d7004db13e835c679bfa6d78a9692ed7b28d

lib/core-integration/src/Test/Integration/Scenario/API/ByronWallets.hs
```diff
@@ -399,6 +407,21 @@ spec = do
...
+        m <- genMnemonics @12
+        let passw = "Secure Passphrase"
+        w <- emptyByronWalletWith ctx "random"
+             ("Byron Wallet", m, passw)
+        rd <- request
+              @ApiByronWallet ctx (Link.deleteWallet @'Byron w) Default Empty
+
+        let (Right seed) = fromMnemonic @'[12] m
+        let _rootXPrv =
+                generateKeyFromSeed seed (Passphrase $ BA.convert $ T.encodeUtf8 passw)
```
KtorZ, line 421: @paweljakubas may you clarify what this call is for :thinking: ? Or is there actually an API call missing from here?

paweljakubas: yeah, this is not finished integration tests. I was waiting for your branch to me merged to use stuff hashing password. As agreed we will do it in next PR

KtorZ: Ah! Damn. I actually removed that from 'encryptPassphrase` because there's actually no reason we would try to encrypt a passphrase with Scrypt on the new codebase :s 

May I suggest to use this passphrase hash instead:

```
31347c387c317c574342652b796362417576356c2b4258676a344a314c
6343675375414c2f5653393661364e576a2b7550766655513d3d7c2f37
6738486c59723174734e394f6e4e753253302b6a65515a6b5437316b45
414941366a515867386539493d
```

It was generated using cardano-sl, it corresponds to the passphrase "patate".

KtorZ (approved):

KtorZ:
Tested locally both unit and integration tests pass. Hydra seems a bit overloaded. Merging this.

PR #1456, opened 2020-03-18
Distinguish (internally) between legacy and new passphrase encryptions

KtorZ:
# Issue Number

#1436 

# Overview

- c433d811172cf0983ceb741a304ea9785958e7f5
  Augment passphrase metadata info with encryptiohttps://github.com/input-output-hk/cardano-wallet/compare/paweljakubas/1436/byron-wallet-restoration-from-xprv-endpoint-impl...KtorZ/1436/encrypt-with-scrypt?expand=1n scheme
  This to allow distinguish legacy passphrase hashes imported from Daedalus that are
encrypted through Scrypt and the new passphrases that we now encrypt with PBKDF2.

- 413498eb1aa0d389ec2877465d243daaa76743a7
  Actually implement 'checkPassphrase' and 'encryptPassphrase' for Scrypt
  This still misses some test vectors to compare with the legacy implementation. I'll
produce that soon using both Daedalus + cardano-sl code.

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
@@ -57,10 +57,11 @@ share
 
 -- Wallet IDs, address discovery state, and metadata.
 Wallet
-    walId                       W.WalletId     sql=wallet_id
-    walCreationTime             UTCTime        sql=creation_time
-    walName                     Text           sql=name
-    walPassphraseLastUpdatedAt  UTCTime Maybe  sql=passphrase_last_updated_at
+    walId                       W.WalletId                sql=wallet_id
+    walCreationTime             UTCTime                   sql=creation_time
+    walName                     Text                      sql=name
+    walPassphraseLastUpdatedAt  UTCTime Maybe             sql=passphrase_last_updated_at
+    walPassphraseScheme         W.PassphraseScheme Maybe  sql=passphrase_scheme
 
     Primary walId
     deriving Show Generic
--- a/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
@@ -119,7 +119,7 @@ import Control.DeepSeq
 import Control.Monad
     ( unless, when )
 import Crypto.Hash
-    ( Digest, HashAlgorithm )
+    ( Blake2b_256, Digest, HashAlgorithm, hash )
 import Crypto.KDF.PBKDF2
     ( Parameters (..), fastPBKDF2_SHA512 )
 import Crypto.Random.Types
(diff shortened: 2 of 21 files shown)


lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
```diff
@@ -57,10 +57,11 @@ share
...
 Wallet
-    walId                       W.WalletId     sql=wallet_id
-    walCreationTime             UTCTime        sql=creation_time
-    walName                     Text           sql=name
-    walPassphraseLastUpdatedAt  UTCTime Maybe  sql=passphrase_last_updated_at
+    walId                       W.WalletId                sql=wallet_id
+    walCreationTime             UTCTime                   sql=creation_time
+    walName                     Text                      sql=name
+    walPassphraseLastUpdatedAt  UTCTime Maybe             sql=passphrase_last_updated_at
+    walPassphraseScheme         W.PassphraseScheme Maybe  sql=passphrase_scheme
```
KtorZ, line 64: I bet this will require some database migration, I need to look into this.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -495,27 +499,60 @@ instance
...
+        let msg = preparePassphrase pwd
+        return $ Hash $ Scrypt.getEncryptedPass $
+            Scrypt.encryptPass Scrypt.defaultParams salt msg
+
+-- | Manipulation done on legacy passphrases before getting encrypted.
+--
+-- Yes there's CBOR. Yes it's hashed before being encrypted. You don't actually
+-- pronounce the 'l' in salmon. A lot of things in the world don't make sense
+-- but we still live just fine.
+preparePassphrase
```
paweljakubas, line 516: probably we will need to export it in the module as we will need it in integration test or even better to expose : 
```
        salt <- Scrypt.Salt <$> getRandomBytes @_ @ByteString 32
        let msg = preparePassphrase pwd
        return $ Hash $ Scrypt.getEncryptedPass $
            Scrypt.encryptPass Scrypt.defaultParams salt msg
```
in a function to avoid importing Scrypt and Cbor in core-integration

KtorZ: We already expose `encryptPassphrase` and `checkPassphrases`, and that's all which is needed :)

paweljakubas (approved):
very good idea to introduce `PassphraseScheme` 
It would be good also to : 
1. update swagger's https://github.com/input-output-hk/cardano-wallet/blob/master/specifications/api/swagger.yaml#L279 if it is not 128 bytes
2. finish integration test showing restoration for byron wallet from master root key works

we will need also to have patch nix stuff.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
paweljakubas: indeed!

Issue #1462, opened 2020-03-19
cardano-node: Cannot send transaction on testnet (cardano-node + cardano-wallet-byron) 


piotr-iohk:
# Context
 

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | cardano-wallet: `2020.3.16 (git revision: 15d6b4e4d6a33fab19550581ddc967d2121ae537)`, cardano-node 1.8 |
| Platform     | Linux        |
| Installation | from source           |

# Steps to Reproduce

0. start cardano-node + wallet on testnet
 - `nix-build -A scripts.testnet.node -o testnet.node.local && testnet.node.local`
 -  `cardano-wallet-byron serve --node-socket ~/wb/cardano-node/state-node-testnet/node.socket --testnet /nix/store/7106qjwral616j6p5049m7hhi5ih0r33-testnet-genesis.json --database ./testnet`

1. Restore wallet with funds.
```
curl  -vX POST http://localhost:8090/v2/byron-wallets \
  -H "Content-Type: application/json; charset=utf-8" \
  -d '{
"style": "random",
"name": "Byron",
"mnemonic_sentence": ["recycle","august","horse","awesome","shuffle","agent","almost","cheap","fury","speak","duty","enemy"],
"passphrase": "Secure Passphrase"
}' --http1.1  |  jq
```
2. Run TX:
```
curl  -vX POST http://localhost:8090/v2/byron-wallets/d3358ac0fc699784af021710e7aba2a3e31fbbd1/transactions \
  -H "Accept: application/json; charset=utf-8" \
  -H "Content-Type: application/json; charset=utf-8" \
  -d '{
	"payments": [
		{
			"address": "37btjrVyb4KFcfEEYyBU8j4VazP5he8iktF5uyD95avRpUHU6aWuLhuhGMmyFiJjtinJQPSjgGmh5w8qmuxDpHjK2JmCnPALuA4dyUkMpHF5Rf4PBd",
			"amount": {
				"quantity": 1,
				"unit": "lovelace"
			}
		}
		],
	"passphrase": "Secure Passphrase"
}' --http1.1  |  jq
```

## Expected behavior

Transaction is successful
## Actual behavior
 
Response:
```
{
  "code": "network_unreachable",
  "message": "The node backend is unreachable at the moment. Trying again in a bit might work."
}


```

cardano-wallet log:
```
[...]

PR #1463, opened 2020-03-19
Resurrect (and upgrade) golden tests for signed txs from http-bridge / cardano-sl

KtorZ:
# Issue Number

#1462 

# Overview

- 5fbb7fdd6f80985438c80d651ba58c00430ca1ca
  resurrect golden integration tests for signed tx from http-bridge
  Turned out to be a dead-end. The golden tests are just fine, but transactions still won't
go through on a testnet-like network

# Comments

At least we have golden tests back now :man_shrugging:


piotr-iohk (approved):
:+1:

PR #1469, opened 2020-03-20
Update test node config for 1.8

Anviking:
# Issue Number

None.

# Overview

- Pass in relative path to the genesis file in the node config
- use shorter tmp dirs to avoid a unix 108 char limit on socket paths
- Disable node forge metric logs which are "critical", but fine.
- set PBftSignatureThreshold: 1, which is the max value according to Duncan

# Comments

### DIFF
--- a/lib/byron/test/data/cardano-node/node.config
+++ b/lib/byron/test/data/cardano-node/node.config
@@ -10,8 +10,9 @@
 
 NodeId:
 NumCoreNodes: 1
-PBftSignatureThreshold: 1000000000000000
+PBftSignatureThreshold: 1
 Protocol: RealPBFT
+GenesisFile: genesis.json
 RequiresNetworkMagic: RequiresNoMagic
 TurnOnLogMetrics: False
 TurnOnLogging: True
@@ -61,6 +62,9 @@ options:
     cardano.node.metrics.ChainDB: []
     cardano.node.metrics: []
     cardano.node.metrics: []
+  mapSubtrace:
+    cardano.node.Forge.metrics:
+      subtrace: NoTrace
 
 # these backends are initialized:
 setupBackends:
--- a/lib/byron/src/Cardano/Wallet/Byron/Launch.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron/Launch.hs
@@ -153,7 +153,7 @@ withConfig tdir action =
               )
     setupConfig = do
         dir <- getCanonicalTemporaryDirectory
-            >>= \tmpRoot -> createTempDirectory tmpRoot "cardano-wallet-byron"
+            >>= \tmpRoot -> createTempDirectory tmpRoot "cw-byron"
 
         let nodeConfigFile   = dir </> "node.config"
         let nodeDatabaseDir  = dir </> "node.db"


lib/byron/test/data/cardano-node/node.config
```diff
@@ -12,6 +12,7 @@ NodeId:
 NumCoreNodes: 1
 PBftSignatureThreshold: 1000000000000000
 Protocol: RealPBFT
+GenesisFile: genesis.yaml
```
KtorZ: genesis has to a JSON file though, or at least, it used to be. We'll to inject a path to the genesis file at runtime, like we do for the start time in the genesis.yaml ---> https://github.com/input-output-hk/cardano-wallet/blob/master/lib/byron/test/integration/Cardano/Wallet/Byron/Config.hs#L159-L161

Anviking: Yeah, but I think we'll be fine with a relative `genesis.json`?

lib/byron/test/integration/Cardano/Wallet/Byron/Config.hs
```diff
@@ -145,8 +145,7 @@ withConfig action =
               , (BlockchainParameters, NodeVersionData)
               )
     setupConfig = do
-        dir <- getCanonicalTemporaryDirectory
-            >>= \tmpRoot -> createTempDirectory tmpRoot "cardano-wallet-byron"
+        let dir = "/tmp/cardano-wallet-byron"
```
Anviking: The node was complaining about the socket path being too long. This is temporary.

KtorZ: > The node was complaining about the socket path being too long. This is temporary.

What o.O ?

KtorZ: Could you report (i.e. open a GH ticket) that to ouroboros-network ?

Anviking: 👍 I'll verify I get the same error and report.

Anviking:
```
{-# LINE 1062 "Network/Socket/Types.hsc" #-}
unixPathMax :: Int
unixPathMax = 108
{-# LINE 1064 "Network/Socket/Types.hsc" #-}

{-# LINE 1065 "Network/Socket/Types.hsc" #-}

-- We can't write an instance of 'Storable' for 'SockAddr' because
-- @sockaddr@ is a sum type of variable size but
-- 'Foreign.Storable.sizeOf' is required to be constant.

-- Note that on Darwin, the sockaddr structure must be zeroed before
-- use.

-- | Write the given 'SockAddr' to the given memory location.
pokeSockAddr :: Ptr a -> SockAddr -> IO ()

{-# LINE 1076 "Network/Socket/Types.hsc" #-}
pokeSockAddr p sa@(SockAddrUnix path) = do
    when (length path > unixPathMax) $ error "pokeSockAddr: path is too long"
    zeroMemory p $ fromIntegral $ sizeOfSockAddr sa
```

I suspect there's nothing they could do. Won't report.

https://unix.stackexchange.com/questions/367008/why-is-socket-path-length-limited-to-a-hundred-chars

Changing the `cardano-wallet-byron` subdir to `cw-byron` seems to do it though.

lib/byron/test/data/cardano-node/node.config
KtorZ: No, not really... Right now it might works because the PR removing the `--genesis-file` override from the command-line is not part of 1.8.0 but after that, the genesis file is actually generated dynamically and put in temporary location. The node isn't started from this same temporary directory thus it won't probably find the genesis file if specified relatively. 

Your change is however sufficient to get us going with 1.8.0 for it fixes the parsing issue and the value is overridden anyway by the command-line option.

Anviking: > `--genesis-file`

Right, that explains why it works.

lib/byron/test/data/cardano-node/node.config
```diff
@@ -10,8 +10,9 @@
 
 NodeId:
 NumCoreNodes: 1
-PBftSignatureThreshold: 1000000000000000
+PBftSignatureThreshold: 1
```
KtorZ, line 13: Why this change? This was actually necessary to get the self node keeps producing block alone.

Anviking: Duncan said 1 was the max value, and it seemed to work just as well with 1.

> to get the self node keeps producing block alone.

It didn't produce blocks otherwise? Now it seems to at least.

PR #1473, opened 2020-03-22
Use less verbose QuickCheck helpers in tests

jhrcek:
# Issue Number

N/A

# Overview

Using less verbose QuickCheck helpers where applicable. In particular:
- `vectorOf n arbitrary == vector n`
- `oneof [pure x, pure y, ..] == elements [x, y, ..]`
- `X <$> arbitrary <*> arbitrary <*> arbitrary == applyArbitrary3 X`
- other minor simplifications

# Comments

### DIFF
--- a/lib/text-class/test/unit/Data/Text/ClassSpec.hs
+++ b/lib/text-class/test/unit/Data/Text/ClassSpec.hs
@@ -38,11 +38,13 @@ import Test.QuickCheck
     ( Arbitrary (..)
     , UnicodeString (..)
     , arbitraryBoundedEnum
+    , arbitrarySizedNatural
     , choose
     , classify
     , elements
     , genericShrink
     , property
+    , shrinkIntegral
     , vectorOf
     , (===)
     , (==>)
@@ -71,8 +73,6 @@ spec = do
                     <> show (maxBound @Int)
                     <> "."
             in fromText @Int "patate" === Left (TextDecodingError err)
-        it "fromText . toText === pure"
-            $ property $ \(i :: Int) -> (fromText . toText) i === pure i
         it "fromText ~ fromTextMaybe" $
             property $ \(Digits t) ->
                 classify (isNothing (fromTextMaybe @Int t))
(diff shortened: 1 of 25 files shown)


lib/text-class/test/unit/Data/Text/ClassSpec.hs
```diff
@@ -71,8 +73,6 @@ spec = do
                     <> show (maxBound @Int)
                     <> "."
             in fromText @Int "patate" === Left (TextDecodingError err)
-        it "fromText . toText === pure"
```
jhrcek, line 74: Replaced these 3 tests with 3 functionally equivalent `textRoundtrip` cases below.

lib/text-class/test/unit/Data/Text/ClassSpec.hs
```diff
@@ -159,14 +157,8 @@ instance Arbitrary Digits where
...
 instance Arbitrary Natural where
-    shrink x =
-        [ fromIntegral x
-        | x' <- shrink (fromIntegral x)
-        , x' >= 0
-        , x' /= x
-        ]
-    arbitrary =
-        fromIntegral . abs <$> arbitrary @Integer
+    shrink = shrinkIntegral
```
rvl, line 160: Does this work without arithmetic underflow exceptions? I notice in the definition of `shrinkIntegral x` there is `-x`.

rvl (commented):
Nice clean ups, thanks! :+1: 
Just one question...

lib/text-class/test/unit/Data/Text/ClassSpec.hs
jhrcek: Yes, I double checked the implementation of [shrinkIntegral](http://hackage.haskell.org/package/QuickCheck-2.13.2/docs/src/Test.QuickCheck.Arbitrary.html#shrinkIntegral) carefully before doing this change. The `-x` is negating those `x`-es which are negative in the first place -> makes them positive. But when shrinking Naturals there won't be any negative x-es in the first place, so that code path won't even be taken. The additional confirmation that I'm doing the right thing was the same implementation of `Arbitrary Natural` is used in [quickcheck-instances](http://hackage.haskell.org/package/quickcheck-instances-0.3.22/docs/src/Test.QuickCheck.Instances.Natural.html#line-19)

rvl (approved):

rvl:
bors try

rvl:
@jhrcek You won't be able to see it, but the build failure was simply because stylish-haskell needs to be run over the code. If you do that then it should pass.

jhrcek:
Thank you, I reformatted the code via stylish-haskell. Love the attention to details in this codebase :)

Issue #1446, opened 2020-03-17
Faling test case: list byron transactions


KtorZ:
# Context
 
 

# Test Case

https://buildkite.com/input-output-hk/cardano-wallet/builds/6328#978398e3-7432-4507-8599-0777fbfe2292/6-7753

From the logs, the only error available which seems to match the test is the following:

```
[cardano-wallet.wallet-db:Info:11923] [2020-03-16 17:55:41.05 UTC] Waiting for 1 withDatabase dd66acf0...78024d2a call(s) to finish
[cardano-wallet.wallet-engine:Notice:12037] [2020-03-16 17:55:41.05 UTC] Worker has exited: main action is over.
[cardano-wallet.wallet-engine:Error:11707] [2020-03-16 17:55:41.07 UTC] Worker has exited unexpectedly: SQLite3 returned ErrorBusy while attempting to perform close: unable to close due to unfinalized statements or unfinished backups
[cardano-wallet.wallet-db:Info:11923] [2020-03-16 17:55:41.10 UTC] Database dd66acf0...78024d2a is ready to be deleted
[cardano-wallet.wallet-db:Info:11923] [2020-03-16 17:55:41.10 UTC] Removing wallet's database. Wallet id was dd66acf0...78024d2a
```

This is kinda weird and I don't think particularly connected to this test. 

# Failure / Counter-example

---

# Resolution 

---

# QA

piotr-iohk:
Another occurence -> https://buildkite.com/input-output-hk/cardano-wallet/builds/6400#62c6c6c5-a54c-4a09-9e8c-6fddba5966c2

PR #1474, opened 2020-03-23
Mark wallet as "not_responding" when underlying workers die

KtorZ:
# Issue Number

#1292 (in particular, last comment: https://github.com/input-output-hk/cardano-wallet/issues/1292#issuecomment-581822625)

# Overview

- 732c6f4ad76ebef0c595beb78cea589567f9c5d0
  :round_pushpin:**move list databases to enable server calling it from API handlers**
  When a wallet worker dies, it is removed from the registry and therefore, is no longer
available for listing. Yet, the file might still exist on disk, so looking at the database
files is more reliable than looking at the registered workers. Should a worker be dead
however, operation on a wallet should be degraded

- b5f560be8667267144f0a75b4a70c8c58626af4b
  :round_pushpin:**create a new 'ErrDeadWallet' to capture dead workers**
  
- da31d825a2d35402b36483c6053fc8c951170972
  :round_pushpin:**list workers through known databases and allow recovering from a dead worker**
  With this, we could still return wallet as degraded when listing and fetching them independently. At
least, it indicates client that something went wrong and despite being stuck, wallets remains accessible.

- 3e499186bc2e0e94762fec82480b88062db5bf11
  :round_pushpin:**marks wallets as 'dead' when fetching them despite a worker being dead**
  
- 33ce4bbfd189e9df1f2c79b39fe952058133747a
  :round_pushpin:**rename 'dead' to 'not_responding'**
  Because 'dead' is a bad word.
 
# Comments

Kinda hard to test because workers ain't supposed to die in practice :grimacing: ...
So, I injected a temporary fault in the restoration loop to make the worker dies immediately and it works as expected:

<details>
    <summary>curl http://localhost:8090/v2/byron-wallets | jq</summary>

```json
[
  {
    "passphrase": {
      "last_updated_at": "2020-03-23T11:36:12.582891065Z"
    },
    "state": {
      "status": "dead"
    },
    "balance": {
      "total": {
        "quantity": 0,
        "unit": "lovelace"
      },
      "available": {
        "quantity": 0,
        "unit": "lovelace"
      }
    },
    "name": "MyWallet",
    "id": "adc78398110d92af3132488546c6977f61dc0c7e",
    "tip": {
      "height": {
        "quantity": 0,
        "unit": "block"
      },
      "epoch_number": 0,
      "slot_number": 0
    }
  }
]
```
</details>

<details>
    <summary>curl http://localhost:8090/v2/byron-wallets/{walletId} | jq</summary>

```json
{
  "passphrase": {
    "last_updated_at": "2020-03-23T11:36:12.582891065Z"
  },
  "state": {
    "status": "dead"
  },
  "balance": {
    "total": {
      "quantity": 0,
      "unit": "lovelace"
    },
    "available": {
      "quantity": 0,
      "unit": "lovelace"
    }
  },
  "name": "MyWallet",
  "id": "adc78398110d92af3132488546c6977f61dc0c7e",
  "tip": {
    "height": {
      "quantity": 0,
      "unit": "block"
    },
    "epoch_number": 0,
    "slot_number": 0
  }
}
```
</details>

<details>
    <summary>Anything else</summary>

```json
{
  "code": "wallet_is_dead",
  "message": "That's embarassing. My associated worker for adc78398110d92af3132488546c6977f61dc0c7e is no longer responding. This is not something that is supposed to happen. The worker must have left a trace in the logs of severity 'Error' when it died which might explain the cause. Said differently, this wallet won't be accessible until the server is restarted but there are good chances it'll recover itself upon restart."
}
```
</details>

### DIFF
--- a/lib/core/src/Cardano/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/DB/Sqlite.hs
@@ -374,7 +374,14 @@ data DBLog
     | MsgManualMigrationNeeded DBField Text
     | MsgManualMigrationNotNeeded DBField
     | MsgUpdatingForeignKeysSetting ForeignKeysSetting
+    | MsgFoundDatabase FilePath Text
+    | MsgUnknownDBFile FilePath
     deriving (Generic, Show, Eq, ToJSON)
+
+
+{-------------------------------------------------------------------------------
+                                    Logging
+-------------------------------------------------------------------------------}
 
 data DBField where
     DBField
@@ -442,6 +449,8 @@ instance DefineSeverity DBLog where
         MsgManualMigrationNeeded{} -> Notice
         MsgManualMigrationNotNeeded{} -> Debug
         MsgUpdatingForeignKeysSetting{} -> Debug
+        MsgFoundDatabase _ _ -> Info
+        MsgUnknownDBFile _ -> Notice
 
 instance ToText DBLog where
     toText = \case
@@ -493,6 +502,12 @@ instance ToText DBLog where
             [ "Updating the foreign keys setting to: "
             , T.pack $ show value
             , "."
+            ]
+        MsgFoundDatabase _file wid ->
+            "Found existing wallet: " <> wid
+        MsgUnknownDBFile file -> mconcat
+            [ "Found something other than a database file in "
+            , "the database folder: ", T.pack file
             ]
 
 {-------------------------------------------------------------------------------
--- a/lib/core/src/Cardano/Wallet/Primitive/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types.hs
@@ -1283,11 +1283,14 @@ data SlotParameters = SlotParameters
 data SyncProgress
     = Ready
     | Syncing !(Quantity "percent" Percentage)
+    | Dead
     deriving (Generic, Eq, Show)
 
 instance NFData SyncProgress
 
 instance Ord SyncProgress where
+    Dead <= _ = True
+    _ <= Dead = False
     Ready <= Ready = True
     Ready <= Syncing _ = False
     Syncing _ <= Ready = True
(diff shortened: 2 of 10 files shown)


KtorZ:
bors try

lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -442,6 +449,8 @@ instance DefineSeverity DBLog where
         MsgManualMigrationNeeded{} -> Notice
         MsgManualMigrationNotNeeded{} -> Debug
         MsgUpdatingForeignKeysSetting{} -> Debug
+        MsgFoundDatabase _ _ -> Info
+        MsgUnknownDBFile _ -> Notice
```
paweljakubas, line 453: wonder if unknown db file should not be treated more severely. Is it not critical?

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -1283,11 +1283,14 @@ data SlotParameters = SlotParameters
...
     = Ready
     | Syncing !(Quantity "percent" Percentage)
+    | Dead
     deriving (Generic, Eq, Show)
 
 instance NFData SyncProgress
 
 instance Ord SyncProgress where
+    Dead <= _ = True
+    _ <= Dead = False
```
paweljakubas: could you give rationale behind those ^^^ ?

paweljakubas (approved):
Looks fine

lib/core/src/Cardano/DB/Sqlite.hs
KtorZ: It is not critical as the file is just ignored. It won't even prevent the system from working just fine, hence the `Notice`.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
KtorZ: We needed an ordering ^^" ... So I thought we might as well just say that "Dead" is smaller than anything else. That's purely arbitrary.

paweljakubas: yeah, I know we need ordering, just wanted if there is some thinking behind this that will address some edge case, etc.

KtorZ: Nope, purely arbitrary here :man_shrugging:

KtorZ:
Another occurence of: #1446

PR #1475, opened 2020-03-23
make NETWORK_BYRON work on cardano-node and jormungandr

piotr-iohk:
# Issue Number

# Overview

- 791bc1960b1952577a271084e3be2d150c725dba
  remove waitForNextEpoch from NETWORK_BYRON to make it work on cardano-node

# Comments

The test was failing on cardano-node as it was using `waitForNextEpoch`. I removed it and put the whole action + verification into `eventually`, so hopefully it is not flaky on neither cardano-node nor jormungandr.


KtorZ (approved):
:thinking: ... It should work yes.

Issue #1477, opened 2020-03-23
Refactor cardano-cli to re-use commands for both Byron & Shelley endpoints


KtorZ:
# Context

Commands in cardano-cli currently "suffer" from two issues:

a) They require a type-level network discriminant which must be provided **at compile-time**. As a direct consequence, it is not yet possible to provide a single command-line that works with two different network discriminations.

b) They are hard-wired to call the "Shelley" endpoints of the wallet, which in the Byron setup, all returns "501 Not Implement". However, the underlying APIs are _mostly_ the same (except for wallet creation). 

# Decision

- Refactor the command-line setup to work with both Byron and Shelley endpoint
- Refactor the command-line setup to accept network discriminant to be set at runtime
- Keep cardano-wallet-jormungandr command-line unchanged, as much as possible

# Acceptance Criteria

1. cardano-wallet-byron _should_ offer mostly the same functionality as Jormungandr transactions and wallet commands. 

2. The creation of wallet _should_ however be different and adjusted for byron.

3. cardano-wallet-jormungandr _should_ remain unchanged, from a user-perspective.

---

# Development 

- Refactor command-line ==> #1478
- Implement address endpoints for Byron + address command-line ==> #1493 

# QA 

- Integration tests covering transaction manipulation via CLI --> https://github.com/input-output-hk/cardano-wallet/blob/master/lib/byron/test/integration/Test/Integration/Byron/Scenario/CLI/Transactions.hs
- Superficial manual tests done for address commands here: 

<details>
    <summary>cardano-wallet address create --help</summary>

```
Usage: cardano-wallet-byron address create [--port INT] [--address-index INDEX]
                                           WALLET_ID
  Create a new random address. Only available for random wallets. The address
  index is optional, give none to let the wallet generate a random one.

Available options:
  -h,--help                Show this help text
  --port INT               port used for serving the wallet API. (default: 8090)
  --address-index INDEX    A derivation index for the address
```
</details>

<details>
    <summary>cardano-wallet address list adc78398110d92af3132488546c6977f61dc0c7e</summary>

```
Ok.
[]
```
</details>

<details>
    <summary>cardano-wallet address create adc78398110d92af3132488546c6977f61dc0c7e</summary>

```
Please enter your passphrase: ****
passphrase is too short: expected at least 10 characters

Please enter your passphrase: ****************
[...]

PR #1478, opened 2020-03-23
Support for wallet & transactions commands in cardano-wallet-byron

KtorZ:
# Issue Number

#1477 

# Overview

- 12339ede10eab98e6543650b5a2ed0aec3ded35b
  :round_pushpin: **capture network discriminant from command-line options**
  This enables mounting commands in the command-line to work with both
mainnet and testnet; chosen at runtime. This commit won't work properly
with Jörmungandr and need to be adjusted to work fine there too.

- 403e031c298e9031bcf8c3dde9222de0e2fd6755
  :round_pushpin: **Allow caller to choose how to parse network discriminant**
  This enables correct support of the network discriminant in the Byron CLI while,
at the same time, keep the interface for Jörmungandr unchanged by forcing the
network discriminant to testnet, as before.

- 29081d293264af115b39341735773a1352f2ce51
  :round_pushpin: **wire the new discriminant option in existing command-lines**
  
- 9b5a89f61273e7e5dbeef882104175833374e00f
  :round_pushpin: **break 'walletClient' into several smaller clients**
  This enables to provide different clients for each commands, and in particular
to define and make use of clients for byron wallet endpoints where needed.
There's still an annoying bit regarding the creation of wallets that differs
between Byron and new ones. This will be addressed in another commit

- 3a33b6e6e91a61818ffdc6d90e91cea6e4ea20cf
  :round_pushpin: **remove now obsolete discriminant flag for wallet commands**
  The wallet client split introduced in the previous commmit makes this option now
pointless. Indeed, it was needed to instantiate a (WalletClient n) in the past
because of the addresses, transactions and stake pools endpoints. Now that the
wallet client is only about wallets, the network discrimination matters not.

- a3ecef20116f62133c59a707349be521c5163622
  :round_pushpin: **implement cmdByronWalletCreate for creating wallets from mnemonic**
  This can be extended later on to create a wallet from a raw XPrv. Didn't do it yet though
as the endpoint is still under testing.

- 4e444482e50b23b7a05cab6d89a5ef6363c67f27
  :round_pushpin: **Fix postTransactionFeeViaCLI (which is not interactive CLI)**
  
- 7ee58476744dd25f743b2ba6bdbdbbd3f4143d75
  :round_pushpin: **Integration tests for transactions via CLI for cardano-node**
  
- 02557e613b0dbc706e22cc4901577b81bfe6966e
  :round_pushpin: **use Haskell magic and strong _family_ bonds to remove the need for passing discriminant as option**
  The main trick here was to make the 'Api' type poly-kinded and then allow some of the problematic types
to be defined as family of the given type parameter. This enables us to use more flexible data-types in
the Haskell typed client (like plain Aeson.Value) since the client is in the end simply returning JSON.

what we loose however by doing this is the ability to catch _early_ (i.e. in the command-line client code)
that an address may be of the wrong discriminant. This is deferred until the server sees the address and
reject the API call. I think this is really acceptable.

- e855f4fd5e276eb9e1b61ceb07953b68a8a594de
  :round_pushpin: **Fix protocol magic not being verified when decoding addresses**
  Pretty big deal and which explains why the wallet was actually crashing when submitting an address that
was illed-formed and the API not catching it.

- aedcd00e252abeeb8b66df52d6163fa81727be58
  :round_pushpin: **defer network discriminant check from command-line client to the API**
  This way, we need not to know or pass the network discriminant as an option when parsing payments
from the command-line. If an address has an invalid network discrimination, the API will reject it

- c52a7d8a6c7048add3880a013cef70d4eaaf4779
  :round_pushpin: **more integration tests for list/forget tx**
  
- e6e04230010c3ce898bb937875ccfb98d1774d3e
  :round_pushpin: **do not enforce network discriminant for legacy address on the ITN...**
  
- f243a7de02a0aae4b89cdf08e9af780b49f7b21b
  :round_pushpin: **re-generate nix machinery**

# Comments

:warning: I've only tested this superficially. So I'd be keen not to merge that before this week's release but only after while we work out the integration tests. Most of existing scenarios should be portable (modulo the endpoints that aren't yet implemented on the byron API).

### DIFF
--- a/lib/byron/cardano-wallet-byron.cabal
+++ b/lib/byron/cardano-wallet-byron.cabal
@@ -170,6 +170,7 @@ test-suite cardano-node-integration
     , cardano-wallet-core-integration
     , cardano-wallet-launcher
     , cardano-wallet-test-utils
+    , command
     , generic-lens
     , hspec
     , http-client
@@ -189,3 +190,4 @@ test-suite cardano-node-integration
   other-modules:
       Cardano.Wallet.Byron.Faucet
       Test.Integration.Byron.Scenario.API.Transactions
+      Test.Integration.Byron.Scenario.CLI.Transactions
--- a/lib/byron/exe/cardano-wallet-byron.hs
+++ b/lib/byron/exe/cardano-wallet-byron.hs
@@ -38,8 +38,11 @@ import Cardano.CLI
     , cmdKey
     , cmdMnemonic
     , cmdNetwork
+    , cmdTransaction
     , cmdVersion
+    , cmdWallet
     , databaseOption
+    , defaultDiscriminantOption
     , enableWindowsANSI
     , helperTracing
     , hostPreferenceOption
@@ -144,6 +147,8 @@ main = withUtf8Encoding $ do
         <> cmdServe
         <> cmdMnemonic
         <> cmdKey
+        <> cmdWallet defaultDiscriminantOption
+        <> cmdTransaction defaultDiscriminantOption
         <> cmdNetwork
         <> cmdVersion
 
--- a/lib/byron/exe/cardano-wallet-byron.hs
+++ b/lib/byron/exe/cardano-wallet-byron.hs
@@ -150,10 +150,7 @@ main = withUtf8Encoding $ do
         <> cmdServe
         <> cmdMnemonic
         <> cmdKey
-        <> cmdWallet
-            defaultDiscriminantOption
-            cmdByronWalletCreate
-            byronWalletClient
+        <> cmdWallet cmdByronWalletCreate byronWalletClient
         <> cmdTransaction
             defaultDiscriminantOption
             byronTransactionClient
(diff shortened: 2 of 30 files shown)


KtorZ:
bors try

piotr-iohk (approved):
Added integration tests for tx create/list/forget
Manually tested CRUD (create / get / delete / list wallets)

There are integration tests missing for CLI CRUD and transaction submit (can be added in another PR)

KtorZ:
_running `$ stack test cardano-wallet-byron:cardano-node-integration` locally_

KtorZ:
bors try

KtorZ:
bors try

KtorZ:
```
stack test cardano-wallet-byron:cardano-node-integration
[...]
Finished in 402.7320 seconds
123 examples, 0 failures
```

KtorZ:
bors try

PR #1479, opened 2020-03-23
Tmp: use nullTracer for network layer

Anviking:
# Issue Number

A bit related to #1398 

# Overview

- Need windows artefact from CI for this

# Comments


Anviking:
Proper fix in #1484

PR #1482, opened 2020-03-24
Re-add old wallet restoration benchmark

Anviking:
# Issue Number

#1458 

# Overview

- [x] Restore old benchmark measuring the time it takes to restore wallets.
- [x] Make benchmark launch a cardano-node

# Comments
Successful run:
```
All results:
--
  | restore mainnet seq: 1016 s
  | restore mainnet 1% ownership: 3571 s
  | restore mainnet 2% ownership: 6687 s
```

https://buildkite.com/input-output-hk/cardano-wallet/builds/6536#e9097620-b59a-4b2d-9c44-c41982eea313

### DIFF
--- a/lib/byron/bench/Restore.hs
+++ b/lib/byron/bench/Restore.hs
@@ -75,7 +75,7 @@ import Cardano.Wallet.Primitive.AddressDiscovery.Any.TH
 import Cardano.Wallet.Primitive.AddressDiscovery.Random
     ( RndState, mkRndState )
 import Cardano.Wallet.Primitive.Model
-    ( currentTip, totalBalance, totalUTxO )
+    ( currentTip, totalUTxO )
 import Cardano.Wallet.Primitive.Types
     ( Address
     , Block (..)
@@ -84,9 +84,10 @@ import Cardano.Wallet.Primitive.Types
     , ChimericAccount
     , SlotId (..)
     , SyncProgress (..)
-    , UTxO (..)
     , WalletId (..)
     , WalletName (..)
+    , computeUtxoStatistics
+    , log10
     , mkSyncTolerance
     , slotAt
     , slotParams
@@ -123,7 +124,7 @@ import Data.Time.Clock.POSIX
 import Database.Persist.Sql
     ( runMigrationSilent )
 import Fmt
-    ( fmt, pretty, (+|), (+||), (|+), (||+) )
+    ( build, fmt, pretty, (+|), (+||), (|+), (||+) )
 import Ouroboros.Network.NodeToClient
     ( NodeToClientVersionData (..) )
 import Say
@@ -151,7 +152,6 @@ import qualified Cardano.Wallet.DB.Sqlite as Sqlite
 import qualified Cardano.Wallet.Primitive.AddressDerivation.Byron as Byron
 import qualified Data.ByteString.Char8 as B8
 import qualified Data.List.NonEmpty as NE
-import qualified Data.Map.Strict as Map
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
 
@@ -335,9 +335,8 @@ bench_restoration tracer socketPath progressLogFile (wid, wname, s) = do
                 waitForWalletSync @n w wallet
                 (wallet', _, pending) <- unsafeRunExceptT $ W.readWallet w wid
                 sayErr "Wallet restored!"
-                sayErr . fmt $ "Balance: " +|| totalBalance pending wallet' ||+ " lovelace"
-                sayErr . fmt $
-                    "UTxO: " +|| Map.size (getUTxO $ totalUTxO pending wallet') ||+ " entries"
+                sayErr . fmt . build $
+                    computeUtxoStatistics log10 (totalUTxO pending wallet')
                 unsafeRunExceptT $ W.deleteWallet w wid
 
 traceProgressForPlotting :: Tracer IO Text -> Tracer IO WalletLog
--- a/lib/byron/bench/Restore.hs
+++ b/lib/byron/bench/Restore.hs
@@ -1,0 +1,460 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedLabels #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+module Main where
+
+import Prelude
+
+import Cardano.BM.Configuration.Static
+    ( defaultConfigStdout )
+import Cardano.BM.Data.Severity
+    ( Severity (..) )
+import Cardano.BM.Setup
+    ( setupTrace_ )
+import Cardano.BM.Trace
+    ( Trace, nullTracer )
+import Cardano.DB.Sqlite
+    ( destroyDBLayer, unsafeRunQuery )
+import Cardano.Launcher
+    ( Command (..), StdStream (..), withBackendProcess )
+import Cardano.Startup
+    ( installSignalHandlers )
+import Cardano.Wallet
+    ( WalletLayer (..), WalletLog (..) )
+import Cardano.Wallet.Byron.Compatibility
+    ( Byron
+    , KnownNetwork (blockchainParameters, versionData)
+    , emptyGenesis
+    , fromByronBlock
+    , fromNetworkMagic
+    )
+import Cardano.Wallet.Byron.Network
+    ( withNetworkLayer )
+import Cardano.Wallet.Byron.Transaction
+    ( newTransactionLayer )
+import Cardano.Wallet.Byron.Transaction.Size
+    ( MaxSizeOf (..) )
+import Cardano.Wallet.DB
+    ( DBLayer )
+import Cardano.Wallet.DB.Sqlite
+    ( PersistState, newDBLayer )
+import Cardano.Wallet.Logging
+    ( trMessageText )
+import Cardano.Wallet.Network
+    ( FollowLog (..), NetworkLayer (..) )
+import Cardano.Wallet.Primitive.AddressDerivation
+    ( Depth (..)
+    , NetworkDiscriminant (..)
+    , NetworkDiscriminantVal (..)
+    , PersistPrivateKey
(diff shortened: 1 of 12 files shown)


nix/haskell.nix
```diff
@@ -85,6 +85,24 @@ let
...
 
+
+        # TODO: How do we bundle the config in the derivation, and
+        # do we even want to do this? Maybe there are better alternatives.
+        packages.cardano-wallet-byron.components.benchmarks.restore = {
+           build-tools = [ pkgs.makeWrapper ];
+           preBuild = ''
+             cat ${src + /lib/byron/test/data/bench-mainnet-configuration.yaml}
+             ls $out
+             cp ${src + /lib/byron/test/data/bench-mainnet-configuration.yaml} $out/config.yaml
```
Anviking: We — probably — need a custom config. 

1. To stop the node trying to write logs the the working directory where it doesn't have permissions
2. To disable the node's LiveView GUI

I suspect something like this could be done, but this in particular didn't work

Anviking: @rvl Does bundling the config like this sound like a sane idea to begin with, and do you know how to? (I suspect it's simple)

rvl: We could include our own logging config, but in this case I don't think we need to.

The liveview stuff probably won't affect the benchmark so it can stay. The other logging settings provided by cardano-node might help with debugging the benchmark.

The "trick" I used in cardano-launcher is to change working directory to somewhere where cardano-node is permitted to write its logs. So remove the `--run "cd $src"` line from below, because that forces the cwd to be a read-only location.

lib/byron/bench/Restore.hs
```diff
@@ -0,0 +1,419 @@
...
+
+
+    -- TODO: We might want to support testnet
+    let network = Mainnet
+
+    -- Reading setup from env vars.
+    --
+    -- TODO: Look over whether this is sane or not.
+    configs <- getEnv "BYRON_CONFIGS"
+    nodeDB <- getEnv "NODE_DB"
```
rvl, line 184: Maybe it would be easiest to set nodeDB to something like `./db` and socketPath to `./cardano-node.socket`. The logs will get dropped in `./logs` anyway. Then the script which executes the benchmark can take care of changing to the correct directory.

Anviking: I found it useful to be able to set NODE_DB when running locally, without nix.

`./cardano-node.socket` sounds sane though.

lib/byron/bench/Restore.hs
```diff
@@ -0,0 +1,468 @@
...
+main = do
+    hSetBuffering stdout NoBuffering
+    hSetBuffering stderr NoBuffering
+
+    (_logCfg, tr) <- initBenchmarkLogging Info
+    installSignalHandlers (return ())
+
+    -- TODO: We might want to support testnet via a
+    -- `--benchmark-arguments "testnet"`
+    let network = Mainnet
```
Anviking, line 172: I think this can go into a separate PR.

Anviking:
Btw, empty wallet, vs 1% vs 2% owned addresses using the logs from buildkite
Skärmavbild 2020-03-27 kl  13 15 59

lib/byron/bench/Restore.hs
```diff
@@ -0,0 +1,468 @@
...
+    -- it can be useful to set
+    --
+    --     export BYRON_CONFIGS=$CARDANO_NODE_DIR/configuration
+    --     export NODE_DB=$CARDANO_NODE_DIR/db
+    --
+    -- before running this benchmark manually with stack.
+    configs <- getEnv "BYRON_CONFIGS"
+    nodeDB <- getEnv "NODE_DB"
+
+    let genesisHash = "5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb"
```
KtorZ: Note that this is already hard-coded in `Cardano.Wallet.Byron.Compatibility`, would be good limit the number of places where this gets written literally. 

```hs
let genesisHash = getGenesisBlockHash $ blockchainParameters @'Mainnet
```

lib/byron/bench/Restore.hs
```diff
@@ -0,0 +1,468 @@
...
+    -- however.
+    let cmd = Command "cardano-node" args (return ()) NoStream NoStream
+    void $ withBackendProcess nullTracer cmd $ do
+        case network of
+            Testnet _ -> error "Testnet: not supported yet"
+            Mainnet -> do
+                prepareNode (Proxy @'Mainnet) socketPath
+                runBenchmarks
+                    [ bench ("restore " <> "mainnet" <> " seq")
+                        (bench_restoration @'Mainnet @ShelleyKey
```
KtorZ: Doh! ShelleyKey aren't actually a thing on the Byron network. Byron network only uses `ByronKey` and `IcarusKey`! 
Benchmarks should be made for these two! The % one can be limited to only the IcarusKey though.

lib/byron/bench/Restore.hs
```diff
@@ -0,0 +1,468 @@
...
+                        tl
+                        db
+                wallet <- unsafeRunExceptT $ W.createWallet w wid wname s
+                void $ forkIO $ unsafeRunExceptT $ W.restoreWallet @_ @s @t @k w wid
+                waitForWalletSync @n w wallet
+                (wallet', _, pending) <- unsafeRunExceptT $ W.readWallet w wid
+                sayErr "Wallet restored!"
+                sayErr . fmt $ "Balance: " +|| totalBalance pending wallet' ||+ " lovelace"
+                sayErr . fmt $
+                    "UTxO: " +|| Map.size (getUTxO $ totalUTxO pending wallet') ||+ " entries"
```
KtorZ: What about printing the UTxO statistics also here? To get an overview of the distribution.

lib/byron/bench/Restore.hs
```diff
@@ -0,0 +1,468 @@
...
+        slotDur = 20
+        epochDur = slotDur * 21600
+        ep = floor (d / epochDur)
+        idx = floor ((d - (fromIntegral ep) * epochDur) / slotDur)
+
+    startTime :: MonadFail m => m POSIXTime
+    startTime = pure
+        $ (\(StartTime t) -> utcTimeToPOSIXSeconds t)
+        $ getGenesisBlockDate
+        $ blockchainParameters @n
```
KtorZ: I assume this function was taken from the old http-bridge implementation. We know have:

```
slotAt :: SlotParameters -> UTCTime -> Maybe SlotId
```

in `Cardano.Wallet.Primitive.Types` which can get us exactly the same result, what about using this instead of duplicating the logic :pray:  ?

lib/byron/bench/Restore.hs
```diff
@@ -0,0 +1,468 @@
...
+
+logQuiet :: SlotId -> SlotId -> IO ()
+logQuiet _ _ = pure ()
+
+instance MaxSizeOf Address 'Mainnet ShelleyKey where
+    maxSizeOf = error "maxSizeOf should not be called by restore benchmark"
```
KtorZ: It's more that "ShelleyKey" shouldn't be used in restoration benchmark :upside_down_face:

lib/byron/bench/Restore.hs
```diff
@@ -0,0 +1,468 @@
...
+    -> FilePath
+       -- ^ Socket path
+    -> FilePath
+       -- ^ Log output
+    -> (WalletId, WalletName, s)
+    -> IO ()
+bench_restoration tracer socketPath progressLogFile (wid, wname, s) = do
+    let bp = blockchainParameters @n
+    let vData = versionData @n
+    let network = "mainnet" :: String
```
KtorZ:
```suggestion
    let network = networkDiscriminantVal @n
```

lib/byron/bench/Restore.hs
Anviking: 👍 6632d448cfdc1fdedb7353cf7140aa77937d91f0

lib/byron/bench/Restore.hs
Anviking: 6632d448cfdc1fdedb7353cf7140aa77937d91f0

lib/byron/bench/Restore.hs
Anviking: 5dcfbfc4dcd3d375c9ada680eb7d69e3f809acda

lib/byron/bench/Restore.hs
Anviking: ebf7978769b6876b7c31db914e41777780a058cc

lib/byron/bench/Restore.hs
Anviking: 885ffee71e4e53fc0113ed19dcd2be872f242c59

lib/byron/bench/Restore.hs
```diff
@@ -0,0 +1,460 @@
...
+
+{-------------------------------------------------------------------------------
+                                  Benchmarks
+-------------------------------------------------------------------------------}
+
+{-# ANN bench_restoration ("HLint: ignore Use camelCase" :: String) #-}
+bench_restoration
+    :: forall (n :: NetworkDiscriminant) (k :: Depth -> * -> *) s t.
+        ( IsOurs s Address
+        , IsOurs s ChimericAccount
```
KtorZ, line 289: Why is this constraint needed ?

lib/byron/bench/Restore.hs
```diff
@@ -0,0 +1,460 @@
...
+            rngSeed = 0
+            s = mkRndState xprv rngSeed
+        in
+            (wid, wname, s)
+
+runBenchmarks :: [IO (Text, Double)] -> IO ()
+runBenchmarks bs = do
+    initializeTime
+    -- NOTE: Adding an artificial delay between successive runs to get a better
+    -- output for the heap profiling.
```
KtorZ, line 254: :+1:

KtorZ (approved):

Anviking:
> What about printing the UTxO statistics also here?

Can go in a later PR.

ℹ️ Testing once more on normal CI 6fb36ec (should be dropped) before merging

paweljakubas:
@Anviking  @KtorZ would be good also to add also latency benchmark in some future PR

Anviking:
> would be good also to add also latency benchmark in some future PR

So a byron version of the existing (jormungandr) latency benchmark in the nightly. Sounds like a good idea.

lib/byron/bench/Restore.hs
Anviking: For [`createWallet`](https://github.com/input-output-hk/cardano-wallet/blob/a3fecaccdb0600b0a5c76b1027441581cd2f43e6/lib/core/src/Cardano/Wallet.hs#L482), all the way to [`prefilterBlock`](https://github.com/input-output-hk/cardano-wallet/blob/a3fecaccdb0600b0a5c76b1027441581cd2f43e6/lib/core/src/Cardano/Wallet/Primitive/Model.hs#L370-L377).

PR #1484, opened 2020-03-25
disable Mux & Chain sync traces

KtorZ:
# Issue Number

#1398 

# Overview

- 6f9d4c15151669ca75654b031805d5735e4cc127
  :round_pushpin: **disable Mux & Chain sync traces**
  These traces, even if only printed at lower severities, are causing tremendous
slowness in the wallet and in particular in the restoration loop. They are
quite verbose, even for a 'Debug' output actually.

Also bumped the Handshake min severity from Debug to Info as this trace is
quite useful and only occuring once when a new client establishes a connection
with the node.

# Comments

### DIFF
--- a/lib/byron/src/Cardano/Wallet/Byron/Network.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron/Network.hs
@@ -98,7 +98,7 @@ import Control.Retry
     , retrying
     )
 import Control.Tracer
-    ( Tracer, contramap, traceWith )
+    ( Tracer, contramap, nullTracer, traceWith )
 import Data.ByteString.Lazy
     ( ByteString )
 import Data.Function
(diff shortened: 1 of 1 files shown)


Anviking (approved):

Anviking:
> Also bumped the Handshake min severity from Debug to Info as this trace is
quite useful and only occuring once when a new client establishes a connection
with the node.

I don't see this in the diff though. Looks like it is still `MsgHandshakeTracer{}       -> Debug`

KtorZ:
Ho! Indeed, I wrongly resolved the merge conflict when rebasing. Fixing.

Done --> https://github.com/input-output-hk/cardano-wallet/compare/6f9d4c15151669ca75654b031805d5735e4cc127..7c387cd5e87b79bf331ad7a2288e1b3f8461211e

PR #1487, opened 2020-03-26
nix: add cardano-node configuration to the windows build products

rvl:
This will let us download the correct config files when running cardano-launcher CI for windows.


nix/package-cardano-node.nix
```diff
@@ -28,6 +28,7 @@ pkgs.stdenv.mkDerivation rec {
   '' + (if pkgs.stdenv.hostPlatform.isWindows then ''
     cp --remove-destination -v ${pkgs.libffi}/bin/libffi-6.dll $out/bin
     cp --remove-destination -v ${cardano-node}/bin/* $out/bin
+    cp -Rv ${cardano-node.configs} $out/configuration
```
disassembler, line 31: While I don't particularly have a problem with this, these configs are not used by any of our products we document/release. Daedalus, the scripts in cardano-node, the NixOS services, etc... all default to grabbing the configuration from `iohkNix.cardanoLib.environments.${environment}`.

disassembler (approved):

nix/package-cardano-node.nix
rhyslbw: These are for [native windows tests](https://github.com/input-output-hk/cardano-launcher/runs/544376614?check_suite_focus=true#step:7:2). We could easily switch to a supported artifact produced by nix though

Issue #1483, opened 2020-03-25
Byron wallet CRUD - update-name, update-spending password


paweljakubas:
# Context

Those functionalities are needed for Byron Reboot

# Decision

1. Update-name will be as it was done for Jormungandr
2. get utxo statistics will be as it was done for Jormungandr
3. Update spending password will need to take into consideration below requirements:
(a) no password -> valid password   LEGAL
(b) valid password -> valid password   LEGAL
(c) valid password -> no password  ILLEGAL
Moreover, restoration should be reviewed to accept either no password or current password.

# Acceptance Criteria

1. Byron Api supports update-name, get utxo statistics and update-spending password
2. Byron Api restoring accepts also no password
3. Byron Api of update-spending password supports transitions specified in Decision section

---

# Development 

# QA 

It would be worth checking the changes in : 
1. swagger.yaml
2. check integration tests introduced which can be run via
```
stack test --ta '-m "BYRON_UPDATE_NAME"' cardano-wallet-byron:cardano-node-integration
stack test --ta '-m "BYRON_UTXO"' cardano-wallet-byron:cardano-node-integration
```

piotr-iohk:
lgtm.

PR #1495, opened 2020-03-27
Update password endpoint for byron

paweljakubas:
# Issue Number

#1483 

# Overview

- [x] I have added `ByronWalletPasspphraseData`
- [x] I have added endpoint to Api modules
- [x] I have added endpoint to swagger
- [x] I have updated Api.TypesSpec and Malformed
- [x] I have added integration tests for updating valid paswd->valid passwd
- [x] I have introduced Password "byron-raw"
- [x] I have enabled null passwords in `ByronWalletPostData` 
- [x] I have updated the rest of codebase bacause of it
- [x] I have added integration tests for updating no passwd -> valid passwd 

# Comments


paweljakubas:
bors try

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -469,6 +469,11 @@ data WalletPutPassphraseData = WalletPutPassphraseData
     , newPassphrase :: !(ApiT (Passphrase "raw"))
     } deriving (Eq, Generic, Show)
 
+data ByronWalletPutPassphraseData = ByronWalletPutPassphraseData
+    { oldPassphrase :: !(Maybe (ApiT (Passphrase "raw")))
```
KtorZ: This won't work unfortunately ... not only does the old passphrase can be empty, but it can also be less than 10 characters (which the new backend doesn't have as a limitation). 

My suggestion would be to have a `ApiT (Passphrase "byron-raw")`, for which we can define a different minimum length that is different from `Passphrase "raw"`, and coerce the types in the API handlers.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1045,22 +1047,39 @@ cmdWalletUpdatePassphrase mkClient =
...
+instance PassphraseUpdate ByronWalletPutPassphraseData where
+    producePassphraseData (OldNewPassword old new) =
+        if old == Passphrase @"raw" "" then
+            ByronWalletPutPassphraseData Nothing (ApiT new)
+        else
+            ByronWalletPutPassphraseData (Just $ ApiT old) (ApiT new)
+
+instance PassphraseUpdate WalletPutPassphraseData where
+    producePassphraseData (OldNewPassword old new) =
+        WalletPutPassphraseData (ApiT old) (ApiT new)
```
KtorZ: The extra parameter and instances above bring quite some overhead. I'd suggest to leave the command as it is. It won't be possible to update legacy wallets that have no or a invalid passphrase set via the command-line. That's an acceptable limitation :+1:

lib/core/src/Cardano/Wallet/Api/Types.hs

lib/cli/src/Cardano/CLI.hs
paweljakubas: reverted support for CLI

paweljakubas:
bors try

lib/core/src/Cardano/Wallet/Api/Client.hs
```diff
@@ -228,7 +229,7 @@ byronWalletClient =
             , listWallets = _listWallets
             , postWallet = _postWallet
             , putWallet = _putWallet
-            , putWalletPassphrase = error "TODO: putWalletPassphrase"
+            , putWalletPassphrase = error "putWalletPassphrase: not impl for byron"
```
KtorZ: This should be `_putByonWalletPassphrase` now, no :) ?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -899,7 +908,11 @@ postRandomWallet ctx body = do
         W.createWallet @(WorkerCtx ctx) @s @k wrk wid wName s
   where
     wName = getApiT (body ^. #name)
-    pwd   = preparePassphrase EncryptWithPBKDF2 $ getApiT (body ^. #passphrase)
+    pwd  = case body ^. #passphrase of
+        Nothing ->
+            preparePassphrase EncryptWithPBKDF2 $ Passphrase ""
+        Just (ApiT (Passphrase passwd)) ->
+            preparePassphrase EncryptWithPBKDF2 $ Passphrase passwd
```
KtorZ: suggestion:

```
pwd = preparePassphrase EncryptWithPBKDF2 $ fromMaybe mempty $ getApiT (body ^. #passphrase)
```

lib/core/src/Cardano/Wallet/Api/Client.hs
paweljakubas: I have this problem of using `ByronWalletPutPassphraseData` and not `WalletPutPassphraseData` in `WalletClient` and it contributed to the solution you did not liked  done in commit https://github.com/input-output-hk/cardano-wallet/pull/1495/commits/e485e7bad1fd04c6fab3b7997715a354b0d60e85

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -430,7 +430,7 @@ data SomeByronWalletPostData
 data ByronWalletPostData mw = ByronWalletPostData
     { mnemonicSentence :: !(ApiMnemonicT mw)
     , name :: !(ApiT WalletName)
-    , passphrase :: !(ApiT (Passphrase "raw"))
+    , passphrase :: !(Maybe (ApiT (Passphrase "byron-raw")))
```
KtorZ: Why this change? The "standard" API shouldn't allow creating Byron wallets with an empty passphrase, this should only be possible through the legacy endpoint when providing the encrypted XPrv. Sorry for the confusion, I misread your question on slack :(, I thought you were talking about the put passphrase endpoint.

paweljakubas: ok, then we cannot test it easily. I mean now we can and it passes but I can remove the last two commits and should be fine

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1136,6 +1161,27 @@ putWalletPassphrase ctx (ApiT wid) body = do
...
+putByronWalletPassphrase ctx (ApiT wid) body = do
+    let (ByronWalletPutPassphraseData oldM (ApiT new)) = body
+    withWorkerCtx ctx wid liftE liftE $ \wrk -> liftHandler $
+        case oldM of
+            Just (ApiT (Passphrase old')) -> do
+                let old = Passphrase @"raw" old'
+                W.updateWalletPassphrase wrk wid (old, new)
+            _ -> do
+                let old = Passphrase @"raw" ""
+                W.updateWalletPassphrase wrk wid (old, new)
```
KtorZ: suggestion:

```hs
let old = maybe mempty (coerce . getApiT) oldM  
W.updateWalletPassphrase wrk wid (old, new)
```

lib/core/src/Cardano/Wallet/Api/Client.hs
KtorZ: Okay, I'll look into this :+1:

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: ~~Hmmm. Okay, let's keep it.~~ .. That's a tricky problem. Because now, we need to change the passphrase type on every endpoint that requires a passphrase.

KtorZ (approved):

KtorZ:
Another occurrence of the "500" error sometimes happening on CI these days. Merging anyway.

PR #1497, opened 2020-03-27
 Integration test updating wallet name, that is restored from Xprv

piotr-iohk:
# Issue Number

#1436

# Overview

- 24d13f8ddf0da0186ae7790ca1851894f330dc24
  Fix swagger operationID for getUTxOsStatisticsByron so it there is no duplicate ids on rendered API doc
  
- df2e98c853b0ad98d33cd44e4e11a2d054d7d499
  Integration test updating wallet name, that is restored from Xprv

# Comments

### DIFF
--- a/lib/core-integration/cardano-wallet-core-integration.cabal
+++ b/lib/core-integration/cardano-wallet-core-integration.cabal
@@ -49,6 +49,7 @@ library
     , http-api-data
     , http-client
     , http-types
+    , memory
     , OddWord
     , memory
     , process
--- a/specifications/api/swagger.yaml
+++ b/specifications/api/swagger.yaml
@@ -2160,7 +2160,7 @@ paths:
 
   /byron-wallets/{walletId}/statistics/utxos:
     get:
-      operationId: getUTxOsStatistics
+      operationId: getUTxOsStatisticsByron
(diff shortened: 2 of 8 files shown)


specifications/api/swagger.yaml
```diff
@@ -2065,7 +2065,7 @@ paths:
 
   /byron-wallets/{walletId}/statistics/utxos:
     get:
-      operationId: getUTxOsStatistics
+      operationId: getUTxOsStatisticsByron
```
paweljakubas: then you need to update https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/Api.hs#L401 and type name below, probably -> check cardano-wallet-core:unit

lib/core-integration/cardano-wallet-core-integration.cabal
```diff
@@ -49,6 +49,7 @@ library
     , http-api-data
     , http-client
     , http-types
+    , memory
```
paweljakubas, line 52: then you need to have nix patch

paweljakubas (approved):
looks fine modulo comments I made

KtorZ (approved):

PR #1506, opened 2020-03-30
Update byron launcher and fix integration tests

paweljakubas:
# Issue Number

# Overview

- [x] I have updated laucher to work with 1.9* cardano-node

# Comments

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ cabal.project.local~
 ### Nix ###
 result*
 .stack-to-nix.cache
+
+### cardano-node ###
+lib/byron/test/data/cardano-node/genesis.json
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,3 @@ cabal.project.local~
 ### Nix ###
 result*
 .stack-to-nix.cache
-
-### cardano-node ###
-lib/byron/test/data/cardano-node/genesis.json
--- a/lib/byron/src/Cardano/Wallet/Byron/Launch.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron/Launch.hs
@@ -63,7 +63,6 @@ import System.IO.Temp
 import qualified Data.Aeson as Aeson
 import qualified Data.HashMap.Strict as HM
 import qualified Data.Text as T
-import qualified Data.Text.IO as T
 import qualified Data.Yaml as Yaml
 
 data CardanoNodeConfig = CardanoNodeConfig
@@ -158,10 +157,9 @@ withConfig tdir action =
         let nodeTopologyFile = dir </> "node.topology"
 
         -- we need to specify genesis file location every run in tmp
-        configTxt <- T.readFile (source </> "node.config")
-        let configTxtExtended =
-                T.append configTxt ("\nGenesisFile: "<> T.pack nodeGenesisFile)
-        T.writeFile (dir </> "node.config") configTxtExtended
+        Yaml.decodeFileThrow (source </> "node.config")
+            >>= withObject (addGenesisFilePath (T.pack nodeGenesisFile))
+            >>= Yaml.encodeFile (dir </> "node.config")
 
         Yaml.decodeFileThrow @_ @Aeson.Value (source </> "genesis.yaml")
             >>= withObject updateStartTime
@@ -209,6 +207,14 @@ updateStartTime
 updateStartTime m = do
     time <- round @_ @Int <$> getPOSIXTime
     pure $ HM.insert "startTime" (toJSON time) m
+
+-- | Add a "GenesisFile" field in a given object with the current path of
+-- genesis.json in tmp dir as value.
+addGenesisFilePath
+    :: Text
+    -> Aeson.Object
+    -> IO Aeson.Object
+addGenesisFilePath path = pure . HM.insert "GenesisFile" (toJSON path)
 
 -- | Do something with an a JSON object. Fails if the given JSON value isn't an
 -- object.
--- a/lib/byron/src/Cardano/Wallet/Byron/Launch.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron/Launch.hs
@@ -64,7 +64,6 @@ import System.IO.Temp
 
 import qualified Data.Aeson as Aeson
 import qualified Data.HashMap.Strict as HM
-import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
 import qualified Data.Yaml as Yaml
 
@@ -107,8 +106,6 @@ withCardanoNode tr tdir action =
         , "--topology", nodeTopologyFile cfg
         , "--database-path", nodeDatabaseDir cfg
         , "--socket-path", nodeSocketFile cfg
-        , "--genesis-file", nodeGenesisFile cfg
-        , "--genesis-hash", T.unpack (nodeGenesisHash cfg)
         , "--port", show port
         ]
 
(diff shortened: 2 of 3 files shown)


paweljakubas:
bors try

lib/byron/test/data/cardano-node/node.config
```diff
@@ -12,7 +12,7 @@ NodeId:
 NumCoreNodes: 1
 PBftSignatureThreshold: 1
 Protocol: RealPBFT
-GenesisFile: genesis.json
+GenesisFile: test/data/cardano-node/genesis.json
```
Anviking: Does this work from nix? (the previous `genesis.json` was probably overridden by the CLI arg, and probably didn't work)

lib/byron/src/Cardano/Wallet/Byron/Launch.hs
```diff
@@ -155,18 +150,14 @@ withConfig tdir action =
         let nodeConfigFile   = dir </> "node.config"
         let nodeDatabaseDir  = dir </> "node.db"
         let nodeDlgCertFile  = dir </> "node.cert"
-        let nodeGenesisFile  = dir </> "genesis.json"
-        let nodeGenesisFile1  = "test/data/cardano-node/genesis.json"
+        let nodeGenesisFile  = "test/data/cardano-node/genesis.json"
```
piotr-iohk: why not the same pattern as others?
```
let nodeGenesisFile  = dir </> "genesis.json"
```

lib/byron/test/data/cardano-node/node.config
paweljakubas: as far as I understand cardano-node team decided to point to genesis file and resign from `genesis-file` and `genesis-hash` option. So I take `genesis.yaml` transform it (with adding `startTime`) and it always lands in this directory. It does not work with just tmp directory for me. I do not know if it works with nix, though. Feel free to experiment and push on top

piotr-iohk (approved):
lgtm modulo comment.
(locally works for me)

lib/byron/src/Cardano/Wallet/Byron/Launch.hs
paweljakubas: I need to set it in node.config as it is new requirement (to my best knowledge) and tmp path is not deterministic. we have something like that /tmp/cardano-wallet-byron-2cbb1ea94edb1cea/genesis.json which is ok if we know `dir`. So as **fast** fix I decided to make it constant. Which can be probably improved later

paweljakubas:
failing on
```
src/Test/Integration/Framework/DSL.hs:1179:9:
--
  | 1) API Specifications BYRON_UPDATE_NAME_01 - Update names of wallets
  | expected: Status {statusCode = 204, statusMessage = "No Content"}
  | but got: Status {statusCode = 500, statusMessage = "Internal Server Error"}
  |  
  | from the following response: Left (HttpException ResponseTimeout)
```
plus nix voodoo

paweljakubas:
bors try

lib/byron/src/Cardano/Wallet/Byron/Launch.hs
paweljakubas: we will probably need here something like this : 
when we write to 
```
node.config -> tmp/cardano-wallet-byron-xxxxxxx/node.config
```
we add field upon copying: 
```
GenesisFile: tmp/cardano-wallet-byron-xxxxxxx/genesis.json 
```

.gitignore
```diff
@@ -28,3 +28,6 @@ cabal.project.local~
 ### Nix ###
 result*
 .stack-to-nix.cache
+
+### cardano-node ###
+lib/byron/test/data/cardano-node/genesis.json
```
piotr-iohk: Hm, that might be a bit confusing if we'll need to modify genesis.json.

lib/byron/src/Cardano/Wallet/Byron/Launch.hs
piotr-iohk: yes, also in such :point_up_2:  solution there is no need to put `lib/byron/test/data/cardano-node/genesis.json` into .gitignore.

.gitignore
paweljakubas: yes, working on it

lib/byron/test/data/cardano-node/node.config
paweljakubas: addressed in https://github.com/input-output-hk/cardano-wallet/pull/1506/commits/c1b1c8e42e57016eb3e42d3837974ca698fc6142

lib/byron/src/Cardano/Wallet/Byron/Launch.hs
paweljakubas: addressed in https://github.com/input-output-hk/cardano-wallet/pull/1506/commits/c1b1c8e42e57016eb3e42d3837974ca698fc6142

.gitignore
paweljakubas: reverted in https://github.com/input-output-hk/cardano-wallet/pull/1506/commits/c1b1c8e42e57016eb3e42d3837974ca698fc6142

paweljakubas:
bors try

lib/byron/test/data/cardano-node/node.config
```diff
@@ -12,7 +12,6 @@ NodeId:
 NumCoreNodes: 1
 PBftSignatureThreshold: 1
 Protocol: RealPBFT
-GenesisFile: test/data/cardano-node/genesis.json
```
paweljakubas: this one will be determined upon creation in `/tmp` knowing the location of `genesis.json` in `tmp`

lib/byron/src/Cardano/Wallet/Byron/Launch.hs
```diff
@@ -163,27 +157,30 @@ withConfig tdir action =
         let nodeSocketFile   = dir </> "node.socket"
         let nodeTopologyFile = dir </> "node.topology"
 
+        -- we need to specify genesis file location every run in tmp
+        configTxt <- T.readFile (source </> "node.config")
+        let configTxtExtended =
+                T.append configTxt ("\nGenesisFile: "<> T.pack nodeGenesisFile)
```
paweljakubas: wonder if it ("\n") is acceptable for windows env

KtorZ: Probably not. Good question. You have two options here IMO:

either, pattern match on `nativeNewLine` from https://hackage.haskell.org/package/base-4.12.0.0/docs/System-IO.html#t:Newline, and put the right mark depending on the result. 

Or, instead of appending a raw line, you could Yaml.decodeFileThrow the node.config, use `withObject` helper from this file, re-encode the modified node.config and write it directly to the temporary config folder (very much like it's node for the genesis file).

KtorZ (approved):
It'll be better to use the same pattern as what's done for the genesis file (i.e., instead of appending a raw line, decode it to a JSON value, modify the value, and write the modified JSON to the config dir). 

Also, this and the base branch need to be rebased on top of master. I'll rebase the base branch.

paweljakubas:
@KtorZ  used the same Yaml machinery like for genesis file. see https://github.com/input-output-hk/cardano-wallet/pull/1506/commits/f1776b1c1f8177b9e679a2658b423ded948987ea
Also I am rebased against master

KtorZ (approved):

paweljakubas:
bors try

PR #1508, opened 2020-03-30
skip jormungandr integration tests in CI, but enable cardano-node instead

KtorZ:
# Issue Number

N/A

# Overview

Next releases are going to be about cardano-node and we aren't touching jormungandr a lot more these days. So instead, Jörmungandr integration tests should eventually be run on a nightly base whereas cardano-wallet-byron will be on a much more frequent bases. We are getting a lot of this flaky "500 Internal Server Error" these days in CI on the Jörmungandr integration. Unfortunately, we don't have much to investigate further.. 

Plus, we really need cardano-node integration tests to be running in CI! So, I suggest we add a bullet point to the release note about running cardano-wallet-jormungandr:integration manually before any release.

### DIFF
--- a/.buildkite/rebuild.hs
+++ b/.buildkite/rebuild.hs
@@ -166,7 +166,7 @@ buildStep dryRun bk =
             [ color "always"
             , [ "test" ]
             , fast opt
-            , skip "cardano-node-integration"
+            , skip "jormungandr-integration"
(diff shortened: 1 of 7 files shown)


KtorZ:
bors try

piotr-iohk (approved):
+1

PR #1509, opened 2020-03-30
Fix fee calculation on Byron

KtorZ:
# Issue Number

#1462 

# Overview

- 7c1c4643148956ee20ff7c6dc4648133e2651adf
  :round_pushpin: **resurrect transaction unit test for fee estimation**
  
- 60c1f4e0d4c3b53c11cdaaac69d7b505f36071ba
  :round_pushpin: **fix fee estimation for byron**
  I can't count apparently. Aaaaaand, the encrypted derivation path can actually be
of variable size on random wallets

# Comments

I tried to make the property counter example more useful than it used to be for debugging.. 

```
Falsified (after 1 test and 3 shrinks):
 real size inf:   285
 real size sup:   289
 estimated size: 320

 Coin Selection:
 inputs: - 2nd 01010000 (~ 198202 @ 82d81858...0eb7bd45)
 outputs: - 197171 @ 82d81858...1b6b755e
 change: [1031]

 Raw Transaction:
 82  # list(2)
    83  # list(3)
       9f  # list(*)
          82  # list(2)
             00  # word(0)
             d8 18  # tag(24)
                58 24 82 58 20 01 01 00 00 01 01 00 00 01 00 00
                00 00 01 01 01 00 01 00 01 00 00 01 00 01 01 00
                01 01 00 01 00 01  # bytes(36)
       ff  # break
       9f  # list(*)
          82  # list(2)
             82  # list(2)
                d8 18  # tag(24)
                   58 21 83 58 1c d0 6d 92 32 a2 ab 1c 0f 8f ce ef
                   1a d0 2b 6f e4 02 e4 52 25 cc d7 9e 0e b6 79 fb
                   a7 a0 00  # bytes(33)
                1a 1b 6b 75 5e  # word(460027230)
             1a 00 03 02 33  # word(197171)
          82  # list(2)
             82  # list(2)
                d8 18  # tag(24)
                   58 21 83 58 1c 7c 81 05 9c ac 86 24 8b 41 40 32
                   ad 48 96 85 56 90 fa b7 10 55 2e 9a 09 b3 3b ab
                   1c a0 00  # bytes(33)
                1a d5 25 0e 6e  # word(3575975534)
             19 04 07  # word(1031)
       ff  # break
       a0  # map(0)
    81  # list(1)
       82  # list(2)
          00  # word(0)
          d8 18  # tag(24)
             58 85 82 58 40 00 00 00 00 00 00 00 00 00 00 00
             00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
             00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
             00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
             00 00 00 00 00 58 40 00 00 00 00 00 00 00 00 00
             00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
             00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
             00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
             00 00 00 00 00 00 00  # bytes(133)
```

### DIFF
--- a/lib/byron/src/Cardano/Wallet/Byron/Transaction/Size.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron/Transaction/Size.hs
@@ -56,14 +56,12 @@ sizeOfTx inps outs = 6
 
 -- SIGNED-TX
 --     = CBOR-LIST-LEN (2)    -- 1 byte
---     | U8                   -- 1 byte
---     | CBOR-LIST-LEN (2)    -- 1 byte
 --     | TX                   -- sizeOf(TX) bytes
 --     | CBOR-LIST-LEN (n)    -- 1-2 bytes (assuming n < 255)
 --     | *WITNESS             -- n * 139 bytes
 --                            == 3 + sizeOf(TX) + 1-2 + n * 139
 sizeOfSignedTx :: [TxIn] -> [TxOut] -> Int
-sizeOfSignedTx inps outs = 3
+sizeOfSignedTx inps outs = 1
     + sizeOfTx inps outs
     + sizeOf (CBOR.encodeListLen $ fromIntegral n)
     + n * sizeOfTxWitness
@@ -154,8 +152,8 @@ class MinSizeOf (t :: *) (n :: NetworkDiscriminant) (k :: Depth -> * -> *) where
 --             | 28OCTET              --    28 bytes
 --             | ATTRIBUTES (Ø)       --     1 byte
 --             | U8                   --     1 bytes
-instance MaxSizeOf Address 'Mainnet IcarusKey where maxSizeOf = 44
-instance MinSizeOf Address 'Mainnet IcarusKey where minSizeOf = 40
+instance MaxSizeOf Address 'Mainnet IcarusKey where maxSizeOf = 43
+instance MinSizeOf Address 'Mainnet IcarusKey where minSizeOf = 39
 
 -- ADDRESS (TestNet, Icarus)
 --     = CBOR-LIST-LEN (2)    --     1 byte
@@ -180,8 +178,8 @@ instance MinSizeOf Address 'Mainnet IcarusKey where minSizeOf = 40
 --             | 28OCTET              --    28 bytes
 --             | ATTRIBUTES (8)       --     8 bytes
 --             | U8                   --     1 bytes
-instance MaxSizeOf Address ('Testnet pm) IcarusKey where maxSizeOf = 51
-instance MinSizeOf Address ('Testnet pm) IcarusKey where minSizeOf = 47
+instance MaxSizeOf Address ('Testnet pm) IcarusKey where maxSizeOf = 50
+instance MinSizeOf Address ('Testnet pm) IcarusKey where minSizeOf = 46
 
 -- ADDRESS (MainNet, Random)
 --     = CBOR-LIST-LEN (2)    --     1 byte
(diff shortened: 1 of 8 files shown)


lib/byron/test/unit/Cardano/Wallet/Byron/TransactionSpec.hs
```diff
@@ -46,29 +54,85 @@ import Cardano.Wallet.Primitive.Types
...
+
+        it "Icarus / Mainnet" $ property $
+            propSizeEstimation @'Mainnet @IcarusKey mainnetMagic
+                (genSelection @'Mainnet @IcarusKey)
+                (\n -> vectorOf n $ genLegacyAddress @'Mainnet @IcarusKey)
+
+        it "Icarus / Testnet" $ property $
+            propSizeEstimation @('Testnet 459045235) @IcarusKey (testnetMagic @459045235)
+                (genSelection @('Testnet 459045235) @IcarusKey)
+                (\n -> vectorOf n $ genLegacyAddress @('Testnet 459045235) @IcarusKey)
```
KtorZ, line 134: Not as pretty as I'd like, many repetitions but... I have no time :(

stack.yaml
```diff
@@ -66,17 +66,17 @@ extra-deps:
...
 - network-3.1.0.1
 - connection-0.3.1
 - Unique-0.4.7.6
 - moo-1.2
 - gray-code-0.3.1
 
-# - bimap-0.4.0
-# - hedgehog-1.0
-# - tasty-hedgehog-1.0.0.1
+- git: https://github.com/well-typed/cborg
```
rvl, line 76: Is this a fix for `pprint: Unexpected token` or something else?

lib/byron/src/Cardano/Wallet/Byron/Transaction/Size.hs
```diff
@@ -194,20 +192,20 @@ instance MinSizeOf Address ('Testnet pm) IcarusKey where minSizeOf = 47
...
+--         62-66OCTET
 --             = CBOR-LIST-LEN (3)    --     1 byte
 --             | CBOR-BYTES (28)      --     2 bytes
 --             | 28OCTET              --    28 bytes
---             | ATTRIBUTES (34)      --    34 bytes
+--             | ATTRIBUTES (34)      -- 30-34 bytes
 --             | U8                   --     1 bytes
-instance MaxSizeOf Address 'Mainnet ByronKey where maxSizeOf = 77
-instance MinSizeOf Address 'Mainnet ByronKey where minSizeOf = 73
+instance MaxSizeOf Address 'Mainnet ByronKey where maxSizeOf = 76
```
rvl, line 207: These `maxSizeOf` values doesn't include the list len byte. I assume the code is correct and comment ambiguous perhaps.

lib/byron/test/unit/Cardano/Wallet/Byron/TransactionSpec.hs
```diff
@@ -159,6 +223,197 @@ spec = do
...
+       , MaxSizeOf Address n k
+       )
+    => ProtocolMagic
+    -> Gen CoinSelection
+    -> (Int -> Gen [Address])
+    -> Property
+propSizeEstimation pm genSel genChngAddrs =
+    withMaxSuccess 1000 $
+    forAllShrinkBlind genSel shrinkSelection $ \sel -> let nc = length (change sel) in
+    forAllBlind (genChngAddrs nc) $ \chngAddrs ->
```
rvl, line 244: OK so the forAll . forAll is because the generators are parameters of this property.

lib/byron/test/unit/Cardano/Wallet/Byron/TransactionSpec.hs
```diff
@@ -159,6 +223,197 @@ spec = do
...
+        , Arbitrary (LegacyAddress n k)
+        )
+    => Gen Address
+genLegacyAddress = do
+    LegacyAddress (_ :: Proxy n, _ :: Proxy k, addr) <- arbitrary
+    pure addr
+
+-- 'CoinSelection' are generated by running the @largestFirst@ algorithm on an
+-- arbitrary UTxO and to cover an arbitrary set of 'TxOut'.
+genSelection
```
rvl, line 323: Nice generators. I was wondering if it is actually necessary to generate a UTxO then run a coin selection, or are CoinSelection values enough.

rvl (approved):
Looks good to me.

stack.yaml
KtorZ: It is indeed.

lib/byron/test/unit/Cardano/Wallet/Byron/TransactionSpec.hs
KtorZ: Yeah, the issue here is that generators depends on which network discrimination and the type of address / key that is chosen. Thus in practice, we can't just have `Arbitrary Address`, because there are 4 types of `Address` in this test. We could define arbitrary instances on newtypes that wraps an `Address` with a `Proxy` or alike, but I found that passing the generator as parameters and using forAll was just simpler.

lib/byron/test/unit/Cardano/Wallet/Byron/TransactionSpec.hs
KtorZ: I tried generating CoinSelection at first, but there are just so many cases for invalid coin selections... Since it's quite easy however to generate valid UTxOs, I went down this path.

lib/byron/src/Cardano/Wallet/Byron/Transaction/Size.hs
KtorZ: Yeah, the comment is a bit ambiguous. I tried to improve things compare to what it was back on the http-bridge, but it's still far from perfect. I guess the next update is going to be Shelley with a non-ambiguous CDDL spec.

jonathanknowles:
It looks as though `cborg.nix` is missing from the `nix` directory:
```
getting status of '/var/lib/buildkite-agent/builds/packet-ipxe-3-ci3-2/input-output-hk/cardano-wallet/nix/.stack.nix/cborg.nix': No such file or directory
```
source: https://buildkite.com/input-output-hk/cardano-wallet/builds/6667#7f465c25-b83c-4227-a510-79a2fb244f00/22-52

PR #1511, opened 2020-03-31
Fix passphrase check when coming from empty legacy passphrase

KtorZ:
# Issue Number

#1483 

# Overview

- f29cb4c1d15d55fec378c73970b2b03e7127905f
  :round_pushpin: **add regression test for passphrase update generated from cardano-sl**
  So we _almost_ got it right... but we did forget that hashing a empty bytestring doesn't spit out
an empty bytestring, but an actual hash; It's not an identity. Therefore, we have to be careful
when 'preparing' the passphrase to NOT hash it if it's an empty one.

- 35574df4db0db4b095f177387c5fba7220cbe3a9
  :round_pushpin: **only hash legacy passphrases when they are not empty.**
  NOTE: this compare a 'Passphrase' with a 'Passphrase', which uses ScrubbedBytes under the hood for
constant time comparison.

Aaaand.. yes, we still need to serialize the empty passphrase to CBOR though

# Comments

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/ByronWallets.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/ByronWallets.hs
@@ -503,10 +503,28 @@ spec = do
             [ expectResponseCode @IO HTTP.status204
             ]
 
-    it "BYRON_UPDATE_PASS_04 - Updating passphrase with no password wallets" $ \ctx -> do
+    it "BYRON_UPDATE_PASS_04a - Updating passphrase with no password wallets" $ \ctx -> do
         w <- emptyRandomWalletWithPasswd ctx ""
         request @ApiByronWallet ctx (Link.getWallet @'Byron w) Default Empty
             >>= flip verify [ expectField #passphrase (`shouldSatisfy` isNothing) ]
+        let payload = updatePassPayload "" "correct-password"
+        r <- request @ApiByronWallet ctx
+            (Link.putWalletPassphrase @'Byron w) Default payload
+        verify r
+            [ expectResponseCode @IO HTTP.status204
+            ]
+
+    it "BYRON_UPDATE_PASS_04b - Regression test" $ \ctx -> do
+        let key = "38e8de9c583441213fe34eecc4e28265267466877ba4048e3ab1fa99563\
+                  \66947aefaf5ba9779db67eead7fc9cd1354b994a5d8d9cd40ab874bfeb1\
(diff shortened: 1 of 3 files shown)


jonathanknowles:
The test suite for `cardano-wallet-core` currently fails with:
```hs
  test/unit/Cardano/Wallet/Primitive/AddressDerivationSpec.hs:274:13:
  1) Cardano.Wallet.Primitive.AddressDerivation, golden test legacy passphrase encryption, compare new implementation with cardano-sl - empty password
       expected: Right ()
        but got: Left ErrWrongPassphrase

  To rerun use: --match "/Cardano.Wallet.Primitive.AddressDerivation/golden test legacy passphrase encryption/compare new implementation with cardano-sl - empty password/"

Randomized with seed 513786889

Finished in 534.0376 seconds
1741 examples, 1 failure, 1 pending
```
Link to failure: https://buildkite.com/input-output-hk/cardano-wallet/builds/6658#c13e52a3-ef85-4c3c-8a3e-0ef77d83db51/1730-5112

lib/core-integration/src/Test/Integration/Scenario/API/ByronWallets.hs
```diff
@@ -514,6 +514,24 @@ spec = do
...
+                  \66947aefaf5ba9779db67eead7fc9cd1354b994a5d8d9cd40ab874bfeb1\
+                  \b33649280cd33651377731e0e59e0233425a55257782c5adaa768da0567\
+                  \f43c1c6c0c18766ed0a547bb34eb472c120b170a8640279832ddf180028\
+                  \87f03c15dea59705422d"
+        let pwd = "31347c387c317c574342652b796362417576356c2b4258676a344a314c6\
+                  \343675375414c2f5653393661364e576a2b7550766655513d3d7c6f7846\
+                  \36654939734151444e6f38395147747366324e653937426338372b484b6\
+                  \b4137756772752f5970673d"
+        w <- emptyByronWalletFromXPrvWith ctx "random" ("Random Wallet", key, pwd)
+        let payload = updatePassPayload "" "correct-password"
```
Anviking, line 528: Shouldn't the `""` be `pwd` here? Or `pwd` on the previous line `""`?

paweljakubas: @Anviking no, because `pwd` here is hash of password which is "" that is produced by Scrypt function, and `updatePassPayload` operates on `curl` level.

paweljakubas (approved):
looks fine

jonathanknowles (approved):

PR #1512, opened 2020-03-31
report empty legacy passphrase as no passphrase

KtorZ:
# Issue Number

#1483 

# Overview

- 678f595e8e3b6d14044a0d21e623b4f55b73a7e2
  :round_pushpin: **report empty legacy passphrase as no passphrase**
    Legacy wallets imported through via XPrv might have an empty passphrase set. The passphrase is empty from a client perspective, but in practice it still exists (it is a CBOR-serialized empty bytestring!).

  Therefore, if we detect an empty passphrase, we chose to return the metadata as if no passphrase was set, so that clients can act in consequences.

# Comments


jonathanknowles (approved):
Looks good to me!

PR #1513, opened 2020-03-31
nix: Fix build of cardano-wallet-jormungandr-tests-win64

rvl:
Fix a failing build: https://hydra.iohk.io/build/2178754#tabs-summary

Use of makeWrapper in the newly added restore benchmark needs to be disabled on Windows.

### DIFF
--- a/nix/haskell.nix
+++ b/nix/haskell.nix
@@ -87,14 +87,15 @@ let
 
         # cardano-node will want to write logs to a subdirectory of the working directory.
         # We don't `cd $src` because of that.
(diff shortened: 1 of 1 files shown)


Anviking (approved):
So `makeWrapper` doesn't work on windows?

rvl:
Yeah `makeWrapper` can't work on windows. What it does is create a bash script wrapper and manipulates the PATH variable, etc, in a posix-specific way. There is no equivalent, so it causes the build to fail.

bors r+

PR #1515, opened 2020-03-31
Update cardano-node matrix

paweljakubas:
# Issue Number

# Overview

- [x] I have updated cardano-node matrix to reflect current situation on master

# Comments


KtorZ (approved):
Good point.

PR #1516, opened 2020-03-31
Use more convenient Map functions

jhrcek:
# Issue Number

N/A

# Overview

Another batch of minor simplifications I saw were possible when reading through the sources. Mostly reusing existing Map functions where appropriate.

# Comments


lib/core/src/Cardano/Wallet/Primitive/CoinSelection/LargestFirst.hs
```diff
@@ -60,7 +60,7 @@ largestFirst opt outs utxo = do
         Nothing -> do
             let moneyRequested = sum $ (getCoin . coin) <$> (descending outs)
             let utxoBalance = fromIntegral $ balance utxo
-            let nUtxo = fromIntegral $ L.length $ (Map.toList . getUTxO) utxo
+            let nUtxo = fromIntegral $ Map.size $ getUTxO utxo
```
jhrcek, line 63: Asymptotic improvement from `O(n)` to `O(1)`

lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -197,7 +195,7 @@ newDBLayer trace fp = do
                     poolMarginN
                     poolMarginD
                     poolCost_
-            insertMany_ $ uncurry (PoolOwner poolId) <$> zip poolOwners [0..]
+            insertMany_ $ zipWith (PoolOwner poolId) poolOwners [0..]
```
jhrcek, line 198: Avoid constructing and then immediately deconstructing tuples

lib/core/src/Cardano/Pool/DB/Model.hs
```diff
@@ -219,7 +219,7 @@ mReadSystemSeed db@PoolDatabase{seed} =
 
 mReadCursor :: Int -> ModelPoolOp [BlockHeader]
 mReadCursor k db@PoolDatabase{pools} =
-    let allHeaders = foldMap snd $ Map.toList pools
+    let allHeaders = fold pools
```
jhrcek, line 222: Using the fact that Map is Foldable and the list (values in the pools map) is Monoid we can avoid unnecessary traversal.

lib/core/src/Cardano/Pool/DB.hs
```diff
@@ -47,6 +47,7 @@ import System.Random
 -- with @atomically@:
 --
 -- Example:
+--
```
jhrcek, line 50: This code example was unreadable, because it was made part of paragraph. Separating the code block from surrounding text fixes formatting.

lib/core/test/unit/Cardano/Wallet/DB/StateMachine.hs
```diff
@@ -828,7 +828,7 @@ tag = Foldl.fold $ catMaybes <$> sequenceA
         update matches ev =
             case match ev of
                 Just wid ->
-                    Map.alter (Just . (+1) . fromMaybe 0) wid matches
+                    Map.insertWith (+) wid 1 matches
```
KtorZ, line 831: :+1:

lib/core/src/Cardano/Pool/DB/Model.hs
KtorZ: I'd expect the compiler to optimize list fusions for me but maybe that's too naive / too strong faith in GHC :grimacing:

KtorZ (approved):
Thank you :)

PR #1520, opened 2020-03-31
Fix passphrase conversion when coming from legacy wallets

KtorZ:
# Issue Number

#1483 

# Overview

- 0a72efc8a84a700d3d524c3c9883b35bd3bf8eb5
  :round_pushpin: **only CBOR-serialize the passphrase when checking its hash**
  The actual encryption passphrase that is used is the hash provided to the API
by clients, it can be either null or a 32-byte string. Then, this passphrase
is serialized to CBOR before being hashed, but, that is the non-serialized
passphrase that is used for encrypted the root key!

- 6c88eead09e29614a86d235b79b9e2b177695b7f
  :round_pushpin: **use golden string generated from cardano-sl implementation!**
  The previous strings were generated from our implementation, which disqualifies them from being valid
test cases. When comparing with a legacy implementation, the data MUST come from the legacy implementation
otherwise the test is pointless.

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
@@ -527,21 +527,13 @@ encryptPassphrase  (Passphrase bytes) = do
         <> BA.convert @ByteString (fastPBKDF2_SHA512 params bytes salt)
 
 -- | Manipulation done on legacy passphrases before getting encrypted.
---
--- Yes there's CBOR. Yes it's hashed before being encrypted. You don't actually
--- pronounce the 'l' in salmon. A lot of things in the world don't make sense
--- but we still live just fine.
 preparePassphrase
     :: PassphraseScheme
     -> Passphrase "raw"
     -> Passphrase "encryption"
 preparePassphrase = \case
     EncryptWithPBKDF2 -> coerce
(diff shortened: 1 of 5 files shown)


paweljakubas (approved):

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -527,21 +527,13 @@ encryptPassphrase  (Passphrase bytes) = do
         <> BA.convert @ByteString (fastPBKDF2_SHA512 params bytes salt)
 
 -- | Manipulation done on legacy passphrases before getting encrypted.
---
--- Yes there's CBOR. Yes it's hashed before being encrypted. You don't actually
--- pronounce the 'l' in salmon. A lot of things in the world don't make sense
```
paweljakubas, line 532: I will miss this comment ;-)

KtorZ: Me too. It'll live in the git history and will be our little secret 🤫

paweljakubas:
so we need to revert this commit https://github.com/input-output-hk/cardano-wallet/pull/1511/commits/702cc229faaa33ba1117be690632d9db7400195b right?

PR #1507, opened 2020-03-30
Store bench node socket in tmp dir

Anviking:
# Issue Number

#1458 

# Overview

- [x] Make sure the socket path is in a temporary directory

# Comments

PR #1521, opened 2020-03-31
Fix broken bench with up-to-date config from nix

Anviking:
# Issue Number

#1458 

# Overview

- [x] I use a nix helper to retrieve a valid mainnet config using nix

# Comments

The old approach of pointing to files in the `cardano-node/configuration` directory had two problems:
- There is no testnet specific topology or config
- The configuration/configuration-mainnet.yaml contains relative paths

### DIFF
--- a/lib/byron/bench/Restore.hs
+++ b/lib/byron/bench/Restore.hs
@@ -170,17 +170,14 @@ main = do
     -- `--benchmark-arguments "testnet"`
     let network = Mainnet
 
-    -- Reading setup from env vars.
-    --
-    -- If you usually run ./scripts/mainnet.sh in the cardano-node repository,
-    -- it can be useful to set
-    --
-    --     export BYRON_CONFIGS=$CARDANO_NODE_DIR/configuration
-    --     export NODE_DB=$CARDANO_NODE_DIR/db
-    --
-    -- before running this benchmark manually with stack.
-    configs <- getEnv "BYRON_CONFIGS"
+    ----------------------------------------------------------------------------
+    -- Environment variables set by nix/haskell.nix (or manually)
+    topology <- getEnv "NODE_TOPOLOGY"
+    config <- getEnv "NODE_CONFIG"
+    ----------------------------------------------------------------------------
+    -- Environment variables set by ./buildkite/bench-restore.sh (or manually)
     nodeDB <- getEnv "NODE_DB"
+    ----------------------------------------------------------------------------
 
     -- Temporary directory for storing socket
     tmpDir <- getCanonicalTemporaryDirectory
@@ -190,16 +187,16 @@ main = do
     let args =
             [ "run"
             , "--database-path", nodeDB
-            , "--topology", configs </> "mainnet-topology.json"
+            , "--topology", topology
             , "--socket-path", socketPath
-            , "--config", configs </> "configuration-mainnet.yaml"
+            , "--config", config
             , "--port", "7776"
             ]
-    -- NOTE: It is warned against using @NoStream@ to supress output on POSIX
-    -- systems, since it could lead to the process throwing errors.
-    -- It appears to work fine stopping the live view of cardano-node,
-    -- however.
-    let cmd = Command "cardano-node" args (return ()) NoStream NoStream
+    let cmd = Command "cardano-node" args (return ()) Inherit Inherit
+
+    sayErr "Starting node with command:"
+    sayErr $ pretty cmd
+
     void $ withBackendProcess nullTracer cmd $ do
         case network of
             Testnet _ -> error "Testnet: not supported yet"
--- a/nix/haskell.nix
+++ b/nix/haskell.nix
@@ -18,8 +18,8 @@
 let
   haskell = pkgs.haskell-nix;
   jmPkgs = pkgs.jmPkgs;
-  commonLib' = import ./default.nix {};
-  commonLib = commonLib'.commonLib;
+  # commonLib = (import ./default.nix {}).commonLib;  # option a - shorter
+  inherit (import ./default.nix {}) commonLib; # option b - even shorter
 
   # our packages
   stack-pkgs = import ./.stack.nix/default.nix;
--- a/nix/haskell.nix
+++ b/nix/haskell.nix
@@ -18,6 +18,8 @@
 let
   haskell = pkgs.haskell-nix;
   jmPkgs = pkgs.jmPkgs;
+  commonLib' = import ./default.nix {};
+  commonLib = commonLib'.commonLib;
 
   # our packages
   stack-pkgs = import ./.stack.nix/default.nix;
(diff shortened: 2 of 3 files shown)


lib/byron/bench/Restore.hs
```diff
@@ -190,16 +191,16 @@ main = do
...
-            , "--config", configs </> "configuration-mainnet.yaml"
+            , "--config", config
             , "--port", "7776"
             ]
-    -- NOTE: It is warned against using @NoStream@ to supress output on POSIX
-    -- systems, since it could lead to the process throwing errors.
-    -- It appears to work fine stopping the live view of cardano-node,
-    -- however.
-    let cmd = Command "cardano-node" args (return ()) NoStream NoStream
+    let cmd = Command "cardano-node" args (return ()) Inherit Inherit
```
Anviking, line 195: The node config is not using `LiveView` anymore but `SimpleView`, which indeed breaks when we use `NoStream`. Using `silently` seemed complicated since we need to change `withBackendProcess` or `Command`. I found the current output to stdout to be fine though, and shouldn't contaminate any data.

(detailed progress & timestamps are written to a separate file)

nix/haskell.nix
```diff
@@ -87,13 +89,16 @@ let
...
+				#
+				# TODO: Using the configuration dir works for mainnet, but to add testnet
+        # support, we need to retrieve it properly.
         packages.cardano-wallet-byron.components.benchmarks.restore =
           pkgs.lib.optionalAttrs (!pkgs.stdenv.hostPlatform.isWindows) {
             build-tools = [ pkgs.makeWrapper ];
             postInstall = ''
               wrapProgram $out/bin/restore \
-                --set BYRON_CONFIGS ${pkgs.cardano-node.configs} \
-                --prefix PATH : ${pkgs.cardano-node}/bin
```
Anviking, line 103: So — why not use `scripts.mainnet.node` for a pre-configured script?

That script would set `--socket-path` and `database-path` relative to cwd.

We need:
1. The `--database-path` to be in the cache
2. The `--socket-path` to be in a temporary directory to not break CI (#1507)

We cannot simply change the cwd to satisfy both constraints.

There might be some leeway and alternatives, but building the launch script ourselves seemed like the safest option.

nix/pkgs.nix
```diff
@@ -3,10 +3,22 @@
...
+  commonLib' = import ./default.nix {};
+  commonLib = commonLib'.commonLib;
 in pkgs: super: with pkgs; {
   jmPkgs = import ./jormungandr.nix { inherit (pkgs) commonLib; inherit pkgs; };
+  cardanoNodePkgs = import sources.cardano-node { inherit system crossSystem config; };
   cardano-node = cardanoNodePkgs.cardano-node // {
+    mainnet = {
+      configFile =
+        let
+          rawCfg = commonLib.cardanoLib.environments.mainnet.nodeConfig // {
```
Anviking, line 14: Would be much neater and less risky for typos to not replicate `mainnet` three times, but didn't try figuring out how to do that… yet.

Saw that it was [possible](https://github.com/input-output-hk/cardano-node/blob/f9d541c51720811897310d09df8b8e43deae9570/nix/scripts.nix#L90).

nix/haskell.nix
Anviking: Another option might be to get access to the internal [`mkNodeScript`](https://github.com/input-output-hk/cardano-node/blob/f9d541c51720811897310d09df8b8e43deae9570/nix/scripts.nix#L16-L35) of cardano-node… 🤔

Anviking:
Testing with https://buildkite.com/input-output-hk/cardano-wallet/builds/6728#88d45b71-6eb1-44a0-b709-29859a009e77 in #1496 it doesn't seem like the node is even starting.

Works locally for me using both `./.buildkite/bench-restore.sh` and ` nix-build -A benchmarks.cardano-wallet-byron.restore -o restore`

**Edit** forgot to keep the `cardano-node` in the PATH in `wrapProgram`. Fixed.

lib/byron/bench/Restore.hs
Anviking: https://buildkite.com/input-output-hk/cardano-wallet/builds/6731#9c9e2709-e4a2-4574-8d06-b0aef9221c68 Logs exceed 1MB… should probably limit them.

Anviking: https://github.com/input-output-hk/cardano-wallet/pull/1521/commits/c48cbee1cd7cf488e4df0ee684655f9c107588d8

nix/haskell.nix
```diff
@@ -18,6 +18,8 @@
 let
   haskell = pkgs.haskell-nix;
   jmPkgs = pkgs.jmPkgs;
+  commonLib' = import ./default.nix {};
+  commonLib = commonLib'.commonLib;
```
rvl:
```suggestion
  # commonLib = (import ./default.nix {}).commonLib;  # option a - shorter
  inherit (import ./default.nix {}) commonLib; # option b - even shorter
```

nix/pkgs.nix
rvl: Sometimes a little repetition is ok if the factorized version is complicated.

This seems ok anyway. I had to do a similar thing in [cardano-launcher/test/utils.ts](https://github.com/input-output-hk/cardano-launcher/blob/1bfb55114d7d5c4fe0f9bd3454d06cdc9f6f0a43/test/utils.ts#L125-L173) (except that I edited the config file with javascript rather than nix).

nix/haskell.nix
rvl: Yes fair enough. The config system in cardano-node could be tweaked to better suit our purposes.

lib/byron/bench/Restore.hs
rvl: Maybe not in this PR, but it would be good if cardano-node logs were still written to file (perhaps even uploaded as buildkite artifacts). Then we can refer back to them if a benchmark runs badly.

lib/byron/bench/Restore.hs
```diff
@@ -190,16 +187,16 @@ main = do
...
```
rvl, line 195: There is actually a `Cardano.Wallet.Byron.Launch.withCardanoNode` function.
Would this also be suitable for the restore benchmark?

nix/pkgs.nix
```diff
@@ -3,10 +3,23 @@
 let
   sources = import ./sources.nix {};
   pkgs1903 = import sources."nixpkgs-19.03" {};
-  cardanoNodePkgs = import sources.cardano-node { inherit system crossSystem config; };
+  commonLib' = import ./default.nix {};
+  commonLib = commonLib'.commonLib;
```
rvl:
```suggestion
  inherit (import ./default.nix {}) commonLib;
```

rvl (approved):
Looking good

Anviking:
Hope this nightly will work now bors r+

lib/byron/bench/Restore.hs
Anviking: That module currently only contains self-node/integration-style setup… Maybe the module should be re-named if we don't end up adding any "normal" launch code

nix/pkgs.nix
Anviking: > Sometimes a little repetition is

I know. But here I suspect the factorised version should be clearly better. I was very close to returning the testnet config for this mainnet helper by accident.

PR #1526, opened 2020-04-01
nix: Add fully static build of cardano-wallet-byron

rvl:
### Overview

I saw the musl static build of `cardano-wallet-byron` was failing and it's pretty simple to fix.

# Comments

Download links:
- https://hydra.iohk.io/job/Cardano/cardano-wallet-pr-1526/cardano-wallet-byron-linux64/latest-finished
- https://hydra.iohk.io/job/Cardano/cardano-wallet-pr-1526/cardano-wallet-byron-macos64/latest-finished

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -149,21 +149,19 @@ let
       migration-tests = jobs.x86_64-w64-mingw32.migration-tests.x86_64-linux;
     };
 
-    # Fully-static linux binary (placeholder - does not build)
-    cardano-wallet-jormungandr-linux64 = let
-      name = "cardano-wallet-jormungandr-${project.version}";
-      tarname = "${name}-linux64.tar.gz";
-    in pkgs.runCommand "${name}-linux64" {
-      buildInputs = with pkgs.buildPackages; [ gnutar gzip binutils ];
-    } ''
-      cp -R ${jobs.musl64.cardano-wallet-jormungandr.x86_64-linux}/bin ${name}
-      chmod -R 755 ${name}
-      strip ${name}/*
-
-      mkdir -p $out/nix-support
-      tar -czf $out/${tarname} ${name}
-      echo "file binary-dist $out/${tarname}" > $out/nix-support/hydra-build-products
-    '';
+    # Fully-static linux binaries
+    cardano-wallet-jormungandr-linux64 = import ./nix/linux-release.nix {
+      inherit pkgs;
+      exes = [ jobs.musl64.cardano-wallet-jormungandr.x86_64-linux ];
+    };
+    cardano-wallet-byron-linux64 = import ./nix/linux-release.nix {
+      inherit pkgs;
+      exes = [
+        jobs.musl64.cardano-wallet-byron.x86_64-linux
+        # SRE-83 dependencies fail to build
+        # jobs.musl64.cardano-node.x86_64-linux
+      ];
+    };
 
     # macOS binary and dependencies in tarball
     cardano-wallet-jormungandr-macos64 = let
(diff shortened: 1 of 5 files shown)


KtorZ:
I just noticed we don't have any full static build for macos64 either :s, but only have it for windows:

https://hydra.iohk.io/job/Cardano/cardano-wallet/cardano-wallet-byron-win64/latest

Anviking:
_just rebasing for more recent artifacts_

rvl:
I don't think it's possible to make fully static builds for macOS.
What's missing is a hydra download link for `cardano-wallet-byron-macos64.tar.gz`. This is set up in `release.nix`. There is already one for `cardano-wallet-jormungandr-macos64`.

rvl:
@KtorZ I added a .tar.gz download for macos64 as well.

release.nix
```diff
@@ -149,36 +149,32 @@ let
...
-    '';
+    cardano-wallet-jormungandr-macos64 = import ./nix/macos-release.nix {
+      inherit pkgs;
+      exes = [ jobs.native.cardano-wallet-jormungandr.x86_64-darwin ];
+    };
+    cardano-wallet-byron-macos64 = import ./nix/macos-release.nix {
+      inherit pkgs;
+      exes = [
+        jobs.native.cardano-wallet-byron.x86_64-darwin
+        jobs.native.cardano-node.x86_64-darwin
```
KtorZ, line 175: Why bundle `cardano-node` with the wallet here? I don't think our release assets should contain anything else than the wallet and the DLLs / runtime dependencies needed to run it.

Anviking:
Ok, I note we are now using the name e.g. `cardano-wallet-byron-2020.4.1-win64.zip` instead of `cardano-wallet-jormungandr-win64-v2020-04-01.zip`

release.nix
rvl: To save users the not-insignificant hassle of finding the correct version of cardano-node and downloading then installing the macos build for it. To save ourselves from support requests where users are running the wrong version of cardano-node. Because cardano-node cannot yet report its own version. And I don't think there are even macos binaries of cardano-node to available to download.
We can remove it sure, but it costs so little to include in the tarball.

rvl:
Yes `2020.4.1` is the Cabal package version (these cannot contain version numbers with leading zero digits) and `v2020-04-01` is the git tag. It's possible to do a little regex to convert the Cabal package version, if you want.

release.nix
KtorZ: > And I don't think there are even macos binaries of cardano-node to available to download.

:man_facepalming: ... you have a point. There are actually not even Linux assets at the moment :upside_down_face: ... 
And since it is not possible to (easily) know the version of the `cardano-node` executable at the moment .... 

Fair enough!

KtorZ:
> (these cannot contain version numbers with leading zero digits)

Note that this is only a constraints on old versions of cabal. Starting from ... < I can't remember which version, I think 1.18 >, cabal accepts leading zeros just fine. I have nevertheless always been quite afraid of the impact that  bumping the min bound on  `cabal-version:       >=1.10` would have on the nix builds. I currently have cabal 3.0.2.0 behind the scene, and I am not even sure that we can actually build the software with 1.10 anyway :| ... So maybe we simply bump the min bound?

KtorZ (approved):

rvl:
I did previously attempt to increase the Cabal minimum version to see if it would allow version numbers with leading zeros. That didn't seem to work for me, but perhaps someone else will have more luck.

The nix builds will work fine with new Cabal file versions. I added test cases to Haskell.nix for this. :-)

rvl:
Eval issues on Hydra have been fixed.

bors r+

PR #1528, opened 2020-04-01
update cardano-node matrix

paweljakubas:
# Issue Number

# Overview

- [x] I have updated cardano-matrix to point also 1.9.3 as valid version

# Comments


piotr-iohk (approved):

Anviking (approved):
Why not:
- Either mention `1.9.1, 1.9.2, 1.9.3` or just the latest `1.9.3`? (Just wondering)
- Separate the versions by comma instead of just a space (think it would look neater)

paweljakubas:
yeah, we can just put 1.9.3 I believe

PR #1532, opened 2020-04-01
Re-write and automate release note template

Anviking:
# Issue Number

None.

# Overview

- [x] Let's rewrite the release-notes so the are more slim to read
- [x] Let's add some form of templating (e.g.`{{GIT_TAG}}`) such that we never forget to change all dates appropriately.
- [ ] Let's add & integrate with @KtorZ 's [change-log script](https://gist.github.com/KtorZ/5098d42611658c65a5df835b0e73336f)
- [ ] Would be great if CI could set the correct template-variables and generate change-log all by itself

# Comments

- I **know** there is an existing template in `.github/RELEASE_TEMPLATE.md`. We shouldn't duplicate it.


Anviking:
Might re-open later.

PR #1538, opened 2020-04-03
Rename projects, executables and update travis deploy step

Anviking:
# Issue Number

None.

# Overview

- [ ] WIP: Make sure binaries are uploaded to travis
- [x] Instead of `cardano-wallet-byron` and `cardano-wallet-jormungandr`, let's have `cardano-wallet` and `cardano-wallet-itn`.

# Comments


Anviking:
bors try

KtorZ:
I am afraid of the number of things this would break in the nix pipeline, hydra and al ... We should just go with the Travis change (and the binaries renamed there first) and possibly, rename then the cabal packages.

Anviking:
Superseded by #1544

rvl:
@Anviking @KtorZ One thing that could also be done is move the Cabal packages up to the top-level, instead of having them inside `lib`.

Issue #1535, opened 2020-04-02
Icarus wallets are created with Mainnet addresses even if the server is in 'Testnet' mode


piotr-iohk:
# Context
... what the heck...

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | `2020.4.1 (git revision: df83d3f54b96ebe9f516a7a216c0d1cc9f169aa0)` cardano-node 1.9.3 |
| Platform     | linux, all?          |
| Installation | from release          |

# Steps to Reproduce

0. start cardano-node + wallet on testnet
 - `nix-build -A scripts.testnet.node -o testnet.node.local && testnet.node.local`
 -  `cardano-wallet-byron serve --node-socket ~/wb/cardano-node/state-node-testnet/node.socket --testnet /nix/store/7106qjwral616j6p5049m7hhi5ih0r33-testnet-genesis.json --database ./testnet`

1. Restore a wallet.
```
curl  -vX POST http://localhost:8090/v2/byron-wallets \
  -H "Content-Type: application/json; charset=utf-8" \
  -d '{
"style": "icarus",
"name": "Icarus",
"mnemonic_sentence": ["merry", "output", "wrestle", "second", "jealous", "electric", "network", "course", "gravity", "valid", "walnut", "trumpet", "sniff", "behind", "switch"],
"passphrase": "Secure Passphrase"
}' --http1.1  |  jq
```
2. Verify addresses generated for that wallet
```
curl  -vX GET http://localhost:8090/v2/byron-wallets/480fcd62295228616a6b0bdf5b818f40beb688c3/addresses |  jq

```
they are like:
```
Ae2tdPwUPEZMhXT3psnE9aaDJVhFFzE6NHjKyeU1xCN2YFmruAJtpK3BCTk
Ae2tdPwUPEZD7cMRTsUcQsZoa1LnSkYtRGLrBc7yQGw4GJWF22qjhh7s4xG
Ae2tdPwUPEZF5LBV9LPE5VMPLh5WMnQnqqT2noewK7Mp6g9bUPNNf1TifyG
Ae2tdPwUPEZLP3McqvFkiB842jRoVnYCxyofGgdkMSuP4gWguZMoS8tiYWP
Ae2tdPwUPEZJUAWD7JvnqB6iVUEr4hD8sfwswS4G7kuGVktSQ7G7ttuAfYN
[...]

PR #1540, opened 2020-04-03
Fix wrong key fingerprint for Icarus addresses on Testnet

KtorZ:
# Issue Number

#1535

# Overview

- 942f536017be58222b1008dbede17dd3c27e83ad
  :round_pushpin: **repair fingerprint conversion for Icarus addresses**
  This was forcing 'Mainnet by default because of how we consider Icarus addresses
on the ITN. However, now that we have integrated with cardano-node, this exception
is no longer valid.

Note that we can't really take anything else than the full address for a
fingerprint on Byron addresses because both the public key and the
network discriminant gets hashed in the address root! Thus there's no
way to recover them from a raw address. This also forces us to provide a
network discriminant in order to compute the fingerprint!

- f73322046afbf77b282a404cc9751d11a2ec0ca3
  :round_pushpin: **add simple golden tests on address decoding**
  We had 0 so far and it's always good to check that we can at least decode existing addresses..
These addresses were taken from the mainnet and testnet explorers

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -530,13 +530,13 @@ createIcarusWallet
     -> (k 'RootK XPrv, Passphrase "encryption")
     -> ExceptT ErrWalletAlreadyExists IO WalletId
 createIcarusWallet ctx wid wname credentials = db & \DBLayer{..} -> do
-    let s = mkSeqStateFromRootXPrv credentials (mkUnboundedAddressPoolGap 10000)
+    let s = mkSeqStateFromRootXPrv @n credentials (mkUnboundedAddressPoolGap 10000)
     let (hist, cp) = initWallet block0 bp s
     let addrs = map address . concatMap (view #outputs . fst) $ hist
     let g  = defaultAddressPoolGap
     let s' = SeqState
-            (shrinkPool (liftPaymentAddress @n) addrs g (internalPool s))
-            (shrinkPool (liftPaymentAddress @n) addrs g (externalPool s))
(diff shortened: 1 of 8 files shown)


piotr-iohk (approved):
lgtm.

Generated Icarus addresses on Testnet are:
```
2cWKMJemoBahmgvHu12VNW64Z65qMJfhyqaUbNNpjntXMzMpRCwPJkZvia4TzhzYFsZhn
2cWKMJemoBajoEJeJNtZGc3SURzZTs7qZdPy3Dp8dBeaUMctA2xMdm6swwQA7KMjcKYBU
2cWKMJemoBahhRKCZNqeAVaJxXA5zFa7zPxSB6rX4E2qRPp8iHuP2NkZc2Q7bB6y9hrEo
2cWKMJemoBahXDtSH819MLxDk8w8ddJN7GGLU75bXNV56aRuFkS4yTSfetQxdpqXwVAKj
...
```

Generated Icarus addresses on Mainnet are:
```
Ae2tdPwUPEYw5pkUEQSN2GKF6NcERHUnv6uwpnowAhssV38MKvkHrcnUzhV
Ae2tdPwUPEZ6GdqauDg5veWykKbALtpebNxnduwohTjAci4hjGML1FU4Rk1
Ae2tdPwUPEZAWBZwDgSF4yitDvPJXpko5BFaxFmiXDDrGbXgcFi2Y8tGSPv
Ae2tdPwUPEZL7nfhE2Fe1PLQirxRFvVVtnsA9eoWAV1NYBodz4auRTYq1e3
...
```

PR #1541, opened 2020-04-03
Make sure to only generate addresses using the Hardened index convention

KtorZ:
# Issue Number

# Overview

- be2cd3d3b2b921db2d674cc031ff3c9c2ace536c
  :round_pushpin: **Make sure to only generate addresses using the Hardened index convention**
  It seems that the old / legacy wallet has trouble digesting addresses that are using indexes in the
soft convention. Note that the derivation is always hardened, regardless of the index value but
the convention was switched a while ago on -sl to use only indexes beyond 2^32...

Note that there are still many addresses on the current Mainnet that do not abide by this convention
and these addresses seems to break legacy wallet. So, if we can avoid generating potentially
problematic addresses in the new version, that's good :) ...

We still recognize address and account indexes across the whole domain though

- 91ecb305093549121ebca03c30eb6b0292789dd2
  :round_pushpin: **hack, hopefully temporarily, rnd state reading from the database**
  See comment in DB/Sqlite.

- 5e7a6041610918d9df41d0d3a15940422de5b010
  :round_pushpin: **re-generate API golden test for PostRandomAddressData**
  This now contains indexes that are considered invalid :)...

- b93c623486ddc26edd3415d74c6e82b2be632162
  :round_pushpin: **remove soft derivation index addresses from known addresses**
    smoke tested with wallet made from the previous release:

  ```
  $ cardano-wallet address list aef25f56c818fc31494183702016b82227cd4702
  Ok.
  [
    {
        "state": "unused",
        "id": "2w1sdSJu3GVfRsedS1xiTkpHAg8DuBZZdEXVY971uAdebdmyD75c271fMdj2j33dndqKchWis5Leaxzkxc6ksi2uupu68Ykk7V1"
    },
    {
        "state": "unused",
        "id": "2w1sdSJu3GVix5rkaLKd4fNMbUsY9ufWz1A6zWq6nyG3QFcris1rW9yaipuHDfrL8NgBWStDmAwm7TdSpqbcsMYo333tudcQYcV"
    }
  ]

  -- Restarting the server using the new version, migration applies automatically.

  $ ./cardano-wallet address list aef25f56c818fc31494183702016b82227cd4702
  Ok.
  []
  ```

# Comments

```
$ cardano-wallet-byron address create 924e219963cfcb5e36e56447bf63848a21134c75
Please enter your passphrase: **********
Ok.
{
    "state": "unused",
    "id": "DdzFFzCqrhsv6mq5SVNwaUkkFXBAcVpss3tmWXhxMLRgQ17iMRWEMik3a1bDYuymnQ8cgP7N4uCa5BFFBXK2raAPzqsEVZdPNKxNU5H9"
}
```

### DIFF
--- a/lib/byron/test/integration/Test/Integration/Byron/Scenario/API/Addresses.hs
+++ b/lib/byron/test/integration/Test/Integration/Byron/Scenario/API/Addresses.hs
@@ -225,7 +225,7 @@ scenario_ADDRESS_CREATE_05 = it title $ \ctx -> do
     w <- emptyRandomWallet ctx
     let payload = Json [json|
             { "passphrase": "Secure Passphrase"
-            , "address_index": 14
+            , "address_index": 2147483662
             }|]
     r <- request @(ApiAddress n) ctx (Link.postRandomAddress w) Default payload
     verify r
@@ -245,7 +245,7 @@ scenario_ADDRESS_CREATE_06 = it title $ \ctx -> do
     w <- emptyRandomWallet ctx
     let payload = Json [json|
             { "passphrase": "Secure Passphrase"
-            , "address_index": 14
+            , "address_index": 2147483662
             }|]
     r0 <- request @(ApiAddress n) ctx (Link.postRandomAddress w) Default payload
     verify r0 [ expectResponseCode @IO HTTP.status201 ]
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -1302,7 +1302,7 @@ cmdAddressList mkClient =
 -- | Arguments for 'address create' command
 data AddressCreateArgs = AddressCreateArgs
     { _port :: Port "Wallet"
-    , _addressIndex :: Maybe (Index 'WholeDomain 'AddressK)
+    , _addressIndex :: Maybe (Index 'Hardened 'AddressK)
     , _id :: WalletId
     }
 
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -187,6 +187,7 @@ import Cardano.Wallet.Primitive.AddressDerivation
     , checkPassphrase
     , deriveRewardAccount
     , encryptPassphrase
+    , liftIndex
     , preparePassphrase
     , toChimericAccount
     )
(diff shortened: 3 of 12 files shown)


Anviking:
I still get "Address Verification Failed" when sending from Binance to an address generated with this PR… (and address still starting with `2w1sdS` also, not `DdzFFz`)

Anviking:
With the new addresses I get a derivation path of e.g. 
`(Index {getIndex = 0},Index {getIndex = 2206972976})`.

The legacy addresses seem to have a derivation path of e.g.
`(Index {getIndex = 2147483648},Index {getIndex = 2147483648})`

Is the `0` correct? 
_was I really on the right version when I generated that address?_ 🤔

Anviking:
I think the wallet was reading the existing rndState from the DB with account index of `0` for me:  https://github.com/input-output-hk/cardano-wallet/blob/95a964bbca2ddffd948515d0cfabc9bfd701b0bd/lib/core/src/Cardano/Wallet.hs#L894-L896

Hence no effect on my existing wallet.

Anviking:
After deleting the wallet and re-restoring it I am able to generate `DdzFFz…` addresses with the account index `2147483648` that Binance accepts.

Anviking (approved):
We might want to overwrite the previous "invalid" account index in the DB.

KtorZ:
> I still get "Address Verification Failed" when sending from Binance to an address generated with this PR… (and address still starting with 2w1sdS also, not DdzFFz)

Yeah I forgot to mention that but, since we do store the account used for derivation in the database currently, without any changes, that's the one still used for derivation. I'll add something to overwrite this at runtime, because we've already pushed out a version of Daedalus flight which does this :(

KtorZ:
Okay, fixed at the database level.. I am actually even surprised it didn't raised any issue in the first place but we are being quite conservative when deserializing indexes :grimacing: ...

Without the fix:

```
  1) Cardano.Wallet.DB.Sqlite, Sqlite (RndState), insertState . selectState (regression account index)
       expected: Just (Index {getIndex = 2147483648})
        but got: Just (Index {getIndex = 0})
```

After the fix:

```
  Sqlite (RndState)
     insertState . selectState (regression account index)
Finished in 0.0059 seconds
1 example, 0 failures

```

PR #1543, opened 2020-04-06
Better error message when trying to generate address on icarus wallet.

piotr-iohk:
# Issue Number

# Overview

- e3d60368faa266c9445c12edf329afe9e9ef05af
  Better error message when trying to generate address on icarus wallet.
# Comments

Was:
```
$ cardano-wallet-byron address create 3fe1acf86fefbf66b7481470ea81244eadde079d
Please enter your passphrase: *****************
It looks like something unexpected went wrong. Unfortunately I don't yet know how to handle this type of situation. Here's some information about what happened:
```

is:
```
$ cardano-wallet-byron address create 3fe1acf86fefbf66b7481470ea81244eadde079d
Please enter your passphrase: *****************
I cannot derive new unused address for this wallet type. Make sure to use Byron random wallet id.

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -1764,6 +1764,7 @@ data ErrCreateRandomAddress
     = ErrIndexAlreadyExists (Index 'Hardened 'AddressK)
     | ErrCreateAddrNoSuchWallet ErrNoSuchWallet
     | ErrCreateAddrWithRootKey ErrWithRootKey
+    | ErrCreateAddressNotAByronWallet
     deriving (Generic, Eq, Show)
 
 {-------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by KtorZ — APPROVED** [review]
bors r+

**Comment by KtorZ** [comment]
My bad. Bors will fail to merge :point_down: .

**Comment by piotr-iohk** [comment]
> My bad. Bors will fail to merge point_down .

Why?

**Review by paweljakubas — APPROVED** [review]
LGTM

File: lib/core/src/Cardano/Wallet.hs [hunk 1]
```diff
@@ -1764,6 +1764,7 @@ data ErrCreateRandomAddress
     = ErrIndexAlreadyExists (Index 'Hardened 'AddressK)
     | ErrCreateAddrNoSuchWallet ErrNoSuchWallet
     | ErrCreateAddrWithRootKey ErrWithRootKey
+    | ErrCreateAddressInvalidWalletType
```
Anviking: The constructor name`ErrCreateAddressInvalidWalletType` makes it seem like the error can be thrown for any wallet, but the error message contains: `Make sure to use Byron random wallet id.` — i.e. specific to Byron wallets.

A bit nit-picky, but maybe mention Byron in the name, like `ErrCreateAddressNotAByronWallet`, or take the expected wallet type name as a string.

Anviking (approved):

lib/core/src/Cardano/Wallet.hs
piotr-iohk: `ErrCreateAddressNotAByronWallet` it is.

PR #1557, opened 2020-04-09
Make the clean shutdown handler simpler

rvl:
This modifies the clean shutdown handler to use a FD from the command line, rather than taking stdin.

It also makes the logic even simpler, as shown in input-output-hk/cardano-node#767

The command-line arguments are changed from `--shutdown-handler` to `--shutdown-ipc FD`
so that they are identical to cardano-node.


KtorZ (approved):
Awesome :heart:

rvl:
I added another unit test and made the tests also build on Windows. But the unit tests seem to have a race condition somewhere, and don't always pass.

KtorZ:
@rvl Do you feel like wrapping this one up anytime soon :grimacing: ?

rvl:
I had a little trouble getting the tests to pass, so put this on the backburner.

I rebased and had another try at fixing it.

rvl:
Will fix it later.

Issue #1505, opened 2020-03-30
Cannot force resync wallet and cardano-node crash as a result


piotr-iohk:
# Context
 
Observed on testnet and staging-shelley and mainnet...

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | cardano-wallet: `2020.3.16 (git revision: fff00d48db58bc73f3b927955944fbeb0f23d23f)`, cardano-node 1.9.1 |
| Platform     | All        |
| Installation | from source           |

# Steps to Reproduce

0. start cardano-node + wallet on testnet
 - `nix-build -A scripts.testnet.node -o testnet.node.local && testnet.node.local`
 -  `cardano-wallet-byron serve --node-socket ~/wb/cardano-node/state-node-testnet/node.socket --testnet /nix/store/7106qjwral616j6p5049m7hhi5ih0r33-testnet-genesis.json --database ./testnet`

1. Restore a wallet.
```
curl  -vX POST http://localhost:8090/v2/byron-wallets \
  -H "Content-Type: application/json; charset=utf-8" \
  -d '{
"style": "random",
"name": "Byron",
"mnemonic_sentence": ["recycle","august","horse","awesome","shuffle","agent","almost","cheap","fury","speak","duty","enemy"],
"passphrase": "Secure Passphrase"
}' --http1.1  |  jq
```
2. Force resync wallet:
```
curl  -vX PUT http://localhost:8090/v2/byron-wallets/d3358ac0fc699784af021710e7aba2a3e31fbbd1/tip \
  -H "Content-Type: application/json; charset=utf-8" \
  -d '{
"slot_number": 0,
"epoch_number": 0
}' --http1.1  |  jq
```

## Expected behavior

Force resync is successful

## Actual behavior
 
Wallet is not resyncing. (state is 0% syncing)

Rolling messages cardano-wallet log:
```
[cardano-wallet.wallet-engine:Error:17] [2020-03-30 10:34:08.44 UTC] c29a2bfb: Unexpected error following the chain: user error (restoreBlocks: given chain isn't a valid continuation. Wallet is at: 8f860283-[0.0#0] but the given chain continues starting from: 8f860283-[0.0#0])
[cardano-wallet.network:Debug:4492] [2020-03-30 10:34:08.44 UTC] Couldn't connect to node (x1). Retrying in a bit...
[...]

PR #1559, opened 2020-04-09
Correct `toPoint` to handle genesis correctly 

Anviking:
# Issue Number

#1505 

# Overview

- [x] Now we map `BlockHeader`s to `Origin` **only** if the header hash matches the genesis hash.
- [x] Added roundtrip test that covers this bug

# Comments

- Previously we would map the "first real block", which also has `SlotId 0 0`, to `Origin`. The node would then believe the wallet was at genesis, instead of having the first real block applied.


lib/byron/test/unit/Cardano/Wallet/Byron/CompatibilitySpec.hs
```diff
@@ -40,3 +53,36 @@ spec = do
...
+    arbitrary = Hash . BS.pack <$> vector 32
+
+instance Arbitrary (Hash "BlockHeader") where
+    arbitrary = Hash . BS.pack <$> vector 32
+
+instance (Arbitrary (Tip ByronBlock)) where
+    arbitrary = oneof
+        [ return TipGenesis
+        , arbitraryTip
+        ]
```
KtorZ, line 73: Would be slightly better to use `frequency` here and give a lesser weight to `TipGenesis`, otherwise, half of the tests are actually testing the same thing.

lib/byron/src/Cardano/Wallet/Byron/Compatibility.hs
```diff
@@ -248,10 +248,22 @@ toEpochSlots :: W.EpochLength -> EpochSlots
...
+    -> W.EpochLength
+    -> W.BlockHeader
+    -> Point ByronBlock
+toPoint genesisH epLength (W.BlockHeader sid _ h _)
+  | h == (conv genesisH) = O.GenesisPoint
+  | otherwise =
+    O.Point $ Point.block (toSlotNo epLength sid) (toByronHash h)
+  where
+    conv :: W.Hash "Genesis" -> W.Hash "BlockHeader"
+    conv (W.Hash x) = W.Hash x
```
KtorZ: Or simply `coerce` from `Data.Coerce`

lib/byron/src/Cardano/Wallet/Byron/Compatibility.hs
```diff
@@ -248,10 +248,22 @@ toEpochSlots :: W.EpochLength -> EpochSlots
...
+
+toPoint
+    :: W.Hash "Genesis"
+    -> W.EpochLength
+    -> W.BlockHeader
+    -> Point ByronBlock
+toPoint genesisH epLength (W.BlockHeader sid _ h _)
+  | h == (conv genesisH) = O.GenesisPoint
+  | otherwise =
+    O.Point $ Point.block (toSlotNo epLength sid) (toByronHash h)
```
KtorZ: The indentation is weird here. Indentation should enforce a hierarchy in the information. Here `otherwise` is at the same level as `O.Point`, so the eye treat them similarly: as another branch / guard of the function definition. Indenting it properly makes it easier to process:

```hs
  | h == (conv genesisH) = O.GenesisPoint
  | otherwise =
    O.Point $ Point.block (toSlotNo epLength sid) (toByronHash h)
```

vs

```hs
  | h == (conv genesisH) = O.GenesisPoint
  | otherwise =
      O.Point $ Point.block (toSlotNo epLength sid) (toByronHash h)
```

lib/byron/src/Cardano/Wallet/Byron/Compatibility.hs
```diff
@@ -248,10 +248,22 @@ toEpochSlots :: W.EpochLength -> EpochSlots
 toEpochSlots =
     EpochSlots . fromIntegral . W.unEpochLength
 
-toPoint :: W.EpochLength -> W.BlockHeader -> Point ByronBlock
-toPoint epLength (W.BlockHeader sid _ h _)
-    | sid == W.SlotId 0 0 = genesisPoint
-    | otherwise = O.Point $ Point.block (toSlotNo epLength sid) (toByronHash h)
+-- | Magic value for the absence of a block.
+hashOfNoParent :: W.Hash "BlockHeader"
+hashOfNoParent = W.Hash . BS.pack $ replicate 0 32
```
KtorZ, line 253: :+1:

KtorZ (approved):

lib/byron/src/Cardano/Wallet/Byron/Compatibility.hs
Anviking: Turns out the newline wasn't needed anyway.

PR #1560, opened 2020-04-09
use cardano-haskell snapshots instead of extra-deps

KtorZ:
# Issue Number

N/A

# Overview

- [x] I have used cardano-haskell snapshots instead of extra-deps

# Comments


rvl (approved):
Much simpler.

Issue #1567, opened 2020-04-14
Weird tx history and balance when force-resyncing


Anviking:
# Context

The force-resync endpoint allows artificially rolling back the wallet to (SlotId 0 0).

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | includes the #1559 fix  |
| Platform     | mac          |
| Installation |  Official FC with cardano-wallet manually replaced  |

# Steps to Reproduce

1. Launch Daedalus FC with a wallet
2. 
```
curl  -vX PUT https://localhost:50078/v2/byron-wallets/021028cf0b109aab8f4fde7dd124a89efeeb4ed8/tip \
       -H "Content-Type: application/json; charset=utf-8" \
       -d '{ "slot_number": 0, "epoch_number": 0 }' --http1.1 --cert $TLS/client.pem --cacert $TLS/ca.crt  |  jq
```

## Expected behavior

I'd expect:
1. <s>All tx history disappears</s>

But otherwise:
2. <s>All of the tx history becomes pending but otherwise don't change</s>

And always:
3. Tx history should be in sync with the wallet balance
4. The wallet balance should (here) be 0 immediately after force-resyncing

## Actual behavior

- Incoming txs disappear from history. Outgoing txs are turned to pending.
- The pending outgoing txs have the wrong dates
- The wallet balance does not match the tx history
- The wallet balance is not 0 at the start of the chain. The balance decreases to become more accurate while syncing, despite the wallet essentially only having _incoming_ transactions.
 

---

# Resolution 

---

# QA

KtorZ:
- **Incoming txs disappear from history. Outgoing txs are turned to pending.**

This is the actual expected behavior. When rolling back, incoming txs are forgotten (and will eventually be re-discovered). On the contrary, outgoing txs may still be in a mempool and be accepted somewhere. Possibly, they'll end up invalid because they reference a no-longer existing UTxO. But it is so far up to users to cancel them. 
Later, we will add automatic re-submission of transaction, using a transaction scheduler. 

- **The pending outgoing txs have the wrong dates.**

This is also expected. The transaction is becoming pending at the precise time the rollback happened. 

- **The wallet balance does not match the tx history**
- **The wallet balance is not 0 at the start of the chain. The balance decreases to become more accurate while syncing, despite the wallet essentially only having incoming transactions.**

[...]

PR #1573, opened 2020-04-15
[WIP] Add mock golden rollback test with DB

Anviking:
# Issue Number

Relates to #1567, #1575 

# Overview

- [x] Use a mock chain to easily test corner cases
- [x] Demonstrate that `totalBalance` may appear to "duplicate" money due to inclusion of change from invalid pending tx
- [ ] TODO: Factor out and create more tests

# Comments

- Abstracting further to some ultra-concise DSL might be useful
- Seems alarming to me that we don't have any properties that prevented the `totalBalance` problem. Want to understand our divergence from the spec better. But this is an orthogonal issue to this PR.

### DIFF
--- a/lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
+++ b/lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
@@ -553,6 +553,89 @@ fileModeSpec =  do
                 -- getAvailableBalance `shouldReturn` 5
                 -- getTotalBalance `shouldReturn` 7
 
+            it "(NEEDS FIX, #1567) Pending txs can duplicate balance 2" $ \f -> do
+                -- Similar test case as above. But this time the pending txs will
+                -- have "unexisting" txins that cannot be spent.
+                --
+                -- The spec calls for removing pending transactions when
+                -- their inputs are spent in a new block.
+                --
+                -- Doing so would not fix the - arguably surprising - result of
+                -- this test.
+                (_ctx, db@DBLayer{..}) <- newDBLayer' (Just f)
+
+                let ourAddrs = knownAddresses (getState testCp)
+
+                atomically $ do
+                    unsafeRunExceptT $ initializeWallet testPk testCp testMetadata mempty
+
+                let mockApply h mockTxs = do
+                        Just cpA <- atomically $ readCheckpoint testPk
+                        let slotA = slotId $ currentTip cpA
+                        let hashA = headerHash $ currentTip cpA
+                        let fakeBlock = Block
+                                (BlockHeader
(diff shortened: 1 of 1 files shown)


lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
```diff
@@ -447,10 +465,204 @@ fileModeSpec =  do
...
+                getAvailableBalance db `shouldReturn` 2
+                getTotalBalance db `shouldReturn` 2
+                getTxsInLedger db `shouldReturn` [(Outgoing, 3), (Incoming, 5)]
+                getPending db `shouldReturn` []
+
+
+
+                atomically $ void $ unsafeRunExceptT $ rollbackTo testPk (SlotId 0 0)
+                getTxsInLedger db `shouldReturn` []
+                -- getPending `shouldReturn` []
```
KtorZ, line 530: I disagree here. getPending should return the outgoing transaction made at slot 2

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
```diff
@@ -447,10 +465,204 @@ fileModeSpec =  do
...
+                -- but got: [(Outgoing, 3)]
+
+                getAvailableBalance db `shouldReturn` 0
+                -- getTotalBalance `shouldReturn` 0
+                -- but got: 2
```
KtorZ, line 535: `2` is also expected here.

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
```diff
@@ -447,10 +465,204 @@ fileModeSpec =  do
...
+                                (blockHash "tx2b")
+                                [ (TxIn (blockHash "tx1") 0, Coin 5)
+                                ]
+                                [ TxOut (ourAddrs !! 3) (Coin 3)
+                                , TxOut (ourAddrs !! 2) (Coin 2)
+                                ]
+                            ]
+                -- THESE ARE NOT EXPECTED, BUT PASS!!
+                -- getAvailableBalance `shouldReturn` 5
+                -- getTotalBalance `shouldReturn` 7
```
KtorZ, line 554: I don't quite see why they "are not expected" though. This is 100% expected, but I agree, maybe not the best behavior we would like here. But, the code is simply doing doing what's it told to do.

Anviking:
Closing, but I think we need to re-visit the wallet's behaviour regarding pending txs, balance and rollback.

_[thread with some notes / thoughts](https://input-output-rnd.slack.com/archives/GBT05825V/p1587052833056000?thread_ts=1586967724.044500&cid=GBT05825V)_

PR #1576, opened 2020-04-16
Buildkite nightly: Remove cache clean jobs

rvl:
The cache cleaning is now done with a cron job on the buildkite agents. These pipeline steps never worked well anyway because it was not possible to run them on all buildkite agents at the same time.


Anviking (approved):
> These pipeline steps never worked well anyway because it was not possible to run them on all buildkite agents at the same time.

👌

Issue #1575, opened 2020-04-15
DB.rollbackTo may cause inconsistencies between checkpoints and other data


Anviking:
# Context
 
 

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      |  |
| Platform     |           |
| Installation |             |

# Steps to Reproduce

If our DB stores:
```
Checkpoints: cpA                cpB
TxMeta:          m1
Certs:                     c1
--------------------------------------> Increasing slotId
                              |
                              |
                              + p, rollback to here
```
And we rollback to some point p between c1 and cpB.

## Expected behavior

- We rollback to `cpA`, and make sure all other data is consistent with that checkpoint.

## Actual behavior

We rollback to `cpA`
We don’t remove txMeta1 and cert1 because they are earlier than the _requested_ rollback point _p_.

And our new state is:
```
Checkpoints: cpA            
TxMeta:          m1
Certs:                     c1
--------------------------------------> Increasing slotId
```
The wallet will continue restoring from cpA. `m1` is a txMeta from a different chain, and shouldn't exist.

- [Example test 1](https://github.com/input-output-hk/cardano-wallet/blob/5333afc469294bfa74e07de3317578545661ea05/lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs#L458-L503)
- [Example test 2](https://github.com/input-output-hk/cardano-wallet/pull/1573/files#diff-14a8238703a0aa58c4ac6e0fcda7e6eaR556-R663)
 

---

# Resolution 

- Clear certs, txMeta etc based on the actual rollback point instead of the requested rollback point #1579 

---

# QA 

- Golden test in DB close which is almost a "mockchain" that illustrates problem
- Adjusted both Sqlite DB, and Model DB
- Also had to tweak an existing property test

piotr-iohk:
lgtm.

PR #1579, opened 2020-04-16
Fix rollback of TxMeta and Certs

Anviking:
# Issue Number

#1575 

# Overview

- [x] I have constructed a regression test using the DB and a mock chain.
- [x] I tweaked the `rollbackTo` code to clear from `nearestPoint` instead of `point`. I.e. the actual rollback point instead of the requested rollback point.
- [x] Improved state machine counterexamples slightly

# Comments

- We should perhaps think about why no higher-level property caught this bug previously.


lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
```diff
@@ -10,7 +10,7 @@
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeApplications #-}
 
-{-# OPTIONS_GHC -fno-warn-orphans #-}
+{-# OPTIONS_GHC -fno-warn-orphans -fno-warn-unused-imports #-}
```
Anviking: Oops

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -505,21 +505,21 @@ newDBLayer trace defaultFieldValues mDatabaseFile = do
                     Just _  -> lift $ throwIO (ErrNoOlderCheckpoint wid point)
                 Just nearestPoint -> do
                     deleteCheckpoints wid
-                        [ CheckpointSlot >. point
+                        [ CheckpointSlot >. nearestPoint
```
Anviking, line 494: 🤔 Should work, but the change should also not be necessary.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -505,21 +505,21 @@ newDBLayer trace defaultFieldValues mDatabaseFile = do
...
                         ]
                     deleteDelegationCertificates wid
                         [ CertSlot >. point
```
Anviking: Should also be `nearestPoint`

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
Anviking: Keeping it for consistency.

Anviking:
State machine tests will fail?

bors try

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
```diff
@@ -471,6 +484,88 @@ fileModeSpec =  do
...
+                -- Since we increment the blockHeight faster than naturally,
+                -- The wallet should have pruned everything but the first and last
+                -- checkpoint.
+                --
+                -- So when we rollback to the previous step...
+                atomically . void . unsafeRunExceptT $
+                    rollbackTo testPk (SlotId 2 0)
+                Just cp <- atomically $ readCheckpoint testPk
+
+                -- We end up at genesis.
```
KtorZ: That comment is superfluous. Comments should give detail about "why" , not "what". What is written just below and already perfectly readable :ok_hand:

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
```diff
@@ -471,6 +484,88 @@ fileModeSpec =  do
...
+                slotId (currentTip cp) `shouldBe` (SlotId 0 0)
+
+                -- (Outgoing, 2) should /not/ be a "InLedger" transaction!
+                --
+                -- Whether is is now pending, or removed completely is not the
+                -- concern of this test.
```
KtorZ: Same as above, the comment is superfluous.

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
```diff
@@ -471,6 +484,88 @@ fileModeSpec =  do
...
+                        Just cpA <- atomically $ readCheckpoint testPk
+                        let slotA = slotId $ currentTip cpA
+                        let Quantity bhA = blockHeight $ currentTip cpA
+                        let hashA = headerHash $ currentTip cpA
+                        let fakeBlock = Block
+                                (BlockHeader
+                                { slotId = slotA { epochNumber = (epochNumber slotA) + 1}
+                                -- Increment blockHeight by steps greater than k
+                                -- Such that old checkpoints are always pruned.
+                                , blockHeight = Quantity $ bhA + 5000
```
KtorZ, line 495: I am not sure what the pruning has to do with this test and why it matters to introduce this extra complexity in the test?

KtorZ (commented):
Looks good, although I wonder why not writing a simple property here using arbitrary data rather using mock and fake and dummy everywhere :thinking:  ?

Also, seems like `prop_rollbackTxHistory`, for a good reason it seems. The test isn't actually taking care of any consistency consideration between checkpoints and history. 

Also, the database model needs to be updated accordingly, hence the state-machine failures.

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
Anviking: > That comment is superfluous

Yes, perhaps. I totally agree in general.

My reasoning might have been: There is a lot of boilerplate here and "What"-comments highlight the important stuff.

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
Anviking: Hm… indeed

Anviking: No, this is essential. Without pruning there is no problem.

If a checkpoint exists for every slot, then `nearestPoint` and `requestedPoint` are going to be equal, and the difference does not matter (Also see diagram in #1575)

lib/core/test/unit/Cardano/Wallet/DB/Properties.hs
```diff
@@ -726,10 +726,13 @@ prop_rollbackTxHistory db@DBLayer{..} (InitialCheckpoint cp0) (GenTxHistory txs0
...
             unsafeRunExceptT $ putTxHistory wid txs0
 
-    prop wid point = do
-        _ <- run $ unsafeRunExceptT $ mapExceptT atomically $ rollbackTo wid point
+    prop wid requestedPoint = do
+        point <- run $ unsafeRunExceptT $ mapExceptT atomically $ rollbackTo wid requestedPoint
         txs <- run $ atomically $ readTxHistory wid Descending wholeRange Nothing
-        monitor $ counterexample $ "\nTx history after rollback: \n" <> fmt txs
+
+        monitor $ counterexample $ "\n" <> "Actual Rollback Point:\n" <> (pretty point)
```
Anviking, line 734: This is weird. 
```
         Requested Rollback point:
         79.0

         Actual Rollback Point:
         0.0

         Original tx history:
         -  9eb44fb2 14.6 in ledger outgoing 2

         New tx history:
         -  9eb44fb2 14.6 in ledger outgoing 2
         Outgoing txs are reschuled ✗

  To rerun use: --match "/Cardano.Wallet.DB.MVar/MVar/rollback/Correctly re-construct tx history on rollbacks/"
```

Through experimentation it seems the `TxMetaSlot >. nearestPoint` predicate is ineffective. (If removed all outgoing become pending, all incoming are removed). But still: wat?

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
KtorZ: Functions should be used then, they are clearer than comments, and re-usable.

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
KtorZ: But you don't have to insert a checkpoint at every slot for the test in the first place?

lib/core/test/unit/Cardano/Wallet/DB/Properties.hs
Anviking: Not sure what fixed it 🤷‍♂️

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
Anviking: I wanted to include [`prune`](https://github.com/input-output-hk/cardano-wallet/blob/7e0aa90186462721397b22a3ca90967862716fbb/lib/core/src/Cardano/Wallet.hs#L767) similar to `restoreBlocks`, to convince myself that the bug could happen in reality.

But, yes, we could also specify which checkpoints to write manually.

Anviking:
Could move
- Make (Wallet s) a plain ADT to derive Generic 
- Add more shrinking to State Machine Tests
- Derive ToExpr instances using generic

to separate PR if desired.

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
Anviking: 755065b

lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs
```diff
@@ -413,6 +414,14 @@ instance Arbitrary UTxO where
             <*> vector n
         return $ UTxO $ Map.fromList u
 
+
+instance Arbitrary a => Arbitrary (Range a) where
+    arbitrary = Range <$> arbitrary <*> arbitrary
+
+    shrink (Range from to) =
+        [Range from' to | from' <- shrink from]
+        ++ [Range from to' | to' <- shrink to]
```
KtorZ, line 424: Are we sure that `to'` is necessarily bigger than `from` here ? For `from'`, it should shrink to a value smaller than `from` so there's no obvious problem, but for `to'` :shrug:

KtorZ (approved):
LGTM. I wonder how this previously property got fixed though...

Issue #1577, opened 2020-04-16
LocalStateQuery: Use params from chain when building txs


rvl:
# Context

Part of [ADP-83](https://jira.iohk.io/projects/ADP/issues/ADP-83).

The wallet server must keep track of certain protocol parameters that may change through soft protocol updates. The current parameters of interest both relate to transaction construction:

 - max tx size
 - fee policy

These two parameters are needed when building a transaction before running the coin selection.

Relevant parameter functions are:

```haskell
byronGenesisConfig
  :: Ouroboros.Consensus.Byron.Ledger.Config.BlockConfig
  -> Cardano.Chain.Genesis.Config

configGenesisData :: Cardano.Chain.Genesis.Config -> GenesisData

gdProtocolParameters
  :: Cardano.Chain.Genesis.GenesisData
  -> Cardano.Chain.Update.ProtocolParameters

ppMaxTxSize :: ProtocolParameters -> Natural

ppTxFeePolicy :: ProtocolParameters -> TxFeePolicy
```

The state machine types are in `Ouroboros.Network.Protocol.LocalStateQuery.Client`.
With cardano-node since revision 0172a96e16f46f1688da6dd8294481bb995fd72d, a new entry is added to NodeToClientProtocols that we can use. In our network layer `Cardano.Wallet.Byron.Network` we need to make room for this protocol.

Example code exists in `Ouroboros.Network.Protocol.LocalStateQuery.Example`:

``` haskell
localStateQueryClient
  :: forall block query result m. Applicative m
  => [(Point block, query result)]
  -> LocalStateQueryClient block query m
                           [(Point block, Either AcquireFailure result)]
```

In the wallet engine are already blockchain parameters stored in the database with every wallet checkpoint.

Note that for transaction construction, only the parameters for the _network tip_ are relevant. A transaction obviously cannot be inserted into the chain under the network tip.

The node only maintains local state for the last _k_ blocks from its tip.

The type `Cardano.Chain.Update.ProtocolParameters` is the values that the node supports updates for. The intersection of this type and `Cardano.Wallet.Primitive.Types.BlockchainParameters` is the following three fields:
      - ppSlotDuration (getSlotLength)
      - ppMaxTxSize
      - ppTxFeePolicy

# Decision

## Option A - parameters from checkpoint

The present code structure has the full bundle of parameters stored in each wallet checkpoint. So it's convenient to use the parameters from the _wallet tip_ (not network tip) when constructing a transaction.

[...]

PR #1580, opened 2020-04-17
Add LocalStateQuery to byron network layer

rvl:
### Issue Number

Relates to #1577 / [ADP-83](https://jira.iohk.io/projects/ADP/issues/ADP-83).

### Overview

- [x] Updates ouroboros-network and cardano-node to latest master versions
- [x] Initial try at LocalStateQuery client in network layer.
- [x] Get latest tx parameters when fetching the block tip.
- [x] Store tx parameters in database while following the chain.
- [x] Use tx parameters from database when creating transactions.

### Comments

The TRANS_CREATE integration tests which check the fee should cover this. The only way the wallet can get TxParameters is through the node protocol parameters state.

### Next PRs
- [ ] Marcin says to use `defaultCodecs` or `clientCodecs` rather than constructing our own codecs. These functions require a `BlockConfig ByronBlock`. Which means that we would need to modify parseGenesisData, etc, etc. So this can be another PR ⇒ #1606.
- [ ] Integration test where new protocol parameters are proposed and accepted. It should test that the new tx max size and fee policy apply. This will require cardano-cli to update the node's parameters. However the update proposal function of cardano-cli is not yet complete.

### DIFF
--- a/lib/byron/src/Cardano/Wallet/Byron/Compatibility.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron/Compatibility.hs
@@ -119,7 +119,6 @@ import Ouroboros.Network.NodeToClient
 import Ouroboros.Network.Point
     ( WithOrigin (..) )
 
-import qualified Cardano.Chain.Common as CC
 import qualified Cardano.Chain.Update as Update
 import qualified Cardano.Chain.Update.Validation.Interface as Update
 import qualified Cardano.Crypto.Hashing as CC
@@ -371,7 +370,7 @@ fromTxFeePolicy (TxFeePolicyTxSizeLinear (TxSizeLinear a b)) =
     W.LinearFee
         (Quantity (lovelaceToDouble a))
         (Quantity (rationalToDouble b))
-        (Quantity 0)
+        (Quantity 0) -- certificates do not exist for Byron
   where
     lovelaceToDouble :: Lovelace -> Double
     lovelaceToDouble = fromIntegral . unsafeGetLovelace
@@ -394,6 +393,17 @@ fromBlockCount (BlockCount k) =
 fromMaxTxSize :: Natural -> Quantity "byte" Word16
 fromMaxTxSize =
     Quantity . fromIntegral
+
+txParametersFromPP :: Update.ProtocolParameters -> W.TxParameters
+txParametersFromPP pp = W.TxParameters
+    { getFeePolicy = fromTxFeePolicy $ Update.ppTxFeePolicy pp
+    , getTxMaxSize = fromMaxTxSize $ Update.ppMaxTxSize pp
+    }
+
+-- | Pluck the blockchain parameters relevant to creating transactions out of
+-- the cardano-chain update state record.
+txParametersFromUpdateState :: Update.State -> W.TxParameters
+txParametersFromUpdateState = txParametersFromPP . Update.adoptedProtocolParameters
 
 -- | Convert non AVVM balances to genesis UTxO.
 fromNonAvvmBalances :: GenesisNonAvvmBalances -> [W.TxOut]
@@ -418,12 +428,7 @@ fromGenesisData (genesisData, genesisHash) =
             , getActiveSlotCoefficient =
                 W.ActiveSlotCoefficient 1.0
             }
-        , txParameters = W.TxParameters
-            { getFeePolicy =
-                fromTxFeePolicy . ppTxFeePolicy . gdProtocolParameters $ genesisData
-            , getTxMaxSize =
-                fromMaxTxSize . ppMaxTxSize . gdProtocolParameters $ genesisData
-            }
+        , txParameters = txParametersFromPP . gdProtocolParameters $ genesisData
         }
     , fromNonAvvmBalances . gdNonAvvmBalances $ genesisData
     )
@@ -434,20 +439,3 @@ fromNetworkMagic (NetworkMagic magic) =
 
 fromProtocolMagicId :: ProtocolMagicId -> W.ProtocolMagic
 fromProtocolMagicId = W.ProtocolMagic . fromIntegral . unProtocolMagicId
-
--- | Pluck the blockchain parameters relevant to creating transactions out of
--- the cardano-chain update state record.
-txParametersFromUpdateState :: Update.State -> W.TxParameters
-txParametersFromUpdateState = mk . Update.adoptedProtocolParameters
-  where
-    mk :: Update.ProtocolParameters -> W.TxParameters
-    mk pp = W.TxParameters
-        { getFeePolicy = feePolicy $ Update.ppTxFeePolicy pp
-        , getTxMaxSize = Quantity . fromIntegral $ Update.ppMaxTxSize pp
-        }
-    feePolicy :: CC.TxFeePolicy -> W.FeePolicy
-    feePolicy (CC.TxFeePolicyTxSizeLinear (CC.TxSizeLinear a m)) =
-        W.LinearFee
-            (Quantity (fromIntegral (CC.lovelaceToInteger a)))
-            (Quantity (fromRational m))
-            (Quantity 0) -- certificates do not exist for Byron
--- a/lib/byron/src/Cardano/Wallet/Byron/Compatibility.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron/Compatibility.hs
@@ -136,24 +136,28 @@ type NodeVersionData =
 -- Chain Parameters
 
 
-mainnetBlockchainParameters :: W.BlockchainParameters
-mainnetBlockchainParameters = W.BlockchainParameters
-    { getGenesisBlockHash = W.Hash $ unsafeFromHex
-        "5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb"
-    , getGenesisBlockDate =
-        W.StartTime $ posixSecondsToUTCTime 1506203091
-    , getFeePolicy =
-        W.LinearFee (Quantity 155381) (Quantity 43.946) (Quantity 0)
-    , getSlotLength =
-        W.SlotLength 20
-    , getEpochLength =
-        W.EpochLength 21600
-    , getTxMaxSize =
-        Quantity 8192
-    , getEpochStability =
-        Quantity 2160
-    , getActiveSlotCoefficient =
-        W.ActiveSlotCoefficient 1.0
+mainnetBlockchainParameters :: W.GenesisBlockParameters
+mainnetBlockchainParameters = W.GenesisBlockParameters
+    { staticParameters = W.BlockchainParameters
+        { getGenesisBlockHash = W.Hash $ unsafeFromHex
+            "5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb"
+        , getGenesisBlockDate =
+            W.StartTime $ posixSecondsToUTCTime 1506203091
+        , getSlotLength =
+            W.SlotLength 20
+        , getEpochLength =
+            W.EpochLength 21600
+        , getEpochStability =
+            Quantity 2160
+        , getActiveSlotCoefficient =
+            W.ActiveSlotCoefficient 1.0
+        }
+    , txParameters = W.TxParameters
+        { getFeePolicy =
+            W.LinearFee (Quantity 155381) (Quantity 43.946) (Quantity 0)
+        , getTxMaxSize =
+            Quantity 8192
+        }
     }
 
 -- NOTE
@@ -392,25 +396,29 @@ fromNonAvvmBalances (GenesisNonAvvmBalances m) =
     fromTxOut . uncurry TxOut <$> Map.toList m
 
 -- | Convert genesis data into blockchain params and an initial set of UTxO
-fromGenesisData :: (GenesisData, GenesisHash) -> (W.BlockchainParameters, [W.TxOut])
+fromGenesisData :: (GenesisData, GenesisHash) -> (W.GenesisBlockParameters, [W.TxOut])
 fromGenesisData (genesisData, genesisHash) =
-    ( W.BlockchainParameters
-        { getGenesisBlockHash =
-            W.Hash . CC.hashToBytes . unGenesisHash $ genesisHash
-        , getGenesisBlockDate =
-            W.StartTime . gdStartTime $ genesisData
-        , getFeePolicy =
-            fromTxFeePolicy . ppTxFeePolicy . gdProtocolParameters $ genesisData
-        , getSlotLength =
-            fromSlotDuration . ppSlotDuration . gdProtocolParameters $ genesisData
-        , getEpochLength =
-            fromBlockCount . gdK $ genesisData
-        , getTxMaxSize =
-            fromMaxTxSize . ppMaxTxSize . gdProtocolParameters $ genesisData
-        , getEpochStability =
-            Quantity . fromIntegral . unBlockCount . gdK $ genesisData
-        , getActiveSlotCoefficient =
-            W.ActiveSlotCoefficient 1.0
+    ( W.GenesisBlockParameters
+        { staticParameters = W.BlockchainParameters
+            { getGenesisBlockHash =
+                W.Hash . CC.hashToBytes . unGenesisHash $ genesisHash
+            , getGenesisBlockDate =
+                W.StartTime . gdStartTime $ genesisData
+            , getSlotLength =
+                fromSlotDuration . ppSlotDuration . gdProtocolParameters $ genesisData
+            , getEpochLength =
+                fromBlockCount . gdK $ genesisData
+            , getEpochStability =
+                Quantity . fromIntegral . unBlockCount . gdK $ genesisData
+            , getActiveSlotCoefficient =
+                W.ActiveSlotCoefficient 1.0
+            }
+        , txParameters = W.TxParameters
+            { getFeePolicy =
+                fromTxFeePolicy . ppTxFeePolicy . gdProtocolParameters $ genesisData
+            , getTxMaxSize =
+                fromMaxTxSize . ppMaxTxSize . gdProtocolParameters $ genesisData
+            }
         }
     , fromNonAvvmBalances . gdNonAvvmBalances $ genesisData
     )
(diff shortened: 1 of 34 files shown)


rvl:
@KtorZ This still needs to be cleaned up and finished but can you please check the changes in `Byron/Network.hs`?

lib/byron/src/Cardano/Wallet/Byron/Compatibility.hs
```diff
@@ -136,24 +141,28 @@ type NodeVersionData =
...
+        , getEpochStability =
+            Quantity 2160
+        , getActiveSlotCoefficient =
+            W.ActiveSlotCoefficient 1.0
+        }
+    , txParameters = W.TxParameters
+        { getFeePolicy =
+            W.LinearFee (Quantity 155381) (Quantity 43.946) (Quantity 0)
+        , getTxMaxSize =
+            Quantity 8192
```
KtorZ: To be double-checked. I am pretty sure that we put there the _current_ tx max size, but I am not sure it was the genesis one. I know it used to be 4096 bytes _at some point_.

KtorZ: So, I just checked and indeed:

```json
        "maxTxSize": "4096",
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -788,6 +803,20 @@ restoreBlocks ctx wid blocks nodeTip = db & \DBLayer{..} -> do
...
+    :: forall ctx s k.
+        ( HasLogger WalletLog ctx
+        , HasDBLayer s k ctx
+        )
+    => ctx
+    -> WalletId
+    -> TxParameters
+    -> ExceptT ErrNoSuchWallet IO ()
+saveParams ctx wid params = db & \DBLayer{..} ->
+   mapExceptT atomically $ putTxParameters (PrimaryKey wid) params
```
KtorZ, line 819: What about trying to get this done in the same atomic block storing the corresponding checkpoint to keep the data consistent in case a failure arises just at the bad time.

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
```diff
@@ -133,17 +133,22 @@ Checkpoint
...
     Primary checkpointWalletId checkpointSlot
     Foreign Wallet checkpoint checkpointWalletId ! ON DELETE CASCADE
     deriving Show Generic
 
+TxParameters
+    txParametersWalletId          W.WalletId  sql=wallet_id
+    txParametersFeePolicy         W.FeePolicy sql=fee_policy
+    txParametersTxMaxSize         Word16      sql=tx_max_size
+    Primary txParametersWalletId
+    deriving Show Generic
```
KtorZ, line 151: There's no slot or epoch reference here. How do we deal with rollbacks, especially on epoch boundaries?
This is the reason why parameters ended up being stored with checkpoints in the first place (although a bit excessive since most of the stored parameters are actually static ... ). But these needs to be consistent with checkpoints, so when we rollback, we ideally need to also rollback to the previous tx parameters values. 

So we either need to store them with checkpoints, or make sure to keep this table and the checkpoint one in sync :thinking: ... Tx parameters however don't necessarily change on every rollback though, but only when we cross an epoch boundary.

KtorZ: Also, on a slightly different note, if we really end up with a different table here, what about choosing a slightly more general name "ProtocolParameters" or "UpdatableParameters" because, there could be more in practice and it'll be sad to have the table called "TxParameters" if we end up adding more params that aren't tx-specific. 

My suggestion would be however to do the opposite here:

a) Keep the tx-specific parameters inside the `checkpoint` table. 
b) Move the static parameters (slot length, epoch length...) inside a "static" table (or just keep them in a record in-memory; we know mainnet, and testnet are provided at runtime anyway).

lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -354,7 +368,8 @@ follow nl tr cps yield header =
         Right (RollForward cursor' tip (blockFirst : blocksRest)) -> do
             let blocks = blockFirst :| blocksRest
             traceWith tr $ MsgApplyBlocks (header <$> blocks)
-            action <- yield blocks tip
+            params <- getTxParameters nl
```
KtorZ, line 371: I wonder what's the cost of this. In practice, it should only happen every 1000 blocks in "restoration mode" so I assume it's a small cost. Yet, since we know it can only change with epoch changes, if it ends up making the wallet choke every 1000 blocks we may need something clever here :thinking:

KtorZ (commented):
I've checked everything but `Wallet/Byron/Network` which I'll keep for a bit later. 
I wonder what motivated the removal of the tx parameters from the checkpoints, since they were initially added to the checkpoint precisely for this moment when we would have to track protocol updates :thinking: ?

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
Anviking: Do we actually need to store the `TxParameters` in the DB?

Why not refetch them on each tx creation (or somewhat cached), and not have to worry about rollbacks and consistency?

rvl:
The only Tx parameters which are relevant for creating transactions are the Tx parameters for the network tip. So there is no point worrying about old tx parameters. If you look at the network layer changes, it is updating the parameters immediately after the tip changes. So chain switches are handled and there is no need to worry about rollback.

If you look at the blockchain parameters which are being stored in the database for each checkpoint:
- 2 of them are completely static and will never change
- 3 are not covered by the UpdateParameters data returned from the node
- 2 are relevant to transactions, so historical data doesn't matter.
- and only one - slot length - can possibly change. But our app can't handle updates to slot length.
So I think it was premature implementation to add this table to the database.

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
rvl: @KtorZ response to your first comment is below.

@Anviking You are right there is no need to store these parameters in the DB. However, if we do cache them in the DB, it is possible to create transactions without the network layer being active.

lib/core/src/Cardano/Wallet/Network.hs
rvl: There is almost zero cost -- it's reading a TVar.

KtorZ:
> So chain switches are handled and there is no need to worry about rollback.

This is surely true if the all forks include the protocol update. But are we sure of that? What is the actual threshold for an update to land on the chain? And, regardless, it is possible that we end up on an adversarial fork that does not include a particular update :thinking:  ? On the principle, I agree with you. It's probably enough to keep only the latest value, although I wonder if we aren't overlooking something.

> and only one - slot length - can possibly change. 

Not even. This is only true for Byron and will just not happen. This was however removed from the updatable parameters in Shelley because of the obvious complications it causes. So you're right, it really isn't a concern.

Screenshot from 2020-04-27 12-03-29

rvl:
bors try

lib/core/src/Cardano/Wallet.hs
rvl: I don't think it will be necessary to do this because the tip tx parameters and wallets checkpoints are updated independently and asynchronously anyway.

lib/byron/src/Cardano/Wallet/Byron/Compatibility.hs
rvl: OK fixed.

rvl:
bors try

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -594,6 +596,19 @@ newDBLayer trace defaultFieldValues mDatabaseFile = do
...
+        {-----------------------------------------------------------------------
+                                 Blockchain Parameters
+        -----------------------------------------------------------------------}
+
+        , putTxParameters = \(PrimaryKey wid) txp -> ExceptT $ do
+            selectWallet wid >>= \case
+                Nothing -> pure $ Left $ ErrNoSuchWallet wid
+                Just _  -> Right <$> do
+                    deleteWhere [TxParametersWalletId ==. wid]
+                    insert_ (mkTxParametersEntity wid txp)
```
KtorZ: or simply [`repsert`](https://hackage.haskell.org/package/persistent-2.10.1/docs/Database-Persist-Class.html#v:repsert)

lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -221,46 +241,52 @@ withNetworkLayer
...
 
-    -- NOTE: We keep a client connection running just for accessing the node tip.
+    -- NOTE: We keep a client connection running for accessing the node tip,
+    -- submitting transactions, and querying parameters.
     -- It is safe to retry when the connection is lost here because this client
     -- doesn't really do anything but sending dummy messages to get the node's
     -- tip. It doesn't rely on the intersection to be up-to-date.
-    nodeTipQ <- atomically newTQueue
-    let nodeTipClient = const $ mkNetworkClient tr bp nodeTipQ localTxSubmissionQ
+    nodeTipVar <- atomically $ newTVar TipGenesis
```
KtorZ, line 261: :+1: Also.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -221,46 +241,52 @@ withNetworkLayer
...
+    txParamsVar <- atomically $ newTVar (W.txParameters gbp)
+    nodeTipClient <- mkTipSyncClient tr bp
+        localTxSubmissionQ
+        (atomically . writeTVar nodeTipVar)
+        (atomically . writeTVar txParamsVar)
```
KtorZ, line 266: Okay. Works because `writeTVar` is non-blocking, like `readTVar` (which I still find very error-prone / confusing when compared with the MVar API...).

lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -406,43 +435,102 @@ type NetworkClient m = OuroborosApplication
...
-    -> W.BlockchainParameters
+    => W.BlockchainParameters
         -- ^ Static blockchain parameters
     -> TQueue m (ChainSyncCmd m)
         -- ^ Communication channel with the ChainSync client
-    -> TQueue m (LocalTxSubmissionCmd m)
-        -- ^ Communication channel with the LocalTxSubmission client
     -> NetworkClient m
-mkNetworkClient tr bp chainSyncQ localTxSubmissionQ =
+mkWalletClient bp chainSyncQ =
```
KtorZ, line 447: Yeah. Good thing to split up the clients. It wasn't really clear before how transactions were actually pushed through any of the client.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -406,43 +435,102 @@ type NetworkClient m = OuroborosApplication
...
+            queryLocalState (getTipPoint tip)
+
+        queryLocalState pt =
+            (localStateQueryQ `send` CmdQueryLocalState pt) >>= \case
+                Left (e :: ErrNetworkUnavailable) ->
+                    traceWith tr $ MsgLocalStateQueryError $ show e
+                Right (Left (e :: AcquireFailure)) ->
+                    traceWith tr $ MsgLocalStateQueryError $ show e
+                Right (Right ls) ->
+                    onTxParamsUpdate' $ txParametersFromUpdateState ls
```
KtorZ: slight preference for `where` clause here to make the overall function body more readable.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -406,43 +435,102 @@ type NetworkClient m = OuroborosApplication
...
+    localStateQueryQ <- atomically newTQueue
+
+    onTxParamsUpdate' <- debounce $ \txParams -> do
+        traceWith tr $ MsgTxParameters txParams
+        onTxParamsUpdate txParams
+
+    let
+        onTipUpdate' tip = do
+            traceWith tr $ MsgNodeTip tip
+            onTipUpdate tip
```
KtorZ, line 486: Why not doing the tracing inside the callback itself :thinking: ?

lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -739,6 +893,105 @@ localTxSubmission tr queue channel =
...
+--        │           ▼       │
+--        │       ┌───────────┴───┐              Result
+--        │       │   Acquiring   │◀─────────────────────┐
+--        │       └───┬───────────┘                      │
+-- Release│           │       ▲                          │
+--        │           │       │                          │
+--        │  Acquired ▼       │ ReAcquire                │
+--        │       ┌───────────┴───┐             ┌────────┴───────┐
+--        └───────┤   Acquired    │────────────>│   Querying     │
+--                └───────────────┘             └────────────────┘
```
KtorZ, line 925: :heart_eyes:

lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -680,6 +744,96 @@ chainSyncWithBlocks tr bp queue channel = do
...
+        -- After an intersection is found, we return to idle with the sync flag
+        -- set.
+        clientStIntersect
+            :: ClientStIntersect ByronBlock (Tip ByronBlock) m Void
+        clientStIntersect = ClientStIntersect
+            { recvMsgIntersectFound = \_intersection _tip ->
+                ChainSyncClient $ clientStIdle True
+            , recvMsgIntersectNotFound = \_tip ->
+                ChainSyncClient $ clientStIdle False
+            }
```
KtorZ, line 837: Okay, I think this works fine because the node will always start by replying with a `Rollbackward` to a `RequestNext`. Without that, we would have to wait at least for the next request to override the initial value for the tip. 

:+1:

lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -739,6 +893,105 @@ localTxSubmission tr queue channel =
...
+        clientStAcquired1
+            :: (LocalStateQueryResult -> m ())
+            -> LocalStateQuery.ClientStAcquired ByronBlock (Query ByronBlock) m Void
+        clientStAcquired1 respond = LocalStateQuery.SendMsgQuery GetUpdateInterfaceState (clientStQuerying respond)
+
+        clientStAcquired2
+            :: m (LocalStateQuery.ClientStAcquired ByronBlock (Query ByronBlock) m Void)
+        clientStAcquired2 = atomically (readTQueue queue) <&> \case
+            CmdQueryLocalState pt respond ->
+                LocalStateQuery.SendMsgReAcquire pt (clientStAcquiring respond)
```
KtorZ: Is this (re-acquiring) preferred over releasing the state and acquiring it back later ?

KtorZ (approved):
Looks good overall. I haven't quite played with the code but the changes in the networking side looks sensible, as well as in the database.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
rvl: TVars don't offer any synchronisation apart from atomically updating the value. TVars are like a basketball scoreboard and MVars are like boxes.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
rvl: Yeah, there were multiple threads reading the queue, so I think the tx would have been sent via the node client of the first thread to wake up after the message was pushed to the queue.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
rvl: Ok I have split the result handling out.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
rvl: It would be slightly more code and this way keeps all the tracing (both successes and failures) in the same function.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
rvl: Yes the both the `RollForward` and `RollBackward` messages gives us a new tip, so we update the TVar and then ask the node what's next.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
rvl: Yes, I added a comment.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
rvl: Good idea, thanks. I lost faith in the persistent-sqlite update functions...

lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -406,43 +435,104 @@ type NetworkClient m = OuroborosApplication
...
-            doNothingProtocol
+            let tr' = contramap MsgLocalStateQuery tr in
+            InitiatorProtocolOnly $ MuxPeerRaw $ \channel ->
+                localStateQuery tr' localStateQueryQ channel
         }
         NodeToClientV_2
 
+-- | Return a function to run an action only if its single parameter has changed
+-- since the previous time it was called.
+debounce :: (Eq a, MonadSTM m) => (a -> m ()) -> m (a -> m ())
```
Anviking, line 517: 🆒

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
```diff
@@ -133,17 +133,23 @@ Checkpoint
     checkpointBlockHeight       Word32       sql=block_height
     checkpointGenesisHash       BlockId      sql=genesis_hash
     checkpointGenesisStart      UTCTime      sql=genesis_start
-    checkpointFeePolicy         W.FeePolicy  sql=fee_policy
     checkpointSlotLength        Word64       sql=slot_length
     checkpointEpochLength       Word32       sql=epoch_length
-    checkpointTxMaxSize         Word16       sql=tx_max_size
```
Anviking, line 139: 👍

Anviking (approved):
lgtm

rvl:
Thanks. Rebased and squashed a little.

bors r+

PR #1581, opened 2020-04-19
Revert "Lower node logging severity to Notice in bench:restore"

Anviking:
# Issue Number

#1458 

# Overview

- [x] Lowering to notice was probably a bad idea. Logs were too large when syncing without cache causing bench to fail.

# Comments

Last logged line `a9b4bb2@2927618`, ≈500 MB size, https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/478


rvl:
So using the default cardano-node logging config and minimum severity at Notice, a chain sync generates 500MB of logs? That seems like a bug to me.

KtorZ:
Last time I synced a node on AWS, it also destroyed my cloudwatch free-tier in a few hours only :grimacing: .... The node is ... verbose in its default configuration.

Anviking:
Overwhelming output looks like a known issue: https://github.com/input-output-hk/cardano-node/issues/733

rvl (approved):
Looks like cardano-node will get a fix soon - input-output-hk/cardano-node#773.

But if you want to merge this in the mean time, do add a source comment wlth a link to the issue.

Anviking:
Benchmarks seem to be passing now… so seems we don't have to merge this.

PR #1584, opened 2020-04-21
remove passphrase validations on endpoints using passphrase

KtorZ:
# Issue Number

N/A

# Overview

- 0364e701cbc0f3bcb7eee740f59137ae377de701
  :round_pushpin: **remove passphrase validations on endpoints using passphrase**
  It is still required to provide a valid and "strong" passphrase
  when creating a wallet though. But if a wallet has been created
  with a passphrase that is shorter than the expected validation,
  then the API won't reject it and simply go forward trying to use it.

  This only use-case and possible way for introducing a shorter
  passphrase is by importing a legacy wallet from an already encrypted
  private key. In this case, we can't control rules regarding the
  passphrase and, the passphrase might as well be one single character.

# Comments

:warning: I am expecting some CI failures here, i haven't particularly look at the integration tests. Most should be fine because it won't impact current behavior, but maybe we have some scenario that are precisely checking that we can't use a short passphrase on legacy wallets.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -714,7 +714,7 @@ byronServer byron icarus ntp =
     byronTransactions =
              (\wid tx -> withLegacyLayer wid
                  (byron , do
-                    let pwd = getApiT $ tx ^. #passphrase
+                    let pwd = coerce (getApiT $ tx ^. #passphrase)
                     genChange <- rndStateChange byron wid pwd
                     postTransaction byron genChange wid tx
                  )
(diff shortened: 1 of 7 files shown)


KtorZ:
bors try

paweljakubas (approved):
LGTM (ie., hack to further service invalid spending passwords) - also feel some integration wallet tests will fail, maybe some stuff will not compile because of "byron-raw" -> "lenient"

Maybe some test added to illustrate the boundary case it should service?

KtorZ:
Weird.. The build passed but the result wasn't properly to Github. Merging manually.

https://buildkite.com/input-output-hk/cardano-wallet/builds/7049

PR #1587, opened 2020-04-22
Cabal: Set -Werror and -Ox only with release flag

rvl:
Makes Cabal builds faster, like the `stack build --fast` flag.

Nothing is changed when the development flag is unset.

Some components are missing `-O`, and most others are at `-O2`. However, from what I've read, `-O2` often doesn't help, so you only use it if your benchmarks say so.


KtorZ:
Note that for the libraries, I went the other way around, and chose to enable Werrors and O2 only when the "release" flag is set. So by default, it builds with no optimization and doesn't fail on warnings. 

Thoughts?

rvl:
For libraries that other people will use, potentially with a different GHC version, `-Werror` by default is a bad option. Also for development `-Werror` is annoying.

So I am happy to exchange a release flag which is the inverse of development, and have it disabled by default.

I also think that unless there is a known good reason to use `-O2`, just enable optimization at the GHC default level with `-O`.

.buildkite/pipeline.yml
```diff
@@ -21,7 +21,10 @@ steps:
       system: x86_64-linux
 
   - label: 'Check Cabal Configure'
-    command: 'nix-shell --run "cabal v2-update && cabal v2-configure --enable-tests --enable-benchmarks"'
+    command:
+      - 'nix-shell --run "cabal v2-update"'
+      - 'nix-shell --run "cabal v2-configure --enable-tests --enable-benchmarks"'
+      - 'nix-shell --run "cabal v2-configure -frelease --enable-tests --enable-benchmarks"'
```
KtorZ, line 27: Running it twice, one without and once with the flag? Is this intended? (why not only with the flag)

rvl: It was intended. In case someone puts something silly under `if (!flag(release))`

KtorZ (approved):
Ok. This will still however runs the tests and compile the doc without werror in CI right? I'd assume we would want the opposite, so that's what's on master always compile with `-frelease`

rvl:
I forgot to do that. Do you know how to set a flag for all packages in a stack build?
Otherwise it looks like the build command will need to be like this:
```
stack build --flag cardano-wallet-core:release --flag cardano-wallet-jormungandr:release --flag cardano-wallet-byron:release ...
```

KtorZ:
@rvl the only "trick" I know is to add the flags directly in the `stack.yaml`, but it still has to be done for each package. Stack "supports" wildcards here, but it'll apply the wildcard to all local packages AND extra dependencies, when it applies. So unless our flag name is unique, bad things will probably happen with `--flag *:release` :upside_down_face: ...

rvl:
OK I fixed the stack package flags with a little bit of turtle scripting.

.buildkite/rebuild.hs
```diff
@@ -216,6 +220,20 @@ checkUnclean dryRun dir = do
...
+    logCommand cmd args
+    inprocWithErr cmd args empty
+        & useStderrOnly
+        & grep (invert (begins "Cloning"))
+        & textLines
+  where
+    useStderrOnly = fmap (either id (const ""))
+    textLines shell = fold (lineToText <$> shell) Fold.list
+    cmd = "stack"
+    args = ["ide", "packages"]
```
KtorZ: :tada:

KtorZ (approved):

KtorZ:
Some pretty interesting flags set in there:

```
--flag 'trace: Using IOHK default nixpkgs:release' --flag 'trace: Using IOHK default nixpkgs:release' --flag 'trace: gitSource.nix: /nix/store/bsxg3s82zwjidjnnfx7hpf1yzq3zhh2z-source does not seem to be a git repository,:release' --flag 'assuming it is a clean checkout.:release' --flag 'trace: WARNING: `cleanGit` called on /nix/store/bsxg3s82zwjidjnnfx7hpf1yzq3zhh2z-source without a `name`. Consider adding `name = "bsxg3s82zwjidjnnfx7hpf1yzq3zhh2z-source;"`:release' --flag 'trace: Using index-state: 2020-04-01T00:00:00Z for source:release' --flag 'trace: Using IOHK default nixpkgs:release' --flag 'trace: WARNING: license "GPL-2.0-or-later" not found:release' --flag 'trace: [alex-plan-to-nix-pkgs] cabal new-configure --with-ghc=ghc --with-ghc-pkg=ghc-pkg:release' --flag 'trace: [happy-plan-to-nix-pkgs] cabal new-configure --with-ghc=ghc --with-ghc-pkg=ghc-pkg:release' --flag 'trace: [hscolour-plan-to-nix-pkgs] cabal new-configure --with-ghc=ghc --with-ghc-pkg=ghc-pkg:release'
```

I guess the `& grep (invert (begins "Cloning"))` is grepping more than it should :sweat_smile:

rvl:
> Some pretty interesting flags set in there

Oh damn... I guess I could add `begins "trace:"` as well.

The `stack ide` command has stolen defeat from the jaws of victory by printing its result on stderr not stdout.

rvl:
I changed it to just list `**/*.cabal` to find the package names.

PR #1588, opened 2020-04-22
Support public key derivation in CLI

Anviking:
## Issue Number

ADP-190, #1594

## Overview

- [x] Add new `XPrvOrXPub` type
- [x] Add hex coders for `XPrvOrXPub`
- [x] Test `derive . toPublic == toPublic . derive (for soft indices)`
- [x] Implement `inspect` and `toPublic` using the `CliKeyScheme` instead of separate functions
- [x] `derive` and `inspect` now accepts public keys as well
- [x] Extend roundtrip tests to test `inspect` and `toPublic`
- [ ] <s>Remove some boilerplate</s>
- [ ] <s>Consider parameterising by `prv pub any` instead of just `key` for slightly more helpful error messages.</s>
- [ ] <s>Maybe `Data.Codec` could be used?</s>

### DIFF
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -73,11 +73,13 @@ module Cardano.CLI
     , CliKeyScheme (..)
     , DerivationIndex (..)
     , DerivationPath (..)
+    , XPrvOrXPub (..)
 
     , newCliKeyScheme
-    , xPrvToTextTransform
+    , keyHexCodec
     , hoistKeyScheme
     , mapKey
+    , firstHardenedIndex
 
 
     -- * Logging
@@ -167,6 +169,7 @@ import Cardano.Wallet.Primitive.AddressDerivation
     , SomeMnemonic (..)
     , WalletKey (..)
     , XPrv
+    , XPub
     , deriveRewardAccount
     , hex
     , unXPrv
@@ -197,8 +200,6 @@ import Data.Bifunctor
     ( bimap )
 import Data.ByteArray.Encoding
     ( Base (Base16), convertFromBase )
-import Data.ByteString
-    ( ByteString )
 import Data.Char
     ( toLower )
 import Data.List.Extra
@@ -389,7 +390,11 @@ data CliKeyScheme key m = CliKeyScheme
     { allowedWordLengths :: [Int]
     , mnemonicToRootKey :: [Text] -> m key
     , deriveChildKey :: key -> DerivationIndex -> m key
+    , toPublic :: key -> m key
+    , inspect :: key -> m Text
     }
+
+data XPrvOrXPub = AXPrv XPrv | AXPub XPub
 
 -- | Change the underlying monad of a @CliKeyScheme@.
 hoistKeyScheme
@@ -400,27 +405,38 @@ hoistKeyScheme eta s = CliKeyScheme
     { allowedWordLengths = allowedWordLengths s
     , mnemonicToRootKey = eta . mnemonicToRootKey s
     , deriveChildKey = \k i -> eta $ deriveChildKey s k i
+    , toPublic = eta . toPublic s
+    , inspect = eta . inspect s
     }
 
 -- | Pair of functions representing the bidirectional transformation between
--- a @XPrv@ and its 96-byte hex-encoded form.
-xPrvToTextTransform :: (XPrv -> Either String Text, Text -> Either String XPrv)
-xPrvToTextTransform = (xPrvToHexText, hexTextToXPrv)
+-- a @XPrvOrXPub@ and its hex-encoded form.
+keyHexCodec :: (XPrvOrXPub -> Either String Text, Text -> Either String XPrvOrXPub)
+keyHexCodec = (encode, decode)
   where
-    xPrvToHexText :: XPrv -> Either String Text
-    xPrvToHexText =
+    encode :: XPrvOrXPub -> Either String Text
+    encode (AXPub xpub) = return . T.pack . B8.unpack . hex . CC.unXPub $ xpub
+    encode (AXPrv xprv) =
         fmap (T.pack . B8.unpack . hex)
         . left showErr
-        . unXPrvStripPubCheckRoundtrip
+        . unXPrvStripPubCheckRoundtrip $ xprv
       where
         -- NOTE: This error should never happen from using the CLI.
         showErr ErrCannotRoundtripToSameXPrv =
             "Internal error: Failed to safely encode an extended private key"
 
-    hexTextToXPrv :: Text -> Either String XPrv
-    hexTextToXPrv txt = do
+    decode :: Text -> Either String XPrvOrXPub
+    decode txt = do
         bytes <- fromHex $ B8.pack $ T.unpack . T.strip $ txt
-        left showErr $ xPrvFromStrippedPubXPrvCheckRoundtrip bytes
+        case BS.length bytes of
+            96 -> fmap AXPrv $ left showErr $ xPrvFromStrippedPubXPrvCheckRoundtrip bytes
+            64 -> fmap AXPub . CC.xpub $ bytes
+            n -> Left . mconcat $
+                [ "Expected key to be 96 bytes in the case of a private key"
+                , " and, 64 bytes for public keys. This key is "
+                , show n
+                , " bytes."
+                ]
       where
         showErr (ErrInputLengthMismatch expected actual) = mconcat
             [ "Expected extended private key to be "
@@ -451,6 +467,8 @@ mapKey (f, g) s = CliKeyScheme
     , deriveChildKey = \k i -> do
         k' <- g k
         (deriveChildKey s k' i) >>= f
+    , toPublic = g >=> toPublic s >=> f
+    , inspect = g >=> inspect s
     }
 
 eitherToIO :: Either String a -> IO a
@@ -468,7 +486,7 @@ instance ToText DerivationPath where
 
 newtype DerivationIndex = DerivationIndex { unDerivationIndex :: Word32 }
     deriving (Show, Eq)
-    deriving newtype (Bounded, Enum)
+    deriving newtype (Bounded, Enum, Ord)
 
 firstHardenedIndex :: Word32
 firstHardenedIndex = getIndex $ minBound @(Index 'Hardened 'AddressK)
@@ -531,7 +549,7 @@ instance ToText DerivationIndex where
             then show (i - firstHardenedIndex) ++ "H"
             else show i
 
-newCliKeyScheme :: ByronWalletStyle -> CliKeyScheme XPrv (Either String)
+newCliKeyScheme :: ByronWalletStyle -> CliKeyScheme XPrvOrXPub (Either String)
 newCliKeyScheme = \case
     Random ->
         -- NOTE
@@ -545,24 +563,30 @@ newCliKeyScheme = \case
         in
             CliKeyScheme
                 (apiAllowedLengths proxy)
-                (fmap icarusKeyFromSeed . seedFromMnemonic proxy)
+                (fmap (AXPrv . icarusKeyFromSeed) . seedFromMnemonic proxy)
                 (derive CC.DerivationScheme2)
+                (toPub)
+                insp
     Trezor ->
         let
             proxy = Proxy @'API.Trezor
         in
             CliKeyScheme
                 (apiAllowedLengths proxy)
-                (fmap icarusKeyFromSeed . seedFromMnemonic proxy)
+                (fmap (AXPrv . icarusKeyFromSeed) . seedFromMnemonic proxy)
                 (derive CC.DerivationScheme2)
+                (toPub)
+                insp
     Ledger ->
         let
             proxy = Proxy @'API.Ledger
         in
             CliKeyScheme
                 (apiAllowedLengths proxy)
-                (fmap ledgerKeyFromSeed . seedFromMnemonic proxy)
+                (fmap (AXPrv . ledgerKeyFromSeed) . seedFromMnemonic proxy)
                 (derive CC.DerivationScheme2)
+                (toPub)
+                insp
   where
     seedFromMnemonic
         :: forall (s :: ByronWalletStyle).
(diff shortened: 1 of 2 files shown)


lib/cli/src/Cardano/CLI.hs
```diff
@@ -400,27 +405,38 @@ hoistKeyScheme eta s = CliKeyScheme
...
 
-    hexTextToXPrv :: Text -> Either String XPrv
-    hexTextToXPrv txt = do
+    decode :: Text -> Either String XPrvOrXPub
+    decode txt = do
         bytes <- fromHex $ B8.pack $ T.unpack . T.strip $ txt
-        left showErr $ xPrvFromStrippedPubXPrvCheckRoundtrip bytes
+        case BS.length bytes of
+            96 -> fmap AXPrv $ left showErr $ xPrvFromStrippedPubXPrvCheckRoundtrip bytes
+            64 -> fmap AXPub . CC.xpub $ bytes
```
Anviking, line 433: Should use named constant

lib/cli/src/Cardano/CLI.hs
```diff
@@ -587,18 +611,58 @@ newCliKeyScheme = \case
...
+        return $ mconcat
+            [ "extended private key: "
+            , encodeToHex xprv
+            , "\n"
+            , "chain code: "
+            , encodeToHex cc
+            ]
+    insp (AXPub key) = do
+        let bytes = CC.unXPub key
+        let (xpub, cc) = BS.splitAt 32 bytes
```
Anviking, line 657: Need to check that this is right.

lib/cli/src/Cardano/CLI.hs
KtorZ: This function really looks like it should be two functions :no_mouth: ...

KtorZ: See for instance:

- https://input-output-hk.github.io/cardano-addresses/haddock/Cardano-Address-Derivation.html#v:xpubFromBytes
- https://input-output-hk.github.io/cardano-addresses/haddock/Cardano-Address-Derivation.html#v:xprvFromBytes

Anviking: I intentionally wanted a single coder for `XPrvOrXPub` instead of ones for `XPub` and `XPrv` individually.

This way the "is the key private or public?" question `key child` has to answer can be handled by the coder and tested via roundtrips.

Anviking: Or are you referring to the fact that it is written as a tuple `(encode, decode)`?

I wanted something more opaque than
```
mapKey (encodeHex, decodeHex) scheme
```

since it is important that the transformation is bidirectional

That is why it was called:
```
mapKey xPrvToTextTransform scheme
```

need to rename it now, maybe
```
mapKey hexKeyEncoding scheme
```
or something 🤷‍♂️

lib/cli/src/Cardano/CLI.hs
Anviking: Yes https://github.com/input-output-hk/cardano-crypto/blob/2547ad1e80aeabca2899951601079408becbc92c/src/Cardano/Crypto/Wallet.hs#L144-L150

lib/cli/src/Cardano/CLI.hs
KtorZ: I don't know, it feels wrong to have a single function that decodes an xpub or xprv, I'd expect two here since, whether you want one or the other is clear from the context where these functions are called. Although I get your point, if we want one common interface where we can just call `cardano-wallet key child` irrespective of whether the key is private or public then, it's not clear from the surrounding context anymore which decoder you should call. 

Still, I'd find it nicer to go for an alternative, trying one, and then the other.

Anviking: > Still, I'd find it nicer to go for an alternative, trying one, and then the other.

`<|>` gives you less control over error messages, so I'm trying to avoid it.

>  I'd expect two here since

Well, not needed. Unless we differentiated between `xprv xpub any` and had separate encodings for all three. This way `key public` could avoid suggesting public keys are accepted 
```
$ cardano-wallet-byron key public 0000
Expected key to be 96 bytes in the case of a private key and, 64 bytes for public keys. This key is 2 bytes.
```
but then not:
```
$ cardano-wallet-byron key public 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
Input is already a public key.
```

But since it is only one command, and fairly self-explanatory I thought it would be overkill.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -400,27 +405,38 @@ hoistKeyScheme eta s = CliKeyScheme
...
             "Internal error: Failed to safely encode an extended private key"
 
-    hexTextToXPrv :: Text -> Either String XPrv
-    hexTextToXPrv txt = do
+    decode :: Text -> Either String XPrvOrXPub
+    decode txt = do
         bytes <- fromHex $ B8.pack $ T.unpack . T.strip $ txt
-        left showErr $ xPrvFromStrippedPubXPrvCheckRoundtrip bytes
+        case BS.length bytes of
+            96 -> fmap AXPrv $ left showErr $ xPrvFromStrippedPubXPrvCheckRoundtrip bytes
```
KtorZ, line 432: Gotta love these names.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -587,18 +611,58 @@ newCliKeyScheme = \case
...
+        maybe (Left err) (return . AXPub)
+            $ CC.deriveXPub
+                scheme1Or2
+                k
+                (unDerivationIndex i)
+      where
+        err = mconcat
+            [ T.unpack (toText i)
+            , " is a hardened index. Public key derivation is only possible for"
+            , " soft indices. \nIf the index is correct, please use the"
```
KtorZ, line 639: Small consistency tweak. This message uses both "indices" and "index" interchangeably, perhaps it's good to stick to one consistent wording otherwise people may start to wonder what's the difference between an indice and an index.

lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -809,17 +855,58 @@ prop_roundtripCliKeySchemeKeyViaHex :: ByronWalletStyle -> Property
...
+--                +                       +
+--                |                       |
+--                |                       |
+-- deriveChildKey |                       | deriveChildKey
+--                |                       |
+--                |                       |
+--                v                       v
+--           child xprv +-----------> child xpub
+--
+--                        toPublic
```
KtorZ, line 883: Noice :heart_eyes:

lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -809,17 +855,58 @@ prop_roundtripCliKeySchemeKeyViaHex :: ByronWalletStyle -> Property
...
+-- @
+--
+prop_publicKeyDerivation
+    :: ByronWalletStyle
+    -> DerivationIndex
+    -> XPrv
+    -> Property
+prop_publicKeyDerivation style i key = do
+    if isSoftIndex i
+    then (public k >>= derive) === (derive k >>= toPublic s)
```
KtorZ, line 893: Noice x2

KtorZ (approved):
:+1:

lib/cli/src/Cardano/CLI.hs
Anviking: I believe both `indices` and `indexes` are valid plural forms of `index`

- https://www.nasdaq.com/articles/indexes-or-indices-whats-the-deal-2016-05-12
- https://developer.apple.com/documentation/swift/collection/1641719-indices

Actually we are already using both even without this PR (though indices much less):

```
git grep "indices"
.travis.yml:  - rm -rf $HOME/.stack/indices/Hackage/00-index.tar $HOME/.stack/indices/Hackage/00-index.tar.gz
lib/cli/src/Cardano/CLI.hs:            , " soft indices. \nIf the index is correct, please use the"
lib/cli/test/unit/Cardano/CLISpec.hs:                    \derivation is only possible for soft indices. \nIf the\
lib/cli/test/unit/Cardano/CLISpec.hs:        it "derive . toPublic == toPublic . derive (for soft indices)"
lib/cli/test/unit/Cardano/CLISpec.hs:-- | For soft indices, public key derivation should be "equivalent" to private
lib/core/src/Cardano/Wallet/DB/Sqlite.hs:    -- Note: there are sorted indices on these columns.
lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs:-- Mapping of pool addresses to indices, and the slot
lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Byron.hs:    -- ^ The address derivation indices for the level of this key.
lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Byron.hs:-- | The hierarchical derivation indices for a given level/depth.
lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Random.hs:    -- that addresses will be _discovered_ from any and all account indices,

git grep "indexes"
lib/byron/test/data/cardano-node/genesis.yaml:  # Address indexes below are #501, #502, ..., #510
lib/byron/test/integration/Test/Integration/Byron/Scenario/CLI/Addresses.hs:        describe "CLI_ADDRESS_CREATE_07 - False indexes" $ do
lib/byron/test/unit/Cardano/Wallet/Byron/TransactionSpec.hs:    <*> scale (`mod` 3) arbitrary -- No need for a high indexes
lib/cli/src/Cardano/CLI.hs:                , "Please use \"H\" to denote hardened indexes. "
lib/cli/test/unit/Cardano/CLISpec.hs:                     \Please use \"H\" to denote hardened indexes. Did you \
lib/core-integration/src/Test/Integration/Scenario/API/Byron/Wallets.hs:    it "BYRON_RESTORE_08 - Icarus wallet with high indexes" $ \ctx -> do
lib/core-integration/src/Test/Integration/Scenario/API/Byron/Wallets.hs:        -- Special Icarus mnemonic where address indexes are all after the index
lib/core/src/Cardano/Byron/Codec/Cbor.hs:                , "expected two indexes but got: "
lib/core/src/Cardano/Wallet/DB/Sqlite.hs:                -- wallet backend was generating addresses indexes across the
lib/core/src/Cardano/Wallet/DB/Sqlite.hs:                -- We later changed that to instead use "hardened indexes". Yet,
lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs:-- Sequential address discovery scheme -- pending change indexes
lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs:-- | Type of derivation that should be used with the given indexes.
lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Shelley.hs:-- indexes and soft (non-hardened) indexes for various level in a non-ambiguous
lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs:-- in order to discover addresses with indexes separated by possible huge gaps.
lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs:-- expected to be ordered in sequence (first indexes, first in the list).
lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs:-- | An ordered set of pending indexes. This keep track of indexes used
lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs:-- | An empty pending set of change indexes.
lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs:-- | Update the set of pending indexes by discarding every indexes _below_ the
lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs:-- to know from where we can generate new indexes.
lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs:-- | Construct a 'PendingIxs' from a list, ensuring that it is a set of indexes
lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs:-- | Get the next change index; If every available indexes have already been
lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs:-- taken, we'll rotate the pending set and re-use already provided indexes.
lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs:        <> indentF 4 ("Change indexes: " <> indentF 4 chgsF)
lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs:    -- | We pick indexes in sequence from the first known available index (i.e.
lib/core/src/Cardano/Wallet/Primitive/Types.hs:        -- their indexes. It matters also for serialization.
lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs:        -- NOTE: No need for a crazy high indexes
lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs:        <*> scale (`mod` 3) arbitrary -- No need for a high indexes
lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs:-- indexes. For example:
lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs:-- - put checkpoint with pending indexes
lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs:-- - put checkpoint with empty pending indexes
lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs:-- currently, it'll also contains the pending indexes from the previously
lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivation/ByronSpec.hs:        it "Key derivation from seed works for various indexes" $
lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivation/ByronSpec.hs:        it "Key derivation from master key works for various indexes" $
lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivation/IcarusSpec.hs:        it "deriveAccountPrivateKey works for various indexes" $
lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivation/ShelleySpec.hs:        it "deriveAccountPrivateKey works for various indexes" $
lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/RandomSpec.hs:    it "findUnusedPath: indexes are always in the 'hardened' realm" $
lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelectionSpec.hs:        <*> scale (`mod` 3) arbitrary -- No need for a high indexes
lib/core/test/unit/Cardano/Wallet/Primitive/FeeSpec.hs:        <*> scale (`mod` 3) arbitrary -- No need for a high indexes
lib/core/test/unit/Cardano/Wallet/Primitive/TypesSpec.hs:        <*> scale (`mod` 3) arbitrary -- No need for a crazy high indexes
lib/jormungandr/test/data/jormungandr/genesis.yaml:  # Address indexes below are #501, #502, ..., #510
lib/jormungandr/test/integration/Main.hs:-- outputs and the values of the input indexes.
```

ouroboros-network has more occurrences of indices than indexes though

😅 🔫

PR #1591, opened 2020-04-23
Support bech32 encoding for CLI key derivation

Anviking:
## Issue Number

ADP-195. No issue yet.

## Overview

- [x] Added a `keyBech32Codec` abstraction for inferring the encoding automatically
- [x] Added support for reading input from stdin and chaining commands with `|`.

## Comments

Commands can be chained and work with both hex and bech32 encodings!

```
$ cardano-wallet-byron key root --wallet-style icarus --encoding bech32 -- express theme celery coral permit dizzy nothing stadium lemon tube olive brass dutch find mask \
    | cardano-wallet-byron key public
xpub1k365denpkmqhj9zj6qpaxat4dh0xetpz2qyneerrputtnwwq4jr5zyehhkjdt0qzzerzfq9cpxpyl76g79lq3k26h8cmj8fereywv6c0g4qe2

$ cardano-wallet-byron key root --wallet-style icarus -- express theme celery coral permit dizzy nothing stadium lemon tube olive brass dutch find mask \
     | cardano-wallet-byron key public
b47546e661b6c1791452d003d375756dde6cac2250093ce4630f16b9b9c0ac87411337bda4d5bc0216462480b809824ffb48f17e08d95ab9f1b91d391e48e66b
```

### DIFF
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -84,6 +84,7 @@ module Cardano.CLI
     , hoistKeyScheme
     , mapKey
     , firstHardenedIndex
+    , fullKeyEncodingDescription
 
     , Codec (..)
     , inverse
@@ -263,6 +264,7 @@ import Options.Applicative
     , flag'
     , footer
     , header
+    , headerDoc
     , help
     , helpDoc
     , helper
@@ -378,7 +380,16 @@ runCli = join . customExecParser preferences
 
 cmdKey :: Mod CommandFields (IO ())
 cmdKey = command "key" $ info (helper <*> cmds) $ mempty
-    <> progDesc "Derive keys from mnemonics."
+    <> progDesc "Derive and manipulate keys."
+    <> headerDoc (Just $ string $ mconcat
+        [ "Keys can be passed as arguments or read as standard input. Both "
+        , "bech32- and hexadecimal encodings are supported."
+        , "\n\n"
+        , "For instance:\n"
+        , "$ cardano-wallet-byron key root --wallet-style icarus --encoding bech32 -- express theme celery coral permit ... \\\n"
+        , "    | cardano-wallet-byron key public\n"
+        , "xpub1k365denpkmqhj9zj6qpax..."
+        ])
   where
     cmds = subparser $ mempty
         <> cmdKeyRoot
@@ -555,7 +566,17 @@ anyKeyCodec = Codec encode decode
     decode x
         | "xprv1" `T.isPrefixOf` x = decodeBech32 x
         | "xpub1" `T.isPrefixOf` x = decodeBech32 x
-        | otherwise                = decodeHex x
+        | otherwise                = decodeHex x <|>
+            (Left fullKeyEncodingDescription)
+
+fullKeyEncodingDescription :: String
+fullKeyEncodingDescription = mconcat
+    [ "Invalid key. Expected one of the following:\n"
+    , "- 96 bytes long hex (extended private key)\n"
+    , "- 64 bytes long hex (extended public key)\n"
+    , "- xprv1... (bech32-encoded extended private key)\n"
+    , "- xpub1... (bech32-encoded extended public key)"
+    ]
 
 -- | Map over the key type of a CliKeyScheme.
 --
@@ -780,15 +801,6 @@ defaultScheme =
     mapKey anyKeyCodec
     . mapKey keyByteStringCodec
     . newCliKeyScheme
-  where
---    setDefaultErr "" = mconcat
---        [ "Invalid key. Expected one of the following:\n"
---        , "- 96 bytes hex-encoded extended private key\n"
---        , "- 64 bytes hex-encoded extended public key\n"
---        , "- bech32-encoded extended private key starting with xprv1\n"
---        , "- bech32-encoded extended public key starting with xpub1\n"
---        ]
---    setDefaultErr x = x
 
 data KeyRootArgs = KeyRootArgs
     { _walletStyle :: ByronWalletStyle
@@ -806,8 +818,7 @@ cmdKeyRoot =
         <*> mnemonicWordsArgument
         <*> keyEncodingOption
     exec (KeyRootArgs keyType ws enc) = do
-        xprv <- mnemonicToRootKey scheme ws
-        TIO.putStrLn xprv
+        mnemonicToRootKey scheme ws >>= TIO.putStrLn
       where
         scheme :: CliKeyScheme Text IO
         scheme =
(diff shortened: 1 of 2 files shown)


lib/cli/src/Cardano/CLI.hs
```diff
@@ -409,25 +419,94 @@ hoistKeyScheme eta s = CliKeyScheme
...
+-- along with the result:
+-- >>> let (Codec _ decode) = ...
+-- >>> (x, encodeBack) <- decode
+--
+-- This allows us to have a @Codec@ for /both/ hex and bech32, and makes sure
+-- that we always encode the result back in the same encoding as we first got.
+--
+-- NOTE: Only supports one step of specialization. Could probably be fixed
+-- easily, but not needed now.
+data Codec m a b = Codec (a -> m (b, b -> m a)) (b -> m (a, a -> m b))
```
Anviking: 😬 

Not sure about the name either.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -409,25 +419,94 @@ hoistKeyScheme eta s = CliKeyScheme
...
+        bytes <- maybe (Left "todo") Right $ Bech32.dataPartToBytes dp
+        case hrp of
+            h | h == xpubHrp -> return bytes
+            h | h == xprvHrp -> return bytes
+            _ -> Left "unrecognized Bech32 Human Readable Part"
+
+    encode :: ByteString -> Either String Text
+    encode bs =
+        case BS.length bs of
+            96 -> return $ bech32 xprvHrp bs
```
Anviking, line 515: I think this can now be re-written as ` Codec (Either String) AXPrvOrXPub Text` which would remove the need to case-switch on the magic lengths here. It would also remove the need for codec composition — I think

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1721,8 +1845,23 @@ walletStyleOption defaultStyle accepted = option (eitherReader fromTextS)
     prettyStyle s =
         "  " ++ T.unpack (toText s) ++ " (" ++ fmtAllowedWords s ++ ")"
 
+
+instance FromText KeyEncoding where
+    fromText "bech32" = return Bech32Encoding
+    fromText "hex" = return HexEncoding
+    fromText _ = Left $ TextDecodingError "Invalid key encoding option."
```
Anviking, line 1869: Could have `Codec` directly and skip the intermediary `KeyEncoding´ type, I think

lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -839,8 +847,14 @@ spec = do
                 (newCliKeyScheme s)
                 (newCliKeyScheme s)
 
-        it "scheme == mapKey (fromHex . toHex) scheme"
-            $ property prop_roundtripCliKeySchemeKeyViaHex
+        it "scheme == mapKey (fromBS . toBS) scheme" $ property
+            $ propCliKeySchemeCodecRoundTrip keyByteStringCodec
```
Anviking, line 880: The notion of "roundtripping" is now made a bit more complicated when the roundtrip can depend on the values themselves.

I should have a closer look at these to make sure they are testing something valuable.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -414,10 +419,48 @@ hoistKeyScheme eta s = CliKeyScheme
...
+--
+-- This allows us to have a @Codec@ for /both/ hex and bech32, and makes sure
+-- that we always encode the result back in the same encoding as we first got.
+--
+-- NOTE: Only supports one step of specialization. Could probably be fixed
+-- easily, but not needed now.
+data Codec m a b = Codec (a -> m (b, b -> m a)) (b -> m (a, a -> m b))
+
+-- | A @Category@ instance for @Codec@ allows us to use the @(.)@ operator.
+instance Monad m => Cat.Category (Codec m) where
```
KtorZ: Why would we ever want such a thing?

lib/cli/src/Cardano/CLI.hs
```diff
@@ -409,25 +431,103 @@ hoistKeyScheme eta s = CliKeyScheme
...
+-- | Bidirectional encoding of data that allows specialization.
+--
+-- When decoding, the correponding, potentially specialized encoder is returned
+-- along with the result:
+-- >>> let (Codec _ decode) = ...
+-- >>> (x, encodeBack) <- decode
+--
+-- This allows us to have a @Codec@ for /both/ hex and bech32, and makes sure
+-- that we always encode the result back in the same encoding as we first got.
+data Codec m a b = Codec (a -> m (b, Codec m b a)) (b -> m (a, Codec m a b))
```
KtorZ, line 444: This type seems overly-abstracted to me. It _could_ (maybe) make sense in an abstract library for manipulating codecs and even there, you'd most likely want to pick a concrete type for your target encoding which will end up being `ByteString`. There's no other pragmatic use-case for a Codec.

Plus, it is very unlikely than encoding a value would fail, and in particular, would fail with the same error type. That would be a silly API. 

What about simplifying this as:

Consider instead:

```hs
data Codec (m :: * -> *) = Codec 
    { encode :: ByteString -> Text
    , decode :: Text -> m ByteString
    }
```

(and even here, it is already _too abstract_ for our need IMO ..) 

Aim for _simple_ code. Category theory is cool and can help coming up with nice abstractions. But remember that we have to maintain this code afterwards and here, you don't need to represent Codec as a utterly abstract concept to get a code that is flexible enough to support multiple types of encoding. There's a balance between flexibility and simplicity. Here, it is clearly too much on the side of flexibility.

lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -558,15 +570,15 @@ spec = do
...
+            [ "Usage:  key public [XPRV]"
             , "  Extract public key from a private key."
             , ""
             , "Available options:"
             , "  -h,--help                Show this help text"
             ]
 
         ["key", "inspect", "--help"] `shouldShowUsage`
-            [ "Usage:  key inspect XPRV"
+            [ "Usage:  key inspect [XPRV]"
```
Anviking, line 581: > [XPRV]

This can now be an XPUB also. Forgot to update it

lib/cli/src/Cardano/CLI.hs
Anviking:
```haskell
data Codec (m :: * -> *) = Codec 
    { encode :: ByteString -> Text
    , decode :: Text -> m ByteString
    }
```
Wouldn't work if we want `bech32`, `hex`, `any` to all be values of `Codec`. This is because we need to make sure to encode the value back using the same codec as we got it in, as described in the comment.

In that case one should remove even more abstractions ->  #1599.

Anviking:
Closing in favour of #1599.

While interesting, encapsulating all logic inside `CliKeyScheme` for the sake of roundtrips might be overkill here. The golden tests in #1599 should be enough for catching errors in the glue code.

PR #1592, opened 2020-04-23
NTP fix: block and force check only when asked

KtorZ:
# Issue Number

N/A

# Overview

- 2fa2f9c58ace8ab5dd542af9ae9e98995dd599e4
  :round_pushpin: **extend swagger specification: add forceNtpCheck flag to the API**
  
- d9a9ce0de466586d34456aa5fce1f39ac85dfd49
  :round_pushpin: **only force NTP check when required to do so. Otherwise, used cached result**
    Also added this to the command line, so that we can:

  ```
  cardano-wallet network clock --force-ntp-check
  ```

- 18340b7f6336d35f46c5c760b0bbc2266bad6d89
  :round_pushpin: **increase the number of required result from 2 to 3**
  aligned with what we used to have

- 96f5070de0c3f6e63bd990d5f781744881a3ba2c
  :round_pushpin: **add extra integration scenario with a forced NTP check**
  
- 3df212532ad60888d7a98e78285cba395356cc25
  :round_pushpin: **clarify usage of ntpQueryBlocking vs ntpGetStatus with comments**
  
- e29069bb024bf6b627f5d14753647578f47fce79

# Comments

Screenshot from 2020-04-23 11-46-35

### DIFF
--- a/lib/core/src/Network/Ntp.hs
+++ b/lib/core/src/Network/Ntp.hs
@@ -26,8 +26,8 @@ import Cardano.BM.Data.Tracer
     ( HasPrivacyAnnotation (..), HasSeverityAnnotation (..) )
 import Cardano.Wallet.Api.Types
     ( ApiNetworkClock (..), ApiNtpStatus (..), NtpSyncingStatus (..) )
-import Control.Monad.STM
-    ( atomically )
+import Control.Concurrent.STM
+    ( atomically, check )
 import Control.Tracer
     ( Tracer )
 import Data.Quantity
@@ -130,12 +130,17 @@ getNtpStatus
     -> IO ApiNetworkClock
 getNtpStatus client forceCheck = (ApiNetworkClock . toStatus) <$>
     if forceCheck
-    then ntpQueryBlocking client
-        -- ^ Forces an NTP check / query on the central servers, use with care
-    else atomically (ntpGetStatus client)
-        -- ^ Reads a cached NTP status from an STM.TVar so we don't get
-        -- blacklisted by the central NTP "authorities" for sending too many NTP
-        -- requests.
+    -- Forces an NTP check / query on the central servers, use with care
+    then do
+        ntpQueryBlocking client
+
+    else atomically $ do
+      -- Reads a cached NTP status from an STM.TVar so we don't get
+      -- blacklisted by the central NTP "authorities" for sending too many NTP
+      -- requests.
+      s <- ntpGetStatus client
+      check (s /= NtpSyncPending)
+      pure s
   where
     toStatus = \case
         NtpSyncPending ->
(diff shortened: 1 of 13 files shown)


KtorZ:
bors try

lib/core/src/Network/Ntp.hs
```diff
@@ -121,8 +123,11 @@ instance HasSeverityAnnotation NtpTrace where
         NtpTracePacketReceived _ _ -> Debug
         NtpTraceWaitingForRepliesTimeout _ -> Notice
 
-getNtpStatus :: NtpClient -> IO ApiNetworkClock
-getNtpStatus = fmap (ApiNetworkClock . toStatus) . ntpQueryBlocking
+getNtpStatus :: NtpClient -> Bool -> IO ApiNetworkClock
+getNtpStatus client forceCheck = (ApiNetworkClock . toStatus) <$>
+    if forceCheck
+    then ntpQueryBlocking client
+    else atomically (ntpGetStatus client)
```
rvl: Perhaps a comment that ntpQueryBlocking does not use a cached result but ntpGetStatus does.
It's not really obvious from the names.

rvl (approved):
LGTM

lib/core/src/Network/Ntp.hs
KtorZ: Good point :+1:

KtorZ: I've commented everywhere else but here :upside_down_face: .....

KtorZ: done in https://github.com/input-output-hk/cardano-wallet/commit/3df212532ad60888d7a98e78285cba395356cc25

coot: @KtorZ I think you want to do this:
```
atomically $ do
  s <- ntpGetStatus client
  check (s /= NtpSyncPending)
  pure s
```
Which will block for at most `1s`, if the ntp is currently running.

coot: @rvl if it's blocking it surly must do something; We could rename `ntpGetStatus` as `ntpGetStatusNonBlocking` but since it returns an `STM` action it cannot run the ntp protocol (unless we are evil and run `IO` inside `STM` ;).

coot (approved):

lib/core/src/Network/Ntp.hs
KtorZ: @coot rather, `ntpGetCachedStatus` could better convey the idea IMO :+1: 

> (unless we are evil and run IO inside STM

Expect the unexpectable. We will stop at nothing.

KtorZ:
:point_up: Hydra eval errors timed out somehow...

PR #1599, opened 2020-04-27
Support Bech32 keys in CLI (and rework the implementation to be less fancy)

Anviking:
# Issue Number

#1603 

Alternative implementation to PR #1591 

# Overview
 
- [x] I removed the inspect, child, public fields from `CliKeyScheme` which was the cause for several abstractions.
- [x] I replaced the roundtrip property with more basic ones.
- [x] This opens up for more untested mistakes in the CLI handlers themselves. But I tried to counteract this using point-free notation to avoid potential mixup of `k`, `k'`, `key``, etc.
- [x] Added integration tests making sure `key child`, `key public`, `key inspect` accept keys from stdin. Integration tests also hit both hex and bech32 encodings, as well as public and private derivation.

# Comments


lib/cli/src/Cardano/CLI.hs
```diff
@@ -866,14 +765,13 @@ cmdKeyPublic =
...
-    exec (KeyPublicArgs key) = do
-        orReadStdin key >>= toPublic scheme >>= TIO.putStrLn
-
-    scheme :: CliKeyScheme Text IO
-    scheme =
-        hoistKeyScheme eitherToIO
-        $ defaultScheme Icarus
-
+    exec (KeyPublicArgs key) =
+        orReadStdin key
```
Anviking, line 774: This `orReadStdin key` feels a bit strange.

`key` is the key passed as argument. `orReadStdin key` reads the key from stdIn if the arguments are empty.

It's easy to forget it, or make the mistake of using `key` directly.

Maybe it can be moved to `cmd`.

lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -864,36 +863,33 @@ spec = do
...
-            propCliKeySchemeEquality
-                (newCliKeyScheme s)
-                (newCliKeyScheme s)
+        it "derive . toPublic == toPublic . derive (for soft indices)"
+                    $ property prop_publicKeyDerivation
 
-        it "scheme == mapKey (fromBS . toBS) scheme" $ property
-            $ propCliKeySchemeCodecRoundTrip keyByteStringCodec
+        describe "Key Encoding Roundtrips" $ do
+            let decodeAny x = detectEncoding x >>= flip decodeKey x
```
Anviking: Should use the existing `decodeAnyKey `

rvl:
I'm a bit lost here but it looks like there's less code now, which is good.

Anviking:
bors try

lib/cli/src/Cardano/CLI.hs
```diff
@@ -367,114 +377,139 @@ runCli = join . customExecParser preferences
...
 cmdKey :: Mod CommandFields (IO ())
 cmdKey = command "key" $ info (helper <*> cmds) $ mempty
-    <> progDesc "Derive keys from mnemonics."
+    <> progDesc "Derive and manipulate keys."
+    <> headerDoc (Just $ string $ mconcat
+        [ "Keys can be passed as arguments or read as standard input. Both "
+        , "bech32- and hexadecimal encodings are supported."
+        , "\n\n"
+        , "For instance:\n"
+        , "$ cardano-wallet-byron key root --wallet-style icarus --encoding bech32 -- express theme celery coral permit ... \\\n"
```
KtorZ: We should avoid hard-coding the name of the binary here, since it actually depends on where this command is mounted. I'd suggest to simply stick with the commands:

```
key root --wallet-style icarus ...
```

lib/cli/src/Cardano/CLI.hs
```diff
@@ -367,114 +377,139 @@ runCli = join . customExecParser preferences
...
--- a @XPrvOrXPub@ and its hex-encoded form.
-keyHexCodec :: (XPrvOrXPub -> Either String Text, Text -> Either String XPrvOrXPub)
-keyHexCodec = (encode, decode)
+data KeyEncoding = Hex | Bech32
+    deriving (Show, Eq, Enum, Bounded)
+
+detectEncoding :: Text -> Either String KeyEncoding
+detectEncoding x
+    | "xprv1" `T.isPrefixOf` x = pure Bech32
+    | "xpub1" `T.isPrefixOf` x = pure Bech32
```
KtorZ, line 430: Detecting bech32 encoding by detecting the prefix is sort of wrong. A value could be bech32 encoded but simply has a wrong prefix. Yet, a value can't be both bech32 and hex encoded, so we may as well simply try to decode them, and then, check for the prefix in case it's bech32.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -367,114 +377,139 @@ runCli = join . customExecParser preferences
...
+        bytes <- maybe (Left dpErr) Right $ Bech32.dataPartToBytes dp
+        case hrp of
+            h | h == xpubHrp -> keyFromBytes bytes
+            h | h == xprvHrp -> keyFromBytes bytes
+            _ -> Left "unrecognized Bech32 Human Readable Part"
+      where
+        dpErr = "Internal error: Unable to extract bytes from bech32 data part"
+
+decodeHex :: Text -> Either String ByteString
+decodeHex txt = fromHex $ B8.pack $ T.unpack . T.strip $ txt
```
KtorZ: This works because hex-encoded characters are ASCII characters, but the "good way" of doing this would be by using `encodeUtf8` from `Data.Text.Encoding`.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -367,114 +377,139 @@ runCli = join . customExecParser preferences
...
+ where
+   fromHex = left (const "Invalid hex.")
+       . convertFromBase Base16
+
+encodeKey :: KeyEncoding -> XPrvOrXPub -> Either String Text
+encodeKey enc key = case enc of
+    Hex -> T.pack . B8.unpack . hex <$> bytes
```
KtorZ: Same as above, see `Data.Text.Encoding`.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -612,65 +638,71 @@ newCliKeyScheme = \case
...
+
+deriveChildKey
+    :: XPrvOrXPub
+    -> DerivationIndex
+    -> Either String XPrvOrXPub
+deriveChildKey (AXPrv k) i =
+    return . AXPrv
+        $ CC.deriveXPrv
+            CC.DerivationScheme2
+            (mempty :: Passphrase "Encryption")
```
KtorZ, line 672: Hmmm. More like `mempty :: ScrubbedBytes` or `mempty :: ByteString` but, why not. Since `Passphrase` has a `ByteArrayAccess`, it probably doesn't matter.

lib/cli/src/Cardano/CLI.hs
KtorZ: Or, simply make it the only way to provide a value. So, for key commands, inputs HAVE to be provided via stdin, and it's very easy to redirect "literals" if needed to make it like standard arguments:

```
echo ... | cardano-wallet key child
```

lib/cli/src/Cardano/CLI.hs
```diff
@@ -755,14 +774,13 @@ cmdKeyInspect =
...
-        hoistKeyScheme eitherToIO
-        . mapKey keyHexCodec
-        $ newCliKeyScheme Icarus
+    exec (KeyInspectArgs key) = do
+        orReadStdin key
+            >>= eitherToIO . go
+            >>= TIO.putStrLn
+      where
+        go :: Text -> Either String Text
+        go = fmap (\(k, _) -> inspect k) . decodeAnyKey
```
KtorZ: small note but `go` is typically used for recursive functions. So it feels like something is going to recurse but it isn't. Perhaps `action` or `run` is a better fit.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1732,8 +1750,23 @@ walletStyleOption defaultStyle accepted = option (eitherReader fromTextS)
     prettyStyle s =
         "  " ++ T.unpack (toText s) ++ " (" ++ fmtAllowedWords s ++ ")"
 
+instance FromText (XPrvOrXPub -> Either String Text) where
```
KtorZ: This is, interesting. Why not simply have an instance from `Text` to `Encoding` ? The encoding as a data-type is perfectly fine to carry on especially because the `encodeKey` arguments are homogeneous here. So this looks like needlessly complex.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -612,65 +638,71 @@ newCliKeyScheme = \case
...
-            ]
-    insp (AXPub key) = do
-        let bytes = CC.unXPub key
-        let (xpub, cc) = BS.splitAt 32 bytes
-        let encodeToHex = T.pack . B8.unpack . hex
-        return $ mconcat
+
+toPublic :: XPrvOrXPub -> Either String XPrvOrXPub
+toPublic (AXPrv xprv) = return . AXPub . CC.toXPub $ xprv
+toPublic (AXPub _) = Left "Input is already a public key."
```
KtorZ, line 662: Wouldn't it be find though to just have an identity here :thinking: ? It's kinda nicer for building scripts that accepts either a private or public key. There's nothing wrong in practice although, it's nice to also have an explicit message indicating the user he's probably doing something he didn't intended. 

No strong opinion here, but just a thought.

KtorZ (approved):
Looks good, although I have some reserve about the dual argument / stdin support. I think we should only have one way for passing inputs (stdin) and stick to it.

lib/cli/src/Cardano/CLI.hs
rvl: If you really want the program name, I think it would be fine to do `unsafePerformIO getProgName`.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -367,114 +377,139 @@ runCli = join . customExecParser preferences
...
-hoistKeyScheme eta s = CliKeyScheme
-    { allowedWordLengths = allowedWordLengths s
-    , mnemonicToRootKey = eta . mnemonicToRootKey s
-    , deriveChildKey = \k i -> eta $ deriveChildKey s k i
-    , toPublic = eta . toPublic s
-    , inspect = eta . inspect s
-    }
+eitherToIO :: Either String a -> IO a
+eitherToIO (Right a) = return a
+eitherToIO (Left e) = hPutStrLn stderr e >> exitFailure
```
rvl:
```suggestion
eitherToIO = either die return
```

rvl (commented):
Looks good

lib/cli/src/Cardano/CLI.hs
Anviking: 👍

lib/cli/src/Cardano/CLI.hs
Anviking: 👍 Think I didn't have `KeyEncoding` in the same sense when I added this.

lib/cli/src/Cardano/CLI.hs
Anviking: The problem is that I want `valid_xprv1 <> "ö"` to fail with a `bech32` specific error message, but `0000` to fail with a hex-specific one.

Not sure how to achieve that otherwise.

I experimented some with:

```haskell
    let r1 = (,Hex) <$> decodeHex txt
    let r2 = (,Bech32) <$> decodeBech32 txt
    case (r1, r2) of
        (Right x, _) -> return x
        (_, Right x) -> return x
        (Left (Recognized e), Left (UnrecognizedEncoding _))
            -> Left e
        (Left (UnrecognizedEncoding _), Left (Recognized e))
            -> Left e
        (Left (UnrecognizedEncoding _), Left (UnrecognizedEncoding _))
            -> Left fullKeyEncodingDescription
        (Left (Recognized _), Left (Recognized _))
            -> Left "decodeAnyKey: internal error. The key is recognized as \
                    \bech32- and hex-encoded at the same time, but fails with \
                    \both."
```
but it didn't seem to work out fully.

lib/cli/src/Cardano/CLI.hs
Anviking: It occurred to me, but went with this because of:
>  it's nice to also have an explicit message indicating the user he's probably doing something he didn't intended

But I can't really imagine any realistic scenarios motivating either decision.

Me trying really hard to come up with a real argument for erroring:
```
# someone makes a CLI program which accepts both private and public keys
publishAddressOnline = mkAddress | cardano-wallet key public

# The authors forget that the program also accepts private keys

# The authors remove the public command
publishAddressOnline = mkAddress

# Users still think the program accepts private keys

# mkAddress is somehow flawed and leaks the users' private key
```

lib/cli/src/Cardano/CLI.hs
Anviking: I added golden tests for failures in 1d0cc5effcf02321e0ef0f3d3069d54328d5c726

lib/cli/src/Cardano/CLI.hs
Anviking: Maybe let's wait and see.

If we realise it will be useful for say, the address CLI commands, then let's add it then.

Anviking:
Thanks for the suggestions both!

> I think we should only have one way for passing inputs (stdin) and stick to it.

🤔 sure. Separate PR. Might need to re-organize the tests a bit for that.

bors r+

PR #1604, opened 2020-04-28
nix: remove socket files from source trees before building

rvl:
## Overview

Any `cardano-node.socket` files present in the source tree will cause an error in the nix build.

This happened in the nightly build. I do not know why the `git clean` command that Buildkite ran did not remove `cardano-node.socket` from the git checkout directory. But this change will prevent the error.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -59,9 +59,10 @@
 with pkgs; with commonLib; with pkgs.haskell-nix.haskellLib;
 
 let
-  src = pkgs.haskell-nix.cleanSourceHaskell {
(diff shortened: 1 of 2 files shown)


Anviking (approved):
🆗

jbgi (approved):

rvl:
There was a cached failure of cardano-wallet-byron:test:integration test case BYRON_WALLETS_RESYNC_01 on macOS. I restarted it and it passed this time.

bors r+

PR #1606, opened 2020-04-29
Don't deserialise blocks in the chain sync node client

rvl:
# Issue Number

Relates to #1577 / ADP-83.

# Overview

The chain tip following node client doesn't need to serialise blocks. So use bytestring representation for blocks in this part. 

# Comments

I attempted to change it to use `defaultCodecs` (serialised) / `clientCodecs` (deserialised) to help when upgrading the node version, but it is quite difficult to set up the arguments `mkByronConfig`. Not to mention there are two different `NodeToClientVersion` types that need to be imported. So I gave up on this.


KtorZ (approved):
Sound good.

PR #1607, opened 2020-04-29
[ADP-286] Enable tagging latest container

sevanspowell:
# Issue Number

ADP-286

# Overview

- https://jira.iohk.io/browse/ADP-286
- Adds a GitHub action that, on release, triggers a Buildkite action
  to tag the latest docker image with the "latest" tag.
- Note I've also added a "Buildkite_Token" GitHub secret so the action
  can talk to BuildKite.
- Note I've also added the pattern "latest" to the Branch Filter
  Pattern of this repo's Buildkite pipeline. This is so when Buildkite
  receives the tag "latest", it triggers a build.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,21 @@
+name: Latest Release
+
+on:
+  release:
+    types:
+      - created
+
+jobs:
+    test-job:
+      name: "Tag / Build / Push Container"
+      runs-on: ubuntu-latest
+      steps:
+      - name: Trigger Buildkite Pipeline
+        uses: buildkite/trigger-pipeline-action@v1.2.0
+        env:
+          BUILDKITE_API_ACCESS_TOKEN: ${{ secrets.Buildkite_Token }}
+          PIPELINE: ${{ github.repository }}
+          COMMIT: "HEAD"
+          BRANCH: "master"
+          MESSAGE: ":github: Triggered from a GitHub Action"
+          BUILD_ENV_VARS: "{\"BUILDKITE_TAG\": \"latest\"}"


.github/workflows/release.yaml
```diff
@@ -0,0 +1,21 @@
...
+      steps:
+      - name: Trigger Buildkite Pipeline
+        uses: buildkite/trigger-pipeline-action@v1.2.0
+        env:
+          BUILDKITE_API_ACCESS_TOKEN: ${{ secrets.Buildkite_Token }}
+          PIPELINE: ${{ github.repository }}
+          COMMIT: "HEAD"
+          BRANCH: "master"
+          MESSAGE: ":github: Triggered from a GitHub Action"
+          BUILD_ENV_VARS: "{\"BUILDKITE_TAG\": \"latest\"}"
```
rvl, line 21: This needs to be a git tag I think.
The buildkite docker image build script will take care of the rest ... it has the logic for applying the "latest" tag.

rvl (commented):
Cool, thanks!

Do you have a link to an example of a successful trigger + build?

.github/workflows/release.yaml
craigem: https://github.com/input-output-hk/latest_docker_dev

https://hub.docker.com/r/craigem/latest_docker_dev/tags

.github/workflows/release.yaml
```diff
@@ -0,0 +1,21 @@
+name: Latest Release
```
rhyslbw, line 1:
```suggestion
name: Release
```

rvl:
Thanks for this PR @sevanspowell @craigem. 
Because of the tag naming scheme in cardano-wallet, we can tweak the buildkite ref filter so that tag builds work. So no further action required.

Screenshot from 2020-04-11 07-53-46

Ideally, we could remove the buildkite branch filters, but they are required to prevent the `hydra-eval-errors` script from failing.

The `trigger-pipeline-action` looks handy though. I have used it in #1610.

PR #1610, opened 2020-04-29
Buildkite all-tests-pass branch

rvl:
### Overview

- Some buildkite fixes.
- Trigger buildkite nightly pipeline from github actions, after the windows tests pass.
- Push the `all-tests-pass` branch after the nightly job successfully completes.
- Add a github tag filter to the docker image push script. Otherwise, if the buildkite ref filter were removed, and someone pushed any tag, then the "latest" docker tags would be updated to that tag.
- Bump haskell.nix revision while we're at it.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -7,6 +7,7 @@ env:
 steps:
   - label: 'hydra-eval-errors'
     command: 'nix-build ./nix -A iohkNix.hydraEvalErrors && ./result/bin/hydra-eval-errors.py'
+    if: 'build.tag == null'
     agents:
       system: x86_64-linux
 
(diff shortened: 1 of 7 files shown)


.buildkite/pipeline.yml
```diff
@@ -7,6 +7,7 @@ env:
 steps:
   - label: 'hydra-eval-errors'
     command: 'nix-build ./nix -A iohkNix.hydraEvalErrors && ./result/bin/hydra-eval-errors.py'
+    if: 'build.tag != null'
```
KtorZ: I am not sure to get this one?

KtorZ (approved):

.buildkite/pipeline.yml
rvl: Oops, I meant the opposite. That script fails unconditionally for tag builds, so I want to disable it. Fixed.

PR #1612, opened 2020-04-30
Fix caching in first scenarion latency benchmarking

paweljakubas:
# Issue Number

Just make all network measurements comparable in all scenarios and runs. Always rely on cashing present.

# Overview

- [ ] I have ...

# Comments

### DIFF
--- a/lib/byron/bench/Latency.hs
+++ b/lib/byron/bench/Latency.hs
@@ -313,6 +313,11 @@ main = withUtf8Encoding $ withLatencyLogging $ \logging tvar ->
             (Link.getTransactionFee @'Byron wal1) Default payload
         fmtResult "postTransactionFee " t6
 
+        -- this one is to have comparable results from first to last measurement
+        -- otherwise the first one would be without cashing in contrast to other
+        -- measurements
+        _t7bare <- measureApiLogs tvar $ request @ApiNetworkInformation ctx
+            Link.getNetworkInfo Default Empty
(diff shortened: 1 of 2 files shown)


KtorZ (approved):

piotr-iohk (approved):

paweljakubas:
another occurrence of : 
```
  src/Test/Integration/Scenario/API/Byron/Wallets.hs:557:9:
  1) API Specifications, BYRON_WALLETS_RESYNC_01, emptyRandomWallet - force resync eventually get us back to the same point
       user error (waited more than 3min for action to eventually resolve. Action: "Wallet re-syncs successfully")

  To rerun use: --match "/API Specifications/BYRON_WALLETS_RESYNC_01/emptyRandomWallet - force resync eventually get us back to the same point/"

  src/Test/Integration/Scenario/API/Byron/Wallets.hs:560:9:
  2) API Specifications, BYRON_WALLETS_RESYNC_01, emptyIcarusWallet - force resync eventually get us back to the same point
       user error (waited more than 3min for action to eventually resolve. Action: "Wallet re-syncs successfully")

  To rerun use: --match "/API Specifications/BYRON_WALLETS_RESYNC_01/emptyIcarusWallet - force resync eventually get us back to the same point/"
```

rvl (approved):
Maybe "warming up" caches should be part of the `measureLatency` function, if you want to measure like that.

Which value is being cached? Is it the node tip? Because the way that's fetched has changed with #1580.

PR #1613, opened 2020-04-30
Remove force resync feature

piotr-iohk:
# Issue Number

N/A

# Overview

- 05ae60a7f6dd8fb7b87b7cc07e41a4bc77d22f64
  remove integration tests
  
- e650619499adebb16ae798665b9e0f57d199c14f
  remove forceResync feature
  
- a479fad727fa9f12a31d78415c65f01e44bd1eef
  remove forceResync from swagger spec

# Comments

_All tests passed for me locally._

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -42,7 +42,6 @@ module Cardano.Wallet.Api.Server
     , delegationFee
     , deleteTransaction
     , deleteWallet
-    , forceResyncWallet
     , getMigrationInfo
     , getNetworkClock
     , getNetworkInformation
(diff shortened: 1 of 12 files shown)


lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -979,44 +976,6 @@ getUTxOsStatistics ctx (ApiT wid) = do
...
 
-forceResyncWallet
-    :: forall ctx s t k.
-        ( ctx ~ ApiLayer s t k
-        , IsOurs s ChimericAccount
-        , IsOurs s Address
-        )
-    => ctx
-    -> ApiT WalletId
-    -> ApiNetworkTip
```
KtorZ, line 990: I wonder if this one (`ApiNetworkTip` is used for anything else or could also be removed.

KtorZ (approved):
:+1:

lib/core/src/Cardano/Wallet/Api/Server.hs
piotr-iohk: I think it is -> https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/Api/Types.hs#L571

KtorZ:
bors retry

Issue #1596, opened 2020-04-27
Mock metadata aggregation server


KtorZ:
# Context

The specification for the new metadata aggregation server is captured in the following swagger document:

https://gist.github.com/KtorZ/4ee1bf8136c91f2d4c36ed850023cab6

# Decision

In order to test the aggregation server, we'll need to setup a mock HTTP server that satisfies this specification. 

# Acceptance Criteria

---

# Development 

- [ ] Create a simple servant application that "mocks" the swagger specification ==> #1614 

# QA 

- The mock server is property tested checking that:
    a) It returns valid responses that satisfy the swagger schema mentioned above
    b) Returns both positive and negative responses (404), but does it deterministically (same pool id, same response always) so that we can use it to test correct behavior.
   
see property here: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Cardano/Pool/MetadataSpec.hs#L117-L137

The relevant bits of the schema were written directly using the Swagger DSL in Haskell: 

https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Cardano/Pool/MetadataSpec.hs#L262-L290

piotr-iohk:
LGTM.

PR #1614, opened 2020-04-30
Metadata aggregation client & mock server

KtorZ:
# Issue Number

#1596

# Overview

- 92a8d1c4355d0956c5087148d9b5e7a34d5685a9
  :round_pushpin: **change the signature of 'mkSeed' to return an Int instead of 'StdGen'**
  For better re-use across code

- 83b04d511d4711504da7cbe13c4e54dbdc71eb75
  :round_pushpin: **add simple servant client for querying off-chain metadata on an aggregation server**
  
- a280efae8ef384a540263dbf6644fe61d24473c5
  :round_pushpin: **write mock metadata aggregation server, with some smoke tests**
    An example of failure:

  ```
  Assertion failed (after 2 tests):
    PoolId {getPoolId = "\NUL\NUL\SOH\NUL\SOH\SOH\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\SOH\SOH\NUL\SOH\NUL\NUL\NUL\SOH\SOH\NUL\SOH\NUL\NUL\SOH\SOH\NUL\NUL\NUL"}
    {"homepage":"https://.io","name":"","ticker":"TWUMK","description":""}
    ["string is too short (length should be >=1)","string is too short (length should be >=1)"]

  Assertion failed (after 1 test):
    PoolId {getPoolId = "\SOH\NUL\NUL\SOH\SOH\SOH\SOH\NUL\SOH\NUL\SOH\SOH\SOH\SOH\SOH\NUL\SOH\SOH\NUL\NUL\SOH\SOH\SOH\SOH\NUL\SOH\NUL\SOH\NUL\SOH\SOH\NUL"}
    unexpected response from server: ConnectionError (HttpExceptionRequest Request {
      host                 = "Host \"127.0.0.1\""
      port                 = 39839
      secure               = False
      requestHeaders       = [("Accept","application/json;charset=utf-8,application/json")]
      path                 = "/api/v1/metadata/0100000101010100010001010101010001010000010101010001000100010100"
      queryString          = ""
      method               = "GET"
      proxy                = Nothing
      rawBody              = False
      redirectCount        = 10
      responseTimeout      = ResponseTimeoutDefault
      requestVersion       = HTTP/1.1
    }
    (ConnectionFailure Network.Socket.getAddrInfo (called with preferred socket type/protocol: AddrInfo {addrFlags = [AI_ADDRCONFIG], addrFamily = AF_UNSPEC, addrSocketType = Stream, addrProtocol = 0, addrAddress = <assumed to be undefined>, addrCanonName = <assumed to be undefined>}, host name: Just "Host \"127.0.0.1\"", service name: Just "39839"): does not exist (Name or service not known)))
  ```

  And, passes with:

  ```
  Mock Server works as intended
    +++ OK, passed 100 tests:
    54% No Corresponding Metadata
    46% Got Valid Metadata
  ```

- 8b10ed9c6de3b7f828716a9e453a0f4129a281b0
  :round_pushpin: **add top-level module comment giving some insights about the Metadata module**

# Comments


lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -0,0 +1,77 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
+
+module Cardano.Pool.Metadata
```
Anviking, line 15: A comment explaining some basics would be nice. Might be useful to answer *any* or multiple of:
- What is a Metadata Aggregation Server?
- Any urls where I can read more about them? (ok, I see https://jira.iohk.io/browse/CAD-736, but not public)
- Any haddock items I should check out in particular? (`StakePoolOffChainMetadata`? `Client`?)

Also, just curious: is it known whether the Voltaire-testnet will use these servers, or a repo like ITN?

Anviking: Maybe https://jira.iohk.io/browse/CAD-736 could be linked regardless with a TODO to replace with a public link if/once it exists?

lib/core/test/unit/Cardano/Pool/MetadataSpec.hs
```diff
@@ -0,0 +1,239 @@
...
+        \Client{getStakePoolMetadata} -> property $ \pid -> monadicIO $ do
+            run (getStakePoolMetadata pid) >>= \case
+                Left (FailureResponse _ res) -> do
+                    monitor $ label "No Corresponding Metadata"
+                    monitor $ counterexample $ mconcat
+                        [ show (responseStatusCode res)
+                        , " =/= "
+                        , show status404
+                        ]
+                    assert (responseStatusCode res == status404)
```
Anviking, line 112: Might be nicer to use something like [`assertWith`](https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Cardano/Wallet/DB/Properties.hs#L391-L396) to couple the assertion with the error message

Now the `counterexample` is dependent on the next line.

… though perhaps not worth for just one instance.

lib/core/src/Cardano/Pool/Metadata.hs
KtorZ: > A comment explaining some basics would be nice

Indeed! Will add.

> Any urls where I can read more about them?

Actually, we could link to: https://github.com/input-output-hk/smash/

> Is it known whether the Voltaire-testnet will use these servers, or a repo like ITN?

That is a good question. I am still unsure about the scope of the Voltaire testnet but the direction it seems to take is that it'll be an extension of the ITN and therefore, the old metadata registry on the CF repository still applies.

lib/core/test/unit/Cardano/Pool/MetadataSpec.hs
KtorZ: Yeah, I wish `assertWith` was part of QuickCheck though. I might as well do a PR actually :)

lib/core/test/unit/Cardano/Pool/MetadataSpec.hs
```diff
@@ -0,0 +1,239 @@
...
+                Left e -> do
+                    monitor $ counterexample $
+                        "unexpected response from server: " <> show e
+                    assert False
+                Right !metadata -> do
+                    monitor $ label "Got Valid Metadata"
+                    let json = toJSON $ ApiT metadata
+                    let errs = validateJSON mempty metadataSchema json
```
Anviking, line 120: 👌 

Would it be interesting to see how the client reacts when the response does _not_ conform to the swagger spec?

Anviking (approved):
lgtm

lib/core/test/unit/Cardano/Pool/MetadataSpec.hs
KtorZ: There's no logic in the client so far; I have another chunk of work adding the interesting logic of fetching / caching the metadata. Here it'll be more interesting to test faulty behavior. 

Also, the server not conforming to the swagger schema will likely result in the JSON parser failing, and therefore, a `Left ClientError` on the client. The JSON parser is quite lenient however, it doesn't check for things like minimal or maximal length. 

Having said that, the server not satisfying the interface is a "contract violation" and something that would require a fix on the server, if / when possible.

lib/core/test/unit/Cardano/Pool/MetadataSpec.hs
KtorZ: https://github.com/nick8325/quickcheck/pull/294 :man_shrugging:

lib/core/src/Cardano/Pool/Metadata.hs
KtorZ: 8b10ed9c6de3b7f828716a9e453a0f4129a281b0

KtorZ:
Hydra eval errors stuck for 40 minutes. :| ... Merging this.

PR #1615, opened 2020-04-30
ChainSync Pipelined: moar speed. 

KtorZ:
# Issue Number

ADP-270

# Overview

- 441eaf179d344d8c82e4f07204a0087017733038
  :round_pushpin: **replace the chain sync client with its pipelined version**
  No pipeline is done yet, this still uses the standard sequential way for sending requests,
but now using the 'ChainSyncClientPipelined' which has extra constructor we can now use
for pipelining

- 178b0dbe42f9b687f02e10c47b2fe00332af7632
  :round_pushpin: **implementing pipelining, pipeline by batches of 1000 requests. It's fast.**
  This is still work in progress for the 'edge' cases aren't properly handled. Rollbacks are just ignored
and, the area near the tip shouldn't be pipelined but is. Yet, this is a starter that gives some idea.

I haven't tweaked it much, nor tried different settings but so far, I've restored a wallet in ~6:30 minutes
on Mainnet, so more than 10k blocks per second.

- 37cdeb5bf210a72517494f86f0be9de5c64ad6d4
  :round_pushpin: **only pipeline when it's safe**
    This also makes sure to properly handle the starting and terminating
  conditions. This is a bit tricky because, ideally, to know whether we
  can safely pipeline, we want to check whether any of the pipelined
  request would be made within k blocks from the tip. The tricky part is
  "blocks". During rollbacks, we have access to points on the chain which
  are only referencing slots yet, we need to compare block height!

  To solve this, I've adopted two strategies: pipeline and one-by-one,
  defaulting to the latter and only using the former when we have enough
  information to make a sound decision. This means that the client always
  starts by restoring a single block from which it can now figure out a
  blockheight and compare this with the node's tip. In case it's far
  enough, we go full speed, in steps; checking at each step whether it is
  still safe to proceed with the next step.

  The last `k` blocks are therefore fetched one-by-one, which is quite
  slow but, it's `k` blocks out of millions. We _could_ possibly make this
  last part even faster by using the same batching strategy as before
  (where we would basically check after each request whether the node has
  rolled back or not) but it comes with some added complexity I am not
  necessarily fond of introducing. This new approach already cuts the
  overall restoration time by 2.5!

  e.g. on Mainnet:

  ```
  -- pipelined -----------------
  - [2020-05-06 12:12:34.40 UTC] 87a2d760: syncProgress: still restoring (0.00%)
  - [2020-05-06 12:15:12.80 UTC] 87a2d760: syncProgress: still restoring (50.01%)
  -- one by one ----------------
  - [2020-05-06 12:18:20.34 UTC] 87a2d760: syncProgress: still restoring (99.94%)
  - [2020-05-06 12:18:37.38 UTC] 87a2d760: syncProgress: restored
  ------------------------------
  6 minutes 3s / ~11366 block/s

  ```

- 485a324d48fa180bcd21721b29ce653bbe87d90a
  
- 62d47077ba36b78ba78a718240079558b943a98c
  :round_pushpin: **adjust log outputs for MsgNodeTip and MsgTxParameters**
    Going from:

  ```
  [cardano-wallet.network:Debug:16] [2020-05-07 09:44:29.79 UTC] Network node tip block height is 4129973 at hash 3cf1a60a
  [cardano-wallet.network:Info:17] [2020-05-07 09:31:03.66 UTC] TxParams for tip are:  Fee policy:         155381.0 + 43.0x + 0.0y
   Tx max size:        8192
  ``` 

  to:
  ```
  [cardano-wallet.network:Debug:17] [2020-05-07 09:40:30.88 UTC] Network node tip is fe158920-[191.6466#4129961]
  [cardano-wallet.network:Info:17] [2020-05-07 09:40:30.88 UTC] TxParams for tip are: [Fee policy: 155381.0 + 43.0x + 0.0y, Tx max size: 8192]
  ```

  a) avoid multi-line log output with akward structure for tx params
  b) Show tip in a consistent way with other loggers (for instance, in the wallet engine).

- feb1ecff96403b91a9a38fcc44c68a260b38f254
  :round_pushpin: **fix asynchronous exception handling**
    There are two sides to this commit:

  a) It fixes the asynchronous exception "catcher" in 'follow' that would
     fail to catch any exception that occurs while the thread is sleeping. I
     noticed that sometimes the logs wouldn't display a proper error.

  b) It also fixes the 'runQuery' combinator for our SQLite implementation
     to make sure that asynchronous exceptions are masked while executing
     sql queries. I falsy thought this was already properly handled by
     'runSqlConn' from persistent but it seems to be source of most of our
     'ErrorBusy' / 'statementAlreadyFinalized' issues.

  :finger_crossed:

# Comments

It's fast :man_shrugging:

### DIFF
--- a/lib/byron/src/Cardano/Wallet/Byron/Network.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron/Network.hs
@@ -132,6 +132,8 @@ import Network.Mux
     ( AppType (..), MuxError (..), MuxErrorType (..), WithMuxBearer )
 import Network.TypedProtocol.Codec
     ( Codec )
+import Network.TypedProtocol.Pipelined
+    ( N (..), Nat (..), natToInt )
 import Ouroboros.Consensus.Byron.Ledger
     ( ByronBlock (..)
     , GenTx
@@ -172,7 +174,7 @@ import Ouroboros.Network.Codec
 import Ouroboros.Network.CodecCBORTerm
     ( CodecCBORTerm )
 import Ouroboros.Network.Driver.Simple
-    ( TraceSendRecv, runPeer )
+    ( TraceSendRecv, runPeer, runPipelinedPeer )
 import Ouroboros.Network.Mux
     ( MuxPeer (..), OuroborosApplication (..), RunMiniProtocol (..) )
 import Ouroboros.Network.NodeToClient
@@ -196,6 +198,12 @@ import Ouroboros.Network.Protocol.ChainSync.Client
     , ClientStIntersect (..)
     , ClientStNext (..)
     , chainSyncClientPeer
+    )
+import Ouroboros.Network.Protocol.ChainSync.ClientPipelined
+    ( ChainSyncClientPipelined (..)
+    , ClientPipelinedStIdle (..)
+    , ClientPipelinedStIntersect (..)
+    , chainSyncClientPeerPipelined
     )
 import Ouroboros.Network.Protocol.ChainSync.Codec
     ( codecChainSync, codecChainSyncSerialised )
@@ -231,6 +239,7 @@ import qualified Codec.CBOR.Term as CBOR
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
 import qualified Ouroboros.Network.Protocol.LocalStateQuery.Client as LSQ
+import qualified Ouroboros.Network.Protocol.ChainSync.ClientPipelined as P
 
 {- HLINT ignore "Use readTVarIO" -}
 
@@ -442,7 +451,7 @@ type NetworkClient m = OuroborosApplication
 -- | Construct a network client with the given communication channel, for the
 -- purposes of syncing blocks to a single wallet.
 mkWalletClient
-    :: (MonadThrow m, MonadST m, MonadTimer m)
+    :: (MonadThrow m, MonadST m, MonadTimer m, MonadAsync m)
     => W.BlockchainParameters
         -- ^ Static blockchain parameters
     -> TQueue m (ChainSyncCmd m)
@@ -662,7 +671,7 @@ handleMuxError tr onResourceVanished = pure . errorType >=> \case
 chainSyncWithBlocks
     :: forall m protocol.
         ( protocol ~ ChainSync ByronBlock (Tip ByronBlock)
-        , MonadThrow m, MonadST m, MonadSTM m
+        , MonadThrow m, MonadST m, MonadSTM m, MonadAsync m
         )
     => Tracer m (TraceSendRecv protocol)
         -- ^ Base tracer for the mini-protocols
@@ -679,7 +688,7 @@ chainSyncWithBlocks
         -- socket).
     -> m Void
 chainSyncWithBlocks tr bp queue channel = do
-    runPeer tr codec channel (chainSyncClientPeer client)
+    runPipelinedPeer tr codec channel (chainSyncClientPeerPipelined client)
   where
     W.BlockchainParameters
         { getGenesisBlockHash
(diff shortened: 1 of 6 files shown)


lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -676,75 +698,114 @@ chainSyncWithBlocks
...
-    client :: ChainSyncClient ByronBlock (Tip ByronBlock) m Void
-    client = ChainSyncClient clientStIdle
+    -- Return the _number of slots between two tips.
+    tipDistance :: BlockNo -> Tip ByronBlock -> Natural
+    tipDistance (BlockNo n) TipGenesis =
+        1 + fromIntegral n
+    tipDistance (BlockNo n) (Tip _ _ (BlockNo n')) = fromIntegral $
+        if n' >= n
+        then n' - n
+        else  n  - n'
```
coot: Why not use `abs`?

lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -676,75 +698,114 @@ chainSyncWithBlocks
...
-            | otherwise = ClientStNext
-                { recvMsgRollBackward = onRollback
-                , recvMsgRollForward = \block _ ->
-                    ChainSyncClient $ pure $ SendMsgRequestNext
-                        (clientStNext (block:blocks,n-1) respond)
-                        (pure $ clientStNext (block:blocks,1) respond)
-                }
-          where
-            onRollback point _tip = ChainSyncClient $ do
+        -- Simple strategy that sends a request and wait for the answer.
```
coot: grammar: it should be `.. and waits for an answer`

lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -676,75 +698,114 @@ chainSyncWithBlocks
...
+            -> [ByronBlock]
+            -> Nat n
+            -> P.ClientStNext n ByronBlock (Tip ByronBlock) m Void
+        collectResponses respond blocks Zero = P.ClientStNext
+            { P.recvMsgRollForward = \block tip -> do
+                let cursor' = blockPoint block
+                let blocks' = reverse (block:blocks)
+                let tip'    = fromTip getGenesisBlockHash getEpochLength tip
+                respond (RollForward cursor' tip' blocks')
+                distance <- tipDistance (blockNo block) <$> getNodeTip
```
coot: Can't you use `tip`, instead of `getNodeTip`?

lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -676,75 +698,114 @@ chainSyncWithBlocks
...
+        --     |---------------------------|-------------->
+        --               pipelined            one by one
+        --
+        pipeline
+            :: Nat n
+            -> RequestNextStrategy m n
+        pipeline (Succ n) respond | natToInt (Succ n) == fromIntegral maxInFlight =
+            P.CollectResponse Nothing $ collectResponses respond [] n
+        pipeline n respond =
+            P.SendMsgRequestNextPipelined $ pipeline (Succ n) respond
```
coot, line 781: I would use `CollectResponse (Just (SendMsgRequestNextPipelined ..)) ...`. This pokes the receiving queue and if we got something it will be available in the next step, so more promptly.  You could experiment with some strategies when to use `CollectResponse (Just ...) ..` and when `SendMsgRequestNextPipelined`.  But maybe you're not doing so because of `Jörmungandr` support?

lib/byron/src/Cardano/Wallet/Byron/Network.hs
KtorZ: Because I didn't notice `BlockNo` had a `Num` instance :no_mouth:, good point.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
KtorZ: Actually yes. I went through several refactors and trials on this one, so this is a leftover :+1:

lib/byron/src/Cardano/Wallet/Byron/Network.hs
KtorZ: > But maybe you're not doing so because of Jörmungandr support?

Not really. We process blocks by batches through the wallet, to avoid having to trigger the database and create checkpoint after processing every block. The restoration loop is driving the networking side by asking for next blocks; it is in essence similar to how the ChainSync works; except that instead of request a single block, it requests a batch. 

So here, we simply leverage pipelining to collect the next batch more rapidly; it queues "maxInFlight" requests, and then collect them all. Before, we would simply call `RequestNext` several time, but waiting for each request to complete before calling the next one.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
KtorZ: Fixed https://github.com/input-output-hk/cardano-wallet/compare/c1ef56bb5e3b9c924673981ee6d4ebb36248e3d4..0c1a920da3563ca8cf09e37d620b78b653f17555

lib/byron/src/Cardano/Wallet/Byron/Network.hs
KtorZ: Fixed https://github.com/input-output-hk/cardano-wallet/compare/c1ef56bb5e3b9c924673981ee6d4ebb36248e3d4..0c1a920da3563ca8cf09e37d620b78b653f17555

KtorZ:
bors try

lib/byron/src/Cardano/Wallet/Byron/Network.hs
coot: I see; then you could pipline even more, collect and buffer things as they come from the network (on your side) and send batches for processing. If you do your own buffering you will be able to saturate the network.

KtorZ:
>  then you could pipline even more, collect and buffer things as they come from the network (on your side) and send batches for processing

Yes, I thought of that but it requires a quite fundamental change in the way our networking layer currently operate. In theory, we could actually have the network layer do its own buffering indeed and, while the wallet engine is busy processing blocks and creating checkpoints, the networking layer could already be pipelining the next queries! 

Yet, the networking layer is currently driven by the engine on a pull-based mechanism, so it only request the next blocks when asked and this is a synchronous chain of events; I'll try make this a bit more concurrent to see if the complexity is worth adding. Getting below 5 minutes restoration time would be pretty nice :D

Anviking (approved):
lgtm, and tried restoring with Daedalus Mainnet

KtorZ:
```
    [cardano-wallet.wallet-engine:Error:68351] [2020-05-07 12:23:58.46 UTC] b502fca0: Unexpected error following the chain: user error (restoreBlocks: given chain isn't a valid continuation. Wallet is at: fd77ead9-[0.0#0] but the given chain continues starting from: dc23b4f4-[0.1#2])
```

Interesting.

KtorZ:
Hydra timing out again.

PR #1617, opened 2020-05-01
[ADP-286] Enable tagging latest container

sevanspowell:
# Issue Number

[ADP-286](https://jira.iohk.io/browse/ADP-286)

# Overview

- Adds a GitHub action that triggers the "docker-build-push.nix"
  script on release.
- Modified the "docker-build-push.nix" script:
    On release:
      - upload byron image as "latest"
      - upload images: "<version>-byron" and "<version>-jormungandr"
      - upload images: "byron" and "jormungandr"
    On master branch:
      - upload "dev-master-byron" and "dev-master-jormungandr"
    On every commit:
      - upload "<COMMIT_ID>-byron" and "<COMMIT_ID>-jormungandr"
- The difference here is that we're using a release workflow
  consistent with the other repo's addressed under [ADP-286].
- Additionally, we're uploading a image with the commit ID on every
  build.
- Note I've also added a "Buildkite_Token" GitHub secret so the action
  can talk to BuildKite.

# Comments

- The idea behind this was to bring this repo's
  "docker-build-push.nix" further into aligment with the rest of the
  repo's under [ADP-286], but I'm not particularly sure I've achieved
  that. This repo has quite different image requirements, namely:
    - "dev-master" images
    - "date-backend" images

### DIFF
--- a/.buildkite/docker-build-push.nix
+++ b/.buildkite/docker-build-push.nix
@@ -70,36 +70,45 @@ in
     ''}
 
   '' + concatMapStringsSep "\n" (image: ''
+    gitrev=${pkgs.iohkNix.commitIdFromGitRepoOrZero ../.git}
     branch="''${BUILDKITE_BRANCH:-}"
-    tag="''${BUILDKITE_TAG:-}"
-    extra_tag=""
-    if [[ -n "$tag" ]]; then
-      tag="${image.imageTag}"
-      extra_tag="${image.backend}"
-    elif [[ "$branch" = master ]]; then
-      tag="$(echo ${image.imageTag} | sed -e s/${image.version}/''${BUILDKITE_COMMIT:-dev-$branch}/)"
-      extra_tag="$(echo ${image.imageTag} | sed -e s/${image.version}/dev-$branch/)"
-    else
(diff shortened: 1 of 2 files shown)


KtorZ (commented):
I am not sure I am willing to approve this. This setup is more complicated than the current one which already does give us the tagging we want. 

Although I understand the desire for aligning the setup in different repositories, it's still different files in different repositories with different needs. There's no way these are going to be maintained in sync -- or perhaps at a relatively high and painful cost. 

The end goal of ADP-286 is to make sure various repositories which expose services can be pulled with the default docker tags. cardano-wallet already satisfies this criteria :+1:

craigem:
Hi @KtorZ 

The brief for AD-286 as it was given to us and as implemented elsewhere, was that all containers tagged `latest` must be triggered from the GitHub release process.

The existing workflow does not do this, hence this PR which does.

We have no ownership over this workflow, we're just delivering what was asked of us. If it's not actually wanted, then this PR brings no value :smile:

rvl:
Thanks, we continue to trigger builds with `v20*` tag pushes for now because they work well enough.

PR #1618, opened 2020-05-01
Unit test for LocalStateQuery

rvl:
### Issue Number

Relates to #1577 / ADP-83.

### Overview

Adds a test of NetworkLayer.getTxParameters against the self-node


lib/byron/test/unit/Cardano/Wallet/Byron/NetworkSpec.hs
```diff
@@ -0,0 +1,67 @@
...
+spec = describe "getTxParameters" $ do
+    it "Correct values are queried" $ do
+        withTestNode $ \gbp sock vData -> withLogging $ \(tr, getLogs) -> do
+            let gbp' = gbp { txParameters = fakeTxParameters }
+            withNetworkLayer tr gbp' sock vData $ \nl -> do
+                let retryPolicy = constantDelay 1_000_000 <> limitRetries 10
+                recoverAll retryPolicy $ const $
+                    getTxParameters nl `shouldReturn` txParameters gbp
+            msg <- mapMaybe isMsgTxParams <$> getLogs
+            msg `shouldBe` [txParameters gbp]
```
KtorZ, line 54: Feels more like an integration test than a unit test though.

KtorZ (approved):

lib/byron/test/unit/Cardano/Wallet/Byron/NetworkSpec.hs
rvl: The byron network layer is not set up very well for unit testing, so using the real cardano-node was the best that I could do.

rvl:
Socket file issue on macos - fixed.

bors r+

Issue #1603, opened 2020-04-28
CLI: Support bech32 encoding


Anviking:
## Context

ADP-195

## Decision

Add support for bech32 key encoding.

## Acceptance criterea

- `key inspect`, `key child` and `key public` must accept both bech32- and hex-encoded keys.
- `key root` should still default to outputting a hex-encoded key
- `key root` should output bech32 when given `--encoding bech32`
- Chaining commands with pipes (`|`) should be possible.

## Development 

- [x] #1599 Add bech32 support
- [x] #1619 Render pretty bech32 error messages
- [x] #1622 Breaking: Only allow reading keys from stdin
- [x] #1643 Make rendering of bech32 errors more correct

## QA 

- Integration tests making sure commands can be chained with both hex and bech32 encoding https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core-integration/src/Test/Integration/Scenario/CLI/Keys.hs#L17-L51
- Golden tests for key decoding error messages https://github.com/input-output-hk/cardano-wallet/blob/master/lib/cli/test/unit/Cardano/CLISpec.hs#L607-L664
- Not sure if the ANSI red works on windows https://github.com/input-output-hk/cardano-wallet/pull/1619#issuecomment-624088900, but its just for an error message, and think there are more important things to worry about.
- Added two properties about the rendering of ANSI red in #1643 (but don't think it says anything about windows)

piotr-iohk:
LGTM.

PR #1619, opened 2020-05-04
Better bech32 error messages

Anviking:
# Issue Number

#1603 

# Overview

- [x] Addded custom error messages for bech32 errors
- [x] Mark invalid characters red <s>and underline them with `^`</s>

# Comments

- Not sure if/how we should test
- Not sure if the (non-IO) ANSI red works on windows
- <s>Not sure if the `^` underlining is a great idea considering line wrapping is likely</s> removed

Skärmavbild 2020-05-04 kl  12 02 28

Skärmavbild 2020-05-04 kl  12 02 49

### DIFF
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -188,6 +188,8 @@ import Cardano.Wallet.Primitive.Types
     ( AddressState, Hash, SortOrder, SyncTolerance (..), WalletId, WalletName )
 import Cardano.Wallet.Version
     ( gitRevision, showFullVersion, version )
+import Codec.Binary.Bech32
+    ( CharPosition (..), DecodingError (..) )
 import Control.Applicative
     ( optional, some, (<|>) )
 import Control.Arrow
@@ -295,6 +297,7 @@ import System.Console.ANSI
     , hCursorBackward
     , hSetSGR
     , hSupportsANSIWithoutEmulation
+    , setSGRCode
     )
 import System.Directory
     ( XdgDirectory (..)
@@ -492,7 +495,7 @@ decodeKey = \case
   where
     decodeBech32 :: Text -> Either String XPrvOrXPub
     decodeBech32 t = do
-        (hrp, dp) <- left show $ Bech32.decodeLenient t
+        (hrp, dp) <- left bech32Err $ Bech32.decodeLenient t
         bytes <- maybe (Left dpErr) Right $ Bech32.dataPartToBytes dp
         case hrp of
             h | h == xpubHrp -> xpubFromBytes bytes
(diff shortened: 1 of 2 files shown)


KtorZ:
Alternatively, to avoid the issue with the line wrapping, it is perhaps possible to only show part of the encoded string? Like the 5 characters that precede the error, and the 5 after and at the exact position in the error message:

```
Bech32 error: Invalid character(s) in string at position 78/127:
[...]90aa66mvaux[...]
          ^
```

:thinking: ?

lib/cli/src/Cardano/CLI.hs
```diff
@@ -500,6 +503,48 @@ decodeKey = \case
...
+arrowsAtIndices :: Integral i => [i] -> String
+arrowsAtIndices ixs = go 0 ixs ++ def
+  where
+    go _c [] = mempty
+    go c (x:xs)
+        | c == x    = red ++ "^" ++ go (c + 1) xs
+        | otherwise = ' ' : go (c + 1) (x:xs)
+
+    red = setSGRCode [SetColor Foreground Vivid Red]
+    def = setSGRCode [Reset]
```
KtorZ: See: https://input-output-hk.github.io/cardano-wallet/haddock/edge/cardano-wallet-cli-2020.4.28/src/Cardano.CLI.html#withSGR which also properly handles non-terminal environments :)

lib/cli/src/Cardano/CLI.hs
```diff
@@ -500,6 +503,48 @@ decodeKey = \case
       where
         dpErr = "Internal error: Unable to extract bytes from bech32 data part"
 
+        bech32Err  = ("Bech32 error: " <>) . \case
+            StringToDecodeTooLong -> "string is too long"
+            StringToDecodeTooShort -> "string is too short"
+            StringToDecodeHasMixedCase -> "mixed case"
```
KtorZ: for consistency: "string has mixed case" would sound better.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -500,6 +503,48 @@ decodeKey = \case
...
+            StringToDecodeMissingSeparatorChar -> "missing separator char"
```
KtorZ: for consistency: "string has no separator char" would sound better.

lib/cli/src/Cardano/CLI.hs
Anviking: I'm not in IO here though 🤔

KtorZ: Well, yeah, the function needs IO to determines whether the code runs in a terminal or not.

Anviking:
Skärmavbild 2020-05-04 kl  16 44 06

Tbh, it gets a bit non-trivial implementation-wise. Need to cluster neighbouring indices, pad,  etc. 

Might consider just printing the full key with coloured characters (and not `^`).

Anviking:
I think we should go with the simple option of printing the full key and marking the invalid characters red.

Unless that's somehow a security concern.

Anviking:
bors try

rvl:
I would be fine with just printing the faulty address and character offset at which the error occurred.
This is enough information to identify the error.

KtorZ (approved):

Anviking:
> This variant outputs the sequence of control characters as a String, which can be added to any other bit of text before being output. The use of these codes is generally discouraged because they will not work on legacy versions of Windows where the terminal in use is not ANSI-enabled (see further above). On Windows, where emulation has been necessary, these variants will always output the empty string. That is done so that it is possible to use them portably; for example, coloring console output on the understanding that you will see colors only if you are running on a Unix-like operating system or a version of Windows where emulation has not been necessary. If the control characters are always required, see module System.Console.ANSI.Codes.

[src](https://hackage.haskell.org/package/ansi-terminal-0.10.3/docs/System-Console-ANSI.html)

this is why I'm not sure to what extent it will work on windows, but should be a minor problem even if there is one

bors r+

Anviking:
🤦‍♂️

Issue #1597, opened 2020-04-27
Module for fetching and caching metadata from the aggregation server


KtorZ:
# Context

See:

- https://jira.iohk.io/browse/ADP-258
- #1595
- #1596 

# Decision

Write a self-isolated module that can offer a simple interface to fetch and refresh metadata from an aggregation server. We'll likely cache these data in a database or on-disk in a second step.

# Acceptance Criteria

---

# Development 

- Cached client fetching data from an aggregation server ==> #1623 
- ~~Database interface for storing and fetching cached metadata~~ ==> Will actually do when we have more confidence about the final format of the metadata. This should be pretty straightforward but doing it now could force us to have to work on some unfortunate db migration later :| 

# QA 

- The cached client is tested in various scenarios with existing and non existing metadata [here](https://github.com/input-output-hk/cardano-wallet/blob/a218df265f67d2e8c941deb383129c521197a114/lib/core/test/unit/Cardano/Pool/MetadataSpec.hs#L139-L169).

Anviking:
Btw, did you know about [`managerRetryableException`](https://hackage.haskell.org/package/http-client-0.6.1/docs/src/Network.HTTP.Client.Types.html#managerRetryableException)? Maybe it could have been used for cardano-wallet-jormungandr, or can be used here, if we want retrying.

KtorZ:
Nope I didn't! Sounds like indeed something that is pretty handy. I've seen the servant client automatically retry on 500 errors. I wonder if it's manual or using this setting from the manager.

piotr-iohk:
LGTM.

PR #1623, opened 2020-05-04
Mock metadata aggregation server

KtorZ:
# Issue Number

#1597

# Overview

-  64052c5449d21ef3002d6ddf21dc06669e4c9315
  :round_pushpin: **Basic skeleton for refreshing metadata.**
  Still requires proper log messages and error handling. But this already allows some level of testing.

- 7be9a83dd78c02cef31b1ba71d08c5838709e98c
  :round_pushpin: **log events while refreshing metadata**
  
- 566d95d9482dd32eb4098288ecca99ac994bb36f
  :round_pushpin: **handle 404 in the metadata client directly**
  So that the calling code needs not to worry about the 'ClientError'. Any error becomes unexpected and all errors can be treated consistently.

- 398bb7839e105fabeea062292d6cbfe40e01ed8b
  :round_pushpin: **merge 'refresh' and 'Client' so that the client is already caching results by itself.**
  This makes it slightly easier to use from a consumer perspective. Once one has a handle on a 'Client IO Api' it can makes requests without having to
worry about whether they get cached or not. Doing so, I also removed the 'ClientError' from the response. The client does handle error by itself and
log messages accordingly. A caller would likely not do anything more than relogging the error.
Plus, from the caller perspective, a metadata not being there and the request failing has pretty much the same end: no metadata.

- 98a499d1e5fdbf20361074a6fe30be7766747852
  :round_pushpin: **write small unit / property tests to exercise the client.**
    Written as property so we get some fuzzy testing and test cases with and without metadata
  Fails nicely with each log messages obtained in the run:
  ```
  Assertion failed (after 1 test):
    PoolId {getPoolId = "\SOH\NUL\SOH\NUL\NUL\NUL\NUL\NUL\NUL\SOH\SOH\NUL\SOH\SOH\SOH\SOH\NUL\SOH\NUL\SOH\SOH\SOH\NUL\SOH\NUL\SOH\NUL\NUL\NUL\SOH\SOH\NUL"}
    MsgRefreshingMetadata (PoolId {getPoolId = "\SOH\NUL\SOH\NUL\NUL\NUL\NUL\NUL\NUL\SOH\SOH\NUL\SOH\SOH\SOH\SOH\NUL\SOH\NUL\SOH\SOH\SOH\NUL\SOH\NUL\SOH\NUL\NUL\NUL\SOH\SOH\NUL"}) (Just (StakePoolOffChainMetadata {ticker = StakePoolTicker {unStakePoolTicker = "VUHQ"}, name = "_-NfdG\21785\&0?-wjm'\\.G{0MtJ4E", description = "_-NfdG\21785\&0?-wjm'\\.G{0MtJ4E", homepage = "https://\tXuP9h>v\ETXXM\FSUHd\ACKp\SOH\FS\SOHg\DEL$.io"}),2020-05-04 23:31:26.917183969 UTC)
    MsgUsingCached (PoolId {getPoolId = "\SOH\NUL\SOH\NUL\NUL\NUL\NUL\NUL\NUL\SOH\SOH\NUL\SOH\SOH\SOH\SOH\NUL\SOH\NUL\SOH\SOH\SOH\NUL\SOH\NUL\SOH\NUL\NUL\NUL\SOH\SOH\NUL"}) 2020-05-04 23:31:26.917183969 UTC
  ```

- 51cf00ad9c6391dcc1a600298150c35909b1cd3f
  :round_pushpin: **add an extra scenario illustrating a unhappy path**
    ```
  Cardano.Pool.Metadata
    Metadata - MockServer
      Mock Server works as intended
        +++ OK, passed 100 tests:
        51% Got Valid Metadata
        49% No Corresponding Metadata
      Cache metadata when called twice within the TTL
        +++ OK, passed 10 tests.
      Fetch them again when fetching outside of the TTL
        +++ OK, passed 10 tests.
      Returns 'Nothing' and a warning log message on failure
        +++ OK, passed 1 test.
  ```

- c76f29d7b3eb338e26889d9ec41f8ecb37cd08d3
  :round_pushpin: **move 'count' to 'Test.Util.Trace' with extra comments'**

### DIFF
--- a/lib/core/src/Cardano/Pool/Metadata.hs
+++ b/lib/core/src/Cardano/Pool/Metadata.hs
@@ -21,7 +21,10 @@ module Cardano.Pool.Metadata
     ( Api
     , Client(..)
     , mkClientIO
+
+    -- * Refreshing Metadata
     , refresh
+    , MetadataRegistryLog (..)
 
     -- * Re-export
     , BaseUrl (..)
@@ -33,16 +36,26 @@ module Cardano.Pool.Metadata
 
 import Prelude
 
+import Cardano.BM.Data.Severity
+    ( Severity (..) )
+import Cardano.BM.Data.Tracer
+    ( HasPrivacyAnnotation (..), HasSeverityAnnotation (..) )
 import Cardano.Wallet.Api.Types
     ( ApiT (..) )
 import Cardano.Wallet.Primitive.Types
     ( PoolId, StakePoolOffChainMetadata (..) )
 import Control.Monad
     ( forM_ )
+import Control.Tracer
+    ( Tracer, traceWith )
 import Data.Proxy
     ( Proxy (..) )
+import Data.Text.Class
+    ( ToText (..) )
 import Data.Time.Clock
     ( NominalDiffTime, UTCTime, diffUTCTime, getCurrentTime )
+import Fmt
+    ( pretty )
 import Network.HTTP.Client
     ( Manager, defaultManagerSettings, newManager )
 import Servant
@@ -56,6 +69,8 @@ import Servant.Client
     , mkClientEnv
     , runClientM
     )
+
+import qualified Data.Text as T
 
 --
 -- Api
@@ -128,21 +143,66 @@ data MetadataConfig (m :: * -> *) = MetadataConfig
 --
 -- TODO: Add some typed log messages in here.
 refresh
-    :: MetadataConfig IO
+    :: Tracer IO MetadataRegistryLog
+        -- ^ Logging object for capturing events in the 'refresh' function.
+        --
+    -> MetadataConfig IO
         -- ^ Configuration and callbacks for persistence of metadata.
 
     -> [PoolId]
         -- ^ A list of pool ids for which metadata need to be fetched or
         -- refreshed.
     -> IO ()
-refresh cfg pids = do
+refresh tr cfg pids = do
     now <- getCurrentTime
     forM_ pids $ \pid -> getModificationTime pid >>= \case
         Just timestamp | diffUTCTime now timestamp < cacheTTL -> do
-            pure ()
-        _expiredOrNotCached -> do
-            meta <- either onClientError (pure . Just) =<< getStakePoolMetadata pid
-            saveMetadata (meta, now)
+            traceWith tr $ MsgUsingCached pid timestamp
+        _expiredOrNotCached -> getStakePoolMetadata pid >>= \case
+            -- FIXME: Handle errors properly:
+            --
+            --   - 404 not found are somewhat expected
+            --   - Other errors should be reported as warning.
+            Left e ->
+                traceWith tr $ MsgUnexpectedError e
+            Right meta -> do
+                traceWith tr $ MsgRefreshingMetadata pid (Just meta, now)
+                saveMetadata pid (Just meta, now)
   where
     MetadataConfig{saveMetadata,getModificationTime,apiClient,cacheTTL} = cfg
     Client{getStakePoolMetadata} = apiClient
+
+-- | Capture log events for the 'refresh' function.
+data MetadataRegistryLog
+    = MsgUsingCached PoolId UTCTime
+    | MsgRefreshingMetadata PoolId (Maybe StakePoolOffChainMetadata, UTCTime)
+    | MsgUnexpectedError ClientError
+    deriving (Show, Eq)
+
+instance HasPrivacyAnnotation MetadataRegistryLog
+instance HasSeverityAnnotation MetadataRegistryLog where
+    getSeverityAnnotation = \case
+        MsgUsingCached{} -> Debug
+        MsgRefreshingMetadata{} -> Debug
+        MsgUnexpectedError{} -> Warning
+
+instance ToText MetadataRegistryLog where
+    toText = \case
+        MsgUsingCached pid time -> T.unwords
+            [ "Using cached result for"
+            , pretty pid
+            , "last modified at"
+            , T.pack (show time)
+            ]
+        MsgRefreshingMetadata pid (meta, time) -> T.unwords
+            [ "Setting metadata for "
+            , pretty pid
+            , "="
+            , maybe "ø" (T.pack . show) meta
+            , ", last modified at"
+            , T.pack (show time)
+            ]
+        MsgUnexpectedError e -> T.unwords
+            [ "Unexpected error from the aggregation server:"
+            , T.pack (show e)
+            ]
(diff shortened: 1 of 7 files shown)


lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -65,23 +95,123 @@ type Api
...
         client (Proxy @Api)
+
+-- | A configuration for managing metadata with the aggregation server.
+-- Callbacks and parameterized effects allows for easier testing while a real
+-- specialization would wire a database connector in here.
+data MetadataConfig (m :: * -> *) = MetadataConfig
+    { saveMetadata
+        :: PoolId -> (Maybe StakePoolOffChainMetadata, UTCTime) -> m ()
+        -- ^ A callback action for storing an off-chain metadata. The callback
+        -- may be called with 'Nothing' to store that no metadata were found for
```
Anviking, line 128: > may be called with 'Nothing' to store that no metadata were found

Does this mean any existing metadata will be removed?

> this allows for not constantly re-fetching data for pools that are known to have no metadata

Not quite understanding this from the comment

lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -65,23 +95,123 @@ type Api
...
+    :: Tracer IO MetadataRegistryLog
+        -- ^ Logging object for capturing events in the 'refresh' function.
+        --
+    -> MetadataConfig IO
+        -- ^ Configuration and callbacks for persistence of metadata.
+
+    -> [PoolId]
+        -- ^ A list of pool ids for which metadata need to be fetched or
+        -- refreshed.
+    -> IO ()
```
Anviking: Why is `refresh` needed? Can we not just put the caching inside `Client`?
```haskell
cachedClient 
    :: (PoolId -> m (Maybe StakePoolOffChainMetadata, UTCTime)) 
    -> (PoolId -> Maybe (StakePoolOffChainMetadata, UTCTime) -> m ()) 
    -> Client Api m 
    -> Client Api m
cachedClient pid putCached getCached rawClient = Client { getStakePoolMetadata
    if stale 
    then (getStakePoolMetadata rawClient) >>= \x -> putCache x >> return x 
    else getCached
}
```

lib/core/src/Cardano/Pool/Metadata.hs
KtorZ: > Does this mean any existing metadata will be removed?

No. This would be done automatically by the DB when pushing a de-registration certificate.

> Not quite understanding this from the comment

The thing is: we don't want to only cache metadata when they exist, but also when they aren't available. If we only cache metadata hit, then everytime we'll "refresh" metadata we'll have to make a query for all the pools that didn't have metadata. 

Yet, if we cache both the presence or absence of metadata, we only need a request for pools that are new and for which the state is unknown.

lib/core/src/Cardano/Pool/Metadata.hs
KtorZ: That is a valid point. Felt simpler to write both decoupled initially but, simply having a cached client make sense.

lib/core/test/unit/Cardano/Pool/MetadataSpec.hs
```diff
@@ -237,3 +321,14 @@ generateWith seed (MkGen gen) =
     --
     -- https://hackage.haskell.org/package/QuickCheck-2.13.2/docs/src/Test.QuickCheck.Gen.html#generate
     size = 30
+
+-- | Count elements in the list matching the given Prism.
+--
+-- >>> count (_Ctor @"MyConstructor") xs
+-- 2
+count :: Prism' s a -> [s] -> Int
```
rvl: Maybe put this in a tests common module, because other tests which count log messages can use it.

rvl (approved):

lib/core/test/unit/Cardano/Pool/MetadataSpec.hs
KtorZ: c76f29d7b3eb338e26889d9ec41f8ecb37cd08d3

KtorZ:
bors try

lib/core/src/Cardano/Pool/Metadata.hs
Anviking: I was more curious about a `Client Api m -> Client Api m` function, but sure, what you have now works too.

lib/core/test/unit/Cardano/Pool/MetadataSpec.hs
```diff
@@ -83,54 +93,120 @@ import Test.QuickCheck
...
 
+    around (withMockServer inMemoryCache)
+        $ it "Cache metadata when called twice within the TTL"
+        $ \mkClient -> withMaxSuccess 10 $ property $ \pid -> monadicIO $ do
+            (logs, _) <- run $ captureLogging $ \tr -> do
+                let Client{getStakePoolMetadata} = mkClient tr
+                void $ getStakePoolMetadata pid
+                void $ getStakePoolMetadata pid
+            monitor $ counterexample $ unlines $ show <$> logs
+            assert $ countMsg (_Ctor @"MsgRefreshingMetadata") logs == 1
```
Anviking, line 147: Cool approach!

lib/core/test/unit/Cardano/Pool/MetadataSpec.hs
```diff
@@ -83,54 +93,120 @@ import Test.QuickCheck
...
+    around (withMockServer inMemoryCache)
+        $ it "Fetch them again when fetching outside of the TTL"
+        $ \mkClient -> withMaxSuccess 10 $ property $ \pid -> monadicIO $ do
+            (logs, _) <- run $ captureLogging $ \tr -> do
+                let Client{getStakePoolMetadata} = mkClient tr
+                void $ getStakePoolMetadata pid
+                threadDelay' (2 * defaultCacheTTL)
+                void $ getStakePoolMetadata pid
+            assert $ countMsg (_Ctor @"MsgRefreshingMetadata") logs == 2
+            assert $ countMsg (_Ctor @"MsgUsingCached")        logs == 0
```
Anviking, line 159: But now, what if the data becomes unavailable? How should it react?

Anviking (approved):

lib/core/test/unit/Cardano/Pool/MetadataSpec.hs
KtorZ: If the metadata is removed, then it'll be updated when the cache is refreshed.

Issue #1621, opened 2020-05-04
Database migration tests is failing


piotr-iohk:
# Context
 
Seems to be failing:

```
*** Migrating from v2019-12-13 ***
*** Migrating from v2019-12-16 ***
```

# Test Case

https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/502#3cee1ff2-e03a-4e1c-abca-3fac2ff3b1d6

# Failure / Counter-example

https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/502#3cee1ff2-e03a-4e1c-abca-3fac2ff3b1d6

---

# Resolution 

---

# QA

piotr-iohk:
BTW -> https://github.com/input-output-hk/cardano-wallet/blob/master/nix/migration-tests.nix#L44-L61

It seems that we are testing against fixed versions, which are quite old now. Not sure what would make sense to support in migration, 3 versions back, 5 versions back? Or maybe we need to support migration from any version that was released?

rvl:
Oh I see. It wants to drop and recreate the checkpoint table when non-nullable columns are removed. That will be an issue for upgrading Daedalus.

Definitely, we should only test migrating from the versions that we care about. And we should be testing cardano-wallet-byron.

KtorZ:
Seems like there's still an issue in the migration tests: https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/505

Unless this nightly was triggered before #1624 was merged somehow.

rvl:
Yes that build was triggered before #1624 got merged.
I kicked off another test where it has passed:
https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/506

piotr-iohk:
Migration tests passed in https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/506

PR #1624, opened 2020-05-05
Fix migration error on checkpoint table

rvl:
### Issue Number

Relates to #1577 / ADP-83 and  #1621.

### Overview

Fix migration error on checkpoint table found in nightly tests.

SQLite doesn't support removing columns, so we put them back as "unused" fields (https://www.sqlite.org/lang_altertable.html#altertableishard).

### DIFF
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
@@ -133,8 +133,10 @@ Checkpoint
     checkpointBlockHeight       Word32       sql=block_height
     checkpointGenesisHash       BlockId      sql=genesis_hash
     checkpointGenesisStart      UTCTime      sql=genesis_start
+    checkpointFeePolicyUnused   Text         sql=fee_policy
     checkpointSlotLength        Word64       sql=slot_length
     checkpointEpochLength       Word32       sql=epoch_length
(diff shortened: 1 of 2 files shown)


lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
```diff
@@ -133,8 +133,10 @@ Checkpoint
     checkpointBlockHeight       Word32       sql=block_height
     checkpointGenesisHash       BlockId      sql=genesis_hash
     checkpointGenesisStart      UTCTime      sql=genesis_start
+    checkpointFeePolicyUnused   Text         sql=fee_policy
     checkpointSlotLength        Word64       sql=slot_length
     checkpointEpochLength       Word32       sql=epoch_length
+    checkpointTxMaxSizeUnused   Word16       sql=tx_max_size
     checkpointEpochStability    Word32       sql=epoch_stability
     checkpointActiveSlotCoeff   Double       sql=active_slot_coeff
```
KtorZ, line 141: What about grouping the unused fields and separate them from the rest, so it visually stands out.

KtorZ (approved):
Arf.

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
rvl: I would like to do this but alas the reordering of columns causes a destructive auto-migration.

PR #1626, opened 2020-05-05
Initial integration with shelley node

Anviking:
# Issue Number

ADP-289, #1657

# Overview

- [x] Copied the `cardano-wallet-byron` structure and made it compile for shelley
- [x] Completely unimplemented `TransactionLayer`
- [x] A few integration tests and unit tests are running (that don't need txs)
- [x] Add appropriate CLI and API
- [x] Make sure a wallet can be restored

# Comments


KtorZ:
@Anviking can you reach out to the networking & consensus team on #ouroboros-network-dev and find out about _"Seems local state query is not supported yet since there is no QueryLedger (ShelleyBlock c) instance in ouroboros-network."_ ?

lib/shelley/exe/cardano-wallet-shelley.hs
```diff
@@ -0,0 +1,309 @@
...
+main :: IO ()
+main = withUtf8Encoding $ do
+    enableWindowsANSI
+    runCli $ cli $ mempty
+        <> cmdServe
+        <> cmdMnemonicByron
+        <> cmdKey
+        <> cmdWallet cmdByronWalletCreate byronWalletClient
+        <> cmdAddress byronAddressClient
+        <> cmdTransaction byronTransactionClient byronWalletClient
```
KtorZ: These will have to be change to point to the "normal" API. We could perhaps still have them behind a `legacy-wallet` and `legacy-address` sets of command.

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
```diff
@@ -0,0 +1,256 @@
...
+
+    wallets :: Server Wallets
+    wallets =
+             (\_ -> throwError err501)
+        :<|> (\_ -> throwError err501)
+        :<|> throwError err501
+        :<|> (\_ -> throwError err501)
+        :<|> (\_ _ -> throwError err501)
+        :<|> (\_ _ -> throwError err501)
+        :<|> (\_ -> throwError err501)
```
KtorZ: These will now be accessible in Shelley.

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
```diff
@@ -0,0 +1,256 @@
...
+    :<|> stakePools
+    :<|> byronWallets
+    :<|> byronAddresses
+    :<|> byronTransactions
+    :<|> byronMigrations
+    :<|> network
+    :<|> proxy
+  where
+    byron :: ApiLayer (RndState n) t ByronKey
+    byron = error "todo: should we have a byron api here?"
```
KtorZ: Yes.

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
```diff
@@ -0,0 +1,256 @@
...
+    ( (:<|>) (..), Server, err501, throwError )
+
+-- | A diminished servant server to serve Byron wallets only.
+server
+    :: forall t n.
+        ( Buildable (ErrValidateSelection t)
+        , PaymentAddress n IcarusKey
+        , PaymentAddress n ByronKey
+        )
+    => ApiLayer (SeqState n IcarusKey) t IcarusKey
```
KtorZ: Similarly to Jörmungandr, we'll have to instantiate 3 ApiLayer: Byron, Icarus and Shelley. 
Although, what we currently call `Shelley` really is `Jormungandr` so it'll have to be renamed (and moved) first.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -0,0 +1,454 @@
...
+toSlotNo :: W.EpochLength -> W.SlotId -> SlotNo
+toSlotNo epLength =
+    SlotNo . W.flatSlot epLength
+
+-- | SealedTx are the result of rightfully constructed byron transactions so, it
+-- is relatively safe to unserialize them from CBOR.
+toGenTx :: HasCallStack => W.SealedTx -> GenTx (ShelleyBlock c)
+toGenTx = error "todo: toGenTx not implemented"
+
+fromShelleyTx :: SL.Tx c -> W.Tx
```
KtorZ: minor note: the conventions I have been trying to enforce in this file were that:

a) Identifiers from cardano-wallet are used qualified.
b) Identifiers from networking, consensus or ledger were used unqualified.  
c) Function are named after the data-type
d) `toXXX` when they translate from a cardano-wallet type to a networking, consensus & ledger data-type. `fromXXX` when they do the opposite (from networking to cardano-wallet).

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -0,0 +1,454 @@
...
+            { slotId =
+                fromSlotNo epLength $ SL.bheaderSlotNo header
+            , blockHeight =
+                fromBlockNo $ SL.bheaderBlockNo header
+            , headerHash = fromShelleyHash headerHash
+            , parentHeaderHash = fromPreviousHash hashOfNoParent $
+                SL.bheaderPrev header
+            }
+        , transactions = txs
+        , delegations  = []
```
KtorZ, line 282: Is this always the case for now or are delegation certificates somewhat already specified in the ShelleyBlock type?

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -0,0 +1,962 @@
...
+-- License: Apache-2.0
+--
+-- Network Layer for talking to Haskell re-written nodes.
+--
+-- Good to read before / additional resources:
+--
+-- - Module's documentation in `ouroboros-network/typed-protocols/src/Network/TypedProtocols.hs`
+-- - Data Diffusion and Peer Networking in Shelley (see: https://raw.githubusercontent.com/wiki/input-output-hk/cardano-wallet/data_diffusion_and_peer_networking_in_shelley.pdf)
+--     - In particular sections 4.1, 4.2, 4.6 and 4.8
+module Cardano.Wallet.Shelley.Network
```
KtorZ, line 24: Make sure to base this "fork" on the latest version of the Byron Network code which includes Rodney's latest refactoring and inclusion of the local state query (you can use a "null client" here if the protocol isn't available yet on the node's end, but at least, we have the skeleton in place on the wallet's side).

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -0,0 +1,116 @@
...
+        , parentHeaderHash =
+            Hash (BS.replicate 32 0)
+        }
+    , transactions = mkTx <$> outs
+    }
+  where
+    -- TODO: Not sure why we need this function, but should we call blake2b256
+    -- for Shelley really?
+    mkTx out@(TxOut (Address bytes) _) =
+        Tx (Hash $ blake2b256 bytes) [] [out]
```
KtorZ, line 99: This function is needed for our integration test setup. With Jörmungandr, the genesis file is mapped to a genesis block which contains a set of initial transaction (although with no inputs). On cardano-node however, the genesis contains an initial set of Genesis TxOut, so this function artificially maps that to an initial "Block". 

May you clarify this with a comment for future readers?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction/Size.hs
```diff
@@ -0,0 +1,234 @@
...
+import Cardano.Wallet.Primitive.AddressDerivation.Icarus
+    ( IcarusKey )
+import Cardano.Wallet.Primitive.Types
+    ( Address (..), Coin (..), TxIn (..), TxOut (..) )
+
+import qualified Codec.CBOR.Encoding as CBOR
+import qualified Codec.CBOR.Write as CBOR
+import qualified Data.ByteString as BS
+import qualified Data.ByteString.Lazy as BL
+
```
KtorZ: I assume none of these :point_down: will apply to Shelley on-chain data-types. We should adjust that based on the CDDL spec.

lib/shelley/test/data/cardano-node/node.cert
```diff
@@ -0,0 +1,8 @@
+{ "omega": 0
+, "issuerPk":
+    "KbEP1cIgT+ITFB/+isxQYT+yN0T47WHTZ43gvgy8kROJt9/Lsf25kpxRNU/AFsXeCnARfnwfIf2pCkTcaO1z6Q=="
+, "delegatePk":
+    "b8j8qcnO/HyB5PcDgROOrF3fGHyAe1Ec2vitqVas+ZfWdrcaTfRDIhcKT+6jsNh0oQuKi7I23shW+1X7V3Y2Pw=="
+, "cert":
+    "ea983bfd989be8d1ac1529e0d9f9ef723766397ad58ea24b89d09c40fc3122a8728f3efaca35f16ab994aa02fff76098006889f20810c40d47a34a3a6b9d3e05"
+}
```
KtorZ: I'd assume that this only applies to Byron. Shelley uses VRF/KES credentials for stake pools.

lib/shelley/test/integration/Cardano/Wallet/Shelley/Faucet.hs
```diff
@@ -0,0 +1,21 @@
+{-# LANGUAGE DataKinds #-}
+
+module Cardano.Wallet.Byron.Faucet
```
KtorZ: *Shelley.Faucet

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
Anviking: > if the protocol isn't available yet on the node's end

It was, was I was grepping wrong.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction/Size.hs
Anviking: Can do when adding tx support, I'd say.

KtorZ: Yes, but in the meantime, I'd not add this at all since we know it doesn't apply already.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
Anviking: Looks like we can access them

https://github.com/input-output-hk/cardano-ledger-specs/blob/caa0b377698aace2550d1a9948a050bfa296a3e0/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/TxData.hs#L315-L328

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
Anviking: ouroboros-network seems to consistently imports everything from the Shelley ledger as qualified `as SL`.

There are name clashes (e.g. `Block`, `Tx`) between ouroboros-network and the specs.

I think we have to use `qualified` to some extent. At least for `SL` and `W`.

lib/byron/src/Cardano/Wallet/Byron/Transaction.hs
```diff
@@ -16,7 +16,7 @@
 
 module Cardano.Wallet.Byron.Transaction
     ( newTransactionLayer
-    , fromGenesisTxOut
+    , genesisBlockFromTxOuts
```
KtorZ, line 19: :+1:

lib/shelley/src/Cardano/Wallet/Shelley.hs
```diff
@@ -0,0 +1,464 @@
...
+        -> SomeNetworkDiscriminant
+
+deriving instance Show SomeNetworkDiscriminant
+
+-- | The @cardano-wallet-shelley@ main function. It takes the configuration
+-- which was passed from the CLI and environment and starts all components of
+-- the wallet.
+serveWallet
+    :: forall t c.
+        ( t ~ IO (Shelley c)
```
KtorZ: Is there any value for us to carry this `c` parameter around? Certainly, we can already make a choice for the underlying Crypto scheme (for instance: `TPraosStandardCrypto`) as it seems to be done in:

https://github.com/input-output-hk/cardano-node/blob/master/cardano-api/src/Cardano/Api/Types.hs#L127-L146

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -0,0 +1,443 @@
...
+
+import Prelude
+
+
+import Cardano.Binary
+    ( serialize' )
+import Cardano.Chain.Common
+    ( unsafeGetLovelace )
+import Cardano.Chain.UTxO
+    ( Tx (..), TxAux, TxIn (..), TxOut (..), taTx )
```
KtorZ: I don't think any of these apply to `Shelley`. 
We should try to use things from https://github.com/input-output-hk/cardano-node/blob/master/cardano-api/src/Cardano/Api/Types.hs#L108-L114, or better, use `Cardano.Api` here to interface with cardano-wallet

lib/shelley/src/Cardano/Wallet/Shelley.hs
Anviking: Was wondering the same. Went with it to see how it would work out. Maybe we would want to use some mock crypto for… performance at some point?

But certainly we can remove it.

lib/shelley/src/Cardano/Wallet/Shelley/GenesisDataOrphans.hs
```diff
@@ -0,0 +1,197 @@
...
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+{-# OPTIONS_GHC -Wno-missing-methods #-}
+{- HLINT ignore -}
+
+-- Module copied from cardano-node.
```
KtorZ: Couldn't the module be imported instead so that the instances are put in Context?

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -0,0 +1,1101 @@
...
+--     -------------------------------------------------------------------------
+--     Client has agency*                | Idle
+--     Server has agency*                | Intersect, Next
+--
+--     * A peer has agency if it is expected to send the next message.
+--
+--      *-----------*
+--      | Intersect |◀══════════════════════════════╗
+--      *-----------*         FindIntersect         ║
+--            │                            :750
```
KtorZ: :thinking:

lib/shelley/src/Cardano/Wallet/Shelley/GenesisDataOrphans.hs
Anviking: I got the sense that depending on cardano-node would be cumbersome due to all the dependencies.

lib/shelley/src/Cardano/Wallet/Shelley.hs
KtorZ: When the crypto becomes our bottleneck in performances here, we would have come a long way :grimacing:

lib/shelley/src/Cardano/Wallet/Shelley/GenesisDataOrphans.hs
Anviking: Done, wasn't that cumbersome, but needed to add a new Iohk-monitoring plugin to the snapshot.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
Anviking: 👍Fixed and changed the commits to be more
1. Copy byron implementation
2. Tweak

for nicer diffs.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -698,131 +670,98 @@ chainSyncWithBlocks
...
-        -- We only pipeline requests when it's safe to pipeline them. Safe
-        -- is when all blocks are considered 'stable' and immutable, i.e.
-        -- when we land more than `k` blocks from the tip, so when we start
-        -- more than `k + maxInFlight` blocks from the tip
-        --
-        --     0                       tip - k           tip
-        --     |---------------------------|-------------->
-        --               pipelined            one by one
-        --
-        pipeline
```
Anviking: When I started, pipelining wasn't on master. 

Should be simple to revert this deletion now.

Could also factor out the logic to a block-agnostic shared location… or maybe in the next PR.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -1132,6 +1070,12 @@ instance HasSeverityAnnotation NetworkLayerLog where
         MsgIntersectionFound{}     -> Info
         MsgPostSealedTx{}          -> Debug
         MsgLocalStateQuery{}       -> Debug
-        MsgNodeTip{}               -> Debug
+        MsgNodeTip{}               -> Info
```
Anviking: Should revert this and similar things if any.

lib/shelley/exe/cardano-wallet-shelley.hs
Anviking: > These will have to be change to point to the "normal" API.

Done.

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
Anviking: The shelley wallet endpoints are now added.

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
Anviking: Should be done.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
Anviking: 👍 I simply removed them for now.

Anviking:
bors try

Anviking:
Ok, need to add `cardano-node` to the CI environment

Anviking:
bors try

lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -378,6 +354,12 @@ mkWalletClient bp chainSyncQ =
         , getGenesisBlockHash
         } = bp
 
+    codecs :: MonadST m => ClientCodecs ByronBlock m
+    codecs = clientCodecs
+        (ByronCodecConfig (toEpochSlots getEpochLength))
+        ByronNodeToClientVersion1
```
KtorZ: Aren't we using `Version2` ?

lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -458,20 +427,24 @@ mkTipSyncClient tr gbp localTxSubmissionQ onTipUpdate onTxParamsUpdate = do
...
             let
                 tr' = contramap MsgLocalStateQuery tr
-                codec = codecLocalStateQuery
-                    (encodePoint encodeByronHeaderHash)
-                    (decodePoint decodeByronHeaderHash)
-                    nodeEncodeQuery
-                    nodeDecodeQuery
-                    nodeEncodeResult
-                    nodeDecodeResult
+                codec = cStateQueryCodec codecs
```
KtorZ, line 430: Sweet. Nice simplification.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
Anviking: Hm… I was going to say that this is another kind of version… but yes — we should probably use https://github.com/input-output-hk/ouroboros-network/blob/02a808741ef3fc8b244c4be221616e4fc92a8310/ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Ledger/NetworkProtocolVersion.hs#L32

lib/shelley/src/Cardano/Wallet/Shelley.hs
```diff
@@ -0,0 +1,454 @@
...
+
+-- | Encapsulate a network discriminant and the necessary constraints it should
+-- satisfy.
+data SomeNetworkDiscriminant where
+    SomeNetworkDiscriminant
+        :: forall (n :: NetworkDiscriminant).
+            ( NetworkDiscriminantVal n
+            , PaymentAddress n IcarusKey
+            , PaymentAddress n ByronKey
+            , PaymentAddress n ShelleyKey
```
KtorZ, line 157: This is incorrect at the moment, but we can solve later. We need to:

a) Rename 'ShelleyKey' to 'JormungandrKey'
b) Port https://github.com/input-output-hk/cardano-addresses/blob/master/src/Cardano/Address/Style/Shelley.hs

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -0,0 +1,413 @@
...
+fromMaxTxSize :: Natural -> Quantity "byte" Word16
+fromMaxTxSize =
+    Quantity . fromIntegral
+
+fromPParams :: SL.PParams -> W.TxParameters
+fromPParams pp = W.TxParameters
+    { getFeePolicy = W.LinearFee
+        (Quantity (naturalToDouble $ SL._minfeeA pp))
+        (Quantity (fromIntegral $ SL._minfeeB pp))
+        (Quantity 0) -- TODO: it's not as simple as this?
```
KtorZ, line 354: No it's not indeed. To be addressed when we re-introduce certificates.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -0,0 +1,413 @@
...
+    mkTxOut (addr, coin) = W.TxOut
+        (W.Address $ serialize' addr)
+        (W.Coin $ fromIntegral coin)
+
+fromNetworkMagic :: NetworkMagic -> W.ProtocolMagic
+fromNetworkMagic (NetworkMagic magic) =
+    W.ProtocolMagic (fromIntegral magic)
+
+fromProtocolMagicId :: ProtocolMagicId -> W.ProtocolMagic
+fromProtocolMagicId = W.ProtocolMagic . fromIntegral . unProtocolMagicId
```
KtorZ: I thought the concept of protocolMagic was gone in Shelley :thinking: ? This one is probably unused.

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -0,0 +1,373 @@
...
+import qualified Data.Aeson as Aeson
+import qualified Data.ByteString.Lazy as BL
+import qualified Data.HashMap.Strict as HM
+import qualified Data.Text as T
+import qualified Data.Yaml as Yaml
+
+data NetworkConfiguration where
+    MainnetConfig
+        :: (SomeNetworkDiscriminant, NodeVersionData)
+        -> NetworkConfiguration
```
KtorZ: I wonder if we shouldn't remove anything regarding "Shelley mainnet" at this stage.

lib/shelley/test/integration/Main.hs
```diff
@@ -0,0 +1,220 @@
...
+main :: forall t . t ~ Shelley => IO ()
+main = withUtf8Encoding $ withLogging Nothing Info $ \(_, tr) -> do
+    hSetBuffering stdout LineBuffering
+    hspec $ do
+        describe "No backend required" $ do
+            describe "Mnemonics CLI tests" $ parallel (MnemonicsCLI.spec @t)
+            describe "Miscellaneous CLI tests" $ parallel (MiscellaneousCLI.spec @t)
+            describe "Key CLI tests" $ parallel (KeyCLI.spec @t)
+        describe "API Specifications" $ specWithServer tr $ do
+            Network.spec
```
KtorZ, line 129: What about the wallet specs?

Anviking: They need transactions. For faucet txs I think.

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -0,0 +1,1124 @@
...
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NumericUnderscores #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE TypeApplications #-}
+
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+module Cardano.Wallet.Shelley.TransactionSpec
```
KtorZ: Most of this module is irrelevant for Shelley since the transaction format has completely changed since Byron. I'd suggest to remove it; it'll be re-implement once we implement the transaction layer for Shelley.

stack.yaml
```diff
@@ -1,4 +1,4 @@
-resolver: https://raw.githubusercontent.com/input-output-hk/cardano-haskell/master/snapshots/cardano-1.11.0.yaml
+resolver: https://gist.githubusercontent.com/Anviking/3c48b7343002d58f218b87436bfdaf0c/raw/877a252c197b36632dc52f14370fae3f3ea9c834/cardano-1.12.2.yaml
```
KtorZ: Let's have this as a pre-1.12.0 snapshot on cardano-haskell

KtorZ (approved):

lib/shelley/test/integration/Main.hs
KtorZ: > For faucet txs I think.

Precisely not :) The faucets are genesis UTxOs, so most of the wallet scenarios are actually about restoring wallets from on-chain data.

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
Anviking: 48cf00bd5

lib/shelley/test/integration/Main.hs
Anviking: Ok. The node has stopped producing blocks for some reason 🤔

Anviking: There are a few scenarios requiring transactions. I can't mark them pending since they exist in core-integration. Let's deal with this later.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
Anviking: Done

stack.yaml
Anviking: Done.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
Anviking: Done

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
Anviking: Good point, probably. Done.

Anviking:
bors try

Anviking:
bors try

rvl:
I have updated the version of cardano-node used for Buildkite. With this version, `cardano-wallet-shelley:test:integration` passes for me locally. I have also added more shelley to the nix.

bors try

rvl:
In that last bors run the shelley integration tests have passed but not the jormungandr unit tests. There was also a problem with the nix build which I think I have fixed.

Anviking:
Thanks!

> jormungandr unit tests

🤔 
bors try

Anviking:
```
ined reference to `_gcry_mpih_addmul_1'
/nix/store/vcnbgxj6qn1xmawrcsw63vd9bs1vllrc-x86_64-unknown-linux-musl-binutils-2.31.1/bin/x86_64-unknown-linux-musl-ld: ./.libs/libgcrypt.so: undefined reference to `_gcry_mpih_rshift'
/nix/store/vcnbgxj6qn1xmawrcsw63vd9bs1vllrc-x86_64-unknown-linux-musl-binutils-2.31.1/bin/x86_64-unknown-linux-musl-ld: ./.libs/libgcrypt.so: undefined reference to `_gcry_mpih_add_n'
/nix/store/vcnbgxj6qn1xmawrcsw63vd9bs1vllrc-x86_64-unknown-linux-musl-binutils-2.31.1/bin/x86_64-unknown-linux-musl-ld: ./.libs/libgcrypt.so: undefined reference to `_gcry_mpih_mul_1'
/nix/store/vcnbgxj6qn1xmawrcsw63vd9bs1vllrc-x86_64-unknown-linux-musl-binutils-2.31.1/bin/x86_64-unknown-linux-musl-ld: ./.libs/libgcrypt.so: undefined reference to `_gcry_mpih_submul_1'
/nix/store/vcnbgxj6qn1xmawrcsw63vd9bs1vllrc-x86_64-unknown-linux-musl-binutils-2.31.1/bin/x86_64-unknown-linux-musl-ld: ./.libs/libgcrypt.so: undefined reference to `_gcry_mpih_lshift'
/nix/store/vcnbgxj6qn1xmawrcsw63vd9bs1vllrc-x86_64-unknown-linux-musl-binutils-2.31.1/bin/x86_64-unknown-linux-musl-ld: ./.libs/libgcrypt.so: undefined reference to `_gcry_mpih_sub_n'
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:728: mpicalc] Error 1
make[2]: Leaving directory '/build/libgcrypt-1.8.5/src'
make[1]: *** [Makefile:497: all-recursive] Error 1
make[1]: Leaving directory '/build/libgcrypt-1.8.5'
make: *** [Makefile:429: all] Error 2
builder for '/nix/store/v3s67c98zmc3jlw9jsryk4yy7hf1vz0k-libgcrypt-1.8.5-x86_64-unknown-linux-musl.drv' failed with exit code 2
```

Anviking:
bors r-

Seems problems due to timeouts are expected today
https://input-output-rnd.slack.com/archives/CAP8NM7N0/p1589424744022100

Will try again at the end of the day

PR #1630, opened 2020-05-06
Reduce logging from network layer

rvl:
### Issue Number

Relates to #1577 / ADP-83.

### Overview

- Set network tip log level to debug
- Improve formatting of text log message
- Only log the tip when it changes, because the node can seem to send updates when nothing changes

### Comments

This is what it looks like now:

```
[cardano-wallet.network:Debug:18] [2020-05-06 07:49:54.10 UTC] Network node tip block height is 4125310 at hash c5cb3bcd
```

### DIFF
--- a/lib/byron/src/Cardano/Wallet/Byron/Network.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron/Network.hs
@@ -40,6 +40,7 @@ import Cardano.Chain.Byron.API
     ( ApplyMempoolPayloadErr (..) )
 import Cardano.Wallet.Byron.Compatibility
     ( Byron
+    , fromByronHash
     , fromChainHash
     , fromSlotNo
     , fromTip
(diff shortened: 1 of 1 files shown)


lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -497,6 +494,11 @@ mkTipSyncClient tr localTxSubmissionQ onTipUpdate onTxParamsUpdate = do
             Right (Right ls) ->
                 onTxParamsUpdate' $ txParametersFromUpdateState ls
 
+    onTipUpdate' <- debounce $ \tip -> do
```
jonathanknowles, line 497: :+1:

jonathanknowles (approved):
Nice!

I particularly like the use of the `debounce` function.

rvl:
Thanks

bors r+

PR #1632, opened 2020-05-06
Prevent bors from merging into any branch except master

rvl:
I got bitten by this today, doh.

PR #1635, opened 2020-05-07
Prevent bors from merging into any branch except master

rvl:
Another try at #1632.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -5,11 +5,13 @@ env:
   LC_ALL: "en_US.UTF-8"
 
 steps:
-  - label: 'hydra-eval-errors'
-    command: 'nix-build ./nix -A iohkNix.hydraEvalErrors && ./result/bin/hydra-eval-errors.py'
-    if: 'build.tag == null'
+  - label: 'Prevent merging to wrong branch'
+    if: 'build.branch == "bors/staging"'
+    command: './.buildkite/check-bors.sh'
     agents:
       system: x86_64-linux
+  - wait: ~
+    if: 'build.branch == "bors/staging"'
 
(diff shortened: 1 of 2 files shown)


KtorZ (approved):

KtorZ:
:point_up: should fail.

KtorZ:
--> https://buildkite.com/input-output-hk/cardano-wallet/builds/7370#edb2b04f-54c1-4156-812f-9cac2b67ac3d

So this fails, but the rest of the job is still running. Is there a "fail fast" option on bors :thinking: ? Or on buildkite actually, because bors has probably no power here and is simply waiting for Buildkite to notify.

KtorZ:
Seems like we could mark all other steps to `depends_on` this one, to make sure they only run if necessary.

.buildkite/pipeline.yml
```diff
@@ -5,11 +5,13 @@ env:
...
-  - label: 'hydra-eval-errors'
-    command: 'nix-build ./nix -A iohkNix.hydraEvalErrors && ./result/bin/hydra-eval-errors.py'
-    if: 'build.tag == null'
+  - label: 'Prevent merging to wrong branch'
+    if: 'build.branch == "bors/staging"'
+    command: './.buildkite/check-bors.sh'
     agents:
       system: x86_64-linux
+  - wait: ~
+    if: 'build.branch == "bors/staging"'
```
KtorZ, line 14: Also :+1:

KtorZ (approved):

rvl:
I added a conditional wait step which should make it fail fast.
May need to wait a little while to test because buildkite are doing some unscheduled db maintenance.

KtorZ:
Hydra timing out again.

PR #1638, opened 2020-05-07
Modify docker-compose.yml example so it is possible to run mainnet and testnet interchangeably

piotr-iohk:
# Issue Number

N/A

# Overview

- [ ] I have modified docker-compose.yml example so one can run it for `testnet` and `mainnet` interchangeably.

# Comments

With current docker-compose.yml if one starts:
```
NETWORK=testnet docker-compose up
```
...runs it for a while and then starts:

```
NETWORK=mainnet docker-compose up
```
then she will get failure:
```
cardano-node_1    | Shutting down..
cardano-node_1    | ProtocolMagicIdMismatch "/data/db/protocolMagicId" (ProtocolMagicId {unProtocolMagicId = 764824073}) (ProtocolMagicId {unProtocolMagicId = 1097911063})
cardano-node_1    |
```
as the volumes will interfere with each other. 

After this change one can run it for `testnet` and `mainnet` interchangeably...

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ start with Daedalus.
 > [cardano-sl](https://github.com/input-output-hk/cardano-sl)
 > repository.
 
-## Getting Started 
+## Getting Started
 
 ```
 wget https://raw.githubusercontent.com/input-output-hk/cardano-wallet/master/docker-compose.yml
@@ -52,6 +52,12 @@ Fantastic! The server is up-and-running, waiting for HTTP requests on `localhost
 
 ```
 curl http://localhost:8090/v2/network/information
+```
+
+or to be accessed via CLI, e.g.:
+
+```
+docker run --network host --rm inputoutput/cardano-wallet network information
 ```
 
 See also [Wiki - Docker](https://github.com/input-output-hk/cardano-wallet/wiki/Docker) for more information about using docker.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ services:
     environment:
       NETWORK:
     volumes:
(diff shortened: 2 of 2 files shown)


docker-compose.yml
```diff
@@ -30,7 +30,9 @@ services:
     restart: on-failure
 
 volumes:
-  node-db:
+  node-mainnet-db:
+  node-testnet-db:
+  wallet-mainnet-db:
+  wallet-testnet-db:
```
KtorZ, line 36: out of curiosity, did you try templating this with an ENV var too :thinking: ? 

```
node-${NETWORK}-db: 
wallet-${NETWORK}-db:
```

?

README.md
```diff
@@ -54,6 +54,12 @@ Fantastic! The server is up-and-running, waiting for HTTP requests on `localhost
 curl http://localhost:8090/v2/network/information
 ```
 
+or to be accessed via CLI, e.g.:
+
+```
+docker run --network host inputoutput/cardano-wallet network information
```
KtorZ: `run --rm`, otherwise things rapidly get out of hands ^^"

KtorZ (approved):

docker-compose.yml
piotr-iohk: I did, it doesn't work :(

```
ERROR: The Compose file './docker-compose.yml' is invalid because:
volumes value 'node-${NETWORK}-db', 'wallet-${NETWORK}-db' do not match any of the regexes: '^[a-zA-Z0-9._-]+$'
```

README.md
piotr-iohk: fixed.

Issue #1637, opened 2020-05-07
Use address derivation from cardano-addresses


paweljakubas:
# Context

We have separate cardano-addresses repo that contains implementation of mnemonics and address derivation interface along with byron, icarus, jomrungandr and shelley implementations

# Decision

We want to use address-derivation's interfaces and impl inside cardano-wallet, and replace it with current ones

# Acceptance Criteria

1. `HardDerivation`, `SoftDerivation`, `StakingDerivation`, `PaymentAddress` and `DelegationAddress`
2. Use crypto primitives from cardano-addresses 
3. Use `Depth`, `Index`, DerivationType`, `AccountStyle` from cardano-addresses
4. Styles and corresponding unit tests are removed
5. The rest codebase is adjusted

---

# Development 

# QA

KtorZ:
Tracked in JIRA as: https://jira.iohk.io/browse/ADP-438

PR #1640, opened 2020-05-07
Absorbing cardano addresses - crypto bits

paweljakubas:
# Issue Number

#1637 

# Overview

- [x] I have used crypto primitives from cardano-addresses

# Comments


lib/cli/src/Cardano/CLI.hs
```diff
@@ -177,11 +179,8 @@ import Cardano.Wallet.Primitive.AddressDerivation
     , PassphraseMaxLength
     , PassphraseMinLength
     , WalletKey (..)
-    , XPrv
-    , XPub
     , deriveRewardAccount
     , hex
-    , unXPrv
     , unXPrvStripPubCheckRoundtrip
     , xPrvFromStrippedPubXPrvCheckRoundtrip
```
Anviking, line 186: If `xprvFromBytes` never modifies keys (but fails), we could replace `xPrvFromStrippedPubXPrvCheckRoundtrip` and `unXPrvStripPubCheckRoundtrip` too I suspect.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -680,35 +671,14 @@ data ErrUnXPrvStripPub
 -- - the key is an old byron key
 unXPrvStripPubCheckRoundtrip :: XPrv -> Either ErrUnXPrvStripPub ByteString
 unXPrvStripPubCheckRoundtrip k = do
-    let res = unXPrvStripPub k
-    case (fmap unXPrv . xPrvFromStrippedPubXPrv $ res) of
+    let res = xprvToBytes k
+    case (fmap CC.unXPrv . xPrvFromStrippedPubXPrv $ res) of
```
KtorZ, line 674: This is just `xprvFromBytes`

lib/jormungandr/cardano-wallet-jormungandr.cabal
```diff
@@ -187,7 +189,7 @@ test-suite unit
       Cardano.Wallet.Jormungandr.RewardsSpec
       Cardano.Wallet.Jormungandr.TransactionSpec
 
-test-suite jormungandr-integration
+test-suite integration
```
KtorZ: Careful here. This change was intended to allow filtering out Jörmungandr integration tests from buildkite.

paweljakubas: ok, I just find it very clumsy to run tests using `...jormungandr:jormunagndr-integration`. I will revert it then...

lib/cli/src/Cardano/CLI.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/1640/commits/e9d860edc854e575712c720405e06aa515ea77f7

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/1640/commits/e9d860edc854e575712c720405e06aa515ea77f7

lib/jormungandr/cardano-wallet-jormungandr.cabal
paweljakubas: reverted in https://github.com/input-output-hk/cardano-wallet/pull/1640/commits/e9d860edc854e575712c720405e06aa515ea77f7

paweljakubas:
bors try

paweljakubas:
bors try

lib/cli/src/Cardano/CLI.hs
Anviking: a6ba786d7bdd1a3128161789ba4ed6a4380dceea Ok, so I guess it didn't work.

Btw, have we thought about whether `xprvFromBytes` is dangerous? Can users of `cardano-addresses` accidentally mess up by having their keys silently modified ("tweaked")?

paweljakubas:
I see : 
```
1) API Specifications, BYRON_ADDRESSES, ADDRESS_CREATE_06 - Cannot create an address that already exists
```

lib/core-integration/src/Test/Integration/Scenario/CLI/Keys.hs
```diff
@@ -74,17 +74,13 @@ spec  = do
...
               \5d3661deb9064f2d0e03fe85d68070b2fe33b4916059658e28ac7f7f91ca4b12"
 
-        it "bryon keys fail" $ do
-            keyStderr ["child", "--path", "0"] byronKey
-                `shouldReturn`
-                "That extended private key looks \
-                \weird. Is it encrypted? Or is it an old Byron key?\n"
+        it "byron keys does not fail anymore" $ do
+            keySuccess ["child", "--path", "0"] byronKey
+                `shouldReturn` ""
```
paweljakubas: please double check this change

lib/core-integration/src/Test/Integration/Scenario/CLI/Keys.hs
```diff
@@ -74,17 +74,13 @@ spec  = do
...
 
-        it "encrypted keys fail" $ do
-            keyStderr ["child", "--path", "0"] encryptedKey
-            `shouldReturn`
-                "That extended private key looks \
-                \weird. Is it encrypted? Or is it an old Byron key?\n"
+        it "encrypted byron keys does not fail anymore" $ do
+            keySuccess ["child", "--path", "0"] encryptedKey
+                `shouldReturn` ""
```
paweljakubas, line 101: please double check this change

lib/cli/src/Cardano/CLI.hs
paweljakubas: to my best knowledge, as the `cardano-address` implementation works with Byron key, which wasn't the case for `cardano-wallet` impl, we are fine here with this change.

lib/core-integration/src/Test/Integration/Scenario/CLI/Keys.hs
Anviking: What does "byron keys does not fail anymore" and `shouldReturn ""` mean? 🧐

lib/cli/src/Cardano/CLI.hs
paweljakubas: but i need to change the expected behavior of two  Keys tests in CLI as they exploited the tweak present in `cardano-wallet` - see my comment (asking for double-check) below for those tests

lib/core-integration/src/Test/Integration/Scenario/CLI/Keys.hs
Anviking: Ok, right, `keySuccess` is new. So this means that the stderr is empty, but you don't check the actually returned key.

paweljakubas: yeah, improved on this test : https://github.com/input-output-hk/cardano-wallet/pull/1640/commits/014e26b82df20b797eb2187373cc8017a72ad290

paweljakubas:
bors try

paweljakubas:
tried locally to reproduce :
```
1) API Specifications, BYRON_ADDRESSES, ADDRESS_CREATE_06 - Cannot create an address that already exists
```
but CANNOT

lib/cli/src/Cardano/CLI.hs
Anviking: > to my best knowledge, as the cardano-address implementation works with Byron key, which wasn't the case for cardano-wallet impl, we are fine here with this change.

Ok, good, that was what I hoped.

I tried this to be completely convinced, and it passes

```haskell
prop_roundtrip
    :: XPrv
    -> Index 'WholeDomain 'AddressK
    -> Property
prop_roundtrip xprv ix =
    xprvFromBytes (xprvToBytes k) === Just k
  where
    rootXPrv = genMasterKeyFromXPrv xprv :: Byron 'RootK XPrv
    k = deriveXPrv DerivationScheme1 (getKey rootXPrv) ix
```

paweljakubas:
once again ```1) API Specifications, BYRON_ADDRESSES, ADDRESS_CREATE_04 - Can list address after creating it```
locally cannot I replicate this ;-/

lib/jormungandr/cardano-wallet-jormungandr.cabal
rvl: I've not tried it but you _might_ be able to use to rename back to the slightly shorter component name and use `--skip cardano-wallet-jormungandr:test:integration` in `.buildkite/rebuild.hs`.

rvl:
The integration test failures are probably due to #1644.

rvl (approved):
OK so this is replacing the types and functions from Cardano.Crypto with the wrapper from Cardano.Address.Derivation.
Tests still use CC functions for calculating expected values.

Should I be expecting to see any deleted files, or not yet?

paweljakubas:
So the plan is : 
1. reuse crypto bits from cardano-addresses we can
2. replace AddressDerivation, Index,... with the ones from cardano-addresses - and here a number of files will be deleted (along with corresponding unit tests).

Point 1 - this PR
Point 2 - it was decided to do it after shelley integrationas part of not-so-small-refatoring (@KtorZ )

paweljakubas:
bors try

lib/core-integration/src/Test/Integration/Scenario/CLI/Keys.hs
```diff
@@ -74,17 +74,21 @@ spec  = do
...
+                \3b209bcb6f8af16317702127ab4169d2d7f11807d464b2f427b3ca45d5d9f0\
+                \578288\n"
+
+        it "encrypted byron keys does not fail anymore" $ do
+            res <- key ["child", "--path", "0"] encryptedKey
+            res `shouldBe`
+                "7d0ef39b2f7de32e8ca24cfe7a9cf1663a860778c42c70ef41ad851fa6606c\
+                \bbcfee7fbacf0e7141e449aa18e8d0c7206c0b7a71d3129751b516bf0a2abb\
+                \2383ca7b962e3bc5aa7d049f33606c8851a7ba94432ef335e47d77575e2fce\
+                \a9813c\n"
```
KtorZ, line 100: Why though? I'd still expect these (= encrypted keys) to fail.

lib/core/src/Cardano/Byron/Codec/Cbor.hs
```diff
@@ -593,7 +594,7 @@ decodeUpdateProof = do
 --   passphrases). This is just scheme-specific and is better left out of this
 --   particular function
 encodeAddress :: XPub -> [CBOR.Encoding] -> CBOR.Encoding
-encodeAddress (XPub pub (ChainCode cc)) attrs =
+encodeAddress (CC.XPub pub (CC.ChainCode cc)) attrs =
```
KtorZ: The import of `CC` is unnecessary here. This pattern matchin & `pub <> cc` below could be replaced with `xpubToBytes xpub`

KtorZ (approved):
LGTM, although one minor reserve about one of the test which was changed. I don't understand why this test is now passing.

lib/core-integration/src/Test/Integration/Scenario/CLI/Keys.hs
Anviking: I think it used to fail because the encrypted keys didn't (appear) to have the correct tweaked bits.

Ignoring the tweak, there's no way the tell that the key is encrypted. (I think)

KtorZ: > there's no way the tell that the key is encrypted. 

We can try changing its passphrase to `""`. If the key was indeed non encrypted, this should be a no-op :thinking:

Anviking: > We can try changing its passphrase to "".

Not sure how that works, but keep in mind that the 96-byte format — unlike a `XPrv` — doesn't contain the public key.

lib/core/src/Cardano/Byron/Codec/Cbor.hs
paweljakubas: indeed - done in https://github.com/input-output-hk/cardano-wallet/pull/1640/commits/19cf8232cf85c76e63afd7a10e202abfd7d7eea9

Anviking (approved):
Lgtm!

Though not sure how good or bad idea it is to support byron and encrypted keys in general.

think we can remove the checkRoundtrip functions -> created #1656

paweljakubas:
once again problem with : 
```
1) API Specifications, BYRON_ADDRESSES, ADDRESS_CREATE_06 - Cannot create an address that already exists
```

KtorZ:
@Anviking it won't work since hydra will re-use the failing cache build. The PR needs to be updated with a new commit hash.

PR #1643, opened 2020-05-07
Add property and fix for markCharsRedAtIndices

Anviking:
# Issue Number

#1603 

# Overview

- [x] Add property for and fix questionable behaviour of `markCharsRedAtIndices`

# Comments

- Step by step property failures in  72d5bb9
- Don't think the problems affected the error rendering (or at least not commonly), or I'd noticed yesterday.
- Should still have a another calm look at this to be safe 😬

### DIFF
--- a/lib/cli/cardano-wallet-cli.cabal
+++ b/lib/cli/cardano-wallet-cli.cabal
@@ -76,6 +76,7 @@ test-suite unit
     , cardano-wallet-cli
     , cardano-wallet-core
     , cardano-wallet-launcher
+    , containers
     , filepath
     , hspec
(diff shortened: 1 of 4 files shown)


Anviking:
bors try

rvl (approved):
The little functions are usually the tricky ones. :-)

For the property, can I suggest that it be "all the red characters correspond to errors"? So scan through the error string and get all the characters that would appear red in a terminal. Assert that this is equal to the characters of the error indices (sorted).

KtorZ:
merging manually until we figure out how to fix Hydra.

PR #1645, opened 2020-05-08
Import legacy Byron addresses

KtorZ:
# Issue Number

ADP-292

# Overview

- f71d8a890417cc0d029dbebba43337a3951d979b
  :round_pushpin: **add 'putRandomAddress' to the swagger specifications and API servant definition**
  
- db29babc7fdeff92133a48ed40196b04aa81affe
  :round_pushpin: **write dummy server implementation for 'putRandomAddress' and fill the gap in various APIs**
  
- 46aa7c5c14dfa0a8865eda0fab7575805f4c6309
  :round_pushpin: **add integration tests for byron address restoration**
  
- 3f255e77a21b5ea4d2635a6378fa020531be1860
  :round_pushpin: **implement server and application handlers for importing an address**
  
- f7ba8fb485946175ca85775e675c7989c0eafed8
  :round_pushpin: **add command for 'address import' mapping to the newly introduced endpoint**
  
- 2ae81f6049425cdfa8892cd6c0ada71306cc2f62
  :round_pushpin: **add some integration tests scenario for 'address import'**

- 73294aeb92297b890e0a67d9848dbdaf7c359fae
  :round_pushpin: **fix non-atomic checkpoint modification in {import,create}RandomAddress**
    This is subtle but, everytime we are tempted to write:

  ```hs
  cp <- atomically $ readCheckpoint wid
  {- Do some stuff... -}
  atomically $ putCheckpoint wid cp'
  ```

  we should in reality really go for:

  ```hs
  atomically $ do
    cp <- readCheckpoint wid
    {- Do some stuff... -}
    putCheckpoint wid cp'
  ```

  because there's no guarantee that the checkpoint hasn't already
  changed between the moment we fetched it, and the moment we insert a
  modification. This could have disastreous effects like actually
  updating a checkpoint in the past while a new one has already been
  created with now oudated data; or worse, in the case of EBB, it'll
  simply override the newly written checkpoint and makes the database
  silently roll back of exactly 1 block!

# Comments

Screenshot from 2020-05-08 18-05-25

```
$ cardano-wallet-byron address import --help
Usage: cardano-wallet-byron address import [--port INT] WALLET_ID ADDRESS
  Import a random address generated elsewhere. Only available for random
  wallets. The address must belong to the target wallet.

Available options:
  --port INT               port used for serving the wallet API. (default: 8090)
```

### DIFF
--- a/specifications/api/swagger.yaml
+++ b/specifications/api/swagger.yaml
@@ -1144,6 +1144,15 @@ x-parametersStakePoolId: &parametersStakePoolId
     maxLength: 64
     minLength: 64
 
+x-parametersAddressId:  &parametersAddressId
+  in: path
+  name: addressId
+  required: true
+  schema:
+    type: string
+    format: base58
+    example: DdzFFzCqrhtCNjPk5Lei7E1FxnoqMoAYtJ8VjAWbFmDb614nNBWBwv3kt6QHJa59cGezzf6piMWsbK7sWRB5sv325QqWdRuusMqqLdMt
+
 x-parametersStartDate: &parametersStartDate
   in: query
   name: start
@@ -1597,6 +1606,13 @@ x-responsesPostRandomAddress: &responsesPostRandomAddress
       application/json:
         schema: *ApiAddress
 
+x-responsesPutRandomAddress: &responsesPutRandomAddress
+  <<: *responsesErr400
+  <<: *responsesErr403
+  <<: *responsesErr405
+  204:
+    description: No Content
+
 #############################################################################
 #                                                                           #
 #                                  PATHS                                    #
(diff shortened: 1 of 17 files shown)


disassembler (approved):

KtorZ:
bors try

specifications/api/swagger.yaml
```diff
@@ -2031,6 +2047,22 @@ paths:
               schema: *ApiPostRandomAddressData
         responses: *responsesPostRandomAddress
 
+  /byron-wallets/{walletId}/addresses/{addressId}:
+      put:
```
rvl, line 2051: Don't the REST dogmas say that PUT is for updating and POST is for creating?

rvl (approved):
Looks good.

specifications/api/swagger.yaml
KtorZ: Not necessarily. `PUT` is for pushing an already created resource  assuming that a client does have the ability to create such resource entirely. This is the case here where the address pre-exists to the call, and is therefore pushed to the server. 

`POST` would be for creating a resource from partial data, or from a "constructor". There are resources where this is not possible to do differently, like the wallet. We can't possibly allow a client to "put" its own balance or sync progress, hence the POST.

KtorZ: Then, for choosing between `PUT` and `PATCH`, it's more of a matter of perspective. We are not using `PATCH` much in the API where it would be legitimate IMO (like when updating the passphrase of metadata, i.e. when doing a partial update of a resource). Yet, we've stick to `PUT` here because people tend to be somewhat confused by the `PATCH` method. And, from the perspective of a sub-resource, a `PUT` is still acceptable for partially replacing an entity. But this is why we write:

```
PUT /wallets/{walletId}/passphrase
```

and not:

```
PUT /wallets/{walletId}

{ "passphrase:" ... }
```

KtorZ:
waiting for Hydra again takes ages :(

Issue #1648, opened 2020-05-11
Add statistics to fee estimation API endpoint


rvl:
# Context

Jira user story is ADP-283 Reduce variability in fee estimation

> As an API user,
> I want to be able to rely on a somewhat stable fee estimation,
> so that I can better inform my users about the cost of a transaction.

Fee estimation in its current form can be a little confusing. Because the underlying coin selection is a (non-seeded) random algorithm, it is hard to know which coins will be picked for payment in advance.

The current fee estimation in cardano-wallet is quite naive: it applies the normal coin selection process, balance the selection and returns the fees corresponding to that particular selection. Yet there's no guarantee than upon making the real payment, the same coins will be selected, and therefore, the same fee amount will be paid.

Of course, in practice the variability on small UTxO sets is quite small, but it still exists. So, a better alternative would be to run multiple selections, and returns a more elaborated response to the API client with:

 * max fee
 * 9th decile

so we can provide more reliable statistical answers like "90% of the transactions cost at least X, but not more than Y."

# Decision

1. Add a new interface to the fee estimation function which repeats the process and returns statistics along with the result.
2. Change cardano-wallet-core to use this new function.
3. Adjust API definitions to add more information to the `/wallets/{walletId}/payment-fees` endpoint.

## To be decided

1. How many times to run the estimation. e.g. run it until the statistical distribution is confidently known.
2. Which parts to implement in cardano-coin-selection vs cardano-wallet-core.
3. Check that monte carlo simulation is the best method to get this answer.

# Acceptance Criteria

1. The fee estimation API _must_ provide statistical background information about its result.
2. The following statistics _must_ be returned from the API:
   - max fee
   - 9th decile

---

# Development 

1. Implement in cardano-wallet-core for now.
2. Ensure that the estimation function can be unit tested.
3. Repeat the sampling a fixed number of times - 100 samples.
4. Update swagger spec and notify Daedalus team.

# QA 

1. This adds a property test for the estimation function in `lib/core/test/unit/Cardano/WalletSpec.hs`.
2. Integration tests that deal with transaction estimates are updated to use estimatedMin and estimatedMax.

piotr-iohk:
@rvl two questions:

1. Currently the "fee" response is as follows:
```
{
"estimated_min": {
"quantity": 42000000,
"unit": "lovelace"
},
"estimated_max": {
"quantity": 42000000,
"unit": "lovelace"
}
}
```
[...]

PR #1653, opened 2020-05-12
Run multiple coin selections to estimate fees

rvl:
### Issue Number

Relates to ADP-283 / #1648.

### Overview

- Extends `POST /wallets/{id}/payment-fees` API endpoint with a "better" estimation. Adds fields for expected minimum and maximum fees.

### Comments

- [x] decide on whether the api change should be breaking or backwards compatible
- [x] also do similar fee estimation for delegation transactions
- [x] better handling of coin selection errors while sampling
- [ ] determine how many samples to take, or a criteria for stopping sampling.
- [x] decide on quantile calculation method (currently using median).
- [x] unit test `estimateFeeForCoinSelection`

### DIFF
--- a/lib/byron/test/integration/Test/Integration/Byron/Scenario/CLI/Transactions.hs
+++ b/lib/byron/test/integration/Test/Integration/Byron/Scenario/CLI/Transactions.hs
@@ -582,7 +582,9 @@ scenario_TRANS_ESTIMATE_01_02 fixtureSource fixtures = it title $ \ctx -> do
             , nChanges = length fixtures
             }
     verify r
-        [ expectCliField #amount $ between (Quantity feeMin, Quantity feeMax) ]
+        [ expectCliField #estimatedMin $
+            between (Quantity feeMin, Quantity feeMax)
+        ]
   where
     title = "CLI_TRANS_ESTIMATE_01/02 - " ++ show (length fixtures) ++ " recipient(s)"
 
@@ -751,7 +753,9 @@ scenario_TRANS_ESTIMATE_04b = it title $ \ctx -> do
             , nChanges = 0
             }
     verify r
-        [ expectCliField #amount $ between (Quantity feeMin, Quantity feeMax) ]
+        [ expectCliField #estimatedMin $
+            between (Quantity feeMin, Quantity feeMax)
+        ]
   where
     title = "CLI_TRANS_ESTIMATE_04 - Can't cover fee"
 
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -95,6 +95,7 @@ module Cardano.Wallet
     -- ** Payment
     , selectCoinsExternal
     , selectCoinsForPayment
+    , estimateFeeForPayment
     , signPayment
     , ErrSelectCoinsExternal (..)
     , ErrSelectForPayment (..)
@@ -112,6 +113,7 @@ module Cardano.Wallet
     , joinStakePool
     , quitStakePool
     , selectCoinsForDelegation
+    , estimateFeeForDelegation
     , signDelegation
     , guardJoin
     , guardQuit
@@ -121,6 +123,10 @@ module Cardano.Wallet
     , ErrCannotQuit (..)
     , ErrSelectForDelegation (..)
     , ErrSignDelegation (..)
+
+    -- ** Fee Estimation
+    , FeeEstimation (..)
+    , estimateFeeForCoinSelection
 
     -- ** Transaction
     , forgetPendingTx
@@ -214,7 +220,11 @@ import Cardano.Wallet.Primitive.AddressDiscovery.Sequential
     , shrinkPool
     )
 import Cardano.Wallet.Primitive.CoinSelection
-    ( CoinSelection (..), CoinSelectionOptions (..), ErrCoinSelection (..) )
+    ( CoinSelection (..)
+    , CoinSelectionOptions (..)
+    , ErrCoinSelection (..)
+    , feeBalance
+    )
 import Cardano.Wallet.Primitive.CoinSelection.Migration
     ( depleteUTxO, idealBatchSize )
 import Cardano.Wallet.Primitive.Fee
@@ -292,7 +302,7 @@ import Cardano.Wallet.Transaction
 import Control.Exception
     ( Exception )
 import Control.Monad
-    ( forM, forM_, unless, when )
+    ( forM, forM_, replicateM, unless, when )
 import Control.Monad.IO.Class
     ( MonadIO, liftIO )
 import Control.Monad.Trans.Class
@@ -309,6 +319,8 @@ import Data.ByteString
     ( ByteString )
 import Data.Coerce
     ( coerce )
+import Data.Either
+    ( partitionEithers )
 import Data.Either.Extra
     ( eitherToMaybe )
 import Data.Foldable
@@ -347,6 +359,8 @@ import Numeric.Natural
     ( Natural )
 import Safe
     ( lastMay )
+import Statistics.Quantile
+    ( medianUnbiased, quantiles )
 
 import qualified Cardano.Wallet.Primitive.AddressDiscovery.Random as Rnd
 import qualified Cardano.Wallet.Primitive.CoinSelection.Random as CoinSelection
@@ -356,6 +370,7 @@ import qualified Data.List.NonEmpty as NE
 import qualified Data.Map.Strict as Map
 import qualified Data.Set as Set
 import qualified Data.Text as T
+import qualified Data.Vector as V
 
 -- $Development
 -- __Naming Conventions__
@@ -1068,6 +1083,20 @@ selectCoinsForDelegation ctx wid = do
     tl = ctx ^. transactionLayer @t @k
     tr = ctx ^. logger @WalletLog
 
+-- | Estimate fee for 'selectCoinsForDelegation'.
+estimateFeeForDelegation
+    :: forall ctx s t k.
+        ( HasTransactionLayer t k ctx
+        , HasLogger WalletLog ctx
+        , HasDBLayer s k ctx
+        )
+    => ctx
+    -> WalletId
+    -> ExceptT ErrSelectForDelegation IO FeeEstimation
+estimateFeeForDelegation ctx wid =
+    estimateFeeForCoinSelection $
+    selectCoinsForDelegation @_ @s @t @k ctx wid
+
 -- | Constructs a set of coin selections that select all funds from the given
 --   source wallet, returning them as change.
 --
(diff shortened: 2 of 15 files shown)


lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -498,8 +498,10 @@ newtype PostExternalTransactionData = PostExternalTransactionData
     { payload :: ByteString
     } deriving (Eq, Generic, Show)
 
-newtype ApiFee = ApiFee
-    { amount :: (Quantity "lovelace" Natural)
+data ApiFee = ApiFee
+    { amount :: !(Quantity "lovelace" Natural)
+    , estimatedMin :: !(Quantity "lovelace" Natural)
+    , estimatedMax :: !(Quantity "lovelace" Natural)
```
jonathanknowles, line 502: I guess the motivation for keeping `amount` around is backwards compatibility?

Perhaps an alternative option is to leave the existing API call(s) alone (adding a deprecated tag), and introduce a completely new API call that uses a new datatype:

```hs
data ApiFeeEstimate = ApiFeeEstimate
    { estimateMin :: !(Quantity "lovelace" Natural)
    , estimateMax :: !(Quantity "lovelace" Natural)
```

lib/core/test/unit/Cardano/WalletSpec.hs
```diff
@@ -488,6 +496,70 @@ walletListTransactionsSorted wallet@(wid, _, _) _order (_mstart, _mend) history
...
+-- 1. There is no coin selection with a fee above the estimated maximum.
+-- 2. The minimum estimated fee is no greater than the maximum estimated fee.
+-- 3. Around 10% of fees are below the estimated minimum.
+prop_estimateFee :: NonEmptyList (Either String FeeGen) -> Property
+prop_estimateFee (NonEmpty results) = case actual of
+    Left err -> label "errors: all" $
+        Left err === head results
+    Right (W.FeeEstimation minFee maxFee) ->
+        label ("errors: " <> if any isLeft results then "some" else "none") $
+        counterexample ("minFee=" <> show minFee <> " maxFee=" <> show maxFee) $
```
jonathanknowles:
```suggestion
    Right estimation@(W.FeeEstimation minFee maxFee) ->
        label ("errors: " <> if any isLeft results then "some" else "none") $
        counterexample (show estimation) $
```

(Since `W.FeeEstimation` already has a `Show` instance.)

lib/core/test/unit/Cardano/WalletSpec.hs
```diff
@@ -488,6 +496,70 @@ walletListTransactionsSorted wallet@(wid, _, _) _order (_mstart, _mend) history
...
+
+-- | Manufacture a coin selection that would result in the given fee.
+coinSelectionForFee :: FeeGen -> CoinSelection
+coinSelectionForFee (FeeGen (Coin fee)) = CoinSelection
+    { inputs = [(TxIn (Hash "") 0, TxOut (Address "") (Coin (1 + fee)))]
+    , outputs = [TxOut (Address "") (Coin 1)]
+    , change = []
+    }
+
+
```
jonathanknowles:
```suggestion

```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1109,6 +1137,59 @@ selectCoinsForMigration ctx wid = do
...
+    , estMaxFee :: Word64
+    -- ^ Most coin selections will result in a fee lower than this.
+    } deriving (Show, Eq)
+
+estimateFeeForCoinSelection
+    :: forall m err. Monad m
+    => ExceptT err m CoinSelection
+    -> ExceptT err m FeeEstimation
+estimateFeeForCoinSelection =
+    fmap deciles . handleErrors . replicateM repeats . runExceptT . fmap feeBalance
```
```suggestion
estimateFeeForCoinSelection
    = fmap deciles
    . handleErrors
    . replicateM repeats
    . runExceptT
    . fmap feeBalance
```

jonathanknowles (commented):
Hi Rodney

I've not looked into the statistics yet, but this looks very reasonable.

I do have a question regarding backwards compatibility.

The approach taken (so far) is to add a pair of fields to the existing `ApiFee` type:
```diff
- newtype ApiFee = ApiFee
-     { amount :: !(Quantity "lovelace" Natural) }
+ data ApiFee = ApiFee
+     { amount :: !(Quantity "lovelace" Natural)
+     , estimatedMin :: !(Quantity "lovelace" Natural)
+     , estimatedMax :: !(Quantity "lovelace" Natural)
+    }
```
Values of this type are returned by the following endpoint:
```
POST /wallets/{id}/payment-fees
```

As an alternative to modifying this endpoint, have you considered introducing a new endpoint and new type? Something like:
```
POST /wallets/{id}/payment-fee-estimates
```
With the following type:
```hs
data ApiFeeEstimate = ApiFeeEstimate
    { estimatedMin :: !(Quantity "lovelace" Natural)
    , estimatedMax :: !(Quantity "lovelace" Natural)
   }
```

lib/byron/test/integration/Test/Integration/Byron/Scenario/CLI/Transactions.hs
```diff
@@ -582,7 +582,7 @@ scenario_TRANS_ESTIMATE_01_02 fixtureSource fixtures = it title $ \ctx -> do
             , nChanges = length fixtures
             }
     verify r
-        [ expectCliField #amount $ between (Quantity feeMin, Quantity feeMax) ]
+        [ expectCliField #estimatedMin $ between (Quantity feeMin, Quantity feeMax) ]
```
jonathanknowles:
```suggestion
        [ expectCliField #estimatedMin $
            between (Quantity feeMin, Quantity feeMax)
        ]
```

(https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#limit-line-length-to-80-characters)

lib/byron/test/integration/Test/Integration/Byron/Scenario/CLI/Transactions.hs
```diff
@@ -751,7 +751,7 @@ scenario_TRANS_ESTIMATE_04b = it title $ \ctx -> do
             , nChanges = 0
             }
     verify r
-        [ expectCliField #amount $ between (Quantity feeMin, Quantity feeMax) ]
+        [ expectCliField #estimatedMin $ between (Quantity feeMin, Quantity feeMax) ]
```
jonathanknowles:
```suggestion
        [ expectCliField #estimatedMin $
            between (Quantity feeMin, Quantity feeMax)
        ]
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1586,6 +1631,56 @@ quitStakePool ctx wid argGenChange pwd = db & \DBLayer{..} -> do
...
+
+    -- Remove failed coin selections from samples. Unless they all failed, in
+    -- which case pass on the error.
+    handleErrors :: m [Either err a] -> ExceptT err m [a]
+    handleErrors = ExceptT . fmap skipFailed
+      where
+        skipFailed samples = case partitionEithers samples of
+            ([], []) -> error "estimateFeeForCoinSelection: impossible empty list"
+            ((e:_), []) -> Left e
+            (_, samples') -> Right samples'
```
```suggestion
            ([], []) ->
                error "estimateFeeForCoinSelection: impossible empty list"
            ((e:_), []) ->
                Left e
            (_, samples') ->
                Right samples'
```

https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#limit-line-length-to-80-characters

lib/core/test/unit/Cardano/WalletSpec.hs
```diff
@@ -489,6 +497,68 @@ walletListTransactionsSorted wallet@(wid, _, _) _order (_mstart, _mend) history
...
+-- 2. The minimum estimated fee is no greater than the maximum estimated fee.
+-- 3. Around 10% of fees are below the estimated minimum.
+prop_estimateFee :: NonEmptyList (Either String FeeGen) -> Property
+prop_estimateFee (NonEmpty results) = case actual of
+    Left err -> label "errors: all" $
+        Left err === head results
+    Right estimation@(W.FeeEstimation minFee maxFee) ->
+        label ("errors: " <> if any isLeft results then "some" else "none") $
+        counterexample (show estimation) $
+            maxFee <= maximum (map (getRight 0) results) .&&.
```
jonathanknowles, line 516: I could be wrong here, but shouldn't this be an equality test?

So:
```suggestion
            maxFee == maximum (map (getRight 0) results) .&&.
```

jonathanknowles (changes requested):
Hi @rvl 

Given the requirements as stated, this PR looks very sensible to me!

Ideally, we'd have some experimental data available to indicate the time and space costs of making multiple coin selections for different kinds of UTxO distributions. Unfortunately we don't have such data yet. (Though I'd very much like to create such a benchmark in future.)

Having said that, our current coin selection property tests do give us some (limited) information. With our current configuration, we're able to generate around 1,000 coin selections per second on our test machine. But this time includes the time taken to generate arbitrary UTxO sets. We currently generate UTxOs with between 1 and 256 entries, chosen according to a uniform random distribution. If we were to start with the same UTxO every time (to match what this PR is doing), this would almost certainly reduce the calculation time.

Assuming 100 samples from a similarly-sized UTxO set, the fee estimation endpoint could take around 1/10 second on a similar machine. So this seems reasonable as a starting point. :+1: 

If we wanted to err on the side of caution, then we _could_ watch the clock while taking samples, and stop prematurely if we've taken too long. (Though perhaps that could be a future PR.)

As for the rest of the PR: I only have a minor request, which is to fix a couple of formatting issues (see comments with suggested changes included).

Also, I have a question regarding the property test (relating to the fee maximum).

Apart from that, all looks good!

lib/core/test/unit/Cardano/WalletSpec.hs
rvl: It can't be an equality test because the real maximum fee might not have been sampled.

rvl:
We are going to change the estimate-fees api without making it backwards compatible.

lib/core/src/Cardano/Wallet/Api/Types.hs
jonathanknowles: Marking as resolved, as we decided to break backwards compatibility.

lib/core/test/unit/Cardano/WalletSpec.hs
jonathanknowles: > It can't be an equality test because the real maximum fee might not have been sampled.

Okay that makes sense. Thanks for clarifying!

lib/core/src/Cardano/Wallet/Api/Types.hs
jonathanknowles: > Perhaps an alternative option is to leave the existing API call(s) alone (adding a deprecated tag), and introduce a completely new API call that uses a new datatype:

Marking resolved, as we decided to break backwards compatibility.

jonathanknowles (approved):

rvl:
Thanks - rebased and squashed.

bors merge

KtorZ:
bors rety

KtorZ:
bors retry

PR #1656, opened 2020-05-12
Remove xPrvFromStrippedPubXPrvCheckRoundtrip and related

Anviking:
# Issue Number

#1637, relates to PR #1640 

# Overview

- [x] `xPrvFromStrippedPubXPrvCheckRoundtrip` and related functions no longer serve any purpose and can be removed.

# Comments

- They were introduced do cope with the underlying function silently modifying the keys. `xprvFromBytes` in cardano-addresses will not modify the keys.

### DIFF
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -114,7 +114,14 @@ import Prelude hiding
     ( getLine )
 
 import Cardano.Address.Derivation
-    ( XPrv, XPub, toXPub, xpubToBytes )
+    ( XPrv
+    , XPub
+    , toXPub
+    , xprvFromBytes
+    , xprvToBytes
+    , xpubFromBytes
+    , xpubToBytes
+    )
 import Cardano.BM.Backend.Switchboard
     ( Switchboard )
 import Cardano.BM.Configuration.Static
@@ -173,8 +180,6 @@ import Cardano.Wallet.Network
 import Cardano.Wallet.Primitive.AddressDerivation
     ( Depth (..)
     , DerivationType (..)
-    , ErrUnXPrvStripPub (..)
-    , ErrXPrvFromStrippedPubXPrv (..)
     , Index (..)
     , Passphrase (..)
     , PassphraseMaxLength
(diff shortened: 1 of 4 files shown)


Anviking:
bors try

Anviking:
bors try

lib/cli/src/Cardano/CLI.hs
```diff
@@ -451,29 +454,37 @@ fullKeyEncodingDescription = mconcat
...
     ]
 
-xprvFromBytes :: ByteString -> Either String XPrvOrXPub
-xprvFromBytes = fmap AXPrv . left showErr . xPrvFromStrippedPubXPrvCheckRoundtrip
+aXPrvFromBytes :: ByteString -> Either String XPrvOrXPub
+aXPrvFromBytes x = maybe errLength (return . AXPrv) $ xprvFromBytes x
   where
-    showErr (ErrInputLengthMismatch expected actual) = mconcat
+    actual = BS.length x
+    expectedBytes = 96 :: Int
```
Anviking, line 462: Should probably ideally be defined in `cardano-addresses` though

paweljakubas: yes, `96` maybe should be explicitly exposed here : https://github.com/input-output-hk/cardano-addresses/blob/master/src/Cardano/Address/Derivation.hs#L166

lib/cli/src/Cardano/CLI.hs
```diff
@@ -451,29 +454,37 @@ fullKeyEncodingDescription = mconcat
...
-        , "Is it encrypted? Or is it an old Byron key?"
-        ]
 
-xpubFromBytes :: ByteString -> Either String XPrvOrXPub
-xpubFromBytes = fmap AXPub . CC.xpub
+aXPubFromBytes :: ByteString -> Either String XPrvOrXPub
+aXPubFromBytes x = maybe errLength (return . AXPub) $ xpubFromBytes x
+  where
+    actual = BS.length x
+    expectedBytes = 64 :: Int
```
paweljakubas, line 476: `64` as the length of public key with chain code probably also.
the same for `32` as public key without chain code.

paweljakubas (approved):
Nice catch and refactoring giving rise to 150-line code cutting :)

lib/cli/src/Cardano/CLI.hs
Anviking: Added a todo in the code

lib/cli/src/Cardano/CLI.hs
Anviking: > 32 as public key without chain code

Not sure we ever deal with those?

Issue #1644, opened 2020-05-08
Integration BYRON_ADDRESS tests fail intermittently


piotr-iohk:
# Context
 
https://buildkite.com/input-output-hk/cardano-wallet/builds/7409#d6d784e5-27e0-47ab-98ee-196d96c12868/6-7586

# Test Case

Those seems to be failing randomly:
 - ADDRESS_CREATE_04, ADDRESS_CREATE_06
 - CLI_ADDRESS_CREATE_04, CLI_ADDRESS_CREATE_06

**On [buildkite](https://buildkite.com/input-output-hk/cardano-wallet/builds/7409#d6d784e5-27e0-47ab-98ee-196d96c12868/6-7586):**
```

Failures:
--
  |  
  | test/integration/Test/Integration/Byron/Scenario/API/Addresses.hs:205:30:
  | 1) API Specifications, BYRON_ADDRESSES, ADDRESS_CREATE_04 - Can list address after creating it
  | uncaught exception: IOException of type UserError
  | user error (expectListField: trying to access the #0 element from a list but there's none! )
  |  
  | To rerun use: --match "/API Specifications/BYRON_ADDRESSES/ADDRESS_CREATE_04 - Can list address after creating it/"
  |  
  | test/integration/Test/Integration/Byron/Scenario/API/Addresses.hs:257:11:
  | 2) API Specifications, BYRON_ADDRESSES, ADDRESS_CREATE_06 - Cannot create an address that already exists
  | expected: Status {statusCode = 409, statusMessage = "Conflict"}
  | but got: Status {statusCode = 201, statusMessage = "Created"}

```

**Locally:**

```
stack test cardano-wallet-byron:integration --test-arguments "--match ADDRESS"
```

Most of the time it passes, but every once in a while I was getting those errors as above or from the equivalent CLI tests:

```
  test/integration/Test/Integration/Byron/Scenario/CLI/Addresses.hs:239:30: 
  1) CLI Specifications, BYRON_CLI_ADDRESSES, CLI_ADDRESS_CREATE_04 - Can list address after creating it
       uncaught exception: IOException of type UserError
       user error (expectCliListField: trying to access the #0 element from a list but there's none! )


  test/integration/Test/Integration/Byron/Scenario/CLI/Addresses.hs:290:5: 
  1) CLI Specifications, BYRON_CLI_ADDRESSES, CLI_ADDRESS_CREATE_06 - Cannot create an address that already exists
[...]

PR #1663, opened 2020-05-14
Time and space optimization of address persistence

KtorZ:
# Issue Number

N/A (was working on debugging #1644 and noticed this so I gave it a shot).

# Overview

- 2fa27f78213361d9054675467f06af4a641192e5
  :round_pushpin: **time and space optimization of address persistence**
    We currently re-write the entire address state for each checkpoint, whereas the address state is by essence, an ever growing list of addresses. Unlike UTxOs, used addresses aren't disappearing from the state, so an easy win to avoid inserting the same addresses over and over is to actually make a diff and only store the diff, if necessary.

  Analyzing our 2 cases:

  ##### Sequential addresses

  Since addresses are derived sequentially, it suffices to look at the higher index stored in the database, and only store addresses with indexes beyond that.

  Benchmark                     | Before  | After
  ----------------------------- | ------- | -------
  SeqState/10  CP x 10    addr  | 14.86ms | 10.94ms
  SeqState/10  CP x 10000 addr  | 2.019s  | 485.5ms
  SeqState/100 CP x 10    addr  | 78.60ms | 55.99ms
  SeqState/100 CP x 10000 addr  | 21.53s  | 3.814s

  ##### Random addresses

  These are a bit more complicated. To make a proper diff, we need to first lookup all known indexes from the database, and remove them from the in-memory map representing currently known addresses. And, store the diff. It isn't necessarily faster to select all indexes, make a diff in memory, and store the result, but it isn't significantly slower either. Random wallets are already deprecated so this is simply another good reason to stop using them. 

  Incidentally, this approach still optimizes the space needed to store these addresses, so it's still a win!

  Benchmark                     | Before  | After
  ----------------------------- | ------- | -------
  RndState/10  CP x 10    addr  | 6.812ms | 8.987ms
  RndState/10  CP x 10000 addr  | 754.2ms | 819.9ms
  RndState/100 CP x 10    addr  | 44.57ms | 44.35ms
  RndState/100 CP x 10000 addr  | 7.529s  | 8.281s

  Now, this introduces a fundamental change in how we manage this address state. Before, we would have a set of rows for each slot at which a checkpoint was created; These rows would be ALL the addresses known at this particular slot. Now, we only store the newly discovered addresses at that slot!

  ```
        s0 s1 s2 s3 s4                     s0 s1 s2 s3 s4
       ┌──┬──┬──┬──┬──┐                   ┌──┬──┬──┬──┬──┐
       │a0│a0│a0│a0│a0│    __   _____     │a0│  │  │  │  │
       │  │a1│a1│a1│a1│    \ \ / / __|    │  │a1│  │  │  │
       │  │  │  │a2│a2│     \ V /\__ \    │  │  │  │a2│  │
       │  │  │  │a3│a3│      \_/ |___/    │  │  │  │a3│  │
       │  │  │  │  │a4│                   │  │  │  │  │a4│
       └──┴──┴──┴──┴──┘                   └──┴──┴──┴──┴──┘
  ```

  This requires thereby some changes:

  a. Remove the foreign key to the checkpoint table with cascade delete
  b. Remove address state manually when relevant (rollbacks)
  c. Do not prune the address state, it is pruned by construction
  d. To reconstruct the state, select not only addresses from a slot S
     but also all the addresses from previous slots.

- a8faeb13aaa99a301eb820b0d32ffc150d5bf79f
  :round_pushpin: **Resize QSM checkpoints generator in terms of the number of steps**
    The main effect of this change is to make sure that checkpoints
  generated in by the QSM tests have increasing slot numbers, and
  that they also generate address states that are bigger and bigger.

  Before this change, we would see commands such as:

  - putCheckpoint (RndState [addr1, addr2, addr3])
  - putCheckpoint (RndState [])

  Which is basically not possible in practice; addresses don't disappear. Another approach could have been to adjust a bit our database model to keep track of address state in a sparse way as
  well. Ideally, we should also decrease the "size" factor with rollbacks although, to do it well, we need to have a coupling between the generator and the transition function advancing the model (to make sure to shrink the size by the right factor).

# Comments

### DIFF
--- a/lib/byron/bench/Cardano/Wallet/Primitive/AddressDiscovery/Any.hs
+++ b/lib/byron/bench/Cardano/Wallet/Primitive/AddressDiscovery/Any.hs
@@ -43,7 +43,7 @@ import Data.Text
 import Data.Word
     ( Word32 )
 import Database.Persist.Sql
-    ( entityVal, insert_, selectFirst, (==.) )
+    ( deleteWhere, entityVal, insert_, selectFirst, (==.), (>.) )
 import GHC.Generics
     ( Generic )
 
@@ -99,6 +99,10 @@ instance PersistState AnyAddressState where
             , DB.AnyAddressStateCheckpointSlot ==. sl
             ] []
         return (AnyAddressState s)
+    deleteState (wid, sl) = deleteWhere
+        [ DB.AnyAddressStateWalletId ==. wid
+        , DB.AnyAddressStateCheckpointSlot >. sl
+        ]
 
 initAnyState :: Text -> Double -> (WalletId, WalletName, AnyAddressState)
 initAnyState wname p = (walletId cfg, WalletName wname, cfg)
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
@@ -132,6 +132,8 @@ import Data.Proxy
     ( Proxy (..) )
 import Data.Quantity
     ( Quantity (..) )
+import Data.Set
+    ( Set )
 import Data.Text.Class
     ( ToText (..), fromText )
 import Data.Typeable
@@ -180,6 +182,7 @@ import qualified Cardano.Wallet.Primitive.AddressDiscovery.Sequential as Seq
 import qualified Cardano.Wallet.Primitive.Model as W
 import qualified Cardano.Wallet.Primitive.Types as W
 import qualified Data.Map as Map
+import qualified Data.Set as Set
 import qualified Data.Text as T
 import qualified Database.Sqlite as Sqlite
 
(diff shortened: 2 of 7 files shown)


KtorZ:
Note: this probably deserve some attention regarding pre-existing database. I assume we can simply get away by "nubbing" duplicates that would be selected :thinking:

rvl:
The idea of writing diffs to the database sounds good. It's more suited to the wallet than "put" operations a la mongodb. I also had a branch which changed the return type of `applyBlocks` to be a checkpoint diff.

KtorZ:
> I also had a branch which changed the return type of applyBlocks to be a checkpoint diff.

Yes, I recall you groomed on that a while ago about the UTxO. It sounded more difficult there though? Do you still have that somewhere?

paweljakubas (approved):
looks reasonable!

PR #1665, opened 2020-05-14
Pipeline requests near the chain tip too, with care

### LINKED ISSUE #1650: Restoration benchmark on mainnet for 1% and 2% ownership unexpectedly slow
# Context

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | `2020.5.6 (git revision: 655a45952f3bbfa862140d12843e6d87a3577a6e)` |
| Platform     | All?          |
| Installation | From source          |
 
Looking at recent restoration benchmark results after introducing pipelining (https://github.com/input-output-hk/cardano-wallet/pull/1615):

 - TESTNET ([see chart](http://cardano-wallet-benchmarks.herokuapp.com/testnet-restoration)). Last 3 builds show the restoration times being 2 or 3 times faster for all 3 benchmarks :tada: 

Screenshot from 2020-05-11 13-55-40

 - MAINNET ([see chart](http://cardano-wallet-benchmarks.herokuapp.com/mainnet-restoration)). Shows improvement for the first benchmark (`restore mainnet seq`), but it shows that restoration for 1% and 2% as being almost two times slower...

Screenshot from 2020-05-11 13-55-21

:wave: :arrow_down: 
Look also at the memory profiles and blockchain height vs. time plot for before and after pipelining, interesing how they differ:

|Before pilelining| After pipelining  |
|--|--|
| [508](http://cardano-wallet-benchmarks.herokuapp.com/nightbuilds/508) |   [511](http://cardano-wallet-benchmarks.herokuapp.com/nightbuilds/511) |

# Steps to Reproduce

Run restoration benchmarks.

## Expected behavior

Expected would be to observe improvement in restoration times on all benchmarks.

## Actual behavior
 
Slower restoration for 1% and 2% ownership for mainnet.

---

# Resolution 

- Avoid creating too many checkpoints near the chain tip and either batch or pipeline there. See #1665

---

# QA

KtorZ:
# Issue Number

#1650 

# Overview

- 0a90f237a76d5895e0623e9a3f1a57b724512b13
  :round_pushpin: **Pipeline requests near the chain tip too, with care**
    This is a bit tricky because, as soon as we enter the `tip - k` area,
  any 'RequestNext' may yield a 'RollBackward'. The strategy is however
  quite simple and boils down to two cases:

  - Either we have to rollback within blocks we have just collected but
    not yet sent for processing.

  - Either, we have to rollback beyond this.

  In the former case, we can simply discard blocks from what we have
  collected and continue as normal. In the latter, we have to forward
  the rollback requests to the wallet engine, but we have to keep on
  collecting the results for the requests we have already pipelined!

  We do so by stashing the results in a queue, so that they'd be
  available for the next 'RequestNext' command from the wallet.

  This effectively extend pipelining up to the point where we have
  caught up with the tip and start collecting requests one by one, in a
  calmly manner. This is visible from the logs:

  ```
  ...
  Applying blocks [192.13732 ... 192.14731]
  Applying blocks [192.14732 ... 192.15731]
  Applying blocks [192.15732 ... 192.15977]
  Applying blocks [192.15978 ... 192.15978]
  Applying blocks [192.15979 ... 192.15979]
  Applying blocks [192.15980 ... 192.15980]
  ```

# Comments

This should fix #1650

### DIFF
--- a/lib/byron/src/Cardano/Wallet/Byron/Network.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron/Network.hs
@@ -328,7 +328,7 @@ mkWalletClient
         -- ^ Communication channel with the ChainSync client
     -> m (NetworkClient m)
 mkWalletClient bp chainSyncQ = do
-    stash <- atomically newTQueue
+    responsesBuffer <- atomically newTQueue
     pure $ nodeToClientProtocols NodeToClientProtocols
         { localChainSyncProtocol =
             let
@@ -339,7 +339,7 @@ mkWalletClient bp chainSyncQ = do
             InitiatorProtocolOnly $ MuxPeerRaw
                 $ \channel -> runPipelinedPeer nullTracer codec channel
                 $ chainSyncClientPeerPipelined
-                $ chainSyncWithBlocks fromTip' chainSyncQ stash
+                $ chainSyncWithBlocks fromTip' chainSyncQ responsesBuffer
 
         , localTxSubmissionProtocol =
             doNothingProtocol
--- a/lib/byron/src/Cardano/Wallet/Byron/Network.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron/Network.hs
@@ -251,9 +251,9 @@ withNetworkLayer tr gbp addrInfo versionData action = do
 
     _initCursor headers = do
         chainSyncQ <- atomically newTQueue
-        let client = const $ mkWalletClient bp chainSyncQ
+        client <- mkWalletClient bp chainSyncQ
         let handlers = failOnConnectionLost tr
-        link =<< async (connectClient tr handlers client versionData addrInfo)
+        link =<< async (connectClient tr handlers (const client) versionData addrInfo)
         let points = reverse $ genesisPoint : (toPoint getGenesisBlockHash getEpochLength <$> headers)
         let policy = constantDelay 500
         let findIt = chainSyncQ `send` CmdFindIntersection points
@@ -326,9 +326,10 @@ mkWalletClient
         -- ^ Static blockchain parameters
     -> TQueue m (ChainSyncCmd ByronBlock m)
         -- ^ Communication channel with the ChainSync client
-    -> NetworkClient m
-mkWalletClient bp chainSyncQ =
-    nodeToClientProtocols NodeToClientProtocols
+    -> m (NetworkClient m)
+mkWalletClient bp chainSyncQ = do
+    stash <- atomically newTQueue
+    pure $ nodeToClientProtocols NodeToClientProtocols
         { localChainSyncProtocol =
             let
                 fromTip' =
@@ -338,7 +339,7 @@ mkWalletClient bp chainSyncQ =
             InitiatorProtocolOnly $ MuxPeerRaw
                 $ \channel -> runPipelinedPeer nullTracer codec channel
                 $ chainSyncClientPeerPipelined
-                $ chainSyncWithBlocks getEpochStability fromTip' chainSyncQ
+                $ chainSyncWithBlocks fromTip' chainSyncQ stash
 
         , localTxSubmissionProtocol =
             doNothingProtocol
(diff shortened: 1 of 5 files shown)


lib/byron/1-percent.timelog
```diff
@@ -0,0 +1,4320 @@
+1589384183.268922344s -0
+1589384183.34260801s 1000
+1589384183.401558201s 2000
+1589384183.462808633s 3000
```
Anviking: Shouldn't be committed!

piotr-iohk: That's for boosting GH statistics ;P

Screenshot from 2020-05-14 16-53-04

KtorZ: I might have abused `git add lib` :smile: ...

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -340,24 +355,61 @@ chainSyncWithBlocks (Quantity epochStability) fromTip queue =
...
+        -- a) Either we rollback to a point we have just collected, so it
+        -- suffices to discard blocks from the list and continue.
+        --
+        -- b) Or, we need to reply immediately, but we still have to collect the
+        -- remaining responses. BUT, we can only reply once to a given command.
+        -- So instead, we stash all the remaining responses in a queue and, upon
+        -- receiving future requests, we'll simply read them from the stash!
+        , P.recvMsgRollBackward = \point _tip ->
+            case rollback point blocks of
+                [] -> do -- b)
```
Anviking, line 403: It probably doesn't matter, but I think you can't differentiate between:
- Rolling back to the latest persisted block (`[]`, but no `respond (Rollback point)` needed)
- Rolling back to one before the latest persisted block (rollback needed)

Anviking (approved):
lgtm.
I guess testing that everything works if the chain forks in the middle of collecting responses, etc, would be nice, but not sure how one would do that.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -326,9 +326,10 @@ mkWalletClient
         -- ^ Static blockchain parameters
     -> TQueue m (ChainSyncCmd ByronBlock m)
         -- ^ Communication channel with the ChainSync client
-    -> NetworkClient m
-mkWalletClient bp chainSyncQ =
-    nodeToClientProtocols NodeToClientProtocols
+    -> m (NetworkClient m)
+mkWalletClient bp chainSyncQ = do
+    stash <- atomically newTQueue
```
rvl: This could maybe be called "responseBuffer"

rvl (approved):
Looks good to me.

Would the design be simpler if we changed network layer from pull (nextBlocks) to push (onNextBlock)?

It would also be good to allocate some time to unit testing the cardano-node network layers because they are not totally simple any more.

KtorZ:
> Would the design be simpler if we changed network layer from pull (nextBlocks) to push (onNextBlock)?

Yes, having the buffering done in the network clients is awkward and it seems more and more suggesting that a push approach would be better suited. Johannes also suggested that a few times; I am still worried about the amount of work this would require on both cardano-node + jormungandr :/

KtorZ:
> I guess testing that everything works if the chain forks in the middle of collecting responses, etc, would be nice, but not sure how one would do that.
> It would also be good to allocate some time to unit testing the cardano-node network layers because they are not totally simple any more.

Yes, in particular, the behavior on roll backs is non trivial :(

lib/core/src/Ouroboros/Network/Client/Wallet.hs
KtorZ: Yes, this would induce a rollback to where the wallet already is at, which is a no-op for the wallet and database. So, in principle, there's nothing wrong from it. Yet indeed, the client is blind to what the latest persisted block is so it can't really make this distinction.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
KtorZ: done in: 0446953c8f6a888a6af5d4d19895d8b918a5884f :+1:

KtorZ:
Weird, the build succeeded but somehow bors never reported to github.

https://buildkite.com/input-output-hk/cardano-wallet/builds/7548

Issue #1672, opened 2020-05-18
(ADP-319) Add transaction support to cardano-wallet-shelley


Anviking:
# Context

ADP-319

# Decision

Make sure `cardano-wallet-shelley` supports making transactions by implementing `TransactionLayer` and using the cardano-address-Shelley-implementation.

# Acceptance Criteria

The following existing integration test scenarios should be passing:
- Wallets
- <s>ByronWallets</s>
- Transactions
- <s>ByronTransactions</s>

---

# Development 

- #1673 We added transaction support in one huge PR.

# QA 

- Byron addresses are not yet supported, and we cannot test anything with them.
- Wallet- and Transaction- integration tests are running otherwise.

PR #1673, opened 2020-05-18
Add Shelley `TransactionLayer`

Anviking:
# Issue Number

#1672 

# Overview

- [x] Added `toCardano` conversion functions interacting with `cardano-api`
- [x] Fixed mistakes in existing conversion functions
- [x] Implemented Shelley `TransactionLayer` and enabled `Wallet` and `Transaction` integration tests.
- [x] Property test making sure estimateSize never underestimates
# Comments

Main commit: 4f6ea17


lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -74,6 +105,64 @@ newTransactionLayer _proxy _protocolMagic = TransactionLayer
...
+                - sum (map (getCoin . coin) outs)
+        let certs = []
+
+        -- TODO: 1. Get the current epoch length
+        -- TODO: 2. The epoch length can change in the era change between byron and
+        -- shelley? It would probably be better to retrive a @SlotNo@ directly
+        -- without converting from @SlotId@.
+        let epochLength = EpochLength 1215
+
+        let Cardano.TxUnsignedShelley unsigned = Cardano.buildShelleyTransaction
```
Anviking: Would be nicer to use [`signTransaction`](https://github.com/input-output-hk/cardano-node/blob/c9ce637a260f415c8af21d75ddc6a1bb2efac23b/cardano-api/src/Cardano/Api.hs#L482-L496).

But I thought the way cardano-crypto carries an _encrypted_ XPrv and a password to the very end was intentional for… some security reasons? To construct `SignKeyDSIGN` we'd need to provide the unencrypted key.

KtorZ: >  was intentional for… some security reasons

Partly. Having key encrypted in-memory was an explicit request from the previous development team which was several times depicted as useless by cryptographers. What matters is storing them encrypted (cf the wallet-chain-libs rust implementation). We still use cardano-crypto for lack of alternative.

Anviking:
We could split this if desired, but I think it should be manageable.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Shelley.hs
```diff
@@ -305,20 +305,20 @@ instance DelegationAddress ('Testnet pm) ShelleyKey where
...
         Address $ BL.toStrict $ runPut $ do
             putWord8 (base + networkId)
-            putByteString . blake2b224 $ paymentK
-            putByteString . blake2b224 $ stakingK
+            putByteString . blake2b256 $ paymentK
+            putByteString . blake2b256 $ stakingK
       where
         base = 0
-        networkId = 0
+        networkId = 1 -- NOTE: Should be 0. Hack to support FF.
```
KtorZ: site-note: this should be `FIXME`, not `NOTE`. FIXME is used for code that will require change otherwise the software won't work. NOTE is for giving precision. (and TODO is more of a nice-to-have, but the software will work without).

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -15,92 +14,166 @@
...
+        -> Either ErrMkTx (Tx, SealedTx)
+    _mkStdTx keyFrom slot ownedIns outs = do
+        let ins' = map (toCardanoTxIn . fst) ownedIns
+        let outs' = map toCardanoTxOut outs
+        let fee = Coin $ sum (map (getCoin . coin . snd) ownedIns)
+                - sum (map (getCoin . coin) outs)
+        let certs = []
+
+        -- TODO: The SlotId-SlotNo conversion based on epoch length would not
+        -- work if the epoch length changed in a hard fork.
```
KtorZ, line 119: Although, the transaction itself would be invalid before the hard-fork. True that, subsequent hardfork after shelley may be a risk but for `Byron` -> `Shelley`, there's no problem here if we know the new epoch length prior to the hardfork.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -15,92 +14,166 @@
...
+        let timeToLive = (toSlotNo epochLength slot) + 7200
```
KtorZ, line 124: 7200 ? Avoid magic numbers, make it a named constant and explain the rationale of choosing `7200` (ultimately, we want that configurable via the API). Can you make sure to let Alex know about this?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -15,92 +14,166 @@
...
+        lookupPrivateKey addr =
+            maybe (Left $ ErrKeyNotFoundForAddress addr) Right (keyFrom addr)
+
+        signBody :: ByteString -> (k 'AddressK XPrv, Passphrase "encryption") -> SL.WitVKey TPraosStandardCrypto
+        signBody body k =
+            let
+                sig = SignedDSIGN $ fromMaybe (error "error converting signatures")
+                        $ rawDeserialiseSigDSIGN
+                        $ k `sign` body
+                vKey = SL.VKey $ VerKeyEd25519DSIGN $  throwCryptoError $ Ed25519.publicKey $ BS.take 32 $ CC.unXPub $ CC.toXPub $ getRawKey $ fst k
```
KtorZ: `BS.take 32 $ CC.unXPub $ CC.toXPub`

Use : `xpubPublicKey` instead from `Cardano.Address.Derivation`

lib/shelley/src/Cardano/Wallet/Shelley/Transaction/Size.hs
```diff
@@ -39,26 +37,21 @@ class MinSizeOf (t :: *) (n :: NetworkDiscriminant) (k :: Depth -> * -> *) where
...
+  where
+    -- According to the ledger specs, the 40 and 65 should be /the/ values, but
+    -- this integration test failure suggests otherwise:
+    -- ApplyTxError [LedgerFailure (UtxowFailure (UtxoFailure (FeeTooSmallUTxO
+    -- {pfUTXOminFee = Coin 553001000, pfUTXOgivenFee = Coin 520001000})))]\"}")
+    --
+    -- Maybe that value wasn't taking delegation keys into account?
+    --
+    -- Regardless:
+    -- TODO: Implement properly.
```
KtorZ: The length of the output will actually depend on the value and the type of addresses used. We can assume worst-case scenario which are currently 65 bytes for addresses indeed. 32 bytes for just the value sounds huge.

KtorZ (approved):
Nice work. I think we can go with this first big chunk and do improvements / additions on top.

KtorZ:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
Anviking:
```
era  Byron | Shelley
el       3 | 2
epoch   0 1 2 3
=========================================
slot 0 |0 3 6 8
  V  1 |1 4 7 9
     2 |2 5
```

If we are here given a correct `SlotId 3 1` at `SlotNo 9` (which arguably would be unlikely, since there's a similar conversion to create the SlotId)

The we cannot simply convert back to `SlotNo` using the genesis epoch length (here 3):

```
(SlotNo 3*3+1) == (SlotNo 10)
But the correct answer is: SlotNo 9
```

Anviking:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -131,10 +132,10 @@ newTransactionLayer _proxy _protocolMagic epochLength = TransactionLayer
         :: CoinSelection
         -> Quantity "byte" Int
     _estimateSize (CoinSelection inps outs chngs) = Quantity $
-        sizeOfSignedTx (fst <$> inps) (outs <> map dummyOutput chngs)
+        sizeOfSignedTx inps (outs <> map dummyOutput chngs)
       where
         dummyOutput :: Coin -> TxOut
-        dummyOutput = TxOut $ Address "\130\ACKX \217\NULYA\205=\179d\160d\203\232\SYN\138C\SOH\\MGO\141\201\228\161\138\146\136\179\235M\217v"
+        dummyOutput = TxOut $ Address $ BS.replicate 65 0
```
KtorZ: 65 should be `maxSizeOf @Address`

Anviking:
bors try

KtorZ:
I force pushed the "regenerate nix" commit like a retard. :|

PR #1679, opened 2020-05-21
Reduce running time of fee estimation endpoint

rvl:
### Issue Number

Relates to ADP-283 / #1648.

### Overview

The nightly api endpoint latency benchmark got 100× slower for `postTransactionFee`. This fixes it by moving common setup for coin selection outside of the loop.

### Comments

```
    Latencies for 10 fixture wallets scenario
        postTransactionFee - before - 102.2 ms
        postTransactionFee  - after - 4.8 ms
```


jonathanknowles (approved):
If I've understood correctly, the general pattern seems to be:

1. Fetch the wallet's UTxO set and also the current fee policy.
2. Run a coin selection on the UTxO set.

Given that the estimation endpoints need to run the second phase multiple times, it seems very sensible to have this separation. :+1:

Issue #1649, opened 2020-05-11
DB migration tests failing intermittently


piotr-iohk:
# Context
 
DB migration test seem to be failing intermittently. In last 3 nightly builds (all of them being run on the same revision https://github.com/input-output-hk/cardano-wallet/commit/655a45952f3bbfa862140d12843e6d87a3577a6e) they failed twice, i.e.:

[NB 509](https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/509) - passed (:heavy_check_mark:)
[NB 510](https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/510#60af6023-b991-4dcb-8715-572668288d9f) - failed (:x:)
[NB 511](https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/511#e16f28bd-fab4-4d1b-a3a4-51ad7737a67f) - failed (:x:)

# Test Case

DB migration test

# Failure / Counter-example

**NB 510**
https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/510#60af6023-b991-4dcb-8715-572668288d9f

Seems that Wallet stopped working (?) after migrating from `v2020-04-28` => `v2020-05-06`
```

[migration-test:Notice:4] [2020-05-09 01:13:33.74 UTC] Starting process cardano-wallet-jormungandr
--
  | launch
  | --state-dir ./state-migration-test-v2020-04-28
  | --genesis-block /nix/store/8b6in2dygz6hk29p9zpyxzhrs17cb3ni-cardano-wallet-src/lib/jormungandr/test/data/jormungandr/block0.bin
  | --
  | --secret /nix/store/8b6in2dygz6hk29p9zpyxzhrs17cb3ni-cardano-wallet-src/lib/jormungandr/test/data/jormungandr/secret.yaml
  | --config /nix/store/8b6in2dygz6hk29p9zpyxzhrs17cb3ni-cardano-wallet-src/lib/jormungandr/test/data/jormungandr/config.yaml
  |  
  | [migration-test:Info:4] [2020-05-09 01:13:33.74 UTC] [cardano-wallet-jormungandr.128152] Process started
  | [migration-test:Debug:13] [2020-05-09 01:13:33.74 UTC] [cardano-wallet-jormungandr.128152] About to waitForProcess
  | [migration-test:Debug:12] [2020-05-09 01:13:33.74 UTC] [cardano-wallet-jormungandr.128152] Running withBackend action
[...]

PR #1680, opened 2020-05-21
Fix port selection in the db migration tests

rvl:
### Issue number

Relates to #1649

### Overview

Each migration test running in turn was using the same port.

However due to the `SO_LINGER` TCP socket option, the port remains in use for a short time after the wallet server process exits.

So use a different port for each test.

### Comments

Tested with:

    nix-build -A migration-tests -o migration-tests && ./migration-tests/runall.sh

### DIFF
--- a/lib/jormungandr/test/migration/migration-test.hs
+++ b/lib/jormungandr/test/migration/migration-test.hs
@@ -65,7 +65,7 @@ import Data.Aeson
 import Data.Aeson.Lens
     ( key, values, _Integral, _String )
 import Data.Maybe
-    ( isNothing )
+    ( fromMaybe, isNothing )
 import Data.Text
     ( Text )
 import Network.Wreq
@@ -74,6 +74,8 @@ import System.Environment
     ( getArgs )
 import System.Exit
     ( ExitCode (..), exitFailure, exitWith )
+import Text.Read
+    ( readMaybe )
 
 import qualified Data.Text as T
 
@@ -83,7 +85,7 @@ main = do
 
     cfg <- defaultConfigStdout
     withTrace cfg "migration-test" $ \tr ->
-        testMain @('Testnet 0) tr 9090 testAction launchArgs >>= exitWith
+        testMain @('Testnet 0) tr testAction launchArgs >>= exitWith
 
 -- | Something to do while the server is running.
 type TestAction (t :: NetworkDiscriminant) = Trace IO Text -> ApiBase -> IO ExitCode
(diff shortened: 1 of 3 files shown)


nix/launch-migration-test.sh
```diff
@@ -30,7 +30,14 @@ if [ -z "${configFile:-}" ]; then
   exit 1
 fi
 
+# Find an unused TCP port to run the test on.
+# Source: https://unix.stackexchange.com/a/132524
+find_unused_port() {
+  python -c 'import socket; s=socket.socket(); s.bind(("", 0)); print(s.getsockname()[1]); s.close()'
+}
```
jonathanknowles, line 37: :+1:

nix/migration-tests.nix
```diff
@@ -112,6 +112,7 @@ let
           migrationTest
           pkgs.bash
           pkgs.coreutils
+          pkgs.python3
```
jonathanknowles, line 115: Will pulling this in lead to an increase in build times for CI? (Or is everything cached?)

lib/jormungandr/test/migration/migration-test.hs
```diff
@@ -113,6 +115,15 @@ testMain tr serverPort testAction launchArgs = do
             logError tr ("Process exited with status " <> T.pack (show st))
             pure (ExitFailure 14)
 
+-- | Scan through cardano-wallet server CLI arguments to find a port.
+-- Returns the default 8090 if none found.
+getServerPort :: [String] -> Int
+getServerPort = fromMaybe 8090 . portArg
```
jonathanknowles, line 121: Hi @rvl

Here the default seems to have changed from `9090` to `8090`.

Is that your intention?

jonathanknowles (commented):
Hi @rvl 

This looks good to me. Though I have a couple of questions (see comments).

nix/migration-tests.nix
rvl: It's all cached, and python3 is already used by openapi-spec-validator anyway.

lib/jormungandr/test/migration/migration-test.hs
rvl: It's a bit of a magic assumption that the cardano-wallet server default is 8090 when the `--port` arg it missing (which it isn't). When we port the db migration test to byron I have a plan to fix this properly.

nix/migration-tests.nix
jonathanknowles: Okay, thanks for confirming!

jonathanknowles (approved):
Looks good to me!

Issue #1658, opened 2020-05-13
(ADP-257) Change ByronMigration endpoints


paweljakubas:
# Context

As an API Client,
I want to be able to empty a wallet and send all funds to a list of addresses,
So that I can easily move funds around and have the wallet dealing with fees for me.

# Decision

We first add new migration endpoint that enables to transfer a given wallet to the addresses specified. The following cases are to be supported : 
(a) byron wallet -> shelley addresses
(b) byron wallet -> byron addresses
Moreover, MigrationWalletInfo is to be operative 

This is jormungandr context!

# Acceptance Criteria

1. New endpoints in ByronWallet are changed in all needed places in codebase (swagger, API endpoint).
2. Needed unit tests for types are added.
3. Implementation is added and integration tests work

---

# Development 

# QA 

1. We changed how Byron migration endpoints work. Now they accept source wallet id and reciepent addresses. This gave rise to many changes in our codebase, from swagger up to Wallet core-integration tests. Would be worth at this changes
2. Added integration tests for checking address cycling https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs#L1344
3. added a number of property tests for `assignMigrationAddresses` https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelectionSpec.hs#L92-L101

Anviking:
What is the reason to not also support "Shelley wallet -> Shelley addresses"?

paweljakubas:
Because in this task I am after refactoring ByronMigration. In the next task I will add address shelley -> shelley in a sense that the endpoints will be added to shelley transactions endpoint family

piotr-iohk:
Doing some exploratory testing and encountered this case on jormungandr. When trying to migrate to an address `DdzFFzCqrhtCNjPk5Lei7E1FxnoqMoAYtJ8VjAWbFmDb614nNBWBwv3kt6QHJa59cGezzf6piMWsbK7sWRB5sv325QqWdRuusMqqLdMt` I am getting `500`.

To reproduce:
1. Launch Jorm:
```
cd /lib/jormungandr/test/data/jormungandr/test_scripts
./wallet_launch
```
:point_down: 

2. Create byron wallet with funds:
```
arctic decade pink easy jar index base bright vast ocean hard pizza

[...]

PR #1682, opened 2020-05-22
Scenario to illustrate the migration 500 error on Jormungandr

piotr-iohk:
# Issue Number

https://github.com/input-output-hk/cardano-wallet/issues/1658

# Overview

- 22ddb84bd36c5ea9ecc2bdd36333a25e3d06acc5
  Scenario to illustrate the migration 500 error on Jormungandr

# Comments

This is regarding https://github.com/input-output-hk/cardano-wallet/issues/1658#issuecomment-632137152.

In the test I am using byron mainnet and byron testnet address and currently the response from the server is `500` (with Jormungandr), and that is what I left in the test for the time being. I think that it should be probably something more user firendly...

### DIFF
--- a/lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/Transactions.hs
+++ b/lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/Transactions.hs
@@ -54,6 +54,8 @@ import Cardano.Wallet.Primitive.Types
     )
 import Cardano.Wallet.Transaction
     ( TransactionLayer (..) )
+import Control.Monad
+    ( forM_ )
 import Data.ByteArray.Encoding
     ( Base (Base16, Base64), convertFromBase, convertToBase )
 import Data.Generics.Internal.VL.Lens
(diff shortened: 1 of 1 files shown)


lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/Transactions.hs
```diff
@@ -353,6 +356,24 @@ spec = do
...
+            sWallet <- fixtureRandomWallet ctx
+            r <- request @[ApiTransaction n] ctx
+                (Link.migrateWallet sWallet)
+                Default
+                (Json [json|
+                    { passphrase: #{fixturePassphrase}
+                    , addresses: [#{addr}]
+                    }|])
+            verify r
+                [ expectResponseCode @IO HTTP.status500
```
rvl: Could we expect `HTTP.status400` and set the test to pending?

rvl (approved):

lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/Transactions.hs
piotr-iohk: Ok, corrected.

Issue #1675, opened 2020-05-20
(ADP-257) Add Shelley migration endpoints and Byron (haskell impl) support of migrateWallet


paweljakubas:
# Context

As an API Client,
I want to be able to empty a wallet and send all funds to a list of addresses,
So that I can easily move funds around and have the wallet dealing with fees for me.

# Decision

We are adding migration endpoints as done in https://github.com/input-output-hk/cardano-wallet/issues/1658 in Shelley

# Acceptance Criteria

1. Shelley has migration sendAll endpoints and cost estimation endpoint for this.
2. We have Shelley migration post data unit tests.
3. Corresponding changes in swagger are done. 
4. Migration tests are added to cardano-node's byron integration tests.

Comment: as transaction are not available for shelley we cannot check it at this moment in integration test mode.

---

# Development 

# QA 

1. migrateWallet was enabled for byron (haskell impl) and several integration tests were added in lib/byron.
2. The subtle bug was found in Transaction layer of byron (haskell impl) and TransactionSpec was adjusted.
3. migrateWallet was enabled also for shelley (haskell impl), but at this moment it was not impossible to test it as this lib misses transaction layer impl (although it should go very soon as #1673 - so it would be worth testing this (maybe copy integration tests from byron lib concerning migration)

KtorZ:
When reviewing #1658 I noticed that the current migration endpoints are actually associated with a `/migrations` namespace / resource. So, perhaps it makes sense to keep the separation between transactions / migrations in our API types as well? @paweljakubas

piotr-iohk:
I have updated swagger spec -> #1687. 
It would be nice to also address this -> #1682.

piotr-iohk:
I'll leave it in QA for now because Shelley migrations are not quite testable currently without transactions.

piotr-iohk:
I encounter a `500` response when running migrations from **Byron** wallet on F&F testnet (for Shelley wallets migrations seem to work fine).

Steps to reproduce:
1. Run migration on Byron wallet having funds and as a result you'll have `500` response:

```
POST http://localhost:8090/v2/byron-wallets/94c0af1034914f4455b7eb795ebea74392deafe9/migrations

The response was:
Code = 500,
[...]

PR #1684, opened 2020-05-25
clarify migrate endpoint end-user documentation

KtorZ:
# Issue Number

#1675 

# Overview

The sense of 'migrate' seemed unclear to people. So this attempts to give more precisions about
what it really is about. The same description should be copied over for the Shelley equivalent.

# Comments

### DIFF
--- a/specifications/api/swagger.yaml
+++ b/specifications/api/swagger.yaml
@@ -2180,7 +2180,12 @@ paths:
       description: |
         <p align="right">status: <strong>stable</strong></p>
 
-        Migrate all funds from a Byron wallet to set of addresses.
+        Create one or several transactions sending all funds from a Byron wallet to set of addresses.
+        This operation tries to preserve as much as possible the current UTxO shape of a wallet (hence
+        why several transactions may be needed).
+
+        A typical usage would be when one wants to move all funds from an old wallet to another (Shelley
+        or Byron) by providing addresses coming from the new wallet.
       parameters:
         - <<: *parametersWalletId
           name: walletId


piotr-iohk (approved):
Question about the corresponding `cost` endpoint. The description for endpoint `calculate migration cost` reads:
```
Calculate the exact cost of migrating from a Byron wallet to the specified set of addresses.
```
Addresses are not provided as a parameter to the `cost` endpoint, as they are for the `migrate` endpoint. (Is the cost not influenced by the number of addresses?)

KtorZ:
> Is the cost not influenced by the number of addresses?

Hmmm. No. The algorithm can be seen as two steps: 

a) Construct coin selections needed to migrate funds. 
b) Assign an address to each change output, cycling them if necessary.

So, the number of addresses is known, from the migration itself. It is true that in principle, the address could influence the selection, because their _individual size_ is not known and, is not fixed in Byron. The fee estimation is using the worst-case scenario (i.e. random addresses with derivation payload) to do calculation, but Shelley addresses are much shorter. This could be a potential improvement of the algorithm / fee estimator used by the algorithm.

piotr-iohk:
OK, but then saying: 

_Calculate the exact cost of migrating from a Byron wallet **to the specified set of addresses.**_ 

sounds  a bit awkward when you don't specify any addresses.

paweljakubas:
@piotr-iohk you need to specify at least one address -> https://github.com/input-output-hk/cardano-wallet/blob/master/specifications/api/swagger.yaml#L416 don't you :-)

paweljakubas (approved):

piotr-iohk:
@paweljakubas I'm referring to `calculate migration cost` endpoint -> https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/getByronWalletMigrationInfo

You don't specify addresses on that endpoint but the description reads:

_Calculate the exact cost of migrating from a Byron wallet to the **specified** set of addresses._

paweljakubas:
ah, ok. my bad. in the context of getByronWalletMigrationInfo the "to the specified set of addresses" is not needed, indeed! Agree here. It is the migration cost of whole funds in the wallet specified by wallet id.

Issue #1670, opened 2020-05-18
Different behavior regarding resolved inputs when listing transactions with and without query parameters


KtorZ:
# Context
 
 

Reported by an exchange on Slack.

> __Quote:__
> 
> "We need to check input, output values in a transaction.  However, if we use `start`, `end` parameters in `/transactions` api, input values **don't appear in the response**. If we fetch all the transactions without `start` and `end` parameters, **they appears**. But we don't want to fetch all the transactions belonging the wallet neither due to the performance concerns."

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | Unknown  |
| Platform     | Unknown           |
| Installation | Unknown             |

# Steps to Reproduce

see #1685 

## Expected behavior

Responses are coherent with and without `start` / `end` 

## Actual behavior
 

Responses are different.

---

# Resolution 

- Resolve transaction inputs in the database lookup, when fetching inputs and outputs ==> #1685 

---

# QA 

- Added a regression test to capture the scenario described below in comment as `TRANS_REG_1670`

KtorZ:
Going quickly through the code, one of the reason might be:

```hs
    -- This relies on DB.readTxHistory returning all necessary transactions
    -- to assemble coin selection information for outgoing payments.
    assemble
```

So, we the history returned is used both as the result, but also as an index for resolving transaction inputs. Therefore, if a transaction input refers to a known transaction that is discarded by the range parameters, then it won't be present in the return history and will fail the lookup. 

Making `assemble` fetch transactions from the database would probably fix the issue, but at what cost? This would require at most 1 request per input (so possibly up to 80/90 requests per transactions) although it'd still be better than fetching the entire history in memory every time. 

Another alternative could be to list transactions as already resolved from the database with an inner join; haven't quite thought about it but it seems doable from the top of my head and should not induce that much overhead. 

cc @piotr-iohk : if this intuition is correct :point_up: , we should be able to reproduce this by making two transactions, the second one using at least one output of the first one, and then, by listing all transactions  and listing all transaction with a range only including the second transaction.

psychoplasma:
Here is an example case from my test wallet

`cardano-wallet transaction list <wallet_id> --order ascending --start 2020-05-14T05:53:31Z`

```
Ok.
[
    {
        "inserted_at": {
            "time": "2020-05-14T05:53:31Z",
            "block": {
                "height": {
                    "quantity": 4159520,
                    "unit": "block"
                },
[...]

PR #1685, opened 2020-05-25
Fix Input resolution for inputs only present in db

KtorZ:
# Issue Number

#1670

# Overview

- 1502136da13c0f9e6ba47b74c12c31912099da62
  :round_pushpin: **add regression test illustrating the issue**
  Input that should be resolved aren't. In order to resolve inputs, we are actually wrongly using
the transactions available from the history that was fetched from the database, instead of using
the entire database! So, giving different query parameters restricting the set of outputs has an
influence on something it shouldn't.

- 8ec012359e5faaeeec9f73f255184575b6b5d0c1
  :round_pushpin: **resolve transaction inputs from db when fetching the history**
  Before, this was done in the application layer, from the available history ('assemble' in 'Cardano.Wallet'). This requires exactly two extra lookups. One in the TxOut table, and one in the Checkpoint table so the performance impact should be quite small. It simplifies the application logic quite a bit and doesn't much impact the database logic, so it's almost a win-win. Plus, it fixes the issue where resolvable inputs would appear as non resolved

- 64bf4db742dc029ea3a18b1c4e957e22f612ab47
  :round_pushpin: **fix DB model to also return full 'TransactionInfo' when fetching the tx history**
  
- 21d75d20ae9b11707d16a4115159ef617e6bcdb0
  :round_pushpin: **Fix db benchmark build**
  Benchmarks may produce higher timings after this change, but it's OK
since the answers are more correct now, and the work has just shifted
from the wallet layer into the db layer.

- fffe476a58673fcb23a3c43136d8eded2d46f999
  :round_pushpin: **re-add lost comment about input resolution for incoming/outgoing transactions**
  
- 67adc07a71a3e8c3e5430654b0845173e4d189e2
  :round_pushpin: **rename 'txs' -> 'pending' to better match context**

# Comments

### DIFF
--- a/lib/byron/test/integration/Test/Integration/Byron/Scenario/API/Transactions.hs
+++ b/lib/byron/test/integration/Test/Integration/Byron/Scenario/API/Transactions.hs
@@ -26,6 +26,7 @@ import Cardano.Wallet.Api.Types
     , ApiWalletMigrationInfo
     , DecodeAddress (..)
     , EncodeAddress (..)
+    , Iso8601Time (..)
     , WalletStyle (..)
     )
 import Cardano.Wallet.Primitive.AddressDerivation
@@ -40,6 +41,8 @@ import Control.Monad
     ( forM, forM_ )
 import Data.Generics.Internal.VL.Lens
     ( view, (^.) )
+import Data.Maybe
+    ( fromJust, isJust )
 import Data.Quantity
     ( Quantity (..) )
 import Data.Text
@@ -161,6 +164,8 @@ spec = do
         scenario_TRANS_ESTIMATE_04a @n
         scenario_TRANS_ESTIMATE_04b @n
         scenario_TRANS_ESTIMATE_04c @n
+
+        scenario_TRANS_REG_1670 @n (fixtureIcarusWalletWith @n)
 
     describe "BYRON_RESTORATION" $ do
         scenario_RESTORE_01 @n fixtureRandomWallet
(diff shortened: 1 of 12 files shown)


lib/byron/test/integration/Test/Integration/Byron/Scenario/API/Transactions.hs
```diff
@@ -690,6 +696,71 @@ scenario_TRANS_UTXO_01 fixtureSource fixtureTarget = it title $ \ctx -> do
...
 
+scenario_TRANS_REG_1670
+    :: forall (n :: NetworkDiscriminant) t.
+        ( DecodeAddress n
+        , EncodeAddress n
+        , PaymentAddress n IcarusKey
+        )
+    => (Context t -> [Natural] -> IO ApiByronWallet)
+    -> SpecWith (Context t)
+scenario_TRANS_REG_1670 fixture = it title $ \ctx -> do
```
KtorZ, line 800: This fails before the fix.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1493,45 +1493,6 @@ listTransactions ctx wid mStart mEnd order = db & \DBLayer{..} -> do
         _ ->
             pure $ slotRangeFromTimeRange sp $ Range mStart mEnd
 
-    -- This relies on DB.readTxHistory returning all necessary transactions
-    -- to assemble coin selection information for outgoing payments.
-    -- To reliably provide this information, it should be looked up when
-    -- applying blocks, but that is future work (issue #573).
-    assemble
```
KtorZ, line 1500: no longer needed, the db now returns `[TransactionInfo]` directly.

lib/byron/test/integration/Test/Integration/Byron/Scenario/API/Transactions.hs
rvl: Since the tx input resolving is moved into the DBLayer, this test could probably be more easily done at the unit level.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1016,15 +1033,35 @@ selectUTxO cp = fmap entityVal <$>
...
   where
     select txids = do
-        ins <- fmap entityVal <$> selectList [TxInputTxId <-. txids]
+        inputs <- fmap entityVal <$> selectList
+            [TxInputTxId <-. txids]
             [Asc TxInputTxId, Asc TxInputOrder]
-        outs <- fmap entityVal <$> selectList [TxOutputTxId <-. txids]
+
+        resolvedInputs <- toOutputMap . fmap entityVal <$> selectList
+            [TxOutputTxId <-. (txInputSourceTxId <$> inputs)]
```
rvl, line 1054: If only persistent could support SQL joins (without raw sql).

lib/core/src/Cardano/Wallet.hs
```diff
@@ -599,7 +599,7 @@ readWallet ctx wid = db & \DBLayer{..} -> mapExceptT atomically $ do
     cp <- withNoSuchWallet wid $ readCheckpoint pk
     meta <- withNoSuchWallet wid $ readWalletMeta pk
     txs <- lift $ readTxHistory pk Descending wholeRange (Just Pending)
-    pure (cp, meta, Set.fromList (fst <$> txs))
+    pure (cp, meta, Set.fromList (fromTransactionInfo <$> txs))
```
rvl: Is it possible to rename `txs` here to `pending`?

lib/core/src/Cardano/Wallet.hs
rvl: Is it still noted somewhere that coin selection information can't (yet) be provided for outgoing payments?

rvl (approved):
LGTM

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
KtorZ: If only yeah...

lib/core/src/Cardano/Wallet.hs
KtorZ: I've rewritten the comment and linked to the github issue in the relevant place in the db layer :+1: , good point

KtorZ:
Whoops. I failed to push because there was a change upstream I didn't see. Rebasing and pushing.

PR #1687, opened 2020-05-26
Display Shelley migrations in Api doc

piotr-iohk:
# Issue Number

#1675

# Overview

- 186726655c405149af9b3903dbf61c38b64be92e
  Display Shelley migrations in swagger spec.
  
- 508e41b4d94be8861ab9c8e407a1584975ad6dfa
  Reword migration description according to review suggestions

# Comments

Shelley migrations, although included in `swagger.yaml`, were not being displayed in the [API doc](https://input-output-hk.github.io/cardano-wallet/api/edge/). This pr fixes this. Also includes minor rewording of the migration endpoints descriptions.

Screenshot from 2020-05-26 16-38-13

### DIFF
--- a/specifications/api/swagger.yaml
+++ b/specifications/api/swagger.yaml
@@ -1987,10 +1987,12 @@ paths:
       description: |
         <p align="right">status: <strong>in development</strong></p>
 
-        Migrate all funds from a Shelley wallet to a specified set of addresses.
-        Create one or several transactions sending all funds from a Shelley wallet to set of addresses.
-        This operation tries to preserve as much as possible the current UTxO shape of a wallet (hence
-        why several transactions may be needed).
+        Submit one or more transactions which transfers all funds from a Shelley
+        wallet to a set of addresses.
+
+        This operation attempts to preserve the UTxO "shape" of a wallet as far as possible.
+        That is, coins will not be agglomerated. Therefore, if the wallet has
+        a large UTxO set, several transactions may be needed.
 
         A typical usage would be when one wants to move all funds from an old wallet to another
         by providing addresses coming from the new wallet.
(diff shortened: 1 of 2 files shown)


specifications/api/swagger.yaml
```diff
@@ -1660,6 +1660,7 @@ x-tagGroups:
     - Addresses
     - Coin Selections
     - Transactions
+    - Migrations
```
piotr-iohk, line 1663: Shelley migrations were not displayed in api doc because of this :point_up_2:

specifications/api/swagger.yaml
```diff
@@ -1967,26 +1968,32 @@ paths:
...
+      tags: ["Migrations"]
       summary: Migrate
       description: |
-        <p align="right">status: <strong>stable</strong></p>
+        <p align="right">status: <strong>in development</strong></p>
 
         Migrate all funds from a Shelley wallet to a specified set of addresses.
+        Create one or several transactions sending all funds from a Shelley wallet to set of addresses.
+        This operation tries to preserve as much as possible the current UTxO shape of a wallet (hence
+        why several transactions may be needed).
```
paweljakubas: I wonder whether word `why` is needed here :thinking:  @rvl / @jonathanknowles  could you voice your opinion?

paweljakubas (approved):

specifications/api/swagger.yaml
rvl: How about this

> Submit one or more transactions which transfers all funds from a Shelley wallet to a set of addresses.
> This operation attempts to preserve the UTxO "shape" of a wallet as far as possible. That is, coins will not be agglomerated.
> Therefore, if the wallet has a large UTxO set, several transactions may be needed.

piotr-iohk: Reworded in 508e41b4d94be8861ab9c8e407a1584975ad6dfa.

PR #1691, opened 2020-05-27
Fix for release.nix sourcesOverride arg

hamishmack:


rvl:
Couldn't niv be used for this? For example:

```
niv drop iohk-nix
niv add local --name iohk-nix --path ../iohk-nix
```

Issue #1690, opened 2020-05-27
Remove `epochId` argument from `getNetworkParameters`.


jonathanknowles:
# Context

The presence of the `epochId` path parameter in the `/network/parameters/<epochId>` endpoint might give callers the _impression_ that calling `GET` on the endpoint will retrieve parameters corresponding to a particular epoch.

In reality, however, this endpoint only ever returns parameters corresponding to the _latest_ epoch, _regardless_ of which epoch is specified.

# Decision

Remove the `epochId` argument from this endpoint.

# Acceptance Criteria

- [x] The `epochId` argument _must_ be removed from the `getNetworkParameters` endpoint.
- [x] The `getNetworkParameters` endpoint _must_ always return parameters that correspond to the _latest_ epoch.
- [x] We _must_ provide assistance to projects that consume `cardano-wallet`, to help them transition to the modified endpoint. (See https://input-output-rnd.slack.com/archives/C7PJ29FME/p1590569319044000)

# Development 

Remove the following functions and types:

- [x] `ApiEpochNumber`
- [x] `errMsg404NoEpochNo`
- [x] `getNetworkParamsViaCliExpectingFailure`

In addition:

- [x] Adjust description of CLI `network parameters` command to read "View network parameters for the current epoch".

# QA 

- [x] Adjust existing API and CLI tests. We no longer need to test the `epochId` parameter as it no longer exists, but we should retain basic tests to check that the API function and CLI command both still work.

rvl:
Just noting here that the only parameter in the current ApiNetworkParameters that will ever update in shelley is activeSlotCoeff. The rest come from the genesis block.

jonathanknowles:
@rvl wrote:

> Just noting here that the only parameter in the current ApiNetworkParameters that will ever update in shelley is activeSlotCoeff. The rest come from the genesis block.

That's a great point. Perhaps we could break this type up into two separate types: one for static parameters, and another for parameters that can be updated.

I've started a discussion thread about this here: https://input-output-rnd.slack.com/archives/GBT05825V/p1590642003077400

piotr-iohk:
lgtm.

Issue #1692, opened 2020-05-27
Add skeletal `decentralizationLevel` field to `ApiNetworkParameters`.


jonathanknowles:
# Context

We need to provide a way for API users to query the current level of decentralization in the Cardano network, represented by the `d` network parameter.

# Decision

Add a field `decentralizationLevel` (of type `Quantity "percent" Percentage`) to the `ApiNetworkParameters` record type, and make corresponding adjustments to the swagger API definition.

For _this_ task, we will arrange that the value of this field is always set to **0%**. _Future_ tasks will cause this field to be set to the true value reported by the network.

# Acceptance Criteria

- [x] The `ApiNetworkParameters` type _must_ have a `decentralizationLevel` field, and the swagger API definition _must_ have a corresponding field that matches.
- [x] The `decentralizationLevel` field _should_ have type `Quantity "percent" Percentage` (already defined elsewhere), where **0%** indicates that the network is completely federalized, and **100%** indicates that the network is completely decentralized.
- [x] The `decentralizationLevel` field _may_ be assigned a constant value of **0%**. (Reporting the _true_ network value is _not_ part of the acceptance criteria of this task.)

# Development 

- [x] Add skeletal `decentralizationLevel` field to `ApiNetworkParameters`.

# QA 

- [x] Write an integration test to check that the API really does return a value of 0% for the decentralization level. In future, we can adjust our test to expect the live value.

piotr-iohk:
lgtm.

PR #1694, opened 2020-05-27
Remove `epochId` argument from `getNetworkParameters`.

jonathanknowles:
# Issue Number

#1690 

# Overview

This PR:

- [x] Removes the `epochId` argument from the `getNetworkParameters` API function.
- [x] Removes the `epoch` argument from the `network parameters` CLI command.
- [x] Removes the now unused `ApiEpochNumber` type. (We can reinstate it later, if necessary.)
- [x] Removes the now unused `ErrNoSuchEpoch` type. (We can reinstate it later, if necessary.)
- [x] Updates the swagger specification.
- [x] Updates the CLI integration tests.
- [x] Updates the API integration tests.

# Comments

**:warning: This PR breaks API compatibility with older versions! :warning:** 

- [x] Before merging this PR, we should notify projects that consume `cardano-wallet`, and help them transition to the modified endpoint. (See https://input-output-rnd.slack.com/archives/C7PJ29FME/p1590569319044000)

### DIFF
--- a/lib/cli/test/unit/Cardano/CLISpec.hs
+++ b/lib/cli/test/unit/Cardano/CLISpec.hs
@@ -518,14 +518,13 @@ spec = do
             ]
 
         ["network", "parameters", "--help"] `shouldShowUsage`
-            [ "Usage:  network parameters [--port INT] EPOCH_NUMBER"
+            [ "Usage:  network parameters [--port INT]"
             , "  View network parameters."
             , ""
             , "Available options:"
             , "  -h,--help                Show this help text"
             , "  --port INT               port used for serving the wallet"
             , "                           API. (default: 8090)"
-            , "  EPOCH_NUMBER             epoch number parameter or 'latest'"
             ]
 
         ["network", "clock", "--help"] `shouldShowUsage`
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -147,7 +147,6 @@ import Cardano.Wallet.Api.Types
     , ApiCoinSelection (..)
     , ApiCoinSelectionInput (..)
     , ApiEpochInfo (..)
-    , ApiEpochNumber (..)
     , ApiErrorCode (..)
     , ApiFee (..)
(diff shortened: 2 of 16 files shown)


piotr-iohk (approved):
LGTM.

lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -518,14 +518,13 @@ spec = do
             ]
 
         ["network", "parameters", "--help"] `shouldShowUsage`
-            [ "Usage:  network parameters [--port INT] EPOCH_NUMBER"
+            [ "Usage:  network parameters [--port INT]"
```
rvl, line 521: + "View network parameters for the current epoch"

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1430,32 +1429,12 @@ getNetworkInformation (_block0, gbp, st) nl = do
...
-                liftHandler $ throwE $ ErrNoSuchEpoch
-                    { errGivenEpoch = epochNum
-                    , errCurrentEpoch = currentEpochNum
-                    }
-            pure (toApiNetworkParameters bp)
-
-        ApiEpochNumberLatest ->
-            pure (toApiNetworkParameters bp)
+getNetworkParameters (_block0, gbp, _st) =
+    pure $ toApiNetworkParameters bp
```
rvl, line 1434: Consider removing the where clause and putting it all on one line.

rvl (approved):
:+1:

lib/core/src/Cardano/Wallet/Api/Server.hs
jonathanknowles: Good point! I've added this to the list of things to do on the next PR: https://github.com/input-output-hk/cardano-wallet/issues/1692

lib/cli/test/unit/Cardano/CLISpec.hs
jonathanknowles: @rvl I've created a PR for this here: https://github.com/input-output-hk/cardano-wallet/pull/1698

PR #1696, opened 2020-05-27
sql: Fix input resolution when number of inputs is large

rvl:
### Issue Number

#1670 / #1685

### Overview

The "too many SQL variables" problem strikes again in the nightly benchmark.

One way to fix it is to do a SQL join or subquery.

Another way to fix it (this PR) is to make the select query in chunks.

### Comments

Tested with:

    cabal run cardano-wallet-core:bench:db  -- --match prefix "TxHistory (Read)/1000"

or

    stack bench cardano-wallet-core:bench:db --ba '--match prefix "TxHistory (Read)/1000"'


KtorZ (approved):
:+1:

PR #1697, opened 2020-05-27
actually identify output by their transaction AND output index...

KtorZ:
# Issue Number

#1670 

# Overview

The transaction id doesn't uniquely identify an output! So, if a transaction had several output, they'd all be conflated under the same key... :facepalm:

I wonder why the state machine didn't catch that though.. because the database model was doing it correctly by using the full `TxIn` as a key.

# Comments

### DIFF
--- a/lib/byron/test/integration/Test/Integration/Byron/Scenario/API/Transactions.hs
+++ b/lib/byron/test/integration/Test/Integration/Byron/Scenario/API/Transactions.hs
@@ -841,26 +841,47 @@ scenario_TRANS_REG_1670 fixture = it title $ \ctx -> do
         pure $ Iso8601Time $ maximum $ getTime <$> getFromResponse id rTxs
 
     -- ACTION
-    rTxs <- request @[ApiTransaction n] ctx
+    rTxsFromDate <- request @[ApiTransaction n] ctx
         (Link.listTransactions' @'Byron wSrc (Just start) Nothing Nothing)
         Default
         Empty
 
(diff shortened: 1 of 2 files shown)


rvl:
Hmm I will investigate the generators for the state machine tests.

piotr-iohk (approved):
LGTM.

Maybe simple integration test checking that sum of inputs >= sum of outputs of the sent tx could be also handy.

rvl (approved):

lib/byron/test/integration/Test/Integration/Byron/Scenario/API/Transactions.hs
```diff
@@ -847,6 +847,14 @@ scenario_TRANS_REG_1670 fixture = it title $ \ctx -> do
         Empty
 
     -- ASSERTIONS
+    let transaction = head $ getFromResponse id rTxs
```
KtorZ: @piotr-iohk There's no guarantee that the first tx is the one we are interested in here :thinking:
It'll be safer to simply run this check for every transaction present in the wallet, it should hold anyway.

piotr-iohk: Added verification for all outgoing transactions -> https://github.com/input-output-hk/cardano-wallet/pull/1697/commits/f8bcfe23df814fb3d6769e2e27984c26b523cba3. Incoming txs not necessarily have input amounts.

PR #1698, opened 2020-05-28
Clarify the description of `getNetworkParameters` in the API and CLI.

jonathanknowles:
# Related Issues

#1690

# Overview

This PR clarifies the descriptions of the `getNetworkParameters` API endpoint and the associated `network parameters` CLI command, in response to [review feedback](https://github.com/input-output-hk/cardano-wallet/pull/1694/files#r431011565) for #1694,


rvl (approved):

rvl (approved):
Actually ... technically those parameters which are returned are from genesis block because none of them are updatable. They will never change.

And also technically, suppose that there are updatable parameters in there. They will be for latest epoch which the node has synced, which is not necessarily the current epoch. Although I think it's an acceptable simplification to say "current epoch" in this context.

PR #1702, opened 2020-06-01
Remove types for the greater good: Value-level NetworkDiscriminant

Anviking:
# Overview

We use a type-level `NetworkDiscriminant` and type-classes for much logic. I believe the benefits are negligible, with greater-than-realised costs.

This PR makes `NetworkDiscriminant` used at the value-level. I hope this opens up for lots of additional future simplifications.

# Current state

We have code like this:
```haskell
data SomeNetworkDiscriminant where
    SomeNetworkDiscriminant
        :: forall (n :: NetworkDiscriminant).
            ( NetworkDiscriminantVal n
            , PaymentAddress n IcarusKey
            , PaymentAddress n ByronKey
            , DecodeAddress n
            , EncodeAddress n
            , MaxSizeOf Address n IcarusKey
            , MaxSizeOf Address n ByronKey
            )
        => Proxy n
        -> SomeNetworkDiscriminant
```

## Benefits with current approach

- With `EncodeAddress n` and `DecodeAddress n`, Servant can automatically convert between the textual (e.g. hex, bech32) encodings of addresses and the binary.

## Problems with current approach

- It can be complicated to work with type-level `NetworkDiscriminant` (calls for code like above), and compiler error messages can be frustrating.
- Since it is tricky to work with, we may end up not making the abstractions we "should" be making.
- Can be tricky to realise what effect `@n` has, since type-class instances are scattered everywhere.
- High-level functions often need many constraint-annotations.
    - IIRC, sometimes when changing a small detail, we have to update lots of constraints everywhere. This is annoying. 

# Overview

- [x] I removed the `MkKeyFingerprint`, `GenChange`, `PaymentAddress`, `DelegationAddress`, `NetworkDiscriminantVal` classes.
- [x] I added a `AddressScheme k` record.
- [x] I made the outside Servant-API use `Text` as type for Address, and use 
- [ ] **TODO** 🏗️ Handle Address decoding errors, and make all targets and tests compile and pass
    - [ ] Make core compile without TODOs

## Future work
- We could remove `KeyFingerprint` and `Address <-> Text` conversions by parameterising more over `address`.
- We could seek to reduce duplication in integration tests
    -  I think a record-layer-type for transaction capabilities could be great
- We could seek to reduce duplication in `Api.Server`
- We could allow target-specific `NetworkDiscriminants`
    - In shelley we could use it to allow the `networkMagic` from the genesis file to _actually be used_, instead of re-calculating it ourselves.

## Comments

- This PR has a fundamental preference for records over type-classes.

### DIFF
--- a/lib/byron/src/Cardano/Wallet/Byron.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron.hs
@@ -26,8 +26,7 @@
 -- "Cardano.Wallet.Byron.Transaction"
 
 module Cardano.Wallet.Byron
-    ( SomeNetworkDiscriminant (..)
-    , serveWallet
+    ( serveWallet
 
       -- * Tracing
     , Tracers' (..)
@@ -59,8 +58,11 @@ import Cardano.Wallet.Api
     ( ApiLayer, ApiV2 )
 import Cardano.Wallet.Api.Server
     ( HostPreference, Listen (..), ListenError (..), TlsConfiguration )
+<<<<<<< HEAD
 import Cardano.Wallet.Api.Types
     ( ApiStakePool, DecodeAddress, EncodeAddress, EncodeStakeAddress )
+=======
+>>>>>>> 59d9eb545... Refactor type-level NetworkDiscriminant
 import Cardano.Wallet.Byron.Api.Server
     ( server )
 import Cardano.Wallet.Byron.Compatibility
@@ -69,8 +71,6 @@ import Cardano.Wallet.Byron.Network
     ( NetworkLayerLog, withNetworkLayer )
 import Cardano.Wallet.Byron.Transaction
     ( newTransactionLayer )
-import Cardano.Wallet.Byron.Transaction.Size
-    ( MaxSizeOf )
 import Cardano.Wallet.DB.Sqlite
     ( DefaultFieldValues (..), PersistState )
 import Cardano.Wallet.Logging
@@ -78,18 +78,16 @@ import Cardano.Wallet.Logging
 import Cardano.Wallet.Network
     ( NetworkLayer (..) )
 import Cardano.Wallet.Primitive.AddressDerivation
-    ( Depth (..)
+    ( AddressScheme
+    , Depth (..)
     , NetworkDiscriminant (..)
-    , NetworkDiscriminantVal
-    , PaymentAddress
     , PersistPrivateKey
     , WalletKey
-    , networkDiscriminantVal
     )
 import Cardano.Wallet.Primitive.AddressDerivation.Byron
-    ( ByronKey )
+    ( ByronKey, byronScheme )
 import Cardano.Wallet.Primitive.AddressDerivation.Icarus
-    ( IcarusKey )
+    ( IcarusKey, icarusScheme )
 import Cardano.Wallet.Primitive.AddressDiscovery
     ( IsOurs )
 import Cardano.Wallet.Primitive.AddressDiscovery.Random
@@ -149,6 +147,7 @@ import qualified Cardano.Wallet.DB.Sqlite as Sqlite
 import qualified Data.Text as T
 import qualified Network.Wai.Handler.Warp as Warp
 
+<<<<<<< HEAD
 -- | Encapsulate a network discriminant and the necessary constraints it should
 -- satisfy.
 data SomeNetworkDiscriminant where
@@ -169,6 +168,8 @@ data SomeNetworkDiscriminant where
 
 deriving instance Show SomeNetworkDiscriminant
 
+=======
+>>>>>>> 59d9eb545... Refactor type-level NetworkDiscriminant
 -- | The @cardano-wallet-shelley@ main function. It takes the configuration
 -- which was passed from the CLI and environment and starts all components of
 -- the wallet.
@@ -176,7 +177,7 @@ serveWallet
     :: forall t.
         ( t ~ IO Byron
         )
-    => SomeNetworkDiscriminant
+    => NetworkDiscriminant
     -- ^ Proxy for the network discriminant
     -> Tracers IO
     -- ^ Logging config.
@@ -206,7 +207,7 @@ serveWallet
     -- ^ Callback to run before the main loop
     -> IO ExitCode
 serveWallet
-  (SomeNetworkDiscriminant proxy)
+  n
   Tracers{..}
   sTolerance
   databaseDir
@@ -217,9 +218,8 @@ serveWallet
   block0
   (np, vData)
   beforeMainLoop = do
-    let ntwrk = networkDiscriminantValFromProxy proxy
     traceWith applicationTracer $ MsgStarting socketPath
-    traceWith applicationTracer $ MsgNetworkName ntwrk
+    traceWith applicationTracer $ MsgNetworkName (toText n)
     Server.withListeningSocket hostPref listen $ \case
         Left e -> handleApiServerStartupError e
         Right (_, socket) -> serveApp socket
@@ -228,19 +228,13 @@ serveWallet
         withNetworkLayer networkTracer np socketPath vData $ \nl -> do
             withWalletNtpClient io ntpClientTracer $ \ntpClient -> do
                 let pm = fromNetworkMagic $ networkMagic $ fst vData
-                randomApi  <- apiLayer (newTransactionLayer proxy pm) nl
-                icarusApi  <- apiLayer (newTransactionLayer proxy pm) nl
-                startServer proxy socket randomApi icarusApi ntpClient
+                randomApi  <- apiLayer (newTransactionLayer n pm) nl (byronScheme n)
+                icarusApi  <- apiLayer (newTransactionLayer n pm) nl (icarusScheme n)
+                startServer socket randomApi icarusApi ntpClient
                 pure ExitSuccess
 
-    networkDiscriminantValFromProxy
-        :: forall n. (NetworkDiscriminantVal n)
-        => Proxy n
-        -> Text
-    networkDiscriminantValFromProxy _ =
-        networkDiscriminantVal @n
-
     startServer
+<<<<<<< HEAD
         :: forall n.
             ( PaymentAddress n IcarusKey
             , PaymentAddress n ByronKey
@@ -253,31 +247,43 @@ serveWallet
         -> Socket
         -> ApiLayer (RndState n) t ByronKey
         -> ApiLayer (SeqState n IcarusKey) t IcarusKey
+=======
+        :: Socket
+        -> ApiLayer RndState t ByronKey
+        -> ApiLayer (SeqState IcarusKey) t IcarusKey
+>>>>>>> 59d9eb545... Refactor type-level NetworkDiscriminant
         -> NtpClient
         -> IO ()
-    startServer _proxy socket random icarus ntp = do
+    startServer socket random icarus ntp = do
         sockAddr <- getSocketName socket
         let settings = Warp.defaultSettings & setBeforeMainLoop
                 (beforeMainLoop sockAddr)
+<<<<<<< HEAD
         let application = Server.serve (Proxy @(ApiV2 n ApiStakePool)) $
                 server random icarus ntp
+=======
+        let application = Server.serve (Proxy @ApiV2) $
+                server n random icarus ntp
+>>>>>>> 59d9eb545... Refactor type-level NetworkDiscriminant
         Server.start settings apiServerTracer tlsConfig socket application
 
     apiLayer
         :: forall s k.
             ( IsOurs s Address
             , IsOurs s ChimericAccount
-            , PersistState s
+            , PersistState s k
             , PersistPrivateKey (k 'RootK)
             , WalletKey k
             )
         => TransactionLayer t k
         -> NetworkLayer IO t ByronBlock
+        -> AddressScheme k
         -> IO (ApiLayer s t k)
-    apiLayer tl nl = do
+    apiLayer tl nl addrScheme = do
         let params = (block0, np, sTolerance)
         db <- Sqlite.newDBFactory
             walletDbTracer
+<<<<<<< HEAD
             (DefaultFieldValues
                 { defaultActiveSlotCoefficient =
                     getActiveSlotCoefficient gp
@@ -293,6 +299,12 @@ serveWallet
             databaseDir
         Server.newApiLayer walletEngineTracer params nl' tl db
             Server.idleWorker
+=======
+            (DefaultFieldValues $ getActiveSlotCoefficient gp)
+            addrScheme
+            databaseDir
+        Server.newApiLayer walletEngineTracer params nl' tl db addrScheme
+>>>>>>> 59d9eb545... Refactor type-level NetworkDiscriminant
       where
         gp = genesisParameters np
         nl' = fromByronBlock gp <$> nl
--- a/lib/byron/src/Cardano/Wallet/Byron/Api/Server.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron/Api/Server.hs
@@ -23,6 +23,7 @@ import Cardano.Wallet
     ( ErrCreateRandomAddress (..)
     , ErrNotASequentialWallet (..)
     , ErrValidateSelection
+    , addressScheme
     , genesisData
     , networkLayer
     )
@@ -81,7 +82,7 @@ import Cardano.Wallet.Api.Server
 import Cardano.Wallet.Api.Types
     ( ApiStakePool, ApiT (..), SomeByronWalletPostData (..) )
 import Cardano.Wallet.Primitive.AddressDerivation
-    ( PaymentAddress (..) )
+    ( NetworkDiscriminant )
 import Cardano.Wallet.Primitive.AddressDerivation.Byron
     ( ByronKey )
 import Cardano.Wallet.Primitive.AddressDerivation.Icarus
@@ -89,7 +90,7 @@ import Cardano.Wallet.Primitive.AddressDerivation.Icarus
 import Cardano.Wallet.Primitive.AddressDiscovery.Random
     ( RndState )
 import Cardano.Wallet.Primitive.AddressDiscovery.Sequential
-    ( SeqState )
+    ( SeqState, seqGenChange )
 import Control.Applicative
     ( liftA2 )
 import Control.Monad.Trans.Except
@@ -113,6 +114,7 @@ import Type.Reflection
 
 -- | A diminished servant server to serve Byron wallets only.
 server
+<<<<<<< HEAD
     :: forall t n.
         ( Buildable (ErrValidateSelection t)
         , PaymentAddress n IcarusKey
@@ -124,6 +126,15 @@ server
     -> NtpClient
     -> Server (Api n ApiStakePool)
 server byron icarus ntp =
+=======
+    :: forall t.  ( Buildable (ErrValidateSelection t))
+    => NetworkDiscriminant
+    -> ApiLayer RndState t ByronKey
+    -> ApiLayer (SeqState IcarusKey) t IcarusKey
+    -> NtpClient
+    -> Server Api
+server _n byron icarus ntp =
+>>>>>>> 59d9eb545... Refactor type-level NetworkDiscriminant
          wallets
     :<|> addresses
     :<|> coinSelections
@@ -148,13 +159,13 @@ server byron icarus ntp =
         :<|> (\_ _ -> throwError err501)
         :<|> (\_ -> throwError err501)
 
-    addresses :: Server (Addresses n)
+    addresses :: Server Addresses
     addresses _ _ = throwError err501
 
-    coinSelections :: Server (CoinSelections n)
+    coinSelections :: Server CoinSelections
     coinSelections _ _ = throwError err501
 
-    transactions :: Server (Transactions n)
+    transactions :: Server Transactions
     transactions =
              (\_ _ -> throwError err501)
         :<|> (\_ _ _ _ _ -> throwError err501)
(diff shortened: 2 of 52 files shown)


lib/byron/src/Cardano/Wallet/Byron/Api/Server.hs
```diff
@@ -179,20 +177,20 @@ server byron icarus ntp =
              )
         :<|> (\wid -> withLegacyLayer' wid
                 ( byron
-                , fst <$> getWallet byron  mkLegacyWallet wid
-                , const (fst <$> getWallet byron  mkLegacyWallet wid)
+                , fst <$> getWallet byron  (mkLegacyWallet @_ @RndState @ByronKey) wid
```
Anviking, line 230: Hopefully we can get rid of the need for type applications / make this nicer in the future.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1387,59 +1374,3 @@ taggedSumTypeOptions base opts = base
...
--- defining custom types in the API client with a minimal overhead and, without
--- having to actually rewrite any of the API definition.
---
--- We use an open type family so it can be extended by other module in places.
-type family ApiAddressT (n :: k) :: *
-type family ApiAddressIdT (n :: k) :: *
-type family ApiCoinSelectionT (n :: k) :: *
-type family ApiSelectCoinsDataT (n :: k) :: *
-type family ApiTransactionT (n :: k) :: *
-type family PostTransactionDataT (n :: k) :: *
```
Anviking, line 1539: All this can be removed now

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Shelley.hs
```diff
@@ -248,77 +245,40 @@ instance WalletKey ShelleyKey where
...
+    , addressToText = error "todo: shelley addressToText"
+    , addressFromText = error "todo: shelley addressFromText"
+
+    }
+  where
+    enterprise = 96
+    base = 0
+    networkId = case net of
+        Mainnet   -> 1
+        Testnet _ -> 1 -- FIXME: Should be 0. Hack to support FF testnet.
```
Anviking: I think this is much neater.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1175,11 +1180,11 @@ checkpointId cp = (checkpointWalletId cp, checkpointSlot cp)
...
 -- | Functions for saving/loading the wallet's address discovery state into
 -- SQLite.
-class PersistState s where
+class PersistState s k where
     -- | Store the state for a checkpoint.
-    insertState :: (W.WalletId, W.SlotId) -> s -> SqlPersistT IO ()
+    insertState :: (W.WalletId, W.SlotId) -> s -> AddressScheme k-> SqlPersistT IO ()
     -- | Load the state for a checkpoint.
-    selectState :: (W.WalletId, W.SlotId) -> SqlPersistT IO (Maybe s)
+    selectState :: (W.WalletId, W.SlotId) -> AddressScheme k -> SqlPersistT IO (Maybe s)
```
Anviking: Arguably looks a bit weird to take `AddressScheme k`, but I think it is fine, and can be improved upon in the future (…perhaps)

lib/byron/src/Cardano/Wallet/Byron.hs
```diff
@@ -224,57 +198,45 @@ serveWallet
...
-        => Proxy n
-        -> Text
-    networkDiscriminantValFromProxy _ =
-        networkDiscriminantVal @n
-
     startServer
-        :: forall n.
-            ( PaymentAddress n IcarusKey
-            , PaymentAddress n ByronKey
-            , DecodeAddress n
```
Anviking: We were forced to explicitly specify these constraints because of the `n` parameter.

I think they are implementation details, and being able to remove them here is a good thing.

lib/byron/src/Cardano/Wallet/Byron.hs
```diff
@@ -202,7 +177,7 @@ serveWallet
     -- ^ Callback to run before the main loop
     -> IO ExitCode
 serveWallet
-  (SomeNetworkDiscriminant proxy)
+  n
```
KtorZ, line 210: May I suggest a better / longer name than `n` ? This is used across a function that spread over a hundred lines. It deserves a longer name than `n` ;)

lib/byron/src/Cardano/Wallet/Byron/Api/Server.hs
```diff
@@ -218,16 +216,16 @@ server byron icarus ntp =
...
                  (byron , do
                     let pwd = coerce (getApiT $ tx ^. #passphrase)
-                    genChange <- rndStateChange byron wid pwd
-                    postTransaction byron genChange wid tx
+                    genChange <- rndStateChange @_ @RndState byron wid pwd
+                    postTransaction @_ @RndState byron genChange wid tx
                  )
                  (icarus, do
-                    let genChange k _ = paymentAddress @n k
+                    let genChange = seqGenChange (icarus ^. addressScheme)
```
KtorZ, line 288: This feels quite redundant / unnecessary now. If the address scheme is represented as a record, there's no need for type families and passing around of this `ArgGenChange`.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1475,7 +1477,8 @@ cmdAddressImport mkClient =
         <*> walletIdArgument
         <*> addressIdArgument
     exec (AddressImportArgs wPort wId addr) = do
-        runClient wPort (const "") $ putRandomAddress mkClient (ApiT wId) addr
+        runClient wPort (const "") $ putRandomAddress mkClient (ApiT wId)
+            (error "todo" $ addr)
```
KtorZ, line 942: :point_up:

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1306,14 +1308,14 @@ cmdTransactionFees mkTxClient mkWalletClient =
         <*> fmap NE.fromList (some paymentOption)
     exec (TransactionCreateArgs wPort wId wAddressAmounts) = do
         wPayments <- either (fail . getTextDecodingError) pure $
-            traverse (fromText @(AddressAmount Text)) wAddressAmounts
+            traverse (fromText @(AddressAmount ApiAddress)) wAddressAmounts
```
KtorZ, line 749: It now seems unnecessary for `AddressAmount` to have a type parameter. It was there before exactly to cope with the command-line.

lib/core-integration/src/Test/Integration/Faucet.hs
```diff
@@ -1353,10 +1354,11 @@ genByronFaucets = genFaucet encodeAddress genAddresses
             addrXPrv =
                 Byron.deriveAddressPrivateKey pwd accXPrv
         in
-            [ paymentAddress @'Mainnet
+            [ addressFromKey
                 $ publicKey $ addrXPrv $ liftIndex @'Hardened ix
             | ix <- [minBound..maxBound]
             ]
+    AddressScheme{addressFromKey} = Byron.byronScheme Mainnet
```
KtorZ, line 1434: What about keeping the same naming convention as before? `paymentAddress` captured it quite well and is also used in `cardano-addresses`. It'll make it easier to bridge between packages.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -462,6 +460,7 @@ type HasLogger msg = HasType (Tracer IO msg)
 -- | This module is only interested in one block-, and tx-type. This constraint
 -- hides that choice, for some ease of use.
 type HasNetworkLayer t = HasType (NetworkLayer IO t Block)
+type HasAddressScheme (k :: Depth -> * -> *) = HasType (AddressScheme k)
```
KtorZ, line 496: I am going to be "picky" here, but there's a newline between each of these declarations. Let's stick to the existing style when changing code :+1:

lib/core/src/Cardano/Wallet.hs
```diff
@@ -568,6 +573,12 @@ createIcarusWallet ctx wid wname credentials = db & \DBLayer{..} -> do
         initializeWallet pk (updateState s' cp) meta hist txp $> wid
   where
     db = ctx ^. dbLayer @s @k
+    addrScheme@AddressScheme{addressFromFingerprint} = ctx ^. addressScheme
+    -- We shouldn't have to add it to the AddressSchema. Instead:
+    -- 1. Use Icarus-specific function
+    -- 2. Allow the SeqState to shrink itself when applying blocks (drastic
```
KtorZ, line 614: NO! Certainly not! This shrinking above is really to cope with the "snapshot" approach of the ITN. Under no circumstances one would want to shrink an address pool when applying blocks.

Note however that the address fingerprint is needed for every address; this is what makes it possible to manipulate the "spending" part of an address, regardless of its delegation settings. Most addresses in the wallet are attached to the same stake key, but it's not a golden rule. One could have an two address with the same payment key and two different stake key, or one with no stake key at all. Still, this address is the same from the point of view of the sequential discovery! We aren't simply using a public key or a hash of a public key here because, not all addresses embeds these; in particular, Byron & Icarus addresses only have a hashed "root" which includes many things apart from the public key; but still, Icarus addresses are using sequential discovery. 

The address fingeprint captures this as it gives you the actual address identifier used for discovery. Perhaps the name isn't ideal, but that concept needs to be available for any address and isn't icarus-specific!

lib/core/src/Cardano/Wallet.hs
```diff
@@ -988,17 +996,11 @@ importRandomAddress ctx wid addr = db & \DBLayer{..} -> mapExceptT atomically $
...
-        )
-    => s
+    :: s
     -> Address
     -> Maybe Address
-normalizeDelegationAddress s addr = do
-    fingerprint <- eitherToMaybe (paymentKeyFingerprint addr)
-    pure $ liftDelegationAddress @n fingerprint (rewardAccount s)
+normalizeDelegationAddress _s _addr = error "todo: Can we remove this by putting \
+    \it directly into the target-specific listAddresses?"
```
KtorZ: Not really. Or not that easily. Because you need to apply this conditionally based on whether or not the given wallet support delegation. This isn't given by `s`, but is given by `k`. It can probably be included in the newly created `addressScheme` and be an identity function for legacy addresses.

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -178,19 +178,19 @@ import Servant.API.Verbs
...
+    :<|> Transactions
+    :<|> ShelleyMigrations
+    :<|> StakePools
     :<|> ByronWallets
-    :<|> ByronAddresses n
-    :<|> ByronTransactions n
-    :<|> ByronMigrations n
+    :<|> ByronAddresses
+    :<|> ByronTransactions
+    :<|> ByronMigrations
```
KtorZ, line 206: Feels good.

lib/core/src/Cardano/Wallet/Api/Link.hs
```diff
@@ -506,14 +504,6 @@ instance Discriminate 'Shelley where
 instance Discriminate 'Byron where
     discriminate _ a = a
 
--- | Some endpoints are parameterized via a network discriminant in order to
--- correctly encode their end type (for example, 'CreateTransaction n'). Yet, in
--- the context of this module, the network discrimination doesn't matter for it
--- has no influence on the endpoint's value and/or path parameters.
---
--- To ease type signatures, we therefore arbitrarily fix the network to Mainnet.
-type Net = 'Mainnet
```
KtorZ, line 561: :ok_hand:

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -535,7 +529,9 @@ postShelleyWallet
     -> WalletPostData
     -> Handler ApiWallet
 postShelleyWallet ctx generateKey body = do
-    let state = mkSeqStateFromRootXPrv (rootXPrv, pwd) g
+    let net = error "todo: value level network"
```
KtorZ: :point_up:

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -535,7 +529,9 @@ postShelleyWallet
...
+    let addrScheme = shelleyScheme net (Just $ error "todo: stakingK")
```
KtorZ: :point_up:

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1004,64 +997,73 @@ getUTxOsStatistics ctx (ApiT wid) = do
...
-    withWorkerCtx ctx wid liftE liftE
-        $ \wrk -> liftHandler $ W.importRandomAddress @_ @s @k wrk wid addr
+putRandomAddress ctx (ApiT wid) (ApiAddress addr)  = do
+    withWorkerCtx @_ @s @k ctx wid liftE liftE
+        $ \wrk -> liftHandler $
+            W.importRandomAddress @_ @s @k wrk wid (decodeAddr addr)
     pure NoContent
+  where
+    decodeAddr = either
+            (error "TODO: handle addr decode failures in coerceCoin")
```
KtorZ: :point_up:

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1573,13 +1588,20 @@ mkApiTransaction txid ins outs (meta, timestamp) setTimeReference =
...
 
-coerceCoin :: forall (n :: NetworkDiscriminant). AddressAmount (ApiT Address, Proxy n) -> TxOut
-coerceCoin (AddressAmount (ApiT addr, _) (Quantity c)) =
-    TxOut addr (Coin $ fromIntegral c)
+coerceCoin :: AddressScheme k -> AddressAmount ApiAddress -> TxOut
+coerceCoin addrScheme (AddressAmount (ApiAddress addr) (Quantity c)) =
+    TxOut (decodeAddr addr) (Coin $ fromIntegral c)
+  where
+    decodeAddr = either
+            (error "TODO: handle addr decode failures in coerceCoin")
```
KtorZ, line 1859: :point_up:

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: Indeed.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
KtorZ: What is weird exactly? I don't follow ?

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1175,11 +1180,11 @@ checkpointId cp = (checkpointWalletId cp, checkpointSlot cp)
 
 -- | Functions for saving/loading the wallet's address discovery state into
 -- SQLite.
-class PersistState s where
+class PersistState s k where
     -- | Store the state for a checkpoint.
-    insertState :: (W.WalletId, W.SlotId) -> s -> SqlPersistT IO ()
+    insertState :: (W.WalletId, W.SlotId) -> s -> AddressScheme k-> SqlPersistT IO ()
```
KtorZ: Why would instances require an 'AddressScheme' for inserting the state? I don't think any does. If it's for symmetry with `selectState`, I think this is a wrong reason :)

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -136,6 +132,63 @@ import qualified Data.ByteString as BS
...
 
+-- @AddressScheme@ correlates @key@s with @Addresses@.
+--
+-- In a @SeqState@ we need tell if an @Address@ was created using a given @key@.
+--
+-- A @key@ may corrspond to multiple @Address@es. We cannot* convert
+-- an address into a public key.
+--
+-- To be able to compare @key@s and addresses, we convert them both to
+-- fingerprints, which are hashes of the public keys.
```
KtorZ, line 145: Nope. This is not it. cf my previous comment. The fingerprint is an identifier for the spending key of an address that is unaffected by whether or not the address is a delegation address or not. Its main use is for comparing addresses with addresses.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -136,6 +132,63 @@ import qualified Data.ByteString as BS
...
+--
+-- @@
+--
+--            paymentAddress         delegationAddress
+--
+--    Address ◀────────────── Key ──────────────▶ Address
```
KtorZ, line 151: Nice but the diagram feels rather inaccurate. There are actually 2 keys involved, which created the need for the fingerprint thingy. 

```
Payment Address     ───────────────▶ Fingerprint 
  | Payment Key PK1                       ▲
                                          │ ==
                                          ▼
Delegation Address  ───────────────▶ Fingerprint 
  | Payment Key PK1                       ▲
  | Stake Key   SK1                       │ ==
                                          ▼
Delegation Address  ───────────────▶ Fingerprint
  | Payment Key PK1
  | Stake Key   SK2 
```

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -136,6 +132,63 @@ import qualified Data.ByteString as BS
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
 
+-- @AddressScheme@ correlates @key@s with @Addresses@.
+--
+-- In a @SeqState@ we need tell if an @Address@ was created using a given @key@.
+--
+-- A @key@ may corrspond to multiple @Address@es. We cannot* convert
+-- an address into a public key.
```
KtorZ, line 142: These two statements are false. Jörmungandr does embed the full public key, so it's possible to go back and forth. And, a key itself correspond to only a single address; unless you meant that a key can be part of many addresses when grouped with some other key or script hash.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -136,6 +132,63 @@ import qualified Data.ByteString as BS
...
+-- @@
+--
+-- *) Except in jormungandr
+--
+-- TODO: @addressToText@ and @addressFromText@ could /maybe/ be removed if we made
+-- Address an anbstract type. But something entirely differnt could be desired
+-- as well.
+data AddressScheme (key :: Depth -> * -> *) = AddressScheme
+    { addressFromKey
+        :: key 'AddressK XPub -> Address
```
KtorZ, line 174: This is missing the delegation addresses where addresses are constructed from two keys.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -136,6 +132,63 @@ import qualified Data.ByteString as BS
...
+    , addressFromText
+        :: Text -> Either TextDecodingError Address
+    , addressFromFingerprint
+        :: KeyFingerprint -> Address
+        -- TODO: Can we remove the need for this somehow?
+        -- I think we can, by parameterizing AddressPool over address.
+    } deriving Generic
+
+instance Show (AddressScheme k) where
+    show _ = "<AddressScheme>"
```
KtorZ, line 194: We can probably do something a little nicer by requiring `Typeable k`.

KtorZ (changes requested):

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
Anviking: > Jörmungandr does embed the full public key

Hence the `*` asterisk :)

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
Anviking: > This is missing the delegation addresses where addresses are constructed from two keys.

I thought we could conflate `paymentAddress` and `delegationAddress` into one `addressFromKey`, where e.g. the `shelleyScheme` constructor takes the `Maybe StakingKey` as an argument, and produces `paymentAddress` if `Nothing`, and `delegationAddress` if `Just`.

https://github.com/input-output-hk/cardano-wallet/pull/1702/files/7a551ce9e87e717d3d0df85697db8a9a1d8ebedb#diff-708137f7529bf5efbae2de60278a966aR248-R263

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
Anviking: Yup, mistake, not intentional!

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
Anviking: Hm... Good idea!

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
Anviking: >  The fingerprint is an identifier for the spending key of an address that is unaffected by whether or not the address is a delegation address or not.

Yes, this was basically the notion I had. Don't see anything _clearly_ wrong (though perhaps your suggested diagram would indeed be much better) Will have a closer look later.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
KtorZ: Hmmm. That would probably work fine for now indeed with the hypotheses under which the wallet is operating (i.e. one stake key per wallet and, using only delegated addresses).

lib/byron/src/Cardano/Wallet/Byron/Api/Server.hs
Anviking: `ArgGenChange` _is_ removed. I am instead passing around a `s -> (Address, s)` function (same effect, but more down-to-earth).

Anviking:
Closing for now.

Issue #1701, opened 2020-06-01
Wallet "passphrase" info gets lost after updating cardano-wallet


nikolaglumac:
# Context
 
We are currently working on a next Daedalus ITN release ([2.4.0-ITN1](https://buildkite.com/input-output-hk/daedalus/builds/12987)) which uses cardano-wallet `031835950a38384b289f3adc43c671a00ef9000b` revision.
We've found an issue that is currently blocking us from proceeding with our release.
Given that you had a wallet created in the current production Daedalus ITN version ([2.3.1-ITN1](https://buildkite.com/input-output-hk/daedalus/builds/11329)) which uses cardano-wallet `8316bffdd05596e8830da9129f3a3907c6fba684` revision and then you update Daedalus to this next release, the wallet's passphrase info is lost - which makes Daedalus think this wallet has no password set.
**There is no option to have a passwordless wallet in any of the Daedalus ITN builds**

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | 031835950a38384b289f3adc43c671a00ef9000b |
| Platform     | All         |
| Installation | [Daedalus 2.4.0-ITN1](https://buildkite.com/input-output-hk/daedalus/builds/12987)           |

# Steps to Reproduce

1. Install production Daedalus ITN version: [2.3.1-ITN1](https://buildkite.com/input-output-hk/daedalus/builds/11329)
2. Create/restore a wallet (doesn't matter if it is a Byron/Shelley wallet)
3. Update Daedalus to the pre-production version: [2.4.0-ITN1](https://buildkite.com/input-output-hk/daedalus/builds/12987)
4. Daedalus will ask you to set a password for your existing wallet which will fail as the wallet already has a password set but the Api doesn't return this info correctly.
5. If you then proceed and create a new wallet in this version of Daedalus it will correctly display passphrase info.

### Postman GET wallets response
![screenshot](https://user-images.githubusercontent.com/376611/83407475-11e76c80-a411-11ea-8650-b71e6604e4a2.png)

☝️ as can be seen from this screenshot, the 1st wallet has no "passphrase" attribute, while the 2nd one (created after the update) has.

## Expected behavior

Wallet passphrase info should be preserved after the update.

## Actual behavior
 
Wallet passphrase info is lost after the update.

---

# Resolution 

---

# QA

piotr-iohk:
[...]

PR #1704, opened 2020-06-02
assign default encryption scheme for passphrase with none.

KtorZ:
# Issue Number

#1701 

# Overview

- a4dd767006047e0fc14fb6de7b20920ae854dc03
  :round_pushpin: **assign default encryption scheme for passphrase with none.**

# Comments

cc @nikolaglumac


paweljakubas (approved):
looks like a fix that addresses the case when we had updated password that had not passphrase schema  and was not exposed (because resulted in Nothing)   

Alternatively, we could solve this via manual migration...

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -378,6 +414,7 @@ migrateManually tr defaultFieldValues =
                 _ <- Sqlite.step addColumn
                 Sqlite.finalize addColumn
       where
+        activeSlotCoeff = DBField CheckpointActiveSlotCoeff
```
KtorZ, line 417: moved from above.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -353,6 +388,7 @@ migrateManually tr defaultFieldValues =
                 Sqlite.finalize stmt
       where
         hardLowerBound = toText $ fromEnum $ minBound @(Index 'Hardened _)
+        rndAccountIx   = DBField RndStateAddressAccountIndex
```
KtorZ, line 391: moved from above.

KtorZ:
bors retry

Issue #1707, opened 2020-06-02
BYRON_MIGRATE_05 - failing on Jormungandr


piotr-iohk:
# Context
 
 

# Test Case

https://buildkite.com/input-output-hk/cardano-wallet/builds/7820#af35c97c-e5fa-47ed-b31f-a1970e11acc9

# Failure / Counter-example

```

Failures:
--
  |  
  | src/Test/Integration/Scenario/API/Shelley/Wallets.hs:1196:19:
  | 1) API Specifications, BYRON_MIGRATE_05 - migrating from inappropriate wallet types, Byron
  | expected: Status {statusCode = 404, statusMessage = "Not Found"}
  | but got: Status {statusCode = 400, statusMessage = "Bad Request"}
  |  
  | from the following response: Left (DecodeFailure "Error in $: parsing [] failed, expected Array, but encountered Object: Response {responseStatus = Status {statusCode = 400, statusMessage = \"Bad Request\"}, responseVersion = HTTP/1.1, responseHeaders = [(\"Transfer-Encoding\",\"chunked\"),(\"Date\",\"Tue, 02 Jun 2020 01:20:44 GMT\"),(\"Server\",\"Warp/3.3.5\"),(\"Content-Type\",\"application/json;charset=utf-8\")], responseBody = \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"Error in $.addresses[0]: This type of address is not supported: [97].\\\"}\", responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}")
  |  
  | To rerun use: --match "/API Specifications/BYRON_MIGRATE_05 - migrating from inappropriate wallet types/Byron/"
  |  
  | src/Test/Integration/Scenario/API/Shelley/Wallets.hs:1213:19:
  | 2) API Specifications, BYRON_MIGRATE_05 - migrating from inappropriate wallet types, Icarus
  | expected: Status {statusCode = 404, statusMessage = "Not Found"}
  | but got: Status {statusCode = 400, statusMessage = "Bad Request"}
  |  
[...]

PR #1711, opened 2020-06-02
Rewrite BYRON_MIGRATE_05 to try to migrate to different addresses (sh…

piotr-iohk:
…elley, byron, icarus) instead of trying to migrate from deleted wallet

# Issue Number

https://github.com/input-output-hk/cardano-wallet/issues/1707

# Overview

- b1ba795a998c9cd8899aff3b983e3de5410cbe9d
  Rewrite BYRON_MIGRATE_05 to try to migrate to different addresses (shelley, byron, icarus) instead of trying to migrate from deleted wallet
  
- 1a6fb36dd53dbc7a27a26c6c6bba7e9db6510960
  Use shelley addresses from the wallet in the test
 

# Comments

In attempt of fixing BYRON_MIGRATE_05 (https://github.com/input-output-hk/cardano-wallet/issues/1707) failing on Jormungandr, I have rewritten it a bit, so it is trying to migrate to different address types. This works fine on `cardano-node` but fails for `jormungandr`.

It seems that the address that's causing the issue is `addr1vxyuj54gmnzxxhtrgeqd84df9zeudgvnaxy8lwh8ne29xp5mnl2v55m8r2n` which triggers following response:

```
Code = 400,
Message = {"code":"bad_request","message":"Error in $.addresses[0]:This type of address is not supported: [97]."}
```

The address is produced by the code:

```haskell
        addrShelley <- runIO
            $ encodeAddress @n . head . shelleyAddresses @n
            . entropyToMnemonic @15 <$> genEntropy
```

Cannot reproduce it manually (i.e. taking address from a shelley wallet)

---
**UPDATE:**
In 1a6fb36dd53dbc7a27a26c6c6bba7e9db6510960 I have "fixed" it so the shelley address in test is generated like this:
```
                --shelley address
                wShelley <- emptyWallet ctx
                addrs <- listAddresses @n ctx wShelley
                let addrShelley = (addrs !! 1) ^. #id
```

This works for haskell node and jormungandr...

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
@@ -42,6 +42,10 @@ import Cardano.Wallet.Api.Types
     )
 import Cardano.Wallet.Primitive.AddressDerivation
     ( PassphraseMaxLength (..), PassphraseMinLength (..), PaymentAddress )
+import Cardano.Wallet.Primitive.AddressDerivation.Byron
+    ( ByronKey )
+import Cardano.Wallet.Primitive.AddressDerivation.Icarus
+    ( IcarusKey )
 import Cardano.Wallet.Primitive.AddressDerivation.Shelley
     ( ShelleyKey )
 import Cardano.Wallet.Primitive.AddressDiscovery.Sequential
@@ -99,9 +103,11 @@ import Test.Integration.Framework.DSL
     , fixtureRandomWallet
     , fixtureWallet
     , getFromResponse
+    , icarusAddresses
     , json
     , listAddresses
     , notDelegating
+    , randomAddresses
     , request
     , selectCoins
     , shelleyAddresses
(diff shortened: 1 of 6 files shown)


paweljakubas:
the failure above because of https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Jormungandr.hs#L389 ?
ie., wrong first byte so wrong network used in address?

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
```diff
@@ -1174,68 +1180,42 @@ spec = do
...
-                ]
+    describe "BYRON_MIGRATE_05 - I could migrate to any valid address" $ do
+        forM_ [ ("Byron", emptyRandomWallet)
+              , ("Icarus", emptyIcarusWallet)
+              ] $ \(walType, destWallet) -> do
+
+            it ("From wallet type: " ++ walType) $ \ctx -> do
+                --shelley address
+                wShelley <- emptyWallet ctx
+                addrs <- listAddresses @n ctx wShelley
```
paweljakubas, line 1191: :+1:

paweljakubas (approved):
looks ok

PR #1700, opened 2020-05-29
Introduce `ProtocolParameters` type.

jonathanknowles:
# Related Issue

#1693 

# Overview

This PR prepares the ground for reporting the live value of the **_d_** protocol parameter.

We introduce the following type to hold **protocol parameters**:

```hs
data ProtocolParameters = ProtocolParameters  
    { decentralizationLevel
        :: Quantity "percent" Percentage      
    , txParameters
        :: TxParameters       
    }
```

Using the above type, we can rewrite the following type:

```diff
- data GenesisBlockParameters = GenesisBlockParameters
-     { staticParameters :: BlockchainParameters
-     ,     txParameters ::         TxParameters
-     }
+ data NetworkParameters = NetworkParameters
+     {  genesisParameters ::  GenesisParameters
+     , protocolParameters :: ProtocolParameters
+     }
```

Finally, we change the type of `toApiNetworkParameters` to:
```diff
- toApiNetworkParameters :: BlockchainParameters -> ApiNetworkParameters
+ toApiNetworkParameters ::    NetworkParameters -> ApiNetworkParameters
```

## Outstanding Work

- [x] Ensure that related functions and values are named consistently.

Issue #1693, opened 2020-05-27
Report live value of `d` parameter in `decentralizationLevel`.


jonathanknowles:
# Context

Task #1692 will have added a skeletal `decentralizationLevel` parameter to the `ApiNetworkParameters` object.

# Decision

Adjust the code so that this parameter reflects the true value of the `d` parameter, as reported by the network.

# Acceptance Criteria

- The `GetNetworkParameters` API endpoint **must** return a decentralization level percentage that reflects the true value of the `d` parameter, as reported by the network.

# Development 

- [x] (https://github.com/input-output-hk/cardano-wallet/pull/1700)
    Introduce `ProtocolParameters` type to hold (a subset of) the protocol parameters defined in [a Formal Specification of the Cardano Ledger](https://hydra.iohk.io/build/2839558/download/1/ledger-spec.pdf).

    ```hs
    data ProtocolParameters = ProtocolParameters  
        { decentralizationLevel
            :: Quantity "percent" Percentage      
        , txParameters
            :: TxParameters       
        }
    ```

- [x] (https://github.com/input-output-hk/cardano-wallet/pull/1700)
    Rewrite the `GenesisBlockParameters` type to include a `ProtocolParameters` value (whose type is defined above):
    ```patch
    - data GenesisBlockParameters = GenesisBlockParameters
    -     { staticParameters :: BlockchainParameters
    -     ,     txParameters ::         TxParameters
    -     }
    + data NetworkParameters = NetworkParameters
    +     {  genesisParameters ::  GenesisParameters
    +     , protocolParameters :: ProtocolParameters
    +     }
    ```
- [x] (https://github.com/input-output-hk/cardano-wallet/pull/1700)
    Change the type of `toApiNetworkParameters` to:

    ```patch
    - toApiNetworkParameters :: BlockchainParameters -> ApiNetworkParameters
    + toApiNetworkParameters ::    NetworkParameters -> ApiNetworkParameters
    ```

- [x] (https://github.com/input-output-hk/cardano-wallet/pull/1715)
    Revise documentation comments for the following types:
    - [x] `NetworkParameters`
    - [x] `GenesisParameters`
    - [x] `ProtocolParameters`<br/><br/>

- [x] (https://github.com/input-output-hk/cardano-wallet/pull/1725)
    Generalize the wallet, network, and database layers to use `ProtocolParameters` instead of `TxParameters`. (`ProtocolParameters` includes `TxParameters`).
 
[...]

PR #1715, opened 2020-06-03
Revise documentation for `NetworkParameters` and related types.

jonathanknowles:
# Related Issues

#1693 

# Overview

PR #1700 changed the way that network parameters are defined.

This PR revises and adds documentation for the following types:

- [x] `NetworkParameters`
- [x] `GenesisParameters`
- [x] `ProtocolParameters`
- [x] `TxParameters`

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types.hs
@@ -1299,6 +1299,12 @@ instance NFData NetworkParameters
 instance Buildable NetworkParameters where
     build (NetworkParameters gp pp) = build gp <> build pp
 
+-- | Parameters defined by the __genesis block__.
+--
+-- At present, these values cannot be changed through the update system.
+--
+-- They can only be changed through a soft or hard fork.
+--
 data GenesisParameters = GenesisParameters
     { getGenesisBlockHash :: Hash "Genesis"
         -- ^ Hash of the very first block
--- a/lib/core/src/Cardano/Wallet/Primitive/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types.hs
@@ -1360,8 +1360,9 @@ data ProtocolParameters = ProtocolParameters
     { decentralizationLevel
         :: Quantity "percent" Percentage
         -- ^ The current level of decentralization in the network.
-        --   * '  0%' indicates that the network is /completely federalized/.
-        --   * '100%' indicates that the network is /completely decentralized/.
+        --   (also known as the 'd' parameter).
+        -- * '  0%' indicates that the network is /completely federalized/.
+        -- * '100%' indicates that the network is /completely decentralized/.
     , txParameters
         :: TxParameters
         -- ^ Parameters relating to transactions.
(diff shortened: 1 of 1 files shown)


lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -1351,11 +1354,17 @@ slotParams gp =
         (gp ^. #getGenesisBlockDate)
         (gp ^. #getActiveSlotCoefficient)
 
+-- | Protocol parameters that can be changed through the update system.
+--
 data ProtocolParameters = ProtocolParameters
     { decentralizationLevel
         :: Quantity "percent" Percentage
+        -- ^ The current level of decentralization in the network.
```
rvl, line 1363: You could add that this is the _d_ parameter from the formal spec.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -1351,11 +1354,17 @@ slotParams gp =
...
+        --   * '  0%' indicates that the network is /completely federalized/.
+        --   * '100%' indicates that the network is /completely decentralized/.
     , txParameters
         :: TxParameters
+        -- ^ Parameters relating to transactions.
```
rvl:
```suggestion
        -- ^ Parameters which affect transaction creation.
```

rvl (approved):
Better

lib/core/src/Cardano/Wallet/Primitive/Types.hs
jonathanknowles: Fixed in a417440ceb99cfc997651e6055eefc6453048470.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
jonathanknowles: Fixed in cfaf9ac6a561b94034465d28ad883835a13ac38b.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -1279,26 +1279,29 @@ computeUtxoStatistics btype utxos =
     (^!) = (^)
 
 {-------------------------------------------------------------------------------
-                             Blockchain Parameters
+                              Network Parameters
 -------------------------------------------------------------------------------}
 
--- | Initial blockchain parameters loaded from the application configuration and
--- genesis block.
+-- | Records the complete set of parameters currently in use by the network.
```
rvl: Actually this isn't the complete set of parameters. Just those which are relevant to the wallet.

rvl (approved):

lib/core/src/Cardano/Wallet/Primitive/Types.hs
jonathanknowles: Fixed in e26d3ae8bce792b5530bd8f82324accd95a0c668.

rvl:
The PR #1716 should fix those `ci/hydra-eval` failures.

Issue #1552, opened 2020-04-08
Migration issue on Jormungandr (v2020-03-16 => v2020-04-01 | 07)


piotr-iohk:
# Context
 
After migration `v2020-03-16` => `v2020-04-01` (or `v2020-04-07`) one cannot send tx from Shelley wallet or migrate Byron wallet... 

Note: as a workaround one can remove and restore the wallet again.

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | cardano-wallet  `v2020-03-16` => `v2020-04-01`, jormungandr 0.8.15  |
| Platform     | Probably all, discovered on Linux          |
| Installation | from release          |

# Steps to Reproduce

1. Launch server using cardano-wallet `v2020-03-16`
```
cd lib/jormungandr/test/data/jormungandr/test_scripts
./wallet_launch.sh
```
2. Create a wallet (e.g. faucet shelley wallet from integration tests: `absurd seat ball together donate bulk sustain loop convince capital peanut mutual notice improve jewel`)
3. Stop wallet and start wallet  `v2020-04-01` or `v2020-04-07`
4. Submit transaction from that wallet

## Expected behavior

TX is successful.

## Actual behavior
 
```
Code = 403,
Json = {"code":"no_root_key","message":"I couldn't find a root private key for the given wallet: 6e262e86ac68ab5f71d0b6dfe27cdc0609db5198. However, this operation requires that I do have such a key. Either there's no such wallet, or I don't fully own it."}
```

---

# Resolution 

---

# QA

Anviking:
`cardano-wallet-byron` is not affected?

piotr-iohk:
No, there is no `v2020-03-16` for cardano-wallet-byron. Migration `v2020-04-01` => `v2020-04-07` seems to be fine.

piotr-iohk:
fixed with #1717.

PR #1717, opened 2020-06-03
Replace 'Maybe Bool'  with a dedicated sum-type. 

KtorZ:
# Issue Number

#1701

# Overview

- 9aaf4ad62c6b5982ef381073f84f2b76d5251f2f
  :round_pushpin: **log when no migration is needed**
  
- 7c3d459adfbd86353bd84b31f8d9220bd9c9532f
  :round_pushpin: **replace 'Maybe Bool' with a dedicated sum-type for more clarity**
  
# Comments

Caught me one time already... Won't caught anyone else.

### DIFF
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
@@ -352,6 +352,14 @@ migrateManually tr defaultFieldValues =
                 traceWith tr $ MsgManualMigrationNotNeeded passphraseScheme
             ColumnMissing -> do
                 traceWith tr $ MsgManualMigrationNotNeeded passphraseScheme
+                query <- Sqlite.prepare conn $ T.unwords
+                    [ "ALTER TABLE", tableName passphraseScheme
+                    , "ADD COLUMN", fieldName passphraseScheme
+                    , fieldType passphraseScheme, " NULL"
+                    , ";"
+                    ]
+                Sqlite.step query *> Sqlite.finalize query
+                assignDefaultPassphraseScheme conn -- loop to apply case below
             ColumnPresent  -> do
                 value <- either (fail . show) (\x -> pure $ "\"" <> x <> "\"") $
                     fromPersistValueText (toPersistValue W.EncryptWithPBKDF2)
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
@@ -341,7 +341,8 @@ migrateManually tr defaultFieldValues =
     assignDefaultPassphraseScheme :: Sqlite.Connection -> IO ()
     assignDefaultPassphraseScheme conn = do
         isFieldPresent conn passphraseScheme >>= \case
-            Nothing -> pure ()
+            Nothing -> do
+                traceWith tr $ MsgManualMigrationNotNeeded passphraseScheme
             Just _  -> do
                 value <- either (fail . show) (\x -> pure $ "\"" <> x <> "\"") $
                     fromPersistValueText (toPersistValue W.EncryptWithPBKDF2)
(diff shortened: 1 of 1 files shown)


lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -306,6 +306,13 @@ findDatabases tr dir = do
   where
     expectedPrefix = T.pack $ keyTypeDescriptor $ Proxy @k
 
+-- | A data-type for capturing column status. Used to be represented as a
+-- 'Maybe Bool' which is somewhat confusing to interpret.
+data SQLColumnStatus
+    = TableDoesntExist
+    | ColumnDoesntExist
+    | ColumnExists
```
jonathanknowles: Perhaps this wording would be a little more concise:
```suggestion
    = TableMissing
    | ColumnMissing
    | ColumnPresent
```

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -306,6 +306,13 @@ findDatabases tr dir = do
   where
     expectedPrefix = T.pack $ keyTypeDescriptor $ Proxy @k
 
+-- | A data-type for capturing column status. Used to be represented as a
+-- 'Maybe Bool' which is somewhat confusing to interpret.
+data SQLColumnStatus
```
jonathanknowles: Suggestion: use PascalCase here, treating SQL as a word.

```suggestion
data SqlColumnStatus
```

Similar examples:
- [SqlBackend](https://hackage.haskell.org/package/persistent-2.10.5.1/docs/Database-Persist-Sql.html#t:SqlBackend)
- [InsertSqlResult](https://hackage.haskell.org/package/persistent-2.10.5.1/docs/Database-Persist-Sql.html#t:InsertSqlResult)

KtorZ: I do actually prefer the golang guidelines on acronyms, but whatever ^.^

paweljakubas (approved):

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -306,6 +306,13 @@ findDatabases tr dir = do
   where
     expectedPrefix = T.pack $ keyTypeDescriptor $ Proxy @k
 
+-- | A data-type for capturing column status. Used to be represented as a
+-- 'Maybe Bool' which is somewhat confusing to interpret.
```
jonathanknowles, line 310: Suggestion: remove the line starting with "Used to be represented...".

Justification: the new code is already very clear. :+1: 

(We don't need to remind the reader how confusing it used to be :smile_cat:)

```suggestion
-- | Indicates whether or not a column and its parent table are present
--   in the database.
```

piotr-iohk:
Unfortunately I still see behavior as in: https://github.com/input-output-hk/cardano-wallet/issues/1552

To reproduce:
1. Launch server using cardano-wallet `v2020-03-16`
```
cd lib/jormungandr/test/data/jormungandr/test_scripts
./wallet_launch.sh
```
2. Create a wallet (e.g. faucet shelley wallet from integration tests: `absurd seat ball together donate bulk sustain loop convince capital peanut mutual notice improve jewel`)
3. Stop wallet and start wallet  with version from this branch
4. Submit transaction from that wallet
Result:

```
Code = 403,
Message = {"code":"no_root_key","message":"I couldn't find a root private key for the given wallet: f425f750be117385fb7c4280dee5cf9be2f0e71c. However, this operation requires that I do have such a key. Either there's no such wallet, or I don't fully own it."}
```

piotr-iohk:
What is interesting that after point 4 (from the above) when you **restart wallet again** (using the same version) you are able to send transaction... :thinking:

KtorZ:
@piotr-iohk My guess here is that, when you first restart it, the automatic migration hasn't happened _yet_, so the column doesn't yet exists and can't be populated (so, running into the bug as before). Manual migrations are ran BEFORE automatic migration. which is, somewhat necessary for the active slot coeff as it requires to alter the tables before attempting an automatic migration, but problematic for this one which requires the passphrase_scheme to be set :| ... 

We could "force-create" the column if it doesn't exist. It won't hurt and we are sure it's there.

piotr-iohk (approved):
works on my computer :tada:

PR #1725, opened 2020-06-04
Report live value of decentralization parameter d.

jonathanknowles:
# Issue Number

#1693 

# Overview

This PR:

- [x] Introduces the `DecentralizationLevel` type (based on the `Percentage` type).

    _Greater_ values indicate _greater_ levels of decentralization, as follows:

    | Value | Meaning |
    | --: | :-- |
    | 0% | Network is completely _federalized_ |
    | 100% | Network is completely _decentralized_ |

- [x] Adjusts the **wallet API server** to convert from values of type `DecentralizationLevel` to values of type `Quantity "percent" Percentage`.

- [x] Adjusts the **wallet layer** to store a value of type `ProtocolParameters` instead of `TxParameters`. (Note that `ProtocolParameters` _includes_ `TxParameters`.)

- [x] Adjusts the **network layer** to provide a `getProtocolParameters` function instead of `getTxParameters`.

- [x] Adjusts the **database layer** to use a `protocol_parameters` table instead of a `tx_parameters` table. (Note that `protocol_parameters` is a superset of `tx_parameters`.)

- [x] Adjusts the **chain-following code** to extract the live value of `d`, in the same way as we already do for transaction-related protocol parameters.

- [x] Adds a **conversion function** to convert from `d` parameter values to values of type `DecentralizationLevel`.

    This function performs the following mapping:
    ```
    {1 → 0%, 0.9 → 10%, 0.8 → 20%, ..., 0 → 100%}
    ```

# Comments

Starting a wallet with this PR applied will result in the creation of a new `protocol_parameters` table, if one does not already exist. (We rely on the automatic migration mechanism to create the table, and the network layer to add rows to the table.)

The redundant `tx_parameters` table, if it exists, will be ignored, but _not_ deleted.

In a future PR, we can add a manual migration step that removes the `tx_parameters` table.

See Issue https://github.com/input-output-hk/cardano-wallet/issues/1727.

### DIFF
--- a/lib/byron/src/Cardano/Wallet/Byron/Network.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron/Network.hs
@@ -603,8 +603,8 @@ instance ToText NetworkLayerLog where
             [ "Network node tip is"
             , pretty bh
             ]
-        MsgProtocolParameters params -> T.unwords
-            [ "ProtocolParams for tip are:"
+        MsgProtocolParameters params -> T.unlines
+            [ "Protocol parameters for tip are:"
             , pretty params
             ]
         MsgLocalStateQueryError e -> T.unwords
--- a/lib/byron/src/Cardano/Wallet/Byron/Network.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron/Network.hs
@@ -42,10 +42,10 @@ import Cardano.Wallet.Byron.Compatibility
     , fromChainHash
     , fromSlotNo
     , fromTip
+    , protocolParametersFromUpdateState
     , toEpochSlots
     , toGenTx
     , toPoint
-    , txParametersFromUpdateState
     )
 import Cardano.Wallet.Network
     ( Cursor, ErrPostTx (..), NetworkLayer (..), mapCursor )
@@ -214,12 +214,11 @@ withNetworkLayer tr np addrInfo versionData action = do
     -- doesn't really do anything but sending dummy messages to get the node's
     -- tip. It doesn't rely on the intersection to be up-to-date.
     nodeTipVar <- atomically $ newTVar TipGenesis
-    txParamsVar <- atomically $ newTVar $
-        W.txParameters $ W.protocolParameters np
+    protocolParamsVar <- atomically $ newTVar $ W.protocolParameters np
     nodeTipClient <- mkTipSyncClient tr np
         localTxSubmissionQ
         (atomically . writeTVar nodeTipVar)
-        (atomically . writeTVar txParamsVar)
+        (atomically . writeTVar protocolParamsVar)
     let handlers = retryOnConnectionLost tr
     link =<< async
         (connectClient tr handlers nodeTipClient versionData addrInfo)
@@ -230,7 +229,7 @@ withNetworkLayer tr np addrInfo versionData action = do
             , nextBlocks = _nextBlocks
             , initCursor = _initCursor
             , cursorSlotId = _cursorSlotId
-            , getTxParameters = atomically $ readTVar txParamsVar
+            , getProtocolParameters = atomically $ readTVar protocolParamsVar
             , postTx = _postTx localTxSubmissionQ
             , stakeDistribution = _stakeDistribution
             , getAccountBalance = _getAccountBalance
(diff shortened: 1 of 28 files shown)


lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -345,18 +346,36 @@ fromMaxTxSize :: Natural -> Quantity "byte" Word16
...
+fromPParams :: SL.PParams -> W.ProtocolParameters
+fromPParams pp = W.ProtocolParameters
+    { decentralizationLevel = decentralizationLevelFromPParams pp
+    , txParameters = W.TxParameters
+        { getFeePolicy = W.LinearFee
+            (Quantity (naturalToDouble (SL._minfeeB pp)))
+            (Quantity (fromIntegral (SL._minfeeA pp)))
+            (Quantity 0) -- TODO: it's not as simple as this?
+        , getTxMaxSize = fromMaxTxSize $ SL._maxTxSize pp
+        }
```
jonathanknowles: Perhaps we should also extract a `txParametersFromPParams` function here.

jonathanknowles: Fixed in 53352b61478ce658986a596b03d8691c2b80120f.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -597,8 +603,8 @@ instance ToText NetworkLayerLog where
             [ "Network node tip is"
             , pretty bh
             ]
-        MsgTxParameters params -> T.unwords
-            [ "TxParams for tip are:"
+        MsgProtocolParameters params -> T.unwords
+            [ "ProtocolParams for tip are:"
```
rvl: How does this log message look? Would it be better as a multiline text (`T.unwords` → `T.unlines`)?

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -345,8 +346,50 @@ fromMaxTxSize :: Natural -> Quantity "byte" Word16
...
+    :: SL.PParams
+    -> W.DecentralizationLevel
+decentralizationLevelFromPParams pp =
+    either reportInvalidValue W.DecentralizationLevel
+        $ mkPercentage
+        $ (1 -)
+        $ SL.intervalValue d
+  where
+    d = SL._d pp
+    reportInvalidValue = error $ mconcat
```
rvl, line 395: What about adding a `HasCallStack` constraint to this function and its callers?

rvl (approved):
Looks good so far

rvl:
> Write a manual DB upgrade step to cope with the changed schema.

Rather than doing this migration step, you could change `readWalletProtocolParameters` to use the genesis block as a fallback if the data is missing. That might be easier.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
jonathanknowles: Fixed in 4bc118bf324f63f183be82c85731e4042d3c87f5.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
jonathanknowles: > How does this log message look? Would it be better as a multiline text (`T.unwords` → `T.unlines`)?

Currently it looks like this:

```
ProtocolParams for tip are: [Decentralization level: 0.00%, Transaction parameters: [Fee policy: 14.0 + 42.0x + 5.0y, Tx max size: 8192]]
```

Using `T.unlines` makes it look like this:
```
ProtocolParams for tip are:
[Decentralization level: 0.00%, Transaction parameters: [Fee policy: 14.0 + 42.0x + 5.0y, Tx max size: 8192]]
```

I'll try to make some further improvements.

jonathanknowles: With commit 4de805b147e5626614c36d12fa65958db7bb9220, the log output now looks like this:

```
Protocol parameters for tip are:
 Decentralization level: 0.00%
 Transaction parameters: [Fee policy: 14.0 + 42.0x + 5.0y, Tx max size: 8192]
```

jonathanknowles:
> Rather than doing this migration step, you could change `readWalletProtocolParameters` to use the genesis block as a fallback if the data is missing. That might be easier.

After a discussion with @rvl, we decided to leave `readWalletProtocolParameters` unchanged.

Justification:

1. If a user starts with an old version of the wallet DB (one with a `tx_parameters` table instead of a `protocol_parameters` table), then the automatic migration mechanism will kick in and create an empty `protocol_parameters` table.
2. Once the network layer has started, it will populate the empty `protocol_parameters` with a row for the wallet.
3. The `readWalletProtocolParameters` function is only ever called by the `selectCoinsSetup` function.
4. This means that immediately after a migration from an older database without a `protocol_parameters` table, attempting to create a coin selection will fail if the network layer has not yet started, because we rely on the network layer to populate the `protocol_parameters` table.
5. However, in normal startup situations without a migration, if the `protocol_parameters` table already exists and is already populated, then creating a coin selection will succeed even if the network layer has not yet started.

So the corner case here is: attempting to create a coin selection _immediately after_ a migration, but _before_ the network layer has started. In this corner case, creating a coin selection will fail.

On the assumption that this corner case is relatively rare and easy to work around (by waiting for the network layer to start), we've decided to leave the `readWalletProtocolParameters` function unchanged, for the moment.

rvl (approved):
Great! :+1:

PR #1729, opened 2020-06-05
Fix mistakes in .travis.yml deployment preventing shelley binaries from being deployed

Anviking:
# Issue Number

For v2020-06-05 release.

# Overview

- Rename `cardano-wallet` to `cardano-wallet-byron`
- Make sure shelley binaries are called `cardano-wallet-shelley`, and remember to add the output to the `files` list

# Comments

Should work now:

Skärmavbild 2020-06-08 kl  15 50 21

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -114,21 +114,21 @@ jobs:
     # Fetch the linux archive
     - travis_retry curl -L https://hydra.iohk.io/job/Cardano/cardano-wallet/cardano-wallet-shelley-linux64/latest/download/1 | tar xz
     - find . -maxdepth 1 -type d -name "cardano-wallet-shelley-*" -exec mv \{} cardano-wallet-shelley-linux64 \;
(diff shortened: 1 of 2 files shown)


KtorZ:
hmm. That is the timeout of the _cache_ (2h is a little excessive... 37 minutes was already excessive, but since we're uploading ~1Go of cache the default of 3min was too small); the job are limited to 50 minutes max and there's no way around that on the free-tier. The travis "pro" version allows jobs up 2h but we don't have that here.

Anviking:
Btw, I can't have been in my right mind when converting `1337 s`, because it is ≈22 minutes, not 37 😅 

> the job are limited to 50 minutes

I've set it to 40 min / 2400 s now.

Anviking:
Although it seems the job already ran for 50 min 🤔  https://travis-ci.org/github/input-output-hk/cardano-wallet/jobs/695076828

KtorZ (approved):
LGTM, modulo what we discussed on Slack.

PR #1731, opened 2020-06-08
Foundational work to get a cluster of pools for integration testing

KtorZ:
# Issue Number

N/A

# Overview

- 5a263bc2f6db213ca9a222968525607b0a45e117
  :round_pushpin: **move content of 'Test.Utils.Ports' to 'Cardano.Wallet.Network.Ports'**
  
- 6ed9efca634aa960137d1966be49215be7bede0f
  :round_pushpin: **rework the shelley launcher code so that we can start a cluster of nodes**
  using 'withCluster ... ... 0' will result in the same behavior as before. Right now, pools would
exists as process but not on chain. In order to exist on chain, pools need to be registered, which
will be done in a separate commit.

- 1a8b720a95870a0834e737f64656fd0ce597dfbc
  :round_pushpin: **fix database path**
  
- 6745ea944d20752353d6e4c978b6670ae7b75be2
  
- 6b7b94763e4d3e4ceb9663dee0beb530b0a49e15
  :round_pushpin: **assign sensible default ordering to 'listAllTransactions'**
  
# Comments

This should make it possible to run a cluster of pools as simply as:

```hs
        let numberOfPools = 3
        withCluster tr Info numberOfPools $ \socketPath block0 (gp,vData) -> ...
```

So far, the pools won't have any _stake_ and won't therefore be any useful.. I'll give them some stake in another PR.


lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -163,152 +179,250 @@ parseGenesisData = \case
...
-    -> IO a
-withCardanoNode tr tdir severity action =
-    orThrow $ withConfig tdir severity $ \cfg block0 (np, vData) -> do
-        nodePort <- getRandomPort
-        let args = mkArgs cfg nodePort
+    -- ^ Action to run with the cluster up
+    -> IO (Either ProcessHasExited a)
+withCluster tr severity n action = do
+    ports <- randomUnusedTCPPorts (n + 1)
+    print $ permutations ports
```
paweljakubas: probably we want to use our logging infrastructure someday in the future - if not doing this now maybe add TO_DO to not miss it later

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -163,152 +179,250 @@ parseGenesisData = \case
...
+                (ExitFailure 1)
+        else do
+            action socket block0 params `finally` cancelAll
+  where
+    -- | Get permutations of the size (n-1) for a list of n elements, alongside with
+    -- the element left aside. `[a]` is really expected to be `Set a`.
+    --
+    -- >>> permutations [1,2,3]
+    -- [(1,[2,3]), (2, [1,3]), (3, [1,2])]
+    permutations :: Eq a => [a] -> [(a, [a])]
```
paweljakubas: maybe reuse https://hackage.haskell.org/package/base-4.14.0.0/docs/src/Data.OldList.html#permutations

paweljakubas: and then fmap with `f(x:xs)=[x,[xs]]` ?

paweljakubas (approved):
looks like sound scaffolding for what is coming next

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
KtorZ: What do you mean?

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -163,152 +179,250 @@ parseGenesisData = \case
...
+    print $ permutations ports
+    withBFTNode tr severity (head $ permutations ports) $ \socket block0 params -> do
+        waitGroup <- newChan
+        doneGroup <- newChan
+        let waitAll   = replicateM  n (readChan waitGroup)
+        let cancelAll = replicateM_ n (writeChan doneGroup ())
+
+        forM_ (init $ permutations ports) $ \(port, peers) -> do
+            withAsync
+                (withStakePool tr severity (port, peers) $ readChan doneGroup)
```
Anviking, line 217: Why not put the n-th `withStakePool` in the n-1-th's action?

Will all nodes stop if on stake pool goes down?

Anviking (approved):
looks neat

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
paweljakubas: I mean, we have now `print` - and maybe use our logging here?

Anviking: Also, something more descriptive than
> [(32403,[])]

would be nice

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
KtorZ: > Why not put the n-th withStakePool in the n-1-th's action?

What?

> Will all nodes stop if on stake pool goes down?

No. If one pools doesn't start for some reason, others will be cancelled. But once they've all started, if one dies, the cluster will keep running.

Anviking: > What?

Something like this:

```haskell
withStakePool args $ do
    withStakePool args $ do
        withStakePool args $ do
            .
              . n times
                .
```

Slower to start, but maybe simpler, and should stop all if one goes down.

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
KtorZ: It won't work because it's does not actually compute _all_ permutations (unlike Data.List.permutations). Using Data.List.permutations on the example above, we would also get [(1, [3,2]), (2, [3,1]), (3, [2,1])]. So, we should also nub + sort the second element which is arguably simpler than the current function :thinking: .. 

Changed for:

```hs
    rotate :: Ord a => [a] -> [(a, [a])]
    rotate = nub . fmap (\(x:xs) -> (x, sort xs)) . permutations
```

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
KtorZ: Ah sorry. I was really confused by your comment at first and didn't get why. This line is actually an oversight. I removed it :+1:

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -163,152 +179,242 @@ parseGenesisData = \case
...
+        forM_ (init $ rotate ports) $ \(port, peers) -> do
+            withAsync
+                (withStakePool tr severity (port, peers) $ readChan doneGroup)
+                (writeChan waitGroup . Just) -- FIXME: register pool
+              `finally` (writeChan waitGroup Nothing)
+
+        group <- waitAll
+        if length (catMaybes group) /= n then do
+            cancelAll
+            throwIO $ ProcessHasExited
```
rvl, line 224: This looks like it is mixing thrown exceptions with `IO (Either e a)`. If the bft node fails, it will return `Left`. If the others fail to start, they will throw.

lib/shelley/test/data/cardano-node-shelley/genesis.yaml
```diff
@@ -34,10 +34,10 @@ initialFunds: {}
 maxLovelaceSupply: 45000000000000000
 networkMagic: 1
 networkId: Mainnet
-epochLength: 1215
+epochLength: 100
 staking:
 slotsPerKESPeriod: 86400
-slotLength: 8
+slotLength: 0.2
```
rvl, line 40: Wow, that's a short slot. So if the tests get paused for 100x0.2=20secs (for whatever reason), they will not be able to continue.

lib/core/src/Cardano/Wallet/Network/Ports.hs
```diff
@@ -26,16 +26,20 @@ module Cardano.Wallet.Network.Ports
     -- * Helpers
     , waitForPort
     , unsafePortNumber
+    , findPort
+    , randomUnusedTCPPorts
```
rvl, line 30: Ideally we wouldn't have test util functions in the core library.
But I think you couldn't put it into the core-integration library due to a package cycle. Is that correct?

rvl (approved):

rvl:
The error from bors does look like a legitimate test failure:

```
Failures:

  src/Test/Integration/Scenario/API/Shelley/Transactions.hs:887:5:
  1) API Specifications TRANS_LIST_02,03x - Can limit/order results with start, end and order
       user error (expected a successful response but got an error: DecodeFailure "Error in $: parsing [] failed, expected Array, but encountered Object: Response {responseStatus = Status {statusCode = 400, statusMessage = \"Bad Request\"}, responseVersion = HTTP/1.1, responseHeaders = [(\"Transfer-Encoding\",\"chunked\"),(\"Date\",\"Mon, 08 Jun 2020 12:15:56 GMT\"),(\"Server\",\"Warp/3.3.5\")], responseBody = \"{\\\"code\\\":\\\"start_time_later_than_end_time\\\",\\\"message\\\":\\\"The specified start time '2020-06-08T12:15:56.471958Z' is later than the specified end time '2020-06-08T12:15:55.871958Z'.\\\"}\", responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}")

  To rerun use: --match "/API Specifications/TRANS_LIST_02,03x - Can limit/order results with start, end and order/"

Randomized with seed 1856737488

Finished in 223.8642 seconds
549 examples, 1 failure, 11 pending
builder for '/nix/store/6f339cm8hhd0k0xrdxmwbldkv8awg4yg-cardano-wallet-shelley-test-integration-2020.6.5-check.drv' failed with exit code 1
```

lib/shelley/test/data/cardano-node-shelley/genesis.yaml
KtorZ: I've tweaked according to what the ledger / consensus teams are using for testing as well; it seems fairly okay though. I don't get why the tests wouldn't be able to continue on if they pause for 20s :thinking: ?

lib/core/src/Cardano/Wallet/Network/Ports.hs
KtorZ: Actually no. I needed to share them and `Cardano.Wallet.Network.Ports` felt like a logical place to put them. I see your point about the "non production" aspect of these two. I believe they could fit in core-integration as well :thinking:

KtorZ:
@rvl Yes it does, although I thought I fixed it in https://github.com/input-output-hk/cardano-wallet/pull/1731/commits/6b7b94763e4d3e4ceb9663dee0beb530b0a49e15 but I obviously didn't. I only re-run once locally after this commit and it was probably luck that made it pass :/

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
KtorZ: Yes, I've changed that later actually while continuing working on this, but this isn't visible from this PR _yet_

lib/shelley/test/data/cardano-node-shelley/genesis.yaml
rvl: The epoch length is 20s and I _think_ if an entire epoch's worth of slots are missed then it will not be able to elect slot leaders for the next epoch. But it should be ok anyway. Nobody's going to press ctrl-z during integration tests and expect them to pass.

KtorZ: The cluster is made of one BFT leader + 3 pools though, so there's always the BFT leader capable of producing blocks. I am not sure how that influence the slot leader election however :thinking:

KtorZ: > Nobody's going to press ctrl-z during integration tests and expect them to pass.

I guess that's also a fair assumption :)

PR #1734, opened 2020-06-09
Cluster of stake pools in integration tests

KtorZ:
# Issue Number

N/A

# Overview

- de6fec030e2ec6ba0ea232df79f4ff133fe82f5a
  :round_pushpin: **add command-line helpers for generating stake key and certificates**
  
- 50ae86187a8dae89d67840c9d801df193f4d59fb
  :round_pushpin: **improve error handling and reporting when member of the cluster fail to start**
  It's actually nice to be able to tell what's going on. The command-line will likely undergo changes so it's really likely
that a command that used to work will fail. The previous message was sort of opaque / cryptic. Now we get a better error.

- 772c813dcfa9cc7831119756d92d432590eb9417
  :round_pushpin: **fixup add new gen methods**
  
- bb2b82e906f1df0c8fcaea6fa5e903578c997ef8
  :round_pushpin: **register stake pool and delegate funds to it using genesis faucets.**
  This makes sure that a stake pool joins the cluster only after it has been registered and is ready to process block.

- e08d5cd6f973f6be70fbabb3019d15c2d92a6a7c
  :round_pushpin: **Drop use of 'withAsync' so pools boot in parallel**
  The issue with 'withAsync' is that it's actually synchronous.
While the underlying action is executed in a thread, 'withAsync' will
only resolve when the async action is done. Yet, in our case, we want
all async actions to run in parallel and to be stopped when the test
suite is completed.

- de7559f772f3d173fbddcd5afb27afbf35bac69b
  :round_pushpin: **make the node a little less verbose regarding forging**
  With a slot length of 0.2, blocks are produced at a much faster pace.

- b2b3c0b97d5a9283380285a640765c9306d086a4
  :round_pushpin: **tweak genesis parameters with sensible values**
  A value of '1' for the active slot coefficient makes actually no sense. Only small values of it are
considered sensible. Tweaking also the decentralization level so that some blocks are deferred to
registered stake pools

- adf9dda8b854d148df22486a162f32de8ddb0af1
  :round_pushpin: **start a cluster of 3 pools when running integration tests**
  
- aaf7d6d1d04fb299126e40f121121db52320a7f2
  :round_pushpin: **show little disclaimer when starting integration tests**
  
- 2d7ece596e0c99d7b4846875399d069b2707cf75
  :round_pushpin: **do not restart the cluster between CLI and API spec**
  We have a limited number of faucets, and it takes quite some time now

# Comments

With this, we should be able to test all sort of things regarding pools o/

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
@@ -104,15 +104,9 @@ spec = do
         let networkParams = getFromResponse id r
         networkParams `shouldBe`
             toApiNetworkParameters (ctx ^. #_networkParameters)
-        let Right zeroPercent = Quantity <$> mkPercentage 0
+        let Right d = Quantity <$> mkPercentage .75 -- d is set to 0.25 in genesis
         verify r
-            -- NOTE: Currently, the decentralization level is hard-wired to 0%.
-            -- TODO: Adjust this test to expect the live value.
-            --
-            -- Related issue:
-            -- https://github.com/input-output-hk/cardano-wallet/issues/1693
-            --
-            [ expectField (#decentralizationLevel) (`shouldBe` zeroPercent) ]
+            [ expectField (#decentralizationLevel) (`shouldBe` d) ]
 
     it "NETWORK_CLOCK - Can query network clock" $ \ctx -> do
         sandboxed <- inNixBuild
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
@@ -14,10 +14,8 @@ import Cardano.Wallet.Api.Types
     , ApiEpochInfo (..)
     , ApiNetworkClock
     , ApiNetworkInformation
-    , ApiNetworkParameters (..)
     , NtpSyncingStatus (..)
     , WalletStyle (..)
-    , toApiNetworkParameters
     )
 import Cardano.Wallet.Primitive.Types
     ( SyncProgress (..) )
@@ -25,10 +23,6 @@ import Control.Monad
     ( when )
 import Control.Monad.IO.Class
     ( liftIO )
-import Data.Generics.Internal.VL.Lens
-    ( (^.) )
-import Data.Quantity
-    ( Quantity (..), mkPercentage )
 import Data.Time.Clock
     ( getCurrentTime )
 import Test.Hspec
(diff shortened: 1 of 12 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
```diff
@@ -104,14 +104,13 @@ spec = do
         let networkParams = getFromResponse id r
         networkParams `shouldBe`
             toApiNetworkParameters (ctx ^. #_networkParameters)
-        let Right zeroPercent = Quantity <$> mkPercentage 0
+        let Right zeroPercent = Quantity <$> mkPercentage 75
```
rvl: ?

rvl (approved):

lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
paweljakubas: yeah, why `75`?

lib/shelley/test/data/cardano-node-shelley/genesis.yaml
```diff
@@ -32,7 +31,8 @@ updateQuorum: 5
 maxMajorPV: 25446
 initialFunds: {}
 maxLovelaceSupply: 45000000000000000
-networkMagic: 1
+protocolMagicId: 1
+networkMagic: 764824073
```
paweljakubas, line 35: why we need this?

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -401,19 +442,214 @@ genVrfKeyPair dir = do
...
+    let file = dir </> "dlg.cert"
+    void $ cli
+        [ "shelley", "stake-address", "delegation-certificate"
+        , "--staking-verification-key-file", stakePub
+        , "--stake-pool-verification-key-file", opPub
+        , "--out-file", file
+        ]
+    pure file
+
+-- | Generate a raw transaction. We kill two birds one stone here by also
```
paweljakubas, line 519: aha

paweljakubas (approved):

lib/core-integration/src/Test/Integration/Scenario/API/Network.hs
KtorZ: Because the d-parameter is now set to `0.25` (was 1) to allow both the BFT node and pools to produce blocks. We actually report (1 - d) in the API, hence 75% here. We are 75% decentralized in the integration test.

lib/shelley/test/data/cardano-node-shelley/genesis.yaml
KtorZ: While new Shelley addresses uses the new protocolMagic (`1`), the **network** still uses the old magic (`764824073`). This is therefore needed to match the right network configuration of the cluster.

KtorZ:
bors try

rvl:
bors try

PR #1735, opened 2020-06-10
Clean up old TxParameters SQL table

### LINKED ISSUE #1727: Remove redundant `tx_parameters` table.
# Context

Issue https://github.com/input-output-hk/cardano-wallet/issues/1693 will introduce a new database table `protocol_parameters`. This table will include all columns included in the pre-existing `tx_parameters` table, and add a new `decentralization_level` column.

The pre-existing `tx_parameters` table will become redundant.

# Decision

Create a startup task to remove the `tx_parameters` table, if it exists.

# Acceptance Criteria

- [ ] The wallet _should_ remove the `tx_parameters` table on startup, if it exists.

# Development 

_To be written._

# QA 

_To be written._

rvl:
### Issue Number

Closes #1727

### Overview

Remove an old SQL table, if it exists.

### DIFF
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
@@ -331,6 +331,8 @@ migrateManually tr defaultFieldValues =
         -- really be removed as soon as we have a fix for the cardano-sl:wallet
         -- currently in production.
         removeSoftRndAddresses conn
+
+        removeOldTxParametersTable conn
   where
     -- NOTE
     -- Wallets created before the 'PassphraseScheme' was introduced have no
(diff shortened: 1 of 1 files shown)


lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -436,6 +438,13 @@ migrateManually tr defaultFieldValues =
             $ W.unActiveSlotCoefficient
             $ defaultActiveSlotCoefficient defaultFieldValues
 
+    -- | This table became @protocol_parameters@.
+    removeOldTxParametersTable :: Sqlite.Connection -> IO ()
+    removeOldTxParametersTable conn = do
+        dropTable <- Sqlite.prepare conn "DROP TABLE IF EXISTS tx_parameters;"
```
KtorZ, line 444: Could be nice to first look for the table using `isFieldPresent` and then, log accordingly depending on the outcome?

KtorZ (approved):

rvl:
From the buildkite log it looks like the integration tests may have a fd leak.

KtorZ:
@rvl Yes, I've observed this a few times already and I am trying to pinpoint the issue without success. Sometimes, there's _something_ causing an internal error on the server, after what, workers stop being removed and simply pile up as the test move forward. Listing wallets returns a bazillion number of wallets while also causing all subsequent test to fail.  The cause is unclear.

Issue #1718, opened 2020-06-03
List id and stake of pools


Anviking:
# Context
We want to implement this endpoint for (the cardano-node-) Shelley.
https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/listStakePools

There is a `LocalStateQuery` command to [`GetStakeDistribution`](https://github.com/input-output-hk/ouroboros-network/blob/e24f14f837d3ea4ee1942121ecdb4a4169a8df99/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs#L313-L314). Although this returns the stake as a fraction, not in real ada.

# Decision

- Use `GetStakeDistribution` list stake pools, providing their id and stake, as a first step. Other fields may be omitted or use dummy values.

# Acceptance Criteria

- [ ] API and CLI must be able to list stake pools
- [ ] the `id` field must be correct
- [ ] the `controlled_stake` field must be correct
- [ ] the `saturation` field _may_ be implemented

---

# Development 

- [ ] Can we get the total stake, or would it be acceptable to use the fractional stake we have?

# QA 

- See https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs#L443-L559

piotr-iohk:
lgtm

PR #1738, opened 2020-06-10
Parameterise `StakePools` API over apiPool type

Anviking:
# Issue Number

ADP-311, #1718 

# Overview

- [x] I parameterised `StakePools` API over `apiPool`
- [x] I renamed `ApiStakePool` to `ApiJormungandrStakePool`, such that we can later introduce a haskell `ApiStakePool`
- [x] I made the default `Api` type use `ApiJormungandrStakePool`. We should later switch to `ApiStakePool`, and update the swagger scheme such that it targets haskell, and not jormungandr.
- [x] I moved `StakePoolLayer` to the jormungandr package.

# Comments

- I am keeping the Pool DB in core, since I think we still want it. But I have not thought about how it would work with shelley.
- changelog: this is pure refactoring, but in combination with follow-up PRs this is a new feature.


Anviking:
bors try

lib/core/src/Cardano/Pool.hs
```diff
@@ -1,28 +0,0 @@
...
-import Prelude
-
-import Cardano.Wallet.Primitive.Types
-    ( PoolId, StakePool, StakePoolMetadata )
-import Control.Monad.Trans.Except
-    ( ExceptT )
-
--- | @StakePoolLayer@ is a thin layer ontop of the DB. It is /one/ value that
--- can easily be passed to the API-server, where it can be used in a simple way.
-data StakePoolLayer e m = StakePoolLayer
```
Anviking, line 19: Moved to `Cardano.Pool.Jormungandr.Metrics`

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1237,19 +1201,21 @@ joinStakePool
         , ctx ~ ApiLayer s t k
         )
     => ctx
-    -> StakePoolLayer e IO
+    -> IO [PoolId]
+       -- ^ Known pools
+       -- We could maybe replace this with a @IO (PoolId -> Bool)@
```
Anviking, line 1205: No need to have a common `StakePoolLayer` at this point. No need to have a common `StakePool` type in core either.

We just need to make sure the target packages can provide a `Handler (ListStakePool apiPool)`, for some `apiPool`, and can provide this `IO [PoolId]`.

lib/jormungandr/src/Cardano/Pool/Jormungandr/Metrics.hs
```diff
@@ -139,6 +145,48 @@ data Block = Block
...
+--------------------------------------------------------------------------------
+-- Api Handler
+--------------------------------------------------------------------------------
+
+listPools
+    :: LiftHandler e
+    => StakePoolLayer e IO
+    -> Handler [ApiJormungandrStakePool]
+listPools spl =
+    liftHandler $ map (uncurry mkApiJormungandrStakePool) <$> listStakePools spl
```
Anviking: Oh, yeah, I put the API handler here.

Any strong objections or suggestions for a better place?

For the shelley version particularly, I was thinking that a `StakePoolLayer` _might_ be unnecessary indirection, and that we should implement the handler directly.

KtorZ: > Any strong objections or suggestions for a better place?

Yes --> 

```hs
module Cardano.Wallet.Jormungandr.Api.Server
```

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -183,13 +183,16 @@ import Servant.API.Verbs
 
 type ApiV2 n = "v2" :> Api n
 
+-- | The full cardano-wallet API.
+--
+-- The API used in cardano-wallet-jormungandr may differ from this one.
 type Api n =
```
KtorZ, line 189: I assume this one can be fully moved to each target package, alongside the server instantiation.

KtorZ (approved):

lib/core/src/Cardano/Wallet/Api.hs
Anviking: No! Sadly not easily, because we have a swagger spec in core!

My plan is to keep this type, but let jormungandr use a custom `Api` type.

KtorZ: Hmmm. Well no, we can also declare an "Api" type in the relevant test.

lib/jormungandr/src/Cardano/Pool/Jormungandr/Metrics.hs
```diff
@@ -139,6 +145,48 @@ data Block = Block
     -- ^ Any stake pools that were registered in this block.
     } deriving (Eq, Show, Generic)
 
+
+-- | @StakePoolLayer@ is a thin layer ontop of the DB. It is /one/ value that
+-- can easily be passed to the API-server, where it can be used in a simple way.
+data StakePoolLayer e m = StakePoolLayer
```
paweljakubas, line 144: don't we want to have StakePoolLayer interface in `core`? And in jormunagdr and shelley implementations of it?

paweljakubas (approved):

lib/jormungandr/src/Cardano/Pool/Jormungandr/Metrics.hs
Anviking: I believe we/you currently only need the `knownStakePools :: IO [PoolId]` field to check that the pool we're joining actually exists. So I see no point in keeping the entire `StakePoolLayer` then, since it would just be more work trying to unify the different pool-types under one common interface.

lib/core/src/Cardano/Wallet/Api.hs
Anviking: > Well no, we can also declare an "Api" type in the relevant test

That might make our test meaningless since if it is no longer testing the `Api` type that is actually used.

I say we keep the haskell API in `core` as the most important api, and test it, but not the jormungandr one.

…or, we define and test both APIs in core 🤔 

…or, maybe one could do something with a `ApiStakePool | ApiJormungandrStakePool` sum type 🤔

lib/jormungandr/src/Cardano/Pool/Jormungandr/Metrics.hs
paweljakubas: a ok, got it . so now we will sneak `IO [PoolId]` instead of `StakePoolLayer` on wallet level. and just call it (with liftIO). and this will be `knownPools` type argument...

Anviking: Yeah (see diff at https://github.com/input-output-hk/cardano-wallet/pull/1738#discussion_r437978274)

I was curious about using a `IO (PoolId -> Bool)` also, but didn't do it.

lib/core/src/Cardano/Wallet/Api.hs
Anviking: Can sort out later.

KtorZ: > …or, maybe one could do something with a ApiStakePool | ApiJormungandrStakePool sum type thinking

Yeah, I thought of it. Since the stake pools are only something the server produces, it might just be okay to go down that path and fairly transparent for clients.

KtorZ:
Hydra will cache the previous build and it'll fail all-the-same.

KtorZ:
bors try

KtorZ:
Twice in a row, the unit tests on Windows aren't passing :thinking: 

```
cardano-wallet-core-test-unit-2020.6.5-x86_64-w64-mingw32-check-x86_64-w64-mingw32.drv
```

KtorZ:
I don't see anything _obviously wrong_ on Windows in the changes from this PR though...

Anviking:
This doesn't make sense, because this is not the actual code:
```
src/Cardano/Wallet/Shelley/Api/Server.hs:23:1: error:
    Could not find module ‘Cardano.Pool’
    Perhaps you meant
      Cardano.Pool.DB (from cardano-wallet-core-2020.6.5)
    Use -v to see a list of the files searched for.
   |
23 | import Cardano.Pool
   | ^^^^^^^^^^^^^^^^^^^...
[2 of 6] Compiling Cardano.Wallet.Shelley.Compatibility ( src/Cardano/Wallet/Shelley/Compatibility.hs, dist/build/Cardano/Wallet/Shelley/Compatibility.o )
[3 of 6] Compiling Cardano.Wallet.Shelley.Network ( src/Cardano/Wallet/Shelley/Network.hs, dist/build/Cardano/Wallet/Shelley/Network.o )
[4 of 6] Compiling Cardano.Wallet.Shelley.Transaction ( src/Cardano/Wallet/Shelley/Transaction.hs, dist/build/Cardano/Wallet/Shelley/Transaction.o )
builder for '/nix/store/q9jxhhslcw0j4f4aqdkn3xpvyyx0y9z6-cardano-wa
```
The build failure can't be cached, since master has changed?

Anviking:
Ok, it is because of the merge 🤔

PR #1739, opened 2020-06-10
Migration tests clean up

piotr-iohk:
# Issue Number

#1675

# Overview

- 78a6a1c93caf529a419f6fb2daf67c090779d1ba
  Extract Migrations for byron to separate scenario file
  
- 64a0416438f3ec110b2e7d203369e88022a4ae21
  Extract Migration tests for byron in core-integration
  
- 96bd585db408c6b96d0b74c510c9ae0479565d3e
  Migration tests for Shelley
  
- 90853c3adff2a69818757692a7c4dbc7cdc8516d
  Remove `pending` from byron migration tests
  
- 5e2f26c20fdc2b65e8cb87db6fe3daa14c40eabc
  Make `SHELLEY_MIGRATE_01_big_wallet` test pending due to issue #1751
  
- 5c25b012946eed57c767abb2dcdfdcbf8fbf6bff
  - Add ByronWallets tests to Shelley integration suite.  
  - Make `Todo comment` which tests to enable when Byron addresses/transactions are supported on the cardano-node.  
  - temporarily move BYRON_RESTORE special cases involving restoring 'funky' ledger and icarus wallets to Transactions so we can run ByronWallet tests on Shelley integration suite

# Comments

Attempting to "clean-up" migration tests and add migration tests for Shelley and Byron.

~I have added some byron (random + icarus) wallets into genesis.yaml, but... It doesn't really work, node crashes... :thinking:~ ... because they are not supported yet.


lib/shelley/test/data/cardano-node-shelley/genesis.yaml
```diff
@@ -42,12 +42,2212 @@ maxKESEvolutions: 90
...
+  82d818582183581ce34112bfc8f87c19318889d3acaaaa8cdff920e5901ddacdf681a3e1a0001ab3429726: 100000000000
+  82d818582183581c821895116b4d437d028efcb0601df97b82bc8ef565803ecbbb2fab0ea0001a7e988014: 100000000000
+  82d818582183581cca4cf32957ac53ecadd6a3dc4ee5a3b8cc5328ad20b3edc9aa1c6474a0001a39729758: 100000000000
+  82d818582183581cb26fe0f0908be9722252a786b3dec40aa74b4987084b5660e38026d8a0001a1bc3fd38: 100000000000
+  82d818582183581c4921db5cb3c34576c37363a4c72c6e49a8e516307891a2e36d54960fa0001ad463779b: 100000000000
+  82d818582183581c410f8fb53e34107eafdc5dd297bc4cd20d245dd153d5c8bed7e7c832a0001a56295cfb: 100000000000
+  82d818582183581ca02e0d8a3c4bede18c1cf8229ddb02769c03ac0dd3776b8c34f17cc8a0001af6579671: 100000000000
+  82d818582183581c963f621368a253a75fc950c25493216ef6543e7dbe5dbe652a8eb978a0001a8b5d87da: 100000000000
+  82d818582183581c812f3071c5c9f9a7ea95d6e98f910173f912097a6a75351322c1c867a0001afd55c87f: 100000000000
+
```
KtorZ: I believe that these would need to be base58-encoded as being "byron addresses". 
Can you also double-check with the node's guys whether Byron addresses as initial genesis funds are supported _yet_ ?

piotr-iohk: Currently when you create a byron wallet (e.g. on FF) the address is:
For random:
`82d818584683581cd7c21027dd1037acc53b9ce171f7f909ba68972b56886c00e7122392a201581e581c427a66d6279d26c4538f97f8df0ce0d98ce5de06cbf45a7ede422b1b0242182a001ac4109f66`

For icarus:
`82d818582583581c4b3a47fcc6ebc7018955105eaf4fa37ca215d084370f299b3be00f4ea10242182a001afe67739c`

Anyway Byron transactions are not supported yet, I learned, so I guess, it's still too early for that...

piotr-iohk: I have removed Byron addresses from genesis for now.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
```diff
@@ -0,0 +1,481 @@
...
+        \after a migration operation successfully completes, the correct \
+        \amount eventually becomes available in the target wallet for arbitrary \
+        \ number of specified addresses. Balance of source wallet = 0."
+        $ do
+              testAddressCycling 1
+              testAddressCycling 3
+              testAddressCycling 10
+
+    it "SHELLEY_MIGRATE_01 - \
+        \ migrate a big wallet requiring more than one tx" $ \ctx -> do
```
piotr-iohk, line 168: This test is failing with:

```
Failures:

  src/Test/Integration/Scenario/API/Shelley/Migrations.hs:221:15: 
  1) API Specifications SHELLEY_MIGRATE_01 -  migrate a big wallet requiring more than one tx
       expected: Status {statusCode = 202, statusMessage = "Accepted"}
        but got: Status {statusCode = 500, statusMessage = "Internal Server Error"}
       
       from the following response: 
Left (DecodeFailure 
"{\"code\":\"created_invalid_transaction\",
  \"message\":\"That's embarrassing. It looks like I've created an invalid transaction that could not be 
    parsed by the node. Here's an error message that may help with debugging: 
    ApplyTxError [LedgerFailure (UtxowFailure (UtxoFailure (MaxTxSizeUTxO 20614 4096)))]\"}")

```

piotr-iohk:
bors try

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
KtorZ: Yes. The max number of inputs is currently hard-coded to `100` in the Shelley transaction layer. We need something a bit better here :)

lib/shelley/test/integration/Main.hs
```diff
@@ -133,21 +135,22 @@ main = withUtf8Encoding $ withLogging Nothing Info $ \(_, tr) -> do
...
-                Network.spec
-                Network_.spec
-            describe "CLI Specifications" $ do
-                AddressesCLI.spec @n
-                TransactionsCLI.spec @n
-                WalletsCLI.spec @n
-                HWWalletsCLI.spec @n
-                PortCLI.spec @t
-                NetworkCLI.spec @t
+        describe "API Specifications" $ specWithServer tr $ do
```
KtorZ: Hey! No! 

There's a wrong merge-conflict resolution here I think ^.^ .. I moved the `specWithServer` above to be around ALL, otherwise, this actually restarts a cluster between both steps. Not only does it takes time, but it also doubles the number of genesis credentials we need for starting stake pools. The `Migrations.spec @n` can be added all the same.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
```diff
@@ -0,0 +1,481 @@
...
+
+spec :: forall n t.
+    ( DecodeAddress n
+    , EncodeAddress n
+    , PaymentAddress n ShelleyKey
+    , PaymentAddress n IcarusKey
+    , PaymentAddress n ByronKey
+    ) => SpecWith (Context t)
+spec = do
+    it "SHELLEY_CALCULATE_01 - \
```
KtorZ, line 97: Do you think we could mimic the "new clean way of writing scenarios" that we have in recently written integration spec files? (i.e., declare scenario as separate function and reference them in the main "spec")

lib/shelley/test/integration/Main.hs
piotr-iohk: oh ok, whatever it was... it was done automatically (I think) :sweat_smile: I'll fix.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
piotr-iohk: To be honest I don't find that new way handy for me. I recognize this is cleaner/more aesthetic maybe:

```
    describe "BYRON_MIGRATE" $ do
        scenario_MIGRATE_01 @n fixtureRandomWallet
        scenario_MIGRATE_02 @n fixtureRandomWallet 1
        scenario_MIGRATE_02 @n fixtureRandomWallet 3
        scenario_MIGRATE_02 @n fixtureRandomWallet 10
        scenario_MIGRATE_02 @n fixtureIcarusWallet 1
        scenario_MIGRATE_02 @n fixtureIcarusWallet 3
        scenario_MIGRATE_02 @n fixtureIcarusWallet 10
```
But you have to scroll down/scroll up to see what particular test is doing. This is ok(ish) when there is small number of scenarios in a file, but when there are many of them it becomes painful (to me).

Ok, maybe that's my personal preference (or I just got used to it so much), but I find good ol `spec -> describe -> it` scheme (+ some helper methods for repeatable actions) a good compromise between "clean" and sufficient level of detail in the (integration) test suite such that I can immediately tell what is going on...

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
piotr-iohk: Shall I make this test pending and create a ticket for that perhaps?

lib/shelley/test/integration/Main.hs
```diff
@@ -139,6 +141,8 @@ main = withUtf8Encoding $ withLogging Nothing Info $ \(_, tr) -> do
         specWithServer tr $ do
             describe "API Specifications" $ do
                 Addresses.spec @n
+                ByronMigrations.spec @n
```
KtorZ: I think it's a bit too soon to enable this one on the shelley node.

KtorZ (approved):

lib/shelley/test/integration/Main.hs
piotr-iohk: They have been enabled (those tests were part of Wallets.spec, but they were made "pending"), but since they are now extracted to separate spec, then maybe they just don't have to be enabled for shelley indeed.

piotr-iohk: I have removed those tests from Shelley suite and also added ByronWallet.spec. In https://github.com/input-output-hk/cardano-wallet/pull/1739/commits/5c25b012946eed57c767abb2dcdfdcbf8fbf6bff.

piotr-iohk:
bors try

piotr-iohk:
bors try

PR #1744, opened 2020-06-11
Add api swagger spec for shelley stake pools

Anviking:
# Issue Number

#1718 

# Overview

- [x] Added swagger spec for shelley pools
- [x] The entire Api type is now parameterised over `apiPool`
- [x] ⚠️ Breaking: Make both the haskell and jormungadnr API require a wallet id to list stake pools

# Comments

### DIFF
--- a/lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
+++ b/lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
@@ -51,7 +51,7 @@ import Cardano.Wallet.Api.Malformed
 import Cardano.Wallet.Api.Server
     ( LiftHandler (..) )
 import Cardano.Wallet.Api.Types
-    ( DecodeAddress (..), EncodeAddress (..) )
+    ( ApiStakePool, DecodeAddress (..), EncodeAddress (..) )
 import Cardano.Wallet.Primitive.AddressDerivation
     ( NetworkDiscriminant (..) )
 import Cardano.Wallet.Primitive.Types
@@ -303,10 +303,11 @@ application :: Application
 application = serve api server
     & handleRawError (curry handler)
 
-api :: Proxy (Api ('Testnet 0))
-api = Proxy @(Api ('Testnet 0))
+-- Note: Doesn't validate the jormungandr api.
+api :: Proxy (Api ('Testnet 0) ApiStakePool)
+api = Proxy
 
-server :: Server (Api ('Testnet 0))
+server :: Server (Api ('Testnet 0) ApiStakePool)
 server = error
     "No test from this module should actually reach handlers of the server. \
     \Tests are indeed all testing the internal machinery of Servant + Wai and \
--- a/specifications/api/swagger.yaml
+++ b/specifications/api/swagger.yaml
@@ -492,17 +492,6 @@ x-transactionStatus: &transactionStatus
     - pending
     - in_ledger
 
-x-stakePoolMetrics: &stakePoolMetrics
-  type: object
-  required:
-    - controlled_stake
-    - produced_blocks
-  properties:
-    controlled_stake: *transactionAmount
-    produced_blocks:
-      <<: *numberOfBlocks
-      description: Number of blocks produced by a given stake pool
-
 x-stakePoolApparentPerformance: &stakePoolApparentPerformance
   description: |
     Apparent performance of the stake pool over past epochs. This indicator is computed
@@ -588,6 +577,45 @@ x-stakePoolDesirability: &stakePoolDesirability
     How desirable / attractive a pool is. To determine a pool's _rank_, we order pools by decreasing desirability.
     The most desirable pool gets rank 1, the second most desirable pool gets rank 2 and so on.
   example: 42
+
+x-non-myopic-member-rewards: &nonMyopicMemberRewards
+  <<: *amount
+  description: |
+    The rewards the wallet can expect to receive at the end of an epoch, in the long term, if delegating to
+    this pool.
+
+    For more details, see the
+    [Design Specification for Delegation and Incentives in Cardano](https://hydra.iohk.io/job/Cardano/cardano-ledger-specs/delegationDesignSpec/latest/download-by-type/doc-pdf/delegation_design_spec)
+    document.
+
+x-stakePoolMetrics: &stakePoolMetrics
+  type: object
+  required:
+    - relative_stake
+    - non_myopic_member_rewards
+    - produced_blocks
+    - saturation
+  properties:
+    non_myopic_member_rewards: *nonMyopicMemberRewards
+    relative_stake:
+      <<: *percentage
+      description: The pools stake compared to the stake of all pools.
+    saturation: *stakePoolSaturation
+    produced_blocks:
+      <<: *numberOfBlocks
+      description: Number of blocks produced by a given stake pool
+
+x-jormungandrStakePoolMetrics: &jormungandrStakePoolMetrics
+  type: object
+  required:
+    - controlled_stake
+    - produced_blocks
+  properties:
+    controlled_stake: *transactionAmount
+    produced_blocks:
+      <<: *numberOfBlocks
+      description: Number of blocks produced by a given stake pool
+
 
 x-networkInformationSyncProgress: &networkInformationSyncProgress
   <<: *syncProgress
@@ -755,6 +783,19 @@ components:
       type: object
       required:
         - id
+        - cost
+        - margin
+      properties:
+        id: *stakePoolId
+        metrics: *stakePoolMetrics
+        cost: *stakePoolCost
+        margin: *stakePoolMargin
+        metadata: *stakePoolMetadata
+
+    ApiJormungandrStakePool: &ApiJormungandrStakePool
+      type: object
+      required:
+        - id
         - metrics
         - apparent_performance
         - cost
@@ -763,7 +804,7 @@ components:
         - desirability
       properties:
         id: *stakePoolId
-        metrics: *stakePoolMetrics
+        metrics: *jormungandrStakePoolMetrics
         apparent_performance: *stakePoolApparentPerformance
         cost: *stakePoolCost
         margin: *stakePoolMargin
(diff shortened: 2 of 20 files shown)


specifications/api/swagger.yaml
```diff
@@ -752,6 +762,32 @@ components:
...
     ApiStakePool: &ApiStakePool
+      type: object
+      required:
+        - id
+        - relative_stake
+        - non_myopic_member_rewards
+        - produced_blocks
+        - apparent_performance
+        - cost
+        - margin
```
Anviking, line 787: On jormungandr we have a `metrics` object containing `produced_blocks` and `controlled_stake`. I don't see what is so special about those two fields compared to the rest, so I kept everything flat here.

KtorZ: The rationale at the time was that what's in "metrics" was measured, whereas other parameters were simply read from the chain.

Anviking: Then why was not Apparent performance included?

Anviking: …and saturation, and desirability?

specifications/api/swagger.yaml
```diff
@@ -1863,17 +1903,23 @@ paths:
...
       description: |
         <p align="right">status: <strong>stable</strong></p>
 
-        List all known stake pools ordered by descending `desirability`.
+        List all known stake pools ordered by descending `non_myopic_member_rewards`.
+
         Some pools _may_ also have metadata attached to them if they have been registered
         to the metadata registry.
+
+        The `non_myopic_member_rewards` — and thus the ording — depends on the balance of the given wallet.
```
KtorZ: ording -> ordering ?

specifications/api/swagger.yaml
```diff
@@ -1863,17 +1903,23 @@ paths:
...
       tags: ["Stake Pools"]
       summary: List
       description: |
         <p align="right">status: <strong>stable</strong></p>
 
-        List all known stake pools ordered by descending `desirability`.
+        List all known stake pools ordered by descending `non_myopic_member_rewards`.
+
         Some pools _may_ also have metadata attached to them if they have been registered
         to the metadata registry.
```
KtorZ: What about cutting out the last part of the sentence after if?

specifications/api/swagger.yaml
Anviking: 👍 I should probably have looked at this with spell-check enabled.

specifications/api/swagger.yaml
```diff
@@ -1863,17 +1903,23 @@ paths:
...
+
+        The `non_myopic_member_rewards` — and thus the ording — depends on the balance of the given wallet.
+
+        On cardano-wallet-jormungandr, pools are instead ordered by decending `desirability`, which is a less
```
KtorZ: Let's say instead: "On the incentivized testnet, ..."

specifications/api/swagger.yaml
```diff
@@ -1863,17 +1903,23 @@ paths:
...
+        complete metric compared to `non_myopic_member_rewards`.
```
KtorZ: No need for this last part `which ...`, it feels unnecessary. "pools are instead ordered by descending `desirability`."

specifications/api/swagger.yaml
```diff
@@ -752,6 +761,32 @@ components:
...
+          description: The pools stake compared to the stake of all pools.
+        cost: *stakePoolCost
+        margin: *stakePoolMargin
+        saturation: *stakePoolSaturation
+        non_myopic_member_rewards: *nonMyopicMemberRewards
+        metadata: *stakePoolMetadata
+        produced_blocks:
+          <<: *numberOfBlocks
+          description: Number of blocks produced by a given stake pool.
+        apparent_performance: *stakePoolApparentPerformance
```
KtorZ: Do we actually still have this one on cardano-node ?

specifications/api/swagger.yaml
KtorZ: Because many people modify many parts of the code and many of these fields were added in a rush resulting in an incoherent design.

specifications/api/swagger.yaml
Anviking: Are we going to display some form of `produced_blocks`? Then why not? Though it will be a different apparent performance from that used to calculate the non-myopic rewards.

specifications/api/swagger.yaml
Anviking: ok, then indeed, let's move all relevant fields to `metrics` (for shelley)

specifications/api/swagger.yaml
KtorZ: We also need the pool's stake on the previous epoch, do we have that?

Anviking: I'm not sure. If LSQ worked for k blocks old, then we have stake distributions for the k latest blocks.

Should we remove?

Should we not also remove `produced_blocks`? (Or have say `lifetime_blocks` if we keep it?)

specifications/api/swagger.yaml
Anviking: Sure. But jormungandr and cardano-wallet-jormungandr may be used outside the ITN?

specifications/api/swagger.yaml
```diff
@@ -589,6 +578,46 @@ x-stakePoolDesirability: &stakePoolDesirability
     The most desirable pool gets rank 1, the second most desirable pool gets rank 2 and so on.
   example: 42
 
+x-non-myopic-member-rewards: &nonMyopicMemberRewards
+  <<: *amount
+  description: |
+    Non-myopic member rewards is the recomended way to rank stake pools. It is the rewards the
```
Anviking: Perhaps this is a weird way to describe it.

E.g. for what unit of time do the rewards correspond to? (I think it is an epoch.)

KtorZ: Non-myopic member rewards = rewards one can expect at the end of an epoch from a fully saturated pool.

specifications/api/swagger.yaml
KtorZ: What Jörmungandr is use for is irrelevant. cardano-wallet-jormungandr is only used for the ITN. For Voltaire, we'll see and adjust wording consequently if needed :+1:

specifications/api/swagger.yaml
KtorZ: Ideally we would have both last production and total production. Though we do not depend on the LSQ for these, we have this information from the chain sync data. 

> I'm not sure. If LSQ worked for k blocks old, then we have stake distributions for the k latest blocks. Should we remove?

Well, without the stake distribution of past epochs, we are running into the same problem / awkwardness as during the ITN where we would have holes in our dataset. I'd rather not go through this again.

Anviking: > Well, without the stake distribution of past epochs, we are running into the same problem / awkwardness as during the ITN where we would have holes in our dataset. I'd rather not go through this again.

To be fair: We are _guaranteed_ to have `k` blocks with corresponding stake distributions (except for at genesis). On the ITN we weren't guaranteed to have that.

Anviking: What would the point of  `last_production` be, and what would the window be? Wouldn't it just be to give users some notion of the apparent performance?

Then why not either remove _both_ `last_production` and `apparent_performance`, or keep both? 🤔

specifications/api/swagger.yaml
Anviking: I rephrased in 44e6ff870c7c0805661b425d9ca934b9892a576b 

> epoch from a fully saturated pool

That is misleading though? It is assuming the top k pools are saturated, and the rest are empty (except the user's stake… and pledge?), or something like that. I didn't mention on those specifics in the commit.

specifications/api/swagger.yaml
```diff
@@ -589,6 +578,45 @@ x-stakePoolDesirability: &stakePoolDesirability
     The most desirable pool gets rank 1, the second most desirable pool gets rank 2 and so on.
   example: 42
 
+x-non-myopic-member-rewards: &nonMyopicMemberRewards
+  <<: *amount
+  description: |
+    The rewards the wallet can expect to receive at the end of an epoch, in the long term, if delegating to
+    this pool.
+
+    For more details, see the "Design Specification for Delegation and Incentives in Cardano" document.
```
KtorZ: We could _link_ it here (the description is actually plain markdown)

specifications/api/swagger.yaml
```diff
@@ -1863,17 +1909,23 @@ paths:
...
         <p align="right">status: <strong>stable</strong></p>
 
-        List all known stake pools ordered by descending `desirability`.
-        Some pools _may_ also have metadata attached to them if they have been registered
-        to the metadata registry.
+        List all known stake pools ordered by descending `non_myopic_member_rewards`.
+
+        Some pools _may_ also have metadata attached to them.
+
+        The `non_myopic_member_rewards` — and thus the ordering — depends on the balance of the given wallet.
```
KtorZ: Small note, I'd actually swap these two sentences. So that it we give precision on the ordering after actually mentioning that pools are ordered by descending member rewards.

KtorZ: List all known stake pools ordered by descending `non_myopic_member_rewards`. The `non_myopic_member_rewards` — and thus the ordering — depends on the balance of the given wallet.

> /!\ On the incentivized testnet, pools are instead ordered by descending `desirability`.
        
Some pools _may_ also have metadata attached to them.

specifications/api/swagger.yaml
```diff
@@ -589,6 +578,45 @@ x-stakePoolDesirability: &stakePoolDesirability
...
+  properties:
+    non_myopic_member_rewards: *nonMyopicMemberRewards
+    relative_stake:
+      <<: *percentage
+      description: The pools stake compared to the stake of all pools.
+    saturation: *stakePoolSaturation
+    produced_blocks:
+      <<: *numberOfBlocks
+      description: Number of blocks produced by a given stake pool
+    apparent_performance: *stakePoolApparentPerformance
```
KtorZ: Let's remove that one as discussed.

KtorZ (approved):

paweljakubas (approved):
LGTM!

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
```diff
@@ -299,14 +303,20 @@ instance GenericApiSpec (Map [Text] [Method])
 -- Construct test cases from the API
 --
 
+-- | A This combines the haskell API type with the part of the jormungadnr api
+-- that differs, such that we validate both.
+type CustomApi n =
+    Api n ApiStakePool
+    :<|> ("jormungandr" :> StakePools n ApiJormungandrStakePool)
```
Anviking: @KtorZ I hope you don't mind this

Anviking:
bors try

PR #1746, opened 2020-06-12
encode shelley addresses with their final encoding (bech32 / base58)

KtorZ:
# Issue Number

N/A

# Overview

Still, I kept support for base16 when decoding addresses, so that we don't necessarily break compatibility with cardano-cli.

# Comments


lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -560,31 +573,88 @@ toCardanoTxOut (W.TxOut addr coin) =
...
 
 instance EncodeAddress ('Testnet pm) where
-    encodeAddress = T.decodeUtf8 . hex . W.unAddress
+    encodeAddress = _encodeAddress
 
-_decodeAddress :: Text -> Either TextDecodingError W.Address
-_decodeAddress x = validateWithLedger =<< W.Address <$> fromHex x
+_encodeAddress :: W.Address -> Text
+_encodeAddress (W.Address bytes) =
+    if isJust (CBOR.deserialiseCbor CBOR.decodeAddressPayload bytes)
```
Anviking, line 680: A golden here would be nice; I don't think a mistake here would be caught

Anviking (approved):
👍

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
KtorZ: I agree. I intended to add some tests, hence the _draft_ :+1:

KtorZ:
bors try

Issue #1709, opened 2020-06-02
Implement join/quitStakePool and delegation fee in Shelley


paweljakubas:
# Context

ADP-157

We would like to have an ability to join, and quit stake pool and check delegation fee in cardano-wallet-shelley.

# Decision

We will use `deriveRewardAccount` to get sks and then vsk (public key of sks) and treat it as stake key registration certificate. Then we will try to reuse mkUnsignedTx to make certificate registration and joining pool. We will not expose registrate/deregistrate certificate to the user. In detail, we will have three situations : 
1. User wants to join the pool (and is not registered) - so we register him and join at once.
2. User want to rejoin - here tx will have join component only (or we can in the first approach send both registration and joining as in point 1 - even if it means additional costs)
3. User want to quit - here tx will have deregister component only
 In this task we are interested in point 1 and 2. Point 3 is going to be next task within the milestone.
4. User want to check the cost of delegation

# Acceptance Criteria

1. `joinStakePool` endpoint works
2. `quitStakePool` endpoint works
3. `delegationFee` endpoint works
2. Integration tests illustrating the correctness are added

---

# Development 

# QA 

see endpoint in shelley plus integration tests added in (please notice that at the time of doing listStakePools was not implemented and we hardcoded pool id from cluster associated with integration tests) `lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs`

piotr-iohk:
@paweljakubas I am not able to join stake-pool that is on Shelley testnet. They all seem to give me 404.

To reproduce:
1. start cardano-node + cardano-wallet on testnet.
e.g.:
cardano-node:
```
nix-build -A scripts.ff.node -o ff-node-local && ./ff-node-local
```

cardano-wallet-shelley:
```
cardano-wallet-shelley serve --port 8090 \
	--node-socket /path/to/node.socket  \
	--testnet /nix/store/v8lb35176p73f6ksp7p5rhkqmygi1zj9-ff-genesis.json \
        --database ./testnet-ff
```

2. Query stake pools registered on testnet:
```
CARDANO_NODE_SOCKET_PATH=/path/to/node.socket cardano-cli shelley query stake-distribution --testnet-magic 42
```
:point_down: 
```
                           PoolId                                 Stake frac
------------------------------------------------------------------------------
0087ac2efc783ae2aff9edc9ddb7f4a1d892ecb2041c12d30dc21b6c2971c3c9   3.322e-3
[...]

PR #1747, opened 2020-06-12
Delegation fee shelley

paweljakubas:
# Issue Number

#1709 

# Overview

- [x] I have enabled delegation fee in Shelley server
- [x] I have ported integration tests from jormungandr
- [x] I have adjusted feeMin

# Comments


rvl (approved):
Looks good

KtorZ:
@paweljakubas I don't think we can have such assertions with cardano-node and the way we do fee estimations there. What we can say however is that real fee should be at most what the `minimumFee` from our estimator return.

Issue #1750, opened 2020-06-12
Get reward account balance


rvl:
# Context

Jira ticket: [ADP-302](https://jira.iohk.io/browse/ADP-302)

### User Story

> As a user, I want to check the rewards I have currently accumulated
>
> Given my wallet is in sync
>
> And I have the public key
> 
> Or I have the mnemonic 
>
> Then I can check the current delegation certificates

### Overview

The delegation rewards balance is shown as part of the shelley wallet balance in the API.

This task is to implement the function in the shelley network layer which queries the cardano-node backend for the reward account balance.

There is some overlap with [ADP-301 Check current delegation preference](https://jira.iohk.io/browse/ADP-301) because the same cardano-node query method is used.

# Decision

The wallet rewards address is already known. It is passed as the parameter to this network layer method.

```haskell
getAccountBalance
    :: NetworkLayer IO (IO Shelley) ShelleyBlock
    -> ChimericAccount
    -> ExceptT ErrGetAccountBalance m (Quantity "lovelace" Word64)
```

The shelley implementation of this method should use the node local state query protocol to get the stake address info.

A model for this could be the `cardano-cli shelley query stake-address-info` command. Its implementation is in `Cardano.Api.LocalStateQuery.queryDelegationsAndRewardsFromLocalState` (`cardano-node` repo / `cardano-api` package).

The `ChimericAccount` address must be converted to a shelley stake credential.

In the shelley network layer, create a queue of `LocalStateQueryCmd ShelleyBlock GetFilteredDelegationsAndRewardAccounts`

```haskell
-- cardano-wallet-shelley
data LocalStateQueryCmd block state (m :: * -> *)
    = CmdQueryLocalState
        (Point block)
        (Query block state)
        (LocalStateQueryResult state -> m ())

-- ouroboros-consensus-shelley
-- Ouroboros.Consensus.Shelley.Ledger.Ledger
instance TPraosCrypto c => QueryLedger (ShelleyBlock c) where
  data Query (ShelleyBlock c) :: Type -> Type where

    -- ...

    GetFilteredDelegationsAndRewardAccounts
      :: Set (SL.Credential 'SL.Staking c)
      -> Query (ShelleyBlock c) (Delegations c, SL.RewardAccounts c)
```

In `mkWalletClient`, add a local state query protocol which handles the reward account query commands. The block queried is tip of the wallet's chain sync protocol.

# Acceptance Criteria

[...]

PR #1752, opened 2020-06-12
Shelley get reward account balance

rvl:
### Issue Number

ADP-302 / #1750 

### Overview

Adds implementation of network layer `getAccountBalance` for shelley.

### Comments

The integration test is not passing because stake pools are not connected with the test cluster network.

### DIFF
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
@@ -38,17 +38,27 @@ import Cardano.BM.Data.Severity
 import Cardano.BM.Data.Tracer
     ( HasPrivacyAnnotation (..), HasSeverityAnnotation (..) )
 import Cardano.Wallet.Network
-    ( Cursor, ErrPostTx (..), NetworkLayer (..), mapCursor )
+    ( Cursor
+    , ErrGetAccountBalance (..)
+    , ErrNetworkUnavailable (..)
+    , ErrPostTx (..)
+    , NetworkLayer (..)
+    , mapCursor
+    )
 import Cardano.Wallet.Shelley.Compatibility
-    ( Shelley
+    ( Delegations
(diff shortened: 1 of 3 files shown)


paweljakubas (approved):
great job! just make the test pending before we can run it, also change `Delegations` type alias a little bit using the type aliases from cardano-api (as we use them below):
```
type ShelleyCredentialStaking            = Shelley.Credential Shelley.Staking Shelley.TPraosStandardCrypto
type ShelleyVerificationKeyHashStakePool = Shelley.KeyHash Shelley.StakePool Shelley.TPraosStandardCrypto
```
(https://github.com/input-output-hk/cardano-node/blob/master/cardano-api/src/Cardano/Api/Types.hs#L144 and https://github.com/input-output-hk/cardano-node/blob/master/cardano-api/src/Cardano/Api/Types.hs#L153)

Anviking (approved):
lgtm

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -206,43 +218,57 @@ withNetworkLayer
...
+        connectNodeTipClient handlers
+
+    queryRewardQ <- connectDelegationRewardsClient handlers
 
     action
         NetworkLayer
             { currentNodeTip = liftIO $ _currentNodeTip nodeTipVar
             , nextBlocks = _nextBlocks
-            , initCursor = _initCursor
+            , initCursor = _initCursor queryRewardQ
```
Anviking, line 238: Wait, why is `queryRewardQ` both read from by the shared `connectDelegationRewardsClient` and each wallet client? That seems odd. Shouldn't `queryRewardQ` _only_ be handled by the shared client?

_trying removing it on my branch_

rvl: Oops. I meant to remove that.

PR #1753, opened 2020-06-12
List stake pools

### LINKED ISSUE #1718: List id and stake of pools

### LINKED ISSUE #1720: Order stake pools by non-myopic member rewards
# Context

We want to implement this endpoint for (the cardano-node-) Shelley.
https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/listStakePools

To help users chose between pools, the endpoint, which was added during the ITN, used to provide:
- `apparent_performance`
- `desirability`
- (and `saturation`)

According to the delegation design spec pools should be sorted by non-myopic member rewards.

The node's has a local state query for non-myopic member rewards which we can use.
# Decision

We should:
- Order stake pools by non-myopic member rewards
- Add non-myopic member rewards to the API response

And then:
- Remove the fields `apparent_performance`, `desirability`, `saturation`?

# Acceptance Criteria

- `GET /stake-pools` must be ordered by non-myopic member rewards
- `GET /stake-pools` should contain non-myopic member rewards in the response

---

# Development 

# QA 

- Integration test https://github.com/input-output-hk/cardano-wallet/pull/1783 that rewards are non-zero and pools are order descending by rewards in the integration test setup.
- Fetching of non-myopic rewards https://github.com/input-output-hk/cardano-wallet/pull/1769 (the relevant logic on the wallet-side is pretty trivial)
- Tested manually on testnet

Anviking:
# Issue Number

#1718, #1720

# Overview

- [x] Allow heterogeneous queues of LSQ queries such that we can query for stake, rewards, and pparams using the same queue.
- [x] Retrieve pool id, stake and saturation and make the API return it
- [x] Stubbed values for non-myopic member rewards, margin, cost, produced blocks

# Comments

- Actually fetching non-myopic member rewards depends on a consensus change reaching cardano-node.

- Working chain following may be a non-goal for this PR, but trying to make room for it.

### DIFF
--- a/lib/jormungandr/src/Cardano/Pool/Jormungandr/Metrics.hs
+++ b/lib/jormungandr/src/Cardano/Pool/Jormungandr/Metrics.hs
@@ -9,6 +9,7 @@
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TupleSections #-}
+{-# LANGUAGE TypeFamilies #-}
 
 -- | This module can fold over a blockchain to collect metrics about
 -- Stake pools.
@@ -54,12 +55,17 @@ import Cardano.Pool.Jormungandr.Performance
     ( readPoolsPerformances )
 import Cardano.Pool.Jormungandr.Ranking
     ( EpochConstants (..), unsafeMkNonNegative )
+import Cardano.Wallet.Jormungandr.Compatibility
+    ( Jormungandr )
+import Cardano.Wallet.Jormungandr.Network
+    ()
 import Cardano.Wallet.Network
     ( ErrCurrentNodeTip
     , ErrNetworkUnavailable
     , FollowAction (..)
     , FollowExit (..)
     , FollowLog
+    , GetStakeDistribution
     , NetworkLayer (currentNodeTip, stakeDistribution)
     , follow
     )
@@ -163,7 +169,8 @@ data StakePoolLayer e m = StakePoolLayer
 -- The pool productions and stake distrubtions in the db can /never/ be from
 -- different forks such that it's safe for readers to access it.
 monitorStakePools
-    :: Tracer IO StakePoolLog
+    :: GetStakeDistribution t IO ~ GetStakeDistribution Jormungandr IO
+    => Tracer IO StakePoolLog
     -> (Block, Quantity "block" Word32)
         -- ^ Genesis block and 'k'
     -> NetworkLayer IO t Block
@@ -201,7 +208,7 @@ monitorStakePools tr (block0, Quantity k) nl db@DBLayer{..} = do
         distributions <- forM epochs $ \ep -> do
             liftIO $ traceWith tr $ MsgStakeDistribution ep
             withExceptT ErrMonitorStakePoolsNetworkUnavailable $
-                (ep,) <$> stakeDistribution nl ep
+                (ep,) <$> (stakeDistribution nl ep)
 
         currentTip <- withExceptT ErrMonitorStakePoolsCurrentNodeTip $
             currentNodeTip nl
@@ -253,7 +260,7 @@ newStakePoolLayer
     -- ^ Genesis block header
     -> (EpochNo -> Quantity "lovelace" Word64 -> EpochConstants)
     -> DBLayer IO
-    -> NetworkLayer IO t Block
+    -> NetworkLayer IO Jormungandr Block
     -> FilePath
     -- ^ A directory to cache downloaded stake pool metadata. Will be created if
     -- it does not exist.
--- a/lib/shelley/src/Cardano/Wallet/Shelley.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley.hs
@@ -57,7 +57,7 @@ import Cardano.DB.Sqlite
 import Cardano.Wallet
     ( WalletLog )
 import Cardano.Wallet.Api
-    ( ApiLayer, ApiV2 )
+    ( ApiLayer, ApiV2, ListStakePools )
 import Cardano.Wallet.Api.Server
     ( HostPreference, Listen (..), ListenError (..), TlsConfiguration )
 import Cardano.Wallet.Api.Types
@@ -94,6 +94,7 @@ import Cardano.Wallet.Primitive.Types
     ( Address
     , Block
     , ChimericAccount
+    , Coin (..)
     , GenesisParameters (..)
     , NetworkParameters (..)
     , PoolId
@@ -109,12 +110,12 @@ import Cardano.Wallet.Shelley.Compatibility
     ( Shelley, ShelleyBlock, fromNetworkMagic, fromShelleyBlock )
 import Cardano.Wallet.Shelley.Network
     ( NetworkLayerLog, withNetworkLayer )
+import Cardano.Wallet.Shelley.Pools
+    ( StakePoolLayer (..), newStakePoolLayer )
 import Cardano.Wallet.Shelley.Transaction
     ( newTransactionLayer )
 import Cardano.Wallet.Transaction
     ( TransactionLayer )
-import Cardano.Wallet.Unsafe
-    ( unsafeFromHex )
 import Control.Applicative
     ( Const (..) )
 import Control.Tracer
@@ -141,6 +142,8 @@ import Ouroboros.Network.CodecCBORTerm
     ( CodecCBORTerm )
 import Ouroboros.Network.NodeToClient
     ( NodeToClientVersionData (..) )
+import Servant
+    ( Server )
 import System.Exit
     ( ExitCode (..) )
 import System.IOManager
@@ -230,7 +233,17 @@ serveWallet
                 randomApi <- apiLayer (newTransactionLayer proxy pm el) nl
                 icarusApi  <- apiLayer (newTransactionLayer proxy pm el ) nl
                 shelleyApi <- apiLayer (newTransactionLayer proxy pm el) nl
-                startServer proxy socket randomApi icarusApi shelleyApi mockKnownPools ntpClient
+                let spl = newStakePoolLayer (genesisParameters np) nl
+                startServer
+                    proxy
+                    socket
+                    randomApi
+                    icarusApi
+                    shelleyApi
+                    (\_wid -> listStakePools spl (Coin 0))
+                    -- TODO: read wallt balance
+                    (knownPools spl)
+                    ntpClient
                 pure ExitSuccess
 
     networkDiscriminantValFromProxy
@@ -253,15 +266,16 @@ serveWallet
         -> ApiLayer (RndState n) t ByronKey
         -> ApiLayer (SeqState n IcarusKey) t IcarusKey
         -> ApiLayer (SeqState n ShelleyKey) t ShelleyKey
+        -> Server (ListStakePools ApiStakePool)
         -> IO [PoolId]
         -> NtpClient
         -> IO ()
-    startServer _proxy socket byron icarus shelley spl ntp = do
+    startServer _proxy socket byron icarus shelley listPoolsServer kp ntp = do
         sockAddr <- getSocketName socket
         let settings = Warp.defaultSettings & setBeforeMainLoop
                 (beforeMainLoop sockAddr)
         let application = Server.serve (Proxy @(ApiV2 n ApiStakePool)) $
-                server byron icarus shelley spl ntp
+                server byron icarus shelley kp listPoolsServer ntp
         Server.start settings apiServerTracer tlsConfig socket application
 
     apiLayer
@@ -303,19 +317,6 @@ exitCodeApiServer = \case
     ListenErrorInvalidAddress _ -> 11
     ListenErrorAddressAlreadyInUse _ -> 12
     ListenErrorOperationNotPermitted -> 13
-
--- | FIXME: Temporary mock stake pool layer until we can get the stake pool
--- listing working. These IDs match hard-wired operator credentials in our
--- integration setup. See 'Cardano.Wallet.Shelley.Launch'.
-mockKnownPools :: IO [PoolId]
-mockKnownPools = pure
-        [ PoolId $ unsafeFromHex
-            "5a7b67c7dcfa8c4c25796bea05bcdfca01590c8c7612cc537c97012bed0dec35"
-        , PoolId $ unsafeFromHex
-            "775af3b22eff9ff53a0bdd3ac6f8e1c5013ab68445768c476ccfc1e1c6b629b4"
-        , PoolId $ unsafeFromHex
-            "c7258ccc42a43b653aaf2f80dde3120df124ebc3a79353eed782267f78d04739"
-        ]
 
 {-------------------------------------------------------------------------------
                                     Logging
--- a/lib/shelley/src/Cardano/Wallet/Shelley.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley.hs
@@ -57,7 +57,7 @@ import Cardano.DB.Sqlite
 import Cardano.Wallet
     ( WalletLog )
 import Cardano.Wallet.Api
-    ( ApiLayer, ApiV2, ListStakePools )
+    ( ApiLayer, ApiV2 )
 import Cardano.Wallet.Api.Server
     ( HostPreference, Listen (..), ListenError (..), TlsConfiguration )
 import Cardano.Wallet.Api.Types
@@ -94,11 +94,8 @@ import Cardano.Wallet.Primitive.Types
     ( Address
     , Block
     , ChimericAccount
-    , Coin (..)
     , GenesisParameters (..)
     , NetworkParameters (..)
-    , PoolId
-    , PoolId (..)
     , SyncTolerance
     , WalletId
     )
@@ -142,8 +139,6 @@ import Ouroboros.Network.CodecCBORTerm
     ( CodecCBORTerm )
 import Ouroboros.Network.NodeToClient
     ( NodeToClientVersionData (..) )
-import Servant
-    ( Server )
 import System.Exit
     ( ExitCode (..) )
 import System.IOManager
@@ -240,9 +235,7 @@ serveWallet
                     randomApi
                     icarusApi
                     shelleyApi
-                    (\_wid -> listStakePools spl (Coin 0))
-                    -- TODO: read wallt balance
-                    (knownPools spl)
+                    spl
                     ntpClient
                 pure ExitSuccess
 
@@ -266,16 +259,15 @@ serveWallet
         -> ApiLayer (RndState n) t ByronKey
         -> ApiLayer (SeqState n IcarusKey) t IcarusKey
         -> ApiLayer (SeqState n ShelleyKey) t ShelleyKey
-        -> Server (ListStakePools ApiStakePool)
-        -> IO [PoolId]
+        -> StakePoolLayer
         -> NtpClient
         -> IO ()
-    startServer _proxy socket byron icarus shelley listPoolsServer kp ntp = do
+    startServer _proxy socket byron icarus shelley spl ntp = do
         sockAddr <- getSocketName socket
         let settings = Warp.defaultSettings & setBeforeMainLoop
                 (beforeMainLoop sockAddr)
         let application = Server.serve (Proxy @(ApiV2 n ApiStakePool)) $
-                server byron icarus shelley kp listPoolsServer ntp
+                server byron icarus shelley spl ntp
         Server.start settings apiServerTracer tlsConfig socket application
 
     apiLayer
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
@@ -37,6 +37,7 @@ import Cardano.Wallet.Api
     , ByronTransactions
     , ByronWallets
     , CoinSelections
+    , ListStakePools
     , Network
     , Proxy_
     , ShelleyMigrations
@@ -116,9 +117,8 @@ import Fmt
 import Network.Ntp
     ( NtpClient )
 import Servant
-    ( (:<|>) (..), Server, err501, throwError )
+    ( (:<|>) (..), Server )
 
--- | A diminished servant server to serve Byron wallets only.
 server
     :: forall t n.
         ( Buildable (ErrValidateSelection t)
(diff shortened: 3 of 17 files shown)


Anviking:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -433,7 +423,6 @@ codecs = clientCodecs ShelleyCodecConfig ShelleyNodeToClientVersion1
 serialisedCodecs :: MonadST m => DefaultCodecs ShelleyBlock m
 serialisedCodecs = defaultCodecs ShelleyCodecConfig ShelleyNodeToClientVersion1
 
--- | Construct a network client with the given communication channel, for the
```
Anviking: Oops, mistake from rebasing.

rvl (commented):
I'm not sure if you would like a review yet, but I see that you are using the `Cursor` fields to access the `lsqQ`. Perhaps we would just need another `NetworkLayer` method for this. Because for Jormungandr, we needed to track pools, but for cardano-node, we can simply query.

Anviking:
> Perhaps we would just need another NetworkLayer method for this.

The `NetworkLayer` is shared between jormungandr and haskell, and I don't see the point in trying to generalise over the two where they are different. We'd need to move a bunch of Shelley-specific types to core, and we'd need to add `error "not implemented"` on the jormungandr side.

> we needed to track pools

We are still going to track pool registration certificates. It could perhaps be useful to allow the LSQ to fetch metrics from the chain-follower tip, instead of the NetworkLayer's currentNodeTip, to guarantee consistency — but just a guess.

KtorZ:
> to allow the LSQ to fetch metrics from the chain-follower tip, 

I don't think this is generally possible. The LSQ has only a certain history of the past. It'd be nice if we could move the chain sync cursor forward, fetch some ledger state data, move forward, and so forth. 
Yet in practice, they are separate protocols and they don't really share state here. Thus, unless your chain sync tip is within 2k from the node's tip, you won't be able to fetch corresponding ledger data.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -513,6 +523,39 @@ fromNetworkMagic :: NetworkMagic -> W.ProtocolMagic
...
+
+-- TODO: Change to return a map of maps, instead of using head
+fromRewards
+    :: O.NonMyopicMemberRewards TPraosStandardCrypto
+    -> Map W.PoolId (Quantity "lovelace" Word64)
+fromRewards =
+    Map.map (Quantity . fromIntegral)
+    . Map.mapKeys fromPoolId
+    . snd
+    . head
```
Anviking: Not sure what we want to do here. We are currently only requesting rewards for one wallet at a time, so in a way perhaps we should keep it (albite with a better error than `head`)

Anviking:
bors try

lib/shelley/src/Cardano/Wallet/Shelley.hs
```diff
@@ -230,7 +233,17 @@ serveWallet
...
                 shelleyApi <- apiLayer (newTransactionLayer proxy pm el) nl
-                startServer proxy socket randomApi icarusApi shelleyApi mockKnownPools ntpClient
+                let spl = newStakePoolLayer (genesisParameters np) nl
+                startServer
+                    proxy
+                    socket
+                    randomApi
+                    icarusApi
+                    shelleyApi
+                    (\_wid -> listStakePools spl (Coin 0))
```
Anviking: Ok, this could certainly be filled in now

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
```diff
@@ -130,9 +130,10 @@ server
     -> ApiLayer (SeqState n IcarusKey) t IcarusKey
     -> ApiLayer (SeqState n ShelleyKey) t ShelleyKey
     -> IO [PoolId]
+    -> Server (ListStakePools ApiStakePool)
```
KtorZ: This is actually more of an `Handler` than a `Server`.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
KtorZ: I agree that having `head` _here_ feels kinda wrong. If anything, the caller of `fromRewards` should make the decision about how to treat the result but this function `fromRewards` should simply do the type translation, nothing more.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -513,6 +523,39 @@ fromNetworkMagic :: NetworkMagic -> W.ProtocolMagic
...
+    . Map.toList
+    . O.unNonMyopicMemberRewards
+
+optimumNumberOfPools :: SL.PParams -> Int
+optimumNumberOfPools = safeConvert . SL._nOpt
+  where
+    safeConvert :: Natural -> Int
+    safeConvert = fromIntegral
```
KtorZ: Is it though :wink: 

```
λ> safeConvert 9223372036854775808
-9223372036854775808
```

KtorZ: I don't think going from an unbounded type to a bounded type will ever be "safe". Although realistic values of _optinumNumberOfPools_ are most likely within bounds.

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
Anviking: I'm considering using `StakePoolLayer` instead, now that the shape of it has settled 🤔

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -549,6 +592,12 @@ fromShelleyCoin (SL.Coin c) = W.Coin $ unsafeCast c
     unsafeCast :: Integer -> Word64
     unsafeCast = fromIntegral
 
+toShelleyCoin :: W.Coin -> SL.Coin
+toShelleyCoin (W.Coin c) = SL.Coin $ safeCast c
+  where
+    safeCast :: Word64 -> Integer
+    safeCast = fromIntegral
```
KtorZ, line 599: This one sounds safe indeed :)

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
Anviking: Yes, this should be labeled "unsafeConvert" (perhaps with a comment). Not sure how I managed to get to this.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -0,0 +1,202 @@
...
+data ErrFetchMetrics = ErrFetchMetrics
+  deriving Show
+
+askNode
+    :: MonadSTM m
+    => TQueue m (LocalStateQueryCmd ShelleyBlock m)
+    -> Point ShelleyBlock
+    -> Coin
+    -> ExceptT ErrFetchMetrics m (Map PoolId PoolLsqMetrics)
+askNode queue pt coin = do
```
KtorZ: It's good to avoid _too general_ verbs like "ask" , "do", "run" etc .. in function names because they are often meaningless. `fetchMetrics` sounds a bit more transparent.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -0,0 +1,202 @@
...
+import qualified Cardano.Wallet.Api.Types as Api
+import qualified Data.Map.Merge.Strict as Map
+import qualified Data.Map.Strict as Map
+import qualified Data.Set as Set
+import qualified Ouroboros.Consensus.Shelley.Ledger as OC
+
+-- | Stake Pool Data fields fetched from the node via LSQ
+data PoolLsqMetrics = PoolLsqMetrics
+    { nonMyopicMemberRewards :: Quantity "lovelace" Word64
+    , stake :: Percentage
```
KtorZ: perhaps: `relativeStake`

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -0,0 +1,202 @@
...
+        -> Map PoolId PoolLsqMetrics
+    combine nOpt =
+        Map.merge stakeButNoRew rewardsButNoStake bothPresent
+      where
+        -- calculate the saturation from the relative stake
+        sat s = fromRational $ (getPercentage s) / (1 / fromIntegral nOpt)
+
+        -- Haven't figured out how to fetch non-myopic member rewards properly yet.
+        -- Let's provide a default value of 0, at least for now.
+        stakeButNoRew     = traverseMissing $ \_k s -> pure $ PoolLsqMetrics
```
KtorZ: NoRew? :thinking:

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -0,0 +1,202 @@
...
+    combine
+        :: Int -- ^ Desired number of pools
+        -> Map PoolId Percentage
+        -> Map PoolId (Quantity "lovelace" Word64)
+        -> Map PoolId PoolLsqMetrics
+    combine nOpt =
+        Map.merge stakeButNoRew rewardsButNoStake bothPresent
+      where
+        -- calculate the saturation from the relative stake
+        sat s = fromRational $ (getPercentage s) / (1 / fromIntegral nOpt)
```
KtorZ: :+1:

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -0,0 +1,202 @@
...
+  where
+    handleQueryFailure = withExceptT (const ErrFetchMetrics) . ExceptT
+
+    combine
+        :: Int -- ^ Desired number of pools
+        -> Map PoolId Percentage
+        -> Map PoolId (Quantity "lovelace" Word64)
+        -> Map PoolId PoolLsqMetrics
+    combine nOpt =
+        Map.merge stakeButNoRew rewardsButNoStake bothPresent
```
KtorZ: How is it actually possible with the LSQ to have cases where a poolId wouldn't be in both maps? From the excerpt above, I get that these are two separate requests, but the local state query allows for "locking" a state and making several queries on that locked state if I recall correctly? So in principle, that's sort of a guarantee that the list of pools fetched from the stake distribution is exactly the same as the list of pools get from the member rewards.

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
Anviking: Done in cdc4ad7

lib/shelley/src/Cardano/Wallet/Shelley.hs
Anviking: Done in cdc4ad7

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
Anviking: b36cfe6bebe49b961356d60cd36196b47a9b901e
(But still treating this as an `error`)

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
Anviking: 5f93fcd9cee368ea11069a0337051f8894558b45

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
Anviking: 05256b73f

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
Anviking: b65293b4c54fc298b10f2adcfdcde61bc5dec4e1

Went with `fetchLsqPoolMetrics`. Want to highlight that this just fetches part of the data we want with the `Lsq` part.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
Anviking: c6ddbea6415ea12ef38be1a43a1555a7a17e900e `Rew` -> `Rewards`.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
Anviking: The poolId may not be in the rewards map (clarified comment in cacff5590ed5f77ba6aa0d5e6420e968733f1f99)

As for the `rewardsButNoStake` case: I _suspect_ that is impossible. (Maybe @rvl knows for sure though?)

I guess we should have some `InconcistentMetrics` similar to what we had for jormungandr, instead of the current silent `dropMissing`.

We will have more problems like this when we start folding over pool registration certificates on the chain.

Anviking: > Yes, I suspect that is impossible

But I guess regardless of locking or not, we are making the queries with the same specific `Point`, so they should all be consistent if they succeed 🤔

KtorZ: > I guess we should have some InconcistentMetrics similar to what we had for jormungandr, instead of the current silent dropMissing.

Not really though. In the case of Jörmungandr, we knew we could end up with inconsistent metrics because of the concurrent nature of our logic + the stateless API from Jörmungandr. So it was kind of _expected_. With cardano-node here, this is really something we do not expect, and therefore,not something we're willing to push onto the caller of this function.

specifications/api/swagger.yaml
```diff
@@ -1927,7 +1927,9 @@ paths:
         Some pools _may_ also have metadata attached to them.
       parameters:
         - *parametersWalletId
-      responses: *responsesListStakePools
+      responses:
+        - *responsesListStakePools
+        - *responsesErr404
```
KtorZ: Instead, add `responsesErr404` to the possible responses of `responsesListStakePools`:

```yaml
x-responsesListStakePools: &responsesListStakePools
  <<: *responsesErr404
  <<: *responsesErr405
  200:
    description: Ok
    content:
      application/json:
        schema:
          oneOf:
            - type: array
              items: *ApiStakePool
              title: "cardano-node stake pools"
            - type: array
              items: *ApiJormungandrStakePool
              title: "jormungandr stake pools"
```

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
```diff
@@ -176,8 +183,13 @@ server byron icarus shelley knownPools ntp =
 
     stakePools :: Server (StakePools n ApiStakePool)
     stakePools =
-             (\_ -> throwError err501)
-        :<|> joinStakePool shelley knownPools
+        (\(ApiT wid) -> do
+            stake <- withWorkerCtx shelley wid liftE liftE $ \wrk -> do
+                (w, _, pending) <- liftHandler $ W.readWallet wrk wid
+                return $ Coin $ fromIntegral $ totalBalance pending w
+            liftHandler $ listStakePools spl stake
```
KtorZ, line 190: This should eventually include the reward balance as well.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -329,8 +341,36 @@ withNetworkLayer tr np addrInfo versionData action = do
...
+    handleQueryFailure = withExceptT
+        (\e -> ErrNetworkUnreachable $ T.pack $ "Unexpected" ++ show e) . ExceptT
+    _stakeDistribution queue pt coin = do
+        stakeMap <- fromPoolDistr <$> handleQueryFailure
+            (queue `send` CmdQueryLocalState pt OC.GetStakeDistribution)
+        let toStake = Set.singleton $ Left $ toShelleyCoin coin
+        rewardsPerAccount <- fromNonMyopicMemberRewards <$> handleQueryFailure
+            (queue `send` CmdQueryLocalState pt (OC.GetNonMyopicMemberRewards toStake))
+        pparams <- handleQueryFailure
+            (queue `send` CmdQueryLocalState pt OC.GetCurrentPParams)
```
KtorZ, line 353: This one feels quite redundant / awkward. It sure is needed by the caller to compute few things on the pool data, but the caller should simply call `getProtocolParameters`, don't you think ?

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -622,13 +658,6 @@ handleMuxError tr onResourceVanished = pure . errorType >=> \case
         traceWith tr Nothing
         pure onResourceVanished
 
---------------------------------------------------------------------------------
---
--- Temporary
-
-notImplemented :: HasCallStack => String -> a
-notImplemented what = error ("Not implemented: " <> what)
```
KtorZ, line 630: :tada:

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -0,0 +1,148 @@
...
+    sat s = fromRational $ (getPercentage s) / (1 / fromIntegral nOpt)
+
+    -- If we fetch non-myopic member rewards of pools using the wallet
+    -- balance of 0, the resulting map will be empty. So we set the rewards
+    -- to 0 here:
+    stakeButNoRewards = traverseMissing $ \_k s -> pure $ PoolLsqMetrics
+        { nonMyopicMemberRewards = Quantity 0
+        , relativeStake = s
+        , saturation = (sat s)
+        }
```
KtorZ, line 77: Doesn't the spec suggests something about this? Should we order pool randomly in that case?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -0,0 +1,148 @@
...
+    el = getEpochLength gp
+    gh = getGenesisBlockHash gp
+    getTip = fmap (toPoint gh el) . liftIO $ unsafeRunExceptT $ currentNodeTip nl
+
+    _knownPools
+        :: IO [PoolId]
+    _knownPools = do
+        pt <- getTip
+        res <- runExceptT $ map fst . Map.toList
+            . combineLsqData <$> stakeDistribution nl pt dummyCoin
```
KtorZ, line 116: This shouldn't use the network layer but use the database. Yet, you'll need the work on my branch for that. I'll modify this function in there.

KtorZ (approved):

KtorZ:
bors try

lib/jormungandr/src/Cardano/Pool/Jormungandr/Metrics.hs
```diff
@@ -163,7 +169,8 @@ data StakePoolLayer e m = StakePoolLayer
 -- The pool productions and stake distrubtions in the db can /never/ be from
 -- different forks such that it's safe for readers to access it.
 monitorStakePools
-    :: Tracer IO StakePoolLog
+    :: GetStakeDistribution t IO ~ GetStakeDistribution Jormungandr IO
```
KtorZ, line 172: Any reason for not writing:

```hs
    :: GetStakeDistribution Jormungandr IO
```

?

Anviking: Yes - allowing other targets than `Jormungandr` in the tests :S  https://github.com/input-output-hk/cardano-wallet/blob/1fab8e511360af8e567a6f2fd4e9e7b77c3df917/lib/jormungandr/test/unit/Cardano/Pool/Jormungandr/MetricsSpec.hs#L278-L281

Anviking: I am thinking we should get rid of the `NetworkLayer` (or break it up) later and the need for the type families (and this weirdness)

KtorZ: Okay, this is required for `module Cardano.Pool.Jormungandr.MetricsSpec` and the polymorphism on the target.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
Anviking: 👍 Though why?

I know `knownPools` are used to validate the users's choice of delegation.

The pool worker is less likely to be in sync with the network (more downstream) than the node.

Wouldn't implementing it in terms of the LSQ make more sense?

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
Anviking: This code guarantees that the `pparams` are valid at the same point `pt`.

If we were to instead use
```haskell
getProtocolParameters :: m ProtocolParameters
```
We lose that control, and introduce potential race conditions. Even if unlikely, _that_ feels incredibly wrong to me.

lib/jormungandr/src/Cardano/Pool/Jormungandr/Metrics.hs
KtorZ: > I am thinking we should get rid of the NetworkLayer (or break it up) later and the need for the type families (and this weirdness)

Although the source of the problem isn't much the `NetworkLayer` here, but the double-support Jormungandr / cardano-node :upside_down_face: 
Another abstraction could work better, but it'd still have to deal with backend discrepancies.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
Anviking: I thought that was when we don't have enough data.

I would think that ranking according to the rewards _if_ the user had money, would be a better thing to do.

A user can't delegate without any money anyway.

We could include ROI in % and return `null` for the non-myopic member rewards.

PR #1760, opened 2020-06-15
Server: Use the network layer to get protocol parameters

rvl:
### Issue Number

ADP-244 / #1693

### Overview

The API server was previously getting all network parameters from the genesis block. This change makes it use the network layer to get the latest protocol parameters.

### Comments

An integration test for this is going to be hairy.


piotr-iohk (approved):
ok

KtorZ (approved):

paweljakubas (approved):

Issue #1761, opened 2020-06-15
Extract stake pool registration from on-chain data


KtorZ:
# Context

As a wallet user I want to Check stakepool metadata so that I can make an informed delegation choice.

Stake pool metadata are registered on chain inside stake pool registrations alongside some interesting data like the margin, cost or pledge amount. These registrations can be parsed and the information inside them extracted as we crawl the chain so that we can access it later.

# Decision

- Parse on-chain data and store stake-pool registration certificates in the stake pool database
- Also parse and store metadata details so that they can be fetched from remote servers in a separate worker (see other task)

# Acceptance Criteria

- Pool registrations present on-chain are captured and stored in a dedicated database
- Pool metadata details are also stored and queued but not yet fetched.

---

# Development 

- #1763 

# QA 

- This is relatively hard to test in _isolation_ because to get proper testing, we need to somehow simulate a blockchain with registration and simulate interesting cases. Ultimately, we'll be able to confirm that this works as intended with e2e tests in the integration scenarios, although we need to finalize stake pools listing first (see #1769)

- The PR piggy-back on existing tables and interfaces developed for the ITN, and tests written at this time still apply:
    - https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Cardano/Pool/DB/Properties.hs#L441-L462
    - https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Cardano/Pool/DB/Properties.hs#L408-L439   
    - https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Cardano/Pool/DB/Properties.hs#L389-L406

piotr-iohk:
lgtm

PR #1763, opened 2020-06-15
extract stake pool registrations

KtorZ:
# Issue Number

#1761

# Overview

- a57cf755140ea247a9692ea229894a26acb4b505
  :round_pushpin: **extend pool database with two new tables to store metadata**
  one of them will be used to persist a queue of metadata to fetch, while the other
is a less volatile storage where metadata gets dumped once fetched.
There's in principle no need for caching or timestamps here since
metadata can't be updated (hash is stored on-chain as part of the
registration) which simplifies a bit the problem.

- e776d1c72089ef60bd7befae7f3104e33b43a9ed
  :round_pushpin: **extract pool registration certificate from blocks**
  
- 29166e4ba7c4b4f4a6150c6abef008157005be29
  :round_pushpin: **add logic for monitoring shelley stake pools registration certificates**
  Still rudimentary, and doesn't keep track of metadata information _yet_ but it's coming.

- 1331015b527254d0b1a8d23029f01ec14c45ae71
  :round_pushpin: **write stake pool monitoring in start-up code, with proper loggers**
  
- 7e626619be7244f04f587653b1f5013acdb5d37e
  :round_pushpin: **destroy node's connection when they're no longer needed**
  
- 60651346585eaf503304308ad9970e73c294a804
  :round_pushpin: **do not force rollback when the list of checkpoints is empty**
  An empty list basically assumes that there's no checkpoint on the driver, so nothing to rollback.

- f49a7bf40e0c3d9a11048be419cb9106faf66cd1
  :round_pushpin: **extract and store pool metadata references from the chain**
  
- 4ef1b8097af05063f21669314ccf6969f5dbf0d0
  :round_pushpin: **add basic property tests for checking new pool db functions.**
  
- 6ef598028f0dbeb3230eb66eea6a9b0a7aff64cf
  :round_pushpin: **store block production in the pool monitoring worker**
  
- e8e1d8302310973173f9e1b7af5e9192dc16f094
  :round_pushpin: **change database fn for fetching metadata ref to return them one-by-one**
  This is how the worker uses it in the end, and it makes the signature looks less awkward.

- 2c63840b9ddcc8dcd28ae3131fdb1e1fb8a3ac3a
  :round_pushpin: **avoid intermediary akward 'metadataQueue' table**
    Instead, we can store metadata references alongside the registration
  (with blank / default values for Jörmungandr)

- 279cd62c35a90b00329c4767e53150ce4c634955
  :round_pushpin: **remove unused module 'Cardano.Pool.Metadata'**
  Turns out that we won't be able to fetch metadata from Smash as it was
originally intended:

a) Metadata are fetchted by metadata hash
b) Metadata needs to be fetched as raw bytestring, and decoded to JSON
   _after_ so that we are able to compute their corresponding hash and
   verify it. If we immediately deserialize to JSON, then we would have
   to re-encode them but there's no particular canonical encoding
   enforced when registering metadata, which leaves many possibility
   about how to encode a single file.

Plus, there's no need for caching at the request-level since once
fetched, there's no reason to fetch a result again. This module is
therefore useless.

- d393b7db500400330de5aec15cdbdb5058c25c5c
  :round_pushpin: **move jormungandr-specific data-type into jormungandr package target**
  This cleans up a bit the metadata-related types because it had become a mess. In the end, we'll keep in the core package
the types as they're expected by cardano-wallet-shelley and simply move the exceptional ones (jormungandr) inside its
relevant package.

# Comments

Kinda hard to fully test without wiring it up into the stake pool listing, but logs are quite positive:

```
[cardano-wallet.pools-engine:Info:1071] [2020-06-17 10:22:36.17 UTC] Monitoring stake pools. Currently at genesis
[cardano-wallet.network:Info:1071] [2020-06-17 10:22:36.17 UTC] Looking for an intersection with the node's local chain with: 
[cardano-wallet.network:Info:1071] [2020-06-17 10:22:36.18 UTC] Intersection found: 01010101
[cardano-wallet.pools-engine:Info:1071] [2020-06-17 10:22:36.18 UTC] Applying blocks [0.0 ... 0.0]
[cardano-wallet.pools-engine:Info:1071] [2020-06-17 10:22:36.19 UTC] Applying blocks [0.12 ... 2.0]
[cardano-wallet.pools-engine:Info:1071] [2020-06-17 10:22:36.19 UTC] Discovered stake pool registration: Registration of 5a7b67c7 owned by [ed25519_pk1h3cqm8h6glw56hqa5mgy0gl97jxcawzpphxy5hwtchq8hra6zcwq70fndy]
[cardano-wallet.pools-engine:Info:1071] [2020-06-17 10:22:36.19 UTC] Discovered stake pool registration: Registration of c7258ccc owned by [ed25519_pk1mnjxa09v9hmn48awwcgyrfmnnq4dvs7fsege3glcg6yn0vckn0xsrm4zqy]
[cardano-wallet.pools-engine:Info:1071] [2020-06-17 10:22:36.19 UTC] Discovered stake pool registration: Registration of 775af3b2 owned by [ed25519_pk1qud4ay0zapd0r43c88lc9av8fnesx2hu0r36dsf46lprdp9p3zksx7q880]
```


lib/shelley/src/Cardano/Wallet/Shelley.hs
```diff
@@ -229,15 +234,16 @@ serveWallet
...
-                startServer
-                    proxy
-                    socket
-                    randomApi
-                    icarusApi
-                    shelleyApi
-                    spl
-                    ntpClient
-                pure ExitSuccess
+                withPoolsMonitoring databaseDir (genesisParameters np) nl $ do
```
Anviking, line 237: Since `withPoolsMonitoring` hides the `DBLayer`, should we make it pass a `StakePoolLayer` to the action?

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -303,33 +305,56 @@ toSlotNo :: W.EpochLength -> W.SlotId -> SlotNo
...
-            }
+        { header = toBlockHeader genesisHash epLength blk
         , transactions = txs
         , delegations  = mconcat certs
         }
 
+fromShelleyBlock'
+    :: W.EpochLength
+    -> ShelleyBlock
+    -> (W.SlotId, [W.PoolRegistrationCertificate])
```
Anviking, line 371: We should extract retirement certificates here too, eventually?

Btw, to me scattering the logic like this, while consistent, feels a bit wrong. Wouldn't matching on the `PoolCert` constructors in the `Shelley.Pool` module directly be clearer? https://github.com/input-output-hk/cardano-ledger-specs/blob/036ff8223310ecdaf7f0ed8c211c691cb0dcb8f3/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/TxData.hs#L287-L292

Anviking: Ok, I saw your:
> FIXME We need to acknowledge pool retirement

lib/shelley/src/Cardano/Wallet/Shelley.hs
KtorZ: I don't follow? The stake pool layer is independent of the pools monitoring?

KtorZ:
bors try

lib/core/src/Cardano/Pool/DB.hs
```diff
@@ -116,6 +117,23 @@ data DBLayer m = forall stm. (MonadFail stm, MonadIO stm) => DBLayer
...
+    , deletePoolMetadataRef
+        :: PoolId
+        -> stm ()
+        -- ^ Remove pool metadata references from the database.
+
+    , readPoolMetadataRef
+        :: Int
+        -> stm [(PoolId, StakePoolMetadataRef)]
+        -- ^ Peek at some pool metadata ref. Returns at most 'n' elements, where
+        -- 'n' is the first parameter.
```
Anviking: Why the limit?

_To allow the worker to work with batches I presume?_

KtorZ: Initially yes, but turns out I am going one-by-one with the worker :shrug:

KtorZ: Okay, I've changed this to `peekPoolMetadataRef` and to return only one. Also adjusted the property tests, see: https://github.com/input-output-hk/cardano-wallet/commit/e8e1d8302310973173f9e1b7af5e9192dc16f094

KtorZ:
bors try

lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -214,6 +216,15 @@ newDBLayer trace fp = do
             fmap (poolRegistrationPoolId . entityVal) <$> selectList [ ]
                 [ Desc PoolRegistrationSlot ]
 
+        , putPoolMetadataRef = \poolId ref -> do
+            putMany [PoolMetadataQueue poolId (metadataURL ref) (metadataHash ref)]
+
+        , deletePoolMetadataRef = \poolId -> do
+            deleteWhere [PoolMetadataQueuePoolId ==. poolId]
```
Anviking: `deletePoolMetadataRef` seems unused. Can it be removed?

You're relying on overwrites to handle rollback? (keyed by pool id?) How about either
1. Rolling back `pool_metadata` and `pool_metadata_queue` with `SlotId` like normally?
2. Not including `PoolId` in  `pool_metadata` and `pool_metadata_queue`, but rather only using the metadata hash for the relationships, and primary key? (No fear for race conditions this way, I think)

Anviking:
Partly related question:

Why does a pool both have `owner :: PoolOwner` and `owners :: [PoolOwner]`?
```haskell
-- | Pool ownership data from the stake pool registration certificate.
data PoolRegistrationCertificate = PoolRegistrationCertificate
    { poolId :: !PoolId
    , poolOwners :: ![PoolOwner]
    , poolMargin :: Percentage
    , poolCost :: Quantity "lovelace" Word64
    } deriving (Generic, Show, Eq, Ord)

data StakePoolMetadata = StakePoolMetadata
    { owner :: PoolOwner
    -- ^ Bech32-encoded ed25519 public key.
    , ticker :: StakePoolTicker
    -- ^ Very short human-readable ID for the stake pool.
    , name :: Text
    -- ^ Name of the stake pool.
    , description :: Maybe Text
    -- ^ Short description of the stake pool.
    , homepage :: Text
    -- ^ Absolute URL for the stake pool's homepage link.
    , pledgeAddress :: Text
    -- ^ Bech32-encoded address.
    } deriving (Eq, Show, Generic)
```
Do we use the wrong terminology for one of them?

Anviking (changes requested):
Mostly not sure about how resistant this is to rollbacks and race-conditions (see comment). Otherwise looks promising (certs are discovered on FF in my PR, worker seems to run smoothly)

lib/core/src/Cardano/Pool/DB/Sqlite.hs
KtorZ: `deletePoolMetadataRef` is used by the worker fetching the metadata, coming in another PR.

> You're relying on overwrites to handle rollback?

Indeed. 

1. For the `pool_metadata_queue` it could make sense indeed, but not for the pool_metadata. Metadata aren't on-chain (although I believe your idea would be to assign them the slot id of the pool registration?). 

2. The `PoolId` is needed if you want to be able to map a particular pool to its metadata. Yet you're right, the metadata-hash feels like a better primary key. 

If we indeed go for the metadata-hash, it also reinforces my belief that rolling back pool metadata is awkward; we'd have to re-fetch them again even if the hash hasn't changed, which sounds like a waste of time / resources.

KtorZ: Also, _why_ isn't `pool_metadata_queue` storing slot id and going through the normal rollback mechanism is because this table is ephemeral. Data in there are stored, and then removed. It's really just a queue (hence the name) and not used for long-term storage.

Anviking: 👍 So let the metadata only be keyed by hash, and not pool id?

(Agree on the storing slotId seems suboptimal point)

Anviking: Also: What are the reasons for keeping `PoolMetadataQueue` and `PoolRegistration` separate tables?

The worker _could_ select registration certificates whose metadata hashes are not in the metadata table?

lib/core/src/Cardano/Pool/DB/Sqlite/TH.hs
```diff
@@ -89,4 +93,26 @@ PoolRegistration sql=pool_registration
...
+PoolMetadataQueue sql=pool_metadata_queue
+    poolMetadataQueuePoolId            W.PoolId           sql=pool_id
+    poolMetadataQueueUrl               Text               sql=metadata_url
+    poolMetadataQueueHash              ByteString         sql=metadata_hash
+
+    Primary poolMetadataQueuePoolId
+    deriving Show Generic
+
+-- Cached metadata after they've been fetched from a remote server.
+PoolMetadata sql=pool_metadata
```
Anviking, line 99: `StakePoolMetadata` contains a field `pledgeAddress :: Text`. Should this type too?

KtorZ: We don't have any `PledgeAddress` with cardano-node, but we do have a pledge amount. I've started to rework a bit the metadata types because it's sort of hard to follow with both jormungandr and cardano-node.

lib/core/src/Cardano/Pool/DB/Sqlite/TH.hs
```diff
@@ -85,34 +83,28 @@ PoolOwner sql=pool_owner
...
-    Primary poolMetadataPoolId
+    poolMetadataHash                   W.StakePoolMetadataHash sql=metadata_hash
+    poolMetadataPoolId                 W.PoolId                sql=pool_id
+    poolMetadataName                   Text                    sql=name
+    poolMetadataTicker                 W.StakePoolTicker       sql=ticker
+    poolMetadataDescription            Text Maybe              sql=description
+    poolMetadataHomepage               Text                    sql=homepage
+
+    Primary poolMetadataHash
+    Foreign PoolRegistration fk_registration_metadata_hash poolMetadataPoolId ! ON DELETE CASCADE
```
Anviking, line 108:
```
Foreign PoolRegistration fk_registration_metadata_hash poolMetadataPoolId ! ON DELETE CASCADE
```
Does this delete the metadata when the registration certificate with the same pool id is rolled deleted/rolled back?

KtorZ:
bors try

lib/jormungandr/test/unit/Cardano/Wallet/Jormungandr/ApiSpec.hs
```diff
@@ -152,13 +189,68 @@ spec = do
...
+        "verify that every type used with JSON content type in a servant API \
+        \has compatible ToJSON and ToSchema instances using validateToJSON." $ do
+        validateEveryToJSON
+            (Proxy :: Proxy (ListStakePools ApiStakePool))
+
+
+    describe
+        "can perform roundtrip JSON serialization & deserialization, \
+        \and match existing golden files" $ do
+            jsonRoundtripAndGolden $ Proxy @ApiStakePool
```
Anviking, line 202: You've moved the types and preserved the tests 💯

lib/byron/src/Cardano/Wallet/Byron/Network.hs
```diff
@@ -617,6 +626,10 @@ instance ToText NetworkLayerLog where
             [ "Error when querying local state parameters:"
             , T.pack e
             ]
+        MsgDestroyCursor threadId -> T.unwords
+            [ "Destroying cursor connection at"
```
Anviking, line 630: Was there a specific story for the addition of `destroyCursor`? It was called when workers crashed, right? You noticed chain sync threads staying around?

lib/core/src/Cardano/Pool/DB/Sqlite/TH.hs
```diff
@@ -81,12 +83,28 @@ PoolOwner sql=pool_owner
...
+-- Cached metadata after they've been fetched from a remote server.
+PoolMetadata sql=pool_metadata
+    poolMetadataHash                   W.StakePoolMetadataHash sql=metadata_hash
+    poolMetadataPoolId                 W.PoolId                sql=pool_id
+    poolMetadataName                   Text                    sql=name
+    poolMetadataTicker                 W.StakePoolTicker       sql=ticker
+    poolMetadataDescription            Text Maybe              sql=description
+    poolMetadataHomepage               Text                    sql=homepage
+
+    Primary poolMetadataHash
```
Anviking, line 107: 👍

Anviking (approved):
Awesome! Lgtm. I think the further jormungandr-haskell differentiation and new metadata ref approach is good.

lib/core/src/Cardano/Pool/DB/Sqlite/TH.hs
KtorZ: Yes.

KtorZ: Which might not be desirable actually.

lib/byron/src/Cardano/Wallet/Byron/Network.hs
KtorZ: Yeah, this is a bit of a side-story within the story. While I was debugging the worker looping and crashing I noticed that we weren't actually closing connections to the node and eventually, the system would run out of file descriptors. 

These doesn't quite fix the problem though as the it seems that even killing the thread doesn't close the connection (not immediately at least) and I was still able to reach the maximum number of sockets allowed by the system fairly quickly :upside_down_face: ... that's a first step in the right direction I believe. Perhaps some little delay before looping in case of errors would be nice to give time for the connection to be cleared up.

KtorZ:
Please.

Issue #1751, opened 2020-06-12
Cannot migrate big wallet


piotr-iohk:
# Context

There is -> https://github.com/input-output-hk/cardano-wallet/blob/master/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs#L215
 
```
    _estimateMaxNumberOfInputs _ _ =
        -- FIXME Implement.
        100
```
which causes that big wallets cannot be migrated using migration endpoint.

Integration test -> `SHELLEY_MIGRATE_01_big_wallet` (PR: https://github.com/input-output-hk/cardano-wallet/pull/1739)

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | `2020.6.5 (git revision: 3bafec15d862c81c052c4cf450fa7e6cf29d0f76)`|
| Platform     | Linux, All         |
| Installation | from source           |

# Steps to Reproduce

1. `stack test cardano-wallet-shelley:integration --ta "--match SHELLEY_MIGRATE_01_big_wallet"`

## Expected behavior
Test passes.

## Actual behavior
 
Test fails:

```
  7) API Specifications SHELLEY_MIGRATE_01 -  migrate a big wallet requiring more than one tx
       expected: Status {statusCode = 202, statusMessage = "Accepted"}
        but got: Status {statusCode = 500, statusMessage = "Internal Server Error"}
       
       from the following response: Left (DecodeFailure "{\"code\":\"created_invalid_transaction\",\"message\":\"That's embarrassing. It looks like I've created an invalid transaction that could not be parsed by the node. Here's an error message that may help with debugging: ApplyTxError [LedgerFailure (UtxowFailure (UtxoFailure (MaxTxSizeUTxO 20614 4096)))]\"}")

```

---

# Resolution 

There is -> https://github.com/input-output-hk/cardano-wallet/blob/master/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs#L215
 
```
    _estimateMaxNumberOfInputs _ _ =
        -- FIXME Implement.
        100
```

- #1764

---

# QA 

[...]

PR #1764, opened 2020-06-15
Implement the estimation of maximum number of inputs for Shelley and show big wallet migration

paweljakubas:
# Issue Number

#1751 

# Overview

- [x] I have refactored computeTxSize to use later
- [x] I have implemented `_estimateMaximumNumberOfInputs`
- [x] I have unpended integration test showing the case - big wallet migration
- [x] I have added property tests

# Comments


paweljakubas:
bors try

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
```diff
@@ -246,10 +242,9 @@ spec = do
             Empty >>= flip verify
             [ expectField
                 #distribution
-                ((`shouldBe` (Just 400)) . Map.lookup 10_000_000_000)
+                ((`shouldBe` (Just 129)) . Map.lookup 100_000_000_000)
```
piotr-iohk: This does not appear to be right... :thinking: 
The fixture wallet has 400 utxo * 100_000_000_000 + dust (100 utxo * 1), so I'd expect target wallet ditribution to be 400, not 129...

https://github.com/input-output-hk/cardano-wallet/blob/master/lib/shelley/test/data/cardano-node-shelley/genesis.yaml#L1217

paweljakubas: probably addresses were reused and hence we have decreased number of utxos, BUT balance is ok. if so the test should be corrected in such a way that we are sure 400 distinct addresses are used in the migration. That's my guess

paweljakubas: I do not confirm original balance 400 * ...
I see : 
```
################## originalBalance:12900000000000
################## expectedFee:3574500
################## expectedBalance:12899996425500
```

piotr-iohk: Yes, but wallet from the genesis file has the following: 400 * 100_000_000_000 and 100 * 1... so  the original balance should be 40000000000500... :thinking:

paweljakubas:
bors try

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
paweljakubas: I think it should be subject of another BUG ...

KtorZ (approved):

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
KtorZ: The problem comes from:

a) Funds are defined as a key-value "map" in the genesis file. 
b) This is therefore parsed as a map by ledger's functions & al. 
c) These identify _sequential_ addresses, that need to be discovered in order! However, the map "destroys" any sort of order we could have on keys... and there's no guarantee that we process addresses in order. Actually, we don't, and are only able to discover 129 of those addresses.

KtorZ: Was harder than I thought, but I got it:

```
(Status {statusCode = 200, statusMessage = "OK"},Right (ApiUtxoStatistics {total = Quantity {getQuantity = 40000000000100}, scale = ApiT {getApiT = Log10}, distribution = fromList [(10,100),(100,0),(1000,0),(10000,0),(100000,0),(1000000,0),(10000000,0),(100000000,0),(1000000000,0),(10000000000,0),(100000000000,400),(1000000000000,0),(10000000000000,0),(100000000000000,0),(1000000000000000,0),(10000000000000000,0),(45000000000000000,0)]}))
```

KtorZ:
bors retry

KtorZ:
bors retry

paweljakubas:
bors retry

paweljakubas:
once again : 
```
Progress 6/7: cardano-wallet-shelley-2020.6.5
Timeout after 45 minutes.
```

KtorZ:
bors retry

KtorZ:
bors retry

Issue #1759, opened 2020-06-15
List/create shelley wallet while cardano-node is syncing takes a long time


piotr-iohk:
# Context

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | `2020.6.5 (git revision: 0fd72625c3ed83999a59c6187445774210ac5d12)` |
| Platform     | Linux, all?        |
| Installation | From source            |

# Steps to Reproduce

1. start cardano-node + cardano-wallet on Shelley testnet.
e.g.:
cardano-node:
```
nix-build -A scripts.ff.node -o ff-node-local && ./ff-node-local
```

cardano-wallet-shelley:
```
cardano-wallet-shelley serve --port 8090 \
	--node-socket /path/to/node.socket  \
	--testnet /nix/store/v8lb35176p73f6ksp7p5rhkqmygi1zj9-ff-genesis.json \
        --database ./testnet-ff
```

2. While node is in the process of syncing, create shelley wallet and/or try to list shelley wallets.

## Expected behavior

Results should appear.

## Actual behavior
 
It takes quite a long time. Occasionally one can get:

```
Code = 503,
Message = {"code":"network_unreachable","message":"The node backend is unreachable at the moment. Trying again in a bit might work."}
```

When it takes long time, it appears to be co-related with the log entries for "Querying the reward account balance" and "Result for account":

```
[RequestId 38] [GET] /v2/wallets
[cardano-wallet.wallet-db:Info:1940] [2020-06-15 11:25:15.16 UTC] Found existing wallet: 54dd899cfa7f27b8e02a127f6b9088b77fd150e1
[cardano-wallet.network:Info:1940] [2020-06-15 11:25:15.16 UTC] Querying the reward account balance for 256db575ccb98453aae20137dbbc45311150a9732278fe64fc3233379eed3f66 at 70617265<-[c2b78cdb-8.12970#8927]

```
...waiting for result...
```
[cardano-wallet.network:Info:1940] [2020-06-15 11:25:15.17 UTC] Result for account 256db575ccb98453aae20137dbbc45311150a9732278fe64fc3233379eed3f66:
  delegations = fromList []
  rewards = fromList []
```
result received.

---

# Resolution 

- This is temporarily patched by always returning a zero reward account balance.
- It will be resolved properly by caching the reward account balance in a database table and updating it in a worker thread ⇒ #1768.

---

# QA 

- Manually testing reward balance and performance on the shelley_testnet. :-/

[...]

PR #1768, opened 2020-06-16
Store delegation reward account balances in the database

rvl:
### Issue Number

ADP-302 / #1750 / #1759

### Overview

Turns out querying reward account balances can be slow.
This is not good if it's done when listing/getting wallets.
Better to do these queries in a separate thread and cache the results in the database.

- [x] Add new database table for the latest known wallet reward account balance.
- [x] Use the database when getting the reward account balance.
- [x] Add a network layer method to watch the node tip.
- [x] Update the account balance in each wallet when the node tip changes (asynchronously).
- [x] Store reward account balance in db as it arrives.
- [x] Adapt Jörmungandr network layer


lib/core/src/Cardano/Wallet.hs
```diff
@@ -845,11 +848,29 @@ deleteWallet ctx wid = db & \DBLayer{..} -> do
...
+        ( HasDBLayer s k ctx
+        )
+    => ctx
+    -> WalletId
+    -> ExceptT ErrFetchRewards IO (Quantity "lovelace" Word64)
+fetchRewardBalance ctx wid = db & \DBLayer{..} -> do
+    _ <- withExceptT ErrFetchRewardsNoSuchWallet
+        . mapExceptT atomically
+        . withNoSuchWallet wid
+        $ readCheckpoint pk
```
KtorZ: Minor note but If this is for triggering a "NoSuchWallet" error, I don't think it is necessary. Since each wallet is stored in an independent database, the only way to get a `ctx` that has a `DBLayer` is to open a corresponding database for that wallet.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -881,6 +902,40 @@ fetchRewardBalance ctx wid = db & \DBLayer{..} -> do
...
+        , HasRewardAccount s
+        , k ~ RewardAccountKey s
+        , WalletKey k
+        , ctx ~ WalletLayer s t k
+        )
+    => ctx
+    -> WalletId
+    -> IO ()
+manageRewardBalance ctx wid = db & \DBLayer{..} ->
+    unsafeRunExceptT $ -- TODO: map error type, not that it really matters
```
KtorZ: I think this is _fine_ though, provided that the thread running this method would catch and log properly the error. If it's using the "Worker" machinery, then it'll be done "for free".

lib/core/src/Cardano/Wallet.hs
```diff
@@ -881,6 +902,40 @@ fetchRewardBalance ctx wid = db & \DBLayer{..} -> do
...
+         res <- lift $ runExceptT (queryRewardBalance @ctx @s @t @k ctx wid)
+         lift $ traceWith tr $ MsgRewardBalanceResult res
+         case res of
+            Right amt ->
+                withExceptT show $
+                mapExceptT atomically $ putDelegationRewardBalance pk amt
+            Left _err ->
+                -- Occasionaly failing to query is generally not fatal. It will
+                -- just update the balance next time the tip changes.
+                pure ()
```
KtorZ, line 940: Not fatal, yes. Shouldn't it deserve a small warning though?

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2010,3 +2075,6 @@ instance HasSeverityAnnotation WalletLog where
         MsgDelegationCoinSelection _ -> Debug
         MsgPaymentCoinSelection _ -> Debug
         MsgPaymentCoinSelectionAdjusted _ -> Debug
+        MsgRewardBalanceQuery _ -> Debug
+        MsgRewardBalanceResult (Right _) -> Debug
+        MsgRewardBalanceResult (Left _) -> Notice
```
KtorZ, line 2173: :+1:

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -580,17 +586,13 @@ mkShelleyWallet
...
         , HasWorkerRegistry s k ctx
         )
     => MkApiWallet ctx s ApiWallet
 mkShelleyWallet ctx wid cp meta pending progress = do
-    -- TODO: issue #1750 re-enable querying reward balance when it's faster
     reward <- withWorkerCtx @_ @s @k ctx wid liftE liftE $ \wrk -> liftHandler $
-        if False
-            then W.fetchRewardBalance @_ @s @t @k wrk wid
-            else pure $ Quantity 0
+        W.fetchRewardBalance @_ @s @k wrk wid
```
KtorZ, line 613: Would this call actually _fails_ if the reward balance isn't yet available?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1613,8 +1617,9 @@ registerWorker ctx wid =
         , workerMain = \ctx' _ -> do
             -- FIXME:
             -- Review error handling here
-            unsafeRunExceptT $
-                W.restoreWallet @(WorkerCtx ctx) @s @t ctx' wid
+            concurrently_
+                (unsafeRunExceptT $ W.restoreWallet @(WorkerCtx ctx) @s @t ctx' wid)
+                (coworker @(WorkerCtx ctx) ctx' wid)
```
KtorZ: It'll be necessary to do the same thing for `initWorker` too. Otherwise, newly created wallet won't have any reward monitoring in place.

lib/core/src/Cardano/Wallet/DB.hs
```diff
@@ -184,6 +184,24 @@ data DBLayer m s k = forall stm. (MonadIO stm, MonadFail stm) => DBLayer
...
+        -- node tip.
+        -- This is separate from putWalletMeta because it's not meta data
+
+    , readDelegationRewardBalance
+        :: PrimaryKey WalletId
+        -> stm (Quantity "lovelace" Word64)
+        -- ^ Get the reward account balance.
+        --
+        -- Returns zero if the wallet isn't found or if wallet hasn't delegated
+        -- stake.
```
KtorZ, line 203: Okay :+1: , this answers my previous question.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -673,6 +674,23 @@ newDBLayer trace defaultFieldValues mDatabaseFile = do
...
+                                 Delegation Rewards
+        -----------------------------------------------------------------------}
+
+        , putDelegationRewardBalance =
+            \(PrimaryKey wid) (Quantity amt) -> ExceptT $ do
+            selectWallet wid >>= \case
+                Nothing -> pure $ Left $ ErrNoSuchWallet wid
+                Just _  -> Right <$> repsert
+                    (DelegationRewardKey wid)
+                    (DelegationReward wid amt)
```
KtorZ, line 697: Thinking about rollbacks for a second: it is possible to rollback to a _previous epoch_. Rewards between two forks might actually be slightly different. I don't think having the reward balance _slightly_ off for a short period of time is much of an issue, because the result will _eventually_ become correct. 

Thoughts?

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -669,6 +687,51 @@ handleMuxError tr onResourceVanished = pure . errorType >=> \case
...
+                    -- signal watcher to stop waiting
+                    finish $ Left msg
+    link worker
+
+    pure register
+
+stateTVar :: MonadSTMTx m => TVar_ m a -> (a -> (b, a)) -> m b
+stateTVar var f = do
+   s <- readTVar var
+   let (a, s') = f s -- since we destructure this, we are strict in f
```
KtorZ: "destructure", that's the sign you've been doing too much JavaScript indeed :smile:

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -669,6 +687,51 @@ handleMuxError tr onResourceVanished = pure . errorType >=> \case
         traceWith tr Nothing
         pure onResourceVanished
 
+{-------------------------------------------------------------------------------
+                                  Tip Watcher
+-------------------------------------------------------------------------------}
+
+setupWatchers
+    :: Tracer IO (WatcherLog e)
+    -> TQueue IO v
```
KtorZ: There's already a lof of concurrent mutable variables down there, since the function only needs an accessor on the tip, I'd actually change this for just a `IO v` and not even mention the `TQueue` here.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -669,6 +687,51 @@ handleMuxError tr onResourceVanished = pure . errorType >=> \case
...
+    -> IO ((v -> IO (Either e ())) -> IO (IO (Either e a)))
+setupWatchers tr nodeTipQ = do
+    watchersVar <- atomically $ newTVar (0 :: Int, [])
+
+    let register cb = do
+            done <- atomically $ newEmptyTMVar
+            let finish = atomically . putTMVar done
+            i <- atomically $ stateTVar watchersVar $ \(nextId, watchers) ->
+                (nextId, (nextId + 1, ((nextId, cb, finish):watchers)))
```
KtorZ: :+1:

KtorZ: Alternatively, `register` could take an explicit `name` for the watcher.

KtorZ (approved):

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
rvl: Yes - I think "eventually correct" is the best we that we can do, because it's not feasible to synchronise wallet applyBlock updates with the LocalStateQuery result messages.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
rvl: Actually I just copy&pasted this function from the stm library... which is maybe also a sign that I've been doing too much JavaScript.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
rvl: OK good idea

lib/core/src/Cardano/Wallet.hs
rvl: It's logged a few lines above.

lib/core/src/Cardano/Wallet/Api/Server.hs
rvl: I added a commend

lib/core/src/Cardano/Wallet.hs
rvl: OK I have adjusted a few things to take into account that the WorkerCtx implies an existing wallet.

lib/core/src/Cardano/Wallet/Api/Server.hs
rvl: OK fixed, thanks

KtorZ:
bors retry

KtorZ:
bors retry

rvl:
bors try

KtorZ:
(just have some little changes to push)

KtorZ:
bors retry

rvl:
Integration test failure seems to be due to a rolled back transaction ⇒ ADP-93. Retrying.

bors r+

KtorZ:
Merging, the failure above seems related to something else: https://github.com/input-output-hk/cardano-wallet/issues/1831

Issue #1719, opened 2020-06-03
Margin, cost & pledge for pools


Anviking:
# Context

We want to implement this endpoint for (the cardano-node-) Shelley.
https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/listStakePools

Pool margin and cost cannot be retrieved through the node's local state query.

# Decision

It seems we have to traverse the chain for pool registration certificates, to get this data?

# Acceptance Criteria

- Stake pool `margin` must be available in API (and CLI)
- Stake pool `cost` must be available in API (and CLI)
- Stake pool `pledge` must be available in API (and CLI)

---

# Development 

# QA 

- See https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/listStakePools

- And https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs#L464-L496

Anviking:
Superseded by #1761

Anviking:
That ticket seems more about the writing of it, so re-opening this for the reading-aspect.

piotr-iohk:
lgtm

PR #1769, opened 2020-06-16
Use and combine chain data with LSQ data for stake pools

Anviking:
# Issue Number

ADP-311, #1719 

# Overview

- [x] Add Shelley `MsgFetchedNodePoolLsqData` trace
- [x] Use and combine chain data with LSQ when listing pools
- [x] Update swagger to make pool cost and margin nullable
- [x] Make API types nullable
- [x] Add model implementation for `readPoolMetadata`

# Comments

- [ ] We should test that we order by non-myopic rewards, but that would require the pools producing blocks. 
- [ ] Maybe we should have a DB test checking that `readMetadata` returns what we previously put in the DB.

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
@@ -26,9 +26,21 @@ import Cardano.Wallet.Primitive.AddressDerivation
 import Cardano.Wallet.Primitive.AddressDerivation.Shelley
     ( ShelleyKey )
 import Cardano.Wallet.Primitive.Types
-    ( Coin (..), Direction (..), PoolId (..), TxStatus (..) )
+    ( Coin (..)
+    , Direction (..)
+    , PoolId (..)
+    , StakePoolMetadata (..)
+    , StakePoolTicker (..)
+    , TxStatus (..)
+    )
+import Cardano.Wallet.Unsafe
+    ( unsafeMkPercentage )
 import Data.Generics.Internal.VL.Lens
     ( view, (^.) )
+import Data.List
+    ( sortOn )
+import Data.Maybe
+    ( mapMaybe )
 import Data.Quantity
     ( Quantity (..) )
 import Data.Text.Class
@@ -65,7 +77,6 @@ import Test.Integration.Framework.DSL
     , walletId
     , (.<=)
     , (.>)
-    , (.>=)
     )
 import Test.Integration.Framework.TestData
     ( errMsg403DelegationFee
@@ -415,68 +426,107 @@ spec = do
             , expectErrorMessage $ errMsg403DelegationFee fee
             ]
 
-    it "STAKE_POOLS_LIST_01 - List stake pools" $ \ctx -> do
-        eventually "Listing stake pools shows expected information" $ do
-            r <- request @[ApiStakePool] ctx
-                (Link.listStakePools arbitraryStake) Default Empty
+    let fixTypeInference = True `shouldBe` True
+    let listPools ctx = request @[ApiStakePool] @IO ctx (Link.listStakePools arbitraryStake) Default Empty
+    describe "STAKE_POOLS_LIST_01 - List stake pools" $ do
+        it "immediately has non-zero saturation & stake" $ \ctx -> do
+            r <- listPools ctx
             expectResponseCode HTTP.status200 r
             verify r
                 [ expectListSize 3
-
--- Pending a mock metadata registry
---                , expectListField 0
---                    #metadata ((`shouldBe` Just "Genesis Pool") . fmap (view #name . getApiT))
---                , expectListField 1
---                    #metadata ((`shouldBe` Just "Genesis Pool") . fmap (view #name . getApiT))
---                , expectListField 2
---                    #metadata ((`shouldBe` Just "Genesis Pool") . fmap (view #name . getApiT))
-
+                -- At the time of setup, the pools have 1/3 stake each, but
+                -- this could potentially be changed by other tests. Hence,
+                -- we try to be forgiving here.
                 , expectListField 0
-                    #cost (`shouldBe` (Quantity 0))
+                    (#metrics . #relativeStake)
+                        (.> Quantity (unsafeMkPercentage 0))
                 , expectListField 1
-                    #cost (`shouldBe` (Quantity 0))
+                    (#metrics . #relativeStake)
+                        (.> Quantity (unsafeMkPercentage 0))
                 , expectListField 2
-                    #cost (`shouldBe` (Quantity 0))
-
+                    (#metrics . #relativeStake)
+                        (.> Quantity (unsafeMkPercentage 0))
                 , expectListField 0
-                    #margin (`shouldBe` (Quantity minBound))
+                    (#metrics . #saturation) (.> 0)
                 , expectListField 1
-                    #margin (`shouldBe` (Quantity minBound))
+                    (#metrics . #saturation) (.> 0)
                 , expectListField 2
-                    #margin (`shouldBe` (Quantity minBound))
-
--- Pending stake pools producing blocks in our setup,
--- AND pending keeping track of block producions
---                , expectListField 0
---                    (#metrics . #producedBlocks) (.>= Quantity 0)
---                , expectListField 1
---                    (#metrics . #producedBlocks) (.>= Quantity 0)
---                , expectListField 2
---                    (#metrics . #producedBlocks) (.>= Quantity 0)
---
---                , expectListField 0
---                    (#metrics . #nonMyopicMemberRewards) (.>= Quantity 0)
---                , expectListField 1
---                    (#metrics . #nonMyopicMemberRewards) (.>= Quantity 0)
---                , expectListField 2
---                    (#metrics . #nonMyopicMemberRewards) (.>= Quantity 0)
-
-                , expectListField 0
-                    (#metrics . #saturation) (.>= 0)
-                , expectListField 1
-                    (#metrics . #saturation) (.>= 0)
-                , expectListField 2
-                    (#metrics . #saturation) (.>= 0)
+                    (#metrics . #saturation) (.> 0)
                 ]
 
+        it "eventually has correct margin and cost" $ \ctx -> do
+            eventually "pool worker finds the certificate" $ do
+                r <- listPools ctx
+                expectResponseCode HTTP.status200 r
+                verify r
+                    [ expectListField 0
+                        (#cost) (`shouldBe` Just (Quantity 0))
+                    , expectListField 1
+                        (#cost) (`shouldBe` Just (Quantity 0))
+                    , expectListField 2
+                        (#cost) (`shouldBe` Just (Quantity 0))
+
+                    , expectListField 0
+                        #margin (`shouldBe` Just
+                            (Quantity $ unsafeMkPercentage 0.1 ))
+                    , expectListField 1
+                        #margin (`shouldBe` Just
+                            (Quantity $ unsafeMkPercentage 0.1 ))
+                    , expectListField 2
+                        #margin (`shouldBe` Just
+                            (Quantity $ unsafeMkPercentage 0.1 ))
+
+                    , expectListField 0
+                        (#metrics . #producedBlocks) (.> Quantity 0)
+                    , expectListField 1
+                        (#metrics . #producedBlocks) (.> Quantity 0)
+                    , expectListField 2
+                        (#metrics . #producedBlocks) (.> Quantity 0)
+                    -- TODO: Test that we have non-zero non-myopic member
+                    -- rewards, and sort by it.
+                    ]
+
+        it "contains pool metadata" $ \ctx -> do
+            eventually "metadata is fetched" $ do
+                r <- listPools ctx
+                let metadataList =
+                        [ StakePoolMetadata
+                            { ticker = (StakePoolTicker "GPA")
+                            , name = "Genesis Pool A"
+                            , description = Nothing
+                            , homepage = "https://iohk.io"
+                            }
+                        , StakePoolMetadata
+                            { ticker = (StakePoolTicker "GPB")
+                            , name = "Genesis Pool B"
+                            , description = Nothing
+                            , homepage = "https://iohk.io"
+                            }
+                        , StakePoolMetadata
+                            { ticker = (StakePoolTicker "GPC")
+                            , name = "Genesis Pool C"
+                            , description = Just "Lorem Ipsum Dolor Sit Amet."
+                            , homepage = "https://iohk.io"
+                            }
+                        ]
+
+                verify r
+                    [ expectListSize 3
+                    , expectField id $ \pools ->
+                        -- To ignore the arbitrary order,
+                        -- we sort on the names before comparing
+                        sortOn name ( mapMaybe (fmap getApiT . view #metadata) pools)
+                        `shouldBe` metadataList
+                    ]
+
     it "STAKE_POOLS_LIST_05 - Fails without query parameter" $ \ctx -> do
-        _w <- fixtureWallet ctx -- Ambiguous type error without this line
+        fixTypeInference
         r <- request @[ApiStakePool] ctx
             (Link.listStakePools Nothing) Default Empty
         expectResponseCode HTTP.status400 r
 
     it "STAKE_POOLS_LIST_06 - NonMyopicMemberRewards are 0 when stake is 0" $ \ctx -> do
-        _w <- fixtureWallet ctx
+        fixTypeInference
         let stake = Just $ Coin 0
         r <- request @[ApiStakePool] ctx (Link.listStakePools stake) Default Empty
         expectResponseCode HTTP.status200 r
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
@@ -46,7 +46,7 @@ import Data.Quantity
 import Data.Text.Class
     ( toText )
 import Test.Hspec
-    ( SpecWith, describe, it, shouldBe, xit )
+    ( SpecWith, describe, it, shouldBe, shouldSatisfy, xit )
 import Test.Integration.Framework.DSL
     ( Context (..)
     , Headers (..)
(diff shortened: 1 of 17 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -417,58 +429,80 @@ spec = do
             ]
 
     it "STAKE_POOLS_LIST_01 - List stake pools" $ \ctx -> do
-        eventually "Listing stake pools shows expected information" $ do
```
Anviking, line 419: I think there's a _theoretical_ chance that the pool registration certificates haven't been discovered yet, but I'm not sure it's possible in practice.

It fails quicker without the default delay.

I guess we could use `eventuallyUsingDelay (5 * second)` also.

lib/core/src/Cardano/Pool/DB.hs
```diff
@@ -113,10 +113,13 @@ data DBLayer m = forall stm. (MonadFail stm, MonadIO stm) => DBLayer
         -- ^ Find a registration certificate associated to a given pool
 
     , unfetchedPoolMetadataRefs
-        :: stm [(PoolId, StakePoolMetadataUrl, StakePoolMetadataHash)]
+        :: Int
+        -> stm [(PoolId, StakePoolMetadataUrl, StakePoolMetadataHash)]
```
Anviking: 🧐 bad rebase

Anviking:
bors try

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -416,68 +428,108 @@ spec = do
...
+                                (Quantity $ unsafeMkPercentage 0.1 ))
+
+                        -- TODO: Switch from (.>=) to (.>) (and perhaps an eventually) when
+                        -- our stake pool cluster are producing blocks.
+                        , expectListField 0
+                            (#metrics . #producedBlocks) (.>= Quantity 0)
+                        , expectListField 1
+                            (#metrics . #producedBlocks) (.>= Quantity 0)
+                        , expectListField 2
+                            (#metrics . #producedBlocks) (.>= Quantity 0)
```
Anviking: Now after rebasing I see them producing blocks, but I still see no rewards 🤔 

_or I wasn't fully rebased… trying again_

_I think some monetary protocol parameters need to be tweaked in the genesis file_

KtorZ: > I think some monetary protocol parameters need to be tweaked in the genesis file

That is possible.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -416,68 +428,108 @@ spec = do
...
-                    (#metrics . #saturation) (.>= 0)
-                , expectListField 1
-                    (#metrics . #saturation) (.>= 0)
-                , expectListField 2
-                    (#metrics . #saturation) (.>= 0)
-                ]
+    let fixTypeInference = True `shouldBe` True
+    let listPools ctx = request @[ApiStakePool] @IO ctx (Link.listStakePools arbitraryStake) Default Empty
+    describe "STAKE_POOLS_LIST_01 - List stake pools" $ do
+            it "immediately has non-zero saturation & stake" $ \ctx -> do
```
KtorZ: neat-picking but why the double indentation here?

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -416,68 +428,108 @@ spec = do
...
+                                , homepage = "https://iohk.io"
+                                }
+                            ]
+
+                    verify r
+                        [ expectListSize 3
+                        , expectField id $ \pools ->
+                            -- To ignore the arbitrary order,
+                            -- we sort on the names before comparing
+                            sortOn name ( mapMaybe (fmap getApiT . view #metadata) pools)
```
KtorZ: :+1:

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -416,68 +428,108 @@ spec = do
...
+                                , description = Nothing
+                                , homepage = "https://iohk.io"
+                                }
+                            , StakePoolMetadata
+                                { ticker = (StakePoolTicker "GPC")
+                                , name = "Genesis Pool C"
+                                , description = Just "Lorem Ipsum Dolor Sit Amet."
+                                , homepage = "https://iohk.io"
+                                }
+                            ]
```
KtorZ: Maybe we could avoid hard-coding the list here since it's available as `operators` in `Cardano.Wallet.Shelley.Launch`.

lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -463,6 +465,7 @@ instance ToText DBLog where
         MsgQuery stmt _ -> stmt
         MsgRun False -> "Running database action - Start"
         MsgRun True -> "Running database action - Finish"
+        MsgWillOpenDB fp -> "Will open db at " <> (maybe "in-memory" T.pack fp)
```
KtorZ, line 468: :+1: Anything particular you were looking for :thinking: ?

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -440,6 +440,7 @@ withStakePool tr idx cfg action =
     bracketTracer' tr "withStakePool" $
         withTempDir tr name $ \dir -> do
             withStaticServer dir $ \url -> do
+                putStrLn $ "Launching static server at url: " <> url
```
KtorZ: :grimacing: ... There's litteraly a tracer `tr` in scope that can be used instead of printing to stdout!

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -98,13 +111,130 @@ import qualified Data.Map.Merge.Strict as Map
...
+    , listStakePools = _listPools
+    }
+  where
+    _knownPools
+        :: IO [PoolId]
+    _knownPools = do
+        tip <- getTip
+        let dummyCoin = Coin 0
+        res <- runExceptT $ map fst . Map.toList
+            . combineLsqData <$> stakeDistribution nl tip dummyCoin
```
KtorZ, line 145: known pools should be implemented using the _database_ only. That is its sole purpose, otherwise it is utterly redundant with listPools. The goal is to have a fast access to known pools ids to avoid having to query the network layer when delegating / undelegating to pools.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -98,13 +111,130 @@ import qualified Data.Map.Merge.Strict as Map
...
+        -> ExceptT ErrNetworkUnavailable IO [Api.ApiStakePool]
+    _listPools userStake = do
+            tip <- liftIO getTip
+            lsqData <- combineLsqData <$> stakeDistribution nl tip userStake
+            chainData <- liftIO $ readDBPoolData db
+            return
+              . sortOn (Down . (view (#metrics . #nonMyopicMemberRewards)))
+              . map snd
+              . Map.toList
+              $ combineDBAndLsqData lsqData chainData
```
KtorZ: _You can ignore this comment_

The mixed case style somewhat triggers some OCDs in me. 

```
combineDBAndLSQData
```

or 

```
combineDbAndLsqData
```

Mixing both looks weird.

lib/core/src/Cardano/DB/Sqlite.hs
Anviking: I wanted to browse manually to check 1) if pool registration certificates had metadata references 2) if any metadata had been fetched

I believe I was simply missing an `eventually` though.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -98,13 +111,130 @@ import qualified Data.Map.Merge.Strict as Map
...
+    -> Map PoolId Api.ApiStakePool
+combineDBAndLsqData =
+    Map.merge lsqButNoChain chainButNoLsq bothPresent
+  where
+    lsqButNoChain = traverseMissing $ \k lsq -> pure $ mkApiPool k lsq Nothing
+
+    -- TODO: Maybe we have simply missed a retirement certificate. Maybe we
+    -- should simply dropMissing here.
+    chainButNoLsq = traverseMissing $ \k _ ->
+        error $ "Pool exists in DB but not in LSQ: " <> show k
```
KtorZ: Yes we should! We do not track retirement certificates yet, so we definitely need this to account for them, same as we did for the ITN. When there's a pool in our database, but not returned by the LSQ, we fully trust the node and discard the pool on our end :+1:

Anviking: Ok, will change now

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
Anviking: What do you say about: https://github.com/input-output-hk/cardano-wallet/pull/1753#discussion_r440780076 ?

> I know knownPools are used to validate the users's choice of delegation.
>
> The pool worker is less likely to be in sync with the network (more downstream) than the node.
>
> Wouldn't implementing it in terms of the LSQ make more sense?

Though perhaps, having the pool worker out of sync with the node is unlikely, since they Daedalus users will always launch the node and wallet at the same time?

KtorZ: Because that is why we have this function in the first place. We can already get the full list of pool ids from the `listPools` and it was how it was done _before_. Yet, this was slow and sometimes _unreliable_ because it needed to go through much more work. As a result, requests for joining / quitting pools were often failing because we couldn't reliably get a list of pools ids. 

Yet, getting the list of pool ids is straightforward and very reliable when fetched from the local database. Implementing this function using the LSQ completely defies its original purpose which was to circumvent the network layer!

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
Anviking: I created the related https://github.com/input-output-hk/cardano-wallet/pull/1783, which might be good for investigating. We can leave the todo for rewards in this PR for now.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
Anviking: A remnant, fixed.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
Anviking: Don't think we can (or want) `cardano-wallet-core-integration` depend on `cardano-wallet-shelley`… :S  We _could_ move `API.Shelley.StakePools` to `cardano-wallet-shelley:integration` though?

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -27,11 +27,23 @@ import Cardano.Wallet.Primitive.AddressDerivation
...
+import Cardano.Wallet.Unsafe
+    ( unsafeMkPercentage )
 import Data.Generics.Internal.VL.Lens
     ( view, (^.) )
+import Data.Maybe
+    ( mapMaybe )
 import Data.Quantity
     ( Quantity (..) )
+import Data.Sort
+    ( sortOn )
```
rvl: You don't need the sort package. It's in base `Data.List.sortOn`.

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -1038,6 +1039,7 @@ withSystemTempDir tr name action = do
 data ClusterLog
     = MsgCartouche
     | MsgLauncher String LauncherLog
+    | MsgServerForMetadataStarted FilePath
```
rvl: Might be good to log both the directory and url - the url type being String.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -123,76 +253,53 @@ combineLsqData NodePoolLsqData{nOpt, rewards, stake} =
...
+-- NOTE: If performance becomes a problem, we could try replacing all
+-- the individual DB queries, and combbination functions with a single
+-- hand-written Sqlite query.
+readDBPoolData
+    :: DBLayer IO
+    -> IO (Map PoolId PoolDBMetrics)
+readDBPoolData DBLayer{..} = atomically $ do
+    pools <- listRegisteredPools
+    certMap <- Map.fromList . catMaybes <$> forM pools
+        (\poolId -> ((fmap (poolId,)) <$> readPoolRegistration poolId))
```
rvl: A bit simpler...

```suggestion
    registrations <- mapM readPoolRegistration pools
    let certMap = Map.fromList [(poolId, cert) | (poolId, Just cert) <- zip pools registrations]
```

rvl (commented):
Looks decent

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
KtorZ: > cardano-wallet-core-integration depend on cardano-wallet-shelley

Arf. I didn't realize that. That's quite lame indeed.

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
Anviking: 👍 now
```
cardano-wallet.integration:Info:43] [2020-06-23 09:01:53.84 UTC] Started a static server for http://localhost:60431/ at /private/var/folders/h_/mlzym27d5lz_fx3ph2d547700000gn/T/integration-62aa8274a6e3655b/stake-pool-2
[cardano-wallet.integration:Info:41] [2020-06-23 09:01:53.84 UTC] Started a static server for http://localhost:60430/ at /private/var/folders/h_/mlzym27d5lz_fx3ph2d547700000gn/T/integration-62aa8274a6e3655b/stake-pool-0
[cardano-wallet.integration:Info:46] [2020-06-23 09:01:53.84 UTC] Started a static server for http://localhost:60432/ at /private/var/folders/h_/mlzym27d5lz_fx3ph2d547700000gn/T/integration-62aa8274a6e3655b/stake-pool-1
```

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
Anviking: Went with `combineDbAndLsqData`. I can't find a guideline [here](https://github.com/input-output-hk/adrestia/wiki/Coding-Standards), but I thought `Lsq` was preferred to `LSQ`.

We do use `DB` a lot in our code.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
Anviking: Fixed.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
Anviking: `zip` feels a bit wrong to me, but sure, it does look neat

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
Anviking: Fixed d31e78541651f28bc38b04be24f3122c1712e226

Anviking:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
Anviking: In jormungandr we required the DB to be in sync with the node. For haskell we don't.

If the user ends up with a synced node and no stake-pools.sqlite, and lists pools, he will see pools based on LSQ data from the node tip. If he were to try to delegate, I fear the wallet wouldn't let him, claiming that the pool he selected doesn't exist.

It just seems weird to me that we would treat the node as the source of [truth](https://github.com/input-output-hk/cardano-wallet/blob/2f1a21f67f1ab3f5ae8e57da5c97842bf0d0532d/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs#L201), but then validate the delegation choice based a different one.

Anviking:
bors try

Anviking:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
KtorZ: > If he were to try to delegate, I fear the wallet wouldn't let him, claiming that the pool he selected doesn't exist.

Which is fine, the pool worker won't be _that_ behind and will catch up quite quickly. We can leave it as it is; if the networking layer is fast enough and reliable, then it _should be okay_. Let's see.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -415,68 +426,107 @@ spec = do
             , expectErrorMessage $ errMsg403DelegationFee fee
             ]
 
-    it "STAKE_POOLS_LIST_01 - List stake pools" $ \ctx -> do
-        eventually "Listing stake pools shows expected information" $ do
-            r <- request @[ApiStakePool] ctx
-                (Link.listStakePools arbitraryStake) Default Empty
+    let fixTypeInference = True `shouldBe` True
```
KtorZ: You just need a `@IO` down below on `request`

lib/shelley/test/data/cardano-node-shelley/genesis.yaml
```diff
@@ -6,7 +6,7 @@ protocolParams:
   protocolVersion:
     minor: 0
     major: 0
-  decentralisationParam: 0.97 # means 3% decentralised
+  decentralisationParam: 0.1 # means 90% decentralised
```
KtorZ: This was very high to avoid too frequent rollbacks until we have a proper transaction scheduler. I don't think you'll get anything merged with 90% decentralization :)

KtorZ: We can try, but I am afraid this is going to block many PRs after that :/

KtorZ (approved):
reviewed, tweaked a few things :+1:

PR #1756, opened 2020-06-13
Tx fee estimation edge cases bug

paweljakubas:
finish integration test

# Issue Number

#1740

# Overview

- [x] I have pin down the edge value characteristic in integration test (by bisection). This is different from Shelley testnet value reported by @johnalotoski  and @piotr-iohk , ie., `165281`
- [x] I have added integration test that reproduces issue
- [x] I have fixed the issue

# Comments
Running : 
```
stack test cardano-wallet-shelley:integration --ta '-m "TRANS_ESTIMATE_08 - Fee estimation edge case"'
```

My observations:
1. feeMin = 129 000
2. When estimating with payload amount, we have:
(a) payload < walletBalance - 129 000 => fee est = 129 000
(b) payload is (walletBalance - 129 000, walletBalance - 122 500) => fee est = 129 000 - (payload - 129 000)
(c) payload > walletBalance - 122 500 => fee est = walletBalance + (payload - 122 500 )

So when (a) fee estimation is correct, then when (b) it is decreasing linearly until we want to send `walletBalance - 122500`. Then it is increasing linearly in (c). I encoded those thresholds in integration test 

Investigation details: 
we have 3 cases (the last two give wrong fee):
1 the last proper one when 
```
CoinSel {inputs=[5_000_000], outputs=[4_871_000], change=[129_000]} this goes to rebalanceChangeOutputs twice 
-> branch : φ_original > δ_original && not (null (change s))
φ_original=129_000 δ_original=0 (change s)=[129_000]
-> branch : φ_dangling >= δ_original && φ_dangling > δ_dangling
φ_original=122_500 δ_original=129_000 φ_dangling=129_000 δ_dangling=122_500 (splitChange extraChng (change s))=[6_500]
So in senderPaysFee we have in output CoinSel {inputs=[5_000_000], outputs=[4_871_000], change=[]} as remFee=0
As a result, feeBalance is 5_000_000 - 4_871_000= 129_000 (OK)
```
2. First not proper (type 1):
```
CoinSel {inputs=[5_000_000], outputs=[4_871_001], change=[128_999]} this goes to rebalanceChangeOutputs twice 
-> branch : φ_original > δ_original && not (null (change s))
φ_original=129_000 δ_original=0 (change s)=[128_999]
-> branch : φ_dangling >= δ_original && φ_dangling > δ_dangling
φ_original=122_500 δ_original=128_999 φ_dangling=129_000 δ_dangling=122_500 (splitChange extraChng (change s))=[6_499]
[...]

Issue #1740, opened 2020-06-10
 Tx fee estimation is off when tx amount approaches wallet's balance


johnalotoski:
# Context

1) When calculating a transaction fee, if the amount to pay is set to the same value as the wallet holds (in a single address in this test case) or close to it, the TX fee is about as large as the wallets value.  This likely has no functional impact as it looks like these TXs should fail anyway due to insufficient fee ADA.  But in case it impacts something somewhere else, just FYI.

2) Network discrimination for the send to address is not tested for when calculating fees.  This might be by design and not an issue.  Just FYI

| Information  | -  |
| --- | --- 
| Version      |  2020.6.5 (git revision: 5290acf902da32e1e541b5dc227629c9a300f1be) |
| Platform     | Linux |
| Installation | Nix build |

# Steps to Reproduce

1) Check the wallet balance
```
$ cardano-wallet-shelley wallet get $WALLETID
...
        "total": {
            "quantity": 99627083,
            "unit": "lovelace"
        },
        "available": {
            "quantity": 99627083,
            "unit": "lovelace"
        }
...
```

2) Estimate fees by CLI or API:
```
# Set the payment about ~0.13 ADA lower than the wallet value; get a large estimated fee
#
$ cardano-wallet-shelley transaction fees $WALLETID --payment 99500000@$SEND_TO
Ok.
{
    "estimated_min": {
        "quantity": 99665281,
        "unit": "lovelace"
    },
    "estimated_max": {
        "quantity": 99665281,
        "unit": "lovelace"
    }
}

# Set the payment about ~1.13 ADA lower than the wallet value; get an accurate fee
#
$ cardano-wallet-shelley transaction fees $WALLETID --payment 98500000@$SEND_TO
Ok.
{
    "estimated_min": {
        "quantity": 168141,
        "unit": "lovelace"
    },
    "estimated_max": {
        "quantity": 168141,
        "unit": "lovelace"
    }
}
```

3) Perhaps not an issue, the network discriminator of the send to address doesn't appear to be checked during fee estimates.  Repeat one of the above fee estimates, this time flipping the second digit of the address from a `1` to a `0` or visa-versa.  The command completes the same way, successfully, despite one of those addresses not being valid on the network.
```
$ cardano-wallet-shelley transaction fees $WALLETID \
  --payment 99500000@$SEND_TO_WITH_MODIFIED_SECOND_DIGIT
Ok.
{
    "estimated_min": {
        "quantity": 99665281,
[...]

PR #1782, opened 2020-06-22
Fix fee estimation error handler (cannot cover fee) 

KtorZ:
# Issue Number

#1740 

# Overview

- 5f6a28f54f3e089bf8b279a423d2944c4edc2628
  :round_pushpin: **remove unused 'ErrInvalidTxOutSubmit' data-type from Shelley transaction layer**
  
- 41530d62058ac87c37468c21dcb8bf198fe93b02
  :round_pushpin: **move error handler one layer down, out of the API layer**
  So that it can be more tested more easily at the unit level.

- d9d2cd022e9ed8606c2a1f37089d9a037c985e59
  :round_pushpin: **write unit test to illustrate #1740 at the unit-level.**
  The test currently fails, so the regression is indeed identified.

- 0bd806a7acea3c736e8706c5c3822f6e437de403
  :round_pushpin: **slightly change type signature of 'estimateFeeForCS' to allow mounting the error handler on each selection.**
  
- ed03123f42c45107d070a7fec4abf45686235561
  :round_pushpin: **fix error handler in estimated fee, take recipients into account.**

# Comments

The issue wasn't actually in the coin selection or fee estimation procedures, but in an extra error handler that was set only in the API layer. 
This PR supersedes #1756

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -127,6 +127,9 @@ module Cardano.Wallet
     -- ** Fee Estimation
     , FeeEstimation (..)
     , estimateFeeForCoinSelection
+    , feeOpts
+    , coinSelOpts
+    , handleCannotCover
 
     -- ** Transaction
     , forgetPendingTx
@@ -275,6 +278,7 @@ import Cardano.Wallet.Primitive.Types
     , TxMeta (..)
     , TxOut (..)
     , TxStatus (..)
+    , UTxO
     , UTxOStatistics
     , UnsignedTx (..)
     , WalletDelegation (..)
@@ -309,7 +313,14 @@ import Control.Monad.IO.Class
 import Control.Monad.Trans.Class
     ( lift )
 import Control.Monad.Trans.Except
-    ( ExceptT (..), except, mapExceptT, runExceptT, throwE, withExceptT )
+    ( ExceptT (..)
+    , catchE
+    , except
+    , mapExceptT
+    , runExceptT
+    , throwE
+    , withExceptT
+    )
 import Control.Monad.Trans.Maybe
     ( MaybeT (..), maybeToExceptT )
 import Control.Monad.Trans.State.Strict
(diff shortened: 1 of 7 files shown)


piotr-iohk:
I have tried this branch on Shelley **QA** testnet and here's the result:

wallet balance = 5 000 000 lovelace

estimate fee for 1 => **168141** lovelace
estimate fee for 1000000 => **168141** lovelace
estimate fee for 4834710 => **165290** lovelace
estimate fee for 4834718 => **165282** lovelace

estimate fee for 4834719 => 165281 lovelace
estimate fee for 4834720 => 165281 lovelace
estimate fee for 5000000 => 165281 lovelace
estimate fee for 5000001 => not_enough_money

This is slightly different than -> https://github.com/input-output-hk/cardano-wallet/pull/1756#issuecomment-644115752 (where fee was rather flat = 168141, I think it was the same for any value between 0 and 5000000). Not sure if that's good or bad, but sharing this observation.

Also the previous result is from Shelley testnet and this one is from Shelley QA tesntet, but they both have the same in genesis:

```
    "minFeeA": 44,
    "minFeeB": 155381, 
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1192,8 +1204,27 @@ estimateFeeForPayment
 estimateFeeForPayment ctx wid recipients = do
     (utxo, txp) <- withExceptT ErrSelectForPaymentNoSuchWallet $
         selectCoinsSetup @ctx @s @k ctx wid
-    estimateFeeForCoinSelection $
-        selectCoinsForPaymentFromUTxO @ctx @t @k @e ctx utxo txp recipients
+    let selectCoins =
+            selectCoinsForPaymentFromUTxO @ctx @t @k @e ctx utxo txp recipients
+    estimateFeeForCoinSelection $ (Fee . feeBalance <$> selectCoins)
+        `catchE` handleCannotCover utxo recipients
```
paweljakubas, line 1210: this addresses what should be addressed :-)

paweljakubas (approved):
The PR targets deficiency in how api layer treats `ErrCannotCoverFee`. It is nice that the burden of testing is put in unit testing layer rather than integration layer. I second the question raised by Piotr about fee estimation, rather than being flat = 168141 in (0, 5000000) it is constant up to some value (168141),then abruptly decreases to 165290 until it reaches 165281 and becomes constant. The two thresholds are set by : 
```
         let (feeMin, _) = ctx ^. #_feeEstimator $ PaymentDescription                                                                                                                                          
                 { nInputs = 1                                                                                                                                                                                  
                 , nOutputs = 1                                                                                                                                                                                 
                 , nChanges = 1                                                                                                                                                                                 
                 }
```
and 
```
         let (feeMin', _) = ctx ^. #_feeEstimator $ PaymentDescription                                                                                                                                          
                 { nInputs = 1                                                                                                                                                                                  
                 , nOutputs = 1                                                                                                                                                                                 
                 , nChanges = 0                                                                                                                                                                                 
                 } 
```
I am not saying this is incorrect but maybe unexpected for people playing a lot with api (like me:-)

KtorZ:
Indeed, without diving much further in:

- That the fees become "constant" when reaching a certain window close to the wallet balance is expected. This is when the "special" handler try to re-conciliate the "ErrCannotCoverForFee" coming from the fee balancing algorithm. When estimating fee, returning this error is sort of misleading, so instead, we give a bets-effort estimation using the wallet balance, the requested output and the missing amount for the "cover for fee". It's basically a computation like: "fee = balance - output + missing". If output increases, then missing also decreases of the same amount so it balances itself out (no pun intended) and remains constant once we are in this special case.

- The fee `168141` is the minimal fee for a transaction with one input, one output and one change output. However, when the output amount is big enough (i.e. close to the wallet's balance), there's no change output to leave. So the fee looks slightly smaller: `165281`, the difference is the cost of a change output, ~61 bytes. Arguably, we could make the estimation a bit better here and return the cost _assuming a change output_ because a transaction is actually very likely to have one. It might be confusing for people unless they're sending _just_ the right amount. Imagine the following scenario:

    - I have `5000000`, trying to send `4834720`. The API tells me that estimated fee are `165281` so I send `100` to my wallet, you know, because I like it round although I only need 1 lovelace. 
    - Now, the transaction requires a change output and fees are `168141`. So I still don't have enough on my wallet. 

Alternatively, we could return `feeMin` / `feeMax` where `feeMin` is the amount without change and `feeMax` the amount with change.

PR #1783, opened 2020-06-22
Test that pools have non-myopic member rewards

Anviking:
# Issue Number

#1276 

# Overview

- [x] Test that pools have non-zero non-myopic member rewards in `GET /stake-pools`
- [x] Test that pools are sorted by descending rewards
- [x] Add `MsgWillQueryRewardsForStake` trace

# Comments

### DIFF
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
@@ -367,6 +367,7 @@ withNetworkLayer tr np addrInfo versionData action = do
         stakeMap <- fromPoolDistr <$> handleQueryFailure
             (queue `send` CmdQueryLocalState pt OC.GetStakeDistribution)
         let toStake = Set.singleton $ Left $ toShelleyCoin coin
+        liftIO $ traceWith tr $ MsgWillQueryRewardsForStake coin
         rewardsPerAccount <- fromNonMyopicMemberRewards <$> handleQueryFailure
             (queue `send` CmdQueryLocalState pt (OC.GetNonMyopicMemberRewards toStake))
         pparams <- handleQueryFailure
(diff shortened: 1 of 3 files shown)


lib/shelley/test/data/cardano-node-shelley/genesis.yaml
```diff
@@ -19,11 +19,11 @@ protocolParams:
   keyDeposit: 0
   keyDecayRate: 0
   nOpt: 100
-  rho: 0
+  rho: 0.178650067
```
Anviking: Documentation: https://github.com/input-output-hk/cardano-ledger-specs/blob/036ff8223310ecdaf7f0ed8c211c691cb0dcb8f3/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/PParams.hs#L127,

(This particular value is copied from FF, but increased by some factors of 10 when it didn't work)

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -790,6 +792,8 @@ instance ToText NetworkLayerLog where
             [ "Destroying cursor connection at"
             , T.pack (show threadId)
             ]
+        MsgWillQueryRewardsForStake c ->
+            "Will query non-myopic rewards using the stake " <> pretty c
```
Anviking, line 796: Is the balance private and not to be logged? (not necessarily the user's balance, but likely correlated)

KtorZ: Nope, fine. Balances are already logged as part of the wallet. Current logs aren't much private.

KtorZ (approved):

KtorZ:
(trying the branch locally)

KtorZ:
Locally, the test pass when run in isolation. But it fails when run with the entire test-suite. One of two conditions is not verified.

Anviking:
It seems the pools all have the same non-myopic member rewards. Perhaps leaving them running for long removes the noise.

I tried giving one pool twice the amount of pledge/stake, and it seemed to work. But perhaps giving each pool different costs, or margins would be neater 🤔

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

PR #1748, opened 2020-06-12
Remove redundant integration test jobs on hydra

rvl:
### Overview

Remove redundant integration test jobs on hydra
    
We don't need to run integration tests for both linux glibc and musl. Choose musl, because that's what our release binaries are.

And remove the windows cardano-node integration test jobs, because ouroboros-network doesn't fully work under Wine.

- [PR Jobset](https://hydra.iohk.io/jobset/Cardano/cardano-wallet-pr-1748#tabs-jobs)

~And while we're here, update Haskell.nix and iohk-nix, and fix resulting errors.~ ⇒  #1791

### Comments

- First attempt editing `nix/haskell.nix` did not work because the jobset evaluator is always a linux system. The job filtering needs to be done in `release.nix`.

PR #1788, opened 2020-06-23
Bump cardano-node to 1.14.1

Anviking:
# Overview

- c9e5fad9897f09080389894fb0210bf05d6aa1ce
  :round_pushpin: **Bump cardano-node to 1.14.0**
  
- 289b857cc7265ea376cc041e50b96c7517119214
  :round_pushpin: **iohk-nix: use crypto overlay for custom libsodium**
  
- 0e3a1c2ace0be1ce18e0adb401933c59a9c5c31d
  :round_pushpin: **cardano-node: fix version**
  
- cd5138102b78477628e0f3b1120c25fdd091286a
  :round_pushpin: **Make cardano-wallet-shelley compile**
  
- a1982bf8511d15da02da9158302b5bf785dd9900
  :round_pushpin: **fixup: tweak snapshot revision**
  
- cf2017566f8ea9ed80bcf285a1cf93139311ed01
  :round_pushpin: **Update compatibility matrix in README.md**
  
- e50d75854443ad0e45e710aba244f465d75c4225
  :round_pushpin: **wip: use 28-bit digests for addresses and regen setup**
  - Regenerated faucets
- Regenerated integration setup using the regenerate.sh script
- Regenerated stake pool setup faucet funds

- 4c6b8fb1b3cac48913181e6a6b8e8d7540f5710d
  :round_pushpin: **Add libsodium to stack --nix shell**
  
- a384d45155c20d6c370688eb286b52abe8561687
  :round_pushpin: **force the libsodium fork to be bundled with cardano-crypto-praos**
  
- c242a3edae45d32d185a56abf8bde22e599053a4
  :round_pushpin: **add disclaimer to genesis.yaml & changed initialFunds to be a list of tuples**
  Why? See the disclaimer.

- 01c98f535d19cb25a2c8531908338f5b57d1fd80
  :round_pushpin: **re-generate bft leaders keys, and use them as the actual keys for launching the bft leader!**
  I've also dropped the dependency on 'yq' which seems to have a different interface on mac than on Linux. At least, the --yaml-output
isn't a thing on the Ubuntu version. In the end, plain JSON is valid YAML, so we can just output JSON here.

- 6b0b943c21d6fe16751ebad548ec1e72a861518b
  :round_pushpin: **make all key hashes 28-byte digests in shelley**
  
- 43f5eec45d4e93d0083b4c4618c3a40cf7ba6404
  
- d70d44bb0ec82704f9bb2232ac8aefc88d5e42a4
  :round_pushpin: **hopefully add libsodium DLLs to the Nix configuration**
  
- b5ae93bbeec1c81f170f2085a4e65bb638d16075
  :round_pushpin: **remove 'golden' comparing addresses tests now obsolete**
  The address format has changed. Also, I am not quite sure if these were testing anything in the first place...
Seems to be comparing the wallet implementation with itself.

- d498b64e7913b6112f1f96a375237f584ad23472
  :round_pushpin: **Clean up nix/haskell.nix a little**
  
- c11f00356d7488d7734e7c25f32be85f267b70d3
  :round_pushpin: **Update cardano-node to 1.14.1 tag**
  
- 76a0c539cb69baa4dfe9b459ebe44f628ad2b63b
  :round_pushpin: **Provide libsodium.dll to cardano-wallet-byron tests**
  The DLL is needed to run cardano-node.

- 761e2175029fe2f3515b9202190800646d25d4db
  :round_pushpin: **Bump cardano-node niv dep so that libsodium is included.**
  
- 213699dcf578ee410a44d4e7d569cd79eb6d5fa3
  :round_pushpin: **increase number of UTxO in assertion for big wallet migration**
  We just changed the address format to 28-byte, so it seems like transactions have become a bit less costly,
and as a result, we are able to migrate one extra UTxO! (or more exactly, to have a bigger value)

# Comments

- Pending that it works


KtorZ:
We are missing libsodium in the nix setup.

rvl:
@disassembler There were some issues resulting from updating iohk-nix - I fixed two. But one of the wine tests will fail. You can see it already on the hydra build of #1748.

lib/shelley/test/data/cardano-node-shelley/genesis.yaml
```diff
@@ -1200,519 +1198,3 @@ initialFunds:
   6114de519ddd182962ed9ed061afb654911401415e17d0f934e3c83da3c6ee2cde: 100000000000
   613139b22b1d328000063beb4b8d59935674458f74168459d07fd4ec766328163b: 100000000000
   614f27f4008ffa5454efeb925b98f743e144f3780969d36423e7940a8ba3f5b6d3: 100000000000
-
```
KtorZ, line 1204: Why are these removed? Oversight?

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -768,44 +768,44 @@ takeFaucet dir = takeMVar faucets >>= \case
...
           ]
       )
-    , ( "82a8981af11569eef547e012232f5306b85016d27aff4c1bbd5801800f83baab#0"
+    , ( " 5820e96f612fbff3df3d8eef4ea3a07e3dc98769020545ced0167998a85a4cc50aa7"
       , unlines
           [ "type: Genesis UTxO signing key"
           , "title: Genesis initial UTxO key"
           , "cbor-hex:"
-          , "  58207b9de02388f78146cb53e4fb4a4d8917ac367cdfe0ff1d4008cc9da52af6ac74"
+          , "  5820e96f612fbff3df3d8eef4ea3a07e3dc98769020545ced0167998a85a4cc50aa7"
```
Anviking, line 849: @KtorZ Is there another place that needs updating? With this commit there are no errors logged, but the pools never get any `Stake frac`, and the setup times out.

lib/shelley/test/data/cardano-node-shelley/genesis.yaml
Anviking: Thanks — oversight.

I regenerated the faucet addresses since I believe we now have the new addresses with shorter hashes. But we need to change the address generation first, for the regeneration to be effective… and indeed, make sure these are preserved.

KtorZ: > But we need to change the address generation first

Indeed, we need to switch to 28-bit digest for Shelley addresses.

> make sure these are preserved.

I don't understand?

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
KtorZ: The key associated to these should match what's in the genesis file. Note that the TxIn of this particular one is not right.

lib/shelley/test/data/cardano-node-shelley/genesis.yaml
Anviking: I switched to 28 bit (actually — you mean "byte", right?), and regenerated everything (including these special mnemonics) in 
eae65e2

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
Anviking: Still doesn't work though 🤔

KtorZ (approved):
reviewed, currently running tests locally.

rvl:
Some windows tests are still failing because `libsodium-23.dll` is missing from the bin directory of _cardano-node.exe_ and _cardano-cli.exe_.
Would someone @input-output-hk/devops  be able to help fix that in cardano-node, by adding a windows-only postInstall?

KtorZ:
bors try

disassembler:
bors try

PR #1790, opened 2020-06-24
Fix logging in shelley integration test cluster

rvl:
### Issue Number

ADP-302 / #1750

### Overview

1. Block forging and adoption was not being logged by the integration tests cluster. Fixed it. The default minimum severity for stdout remains as Error. The default minimum severity for the log file is Debug.

2. This adds a passive node which the wallet connects to, rather than using the bft node. My hope is that this will cause transactions to be broadcast to _all_ stake pools, meaning that transactions won't be lost with rollbacks. But in any case, this updated topology makes for a more "realistic" integration test.


lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -409,6 +411,44 @@ withBFTNode tr baseDir (NodeParams severity systemStart (port, peers)) action =
     name = "bft"
     dir = baseDir </> name
 
+withPassiveNode
```
KtorZ, line 414: It could be nice to add a comment explaining _why_ we are using this indirection.

KtorZ (approved):

Issue #1789, opened 2020-06-23
Get transaction by its id


paweljakubas:
# Context

I want to get information about a transaction based on an ID

# Decision

# Acceptance Criteria

1. the endpoint capable of doing that is added.
2. integration test illustrating transaction info lifecycle from valid id and invalid id behavior is added

---

# Development 

# QA 

A number of things were added in this task that are worth checking : 
1. getTx in DBlayer and corresponding property tests
2. get tx by id endpoint in wallets and byron-wallets family - here integration tests 
3. CLI support - here integration tests
4. swagger was updated

piotr-iohk:
Looking good. I have added additional checks in integration tests for Byron wallets (API & CLI) in #1804. 

Also added tests for "get tx by id" into "nightly shelley testnet tests". (And hopefully fixed them to run -> https://github.com/piotr-iohk/cardano-wallet-rb/actions?query=workflow%3ANightly)

piotr-iohk:
lgtm

PR #1795, opened 2020-06-24
Get tx by id

paweljakubas:
# Issue Number

#1789

# Overview

- [x] I have added getTransaction endpoint, and all needed implementations, errors
- [x] I have updated swagger
- [x] I have added integration test
- [x] I have added CLI support
- [x] I have updated CLI unit tests
- [x] I have added CLI integration tests
- [x] I have added support in ByronTransaction 
- [x] I have added getTx to DB
- [x] I have implemented getTx in Sqlite
- [x] I have implemented getTx in MVar
- [x] I added properties for getTx and improved on MVar impl as a result

# Comments

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
@@ -103,6 +103,7 @@ import Test.Integration.Framework.TestData
     , errMsg403UTxO
     , errMsg403WrongPass
     , errMsg404CannotFindTx
+    , errMsg404NoTransaction
     , errMsg404NoWallet
     )
 import Web.HttpApiData
@@ -1209,6 +1210,27 @@ spec = do
         r <- request @(ApiTransaction n) ctx link Default Empty
         expectResponseCode @IO HTTP.status404 r
         expectErrorMessage (errMsg404NoWallet $ w ^. walletId) r
+
+    it "TRANS_GET_03 - Using wrong transaction id" $ \ctx -> do
+        (wSrc, wDest) <- (,) <$> fixtureWallet ctx <*> emptyWallet ctx
+        -- post tx
+        let amt = (1 :: Natural)
+        rMkTx <- postTx ctx
+            (wSrc, Link.createTransaction @'Shelley, "cardano-wallet")
+            wDest
+            amt
+        verify rMkTx
+            [ expectSuccess
+            , expectResponseCode HTTP.status202
+            , expectField (#direction . #getApiT) (`shouldBe` Outgoing)
+            , expectField (#status . #getApiT) (`shouldBe` Pending)
+            ]
+
+        let txid =  Hash $ BS.pack $ replicate 32 1
+        let link = Link.getTransaction wSrc (ApiTxId $ ApiT txid)
+        r <- request @(ApiTransaction n) ctx link Default Empty
+        expectResponseCode @IO HTTP.status404 r
+        expectErrorMessage (errMsg404NoTransaction $ toText txid) r
 
 
     it "TRANS_DELETE_01 -\
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
@@ -29,7 +29,7 @@ import Cardano.Wallet.Api.Types
     , time
     )
 import Cardano.Wallet.Primitive.Types
-    ( Direction (..), SortOrder (..), TxStatus (..), WalletId )
+    ( Direction (..), Hash (..), SortOrder (..), TxStatus (..), WalletId )
 import Control.Monad
     ( forM_ )
 import Data.Aeson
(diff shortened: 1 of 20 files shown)


paweljakubas:
bors try

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -1148,6 +1149,89 @@ spec = do
...
+            ]
+
+        -- Verify Tx in source wallet is Outgoing and Pending
+        let linkSrc = Link.getTransaction @'Shelley wSrc (ApiTxId txid)
+        r1 <- request @(ApiTransaction n) ctx linkSrc Default Empty
+        verify r1
+            [ expectResponseCode HTTP.status200
+            , expectField (#direction . #getApiT) (`shouldBe` Outgoing)
+            , expectField (#status . #getApiT) (`shouldBe` Pending)
+            ]
```
KtorZ: Note that this is subject to race-condition, especially with a slot-length of 0.2s. There are good chances that the transaction is already inserted at this stage. I'd remove this assertion and the next one about the wallet balance (which is collateral to this test), and simply keep the last two assertions of this scenario behind an `eventually`. 

These would be sufficient for this test to show that transactions can be obtained using their ID :+1:

lib/core-integration/src/Test/Integration/Scenario/CLI/Shelley/Transactions.hs
```diff
@@ -839,7 +840,112 @@ spec = do
               oJson2 <- expectValidJSON (Proxy @([ApiTransaction n])) o2
               length <$> [oJson1, oJson2] `shouldSatisfy` all (== 0)
 
-    it "TRANS_DELETE_01 - Cannot forget pending transaction when not pending anymorevia CLI" $ \ctx -> do
+    it "TRANS_GET_01 - Can get Incoming and Outgoing transaction" $ \ctx -> do
```
KtorZ, line 843: Same comment than for the API scenario applies here too :+1:

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1536,6 +1539,29 @@ listTransactions ctx wid mStart mEnd order = db & \DBLayer{..} -> do
...
+    -> Hash "Tx"
+    -> ExceptT ErrGetTransaction IO TransactionInfo
+getTransaction ctx wid tid = db & \DBLayer{..} -> do
+    let pk = PrimaryKey wid
+    mapExceptT atomically $ do
+        _ <- withExceptT ErrGetTransactionNoSuchWallet $
+            withNoSuchWallet wid $ readCheckpoint pk
+        txs <- lift $ readTxHistory pk Descending wholeRange Nothing
+        case (filter txPresent txs) of
+            [] -> do
```
KtorZ: Arg! This defies a bit the purpose of such endpoint. Being able to fetch a single transaction should be quick and swift; here we are loading the entire transaction history in memory and filtering it by hand! For exchanges processing MANY transactions a day, that's not going to work (especially because they'll be the primary consumer of this endpoint). 

This should be replaced by a single SQL query to retrieve a single transaction from the database :+1:

KtorZ (changes requested):
Nice work, very clean. 
Yet we gotta need a new database query for fetching a single transaction from the database; we can't possibly load the entire history in memory.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
paweljakubas: agree here. this poses potential race condition - corrected in https://github.com/input-output-hk/cardano-wallet/pull/1795/commits/d898b0d84bba687166dcd71bfd440d29cd26f063

lib/core-integration/src/Test/Integration/Scenario/CLI/Shelley/Transactions.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/1795/commits/d898b0d84bba687166dcd71bfd440d29cd26f063

lib/core/src/Cardano/Wallet.hs
paweljakubas: introduced getTx in  https://github.com/input-output-hk/cardano-wallet/pull/1795/commits/d898b0d84bba687166dcd71bfd440d29cd26f063 and https://github.com/input-output-hk/cardano-wallet/pull/1795/commits/9f5eba327dde35772a26d451d35c0718f3524341

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1536,6 +1539,27 @@ listTransactions ctx wid mStart mEnd order = db & \DBLayer{..} -> do
...
+    let pk = PrimaryKey wid
+    res <- lift $ atomically $ runExceptT $ getTx pk tid
+    case res of
+        Left err -> do
+            throwE (ErrGetTransactionNoSuchWallet err)
+        Right Nothing -> do
+            let err' = ErrNoSuchTransaction tid
+            throwE (ErrGetTransactionNoSuchTransaction err')
+        Right (Just tx) ->
+            pure tx
```
KtorZ, line 1574: :+1:

KtorZ (approved):

PR #1797, opened 2020-06-25
Implement log rotation in docker-compose file

rhyslbw:
See https://github.com/input-output-hk/cardano-graphql/pull/197 for context

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,6 +10,12 @@ services:
       - node-ipc:/ipc
       - node-config:/nix/store
     restart: on-failure
+    logging:
+      driver: "json-file"
+      options:
+        compress: "true"
+        max-file: "10"
+        max-size: "50m"
 
   cardano-wallet:
     image: inputoutput/cardano-wallet:latest
@@ -28,6 +34,12 @@ services:
       CMD: "cardano-wallet-byron serve --node-socket /ipc/node.socket --database /wallet-db --listen-address 0.0.0.0"
       NETWORK:
     restart: on-failure
+    logging:
+      driver: "json-file"
+      options:
+        compress: "true"
+        max-file: "10"
+        max-size: "50m"
 
 volumes:
   node-mainnet-db:


rvl (commented):
How long does cardano-node take to emit 2000k of logs under this configuration?
You would want it to retain logs for at least a couple of days.

rhyslbw:
@rvl Good call, my assessment was based on my own config, so am now running the stack against `NETWORK=mainnet` for a few hours to determine a more suitable default.

rhyslbw:
`cardano-node` emits ~ 3.75MB per epoch on the mainnet, so the resulting mainnet sync from genesis emits around 750MB of logs within the first few hours. What's your preference here? Dedicating that chunk of disk seems unreasonable when all we're doing is fast-forwarding through time, with the relevance of _old_ epochs having the same meaning. I expect critical errors in the sync should be on disk if it's needing to be debugged. To ensure the current and previous epochs, plus their transitions are included, I propose: 

``` json
  max-size: "400k"
  max-file: "20"
```

rhyslbw:
The base for `cardano-wallet` is the clock check, 5mins apart:
``` console
cardano-wallet_1  | [cardano-wallet.ntp-client:Info:14] [2020-06-29 01:41:17.78 UTC] query to ntp client invoked
cardano-wallet_1  | [cardano-wallet.ntp-client:Info:14] [2020-06-29 01:41:18.93 UTC] local clock is drifting by -27088μs
cardano-wallet_1  | [cardano-wallet.ntp-client:Info:14] [2020-06-29 01:46:18.96 UTC] query to ntp client invoked
cardano-wallet_1  | [cardano-wallet.ntp-client:Info:14] [2020-06-29 01:46:20.17 UTC] local clock is drifting by -20103μs
```

Do you have a simulation that can provide some insight into the `cardano-wallet` logging behaviour, or is what I've submitted here satisfactory?

rvl:
Can we just say that 500MB of logs is too much. Anything less than 500MB of logs is OK. This is 1/10th of the mainnet chain db size.

So make it 10 rotations of 50MB each. And also there is a log compression flag which may help reduce impact of logs: https://docs.docker.com/config/containers/logging/json-file/#options

rvl (approved):
Looks fine

Issue #1794, opened 2020-06-24
Wallet on new testnet fails with UNIQUE constraint failed: pool_registration.pool_id


Anviking:
# Context
 
Using:
- `shelley_testnet` https://hydra.iohk.io/job/Cardano/iohk-nix/cardano-deployment/latest/download/1/index.html
- cardano-node 1.14.0

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | a1db69e710d492e4297cdea4c4d8c73b47315495 |
| Platform     |           |
| Installation |             |

# Steps to Reproduce

1. `cardano-node run --database-path db --topology shelley_testnet-topology.json --socket-path socket --config shelley_testnet-config.json --port 7776`

2. `cardano-wallet-shelley serve --node-socket socket --testnet shelley_testnet-genesis.json --database wallet-db`

## Expected behavior

- Stake pool monitoring worker runs smoothly

## Actual behavior

After some second the worker panics:

```
[cardano-wallet.pools-engine:Info:30] [2020-06-24 15:32:16.17 UTC] Discovered stake pool registration: Registration of 3d51d7f3 owned by [ed25519_pk1yeq965dl9fg47us5n7nvts63gg0van60ahpf40qyjfec7xhdv8e]
[cardano-wallet.pools-engine:Info:30] [2020-06-24 15:32:16.18 UTC] Discovered stake pool registration: Registration of 31d6318b owned by [ed25519_pk170dxgh2evajf994s7t2cj4vszrnu7fvsk6xpay4gwt6y70hvgx7]
[cardano-wallet.pools-engine:Info:30] [2020-06-24 15:32:16.18 UTC] Discovered stake pool registration: Registration of 27e33e8d owned by [ed25519_pk178axspgs7ktsxk8d4xmrqh238p95gpthvwk5npd502t5gg5xv3g]
[cardano-wallet.pools-engine:Info:30] [2020-06-24 15:32:16.25 UTC] Applying blocks [0.20360 ... 1.18740]
[cardano-wallet.pools-engine:Info:30] [2020-06-24 15:32:16.25 UTC] Discovered stake pool registration: Registration of afd3cbad owned by [ed25519_pk1epdkas5l7t3nw0ntdhc30gd4hyl5eyz9qsr0ld4ar628gm26jdm]
[...]

PR #1798, opened 2020-06-25
Fix pool listing on testnet

Anviking:
# Issue Number

#1794 

# Overview

- [x] Support pool-reregistration (instead of crashing)
- [x] Log pool retirements (But not doing anything else with them… yet)
- [x] Allow pools to have rewards but no stake (not sure why this happens 🤷‍♂️ ) Asked about it [here](https://input-output-rnd.slack.com/archives/CCRB7BU8Y/p1593085379334200)

# Comments

### DIFF
--- a/lib/core/src/Cardano/Pool/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Pool/DB/Sqlite.hs
@@ -86,6 +86,7 @@ import Database.Persist.Sql
     , insert_
     , putMany
     , rawSql
+    , repsert
     , selectFirst
     , selectList
     , (<.)
@@ -208,7 +209,8 @@ newDBLayer trace fp = do
             let poolPledge_ = getQuantity poolPledge
             let poolMetadataUrl = fst <$> poolMetadata
             let poolMetadataHash = snd <$> poolMetadata
-            insert_ $ PoolRegistration
+            deleteWhere [PoolOwnerPoolId ==. poolId]
+            _ <- repsert (PoolRegistrationKey poolId) $ PoolRegistration
                     poolId
                     point
                     poolMarginN
--- a/lib/core/src/Cardano/Pool/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Pool/DB/Sqlite.hs
@@ -209,8 +209,8 @@ newDBLayer trace fp = do
             let poolPledge_ = getQuantity poolPledge
             let poolMetadataUrl = fst <$> poolMetadata
             let poolMetadataHash = snd <$> poolMetadata
-            deleteWhere [PoolOwnerPoolId ==. poolId]
-            _ <- repsert (PoolRegistrationKey poolId) $ PoolRegistration
+            deleteWhere [PoolOwnerPoolId ==. poolId, PoolOwnerSlot ==. point]
+            _ <- repsert (PoolRegistrationKey poolId point) $ PoolRegistration
                     poolId
                     point
                     poolMarginN
@@ -219,10 +219,12 @@ newDBLayer trace fp = do
                     poolPledge_
                     poolMetadataUrl
                     poolMetadataHash
-            insertMany_ $ zipWith (PoolOwner poolId) poolOwners [0..]
+            insertMany_ $ zipWith (PoolOwner poolId point) poolOwners [0..]
 
         , readPoolRegistration = \poolId -> do
-            selectFirst [ PoolRegistrationPoolId ==. poolId ] [] >>= \case
+            let filterBy = [ PoolRegistrationPoolId ==. poolId ]
+            let orderBy  = [ Desc PoolRegistrationSlot ]
+            selectFirst filterBy orderBy >>= \case
                 Nothing -> pure Nothing
                 Just meta -> do
                     let PoolRegistration
@@ -240,7 +242,7 @@ newDBLayer trace fp = do
                     let poolMetadata = (,) <$> poolMetadataUrl <*> poolMetadataHash
                     poolOwners <- fmap (poolOwnerOwner . entityVal) <$> selectList
                         [ PoolOwnerPoolId ==. poolId ]
-                        [ Asc PoolOwnerIndex ]
+                        [ Desc PoolOwnerSlot, Asc PoolOwnerIndex ]
                     pure $ Just $ PoolRegistrationCertificate
                         { poolId
                         , poolOwners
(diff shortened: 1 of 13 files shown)


lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -208,7 +209,8 @@ newDBLayer trace fp = do
             let poolPledge_ = getQuantity poolPledge
             let poolMetadataUrl = fst <$> poolMetadata
             let poolMetadataHash = snd <$> poolMetadata
-            insert_ $ PoolRegistration
+            deleteWhere [PoolOwnerPoolId ==. poolId]
```
KtorZ: The delete is redundant with the use of `repsert`, isn't ?

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -1866,6 +1875,26 @@ instance Buildable PoolRegistrationCertificate where
         <> " owned by "
         <> build o
 
+data PoolRetirementCertificate = PoolRetirementCertificate
+    { poolId :: !PoolId
+
+    -- | TODO: Check what this EpochNo actually means.
+    --
+    -- But a guess would be that it is the /first/ epoch that the pool will be
+    -- retired in.
```
KtorZ: Indeed. it should be the epoch at which the pool becomes inactive.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -251,8 +253,17 @@ combineLsqData NodePoolLsqData{nOpt, rewards, stake} =
         , saturation = (sat s)
         }
 
-    rewardsButNoStake = traverseMissing $ \k r ->
-        error $ "Rewards but no stake: " <> show (k, r)
+    -- TODO: This case seems possible on shelley_testnet, but why, and how
+    -- should we treat it?
+    --
+    -- The pool with rewards but not stake didn't seem to be retiring.
```
KtorZ, line 259: Could it be because the leader schedule is based on the stake from the previous epoch? So a pool can have 0 stake now, but had some in the previous epoch and therefore, is getting rewards in this epoch.

KtorZ (approved):
LGTM

lib/core/src/Cardano/Pool/DB/Sqlite.hs
Anviking: No, this is removing the pool _owners_ (before we later `insertMany_ `)

The `repsert` is used with the pool _registration certificate_

Anviking: However — I think my solution here is flawed in the context of rollbacks :/

We need to preserve the old certificate, in case the new certificate gets rolled back.

Maybe we can get away with something simple though. Like just keeping both and reading the latest one, when we need to read it.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
Anviking: > So a pool can have 0 stake now

I think that I confirmed that pools in the stake distribution may have 0 stake. I.e. if this pool is entirely _missing_, something else is going on.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
KtorZ: Indeed.. rollbacks. What about simply changing the PrimaryKey to:

```
    Primary poolRegistrationPoolId poolRegistrationSlot
```

In the rare case we have two certificates at the same slot, then only the latest one will prevail (we still need the repsert here) as it should, and it'll preserve old certificates on past slots if there are any. 

We'll have to be careful when _reading_ the certificates from the db now, and make sure they're ordered by descending slot. As for the owners :/ ... it's more complicated.

Anviking: > As for the owners :/ ... it's more complicated

Could we not store the owners as some kind of comma-separated list as part of the `pool_registration` table?

Anviking: …or perhaps that's not possible.

KtorZ: :/ ... What about using MongoDB then? 

Not really, this is not quite the problem, and it won't solve the need for a migration. Ideally, we should also attach a slot to the pool owner, or a certificate number, so that when we fetch a certificate, we can fetch _just_ the owner corresponding to that certificate.

Anviking: Are you sure it is possible to use both slot and poolId as primary key, while using them as a foreign key at the same time?

I added (to the `PoolRegistration` table):
```
    Foreign PoolRegistration fk_registration_pool_id poolOwnerPoolId
    Foreign PoolRegistration fk_registration_slot poolOwnerPoolSlot
```

But get:
```

found 1 fkeys and 2 pkeys: fdef=ForeignDef {foreignRefTableHaskell = HaskellName {unHaskellName = "PoolRegistration"}, foreignRefTableDBName = DBName {unDBName = "pool_registration"}, foreignConstraintNameHaskell = HaskellName {unHaskellName = "fk_registration_pool_id"}, foreignConstraintNameDBName = DBName {unDBName = "pool_ownerfk_registration_pool_id"}, foreignFields = [], foreignAttrs = [], foreignNullable = False} pdef=CompositeDef {compositeFields = [FieldDef {fieldHaskell = HaskellName {unHaskellName = "poolRegistrationPoolId"}, fieldDB = DBName {unDBName = "pool_id"}, fieldType = FTTypeCon (Just "W") "PoolId", fieldSqlType = SqlOther "SqlType unset for poolRegistrationPoolId", fieldAttrs = ["sql=pool_id"], fieldStrict = True, fieldReference = NoReference, fieldComments = Nothing},FieldDef {fieldHaskell = HaskellName {unHaskellName = "poolRegistrationSlot"}, fieldDB = DBName {unDBName = "slot"}, fieldType = FTTypeCon (Just "W") "SlotId", fieldSqlType = SqlOther "SqlType unset for poolRegistrationSlot", fieldAttrs = ["sql=slot"], fieldStrict = True, fieldReference = NoReference, fieldComments = Nothing}], compositeAttrs = []}
```

KtorZ: Yes, you need the foreign key to match the primary key of the foreign table. See: https://github.com/input-output-hk/cardano-wallet/pull/1798/commits/81fe0c38b2db9a928c91af52b47729c70cc72438

KtorZ: I think this should do.

Anviking:
```
$ cardano-wallet-shelley serve --node-socket socket --testnet shelley_testnet-genesis.json --database wallet-db

[cardano-wallet.pools-engine:Error:30] [2020-06-25 22:10:54.66 UTC] 
  Unexpected error following the chain: 
    SQLite3 returned ErrorConstraint while attempting to perform step:
       UNIQUE constraint failed: pool_owner.pool_id, pool_owner.pool_owner, pool_owner.pool_owner_index
```

KtorZ:
Ah! crap, I forgot to add the `slot` to the primary key of PoolOwner too.

KtorZ (approved):

KtorZ:
bors try

Issue #1799, opened 2020-06-25
Cannot join stake-pool on shelley_testnet


piotr-iohk:
# Context

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | 55ffb6b4abcbc7f7ab15c60d2bc4b1c61208cfff `anviking/fix-pools-for-testnet`|
| Platform     | Linux        |
| Installation | From source           |

# Steps to Reproduce

https://hydra.iohk.io/job/Cardano/iohk-nix/cardano-deployment/latest/download/1/index.html

1. 
```
cardano-node run \
        --config /root/node/$type/*-config.json \
        --database-path /root/node/$type/db \
        --socket-path /root/node/$type/node.socket \
        --topology /root/node/$type/*-topology.json \
        --host-addr 127.0.0.1 \
        --port 3001}
```
```
cardano-wallet-shelley serve --port 8090 \
                --node-socket /root/node/$type/node.socket \
                --testnet /root/node/$type/*-genesis.json \
                --database /root/wallet/$type/db

```

2. Join stake-pool on shelley testnet with wallet having funds: (`fuel gentle purse blood bomb advice hundred vehicle maze special pull category smart control shallow`)

```
PUT http://localhost:8090/v2/stake-pools/1fd1450ad92fc024235e2bf908a9b178a6e13f2e653ff0dec5414c46/wallets/98ae2ce26143a126a19a2d47e60b539a596b4d2b
Body: {"passphrase":"Secure Passphrase"}
```
## Expected behavior

I can join.

## Actual behavior
 
```
Code = 500,
Message = {"code":"created_invalid_transaction","message":"That's embarrassing. It looks like I've created an invalid transaction that could not be parsed by the node. Here's an error message that may help with debugging: ApplyTxError [LedgerFailure (UtxowFailure (UtxoFailure (ValueNotConservedUTxO (Coin 994831859) (Coin 995231859))))]"}
```

---

# Resolution 

- Take stake key deposit into account when delegating and quitting delegation ==> #1806 

---

# QA 

- I've changed the genesis configuration to have a non-null key deposit (https://github.com/input-output-hk/cardano-wallet/blob/78074621d87f5a50c833693a38fc4e6aa1a885e6/lib/shelley/test/data/cardano-node-shelley/genesis.yaml#L34) which, as a result, caused most STAKE_POOLS integration scenarios to fail with a similar error to the above :point_up: 

[...]

PR #1806, opened 2020-06-26
Handle key deposit for Shelley transactions

KtorZ:
# Issue Number

#1799 

# Overview

- 8085f82b7352cf2912f41812af422e543ef0a9c9
  :round_pushpin: **add a non-null key deposit to genesis, and account for it for launching pools**
  
- 8240a15c825fab379b5fe3baaee259b3803bddc6
  :round_pushpin: **account for key deposit when registering a stake key**
  This is.. tricky, because the deposit is sort of implicit but, because of the dual support with Jörmungandr, we have to treat it as a fee. So, there's a bit of weird logic / work-around when creating the actual transaction.

- 285d0275c0459bb3012bdc213f700ddc7484a7b5
  :round_pushpin: **fix fee calculation for Shelley**
  - using 28-byte hash length for key hashes
- adding certificate witnesses to the estimation

- 0495b965b48150b4ebaf9941063cda742e71be5a
  :round_pushpin: **give deposit back when de-registing a stake key**
  
- a19bc0e3592ddd4b5bd9b69f154d0c03e9ce7c8d
  :round_pushpin: **adjust test scenarios according to recent API changes**

# Comments

Not 100% ideal, but works without _too much_ added complexity.

### DIFF
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
@@ -165,18 +165,28 @@ newTransactionLayer _proxy _protocolMagic epochLength = TransactionLayer
 
     _mkDelegationJoinTx
         :: FeePolicy
+            -- Latest fee policy
         -> WalletDelegation
+            -- Wallet current delegation status
         -> PoolId
+            -- Pool Id to which we're planning to delegate
         -> (k 'AddressK XPrv, Passphrase "encryption")
+            -- Reward account
         -> (Address -> Maybe (k 'AddressK XPrv, Passphrase "encryption"))
+            -- Key store
         -> SlotId
+            -- Tip of the chain, for TTL
         -> [(TxIn, TxOut)]
+            --  Resolved inputs
         -> [TxOut]
+            -- Outputs
+        -> [TxOut]
+            -- Change, with assigned address
         -> Either ErrMkTx (Tx, SealedTx)
-    _mkDelegationJoinTx policy wDeleg poolId (accXPrv, pwd') keyFrom slot ownedIns outs = do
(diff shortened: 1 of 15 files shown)


lib/shelley/test/data/cardano-node-shelley/genesis.yaml
```diff
@@ -31,7 +31,7 @@ protocolParams:
   extraEntropy:
     tag: NeutralNonce
   maxBlockHeaderSize: 217569
-  keyDeposit: 0
+  keyDeposit: 100
```
KtorZ, line 34: This little one made pretty much all integration tests to fail, as they should.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -280,51 +356,74 @@ _decodeSignedTx bytes = do
...
+                Cardano.shelleyRegisterStakingAddress dummyKeyHash
+            KeyDeRegistrationCertificate ->
+                Cardano.shelleyDeregisterStakingAddress dummyKeyHash
+
+    (addrWits, certWits) =
+        ( Set.map dummyWitness $ Set.fromList (fst <$> inps)
+        , if null certs
+            then Set.empty
+            else Set.singleton (dummyWitness dummyTxIn)
+        )
```
KtorZ, line 409: Had to revise the size estimation too because it was wrong :/ ... 

a) The hash length were still 32 bytes 
b) Certificate witnesses weren't counted

Turned out that, until I introduced the deposit amount, b) and a) were both balancing each other (we were allocating too much fee because of the wrong hash sizes, and it was coping with the lack of witness for the certificate :shrug: )

KtorZ:
bors try

KtorZ:
bors try

KtorZ:
bors retry

Issue #1796, opened 2020-06-24
Conflicting pool production


Anviking:
# Context
 
 

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      |  |
| Platform     |           |
| Installation |             |

# Steps to Reproduce

Haven't managed to reproduce yet. Think I saw it multiple times when tweaking genesis params when running integration tests.

## Actual behavior

Sometimes: 

<img width="764" alt="ska__rmavbild_2020-06-22_kl _14 31 29" src="https://user-images.githubusercontent.com/304423/85590221-83978c80-b644-11ea-86a2-6d76b1f59c51.png">
 

---

# Resolution 

- After investigation (see below), we've identified the root cause of the issue and changed the approach in the chain follower to cope with the issue. See #1808

---

# QA 

- Testing this thoroughly is quite delicate as it would require "manipulating" the chain in such way that this particular edge-case would be met. This however touches the very core of the chain following function, which is used by all three byron, jormungandr and shelley versions -- and integrations tests still pass. 

- I wasn't able to reproduce or re-observe on the shelley testnet after the fix. It is probably best to keep this ticket open a little while until we feel confident about this issue being gone,

Anviking:
I see a lot of these errors on #1783 where activeSlotCoeff is increased, and d decreased.
```
[cardano-wallet.pools-engine:Info:235] [2020-06-26 14:51:21.54 UTC] Applying blocks [3.92 ... 3.92]
[cardano-wallet.pools-engine:Info:235] [2020-06-26 14:51:21.73 UTC] Applying blocks [3.93 ... 3.93]
[cardano-wallet.pools-engine:Info:252] [2020-06-26 14:51:21.76 UTC] Taking a little break from fetching metadata, back to it in about 0.4s
[cardano-wallet.pools-engine:Info:235] [2020-06-26 14:51:21.80 UTC] Applying blocks [3.93 ... 3.93]
[cardano-wallet.pools-engine:Error:235] [2020-06-26 14:51:21.80 UTC] Couldn't store production for given block before it conflicts with another block. Conflicting block header is: 57c080cb<-[7fe787a4-3.93#191]
```
Despite seeing many errors, I see no rollbacks, which is interesting.

Anviking:
When adding a more low-level `ChainSyncLog`:
```
[...]

PR #1807, opened 2020-06-26
keep track of the last applied checkpoint in the chain follower

KtorZ:
# Issue Number

#1796 

# Overview

- 6a1f20ec31c9659c1f9e385e61e1951aef3a8101
  :round_pushpin: **keep track of the last applied checkpoint in the chain follower**
  So that we avoid discarding rollbacks when they occur beyond the initial step.

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/Network.hs
+++ b/lib/core/src/Cardano/Wallet/Network.hs
@@ -327,8 +327,8 @@ follow
     -> (block -> BlockHeader)
     -- ^ Getter on the abstract 'block' type
     -> IO FollowExit
-follow nl tr cps yield header =
-    sleep 0 =<< initCursor nl cps
+follow nl tr cps0 yield header =
+    sleep 0 cps0 =<< initCursor nl cps0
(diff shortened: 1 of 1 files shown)


lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -361,27 +361,27 @@ follow nl tr cps yield header =
           where
             eT = T.pack (show e)
 
-    step :: Int -> Cursor target -> IO FollowExit
-    step delay cursor = runExceptT (nextBlocks nl cursor) >>= \case
+    step :: Int -> [BlockHeader] -> Cursor target -> IO FollowExit
```
Anviking, line 364: I think `[BlockHeader]` is a very confusing type here. Can we not have `BlockHeader` and call it say `followerTip`?

Anviking: Gets a bit weird with the `last cps0` not necessarily being apart of the chain though

PR #1808, opened 2020-06-26
Alternate fix for Shelley pool worker rollback problem

Anviking:
# Issue Number

# Overview

- [x] I believe ignoring the first rollback is a simpler, but still accurate approach.
- [x] A comment explaining why our options are so limited here, and what a better approach could be

# Comments

- ⚠️ This is a more drastic change. Not entirely sure it's sound.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Network.hs
+++ b/lib/core/src/Cardano/Wallet/Network.hs
@@ -327,8 +327,8 @@ follow
     -> (block -> BlockHeader)
     -- ^ Getter on the abstract 'block' type
     -> IO FollowExit
-follow nl tr cps0 yield header =
-    sleep 0 cps0 =<< initCursor nl cps0
+follow nl tr cps yield header =
+    sleep 0 False =<< initCursor nl cps
   where
     delay0 :: Int
     delay0 = 500*1000 -- 500ms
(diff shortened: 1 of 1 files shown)


lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -361,57 +361,65 @@ follow nl tr cps0 yield header =
...
+            -- [] to `SlotId 0 0`. The problem is that the wallet associates
+            -- `SlotId 0 0` with both the first slot and the genesis/origin
+            -- point. (NOTE: I do believe this is correct, but not 100% sure)
+            --
+            -- Some alternative solutions would be to:
+            -- 1. Make sure we have a @BlockHeader@/@SlotId@ for @Origin@
+            -- 2. Stop forcing @follow@ to quit on rollback
+            if hasRolledForward
+            then destroyCursor nl cursor' $> FollowRollback sl
+            else step delay0 cps cursor'
```
KtorZ: I like the approach, but the part checking for 

```hs
                (sl, _:_) | sl == slotId (last cps) ->
            --
                    step delay0 cps cursor'
```

is actually necessary. Imagine the following scenario:

- We have a set of checkpoints `[a,b,c]` initially provided to find an intersection. 
- The node find an intersection with `b`, and ask to rollback to `b`. 

At this point, if we don't acknowledge this rollback, on the next attempt to roll forward, the worker will crash because the block yielded won't be a valid continuation. The worker will restart itself, and provides `[a,b,c]` as a list of checkpoints... and loop ad infinitum.

Anviking: > is actually necessary.

🤦‍♂️damn! Yeah, you're right.

KtorZ (approved):

PR #1826, opened 2020-06-29
Add more debug-level chain-sync- and follow traces

Anviking:
# Issue Number

Related to / was used to debug #1794 

# Overview

The chain-following code is complex. We usually log when the `follow` function asks us to roll forward or backward, but we don't have traces for the underlying logic.

- [x] Trace when `follow` ignores (hopefully) idempotent rollbacks, and why
- [x] Trace low-level chain-sync commands

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/Network.hs
+++ b/lib/core/src/Cardano/Wallet/Network.hs
@@ -401,15 +401,20 @@ follow nl tr cps yield header =
             -- So this becomes a bit intricate:
 
             case (cursorSlotId nl cursor', cps, hasRolledForward) of
-                (_, [], False) ->
+                (sl, [], False) -> do
                     -- The following started from @Origin@.
                     -- This is the initial rollback.
                     -- We can infer that we are asked to rollback to Origin, and
                     -- we can ignore it.
-                     step delay0 hasRolledForward cursor'
-                (sl, _:_, False) | sl == slotId (last cps) ->
-                     step delay0 hasRolledForward cursor'
-                (sl, _, _) ->
+                    traceWith tr $ MsgWillIgnoreRollback sl "initial rollback, \
+                        \cps=[]"
+                    step delay0 hasRolledForward cursor'
+                (sl, _:_, False) | sl == slotId (last cps) -> do
+                    traceWith tr $ MsgWillIgnoreRollback sl "initial rollback, \
+                        \rollback point equals the last checkpoing"
+                    step delay0 hasRolledForward cursor'
+                (sl, _, _) -> do
+                    traceWith tr $ MsgWillRollback sl
                     destroyCursor nl cursor' $> FollowRollback sl
             -- Some alternative solutions would be to:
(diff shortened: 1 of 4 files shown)


lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -401,15 +401,20 @@ follow nl tr cps yield header =
...
-                     step delay0 hasRolledForward cursor'
-                (sl, _:_, False) | sl == slotId (last cps) ->
-                     step delay0 hasRolledForward cursor'
-                (sl, _, _) ->
+                    traceWith tr $ MsgWillIgnoreRollback sl "initial rollback, \
+                        \cps=[]"
+                    step delay0 hasRolledForward cursor'
+                (sl, _:_, False) | sl == slotId (last cps) -> do
+                    traceWith tr $ MsgWillIgnoreRollback sl "initial rollback, \
+                        \rollback point equals the last checkpoing"
```
KtorZ, line 414: checkpoint*

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -348,6 +361,7 @@ chainSyncWithBlocks fromTip queue responseBuffer =
         -> P.ClientStNext n block (Tip block) m Void
     collectResponses respond blocks Zero = P.ClientStNext
         { P.recvMsgRollForward = \block tip -> do
+            traceWith tr $ MsgChainRollForward block
```
KtorZ, line 366: Logging every `rollForward` sounds like something that would cause great slowness. Have you measured the syncing time against mainnet and a non-null tracer on the Byron wallets?

Anviking: These logs should't be visible per default since they are marked `Debug` — but perhaps still worth a check. Haven't tried yet…

KtorZ: We had the case in the past, where simply having a logger would cause tremendous performance regression because the system still needs, on each block, to assess whether or not it should be logged.

Anviking:
[This pr](https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/563#4025dc4b-9a02-49c9-8bc1-3608655b5efc)
```
restore mainnet seq: 387.6 s
--
  | restore mainnet 1% ownership: 3073 s
  | restore mainnet 2% ownership: 6333 s

```

[Recent build](https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/560#9f41bfae-f74c-418f-9c43-0ac97a071bf8) from master:
```
restore mainnet seq: 385.9 s
--
  | restore mainnet 1% ownership: 3027 s
  | restore mainnet 2% ownership: 6241 s

```

Looks like no noticeable change. @KtorZ, thoughts?

KtorZ:
Looks convincing enough :)

KtorZ (approved):

Issue #1822, opened 2020-06-29
Introduce K parameter


paweljakubas:
# Context

As a wallet user
I want to receive the K(optimal pool number) value
so that I can know the maximum amount of ADA I can delegate in one delegation certificate

# Decision

we want to have K parameter available

# Acceptance Criteria

1. Extend protocol parameters and propagate the change throughout codebase
2. Update sqlite tables
3. Add update parameter tests

---

# Development 

# QA 

Desired number of stake pools was added to ApiNetworkParameters. For byron it is set to 0, for shelley it get PParams from ledger, for jormungandr it takes maximum value of pool capping. I checked in integration tests the value for shelley.

piotr-iohk:
lgtm.

I have added tests for Byron and Jormungandr -> https://github.com/input-output-hk/cardano-wallet/pull/1848.

For Shelley I was also thinking about testing the parmeter (in integration test) by comparing the value with the nOpt parameter taken from `cardano-cli shelley query protocol-parameters --testnet-magic 42 --out-file protocol.json` with the desiredPoolNumber reported by the endpoint. After discussion with @paweljakubas I think it might be a bit of overkill.

Anyway looking on shelley testnet the parameter reported by the endpoint matches with the nOpt param reported by cardano-cli.

PR #1827, opened 2020-06-29
Introduce k parameter

paweljakubas:
# Issue Number

#1822 

# Overview

- [x] I have extended ApiNetworkParameters for desired number of pools. This is optional as it does not make sense for byron, it is unknown in jormungandr 
- [x] I have made the changes in swagger
- [x] I have updated unit tests in core, especially made a change to Arbitrary for ApiNetworkParameters to allow Word16. 
- [x] I have added 0 to byron and used max value of participation capping in jormungandr
- [x] I have added impl in shelley using ledger accessors
- [x] I have updated all other places in the code to account for this change 
- [x] I have extended integration network test for shelley and see that desired number of pools is 3.

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
@@ -151,7 +151,7 @@ ProtocolParameters
     protocolParametersFeePolicy             W.FeePolicy     sql=fee_policy
     protocolParametersTxMaxSize             Word16          sql=tx_max_size
     protocolParametersDecentralizationLevel Percentage      sql=decentralization_level
-    protocolParametersDesiredNumberOfPools  Word16 Maybe    sql=desired_pool_number
+    protocolParametersDesiredNumberOfPools  Word16          sql=desired_pool_number
     Primary protocolParametersWalletId
     Foreign Wallet fk_wallet_protocol_parameters protocolParametersWalletId ! ON DELETE CASCADE
     deriving Show Generic
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
@@ -147,11 +147,11 @@ Checkpoint
     deriving Show Generic
 
 ProtocolParameters
-    protocolParametersWalletId              W.WalletId  sql=wallet_id
-    protocolParametersFeePolicy             W.FeePolicy sql=fee_policy
-    protocolParametersTxMaxSize             Word16      sql=tx_max_size
-    protocolParametersDecentralizationLevel Percentage  sql=decentralization_level
-    protocolParametersDesiredNumberOfPools  Word16      sql=desired_pool_number
+    protocolParametersWalletId              W.WalletId      sql=wallet_id
+    protocolParametersFeePolicy             W.FeePolicy     sql=fee_policy
+    protocolParametersTxMaxSize             Word16          sql=tx_max_size
+    protocolParametersDecentralizationLevel Percentage      sql=decentralization_level
+    protocolParametersDesiredNumberOfPools  Word16 Maybe    sql=desired_pool_number
     Primary protocolParametersWalletId
     Foreign Wallet fk_wallet_protocol_parameters protocolParametersWalletId ! ON DELETE CASCADE
     deriving Show Generic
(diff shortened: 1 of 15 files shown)


paweljakubas:
bors try

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
```diff
@@ -147,10 +147,11 @@ Checkpoint
...
 ProtocolParameters
-    protocolParametersWalletId              W.WalletId  sql=wallet_id
-    protocolParametersFeePolicy             W.FeePolicy sql=fee_policy
-    protocolParametersTxMaxSize             Word16      sql=tx_max_size
-    protocolParametersDecentralizationLevel Percentage  sql=decentralization_level
+    protocolParametersWalletId              W.WalletId      sql=wallet_id
+    protocolParametersFeePolicy             W.FeePolicy     sql=fee_policy
+    protocolParametersTxMaxSize             Word16          sql=tx_max_size
+    protocolParametersDecentralizationLevel Percentage      sql=decentralization_level
+    protocolParametersDesiredNumberOfPools  Word16 Maybe    sql=desired_pool_number
```
KtorZ: Why is this a `Maybe` :thinking:  ?

paweljakubas: I looked at PParams in ledger for byron - there is no such a thing. So it does not make sense to talk about it (of course there is no staking for byron). So I thought Nothing there make sense. For jormungandr there is no such thing exposed so also set Nothing (although it is sensible as stake pools are there). For shelley we have this parameter in ledger `nOpt` and we can query the desired number of stake pool (which is 100 btw)

KtorZ: There's `pool_participation_capping` for Jörmungandr, which has the same meaning. For Byron, we could default to 0 here, I think it'll be less misleading than having to carry a Maybe around for Shelley :/

paweljakubas: ok, will use Word16 instead of Maybe Word16 with 0 for Byron

paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/1827/commits/3957dbcaf320cc884c13d43aef0294c7b4b8123e (also used pool participation capping max value for jormunagndr)

paweljakubas:
bors try

paweljakubas:
error not related to this PR to my best knowledge: 
```
Left (DecodeFailure "Error in $: parsing Cardano.Wallet.Api.Types.ApiTransaction(ApiTransaction) failed, key \"id\" not found: Response {responseStatus = Status {statusCode = 404, statusMessage = \"Not Found\"}, responseVersion = HTTP/1.1, responseHeaders = [(\"Transfer-Encoding\",\"chunked\"),(\"Date\",\"Tue, 30 Jun 2020 12:09:51 GMT\"),(\"Server\",\"Warp/3.3.5\")], responseBody = \"{\\\"code\\\":\\\"no_such_pool\\\",\\\"message\\\":\\\"I couldn't find any stake pool with the given id: 1b3dc19c6ab89eaffc8501f375bb03c11bf8ed5d183736b1d80413d6\\\"}\", responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}")
--
```

rvl:
@paweljakubas yes - that test failure is due to #1831.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -1409,6 +1413,7 @@ instance Buildable ProtocolParameters where
     build pp = blockListF' "" id
         [ "Decentralization level: " <> build (pp ^. #decentralizationLevel)
         , "Transaction parameters: " <> build (pp ^. #txParameters)
+        , "K parameter: " <> build (pp ^. #desiredNumberOfStakePools)
```
KtorZ: I'd actually go for "desired number of pools" here too. A bit clearer since there are at least 3 things called `k` in the shelley specs ^^"

KtorZ (approved):

lib/core/src/Cardano/Wallet/Primitive/Types.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/1827/commits/499d4eb76bf2bac9b957940c62e357ffbb4a27d1

paweljakubas:
bors retry

Issue #1825, opened 2020-06-29
Estimate fee Api endpoint response taking ~10 seconds


nikolaglumac:
#1569 
Context
 
There is a regression in the performance of the "Estimate fee" Api endpoint (https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/postTransactionFee) for which the response time has been increased from ~100 milliseconds to ~10 seconds. This happens regardless of the wallet's UTXO fragmentation (it is the same for wallets with a single UTXO as for the ones with >100 UTXOs) and is reproducible every time.
Please note that there is no impact on the "Delegation fee estimation" Api endpoint (https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/getDelegationFee) which has almost instant response...

Logs: [logs.zip](https://github.com/input-output-hk/cardano-wallet/files/4847139/logs.zip)

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | f6d5a80173fd93cc6a08baf3ba841fd9be95207f (master branch) |
| Platform     | All          |
| Installation | Daedalus, develop build            |

# Steps to Reproduce

1. Restore a wallet with funds (or send funds to an empty wallet)
2. Proceed to "Send" screen in Daedalus
3. Enter valid receiving address (it can be an address from the same wallet)
4. Enter valid ada amount
5. Fee estimation endpoint request is triggered upon entering the amount
6. Fees will be visible only after ~10 seconds.

## Expected behavior

Fees should be visible almost instantly.

## Actual behavior
 
Fees are visible only after ~10 seconds.

---

# Resolution 

1. Make shelley transaction size calculation faster ⇒ #1834.
2. Set up shelley versions of API response time benchmarks, to detect performance issues earlier ⇒ #1829 #1845.

---

# QA 

1. Existing tests check that results are the same as before.
2. A new benchmark for the shelley API runs in the nightly tests.

rvl:
I did some more testing - even estimating transaction fees on a completely empty wallet is slow! (it takes 10 seconds to say there is not enough ada)

* Interestingly, on the cardano-wallet-byron backend, fee estimation is fast. On both byron and shelley, the estimation is run 100 times and statistics gathered.

   We only have a nightly benchmark running for byron, not shelley. These byron benchmarks are showing reasonable fee estimation performance for wallets with up to 1000 UTxO.

* Delegation fee estimation is not slow, because this does not require coin selection. That is the part that is running slowly for shelley.

[...]

PR #1829, opened 2020-06-30
Begin porting latency benchmark to shelley

rvl:
Relates to #1825

# Overview

These refactors are the start of porting the latency bench from the byron backend to the shelley backend.

The measurement and reporting code is split into a separate module.

The bench scenario functions are reworked so that swapping backend will be easier.


piotr-iohk (approved):
lgtm

PR #1841, opened 2020-07-01
Bump cardano-ledger-specs to include a fix on non-myopic member reward ordering

KtorZ:
# Issue Number

# Overview

- 04b8141420da8f7379535dcd261b1f77c6ebed97
  :round_pushpin: **Bump cardano-node to include a fix on non-myopic member reward ordering**
  
- d064a2b6e15a9866158b778112601254392e1e68
  :round_pushpin: **update compatibility matrix**
  
- 67865c9ed6f5ba068cca1bea5ae17d38c0b3a8a8
  :round_pushpin: **Also bump cardano-node in niv**
  Matches revision in snapshot.yaml.

I can fix it later so that the cardano-node rev is defined in only one
place, in the stack snapshot.

- c5719173e4795be0e10d3f0c6452ff4db9e0def7
  :round_pushpin: **Unfortunately stack-to-nix does not support recursive snapshot URLs**
  
- 02f61662ac90fd213a00adb68e529bc48bc8fdd4
  

# Comments

As requested by Darko / Daedalus.


Anviking:
I thought it was cardano-node that needed to have the fix, not us?

The calculation happens on the "server-side" i.e. the node, so bumping the specs for the client (us) would be meaningless?

KtorZ:
You're probably right. It'll likely mess up with the whole build actually... because there would be two different versions in cardano-node and cardano-wallet.

Anviking:
But we could probably fork the node from 1.14.2, fork cardano-ledger-specs from here https://github.com/input-output-hk/cardano-node/blob/1.14.2/stack.yaml#L62, cherry-pick in the fix, and target the resulting cardano-node from the wallet

Anviking (approved):

paweljakubas:
@KtorZ what about also to update matrix in README ?

KtorZ:
> Unfortunately stack-to-nix does not support recursive snapshot URLs

:cry: :disappointed:

KtorZ:
bors retry

KtorZ:
bors try

Issue #1835, opened 2020-06-30
Cannot join pool with ITN rewards wallet on shelley testnet


piotr-iohk:
# Context
 
Cannot join pool from ITN reward wallet.
_(Known but haven't seen that reported, so reporting.)_

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | node 1.14.2, wallet `bcc0ff009b4695589c107fdf21272a470b2362c0`|
| Platform     | All         |
| Installation | From source            |

# Steps to Reproduce

1. Restore ITN rewards wallet on Shelley Testnet. => wallet restored. Rewards account has funds.
2. Get some ADA from faucet onto that wallet.
3. Try to join any stake-pool.

## Expected behavior

Should be able to join.

## Actual behavior
 
```
Code = 500,
Message = {"code":"created_invalid_transaction","message":"That's embarrassing. It looks like I've created an invalid transaction that could not be parsed by the node. 
Here's an error message that may help with debugging: 
ApplyTxError [LedgerFailure (DelegsFailure (DelplFailure (DelegFailure (StakeKeyAlreadyRegisteredDELEG (KeyHashObj (KeyHash 96b3e636fddcfb732335f938209812949b5ce355ac80dffb8d4051b7))))))]"}
```

---

# Resolution 

## A plan:
1. Make it easy to create shelley transactions that only register the key, without delegating
2. Create a failing integration test where we A. register key, B. try to delegate
3. Keep track of key registration and key de-registration certificates in the DB

---

# QA 

- Our integration test cluster are now pre-registering a stake key before starting, so that we can test restoring a wallet with a pre-existing stake key and delegating from it (to simulate the case of an ITN) wallet.
    - see here: https://github.com/input-output-hk/cardano-wallet/blob/5e173ea7b1faa883eb41c472e102155c557be4f4/lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs#L407-L412
    - and here for the actual scenario: https://github.com/input-output-hk/cardano-wallet/blob/5e173ea7b1faa883eb41c472e102155c557be4f4/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs#L374-L395

Anviking:
Summarizing info from Jared:
A pool retiring "in" epoch 5 will:
- Disappear from the stake distribution query at the 6/7 epoch boundary
- Be un-joinable at the 5/6 epoch boundary

rvl:
[...]

PR #1842, opened 2020-07-01
Add pending test for joining pool with pre-existing stake key

Anviking:
# Issue Number

#1835 and ADP-358 

# Overview

- [x] Begun abstracting the notion of creating a tx with an arbitrary payload
- [x] At the start of the integration tests, register the stake key of a special wallet
- [x] Test that this wallet can delegate to a pool (pending the actual fix)

# Comments

Failure from `STAKE_POOLS_JOIN_05 - Can join when stake key already exists`:

```
       from the following response: Left (DecodeFailure "{\"code\":\"created_invalid_transaction\",\"message\":\"That's embarrassing. It looks like I've created an invalid transaction that could not be parsed by the node. Here's an error message that may help with debugging: ApplyTxError [LedgerFailure (DelegsFailure (DelplFailure (DelegFailure (StakeKeyAlreadyRegisteredDELEG (KeyHashObj (KeyHash 342dac9f95ee4f92567f652a736d64bfa0afaa2da795dda8de658580))))))]\"}")

```

### DIFF
--- a/lib/shelley/test/integration/Main.hs
+++ b/lib/shelley/test/integration/Main.hs
@@ -13,6 +13,8 @@ module Main where
 
 import Prelude
 
+import Cardano.Address.Derivation
+    ( XPrv, xprvFromBytes, xpubFromBytes )
 import Cardano.BM.Data.Severity
     ( Severity (..) )
 import Cardano.BM.Data.Tracer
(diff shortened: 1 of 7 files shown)


lib/shelley/test/integration/Main.hs
```diff
@@ -324,3 +340,40 @@ minSeverityFromEnv def var = lookupEnv var >>= \case
...
     Just "" -> pure def
     Just arg -> either die pure (parseLoggingSeverity arg)
+
+-- | Pre-register a staking key for the STAKE_POOLS_JOIN_05 test.
+preregisterStakingKeysForTests :: NetworkLayer IO t b -> IO ()
+preregisterStakingKeysForTests nl = do
+    let payload = TxPayload
+            [toStakeKeyRegCert rewardPub]
+            (const mempty)
+            (Fee 9999999900)
```
Anviking, line 350: 🤔 Not having to hard-code this would be nice

Anviking:
bors try

Anviking:
Including the test in #1851 together with the fix

Issue #1821, opened 2020-06-29
Tentatively allow having an initial reserve amount for running coin selection.


KtorZ:
# Context

In the context of Shelley, Coin Selection becomes a bit more complex. There are now "things" that can be on the left side of the balance when making transaction that aren't unspent transaction inputs. In particular, we can now find:

- deposit sent back to wallet
- rewards from a stake address 

# Decision

These special inputs will be treated separately from standard UTxO inputs, but fully part of the fee balancing algorithm to allow interacting with them when building transactions. Both rewards and deposit are a bit special in the sense that they require at least a UTxO in a transaction. 

As a first step, I'll create an extension of the current `CoinSelection` data-type as an `UnsignedTransaction`, allowing it to carry an extra reserve amount that can (and will) be used in priority when paying for fees. 

# Acceptance Criteria

1. Fee balancing can be run with an extra positive "reserve" amount that allows for providing a balance to consume. 
2. The behavior of the fee balancing algorithm remains unchanged when `reserve = 0`

---

# Development 

- Added reserve as two separate fields `reclaim` and `withdrawal`, as well as  a new `deposit` for implicit deposit output. Also took care of all the internal plumbing to handle these correctly, including reviewing the fee balancing algorithm and adding the correct witnesses for withdrawal! ==> #1843 

# QA 

- All existing integration & unit tests pass, except one on Jörmungandr's I had to temporarily disable. I did some rework in fee estimator from the integration tests but we've lost some precision along the way which makes it quite harder to run this scenario again. The main problem being that we use two differents ways of calculating fees in integration scenarios and in the real code! ==> https://github.com/input-output-hk/cardano-wallet/blob/master/lib/jormungandr/test/integration/Test/Integration/Jormungandr/Scenario/API/StakePools.hs#L643-L657

- I've extended the unit test for the fee balancing function, with extra assertions. I've also extended the generators so they would cover more cases and in particular cases which mimic the way we "abuse" the algorithm for the delegation selection. In particular we show that:
  1. If at the end of the balancing, remaining fee are null, then the selection is indeed balanced.
  2. If the selection is balanced, then inputs are necessarily not null (we need at least 1 UTxO input)
  3. If the selection is balanced, then it is _valid_, where valid means that the sum of its inputs (including reclaim and withdrawal) does not exceed the sum of its output (including deposit).
[...]

PR #1843, opened 2020-07-01
Allow coin-selection to carry an extra "reserve input"

KtorZ:
# Issue Number

#1821 

# Overview

- 859f53bdeed9fc641f83a98201e184254eb14591
  :round_pushpin: **extend 'CoinSelection' type to carry an extra reserve amount that can be used as input**
  For now, the reserve is always 'Nothing', which should not alter any of the existing behavior. Then, it can be
set to a specific value (e.g. the reward balance) to be used on the left side of the balance (i.e. the input side).

- eebc981a6e80d9403f0df5c9dc5a9917cf5289e0
  :round_pushpin: **rename 'rebalanceChangeOutput' in 'rebalanceSelection' and allow picking from the reserve**
  The approach is still very 'simple', before trying to deplete any change output, we first try to remove fees
from the reserve. Once the reserve is empty, we'll start depleting change outputs.

- 08a6cbe852793398b642d70faeae834d64620406
  :round_pushpin: **move prop_rebalanceSelection from byron package to core package**
  It was originally in `-byron` because it was using the fee policy and
fee estimation from Byron, so, while moving it, I also changed the way
the transaction size is estimated to make it mimics the way it's done on
shelley and byron, with rather realistic fee values.

- 338dde73e990924e3004d4e3a11825deb0382a2e
  :round_pushpin: **generate arbitrary selection with reserve, add more classification and assertions to properties**
  
- 20c161bd8cb528d1c6a16cfada568dcebc5dd137
  :round_pushpin: **do not mutate the reserve as part of the fee balancing process**
  Everything in the past coin selection (without reserve) was easy to reason about. One could always look at the sum
of inputs and compare it with the sum of outputs and changes, and there were some nice properties about this. Having
mutable reserve makes it somewhat very confusing because, there's no trace of where the money comes from (turning
reserve into changes shouldn't make it disappear).

- 22e17144e678bd0292acc3891c186af2fa0d4cc4
  :round_pushpin: **expand 'reserve' to also cover deposit and reclaim**
  So that types are clear from the coin selection and a bit less confusing.

- 43cace27cff105b6d3a8428bbec2998ada0398a0
  :round_pushpin: **extend transaction layer with an 'initDelegationSelection'**
    This in order to defer to the relevant backend the necessary bits of logic regarding withdrawal, deposits and reclaims.
  I've also removed the 'WalletDelegation' from 'mkDelegationJoinTx' as this can now be inferred from the presence of a deposit in the coin selection.

  We have the following situation:

  object     | cardano-node | jormungandr
   ----      | ----         | ----
  inputs     | explicit     | explicit
  reclaim    | implicit     | N/A
  withdrawal | explicit     | N/A
  ---        | ---          | ---
  outputs    | explicit     | explicit
  deposit    | implicit     | N/A

  Regardless of the situation, we must always have at least one input (replay protection) and output (although outputs may be entirely change outputs).
  Then, depending on the situation, we may have 'withdrawal', 'reclaim' and or 'deposit' present.

  `reclaim` and `deposit` are "opposed" and can't be present together. `withdrawal` can in theory be present with both, but in practice, will only be present alone: we'll only
  allow users to withdraw their reward as part of standard transaction. `deposit` are only present when registering a stake-key for the first time. In summary, this gives us:

  type of transaction  | objects
  -------------------  | --------
  standard transaction | inputs, outputs, withdrawal? (if users said so)
  initial delegation   | inputs, outputs, deposit
  re-delegation        | inputs, outputs
  de-registration      | inputs, outputs, reclaim

- 227dcd9d9287d4ef3d76f602fc4cdd5b5e15d347
  :round_pushpin: **only count reclaim and withdrawal as part of the input balance if there's already a UTxO input**
  
- dd003064e32d82cdb53bfeb164ad3614c8b0fb93
  :round_pushpin: **tweak arbitrary coin selection generator to produce more realistic cases**
  The generator was good for 'payment' kind of selection. Yet, we also sort of abuse the fee balancing when creating selection for delegation with special delegation
with no inputs and no outputs. This was therefore completely untested by the current generator! Now it does, sometimes, create typical selections we would create when
delegating or undelegating. It nicely caught an error in the fee balancing algorithm which this commit also fixes

- 523ca66a8c5c7cc8417836830c4d467c29381ffa
  :round_pushpin: **re-implement shelley's transaction layer accordingly**
  In the end, it is much simpler and cleaner, no more hacky-hack to get things working. The newly introduced 'initDelegationSelection' makes it easier to construct a correct coin selection from the start and simply go with it.

- 7fa1162927b32e4510d93ae2c52753c2eae5ca55
  :round_pushpin: **review integration 'feeEstimator' to use 'DelegationAction'**
  We have two different ways of computing fees, and this is because in Byron, it used to be complicated to evaluate fees for a transaction. Now, this fee estimator feels sort of redundant and we should solely rely on the result from the fee estimation endpoint. In the meantime, I've adjusted it to make a bit more consistent with how fees are calculated elsewhere.

I've also adjusted the genesis file to have a much bigger deposit key.  The previous code was actually handling things in a very wrong way, but it went unnoticed because two errors were cancelling each others: fee were slightly over-evaluated, and the deposit for key was small enough to compensate.

- f8faf55502f57a203be4470954820df5b31fd8b3
  :round_pushpin: **re-implement byron & jormungandr transaction layers accordingly**
  
- ff7ee223cb549538025b75f6cd4cddccd3f3d7a7
  :round_pushpin: **add withdrawal on each payment transaction**
  We could make it slightly smarter here and requires that the withdrawal is only requested when the balance is somewhat sufficient.
Sufficient would mean, bigger than the cost of adding it. Indeed, adding a withdrawal would only make sense if the cost of adding it
isn't greater to its value.

- 98ace73a0f2e8d43faa0fa0859ea9ca2d2316def
  :round_pushpin: **only set withdrawal when they are sufficiently big**
  Otherwise, we end-up paying for more than the value they offer.

- d6b1db1038f2c455433dae5dc0b4dc079bc1b7fd
  :round_pushpin: **remove now obsolete 'ErrChangeIsEmptyForRetirement'**
  
- ba174deb629f70a9f19f6c406384f733543c05ca
  :round_pushpin: **prepare foundation to support withdrawals witnesses in standard transaction**
  I've once again changed a bit the transaction layer to be more aligned with the recent changes (now takes a full coin selection).
Also, it now requires a reward account to be given; this is relatively 'unsound' for Byron and Icarus wallet since they have no
reward account, yet, the account is only used when a non-null withdrawal is provided which should never happen for these wallet types
since they aren't receiving rewards at all.

- e7ab562ade96ec58d7d47aefd61da5c814164121
  :round_pushpin: **add property showing that withdrawals are correctly estimated by Shelley tx layer**
  
- 3035ec7dffe0ca0b63cb1185b73b5af0d1a19464
  :round_pushpin: **add integration test showing that rewards are spent with a transaction**
  
- 10e04f46d1d6efab8d6d5a74757c26aa31fd6c27
  :round_pushpin: **count rewards as part of the total and available balance**

# Comments

- ~~TODO: need to create the a corresponding withdrawal witness inside payment transactions~~
- ~~TODO: need to NOT add a withdrawal when it's below certain threshold~~
- ~~TODO: need to add an integration test showing that reward are withdrawn successfully.~~

### DIFF
--- a/lib/byron/test/unit/Cardano/Wallet/Byron/TransactionSpec.hs
+++ b/lib/byron/test/unit/Cardano/Wallet/Byron/TransactionSpec.hs
@@ -48,12 +48,7 @@ import Cardano.Wallet.Primitive.AddressDerivation.Byron
 import Cardano.Wallet.Primitive.AddressDerivation.Icarus
     ( IcarusKey (..) )
 import Cardano.Wallet.Primitive.CoinSelection
-    ( CoinSelection (..)
-    , CoinSelectionOptions (..)
-    , changeBalance
-    , inputBalance
-    , outputBalance
-    )
+    ( CoinSelection (..), CoinSelectionOptions (..) )
 import Cardano.Wallet.Primitive.CoinSelection.LargestFirst
     ( largestFirst )
 import Cardano.Wallet.Primitive.Fee
@@ -62,7 +57,6 @@ import Cardano.Wallet.Primitive.Fee
     , FeePolicy (..)
     , OnDanglingChange (..)
     , adjustForFee
-    , rebalanceSelection
     )
 import Cardano.Wallet.Primitive.Types
     ( Address (..)
@@ -164,11 +158,6 @@ spec = do
             runExceptT (adjustForFee opts utxo csel) >>= \case
                 Left e -> expectationFailure $ "failed with: " <> show e
                 Right{}-> pure ()
-
-    it "1561 - The fee balancing algorithm converges for any coin selection."
-        $ property
-        $ withMaxSuccess 10000
-        $ forAllBlind (genSelection @'Mainnet @ByronKey) prop_rebalanceSelection
 
     describe "Fee estimation calculation" $ do
         it "Byron / Mainnet" $ property $
@@ -284,38 +273,6 @@ spec = do
 {-------------------------------------------------------------------------------
                                 Properties
 -------------------------------------------------------------------------------}
-
-prop_rebalanceSelection
-    :: CoinSelection
-    -> OnDanglingChange
-    -> Property
-prop_rebalanceSelection sel onDangling = do
-    let (sel', fee') = rebalanceSelection opts sel
-    let prop = case onDangling of
-            PayAndBalance ->
-                fee' /= Fee 0 || Fee (delta sel') == estimateFee opts sel'
-            SaveMoney ->
-                fee' /= Fee 0 || Fee (delta sel') >= estimateFee opts sel'
-    property prop
-        & counterexample (unlines
-            [ "selection (before):", pretty sel
-            , "selection (after):", pretty sel'
-            , "delta (before): " <> show (delta sel)
-            , "delta (after):  " <> show (delta sel')
-            , "remaining fee:  " <> show (getFee fee')
-            ])
-  where
-    delta s = inputBalance s - (outputBalance s + changeBalance s)
-    opts = FeeOptions
-        { estimateFee = minimumFee tlayer feePolicy []
-        , dustThreshold = minBound
-        , onDanglingChange = onDangling
-        }
-      where
-        tlayer =
-            newTransactionLayer @'Mainnet @ByronKey Proxy mainnetMagic
-        feePolicy =
-            LinearFee (Quantity 155381) (Quantity 43) (Quantity 0)
 
 propSizeEstimation
     :: forall n k.
--- a/lib/byron/test/unit/Cardano/Wallet/Byron/TransactionSpec.hs
+++ b/lib/byron/test/unit/Cardano/Wallet/Byron/TransactionSpec.hs
@@ -157,6 +157,8 @@ spec = do
                         ]
                     , change =
                         [Coin 30_556]
+                    , reserve =
+                        Nothing
                     }
 
             runExceptT (adjustForFee opts utxo csel) >>= \case
@@ -365,7 +367,7 @@ propSizeEstimation pm genSel genChngAddrs =
         idPolicy = LinearFee (Quantity 0) (Quantity 1) (Quantity 0)
 
     fromCoinSelection :: CoinSelection -> [Address] -> CBOR.Encoding
-    fromCoinSelection (CoinSelection inps outs chngs) chngAddrs =
+    fromCoinSelection (CoinSelection inps outs chngs _) chngAddrs =
         CBOR.encodeSignedTx (fst <$> inps, outs <> outs') wits
       where
         dummySig =
@@ -444,7 +446,7 @@ genSelection = do
         }
 
 shrinkSelection :: CoinSelection -> [CoinSelection]
-shrinkSelection sel@(CoinSelection inps outs chgs) = case (inps, outs, chgs) of
+shrinkSelection sel@(CoinSelection inps outs chgs rsv) = case (inps, outs, chgs) of
     ([_], [_], []) ->
         []
     _ ->
@@ -457,24 +459,25 @@ shrinkSelection sel@(CoinSelection inps outs chgs) = case (inps, outs, chgs) of
             chgs'' = drop 1 chgs
         in
             filter (\s -> s /= sel && isValidSelection s)
-                [ CoinSelection inps' outs' chgs'
-                , CoinSelection inps' outs chgs
-                , CoinSelection inps outs chgs'
-                , CoinSelection inps outs' chgs
-                , CoinSelection inps'' outs'' chgs''
-                , CoinSelection inps'' outs chgs
-                , CoinSelection inps outs'' chgs
-                , CoinSelection inps outs chgs''
+                [ CoinSelection inps' outs' chgs' rsv
+                , CoinSelection inps' outs chgs rsv
+                , CoinSelection inps outs chgs' rsv
+                , CoinSelection inps outs' chgs rsv
+                , CoinSelection inps'' outs'' chgs'' rsv
+                , CoinSelection inps'' outs chgs rsv
+                , CoinSelection inps outs'' chgs rsv
+                , CoinSelection inps outs chgs'' rsv
                 ]
 
 isValidSelection :: CoinSelection -> Bool
-isValidSelection (CoinSelection i o c) =
+isValidSelection (CoinSelection i o c r) =
     let
         oAmt = sum $ map (fromIntegral . getCoin . coin) o
         cAmt = sum $ map (fromIntegral . getCoin) c
         iAmt = sum $ map (fromIntegral . getCoin . coin . snd) i
+        rAmt = maybe (0 :: Integer) (fromIntegral . getCoin) r
     in
-        (iAmt :: Integer) >= (oAmt + cAmt)
+        iAmt + rAmt >= oAmt + cAmt
 
 genTxIn :: Gen TxIn
 genTxIn = TxIn
--- a/lib/byron/test/unit/Cardano/Wallet/Byron/TransactionSpec.hs
+++ b/lib/byron/test/unit/Cardano/Wallet/Byron/TransactionSpec.hs
@@ -496,8 +496,10 @@ goldenTestSignedTx proxy pm nOuts xprvs expected = it title $ do
     let keyFrom a = (,mempty) <$> Map.lookup a s
     let inps = mkInput <$> zip addrs [0..]
     let outs = take nOuts $ mkOutput <$> cycle addrs
+    let cs = mempty { inputs = inps, outputs = outs }
     let curSlot = error "current slot not needed in byron mkStdTx"
-    let res = mkStdTx (newTransactionLayer proxy pm) keyFrom curSlot inps outs
+    let rewardAcnt = error "reward account not needed in byron mkStdTx"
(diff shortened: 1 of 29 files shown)


lib/byron/test/unit/Cardano/Wallet/Byron/TransactionSpec.hs
```diff
@@ -157,17 +151,14 @@ spec = do
...
                     }
 
             runExceptT (adjustForFee opts utxo csel) >>= \case
                 Left e -> expectationFailure $ "failed with: " <> show e
                 Right{}-> pure ()
 
-    it "1561 - The fee balancing algorithm converges for any coin selection."
-        $ property
-        $ withMaxSuccess 10000
-        $ forAllBlind (genSelection @'Mainnet @ByronKey) prop_rebalanceChangeOutputs
```
KtorZ, line 169: Moved to core package.

lib/byron/test/unit/Cardano/Wallet/Byron/TransactionSpec.hs
```diff
@@ -283,38 +274,6 @@ spec = do
...
-        { estimateFee = minimumFee tlayer feePolicy []
-        , dustThreshold = minBound
-        , onDanglingChange = onDangling
-        }
-      where
-        tlayer =
-            newTransactionLayer @'Mainnet @ByronKey Proxy mainnetMagic
-        feePolicy =
-            LinearFee (Quantity 155381) (Quantity 43) (Quantity 0)
-
```
KtorZ, line 317: Moved to core package.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
```diff
@@ -48,6 +48,8 @@ data CoinSelection = CoinSelection
       -- ^ Picked outputs
     , change  :: [Coin]
       -- ^ Resulting changes
+    , reserve :: Maybe Coin
```
paweljakubas: it can be only "one" ? not list?

lib/core/src/Cardano/Wallet/Primitive/Fee.hs
```diff
@@ -213,16 +213,33 @@ coverRemainingFee (Fee fee) = go [] where
...
     -> CoinSelection
     -> (CoinSelection, Fee)
-rebalanceChangeOutputs opts s
+rebalanceSelection opts s
     -- selection is now balanced, nothing to do.
     | φ_original == δ_original =
         (s, Fee 0)
 
+    -- some fee left to pay, and the reserve is non-empty, use it first.
+    | φ_original > δ_original && reserve s > Just (Coin 0) =
```
paweljakubas: :+1:

lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
KtorZ: See comments above. In theory, we could also have deposits here. So I am thinking about having a list of `Reserve` where `Reserve = Withdrawal | Deposit`

lib/core/src/Cardano/Wallet/Primitive/Fee.hs
```diff
@@ -213,16 +213,33 @@ coverRemainingFee (Fee fee) = go [] where
...
+        let
+            -- Safe because of the above guard.
+            Just (Coin r) = reserve s
+            r' = if r > φ_original
+                then Coin (r - φ_original)
```
paweljakubas: it was supposed to turn all reserve (if there is left) into change. So `rebalanceSelection` will be called in with decreased reserve and some branch will decrease it to zero and after that it will return in `senderPaysFee` with reserve (Just (Coin 0)). Is that right?

lib/core/test/unit/Cardano/Wallet/Primitive/FeeSpec.hs
```diff
@@ -472,6 +491,66 @@ propDivvyFeeInvariantEmptyList (fee, outs) =
...
+            , "remaining fee:  " <> show (getFee fee')
+            ])
+        & classify (reserveNonNull && feeLargerThanDelta)
+            "reserve > 0 && fee > delta"
+        & classify (reserveLargerThanFee && feeLargerThanDelta)
+            "reserve > fee && fee > delta"
+        & classify reserveLargerThanFee
+            "reserve > fee"
+        & classify feeLargerThanDelta
+            "fee > delta"
```
paweljakubas, line 542: any chance to test `ErrValidateSelection` here or maybe at integration testing/backend TransactionSpecs ? :thinking:

lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
paweljakubas: Will Withdraw and Deposit, if introduced will change on this level? Will there be different conditions in `rebalanceSelection` for them? for example Withdraw will deplete and transform into change at first, and only then Deposit?  Or other rules?

KtorZ: No, they'll have the exact same effect hence why I initially went for a simple `Maybe Coin`. Perhaps it's not worth abstracting further here, in the end, both the deposit and withdrawal can be represent as a sum for this reserve. 
The difference lies in how the transaction is then constructed / signed / serialized. Withdrawals require a special entry as .. "withdrawals" in the transaction, as well as a special witness. 

Whereas deposit are implicit, when de-registering a stake key.

lib/core/test/unit/Cardano/Wallet/Primitive/FeeSpec.hs
KtorZ: In the fee spec, no chance, and in integration also quite hard. This is actually a server error (500), so something that we shouldn't be able to trigger under any circumstances, in theory.

lib/core/src/Cardano/Wallet/Primitive/Fee.hs
KtorZ: Correct.

KtorZ: Hence the added check in the corresponding property checking that the reserve is always 0 at the end of the balancing.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
```diff
@@ -44,10 +44,16 @@ import GHC.Generics
 data CoinSelection = CoinSelection
     { inputs  :: [(TxIn, TxOut)]
       -- ^ Picked inputs
+    , withdrawal :: Word64
+      -- ^ An available withdrawal amount, counting as an extra input
+    , reclaim :: Word64
+      -- ^ Claim back a deposit, counting as a an extra input
```
paweljakubas, line 50: :+1:

lib/core/src/Cardano/Wallet/Transaction.hs
```diff
@@ -97,30 +89,33 @@ data TransactionLayer t k = TransactionLayer
...
+            -- A balanced coin selection where all change addresses have been
+            -- assigned.
         -> Either ErrMkTx (Tx, SealedTx)
         -- ^ Construct a transaction containing a certificate for quiting from
         -- a stake pool.
         --
         -- The certificate is the public key of the reward account.
 
-    , minimumFee :: FeePolicy -> [Certificate] -> CoinSelection -> Fee
+    , initDelegationSelection
```
paweljakubas, line 105: :+1:

lib/core/test/unit/Cardano/Wallet/Primitive/FeeSpec.hs
```diff
@@ -567,13 +663,58 @@ genTxOut coins = do
...
+    , (1, do
+        deposit_ <- genDeposit 100000
+        pure $ mempty { deposit = deposit_ }
+      )
+    ]
+
+genWithdrawal :: Gen Word64
+genWithdrawal = frequency
+    [ (3, pure 0)
+    , (1, oneof
```
paweljakubas, line 688: why this oneof?

paweljakubas (approved):
A lot of changes in the code due to two ideas implemented (not all was able to follow, but I bet the author mainly followed compiler here). When reviewing focused on the ideas, ie. 1. extension of CoinSelection to account for deposit, withdrawal and reclaim, 2. changing interface in Cardano.Wallet.Transaction - and here using CoinSelection in number of methods instead of inputs/outputs and introducing `initDelegationSelection` to handle initial selection for different backends, 3. testing rebalancingSelection in unit tests. All three point looks solid to me!

lib/core/test/unit/Cardano/Wallet/Primitive/FeeSpec.hs
KtorZ: To favor some edge-cases a bit more in the generator. The type is a `Word64`, so something like `arbitrary` will span over the entire domain. Yet, we know that bugs are usually found at edges, so I've favored small values, close to the fee amount so see how it behaves when a bit smaller than the fee, or a bit bigger, and still, allowed for big values far from this edge.

Issue #1819, opened 2020-06-29
Extend Shelley `listPools` operation to provide real pool retirement data.


jonathanknowles:
# Context

As part of the work to track pool retirements, we need to ensure that the API server actually reports real pool retirement data as stored in the database (rather than always returning `Nothing`).

# Decision

Extend the Shelley `listPools` operation to provide real pool retirement data:

- [x] Extend `PoolDBMetrics` to include retirement data.
- [x] Extend `readDBPoolData` to provide pool retirement data.
- [x] Adjust `combineDbAndLsqData` to write retirement data into `ApiStakePool`.

# Acceptance Criteria

- [x] The `ListStakePools` endpoint **_must_** return `ApiStakePool` objects with real `retirement` field values and not skeleton values, reflecting the actual contents of the database.

# Development 

- https://github.com/input-output-hk/cardano-wallet/pull/1847

# QA 

Add integration tests to check that:

- [x] retiring a pool eventually results in a correct update to the `retirement` field of `ApiStakePool`, when read through the `ListStakePools` operation.

# Notes

Note that ideally, we would also test that retiring and then re-registering a pool clears the `retirement` field. However, the static nature of our test pool cluster means that building such a test is currently impractical. See: https://jira.iohk.io/browse/ADP-385

piotr-iohk:
ok

Issue #1817, opened 2020-06-29
Extend pool monitoring code to store retirements in the DB.


jonathanknowles:
# Context

As part of the work to track pool retirements, we need to make sure that when the wallet encounters a pool retirement certificate, that the relevant information is recorded in the DB.

# Decision

Extend the pool monitoring code to store pool retirements in the DB. Adjust the `forward` function within `monitorStakePools` to call `putPoolRetirement`, along with the corresponding _slot_ and _slot internal index_.

For a discussion of slot internal indices, see https://github.com/input-output-hk/cardano-wallet/issues/1815.

# Acceptance Criteria

- [x] The pool monitoring code, on encountering a pool retirement certificate, **_must_** ensure that the retirement epoch of that pool is recorded in the DB, along with the slot and slot internal index corresponding to that certificate.

# Development 

- https://github.com/input-output-hk/cardano-wallet/pull/1847

# QA 

Write integration tests to check that:

- [x] retiring a pool eventually results in a correct update to the `retirement` field of `ApiStakePool`, when read through the `ListStakePools` operation. (This field is only written as a consequence of the pool monitoring code writing retirements to the DB.)

piotr-iohk:
lgtm

Issue #1816, opened 2020-06-29
Extend DB layer to allow reading and writing of pool retirement data.


jonathanknowles:
# Context

As part of the work to track pool retirements, we need to provide a way for users of the DB layer to record that a pool will retire at a given epoch, or to query the DB to determine whether such a retirement has already been recorded. 

# Decision

Add the following operations (or equivalent operations) to the DB layer:

```hs
putPoolRetirement
    :: SlotId
    -> SlotInternalIndex
    -> PoolRetirementCertificate
    -> stm ()

readPoolRetirement
    :: PoolId
    -> stm (Maybe (SlotId, SlotInternalIndex, PoolRetirementCertificate))
```

For a discussion of `SlotInternalIndex` see https://github.com/input-output-hk/cardano-wallet/issues/1815.

# Acceptance Criteria

- [x] It **_must_** be possible to use the DB layer to record a pool retirement for a given epoch announced at a given slot and slot internal index.
- [x] It **_must_** be possible to use the DB layer to look up the _latest_ pool retirement certificate for a pool, and if one exists, the slot and slot internal index corresponding to that certificate.

# Development 

- https://github.com/input-output-hk/cardano-wallet/pull/1847

# QA 

See the following property tests:
- [x] `putPoolRetirement`
- [x] `prop_multiple_putPoolRetirement_single_readPoolRetirement`

jonathanknowles:
Hi @piotr-iohk!

I believe the work for this ticket is complete. Would you be okay with me closing it?

piotr-iohk:
lgtm.

Issue #1815, opened 2020-06-29
Extend SQLite DB to store pool retirement data.


jonathanknowles:
# Context

As part of the work to track pool retirements, we need a way of storing **pool retirement data** in the database.

This is complicated by the fact that:

1. A _single block_ can contain _multiple_ certificates relating to the same pool.
2. A _single transaction_ can contain _multiple_ certificates relating to the same pool.
3. The _relative order_ of these certificates is _significant_.
4. In _general_, earlier certificates are made obsolete by later certificates, but in particular:
    - a registration certificate will always _supersede_ an earlier retirement certificate (making that earlier retirement certificate obsolete);
    - a retirement certificate will always _augment_ the latest known registration certificate (meaning that data supplied as part of the registration remains relevant until the pool actually retires).

These rules follow from the "pool inference rule", as described in "[A Formal Specification of the Cardano Ledger](https://hydra.iohk.io/build/3202141/download/1/ledger-spec.pdf)".

# Decision

Extend the pool DB schema to make it possible to store:

- [x] retirement certificates.
- [x] the order of publication of certificates _within the same block_, for both registration and retirement certificates.

# Acceptance Criteria

- [x] The pool database _must_ provide a way to record all retirement certificates for any given pool.
- [x] The pool database _must_ provide a way to record the _order of publication_ of registration and retirement certificates within a block.

# Development 

Add a `slot_internal_index` row to the `PoolRegistration` table (and similarly to the `PoolOwners` table):
```patch
PoolRegistration sql=pool_registration
    poolRegistrationPoolId            W.PoolId            sql=pool_id
    poolRegistrationSlot              W.SlotId            sql=slot
+   poolRegistrationSlotInternalIndex W.SlotInternalIndex sql=slot_internal_index        
```
This will allow the database to record the position of each certificate in relation to other certificates in the same block.

Add a table to store pool retirements:

```hs
PoolRetirement sql=pool_retirement                                                                                                                                                    
    poolRetirementPoolId              W.PoolId            sql=pool_id                                                                                                                 
    poolRetirementSlot                W.SlotId            sql=slot                                                                                                                    
    poolRetirementSlotInternalIndex   W.SlotInternalIndex sql=slot_internal_index                                                                                                     
    poolRetirementEpoch               Word64              sql=epoch
```
For any given pool **_p_**:

[...]

PR #1847, opened 2020-07-03
Track Stakepool Retirements in the DB

jonathanknowles:
# Related Issues

#1815 
#1816 
#1817 
#1819 
#1880 

# Overview

This PR:

- [x] Extends the pool DB schema to make it possible to store:
    - [x] retirement certificates.
    - [x] the _order_ of publication of certificates _within the same block_, for both registration and retirement certificates. This is necessary, as the intra-block order is _significant_.<br><br>

- [x] Adds the `PoolLifeCycleStatus` type, which indicates the current lifecycle stage of a pool.

    There are currently _three_ possibilities:
    1. `NotRegistered`
        (indicates that a pool has never been registered)
    2. `Registered`
        (indicates that a pool has been registered but _not_ retired)
        (provides a registration certificate)
    3. `RegisteredAndRetired`
        (indicates that a pool has been registered _and_ retired)
        (provides both a registration and a retirement certificate)<br><br>

- [x] Adds the `readPoolLifeCycleStatus` function to the pool DB layer, making it possible to determine the current lifecycle status of a pool. 

- [x] Updates the `retirement` field of `ApiStakePool` with live information.

# Property Tests

This PR adds property tests to ensure that:

- [x] the `determinePoolLifeCycleStatus` function respects the correct order of precedence for registration and retirement certificates:
    - for a given pool, a registration certificate always _supercedes_ a prior retirement certificate.
    - for a given pool, a retirement certificate always _augments_ the latest registration certificate.
- [x] `readPoolLifeCycleStatus` queries work correctly in the presence of blocks containing multiple certificates for the same pool.
- [x] database rollback works correctly in the presence of retirement certificates.

# Future Work

A **_future_** PR will add integration tests to ensure that:

- retiring a pool eventually results in a correct update to the `retirement` field of `ApiStakePool`, when read through the `ListStakePools` operation.
- retiring and then re-registering a pool clears the `retirement` field. 

See QA section of https://github.com/input-output-hk/cardano-wallet/issues/1819.


lib/core/src/Cardano/Pool/DB/Model.hs
```diff
@@ -180,22 +192,60 @@ mReadStakeDistribution epoch db@PoolDatabase{distributions} =
     , db
     )
 
-mPutPoolRegistration :: SlotId -> PoolRegistrationCertificate -> ModelPoolOp ()
-mPutPoolRegistration sl registration db@PoolDatabase{owners,registrations} =
+mPutPoolRegistration
+    :: CertificatePublicationTime
+    -> PoolRegistrationCertificate
+    -> ModelPoolOp ()
+mPutPoolRegistration sp registration db =
```
jonathanknowles: Given that the type is `CertificatePublicationTime`, we should probably also rename `sp` to `cpt` for consistency.

```suggestion
mPutPoolRegistration cpt registration db =
```

lib/core/src/Cardano/Pool/DB/Model.hs
```diff
@@ -180,22 +192,60 @@ mReadStakeDistribution epoch db@PoolDatabase{distributions} =
...
+    only k (_, k') _ = k == k'
 
-mReadPoolRegistration :: PoolId -> ModelPoolOp (Maybe PoolRegistrationCertificate)
-mReadPoolRegistration poolId db@PoolDatabase{registrations} =
-    ( Right $ fmap snd $ Map.lookupMax $ Map.filterWithKey (only poolId) registrations
+mPutPoolRetirement
+    :: CertificatePublicationTime
+    -> PoolRetirementCertificate
+    -> ModelPoolOp ()
+mPutPoolRetirement sp retirement db =
```
jonathanknowles: Given that the type is `CertificatePublicationTime`, we should probably also rename sp to cpt for consistency.

```suggestion
mPutPoolRetirement cpt retirement db =
```

lib/core/src/Cardano/Pool/DB/Sqlite/TH.hs
```diff
@@ -81,21 +81,31 @@ PoolOwner sql=pool_owner
     poolOwnerIndex      Word8        sql=pool_owner_index
 
     Primary poolOwnerPoolId poolOwnerSlot poolOwnerOwner poolOwnerIndex
-    Foreign PoolRegistration fk_registration_pool_id poolOwnerPoolId poolOwnerSlot ! ON DELETE CASCADE
```
jonathanknowles, line 84: We should replace this with suitable foreign key constraints.

lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs
```diff
@@ -315,6 +316,22 @@ instance PathPiece SlotId where
...
+
+instance PersistField SlotInternalIndex where
+    toPersistValue = toPersistValue . unSlotInternalIndex
+    fromPersistValue = fmap SlotInternalIndex . fromPersistValue
+
+instance ToJSON SlotInternalIndex where
+    toJSON = toJSON . unSlotInternalIndex
+
+instance FromJSON SlotInternalIndex where
+    parseJSON = fmap SlotInternalIndex . parseJSON
```
jonathanknowles: We should add round trip serialization tests for theses instances.

lib/core/src/Cardano/Pool/DB.hs
```diff
@@ -166,6 +187,64 @@ data DBLayer m = forall stm. (MonadFail stm, MonadIO stm) => DBLayer
         -- For a Sqlite DB, this would be "run a query inside a transaction".
     }
 
+-- | Represents an abstract notion of a certificate publication time.
+--
+-- Certificates published at later times take precedence over certificates
+-- published at earlier times.
+--
+type CertificatePublicationTime = (SlotId, SlotInternalIndex)
```
jonathanknowles: Consider making this into a `newtype`.

lib/core/src/Cardano/Pool/DB/Model.hs
jonathanknowles: Fixed in a75c879aabde68edcc1389a38a9e09b0bbdf17a7.

lib/core/src/Cardano/Pool/DB/Model.hs
jonathanknowles: Fixed in a75c879aabde68edcc1389a38a9e09b0bbdf17a7.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -354,22 +389,39 @@ monitorStakePools tr gp nl db@DBLayer{..} = do
...
+            -- their precedence relative to one another.
+            --
+            -- Certificates that appear later in a block have higher precedence
+            -- than certificates that appear earlier.
+            --
+            let certificateIndices = [minBound ..]
+            forM_ (certificateIndices `zip` certificates) $ \case
+                (slotInternalIndex, Registration cert) -> do
+                    liftIO $ traceWith tr $ MsgStakePoolRegistration cert
+                    putPoolRegistration (slot, slotInternalIndex) cert
```
KtorZ: isn't this slot internal index superfluous? 

a) All certificates within a block are inserted _atomically_, so they're either all inserted, or none of them are. 
b) A certificate present earlier in a block will be overridden by one later in the block
c) This is a strategy we've been using so far for all certificates, so I find the discrepancy here rather confusing. It suggests there's something special about these certificates, although there isn't (or am I missing something?) 

I'd suggest to drop the `slotInternalIndex` to not add on top of the existing complexity.

jonathanknowles: > isn't this slot internal index superfluous?

It turns out that:

1. A _single block_ can contain multiple certificates relating to the same pool.
2. A _single transaction_ can contain multiple certificates relating to the same pool.
3. The _relative order_ of these certificates is _significant_.
4. In general, earlier certificates are made obsolete by later certificates, but in particular:
    - a registration certificate will always _supersede_ an earlier retirement certificate (making that earlier retirement certificate obsolete);
    - a retirement certificate will always _augment_ the latest known registration certificate.

See [this thread](https://input-output-rnd.slack.com/archives/CCRB7BU8Y/p1593584824393100) for more information.

After discussing this with @rvl and @Anviking, we identified two potential methods of dealing with this issue:
- **Option A**: when processing a given block in a given slot **_s_**, if there are multiple registration and/or retirement certificates for a given pool **_p_**, then only record the latest registration and retirement certificates for that pool in the database, and ignore all prior certificates (subject to the rules stated above).
- **Option B**: when processing a given block in a given slot **_s_**, always record _all_ certificates for a given pool **_p_**, but also record the _order_ in which they appeared within the block. When processing a query for a given pool, fetch the latest registration and retirement certificates, and at _that point_ determine their relative precedence.

Comparing these two options:
* Option A has the advantage of retaining a simpler database design.
* Option B has the advantage of retaining more information to aid with debugging.

On balance, we felt that option B is the better choice. While it is true that this option requires more invasive changes to the pool database (an extra column in three tables to record the intra-slot order), in the event we need to perform debugging, the extra information (a record of all certificates that we've encountered, rather than just a subset) could prove to be very helpful.

lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs
jonathanknowles: > We should add round trip serialization tests for theses instances.

Fixed in cbab6b40db6d770a5dc2573969b142c44f3987c0.

lib/core/src/Cardano/Pool/DB/Sqlite/TH.hs
jonathanknowles: > We should replace this with suitable foreign key constraints.

Fixed in b28504ec32370917e83bd4b593b1cf7c45c76cfc.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -221,16 +230,24 @@ combineDbAndLsqData =
...
+            fmap fromIntegral . poolCost . registrationCert <$> dbData
+        , Api.pledge =
+            fmap fromIntegral . poolPledge . registrationCert <$> dbData
+        , Api.margin =
+            Quantity . poolMargin . registrationCert <$> dbData
           -- TODO: Report the actual retirement status of a pool.
           -- For the moment, we always report that a pool will never retire.
           -- See https://github.com/input-output-hk/cardano-wallet/milestone/89
-        , Api.retirement = Nothing
+        , Api.retirement =
```
rvl, line 238: Is the TODO comment above still valid?

lib/core/src/Cardano/Pool/DB.hs
rvl: It would need to be a `data` because there are two values.

I think remove `newtype SlotInternalIndex` and make this a data type.

Can you find a less wordy name for it?

lib/core/src/Cardano/Pool/DB.hs
```diff
@@ -166,6 +190,98 @@ data DBLayer m = forall stm. (MonadFail stm, MonadIO stm) => DBLayer
...
+
+-- | Given the latest registration and retirement certificates for a pool,
+-- determine the pool's current registration status based on the relative
+-- order in which the certificates were published.
+determinePoolRegistrationStatus
+    :: Ord certificatePublicationTime
+    => Maybe (certificatePublicationTime, PoolRegistrationCertificate)
+    -> Maybe (certificatePublicationTime, PoolRetirementCertificate)
+    -> PoolRegistrationStatus
+determinePoolRegistrationStatus = f
```
rvl: I'm pretty sure this could be made shorter with `case` and guards, without the need for helper functions f and g.
```suggestion
determinePoolRegistrationStatus a b = case (a, b) of
```

lib/core/src/Cardano/Pool/DB.hs
```diff
@@ -166,6 +190,98 @@ data DBLayer m = forall stm. (MonadFail stm, MonadIO stm) => DBLayer
...
+-- published at earlier times.
+--
+type CertificatePublicationTime = (SlotId, SlotInternalIndex)
+
+-- | Indicates the current registration status of a pool.
+--
+-- Use the 'readPoolRegistrationStatus' function to query the registration
+-- status for a particular pool and database backend.
+--
+data PoolRegistrationStatus
```
rvl: Why not represent this status as `Maybe (Registration, Maybe Retirement)`? I find the standard types easier to understand.

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
```diff
@@ -130,8 +158,20 @@ properties = do
             (property . prop_putStakeReadStake)
         it "putPoolRegistration then readPoolRegistration yields expected result"
             (property . prop_poolRegistration)
+        it "putPoolRetirement then readPoolRetirement yields expected result"
```
rvl, line 161: I can't seem to find any modifications to the DB state machine tests here.

I think you will get more thorough testing for your effort by updating the state machine generators and adding a few tags to check that they work.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -172,27 +179,29 @@ newStakePoolLayer gp nl db = StakePoolLayer
 --
 --
 
--- | Stake Pool data fields that we can fetch from the node over Local State
--- Query.
-data PoolLsqMetrics = PoolLsqMetrics
+-- | Stake pool-related data that has been read from the node using a local
+--   state query.
+data PoolLsqData = PoolLsqData
```
rvl, line 181: Since you're changing the name anyway... I don't think `Lsq` is descriptive enough of what the data is. It only describes the mechanism used to fetch it. How about `PoolStateFromLedger`?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -293,26 +312,42 @@ combineChainData =
...
+                (certificatesFromRegistrationStatus <$> registrationStatuses)
+            ]
     prodMap <- readTotalProduction
     metaMap <- readPoolMetadata
     return $ Map.map (lookupMetaIn metaMap) (combineChainData certMap prodMap)
   where
+    certificatesFromRegistrationStatus
+        :: PoolRegistrationStatus
+        -> Maybe (PoolRegistrationCertificate, Maybe PoolRetirementCertificate)
+    certificatesFromRegistrationStatus = \case
```
rvl, line 325: If PoolRegistrationStatus were replaced with the double Maybe then this function would not be needed.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -354,22 +389,39 @@ monitorStakePools tr gp nl db@DBLayer{..} = do
...
+            -- the same pool.
+            --
+            -- The order in which certificates appear within a block determines
+            -- their precedence relative to one another.
+            --
+            -- Certificates that appear later in a block have higher precedence
+            -- than certificates that appear earlier.
+            --
+            let certificateIndices = [minBound ..]
+            forM_ (certificateIndices `zip` certificates) $ \case
```
rvl: Can we please just say `zip [0..] certificates`.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -354,22 +389,39 @@ monitorStakePools tr gp nl db@DBLayer{..} = do
...
+
+            -- A Shelley block can contain multiple certificates relating to
+            -- the same pool.
+            --
+            -- The order in which certificates appear within a block determines
+            -- their precedence relative to one another.
+            --
+            -- Certificates that appear later in a block have higher precedence
+            -- than certificates that appear earlier.
+            --
```
rvl:
```suggestion
            -- When there are multiple certificates for a pool in one block,
            -- certificates that appear later in the block have higher precedence.
            -- It is possible to have a pool (re-)registration and retirement within the same block.
```

lib/core/src/Cardano/Pool/DB.hs
```diff
@@ -166,6 +190,98 @@ data DBLayer m = forall stm. (MonadFail stm, MonadIO stm) => DBLayer
...
+    f (Just (_, regCert)) Nothing =
+        PoolRegistered regCert
+    f (Just reg) (Just ret) =
+        g reg ret
+
+    g (regTime, regCert) (retTime, retCert)
+        | regPoolId /= retPoolId =
+            differentPoolsError
+        | regTime > retTime =
+            PoolRegistered regCert
```
rvl: Is this case correct?
I think this is the only non-obvious about this function, and it's worth a comment.

rvl (changes requested):
Looks pretty good but I have a few niggles...

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
jonathanknowles: > Is the TODO comment above still valid?

You're right! It's redundant.

Fixed in 1ab38cd965f5d12b210aa841856dc50797de2a12.

lib/core/src/Cardano/Pool/DB.hs
jonathanknowles: > I'm pretty sure this could be made shorter with `case` and guards, without the need for helper functions f and g.

It does indeed save a few lines (8 lines added in return for 12 lines deleted).

I've fixed this in c11624eb76afacf8448f279656da74b940f8b443.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
jonathanknowles: > Since you're changing the name anyway... I don't think `Lsq` is descriptive enough of what the data is. It only describes the mechanism used to fetch it. How about `PoolStateFromLedger`?

I do tend to agree with you here.

But on the other hand, the data for `PoolLsqData` comes (in part) from another data type called `NodePoolLsqData` (defined in `Cardano.Wallet.Shelley.Network`). See the function `combineLsqData` for more info.

If we were to change this name, I feel that we should also change the names of `NodePoolLsqData` and `combineLsqData` for consistency, as these manipulate the same sets of data.

But perhaps that is outside the scope of this PR? We could always make another PR with these changes (to avoid making this PR too big).

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
jonathanknowles: @rvl I've revised this comment a bit to make it more useful. (See 9efad1d76dc39641e79ed8caeff7f7003493644e)

Let me know if you think it could be revised further. :+1:

lib/core/src/Cardano/Pool/DB.hs
jonathanknowles: > Is this case correct?

Yes, it's correct. This case represents a re-registration (a registration that occurs after the prior publication of a retirement certificate), in which case the prior retirement is deemed to be obsolete. (See "pool inference rule" from the [Shelley ledger specification](https://hydra.iohk.io/build/3202141/download/1/ledger-spec.pdf))

> I think this is the only non-obvious about this function, and it's worth a comment.

Great point. I've added more comments in fcdfde6ce27c6f7744ecf7728ff1151908d4c959, which also adds a link to the specification.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
jonathanknowles: > Can we please just say `zip [0..] certificates`.

So we can't use `0` here, as `SlotInternalIndex` doesn't implement `Num`. But it does implement `Bounded`, hence the use of `minBound`.

As for using the name `certificateIndices`: I'm not sure we should get rid of that, as having the name serves as a kind of documentation for what these indices are supposed to be.

Perhaps there is a better name though?

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
jonathanknowles: > I can't seem to find any modifications to the DB state machine tests here.
> 
> I think you will get more thorough testing for your effort by updating the state machine generators and adding a few tags to check that they work.

Definitely agree with this.

Unfortunately there doesn't seem to be a state machine test suite for the pool database.

Perhaps we can add one in a future PR? I've started a discussion in the team channel here: https://input-output-rnd.slack.com/archives/GBT05825V/p1594183858000500

lib/core/src/Cardano/Pool/DB.hs
jonathanknowles: > Why not represent this status as `Maybe (Registration, Maybe Retirement)`? I find the standard types easier to understand.

Having a dedicated type means we have a single, obvious place to document what each of the states means. A single source of truth, if you like.

Consider the following pair of functions, both of which return the same type:

```hs
determinePoolLifeCycleStatus
    :: Ord publicationTime
    => Maybe (publicationTime, PoolRegistrationCertificate)
    -> Maybe (publicationTime, PoolRetirementCertificate)
    -> PoolLifeCycleStatus
```
```hs
readLifeCycleStatus
    :: DBLayer m
    -> PoolId
    -> m PoolLifeCycleStatus
```
A reader who is curious about what `PoolLifeCycleStatus` means can simply navigate to the definition and read the documentation for each case:

```hs
data PoolLifeCycleStatus
    = PoolNotRegistered
        -- ^ Indicates that a pool is not registered.
    | PoolRegistered
        PoolRegistrationCertificate
        -- ^ Indicates that a pool is registered BUT NOT marked for retirement.
        -- Records the latest registration certificate.
    | PoolRegisteredAndRetired
        PoolRegistrationCertificate
        PoolRetirementCertificate
        -- ^ Indicates that a pool is registered AND ALSO marked for retirement.
        -- Records the latest registration and retirement certificates.
```

IMO, even though the type `Maybe (PoolRegistrationCertificate, Maybe PoolRetirementCertificate)` is isomorphic to `PoolLifeCycleStatus`, it has the disadvantage of not being immediately clear to the uninformed reader what is meant by each of the cases.

We could document the meaning of each of the cases in all functions that consume or return this type, but that seems to be worse than just providing a dedicated type with its own documentation.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
jonathanknowles: > If PoolRegistrationStatus were replaced with the double Maybe then this function would not be needed.

That's true. But IMO this would reduce the readability of the following functions, both of which use the `PoolRegistrationStatus` type.

* `determinePoolRegistrationStatus`
* `readPoolRegistrationStatus`

See my comment here for a justification: https://github.com/input-output-hk/cardano-wallet/pull/1847#discussion_r451304532

lib/core/src/Cardano/Pool/DB.hs
jonathanknowles: I've promoted this type synonym into a concrete `data` type. (See 01fc73d8193597645069f1d1a4a5d7c0e79eeae2.)

jonathanknowles: @rvl wrote:
> Can you find a less wordy name for it?

I can't (currently) think of a shorter name for this.

For example, we could use `SlotIndex`, but IMO that doesn't make it clear enough that the index is _internal_ to a slot.

Do you have any ideas for a better name?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
jonathanknowles: @rvl I've rewritten this part of the code in 01fc73d8193597645069f1d1a4a5d7c0e79eeae2.

Let me know if you think it could be revised further!

lib/core/src/Cardano/Pool/DB.hs
```diff
@@ -166,6 +188,87 @@ data DBLayer m = forall stm. (MonadFail stm, MonadIO stm) => DBLayer
...
+determinePoolRegistrationStatus mReg mRet = case (mReg, mRet) of
+    (Nothing, _) ->
+        PoolNotRegistered
+    (Just (_, regCert), Nothing) ->
+        PoolRegistered regCert
+    (Just (regTime, regCert), Just (retTime, retCert))
+        | regPoolId /= retPoolId ->
+            -- Comparing certificates from different pools is a programming
+            -- error.
+            differentPoolsError
```
rvl: You may as well remove the comment and inline the differentPoolsError string -- it's self-documenting.

lib/core/src/Cardano/Pool/DB.hs
```diff
@@ -166,6 +188,87 @@ data DBLayer m = forall stm. (MonadFail stm, MonadIO stm) => DBLayer
...
+            -- A re-registration always /supercedes/ a prior retirement.
+            PoolRegistered regCert
+        | regTime < retTime ->
+            -- A retirement always /augments/ the latest known registration.
+            PoolRegisteredAndRetired regCert retCert
+        | otherwise ->
+            -- If a registration certificate and a retirement certificate
+            -- for the same pool appear to have been published at exactly
+            -- the same time, this indicates a programming error.
+            timeCollisionError
```
rvl: Same here - also the error messages should mention "programming error", so that if it ever occurs - which it won't - the user knows that it's not their fault.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -196,62 +200,114 @@ newDBLayer trace fp = do
...
+
+        , readPoolRetirement = \poolId -> do
+            let filterBy = [ PoolRetirementPoolId ==. poolId ]
+            let orderBy =
+                    [ Desc PoolRetirementSlot
+                    , Desc PoolRetirementSlotInternalIndex
+                    ]
+            selectFirst filterBy orderBy >>= \case
+                Nothing -> pure Nothing
+                Just meta -> do
```
rvl: How about this way?

```suggestion
            res <- selectFirst [ PoolRetirementPoolId ==. poolId ]
                    [ Desc PoolRetirementSlot
                    , Desc PoolRetirementSlotInternalIndex
                    ]
            forM res $ \meta -> do
```

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -1771,6 +1784,37 @@ instance Buildable PoolRetirementCertificate where
...
+-- | Represents an abstract notion of a certificate publication time.
+--
+-- Certificates published at later times take precedence over certificates
+-- published at earlier times.
+--
+data CertificatePublicationTime = CertificatePublicationTime
+    { slotId
+        :: SlotId
+    , slotInternalIndex
+        :: SlotInternalIndex
```
rvl: Adding this newtype only clouds the meaning. Almost any integral type would be better.

rvl (approved):

lib/core/src/Cardano/Pool/DB.hs
```diff
@@ -166,6 +188,87 @@ data DBLayer m = forall stm. (MonadFail stm, MonadIO stm) => DBLayer
...
+
+        timeCollisionError = error $ mconcat
+            [ "determinePoolRegistrationStatus:"
+            , " called with identical certificate publication times:"
+            , " pool id of registration certificate: "
+            , show regPoolId
+            , " pool id of retirement certificate: "
+            , show retPoolId
+            , " publication time of registration certificate: "
+            , show regPoolId
```
jonathanknowles: This should be a publication time, rather than a pool ID.

lib/core/src/Cardano/Pool/DB.hs
```diff
@@ -166,6 +188,87 @@ data DBLayer m = forall stm. (MonadFail stm, MonadIO stm) => DBLayer
...
+            , " publication time of retirement certificate: "
+            , show retPoolId
```
jonathanknowles: This should be a publication time, rather than a pool ID.

lib/core/src/Cardano/Pool/DB.hs
jonathanknowles: > You may as well remove the comment and inline the differentPoolsError string -- it's self-documenting.

Thanks! I've removed the comment, and revised the error string to make it clear that it's a programming error.

(The only thing I haven't changed is to inline `differentPoolsError`, in order to keep the pattern matching fairly compact.)

lib/core/src/Cardano/Pool/DB.hs
jonathanknowles: > Same here - also the error messages should mention "programming error", so that if it ever occurs - which it won't - the user knows that it's not their fault.

> You may as well remove the comment and inline the differentPoolsError string -- it's self-documenting.

Thanks! I've removed the comment, and revised the error string to make it clear that it's a programming error.

lib/core/src/Cardano/Pool/DB.hs
jonathanknowles: Fixed in eae043b6cb8510cb292cb23aec211d3296edfd20.

lib/core/src/Cardano/Pool/DB.hs
jonathanknowles: Fixed in eae043b6cb8510cb292cb23aec211d3296edfd20.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
jonathanknowles: > How about this way?

Nice!

The function `readPoolRegistration` was originally written this way, so I extended the style when writing `readPoolRetirement`. But your way is shorter and clearer.

I've applied it to the pair of functions `readPool{Registration,Retirement}`

Fixed in: c79a3300883dfc4bd62c18ec23aa730bb4a208b6.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
jonathanknowles: > Adding this newtype only clouds the meaning. Almost any integral type would be better.

Fair enough!

I've removed the `SlotInternalIndex` type by inlining it into `CertificatePublicationTime`, resulting in the following type:
```hs
data CertificatePublicationTime = CertificatePublicationTime
    { slotId
        :: SlotId
    , slotInternalIndex
        :: Word64
        -- ^ Indicates the relative position of a publication within a slot.
    }
    deriving (Eq, Generic, Ord, Show)
```
Fixed in c7e5da754889c5ba684fbd82fa3d899996ff0624.

lib/core/src/Cardano/Pool/DB.hs
jonathanknowles: I've removed the `SlotInternalIndex` type, inlining the definition into `CertificatePublicationTime`. (See c7e5da754889c5ba684fbd82fa3d899996ff0624.)

jonathanknowles:
bors retry

lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -196,62 +203,104 @@ newDBLayer trace fp = do
...
-                    poolMarginD
-                    poolCost_
-                    poolPledge_
-                    poolMetadataUrl
-                    poolMetadataHash
-            insertMany_ $ zipWith (PoolOwner poolId point) poolOwners [0..]
+        , putPoolRegistration = \cpt cert -> do
+            let CertificatePublicationTime {slotId, slotInternalIndex} = cpt
+            let poolId = view #poolId cert
+            deleteWhere [PoolOwnerPoolId ==. poolId, PoolOwnerSlot ==. slotId]
```
jonathanknowles: This looks as though it may be deleting _too many_ rows.

We probably want to restrict this to deleting rows with a matching value of `PoolOwnerSlotInternalIndex`.

jonathanknowles: > This looks as though it may be deleting _too many_ rows.
> 
> We probably want to restrict this to deleting rows with a matching value of `PoolOwnerSlotInternalIndex`.

Fixed in 31f8e0fd8a15814b4624dde566c608de0fb44ec7.

jonathanknowles:
bors try

jonathanknowles:
bors retry

Issue #1844, opened 2020-07-01
Introduce diverse witnesses in shelley


paweljakubas:
# Context

As a user
I want to move my funds from my Byron wallet to my shelley wallet
so that I manage my funds

# Decision

We want to enable sending txs with Icarus/Byron witnesses within shelley style wallet. This will require introducing `TxWitnessTagFor` typeclass on key parameter, add the constraint in `newTransactionLayer`. This will allow adding dedicated `mkWitness` implementation, also differentiate `postTx`

# Acceptance Criteria
1. `TxWitnessTagFor` typeclass and impl for IrarusKey, ByronKey, and ShelleyKey are introduced.
2. `newTransactionLayer` is constraint with `TxWitnessTagFor`
3. Specialized impl of `mkWitness` for keys are added.
4. Estimated fee is revised (IcarusKey and ByronKey witnesses are more costly)
5. Integration tests illustrating the case are added. 

---

# Development 

# QA 

I have demonstrated that decodeTx for shelley/byron witnesses work -> see TransactionSpec tests in shelley 
Also showed that byron/icarus txs to shelley work! 
It would be good to enable ByronTransactions for shelley integration testing and see byron/icarus -> byron/icarus txs work.
Also, we resigned from mkFeeestimator and use fee estimation endpoint - it would be worth to propagate this change throught code

piotr-iohk:
LGTM, I have enabled additional tests for Byron txs, migrations (and also tests for byron addresses, and hw wallets) in integration tests. Also extracted the newly introduced integration test byron tx -> shelley to separate file not to run it on Jormungandr. (Jorm does not support Byron txs so it failed nightly last night). All in -> #1900.

piotr-iohk:
Closing. This is basically working OK. Still #1900 with additional tests to be merged.

PR #1849, opened 2020-07-03
Diverse witnesses in shelley

paweljakubas:
# Issue Number

#1844 

# Overview

- [x] I have added WitnessTag
- [x] I have added impl of mkTx for Byron/Icarus wtinesses
- [x] I have added unit tests verifying decodeTx for Byron/Icarus witnesses
- [x] I have updated fee estimation
- [x] I have added integration tests illustrating the case 

# Comments


lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -298,3 +382,15 @@ dummyAddress = Address $ BS.pack $ 1 : replicate 64 0
...
+    :: forall mw ent csz.
+     ( ConsistentEntropy ent mw csz
+     , EntropySize mw ~ ent
+     )
+    => Gen (Mnemonic mw)
+genMnemonic = do
+    let n = fromIntegral (natVal $ Proxy @(EntropySize mw)) `div` 8
+    bytes <- BS.pack <$> vector n
+    let ent = unsafeMkEntropy @(EntropySize mw) bytes
+    return $ entropyToMnemonic ent
```
KtorZ: Why not simply use `genEntropy` with `EntropySize mw` to get the right number of bytes?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -355,6 +384,26 @@ _decodeSignedTx bytes = do
...
+    let (Cardano.TxUnsignedByron _ _ txHash) =
+            mkUnsignedByronTx ownedIns outs
+    let tx = (fst <$> ownedIns, outs)
+    let sigData = Byron.hashToBytes txHash
+    witnesses <- forM ownedIns $ \(_, TxOut addr _) -> do
+        ks <- lookupPrivateKey keyFrom addr
+        pure $ mkByronWitness pm sigData ks
+    pure
+        ( Tx (Hash sigData) (second coin <$> ownedIns) outs
+        , SealedTx $ CBOR.toStrictByteString $ CBOR.encodeSignedTx tx witnesses
```
KtorZ: I don't think this is right. This is how transaction were serialized in Byron, but in Shelley, they are following a different scheme. Therefore, we should serialize transaction using the new shelley methods; only the witnesses are different!

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -164,25 +194,34 @@ mkTx
...
+                <> mkExtraWits unsigned
 
-    let wits = (SL.WitnessSet (addrWits <> withdrawalsWits) mempty mempty)
-            <> mkExtraWits unsigned
+        TxWitnessByronUTxO -> do
+            bootstrapWits <- fmap Set.fromList $ forM (CS.inputs cs) $ \(_, TxOut addr _) -> do
+                (k, pwd) <- lookupPrivateKey keyFrom addr
+                pure $ mkByronWitness unsigned pm (getRawKey k, pwd)
+            pure $ SL.WitnessSet mempty mempty bootstrapWits
+                <> mkExtraWits unsigned --- TO_DO - this is needed?
```
paweljakubas: this one is maybe not needed - I do not know at this moment

Anviking: Yes, if extra witnesses are needed for certificates or whatnot, they end up here

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -421,7 +468,32 @@ computeTxSize proxy action cs =
...
-    wits = SL.WitnessSet (Set.union addrWits certWits) mempty mempty
+    byronWits = Set.map dummyWitnessUniq $ Set.fromList (fst <$> CS.inputs cs)
+      where
+        dummyWitness :: BL.ByteString -> SL.BootstrapWitness TPraosStandardCrypto
+        dummyWitness chaff =
+            SL.makeBootstrapWitness credential signingKey addrAttr
+          where
+            credential = Hash.UnsafeHash $ L8.toStrict chaff
+            addrAttr = Byron.mkAttributes $ Byron.AddrAttributes
+                       (Just $ Byron.HDAddressPayload (bloatChaff keyLen))
```
KtorZ: Why `keyLen` here :thinking:  ? The HDAddressPayload is a ChaChaPoly digest + nonce that totals 28 bytes + 2 bytes because it's CBOR-encoded. I believe the `keyLen` is 32 bytes here so it's probably not a bit deal, but that is quite misleading and error-prone.

More importantly, there's only a HDPayload for /Byron/ type of addresses, Icarus do not.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -421,7 +468,32 @@ computeTxSize proxy action cs =
...
+                       (toByronNetworkMagic mainnetMagic)
```
KtorZ: mainnet is hard-coded here which is sort of wrong because mainnet addresses don't embed any discrimination attribute whereas testnet address do. So there's a significant size variation between the two.

KtorZ: suggestion: rely on the `proxy` above to know whether mainnet or testnet is needed. See also: `fromNetworkDiscriminant`

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -504,6 +576,20 @@ unsafeMkEd25519 :: XPub -> Ed25519.PublicKey
...
+mkByronWitness
+    :: SL.TxBody TPraosStandardCrypto
+    -> ProtocolMagic
+    -> (XPrv, Passphrase "encryption")
+    -> SL.BootstrapWitness TPraosStandardCrypto
+mkByronWitness body protocolMagic (prv, _pwd) =
+    let signingKey = Crypto.SigningKey prv
+        bytes = xprvToBytes prv
+        addrAttr = Byron.mkAttributes $ Byron.AddrAttributes
+            (Just $ Byron.HDAddressPayload bytes)
```
KtorZ: I don't think we should send the user's private key over in a witness! 
The HDAddressPayload corresponds to the encrypted derivation path that is present only on Byron-Random addresses. I assume that what we needs here is: 

a) To pass down the corresponding address to `mkByronWitness`
b) Have a function in the Compatibility module to convert an `Address` into a `AddrAttributes`. We have all the information we need in the Address already.

KtorZ (changes requested):
The witness for Byron needs to be corrected, in particular the HD payload needs to be inferred from the address!

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
paweljakubas: have a try to deliver proper solution in https://github.com/input-output-hk/cardano-wallet/pull/1849/commits/72c70d0f90a40ac481bdbebd48199d67f0745163

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
paweljakubas: addressed here https://github.com/input-output-hk/cardano-wallet/pull/1849/commits/1209bca45f0c0dd9cd8615355eee19df11893e46

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
paweljakubas: can I retrieve protocolMagic easily from proxy? :thinking: 
I have above (I agree it was quick hack) : 
```
763 toByronNetworkMagic :: W.ProtocolMagic -> Byron.NetworkMagic                                                                                                                                                   
764 toByronNetworkMagic pm@(W.ProtocolMagic magic) =                                                                                                                                                               
765     if pm == W.mainnetMagic then                                                                                                                                                                               
766         Byron.NetworkMainOrStage                                                                                                                                                                               
767     else                                                                                                                                                                                                       
768         Byron.NetworkTestnet (fromIntegral magic)
```

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -164,25 +195,34 @@ mkTx
...
-    addrWits <- fmap Set.fromList $ forM (CS.inputs cs) $ \(_, TxOut addr _) -> do
-        (k, pwd) <- lookupPrivateKey keyFrom addr
-        pure $ mkWitness unsigned (getRawKey k, pwd)
+    wits <- case (txWitnessTagFor @k) of
+        TxWitnessShelleyUTxO -> do
+            addrWits <- fmap Set.fromList $ forM (CS.inputs cs) $ \(_, TxOut addr _) -> do
+                (k, pwd) <- lookupPrivateKey keyFrom addr
+                pure $ mkShelleyWitness unsigned (getRawKey k, pwd)
+
+            let withdrawalsWits
```
Anviking, line 211: I was first wondering: why not factor out `fmap Set.fromList $ forM (CS.inputs cs) $ \(_, TxOut addr _)` from both cases, and just differentiate on the witness construction?

Then I see `withdrawalsWits` are only in the `TxWitnessShelleyUTxO` case. Wouldn't it _technically_ be possible (at least for people using cardano-wallet as a library) to spend both byron funds and withdrawals? Would it not be as well to construct `withdrawalsWits` in both cases then?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
paweljakubas: yeah, but in byron transactions it should not happen to my best knowledge, right?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
paweljakubas: that's good question - I have also this in the back of my head. I planned to have working byron/icarus tx...and then in the next PR check this possibility

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
KtorZ: In theory they should match indeed.

paweljakubas:
bors try

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -464,6 +465,70 @@ spec = do
...
+                (Link.getWallet @'Shelley wShelley) Default Empty
+            expectField
+                (#balance . #getApiT . #available)
+                (`shouldBe` Quantity (faucetAmt + amt)) rb
+
+            ra2 <- request @ApiByronWallet ctx
+                (Link.getWallet @'Byron wByron) Default Empty
+            expectField
+                (#balance . #available)
+                (`shouldBe` Quantity (faucetAmt - feeEstMax - amt)) ra2
```
KtorZ, line 534: :+1:

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
KtorZ: > to spend both byron funds and withdrawals?

Technically yes, but not with the wallet, because there are no withdrawals coming with our definition of Byron wallets.

KtorZ (approved):
:+1:

PR #1850, opened 2020-07-03
Make quitting on a non-empty reward account forbidden

KtorZ:
# Issue Number

# Overview

The Cardano ledger forbids to de-register a stake key, if there are
still rewards associated with it. We could have implemented some more
convoluted logic where the wallet would automatically collect rewards as
part of a de-registration, but we are a bit lacking time here so we'll
simply make this a manual operation that users have to undergo. Yet, the
backend will still return a meaningful error message explaining the
situation.

# Comments

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
@@ -95,6 +95,7 @@ import Test.Integration.Framework.DSL
     )
 import Test.Integration.Framework.TestData
     ( errMsg403DelegationFee
+    , errMsg403NonNullReward
     , errMsg403NotDelegating
     , errMsg403PoolAlreadyJoined
     , errMsg403WrongPass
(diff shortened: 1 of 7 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -339,38 +363,11 @@ spec = do
...
---        reward <- getFromResponse (#balance . #getApiT . #reward) <$>
---            request @ApiWallet ctx (Link.getWallet @'Shelley w) Default Empty
---
---        waitForNextEpoch ctx
---        request @ApiWallet ctx (Link.getWallet @'Shelley w) Default Empty >>= flip verify
---            [ expectField
---                    (#balance . #getApiT . #reward)
---                    (`shouldBe` reward)
---            ]
+        -- Can't quite if unspoiled rewards.
```
Anviking, line 366:
```suggestion
        -- Can't quit if unspoiled rewards.
```
But also, why not "Can't quit with remaining rewards.", or even "Can't quit with unredeemed rewards."?

The term "unspoiled" seems a bit weird to me at least.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2131,11 +2133,16 @@ guardJoin knownPools WalletDelegation{active,next} pid = do
...
+    -> Quantity "lovelace" Word64
     -> Either ErrCannotQuit ()
-guardQuit WalletDelegation{active,next} = do
+guardQuit WalletDelegation{active,next} rewards = do
     let last_ = maybe active (view #status) $ lastMay next
+
     unless (isDelegatingTo anyone last_) $
         Left ErrNotDelegatingOrAboutTo
+
+    unless (rewards == Quantity 0) $
```
Anviking, line 2144: > We could have implemented some more
convoluted logic where the wallet would automatically collect rewards as
part of a de-registration, but we are a bit lacking time here so we'll
simply make this a manual operation that users have to undergo.

👌

Anviking (approved):
lgtm

KtorZ:
bors try

PR #1851, opened 2020-07-03
Track stake key registrations independently from delegation

Anviking:
# Issue Number

#1835 

# Overview

- [x] Keep track of stake key registrations such that we don't try to register a key twice
- [ ] TODO: model impl
- There was a integration test, which I saw passing at some point, but I've not rebased it to this PR
- Some DB model-like tests would have been nice, but no time now

# Comments

To delegate in shelley we need to know whether we need to register the
stake key, or if it already exists. Trying to register the same key
twice will cause the delegation to fail.

In shelley there are three delegation certs:
- reg
- dereg
- delegate

We need to keep track of two things:
- Delegation status
- Is the stake key registered or not? (new in this commit)

Also new in this commit:
- Stake key de-registration inserts an uptdate to the delegation status
table that says that it is not delegating.


lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -604,9 +608,32 @@ newDBLayer trace defaultFieldValues mDatabaseFile = do
...
-                    (DelegationCertificateKey wid sl)
-                    (DelegationCertificate wid sl (W.dlgCertPoolId cert))
+                Just _  -> case cert of
+                    W.CertDelegateNone _ -> do
+                        repsert
+                            (DelegationCertificateKey wid sl)
+                            (DelegationCertificate wid sl Nothing)
+                        pure <$> repsert
+                            (StakeKeyCertificateKey wid sl)
+                            (StakeKeyCertificate wid sl False)
```
Anviking: Here we insert "deregistration" into both tables

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1712,13 +1712,17 @@ joinStakePool
...
+    (isKeyReg, walMeta) <- mapExceptT atomically $ withExceptT ErrJoinStakePoolNoSuchWallet $
+        (,)
+            <$> readIsStakeKeyRegistered (PrimaryKey wid)
+            <*> withNoSuchWallet wid (readWalletMeta (PrimaryKey wid))
 
     withExceptT ErrJoinStakePoolCannotJoin $ except $
         guardJoin pools (walMeta ^. #delegation) pid
 
-    let action = nextJoinAction pid walMeta
+    let action = if isKeyReg then Join pid else RegisterKeyAndJoin pid
```
Anviking, line 1723: To preserve the `isKeyReg then` for readability (and not have to add a new type) I removed the `nextJoinAction`-helper.

Slight duplication, but I think it's ok.

KtorZ:
> Stake key de-registration inserts an uptdate to the delegation status
table that says that it is not delegating.

This seems wrong. Stake key de-registration should be tracked separately as well, and the delegation status should be inferred from the database current status (as it is now).

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -604,9 +608,32 @@ newDBLayer trace defaultFieldValues mDatabaseFile = do
...
+                    W.CertDelegateFull _ pool ->
+                        pure <$> repsert
+                            (DelegationCertificateKey wid sl)
+                            (DelegationCertificate wid sl (Just pool))
+                    W.CertRegisterKey _ ->
+                        pure <$> repsert
+                            (StakeKeyCertificateKey wid sl)
+                            (StakeKeyCertificate wid sl True)
```
KtorZ: ~~Hmmm. This isn't quite right. Stake key are registered through their own certificates. So this is what we should be tracking. In particular, the case of ITN, the key is registered by _another mean_ than the wallet backend so we never get to create a registration certificate in the first place.~~

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2239,6 +2244,7 @@ instance HasSeverityAnnotation WalletLog where
         MsgPaymentCoinSelectionStart{} -> Debug
         MsgPaymentCoinSelection _ -> Debug
         MsgPaymentCoinSelectionAdjusted _ -> Debug
+        MsgIsStakeKeyRegistered _ -> Info
```
KtorZ, line 2254: What about:

- `Info` when `False`
- `Debug` otherwise.

Anviking: `False` would be expected for new wallets.

`True` would be common for people will ITN rewards.

Why not just `Info` for both?

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
```diff
@@ -156,6 +156,16 @@ ProtocolParameters
     Foreign Wallet fk_wallet_protocol_parameters protocolParametersWalletId ! ON DELETE CASCADE
     deriving Show Generic
 
+-- Track whether the wallet's stake key is registered or not.
+StakeKeyCertificate
+    stakeKeyCertWalletId             W.WalletId     sql=wallet_id
+    stakeKeyCertSlot                 W.SlotId       sql=slot
+    stakeKeyCertIsReg                Bool           sql=is_reg
```
KtorZ: I'd recommend using something more explicit than a boolean here. Boolean convey very little information and they are easy to misinterpret (like I did on my first review...). I'd suggest something like `Registration | Deregistration`, stored as plain text lowercased in the database, so it's also readable when debugging / inspecting.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
KtorZ: Okay, went through too quickly at first when I saw the `repsert` and the boolean. Looks like this would work indeed and actually play nicely with both Jörmungandr and cardano-node.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -604,9 +608,32 @@ newDBLayer trace defaultFieldValues mDatabaseFile = do
...
+
+        , readIsStakeKeyRegistered = \(PrimaryKey wid) -> ExceptT $ do
+            val <- fmap entityVal <$> selectFirst
+                    [StakeKeyCertWalletId ==. wid]
+                    [Desc StakeKeyCertSlot]
+
+            return $ case val of
+                Nothing -> Left $ ErrNoSuchWallet wid
```
KtorZ: Hmmm. No. There could be a wallet but no key. This should return negatively about the existing of a key, but not return an error. To know whether there's a wallet, this should use the same pattern as for other functions:

```hs
            selectWallet wid >>= \case
                Just{} -> {- ... do actual stuff ... -}
```

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
```diff
@@ -156,6 +156,16 @@ ProtocolParameters
...
+
+    Primary stakeKeyCertWalletId stakeKeyCertSlot
```
KtorZ, line 165: :+1:

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -676,7 +676,8 @@ fromShelleyDelegationCert = \case
     SL.DCertDeleg (SL.DeRegKey credentials) ->
         Just $ W.CertDelegateNone (fromStakeCredential credentials)
 
-    SL.DCertDeleg SL.RegKey{} -> Nothing
+    SL.DCertDeleg (SL.RegKey cred) ->
+        Just $ W.CertRegisterKey $ fromStakeCredential cred
```
KtorZ, line 680: :+1:

lib/core/src/Cardano/Wallet.hs
KtorZ: It's true that we aren't going to see this log line often actually. Only occurs when changing delegation settings, so I guess it's fine :+1:

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -626,14 +626,15 @@ newDBLayer trace defaultFieldValues mDatabaseFile = do
...
-
+              selectWallet wid >>= \case
+                Nothing -> pure $ Left $ ErrNoSuchWallet wid
+                Just{} -> do
+                    val <- fmap entityVal <$> selectFirst
+                        [StakeKeyCertWalletId ==. wid]
+                        [Desc StakeKeyCertSlot]
+                    return $ case val of
+                        Nothing -> Right False
+                        Just (StakeKeyCertificate _ _ isReg) -> Right isReg
```
KtorZ: :+1:

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -369,6 +373,30 @@ spec = do
...
+                , "stamp" , "chest", "owner", "excess", "omit"
+                , "bid", "raccoon", "spin" , "reduce", "rival"
+                ]
+        let payload = Json [json| {
+                "name": "Wallet with pre-registered stake key",
+                "mnemonic_sentence": #{walletWithPreRegKey},
+                "passphrase": "Secure Passphrase"
+                } |]
+
+        threadDelay 4000000
```
Anviking: Should use `eventually` I guess

lib/shelley/test/integration/Main.hs
```diff
@@ -321,3 +337,53 @@ minSeverityFromEnv def var = lookupEnv var >>= \case
     Nothing -> pure def
     Just "" -> pure def
     Just arg -> either die pure (parseLoggingSeverity arg)
+
+-- | Pre-register a staking key for the STAKE_POOLS_JOIN_05 test.
+preregisterStakingKeysForTests :: NetworkLayer IO t b -> IO ()
+preregisterStakingKeysForTests nl = do
```
KtorZ: I am not fond of having this function here but I guess it'll do for now. Ideally, I'd have put that where the rest of the "setup & fixture" work is happening (i.e. in the "Launch" module).

lib/shelley/test/integration/Main.hs
```diff
@@ -321,3 +337,53 @@ minSeverityFromEnv def var = lookupEnv var >>= \case
...
+    let dummyOut = TxOut addr (Coin 10000000000)
+
+    let cs = CoinSelection
+            { inputs = [(txIn, dummyOut)]
+            , withdrawal = 0
+            , reclaim = 0
+            , outputs = []
+            , change = []
+            , deposit = 100000 -- keyDeposit
+            }
```
KtorZ: CoinSelection is a `Monoid`, so you can write instead: 

```hs
mempty { inputs = [(txIn, dummyOut)], deposit = 100000 }
```

Also, note that the key deposit is available in protocol parameters, would be better than hard-coding it.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -1844,19 +1845,30 @@ instance NFData ChimericAccount
...
     | CertDelegateFull ChimericAccount PoolId
+    | CertRegisterKey ChimericAccount
     deriving (Generic, Show, Eq, Ord)
 
 instance NFData DelegationCertificate
 
+data StakeKeyCertificate
+    = StakeKeyRegistration
+    | StakeKeyDeregistration
+    deriving (Generic, Show, Read, Eq)
```
Anviking, line 1856: There are so many different kinds of certificate types in here already, for multiple purposes. I'd preferred to keep it closer to the DB.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
KtorZ: Why the delay?

lib/shelley/test/integration/Main.hs
Anviking: > key deposit is available in protocol parameters

👍 

> CoinSelection is a Monoid

Didn't think about it, but in this case I found it useful to explicitly see the new fields of `CoinSelection`.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
Anviking: We try to join immediately when the wallet is created. It has no time to find the existing key reg cert.

Should probably use `eventually` though. Just sprinkled it in to get it passing.

KtorZ (approved):
Reviewed, partially fixed and partially extended.

PR #1854, opened 2020-07-04
 assign correct network discrimination to withdrawals

KtorZ:
# Issue Number

#1821 

# Overview

- [ ] I have assigned correct network discrimination to withdrawals. This was hard-coded to mainnet, for some reasons.

# Comments

### DIFF
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
@@ -73,6 +73,7 @@ import Cardano.Wallet.Primitive.Types
 import Cardano.Wallet.Shelley.Compatibility
     ( Shelley
     , TPraosStandardCrypto
+    , fromNetworkDiscriminant
     , toCardanoLovelace
     , toCardanoTxIn
     , toCardanoTxOut
@@ -105,8 +106,6 @@ import Data.Proxy
     ( Proxy (..) )
 import Data.Quantity
     ( Quantity (..) )
-import Data.Type.Equality
-    ( testEquality )
 import Data.Word
     ( Word16, Word64, Word8 )
 import Ouroboros.Consensus.Shelley.Protocol.Crypto
@@ -114,7 +113,7 @@ import Ouroboros.Consensus.Shelley.Protocol.Crypto
 import Ouroboros.Network.Block
     ( SlotNo )
 import Type.Reflection
-    ( Typeable, typeRep )
+    ( Typeable )
 
 import qualified Cardano.Api as Cardano
 import qualified Cardano.Byron.Codec.Cbor as CBOR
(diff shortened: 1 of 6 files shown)


lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -459,19 +473,24 @@ mkUnsignedTx ttl cs withdrawals certs =
...
           , SL.Coin $ fromIntegral amount
           )
         ]
   where
     keyHashObj = SL.KeyHashObj $ SL.KeyHash $ Hash.UnsafeHash keyHash
-
+    network =
+        case testEquality (typeRep @n) (typeRep @'Mainnet) of
+            Just{}  -> SL.Mainnet
+            Nothing -> SL.Testnet
```
KtorZ: relevant part is here :point_up:

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -193,24 +199,25 @@ newTransactionLayer
...
         , WalletKey k
+        , Typeable n
         )
     => Proxy n
     -> ProtocolMagic
     -> EpochLength
     -> TransactionLayer t k
-newTransactionLayer _proxy _protocolMagic epochLength = TransactionLayer
-    { mkStdTx = \acc ks tip -> mkTx emptyTxPayload (defaultTTL epochLength tip) acc ks
+newTransactionLayer proxy _protocolMagic epochLength = TransactionLayer
```
paweljakubas, line 207: in the light of this change - we need still protocolMagic, right?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
paweljakubas: just wondering if this part should not go to Compatibility, we have there number of network translation like  https://github.com/input-output-hk/cardano-wallet/blob/master/lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs#L882 or https://github.com/input-output-hk/cardano-wallet/blob/master/lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs#L572

paweljakubas (approved):
LGTM

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
KtorZ: Ho yes, it probably should. Happy to do it on Monday :sweat_smile:

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
KtorZ: Actually no :thinking: ... Since there's only one testnet in Shelley, the network is either mainnet or testnet.

paweljakubas: hehe, it was supposed to be a question - **so do we still need protocol magic?** (probably redundant - agree here, hence stopped at this place)

paweljakubas: on the other hand we will need protocol magic for icarus/byron witnesses very soon

KtorZ: We do need it for the Byron transaction layer indeed, and for the associated witnesses :)

KtorZ:
bors retry

PR #1859, opened 2020-07-06
install a query flag for driving withdrawals creation on Shelley tx

KtorZ:
# Issue Number

# Overview

- b448b07816a7efd42e65393fb92257693a16e9b4
  :round_pushpin: **install a query flag for driving withdrawals creation on Shelley transactions**
  Withdrawals currently occur by default, implicitely which turns out to
be very confusing for end-users (seeing their rewards balance disappear
freak them out, as well as transaction which seems much bigger than what
they requested).

# Comments

Screenshot from 2020-07-06 07-53-08

### DIFF
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -128,6 +128,7 @@ import Cardano.Wallet.Api.Types
     , ApiT (..)
     , ApiTxId (ApiTxId)
     , ApiWallet
+    , ApiWithdrawRewards (..)
     , ByronWalletPostData (..)
     , ByronWalletStyle (..)
     , Iso8601Time (..)
@@ -689,6 +690,7 @@ data TransactionCreateArgs t = TransactionCreateArgs
     { _port :: Port "Wallet"
     , _id :: WalletId
     , _payments :: NonEmpty Text
+    , _withdrawRewards :: Bool
     }
 
 cmdTransactionCreate
@@ -704,7 +706,8 @@ cmdTransactionCreate mkTxClient mkWalletClient =
         <$> portOption
         <*> walletIdArgument
         <*> fmap NE.fromList (some paymentOption)
-    exec (TransactionCreateArgs wPort wId wAddressAmounts) = do
+        <*> withdrawRewardsFlag
+    exec (TransactionCreateArgs wPort wId wAddressAmounts wWithdraw) = do
         wPayments <- either (fail . getTextDecodingError) pure $
             traverse (fromText @(AddressAmount Text)) wAddressAmounts
         res <- sendRequest wPort $ getWallet mkWalletClient $ ApiT wId
@@ -714,6 +717,7 @@ cmdTransactionCreate mkTxClient mkWalletClient =
                 runClient wPort Aeson.encodePretty $ postTransaction
                     mkTxClient
                     (ApiT wId)
+                    (ApiWithdrawRewards wWithdraw)
                     (Aeson.object
                         [ "payments" .= wPayments
                         , "passphrase" .= ApiT wPwd
@@ -735,7 +739,8 @@ cmdTransactionFees mkTxClient mkWalletClient =
         <$> portOption
         <*> walletIdArgument
         <*> fmap NE.fromList (some paymentOption)
-    exec (TransactionCreateArgs wPort wId wAddressAmounts) = do
+        <*> withdrawRewardsFlag
+    exec (TransactionCreateArgs wPort wId wAddressAmounts wWithdraw) = do
         wPayments <- either (fail . getTextDecodingError) pure $
             traverse (fromText @(AddressAmount Text)) wAddressAmounts
         res <- sendRequest wPort $ getWallet mkWalletClient $ ApiT wId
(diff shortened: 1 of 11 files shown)


lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1040,7 +1040,9 @@ selectCoins ctx gen (ApiT wid) body =
     fmap mkApiCoinSelection
     $ withWorkerCtx ctx wid liftE liftE
     $ \wrk -> do
-        withdrawal <- liftIO $ W.readNextWithdrawal @_ @s @t @k wrk wid
+        -- TODO:
+        -- Allow representing withdrawals as part of external coin selections.
+        let withdrawal = Quantity 0
```
KtorZ, line 1045: I went quite fast in the first iteration here, but this `withdrawal` value actually has no effect because `selectCoinsExternal` can't actually process nor represent them, so changing this to be `0` and left a little TODO.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -184,10 +185,12 @@ spec = do
...
-        request @(ApiTransaction n) ctx (Link.createTransaction @'Shelley w) Default (Json payload) >>= flip verify
-            [ expectField #amount (.> (Quantity coin))
-            , expectField (#direction . #getApiT) (`shouldBe` Outgoing)
-            ]
+        request @(ApiTransaction n) ctx
+            (Link.createTransaction' @'Shelley w (ApiWithdrawRewards True))
+            Default (Json payload) >>= flip verify
+                [ expectField #amount (.> (Quantity coin))
+                , expectField (#direction . #getApiT) (`shouldBe` Outgoing)
+                ]
```
KtorZ, line 193: This scenario was relying on withdrawals before, now it needs to be explicitly requested yet still pass as normal :+1:

KtorZ:
bors try

lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -261,10 +261,14 @@ spec = do
             , "  --payment PAYMENT        address to send to and amount to send"
             , "                           separated by @, e.g."
             , "                           '<amount>@<address>'"
+            ,"  --withdraw-rewards       Withdraw rewards as change in this"
```
paweljakubas, line 264:
```suggestion
            ,"  --withdraw-rewards       Withdraw rewards as a change in this"
```

lib/cli/test/unit/Cardano/CLISpec.hs
```diff
@@ -274,6 +278,9 @@ spec = do
...
```
paweljakubas, line 281:
```suggestion
            ,"  --withdraw-rewards       Withdraw rewards as a change in this"
```

specifications/api/swagger.yaml
```diff
@@ -1365,6 +1365,16 @@ x-parametersForceNtpCheck: &parametersForceNtpCheck
   schema:
     type: boolean
 
+x-parametersWithdrawal: &parametersWithdrawal
+  in: query
+  name: withdrawRewards
+  allowEmptyValue: true
+  description: |
+    Withdraw rewards as change in this transaction, provided they contribute positively towards the balance (i.e. adding them results in at least as much value as their value after fee calculation).
```
paweljakubas, line 1373:
```suggestion
    Withdraw rewards as a change in this transaction, provided they contribute positively towards the balance (i.e. adding them results in at least as much value as their value after fee calculation).
```

lib/cli/test/unit/Cardano/CLISpec.hs
KtorZ: Hmmm. I disagree here, this is not referring to a single change output, but "change" in general.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1335,6 +1341,13 @@ addressIdArgument :: Parser Text
 addressIdArgument = argumentT $ mempty
     <> metavar "ADDRESS"
 
+-- | [--withdraw-rewards]
+withdrawRewardsFlag :: Parser Bool
+withdrawRewardsFlag = switch $ mempty
+    <> long "withdraw-rewards"
+    <> help "Withdraw rewards as change in this transaction, provided they \
```
paweljakubas, line 1348:
```suggestion
    <> help "Withdraw rewards as a change in this transaction, provided they \
```

paweljakubas (approved):
Looks fine. Would be good to see integration tests with True/False flag side by side at some point.

KtorZ:
> Would be good to see integration tests with True/False flag side by side at some point.

Well, my point of view here is that: we have already hundred of scenarios where the flag is not set (by default), and one scenario that shows what happens when the flag is set :grimacing: ... Nevertheless, I admit that there's no scenario showing what happens when there are rewards AND the flag is not set.

lib/cli/test/unit/Cardano/CLISpec.hs
KtorZ: Checked with our fellow native speaker, and he confirms "change is already in a collective context" :+1:

PR #1864, opened 2020-07-06
Less verbose LSQ pool response trace

Anviking:
# Overview

- [x] Log the full response as DEBUG, not INFO
- [x] Log the _size_ of the stake distribution and reward map as INFO 

# Comments

We now see:
```
[cardano-wallet.network:Info:65] [2020-07-06 15:07:56.63 UTC] Will query non-myopic rewards using the stake 100000000000
[cardano-wallet.network:Info:65] [2020-07-06 15:07:56.82 UTC] Fetched pool data from node tip using LSQ. Got 360 pools in the stake distribution, and 360 pools in the non-myopic member reward map.
```

With:
```
--trace-network=DEBUG
```
we still see the full response.

### DIFF
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
@@ -381,6 +381,9 @@ withNetworkLayer tr np addrInfo versionData action = do
                 (optimumNumberOfPools pparams)
                 rewardMap
                 stakeMap
+        liftIO $ traceWith tr $ MsgFetchedNodePoolLsqDataSummary
+            (Map.size stakeMap)
+            (Map.size rewardMap)
         liftIO $ traceWith tr $ MsgFetchedNodePoolLsqData res
         return res
 
@@ -724,6 +727,9 @@ data NetworkLayerLog
     | MsgDestroyCursor ThreadId
     | MsgWillQueryRewardsForStake W.Coin
     | MsgFetchedNodePoolLsqData NodePoolLsqData
+    | MsgFetchedNodePoolLsqDataSummary Int Int
+        -- ^ Number of pools in stake distribution, and rewards map,
+        -- respectively.
     | MsgWatcherUpdate W.BlockHeader BracketLog
 
 data QueryClientName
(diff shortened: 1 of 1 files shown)


lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -381,6 +381,9 @@ withNetworkLayer tr np addrInfo versionData action = do
                 (optimumNumberOfPools pparams)
                 rewardMap
                 stakeMap
+        liftIO $ traceWith tr $ MsgFetchedNodePoolLsqDataSummary
+            (Map.size stakeMap)
+            (Map.size rewardMap)
         liftIO $ traceWith tr $ MsgFetchedNodePoolLsqData res
```
KtorZ, line 387: That's one way to do it, what about using only a single message, and depending on the severity, show either the full result, or a summary :) ?

Anviking: That was how I first imagined it, but I don't see how.

We need to _provide_ a severity, _given_ a `MsgFetchedNodePoolLsqData`. Not the other way around. We don't know the global severity level.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -724,6 +727,9 @@ data NetworkLayerLog
     | MsgDestroyCursor ThreadId
     | MsgWillQueryRewardsForStake W.Coin
     | MsgFetchedNodePoolLsqData NodePoolLsqData
+    | MsgFetchedNodePoolLsqDataSummary Int Int
```
Anviking, line 730: Wish we had some kind of anonymous records / named-tuples (e.g. `(stakeDistrSize :: Int, rewardMapSize :: Int)`) for situations like these :/

Think a separate record-type is overkill and not sure `Quantity "pools in stakeDistr"` helps _that_ much. Or maybe the latter?

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
KtorZ: Hmmm... right, the `ToText` instance is actually not aware of the severity!

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
KtorZ: > anonymous records / named-tuples

Yeah! One of the thing I really like in Elm <3

KtorZ (approved):

piotr-iohk (approved):

Anviking:
bors r+
should be manually restarted now

Issue #1861, opened 2020-07-06
Include withdrawals one level down, during the coin selection (instead of during fee balancing).


KtorZ:
# Context

In a first iteration, I added withdrawals as a part of the coin selection data type. The withdrawals were taken into account during fee balancing, such that they could be used to pay for fees. Adding them there was easier in a first step as they required very little change to coin selection and fee balancing logic.

# Decision

Yet, being able to use the reward account _during_ the coin selection would be a big plus and would avoid weird cases where the wallet would fail to pay for a transactions despite having enough funds (indeed, in the current state, the wallet needs to be able to cover the entire requested amount with only UTxO and will only use the reward account during fee balancing. This leads to rather misleading API errors. 

# Acceptance Criteria

1. Withdrawals _must_ be used during the coin selection.
2. Withdrawals _must_ still be usable during fee balancing.

---

# Development 

- #1865 

# QA 

- The previous integration tests consuming the rewards still pass.

- The previous unit tests about fee balancing still pass.

- I've added new unit tests showing how the **random** coin selection behaves in the presence  of withdrawals:
    - https://github.com/input-output-hk/cardano-wallet/blob/368c038107e457bfd2ad5298b396ccdd2dfbb422/lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/RandomSpec.hs#L166-L178
    - https://github.com/input-output-hk/cardano-wallet/blob/368c038107e457bfd2ad5298b396ccdd2dfbb422/lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/RandomSpec.hs#L180-L192
    - https://github.com/input-output-hk/cardano-wallet/blob/368c038107e457bfd2ad5298b396ccdd2dfbb422/lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/RandomSpec.hs#L194-L210
    - https://github.com/input-output-hk/cardano-wallet/blob/368c038107e457bfd2ad5298b396ccdd2dfbb422/lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/RandomSpec.hs#L212-L237
    - https://github.com/input-output-hk/cardano-wallet/blob/368c038107e457bfd2ad5298b396ccdd2dfbb422/lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/RandomSpec.hs#L239-L251
[...]

PR #1865, opened 2020-07-06
withdrawals as part of the coin selection

KtorZ:
# Issue Number

#1861 

# Overview

- 12b28f75b69c4bef47e1d1b7f2b765e9d544f800
  :round_pushpin: **take withdrawals into account, one level earlier, during the coin selection**
  Still to be done:

    - Make sure it's correctly done in the largest-first algorithm
    - Add some test scenario that show the influence of the withdrawal

- 547b1301fe7010cc0285a651aa8448cf88b6abc7
  :round_pushpin: **unit-test the newly introduced 'proportionallyTo'**
  Getting this one wrong would be quite bad :s

- 975f72648879a4f9ee513592e322dd10565f5739
  :round_pushpin: **add unit tests showing how withdrawal impacts the random coin selection**

> :bulb: NOTE
>
> I've chosen not to treat the withdrawal as a single _input_, but more as a "money pot" that is proportionally distributed amongst change output, so that it contributes to every output, based on their size. This is to avoid having a small output consuming the entire withdrawal for itself. Note sure if I'll keep the approach in the end, I'll have the night to think about it.

# Comments

- [ ] TODO: take into account the withdrawal when doing largest-first (and testing it)
- [ ] TODO: echo the error message change on "ErrInputsDepleted" to integration tests relying on that message.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -1862,13 +1862,6 @@ instance Buildable e => LiftHandler (ErrCoinSelection e) where
                 , " Lovelace (excluding fee amount) in order to proceed "
                 , " with the payment."
                 ]
-        ErrUtxoNotEnoughFragmented nUtxo nOuts ->
-            apiError err403 UtxoNotEnoughFragmented $ mconcat
-                [ "When creating new transactions, I'm not able to re-use "
-                , "the same UTxO for different outputs. Here, I only have "
-                , showT nUtxo, " available, but there are ", showT nOuts
-                , " outputs."
-                ]
         ErrMaximumInputsReached n ->
             apiError err403 TransactionIsTooBig $ mconcat
                 [ "I had to select ", showT n, " inputs to construct the "
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -1878,10 +1878,9 @@ instance Buildable e => LiftHandler (ErrCoinSelection e) where
                 ]
         ErrInputsDepleted ->
             apiError err403 InputsDepleted $ mconcat
-                [ "I had to select inputs to construct the "
-                , "requested transaction. Unfortunately, one output of the "
-                , "transaction depleted all available inputs. "
-                , "Try sending a smaller amount."
+                [ "I cannot select enough UTxO from your wallet to construct "
+                , "an adequate transaction. Try sending a smaller amount or "
+                , "increasing the number of available UTxO."
                 ]
         ErrInvalidSelection e ->
             apiError err403 InvalidCoinSelection $ pretty e
--- a/lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
@@ -20,6 +20,7 @@ module Cardano.Wallet.Primitive.CoinSelection
     , outputBalance
     , changeBalance
     , feeBalance
+    , proportionallyTo
     , ErrCoinSelection (..)
     , CoinSelectionOptions (..)
     ) where
@@ -30,6 +31,8 @@ import Cardano.Wallet.Primitive.Types
     ( Coin (..), TxIn, TxOut (..) )
 import Data.List
     ( foldl' )
+import Data.Ratio
+    ( Ratio, denominator, numerator )
 import Data.Word
     ( Word64, Word8 )
 import Fmt
@@ -122,6 +125,25 @@ addTxOut total = addCoin total . coin
 
 addCoin :: Integral a => a -> Coin -> a
 addCoin total c = total + (fromIntegral (getCoin c))
+
+-- | Compute the fraction of the first input to match the given ratio, rounded
+-- down.
+--
+-- /invariant:/ The ratio mustn't be greater than 1 without risk to overflow.
+--
+-- >>> 10 `proportionallyTo` 1%1
+-- 10
+--
+-- >>> 10 `proportionallyTo` 1%2
+-- 5
+--
+-- >>> 10 `proportionallyTo` 1%3
+-- 3
+proportionallyTo :: Integral a => a -> Ratio a -> a
+proportionallyTo n r
+    | r > 1 = error "proportionallyTo: ratio is greater than 1!"
+    | otherwise = fromIntegral $
+        toInteger n * toInteger (numerator r) `div` toInteger (denominator r)
 
 data ErrCoinSelection e
     = ErrNotEnoughMoney Word64 Word64
--- a/lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
@@ -20,6 +20,7 @@ module Cardano.Wallet.Primitive.CoinSelection
     , outputBalance
     , changeBalance
     , feeBalance
+    , totalBalance
     , proportionallyTo
     , ErrCoinSelection (..)
     , CoinSelectionOptions (..)
@@ -28,9 +29,11 @@ module Cardano.Wallet.Primitive.CoinSelection
 import Prelude
 
 import Cardano.Wallet.Primitive.Types
-    ( Coin (..), TxIn, TxOut (..) )
+    ( Coin (..), TxIn, TxOut (..), balance' )
 import Data.List
     ( foldl' )
+import Data.Quantity
+    ( Quantity (..) )
 import Data.Ratio
     ( Ratio, denominator, numerator )
 import Data.Word
@@ -119,6 +122,10 @@ changeBalance = foldl' addCoin 0 . change
 
 feeBalance :: CoinSelection -> Word64
 feeBalance sel = inputBalance sel - outputBalance sel - changeBalance sel
+
+-- | Total UTxO balance + withdrawal.
+totalBalance :: Quantity "lovelace" Word64 -> [(TxIn, TxOut)] -> Word64
+totalBalance (Quantity withdraw) inps = balance' inps + withdraw
 
 addTxOut :: Integral a => a -> TxOut -> a
 addTxOut total = addCoin total . coin
--- a/lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
@@ -21,7 +21,6 @@ module Cardano.Wallet.Primitive.CoinSelection
     , changeBalance
     , feeBalance
     , totalBalance
-    , proportionallyTo
     , ErrCoinSelection (..)
     , CoinSelectionOptions (..)
     ) where
@@ -34,8 +33,6 @@ import Data.List
     ( foldl' )
 import Data.Quantity
     ( Quantity (..) )
-import Data.Ratio
-    ( Ratio, denominator, numerator )
 import Data.Word
     ( Word64, Word8 )
 import Fmt
@@ -132,25 +129,6 @@ addTxOut total = addCoin total . coin
 
 addCoin :: Integral a => a -> Coin -> a
 addCoin total c = total + (fromIntegral (getCoin c))
-
--- | Compute the fraction of the first input to match the given ratio, rounded
--- down.
---
--- /invariant:/ The ratio mustn't be greater than 1 without risk to overflow.
---
--- >>> 10 `proportionallyTo` 1%1
--- 10
---
--- >>> 10 `proportionallyTo` 1%2
--- 5
---
--- >>> 10 `proportionallyTo` 1%3
--- 3
-proportionallyTo :: Integral a => a -> Ratio a -> a
-proportionallyTo n r
-    | r > 1 = error "proportionallyTo: ratio is greater than 1!"
-    | otherwise = fromIntegral $
-        toInteger n * toInteger (numerator r) `div` toInteger (denominator r)
 
 data ErrCoinSelection e
     = ErrNotEnoughMoney Word64 Word64
--- a/lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
@@ -135,10 +135,6 @@ data ErrCoinSelection e
     -- ^ UTxO exhausted during input selection
     -- We record the balance of the UTxO as well as the size of the payment
     -- we tried to make.
-    | ErrUtxoNotEnoughFragmented Word64 Word64
-    -- ^ UTxO is not enough fragmented for the number of transaction outputs
-    -- We record the number of UTxO entries as well as the number of the
-    -- outputs of the transaction.
     | ErrMaximumInputsReached Word64
     -- ^ When trying to construct a transaction, the max number of allowed
     -- inputs was reached.
--- a/lib/core/src/Cardano/Wallet/Primitive/CoinSelection/LargestFirst.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/CoinSelection/LargestFirst.hs
@@ -17,9 +17,13 @@ module Cardano.Wallet.Primitive.CoinSelection.LargestFirst (
 import Prelude
 
 import Cardano.Wallet.Primitive.CoinSelection
-    ( CoinSelection (..), CoinSelectionOptions (..), ErrCoinSelection (..) )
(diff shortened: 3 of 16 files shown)


lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1878,10 +1878,9 @@ instance Buildable e => LiftHandler (ErrCoinSelection e) where
                 ]
         ErrInputsDepleted ->
             apiError err403 InputsDepleted $ mconcat
-                [ "I had to select inputs to construct the "
-                , "requested transaction. Unfortunately, one output of the "
-                , "transaction depleted all available inputs. "
-                , "Try sending a smaller amount."
+                [ "I cannot select enough UTxO from your wallet to construct "
+                , "an adequate transaction. Try sending a smaller amount or "
+                , "increasing the number of available UTxO."
```
KtorZ, line 1876: `ErrInputsDepleted` is now used also when trying to select coins on a null UTxO, despite enough balance in the reward account.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Random.hs
```diff
@@ -112,64 +114,89 @@ random
...
     let nOuts = fromIntegral $ NE.length outs
     let maxN = fromIntegral $ maximumNumberOfInputs opt nOuts
+    let totalOut = sum (getCoin . coin <$> outs)
     randomMaybe <- lift $ runMaybeT $
-        foldM makeSelection (maxN, utxo, []) (descending outs)
+        foldM
+            (\acc out ->
+                let
+                    withdraw = totalWithdraw
+                        `proportionallyTo` (getCoin (coin out) % totalOut)
```
KtorZ: So here, we are guaranteed that the ratio given as second argument is always `<= 1`

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Random.hs
```diff
@@ -112,64 +114,89 @@ random
...
+            (maxN, utxo, [])
+            (descending outs)
     case randomMaybe of
         Just (maxN', utxo', res) -> do
             (_, sel, remUtxo) <- lift $
                 foldM improveTxOut (maxN', mempty, utxo') (reverse res)
-            guard sel $> (sel, remUtxo)
+            -- NOTE re-assigning total withdrawal to cope with potential
+            -- rounding issues.
+            guard sel $> (sel { withdrawal = totalWithdraw }, remUtxo)
```
KtorZ: `sel` here is constructed by appending multiple selection together from left to right, using the `Monoid` instance on `CoinSelection`. So in the end, the total withdrawal that we get is the sum of all withdrawals that were assigned to each output, but because `proportionallyTo` rounds **down**, the final sum may end up being slightly less than the total (an error of 1 per output at most) yet, we need it to be _precisely equal_ to the reward's balance without what the Ledger will reject the withdrawal. 

Thus, this is reassigned here, and the possible exceeding will be balanced out during fee balancing anyway.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
```diff
@@ -117,12 +123,35 @@ changeBalance = foldl' addCoin 0 . change
 feeBalance :: CoinSelection -> Word64
 feeBalance sel = inputBalance sel - outputBalance sel - changeBalance sel
 
+-- | Total UTxO balance + withdrawal.
+totalBalance :: Quantity "lovelace" Word64 -> [(TxIn, TxOut)] -> Word64
+totalBalance (Quantity withdraw) inps = balance' inps + withdraw
```
paweljakubas, line 125: :+1:

lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
```diff
@@ -117,12 +123,35 @@ changeBalance = foldl' addCoin 0 . change
...
+--
+-- >>> 10 `proportionallyTo` 1%1
+-- 10
+--
+-- >>> 10 `proportionallyTo` 1%2
+-- 5
+--
+-- >>> 10 `proportionallyTo` 1%3
+-- 3
+proportionallyTo :: Integral a => a -> Ratio a -> a
```
paweljakubas: any chance to have unit test for this or something that uses this inside?

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/LargestFirst.hs
```diff
@@ -59,11 +69,15 @@ largestFirst opt outs utxo = do
             guard s $> (s, UTxO $ Map.fromList utxo')
         Nothing -> do
             let moneyRequested = sum $ (getCoin . coin) <$> (descending outs)
-            let utxoBalance = fromIntegral $ balance utxo
+            let utxoList = Map.toList $ getUTxO utxo
+            let total = totalBalance withdraw utxoList
```
paweljakubas: :+1:

lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
paweljakubas: ok, so it is used with `totalWithdraw` and checked in random...

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Random.hs
paweljakubas: real question, maybe trivial... but why do we need to use `proportionallyTo` at all? why this rule to transfer upon withdrawing proportionally to relative weight in total outs?  Why not to transfer randomly (adding up to withdrawal)? Or almost everything to one plus 1 + 1 + ... + 1? Or evenly? Or create one change?

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Random.hs
paweljakubas: if we transfer withdraw to one change we would not have this problem;-) But, when used `proportionallyTo` indeed we need to take care of this... So what happens in detail when  `proportionallyTo` rounding does consume not all withdraw? To which changes this additional `1`s go ? To those that correspond to outputs that were affected by this rounding, right?

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/RandomSpec.hs
```diff
@@ -151,15 +160,106 @@ spec = do
...
                 , utxoInputs = [oneAda, oneAda, oneAda, oneAda]
                 , txOutputs = 2*oneAda :| [oneAda `div` 2]
+                , totalWithdrawal = 0
                 })
 
+        coinSelectionUnitTest random "withdrawal simple"
+            (Right $ CoinSelectionResult
+                { rsInputs = [1]
+                , rsChange = []
+                , rsOutputs = [2]
```
paweljakubas, line 170: wy it is not `3` here? :thinking:

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/RandomSpec.hs
```diff
@@ -151,15 +160,106 @@ spec = do
...
+            (Right $ CoinSelectionResult
+                { rsInputs = [1,1]
+                , rsChange = []
+                , rsOutputs = [2,2]
+                })
+            (CoinSelectionFixture
+                { maxNumOfInputs = 100
+                , validateSelection = noValidation
+                , utxoInputs = [1,1]
+                , txOutputs = 2 :| [2]
```
paweljakubas, line 190: and here `[3,3]` ?

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/RandomSpec.hs
```diff
@@ -151,15 +160,106 @@ spec = do
                 , validateSelection = noValidation
                 , utxoInputs = [oneAda, oneAda, oneAda, oneAda]
                 , txOutputs = 2*oneAda :| [oneAda `div` 2]
+                , totalWithdrawal = 0
                 })
 
+        coinSelectionUnitTest random "withdrawal simple"
+            (Right $ CoinSelectionResult
+                { rsInputs = [1]
+                , rsChange = []
```
paweljakubas, line 169: why it is not `[1]` ?

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/RandomSpec.hs
```diff
@@ -151,15 +160,106 @@ spec = do
...
+                , validateSelection = noValidation
+                , utxoInputs = [1]
+                , txOutputs = 2 :| []
+                , totalWithdrawal = 1
+                })
+
+        coinSelectionUnitTest random "withdrawal multi-output"
+            (Right $ CoinSelectionResult
+                { rsInputs = [1,1]
+                , rsChange = []
```
paweljakubas, line 183: why it is not `[1,1]` ?

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/RandomSpec.hs
paweljakubas: because of the `proportionallyTo` rounding?

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/RandomSpec.hs
```diff
@@ -151,15 +160,106 @@ spec = do
...
+                })
+            (CoinSelectionFixture
+                { maxNumOfInputs = 100
+                , validateSelection = noValidation
+                , utxoInputs = [1]
+                , txOutputs = 10 :| [10]
+                , totalWithdrawal = 20
+                })
+
+        coinSelectionUnitTest random "withdrawal requires at least one input"
```
paweljakubas, line 253: this one I understand :+1:

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/RandomSpec.hs
paweljakubas: ah, ok. we have input 1 plus withdraw 1 to cover 2

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/RandomSpec.hs
paweljakubas: inps  + withdrawals : `[1,1,2]` to  cover `[2,2]` gives `[]` change

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/RandomSpec.hs
```diff
@@ -151,15 +160,106 @@ spec = do
...
+                })
+
+        coinSelectionUnitTest random "withdrawal not even"
+            -- 10 Ada available as withdrawal
+            --
+            -- - 7 Ada goes to the first output (10/14 * 10 ~= 7)
+            -- - 2 Ada goes to the other output ( 4/14 * 10 ~= 2)
+            (Right $ CoinSelectionResult
+                { rsInputs = [5,5]
+                , rsChange = [2,3]
```
paweljakubas: and here we see that in spite of missing `1` by proportionallyTo the money is counting right

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/RandomSpec.hs
```diff
@@ -151,15 +160,106 @@ spec = do
...
+            (Right $ CoinSelectionResult
+                { rsInputs = [1]
+                , rsChange = [1]
+                , rsOutputs = [10, 10]
+                })
+            (CoinSelectionFixture
+                { maxNumOfInputs = 100
+                , validateSelection = noValidation
+                , utxoInputs = [1]
+                , txOutputs = 10 :| [10]
```
paweljakubas: in old days it would trigger not enough fragmentation if the number of inputs does not match number of outputs. now I see withdrawals help in this sense :+1:  wonder if there are some integration tests checking this edge conditions that needs to be updated

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelectionSpec.hs
```diff
@@ -99,6 +109,16 @@ spec = do
         prop "All inputs are used" prop_allInputsAreUsed
         prop "All inputs are used per transaction" prop_allInputsAreUsedPerTx
         prop "Addresses are recycled fairly" prop_fairAddressesRecycled
+
+    describe "proportionallyTo" $ do
```
paweljakubas: ah, this one I was asking above - persuasive :100:

lib/core/test/unit/Cardano/Wallet/Primitive/FeeSpec.hs
```diff
@@ -707,10 +709,10 @@ genSelectionFor :: NonEmpty TxOut -> Gen CoinSelection
 genSelectionFor outs = do
     let opts = CS.CoinSelectionOptions (const 100) (const $ pure ())
     utxo <- vector (NE.length outs * 3) >>= genUTxO
-    case runIdentity $ runExceptT $ largestFirst opts outs utxo of
+    withdrawal_ <- genWithdrawal
+    case runIdentity $ runExceptT $ largestFirst opts outs (Quantity withdrawal_) utxo of
```
paweljakubas, line 713: Dusting is going to affect this somehow ? Or we are confident it is going to be fine?

lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
paweljakubas: it is checked below

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/RandomSpec.hs
paweljakubas: what about this example : 
```
            (CoinSelectionFixture
                { maxNumOfInputs = 100
                , validateSelection = noValidation
                , utxoInputs = [1]
                , txOutputs = 1 :| [1,1,1,1,1]
                , totalWithdrawal = 5
                })
```
?

lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
KtorZ: Indeed!

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Random.hs
KtorZ: > To which changes this additional 1s go

Most likely, they'll be used to partially cover fee in the fee balancing algorithm here. What will happen in case of rounding issues is that the transaction will be _slightly_ unbalanced (positively) and have a few lovelace extra on the input side assigned to "nothing". Following the fee balancing algorithm, it'll start by consuming that delta, and then, starts to deplete (proportionally again) change outputs. So in the end, the little exceeding isn't lost but goes directly for fee; which is roughly similar to assigning it to any of the change output, since they'll be depleted for fees regardless.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/RandomSpec.hs
KtorZ: These examples are indeed a bit tricky to follow, I tried to left some note above each test case to make it a bit easier.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/RandomSpec.hs
KtorZ: > what about this example

Following the _current_ logic, each output would receive `0` withdrawal bonus (1/6 * 5, rounded down) and therefore, this would fail on the second output because there wouldn't be enough inputs to select. Although indeed, if instead of using the withdrawal proportionally, we were using them fully to cover "just" the right amount, it'd give better result.

lib/core/test/unit/Cardano/Wallet/Primitive/FeeSpec.hs
KtorZ: I am not sure to follow your concern on this one :thinking:  ?

KtorZ:
bors try

paweljakubas (approved):
LGTM!

PR #1874, opened 2020-07-07
test that rewards are smaller for smaller stake

piotr-iohk:
# Issue Number

#1720

# Overview

- 3fbdeed9b06d46619dcdbf685cae98cd16711381
  test that rewards are smaller for smaller stake
  
- e94100736e35a1a98e75d39c66d161690d5a4049
  compare sum of rewards from pools for small and big stake

# Comments

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
@@ -508,7 +508,7 @@ spec = do
             , expectErrorMessage $ errMsg403DelegationFee fee
             ]
 
-    let listPools ctx = request @[ApiStakePool] @IO ctx (Link.listStakePools arbitraryStake) Default Empty
+    let listPools ctx stake = request @[ApiStakePool] @IO ctx (Link.listStakePools stake) Default Empty
     describe "STAKE_POOLS_LIST_01 - List stake pools" $ do
         -- TODO: Add a /short/ eventually (not currently possible) as this test
         -- fails when run in isolation.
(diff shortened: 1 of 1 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -606,14 +606,27 @@ spec = do
...
+            eventually "rewards are smaller for smaller stakes" $ do
+                let stakeSmall = Just (Coin 10_000_000)
+                let stakeBig = Just (Coin 1000_000_000)
+                Right poolsStakeSmall <- snd <$> listPools ctx stakeSmall
+                Right poolsStakeBig <- snd <$> listPools ctx stakeBig
+                let rewards = view (#metrics . #nonMyopicMemberRewards)
+                let rewardsStakeSmall = rewards <$> poolsStakeSmall
+                let rewardsStakeBig = rewards <$> poolsStakeBig
+
+                rewardsStakeSmall `shouldSatisfy` (< rewardsStakeBig)
```
Anviking: I think our custom `.<` will have a yield nicer failure reasons than `shouldSatisfy`

Anviking: You're relying on `Ord [a]`. I'm not sure exactly how that works
```haskell
λ> [3,4] > [3,4]
False
λ> [3,5] > [3,4]
True
λ> [2,5] > [3,4]
False
```
🤔 

Maybe we are looking for two checks
- For one or more pools, higher stake results in _strictly_ higher member rewards
- For all pools, higher stake results in _higher or the same_ member rewards (could imagine them being zero)
?

Or perhaps that is an unnecessarily complex check. Also, the ordering of pools _could_ change between the requests (on rare occasions)

Maybe just comparing the `sum` of the rewards would be more resilient, simple and enough?

Anviking (approved):
Nice!

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
piotr-iohk: okay, I went for comapring sums of rewards. :sweat_smile:

Issue #1838, opened 2020-07-01
Expire transactions when TTL reached


rvl:
# Context

ADP-93 Transaction scheduler

Introduce a time-to-live (TTL) value for transactions, stating that the transaction must be included in the blockchain before a certain slot and simply dropped otherwise.

# Decision

- TTL value is a number of slots, and hardcoded initially.
- The Tx expiry slot is calculated when the transaction is created, and recorded in the DB with pending transactions.
- Add a new TxStatus "Expired", which means that the TTL has been reached.
- In the wallet layer, when applying blocks, check the expiry slot of pending transactions. Any expired transactions are removed from the pending set and marked with Expired state.
- Tx expiry is not something that can be rolled back.

# Acceptance Criteria

- Transaction expiry must be the time when the transaction was created plus the TTL. 
- Transactions must not be still pending after their TTL has been reached.

---

# Development 

As above.

# QA 

1. Integration tests are added in `lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs`
    - TRANS_TTL_01
    - TRANS_TTL_02
    - TRANS_TTL_03
    - TRANS_TTL_04
    - TRANS_TTL_DELETE_01

jonathanknowles:
@rvl 

> Add a new transaction state "Expired", which means that the TTL has been reached.

Just to clarify, is the intention to represent this as an additional constructor on `TxStatus` (and corresponding DB field `txMetaStatus`)?

rvl:
@jonathanknowles Yes, I think so, but after further digging this decision may change.

HeinrichApfelmus:
This is a very old issue which has been implemented.

The design has changed compared to the issue description. The [current design][1] is that

* `Expired` transactions can be rolled back, because they can actually still be included in the blockchain after a rollback.
* The pending transactions are kept in a separate database table from the transactions that are in ledger.

[1]: https://github.com/cardano-foundation/cardano-wallet/blob/4d1575023d29be39cfb7867f010c6c6f0de61243/specifications/Cardano/Wallet/Submissions/design.md

PR #1879, opened 2020-07-08
Add transaction expiry slot for pending transactions

rvl:
### Issue Number

Relates to ADP-93 / #1838.

### Overview

- Adds expiry slot to pending transactions in order to implement transaction TTL.
- Expiry slot and time is reported in the transaction history.
- At the expiry slot, pending transactions are marked Expired.
- Default TTL of 7200 slots is still hard coded.


lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -916,6 +919,25 @@ instance FromText TxStatus where
...
+data PendingTx = PendingTx
+    { pendingTx :: !Tx
+    -- ^ The transaction.
+    , expiry :: !SlotId
+    -- ^ The future slot at which the transaction will expire.
+    }
+    deriving (Generic, Eq, Show)
+
+instance Ord PendingTx where
+    compare (PendingTx a _) (PendingTx b _) = compare a b
```
jonathanknowles:
```suggestion
    compare = comparing pendingTx
```

(Where `comparing` comes from `Data.Ord`.)

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -983,6 +1005,12 @@ fromTransactionInfo info = Tx
     , outputs = txInfoOutputs info
     }
 
+fromTransactionInfoPending :: TransactionInfo -> PendingTx
+fromTransactionInfoPending info = PendingTx
```
jonathanknowles: Perhaps the `xFromY` naming scheme would be more understandable here? So:

```suggestion
pendingTxFromTransactionInfo :: TransactionInfo -> PendingTx
pendingTxFromTransactionInfo info = PendingTx
```

lib/core/src/Cardano/Wallet/Primitive/Types.hs
rvl: Thanks - good idea.

rvl:
bors try

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -593,6 +594,36 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
...
+
+        verify r
+            [ expectSuccess
+            , expectResponseCode HTTP.status202
+            , expectField (#direction . #getApiT) (`shouldBe` Outgoing)
+            , expectField (#status . #getApiT) (`shouldBe` Pending)
+            , expectField #expiresAt (`shouldSatisfy` isJust)
+            ]
+
+        -- This stuff would be easier with Control.Lens...
```
KtorZ, line 613: Are we tracking that in a technical debt ticket :thinking: ?

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -593,6 +594,36 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
...
+
+        -- Get insertion slot and out of response.
+        let (_, Right apiTx) = r
+        let (Just (ApiTimeReference _ sinceBlock)) = apiTx ^. #pendingSince
+        let sl = sinceBlock ^. #absoluteSlotNumber . #getApiT
+
+        -- The expected expiry slot (adds the hardcoded default ttl)
+        let ttl = sl + SlotNo 7200
```
KtorZ: Should we perhaps expose that default "magic number" ? Or at least, have it behind a named constant in the integration DSL so that we do not duplicate it all over the place.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -840,6 +840,7 @@ restoreBlocks ctx wid blocks nodeTip = db & \DBLayer{..} -> mapExceptT atomicall
     let k = gp ^. #getEpochStability
     let localTip = currentTip $ NE.last cps
 
+    updatePendingTx (PrimaryKey wid) (view #slotNo localTip)
```
KtorZ: `update` feels a bit general here. When reading this, my first thought is "update what?". What about using a more specific qualifier like: "discardNowExpiredTx" or, whatever you'd come up with ?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1332,9 +1333,11 @@ mkApiTransactionFromInfo ti (TransactionInfo txid ins outs ws meta depth txtime
         case meta ^. #status of
             Pending  -> #pendingSince
             InLedger -> #insertedAt
+            Expired  -> #pendingSince
```
KtorZ, line 1339: Shall we perhaps use yet-another-field for tracking the time since which the transaction has expired?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -663,6 +665,11 @@ data ApiBlockReference = ApiBlockReference
     , absoluteSlotNumber :: !(ApiT SlotNo)
     } deriving (Eq, Generic, Show)
 
+data ApiSlotReference = ApiSlotReference
+    { time :: !UTCTime
+    , absoluteSlotNumber :: !(ApiT SlotNo)
+    } deriving (Eq, Generic, Show)
```
KtorZ, line 679: What about including also the epoch and slot within epoch to be consistent with other time references? We use:

```hs
data ApiTimeReference = ApiTimeReference
    { time :: !UTCTime
    , block :: !ApiBlockReference

data ApiBlockReference = ApiBlockReference
    { epochNumber :: !(ApiT EpochNo)
    , slotNumber :: !(ApiT SlotInEpoch)
    , height :: !(Quantity "block" Natural)
    , absoluteSlotNumber :: !(ApiT SlotNo)
```

Although indeed, the TTL is about slots and not blocks so the _height_ don't make much sense. The other however do. So I'd suggest to have something along the lines of:

```hs
data ApiTimeReferenceBySlot = ApiTimeReferenceBySlot
    { time :: !UTCTime
    , slot :: !ApiSlotReference

data ApiSlotReference = ApiBlockReference
    { epochNumber :: !(ApiT EpochNo)
    , slotNumber :: !(ApiT SlotInEpoch)
    , absoluteSlotNumber :: !(ApiT SlotNo)
```

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
rvl: OK done.

lib/core/src/Cardano/Wallet.hs
rvl: `updatePendingTxForExpiry`

lib/core/src/Cardano/Wallet/Api/Server.hs
rvl: I think "pending_since" is the correct field. The time at which the transaction expired is in "expires_at".
So both times are captured for expired transactions.

lib/core/src/Cardano/Wallet/Api/Types.hs
rvl: Not only is TTL about slots, but it's about slots in the _future_ -- which cannot have a block height.

OK I have added epoch and slot numbers.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
rvl: No, because I sent a message about this to slack, and got exactly 0 replies.

lib/core/src/Cardano/Wallet/Api/Server.hs
KtorZ: Fair enough!

lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -522,11 +526,20 @@ walletId =
...
+-- Constants
 --
+
+-- | Min UTxO parameter for the test cluster.
 minUTxOValue :: Natural
 minUTxOValue = 1_000_000
 
+-- | Wallet server's chosen transaction TTL value (in slots) when none is given.
+defaultTxTTL :: SlotNo
+defaultTxTTL = 7200
```
KtorZ, line 538: :+1:

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: For consistency, It'd be better to have them nested under a `slot` field, don't you think :thinking: ?

lib/core/src/Cardano/Wallet/DB/MVar.hs
```diff
@@ -174,6 +175,9 @@ newDBLayer timeInterpreter = do
                                        Pending Tx
         -----------------------------------------------------------------------}
 
+        , updatePendingTxForExpiry = \pk tip -> ExceptT $ do
+            alterDB errNoSuchWallet db (mUpdatePendingTx pk tip)
```
KtorZ, line 179: :+1:

lib/core/src/Cardano/Wallet/Api/Types.hs
rvl: Yeah maybe - I am going to straighten out the types a bit (without changing existing API fields).

rvl:
@KtorZ I made the API types more consistent (last 2 commits).

ApiBlockReference is just ApiSlotReference with block information added.

ApiNetworkTip is replaced with ApiSlotReference.

The end result is that all references to a slot/time look the same in the API.

The Haskell API types have SlotId and block info nested rather than inline. I hope in future that we can change the swagger spec to match. But for now there are some non-generic Aeson instances to maintain backwards compatibility.

KtorZ (approved):
Nice work!

KtorZ:
> The Haskell API types have SlotId and block info nested rather than inline. I hope in future that we can change the swagger spec to match. But for now there are some non-generic Aeson instances to maintain backwards compatibility.

I pointed that out in my review from yesterday which I apparently didn't submit :| ... I agree that having one flat and the other nested is sort of inconsistent and I'd very much prefer getting rid of this inconsistency but its also neat-picking at this stage. So happy to move forward with that one.

KtorZ:
bors merge

Issue #1883, opened 2020-07-08
Allow configuring an additional pool metadata registry when starting wallet server


KtorZ:
# Context

The wallet currently fetches pool metadata from all pool remote servers. Not only this present some security _risks_ for cardano-wallet (denial of service, although the wallet is being quite defensive about things here), but it is also inefficient since every wallets have to fetch many pool metadata servers (which can be arbitrarily located on the planet). 

Beside, IOHK has developed a pool metadata aggregation server that has been deployed. The server can be used as a proxy for fetching pool metadata. 

# Decision

- Revise the logic fetching metadata to allow fetching them from a different URL, falling back to the URL specified on-chain when the connection with the proxy fails.

- Allow configuring a proxy via a new command-line optional option `[--pool-metadata-proxy URI]`; when provided, this should become the default URL from which metadata are fetched. When not set, the wallet should fallback to the original metadata URL defined in the certificate.

# Acceptance Criteria

1. There _must_ be a new option in the `serve` command for defining a metadata proxy
2. When set, the wallet _must_ try fetching from this URL first. 
3. Upon **HTTP Exception**, the wallet _may_ fallback to the original URL specified in the certificate. 
4. There _should_ be sufficient logging to observe what is happening. 

Some precision about 3. => The registry proxy also supports pools' blacklisting. So it is possible that the server would reply, but not positively, in which case, the metadata should NOT be retried. The wallet may only retry when it fails to establish the connection with the proxy (because down or anything). 

---

# Development 

- Add a command-line option to allow specifying a metadata proxy. See #1884 

# QA

KtorZ:
:+1: tested manually against mainnet and latest deployed verison of SMASH.

PR #1884, opened 2020-07-08
SMASH Integration

KtorZ:
# Issue Number

#1883 

# Overview

- d790880de0392f22e8d2d2d4096e1d44c3e60616
  :round_pushpin: **allow different ways of building metadata URL from metadata information**
  The idea is the following:

- We have a list of "URL builders" that will be tried in order, but only
  if the first builder replied with a connection failure (so, only if the
  server is somewhat down)

- We can always have our current behavior by using the
  'identityUrlBuilder'.

- Then, we can augment this behavior if given a specific registry URL
  via the command-line options, which will add a "registryUrlBuilder"
  which takes predecence over the identity builder.

- 48884cd3a0c436b0e4b4e817e8df2c7319440374
  :round_pushpin: **try fetching metadata from the next builder on HTTP exception**
  
- c133609672d5b1173a00a20595bc00ac9054ee10
  :round_pushpin: **add command-line option for enabling a pool metadata proxy**
  Going for a command-line option so that this can easily be turned on and off at will in case of any issue. Plus, not everyone might agree with the idea of going through a registry so hard-coding the registry URL is a no go. cardano-wallet-jormungandr uses an ad-hoc environment variable for this, but a documented option in the command-line is definitely cleaner.

- e9dbf56e160b869af3b677e82449e56b8b8a37af
  :round_pushpin: **revise logging around pool metadata fetching to have more insight _inside_ the fetcher**
  So that we can clearly log fallback attempts and keep it clean.

- cf00db653779cf80a579025e8df8a9576c14b46a
  
- 08305897348945dd16afd484d2b500fd2b298e52
  :round_pushpin: **use 'Either HttpException' as a base monad instead of 'ExceptT'**
    The underlying URL builder relies on the 'MonadThrow' class to send errors. Yet, when using 'ExceptT', the error is thrown _in the base monad_!!
  So, using 'Either' instead in conjunction with `except` to catch the error as expected in the loopl, instead of throwing hard in the parent context.

- dd00d12e38defb608db5565f181c0052c3713e44
  :round_pushpin: **be more specific when interpreting responses from the aggregation server**
  
- 28b9d575c23413047b73819aa60afb3bb65d6427
  :round_pushpin: **remove fallback when a pool metadata proxy is supplied.**
  The fallback made sense in the early days, but we can now make it a proper switch: it's either fetching from pools, or it's fetching from the registry

# Comments

- [ ] :warning: TODO: show that it works. So far, on the shelley testnet:

Some quick manual / sanity testing:

```console
$ cardano-wallet-shelley serve \
  --node-socket node.socket \
  --testnet shelley_testnet-genesis.json \
  --pool-metadata-proxy https://smash.shelley-testnet.dev.cardano.org/api/v1/metadata

[...]

[2020-07-08 19:57:09.29 UTC] Fetching metadata with hash c9ed7096 from https://smash.shelley-testnet.dev.cardano.orgapi/v1/metadata/c9ed70969c964284ff854d61e43c2591b67ceb7e978347cb1b50ba3ed1d3edb4
[2020-07-08 19:57:09.45 UTC] Successfully fetched metadata with hash c9ed7096: StakePoolMetadata {ticker = StakePoolTicker {unStakePoolTicker = "IOHK1"}, name = "IOHK1", description = Just "IOHK Pool", homepage = "https://iohk.io"}
```

```console
$ cardano-wallet-shelley serve \
  --node-socket node.socket \
  --testnet shelley_testnet-genesis.json \
  --pool-metadata-proxy patate
option --pool-metadata-proxy: Invalid URI. Make sure the URI is well formed, with a protocol and a host.
```


lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -70,45 +104,129 @@ defaultManagerSettings =
...
+            -- - A name (at most 50 UTF-8 bytes)
+            -- - An optional description (at most 250 UTF-8 bytes)
+            -- - A ticker (at most 5 UTF-8 bytes)
+            -- - A homepage (at most 100 UTF-8 bytes)
+            --
+            -- So, the total, including a pretty JSON encoding with newlines ought
+            -- to be less than 512 bytes. For security reasons, we only download the
+            -- first 512 bytes.
+            if responseStatus res /= status200 then
+                return $ Left "Server did reply, but not positively. This metadata may be blacklisted."
```
piotr-iohk: Does SMASH respond with specific code when metadata is "blacklisted"? If so maybe it could be specified for this particular case.

Currently the solution falls back only if it SMASH url cannot be reached, but for example if provided SMASH url returns `404` it could be perfectly falling back as well...

KtorZ: The blacklisting isn't yet implemented in SMASH but, when it does it'll probably return a `404` (according to the specification). There really are two cases:

a) The metadata aren't yet available on SMASH because it hasn't yet downloaded the metadata from a pool (that could be the case for recent pools). Then, it's okay, we'll get a `404` and will retry the url _later_ (exponential backoff), up to several days after the first try and up to 10 times.

b) The metadata is blacklisted, in which case SMASH should also reply with a 404 (very much as if the metadata was absent), and here, we'll retry all-the-same because we can't actually tell the difference with the first case. After 10 retries, we'll stop and consider this metadata "unfetchable". 

When blacklisting is implemented, we might revise this behavior and "acknowledge" blacklisted pools if possible so that we can skip the 10 retries.

piotr-iohk: Then maybe we can be more specific than `/= status200`, and just implement this behaviour on `== status404`? I can imagine that in case of mainetenace, the server might for example return `502` or something else in which case we could just fall back.

piotr-iohk: Also as far as I see SMASH actually responds with 200 and a message, in case of data missing:

```
{"code":"DbLookupTxMetadataHash","description":"The hash 534b6a6bd10d23a2eb060a51d61df263bf2fa3f10b6f5a701fd9a77d6c81e67011111111111111111111 is missing from the DB."}
```

KtorZ: > Also as far as I see SMASH actually responds with 200 and a message, in case of data missing:

:facepalm: really?

KtorZ: https://github.com/input-output-hk/smash/issues/16

piotr-iohk:
Although fallback is removed I see there is a fallback going on in case of _SMASH with expired SSL_:

```
$ cardano-wallet-shelley serve \
  --node-socket node.socket \
  --testnet shelley_testnet-genesis.json \
  --pool-metadata-proxy https://expired.badssl.com/
```

```
[cardano-wallet.pools-engine:Warning:52] [2020-07-22 13:23:37.43 UTC] Couldn't reach server at https://expired.badssl.com4f3b3be782e93761220942c55cbc13358b8bcd12bf200f1548e4daa5e99bdfcd. Falling back using a different strategy.
[cardano-wallet.pools-engine:Info:52] [2020-07-22 13:23:37.43 UTC] Fetching metadata with hash 4f3b3be7 from https://cent.stakepoolcentral.com/resources/CENT.json

```

piotr-iohk:
There is minor issue in the log, missing `/` in the url (`orgapi`):

```
Fetching metadata with hash f0022651 from 
https://smash.shelley-testnet.dev.cardano.orgapi/v1/metadata/f0022651a06ba5b62bc6c9e223ee15aa957867cb9ba9b3ca4f4f4a9ff4f47964
```

KtorZ:
Fixed the URL in the log (was missing a trailing slash in the `URI` path representation, but oddly enough, the underlying library building the request from the URL still worked just fine :man_shrugging:). 

Also, I've double-checked about the fallback and I do observe the correct behavior:

##### --pool-metadata-proxy https://smash.shelley-testnet.dev.cardano.org/api/v1/metadata

```
[cardano-wallet.pools-engine:Info:53] [2020-07-23 06:42:36.15 UTC] Fetching metadata with hash 2633c85f from https://smash.shelley-testnet.dev.cardano.org/api/v1/metadata/2633c85fdcd1c1193b92c785facdc9b849e5ab0cd70833faa383157103f40fc4
```

##### --pool-metadata-proxy  https://expired.badssl.com/

```
[cardano-wallet.pools-engine:Info:64] [2020-07-23 06:43:44.44 UTC] Fetching metadata with hash 05eb1bd0 from https://expired.badssl.com/05eb1bd033853b0f90625fed274578605a0f00fd19be320f2b5f81c04498ec2f
[cardano-wallet.pools-engine:Warning:64] [2020-07-23 06:43:44.85 UTC] Couldn't reach server at https://expired.badssl.com/05eb1bd033853b0f90625fed274578605a0f00fd19be320f2b5f81c04498ec2f.
[cardano-wallet.pools-engine:Warning:64] [2020-07-23 06:43:44.85 UTC] Failed to fetch metadata with hash 05eb1bd0: Metadata server(s) didn't reply in a timely manner.
[cardano-wallet.pools-engine:Info:64] [2020-07-23 06:43:44.85 UTC] Fetching metadata with hash fb4086de from https://expired.badssl.com/fb4086de12dd20e71817ad12220fe59b500488445e7278614eef9fb87fa7743f
[cardano-wallet.pools-engine:Warning:64] [2020-07-23 06:43:45.23 UTC] Couldn't reach server at https://expired.badssl.com/fb4086de12dd20e71817ad12220fe59b500488445e7278614eef9fb87fa7743f.
[cardano-wallet.pools-engine:Warning:64] [2020-07-23 06:43:45.23 UTC] Failed to fetch metadata with hash fb4086de: Metadata server(s) didn't reply in a timely manner.
[cardano-wallet.pools-engine:Info:64] [2020-07-23 06:43:45.23 UTC] Fetching metadata with hash 9d355af4 from https://expired.badssl.com/9d355af4aa94a75e1c095465e778142d920cb44976ab527faffe9543818c3948
[cardano-wallet.pools-engine:Warning:64] [2020-07-23 06:43:45.57 UTC] Couldn't reach server at https://expired.badssl.com/9d355af4aa94a75e1c095465e778142d920cb44976ab527faffe9543818c3948.
```

lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -70,45 +102,133 @@ defaultManagerSettings =
...
+    :: URI
+    -> StakePoolMetadataUrl
+    -> StakePoolMetadataHash
+    -> Either HttpException URI
+registryUrlBuilder baseUrl _ (StakePoolMetadataHash bytes) =
+    Right $ baseUrl
+        { uriPath = "/" <> intercalate "/" (pathSegments baseUrl ++ [hash])
+        }
+  where
+    hash = T.unpack $ T.decodeUtf8 $ convertToBase Base16 bytes
```
rvl, line 126:
```suggestion
    hash = B8.unpack $ convertToBase Base16 bytes
```

lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -70,45 +102,133 @@ defaultManagerSettings =
...
+        Left msg ->
+            Nothing <$ traceWith tr (MsgFetchPoolMetadataFailure hash msg)
+
+        Right meta ->
+            Just meta <$ traceWith tr (MsgFetchPoolMetadataSuccess hash meta)
+
+    -- Try each builder in order, but only if the previous builder led to a
+    -- "ConnectionFailure" exception. Other exceptions mean that we could
+    -- likely reach the server.
+    fromFirst (builder:rest) = do
```
rvl: I found this a little confusing. Can we separate the function which makes a request and reads at most 512 bytes from the response from the function which tries each builder and uses the first which succeeds?

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1296,6 +1299,21 @@ tlsOption = TlsConfiguration
         <> metavar "FILE"
         <> help "The RSA Server key which signed the x.509 server certificate."
 
+poolMetadataProxyOption
+    :: Parser URI
+poolMetadataProxyOption = option (eitherReader reader) $ mempty
+    <> long "pool-metadata-proxy"
```
rvl: I would prefer to not use the word "proxy" where it could be confused with actual HTTP proxies.
We could just say `--smash-server`.
And what about making the CF smash server the default, with an option to disable the metadata aggregator?

rvl (approved):

lib/core/src/Cardano/Pool/Metadata.hs
KtorZ: I can try that :+1:

lib/cli/src/Cardano/CLI.hs
KtorZ: > I would prefer to not use the word "proxy" where it could be confused with actual HTTP proxies.

Fair point. Although I dislike the "--smash-server" naming for this because in principle, smash is one implementation of a metadata registry. There could be others by the community that aren't smash. Hence why I wanted to keep the name rather "generic". Perhaps `--pool-metadata-registry` ?

>  And what about making the CF smash server the default, with an option to disable the metadata aggregator?

There's no CF server yet as far as I know :grimacing: ... plus, with all the networks and candidates at the moment, it's sort of hard to know what a sensible default would be (other than, no default). Long-term, I do however agree, once the dust settle down a bit.

piotr-iohk:
> Also, I've double-checked about the fallback and I do observe the correct behavior:

Indeed, I must have not updated the branch, sorry.

Ok, small thing, but maybe we would like to consider allowing only `https` as a protocol in the URL that is provided to --pool-metadata-proxy? Currently the validation of URL is very simple and in particular you can:

```
--pool-metadata-proxy protocol:host
```
which results in:
```
[cardano-wallet.pools-engine:Error:49] [2020-07-23 08:45:29.21 UTC] Worker has exited unexpectedly: InvalidUrlException "protocol:/host/65e023d50434d576f6d8f901a10adb8bb55f2550ee08fe017e10c7911d4f97f2" "Invalid scheme"
```

Just a thought.

piotr-iohk (approved):

lib/cli/src/Cardano/CLI.hs
rvl: I thought "smash" _was_ a generic name, or at least something that implements the smash API.
I have asked for clarification input-output-hk/smash#22.

KtorZ: Anyway, I'll for "--smash-server" since it's more transparent for now. We can always change later.

lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -70,45 +102,138 @@ defaultManagerSettings =
...
-    StakePoolMetadataHash hash = hash_
+    runExceptTLog
+        :: ExceptT String IO StakePoolMetadata
+        -> IO (Maybe StakePoolMetadata)
+    runExceptTLog action = runExceptT action >>= \case
+        Left msg ->
+            Nothing <$ traceWith tr (MsgFetchPoolMetadataFailure hash msg)
+
+        Right meta ->
+            Just meta <$ traceWith tr (MsgFetchPoolMetadataSuccess hash meta)
```
rvl, line 153: Previously I have combined Success/Failure log messages into a single message type, e.g.
```
    | MsgFetchPoolMetadataResult StakePoolMetadataHash (Either String StakePoolMetadata)
```
This way is no less expressive for logging, and means you can use a single `traceWith` statement and avoid needing to use constructs like `runExceptTLog`.

rvl (approved):

rvl:
> Ok, small thing, but maybe we would like to consider allowing only https as a protocol in the URL

I added this @piotr-iohk so that the error can be caught earlier.

rvl:
Forgot a cabal dep, oops.

bors r+

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1296,6 +1299,26 @@ tlsOption = TlsConfiguration
...
+    reader = maybe (Left err) Right . (parseAbsoluteURI >=> isHttp)
+
+    isHttp uri
+        | uriScheme uri `elem` ["http:", "https:"] = Just uri
+        | otherwise = Nothing
+
+    err :: String
+    err =
+        "Invalid URI. Make sure it is a well-formed \
+        \ absolute http or https URI."
```
KtorZ, line 1320: :+1:

lib/shelley/test/integration/Main.hs
```diff
@@ -165,9 +165,7 @@ testPoolConfigs =
     [ -- This pool should never retire:
       PoolConfig {retirementEpoch = Nothing}
       -- This pool should retire almost immediately:
-    , PoolConfig {retirementEpoch = Just 3}
-      -- This pool should retire, but not within the duration of a test run:
-    , PoolConfig {retirementEpoch = Just 100_000}
+    , PoolConfig {retirementEpoch = Just 1_000}
```
KtorZ: Arf! I think there has been a wrong rebase / conflict resolution here! Hence the integration failure.

rvl: Oops

Issue #1868, opened 2020-07-07
Properly support BlockHeader <-> HF.Point conversions


Anviking:
# Context

We can no longer (correctly) convert `SlotNo` to `SlotId` using the genesis epoch length when
the hard fork arrives. Shelley may use a different epoch length.

To interact with the node we need to provide points which are `(slotNo, headerHash)`.

Rollbacks to EBBs are not a concern as they have already ceased to be produced.

# Decision

- Replace SlotId with SlotNo in DB
- Concider using a special value for genesis (like `NULL`)

# Acceptance Criteria

- Wallet _must_ not break when the hard fork arrives
- The DB/persisted slots _should_ be easy to reason about for developers (i.e. not just work by accident) 
- Don't break cardano-wallet-jormungandr

---

# Development 

- #1887 
- #1890 
- #1901 

# QA 

- Existing tests pass

> Wallet _must_ not break when the hard fork arrives

Bad AC, but we don't need to convert between `SlotNo` and `SlotId` anymore to talk to the node.

- Here are the new golden tests for sync progress calculation: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Cardano/Wallet/Primitive/SyncProgressSpec.hs#L125-L135

Anviking:
I think it would be neat to even use the _very same_ `SlotNo` definition as other projects https://github.com/input-output-hk/cardano-base/blob/5c6db42be8ed83463f89e6257d8eb94ca4df8ca9/slotting/src/Cardano/Slotting/Slot.hs#L31-L35

piotr-iohk:
lgtm

PR #1887, opened 2020-07-08
Rename our SlotNo to SlotInEpoch

Anviking:
# Issue Number

#1868

# Overview

- [x] Renamed `SlotNo` to `SlotInEpoch` (for consistency with other components)

# Comments

- `W.SlotNo` == `Cardano.SlotInEpoch` /= `Cardano.SlotNo`

```
Epoch       0   1   2   3   4
SlotNo*     0 1 2 3 4 5 6 7 8 9
SlotInEpoch 0 1 0 1 0 1 0 1 0 1
```

*) I.e. the one defined in cardano-base and used in ourobouros-consensus.

### DIFF
--- a/lib/byron/src/Cardano/Wallet/Byron/Compatibility.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron/Compatibility.hs
@@ -39,7 +39,7 @@ module Cardano.Wallet.Byron.Compatibility
     , toByronHash
     , toGenTx
     , toPoint
-    , toSlotNo
+    , toSlotInEpoch
 
     , fromBlockNo
     , fromByronBlock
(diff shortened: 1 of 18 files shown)


rvl (approved):
Good idea to match the names in ouroboros-consensus.

It would be nice if there were a small library of slotting types that we could import.

erikd:
> It would be nice if there were a small library of slotting types that we could import.

https://github.com/input-output-hk/cardano-base/tree/master/slotting maybe?

Anviking:
> https://github.com/input-output-hk/cardano-base/tree/master/slotting maybe?

Let's not get ahead of ourselves 😄 but I'm hoping we can switch to that soon.

Think we might end up with a few orphans in the wallet (e.g. for a `Buildable` instance), or the need to resolve that somehow.

Anviking:
bors r+
(master has changed, should not be cached anymore I think)

Issue #1869, opened 2020-07-07
Support SlotNo <-> UTCTime conversions


Anviking:
# Context

⚠️ **Ticket may be partly wrong. In the Shelley phase of Byron;Shelley, we have unlimited forecast range** ⚠️ 

We currently have naive point -> time conversions that are going to break with the hard fork. We:
- Show the time transactions were inserted to the ledger
- Show the time of the next epoch
- Show the time of future changes to the delegation status

To help us, there will be a query from consensus that will enable us to do time conversions locally. We will
/not/ have to send one request per conversion. But we will have to update the conversion-info 
from time to time. (I would guess some time per epoch.)

> For the wallet the most annoying thing will be that you'll need to update this from time to
> time, so you'd need some kind of stateful abstraction that if a conversion is out of range,
> asks for an updated interpreter, tries again; if still out of date, throw an exception (this
> would then be a bug in the wallet, trying to look ahead too far).

All answers are certain and not subject to rollback.

> You get one of two answers: "Yes, I can convert this thing, here's the result [and it is not
> subject to rollback]" or ‚"No, we don't know that yet".

### Times of future epochs

We can convert historial times. There is a forecast range. We can know the time of the next epoch 
start, but not anything beyond it.

So this will still work:
<img width="1016" alt="Skärmavbild 2020-07-07 kl  11 19 11" src="https://user-images.githubusercontent.com/304423/86768979-aff0d700-c04e-11ea-880e-2e76a3d57389.png">

But we can no longer provide `epochStartTime` for changes to the delegation:

```haskell
data ApiEpochInfo = ApiEpochInfo
    { epochNumber :: !(ApiT EpochNo)
    , epochStartTime :: !UTCTime
    } deriving (Eq, Generic, Show)
```

Luckily it seems Daedalus doesn't use it:

<img width="226" alt="Skärmavbild 2020-07-07 kl  11 19 15" src="https://user-images.githubusercontent.com/304423/86769056-ce56d280-c04e-11ea-9be5-e3c8885eec5e.png">

### Sync progress

We can no longer convert the current wall clock to a slot while syncing. We are currently relying on this for the sync progress calculation. Doing the opposite however: converting the genesis and wallet slots to times, and comparing with the current time, should work. But perhaps not as well on jormungandr where there may be a large gap of empty slots between block0 and block1.

# Decision

- Abstraction for `point <-> UTCTime` conversions for txs _must_ be added #1901 
[...]

Issue #1870, opened 2020-07-07
Actually integrate with HFC (byron-shelley) node


Anviking:
# Context

User story: ADP-356

Creating this task for the actual integration with the HFC node, where as other tasks in the milestone are more preparatory.

### Questions

1. Q: We probably need to switch dynamically between the byron and shelley transaction layer.
    A: This is no longer the case - we don't need to support single-era mode any more.

2. Q: Can there be race conditions where tip is in shelley and wallet is in byron when we are creating txs?
    A: It does not matter how far along the wallet is synced, as long as its UTxO are spendable. For submitting txs, only the latest state of the network tip is relevant.

# Decision

1. [x] Update node libraries to latest versions (dcoutts/update-deps branch) ⇒ #1909

2. [x] Update cardano-node to 1.15.1 ⇒ #1910 or #1909 

3. [x] Move to using the new `Cardano.Api.Typed` module in `NetworkLayer` ⇒ #1909

4. [x] Move to using the new `Cardano.Api.Typed` module in `TransactionLayer`.

5. [x] Change network layer to use `CardanoBlock` instead of `ShelleyBlock` ⇒ #1909.
    - LocalChainSync - network tip follower
    - LocalChainSync - wallet block fetcher
    - LocalStateQuery - protocol parameters
    - LocalStateQuery - stake pools
    - LocalTxSubmission

6. [/] ~Adjust tx parameters query in network layer to include era information. This will be used for the TransactionLayer switch.~

7. [x] Add a network layer function to get latest Interpreter ⇒ #1909.

8. [/] ~Functions merged in input-output-hk/ouroboros-network#2399 allow for a single LocalStateQuery to be used for both shelley and HFC nodes. Attempt to support both shelley_testnet (shelley only) and mainnet_candidate (HFC) with this wrapper.~

9. [x] Integration test - launch a separate cluster for this test case. This modified cluster will do a hard fork. ⇒ #1924

10. [x] Adjust `cardano-wallet-shelley serve` CLI to support dual-era mode.

11. [ ] More details are yet to be discovered.

# Acceptance Criteria

1. Wallets must sync across hard fork boundary.
2. Transactions must be able to be submitted in ~both byron and~ shelley eras.

---

# Development 

(As above)

# QA 

1. Shelley integration tests are modified to do a hard fork before starting tests (`lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs`).
[...]

PR #1890, opened 2020-07-09
Move slotting to separate module

Anviking:
# Issue Number

ADP-356 #1868, #1869 

# Overview

- [x] I have moved the existing slotting logic (`flatSlot`, `slotAt`, etc.) to a new module `Cardano.Wallet.Primitive.Slotting`
    - It is easier to keep track of remaining uses of static slotting this way
- [x] I started adding a ouroboros-consensus wrapper on top (currently unused)
    - [x] Added property `runQuery epochOf singleEraInterpreter == epochNumber . fromFlatSlot`

# Comments

- I left tests for slotting arithmetic in TypesSpec because the boundary between `Range` tests and slotting tests seemed unclear.
- Next step: Wrap all existing slotting logic using `Qry`/`Interpreter`.
- Could split into two PRs if you want

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
@@ -11,8 +11,13 @@
 -- @UTCTime@.
 
 module Cardano.Wallet.Primitive.Slotting
-    ( -- * Legacy functions
-      unsafeEpochNo
+    ( -- * New api using ouroboros-concensus
+      epochOf
+    , singleEraInterpreter
+    , runQuery
+
+      -- * Old functions
+    , unsafeEpochNo
     , epochStartTime
     , epochPred
     , epochSucc
@@ -62,6 +67,7 @@ import Numeric.Natural
 import Ouroboros.Consensus.HardFork.History.EraParams
     ( EraParams (..), noLowerBoundSafeZone )
 import Ouroboros.Consensus.HardFork.History.Qry
+    ( Qry, runQuery, slotToEpoch )
 import Ouroboros.Consensus.HardFork.History.Summary
     ( Summary (..), neverForksSummary )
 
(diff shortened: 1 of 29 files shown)


lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
```diff
@@ -0,0 +1,261 @@
...
+    ( NominalDiffTime, UTCTime, addUTCTime, diffUTCTime )
+import Data.Word
+    ( Word64 )
+import GHC.Generics
+    ( Generic )
+import Numeric.Natural
+    ( Natural )
+import Ouroboros.Consensus.HardFork.History.EraParams
+    ( EraParams (..), noLowerBoundSafeZone )
+import Ouroboros.Consensus.HardFork.History.Qry
```
Anviking, line 69: Oops. Missing explicit imports.

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
```diff
@@ -0,0 +1,262 @@
...
+    ( Qry, runQuery, slotToEpoch )
+import Ouroboros.Consensus.HardFork.History.Summary
+    ( Summary (..), neverForksSummary )
```
rvl, line 72: Importing of Ouroboros.Consensus doesn't seem right for the cardano-wallet-core package - it is cardano-node backend specific. Is there any way to move this to the cardano-wallet-shelley package?

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
```diff
@@ -0,0 +1,262 @@
...
+epochOf :: Cardano.SlotNo -> Qry EpochNo
+epochOf slot = do
+    (e, _, _) <- slotToEpoch slot
+    return $ EpochNo $ fromIntegral $ Cardano.unEpochNo e
+
+-- | An 'Interpreter' for a single era, where the slotting from
+-- @GenesisParameters@ cannot change.
+--
+-- TODO: The type should be changed to @Interpreter@ when we bump
+-- ouroboros-consensus.
```
rvl, line 90: Should we do this? Another PR?

rvl (approved):
LGTM

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
Anviking: No rush. I don't think we need it until #1870 is starting.

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
Anviking: Core already depends on ouroboros-network: https://github.com/input-output-hk/cardano-wallet/blob/315b875cd3c6a78b8c4d4e58bfec6bd4d9acffcd/lib/core/src/Ouroboros/Network/Client/Wallet.hs#L12-L20


- To avoid having `Qry` in core, we'd need to duplicate the interface
- Jormungandr can satisfy the `Qry` interface

I do think it is fine.

_Slightly_ related thread: [point vs forcing jormungandr to use SlotNo](https://input-output-rnd.slack.com/archives/GBT05825V/p1594312980015000)

Anviking: The boundary between jormungandr and byron/shelley could also be revisited later.

Issue #1881, opened 2020-07-08
Add withdrawals to core transaction data-type 


KtorZ:
# Context

In order to correctly report an history of all withdrawals made by a user, we need to be able to filter the transaction history based on whether or not they included withdrawals. 

# Decision

- Add a new database table to store withdrawals alongside input and output
- Wire down everything from the block discovery so that withdrawals are correctly inserted into the database for transactions.

# Acceptance Criteria

1. The API _must_ now return a list of withdrawals with each transactions
2. Withdrawals _should_ be represented as a key:value map of chimeric account mapping to a quantity of lovelace
3. Withdrawals belonging to a wallet must be correctly reported in the API

---

# Development 

- #1892 

# QA 

- Transactions now have a "withdrawals" field documented in the API, reflecting a new field added to the core transaction type. Withdrawals are only present in Shelley, and they are simply defaulted to an empty map for Byron and Jormungandr.

- Withdrawals are made of a stake address (hashed reward account, with network discrimination, bech32 encoded). I've tested the encoding / decoding using a classical roundtrip property:
  https://github.com/input-output-hk/cardano-wallet/blob/26064b21e3dd26fef71913ab8a248ba5ad2f5051/lib/shelley/test/unit/Cardano/Wallet/Shelley/CompatibilitySpec.hs#L127-L136

- There's a new **table** in the database for storing withdrawals. This doesn't require any manual migration since persistent will create the new table automatically. There's one 'gotcha' however, and it is that pre-existing wallet with withdrawals will not "sync back", so it is possible that they have missed withdrawals that have already happened. Restoring the wallet from scratch would solve this. This shouldn't be a problem for end-users since withdrawals haven't yet been introduced to them.

- I made sure to adjust our Arbitrary generators to generate transactions _with_ withdrawals, so that every existing unit tests and state-machine tests for the database functions could be tested. They pass. 
  https://github.com/input-output-hk/cardano-wallet/blob/26064b21e3dd26fef71913ab8a248ba5ad2f5051/lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs#L378-L400

piotr-iohk:
lgtm

PR #1892, opened 2020-07-09
Withdrawals as part of core and API transactions

KtorZ:
# Issue Number

#1881 

# Overview

- 74abe1ca1af0dd31ff90dfa95d8f718fc463c8d0
  :round_pushpin: **add withdrawals to API & core transaction data types**
  Still to be done: withdrawals must be discovered when processing blocks

- 0211c5c20b46fec3df6f143ba65eec545db94374
  :round_pushpin: **upgrade Byron & Jormungandr compatibility code with empty withdrawals**
  Withdrawals are only a thing in Shelley, so for this targets, we can safely replace them by empty lists

- 861a79250cb37e730b41b6db38033f290a4ac1fe
  :round_pushpin: **wire withdrawals in Shelley compatibility code**
  Here, we extract withdrawals directly from the TxBody, the conversion is straightforward

- c36f23a8e673bf32dc7e8554e5bf7fe939d39af5
  :round_pushpin: **include withdrawals where relevant in test and arbitrary generators**
  
- 9f6738ed03318f1d93e7bb2f91665dbe693ba73b
  :round_pushpin: **add withdrawals to the API Swagger specification**
  We represent them as a list and not a key:value map with stake address as key because clients typically don't know the stake address associated with their wallet, so it'd be quite impractical for them to access this data if returned as a map...

- e7fcedb274768d8514b787c746eb758997fa9ef9
  :round_pushpin: **add roundtrip tests for stake address and show that withdrawals are returned from the API**
  
- 00c2d697db0e104dea24fea105df78d6200d6151
  :round_pushpin: **Add stake address constraints to Jormungandr and Byron**
  
- 533b772fb7567f855afe6753ca39d51830504227
  :round_pushpin: **add 'minWithdrawal' query parameter to the swagger specification**
  
- 07a0f235103882d2ba24416e0a45b34976d8ff2a
  :round_pushpin: **extend transaction API to allow filtering by minimal withdrawals amount**
  This is a 'any' sort of filter, such that it returns any transaction where there's at least one withdrawal matching the criteria. The use-case is typically to ask for all transactions where there's a min withdrawal of 1, such that we get the entire withdrawal history

- 3dd7257a2c76ffe453092d7764effa562e3c5e8c
  :round_pushpin: **review integration and unit tests, add additional checks to cover withdrawals**

# Comments

:warning: Sorry, the PR spans over many files because I needed to add a new class constraints `DecodeStakeAddress` which goes side-by-side with the current `DecodeAddress`.


lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -491,6 +463,30 @@ migrateManually tr defaultFieldValues =
...
+                query <- Sqlite.prepare conn $ T.unwords
+                    [ "ALTER TABLE", tableName field
+                    , "ADD COLUMN", fieldName field
+                    , fieldType field, "NOT NULL", "DEFAULT", value
+                    , ";"
+                    ]
+                _ <- Sqlite.step query
+                Sqlite.finalize query
+            ColumnPresent ->
+                traceWith tr $ MsgManualMigrationNotNeeded field
```
KtorZ: I originally started adding the withdrawal to the `TxMeta` table but I reverted and went for a new table instead since it behaves much more like current TxIn and TxOut. I however kept this refactor as I thought it wouldn't hurt :blush:

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
```diff
@@ -125,6 +125,18 @@ TxOut
...
+--
+-- There is no wallet ID because these values depend only on the transaction,
+-- not the wallet. txOutputTxId is referred to by TxMeta
+TxWithdrawal
+    txWithdrawalTxId    TxId                sql=tx_id
+    txWithdrawalAmount  W.Coin              sql=amount
+    txWithdrawalAccount W.ChimericAccount   sql=account
+
+    Primary txWithdrawalTxId txWithdrawalAccount
+    deriving Show Generic
```
KtorZ, line 138: Interesting stuff actually here :point_up:

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -820,16 +820,21 @@ data Tx = Tx
         -- ^ NOTE: Order of outputs matters in the transaction representations. Outputs
         -- are used as inputs for next transactions which refer to them using
         -- their indexes. It matters also for serialization.
+    , withdrawals
+        :: !(Map ChimericAccount Coin)
+        -- ^ Withdrawals (of funds from a registered reward account) embedded in
+        -- a transaction. The order does not matter.
```
KtorZ, line 828: And here :point_up_2:

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -816,6 +832,61 @@ toStakePoolDlgCert xpub (W.PoolId pid) =
...
+                Left msg
+
+            Right (remaining,_,_) | not (BL.null remaining) ->
+                Left msg
+
+            Right (_,_,a) ->
+                Right a
+      where
+        msg = TextDecodingError
+            "Unable to decode stake-address: not a well-formed address."
```
KtorZ: I'll add a roundtrip test for these instances.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -911,9 +983,6 @@ instance Buildable addr => Buildable (ConnectionId addr) where
 instance Buildable LocalAddress where
     build (LocalAddress p) = build p
 
-instance Buildable W.ChimericAccount where
-    build (W.ChimericAccount addr) = hexF addr
```
KtorZ, line 947: Found this little orphan :upside_down_face: ... No longer needed since it's define in the core package next to its relevant type declaration.

specifications/api/swagger.yaml
```diff
@@ -460,6 +465,20 @@ x-transactionOutputs: &transactionOutputs
...
+  minItems: 0
+  items:
+    type: object
+    additionalProperties: false
+    required:
+      - stake_address
+      - amount
+    properties:
+      stake_address: *stakeAddress
+      amount: *amount
```
KtorZ, line 480: Here, I've chosen not to represent withdrawals as a key:value map in the API because it would otherwise be a key:value map where all keys are pretty much unknown by the client :no_mouth: ... wallet clients aren't typically aware of what their stake address is, so having something like:

```json
{ "stake_addr1sjck9mdmfyhzvjhydcjllgj9vjvl522w0573ncustrrr2rg7h9azg4cyqd36yyd48t5ut72hgld0fg2x": 42 }
```

Would make it quite hard for them to inspect the content of that value ^^"... Hence, in the API, I am turning the `Map` we have internally into a list of object with a `stake_address` and `amount` properties.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -534,9 +539,15 @@ data ApiTransaction (n :: NetworkDiscriminant) = ApiTransaction
     , direction :: !(ApiT Direction)
     , inputs :: ![ApiTxInput n]
     , outputs :: ![AddressAmount (ApiT Address, Proxy n)]
+    , withdrawals :: ![ApiWithdrawal n]
     , status :: !(ApiT TxStatus)
     } deriving (Eq, Generic, Show)
 
+data ApiWithdrawal n = ApiWithdrawal
```
paweljakubas, line 549: :+1:

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -660,15 +671,20 @@ fromShelleyTx
...
     ( W.Tx
         (fromShelleyTxId $ SL.txid bod)
         (map ((,W.Coin 0) . fromShelleyTxIn) (toList ins))
         (map fromShelleyTxOut (toList outs))
+        (fromShelleyWdrl wdrls)
     , mapMaybe fromShelleyDelegationCert (toList certs)
     , mapMaybe fromShelleyRegistrationCert (toList certs)
     )
 
+fromShelleyWdrl :: SL.Wdrl TPraosStandardCrypto -> Map W.ChimericAccount W.Coin
```
paweljakubas, line 696: :+1:

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -387,15 +387,15 @@ computeTxSize proxy action cs =
 
     dummyKeyHashRaw = BS.pack (replicate 28 0)
 
-    withdrawals = mkWithdrawals
+    wdrls = mkWithdrawals
```
paweljakubas, line 440: frankly speaking like more previous naming, but understand the probable rationale (80-line styling)

specifications/api/swagger.yaml
paweljakubas: yeah, sound decision

paweljakubas: Was thinking before how absent withdrawals are going to be represented:  either by 0-length array or by making `withdrawals` in `ApiTransaction` not required... Why the first option is better?

paweljakubas (approved):
LGTM! would be good not to forget `_decodeStakeAddress . _encodeStakeAddress` roundtrip. Also some changes are needed in `/build/cardano-wallet/lib/core/test/bench/db/Main.hs:408` but it is no surprise when changes in such a core place are made.

specifications/api/swagger.yaml
KtorZ: It'll be a 0-length array.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
KtorZ: Yeah, I usually try to keep naming consistent with the surrounding context.

Issue #1889, opened 2020-07-09
inputs_depleted error when sending tx / estimating fee / running coin selection on wallet with total balance = 0


piotr-iohk:
# Context
 
The error used to be `not_enough_money`, now it is:
```
{"code":"inputs_depleted"
,"message":"I cannot select enough UTxO from your wallet to construct an adequate transaction. Try sending a smaller amount or increasing the number of available UTxO."}
```

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | `2020-06-07` |
| Platform     | Linux, all          |
| Installation | from source            |

# Steps to Reproduce

1. Try to send tx / estimate fee or run random coin selection on either Shelley or Byron wallet with total balance = 0 (i.e. UTxO = 0 and rewards = 0)

## Expected behavior

```
not_enough_money
```

## Actual behavior
 
```
{"code":"inputs_depleted"
,"message":"I cannot select enough UTxO from your wallet to construct an adequate transaction. Try sending a smaller amount or increasing the number of available UTxO."}
```

---

# Resolution 

#1896 

---

# QA 

- After fixing the bug, I had to revert back some error messages in integration tests that were changed during the withdrawals introduction. Messages are now back to what they used to be, which the behavior detailed above.

piotr-iohk:
lgtm

PR #1896, opened 2020-07-10
adjust LargestFirst error handling to return 'not enough money' when there's no UTxO

KtorZ:
# Issue Number

#1889 

# Overview

- 912cf3ade3e523b6413d56f0397fcceb3fc6b7a2
  :round_pushpin: **adjust LargestFirst error handling to return 'not enough money' when there's no UTxO**
  

# Comments

I imagine there's at least an integration test that would fail and for which the message will need to be changed. I was too lazy to look for it so I'll wait for the CI to fail and report it.


piotr-iohk (approved):
ok, works as expected.

paweljakubas (approved):

PR #1897, opened 2020-07-10
run integration tests twice if the first time failed

KtorZ:
# Issue Number

N/A

# Overview

- 9aa974bd3adec2ce3b330639c6c2f4273c9d48ae
  :round_pushpin: **run integration tests twice if the first time failed**
    This is to cope with their inherent flakyness due to the very 'volatile' nature of the chain. Indeed, in Praos, rollbacks happens quite often and, while this is not a problem under normal circumstances, it can be pretty annoying when fixtures are being rolled back or when the wallet hasn't got time to catch up between a rollback and a test assertion.

  So, instead of retrying every single assertion, we simply retry every scenario once if they fail.

  Hspec 'aroundAll', 'before', 'beforeAll' etc.. are built with monad composition and exception primitivies ('finally'), so re-running the IO action should also re-run automatically the necessary tear up and down code

# Comments

### DIFF
--- a/lib/test-utils/src/Test/Hspec/Extra.hs
+++ b/lib/test-utils/src/Test/Hspec/Extra.hs
@@ -12,6 +12,7 @@
 
 module Test.Hspec.Extra
     ( aroundAll
+    , it
     ) where
 
 import Prelude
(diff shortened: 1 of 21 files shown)


lib/test-utils/src/Test/Hspec/Extra.hs
```diff
@@ -99,6 +101,12 @@ aroundAll acquire =
                 unlock release
                 await done
 
+-- | A drop-in replacement for 'it' that'll automatically retry a scenario once
+-- if it fails, to cope with potentially flaky tests.
+it :: HasCallStack => String -> ActionWith ctx -> SpecWith ctx
```
paweljakubas, line 106: :+1:

paweljakubas (approved):

Issue #1708, opened 2020-06-02
Cardano-wallet-shelley workers crash


piotr-iohk:
# Context
| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | cardano-wallet: 2020.5.6 (git revision: 3fb783f5545862fbd5cd7c55c8251167e37d1377), cardano-node: 1.12 044e82d68a5e5d05258c93eda6c00d8541e4d977   |
| Platform     | Linux         |
| Installation | From source            |

### 1.
I observed that `workers` on cardano-wallet tend to die while wallet is just running on top of cardano-node. After running wallet overnight on a VPS server I had:
  - 2 out of 3 of my Byron wallets showing status: not_responding
 -  Listing of Shelley wallets was giving `500` :
```
Code = 500,
Message = {"code":"wallet_not_responding","message":"That's embarassing. My associated worker for e9bc1ad5c4273eed7813483a40342e01ad47e243 is no longer responding. This is not something that is supposed to happen. The worker must have left a trace in the logs of severity 'Error' when it died which might explain the cause. Said differently, this wallet won't be accessible until the server is restarted but there are good chances it'll recover itself upon restart."}
```
### 2.
I was able to reproduce the situation by starting and stopping `cardano-node` while wallet was running. (however it was intermittent, i.e. not every cardano-node restart was causing that). 

There was in the wallet.log upon restarting cardano-node (and workers crash as a result):
```
[cardano-wallet.network:Info:119][2020-06-02 08:16:13.26 UTC] Mux ConnectionId {localAddress = LocalAddress {getFilePath = ""}, remoteAddress = LocalAddress {getFilePath = "/root/node/relay/node.socket"}} Send MsgProposeVersions (fromList [(NodeToClientV_2,TInt 42)])
[cardano-wallet.network:Warning:121][2020-06-02 08:16:13.26 UTC] Couldn't connect to node (x11). Retrying in a bit...
[cardano-wallet.network:Info:121][2020-06-02 08:16:13.26 UTC] Mux ConnectionId {localAddress = LocalAddress {getFilePath = ""}, remoteAddress = LocalAddress {getFilePath = "/root/node/relay/node.socket"}} Send MsgProposeVersions (fromList [(NodeToClientV_2,TInt 42)])
[cardano-wallet.network:Warning:125][2020-06-02 08:16:13.27 UTC] Couldn't connect to node (x11). Retrying in a bit...
[...]

PR #1900, opened 2020-07-13
Testing byron txs and migrations in Shelley

piotr-iohk:
# Issue Number

https://github.com/input-output-hk/cardano-wallet/issues/1844, https://github.com/input-output-hk/cardano-wallet/issues/1675

# Overview

- 66e5151f26172f957f50eab75e400b4e4c57384c
  Enable more integration tests for byron transactions, migrations, addresses and HW wallets in shelley
  
- d773d955a9990e33c3a51ba6613b87a365d7c7d3
  Add some special Byron and Icarus wallets to genesis for testing migration scenarios
  
- 834923ede8987b78eaf0cc9a32aa0f5b3582347d
  Use regular HSpec.it for tests that use explicit/special mnemonics, otherwise in case they fail, on repeat there will be a 409 as such wallet would exist
  
- 18d1c3487915013512312e413ec2f8ce17f81609
  Don't run Byron tx test on Jormungandr

  

# Comments

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Byron/Transactions.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Byron/Transactions.hs
@@ -62,6 +62,7 @@ import Test.Integration.Framework.TestData
 import qualified Cardano.Wallet.Api.Link as Link
 import qualified Data.Text as T
 import qualified Network.HTTP.Types.Status as HTTP
+import qualified Test.Hspec as Hspec
 
 data TestCase a = TestCase
     { query :: T.Text
@@ -72,9 +73,9 @@ spec :: forall n t.
     ( DecodeAddress n
     , DecodeStakeAddress n
     ) => SpecWith (Context t)
-spec = do
+spec = describe "BYRON_TX_COMMON" $ do
 
-    it "BYRON_RESTORE_08 - Icarus wallet with high indexes" $ \ctx -> do
+    Hspec.it "BYRON_RESTORE_08 - Icarus wallet with high indexes" $ \ctx -> do
         -- NOTE
         -- Special Icarus mnemonic where address indexes are all after the index
         -- 500. Because we don't have the whole history, restoring sequential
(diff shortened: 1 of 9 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Byron/Transactions.hs
```diff
@@ -98,27 +99,27 @@ spec = do
             , expectField (#balance . #available) (`shouldBe` Quantity faucetAmt)
             ]
 
-    it "BYRON_RESTORE_09 - Ledger wallet" $ \ctx -> do
-        -- NOTE
-        -- Special legacy wallets where addresses have been generated from a
-        -- seed derived using the auxiliary method used by Ledger.
-        let mnemonics =
```
piotr-iohk, line 105: I don't know how to generate this special legacy ledger wallet. For the moment I commented out this test.

lib/core-integration/src/Test/Integration/Scenario/API/Byron/TransactionsShelley.hs
```diff
@@ -0,0 +1,171 @@
...
+              between
+                  ( Quantity (faucetAmt - feeEstMax - amt)
+                  , Quantity (faucetAmt - feeEstMin - amt)
+                  )
+          , expectField
+                  (#balance . #available)
+                  (.>= Quantity (faucetAmt - faucetUtxoAmt))
+          ]
+
+      eventually "wa and wb balances are as expected" $ do
```
paweljakubas:
```suggestion
      eventually "wByron and wShelley balances are as expected" $ do
```

paweljakubas (approved):
LGTM!

piotr-iohk:
Tests timeout in CI. Locally running integration tests and they run already for an ~1h. I think the problem is that there is a manifestation of this error (https://github.com/input-output-hk/cardano-wallet/issues/1708) or something similar going on which makes a lot of test stuck on `eventually` step and fail... (Trying to list shelley wallets on the wallet from integration tests, while they are running shows `500`)

Perhaps some tests I enabled are causing this... I'll try to isolate.

rvl:
@piotr-iohk it says the tests were taking longer than 45 minutes.
Are they getting stuck, or do we need to increase the timeout?
Timeout is here: https://github.com/input-output-hk/cardano-wallet/blob/master/.buildkite/rebuild.hs#L151
In Buildkite you can see the stack log files under the "Artifacts" tab.

piotr-iohk:
Running locally I got:
```
Finished in 4349.0001 seconds
643 examples, 297 failures, 3 pending
```
All failures are:
```
       DecodeFailure "{\"code\":\"wallet_not_responding\",
\"message\":\"That's embarassing. My associated worker for 07981b66dda508d67d8b212ca3ecacafa6d505fd is no longer responding. This is not something that is supposed to happen. 
The worker must have left a trace in the logs of severity 'Error' when it died which might explain the cause. Said differently, this wallet won't be accessible until the server is restarted but there are good chances it'll recover itself upon restart.\"}"

```
@rvl so I don't think increasing timeout will help...
Some test could be causing this or this (#1708) "just" happens at some point which causes subsequent test to fail.

piotr-iohk:
I'm not able to isolate a specific test but the issue seems to be happening while running tests tagged `BYRON_MIGRATIONS`... running tests from there 1 by 1 appears to be fine... :thinking:

piotr-iohk:
tests passed for me locally after rebasing. :man_shrugging:

piotr-iohk:
bors try

rvl (approved):

paweljakubas:
@rvl  @piotr-iohk  I wonder if we don't need to increase CI timeouts to accommodate those new scenarios

piotr-iohk:
bors try

piotr-iohk:
I'm not sure what is the problem:
```
killed after writing more than 67108864 bytes of log output
```
(included in the CI failures spreadsheet)

The suite passes now for me locally:
```
Finished in 1365.8492 seconds
646 examples, 0 failures, 3 pending
```

piotr-iohk:
Again:
```
killed after writing more than 67108864 bytes of log output
```
:(

KtorZ (changes requested):
Please let's merge this one after #1909. In particular, #1909 will re-introduce scenarios for Byron wallets as well as migration from Byron to Shelley already.

piotr-iohk:
We might as well close this one if #1909 has equivalent scenarios.

PR #1901, opened 2020-07-13
Use SlotNo and TimeInterpreter

Anviking:
# Issue Number

ADP-356 #1868 #1869 

# Overview

- [x] Add TimeInterpreter, necessary queries, and a way to run them
Add tests comparing them with old implementation

- [x] Fundamentally change wallet to use SlotNo instead of SlotId
    - Patch together cardano-wallet-jormungandr to still work, despite the
binary using SlotId.

- [x] Remove redundant byron code

- [x] Sync progress calculation compares times now instead of slots

- [x] Rename the `slot` tables to `slot_no` to force automatic migration, preventing catastrophic re-interpretation of old data

# Comments

- Depends on #1890

### DIFF
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
@@ -92,7 +92,7 @@ TxMeta
     txMetaWalletId     W.WalletId   sql=wallet_id
     txMetaStatus       W.TxStatus   sql=status
     txMetaDirection    W.Direction  sql=direction
-    txMetaSlot         SlotNo       sql=slot
+    txMetaSlot         SlotNo       sql=slot_no
     txMetaBlockHeight  Word32       sql=block_height
     txMetaAmount       Natural      sql=amount
 
@@ -143,7 +143,7 @@ TxWithdrawal
 -- Volatile checkpoint data such as AD state will refer to this table.
 Checkpoint
     checkpointWalletId          W.WalletId   sql=wallet_id
-    checkpointSlot              SlotNo       sql=slot
+    checkpointSlot              SlotNo       sql=slot_no
     checkpointHeaderHash        BlockId      sql=header_hash
     checkpointParentHash        BlockId      sql=parent_header_hash
     checkpointBlockHeight       Word32       sql=block_height
@@ -174,7 +174,7 @@ ProtocolParameters
 -- Track whether the wallet's stake key is registered or not.
 StakeKeyCertificate
     stakeKeyCertWalletId             W.WalletId            sql=wallet_id
-    stakeKeyCertSlot                 SlotNo                sql=slot
+    stakeKeyCertSlot                 SlotNo                sql=slot_no
     stakeKeyCertType                 W.StakeKeyCertificate sql=type
 
     Primary stakeKeyCertWalletId stakeKeyCertSlot
@@ -184,7 +184,7 @@ StakeKeyCertificate
 -- Store known delegation certificates for a particular wallet
 DelegationCertificate
     certWalletId             W.WalletId     sql=wallet_id
-    certSlot                 SlotNo         sql=slot
+    certSlot                 SlotNo         sql=slot_no
     certPoolId               W.PoolId Maybe sql=delegation
 
     Primary certWalletId certSlot
@@ -208,7 +208,7 @@ UTxO                                sql=utxo
 
     -- The wallet checkpoint (wallet_id, slot)
     utxoWalletId        W.WalletId  sql=wallet_id
-    utxoSlot            SlotNo      sql=slot
+    utxoSlot            SlotNo      sql=slot_no
 
     -- TxIn
     utxoInputId         TxId        sql=input_tx_id
@@ -239,7 +239,7 @@ SeqState
 -- when they were discovered.
 SeqStateAddress
     seqStateAddressWalletId         W.WalletId     sql=wallet_id
-    seqStateAddressSlot             SlotNo         sql=slot
+    seqStateAddressSlot             SlotNo         sql=slot_no
     seqStateAddressAddress          W.Address      sql=address
     seqStateAddressIndex            Word32         sql=address_ix
     seqStateAddressAccountingStyle  W.AccountingStyle  sql=accounting_style
@@ -277,7 +277,7 @@ RndState
 -- The set of discovered addresses.
 RndStateAddress
     rndStateAddressWalletId      W.WalletId        sql=wallet_id
-    rndStateAddressSlot          SlotNo            sql=slot
+    rndStateAddressSlot          SlotNo            sql=slot_no
     rndStateAddressAccountIndex  Word32            sql=account_ix
     rndStateAddressIndex         Word32            sql=address_ix
     rndStateAddressAddress       W.Address         sql=address
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
@@ -92,7 +92,7 @@ TxMeta
     txMetaWalletId     W.WalletId   sql=wallet_id
     txMetaStatus       W.TxStatus   sql=status
     txMetaDirection    W.Direction  sql=direction
-    txMetaSlot         SlotNo       sql=slot_no
+    txMetaSlot         SlotNo       sql=slot
     txMetaBlockHeight  Word32       sql=block_height
     txMetaAmount       Natural      sql=amount
 
@@ -143,7 +143,7 @@ TxWithdrawal
 -- Volatile checkpoint data such as AD state will refer to this table.
 Checkpoint
     checkpointWalletId          W.WalletId   sql=wallet_id
-    checkpointSlot              SlotNo       sql=slot_no
+    checkpointSlot              SlotNo       sql=slot
     checkpointHeaderHash        BlockId      sql=header_hash
     checkpointParentHash        BlockId      sql=parent_header_hash
     checkpointBlockHeight       Word32       sql=block_height
@@ -174,7 +174,7 @@ ProtocolParameters
 -- Track whether the wallet's stake key is registered or not.
 StakeKeyCertificate
     stakeKeyCertWalletId             W.WalletId            sql=wallet_id
-    stakeKeyCertSlot                 SlotNo                sql=slot_no
+    stakeKeyCertSlot                 SlotNo                sql=slot
     stakeKeyCertType                 W.StakeKeyCertificate sql=type
 
     Primary stakeKeyCertWalletId stakeKeyCertSlot
@@ -184,7 +184,7 @@ StakeKeyCertificate
 -- Store known delegation certificates for a particular wallet
 DelegationCertificate
     certWalletId             W.WalletId     sql=wallet_id
-    certSlot                 SlotNo         sql=slot_no
+    certSlot                 SlotNo         sql=slot
     certPoolId               W.PoolId Maybe sql=delegation
 
     Primary certWalletId certSlot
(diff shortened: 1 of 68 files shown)


lib/core/test/unit/Cardano/Pool/DB/Properties.hs
```diff
@@ -799,9 +798,11 @@ prop_listRegisteredPools DBLayer {..} entries =
         L.nub poolOwners /= poolOwners
 
     prop = do
+        -- TODO#1901: Is this change ok? Previously only (SlotInEpoch 0) were
+        -- generated.
```
Anviking: @jonathanknowles do you know off the top of your head if there was a reason for only `SlotId`s with slotInEpoch 0 were generated?
(you recently edited here)

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
```diff
@@ -66,31 +78,100 @@ import Numeric.Natural
...
+-- A @TimeInterpreter@ is a way for the wallet to run things of type @Qry a@.
+--
+-- NOTE:
+-- A @TimeInterpreter@ could in theory decide to update the era summary from the
+-- node when running a query.
+--
+-- We cannot manually specify when the fetching happens.
+--
+-- This may or may not be what we actually want.
+type TimeInterpreter m = forall a. Qry a -> m a
```
Anviking, line 199: ☝️ We may now have a `TimeInterpreter IO` that _sometimes_ takes a long time to run a conversion query (when it needs to refetch the summary over LSQ).

Not sure if we should change…

The re-fetching should happen on very rare occasions though…

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
```diff
@@ -90,7 +92,7 @@ TxMeta
     txMetaWalletId     W.WalletId   sql=wallet_id
     txMetaStatus       W.TxStatus   sql=status
     txMetaDirection    W.Direction  sql=direction
-    txMetaSlot         W.SlotId     sql=slot
+    txMetaSlot         SlotNo       sql=slot
```
Anviking, line 95: I suspect we may need manual migration

rvl: Don't fret over DB migration if it's too hard. We can start from a fresh state directory if necessary.

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
rvl: I think it would be better to have the network layer query the Interpreter on a schedule, and stash it in a mutable variable. So that all queries can be pure.

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
Anviking: On second thought, I fear/wonder if users would silently get corrupted data, as the meaning of the `slot` table has changed, but not the name.

Is there a good way to test / detect this?

_Or I guess testing manually using CLI works. I tried using the Daedalus mainnet installation, and replacing the wallet and node manually, but I think too many other things have changed for it to work._

rvl: If the type of the column changes, I think persistent auto migrations will either throw an exception or helpfully delete the table.

Let's decide next week whether to automatically migrate or make a fresh state directory.

rvl:
What's the benefit of wrapping the provided Interpreter in our own type? Can we use the bare Interpreter?

Anviking:
> What's the benefit of wrapping the provided Interpreter in our own type?

The original interpreter only works with `RelativeTime`, which is relative to genesis.

The wrapper carries the genesis start date along with it.

Anviking:
bors try

Anviking:
bors try

Anviking:
I am seeing rare failures of this sort. I am investigating
```haskell
  test/unit/Cardano/Pool/DB/Properties.hs:348:17:
  1) Cardano.Pool.DB.Sqlite.Sqlite, Stake Pool properties, readPoolProduction for a given epoch should always give slots from given epoch
       Falsifiable (after 10 tests and 1 shrink):
         StakePoolsFixture {poolSlots = [(PoolId {getPoolId = "w\EOTe\227Aa\169\r=\fK%v>C\178\r/\ETX\v\FS\131>\ACK\225\aG$u\a\v>"},BlockHeader {slotNo = SlotNo {unSlotNo = 10}, blockHeight = Quantity {getQuantity = 0}, headerHash = Hash {getHash = "00000000000000000000000000000001"}, parentHeaderHash = Hash {getHash = "00000000000000000000000000000000"}}),(PoolId {getPoolId = "\138P0PcK\141\161.\a\DC1\DC3L\ESC\a\SOT\FS\US\245[\209&Q\CAN\DLEWU\249\DC1\165?"},BlockHeader {slotNo = SlotNo {unSlotNo = 0}, blockHeight = Quantity {getQuantity = 0}, headerHash = Hash {getHash = "00000000000000000000000000000001"}, parentHeaderHash = Hash {getHash = "00000000000000000000000000000000"}}),(PoolId {getPoolId = "w\EOTe\227Aa\169\r=\fK%v>C\178\r/\ETX\v\FS\131>\ACK\225\aG$u\a\v>"},BlockHeader {slotNo = SlotNo {unSlotNo = 8}, blockHeight = Quantity {getQuantity = 0}, headerHash = Hash {getHash = "00000000000000000000000000000001"}, parentHeaderHash = Hash {getHash = "00000000000000000000000000000000"}})], rollbackSlots = [SlotNo {unSlotNo = 10},SlotNo {unSlotNo = 0},SlotNo {unSlotNo = 8}]}
       expected: fromList [SlotNo {unSlotNo = 0},SlotNo {unSlotNo = 8}]
        but got: fromList [SlotNo {unSlotNo = 10}]
```

Anviking:
bors try

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
Anviking: It seems to work 🤷‍♂️

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
Anviking: >  So that all queries can be pure

I presume it would still be in `IO` (to access the var), so no change to the interface.

In that case I'd be concerned with what callers would see if there is nothing in the var. There are many call-sites, and I think blocking until it is available is the simplest thing to. Which is somewhat like this.

But preemptively refreshing _before_ a conversion fails, sounds sane to do… optimally.

Anviking:
bors try

Anviking:
9eb843b broke a migration test
```
  test/unit/Cardano/Wallet/DB/SqliteSpec.hs:236:9:
  1) Cardano.Wallet.DB.Sqlite, Manual migrations, 'migrate' db with no passphrase scheme set.
       uncaught exception: MigrationError
       MigrationError {getMigrationErrorMessage = "PersistError
```

But it also seems that automatic migration works just fine without renaming the column, so -> reverted with 0df6556

Anviking:
bors try

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
rvl: Let's see. I was thinking that the initial Interpreter will be one constructed from genesis parameters.
It will get updated once the node has synced far enough. (Question - how far is enough?)
This would prevent the wallet from being unresponsive while the node is syncing.

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
```diff
@@ -59,54 +77,198 @@ import Data.Quantity
...
-            $ gp ^. #getSlotLength
-
-        , eraSafeZone =
-            noLowerBoundSafeZone (k * 2)
-        }
+singleEraInterpreter :: GenesisParameters -> TimeInterpreter Identity
+singleEraInterpreter gp q = either bomb return $ runQuery q int
   where
-    k = fromIntegral $ getQuantity $ getEpochStability gp
+    bomb x = error $ "singleEraInterpreter: " <> show x
```
rvl: Could you add a comment saying that these queries will never fail?

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -20,6 +20,10 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+-- Needed for (Buildable SlotNo)
+-- TODO#1901: How do we want to deal with the orphan (Buildable SlotNo) instance?
```
rvl: I would add an `Cardano.Orphans` module or something like that, where it would be acceptable to have `-fno-warn-orphans`.

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
```diff
@@ -12,11 +16,20 @@
...
+    , ongoingSlotAt
+
+    -- ** Running queries
+    , TimeInterpreter
     , singleEraInterpreter
-    , runQuery
+    , Qry
 
-      -- * Old functions
+      -- * Legacy api
```
rvl, line 32: Will these all be removed in a future PR, or moved to a tests-only module?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -559,15 +555,11 @@ mkWithdrawals proxy (ChimericAccount keyHash) amount
...
--- work if the epoch length changed in a hard fork.
-
 -- NOTE: The (+7200) was selected arbitrarily when we were trying to get
 -- this working on the FF testnet. Perhaps a better motivated and/or
 -- configurable value would be better.
-defaultTTL :: EpochLength -> SlotId -> SlotNo
-defaultTTL epochLength slot =
-    (toSlotNo epochLength slot) + 7200
+defaultTTL ::  SlotNo -> SlotNo
+defaultTTL slot = slot + 7200
```
rvl, line 562: I like this. :-)

rvl (approved):
Nice work - this is a really big PR.
I have checked Cardano.Wallet.Primitive and Cardano.Wallet.Shelley.
Is there anything else I should pay attention to?

rvl:
With the database migration issue, we can't re-use the `slot` columns if the meaning of their values has changed. So we have to do something about this situation.

We could either:
1. implement some "manual migration" code to restore all wallets from genesis if loading a DB from before this change.
2. say that the DB is incompatible with previous versions, and a fresh state directory must be used.

In the case of 2, the DB migration test can be disabled, because there is no migration to test. I like option 2, because there are other user stories such as transaction TTL that also need major DB schema changes.

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
Anviking: I suppose we should remove them completely, and re-write the tests to use the new ones.

Now we are testing the old ones, and that the new ones are equivalent to the old ones (Generators aren't perfect though).

Anviking:
Migrating from 2020-07-06 works. Migrating from the current flight candidate to this PR doesn't:

<details>

Skärmavbild 2020-07-17 kl  12 47 59

</details>

But in the real world, users would be upgrading from an older Daedalus Mainnet release, which might trigger automatic migration.

> In the case of 2, the DB migration test can be disabled, because there is no migration to test. I like option 2, because there are other user stories such as transaction TTL that also need major DB schema changes.

✔️ 

Regardless, I think thinking about this later is fine, as you suggested in another comment.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
Anviking: 👍  Done

PR #1902, opened 2020-07-14
Update cardano-addresses dep

rvl:
### Issue Number

input-output-hk/cardano-addresses#41

### Overview

- The latest cardano-addresses has fixes for windows.
- ~Update bech32 to the Hackage release.~ ⇒ #1906

### Comments

There seems to be an occasional memory leak in the cardano-wallet-core unit tests on this branch, but not on master.


jonathanknowles:
I think we still might need to include `bech32-th`, the latest version of which is still `1.0.2`.

rvl:
Right thanks, good point.

rhyslbw (approved):

jonathanknowles (approved):

rvl:
Unit tests were killed ... they work fine locally ... one for the failure tracking sheet.

bors r+

PR #1909, opened 2020-07-15
Cardano.Api -> Cardano.Api.Typed and HardForkBlock

rvl:
### Issue Number

Relates to #1870.

### Overview

- Updates cardano-node to latest master version.
- Updates libraries to version from cardano-node master.
- Cardano.Api → Cardano.Api.Typed changes.
- Change `ShelleyBlock` → `CardanoBlock sc` in "Compatibility" module
- Change `ShelleyBlock` → `CardanoBlock sc` in network module
- Add in byron implementations of conversion functions for two-era mode.
- Add CLI support for two-era mode.
- Use LocalStateQuery to get the history Interpreter from the node.
- Update shelley transaction layer to use new extended keys support in Cardano.Api.Typed.

### Comments

- [x] Fix all the type errors in packages other than cardano-wallet-shelley.
- [x] Transaction layer works during shelley era
- [ ] Transaction layer works during byron era
- [x] Manual testing against [mainnet_candidate](https://hydra.iohk.io/job/Cardano/iohk-nix/cardano-deployment/latest/download/1/index.html) which is in byron era -- can sync and restore empty wallet
- [x] Manual testing against [shelley_qa](https://hydra.iohk.io/job/Cardano/iohk-nix/cardano-deployment/latest/download/1/index.html) which is in shelley era -- can sync and restore empty wallet
- [ ] Unit tests passing
- [x] Merge integration test cluster setup from #1924, and fix.


lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -192,23 +197,30 @@ mkTx
...
 
-    wits <- case (txWitnessTagFor @k) of
+    sealed <- case (txWitnessTagFor @k) of
         TxWitnessShelleyUTxO -> do
+            let paymentKey = _ (getRawKey k, pwd)
+            let stakeKey
+                     | null wdrls = []
+                     | otherwise = [_ (getRawKey rewardAcnt, pwdAcnt)]
+
+            tx <- signShelleyTransaction unsigned (paymentKey:stakeKey)
```
Anviking: This seems weird to me, what was the problem here?

Anviking: Also, it seems you almost went in the direction of creating a byron vs Shelley tx based on the key/witness tag, but we may spend from byron addresses in shelley.

 So we need to either on the era we're in (or if there's something in cardano-api that does that automatically)?

rvl: Basically what happened is there were functions removed that we were using, so I attempted to use the new API and broke it.

KtorZ:
>  Transaction layer works during byron era

We can tackle that in a second time..... Or not. With the time constraints we have, there will likely be very little time between the release and the hard-fork. So we may as well leave it as it is.

lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -141,18 +143,20 @@ data NetworkLayer m target block = NetworkLayer
...
+
     , timeInterpreter
-        :: TimeInterpreter m
+        :: HasCallStack => TimeInterpreter m
     }
 
 instance Functor m => Functor (NetworkLayer m target) where
     fmap f nl = nl
         { nextBlocks = fmap (fmap f) . nextBlocks nl
+        , stakeDistribution = error "fixme: functor instance"
```
KtorZ: :point_up: Just leaving a marker here.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -416,39 +383,45 @@ computeTxSize proxy pm witTag action cs =
...
         dummyOutput = TxOut $ Address $ BS.pack (1:replicate 56 0)
 
-        dummyKeyHash = SL.KeyHash . Hash.UnsafeHash $ dummyKeyHashRaw
+        dummyStakeCred = toCardanoStakeCredential
+            $ ChimericAccount dummyKeyHashRaw
+
+        dummyPoolId :: Cardano.PoolId
+        dummyPoolId = fromMaybe (error "dummyPoolId couldn't be constructed")
+            $ Cardano.deserialiseFromRawBytes (Cardano.AsHash Cardano.AsStakePoolKey)
+            $ BS.pack $ replicate 32 0
```
KtorZ: This needs to be 28 bytes.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -416,39 +383,45 @@ computeTxSize proxy pm witTag action cs =
...
         (ChimericAccount dummyKeyHashRaw)
         (withdrawal cs)
 
     (addrWits, certWits) =
-        ( Set.union
-            (Set.map dummyWitnessUniq $ Set.fromList (fst <$> CS.inputs cs))
-            (if Map.null wdrls then Set.empty else Set.singleton (dummyWitness "0"))
+        (
+            map dummyWitnessUniq (fst <$> CS.inputs cs)
+            <> [dummyWitness "0" | null wdrls]
```
KtorZ:
```hs
>>> [14|True]
[14]

>>> [14|False]
[]
```

This creates a dummy witness if the witness set is empty. Which is, the opposite of what we want :grimacing: ... The list comprehension is shorter / somewhat more elegant but also more confusing, so I'd rather put back the good old `if null then [] else [...]` which is less error-prone.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -506,11 +483,13 @@ computeTxSize proxy pm witTag action cs =
     bloatChaff :: Word -> BL.ByteString -> ByteString
     bloatChaff n = BL.toStrict . BL.take (fromIntegral n) . BL.cycle
 
+    -- TODO: Surely we can allow byron witnesses paying for certificates?
+    -- Should be no reason to case here.
```
KtorZ: Actually we do not. We don't have hybrid wallets, so a wallet that would require Byron witnesses is by construction a "Byron wallet" with no delegation / rewards capability.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
Anviking: Sorry, hlint asked for it, and thought: fine.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -522,87 +501,114 @@ lookupPrivateKey keyFrom addr =
...
+          SL.bwChainCode  = chainCode,
+          SL.bwAttributes = attributes
+        }
   where
-    (SL.TxId txHash) = SL.txid body
-    signingKey = Crypto.SigningKey $ CC.xPrvChangePass pwd BS.empty prv
-    addrAttr = Byron.mkAttributes $ Byron.AddrAttributes
-        (toHDPayloadAddress addr)
-        (toByronNetworkMagic protocolMagic)
+    sk = CC.SigningKey $ Crypto.HD.xPrvChangePass pwd BS.empty prv
```
KtorZ: Would be nice to be consistent with the above. I'll refactor.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
Anviking: To me it seems like there is not benefit to not allowing it, and limits our options for e.g. integration test setup.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
KtorZ: hlint is stupid. We are not (arguably, but I will argue anytime). Sometimes, when the suggestion isn't sound, we might as well turn it off. I reverted this to a more readable / less error-prone version.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
KtorZ: The benefit is lowering complexity. There are a lot of things that hold together precisely because only Shelley wallets are able to delegate and have withdrawals. Changing this would add a great deal of complexity AND, it'll remove the incentive there's for people to move to Shelley wallets (remember that Byron wallets have security flaws).

Anviking: Changing `mkTx` should have no user-facing effect since shelley wallets' won't select byron inputs.

I also suspect the code here would be more elegant, and we would have the option to use a slightly more flexible `mkTx` in integration tests / for scripts.

KtorZ: From within the transaction layer it shouldn't really have any visible impact indeed. And it's perhaps best to not make any assumption about the "outside context" within the transaction layer actually.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -522,87 +506,117 @@ lookupPrivateKey keyFrom addr =
...
+    :: Cardano.TxBody Cardano.Shelley
+    -> Cardano.NetworkId
     -> Address
     -> (XPrv, Passphrase "encryption")
-    -> SL.BootstrapWitness TPraosStandardCrypto
-mkByronWitness body protocolMagic addr (prv, Passphrase pwd) =
-    SL.makeBootstrapWitness txHash signingKey addrAttr
+    -> Cardano.Witness Cardano.Shelley
+mkByronWitness (Cardano.ShelleyTxBody txbody _) nw addr encryptedKey =
+    Cardano.ShelleyBootstrapWitness $
```
KtorZ, line 598: Any reason why we refactored all these actually and no longer use `makeBootstrapWitness` :thinking: ? This code was quite tricky to get right in the first place, so I am still very doubtful about it working correctly now.

KtorZ: I ended up changing it back in 59535f54c3d49d98c20d432f19dc60c8f1eef172

rvl: Yes it was confusion related to `Cardano.Api.Typed.makeShelleyBootstrapWitness`.

KtorZ:
bors try

KtorZ (approved):

KtorZ:
bors retry

KtorZ:
bors retry

KtorZ:
bors r+

please.

KtorZ:
bors retry

AlexIOHK:
🎉

andy521md:
How to creat an acount




| |
韩守根
|
|
邮箱：jtws302007@126.com
|

签名由 网易邮箱大师 定制

On 07/26/2020 01:15, Alex Apeldoorn wrote:

🎉

Issue #1903, opened 2020-07-14
Performance issues importing many addresses into byron random wallet


rvl:
# Context
 
Using the import byron random address endpoint for transferring unused addresses from a cardano-sl random address derivation wallet is slow.

Related to #1713.

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | 2020-07-06 |
| Platform     | All          |
| Installation |         |

# Steps to Reproduce

1. Use `PUT /v2/byron-wallets/{walletId}/addresses/{addressId}` to import 130000 addresses into a wallet.

Script to make 50,000 adresses:
```
#!/usr/bin/env bash

set -euo pipefail

cardano-address recovery-phrase generate --size 12 > addresses.mnem
cardano-address key from-recovery-phrase Byron > addresses.prv < addresses.mnem

net=764824073
acct_idx=14

for addr_idx in `seq 0 50000`; do
  path=${acct_idx}H/${addr_idx}H
  cat addresses.prv \
    | cardano-address key child $path \
    | cardano-address key public \
    | cardano-address address bootstrap $(cat addresses.prv | cardano-address key public) \
        --network-tag $net --path $path
  echo
done
```

Script to import addresses from addresses.txt individually:
```
#!/usr/bin/env bash

while read addr; do
  curl -i -H 'Content-type: application/json' -XPUT "http://localhost:8092/v2/byron-wallets/59a4c8bc7fcbbfd8ca2473f4752bc47d2da09a3f/addresses/$addr"
done <addresses.txt
```

## Expected behavior

All addresses can be imported in a reasonable amount of time.

## Actual behavior
 
- Users report that API client starts timing out after adding about 6000 addresses.
- From testing on my system, importing addresses starts quickly and gradually gets slower. After 40000 addresses, it takes 2s per address to import.
- The byron address list API can return those 40000 adresses in 12s.

---

# Resolution 

Plan

1. Add integration test reproducing issue ⇒ #1907
2. Add RndState to database benchmarks
3. New endpoint for bulk import of address ⇒ #1911 

---

# QA

piotr-iohk:
LGTM!
Importing 130k addresses => ~7s.
Listing 130k addresses => ~3s.

PR #1911, opened 2020-07-15
Enable bulk import of random addresses.

paweljakubas:
# Issue Number
#1903 

# Overview

- [x] we have added PUT endpoint to `/byron-wallet/{walletId}/addresses` that accepts array of addresses at once
- [x] we have added integration test that illustrate the point. the results are following: 
```
ADDRESS_IMPORT_05 - I can import 100 of addresses --> 0.02 s
ADDRESS_IMPORT_05 - I can import 200 of addresses --> 0.06 s
ADDRESS_IMPORT_05 - I can import 400 of addresses --> 0.11 s
ADDRESS_IMPORT_05 - I can import 800 of addresses --> 0.22 s
ADDRESS_IMPORT_05 - I can import 1600 of addresses --> 0.40 s
ADDRESS_IMPORT_05 - I can import 3200 of addresses --> 0.65 s
ADDRESS_IMPORT_05 - I can import 10000 of addresses --> 1.76 s
ADDRESS_IMPORT_05 - I can import 50000 of addresses --> 18.92 s
ADDRESS_IMPORT_05 - I can import 150000 of addresses --> 186.75 s
```
 - [x] we have updated swagger and core unit tests
-  [x] @piotr-iohk tested endpoint and the results are following : 
```
Start bulk import
  Import 130k addressesFinished in 6.76 seconds (files took 0.33932 seconds to load)
1 example, 0 failures
``` 
Also listing addresses got better. results measured by Piotr:
```
  List 130k addressesFinished in 3.39 seconds (files took 0.34561 seconds to load)
```

# Comments
now, this works : 
```
curl  -vX PUT http://localhost:8090/v2/byron-wallets/0b3a38e4078206ccd93cc353a93cc3a37dbbb4fe/addresses \
  -H "Accept: application/json; charset=utf-8" \
  -d '{
	"addresses": ["2RhQhCGqYPDnCufbC5UrxikckjfXHqogwurY4HhGattJQpuvrDreGSbacwQL9x8o4ysdmJja6TJ2JzDwpj6MzuHHrc733MRyZBHW68uWqAsnKm"]
}' --http1.1  |  jq
```


paweljakubas:
bors try

paweljakubas:
bors try

paweljakubas:
bors try

specifications/api/swagger.yaml
```diff
@@ -2358,6 +2374,25 @@ paths:
...
         responses: *responsesListAddresses
 
+      put:
+        operationId: importAddresses
+        tags: ["Byron Addresses"]
+        summary: Import Addresses
+        description: |
+          <p align="right">status: <strong>stable</strong></p>
+
+          ⚠️  This endpoint is available for `random` wallets only. Any
```
Anviking, line 2384: Maybe we should elaborate briefly on the difference between the endpoints such that people don't just discover the slow one, without realising that this one exists and is better suited for a large number of addresses.

Anviking (approved):
LGTM, but haven't reviewed the details

specifications/api/swagger.yaml
paweljakubas: yes, it would be good, especially if the two survive..but as we talked today, the other one is rather redundant. We can decide tomorrow and if the other survive we will change it as you suggested

KtorZ: > especially if the two survive..but as we talked today, the other one is rather redundant.

:grimacing: I'd argue that the new one is redundant ^.^ ... or more exactly, shall be made redundant once we fixed the issue with the database / checkpoint insertion. 

This should help: https://github.com/input-output-hk/cardano-wallet/pull/1663 although it'd need some more thoughts on the byron case.

Issue #1853, opened 2020-07-04
Prevent staking to a retired pool.


jonathanknowles:
# Context

As part of tracking stake pool retirements, we need to return an error if a user attempts to stake to a pool that has already retired.

# Decision

- Adjust `joinStakePool` to return `ErrNoSuchPool` if the current epoch is later than or equal to the retirement epoch of the specified pool.

# Acceptance Criteria

- [x] The wallet _must_ produce an error if a user attempts to stake to pool that has already retired.

# Development 

- https://github.com/input-output-hk/cardano-wallet/pull/1913

# QA 

- [x] Add an integration test that performs the following steps:

1. Create a new pool **_p_**.
1. Issue a retirement certificate for epoch **_e_**<sub>retirement</sub>.
1. Wait until epoch **_e_**<sub>current</sub> >= epoch **_e_**<sub>retirement</sub>.
1. Verify that attempting to join pool p **does** result in an `ErrNoSuchPool` error.

# Notes

Note that the above integration test is a compromise.

Ideally, a more complete integration test would verify that staking to a pool is still possible right up to the moment before the pool has actually retired. See: https://jira.iohk.io/browse/ADP-385

piotr-iohk:
Works as expected. Also checked that pool that is about to retire (but not yet retired) is still joinable.

PR #1913, opened 2020-07-16
Prevent staking to a retired pool.

jonathanknowles:
# Issue Number

#1853

# Overview

This PR:

- [x] Adjusts `joinStakePool` to return `ErrNoSuchPool` if the current epoch is _later than or equal to_ the retirement epoch of the specified pool.

# Comments

Integration tests specifically designed to test the rejection mechanism are **not** included in this PR, but will be tackled in a future PR. (See QA section of issue https://github.com/input-output-hk/cardano-wallet/issues/1853.)

However, existing tests already verify that it is possible to stake to a _non-retired_ pool, so we can be fairly confident we are not breaking existing functionality.

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -1743,12 +1743,13 @@ joinStakePool
         , AddressIndexDerivationType k ~ 'Soft
         )
     => ctx
+    -> [PoolId]
+    -> PoolId
     -> WalletId
-    -> (PoolId, [PoolId])
     -> ArgGenChange s
     -> Passphrase "raw"
     -> ExceptT ErrJoinStakePool IO (Tx, TxMeta, UTCTime)
-joinStakePool ctx wid (pid, pools) argGenChange pwd = db & \DBLayer{..} -> do
+joinStakePool ctx pools pid wid argGenChange pwd = db & \DBLayer{..} -> do
     (isKeyReg, walMeta) <- mapExceptT atomically
         $ withExceptT ErrJoinStakePoolNoSuchWallet
         $ (,) <$> isStakeKeyRegistered (PrimaryKey wid)
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -2117,6 +2117,7 @@ instance Exception ErrCheckWalletIntegrity
 data ErrCannotJoin
     = ErrAlreadyDelegating PoolId
     | ErrNoSuchPool PoolId
+    | ErrPoolAlreadyRetired PoolId W.EpochNo
     deriving (Generic, Eq, Show)
 
 data ErrCannotQuit
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -114,6 +114,7 @@ module Cardano.Wallet
     , ErrSelectForMigration (..)
 
     -- ** Delegation
+    , PoolRetirementEpochInfo (..)
     , joinStakePool
     , quitStakePool
     , selectCoinsForDelegation
@@ -2158,6 +2159,16 @@ withNoSuchWallet
     -> ExceptT ErrNoSuchWallet m a
 withNoSuchWallet wid =
     maybeToExceptT (ErrNoSuchWallet wid) . MaybeT
+
+data PoolRetirementEpochInfo = PoolRetirementEpochInfo
+    { retirementEpoch
+        :: W.EpochNo
+        -- ^ The retirement epoch of a pool.
+    , currentEpoch
+        :: W.EpochNo
+        -- ^ The current epoch.
+    }
+    deriving (Eq, Generic, Show)
 
 guardJoin
     :: [PoolId]
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -2135,7 +2135,6 @@ instance Exception ErrCheckWalletIntegrity
 data ErrCannotJoin
     = ErrAlreadyDelegating PoolId
     | ErrNoSuchPool PoolId
-    | ErrPoolAlreadyRetired PoolId PoolRetirementEpochInfo
     deriving (Generic, Eq, Show)
 
 data ErrCannotQuit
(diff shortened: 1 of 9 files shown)


jonathanknowles:
bors try

jonathanknowles:
bors try-

jonathanknowles:
bors try

jonathanknowles:
bors try

jonathanknowles:
bors try-

jonathanknowles:
bors try

jonathanknowles:
bors try

jonathanknowles:
bors try

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1715,36 +1718,53 @@ joinStakePool
...
-        guardJoin pools (walMeta ^. #delegation) pid
+        (isKeyReg, walMeta) <- mapExceptT atomically
+            $ withExceptT ErrJoinStakePoolNoSuchWallet
+            $ (,) <$> isStakeKeyRegistered (PrimaryKey wid)
+                  <*> withNoSuchWallet wid (readWalletMeta (PrimaryKey wid))
 
-    let action = if isKeyReg then Join pid else RegisterKeyAndJoin pid
-    liftIO $ traceWith tr $ MsgIsStakeKeyRegistered isKeyReg
+        currentTime <- liftIO getCurrentTime
+        let currentEpoch = view #epochNumber
```
Anviking: Could we please rebase this off #1901 ? 🙏 

Or, actually, no it's probably better to merge this first.

rvl (commented):
Is there a reason that we track retired pools? I don't think we would need info about pools which can't be delegated to. Shouldn't we purge them from the database, and return "pool not found" in this case?

jonathanknowles:
> Is there a reason that we track retired pools?

Yes, it's just what happens by default. 

We already track pool retirement certificates as we encounter them. In the current setup, a retired pool is simply a pool with a retirement certificate whose retirement epoch number is less than or equal to the current epoch.

> I don't think we would need info about pools which can't be delegated to. Shouldn't we purge them from the database, and return "pool not found" in this case?

Perhaps there are two separate questions here:

1. How should we handle garbage collection of retired pools from the pool DB?

2. What error should we return if a user attempts to delegate to a pool that has already retired, but has not yet been garbage collected?

I think we can defer the first question (garbage collection) for a later PR, as it's not strictly required by the [Track Stakepool Retirements](https://github.com/input-output-hk/cardano-wallet/milestone/89) milestone. There's a ticket specifically for this [here](https://jira.iohk.io/browse/ADP-376).

But we certainly need to answer the second question in order to finalize this PR. How about if we modify this PR so that the error returned is just `ErrNoSuchPool`?

jonathanknowles:
bors try

jonathanknowles:
bors try

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1726,36 +1729,53 @@ joinStakePool
...
+                $ fromMaybe W.slotMinBound
+                $ W.slotAt' sp currentTime
```
Anviking: I wasn't able to fully remove `slotAt'` yet in #1901, but we need to use the `TimeInterpreter` for this. Otherwise the calculation will be incorrect.

I suggest using `epochOf`, `ongoingSlotAt`, and running the resulting `Qry` using the `timeInterpreter` field from `networkLayer`. Happy to help if that doesn't work out for some reason.

jonathanknowles: > I wasn't able to fully remove `slotAt'` yet in #1901, but we need to use the `TimeInterpreter` for this. Otherwise the calculation will be incorrect.

Thanks for letting me know. When are you planning to remove the `slotAt'` function? The `getNetworkInformation` endpoint also uses this function in a similar way to determine the _next_ epoch. If that function is updated to use the new API, then we would presumably have to do something similar for `joinStakePool`?

Anviking: > If that function is updated to use the new API, then we would presumably have to do something similar for joinStakePool?

Yes, they should be similar.

Here you handle the `Nothing` case, where it means the current time is behind the genesis date, I think.

With the hard fork, there the query may (also) fail if the node is not in sync, if the current time is beyond the forecast range of the interpreter (think this is up to the next epoch, but not sure exactly)

jonathanknowles:
> 2. What error should we return if a user attempts to delegate to a pool that has already retired, but has not yet been garbage collected?

For the record, we decided to return `ErrNoSuchPool` instead of `ErrPoolAlreadyRetired`.

Fixed in 76b4ab1c6d67b75378b229cef10543a0cd90ff6d.

lib/core/src/Cardano/Wallet.hs
jonathanknowles: > > If that function is updated to use the new API, then we would presumably have to do something similar for joinStakePool?
> 
> Yes, they should be similar.
> 
> Here you handle the `Nothing` case, where it means the current time is behind the genesis date, I think.
> 
> With the hard fork, there the query may (also) fail if the node is not in sync, if the current time is beyond the forecast range of the interpreter (think this is up to the next epoch, but not sure exactly)

Hi @Anviking 

Thanks for your help with this!

I've (fingers crossed) fixed this with the following commits:

* 255210f8b5e6b58a7f4ce47ab83e26b954bc915b
* 7660579b295bd8ad78085da453c6c1c670f45bd5
* 18ef12a77a7a600738e52f85efe292de87168eef
* 23010f8ab79daa7f0ff86d0693b0f02b20c97c0a

jonathanknowles:
bors try

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1877,6 +1899,12 @@ newtype ErrRejectedTip = ErrRejectedTip ApiNetworkTip
 showT :: Show a => a -> Text
 showT = T.pack . show
 
+instance LiftHandler ErrCurrentEpoch where
+    handler = \case
+        ErrUnableToDetermineCurrentEpoch ->
+            apiError err500 UnableToDetermineCurrentEpoch
+                "I'm unable to determine the current epoch."
```
rvl, line 1906: Why would "I" be unable to do that?
Because the node is not synced?
What can the user do about it?

rvl (approved):

lib/core/src/Cardano/Wallet/Api/Server.hs
jonathanknowles: > Why would "I" be unable to do that?
> Because the node is not synced?
> What can the user do about it?

I'm pretty sure there's not much the user can do about this.

I'm going to create another PR with an enhancement to this error, once I can determine the range of conditions under which it's not possible to determine the current epoch (which is derived from the current slot).

jonathanknowles:
bors retry

lib/core/src/Cardano/Wallet/Api/Server.hs
jonathanknowles: Issue created here: https://github.com/input-output-hk/cardano-wallet/issues/1939

PR #1918, opened 2020-07-17
chore: Support Jormungandr-based network address validation in Daedalus

rhyslbw:
Daedalus Inherits the dependency via cardano-wallet
https://github.com/input-output-hk/daedalus/pull/2041

# Issue Number

https://github.com/input-output-hk/daedalus/pull/2041

# Overview

To provide a frontend address validation utility, cardano-addresses is called from the electron main process, and the result passed back over IPC. This removes the parallel development/maintenance overhead with an implementation in JS.
The dependency is not managed directly for the other variants, so is required to be defined here too for alignment with the packaging concerns.

### DIFF
--- a/.weeder.yaml
+++ b/.weeder.yaml
@@ -74,6 +74,12 @@
         - identifier:
           - C
           - Consumer
+  - section:
+    - name: exe:cardano-wallet-jormungandr
+    - message:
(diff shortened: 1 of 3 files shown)


paweljakubas (approved):
should be ok now

Issue #1888, opened 2020-07-09
Nix eval is slow and memory hungry


rvl:
# Context
 
The Hydra evaluator requires gigabytes of memory to evaluate `release.nix`. This leads to occasional OOM errors on Hydra when multiple builds are being evaluated in parallel.

It is also rather slow -- typically 10-15 minutes!

# Steps to Reproduce

Run the following from a cardano-wallet repo clone:

```
nix-build '<nixpkgs>' -A pkgs.hydra-unstable -o hydra
time ./hydra/bin/hydra-eval-jobs \
  --option allowed-uris "https://github.com/NixOS https://github.com/input-output-hk" \
  --arg supportedSystems '[ "x86_64-linux" ]' \
  -I . release.nix | tee eval.json
```

## Expected behavior

Builds start quickly.

## Actual behavior
 
Builds start slowly.

---

# Resolution 

1. #2300
2. #2301
3. #2313 - Remove jormungandr - less dependencies to evaluate

---

# QA 

1. Eval times are shown here: https://hydra.iohk.io/jobset/Cardano/cardano-wallet#tabs-evaluations

rvl:
Eval times are now about 1/3 of what they were.
Still not great, but closing this for the time being.

PR #1922, opened 2020-07-17
bors: Remove ci/hydra-eval from required statuses

rvl:
### Issue Number

Partial workaround for #1888.

### Overview

Recently we've seen several Bors merge failures from `ci/hydra-eval` that terminate with

    error: unable to fork: Cannot allocate memory

This works around the issue by removing `ci/hydra-eval` from statuses required for Bors to merge.

This status will instead be added as a requirement to the GitHub branch protection rules of this repo.

If the Hydra eval fails the first time around for the Bors merge commit, Bors won't fail because eval isn't required for Bors anymore. Hydra will automatically fix the eval, likely on the second pass, and then build the job at which point Bors will get the build notification and merge assuming the build is fine (which it should be if the PR CI status is already passed).

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -1,6 +1,5 @@
 status = [
   "buildkite/cardano-wallet",
-  "ci/hydra-eval",
   "ci/hydra-build:required",
 ]
 timeout_sec = 7200


Anviking (approved):
🆗

paweljakubas (approved):

KtorZ:
bors retry

Issue #1920, opened 2020-07-17
Hard fork integration tests


rvl:
# Context

ADP-356

With the hard fork combinator, we now have a wallet that must work in both byron and shelley eras.

Therefore we should run our integration tests during both eras.

# Decision

1. Adjust the local cardano cluster setup so that it runs in dual era mode and is able to hard fork.
2. Modify cardano-wallet-shelley:test:integration API specs so that they are run initially in byron mode.
3. Command the cluster to complete a hard fork in the local cluster as a test setup step for...
4. Running the integration API specs again.

# Acceptance Criteria

1. Integration test suide API specs must pass in both byron and shelley era on a local cardano cluster.

---

# Development 

- #1924 test cluster setup, which was merged into #1909 

# QA 

- With #1909, all shelley integration tests are run on Byron;Shelley _after_ the fork.
- We can no longer add faucet funds in the Shelley-genesis. After the fork we manually construct txs from byron faucets to all faucet address defined [here](https://github.com/input-output-hk/cardano-wallet/blob/71c0b8fd10aaf7e09c62c42571780f7676381584/lib/core-integration/src/Test/Integration/Faucet.hs#L1472-L1499) 
- I think Matthias disabled two flaky tests relating to hardware wallet address listing
- Relevant test setup code: https://github.com/input-output-hk/cardano-wallet/blob/71c0b8fd10aaf7e09c62c42571780f7676381584/lib/shelley/test/integration/Main.hs#L221-L242
- We don't run any tests _before_ the fork. We are probably going to release very close to the fork anyway.

piotr-iohk:
lgtm

PR #1924, opened 2020-07-17
Integration test cluster for Byron;Shelley

Anviking:
# Issue Number

#1920 

# Overview

- [x] Add support for Byron;Shelley integration test cluster
- [x] Adjusted configs to work with the new version of cardano-node
- [x] withCluster now takes three actions:
      - onByron
      - onFork (when in shelley. We can send money to the shelley faucets here)
      - onClusterStart (when the stake pools have fully started)
- [x] CLI faucets are now byron facuets, and are generated automatically. This is both easier and more elegant.
- [x] The underlying logic of withCluster was kept somewhat flexible for easier experimentation.
- [x] **Todo:** For the Shelley integration tests to work, we need to send funds to the shelley faucet addresses, as `initialFunds` doesn't work when entering shelley through a hard-fork.
- [x] **Todo** clean up faucet config/code, and make sure special wallets are included

# Comments

```haskell
$ stack test cardano-wallet-shelley:integration
```


Anviking:
@rvl @KtorZ I have rebased this, and the integration tests are starting, and some passing.

I'll ping/tag for review when the the final todo is done, unless you think there will be merge conflicts and we should merge this.

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -452,49 +549,76 @@ withBFTNode
...
+        let shelleyKeys x =
+                    x { nodeOpCertFile = Just opCert
+                      , nodeKesKeyFile = Just kesPrv
+                      , nodeVrfKeyFile = Just vrfPrv
+                      }
+
+        let makeModeSpecific x = case mode of
+                 RunByronShelley _ -> byronKeys . shelleyKeys $ x
+                 RunByron -> byronKeys x
+                 RunShelley -> shelleyKeys x
```
KtorZ: Any particular reason to try supporting all three modes? Wouldn't it be easier to just support the Byron/Shelley ?

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -452,49 +549,76 @@ withBFTNode
...
+                , nodeVrfKeyFile = Nothing
                 , nodePort = Just (NodePort port)
                 , nodeLoggingHostname = Just name
                 }
 
         withCardanoNodeProcess tr name cfg $ \(CardanoNodeConn socket) -> do
-            setEnv "CARDANO_NODE_SOCKET_PATH" socket
-            (rawTx, faucetPrv) <- prepareKeyRegistration tr dir
-            tx <- signTx tr dir rawTx [faucetPrv]
-            submitTx tr "pre-registered stake key" tx
```
KtorZ, line 490: This is important and shouldn't be removed. I get that it can only happen after the hard-fork, but I don't see it moved at the relevant place.

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -452,49 +549,76 @@ withBFTNode
...
     -> NodeParams
     -- ^ Parameters used to generate config files.
     -> (FilePath -> Block -> (NetworkParameters, NodeVersionData) -> IO a)
     -- ^ Callback function with genesis parameters
     -> IO a
-withBFTNode tr baseDir (NodeParams severity systemStart (port, peers)) action =
+withBFTNode
+    tr
+    baseDir
+    i
```
KtorZ: I don't understand why this index is added. This suggests there could be several BFT nodes, but it it's not the case.

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -965,7 +1217,16 @@ submitTx tr name signedTx = do
     void $ cliRetry tr ("Submitting transaction for " ++ name)
         [ "shelley", "transaction", "submit"
         , "--tx-file", signedTx
-        , "--mainnet"
+        , "--mainnet", "--cardano-mode"
+        ]
+
+-- | Submit a transaction through a running node.
+submitTxNoRetry :: Tracer IO ClusterLog -> FilePath -> IO ()
+submitTxNoRetry tr signedTx = do
```
KtorZ: What's the use for this function? Why would we want to drop the retry behavior?

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
Anviking: Yes, you're right.

I mentioned it in the PR description as a remaining todo:
> Todo clean up faucet config/code, and make sure special wallets are included

Edit: Ah, right, I only thought about the mnemonic. This also (should) exist, yes.

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
Anviking: Yes, not needed anymore 👍

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
Anviking: I hoped it could let us still run the integration tests using a pure shelley chain (as an option), but that didn't work. So yeah, we can remove it.

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
Anviking: 🤔 Yeah, you're right.  `submitTx` works fine, and we should prefer it to this.

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -362,13 +424,22 @@ newtype PoolConfig = PoolConfig
...
+
 -- | Execute an action after starting a cluster of stake pools. The cluster also
 -- contains a single BFT node that is pre-configured with keys available in the
 -- test data.
 --
 -- This BFT node is essential in order to bootstrap the chain and allow
 -- registering pools. Passing `0` as a number of pool will simply start a single
 -- BFT node.
+--
+-- The callback actions are not guaranteed to use the same node.
```
KtorZ, line 392: This comment looks scary. It's quite important that we use the relay node for sending transactions, otherwise they might be rolled back and screw the entire test setup.

Anviking: The existing pool setup are using the BFT node to setup the stake pools.

As the PR is now, the BFT would also be used to transfer funds to the shelley faucet addresses.

Are you sure that is a problem? I think the code is looks slightly nicer if we don't have to start the relay until the end, as we used to.

Anviking:
I'll want to squash appropriately at some point, but not yet.

lib/shelley/test/integration/Main.hs
```diff
@@ -220,25 +226,42 @@ specWithServer (tr, tracers) = aroundAll withContext . after tearDown
...
+                onByron
+                (afterFork dir)
+                (onClusterStart action dir)
+
+    tr' = contramap MsgCluster tr
+    onByron _ = pure ()
+    afterFork dir _ = do
+        let encodeAddr = T.unpack . encodeAddress @'Mainnet
+        let addresses = map (first encodeAddr) shelleyIntegrationTestFunds
+        sendFaucetFundsTo stdoutTextTracer dir addresses
```
Anviking, line 218: The setup is sprinkled across integration/Main.hs, Faucet.hs, and Launch.hs, which feels a bit weird. Maybe we can resolve that sometime.

lib/shelley/test/data/cardano-node-shelley/shelley-genesis.yaml
```diff
@@ -0,0 +1,63 @@
+# IMPORTANT NOTES ABOUT THIS FILE
+#
+# a) cardano-node does not parse 'yaml' but only 'json'. We use yaml as a nicer/simpler format
+#    and convert it as json when generating the underlying configuration for the node.
+#
+# b) the `systemStart` is hard-coded here to please the parser but is replaced dynamically
+#    by the same code generating the final node configuration for integration.
+#
+# c) initialFunds are defined as a list of singleton, and not as a key-value map like in the
```
Anviking: Hm.🤔 We don't really control the order when we fill the shelley faucets now. Maybe this might cause some failures @KtorZ .

Maybe we have to make sure to always keep funds for a sequential wallet in a single tx (but we may have as many as 200 addresses to fill for a single wallet).

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
KtorZ: It shouldn't be a problem indeed since there won't be any stake pools at the time the faucet funds are transferred to cause any rollback.

lib/shelley/test/data/cardano-node-shelley/shelley-genesis.yaml
KtorZ: This will be a problem for big wallets only, which might explain why the scenario is currently failing. I thought initial funds in the shelley genesis were ignored though?

Anviking: > I thought initial funds in the shelley genesis were ignored though?

Yes. 

(I came to think about it when reading the outdated documentation comment. Now updated.)

rvl:
I'm going to merge into #1909 and we can continue there...

Issue #1926, opened 2020-07-18
Minimum utxo value should not influence inputs


paweljakubas:
# Context

We have #1916 dealing with outputs in the context of minimum value utxo. Also in #1894 we make dust determined by it. 
The min value rule in the ledger is simple: new outputs must be bigger than the minimum.
The min value can be changed, and this can lead to existing UTxOs being below the current minimum for new outputs. Those existing UTxOs are still spendable.

# Decision

Revert the dependence of input dusting on minmium UTxO value

# Acceptance Criteria

---

# Development 

# QA

piotr-iohk:
Do I understand correctly that this change will be discarded? @paweljakubas, @KtorZ

KtorZ:
This issue is to be discarded indeed. The dust threshold has no influence on the inputs but operate on the change outputs.

piotr-iohk:
Closing.

PR #1927, opened 2020-07-18
unwind input's dust dependence on minimum utxo value

paweljakubas:
# Issue Number

#1926 

# Overview

- [x] I have make sure input's dust does not depend on minimum utxo value

# Comments


paweljakubas:
bors try

rvl (approved):

Issue #1915, opened 2020-07-16
Reward history: minWithdrawal = 0 behaves the same as minWithdrawal = 1


piotr-iohk:
# Context
 
https://github.com/input-output-hk/cardano-wallet/issues/1882

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | `2020.7.6 (git revision: 53805cc60e6264054c501da85a5d05ab8df1ae10)` |
| Platform     | Linux all          |
| Installation | from source            |

# Steps to Reproduce

1. Have a wallet with some transactions with withdrawals on it.
2. list transactions with `minWithdrawal = 1` :point_right:  all transaction with rewards are listed
3. list transactions with `minWithdrawal = 0`

## Expected behavior

Either all transactions should be listed (with and without withdrawals) or minWithdrawal parameter value should be restricted to => 1.

## Actual behavior
 
Also all transactions *with withdrawals* are listed, same as with minWithdrawal = 1.
This is small "logical inconherence", because in fact regular transactions (without withdrawals), also have minWithdrawal = 0.

---

# Resolution 

---

# QA

piotr-iohk:
lgtm

PR #1928, opened 2020-07-18
change minimum withdrawal to 1

paweljakubas:
# Issue Number

#1915 

# Overview

- [x] I have changed minimum withdrawal to 1. it is optional parameter so 1 is minimal reasonable value. not specifying this parameter will return all txs (with both withdrawals and not). so withdrawal=0 seems to be redundant 
- [x] Added wallet error
- [x] Added integration test illustrating the case

# Comments


piotr-iohk:
This does not seem to be enough. Still `minWithdrawal = 0` is acceptable as a param and behaves the same as `minWithdrawal = 1`.

paweljakubas:
thanks @piotr-iohk  will look further into this soon

paweljakubas:
bors try

piotr-iohk (approved):
LGTM. Added additional test for minWithdrawal = 1.

PR #1931, opened 2020-07-20
Basic integration tests for pool registrations and retirements.

jonathanknowles:
# Issue Number

#1819 

# Overview

This PR provides a first step towards a full suite of integration tests for pool registrations and retirements.

- [x] Adds function `issuePoolRetirementCert` capable of creating Shelley pool retirement certificates.
- [x] Adjusts the initialization of the test pool cluster so that a subset of the pools are configured to retire after different time intervals.
- [x] Adds a basic integration test to verify that the `ListStakePools` API operation eventually shows the correct retirement epochs for the test pools.

# Comments

Future PRs will:

- Simulate the effect of re-registering a pool after a retirement certificate has already been published.
- Verify that it is not possible to join a stake pool that has already retired (see #1913).

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
@@ -593,6 +593,20 @@ spec = do
                             (.> Quantity (unsafeMkPercentage 0))
                     ]
 
+        it "pools have the correct retirement information" $ \ctx -> do
+            eventually "pools have the correct retirement information" $ do
+                r <- listPools ctx arbitraryStake
+                expectResponseCode HTTP.status200 r
+                verify r
+                    [ expectListSize 3
+                    -- At the time of setup, the pools have 1/3 stake each, but
+                    -- this could potentially be changed by other tests. Hence,
+                    -- we try to be forgiving here.
+                    , expectListField 0 #retirement (`shouldBe` Nothing)
+                    , expectListField 1 #retirement (`shouldBe` Nothing)
+                    , expectListField 2 #retirement (`shouldBe` Nothing)
+                    ]
+
         it "eventually has correct margin, cost and pledge" $ \ctx -> do
             eventually "pool worker finds the certificate" $ do
                 r <- listPools ctx arbitraryStake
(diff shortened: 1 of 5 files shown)


jonathanknowles:
bors try

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -593,6 +596,23 @@ spec = do
...
+                expectResponseCode HTTP.status200 r
+                let retirementEpochIs n = \case
+                        Nothing -> False
+                        Just ei -> epochNumber ei == ApiT (EpochNo n)
+                verify r
+                    [ expectListSize 3
+                    , expectListField 0 #retirement
+                        (`shouldBe` Nothing)
+                    , expectListField 1 #retirement
+                        (`shouldSatisfy` retirementEpochIs 10)
```
piotr-iohk: This test passed when I run it alone... but it was not passing when I was running it with a test suite. 

I started to monitor the cluster while the tests were running and my observation is that this pool that is scheduled to retire in epoch 10 disappears from the `list stake pools` at the beginning of epoch 14... :thinking:  so, we're starting with 3 pools being registered in "setup" but since epoch 14 wallet actually shows only 2 pools!

jonathanknowles: @piotr-iohk wrote:

> This test passed when I run it alone... but it was not passing when I was running it with a test suite.
> 
> I started to monitor the cluster while the tests were running and my observation is that this pool that is scheduled to retire in epoch 10 disappears from the `list stake pools` at the beginning of epoch 14... thinking so, we're starting with 3 pools being registered in "setup" but since epoch 14 wallet actually shows only 2 pools!

Thanks for analysing this!

I'm a little surprised that it just disappears from the list, as we don't (yet) actively perform any garbage collection of pools after they retire. If I were to take a guess, the cause would be related to the way we are merging data in order to produce the list of stake pools. I'll put this next on my list for immediate investigation.

For now, I've uploaded a patch (e80b7f5b46bb8d6e94147b6f26c2c86925d092c8) which prevents any stake pools from prematurely retiring during a run of the integration test suite, which should fix this PR.

Later on, in a further PR, I'll add more integration tests to check the behaviour of pools after they retire.

jonathanknowles:
bors try

jonathanknowles:
bors try

jonathanknowles:
bors try

jonathanknowles:
bors try

piotr-iohk (approved):

Issue #1930, opened 2020-07-20
Stakepools are not shown immediately in wallet after they are registered. 


piotr-iohk:
# Context
 https://input-output-rnd.slack.com/archives/GBT05825V/p1595052799003600

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | `2020.7.6 (git revision: cf3847dd08d1c1848c675a1ee94ae12c1d85aadf)` |
| Platform     | all         |
| Installation | Daedalus / From source          |

# Steps to Reproduce

1. Register a stake pool in `epoch` (stake pool registered successfully and visible in the ledger state)
2. Run list stake pools in the wallet.

## Expected behavior

Stake pool that was registered in `epoch` should be immediately visible in the wallet in the same epoch.

## Actual behavior
 
Stake pool is visible in `epoch + 2`

---

# Resolution 

---

# QA

Anviking:
Copying from slack for visibility:
> I believe we should also make sure to, in the case pools have the same non-myopic rewards (they might all have 0 in the first epoch?), then order them randomly based on a user-specific seed. We did this for jormungandr (for the first epoch).
>
> Think this is mentioned in the delegation design spec

even if that might technically be a separate ticket

KtorZ:
This is a necessary limitation until we are done with the retirement certificates story. As a matter of fact, we currently rely on the ledger's state to know whether pools are active or not.

We are processing and discovering pool's registration certificates ourselves, indeed but I am not quite sure about the status of the retirement discovery and handling at the moment? cc @jonathanknowles

jonathanknowles:
@KtorZ wrote:

> This is a necessary limitation until we are done with the retirement certificates story. As a matter of fact, we currently rely on the ledger's state to know whether pools are active or not.
> We are processing and discovering pool's registration certificates ourselves, indeed but I am not quite sure about the status of the retirement discovery and handling at the moment?

I'll try to shed some light on this.

## Current status

When constructing the list of pools to return in `GetStakePools`, we currently rely on **two data sources**:

1. Our database record of _registration_ and _retirement_ certificates.
2. The data returned by local state query (via `getStakeDistribution`).

As for the former (our database record), we update this whenever we process a new block, and **we currently record all registration and retirement certificates**, together with the _order_ in which we see them. So from the perspective of populating the database with the correct data, the work for [ADP-313](https://github.com/input-output-hk/cardano-wallet/milestone/89) is already complete (apart from any further integration test work).

[...]

PR #1936, opened 2020-07-22
Revise data source consolidation for `ListStakePools`.

jonathanknowles:
# Issue Number

#1930 
#1948 

# Overview

The `ListStakePools` endpoint produces a list of `ApiStakePool` records by merging data from (at least) **two sources**:

- DB data: data fetched from the local pool database. The presence of DB data for a pool indicates that the wallet has seen one or more registration certificates for that pool. 
- LSQ data: pool metrics data fetched with local state query. 

This PR makes **two changes** to the set of pools returend by the `ListStakePools` endpoint:

| DB data available? | LSQ data available? | Old Behaviour | New Behaviour |
|---|---|---|---|
| | | | |
| :heavy_check_mark: | :negative_squared_cross_mark: | pool **not** listed | pool **listed** |
| | | | stake pool metrics are assigned a default value of zero |
| | | | |
| :negative_squared_cross_mark:  | :heavy_check_mark: | pool **listed** | pool **not** listed |
| | | the following fields have missing values:<br> - `cost` <br> - `margin` <br> - `pledge` |  the following fields are now required:<br> - `cost` <br> - `margin` <br> - `pledge` |

### DIFF
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
@@ -120,6 +120,7 @@ import Ouroboros.Consensus.Shelley.Protocol
     ( TPraosCrypto )
 
 import qualified Cardano.Wallet.Api.Types as Api
+import qualified Data.Foldable as F
 import qualified Data.Map.Merge.Strict as Map
 import qualified Data.Map.Strict as Map
 import qualified Data.Text as T
@@ -171,14 +172,8 @@ newStakePoolLayer
 
     _knownPools
         :: IO [PoolId]
-    _knownPools = do
-        tip <- getTip
-        let dummyCoin = Coin 0
-        res <- runExceptT $ map fst . Map.toList
-            . combineLsqData <$> stakeDistribution tip dummyCoin
-        case res of
-            Right x -> return x
-            Left _e -> return []
+    _knownPools =
+        F.toList . Map.keysSet <$> liftIO (readPoolDbData db)
 
     _listPools
         :: EpochNo
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
@@ -239,15 +239,15 @@ combineDbAndLsqData
     -> Map PoolId PoolDbData
     -> Map PoolId Api.ApiStakePool
 combineDbAndLsqData sp =
-    Map.merge lsqButNoChain chainButNoLsq bothPresent
+    Map.merge lsqButNoDb dbButNoLsq bothPresent
   where
-    lsqButNoChain = traverseMissing $ \k lsq -> pure $ mkApiPool k lsq Nothing
+    lsqButNoDb = traverseMissing $ \k lsq -> pure $ mkApiPool k lsq Nothing
 
     -- In case our chain following has missed a retirement certificate, we
     -- treat the lsq data as the source of truth, and dropMissing here.
-    chainButNoLsq = dropMissing
+    dbButNoLsq = dropMissing
 
-    bothPresent = zipWithMatched  $ \k lsq chain -> mkApiPool k lsq (Just chain)
+    bothPresent = zipWithMatched  $ \k lsq db -> mkApiPool k lsq (Just db)
 
     mkApiPool
         :: PoolId
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
@@ -242,11 +242,23 @@ combineDbAndLsqData
 combineDbAndLsqData sp =
     Map.merge lsqButNoDb dbButNoLsq bothPresent
   where
-    lsqButNoDb = traverseMissing $ \k lsq -> pure $ mkApiPool k lsq Nothing
+    lsqButNoDb = traverseMissing $ \k lsq ->
+        pure $ mkApiPool k lsq Nothing
 
-    -- In case our chain following has missed a retirement certificate, we
-    -- treat the lsq data as the source of truth, and dropMissing here.
-    dbButNoLsq = dropMissing
+    -- When a pool is registered (with a registration certificate) but not
+    -- currently active (and therefore not causing pool metrics data to be
+    -- available over local state query), we use a default value of /zero/
+    -- for all stake pool metric values so that the pool can still be
+    -- included in the list of all known stake pools:
+    --
+    dbButNoLsq = traverseMissing $ \k db ->
+        pure $ mkApiPool k lsqDefault $ Just db
+      where
+        lsqDefault = PoolLsqData
+            { nonMyopicMemberRewards = minBound
+            , relativeStake = minBound
+            , saturation = 0
+            }
 
     bothPresent = zipWithMatched $ \k lsq db -> mkApiPool k lsq (Just db)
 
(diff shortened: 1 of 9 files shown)


jonathanknowles:
bors try

jonathanknowles:
> # Build failed
> [ci/hydra-build:required](https://hydra.iohk.io/build/3564934)

If the line position indicated in the [log](https://hydra.iohk.io/build/3564934/nixlog/1/tail) is reliable, then the following expectation failed:
https://github.com/input-output-hk/cardano-wallet/blob/2f2cf55296c3b6a988ecf267b1045c19fdce0404/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs#L610-L611

The value that actually appeared was:
```hs
Just (ApiEpochInfo {epochNumber = ApiT {getApiT = EpochNo {unEpochNo = 1000000}}, epochStartTime = 2021-10-28 04:39:02.004223 UTC})")
```
In other words, `1_000_000` appeared when `1_000` was expected.

These two values come from:
https://github.com/input-output-hk/cardano-wallet/blob/2f2cf55296c3b6a988ecf267b1045c19fdce0404/lib/shelley/test/integration/Main.hs#L181-L186

Perhaps it's possible that retirement certificates for different test pools have somehow been transposed during stake pool cluster initialization.

jonathanknowles:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -198,15 +198,27 @@ combineDbAndLsqData
...
+    -- currently active (and therefore not causing pool metrics data to be
+    -- available over local state query), we use a default value of /zero/
+    -- for all stake pool metric values so that the pool can still be
+    -- included in the list of all known stake pools:
+    --
+    dbButNoLsq = traverseMissing $ \k db ->
+        pure $ mkApiPool k lsqDefault $ Just db
+      where
+        lsqDefault = PoolLsqData
+            { nonMyopicMemberRewards = minBound
```
KtorZ, line 253: This needs to be checked / justified against the specs. I am not sure that `0` would be a significant value for new pools.

jonathanknowles:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -208,15 +208,27 @@ combineDbAndLsqData
...
 combineDbAndLsqData sp =
-    Map.merge lsqButNoChain chainButNoLsq bothPresent
+    Map.merge lsqButNoDb dbButNoLsq bothPresent
   where
-    lsqButNoChain = traverseMissing $ \k lsq -> pure $ mkApiPool k lsq Nothing
-
-    -- In case our chain following has missed a retirement certificate, we
-    -- treat the lsq data as the source of truth, and dropMissing here.
-    chainButNoLsq = dropMissing
+    lsqButNoDb = traverseMissing $ \k lsq ->
```
Anviking: Should we not rely 100% on the DB now, and `dropMissing` on `lsqButNoDb`?

Anviking:
Looking good (there are more pools at least):
Before:
Skärmavbild 2020-07-26 kl  23 28 07

With this PR:
Skärmavbild 2020-07-26 kl  23 28 25

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
jonathanknowles: > Should we not rely 100% on the DB now, and `dropMissing` on `lsqButNoDb`?

We could indeed do that. If we did, we'd be able to guarantee the presence of the following fields for each `ApiStakePool` object we produce:

- `cost`
- `margin`
- `pledge`

(See: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/Api/Types.hs#L406-L408)

We'd then be able to make the following simplification:
```patch
@@ -403,9 +403,9 @@ data ApiStakePool = ApiStakePool
     { id :: !(ApiT PoolId)
     , metrics :: !ApiStakePoolMetrics
     , metadata :: !(Maybe (ApiT StakePoolMetadata))
-    , cost :: !(Maybe (Quantity "lovelace" Natural))
-    , margin :: !(Maybe (Quantity "percent" Percentage))
-    , pledge :: !(Maybe (Quantity "lovelace" Natural))
+    , cost :: !(Quantity "lovelace" Natural)
+    , margin :: !(Quantity "percent" Percentage)
+    , pledge :: !(Quantity "lovelace" Natural)
     , retirement :: !(Maybe ApiEpochInfo)
```

jonathanknowles: > Should we not rely 100% on the DB now, and dropMissing on lsqButNoDb?

Fixed in 457ad036c1e0f7f93e637c8453e6ec5f49c9153c.

jonathanknowles: > We'd then be able to make the following simplification:
> ```patch
> @@ -403,9 +403,9 @@ data ApiStakePool = ApiStakePool
>      { id :: !(ApiT PoolId)
>      , metrics :: !ApiStakePoolMetrics
>      , metadata :: !(Maybe (ApiT StakePoolMetadata))
> -    , cost :: !(Maybe (Quantity "lovelace" Natural))
> -    , margin :: !(Maybe (Quantity "percent" Percentage))
> -    , pledge :: !(Maybe (Quantity "lovelace" Natural))
> +    , cost :: !(Quantity "lovelace" Natural)
> +    , margin :: !(Quantity "percent" Percentage)
> +    , pledge :: !(Quantity "lovelace" Natural)
>      , retirement :: !(Maybe ApiEpochInfo)
>     } deriving (Eq, Generic, Show)
> ```

Fixed in:
- efbfd59beee7bc3feac20c0d3d365c073a5c45b4 (Simplify field types for `ApiStakePool`)
- 4a0274329190c6c8fba9a64a2d063028c3782c14 (Mark fields as required in specification for `ApiStakePool`)
- bec56c630fa48a0614b39f8ce2b8b56ed146eb46 (Regenerate golden test data for `ApiStakePool`)

Anviking (approved):

jonathanknowles:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
KtorZ: @jonathanknowles after checking, the ledger team suggests that (given the time constraints) we use the average non myopic member rewards of the top `k` pools (where `k` is the desired number of pools).

Anviking:
We need to update [`knownPools`](https://github.com/input-output-hk/cardano-wallet/blob/master/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs#L174-L180)
Otherwise the new pools are unjoinable.

jonathanknowles:
@Anviking wrote:

> We need to update [`knownPools`](https://github.com/input-output-hk/cardano-wallet/blob/master/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs#L174-L180)
> Otherwise the new pools are unjoinable.

You're right. The result of `knownPools` is passed as an argument to `joinStakePool`, which checks the desired pool against the known set.

But `knownPools` currently returns the result of what is available over local state query, and the desired pool might not yet be included in this set (even though a certificate is already available).

We should probably change this to return the set of pools recorded in the database.

jonathanknowles:
> The result of `knownPools` is passed as an argument to `joinStakePool`, which checks the desired pool against the known set.
> But `knownPools` currently returns the result of what is available over local state query, and the desired pool might not yet be included in this set (even though a certificate is already available).
> We should probably change this to return the set of pools recorded in the database.

Fixed in 1ad8451cbb5155578bd52287fc10334e91de2e42.

jonathanknowles:
bors try

jonathanknowles:
bors try

KtorZ:
I'll do the average non-myopic reward on top of another PR.

Issue #1940, opened 2020-07-23
`STAKE_POOLS_LIST` tests fail due pools appearing out of order.


jonathanknowles:
# Context

As part of the stake pool integration tests, we verify that pools within the test cluster have the  retirement epochs that we expect.

However, this test relies on pools being listed in a certain order, which turns out to be a condition that we can't currently rely on.

# Test Case

`STAKE_POOLS_LIST_01 / pools have the correct retirement information`

# Failure / Counter-example

https://hydra.iohk.io/build/3564934

According to the [log](https://hydra.iohk.io/build/3564934/nixlog/1/tail), the following expectation failed:

https://github.com/input-output-hk/cardano-wallet/blob/2f2cf55296c3b6a988ecf267b1045c19fdce0404/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs#L610-L611

The value that _actually_ appeared was:
```hs
Just (ApiEpochInfo {epochNumber = ApiT {getApiT = EpochNo {unEpochNo = 1000000}}, epochStartTime = 2021-10-28 04:39:02.004223 UTC})")
```

In other words, `1_000_000` appeared when `1_000` was expected.

These two values are defined here:
https://github.com/input-output-hk/cardano-wallet/blob/2f2cf55296c3b6a988ecf267b1045c19fdce0404/lib/shelley/test/integration/Main.hs#L181-L186

This appears to be a simple out-of-order issue.

# Suggested Resolution 

Make the test agnostic to the order in which stake pools appear.

# QA 

https://github.com/input-output-hk/cardano-wallet/pull/1941

piotr-iohk:
ok

PR #1941, opened 2020-07-23
Ignore pool order in `STAKE_POOLS_LIST_01`.

jonathanknowles:
# Issue Number

#1940 

# Overview

This PR changes the API `STAKE_POOLS_LIST_01` test to **ignore** the order in which pools are listed by the `ListStakePools` API operation.

It prevents spurious failures caused by pools being listed in unexpected orders.

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
@@ -607,11 +607,12 @@ spec = do
             eventually "pools have the correct retirement information" $ do
                 response <- listPools ctx arbitraryStake
                 expectResponseCode HTTP.status200 response
+
                 let actualRetirementEpochs = response
                         & snd
                         & either (error . show) Prelude.id
                         & fmap (view #retirement)
-                        & fmap (fmap (getApiT . view #epochNumber))
+                        & fmap (fmap (view (#epochNumber . #getApiT)))
                         & Set.fromList
                 actualRetirementEpochs `shouldBe` expectedRetirementEpochs
 
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
@@ -615,9 +615,8 @@ spec = do
                         .
                         view #retirement
 
-                let actualRetirementEpochs = response
-                        & snd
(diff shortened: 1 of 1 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -597,22 +597,24 @@ spec = do
...
 
+            eventually "pools have the correct retirement information" $ do
+                response <- listPools ctx arbitraryStake
+                expectResponseCode HTTP.status200 response
+                let actualRetirementEpochs = response
+                        & snd
+                        & either (error . show) Prelude.id
+                        & fmap (view #retirement)
+                        & fmap (fmap (getApiT . view #epochNumber))
+                        & Set.fromList
```
KtorZ, line 621: Suggestion: use `getFromResponse` and `#getApitT` instead:

```hs
retirements <- fmap (view (#epochNumber . #getApiT)) <$> getFromResponse #retirement response
```

KtorZ (approved):

paweljakubas (approved):
nice catch!

KtorZ:
bors retry

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -597,22 +598,29 @@ spec = do
...
+                expectResponseCode HTTP.status200 response
+
+                let getRetirementEpoch :: ApiStakePool -> Maybe EpochNo
+                    getRetirementEpoch =
+                        fmap (view (#epochNumber . #getApiT))
+                        .
+                        view #retirement
+
+                let actualRetirementEpochs =
+                        getFromResponse Prelude.id response
```
KtorZ, line 619: by the way: getFromResponse expects a lens as a first argument, so `getFromResponse #retirement` works just fine.

jonathanknowles: > by the way: getFromResponse expects a lens as a first argument, so `getFromResponse #retirement` works just fine.

We're dealing with a list here though, so the expression `getFromResponse #retirement response` produces the error:

```
The type [ApiStakePool] does not contain a field named 'retirement'
```

KtorZ: Ah! It's a list of ApiStakePool, indeed.

Issue #1923, opened 2020-07-17
Transaction with withdrawRewards=true shows misleading amount sent


piotr-iohk:
# Context

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | `2020.7.6 (git revision: cf3847dd08d1c1848c675a1ee94ae12c1d85aadf)` |
| Platform     | Linux, all         |
| Installation | Daedalus           |

# Steps to Reproduce

On shelley testnet:
0. A wallet, total balance = 1,1M ada, rewards = 162,5k
1. Sent 200k from that wallet with `withdrawRewards=true`

## Expected behavior

Amount sent should be reported as 200k + fees.

## Actual behavior
 
Amount sent on that transaction shows = 37,5k, which means (I suppose) part of the transaction has been financed from the reward account (which had 162,5k).

This is misleading, as the amount actually sent is 200k + fees.

![Screenshot from 2020-07-17 15-05-00](https://user-images.githubusercontent.com/42900201/87791323-594f7e00-c842-11ea-8447-8b4c3280519e.png)

Yes, actually 162476375448+37523798985 = 200000174433, that is 200k + fees.
That value should be reported as amount sent for the transaction.

---

# Resolution 

---

# QA

piotr-iohk:
This translates basically to wrong sent amount being displayed in Daedalus in case of transactions with withdrawals:

![Screenshot from 2020-07-17 14-45-25](https://user-images.githubusercontent.com/42900201/87849892-75a7f500-c8ec-11ea-8ea7-c8ca6963eac0.png)
![Screenshot from 2020-07-17 14-51-47](https://user-images.githubusercontent.com/42900201/87849897-79d41280-c8ec-11ea-81d5-00552de9241a.png)

piotr-iohk:
lgtm.

PR #1942, opened 2020-07-23
Account withdrawals in amount

paweljakubas:
# Issue Number

#1923 

# Overview

- [x] I have changed `mkTxMeta`
- [x] I have changed `applyTx` in `Primitive.Model`
- [x] I have extended test to test consistency between amount in Pending and InLedger after tx withWithdrawal=true is sent

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -357,6 +357,8 @@ import Data.List
     ( scanl' )
 import Data.List.NonEmpty
     ( NonEmpty )
+import Data.Map.Strict
+    ( Map )
 import Data.Maybe
     ( mapMaybe )
 import Data.Quantity
@@ -388,7 +390,7 @@ import qualified Cardano.Wallet.Primitive.CoinSelection.Random as CoinSelection
 import qualified Cardano.Wallet.Primitive.Types as W
 import qualified Data.List as L
 import qualified Data.List.NonEmpty as NE
-import qualified Data.Map as Map
+import qualified Data.Map.Strict as Map
 import qualified Data.Set as Set
 import qualified Data.Text as T
 import qualified Data.Vector as V
@@ -1393,6 +1395,7 @@ signPayment
         , HasDBLayer s k ctx
         , HasNetworkLayer t ctx
         , IsOwned s k
+        , IsOurs s ChimericAccount
         , GenChange s
         , HardDerivation k
         , Bounded (Index (AddressIndexDerivationType k) 'AddressK)
@@ -1419,7 +1422,7 @@ signPayment ctx wid argGenChange pwd cs = db & \DBLayer{..} -> do
             (tx, sealedTx) <- withExceptT ErrSignPaymentMkTx $ ExceptT $ pure $
                 mkStdTx tl (rewardAcnt, pwdP) keyFrom (nodeTip ^. #slotNo) cs'
 
-            (time, meta) <- liftIO $ mkTxMeta ti (currentTip cp) s' cs'
+            (time, meta) <- liftIO $ mkTxMeta ti (tx ^. #withdrawals) (currentTip cp) s' cs'
             return (tx, meta, time, sealedTx)
   where
     ti :: TimeInterpreter IO
@@ -1433,6 +1436,7 @@ signTx
     :: forall ctx s t k.
         ( HasTransactionLayer t k ctx
         , HasDBLayer s k ctx
+        , IsOurs s ChimericAccount
         , HasNetworkLayer t ctx
         , IsOwned s k
         , HardDerivation k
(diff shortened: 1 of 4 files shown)


paweljakubas:
bors try

KtorZ (approved):
LGTM. Shall we adjust one of the integration test perhaps to show this behavior? We should have caught that in the integration test suite in the first place.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1694,17 +1694,26 @@ mkApiTransaction ti txid ins outs ws (meta, timestamp) setTimeReference = do
...
         }
 
+    toApiWithdrawal :: [ApiWithdrawal n]
+    toApiWithdrawal = mkApiWithdrawal @n <$> Map.toList ws
+
+    amountWithWithdrawals :: Quantity "lovelace" Natural
+    amountWithWithdrawals =
+        let amtFromMeta = meta ^. #amount
+            amtFromWithdrawal = map (\(ApiWithdrawal _ amt) -> amt) toApiWithdrawal
+        in Quantity $ sum $ getQuantity <$>  (amtFromMeta : amtFromWithdrawal)
```
KtorZ: On a second thought, what about doing this calculation earlier when the `meta ^. #amount` is actually first created? Otherwise, we get to carry this inconsistency everywhere instead of being solved in a more holistic manner.

KtorZ (changes requested):
After discussing, I realized that needs to be fixed at a different level. There's actually a bug were the withdrawal amount is counted twice when creating a transaction meta from a transaction we send, but like an input!

And, it's also completely ignored from the model that discovers transactions (so if transactions are fired from another instance of the wallet for instance).

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1614,6 +1623,10 @@ mkTxMeta interpretTime blockHeader wState cs =
         if fst (isOurs addr wState)
         then Just (fromIntegral val)
         else Nothing
+    possibleWithdrawals =
+        map (fromIntegral . getCoin) $
+        Map.elems $
+        Map.filterWithKey (\chAcct _ -> fst $ isOurs chAcct wState) chimericAccts
```
KtorZ, line 1629: :+1:

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -376,7 +377,8 @@ prefilterBlock b u0 = runState $ do
         ourU <- state $ utxoOurs tx
         let ourIns = Set.fromList (inputs tx) `Set.intersection` dom (u <> ourU)
         let u' = (u <> ourU) `excluding` ourIns
-        let received = fromIntegral @_ @Integer $ balance ourU
+        let withdrawalAmt = sum $ map (fromIntegral . getCoin) $ Map.elems $ withdrawals tx
+        let received = fromIntegral @_ @Integer $ (withdrawalAmt + balance ourU)
```
KtorZ: withdrawals here also need to be filtered based on whether they're ours or not. Unfortunately, that's the same bits of logic in two different places.

paweljakubas: addressed in https://github.com/input-output-hk/cardano-wallet/pull/1942/commits/f200ae0bb1c9d9e7184a3a073f0b85bb6524bf1a

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -368,15 +377,17 @@ prefilterBlock b u0 = runState $ do
...
     applyTx (!txs, !u) tx = do
         ourU <- state $ utxoOurs tx
         let ourIns = Set.fromList (inputs tx) `Set.intersection` dom (u <> ourU)
         let u' = (u <> ourU) `excluding` ourIns
-        let received = fromIntegral @_ @Integer $ balance ourU
+        ourWithdrawals <- mapMaybeM ourChimericAccount $ Map.toList $ withdrawals tx
+        let received =
+                fromIntegral @_ @Integer $ (sum ourWithdrawals) + balance ourU
         let spent = fromIntegral @_ @Integer $ balance (u `restrictedBy` ourIns)
         let amt = fromIntegral $ abs (received - spent)
```
KtorZ, line 392: Okay, now this is only missing the `reclaim` from the input balance when there are some :+1:

KtorZ: Although... the reclaims aren't explicit in the transaction. There's no "input" for them. You know that there are some by looking at the _fees_ and the delta between input and output :thinking: ...

So, right now, if we discover a transaction with a reclaimed deposit, it may have a negative amount (because the "spent" will likely be bigger than the total "received").... This is a bit out of scope for this PR however, so may you just open a ticket about taking reclaims into account in the primitive model when prefiltering blocks?

paweljakubas:
bors try

lib/core/src/Cardano/Wallet/Primitive/Model.hs
paweljakubas: https://github.com/input-output-hk/cardano-wallet/issues/1944

KtorZ (approved):

paweljakubas:
closing in favor of #1946

PR #1943, opened 2020-07-23
Improvements to slotting conversions / TimeInterpreter

### LINKED ISSUE #1960: GET http://localhost:8090/v2/network/information returns 500 on mainnet while wallet is syncing through Byron era.
# Context
 
 

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | `2020.7.6 (git revision: 93c177b179fc4fc500858ebfbbf4ed7ecfc2c6bc)` |
| Platform     | Linux       |
| Installation | From source            |

# Steps to Reproduce

1. Start cardano-wallet-shelley on mainnet. (https://hydra.iohk.io/build/3627080/download/1/index.html)
2. While it is syncing (and also after it is synced, but still before HF) query 
```
GET http://localhost:8090/v2/network/information 
```

## Expected behavior

Information is available.

## Actual behavior
```
The response was:
Code = 500,
Message = Something went wrong
```
in the log there is:

```
[cardano-wallet.network:Critical:206] [2020-07-27 13:54:16.34 UTC] Time interpreter queried past the horizon. This should not have happened.
Query is:
sqBind qBind QBind QBind QBind QAbsToRelEpoch EpochNo {unEpochNo = 208}
tHorizon [("runQuery",SrcLocEras are:
EraSummary {eraStart = Bound {boundTime = RelativeTime {getRelativeTime = 0s}, boundSlot = SlotNo {unSlotNo = 0}, boundEpoch = EpochNo {unEpochNo = 0}}, eraEnd = EraEnd (Bound {boundTime = RelativeTime {getRelativeTime = 89856000s}, boundSlot = SlotNo {unSlotNo = 4492800}, boundEpoch = EpochNo {unEpochNo = 208}}), eraParams = EraParams {eraEpochSize = EpochSize {unEpochSize = 21600}, eraSlotLength = SlotLength {getSlotLength = 20s}, eraSafeZone = StandardSafeZone 4320 NoLowerBound}}
  runQuery, called at src/Ouroboros/Consensus/HardFork/History/Qry.hs:302:44 in ouroboros-consensus-0.1.0.0-GJeJBwzkEl7AcixwXFgXER:Ouroboros.Consensus.HardFork.History.Qry
  interpretQuery, called at src/Cardano/Wallet/Primitive/Slotting.hs:309:26 in cardano-wallet-core-2020.7.6-AamOEQSEhy42ld6T94mTEW:Cardano.Wallet.Primitive.Slotting
  runQuery, called at src/Cardano/Wallet/Primitive/Slotting.hs:275:27 in cardano-wallet-core-2020.7.6-AamOEQSEhy42ld6T94mTEW:Cardano.Wallet.Primitive.Slotting
  mkTimeInterpreter, called at src/Cardano/Wallet/Shelley/Network.hs:456:60 in cardano-wallet-shelley-2020.7.6-2hcgUDLMQFX1xdHvgaYBva:Cardano.Wallet.Shelley.Network
  _timeInterpreterQuery, called at src/Cardano/Wallet/Shelley/Network.hs:307:33 in cardano-wallet-shelley-2020.7.6-2hcgUDLMQFX1xdHvgaYBva:Cardano.Wallet.Shelley.Network
  timeInterpreter, called at src/Cardano/Wallet/Network.hs:153:7 in cardano-wallet-core-2020.7.6-AamOEQSEhy42ld6T94mTEW:Cardano.Wallet.Network
```

---

# Resolution 

Two separate problems:
1. Even when synced, in Byron, it returns 500
2. When un-synced it returns 500

Explanation:
1. We use `startTime =<< firstSlotInEpoch (e + 1)` to get the time of the next epoch. This is _just_ outside the horizon.
2. We throw if we can't get the `network_tip` or `next_epoch`.

Solution:
1. Get the end time of the current epoch, instead of the start time of the next one
2. Don't throw. Make `network_tip` and `next_epoch` optional

---

# QA 
Tested manually on mainnet.

When not in sync:
```bash
$ cardano-wallet-shelley network information
Ok.
{
    "node_tip": {
        "height": {
            "quantity": 2298589,
            "unit": "block"
        },
        "epoch_number": 106,
        "slot_number": 10480
    },
    "sync_progress": {
        "status": "syncing",
        "progress": {
            "quantity": 51.26,
            "unit": "percent"
        }
    }
}
```

When in sync:
```
{
    "network_tip": {
        "epoch_number": 207,
        "slot_number": 15728
    },
    "node_tip": {
        "height": {
            "quantity": 4484639,
            "unit": "block"
        },
        "epoch_number": 207,
        "slot_number": 15728
    },
    "sync_progress": {
        "status": "ready"
    },
    "next_epoch": {
        "epoch_number": 208
    }
```

Anviking:
# Issue Number

#1869 / #1960 

# Overview

- [x] Block instead of using a `singleEraInterpreter` if we don't yet have a proper `TimeInterpreter`.
- [x] Remove TODO; explain why delegation time conversion never should fail
- [x] Replace `TimeInterpreter IO` with `TimeInterpreter (ExceptT IO ErrPastHorizon)`
- [x] Add, test and use `endTimeOfEpoch e` instead of `timeOf =<< firstSlotInEpoch (e+1)`. This is needed to stay inside the forecast range when showing `next_epoch` when the node is in-sync, in Byron. Fixes #1960 
- [x] Make `next_epoch` and `network_tip` optional in the API. When the node is still syncing the Byron chain, it cannot know any slotting info close to the current time. Solving related problem to #1960 
- [x] Revise the names of the constructors of `TimeInterpreter`

# Comments

- I thought queries in Shelley could fail. This is wrong. Once in Shelley (for now the final era), they have unlimited forecast range. However, if the node is in Byron, we need to be careful.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -254,9 +254,7 @@ import Cardano.Wallet.Primitive.CoinSelection
 import Cardano.Wallet.Primitive.Model
     ( Wallet, availableBalance, currentTip, getState, totalBalance )
 import Cardano.Wallet.Primitive.Slotting
-    ( PastHorizonException (..)
-    , Qry
-    , TimeInterpreter
+    ( TimeInterpreter
     , currentEpoch
     , endTimeOfEpoch
     , epochSucc
@@ -313,6 +311,8 @@ import Control.Monad
     ( forM, forever, void, (>=>) )
 import Control.Monad.IO.Class
     ( MonadIO, liftIO )
+import Control.Monad.Trans.Class
+    ( lift )
 import Control.Monad.Trans.Except
     ( ExceptT (..), runExceptT, throwE, withExceptT )
 import Control.Monad.Trans.Maybe
@@ -660,7 +660,7 @@ mkShelleyWallet ctx wid cp meta pending progress = do
         }
   where
     ti :: TimeInterpreter IO
-    ti = unsafeRunExceptT . timeInterpreter (ctx ^. networkLayer @t)
+    ti = timeInterpreter (ctx ^. networkLayer @t)
 
     -- This may fail
     -- 1. In Byron when running Byron;Shelley
@@ -783,7 +783,7 @@ mkLegacyWallet ctx wid cp meta pending progress = do
         }
   where
     ti :: TimeInterpreter IO
-    ti = unsafeRunExceptT . timeInterpreter (ctx ^. networkLayer @t)
+    ti = timeInterpreter (ctx ^. networkLayer @t)
 
     matchEmptyPassphrase
         :: WorkerCtx ctx
@@ -1229,7 +1229,7 @@ postTransaction ctx genChange (ApiT wid) withdrawRewards body = do
         #pendingSince
   where
     ti :: TimeInterpreter IO
-    ti = unsafeRunExceptT . timeInterpreter (ctx ^. networkLayer @t)
+    ti = timeInterpreter (ctx ^. networkLayer @t)
 
 deleteTransaction
     :: forall ctx s t k. ctx ~ ApiLayer s t k
@@ -1266,7 +1266,7 @@ listTransactions ctx (ApiT wid) mMinWithdrawal mStart mEnd mOrder = do
     -- In the Shelley of Byron;Shelley this is fine, but otherwise,
     -- supplying a time range in the future may cause a err500.
     ti :: TimeInterpreter IO
-    ti = unsafeRunExceptT . timeInterpreter (ctx ^. networkLayer @t)
+    ti = timeInterpreter (ctx ^. networkLayer @t)
 
 getTransaction
     :: forall ctx s t k n. (ctx ~ ApiLayer s t k)
@@ -1280,7 +1280,7 @@ getTransaction ctx (ApiT wid) (ApiTxId (ApiT (tid))) = do
     liftIO $ mkApiTransactionFromInfo ti tx
   where
     ti :: TimeInterpreter IO
-    ti = unsafeRunExceptT . timeInterpreter (ctx ^. networkLayer @t)
+    ti = timeInterpreter (ctx ^. networkLayer @t)
 
 -- Populate an API transaction record with 'TransactionInfo' from the wallet
 -- layer.
@@ -1370,7 +1370,7 @@ joinStakePool ctx knownPools getPoolStatus apiPoolId (ApiT wid) body = do
   where
     -- Not forecasting into the future. Should be safe.
     ti :: TimeInterpreter IO
-    ti = unsafeRunExceptT . timeInterpreter (ctx ^. networkLayer @t)
+    ti = timeInterpreter (ctx ^. networkLayer @t)
 
 delegationFee
     :: forall ctx s t n k.
@@ -1416,7 +1416,7 @@ quitStakePool ctx (ApiT wid) body = do
   where
     -- Not forecasting into the future. Should be safe.
     ti :: TimeInterpreter IO
-    ti = unsafeRunExceptT . timeInterpreter (ctx ^. networkLayer @t)
+    ti = timeInterpreter (ctx ^. networkLayer @t)
 
 {-------------------------------------------------------------------------------
                                 Migrations
@@ -1488,7 +1488,7 @@ migrateWallet ctx (ApiT wid) migrateData = do
     addrs = getApiT . fst <$> migrateData ^. #addresses
     -- Not forecasting into the future. Should be safe.
     ti :: TimeInterpreter IO
-    ti = unsafeRunExceptT . timeInterpreter (ctx ^. networkLayer @t)
+    ti = timeInterpreter (ctx ^. networkLayer @t)
 
 
 -- | Transform the given set of migration coin selections (for a source wallet)
@@ -1531,13 +1531,12 @@ getCurrentEpoch
     => ctx
     -> Handler W.EpochNo
 getCurrentEpoch ctx = do
-    res <- liftIO $ runExceptT (currentEpoch ti)
+    res <- liftIO $ currentEpoch ti
     case res of
-        Left _ -> liftE ErrUnableToDetermineCurrentEpoch
-        Right Nothing -> liftE ErrUnableToDetermineCurrentEpoch
-        Right (Just x) -> pure x
+        Nothing -> liftE ErrUnableToDetermineCurrentEpoch
+        Just x  -> pure x
   where
-    ti :: TimeInterpreter (ExceptT PastHorizonException IO)
+    ti :: TimeInterpreter IO
     ti = timeInterpreter (ctx ^. networkLayer @t)
 
 getNetworkInformation
@@ -1548,16 +1547,9 @@ getNetworkInformation
 getNetworkInformation (_block0, _, st) nl = do
     now <- liftIO getCurrentTime
     nodeTip <-  liftHandler (NW.currentNodeTip nl)
-    apiNodeTip <- liftIO $ unsafeRunExceptT $ mkApiBlockReference ti nodeTip
-
+    apiNodeTip <- liftIO $ mkApiBlockReference ti nodeTip
     nowInfo <- liftIO $ runMaybeT $ networkTipInfo now
-
-    -- I don't /think/ this should fail, but in case we do, it seems better to
-    -- return NotResponding than to throw an uncaught exception.
-    progress <- liftIO $ runExceptT (syncProgress st ti nodeTip now) >>= \case
-        Right x -> pure x
-        Left _ -> pure NotResponding
-
+    progress <- liftIO $ syncProgress st ti nodeTip now
     pure $ Api.ApiNetworkInformation
         { Api.syncProgress = ApiT progress
         , Api.nextEpoch = snd <$> nowInfo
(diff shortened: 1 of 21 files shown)


lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -456,10 +456,10 @@ withNetworkLayer tr np addrInfo versionData action = do
             bracketTracer (contramap (MsgWatcherUpdate header) tr) $
                 cb header
 
-    _timeInterpreterQuery :: HasCallStack => TVar IO (Maybe (CardanoInterpreter sc)) -> TimeInterpreter IO
+    _timeInterpreterQuery :: HasCallStack => MVar (CardanoInterpreter sc) -> TimeInterpreter IO
     _timeInterpreterQuery var query = do
-        cached <- atomically (readTVar var)
-        let interpret = maybe (interpreterFromGenesis gp) (mkTimeInterpreter getGenesisBlockDate) cached
+        cached <- takeMVar var
+        let interpret = (mkTimeInterpreter getGenesisBlockDate) cached
```
KtorZ: I am quite failing to grasp the consequences of this.

Anviking: Goal:
```
On launch, the TimeInterpreter may not have been fetched from the node.

Instead returning the singleEraInterpreter for the first era, it seems
safer to block until fetched.

I imagine there could otherwise be race conditions where we would sometimes
return completely wrong time data when in Shelley, just after starting the
node.
```
But:
- I should use `readMVar` instead of `takeMVar` 😅

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -450,15 +461,18 @@ withNetworkLayer tr np addrInfo versionData action = do
...
-        case interpret query of
-            Right res -> pure res
+        interpreter <- liftIO $ atomically $ readTMVar var
+        res <- liftIO $ runExceptT $ mkTimeInterpreter getGenesisBlockDate interpreter query
+        case res of
+            Right r -> pure r
             Left e -> do
-                traceWith tr $ MsgInterpreterPastHorizon (pretty query) e
-                throwIO e
+                liftIO $ traceWith tr $ MsgInterpreterPastHorizon (pretty query) e
```
Anviking, line 489: `MsgInterpreterPastHorizon` shouldn't be critical anymore, as it might be expected.

Anviking:
bors try

Anviking:
Connecting to mainnet, when the node is still syncing, I often see 
```
cardano-wallet-shelley network information
ConnectionError (HttpExceptionRequest Request {
  host                 = "localhost"
  port                 = 8090
  secure               = False
  requestHeaders       = [("Accept","application/json;charset=utf-8,application/json")]
  path                 = "/v2/network/information"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
 NoResponseDataReceived)
```
Other times I saw the response without the `next_epoch` or `network_tip`, which would be expected, **but** with an unexpected `ready` `sync_progress` status.
```
{
    "node_tip": {
        "height": {
            "quantity": 435869,
            "unit": "block"
        },
        "epoch_number": 20,
        "slot_number": 3925
    },
    "sync_progress": {
        "status": "ready"
    }
}
```

Anviking:
With synced node in Byron, looking good:
```
cardano-wallet-shelley network information
Ok.
{
    "network_tip": {
        "epoch_number": 207,
        "slot_number": 12408
    },
    "node_tip": {
        "height": {
            "quantity": 4481319,
            "unit": "block"
        },
        "epoch_number": 207,
        "slot_number": 12408
    },
    "sync_progress": {
        "status": "ready"
    },
    "next_epoch": {
        "epoch_number": 208
    }
}
```

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
Anviking: I lowered to `Info` and removed the log line about it being unexpected.

Anviking:
It seems to work as I'd expect against Byron mainnet now. After 4b76ac2 I don't see any weird errors, or wrong sync progress.

Preferably I should add a regression test. 

But should be reviewable now.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1349,8 +1366,9 @@ joinStakePool ctx knownPools getPoolStatus apiPoolId (ApiT wid) body = do
         (txMeta, txTime)
         #pendingSince
   where
+    -- Not forecasting into the future. Should be safe.
     ti :: TimeInterpreter IO
-    ti = timeInterpreter (ctx ^. networkLayer @t)
+    ti = unsafeRunExceptT . timeInterpreter (ctx ^. networkLayer @t)
```
Anviking: It doesn't seem optimal to have to argue "should be safe". But I'm not sure that is easy to solve for the moment.

Anviking:
Added `syncProgress` regression test -> 95a9564

lib/core/src/Cardano/Wallet/Api/Server.hs
KtorZ: `unsafe` will throw and cause a 500, we shouldn't use it if that can indeed happen. What about instead having a proper error, even if it's just a server error, but at least not an opaque 500.

Anviking:
bors try

KtorZ:
I am a little bit _reluctant_ with all these `unsafeExceptT` ... If the error indeed is a true exception that callers shouldn't look for handling, then why is it returned as a value? 
If that `unsafeExceptT` is there as a quick'n'dirty way of getting everything compiled without creating proper error types, then this is wrong.

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
```diff
@@ -26,13 +26,14 @@ module Cardano.Wallet.Primitive.Slotting
...
     , ongoingSlotAt
+    , endTimeOfEpoch
 
     -- ** Running queries
     , TimeInterpreter
     , singleEraInterpreter
-    , interpreterFromGenesis
     , mkTimeInterpreter
-    , MyInterpreter(..)
+    , unsafeNoForecasts
```
Anviking: I should either use this consistently over `unsafeRunExceptT`, or remove it.

Anviking:
> If that unsafeExceptT is there as a quick'n'dirty way of getting everything compiled without creating proper error types, then this is wrong.

No. The errors `unsafeExceptT` here could be thrown in #1909 too (with `throwIO`, but hidden in the NetworkLayer)
Where it matters the most, this PR adds proper error handling.

> If the error indeed is a true exception that callers shouldn't look for handling, then why is it returned as a value?

There are cases where error handling is needed, and here are cases where it is not. There are also places that work fine in Byron;Shelley, but will need to be re-thought in Byron;Shelley;Gougen (the time of delegation changes, and pool retirements).

Anviking:
bors try

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: > instead having a proper error

I realised that we should at least use a custom error message instead of `unsafeRunExceptT`, and have the motivation why it is impossible _there_ instead of in a comment.

rvl (commented):
There is so much `unsafeRunExceptT` here. It makes our code look unsafe.
This is replacing an actual exception type `PastHorizonException` with `userError` and a string message, which is worse.

Why not let all the time/slot conversion functions throw PastHorizonException, wherever unsafeRunExceptT is currently used. In the other "safe" cases, catch the exception.

Alternatively, replace `unsafeRunExceptT` with a new util `runNonFailingQuery` (or some name like that). This will be basically `runNonFailingQuery q = q >>= either throwIO pure`.  The advantage of this is that it's obvious where the exception could occur, and userError is not used.


And also we should sprinkle HasCallStack liberally - otherwise tracking down errors will be hard.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -1024,4 +1037,4 @@ instance HasSeverityAnnotation (NetworkLayerLog b) where
         MsgWatcherUpdate{}                 -> Debug
         MsgChainSyncCmd cmd                -> getSeverityAnnotation cmd
         MsgInterpreter{}                   -> Debug
-        MsgInterpreterPastHorizon{}        -> Critical
+        MsgInterpreterPastHorizon{}        -> Info
```
KtorZ: Is this change really sound :thinking: ?

rvl: I think the reasoning is that we have actually some cases where the PastHorizonException is "normal" and somewhat expected.

But perhaps we can divide slot/time conversions into two categories:
1. Where it's our bug if the conversion fails.
2. Where any conversion errors are because of the user inputting a date which is beyond the safe zone. So an error not a bug.

KtorZ:
bors try

Anviking:
bors try

KtorZ (approved):

KtorZ:
Waited for the "trying" branch to pass. Then pushed a little update that reshaping the git history and, lowering the "PastHorizon" error from `Critical` to `Error`. Here was the diff:

```
$ git diff HEAD origin/anviking/1869/improve-time-interpreter 
diff --git a/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs b/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
index 833ff04f2..971cfbfc4 100644
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
@@ -1051,4 +1051,4 @@ instance HasSeverityAnnotation (NetworkLayerLog b) where
         MsgWatcherUpdate{}                 -> Debug
         MsgChainSyncCmd cmd                -> getSeverityAnnotation cmd
         MsgInterpreter{}                   -> Debug
-        MsgInterpreterPastHorizon{}        -> Error
+        MsgInterpreterPastHorizon{}        -> Critical
```

Issue #1937, opened 2020-07-22
Do not list retired pools in the API.


jonathanknowles:
# Context

As part of the work to track stake pool retirements, we need to prevent pools that have already retired from being listed in public API operations.

# Decision

- Prevent the `ListStakePools` API operation from listing pools that have already retired.
- If practical, use a database query to filter out retired pools, rather than in-memory filter.

# Acceptance Criteria

- [x] The wallet must not include pools that have already retired in the list of known pools returned by the API `ListStakePools` operation.

# Development 
- https://github.com/input-output-hk/cardano-wallet/pull/1945

# QA 

- [x] Provide an integration test that:

    1. Registers a pool _**p**_
    1. Publishes a retirement certificate for pool _**p**_ with retirement epoch _**e**_.
    1. Verifies that when the current epoch is equal to or later than _**e**_, pool _**p**_ does not appear in the list of stake pools returned by `ListStakePools`.

# Notes

Note that the above integration test is a compromise.

Ideally, a more complete integration test would verify that a pool is still visible right up to the moment before the pool has actually retired. See: https://jira.iohk.io/browse/ADP-385

piotr-iohk:
lgtm

PR #1945, opened 2020-07-24
Exclude retired pools from the `ListStakePools` API function result.

jonathanknowles:
# Issue Number

#1937 

# Overview

This PR:

- [x] Applies a very simple filter to the Shelley `ListStakePools` API operation that excludes all pools with a retirement epoch earlier than or equal to the current epoch.
- [x]  Adds an integration test to verify that if pool _**p**_ has already retired, it will **not** be listed by `ListStakePools`.

# Future Improvements

The current pool tracking implementation (which predates this PR) has two potential areas for improvement:

- Reducing the number of database queries that need to be executed:
    See: https://jira.iohk.io/browse/ADP-383
- Performing garbage collection of retired pools from the database:
    See: https://jira.iohk.io/browse/ADP-376

These are **not** tackled in this PR, but are instead recorded in the above tickets, so that we may schedule time to fix them at a later date.

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
@@ -660,7 +660,7 @@ spec = do
         it "contains pool metadata" $ \ctx -> do
             eventually "metadata is fetched" $ do
                 r <- listPools ctx arbitraryStake
-                let metadataList =
+                let metadataPossible = Set.fromList
                         [ StakePoolMetadata
                             { ticker = (StakePoolTicker "GPA")
                             , name = "Genesis Pool A"
@@ -679,16 +679,21 @@ spec = do
                             , description = Just "Lorem Ipsum Dolor Sit Amet."
                             , homepage = "https://iohk.io"
                             }
+                        , StakePoolMetadata
+                            { ticker = (StakePoolTicker "GPD")
+                            , name = "Genesis Pool D"
+                            , description = Just "Lorem Ipsum Dolor Sit Amet."
+                            , homepage = "https://iohk.io"
+                            }
                         ]
 
                 verify r
                     [ expectListSize 3
-                    , expectField Prelude.id $ \pools ->
-                        -- To ignore the arbitrary order,
-                        -- we sort on the names before comparing
-                        sortOn name
-                            (mapMaybe (fmap getApiT . view #metadata) pools)
-                            `shouldBe` metadataList
+                    , expectField Prelude.id $ \pools -> do
+                        let metadataActual = Set.fromList $
+                                mapMaybe (fmap getApiT . view #metadata) pools
+                        metadataActual
+                            `shouldSatisfy` (`Set.isSubsetOf` metadataPossible)
                     ]
 
         it "contains and is sorted by non-myopic-rewards" $ \ctx -> do
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
@@ -1122,6 +1122,33 @@ operators = unsafePerformIO $ newMVar
           , "homepage" .= Aeson.String "https://iohk.io"
           ]
       )
+    , ( PoolId $ unsafeFromHex
+          "5a7b67c7dcfa8c4c25796bea05bcdfca01590c8c7612cc537c97012bed0dec36"
+      , Aeson.object
+          [ "type" .= Aeson.String "Node operator verification key"
+          , "description" .= Aeson.String "Stake Pool Operator Verification Key"
+          , "cborHex" .= Aeson.String
+              "58203263e07605b9fc0100eb520d317f472ae12989fbf27fc71f46310bc0f24f2970"
+          ]
+      , Aeson.object
+          [ "type" .= Aeson.String "Node operator signing key"
+          , "description" .= Aeson.String "Stake Pool Operator Signing Key"
+          , "cborHex" .= Aeson.String
+              "58208f50de27d74325eaf57767d70277210b31eb97cdc3033f632a9791a3677a64d2"
+          ]
+      , Aeson.object
+          [ "type" .= Aeson.String "Node operational certificate issue counter"
+          , "description" .= Aeson.String "Next certificate issue number: 0"
+          , "cborHex" .= Aeson.String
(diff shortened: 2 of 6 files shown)


jonathanknowles:
bors try

jonathanknowles:
bors try

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -679,16 +679,21 @@ spec = do
...
-                        -- To ignore the arbitrary order,
-                        -- we sort on the names before comparing
-                        sortOn name
-                            (mapMaybe (fmap getApiT . view #metadata) pools)
-                            `shouldBe` metadataList
+                    , expectField Prelude.id $ \pools -> do
+                        let metadataActual = Set.fromList $
+                                mapMaybe (fmap getApiT . view #metadata) pools
+                        metadataActual
+                            `shouldSatisfy` (`Set.isSubsetOf` metadataPossible)
```
piotr-iohk, line 696: Nitpicking (:sweat_smile:), but maybe we could add a stronger assertion that the list of returned pools does not contain GPD?

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -1122,6 +1122,33 @@ operators = unsafePerformIO $ newMVar
           , "homepage" .= Aeson.String "https://iohk.io"
           ]
       )
+    , ( PoolId $ unsafeFromHex
+          "5a7b67c7dcfa8c4c25796bea05bcdfca01590c8c7612cc537c97012bed0dec36"
```
piotr-iohk, line 1126: :tada:

piotr-iohk (commented):
There is one problem I think. Playing around with the CI cluster I've observed that the pool scheduled to be retired in epoch `n` is still visible through the whole epoch `n` and disappears from the list in epoch `n+1`. While it is visible one can still join it but gets:
```
{"code":"no_such_pool"
,"message":"I couldn't find any stake pool with the given id: b45768c1a2da4bd13ebcaa1ea51408eda31dcc21765ccbd407cda9f2"}
```

I guess pool retired to be scheduled in `n` should not be shown from `n` onwards...

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
jonathanknowles: > Nitpicking (sweat_smile), but maybe we could add a stronger assertion that the list of returned pools does not contain GPD?

Unfortunately we can't currently do this, as there is no guarantee that pool `GPD` will be the one to be assigned with an early retirement epoch. It could be _any_ of the four pools that gets retired early.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -166,17 +175,35 @@ newStakePoolLayer gp nl db@DBLayer {..} = StakePoolLayer
...
+        -- See: https://jira.iohk.io/projects/ADP/issues/ADP-383
         return
             . sortOn (Down . (view (#metrics . #nonMyopicMemberRewards)))
+            . filter (not . poolIsRetired)
             . map snd
             . Map.toList
             $ combineDbAndLsqData (slotParams gp) lsqData dbData
+      where
+        epochIsInFuture :: EpochNo -> Bool
+        epochIsInFuture = (>= currentEpoch)
```
jonathanknowles: Surely, this should be:

```suggestion
        epochIsInFuture = (> currentEpoch)
```

jonathanknowles:
> There is one problem I think. Playing around with the CI cluster I've observed that the pool scheduled to be retired in epoch `n` is still visible through the whole epoch `n` and disappears from the list in epoch `n+1`. While it is visible one can still join it but gets:
> 
> ```
> {"code":"no_such_pool"
> ,"message":"I couldn't find any stake pool with the given id: b45768c1a2da4bd13ebcaa1ea51408eda31dcc21765ccbd407cda9f2"}
> ```
> 
> I guess pool retired to be scheduled in `n` should not be shown from `n` onwards...

Nice catch!

I think I've found the source of this here:

https://github.com/input-output-hk/cardano-wallet/pull/1945/files#r460378458

I'll post a patch shortly.

EDIT: fixed in 6ec42fbe22762b91d2607f1b2bdab8e2312d6eb7

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
jonathanknowles: I suspect we could solve this problem entirely by changing our test framework to allow pools to be registered and retired dynamically, as in: https://jira.iohk.io/browse/ADP-385

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
jonathanknowles: Fixed in 6ec42fbe22762b91d2607f1b2bdab8e2312d6eb7.

piotr-iohk (approved):
LGTM!

Corner case that may be considered in future PR/story:
 - pool A schedule to retire in `n`
 - wallet joins A in `n-1` (shows as delegation next: `n + 1`, target: A)
 - pool disappears from the list in `n` (wallet still shows as delegation next: `n + 1`, target: A)
 - since `n + 1` wallet shows as delegation active, target: A

jonathanknowles:
> Corner case that may be considered in future PR/story:
>     * pool A schedule to retire in `n`
>     * wallet joins A in `n-1` (shows as delegation next: `n + 1`, target: A)
>     * pool disappears from the list in `n` (wallet still shows as delegation next: `n + 1`, target: A)
>     * since `n + 1` wallet shows as delegation active, target: A

I agree we should definitely discuss what we want the `GetWallet` and `ListWallets` operations to show in the case where a pool has already retired, including the above corner case.

I've opened the following ticket to capture this discussion: https://github.com/input-output-hk/cardano-wallet/issues/1949.

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
KtorZ: Actually no :/ ... The pool id is a hash of the operator verification key. I am not sure how this successfully got merged but it's now causing issue.

PR #1946, opened 2020-07-24
Withdrawals accounted in received side of prefiler

paweljakubas:
# Issue Number

#1923 

# Overview

- [x] I have changed how received in prefilterblock are calculated when withdrawals are present
- [x] I have captured that in the test

# Comments


lib/core/src/Cardano/.#Wallet.hs
```diff
@@ -0,0 +1 @@
+pawel@pure.5413:1595575265
```
KtorZ: Whoops :no_mouth:

paweljakubas:
so in tx with withdrawal we have at this moment : 
```
Pending ->  amount: 142900, inps: [100000000000], outs: [1, 100002984537], withdrawals: [3127438]
InLedger -> amount: 6111976, inps: [100000000000], outs: [1, 100002984537], withdrawals: [3127438]
```
in Pending we have the following obeyed : 
```
amount = sum(withdrawals)+sum(inputs) - sum(outs)
```

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -376,7 +377,9 @@ prefilterBlock b u0 = runState $ do
         ourU <- state $ utxoOurs tx
         let ourIns = Set.fromList (inputs tx) `Set.intersection` dom (u <> ourU)
         let u' = (u <> ourU) `excluding` ourIns
-        let received = fromIntegral @_ @Integer $ balance ourU
+        let withdrawalAmt =
+                sum $ map (fromIntegral . getCoin) $ Map.elems $ withdrawals tx
+        let received = fromIntegral @_ @Integer $ (withdrawalAmt + balance ourU)
         let spent = fromIntegral @_ @Integer $ balance (u `restrictedBy` ourIns)
```
KtorZ, line 383: So actually, the withdrawals count towards the `spent` here (it's like an input that has been spent!)

paweljakubas:
bors try

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -253,6 +254,7 @@ spec = do
             verify rWithdrawal
                 [ expectResponseCode HTTP.status200
                 , expectField #withdrawals (`shouldSatisfy` (not . null))
+                , expectField #amount (`shouldBe` totalAmtAfterWithdrawals)
```
paweljakubas, line 257: so here basically we check that amount in tx with withdrawal is the same as tx goes from Pending -> InLedger

KtorZ (approved):

KtorZ:
bors retry

rvl:
bors retry

piotr-iohk:
I have been looking on this branch on testnet and experienced "weird" observation.
Ok, first of all, amt sent is correct now (both when tx is `pending` or `in_ledger`)... however I see that the balances (total, available, rewards) are "stalled" after transaction is done (and already in ledger)... the wallet still shows previous balance for upto 15-20s before it gets updated.

E.g:
0. Wallet with 
```
Balance: (Total: 1824872492266 , Available: 962475626890 , Rewards: 862396865376)
```
1. Sending tx with amt = 862396865376 (exactly the level of tx)
2. Tx is sent, amt reported on `pending` tx = 862396865376 + fee :heavy_check_mark: 
3. Tx is in ledger, amt reported on `in_ledger` tx = 862396865376 + fee :heavy_check_mark: 
4. Looking at balance, it is still:
```
Balance: (Total: 1824872492266 , Available: 962475626890 , Rewards: 862396865376)
```
5. Balance gets updated after 15 seconds.

KtorZ:
@piotr-iohk requesting the reward balance on each call is quite expensive and has been shown unreliable; so instead, the reward balance is polled every now and then (every _block_ I think) and cached. So here, it's probably just the time for the cache to get updated.

piotr-iohk:
@KtorZ @paweljakubas - OK, just sharing what I see ... it might have to be then reported to Daedalus guys, that this is somewhat expected on transactions with withdrawals.

Issue #1952, opened 2020-07-26
Incoming tx with withdrawals shows wrong amount received


piotr-iohk:
# Context
 
https://github.com/input-output-hk/cardano-wallet/issues/1923 addresses amount being displayed on **outgoing** transaction, however we missed the **incoming** tx part...

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | `2020.7.6 (git revision: 71c0b8fd10aaf7e09c62c42571780f7676381584)` |
| Platform     | Linux, all        |
| Installation | Daedalus, from source           |

# Steps to Reproduce

1. Send tx with `withdrawRewards=true` amount from the wallet with reward = X
2. Send tx amt = X + n

## Expected behavior

Outgoing tx amount = X + n + fee
Incoming tx amount = X + n

## Actual behavior
 
Outgoing tx amount = X + n + fee
Incoming tx amount = n

For instance transaction where amount sent was exactly the level of rewards (_outgoing_, on the left, _incoming_ on the right). Incoming transaction (on the right) shows amount received = 0:

![Screenshot from 2020-07-26 08-39-43](https://user-images.githubusercontent.com/42900201/88473317-98559180-cf1c-11ea-9d62-fcbb56d929af.png)

Transaction where amount sent  was _rewards + 1000 000 lovelace_ (_outgoing_ on the left, _incoming_ on the right). 
Incoming transaction (on the right) shows amount received = 1000 000 lovelace

![Screenshot from 2020-07-26 08-43-41](https://user-images.githubusercontent.com/42900201/88473357-fa15fb80-cf1c-11ea-859f-82184f5ee0f5.png)

---

# Resolution 

- #1954 

---

# QA 

- Changed our current scenario that is spending rewards to send rewards to _another_ wallet (was sending them to itself before) so that we could add assertion from both perspective (sender & receiver). In particular, added:

    https://github.com/input-output-hk/cardano-wallet/blob/905277f370093ab43c9246ee6d7c32df7114fe1d/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs#L264-L270

    which failed initially without fix. Note that this assertion expects _two_ transactions because the scenarios makes 2 (one without withdrawal, and one with withdrawal). But, from the perspective of the received, the reported amount of the transaction is exactly the amount sent (since the withdrawal is sent back to the source wallet as change).

piotr-iohk:
lgtm

PR #1954, opened 2020-07-26
Fix reported amount on incoming transactions carrying withdrawals

KtorZ:
# Issue Number

#1952 

# Overview

- d1c5618b79ac861600e94b7200a94c89a7b80f87
  :round_pushpin: **modify scenario spending rewards to send money to _another_ wallet and assess right amount**
    We didn't quite catch a bug here because we were only looking at one side of the coin by sending the transaction to ourselves (so, we couldn't really assess the amount on an _incoming_ transaction, from a different perspective. That is now done properly such that 1 lovelace is send to another wallet (and, which should be the amount reported by this wallet). The outgoing amount however should contain any output sent to _another_ wallet, as well as fees.

- 6932d1d7d73677d840a5cebc7ac96d31eca96f42
  :round_pushpin: **revise 'amount' calculation in the primitive model**
    Such that we only count withdrawals in outgoing transactions. In principle, withdrawals should be necessarily ours on a transaction.. but it's not excluded that a transaction may contain multiple withdrawals from different wallets (especially in the context of multisig). Beside, withdrawals should be completely out of the picture for _incoming_ transactions.

- aa9fdf9ba996898a6978a4eb4af638361e466bcc
  :round_pushpin: **add test showing oddities on total balance + rewards when there are pending withdrawals + fix it**
    This was trickier than I originally thought because crafting a generator of pending transactions which is somewhat meaningful and can have withdrawals turned out to be a bit complex. The property originally failed (without the fix) and clearly showed the bug described in #1955. The fix is relatively straightforward and is about adding the reward amount only if there are no pending withdrawal

- 5867d38f818e14c802de8293aed1ac425227e478
  :round_pushpin: **fix SlotNo generator in database arbitrary instances**
    This one was wrongly changed to browse the entire Word32 space instead of staying within small bounds as originally. There are many things in tests and other generators derived from the slot number value, and too big values can really do bad things here.

- 8d1eadb42cc263116f27b625a8cb5438653b63f9
  :round_pushpin: **prevent users from creating two withdrawals simultaneously on the same wallet**
    Rewards ought to be spent fully, therefore, making two concurrent withdrawals has no sense since only one transaction will eventually be validated.

# Comments

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
@@ -150,12 +150,13 @@ spec = do
     it "STAKE_POOLS_JOIN_01 - \
         \Can join a pool, earn rewards and collect them" $ \ctx -> do
         -- Setup
-        w <- fixtureWallet ctx
+        src <- fixtureWallet ctx
+        dest <- emptyWallet ctx
 
         -- Join Pool
         pool:_ <- map (view #id) . snd <$> unsafeRequest @[ApiStakePool] ctx
             (Link.listStakePools arbitraryStake) Empty
-        joinStakePool @n ctx pool (w, fixturePassphrase) >>= flip verify
+        joinStakePool @n ctx pool (src, fixturePassphrase) >>= flip verify
             [ expectResponseCode HTTP.status202
             , expectField (#status . #getApiT) (`shouldBe` Pending)
             , expectField (#direction . #getApiT) (`shouldBe` Outgoing)
@@ -166,7 +167,7 @@ spec = do
         waitForNextEpoch ctx
         (previousBalance, walletRewards) <-
             eventually "Wallet gets rewards" $ do
-                r <- request @ApiWallet ctx (Link.getWallet @'Shelley w)
+                r <- request @ApiWallet ctx (Link.getWallet @'Shelley src)
                     Default Empty
                 verify r
                     [ expectField
@@ -180,7 +181,7 @@ spec = do
                 pure (availableBalance, rewardBalance)
 
         -- Try to use rewards
-        addrs <- listAddresses @n ctx w
+        addrs <- listAddresses @n ctx dest
         let coin = 1 :: Natural
         let addr = (addrs !! 1) ^. #id
         let payload = [json|
@@ -197,18 +198,18 @@ spec = do
 
         -- cannot use rewards by default
         r1 <- request @(ApiTransaction n) ctx
-            (Link.createTransaction @'Shelley w)
+            (Link.createTransaction @'Shelley src)
             Default (Json payload)
         expectResponseCode HTTP.status202 r1
         eventually "Wallet has not consumed rewards" $ do
           let linkSrc = Link.getTransaction @'Shelley
-                  w (getFromResponse Prelude.id r1)
+                  src (getFromResponse Prelude.id r1)
           request @(ApiTransaction n) ctx linkSrc Default Empty
               >>= flip verify
                   [ expectField
                       (#status . #getApiT) (`shouldBe` InLedger)
                   ]
-          request @ApiWallet ctx (Link.getWallet @'Shelley w) Default Empty
+          request @ApiWallet ctx (Link.getWallet @'Shelley src) Default Empty
               >>= flip verify
                   [ expectField
                       (#balance . #getApiT . #reward) (`shouldBe` walletRewards)
@@ -216,7 +217,7 @@ spec = do
 
         -- there's currently no withdrawals in the wallet
         rw1 <- request @[ApiTransaction n] ctx
-            (Link.listTransactions' @'Shelley w (Just 1)
+            (Link.listTransactions' @'Shelley src (Just 1)
                 Nothing Nothing Nothing)
             Default Empty
         verify rw1 [ expectListSize 0 ]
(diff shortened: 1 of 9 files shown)


KtorZ:
bors try

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -244,30 +245,38 @@ spec = do
...
+                , expectField #amount (`shouldBe` txAmount)
                 ]
             rw2 <- request @[ApiTransaction n] ctx
-                (Link.listTransactions' @'Shelley w (Just 1)
+                (Link.listTransactions' @'Shelley src (Just 1)
                     Nothing Nothing Nothing)
                 Default Empty
             verify rw2 [ expectListSize 1 ]
 
+        eventually "There's one incoming transaction with correct amount" $ do
```
paweljakubas, line 264: :+1:

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -369,27 +376,25 @@ prefilterBlock b u0 = runState $ do
...
     applyTx (!txs, !u) tx = do
         ourU <- state $ utxoOurs tx
         let ourIns = Set.fromList (inputs tx) `Set.intersection` dom (u <> ourU)
         let u' = (u <> ourU) `excluding` ourIns
-        let withdrawalAmt =
-                sum $ map (fromIntegral . getCoin) $ Map.elems $ withdrawals tx
-        let received = fromIntegral @_ @Integer $ balance ourU
-        let spent = fromIntegral @_ @Integer $ balance (u `restrictedBy` ourIns)
-        let amt = fromIntegral $ abs (received - spent - withdrawalAmt)
+        ourWithdrawals <- mapMaybeM ourChimericAccount $ Map.toList $ withdrawals tx
```
paweljakubas, line 396: yes, we need to check if withdrawals are ours

paweljakubas (approved):
looks like a fix! extension to testing persuasive

Anviking:
Not specific to PR, but shouldn't [`availableUTxO`](https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/Primitive/Model.hs#L308) be changed to exclude outgoing withdrawals too? 

edit -> #1955

Anviking (approved):
Fix looks good to me. Wanted to check that the adjusted tests cover the fix, but I only get socket errors  when I try to run the tests locally…

piotr-iohk (approved):
Looks good also testing manually.

KtorZ:
I'll include a fix for #1955 in the meantime.

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -214,6 +229,40 @@ prop_initialBlockHeight s =
...
+    $ counterexample ("Pending Transactions:\n" <> pretty' (blockListF pending))
+    $ counterexample ("Pending balance:            " <> show pendingBalance)
+    $ counterexample ("Total Rewards:              " <> show (getCoin rewards))
+    $ counterexample ("Total Balance (w/ pending): " <> show totalWithPending)
+    $ counterexample ("Total Balance (no pending): " <> show totalWithoutPending)
+    $ classify hasPending "has pending"
+    $ classify hasWithdrawals "has withdrawals"
+    $ if hasPending
+      then property (totalWithPending < totalWithoutPending)
+      else property (totalWithPending == totalWithoutPending)
```
KtorZ, line 255: The property is actually more precise than it says "counting rewards once". It really shows that having pending transactions result in a smaller balance for the wallet, regardless of whether there are withdrawals or not. 

Of course, this only holds in the presence of fees that are non-null.

KtorZ:
```
only counts rewards once.
      +++ OK, passed 10000 tests:
      45.33% has pending
       5.21% has withdrawals
```

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -641,8 +642,8 @@ mkShelleyWallet ctx wid cp meta pending progress = do
         { addressPoolGap = ApiT $ getState cp ^. #externalPool . #gap
         , balance = ApiT $ WalletBalance
             { available = Quantity $ availableBalance pending cp
-            , total = Quantity $ reward + totalBalance pending cp
-            , reward = Quantity reward
+            , total = Quantity $ totalBalance pending reward cp
```
Anviking, line 645: 👍

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -289,14 +289,23 @@ availableBalance :: Set Tx -> Wallet s -> Natural
...
-    :: IsOurs s Address
+    :: (IsOurs s Address, IsOurs s ChimericAccount)
     => Set Tx
+    -> Quantity "lovelace" Natural
     -> Wallet s
     -> Natural
-totalBalance pending =
-    balance . totalUTxO pending
+totalBalance pending (Quantity rewards) s =
+    balance (totalUTxO pending s) + if hasPendingWithdrawals then 0 else rewards
```
Anviking, line 300: 👍

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
Anviking: 👍

Anviking:
Lgtm, but I can't check what happens in reality as I don't have a wallet containing 99% rewards anymore.

I'm wondering if there could still be strange behaviour from race-conditions from the fact that the reward-balance and wallet utxo are not guaranteed to be in-sync.

KtorZ:
> I'm wondering if there could still be strange behaviour from race-conditions from the fact that the reward-balance and wallet utxo are not guaranteed to be in-sync.

What I am wondering at the moment is: what exactly happens when you submit two withdrawals. Surely, the wallet doesn't really check for any pending withdrawals and will therefore "allow it" (modulo available UTxO) to create 2 pending transactions with withdrawals. An immediate problem will be that because only one transaction will eventually be inserted, the other one will stay pending forever (or until canceled) and, the rewards balance of the wallet would not be counted towards the total balance so long as that transaction is pending :thinking: ...

PR #1961, opened 2020-07-27
Handle Network.Socket.connect: resource exhausted exceptions

rvl:
### Issue Number

Relates to #1708

### Overview

- Adds regression test
- Fixes unhandled exception resulting from parallel local snocket connections (see issue description for more details).

### DIFF
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
@@ -804,6 +804,12 @@ handleIOException tr onResourceVanished e
     | isDoesNotExistError e =
         pure True
 
+    -- If the nonblocking UNIX domain socket connection cannot be completed
+    -- immediately (i.e. connect() returns EAGAIN), try again. This happens
+    -- because the node's listen queue is quite short.
+    | isTryAgainError e =
+        pure True
+
     | isResourceVanishedError e = do
         traceWith tr $ Just e
         pure onResourceVanished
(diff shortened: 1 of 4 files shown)


lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -804,6 +804,12 @@ handleIOException tr onResourceVanished e
     | isDoesNotExistError e =
         pure True
 
+    -- If the nonblocking UNIX domain socket connection cannot be completed
+    -- immediately (i.e. connect() returns EAGAIN), try again. This happens
+    -- because the node's listen queue is quite short.
+    | isTryAgainError e =
+        pure True
```
KtorZ, line 811: Interesting :thinking:

lib/shelley/test/unit/Cardano/Wallet/Shelley/NetworkSpec.hs
```diff
@@ -0,0 +1,47 @@
...
+
+{-------------------------------------------------------------------------------
+                                      Spec
+-------------------------------------------------------------------------------}
+
+spec :: Spec
+spec = describe "NetworkLayer regression test #1708" $ do
+    it "Parallel local socket connections" $
+        withTestNode $ \np sock vData -> withLogging $ \(tr, _getLogs) ->
+            replicateConcurrently_ 5 $
```
KtorZ, line 36: Was this enough to observe the failure? (I assume so)

KtorZ (approved):
:+1: Nice!

KtorZ:
Failure above is another instance of the unit test OOM issue, fixed in another PR.

lib/shelley/test/unit/Cardano/Wallet/Shelley/NetworkSpec.hs
rvl: Yes - probably 3 concurrently would be enough, but why not 5.

Issue #1959, opened 2020-07-27
Sort pools with equal rewards randomly


Anviking:
# Context

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      |  |
| Platform     |           |
| Installation |             |

# Steps to Reproduce

1. List pools

## Expected behavior

1. Having a poolId starting with `0` shouldn't be better than one starting with `f`
2. Pools with equal rewards are sorted randomly
3. The random order is unique to each wallet/user
4. The random order is the consistent for the same wallet/user

## Actual behavior

1. It seems pools are ranked alphabetically on poolId, when they have equal reards
 

---

# Resolution 

- Make sure to sort pools that have identical rewards 

---

# QA 

- Pools are now sorted "twice", a first time based on weight that is randomly attributed to them. The weights are generate from a seed that is common to each wallet of a server (so each wallet gets the same randomization), but different across server instances. Then, that randomized list of pools is sorted again based on their non-myopic rewards. In particular, Haskell's `sortOn` function used for the second sorting has the nice property of _"keeping duplicates in the order they appeared in the input."_, which makes the whole thing work nicely. 

![Screenshot from 2020-07-28 14-49-46](https://user-images.githubusercontent.com/5680256/88667405-9e489f80-d0e1-11ea-9134-74d8282d4f08.png)

https://github.com/input-output-hk/cardano-wallet/blob/aab4ddf15fd727b4ed476fa17f6be6c69e94a6e7/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs#L227-L248

- Existing integration tests still pass, which means that pools are still ordered by descending non-myopic rewards. Testing however that pools with identical rewards are correctly randomized is tricky.

- Some additional testing was done on the mainnet candidate manually where it appeared "obvious" that new pools were no longer ordered by pool's id.

piotr-iohk:
lgtm. Also looking a bit on mainnet candidate and looks like pools on different wallets are sorted differently.

PR #1962, opened 2020-07-27
Assign average member rewards to new pools & sort equal pools arbitrarily 

KtorZ:
# Issue Number

#1959 & #1930

# Overview

- e71ab7d4330793fbe7c7e6a648fea952ea4e666a
  :round_pushpin: **use average member rewards of the top k pools for freshly registered pools**
    So that everyone gets a chance. Note that this is a potentially
  temporary solution until. Another approach is currently under
  discussion by the ledger team which may end up implementing something
  directly within the query itself.

- fda53fd2b5c3280dbef59ff2a14f9ebb82f62e32
  :round_pushpin: **randomly sort stake pools with identical non-myopic rewards.**
    The nice thing about this approach is that it works for both:

  - initial conditions, when all pools may have a the same non-myopic rewards of '0' due to the network just starting
  - long-term cases where there are pools that have different non-myopic member rewards but, some of them have equal rewards.

# Comments

Testing --> I expect everything in integration tests to still pass. For the rest, it'll be manual testing on MCX

### DIFF
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
@@ -37,7 +37,8 @@ import Cardano.Pool.DB
 import Cardano.Wallet.Api.Types
     ( ApiT (..) )
 import Cardano.Wallet.Network
-    ( ErrNetworkUnavailable
+    ( ErrCurrentNodeTip (..)
+    , ErrNetworkUnavailable (..)
     , FollowAction (..)
     , FollowExit (..)
     , FollowLog
@@ -58,7 +59,7 @@ import Cardano.Wallet.Primitive.Types
     , PoolLifeCycleStatus (..)
     , PoolRegistrationCertificate (..)
     , PoolRetirementCertificate (..)
-    , ProtocolParameters
+    , ProtocolParameters (..)
     , SlotLength (..)
     , SlotNo
     , StakePoolMetadata
@@ -77,7 +78,7 @@ import Cardano.Wallet.Shelley.Compatibility
 import Cardano.Wallet.Shelley.Network
     ( NodePoolLsqData (..) )
 import Cardano.Wallet.Unsafe
-    ( unsafeMkPercentage, unsafeRunExceptT )
+    ( unsafeMkPercentage )
 import Control.Concurrent
     ( threadDelay )
 import Control.Monad
@@ -85,7 +86,7 @@ import Control.Monad
 import Control.Monad.IO.Class
     ( liftIO )
 import Control.Monad.Trans.Except
-    ( ExceptT (..), runExceptT )
+    ( ExceptT (..), runExceptT, withExceptT )
 import Control.Tracer
     ( Tracer, contramap, traceWith )
 import Data.Function
@@ -122,6 +123,7 @@ import Ouroboros.Consensus.Shelley.Protocol
     ( TPraosCrypto )
 
 import qualified Cardano.Wallet.Api.Types as Api
+import qualified Data.List as L
 import qualified Data.Map.Merge.Strict as Map
 import qualified Data.Map.Strict as Map
 import qualified Data.Text as T
@@ -158,8 +160,7 @@ newStakePoolLayer
     -> NetworkLayer IO (IO Shelley) (CardanoBlock sc)
     -> DBLayer IO
     -> StakePoolLayer
-newStakePoolLayer
-    gp NetworkLayer{stakeDistribution,currentNodeTip} db@DBLayer {..} =
+newStakePoolLayer gp nl db@DBLayer {..} =
         StakePoolLayer
             { getPoolLifeCycleStatus = _getPoolLifeCycleStatus
             , knownPools = _knownPools
@@ -182,9 +183,10 @@ newStakePoolLayer
         -> Coin
         -> ExceptT ErrNetworkUnavailable IO [Api.ApiStakePool]
     _listPools currentEpoch userStake = do
-        tip <- liftIO getTip
-        lsqData <- combineLsqData <$> stakeDistribution tip userStake
+        tip <- withExceptT fromErrCurrentNodeTip $ currentNodeTip nl
+        lsqData <- combineLsqData <$> stakeDistribution nl tip userStake
         dbData <- liftIO $ readPoolDbData db
+        pp <- liftIO $ getProtocolParameters nl
         -- TODO:
         -- Use a more efficient way of filtering out retired pools.
         -- See: https://jira.iohk.io/projects/ADP/issues/ADP-383
@@ -193,8 +195,13 @@ newStakePoolLayer
             . filter (not . poolIsRetired)
             . map snd
             . Map.toList
-            $ combineDbAndLsqData (slotParams gp) lsqData dbData
+            $ combineDbAndLsqData (slotParams gp, pp) lsqData dbData
       where
+        fromErrCurrentNodeTip :: ErrCurrentNodeTip -> ErrNetworkUnavailable
+        fromErrCurrentNodeTip = \case
+            ErrCurrentNodeTipNetworkUnreachable e -> e
+            ErrCurrentNodeTipNotFound -> ErrNetworkUnreachable "tip not found"
+
         epochIsInFuture :: EpochNo -> Bool
         epochIsInFuture = (> currentEpoch)
 
@@ -206,8 +213,6 @@ newStakePoolLayer
         poolRetirementEpoch p = p
             & view #retirement
             & fmap (view (#epochNumber . #getApiT))
-
-    getTip = liftIO $ unsafeRunExceptT currentNodeTip
 
 --
 -- Data Combination functions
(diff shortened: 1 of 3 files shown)


lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -207,7 +219,42 @@ newStakePoolLayer
             & view #retirement
             & fmap (view (#epochNumber . #getApiT))
 
-    getTip = liftIO $ unsafeRunExceptT currentNodeTip
```
KtorZ, line 210: Not sure what this `unsafeRunExceptT` was doing here, but nothing good could come from it for sure.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -207,7 +219,42 @@ newStakePoolLayer
...
+        -- Sort by non-myopic member rewards, making sure to also randomly sort
+        -- pools that have equal rewards (hence the IO).
+        --
+        -- NOTE: we discard the final value of the random generator because we
+        -- do actually want the order to be stable between two identical
+        -- requests. The order simply needs to be different between _wallets_.
```
Anviking: The pool db, and thus the seed is shared for all wallets on a single wallet installation, so `_wallets_` should be something else, right?

KtorZ: Right, it's actually _server_. The server will return a consistent ordering (which is what we probably want to not be really confusing). Yet, between different user machines, the seed will be different.

Anviking:
As said in DM:

I tried the ranking and it looks better:
<details>
<summary>
Top and bottom pools from Testnet v6, using this PR, and 10 ada stake
</summary>

**But ranking will change, so I guess the screenshots can't be used for comparison**

Skärmavbild 2020-07-27 kl  22 26 52
Skärmavbild 2020-07-27 kl  22 26 57

</details>

- Looks roughly similar to how it was before
- But it is no longer sorted by `poolId` (you could previously clearly tell e.g. by looking at the bottom pools with, presumably, all 0 non-myopic member rewards)

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -207,7 +219,43 @@ newStakePoolLayer
...
+            -- an weight that was randomly attributed to them. The weight may
+            -- still equal (in which case the order will depend on the sortBy
+            -- implementation, but that is fine).
+            fn :: (Int, Api.ApiStakePool) -> (Int, Api.ApiStakePool) -> Ordering
+            fn (weightA, poolA) (weightB, poolB)
+                | rewards poolA > rewards poolB = LT
+                | rewards poolA < rewards poolB = GT
+                | otherwise {- == -}            = compare weightA weightB
+              where
+                rewards = view (#metrics . #nonMyopicMemberRewards)
```
jonathanknowles: **Suggestion:** without changing the overall time complexity of O (n log n), we can simplify this code by composing two simpler sorts together:

**Sort # 1**: Randomize the order.
**Sort # 2**: Sort by descending order of rewards.

So, for example (see 1bb3bc4d2de3b282c0316016ea75fd9bc5e4d09c):

```hs
        sortByReward g0 =
            map stakePool
            . L.sortOn (Down . rewards . stakePool)
            . L.sortOn randomWeight
            . evalState' g0
            . traverse withRandomWeight
          where
            evalState' :: s -> State s a -> a
            evalState' = flip evalState

            withRandomWeight :: RandomGen g => a -> State g (Int, a)
            withRandomWeight a = (, a) <$> state random

            (randomWeight, stakePool) = (fst, snd)
            rewards = view (#metrics . #nonMyopicMemberRewards)
```

As a result, we can remove the separate sorting function `fn`.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -182,19 +187,26 @@ newStakePoolLayer
...
+        lsqData <- combineLsqData <$> stakeDistribution nl tip userStake
         dbData <- liftIO $ readPoolDbData db
+        pp <- liftIO $ getProtocolParameters nl
+        seed <- liftIO $ atomically readSystemSeed
         -- TODO:
         -- Use a more efficient way of filtering out retired pools.
         -- See: https://jira.iohk.io/projects/ADP/issues/ADP-383
         return
-            . sortOn (Down . (view (#metrics . #nonMyopicMemberRewards)))
+            $ sortByReward seed
```
jonathanknowles: **Suggestion**: use composition operator here.

```suggestion
            . sortByReward seed
```
Alternatively, change it to use a chain of `$`s.

Justification: consistency and readability. (The remainder of the expression uses a chain of `.` operators.)

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -207,7 +219,43 @@ newStakePoolLayer
             & view #retirement
             & fmap (view (#epochNumber . #getApiT))
 
-    getTip = liftIO $ unsafeRunExceptT currentNodeTip
+        -- Sort by non-myopic member rewards, making sure to also randomly sort
+        -- pools that have equal rewards (hence the IO).
```
jonathanknowles:
```suggestion
        -- pools that have equal rewards.
```

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -250,23 +300,35 @@ combineDbAndLsqData sp =
...
+            { nonMyopicMemberRewards = freshmenMemberRewards
             , relativeStake = minBound
             , saturation = 0
             }
 
-    bothPresent = zipWithMatched mkApiPool
+    -- To give a chance to freshly registered pools that haven't been part of
+    -- any leader schedule, we assign them the average reward of the top @k@
+    -- pools.
+    freshmenMemberRewards
```
jonathanknowles:
```suggestion
    freshmanMemberRewards
```
(singular rather than plural)

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -250,23 +300,35 @@ combineDbAndLsqData sp =
...
         -> PoolLsqData
         -> PoolDbData
         -> Api.ApiStakePool
-    mkApiPool
-        pid
-        (PoolLsqData prew pstk psat)
-        dbData
-        = Api.ApiStakePool
+    mkApiPool pid (PoolLsqData prew pstk psat) dbData =
+        Api.ApiStakePool
```
jonathanknowles, line 321: You could save a line here if you wanted to, without going over the line limit.
```suggestion
    mkApiPool pid (PoolLsqData prew pstk psat) dbData = Api.ApiStakePool
```

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -250,23 +300,35 @@ combineDbAndLsqData sp =
...
+    -- pools.
+    freshmenMemberRewards
+        = Quantity
+        $ average
+        $ L.take (fromIntegral $ desiredNumberOfStakePools pp)
+        $ L.sort
+        $ map (Down . getQuantity . nonMyopicMemberRewards)
+        $ Map.elems lsqData
+      where
+        average [] = 0
```
jonathanknowles, line 309: :+1:

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -207,7 +219,43 @@ newStakePoolLayer
...
+            withRandomWeight :: RandomGen g => a -> State g (Int, a)
+            withRandomWeight a = do
+                weight <- state random
+                pure (weight, a)
+
+            -- Sort pools by descending non-myopic member rewards (i.e. pools
+            -- with larger rewards are at the beginning of the list).
+            --
+            -- In case pools have exactly the same reward, then compare them via
+            -- an weight that was randomly attributed to them. The weight may
```
jonathanknowles:
```suggestion
            -- a weight that was randomly attributed to them. The weight may
```

jonathanknowles (approved):
This looks good to me.

I do have a few suggestions, including a simplification to the sorting function. (See review comments and 1bb3bc4d2de3b282c0316016ea75fd9bc5e4d09c.)

But given the time constraints, I think the existing solution is just fine. :+1:

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
KtorZ: Indeed, I ended up changing this :grimacing:

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
KtorZ: I do love newlines actually, Elm formatted me here :grimacing:

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
KtorZ:
```
sortOn :: ...

Elements are arranged from from lowest to highest, keeping duplicates in the order they appeared in the input.
```

Interesting, I always thought that this was an implementation detail, and not an actual property of the sort function here.

Issue #1965, opened 2020-07-27
Redeem ITN rewards from mainet wallet 


KtorZ:
# Context

- As a ADA stakeholder who participated in the ITN,
- I want to be able to redeem my rewards from the ITN to my mainnet wallet,
- Without having to send a transaction to my ITN account first.
- So that I can enjoy my ITN rewards. 

# Decision

- Make it possible to redeem rewards of a wallet A from the API using another wallet B. 

# Acceptance Criteria

1. It _must not_ be necessary to restore the ITN wallet in order to redeem its rewards. 

---

# Development 

- Add a new `withdrawal` field to Shelley transactions to replace the `?withdrawReward` and extend its behavior ==> #1967 

# QA 

- See documentation on:
    - [postTransaction](https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/postTransaction)
    - [postTransactionFee](https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/postTransactionFee)

- Added extensive test scenarios to cover for various cases. The test scenarios use newly introduced "rewardWallet" faucets which are wallets which come with pre-existing rewards. 

    - https://github.com/input-output-hk/cardano-wallet/blob/5af6e35a4f70ca8e1cebe413b19a99439e04addb/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs#L1258-L1287

    - https://github.com/input-output-hk/cardano-wallet/blob/5af6e35a4f70ca8e1cebe413b19a99439e04addb/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs#L1289-L1324

    - https://github.com/input-output-hk/cardano-wallet/blob/5af6e35a4f70ca8e1cebe413b19a99439e04addb/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs#L1326-L1357

    - https://github.com/input-output-hk/cardano-wallet/blob/5af6e35a4f70ca8e1cebe413b19a99439e04addb/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs#L1359-L1378

    - https://github.com/input-output-hk/cardano-wallet/blob/5af6e35a4f70ca8e1cebe413b19a99439e04addb/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs#L1380-L1398

darko-mijic:
Looking forward to this one. :)

[...]

PR #1967, opened 2020-07-28
Redeem rewards from another wallet

KtorZ:
# Issue Number

#1965 

# Overview

- 8f97407f299118fbf1bd00b1d01ae5f673100371
  :round_pushpin: **let the caller decide how to generate a reward account for signing withdrawals**
    The goal is to allow a mnemonic to be passed around as a function
  argument. If it is, then that account from the mnemonic is used.
  Otherwise, it will defaults to the wallet's own account.

- e26fc387183bf2f6a560edc1c3143fd9752ed10d
  :round_pushpin: **make it possible to query rewards of any arbitrary reward account**
    Yet, the server still query rewards from standard wallet reward
  account at the moment. Next commits will make it possible for users to
  pass a mnemonic to the API.

- 657e541319545f47d35bd5d742209007adc4eb5b
  :round_pushpin: **remove duplicated logic between 'readNextWithdrawal' and 'queryRewardBalance'**
    This partially revert some of the previous commit, but that's for the best.

- fc14a3b9e0beb87147f784eb1eded509fdf4c849
  :round_pushpin: **extend 'POST' transaction Shelley endpoint to support redeeming rewards**
    So, now the transaction endpoint accepts two kind of payloads, either a list of payments like before, or simply a source mnemonic from which rewards would be drained / withdrawn.

- 9161f9f7af25996fd655b60a95646a277373940f
  :round_pushpin: **update 'postTransaction' handler to handle payments and withdrawals**
  
- 7a71a7e1b72f2ed9258925d74314adaeaf9e8b11
  :round_pushpin: **separate 'toChimericAccount' from 'HasRewardAccount' type-class**
    This makes it slightly easier to deal with all these constraints in the API, in particular in the context of withdrawals.
  Also, I've completely removed the 'HasRewardAccount' constraint from the 'readChimericAccount' function. Instead, that function will
  dynamically check that the type is a Shelley type and fails with a proper error otherwise.

- 04c96ea50d65d924de4a42cb74ab6d608f1629a6
  :round_pushpin: **actually get rid of 'HasRewardAccount' type-class altogether**
    This type class only existed in order to be able to get a reward account from a sequential state to assess whether an account was ours or not.
  We had to defined an unsound 'HasRewardAccount' on 'IcarusKey' because of that and carry this instance in many places in the API. Instead, we
  can simply carry a 'IsOurs _ ChimericAccount' constraints up to API instantiation.

- 03da4762d23f1f1a7e663780bfd5961e627c7f18
  :round_pushpin: **unify withdrawal requests frmo self or external wallets under one common API**
    This is so-to-speak a breaking change compared to what we just released yet, it's for the best. Now there's one consistent way of creating withdrawals! This also gets rid of a few edge-case that are not possible anymore by design.

- 3e0e9db5bef5f8d0e2bcda67fc128d1e21812a71
  :round_pushpin: **add pre-funded wallets with rewards using MIR certificates to the integration cluster**
  
- 3183a7fbe1fb19370b4b66550a52c740378c5adb
  :round_pushpin: **add integration scenarios for reward redemption**
    Trying various scenarios, from self and from other wallets.

- d3ebc72b596c5408e86d569c6cb9fb192b677aa5
  :round_pushpin: **prevent users from withdrawing from empty reward accounts.**

# Comments

Screenshot from 2020-07-29 23-38-52 Screenshot from 2020-07-29 23-39-05


lib/core/src/Cardano/Wallet.hs
```diff
@@ -901,14 +907,19 @@ readNextWithdrawal
...
     => ctx
     -> WalletId
-    -> IO (Quantity "lovelace" Word64)
-readNextWithdrawal ctx wid = db & \DBLayer{..} -> do
-    (pp, withdrawal) <- atomically $ (,)
-        <$> readProtocolParameters pk
-        <*> fmap getQuantity (readDelegationRewardBalance pk)
+    -> Either WalletId SomeMnemonic
+    -> ExceptT ErrFetchRewards IO (Quantity "lovelace" Word64)
+readNextWithdrawal ctx wid src = db & \DBLayer{..} -> do
```
paweljakubas: `src` was unexpected name for me. tbh. maybe `acctSrc`, `chimericOrigin` ?

lib/core/src/Cardano/Wallet.hs
```diff
@@ -928,11 +939,38 @@ readNextWithdrawal ctx wid = db & \DBLayer{..} -> do
...
     minFee policy = fromIntegral . getFee . minimumFee tl policy Nothing
 
+    handleErr = \case
+        Right x -> Right x
+        Left ErrGetAccountBalanceAccountNotFound{} ->
+            Right $ Quantity 0
+        Left (ErrGetAccountBalanceNetworkUnreachable e) ->
+            Left $ ErrFetchRewardsNetworkUnreachable e
+
+    fromOwnWallet
```
paweljakubas: :+1:

lib/core/src/Cardano/Wallet.hs
```diff
@@ -928,11 +939,38 @@ readNextWithdrawal ctx wid = db & \DBLayer{..} -> do
...
+        :: WalletId
+        -> ExceptT ErrFetchRewards IO ChimericAccount
+    fromOwnWallet _wid = db & \DBLayer{..} -> do
```
paweljakubas: :+1:

lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -159,6 +165,25 @@ instance Functor m => Functor (NetworkLayer m target) where
         { nextBlocks = fmap (fmap f) . nextBlocks nl
         }
 
+-- Fetching the account balance may sometimes fail, especially on epoch
+-- boundaries. Therefore, we add some retrying mecanism here to cope with
```
paweljakubas:
```suggestion
-- boundaries. Therefore, we add some retrying mechanism here to cope with
```

lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -159,6 +165,25 @@ instance Functor m => Functor (NetworkLayer m target) where
...
+-- that nicely. 'readNextWithdrawal' is typically called by one-time
+-- handlers like 'postTransactions' and not by workers polling things
+-- regularly, so retrying a few times is okay here.
+getAccountBalanceRetry
+    :: NetworkLayer IO target block
+    -> ChimericAccount
+    -> ExceptT ErrGetAccountBalance IO (Quantity "lovelace" Word64)
+getAccountBalanceRetry nl account = ExceptT $
+    retrying policy decision (const $ runExceptT $ getAccountBalance nl account)
```
paweljakubas: I assume here, that there will be a moment, where cardano-node will be able to discover itn rewards basing on chimeric account? so itn and shelley chimeric account will be married, the same format, etc ? and we can reuse the same local query infrastructure already present?

paweljakubas:
so far so good (although I liked more `fromOwnWallet` and `fromExternalWallet` and they changed in last commit ;-) 
of course, waiting for proof of concept in integration testing (the question is whether itn has been absorbed already in cardano-node and it is possible now )

lib/core/src/Cardano/Wallet.hs
KtorZ: Good point. My creativity for names at 4 a.m. was somewhat diminished :joy:

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -491,10 +499,27 @@ data PostTransactionData (n :: NetworkDiscriminant) = PostTransactionData
     , passphrase :: !(ApiT (Passphrase "lenient"))
     } deriving (Eq, Generic, Show)
 
+data PostWithdrawalData = PostWithdrawalData
+    { source :: !(ApiMnemonicT '[15,18,21,24])
```
paweljakubas: why not to use `mnemonicSentence` like in other places?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -491,10 +499,27 @@ data PostTransactionData (n :: NetworkDiscriminant) = PostTransactionData
...
+    , passphrase :: !(ApiT (Passphrase "lenient"))
+    } deriving (Eq, Generic, Show)
+
+newtype PostPaymentOrWithdrawalData n = PostPaymentOrWithdrawalData
+    { postData :: Either PostWithdrawalData (PostTransactionData n)
+    } deriving (Eq, Generic, Show)
+
+newtype PostWithdrawalFeeData = PostWithdrawalFeeData
+    { source :: ApiMnemonicT '[15,18,21,24]
```
paweljakubas: here also, why not to use `mnemonicSentence` like in other places?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1144,11 +1169,47 @@ instance DecodeAddress t => FromJSON (PostTransactionData t) where
...
+instance ToJSON PostWithdrawalData where
+    toJSON = genericToJSON defaultRecordTypeOptions
+
+instance EncodeAddress t => ToJSON (PostPaymentOrWithdrawalData t) where
+    toJSON (PostPaymentOrWithdrawalData (Left x))  = toJSON x
+    toJSON (PostPaymentOrWithdrawalData (Right x)) = toJSON x
+instance DecodeAddress t => FromJSON (PostPaymentOrWithdrawalData t) where
+    parseJSON val = flip (withObject "PostPaymentOrWithdrawalData") val $ \o -> do
+        case (HM.lookup "source" o, HM.lookup "payments" o) of
+            (Just{}, Just{})   -> empty
```
paweljakubas: :+1:

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1144,11 +1169,47 @@ instance DecodeAddress t => FromJSON (PostTransactionData t) where
...
+instance FromJSON PostWithdrawalFeeData where
+    parseJSON = genericParseJSON defaultRecordTypeOptions
+instance ToJSON PostWithdrawalFeeData where
+    toJSON = genericToJSON defaultRecordTypeOptions
+
+instance EncodeAddress t => ToJSON (PostPaymentOrWithdrawalFeeData t) where
+    toJSON (PostPaymentOrWithdrawalFeeData (Left x))  = toJSON x
+    toJSON (PostPaymentOrWithdrawalFeeData (Right x)) = toJSON x
+instance DecodeAddress t => FromJSON (PostPaymentOrWithdrawalFeeData t) where
+    parseJSON val = flip (withObject "PostPaymentOrWithdrawalFeeData") val $ \o -> do
```
paweljakubas: what about reusing what was done in impl of `instance DecodeAddress t => FromJSON (PostPaymentOrWithdrawalData t)` here?

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -385,17 +374,17 @@ prefilterBlock b u0 = runState $ do
...
+        :: (IsOurs s Address)
         => ([(Tx, TxMeta)], UTxO)
         -> Tx
         -> State s ([(Tx, TxMeta)], UTxO)
     applyTx (!txs, !u) tx = do
         ourU <- state $ utxoOurs tx
         let ourIns = Set.fromList (inputs tx) `Set.intersection` dom (u <> ourU)
         let u' = (u <> ourU) `excluding` ourIns
-        ourWithdrawals <- mapMaybeM ourChimericAccount $ Map.toList $ withdrawals tx
+        let wdrls = fromIntegral . getCoin <$> Map.elems (withdrawals tx)
```
paweljakubas, line 385: could you explain this change?

lib/core/src/Cardano/Wallet/Network.hs
KtorZ: Indeed. In principle, on the 3rd of August, rewards will be transferred to the Mainnet based on the same public key that was used on the ITN.

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: Because the JSON object isn't exactly the same (FeeData has no passphrase).

lib/core/src/Cardano/Wallet/Primitive/Model.hs
KtorZ: Yes. Since we can now have withdrawals in our wallet that aren't from our chimeric account (but will still contribute  positively to the input balance of a transaction), we need to account for pretty much all withdrawals in transactions (or otherwise, more complicated, "remember" all chimeric account we've made withdrawals from...). 

This will work fine until we introduce multisig (where some inputs will not be ours... right now, if one input is ours, then they all are so the problem is quite simplified). By then, we'll have to review this.

specifications/api/swagger.yaml
```diff
@@ -1270,13 +1270,35 @@ components:
           <<: *lenientPassphrase
           description: The wallet's master passphrase.
 
+    ApiPostWithdrawalData: &ApiPostWithdrawalData
+      type: object
+      required:
+        - source
```
paweljakubas: we use `- mnemonic_sentence` all around for this bit

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -1240,6 +1254,148 @@ spec = do
             [ expectResponseCode @IO HTTP.status404
             , expectErrorMessage (errMsg404NoWallet wid)
             ]
+
+    it "SHELLEY_TX_REDEEM_01 - Can redeem rewards from self" $ \ctx -> do
+        (wSrc,_) <- rewardWallet ctx
```
KtorZ, line 1259: cc @piotr-iohk

KtorZ:
bors try

lib/byron/test/integration/Cardano/Wallet/Byron/Faucet.hs
```diff
@@ -11,11 +11,12 @@ import Prelude
...
 import Test.Integration.Faucet
-    ( Faucet (..), icaMnemonics, rndMnemonics, seqMnemonics )
+    ( Faucet (..), icaMnemonics, mirMnemonics, rndMnemonics, seqMnemonics )
 
 initFaucet :: IO Faucet
 initFaucet = Faucet
     <$> newMVar seqMnemonics
     <*> newMVar icaMnemonics
     <*> newMVar rndMnemonics
+    <*> newMVar mirMnemonics
```
paweljakubas, line 21: :+1:

KtorZ: I generated 10 of them, used 2 or 3 I think in the scenarios I wrote, so there is still some room for some :p

lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -785,6 +786,27 @@ emptyWalletWith ctx (name, passphrase, addrPoolGap) = do
     expectResponseCode @IO HTTP.status201 r
     return (getFromResponse id r)
 
+rewardWallet :: Context t -> IO (ApiWallet, Mnemonic 24)
```
paweljakubas, line 789: why not to call it `readMIRwallet` to better reflect its nature?

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -1240,6 +1254,148 @@ spec = do
...
+        let payload = Json [json|{
+                "withdrawal": #{mnemonicToText mw},
+                "payments": [{
+                    "address": #{addr},
+                    "amount": { "quantity": 1, "unit": "lovelace" }
+                }],
+                "passphrase": #{fixturePassphrase}
+            }|]
+
+        -- Withdraw rewards from the other wallet.
```
paweljakubas, line 1340: :+1:

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -1240,6 +1254,148 @@ spec = do
...
+            (Link.createTransaction @'Shelley wSelf) Default payload
+        eventually "rewards disappear from other" $ do
+            rWOther <- request @ApiWallet ctx
+                (Link.getWallet @'Shelley wOther) Default payload
+            verify rWOther
+                [ expectField (#balance . #getApiT . #reward)
+                    (`shouldBe` Quantity 0)
+                ]
+
+        -- Try withdrawing AGAIN, rewards that aren't there anymore.
```
paweljakubas, line 1351: :+1:

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -1240,6 +1254,148 @@ spec = do
...
+            }|]
+
+        rTx <- request @(ApiTransaction n) ctx
+            (Link.createTransaction @'Shelley wSelf) Default payload
+        verify rTx
+            [ expectResponseCode HTTP.status403
+            , expectErrorMessage errMsg403WithdrawalNotWorth
+            ]
+
+    it "SHELLEY_TX_REDEEM_05 - Can't redeem rewards using byron wallet" $ \ctx -> do
```
paweljakubas, line 1380: :+1:

lib/core/src/Cardano/Wallet.hs
```diff
@@ -904,61 +920,70 @@ readNextWithdrawal
...
 
-            let costOfWithdrawal =
+                costOfWithdrawal =
                     minFee policy (mempty { withdrawal })
                     -
                     minFee policy mempty
 
-            pure $ if toInteger withdrawal < 2 * costOfWithdrawal
+            in
+                if toInteger withdrawal < 2 * costOfWithdrawal
```
paweljakubas, line 939: could you recall me why we have here `2 * costOfWithdrawal` , ie., why `2` ?

lib/core-integration/src/Test/Integration/Framework/DSL.hs
KtorZ: I was shared between calling everything `MIR` (which is sort of cardano-node specific terminology) or simply `Reward` (which is more generic). 

I ended up referring to "MIR" in the setup code, because it makes it clearer to what it connects behind the scene, but resorted to "reward" for everything in the DSL and scenarios. That the rewards come from MIR certificates is sort of an implementation detail.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2396,6 +2453,16 @@ instance LiftHandler ErrNotASequentialWallet where
                 , "Make sure to use a sequential wallet style, like Icarus."
                 ]
 
+instance LiftHandler ErrWithdrawalNotWorth where
+    handler = \case
+        ErrWithdrawalNotWorth ->
+            apiError err403 WithdrawalNotWorth $ mconcat
+                [ "I've noticed that you're requesting a withdrawal from an "
+                , "account that is either empty or doesn't have a balance big "
```
paweljakubas, line 2461: it handles also situations, where we got already rewards... saying that account is empty covers this. We don't want to say explicitly something like : `presumably the account was already withdrawn` ?

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -315,6 +318,12 @@ class HardDerivation key => SoftDerivation (key :: Depth -> * -> *) where
         -> Index 'Soft 'AddressK
         -> key 'AddressK XPub
 
+-- | Derivation of a reward account, as a type-class because different between
+-- key types (in particular, Jörmungandr vs Shelley).
+class ToChimericAccount k where
```
paweljakubas, line 323: this should also be reflected in `cardano-addresses` ?

lib/core/src/Cardano/Wallet.hs
KtorZ: The rationale was that: "withdrawal should bring **at least much value** as they cost.". 

This is to prevent withdrawals from either loosing money, or being consumed too early when they're really small (imagine an account with `0.31` Ada, and the withdrawal cost `0.3` Ada. If we were just ensuring the withdrawal to be `> costOfWithdrawal` (without `2 *`), then the resulting withdrawal would be `0.01` Ada :/ .. And it really feels not worth it given the fees. Hence the times two.

lib/core/src/Cardano/Wallet/Api/Server.hs
KtorZ: Well, an account could be empty for two reasons: either it hasn't received rewards yet, or it has already withdrawn them. I didn't feel like emphasizing on any of these aspects but we could indeed.

lib/shelley/test/data/cardano-node-shelley/shelley-genesis.yaml
```diff
@@ -40,7 +40,7 @@ genDelegs:
   8ae01cab15f6235958b1147e979987bbdb90788f7c4e185f1632427a:
     delegate: b7bf59bb963aa785afe220f5b0d3deb826fd0bcaeeee58cb81ab443d
     vrf: 4ebcf8b4c13c24d89144d72f544d1c425b4a3aa1ace30af4eb72752e75b40d3e
-updateQuorum: 5
+updateQuorum: 1
```
paweljakubas, line 43: could you explain why this change was needed?

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
KtorZ: cardano-addresses doesn't actually treat with chimeric account or stake address at all at the moment. We should eventually though.

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -1229,6 +1232,12 @@ instance Arbitrary a => Arbitrary (AddressAmount a) where
     shrink _ = []
 
 instance Arbitrary (PostTransactionData t) where
+    arbitrary = PostTransactionData
+        <$> arbitrary
+        <*> arbitrary
+        <*> elements [Just SelfWithdrawal, Nothing]
```
paweljakubas, line 1238: we don't want to include external withdrawals here?

paweljakubas (approved):
very persuasive test scenarios! I like refactoring of `ApiPostTransactionData` and how posting tx api changed, ie that `withdrawal` was embedded in `PostTransactionData***`. we have breaking change, but I believe it is worth it

lib/shelley/test/data/cardano-node-shelley/shelley-genesis.yaml
KtorZ: Ah! Took me some time to figure that out (still dreaming about a documentation of the genesis parameters one day :pray:). 
MIR certificates are used to transfer rewards to pre-registered stake key. This is not an operation that _anyone_ can do (that would be sort of problematic :sweat_smile:), but only the core delegates can (same goes for other "governance" operations like updating the protocol parameters). 

This `updateQuorum` parameters controls how many genesis delegate signatures are required for a governance type of request to be considered valid on-chain. Since we only have a single BFT leader in this setup, well, we can only have at most one signature! Thus I had to tweak this so that our BFT leader could act in full dictatorial mode :)

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
paweljakubas: yes, I was saying about that. We need to remember this in the context of `cardano-addresses`

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
KtorZ: This is a bit tricky because the swagger2 Haskell libraries doesn't support the `oneOf` constructor :( ... And here, the swagger schema is a oneOf between a string and a mnemonic sentence; so I had to pick one to not overly complexifies things here; since the mnemonic sentences schema and JSON are already well-tested through other API types, I opted for the self withdrawal one which is completely new and only exists in this context.

PR #1968, opened 2020-07-28
Test that it's not possible to join a retired stake pool.

jonathanknowles:
# Issue Number

#1853 

# Overview

This PR adds an integration test to demonstrate that a wallet cannot join a stake pool that has already retired.

# Comments

Note that the integration test included in this PR is a compromise.

Ideally, a more complete integration test would verify that staking to a pool is still possible right up to the moment before the pool has actually retired. See: https://jira.iohk.io/browse/ADP-385


jonathanknowles:
bors try

paweljakubas (approved):

KtorZ:
> https://github.com/input-output-hk/cardano-wallet/pull/1968#issuecomment-664866562

I think all the BYRON_ scenarios in the API are still using the non-retryable `it`. They should instead use:

https://github.com/input-output-hk/cardano-wallet/blob/master/lib/test-utils/src/Test/Hspec/Extra.hs#L104-L109

piotr-iohk (approved):
lgtm.

PR #1969, opened 2020-07-28
Add more HasCallStack to shelley network layer

rvl:
### Issue Number

Relates to #1708

### Overview

- Adds HasCallStack annotations where possible, based on the principle that future bugs often appear in the same place as past bugs.
- Make the parallel socket connections test more robust by removing the threadDelay.

### Comments

- I was unable to add `HasCallStack` to `connectClient`. This requires a change in the ouroboros-network library.


lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -782,10 +801,11 @@ connectClient tr handlers client (vData, vCodec) addr = withIOManager $ \iocp ->
     policy :: RetryPolicyM IO
     policy = fibonacciBackoff 250_000 & capDelay 2_000_000
 
+-- | Shorthand for the list of exception handlers used with 'recovering'.
+type RetryHandlers = [RetryStatus -> Handler IO Bool]
```
paweljakubas, line 805: :+1:

paweljakubas (approved):
cool!

KtorZ (approved):

PR #1973, opened 2020-07-28
do not migrate hardfork column as non-null

KtorZ:
# Issue Number

N/A

# Overview

- [ ] I have adjusted the migration of the hardfork countdown column.

# Comments

```
'hardfork_epoch'. Adding this field with a default value of NULL.
SQLite3 returned ErrorError while attempting to perform prepare "ALTER TABLE protocol_parameters ADD COLUMN hardfork_epoch INTEGER NOT NULL DEFAULT NULL ;": Cannot add a NOT NULL column with default value NULL
```


piotr-iohk (approved):
Migration error is not occurring any longer.

paweljakubas (approved):
fixed the issue

PR #1975, opened 2020-07-28
Update docker-compose.yml to work with recent testnets and staging networks.

melmccann:
Should we use cardano-wallet-shelley instead of cardano-wallet-byron?


KtorZ:
Indeed. That was going to be part of our next steps.

KtorZ (approved):

PR #1977, opened 2020-07-28
Test getNetworkInformation handling of PastHorizonException

Anviking:
# Issue Number

#1869 /  #1960

# Overview

- [x] Added a mock interpreter & network layer to produce a single unit test
- [x] I think we can make this a monadic property test easily

# Comments

### DIFF
--- a/lib/byron/src/Cardano/Wallet/Byron/Api/Server.hs
+++ b/lib/byron/src/Cardano/Wallet/Byron/Api/Server.hs
@@ -285,12 +285,12 @@ server byron icarus ntp =
 
     network :: Server Network
     network =
-        getNetworkInformation genesis nl
+        getNetworkInformation syncTolerance nl
         :<|> getNetworkParameters genesis nl
         :<|> getNetworkClock ntp
       where
         nl = icarus ^. networkLayer @t
-        genesis = icarus ^. genesisData
+        genesis@(_,_,syncTolerance) = icarus ^. genesisData
 
     proxy :: Server Proxy_
     proxy = postExternalTransaction icarus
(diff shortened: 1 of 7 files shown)


rvl (approved):
I like that we now have this test.

But two things don't sit right with me.
1. We have started using cardano-node/shelley specific libraries in cardano-wallet-core. Perhaps this is unavoidable though.
2. The amount of effort required to set up a simple test is quite high, don't you think? At least the handler function can be tested independently of the API server, but still.

Anviking:
> We have started using cardano-node/shelley specific libraries in cardano-wallet-core. Perhaps this is unavoidable though.

Yes. I think this is unavoidable for now.

> At least the handler function can be tested independently of the API server, but still.

Yes, I was pleasantly surprised that this was possible.

> The amount of effort required to set up a simple test is quite high, don't you think?

Yes. 

1. Constructing the `Interpreter` is a bit complicated, but not sure what the workaround would be.
2. Having to construct a `NetworkLayer` just to mock the `currentNodeTip` is also a bit annoying
3. Having to mimic the exact `Either PastHorizon ~> IO` using `throwIO` that the implementation uses (I think passing around `TimeInterpreter (ExcepT PastHorizonException IO)` would be better for this reason)

We could alleviate 2 by adding a new helper
```haskell
getNetworkInformation
    :: forall t. (HasCallStack)
    => SyncTolerance
    -> NetworkLayer IO t Block
    -> Handler ApiNetworkInformation

getNetworkInformation'
    :: forall t. (HasCallStack)
    => SyncTolerance
    -> TimeInterpreter IO
    -> SlotNo -- ^ Tip
    -> Handler ApiNetworkInformation
```
which I'm intrigued by, but maybe it should be done as part of some greater, consistent re-organization of the Server module.

We already have several mock layer records with mostly unimplemented fields. As we add more, we might start seeing the patterns.

Anviking:
Thanks! Merging. 

bors r+

Issue #1981, opened 2020-07-29
Don't fetch nOpt another time, separately from stake distribution


Anviking:
# Context

ADP-388

# Decision

Re-use the existing nOpt.

# Acceptance Criteria

- `nOpt` is not fetched again with `getProtocolParameters`

---

# Development 

See decision.

# QA 

- This issue is a tiny cleanup / refactoring.

piotr-iohk:
ok.

PR #1982, opened 2020-07-29
Don't fetch nOpt another time

Anviking:
# Issue Number

#1981 

# Overview

- [x] Re-use the existing `nOpt` instead of fetching it again

# Comments

### DIFF
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
@@ -187,9 +187,9 @@ newStakePoolLayer nl db@DBLayer {..} =
         -> ExceptT ErrNetworkUnavailable IO [Api.ApiStakePool]
     _listPools currentEpoch userStake = do
         tip <- withExceptT fromErrCurrentNodeTip $ currentNodeTip nl
-        lsqData <- combineLsqData <$> stakeDistribution nl tip userStake
+        rawLsqData <- stakeDistribution nl tip userStake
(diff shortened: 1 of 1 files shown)


lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -272,11 +276,11 @@ data PoolDbData = PoolDbData
 combineDbAndLsqData
     :: forall m. (Monad m)
     => TimeInterpreter m
-    -> ProtocolParameters
+    -> Int -- ^ nOpt; desired number of pools
```
Anviking, line 279: Do we need a `newtype DesiredNumberOfPools`? 

I'm not sure the need is really there yet, and making all the other packages compile again can be slightly annoying, so I didn't do it.

paweljakubas: hmm, I do not see this newtype: 
```
pawel@pure:~/Work/IOHK/cardano-wallet (master)$ ag DesiredNumberOfPools
lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
168:    protocolParametersDesiredNumberOfPools  Word16          sql=desired_pool_number

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
450:        addColumn conn True (DBField ProtocolParametersDesiredNumberOfPools) value
```

paweljakubas (approved):
one less rountrip. lgtm !

Issue #1872, opened 2020-07-07
Remove `GenesisParameters` from wallet checkpoints


Anviking:
# Context

In ADP-356 we are integrating with Hard Fork Combinator.

In each wallet checkpoint we are currently storing:

```
checkpointGenesisHash       BlockId      sql=genesis_hash
checkpointGenesisStart      UTCTime      sql=genesis_start
checkpointFeePolicyUnused   Text         sql=fee_policy
checkpointSlotLength        Word64       sql=slot_length
checkpointEpochLength       Word32       sql=epoch_length
checkpointTxMaxSizeUnused   Word16       sql=tx_max_size
checkpointEpochStability    Word32       sql=epoch_stability
checkpointActiveSlotCoeff   Double       sql=active_slot_coeff
```

Note that
- `epochLength`, `slotLength` may change in the byron-to-shelley transition, but I don't think wallet checkpoints are suitable for keeping track of this.
- `epochStability` (i.e. `k`/ `securityParam`) is completely static
- We store recent `TxParameters` in a separate table
- We verify the `genesisHash` of the stored genesis checkpoint (seems useful)

# Decision

I think we should remove everything but `genesisHash`.

# Acceptance Criteria

- [ ] Redundant and potentially confusing information should be removed from wallet checkpoints

---

# Development 

#1983

# QA

Anviking:
Maybe we should abort this for now because of https://input-output-rnd.slack.com/archives/GBT05825V/p1596197796271100?thread_ts=1596197783.271000&cid=GBT05825V

rvl:
Yes when we have a better db migration system we can revisit this.

PR #1983, opened 2020-07-29
Don't store GenesisParameters in checkpoints

Anviking:
# Issue Number

#1872 

# Overview

- [x] Remove all genesis params but genesis hash from checkpoints
- [x] Pass in the `EpochStability`/a.k.a `k` to the DB, from the `GenesisParameters` we read at startup

# Comments

```
Most GenesisParameters are unneeded in the checkpoints, and storing them
there may give the impression that they may be changed by protocol
updates and forks. They won't. (There is a separate ProtocolParameters
table, however)

This commit keeps the genesis hash in the checkpoints, such that we
still can validate the integrity of the db.

For pruning checkpoints, the db needs the security parameter k. Since we
no longer can read it from the DB, we pass it in when constructing the
db.
```

### DIFF
--- a/lib/byron/bench/Restore.hs
+++ b/lib/byron/bench/Restore.hs
@@ -368,7 +368,7 @@ bench_restoration _proxy tracer socketPath np vData progressLogFile (wid, wname,
         let gp = genesisParameters np
         let convert = fromByronBlock gp
         let nw = convert <$> nw'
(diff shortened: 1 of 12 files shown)


Anviking:
Hm. I think manual migration might be required. @rvl Do you think #1872 is worth the effort?

Anviking:
Closing for now since manual migration seems tricky.

Issue #1984, opened 2020-07-30
Impose minimum execution time on `withRootKey` in the event of failure.


jonathanknowles:
# Context

For API operations that require the user to specify their passphrase, we wish to impose a minimum response time in the event of an authentication failure, where the minimum response time is:

- significantly longer than the actual time required to service the request (seconds, rather than milliseconds).
- independent of the actual time required to service the request (so that no timing information is leaked to the caller).

## Affected Operations

The following operations would be affected:

- Addresses
    - PostByronAddress
- Migrations
    - MigrateByronWallet
    - MigrateShelleyWallet
- Passphrases
    - PutWalletPassphrase
    - PutByronWalletPassphrase
- StakePools
    - JoinStakePool
    - QuitStakePool
Transactions
    - CreateTransaction
    - CreateByronTransaction

# Decision

Adjust the `withRootKey` function to incorporate a minimum execution time in the event of failure.

# Acceptance Criteria

- [x] In the event of a _successful_ authentication, there must be _no_ delay.
- [x] In the event of an _unsuccessful_ authentication, the total execution time of the `withRootKey` function must be:
    - [x] _equal than or greater to_ a defined minimum execution time constant.
    - [x] _independent_ of the actual time required to perform the authentication.

# Development 

- https://github.com/input-output-hk/cardano-wallet/pull/1985

# QA 

- [x] Add integration test to confirm that an _unsuccessful_ attempt to authenticate results in an appropriate delay.
- [x] Add integration test to confirm that a _successful_ attempt to authenticate does _not_ result in such a delay.

PR #1985, opened 2020-07-30
Minimum execution time for `withRootKey` on authentication failure.

jonathanknowles:
# Issue Number

#1984 

# Overview

This PR adjusts the `withRootKey` function to incorporate a minimum execution time in the event of failure.

- In the event of a _successful_ authentication, there is _no_ delay.
- In the event of an _unsuccessful_ authentication, the total execution time of the `withRootKey` function is:
    -  equal to (or greater than) `minimumExecutionTimeOnAuthFailure`.
    -  independent of the actual time required to perform the authentication.

Where `minimumExecutionTimeOnAuthFailure` is defined as:
```hs
-- | Specifies a minimum bound on the length of time that is allowed to elapse
--   before an authentication failure can be reported to the user.
--
minimumExecutionTimeOnAuthFailure :: NominalDiffTime   
minimumExecutionTimeOnAuthFailure = threeSeconds   
  where   
    threeSeconds = fromRational 3   
```

In the future, we could make this configurable (if the need arises).

# Testing

- [x] Add integration test to confirm that an unsuccessful attempt to authenticate results in a delay of `minimumExecutionTimeOnAuthFailure` or longer.
- [x] Add integration test to confirm that a successful attempt to authenticate does not result in such a delay.

### DIFF
--- a/lib/core/src/Data/Time/Utils.hs
+++ b/lib/core/src/Data/Time/Utils.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE NumericUnderscores #-}
+
 -- |
 -- Copyright: © 2018-2020 IOHK
 -- License: Apache-2.0
@@ -5,14 +7,21 @@
 -- Utility functions for manipulating time values.
 
 module Data.Time.Utils
-    ( utcTimePred
+    ( nominalDiffTimeToMicroseconds
+    , utcTimePred
     , utcTimeSucc
     ) where
 
 import Prelude
 
 import Data.Time
-    ( UTCTime, addUTCTime )
+    ( NominalDiffTime, UTCTime, addUTCTime )
+
+-- | Converts the specified time difference into an integral number of
+--   microseconds.
+--
+nominalDiffTimeToMicroseconds :: Integral i => NominalDiffTime -> i
+nominalDiffTimeToMicroseconds = ceiling . (* 1_000_000) . toRational
 
 -- | For a given time 't0', get the closest representable time 't1' to 't0'
 --   for which 't0 < t1'.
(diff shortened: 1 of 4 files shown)


jonathanknowles:
bors try

jonathanknowles:
bors try

jonathanknowles:
bors try

lib/core/src/Data/Time/Utils.hs
```diff
@@ -23,3 +37,16 @@ utcTimeSucc = addUTCTime $ succ 0
...
+-- | Suspends the current thread until after the specified time has passed.
+--
+-- There is no guarantee that the thread will be rescheduled promptly when the
+-- delay has expired, but the thread will never continue to run earlier than
+-- specified.
+--
+waitUntil :: MonadIO m => UTCTime -> m ()
+waitUntil futureTime = liftIO $ do
+    currentTime <- getCurrentTime
+    threadDelay
```
rvl, line 50: Does this work OK when the difference is negative?
Perhaps you could put a `max 0` there to make sure.

rvl (commented):
It looks OK at first glance if we are going to implement this.

lib/core/src/Data/Time/Utils.hs
jonathanknowles: > Does this work OK when the difference is negative?
> Perhaps you could put a `max 0` there to make sure.

I think it should work, as informal testing of `threadDelay` shows that it exits immediately when called with a negative value.

However, this behaviour appears to be undocumented, and so I agree it's a good idea to use `max 0`, as then we won't have to rely on undocumented behaviour.

Fixed in bcdeeb8d8327d9b91aacbd6ec71e5545547c4989.

KtorZ:
Closing for now. Will re-open once we've got feedback from the security auditors.

Issue #1939, opened 2020-07-23
Provide more information on failure to determine current epoch.


jonathanknowles:
# Context

As part of the work to track stake pool retirements, we must reject attempts to join stake pools that have already retired.

In order to do this, we must compare the pool's retirement epoch with the _current_ epoch.

However, determining the current epoch is an operation that might fail:

https://github.com/input-output-hk/cardano-wallet/blob/05d377418b02e59e1299a0eaee135802c52735d2/lib/core/src/Cardano/Wallet/Primitive/Slotting.hs#L108-L109

If we encounter this condition, we currently report the following error:

https://github.com/input-output-hk/cardano-wallet/blob/05d377418b02e59e1299a0eaee135802c52735d2/lib/core/src/Cardano/Wallet/Api/Server.hs#L1902-L1906

# Decision

Determine the conditions under which this operation may fail, and if necessary, update the error message to be more informative.

# Acceptance Criteria

- [x] The error message associated with `ErrUnableToDetermineCurrentEpoch` should describe a course of action for the user to take.

# Development 

See https://github.com/input-output-hk/cardano-wallet/pull/1992

# QA 

See https://github.com/input-output-hk/cardano-wallet/pull/1992

jonathanknowles:
@Anviking Do you have any opinion on how we might make this particular error more informative?

Anviking:
It's possible with #1943, with the change from `TimeInterpreter IO` to `TimeInterpreter (ExceptT IO PastHorizonException)` (Sorry, I think I indicated some time that the `Nothing` of `currentEpoch` could "include" `PastHorizonException`, but that is not the case.)

There should be reasons for failure:
- Current time is behind genesis. Action: 🤷‍♂️
- Node is in the Byron era of Byron;Shelley, and the current time is beyond our forecast zone. Action: Wait for the node to sync more

A concrete example of `PastHorizonException` behaviour [here](https://input-output-rnd.slack.com/archives/CFKLUH4R0/p1595511029493900), in case it is useful for context.

PR #1992, opened 2020-08-03
Add user advice to `UnableToDetermineCurrentEpoch` error.

jonathanknowles:
# Issue Number

#1939 

# Overview

This PR revises the error we show if we cannot determine the current epoch, inviting the user to wait and try again.

# Comments

Ideally, we'd be able to give the user some indication of _how long_ they should wait, but assuming we cannot reliably know this information, letting them know that waiting may resolve the issue is arguably better than giving no advice at all.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -1997,8 +1997,10 @@ showT = T.pack . show
 instance LiftHandler ErrCurrentEpoch where
     handler = \case
         ErrUnableToDetermineCurrentEpoch ->
-            apiError err500 UnableToDetermineCurrentEpoch
-                "I'm unable to determine the current epoch."
+            apiError err500 UnableToDetermineCurrentEpoch $ mconcat
+                [ "I'm unable to determine the current epoch. "
+                , "Please wait a while for the node to sync and try again."
+                ]
 
 instance LiftHandler ErrUnexpectedPoolIdPlaceholder where
     handler = \case


lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1997,8 +1997,10 @@ showT = T.pack . show
 instance LiftHandler ErrCurrentEpoch where
     handler = \case
         ErrUnableToDetermineCurrentEpoch ->
-            apiError err500 UnableToDetermineCurrentEpoch
-                "I'm unable to determine the current epoch."
+            apiError err500 UnableToDetermineCurrentEpoch $ mconcat
+                [ "I'm unable to determine the current epoch. "
+                , "Please wait a while for the node to sync and try again."
```
rhyslbw, line 2002: Since it's unexpected for a node client to not know which epoch they are in at any time, it might be helpful to add a little detail as to why this cannot be known. If it's too complex to provide over an error message, maybe a link to some docs would be something to add later when available. I assume this is a common issue for chain-sync clients.

rhyslbw (approved):

lib/core/src/Cardano/Wallet/Api/Server.hs
jonathanknowles: > If it's too complex to provide over an error message, maybe a link to some docs would be something to add later when available.

This sounds like a sensible suggestion. :+1: 

From conversations with @Anviking, it seems that this situation will only occur very rarely, around the time of a hard fork, and when a node is not yet fully in sync. I think if we do start seeing reports of this error in the wild, we can look into adding more information to the error.

PR #1995, opened 2020-08-03
revised SQL query returning next pool metadata to fetch

KtorZ:
# Issue Number

# Overview

- fccb250ff7628ad0d2a30f2aa36a4f35f542be1b
  :round_pushpin: **revised SQL query returning next pool metadata to fetch**
    There are two fundamentals change here:

  a) This now does a LEFT JOIN on the fetch attempts table on BOTH the metadata_hash and
  metadata_url. Before, we would discard metadata based solely on their metadata_hash. Now,
  we select on the joined table directly and keep all pool metadata with either no fetch attempts,
  or a fetch_after that is prior to the current datetime.

  b) It now _sorts_ the result based on their fetch_after date, the most "urgent" first. This is to cope
  with the arbitrary batch limit of 100 pools and make sure that all metadata are eventually fetched.
  Before, we could potentially retry fetching the same metadata over and over again until they expire.

- 5c22399b59f55b55409a44b8269fe5359296a2ed
  :round_pushpin: **use 'Either InvalidUrlException' as a base monad for requestFromURI**
    Again... errors from 'requestFromURI' are thrown via 'MonadThrow', which, in the case of 'ExceptT' throws in the base monad, which is IO in this case.

# Comments

On a database synced against latest mainnet, the old query would return the same 100 pools over and over. Whereas, there's in total `524` pools that still need fetching! 

With the query, we indeed get 100 pools that haven't been fetched yet, and all pools are eventually getting fetched (if any of these metadata fetch attempt fail, it is recorded and put later in the queue, so that new pools are tried out until eventually only remains those that are failing). Using the database above, I've re-started the wallet with this patch on mainnet, and after a few seconds, the number of pools that remain to fetch are indeed fetched. In the end, 702 metadata are fetched, and only 274 remains unfetched due to errors.

### DIFF
--- a/lib/core/src/Cardano/Pool/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Pool/DB/Sqlite.hs
@@ -268,22 +268,32 @@ newDBLayer trace fp timeInterpreter = do
             let fetchAttempts = tableName (DBField PoolFetchAttemptsMetadataHash)
             let metadata      = tableName (DBField PoolMetadataHash)
             let query = T.unwords
-                    [ "SELECT"
-                    , metadataUrl, ",", metadataHash
-                    , "FROM", registrations
+                    [ "SELECT", "a." <> metadataUrl, ",", "a." <> metadataHash
+                    , "FROM", registrations, "AS a"
+                    , "LEFT JOIN", fetchAttempts, "AS b"
+                    , "ON"
+                        , "a." <> metadataUrl,  "=", "b." <> metadataUrl, "AND"
+                        , "a." <> metadataHash, "=", "b." <> metadataHash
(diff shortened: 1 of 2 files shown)


paweljakubas (approved):
looks valid. I bet we want to have queryable pool metadata that has ` past retryAfter` OR  `not retried at all` AND order by those starting from latest

KtorZ:
@paweljakubas I am currently trying this out on my mainnet wallet, which has `524` unfetched pool metadata in it :s, including a lot of very recent pool metadata that are _valid_ but are not being fetched because of other older metadata retrials still ongoing. I'll post an update in a bit.

KtorZ:
bors try

Issue #949, opened 2019-11-04
Let CLI client connect to remote hosts


rvl:
# Context

The wallet server can listen on interfaces other than localhost (#833).

The wallet CLI client ought to be able to connect to addresses other than localhost. Relevant to these situations:

- Connecting to a wallet server on a remote host.
- Connecting to the wallet server within a docker container.

# Decision

Change the CLI `--port=8090` argument to something like `--server=localhost:8090`.
Use the given hostname when forming the API client base URL.

# Acceptance Criteria

- [ ] It's possible to use the CLI client to connect to a wallet server listening on a remote host.
- [ ] The default host and port must remain localhost:8090.

---

# Development Plan

# PR

| Number   | Base      |
| ---      | ---       |
| #?       | `master`  |

# QA

KtorZ:
Moved to `cardano-wallet-laboratory` -> https://github.com/input-output-hk/cardano-wallet-laboratory/issues/14

runeksvendsen:
I would really appreciate this feature. Would you accept a PR? The https://github.com/input-output-hk/cardano-wallet-laboratory/issues/14 gives me a 404.

PR #1999, opened 2020-08-04
docker: add nsswitch.conf to base layer

rvl:
### Overview

Configures the docker image to use `/etc/hosts` ahead of DNS.

This works around the issue of the CLI being hardcoded to use localhost.

### Comments

Before:

    docker run --network host --rm -ti --entrypoint=bash inputoutput/cardano-wallet:2020.7.28-shelley
    bash-4.4# ping localhost
    PING localhost(localhost.lan (::1)) 56 data bytes
    64 bytes from localhost.lan (::1): icmp_seq=1 ttl=64 time=0.043 ms
    64 bytes from localhost.lan (::1): icmp_seq=2 ttl=64 time=0.087 ms
    64 bytes from localhost.lan (::1): icmp_seq=3 ttl=64 time=0.092 ms
    ^C
    --- localhost ping statistics ---
    3 packets transmitted, 3 received, 0% packet loss, time 2043ms
    rtt min/avg/max/mdev = 0.043/0.074/0.092/0.022 ms
    bash-4.4# cat /etc/nsswitch.conf
    cat: /etc/nsswitch.conf: No such file or directory

After:

    docker run --network host --rm -ti --entrypoint=bash inputoutput/cardano-wallet:2020.7.28-shelley
    bash-4.4# ping localhost
    PING localhost (127.0.0.1) 56(84) bytes of data.
    64 bytes from localhost (127.0.0.1): icmp_seq=1 ttl=64 time=0.047 ms
    64 bytes from localhost (127.0.0.1): icmp_seq=2 ttl=64 time=0.081 ms
    64 bytes from localhost (127.0.0.1): icmp_seq=3 ttl=64 time=0.078 ms
    ^C
    --- localhost ping statistics ---
    3 packets transmitted, 3 received, 0% packet loss, time 2062ms
    rtt min/avg/max/mdev = 0.047/0.068/0.081/0.015 ms
    bash-4.4# cat /etc/nsswitch.conf
    hosts: files dns

### DIFF
--- a/nix/docker.nix
+++ b/nix/docker.nix
@@ -8,7 +8,7 @@
 #
 ############################################################################
 
-{ runtimeShell, writeScriptBin, runCommand, dockerTools
+{ runtimeShell, writeScriptBin, writeTextFile, runCommand, dockerTools
 
 # The main contents of the image.
 , exe
@@ -40,11 +40,18 @@ let
     exec ${exe}/bin/cardano-wallet-${backend} "$@"
   '';
 
+  # Config file needed for container/host resolution.
+  nsswitch-conf = writeTextFile {
+    name = "nsswitch.conf";
+    text = "hosts: files dns";
+    destination = "/etc/nsswitch.conf";
+  };
+
   # Layer of tools which aren't going to change much between versions.
   baseImage = dockerTools.buildImage {
     name = "${repoName}-env";
     contents = [
-      glibcLocales iana-etc cacert
+      glibcLocales iana-etc cacert nsswitch-conf
       bashInteractive coreutils utillinux iproute iputils curl socat
     ];
     # set up /tmp (override with TMPDIR variable)


KtorZ (approved):
LGTM. What about also changing the BaseUrl constructed in the CLI to not use localhost?

rvl:
The issue for that is (was) #949.

runeksvendsen:
> This works around the issue of the CLI being hardcoded to use localhost.

How do I work around this issue, exactly?

rvl:
Hello @runeksvendsen. There is currently no way to use the cardano-wallet CLI client with a non-localhost server.

If you need to call the API of a cardano-wallet server running on another host, you can:
1. Use `curl` to make your API requests instead of `cardano-wallet`.
2. port forward the remote server to a localhost port, then use `cardano-wallet --port=PORT`

Just make sure that the API server is running with the correct `--listen-address` option.

The changes in this PR achieved something different to what you probably want. This PR makes it at least possible to use the cardano-wallet CLI client inside the same Docker container which the server is running on.

PR #2001, opened 2020-08-04
review guard condition on self withdrawl redemption

KtorZ:
# Issue Number

#1965 

# Overview

- 728ab75ab077c06da4a3811772e05293c3072072
  :round_pushpin: **change integration scenario to always allow redeeming from self**
    This makes it a lot easier for clients to deal with rewards. One can always give 'self' and simply let the backend decides whether or not rewards should be redeemed. For external wallets however, it could be quite counter-intuitive to succeed in the case there are no rewards in the target wallet so we still error in that case.

- 278c9a359a59823b236522ab42b2767d7c0d1047
  :round_pushpin: **move rewards redemption guard to only apply to external redemptions**
    As per previous commit.

- 029da7f0922e09ce5724c63ec8fc4c06e73d14db
  :round_pushpin: **improve and adjust API documentation on withdrawal redemption**

# Comments

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
@@ -1323,7 +1323,7 @@ spec = do
                     (.> (wSelf ^. #balance . #getApiT . #available))
                 ]
 
-    it "SHELLEY_TX_REDEEM_03 - Can't redeem rewards if none left" $ \ctx -> do
+    it "SHELLEY_TX_REDEEM_03 - Can't redeem rewards from other if none left" $ \ctx -> do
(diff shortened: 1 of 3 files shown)


KtorZ:
bors try

paweljakubas (approved):
LGTM. Before this PR, we had the check always ON, for both `ExternalWithdrawal` and `SelfWithdrawal` cases when withdrawal = 0 and this could impede `SelfWithdrawal` txs we don't want to be stopped. Now we will only do it for `ExternalWithdrawal` case, meaning we will not end up with `ErrWithdrawalNotWorth` error in `SelfWithdrawal` case. Which is reasonable change.

KtorZ:
Hydra is failing on the MacOS integration test, failure seems completely unrelated. So, moving on :+1:

Issue #2000, opened 2020-08-04
Swagger docs say payments list >= 0


disassembler:
# Context
 
When you try to create a tx with just withdrawals (and no payments) wallet returns:
```
{"code":"bad_request","message":"Error in $.payments: parsing NonEmpty failed, unexpected empty list"}
```

But swagger docs show `payments` is a list `>=` 0. @KtorZ confirmed the behavior is correct but docs are wrong.

# Steps to Reproduce

1. ...

## Actual behavior
 

---

# Resolution 

---

# QA

KtorZ:
:+1:

PR #2002, opened 2020-08-05
Specify that `payments` should be a non-empty list.

jonathanknowles:
# Issue Number

#2000 

# Overview

This PR:

- [x] adjusts the Swagger specification to indicate that `payments` (for `ApiPostTransactionData` and other related types) should be a non-empty list.
- [x] adjusts the unit tests to only generate `ApiTransaction` values with non-empty lists of outputs.
- [x] adds a reminder to investigate whether or not the `outputs` field of the `ApiTransaction` type should also be a non-empty list. (See https://jira.iohk.io/browse/ADP-400)

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -549,6 +549,9 @@ data ApiTransaction (n :: NetworkDiscriminant) = ApiTransaction
     , depth :: !(Maybe (Quantity "block" Natural))
     , direction :: !(ApiT Direction)
     , inputs :: ![ApiTxInput n]
+      -- TODO: Investigate whether the list of outputs should be non-empty, and
+      -- if so, whether the 'outputs' field can be encoded as a non-empty list.
+      -- See: https://jira.iohk.io/browse/ADP-400
     , outputs :: ![AddressAmount (ApiT Address, Proxy n)]
     , withdrawals :: ![ApiWithdrawal n]
     , status :: !(ApiT TxStatus)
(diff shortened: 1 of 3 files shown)


lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -549,6 +549,9 @@ data ApiTransaction (n :: NetworkDiscriminant) = ApiTransaction
     , depth :: !(Maybe (Quantity "block" Natural))
     , direction :: !(ApiT Direction)
     , inputs :: ![ApiTxInput n]
+      -- TODO: Investigate whether the list of outputs should be non-empty, and
+      -- if so, whether the 'outputs' field can be encoded as a non-empty list.
+      -- See: https://jira.iohk.io/browse/ADP-400
```
rvl, line 554: I don't think using NonEmpty list buys us anything here. The ApiTransaction is derived from the core type which is already valid.
Best to either address the TODO right now or remove it.

rvl (approved):

lib/core/src/Cardano/Wallet/Api/Types.hs
jonathanknowles: > I don't think using NonEmpty list buys us anything here.

If this list _can_ be empty, then I think we should at least have an comment here explaining _why_ the list can be empty, especially when _all_ the other API types involving payments use `NonEmpty`:

- `ApiSelectCoinsData`
- `ApiCoinSelection`
- `PostTransactionData`
- `PostTransactionFeeData`

OTOH, if it really _can't_ be empty, then using `NonEmpty` would give the reader strong evidence that this list cannot be empty. Without this, readers of the code are left wondering whether or not the list really can be empty, which seems somewhat unsatisfactory (given the inconsistency with the rest of the API).

I did actually have a go at changing this to use a `NonEmpty` list, but it became clear that this was a non-trivial change. Hence, I added this comment with a reminder to tackle this in future.

jonathanknowles:
bors retry

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: I am planning to make that thing non-empty in a very near future actually, but under special conditions that I intend to specify indeed :smile: (when creating a withdrawal for instance).

Issue #2009, opened 2020-08-07
Failure after quitting a stake pool from API


disassembler:
# Context
 
I quit a stake pool with intention of getting my deposit back and now my wallet throws this error in the logs:

```
[cardano-wallet.wallet-engine:Error:43] [2020-08-07 07:32:45.73 UTC] 237779f4: Unexpected error following the chain: arithmetic underflow
```

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      |  |
| Platform     |           |
| Installation |             |

# Steps to Reproduce

1. ...

## Actual behavior
 

---

# Resolution 

- Use a safe "distance" when computing amounts of transactions #2010 

---

# QA 

- https://github.com/input-output-hk/cardano-wallet/blob/d5b43356e4dbe9267ede9c8e912da6c893bdda85/lib/core/src/Cardano/Wallet.hs#L1735-L1737

- https://github.com/input-output-hk/cardano-wallet/blob/d5b43356e4dbe9267ede9c8e912da6c893bdda85/lib/core/src/Cardano/Wallet/Primitive/Model.hs#L409-L413

Fell-x27:
Got the same error today,:

`[cardano-wallet.wallet-engine:Error:1114] [2020-08-07 18:20:38.15 UTC] 4ec9262c: Unexpected error following the chain: arithmetic underflow
`

The wallet can't sync and spamming node with:

`[developM:cardano.node.LocalErrorPolicy:Error:62] [2020-08-07 18:20:41.11 UTC] [String "ErrorPolicyUnhandledApplicationException (MuxError MuxBearerClosed \"<socket: 82> closed when reading data, waiting on next header True\")",String "ErrorPolicyTrace",String "LocalAddress {getFilePath = \"\"}"]
`

So node even gets desync and can't work normally.

Trying understand why this wallet (4ec9262c) is causing it.

Just one of ~100 wallets. Others work fine.

Fell-x27:
>Trying understand why this wallet (4ec9262c) is causing it.

Fortunately I discovered that it's a person I know :)
The wallet's owner is an enthusiast, who tried to _undelegate_ his stake today :)


The full log section:

`[cardano-wallet.wallet-engine:Error:32] [2020-08-07 19:38:16.50 UTC] 4ec9262c: Unexpected error following the chain: arithmetic underflow
[...]

PR #2010, opened 2020-08-07
Yet another fix for reported withdrawals in transactions + underflow in case of key deposit reclaim

KtorZ:
# Issue Number

#2009 + TODO

# Overview

- 5757a5500a143e70cd75e2212dddc59b93fd039a
  :round_pushpin: **highlight reported failure in integration scenarios by adding more assertions**
  
- d5b43356e4dbe9267ede9c8e912da6c893bdda85
  :round_pushpin: **report our withdrawals differently from external withdrawals.**
    We haven't been quite careful here when introducing the reward redemption and the transaction amount are looking weird again.
  This commit fixes several issues:

  1. It only counts withdrawals on the "spent" side of the balance if they are coming from OUR reward account. Indeed, in the case of external withdrawals, the money is coming from elsewhere and not from the wallet itself.

  2. Fix an underflow in the amount calculation in the case where we spent less than we receive. This can be the case when:
    a. We are redeeming from an external account and the reward brings more than the fee. From the redeeming wallet, it'll look like the wallet is receiving money.
    b. We are reclaiming a key deposit back, and it brings more money than what's actually spent.

  3. Discover transactions that are spending our withdrawals without belonging to our wallet. This happens when the reward is redeemed from another wallet. That transaction should still show up in the redeemed wallet, without which rewards would just "vanish" without any trace reported by the wallet.

# Comments

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
@@ -1301,13 +1301,20 @@ spec = do
                 }],
                 "passphrase": #{fixturePassphrase}
             }|]
+        let fee = 143600
+
         rTx <- request @(ApiTransaction n) ctx
             (Link.createTransaction @'Shelley wSelf) Default payload
         verify rTx
             [ expectResponseCode HTTP.status202
             , expectField #withdrawals
                 (`shouldSatisfy` (not . null))
+            , expectField (#direction . #getApiT)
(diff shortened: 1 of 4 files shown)


paweljakubas (approved):
I thought we have already `ourWithdrawal` in model ... and what is left to be accounted are inputs when multisig txs are added. Using `distance` seems fine to avoid underflow error. So I bet in this or another PR the `ourWithdrawal` logic is going to be only in Model, right?

Issue #2006, opened 2020-08-06
 Fees calculation with not null withdrawal param takes 3-10mins 


Fell-x27:
# Context
 
Splitted from https://github.com/input-output-hk/cardano-wallet/issues/2004

When I'm trying to calculate fees with "withdrawal"!="null", it takes up to 10 minutes. 5 as average.

It doesn't depend on wallet size\UTxO amount\Reward amount. The same result for funded wallet without reward, rewarded without UTxO, and even for zero-balanced one.

Probably, this problem is related to the count of wallets because there are dozens in my case. 

The node and wallet are **full synced**.

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | v2020-08-03 |
| Platform     | Linux          |
| Installation | From binaries           |

# Steps to Reproduce

1. Run wallet over full-synced node.
2. Try to calculate fees with withdrawal included
3. ...
4. PROFIT!1

## Expected behavior

Fees are calculated for a couple seconds or faster.

## Actual behavior
 
It takes much much more time. Up to 10 mins. For the casual request which can't be prefetched and cached by application :(

---

# Resolution 

Seems likely to be the same cause as #2005.

---

# QA 

Pretty sure this is exactly the same as #2005, but tested once manually with the 1.18.0 node.

```bash
echo '{"payments": [{"address": "addr1qyx5s6kth3ta38qv782yajnacap97rt6jfugck36ku5n6za3veh8xatdfdra9kx57gray20m28t4mhu9l5pqaeemasdsaw4l0s", "amount": {"unit": "lovelace", "quantity": 0}}], "withdrawal": "self"}' | http post localhost:8090/v2/wallets/9d1e1558177c61d19e93aa25a9cbee774fd08e46/payment-fees
```
And indeed:
- 2020.8.3 (git revision: 40a8b5daa1da77480e4b818c49f2a4977aed4a71) -> 1min31s
- 2020.8.3 (git revision: fece720cb18d9676c4f1f6b0d3a69a37fab93f1c) -> 700ms

KtorZ:
May you provide the result of the `network information` command please ?

Fell-x27:
```
{                                                                                                                          
    "network_tip": {                                                                                                       
        "epoch_number": 209,                                                                                               
        "slot_number": 210623                                                                                              
    },                                                                                                                     
    "node_tip": {                                                                                                          
        "height": {                                                                                                        
            "quantity": 4522589,                                                                                           
[...]

PR #2014, opened 2020-08-10
Run coin selection properties with non-zero withdrawal

Anviking:
# Issue Number

From looking at  #2006, but surely not related.

# Overview

- [x] I added a withdrawal field to `CoinSelProp` and used it in coin selection properties

# Comments

- I'm not that familiar with the coin selection, but this seems like a good idea, right?

### DIFF
--- a/lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelectionSpec.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelectionSpec.hs
@@ -68,6 +68,7 @@ import Test.QuickCheck
     , counterexample
     , cover
     , elements
+    , frequency
     , generate
     , scale
     , vector
@@ -82,6 +83,8 @@ import qualified Data.List.NonEmpty as NE
 import qualified Data.Map.Strict as Map
 import qualified Data.Set as Set
 import qualified Test.QuickCheck.Monadic as QC
+
+{-# ANN module ("HLint: ignore Use <$>" :: String) #-}
 
 spec :: Spec
 spec = do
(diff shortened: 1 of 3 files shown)


lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelectionSpec.hs
```diff
@@ -339,9 +346,17 @@ instance Arbitrary a => Arbitrary (NonEmpty a) where
         NE.fromList <$> vector n
 
 instance Arbitrary CoinSelProp where
-    shrink (CoinSelProp utxo outs) = uncurry CoinSelProp
-        <$> zip (shrink utxo) (shrink outs)
-    arbitrary = applyArbitrary2 CoinSelProp
+    shrink (CoinSelProp utxo wdrl outs) =
+        tail $ CoinSelProp <$> shrinkUTxO <*> shrinkWithdrawal <*> shrinkOuts
```
Anviking: Think list comprehensions are more idiomatic

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelectionSpec.hs
```diff
@@ -339,9 +345,17 @@ instance Arbitrary a => Arbitrary (NonEmpty a) where
...
-    arbitrary = applyArbitrary2 CoinSelProp
+    shrink (CoinSelProp utxo wdrl outs) =
+        [ CoinSelProp utxo' wdrl outs | utxo' <- shrink utxo ]
+        ++ [ CoinSelProp utxo wdrl' outs | wdrl' <- shrinkWdrl wdrl ]
+        ++ [ CoinSelProp utxo wdrl outs' | outs' <- shrink outs ]
+      where
+        shrinkWdrl = map Quantity . shrink . getQuantity
+    arbitrary = do
+        utxo <- arbitrary
+        wdrl <- Quantity <$> frequency [(65, return 0), (35, arbitrary)]
```
Anviking, line 356: 65% no withdrawal like previously

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelectionSpec.hs
```diff
@@ -227,14 +230,17 @@ data CoinSelectionsSetup = CoinSelectionsSetup
 data CoinSelProp = CoinSelProp
     { csUtxO :: UTxO
         -- ^ Available UTxO for the selection
+    , csWithdrawal :: Quantity "lovelace" Word64
+    -- ^ Availible Withdrawal
```
paweljakubas:
```suggestion
    -- ^ Available Withdrawal
```

paweljakubas (approved):
LGTM I wonder if we can add some properties, especially when withdrawal is edge case in comparison with inputs :thinking:

Issue #2016, opened 2020-08-11
Add pool DB operation `listRetiredPools`.


jonathanknowles:
# Context

As part of the work to perform garbage collection of retired pools from the database, we need to be able to determine _which_ pools have retired.

# Decision

Create a pool database operation `listRetiredPools`:

```hs
listRetiredPools
    :: EpochNo
    -> stm [PoolId]
    -- ^ List all pools with an active retirement epoch that is
    -- earlier than or equal to the specified epoch.
````

## SQLite implementation

The SQLite implementation of this operation can be based on the following **view**, which lists all pools that have active retirement certificates associated with them:

<details><summary>Click to expand</summary>

```sql
CREATE VIEW active_pool_retirements AS
SELECT * FROM (
    SELECT
        pool_id,
        retirement_epoch
    FROM (
        SELECT row_number() OVER w AS r, *
        FROM (
            SELECT
                pool_id, slot, slot_internal_index, NULL as retirement_epoch
                FROM pool_registration
            UNION
            SELECT
                pool_id, slot, slot_internal_index, epoch as retirement_epoch
                FROM pool_retirement
        )
        WINDOW w AS (ORDER BY pool_id, slot desc, slot_internal_index desc)
    )
    GROUP BY pool_id
)
WHERE retirement_epoch IS NOT NULL;
```

</details>

# Acceptance Criteria

- [x] The pool DB _must_ provide a `listRetiredPools` operation.
- [x] The `listRetiredPools` operation _must not_ include pools that have never had retirement certificates associated with them.
- [x] The `listRetiredPools` operation _must not_ include pools that have had their most-recently-published retirement certificates revoked by subsequent re-registration certificates.
- [x] The `listRetiredPools` operation _must not_ include pools with retirement epochs that are later than the specified epoch.

# Development 

# QA 

Create property tests for the **model** and **SQLite** implementations of `listRetiredPools` to:
- [x] check that they return **the same results** given the same inputs and initial conditions.
- [x] check that they **do not** include pools that have never had retirement certificates associated with them.
- [x] check that they **do not** include pools that have had their most-recently-published retirement certificates revoked by subsequent re-registration certificates.
- [x] check that they **do not** include pools with active retirement epochs that are later than the specified epoch.

piotr-iohk:
LGTM.

@jonathanknowles this introduces new DB view `active_pool_retirements`. Do we need to consider any additional migration testing?

jonathanknowles:
[...]

PR #2024, opened 2020-08-11
Add pool DB operation `listRetiredPools`.

jonathanknowles:
# Issue Number

#2016 

# Overview

This PR:

- [x] Adds a `listRetiredPools` operation to the pool DB layer:<br><br>
    ```hs
    listRetiredPools
        :: EpochNo
        -> stm [PoolId]
        -- ^ List all pools with an active retirement epoch that is
        -- earlier than or equal to the specified epoch.
    ````
- [x] Provides the following implementations of `listRetiredPools`.
    - [x] `Pool.DB.MVar`
    - [x] `Pool.DB.SQLite`.

The SQLite implementation is based on an **SQL view** ([see definition](https://github.com/input-output-hk/cardano-wallet/pull/2024/files#diff-08afa2852f50fa31f5757942d004ce8aR513)) which determines the currently-active set of retirement certificates.

<details><summary>Click to view sample data</summary>

active_pool_retirements

</details>

# Property Testing

This PR also adds property tests to check that:
- [x] `listRetiredPools` **does not** return:
    - [x] pools that have never had retirement certificates associated with them.
    - [x] pools that have had their most-recently-published retirement certificates revoked by subsequent re-registration certificates.
    - [x] pools with effective retirement epochs that are later than the specified epoch.
- [x] `listRetiredPools` returns the correct set of pools in the following contexts:
    - [x] where there are **multiple pools**.
    - [x] where there are **multiple certificates per pool**
        (registrations, retirements, re-registrations, re-retirements).
    - [x] where the publications of certificates affecting different pools are **interleaved together in time**.

# Mutation Testing

To increase confidence that the `activePoolRetirements` view is correctly defined, I have manually performed the following mutations and verified that the `prop_listRetiredPools_multiplePools_multipleCerts` property fails appropriately.

### Mutation 1

```patch
-    WINDOW w AS (ORDER BY pool_id, slot desc, slot_internal_index desc)
+    WINDOW w AS (ORDER BY pool_id, slot     , slot_internal_index desc)
 )
 GROUP BY pool_id
```

### Mutation 2

```patch
-    WINDOW w AS (ORDER BY pool_id, slot desc, slot_internal_index desc)
+    WINDOW w AS (ORDER BY pool_id, slot desc, slot_internal_index     )
 )
 GROUP BY pool_id
```

### Mutation 3

```patch
-    WINDOW w AS (ORDER BY pool_id, slot desc, slot_internal_index desc)
+    WINDOW w AS (ORDER BY pool_id, slot     , slot_internal_index     )
 )
 GROUP BY pool_id
```

### Mutation 4

```patch
-    WINDOW w AS (ORDER BY pool_id, slot desc, slot_internal_index desc)
+    WINDOW w AS (ORDER BY pool_id, slot desc,                         )
 )
 GROUP BY pool_id
```

### DIFF
--- a/lib/core/src/Cardano/Pool/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Pool/DB/Sqlite.hs
@@ -122,6 +122,7 @@ import Cardano.Pool.DB.Sqlite.TH
 
 import qualified Data.Map.Strict as Map
 import qualified Data.Text as T
+import qualified Data.Text.Class as T
 import qualified Database.Sqlite as Sqlite
 
 -- | Return the preferred @FilePath@ for the stake pool .sqlite file, given a
@@ -354,7 +355,19 @@ newDBLayer trace fp timeInterpreter = do
                 , Desc PoolRegistrationSlotInternalIndex
                 ]
 
-        , listRetiredPools = \_epochNo -> pure []
+        , listRetiredPools = \epochNo -> do
+            let query = T.unwords
+                    [ "SELECT * FROM "
+                    , databaseViewName activePoolRetirements
+                    , "WHERE retirement_epoch <="
+                    , T.toText epochNo
+                    , ";"
+                    ]
+            let safeCast (Single poolId, Single retirementEpoch) =
+                    PoolRetirementCertificate
+                        <$> fromPersistValue poolId
+                        <*> fromPersistValue retirementEpoch
+            rights . fmap safeCast <$> rawSql query []
 
         , rollbackTo = \point -> do
             -- TODO(ADP-356): What if the conversion blocks or fails?
--- a/lib/core/src/Cardano/Pool/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Pool/DB/Sqlite.hs
@@ -354,6 +354,8 @@ newDBLayer trace fp timeInterpreter = do
                 , Desc PoolRegistrationSlotInternalIndex
                 ]
 
+        , listRetiredPools = \_epochNo -> pure []
+
         , rollbackTo = \point -> do
             -- TODO(ADP-356): What if the conversion blocks or fails?
             --
--- a/lib/core/src/Cardano/Pool/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Pool/DB/Sqlite.hs
@@ -5,6 +5,7 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedLabels #-}
+{-# LANGUAGE QuasiQuotes #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TupleSections #-}
@@ -25,6 +26,7 @@ module Cardano.Pool.DB.Sqlite
     ( newDBLayer
     , withDBLayer
     , defaultFilePath
+    , DatabaseView (..)
     ) where
 
 import Prelude
@@ -81,6 +83,10 @@ import Data.Quantity
     ( Percentage (..), Quantity (..) )
 import Data.Ratio
     ( denominator, numerator, (%) )
+import Data.String.QQ
+    ( s )
+import Data.Text
+    ( Text )
 import Data.Time.Clock
     ( UTCTime, addUTCTime, getCurrentTime )
 import Data.Word
@@ -116,6 +122,7 @@ import Cardano.Pool.DB.Sqlite.TH
 
 import qualified Data.Map.Strict as Map
 import qualified Data.Text as T
+import qualified Database.Sqlite as Sqlite
 
 -- | Return the preferred @FilePath@ for the stake pool .sqlite file, given a
 -- parent directory.
@@ -166,7 +173,7 @@ newDBLayer
     -> IO (SqliteContext, DBLayer IO)
 newDBLayer trace fp timeInterpreter = do
     let io = startSqliteBackend
-            (ManualMigration mempty)
+            (migrateManually trace)
             migrateAll
             trace
             fp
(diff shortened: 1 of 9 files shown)


lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -347,6 +355,20 @@ newDBLayer trace fp timeInterpreter = do
...
+                    [ "SELECT * FROM "
+                    , databaseViewName effectivePoolRetirements
+                    , "WHERE retirement_epoch <="
+                    , T.toText epochNo
+                    , ";"
+                    ]
+            let safeCast (Single poolId, Single retirementEpoch) =
+                    PoolRetirementCertificate
+                        <$> fromPersistValue poolId
+                        <*> fromPersistValue retirementEpoch
```
KtorZ, line 369: Hmmm. Why is the unsafe query needed here? We should stick with the persistent DSL as much as possible and only resort to unsafe queries when the DSL is too limiting or becomes a real performance impediment.

jonathanknowles: @KtorZ wrote:
> Why is the unsafe query needed here?

When determining the current set of effective retirements, we need to combine results from **two different tables**:

1. `pool_registrations`
2. `pool_retirements`

This is because registration and retirement certificates can _cancel each other out_.

The only way to find out whether a retirement certificate is still in effect is to verify that there are no other retirement certificates _or_ registration certificates that cancel it out.

> we should only resort to unsafe queries when the DSL is too limiting

I do agree with this.

However, the implementation in this PR determines the set of currently-active retirement certificates with a **single query** that uses:
- the SQL `UNION` function;
- the SQL `row_number` window function.

As far as I know, `persistent` supports neither of the above functions (at least out of the box).

Furthermore, defining this query in terms of an SQL **view** can help us if we need to debug a real pool database. By opening the database and fetching the `active_pool_retirements` view, we can immediately see which retirement certificates the wallet currently considers to be active.

jonathanknowles: Here's the `active_pool_retirements` view for mainnet:

active_pool_retirements

To implement `listRetiredPools`, we just need to select all rows from this view with a `retirement_epoch` that is _earlier than or equal_ to the given epoch argument.

jonathanknowles:
bors try

jonathanknowles:
bors try

jonathanknowles:
bors retry

jonathanknowles:
bors try

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
```diff
@@ -465,13 +477,20 @@ prop_poolRegistration DBLayer {..} entries =
...
             , "Read from DB: "
             , show entriesOut
+            , "All pools that are marked to retire: "
+            , unlines (("\n" <>) . show <$> poolsMarkedToRetire)
             ]
-        assert (entriesIn == entriesOut)
+        assertWith "entriesIn == entriesOut"
+            $ entriesIn == entriesOut
+        assertWith "no pools are marked to retire"
+            $ null poolsMarkedToRetire
```
KtorZ, line 493: :+1:

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
```diff
@@ -488,13 +507,22 @@ prop_poolRetirement DBLayer {..} entries =
...
+            , "All pools that are marked to retire: "
+            , unlines (("\n" <>) . show <$> poolsMarkedToRetire)
             ]
-        assert (entriesIn == entriesOut)
+        assertWith "entriesIn == entriesOut"
+            $ entriesIn == entriesOut
+        assertWith "all pools are marked to retire"
+            $ (==)
+                (Set.fromList $ snd <$> entriesIn)
+                (Set.fromList poolsMarkedToRetire)
```
KtorZ, line 525: :+1:

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
```diff
@@ -526,9 +556,11 @@ prop_multiple_putPoolRegistration_single_readPoolRegistration
             , "\nAll certificate publications: "
             , unlines (("\n" <>) . show <$> certificatePublications)
             ]
-        assert $ (==)
+        assertWith "retrieved certificate matches expectations" $ (==)
             mRetrievedCertificatePublication
             mExpectedCertificatePublication
+        assertWith "pool is not marked to retire" $
+            null poolsMarkedToRetire
```
KtorZ, line 563: :+1:

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
```diff
@@ -822,6 +871,91 @@ prop_listRegisteredPools DBLayer {..} entries =
...
+  where
+    setup = run $ atomically cleanDB
+
+    prop = do
+        run $ atomically $ do
+            mapM_ (uncurry putCertificate) allPublicationsSerialized
+        lifeCycleStatuses <- run $ atomically $ do
+            mapM readPoolLifeCycleStatus allPoolIds
+        let poolsMarkedToRetire = catMaybes $
+                getPoolRetirementCertificate <$> lifeCycleStatuses
```
KtorZ, line 916: I'd be better not to rely on `lifeCycleStatuses` to determine which pools are supposed to be retired. The risk is to have a similar logic bug in both `lifeCycleStatuses` and `listRetiredPools` that remain uncaught, especially if the same approach is used behind the scene.

KtorZ (approved):

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
jonathanknowles: > I'd be better not to rely on `lifeCycleStatuses` to determine which pools are supposed to be retired. The risk is to have a similar logic bug in both `lifeCycleStatuses` and `listRetiredPools` that remain uncaught, especially if the same approach is used behind the scene.

I definitely agree, this is indeed a risk.

To counteract this risk:

1. We do have a suite of tests specifically aimed at `determinePoolLifeCycleStatus`:
    - `prop_determinePoolLifeCycleStatus_orderCorrect`
    - `prop_determinePoolLifeCycleStatus_neverRegistered`
    - `prop_determinePoolLifeCycleStatus_differentPools`<br><br>

2. The definition of `listRetiredPools` uses a completely different approach in the form of a handwritten query. In the property test for this query, we check it against the `determinePoolLifeCycleStatus` function, which we consider to be the source of truth.

jonathanknowles:
bors ry

jonathanknowles:
bors retry

PR #2027, opened 2020-08-11
Allow to build with cabal

hasufell:
The process to create these files is roughly:

1. run `stack2cabal`
2. work around cabal bugs (see https://github.com/haskell/cabal/issues/5444)
3. disable tests to avoid build failures with --enable-tests from
   unrelated packages (cabal considers all packages defined via
   source-repository-package as *local* and will try to build its tests)
4. fix package constraints in cabal.project.freeze, where stackage
   constraints were overwritten in stack.yaml files (usually raising
   a version number)

Rationale:

1. good to support all available tools as options for interested users
2. cabal allows easier managing of package bounds (and testing against
   latest package versions)
3. #1303 is an excellent workaround for nix users, but requires nix.
   This should be the go-to fallback in case cabal project files
   get out of sync for too long.


rvl:
Thanks for the PR, but last time someone did this the cabal build broke pretty much the next day.
And this will do the same.
The process to keep the `cabal.project` file updated is manual and undocumented, with debugging steps required, and no CI checks.

But even if the `cabal.project` file were fastidiously maintained, the package set which the cabal solver chooses will not be the same as the LTS snapshot from the stack build. So cabal users will have different package versions and therefore different build/test results than stack users.

It is not hard to install stack to build the project.

It is also not that hard to install nix. And by building with cabal inside a nix-shell you get the correct versions of Haskell dependencies prebuilt from cache, as well as all required system libraries (e.g. systemd, libsodium), and all the build tools, including cabal itself.

hasufell:
>  Thanks for the PR, but last time someone did this the cabal build broke pretty much the next day. And this will do the same.

I'm using cabal as my primary tool, so I'd like to maintain these files. But I agree that CI should run cabal as well. Once I'm more familiar with those pieces, I'll be happy to provide a PR for that part as well.

> The process to keep the cabal.project file updated is manual and undocumented, with debugging steps required, and no CI checks.

That is unfortunate, but I believe that is true for nix as well (which also disables systemd for example). Synchronizing build systems always has a maintenance cost. I believe it not to be high though.

> The package set which the cabal solver chooses will not be the same as the LTS snapshot from the stack build. So cabal users will have different package versions and therefore different build/test results than stack users.

The cabal.freeze file is generated from the LTS snapshot, so the solver will choose the same, unless cabal.freeze file is modified afterwards.

> It is not hard to install stack to build the project.

I think choice is good and it is consistent with what cardano-node does, especially since this isn't just about developers, but also users trying to build the project.

cabal.project
```diff
@@ -1,2 +1,438 @@
...
+
+package ouroboros-network-framework
+  tests: False
+
+package ouroboros-network
+  tests: False
+
+-- force systemd off until there's a solution to
+-- https://github.com/haskell/cabal/issues/5444
+flags: -systemd
```
hasufell: This is being debated in https://github.com/input-output-hk/cardano-node/pull/1666 as well

hasufell:
ping

gen-cabal-nosystemd.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+# force systemd off until there's a solution to
+# https://github.com/haskell/cabal/issues/5444
+
+set -eux
```
KtorZ: let's move this one under "scripts" :+1:

hasufell:
Instructions on how to update the cabal files is in the commit:

> 1. Run stack2cabal
> 2. Disable tests and benchmarks of local packages
>    (those specified via source-repository-package...
>    this won't be needed with cabal-3.4 anymore, because it won't treat
>    them as local anymore)
> 3. Hackage packages in `packages` in `cardano-1.19.x.yaml` or `extra-deps`
>    in `stack.yaml`, which are specified as git repositories might
>    have different versions than the stack resolver. `stack2cabal`
>    cannot pick these up, so these must be adjusted manually (or deleted) in
>    `cabal.project.freeze` (at the time of writing: http-client and persistent).

hasufell:
Another good argument for this is that it unbreaks cabal on windows.

rvl:
> Another good argument for this is that it unbreaks cabal on windows.

Have you actually tried building this branch with cabal on windows?

hasufell:
> Have you actually tried building this branch with cabal on windows?

@Anviking did I think?

hasufell:
> Have you actually tried building this branch with cabal on windows?

I just did and it's a major step forward for windows support.

I described how to build cardano-wallet with cabal on windows here: https://gist.github.com/hasufell/0deaeb316eb8a3ad8ad9bf1266b332f8

This also has the advantage over stack that you have a properly up2date msys2 installation and it allows easy adding of native dependencies via `mingw-pkg`.

hasufell:
Anything blocking this? There's been no further objections and the process to update the cabal files is [documented](https://github.com/input-output-hk/cardano-wallet/wiki/Building#syncing-cabal-with-stack).

KtorZ:
@hasufell may you modify the wiki as we talked about? i.e. illustrate the steps with examples and give more details about vague words like "adjust" or "disable". Consider this from the point of view of someone who is not familiar with cabal.

hasufell:
> @hasufell may you modify the wiki as we talked about? i.e. illustrate the steps with examples and give more details about vague words like "adjust" or "disable". Consider this from the point of view of someone who is not familiar with cabal.

I'm not sure what is missing. All the steps necessary are explained. If you would give it a shot and let me know what specifically didn't work or wasn't clear, that would be helpful.

I'd suggest to use ghcup to install `cabal-3.4.0.0-rc3`, so the second step can be skipped.

hasufell:
Rebased and wiki updated.

hasufell:
bors try

rvl (approved):

Issue #2005, opened 2020-08-06
The pools list fetching takes 3-10 mins 


Fell-x27:
# Context
 
Splitted from https://github.com/input-output-hk/cardano-wallet/issues/2004

When I try to fetch pools list, it takes too much time. The same problem appeared within ITN either and it could be solved by shadow-caching at the application level. But I'm not sure if that behaviour is fine. 

I noticed almost linear progression for time spending depends on pools count. So, this problem not appears in the testnet with low pools count. 

Probably, this problem is related to the count of wallets because there are dozens in my case. 

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | v2020-08-03 |
| Platform     | Linux          |
| Installation | From binaries           |

# Steps to Reproduce

1. Run the wallet over full-synced node.
2. Request pools list
3. ...
4. PROFIT!1

## Expected behavior

Fast response about couple seconds. Or 10-20 seconds may be.

## Actual behavior
 
You have to wait for 3-10 mins. My avg value is about 5 mins.

---

# Resolution 

Cache account balances, and only refetch them on tip-changes (all wallets in a single query) #2034

Related to: input-output-hk/cardano-node#1674

Outstanding TODO: https://github.com/input-output-hk/cardano-wallet/issues/2042 (arguments for #2034 being ok for now is provided there)

---

# QA 

Tested manually:

## Without fix

Using:
1.18.0 node
Master cardano-wallet pre-fix ( e.g. 51221c7f06fa119048b763b0737202c646100456)
13 wallets syncing at ≈25%

Steps:
Repeatedly run `cardano-wallet stake-pool list --stake 100`

Results 
The first couple of times it is fast (~10s). 

After trying a lot, and a couple of minutes it takes 4 minutes. 👎 

## With fix
Using:
Same db and node as without the fix
A wallet that includes the fix (e.g. 90c1e5953a48ffc65f004a68abd545ef68574c4b).

Steps:
Repeatedly run `cardano-wallet stake-pool list --stake 100`

Result
At first: ~10s. 
After waiting 1 min: ~12s 👍

KtorZ:
Testing locally against the current mainnet, listing stake pools takes between 1.2 and 1.5 **seconds** for me. 

[...]

Issue #1997, opened 2020-08-03
Wallets are extremely slow to restore for large wallets (e.g. exchanges).


KtorZ:
# Context
 
 

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | edge  |
| Platform     | All           |
| Installation |             |

# Steps to Reproduce

1. Have a big wallet (50k+ pending addresses and UTxO). 
2. Restore a wallet an observe

## Expected behavior

1. The wallet _just work_. 

## Actual behavior
 

1. Things get really slow.

---

# Resolution 

1. Try adding more logs to measure various actions of the restoration process to better pinpoint the problem ==> #1994 
2. ~Try using sequential wallets that are storing _diffs_ instead of entire checkpoints~. Aborted, it seems that the performance gain is pretty minor for this, even for big wallets. 
3. Add logging to benchmarks to see the influence of tracers on benchmarks ==> #2030
4. Fix usage of `filterSeverity` and apply it one level earlier (one the Tracer, and not the Trace itself, filtering needs to happen _before_ the `LogObject` is constructed) ==> #2028 

---

# QA 

- Added read/write of RndState in the database benchmark to also measure db operation for random wallets.

- Enabled a non-null tracer in benchmarks to reproduce the production setup. Benchmarks went 2 to 4 times slower, applying the fix turned them back to time comparable with a `nullTracer` (see comments on https://github.com/input-output-hk/cardano-wallet/pull/2030).

- Our partners that had difficulty restoring confirmed that they were now able to restore their wallets.

Anviking:
This is from a recent benchmark:
<img width="853" alt="Skärmavbild 2020-08-10 kl  12 40 45" src="https://user-images.githubusercontent.com/304423/89775381-f501a080-db07-11ea-899e-fa015cab7704.png">

The _shape_ of 1% seems very peculiar, and I think it used to be [≈50 min](https://cardano-wallet-benchmarks.herokuapp.com/mainnet-restoration) instead of ≈ 1h35 min

KtorZ:
I'd actually somewhat expect this shape for the last k blocks (since checkpoints are much more frequent in this area). but, the behavior seems to drastically change around block #600.000 o.O ... That makes no sense. 

**edit:** The restoration actually failed around 12%: 

![Screenshot from 2020-08-10 13-05-45](https://user-images.githubusercontent.com/5680256/89776474-46129400-db0a-11ea-8f09-647dca3627f7.png)

PR #2028, opened 2020-08-12
Filter Tracers before adding LogObject and converting to Traces

rvl:
### Issue Number

~Relates to #2005, I think.~

**edit by @KtorZ**: relates to #1997 

### Overview

Changes the tracer minimum severity so that it filters logs before adding `LogObject` and converting to `Trace`.

Users noticed that setting `--trace-db=off` improved performance, especially when restoring many wallets in parallel.

### Comments

Benchmarks not fixed yet.


KtorZ (approved):
Ah! Good. I was about to do it :) 
I've updated the benchmark test suite to use a non-null tracer as well, showing the effect of logging on benchmarks (which should be in theory, negligible).

PR #2031, opened 2020-08-12
Time LSQ requests (and trace the times)

Anviking:
# Issue Number

# Overview

- [x] Measure and trace LSQ queries such that we can correlate the slowness with that of #2005 and #2006

# Comments

E.g.
```bash
[cardano-wallet.network:Info:130] [2020-08-12 11:03:17.03 UTC] Query getAccountBalance took 51.175731s
[cardano-wallet.network:Info:145] [2020-08-12 11:03:22.08 UTC] Query getAccountBalance took 51.244343s
```

### DIFF
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
@@ -109,7 +109,7 @@ import Control.Monad.Class.MonadThrow
 import Control.Monad.Class.MonadTimer
     ( MonadTimer, threadDelay )
 import Control.Monad.IO.Class
-    ( liftIO )
+    ( MonadIO, liftIO )
 import Control.Monad.Trans.Except
     ( ExceptT (..), throwE, withExceptT )
 import Control.Retry
@@ -138,6 +138,8 @@ import Data.Text
     ( Text )
 import Data.Text.Class
     ( ToText (..) )
+import Data.Time.Clock
+    ( NominalDiffTime, diffUTCTime, getCurrentTime )
 import Data.Void
     ( Void )
 import Data.Word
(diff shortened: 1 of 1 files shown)


lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -1051,4 +1083,5 @@ instance HasSeverityAnnotation (NetworkLayerLog b) where
         MsgWatcherUpdate{}                 -> Debug
         MsgChainSyncCmd cmd                -> getSeverityAnnotation cmd
         MsgInterpreter{}                   -> Debug
+        MsgQueryTime{}                     -> Info
```
Anviking, line 1086: Could increase the severity if the time is long.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -777,6 +781,31 @@ debounce action = do
...
+timeQryAndLog
+    :: MonadIO m
+    => String
+    -- ^ Label to identify the query
+    -> Tracer m (NetworkLayerLog sc)
+    -- ^ Tracer to which the measurement will be logged
+    -> m a
+    -- ^ The action that submits the query.
+    -> m a
+timeQryAndLog label tr act = do
```
KtorZ, line 801: What about using the `bracketTracer` from the logging like in: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/DB/Sqlite.hs#L222-L223

The main benefits is that you still get some log even if the action doesn't finish and we don't have to compute time on top of the logger which already comes with timestamps.

rvl:
This is a good idea.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
Anviking: Interesting, but then we'd need some state machine to write `Trace BracketLog -> Trace NominalDiffTime`. That seems _more_ complicated.

I think logging the `NominalDiffTime` is much better than requiring us to look and compare timestamps later.

We could add `MonadCatch` to `timeQryAndLog`, but:
- We don't expect the query `send`s to throw in `IO`
- When it does (I think socket disconnects), do we really care about the time?
- I think `(MonadCatch m, MonadIO m)` wouldn't catch `IO` exceptions if  `m ~ ExceptT`, which could be surprising (we could make `timeQryAndLog` work with a concrete `IO`, but 🤷‍♂️ )

KtorZ (approved):
bors r+

Issue #2032, opened 2020-08-12
'listAddresses' is extremely slow / unusable for large wallets.


KtorZ:
# Context
 
 

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | 2020-08-03  |
| Platform     | All           |
| Installation | Source             |

# Steps to Reproduce

1. Restore a wallet with a large UTxO (> 10K) 
2. List addresses via the corresponding endpoint

## Expected behavior

- Clients receive a response in a reasonable amount of time.

## Actual behavior
 

- The request is timing out even after several minutes.

---

# Resolution 

1. The main issue of this handler is that it loads in memory and crawls the ENTIRE transaction history of the wallet. This is done merely to know whether addresses have been used or not. This is unnecessary for random addresses because used and unused addresses are already separated in two different Maps. For sequential wallets it's a bit more subtle but, we can mark addresses as "Used" during discovery and store this information in the database. See #2033 

2. The server currently returns the entire address space as one big chunk and only replies once all addresses are returned. We could leverage the HTTP streaming capabilities and stream addresses as they are processed by the server so that the client get a response immediately. Exploring this idea in #2052 

3. There's no way to filter out addresses at the moment, so it'd be nice to also have a way to filter addresses based on:
  a. Their state (used vs unused)
  b. Their ID 

  TODO.

4. We need to be able to catch these errors upfront before they reach exchanges' production environment. However, generating such large wallets is tricky. We already have a special "unsound wallet" we use in nightly benchmarks which recognizes addresses based on a checksum and a proportion defined upfront (like, 1% of all addresses). The downside of this special scheme is that the internal address state is very simple (only a double for the proportion) and does not exercise the wallet fully when it comes to storing and retrieving addresses. We could extend this concept to wrap the random and sequential wallets around this, so that we also get to test functionalities related to the address space too. Done in #2081 #2084

---

# QA 

- Last nightly benchmarks results (https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/636#e79c244b-d2c0-43a5-8d24-36d07f5bc7bf)

```
BenchSeqResults:
  listAddressesTime: 6.834 ms
  utxoStatistics:
    = Total value of -0 lovelace across 0 UTxOs
 
BenchRndResults:
  listAddressesTime: 385.6 μs
  utxoStatistics:
    = Total value of -0 lovelace across 0 UTxOs
 
BenchRndResults:
  listAddressesTime: 142.0 ms
[...]

PR #2033, opened 2020-08-12
Mark addresses as used while discovering them.

KtorZ:
# Issue Number

#2032 

# Overview

- dd3cff59cef4a66bb792adda6491f54b8ea8ca23
  :round_pushpin: **mark addresses as used as they are discovered**
    The main issue of the current list address handler is that it loads in memory and crawls the ENTIRE transaction history of the wallet. This is done merely to know whether addresses have been used or not. This is unnecessary for random addresses because used and unused addresses are already separated in two different Maps. For sequential wallets it's a bit more subtle but, we can mark addresses as "Used" during discovery and store this information in the database.

- 61bbd7b3f0755386a815490868e2e556e13a8e8d
  :round_pushpin: **add two properties for Random & Sequential wallets showing the effect of IsOurs on the address state**
  
- eed460c03a8ce2f5093e59f7d58acf3422653d49
  :round_pushpin: **fix random address import to not mark every imported address as 'Used'**
  

# Comments

- [ ] **NOTE 1**: This change will require a database migration forcing wallets to re-synchronize from scratch. 

- [x] ~~**NOTE 2**: I haven't yet fixed unit tests which need some adjustment since some of the interfaces have changed.~~

- [ ] **NOTE 3**: I'll try measuring and comparing the behavior with a stress-wallet that owns many addresses.

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -13,6 +13,7 @@
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
@@ -752,7 +753,8 @@ restoreWallet ctx wid = db & \DBLayer{..} -> do
             restoreBlocks @ctx @s @k @t ctx wid bs h
             saveParams @ctx @s @k ctx wid ps
     liftIO (follow nw tr cps forward (view #header)) >>= \case
-        FollowInterrupted -> pure ()
+        FollowInterrupted ->
+            pure ()
         FollowFailure ->
             restoreWallet @ctx @s @t @k ctx wid
         FollowRollback point -> do
@@ -1044,7 +1046,6 @@ manageRewardBalance _ ctx wid = db & \DBLayer{..} -> do
 listAddresses
     :: forall ctx s k.
         ( HasDBLayer s k ctx
-        , IsOurs s Address
         , CompareDiscovery s
         , KnownAddresses s
         )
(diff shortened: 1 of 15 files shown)


lib/core-integration/src/Test/Integration/Framework/Request.hs
```diff
@@ -176,23 +183,38 @@ request ctx (verb, path) reqHeaders body = do
...
+    -- in case it's streaming responses. In practice, a smart client would process
+    -- chunk one by one, but within our integration framework, it is easier to
+    -- consider streams as a list.
+    --
+    -- Therefore, in case where the response is stream, we simply decode each
+    -- chunk as a generic JSON, re-encode them as a list (trimming the '\n'
+    -- separator), and decoding them as a JSON list.
+    decodeOne = Aeson.decodeStrict . mconcat
+    decodeStream chunks = do
+        xs <- traverse (Aeson.decodeStrict @Aeson.Value . B8.init) chunks
```
rvl: I don't think you can necessarily assume that the chunks are split nicely at the newline.
The OS might interrupt the socket read early, and you will have a line in two pieces, neither of which are well-formed json.

This streaming change is worthy of its own PR, where the details can be sorted out.

I like the idea of streaming in general, because it removes one case of unbounded memory usage.

But on the other hand, if you have 1 million addresses, that would be perhaps 80MB of encoded json, as an indication of size. So the memory cost of not streaming is smallish. Without streaming there is a delay while the server is reading the list from sqlite, but unless the clients also have stream processing, so what?

rvl (commented):
I wholly approve of the change to mark sequential addresses as used once they are discovered, so that this information is easily known when listing addresses.

I also like the idea of streaming responses, but the topic needs further discussion in a separate PR.

lib/core-integration/src/Test/Integration/Framework/Request.hs
KtorZ: @rvl actually.. there's one test failing because of this which shows that sometimes, there are chunks that only contains half of an address (it fails on the "import 15.000 addresses and list them" scenario, I haven't noticed this below 100 addresses). 

So yeah, this require some additional thoughts maybe. I'll split the streaming approach into a separate PR.

KtorZ:
@rvl I ended up adding a simple migration for this. Reviewing the migration approach of the wallet can be done as a separate task / debt. I've tested the query with a hand-crafted database with 100k address and txout rows arbitrarily generated (with however existing relationships between tables) and the query is executed almost instantly.

rvl (approved):
Migration looks OK

PR #2034, opened 2020-08-12
Call getAccountBalance for all wallets at once on tip change

Anviking:
# Issue Number

#2005 

# Overview

- [x] Only send the LSQ query `GetFilteredDelegationsAndRewardAccounts` when the tip changes
- [x] Query _all_ wallet reward accounts in _the same_ `GetFilteredDelegationsAndRewardAccounts`

Listing stake pools with 10 wallets now takes ~5 seconds instead of 2-3 minutes!

- [x] Make the implementation easier to follow and actually sane

# Comments

- [ ] Not completely sure this doesn't mess up some some intricate interactions between the reward balance state and utxo state. The integration tests pass. But maybe there are some rollback scenarios which could lead to the reward balance being incorrectly reported for a while… Would be nice to think through properly and synchronise them. 

I think this happens on tip changes:

1. Wallet worker updates UTxO
2. Tip worker re-feches balances and writes to TVar
3. (I think on next tip change) Wallet reads the balance from the TVar and writes it to Sqlite
4. New reward balance is queryable from API

With this PR, I wonder if it may not risk being out of date for one more tip-change


Anviking:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -328,6 +349,8 @@ withNetworkLayer tr np addrInfo versionData action = do
     connectNodeTipClient
         :: HasCallStack
         => RetryHandlers
+        -> TVar IO (Map W.ChimericAccount (Maybe W.Coin))
```
KtorZ: What about using `0` as a default value instead of `Nothing`. We do already default to `0` when it comes to the reward balance and it is quite plausible / will not cause any particular harm to report a balance as 0 until we fetch it.

KtorZ (commented):
I like the principle here of fetching them all in one query. We need to make sure however to clean accounts from the list when wallets are deleted. Otherwise, we end up fetching unused account over and over. We need some form "pub / sub" mechanism.

Anviking:
Meant:
bors try

Anviking:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -984,15 +1092,14 @@ instance TPraosCrypto sc => ToText (NetworkLayerLog sc) where
         MsgLocalStateQueryEraMismatch mismatch ->
             "Local state query for the wrong era - this is fine. " <>
             T.pack (show mismatch)
-        MsgGetRewardAccountBalance bh acct -> T.unwords
+        MsgGetRewardAccountBalance tip accts -> T.unwords
             [ "Querying the reward account balance for"
-            , pretty acct
+            , fmt $ listF accts
```
Anviking, line 1163: With 10 wallets:
```
[cardano-wallet.network:Info:15] [2020-08-17 10:37:49.31 UTC] Querying the reward account balance for  [1ed0bdd4, 2a98402c, 5237aa81, 634bea28, 69e2c672, 75765c1e, 84132baa, c662444a, d55deafa, f4ea9095] at 70617265<-[c90c882f-6094377#4570261]
```

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
Anviking: I wanted to keep "we already default to 0" separate from the core caching logic.

With the slight refactoring, I made new entries get added to a separate `toBeObserved :: Set`, instead of `Nothing` values inside the `Map`, but the "query" still returns `Maybe W.Coin` like here, requiring an explicit `fromMaybe`.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -767,6 +761,92 @@ mkTipSyncClient tr np localTxSubmissionQ onTipUpdate onPParamsUpdate onInterpret
...
+        -> Observer IO W.ChimericAccount W.Coin
+    observer cacheVar toBeObservedVar =
+        Observer
+            { startObserving = \k ->
+                atomically $ do
+                    modifyTVar' toBeObservedVar (Set.insert k)
+            , stopObserving = \k ->
+                atomically $ do
+                    modifyTVar' toBeObservedVar (Set.delete k)
+                    modifyTVar' cacheVar (Map.delete k)
```
Anviking: Integration tests / [ STAKE_POOLS_JOIN_01](https://github.com/input-output-hk/cardano-wallet/blob/f66a86e01b89a0fbe32df6623f3923fc3d9fd6bd/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs#L152-L307) pass, meaning this should work.

But if this `Observer` abstraction is something we want to keep, it would be nice to add more extensive unit tests. `newRewardBalanceFetcher` would probably have to be made more abstract. (Because the record itself provides very little structure and can't be tested)

I also guess there is similarities between this and the `WorkerRegistry`, that perhaps could be unified in the future.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -771,6 +761,92 @@ mkTipSyncClient tr np localTxSubmissionQ onTipUpdate onPParamsUpdate onInterpret
...
+                m <- atomically (readTVar cacheVar)
+                return $ Map.lookup k m
+            }
+
+    refresh
+        :: TVar IO (Map W.ChimericAccount W.Coin)
+        -> TVar IO (Set W.ChimericAccount)
+        -> Tip (CardanoBlock sc)
+        -> IO ()
+    refresh cacheVar toBeObservedVar tip = do
```
KtorZ: I find it surprising that `refresh` is not part of the `Observer` "interface". The function clearly has a privileged access to both TVar and make strong assumption about how they are manipulated. So to me, it seems like it belong to the same abstraction and having it as a separate function is a bit confusing.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -771,6 +761,92 @@ mkTipSyncClient tr np localTxSubmissionQ onTipUpdate onPParamsUpdate onInterpret
...
+newRewardBalanceFetcher
+    :: forall sc. Tracer IO (NetworkLayerLog sc)
+    -> W.GenesisParameters
+    -- ^ Used to convert tips for logging
+    -> TQueue IO (LocalStateQueryCmd (CardanoBlock sc) IO)
+    -> IO ( Observer IO W.ChimericAccount W.Coin
+          , Tip (CardanoBlock sc) -> IO ()
+            -- Call on tip-change to refresh
+          )
+newRewardBalanceFetcher tr gp queryRewardQ = do
```
KtorZ: It'd be nice to be able to test this bit of logic in isolation. This is currently hard because the management of the keys / accounts is mixed with the logic for fetching the reward. I'd suggest to replace the `ChimericAccount` with a `Ord k => k` and to replace `Coin` with an abstract `a`, and pass a function that yield `a` as argument. So, something like:


 ```hs
data Observer m key value = Observer
    { startObserving :: key -> m ()
    , stopObserving :: key -> m ()
    , query :: key -> m (Maybe value)
    , refresh :: tip -> m tip
    }

newObserver 
    :: Ord key
    => (tip -> [key] -> m (Map key value))
    -> Observer m key value
newObserver getValues = ...
```

Comment (KtorZ) [reply lib/shelley/src/Cardano/Wallet/Shelley/Network.hs]: This way, we can separate the network logic from the observer logic and test them separately (at least, test the observer).

Comment (Anviking) [reply lib/shelley/src/Cardano/Wallet/Shelley/Network.hs]: > It'd be nice to be able to test this bit of logic in isolation. 

Yes 👍  I alluded to it in https://github.com/input-output-hk/cardano-wallet/pull/2034#discussion_r471493798, but didn't want to explore it before hearing your feedback.

File: lib/shelley/src/Cardano/Wallet/Shelley/Network.hs (hunk 4 above)
Comment (Anviking) [reply lib/shelley/src/Cardano/Wallet/Shelley/Network.hs]: To me, `Observer` is concerned about _subscribing to_ and _accessing_ values. It theoretically be passed around and escape the `Network` module. 

`refresh` is coupled to the _creation_ of the `Observer`. It's more like an implementation detail, specific to the `Network` module.

If say the wallet registry were to have access to an `Observer`, it should not be allowed to manually call `refresh`.

File: lib/shelley/src/Cardano/Wallet/Shelley/Network.hs [hunk 6]
```diff
@@ -771,6 +766,121 @@ mkTipSyncClient tr np localTxSubmissionQ onTipUpdate onPParamsUpdate onInterpret
...
+    observer
+        :: TVar m (Map key value)
+        -> TVar m (Set key)
+        -> Observer m key value
+    observer cacheVar observedKeysVar =
+        Observer
+            { startObserving = \k -> do
+                atomically $ do
+                    modifyTVar' observedKeysVar (Set.insert k)
+                traceWith tr $ MsgAddedObserver k
```
KtorZ: Since "startObserving" is called every time `getAccountBalance` is called, this would log a line `MsgAddedObserver` in a rather misleading way and much more than it needs. We could use something like `Set.alterF` and return `Just _` or `Nothing` depending on whether the element was added (and log accordingly).

Anviking: Yup. It's connected to `nullTracer` now, so one doesn't notice, but should connect to the real tracer and change this.

lib/shelley/test/unit/Cardano/Wallet/Shelley/NetworkSpec.hs
```diff
@@ -47,6 +54,54 @@ spec = describe "NetworkLayer regression test #1708" $ do
                     race_ (watchNodeTip nl (putMVar waiter)) (takeMVar waiter)
             void $ waitAnyCancel tasks
 
+    describe "Observer" $ do
+        describe "(query k) with typical use" $ beforeAll mockObserver $ do
+            -- Using monadic-property tests /just/ for the sake of testing with
+            -- multiple keys seem worthless.
```
KtorZ: It'd be nice to test with multiple keys, since this is the main _point_ of this PR actually. Coming up with some properties would help testing possible edge-cases that aren't caught here, although may be a bit tricky here. At least, we should add two additional test scenarios:

- registering the same key twice
- registering multiple keys

Anviking: I should say, I’m much more worried about this causing further balance discrepancies than that it doesn’t handle registering the same key twice, or multiple keys, etc (though sure, good to test).

lib/shelley/test/unit/Cardano/Wallet/Shelley/NetworkSpec.hs
```diff
@@ -47,6 +67,155 @@ spec = describe "NetworkLayer regression test #1708" $ do
...
+                        let expectedValue = length k
+                        (query observer k)`shouldReturn` Just expectedValue
+
+                it "traced MsgAddedObserver, MsgWillFetch, MsgDidFetch"
+                    $ \(_, _, trVar) -> do
+                        trVar `shouldHaveTraced`
+                            [ MsgAddedObserver k
+                            , MsgWillFetch $ Set.singleton k
+                            , MsgDidFetch $ Map.singleton k v
+                            ]
```
KtorZ, line 133: I don't get this test, unless this test depends on the previous ones which would be quite bad / misleading. Tests should work in isolation.

Anviking: I wanted to experiment with slightly different usage of `describe`/`it` for more readable test output.

Here we get:
```haskell
        when refresh fails
          (query k) returns the existing v
          only MsgWillFetch is traced
```
which I think is quite neat.

(And in total:)
```
      typical use
        startObserving
          (query k) returns Nothing before startObserving
          (query k) returns v after (startObserving k >> refresh)
          traced MsgAddedObserver, MsgWillFetch, MsgDidFetch
        calling startObserving a second time
          (query k) is still v
        when refresh fails
          (query k) returns the existing v
          only MsgWillFetch is traced
        stopObserving
          makes (query k) return Nothing
```

With no shared state, the whole `typical use` tree would be a single `it`.

I'll admit it gets a bit weird here though, where the `refresh False` belongs to the whole `when refresh fails` tree, but is run in the first `it`. (We cannot have it as a `runIO` or `beforeAll`, because we cannot access the `refresh` function from outside an `it`).

We'd need some
```haskell
mapBeforeAll :: (a -> IO b) -> SpecWith b -> Spec a
```
which doesn't exist

Anviking: We _could_ probably add a `Arbitrary (Observer, refresh)`, and _do_ run most of these checks in isolation though 🤔

KtorZ: I admit that the separation is neat, but it is puzzling. So at the very least, there should be a clear note as comment explaining basically what you just explained here.

Anviking:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -770,6 +764,151 @@ mkTipSyncClient tr np localTxSubmissionQ onTipUpdate onPParamsUpdate onInterpret
...
+  where
+    observer
+        :: TVar m (Map key value)
+        -> TVar m (Set key)
+        -> Observer m key value
+    observer cacheVar observedKeysVar =
+        Observer
+            { startObserving = \k -> do
+                wasAdded <- atomically $ do
+                    notAlreadyThere <- Set.notMember k <$> readTVar observedKeysVar
```
Anviking, line 883: Seems `Set.alterF` doesn't exist with our version of `containers` , and I didn't figure out how it worked, but think a separate `readTVar` inside `atomically` is perfectly fine.

Fell-x27:
There is a bug. This build doesn't sync reward balance for fresh-restored wallets. Probably it also doesn't get reward from pools incoming after wallet sync.

Fell-x27:
Hmm..looks like the very last build (03ed32945e01086fbc918ca0b3ff22ba4b1fef73) doesn't have this problem. Prev had.

KtorZ:
Do you mind describing the context, expected behavior and observed behavior?

There are a few things that are counter intuitive at first when dealing with reward balances. And there's also a known "latency" issue we are tracking as part of anothrr ticket.

Fell-x27:
> Do you mind describing the context, expected behavior and observed behavior?
> 
> There are a few things that are counter intuitive at first when dealing with reward balances. And there's also a known "latency" issue we are tracking as part of anothrr ticket.

I mean, fresh-restored wallets were restored without reward balance. And even pre-restored wallets with non-zero reward balance randomly "lost" rewards. But with the last build problem is gone. And rewards were back. But there is another one issue with fees described in https://github.com/input-output-hk/cardano-wallet/issues/2005#issuecomment-677911996

KtorZ (approved):

Issue #2017, opened 2020-08-11
Add pool DB operation `removePools`.


jonathanknowles:
# Context

As part of the work to perform garbage collection of retired pools from the database, we need a convenient and safe way to _remove_ pools that have retired.

# Decision

Create a pool database operation `removePools`:

```hs
removePools
    :: [PoolId]
    -> stm ()
    -- ^ Remove all data relating to the specified pools.
````

# Acceptance Criteria

- [x] The pool DB _must_ provide a `removePools` operation.
- [x] The `removePools` operation _must_ remove each pool _atomically_.
- [x] The `removePools` operation _must_ write an entry to the log before removing a pool.

# Development 

# QA 

- [x] Create a property test that compares the SQLite and `MVar` implementations of `removePools` and confirms that they have the same effect, given the same inputs and initial conditions:
    - [x] After a `removePools` operation, `listPools` should not include any of the pools that were removed.

piotr-iohk:
I guess that's this one that needs to be merged first: https://github.com/input-output-hk/cardano-wallet/pull/2038.

jonathanknowles:
> I guess that's this one that needs to be merged first: #2038.

Yes, that's the one!

piotr-iohk:
lgtm.

PR #2038, opened 2020-08-13
Add pool DB operation `removePools`

jonathanknowles:
# Issue Number

#2017 

# Overview

This PR:

- [x] Adds a `removePools` operation to the pool DB layer:<br><br>
    ```hs
    removePools
        :: [PoolId]
        -> stm ()
        -- ^ Remove all data relating to the specified pools.
    ```
- [x] Provides the following implementations of `removePools`:
    - [x] `Pool.DB.MVar`
    - [x] `Pool.DB.SQLite`<br><br>
- [x] Writes a `MsgRemovingPool` message to the log whenever a pool is removed:<br><br>
    ```hs
    Removing the following pool from the database: XXXX.
    ```
- [x] Adds a property test to check that only the specified pools are removed, and no other pools.

### DIFF
--- a/lib/core/src/Cardano/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/DB/Sqlite.hs
@@ -372,6 +372,7 @@ data DBLog
     | MsgWaitingForDatabase Text (Maybe Int)
     | MsgRemovingInUse Text Int
     | MsgRemoving Text
+    | MsgRemovingDatabaseEntity Text
     | MsgRemovingDatabaseFile Text DeleteSqliteDatabaseLog
     | MsgManualMigrationNeeded DBField Text
     | MsgManualMigrationNotNeeded DBField
@@ -448,6 +449,7 @@ instance HasSeverityAnnotation DBLog where
         MsgWaitingForDatabase _ _ -> Info
         MsgRemovingInUse _ _ -> Notice
         MsgRemoving _ -> Info
+        MsgRemovingDatabaseEntity _ -> Notice
         MsgRemovingDatabaseFile _ msg -> getSeverityAnnotation msg
         MsgManualMigrationNeeded{} -> Notice
         MsgManualMigrationNotNeeded{} -> Debug
@@ -484,6 +486,8 @@ instance ToText DBLog where
             "Attempting to remove the database anyway."
         MsgRemoving wid ->
             "Removing wallet's database. Wallet id was " <> wid
+        MsgRemovingDatabaseEntity msg ->
+            "Removing the following entity from the database: " <> msg
         MsgRemovingDatabaseFile wid msg ->
             "Removing " <> wid <> ": " <> toText msg
         MsgManualMigrationNeeded field value -> mconcat
--- a/lib/core/src/Cardano/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/DB/Sqlite.hs
@@ -372,7 +372,6 @@ data DBLog
     | MsgWaitingForDatabase Text (Maybe Int)
     | MsgRemovingInUse Text Int
     | MsgRemoving Text
-    | MsgRemovingDatabaseEntity Text
     | MsgRemovingDatabaseFile Text DeleteSqliteDatabaseLog
     | MsgManualMigrationNeeded DBField Text
     | MsgManualMigrationNotNeeded DBField
@@ -449,7 +448,6 @@ instance HasSeverityAnnotation DBLog where
         MsgWaitingForDatabase _ _ -> Info
         MsgRemovingInUse _ _ -> Notice
         MsgRemoving _ -> Info
-        MsgRemovingDatabaseEntity _ -> Notice
         MsgRemovingDatabaseFile _ msg -> getSeverityAnnotation msg
         MsgManualMigrationNeeded{} -> Notice
         MsgManualMigrationNotNeeded{} -> Debug
@@ -486,8 +484,6 @@ instance ToText DBLog where
             "Attempting to remove the database anyway."
         MsgRemoving wid ->
             "Removing wallet's database. Wallet id was " <> wid
-        MsgRemovingDatabaseEntity msg ->
-            "Removing the following entity from the database: " <> msg
         MsgRemovingDatabaseFile wid msg ->
             "Removing " <> wid <> ": " <> toText msg
         MsgManualMigrationNeeded field value -> mconcat
--- a/lib/core/test/unit/Cardano/Pool/DB/Arbitrary.hs
+++ b/lib/core/test/unit/Cardano/Pool/DB/Arbitrary.hs
@@ -160,8 +160,9 @@ instance Arbitrary PoolOwner where
 
 instance Arbitrary PoolRegistrationCertificate where
     shrink (PoolRegistrationCertificate p xs m c pl md) =
-        (\xs' -> PoolRegistrationCertificate p xs' m c pl md)
-            <$> shrinkList (const []) xs
+        (\p' xs' -> PoolRegistrationCertificate p' xs' m c pl md)
+            <$> shrink p
+            <*> shrinkList (const []) xs
     arbitrary = PoolRegistrationCertificate
         <$> arbitrary
         <*> scale (`mod` 8) (listOf arbitrary)
--- a/lib/core/test/unit/Cardano/Pool/DB/Arbitrary.hs
+++ b/lib/core/test/unit/Cardano/Pool/DB/Arbitrary.hs
@@ -133,9 +133,24 @@ arbitraryChainLength = 10
 
 -- NOTE Expected to have a high entropy
 instance Arbitrary PoolId where
-    arbitrary = do
-        bytes <- vector 32
-        return $ PoolId $ B8.pack bytes
+    arbitrary = PoolId . BS.pack <$> vector 32
+    shrink (PoolId p) = do
+        let s = BS.unpack p
+        result <-
+            [ -- Zero out everything:
+              replicate 32 z
+              -- Zero out different halves:
+            ,              replicate 16 z <> drop 16 s
+            , take 16 s <> replicate 16 z
+              -- Zero out different quarters:
+            ,              replicate 8 z <> drop  8 s
+            , take  8 s <> replicate 8 z <> drop 16 s
+            , take 16 s <> replicate 8 z <> drop 24 s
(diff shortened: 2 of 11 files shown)


lib/core/test/unit/Cardano/Pool/DB/Properties.hs
```diff
@@ -188,6 +193,8 @@ properties = do
             (property . prop_rollbackRegistration)
         it "rollback of PoolRetirement"
             (property . prop_rollbackRetirement)
+        it "removePools"
+            (property . prop_removePools)
```
rvl, line 197: Still no state machine tests for stake pools?

lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -479,11 +480,18 @@ instance ToText DBLog where
             "Database "+|wid|+" is ready to be deleted"
         MsgWaitingForDatabase wid (Just count) ->
             "Waiting for "+|count|+" withDatabase "+|wid|+" call(s) to finish"
-        MsgRemovingInUse wid count ->
-            "Timed out waiting for "+|count|+" withDatabase "+|wid|+" call(s) to finish. " <>
-            "Attempting to remove the database anyway."
+        MsgRemovingInUse wid count -> mconcat
```
rvl: Why replace the combinators from a _string formatting library_ with `mconcat`? 
We should be doing the reverse.
In any case, it's not related to the issue.

lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -372,6 +372,7 @@ data DBLog
     | MsgWaitingForDatabase Text (Maybe Int)
     | MsgRemovingInUse Text Int
     | MsgRemoving Text
+    | MsgRemovingDatabaseEntity Text
```
rvl: How about being more specific?
`MsgRemovePool PoolId`

rvl (changes requested):

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
jonathanknowles: > Still no state machine tests for stake pools?

I agree this is something we should definitely add, but I also think it's a significant piece of work in its own right, and that we should therefore tackle this in a different PR.

I've added an item of technical debt here: https://jira.iohk.io/browse/ADP-406

lib/core/src/Cardano/DB/Sqlite.hs
jonathanknowles: > Why replace the combinators from a _string formatting library_ with `mconcat`?

Thanks for spotting that: removing the combinators was unintentional.

I've removed this commit.

lib/core/src/Cardano/DB/Sqlite.hs
jonathanknowles: > How about being more specific?
> `MsgRemovePool PoolId`

I do agree that we should use specific types, wherever possible, rather than `Text`. 

However, one issue is that `DBLog` type is **general** to all databases. Ideally, `DBLog` shouldn't use types that are specific to _pools_ or _wallets_.

An obvious solution would be to introduce `WalletDbLog` and `PoolDbLog` types, of which the latter would have the constructor that you suggest: `MsgRemovePool`. 

However, this would require changing references to `DBLog` in quite a few places. If we decide to do this, I suggest we create a separate PR (or make an item of technical debt), in order to avoid complicating this one too much.

What do you think?

jonathanknowles: I've created a technical debt ticket to cover the idea of defining separate `WalletDbLog` and `PoolDbLog` types here: https://jira.iohk.io/browse/ADP-407.

While tackling the above ticket, we could also go through the existing `DbLog` constructors and replace the "stringly-typed" constructors with strongly-typed equivalents in `WalletDbLog` and `PoolDbLog`. Hopefully, this should also clean up parts of the code that are obscured with log formatting logic.

lib/core/test/unit/Cardano/Pool/DB/Arbitrary.hs
```diff
@@ -133,9 +133,24 @@ arbitraryChainLength = 10
...
-        return $ PoolId $ B8.pack bytes
+    arbitrary = PoolId . BS.pack <$> vector 32
+    shrink (PoolId p) = do
+        let s = BS.unpack p
+        result <-
+            [ -- Zero out everything:
+              replicate 32 z
+              -- Zero out different halves:
+            ,              replicate 16 z <> drop 16 s
+            , take 16 s <> replicate 16 z
```
Anviking, line 144: 🆒 

Two just-thoughts:
1. I imagine generalising over [whole, half, quarter, ...] would be possible, if one really wanted to
2. I guess we don't expect the contents of the `PoolId` to ever matter, and such, zeroing out everything _should_ be enough

But mainly:

Would it make sense to move to `Cardano.Wallet.Gen` for re-usability?

jonathanknowles: > I guess we don't expect the contents of the PoolId to ever matter, and such, zeroing out everything should be enough

Unfortunately if we always zero everything out, QuickCheck often fails to find a smaller counterexample, as zeroing out everything is just a little too drastic. :)

If we zero out smaller parts, then we can often still make progress in shrinking towards a smaller counterexample. I did actually manage to find a bug this way (which I eventually fixed).

> I imagine generalising over [whole, half, quarter, ...] would be possible, if one really wanted to

We could indeed! :+1: On this occasion I chose to save the time budget for other things. Though would be interested to see if you can express it in a more concise way?

> Would it make sense to move to Cardano.Wallet.Gen for re-usability?

I agree we could do that. Do you think it would be okay to tackle this in a future task that tackles the duplication among our generators? (I'm trying to limit the scope of this PR as much as possible.)

I think we could actually make a technical debt task for this, as there are many duplicated generators in our source code.

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
```diff
@@ -844,6 +851,90 @@ prop_rollbackRetirement DBLayer{..} certificates =
...
+            mapM_ (uncurry putCertificate) certificatePublications
+        -- Next, read the latest certificates for all pools:
+        poolIdsWithRegCertsAtStart <- run poolIdsWithRegCerts
+        poolIdsWithRetCertsAtStart <- run poolIdsWithRetCerts
+        -- Next, remove a subset of the pools:
+        run $ atomically $ removePools $ Set.toList poolsToRemove
+        -- Finally, see which certificates remain:
+        poolIdsWithRegCertsAtEnd <- run poolIdsWithRegCerts
+        poolIdsWithRetCertsAtEnd <- run poolIdsWithRetCerts
+        monitor $ counterexample $ T.unpack $ T.unlines
```
Anviking, line 881: Is this amount of set-arithmetic really needed / worth it?

Though I guess the error messages are nicer than if the counterexample was just the specific element that caused the failure.

Anviking (approved):
lgtm!

lib/core/src/Cardano/DB/Sqlite.hs
rvl: I would like this to be fixed now, rather than kicking the can down the road.

jonathanknowles:
bors try

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
jonathanknowles: > Is this amount of set-arithmetic really needed / worth it?

I think you're right that there is some redundancy in the set arithmetic checks at the end. :+1:  But I think this redundancy is okay, as it can help to make failures a bit clearer when debugging.

lib/core/src/Cardano/DB/Sqlite.hs
jonathanknowles: @rvl wrote:
> I would like this to be fixed now, rather than kicking the can down the road.

It wasn't clear to me that your original feedback wasn't just a suggestion:

> How about being more specific?
> `MsgRemovePool PoolId`

I agree that this would be better, but IMO fixing the lack of type specialization in `DBLog` is really orthogonal to the purpose of this PR, whose main focus is the removal of pools from the database. That the log messages use `Text` rather than `PoolId` seems like a minor issue, and not really a valid reason to delay the whole PR. (The other constructors in this type also use `Text`, so it's not inconsistent with the original code.)

IMO, it would be better to make these changes in separate PRs:
* **PR A**: Implement the `removePools` operation.
* **PR B**: Divide `DBLog` into `PoolDbLog` and `WalletDbLog`, and convert the various "stringly-typed" constructors within `DBLog` into strongly-typed constructors.

Indeed, we already had a ticket in place to perform the refactoring here https://jira.iohk.io/browse/ADP-407

But by attempting to do the refactoring in this PR, we risk adding extra delay, as any lack of consensus about how to perform the refactoring will end up delaying the PR as a whole.

Nevertheless, I've added the following commits:

* bcec1b9173d0e5931181965c7752d3218cb777c9 (Provide database-schema-specific logging types.)
* 44a72ecce7bf23e702cb0e400f161581b4b9a31b (Introduce log constructor `MsgRemovingPool`.)
* 8e5449fb3a5f04b36b9fdcf9229c784ce196d4e0 (Use `MsgRemovingPool` in `removePools` operation.)
* b5f15796c7f182cba7a6ca95c2b325b11a06911d (Remove unused `MsgRemovingDatabaseEntity` log constructor.)

Could have a look at these commits and let me know your thoughts?

Thanks!

rvl (commented):
For the logging changes - thanks for trying to fix it.
I would go with something simpler.
Create a `PoolDBLog` data type which embeds DBLog plus your new log messages.
Then all the changes will be local to the `Pool.DB.Sqlite` module, and we don't need a `DBLog schema` type parameter everywhere.

jonathanknowles:
> Create a `PoolDBLog` data type which embeds DBLog plus your new log messages.
> Then all the changes will be local to the `Pool.DB.Sqlite` module, and we don't need a `DBLog schema` type parameter everywhere.

Thanks for this suggestion. I've changed the types as you suggest.

See commits from https://github.com/input-output-hk/cardano-wallet/pull/2038/commits/82be5049b5bd2a249bdcf6ef9385b53f50735561 onwards.

rvl (approved):
Thanks!

Issue #2044, opened 2020-08-15
BUG: cardano-wallet docker container exits with code 127 - command not found


philkunz:
# Context

I'm in the process of creating a MIT licensed TypeScript client for the the cardano wallet api, that replicates daedalus functionality in TypeScript.

For testing I checked out the repo and started the docker compose file with `NETWORK=mainnet docker-compose up`

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | Docker Images: inputoutput/cardano-node:1.18.0 and inputoutput/cardano-wallet:latest as in docker-compose file |
| Platform     | Latest Ubuntu with latest Docker |
| Installation | Docker Compose from master of this repository  |

## Expected behaviour

Everything runs normal and `curl http://localhost:8090/v2/network/information` gives a status 200 response

## Actual behaviour

cardano-node runs normal, cardano-wallet container exits with code 127, command not found. 

---

# Resolution 

---

# QA

rvl:
@philkunz Your project sounds interesting. Perhaps you could add it to [cardano-launcher](https://github.com/input-output-hk/cardano-launcher)?

philkunz:
@rvl I took a look at cardano-launcher and created a ticket there https://github.com/input-output-hk/cardano-launcher/issues/86

My goal is to make cardano consumable to the average TypeScript dev implementing a website or simple shop by nothing a more than a simple `npm install [packagename]`. I'm more than happy to include this work into iohk later should it find your approval upon completion (it is MIT licensed anyway).

Anviking:
It [seems](https://hub.docker.com/r/inputoutput/cardano-wallet/tags) `latest` is not updated immediately, but after setting this:
```bash
    image: inputoutput/cardano-wallet:dev-master-shelley
```
in `docker-compose.yml`, ` NETWORK=mainnet docker-compose up` starts correctly.

I suppose `latest` will be updated next time we make a wallet release, which might be soon, so no need to PR docker-compose-yml. cc @piotr-iohk @rvl

KtorZ:
@Anviking latest correspond to the latest release. So, on August 3rd at the moment.

piotr-iohk:
works fine in `v2020.9.11`

Issue #1919, opened 2020-07-17
Remove cardano-wallet-byron package


rvl:
# Context

ADP-356

With the hard fork combinator we have a wallet which works on both byron and shelley networks. This will be in the cardano-wallet-shelley package.

We no longer require wallets that can only work with single-era nodes.

# Decision

Remove the cardano-wallet-byron package.

Salvage any tests or benchmarks and copy them across to the cardano-wallet-shelley package.
- Restore benchmarks ⇒ #1695.

Note: grep is your friend, especially when tracking down byron references in the nix build.

# Acceptance Criteria

1. There must be only one cardano-node backed cardano-wallet.
2. Test cases, nightly tests, benchmarks, etc should not be lost because we need more automated QA not less.

---

# Development 

- [x] Add cardano-wallet-shelley restore bench to nightly #1695
- [x] Remove cardano-wallet-byron #2012 
- [x] Rename cardano-wallet-shelley to cardano-wallet #2013
- [x] Adjust wiki again
- [x] Ensure travis release script still works
- [ ] Consider moving `lib/*` to `.`
- [ ] Consider renaming `shelley` dir to `cardano-wallet`
- [x] Fix restore bench and docker #2045
- [x] Fix travis release deployment #2048

# QA 

- `cardano-wallet-byron` was removed.
- `cardano-wallet-shelley` was renamed to `cardano-wallet`
- Existing shelley tests pass.
    - Some Byron-specific integration tests were removed in #2012. Is this ok?

piotr-iohk:
I is this completed? I suppose there are integration tests to be "merged" to shelley, e.g. -> https://github.com/input-output-hk/cardano-wallet/blob/master/lib/byron/test/integration/Main.hs#L128, not sure about other stuff?

QA section filled would be helpful :sweat_smile:

Anviking:
@piotr-iohk This is not in the QA column. Are you talking about another ticket?

piotr-iohk:
Sorry, my bad. I thought I saw it in QA... Jesus :tired_face:

piotr-iohk:
> Test cases, nightly tests, benchmarks, etc should not be lost because we need more automated QA not less.

Perhaps that is not a result of this change but Benchmarks and Windows tests are failing for quite a while now. (Although benchmarks are hopefully fixed :crossed_fingers: )

piotr-iohk:
[...]

PR #2045, opened 2020-08-16
Fix restore bench and docker

Anviking:
# Issue Number

#1919, #2044 

# Overview

- [x] Fix broken restore bench  https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/608#f02f336b-c7be-4b3f-b86b-bc825c9b5ac2
- [ ] _Hope_ `docker` should be fixed, but can't easily test locally ⚠️ 

# Comments

### DIFF
--- a/nix/docker.nix
+++ b/nix/docker.nix
@@ -27,7 +27,9 @@ let
   defaultPort = "8090";
   dataDir = "/data";
 
-  startScript = writeScriptBin "start-cardano-wallet-${backend}" ''
+  suffix = if backend == "shelley" then "" else "-" + backend;
+  walletExeName = "cardano-wallet" + suffix;
+  startScript = writeScriptBin "start-wallet" ''
     #!${runtimeShell}
     set -euo pipefail
 
@@ -37,7 +39,7 @@ let
     ln -s ${dataDir} /cardano-wallet
 
     export LOCALE_ARCHIVE="${glibcLocales}/lib/locale/locale-archive"
-    exec ${exe}/bin/cardano-wallet-${backend} "$@"
+    exec ${exe}/bin/${walletExeName} "$@"
   '';
 
   # Config file needed for container/host resolution.
@@ -68,7 +70,7 @@ in
       startScript
     ];
     config = {
-      EntryPoint = [ "start-cardano-wallet-${backend}" ];
+      EntryPoint = [ "start-wallet" ];
       ExposedPorts = {
         "${defaultPort}/tcp" = {}; # wallet api
       };
--- a/.buildkite/bench-restore.sh
+++ b/.buildkite/bench-restore.sh
@@ -3,27 +3,26 @@
 
 set -euo pipefail
 
-if [ "$#" -lt 2 ]; then
-  echo "usage: $0 TARGET NETWORK"
+if [ "$#" -lt 1 ]; then
+  echo "usage: $0 NETWORK"
   exit 1
 fi
 
-target=$1
-network=$2
-artifact_name=restore-$target-$network
+network=$1
+artifact_name=restore-$network
 log=restore.log
-results=restore-$target-$network.txt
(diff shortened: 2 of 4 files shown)


nix/docker.nix
```diff
@@ -27,18 +27,22 @@ let
...
+      # set up data volume
+      export XDG_DATA_HOME=/
+      mkdir -p ${dataDir}
+      ln -s ${dataDir} /cardano-wallet
 
-    export LOCALE_ARCHIVE="${glibcLocales}/lib/locale/locale-archive"
-    exec ${exe}/bin/cardano-wallet-${backend} "$@"
-  '';
+      export LOCALE_ARCHIVE="${glibcLocales}/lib/locale/locale-archive"
+      exec ${exe}/bin/${walletExeName} "$@"
```
rvl: Since it's a bash script, you could "cheat" and use a glob.
```
exec ${exe}/bin/cardano-wallet* "$@"
```

nix/docker.nix
```diff
@@ -12,7 +12,7 @@
 
 # The main contents of the image.
 , exe
-# Short name of the backend, e.g. jormungandr or shelley
+# "", or "-jormungandr" (as suffix to cardano-wallet)
```
rvl: The backend function argument will still be either jormungandr or shelley, so this comment doesn't need to be changed.

nix/docker.nix
```diff
@@ -27,18 +27,22 @@ let
...
   dataDir = "/data";
 
-  startScript = writeScriptBin "start-cardano-wallet-${backend}" ''
-    #!${runtimeShell}
-    set -euo pipefail
+  startScript =
+    let
+      suffix = if backend == "shelley" then "" else "-" + backend;
+      walletExeName = "cardano-wallet" + suffix;
+    in writeScriptBin "start-cardano-wallet-${backend}" ''
```
rvl: The name of this start script actually doesn't matter. It could be called globbits and the image would still work.
So you can remove the backend from its name.

nix/docker.nix
```diff
@@ -27,18 +27,22 @@ let
   defaultPort = "8090";
   dataDir = "/data";
 
-  startScript = writeScriptBin "start-cardano-wallet-${backend}" ''
-    #!${runtimeShell}
-    set -euo pipefail
+  startScript =
+    let
+      suffix = if backend == "shelley" then "" else "-" + backend;
```
rvl: These variables can just as well be defined above or below to avoid a `let..in` and indentation.

nix/docker.nix
Anviking: Oops, indeed 👍

nix/docker.nix
Anviking: This I noticed, but in-doubt kept the name as before.

`start-cardano-wallet` could make you believe it was shelley specific, even when the backend is `jormungandr`.

Maybe just `start` or `start-wallet`?

nix/docker.nix
Anviking: Done.

nix/docker.nix
Anviking: I went with `start-wallet` now.

nix/docker.nix
Anviking: On second thought, that would break if we for some reason would end up with multiple matches?

If the cost is the two lines:
```nix
  suffix = if backend == "shelley" then "" else "-" + backend;
  walletExeName = "cardano-wallet" + suffix;
```
then maybe being explicit is worth it.

rvl (approved):

rvl:
Looks good, works for me.

bors r+

nix/docker.nix
rvl: The glob will be safe to use because this is the bin directory within a backend-specific package. But explicit filenames are probably easier to maintain.

Issue #2019, opened 2020-08-11
Call `removeRetiredPools` from `monitorStakePools`.


jonathanknowles:
# Context

As part of the work to garbage collect retired pools from the database, we need to periodically _initiate_ the garbage collector.

# Decision

Within `monitorStakePools`, at appropriate points call [`removeRetiredPools`](https://github.com/input-output-hk/cardano-wallet/issues/2018) with epoch number _**e**_, where:

> _**e**_ = _**e**_<sub>current</sub> − 2.

The choice to wait two epochs after retirement is a conservative one, and ensures that we will never garbage collect pools that are still within the rollback window.

# Acceptance Criteria

- [x] The pool monitoring code _must not_ call `removeRetiredPools` with an epoch number that is within the rollback window.
- [x] Any calls to `removeRetiredPools` _must_ be logged to assist with debugging.

# Development 

- https://github.com/input-output-hk/cardano-wallet/pull/2057

# QA

See:
- #2016 
- #2017

jonathanknowles:
Hi @piotr-iohk 

This was finished some time ago, but I forgot to move it into the QA bucket. My apologies!

Hopefully there is nothing more to do, but let me know if you spot anything.

If all is okay, then we can also close https://jira.iohk.io/browse/ADP-376.

piotr-iohk:
@jonathanknowles, so I'm looking into integration tests while there are running and we have there 4 pools initially with one of them that is scheduled to be retired in epoch 3. The pool is no longer displayed in list pools since epoch 3 which is OK. I expected however that since epoch 5, data related to this retired pool will be removed from DB, but I still see bits of this pool in DB in epoch 5 and onwards:
E.g. (there are 4 pools and I was expecting 3)

![Screenshot from 2020-09-15 09-26-36](https://user-images.githubusercontent.com/42900201/93179402-ebd2a780-f735-11ea-9798-b2bb46a5a6b4.png)

![Screenshot from 2020-09-15 09-26-51](https://user-images.githubusercontent.com/42900201/93179423-f2611f00-f735-11ea-9a0e-930009bc100f.png)

![Screenshot from 2020-09-15 09-27-01](https://user-images.githubusercontent.com/42900201/93179427-f4c37900-f735-11ea-8266-b096f69e1b89.png)

![Screenshot from 2020-09-15 09-27-09](https://user-images.githubusercontent.com/42900201/93179436-f7be6980-f735-11ea-9632-7793866cb197.png)

[...]

Issue #2018, opened 2020-08-11
Add pool DB operation `removeRetiredPools`


jonathanknowles:
# Context

As part of the work to perform garbage collection of retired pools from the database, we need to be able to conveniently and safely remove _retired_ pools.

# Decision

Create a pool database operation `removeRetiredPools`:

```hs
removeRetiredPools
    :: EpochNo
    -> stm [PoolId]
    -- ^ Remove all pools with an active retirement epoch that is
    -- earlier than or equal to the specified epoch.
```

This operation _reuses_ the existing [`listRetiredPools`](https://github.com/input-output-hk/cardano-wallet/issues/2016) and [`removePools`](https://github.com/input-output-hk/cardano-wallet/issues/2017) operations.

# Acceptance Criteria

- [x] The pool DB _must_ provide a `removeRetiredPools` operation.
- [x] The `removeRetiredPools` operation _must_ remove each pool _atomically_.
- [x] The `removeRetiredPools` operation _must_ write an entry to the log before removing a pool.

# Development 

- #2047 

# QA 

Property test:
- https://github.com/input-output-hk/cardano-wallet/pull/2078

This operation _reuses_ the existing [`listRetiredPools`](https://github.com/input-output-hk/cardano-wallet/issues/2016) and [`removePools`](https://github.com/input-output-hk/cardano-wallet/issues/2017) operations. The tests for these operations are described in the following tasks:
- #2016 
- #2017

piotr-iohk:
lgtm

PR #2047, opened 2020-08-18
 Add pool DB operation `removeRetiredPools`. 

jonathanknowles:
# Issue Number

#2018 

# Overview

Building on the work added in PRs #2024 and #2038, this PR:

- [x] Adds a `removeRetiredPools` operation to the pool DB layer:<br><br>
    ```hs
    removeRetiredPools  
        :: EpochNo
        -> stm [PoolRetirementCertificate]
        -- ^ Remove all pools with an active retirement epoch that is earlier 
        -- than or equal to the specified epoch.  
        --
        -- Returns the retirement certificates of the pools that were removed.
    ```
- [x] Writes a bracketed `MsgRemovingRetiredPools` message to the log whenever the `removeRetiredPools` operation is called:<br><br>
    ```
    Looking for pools that retired in or before epoch 1000.
    Removing the following retired pools:
        Pool 4355a46b with retirement epoch 999
        Pool 53c234e5 with retirement epoch 999
        Pool 121cfccd with retirement epoch 1000
        Pool 917df332 with retirement epoch 1000
    Finished removing retired pools.
    ```

# Further Work

Before garbage collection can actually happen, the wallet needs to call `removeRetiredPools` at appropriate intervals. This work is deferred to a future PR.

See issue #2019.

### DIFF
--- a/lib/core/src/Cardano/Pool/DB.hs
+++ b/lib/core/src/Cardano/Pool/DB.hs
@@ -203,6 +203,19 @@ data DBLayer m = forall stm. (MonadFail stm, MonadIO stm) => DBLayer
         -> stm ()
         -- ^ Remove all data relating to the specified pools.
 
+    , removeRetiredPools
+        :: EpochNo
+        -> stm [PoolRetirementCertificate]
+        -- ^ Remove all pools with an active retirement epoch that is earlier
+        -- than or equal to the specified epoch.
+        --
+        -- Returns the retirement certificates of the pools that were removed.
+        --
+        -- See also:
+        --
+        --    - 'listRetiredPools'.
+        --    - 'removePools'.
+
     , cleanDB
         :: stm ()
         -- ^ Clean a database
--- a/lib/core/src/Cardano/Pool/DB.hs
+++ b/lib/core/src/Cardano/Pool/DB.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE ExistentialQuantification #-}
 {-# LANGUAGE GADTs #-}
+{-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedLabels #-}
 {-# LANGUAGE RankNTypes #-}
 
@@ -16,6 +17,7 @@ module Cardano.Pool.DB
 
       -- * Utilities
     , determinePoolLifeCycleStatus
+    , removeRetiredPools
 
       -- * Errors
     , ErrPointAlreadyExists (..)
@@ -23,6 +25,8 @@ module Cardano.Pool.DB
 
 import Prelude
 
+import Cardano.DB.Sqlite
+    ( DBLog (..) )
 import Cardano.Wallet.Primitive.Types
     ( BlockHeader
     , CertificatePublicationTime (..)
@@ -39,19 +43,27 @@ import Cardano.Wallet.Primitive.Types
 import Control.Monad.Fail
     ( MonadFail )
 import Control.Monad.IO.Class
-    ( MonadIO )
+    ( MonadIO, liftIO )
 import Control.Monad.Trans.Except
     ( ExceptT )
+import Control.Tracer
+    ( Tracer, traceWith )
 import Data.Generics.Internal.VL.Lens
     ( view )
 import Data.Map.Strict
     ( Map )
 import Data.Quantity
     ( Quantity (..) )
+import Data.Text.Class
+    ( toText )
 import Data.Word
     ( Word64 )
+import Fmt
+    ( pretty )
 import System.Random
     ( StdGen )
+
+import qualified Data.Text as T
 
 -- | A Database interface for storing pool production in DB.
 --
@@ -203,19 +215,6 @@ data DBLayer m = forall stm. (MonadFail stm, MonadIO stm) => DBLayer
         -> stm ()
         -- ^ Remove all data relating to the specified pools.
 
-    , removeRetiredPools
-        :: EpochNo
-        -> stm [PoolRetirementCertificate]
-        -- ^ Remove all pools with an active retirement epoch that is earlier
-        -- than or equal to the specified epoch.
-        --
-        -- Returns the retirement certificates of the pools that were removed.
-        --
-        -- See also:
-        --
-        --    - 'listRetiredPools'.
-        --    - 'removePools'.
-
     , cleanDB
         :: stm ()
         -- ^ Clean a database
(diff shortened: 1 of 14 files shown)


lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -421,6 +419,20 @@ newDBLayer trace fp timeInterpreter = do
         , atomically = runQuery
         })
     where
+        listRetiredPools_ epochNo = do
+            let query = T.unwords
+                    [ "SELECT * FROM "
+                    , databaseViewName activePoolRetirements
+                    , "WHERE retirement_epoch <="
+                    , T.toText epochNo
```
rvl: Isn't there a `?` parameter syntax that we should be using here.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -383,16 +370,27 @@ newDBLayer trace fp timeInterpreter = do
...
+                [ "Looking for pools that retired in or before epoch "
+                , toText epoch
+                , "."
+                ]
+            listRetiredPools_ epoch >>= \case
+                [] -> do
+                    report "Found no retired pools."
+                    pure []
+                retirementCerts -> do
+                    report $ T.unlines
```
rvl: Part of the benefit of typed tracing is that you don't need to obscure the code with log string formatting.

I would make two messages:
 - `MsgRemoveRetiredPools EpochNumber BracketLog` (covers start and finish - use with `bracketTracer`)
 - `MsgRemoveRetiredPoolsCerts [Cert]`

Then put all the message formatting in the ToText instance.

rvl (changes requested):

lib/core/src/Cardano/Pool/DB/Sqlite.hs
jonathanknowles: > Isn't there a `?` parameter syntax that we should be using here.

I don't know, is there?

lib/core/src/Cardano/Pool/DB/Sqlite.hs
jonathanknowles: @rvl wrote:
> Part of the benefit of typed tracing is that you don't need to obscure the code with log string formatting.

Thanks for this feedback.

I do agree with you. However, one issue is that `DBLog` type is _general_ to all databases. Ideally, `DBLog` itself shouldn't use types that are specific to _pools_ or _wallets_. 

An obvious solution would be to introduce `WalletDbLog` and `PoolDbLog` types, of which the latter would have the constructor that you suggest: `MsgRemoveRetiredPools`. However, this would require changing references to `DBLog` in quite a few places. If we decide to do this, I suggest we create a **separate PR** (or make an item of technical debt), in order to avoid complicating this PR too much.

> I would make two messages:
>   * `MsgRemoveRetiredPools EpochNumber BracketLog` (covers start and finish - use with `bracketTracer`)
>   * `MsgRemoveRetiredPoolsCerts [Cert]`

Thanks for this suggestion: I didn't know about `bracketTracer`.

I've adjusted `removeRetiredPools` to use bracket-style tracing, with the following general constructors (defined within `DBLog`):

```hs
| MsgGarbageCollection Text BracketLog
| MsgGarbageCollectionStep Text
```

See: c912a269501343563283281b5771c66b619e5d32

jonathanknowles:
bors try

lib/core/src/Cardano/Pool/DB/Sqlite.hs
jonathanknowles: I've created a technical debt ticket to cover the idea of defining separate `WalletDbLog` and `PoolDbLog` types here: https://jira.iohk.io/browse/ADP-407.

While tackling the above ticket, we could also go through the existing `DbLog` constructors and replace the "stringly-typed" constructors with strongly-typed equivalents in `WalletDbLog` and `PoolDbLog`. Hopefully, this should also clean up parts of the code that are obscured with log formatting logic.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
rvl: Yes we must maintain a good habit of always using `?` value placeholders. Even though in this particular instance it is safe.

Injection is the number 1 OWASP risk.

Also I am not fond of using views when normal queries would work just as well.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
jonathanknowles: @rvl we now have the following types:

https://github.com/input-output-hk/cardano-wallet/blob/ab68664b81304faaa3037ce5c1e7a772f555ba45/lib/core/src/Cardano/Pool/DB/Log.hs#L35-L36

lib/core/src/Cardano/Pool/DB/Sqlite.hs
jonathanknowles: > Yes we must maintain a good habit of always using ? value placeholders. Even though in this particular instance it is safe.

Fair enough.

Since this PR no longer touches this area of the code, I will create a separate PR to add `?` value placeholders.

**EDIT: See separate PR here: https://github.com/input-output-hk/cardano-wallet/pull/2053**

lib/core/src/Cardano/Pool/DB.hs
```diff
@@ -280,6 +288,33 @@ determinePoolLifeCycleStatus mReg mRet = case (mReg, mRet) of
...
+removeRetiredPools
+    DBLayer {atomically, listRetiredPools, removePools} trace epoch =
+        bracketTracer (contramap actionMessage trace) action
+  where
+    actionMessage = MsgRemovingRetiredPoolsForEpoch epoch
+
+    action = atomically (listRetiredPools epoch) >>=
+        \retirementCerts -> do
+            traceWith trace $ MsgRemovingRetiredPools retirementCerts
+            atomically $ removePools (view #poolId <$> retirementCerts)
```
rvl: The list and remove queries should be within a single `atomically` block. You should be able to liftIO the logging statement.

rvl (changes requested):

lib/core/src/Cardano/Pool/DB.hs
jonathanknowles: > The list and remove queries should be within a single atomically block. You should be able to liftIO the logging statement.

Good point.

Fixed in 5f9c7df9f9a7c66c38d7bc74ad502737ec12febc.

lib/core/src/Cardano/Pool/DB/Log.hs
```diff
@@ -0,0 +1,67 @@
...
+    deriving (Eq, Show)
+
+instance HasPrivacyAnnotation PoolDbLog
+
+instance HasSeverityAnnotation PoolDbLog where
+    getSeverityAnnotation = \case
+        MsgGeneric e -> getSeverityAnnotation e
+        MsgRemovingPool {} -> Notice
+        MsgRemovingRetiredPools {} -> Notice
+        MsgRemovingRetiredPoolsForEpoch {} -> Notice
```
rvl: I don't think all of these should be `Notice` level. Ideally there should be one notice for when a pool is registered and just one more notice for when it retires. The rest is probably debug.

lib/core/src/Cardano/Pool/DB/Log.hs
```diff
@@ -0,0 +1,67 @@
...
+        MsgRemovingRetiredPools poolRetirementCerts -> T.unlines
+            [ "Removing the following retired pools:"
+            , T.unlines (pretty <$> poolRetirementCerts)
+            ]
+        MsgRemovingRetiredPoolsForEpoch epoch nestedMessage -> T.concat
+            [ "Removing pools that retired in or before epoch "
+            , toText epoch
+            , ": "
+            , toText nestedMessage
+            ]
```
rvl, line 67: The [`Fmt`](http://hackage.haskell.org/package/fmt-0.6.1.2/docs/Fmt.html) module has better functions for constructing string messages.

rvl (approved):
Two comments about logging for you to address. Since `removeRetiredPools` is not used anywhere I assume there will be another PR.

jonathanknowles:
> Since `removeRetiredPools` is not used anywhere I assume there will be another PR.

That's correct. From the PR description:

> Before garbage collection can actually happen, the wallet needs to call `removeRetiredPools` at appropriate intervals. This work is deferred to a future PR. See issue #2019.

lib/core/src/Cardano/Pool/DB/Log.hs
jonathanknowles: > I don't think all of these should be Notice level. Ideally there should be one notice for when a pool is registered and just one more notice for when it retires. The rest is probably debug.

Fixed in 024d36b7c51dac6cb9624697c88fcfa32b82f997.

lib/core/src/Cardano/Pool/DB/Log.hs
jonathanknowles: > The [`Fmt`](http://hackage.haskell.org/package/fmt-0.6.1.2/docs/Fmt.html) module has better functions for constructing string messages.

Thanks. I've made a note to research this a bit more.

PR #2048, opened 2020-08-18
Adjust travis deployment for renamed cardano-wallet binary

Anviking:
# Issue Number

#1919 

# Overview

- [x] Fix naming issue caused by the renaming from `cardano-wallet-shelley` to `cardano-wallet`
- [x] ~Simply use the archives from hydra as-is. No need to unzip an re-zip. We don't rename the executable anymore.~ @rvl has made it use same filenames as previous releases.
- [x] Use a script for picking builds of the exact revision from hydra.

# Comments

- The adjustment of the cardano-wallet-jormungandr archives is left as-is.

Comparing latest release to manually calling `curl -L https://hydra.iohk.io/job/Cardano/cardano-wallet/cardano-wallet-linux64/latest/download/1 --output cardano-wallet-$TRAVIS_TAG-linux64.tar.gz`

The extracted folder appears different. Does that matter?

Skärmavbild 2020-08-18 kl  11 58 07

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,7 @@ jobs:
     ############################
 
     # Fetch the linux archive
-    - travis_retry curl -L https://hydra.iohk.io/job/Cardano/cardano-wallet/cardano-wallet-linux64/latest/download/1 | --output cardano-wallet-$TRAVIS_TAG-linux64.tar.gz
+    - travis_retry curl -L https://hydra.iohk.io/job/Cardano/cardano-wallet/cardano-wallet-linux64/latest/download/1 --output cardano-wallet-$TRAVIS_TAG-linux64.tar.gz
     # Fetch the MacOS archive
     - travis_retry curl -L https://hydra.iohk.io/job/Cardano/cardano-wallet/cardano-wallet-macos64/latest/download/1 --output cardano-wallet-$TRAVIS_TAG-macos64.tar.gz
     # Fetch the Windows archive
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,10 @@ jobs:
     if: type != pull_request AND (tag =~ ^v OR commit_message =~ /TRAVIS_TRIGGER_RELEASE/)
     name: "Executables"
     script:
+    # Fetch the builds from Hydra
+    - npm install --no-save axios lodash
+    - node scripts/travis-download-builds.js
+
     ############################
     # cardano-node
     ############################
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -54,26 +54,11 @@ jobs:
     ############################
 
     # Fetch the linux archive
-    - travis_retry curl -L https://hydra.iohk.io/job/Cardano/cardano-wallet/cardano-wallet-linux64/latest/download/1 | tar xz
-    - find . -maxdepth 1 -type d -name "cardano-wallet-*" -exec mv \{} cardano-wallet-linux64 \;
```
Anviking, line 58: `cardano-wallet-*` wouldn't match `cardano-wallet`.

.travis.yml
```diff
@@ -54,26 +54,11 @@ jobs:
...
-    - mv cardano-wallet-linux64/cardano-wallet cardano-wallet-linux64/cardano-wallet
-    - tar czf cardano-wallet-$TRAVIS_TAG-linux64.tar.gz cardano-wallet-linux64
-    - rm -rf cardano-wallet-linux64
-
+    - travis_retry curl -L https://hydra.iohk.io/job/Cardano/cardano-wallet/cardano-wallet-linux64/latest/download/1 | --output cardano-wallet-$TRAVIS_TAG-linux64.tar.gz
```
Anviking:
```suggestion
    - travis_retry curl -L https://hydra.iohk.io/job/Cardano/cardano-wallet/cardano-wallet-linux64/latest/download/1 --output cardano-wallet-$TRAVIS_TAG-linux64.tar.gz
```
😅

rvl:
I think the travis script needed to do some renaming and cleanup. Some filenames from Hydra were not right for example. But perhaps that could be fixed.

I also added a script for downloading builds for the tagged version, rather than just latest master. This can be used instead of the `curl` commands.

Anviking:
> I think the travis script needed to do some renaming and cleanup. Some filenames from Hydra were not right for example. But perhaps that could be fixed.

I believe it was only used to rename the `cardano-wallet-shelley` binary to `cardano-wallet`, which is not needed now.

> I also added a script for downloading builds for the **tagged version, rather than just latest master**. This can be used instead of the curl commands.

💯 🙏

rvl (approved):

rvl:
I made it follow naming conventions from previous releases (except for the byron/shelley split).

And I have tested it by copying and pasting commands from `.travis.yaml` into a temporary bash script file.

bors r+

PR #2049, opened 2020-08-18
Reorganize: move lib/* to .

Anviking:
# Issue Number

#1919 

# Overview

- [x] Move the package directories in `lib/` to the top level, as suggested by @rvl 
- [x] Rename `shelley` directory to `cardano-wallet` (matching the package name `cardano-wallet`)

# Comments

Discussion in https://input-output-rnd.slack.com/archives/GBT05825V/p1597051716330400


Anviking:
bors try

rvl (approved):
:+1: You may wish to give some warning before merging this.
The `cardano-wallet` package subdirectory looks a bit odd though. I thought it was OK to keep it as `shelley`.

jonathanknowles (commented):
> Move the package directories in `lib/` to the top level

Sorry to be contrarian, but I would actually prefer that we didn't do this! :smile_cat: 

I rather like the fact that we have a directory just for packages. By moving everything to the top level, it's no longer possible to tell, at a glance, which directories correspond to packages, and which do not. When using source code manipulation tools (like `sed -i`), it's easy to point them at the `lib` subdirectory rather than having to go to the effort of filtering out all the other stuff from the top level.

Having said that, I'm not strongly attached to the name `lib`, and wouldn't mind if it's renamed to something else like `packages`, if people felt that was more appropriate.

Anviking:
Closing, as gathering consensus seems difficult.

PR #2050, opened 2020-08-18
Increase the maximum address pool gap to 100 000

KtorZ:
# Issue Number

N/A

# Overview

- [x] I have increased the maximum address pool gap to 100 000

# Comments

- NOTE 1: This was done originally when exploring options to help exchanges with large Byron wallets, discarded, but now resurrected as needed by other exchanges who wants to leverage Shelley wallets NOW. I still think that we should work on a dedicated wallet scheme for exchange, but in the meantime, this would do. 

- NOTE 2: I originally tested to restore a Shelley wallet with a gap of 100k on Mainnet, it took a bit less than 8 hours BUT, this was before we find and fix the issue with the logger so I'd like to re-conduct such test with the fix on to see.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
@@ -7,6 +7,7 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TupleSections #-}
@@ -152,7 +153,7 @@ instance ToText (AddressPoolGap) where
 
 instance Bounded AddressPoolGap where
     minBound = AddressPoolGap 10
-    maxBound = AddressPoolGap 100
+    maxBound = AddressPoolGap 100_000
 
 instance Enum AddressPoolGap where
     fromEnum (AddressPoolGap g) = fromEnum g
--- a/lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/SequentialSpec.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/SequentialSpec.hs
@@ -62,6 +62,7 @@ import Cardano.Wallet.Primitive.AddressDiscovery.Sequential
     , mkAddressPoolGap
     , mkSeqStateFromAccountXPub
     , mkSeqStateFromRootXPrv
+    , mkUnboundedAddressPoolGap
     , shrinkPool
     )
 import Cardano.Wallet.Primitive.Types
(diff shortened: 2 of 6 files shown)


lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
```diff
@@ -152,7 +152,7 @@ instance ToText (AddressPoolGap) where
 
 instance Bounded AddressPoolGap where
     minBound = AddressPoolGap 10
-    maxBound = AddressPoolGap 100
+    maxBound = AddressPoolGap 100000
```
jonathanknowles: Just a suggestion, but I find these long constants easier to read with `NumericUnderscores`, so:

```suggestion
    maxBound = AddressPoolGap 100_000
```
Perhaps my eyes are getting old :smile_cat:

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/SequentialSpec.hs
```diff
@@ -161,15 +162,15 @@ spec = do
 
     describe "AddressPoolGap - Text Roundtrip" $ do
         textRoundtrip $ Proxy @AddressPoolGap
-        let err = "An address pool gap must be a natural number between 10 and 100."
+        let err = "An address pool gap must be a natural number between 10 and 100000."
```
jonathanknowles, line 170:
```suggestion
        let err = "An address pool gap must be a natural number between 10 and 100,000."
```

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/SequentialSpec.hs
```diff
@@ -596,7 +597,7 @@ deriving instance Arbitrary a => Arbitrary (ShowFmt a)
 
 instance Arbitrary AddressPoolGap where
     shrink _ = []
-    arbitrary = arbitraryBoundedEnum
+    arbitrary = mkUnboundedAddressPoolGap <$> choose (10, 20)
```
jonathanknowles, line 620: Hi @KtorZ 

Just out of curiosity, what's the reason for the upper limit of `20` here?

Is it to keep the running time of tests within a reasonable limit?

jonathanknowles (approved):
LGTM. I just have a question and a couple of suggestions. (See comments.)

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/SequentialSpec.hs
KtorZ: > Is it to keep the running time of tests within a reasonable limit?

Precisely. Remember that the gap serves as an upper bound for how many addresses we need to generate. Generating 100k addresses take roughly 30s on my machine. For a property test that runs thousands of scenarios,  that is not acceptable. 

Plus, having such large gap is not recommended (I wanted to actually add a disclaimer in the swagger doc which I'll do right away) so we can't really (nor are willing to) provide any strong guarantees about performances under such conditions. "Use at your own risks".

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
KtorZ: Done :+1: , I thought I did it already :shrug: ... I thought of it for sure.

KtorZ:
bors retry

PR #2052, opened 2020-08-20
Explore streaming for list endpoints.

KtorZ:
# Issue Number

#2032 

# Overview

- [ ] I have made the API return a stream of addresses instead of a list

    The list of address can be quite large, and even, too large for just sending it back as one big byte chunk. Leverage HTTP streaming, we can have the server immediately replying to the client, sending one address per line so that client can start processing the response as it receives it. This is much more efficient and we should probably consider it for all endpoints returning arbitrary long lists at the moment (e.g. transactions, stake pools). The major 'gotcha' however is that we need to rework some of code that interact with the API to actually handle streams properly.

# Comments

Opening for discussions / explorations on the topic. As @rvl pointed out in a previous PR, even returning 100K addresses is still relatively reasonable (< 200MB memory footprint with a very broad estimation). 

There are obviously pros and cons which I'd summarize as such:

**pros**

- Pretty much instantaneous response from the server since data are already in-memory and can be streamed right away.
- Clients can start processing objects as they receive them (modulo some smart client to also wait for partial chunks). 
- It scales pretty well and keep the API simple (no need for extra query parameters or headers).
- The approach can easily be re-used for other endpoints returning large lists
- Clients that aren't able / willing to process data in a streaming fashion can always fall back to waiting for the entire chunk to be received.

**cons**

- It becomes a bit more awkward to document and deal with because the stream itself is no longer a valid JSON object, but a newline-separated list of JSON objects.

- Server-side, there's no particular gain in doing this since the data already is stored entirely in memory. It'd be more beneficial if we were reading from the database also as a stream, but here it arguably adds extra complexity to clients without much serving any performance improvement on the server.

- Despite being a very standard side of HTTP, streams are still quite "exotic" and may confuse potential users (:disappointed:) in the same way some got confused by the use of HTTP verb other than `GET` and `POST`.

### DIFF
--- a/lib/cli/cardano-wallet-cli.cabal
+++ b/lib/cli/cardano-wallet-cli.cabal
@@ -43,6 +43,7 @@ library
     , http-client
     , iohk-monitoring
     , network-uri
+    , servant
     , servant-client
     , servant-client-core
     , text
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -156,6 +156,8 @@ import Cardano.Wallet.Primitive.SyncProgress
     ( SyncTolerance (..) )
 import Cardano.Wallet.Primitive.Types
     ( AddressState, Coin (..), Hash, SortOrder, WalletId, WalletName )
+import Cardano.Wallet.Unsafe
+    ( unsafeRunExceptT )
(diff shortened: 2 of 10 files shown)


KtorZ:
closing this until we do actually have some time exploring this. Will open a JIRA ticket for tracking this in backlog.

PR #2055, opened 2020-08-21
Fix user facing encoding of stake pool ids (from base16 to bech32).

hasufell:
So I kept the legacy parsing for both json and the query params to not break API for existing users.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -177,6 +177,8 @@ import Cardano.Wallet.Primitive.Types
     , WalletBalance (..)
     , WalletId (..)
     , WalletName (..)
+    , decodePoolIdBech32
+    , encodePoolIdBech32
     , isValidCoin
     , unsafeEpochNo
     )
@@ -1096,9 +1098,11 @@ instance ToJSON ApiByronWalletBalance where
     toJSON = genericToJSON defaultRecordTypeOptions
 
 instance FromJSON (ApiT PoolId) where
-    parseJSON = parseJSON >=> eitherToParser . bimap ShowFmt ApiT . fromText
+    parseJSON = parseJSON >=> eitherToParser
+           . bimap ShowFmt ApiT
+           . decodePoolIdBech32
 instance ToJSON (ApiT PoolId) where
-    toJSON = toJSON . toText . getApiT
+    toJSON = toJSON . encodePoolIdBech32 . getApiT
 
 instance FromJSON ApiWalletDelegationStatus where
     parseJSON = genericParseJSON defaultSumTypeOptions
@@ -1453,12 +1457,16 @@ instance FromHttpApiData ApiPoolId where
         | t == "*" =
             Right ApiPoolIdPlaceholder
         | otherwise =
-            bimap pretty ApiPoolId (fromText t)
+            ApiPoolId <$> case fromText t of
+                Left _ ->
+                    left (T.pack . show . ShowFmt) $ decodePoolIdBech32 t
+                Right r ->
+                    Right r
 
 instance ToHttpApiData ApiPoolId where
     toUrlPiece = \case
         ApiPoolIdPlaceholder -> "*"
-        ApiPoolId pid -> toText pid
+        ApiPoolId pid -> encodePoolIdBech32 pid
 
 {-------------------------------------------------------------------------------
                                 Aeson Options
--- a/lib/core/src/Cardano/Wallet/Primitive/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types.hs
@@ -130,6 +130,8 @@ module Cardano.Wallet.Primitive.Types
     , PoolOwner(..)
     , StakeDistribution (..)
     , poolIdBytesLength
+    , decodePoolIdBech32
+    , encodePoolIdBech32
     , StakePoolMetadata (..)
     , StakePoolMetadataHash (..)
     , StakePoolMetadataUrl (..)
@@ -692,6 +694,28 @@ instance FromText PoolId where
             [ "Invalid stake pool id: expecting a hex-encoded value that is"
             , intercalate " or " (show <$> poolIdBytesLength)
             , "bytes in length."
+            ]
+
+-- | Encode 'PoolId' as Bech32 with "pool" hrp.
+encodePoolIdBech32 :: PoolId -> T.Text
+encodePoolIdBech32 =
+    Bech32.encodeLenient hrp
+        . Bech32.dataPartFromBytes
+        . getPoolId
+  where
+    hrp = [Bech32.humanReadablePart|pool|]
+
+-- | Decode a Bech32 encoded 'PoolId'.
+decodePoolIdBech32 :: T.Text -> Either TextDecodingError PoolId
+decodePoolIdBech32 t =
+    case fmap Bech32.dataPartToBytes <$> Bech32.decodeLenient t of
+        Left _ -> Left textDecodingError
+        Right (_, Just bytes) ->
+            Right $ PoolId bytes
+        Right _ -> Left textDecodingError
+      where
+        textDecodingError = TextDecodingError $ unwords
+            [ "Invalid stake pool id: expecting a Bech32 encoded value with human readable part of 'pool'."
             ]
 
 -- | A stake pool owner, which is a public key encoded in bech32 with prefix
--- a/lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
@@ -489,15 +489,17 @@ spec = do
             |] `shouldBe` (Left @String @(AddressAmount (ApiT Address, Proxy ('Testnet 0))) msg)
 
         it "ApiT PoolId" $ do
-            let msg = "Error in $: Invalid stake pool id: expecting a \
-                      \hex-encoded value that is 28 or 32 bytes in length."
+            let msg =
+                    "Error in $: Invalid stake pool id: expecting a Bech32 \
+                    \encoded value with human readable part of 'pool'."
             Aeson.parseEither parseJSON [aesonQQ|
                 "invalid-id"
             |] `shouldBe` (Left @String @(ApiT PoolId) msg)
 
         it "ApiT PoolId" $ do
-            let msg = "Error in $: Invalid stake pool id: expecting a \
-                      \hex-encoded value that is 28 or 32 bytes in length."
+            let msg =
+                    "Error in $: Invalid stake pool id: expecting a Bech32 \
+                    \encoded value with human readable part of 'pool'."
             Aeson.parseEither parseJSON [aesonQQ|
                 "4c43d68b21921034519c36d2475f5adba989bb4465ec"
             |] `shouldBe` (Left @String @(ApiT PoolId) msg)
--- a/lib/core/test/unit/Cardano/Wallet/Primitive/TypesSpec.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Primitive/TypesSpec.hs
@@ -86,6 +86,8 @@ import Cardano.Wallet.Primitive.Types
     , WalletName (..)
     , balance
     , computeUtxoStatistics
+    , decodePoolIdBech32
+    , encodePoolIdBech32
     , excluding
     , isAfterRange
     , isBeforeRange
@@ -227,6 +229,10 @@ spec = do
             forM_ testAccountIdTexts $ \text ->
                 toText <$> fromText @(Hash "Account") text
                     `shouldBe` Right text
+
+        it "Can roundtrip {decode,encode}PoolIdBech32" $
+            withMaxSuccess 1000 $ property $ \(pid :: PoolId) ->
+                decodePoolIdBech32 (encodePoolIdBech32 pid) === Right pid
 
     describe "Buildable" $ do
         it "WalletId" $ do
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
@@ -926,7 +926,9 @@ _encodeStakeAddress
 _encodeStakeAddress network (W.ChimericAccount acct) =
     Bech32.encodeLenient hrp (dataPartFromBytes bytes)
   where
-    hrp = [Bech32.humanReadablePart|stake_addr|]
+    hrp = case network of
+            SL.Testnet -> [Bech32.humanReadablePart|stake_test|]
+            SL.Mainnet -> [Bech32.humanReadablePart|stake|]
     bytes = BL.toStrict $ runPut $ do
         putWord8 $ (networkIdMask .&. toNetworkId network) .|. stakeAddressPrefix
         putByteString acct
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
@@ -927,8 +927,8 @@ _encodeStakeAddress network (W.ChimericAccount acct) =
     Bech32.encodeLenient hrp (dataPartFromBytes bytes)
   where
     hrp = case network of
-            SL.Testnet -> [Bech32.humanReadablePart|stake_test|]
-            SL.Mainnet -> [Bech32.humanReadablePart|stake|]
+        SL.Testnet -> [Bech32.humanReadablePart|stake_test|]
+        SL.Mainnet -> [Bech32.humanReadablePart|stake|]
     bytes = BL.toStrict $ runPut $ do
         putWord8 $ (networkIdMask .&. toNetworkId network) .|. stakeAddressPrefix
         putByteString acct
--- a/specifications/api/swagger.yaml
+++ b/specifications/api/swagger.yaml
@@ -233,10 +233,8 @@ x-addressState: &addressState
 
 x-stakePoolId: &stakePoolId
   type: string
-  format: hex
-  minLength: 56
-  maxLength: 64
-  example: 42bf330e9653fef8a47a5aab2f2e74db8a5b947e
+  format: hex|bech32
+  example: pool1wqaz0q0zhtxlgn0ewssevn2mrtm30fgh2g7hr7z9rj5856457mm
   description: A unique identifier for the pool.
 
 x-walletAccountPubkey: &walletAccountPubkey
--- a/specifications/api/swagger.yaml
+++ b/specifications/api/swagger.yaml
@@ -218,7 +218,7 @@ x-amount: &amount
 x-stakeAddress: &stakeAddress
   type: string
   format: bech32
-  example: stake_addr1sjck9mdmfyhzvjhydcjllgj9vjvl522w0573ncustrrr2rg7h9azg4cyqd36yyd48t5ut72hgld0fg2x
+  example: stake1sjck9mdmfyhzvjhydcjllgj9vjvl522w0573ncustrrr2rg7h9azg4cyqd36yyd48t5ut72hgld0fg2x
 
 x-addressId: &addressId
   type: string
--- a/lib/core/test/data/Cardano/Wallet/Api/ApiStakePool.json
+++ b/lib/core/test/data/Cardano/Wallet/Api/ApiStakePool.json
@@ -1,324 +1,175 @@
 {
-    "seed": 8281261503581513001,
+    "seed": 1648950576562293098,
     "samples": [
         {
             "metrics": {
-                "saturation": 0.9076337525446171,
+                "saturation": 2.4871362368537793,
                 "non_myopic_member_rewards": {
-                    "quantity": 113239850011,
+                    "quantity": 300829650366,
                     "unit": "lovelace"
                 },
                 "produced_blocks": {
-                    "quantity": 19408109,
+                    "quantity": 12428833,
                     "unit": "block"
                 },
                 "relative_stake": {
-                    "quantity": 64.18,
+                    "quantity": 74.03,
                     "unit": "percent"
                 }
             },
             "retirement": {
-                "epoch_start_time": "1891-05-03T19:32:51Z",
-                "epoch_number": 10740
+                "epoch_start_time": "1878-09-25T02:48:29Z",
+                "epoch_number": 9909
             },
             "cost": {
-                "quantity": 229,
+                "quantity": 78,
                 "unit": "lovelace"
             },
             "margin": {
-                "quantity": 84.56,
+                "quantity": 8.0e-2,
                 "unit": "percent"
             },
             "pledge": {
-                "quantity": 157,
+                "quantity": 255,
                 "unit": "lovelace"
             },
             "metadata": {
-                "homepage": "𻺗\u0013d𔷒񉳷KM\u0017\u0017i\u0005\u0012q`%\u0019{\r奍򷲫𻁏28y񁃔xN3\u0015񗒂Vq𣩵m󵭨,@i\u001a\u001c4[y􂍅}j򭳤`5퇬񥄛(\u001c/\u001a񥕶4",
-                "name": "0bQ񔈡=\\𚂃m5y?򚙈􃳷US򓕇󄧨𾤔s?\u0003eXcVk򣍃򨖍K\u0015\u001a\u0016]𳧐Q\u0007񪿫\u000er",
-                "ticker": "'O",
-                "description": "a2x񘁶󬫺M^\u001f"
+                "homepage": "D\u0017\r9G𥙩\u000bn\u001dZ򨔘\u0011:1A\u001f|G\u001ee򶢹m?򼓘k74D񷖋\u0011N򐹯D􉘃󐿚񯛝\u0013A9=>",
+                "name": "\u0008𾦶\u0004&~-󫡲\u0014@󙖯i𯩁R,񾔶|#𸴟>C\u000f",
+                "ticker": "`\u0007\u001d",
+                "description": "^W'k񬈅F*k&P!C*\u0008=\u0016}𨿋Z 򛫻񲦚aY)\u0003D\u0000W𱐬󁎦񋟙t;\u000c"
             },
-            "id": "7d622efffd4990c8460a92713588f683be1701fc03adc70cd5daf07b"
+            "id": "pool1j4p8d27crkjde00p4l9v5f95hftm23x2yylllp9lzg72w6p4ecm"
         },
         {
             "metrics": {
-                "saturation": 0.43779455326814765,
+                "saturation": 4.206923248389226,
                 "non_myopic_member_rewards": {
-                    "quantity": 847860558162,
+                    "quantity": 939832251301,
                     "unit": "lovelace"
                 },
                 "produced_blocks": {
-                    "quantity": 3350529,
+                    "quantity": 22311798,
                     "unit": "block"
                 },
                 "relative_stake": {
-                    "quantity": 48.91,
+                    "quantity": 86.03,
                     "unit": "percent"
                 }
             },
             "retirement": {
-                "epoch_start_time": "1871-02-28T10:36:03.338376549539Z",
-                "epoch_number": 1210
+                "epoch_start_time": "1885-01-28T19:23:12.506992765083Z",
+                "epoch_number": 19632
             },
             "cost": {
-                "quantity": 179,
+                "quantity": 195,
                 "unit": "lovelace"
             },
             "margin": {
-                "quantity": 64.2,
+                "quantity": 25.17,
                 "unit": "percent"
             },
             "pledge": {
-                "quantity": 70,
+                "quantity": 144,
                 "unit": "lovelace"
             },
             "metadata": {
-                "homepage": "󘩺񍙱^m4Q󪀣9]G𑞣\"򻽑5񌳞\u0012\rgsj7򘩝a(sv𱙄=򠟷 ?\u00132򾯧򞨡󪍿\u000ci:d\u0017𲕼 \u0007\n\u001a#M\u00071X󮀪W𥽂󳅡,&<\u0001",
-                "name": "󹟅\u0013Mf1yU*Y-9񦵜-uB*][J\u001b򰚅񶠽\r-\u0002򥧱5e𯐼\u0004p\u001eLs8ꏡ٩y,JW򨠪W'\u0008|",
-                "ticker": "k򦧮jk"
+                "homepage": "𵍓\u0006\u001d򡗓񍹭",
+                "name": "\u0011&񐔒񕜳\u001a𻧩񣕣Q񙡺󞖋򝷙W񆫽:\u001c򥷪4#",
+                "ticker": "󵬘N;",
+                "description": "<S2G𨿠ꕏi043\u001dU􊨙񨚤>P&u~y򎛊\u0011qDg\u0003\u000e\u0006D^r򃚦\u000f򔇰h񷩙񍓧񓫢:>k񼝛򡜣!k\u000c\u0008x\rs3􃞧d\u0001xvA,W`󩰫󧀿\u001br򓚁񂥳g{\u001cW\u0005\u0002򄗭񄢩@\u001fLEX哠N0 #\u0011\u00082\r#D򘵻򪬏\u0008]򞳭z񞥃񰦑c󗎟򕗀'\u0013\u001b򔌏%8]򱐄6N\n\u0018n#j񜉅:kxX\u0010@yDq\u0011󂻎g򾾡c\u00004򗙧𾤢y󛷥K\\?Ik^Wf^𝽠_򝗆kpuaCz\u001a󖀃\u0016O񁕓9\u000c󮩹\u000bIQ\u0006P\r>f\u00179\u0015!/~u򁽟{)#\u001a\u0011W󧔾?\u00048d&f𚉐(򻙔l'Z\u0015򏼘i[R&𫟾"
             },
-            "id": "993c338cc4c6786d93b8b2e785de803151103646ca2e170c2b5d8cde"
+            "id": "pool1y4zccfz0vrvmt8e96gfn5u5pd0zln5fn2uup25fk8grfw4frwqh"
         },
         {
             "metrics": {
-                "saturation": 2.7648182239544923,
+                "saturation": 2.0262672273589573,
                 "non_myopic_member_rewards": {
-                    "quantity": 851260348936,
+                    "quantity": 688229616774,
                     "unit": "lovelace"
                 },
                 "produced_blocks": {
-                    "quantity": 20522528,
+                    "quantity": 11174627,
                     "unit": "block"
                 },
                 "relative_stake": {
-                    "quantity": 61.06,
+                    "quantity": 1.75,
                     "unit": "percent"
                 }
             },
             "retirement": {
-                "epoch_start_time": "1876-06-30T07:00:00Z",
-                "epoch_number": 31278
+                "epoch_start_time": "1908-07-23T01:18:41.264047239015Z",
+                "epoch_number": 14875
             },
             "cost": {
-                "quantity": 133,
+                "quantity": 175,
                 "unit": "lovelace"
             },
             "margin": {
-                "quantity": 4.29,
+                "quantity": 6.67,
                 "unit": "percent"
             },
             "pledge": {
-                "quantity": 54,
+                "quantity": 230,
                 "unit": "lovelace"
             },
             "metadata": {
-                "homepage": "\u000b񩳽s񍏩`;",
-                "name": "\u000c\u0013\u001b_򐂍\u001aQ󤽢~󖂧v\u000b\"M򉭺𵩣\u0006󰻷0򐻣\"󫣖\n$\u0004񾉽V\u0006񑩳Q\u0003\nd\u0014򩛯\u0015\n-1\u0008򭗬\u0008!",
-                "ticker": "*\u0004w\u001dP",
-                "description": "yRSt\u0006r\u0005󩼤\\&n\u0017\u000b𚻖\u0010\u0018xajxf򫃬d-,\u0013D\u0011􌃳򓟎K𫶡\u001bU狤𾾈m\u0010󤍮򿳂~򘫄\\N5􂢳t\u000b\u0016󝉹\u000c򤌺;:&򢕨4l6\u001d\u0010𵶉bX򃶜!)\u001cuj\u00002Ub[A\u0016@>\u0001\u001f`𻽽\n!򦹿㒂𢦴\u0004񯑑򕎹z\"S3󇾄u񲆖񄇋+p𫾓𲮥M2A򍙙@RTPi򙑣'\u001d;W.Q񤌺H𙽲JH,[-RxC\n6򾑽ty]w\u0015򯦑l<e󞁒󅪷]j\u0013\u0000􄯖!{뉈/񀛨0򱯋0/li~;򶌩򬧲[H\u000b󻮪"
+                "homepage": " 𹙢'Y𴛤-󉡙\u0000\u0016\u0002wm񅨚PG򣁢󡿡Z􇵂\u000e񙄟\u0010jB󞭖􁄽y'$\u001f5񲒖𮡧\u0007^򎴁\u00009sy􏂦W􉑤LNfiQ󉜴%\\Rb\u0001!+򿋳\u001d\u0015l񔺑񃨍D\"𼢴篸\u001b)s\u0012>?",
+                "name": "\u001bL𘜚H\r񳭨\"mC~?U\u0002dvh${rr&)\u0017򦤘񖙢\n򥺪i\u0008\u0006􄓄]1_𹍘\u000b񁾊IfK\"󖈷𳫴򾝢",
+                "ticker": "=us󬟶򷬣",
+                "description": "~`]\u0017򂃱^\u001bIr\u0000,\u0005\u0013m򂋲/󱞉P\u0005\u0005\u0018\u0007jL"
             },
-            "id": "7aac84197185537648f812b85ce9f17005d633e15139763dcdc015f9"
+            "id": "pool1wv5uzfqqrdrns4aprvh4ta8egwqavru05z77024k7h0p6dudkvd"
         },
         {
             "metrics": {
-                "saturation": 3.5908012914491927,
+                "saturation": 2.118516160766028,
                 "non_myopic_member_rewards": {
-                    "quantity": 903252358489,
+                    "quantity": 870783207562,
                     "unit": "lovelace"
                 },
                 "produced_blocks": {
-                    "quantity": 697382,
+                    "quantity": 4904922,
                     "unit": "block"
                 },
                 "relative_stake": {
-                    "quantity": 75.84,
+                    "quantity": 75.06,
                     "unit": "percent"
                 }
             },
             "retirement": {
-                "epoch_start_time": "1893-04-23T12:54:37Z",
-                "epoch_number": 5693
+                "epoch_start_time": "1903-04-06T02:37:17.892934980153Z",
+                "epoch_number": 32025
             },
             "cost": {
-                "quantity": 75,
+                "quantity": 101,
                 "unit": "lovelace"
             },
             "margin": {
-                "quantity": 62.11,
+                "quantity": 2.07,
                 "unit": "percent"
             },
             "pledge": {
-                "quantity": 68,
+                "quantity": 69,
                 "unit": "lovelace"
             },
             "metadata": {
-                "homepage": "񈆉&T􀺦kB쎧",
-                "name": "\u0011GCw:𽆨2;D񅧎@H򜾏𵽛C򷚌~聠u<L񐍄m?\u001e𜈣@\u0010-",
-                "ticker": "𱨕򻋫%񉓀9",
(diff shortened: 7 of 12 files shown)


lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1096,9 +1100,11 @@ instance ToJSON ApiByronWalletBalance where
     toJSON = genericToJSON defaultRecordTypeOptions
 
 instance FromJSON (ApiT PoolId) where
-    parseJSON = parseJSON >=> eitherToParser . bimap ShowFmt ApiT . fromText
+    parseJSON j = legacyParse j <|> (parseJSON >=> eitherToParser . bimap (ShowFmt . displayException) ApiT . decodePoolIdBech32 $ j)
+      where
+        legacyParse = parseJSON >=> eitherToParser . bimap ShowFmt ApiT . fromText
```
jonathanknowles: Hi @hasufell!

This section of code is quite a bit longer than our [line length limit](https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#limit-line-length-to-80-characters) (80 chars).

In some situations (e.g., test data with long strings, unbreakable URLs in comments, and so on), exceeding the limit is unavoidable, but outside of these circumstances, we generally try to adhere to the limit. (Though it is not strictly enforced.)

**Request**: rewrite this section to fit within the limit.

**Suggestion**: How about the following:

**Advantage:** both the "legacy" and "normal" parsers now have names, which makes the intent of the code more obvious.

```suggestion
    parseJSON j = legacyParse j <|> normalParse j
      where
        legacyParse = parseJSON >=> eitherToParser
            . bimap ShowFmt ApiT
            . fromText
        normalParse = parseJSON >=> eitherToParser
            . bimap (ShowFmt . displayException) ApiT
            . decodePoolIdBech32
```

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1453,12 +1459,14 @@ instance FromHttpApiData ApiPoolId where
         | t == "*" =
             Right ApiPoolIdPlaceholder
         | otherwise =
-            bimap pretty ApiPoolId (fromText t)
+            case fmap ApiPoolId (fromText t) of
+              Left _  -> bimap (T.pack . displayException) ApiPoolId . decodePoolIdBech32 $ t
+              Right r -> Right r
```
jonathanknowles: Two points:

1. We generally indent code blocks with 4 spaces, rather than 2, although oddly enough this doesn't appear to be explicitly documented anywhere. We should probably add this to our [coding standards](https://github.com/input-output-hk/adrestia/wiki/Coding-Standards).

2. We have a rule to [avoid variable-length indentation](https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#avoid-variable-length-indentation).

**Suggestion:**

```suggestion
            ApiPoolId <$> case fromText t of
                Left _ ->
                    left (T.pack . displayException) $ decodePoolIdBech32 t
                Right r ->
                    Right r
```

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -694,6 +698,28 @@ instance FromText PoolId where
...
+decodePoolIdBech32 t =
+    case fmap Bech32.dataPartToBytes <$> Bech32.decodeLenient t of
+        Left _ -> throwM textDecodingError
+        Right (_, Just bytes) ->
+            pure $ PoolId bytes
+        Right _ -> throwM textDecodingError
+      where
+        textDecodingError = TextDecodingError $ unwords
+            [ "Invalid stake pool id: expecting a Bech32 encoded value with human readable part of 'pool'."
+            ]
```
jonathanknowles, line 719: **Suggestion:** Report the invalid value to assist with debugging.

```suggestion
        textDecodingError = TextDecodingError $ unwords   
            [ "Invalid stake pool id:"
            , show t  
            , "." 
            , "Expecting a Bech32-encoded value with prefix 'pool'."  
            ]
```

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -489,15 +489,13 @@ spec = do
             |] `shouldBe` (Left @String @(AddressAmount (ApiT Address, Proxy ('Testnet 0))) msg)
 
         it "ApiT PoolId" $ do
-            let msg = "Error in $: Invalid stake pool id: expecting a \
-                      \hex-encoded value that is 28 or 32 bytes in length."
+            let msg = "Error in $: Invalid stake pool id: expecting a Bech32 encoded value with human readable part of 'pool'."
```
jonathanknowles: **Request**: break up overly-long lines (see https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#limit-line-length-to-80-characters) with the line continuation character.

This helps with readability when reading two sources side-by-side.

```suggestion
            let msg = 
                    "Error in $: Invalid stake pool id: expecting a Bech32 \  
                    \encoded value with prefix 'pool'."
```

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -489,15 +489,13 @@ spec = do
...
             Aeson.parseEither parseJSON [aesonQQ|
                 "invalid-id"
             |] `shouldBe` (Left @String @(ApiT PoolId) msg)
 
         it "ApiT PoolId" $ do
-            let msg = "Error in $: Invalid stake pool id: expecting a \
-                      \hex-encoded value that is 28 or 32 bytes in length."
+            let msg = "Error in $: Invalid stake pool id: expecting a Bech32 encoded value with human readable part of 'pool'."
```
jonathanknowles: Similar to above:

```suggestion
            let msg = 
                    "Error in $: Invalid stake pool id: expecting a Bech32 \  
                    \encoded value with prefix 'pool'."
```

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -925,7 +925,9 @@ _encodeStakeAddress
 _encodeStakeAddress network (W.ChimericAccount acct) =
     Bech32.encodeLenient hrp (dataPartFromBytes bytes)
   where
-    hrp = [Bech32.humanReadablePart|stake_addr|]
+    hrp = case network of
+            SL.Testnet -> [Bech32.humanReadablePart|stake_test|]
+            SL.Mainnet -> [Bech32.humanReadablePart|stake|]
```
jonathanknowles: **Request**: use 4-space indents, rather than 8 spaces.

**Justification:** Our [coding standard](https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#editor-configuration-via-editorconfig) specifies 4 spaces for Haskell code via the [`.editorconfig`](https://github.com/input-output-hk/cardano-wallet/blob/fece720cb18d9676c4f1f6b0d3a69a37fab93f1c/.editorconfig#L16) file.

```suggestion
    hrp = case network of
        SL.Testnet -> [Bech32.humanReadablePart|stake_test|]
        SL.Mainnet -> [Bech32.humanReadablePart|stake|]
```

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -694,6 +698,28 @@ instance FromText PoolId where
...
             ]
 
+-- | Encode 'PoolId' as Bech32 with "pool" hrp.
+encodePoolIdBech32 :: PoolId -> T.Text
+encodePoolIdBech32 =
+    Bech32.encodeLenient hrp
+        . Bech32.dataPartFromBytes
+        . getPoolId
+      where
+        hrp = [Bech32.humanReadablePart|pool|]
```
jonathanknowles:
```suggestion
  where
    hrp = [Bech32.humanReadablePart|pool|]
```

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -694,6 +698,28 @@ instance FromText PoolId where
             , "bytes in length."
             ]
 
+-- | Encode 'PoolId' as Bech32 with "pool" hrp.
+encodePoolIdBech32 :: PoolId -> T.Text
+encodePoolIdBech32 =
+    Bech32.encodeLenient hrp
+        . Bech32.dataPartFromBytes
+        . getPoolId
```
jonathanknowles, line 704: :+1:

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -694,6 +698,28 @@ instance FromText PoolId where
...
+      where
+        hrp = [Bech32.humanReadablePart|pool|]
+
+-- | Decode a Bech32 encoded 'PoolId'. Throws 'TextDecodingError' on failure.
+decodePoolIdBech32 :: MonadThrow m => T.Text -> m PoolId
```
jonathanknowles: **Question:** is there an advantage to using `MonadThrow` over `Either` here?

**Suggestion:** use `Either`

```hs
-- | Decode a Bech32-encoded 'PoolId'.
decodePoolIdBech32 :: T.Text -> Either TextDecodingError PoolId
```
**Advantages**:
* the type of failure is made explicit in the type signature
* this type signature is consistent with other existing decoders (for example `hashFromText`).
* the comment can be simpler.

However, I do wonder if there is some advantage to using `MonadThrow` instead?

jonathanknowles (changes requested):
Hi @hasufell.

Looks pretty good so far!

General comments:

1. I'm not sure whether using `MonadThrow` in `decodePoolIdBech32`  gives us much advantage over `Either TextDecodingError`. It's harder for callers to see how the function will fail, as the type signature no longer includes `TextDecodingError`. But perhaps you have a good reason for using `MonadThrow` here?

2. I couldn't find a roundtrip encoding/decoding test specifically for `{decode,encode}PoolIdBech32`. Did you intend to write one? I think it would be worth adding such a test. For inspiration, have a look at the various `textRoundtrip` tests.

3. There are a few small issues with coding style. It's worth taking a look at our [coding standards](https://github.com/input-output-hk/adrestia/wiki/Coding-Standards). These should hopefully be easy to fix. :+1:

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1096,9 +1100,11 @@ instance ToJSON ApiByronWalletBalance where
     toJSON = genericToJSON defaultRecordTypeOptions
 
 instance FromJSON (ApiT PoolId) where
-    parseJSON = parseJSON >=> eitherToParser . bimap ShowFmt ApiT . fromText
+    parseJSON j = legacyParse j <|> (parseJSON >=> eitherToParser . bimap (ShowFmt . displayException) ApiT . decodePoolIdBech32 $ j)
```
KtorZ: Note that the `FromJSON` instance is only really used for testing here, there's no API endpoint which expects a pool id in a JSON body. So the backward compatibility here is unnecessary. 

Users may however provide pool ids as part of the path parameters relying on `FromHttpApiData`; this needs to be updated to accept bech32 while indeed keeping support for base16.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1453,12 +1459,14 @@ instance FromHttpApiData ApiPoolId where
...
-            bimap pretty ApiPoolId (fromText t)
+            case fmap ApiPoolId (fromText t) of
+              Left _  -> bimap (T.pack . displayException) ApiPoolId . decodePoolIdBech32 $ t
+              Right r -> Right r
 
 instance ToHttpApiData ApiPoolId where
     toUrlPiece = \case
         ApiPoolIdPlaceholder -> "*"
-        ApiPoolId pid -> toText pid
+        ApiPoolId pid -> encodePoolIdBech32 pid
```
KtorZ, line 1469: Make sure to also fix the associated `FromHttpApiData` as well!

hasufell: That is part of this PR already https://github.com/input-output-hk/cardano-wallet/pull/2055/files#diff-e1113f15531f8432f00f243428125671R1457-R1464

lib/core/src/Cardano/Wallet/Primitive/Types.hs
hasufell: > Question: is there an advantage to using MonadThrow over Either here?

MonadThrow specialises to Either, Maybe and IOException in IO, so has a richer API. The downside is that `e ~ SomeException`. It is easy to change it, but that was my decision so far.

hasufell:
> There are a few small issues with coding style. It's worth taking a look at our coding standards. These should hopefully be easy to fix. +1

Could you be more specific?

lib/core/src/Cardano/Wallet/Api/Types.hs
hasufell: > This section of code is quite a bit longer than our line length limit (80 chars).

This should be part of the formatter. Is there any reason it is not?

lib/core/src/Cardano/Wallet/Primitive/Types.hs
hasufell: I think last time I wanted to show an invalid value @KtorZ said not to, because it could be malicious input?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1453,12 +1464,16 @@ instance FromHttpApiData ApiPoolId where
         | t == "*" =
             Right ApiPoolIdPlaceholder
         | otherwise =
-            bimap pretty ApiPoolId (fromText t)
+            ApiPoolId <$> case fromText t of
+                Left _ ->
+                    left (T.pack . displayException) $ decodePoolIdBech32 t
```
KtorZ: I'd be best to specialize to `Either TextDecodingError` here and use `ShowFmt` as for other error renderer here. Otherwise, I am afraid that the `displayException` will default to `Show` and renders the error tainted with Haskell syntax instead of something more human-readable for the API.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
KtorZ: And I still maintain this position. We should NOT show invalid values, but instead, show examples of valid values. This should actually be part of our coding standard, I'll start a voting thread on Slack.

lib/core/src/Cardano/Wallet/Api/Types.hs
hasufell: In this case, I wrote a non-Show instance for Exception though

jonathanknowles:
> Could you be more specific?

Sure! I was referring to the comments on the review. :+1: 

AFAIK, all the issues are mentioned in these comments.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
jonathanknowles: >  We should NOT show invalid values, but instead, show examples of valid values.

I think that is definitely a fair position to take, depending on our threat model.

Though if we do decide on such a policy, perhaps we should also audit the code to see if there are existing violations, and take appropriate action. (I suspect we do print out invalid values in other places.)

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1096,9 +1098,11 @@ instance ToJSON ApiByronWalletBalance where
     toJSON = genericToJSON defaultRecordTypeOptions
 
 instance FromJSON (ApiT PoolId) where
-    parseJSON = parseJSON >=> eitherToParser . bimap ShowFmt ApiT . fromText
+    parseJSON = parseJSON >=> eitherToParser
+           . bimap ShowFmt ApiT
+           . decodePoolIdBech32
```
jonathanknowles, line 1103: Request: use 4-space indents.

See https://github.com/input-output-hk/cardano-wallet/pull/2055#discussion_r475345235

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -694,6 +696,28 @@ instance FromText PoolId where
...
+decodePoolIdBech32 :: T.Text -> Either TextDecodingError PoolId
+decodePoolIdBech32 t =
+    case fmap Bech32.dataPartToBytes <$> Bech32.decodeLenient t of
+        Left _ -> Left textDecodingError
+        Right (_, Just bytes) ->
+            Right $ PoolId bytes
+        Right _ -> Left textDecodingError
+      where
+        textDecodingError = TextDecodingError $ unwords
+            [ "Invalid stake pool id: expecting a Bech32 encoded value with human readable part of 'pool'."
```
jonathanknowles, line 718: **Request**: wrap this to meet our [line length limit](https://github.com/input-output-hk/cardano-wallet/wiki/Coding-Standards#limit-line-length-to-80-characters).

(I'm assuming that we've decided not to report the malformed input back to the user in this case.)
```suggestion
            [ "Invalid stake pool id: expecting a Bech32 encoded value "
            , "with human readable part of 'pool'."

```

hasufell:
Wrt the [build failure](https://hydra.iohk.io/build/3913656/nixlog/1)... I cannot reproduce it.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -694,6 +696,28 @@ instance FromText PoolId where
             , "bytes in length."
             ]
 
+-- | Encode 'PoolId' as Bech32 with "pool" hrp.
+encodePoolIdBech32 :: PoolId -> T.Text
```
jonathanknowles, line 700: **Request:** could you add a roundtrip encoding/decoding test for `encodePoolIdBech32` and `decodePoolIdBech32` (see https://github.com/input-output-hk/cardano-wallet/pull/2055#pullrequestreview-473111599).

I'm guessing that the following property should hold:
```
decodePoolIdBech32 (encodePoolIdBech32 p) == Right p
```

We do have similar tests from which you can draw inspiration: have a search for `textRoundtrip`.

Feel free to ask if I can be of any assistance!

jonathanknowles (changes requested):
Hi @hasufell 

Thanks for your changes. I see that you've converted `decodePoolIdBech32` from using `MonadThrow` to returning an ordinary `Either` value, which I think makes the failure modes a bit clearer.

There are still a couple of issues I think we should fix:

1. **Roundtrip encoding tests:** as mentioned in the [previous review](https://github.com/input-output-hk/cardano-wallet/pull/2055#pullrequestreview-473111599), I couldn't find a roundtrip property test for `decodePoolIdBech32` and `encodePoolIdBech32`. Would it be okay to add one?
    
    I'm guessing that the following property should hold:
    ```hs
    decodePoolIdBech32 (encodePoolIdBech32 p) == Right p
    ```
    Note that we do have similar tests from which you can draw inspiration: have a search for `textRoundtrip`, and lots of stuff should pop up. Feel free to ask me if you need any assistance.

2. **Formatting:** There are still a couple of formatting issues to sort out (see review comments).

lib/core/src/Cardano/Wallet/Api/Types.hs
hasufell: Please provide a formatter or an editor config that works. Otherwise these issues will pop up over and over again. I'm not counting spaces while I code.

hasufell:
>     1. As mentioned in the [previous review](https://github.com/input-output-hk/cardano-wallet/pull/2055#pullrequestreview-473111599), I couldn't find a roundtrip property test for `decodePoolIdBech32` and `encodePoolIdBech32`. Would it be okay to add one?

That has been done: https://github.com/input-output-hk/cardano-wallet/pull/2055/commits/b52ee3ffd88efcd72a6e695d2c303998e9db8a65

lib/core/src/Cardano/Wallet/Api/Types.hs
jonathanknowles: @hasufell wrote:
> Please provide a formatter or an editor config that works. Otherwise these issues will pop up over and over again. 

**Regarding automated formatting:** Unfortunately our team doesn't use an automated formatter besides [stylish-haskell](https://github.com/jaspervdj/stylish-haskell) (for imports), but if you feel strongly that we should do, then you could start a discussion in our team channel, and try to convince the team of the benefits. :+1: 

**Regarding an editor config:** We do provide one [here](https://github.com/input-output-hk/cardano-wallet/blob/master/.editorconfig).

@hasufell wrote:
> I'm not counting spaces while I code.

Our [team coding standard](https://github.com/input-output-hk/adrestia/wiki/Coding-Standards) is actually rather simple in this respect: we just indent each level by 4 spaces. (We make an exception for `where` clauses, which are indented by 2 spaces.)

Are you facing an issue that makes it hard to indent in this way? If so, we can discuss ways to fix the issue.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -694,6 +696,30 @@ instance FromText PoolId where
...
+    case fmap Bech32.dataPartToBytes <$> Bech32.decodeLenient t of
+        Left _ -> Left textDecodingError
+        Right (_, Just bytes) ->
+            Right $ PoolId bytes
+        Right _ -> Left textDecodingError
+      where
+        textDecodingError = TextDecodingError $ unwords
+            [ "Invalid stake pool id: expecting a Bech32 encoded value "
+            , "with human readable part of 'pool'."
+
```
jonathanknowles:
```suggestion
```

specifications/api/swagger.yaml
```diff
@@ -233,10 +233,8 @@ x-addressState: &addressState
 
 x-stakePoolId: &stakePoolId
   type: string
-  format: hex
-  minLength: 56
-  maxLength: 64
```
jonathanknowles, line 238: Is there a minimum and maximum length for an encoded stake pool ID string?

lib/core/src/Cardano/Wallet/Api/Types.hs
hasufell: > Regarding automated formatting: Unfortunately our team doesn't use an automated formatter besides stylish-haskell (for imports), but if you feel strongly that we should do, then you could start a discussion in our team channel, and try to convince the team of the benefits. +1

I don't have a strong opinion, except that I avoid formatting discussions as much as possible. I'm just saying these things will pop up again if the tooling (formatter or config) doesn't work consistently. I don't think about indentation while coding and I am not counting spaces while doing so.

I also believe developers time is too valuable to count spaces in pull requests (your time and my time). I'll do whatever formatting you want, as long as I'm provided with the tools to do so.

> Regarding an editor config: We do provide one here.

I'm using that one and it seems it doesn't work consistently. Help appreciated. Here is my editor config https://gogs.hasufell.de/hasufell/vim-config/src/branch/workman

> Are you facing an issue that makes it hard to indent in this way? If so, we can discuss ways to fix the issue.

See above.

specifications/api/swagger.yaml
hasufell: The other bech32 types don't have it, afais.

lib/core/src/Cardano/Wallet/Api/Types.hs
jonathanknowles: @hasufell wrote:
> I'll do whatever formatting you want, as long as I'm provided with the tools to do so.

I'm sorry, I don't have an automated tool to provide you with, otherwise I would. As I mentioned before, if there is a tool that you think we could use, then feel free to discuss it in the team channel. I would happy to help with the discussion!

> I don't think about indentation while coding and I am not counting spaces while doing so.
> I also believe developers time is too valuable to count spaces in pull requests (your time and my time). 

The reason for having a team standard is so that we don't have to waste time during PR reviews debating code style. If we have to discuss this in every PR, it would certainly waste time.

As far as I know, most people in our team just configure their tab key to emit 4 spaces. Then no counting is required. Would that work for you?

KtorZ: @hasufell I had a look at your vim config and... it's .. a lot. Looks quite complicated and there's probably some plugin that you have which is overriding the editorconfig plugin. 

I could share mine if you want but I am not sure it's going to help you much. 

I do agree that arguing about white-spaces over a PR is pointless, but it should just not happened that code is wrongly indented because your editor should be configured to pick up the right config. 
So far, we've all been using that `.editorconfig` without any issues and with various editors (vim, emacs and sublime at least). I can confirm it works perfectly fine with vim 7 & 8.

hasufell: > Looks quite complicated and there's probably some plugin that you have which is overriding the editorconfig plugin.

It is not overridden, because when I press tab it indents 4 spaces. It's possible that there is diverging behavior on pressing `Enter`. I expect an editor to do the right thing when I press enter and not have to manually indent every line I type. I'm not sure what the correct configuration for that is though.

----
Edit:
Possible source: https://github.com/neovimhaskell/haskell-vim#indentation

So, can someone say what the indentation policy for those various constructs is?

jonathanknowles: @hasufell wrote:
> So, can someone say what the indentation policy for those various constructs is?

Which constructs?

jonathanknowles: @hasufell Just in case there's any doubt, we generally indent code with:

- 4 spaces per level of nesting.
- 2 spaces for a `where` clause.

**Example:**
```hs
launchMissiles trace a b c = do
    missiles <- prepareMissiles
    forM_ missiles $ \missile -> do
        log trace Warning $ MsgMissileLaunch missile        
        launchMissile missile
    pure missiles    
  where
    launchMissile m =
        inializeLauncherModule m
        lightFuse m
    prepareMissiles = do
        missiles <- enumerateMissiles
        verifyMissilesIntact missiles
        pure missiles
```

hasufell: > Which constructs?

As was in my link to the haskell-vim readme.

So my configuration for haskell-vim now is

```vim
let g:haskell_indent_if = 0
let g:haskell_indent_case = 4
let g:haskell_indent_let = 4
let g:haskell_indent_where = 6
let g:haskell_indent_before_where = 2
let g:haskell_indent_after_bare_where = 2
let g:haskell_indent_do = 4
let g:haskell_indent_in = 0
let g:haskell_indent_guard = 2
```

This doesn't play nicely with `do` I believe. See below.

----

* do

```hs
foo :: IO Int
foo = do
    x <- pure 1
    do
        pure (2 + x)
```

* if

```hs
foo :: IO Int
foo =
    if True
    then pure 3
    else pure 2
```

* case

```hs
foo :: IO Int
foo =
    case True of
        True -> 3
        False -> 2
```

* let

```hs
foo :: IO Int
foo =
    let x = 3 in
        x + 1

foo :: IO Int
foo =
    let x = 1
    in x + 1
```

* guard

```hs
foo :: IO Int
foo
  | otherwise = 2
```

* where

```hs
foo :: IO Int
foo = x + 1
  where
    x = 3
```

So please either confirm this configuration or suggest an alternative.

lib/core/test/unit/Cardano/Wallet/Primitive/TypesSpec.hs
```diff
@@ -228,6 +230,14 @@ spec = do
                 toText <$> fromText @(Hash "Account") text
                     `shouldBe` Right text
 
+        it "Can roundtrip {decode,encode}PoolIdBech32" $
+            withMaxSuccess 1000 $ property $ \(pid :: PoolId) ->
+                let enc = encodePoolIdBech32 pid
+                    dec = decodePoolIdBech32 enc
+                in case dec of
+                     Left _ -> False
+                     Right dec' -> dec' == pid
```
jonathanknowles: 4-space indents:

```suggestion
                in case dec of
                    Left _ -> False
                    Right dec' -> dec' == pid
```

lib/core/src/Cardano/Wallet/Api/Types.hs
jonathanknowles, line 1103: @hasufell I'm happy to continue the discussion of editor configs in our slack channel.

Though for this PR, there is only [this request](https://github.com/input-output-hk/cardano-wallet/pull/2055#discussion_r476285956):

```suggestion
    parseJSON = parseJSON >=> eitherToParser
        . bimap ShowFmt ApiT
        . decodePoolIdBech32
```
and
https://github.com/input-output-hk/cardano-wallet/pull/2055/files#r477206873

If you can wrap these up, then we can merge this PR immediately. :+1:

KtorZ: I don't see any problem with `do` above ? Also, guards should be 4-space indented like the rest

```
let g:haskell_indent_guard = 4
```

hasufell: The problem with do is if you write something on the same line as do:

```hs
foo :: IO Int
foo = do
    x <- pure 1
    do y <- pure 2
        pure (x + y)
```

This will cause a compile error.

jonathanknowles (approved):
LGTM :+1:

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -925,7 +925,9 @@ _encodeStakeAddress
 _encodeStakeAddress network (W.ChimericAccount acct) =
     Bech32.encodeLenient hrp (dataPartFromBytes bytes)
   where
-    hrp = [Bech32.humanReadablePart|stake_addr|]
+    hrp = case network of
+        SL.Testnet -> [Bech32.humanReadablePart|stake_test|]
+        SL.Mainnet -> [Bech32.humanReadablePart|stake|]
```
KtorZ, line 931: :+1:

lib/core/test/unit/Cardano/Wallet/Primitive/TypesSpec.hs
```diff
@@ -228,6 +230,14 @@ spec = do
                 toText <$> fromText @(Hash "Account") text
                     `shouldBe` Right text
 
+        it "Can roundtrip {decode,encode}PoolIdBech32" $
+            withMaxSuccess 1000 $ property $ \(pid :: PoolId) ->
+                let enc = encodePoolIdBech32 pid
+                    dec = decodePoolIdBech32 enc
+                in case dec of
+                    Left _ -> False
+                    Right dec' -> dec' == pid
```
KtorZ: `===` would be better here to get a proper counter example in case of failure. 

```hs
decodePoolIdBech32 (encodePoolIdBech32 pid) === Right pid
```

KtorZ (approved):

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: That is arguably a horrible choice of formatting and It's perhaps much better that the compiler fails to parse this :)

hasufell:
~~Can't reproduce the build failure, even with the given seed...~~

KtorZ (approved):

KtorZ (changes requested):
actually, on second thoughts. I didn't see any tests that shows that:

a) hex-encoded pool ids are still accepted in path parameters
b) bech32 encoded pool ids are accepted in path parameters

hasufell:
>
> a) hex-encoded pool ids are still accepted in path parameters
> 

These are the tests that fail if hex encoding for path is disabled:


<details><summary>Failing tests</summary>
<p>

```
Failures:

  test/unit/Cardano/Wallet/ApiSpec.hs:137:42:
  1) Cardano.Wallet.Api, Malformed PathParam, (ApiT WalletId) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/śśśśśśśśśśśśśśśśśśśśśśśśśśśśśśśśśśśśśśśś
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected response body \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"wallet id should be a hex-encoded string of 40 characters\\\"}\", but received \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"Invalid stake pool id: expecting a Bech32 encoded value with human readable part of 'pool'.\\\"}\""

  To rerun use: --match "/Cardano.Wallet.Api/Malformed PathParam/(ApiT WalletId) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347\347/"

  test/unit/Cardano/Wallet/ApiSpec.hs:137:42:
  2) Cardano.Wallet.Api, Malformed PathParam, (ApiT WalletId) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/111111111111111111111111111111111111111
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected response body \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"wallet id should be a hex-encoded string of 40 characters\\\"}\", but received \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"Invalid stake pool id: expecting a Bech32 encoded value with human readable part of 'pool'.\\\"}\""

  To rerun use: --match "/Cardano.Wallet.Api/Malformed PathParam/(ApiT WalletId) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/111111111111111111111111111111111111111/"

  test/unit/Cardano/Wallet/ApiSpec.hs:137:42:
  3) Cardano.Wallet.Api, Malformed PathParam, (ApiT WalletId) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/11111111111111111111111111111111111111111
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected response body \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"wallet id should be a hex-encoded string of 40 characters\\\"}\", but received \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"Invalid stake pool id: expecting a Bech32 encoded value with human readable part of 'pool'.\\\"}\""

  To rerun use: --match "/Cardano.Wallet.Api/Malformed PathParam/(ApiT WalletId) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/11111111111111111111111111111111111111111/"

  test/unit/Cardano/Wallet/ApiSpec.hs:142:42:
  4) Cardano.Wallet.Api, Malformed BodyParam, (ApiWalletPassphrase) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected response body \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"Error in $.passphrase: passphrase is too long: expected at most 255 characters\\\"}\", but received \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"Invalid stake pool id: expecting a Bech32 encoded value with human readable part of 'pool'.\\\"}\""

  To rerun use: --match "/Cardano.Wallet.Api/Malformed BodyParam/(ApiWalletPassphrase) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000/"

  test/unit/Cardano/Wallet/ApiSpec.hs:142:42:
  5) Cardano.Wallet.Api, Malformed BodyParam, (ApiWalletPassphrase) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected response body \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"Error in $.passphrase: parsing Text failed, expected String, but encountered Number\\\"}\", but received \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"Invalid stake pool id: expecting a Bech32 encoded value with human readable part of 'pool'.\\\"}\""

  To rerun use: --match "/Cardano.Wallet.Api/Malformed BodyParam/(ApiWalletPassphrase) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000/"

  test/unit/Cardano/Wallet/ApiSpec.hs:142:42:
  6) Cardano.Wallet.Api, Malformed BodyParam, (ApiWalletPassphrase) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected response body \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"Error in $.passphrase: parsing Text failed, expected String, but encountered Array\\\"}\", but received \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"Invalid stake pool id: expecting a Bech32 encoded value with human readable part of 'pool'.\\\"}\""

  To rerun use: --match "/Cardano.Wallet.Api/Malformed BodyParam/(ApiWalletPassphrase) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000/"

  test/unit/Cardano/Wallet/ApiSpec.hs:142:42:
  7) Cardano.Wallet.Api, Malformed BodyParam, (ApiWalletPassphrase) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected response body \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"Error in $.passphrase: parsing Text failed, expected String, but encountered Number\\\"}\", but received \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"Invalid stake pool id: expecting a Bech32 encoded value with human readable part of 'pool'.\\\"}\""

  To rerun use: --match "/Cardano.Wallet.Api/Malformed BodyParam/(ApiWalletPassphrase) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000/"

  test/unit/Cardano/Wallet/ApiSpec.hs:142:42:
  8) Cardano.Wallet.Api, Malformed BodyParam, (ApiWalletPassphrase) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected response body \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"Error in $: parsing Cardano.Wallet.Api.Types.ApiWalletPassphrase(ApiWalletPassphrase) failed, expected Object, but encountered Number\\\"}\", but received \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"Invalid stake pool id: expecting a Bech32 encoded value with human readable part of 'pool'.\\\"}\""

  To rerun use: --match "/Cardano.Wallet.Api/Malformed BodyParam/(ApiWalletPassphrase) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000/"

  test/unit/Cardano/Wallet/ApiSpec.hs:142:42:
  9) Cardano.Wallet.Api, Malformed BodyParam, (ApiWalletPassphrase) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected response body \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"Error in $: parsing Cardano.Wallet.Api.Types.ApiWalletPassphrase(ApiWalletPassphrase) failed, expected Object, but encountered String\\\"}\", but received \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"Invalid stake pool id: expecting a Bech32 encoded value with human readable part of 'pool'.\\\"}\""

  To rerun use: --match "/Cardano.Wallet.Api/Malformed BodyParam/(ApiWalletPassphrase) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000/"

  test/unit/Cardano/Wallet/ApiSpec.hs:142:42:
  10) Cardano.Wallet.Api, Malformed BodyParam, (ApiWalletPassphrase) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected response body \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"trailing junk after valid JSON: endOfInput\\\"}\", but received \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"Invalid stake pool id: expecting a Bech32 encoded value with human readable part of 'pool'.\\\"}\""

  To rerun use: --match "/Cardano.Wallet.Api/Malformed BodyParam/(ApiWalletPassphrase) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000/"

  test/unit/Cardano/Wallet/ApiSpec.hs:142:42:
  11) Cardano.Wallet.Api, Malformed BodyParam, (ApiWalletPassphrase) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected response body \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"I couldn't understand the content of your message. If your message is intended to be in JSON format, please check that the JSON is valid.\\\"}\", but received \"{\\\"code\\\":\\\"bad_request\\\",\\\"message\\\":\\\"Invalid stake pool id: expecting a Bech32 encoded value with human readable part of 'pool'.\\\"}\""

  To rerun use: --match "/Cardano.Wallet.Api/Malformed BodyParam/(ApiWalletPassphrase) PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000/"

  test/unit/Cardano/Wallet/ApiSpec.hs:148:46:
  12) Cardano.Wallet.Api, Malformed Headers, ("Accept") PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected status code 406, but received 400"

  To rerun use: --match "/Cardano.Wallet.Api/Malformed Headers/(\"Accept\") PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000/"

  test/unit/Cardano/Wallet/ApiSpec.hs:155:46:
  13) Cardano.Wallet.Api, Malformed Headers, ("Content-Type") PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected status code 415, but received 400"

  To rerun use: --match "/Cardano.Wallet.Api/Malformed Headers/(\"Content-Type\") PUT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000/"

  test/unit/Cardano/Wallet/ApiSpec.hs:162:42:
  14) Cardano.Wallet.Api, Not Allowed Methods,  GET /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected status code 405, but received 400"

  To rerun use: --match "/Cardano.Wallet.Api/Not Allowed Methods/ GET /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000/"

  test/unit/Cardano/Wallet/ApiSpec.hs:162:42:
  15) Cardano.Wallet.Api, Not Allowed Methods,  POST /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected status code 405, but received 400"

  To rerun use: --match "/Cardano.Wallet.Api/Not Allowed Methods/ POST /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000/"

  test/unit/Cardano/Wallet/ApiSpec.hs:162:42:
  16) Cardano.Wallet.Api, Not Allowed Methods,  PATCH /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected status code 405, but received 400"

  To rerun use: --match "/Cardano.Wallet.Api/Not Allowed Methods/ PATCH /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000/"

  test/unit/Cardano/Wallet/ApiSpec.hs:162:42:
  17) Cardano.Wallet.Api, Not Allowed Methods,  DELETE /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected status code 405, but received 400"

  To rerun use: --match "/Cardano.Wallet.Api/Not Allowed Methods/ DELETE /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000/"

  test/unit/Cardano/Wallet/ApiSpec.hs:162:42:
  18) Cardano.Wallet.Api, Not Allowed Methods,  CONNECT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected status code 405, but received 400"

  To rerun use: --match "/Cardano.Wallet.Api/Not Allowed Methods/ CONNECT /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000/"

  test/unit/Cardano/Wallet/ApiSpec.hs:162:42:
  19) Cardano.Wallet.Api, Not Allowed Methods,  TRACE /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected status code 405, but received 400"

  To rerun use: --match "/Cardano.Wallet.Api/Not Allowed Methods/ TRACE /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000/"

  test/unit/Cardano/Wallet/ApiSpec.hs:162:42:
  20) Cardano.Wallet.Api, Not Allowed Methods,  OPTIONS /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000
       uncaught exception: WaiTestFailure
       WaiTestFailure "Expected status code 405, but received 400"

  To rerun use: --match "/Cardano.Wallet.Api/Not Allowed Methods/ OPTIONS /stake-pools/0000000000000000000000000000000000000000000000000000000000000000/wallets/0000000000000000000000000000000000000000/"
```

</p>
</details>

> b) bech32 encoded pool ids are accepted in path parameters

I believe we just have to adjust this:

```hs
instance Wellformed (PathParam ApiPoolId) where
    wellformed = PathParam $ T.replicate 64 "0"
```

And add a bech32 param there.

KtorZ:
> I believe we just have to adjust this:

I don't think you can provide multiple well-formed data in here. But okay, pretty convincing for the hex backward-compability. 
And I assume that the bech32 encoding is already covered by the integration tests using what's the API outputs.

KtorZ (approved):

PR #2057, opened 2020-08-24
Call `removeRetiredPools` from `monitorStakePools`.

jonathanknowles:
# Issue Number

#2019 

# Overview

PR #2057 provided the [`removeRetiredPools`](https://github.com/input-output-hk/cardano-wallet/blob/fece720cb18d9676c4f1f6b0d3a69a37fab93f1c/lib/core/src/Cardano/Pool/DB.hs#L301) function which, when given an epoch, removes pools from the database that retired **on or before** that epoch.

This PR adjusts [`monitorStakePools`](https://github.com/input-output-hk/cardano-wallet/blob/fece720cb18d9676c4f1f6b0d3a69a37fab93f1c/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs#L481) to actually **call** [`removeRetiredPools`](https://github.com/input-output-hk/cardano-wallet/blob/fece720cb18d9676c4f1f6b0d3a69a37fab93f1c/lib/core/src/Cardano/Pool/DB.hs#L301).

## Frequency of Garbage Collection

Since garbage collection has a **runtime cost** associated with it, we want to avoid doing it more than necessary, even if the cost is relatively small.

Thus, we only perform garbage collection:

- **on demand**, just before we record a new pool registration or retirement certificate in the database.
- **at most once per epoch**.

### A single garbage collection invocation (at most once per epoch)

<details><summary>Click to view log sample</summary>

```hs
[cardano-wallet.pools-engine:Debug:31] [2020-08-31 05:42:24.77 UTC] Performing garbage collection of retired stake pools. Currently in epoch 214. Removing all pools that retired in or before epoch 212.
[cardano-wallet.pools-db:Debug:31] [2020-08-31 05:42:24.77 UTC] Removing pools that retired in or before epoch 212: start
[cardano-wallet.pools-db:Debug:31] [2020-08-31 05:42:24.78 UTC] Removing the following retired pools:
Pool 215622e6 with retirement epoch 212
Pool 5687ecfe with retirement epoch 212
Pool 69d158eb with retirement epoch 212
Pool 7bacd597 with retirement epoch 212
Pool 8e681030 with retirement epoch 212
Pool a6d582d8 with retirement epoch 212
Pool bccf5ce0 with retirement epoch 212
Pool c0b86026 with retirement epoch 212
Pool cfe83240 with retirement epoch 212
Pool f5863386 with retirement epoch 212
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.78 UTC] Removing the following pool from the database: 215622e669c0f0c09828ec4100a4c689b1f9c148f542b0192b6ed00d.
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.80 UTC] Removing the following pool from the database: 5687ecfeaf19ed1a7ec20a9332ff80c8ad6fc5ace97cfb7276d256a2.
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.82 UTC] Removing the following pool from the database: 69d158eb61cbbeff3ae564768eef7c0b52ba8bd33b2301eacf62a24d.
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.84 UTC] Removing the following pool from the database: 7bacd59728317961e6ffb20f40f74706954a5341d1159f7e2be32a68.
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.85 UTC] Removing the following pool from the database: 8e68103060b32595d4eed0a7749795b155d790b95e0615f9f5618b9f.
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.87 UTC] Removing the following pool from the database: a6d582d86cafd861984ec33f3c4bb7ab1e183d0719abc92351dd2dd7.
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.89 UTC] Removing the following pool from the database: bccf5ce0d9b3436cd1616686c98f4584b43760704c9585f1f208ffae.
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.90 UTC] Removing the following pool from the database: c0b86026e65261a0698e789fc2e0b1cf36c85693dbf1cd516067400d.
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.92 UTC] Removing the following pool from the database: cfe83240dd1eb92bac3d739b2550be334f4bb7e4fc11d7f5faeb8769.
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.94 UTC] Removing the following pool from the database: f58633867bea26335d12aa95a8b3b1dee5b6994ffd4bac5348375220.
[cardano-wallet.pools-db:Debug:31] [2020-08-31 05:42:24.95 UTC] Removing pools that retired in or before epoch 212: finish
```
</details>

### Lifecycle of a single pool

<details><summary>Click to view log sample</summary>

```hs
[cardano-wallet.pools-engine:Info:37] [2020-08-27 15:35:54.31 UTC] Discovered stake pool registration: Registration of 00008d6a owned by [ed25519_pk1lrrgks5hnpxqqpyhjxedk7u5nxp0ehp9tm9ksx4a7zepzt6ydnw]
...
[cardano-wallet.pools-engine:Info:37] [2020-08-27 15:35:54.85 UTC] Discovered stake pool retirement: Pool 00008d6a with retirement epoch 209
...
[cardano-wallet.pools-engine:Debug:37] [2020-08-27 15:36:16.07 UTC] Performing garbage collection of retired stake pools. Currently in epoch 211. Removing all pools that retired in or before epoch 209.
[cardano-wallet.pools-db:Notice:37] [2020-08-27 15:36:16.08 UTC] Removing the following pool from the database: 00008d6abd6bf0dafbc7f64b68267f5716e1992d0e845eb22d05065b.
```

</details>

### DIFF
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
@@ -665,14 +665,13 @@ instance ToText StakePoolLog where
             ]
         MsgRollingBackTo point ->
             "Rolling back to " <> pretty point
-        MsgStakePoolGarbageCollection info bkt -> mconcat
+        MsgStakePoolGarbageCollection info _ -> mconcat
             [ "Performing garbage collection of retired stake pools. "
             , "Currently in epoch "
             , toText (currentEpoch info)
             , ". Removing all pools that retired in or before epoch "
             , toText (removalEpoch info)
-            , ". "
-            , toText bkt
+            , "."
             ]
         MsgStakePoolRegistration cert ->
             "Discovered stake pool registration: " <> pretty cert
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
@@ -34,7 +34,11 @@ import Cardano.BM.Data.Severity
 import Cardano.BM.Data.Tracer
     ( HasPrivacyAnnotation (..), HasSeverityAnnotation (..) )
 import Cardano.Pool.DB
-    ( DBLayer (..), ErrPointAlreadyExists (..), readPoolLifeCycleStatus )
+    ( DBLayer (..)
+    , ErrPointAlreadyExists (..)
+    , readPoolLifeCycleStatus
+    , removeRetiredPools
+    )
 import Cardano.Pool.DB.Log
     ( PoolDbLog )
 import Cardano.Pool.Metadata
@@ -51,7 +55,7 @@ import Cardano.Wallet.Network
     , follow
     )
 import Cardano.Wallet.Primitive.Slotting
-    ( TimeInterpreter, firstSlotInEpoch, startTime )
+    ( TimeInterpreter, epochOf, epochPred, firstSlotInEpoch, startTime )
 import Cardano.Wallet.Primitive.Types
     ( ActiveSlotCoefficient (..)
     , BlockHeader
@@ -87,7 +91,7 @@ import Cardano.Wallet.Unsafe
 import Control.Concurrent
     ( threadDelay )
 import Control.Monad
-    ( forM, forM_, forever, when )
+    ( forM, forM_, forever, when, (<=<) )
 import Control.Monad.IO.Class
     ( liftIO )
 import Control.Monad.Trans.Except
@@ -519,9 +523,22 @@ monitorStakePools tr gp nl db@DBLayer{..} = do
                             liftIO $ traceWith tr $ MsgErrProduction e
                         Right () ->
                             pure ()
+                garbageCollectPools slot
                 putPoolCertificates slot certificates
         pure Continue
       where
+        -- Perform garbage collection for pools that have retired.
+        --
+        -- To avoid any issues with rollback, we err on the side of caution and
+        -- only garbage collect pools that retired /at least/ two epochs before
+        -- the current epoch.
+        --
+        garbageCollectPools currentSlot = liftIO $ do
+            currentEpoch <- timeInterpreter nl (epochOf currentSlot)
+            let subtractTwoEpochs = epochPred <=< epochPred
+            forM_ (subtractTwoEpochs currentEpoch) $ \safeRetirementEpoch ->
+                removeRetiredPools db (contramap MsgDb tr) safeRetirementEpoch
+
         putPoolCertificates slot certificates = do
             -- A single block can contain multiple certificates relating to the
             -- same pool.
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
@@ -35,8 +35,6 @@ import Cardano.BM.Data.Tracer
     ( HasPrivacyAnnotation (..), HasSeverityAnnotation (..) )
 import Cardano.Pool.DB
     ( DBLayer (..), ErrPointAlreadyExists (..), readPoolLifeCycleStatus )
-import Cardano.Pool.DB.Log
-    ( PoolDbLog )
 import Cardano.Pool.Metadata
     ( StakePoolMetadataFetchLog )
 import Cardano.Wallet.Api.Types
@@ -620,8 +618,7 @@ monitorMetadata tr gp fetchMetadata DBLayer{..} = forever $ do
         f = unActiveSlotCoefficient (getActiveSlotCoefficient gp)
 
 data StakePoolLog
-    = MsgDb PoolDbLog
-    | MsgFollow FollowLog
+    = MsgFollow FollowLog
     | MsgStartMonitoring [BlockHeader]
     | MsgHaltMonitoring
     | MsgCrashMonitoring
@@ -648,7 +645,6 @@ data PoolGarbageCollectionInfo = PoolGarbageCollectionInfo
 instance HasPrivacyAnnotation StakePoolLog
 instance HasSeverityAnnotation StakePoolLog where
     getSeverityAnnotation = \case
-        MsgDb e -> getSeverityAnnotation e
         MsgFollow e -> getSeverityAnnotation e
         MsgStartMonitoring{} -> Info
         MsgHaltMonitoring{} -> Info
@@ -663,8 +659,6 @@ instance HasSeverityAnnotation StakePoolLog where
 
 instance ToText StakePoolLog where
     toText = \case
-        MsgDb e ->
-            toText e
         MsgFollow e ->
             toText e
         MsgStartMonitoring cursor -> mconcat
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
@@ -85,7 +85,7 @@ import Cardano.Wallet.Unsafe
 import Control.Concurrent
     ( threadDelay )
 import Control.Monad
-    ( forM, forM_, forever, unless, when, (<=<) )
+    ( forM, forM_, forever, unless, void, when, (<=<) )
 import Control.Monad.IO.Class
     ( liftIO )
 import Control.Monad.Trans.Except
@@ -98,6 +98,8 @@ import Data.Function
     ( (&) )
 import Data.Generics.Internal.VL.Lens
     ( view )
+import Data.IORef
+    ( IORef, newIORef, readIORef, writeIORef )
 import Data.List.NonEmpty
     ( NonEmpty (..) )
 import Data.Map
@@ -474,9 +476,11 @@ monitorStakePools
     -> DBLayer IO
     -> IO ()
 monitorStakePools tr gp nl db@DBLayer{..} = do
+    latestGarbageCollectionEpochRef <- mkLatestGarbageCollectionEpochRef
+    let forwardHandler = forward latestGarbageCollectionEpochRef
     cursor <- initCursor
     traceWith tr $ MsgStartMonitoring cursor
-    follow nl (contramap MsgFollow tr) cursor forward getHeader >>= \case
+    follow nl (contramap MsgFollow tr) cursor forwardHandler getHeader >>= \case
         FollowInterrupted -> traceWith tr MsgHaltMonitoring
         FollowFailure -> traceWith tr MsgCrashMonitoring
         FollowRollback point -> do
@@ -489,6 +493,18 @@ monitorStakePools tr gp nl db@DBLayer{..} = do
         , getEpochStability
         } = gp
 
+    -- In order to prevent the pool database from growing too large over time,
+    -- we regularly perform garbage collection by removing retired pools from
+    -- the database.
+    --
+    -- However, there is no benefit to be gained from collecting garbage more
+    -- than once per epoch. In order to avoid invoking the garbage collector
+    -- too frequently, we keep a reference to the latest epoch for which
+    -- garbage collection was performed.
+    --
+    mkLatestGarbageCollectionEpochRef :: IO (IORef EpochNo)
+    mkLatestGarbageCollectionEpochRef = newIORef minBound
+
     initCursor :: IO [BlockHeader]
     initCursor = atomically $ readPoolProductionCursor (max 100 k)
       where k = fromIntegral $ getQuantity getEpochStability
@@ -497,10 +513,11 @@ monitorStakePools tr gp nl db@DBLayer{..} = do
     getHeader = toCardanoBlockHeader gp
 
     forward
-        :: NonEmpty (CardanoBlock sc)
+        :: IORef EpochNo
+        -> NonEmpty (CardanoBlock sc)
         -> (BlockHeader, ProtocolParameters)
         -> IO (FollowAction ())
-    forward blocks (_nodeTip, _pparams) = do
+    forward latestGarbageCollectionEpochRef blocks (_nodeTip, _pparams) = do
         atomically $ forM_ blocks $ \case
             BlockByron _ -> pure ()
             BlockShelley blk -> do
@@ -515,10 +532,8 @@ monitorStakePools tr gp nl db@DBLayer{..} = do
                 unless (null certificates) $ do
                     -- Before adding new pool certificates to the database, we
                     -- first attempt to garbage collect pools that have already
-                    -- retired. By only performing garbage collection when we
-                    -- actually have new certificates to add, we avoid paying
-                    -- the cost of garbage collection in every single slot.
-                    garbageCollectPools slot
+                    -- retired.
+                    garbageCollectPools slot latestGarbageCollectionEpochRef
                     putPoolCertificates slot certificates
         pure Continue
 
@@ -546,16 +561,23 @@ monitorStakePools tr gp nl db@DBLayer{..} = do
     -- that occurred two epochs ago that has not subsquently been superseded,
     -- it should be safe to garbage collect that pool.
     --
-    garbageCollectPools currentSlot = do
+    garbageCollectPools currentSlot latestGarbageCollectionEpochRef = do
         currentEpoch <- liftIO $ timeInterpreter nl (epochOf currentSlot)
         let subtractTwoEpochs = epochPred <=< epochPred
         forM_ (subtractTwoEpochs currentEpoch) $ \latestRetirementEpoch -> do
-            liftIO
-                $ traceWith tr
-                $ MsgStakePoolGarbageCollection
-                $ PoolGarbageCollectionInfo
-                    {currentEpoch, latestRetirementEpoch}
-            removeRetiredPools latestRetirementEpoch
+            latestGarbageCollectionEpoch <-
+                liftIO $ readIORef latestGarbageCollectionEpochRef
+            -- Only perform garbage collection once per epoch:
+            when (latestRetirementEpoch > latestGarbageCollectionEpoch) $ do
+                liftIO $ do
+                    writeIORef
+                        latestGarbageCollectionEpochRef
+                        latestRetirementEpoch
+                    traceWith tr $
+                        MsgStakePoolGarbageCollection $
+                        PoolGarbageCollectionInfo
+                            {currentEpoch, latestRetirementEpoch}
+                void $ removeRetiredPools latestRetirementEpoch
 
     -- For each pool certificate in the given list, add an entry to the
     -- database that associates the certificate with the specified slot
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
@@ -638,7 +638,7 @@ instance HasSeverityAnnotation StakePoolLog where
         MsgHaltMonitoring{} -> Info
         MsgCrashMonitoring{} -> Error
         MsgRollingBackTo{} -> Info
-        MsgStakePoolGarbageCollection{} -> Info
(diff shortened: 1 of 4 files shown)


lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -621,6 +668,15 @@ instance ToText StakePoolLog where
...
             "Rolling back to " <> pretty point
+        MsgStakePoolGarbageCollection info bkt -> mconcat
+            [ "Performing garbage collection of retired stake pools. "
+            , "Currently in epoch "
+            , toText (poolGarbageCollectionCurrentEpoch info)
+            , ". Removing all pools that retired in or before epoch "
+            , toText (poolGarbageCollectionRemovalEpoch info)
+            , ". "
+            , toText bkt
+            ]
```
jonathanknowles, line 701: See if this can be made simpler through use of [`Fmt`](http://hackage.haskell.org/package/fmt-0.6.1.2/docs/Fmt.html).

jonathanknowles:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -577,26 +603,40 @@ monitorMetadata tr gp fetchMetadata DBLayer{..} = forever $ do
...
 instance HasPrivacyAnnotation StakePoolLog
 instance HasSeverityAnnotation StakePoolLog where
     getSeverityAnnotation = \case
+        MsgDb e -> getSeverityAnnotation e
         MsgFollow e -> getSeverityAnnotation e
         MsgStartMonitoring{} -> Info
         MsgHaltMonitoring{} -> Info
         MsgCrashMonitoring{} -> Error
         MsgRollingBackTo{} -> Info
+        MsgStakePoolGarbageCollection{} -> Info
```
jonathanknowles: This should probably be lowered to `Debug` in order to avoid spamming the log. In particular, we don't need to see any informational messages when there are no pools to garbage collect.

jonathanknowles: > This should probably be lowered to `Debug` in order to avoid spamming the log. In particular, we don't need to see any informational messages when there are no pools to garbage collect.

Fixed in b1e4d1506c15f1be05fc1ad2416badee69a47529.

jonathanknowles:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -577,26 +605,40 @@ monitorMetadata tr gp fetchMetadata DBLayer{..} = forever $ do
...
     | MsgErrProduction ErrPointAlreadyExists
     | MsgFetchPoolMetadata StakePoolMetadataFetchLog
     | MsgFetchTakeBreak Int
     deriving (Show, Eq)
 
+data PoolGarbageCollectionInfo = PoolGarbageCollectionInfo
+    { currentEpoch :: EpochNo
+        -- ^ The current epoch at the point in time the garbage collector
+        -- was invoked.
+    , removalEpoch :: EpochNo
```
jonathanknowles: **Suggestion:** The name `removalEpoch` seems a little unclear. Perhaps call it `latestRetirementEpoch`?

**Justification:** the name `latestRetirementEpoch` implies a relationship with `retirementEpoch` in `PoolRetirementCertificate`, which is exactly the implication we want.

```suggestion
    , latestRetirementEpoch :: EpochNo
```

jonathanknowles: > **Suggestion:** The name `removalEpoch` seems a little unclear. Perhaps call it `latestRetirementEpoch`?

Fixed in 78e8e04f888c7288166147b63d22fdff6ccdc7c2.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -517,31 +525,52 @@ monitorStakePools tr gp nl db@DBLayer{..} = do
...
 
+    -- Perform garbage collection for pools that have retired.
+    --
+    -- To avoid any issues with rollback, we err on the side of caution and
+    -- only garbage collect pools that retired /at least/ two epochs before
+    -- the current epoch.
+    --
+    garbageCollectPools currentSlot = liftIO $ do
+        currentEpoch <- timeInterpreter nl (epochOf currentSlot)
+        let subtractTwoEpochs = epochPred <=< epochPred
```
Anviking, line 566: What if the current epoch is 0? (Perhaps not technically possible this is called then, but still…)

jonathanknowles: @Anviking wrote:
> What if the current epoch is 0? (Perhaps not technically possible this is called then, but still…)

I'm glad you asked :smile_cat:.

In that case, we will not perform any garbage collection, as `subtractTwoEpochs currentEpoch` will evaluate to `Nothing`. (Since [`epochPred`](https://github.com/input-output-hk/cardano-wallet/blob/fece720cb18d9676c4f1f6b0d3a69a37fab93f1c/lib/core/src/Cardano/Wallet/Primitive/Slotting.hs#L364) has type `EpochNo -> Maybe EpochNo`.)

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -577,26 +606,41 @@ monitorMetadata tr gp fetchMetadata DBLayer{..} = forever $ do
         f = unActiveSlotCoefficient (getActiveSlotCoefficient gp)
 
 data StakePoolLog
-    = MsgFollow FollowLog
+    = MsgDb PoolDbLog
```
KtorZ: We already have a logger for the pool database, why is this needed here?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -517,31 +525,52 @@ monitorStakePools tr gp nl db@DBLayer{..} = do
...
+                    -- the cost of garbage collection in every single slot.
+                    garbageCollectPools slot
+                    putPoolCertificates slot certificates
         pure Continue
 
+    -- Perform garbage collection for pools that have retired.
+    --
+    -- To avoid any issues with rollback, we err on the side of caution and
+    -- only garbage collect pools that retired /at least/ two epochs before
+    -- the current epoch.
```
KtorZ, line 544: What's the rationale for "at least 2". Why not, "at least 3" for example?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
jonathanknowles: @KtorZ 
> We already have a logger for the pool database, why is this needed here?

Good question.

When we call the `removeRetiredPools` function, it expects to be provided with a `Tracer IO PoolDbLog` value. But we call it from a context where we only have a `Tracer IO StakePoolLog`. Therefore, we need to have some way to transform the `Tracer IO StakePoolLog` into a `Tracer IO PoolDbLog`.

Luckily, the `contramap` function allows us to perform this transformation, but it needs to be provided with something that maps a `PoolDbLog` into a `StakePoolLog`, which is where the `MsgDb` constructor comes in.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
jonathanknowles: > What's the rationale for "at least 2". Why not, "at least 3" for example?

If we've **only just** crossed the boundary from epoch **_e_** into epoch **_e_** **+ 1**, this means that blocks from the previous epoch are still within the rollback window. If a rollback does happen, and if the chain we've just been following has been replaced with another chain in which the retirement certificate for some pool **p** is revoked within epoch **_e_** (with a registration certificate that supersedes it), then the retirement for pool **p** would be cancelled.

But assuming the rollback window is guaranteed to be less than one epoch in length, retirements that occurred two epochs ago can no longer be affected by rollback. So if we see a retirement that occurred two epochs ago and was not superseded, it is safe to perform garbage collection of that pool.

KtorZ: Would it be possible to extend the comment with this rationale :) ?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
KtorZ: That seems quite wrong to me. Why isn't `removeRetiredPools` part of the DBLayer already? This creates an inconsistency with the "normal approach" which seems unjustified to me. 

The problem with having logs database embedded in the wrong logger is that it completely defies the purpose of having separate logger components. Now, if I run `--trace-pool-db off`, I still get pool db logs regarding the garbage collection of pools.

Anviking:
I'm trying this PR out, and my wallet now hangs on `cardano-wallet stake-pool list --stake 10000000000000`.

It could be an unrelated issue (trying to confirm), but I saw it when
1. synced the pools db from scratch
2. used an existing pool db, but moved the garbage collection to outside the `unless` clause, to trigger it without having to wait for a pool cert on chain

Have you seen the garbage collection working using the full wallet / Daedalus?

Anviking:
After seeing the pool worker get stuck, if I comment out the garbage collection, and restart, it's no longer stuck when it reaches the shelley part of the chain.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
jonathanknowles: > Would it be possible to extend the comment with this rationale :) ?

Certainly!

Fixed in 395a64020d9ed0c92d53d0cd4de2d9a38e29a0f9

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
jonathanknowles: > Why isn't `removeRetiredPools` part of the DBLayer already?

That's a good question.

The `removeRetiredPools` function depends **only** on the DBLayer **interface**, and not any particular implementation. It only calls `listRetiredPools` and `removePools`. So this function can sit outside of the `DBLayer` interface itself (in the same way that we do for other function that only depend on the interface).

One alternative would be to add a `removeRetiredPools` function to the `DBLayer` type, and then implement this twice (once in the `SQLite` module and another time in the `MVar` module), but reuse the same function behind the scenes.

KtorZ: > One alternative would be to add a removeRetiredPools function to the DBLayer type, and then implement this twice (once in the SQLite module and another time in the MVar module), but reuse the same function behind the scenes.

Yes. That sounds preferable for the consumers of the DBLayer such that things like the DBLog doesn't leak outside of the DBLayer.

jonathanknowles: @KtorZ wrote:
> The problem with having logs database embedded in the wrong logger is that it completely defies the purpose of having separate logger components. Now, if I run --trace-pool-db off, I still get pool db logs regarding the garbage collection of pools.

The pattern I've used for `MsgDb` is also used by `MsgFollow`, which embeds entries from the `FollowLog`:

```hs
data StakePoolLog 
    = MsgDb PoolDbLog 
    | MsgFollow FollowLog
    ...

data FollowLog
    = MsgFollowAction (FollowAction String)   
    | MsgUnhandledException Text  
    | MsgNextBlockFailed ErrGetBlock 
    ...
```

See: https://github.com/input-output-hk/cardano-wallet/blob/9f93d0a98be7c19035dd603149ebf06ac38d7f2f/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs#L484

Is there a better solution that we can use here?

KtorZ: @jonathanknowles FollowLog is a completely different story, it is not part of any other logger component. There are two components that include a `FollowLog` and these are the `stake-pool-engine` and `wallet-engine` loggers. 

However, the DBLog is already a logger component all by itself. See:

https://github.com/input-output-hk/cardano-wallet/blob/master/lib/shelley/src/Cardano/Wallet/Shelley.hs#L463-L472

jonathanknowles: @KtorZ many thanks for your explanation, that makes sense.

I've made the following changes:
* moved `removeRetiredPools` into the `DBLayer`.
* completely removed the `MsgDb` constructor.

See the following commits:
- 2ed42a2e0b1ba8b9a939a855c6ee40505fd348ff
- 742027cb266cb77e6f7bfd073211bbb84bb404d1

jonathanknowles:
@Anviking wrote:
> I'm trying this PR out, and my wallet now hangs on cardano-wallet stake-pool list --stake 10000000000000.
> After seeing the pool worker get stuck, **if I comment out the garbage collection**, and restart, it's no longer stuck when it reaches the shelley part of the chain.

Many thanks for this feedback.

In `garbageCollectPools`, we rely on the time interpreter to determine the current epoch:
```hs
currentEpoch <- timeInterpreter nl (epochOf currentSlot)
```

I wonder whether that call could be blocking? 

I will experiment with commenting this section out.

jonathanknowles:
bors try

jonathanknowles:
I've updated the branch to perform garbage collection at **_epoch boundaries_**, rather than every time we see a new certificate.
    
This has two advantages:
    
1.  We reduce the frequency of garbage collections that do nothing. Although garbage collection is idempotent and safe to run more than once, there's no real point in running a garbage collection more than once per epoch.
    
2.  We can read the epoch number directly from the epoch boundary block, which means we no longer need to rely on the time interpreter to determine the current epoch.

See: 622cb1a8b5416ff56ff3ba44e108f8a88d5175bc.

jonathanknowles:
> I've updated the branch to perform garbage collection at **_epoch boundaries_**, rather than every time we see a new certificate.

I've just reverted this change after a discussion with @Anviking. It seems the API to interrogate a Shelley boundary block still exists and compiles, but the `headerIsEBB` function will never return anything other than `Nothing`, as EBBs have been completely removed for Shelley.

jonathanknowles:
@Anviking wrote:

> I'm trying this PR out, and my wallet now hangs on `cardano-wallet stake-pool list --stake 10000000000000`.

This is now **fixed** in 523c77817dc1d5b25b4330be72297584037ffd97. The cause was a **nested call to `DBLayer.atomically`**, which has now been removed. :+1:

I've tested the result code with `mainnet`, and have confirmed that garbage collection works as expected. Just to be sure, I synced to the network tip, and then manually verified that:
- every removed pool has a prior retirement certificate.
- every removed pool was removed no earlier than two epochs after its declared retirement epoch.

## Logging Examples

The following log samples are taken from a wallet connecting to `mainnet`.

### A single garbage collection invocation (run at most once per epoch)

<details><summary>Click to view log sample</summary>

```hs
[cardano-wallet.pools-engine:Debug:31] [2020-08-31 05:42:24.77 UTC] Performing garbage collection of retired stake pools. Currently in epoch 214. Removing all pools that retired in or before epoch 212.
[cardano-wallet.pools-db:Debug:31] [2020-08-31 05:42:24.77 UTC] Removing pools that retired in or before epoch 212: start
[cardano-wallet.pools-db:Debug:31] [2020-08-31 05:42:24.78 UTC] Removing the following retired pools:
Pool 215622e6 with retirement epoch 212
Pool 5687ecfe with retirement epoch 212
Pool 69d158eb with retirement epoch 212
Pool 7bacd597 with retirement epoch 212
Pool 8e681030 with retirement epoch 212
Pool a6d582d8 with retirement epoch 212
Pool bccf5ce0 with retirement epoch 212
Pool c0b86026 with retirement epoch 212
Pool cfe83240 with retirement epoch 212
Pool f5863386 with retirement epoch 212
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.78 UTC] Removing the following pool from the database: 215622e669c0f0c09828ec4100a4c689b1f9c148f542b0192b6ed00d.
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.80 UTC] Removing the following pool from the database: 5687ecfeaf19ed1a7ec20a9332ff80c8ad6fc5ace97cfb7276d256a2.
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.82 UTC] Removing the following pool from the database: 69d158eb61cbbeff3ae564768eef7c0b52ba8bd33b2301eacf62a24d.
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.84 UTC] Removing the following pool from the database: 7bacd59728317961e6ffb20f40f74706954a5341d1159f7e2be32a68.
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.85 UTC] Removing the following pool from the database: 8e68103060b32595d4eed0a7749795b155d790b95e0615f9f5618b9f.
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.87 UTC] Removing the following pool from the database: a6d582d86cafd861984ec33f3c4bb7ab1e183d0719abc92351dd2dd7.
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.89 UTC] Removing the following pool from the database: bccf5ce0d9b3436cd1616686c98f4584b43760704c9585f1f208ffae.
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.90 UTC] Removing the following pool from the database: c0b86026e65261a0698e789fc2e0b1cf36c85693dbf1cd516067400d.
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.92 UTC] Removing the following pool from the database: cfe83240dd1eb92bac3d739b2550be334f4bb7e4fc11d7f5faeb8769.
[cardano-wallet.pools-db:Notice:31] [2020-08-31 05:42:24.94 UTC] Removing the following pool from the database: f58633867bea26335d12aa95a8b3b1dee5b6994ffd4bac5348375220.
[cardano-wallet.pools-db:Debug:31] [2020-08-31 05:42:24.95 UTC] Removing pools that retired in or before epoch 212: finish
```
</details>

### Lifecycle of a single pool

<details><summary>Click to view log sample</summary>

```hs
[cardano-wallet.pools-engine:Info:37] [2020-08-27 15:35:54.31 UTC] Discovered stake pool registration: Registration of 00008d6a owned by [ed25519_pk1lrrgks5hnpxqqpyhjxedk7u5nxp0ehp9tm9ksx4a7zepzt6ydnw]
...
[cardano-wallet.pools-engine:Info:37] [2020-08-27 15:35:54.85 UTC] Discovered stake pool retirement: Pool 00008d6a with retirement epoch 209
...
[cardano-wallet.pools-engine:Debug:37] [2020-08-27 15:36:16.07 UTC] Performing garbage collection of retired stake pools. Currently in epoch 211. Removing all pools that retired in or before epoch 209.
[cardano-wallet.pools-db:Notice:37] [2020-08-27 15:36:16.08 UTC] Removing the following pool from the database: 00008d6abd6bf0dafbc7f64b68267f5716e1992d0e845eb22d05065b.
```

</details>

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
Anviking: 👍

Anviking (approved):
lgtm

Issue #1971, opened 2020-07-28
Listing transaction when node is still in the Byron era may fail with 500


Anviking:
# Context
 
 

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | Close to Shelley Mainnet release |
| Platform     |           |
| Installation |             |

With #1869, we use a `TimeInterpreter` from the node for time and slotting conversions. The conversions fails if "Past the Horizon". From genesis to close to the fork, the node will have a limited horizon.

# Steps to Reproduce

1. Have a node and wallet which is running Byron;Shelley, but which has not yet reached Shelley. (E.g. on Cardano mainnet when first syncing from genesis)
1. List transactions with a time range were one of the dates are far ahead of the node tip (which is continuously syncing).

## Expected behavior

1. I get a response, or a helpful error

## Actual behavior

1. (Theorized) `500: Something went wrong`

*Note*: This shouldn't happen when the node is in Shelley.
 

---

# Resolution 

---

# QA

piotr-iohk:
That is the case indeed on mainnet currently:
```
$ cardano-wallet-byron transaction list --port 8886 <wid> --start 2016-11-21T10:15:00Z --end 2020-08-29T10:15:00Z
Something went wrong
```

```
[cardano-wallet.network:Error:160925] [2020-07-29 12:54:20.88 UTC] Time interpreter queried past the horizon. Query is:
PastHorizon [("runQueryqBind qBind qBind qBind UTCTimeToRel 2016-11-21 10:15:00 UTC
"Eras are:
EraSummary {eraStart = Bound {boundTime = RelativeTime {getRelativeTime = 0s}, boundSlot = SlotNo {unSlotNo = 0}, boundEpoch = EpochNo {unEpochNo = 0}}, eraEnd = EraEnd (Bound {boundTime = RelativeTime {getRelativeTime = 89856000s}, boundSlot = SlotNo {unSlotNo = 4492800}, boundEpoch = EpochNo {unEpochNo = 208}}), eraParams = EraParams {eraEpochSize = EpochSize {unEpochSize = 21600}, eraSlotLength = SlotLength {getSlotLength = 20s}, eraSafeZone = StandardSafeZone 4320 NoLowerBound}}
[...]

PR #2059, opened 2020-08-24
Catch PastHorizon Exception and return 400 with descriptive Error

hasufell:
# Issue Number

#1971 

# Overview

- catch the exception and return 400 with descriptive Error

# Comments

- should this be unit tested? If so, how?

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -1844,7 +1844,8 @@ listTransactions ctx wid mMinWithdrawal mStart mEnd order = db & \DBLayer{..} ->
         _ -> do
             liftIO (try $ ti $ slotRangeFromTimeRange $ Range mStart mEnd) >>=
                 \case
-                    Left e@(PastHorizon{}) -> throwE (ErrListTransactionsPastHorizonException e)
+                    Left e@(PastHorizon{}) ->
+                        throwE (ErrListTransactionsPastHorizonException e)
                     Right r -> pure r
 
 
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -271,7 +271,11 @@ import Cardano.Wallet.Primitive.Model
     , updateState
     )
 import Cardano.Wallet.Primitive.Slotting
-    ( TimeInterpreter, slotRangeFromTimeRange, startTime )
+    ( PastHorizonException (..)
+    , TimeInterpreter
+    , slotRangeFromTimeRange
+    , startTime
+    )
 import Cardano.Wallet.Primitive.SyncProgress
     ( SyncProgress, SyncTolerance (..), syncProgress )
 import Cardano.Wallet.Primitive.Types
@@ -327,7 +331,7 @@ import Cardano.Wallet.Transaction
 import Cardano.Wallet.Unsafe
     ( unsafeXPrv )
 import Control.Exception
-    ( Exception )
+    ( Exception, try )
 import Control.Monad
     ( forM, forM_, replicateM, unless, when )
 import Control.Monad.IO.Class
@@ -1838,7 +1842,11 @@ listTransactions ctx wid mMinWithdrawal mStart mEnd order = db & \DBLayer{..} ->
             let err = ErrStartTimeLaterThanEndTime start end
             throwE (ErrListTransactionsStartTimeLaterThanEndTime err)
         _ -> do
-            liftIO $ ti $ slotRangeFromTimeRange $ Range mStart mEnd
+            liftIO (try $ ti $ slotRangeFromTimeRange $ Range mStart mEnd) >>=
+                \case
+                    Left e@(PastHorizon{}) -> throwE (ErrListTransactionsPastHorizonException e)
+                    Right r -> pure r
+
 
 -- | Get transaction and metadata from history for a given wallet.
 getTransaction
@@ -2196,7 +2204,8 @@ data ErrListTransactions
     = ErrListTransactionsNoSuchWallet ErrNoSuchWallet
     | ErrListTransactionsStartTimeLaterThanEndTime ErrStartTimeLaterThanEndTime
     | ErrListTransactionsMinWithdrawalWrong
-    deriving (Show, Eq)
+    | ErrListTransactionsPastHorizonException PastHorizonException
+    deriving (Show)
 
 -- | Errors that can occur when trying to get transaction.
 data ErrGetTransaction
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -2324,10 +2324,12 @@ instance LiftHandler ErrStartTimeLaterThanEndTime where
         ]
 
 instance LiftHandler PastHorizonException where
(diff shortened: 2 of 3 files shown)


lib/core/src/Cardano/Wallet.hs
```diff
@@ -1838,7 +1842,11 @@ listTransactions ctx wid mMinWithdrawal mStart mEnd order = db & \DBLayer{..} ->
             let err = ErrStartTimeLaterThanEndTime start end
             throwE (ErrListTransactionsStartTimeLaterThanEndTime err)
         _ -> do
-            liftIO $ ti $ slotRangeFromTimeRange $ Range mStart mEnd
+            liftIO (try $ ti $ slotRangeFromTimeRange $ Range mStart mEnd) >>=
+                \case
+                    Left e@(PastHorizon{}) -> throwE (ErrListTransactionsPastHorizonException e)
```
hasufell: This is exhaustive https://input-output-hk.github.io/ouroboros-network/ouroboros-consensus/Ouroboros-Consensus-HardFork-History-Summary.html#t:PastHorizonException

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2322,6 +2323,13 @@ instance LiftHandler ErrStartTimeLaterThanEndTime where
         , "'."
         ]
 
+instance LiftHandler PastHorizonException where
+    handler _ = apiError err400 PastHorizon $ mconcat
+        [ "The end time is past the horizon (due to uncertainty about the next"
+        , " hard fork)."
+        , " Wait for the node to finish syncing to the hard fork."
+        ]
```
hasufell, line 2333: Question: Should the eras be listed? (Note that it's a list of Eras)

```hs
-- | We tried to convert something that is past the horizon
--
-- That is, we tried to convert something that is past the point in time
-- beyond which we lack information due to uncertainty about the next
-- hard fork.
--
-- We record the condition we were looking for and the bounds on the summary.
data PastHorizonException = PastHorizon CallStack [EraSummary]

data EraSummary = EraSummary {
      eraStart  :: !Bound     -- ^ Inclusive lower bound
    , eraEnd    :: !EraEnd    -- ^ Exclusive upper bound
    , eraParams :: !EraParams -- ^ Active parameters
    }
  deriving stock    (Show, Eq, Generic)
  deriving anyclass (NoUnexpectedThunks)
```

KtorZ (approved):

KtorZ:
It'd be better to also provide an integration test for this one although it is quite hard to setup with our current test integration framework (which only executes scenarios once the hard-fork has happened!)

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2322,6 +2323,13 @@ instance LiftHandler ErrStartTimeLaterThanEndTime where
         , "'."
         ]
 
+instance LiftHandler PastHorizonException where
+    handler _ = apiError err400 PastHorizon $ mconcat
+        [ "The end time is past the horizon (due to uncertainty about the next"
```
Anviking: The phrase "The end time" is specific to `ErrListTransactionsPastHorizonException`, whereas _this_ instance could theoretically be thrown from any endpoint, I believe.

How about either:
1. Make the the description generic (remove "The end time")
2. Define this message directly in the `ErrListTransactionsPastHorizonException` case

Anviking (approved):
lgtm, modulo a comment!

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: 🤔 I think
> Define this message directly in the ErrListTransactionsPastHorizonException case

would be preferable.

If a `PastHorizonException` were to be thrown in many other places, that would be an unexpected server error (some err5xx), different from this case (err400).

(Having a `LiftHandler PastHorizonException` simply for adding more info than 500 Something wrong in the general unexpected case, could still be good though)

hasufell:
I'm starting to question this is 4xx at all, since the reason the request cannot be fulfilled is that part of the server (the node) isn't ready yet.

hasufell:
So I think 503 is in fact the correct status code: https://stackoverflow.com/a/12281287

lib/core/src/Cardano/Wallet/Api/Server.hs
hasufell: So, IF `PastHorizonException` always means that a later request will succeed, then we can just have the LiftHandler instance return 503. Is that the case?

Anviking:
I think 503 might make sense.

Scenario: Running Byron;Shelly, when node un-synced in Byron

User can wait for either
1. node to reach Shelly
2. node to reach a point in Byron where the supplied time is within the forecast range (if earlier)

Only "but" is that the node is not necessarily guaranteed to reach Shelley, for arbitrary chains. Or if we were to ship Byron;Shelley;Gougen weeks before the hard fork to Gougen were confirmed, it would take… weeks to have a unlimited forecast range.

hasufell:
> Only "but" is that the node is not necessarily guaranteed to reach Shelley, for arbitrary chains. Or if we were to ship Byron;Shelley;Gougen weeks before the hard fork to Gougen were confirmed, it would take… weeks to have a unlimited forecast range.

Good point, but 503 doesn't make any time guarantees, right? It would be nice though if we could give an estimate in the error. Is that possible?

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: > always means that a later request will succeed, then we can just have the LiftHandler instance return 503. Is that the case?

There should exist cases where it's more like "programmer error", and waiting doesn't really help.

For instance:

When listing stake pools, we convert the retirement epochs to `UTCTime`:
https://github.com/input-output-hk/cardano-wallet/blob/90c1e5953a48ffc65f004a68abd545ef68574c4b/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs#L351-L352

In Byron;Shelley;Gougen, before fork to Gougen is confirmed, because the retirements are far in the future, this would likely fail*.

Even if we could wait for the retirement epoch of one pool to enter the forecast range, there could always be additional pools retiring. So waiting won't really help, we need to approach this differently when the time comes.

(*Not a problem currently in Byron;Shelley because pools only exist in Shelley, guaranteeing unlimited forecast)

Anviking: Although, I guess "Wait for the node to finish syncing to the hard fork" is still valid in this case, where waiting for Gougen would technically be a solution, and one would notice if it is thrown in an un-resonable context.

lib/core/src/Cardano/Wallet.hs
jonathanknowles: https://github.com/input-output-hk/cardano-wallet/wiki/Coding-Standards#limit-line-length-to-80-characters

```suggestion
                    Left e@(PastHorizon{}) ->
                        throwE (ErrListTransactionsPastHorizonException e)
```

lib/core/src/Cardano/Wallet/Api/Server.hs
KtorZ: Sounds fine like this already.

KtorZ (approved):

PR #2065, opened 2020-08-25
Fix occasional db tests failure

### LINKED ISSUE #2064: cardano-wallet-core:test:unit sometimes consumes all available memory and fails.
# Context
 
The `cardano-wallet-core:unit` test suite sometimes consumes all available memory and fails.
- On linux it says the test suite exited with status -9 (this is because the oom killer has killed it).
- On macos it says the test suite was killed.
- On windows (wine) it says that a memory allocation failed.

The cause is likely to be an infinite loop, perhaps due to a faulty QuickCheck generator or shrinker.

This error seems to have become a lot more frequent recently.

# Test Case

-  `cardano-wallet-core:test:unit` - not sure where exactly because the test runner logs are buffered.

# Failure / Counter-example

```
    sequential puts replace values in order
      Checkpoint
        +++ OK, passed 100 tests (56% conflicting db entries).
      Wallet Metadata
        +++ OK, passed 100 tests (56% conflicting db entries).
      Tx History
        +++ OK, passed 100 tests (56% conflicting db entries).
      Private Key
        +++ OK, passed 100 tests (56% conflicting db entries).
    parallel puts replace values in _any_ order
      Checkpoint
        +++ OK, passed 10/nix/store/2w7m3lr8ldx8ipdll4j6fk4iid4irrck-stdenv-linux/setup: line 1303:    13 Killed                  /nix/store/gz1cbwbd4a5lvykbcm2z0irh1mldr8q1-cardano-wallet-core-test-unit-2020.8.3-x86_64-unknown-linux-musl/bin/unit
        14 Done                    | tee $out
builder for '/nix/store/wjkmhlj8f2l5bkiw0qswbh3qgrrakszh-cardano-wallet-core-test-unit-2020.8.3-x86_64-unknown-linux-musl-check-x86_64-unknown-linux-musl.drv' failed with exit code 137
```
(it happened twice on PR #2056 for example)

---

# Resolution 

1. Change cabal file so that tests are run with a memory limit. Then the failing test case will be known.
2. Reproduce failing test case in isolation.
3. Try reproducing with shrinking disabled on the failing test case.
   Error message is this:
   ```
   test/unit/Cardano/Wallet/DB/SqliteSpec.hs:243:9: 
   1) Cardano.Wallet.DB.Sqlite, Sqlite State machine tests, Sequential
        uncaught exception: SqliteException
        SQLite3 returned ErrorError while attempting to perform prepare "SELECT \"tx_id\", \"wallet_id\", \"status\", \"direction\", \"slot\", \"block_height\", \"amount\" FROM \"tx_meta\" WHERE (\"wallet_id\"=?) AND (\"tx_id\" IN (?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?)) AND (\"slot\"<=?) AND (\"status\"=?) ORDER BY \"slot\" DESC,\"tx_id\"": too many SQL variables
        (after 1389 tests)
   ```

4. Find the query with too many SQL variables and run it in chunks.
   The problem is `minWithdrawalFilter` in `selectTxHistory`.
   It might actually be better to replace the query with a rawSql join.

---

# QA 

- There should now be less random errors in the unit tests.

rvl:
### Issue Number

Resolves #2064.

### Overview

The DB state machine tests found an extreme case which caused a `too many SQL variables` exception. Unfortunately it was not able to shrink the counterexample without crashing.

- [x] Changes cabal file so that tests are run with a memory limit of 2G.
- [x] Fixes the test case which sometimes fails.

### Comments

- Currently testing with:
   ```
   nix-build -A tests.cardano-wallet-core.unit
   cd lib/core
   while ../../result/bin/unit --match '/Cardano.Wallet.DB.Sqlite/Sqlite State machine tests'; do :; done
   ```
- There is an infinite loop in `Sqlite State machine tests`, it gets stuck somewhere, then takes quite a while to exhaust the heap:
  ```
        There's no checkpoint older than k (+/- 100)
          +++ OK, passed 100 tests.
    Sqlite State machine tests
  1594/100unit: Heap exhausted;
  unit: Current maximum heap size is 2147483648 bytes (2048 MB).
  unit: Use `+RTS -M<size>' to increase it.
   173,889,316,944 bytes allocated in the heap
     9,858,194,528 bytes copied during GC
     2,112,669,264 bytes maximum residency (720 sample(s))
        40,633,672 bytes maximum slop
              2014 MB total memory in use (0 MB lost due to fragmentation)

                                       Tot time (elapsed)  Avg pause  Max pause
    Gen  0     166050 colls,     0 par    7.239s   7.241s     0.0000s    0.0007s
    Gen  1       720 colls,     0 par   2585.867s  2752.903s     3.8235s    6.0089s

    TASKS: 5 (1 bound, 4 peak workers (4 total), using -N1)

    SPARKS: 0(0 converted, 0 overflowed, 0 dud, 0 GC'd, 0 fizzled)

    INIT    time    0.000s  (  0.000s elapsed)
    MUT     time  100.141s  ( 99.704s elapsed)
    GC      time  2593.106s  (2760.143s elapsed)
    EXIT    time    0.002s  (  0.003s elapsed)
    Total   time  2693.249s  (2859.850s elapsed)

    Alloc rate    1,736,438,027 bytes per MUT second

    Productivity   3.7% of total user, 3.5% of total elapsed
  ```
- After disabling shrinking on the state machine tests, the infinite loop turns into a failing test (with quite a large counterexample).
- There is a `too many variables` exception when querying the `tx_meta` table.

### DIFF
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
@@ -123,12 +123,14 @@ import Control.Tracer
     ( Tracer, contramap, traceWith )
 import Data.Coerce
     ( coerce )
+import Data.Ord
+    ( Down (..) )
 import Data.Either
     ( isRight )
 import Data.Generics.Internal.VL.Lens
     ( (^.) )
 import Data.List
-    ( unzip3 )
+    ( sortOn, unzip3 )
 import Data.List.Split
     ( chunksOf )
 import Data.Map.Strict
@@ -1361,11 +1363,11 @@ selectTxs = fmap concatUnzip . mapM select . chunksOf chunkSize
     concatUnzip :: [([a], [b], [c])] -> ([a], [b], [c])
     concatUnzip = (\(a, b, c) -> (concat a, concat b, concat c)) . unzip3
 
-    -- Split a query's input values into chunks, run multiple smaller queries,
-    -- and then concatenate the results afterwards. Used to avoid "too many SQL
-    -- variables" errors for "SELECT IN" queries.
-    combineChunked :: [a] -> ([a] -> SqlPersistT IO [b]) -> SqlPersistT IO [b]
-    combineChunked xs f = concatMapM f $ chunksOf chunkSize xs
+-- | Split a query's input values into chunks, run multiple smaller queries,
+-- and then concatenate the results afterwards. Used to avoid "too many SQL
+-- variables" errors for "SELECT IN" queries.
+combineChunked :: [a] -> ([a] -> SqlPersistT IO [b]) -> SqlPersistT IO [b]
+combineChunked xs f = concatMapM f $ chunksOf chunkSize xs
 
 selectTxHistory
     :: TimeInterpreter IO
@@ -1378,21 +1380,24 @@ selectTxHistory ti wid minWithdrawal order conditions = do
     selectLatestCheckpoint wid >>= \case
         Nothing -> pure []
         Just cp -> do
-            minWithdrawalFilter <- case minWithdrawal of
-                Nothing  -> pure []
+            let txMetaFilter = (TxMetaWalletId ==. wid):conditions
+            metas <- case minWithdrawal of
+                Nothing -> fmap entityVal <$> selectList txMetaFilter sortOpt
                 Just inf -> do
                     let coin = W.Coin $ fromIntegral $ getQuantity inf
-                    wdrls <- fmap entityVal
-                        <$> selectList [ TxWithdrawalAmount >=. coin ] []
-                    pure [ TxMetaTxId <-. (txWithdrawalTxId <$> wdrls) ]
+                    let sortTxId = case order of
+                            W.Ascending -> Desc TxWithdrawalTxId
+                            W.Descending -> Asc TxWithdrawalTxId
+                    let sortSlot = case order of
+                            W.Ascending -> sortOn txMetaSlot
+                            W.Descending -> sortOn (Down . txMetaSlot)
+                    txids <- fmap (txWithdrawalTxId . entityVal)
+                        <$> selectList [ TxWithdrawalAmount >=. coin ] [sortTxId]
 
-            metas <- fmap entityVal <$> selectList
-                ( mconcat
-                    [ [ TxMetaWalletId ==. wid ]
-                    , minWithdrawalFilter
-                    , conditions
-                    ]
-                ) sortOpt
+                    ms <- combineChunked txids (\chunk -> selectList
+                       ((TxMetaTxId <-. chunk):txMetaFilter) sortOpt)
+
+                    pure $ sortSlot $ fmap entityVal ms
 
             let txids = map txMetaTxId metas
             (ins, outs, ws) <- selectTxs txids
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
@@ -193,6 +193,7 @@ import qualified Cardano.Wallet.Primitive.AddressDiscovery.Sequential as Seq
 import qualified Cardano.Wallet.Primitive.Model as W
 import qualified Cardano.Wallet.Primitive.Types as W
 import qualified Data.Map as Map
+import qualified Data.Set as Set
 import qualified Data.Text as T
 import qualified Database.Sqlite as Sqlite
 
@@ -1378,21 +1379,21 @@ selectTxHistory ti wid minWithdrawal order conditions = do
     selectLatestCheckpoint wid >>= \case
         Nothing -> pure []
         Just cp -> do
-            minWithdrawalFilter <- case minWithdrawal of
-                Nothing  -> pure []
+            minWithdrawalPred <- case minWithdrawal of
+                Nothing -> pure (const True)
                 Just inf -> do
+                    -- Tests whether a given TxMeta matches a reward withdrawal
+                    -- above the given minimum. This filtering is done outside
+                    -- of sqlite because joins are not possible with persistent.
                     let coin = W.Coin $ fromIntegral $ getQuantity inf
-                    wdrls <- fmap entityVal
-                        <$> selectList [ TxWithdrawalAmount >=. coin ] []
-                    pure [ TxMetaTxId <-. (txWithdrawalTxId <$> wdrls) ]
+                    wdrls <- fmap entityVal <$> selectList
+                        [ TxWithdrawalAmount >=. coin ]
+                        [ Asc TxWithdrawalTxId ]
+                    let txids = Set.fromAscList $ map txWithdrawalTxId wdrls
+                    pure ((`Set.member` txids) . txMetaTxId)
 
-            metas <- fmap entityVal <$> selectList
-                ( mconcat
-                    [ [ TxMetaWalletId ==. wid ]
-                    , minWithdrawalFilter
-                    , conditions
-                    ]
-                ) sortOpt
+            metas <- filter minWithdrawalPred . fmap entityVal
+                <$> selectList ((TxMetaWalletId ==. wid):conditions) sortOpt
 
             let txids = map txMetaTxId metas
             (ins, outs, ws) <- selectTxs txids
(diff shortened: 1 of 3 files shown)


jonathanknowles:
> 1594/100unit: Heap exhausted;
unit: Current maximum heap size is 2147483648 bytes (2048 MB).
unit: Use `+RTS -M<size>' to increase it.

I think this is nice, as it makes it extra clear that we've exhausted the heap. :+1: 

(Much clearer than getting randomly killed by the OOM killer.)

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1378,21 +1379,21 @@ selectTxHistory ti wid minWithdrawal order conditions = do
...
-                    ]
-                ) sortOpt
+                    wdrls <- fmap entityVal <$> selectList
+                        [ TxWithdrawalAmount >=. coin ]
+                        [ Asc TxWithdrawalTxId ]
+                    let txids = Set.fromAscList $ map txWithdrawalTxId wdrls
+                    pure ((`Set.member` txids) . txMetaTxId)
+
+            metas <- filter minWithdrawalPred . fmap entityVal
+                <$> selectList ((TxMetaWalletId ==. wid):conditions) sortOpt
```
KtorZ: Erf. This means that requests like `/transactions?min_withdrawal=1` will have to load the entire transaction history in memory :/  
The case reached by the state-machine is probably a case where there are more than 1000 withdrawals in the wallet right :thinking: ? So we could re-use a similar approach to what was done for fetching inputs / outputs:

```hs
        resolvedInputs <- toOutputMap . fmap entityVal <$>
            combineChunked inputs (\inputsChunk -> selectList
                [TxOutputTxId <-. (txInputSourceTxId <$> inputsChunk)]
                [Asc TxOutputTxId, Asc TxOutputIndex])
```

where `inputs` here would be `txids`. So we can fetch all transactions from the list of `txids` that match the extra conditions, and do it chunk by chunk.

rvl: Yes - not ideal. If using `combineChunked` we also need to sort again after concatenating chunks. But that would be less bad.

KtorZ: I expect most wallets to have a fairly limited amount of withdrawals. And, in case one has many, fetching all transactions first would be way worse.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1378,21 +1380,24 @@ selectTxHistory ti wid minWithdrawal order conditions = do
...
-            metas <- fmap entityVal <$> selectList
-                ( mconcat
-                    [ [ TxMetaWalletId ==. wid ]
-                    , minWithdrawalFilter
-                    , conditions
-                    ]
-                ) sortOpt
+                    let sortTxId = case order of
+                            W.Ascending -> Desc TxWithdrawalTxId
+                            W.Descending -> Asc TxWithdrawalTxId
```
KtorZ: I am not sure to get the rationale here, surely the extra ordering of the txid has nothing to do with the slot order? So we could pick an arbitrary direction regardless of `order`'s value, right ?

KtorZ (approved):
I've carried with the PR, fixed the shrinker and the withdrawal fetching which was still causing an issue (see commit)

PR #2078, opened 2020-08-26
Test the `removeRetiredPools` function.

jonathanknowles:
## Issue Number

#2018 

## Overview

This PR adds a simple property test to check that the `removeRetiredPools` function only removes pools that the DB layer has indicated as having retired (and does nothing else).

## Assumptions

This property assumes the correctness of the following functions:
| Function | Tested By |
| -- | -- |
| [**`listRetiredPools`**](https://github.com/input-output-hk/cardano-wallet/blob/9f93d0a98be7c19035dd603149ebf06ac38d7f2f/lib/core/src/Cardano/Pool/DB.hs#L182) | [`prop_listRetiredPools_multiplePools_multipleCerts`](https://github.com/input-output-hk/cardano-wallet/blob/9f93d0a98be7c19035dd603149ebf06ac38d7f2f/lib/core/test/unit/Cardano/Pool/DB/Properties.hs#L976)<br>[`prop_multiple_putPoolRegistration_single_readPoolRegistration`](https://github.com/input-output-hk/cardano-wallet/blob/9f93d0a98be7c19035dd603149ebf06ac38d7f2f/lib/core/test/unit/Cardano/Pool/DB/Properties.hs#L540)<br>[`prop_multiple_putPoolRetirement_single_readPoolRetirement`](https://github.com/input-output-hk/cardano-wallet/blob/9f93d0a98be7c19035dd603149ebf06ac38d7f2f/lib/core/test/unit/Cardano/Pool/DB/Properties.hs#L594)<br>[`prop_poolRegistration`](https://github.com/input-output-hk/cardano-wallet/blob/9f93d0a98be7c19035dd603149ebf06ac38d7f2f/lib/core/test/unit/Cardano/Pool/DB/Properties.hs#L475)<br>[`prop_poolRetirement`](https://github.com/input-output-hk/cardano-wallet/blob/9f93d0a98be7c19035dd603149ebf06ac38d7f2f/lib/core/test/unit/Cardano/Pool/DB/Properties.hs#L505) |
| [**`removePools`**](https://github.com/input-output-hk/cardano-wallet/blob/9f93d0a98be7c19035dd603149ebf06ac38d7f2f/lib/core/src/Cardano/Pool/DB.hs#L209) | [`prop_removePools`](https://github.com/input-output-hk/cardano-wallet/blob/9f93d0a98be7c19035dd603149ebf06ac38d7f2f/lib/core/test/unit/Cardano/Pool/DB/Properties.hs#L861) |

## Comments

This PR relies on a mock pool `DBLayer` type, based on a _simplified_ version of the original `DBLayer` type without existential types. This makes it possible to selectively override member functions with standard record update syntax.

See https://gitlab.haskell.org/ghc/ghc/-/issues/16501.

### DIFF
--- a/lib/core/test/unit/Cardano/Pool/DB/Mock.hs
+++ b/lib/core/test/unit/Cardano/Pool/DB/Mock.hs
@@ -1,0 +1,174 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE RecordWildCards #-}
+
+module Cardano.Pool.DB.Mock
+    ( DBLayer (..)
+    , dbLayer
+    , toReal
+    ) where
+
+import Prelude hiding
+    ( fail )
+
+import Cardano.Pool.DB
+    ( ErrPointAlreadyExists (..) )
+import Cardano.Wallet.Primitive.Types
+    ( BlockHeader (..)
+    , CertificatePublicationTime (..)
+    , EpochNo (..)
+    , PoolId
+    , PoolLifeCycleStatus (..)
+    , PoolRegistrationCertificate (..)
+    , PoolRetirementCertificate (..)
+    , SlotNo (..)
+    , StakePoolMetadata
+    , StakePoolMetadataHash
+    , StakePoolMetadataUrl
+    )
+import Control.Monad.Fail
+    ( MonadFail )
+import Control.Monad.IO.Class
+    ( MonadIO )
+import Control.Monad.Trans.Except
+    ( ExceptT )
+import Data.Map.Strict
+    ( Map )
+import Data.Quantity
+    ( Quantity (..) )
+import Data.Word
+    ( Word64 )
+import System.Random
+    ( StdGen )
+
+import qualified Cardano.Pool.DB as Real
+
+-- | A simplified version of 'Real.DBLayer' without existential types.
+--
+-- Since this type does not use existential quantification, member functions
+-- can be overridden with the standard record update syntax.
+--
+data DBLayer m = DBLayer
+    { putPoolProduction
+        :: BlockHeader
+        -> PoolId
+        -> ExceptT ErrPointAlreadyExists m ()
+    , readPoolProduction
+        :: EpochNo
+        -> m (Map PoolId [BlockHeader])
+    , readTotalProduction
+        :: m (Map PoolId (Quantity "block" Word64))
+    , putStakeDistribution
+        :: EpochNo
+        -> [(PoolId, Quantity "lovelace" Word64)]
+        -> m ()
+    , readStakeDistribution
+        :: EpochNo
+        -> m [(PoolId, Quantity "lovelace" Word64)]
+    , readPoolProductionCursor
+        :: Int
+        -> m [BlockHeader]
+    , readPoolLifeCycleStatus
+        :: PoolId
+        -> m PoolLifeCycleStatus
+    , putPoolRegistration
+        :: CertificatePublicationTime
+        -> PoolRegistrationCertificate
+        -> m ()
+    , readPoolRegistration
+        :: PoolId
+        -> m (Maybe (CertificatePublicationTime, PoolRegistrationCertificate))
+    , putPoolRetirement
+        :: CertificatePublicationTime
+        -> PoolRetirementCertificate
+        -> m ()
+    , readPoolRetirement
+        :: PoolId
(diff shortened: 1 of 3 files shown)


piotr-iohk (approved):
:ok_hand: (I guess :sweat_smile: )

lib/core/test/unit/Cardano/Pool/DB/Mock.hs
```diff
@@ -0,0 +1,184 @@
...
+  where
+    fail op = error $ "mock DB layer: unexpected call to: " <> op
+
+-- | Converts a 'DBLayer' into a 'Real.DBLayer'.
+--
+toReal
+    :: (MonadFail m, MonadIO m)
+    => DBLayer m
+    -> Real.DBLayer m
+toReal DBLayer {..} = Real.DBLayer {..}
```
Anviking, line 182: Was going to say that defining a duplicate `DBLayer` seemed a bit overkill, but this `RecordWildCards` line is pretty neat!

Not 100% convinced, but keeping it up-to-date doesn't look _that_ annoying.

jonathanknowles: > Was going to say that defining a duplicate DBLayer seemed a bit overkill

I was also reluctant to use this approach at first, but unfortunately it's not possible to use record update syntax with record types that use existential quantification. See: https://gitlab.haskell.org/ghc/ghc/-/issues/16501

That creates a problem if we want to make a mock in which only a subset of the functions are defined.

So the option I've used here is to define a simpler type where the record fields _can_ be updated, and then map values of that type back to the original type that uses existentials.

>  but this RecordWildCards line is pretty neat!

Thanks!

IMO, the advantage of using this method is that the compiler will tell us if we need to update the `Mock.DBLayer` type.

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
```diff
@@ -937,6 +945,28 @@ prop_removePools
         fmap (Set.fromList . fmap (view #poolId . snd) . catMaybes)
             <$> atomically $ mapM readPoolRetirement $ Set.toList pools
 
+-- Tests that the 'removeRetiredPools' operation does not do anything other
+-- than removing the pools that the DB layer indicates as having retired.
+--
+prop_removeRetiredPools
```
Anviking, line 951: Isn’t the prop_removeRetiredPools essentially just re-iterating the definition of removeRetiredPools though?

Wouldn't we be as fine without it?

jonathanknowles: > Isn’t the prop_removeRetiredPools essentially just re-iterating the definition of removeRetiredPools though? Wouldn't we be as fine without it?

It's admittedly a very simple mock, in this case.

It also checks that `removeRetiredPools` doesn't do anything else, as calling other operations from `DBLayer` will raise an error.

Ultimately, I think we do want to avoid re-testing `listRetiredPools` and `removePools`, as they are tested elsewhere.

Anviking (approved):
As someone who didn't follow the implementation of this story closely, I'd be more interested in knowing e.g. that

```haskell
forall state time. listing pools with garbage collection == listing pools without garbage collection”
```
or 
```
the number of pools we store in db ≈ the number of active pools, even if lots are retired
```
of course these are very vague properties, and I writing them as part of this story might very well be out of scope as you said, if they even are possible to write. (I think they _are_ possible though)

I'm a bit ambivalent about testing that `removeRetiredPools` only… calls `removePools` using `listRetiredPools`, because it seems like it is both very expected and re-iterating the implementation, with the cost of some boilerplate / noise.

> It's not clear to me how we can avoid this. Perhaps you can think of a better way to test this function?

I would be fine with simply not testing this. (Or testing manually / indirectly through existing integration scenarios)

But I suppose I am also fine with this PR as is, just as long as you do think it is useful.

jonathanknowles:
Moving this PR back to draft status for the moment.

I think it's more important to get https://github.com/input-output-hk/cardano-wallet/pull/2057 finalized first, and then to look at this again later on.

I'm also giving some more thought to the arguments that @Anviking raised, that there may be a better way to test this function.

jonathanknowles:
Closing for now. The value gained from testing this function doesn't appear to be worth it compared with other improvements we might make to the test suite.

Issue #2077, opened 2020-08-26
Tx metadata extracted from binary blocks


rvl:
# Context

Metadata must be serialized as CBOR in transactions. However, we should input and output JSON at the system boundary, and convert between JSON/CBOR internally.

### User story ADP-307

> As an API user,
> I want to add metadata to a transaction and view them as JSON,
> So that I can embed custom information within a transaction.

### References

See Appendix E of [Delegation Design Spec](https://hydra.iohk.io/job/Cardano/cardano-ledger-specs/delegationDesignSpec/latest/download-by-type/doc-pdf/delegation_design_spec)

# Decision

There's an existing implementation in cardano-api for doing JSON/CBOR conversions. We can hopefully re-use this. See: https://github.com/input-output-hk/cardano-node/blob/495366939621d4a413d000494ec01a8b0f5edc26/cardano-api/src/Cardano/Api/MetaData.hs

Although it is possible to translate any JSON into CBOR, the reciprocal isn't true. There are therefore arbitrary choices made regarding the transformation JSON/CBOR to make sure it can roundtrip under several assumptions.

# AC

- cardano-wallet must show metadata in transaction that have some (as part of the response object).
- metadata must be shown as JSON format in API responses
- cardano-wallet-jormungandr should default to null values and fail if provided with a metadata object.

---

# Development 

1. Extend `TxMeta` type to include metadata as an aeson `Value`.
2. When getting transactions from a block, use the existing conversion functions to make `Value` from metadata.

# QA 

- I have no actual chain data to use, so there are no golden tests.
- integration test _TRANS_CREATE_10 - Transaction with metadata_ in `lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs` ⇒  #2096.

KtorZ:
> Extend TxMeta type to include metadata as an aeson Value.

@rvl from quickly looking at `cardano-api`, it seems that they provide a `Metadatum` data-type as an intermediate Haskell representation and that transformation are done from this Metadatum type (for JSON & CBOR). Perhaps re-using this data-type is preferable over an opaque aeson Value ? Thoughts?

rvl:
Yes it could be better actually because it helps with validating new transactions.

Also I am thinking it's better to have TxMetadata inside the Tx type rather than TxMeta.

piotr-iohk:
lgtm.

PR #2079, opened 2020-08-26
Get TxMetadata from blocks on chain

rvl:
### Issue Number

ADP-307 / #2077

### Overview

- Initial types, etc.

### Comments

- I don't have any serialized block data to make golden tests. But the testing can be done with integration tests.


lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -819,18 +824,23 @@ data Tx = Tx
         :: !(Map ChimericAccount Coin)
         -- ^ Withdrawals (of funds from a registered reward account) embedded in
         -- a transaction. The order does not matter.
+    , metadata
+        :: !(Maybe TxMetadata)
+        -- ^ Semi-structured application-specific extension data stored in the
+        -- transaction on chain.
```
KtorZ, line 853: :+1:

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -892,6 +902,25 @@ instance Buildable TxMeta where
...
+--
+-- fixme: TxMeta / TxMetadata naming is rather confusing.
+newtype TxMetadata = TxMetadata { unTxMetadata :: Value }
+  deriving (Show, Eq, Generic)
+
+instance NFData TxMetadata
+
+-- Compare metadatas by json encoding
+instance Ord TxMetadata where
+    compare = comparing (Aeson.encode . unTxMetadata)
```
KtorZ: What's the rationale here ? Why not rely on the existing `Ord` instance for `Value` ?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -555,6 +558,9 @@ data ApiTransaction (n :: NetworkDiscriminant) = ApiTransaction
     , outputs :: ![AddressAmount (ApiT Address, Proxy n)]
     , withdrawals :: ![ApiWithdrawal n]
     , status :: !(ApiT TxStatus)
+    -- fixme: #2073 not sure whether empty metadata should be null or missing
+    -- key.
+    , metadata :: !(Maybe (ApiT TxMetadata))
```
KtorZ: My own preference is to have a `null` value so that it is more explicit client side but I admit not being 100% sure about how the default aeson generic options treat this in our API at the moment.

cardano-1.19.x.yaml
```diff
@@ -60,7 +60,7 @@ packages:
 - Win32-2.6.2.0
 
 - git: https://github.com/input-output-hk/cardano-base
-  commit: df8687488449f71dce3d881800c21e41fe1b7fc1
+  commit: 3e2d92b25c675fd923a07237eb1d0859b29a79d3
```
KtorZ: Damn. This completely breaks the linker when building with stack for me. I am trying on a completely clean stack root now because It reminds me of a past issue with stack not correctly figuring out links between same packages of different snapshots.

rvl: Yeah I think the YAML filename needs to be different so that stack knows that something's changed. But I couldn't think of a good name for a version somewhere between 1.19.0 and 1.20.0.

lib/core/src/Cardano/Wallet/Api/Types.hs
rvl: That is also my preference actually. But when submitting a transaction, it is more convenient for users if they can omit the key. Our default generic options are to omit Nothing fields.

cardano-1.19.x.yaml
rvl: I needed to bump this library so that the `shelley-spec-ledger-test` package would build. This was just to get access to the `TxMetadata` test generators. But that has caused more problems with multiple `Arbitrary` instances of other types. It's just not worth it. I will copy&paste the `Arbitrary` instance.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
rvl: `Value` has no `Ord` ... nor does `Cardano.Api.TxMetadata`.
I define it by string encoding so that other types can keep their `Ord` constraint.

lib/core/src/Cardano/Wallet/Api/Types.hs
rvl: Will sort it out in #2089.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -843,18 +847,27 @@ data Tx = Tx
         :: !(Map ChimericAccount Coin)
         -- ^ Withdrawals (of funds from a registered reward account) embedded in
         -- a transaction. The order does not matter.
+    , metadata
+        :: !(Maybe TxMetadata)
+        -- ^ Semi-structured application-specific extension data stored in the
+        -- transaction on chain.
+        --
+        -- This is not to be confused with 'TxMeta', which is information about
+        -- a transaction derived from the ledger.
```
KtorZ, line 856: yeah, we might want to rename `TxMeta` at some point indeed.

lib/core/test/unit/Cardano/Wallet/Primitive/TypesSpec.hs
```diff
@@ -260,6 +265,34 @@ spec = do
...
+                , "    - 65"
+                , "element 20:"
+                , "  - key: \"key\""
+                , "  - val:"
+                , "      list:"
+                , "        - \"v0\""
+                , "        - \"v1\""
+                , "  - key: 0"
+                , "  - val: \"value\""
+                ]
```
KtorZ, line 294: :+1:

KtorZ (approved):
OK

rvl:
Since we have added shelley libs as dependencies to the core library, all executables now link with libsodium. So the nix needed to be updated so that `libsodium-23.dll` was copied for all windows exes.

bors r+

PR #2081, opened 2020-08-26
Extend restoration benchmarks to include large random wallets.

KtorZ:
# Issue Number

#2032 

# Overview

- [ ] I have defined a new special wallet scheme analogous to the any% scheme used until now, but built on top of the `RndState`, so that we still get to store and retrieve the address space too.

- [ ] I have removed the 2% case and "replaced" it with a 1% case on large random wallets.

# Comments

Tested on the testnet, which works fine. The reason to add this one is to be able to create arbitrarily large random wallets that "make sense", and then measure the time taken for other specific operations such as listing addresses or estimating fees (which I'll do as a separate PR).

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Random.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Random.hs
@@ -34,6 +34,10 @@ module Cardano.Wallet.Primitive.AddressDiscovery.Random
     , deriveRndStateAddress
     , findUnusedPath
     , unavailablePaths
+
+    -- ** Benchmarking
+    , RndAnyState (..)
+    , mkRndAnyState
     ) where
 import Prelude
 
@@ -66,16 +70,24 @@ import Control.DeepSeq
     ( NFData (..) )
 import Control.Monad
     ( join )
+import Data.Digest.CRC32
+    ( crc32 )
 import Data.Map
     ( Map )
 import Data.Maybe
     ( isJust )
+import Data.Proxy
+    ( Proxy (..) )
 import Data.Set
     ( Set )
+import Data.Word
+    ( Word32 )
 import Fmt
     ( Buildable (..), blockMapF', indentF, tupleF )
 import GHC.Generics
     ( Generic )
+import GHC.TypeLits
+    ( KnownNat, Nat, natVal )
 import System.Random
     ( RandomGen, StdGen, mkStdGen, randomR )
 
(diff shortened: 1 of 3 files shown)


lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Random.hs
```diff
@@ -268,3 +280,90 @@ instance CompareDiscovery (RndState n) where
...
+    { innerState :: RndState network
+    } deriving (Generic, Show)
+
+instance NFData (RndAnyState n p)
+
+-- | Initialize the HD random address discovery state from a root key and RNG
+-- seed.
+--
+-- The first argument is expected to be a ratio (between 0 and 1) of addresses
+-- we ought to simply recognize as ours. So, giving .5 means that 50% of the
```
Anviking: Seems outdated. Should be `ratio (between 0% and 100%)` and `giving @50 means 50% of the` I presume?

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Random.hs
```diff
@@ -268,3 +280,90 @@ instance CompareDiscovery (RndState n) where
...
     knownAddresses s = Map.elems (addresses s)
+
+--------------------------------------------------------------------------------
+--
+-- RndAnyState
+--
+-- For benchmarking and testing arbitrary large random wallets.
+
+-- | An "unsound" alternative that can be used for benchmarking and stress
+-- testing. It re-uses the same underlying structure as the `RndState` but
```
Anviking, line 291: > same underlying structure

Is there a main difference, say in complexity, between `RndAnyState` and `AnyAddressState`?

Anviking: Would this be true?
```
RndAnyState, like RndState, grow with the number of addresses used. This is unlike SeqState/AnyAddressState, where which has a bounded gap of user addresses it keeps in memory
```
If not, something like it, would be cool to understand what makes `RndAnyState` indicative of performance of random wallets, and why it's not enough to have `AnyAddressState`.

lib/shelley/bench/Restore.hs
```diff
@@ -195,43 +196,45 @@ cardanoRestoreBench tr c socketFile = do
...
-                "2-percent.timelog"
-                (initAnyState "Benchmark 2% Wallet" 0.02))
+        , bench ("restore " <> network <> " 1% naked")
+             (bench_restoration @_ @ShelleyKey
+                 networkProxy
+                 tr
+                 socketFile
+                 np
+                 vData
+                 "1-percent-naked.timelog"
```
Anviking: Maybe just `1-percent.timelog` would be easier

lib/shelley/bench/Restore.hs
```diff
@@ -195,43 +196,45 @@ cardanoRestoreBench tr c socketFile = do
...
                 tr
                 socketFile
                 np
                 vData
-                "1-percent.timelog"
-                (initAnyState "Benchmark 1% Wallet" 0.01))
+                "1-percent-rnd.timelog"
+                (walletRnd $ mkRndAnyState @1))
 
-        , bench ("restore " <> network <> " 2% ownership")
```
Anviking, line 210: Removing because it was too slow, I presume?

Maybe it's worth adding `0.1%`, or `0.5%`, to still have more data points than 0% and 1%.

(Useful, I imagine, if you'd ever want to do some curve fitting on the data)

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Random.hs
Anviking: Re my earlier comment:

Ah, I'm blind 😅 
```haskell
innerState :: RndState network
```
"same underlying structure" makes sense then

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Random.hs
KtorZ: Right, I had it as an argument initially and switched to a type-level parameter later on. 

> giving @50 means 50% of the I presume?

Indeed.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Random.hs
KtorZ: :) ... It still stores the UTxO and transaction history, but the `AnyState` do not store the address space. BUT, we've observed some major slowness when it comes to storing addresses, so I find it quite important to also include that in the benchmark. 

I think that we could have a similar `SeqAnyState` which also include the seq state, and get rid of the `AnyState` altogether.

lib/shelley/bench/Restore.hs
KtorZ: > Removing because it was too slow, I presume?

Correct.

> Maybe it's worth adding 0.1%, or 0.5%, to still have more data points than 0% and 1%.

Good idea :+1:

KtorZ:
@Anviking added .1% and .5% benchmarks + fix comment to correctly mention the type parameter. 

https://github.com/input-output-hk/cardano-wallet/compare/f0a4bf9b25f9d1801fbd1a67b6b294f03096b451..580bd968daa9607431dd80c8d6d3ca7f0b334f68

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Random.hs
```diff
@@ -268,3 +280,90 @@ instance CompareDiscovery (RndState n) where
...
+instance KnownNat p => IsOurs (RndAnyState n p) Address where
+    isOurs addr@(Address bytes) st@(RndAnyState inner) =
+        case isOurs addr inner of
+            (True, inner') ->
+                (True, RndAnyState inner')
+
+            (False, _) | crc32 bytes < p ->
+                let
+                    (path, gen') = findUnusedPath
+                        (gen inner) (accountIndex inner) (unavailablePaths inner)
```
Anviking, line 334: Couldn't this extra `findUnusedPath` call, compared to the inner state implementation, affect the results?

Anviking (approved):

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Random.hs
KtorZ: A bit yes. We don't benchmark this function independently but I believe that even on large wallets with 100K+ addresses, finding an unused path in a space of 2^31 is quite fast.

KtorZ: So I'd expect it to be quite negligible.

PR #2085, opened 2020-08-27
[ADP-416] Allow multiple well-formed values in "/Cardano.Wallet.Api/"…

hasufell:
Related: https://github.com/input-output-hk/cardano-wallet/pull/2055#issuecomment-681954272

[Test log](https://gist.github.com/hasufell/c07ead1a20824bea6c20b2dc9b315bd4) shows that it tries both variants of pool id encodings when combined with #2055

### DIFF
--- a/lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
@@ -117,7 +117,7 @@ newtype Header (headerName :: Symbol) (contentType :: *) =
 --
 
 class Wellformed t where
-    wellformed :: t
+    wellformed :: [t]
 
 class Malformed t where
     malformed :: [(t, ExpectedError)]
@@ -126,7 +126,7 @@ class Malformed t where
 -- Class instances (PathParam)
 --
 instance Wellformed (PathParam (ApiT WalletId)) where
-    wellformed = PathParam $ T.replicate 40 "0"
+    wellformed = [PathParam $ T.replicate 40 "0"]
 
 instance Malformed (PathParam (ApiT WalletId)) where
     malformed = first PathParam <$>
@@ -138,7 +138,7 @@ instance Malformed (PathParam (ApiT WalletId)) where
         msg = "wallet id should be a hex-encoded string of 40 characters"
 
 instance Wellformed (PathParam ApiTxId) where
-    wellformed = PathParam $ T.replicate 64 "0"
+    wellformed = [PathParam $ T.replicate 64 "0"]
 
 instance Malformed (PathParam ApiTxId) where
     malformed = first PathParam <$>
@@ -150,7 +150,9 @@ instance Malformed (PathParam ApiTxId) where
         msg = "Invalid tx hash: expecting a hex-encoded value that is 32 bytes in length."
 
 instance Wellformed (PathParam ApiPoolId) where
-    wellformed = PathParam $ T.replicate 64 "0"
+    wellformed = PathParam <$>
+        [ T.replicate 64 "0"
+        ]
 
 instance Malformed (PathParam ApiPoolId) where
     malformed = first PathParam <$>
@@ -162,8 +164,8 @@ instance Malformed (PathParam ApiPoolId) where
         msg = "Invalid stake pool id: expecting a hex-encoded value that is 28 or 32 bytes in length."
 
 instance Wellformed (PathParam (ApiT Address, Proxy ('Testnet 0))) where
-    wellformed = PathParam
-        "FHnt4NL7yPY7JbfJYSadQVSGJG7EKkN4kpVJMhJ8CN3uDNymGnJuuwcHmyP4ouZ"
+    wellformed = [PathParam
+        "FHnt4NL7yPY7JbfJYSadQVSGJG7EKkN4kpVJMhJ8CN3uDNymGnJuuwcHmyP4ouZ"]
 
 instance Malformed (PathParam (ApiT Address, Proxy ('Testnet 0))) where
     malformed = []
@@ -994,7 +996,7 @@ instance Malformed (BodyParam ApiPostRandomAddressData) where
 --
 instance Wellformed (Header "Content-Type" JSON) where
     wellformed =
-        Header "application/json"
+        [Header "application/json"]
 
 instance Malformed (Header "Content-Type" JSON) where
     malformed = first Header <$>
@@ -1005,7 +1007,7 @@ instance Malformed (Header "Content-Type" JSON) where
 
 instance Wellformed (Header "Content-Type" OctetStream) where
     wellformed =
-        Header "application/octet-stream"
+        [Header "application/octet-stream"]
 
 instance Malformed (Header "Content-Type" OctetStream) where
     malformed = first Header <$>
@@ -1016,7 +1018,7 @@ instance Malformed (Header "Content-Type" OctetStream) where
 
 instance Wellformed (Header "Accept" JSON) where
     wellformed =
-        Header "application/json"
+        [Header "application/json"]
 
 instance Malformed (Header "Accept" JSON) where
     malformed = first Header <$>
--- a/lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
+++ b/lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
@@ -72,7 +72,7 @@ import Data.Function
 import Data.IORef
     ( atomicModifyIORef, newIORef )
 import Data.List
-    ( (\\) )
+    ( foldl', (\\) )
 import Data.Map.Strict
     ( Map )
 import Data.Maybe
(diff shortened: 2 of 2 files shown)


lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
```diff
@@ -150,7 +150,9 @@ instance Malformed (PathParam ApiTxId) where
         msg = "Invalid tx hash: expecting a hex-encoded value that is 32 bytes in length."
 
 instance Wellformed (PathParam ApiPoolId) where
-    wellformed = PathParam $ T.replicate 64 "0"
+    wellformed = PathParam <$>
+        [ T.replicate 64 "0"
```
hasufell, line 154: This is where the bech32 wellformed would go

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
```diff
@@ -251,25 +257,32 @@ instance
         traverseLeft fn xs =
             fmap swap <$> traverse (traverse fn) (swap <$> xs)
 
+        insideOut :: [([x], y)] -> [(x, y)]
+        insideOut = foldl' (\y (xs, b) -> y ++ fmap (, b) xs) []
```
jonathanknowles: **Suggestion**: Use ordinary list comprehension syntax.

**Justification**: Shorter and more readable. (Obviously this is subjective.)

```suggestion
        insideOut zs = [(x, y) | (xs, y) <- zs, x <- xs]
```

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
```diff
@@ -417,63 +430,64 @@ instance
...
+
 
 instance
     ( Wellformed (PathParam t)
     , GEveryEndpoints sub
     ) => GEveryEndpoints (Capture p t :> sub)
   where
     gEveryEndpoint _ =
-        addPathFragment t <$> gEveryEndpoint (Proxy @sub)
+        concat $ fmap (\t -> addPathFragment t <$> gEveryEndpoint (Proxy @sub)) ts
```
jonathanknowles: Suggestion: use `concatMap` here.

```suggestion
        concatMap (\t -> addPathFragment t <$> gEveryEndpoint (Proxy @sub)) ts
```

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
```diff
@@ -591,3 +605,4 @@ titleize proxy req = unwords
     ]
   where
     proxyStr = "(" <> show (typeRep proxy) <> ")"
+
```
jonathanknowles: Request: remove extraneous line from end of file.

Perhaps your editor is inserting an extra CR at the end of the file? :space_invader:

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
```diff
@@ -251,25 +257,32 @@ instance
...
     , Malformed (Header h ct)
-    ) => GenericApiSpec (Header h ct -> Request)
+    ) => GenericApiSpec (Header h ct -> [Request])
   where
-    gSpec toRequest toSpec = toSpec $
-        SomeTest (Proxy @h) (first toRequest <$> malformed @(Header h ct))
+    gSpec toRequest toSpec = do
+        let tests = fmap (\(xs, e) -> fmap (,e) xs)
+                (first toRequest <$> malformed @(Header h ct))
+        forM_ tests $ (toSpec . SomeTest (Proxy @h))
```
jonathanknowles: Suggestion: Remove extraneous `$`. (I think it's unnecessary here.)

```suggestion
        forM_ tests (toSpec . SomeTest (Proxy @h))
```

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
```diff
@@ -218,31 +218,37 @@ instance (GenericApiSpec a, GenericApiSpec b) => GenericApiSpec (a :<|> b) where
...
     ( Typeable a, Malformed (PathParam a)
-    ) => GenericApiSpec (PathParam a -> Request)
+    ) => GenericApiSpec (PathParam a -> [Request])
   where
-    gSpec toRequest toSpec = toSpec $
-        SomeTest (Proxy @a) (first toRequest <$> malformed @(PathParam a))
+    gSpec toRequest toSpec = do
+        let tests = fmap (\(xs, e) -> fmap (,e) xs)
+                (first toRequest <$> malformed @(PathParam a))
+        forM_ tests $ (toSpec . SomeTest (Proxy @a))
```
jonathanknowles: Suggestion: Remove the extraneous `$`. (I think it's unnecessary here, given that you already have the parentheses.)

```suggestion
        forM_ tests (toSpec . SomeTest (Proxy @a))
```

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
```diff
@@ -218,31 +218,37 @@ instance (GenericApiSpec a, GenericApiSpec b) => GenericApiSpec (a :<|> b) where
...
     ( Typeable a, Malformed (BodyParam a)
-    ) => GenericApiSpec (BodyParam a -> IO Request)
+    ) => GenericApiSpec (BodyParam a -> IO [Request])
   where
     gSpec toRequest toSpec = do
-        let tests = first toRequest <$> malformed @(BodyParam a)
-        toSpec . SomeTest (Proxy @a) =<< traverseLeft runIO tests
+        let tests :: [(IO [Request], ExpectedError)]
+            tests = first toRequest <$> malformed @(BodyParam a)
+        toSpec . SomeTest (Proxy @a) =<< fmap insideOut (traverseLeft runIO tests)
```
jonathanknowles: Suggestion: Remove unnecessary `fmap`.

```suggestion
        toSpec . SomeTest (Proxy @a) . insideOut =<< traverseLeft runIO tests
```

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
```diff
@@ -218,31 +218,37 @@ instance (GenericApiSpec a, GenericApiSpec b) => GenericApiSpec (a :<|> b) where
...
 instance
     ( Typeable a, Malformed (PathParam a)
-    ) => GenericApiSpec (PathParam a -> Request)
+    ) => GenericApiSpec (PathParam a -> [Request])
   where
-    gSpec toRequest toSpec = toSpec $
-        SomeTest (Proxy @a) (first toRequest <$> malformed @(PathParam a))
+    gSpec toRequest toSpec = do
+        let tests = fmap (\(xs, e) -> fmap (,e) xs)
+                (first toRequest <$> malformed @(PathParam a))
```
jonathanknowles, line 231: Perhaps it's just me, but I do have a bit of trouble imagining what is happening here.

I wonder if the code could be made simpler with a nested `forM_`?

jonathanknowles (changes requested):
Looks pretty good so far. :+1: 

Just a couple of requested changes, and a few suggestions. (See comments!)

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
hasufell: I don't think that's even the same code. `foldl'` is strict in the accumulator. List comprehension desugars to something monadic, which I believe uses `concat` in the end, which uses `foldr`. So that may be optimised away due to list fusion... or not.

I also believe there should be some freedom to choose which constructs to use. I don't find list comprehension more readable.

Edit: on second thought... I think foldr might be the better choice.

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
hasufell: This code is type tetris... I doubt changing the functions will make it any more accessible. We can add more type signatures though.

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
```diff
@@ -417,63 +431,64 @@ instance
...
     gEveryBodyParam _ req =
-        req { requestMethod = reflectMethod (Proxy @m) }
+        [req { requestMethod = reflectMethod (Proxy @m) }]
 
-    type MkHeaderRequest (NoContentVerb m) = Request
+    type MkHeaderRequest (NoContentVerb m) = [Request]
     gEveryHeader _ req =
-        req { requestMethod = reflectMethod (Proxy @m) }
+        [req { requestMethod = reflectMethod (Proxy @m) }]
+
```
jonathanknowles:
```suggestion

```
https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#use-only-a-single-blank-line-between-top-level-definitions

jonathanknowles (approved):
Hi @hasufell 

This looks good to me. Thanks for addressing the previous feedback! (See comments.) :+1: 

I suspect @KtorZ will also have an opinion on this approach (judging by https://github.com/input-output-hk/cardano-wallet/pull/2055#issuecomment-681990632), so it would be good to get confirmation from him, if you can!

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
```diff
@@ -218,31 +218,38 @@ instance (GenericApiSpec a, GenericApiSpec b) => GenericApiSpec (a :<|> b) where
 instance GenericApiSpec Request where
     gSpec _ _ = pure ()
 
+instance GenericApiSpec a => GenericApiSpec [a] where
+    gSpec xs toSpec = foldr (\x y -> gSpec x toSpec >> y) (pure ()) xs
```
KtorZ, line 222: I mistakenly expected a foldl here but no, foldr is right because of the composition of actions done as such :+1:

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
```diff
@@ -218,31 +218,38 @@ instance (GenericApiSpec a, GenericApiSpec b) => GenericApiSpec (a :<|> b) where
...
     ( Typeable a, Malformed (BodyParam a)
-    ) => GenericApiSpec (BodyParam a -> IO Request)
+    ) => GenericApiSpec (BodyParam a -> IO [Request])
   where
     gSpec toRequest toSpec = do
-        let tests = first toRequest <$> malformed @(BodyParam a)
-        toSpec . SomeTest (Proxy @a) =<< traverseLeft runIO tests
+        let tests :: [(IO [Request], ExpectedError)]
+            tests = first toRequest <$> malformed @(BodyParam a)
+        toSpec . SomeTest (Proxy @a) . insideOut =<< traverseLeft runIO tests
```
KtorZ: :+1:

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
```diff
@@ -251,25 +258,32 @@ instance
...
     , Typeable ct
     , Malformed (Header h ct)
-    ) => GenericApiSpec (Header h ct -> Request)
+    ) => GenericApiSpec (Header h ct -> [Request])
   where
-    gSpec toRequest toSpec = toSpec $
-        SomeTest (Proxy @h) (first toRequest <$> malformed @(Header h ct))
+    gSpec toRequest toSpec = do
+        let tests = fmap (\(xs, e) -> fmap (,e) xs)
+                (first toRequest <$> malformed @(Header h ct))
```
KtorZ: I find the `insideOut` version with the list comprehension much more readable

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
```diff
@@ -417,63 +431,64 @@ instance
...
+        t = wellformed :: [PathParam t]
 
-    type MkHeaderRequest (Capture p t :> sub) = MkHeaderRequest sub
+    type MkHeaderRequest (Capture p t :> sub) = [MkHeaderRequest sub]
     gEveryHeader _ req =
-        gEveryHeader (Proxy @sub) (addPathFragment t req)
+        fmap (\t' -> gEveryHeader (Proxy @sub) (addPathFragment t' req)) t
       where
-        t = wellformed :: PathParam t
+        t = wellformed :: [PathParam t]
```
KtorZ: Would be better to stick to a consistent naming like `ts`  before. These things are actually the same thing so they ought to have the same name.

KtorZ (approved):
Some minor remarks about consistency and readability that I fixed in a commit.
Nice work dealing with this, hope you didn't caught a headache :grimacing:

hasufell:
>  Nice work dealing with this, hope you didn't caught a headache grimacing

I did... and I completely forgot everything I did :rofl: so don't ask me anything about this code.

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
```diff
@@ -565,6 +579,9 @@ instance
 -- Helpers
 --
 
+distributeFirst :: [([x], y)] -> [(x, y)]
```
hasufell, line 582: Thanks... I wasn't able to come up with a sensible name.

The downside is that this won't inline in something like `fmap distributeFirst xs`. But I don't think there's anything performance critical about it.

Issue #2062, opened 2020-08-24
Fees are excessively high sometimes


Fell-x27:
# Context
 
Sometimes when you want to estimate fees, you can get weird result;

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      |2020.8.3 (git revision: b5b292b5a9f0ddccc0fc5d9e98bf3bf80f1a9d13) |
| Platform     | Linux        |
| Installation | From Hydra            |

# Steps to Reproduce

1. Try to estimate fees when payments + expected_fees_for_this_payments >= payments amount
2. ...
3. PROFIT!1

## Expected behavior
Fees for this transaction even if there is not enough money for it. I just want to know estimated fees for transaction like this. It works well for migrations, but there is a problem https://github.com/input-output-hk/cardano-wallet/issues/2061, so I can't use migrations instead.
## Actual behavior
 Your fees are about 1.8446744062438E+19

---

## Comment
Yep, I can handle it by myself with some nasty solutions, and I will, but would be nice if I have not to do this.

KtorZ:
Looks like a proper under/over flow in the fee estimation.

KtorZ:
I've been able to reproduce with an integration test scenario by trying to estimate fee for a wallet using the total balance (available UTxO + rewards). This indeed reaches an edge case of the fee estimation where fees are too big and some extra computation is done to calculate the estimated amount.

KtorZ:
### QA

- Added an extra integration test scenario illustrating the problem (fee calculation fallback with withdrawal) (https://github.com/input-output-hk/cardano-wallet/commit/f6e74dca19d77b05a3d3b1838badc538b7d00e93)

- Also added an invariant in-code to make sure that this does not happen by accident. The invariant checks that fee are within an acceptable bound and crashes the server otherwise.

piotr-iohk:
lgtm.

PR #2086, opened 2020-08-27
Fix underflow in fee estimation in the presence of withdrawal and max amount

KtorZ:
# Issue Number

#2062 

# Overview

10ab5507e out of paranoia, add an invariant to make transactions never leave fees beyond a certain amount.
3d4469ced pass down withdrawal value to 'handleCannotCoverFee' to prevent underflow in fee calculation.
f6e74dca1 add extra scenario covering fee estimation edge-case in the presence of withdrawal

TODO: I'd like to add one property or scenario to exercise the invariant and shows that it operates as expected.

# Comments

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
@@ -102,6 +102,7 @@ import Test.Integration.Framework.DSL
     , utcIso8601ToText
     , verify
     , walletId
+    , (.<=)
     , (.>)
     , (.>=)
     )
@@ -555,7 +556,7 @@ spec = do
                 (Link.getTransactionFee @'Shelley w) Default payload
             expectResponseCode @IO HTTP.status400 r
 
-    it "TRANS_ESTIMATE_03 - we see result when we can't cover fee" $ \ctx -> do
+    it "TRANS_ESTIMATE_03a - we see result when we can't cover fee" $ \ctx -> do
         wSrc <- fixtureWallet ctx
         payload <- mkTxPayload ctx wSrc faucetAmt fixturePassphrase
         r <- request @ApiFee ctx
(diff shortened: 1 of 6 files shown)


lib/core/src/Cardano/Wallet/Primitive/Fee.hs
```diff
@@ -92,6 +115,13 @@ data FeeOptions = FeeOptions
       :: OnDanglingChange
       -- ^ What do to when we encouter a dangling change output.
       -- See 'OnDanglingChange'
+
+    , feeUpperBound
+      :: Fee
+      -- ^ An extra upper-bound computed from the transaction max size. This is
+      -- used to construct invariant after balancing a transaction to make sure
+      -- that our algorithm isn't flawed and does not generate a transaction
+      -- that would inadvertendly leave too much fees.
```
jonathanknowles:
```suggestion
      -- ^ An extra upper bound computed from the transaction max size. This is
      -- used to construct an invariant, used after balancing a transaction, to
      -- make sure that the resultant fee is not unnecessarily high.
```

lib/core/src/Cardano/Wallet/Primitive/Fee.hs
```diff
@@ -140,15 +170,22 @@ newtype ErrAdjustForFee
...
 adjustForFee unsafeOpt utxo coinSel = do
-    let opt = invariant
-            "adjustForFee: fee must be non-null" unsafeOpt (not . nullFee)
-    senderPaysFee opt utxo coinSel
+    let opt = invariant "fee must be non-null" unsafeOpt (not . nullFee)
+    cs <- senderPaysFee opt utxo coinSel
+    let actualFee = Fee (feeBalance cs)
+    when (actualFee > feeUpperBound opt) $
+        error $ pretty $ unlinesF
+            [ "generated a coin selection with excessively large fees."
```
jonathanknowles: There is only one fee, I think?

```suggestion
            [ "generated a coin selection with an excessively large fee."
```

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -555,14 +556,35 @@ spec = do
...
         payload <- mkTxPayload ctx wSrc faucetAmt fixturePassphrase
         r <- request @ApiFee ctx
             (Link.getTransactionFee @'Shelley wSrc) Default payload
         verify r
             [ expectResponseCode HTTP.status202
             , expectField (#estimatedMin . #getQuantity) (.>= 0)
+            , expectField (#estimatedMax . #getQuantity) (.<= oneAda)
+            ]
+
+    it "TRANS_ESTIMATE_03b - we see result when we can't cover fee (with withdrawal)" $ \ctx -> do
```
jonathanknowles, line 570:
```suggestion
    it "TRANS_ESTIMATE_03b - we see result when we can't cover fee \
       \(with withdrawal)" $ \ctx -> do
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1191,14 +1191,21 @@ coinSelOpts tl txMaxSize = CoinSelectionOptions
...
     -> FeeOptions
-feeOpts tl action feePolicy minUtxo = FeeOptions
+feeOpts tl action txp minUtxo = FeeOptions
     { estimateFee = minimumFee tl feePolicy action
     , dustThreshold = minUtxo
     , onDanglingChange = if allowUnbalancedTx tl then SaveMoney else PayAndBalance
+    , feeUpperBound = Fee
+        $ ceiling a
+        + ceiling b * fromIntegral txMaxSize
+        + getCoin minUtxo
```
jonathanknowles, line 1205: **Suggestion**: perhaps extract this out as a separate function?

**Justification**: there is an almost identical calculation in `selectCoinsForMigrationFromUTxO`.

lib/core/src/Cardano/Wallet/Primitive/Fee.hs
```diff
@@ -2,6 +2,7 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NumericUnderscores #-}
```
jonathanknowles, line 5: :+1:

piotr-iohk (approved):
lgtm from black-box functional perspective. 

I have tested on MC4 with `self` and `mnemonic sentence` withdrawal on a wallet with rewards and fee amt is OK. The same fee amt is then applied when sending tx.

Issue #2074, opened 2020-08-26
Tx metadata JSON in API transaction list


rvl:
# Context

ADP-307

When listing transactions in the API, the user should see metadata attached to transactions - if it exists.

# Decision

1. Include metadata in `ApiTxMetadata` response.
2. JSON `null` denotes missing metadata (our generic defaults for aeson are to omit `Nothing` fields).

# Acceptance Criteria

1. Transaction list must show metadata in JSON format, if it exists.
2. Missing metadata should be `null`, to better illustrate that it's missing.
3. When posting a transaction with metadata, the JSON response must also show the metadata.
4. Swagger spec must be updated accordingly.

---

# Development 

As above.

# QA 

1. There are Api types unit tests for encoding/decoding.
2. Integration test `lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs` ⇒ PR #2096
   - TRANS_CREATE_10 - Transaction with metadata -- this tests the create transaction response, as well as checking the transaction list.

piotr-iohk:
lgtm

Issue #2073, opened 2020-08-26
Tx metadata accepted in API for creating a transaction


rvl:
# Context

ADP-307

When creating a transaction, users should be able to attach their metadata in JSON format.

# Decision

1. Use json conversion functions from `Cardano.Api.MetaData`.

# Acceptance Criteria

1. The API must accept metadata when posting to the shelley create transaction endpoint.
2. The API must accept metadata for the shelley transaction fee estimation endpoint.
3. The metadata field must be optional.
4. The swagger API definitions must be updated accordingly.

---

# Development 

- As above.
- Wrap TxMetadata in ApiTxMetadata so that JSON `null` can also denote absence of metadata.

# QA 

1. Api type serialisation/deserialisation is tested with unit tests.
2. There are unit tests in cardano-node for encoding/decoding metadata - see input-output-hk/cardano-node#1797. So we do not need to test multiple invalid metadata cases. One invalid metadata test is enough.
3. Integration tests in `lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs`
   - TRANS_CREATE_01 is modified to check that metadata is missing.
   - TRANS_CREATE_10  - Transaction with metadata
   - TRANS_CREATE_11 - Transaction with invalid metadata -- this is a metadata parsing error.
   - TRANS_CREATE_12 - Transaction with too much metadata

piotr-iohk:
Perhaps this is still in progress, but just to share few observations.

It seems one can send metadata that is _invalid_ in format, should we have more strict validation?

Examples:
1. one can use the same key multiple times which restuls in silently using only first one... maybe there should be validation and such metadata not allowed?
 Metadata: `{"1": 1, "1": 2, "1":123123}` silently results in `{ "1": 1 }`

2. can successfuly send map which is not map (basically list of lists or more generally list of different types)
 
list of lists
`{"4":[["k1", "v1", 1, 2, 4, 5, 6, 7], ["k2", "v2", "v3"]]}`

list of different types
`{111: [1,2,5,"4","4545", [1,2,4]]}`

3. can successfully send "nested" metadata 
`{ "0": {"cardano":"adasd", "2": {3: 4} } }`

4. too big top levels key is are silently made values that are OK (kind of cool, but maybe that's not OK?)
 
[...]

PR #2087, opened 2020-08-28
Transaction metadata in swagger API spec

rvl:
### Issue Number

ADP-307 / #2073 / #2074 

### Overview

Updates the swagger spec for the new metadata field when:
- listing transactions (metadata field is always present but may be null)
- posting a transaction (metadata field is optional)
- estimating fee (as above)

### Comments

[Rendered spec](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/input-output-hk/cardano-wallet/rvl/2073/swagger/specifications/api/swagger.yaml)

### DIFF
--- a/specifications/api/swagger.yaml
+++ b/specifications/api/swagger.yaml
@@ -566,13 +566,17 @@ x-transactionMetadata: &transactionMetadata
     Cardano allows users and developers to embed their own
     authenticated metadata when submitting transactions.
 
-    The metadata format is similar to JSON. String values are limited
-    in length, and metadata is included in the overall transaction
-    size limit.
+    The transaction metadata can be `null`, or a mapping from natural
+    numbers to metadata entries. These top-level keys can be any
+    number between `0` and `2^64 - 1`.
 
-    Note that metadata top-level keys must be integers, and only a subset of JSON
-    is accepted as values. See the example on the right to get an overview and the equivalence
-    table below to understand how metadata are mapped to on-chain data types.
+    Note that metadata is not free-form JSON. Only a subset of JSON is
+    accepted as values. String values are limited in length, and
+    metadata is included in the overall transaction size limit.
+
+    See the example on the right to get an overview and the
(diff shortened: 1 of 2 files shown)


KtorZ (approved):

rvl:
I found it difficult to debug swagger validation errors, so wrote this wiki doc: [Swagger Development](https://github.com/input-output-hk/cardano-wallet/wiki/Swagger-Development).

specifications/api/swagger.yaml
```diff
@@ -559,6 +559,48 @@ x-transactionStatus: &transactionStatus
...
+    numbers to metadata entries. These top-level keys can be any
+    number between `0` and `2^64 - 1`.
+
+    Note that metadata is not free-form JSON. Only a subset of JSON is
+    accepted as values. String values are limited in length, and
+    metadata is included in the overall transaction size limit.
+
+    See the example on the right to get an overview and the
+    equivalence table below to understand how metadata are mapped to
+    on-chain data types.
```
jonathanknowles, line 579: Just out of interest, what does "on the right" refer to here?

Suggestion: if this wording relies on a particular rendering style, perhaps it's better to use a layout-neutral method to refer to examples? (I'm guessing that not everyone will be viewing the API spec through the web tool that we provide.)

jonathanknowles (approved):
LGTM. :+1: 

The description of `transactionMetadata` is very readable.

I do have one question (see review comments) relating to accessibility of the spec.

specifications/api/swagger.yaml
KtorZ: It refers to our current documentation layout indeed.

KtorZ:
@rvl I'll look at the swagger failure above, this is a bit tricky because we use OpenAPI 3.0 but the Haskell library we use expects swagger 2.0 format, so there are some subtleties...

rvl:
Oh ok - I assumed it was maybe because the endpoints in the swagger file need to be added to the servant API.

PR #2088, opened 2020-08-28
Bump cardano-base so that shelley-spec-ledger-test compiles

rvl:
### Issue Number

Found during #2077.

### Overview

If you are keen enough to build the test suites of all our dependencies, you would have found `shelley-spec-ledger-test` failing to compile, because it needs a later revision of `cardano-base`. This will fix it.

### DIFF
--- a/cardano-1.19.dev1.yaml
+++ b/cardano-1.19.dev1.yaml
@@ -1,4 +1,4 @@
-name: cardano-1.19.x
+name: cardano-1.19.dev1
 
 resolver: lts-14.25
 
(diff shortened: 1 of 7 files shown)


hasufell:
Maybe not related directly to this, but it would also be good to bump cardano-sl-x509: https://github.com/input-output-hk/cardano-sl-x509/commit/12925934c533b3a6e009b61ede555f8f26bac037

hasufell (approved):

KtorZ:
@rvl I am afraid Buildkite is running into the same issues as I did because of the cache it's using. We should try purging the cache or; renaming the snapshot to something else otherwise it is going to annoy everyone using stack once this PR is merged.

rvl:
@KtorZ This time I renamed the snapshot file - so I think stack will be able to invalidate its cache. Buildkite master branch still works (the stack build phase at least).

PR #2089, opened 2020-08-28
Add TxMetadata to API

rvl:
### Issue Number

ADP-307 / #2073 / #2074 

### Overview

Add TxMetadata to API.

### Comments

- Still in draft because API swagger tests are failing. It's not liking the schema references.


KtorZ (approved):
Reviewed and fixed the swagger spec verification. The test is rather weak since the swagger2 library we use relies on OpenAPI 2.0 schemas which do not support the `oneOf` combinator... So right now, the only check we can really do on metadata is that they are any valid JSON... which is on the other hand, pretty much all we care about.

PR #2090, opened 2020-08-28
Add metadata to transaction layer

rvl:
### Issue Number

ADP-307 / #2073

### Overview

- Adds metadata as a parameter of `mkStdTx` so that transactions can be submitted with metadata.

### Comments

Based on the branch of PR #2079.

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -302,6 +302,7 @@ import Cardano.Wallet.Primitive.Types
     , TransactionInfo (..)
     , Tx
     , TxMeta (..)
+    , TxMetadata
     , TxOut (..)
     , TxStatus (..)
     , UTxO (..)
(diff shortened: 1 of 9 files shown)


lib/core/src/Cardano/Wallet.hs
```diff
@@ -1591,6 +1592,7 @@ signTx ctx wid pwd (UnsignedTx inpsNE outsNE) = db & \DBLayer{..} -> do
     nl = ctx ^. networkLayer @t
     inps = NE.toList inpsNE
     outs = NE.toList outsNE
+    md = Nothing -- no metadata
```
KtorZ: Shouldn't this be better left as an argument as well ?

rvl: Probably - fixed.

rvl:
bors try

KtorZ (approved):

Issue #2072, opened 2020-08-26
Tx metadata stored in sqlite database


rvl:
# Context

Part of ADP-307

# Decision

Transaction metadata is stored in the SQLite database indexed by TxId.

The metadata is converted to JSON and encoded, and stored in a STRING column.

Schema:
```
TxMetadata
    txMetadataTxId    TxId          sql=tx_id
    txMetadataValue   Text          sql=value

    Primary txMetadataTxId
    deriving Show Generic
```

Pending transactions' metadata are stored in the same manner as confirmed transactions.

Absence of metadata in a transaction corresponds to absence of a `tx_metadata` table row.

I was considering making one table row per top-level metadata map entry, but that's not really needed, slightly more complicated to implement, and makes it harder to differentiate between no metadata and an empty metadata map (if there is in fact any difference on chain).

# Acceptance Criteria

1. Transaction metadata must be stored in the database with the transaction.
2. When transactions are deleted, metadata must also be deleted.

---

# Development 

# QA

- State machine DB tests cover transaction history operations. `lib/core/test/unit/Cardano/Wallet/DB/StateMachine.hs`
- The Tx generator should include arbitrary generated metadata. This arbitrary metadata only includes valid data because metadata is already validated at the API boundary.

piotr-iohk:
lgtm.

PR #2091, opened 2020-08-28
Add tx metadata to sqlite database

rvl:
### Issue Number

ADP-307 / #2072 

### Overview

Includes transaction metadata in the sqlite database transaction history.

### Comments

- Based on PR #2079 branch
- DB state machine tests fail. Possibly due to mutations during the TxMetadata json round-trip.

### DIFF
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
@@ -1280,7 +1280,6 @@ deleteLooseTransactions = do
     deleteLoose "tx_in"
     deleteLoose "tx_out"
     deleteLoose "tx_withdrawal"
-    deleteLoose "tx_metadata"
   where
     -- Deletes all TxIn/TxOuts/TxWithdrawal returned by the sub-select.
     -- The sub-select outer joins TxMeta with TxIn/TxOut/TxWithdrawal.
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
@@ -80,7 +80,6 @@ import Cardano.Wallet.DB.Sqlite.TH
     , StakeKeyCertificate (..)
     , TxIn (..)
     , TxMeta (..)
-    , TxMetadata (..)
     , TxOut (..)
     , TxWithdrawal (..)
     , UTxO (..)
@@ -109,7 +108,7 @@ import Control.Concurrent.MVar
 import Control.Exception
     ( Exception, bracket, throwIO )
 import Control.Monad
-    ( forM, forM_, unless, void, when )
+    ( forM, unless, void, when )
 import Control.Monad.Extra
     ( concatMapM )
 import Control.Monad.IO.Class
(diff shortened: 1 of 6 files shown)


lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -128,13 +129,13 @@ import Data.Either
 import Data.Generics.Internal.VL.Lens
     ( (^.) )
 import Data.List
-    ( nub, sortOn, unzip3 )
+    ( nub, sortOn, unzip4 )
```
KtorZ: I like how it started with just unzip2 and gradually increased. We still have 3 left, after which we have to stop further development :)

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1252,30 +1256,34 @@ updateTxMetas
...
+        [ (TxMetaKey txMetaTxId txMetaWalletId, m)
+        | m@TxMeta{..} <- metas]
     dbChunked repsertMany
         [ (TxInKey txInputTxId txInputSourceTxId txInputSourceIndex, i)
         | i@TxIn{..} <- txins ]
     dbChunked repsertMany
         [ (TxOutKey txOutputTxId txOutputIndex, o)
         | o@TxOut{..} <- txouts ]
+    forM_ (chunksOf chunkSize $ map txMetaTxId metas) $ \txids ->
+        deleteWhere [ TxMetadataTxId <-. txids ]
```
KtorZ: Why the delete? Isn't the point of `repsert` to "replace or insert"?

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1039,13 +1039,14 @@ mkTxHistory wid txs = flatTxHistory
...
-        ( map (\(a,_,_) -> a) entities
-        , concatMap (fst . (\(_,b,_) -> b)) entities
-        , concatMap (snd . (\(_,b,_) -> b)) entities
-        , concatMap (\(_,_,c) -> c) entities
+        ( map (\(a,_,_,_) -> a) entities
+        , concatMap (fst . (\(_,b,_,_) -> b)) entities
+        , concatMap (snd . (\(_,b,_,_) -> b)) entities
+        , concatMap (\(_,_,c,_) -> maybeToList c) entities
+        , concatMap (\(_,_,_,d) -> d) entities
         )
```
KtorZ: Wouldn't be more efficient and simpler to store metadata with the rest of the _static_ data (status, direction etc..)? 

We have to resort to separate database tables for TxIn, TxOut and TxWithdrawals because they are unbounded (modulo tx size) **lists**. Metadata however can be treated as a bunch of bytes that are either present or not.

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
```diff
@@ -100,6 +100,16 @@ TxMeta
...
 
+-- Maps a transaction ID to its on-chain metadata.
+-- The metadata map is encoded in JSON with the same schema
+-- as what is used in the API.
+TxMetadata
+    txMetadataTxId    TxId          sql=tx_id
+    txMetadataValue   W.TxMetadata  sql=value
+
+    Primary txMetadataTxId
+    deriving Show Generic
```
KtorZ: I am a bit doubtful about having this as a separate table (cf earlier comment).

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
rvl: My reason was just that sqlite tables don't cost much and it seemed like it would be confusing to have data in the `TxMeta` table which was actually a field of `Tx`.

KtorZ (approved):
Temporarily worked-around the roundtripping issues until #2098 is fixed. state-machine tests pass for now.

rvl:
Thanks

bors r+

PR #2092, opened 2020-08-30
Fix stack build on windows

rvl:
### Issue Number

None

### Overview

On windows, the stack build would fail with some missing dependencies. There is some stack issue with GHC boot packages. The recommended course of action is to add them as `extra-deps` in `stack.yaml`. This fixes the build.

### Comments

Since the libsodium VRF was added, there is now an extra step to get stack builds working on windows. I have updated our [iohk-setup.ps1](https://github.com/input-output-hk/adrestia/wiki/scripts/iohk-setup.ps1) script to install libsodium and configure stack to use it.

### DIFF
--- a/nix/.stack.nix/default.nix
+++ b/nix/.stack.nix/default.nix
@@ -12,6 +12,9 @@
         "servant-swagger" = (((hackage.servant-swagger)."1.1.8").revisions).default;
         "zip" = (((hackage.zip)."1.3.0").revisions).default;
         "tls" = (((hackage.tls)."1.5.4").revisions).default;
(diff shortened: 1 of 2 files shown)


Anviking (approved):

Anviking:
(It worked! Thanks for this!)

KtorZ:
Reinstalled virtualbox on my new machine and followed the steps in the script above. Everything went just fine, well done!
I will now try running stack and some tests :+1:

Q: The script is a bit in a hidden location at the moment, shall we consider having it cardano-wallet/scripts ?

Issue #2025, opened 2020-08-11
'minLength' must be lower value than 'maxLength'


melmccann:
https://github.com/input-output-hk/cardano-wallet/blob/e096e3403bdd94a67805ede25e2c3818fc875d7a/specifications/api/swagger.yaml#L1404

KtorZ:
That's an interesting one. These two seem to have been swapped.

piotr-iohk:
Fixed in #2093.

Issue #2023, opened 2020-08-11
Wrong user-facing encoding for poold id and reward account


KtorZ:
# Context
 
 

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      |  |
| Platform     |           |
| Installation |             |

- [x] Pool id are currently encoded as base16 strings. 
- [x] Reward accounts are currently encoded as bech32 with a "stake_addr" prefix

# Steps to Reproduce

N/A

## Expected behavior

- Both types are encoded as bech32 using prefixes as specified in [CIP-5](https://github.com/cardano-foundation/CIPs/tree/master/CIP5)

## Actual behavior
 

---

# Resolution 

---

# QA

hasufell:
Where in the API can I reproduce this?

----

Edit:

- pool id: https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/listStakePools
- reward account: e.g. https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/listTransactions

Any other endpoints affected?

The offending code spots are these, afais:

https://github.com/input-output-hk/cardano-wallet/blob/d7994e9c83ed00d798cf24f3101e50f0383ddc17/lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs#L921-L931
 
https://github.com/input-output-hk/cardano-wallet/blob/d7994e9c83ed00d798cf24f3101e50f0383ddc17/lib/core/src/Cardano/Wallet/Primitive/Types.hs#L677-L695

hasufell:
Afais this will change various API endpoints, such as: https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/joinStakePool where the pool ID is a parameter

Having the encoding for parameter and json output diverge is probably going to be confusing?

piotr-iohk:
lgtm.

I have adjusted swagger to show that pool id can be hex or bech32 in https://github.com/input-output-hk/cardano-wallet/pull/2093 which should also address #2025.

hasufell:
@rhyslbw what's missing?

rhyslbw:
> Reward accounts are currently encoded as bech32 with a "stake_addr" prefix

rhyslbw:
Done in https://github.com/input-output-hk/cardano-wallet/pull/2055/commits/c36ba7d897b90e3d6bcaaae90b18f0c4cf7b5d68

PR #2093, opened 2020-08-31
Swagger adjustment: Supported pool id param is hex or bech32

piotr-iohk:
# Issue Number

https://github.com/input-output-hk/cardano-wallet/issues/2023
https://github.com/input-output-hk/cardano-wallet/issues/2025

# Overview

- 01d5d509e65fa9ba18f2be91554953ca2ccdb15d
  Supported pool id param is hex or bech32

# Comments

**Before:**

Screenshot from 2020-08-31 13-42-59

**After:**

Screenshot from 2020-08-31 16-16-37

### DIFF
--- a/specifications/api/swagger.yaml
+++ b/specifications/api/swagger.yaml
@@ -1404,9 +1404,7 @@ x-parametersStakePoolId: &parametersStakePoolId
   required: true
   schema:
     type: string
-    format: hex
-    maxLength: 56
-    minLength: 64
+    format: hex|bech32
 
 x-parametersAddressId:  &parametersAddressId
   in: path
@@ -2226,6 +2224,9 @@ paths:
         <p align="right">status: <strong>stable</strong></p>
 
         Delegate all (current and future) addresses from the given wallet to the given stake pool.
+
+        <strong>Note:</strong> Bech32-encoded stake pool identifiers can vary in length.
+
       parameters:
         - *parametersStakePoolId
         - *parametersWalletId


specifications/api/swagger.yaml
```diff
@@ -1404,9 +1404,9 @@ x-parametersStakePoolId: &parametersStakePoolId
   required: true
   schema:
     type: string
-    format: hex
+    format: hex|bech32
     maxLength: 56
-    minLength: 64
+    minLength: 56
```
piotr-iohk: Empirically I see that pool ids (no matter if in hex or bech32 format) have consistently 56 char length :man_shrugging:

hasufell: @KtorZ is that something we can guarantee? I'm not sure how deterministic bech32 is here.

KtorZ: I don't think this is right indeed, so we could leave it as a comment in the description instead.

Also, pool ids can be 64 chars ..... with jormungandr :s

piotr-iohk: I have added a comment. Pls let me know it that's fine.

specifications/api/swagger.yaml
```diff
@@ -2226,6 +2224,9 @@ paths:
         <p align="right">status: <strong>stable</strong></p>
 
         Delegate all (current and future) addresses from the given wallet to the given stake pool.
+
+        <strong>Note:</strong> Stake pool id lenght can be undeterministic when encoded as bech32.
```
hasufell: `length`

hasufell (approved):

specifications/api/swagger.yaml
piotr-iohk: :facepalm:

specifications/api/swagger.yaml
```diff
@@ -2226,6 +2224,9 @@ paths:
         <p align="right">status: <strong>stable</strong></p>
 
         Delegate all (current and future) addresses from the given wallet to the given stake pool.
+
+        <strong>Note:</strong> Stake pool id length can be undeterministic when encoded as bech32.
```
KtorZ: can be --> is ?

KtorZ (approved):

specifications/api/swagger.yaml
jonathanknowles: I'm not sure that "undeterministic" (which should probably be "nondeterministic") is the best way to word this.

Readers **_might_** infer that the bech32 encoding is nondeterministic in its length. Of course this is not true, as Bech32 encoding is completely deterministic.

Perhaps we could instead say something like:

```
Bech32-encoded stake pool identifiers can vary in length.
```

And perhaps then give some examples to illustrate the different lengths that are possible.

@piotr-iohk What do you think?

jonathanknowles (commented):
@piotr-iohk 

I'm not sure that "undeterministic" (which should probably read "nondeterministic") is the best way to word this.

Unfamiliar readers _might_ infer that the bech32 encoding is nondeterministic in its length. (Of course as we know this isn't true, as Bech32 encoding is completely deterministic.)

Perhaps we could instead say something like:

```
Bech32-encoded stake pool identifiers can vary in length.
```

And perhaps then give some examples to illustrate the different lengths that are possible.

What do you think?

PR #2094, opened 2020-08-31
Return descriptive 503 when listing stake pools with uncomplete sync

hasufell:
Wrt #1971

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -172,6 +172,9 @@ module Cardano.Wallet
 
     -- * Logging
     , WalletLog (..)
+
+    -- * Stake pool listing
+    , ErrListPools (..)
     ) where
 
 import Prelude hiding
@@ -2354,6 +2357,11 @@ data ErrWithdrawalNotWorth
     = ErrWithdrawalNotWorth
     deriving (Generic, Eq, Show)
 
+-- | Errors that can occur when trying to list stake pool.
+data ErrListPools
+    = ErrListPoolsNetworkError ErrNetworkUnavailable
+    | ErrListPoolsPastHorizonException PastHorizonException
+    deriving (Show)
 {-------------------------------------------------------------------------------
                                    Utils
 -------------------------------------------------------------------------------}
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -109,6 +109,7 @@ import Cardano.Wallet
     , ErrImportAddress (..)
     , ErrImportRandomAddress (..)
     , ErrJoinStakePool (..)
+    , ErrListPools (..)
     , ErrListTransactions (..)
     , ErrListUTxOStatistics (..)
     , ErrMkTx (..)
@@ -314,7 +315,7 @@ import Control.Concurrent
 import Control.Concurrent.Async
     ( race_ )
 import Control.Exception
-    ( IOException, bracket, throwIO, tryJust )
+    ( IOException, bracket, throwIO, try, tryJust )
 import Control.Monad
     ( forM, forever, void, when, (>=>) )
 import Control.Monad.Catch
@@ -422,6 +423,7 @@ import qualified Cardano.Wallet.Api.Types as Api
 import qualified Cardano.Wallet.Network as NW
 import qualified Cardano.Wallet.Primitive.AddressDerivation.Byron as Byron
 import qualified Cardano.Wallet.Primitive.AddressDerivation.Icarus as Icarus
+import qualified Cardano.Wallet.Primitive.Slotting as S
 import qualified Cardano.Wallet.Primitive.Types as W
 import qualified Cardano.Wallet.Registry as Registry
 import qualified Data.Aeson as Aeson
@@ -1587,17 +1589,20 @@ assignMigrationAddresses addrs selections =
                                     Network
 -------------------------------------------------------------------------------}
 
-data ErrCurrentEpoch = ErrUnableToDetermineCurrentEpoch
+data ErrCurrentEpoch
+    = ErrUnableToDetermineCurrentEpoch
+    | ErrCurrentEpochPastHorizonException PastHorizonException
 
 getCurrentEpoch
     :: forall ctx s t k . (ctx ~ ApiLayer s t k)
     => ctx
     -> Handler W.EpochNo
 getCurrentEpoch ctx = do
-    res <- liftIO $ currentEpoch ti
+    res <- liftIO $ try $ currentEpoch ti
     case res of
-        Nothing -> liftE ErrUnableToDetermineCurrentEpoch
-        Just x  -> pure x
+        Right Nothing  -> liftE ErrUnableToDetermineCurrentEpoch
+        Right (Just x) -> pure x
+        Left e@(S.PastHorizon{}) -> liftE (ErrCurrentEpochPastHorizonException e)
   where
     ti :: TimeInterpreter IO
     ti = timeInterpreter (ctx ^. networkLayer @t)
@@ -2019,6 +2024,7 @@ instance LiftHandler ErrCurrentEpoch where
                 [ "I'm unable to determine the current epoch. "
                 , "Please wait a while for the node to sync and try again."
                 ]
+        ErrCurrentEpochPastHorizonException e -> handler e
 
 instance LiftHandler ErrUnexpectedPoolIdPlaceholder where
     handler = \case
@@ -2503,6 +2509,11 @@ instance LiftHandler ErrWithdrawalNotWorth where
                 , "enough to deserve being withdrawn. I won't proceed with that "
                 , "request."
                 ]
+
+instance LiftHandler ErrListPools where
+    handler = \case
+        ErrListPoolsNetworkError e -> handler e
+        ErrListPoolsPastHorizonException e -> handler e
 
 instance LiftHandler (Request, ServerError) where
     handler (req, err@(ServerError code _ body headers))
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
@@ -37,6 +37,8 @@ import Cardano.Pool.DB
     ( DBLayer (..), ErrPointAlreadyExists (..), readPoolLifeCycleStatus )
 import Cardano.Pool.Metadata
     ( StakePoolMetadataFetchLog )
+import Cardano.Wallet
+    ( ErrListPools (..) )
 import Cardano.Wallet.Api.Types
     ( ApiT (..) )
 import Cardano.Wallet.Network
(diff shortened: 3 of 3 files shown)


lib/core/src/Cardano/Wallet.hs
```diff
@@ -2316,6 +2319,13 @@ data ErrWithdrawalNotWorth
     = ErrWithdrawalNotWorth
     deriving (Generic, Eq, Show)
 
+-- | Errors that can occur when trying to list stake pool.
```
jonathanknowles, line 2360: There's more than just one stake pool, I hope! :smile_cat: 

```suggestion
-- | Errors that can occur when trying to list stake pools.
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2316,6 +2319,13 @@ data ErrWithdrawalNotWorth
...
+data ErrListPools
+    = ErrListPoolsNetworkError ErrNetworkUnavailable
+    | ErrListPoolsPastHorizonException PastHorizonException
+    deriving (Show)
+
+
```
jonathanknowles: Some extra space has invaded here:
```suggestion

```
https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#use-only-a-single-blank-line-between-top-level-definitions

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -184,17 +190,18 @@ newStakePoolLayer nl db@DBLayer {..} =
         :: EpochNo
         -- Exclude all pools that retired in or before this epoch.
         -> Coin
-        -> ExceptT ErrNetworkUnavailable IO [Api.ApiStakePool]
+        -> ExceptT ErrListPools IO [Api.ApiStakePool]
     _listPools currentEpoch userStake = do
         tip <- withExceptT fromErrCurrentNodeTip $ currentNodeTip nl
-        rawLsqData <- stakeDistribution nl tip userStake
+        rawLsqData <- mapExceptT (fmap (first ErrListPoolsNetworkError))
+                        $ stakeDistribution nl tip userStake
```
jonathanknowles: **Request:** use 4-space indents.

```suggestion
        rawLsqData <- mapExceptT (fmap (first ErrListPoolsNetworkError))
            $ stakeDistribution nl tip userStake
```

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -204,11 +211,16 @@ newStakePoolLayer nl db@DBLayer {..} =
...
+
       where
-        fromErrCurrentNodeTip :: ErrCurrentNodeTip -> ErrNetworkUnavailable
+
+        fromErrCurrentNodeTip :: ErrCurrentNodeTip -> ErrListPools
         fromErrCurrentNodeTip = \case
-            ErrCurrentNodeTipNetworkUnreachable e -> e
-            ErrCurrentNodeTipNotFound -> ErrNetworkUnreachable "tip not found"
+            ErrCurrentNodeTipNetworkUnreachable e -> ErrListPoolsNetworkError e
+            ErrCurrentNodeTipNotFound -> ErrListPoolsNetworkError $ ErrNetworkUnreachable "tip not found"
```
jonathanknowles: **Request:** wrap code to meet the [line length limit](https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#limit-line-length-to-80-characters).

**Justification:** Long lines make it harder to read multiple code sources side-by-side.

```suggestion
            ErrCurrentNodeTipNetworkUnreachable e ->
                ErrListPoolsNetworkError e
            ErrCurrentNodeTipNotFound ->
                ErrListPoolsNetworkError $ ErrNetworkUnreachable "tip not found"
```

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -204,11 +211,16 @@ newStakePoolLayer nl db@DBLayer {..} =
                 (nOpt rawLsqData)
                 lsqData
                 dbData
+        case r of
+            Left e@(PastHorizon{}) -> throwE (ErrListPoolsPastHorizonException e)
+            Right r' -> pure r'
```
jonathanknowles, line 224: **Suggestion**: use the `first` combinator to avoid pattern matching on `Right`.

```suggestion
        handlePastHorizon r
      where
        handlePastHorizon = except . first
            (\e@PastHorizon {} -> ErrListPoolsPastHorizonException e)
```

jonathanknowles (changes requested):
Looks good so far. :+1: 

I've added some suggestions in the comments.

There are still some issues with formatting to address. (See comments.)

lib/core/src/Cardano/Wallet.hs
hasufell: As I'm not counting whitespaces, I'm also not counting newlines between code blocks. So we're back to: use a formatter :)

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
hasufell: why?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
hasufell: This already is 4-spaced indent.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
jonathanknowles: > why?

BTW, this was just a suggestion, rather than a request. If you feel the original is clearer, then feel free to leave it that way.

My reasons for suggesting it:

* Using `first` allows you to omit the `Right` pattern match, which means that you can avoid the extra `r'` binding.
* The `first` and `second` functions (from `Data.Bifunctor`) are used fairly commonly in our code base, so it's a familiar pattern to the team.

hasufell: In this case it doesn't really matter, but I'd like to point out that these expressions aren't strictly the same:

1. pattern matching forces evaluation immediately, so there are slightly different strictness properties
2. depending on how you define the utility function `handlePastHorizon` it may or may not inline and (at worst) cause an additional closure allocation (in this case I believe it does inline, unless some mtl-ism blocks it from doing so)

This is why I'm usually a bit more conservative with rewriting expressions, because even if they look the same, they aren't always.

piotr-iohk:
@hasufell I have taken your branch (`hasufell:hasufell/1971/fix-err-list-stakepools`) for a little spin and I actually still see the same behavior as -> https://github.com/input-output-hk/cardano-wallet/issues/1971#issuecomment-683685247 :thinking: 

Still see 500 and "Somenthing went wrong" on listing stake pools while wallet is syncing through Byron era. (Checked on [testnet](https://hydra.iohk.io/build/3854628/download/1/index.html))

(btw, is there a reason why you are using your fork of cardano-wallet instead of using main repo?)

hasufell:
I can reproduce. This is the problem as described in https://github.com/input-output-hk/cardano-wallet/issues/1971#issuecomment-683728271

It's hard to say where this exception leaks to.

hasufell:
@piotr-iohk should be fixed, please retry

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1580,16 +1582,18 @@ assignMigrationAddresses addrs selections =
 -------------------------------------------------------------------------------}
 
 data ErrCurrentEpoch = ErrUnableToDetermineCurrentEpoch
+                     | ErrCurrentEpochPastHorizonException PastHorizonException
```
jonathanknowles: @hasufell This style is different from our team coding standard ([source](https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#avoid-variable-length-indentation)). Please could you fix it?

We use this style in the code base (newlines with 4-space indents):
```suggestion
data ErrCurrentEpoch
    = ErrUnableToDetermineCurrentEpoch
    | ErrCurrentEpochPastHorizonException PastHorizonException
```

Thank-you.

jonathanknowles (dismissed):
Hi @hasufell

There are still some issues with coding style.

In a previous review, I made these requested changes, and they were marked as resolved, but they don't seem to be have been fixed (or if they were, perhaps the fixes were reverted by a subsequent push?):
* https://github.com/input-output-hk/cardano-wallet/pull/2094/files#r480718996
* https://github.com/input-output-hk/cardano-wallet/pull/2094/files#r480723300
* https://github.com/input-output-hk/cardano-wallet/pull/2094/files#r480716728

Please could you address these issues?

As I [mentioned before](https://github.com/input-output-hk/cardano-wallet/pull/2055#discussion_r476343085), I would love to be able to give you a tool that could automatically apply the correct format, but I don't have one. If you'd like to explore automated solutions such as [brittany](https://hackage.haskell.org/package/brittany), I would not be opposed. But that requires a team discussion which is outside the scope of this PR.

Given that our team does not yet use an automated tool, team members generally make a best effort to make their code follow the standard. Could you do that? I don't think it's too hard to follow given that it's rather short, and only has a few rules. We provide an `.editor` file to make this easier, so the burden should not be too arduous.

As a reminder:
* The purpose of having a written standard is to record things that the team has agreed on, so discussions about style can be avoided during PR reviews.
* The benefit is that we can concentrate on more important issues during PR reviews, such as functionality and correctness.
* Rules are added to the standard as the result of team discussions and voting. The standard is a consensus that we've worked out over time.

We allow ourselves the flexibility to occasionally break our own standard if there is good reason. But I don't think that this flexibility should be interpreted as licence to not follow the standard (on the basis that there is no automated tool), especially when the reviewer has gone to the effort of providing suggestions and feedback.

Thank-you.

hasufell:
Hi @jonathanknowles ,

thank you very much for your detailed explanation!

I believe there must be a misunderstanding though. I wasn't arguing against the coding standard in any of my PRs I believe, nor have I indicated that I don't intend to follow it. In fact I've spent a great deal of time configuring my editor, filing bug reports on various vim plugins and reading through stylish-haskell documentation in order to figure out if some of this can be automated (because we already use this tool). I'm using the `.editorconfig` since the first week, but it doesn't do much.

Maybe you're faster at pattern recognition and adapting to different coding styles. I personally am not and I've never worked with this coding style, so I'm afraid it will take more time before my PRs will satisfy your standards, unless this will be automated. Please bear with me.

Thanks for your time reviewing these points!

piotr-iohk:
> @piotr-iohk should be fixed, please retry

LGTM.

```
GET http://localhost:8090/v2/stake-pools?stake=1000000000

The response was:
Code = 503,
Message = {"code":"past_horizon","message":"Tried to convert something that is past the horizon (due to uncertainty about the next hard fork). Wait for the node to finish syncing to the hard fork. Depending on the blockchain, this process can take an unknown amount of time."}
```

KtorZ (approved):

piotr-iohk (approved):

Issue #2083, opened 2020-08-27
Windows integration tests fail with "The action has timed out" / node dies


Anviking:
# Context
 
https://github.com/input-output-hk/cardano-wallet/actions/runs/203259879
https://github.com/input-output-hk/cardano-wallet/actions/runs/201779298
https://github.com/input-output-hk/cardano-wallet/actions/runs/198447606
https://github.com/input-output-hk/cardano-wallet/actions/runs/209284173
https://github.com/input-output-hk/cardano-wallet/actions/runs/195488381
https://github.com/input-output-hk/cardano-wallet/actions/runs/194028487
https://github.com/input-output-hk/cardano-wallet/actions/runs/192595541
https://github.com/input-output-hk/cardano-wallet/actions/runs/191653223
https://github.com/input-output-hk/cardano-wallet/actions/runs/190681089
https://github.com/input-output-hk/cardano-wallet/actions/runs/189300667
https://github.com/input-output-hk/cardano-wallet/actions/runs/224510919
https://github.com/input-output-hk/cardano-wallet/actions/runs/187857979
https://github.com/input-output-hk/cardano-wallet/actions/runs/207833914
https://github.com/input-output-hk/cardano-wallet/actions/runs/215981659

Occurring 14 times in the 29 most recent CI windows runs

# Test Case

Multiple, I think

# Failure / Counter-example

```
2020-08-15T01:26:36.3384177Z [33m[cardano-wallet.network:Warning:1554][0m [2020-08-15 01:26:36.30 UTC] Connection lost with the node. withIODataPtr (readHandle): resource vanished (The pipe has been ended.)
```

```
2020-08-15T01:26:38.1764565Z [31m[cardano-wallet.integration:Error:823][0m [2020-08-15 01:26:38.17 UTC] bft Child process cardano-node exited with status 252
2020-08-15T01:26:38.8146579Z [33m[cardano-wallet.network:Warning:1552][0m [2020-08-15 01:26:38.81 UTC] Couldn't connect to node (x6). Retrying in a bit...
2020-08-15T01:26:38.8147005Z [33m[cardano-wallet.network:Warning:1554][0m [2020-08-15 01:26:38.81 UTC] Couldn't connect to node (x6). Retrying in a bit...
2020-08-15T01:26:39.4034917Z [33m[cardano-wallet.network:Warning:67771][0m [2020-08-15 01:26:39.40 UTC] Couldn't connect to node (x6). Retrying in a bit...
[...]

PR #2095, opened 2020-08-31
Replace BS.init with T.strip to handle windows newlines (`\r\n`)

Anviking:
# Issue Number

#2083 / related

# Overview

- [x] Replace BS.init with T.strip to handle windows newlines (`\r\n`). Integration tests failed with a pattern match failure here, when I ran locally on windows.

# Comments

- I've never seen CI fail for this reason, which is arguably strange.


KtorZ (approved):
Subtle...  There are maybe other places in the Launch code where we simply assume unix newlines...

rvl:
The reason only you can see the error is because your git on windows is configured to automatically convert newlines when checking out files.

Issue #2082, opened 2020-08-27
Listing stake pools time grows when there are wallets syncing in background


piotr-iohk:
# Context
 
Not sure how is the relation but the time of listing pools seems to grow when there are more wallets being synced.
I don't see any obvious outstanding query time that would explain 50s... the longest query is `GetNonMyopicMemberRewards` which is a bit longer the more wallets are syncing. (Which applies to other querries too actually).

**0 Wallets syncing:**
List sp = 3s
Query GetNonMyopicMemberRewards took 2.028777786s

**1 wallet syncing:**
List sp = 13s
Query GetNonMyopicMemberRewards took 2.39806037s

**2 wallets syncing**
List sp = 26s
Query GetNonMyopicMemberRewards took 2.452248284s

**3 wallets syncing**
List sp = 31s
Query GetNonMyopicMemberRewards took 2.402175446s

**5 wallets syncing**
List sp = 40s
Query GetNonMyopicMemberRewards took 2.821580222s

**10 wallets syncing**
List sp = 47s
Query GetNonMyopicMemberRewards took 3.001514181s

**15 wallets syncing**
List sp = 1m 03s
Query GetNonMyopicMemberRewards took 3.012593506s

**20 wallets syncing**
List sp = 1m 11s
Query GetNonMyopicMemberRewards took 3.519106839s

Times of other querries is also growing a bit too and also the number of queries is growing when there are more wallets:

With 0 wallets (lp exec time = 4s):
```
Query GetNonMyopicMemberRewards took 2.386650034s
Query GetCurrentPParams took 0.000716248s
Query GetEraStart took 0.00137057s
Query getAccountBalance took 0.001327425s
Query GetCurrentPParams took 0.000513555s
Query GetUpdateInterfaceState took 0.000478
```
With 20 wallets (lp exec time = 1:10m):
```
Query GetEraStart took 0.637137392s
Query getAccountBalance took 0.648792572s
Query GetCurrentPParams took 0.626415965s
Query GetUpdateInterfaceState took 0.753813193s
Query GetStakeDistribution took 0.570119185s
Query GetEraStart took 0.771473205s
Query GetCurrentPParams took 0.650308479s
Query GetUpdateInterfaceState took 0.784389s
Query GetNonMyopicMemberRewards took 3.200318526s
Query getAccountBalance took 3.245384946s
Query GetCurrentPParams took 1.100185536s
Query GetEraStart took 0.842396347s
[...]

PR #2111, opened 2020-09-04
Faster listing of stake pools.

jonathanknowles:
## Related Issues

#2082 (_Listing stake pools time grows when there are wallets syncing in background_.)

https://jira.iohk.io/browse/ADP-383 (_Use more efficient DB queries when listing pools_.)

## Overview

This PR:

* changes the implementation of `ListStakePools` to execute a small number of handwritten SQL queries with native joins, rather than issuing multiple queries and joining together the results in memory.

* adds a new operation [`listPoolLifeCycleData`](https://github.com/input-output-hk/cardano-wallet/blob/d00c608e47d2d189a282a02946cc8e46b843375b/lib/core/src/Cardano/Pool/DB.hs#L180). The [SQLite implementation](https://github.com/input-output-hk/cardano-wallet/blob/d00c608e47d2d189a282a02946cc8e46b843375b/lib/core/src/Cardano/Pool/DB/Sqlite.hs#L384) of this operation uses a [single hand-optimized query](https://github.com/input-output-hk/cardano-wallet/blob/d00c608e47d2d189a282a02946cc8e46b843375b/lib/core/src/Cardano/Pool/DB/Sqlite.hs#L595) to return lifecycle data for all pools.

### :star: **Faster execution time**

  The time taken to execute `ListStakePools` is greatly reduced, even when there are multiple wallets syncing in the background.
  <details><summary>Click to view data</summary><br>

  | No. of <br>Wallets<br>Syncing | Time Required<br>(seconds)<br>`master` | Time Required<br>(seconds)<br>`more-efficient-list-pools` |
  | ---: | ---: | ---: |
  | 0 | 8 | 3 |
  | 5 | 30 | 7 |
  | 10 | 60 | 8 |
  | 15 | 72 | 9 |
  | 20 | 95 | 10 |
  </details>

### :star: **Fewer database queries**

  The number of `select` queries required for a single call to `ListStakePools` is greatly reduced.
  <details><summary>Click to view data</summary><br>
  
  | Branch | Network | No. of Pools | No. of Queries |
  | :--- | :--- | ---: | ---: | 
  | `master` | `mainnet` |  1,124 |  12,365 |
  | `more-efficient-list-pools`| `mainnet` | 1,124  | 4 |
  </details>

### :heavy_check_mark: Properties

This PR adds a [property test](https://github.com/input-output-hk/cardano-wallet/blob/d00c608e47d2d189a282a02946cc8e46b843375b/lib/core/test/unit/Cardano/Pool/DB/Properties.hs#L987) for `listPoolLifeCycleData`, consisting of the following steps:
1. Add an arbitrary sequence of pool registration and retirement certificates to the database, for multiple pools.
2. Call `listPoolLifeCycleData` once to fetch lifecycle data for all active pools.
3. Call `readPoolLifeCycleStatus` multiple times, once for each known pool, and coalesce the results. 
4. Test that the results returned in steps 2 and 3 are identical.

### :heavy_check_mark: Live Data

Run the following command to get a list of all stake pools sorted by ID:
```sh
cardano-wallet stake-pool list --stake 1000 | jq 'sort_by(.id)' > stake-pool-list
```
The results should be identical for both `master` and this branch.

### System used for Benchmarking

The system used for benchmarking (4 cores):

```
vendor_id       : GenuineIntel
cpu family      : 6
model           : 85
model name      : Intel(R) Xeon(R) CPU
stepping        : 7
microcode       : 0x1
cpu MHz         : 2800.176
cache size      : 33792 KB
```

### DIFF
--- a/lib/core/src/Cardano/Pool/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Pool/DB/Sqlite.hs
@@ -47,7 +47,7 @@ import Cardano.DB.Sqlite
 import Cardano.Pool.DB
     ( DBLayer (..), ErrPointAlreadyExists (..), determinePoolLifeCycleStatus )
 import Cardano.Pool.DB.Log
-    ( PoolDbLog (..) )
+    ( ParseFailure (..), PoolDbLog (..) )
 import Cardano.Wallet.DB.Sqlite.Types
     ( BlockId (..) )
 import Cardano.Wallet.Logging
@@ -78,7 +78,7 @@ import Control.Monad.Trans.Except
 import Control.Tracer
     ( Tracer (..), contramap, natTracer, traceWith )
 import Data.Either
-    ( rights )
+    ( lefts, rights )
 import Data.Function
     ( (&) )
 import Data.Functor
@@ -211,9 +211,15 @@ newDBLayer trace fp timeInterpreter = do
 
             pure (foldl' toMap Map.empty production)
 
-        readTotalProduction =
-            Map.fromList . rights . fmap parseRow <$> rawSql query []
+        readTotalProduction = do
+            parsedRows <- fmap parseRow <$> rawSql query []
+            mapM_ onParseFailure $ lefts parsedRows
+            pure $ Map.fromList $ rights parsedRows
           where
+            onParseFailure = liftIO
+                . traceWith trace
+                . MsgParseFailure
+                . ParseFailure "readTotalProduction"
             query = T.unwords
                 [ "SELECT pool_id, count(pool_id) as block_count"
                 , "FROM pool_production"
@@ -369,20 +375,29 @@ newDBLayer trace fp timeInterpreter = do
                 ]
 
         listRetiredPools epochNo = do
-            let query = T.unwords
-                    [ "SELECT *"
-                    , "FROM active_pool_retirements"
-                    , "WHERE retirement_epoch <= ?;"
-                    ]
-            let parameters = [ toPersistValue epochNo ]
-            let safeCast (Single poolId, Single retirementEpoch) =
-                    PoolRetirementCertificate
-                        <$> fromPersistValue poolId
-                        <*> fromPersistValue retirementEpoch
-            rights . fmap safeCast <$> rawSql query parameters
+            parsedRows <- fmap parseRow <$> rawSql query parameters
+            mapM_ onParseFailure $ lefts parsedRows
+            pure $ rights parsedRows
+          where
+            query = T.unwords
+                [ "SELECT *"
+                , "FROM active_pool_retirements"
+                , "WHERE retirement_epoch <= ?;"
+                ]
+            parameters = [ toPersistValue epochNo ]
+            onParseFailure = liftIO
+                . traceWith trace
+                . MsgParseFailure
+                . ParseFailure "listRetiredPools"
+            parseRow (Single poolId, Single retirementEpoch) =
+                PoolRetirementCertificate
+                    <$> fromPersistValue poolId
+                    <*> fromPersistValue retirementEpoch
 
-        listPoolLifeCycleData epochNo =
-            rights . fmap parseRow <$> rawSql query parameters
+        listPoolLifeCycleData epochNo = do
+            parsedRows <- fmap parseRow <$> rawSql query parameters
+            mapM_ onParseFailure $ lefts parsedRows
+            pure $ rights parsedRows
           where
             query = T.unwords
                 [ "SELECT *"
@@ -390,6 +405,10 @@ newDBLayer trace fp timeInterpreter = do
                 , "WHERE retirement_epoch IS NULL OR retirement_epoch > ?;"
                 ]
             parameters = [ toPersistValue epochNo ]
+            onParseFailure = liftIO
+                . traceWith trace
+                . MsgParseFailure
+                . ParseFailure "listPoolLifeCycleData"
             parseRow
                 ( Single fieldPoolId
                 , Single fieldRetirementEpoch
--- a/lib/core/src/Cardano/Pool/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Pool/DB/Sqlite.hs
@@ -70,7 +70,7 @@ import Cardano.Wallet.Unsafe
 import Control.Exception
     ( bracket, throwIO )
 import Control.Monad
-    ( forM )
+    ( forM, forM_ )
 import Control.Monad.IO.Class
     ( liftIO )
 import Control.Monad.Trans.Except
@@ -78,7 +78,7 @@ import Control.Monad.Trans.Except
 import Control.Tracer
     ( Tracer (..), contramap, natTracer, traceWith )
 import Data.Either
-    ( lefts, rights )
+    ( partitionEithers, rights )
 import Data.Function
     ( (&) )
 import Data.Functor
@@ -104,6 +104,8 @@ import Data.Word
 import Database.Persist.Sql
     ( Entity (..)
     , Filter
+    , PersistValue
+    , RawSql
     , SelectOpt (..)
     , Single (..)
     , deleteWhere
@@ -211,15 +213,9 @@ newDBLayer trace fp timeInterpreter = do
 
             pure (foldl' toMap Map.empty production)
 
-        readTotalProduction = do
-            parsedRows <- fmap parseRow <$> rawSql query []
-            mapM_ onParseFailure $ lefts parsedRows
-            pure $ Map.fromList $ rights parsedRows
+        readTotalProduction = Map.fromList <$> runRawQuery trace
+            (RawQuery "readTotalProduction" query [] parseRow)
           where
-            onParseFailure = liftIO
-                . traceWith trace
-                . MsgParseFailure
-                . ParseFailure "readTotalProduction"
             query = T.unwords
                 [ "SELECT pool_id, count(pool_id) as block_count"
                 , "FROM pool_production"
@@ -374,10 +370,8 @@ newDBLayer trace fp timeInterpreter = do
                 , Desc PoolRegistrationSlotInternalIndex
                 ]
 
-        listRetiredPools epochNo = do
-            parsedRows <- fmap parseRow <$> rawSql query parameters
-            mapM_ onParseFailure $ lefts parsedRows
-            pure $ rights parsedRows
+        listRetiredPools epochNo = runRawQuery trace $
+            RawQuery "listRetiredPools" query parameters parseRow
           where
             query = T.unwords
                 [ "SELECT *"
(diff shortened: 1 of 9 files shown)


jonathanknowles:
bors try

piotr-iohk:
Awesome :tada: 

My results of listing stake-pools with n wallets syncing on this branch (Intel i7 - 6 cores, 32GB RAM, SSD):

```
0 wallets    - 2-3s
10 wallets   - 4-6s
20 wallets   - 4-5s
50 wallets   - 4-5s
100 wallets  - 5-6s
200 wallets  - 7-8s
```

jonathanknowles:
bors try

jonathanknowles:
bors try

jonathanknowles:
bors try

jonathanknowles:
bors try

jonathanknowles:
bors cancel

jonathanknowles:
bors try

lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -208,14 +211,20 @@ newDBLayer trace fp timeInterpreter = do
...
-        readTotalProduction = do
-            production <- fmap entityVal <$>
-                selectList ([] :: [Filter PoolProduction]) []
-
-            let toMap m (PoolProduction{poolProductionPoolId}) =
-                    Map.insertWith (+) poolProductionPoolId 1 m
-
-            pure $ Map.map Quantity $ foldl' toMap Map.empty production
+        readTotalProduction =
+            Map.fromList . rights . fmap parseRow <$> rawSql query []
```
hasufell: This looks like we hide parse errors?

lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -372,6 +381,55 @@ newDBLayer trace fp timeInterpreter = do
                         <*> fromPersistValue retirementEpoch
             rights . fmap safeCast <$> rawSql query parameters
 
+        listPoolLifeCycleData epochNo =
+            rights . fmap parseRow <$> rawSql query parameters
```
hasufell: Same here, looks like we hide parse errors.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
jonathanknowles: > This looks like we hide parse errors?

For this particular function, I think it would be appropriate to fail fast if there's even a single parse error. It's a very simple `group by` query that should not have any failures at all.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
jonathanknowles: > looks like we hide parse errors.

It does indeed!

I previously added a patch that would cause this function to fail fast in the face of a parse error. However, that had the disadvantage that a single row failing to parse would cause the entire operation to fail.

After giving this some thought, I think the following method would be more appropriate, especially from a user experience perspective:

- for rows that fail to parse: create a log entry for each error.
- for rows that parse successfully: return these as part of the result.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
jonathanknowles: Fixed in 6fbc807c47cfea70f3f7c7e4fa628048cc703ac6.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
jonathanknowles: Fixed in 79f505c41ab68c2bc6d134b43071163add14a3be.

jonathanknowles:
bors try

jonathanknowles:
bors cancel

rvl:
bors try-

rvl:
bors try

jonathanknowles:
bors try

jonathanknowles:
bors try

lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -208,14 +211,24 @@ newDBLayer trace fp timeInterpreter = do
...
-                    Map.insertWith (+) poolProductionPoolId 1 m
-
-            pure $ Map.map Quantity $ foldl' toMap Map.empty production
+        readTotalProduction =
+            Map.fromList . fmap (assertSuccess . parseRow) <$> rawSql query []
+          where
+            assertSuccess = flip either id $ \e -> error $ mconcat
+                [ "readTotalProduction: unexpected parse failure: "
+                , show e
+                ]
```
hasufell: Since we're in IO it's best to throw it as IO errors. `error` triggers on evaluation and can leak through exception handlers. Probably not a big concern here, but better safe than sorry.

```suggestion
        readTotalProduction = do
            rows <- rawSql query []
            parsed <- forM rows $ \row -> case parseRow row of
                Right x -> pure x
                Left e  -> liftIO . throwIO . userError
                    . ("readTotalProduction: unexpected parse failure: " ++)
                    . T.unpack $ e
            pure $ Map.fromList parsed
          where
```

jonathanknowles: Thanks @hasufell, I agree it's better to record parse failures eagerly, as soon as they happen, rather than relying on the lazy evaluation of `error`.

In the end I changed this to use the logging mechanism instead, as I think this will make any errors easier to analyze.

hasufell: Logging doesn't short-circuit and doesn't crash the handler, though. From what I remember, we don't expect parse failures here ever?

hasufell:
I'm not sure what the exact policies wrt committing are, but I feel 29 commits is a bit hard to follow and they should be squashed. After all, it's one atomic change with one topic. Cherry-picking, reverting, bisecting etc will become a lot harder with 29 commits that are inter-connected.

jonathanknowles:
> I'm not sure what the exact policies wrt committing are

I usually try to structure each PR into a series of smaller commits, where each commit:

* is self-contained;
* has a clear logical purpose;
* compiles with no errors;
* passes the test suite*.

The overall goal is a sequence of commits that tells a **_coherent story_** of how the code evolves from one state to another.

> but I feel 29 commits is a bit hard to follow

Perhaps the [Files Changed](https://github.com/input-output-hk/cardano-wallet/pull/2111/files) view will help? This view squashes all commits together, if you prefer to view commits that way.

> Cherry-picking, reverting, bisecting etc will become a lot harder with 29 commits that are inter-connected.

In my experience, when we've discovered bugs not covered by an existing test suite, it's much easier to discover the source of the bug if individual commits are retained, provided that we followed the rules I mentioned above.

With the "one big commit" approach, it often takes quite a bit more effort to figure out where the error was introduced.

For this reason, I generally am not in favour of squashing well-curated histories.

However, if the history is incoherent, or if the history has changes in direction, I am in favour of squashing to clean the history up.

---
* I think there are some useful exceptions to this rule, such as: deliberately introducing a test that fails in one commit (to demonstrate failure), and then fixing it in the next commit.

hasufell:
> each commit is an atomic change that does just one thing

Well, IMO, atomic doesn't mean *small*. Atomic means **self-evident**: the commit can be understood by itself by looking at the changes and the commit message. This isn't the case for most of the commits in this PR. It also doesn't mean a PR must consist of one commit only.

Tight logical dependencies between commits is a sign of non-atomicity.

> In my experience, when we have discovered bugs not covered by an existing test suite, it's much easier to discover the source of the bug if commits are kept small and atomic.

To discover the source of bug you have to read further down the commit history to understand the context, even after you found the bad commit. Now you have to read 10+ commits and still won't know what it does, unless you find and open this PR.

jonathanknowles:
> Well, IMO, atomic doesn't mean _small_. 

By an **_atomic_** commit, I mean that:
- applying the commit will not break the build.
- applying the commit will not cause the tests to fail (unless we are purposefully introducing a failing test that will be fixed by a later commit).
- the commit has **_just one_** purpose, which can be:
    - a refactoring, or
    - a behavioural change (but **_not both_**).

Typically, I prefer to keep refactorings and behavioural changes in **_separate commits_**, as described in [this blog article](http://jakegoulding.com/blog/2012/11/04/refactor-and-make-changes-in-different-commits/).

> Atomic means **self-evident**: the commit can be understood by itself by looking at the changes and the commit message. This isn't the case for most of the commits in this PR.

Can you give an example of a commit in this PR that cannot be understood by looking at the changes and the commit message?

I'd like to better understand your perspective.

hasufell:
> By an atomic commit, I mean that:
> 
>     applying the commit will not break the build.
>     applying the commit will not cause the tests to fail (unless we are purposefully introducing a failing test that will be fixed by a later commit).

I don't think either of these are proper indicators of atomicity. You can write a single logical change in 200 commits without breaking the build or the tests.

>     the commit has just one purpose, which can be:
>         a refactoring, or
>         a behavioural change (but not both).

Yes, I believe this PR could have 2, maybe 3 commits due to this distinction.

> Can you give an example of a commit in this PR that cannot be understood by looking at the changes and the commit message?

https://github.com/input-output-hk/cardano-wallet/pull/2111/commits/161068e76cef7563da09779d1a2f481ea635797b

A type was added to the test suite with an arbitrary instance. It's not self-evident why the type exist, what test it will be used for and why that test (that isn't in this commit) might even be interesting to have.

jonathanknowles:
> Yes, I believe this PR could have 2, maybe 3 commits due to this distinction.

In that case, I suspect we have very different philosophies about how to structure a PR into separate commits. :cat: 

**I'm happy to discuss the topic with you further, but could we discuss it over a call?**

I'd very much like to keep the discussion for this PR focused on its **content**, rather than the **general** question about _how to break up a PR into multiple commits_. (There are clearly multiple schools of thought about how to do this.) I find written discussions can sometimes consume a lot of time, especially when parties have very different points of view to reconcile. I think a call would help us to reach a compromise more quickly.

> A type was added to the test suite with an arbitrary instance. It's not self-evident why the type exist, what test it will be used for and why that test (that isn't in this commit) might even be interesting to have.

The `MultiPoolCertificateSequence` type will be used by two later commits:
1. 8af20d89e733c0beaeb109ef0d96806703cb6af5: (Uses `MultiPoolCertificateSequence` to simplify a property.)
2. 28e734cbe3a37cbe8af0abd6e0895b50aa932a65: (Uses `MultiPoolCertificateSequence` in a completely new property.)

I don't think these commits should be squashed together, as they have different purposes.

hasufell:
> In that case, I suspect we have very different philosophies about how to structure a PR into separate commits.

Sure, I'm suspecting everyone is free to follow their own way of structuring commits then.

rvl (approved):
This is a good improvement.

Automated performance testing is always good... a simple count of queries should be enough in most cases. Perhaps it's possible to test that a call to `listPoolLifeCycleData` never causes more than (say) 5 SQL queries.

Let's have the squabble about boring SQL on slack.

Issue #2051, opened 2020-08-19
Creating an address takes a massive amount of time on large wallets


KtorZ:
# Context
 
 

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      |  |
| Platform     |           |
| Installation |             |

# Steps to Reproduce

1. Have a large Byron wallet (100K+ transactions, 50K+ UTxO, 50K+ existing addresses).
2. Import a single address using the `/byron-wallets/addresses` endpoint

## Expected behavior

1. The address is promptly imported.

## Actual behavior
 

1. The wallet hangs for several minutes before completing the operation.

---

# Resolution 

- A possible started would be to look more closely at the strictness of the `scanl` used in the handler.

---

# QA 

- Benchmark results from last night (https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/636#e79c244b-d2c0-43a5-8d24-36d07f5bc7bf)

```
BenchRndResults:
  importOneAddressTime: 1.510 ms
  importManyAddressesTime: 187.3 ms
  utxoStatistics:
    = Total value of -0 lovelace across 0 UTxOs
 
BenchRndResults:
  importOneAddressTime: 328.9 ms
  importManyAddressesTime: 789.7 ms
  utxoStatistics:
    = Total value of 21515725326443 lovelace across 747 UTxOs
 
BenchRndResults:
  importOneAddressTime: 942.2 ms
  importManyAddressesTime: 1.886 s
  utxoStatistics:
    = Total value of 31020480179736 lovelace across 1392 UTxOs

BenchRndResults:
  importOneAddressTime: 1.770 s
  importManyAddressesTime: 3.551 s
  utxoStatistics:
    = Total value of 69284306530582 lovelace across 2779 UTxOs
```

Note: many = 1000

KtorZ:
Even on a rather "small" wallet, importing 10K addresses took 5s, and then, importing a new address took more than 10minute. So there's something quite fishy.

KtorZ:
On different runs, it seems like the behaviour is not entirely consistent. Somehow, the stake pool restoration worker also seem to have an impact when creating addresses. Once the wallet is fully synced and done fetching metadata, it seems like creating new addresses is done in a decent amount of time. I have however been very inconsistent in my testing here so I am planning to extend the nightly restoration benchmark to see how this actually performs there.

piotr-iohk:
lgtm

PR #2112, opened 2020-09-04
simplify implementation of 'estimateMaxNumberOfInputs' & increase coverage

KtorZ:
# Issue Number

#2051 

# Overview

- I've added scenarios for Byron and Icarus keys
- I've added a 'within' clause to some scenarios in order to make sure that the properties are executed within a reasonable time.

# Comments

This didn't really improve execution time significantly, but it is still useful to keep.

### DIFF
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
@@ -310,16 +310,21 @@ _estimateMaxNumberOfInputs
     -> Word8
     -- ^ Number of outputs in transaction
     -> Word8
-_estimateMaxNumberOfInputs networkId (Quantity maxSize) md nOuts =
-    findMax minBound
+_estimateMaxNumberOfInputs networkId txMaxSize md nOuts =
+    findLargestUntil ((> maxSize) . txSizeGivenInputs) 0
   where
-    findMax :: Word8 -> Word8
-    findMax inf
-        | inf == maxBound    = 0
-        | isTooBig (inf + 1) = inf
-        | otherwise          = findMax (inf + 1)
+    -- | Find the largest amount of inputs that doesn't make the tx too big.
+    -- Tries in sequence from 0 and upward (up to 255, but smaller than 50 in
+    -- practice because of the max transaction size).
+    findLargestUntil :: (Word8 -> Bool) -> Word8 -> Word8
+    findLargestUntil isTxTooLarge inf
+        | inf == maxBound        = maxBound
+        | isTxTooLarge (inf + 1) = inf
+        | otherwise              = findLargestUntil isTxTooLarge (inf + 1)
 
-    isTooBig nInps = size > fromIntegral maxSize
+    maxSize = fromIntegral (getQuantity txMaxSize)
+
+    txSizeGivenInputs nInps = size
       where
         size = computeTxSize networkId (txWitnessTagFor @k) md Nothing sel
         sel  = dummyCoinSel (fromIntegral nInps) (fromIntegral nOuts)
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
@@ -1,5 +1,4 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
-{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleContexts #-}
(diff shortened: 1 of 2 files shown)


lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -312,28 +311,18 @@ _estimateMaxNumberOfInputs
...
-
-    growingFactor = 2
-
-    lowerBound = upperBound `div` growingFactor
-    upperBound = upperBound_ 1
-      where
-        upperBound_ !n | isTooBig n = n
-                       | otherwise  = upperBound_ (n*growingFactor)
+    findMax :: Word8 -> Word8
+    findMax inf
```
Anviking: `findMax minBound` first looks deceptively simple... until you start looking at it. Can we give give an overview about what is going on here? E.g. `Find the largest amount of inputs that doesn't make the tx too big. Tries in sequence from 0 and upward.`

Also, btw, why did you decide to remove the bisection? Just simplicity?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -312,28 +311,18 @@ _estimateMaxNumberOfInputs
...
+        | inf == maxBound    = 0
```
Anviking: What does `inf == maxBound` mean? If `1` input would be too much, `findMax 0` should return 0. But it _also_ returns 0 if all possibilities pass? That seems like an error. (If the max tx size is really big...)

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -312,28 +311,18 @@ _estimateMaxNumberOfInputs
...
+        | isTooBig (inf + 1) = inf
+        | otherwise          = findMax (inf + 1)
 
     isTooBig nInps = size > fromIntegral maxSize
```
Anviking: What would you think about
1. Calling this `txTooBig`
2. Passing it explicitly to the `findMax` function

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
Anviking: Also I think it might make sense to take the condition explicitly (see other comment).

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -162,23 +183,25 @@ spec = do
         res `shouldBe` Right (FeeEstimation 165413 165413)
 
 estimateMaxInputsTests
-    :: Cardano.NetworkId
+    :: forall k. (TxWitnessTagFor k, Typeable k)
+    => Cardano.NetworkId
+    -> [(Word8, Word8)]
```
Anviking: Doc comment? 🙏 

I think a name like _estimateMaxInputs**GivenOutputs**Tests_ would make the call-site tuples more easy to understand too.

Anviking: > I think a name like estimateMaxInputsGivenOutputsTests would make the call-site tuples more easy to understand too.

Ok, I see `estimateMaxInputsTests` is not _only_ goldens. Then I'm less sure...

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -162,23 +183,25 @@ spec = do
         res `shouldBe` Right (FeeEstimation 165413 165413)
 
 estimateMaxInputsTests
-    :: Cardano.NetworkId
+    :: forall k. (TxWitnessTagFor k, Typeable k)
```
Anviking, line 207: 👍

Anviking (commented):
Looks good; a few comments

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -211,39 +234,48 @@ prop_decodeSignedByronTxRoundtrip (DecodeByronSetup utxo outs slotNo network pai
...
+    => Cardano.NetworkId
     -> Quantity "byte" Word16
     -> Word8
     -> Property
-prop_moreOutputsMeansLessInputs net size nOuts = withMaxSuccess 1000 $
-    nOuts < maxBound ==>
-        _estimateMaxNumberOfInputs @ShelleyKey net size Nothing nOuts
+prop_moreOutputsMeansLessInputs net size nOuts
+    = withMaxSuccess 1000
+    $ within 100000
```
Anviking, line 265: 👌 👍

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
KtorZ: > Also, btw, why did you decide to remove the bisection? Just simplicity?

I originally suspected that the bisection could have been causing some slowness, taking to long to find a fix point. I actually measured the time and observed some very long delays until I realized I was measuring the wrong thing. In the meantime, I had changed the implementation to `findMax` which turned out to be slightly master but also much simpler so I kept it.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
KtorZ: Actually you're right, this should just return `maxBound` :+1:

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
KtorZ: 1. Sounds good.
2. Okay, but why ? What's the added benefit in this context?

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
KtorZ: > estimateMaxInputsGivenOutputsTests

Estimating max inputs is **always** given some outputs though.

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
KtorZ: I am _slightly_ worried about Windows here which it notably slower. I can go down to 50000 on my machine without issue, so I took an order of magnitude higher _just to be sure_.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
Anviking: I think `findMax :: Word8 -> Word8` is a kind-of deceptive name and type-signature. It sounds abstract (at first glance), but it is not. I believe it is meant more as a helper of `estimateMaxNumberOfInputs`.

I think it should either _be_ abstract, _or_ have a more specialised name.

I think explicitly showing the dependency on the `txTooBig` condition might be clearer in both cases, but in particular, it _would_ make `findMax` abstract (assuming no other details specific to this use-case).

Modulo, if the type-signature becomes too ugly, perhaps.

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
Anviking: Yes, but I think these tests should be easily understood without being 100% sure about that.

Adding a doc comment to `[(Word8, Word8)]` definitely helps.

But my idea with `estimateMaxInputsGivenOutputsTests` was to hint the same, already at the _call-site_, similar to  the impossible `[(outputs: 1, maxInputs: 2)]`, or the perhaps overkill `EstimateMaxInputsGolden { outputs, maxInputs :: Word8  }`.

Anviking (approved):

PR #2116, opened 2020-09-07
Improve fee estimation performance / fix accidentally quadratic function

KtorZ:
# Issue Number

#2032 & rel

# Overview

- ca9dd1ea2ded5b3539d7362f4fcb74fd2bcadf94
  :round_pushpin: **remove use of DerivingVia in Sqlite types: can't build with --profile with it**
  
- bc1bacc106faa4b17ad2a3cecf75eb29fb459bd6
  :round_pushpin: **avoid re-creating and folding on a map to compute balance of a list of resolved inputs**
    This is somehow causing massive slowness in the fee estimation. Changing this reduces the time in benchmarks from 18,000ms to 3,000ms!!

- f82a7fed3903f46409456c41ec46b292e2fa4bc6
  :round_pushpin: **use standard random generator for picking random UTxO**
    There's really no need for strong cryptographic randomness at this level, this is needlessly slower than it could be an called pretty often during fee calculation.

- 94d4812d459951fef3fa999008c35d9d781b4f0c
  :round_pushpin: **do not re-compute the full balance after picking new inputs**
    Instead, compute the balance incrementally. The 'coverRemainingFee' was computing the balance of the extra inputs selected to cover for fee. In worst-case scenario where many inputs needed to be selected, for example, 1000 extra inputs, it'll recompute the balance of an ever growing list a thousand times! Not only this is accidentally quadratic, but, it is also completely unbounded and will potentially select way too many inputs. In theory, this should not select more than the max allowed inputs. Next commit will fix this.

# Comments

Before:

```
Latency with 500 utxos scenario
    postTransactionFee  - 16756.3 ms
```

After:

```
    Latencies for 2 fixture wallets with 500 utxos scenario
        postTransactionFee  - 119.0 ms
```

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Random.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Random.hs
@@ -38,8 +38,6 @@ import Control.Monad.Trans.Except
     ( ExceptT (..), except )
 import Control.Monad.Trans.Maybe
     ( MaybeT (..), runMaybeT )
-import Crypto.Random.Types
-    ( MonadRandom )
 import Data.Functor
     ( ($>) )
 import Data.List.NonEmpty
@@ -109,12 +107,12 @@ data TargetRange = TargetRange
 -- we set.
 -- @
 random
-    :: forall m e. MonadRandom m
+    :: forall e. ()
     => CoinSelectionOptions e
     -> NonEmpty TxOut
     -> Quantity "lovelace" Word64
     -> UTxO
-    -> ExceptT (ErrCoinSelection e) m (CoinSelection, UTxO)
+    -> ExceptT (ErrCoinSelection e) IO (CoinSelection, UTxO)
 random opt outs (Quantity withdrawal) utxo = do
     let descending = NE.toList . NE.sortBy (flip $ comparing coin)
     let nOuts = fromIntegral $ NE.length outs
@@ -143,10 +141,9 @@ data SelectionState = SelectionState
 
 -- | Perform a random selection on a given output, without improvement.
 makeSelection
-    :: forall m. MonadRandom m
-    => SelectionState
+    :: SelectionState
     -> TxOut
-    -> MaybeT m SelectionState
+    -> MaybeT IO SelectionState
 makeSelection (SelectionState maxN utxo0 withdrawal0 selection0) txout = do
     (selection', utxo') <- coverRandomly ([], utxo0)
     return $ SelectionState
@@ -159,9 +156,8 @@ makeSelection (SelectionState maxN utxo0 withdrawal0 selection0) txout = do
     TargetRange{targetMin} = mkTargetRange $ getCoin $ coin txout
 
     coverRandomly
-        :: forall m. MonadRandom m
-        => ([(TxIn, TxOut)], UTxO)
-        -> MaybeT m (CoinSelection, UTxO)
+        :: ([(TxIn, TxOut)], UTxO)
+        -> MaybeT IO (CoinSelection, UTxO)
     coverRandomly (inps, utxo)
         | L.length inps > fromIntegral maxN =
             MaybeT $ return Nothing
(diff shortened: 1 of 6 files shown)


lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Random.hs
```diff
@@ -109,12 +107,12 @@ data TargetRange = TargetRange
 -- we set.
 -- @
 random
-    :: forall m e. MonadRandom m
+    :: forall e. ()
```
rvl, line 110: Perhaps in future, rather than using IO, we could explicity pass around the DRG state. Or maybe provide an infinite list of random UTxO to the function (if possible).

lib/core/src/Cardano/Wallet/Primitive/Fee.hs
```diff
@@ -239,20 +234,19 @@ senderPaysFee opt utxo sel = evalStateT (go sel) utxo where
...
-    => Fee
-    -> StateT UTxO (ExceptT ErrAdjustForFee m) [(TxIn, TxOut)]
-coverRemainingFee (Fee fee) = go [] where
-    go acc
-        | balance' acc >= fee =
-            return acc
+    :: Fee
+    -> StateT UTxO (ExceptT ErrAdjustForFee IO) ([(TxIn, TxOut)], Coin)
+coverRemainingFee (Fee fee) = go [] 0 where
+    go additionalInputs surplus
```
rvl, line 240: :+1:

lib/core/src/Cardano/Wallet/Primitive/Fee.hs
```diff
@@ -239,20 +234,19 @@ senderPaysFee opt utxo sel = evalStateT (go sel) utxo where
...
+        | surplus >= fee =
+            return (additionalInputs, Coin surplus)
         | otherwise = do
             -- We ignore the size of the fee, and just pick randomly
             StateT (lift . pickRandom) >>= \case
-                Just entry ->
-                    go (entry : acc)
+                Just input@(_, out) ->
+                    go (input : additionalInputs) (getCoin $ coin out)
```
rvl: Does the value of `out` need to be added to `surplus` here?

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -1231,7 +1228,10 @@ balance =
 -- | Compute the balance of a unwrapped UTxO
 balance' :: [(TxIn, TxOut)] -> Word64
 balance' =
-    fromIntegral . balance . UTxO . Map.fromList
+    foldl' fn 0
```
rvl, line 1231: What about just `sum . map (getCoin . coin . snd)`?

rvl (commented):
Very nice improvement but I have 2 questions.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Random.hs
KtorZ: Yes, testing was made a little harder because of this, plus, we also now loose the nice type signature that was nicely indicating which side-effects were needed. I might as well flag it as a technical debt to address later on and restore some purity in this.

lib/core/src/Cardano/Wallet/Primitive/Fee.hs
KtorZ: :facepalm: .... yes. That's the whole point.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
KtorZ: I've went for sum initially and switch to foldl' for consistency with the other balance calculation. It makes it more obvious that the same thing is happening.

lib/core/src/Cardano/Wallet/Primitive/Fee.hs
KtorZ: Fortunately our test suite catches it too :muscle:

KtorZ: Aaaand, fixed, tests are now passing :)

rvl (approved):

PR #2117, opened 2020-09-07
make sure to provide an upper bound for max inputs when adjusting fees

KtorZ:
# Issue Number

TODO (discovered as part of the slow fee estimation investigation, this is slightly related but actually a bug).

# Overview

In order to prevent transactions from becoming too big, we compute an estimated maximum number of inputs that can be selected. This is well respected by coin selection algorithms but was disregarded by the fee adjustment function which would basically pick UTxO until it has either covered fees or depleted the entire UTxO. Although in practice it should be rarely necessary to pick more than 1 extra input, there are edge-cases where we may need many (especially in wallets with a lot of dust). Picking inputs without any upper bound could cause a transaction to go beyond an acceptable size, and be later rejected.

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -1203,8 +1203,9 @@ feeOpts
     -> Maybe TxMetadata
     -> W.TxParameters
     -> W.Coin
+    -> CoinSelection
     -> FeeOptions
-feeOpts tl action md txp minUtxo = FeeOptions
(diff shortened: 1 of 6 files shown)


rvl (approved):
The diff shown in github looks a little bit mixed up with #2116.
I just looked at senderPaysFee and coverRemainingFee and these changes look good.
And the unit test looks fine.

KtorZ:
@rvl Yeah, I need to rebase the branch here, will do once #2116 is merged.

Issue #2120, opened 2020-09-08
Wallet address pool gap is limited to 232


nikolaglumac:
# Context
 
When trying to restore a wallet with a `address_pool_gap` set to `1000` you end up with a wallet with `address_pool_gap` set to `232`.

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | 3feb1d09a4e37b46dcec92585870d8490b6bdf78 |
| Platform     | All         |
| Installation | Daedalus            |

# Steps to Reproduce

1. Trigger wallet restoration with the `address_pool_gap` set to `1000`:
```
"{"name":"Super wallet","mnemonic_sentence":["little","inch","art","moral","entire","alert","blossom","innocent","judge","flower","load","match","what","bargain","museum","sea","crush","rely","baby","rally","during","keen","borrow","carpet"],"passphrase":"Secret1234","address_pool_gap":1000}"
```

## Expected behavior

Wallet is restored and it's `address_pool_gap` is set to `1000`.

## Actual behavior
 
Wallet is restored and it's `address_pool_gap` is set to `232`:
```
"{"passphrase":{"last_updated_at":"2020-09-08T12:27:35.595262Z"},"address_pool_gap":232,"state":{"status":"syncing","progress":{"quantity":0,"unit":"percent"}},"balance":{"reward":{"quantity":0,"unit":"lovelace"},"total":{"quantity":0,"unit":"lovelace"},"available":{"quantity":0,"unit":"lovelace"}},"name":"Super wallet","delegation":{"next":[],"active":{"status":"not_delegating"}},"id":"82324b69f0c16847ab93977a23f4ae2b2f4dc8cc","tip":{"height":{"quantity":0,"unit":"block"},"epoch_number":0,"slot_number":0}}"
```

---

# Resolution 

- Fix the database internal type used for serializing and deserializing address pool gap ==> #2121

---

# QA 

- Test integration scenarios extended to have extra assertions controlling the value of the gap: 
    https://github.com/input-output-hk/cardano-wallet/blob/8e0c0cf1c3c65e6c5f8dd1f2ead008641cef77b9/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs#L457-L470

KtorZ:
Note for whoever pick this:

```
instance PersistFieldSql AddressPoolGap where
    sqlType _ = sqlType (Proxy @Word8)
```

and 

```
1000 `mod` 256 = 232
```

PR #2121, opened 2020-09-08
Fix wallet address pool gap limited to Word8

hasufell:
Although with Persistent Word32 will change the
underlying type from SqlInt32 to SqlInt64, it
doesn't matter for sqlite, since they all end
up as INTEGER.

https://www.sqlite.org/datatype3.html

Issue #2120

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
@@ -473,8 +473,11 @@ spec = do
             let payload = payloadWith' "Secure Wallet" mnemonics24 (fromIntegral addrPoolGap)
             rW <- request @ApiWallet ctx (Link.postWallet @'Shelley) Default payload
             verify rW expectations
+            let w = getFromResponse id rW
             rA <- request @[ApiAddress n] ctx
-                (Link.listAddresses @'Shelley (getFromResponse id rW)) Default Empty
+                (Link.listAddresses @'Shelley w) Default Empty
+            _ <- request @ApiWallet ctx
+                (Link.deleteWallet @'Shelley w) Default Empty
             verify rA
                 [ expectListSize addrPoolGap
                 ]
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
@@ -470,8 +470,11 @@ spec = do
                 ]
         forM_ matrix $ \(title, addrPoolGap, expectations) -> it title $ \ctx -> do
             let payload = payloadWith' "Secure Wallet" mnemonics24 (fromIntegral addrPoolGap)
-            r <- request @ApiWallet ctx (Link.postWallet @'Shelley) Default payload
+            r@(_, Right wallet) <- request @ApiWallet ctx (Link.postWallet @'Shelley) Default payload
             verify r expectations
+            addrs <- listAddresses @n ctx wallet
+            length addrs `shouldBe` addrPoolGap
+
 
     it "WALLETS_CREATE_08 - default address_pool_gap" $ \ctx -> do
         let payload = Json [json| {
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
@@ -116,6 +116,7 @@ import Test.Integration.Framework.TestData
     , mnemonics24
     , mnemonics9
     , payloadWith
+    , payloadWith'
     , polishWalletName
     , russianWalletName
     , simplePayload
@@ -462,16 +463,13 @@ spec = do
                   )
                 , ( show addrPoolMax
                   , addrPoolMax
-                  , [ expectResponseCode @IO HTTP.status201 ]
+                  , [ expectResponseCode @IO HTTP.status201
+                    , expectField (#addressPoolGap . #getApiT) (`shouldBe` maxBound)
+                  ]
                   )
                 ]
         forM_ matrix $ \(title, addrPoolGap, expectations) -> it title $ \ctx -> do
-            let payload = Json [json| {
-                    "name": "Secure Wallet",
-                    "mnemonic_sentence": #{mnemonics24},
-                    "passphrase": "Secure passphrase",
-                    "address_pool_gap": #{addrPoolGap}
-                    } |]
+            let payload = payloadWith' "Secure Wallet" mnemonics24 (fromIntegral addrPoolGap)
             r <- request @ApiWallet ctx (Link.postWallet @'Shelley) Default payload
             verify r expectations
 
@@ -524,6 +522,7 @@ spec = do
                     } |]
             r <- request @ApiWallet ctx (Link.postWallet @'Shelley) headers payload
             verify r expectations
+
 
     it "WALLETS_GET_01 - can get wallet details" $ \ctx -> do
         (_, w) <- unsafeRequest @ApiWallet ctx (Link.postWallet @'Shelley) simplePayload
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
@@ -28,6 +28,7 @@ import Cardano.Mnemonic
     )
 import Cardano.Wallet.Api.Types
     ( AddressAmount (..)
+    , ApiAddress
     , ApiByronWallet
     , ApiCoinSelection
     , ApiNetworkInformation
@@ -470,11 +471,13 @@ spec = do
                 ]
         forM_ matrix $ \(title, addrPoolGap, expectations) -> it title $ \ctx -> do
             let payload = payloadWith' "Secure Wallet" mnemonics24 (fromIntegral addrPoolGap)
-            r@(_, Right wallet) <- request @ApiWallet ctx (Link.postWallet @'Shelley) Default payload
-            verify r expectations
-            addrs <- listAddresses @n ctx wallet
-            length addrs `shouldBe` addrPoolGap
-
+            rW <- request @ApiWallet ctx (Link.postWallet @'Shelley) Default payload
+            verify rW expectations
+            rA <- request @[ApiAddress n] ctx
+                (Link.listAddresses @'Shelley (getFromResponse id rW)) Default Empty
+            verify rA
+                [ expectListSize addrPoolGap
+                ]
 
     it "WALLETS_CREATE_08 - default address_pool_gap" $ \ctx -> do
         let payload = Json [json| {
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
@@ -28,7 +28,6 @@ import Cardano.Mnemonic
     )
 import Cardano.Wallet.Api.Types
     ( AddressAmount (..)
-    , ApiAddress
     , ApiByronWallet
     , ApiCoinSelection
     , ApiNetworkInformation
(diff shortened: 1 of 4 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
```diff
@@ -525,6 +526,33 @@ spec = do
             r <- request @ApiWallet ctx (Link.postWallet @'Shelley) headers payload
             verify r expectations
 
+    -- https://github.com/input-output-hk/cardano-wallet/issues/2120
+    it "WALLETS_CREATE_10 - Address pool gap isn't limited to Word8" $ \ctx -> do
```
piotr-iohk: Would you mind adjusting also test `WALLETS_CREATE_08 - address_pool_gap` to have assertion for `addrPoolMax`... somehow it is missing it.

hasufell: I have a feeling this was omitted precisely of that bug...

hasufell: @piotr-iohk I believe this one is enough then, even and I can delete the other?

lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs
```diff
@@ -393,12 +393,12 @@ instance PersistField AddressPoolGap where
...
     fromPersistValue pv = fromPersistValue >=> mkAddressPoolGap' $ pv
       where
-        mkAddressPoolGap' :: Word8 -> Either Text AddressPoolGap
+        mkAddressPoolGap' :: Word32 -> Either Text AddressPoolGap
         mkAddressPoolGap' = first msg . mkAddressPoolGap . fromIntegral
         msg e = T.pack $ "not a valid value: " <> show pv <> ": " <> show e
 
 instance PersistFieldSql AddressPoolGap where
-    sqlType _ = sqlType (Proxy @Word8)
+    sqlType _ = sqlType (Proxy @Word32)
```
hasufell, line 401: So this corresponds to the newtype

```hs
newtype AddressPoolGap = AddressPoolGap
    { getAddressPoolGap :: Word32 }
    deriving stock (Generic, Show, Eq, Ord)
```

I'm not sure if there is a way to get the inner type of a newtype. Maybe with generics?

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
```diff
@@ -462,16 +463,13 @@ spec = do
                   )
                 , ( show addrPoolMax
                   , addrPoolMax
-                  , [ expectResponseCode @IO HTTP.status201 ]
+                  , [ expectResponseCode @IO HTTP.status201
+                    , expectField (#addressPoolGap . #getApiT) (`shouldBe` maxBound)
+                  ]
```
KtorZ, line 468: :+1:

piotr-iohk:
BTW, from what I see now testing manually on `master`, the issue is only about `"address_pool_gap": 232,` field returned by the wallet, however the list of addresses that is returned for a wallet created with address_pool_gap: 1000 is correct.

lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs
KtorZ:
```hs
{-# LANGUAGE GADTs #-}
{-# LANGUAGE KindSignatures #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE TypeFamilies #-}

import Data.Proxy
import Data.Typeable
import Data.Word
import GHC.Generics

newtype AddressPoolGap = AddressPoolGap
    { getAddressPoolGap :: Word32 }
    deriving (Generic, Show)

innerTypeOf :: (Generic a, GInnerType (Rep a)) => a -> Proxy (Inner (Rep a))
innerTypeOf = gInnerType . from

class GInnerType (f :: * -> *) where
    type Inner f :: *
    gInnerType :: f a -> Proxy (Inner f)

instance (GInnerType f) => GInnerType (M1 i c f) where
    type Inner (M1 i c f) = Inner f
    gInnerType (M1 fp) = gInnerType fp

instance GInnerType (K1 i c) where
    type Inner (K1 i c) = c
    gInnerType (K1 c) = Proxy @c


main :: IO ()
main = print $ typeRep $ innerTypeOf $ AddressPoolGap 42
```

```
$ stack runhaskell Main.hs
Word32
```

KtorZ (approved):

lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs
hasufell: Want it only on type level.

I think this will do:
```hs
type family Inner a where
  Inner (M1 _ _ x) = Inner x
  Inner (K1 _ x) = x
```

KtorZ:
Hydra failure: 

```
  test/unit/Cardano/Wallet/DB/Properties.hs:251:9:1) Cardano.Wallet.DB.Sqlite.Sqlite, sequential puts replace values in order, Tx Historyuncaught exception: PersistExceptionPersistMarshalError "Couldn't parse field `txMetaData` from table `tx_meta`. ConversionErrExpected \"[Aeson.String x]\" \"fromList [(\\\"hex\\\",Number 0.0)]\""(after 57 tests and 24 shrinks)KeyValPairs [(ad54f8e8...71a6e976,- ( b9c05762inputs 3rd 66393366outputs 42873 @ 250d4250...27143352withdrawals (1fae2df0, 89572)metadata:element 4318594985062:- key: "hex"- val: 0,+43.966001 in ledger since 784#201399946 ))]To rerun use: --match "/Cardano.Wallet.DB.Sqlite/Sqlite/sequential puts replace values in order/Tx History/"test/unit/Cardano/Wallet/DB/Properties.hs:273:9:2) Cardano.Wallet.DB.Sqlite.Sqlite, parallel puts replace values in _any_ order, Tx Historyuncaught exception: PersistExceptionPersistMarshalError "Couldn't parse field `txMetaData` from table `tx_meta`. ConversionErrExpected \"[Aeson.String x]\" \"fromList [(\\\"hex\\\",Number 0.0)]\""(after 57 tests and 24 shrinks)KeyValPairs [(PrimaryKey (WalletId {getWalletId = ad54f8e826744f5fa76a1730bdf9b65771a6e976}),GenTxHistory {unGenTxHistory = [(Tx {txId = Hash {getHash = "\185\192Wb\142\238\151%\225\DC1.\NULl\129\152\201\177\191\177\132c\195\243\NUL\143>\"J\\ \164\130"}, resolvedInputs = [(TxIn {inputId = Hash {getHash = "f93f7697c2ec7c5359429610f1e716f9"}, inputIx = 2},Coin {getCoin = 14709})], outputs = [TxOut {address = Address {unAddress = "%\rBP\133\NAKt\SYNrm_z\167\174\253]M}#XU=\ETBJ?1Ee+'\DC43R"}, coin = Coin {getCoin = 42873}}], withdrawals = fromList [(ChimericAccount {unChimericAccount = "\US\174-\240O\227]\213.F\224u\251+\215\137NWjR\148}.\174\ACK\"\224i"},Coin {getCoin = 89572})], metadata = Just (TxMetadata (MetaData' {mdMap = fromList [(4318594985062,Map [(S "hex",I 0)])], mdBytes = "\161\ESC\NUL\NUL\ETX\237\128Q\236f\161chex\NUL"}))},TxMeta {status = InLedger, direction = Incoming, slotNo = SlotNo 784, blockHeight = Quantity {getQuantity = 201399946}, amount = Quantity {getQuantity = 43966001}})]})]
```

Seems unrelated to this PR, and hopefully fixed once cardano-node#1797 gets merged.

rvl (approved):
Looks like a fix :+1:.

What about just replacing `WordN` with `Word` in the cases where the range is smallish and doesn't really matter, or where the bounds don't correspond to machine words.

In most cases 4 wasted bytes is nothing, but using specific sized number types distracts readers.

hasufell:
bors try

piotr-iohk:
>BTW, from what I see now testing manually on master, the issue is only about "address_pool_gap": 232, field returned by the wallet, however the number of addresses that is returned for a wallet created with address_pool_gap: 1000 is correct.

Having said that, perhaps it could be useful to add additional assertion to this test (WALLETS_CREATE_08 - address_pool_gap) checking also that list addresses endpoint shows `addrPoolMin` and `addrPoolMax` number of addresses in respective cases. (As a precaution for the future).

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
```diff
@@ -472,6 +472,14 @@ spec = do
             let payload = payloadWith' "Secure Wallet" mnemonics24 (fromIntegral addrPoolGap)
             r <- request @ApiWallet ctx (Link.postWallet @'Shelley) Default payload
             verify r expectations
+            rl <- request @[ApiWallet] ctx (Link.listWallets @'Shelley) Default Empty
+            verify rl
+                [ expectResponseCode @IO HTTP.status200
+                , expectListSize 1
+                , expectListField 0 (#addressPoolGap . #getApiT . #getAddressPoolGap)
+                    (`shouldBe` (fromIntegral addrPoolGap))
+                ]
```
KtorZ: I don't think that's what Piotr meant though. He meant to check that listing addresses returns the correct number of addresses.

piotr-iohk: Yes, sorry for not being precise. I meant this guy -> https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/listAddresses. The relationship here is that when we create wallet with address_pool_gap = N, the list address endpoint should return precisely N addresses for that wallet.

hasufell: I think should be fine now

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
```diff
@@ -470,8 +470,11 @@ spec = do
                 ]
         forM_ matrix $ \(title, addrPoolGap, expectations) -> it title $ \ctx -> do
             let payload = payloadWith' "Secure Wallet" mnemonics24 (fromIntegral addrPoolGap)
-            r <- request @ApiWallet ctx (Link.postWallet @'Shelley) Default payload
+            r@(_, Right wallet) <- request @ApiWallet ctx (Link.postWallet @'Shelley) Default payload
             verify r expectations
+            addrs <- listAddresses @n ctx wallet
+            length addrs `shouldBe` addrPoolGap
```
KtorZ:
```suggestion
            rW <- request @ApiWallet ctx (Link.postWallet @'Shelley) Default payload 
            verify rW expectations 
            rA <- listAddresses @n ctx (getFromResponse id r)
            verify rA 
                [ expectListSize addrPoolGap
                ]
```

KtorZ (approved):

nikolaglumac:
@KtorZ any ETA on when this one might be merged? 🙏

rvl:
Test failure fixed - passes locally - merging now.

bors r+

nikolaglumac:
Still no luck? :(

hasufell:
I can't reproduce this failure, even with the given seed. Are the integration tests stateful? @piotr-iohk

piotr-iohk:
No, they aren't. Not sure what that could be:

```
user error (expected a successful response but got an error: HttpException IncompleteHeaders)
```

perhaps http client error of some sort? Never seen it before...

KtorZ:
Seems like listing ALL 100K addresses doesn't work well in the integration tests.

KtorZ:
The call takes ~37s on my machine locally and the test pass just fine. The HTTP client manager has an increased request timeout of 180s for the integration tests. Buildkite may be a bit slower but I'd expect the call to resolve within a minute so 180s seems largely enough.

The request does not resolve at all on buildkite. The wallet creation already takes 112s:

```
    [34m[cardano-wallet.api-server:Info:1585][0m [2020-09-10 11:26:33.99 UTC] [RequestId 328] 201 Created in 112.746389206s
```

versus a bit less than 60s on my local machine. So perhaps it is simply timing out.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
```diff
@@ -462,18 +464,23 @@ spec = do
...
-            r <- request @ApiWallet ctx (Link.postWallet @'Shelley) Default payload
-            verify r expectations
+            let payload = payloadWith' "Secure Wallet" mnemonics24 (fromIntegral addrPoolGap)
+            rW <- request @ApiWallet ctx (Link.postWallet @'Shelley) Default payload
+            verify rW expectations
+            let w = getFromResponse id rW
+            rA <- request @[ApiAddress n] ctx
+                (Link.listAddresses @'Shelley w) Default Empty
+            _ <- request @ApiWallet ctx
+                (Link.deleteWallet @'Shelley w) Default Empty
```
piotr-iohk: Test passes for me locally as well... Could this line be somehow problematic? I guess we don't need to remove the wallet (Just thinking out loud :man_shrugging: )

KtorZ: Rodney added this, because of the hard-coded mnemonic and the automatically retried `it`. If you don't remove it, you get the second run to fail because the wallet already exists :/ ... although it should've been cleaned up in between...

piotr-iohk: oh, ok, haven't noticed that it uses hard-coded mnemonic...

Issue #2119, opened 2020-09-08
Less verbose integration test output by redirecting wallet+node logs to file


Anviking:
# Context

An overload (~60MB) of log output make CI integration test failures hard to debug.

Right now, the following are printed to stdout:
1. Wallet logs
2. The logs of 4 nodes (node/relay, BFT, pool-0, pool-1, pool-2)
3. hspec output 
4. Other integration test setup logs, like faucet tx submission

# Decision

Redirect the node and wallet logs to a file.

The hspec output should be uninterrupted like this:
<img width="693" alt="Skärmavbild 2020-09-03 kl  16 42 27" src="https://user-images.githubusercontent.com/304423/92458896-4748e680-f1c6-11ea-8c09-d933dd35279c.png">

# Acceptance Criteria

- [ ] Wallet and node logs should be redirected to a file
- [ ] The full wallet and node logs must still be accessible from CI failures
- [ ] The stdout of the integration tests should mostly contain an uninterrupted Hspec tree

---

# Development 

#2123

# QA 

1. Logs from wallet and nodes can be logged to file by setting`TESTS_LOGDIR=/pathtodir` (logs technically also exist in the temporary directory)
2. Buildkite uploads these as [artifacts](https://buildkite.com/input-output-hk/cardano-wallet/builds/10777#6cfe9d4f-acba-482f-a1d2-6ab614b13690). The other CIs don't. We could fix if we find it needed.
3. `stack test cardano-wallet:integration` has much cleaner standard output.

piotr-iohk:
lgtm :tada:

PR #2123, opened 2020-09-08
Less verbose integration test output

Anviking:
# Issue Number

#2119 

# Overview

- [x] More quiet stdout per default when running integration tests
- [x] Log full wallet, node and test-setup logs to `$TESTS_LOGDIR`, if set.
- [x] Upload `$TESTS_LOGDIR/*.log` as Buildkite artefacts

# Comments

- Regardless of `$TESTS_LOGDIR` setting, logs are still written to the temporary cluster directory as before.
- We can make hydra and GitHub Actions upload logs in a future PR. We might also be fine without it.

Skärmavbild 2020-09-10 kl  17 57 58

### DIFF
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
@@ -1095,7 +1095,7 @@ sendFaucetFundsTo
     -> [(String, Coin)]
     -> IO ()
 sendFaucetFundsTo tr dir allTargets = do
-    forM_ (group 20 allTargets) sendBatch
+    forM_ (group 80 allTargets) sendBatch
   where
     sendBatch targets = do
         (faucetInput, faucetPrv) <- takeFaucet
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE NumericUnderscores #-}
+{-# LANGUAGE QuasiQuotes #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TemplateHaskell #-}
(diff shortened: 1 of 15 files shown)


Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

(this doesn't make sense 🤕 )

Anviking:
bors try

Anviking:
bors try

(It won't run the full integration tests now, so should be quick)

Anviking:
It seems the integration test themselves don't see the `TESTS_LOGDIR` env var as set, despite it being set in `rebuild.hs`

Anviking:
bors try

Anviking:
If I have a small script
```haskell
cat debug.sh
#! /bin/bash
echo $TESTS_LOGDIR
```
and modify `rebuild.hs` to first read the `TESTS_LOGDIR` env var itself, and then run the `debug.sh` script,

and I run
```
export TESTS_LOGDIR=foo
nix-build .buildkite/default.nix -o sr
./sr/bin/rebuild
```
I see
```
rebuild.hs: TESTS_LOGDIR=Just "foo"
foo
ExitSuccess
```
but for some reason, when `stack test` is run instead of the script, the `TESTS_LOGDIR` variable isn't passed a long. I haven't managed to reproduce the latter locally yet though.

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

rvl:
I fixed the error on Hydra, and cleaned up stack.yaml.

bors try

Anviking:
bors try

`nix pure: false` didn't work for me previously, checking with added debug logs

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -777,19 +796,37 @@ genConfig
...
+                , scMinSev = sev
+                , scMaxSev = Critical
+                , scPrivacy = ScPublic
+                , scRotation = Nothing
+                }
+
+    let scribes = catMaybes
+            [ Just $ fileScribe ("cardano-node.log", severity)
+            , fileScribe . first T.pack <$> mExtraLogFile
+            ]
```
KtorZ, line 821: What's the use for two log files here?

lib/shelley/test/integration/Main.hs
```diff
@@ -242,26 +246,29 @@ specWithServer (tr, tracers) = aroundAll withContext . after tearDown
...
                 dir
+                extraLogDir
                 onByron
                 (afterFork dir)
                 (onClusterStart action dir dbDecorator)
 
     tr' = contramap MsgCluster tr
     onByron _ = pure ()
     afterFork dir _ = do
+        traceWith tr MsgSettingUpFacuet
```
KtorZ: *Facuet

lib/shelley/test/integration/Main.hs
```diff
@@ -348,13 +370,19 @@ bracketTracer' tr name = bracketTracer (contramap (MsgBracket name) tr)
...
-    minSeverityFromEnv Error "CARDANO_NODE_TRACING_MIN_SEVERITY"
+    minSeverityFromEnv Info "CARDANO_NODE_TRACING_MIN_SEVERITY"
 
--- Allow configuring integration tests and wallet log level with
+-- Allow configuring wallet log level with
 -- @CARDANO_WALLET_TRACING_MIN_SEVERITY@ environment variable.
 walletMinSeverityFromEnv :: IO Severity
 walletMinSeverityFromEnv =
-    minSeverityFromEnv Info "CARDANO_WALLET_TRACING_MIN_SEVERITY"
+    minSeverityFromEnv Critical "CARDANO_WALLET_TRACING_MIN_SEVERITY"
```
KtorZ, line 379: Is it really sound? Sure we will have less logs but we still needs those logs.

KtorZ:
That doesn't quite work locally for me. 

```
Progress 7/8: cardano-wallet--- TESTS_LOGDIR env:
integration: user error (Pattern match failure in do expression at test/integration/Main.hs:146:5-19)
```

KtorZ:
bors merge

KtorZ (approved):

KtorZ:
bors merge

KtorZ:
bors merge

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
Anviking: It seemed drastic/weird to remove the existing node logs from the temporary directory.

Without setting additional environment variables, we can inspect the dbs, and the logs, in that directory, while the tests are running.

Thought it is also a bit weird to duplicate the logs 🤷‍♂️

lib/shelley/test/integration/Main.hs
Anviking: This is the severity used for `Stdout`.

PR #2131, opened 2020-09-10
Temporarily use metadata revision branch of cardano-api

KtorZ:
# Issue Number

# Overview

- 3698a42a519b55b78cb2a619877ea8e9f5dca2f5
  :round_pushpin: **use temporary patched cardnao-api-1.19.1 with metadata support**
  
- 29050f78d5d0adbc8d12eea4fd3888ecffad29a5
  :round_pushpin: **use released version (2.0.0) for cardano-addresses**
  
- 5dcbee6a9b219245f0357e28df790b03ea23f90e
  :round_pushpin: **revise metadata generator to be generated from valid JSON values**

# Comments

We agreed with @AlexIOHK to move on with that version (so called "B") until the final version (describing metadata as JSON, so called "C") is finalized.


AlexIOHK (approved):
Agreed with using option "B" until further improvements are made.

rvl (approved):
But this scheme makes testing difficult for us. For example, if a metadata value goes through the database, that will be a json roundtrip, and the property tests will find the counterexamples which don't roundtrip properly.

PR #2132, opened 2020-09-11
Make pool DB model interface consistent.

jonathanknowles:
# Issue Number

https://jira.iohk.io/browse/ADP-406

# Overview

This PR adjusts `Pool.DB.Model` to make it more usable with `quickcheck-state-machine`.

**Changes:**

- All operations in the `Model` layer now return values of type `ModelOp a`, rather than the ad-hoc mix that was there before.
    
    _Example:_
    ```patch
    - mListRegisteredPools :: PoolDatabase -> ([PoolId], PoolDatabase)
    + mListRegisteredPools :: ModelOp [PoolId]
   ```

- The `Model` layer now provides **_all_** operations required by `DBLayer`. Previously, some of these were **_missing_**, and were provided only by the `MVar` layer.
    
    _Examples:_
    ```patch
    + mReadPoolLifeCycleStatus :: PoolId  -> ModelOp  PoolLifeCycleStatus
    + mListPoolLifeCycleData   :: EpochNo -> ModelOp [PoolLifeCycleStatus]
    + mRemoveRetiredPools      :: EpochNo -> ModelOp [PoolRetirementCertificate]
    ```

- All _**read-only**_ operations in the `MVar` layer now use `readPoolDB` rather than `adjustPoolDB` or `modifyMVar`.
    
    _Example:_
    ```patch
    listRegisteredPools =
    -   modifyMVar db (pure . swap . mListRegisteredPools)
    +   readPoolDB db mListRegisteredPools
    ```

- Some additional warts are fixed, and record update boilerplate reduced.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -51,6 +51,7 @@
 - ignore: {name: "Redundant do"} # Just an annoying hlint built-in, GHC may remove redundant do if he wants
 - ignore: {name: "Monoid law, left identity"} # Using 'mempty' can be useful to vertically-align elements
 - ignore: {name: "Use ?~"} # It's actually much clearer to do (.~ Just ...) than having to load yet-another-lens-operator
+- ignore: {name: "Redundant multi-way if"} # Using this style might be appropriate given the circumstances.
 
 # Add custom hints for this project
 #
--- a/lib/core/src/Cardano/Pool/DB/MVar.hs
+++ b/lib/core/src/Cardano/Pool/DB/MVar.hs
@@ -122,8 +122,8 @@ newDBLayer timeInterpreter = do
         listRegisteredPools =
             readPoolDB db mListRegisteredPools
 
-        listRetiredPools epochNo =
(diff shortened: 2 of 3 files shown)


rvl (approved):

KtorZ:
We need to prioritize #2131 over this one for the release.

KtorZ:
bors retry

rvl:
I would be cautious about simplifying `ModelOp` before there are state-machine tests. There might be a reason that it's structured like that.

jonathanknowles:
> I would be cautious about simplifying ModelOp before there are state-machine tests. There might be a reason that it's structured like that.

This is a good point.

However, I think we should be okay. We might need to provide something like the following transformation:

```hs
runModelOp :: PoolDatabase -> ModelOp a -> (Either PoolErr a, PoolDatabase)   
runModelOp db op = case runStateT op db of
    Left e -> (Left e, db)
    Right (result, dbUpdated) -> (Right result, dbUpdated)
```

PR #2136, opened 2020-09-14
Run existing pool DB property tests on model implementation.

jonathanknowles:
# Issue Number

https://jira.iohk.io/browse/ADP-406

# Overview

Over time, we've built up a fairly large suite of [property tests for the pool database](https://github.com/input-output-hk/cardano-wallet/blob/ff181966317b50083a61aa3e4f3dfd2ac1bff1a3/lib/core/test/unit/Cardano/Pool/DB/Properties.hs).

However, up until now these tests were only ever run on the **SQLite** implementation. The **model** implementation was not tested.

This PR fixes this by adjusting the test suite to test both the **SQLite** and **model** implementations. For example:

```hs
ghcid \
    --command "stack ghci cardano-wallet-core:lib cardano-wallet-core:test:unit" \
    --test ':run Spec.main -m prop_listPoolLifeCycleData_multiplePools_multipleCerts'

Cardano.Pool.DB.MVar
  MVar
    Stake Pool Properties
      prop_listPoolLifeCycleData_multiplePools_multipleCerts
        +++ OK, passed 100 tests.
Cardano.Pool.DB.Sqlite
  Sqlite
    Stake Pool Properties
      prop_listPoolLifeCycleData_multiplePools_multipleCerts
        +++ OK, passed 100 tests.

Finished in 40.3082 seconds
2 examples, 0 failures
```

## Details

This PR:
1. Generalizes the `Pool.DB.Properties` module to work with any implementation of `Pool.DBLayer`.
2. Adds a `Pool.DB.MVarSpec` module to handle the required setup for testing the model implementation.
3. Fixes the small number of resulting failures.

### DIFF
--- a/lib/core/test/unit/Cardano/Pool/DB/Properties.hs
+++ b/lib/core/test/unit/Cardano/Pool/DB/Properties.hs
@@ -5,21 +5,15 @@
 {-# LANGUAGE OverloadedLabels #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE RecordWildCards #-}
-{-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeFamilies #-}
 
 module Cardano.Pool.DB.Properties
     ( properties
     , withDB
-    , newMemoryDBLayer
     ) where
 
 import Prelude
 
-import Cardano.BM.Trace
-    ( traceInTVarIO )
-import Cardano.DB.Sqlite
-    ( SqliteContext )
 import Cardano.Pool.DB
     ( DBLayer (..)
     , ErrPointAlreadyExists (..)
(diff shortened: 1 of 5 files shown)


lib/core/test/unit/Cardano/Pool/DB/Properties.hs
```diff
@@ -935,7 +911,7 @@ prop_listRegisteredPools DBLayer {..} entries =
         monitor $ counterexample $ unlines
             [ "Read from DB: " <> show pools
             ]
-        assert (pools == (view #poolId <$> reverse entries))
+        assert (Set.fromList pools == Set.fromList (view #poolId <$> entries))
```
rvl: Any reason to remove the check on ordering?

rvl (approved):
It can't hurt to test the tests more.

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
jonathanknowles: > Any reason to remove the check on ordering?

Good question.

The function `listRegisteredPools` doesn't guarantee (currently) to list pools in any particular order, so it doesn't make sense for the property to expect a particular order.

(And as it turns out, the ordering of pools returned by the SQLite implementation of `listRegisteredPools` really is different from the ordering we get back from `Set.toList`. It seems unnecessary to make these orders the same by adding an extra ascending sort to the SQLite implementation.)

jonathanknowles:
bors cancel

Issue #2098, opened 2020-09-01
Ajust jsonToMetadata and jsonFromMetadata from cardano-api


KtorZ:
# Context

These functions are making several implicit conversion when dealing with JSON data which make them very unpractical to use. 

# Decision

a. Encode JSON as CBOR considering only the following "optimizations"
b. Hexadecimal sequences starting with 0x are encoded as CBOR bytestring.
c. JSON keys that are numbers are encoded as CBOR numbers.
d. CBOR is shown as JSON when possible, or as a string representing JSON-encoded data when not (i.e. when map keys aren't numbers or strings).
e. CBOR bytestrings are represented as hexadecimal strings prefixed with 0x

NOTE 1: Roundtripping would only work in one direction, when coming from a JSON value and that's totally fine.
NOTE 2: Complexe metadata using non-string keys in maps will render with funky keys in JSON, that's acceptable.
NOTE 3: Related Slack thread https://input-output-rnd.slack.com/archives/GR599HMFX/p1598966521093500

# Acceptance Criteria

---

# Development 

- See: https://github.com/input-output-hk/cardano-node/pull/1797
- Integrated in: #2131 

# QA 

- Metadata have been added to database property and state machine tests. Metadata are generated from JSON objects, as they would be in the API which guarantees that they roundtrip to JSON and from JSON back. Plus, any arbitrary metadata is guaranteed to be convertible to JSON by the type system. See generators here:

  - https://github.com/input-output-hk/cardano-wallet/blob/9721d5ed729212a4df69d484dfe8f12383800913/lib/core/test/unit/Cardano/Wallet/Gen.hs#L169-L241

- We use a patched version of cardano-node (patched on top of 1.19.1), which adds the fixed json conversions function. These functions are tested in cardano-api directly using a roundtrip property, and some golden tests to make sure we are in the end generating intended CBOR binary data:

  - https://github.com/input-output-hk/cardano-node/blob/eac29083ef25dc619f15854623275fb37e63d2b8/cardano-api/test/Test/Cardano/Api/MetaData.hs#L30-L55

rvl:
I have a slight preference for a JSON representation which can perfectly map one-to-one with the on-chain metadata.

It's not much more burden on API users. There are already several limitations on what metadata they can set. It's definitely not POJO.

piotr-iohk:
lgtm

PR #2142, opened 2020-09-16
cardano-node: 1.19.1-patched -> 1.20.0

rvl:
### Issue Number

ADP-307 / #2098

### Overview

Updates cardano-node to [1.20.0](https://github.com/input-output-hk/cardano-node/releases/tag/1.20.0)

There are quite a few upstream API changes.

Snapshot PR is input-output-hk/cardano-haskell#28.


rvl:
bors try

Anviking (approved):
lgtm.

Seems like swagger<->json tests are failing

specifications/api/swagger.yaml
```diff
@@ -1398,18 +1398,59 @@ components:
         passphrase: *lenientPassphrase
         address_index: *addressIndex
 
-    TransactionMetadatum: &TransactionMetadatum
+    TransactionMetadataValue: &TransactionMetadataValue
```
KtorZ, line 1401: We need to confirm this with @AlexIOHK . In theory, 1.20.0 supports both approaches for metadata, the one we currently use with direct JSON conversion, and this one that is more of a JSON description of the metadata. 

I'd have preferred going for the JSON description initially, but we had to get partners started with something and went for the direct JSON conversion which was ready. If we change that, it'll break existing integrations (although that is rather trivially fixed).

rvl: OK I have changed the API to use the "no schema" mapping from cardano-api. I think it's the same as before. The PR for changing to the detailed schema is #2147.

PR #2150, opened 2020-09-18
Fix a unit test failing on windows due to illegal filename

### LINKED ISSUE #2149: Jormungandr pool metadata unit tests sometimes fail on Windows
# Context

A test case occasionally fails on Windows because the property has chosen a filename which is illegal on Windows.

In this case, the filename was `prn`, but there are a few other filenames that you can't use: [MSDN: Naming Files, Paths, and Namespaces](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#file-and-directory-names).

# Test Case

Test case: `Cardano.Pool.Jormungandr.Metadata.Cardano.Pool.Registry.getStakePoolMetadata` Arbitrary zip files

Environment: windows build, tested under wine on hydra.

# Failure / Counter-example

https://hydra.iohk.io/build/4147712

```
Failures:

  test/unit/Cardano/Pool/Jormungandr/MetadataSpec.hs:163:9: 
  1) Cardano.Pool.Jormungandr.Metadata.Cardano.Pool.Registry.getStakePoolMetadata Arbitrary zip files
       uncaught exception: IOException of type NoSuchThing
       C:\users\nixbld\Temp\registry-spec-71354469d7a75af8\prn.zip: openBinaryFile: does not exist (No such file or directory)
       (after 21 tests and 2 shrinks)
         TestCase {stakePools = [], poolOwners = [], topDir = "prn"}

  To rerun use: --match "/Cardano.Pool.Jormungandr.Metadata/Cardano.Pool.Registry.getStakePoolMetadata/Arbitrary zip files/"
```

---

# Resolution 

The function `System.FilePath.Windows.makeValid` could be useful.

---

# QA 

1. There should no longer be failures due to this flaky test.

rvl:
### Issue Number

Resolves #2149

### Overview

Prevents jormungandr stake pool metadata tests from choosing illegal filenames on Windows.

### DIFF
--- a/lib/test-utils/src/Test/Utils/FilePath.hs
+++ b/lib/test-utils/src/Test/Utils/FilePath.hs
@@ -1,0 +1,33 @@
+-- |
+-- Copyright: © 2018-2020 IOHK
+-- License: Apache-2.0
+--
+-- File path related test utilities.
+
+module Test.Utils.FilePath
+    ( PathElement (..)
+    ) where
+
+import Prelude
+
+import System.FilePath.Windows
+    ( makeValid )
+import Test.QuickCheck
+    ( Arbitrary (..), elements, listOf1, scale )
+import Test.Utils.Windows
+    ( isWindows )
+
+-- | A file or directory name. The 'Arbitrary' instance will generate values
+-- which are valid on Windows and POSIX.
+--
+-- <https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file>
+newtype PathElement = PathElement FilePath deriving (Show, Eq)
+
+instance Arbitrary PathElement where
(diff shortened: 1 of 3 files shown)


lib/test-utils/src/Test/Utils/FilePath.hs
```diff
@@ -0,0 +1,33 @@
...
+-- File path related test utilities.
+
+module Test.Utils.FilePath
+    ( PathElement (..)
+    ) where
+
+import Prelude
+
+import System.FilePath.Windows
+    ( makeValid )
```
KtorZ, line 14: I am both amazed and dismayed that such function exists :sweat_smile:

KtorZ (approved):

PR #2153, opened 2020-09-21
Test pool garbage collection.

jonathanknowles:
# Issue Number

#2019 

# Overview

This PR tests that the **chain-following code** is correctly invoking the pool garbage collector by calling the `removeRetiredPools` operation at appropriate points in time.

Since the effects of the garbage collector are **not** visible through the API, the integration test framework cannot use the API to discover whether garbage collection has taken place.

This PR takes a novel approach of adding a **decorator** to the pool DB layer that records garbage collection events and makes them visible to the integration test layer.

Since the `removeRetiredPools` operation is already covered by property tests, the correctness of this operation is outside the scope of this PR.

# Details

This PR:
- [x] Adjusts the chain-following code to garbage collect pools _even if there are no new certificates_. This makes it possible to see the effects of garbage collection in the context of an integration test, where all certificates are issued at the start of time.
- [x] Limits pool garbage collection to _exactly once per epoch_, and no more.
- [x] Introduces the concept of a _pool database decorator_, which makes it possible to add instrumentation or monitoring to a pool database layer object.
- [x] Uses a pool database decorator to monitor _garbage collection events_ from the integration test framework.
- [x] Adds an integration test to check that retired pools are garbage collected _on schedule_.

# Logging

Here's an extract from the integration test log, showing that garbage collection is run exactly once per epoch:
```hs
[cardano-wallet.integration:Notice:758] [2020-09-21 09:04:59.96 UTC] Intercepted pool garbage collection event for epoch 1. No pools were removed from the database.
[cardano-wallet.integration:Notice:758] [2020-09-21 09:05:36.02 UTC] Intercepted pool garbage collection event for epoch 2. No pools were removed from the database.
[cardano-wallet.integration:Notice:758] [2020-09-21 09:06:16.01 UTC] Intercepted pool garbage collection event for epoch 3. The following pools were removed from the database: PoolRetirementCertificate {poolId = PoolId {getPoolId = "\ESC=\193\156j\184\158\175\252\133\SOH\243u\187\ETX\193\ESC\248\237]\CAN76\177\216\EOT\DC3\214"}, retirementEpoch = EpochNo {unEpochNo = 3}}
[cardano-wallet.integration:Notice:758] [2020-09-21 09:06:56.03 UTC] Intercepted pool garbage collection event for epoch 4. No pools were removed from the database.
[cardano-wallet.integration:Notice:758] [2020-09-21 09:07:36.02 UTC] Intercepted pool garbage collection event for epoch 5. No pools were removed from the database.
[cardano-wallet.integration:Notice:758] [2020-09-21 09:08:16.03 UTC] Intercepted pool garbage collection event for epoch 6. No pools were removed from the database.
[cardano-wallet.integration:Notice:758] [2020-09-21 09:08:56.01 UTC] Intercepted pool garbage collection event for epoch 7. No pools were removed from the database.
[cardano-wallet.integration:Notice:758] [2020-09-21 09:09:36.02 UTC] Intercepted pool garbage collection event for epoch 8. No pools were removed from the database.

```

### DIFF
--- a/lib/core-integration/src/Test/Integration/Framework/Context.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/Context.hs
@@ -1,0 +1,65 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveGeneric #-}
+
+module Test.Integration.Framework.Context
+    ( Context (..)
+    , TxDescription (..)
+    ) where
+
+import Prelude
+
+import Cardano.CLI
+    ( Port (..) )
+import Cardano.Wallet.Primitive.Types
+    ( NetworkParameters )
+import Cardano.Wallet.Transaction
+    ( DelegationAction )
+import Data.Proxy
+    ( Proxy (..) )
+import Data.Text
+    ( Text )
+import GHC.Generics
+    ( Generic )
+import Network.HTTP.Client
+    ( Manager )
+import Numeric.Natural
+    ( Natural )
+import Test.Integration.Faucet
+    ( Faucet )
+
+-- | Context for integration tests.
+--
+data Context t = Context
+    { _cleanup
+        :: IO ()
+        -- ^ A cleanup action.
+    , _manager
+        :: (Text, Manager)
+        -- ^ The underlying base URL and manager used by the wallet client.
+    , _walletPort
+        :: Port "wallet"
+        -- ^ Server TCP port.
+    , _faucet
+        :: Faucet
+        -- ^ Provides access to funded wallets.
+    , _feeEstimator :: TxDescription -> (Natural, Natural)
+        -- ^ A fee estimator.
+    , _networkParameters :: NetworkParameters
+        -- ^ Blockchain parameters for the underlying chain.
+    , _target
+        :: Proxy t
+    }
+    deriving Generic
+
+-- | Describe a transaction in terms of its inputs and outputs.
+data TxDescription
+    = DelegDescription DelegationAction
+    | PaymentDescription
+        { nInputs
+            :: Int
+        , nOutputs
+            :: Int
+        , nChanges
+            :: Int
+        }
+    deriving Show
--- a/lib/core-integration/src/Test/Integration/Framework/Context.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/Context.hs
@@ -3,6 +3,7 @@
 
 module Test.Integration.Framework.Context
     ( Context (..)
+    , PoolGarbageCollectionEvent (..)
     , TxDescription (..)
     ) where
 
@@ -11,9 +12,11 @@ import Prelude
 import Cardano.CLI
     ( Port (..) )
 import Cardano.Wallet.Primitive.Types
-    ( NetworkParameters )
+    ( EpochNo, NetworkParameters, PoolRetirementCertificate )
 import Cardano.Wallet.Transaction
     ( DelegationAction )
+import Data.IORef
+    ( IORef )
 import Data.Proxy
     ( Proxy (..) )
 import Data.Text
@@ -48,8 +51,25 @@ data Context t = Context
         -- ^ Blockchain parameters for the underlying chain.
     , _target
         :: Proxy t
+    , _poolGarbageCollectionEvents
+        :: IORef [PoolGarbageCollectionEvent]
+        -- ^ The complete list of pool garbage collection events.
+        -- Most recent events are stored at the head of the list.
     }
     deriving Generic
+
+-- | Records the parameters and return value of a single call to the
+--   'removeRetiredPools' operation of 'Pool.DB.DBLayer'.
+--
+data PoolGarbageCollectionEvent = PoolGarbageCollectionEvent
+    { poolGarbageCollectionEpochNo
+        :: EpochNo
+        -- ^ The epoch number parameter.
+    , poolGarbageCollectionCertificates
+        :: [PoolRetirementCertificate]
+        -- ^ The pools that were removed from the database.
+    }
+    deriving (Eq, Show)
 
 -- | Describe a transaction in terms of its inputs and outputs.
 data TxDescription
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
@@ -43,10 +43,14 @@ import Cardano.Wallet.Primitive.Types
     )
 import Cardano.Wallet.Unsafe
     ( unsafeFromHex, unsafeMkPercentage )
+import Control.Monad
+    ( forM_ )
 import Data.Function
     ( (&) )
 import Data.Generics.Internal.VL.Lens
     ( view, (^.) )
+import Data.IORef
+    ( readIORef )
 import Data.List
     ( sortOn )
 import Data.Maybe
@@ -67,9 +71,10 @@ import Test.Hspec
     ( SpecWith, describe, pendingWith, shouldBe, shouldSatisfy )
 import Test.Hspec.Extra
     ( it )
+import Test.Integration.Framework.Context
+    ( Context (..), PoolGarbageCollectionEvent (..) )
 import Test.Integration.Framework.DSL
-    ( Context (..)
-    , Headers (..)
+    ( Headers (..)
     , Payload (..)
     , delegating
     , delegationFee
@@ -813,6 +818,56 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
             , expectListField 2
                 (#metrics . #nonMyopicMemberRewards) (`shouldBe` Quantity 0)
             ]
+
+    it "STAKE_POOLS_GARBAGE_COLLECTION_01 - \
+        \retired pools are garbage collected on schedule and not before" $
+        \ctx -> do
+            -- First wait until a few garbage collection cycles have completed.
+            --
+            -- If this test case is run in isolation, this initial waiting
+            -- stage will require a few minutes to complete. However, this
+            -- initial stage is important as it allows the test case to be
+            -- run in isolation, if necessary.
+            --
+            -- If this test case is run at the end of a long integration test
+            -- run, it should run very quickly without much additional delay.
+            --
+            forM_ [1 .. 8] $ \epochNo -> do
+                let stateDescription = mconcat
+                        [ "Garbage has been collected for epoch "
+                        , show epochNo
+                        , "."
+                        ]
+                -- It's important that we wait incrementally, as an individual
+                -- call to 'eventually' must complete within a fairly short
+                -- period of time.
+                eventually stateDescription $ do
+                    events <- readIORef (view #_poolGarbageCollectionEvents ctx)
+                    length events `shouldSatisfy` (>= epochNo)
+
+            -- Check that exactly one pool was garbage collected, and no more:
+            events <- readIORef (view #_poolGarbageCollectionEvents ctx)
+            let certificates = poolGarbageCollectionCertificates =<< events
+            certificates `shouldSatisfy` ((== 1) . length)
+            let [certificate] = certificates
+            let [event] = events &
+                    filter (not . null . poolGarbageCollectionCertificates)
+
+            -- Check that the deleted pool was deleted at the correct epoch:
+            let expectedRetirementEpoch = 3
+            view #retirementEpoch certificate
+                `shouldBe` expectedRetirementEpoch
+            poolGarbageCollectionEpochNo event
+                `shouldBe` expectedRetirementEpoch
+
+            -- Check that the deleted pool was one of the test pools:
+            view #poolId certificate
+                `shouldSatisfy` (`Set.member` testClusterPoolIds)
+
+            -- Check that garbage collection occurred exactly once per epoch:
+            let epochs = poolGarbageCollectionEpochNo <$> events
+            (reverse epochs `zip` [1 ..]) `shouldSatisfy` all (uncurry (==))
+
   where
     arbitraryStake :: Maybe Coin
     arbitraryStake = Just $ ada 10_000_000_000
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
@@ -822,17 +822,38 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
     it "STAKE_POOLS_GARBAGE_COLLECTION_01 - \
         \retired pools are garbage collected on schedule and not before" $
         \ctx -> do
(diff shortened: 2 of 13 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -813,6 +818,56 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
...
+            --
+            -- If this test case is run in isolation, this initial waiting
+            -- stage will require a few minutes to complete. However, this
+            -- initial stage is important as it allows the test case to be
+            -- run in isolation, if necessary.
+            --
+            -- If this test case is run at the end of a long integration test
+            -- run, it should run very quickly without much additional delay.
+            --
+            forM_ [1 .. 8] $ \epochNo -> do
```
piotr-iohk: Why 8 and not 5? :sweat_smile:

jonathanknowles: > Why 8 and not 5? sweat_smile

In reality, it could be any number, so long as it's greater than 5 (as this is the epoch in which our test pool will be garbage collected). But having said that, it's useful to see that several subsequent epochs elapse **_without_** any garbage collection. I'll add a comment explaining this choice. :+1:

jonathanknowles: @piotr-iohk Fixed in 428f483208dff6c742565d6579adcfdc1577ca96.

piotr-iohk:
@jonathanknowles it seems that all tables are garbage collected on epoch 5 except `pool_metadata`:

Screenshot from 2020-09-21 11-53-20

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -813,6 +818,64 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
...
+    it "STAKE_POOLS_GARBAGE_COLLECTION_01 - \
+        \retired pools are garbage collected on schedule and not before" $
+        \ctx -> do
+            -- The epoch for which we expect to see a pool retirement event.
+            -- See 'testPoolConfigs' for the original source of this value.
+            let expectedRetirementEpoch = 3
+
+            -- The last epoch for which we expect to see a garbage collection
+            -- event. This is an arbitrary point in time that lies beyond the
+            -- expected retirement epoch.
```
piotr-iohk:
```suggestion
            -- The last epoch for which we expect to see effective garbage collection
            -- event is epoch 5, since the stake pool retires in epoch 3. 
            -- This is an arbitrary point in time that lies beyond the
            -- expected garbage collection epoch.
```
?

jonathanknowles:
> @jonathanknowles it seems that all tables are garbage collected on epoch 5 except `pool_metadata`:
> 
> Screenshot from 2020-09-21 11-53-20

Well spotted! This appears to be a problem with the `removePools` database operation.

The SQLite implementation of `removePools` is:

```hs
removePools = mapM_ $ \pool -> do
    liftIO $ traceWith trace $ MsgRemovingPool pool
    deleteWhere [ PoolProductionPoolId ==. pool ]
    deleteWhere [ PoolOwnerPoolId ==. pool ]
    deleteWhere [ PoolRegistrationPoolId ==. pool ]
    deleteWhere [ PoolRetirementPoolId ==. pool ]
    deleteWhere [ StakeDistributionPoolId ==. pool ]
```

This is **correct**, in that it removes all rows where `pool_id` is the given pool. However, it's also **incomplete**. The `pool_metadata` table is indexed by `metadata_hash` rather than by `pool_id`, so we'll have to add an extra lookup step to remove this data.

I've created a separate issue (https://github.com/input-output-hk/cardano-wallet/issues/2154) to capture the task of fixing the `removePools` DB operation, as the focus of this PR is to check that the chain-following code is invoking the garbage collector correctly, rather than the correctness of the DB layer.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
jonathanknowles: Fixed in 684d8a55ff4dc8b1df548edc2252155a94a59cfb.

piotr-iohk (approved):
lgtm

lib/core-integration/src/Test/Integration/Framework/Context.hs
```diff
@@ -0,0 +1,85 @@
...
+        :: Faucet
+        -- ^ Provides access to funded wallets.
+    , _feeEstimator :: TxDescription -> (Natural, Natural)
+        -- ^ A fee estimator.
+    , _networkParameters :: NetworkParameters
+        -- ^ Blockchain parameters for the underlying chain.
+    , _target
+        :: Proxy t
+    , _poolGarbageCollectionEvents
+        :: IORef [PoolGarbageCollectionEvent]
```
Anviking, line 55: It feels _a bit_ wrong to me to add such a such a specific thing to the `Context` that is global to all integration tests.

This lead me thinking to: what if we instead added something to the `Context` that allowed individual tests to subscribe whatever subset of the wallet logs they liked? [(Imagined example)](https://input-output-rnd.slack.com/archives/C4Q7MF25U/p1600699953028500) Then it could work for other things than pool garbage collection events, and wouldn't require adding decorators to things.

jonathanknowles: > This lead me thinking to: what if we instead added something to the `Context` that allowed individual tests to subscribe whatever subset of the wallet logs they liked? [(Imagined example)](https://input-output-rnd.slack.com/archives/C4Q7MF25U/p1600699953028500) Then it could work for other things than pool garbage collection events, and wouldn't require adding decorators to things.

@Anviking I do agree that it would be great to have a general mechanism to subscribe to a subset of the wallet logs, instead of this specific mechanism.

However, one issue to solve is that a test might need to access events that occurred **_before_** it started executing.

Specifically, in the case of the `STAKE_POOLS_GARBAGE_COLLECTION_01` test:
- the test needs to examine all historical garbage collection events to check that there was indeed one garbage collection per epoch.
- by the time the test starts, several epochs might already have elapsed.
- if we only start recording events at the start of the test, then we won't be able to access earlier garbage collection events.
- the pool DB decorator is able to solve this problem as it starts recording garbage collection events from the very beginning of time.

Regarding:
> It feels a bit wrong to me to add such a such a specific thing to the Context that is global to all integration tests.

I agree with you that this seems like a very specific addition. However, I think this is likely to be an exceptional case, as it's rare that we need to test something that's not visible through the API.

However, if we do end up wanting to make similar additions in future, I agree it would be worth spending the effort to make a more general mechanism.

I like the idea of using the logging mechanism for this, as the log messages are already typed, meaning that we can compare values directly, without having to parse values from text.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -153,16 +156,49 @@ defaultFilePath = (</> "stake-pools.sqlite")
...
        -- ^ Logging object
     -> Maybe FilePath
        -- ^ Database file location, or Nothing for in-memory database
     -> TimeInterpreter IO
+       -- ^ The time interpreter object.
     -> (DBLayer IO -> IO a)
        -- ^ Action to run.
     -> IO a
-withDBLayer trace fp timeInterpreter action = do
+withDecoratedDBLayer dbDecorator trace fp timeInterpreter action = do
```
KtorZ, line 199: We already have logs as _values_ in the database tracer, so I am not sure to understand the need for an additional decorator?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -462,19 +462,26 @@ monitorStakePools
...
+            let forwardHandler = forward latestGarbageCollectionEpochRef
+            follow nl followTrace cursor forwardHandler getHeader >>= \case
+                FollowInterrupted ->
+                    traceWith tr MsgHaltMonitoring
+                FollowFailure ->
+                    traceWith tr MsgCrashMonitoring
+                FollowRollback point -> do
+                    traceWith tr $ MsgRollingBackTo point
+                    liftIO . atomically $ rollbackTo point
+                    loop
```
KtorZ, line 483: What is the rationale for this change? Is this the fix?

KtorZ (commented):
This was maybe merged a bit too fast...

lib/core/src/Cardano/Pool/DB/Sqlite.hs
jonathanknowles: @KtorZ wrote:
> We already have logs as _values_ in the database tracer, so I am not sure to understand the need for an additional decorator?

What mechanism do you propose for the integration test to get the values it needs?

Specifically, in the case of the `STAKE_POOLS_GARBAGE_COLLECTION_01` test:

* the test needs to examine all historical garbage collection events to check that there was indeed one garbage collection per epoch.
* by the time the test starts, several epochs might already have elapsed.
* if we only start recording events at the start of the test, then we won't be able to access earlier garbage collection events.
* the pool DB decorator is able to solve this problem as it starts recording garbage collection events from the very beginning of time.
* the pool DB decorator doesn't record anything else apart from garbage collection events, so the memory footprint should be minimal.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
jonathanknowles: > What is the rationale for this change? Is this the fix?

This is explained in the message for [the commit](https://github.com/input-output-hk/cardano-wallet/pull/2153/commits/52273ea0f97977d96c762a9d6091fcf856794a68).

> Before this change, it was possible for garbage collection to happen more than once per epoch, as any rollback would recreate the mutable reference to the latest garbage collection epoch, causing the latest garbage collection epoch value to be reset to zero.
>
> This change adjusts the reference creation mechanism so that it only happens once, on startup.

The result is that garbage collection can now only happen once per epoch.

jonathanknowles:
@KtorZ wrote:
> This was maybe merged a bit too fast...

How so?

The PR was created yesterday morning, and approved yesterday afternoon. It was merged today, in accordance with our process.

There was a long discussion about potential solutions here: https://input-output-rnd.slack.com/archives/GBT05825V/p1600328272058800. At the time of the discussion, there were alternative solutions put forward, but each alternative had its own issues. In the end, we had to make a choice to proceed with one of the solutions. The benefits of this solution:

- no API alterations required.
- low memory footprint (as there are very few garbage collection events).
- all historical garbage collection events are available for inspection (and not just those that occur during the test).
- we were able to implement it within a reasonably short period of time, which is important, given our other priorities.

@Anviking put forward a suggestion for a more general solution involving log filtering and subscription [here](https://github.com/input-output-hk/cardano-wallet/pull/2153#discussion_r492180353) and [here](https://input-output-rnd.slack.com/archives/C4Q7MF25U/p1600699953028500). I agree with his point of view. However, it's not clear to me how we would implement such a system right now, at least without further engineering effort.

Though I'd be very happy to see a working proposal, as I'm quite interested in improving on this.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
KtorZ: We have used this approach with the logger for various other tests: 

- The logger can be instrumented to log message in a `TVar`
- That `TVar` can then be read at any time, and logs filtered from it to see for particular events. 

 > the pool DB decorator is able to solve this problem

Yes, but it adds yet-another-layer-of-complexity, for something for which we already had a working solution. It's not a huge addition and the impact on the normal production setup is quite small yet I am usually quite reluctant to add bits of code that are only used in testing.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
KtorZ: ~~How does this make it run once per epoch since the reference is created only once on startup?~~

Nevermind. That is because the `latestGarbageCollectionEpochRef` is an `IORef` that gets mutated in the `forwardHandler`.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
jonathanknowles: > We have used this approach with the logger for various other tests:
> The logger can be instrumented to log message in a TVar
> That TVar can then be read at any time, and logs filtered from it to see for particular events.

This sounds like a useful technique.

I can see there are some property tests that use this technique. (For example: `lib/shelley/test/unit/Cardano/Wallet/Shelley/NetworkSpec.hs`). Though I can't find any integration tests that use the same method.

For the purposes of `STAKE_POOLS_GARBAGE_COLLECTION_01`, it's important that we can inspect all the garbage collection events from the start of time, and not just those that occur during the duration of that one test case.

> Yes, but it adds yet-another-layer-of-complexity, for something for which we already had a working solution.

Sadly, no-one suggested this solution when we originally discussed this issue ([see original discussion](https://input-output-rnd.slack.com/archives/GBT05825V/p1600328272058800)), otherwise I would have been eager to try it out, especially if it would have meant that we didn't need to invent another abstraction.

Anviking: > We have used this approach with the logger for various other tests

I think:
- we _have_ used `traceInTVarIO` for testing small components in isolation.
- the problem here is to test a small component, _but_ in the context of the full integration tests. We can't choose the tracer for the small component specifically. We'd need to setup `traceInTVarIO` globally in the cluster setup.

Although, I guess (🤔) a global `traceInTVarIO` + filter, would work _as well as_ the DB decorator… making it much better since we then don't need the "yet-another-layer-of-complexity".

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
jonathanknowles: > Nevermind. That is because the `latestGarbageCollectionEpochRef` is an `IORef` that gets mutated in the `forwardHandler`.

That's correct.

Coincidentally, it was the integration test that helped us to discover this bug. There are several rollbacks that occur during the integration test, each of which triggered a new garbage collection (before this change was applied).

When testing on `mainnet`, we did not see this behaviour, IIRC.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
Anviking: > Sadly, no-one suggested this solution when we originally discussed this issue

To me, doing stuff with `traceInTVarIO` or to inventing a DB Decorator seem both like implementation details of your first proposed solution:

> 1.  Make garbage collection happen once per epoch without fail (even when there are no new certificates), and then have our integration tests scan the logs for garbage collection messages. (Though we'd really want to avoid accumulating logs in memory.)

PR #2156, opened 2020-09-21
Runtime configuration of pool metadata SMASH server via API

hasufell:
### Issue Number

ADP-427 / #2163

### Comments

TODO:
- [x] Create a GitHub milestone for ADP-427
- [x] Create a GitHub issue detailing implementation plan and assign to milestone and PR.
- [x] add Settings to DB
- [x] add POST endpoint to change Settings
- [x] adjust cli option
- [x] fix metadata fetching logic to start anew when Settings change

Tests:
- [x] POST endpoint
- [ ] integration test
- [x] modify and read settings
- [ ] `delPoolMetadata`


lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -1870,3 +1889,49 @@ invariant msg a predicate =
...
+data Settings = Settings {
+    poolMetadataSource :: PoolMetadataSource
+} deriving (Show, Generic, Eq)
+
+defaultSettings :: Settings
+defaultSettings = Settings {
+    poolMetadataSource = None
+}
+
+instance FromJSON PoolMetadataSource where
```
rvl, line 1939: What we usually do is define aeson instances on `ApiT PoolMetadataSource` and
```haskell
data ApiSettings = ApiSettings { poolMetadataSource :: !(ApiT PoolMetadataSource) }
```
That way affords different JSON encodings for primitive types, if needed.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -1870,3 +1889,49 @@ invariant msg a predicate =
...
+
+instance FromText PoolMetadataSource where
+    fromText "" = Right None
+    fromText "direct" = Right FetchDirect
+    fromText (T.unpack -> uri) =
+        case parseURI uri of
+            Nothing -> Left $ TextDecodingError ("Could not parse URI: " <> uri)
+            Just uri' -> Right (FetchSMASH uri')
+
+data Settings = Settings {
```
rvl, line 1930: HLint is gonna bother you to make this a newtype :shrug:.

lib/core/test/unit/Cardano/Pool/DB/Arbitrary.hs
```diff
@@ -371,3 +378,16 @@ instance Arbitrary StakePoolsFixture where
...
             pool <- elements pools
             return $ (pool,slot):pairs
+
+instance Arbitrary URI where
+    arbitrary = elements
+        [fromJust (parseURI "https://my.little.friend")
+        ,fromJust (parseURI "http://its-friday.com:8000")]
+
+instance Arbitrary PoolMetadataSource where
+    arbitrary = oneof [ pure None
```
rvl: No problem using `genericArbitrary` / `genericShrink` here.

specifications/api/swagger.yaml
```diff
@@ -267,6 +267,28 @@ x-walletName: &walletName
...
+  description: Pool metadata source. null -> no fetching, direct -> direct fetching, uri -> use SMASH server
+  oneOf:
+    - type: string
+      enum:
+        - null
+        - direct
+    - title: URI
+      type: string
+      format: uri
+      example: https://iohk.io
```
rvl: Perhaps we can find the actual CF smash URL and use that as an example.

rvl (commented):
Good start :+1:

lib/core/src/Cardano/Wallet/Primitive/Types.hs
hasufell: There is an instance for `ApiT Settings`:

```hs
instance FromJSON (ApiT W.Settings) where
    parseJSON = fmap ApiT . genericParseJSON defaultRecordTypeOptions
instance ToJSON (ApiT W.Settings) where
    toJSON = genericToJSON defaultRecordTypeOptions . getApiT
```

lib/core/src/Cardano/Wallet/Primitive/Types.hs
hasufell: Any reason we want different strictness properties? Also, we're likely to add more records, so this is morally a data type. Screw hlint :)

lib/core/test/unit/Cardano/Pool/DB/Arbitrary.hs
hasufell: I'm not sure if that will output valid URIs. And it requires a lot of orphan instances.

specifications/api/swagger.yaml
hasufell: I can't find any.

hasufell: Done

lib/core/src/Cardano/Wallet/Primitive/Types.hs
KtorZ: > Screw hlint :)

That's always an option. If you end up doing this, use a proper pragma to ignore the hlint suggestion with a comment explaining your rationale and why it is justified.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -1870,3 +1889,49 @@ invariant msg a predicate =
...
+    | FetchSMASH URI
+    deriving (Show, Generic, Eq)
+
+instance ToText PoolMetadataSource where
+    toText None = (T.pack "")
+    toText FetchDirect = (T.pack "direct")
+    toText (FetchSMASH uri) = T.pack $ uriToString id uri ""
+
+instance FromText PoolMetadataSource where
+    fromText "" = Right None
```
KtorZ: What about making this explicit as `"none"` ?

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -1870,3 +1889,49 @@ invariant msg a predicate =
 distance :: (Ord a, Num a) => a -> a -> a
 distance a b =
     if a < b then b - a else a - b
+
+data PoolMetadataSource
+    = None
```
KtorZ: Little OCD here, but `None` ~> `FetchNone` for consistency :pray:

lib/core/src/Cardano/Wallet/Primitive/Types.hs
KtorZ: That's rule #_154 indeed, no JSON instances in `Primitive.Types`. JSON instances for the API interfaces are located only in API.Types.

specifications/api/swagger.yaml
```diff
@@ -267,6 +267,28 @@ x-walletName: &walletName
   minLength: 1
   example: Alan's Wallet
 
+x-poolMetadataSource: &poolMetadataSource
+  description: Pool metadata source. null -> no fetching, direct -> direct fetching, uri -> use SMASH server
+  oneOf:
+    - type: string
+      enum:
+        - null
```
KtorZ: I don't think `null` is actually of type `string` :grimacing:

specifications/api/swagger.yaml
```diff
@@ -267,6 +267,28 @@ x-walletName: &walletName
...
+        - direct
+    - title: URI
+      type: string
+      format: uri
+      example: https://smash.cardano-mainnet.iohk.io/
```
KtorZ, line 296: :+1:

lib/core/src/Cardano/Wallet/Primitive/Types.hs
hasufell: Done

lib/core/src/Cardano/Wallet/Primitive/Types.hs
hasufell: Done, against my neurologists advice...

hasufell:
@rvl I believe the nix config needs some adjustment?

rvl:
Yes - normally Buildkite would push a commit with the updated files, but this PR comes from a forked repo, so it has no permission to do that.
You could run `./nix/regenerate.sh` locally (give it a little while to fetch repos).
I suggest opening future PRs as branches on the main repo, so that Buildkite is allowed to build them.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
hasufell: I don't see a justification for the rule in the first place. newtype is for:

1. you really want to make sure the data type is strict and zero-cost over the inner type
2. it is morally a newtype (e.g. we want to add instances over an existing type, etc)

Just because a product type currently has only one record doesn't mean you want to turn it into a newtype.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1303,19 +1308,17 @@ tlsOption = TlsConfiguration
...
 
 smashURLOption
-    :: Parser URI
+    :: Parser PoolMetadataSource
 smashURLOption = option (eitherReader reader) $ mempty
     <> long "smash-url"
     <> metavar "URL"
-    <> help "Optional HTTP(S) address of a \
-            \Stakepool Metadata Aggregation Server."
+    <> help "Source of SMASH (Stakepool Metadata Aggregation Server). Possible values: \
```
KtorZ: That is not quite right. The option does not capture the source of SMASH, it captures the strategy for fetching stake pool metadata. As a matter of fact, `none` and `direct` makes no use of SMASH. 

`pool-metadata-fetching-strategy` would better reflect what that option is, although it's arguably a long name so if you have a better suggestion :grimacing: 

(perhaps simply `--pool-metadata-fetching`)

lib/core/src/Cardano/Wallet/Primitive/Types.hs
KtorZ: I do fully agree with you @hasufell :) 
Perhaps a better way forward about that one would be to disable the rule globally in our hlint config.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -592,27 +615,51 @@ monitorStakePools tr gp nl DBLayer{..} =
...
-        threadDelay blockFrequency
+monitorMetadata tr gp DBLayer{..} = do
+    initSettings <- atomically readSettings
+    sref <- newIORef initSettings
+    forever $ do
+        prevSettings <- readIORef sref
+        newSettings <- atomically readSettings
+        writeIORef sref newSettings
+
+        -- delete old metadata table if need be
```
hasufell: Note that this is required to avoid race conditions. We can't drop the table in the PUT logic. Another side effect here is that changing the setting actually takes a bit of time (seconds), because we need to wait for the loop to continue before the new settings get picked up. I feel this is a little awkward @rvl @KtorZ

KtorZ: > that changing the setting actually takes a bit of time (seconds), because we need to wait for the loop to continue before the new settings get picked up

I don't quite see the problem with that? So long as: 

- An API call changing the settings is resolved promptly 
- The new settings is saved before the API call resolves

Then of course, the settings will only enter in effects on the next fetching loop. I don't think this is awkward? Why awkward?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -592,27 +615,51 @@ monitorStakePools tr gp nl DBLayer{..} =
...
+        case (poolMetadataSource prevSettings, poolMetadataSource newSettings) of
+            (_, FetchNone) -> liftIO $ atomically $ delPoolMetadata
```
KtorZ: Why the special case for `FetchNone` ? Isn't that covered already by the next branch?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -592,27 +615,51 @@ monitorStakePools tr gp nl DBLayer{..} =
...
+            (old, new)
+                | old /= new -> liftIO $ atomically $ delPoolMetadata
+            _ -> pure ()
+
+        -- fetch
+        (refs, successes) <- case poolMetadataSource newSettings of
+            FetchNone -> pure ([], [])
+            FetchDirect -> (fetcher Nothing) >>= fetchThem
+            FetchSMASH uri -> (fetcher $ Just uri) >>= fetchThem
```
KtorZ: It's sort of weird to use a `Maybe` here that gets remapped to a fetching strategy below. Wouldn't be more transparent to simply pass the url builder strategy to the `fetcher` function ? 

```hs
FetchDirect -> fetcher identityUrlBuilder >>= fetchThem
FetchSMASH uri -> fetcher registryUrlBuilder >>= fetchThem
```

?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
hasufell: `(FetchNone, FetchNone)` doesn't trigger cleanup. If this was an old database, both prevSettings and newSettings will be `FetchNone`, but the database will still be populated.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -272,14 +272,14 @@ import Web.HttpApiData
...
 import qualified Cardano.Crypto.Wallet as CC
+import qualified Cardano.Wallet.Primitive.Types as W
 import qualified Data.Aeson as Aeson
 import qualified Data.Aeson.Types as Aeson
 import qualified Data.ByteString.Lazy as BL
 import qualified Data.HashMap.Strict as HM
 import qualified Data.Map as Map
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
-import qualified Cardano.Wallet.Primitive.Types as W
```
KtorZ: That change look like a difference in configuration or version of `stylish-haskell` :thinking: ...

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1303,19 +1308,17 @@ tlsOption = TlsConfiguration
...
+    :: Parser PoolMetadataSource
 smashURLOption = option (eitherReader reader) $ mempty
-    <> long "smash-url"
-    <> metavar "URL"
-    <> help "Optional HTTP(S) address of a \
-            \Stakepool Metadata Aggregation Server."
+    <> long "pool-metadata-fetching"
+    <> metavar "STRATEGY"
+    <> help "Pool Metadata fetching strategy. This setting will persist across restarts. Possible values: \
+            \ none (default), \
```
KtorZ: NOTE: there's a `default` combinator for optparse-applicative and also an option to show defaults in the help text auutomatically based on that value.

lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
```diff
@@ -992,6 +993,17 @@ instance Malformed (BodyParam ApiPostRandomAddressData) where
...
         ]
 
+instance Malformed (BodyParam SettingsPutData) where
+    malformed = first (BodyParam . Aeson.encode) <$>
+        [ ( [aesonQQ|
+            { "settings": {
+                          "pool_metadata_source": "not_a_uri"
+                          }
+            }|]
+          , "Error in $['address_index']: parsing Int failed, expected Number, but encountered String"
```
KtorZ: That error message does not sound right **at all**.

lib/shelley/src/Cardano/Wallet/Shelley.hs
```diff
@@ -336,11 +296,11 @@ serveWallet
         -> IO ()
     startServer _proxy socket byron icarus shelley spl ntp = do
         sockAddr <- getSocketName socket
-        let settings = Warp.defaultSettings & setBeforeMainLoop
+        let settings' = Warp.defaultSettings & setBeforeMainLoop
```
KtorZ: perhaps `serverSettings`

lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
hasufell: Copy pasta lol

lib/shelley/src/Cardano/Wallet/Shelley.hs
```diff
@@ -354,21 +314,15 @@ serveWallet
...
-                $ \db -> do
-            let spl = newStakePoolLayer nl db
+                $ \db@DBLayer{..} -> do
+
             void $ forkFinally (monitorStakePools tr gp nl db) onExit
-            fetch <- fetchFromRemote trFetch fetchStrategies
-                <$> newManager defaultManagerSettings
-            void $ forkFinally (monitorMetadata tr gp fetch db) onExit
+            spl <- newStakePoolLayer nl db (forkFinally (monitorMetadata tr gp db) onExit)
+            putSettings spl (fromMaybe defaultSettings settings)
```
KtorZ: isn't there a race-condition here? `monitorMetadata` requires the settings to exists in the database, but on the first run, it is inserted _right after_. 

Also, on a different note, now that I see these `putSettings`, `defaultSettings`, `settings` a bit out of context, it feels like `settings` is a too generic term in itself and it gets quite confusing about _what settings_. Perhaps It'd be good to use more descriptive qualifiers --> `putMetadataFetchingSettings` etc .. ?

lib/shelley/src/Cardano/Wallet/Shelley/Logging.hs
```diff
@@ -0,0 +1,338 @@
...
+    | MsgHaltMonitoring
+    | MsgCrashMonitoring
+    | MsgRollingBackTo SlotNo
+    | MsgStakePoolGarbageCollection PoolGarbageCollectionInfo
+    | MsgStakePoolRegistration PoolRegistrationCertificate
+    | MsgStakePoolRetirement PoolRetirementCertificate
+    | MsgErrProduction ErrPointAlreadyExists
+    | MsgFetchPoolMetadata StakePoolMetadataFetchLog
+    | MsgFetchTakeBreak Int
+    deriving (Show, Eq)
```
KtorZ: Why is the StakePoolLog ending up here with the rest of the application logs? Seems very awkward to me.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -172,19 +179,35 @@ data StakePoolLayer = StakePoolLayer
...
         -> ExceptT ErrListPools IO [Api.ApiStakePool]
+
+    , putSettings :: Settings -> IO ()
+
+    , clearMetadataTable :: IO ()
+
+    , metadataThread :: TVar ThreadId
+
+    , metadataRun
+        :: IO ThreadId
```
KtorZ: What's the difference between `run` and `thread` ?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -196,6 +219,36 @@ newStakePoolLayer nl db@DBLayer {..} =
     _knownPools =
         Set.fromList <$> liftIO (atomically listRegisteredPools)
 
+    -- In order to apply the settings, we have to restart the
+    -- metadata sync thread as well to avoid race conditions.
+    _putSettings :: TVar ThreadId -> Settings -> IO ()
+    _putSettings tv settings = do
+        -- stop syncing thread, so we can apply the settings cleanly
+        mt <- readTVarIO tv
+        killThread mt
```
KtorZ: That thread has no particular exception handler and who knows what will happens if we kill it in the middle of an HTTP query or in the middle of a database call. May I suggest to not kill the thread and to revert the control here? What about:

- Passing an `MVar Settings` to the fetching thread. 
- Have the fetching thread taking that MVar before every fetch, and putting it back once fetched and done (before going to sleep, using a safe bracket-style like `withMVar`). 
- Have the `putSettings` function also take and replace that MVar when handling a query, replacing the content with the new settings, and dropping the table if necessary while holding on that MVar. 

This way, we get rid of potential errors that will occur from killing the thread at arbitrary time, and we still ensure complete synchronization of both thread?

lib/shelley/src/Cardano/Wallet/Shelley.hs
hasufell: > isn't there a race-condition here? monitorMetadata requires the settings to exists in the database, but on the first run, it is inserted right after.

Nope. The database returns `defaultSettings` if the table does not exist.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
hasufell: > That thread has no particular exception handler and who knows what will happens if we kill it in the middle of an HTTP query or in the middle of a database call. 

What exactly can go wrong here? You'd have to be concerned about every ctrl-c by a user as well then, because it causes an async exception too. I'd exepect database operations to be atomic. And if there is a problem with killing this thread it should be fixed in the thread rather than worked around with MVars imo.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
hasufell: thread is the thread id, run runs the thread :man_shrugging:

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
KtorZ: > You'd have to be concerned about every ctrl-c by a user as well the user

We are! Precisely! I think we should avoid killing thread were we don't need to. Async exceptions are hard to deal with as they are without us shooting ourselves in the foot. Killing the thread arbitrarily may result in random errors (which won't probably do anything more than printing an error log, but that's precisely what I want to avoid). We shouldn't raise errors when there are easy ways to avoid them.

hasufell:
I added integration tests for GET/PUT, but I'm not entirely sure how to approach testing the logic itself: that
1. when switching smash server, the pool database is eventually cleared (hard to catch)...
2. ...and then repopulated

Is that something we want in this PR? @piotr-iohk ideas?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
hasufell: Ok can. I think we need to be careful in the endpoint tests then, because of the potential delay.

lib/cli/src/Cardano/CLI.hs
rvl: It's called `value` and `showDefault`.

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1303,19 +1308,17 @@ tlsOption = TlsConfiguration
...
+            \ <SMASH-URL>"
   where
-    reader :: String -> Either String URI
-    reader = maybe (Left err) Right . (parseAbsoluteURI >=> isHttp)
-
-    isHttp uri
-        | uriScheme uri `elem` ["http:", "https:"] = Just uri
-        | otherwise = Nothing
+    reader :: String -> Either String PoolMetadataSource
+    reader = left (const err) . fromText @PoolMetadataSource . T.pack
```
rvl: You can use `fromTextS` instead.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -1870,3 +1888,60 @@ invariant msg a predicate =
 distance :: (Ord a, Num a) => a -> a -> a
 distance a b =
     if a < b then b - a else a - b
+
+-- | Source of Stake Pool Metadata aggregation.
+--
+-- Don't export 'MkFetchSMASH' contructors, so that @fromText@ is our smart constructor.
```
rvl: Another way you could do it, which might be shorter, is to make
```haskell
newtype SmashServer { unSmashServer :: URI } deriving (Show, Generic, Eq)
````
Use `fromText` as the smart constructor of `SmashServer`, and define `PoolMetadataSource` using `SmashServer`, without view patterns.

File: lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs [hunk 23]
```diff
@@ -172,19 +187,38 @@ data StakePoolLayer = StakePoolLayer
...
         -> Coin
         -> ExceptT ErrListPools IO [Api.ApiStakePool]
+
+    , putSettings :: Settings -> IO ()
+
+    , getSettings :: IO Settings
+
+    , clearMetadataTable :: IO ()
+
+    , metadataThread :: TVar ThreadId
```
rvl: I don't think we want to be exposing such implementation details in the stake pool layer interface.

rvl (commented):
:+1:

lib/core/src/Cardano/Wallet/Primitive/Types.hs
hasufell: I hope this doesn't mess with our json encoding.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
hasufell: I'm not really liking dealing with MVars directly. It's very easy to break and end up with deadlocks, infinite loops etc etc

hasufell:
bors try

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -409,6 +418,21 @@ type DelegationFee = "wallets"
     :> "delegation-fees"
     :> Get '[JSON] ApiFee
 
+{-------------------------------------------------------------------------------
+                                  Settings
+
+  See also: https://input-output-hk.github.io/cardano-wallet/api/edge/#tag/Stake-Pools
```
KtorZ: minor note: the link is incorrect here.

lib/shelley/src/Cardano/Wallet/Shelley.hs
```diff
@@ -354,21 +353,22 @@ serveWallet
...
-            let spl = newStakePoolLayer nl db
+                $ \db@DBLayer{..} -> do
+
             void $ forkFinally (monitorStakePools tr gp nl db) onExit
-            fetch <- fetchFromRemote trFetch fetchStrategies
-                <$> newManager defaultManagerSettings
-            void $ forkFinally (monitorMetadata tr gp fetch db) onExit
+            spl <- newStakePoolLayer nl db (\mvar -> forkFinally'
+                (monitorMetadata tr gp db mvar)
+                onExit)
```
KtorZ: suggestion:

```hs
spl <- newStakePoolLayer nl db $ \mvar -> forkFinally'
    (monitorMetadata tr gp db mvar)
    onExit
```

lib/shelley/src/Cardano/Wallet/Shelley.hs
```diff
@@ -354,21 +353,22 @@ serveWallet
...
         onExit = defaultWorkerAfter poolsEngineTracer
-        fetchStrategies = pure $ maybe
-            identityUrlBuilder
-            registryUrlBuilder
-            smashURL
+
+        forkFinally' :: IO a -> (Either SomeException a -> IO ()) -> IO (Async ())
+        forkFinally' action' and_then =
+          mask $ \restore ->
+            async $ try (restore action') >>= and_then
```
KtorZ: Why is this needed exactly? What's wrong with `forkIO` in this context?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -196,6 +222,43 @@ newStakePoolLayer nl db@DBLayer {..} =
...
+            as <- readMVar mas
+            void $ waitCatch as
+
+        -- clean up metadata table if the new sync settings suggests so
+        gcMetadata = do
+            oldSettings <- liftIO $ atomically $ readSettings
+            case (poolMetadataSource oldSettings, poolMetadataSource settings) of
+                (_, FetchNone) -> -- this is necessary if it's e.g. the first time
+                                  -- we start the server with the new feature
+                                  -- and the database has still stuff in it
```
KtorZ, line 247: I still do not understand why this isn't already captured by `old /= new`. Surely, if the database already has stuff in it, it also has a given setting that is not `FetchNone`?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -172,19 +188,29 @@ data StakePoolLayer = StakePoolLayer
...
-            { getPoolLifeCycleStatus = _getPoolLifeCycleStatus
-            , knownPools = _knownPools
-            , listStakePools = _listPools
-            }
+    -> (MVar Bool -> IO (Async ()))
+    -> IO StakePoolLayer
+newStakePoolLayer nl db@DBLayer {..} metadataRun' = do
+    mv <- newMVar False
+    as <- metadataRun' mv
+    mas <- newMVar as
```
KtorZ: Those variables are used across multiple functions and the name are sort of cryptic :thinking:
In particular, it is unclear why the second mvar `mas` is needed?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -592,27 +655,39 @@ monitorStakePools tr gp nl DBLayer{..} =
...
+            FetchNone -> pure ([], [])
+            FetchDirect -> (fetcher [identityUrlBuilder]) >>= fetchThem
+            FetchSMASH (unSmashServer -> uri) -> (fetcher [registryUrlBuilder uri]) >>= fetchThem
+
+        when (null refs || null successes) $ do
+            traceWith tr $ MsgFetchTakeBreak blockFrequency
+            threadDelay blockFrequency
+        readMVar stop >>= \case
+            True -> pure ()
+            False -> loop settings
```
KtorZ: It doesn't seem right that this re-implement the logic for blocking on a variable since we get this for free by using mvar. 
i.e.

by putting the _settings_ themselves in the mvar, and making the `loop` take settings only from that mvar, we ensure synchronization around the data that needs to be synchronize instead of creating additional logic to synchronize the access to the data.

specifications/api/swagger.yaml
```diff
@@ -277,6 +277,28 @@ x-walletName: &walletName
   minLength: 1
   example: Alan's Wallet
 
+x-poolMetadataSource: &poolMetadataSource
+  description: Pool metadata source. null -> no fetching, direct -> direct fetching, uri -> use SMASH server
```
KtorZ: That could be a little clearer and with a bit more prose. This is our end-user docs, so if we need to put details, diagrams and explanations, that's the right place!

KtorZ (changes requested):

lib/shelley/src/Cardano/Wallet/Shelley.hs
hasufell: Doesn't give us an async handle. That means we'd have to do extra work just to make sure the thread terminated (e.g. passing yet another MVar).

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
hasufell: This is the case of e.g. upgrading the wallet from a previous version. The default setting is `FetchNone`, but the database is already populated. Tbf... we might wanna remove this clause and somehow make sure that the database migration wipes out the table? There's more cases that can go wrong. This was the most obvious one that broke local testing.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
hasufell: > It doesn't seem right that this re-implement the logic for blocking on a variable since we get this for free by using mvar.

I cannot follow. We're not blocking here at all. The stop mvar is *never* empty.

> by putting the settings themselves in the mvar

I don't like putting settings into an MVar. This data comes from the database already.

lib/shelley/src/Cardano/Wallet/Shelley.hs
KtorZ: But thread never terminates :sweat_smile:  ?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
KtorZ: > I don't like putting settings into an MVar. This data comes from the database already.

But that's the data for which we want to ensure the synchronization! So instead of dynamically controlling the behavior of the `loop` to make it do nothing or fetch, it is much easier (and far less confusing) to use an MVar to synchronize the settings and make the loop block when the settings are being modified. 

"I don't like it" is a very weak argument :grimacing: 

If you want to make sure *something* is only used by one thread at a time, put that one thing in an MVar. Don't build extra logic around it, the MVar does it well all alone already.

hasufell: > But that's the data for which we want to ensure the synchronization! So instead of dynamically controlling the behavior of the loop to make it do nothing or fetch

That's not what happens. The thread *terminates*... it doesn't do nothing. So you're proposing something completely different. And I haven't seen an argument what's wrong with my approach. This is basically how graceful thread shutdown works.

lib/shelley/src/Cardano/Wallet/Shelley.hs
hasufell: It does.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -902,3 +925,14 @@ fromPoolMeta meta = (poolMetadataHash meta,) $
         , description = poolMetadataDescription meta
         , homepage = poolMetadataHomepage meta
         }
+
+fromSettings
+    :: Settings
+    -> W.Settings
+fromSettings (Settings pms) = W.Settings pms
```
rvl, line 969: Does `coerce` work here? They are `newtypes` aren't they? :slightly_smiling_face:

lib/core/src/Cardano/Pool/DB.hs
```diff
@@ -223,6 +228,15 @@ data DBLayer m = forall stm. (MonadFail stm, MonadIO stm) => DBLayer
...
         --    - 'removePools'.
 
+    , readSettings
+        :: stm Settings
+        -- ^ Get the settings.
+
+    , modSettings
+        :: (Settings -> Settings)
+        -> stm ()
+        -- ^ Modify the settings.
```
rvl: I would prefer `putSettings`, like is done for the other DBLayer methods. There is no need for a modify-type function when you have read and put (and transactions).

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -592,27 +655,39 @@ monitorStakePools tr gp nl DBLayer{..} =
...
+    loop settings = do
+        (refs, successes) <- case poolMetadataSource settings of
+            FetchNone -> pure ([], [])
+            FetchDirect -> (fetcher [identityUrlBuilder]) >>= fetchThem
+            FetchSMASH (unSmashServer -> uri) -> (fetcher [registryUrlBuilder uri]) >>= fetchThem
+
+        when (null refs || null successes) $ do
+            traceWith tr $ MsgFetchTakeBreak blockFrequency
+            threadDelay blockFrequency
+        readMVar stop >>= \case
```
rvl: I agree with @KtorZ - because this thread is a polling loop, it doesn't need to stop. Perhaps we confused ourselves initially thinking that some thread needs to be restarted for changed settings to take effect. No, that's not the case - the settings API request handler can wipe the metadata - next time this loop goes around, it can read settings from the DB and proceed to fetch any missing metadata.

rvl (commented):
Getting there ... 1 comment about monitorMetadata and 2 minor comments.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
hasufell: Good idea

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
hasufell: We're running in circles. I already implemented that approach and we discussed it and that I don't like to poll the database every loop for the settings (it's a waste of resources).

The other two approaches are:

1. Simple **restart-logic**, which is very general and can be applied *regardless* of the internals of the data (this is good). Even places that have no access or knowledge of the data can in theory restart the thread.
2. **Data synchronization** logic... the `putSettings` endpoint will put the new Settings value additionally into an MVar... I dislike this approach for having the mutable Settings in two places and being data specific.

KtorZ: 2. is much simpler and doesn't require extra synchronization and restart logic. It doesn't actually create two points of mutation for the settings, but it does create a synchronized view of them between the thread modifying the database, and the thread which actually needs to access the settings.

rvl: > I don't like to poll the database every loop for the settings (it's a waste of resources).

It's a waste of _a very small amount of_ resources.

hasufell: > is much simpler and doesn't require extra synchronization and restart logic. It doesn't actually create two points of mutation for the settings, but it does create a synchronized view of them between the thread modifying the database, and the thread which actually needs to access the settings.

I think for the problem at hand, all 3 implementations are viable and the difference between them isn't very big. However, I've spent some time implementing the thread restart logic and I quite like it. If I go for 2. then I will have implemented every single one of them in the course of this PR :smile: 

Thread restart logic has also the advantage that we could move more logic into the thread start part:

```hs
monitorMetadata tr gp DBLayer{..} stop = do
    -- thread start logic
    settings <- atomically readSettings

   -- thread loop
    loop settings
```

So we could even move the dropping of the table into this part, instead of convoluting `pushSettings` with that logic and not worry about the performance of the start phase, since it's not part of the loop.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
KtorZ: >  haven't seen an argument what's wrong with my approach.

The argument is the complexity overhead that comes with managing the thread life-cycle. Since there's no need to start and restart a new thread, having to use two MVar to control the execution is complex and could be easily avoided. To be 100% clear, I am not saying that this current solution doesn't work, I am saying that there's a simpler solution which doesn't induce as much complexity (which is good for maintainers). What the loop does for fetching data is all clear now, it won't be in a month. So the simpler, the better. 

Don't get too attached to code, we throw code away every day. That's what the "soft" part in software is for :upside_down_face:

hasufell: I believe thread restarting is the superior approach. And I forgot to mention that it's the only approach that cleanly allows "read after write" and returns a true 200. With the other approaches, you either:

* risk a race condition if you drop the table in `pushSettings`
* can only return 202, because you have no knowledge when the setting was actually applied

I think 200 is far better for daedalus to consume as an API user.

rvl (commented):
@hasufell - Re: https://github.com/input-output-hk/cardano-wallet/pull/2156#discussion_r497356831
> I think for the problem at hand, all 3 implementations are viable and the difference between them isn't very big. However, I've spent some time implementing the thread restart logic and I quite like it.

When reviewing, what I want to be approving is the simplest solution which meets requirements. We have given you an alternative which is simpler.

Please do not add `mask_` blocks. Nobody wants to have to think about these, unless absolutely necessary.

specifications/api/swagger.yaml
```diff
@@ -277,6 +277,34 @@ x-walletName: &walletName
   minLength: 1
   example: Alan's Wallet
 
+x-poolMetadataSource: &poolMetadataSource
+  description: |
+    Pool metadata source. This sets the metadata fetching strategy.
+
+    Possible vales are
```
KtorZ: vales ~> values

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -617,30 +677,42 @@ monitorStakePools tr gp nl DBLayer{..} =
...
-    when (null refs || null successes) $ do
-        traceWith tr $ MsgFetchTakeBreak blockFrequency
-        threadDelay blockFrequency
+monitorMetadata tr gp DBLayer{..} = do
+    settings <- atomically readSettings
+    forever $ do
+        (refs, successes) <- case poolMetadataSource settings of
+            FetchNone -> pure ([], [])
+            FetchDirect -> (fetcher [identityUrlBuilder]) >>= fetchThem
+            FetchSMASH (unSmashServer -> uri) -> (fetcher [registryUrlBuilder uri]) >>= fetchThem
```
KtorZ: The benefits of the view pattern here is debatable :sweat_smile: ...

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -199,6 +224,41 @@ newStakePoolLayer nl db@DBLayer {..} =
...
         Set.fromList <$> liftIO (atomically listRegisteredPools)
 
+    -- In order to apply the settings, we have to restart the
+    -- metadata sync thread as well to avoid race conditions.
+    _putSettings :: TVar (Async ()) -> Settings -> IO ()
+    _putSettings tvAsyncHandle settings = do
+        killSyncThread
+        gcMetadata
+        writeSettings
+        restartSyncThread
```
KtorZ: If anything goes wrong with `gcMetadata` or `writeSettings` we will completely loose the synchronization thread. Also we are currently removing pools from the database and putting the settings in two separate atomic operations. So there's also room for things going wrong here. Suggestion to accommodate these issues:

a) Have `readSettings`, `putSettings` and `delPoolMetadata` in the _same_ database "transaction" (within a single atomic). Otherwise, we open up hard-to-debug concurrency issues, 

b) Move `restartSyncThread` in a `finally` clause so that we can be sure that the thread is restarted no matter what. Could be possibly done with a bracket-style combinator: `bracket killThread restartThread modifySettings`

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -617,30 +677,42 @@ monitorStakePools tr gp nl DBLayer{..} =
...
+            FetchDirect -> (fetcher [identityUrlBuilder]) >>= fetchThem
+            FetchSMASH (unSmashServer -> uri) -> (fetcher [registryUrlBuilder uri]) >>= fetchThem
+
+        when (null refs || null successes) $ do
+            traceWith tr $ MsgFetchTakeBreak blockFrequency
+            threadDelay blockFrequency
   where
+    fetcher fetchStrategies =
+        fetchFromRemote trFetch fetchStrategies
+            <$> newManager defaultManagerSettings
```
KtorZ: Does this create a new manager for every fetch :thinking: ? Wouldn't it be better to factor out the creation of the network manager outside of the forever loop and re-use it on each loop?

KtorZ: Screenshot from 2020-10-01 11-07-16

lib/shelley/src/Cardano/Wallet/Shelley.hs
```diff
@@ -354,21 +352,20 @@ serveWallet
...
-                $ \db -> do
-            let spl = newStakePoolLayer nl db
+                $ \db@DBLayer{..} -> do
+
+            forM_ settings $ atomically . putSettings
             void $ forkFinally (monitorStakePools tr gp nl db) onExit
-            fetch <- fetchFromRemote trFetch fetchStrategies
-                <$> newManager defaultManagerSettings
-            void $ forkFinally (monitorMetadata tr gp fetch db) onExit
+            spl <- newStakePoolLayer nl db $ forkFinally' (monitorMetadata tr gp db) onExit
```
KtorZ: `onExit` here will print an error in the logs every time the worker exits saying that it has exited unexpectedly. That is **not** what we want. 
Logging something is a good idea, but it should not be an error and not suggests that something unexpected happened.

hasufell:
> Please do not add mask_ blocks. Nobody wants to have to think about these, unless absolutely necessary.

This was requested by @KtorZ .

KtorZ:
@hasufell *suggested, not requested :) 
My own preference is still to put settings in an MVar and share that MVar with the monitoring thread without doing any restarting. You dislike this approach, so I was suggesting an alternative based on your approach which reduces the complexity regarding thread management.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
hasufell: I know, it's cool :sunglasses:

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
hasufell: Good catch

hasufell:
bors try

hasufell:
bors try

KtorZ (approved):

hasufell:
bors try

rvl:
Doh weeder.

bors r+

Issue #2154, opened 2020-09-21
DB operation `removePools` doesn't remove pool metadata.


jonathanknowles:
# Context

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | b8d0cfbed5eef5fc110900b990f582b7869acc50 (`master`) |
| Platform     | All  |
| Installation | From Source |

# Steps to Reproduce

1. Create a pools database. This can be done in many ways, but one method is to connect a wallet to `mainnet` and wait for the pool worker to synchronize with the tip of `mainnet`.

2. Manually invoke the `removePools` DB operation for some pool **_p_** that doesn't share a metadata hash with any other pool.

## Expected behaviour

* All rows within `pool_registrations` and `pool_retirements` corresponding to pool **_p_** are removed.
* All rows within `pool_metadata` corresponding to **_p_** are removed, **_provided that_** no other pool references these rows.

## Actual behaviour
 
* All rows within `pool_registrations` and `pool_retirements` corresponding to pool **_p_** are removed.
* No rows within `pool_metadata` corresponding to **_p_** are removed, **_even if_** no other pool references these rows.

# Resolution 

After some discussion, it was decided that we will not fix this (for the moment).

Strictly speaking, when removing pools, we should also be able to remove any pool metadata that becomes unreachable as a result.

However, this is tricky to get right, since it's possible for more than one pool to share the same metadata hash.

For example, if two pools **_p_** and **_q_** issue registration certificates that share the same metadata hash **_h_**, and then we garbage collect pool **_p_**, we have to be careful to not remove the metadata corresponding to **_h_** in order to preserve the metadata for pool **_q_**.

One possible solution would be to remove metadata entries only when we can show that they are no longer reachable: when metadata entries are no longer referenced by any pool.

However, for the moment, we adopt the principle of never removing metadata. We can revise this decision in future if the metadata table grows too large.

# QA 

Test that `removePools` does not remove pool metadata. See https://github.com/input-output-hk/cardano-wallet/pull/2160

jonathanknowles:
After some discussion, it was decided that we will not fix this (for the moment).

Strictly speaking, when removing pools, we should also be able to remove any pool metadata that becomes unreachable as a result.

However, this is tricky to get right, since it's possible for more than one pool to share the same metadata hash.

[...]

PR #2160, opened 2020-09-22
Test that `removePools` does not remove pool metadata.

jonathanknowles:
# Issue Number

#2154 

# Overview

This PR adds documentation and tests that `removePools` does not remove metadata.
    
Strictly speaking, when removing pools, we should also be able to remove any pool metadata that becomes unreachable as a result.

However, this is tricky to get right, since it's possible for more than one pool to share the same metadata hash.

For example, if two pools **_p_** and **_q_** issue registration certificates that share the same metadata hash **_h_**, and then we garbage collect pool **_p_**, we have to be careful to not remove the metadata corresponding to h in order to preserve the metadata for pool **_q_**.

One possible solution would be to remove metadata entries only when we can show that they are no longer reachable: when metadata entries are no longer referenced by any pool.
    
However, for the moment, we adopt the principle of never removing metadata. We can revise this decision in future if the metadata table grows too large.

### DIFF
--- a/lib/core/test/unit/Cardano/Pool/DB/Properties.hs
+++ b/lib/core/test/unit/Cardano/Pool/DB/Properties.hs
@@ -845,10 +845,22 @@ prop_removePools
     -> [PoolCertificate]
     -> Property
 prop_removePools
-    DBLayer {..} certificates =
-        monadicIO (setup >> prop)
+    DBLayer {..} certificates = checkCoverage
+        $ cover 8 (notNull poolsToRemove && notNull poolsToRetain)
+            "remove some pools and retain some pools"
+        $ cover 2 (null poolsToRemove)
+            "remove no pools"
+        $ cover 2 (null poolsToRetain)
+            "retain no pools"
+        $ cover 8 (notNull metadataToRetain)
+            "retain some metadata"
+        $ cover 2 (null metadataToRetain)
+            "retain no metadata"
+        $ monadicIO (setup >> prop)
   where
     setup = run $ atomically cleanDB
+
+    notNull = not . null
 
     prop = do
         -- Firstly, publish an arbitrary set of pool certificates:
--- a/lib/core/test/unit/Cardano/Pool/DB/Properties.hs
+++ b/lib/core/test/unit/Cardano/Pool/DB/Properties.hs
@@ -42,6 +42,8 @@ import Cardano.Wallet.Primitive.Types
     , PoolRegistrationCertificate (..)
     , PoolRetirementCertificate (..)
     , SlotNo (..)
+    , StakePoolMetadata (..)
+    , StakePoolTicker (..)
     , getPoolCertificatePoolId
     , getPoolRetirementCertificate
     )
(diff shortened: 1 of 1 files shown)


lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -483,6 +485,15 @@ newDBLayer trace fp timeInterpreter = do
...
         removePools = mapM_ $ \pool -> do
             liftIO $ traceWith trace $ MsgRemovingPool pool
+            -- Remove all metadata associated with the pool, bearing in mind
+            -- that a single pool can have several historical metadata hashes
+            -- associated with it:
+            metadataHashesToRemove <-
+                mapMaybe (poolRegistrationMetadataHash . entityVal) <$>
+                selectList [ PoolRegistrationPoolId ==. pool ] []
+            forM_ metadataHashesToRemove $ \metadataHash ->
+                deleteWhere [ PoolMetadataHash ==. metadataHash ]
```
KtorZ: Hmmm. That is actually not okay ^^"
Multiple pools could have the same metadata and same metadata hash. This would open up attacks such that pool B registers after pool A with the same metadata, and immediately de-register; clearing up pool's A metadata and making it disappear from the listing.

I suggest instead to add a proper comment to that `removePools` function explaining why metadata are not cleaned up and treated as an immutable data-store.

KtorZ (changes requested):
Cf previous comment, we cannot clear metadata.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
jonathanknowles: @KtorZ wrote:
> Multiple pools could have the same metadata and same metadata hash.
> This would open up attacks such that pool B registers after pool A with the same metadata, and immediately de-register; clearing up pool's A metadata and making it disappear from the listing.

That's a really good point.

> I suggest instead to add a proper comment to that `removePools` function explaining why metadata are not cleaned up and treated as an immutable data-store.

How about only removing metadata when it is _no longer reachable_?

When removing pool **_p_** with metadata hash **_h_**:

1. First remove all rows corresponding to pool **_p_** from `pool_registrations` and `pool_retirements`, as before.
2. Now find the count **_c_** of all rows in `pool_registrations` with `metadata_hash` = **_h_**.
3. If **_c_** **= 0**, then remove all entries from `pool_metadata` with `metadata_hash` == **_h_**, otherwise leave this table _untouched_.

jonathanknowles: After some discussion, it was decided that we will **not** fix this (for the moment).

Strictly speaking, when removing pools, we should also be able to remove any pool metadata that becomes unreachable as a result (as described in the [above comment](https://github.com/input-output-hk/cardano-wallet/pull/2160#discussion_r492565691)).

However, for the moment, to keep things simple, we adopt the principle of never removing metadata. We can revise this decision in future if the metadata table grows too large.

jonathanknowles:
> Cf previous comment, we cannot clear metadata.

@KtorZ I've revised this PR, with the following changes:

- I've added a comment explaining why we've chosen to not garbage collect any metadata.
- I've updated `prop_removePools` to verify that no metadata are removed by either implementation.

If and when we decide to garbage collection unreachable metadata as part of `removePools`, we can revise this property.

KtorZ (approved):
LGTM

KtorZ:
bors merge

PR #2161, opened 2020-09-22
cache latest checkpoint in-memory to avoid deserializing and reserializing it too often

KtorZ:
# Issue Number

# Overview

  This little change has some drastic performance improvements on
  various tasks of the wallet:

  ```
  Wallet Overview
     number of addresses: 29209
     number of transactions: 48819
     number of utxos: 2778 UTxOs

  Bench                   | No Cache | With Cache
  ---                     | --       | ---
  restoreTime             | 63min    | 22min
  readWalletTime          | 715ms    | 53.70 ms
  listAddressesTime       | 1.863s   | 2.169 ms
  listTransactionsTime    | 75.72s   | 12.68 s
  importOneAddressTime    | 2.526s   | 260.4 ms
  importManyAddressesTime | 3.104s   | 1.690 s
  estimateFeesTime        | 2.103s   | 63.11 ms
  ```

  This cache approach however supposes that there's only a single wallet
  per database, which is the case and has been the case for a long time
  now. I'll remove the WalletId from the database interface in a next
  commit to clean things up a bit.

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
@@ -1668,22 +1668,7 @@ instance PersistState (Rnd.RndState t) where
         pendingAddresses <- lift $ selectRndStatePending wid
         pure $ Rnd.RndState
             { hdPassphrase = pwd
-            , accountIndex =
-                -- FIXME
-                -- In the early days when Daedalus Flight was shipped, the
-                -- wallet backend was generating addresses indexes across the
-                -- whole domain which was causing a great deal of issues with
-                -- the legacy cardano-sl:wallet ...
-                --
-                -- We later changed that to instead use "hardened indexes". Yet,
-                -- for the few wallets which were already created, we revert
-                -- this dynamically by replacing the index here.
-                --
-                -- This ugly hack could / should be removed eventually, in a few
-                -- releases from 2020-04-06.
-                if ix == 0
-                then minBound
-                else W.Index ix
+            , accountIndex = W.Index ix
             , addresses = addresses
             , pendingAddresses = pendingAddresses
             , gen = gen
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
@@ -128,6 +128,8 @@ import Data.Either
     ( isRight )
 import Data.Generics.Internal.VL.Lens
     ( (^.) )
+import Data.IORef
+    ( modifyIORef', newIORef, readIORef )
 import Data.List
     ( nub, sortOn, unzip3 )
 import Data.List.Split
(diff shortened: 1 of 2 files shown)


lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -598,6 +600,39 @@ newDBLayer trace defaultFieldValues mDatabaseFile timeInterpreter = do
...
             trace
             mDatabaseFile
+
+    cache <- newIORef Nothing
+
+    let selectLatestCheckpoint
+            :: W.WalletId
+            -> SqlPersistT IO (Maybe (W.Wallet s))
+        selectLatestCheckpoint wid =
+            liftIO (readIORef cache) >>= maybe fromDatabase (pure . Just)
```
Anviking: The cache doesn't live in `atomically`, in contrast to the rest of the DB layer. Couldn't this lead to race-conditions?

KtorZ: There can only be a single execution thread in `atomically` (constructed with an MVar lock), hence the use of a mere IORef here.

Anviking: I was imagining something more like:

```
----------------------------------------------------------------------> time
1: putCheckpoint cp1 
2:           selectLatest >>= f >>= putCheckpoint
                                                     selectLatest >>= g >>= putCheckpoint
```

1. Updates cache, but operation is cancelled before the DB is updated. Cache and DB diverges.
2. DB cp will be overwritten by the cached checkpoint the next time it is modified

But I guess the worst thing that can happen is that a somehow-failing `putCheckpoint` will end up succeeding 🤔

KtorZ: > Updates cache, but operation is cancelled before the DB is updated. Cache and DB diverges.

Cache is updated _after_ the put operation has succeeded though. If it doesn't succeed, the cache isn't updated.

lib/shelley/0.2-percent-rnd.timelog
```diff
@@ -0,0 +1,235 @@
+1600712573.805894689s -0
+1600712573.926944939s 1000
+1600712574.041849965s 2000
+1600712574.158674401s 3000
+1600712574.238137394s 4000
+1600712574.291298178s 5000
+1600712574.365636401s 6000
+1600712574.503498748s 7000
+1600712574.663584257s 8000
```
Anviking: 😅  Perhaps worth adding to `.gitignore`
?
Skärmavbild 2020-09-22 kl  12 13 36

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
Anviking: Ok, Matthias pointed out that "We create an illusion of transaction by having a top-level MVar lock.", so there shouldn't be Sqlite transactions being run and then somehow cancelled.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -598,6 +600,61 @@ newDBLayer trace defaultFieldValues mDatabaseFile timeInterpreter = do
...
+    cache <- newIORef Map.empty
+
+    let readCache :: W.WalletId -> SqlPersistT IO (Maybe (W.Wallet s))
+        readCache wid = Map.lookup wid <$> liftIO (readIORef cache)
+
+    let clearCache :: W.WalletId -> SqlPersistT IO ()
+        clearCache wid = liftIO $ modifyIORef cache $ Map.delete wid
+
+    let writeCache :: W.WalletId -> W.Wallet s -> SqlPersistT IO ()
+        writeCache wid cp = liftIO $ modifyIORef cache $ Map.alter alter wid
```
Anviking: Is there a chance of buildup of thunks here? Though I suppose we'd notice that when testing this

Anviking (approved):
Nice speedup. LGTM.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
KtorZ: I also added some notes in the code about this in place where the cache is created.

lib/shelley/0.2-percent-rnd.timelog
KtorZ: Whoops. Sounds like I a wrong `git add lib` indeed :)

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
KtorZ: Thought of it and:

- Map is strict, so values are evaluated before they are put in the Map 
- We never just write in the cache without reading and evaluating data from it first. 

So no, I don't think there's any chance that large thunks will be built here. I could switch to `modifyIORef'` for the tranquility of mind, we don't need the lazyness anyway since we have in principle already evaluated the cache to WHNF (and that's actually part of why we want to cache it!)

KtorZ:
bors try

KtorZ:
bors retry

KtorZ:
bors merge

PR #2164, opened 2020-09-22
Bump cardano-transactions

Anviking:
# Issue Number

Release.

# Overview

- [x] Bump cardano-transactions

# Comments

- List of downstream PRs included in the commit messages

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -54,7 +54,7 @@ extra-deps:
 # Not strictly a dependency at present, but building it here to get
 # access to the cardano-tx cli.
 - git: https://github.com/input-output-hk/cardano-transactions
-  commit: a68107c1682c3fd119c7e83f5adcbba2c37e1744
+  commit: 218cdc2dfea34b0261f96c96c7b937d620a8780e
 
 flags:
   # Avoid a system library which causes difficulty with cross-compilation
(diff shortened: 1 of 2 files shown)


KtorZ (approved):
We don't need any bump of these in principle. But why not.

Anviking:
> We don't need any bump of these in principle. But why not.

Their binaries are going to be included as release artefacts, so good to have the latest right?

KtorZ:
> so good to have the latest right?

The latest **release** yes. Ideally, we only want to pin released versions of cardano-addresses, cardano-transactions or cardano-node. Cardano-transactions hasn't been released in a while but we're back to it and will release soon. For now, let's just bump to the latest rev. For `cardano-addresses`, the latest release is 2.0.0 and we don't want to bump to something more recent yet.

stack.yaml
```diff
@@ -46,15 +46,15 @@ extra-deps:
 
 # cardano-addresses-2.0.0
 - git: https://github.com/input-output-hk/cardano-addresses
-  commit: 9a64226fd749513ad0ed2e43e663f7132c92faa7
+  commit: de12008ee74faeccff396f46d985cb1066328b68
```
KtorZ: Nope, let's stick to the previous commit here which corresponds to the latest release.

Anviking: Reverted.

Issue #1840, opened 2020-07-01
API/CLI support for transaction expiry


rvl:
# Context

ADP-93 Transaction scheduler

Transaction TTL/expiry should be exposed in the API and the CLI.

# Decision

- Add optional `ttl_seconds` field to  [POST /wallets/{walletId}/transactions](https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/postTransaction). This will be internally converted into slots and determines the expiry time of the transaction.
- Default TTL is 2 hours
- Add expiry information to [GET /wallets/{walletId}/transactions](https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/getTransaction) and transaction list.
  - Present expiry as a time in UTC (only exists on pending transactions).
  - Add an Expired status to the API.
- Adjust [DELETE /wallets/{walletId}/transactions/{transactionId}](https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/deleteTransaction) endpoint to deal with expired transactions.
- Make CLI option for setting transaction TTL when submitting.

# Acceptance Criteria

1. Must be able to set transaction expiry via payments API.
2. Must be able to set transaction expiry via `transaction create` CLI.
3. Time-to-live should be given in seconds rather than slots.
4. Expiry must be presented as a UTC time, and should include slot info.
5. It must be possible to clean up expired transactions from the tx history by deleting them.
6. Specifying TTL must be optional, and a reasonable default should be used.

---

# Development 

As above

# QA 

1. Integration tests are added in `lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs`
    - TRANS_TTL_01
    - TRANS_TTL_02
    - TRANS_TTL_03
    - TRANS_TTL_04
    - TRANS_TTL_DELETE_01

PR #2167, opened 2020-09-23
Add transaction TTL to payments API

rvl:
### Issue Number

ADP-93 / #1840

### Overview

- [x] Add transaction TTL to swagger spec. User provides the value in seconds.
- [x] Add transaction TTL to API types
- [x] Add TTL slot calculation to API handler function
- [x] Adjust mkStdTx to make ttl easier - it now takes the expiry slot directly.
- [x] Integration tests

### Comments

Next PRs
- [x] Allow deleting expired transactions ⇒ #2262
- [x] Add CLI option `cardano-wallet transaction create [--ttl=SECONDS]` ⇒ #2267
- [ ] Perhaps clean up the large number of function arguments and return values in transaction layer functions.

### DIFF
--- a/lib/core-integration/src/Test/Integration/Framework/DSL.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/DSL.hs
@@ -117,6 +117,7 @@ module Test.Integration.Framework.DSL
     , unsafeGetTransactionTime
     , getTxId
     , oneSecond
+    , getTTLSlots
 
     -- * Delegation helpers
     , mkEpochInfo
@@ -283,7 +284,7 @@ import Data.Quantity
 import Data.Text
     ( Text )
 import Data.Time
-    ( UTCTime )
+    ( NominalDiffTime, UTCTime )
 import Data.Time.Text
     ( iso8601ExtendedUtc, utcTimeToText )
 import Data.Word
@@ -540,8 +541,9 @@ walletId =
 minUTxOValue :: Natural
 minUTxOValue = 1_000_000
 
--- | Wallet server's chosen transaction TTL value (in slots) when none is given.
-defaultTxTTL :: SlotNo
+-- | Wallet server's chosen transaction TTL value (in seconds) when none is
+-- given.
+defaultTxTTL :: NominalDiffTime
 defaultTxTTL = 7200
 
 --
@@ -1880,7 +1882,7 @@ pubKeyFromMnemonics mnemonics =
 -- Helper for delegation statuses
 --
 getSlotParams
-    :: (Context t)
+    :: Context t
     -> IO (EpochNo, SlotParameters)
 getSlotParams ctx = do
     r1 <- request @ApiNetworkInformation ctx
@@ -1899,10 +1901,21 @@ getSlotParams ctx = do
     let sp = SlotParameters
             (EpochLength epochL)
             (SlotLength slotL)
-            (genesisBlockDate)
+            genesisBlockDate
             (ActiveSlotCoefficient coeff)
 
     return (currentEpoch, sp)
+
+-- | Converts a transaction TTL in seconds into a number of slots, using the
+-- slot length.
+getTTLSlots
+    :: Context t
+    -> NominalDiffTime
+    -> IO SlotNo
+getTTLSlots ctx dt = do
+    (_, SlotParameters _ (SlotLength _slotLenWrong) _ _) <- getSlotParams ctx
+    let slotLen = 0.2 -- fixme: this is the value from byron genesis
+    pure $ SlotNo $ ceiling $ dt / slotLen
 
 -- | Handy constructor for ApiEpochInfo
 mkEpochInfo
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
@@ -85,7 +85,7 @@ import Network.HTTP.Types.Method
 import Numeric.Natural
     ( Natural )
 import Test.Hspec
-    ( SpecWith, describe )
+    ( SpecWith, describe, pendingWith )
 import Test.Hspec.Expectations.Lifted
     ( shouldBe, shouldNotBe, shouldSatisfy )
 import Test.Hspec.Extra
@@ -602,7 +602,7 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
     let slotDiff a b = if a > b then a - b else b - a
 
     it "TRANS_TTL_01 - Pending transaction expiry" $ \ctx -> do
-        (wa, wb) <- (,) <$> fixtureWallet ctx <*> fixtureWallet ctx
+        (wa, wb) <- (,) <$> fixtureWallet ctx <*> emptyWallet ctx
         let amt = minUTxOValue :: Natural
 
         payload <- mkTxPayload ctx wb amt fixturePassphrase
@@ -633,7 +633,7 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
         slotDiff txExpectedExp txActualExp `shouldSatisfy` (< 50)
 
     it "TRANS_TTL_02 - Custom transaction expiry" $ \ctx -> do
-        (wa, wb) <- (,) <$> fixtureWallet ctx <*> fixtureWallet ctx
+        (wa, wb) <- (,) <$> fixtureWallet ctx <*> emptyWallet ctx
         let amt = minUTxOValue :: Natural
         let testTTL = 42 :: NominalDiffTime
 
@@ -665,11 +665,22 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
         slotDiff txExpectedExp txActualExp `shouldSatisfy` (< 50)
 
     it "TRANS_TTL_03 - Expired transactions" $ \ctx -> do
-        (wa, wb) <- (,) <$> fixtureWallet ctx <*> fixtureWallet ctx
+        pendingWith "#1840 this is flaky -- need a better approach"
+
+        (wa, wb) <- (,) <$> fixtureWallet ctx <*> emptyWallet ctx
         let amt = minUTxOValue :: Natural
 
-        -- this transaction is going to expire really soon.
         basePayload <- mkTxPayload ctx wb amt fixturePassphrase
+        -- Set a transaction TTL that is going to expire really soon.
+        --
+        -- The TTL wants to be small enough that it expires before it can get
+        -- into a block, but large enough that the node allows it into its
+        -- mempool.
+        --
+        -- This is probably impossible to do reliably, so this test is pending.
+        --
+        -- Perhaps we could disconnect the test cluster relay node from its
+        -- peers temporarily while letting the tx expire.
         let payload = addTxTTL 0.1 basePayload
 
         ra <- request @(ApiTransaction n) ctx
@@ -698,7 +709,7 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
             ]
 
     it "TRANS_TTL_04 - Large TTL" $ \ctx -> do
-        (wa, wb) <- (,) <$> fixtureWallet ctx <*> fixtureWallet ctx
+        (wa, wb) <- (,) <$> fixtureWallet ctx <*> emptyWallet ctx
         let amt = minUTxOValue :: Natural
         let hugeTTL = 1e9 :: NominalDiffTime
 
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
@@ -75,7 +75,7 @@ import Data.Text
 import Data.Text.Class
     ( FromText (..), ToText (..) )
 import Data.Time.Clock
-    ( UTCTime, addUTCTime )
+    ( NominalDiffTime, UTCTime, addUTCTime )
 import Data.Time.Utils
     ( utcTimePred, utcTimeSucc )
 import Data.Word
@@ -87,7 +87,7 @@ import Numeric.Natural
 import Test.Hspec
     ( SpecWith, describe )
 import Test.Hspec.Expectations.Lifted
-    ( shouldBe, shouldSatisfy )
+    ( shouldBe, shouldNotBe, shouldSatisfy )
 import Test.Hspec.Extra
     ( it )
 import Test.Integration.Framework.DSL
@@ -115,6 +115,7 @@ import Test.Integration.Framework.DSL
     , fixtureWallet
     , fixtureWalletWith
     , getFromResponse
+    , getTTLSlots
     , json
     , listAddresses
     , listAllTransactions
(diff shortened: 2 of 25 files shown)


rvl:
bors try

rvl:
Fixed code warning, doh.

bors try

lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -1814,11 +1816,22 @@ getSlotParams ctx = do
...
+-- | Converts a transaction TTL in seconds into a number of slots, using the
+-- slot length.
+getTTLSlots
+    :: Context t
+    -> NominalDiffTime
+    -> IO SlotNo
+getTTLSlots ctx dt = do
+    (_, SlotParameters _ (SlotLength _slotLenWrong) _ _) <- getSlotParams ctx
+    let slotLen = 0.2 -- fixme: this is the value from byron genesis
+    pure $ SlotNo $ ceiling $ dt / slotLen
```
KtorZ, line 1918: This could be done using an API call to `/network/parameters` ?

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -614,13 +617,81 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
...
         -- The expected expiry slot (adds the hardcoded default ttl)
-        let ttl = sl + defaultTxTTL
+        ttl <- getTTLSlots ctx defaultTxTTL
+        let txExpectedExp = sl + ttl
+
+        -- The actual expiry slot
+        let Just txActualExp = absSlotS <$> apiTx ^. #expiresAt
+
+        -- Expected and actual are fairly close
+        slotDiff txExpectedExp txActualExp `shouldSatisfy` (< 50)
```
KtorZ, line 633: I wonder how "close" we can really enforce here without making this flaky. 50 slots, that's 10 seconds, I hope it's enough.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1613,6 +1619,27 @@ signPayment ctx wid argGenChange mkRewardAccount pwd md cs = db & \DBLayer{..} -
...
+-- particular reason why we chose that duration).
+getTxExpiry
+    :: TimeInterpreter IO
+    -- ^ Context for time to slot calculation.
+    -> Maybe NominalDiffTime
+    -- ^ Time to live (TTL) in seconds from now.
+    -> IO SlotNo
+getTxExpiry ti maybeTTL = do
+    expTime <- addUTCTime ttl <$> getCurrentTime
+    -- fixme: this will explode if the user provides a TTL past the horizon.
```
KtorZ: :+1:, shall we get this done in this PR? Catching the error + handling it on the server layer should be sufficient?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -528,17 +529,23 @@ data ByronWalletPutPassphraseData = ByronWalletPutPassphraseData
...
+newtype ApiTransactionTTL = ApiTransactionTTL
+    { seconds :: NominalDiffTime
+    } deriving (Eq, Generic, Show)
+
 data PostTransactionData (n :: NetworkDiscriminant) = PostTransactionData
     { payments :: !(NonEmpty (AddressAmount (ApiT Address, Proxy n)))
     , passphrase :: !(ApiT (Passphrase "lenient"))
     , withdrawal :: !(Maybe ApiWithdrawalPostData)
     , metadata :: !(Maybe (ApiT TxMetadata))
+    , ttl :: !(Maybe ApiTransactionTTL)
```
KtorZ: shall we call this: `timeToLive` for the sake of self-documentation. TTL is common enough, but I'd rather avoid acronym in the API.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -528,17 +529,23 @@ data ByronWalletPutPassphraseData = ByronWalletPutPassphraseData
     , newPassphrase :: !(ApiT (Passphrase "raw"))
     } deriving (Eq, Generic, Show)
 
+newtype ApiTransactionTTL = ApiTransactionTTL
+    { seconds :: NominalDiffTime
+    } deriving (Eq, Generic, Show)
```
KtorZ: What about something more consistent with other API fields:

```hs
Quantity "second" NominalDiffTime
```

We have "free" JSON instances for such type already.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1555,7 +1555,7 @@ updatePendingTxForExpiryQuery wid tip =
     isExpired =
         [ TxMetaWalletId ==. wid
         , TxMetaStatus ==. W.Pending
-        , TxMetaSlotExpires >=. Just tip ]
+        , TxMetaSlotExpires <=. Just tip ]
```
KtorZ, line 1557: :+1:

specifications/api/swagger.yaml
```diff
@@ -687,6 +687,33 @@ x-transactionMetadata: &transactionMetadata
...
+    transaction, and the wallet will release the funds allocated to
+    the transaction so they can be used for other payments.
+
+    The TTL should be long enough that the transaction has time to be
+    propagated through the network and confirmed, but short enough so
+    that - in the event of failures - UTxO are returned to the wallet
+    in a timely manner.
+
+    If the TTL is not provided for a payment, a reasonable default
+    value will be used.
```
KtorZ, line 763: :+1:

specifications/api/swagger.yaml
```diff
@@ -687,6 +687,33 @@ x-transactionMetadata: &transactionMetadata
     3: { "list": [ { "int": 14 }, { "int": 42 }, { "string": "1337" } ] }
     4: { "map": [{ "k": { "string": "key" }, "v": { "string": "value" }}, { "k": { "int": 14 }, "v": { "int": 42 } }] }
 
+x-transactionTTL: &transactionTTL
+  description: |
+    The TTL (time to live) is the time period in which the transaction
+    will be accepted into node mempools.
+
+    After the TTL has lapsed, the transaction is considered
```
KtorZ, line 749: genuine English question: when would you favor "lapsed" over "elapsed" ?

specifications/api/swagger.yaml
```diff
@@ -687,6 +687,33 @@ x-transactionMetadata: &transactionMetadata
...
+    value will be used.
+
+  type: object
+  required:
+    - seconds
+  properties:
+    seconds:
+      type: integer
+      minimum: 0
+      description: TTL in seconds. Will be converted to a slot number internally.
```
KtorZ: What semantic do we give to `0` ? Unless I missed something above, this would create a transaction that immediately expire. Sounds a bit counter-intuitive / should be forbidden, don't you think ?

lib/core-integration/src/Test/Integration/Framework/DSL.hs
rvl: Yes, that API call is done with `getSlotParams` - but the slot length value will be wrong until #2246 is merged.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
rvl: In this case, I think 10 seconds is a generous margin, perhaps too generous.

Because the difference, if any, is due to slot rounding.

lib/core/src/Cardano/Wallet.hs
rvl: Yes, shall do.

lib/core/src/Cardano/Wallet/Api/Types.hs
rvl: No problem.

lib/core/src/Cardano/Wallet/Api/Types.hs
rvl: OK, thanks.

specifications/api/swagger.yaml
rvl: Lapsed is for a time period which is "over" at the end, and elapsed is for a time period that continues to pass.
But they tend to be used interchangeably.

specifications/api/swagger.yaml
rvl: Yes. The actual minimum in practice would be one slot, or maybe two. But how long is that in seconds? Impossible to say.
So I just said the minimum is zero, so that at least negative TTLs are illegal.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -617,13 +620,101 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
...
+            verify rr
+                [ expectSuccess
+                , expectField (#status . #getApiT) (`shouldNotBe` Pending)
+                ]
+            pure rr
+
+        verify rb
+            [ expectField (#status . #getApiT) (`shouldBe` Expired)
+            , expectField #expiresAt (`shouldSatisfy` isJust)
+            ]
```
piotr-iohk, line 709: Should we be able to delete expired transactions?

It seems that currently not possible, trying to delete transaction that's expired:

```
DELETE http://localhost:8090/v2/wallets/1ceb45b37a94c7022837b5ca14045f11a5927c65/transactions/a2e35dfa1166747dfd7c85388ea5cadcbcc3bdb57658c7e0dd6b7090428f1689

The response was:
Code = 403,
Message = {"code":"transaction_not_pending","message":"The transaction with id: a2e35dfa1166747dfd7c85388ea5cadcbcc3bdb57658c7e0dd6b7090428f1689 cannot be forgotten as it is not pending anymore."}
```

KtorZ: That is a good question, to be confirmed with Alex. My two cents would be that: yes; such transactions never made it to the chain and wouldn't be restored with the wallet if one were restoring on new machine. They are a local construct so we should be able to clean them up.

rvl: I think it would be a bug if it weren't possible to delete an expired transaction, so thanks for pointing that out.
I will do it in a separate PR #2262.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -617,13 +620,101 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
...
+        let txExpectedExp = sl + ttl
 
-        (view #absoluteSlotNumber <$> (apiTx ^. #expiresAt))
-            `shouldBe` Just (ApiT ttl)
+        -- The actual expiry slot
+        let Just txActualExp = absSlotS <$> apiTx ^. #expiresAt
+
+        -- Expected and actual are fairly close. Any difference should only be
+        -- due to slot rounding.
+        slotDiff txExpectedExp txActualExp `shouldSatisfy` (< 50)
```
KtorZ, line 665: What about also _waiting_ for the transaction to expire and verify that its state eventually becomes reported as expired?

lib/core/src/Cardano/Wallet.hs
```diff
@@ -43,7 +43,7 @@
 
 module Cardano.Wallet
     (
-    -- * Developement
+    -- * Development
```
KtorZ, line 46: 2-year old typo, never too late to fix :sweat_smile:

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1626,6 +1632,29 @@ signPayment ctx wid argGenChange mkRewardAccount pwd md cs = db & \DBLayer{..} -
...
+    :: TimeInterpreter IO
+    -- ^ Context for time to slot calculation.
+    -> Maybe NominalDiffTime
+    -- ^ Time to live (TTL) in seconds from now.
+    -> IO (Either PastHorizonException SlotNo)
+getTxExpiry ti maybeTTL = do
+    expTime <- addUTCTime ttl <$> getCurrentTime
+    try $ ti $ ceilingSlotAt expTime
+  where
+    ttl = fromMaybe defaultTTL maybeTTL
```
KtorZ, line 1653: :+1:

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -571,12 +571,14 @@ data PostTransactionData (n :: NetworkDiscriminant) = PostTransactionData
...
     , withdrawal :: !(Maybe ApiWithdrawalPostData)
     , metadata :: !(Maybe (ApiT TxMetadata))
+    , timeToLive :: !(Maybe (Quantity "second" NominalDiffTime))
     } deriving (Eq, Generic, Show)
 
 data PostTransactionFeeData (n :: NetworkDiscriminant) = PostTransactionFeeData
     { payments :: (NonEmpty (AddressAmount (ApiT Address, Proxy n)))
     , withdrawal :: !(Maybe ApiWithdrawalPostData)
     , metadata :: !(Maybe (ApiT TxMetadata))
+    , timeToLive :: !(Maybe (Quantity "second" NominalDiffTime))
```
KtorZ, line 581: :+1:

KtorZ (approved):

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
rvl: This test case is testing that a custom transaction expiry works, and then I test expiry and reporting of the tx status in the next test case TRANS_TTL_03.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -617,13 +620,101 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
...
+
+    it "TRANS_TTL_03 - Expired transactions" $ \ctx -> do
+        (wa, wb) <- (,) <$> fixtureWallet ctx <*> fixtureWallet ctx
+        let amt = minUTxOValue :: Natural
+
+        -- this transaction is going to expire really soon.
+        basePayload <- mkTxPayload ctx wb amt fixturePassphrase
+        let payload = addTxTTL 0.1 basePayload
```
KtorZ, line 684: Sounds like this is a bit flaky :s, we should choose a time slightly larger in front of the time slot.

rvl: Hmm yes, not good.

The TTL must be small enough that the tx expires before it can get into a block, but large enough that the node allows it into its mempool.

I need another approach to testing this which isn't flaky.

KtorZ: @rvl a possible approach could be to double-spend? If we make two transactions that are using a common UTxO, both should be "accepted" in the mempool and only one would go through ultimately. Then, we can set a large-enough but still short TTL and simply wait for either one to expire. 

So, the test could be:

a) Setup a wallet with a single UTxO
b) Send funds to A with TTL t
c) Forget transaction
d) Send funds to B with TTL t

Yet, this maybe does not test what you intend to test? If we want to check that the transaction is not inserted because it is expired (and not because it is double-spending), then it doesn't work :/

KtorZ:
bors retry

Issue #2172, opened 2020-09-23
500 "Something went wrong" when listing transactions on wallet with old metadata


piotr-iohk:
# Context
 
 

| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | `2020.9.22 (git revision: 0c9ce882bb1663e11af9e08d7f763834636e5344)` |
| Platform     | All         |
| Installation | from release            |

# Steps to Reproduce

1. Have a wallet with transactions with metadata created in 2020-09-11 wallet version
2. Update to wallet 2020-09-22
3. List transactions for the wallet

## Expected behavior

All transactions are listed.

## Actual behavior

Api returns:

"Something went wrong"

In the log there is: 

```
 [GET] /v2/wallets/e36fe2c96998cfab9581b226af0ace8b6a730aae/transactions
PersistMarshalError "Couldn't parse field `txMetaData` from table `tx_meta`. JSON schema error within the metadata item 0: \"cardano\"\nJSON object expected. Unexpected value: \"cardano\""
```

---

# Resolution 

---

# QA

rvl:
This is highly annoying, but thanks for raising the issue.

I feel inclined to solve this by adding a release note saying that if you are one of the tiny number of people who used v2020-09-11 to send or receive transactions with metadata, then you should delete your wallet database and restore from genesis.

KtorZ:
@rvl it should be possible to fallback on the old parsing method when de-serializing from the database, at least temporarily. Then, it'll be replace with the right format when re-inserted.

piotr-iohk:
:+1:

PR #2173, opened 2020-09-24
add fallback for metadata JSON parsing from pre-existing database

KtorZ:
# Issue Number

#2172

# Overview

- [ ] I have added a fallback when deserializing from db

# Comments

  This is NOT 100% safe and it assumes that pre-existing metadata that were serialized with the 'direct' format does not parse with the 'detailed' format. This is however a reasonable assumption to make provided that we only released that for a single week and the number of users was fairly limited.


piotr-iohk (approved):
LGTM.

Checked against few wallets with old metadata on testnet and works fine.

KtorZ:
bors try

Issue #2174, opened 2020-09-24
Create API address derivation path type.


jonathanknowles:
# Context

We need to be able to represent address derivation paths in the API.

# Decision

Create type `ApiAddressDerivationPath` that can encode an address derivation path.

A derivation path should be serialized to JSON as a **list**.

For example, the path `1852H/1815H/0H/0/0` should have the following JSON representation:

```json
[
  {
    "derivation_index": 1852,
    "derivation_type": "hardened"
  },
  {
    "derivation_index": 1815,
    "derivation_type": "hardened"
  },
  {
    "derivation_index": 0,
    "derivation_type": "hardened"
  },
  {
    "derivation_index": 0,
    "derivation_type": "soft"
  },
  {
    "derivation_index": 0,
    "derivation_type": "soft"
  }
]
```

Each entry in the list should have the following pair of fields:

| Field name | Type | Domain |
| -- | -- | -- |
| `derivation_index` | natural number | `[ 0 .. 2^31 − 1 ]` |
| `derivation_type` | enumeration | `[ hardened \| soft ]` | 

# Acceptance Criteria

- [x] There **must** be an `ApiAddressDerivation` type.
- [x] Values of the `ApiAddressDerivation` type **must** be serializable to JSON and deserializable from JSON.
- [x] Deserialization of a previously serialized value **must** yield the original value.
- [x] The JSON representation of an `ApiAddressDerivation` value **should** be a list.

---

# Development 

- https://github.com/input-output-hk/cardano-wallet/pull/2177

# QA 

See https://github.com/input-output-hk/cardano-wallet/pull/2177, which includes:
- JSON roundtrip encoding/decoding tests for all types.
- JSON golden tests for all types.
- negative tests for the `ApiRelativeAddressIndex` type.

paweljakubas:
I wonder if first three paths, ie., are not superfluous 1852H/1815H/0H. the first two for sure if we are dealing with ada. Last one depicting account maybe also... On the other hand, it is only presentation layer so maybe the full information does not hurt.

Another thing : we want to use relative indexes (counting from 0) also for hardened index?

rvl:
Was there anything wrong with just using a string to represent derivation path?
It's pretty simple to parse.

jonathanknowles:
> Was there anything wrong with just using a string to represent derivation path?
> It's pretty simple to parse.

Hi @rvl.

We could indeed have used a string, and you're right, it would not be very hard to parse.

[...]

PR #2177, opened 2020-09-24
Create API address derivation path type.

jonathanknowles:
# Issue Number

#2174 

# Overview

This PR adds a new type `ApiAddressDerivationPath`. A value of this type serializes to a JSON **list**.

For example, the path `1852H/1815H/0H/0/0` has the following JSON representation:

```json
[
  {
    "derivation_index": 1852,
    "derivation_type": "hardened"
  },
  {
    "derivation_index": 1815,
    "derivation_type": "hardened"
  },
  {
    "derivation_index": 0,
    "derivation_type": "hardened"
  },
  {
    "derivation_index": 0,
    "derivation_type": "soft"
  },
  {
    "derivation_index": 0,
    "derivation_type": "soft"
  }
]
```

Each entry in the list has the following pair of fields:

| Field name | Type | Domain |
| -- | -- | -- |
| `derivation_index` | natural number | `[ 0 .. 2^31 − 1 ]` |
| `derivation_type` | enumeration | `[ hardened \| soft ]` | 

# Details

This PR also:

- Adds JSON roundtrip tests for `ApiAddressDerivationPath`.
- Adds a missing JSON roundtrip test for the `ApiAddress` type.
- Fixes the upper bound of `addressIndex` to be `2^32 - 1` rather than `2^32`.
- Adds a matching swagger definition for `ApiAddressDerivationPath`.


specifications/api/swagger.yaml
```diff
@@ -892,10 +892,35 @@ x-delegationTarget: &delegationTarget
 x-addressIndex: &addressIndex
   type: number
   minimum: 0
-  maximum: 4294967296 # 2 ^ 32, whole range (soft and hard) for random addresses.
+  # 2 ^ 32 - 1, whole range (soft and hard) for random addresses:
+  maximum: 4294967295
```
paweljakubas, line 896: the question for me is whether to differentiate soft and hardened index
we need to remember that hardened one can assume values 
```
<0x80000000, maxBound @Word32>
```
and soft index can be 
```
<0, 0x80000000 -1>
```
If we use the same index for both soft and hardened then the index can be between
```
<0, 0x80000000 - 1> = <0,2147483648-1>
```
So the `maximum` is wrong! BUT, I remember about WholeDomain !!! and here we need this maximum in `x-addressIndex`. So maybe introduce `x-addressRelativeIndex` instead? Or better, use the one above with revised maximum, and use for other API using this the above but renamed to `x-addressesLegacyIndex`

paweljakubas (changes requested):
If we use absolute values for soft and hardened index IMHO we should introduce two address indices in swagger and specify properly their minimum and maximum. If we want to use relative index, then maximum is wrong to my best knowledge

specifications/api/swagger.yaml
jonathanknowles: > If we use the same index for both soft and hardened then the index can be between
> ```<0, 0x80000000 - 1> = <0,2147483648-1>```
> So the maximum is wrong! BUT, I remember about `WholeDomain` !!! and here we need this maximum in `x-addressIndex`. So maybe introduce `x-addressRelativeIndex` instead.

Using a relative index sounds like a nice solution here.

If we use the following `x-relativeAddressIndex` definition:

```yaml
x-relativeAddressIndex: &relativeAddressIndex 
  type: number
  minimum: 0  
  # 2 ^ 31 - 1
  maximum: 2147483647 
  description: |  
    A relative address derivation index.      
```

This means that our API type can also be very simple:
```hs
newtype ApiRelativeAddressIndex = ApiRelativeAddressIndex 
    { unApiRelativeAddressIndex :: Word31 
    } deriving (Bounded, Enum, Eq, Generic, Show)
```

Parsing an index value is now very simple, as we just need to check that it lies within the range `[0 ... 2^31 - 1]`.

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -440,6 +449,27 @@ spec = do
                 ["toilet", "toilet", "toilet"]
             |] `shouldBe` (Left @String @(ApiMnemonicT '[12]) msg)
 
+
+        it "ApiRelativeAddressIndex (too small)" $ do
+            let message = mconcat
+                  [ "Error in $: "
+                  , "\"A relative address index must be a natural number "
+                  , "between 0 and 2147483647.\""
```
paweljakubas, line 457: :+1:

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -440,6 +449,27 @@ spec = do
...
+                  ]
+            let value = pred $ toInteger $ unApiRelativeAddressIndex minBound
+            Aeson.parseEither parseJSON [aesonQQ|#{value}|]
+                `shouldBe` Left @String @ApiRelativeAddressIndex message
+
+        it "ApiRelativeAddressIndex (too large)" $ do
```
paweljakubas, line 463: :+1:

paweljakubas (approved):
LGMT!

Issue #2175, opened 2020-09-24
Explore exposing integration tests cluster for a wider audience


KtorZ:
# Context

We use a quite sophisticated setup for our integration tests, with a cluster of pools and pre-registered wallets with funds, rewards etc ... That's useful for testing the wallet, but also for a wider audience looking for testing things with Shelley.

# Decision

# Acceptance Criteria

1. One _must_ be able to start an integration test cluster easily from an executable.
2. The tool _may_ not have a command-line or any sophisticated UX as it is meant for internal use primarily. 

---

# Development 

- Create a standalone executable which can start an integration cluster ==> #2178 

# QA 

- This is for internal use mainly, so I admit doing only very little smoke-testing. The code that spins up the cluster is actually very small and re-use constructors and helpers used for our actual integration tests so I am quite confident that it _just works_ :upside_down_face:

piotr-iohk:
It just works indeed. :tada:

PR #2178, opened 2020-09-24
expose an executable 'shelley-test-cluster' which starts an integration test cluster with faucets

KtorZ:
# Issue Number

#2175 

# Overview

- 9722d827cbfd76ae3ee89cba55a5d0db222063a9
  :round_pushpin: **expose an executable 'shelley-test-cluster' which starts an integration test cluster with faucets**
  
- 6eefd4713ef234a247ff3857a9a79a6b852718a4
  :round_pushpin: **Let the shelley-test-cluster nix build be run from any directory**
  
- 4c4d01efc2aa5a8d6bd51c3f3ab64113cae0a639
  :round_pushpin: **Allow setting different LoggerName when configuring trace with 'withLogging'**
  
- 571e67db538d1f8ed23ef4e1eb0caa1c6467868e
  :round_pushpin: **Add TestLog to shelley-test-cluster**
  
- e99d875d0f55ce6a73dc36e676b75c6cc5eb10ad

# Comments

~~@Anviking there's probably some clashes with what you're also doing with the logging, so happy to get the logging revision merged first and adjust that one.~~

### DIFF
--- a/lib/shelley/exe/shelley-test-cluster.hs
+++ b/lib/shelley/exe/shelley-test-cluster.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeApplications #-}
@@ -11,6 +12,8 @@ import Prelude
 
 import Cardano.BM.Data.Severity
     ( Severity (..) )
+import Cardano.BM.Data.Tracer
+    ( HasPrivacyAnnotation (..), HasSeverityAnnotation (..) )
 import Cardano.CLI
     ( LogOutput (..), withLoggingNamed )
 import Cardano.Startup
@@ -18,7 +21,7 @@ import Cardano.Startup
 import Cardano.Wallet.Api.Types
     ( EncodeAddress (..) )
 import Cardano.Wallet.Logging
-    ( stdoutTextTracer, trMessageText )
+    ( trMessageText )
 import Cardano.Wallet.Primitive.AddressDerivation
     ( NetworkDiscriminant (..) )
 import Cardano.Wallet.Primitive.SyncProgress
@@ -53,9 +56,13 @@ import Control.Arrow
 import Control.Monad
     ( void )
 import Control.Tracer
-    ( traceWith )
+    ( contramap, traceWith )
 import Data.Proxy
     ( Proxy (..) )
+import Data.Text
+    ( Text )
+import Data.Text.Class
+    ( ToText (..) )
 import System.IO
     ( BufferMode (..), hSetBuffering, stdout )
 import Test.Integration.Faucet
@@ -204,24 +211,26 @@ main = do
         $ \(_, trCluster) -> withSystemTempDir (trMessageText trCluster) "testCluster"
         $ \dir -> withTempDir (trMessageText trCluster) dir "wallets"
         $ \db -> withCluster
-            (trMessageText trCluster)
+            (contramap MsgCluster $ trMessageText trCluster)
             nodeMinSeverity
             defaultPoolConfigs
             dir
             Nothing
             whenByron
-            (whenShelley dir)
-            (whenReady trWallet trCluster db)
+            (whenShelley dir (trMessageText trCluster))
+            (whenReady trWallet (trMessageText trCluster) db)
   where
     whenByron _ = pure ()
 
-    whenShelley dir _ = do
+    whenShelley dir trCluster _ = do
+        traceWith trCluster MsgSettingUpFaucet
+        let trCluster' = contramap MsgCluster trCluster
         let encodeAddr = T.unpack . encodeAddress @'Mainnet
         let addresses = map (first encodeAddr) shelleyIntegrationTestFunds
         let accts = concatMap genRewardAccounts mirMnemonics
         let rewards = (,Coin $ fromIntegral oneMillionAda) <$> accts
-        sendFaucetFundsTo stdoutTextTracer dir addresses
-        moveInstantaneousRewardsTo stdoutTextTracer dir rewards
+        sendFaucetFundsTo trCluster' dir addresses
+        moveInstantaneousRewardsTo trCluster' dir rewards
 
     whenReady tr trCluster db (RunningNode socketPath block0 (gp, vData)) = do
         let tracers = setupTracers (tracerSeverities (Just Info)) tr
@@ -239,4 +248,26 @@ main = do
             socketPath
             block0
             (gp, vData)
-            (traceWith (trMessageText trCluster) . MsgListenAddress)
+            (traceWith trCluster . MsgBaseUrl . T.pack . show)
+
+-- Logging
+
+data TestsLog
+    = MsgBaseUrl Text
+    | MsgSettingUpFaucet
+    | MsgCluster ClusterLog
+    deriving (Show)
+
+instance ToText TestsLog where
+    toText = \case
+        MsgBaseUrl addr ->
+            "Wallet backend server listening on " <> T.pack (show addr)
+        MsgSettingUpFaucet -> "Setting up faucet..."
+        MsgCluster msg -> toText msg
+
+instance HasPrivacyAnnotation TestsLog
+instance HasSeverityAnnotation TestsLog where
+    getSeverityAnnotation = \case
+        MsgSettingUpFaucet -> Notice
+        MsgBaseUrl _ -> Notice
+        MsgCluster msg -> getSeverityAnnotation msg
--- a/lib/shelley/exe/shelley-test-cluster.hs
+++ b/lib/shelley/exe/shelley-test-cluster.hs
@@ -12,7 +12,7 @@ import Prelude
 import Cardano.BM.Data.Severity
     ( Severity (..) )
 import Cardano.CLI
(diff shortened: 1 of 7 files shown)


lib/shelley/exe/shelley-test-cluster.hs
```diff
@@ -0,0 +1,224 @@
...
+-- - (Shelley) Has a pre-registered stake key but no delegation.
+--
+--     [ "over", "decorate", "flock", "badge", "beauty"
+--     , "stamp", "chest", "owner", "excess", "omit"
+--     , "bid", "raccoon", "spin", "reduce", "rival"
+--     ]
+--
+-- - (Shelley) Contains only small coins (but greater than the minUTxOValue)
+--
+--     [ "either" , "flip" , "maple" , "shift" , "dismiss"
```
Anviking, line 122: Nice docs, but feels like it might get out of date very easily? Also is a doc comment actually the right place for this information?

Maybe we could add a new `shelley-test-cluster mnemonics`, and change `shelleyIntegrationTestFunds` to also include a `String` description for each wallet?

The output of such a command would be huge. We could add some logic to protect against it, or maybe just something like grep:

E.g.
```bash
$ shelley-test-cluster query-mnemonic "small coins"
shelley/special/seq/Contains only small coins (but greater than the minUTxOValue) 1
    either flip maple shift dismiss bridge sweet reveal green tornado need patient wall stamp pass

shelley/special/seq/Contains only small coins (but greater than the minUTxOValue) 2
    reveal sweet green tornado need patient wall stamp pass either flip maple shift dismiss bridge

```

KtorZ: > Nice docs, but feels like it might get out of date very easily? Also is a doc comment actually the right place for this information?

I wondered the same and actually considered the same idea you're proposing here. Although arguably: 

- we haven't changed those mnemonics in several months and have no reason to change them in the upcoming future. 
- that is for now, mostly an internal tool, so if this end up outdated, we'll most likely notice it quite quickly. 

So, the size of the problem seems not large enough to spend effort on making this better :sweat_smile: ... arguably.

KtorZ: I mean, the same argument applies for the pools configurations and even the cluster setup (what if we add a new node tomorrow? the doc becomes outdated). All this doc describes implementation details, so by its very nature, it's fragile.

Anviking: I remember having to re-regenerate and update faucet mnemonics several for Shelley, which was not fun at all. With the HFC integration we finally got it much more automated.

I think just adding a command to output all mnemonics with a description, requiring manual `| grep ""` piping, would have been really simple.

But sure, it's perhaps unlikely they'll change now. And perhaps still best to keep it comment, for now.

Btw, I wonder if this could be used for ADP-381 eventually.

KtorZ: Perhaps ultimately, we can clean-up our integration tests setup to use that executable. Spawn it in a process and go ahead with the rest of the tests. Right now the integration setup is very complicated with contexts and mvars a bit everywhere waiting for stuff.

rvl (approved):

Anviking:
I can skim over again when rebased.

But trying now I get:
```
  cardano-cli: pokeSockAddr: path is too long
    error, called at Network/Socket/Types.hsc:1077:40 in network-3.1.1.1-Izwsyk64OoDBVuRysfRa9D:Network.Socket.Types
```

```
λ> length "/private/var/folders/h_/mlzym27d5lz_fx3ph2d547700000gn/T/shelley-test-cluster-ff9e0fd9e9507b24/bft/node.socket"
110
```
seems 6 characters [too long](https://unix.stackexchange.com/questions/367008/why-is-socket-path-length-limited-to-a-hundred-chars) on macOS. Using `withSystemTempDir trCluster "shelleycluster"` instead of `"shelley-test-cluster"` seems to fix it.

lib/shelley/exe/shelley-test-cluster.hs
```diff
@@ -0,0 +1,224 @@
...
+-- - CARDANO_WALLET_PORT: choose a port for the API to listen on (default: random)
+--
+-- - CARDANO_NODE_TRACING_MIN_SEVERITY: increase or decrease the logging
+--                                      severity of the nodes. (default: Error)
+--
+-- - CARDANO_WALLET_TRACING_MIN_SEVERITY: increase of decrease the logging
+--                                        severity of cardano-wallet. (default: Info)
+--
+-- - NO_CLEANUP: If set, the temporary directory used as a state directory for
+--               nodes and wallet data won't be cleaned up.
```
Anviking, line 191: Why no `NO_POOLS` option for fast boot?

KtorZ: Oversight :+1:

lib/shelley/exe/shelley-test-cluster.hs
```diff
@@ -0,0 +1,242 @@
...
+            (whenReady trWallet trCluster db)
+  where
+    whenByron _ = pure ()
+
+    whenShelley dir _ = do
+        let encodeAddr = T.unpack . encodeAddress @'Mainnet
+        let addresses = map (first encodeAddr) shelleyIntegrationTestFunds
+        let accts = concatMap genRewardAccounts mirMnemonics
+        let rewards = (,Coin $ fromIntegral oneMillionAda) <$> accts
+        sendFaucetFundsTo stdoutTextTracer dir addresses
```
Anviking: We should use `trCluster` instead of `stdoutTextTracer` here! (Was changed recently in cardano-wallet:integration) Currently we see a huge about of logs about faucet txs when running `shelley-test-cluster`.

KtorZ:
bors merge

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -1648,6 +1715,8 @@ instance ToText ClusterLog where
         MsgGenOperatorKeyPair dir ->
             "Generating stake pool operator key pair in " <> T.pack dir
         MsgCLI args -> T.pack $ unwords ("cardano-cli":args)
+        MsgListenAddress addr ->
+            "Wallet backend server listening on " <> T.pack (show addr)
```
Anviking: I don't think this should be added.

For cardano-wallet:integration there is a separation between `TestsLog` and `ClusterLog`, where `MsgListenAddress` would make sense in `TestsLog`, but _not_ `ClusterLog`.

We could add a similar `TestsLog` for this executable in a follow-up PR.

lib/shelley/exe/shelley-test-cluster.hs
Anviking: Not sure if an additional oversight or not, but still doesn't work.

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -316,6 +327,58 @@ parseGenesisData = \case
...
+-- @CARDANO_NODE_TRACING_MIN_SEVERITY@ environment variable.
+nodeMinSeverityFromEnv :: IO Severity
+nodeMinSeverityFromEnv =
+    minSeverityFromEnv Info "CARDANO_NODE_TRACING_MIN_SEVERITY"
+
+-- Allow configuring integration tests and wallet log level with
+-- @CARDANO_WALLET_TRACING_MIN_SEVERITY@ environment variable.
+walletMinSeverityFromEnv :: IO Severity
+walletMinSeverityFromEnv =
+    minSeverityFromEnv Info "CARDANO_WALLET_TRACING_MIN_SEVERITY"
```
Anviking: This re-adds the verbose logs running `cardano-wallet:integration`

Anviking: Not sure how we want to fix

Anviking (changes requested):
Unless someone wants needs this urgently, can we address at least https://github.com/input-output-hk/cardano-wallet/pull/2178#discussion_r498378506 first?

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
Anviking: A quick fix could be to let it take the default as an argument or return a `Maybe`.

KtorZ: Woops. Probably a wrong merge resolution here :+1:

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
KtorZ: Hmmm. Possibly. Although this one should really be in the application logs of the wallet.

Anviking:
`anviking/2175/integration-test-cluster-exposition`

Suggestion of direction:

```
commit 339fe2bf19689613e663da2bbdaec1ea67804a1f (HEAD -> anviking/2175/integration-test-cluster-exposition, origin/anviking/2175/integration-test-cluster-exposition)
Author: Johannes Lund <johannes.lund@iohk.io>
Date:   Thu Oct 1 18:40:48 2020 +0200

    Add TestLog to shelley-test-cluster
```

Anviking (approved):
👍

lib/shelley/exe/shelley-test-cluster.hs
Anviking: 8da9ccc

KtorZ:
bors retry

KtorZ:
bors merge

Anviking:
_Manually tweaked .nix file_

PR #2181, opened 2020-09-24
When a retried test fails, show the first error

Anviking:
# Issue Number

#2192 

# Overview

- When both `it` attempt fails, show the _first_ error, not the second.

# Comments

If a test uses a finite resource, it cannot be retried, and will fail for another reason the second time. In this case, it is the first failure that is interesting.

In other circumstances, showing the first error instead of the second should still be fine.

Haven't actually tried this yet. Should maybe do that. Or add a test suite for `test-utils`.


KtorZ (approved):
:+1:

KtorZ:
bors merge

Issue #2180, opened 2020-09-24
Provide a way to validate addresses via the API. 


KtorZ:
# Context

Exchanges prefer validating data before submission (even though the API will validate things upon submission). There's no easy to simply validate arbitrary string an know whether they are valid addresses or not. That feature exists in cardano-address but only available via command-line. 

# Decision

Make the API able to inspect addresses.

# Acceptance Criteria

1. There's a new API endpoint to validate addresses
2. The validation is implemented using cardano-address 'inspect' method.

---

# Development 

- Added endpoint to the API ==> #2184 
- Tweaked a bit the swagger documentation to put the endpoint in a separate category ==> #2185 

# QA 

- The `inspect` feature comes from `cardano-addresses` and is well-tested there. 

- Added two straightforward integration tests showing that the API reacts correctly to inputs: 

  https://github.com/input-output-hk/cardano-wallet/blob/d4d5cf3f1dbf478e49186f9e1742939af528bcaa/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs#L229-L237

- Make sure that the generated JSON matches the OpenAPI schema:

  https://github.com/input-output-hk/cardano-wallet/blob/d4d5cf3f1dbf478e49186f9e1742939af528bcaa/lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs#L1524-L1525

  Although, this is a bit of a lie because we don't yet have a proper type to represent the result of an address inspection. What we get from `cardano-addresses` is an opaque JSON value which (from a type perspective) could be _anything_. So we should ultimately replace this with a proper Haskell data-type so that we can have static type-checking on this:

  https://github.com/input-output-hk/cardano-wallet/blob/d4d5cf3f1dbf478e49186f9e1742939af528bcaa/lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs#L1460-L1467

- Ported the test cases from `cardano-addresses` and tested them at the unit-level. Ideally, once we have cardano-addresses to return a proper Haskell data-type, we can make the test a bit more expressive instead of simply `isRight` and `isLeft`

  https://github.com/input-output-hk/cardano-wallet/blob/d4d5cf3f1dbf478e49186f9e1742939af528bcaa/lib/shelley/test/unit/Cardano/Wallet/Shelley/CompatibilitySpec.hs#L232-L292

piotr-iohk:
[...]

PR #2184, opened 2020-09-24
Allow address inspection via the API

KtorZ:
# Issue Number

#2180 

# Overview

- 2daae7e4124147d182bf1a3989c9a08d0abcb7d8
  :round_pushpin: **implement new stateless endpoint for inspecting addresses**
    It takes any arbitrary string as input and tries to deserialize it into some address information if possible, otherwise it fails with a proper error.

- 76d13bad9b1b29bff3336af5f25afeebf31a1694
  :round_pushpin: **add unit and integration tests for the new 'inspectAddress' function & endpoint**

# Comments

Screenshot from 2020-09-24 18-52-44

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
@@ -61,6 +61,7 @@ import Test.Integration.Framework.TestData
     ( errMsg404NoWallet )
 
 import qualified Cardano.Wallet.Api.Link as Link
+import qualified Data.Aeson as Aeson
 import qualified Data.Text as T
 import qualified Network.HTTP.Types.Status as HTTP
 
@@ -224,3 +225,13 @@ spec = describe "SHELLEY_ADDRESSES" $ do
             (Link.listAddresses @'Shelley w) Default Empty
         expectResponseCode @IO HTTP.status404 r
         expectErrorMessage (errMsg404NoWallet $ w ^. walletId) r
+
+    it "ADDRESS_INSPECT_01 - Address inspect OK" $ \ctx -> do
+        let str = "Ae2tdPwUPEYz6ExfbWubiXPB6daUuhJxikMEb4eXRp5oKZBKZwrbJ2k7EZe"
+        r <- request @Aeson.Value ctx (Link.inspectAddress str) Default Empty
+        expectResponseCode @IO HTTP.status200 r
+
+    it "ADDRESS_INSPECT_02 - Address inspect KO" $ \ctx -> do
+        let str = "patate"
+        r <- request @Aeson.Value ctx (Link.inspectAddress str) Default Empty
+        expectResponseCode @IO HTTP.status400 r
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -72,6 +72,8 @@ module Cardano.Wallet.Api.Types
     , ApiTxInput (..)
     , ApiTxMetadata (..)
     , AddressAmount (..)
+    , ApiAddressInspect (..)
+    , ApiAddressInspectData (..)
     , ApiErrorCode (..)
     , ApiNetworkInformation (..)
     , ApiNtpStatus (..)
@@ -244,6 +246,8 @@ import Data.Proxy
     ( Proxy (..) )
 import Data.Quantity
     ( Percentage, Quantity (..) )
+import Data.String
+    ( IsString )
 import Data.Text
     ( Text, split )
 import Data.Text.Class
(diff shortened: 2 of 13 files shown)


rvl (approved):

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
```diff
@@ -224,3 +225,13 @@ spec = describe "SHELLEY_ADDRESSES" $ do
             (Link.listAddresses @'Shelley w) Default Empty
         expectResponseCode @IO HTTP.status404 r
         expectErrorMessage (errMsg404NoWallet $ w ^. walletId) r
+
+    it "ADDRESS_INSPECT_01 - Address inspect OK" $ \ctx -> do
+        let str = "Ae2tdPwUPEYz6ExfbWubiXPB6daUuhJxikMEb4eXRp5oKZBKZwrbJ2k7EZe"
+        r <- request @Aeson.Value ctx (Link.inspectAddress str) Default Empty
+        expectResponseCode @IO HTTP.status200 r
```
paweljakubas, line 232: any chance to also verify the result of this call here?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -606,6 +611,15 @@ data AddressAmount addr = AddressAmount
...
     } deriving (Eq, Generic, Show)
 
+newtype ApiAddressInspect = ApiAddressInspect
+    { unApiAddressInspect :: Aeson.Value }
+    deriving (Eq, Generic, Show)
+
+newtype ApiAddressInspectData = ApiAddressInspectData
+    { unApiAddressInspectData :: Text }
+    deriving (Eq, Generic, Show)
+    deriving newtype (IsString)
```
paweljakubas, line 652: could you comment why this was needed and used? (especially if we have Show above)

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -1004,6 +1010,35 @@ instance DecodeAddress 'Mainnet where
 instance DecodeAddress ('Testnet pm) where
     decodeAddress = _decodeAddress SL.Testnet
 
+decodeBytes :: Text -> Either TextDecodingError ByteString
```
paweljakubas, line 1013: we already have `cardano-addresses` dependency in this project https://github.com/input-output-hk/cardano-wallet/blob/master/stack.yaml#L48 and also have many bits present here waiting for reuse. Probably not in thi sPR but I am in favor of strong reuse of cardano-addresses primitives..

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -1047,15 +1061,27 @@ _decodeAddress serverNetwork text =
...
-                "Unable to decode address: not a well-formed Shelley nor Byron address."
+            Nothing -> Left errMalformedAddress
 
       where
         toNetwork :: Byron.NetworkMagic -> SL.Network
         toNetwork = \case
             Byron.NetworkMainOrStage -> SL.Mainnet
             Byron.NetworkTestnet{}   -> SL.Testnet
 
+-- FIXME: 'cardano-addresses' currently gives us an opaque 'Value'. It'd be
```
paweljakubas, line 1072: yes, agree https://github.com/input-output-hk/cardano-addresses/blob/master/core/lib/Cardano/Address/Style/Shelley.hs#L497

specifications/api/swagger.yaml
```diff
@@ -914,6 +914,69 @@ components:
...
+            maxLength: 56
+          stake_key_hash:
+            type: string
+            format: base16
+            minLength: 56
+            maxLength: 56
+          script_hash:
+            type: string
+            format: base16
+            minLength: 64
```
paweljakubas, line 985: hmm, I think in 1.20.0 in cardano-node they use 28-byte hash for scripts, not 32-byte

paweljakubas (approved):
LGTM! I do think we should reconsider better reuse of lean cardano-addresses library in the future

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: It allows for using the `OverloadedString` extension in a transparent fashion for literal strings like we do for `Text` and `ByteString`. So, one can write: 

```hs
let inspectData = "patate"
```

instead of having to:

```hs
let inspectData = ApiAddressInspectData "patate"
```

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
KtorZ: At some point yes, I agree. I started this PR with only cardano-addresses originally and I ended up reverting because errors were highly inconsistent between endpoints. So I prefer sticking to how it's done now (and also, locate the code doing it in the same place so it's either to change later).

specifications/api/swagger.yaml
KtorZ: Yes, but cardano-addresses still uses 32 bytes :( ... So we need to change it there, bump revision and update here. Which was on my plate today ^^

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
KtorZ: Yes and no. It's technically possible of course. But I'll rather defer that to until we have a proper data-type for representing inspected data (as highlighted in a FIXME comment). But I'd rather introduce this data-type in cardano-addresses directly.

KtorZ:
bors merge

KtorZ:
CI stuck on the mac-mini again. Moving forward with that one :(

PR #2191, opened 2020-09-29
Parallelise integration tests using ResourceT

Anviking:
# Issue Number

ADP-466 — both ADP-529 and ADP-528

# Overview

- [x] Use `runResourceT` in each test scenario to ensure only the wallet created in it are deleted.
- [x] Re-write all `listWallets` expectations to check that the expected wallets are a _subset_ of the actual wallets, instead of checking for equality.
- [x] Enable `-j 8` in buildkite
 

# Comments

- Buildkite now takes ≈35 min instead of 50 min
- On my machine `stack test cardano-wallet:integration --ta '-j 8'` takes 380s instead of ~1800s.
- Haven't enabled parallel tests in hydra yet because of errors
- I now pushed `anviking/ADP-466/parallel` (parallel _everything_) directly here.

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/HWWallets.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/HWWallets.hs
@@ -35,6 +35,10 @@ import Cardano.Wallet.Primitive.Types
     ( AddressState (..) )
 import Control.Monad
     ( forM_ )
+import Control.Monad.IO.Class
+    ( liftIO )
+import Control.Monad.Trans.Resource
+    ( runResourceT )
 import Data.Generics.Internal.VL.Lens
     ( view, (^.) )
 import Data.Quantity
@@ -64,6 +68,7 @@ import Test.Integration.Framework.DSL
     , json
     , listAddresses
     , minUTxOValue
+    , postWallet
     , pubKeyFromMnemonics
     , request
     , restoreWalletFromPubKey
@@ -86,15 +91,15 @@ spec :: forall n t.
     , EncodeAddress n
     ) => SpecWith (Context t)
 spec = describe "SHELLEY_HW_WALLETS" $ do
-    it "HW_WALLETS_01 - Restoration from account public key preserves funds" $ \ctx -> do
+    it "HW_WALLETS_01 - Restoration from account public key preserves funds" $ \ctx -> runResourceT $ do
         wSrc <- fixtureWallet ctx
         -- create wallet
-        mnemonics <- mnemonicToText @15 . entropyToMnemonic <$> genEntropy
+        mnemonics <- liftIO $ mnemonicToText @15 . entropyToMnemonic <$> genEntropy
         let wName = "!st created"
         let payldCrt = payloadWith wName mnemonics
-        rInit <- request @ApiWallet ctx (Link.postWallet @'Shelley) Default payldCrt
+        rInit <- postWallet ctx payldCrt
         verify rInit
-            [ expectResponseCode @IO HTTP.status201
+            [ expectResponseCode HTTP.status201
             , expectField (#balance . #getApiT . #available) (`shouldBe` Quantity 0)
             , expectField (#balance . #getApiT . #total) (`shouldBe` Quantity 0)
             ]
@@ -115,7 +120,7 @@ spec = describe "SHELLEY_HW_WALLETS" $ do
             }|]
         rTrans <- request @(ApiTransaction n) ctx
             (Link.createTransaction @'Shelley wSrc) Default payload
-        expectResponseCode @IO HTTP.status202 rTrans
+        expectResponseCode HTTP.status202 rTrans
 
         eventually "Wallet balance is as expected" $ do
             rGet <- request @ApiWallet ctx
@@ -130,7 +135,7 @@ spec = describe "SHELLEY_HW_WALLETS" $ do
         -- delete wallet
         rDel <-
             request @ApiWallet ctx (Link.deleteWallet @'Shelley wDest) Default Empty
-        expectResponseCode @IO HTTP.status204 rDel
+        expectResponseCode HTTP.status204 rDel
 
         -- restore from account public key and make sure funds are there
         let accXPub = pubKeyFromMnemonics mnemonics
@@ -147,11 +152,11 @@ spec = describe "SHELLEY_HW_WALLETS" $ do
                 ]
 
     describe "HW_WALLETS_03 - Cannot do operations requiring private key" $ do
-        it "Cannot send tx" $ \ctx -> do
+        it "Cannot send tx" $ \ctx -> runResourceT $ do
             (w, mnemonics) <- fixtureWalletWithMnemonics ctx
             let pubKey = pubKeyFromMnemonics mnemonics
             r <- request @ApiWallet ctx (Link.deleteWallet @'Shelley w) Default Empty
-            expectResponseCode @IO HTTP.status204 r
+            expectResponseCode HTTP.status204 r
 
             wSrc <- restoreWalletFromPubKey @ApiWallet @'Shelley ctx pubKey restoredWalletName
             wDest <- emptyWallet ctx
@@ -170,11 +175,11 @@ spec = describe "SHELLEY_HW_WALLETS" $ do
                 }|]
             rTrans <- request @(ApiTransaction n) ctx
                 (Link.createTransaction @'Shelley wSrc) Default payload
-            expectResponseCode @IO HTTP.status403 rTrans
+            expectResponseCode HTTP.status403 rTrans
             expectErrorMessage (errMsg403NoRootKey $ wSrc ^. walletId) rTrans
 
-        it "Cannot update pass" $ \ctx -> do
-            mnemonics <- mnemonicToText @15 . entropyToMnemonic <$> genEntropy
+        it "Cannot update pass" $ \ctx -> runResourceT $ do
+            mnemonics <- liftIO $ mnemonicToText @15 . entropyToMnemonic <$> genEntropy
             let pubKey = pubKeyFromMnemonics mnemonics
             wk <- restoreWalletFromPubKey @ApiWallet @'Shelley ctx pubKey restoredWalletName
 
@@ -182,12 +187,12 @@ spec = describe "SHELLEY_HW_WALLETS" $ do
             let payload = updatePassPayload fixturePassphrase "new-wallet-passphrase"
             rup <- request @ApiWallet ctx
                 (Link.putWalletPassphrase @'Shelley wk) Default payload
-            expectResponseCode @IO HTTP.status403 rup
+            expectResponseCode HTTP.status403 rup
             expectErrorMessage (errMsg403NoRootKey $ wk ^. walletId) rup
 
     describe "HW_WALLETS_04 - Can manage HW wallet the same way as others" $ do
-        it "Can update name" $ \ctx -> do
-            mnemonics <- mnemonicToText @15 . entropyToMnemonic <$> genEntropy
+        it "Can update name" $ \ctx -> runResourceT $ do
+            mnemonics <- liftIO $ mnemonicToText @15 . entropyToMnemonic <$> genEntropy
             let pubKey = pubKeyFromMnemonics mnemonics
             wk <- restoreWalletFromPubKey @ApiWallet @'Shelley ctx pubKey restoredWalletName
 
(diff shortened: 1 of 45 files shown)


Anviking:
bors try

Anviking:
bors try

rvl:
This is an excellent idea @Anviking - a general way to clean up resources allocated during test execution.

1. Just wondering, any reason for choosing the [resourcet](https://hackage.haskell.org/package/resourcet-1.2.4.2/docs/Control-Monad-Trans-Resource.html) library ahead of other possible options such as [managed](http://hackage.haskell.org/package/managed-1.0.8/docs/Control-Monad-Managed.html)?

2. Is there a general strategy for handling errors in the resource destroy function? If a test case throws an exception, and the destructor also throws an exception, which one is reported to the user?

3. Looking at this type which is repeated many times.
   ```haskell
   Context t -> ResourceT m a
   ```
   Since you have introduced a monad transformer, what about stacking a reader on top for the Context?

Anviking:
> Just wondering, any reason for choosing the resourcet library ahead of other possible options such as managed?

I went for resourcet because I think I heard Edsko mention it in Albi, and I didn't know about managed. Looking now, it seems to me like resourcet is more powerful (allows `unprotect`, some stuff for sharing resources across threads), which might come in handy one day.

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try-

Anviking:
Closing for now.

Anviking:
bors try

Anviking:
bors try

Anviking:
> Since you have introduced a monad transformer, what about stacking a reader on top for the Context?

I'd say, in another PR in that case. Maybe we could make it a concrete and opaque newtype, to avoid constraints?

Anviking:
> Is there a general strategy for handling errors in the resource destroy function? If a test case throws an exception, and the destructor also throws an exception, which one is reported to the user?

If I add `liftIO $ fail "foobar"` to `destroy` the tests fail with this:
```
uncaught exception: ResourceCleanupException
       ResourceCleanupException {rceOriginalException = Nothing, rceFirstCleanupException = user error (foobar), rceOtherCleanupExceptions = []}
```

Though in practice, `destroy` doesn't care about the http status of the delete request, so it shouldn't throw.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/HWWallets.hs
```diff
@@ -332,27 +336,29 @@ spec = describe "SHELLEY_HW_WALLETS" $ do
...
             let accXPub = pubKeyFromMnemonics mnemonics
-            _ <- restoreWalletFromPubKey @ApiWallet @'Shelley ctx accXPub restoredWalletName
+            r2' <- restoreWalletFromPubKey @ApiWallet @'Shelley ctx accXPub restoredWalletName
+
+            r1 <- request @ApiWallet ctx (Link.getWallet @'Shelley (getFromResponse id r1')) Default Empty
+            r2 <- request @ApiWallet ctx (Link.getWallet @'Shelley r2') Default Empty
 
             -- both wallets are available
-            rl <- request @[ApiWallet] ctx
-                (Link.listWallets @'Shelley) Default Empty
```
Anviking, line 360: Replaced this `listWallets` call with individually fetching the wallets once again.

IIRC, for some reason there were some wallets that hadn't already been cleaned up, that were messing up this test. And we need to adapt for when enabling `parallel` anyway.

lib/core-integration/src/Test/Integration/Scenario/CLI/Shelley/HWWallets.hs
```diff
@@ -303,46 +305,39 @@ spec = describe "SHELLEY_CLI_HW_WALLETS" $ do
...
-            length rl `shouldBe` 2
-            verify rl
-                [ expectCliListField 0
-                    (#name . #getApiT . #getWalletName)
-                    (`shouldBe` T.pack mnemonicWalName)
-                , expectCliListField 1
-                    (#name . #getApiT . #getWalletName)
-                    (`shouldBe` T.pack pubKeyWalName)
-                ]
+            wids `shouldContain` [mnemonicWal ^. walletId]
```
Anviking, line 346: Now checking that the wallet ids of the listed wallets _contain_ these two, instead of asserting that the first and second wallet in the list have the right names.

lib/core-integration/src/Test/Integration/Scenario/CLI/Shelley/Wallets.hs
```diff
@@ -307,12 +314,9 @@ spec = describe "SHELLEY_CLI_WALLETS" $ do
             out `shouldBe` ""
             T.unpack err `shouldContain` "Invalid number of words: 15, 18, 21\
                 \ or 24 words are expected."
-            (Stdout o, Stderr e) <- listWalletsViaCLI @t ctx
-            o `shouldBe` "[]\n"
-            e `shouldBe` cmdOk
```
Anviking, line 312: 🤔 I guess this is one of the more sub-optimal changes.

We _could_ calculate the corresponding wid, and make sure that no wallet with that wid is created. That would not catch if a wallet with an unexpected/different wid were created (like the wid calculation changing without the test being updated).

Best coverage would come from still running this test in isolation, and keeping the `== []` check.

lib/core-integration/src/Test/Integration/Scenario/CLI/Shelley/Wallets.hs
```diff
@@ -335,9 +339,6 @@ spec = describe "SHELLEY_CLI_WALLETS" $ do
             out `shouldBe` ""
             T.unpack err `shouldContain` "Invalid number of words: 9 or 12\
                 \ words are expected."
-            (Stdout o, Stderr e) <- listWalletsViaCLI @t ctx
-            o `shouldBe` "[]\n"
-            e `shouldBe` cmdOk
```
Anviking, line 340: Same as https://github.com/input-output-hk/cardano-wallet/pull/2191/files#r509169056

lib/core-integration/src/Test/Integration/Scenario/CLI/Shelley/Wallets.hs
```diff
@@ -456,18 +457,22 @@ spec = describe "SHELLEY_CLI_WALLETS" $ do
...
+        let name2 = prefix <> "#2"
+        w1 <- emptyWalletWith' ctx (name1, "secure-passphrase", 21)
+        _ <- emptyWalletWith' ctx (name2, "secure-passphrase", 21)
         (Exit c, Stdout out, Stderr err) <- listWalletsViaCLI @t ctx
         c `shouldBe` ExitSuccess
         err `shouldBe` cmdOk
-        j <- expectValidJSON (Proxy @[ApiWallet]) out
+        let walFromThisTest w =
+                prefix `T.isPrefixOf` getWalletName (getApiT (view #name w))
+        j <- filter walFromThisTest <$> expectValidJSON (Proxy @[ApiWallet]) out
```
Anviking, line 471: Same check as before, but we name the wallets in this test with a unique prefix, such that we can filter out the rest.

lib/core-integration/src/Test/Integration/Scenario/CLI/Shelley/Wallets.hs
```diff
@@ -480,14 +485,17 @@ spec = describe "SHELLEY_CLI_WALLETS" $ do
...
+        w1 <- emptyWalletWith' ctx (prefix <> "1", "secure-passphrase", 20)
+        w2 <- emptyWalletWith' ctx (prefix <> "2", "secure-passphrase", 20)
+        w3 <- emptyWalletWith' ctx (prefix <> "3", "secure-passphrase", 20)
         (Exit c, Stdout out, Stderr err) <- listWalletsViaCLI @t ctx
         c `shouldBe` ExitSuccess
         err `shouldBe` cmdOk
-        j <- expectValidJSON (Proxy @[ApiWallet]) out
+        let walFromThisTest w =
+                prefix `T.isPrefixOf` getWalletName (getApiT (view #name w))
+        j <- filter walFromThisTest <$> expectValidJSON (Proxy @[ApiWallet]) out
```
Anviking, line 498: Similar `filter walFromThisTest` approach

Anviking:
bors try

Anviking:
bors try

rvl (commented):
Wow, nice. Great results. I would say though that there needs to be at least _some_ tests running successfully in parallel for this to be worth merging. I assume the new failures are due to specs using a shared resource. Do you have any idea what that is? Stake pools perhaps?

I would also quite like this to stack a ReaderT on the ResourceT. Then we can make the `ctx` arg implicit everywhere. I think that will enhance readability by reducing boilerplate, if you follow through with a little bit of machinery. 

Check the `liftIO`s around verify and shouldBe, etc. I don't think they are needed if using hspec-expectations-lifted, right?

lib/core-integration/src/Test/Integration/Scenario/CLI/Byron/Wallets.hs
```diff
@@ -90,49 +88,50 @@ spec = describe "BYRON_CLI_WALLETS" $ do
...
-            jr <- expectValidJSON (Proxy @ApiByronWallet) out3
-            verify jr [ expectCliField walletId (`shouldBe` T.pack wid) ]
-            --re-create again? No!
-            (c4, out4, err4) <- createWalletViaCLI @t ctx
-                        args (unwords $ T.unpack <$> mnemonic)
-                        "\n" "secure-passphrase-restored-again"
-            c4 `shouldBe` ExitFailure 1
-            T.unpack err4 `shouldContain` (errMsg409WalletExists wid)
-            out4 `shouldBe` mempty
+            liftIO $ do
```
Anviking: Should de-indent

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
try cancel

Anviking:
bors try-

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
@rvl 
> Wow, nice. Great results. I would say though that there needs to be at least some tests running successfully in parallel for this to be worth merging. I assume the new failures are due to specs using a shared resource. Do you have any idea what that is? Stake pools perhaps?

👍 Enabled the full parallelisation. I think the tests failing due to shared resources were the trivial-to-fix ones. The tricker were race-conditions either created or worsened by the parallelisation. I hope it should be in a good-enough state now though.

> I would also quite like this to stack a ReaderT on the ResourceT. Then we can make the ctx arg implicit everywhere. I think that will enhance readability by reducing boilerplate, if you follow through with a little bit of machinery.

I'd still ultimately lean towards no, I think. Even if it just takes a few hours to implement, I'm not convinced it would be enough of an improvement. We'd lose ~1500 `ctx`, yes, but I think one could argue seeing them explicitly passed around is nice and un-magical, in a way. (Plus it might still be unexpectedly tricky. Did partly try on  `anviking/ADP-466/ctx-reader` though)

> Check the liftIOs around verify and shouldBe, etc. I don't think they are needed if using hspec-expectations-lifted, right?

Very good point, fixed 👍

Anviking:
bors try

Anviking:
bors try

Anviking:
hydra is "unable to fork"; cancelling and retrying
bors try-

Anviking:
bors try

Anviking:
bors try
(failure shouldn't be cached now)

Anviking:
bors try

nix/haskell.nix
```diff
@@ -92,6 +92,14 @@ let
 
           # cardano-node socket path becomes too long otherwise
           unit.preCheck = lib.optionalString stdenv.isDarwin "export TMPDIR=/tmp";
+
+          # Force more integration tests to run in parallel than the
+          # default number of build cores.
+          #
+          # TODO: Figure out why this doesn't work in hydra when it works in buildkite,
+          # and enable.
```
rvl, line 100: The `-j3` flag _should_ be added by this option when running `integration`.
Perhaps there are other things going on. Nix has two settings controlling parallelism - num cores for a build and num parallel builds - perhaps these are overriding our setting.
We can sort it later.

rvl (approved):
Excellent work.

Anviking:
_had accidentally applied the TRANS_EXTERNAL_02 fix to TRANS_EXTERNAL_01 instead of 02_

bors r+

Anviking:
bors r+
(yolo)

Issue #2192, opened 2020-09-30
Tweak `it` for more useful information on failures


Anviking:
# Context

ADP-419

We want to make CI failures easier to understand.

It happens that:
1. CI simply times out on a single test. Identifying _which_ test is non-trivial.
2. Our custom `it` (which automatically retries all tests once on failure), retries something that cannot be retried, and returns an unhelpful error message instead of the first helpful one. (Some tests use hard-coded mnemonics, which can only be spent once.)

# Decision

- Add a long timeout to our custom `it`, as a last resort to turn timeouts into failures. We don't want it interfering with long tests, so 10 minutes should probably work.
- If the retry-attempt also fails, return the _first_ error.

# Acceptance Criteria

- A test that is stuck in a weird place should cause a _test-failure_, not a global _CI timeout_.
- When a test fails, and the retry attempt fails, the original error should be presented.

---

# Development 

#2181 
#2193
#2194

# QA 

> - Add a long timeout to our custom it, as a last resort to turn timeouts into failures. We don't want it interfering with long tests, so 10 minutes should probably work.
> - If the retry-attempt also fails, return the first error.

This is done, and tested by
https://github.com/input-output-hk/cardano-wallet/blob/master/lib/test-utils/test/Test/Hspec/ExtraSpec.hs#L33-L64

rvl:
I was trying to be funny before when suggesting to prefix the function name with `sh`, but there is an element of truth in every joke...

If we persist with using this variation of `it`, we are potentially hiding real bugs in our application - the tricky type which only happen occasionally - and on users' systems not developers'.

So it should be used with caution. I think we should _try to_ put this function away, and only deploy it while we are suffering from test stability issues.

Anviking:
> and only deploy it while we are suffering from test stability issues.

😶 When is that not the case? 😞 

Otherwise I 100% agree.

piotr-iohk:
lgtm

PR #2193, opened 2020-09-30
Add tests for Extra.it

Anviking:
# Issue Number

#2192 

# Overview

- [x] I added a test suite for `cardano-wallet-test-utils` which could be useful
- [x] I added some basic tests for #2181

# Comments

PR #2194, opened 2020-09-30
Add 10 min timeout to `Test.Hspec.Extra.it`

Anviking:
# Issue Number

#2192 

# Overview

- [x] Add 10 min timeout to `it`.

# Comments

### DIFF
--- a/lib/test-utils/src/Test/Hspec/Extra.hs
+++ b/lib/test-utils/src/Test/Hspec/Extra.hs
@@ -13,16 +13,19 @@
 module Test.Hspec.Extra
     ( aroundAll
     , it
+    , itWithCustomTimeout
     ) where
 
 import Prelude
 
+import Control.Concurrent
+    ( threadDelay )
 import Control.Concurrent.Async
-    ( async, race, wait )
+    ( AsyncCancelled, async, race, wait )
 import Control.Concurrent.MVar
     ( MVar, newEmptyMVar, putMVar, takeMVar )
 import Control.Exception
-    ( SomeException, throwIO, try )
+    ( SomeException, catch, throwIO )
 import Test.Hspec
     ( ActionWith
     , HasCallStack
@@ -31,6 +34,7 @@ import Test.Hspec
     , afterAll
     , beforeAll
     , beforeWith
+    , expectationFailure
     , specify
     )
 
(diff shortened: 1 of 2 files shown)


lib/test-utils/src/Test/Hspec/Extra.hs
```diff
@@ -104,17 +113,38 @@ aroundAll acquire =
...
+                res2 <- try $ action ctx
+                case res2 of
+                    -- If the second try fails, return the first error. The
+                    -- second error might not be helpful.
+                    Left (e2 :: SomeException)
+                        | isAsyncCancelled e2 -> throwIO e1
+                        | otherwise           -> throwIO e1
+                    Right r2 -> return r2
+  where
+    isAsyncCancelled = isJust . fromException @AsyncCancelled
```
Anviking: Without this, the `race` cancel is ignored and the timeout doesn't work. Good that I tested it.

Anviking:
bors try

Anviking:
bors try

lib/test-utils/src/Test/Hspec/Extra.hs
```diff
@@ -103,18 +110,41 @@ aroundAll acquire =
...
+                | isAsyncCancelled e1 -> throwIO e1
+                | otherwise           -> do
+                res2 <- try $ action ctx
+                case res2 of
+                    -- If the second try fails, return the first error. The
+                    -- second error might not be helpful.
+                    Left (e2 :: SomeException)
+                        | isAsyncCancelled e2 -> throwIO e1
+                        | otherwise           -> throwIO e1
+                    Right r2 -> return r2
```
KtorZ: :grimacing: Arg. Wouldn't it be better to separate both functions ? A `timeout` combinator would in principle not need to know anything about the execution flow of a particular spec. May I suggest a simpler variant:

```hs
it :: HasCallStack => String -> ActionWith ctx -> SpecWith ctx
it title action = specify title $ \ctx -> timeout _10min $ do
    action ctx
        `catch` (\(_ :: AsyncCancelled) -> return ())
        `catch` (\(e :: SomeException)  -> action ctx
        `catch` (\(_ :: SomeException)  -> throwIO e))
  where
    _10min = 600 :: NominalDiffTime

timeout :: NominalDiffTime -> IO () -> IO ()
timeout t action =
    race (threadDelay (micro t)) action >>= \case
        Right () ->
            return ()
        Left () ->
            expectationFailure $  "timed out in " <> show t
  where
    micro :: NominalDiffTime -> Int
    micro = round . (* 1000) . (* 1000)
```

Anviking: 👍 nice.

KtorZ (approved):

Anviking:
@KtorZ may you also look at #2193 too, which this depends on? 🙏

PR #2199, opened 2020-10-01
Improve changelog generation script

Anviking:
# Issue Number

None / release.

# Overview

- [x] `- #PR | TITLE` -> `- TITLE #PR` Motivation:
    - Reading left to right the TITLE (or description) is the important thing and should come first
    - If we have multiple PR/Issue references in the same bullet, they are variable-length, and mess up the alignment of the titles.
    - Used in previous releases
- [x] Rename Github tag `IMPROVING CODE` to `IMPROVEMENT` (cf commit message) ⚠️ we need to rename the actual label in GitHub if merging this.
- [x] Clarify what to do with unclassified items.

# Comments

### DIFF
--- a/scripts/make_changelog
+++ b/scripts/make_changelog
@@ -23,7 +23,7 @@ PULL_REQUESTS=$(curl -X GET \
   https://api.github.com/search/issues?per_page=500\&q=repo:$REPO+is:pr+is:merged+merged:%3E$1)
 
 PULL_REQUESTS=$(echo $PULL_REQUESTS | jq '.items | map({number:.number,title:.title,label:.labels[0].name}) | group_by(.label)')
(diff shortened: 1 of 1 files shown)


scripts/make_changelog
```diff
@@ -31,7 +31,7 @@ for i in $(seq 0 $I); do
   LABEL=$(echo $LABELS | jq -r ".[$i]")
   case $LABEL in
   "null")
-      echo "## Unclassified"
+    echo "## Unclassified (Move user-facing items to correct section. Delete the rest)"
```
Anviking, line 34: We should also check that the items in the other section are user-facing. We could mention it in either
1. Release checklist 
2. As a HTML comment in the generated changelog

KtorZ (approved):

Anviking:
Renamed GitHub label.

Issue #2196, opened 2020-10-01
Flaky test - STAKE_POOLS_LIST_01 - List stake pools


Anviking:
# Context

https://github.com/input-output-hk/cardano-wallet/pull/2184#issuecomment-698850257 (build 10764)
https://github.com/input-output-hk/cardano-wallet/pull/2185#issuecomment-699868895
https://github.com/input-output-hk/cardano-wallet/pull/2107#issuecomment-700044743
https://github.com/input-output-hk/cardano-wallet/pull/2193#issuecomment-702491672

https://github.com/input-output-hk/cardano-wallet/pull/2198#issuecomment-702514443
https://github.com/input-output-hk/cardano-wallet/pull/2195#issuecomment-701538552
https://github.com/input-output-hk/cardano-wallet/pull/2198#issuecomment-702628123
 

# Test Case

https://github.com/input-output-hk/cardano-wallet/blob/647632e17f1a4add822ab76aef01ba6772a7b813/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs#L726-L735

# Failure / Counter-example

`(Reason "predicate failed on: 0")`

---

# Resolution 

---

Stake pool worker is crashing due to a History Interpreter failure. I think, current epoch check used to garbage collect pools.

But
1. Shouldn't the worker restart?
2. How can the stake pool chain follower be one epoch ahead of the tip-sync client updating the `TimeInterpreter`? In theory this should be possible, but it should be extremely rare, not every third CI run...

# QA

hasufell:
Have you tried increasing the eventually-timeout? Afair it's 90s.

Anviking:
> Have you tried increasing the eventually-timeout? Afair it's 90s.

The pools are set up at the beginning of the tests (probably at least 10 minutes earlier). By this time, they really should have produced blocks, and have saturation 🤔

Anviking:
Running locally I now got
```
Failures:

  src/Test/Integration/Scenario/API/Shelley/StakePools.hs:151:9:
  1) API Specifications, SHELLEY_STAKE_POOLS, STAKE_POOLS_JOIN_01 - Cannot join existent stakepool with wrong password
       uncaught exception: IOException of type UserError
       user error (Pattern match failure in do expression at src/Test/Integration/Scenario/API/Shelley/StakePools.hs:151:9-14)

[...]

PR #2203, opened 2020-10-02
Make pool garbage collection handle an unknown current epoch

Anviking:
# Issue Number

#2196 

# Overview

- [x] If the `currentEpoch` check fails, don't throw, just don't garbage collect.

# Comments

- I think this should be a quick fix for #2196. Still, would be nice to 1) figure out how this can happen, 2) use `Either` instead of `throwIO`.


Anviking:
bors try

KtorZ (approved):
Good catch.

PR #2204, opened 2020-10-04
Update cabal.project via latest stack2cabal

hasufell:
Use the verbatim feature to include the disabled
tests and benchmarks automatically.

Include ghc-options.

----

With the latest 1.0.10 stack2cabal release, this should allow us to drop all steps in https://github.com/input-output-hk/cardano-wallet/wiki/Building#syncing-stack-and-cabal-dependencies except the first!


.editorconfig
```diff
@@ -8,6 +8,9 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
+[stack.yaml]
+trim_trailing_whitespace = false
```
KtorZ, line 12: I do not understand the justification behind this change :thinking: ?

stack.yaml
```diff
@@ -79,3 +79,184 @@ nix:
   # Disabling the pure nix-shell allows environment variables to be
   # passed down to tests. We need this for integration tests.
   pure: false
+
+# !!! This can be dropped when cabal-3.4.0.0 is released !!!
```
KtorZ, line 83: Seems quite wrong to have the cabal config spreading like this :/ 
Couldn't this be figured out automatically by stack2cabal?

hasufell: How is this wrong? Otherwise you have to do it manually. This is a "bug" in cabal-install and a workaround has no place in stack2cabal. That's exactly what this verbatim feature is for.

.editorconfig
hasufell: It confuses the stack2cabal parser for verbatim messages and is annoying. Trailing whitespaces have many good use cases. Special pre-formatted code blocks are one of them.

cabal.project
```diff
@@ -335,6 +337,31 @@ source-repository-package
 allow-older: *
 allow-newer: *
```
rvl, line 338: Why are these needed?

hasufell: 1. Because ppl don't maintain .cabal files properly anymore since stack/nix/cabal-freeze so the version bounds are often wrong and lead to resolution failures
2. our `cardano-<ver>.yaml` snapshots often specify stuff that doesn't resolve and doesn't actually build, but isn't used by any package. E.g. input-output-hk/cardano-sl-x509 doesn't build with lts-14.25... try to add it to cardano-wallet-core.cabal and see it [blow up](https://gist.github.com/hasufell/17854ddfc9aa14897b6bb6e10069129b)... I fixed it [here](https://github.com/input-output-hk/cardano-sl-x509/commit/bf7e67c8af78d48453a12bb2ada85325fe7d26c9). However, cabal doesn't lazily evaluate the build plan like stack and doesn't ignore .cabal version bounds like stack.

So... you get all sorts of resolution errors, but when you ignore them, the build passes. And this matches stack behavior, as you can see.

`allow-older` is a bit more questionable and can indeed lead to issues with packages that are shipped with GHC, when you want a freeze file that works across many GHC versions (not the case for us).

hasufell:
merged in a different PR: https://github.com/input-output-hk/cardano-wallet/pull/2212

PR #2206, opened 2020-10-05
Add script for retrieving CI statistics

Anviking:
# Issue Number

None. Or I guess ADP-419. Or ADP-471.

# Overview

- [x] Added script that retrieves all recent bors-comments and prints them along with some simple statistics 

# Comments

Skärmavbild 2020-10-06 kl  19 09 09

### DIFF
--- a/scripts/bors-stats.sh
+++ b/scripts/bors-stats.sh
@@ -51,6 +51,7 @@ DATA=$(echo $QUERY \
           | select(.bodyText | contains("try") | not)
           | select(.bodyText | contains("Canceled") | not)
           | select(.bodyText | contains("Merge conflict") | not)
+          | select(.bodyText | contains("This PR was included in a batch that successfully built, but then failed to merge into master (it was a non-fast-forward update). It will be automatically retried.") | not)
           | select(.bodyText | contains("Already running a review") | not)
           | . + {succeded: (.bodyText | contains("Build succeeded"))}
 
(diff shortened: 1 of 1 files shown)


scripts/bors-stats.sh
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env bash
+
+# Script for analysing recent CI failures.
+
+# TODO: Add option to output raw JSON.
+# TODO: Add option to specify how many PRs to fetch
+# TODO: Parse attributes from comments to allow categorization
```
Anviking: I think one would want to use Haskell or Ruby for this. The `jq` usage below is already pushing it a bit.

Anviking: Just being able to have a `linkedIssues :: [Int]` attribute, parsed by `^#[\d]+` would be mostly what we need.

scripts/bors-stats.sh
```diff
@@ -0,0 +1,78 @@
...
+}
+}
+END)
+
+DATA=$(echo $QUERY \
+  | tr -d ' ' \
+  | tr -d '\n' \
+  | tr -d '\t' \
+  | jq -aRs '{query: .}' \
+  | http post https://api.github.com/graphql Authorization:"bearer $GITHUB_API_TOKEN" \
```
Anviking: Hm, I probably should use `curl`

scripts/bors-stats.sh
Anviking: It turned out to be pretty straight forward to extract in `jq` too, just using `(.bodyText | split("\n") | map (select(startswith("#")))`.

Anviking:
Closing for now

PR #2209, opened 2020-10-05
change derivation path JSON serialization to be less verbose

KtorZ:
# Issue Number

#2174 

# Overview

- [ ] I have changed  change derivation path JSON serialization to be less verbose and also aligned with other interfaces like cardano-addresses.

# Comments


KtorZ:
included in #2219

PR #2211, opened 2020-10-05
State-machine cache invalidation failure

KtorZ:
# Issue Number

#2210

# Overview

- 2be5813ecee41664fca2ebb955a160aba8a763fb
  :round_pushpin: **do not clear cache when we haven't rollback any blocks.**
    The state-machine tests found a counter example as such:

  - Put checkpoint at T1
  - Rollback to T1
  - Put checkpoint at T2, where T2 < T1
  - Get latest checkpoint

  At this last step, the model returned a checkpoint at `T1` whereas the
  database implementation returned the one at `T2` because the cache had
  been clear during the rollback from T1 to itself

# Comments

The main change lies in this function: 

```hs
    let invalidateCache :: W.WalletId -> SqlPersistT IO ()
        invalidateCache wid = do
            writeCache wid Nothing
            selectLatestCheckpoint wid >>= writeCache wid
```

Which now also immediately _replace_ the cache with whatever it needs to be right away when invalidating it.


Anviking (approved):
Ah, tricky!

PR #2213, opened 2020-10-05
ADP-455: return unsigned delegation certificate

### LINKED ISSUE #2200: Return unsigned delegation certificate
# Context

When joining a stake pool, the process is as follows:

1. Check if the account has already a registered stake key.
2. Check if the requested pool can be joined (it exists and it's not retiring).
3. Select coins to cover for either the delegation, or the delegation AND the key registration.
4. Sign the delegation (provide a witness for the stake key).
5. Submit the transaction.

The user wants to be able to:

1. Stop after step 3
2. Get the unsigned delegation certificate made by the wallet
3. Do step 4 and 5 on the client side

This ticket is about 1. and 2.

# Decision

# Acceptance Criteria

- The endpoint `POST /wallets/{walletId}/coin-selections/random` should be extended to accept a new request body, either `{ "action": "join", "pool": <pool id> }` or `{ "action": "quit" }`.
- Request bodies providing both "payments" and "action" should fail to parse with an appropriate error.
- Request bodies providing "pool" without "action" should fail to parse with an appropriate error.
- When provided with a poolId and an action the handler must verify that the action is valid and be consistent with the existing `POST /stake-pools/{poolId}/wallets/{walletId}` and `DELETE `/stake-pools/*/wallets/{walletId}` handlers (i.e. re-use the same logic).
- If valid, the request must be processed and return a list of resolved inputs, outputs, and a list of "certificates" depending on the scenario. Inputs and outputs should be already balanced to accommodate for the certificates.
- Certificates can be either: (1) a key deregistration certificate, (2) a key registration certificate AND a stake delegation certificate, (3) a stake delegation certificate.
- Certificate must be represented as JSON objects:
    - Key registration certificates should include a tag identifying the certificate, a bech32 serialized public stake key (according to CIP-5) and the derivation path corresponding to that key. (e.g. { "type": "key_registration", "stake_public_key": { "key": "stake_xvk1____", "derivation_path": "1852H/1815H/0H/2/0" } })
    - Key de-registration certificates should include a tag identifying the certificate, a bech32 serialized public stake key and the derivation path corresponding to that key.
    - Stake Pool delegation certificates should include a tag identifying the certificate and a bech32 serialized pool id.

# QA 

1. for the adjustment in the yaml schema check: https://github.com/input-output-hk/cardano-wallet/pull/2213/files#diff-d9433cbec16adf8a7dad1e1440a87cbee063b9fa9ebb7c6a631ce4a3482a4ed1
2. Also on API doc edge: https://input-output-hk.github.io/cardano-wallet/api/edge/#tag/Coin-Selections

Example requestes:

```
{
  "delegation_action": {
    "action": "join",
    "pool": "pool1wqaz0q0zhtxlgn0ewssevn2mrtm30fgh2g7hr7z9rj5856457mm"
  }
}
```
```
{  "delegation_action": {    "action": "quit"  } }
```

integration test scenarios here: https://github.com/input-output-hk/cardano-wallet/pull/2213/files#diff-883f5d29ba439fce1114ee5d9fdf3acf37f84150d24429f2f2ff768de0564d99R561

hasufell:
# Issue Number

#2200 

# Overview

- [x] Added Api types
- [x] Refactored some of the Transaction layer (still WIP)
- [x] Factored out certificate creation from `joinStakePool`
- [x] Added handler for joining stake pool
- [x] Add handler for quitting stake pool
- [x] serialization of certificates
- [x] lots of cleanup
- [x] fix jormungandr
- [x] fix tests
- [ ] add tests

# Comments


lib/core/src/Cardano/Wallet.hs
```diff
@@ -1954,10 +1956,9 @@ joinStakePool
     -> PoolId
     -> PoolLifeCycleStatus
     -> WalletId
-    -> ArgGenChange s
     -> Passphrase "raw"
-    -> ExceptT ErrJoinStakePool IO (Tx, TxMeta, UTCTime)
-joinStakePool ctx currentEpoch knownPools pid poolStatus wid argGenChange pwd =
+    -> ExceptT ErrJoinStakePool IO (CoinSelection, [Cardano.Certificate], DelegationAction)
```
KtorZ: In principle, we don't want to expose a type from `cardano-ledger-specs` or `cardano-api` in here. The code in Cardano.Wallet is supposed to be "backend-agnostic" and this `Cardano.Certificate` is specific to cardano-node (makes no sense for Jormungandr). 

Plus, isn't that redundant with the `DelegationAction` already ?

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1976,21 +1977,62 @@ joinStakePool ctx currentEpoch knownPools pid poolStatus wid argGenChange pwd =
...
 
-        selection <- withExceptT ErrJoinStakePoolSelectCoin $
+        cs <- withExceptT ErrJoinStakePoolSelectCoin $
             selectCoinsForDelegation @ctx @s @t @k ctx wid action
 
+        withRootKey @_ @s ctx wid pwd ErrJoinStakePoolRootKey $ \xprv scheme -> do
+            let pwdP = preparePassphrase scheme pwd
+            let rewardAcnt = getRawKey $ deriveRewardAccount @k pwdP xprv
+            let certs = (mkDelegationCertificates tl) (W.Join pid)
+                    (rewardAcnt, pwdP) cs
```
KtorZ: We should _not_ require the passphrase nor the private key for this. Actually, the main use-case for such function is when used for hardware devices integration in which case we do not have access to the root private key whatsoever. 

It is okay to enforce that `s ~ SeqState n k` here because it makes no sense anyway to call this for a `RndState` (random wallets) which have no delegation capabilities. From there, it is possible to access the `rewardAccountKey` corresponding to that particular account without any need to access the root private key.

lib/core/src/Cardano/Wallet/Transaction.hs
```diff
@@ -68,9 +72,21 @@ data TransactionLayer t k = TransactionLayer
         -- This expects as a first argument a mean to compute or lookup private
         -- key corresponding to a particular address.
 
+    , mkDelegationCertificates
+        :: W.DelegationAction
+            -- Pool Id to which we're planning to delegate
+        -> (XPrv, Passphrase "encryption")
+            -- Reward account
```
KtorZ: This should be passed as a public key, not an encrypted private key.

lib/core/src/Cardano/Wallet/Transaction.hs
```diff
@@ -68,9 +72,21 @@ data TransactionLayer t k = TransactionLayer
...
+        -> CoinSelection
```
KtorZ: Our current logic is a bit convoluted / backward mostly because of the Jormungandr support. In principle, we construct the coin selection from the certificates, not the other way around. We can indeed _infer_ what certificates were assumed for the coin-selection but that information is already solely available from `DelegationAction`. 

RegisterKeyAndJoin -> a stake key certificate + stake delegation certificate
Join -> a stake delegation certificate
Quit -> a stake key de-registration certificate

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -997,11 +1016,42 @@ instance DecodeAddress n => FromJSON (ApiSelectCoinsData n) where
...
+        quitAction obj = do
+          pass <- obj .: "passphrase"
+          ("quit" :: String) <- obj .: "action"
+          pure $ ApiSelectCoinsActionData Quit pass
+
+instance EncodeAddress n => ToJSON (ApiSelectCoinsActionData n) where
+    toJSON (ApiSelectCoinsActionData Quit pass)
+        = object [ "action" .= String "quit", "passphrase" .= pass ]
+    toJSON (ApiSelectCoinsActionData (Join poolid) pass)
+        = object [ "action" .= String "join", "pool" .= ApiT poolid, "passphrase" .= pass ]
```
KtorZ: These would be better derived using Generics. Hand-written JSON usually leads to inconsistencies and are hard to maintain on large APIs.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1712,9 +1764,15 @@ type instance ApiAddressIdT (n :: NetworkDiscriminant) =
...
     ApiCoinSelection n
 
+type instance ApiCoinSelectionCertT (n :: NetworkDiscriminant) =
+    ApiCoinSelectionCert n
+
 type instance ApiSelectCoinsDataT (n :: NetworkDiscriminant) =
     ApiSelectCoinsData n
 
+type instance ApiSelectCoinsActionDataT (n :: NetworkDiscriminant) =
+    ApiSelectCoinsActionData n
```
KtorZ: I don't think we need such type instances here. We need those for data-types which carry _addresses_, because of the type-level network discriminant. But if the underlying data-type do not contain any addresses, then this is superfluous / more complex than necessary.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -402,6 +408,19 @@ newtype ApiSelectCoinsData (n :: NetworkDiscriminant) = ApiSelectCoinsData
...
+    = ApiSelectCoinsActionData
+    { action :: W.DelegationAction
+    , passphrase :: ApiT (Passphrase "raw")
+    }
+    deriving (Eq, Generic, Show)
+
+data ApiCoinSelectionCert (n :: NetworkDiscriminant) = ApiCoinSelectionCert
+    { inputs :: !(NonEmpty (ApiCoinSelectionInput n))
+    , outputs :: !(NonEmpty (AddressAmount (ApiT Address, Proxy n)))
+    , certs :: ! [ApiT Certificate]
```
KtorZ: ~~certs~~ -> `certificates`, bytes are cheap in the API :)

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -402,6 +408,19 @@ newtype ApiSelectCoinsData (n :: NetworkDiscriminant) = ApiSelectCoinsData
...
+    } deriving (Eq, Generic, Show)
```
KtorZ, line 396: What about simply extending the current `ApiCoinSelection` data-type? The list of certificates can be empty for other delegation and we need not to define yet-another-resource.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -402,6 +408,19 @@ newtype ApiSelectCoinsData (n :: NetworkDiscriminant) = ApiSelectCoinsData
     { payments :: NonEmpty (AddressAmount (ApiT Address, Proxy n))
     } deriving (Eq, Generic, Show)
 
+data ApiSelectCoinsActionData (n :: NetworkDiscriminant)
+    = ApiSelectCoinsActionData
+    { action :: W.DelegationAction
+    , passphrase :: ApiT (Passphrase "raw")
```
KtorZ: As outlined in other comments, the `passphrase` should not be required.

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -304,6 +307,13 @@ type SelectCoins n = "wallets"
     :> ReqBody '[JSON] (ApiSelectCoinsDataT n)
     :> Post '[JSON] (ApiCoinSelectionT n)
 
+type SelectCoinsAction n = "wallets"
+    :> Capture "walletId" (ApiT WalletId)
+    :> "coin-selections"
+    :> "random"
+    :> ReqBody '[JSON] (ApiSelectCoinsActionDataT n)
+    :> Post '[JSON] (ApiCoinSelectionCertT n)
```
KtorZ: Interesting. Does `servant` treats `:<|>` as a type-level Alternative now? I am pretty sure in the past that only the right-hand side would be picked by servant when used like this (with identical "signatures" on endpoints). If that acts as a sort-of type-level alternative then that's quite amazing and there are a few other API endpoints we can simplify :thinking:

KtorZ (changes requested):

lib/core/src/Cardano/Wallet.hs
hasufell: How so? We want to serialise the certificates. So we have to know what types they are: https://input-output-hk.github.io/ouroboros-network/shelley-spec-ledger/Shelley-Spec-Ledger-Delegation-Certificates.html#t:DelegCert

lib/core/src/Cardano/Wallet/Api/Types.hs
hasufell: The representation is too different from the data IMO and I believe that's what json roundtrip tests are for.

lib/core/src/Cardano/Wallet/Api.hs
hasufell: https://hackage.haskell.org/package/servant-0.18/docs/Servant-API.html#t::-60--124--62-

lib/core/src/Cardano/Wallet/Api/Types.hs
hasufell: Isn't the point of types to have clear separation? For the old endpoint, `certs` will always be null. That should be expressed in the type system, so a potential client knows it doesn't have to deal with it.

lib/core/src/Cardano/Wallet/Api/Types.hs
hasufell: I am totally out of my depth on this one.

lib/core/src/Cardano/Wallet/Transaction.hs
hasufell: Yeah, I saw that and I was feeling a bit odd about it. Also because we have two `DelegationAction` types now... one just `Join | Quit` (for the API input) and the other what you wrote above.

lib/core/src/Cardano/Wallet.hs
hasufell: In addition, this endpoint won't be available for jormungandr anyway. How would we abstract over this?

KtorZ: But the `DelegationAction` are the certificates already. They capture the exact same concept.

lib/core/src/Cardano/Wallet/Transaction.hs
KtorZ: It'd be okay to use `ApiT DelegationAction` in the API, and have the JSON instance fails on parsing `RegisterKeyAndJoin` so that only `Join` and `Quit` are exposed in the JSON interface.

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: > For the old endpoint,

That's the point though, it is the _same_ endpoint, manipulating the same resource. Just a different request payload. It would indeed be always empty for normal payments, and that is just right because normal payments don't have any certificates. The introduction of certificates done by this PR is an extension of the existing resource, not a new definition.

lib/core/src/Cardano/Wallet.hs
hasufell: We also need `k ~ ShelleyKey` afais, for `getKey`.

KtorZ: You can use `WalletKey k` as a constraint and use `getRawKey`

lib/core/src/Cardano/Wallet.hs
hasufell: So do you mean we keep:

```hs
    _mkDelegationCertificates
        :: DelegationAction
            -- Pool Id to which we're planning to delegate
        -> XPub
            -- Reward account public key
        -> [Cardano.Certificate]
    _mkDelegationCertificates da accXPub =
        case da of
           Join poolId ->
                   [ toStakePoolDlgCert accXPub poolId ]
           RegisterKeyAndJoin poolId ->
                   [ toStakeKeyRegCert  accXPub
                   , toStakePoolDlgCert accXPub poolId
                   ]
           Quit -> [toStakeKeyDeregCert accXPub]
```

private and not in TransactionLayer and then use it when we serialize DelegationAction in the shelley API?

KtorZ: Yes. This is a shelley-specific detail so it shouldn't escape the shelley package.

hasufell: :godmode:

lib/core/src/Cardano/Wallet.hs
hasufell:
```
src/Cardano/Wallet/Jormungandr/Api/Server.hs:191:13: error:
    • Couldn't match type ‘JormungandrKey’
                     with ‘Cardano.Wallet.Primitive.AddressDerivation.Shelley.ShelleyKey’
        arising from a use of ‘joinStakePool’
    • In the first argument of ‘(:<|>)’, namely
        ‘joinStakePool
           jormungandr (knownStakePools spl) (getPoolLifeCycleStatus spl)’
      In the second argument of ‘(:<|>)’, namely
        ‘joinStakePool
           jormungandr (knownStakePools spl) (getPoolLifeCycleStatus spl)
           :<|> quitStakePool jormungandr :<|> delegationFee jormungandr’
      In the expression:
        (listPools spl)
          :<|>
            joinStakePool
              jormungandr (knownStakePools spl) (getPoolLifeCycleStatus spl)
              :<|> quitStakePool jormungandr :<|> delegationFee jormungandr
    |
191 |             joinStakePool jormungandr
```

hasufell: nvm, found it

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -313,6 +314,13 @@ type SelectCoins n = "wallets"
     :> ReqBody '[JSON] (ApiSelectCoinsDataT n)
     :> Post '[JSON] (ApiCoinSelectionT n)
 
+type SelectCoinsAction n = "wallets"
```
hasufell: I forgot if we want to keep this or extend `ApiSelectCoinsData` instead?

```hs
newtype ApiSelectCoinsData (n :: NetworkDiscriminant) = ApiSelectCoinsData
    { payments :: NonEmpty (AddressAmount (ApiT Address, Proxy n))
    , action :: Maybe DelegationAction
```

something like that?

hasufell:
bors try

hasufell:
This is ready for review, except for the missing integration tests... @piotr-iohk anything specific you want covered?

hasufell:
bors try

hasufell:
bors try

hasufell:
bors try

hasufell:
bors try

hasufell:
bors try

piotr-iohk:
> This is ready for review, except for the missing integration tests... @piotr-iohk anything specific you want covered?

@hasufell, I think as far as integration tests are concerned I can think of the following cases:
 - join existing SP
 - join existing SP that's retiring
 - join SP that retired
 - join a valid SP id that never existed

 - quit on a wallet that is delegating 
 - quit on a wallet that is NOT delegating

I am not sure, does the wallet delegation status matter? Perhaps it would be good to check for wallet that is delegating/not delegating/has delegation next etc.

Also I had some questions that I've put on the original ticket -> https://github.com/input-output-hk/cardano-wallet/issues/2200#issuecomment-707643381

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1688,47 +1692,18 @@ selectCoinsExternal ctx wid argGenChange payments withdrawal md = do
...
-        Nothing -> throwE $ err wid
-        Just xs -> pure xs
-
 data ErrSelectCoinsExternal e
     = ErrSelectCoinsExternalNoSuchWallet ErrNoSuchWallet
     | ErrSelectCoinsExternalUnableToMakeSelection (ErrSelectForPayment e)
-    | ErrSelectCoinsExternalUnableToAssignInputs WalletId
-    | ErrSelectCoinsExternalUnableToAssignOutputs WalletId
+    | ErrSelectCoinsExternalUnableToAssignInputs ErrNoSuchWallet
+    | ErrSelectCoinsExternalUnableToAssignOutputs ErrNoSuchWallet
```
KtorZ: I don't understand the rationale here? The `ErrNoSuchWallet` refers to an action which failed because there was no wallet associated with a given id. The `UnableToAssignInputs` and `UnableToAssignOutputs` errors supposes the existence of a wallet because they happened after a coin selection was found on a given wallet! 

Arguably, the `WalletId` in these errors brings very little information in the first place. It'd be more interesting to carry the actual malformed coin-selection in the error context to help debugging.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2494,6 +2601,39 @@ guardCoinSelection minUtxoValue cs@CoinSelection{outputs, change} = do
...
+fullyQualifiedInputs
+    :: forall s e.
+        (IsOurs s Address)
+    => s
+    -> CoinSelection
+    -> e
+    -> ExceptT
+        e
+        IO
+        (NonEmpty (TxIn, TxOut, NonEmpty DerivationIndex))
```
KtorZ: I don't think we need so many newlines here if everything can fit on one line :+1:

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2494,6 +2601,39 @@ guardCoinSelection minUtxoValue cs@CoinSelection{outputs, change} = do
...
+fullyQualifiedInputs s cs e =
+    traverse withDerivationPath (inputs cs) >>= flip ensureNonEmpty e
+  where
+    withDerivationPath
+        :: (TxIn, TxOut)
+        -> ExceptT
+            e
+            IO
+            (TxIn, TxOut, NonEmpty DerivationIndex)
```
KtorZ: Same here.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1972,32 +1947,58 @@ getTransaction ctx wid tid = db & \DBLayer{..} -> do
...
     -> W.EpochNo
     -> Set PoolId
     -> PoolId
     -> PoolLifeCycleStatus
     -> WalletId
-    -> ArgGenChange s
-    -> Passphrase "raw"
-    -> ExceptT ErrJoinStakePool IO (Tx, TxMeta, UTCTime)
-joinStakePool ctx currentEpoch knownPools pid poolStatus wid argGenChange pwd =
+    -> ExceptT ErrJoinStakePool IO (UnsignedTx (TxIn, TxOut, NonEmpty DerivationIndex), DelegationAction, [DerivationIndex])
```
KtorZ: We could possibly use an intermediate type `newtype RewardAccountPath = RewardAccountPath [DerivationIndex]` just for the sake of readability.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1972,32 +1947,58 @@ getTransaction ctx wid tid = db & \DBLayer{..} -> do
...
+        , s ~ SeqState n k
+        )
+    => ctx
+    -> W.EpochNo
+    -> Set PoolId
+    -> PoolId
+    -> PoolLifeCycleStatus
+    -> WalletId
+    -> ExceptT ErrJoinStakePool IO (CoinSelection, DelegationAction, [DerivationIndex])
+joinStakePoolUnsigned' ctx currentEpoch knownPools pid poolStatus wid =
```
KtorZ: Why is this split over two functions? The names of the function are not capturing quite well what they are for and the split seems rather arbitrary?

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2014,9 +2015,52 @@ joinStakePool ctx currentEpoch knownPools pid poolStatus wid argGenChange pwd =
...
+        cs <- withExceptT ErrJoinStakePoolSelectCoin $
             selectCoinsForDelegation @ctx @s @t @k ctx wid action
 
+        cp <- mapExceptT atomically
+            $ withExceptT ErrJoinStakePoolNoSuchWallet
+            $ withNoSuchWallet wid
+            $ readCheckpoint (PrimaryKey wid)
+        let s = getState cp
+            dprefix = Seq.derivationPrefix s
+            sPath = stakePath dprefix
```
KtorZ: We should avoid re-capturing a checkpoint that was already captured earlier in the function handler. By doing this, we may create inconsistency (because we can only ensure that the database is not changed by a concurrent thread within `atomically`). The wallet is already fetched earlier `readWalletMeta` and within the `joinStakePoolUnsigned` function.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2331,6 +2432,8 @@ data ErrStartTimeLaterThanEndTime = ErrStartTimeLaterThanEndTime
 data ErrSelectForDelegation
     = ErrSelectForDelegationNoSuchWallet ErrNoSuchWallet
     | ErrSelectForDelegationFee ErrAdjustForFee
+    | ErrSelectForDelegationUnableToAssignInputs ErrNoSuchWallet
+    | ErrSelectForDelegationUnableToAssignOutputs ErrNoSuchWallet
```
KtorZ, line 2268: Same as above, the `ErrNoSuchWallet` is definitely wrong in this case.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1778,13 +1824,27 @@ rndStateChange ctx (ApiT wid) pwd =
...
     -> ApiCoinSelection n
-mkApiCoinSelection (UnsignedTx inputs outputs) =
+mkApiCoinSelection mcerts (UnsignedTx inputs outputs) =
     ApiCoinSelection
         (mkApiCoinSelectionInput <$> inputs)
         (mkAddressAmount <$> outputs)
+        (fmap (uncurry mkCertificates) mcerts)
   where
+    mkCertificates :: DelegationAction -> [DerivationIndex] -> [Api.ApiCertificate]
+    mkCertificates action (s:sx) =
```
KtorZ: What about using a `NonEmpty DerivationIndex` from the start and carrying it around?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1778,13 +1824,27 @@ rndStateChange ctx (ApiT wid) pwd =
...
+        let apiStakePath = ApiT <$> (s:|sx)
+        in case action of
+            Join pid -> [Api.JoinPool apiStakePath (ApiT pid)]
+            RegisterKeyAndJoin pid ->
+                [ Api.RegisterRewardAccount apiStakePath
+                , Api.JoinPool apiStakePath (ApiT pid)
+                ]
+            Quit-> [Api.QuitPool apiStakePath]
+    mkCertificates _ _ = []
```
KtorZ: This is silently swallowing an error. We should never ever have an empty derivation path here. Having one is clearly a sign that something went really wrong in the internal logic and should be loudly visible / should crash the application.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2137,18 +2197,8 @@ instance Buildable e => LiftHandler (ErrSelectCoinsExternal e) where
...
-                , toText wid
-                ]
-        ErrSelectCoinsExternalUnableToAssignOutputs wid ->
-            apiError err500 UnexpectedError $ mconcat
-                [ "I was unable to assign outputs while generating a coin "
-                , "selection for the specified wallet: "
-                , toText wid
-                ]
+        ErrSelectCoinsExternalUnableToAssignInputs e -> handler e
+        ErrSelectCoinsExternalUnableToAssignOutputs e -> handler e
```
KtorZ: The previous error was actually correct.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -372,18 +378,49 @@ data ApiAddress (n :: NetworkDiscriminant) = ApiAddress
     , state :: !(ApiT AddressState)
     } deriving (Eq, Generic, Show)
 
+-- | Represents a relative address index.
+--
+-- The range of this type is exactly half that of a 'Word32'.
+--
+newtype ApiRelativeDerivationIndex = ApiRelativeDerivationIndex
+    { unApiRelativeDerivationIndex :: Word31
+    } deriving (Bounded, Enum, Eq, Generic, Show)
```
KtorZ: It seems to me that this should not be there? Wrong merge resolution?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -372,18 +378,49 @@ data ApiAddress (n :: NetworkDiscriminant) = ApiAddress
...
+
 data ApiEpochInfo = ApiEpochInfo
     { epochNumber :: !(ApiT EpochNo)
     , epochStartTime :: !UTCTime
     } deriving (Eq, Generic, Show)
 
-newtype ApiSelectCoinsData (n :: NetworkDiscriminant) = ApiSelectCoinsData
+data ApiSelectCoinsData (n :: NetworkDiscriminant)
+    = ApiSelectForPayment (ApiSelectCoinsPayments n)
+    | ApiSelectForAction ApiSelectCoinsAction
```
KtorZ: I think: 

ForAction -> ForDelegation 

would better capture the idea behind?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -372,18 +378,49 @@ data ApiAddress (n :: NetworkDiscriminant) = ApiAddress
...
+    deriving (Eq, Generic, Show)
+
+newtype ApiSelectCoinsPayments (n :: NetworkDiscriminant) = ApiSelectCoinsPayments
     { payments :: NonEmpty (AddressAmount (ApiT Address, Proxy n))
     } deriving (Eq, Generic, Show)
 
+newtype ApiSelectCoinsAction = ApiSelectCoinsAction
+    { delegation_action :: ApiT DelegationAction
```
KtorZ: `delegation_action` -> `delegationAction`, this should use camelCase.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -372,18 +378,49 @@ data ApiAddress (n :: NetworkDiscriminant) = ApiAddress
...
+    = RegisterRewardAccount
+        { reward_account_path :: NonEmpty (ApiT DerivationIndex)
+        }
+    | JoinPool
+        { reward_account_path :: NonEmpty (ApiT DerivationIndex)
+        , pool :: ApiT PoolId
+        }
+    | QuitPool
+        { reward_account_path :: NonEmpty (ApiT DerivationIndex)
+        }
```
KtorZ, line 408: Fields should also be camel-cased here :+1:

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -372,18 +378,49 @@ data ApiAddress (n :: NetworkDiscriminant) = ApiAddress
...
+    deriving (Eq, Generic, Show)
+
 data ApiCoinSelection (n :: NetworkDiscriminant) = ApiCoinSelection
     { inputs :: !(NonEmpty (ApiCoinSelectionInput n))
     , outputs :: !(NonEmpty (AddressAmount (ApiT Address, Proxy n)))
+    , certificates :: Maybe [ApiCertificate]
```
KtorZ: Wouldn't `Maybe (NonEmpty ApiCertificate)` be a slightly better type for this? There cannot be in principle a `Just []`. Either there are no certificates, or there are one or more.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -991,16 +1029,69 @@ instance FromJSON ApiEpochInfo where
...
+    parseJSON = genericParseJSON defaultRecordTypeOptions
+instance EncodeAddress n => ToJSON (ApiSelectCoinsPayments n) where
+    toJSON = genericToJSON defaultRecordTypeOptions
+
+instance DecodeAddress n => FromJSON (ApiSelectCoinsData n) where
+    parseJSON = withObject "DelegationAction" $ \o -> do
+        p <- o .:? "payments"
+        a <- o .:? "delegation_action"
+        case (p, a) of
+            (Just _, Just _) -> fail "Specified both payments and action, pick one"
```
KtorZ, line 1039: :+1:

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -991,16 +1029,69 @@ instance FromJSON ApiEpochInfo where
...
+apiCertificateOptions :: Aeson.Options
+apiCertificateOptions = Aeson.defaultOptions
+      { constructorTagModifier = camelTo2 '_'
+      , tagSingleConstructors = True
+      , fieldLabelModifier = camelTo2 '_' . dropWhile (== '_')
+      , omitNothingFields = True
+      , sumEncoding = TaggedObject
+          {
+            tagFieldName = "delegation_type"
+          , contentsFieldName = "contents"
```
KtorZ: The `delegation_` prefix seems wrong to me here. This is describing the type of "certificate", not delegation. There's no multiple types of delegation actually. What about:

```hs
{ tagFieldName = "type" -- or "certificate_type"
, contentsFieldName = "details"
}
```

 ?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -991,16 +1029,69 @@ instance FromJSON ApiEpochInfo where
...
+    toJSON = genericToJSON apiCertificateOptions
+
+instance FromJSON (ApiT DelegationAction) where
+    parseJSON = withObject "DelegationAction" $ \o ->
+        o .: "action" >>= \case
+            "join" -> do
+                pid <- o .: "pool"
+                pure (ApiT $ Join (getApiT pid))
+            "quit" -> pure $ ApiT Quit
+            val -> fail ("Unexpeced action value: " <> T.unpack val)
```
KtorZ: Do not spit back invalid values to users, instead, show what are expected values :+1:

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -991,16 +1029,69 @@ instance FromJSON ApiEpochInfo where
...
+
+instance ToJSON (ApiT DelegationAction) where
+    toJSON (ApiT (RegisterKeyAndJoin _)) = error "RegisterKeyAndJoin not valid"
```
KtorZ: That we won't have `RegisterKeyAndJoin` as a possible input in the API should not leak up here. May we not make this function partial and simply have a valid instance here :pray: 

The application logic should handle this properly. (or we should use a different type, more constrained that doesn't make it possible).

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -203,6 +211,21 @@ utxoInternal = toEnum $ fromEnum UTxOInternal
 mutableAccount :: Index 'Soft 'RoleK
 mutableAccount = toEnum $ fromEnum MutableAccount
 
+zeroAccount :: Index 'Soft 'AddressK
+zeroAccount = toEnum 0
+
+-- | Full path to the stake key. There's only one.
+stakePath :: DerivationPrefix -> [DerivationIndex]
+stakePath (DerivationPrefix (purpose, coin, acc)) =
```
KtorZ: neat-picking here, but what about: `stakeDerivationPath`

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -203,6 +211,21 @@ utxoInternal = toEnum $ fromEnum UTxOInternal
 mutableAccount :: Index 'Soft 'RoleK
 mutableAccount = toEnum $ fromEnum MutableAccount
 
+zeroAccount :: Index 'Soft 'AddressK
+zeroAccount = toEnum 0
```
KtorZ: It'd be better to rely on the `Bounded` instance here to avoid hard-coding magic numbers :+1:

lib/core/src/Cardano/Wallet/Transaction.hs
```diff
@@ -46,6 +47,8 @@ import Data.Text
 import Data.Word
     ( Word16, Word8 )
 
+import GHC.Generics
```
KtorZ, line 49: Avoid implicit imports, this ages badly and makes it hard to figure out where variable comes from for future reader. Prefer explicit imports :+1:

lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
```diff
@@ -847,12 +847,31 @@ instance Malformed (BodyParam (ApiSelectCoinsData ('Testnet pm))) where
...
+            , ("\"1020344\"", "Error in $: parsing DelegationAction failed, expected Object, but encountered String")
             , ("\"slot_number : \"random\"}", "trailing junk after valid JSON: endOfInput")
             , ("{\"payments : [], \"random\"}", msgJsonInvalid)
+            , ("join", "I couldn't understand the content of your message. If your message is intended to be in JSON format, please check that the JSON is valid.")
+            , ("quit", msgJsonInvalid)
+            ]
+         jsonValid = (first (BodyParam . Aeson.encode) <$> paymentCases) <> jsonValidAction
+         jsonValidAction = first (BodyParam . Aeson.encode) <$>
+            [ ( [aesonQQ| { "action": "join" }|]
+              , "Error in $: No valid parse"
```
KtorZ: The error is not quite satisfactory here, can we come up with something more meaningful which would help user getting to a resolution?

KtorZ (changes requested):

piotr-iohk:
One thing that is not quite correct is that `delegation_action` is available for byron_wallets random selection also (looking at the API doc). Should not be there :sweat_smile: :

Screenshot from 2020-10-13 14-43-55

Invoking it returns cryptic bad request...

```
{
  "code": "bad_request",
  "message": ""
}
```

lib/core/src/Cardano/Wallet.hs
hasufell: Honestly, I'm not too much of a fan of newtypes for readability. We can just use a type synonym or document it. newtypes don't give much type safety, unless they use smart constructors.

lib/core/src/Cardano/Wallet.hs
hasufell: Because they return different types, but basically do the same thing. I can't go back to CoinSelection from `UnsignedTx`.

piotr-iohk:
Something not right here... Although I have this wallet it tells me I don't have it :)

```
curl  -vX POST http://localhost:8090/v2/wallets/1b0aa24994b4181e79116c131510f2abf6cdaa4f/coin-selections/random \
  -d '{
"delegation_action": 
{
	"action": "join",
	"pool":"pool1numvf3nmw65vagz6z6z03l42v3c37nqq20lq8xtckgp67zzl7px"
} 
}' | jq
```
:point_down: 
```
{
  "code": "no_such_wallet",
  "message": "I couldn't find a wallet with the given id: 1b0aa24994b4181e79116c131510f2abf6cdaa4f"
}

```


Note, that it works for payments:

```
curl  -vX POST http://localhost:8090/v2/wallets/1b0aa24994b4181e79116c131510f2abf6cdaa4f/coin-selections/random \
  -d '{
"payments": [
{
	"address": "addr1qqlgm2dh3vpv07cjfcyuu6vhaqhf8998qcx6s8ucpkly6f8l0dw5r75vk42mv3ykq8vyjeaanvpytg79xqzymqy5acmqej0mk7",
	"amount": {"quantity": 1000000, "unit": "lovelace" }
} ]
}' | jq
```
:point_down: 
```
{
  "inputs": [
    {
      "amount": {
        "quantity": 1000000,
        "unit": "lovelace"
      },
...
```

lib/core/src/Cardano/Wallet.hs
hasufell: `readWalletMeta` doesn't return the checkpoint or the state. Can you be a bit more specific?

lib/core/src/Cardano/Wallet/Api/Server.hs
hasufell: That makes sense, since `stakePath` is a static non-empty list

lib/core/src/Cardano/Wallet/Api/Types.hs
hasufell: `contentsFieldName` isn't used

lib/core/src/Cardano/Wallet/Api/Types.hs
hasufell: > That we won't have RegisterKeyAndJoin as a possible input in the API should not leak up here.

Why not? This instance is for `ApiT DelegationAction`, not `DelegationAction`. So it's about our API.

lib/core/src/Cardano/Wallet/Transaction.hs
hasufell: You can use haskell-language-server to figure out where variables come from. This rule is now obsolete. But I'll comply regardless.

hasufell:
@piotr-iohk can you try with the latest commit? It should probably return a more descriptive error.

piotr-iohk:
> @piotr-iohk can you try with the latest commit? It should probably return a more descriptive error.

I get now:

```
{
  "code": "unable_to_assign_input_output",
  "message": "Unable to assign outputs from coin selection: inputs: - 1st 9811e75d (~ 972868630 @ 00bf8614...8094ee36) withdrawal: -0 reclaim: -0 outputs: [] change: [972695869] deposit: -0 "
}
```

but why? My testnet wallet has enough funds (2,7k ADA). The same is for `join` and `quit`. As far as I see it's the same for wallet that is already delegating or not delegating...

_for reference_:
```
# delegation-action quit
curl  -vX POST http://localhost:8090/v2/wallets/1ceb45b37a94c7022837b5ca14045f11a5927c65/coin-selections/random \
  -d '{
"delegation_action": 
{
	"action": "quit"
} 
}' | jq

# delegation-action join
curl  -vX POST http://localhost:8090/v2/wallets/1ceb45b37a94c7022837b5ca14045f11a5927c65/coin-selections/random \
  -d '{
"delegation_action": 
{
	"action": "join",
	"pool":"pool1numvf3nmw65vagz6z6z03l42v3c37nqq20lq8xtckgp67zzl7px"
} 
}' | jq
```

hasufell:
I get:

```
$ curl  -vX POST http://localhost:8090/v2/wallets/bee2d945e46e1b75c36c2157cdfdb980dc39f3d2/coin-selections/random \
    -d '{
  "delegation_action":
  {
  "action": "join", "pool": "pool1numvf3nmw65vagz6z6z03l42v3c37nqq20lq8xtckgp67zzl7px"
  }
  }' | jq
Note: Unnecessary use of -X or --request, POST is already inferred.
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying 127.0.0.1:8090...
* Connected to localhost (127.0.0.1) port 8090 (#0)
> POST /v2/wallets/bee2d945e46e1b75c36c2157cdfdb980dc39f3d2/coin-selections/random HTTP/1.1
> Host: localhost:8090
> User-Agent: curl/7.71.1
> Accept: */*
> Content-Type: application/json; charset=utf-8
> Content-Length: 115
>
} [115 bytes data]
* upload completely sent off: 115 out of 115 bytes
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Transfer-Encoding: chunked
< Date: Wed, 14 Oct 2020 10:12:56 GMT
< Server: Warp/3.3.5
< Content-Type: application/json;charset=utf-8
<
{ [777 bytes data]
100   879    0   764  100   115  30560   4600 --:--:-- --:--:-- --:--:-- 35160
{
  "inputs": [
    {
      "amount": {
        "quantity": 10000000000,
        "unit": "lovelace"
      },
      "address": "addr1qq9ulrlygk4ngxf8j3f4mvpzjhsp7s823ss7wp7vhwluygp4guhvvdgskp44lhpn0medz952f248a6tn8jp5xf0f0evq75rw9q",
      "id": "34f3abfc375be5a166e3aa5d52e846ed9a1fc0dd65c9016e6eaead17b6f002df",
      "derivation_path": [
        "1852H",
        "1815H",
        "0H",
        "0",
        "19"
      ],
      "index": 1
    }
  ],
  "certificates": [
    {
      "certificate_type": "register_reward_account",
      "reward_account_path": [
        "1852H",
        "1815H",
        "0H",
        "2",
        "0"
      ]
    },
    {
      "certificate_type": "join_pool",
      "pool": "pool1numvf3nmw65vagz6z6z03l42v3c37nqq20lq8xtckgp67zzl7px",
      "reward_account_path": [
        "1852H",
        "1815H",
        "0H",
        "2",
        "0"
      ]
    }
  ],
  "outputs": [
    {
      "amount": {
        "quantity": 9997825567,
        "unit": "lovelace"
      },
      "address": "addr1qp3msace6gl4df8a7x5dgxe7m7vealljffkph7lt693agl34guhvvdgskp44lhpn0medz952f248a6tn8jp5xf0f0evq4xdgnn"
    }
  ]
}
```

I'm not sure this output makes sense though. What are we expecting here @KtorZ ?

hasufell:
@piotr-iohk I added all integration tests and fixed the errors. But I'm still not 100% sure the output is what we want.

hasufell:
bors try

lib/core/src/Cardano/Wallet.hs
KtorZ: Instead of fetching _just_ the metadata with `readWalletMeta`, you could fetch the wallet state + metadata, but more importantly, do the coin selection and post-processing within the same `atomically`. Otherwise, there might be another thread interleaving changing the state of the wallet after you ran the coin selection. This is pretty minor here I recall because you're not writing anything back, but that's a potential source of subtle bug. 

Said differently, we should avoid:

```
atomically readCheckpoint
// stuff
atomically writeCheckpoint / readCheckpoint again
// stuff
```

but instead do

```
atomically $ do
   readCheckpoint
   // stuff
   writeCheckpoint 
   // stuff
```

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: Ah! What about indicating that in a comment then :blush: ?

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: We want to avoid partial functions when we can. There's no reason for this function to be partial. That the application logic (even only in the API) do not accept `RegisterKeyAndJoin` as a constructor should not influence the JSON instance itself. We are actually _abusing_ this type in the first place here in the application logic. The proper way to address this would be to have an actual type that captures only the allowed actions; because it's convenient we use the internal type `DelegationAction` wrapped in an `ApiT` but let's not shoot ourselves in the foot on top by making the JSON instances partial

specifications/api/swagger.yaml
```diff
@@ -2892,7 +2947,7 @@ paths:
         required: true
         content:
           application/json:
-            schema: *ApiSelectCoinsData
+            schema: *ApiByronSelectCoinsData
```
KtorZ, line 2954: :+1:

specifications/api/swagger.yaml
```diff
@@ -530,6 +530,41 @@ x-transactionOutputs: &transactionOutputs
...
+    - action
+  properties:
+    action:
+      type: string
+      enum: ["quit", "join"]
+    pool: *stakePoolId
+
+x-rewardAccountPath: &rewardAccountPath
+  type: array
+  minItems: 1
```
KtorZ: we could actually even go further and say:

```
minItems: 5
maxItems: 5
```

specifications/api/swagger.yaml
```diff
@@ -530,6 +530,41 @@ x-transactionOutputs: &transactionOutputs
...
+  items:
+    type: string
```
KtorZ, line 552: Can't this re-use the `derivationIndex` schema instead?

specifications/api/swagger.yaml
```diff
@@ -530,6 +530,41 @@ x-transactionOutputs: &transactionOutputs
       address: *addressId
       amount: *transactionAmount
 
+x-delegationAction: &delegationAction
+  description: A delegation action
```
KtorZ: Perhaps add a comment saying that `pool` is only required for `join` ?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -265,19 +282,17 @@ newTransactionLayer networkId = TransactionLayer
...
+                else mkDelegationCertificates (Join poolId) accXPub
 
         let mkWits unsigned =
                 [ mkShelleyWitness unsigned (accXPrv, pwd')
                 ]
 
         let payload = TxPayload Nothing certs mkWits
-        mkTx networkId payload tip acc keyFrom cs
+        let ttl = defaultTTL tip
+        mkTx networkId payload ttl acc keyFrom cs
```
KtorZ, line 281: Is this the result of a wrong merge-conflict resolution? I don't why this PR would have to change anything regarding transaction TTL?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -172,25 +174,40 @@ instance TxWitnessTagFor IcarusKey where
...
+        unsigned = mkUnsignedTx timeToLive cs md wdrls certs
+    in (unsigned, wdrls)
+
 mkTx
     :: forall k. (TxWitnessTagFor k, WalletKey k)
     => Cardano.NetworkId
     -> TxPayload Cardano.Shelley
     -> SlotNo
-    -- ^ Tip of chain, for calculating TTL
+    -- ^ Time to Live
```
KtorZ: Same here, this change looks _fishy_. I believe this is a wrong merge-conflict resolution?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -172,25 +174,40 @@ instance TxWitnessTagFor IcarusKey where
...
+    -- ^ Reward account
+    -> CoinSelection
+    -> (Cardano.TxBody Cardano.Shelley, [(Cardano.StakeAddress, Cardano.Lovelace)])
+mkTxUnsigned networkId certs md timeToLive rewardAcnt cs =
+    let wdrls = mkWithdrawals
+            networkId
+            (toChimericAccountRaw . toXPub $ rewardAcnt)
+            (withdrawal cs)
+        unsigned = mkUnsignedTx timeToLive cs md wdrls certs
+    in (unsigned, wdrls)
```
KtorZ: What does this function buys us exactly :thinking: ?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -109,6 +109,8 @@ import qualified Cardano.Api.Typed as Cardano
 import qualified Cardano.Chain.Common as Byron
 import qualified Cardano.Crypto as CC
 import qualified Cardano.Crypto.Hash.Class as Crypto
+import Cardano.Crypto.Wallet
+    ( XPub )
```
KtorZ: neat-picking: this import should not be there. We group imports as:

Prelude
Explicit imports
Qualified imports

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
```diff
@@ -285,10 +307,12 @@ server byron icarus shelley spl ntp =
...
-    byronCoinSelections :: Server (CoinSelections n)
-    byronCoinSelections wid x = withLegacyLayer wid
-        (byron, liftHandler $ throwE ErrNotASequentialWallet)
-        (icarus, selectCoins icarus (const $ paymentAddress @n) wid x)
+    byronCoinSelections :: Server (ByronCoinSelections n)
+    byronCoinSelections wid (ApiSelectForPayment x) =
+        withLegacyLayer wid
+            (byron, liftHandler $ throwE ErrNotASequentialWallet)
+            (icarus, selectCoins icarus (const $ paymentAddress @n) wid x)
+    byronCoinSelections _ _ = throwError err400
```
KtorZ: What about having a nicer error message here? Something that tells clients that there's no such thing as selecting coins for delegation on a byron wallet because such wallets have no delegation capabilities whatsoever.

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -1559,7 +1581,22 @@ instance ToSchema (ApiPutAddressesData t) where
...
+    declareNamedSchema _ = do
+        NamedSchema _ paymentData <- declareNamedSchema (Proxy @(ApiSelectCoinsPayments n))
+        NamedSchema _ actionData  <- declareNamedSchema (Proxy @ApiSelectCoinsAction)
+        pure $ NamedSchema Nothing $ mempty
+            & type_ .~ Just SwaggerObject
+            & required .~ []
+            & properties .~ mconcat
+                [ paymentData ^. properties
+                , actionData ^. properties
+                ]
```
KtorZ, line 1602: :+1:

lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
```diff
@@ -847,12 +847,31 @@ instance Malformed (BodyParam (ApiSelectCoinsData ('Testnet pm))) where
...
+            , ("\"1020344\"", "Error in $: parsing DelegationAction failed, expected Object, but encountered String")
             , ("\"slot_number : \"random\"}", "trailing junk after valid JSON: endOfInput")
             , ("{\"payments : [], \"random\"}", msgJsonInvalid)
+            , ("join", "I couldn't understand the content of your message. If your message is intended to be in JSON format, please check that the JSON is valid.")
+            , ("quit", msgJsonInvalid)
+            ]
+         jsonValid = (first (BodyParam . Aeson.encode) <$> paymentCases) <> jsonValidAction
+         jsonValidAction = first (BodyParam . Aeson.encode) <$>
+            [ ( [aesonQQ| { "action": "join" }|]
+              , "Error in $: No valid parse for ApiSelectCoinsAction or ApiSelectCoinsAction"
```
KtorZ: An interesting error. Sounds like a copy-paste that went wrong?

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -203,6 +213,21 @@ utxoInternal = toEnum $ fromEnum UTxOInternal
...
+stakePath :: DerivationPrefix -> NonEmpty DerivationIndex
+stakePath (DerivationPrefix (purpose, coin, acc)) =
+    (fromIndex purpose) :| [
+      fromIndex coin
+    , fromIndex acc
+    , fromIndex mutableAccount
+    , fromIndex zeroAccount]
+  where
+    fromIndex :: Index t l -> DerivationIndex
+    fromIndex (Index ix) = DerivationIndex ix
```
KtorZ: what about simply using `getIndex` that comes with the `Index` data-type ?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2483,6 +2550,14 @@ instance LiftHandler ErrQuitStakePool where
                     , "account! Make sure to withdraw your ", pretty rewards
                     , " lovelace first."
                     ]
+        ErrQuitStakePoolUnableToAssignInputs e ->
+            apiError err403 UnableToAssignInputOutput $ mconcat
+                [ "Unable to assign inputs from coin selection: "
+                , pretty e]
+        ErrQuitStakePoolUnableToAssignOutputs e ->
+            apiError err403 UnableToAssignInputOutput $ mconcat
+                [ "Unable to assign outputs from coin selection: "
```
KtorZ: How is that a client error? This is a server error and it should not happen under any circumstances.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2137,18 +2198,14 @@ instance Buildable e => LiftHandler (ErrSelectCoinsExternal e) where
...
-                , toText wid
-                ]
+        ErrSelectCoinsExternalUnableToAssignInputs e ->
+            apiError err403 UnableToAssignInputOutput $ mconcat
+                [ "Unable to assign inputs from coin selection: "
+                , pretty e]
+        ErrSelectCoinsExternalUnableToAssignOutputs e ->
+            apiError err403 UnableToAssignInputOutput $ mconcat
+                [ "Unable to assign outputs from coin selection: "
+                , pretty e]
```
KtorZ, line 2280: Two remarks here:

a) This is changing the behavior from a server error to a client error. We aren't necessarily against API breaking changes in principle but they should be backed with a rationale. I don't think this is right here, there's nothing that clients can do about this, getting in this situation is a programmer error meaning that we screwed up.

b) May we please keep the first-person tone to remain consistent with the rest of the server errors. It is much more friendly to get errors reported in the first-person :)

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1778,13 +1824,28 @@ rndStateChange ctx (ApiT wid) pwd =
...
+        -> NonEmpty DerivationIndex
+        -> NonEmpty Api.ApiCertificate
+    mkCertificates action xs =
+        let apiStakePath = ApiT <$> xs
+        in case action of
+            Join pid -> Api.JoinPool apiStakePath (ApiT pid) :| []
+            RegisterKeyAndJoin pid ->
+                Api.RegisterRewardAccount apiStakePath :|
+                    [Api.JoinPool apiStakePath (ApiT pid)]
+            Quit-> Api.QuitPool apiStakePath :| []
```
KtorZ: readability suggestion:

```hs
mkCertificates
    :: NonEmpty DerivationIndex
    -> DelegationAction
    -> NonEmpty Api.ApiCertificate
mkCertificates xs =
    \case
        Join pid -> NE.fromList
            [ Api.JoinPool apiStakePath (ApiT pid) 
            ]
  
        RegisterKeyAndJoin pid -> NE.fromList
            [ Api.RegisterRewardAccount apiStakePath 
            , Api.JoinPool apiStakePath (ApiT pid)
            ]
  
        Quit -> NE.fromList
            [ Api.QuitPool apiStakePath 
            ]
  where
    apiStakePath = ApiT <$> xs
```

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -552,6 +555,137 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
...
+
+            -- Join Pool
+            w <- fixtureWallet ctx
+            joinStakePoolUnsigned @n @'Shelley ctx w nonRetiringPoolId >>= \o -> do
+                verify o
+                    [ expectResponseCode HTTP.status200
+                    , expectField #inputs (`shouldSatisfy` (not . null))
+                    , expectField #outputs (`shouldSatisfy` (not . null))
+                    , expectField #certificates (`shouldSatisfy` (not . null))
+                    ]
```
KtorZ, line 593: We could go further in the assertion and make sure that there are 2 certificates, one about joining, and one about the key registration.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -552,6 +555,137 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
...
+                    [ expectField #delegation (`shouldBe` delegating pool [])
+                    ]
+
+            -- Quit Pool
+            quitStakePoolUnsigned @n @'Shelley ctx w >>= \o -> do
+                verify o
+                    [ expectResponseCode HTTP.status200
+                    , expectField #inputs (`shouldSatisfy` (not . null))
+                    , expectField #outputs (`shouldSatisfy` (not . null))
+                    , expectField #certificates (`shouldSatisfy` (not . null))
```
KtorZ: Same here, we could assert that the certificate is indeed of type "quit_pool" (and that there's no other certificates present)

lib/core/src/Cardano/Wallet.hs
hasufell: Does this fix it?
```hs
joinStakePoolUnsigned' ctx currentEpoch knownPools pid poolStatus wid =
    db & \DBLayer{..} -> do
        (wal, walMeta, _) <- withExceptT
            ErrJoinStakePoolNoSuchWallet (readWallet @ctx @s @k ctx wid)
        isKeyReg <- mapExceptT atomically
            $ withExceptT ErrJoinStakePoolNoSuchWallet
            $ isStakeKeyRegistered (PrimaryKey wid)

        let mRetirementEpoch = view #retirementEpoch <$>
                W.getPoolRetirementCertificate poolStatus
        let retirementInfo =
                PoolRetirementEpochInfo currentEpoch <$> mRetirementEpoch

        withExceptT ErrJoinStakePoolCannotJoin $ except $
            guardJoin knownPools (walMeta ^. #delegation) pid retirementInfo

        let action = if isKeyReg then Join pid else RegisterKeyAndJoin pid
        liftIO $ traceWith tr $ MsgIsStakeKeyRegistered isKeyReg

        cs <- withExceptT ErrJoinStakePoolSelectCoin $
            selectCoinsForDelegation @ctx @s @t @k ctx wid action

        let s = getState wal
            dprefix = Seq.derivationPrefix s
            sPath = stakeDerivationPath dprefix

        pure (cs, action, sPath)
```

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
hasufell: It's quite neat :smile: 

I believe the idea was originally to re-use it somewhere, but we don't...

specifications/api/swagger.yaml
hasufell: There is no such schema?

hasufell:
@KtorZ could you reply to https://github.com/input-output-hk/cardano-wallet/pull/2213#issuecomment-708305422 since we're still not sure the output is actually correct

specifications/api/swagger.yaml
KtorZ:
```
x-derivationSegment: &derivationSegment
  description: An individual segment within a derivation path.
  type: string
  example: 1852H

x-derivationPath: &derivationPath
  description: A path for deriving a child key from a parent key.
  type: array
  minItems: 1
  items: *derivationSegment
```

*derivationSegment / derivationPath.

KtorZ:
The output looks quite correct to me. For a first registration, we would expect two certificates, and they are qualified as expected.

piotr-iohk:
Minor thing:
Screenshot from 2020-10-16 13-41-53

But trying with hex pool_id we're getting:
```
curl  -vX POST http://localhost:8090/v2/wallets/1ceb45b37a94c7022837b5ca14045f11a5927c65/coin-selections/random \
  -d '{
"delegation_action": 
{
	"action": "join",
	"pool":"9f36c4c67b76a8cea05a1684f8feaa64711f4c0053fe039978b203af"
} 
}' | jq
```
:point_down: 
```
{ "code": "bad_request",
  "message":
         "Error in $['delegation_action'].pool: Invalid stake pool id: expecting a Bech32 encoded value with human readable part of 'pool'."}
```

hasufell:
> The output looks quite correct to me. For a first registration, we would expect two certificates, and they are qualified as expected.

I was more worried about the amounts in input and output...

KtorZ (approved):

KtorZ:
@hasufell there's one input selected, and a single change output. The delta is: `2 174 433`, which corresponds I believe to 2Ada of key deposit, and ~0.17 Ada for fees which seems right for the Testnet.

KtorZ:
bors merge

piotr-iohk:
Playing with the branch I observe that for the wallet that is **already delegating** to a pool, invoking a join action mostly results in:

```
{"code":"unable_to_assign_input_output"
,"message":"I'm unable to assign outputs from coin selection: inputs: - 1st eefa06df (~ 1000000 @ 00b69bf0...bca0a380) withdrawal: -0 reclaim: -0 outputs: [] change: [] deposit: -0 "}
```

Roughly 1 out of 5 is successful. The rest is this :point_up_2: .

For the case when wallet is not delegating... all seems to be OK.

cc: @hasufell @KtorZ

KtorZ:
@piotr-iohk on which commit did you do these tests :thinking: ?

piotr-iohk:
@KtorZ the last one from this branch. The same is actually now on `master`. -> 2020.10.13 (git revision: 05e59a04cc944cb0b49325b8ea72dfe0db78b98f)

E.g.: I'm seeing this on `testnet` for the wallet: `vacant cheap ghost elegant silly square offer vault announce turtle lesson doll key canyon romance` which is delegating to `pool1frevxe70aqw2ce58c0muyesnahl88nfjjsp25h85jwakzgd2g2l`.

```
curl  -vX POST http://localhost:8090/v2/wallets/1b0aa24994b4181e79116c131510f2abf6cdaa4f/coin-selections/random \
  -d '{
"delegation_action": 
{
	"action": "join",
	"pool":"pool1numvf3nmw65vagz6z6z03l42v3c37nqq20lq8xtckgp67zzl7px"
} 
}' | jq
```
:point_down:  
most of the time:

```
{
  "code": "unable_to_assign_input_output",
  "message": "I'm unable to assign outputs from coin selection: inputs: - 1st eefa06df (~ 1000000 @ 00b69bf0...bca0a380) withdrawal: -0 reclaim: -0 outputs: [] change: [] deposit: -0 "
}
```

sometimes:

```
{
  "inputs": [
    {
      "amount": {
        "quantity": 497657470,
        "unit": "lovelace"
      },
      "address": "addr1qp5aaw59aaa5z82dgvmpn7xryejcn9j00plsv32y4gw6veqhpgwhzykzha0z6w27gxsps40gurvxjt2vxh26009q5wqqmy4ryk",
      "id": "601dd956ade7e3aefdb2618889c05508ced1515711b5ed8da1a8cd495033004b",
      "derivation_path": [
        "1852H",
        "1815H",
        "0H",
        "1",
        "1"
      ],
      "index": 0
    }
  ],
  "certificates": [
    {
      "certificate_type": "join_pool",
      "pool": "pool1numvf3nmw65vagz6z6z03l42v3c37nqq20lq8xtckgp67zzl7px",
      "reward_account_path": [
        "1852H",
        "1815H",
        "0H",
        "2",
        "0"
      ]
    }
  ],
  "outputs": [
    {
      "amount": {
        "quantity": 497484709,
        "unit": "lovelace"
      },
      "address": "addr1qpdunk6q5uw5uwnvptttn7htzqms0amj5ztv98npdekn7aghpgwhzykzha0z6w27gxsps40gurvxjt2vxh26009q5wqq7rgr20"
    }
  ]
}
```

PR #2214, opened 2020-10-05
Allow flexible SMASH selection wrt ADP-427

### LINKED ISSUE #2163: Allow flexible SMASH selection without restarting the wallet
# Context

Currently the wallet needs to be restarted when selecting a SMASH server via cli:

```sh
wallet serve --smash-url URL
```

Daedalus developers would like to have an endpoint to change the smash server while the wallet is running.

# Decision

1 implement the following endpoint: `curl -X PUT --data '{ "settings": { "pool_metadata_source": "my.smash.uri" } }' http://localhost:8090/v2/settings/` ... possible values are `direct`, `none` and an actual url
2. The settings will be stored in the database in a separate table as a single row with one column per setting (as text or json).
3. The sync thread will poll the database for the setting.
4. On PUT, the metadata table will be dropped, forcing it to re-sync automatically with the new setting.
5. When selecting SMASH server on wallet start, compare against database setting and drop the metadata table if they diverge.

## Rationale

1. simple and extendable
2. Alternative would be ephemeral state, but that's more complicated implementation-wise and would require daedalus to have a persistent state about the smash setting.
3. Simplicity of implementation. Otherwise we need a way to restart the sync thread, which also requires mutable ephemeral state.
4. see AC
5. see AC

# Acceptance Criteria

1. The user can select the SMASH server at wallet start
2. The user can select the SMASH server after wallet start without restarting the wallet
3. Changing the SMASH server causes a re-sync of the metadata

---

# Development 

https://github.com/input-output-hk/cardano-wallet/pull/2156

# QA 

1. integration test for Settings change: https://github.com/input-output-hk/cardano-wallet/pull/2214/files#diff-ee9ceec3a14de991a6d049756e4830a4
2. json roundtrip test for SettingsPutData: https://github.com/input-output-hk/cardano-wallet/pull/2214/files#diff-aa8e2781f330251ae7904f42941b1de2
3. updated swagger definition: https://github.com/input-output-hk/cardano-wallet/pull/2214/files#diff-6816a7fbd0d7c7e715b8225b1709c612

hasufell:
Replaces https://github.com/input-output-hk/cardano-wallet/pull/2156

I think there are some weird issued with buildkite, because it's from my fork repo.


hasufell:
bors try

rvl:
Buildkite won't build branches from forks, as a matter of policy.

hasufell:
bors try

specifications/api/swagger.yaml
```diff
@@ -2940,3 +2996,29 @@ paths:
       parameters:
         - *parametersAddressId
       responses: *responsesInspectAddress
+
+  /settings:
+    put:
+      operationId: putSettings
+      tags: ["Settings"]
```
piotr-iohk, line 3004: The tag needs to be added also in one of the tag groups (https://github.com/input-output-hk/cardano-wallet/blob/master/specifications/api/swagger.yaml#L2177) otherwise it will be not visible in https://input-output-hk.github.io/cardano-wallet/api/edge/. 

I think `Miscellaneous` would be good group...

Maybe also the tag could be "Runtime Settings"?

specifications/api/swagger.yaml
```diff
@@ -2940,3 +2997,29 @@ paths:
...
+      summary: Settings
```
piotr-iohk: How about "Update" or "Set"? Otherwise it looks a bit strange:

Screenshot from 2020-10-06 15-26-37

specifications/api/swagger.yaml
```diff
@@ -2940,3 +2997,29 @@ paths:
...
+        required: true
+        content:
+          application/json:
+            schema: *ApiSettingsPutData
+      responses: *responsesPutSettings
+
+    get:
+      operationId: getSettings
+      tags: ["Settings"]
+      summary: Settings
```
piotr-iohk: How about "Get" or "Read"?

specifications/api/swagger.yaml
```diff
@@ -2940,3 +2997,29 @@ paths:
...
+
+  /settings:
+    put:
+      operationId: putSettings
+      tags: ["Settings"]
+      summary: Settings
+      description: |
+        <p align="right">status: <strong>stable</strong></p>
+
+        Overwrite current settings
```
piotr-iohk, line 3009: I think more detailed description would be handy explaining what are the possible values and behavior after updating the settings (like stated in https://github.com/input-output-hk/cardano-wallet/issues/2163 in "Decision" section, if it is still up to date.)

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
```diff
@@ -1263,3 +1263,22 @@ spec = describe "SHELLEY_WALLETS" $ do
                     , expectField (#tip . #slotNumber  . #getApiT) (`shouldBe` slotNum)
                     , expectField (#tip . #height) (`shouldBe` blockHeight)
                     ]
+
+    it "SETTINGS_01 - Can put and read settings" $ \ctx -> do
```
piotr-iohk: Maybe for better clarity it would be good to put test in a separate file? (For instance `Miscellaneous.hs`) or at least inside `StakePools.hs` to which it is more related with its current set of settings... :)

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1314,19 +1319,19 @@ tlsOption = TlsConfiguration
...
-    <> metavar "URL"
-    <> help "Optional HTTP(S) address of a \
-            \Stakepool Metadata Aggregation Server."
+    <> long "pool-metadata-fetching"
+    <> metavar "STRATEGY"
+    <> value FetchNone
+    <> showDefaultWith showT
+    <> help "Pool Metadata fetching strategy. This setting will persist across restarts. Possible values: \
+            \ none, \
+            \ direct, \
```
piotr-iohk, line 1328: What's the difference between "none" and "direct"? :sweat_smile:

lib/core/src/Cardano/Pool/DB.hs
```diff
@@ -191,6 +192,10 @@ data DBLayer m = forall stm. (MonadFail stm, MonadIO stm) => DBLayer
         -> stm ()
         -- ^ Store metadata fetched from a remote server.
 
+    , delPoolMetadata
+        :: stm ()
+        -- ^ Delete all pool metadata.
```
KtorZ, line 197: I think @rvl's already suggested to rename this from `delPool` to `removePoolMetadata` or `prunePoolMetadata` for consistency with the rest of the interface. I second this :+1:

lib/core/src/Cardano/Pool/DB/Model.hs
```diff
@@ -430,6 +444,16 @@ mListHeaders k
     | k > 0 = reverse . take k <$> get #blockHeaders
     | otherwise = reverse <$> get #blockHeaders
 
+mReadSettings
+    :: ModelOp Settings
+mReadSettings = get #settings
+
+{-# HLINT ignore mPutSettings "Use const" #-}
```
KtorZ, line 454: Why? (do not answer to me, but please add a comment in code explaining your rationale for disabling this hint).

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -1870,3 +1883,61 @@ invariant msg a predicate =
...
+    fromText "none" = Right FetchNone
+    fromText "direct" = Right FetchDirect
+    fromText uri = right FetchSMASH . fromText @SmashServer $ uri
+
+unsafeToPMS :: URI -> PoolMetadataSource
+unsafeToPMS = FetchSMASH . SmashServer
+
+-- | Wallet application settings. These are stored at runtime and
+-- potentially mutable.
+{-# HLINT ignore Settings "Use newtype instead of data" #-}
```
KtorZ: Same here, please add a justification for turning off this hint.

KtorZ (approved):
Please make sure to address the few remaining comments before triggering bors :+1:

lib/cli/src/Cardano/CLI.hs
piotr-iohk: I suppose: 
 - "none" - don't look for metadata at all
 - "direct" - look for metadata on chain
 - "uri"  - look in the SMASH server

I imagine "none" would be picked rather seldom. (Only case I can think is some maintenance/testing purposes or maybe exchanges running wallet and don't care about metadata)

Perhaps better default value would be "direct"? Which is also a current default...

specifications/api/swagger.yaml
hasufell: Since settings should be as general as possible, I don't know how to describe the behavior. That depends on the setting. We can only reasonably document it for the `pool_metadata_source`. That is already done in the appropriate places:

```yaml
x-poolMetadataSource: &poolMetadataSource
  description: |
    Pool metadata source. This sets the metadata fetching strategy.

    Possible values are
      * null -> no fetching
      * direct -> direct fetching
      * uri -> use SMASH server
  oneOf:
    - type: string
      enum:
        - none
        - direct
    - title: URI
      type: string
      format: uri
      example: https://smash.cardano-mainnet.iohk.io/

x-settings: &settings
  description: Settings
  type: object
  required:
    - pool_metadata_source
  properties:
    pool_metadata_source:
      <<: *poolMetadataSource
      description: How to fetch pool metadata
```

lib/cli/src/Cardano/CLI.hs
hasufell: `none` was decided as the default in the ticket (ADP-427):

> About (1), the command-line option should be removed altogether and the wallet should instead wait for a strategy to be provided. Without strategy, no metadata will be fetched whatsover.

piotr-iohk: As per: https://input-output-rnd.slack.com/archives/GBT05825V/p1602080513063300

If nothing is provided we should default to the latest setting in use.  If there was no previous setting in use, then it should default to "none".

Currently we always have "none" which erases any settings that were set in runtime.

specifications/api/swagger.yaml
piotr-iohk: Ok, perhaps updating to something like:

```
  properties:
    pool_metadata_source:
      <<: *poolMetadataSource
      description: |
        Select metadata fetching strategy:
          - `none` - metadata is not fetched at all,
          - `direct` - metadata is fetched directly from chain,
          - `uri` - metadata is fetched from the external Stake-Pool Metadata Aggregation Server (SMASH)

        After update existing metadata will be dropped forcing it to re-sync automatically with the new setting.
```

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1314,19 +1319,19 @@ tlsOption = TlsConfiguration
...
-    :: Parser URI
+    :: Parser PoolMetadataSource
 smashURLOption = option (eitherReader reader) $ mempty
-    <> long "smash-url"
-    <> metavar "URL"
-    <> help "Optional HTTP(S) address of a \
-            \Stakepool Metadata Aggregation Server."
+    <> long "pool-metadata-fetching"
+    <> metavar "STRATEGY"
+    <> value FetchNone
```
KtorZ: @piotr-iohk @hasufell 

I suspect the issue is here? There shouldn't be any default value here. Do you concur?

hasufell: yep

piotr-iohk: :tada:

hasufell:
> Build succeeded:

hell-yes-5bd835

Issue #2176, opened 2020-09-24
Provide a way to map addresses to their derivation paths.


jonathanknowles:
# Context

We need a way to map **wallet-owned addresses** to their **derivation paths**, so that they can be exported through the API using the `ApiAddressDerivationPath` type. (See https://github.com/input-output-hk/cardano-wallet/issues/2174)

# Decision

Extend the `IsOurs` interface to return the derivation path corresponding to an address identified as _ours_ (instead of only a Boolean). 

# Acceptance Criteria

1. `IsOurs` must now return a `Maybe` instead of a raw `Bool`
2. `Just [derivationPath]` must be returned when the address belongs to us. 
3. `Nothing` must be returned when the address does not belong to us. 

(the old behavior can be recovered by composing `isOurs` with `isJust`).

# Development 

- #2219 

# QA 

- Extended `IsOurs` interface, existing database tests and model tests still pass (the old behavior can be recovered by composing `isOurs` with `isJust`). 

- Added an entry for golden tests regarding ToJSON / FromJSON serializations:
  https://github.com/input-output-hk/cardano-wallet/blob/364c48021961ee2bd9df3141bf5dec334e922ec9/lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs#L314

- Reworked some existing "negative tests" regarding extremum values
  https://github.com/input-output-hk/cardano-wallet/blob/364c48021961ee2bd9df3141bf5dec334e922ec9/lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs#L457-L475

- Added two migration tests from an Icarus and Shelley databases (there's now a derivation prefix stored with each `SeqState`, this required doing a manual migration). The test show that (a) a previous database can be migrated and (b) after migration, the right prefix has been stored in the database. 
  https://github.com/input-output-hk/cardano-wallet/blob/364c48021961ee2bd9df3141bf5dec334e922ec9/lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs#L249-L263

- Added roundtrip text instances for the newly introduced `DerivationIndex`
  https://github.com/input-output-hk/cardano-wallet/blob/364c48021961ee2bd9df3141bf5dec334e922ec9/lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/SequentialSpec.hs#L148-L149

- Added an extensive comment about `DerivationIndex` 
[...]

PR #2219, opened 2020-10-06
Return derivation path when answering `isOurs`

KtorZ:
# Issue Number

#2176 

# Overview

- 6c3aaead33829e9b3b87d012fd50ccfc21d19aa9
  :round_pushpin: **change derivation path JSON serialization to be less verbose**
    And also aligned with other interfaces like cardano-addresses.

- 1bba794e45c40249da3483ee732a5fa6be260b03
  :round_pushpin: **change 'isOurs' to return a derivation path instead of a boolean**
    This can then be used to figure out what are the derivation path of
  a bunch of addresses when returning raw coin-selections.
  Note that this commit builds but is so-to-speak unsound. We need to
  find a way to feed the purpose, coin type and account index down to
  the 'isOurs' function. The most logical place to do this is as part of
  the state. We can't use arbitrary constant here because both Icarus
  and Shelley use a SeqState, but have different purpose indexes.

- 4b116de1d5a1f8d3c4f33bc12b6eaca1774a6deb
  :round_pushpin: **store seq-state derivation prefix in the database.**
    That prefixes tells us which account corresponds to which state and also, which purpose so can distinguish between Icarus and Shelley wallets. This will require a database migration
  which I'll add in a later commit.

- 185d8fb4ccf8c395614c7f8dd6ffcb1c34b09476
  :round_pushpin: **define manual migrations for seq-state with regards to the derivation prefix**
    I've generated databases for Icarus and Shelley wallets from the latest master and, in a test now trying to open these database and observe that a) it is possible, b) there's a log line indicating that a migration has happened, c) the resulting prefix in each database is exactly what we expect it to be.

# Comments


lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Jormungandr.hs
```diff
@@ -436,8 +410,17 @@ instance MkKeyFingerprint JormungandrKey (Proxy (n :: NetworkDiscriminant), Jorm
...
   where
+    type DerivationPath (SeqState n JormungandrKey) ChimericAccount =
+         DerivationPath (SeqState n JormungandrKey) Address
     isOurs account state =
-        (account == ourAccount, state)
+        let
+            purpose :: Index 'Hardened 'PurposeK
+            coinType :: Index 'Hardened 'CoinTypeK
+            accountIx :: Index 'Hardened 'AccountK
+            (purpose, coinType, accountIx) = undefined -- TODO: have this in the seq state
```
KtorZ: Whoops. Forgot to change that one. Will do.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Shelley.hs
```diff
@@ -350,8 +322,17 @@ instance MkKeyFingerprint ShelleyKey (Proxy (n :: NetworkDiscriminant), ShelleyK
...
   where
+    type DerivationPath (SeqState n ShelleyKey) ChimericAccount =
+         DerivationPath (SeqState n ShelleyKey) Address
     isOurs account state =
-        (account == ourAccount, state)
+        let
+            purpose :: Index 'Hardened 'PurposeK
+            coinType :: Index 'Hardened 'CoinTypeK
+            accountIx :: Index 'Hardened 'AccountK
+            (purpose, coinType, accountIx) = undefined -- TODO: have this in the seq state
```
KtorZ: Whoops. Forgot to change that one. Will do.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
```diff
@@ -551,6 +555,8 @@ data SeqState (n :: NetworkDiscriminant) k = SeqState
         -- (cf: 'PendingIxs')
     , rewardAccountKey :: k 'AddressK XPub
         -- ^ Reward account public key associated with this wallet
+    , derivationPrefix :: DerivationPrefix
+        -- ^ Derivation path prefix from a root key up to the internal account
```
KtorZ, line 567: There are two major changes in this PR (which then ripples to many other files). One of them is here. `DerivationPrefix` is defined below and is simply a 3-tuple of the first three derivation indexes that leads to the account associated with that state. 

This is necessary because we use seq-state for both Shelley and Icarus which have different `purpose` levels. Also, this will build a nice foundation for the up-coming multi-account support by having the actual account index within the state itself too!

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
```diff
@@ -632,17 +707,35 @@ instance
...
-            ours = isJust (internal <|> external)
+
+            ours = case (external, internal) of
+                (Just addrIx, _) ->
+                    Just (purpose, coinType, accountIx, utxoExternal, addrIx)
+
+                (_, Just addrIx) ->
+                    Just (purpose, coinType, accountIx, utxoInternal, addrIx)
+
+                _ -> Nothing
```
KtorZ, line 760: The other main change is here, in the definition of `isOurs`. Instead of returning a mere `Bool`, we now return the full derivation path corresponding to that address.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Random.hs
```diff
@@ -202,14 +203,16 @@ instance RndStateLike (RndState n) where
 -- decoded as a Byron HD random address, and where the wallet key can be used
 -- to decrypt the address derivation path.
 instance IsOurs (RndState n) Address where
+    type DerivationPath (RndState n) Address = DerivationPath
     isOurs addr st =
-        (isJust path, maybe id (addDiscoveredAddress addr Used) path st)
+        (path, maybe id (addDiscoveredAddress addr Used) path st)
```
KtorZ: For random addresses, this is trivially achieved because decrypting the path is exactly how we identify whether a random address is ours or not.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
```diff
@@ -568,13 +574,76 @@ instance
...
+
+instance FromText DerivationPrefix where
+    fromText txt =
+        DerivationPrefix <$> case T.splitOn "/" txt of
+            [purposeT, coinTypeT, accountT] -> (,,)
+                <$> fromText purposeT
+                <*> fromText coinTypeT
+                <*> fromText accountT
+            _ ->
+                Left $ TextDecodingError "expected exactly 3 derivation paths"
```
KtorZ, line 628: TODO: add roundtrip tests to test these instances.

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
```diff
@@ -227,11 +227,12 @@ UTxO                                sql=utxo
...
-    seqStateExternalGap     W.AddressPoolGap  sql=external_gap
-    seqStateInternalGap     W.AddressPoolGap  sql=internal_gap
-    seqStateAccountXPub     B8.ByteString     sql=account_xpub
-    seqStateRewardXPub      B8.ByteString     sql=reward_xpub
+    seqStateWalletId          W.WalletId         sql=wallet_id
+    seqStateExternalGap       W.AddressPoolGap   sql=external_gap
+    seqStateInternalGap       W.AddressPoolGap   sql=internal_gap
+    seqStateAccountXPub       B8.ByteString      sql=account_xpub
+    seqStateRewardXPub        B8.ByteString      sql=reward_xpub
+    seqStateDerivationPrefix  W.DerivationPrefix sql=derivation_prefix
```
KtorZ, line 240: TODO: This will require a database migration. Will add in a later commit.

paweljakubas (approved):
Very fond of renaming initiative taken here - `purposeCIP1852`, `coinTypeAda`,.. all give more clarity. There is **some more work** to do (roundtrips, defined undefined, db migration support..etc) already mentioned. I wonder if some things (like `purposeCI1852`) should not be (some day in future) relocated further to cardano-addresses and imported here in cardano-wallet

KtorZ:
>  I wonder if some things (like purposeCI1852)  should not be (some day in future) relocated further to cardano-addresses and imported here in cardano-wallet

That's a good idea long-term indeed :+1:

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
KtorZ: Done.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery.hs
```diff
@@ -55,10 +55,11 @@ import Cardano.Wallet.Primitive.Types
 -- same time, and this little abstraction can buy us this without introducing
 -- too much overhead.
 class IsOurs s entity where
+    type DerivationPath s entity :: *
     isOurs
         :: entity
         -> s
-        -> (Bool, s)
+        -> (Maybe (DerivationPath s entity), s)
```
hasufell: Comment below could be more descriptive:

```
         -- ^ Returns derivation path if the entity is ours, otherwise Nothing.
```

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -954,9 +960,40 @@ instance FromJSON ApiAddressDerivationPath where
     parseJSON = fmap ApiAddressDerivationPath . parseJSON
 
 instance ToJSON ApiAddressDerivationSegment where
-    toJSON = genericToJSON defaultRecordTypeOptions
+    toJSON (ApiAddressDerivationSegment (ApiRelativeAddressIndex ix) typ)
+        | typ == Hardened = toJSON (show ix <> "H")
```
hasufell: `show` makes me feel uneasy, since it's for debugging purposes mostly, not serialisation, although this use case seems safe.

There's a more elaborate way via Text: https://github.com/haskell/text/issues/218#issue-296457899
```hs
T.toStrict . B.toLazyText . B.decimal $ ix
```

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -954,9 +960,40 @@ instance FromJSON ApiAddressDerivationPath where
...
+        | otherwise = toJSON (show ix)
```
hasufell, line 951: same as above

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
hasufell, line 240: Maybe I'm being pedantic, but do we need to align all these rows? It increases the diff.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -148,7 +151,7 @@ import qualified Data.Text.Encoding as T
 --
 -- We do not manipulate purpose, cointype and change paths directly, so they are
 -- left out of the sum type.
-data Depth = RootK | AccountK | AddressK
+data Depth = RootK | PurposeK | CoinTypeK | AccountK | RoleK | AddressK
```
hasufell, line 151: It seems the documentation doesn't apply anymore.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
```diff
@@ -568,13 +574,76 @@ instance
...
-    build (SeqState intP extP chgs _) = "SeqState:\n"
+    build (SeqState intP extP chgs _ path) = "SeqState:\n"
+        <> indentF 4 ("Derivation prefix: " <> build (toText path))
         <> indentF 4 (build intP)
         <> indentF 4 (build extP)
         <> indentF 4 ("Change indexes: " <> indentF 4 chgsF)
       where
         chgsF = blockListF' "-" build (pendingIxsToList chgs)
 
+newtype DerivationPrefix = DerivationPrefix
```
hasufell, line 606: Maybe it's obvious what this is to others, but some documentation could still help?

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
KtorZ: Indeed!

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
KtorZ: We don't need to. It helps readability, but is arguably superfluous.

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: Yes, for Int / Word type, show is rather safe so I usually don't bother. The result is the same and quite more readable than more elaborated ways.

KtorZ: I think we could also just go for `toText` here since we have such instances for Word / Int I believe. Adding them otherwise would be nice.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
KtorZ: Definitely.

KtorZ: Added:

```hs
-- | Each 'SeqState' is like a bucket of addresses associated with an 'account'.
-- An 'account' corresponds to a subset of an HD tree as defined in BIP-0039.
--
-- cardano-wallet implements two similar HD schemes on top of BIP-0039 that are:
--
-- - BIP-0044 (for so-called Icarus wallets)
-- - CIP-1815 (for so-called Shelley and Jormungandr wallets)
--
-- Both scheme works by considering 5 levels of derivation from an initial root
-- key (see also 'Depth' from Cardano.Wallet.Primitive.AddressDerivation). A
-- SeqState keeps track of indexes from the two last levels of a derivation
-- branch. The 'DerivationPrefix' defines the first three indexes chosen for
-- this particular 'SeqState'.
```

KtorZ:
bors merge

KtorZ:
Damn.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -369,21 +375,21 @@ data ApiAddress (n :: NetworkDiscriminant) = ApiAddress
...
-newtype ApiAddressDerivationPath = ApiAddressDerivationPath
-    { unApiAddressDerivationPath :: NonEmpty ApiAddressDerivationSegment
+newtype ApiDerivationPath = ApiDerivationPath
+    { unApiDerivationPath :: NonEmpty ApiDerivationSegment
     } deriving (Eq, Generic, Show)
 
-data ApiAddressDerivationSegment = ApiAddressDerivationSegment
-    { derivationIndex :: !ApiRelativeAddressIndex
-    , derivationType :: !ApiAddressDerivationType
+data ApiDerivationSegment = ApiDerivationSegment
```
hasufell: I want to use this data type in the `Cardano.Wallet`, which must not use API types. Because this is similar to 

```hs
newtype Index (derivationType :: DerivationType) (level :: Depth) = Index
    { getIndex :: Word32 }
```
which I can't use as in a homogenous List easily. `ApiAddressDerivationSegment` would basically be non-typelevel version of `Index`.

hasufell: Or... we could define a toText instance for Index

KtorZ: What about using the `DerivationIndex` from the `Primitive.Types` instead ?

hasufell: That doesn't preserve the derivation type, don't we need that?

KtorZ: When considering all indexes as a list (or path):

- the depth matters not because what the `Depth` captures is actually the position of the index in that list. 
- the derivationType is captured by representing indexes as plain Word32. The `Soft` / `Hardened` notation is for easing human-readability but in the end, a soft index is simply a value < 2^31, whereas a "hardened" index is simply a value >= 2^31. Therefore, instead of representing indexes as derivationType + relative index within 0 and 2^31, we can represent them as just an index between 0 and 2^32, which is what `DerivationIndex` does.

KtorZ: I should probably add these bits as comments.

KtorZ:
bors merge

KtorZ:
bors retry

KtorZ:
bors retry

KtorZ:
bors merge

PR #2222, opened 2020-10-07
Add regen-cabal.sh

hasufell:
Since there seems to be still confusion on how to regenerate cabal project files, maybe this makes it easier.

* [x] Ran `ShellCheck`
* [x] tested on linux
* [x] tested on mac

### DIFF
--- a/scripts/regen-cabal.sh
+++ b/scripts/regen-cabal.sh
@@ -1,0 +1,335 @@
+#!/usr/bin/env bash
+
+# Make sure we error on subshell errors.
+# https://unix.stackexchange.com/a/48550
+# (more portable than 'shopt -s inherit_errexit')
+set -E
+trap '[ "$?" -ne 77 ] || exit 77' ERR
+
+
+    ###########################
+    #--[ Prelude functions ]--#
+    ###########################
+
+# @FUNCTION: error_message
+# @USAGE: <msg>
+# @DESCRIPTION:
+# Print a red message.
+error_message() {
+  (>&2 printf "\\033[0;31m%s\\033[0m\\n" "$1")
+}
+
+# @FUNCTION: warning_message
+# @USAGE: <msg>
+# @DESCRIPTION:
+# Print a yellow warning message.
+warning_message() {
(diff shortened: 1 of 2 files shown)


scripts/regen-cabal.sh
```diff
@@ -0,0 +1,72 @@
...
+
+SOURCE="$0"
+SCRIPT_DIR="$(dirname "$(realpath "${SOURCE}")")"
+
+cd "${SCRIPT_DIR}"/..
+
+mkdir -p .tmp/bin
+export PATH="$(realpath .tmp/bin):$PATH"
+
+vercomp() {
```
hasufell, line 66: This is a copy-paste... bash doesn't have libraries :)

scripts/regen-cabal.sh
```diff
@@ -0,0 +1,72 @@
...
+
+stack2cabal_ver="1.0.11"
+
+SOURCE="$0"
+SCRIPT_DIR="$(dirname "$(realpath "${SOURCE}")")"
+
+cd "${SCRIPT_DIR}"/..
+
+mkdir -p .tmp/bin
+export PATH="$(realpath .tmp/bin):$PATH"
```
hasufell: So we currently always prefer a local copy of the binary, if there is one (over a system one). I think that's fair.

rvl (commented):
The reason we have Nix is so that you don't need to write bash scripts like this.

hasufell:
> The reason we have Nix is so that you don't need to write bash scripts like this.

I've had more problems with nix than with bash scripts. And I don't want regenerating cabal files take up 6 GB of memory, require downloading half an OS and then deal with all sorts of esoteric errors.

But feel free to rewrite this in nix, because I'm using `stack2cabal` directly anyway. Otherwise I want to move forward with this, so ppl don't ask me every time on how to regenerate those files.

KtorZ:
Closing this one. We do not want to have to maintain yet-another-script, especially bash. What would be good would be to extend a bit the instructions [in the wiki](https://github.com/input-output-hk/cardano-wallet/wiki/Building#syncing-stack-and-cabal-dependencies) regarding the re-generation of cabal.project.

PR #2229, opened 2020-10-09
add verification key endpoint

paweljakubas:
# Issue Number

https://jira.iohk.io/browse/ADP-474

# Overview

- [x] I have added endpoint plus basic types and instances
- [x] I have extended swagger
- [x] I have updated core unit tests
- [x] I have added impl and switched on impl in Shelley
- [x] I have extended Api.Link and core-integration's DSL
- [x] I have added illustrative integration test
- [x] I have added golden tests for explicit mnemonic and picked indices based on cardano-addresses

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api.hs
+++ b/lib/core/src/Cardano/Wallet/Api.hs
@@ -132,7 +132,7 @@ import Cardano.Wallet.Api.Types
     , ApiTransactionT
     , ApiTxId
     , ApiUtxoStatistics
-    , ApiVerificationKeyHash
+    , ApiVerificationKey
     , ApiWallet
     , ApiWalletMigrationInfo
     , ApiWalletMigrationPostDataT
@@ -162,7 +162,6 @@ import Cardano.Wallet.Primitive.Types
     ( AddressState
     , Block
     , Coin (..)
-    , DerivationIndex
     , NetworkParameters
     , SortOrder (..)
     , WalletId (..)
@@ -305,9 +304,9 @@ type WalletKeys =
 type GetWalletKey = "wallets"
     :> Capture "walletId" (ApiT WalletId)
     :> "keys"
-    :> "script"
+    :> Capture "role" (ApiT AccountingStyle)
     :> Capture "index" (ApiT DerivationIndex)
-    :> Get '[JSON] ApiVerificationKeyHash
+    :> Get '[JSON] ApiVerificationKey
 
 {-------------------------------------------------------------------------------
                                   Addresses
--- a/lib/core/src/Cardano/Wallet/Api.hs
+++ b/lib/core/src/Cardano/Wallet/Api.hs
@@ -28,10 +28,10 @@ module Cardano.Wallet.Api
         , PutWallet
         , PutWalletPassphrase
         , GetUTxOsStatistics
-        , SignMetadata
 
     , WalletKeys
         , GetWalletKey
+        , SignMetadata
 
     , Addresses
         , ListAddresses
@@ -242,7 +242,6 @@ type Wallets =
     :<|> PutWallet
     :<|> PutWalletPassphrase
     :<|> GetUTxOsStatistics
-    :<|> SignMetadata
 
 -- | https://input-output-hk.github.io/cardano-wallet/api/#operation/deleteWallet
 type DeleteWallet = "wallets"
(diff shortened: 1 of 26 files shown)


lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -804,6 +810,19 @@ newtype MinWithdrawal = MinWithdrawal
 instance FromHttpApiData MinWithdrawal where
     parseUrlPiece = bimap (T.pack . getTextDecodingError) MinWithdrawal . fromText
 
+instance FromHttpApiData ApiRelativeAddressIndex where
+    parseUrlPiece = bimap (T.pack . getTextDecodingError) ApiRelativeAddressIndex . fromText
+
+instance FromText Word31 where
```
rvl: This could go into the text-class package I think, and perhaps it could be defined in terms of another typeclass like Integral.

rvl (commented):
Looking good

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: done, although now it is Word32. After the rebase many things need to change

paweljakubas:
bors try

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -273,6 +279,22 @@ type GetUTxOsStatistics = "wallets"
...
+-------------------------------------------------------------------------------}
+
+type WalletKeys =
+    GetWalletKey
+
+-- | https://input-output-hk.github.io/cardano-wallet/api/#operation/getWalletKey
+type GetWalletKey = "wallets"
+    :> Capture "walletId" (ApiT WalletId)
+    :> "keys"
+    :> "script"
```
KtorZ: What about capturing the full role here, so that we get better re-usability for this endpoint.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1101,6 +1108,26 @@ getUTxOsStatistics ctx (ApiT wid) = do
...
+    -> ApiT DerivationIndex
+    -> Handler ApiVerificationKeyHash
+getWalletKeyHash ctx (ApiT wid) (ApiT (DerivationIndex index)) =
+    withWorkerCtx @_ @s @k ctx wid liftE liftE $ \wrk -> do
+        (cp, _, _) <- liftHandler $ W.readWallet @_ @s @k wrk wid
+        let accPub = getState cp ^. #externalPool . #accountPubKey
+        let (KeyHash bytes) =
+                hashKey $ deriveMultisigPublicKey
+                (liftXPub $ Shelley.getKey accPub)
+                (toEnum $ fromInteger $ toInteger index)
```
KtorZ: That logic would better be located in `Cardano.Wallet.hs`. What about defining a function along those lines:

```hs
getWalletKeyHash 
    ::  ctx
    -> WalletId
    -> AccountingStyle
    -> Index 'Soft AddressK
    -> KeyHash
```

and re-use it in this layer?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -711,6 +712,10 @@ data ApiWalletMigrationInfo = ApiWalletMigrationInfo
 newtype ApiWithdrawRewards = ApiWithdrawRewards Bool
     deriving (Eq, Generic, Show)
 
+newtype ApiVerificationKeyHash = ApiVerificationKeyHash
+    { verificationKeyHash :: ApiT (Hash "ScriptKey")
+    } deriving (Eq, Generic, Show)
```
KtorZ, line 765: Alternatively, we could have the API return a plain bech32-encoded string on success?

lib/text-class/src/Data/Text/Class.hs
```diff
@@ -116,6 +118,19 @@ instance FromText Natural where
 instance ToText Natural where
     toText = T.pack . show
 
+instance FromText Word32 where
+    fromText =
+        validate <=< (fmap fromIntegral . fromText @Natural)
```
KtorZ, line 129: This will not work :scream:! The `fromIntegral` that transform from `Natural` to `Word32` needs to happen _after_ the validation to be safe. Otherwise, this may just throw if the value can't be converted back. 

Note that @hasufell is doing something similar but re-using constructor from the `text` package which makes things a bit nicer: https://github.com/input-output-hk/cardano-wallet/pull/2213/files#diff-b892068f732dfe3e67cd6f7a464975ba21a3588de3e5ea74022be5dfaa18e476

I'd suggest cherry-picking this change :+1:

stack.yaml
```diff
@@ -44,9 +44,9 @@ extra-deps:
     - persistent-sqlite
     - persistent-template
 
-# cardano-addresses-2.1.0
+# cardano-addresses
 - git: https://github.com/input-output-hk/cardano-addresses
-  commit: e0ab4587266430a08734e1aec1c29d261a9a3b70
+  commit: 02fb4ec7db7c55334ab2216c94cf70b209a33885
```
KtorZ: We need to use released version of cardano-addresses. There are some cleanup to be done in cardano-addresses before we can release a 3.0.0 version. Let's discuss that on Slack.

KtorZ (approved):

KtorZ:
bors merge

KtorZ:
bors merge

PR #2227, opened 2020-10-09
Fix mainnet network parameters

rvl:
### Issue Number

Relates to #2226.

### Overview

- Rearrange byron/shelley code a little.
- Updates hardcoded epoch length (slots not blocks) and active slot coefficient values for mainnet.
- Also updates slot length parameter.
- Also updates the _minFee a_ parameter - but I'm uncertain why the value was like that before.

Issue #2226, opened 2020-10-09
Incorrect mainnet network parameters returned from API


rvl:
### Jira tickets

ADP-483 - epoch length and slot length
ADP-554 - ActiveSlotCoeff parameter

### Context

Daedalus is showing this information about the current epoch for mainnet.
 
![screen_shot_2020-10-08_at_13 48 35](https://user-images.githubusercontent.com/1019641/95543662-8031e080-0a3c-11eb-94a2-cb5aeeb51b01.png)

Cardano-wallet hardcodes the genesis data, including protocol parameters for mainnet. The code was not updated once the shelley genesis data was finalised.

# Steps to Reproduce

1. Run `cardano-wallet serve --mainnet ...`
2. View the `/v2/network/parameters` API endpoint

## Expected behavior

- Epoch length is 432000 (slots).
- Active Slots Coeff is 5%

## Actual behavior
 
- Epoch length is 21600
- Active slots coeff is 100%

---

# Resolution 

### Attempt 1 - bad idea

Update genesis parameters in code, so that they match the shelley era genesis ([cardano deployments page](https://hydra.iohk.io/job/Cardano/iohk-nix/cardano-deployment/latest/download/1/index.html)) ⇒ #2227

### Attempt 2 

1. Stop using genesis data for parameters which aren't covered by protocol parameter updates, but can change.
2. Get epoch length (slots) and slot length (seconds) from HardFork History Interpreter ⇒ #2246.
3. Get _ActiveSlotCoeff_. This one is a bit tricky. We either need to load the shelley genesis (in addition to byron genesis), or find some way of querying the `Shelley.Spec.Ledger.BaseTypes.Globals` structure ⇒ There will be a ledger change to add querying of shelley genesis data.
 
### Can do this as well

In the network parameters endpoint, consider providing more information:
     - an expected number of blocks per epoch (epoch length × active slots coeff)
     - the current block height _within the epoch_.

### PRs

- #2227 - First atttempt
- #2236 - revert of incorrect fix
- #2246 - fix for epoch and slot length parameters.
- #2343 - test hardcoding of shelley mainnet ActiveSlotCoeff.

---

# QA 

1. I have done manual testing of `/network/parameters` and `/network/information` endpoints.
2. Automated testing - epoch length and slot length parameters added to SHELLEY_NETWORK integration test - `lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Network.hs`

rvl:
[...]

PR #2236, opened 2020-10-13
Fix network parameters for both byron and shelley eras

rvl:
### Issue Number

#2226 

### Overview

- [x] Revert genesis block changes from #2227


KtorZ (approved):

KtorZ:
@rvl moving ahead with that one so that we can proceed with today's release. It'll preserve the old behavior which is still slightly buggy, but at least not introduce a new class of bugs and address other points in forthcoming PRs.

KtorZ:
bors merge

Issue #2224, opened 2020-10-08
Flaky STAKE_POOLS_LIST_01 - List stake pools, has non-zero saturation & stake


Anviking:
# Context
 
https://github.com/input-output-hk/cardano-wallet/pull/2214#issuecomment-705172039
https://github.com/input-output-hk/cardano-wallet/pull/2208#issuecomment-705117369

# Test Case

https://github.com/input-output-hk/cardano-wallet/blob/02eecf34d5fee66929a764cf60c669849b0b10bf/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs#L660-L677

But in contrast to #2196, this is the only failing test.

# Failure / Counter-example

```
src/Test/Integration/Scenario/API/Shelley/StakePools.hs:659:9:
--
  | 1) API Specifications, SHELLEY_STAKE_POOLS, STAKE_POOLS_LIST_01 - List stake pools, has non-zero saturation & stake
  | uncaught exception: IOException of type UserError
  | user error (Waited longer than 90s (more than 2 epochs) for an action to resolve. Action: "list pools returns non-empty list". Error condition: Just user error (Quantity {getQuantity = Percentage {getPercentage = 0 % 1}} does not satisfy (> Quantity {getQuantity = Percentage {getPercentage = 0 % 1}})))

```

---

# Resolution 

---

# QA

PR #2237, opened 2020-10-13
Improve integration test failure reasons

Anviking:
# Issue Number

None / #2224 

# Overview

- [x] Add helpers
    - `appendFailureReason :: String -> HUnitFailure -> HUnitFailure`
    - `counterexample :: String -> IO a -> IO a`
- [x] Make `verify` show the full response if a check fails
- [x] Make `eventually` use `appendFailureReason` instead of re-throwing using `fail`, to keep the error nicely rendered
- [x] I guess we should replace usage of `fail` with `expectationFailure`? 🤔  (yes, we definitely should!)

# Comments

### Previously
Skärmavbild 2020-10-13 kl  11 07 01

### Now
Skärmavbild 2020-10-13 kl  11 01 34


lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -1436,8 +1440,36 @@ wantedErrorButSuccess =
     fail . ("expected an error but got a successful response: " <>) . show
 
 -- | Apply 'a' to all actions in sequence
-verify :: (Monad m) => a -> [a -> m ()] -> m ()
-verify a = mapM_ (a &)
+verify :: Show a => a -> [a -> IO ()] -> IO ()
+verify a = counterexample msg . mapM_ (a &)
+  where
+    msg = "While verifying " ++ show a
```
Anviking, line 1457: I guess we could also show the index of the check… unless the `SrcLoc` already leads to it 🤔

Anviking: Looking at the screenshots above, it seems like the source location _now_ does point to the exact item within the `verify` list.

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try-

Anviking:
bors try

Anviking:
bors try

piotr-iohk (approved):
LGTM.

paweljakubas (approved):
Nice! Improvements on more informative reporting when errors are always welcomed!

Anviking:
bors try

Issue #2238, opened 2020-10-13
Hard-coded mnemonics in integration tests cause flakiness


Anviking:
# Context

The test cluster has 3 pools, and forks.

The forks cause tests to fail (_exactly_ how, I don't have in my head, and could be checked closer)

Our solution has been to retry failing tests once with a custom `it`.

However, this doesn't work if the test cannot be retried, for instance due to the test depending on a hard-coded mnemonic with funds that can only be spend once.

# Test Case

This ticket is a grouping of the following failures:
#2233
#2230
#2232
#2207

and seem to cause 15% of total bors r+ to fail, or more.

# Failure / Counter-example

---

# Resolution 

We should make sure all tests can be retried.

---

# QA

Anviking:
Maybe it's not useful to create a super-issue 🤔

Anviking:
Haven't seen in a while. Think #2240 might have fixed it.

PR #2240, opened 2020-10-13
Add threadDelay after eventually conditions to protect against forks

Anviking:
# Issue Number

#2238 

# Overview

- [x] Add `threadDelay` after `eventually` checks in tests that cannot be retried. 

# Comments

I think this might work.

The cluster will roll back. For most tests, our retrying `it` protects
against rollback.

In these cases, hard-coded mnemonics make retrying impossible.

I believe by simply adding a threadDelay after the `eventually`, the relevant txs end up on the new chain.

### DIFF
--- a/lib/core-integration/src/Test/Integration/Framework/DSL.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/DSL.hs
@@ -114,6 +114,7 @@ module Test.Integration.Framework.DSL
     , rootPrvKeyFromMnemonics
     , unsafeGetTransactionTime
     , getTxId
+    , oneSecond
 
     -- * Delegation helpers
     , mkEpochInfo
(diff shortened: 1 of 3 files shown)


Anviking:
We _could_ also consider adding some kind of `paranoidEventually` that only returns after seeing the condition pass _twice_, with some delay in between.

By contrast, with this PR we assume that the condition still holds 10 seconds after it first held. (And also is more unlikely to break due to rollback)

Seemed overkill for now though.

Anviking:
bors try

Anviking:
bors try

paweljakubas (approved):
IMHO the fix is acceptable until we do something with fixed mnemonics tests

Anviking:
bors try

Anviking:
_not entirely sure it is succeeding more often than previously because of this PR, but... can at least not hurt..._
bors r+

PR #2241, opened 2020-10-14
Revise help text for pool metadata source CLI option

rvl:
### Issue Number

ADP-482

### Overview

Noticed while reviewing input-output-hk/cardano-launcher#90.

- Fix variable names to reflect the new option
- Add more detail to CLI help text

### DIFF
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -1323,11 +1323,12 @@ poolMetadataSourceOption
 poolMetadataSourceOption = option (eitherReader reader) $ mempty
     <> long "pool-metadata-fetching"
     <> metavar "( none | direct | SMASH-URL )"
-    <> help ("Stake pool metadata fetching strategy. "
+    <> help ("Sets the stake pool metadata fetching strategy. "
             <> "Provide a URL to specify a SMASH metadata proxy server, "
             <> "use \"direct\" to fetch directly from the registered pool URLs,"
-            <> " or \"none\" to completely disable stake pool metadata. "
-            <> "This setting will persist across restarts.")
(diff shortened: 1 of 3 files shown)


hasufell (approved):

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1318,15 +1318,16 @@ tlsOption = TlsConfiguration
...
-    <> help "Pool Metadata fetching strategy. This setting will persist across restarts. Possible values: \
-            \ none, \
-            \ direct, \
-            \ <SMASH-URL>."
+    <> metavar "( none | direct | SMASH-URL )"
+    <> help ("Stake pool metadata fetching strategy. "
+            <> "Provide a URL to specify a SMASH metadata proxy server, "
+            <> "use \"direct\" to fetch directly from the registered pool URLs,"
+            <> " or \"none\" to completely disable stake pool metadata. "
+            <> "This setting will persist across restarts.")
```
KtorZ: Perhaps: "This setting will persist across restarts AND override any previous setting set via this option or via the API." ?

KtorZ (approved):

lib/cli/src/Cardano/CLI.hs
hasufell: I'm starting to think it was a bad idea to keep the cli switch. It's confusing. Now someone might still think it **pins** the setting or something.

rvl: I think it's fine to have a sticky setting like this. Makes sense to me. Anyway, our users are likely to _either_ configure this by the CLI, _or_ by API, but not both.

Reworded a bit more.

KtorZ (approved):

KtorZ:
bors merge

PR #2243, opened 2020-10-14
Add {100,10,1}-wallet restore benchmarks

Anviking:
# Issue Number

ADP-469

# Overview

- [x] Auxiliary: Make benchmark actually respect passing in `--node-db`
- [x] Auxiliary: Add `--running-node SOCKET_PATH` for easier development
- [x] Benchmark restoring 100, 10 and 1 wallet to 1%
    - [x] Some tweaks to helpers to support this
- [x] Refactor to make the benchmark declarations more high-level and readable
- [x] Make sure the plotting of the results hasn't boken

# Comments

- Currently we won't a get plot for these new multi-wallet benchmarks, like for the existing ones. Could consider adding in another PR.
- Open to disagreement on only restoring 1% ([here](https://input-output-rnd.slack.com/archives/GBT05825V/p1602674480071800?thread_ts=1602142785.076200&cid=GBT05825V) is the old thread), but I think even 10% might be tricky to pull off in this PR. We'd need much more work for that.

### DIFF
--- a/lib/core-integration/src/Cardano/Wallet/BenchShared.hs
+++ b/lib/core-integration/src/Cardano/Wallet/BenchShared.hs
@@ -136,7 +136,7 @@ withNetworkConfiguration args action = do
     withNodeDir $ \dir -> action CardanoNodeConfig
         { nodeDir          = dir
         , nodeConfigFile   = networkDir </> "configuration.json"
-        , nodeDatabaseDir  = "db"
+        , nodeDatabaseDir  = fromMaybe "db" (argNodeDatabaseDir args)
         , nodeDlgCertFile  = Nothing
         , nodeSignKeyFile  = Nothing
         , nodeTopologyFile = networkDir </> "topology.json"
--- a/lib/core-integration/src/Cardano/Wallet/BenchShared.hs
+++ b/lib/core-integration/src/Cardano/Wallet/BenchShared.hs
@@ -60,6 +60,8 @@ import Data.Aeson
     ( ToJSON (..) )
 import Data.Functor
     ( (<&>) )
+import Data.Maybe
+    ( fromMaybe )
 import Data.Text
     ( Text )
 import Fmt
@@ -118,9 +120,13 @@ execBenchWithNode networkConfig action = do
     (_logCfg, tr) <- initBenchmarkLogging Info
     installSignalHandlers (return ())
 
-    void $ withNetworkConfiguration args $ \nodeConfig ->
-        withCardanoNode (trMessageText tr) nodeConfig $ \cp ->
-            action tr (networkConfig args) (nodeSocketFile cp)
+    case argUseAlreadyRunningNodeSocketPath args of
+        Just socket ->
+            action tr (networkConfig args) socket
+        Nothing -> do
+            void $ withNetworkConfiguration args $ \nodeConfig ->
+                withCardanoNode (trMessageText tr) nodeConfig $ \cp ->
+                    action tr (networkConfig args) (nodeSocketFile cp)
 
 withNetworkConfiguration :: RestoreBenchArgs -> (CardanoNodeConfig -> IO a) -> IO a
 withNetworkConfiguration args action = do
@@ -158,6 +164,7 @@ data RestoreBenchArgs = RestoreBenchArgs
     { argNetworkName :: String
     , argConfigsDir :: FilePath
     , argNodeDatabaseDir :: Maybe FilePath
+    , argUseAlreadyRunningNodeSocketPath :: Maybe FilePath
     } deriving (Show, Eq)
 
 restoreBenchArgsParser
@@ -184,6 +191,13 @@ restoreBenchArgsParser envNetwork envConfigsDir envNodeDatabaseDir = RestoreBenc
           <> envDefault "NODE_DB" envNodeDatabaseDir
           <> help "Directory to put cardano-node state. Defaults to $NODE_DB, \
               \falls back to temporary directory"))
+    <*> optional (strOption
+        ( long "running-node"
+          <> metavar "SOCKET"
+          <> envDefault "CARDANO_NODE_SOCKET" envNodeDatabaseDir
+          <> help "Path to the socket of an already running cardano-node. \
+                  \Also set by $CARDANO_NODE_SOCKET. If not set, cardano-node \
+              \will automatically be started."))
   where
     envDefault :: HasValue f => String -> Maybe a -> Mod f a
     envDefault name env = showDefaultWith (const ('$':name))
--- a/lib/core-integration/src/Cardano/Wallet/BenchShared.hs
+++ b/lib/core-integration/src/Cardano/Wallet/BenchShared.hs
@@ -33,7 +33,7 @@ import Cardano.BM.Data.Severity
 import Cardano.BM.Setup
     ( setupTrace_ )
 import Cardano.BM.Trace
-    ( Trace )
+    ( Trace, nullTracer )
 import Cardano.Launcher.Node
     ( CardanoNodeConfig (..)
     , CardanoNodeConn (..)
@@ -83,6 +83,7 @@ import Options.Applicative
     , showDefaultWith
     , strArgument
     , strOption
+    , switch
     , value
     )
 import Say
@@ -117,7 +118,8 @@ execBenchWithNode networkConfig action = do
     hSetBuffering stdout NoBuffering
     hSetBuffering stderr NoBuffering
 
-    (_logCfg, tr) <- initBenchmarkLogging Info
+    (_logCfg, tr') <- initBenchmarkLogging Info
+    let tr = if argQuiet args then nullTracer else tr'
     installSignalHandlers (return ())
 
     case argUseAlreadyRunningNodeSocketPath args of
@@ -165,6 +167,7 @@ data RestoreBenchArgs = RestoreBenchArgs
     , argConfigsDir :: FilePath
     , argNodeDatabaseDir :: Maybe FilePath
     , argUseAlreadyRunningNodeSocketPath :: Maybe FilePath
+    , argQuiet :: Bool
     } deriving (Show, Eq)
 
 restoreBenchArgsParser
@@ -198,6 +201,9 @@ restoreBenchArgsParser envNetwork envConfigsDir envNodeDatabaseDir = RestoreBenc
           <> help "Path to the socket of an already running cardano-node. \
                   \Also set by $CARDANO_NODE_SOCKET. If not set, cardano-node \
               \will automatically be started."))
+    <*> switch
+        ( long ("quiet")
+          <> help "Reduce unnecessary log output.")
   where
     envDefault :: HasValue f => String -> Maybe a -> Mod f a
     envDefault name env = showDefaultWith (const ('$':name))
--- a/lib/shelley/bench/Restore.hs
+++ b/lib/shelley/bench/Restore.hs
@@ -9,6 +9,7 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE NoMonoLocalBinds #-}
 {-# LANGUAGE Rank2Types #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
@@ -29,6 +30,16 @@
 -- @
 --
 -- since it relies on lots of configuration most most easily retrieved with nix.
+--
+-- You can also connect to an already-running node using:
+-- @
+--      stack bench cardano-wallet:bench:restore
+--          --ba 'mainnet -c $CONFIGURATION_DIR
+--          --running-node $SOCKET_PATH
+-- @
+--
+-- This makes iteration easy, but requires you to have the configuration
+-- directory layout setup correctly, and to know how to start a node.
 
 module Main where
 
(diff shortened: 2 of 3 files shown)


lib/shelley/bench/Restore.hs
```diff
@@ -711,7 +769,7 @@ waitForNodeSync tr nw = loop 10
 
 data BenchmarkLog (n :: NetworkDiscriminant)
     = MsgNodeTipTick BlockHeader SyncProgress
-    | MsgRestorationTick SyncProgress
+    | MsgRestorationTick POSIXTime [SyncProgress]
```
Anviking, line 761: Why did I add this? Might not be needed _yet_.

lib/shelley/bench/Restore.hs
```diff
@@ -654,31 +709,34 @@ prepareNode tr proxy socketPath np vData = do
 
 -- | Regularly poll the wallet to monitor it's syncing progress. Block until the
 -- wallet reaches 100%.
```
Anviking: Outdated "100%"

lib/shelley/bench/Restore.hs
```diff
@@ -460,7 +495,7 @@ benchmarksRnd _ w wid wname restoreTime = do
         ([], getState cp)
         [1..n]
       where
-        xprv = Byron.generateKeyFromSeed seed mempty
+        xprv = undefined --Byron.generateKeyFromSeed seed mempty
```
Anviking: 👀

lib/shelley/bench/Restore.hs
```diff
@@ -223,86 +225,70 @@ cardanoRestoreBench tr c socketFile = do
...
-         , bench_restoration @_ @ShelleyKey
-            networkProxy
-            (trMessageText tr)
-            socketFile
-            np
-            vData
-            (walletSeq "0.4-percent-seq" $ mkSeqAnyState' @4 networkProxy)
-            benchmarksSeq
+    runBenchmarks
+        [ benchRestoreMultipleWallets 1
```
Anviking: Takes 6s to reach 1%.

I suspect the extrapolated value (600s to 100%) is a bit wrong.

- Could increase the number of wallets for this test only
- Could leave it as is
- Could increase the restore-to target for this test only
- Could increase restore-to target for all {1,10,100} cases (but 100 would be really slow)

Anviking:
Triggered Buildkite nightly: https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/673#cc255463-69ec-4691-80b0-25042a53eced

lib/shelley/bench/Restore.hs
Anviking: Locally 1 wallet takes 276.4 s, vs 600s extrapolated, so pretty wrong, yes.

I should also note though, that when I compared 100 wallets to 1% extrapolated vs 100 wallets to 10%, they were very similar.

lib/core-integration/src/Cardano/Wallet/BenchShared.hs
```diff
@@ -118,9 +120,13 @@ execBenchWithNode networkConfig action = do
     (_logCfg, tr) <- initBenchmarkLogging Info
     installSignalHandlers (return ())
 
-    void $ withNetworkConfiguration args $ \nodeConfig ->
-        withCardanoNode (trMessageText tr) nodeConfig $ \cp ->
-            action tr (networkConfig args) (nodeSocketFile cp)
+    case argUseAlreadyRunningNodeSocketPath args of
```
paweljakubas, line 125: :+1:

lib/shelley/bench/Restore.hs
```diff
@@ -223,86 +222,70 @@ cardanoRestoreBench tr c socketFile = do
...
+    let benchRestoreRndWithOwnership permillProxy = bench_restoration
             networkProxy
             (trMessageText tr)
             socketFile
             np
             vData
-            (walletRnd "0.4-percent-rnd" $ mkRndAnyState @4)
+            [walletRnd (showPercent permillProxy <> "-percent-rnd")
+                $ mkRndAnyState' permillProxy networkProxy]
+            True -- Write progress to .timelog file
```
paweljakubas: Is this comment correct? The one in line 235 is opposite but the value is also True

lib/shelley/bench/Restore.hs
```diff
@@ -223,86 +222,70 @@ cardanoRestoreBench tr c socketFile = do
...
+    let benchRestoreSeqWithOwnership permillProxy = bench_restoration
             networkProxy
             (trMessageText tr)
             socketFile
             np
             vData
-            (walletSeq "0.2-percent-seq" $ mkSeqAnyState' @2 networkProxy)
+            [walletSeq (showPercent permillProxy <> "-percent-seq")
+                $ mkSeqAnyState' permillProxy networkProxy]
+            True -- Write progress to .timelog file
```
paweljakubas: the same here

lib/shelley/bench/Restore.hs
```diff
@@ -600,6 +610,15 @@ dummyAddress
     | otherwise =
         Address $ BS.pack $ 1 : replicate 56 0
 
+-- | Hash the given text, and construct a mnemonic from it.
+dummySeedFromName :: Text -> SomeMnemonic
+dummySeedFromName = SomeMnemonic @24
+    . entropyToMnemonic
+    . unsafeMkEntropy @256
+    . BS.take 256
```
paweljakubas: surprised here. to my best knowledge Blake2b 256 will create 32-byte bytestring. This line seems unnecessary

lib/shelley/bench/Restore.hs
paweljakubas: well, below you don't use it in toText, so probably superflouous ;-)

paweljakubas: But why not introduce `MsgRestorationTick [(POSIXTime,SyncProgress)]` to have this time for each sync progress? that would be useful probably

lib/shelley/bench/Restore.hs
```diff
@@ -9,6 +9,7 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE NoMonoLocalBinds #-}
```
paweljakubas, line 12: ok, will need to read about it :sweat_smile: At this moment do not understand why it is used

lib/shelley/bench/Restore.hs
paweljakubas: so what's executive summary of the above?

Anviking: Summary:
1. Should restore the 1 wallet to 10% for better extrapolation. But we can keep 10 and 100 wallets restoring to 1%, where it doesn't matter much.
2. There may be some interesting slowdowns (e.g. 10%, 16%), only visible when restoring to 100% (as Piotr also suggested), but I think we can ignore that in this PR.

Based on the data:
- To 1%
    - 1 wallets -> 4.015s (extrapolation is 54% higher than the to-10% value ❌)
    - 10 wallets -> 24.62s (extrapolation is 0.9% higher than the to-10% value 🆗 )
    - 100 wallets -> 345.0s (extrapolation is 0.8% higher than the to-10% value 🆗 )
- To 10%
    - 1 wallets -> 26.13s (extrapolation is 9.3% lower than the to-100% value 🤔 )
    - 10 wallets -> 244.0s (extrapolation is 14.1% lower than the to-100% value 🤔 )
    - 100 wallets -> 3423s (IIRC)
- To 100% (includes the Shelley part of the chain)
    - 1 wallets -> 288s
    - 10 wallets -> 2842s
    - 100 wallets -> 38578s

lib/shelley/bench/Restore.hs
Anviking: Thanks, the value was supposed to be `False` on line 235.

Could have added a new datatype to better represent this, but maybe I should more fundamentally refactor the `bench_restoration` function instead 🤔

lib/shelley/bench/Restore.hs
```diff
@@ -571,23 +563,41 @@ bench_restoration proxy tr socket np vData (wid, wname, s) benchmarks = do
...
                 Aeson.encodeFile resultsFilepath results
-                unsafeRunExceptT (W.deleteWallet w wid)
-                    $> SomeBenchmarkResults results
+                forM_ wallets $ \(wid, _, _) ->
+                    unsafeRunExceptT (W.deleteWallet w wid)
+                pure $ SomeBenchmarkResults ()
   where
-    basename = T.unpack (coerce wname)
+    fst' (x,_,_) = x
+    basename = "foo"
```
Anviking: 🤦‍♂️ Need to fix.

lib/shelley/bench/Restore.hs
Anviking: So we have the main `bench_results` function with a bunch of arguments, type parameters and constraints.

I wanted to define more descriptive helpers where I apply most of the boring arguments, e.g.:
```
    let benchRestoreRndWithOwnership permillProxy = bench_restoration
            networkProxy
            (trMessageText tr)
            socketFile
            np
            vData
            [walletRnd (showPercent permillProxy <> "-percent-rnd")
                $ mkRndAnyState' permillProxy networkProxy]
            True -- Write progress to .timelog file
            (unsafeMkPercentage 1)
            benchmarksRnd

```

But the per-mille ownership is a type parameter, and Haskell will fix it when inferring the type of the let binding. 
```
benchRestoreRndWithOwnership (Proxy @0) -- compiles
benchRestoreRndWithOwnership (Proxy @1) -- Expected type: Proxy 0; Actual type: Proxy 1
```

Moving it to a where-clause, and explicitly specifying `forall (p :: Nat)` didn't seem like a good solution because the full type signature is really big and cumbersome, and I wouldn't have access to the locally bound `vData`, `np`, `networkProxy`.

With `NoMonoLocalBinds` it just works.

Looking [here](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#let-generalisation) it seems that `MonoLocalBinds` is off by default, but enabled automatically by `GADTs`. Didn't know that.

> The extension MonoLocalBinds is implied by TypeFamilies and GADTs. You can switch it off again with NoMonoLocalBinds but type inference becomes less predictable if you do so. (Read the papers!)

lib/shelley/bench/Restore.hs
paweljakubas: Thanks!

lib/shelley/bench/Restore.hs
paweljakubas: Thanks!

paweljakubas (approved):
LGTM

lib/shelley/bench/Restore.hs
```diff
@@ -223,78 +224,67 @@ cardanoRestoreBench tr c socketFile = do
...
-            networkProxy
-            (trMessageText tr)
-            socketFile
-            np
-            vData
-            (walletSeq "0.4-percent-seq" $ mkSeqAnyState' @4 networkProxy)
-            benchmarksSeq
+        [ benchRestoreMultipleWallets 1 (unsafeMkPercentage 0.1)
+        , benchRestoreMultipleWallets 10 (unsafeMkPercentage 0.01)
+        , benchRestoreMultipleWallets 100 (unsafeMkPercentage 0.01)
```
KtorZ, line 290: Why not using the same percentage here? That will skew the benchmark a bit, wouldn't it?

Anviking: > Why not using the same percentage here?

To make it run in a few minutes rather than more than an hour.

See https://github.com/input-output-hk/cardano-wallet/pull/2243#discussion_r505444269

> That will skew the benchmark a bit, wouldn't it?

Well, only if you interpret in the wrong way.

KtorZ: I mean, why not: 

```
        [ benchRestoreMultipleWallets 1 (unsafeMkPercentage 0.01)
        , benchRestoreMultipleWallets 10 (unsafeMkPercentage 0.01)
        , benchRestoreMultipleWallets 100 (unsafeMkPercentage 0.01)
```

The first `0.1` looks like a typo to me ?

Anviking: Ah — the extrapolation from 0.01 to 0.1 is 54% off because it is so fast (imprecise start and stop times + ramping-up time, I take).

Extrapolating from 0.01 to 0.1 in the for 10 and 100 wallets saves an hour of time, but the accuracy of the extrapolation seem within 1%.

Anviking:
https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/680#ea0dfcad-4a8b-4461-99cc-a826b0a4eba1

```
BenchSeqResults:
  benchName: 1-wallets-to-0.1
  restoreTime: 40.50 s
  readWalletTime: 2.820 ms
  listAddressesTime: 156.5 μs
  listTransactionsTime: 552.8 μs
  estimateFeesTime: 9.631 s
  walletOverview:
     number of addresses: 20
     number of transactions: -0
     = Total value of -0 lovelace across 0 UTxOs

BenchSeqResults:
  benchName: 10-wallets-to-0.01
  restoreTime: 32.13 s
  readWalletTime: 37.23 ms
  listAddressesTime: 1.256 ms
  listTransactionsTime: 1.549 ms
  estimateFeesTime: 12.66 s
  walletOverview:
     number of addresses: 20
     number of transactions: -0
     = Total value of -0 lovelace across 0 UTxOs

BenchSeqResults:
  benchName: 100-wallets-to-0.01
  restoreTime: 551.2 s
  readWalletTime: 7.828 ms
  listAddressesTime: 99.44 μs
  listTransactionsTime: 665.3 μs
  estimateFeesTime: 19.78 s
  walletOverview:
     number of addresses: 20
     number of transactions: -0
     = Total value of -0 lovelace across 0 UTxOs
```

Anviking:
🤞 
bors r+

PR #2244, opened 2020-10-15
Add derivation paths to coin selection change outputs

jonathanknowles:
# Issue Number

#2247 
  
# Overview

This PR adjusts the API [`selectCoins`](https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/selectCoins) operation so that:

- _ordinary_ outputs and _change_ outputs are held in separate fields in the returned selection:

    | Field | Description |
    | -- | -- |
    | **`outputs`** | ordinary payments to external wallets |
    | **`change`** | change to be returned to the wallet |
- each change output now has a `derivation_path` field containing the full derivation path for the change output address.

  The format of this field is identical to the format of the existing `derivation_path` field of each input.

<details><summary>Click here to see updated API description</summary>

separate-change-ouputs

</details>

# Internal Changes

This PR introduces a `ApiCoinSelectionChange` type that includes a derivation path:

```hs
data ApiCoinSelectionChange (n :: NetworkDiscriminant) = ApiCoinSelectionChange
    { address :: !(ApiT Address, Proxy n)
    , amount :: !(Quantity "lovelace" Natural)
    , derivationPath :: NonEmpty (ApiT DerivationIndex)
```

In addition, we modify the `ApiCoinSelection` type so that ordinary outputs (**_without_** derivation paths) and change outputs (**_with_** derivation paths) are held in separate fields:

```patch

-   , outputs :: !(NonEmpty (AddressAmount (ApiT Address, Proxy n)))
+   , outputs :: !(NonEmpty (ApiCoinSelectionOutput n))
+   , change :: ![ApiCoinSelectionChange n]

```
# Testing

This PR updates the following integration tests:

- `WALLETS_COIN_SELECTION_01`.
- `WALLETS_COIN_SELECTION_02`.
- `SHELLEY_HW_WALLETS/HW_WALLETS_04`.
- `BYRON_HW_WALLETS/HW_WALLETS_04`.


jonathanknowles:
bors try

jonathanknowles:
bors try

paweljakubas:
Hi @jonathanknowles we will need to revert change to outputs as we in general do not know others derivation paths. We only can re-derive inputs and change. my apologies for wrong suggestion!

jonathanknowles:
bors try

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1624,26 +1633,34 @@ signTx
...
     => ctx
     -> WalletId
     -> Passphrase "raw"
     -> Maybe TxMetadata
-    -> UnsignedTx (TxIn, TxOut)
+    -- This function is currently only used in contexts where all change outputs
+    -- have been assigned with addresses and are included in the set of ordinary
+    -- outputs. We use the 'Void' type here to prevent callers from accidentally
+    -- passing change values into this function:
+    -> UnsignedTx input output Void
```
paweljakubas: for quite a while have not seen Void :-)

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1575,22 +1579,27 @@ migrateWallet ctx (ApiT wid) migrateData = do
 -- the number of addresses in the specified address list, addresses will be
 -- recycled in order of their appearance in the original list.
 assignMigrationAddresses
-    :: [Address]
+    :: forall input output change .
```
paweljakubas: I wonder why to introduce type parametrization and use type equality (very concrete like `output ~TxOut`) rather than stick to what we had?

paweljakubas: this is one place, but this technique was adopted in many places in this PR

lib/core/src/Cardano/Wallet.hs
jonathanknowles: Originally I wrote this by specifying `change ~ Void` in the list of constraints, and then specifying `UnsignedTx input output change` as the parameter type, as I felt that this would be more readable.

However, GHC views this as a redundant constraint. Of course, it's possible to silence the redundant constraint warning, but that would mean silencing this warning for the whole file. So I eventually switched to just writing `Void` inline, even though it is slightly less readable (IMO).

lib/core/src/Cardano/Wallet/Api/Server.hs
jonathanknowles: > I wonder why to introduce type parametrization and use type equality (very concrete like output ~TxOut) rather than stick to what we had?

It's just an attempt to improve readability.

Otherwise, we'd have to write things like:
```hs
UnsignedTx (TxIn, TxOut, NonEmpty DerivationIndex) TxOut (TxChange (NonEmpty DerivationIndex))
```
Which gets repetitive very quickly.

Giving short names to these types also means that we can reuse them within the signatures of inner functions. See `mkApiCoinSelection` for an example of this:

```hs
mkApiCoinSelection
    :: forall n input output change.
        ( input ~ (TxIn, TxOut, NonEmpty DerivationIndex)
        , output ~ TxOut
        , change ~ TxChange (NonEmpty DerivationIndex)
        )
    => UnsignedTx input output change
    -> ApiCoinSelection n
mkApiCoinSelection (UnsignedTx inputs outputs change) =
    ApiCoinSelection
        (mkApiCoinSelectionInput <$> inputs)
        (mkApiCoinSelectionOutput <$> outputs)
        (mkApiCoinSelectionChange <$> change)
  where
    mkApiCoinSelectionInput :: input -> ApiCoinSelectionInput n
    mkApiCoinSelectionInput (TxIn txid index, TxOut addr (Coin c), path) =
        ...
    mkApiCoinSelectionOutput :: output -> ApiCoinSelectionOutput n
    mkApiCoinSelectionOutput (TxOut addr (Coin c)) =
        ...
    mkApiCoinSelectionChange :: change -> ApiCoinSelectionChange n
    mkApiCoinSelectionChange txChange =
        ...
```

jonathanknowles:
bors try

paweljakubas (approved):
so far lgtm.having a way to derive change and inputs begs for some advanced testing sometime in future: we have a wallet with mnemonic1 with one utxo, and send some money to another one. We can now have a handle to output and change. We re-derive the addresses using derivation paths in cardano-addresses. And use them to create another transaction blob in cardano-transactions. Then send it using external api endpoint and see that it works in integration testing . Maybe @piotr-iohk would be interested in adding such a test sometime in the future:-)

jonathanknowles:
bors try

jonathanknowles:
bors try

PR #2245, opened 2020-10-15
cardano-node: 1.21.1 -> 1.22.1

rvl:
### Overview

- Use cardano-node 1.21.2 stack snapshot from input-output-hk/cardano-haskell#33
- Use cardano-node 1.22.1 stack snapshot from input-output-hk/cardano-haskell#34
- Multiple build and test fixes.
- Use a Cardano.Api.Shelley function for metadata type conversion.

### Comments

- Merge input-output-hk/cardano-haskell#34 first.
- Also depends on input-output-hk/cardano-transactions#24 - build fixes required in cardano-transactions package.


rvl:
bors try

rvl:
bors try

rvl:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -626,14 +628,20 @@ withBFTNode tr baseDir params action =
         createDirectoryIfMissing False dir
         source <- getShelleyTestDataPath
 
+        let copyKeyFile f = do
+                let dst = dir </> f
+                copyFile (source </> f) dst
+                restrictFileMode dst
+                pure dst
```
KtorZ, line 681: :+1:

KtorZ (approved):

KtorZ:
bors merge

rvl:
This is now updated to use the 1.22.1 release.

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -81,7 +81,7 @@ import Network.TypedProtocol.Pipelined
     ( N (..), Nat (..), natToInt )
 import Numeric.Natural
     ( Natural )
-import Ouroboros.Consensus.Ledger.Abstract
+import Ouroboros.Consensus.HardFork.Combinator.Ledger.Query
```
KtorZ, line 84: Seems like it changes location at every release, hide and seek 2.0 :upside_down_face:

lib/launcher/src/Cardano/Startup/POSIX.hs
```diff
@@ -29,3 +35,14 @@ installSignalHandlers notify = void $
...
+-- | Restricts the process umask so that any files created are only readable by
+-- their owner.
+setDefaultFilePermissions :: IO ()
+setDefaultFilePermissions = void $ setFileCreationMask mask
+    where mask = groupModes .|. otherModes
+
+-- | Changes permissions of an existing file so that only the owner can read
+-- them.
+restrictFileMode :: FilePath -> IO ()
+restrictFileMode f = setFileMode f ownerReadMode
```
KtorZ, line 48: :+1:

lib/shelley/cardano-wallet.cabal
```diff
@@ -43,6 +43,7 @@ library
     , cardano-crypto
     , cardano-crypto-class
     , cardano-crypto-wrapper
+    , cardano-wallet-launcher
```
KtorZ, line 46: That dependency seems to be in the wrong direction :thinking: ?

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -365,6 +378,10 @@ toCardanoBlockHeader gp = \case
         toByronBlockHeader gp blk
     BlockShelley blk ->
         toShelleyBlockHeader (W.getGenesisBlockHash gp) blk
+    BlockMary _ ->
+        error "TODO: unimplemented mary era"
+    BlockAllegra _ ->
+        error "TODO: unimplemented allegra era"
```
KtorZ, line 387: So it begins.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -609,7 +609,11 @@ codecVersion = verMap ! nodeToClientVersion
     where verMap = supportedNodeToClientVersions (Proxy @(CardanoBlock StandardCrypto))
 
 codecConfig :: W.SlottingParameters -> CodecConfig (CardanoBlock c)
-codecConfig sp = CardanoCodecConfig (byronCodecConfig sp) ShelleyCodecConfig
+codecConfig sp = CardanoCodecConfig
+    (byronCodecConfig sp)
+    ShelleyCodecConfig
+    ShelleyCodecConfig
+    ShelleyCodecConfig
```
KtorZ, line 615: Are those two additional other Codecs supposed to be for Mary and Allegra?

KtorZ (approved):

lib/core/src/Ouroboros/Network/Client/Wallet.hs
rvl: Yes. As far as I can tell, HFC is not yet covered by `Cardano.Api`.

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

rvl:
I rebased again on latest master, but please let me review the TimeInterpreter changes.
Which tests had horizon problems?

Anviking:
@rvl the tests corresponding to the places where `unsafeExtendSafeZone` was added, which is
1. `getTxExpiry` for setting TTL in the future
2. `mkApiTransaction` for showing TTL in the future
3. `mkShelleyWallet` for delegation status changes
4. `combineDbAndLsqData` for future pool retirement dates

lib/core/src/Cardano/Wallet.hs
```diff
@@ -718,7 +719,9 @@ walletSyncProgress ctx w = do
     (_,_,st) = ctx ^. genesisData
 
     ti :: TimeInterpreter IO
-    ti = timeInterpreter (ctx ^. networkLayer @t)
+    ti = neverFails
```
Anviking, line 725: This always felt a bit wrong — to motivate the `neverFails` _outside_ the `syncProgress` function, with `syncProgress` internals as arguments — but I wasn't sure how to fix.

We could make `syncProgress` take a `TimeInterpreter ExcepT PastHorizonException IO` and return `IO SyncProgress`, but then we couldn't use it with `TimeInterpreter Identity` …

_Maybe this still can sort-of work out though... if we hoist the `TimeInterpreter Identity` into `TimeInterpreter PastHorizonException Identity` when needed on the call site_

rvl:
I rebased, squashed a couple commits, fixed some ghc warnings.

Anviking (approved):

lib/core/src/Cardano/Wallet.hs
KtorZ: Although `walletSyncProgress` "only converts times at the tip", there's no guarantee that the tip itself isn't on an unsafe zone, isn't it?

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1640,29 +1643,25 @@ signPayment ctx wid argGenChange mkRewardAccount pwd md ttl cs = db & \DBLayer{.
...
+    :: TimeInterpreter (ExceptT PastHorizonException IO)
     -- ^ Context for time to slot calculation.
     -> Maybe NominalDiffTime
     -- ^ Time to live (TTL) in seconds from now.
-    -> IO (Either PastHorizonException SlotNo)
+    -> IO SlotNo
 getTxExpiry ti maybeTTL = do
     expTime <- addUTCTime ttl <$> getCurrentTime
-    try $ ti $ ceilingSlotAt expTime
+    interpretQuery (unsafeExtendSafeZone ti) (ceilingSlotAt expTime)
```

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -698,9 +699,25 @@ mkShelleyWallet ctx wid cp meta pending progress = do
...
+    -- an unextended @ti@ will simply fail because of uncertainty about the next
+    -- fork.
+    --
+    -- @unsafeExtendSafeZone@ performs the calculation as if no fork will occur.
+    -- This should be fine because:
+    -- 1. We expect the next few eras to have the same epoch length as Shelley
+    -- 2. It shouldn't be the end of the world to return the wrong time.
+    --
+    -- But ultimately, we might want to make the uncertainty transparent to API
+    -- users.
```
KtorZ: Perhaps link this comment to a JIRA technical-debt ticket or this will never happen?

lib/shelley/src/Cardano/Wallet/Shelley.hs
```diff
@@ -333,7 +335,8 @@ serveWallet
                 poolDatabaseDecorator
                 poolsDbTracer
                 (Pool.defaultFilePath <$> dir)
-                (timeInterpreter nl)
+                (neverFails "withPoolsMonitoring never forecasts into the future"
+                    $ timeInterpreter nl)
```
KtorZ, line 339: What about stake pool retirements :thinking:  ? Or does the conversion only happens with the API-layer's time interpreter?

Anviking: Hm, yeah, not obvious. This is used only for
1. [`selectPoolProduction`](https://github.com/input-output-hk/cardano-wallet/blob/4df2ecc5d5259ce333de6b2dfa39b6ac51b17652/lib/core/src/Cardano/Pool/DB/Sqlite.hs#L898-L908) which was jormungandr-specific and we should remove
2. [rolling back stake distributions](https://github.com/input-output-hk/cardano-wallet/blob/4df2ecc5d5259ce333de6b2dfa39b6ac51b17652/lib/core/src/Cardano/Pool/DB/Sqlite.hs#L487), which I guess also is unused and should be removed

Changing `"withPoolsMonitoring"` in the reason to `"pool db"` might help — if we don't want to clean this up now — but I think a separate PR would be better.

piotr-iohk:
> This is now updated to use the 1.22.1 release.

I guess we should now update to 1.23.0.

lib/shelley/cardano-wallet.cabal
rvl: It's just that the package name is misleading. We can fix that later - it's slightly easier to do without cardano-wallet-jormungandr.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
rvl: I don't think so - not at the moment. There is this:
```
data instance BlockConfig (ShelleyBlock era) = ShelleyConfig {..}
```
It fits with AllegraEra and MaryEra.

rvl:
I would like to get #2352 merged first, rewrite the `unsafeExtendSafeZone` fix in terms of that, then merge. Then do 1.23.0 in the next PR.

Anviking:
Rebased off #2352, but need to look through the TimeInterpreter commit with a fresh eye tomorrow.

Also I think changing `unsafeExtendSafeZone` to `unsafeExtendShelleySafeZone` — which only extends the safe zone if the era is Shelley or later might be an idea, to guard against rare extreme cases. If the era is Byron we know that the conversion _will_ be wrong on mainnet.

lib/core/src/Cardano/Wallet.hs
Anviking: The safe-zone is at least a significant portion of an epoch. And we expect the wallet tip to be behind the node tip.

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: Done -> ADP-575

PR #2246, opened 2020-10-15
Get epoch and slot lengths from HFC History Interpreter

rvl:
### Issue Number

#2226

### Overview

- [x] Remove slot length and epoch length from genesis parameters, because we will be getting them with HFC queries.
- [x] Hard-code activeSlotCoeff to 1.0 as before. There needs to be a ledger change to support this.
- [x] Add these parameters to SHELLEY_NETWORK integration test.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -1818,8 +1818,11 @@ getNetworkParameters
     -> Handler ApiNetworkParameters
 getNetworkParameters (_block0, np, _st) nl = do
     pp <- liftIO $ NW.getProtocolParameters nl
-    let (apiNetworkParams, epochNoM) =
-            toApiNetworkParameters np { protocolParameters = pp }
+    let pastHorizon :: PastHorizonException -> IO W.SlottingParameters
+        pastHorizon _ = pure (slottingParameters np)
+    sp <- liftIO $ handle pastHorizon $ NW.getSlottingParametersForTip nl
+    let (apiNetworkParams, epochNoM) = toApiNetworkParameters np
+            { protocolParameters = pp, slottingParameters = sp }
     case epochNoM of
         Just epochNo -> do
             epochStartTime <-
@@ -2170,7 +2173,7 @@ registerWorker
 registerWorker ctx coworker wid =
     void $ Registry.register @_ @ctx re ctx wid config
   where
-    (_, NetworkParameters gp _, _) = ctx ^. genesisData
+    (_, NetworkParameters gp _ _, _) = ctx ^. genesisData
     re = ctx ^. workerRegistry
     df = ctx ^. dbFactory
     config = MkWorker
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -1816,12 +1816,10 @@ getNetworkParameters
     :: (Block, NetworkParameters, SyncTolerance)
     -> NetworkLayer IO t Block
     -> Handler ApiNetworkParameters
-getNetworkParameters (_block0, np, _st) nl = do
+getNetworkParameters (_block0, genesisNp, _st) nl = do
     pp <- liftIO $ NW.getProtocolParameters nl
-    let pastHorizon :: PastHorizonException -> IO W.SlottingParameters
-        pastHorizon _ = pure (slottingParameters np)
-    sp <- liftIO $ handle pastHorizon $ NW.getSlottingParametersForTip nl
-    let (apiNetworkParams, epochNoM) = toApiNetworkParameters np
+    sp <- liftIO $ ignorePastHorizon $ NW.getSlottingParametersForTip nl
+    let (apiNetworkParams, epochNoM) = toApiNetworkParameters genesisNp
             { protocolParameters = pp, slottingParameters = sp }
     case epochNoM of
         Just epochNo -> do
@@ -1833,6 +1831,13 @@ getNetworkParameters (_block0, np, _st) nl = do
                     ApiEpochInfo (ApiT epochNo) epochStartTime }
         Nothing ->
             pure apiNetworkParams
+  where
+    -- A PastHorizonException should never happen here because the ledger is
+    -- being queried for slotting info about its own tip.  So we return the
+    -- genesis slotting parameters as a fallback.
+    ignorePastHorizon = handle pastHorizon
+    pastHorizon :: PastHorizonException -> IO W.SlottingParameters
+    pastHorizon _ = pure (slottingParameters genesisNp)
 
 getNetworkClock :: NtpClient -> Bool -> Handler ApiNetworkClock
 getNetworkClock client = liftIO . getNtpStatus client
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
@@ -451,12 +451,12 @@ migrateManually tr proxy defaultFieldValues =
         hardLowerBound = toText $ fromEnum $ minBound @(Index 'Hardened _)
         rndAccountIx   = DBField RndStateAddressAccountIndex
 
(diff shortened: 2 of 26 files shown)


Anviking:
👌 seems like a good idea

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1079,11 +1079,10 @@ mkCheckpointEntity wid wal =
...
         , checkpointGenesisHash = BlockId (coerce (gp ^. #getGenesisBlockHash))
         , checkpointGenesisStart = coerce (gp ^. #getGenesisBlockDate)
-        , checkpointSlotLength = coerceSlotLength $ gp ^. #getSlotLength
-        , checkpointEpochLength = coerce (gp ^. #getEpochLength)
+        , checkpointSlotLength = 0
+        , checkpointEpochLength = 0
         , checkpointEpochStability = coerce (gp ^. #getEpochStability)
-        , checkpointActiveSlotCoeff =
-            W.unActiveSlotCoefficient (gp ^. #getActiveSlotCoefficient)
+        , checkpointActiveSlotCoeff = 0
```
KtorZ: What about renaming the fields with a `Unused` suffix so it is a bit more transparent / consistent with other unused fields? It's a bit sad that we can't drop columns in SQLite tables :(

lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -248,6 +251,26 @@ defaultRetryPolicy =
...
+getSlottingParametersForTip
+    :: Monad m
+    => NetworkLayer m target block
+    -> m SlottingParameters
+getSlottingParametersForTip nl = do
+    tip <- either (const 0) slotNo <$> runExceptT (currentNodeTip nl)
+
+    -- TODO: Query activeSlotCoeff. Where to get it from though? Need to be able
+    -- to query Globals from ledger.
+    let getActiveSlotCoeff = pure (ActiveSlotCoefficient 1.0)
```
KtorZ, line 267: We can't really hard-code this value because we use a different value for testing than what's use on the mainnet and testnet networks :/ 
I'd suggest to keep the active slot coefficient untouched and keep retrieving it from the Shelley genesis. The value for Byron is always known and equal to `1.0` so we could have this function return `1.0` or the shelley value depending on the era we're in?

In the meantime, let's ask if the `Globals` can be made available through the LSQ protocol.

Anviking: > In the meantime, let's ask if the Globals can be made available through the LSQ protocol.

Think Rodney already asked -> https://input-output-rnd.slack.com/archives/CFKLUH4R0/p1603176743124800, and it looked like they were going to add it.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
rvl: OK done.

lib/core/src/Cardano/Wallet/Network.hs
rvl: Yes, I could add an era switch for hardcoding the mainnet activeSlotCoeff value, or we could wait for it to be implemented in shelley-ledger and cardano-node.

rvl:
bors try

KtorZ:
> Is this a new one?

@rvl , no, seen in the past. This property is sometimes failing on rare cases but we haven't investigated the reason _yet_. The result looks okay everytime, but maybe a `>` instead of `>=` in an assertion is causing this to fail.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1799,8 +1799,11 @@ getNetworkParameters
     -> Handler ApiNetworkParameters
 getNetworkParameters (_block0, np, _st) nl = do
     pp <- liftIO $ NW.getProtocolParameters nl
-    let (apiNetworkParams, epochNoM) =
-            toApiNetworkParameters np { protocolParameters = pp }
+    let pastHorizon :: PastHorizonException -> IO W.SlottingParameters
+        pastHorizon _ = pure (slottingParameters np)
```
Anviking: In the unlikely event this happens, I think we should rather do something else than using the genesis slotting parameters. Say the tip is in Shelley of the `Byron;Shelley;Allegra` mode, then the genesis slotting params would be misleading, no?

Seems unlikely, and maybe not the _most_ important, but maybe add a comment about it at least?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1799,8 +1799,11 @@ getNetworkParameters
     -> Handler ApiNetworkParameters
 getNetworkParameters (_block0, np, _st) nl = do
```
Anviking: Just a thought: Maybe we should call this `genesisNp` or something.

lib/core/test/unit/Cardano/Wallet/Primitive/SyncProgressSpec.hs
```diff
@@ -56,22 +56,19 @@ import Test.QuickCheck.Monadic
 
 spec :: Spec
 spec = do
-    let gp = GenesisParameters
+    let t0 = read "2019-11-09 16:43:02 UTC"
+    let sp = SlottingParameters
             { getEpochLength = EpochLength 21600
             , getSlotLength  = SlotLength 10
-            , getGenesisBlockDate = StartTime (read "2019-11-09 16:43:02 UTC")
-            , getGenesisBlockHash = Hash ""
```
Anviking, line 63: 👍

Anviking (approved):
lgtm! Maybe a comment could be nice about https://github.com/input-output-hk/cardano-wallet/pull/2246#discussion_r517201223 if you agree.

Tried locally, on master:
```
$ cardano-wallet network parameters | jq .
Ok.
{
  "slot_length": {
    "quantity": 20,
    "unit": "second"
  },
  "decentralization_level": {
    "quantity": 48,
    "unit": "percent"
  },
  "epoch_stability": {
    "quantity": 2160,
    "unit": "block"
  },
  "genesis_block_hash": "5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb",
  "blockchain_start_time": "2017-09-23T21:44:51Z",
  "desired_pool_number": 150,
  "epoch_length": {
    "quantity": 21600,
    "unit": "slot"
  },
  "active_slot_coefficient": {
    "quantity": 100,
    "unit": "percent"
  },
  "hardfork_at": {
    "epoch_number": 208
  },
  "minimum_utxo_value": {
    "quantity": 1000000,
    "unit": "lovelace"
  }
}
```

and with this pr:
```bash
$ cardano-wallet network parameters | jq .
Ok.
{
  "slot_length": {
    "quantity": 1,
    "unit": "second"
  },
  "decentralization_level": {
    "quantity": 48,
    "unit": "percent"
  },
  "epoch_stability": {
    "quantity": 2160,
    "unit": "block"
  },
  "genesis_block_hash": "5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb",
  "blockchain_start_time": "2017-09-23T21:44:51Z",
  "desired_pool_number": 150,
  "epoch_length": {
    "quantity": 432000,
    "unit": "slot"
  },
  "active_slot_coefficient": {
    "quantity": 100,
    "unit": "percent"
  },
  "hardfork_at": {
    "epoch_number": 208
  },
  "minimum_utxo_value": {
    "quantity": 1000000,
    "unit": "lovelace"
  }
}
```

lib/core/src/Cardano/Wallet/Api/Server.hs
rvl: Yes, good thought.

lib/core/src/Cardano/Wallet/Api/Server.hs
rvl: I added a comment. It should never happen because we ask the node for slotting parameters of its current tip. That should always be within the horizon. The only time I have seen the PastHorizonException is when the node tip was at genesis.

PR #2253, opened 2020-10-16
ADP-494 - Add POST endpoint for script address

paweljakubas:
# Issue Number

https://jira.iohk.io/browse/ADP-494

# Overview

- [x] I have updated swagger with script as string
- [x]  I have added script as json 
- [x] I have extended Api 
- [x] I have extended core unit tests
- [x] I have added cardano-addresses based impl
- [x] I have added illustrative integration tests 

# Comments

- Uses input-output-hk/cardano-addresses#87


specifications/api/swagger.yaml
```diff
@@ -125,6 +125,11 @@ x-blockReference: &blockReference
     time: *date
     height: *numberOfBlocks
 
+x-networkTag: &networkTag
+  description: Can be null for 'Icarus' and 'Byron' styles.
+  type: integer
+  minimum: 0
```
KtorZ: This should not be necessary because the server is already instantiated for one of the given networks. So the network tag is implicit based on how one started the server (`--mainnet` or `--testnet`)

specifications/api/swagger.yaml
```diff
@@ -321,6 +326,30 @@ x-poolMetadataSource: &poolMetadataSource
       format: uri
       example: https://smash.cardano-mainnet.iohk.io/
 
+x-walletCredentialPubkey: &walletCredentialPubkey
+  description: An extended public key (public key + chain code) for credential
```
KtorZ: Why require an extended key here? The chain code is unnecessary.

specifications/api/swagger.yaml
```diff
@@ -321,6 +326,30 @@ x-poolMetadataSource: &poolMetadataSource
...
+  type: string
+  format: hex
```
KtorZ: What about bech32 ;) ?

specifications/api/swagger.yaml
```diff
@@ -610,6 +640,16 @@ x-signedTransactionBlob: &signedTransactionBlob
   type: string
   format: binary
 
+x-addressParts: &addressParts
```
paweljakubas: enterprise address ->  `{ networkTag, paymentCredential }`
reward account address -> `{ networkTag, stakeCredential }`
stake address -> `{ networkTag, paymentCredential, stakeCredential }`

specifications/api/swagger.yaml
```diff
@@ -321,6 +326,30 @@ x-poolMetadataSource: &poolMetadataSource
...
+  description: An extended public key (public key + chain code) for credential
+  type: string
+  format: hex
+  minLength: 128
+  maxLength: 128
+  example: 1423856bc91c49e928f6f30f4e8d665d53eb4ab6028bd0ac971809d514c92db11423856bc91c49e928f6f30f4e8d665d53eb4ab6028bd0ac971809d514c92db1
+
+x-walletScript: &walletScript
+  description: A script as text with predicates and verification keys. Here probably we should add a full-fledged script object
+  type: string
```
KtorZ, line 326: How is that a `string` :thinking: ? Isn't the goal here to re-use the JSON object instances from cardano-addresses?

paweljakubas: it is going to be json object. just doing this in steps

specifications/api/swagger.yaml
paweljakubas: to my best knowledge we are hashing extended public key with blake2b not just public key -> https://github.com/input-output-hk/cardano-addresses/blob/master/core/lib/Cardano/Address/Style/Shelley.hs#L639

specifications/api/swagger.yaml
paweljakubas: removed in https://github.com/input-output-hk/cardano-wallet/pull/2253/commits/ccd4eb486e7739d6a71497891baf70b0f78eff86

specifications/api/swagger.yaml
paweljakubas: used `xpub` hrp -> https://github.com/input-output-hk/cardano-wallet/pull/2253/commits/ccd4eb486e7739d6a71497891baf70b0f78eff86

specifications/api/swagger.yaml
paweljakubas: my attempt -> https://github.com/input-output-hk/cardano-wallet/pull/2253/commits/ccd4eb486e7739d6a71497891baf70b0f78eff86
But when
```
$ openapi-spec-validator --schema 3.0.0 specifications/api/swagger.yaml 
found undefined alias 'walletCredentialScript'
  in "<file>", line 357, column 14
```
need to recall how to sneak recursive scheme in swagger

specifications/api/swagger.yaml
KtorZ: No we're not :) 

https://github.com/input-output-hk/cardano-addresses/blob/1939121afcd125b6c6300a1b2a8b014007a0356c/core/lib/Cardano/Address/Style/Shelley.hs#L869-L872

`xpubPublicKey` is the public key, without chain code.

specifications/api/swagger.yaml
KtorZ: Look at transaction metadatum. I recall we did something similar.

specifications/api/swagger.yaml
paweljakubas: indeed! Thanks! I made this mistake mainly because of this : 
```
$ cat signingKey1.prv | cardano-address key public --bech32
xpub13pfp60uf4ru7cmj5d57xjdzj6lagn328ah67h35urqrl92jn094z0s07fcy2pvfw3xfhna6fa24na8nk6lraymjcrl5wwj6chsjqtjqlfw44z
```
So we will need a way to get public key only (not chain code included) in cardano-addresses, right?

KtorZ: I think that'd be an interesting feature indeed. Be able to get just the public part, without chain code

KtorZ: Alternatively, we can have the API accepts _both_, but the chain code should not a _required_

paweljakubas: corrected in https://github.com/input-output-hk/cardano-wallet/pull/2253/commits/2f9180366bdb4319b19c3ad90bb06e5f52fc350e

specifications/api/swagger.yaml
paweljakubas: thanks, now it seems to be ok -> https://github.com/input-output-hk/cardano-wallet/pull/2253/commits/2f9180366bdb4319b19c3ad90bb06e5f52fc350e

stack.yaml
```diff
@@ -87,177 +87,177 @@ nix:
 # package cardano-crypto
 #   tests: False
 #   benchmarks: False
-# 
+#
```
KtorZ: I think that stack2cabal relies on these whitespaces and they shouldn't be trimmed. 

@paweljakubas, is your editor correctly picking up the config from: https://github.com/input-output-hk/cardano-wallet/blob/master/.editorconfig#L11-L12

stack.yaml
```diff
@@ -46,7 +46,7 @@ extra-deps:
 
 # cardano-addresses-2.1.0
 - git: https://github.com/input-output-hk/cardano-addresses
-  commit: e0ab4587266430a08734e1aec1c29d261a9a3b70
+  commit: 1399751501b03346265b2e429e5832bacd6df279
```
KtorZ, line 49: We should only pick released version of cardano-addresses. But I am quite reluctant releasing it in the state it is right now :/ ... I'd rather have it compliant with CIP5 for the release.

paweljakubas: agree, pasted it as it is just in order to have a way to interact with what we have in cardano-addresses

stack.yaml
paweljakubas: it seems so - this is config that is taken by my editor for this file: 
```
# EditorConfig for /home/pawel/Work/IOHK/cardano-wallet/stack.yaml                                                                                                                                                 
indent_style = space                                                                                                                                                                                               
indent_size = 2                                                                                                                                                                                                    
end_of_line = lf                                                                                                                                                                                                   
charset = utf-8                                                                                                                                                                                                    
trim_trailing_whitespace = false                                                                                                                                                                                   
insert_final_newline = true                                                                                                                                                                                        
tab_width = 2
```

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1866,6 +1880,61 @@ derivePublicKey ctx (ApiT wid) (ApiT role_) (ApiT ix) = do
...
+      netTag = 1
+      spendingFromKey = CA.PaymentFromKey . CA.liftXPub . toXPub
+      spendingFromScript = CA.PaymentFromScript . CA.toScriptHash
+      stakingFromKey = CA.DelegationFromKey . CA.liftXPub . toXPub
+      stakingFromScript = CA.DelegationFromScript . CA.toScriptHash
+{--
+      netTag = case testEquality (typeRep @n) (typeRep @'Mainnet) of
+          Just{}  -> 1
+          Nothing -> 0
+--}
```
paweljakubas: This one is escaping me -> I want n -> Int ... for now I hacked it to be 1 (Mainnet) to carry on with integration tests

KtorZ: Note that this should not be in `core`, because it only concerns cardano-node and really is cardano-node specific. Plus, you should be able to simply re-use: https://github.com/input-output-hk/cardano-wallet/blob/368b9fc27249e1faf6577c020297b74d4cf6f161/lib/shelley/src/Cardano/Wallet/Shelley.hs#L203-L218

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -384,6 +398,34 @@ data ApiAddress (n :: NetworkDiscriminant) = ApiAddress
...
+data AnyAddressType =
+      EnterpriseDelegating
+    | RewardAccount
+    deriving (Eq, Show, Bounded, Enum)
+
+data AnyAddress = AnyAddress
+    { payload :: ByteString
+    , flavour :: AnyAddressType
+    , network :: Int
+    } deriving (Eq, Generic, Show)
```
KtorZ, line 429: Why not re-use the existing types for Addresses ?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -384,6 +398,34 @@ data ApiAddress (n :: NetworkDiscriminant) = ApiAddress
...
+newtype ApiScript = ApiScript
+    { script :: ApiT Script
+    } deriving (Eq, Generic, Show)
+
+newtype ApiPubKey = ApiPubKey
+    { pubKey :: ByteString
+    } deriving (Eq, Generic, Show)
+
+data Credential = Credential
+    { credential :: Either ApiPubKey ApiScript
```
KtorZ: May I suggest a sum-type instead of wrapping a `Either` ? 

```hs
data Credential 
    = CredentialPubKey ...
    | CredentialScript
```

?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1866,6 +1880,61 @@ derivePublicKey ctx (ApiT wid) (ApiT role_) (ApiT ix) = do
...
+                         , EnterpriseDelegating )
+                (Just (Credential (Right (ApiScript (ApiT script)))), Just (Credential (Left (ApiPubKey bytes)))) ->
+                    pure ( unAddress $
+                           CA.delegationAddress discriminant (spendingFromScript script) (stakingFromKey bytes)
+                         , EnterpriseDelegating )
+                (Just (Credential (Right (ApiScript (ApiT script1)))), Just (Credential (Right (ApiScript (ApiT script2)))) )->
+                    pure ( unAddress $
+                           CA.delegationAddress discriminant (spendingFromScript script1) (stakingFromScript script2)
+                         , EnterpriseDelegating )
+                (Nothing, Nothing) -> fail "At least one credential is required"
```
KtorZ: That's one big chunk :grimacing: 
My recommendation would be to capture this already in the API data-type itself. So, instead of:

```hs
data ApiCredentials = ApiCredentials
    { spending :: !(Maybe Credential)
    , staking :: !(Maybe Credential)
```

have something along the lines of:

```hs
data ApiAddressData 
    = AddrBase Credential Credential
    | AddrPtr Credential Ptr
    | AddrEnterprise Credential
    | AddrBootstrap Address
```

this would be quite more readable and also get rid of the "impossible" case.

specifications/api/swagger.yaml
```diff
@@ -321,6 +321,58 @@ x-poolMetadataSource: &poolMetadataSource
...
+    Script allows establishing either spending or staking guard that could be removed
+    when a special combination of signing keys is present.
+
+    The script can be constructed using four primitives:
+     - verification key hash for which a corresponding signing key is expected
+     - all condition for which all list elements' expectation are met for the condition to be valid
+     - any condition for which any list element' expectation is to be met for the condition to be valid
+     - some condition for which a minimal number of list elements' expectation is to be met for the condition to be valid
+    example:
+      0: "script_vkh18srsxr3khll7vl3w9mqfu55n6wzxxlxj7qzr2mhnyreluzt36ms"
```
rvl: This value should be a yaml list of examples, rather than an object with numeric keys. The single example object is specific to Metadata.

specifications/api/swagger.yaml
```diff
@@ -1474,6 +1525,37 @@ components:
...
+        pub_key: *credentialPubKey
+
+    AnyAddress: &AnyAddress
+      type: object
+      required:
+        - address
+      properties:
+        address: *anyAddress
+
+    ApiCredential: &ApiCredential
```
rvl, line 1577: Could this be renamed to `ScriptAddressCredential`? Are the `Api` prefixes needed for anything? This is a swagger file - it's all API.

specifications/api/swagger.yaml
```diff
@@ -1474,6 +1525,37 @@ components:
...
+        - script
+      properties:
+        script: *script
+
+    ApiPubKey: &ApiPubKey
+      type: object
+      required:
+        - pub_key
+      properties:
+        pub_key: *credentialPubKey
```
rvl: I think these could both be removed and embedded directly into `CredentialValue`

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -384,6 +398,34 @@ data ApiAddress (n :: NetworkDiscriminant) = ApiAddress
     , state :: !(ApiT AddressState)
     } deriving (Eq, Generic, Show)
 
+newtype ApiScript = ApiScript
+    { script :: ApiT Script
```
rvl: I think if the sole field is going to be `ApiT Script`, then you don't need the `ApiScript` newtype.

lib/core/src/Cardano/Wallet/Api/Types.hs
rvl: Should it be called ApiCredential as well for consistency?

lib/core/src/Cardano/Wallet/Api/Types.hs
rvl: Perhaps the newtype might be needed if you want automatic JSON instances, but it looks like they are non-generically defined below.

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -956,6 +969,58 @@ instance Arbitrary ApiEpochInfo where
     arbitrary = ApiEpochInfo <$> arbitrary <*> genUniformTime
     shrink _ = []
 
+instance Arbitrary ApiScript where
+    arbitrary = ApiScript . ApiT <$> arbitrary
+
+instance Arbitrary Script where
+    arbitrary = do
```
rvl: Same comment about generators of arbitrary tree structures applies here as it did in cardano-addresses.

In fact, could the types and generators from cardano-addresses be re-used?

stack.yaml
rvl: I think `trim_trailling_whitespace=true` is a nice thing to have.

When auto-generated whitespace gets removed in a file that you edited, use your editor's git function to discard those irrelevant hunks.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1866,6 +1880,61 @@ derivePublicKey ctx (ApiT wid) (ApiT role_) (ApiT ix) = do
         k <- liftHandler $ W.derivePublicKey @_ @s @k @n wrk wid role_ ix
         pure $ ApiVerificationKey (getRawKey k, role_)
 
+postAnyAddress
+    :: forall n. ApiCredentials
+    -> Handler AnyAddress
+postAnyAddress body = do
```
rvl: It looks like not just _any_ address can be posted. Is that right?

rvl (commented):
Good start :+1:

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/2253/commits/1cdaf25774c19ac6a3b37e0cebb48b9628095f19

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: thanks for this hint. did it in https://github.com/input-output-hk/cardano-wallet/pull/2253/commits/3cf22f47f91eba47a1ebee825df32040dab8d1d7 also did some reshuffling of code connected with HasNetworkId to use it

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: If I use Addresses I will be able to differentiate between `addr` and `addr_test` based on network. but I will not be able to use reward_account's `stake` and `stake_test` (to my best knowledge). This was main reason I introduced this type

specifications/api/swagger.yaml
KtorZ: @rvl we typically keep the same naming as in the `Cardano.Wallet.Api.Types` module for the swagger file references because it creates a straightforward mapping for the automated schema validations we perform in tests. This way, one can write `ToSchema` instances as:

```hs
instance ToSchema (ApiSelectCoinsPayments n) where
    declareNamedSchema _ = declareSchemaForDefinition "ApiSelectCoinsPayments"

instance ToSchema ApiSelectCoinsAction where
    declareNamedSchema _ = declareSchemaForDefinition "ApiSelectCoinsAction"

instance ToSchema (ApiCoinSelection n) where
    declareNamedSchema _ = declareSchemaForDefinition "ApiCoinSelection"

instance ToSchema ApiWallet where
    declareNamedSchema _ = declareSchemaForDefinition "ApiWallet"

instance ToSchema ApiByronWallet where
    declareNamedSchema _ = declareSchemaForDefinition "ApiByronWallet"
```

KtorZ: We could even push it further and have default instances relying on generics and Typeable.

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: good idea, indeed it made things simpler -> https://github.com/input-output-hk/cardano-wallet/pull/2253/commits/1071e78b24b4974a03f4dee7b88c8743e5327775

specifications/api/swagger.yaml
paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/2253/commits/e6c0510a40d5f99f12d63caa03b7654078fe3ae9

specifications/api/swagger.yaml
paweljakubas: I wanted them to be exposed to have a way in TypeSpec to check

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
paweljakubas: did in https://github.com/input-output-hk/cardano-wallet/pull/2253/commits/e6c0510a40d5f99f12d63caa03b7654078fe3ae9

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: yes, I can construct enterprise, reward account and base address for any credential (either from key or script). There is also pointer address missing that could be added in forthcoming PR. That's the rationale I proposed the name AnyAddress

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
```diff
@@ -249,3 +255,86 @@ spec = describe "SHELLEY_ADDRESSES" $ do
         let str = "patate"
         r <- request @Aeson.Value ctx (Link.inspectAddress str) Default Empty
         expectResponseCode @IO HTTP.status400 r
+
+    -- Generating golden test data for enterprise addresses:
+    --- $ cardano-address script hash "$(cat script.txt)" \
+    --- | cardano-address address payment --from-script --network-tag mainnet
+    it "ANY_ADDRESS_POST_01 - Golden tests for enterprise script address - signature" $ \ctx -> do
```
piotr-iohk, line 264: Nitpicking but why golden? Afaik, golden are unit tests with output in separate file which is not the case here.

specifications/api/swagger.yaml
```diff
@@ -321,6 +321,57 @@ x-poolMetadataSource: &poolMetadataSource
...
+      - { "all": [ "script_vkh18srsxr3khll7vl3w9mqfu55n6wzxxlxj7qzr2mhnyreluzt36ms"
+                  , { "any" : [ "script_vkh18srsxr3khll7vl3w9mqfu55n6wzxxlxj7qzr2mhnyreluzt37ms"
+                              , "script_vkh18srsxr3khll7vl3w9mqfu55n6wzxxlxj7qzr2mhnyreluzt38ms"] }
+                  ] }
+      - { "all": [ "script_vkh18srsxr3khll7vl3w9mqfu55n6wzxxlxj7qzr2mhnyreluzt36ms"
+                  , { "some" : { "from" : [ "script_vkh18srsxr3khll7vl3w9mqfu55n6wzxxlxj7qzr2mhnyreluzt37ms"
+                                          , "script_vkh18srsxr3khll7vl3w9mqfu55n6wzxxlxj7qzr2mhnyreluzt38ms"]
+                               , "at_least" : 2
+                               }
+                    } ] }
```
piotr-iohk: There seems to be some formatting issue here:

Screenshot from 2020-10-30 15-37-19

paweljakubas: thanks, will fix

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
paweljakubas: golden because I have generated golden data in cardano-address for all examples, and check if all json/handling/.. machinery comes to the same results

paweljakubas:
I added golden for enterprise, reward account and base with script(s) as credential. I will add more (probably in separate PR) containing public key in credential and mixed after I will add `--pub` to publickey option (now I get `--xpub`) cardano-addresses very soon.

Plus I have problem with this trimming - cannot revert it

paweljakubas:
@rvl, @KtorZ  the last 5 commits added remaining golden tests involving public key. What is important I used this PR to produce them https://github.com/input-output-hk/cardano-addresses/pull/87 

Still having problem with trimming though.

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -968,6 +983,56 @@ instance Arbitrary ApiEpochInfo where
...
+                , RequireSomeOf atLeast scripts
+                ]
+    shrink = genericShrink
+
+instance Arbitrary KeyHash where
+    arbitrary = KeyHash . BS.pack <$> vectorOf 28 arbitrary
+
+instance Arbitrary ApiPubKey where
+    arbitrary =
+        (ApiPubKey . BS.pack) <$> replicateM 32 arbitrary
```
rvl:
```suggestion
        ApiPubKey . BS.pack <$> replicateM 32 arbitrary
```

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -968,6 +983,56 @@ instance Arbitrary ApiEpochInfo where
...
+
+instance Arbitrary ApiCredential where
+    arbitrary = do
+        let pubKeyGen = arbitrary :: Gen ApiPubKey
+        let scriptGen = arbitrary :: Gen ApiScript
+        oneof [ CredentialPubKey <$> pubKeyGen
+              , CredentialScript <$> scriptGen ]
```
rvl:
```suggestion
    arbitrary = oneof [ CredentialPubKey <$> arbitrary, CredentialScript <$> arbitrary ]
```

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -968,6 +983,56 @@ instance Arbitrary ApiEpochInfo where
...
+
+instance Arbitrary ApiAddressData where
+    arbitrary = do
+        credential1 <- arbitrary
+        credential2 <- arbitrary
+        oneof
+            [ pure $ AddrEnterprise credential1
+            , pure $ AddrRewardAccount credential2
+            , pure $ AddrBase credential1 credential2
+            ]
```
rvl:
```suggestion
    arbitrary = do
        credential1 <- arbitrary
        credential2 <- arbitrary
        elements
            [ AddrEnterprise credential1
            , AddrRewardAccount credential2
            , AddrBase credential1 credential2
            ]
```

specifications/api/swagger.yaml
```diff
@@ -321,6 +321,62 @@ x-poolMetadataSource: &poolMetadataSource
...
+  description: |
+    Script allows establishing either spending or staking guard that could be removed
+    when a special combination of signing keys is present.
+
+    The script can be constructed using four primitives:
+     - verification key hash for which a corresponding signing key is expected
+     - all condition for which all list elements' expectation are met for the condition to be valid
+     - any condition for which any list element' expectation is to be met for the condition to be valid
+     - some condition for which a minimal number of list elements' expectation is to be met for the condition to be valid
+    example:
```
rvl: I think this line needs to be de-indented so that it's not included in the description string.

specifications/api/swagger.yaml
rvl: OK fair enough

specifications/api/swagger.yaml
rvl: Aha ok thanks.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1311,6 +1355,81 @@ instance (PassphraseMaxLength purpose, PassphraseMinLength purpose)
 instance ToJSON (ApiT (Passphrase purpose)) where
     toJSON = toJSON . toText . getApiT
 
+instance FromJSON ApiScript where
+    parseJSON =  withObject "Script" $ \o -> do
+        script' <- o .: "script"
+        (ApiScript . ApiT) <$> parseJSON script'
```
rvl: I think this should work...

```suggestion
    parseJSON =  withObject "ApiScript" $ \o ->
        ApiScript . ApiT <$> o .: "script"
```

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1311,6 +1355,81 @@ instance (PassphraseMaxLength purpose, PassphraseMinLength purpose)
...
+    toJSON (ApiPubKey key') = do
+        let hrp = [Bech32.humanReadablePart|addr_vk|]
+        String $ T.decodeUtf8 $ encode (EBech32 hrp) key'
+
+instance FromJSON ApiCredential where
+    parseJSON obj = do
+        key' <-
+            (withObject "ApiCredential" $
+             \o -> o .:? "pub_key" :: Aeson.Parser (Maybe Text)) obj
+        case key' of
```
rvl: You can more succinctly define this parser with `<|>` and applicative style.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1311,6 +1355,81 @@ instance (PassphraseMaxLength purpose, PassphraseMinLength purpose)
...
+                Left (TextDecodingError err) -> fail err
+
+instance ToJSON AnyAddress where
+    toJSON (AnyAddress p addrType net) = do
+        let hrp = case (addrType, net) of
+                (EnterpriseDelegating, 1) -> [Bech32.humanReadablePart|addr|]
+                (RewardAccount, 1) -> [Bech32.humanReadablePart|stake|]
+                (EnterpriseDelegating, 0) -> [Bech32.humanReadablePart|addr_test|]
+                (RewardAccount, 0) -> [Bech32.humanReadablePart|stake_test|]
+                _ -> error "Wrong network tag in AnyAddress"
```
rvl: It should be `fail` not `error`.

You might also be able to say that any network which isn't mainnet is a testnet.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1311,6 +1355,81 @@ instance (PassphraseMaxLength purpose, PassphraseMinLength purpose)
...
+        object [ "address" .= String (T.decodeUtf8 $ encode (EBech32 hrp) p) ]
```
rvl: If there are lots of instances of `(T.decodeUtf8 . encode)`, you could make an `encodeText` function.

Also due to the magic of `(.=)`, the `Aeson.String` constructor isn't required.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1736,6 +1855,46 @@ instance FromText PostExternalTransactionData where
         Right load ->
             pure $ PostExternalTransactionData load
 
+instance FromText ApiPubKey where
+    fromText txt = case detectEncoding (T.unpack txt) of
```
rvl: If the only possible encoding is bech32, then it would be best to just decode in bech32 without trying to detect encoding.

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
```diff
@@ -395,3 +414,40 @@ server byron icarus shelley spl ntp =
...
+                 , EnterpriseDelegating )
+        AddrRewardAccount stakingCred -> do
+            let (Right stakeAddr) =
+                    CA.stakeAddress discriminant (stakingFrom stakingCred)
+            pure ( unAddress stakeAddr, RewardAccount )
+        AddrBase spendingCred stakingCred ->
+            pure ( unAddress $ CA.delegationAddress discriminant
+                     (spendingFrom spendingCred) (stakingFrom stakingCred)
+                 , EnterpriseDelegating )
+    pure $ AnyAddress addr addrType (fromInteger netTag)
```
rvl, line 434: Could we use the `NetworkId` type in `AnyAddress`?

rvl (commented):
Lots of tests - good.

I suppose this is blocked until the cardano-addresses PR is merged?

specifications/api/swagger.yaml
```diff
@@ -2195,6 +2335,17 @@ x-responsesPostExternalTransaction: &responsesPostExternalTransaction
       application/json:
         schema: *ApiTxId
 
+x-responsesPostAnyAddress: &responsesPostAnyAddress
+  <<: *responsesErr400
+  <<: *responsesErr405
```
piotr-iohk: I think it's ok to drop `405` now to be consistent with other endpoints after recent error codes cleanup/documentation. (cf: https://github.com/input-output-hk/cardano-wallet/pull/2258#discussion_r515004888)

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: agree, but if someone gives base16 or base58 then one will be given the error due to this https://github.com/input-output-hk/cardano-addresses/blob/master/core/lib/Codec/Binary/Encoding.hs#L100 and if bech32 format is detected then data part, hrp, etc are later verified

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
paweljakubas: I had at the beginning networkId in AnyAddress and talked about this with @KtorZ  and there was idea to use networkId implicitely, ie., follow the network that the wallet instantiated with...

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/2253/commits/595999169a24d1637760f2f14f02c4af9d3cc4ad

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/2253/commits/595999169a24d1637760f2f14f02c4af9d3cc4ad

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/2253/commits/595999169a24d1637760f2f14f02c4af9d3cc4ad

specifications/api/swagger.yaml
paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/2253/commits/595999169a24d1637760f2f14f02c4af9d3cc4ad

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: indeed, done in https://github.com/input-output-hk/cardano-wallet/pull/2253/commits/595999169a24d1637760f2f14f02c4af9d3cc4ad

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: missed a point why use here alternative :thinking:

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: exposed network id in error. left `error` as I use it to get HRP. how would you use fail here?

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: it seems String is needed, it doesn't want to compile without it

specifications/api/swagger.yaml
paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/2253/commits/595999169a24d1637760f2f14f02c4af9d3cc4ad

paweljakubas:
Thanks @rvl  @KtorZ @piotr-iohk  for all comments/hints/useful suggestions. 

@rvl we do not need to wait for  anything pending in cardano-addresses. We use bits that are already in master there.

lib/core/src/Cardano/Wallet/Api/Types.hs
rvl: I pushed a change to show why. It makes the parser shorter and clearer I think. Alternative means that it tries the parsers in turn from left to right until one succeeds.

lib/core/src/Cardano/Wallet/Api/Types.hs
rvl: Ah OK I see -- this is the toJSON not the fromJSON.
This integer network tag thing seems fishy.
Normally we parse them into types at the program boundaries.
Anyway, I believe that it would be correct to use `_test` for any network that's not mainnet (or staging).
So I pushed a fix for that.

lib/core/src/Cardano/Wallet/Api/Types.hs
rvl: It works without -- I pushed a fix.

rvl (dismissed):

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1360,6 +1404,73 @@ instance (PassphraseMaxLength purpose, PassphraseMinLength purpose)
...
+        parseEnterprise v <|>
+        parseRewardAccount v <|>
+        fail "ApiAddressData must have at least one credential."
+      where
+         parseBaseAddr = withObject "AddrBase" $ \o ->
+             AddrBase <$> o .: "spending" <*> o .: "staking"
+         parseEnterprise = withObject "AddrEnterprise" $ \o ->
+             AddrEnterprise <$> o .: "spending"
+         parseRewardAccount = withObject "AddrRewardAccount" $ \o ->
+             AddrRewardAccount <$> o .: "staking"
```
rvl: I don't think we should use the word "staking" here or anywhere.
"Stake address" is ok, but not "staking address".
The verb is to delegate.

rvl (changes requested):

specifications/api/swagger.yaml
```diff
@@ -2790,6 +2930,16 @@ x-responsesPostExternalTransaction: &responsesPostExternalTransaction
       application/json:
         schema: *ApiTxId
 
+x-responsesPostAnyAddress: &responsesPostAnyAddress
+  <<: *responsesErr400
+  <<: *responsesErr406
+  <<: *responsesErr415
+  202:
```
piotr-iohk, line 2916: As far as I see it is actually `201` on success.

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: I somewhat followed the naming from https://hydra.iohk.io/build/4598683/download/1/delegation_design_spec.pdf - see Fig 2. There is Payment Credential and Stake Credential...so I used spending/staking... You think spending/delegating is better right?

paweljakubas: please notice that `ApiAddressData` has data containing credentials used either for spending or staking or both...

rvl: I think use "payment" and "stake" here then. The word "staking" does not exist in the spec.

paweljakubas: ok, will change to `payment/stake`

specifications/api/swagger.yaml
paweljakubas: nice catch, will make it PostAccepted rather than PostCreated and it will return not 201 but 202as advertised in swagger.

paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/2253/commits/3d003584ca7245da6c5f7012b712a4880d7955be

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/2253/commits/3d003584ca7245da6c5f7012b712a4880d7955be

rvl (approved):

lib/core/src/Cardano/Wallet/Api/Client.hs
```diff
@@ -317,6 +317,7 @@ addressClient =
     let
         _listAddresses
             :<|> _inspectAddress
+            :<|> _postScriptAddress
```
KtorZ, line 320: small note: the endpoint suggest that we can post _any_ address here?

lib/core/test/data/Cardano/Wallet/Api/ApiAddressData.json
```diff
@@ -0,0 +1,347 @@
...
+                }
+            }
+        },
+        {
+            "payment": {
+                "pub_key": "addr_vk1v6835v2g6yx4xqgyzu3x785snq4yks2t0e8mgkldxeyhmx57ksvshkua8r"
+            },
+            "stake": {
+                "pub_key": "addr_vk1ney9mxfedm5gas7x85rdzdz5npgcv5hqpm4au99jr3gumdjf66nqmk5t8r"
+            }
```
KtorZ: It seems inconsistent with other primitives that there's an extra `pub_key` field here. What about simply:

```
{ "payment": "addr_vk1v6835v2g6yx4xqgyzu3x785snq4yks2t0e8mgkldxeyhmx57ksvshkua8r" }
```

and

```
{ "stake": "addr_vk1ney9mxfedm5gas7x85rdzdz5npgcv5hqpm4au99jr3gumdjf66nqmk5t8r" }
```

paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/2253/commits/595a91aa4d4cc2fe1557d7564d32d0dc23daa3ef

lib/core/src/Cardano/Wallet/Api/Client.hs
paweljakubas: yes, I "coined this name" due to the fact that ANY address could be constructed: base, reward account, enterprise, each with any credential. Of course there is also pointer address which was omitted in this PR (but could be added in the forthcoming)

PR #2254, opened 2020-10-16
select coins for delegation simplifications

KtorZ:
# Issue Number

#2200

# Overview

- 16a5f0cda96867b4bc0eaa37a125c4824e6d2a25
  :round_pushpin: **remove duplicated logic with regards to selections for join/quit**
    The code was becoming quite convoluted here with function names that
  were screaming for refactoring (when we start adding `'` to functions,
  it's usually a good sign that something can be simplified).

  The main change is actually in 'selectCoinsExternal' which is now
  parameterized over the selection to run. This way, the steps of
  assigning change addresses are factored out in this function and a lot
  of the duplication goes away.

  Another important change is that I've moved the signing and submission
  of join/quit outside of the body of the function. So that each step
  can be ran independently. This avoid the need for weird intermediate product
  types aggregating more and more information. Now, both functions are
  returning a 'DelegationAction' and merely checking that a join or quit
  is possible.

- 4a5f4d96ea74f89bf0920b2ebed3a57c3bee1ca2
  :round_pushpin: **return structured error for invalid wallets, instead of plain string**

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -125,9 +125,7 @@ module Cardano.Wallet
     -- ** Delegation
     , PoolRetirementEpochInfo (..)
     , joinStakePool
-    , joinStakePoolUnsigned
     , quitStakePool
-    , quitStakePoolUnsigned
     , selectCoinsForDelegation
     , estimateFeeForDelegation
     , signDelegation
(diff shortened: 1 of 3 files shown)


hasufell:
I don't really understand the changes from skimming over it. I'd have to have a deeper look at it if you require a detailed review, otherwise I'm fine with merging.

KtorZ:
@hasufell I have tried to explain the main changes in my PR comment. Did you read it? Do you have any questions or something I could explain?

PR #2255, opened 2020-10-19
do not enforce non-empty outputs for unsigned tx

KtorZ:
# Issue Number

#2200 

# Overview

- c60c5a60a6d1111e8dc5c7482a737bdc4e83c3b2
  :round_pushpin: **do not enforce non-empty outputs for unsigned tx**
    Actually, there are many use-cases for empty outputs. This occurs in
  particular often when performing a selection for delegation: in this
  scenario, there are typically no outputs at all and, depending on the
  size of the selected input(s), there might be no change at all
  (because of the minUTxO value).

# Comments

Should fix: 

- https://github.com/input-output-hk/cardano-wallet/pull/2213#issuecomment-710155543
- https://github.com/input-output-hk/cardano-wallet/pull/2213#issuecomment-711148053

Thanks @piotr-iohk :pray:

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types.hs
@@ -969,8 +969,20 @@ instance ToText TxStatus where
 data UnsignedTx input = UnsignedTx
     { unsignedInputs
         :: NonEmpty input
+        -- Inputs are *necessarily* non-empty because Cardano requires at least
+        -- one UTxO input per transaction to prevent replayable transactions.
+        -- (each UTxO being unique, including at least one UTxO in the
+        -- transaction body makes it seemingly unique).
+
     , unsignedOutputs
-        :: NonEmpty TxOut
+        :: [TxOut]
+        -- Unlike inputs, it is perfectly reasonable to have empty outputs. The
+        -- main scenario where this might occur is when constructing a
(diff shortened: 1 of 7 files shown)


lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -970,7 +970,7 @@ data UnsignedTx input = UnsignedTx
     { unsignedInputs
         :: NonEmpty input
     , unsignedOutputs
-        :: NonEmpty TxOut
+        :: [TxOut]
```
hasufell, line 978: Should add a comment here why we don't want NonEmpty. Will bite someone later, maybe.

hasufell (approved):

KtorZ:
bors merge

lib/core/src/Cardano/Wallet/Primitive/Types.hs
KtorZ: Good point, elaborated here: https://github.com/input-output-hk/cardano-wallet/compare/c60c5a60a6d1111e8dc5c7482a737bdc4e83c3b2..2bdbb29333beb82587055ab576619ec6760888f3

piotr-iohk (approved):
:tada:

KtorZ:
What the heck o.O

KtorZ:
bors merge

PR #2257, opened 2020-10-19
API data-types & specifications for arbitrary metadata signing.

KtorZ:
# Issue Number

ADP-481 > ADP-508

# Overview

- 209781bd4b5aee3df596ec05ae2b8cd7e955e913
  :round_pushpin: **define ToText/FromText instances for 'DerivationIndex'**
    I had to also move it in '...Primitive.AddressDerivation' because of otherwise circular dependencies (since this module is using the Bounded instances of 'Index', relying on the 'DerivationType' as type parameter. This is mostly extracting the text parsing that was already done in the JSON instance, and moving it as a ToText/FromText instances (and re-using it to build the JSON instances). I've however added a quick sanity text roundtrip property test in the meantime.

- c96593646a433eeba840eea6aab1c85c9753e61c
  :round_pushpin: **rename UTxO*ternal -> Utxo*ternal**
    We use the data-type constructor in order to generate text instances.
  However, our automatic casing function is doing pretty bad with the
  mixed cases and we end up with something like:

    - u_tx_o_external
    - u_tx_o_internal
    - mutable_account

  Instead of fixing the instance itself to work around this, I found it
  easier to just fix the type (which isn't used in many places).

- c4cbafe71b3c9a4271afb8c4417a85f1b8fd8171
  :round_pushpin: **do not output back invalid input when parsing text values.**
    This is a rather bad practice; if a value failed to parse, then we have very little idea about _what_ it is and what's it's shape. It may not be even readable / printable. Instead, the error message should show valid expected options.

- fb5ab0d15d1c37a1b70c2fd4caae7c86f4bc78ce
  :round_pushpin: **extend API and specifications to include the 'signMetadata' operation**

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -150,7 +150,7 @@ import Cardano.Mnemonic
     )
 import Cardano.Wallet.Primitive.AddressDerivation
     ( Depth (..)
-    , DerivationType (..)
+    , DerivationIndex (..)
     , Index (..)
     , NetworkDiscriminant (..)
     , Passphrase (..)
@@ -174,7 +174,6 @@ import Cardano.Wallet.Primitive.Types
     , ChimericAccount (..)
     , Coin (..)
     , DecentralizationLevel (..)
-    , DerivationIndex (..)
     , Direction (..)
     , EpochLength (..)
     , EpochNo (..)
@@ -251,8 +250,6 @@ import Data.Proxy
     ( Proxy (..) )
 import Data.Quantity
     ( Percentage, Quantity (..) )
-import Data.Scientific
-    ( Scientific, toBoundedInteger )
 import Data.String
     ( IsString )
 import Data.Text
@@ -281,8 +278,6 @@ import GHC.TypeLits
     ( Nat, Symbol )
 import Numeric.Natural
     ( Natural )
-import Safe
-    ( readMay )
 import Servant.API
     ( MimeRender (..), MimeUnrender (..), OctetStream )
 import Web.HttpApiData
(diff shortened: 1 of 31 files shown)


lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -976,45 +977,10 @@ instance EncodeAddress n => ToJSON (ApiAddress n) where
...
-        castNumber :: Text -> Aeson.Parser Scientific
-        castNumber txt =
-            case readMay (T.unpack txt) of
-                Nothing ->
-                    fail "expected a number as string with an optional 'H' \
-                         \suffix (e.g. \"1815H\" or \"44\""
-                Just s ->
-                    pure s
+    parseJSON = parseJSON
+        >=> fmap ApiT . eitherToParser . first ShowFmt . fromText
```
KtorZ, line 983: Parsing delegated to the `FromText` instance and moved near the data-type declaration.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -976,45 +977,10 @@ instance EncodeAddress n => ToJSON (ApiAddress n) where
     toJSON = genericToJSON defaultRecordTypeOptions
 
 instance ToJSON (ApiT DerivationIndex) where
-    toJSON (ApiT (DerivationIndex ix))
-        | ix >= firstHardened = toJSON (show (ix - firstHardened) <> "H")
-        | otherwise = toJSON (show ix)
-      where
-        firstHardened = getIndex @'Hardened minBound
-
+    toJSON = toJSON . toText . getApiT
```
KtorZ, line 980: Serialization delegated to the `ToText` instance and moved near the data-type declaration.

specifications/api/swagger.yaml
```diff
@@ -1706,18 +1715,18 @@ x-parametersWalletId: &parametersWalletId
...
+  required: false
+  schema:
+    type: integer
+    minimum: 0
+    maximum: 45_000_000_000_000_000 # 45 B ada (in Lovelace)
+  description: |
+    The stake the user intends to delegate in Lovelace. Required.
 
-      > ⚠️  On the incentivized testnet, this parameter is not requred, but rather completely ignored.
+    > ⚠️  On the incentivized testnet, this parameter is not requred, but rather completely ignored.
```
KtorZ, line 1729: This was somewhat indented too far. Sorry for the noise.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -171,8 +169,8 @@ data Depth = RootK | PurposeK | CoinTypeK | AccountK | RoleK | AddressK
 --
 -- FIXME: rename this to 'Role' or 'HDRole'
 data AccountingStyle
-    = UTxOExternal
-    | UTxOInternal
+    = UtxoExternal
+    | UtxoInternal
```
rvl, line 173: I think this could be fixed in `fromTextToBoundedEnum SnakeLowerCase`.
It's spelt UTxO everywhere else.
This important word deserves special case treatment, so that we don't need to mangle our type names just to get the preferred serialisation.

specifications/api/swagger.yaml
```diff
@@ -2293,8 +2331,33 @@ x-tagGroups:
     - Network
     - Proxy
     - Settings
+    - Experimental
 
 paths:
+  /wallets/{walletId}/{role}/{index}/signatures:
```
rvl, line 2337: :+1:

rvl (approved):
:+1:

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
KtorZ: I thought of both options and thought that it was cleaner in the end to rename the type here. It's perhaps a little less pervasive as a change to just deal with it in the toText//fromText instances. Any way requires a database fix / migration / work-around :|

KtorZ:
bors merge

rvl:
Fixed the integration test - baffling failure message though.

bors r+

KtorZ:
bors retry

PR #2258, opened 2020-10-19
ADP-291: Document endpoints error codes in the API documentation

hasufell:
# Issue Number

ADP-291

# Implementation approach

This is all best-effort. There are no static checks to ensure we didn't miss anything. There's ongoing work at servant to make this possible: https://github.com/haskell-servant/servant/issues/841
But even that would require a major refactor.

Testing all possible errors is also an option, but seems quite excessive.

The workflow for figuring out the error codes is roughly:

1. follow the entry point to the the `Handler ()` function e.g.
```hs
listTransactions
    :: forall ctx s t k n. (ctx ~ ApiLayer s t k)
    => ctx
...
    -> Handler [ApiTransaction n]
```
2. check all `liftHandler` calls, which have functions with `ExceptT` as argument, e.g.:

```hs
listTransactions
    :: forall ctx s k t.
        ( HasDBLayer s k ctx
        , HasNetworkLayer t ctx
        )
    => ctx
...
    -> ExceptT ErrListTransactions IO [TransactionInfo]
```

3. find all the LiftHandler instance, e.g.

```hs
instance LiftHandler ErrListTransactions where
    handler = \case
        ErrListTransactionsNoSuchWallet e -> handler e
        ErrListTransactionsStartTimeLaterThanEndTime e -> handler e
        ErrListTransactionsMinWithdrawalWrong ->
            apiError err400 MinWithdrawalWrong
            "The minimum withdrawal value must be at least 1 Lovelace."
        ErrListTransactionsPastHorizonException e -> handler e
```

4. follow the recursive handlers to resolve all errors and add client error types to `swagger.yaml`. The error code is the 3rd `ApiErrorCode` argument to `apiError`.
5. Some errors are implicit, e.g. failed parameter parsing etc. These are also in `ApiErrorCode`. Also see the special instance `instance LiftHandler (Request, ServerError)`
6. repeat until exhaustion

# Overview

- [x] Wallets
- [x] Addresses
- [x] Coin Selections
- [x] Transactions
- [x] Migrations
- [x] Stake Pools
- [x] Utils
- [x] Network
- [x] Proxy
- [x] Settings
- [ ] Byron-specific endpoints

# Remarks

1. I did not double check the byron endpoints. Many of their endpoints share the same types in `swagger.yaml`, so I assumed they have the same behavior wrt error codes.
2. This will generally be hard to maintain, since yaml anchors aren't enough to express the overlaps of error codes and group them sensibly. It would need something like [dhall](https://github.com/dhall-lang/dhall-lang) to do that.
3. I removed 405 errors, because the HTTP method is part of the very spec. If you follow the spec, you can't get 405.

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -2467,7 +2467,6 @@ data ErrJoinStakePool
     | ErrJoinStakePoolSignDelegation ErrSignDelegation
     | ErrJoinStakePoolSubmitTx ErrSubmitTx
     | ErrJoinStakePoolCannotJoin ErrCannotJoin
-    | ErrJoinStakePoolUnableToAssignInputs CoinSelection
     deriving (Generic, Eq, Show)
 
 data ErrQuitStakePool
@@ -2476,7 +2475,6 @@ data ErrQuitStakePool
     | ErrQuitStakePoolSignDelegation ErrSignDelegation
     | ErrQuitStakePoolSubmitTx ErrSubmitTx
     | ErrQuitStakePoolCannotQuit ErrCannotQuit
-    | ErrQuitStakePoolUnableToAssignInputs CoinSelection
     deriving (Generic, Eq, Show)
 
 -- | Errors that can occur when fetching the reward balance of a wallet
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -2451,7 +2451,6 @@ data ErrStartTimeLaterThanEndTime = ErrStartTimeLaterThanEndTime
 data ErrSelectForDelegation
     = ErrSelectForDelegationNoSuchWallet ErrNoSuchWallet
     | ErrSelectForDelegationFee ErrAdjustForFee
-    | ErrSelectForDelegationUnableToAssignInputs ErrNoSuchWallet
     deriving (Show, Eq)
 
 -- | Errors that can occur when signing a delegation certificate.
--- a/lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE FlexibleContexts #-}
@@ -8,6 +9,7 @@
 {-# LANGUAGE NoMonomorphismRestriction #-}
 {-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE OverloadedLabels #-}
+{-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE QuasiQuotes #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -54,6 +56,7 @@ import Cardano.Wallet.Api.Types
     , ApiCoinSelectionInput (..)
     , ApiCoinSelectionOutput (..)
     , ApiEpochInfo (..)
+    , ApiErrorCode (..)
     , ApiFee (..)
     , ApiMnemonicT (..)
     , ApiNetworkClock (..)
@@ -182,17 +185,28 @@ import Cardano.Wallet.Unsafe
 import Control.Lens
     ( at, (.~), (?~), (^.) )
 import Control.Monad
-    ( forM_, replicateM )
+    ( forM, forM_, replicateM )
 import Control.Monad.IO.Class
     ( liftIO )
 import Crypto.Hash
     ( hash )
 import Data.Aeson
-    ( FromJSON (..), ToJSON (..), (.=) )
+    ( FromJSON (..)
+    , Result (..)
+    , ToJSON (..)
+    , fromJSON
+    , withObject
+    , (.:?)
+    , (.=)
+    )
 import Data.Aeson.QQ
     ( aesonQQ )
 import Data.Char
     ( toLower )
+import Data.Data
+    ( dataTypeConstrs, dataTypeOf, showConstr )
+import Data.Either
+    ( lefts )
 import Data.FileEmbed
     ( embedFile, makeRelativeToProject )
 import Data.Function
@@ -272,6 +286,7 @@ import Test.QuickCheck
     , arbitraryPrintableChar
     , arbitrarySizedBoundedIntegral
     , choose
+    , counterexample
     , elements
     , frequency
     , oneof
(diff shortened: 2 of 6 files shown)


lib/core/src/Cardano/Wallet.hs
```diff
@@ -2261,7 +2261,6 @@ data ErrStartTimeLaterThanEndTime = ErrStartTimeLaterThanEndTime
 data ErrSelectForDelegation
     = ErrSelectForDelegationNoSuchWallet ErrNoSuchWallet
     | ErrSelectForDelegationFee ErrAdjustForFee
-    | ErrSelectForDelegationUnableToAssignInputs ErrNoSuchWallet
```
KtorZ, line 2454: :+1: I guess that's also the case for `ErrJoinStakePoolUnableToAssignInputs` and `ErrQuitStakePoolUnableToAssignInputs`.

specifications/api/swagger.yaml
```diff
@@ -1840,12 +1840,363 @@ x-responsesErr: &responsesErr
...
 
+x-responsesErrWalletNotFound: &responsesErrWalletNotFound
+  type: object
+  required:
+    - message
+    - code
+  properties:
+    message:
+      type: string
+      description: A descriptive error message.
```
KtorZ, line 1924: Instead of giving a description of the what the error message look like, wouldn't it be better to give some details about the circumstances which may lead to such error? e.g.

```
description: | 
  May occur when a given walletId does not match with any known 
  wallets (because it has been deleted, or has never existed).
```

specifications/api/swagger.yaml
```diff
@@ -1840,12 +1840,363 @@ x-responsesErr: &responsesErr
...
+
+x-responsesErrBadRequest: &responsesErrBadRequest
+  type: object
+  required:
+    - message
+    - code
+  properties:
+    message:
+      type: string
+      description: Explanation of a good request.
```
KtorZ: In a similar fashion, something that explains the root cause of the error would be better:

```
description: |
  May occur when a request is not well-formed; that is, it fails to parse
  successfully. This could be the case when some required parameters 
  are missing or, when wrong values are provided.
```

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -917,6 +932,41 @@ spec = do
             in
             x' === x .&&. show x' === show x
 
+    describe "Api Errors" $ do
+        it "Every constructor from ApiErrorCode has a corresponding type in the schema" $
+            let res = fromJSON @SchemaApiErrorCode specification
+                errStr = case res of
+                    Error s -> s
+                    _ -> ""
+            in counterexample errStr $ res == Success SchemaApiErrorCode
```
hasufell, line 953: This shows a nice error message:

```
Cardano.Wallet.Api.Types
  Api Errors
    Every constructor from ApiErrorCode has a corresponding type in the schema FAILED [1]

Failures:

  test/unit/Cardano/Wallet/Api/TypesSpec.hs:937:9:
  1) Cardano.Wallet.Api.Types, Api Errors, Every constructor from ApiErrorCode has a corresponding type in the schema
       Falsifiable (after 1 test):
         Missing ApiErrorCode constructors for: ["RejectedTip","InvalidDelegationDiscovery","NotImplemented","WalletNotResponding","AddressAlreadyExists","UtxoTooSmall","WithdrawalNotWorth","PastHorizon","UnableToAssignInputOutput"]
         Each of these need a corresponding swagger type of the form: x-errConstructorName

  To rerun use: --match "/Cardano.Wallet.Api.Types/Api Errors/Every constructor from ApiErrorCode has a corresponding type in the schema/"

Randomized with seed 1387792531
```

KtorZ (approved):

specifications/api/swagger.yaml
```diff
@@ -1887,12 +1887,505 @@ x-responsesErr: &responsesErr
...
+      type: string
+      enum: ['no_such_transaction']
+
+x-errTransactionNotPending: &errTransactionNotPending
+  <<: *responsesErr
+  title: transaction_not_pending
+  properties:
+    message:
+      type: string
+      description: May occur when trying to forget a transaction that is not pending.
```
piotr-iohk, line 1972: Just to note this will change after https://github.com/input-output-hk/cardano-wallet/pull/2262/files#diff-a2398827a22606930a9b4edc67dfc9c99196a94cdb56c56ef0ef5f92512d42e0R2467 merged. Also expired transaction will be "forgettable".

specifications/api/swagger.yaml
```diff
@@ -2138,12 +2693,37 @@ x-responsesGetTransaction: &responsesGetTransaction
...
-  <<: *responsesErr404
-  <<: *responsesErr405
+  400:
+    description: Bad Request
+    content:
+      application/json:
+        schema:
+          oneOf:
+            - <<: *errBadRequest
+            - <<: *errTransactionIsTooBig
```
piotr-iohk, line 2743: I think this should be only in `403`?
Screenshot from 2020-10-30 11-10-06

specifications/api/swagger.yaml
```diff
@@ -2138,12 +2693,37 @@ x-responsesGetTransaction: &responsesGetTransaction
...
+            - <<: *errWrongEncryptionPassphrase
+  <<: *responsesErr404WalletNotFound
   <<: *responsesErr406
-  <<: *responsesErr415
+  410:
+    description: Gone
+    content:
+      application/json:
+        schema:
+          <<: *errNoSuchWallet
```
piotr-iohk, line 2765: Same `410` and errNoSuchWallet.

specifications/api/swagger.yaml
```diff
@@ -1908,19 +2401,12 @@ x-responsesErr404: &responsesErr404
       application/json:
         schema: *responsesErr
 
-x-responsesErr405: &responsesErr405
```
piotr-iohk, line 1939: Do we think it is irrelevant to highlight in API doc? Basically every endpoint may return this...

specifications/api/swagger.yaml
```diff
@@ -2255,22 +2869,67 @@ x-responsesListStakePools: &responsesListStakePools
...
+            - <<: *errNoSuchWallet
+            - <<: *errNoSuchPool
   <<: *responsesErr406
-  <<: *responsesErr415
+  410:
+    description: Gone
+    content:
+      application/json:
+        schema:
+          <<: *errNoSuchWallet
```
piotr-iohk, line 2896: Same `410` and errNoSuchWallet.

specifications/api/swagger.yaml
```diff
@@ -2065,11 +2582,24 @@ x-responsesGetWalletMigrationInfo: &responsesGetWalletMigrationInfo
...
+            - <<: *errWrongEncryptionPassphrase
+  <<: *responsesErr404WalletNotFound
   <<: *responsesErr406
-  <<: *responsesErr415
+  410:
+    description: Gone
+    content:
+      application/json:
+        schema:
+          <<: *errNoSuchWallet
```
piotr-iohk, line 2601: This doesn't feel right to me. I think we should use `404` on 'no such wallet'. Especially looking at some endpoints where:

Screenshot from 2020-10-30 11-25-22

specifications/api/swagger.yaml
```diff
@@ -2255,22 +2869,67 @@ x-responsesListStakePools: &responsesListStakePools
...
+            - <<: *errNotDelegatingTo
+            - <<: *errNonNullRewards
+  <<: *responsesErr404WalletNotFound
+  <<: *responsesErr406
+  410:
+    description: Gone
+    content:
+      application/json:
+        schema:
+          <<: *errNoSuchWallet
```
piotr-iohk, line 2924: Same `410` and errNoSuchWallet.

piotr-iohk (commented):
Good work on documenting error messages. I've left few comments.

specifications/api/swagger.yaml
hasufell: I don't think so. The spec includes which method to use. If you follow the spec, 405 will never be returned.

specifications/api/swagger.yaml
hasufell: This PR isn't really about refactoring or fixing our error codes though. I just added documentation to the existing error handlers. I changed nothing.

specifications/api/swagger.yaml
hasufell:
```
        ErrSelectForPaymentTxTooLarge maxSize maxN  ->
            apiError err400 TransactionIsTooBig $ mconcat
                [ "I am afraid that the transaction you're trying to submit is "
                , "too large! The network allows transactions only as large as "
                , pretty maxSize, "s! As it stands, the current transaction only "
                , "allows me to select up to ", showT maxN, " inputs. Note "
                , "that I am selecting inputs randomly, so retrying *may work* "
                , "provided I end up choosing bigger inputs sufficient to cover "
                , "the transaction cost. Alternatively, try sending to less "
                , "recipients or with smaller metadata."
                ]
```
it is 400

specifications/api/swagger.yaml
hasufell: It is 410

```
        ErrSignPaymentNoSuchWallet e -> (handler e)
            { errHTTPCode = 410
            , errReasonPhrase = errReasonPhrase err410
            }
```

specifications/api/swagger.yaml
hasufell:
```
        ErrSignPaymentNoSuchWallet e -> (handler e)
            { errHTTPCode = 410
            , errReasonPhrase = errReasonPhrase err410
            }
```

specifications/api/swagger.yaml
hasufell:
```
        ErrSignPaymentNoSuchWallet e -> (handler e)
            { errHTTPCode = 410
            , errReasonPhrase = errReasonPhrase err410
            }
```

specifications/api/swagger.yaml
piotr-iohk: Then `406` and `415` should be removed also I guess, no?

specifications/api/swagger.yaml
piotr-iohk: ACK. It revealed those small inconsistencies though (I spotted 410 and 403 only). 
Could be fixed in this go if it's not a lot of work, but don't have to of course.

specifications/api/swagger.yaml
hasufell: 415 maybe, but the swagger spec doesn't document enough for 406 to not be a concern afais.

PR #2260, opened 2020-10-21
Add N={20,40,80} wallet restore benchmarks

Anviking:
# Issue Number

ADP-469

# Overview

- Add N={20,40,80} wallet restore benchmarks, to have more data points to see if it is linear or not.

# Comments

Buildkite nightly build: https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/685

### DIFF
--- a/lib/shelley/bench/Restore.hs
+++ b/lib/shelley/bench/Restore.hs
@@ -287,6 +287,9 @@ cardanoRestoreBench tr c socketFile = do
           -- to-100% time.
           benchRestoreMultipleWallets 1 (unsafeMkPercentage 0.1)
         , benchRestoreMultipleWallets 10 (unsafeMkPercentage 0.01)
+        , benchRestoreMultipleWallets 20 (unsafeMkPercentage 0.01)
+        , benchRestoreMultipleWallets 40 (unsafeMkPercentage 0.01)
+        , benchRestoreMultipleWallets 80 (unsafeMkPercentage 0.01)
         , benchRestoreMultipleWallets 100 (unsafeMkPercentage 0.01)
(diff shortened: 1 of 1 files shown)


lib/shelley/bench/Restore.hs
```diff
@@ -287,6 +287,9 @@ cardanoRestoreBench tr c socketFile = do
           -- to-100% time.
           benchRestoreMultipleWallets 1 (unsafeMkPercentage 0.1)
         , benchRestoreMultipleWallets 10 (unsafeMkPercentage 0.01)
+        , benchRestoreMultipleWallets 20 (unsafeMkPercentage 0.01)
+        , benchRestoreMultipleWallets 40 (unsafeMkPercentage 0.01)
+        , benchRestoreMultipleWallets 80 (unsafeMkPercentage 0.01)
         , benchRestoreMultipleWallets 100 (unsafeMkPercentage 0.01)
```
KtorZ, line 293: What about removing 100, so we keep a nice x2 factor between each bench.

jonathanknowles: Or make it 160? :smile_cat:  (Perhaps that would be **_too_** long.)

Anviking: By keeping 100 we can still reason about 1 vs 10 vs 100.

paweljakubas (approved):

Anviking:
(ty!) 
bors r+

Anviking:
Plot + regression with some recent data from CI:
Skärmavbild 2020-10-23 kl  14 22 56

PR #2266, opened 2020-10-23
add Keys to tags in swagger

paweljakubas:
# Issue Number

https://jira.iohk.io/browse/ADP-474

# Overview

- [x] I have make Keys added to tags

# Comments


piotr-iohk (approved):

piotr-iohk:
@paweljakubas it goes to Misc now:
Screenshot from 2020-10-23 13-21-16


I wonder if it won't be better to put it underneath Shelley wallets, since it is `GET /wallets/{walletId}/keys/{role}/{index}`. What do you think?

PR #2249, opened 2020-10-15
ADP-478: Garbage collect delisted stake pools from SMASH

hasufell:
This is the first step for garbage collecting
stake pools based on SMASH delisting.

X-JIRA-Ticket: ADP-478

----

## Questions / Considerations

1. After looking at the logic, I believe we don't have to adjust any other functions of the DBLayer such as `listRegisteredPools` or `readPoolProduction`. These can still consider all pools. Only the API layer will consider delisted pools and adjust `ApiStakePool` accordingly.
2. There's still the open question whether we want to *really* delete GCed pools via `removePools` at some point.

----

### Feature

- [x] Add `delisted` column to metadata table and populate Pool DBLayer with functions
- [x] Create new `internal_state` table and have GC thread store last sync time
- [x] Add garbage collection thread querying the SMASH server for delisted pools and calling `delistPools` from DBLayer
- [x] have GC thread store last sync time
- [x] return last GC sync time
- [x] Add `POST '{ "maintenance_action": "gc_stake_pools" }' /stake-pools` endpoint

### QA

- [x] database tests
- [x] check json roundtripping works
- [ ] Add integration tests? We don't have SMASH server integration tests yet.

PR #2277, opened 2020-10-28
Record delisted pools in a dedicated table.

jonathanknowles:
# Issue Number

#2249 

# Overview

This PR records delisted pools in a dedicated table instead of using a field in the `pool_registrations` table.

In the updated schema, a pool is delisted if (and only if) there is a single row containing that pool's id in the `delisted_pools` table.

This solution has several advantages:

1.  We only need a single database row to record that a pool is delisted.

2.  We no longer need to carefully to ensure that all registration records for a particular pool have the same delisted status. A pool is either _delisted_ or _not delisted_: the schema rules out all intermediate states.

3.  Pools automatically remain delisted when rollbacks occur or when new certificates are published, with no special effort.

4.  The `putPoolRegistration` function no longer needs to read the  most-recently-written registration entry before adding a new entry.

5.  Each row in the `pool_registrations` table is now just an immutable record of a registration certificate.

### DIFF
--- a/lib/core/src/Cardano/Pool/DB/Model.hs
+++ b/lib/core/src/Cardano/Pool/DB/Model.hs
@@ -432,7 +432,7 @@ mRollbackTo ti point = do
         | otherwise = Nothing
 
 mDelistPools :: [PoolId] -> ModelOp ()
-mDelistPools = modify #delisted . Set.union . Set.fromList
+mDelistPools = modify #delisted . const . Set.fromList
 
 mReadDelistedPools :: ModelOp [PoolId]
 mReadDelistedPools = Set.toList <$> get #delisted
--- a/lib/core/src/Cardano/Pool/DB/Model.hs
+++ b/lib/core/src/Cardano/Pool/DB/Model.hs
@@ -62,6 +62,7 @@ module Cardano.Pool.DB.Model
     , mReadCursor
     , mRemovePools
     , mDelistPools
+    , mReadDelistedPools
     , mRemoveRetiredPools
     , mReadSettings
     , mPutSettings
@@ -80,7 +81,6 @@ import Cardano.Wallet.Primitive.Types
     , CertificatePublicationTime
     , EpochNo (..)
     , InternalState (..)
-    , PoolFlag (..)
     , PoolId
     , PoolLifeCycleStatus (..)
     , PoolOwner (..)
@@ -116,6 +116,8 @@ import Data.Ord
     ( Down (..) )
 import Data.Quantity
     ( Quantity (..) )
+import Data.Set
+    ( Set )
 import Data.Time.Clock.POSIX
     ( POSIXTime )
 import Data.Word
@@ -152,6 +154,8 @@ data PoolDatabase = PoolDatabase
         !(Map (CertificatePublicationTime, PoolId) PoolRetirementCertificate)
     -- ^ On-chain retirements associated with pools
 
+    , delisted :: !(Set PoolId)
+
     , metadata :: !(Map StakePoolMetadataHash StakePoolMetadata)
     -- ^ Off-chain metadata cached in database
 
@@ -184,9 +188,9 @@ instance Eq SystemSeed where
 
 -- | Produces an empty model pool production database.
 emptyPoolDatabase :: PoolDatabase
-emptyPoolDatabase =
-    PoolDatabase mempty mempty mempty mempty mempty mempty mempty NotSeededYet
-        mempty defaultSettings defaultInternalState
+emptyPoolDatabase = PoolDatabase
+    mempty mempty mempty mempty mempty mempty mempty mempty NotSeededYet
+    mempty defaultSettings defaultInternalState
 
 {-------------------------------------------------------------------------------
                                   Model Operation Types
@@ -254,12 +258,10 @@ mPutPoolRegistration
     -> PoolRegistrationCertificate
     -> ModelOp ()
 mPutPoolRegistration cpt cert = do
-    old <- fmap snd <$> mReadPoolRegistration (view #poolId cert)
-    let flag = maybe NoPoolFlag poolFlag old
     modify #owners
         $ Map.insert poolId poolOwners
     modify #registrations
-        $ Map.insert (cpt, poolId) (cert { poolFlag = flag })
+        $ Map.insert (cpt, poolId) cert
   where
     PoolRegistrationCertificate {poolId, poolOwners} = cert
 
@@ -430,16 +432,10 @@ mRollbackTo ti point = do
         | otherwise = Nothing
 
 mDelistPools :: [PoolId] -> ModelOp ()
-mDelistPools poolsToDelist =
-    modify #registrations
-        $ Map.mapWithKey
-        $ \(_, pid) a ->
-            if updateThis pid
-            then a {poolFlag = Delisted}
-            else a
-  where
-    updateThis p = p `Set.member` poolsToDelistSet
-    poolsToDelistSet = Set.fromList poolsToDelist
+mDelistPools = modify #delisted . Set.union . Set.fromList
+
+mReadDelistedPools :: ModelOp [PoolId]
+mReadDelistedPools = Set.toList <$> get #delisted
 
 mRemovePools :: [PoolId] -> ModelOp ()
 mRemovePools poolsToRemove = do
@@ -453,6 +449,8 @@ mRemovePools poolsToRemove = do
         $ Map.filterWithKey $ \(_, p) _ -> retain p
     modify #retirements
         $ Map.filterWithKey $ \(_, p) _ -> retain p
+    modify #delisted
+        $ Set.filter retain
   where
     retain p = p `Set.notMember` poolsToRemoveSet
     poolsToRemoveSet = Set.fromList poolsToRemove
--- a/lib/core/src/Cardano/Pool/DB/Model.hs
+++ b/lib/core/src/Cardano/Pool/DB/Model.hs
@@ -51,6 +51,7 @@ module Cardano.Pool.DB.Model
     , mPutPoolRetirement
     , mReadPoolRetirement
     , mUnfetchedPoolMetadataRefs
+    , mPutDelistedPools
     , mPutFetchAttempt
     , mPutPoolMetadata
     , mListPoolLifeCycleData
@@ -61,7 +62,6 @@ module Cardano.Pool.DB.Model
     , mRollbackTo
     , mReadCursor
     , mRemovePools
-    , mDelistPools
     , mReadDelistedPools
     , mRemoveRetiredPools
     , mReadSettings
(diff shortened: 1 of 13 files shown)


lib/core/src/Cardano/Pool/DB/Model.hs
```diff
@@ -152,6 +154,8 @@ data PoolDatabase = PoolDatabase
         !(Map (CertificatePublicationTime, PoolId) PoolRetirementCertificate)
     -- ^ On-chain retirements associated with pools
 
+    , delisted :: !(Set PoolId)
```
hasufell, line 157: I wonder if we should just enable `StrictData`, but that probably warrants a discussion.

lib/core/src/Cardano/Pool/DB/Model.hs
```diff
@@ -430,16 +432,10 @@ mRollbackTo ti point = do
...
-    modify #registrations
-        $ Map.mapWithKey
-        $ \(_, pid) a ->
-            if updateThis pid
-            then a {poolFlag = Delisted}
-            else a
-  where
-    updateThis p = p `Set.member` poolsToDelistSet
-    poolsToDelistSet = Set.fromList poolsToDelist
+mDelistPools = modify #delisted . Set.union . Set.fromList
```
hasufell: Ok, so I think we don't want union here! The SMASH server has an endpoint to enlist already delisted pools afais:

https://github.com/input-output-hk/smash/blob/f4a179a1e2de22d332ec731df8fedaed5d7d8754/src/Cardano/SMASH/API.hs#L60

That means we really want to overwrite here, not merge.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -500,13 +491,11 @@ newDBLayer trace fp timeInterpreter = do
...
 
-        delistPools pools = do
-            px <- selectList
-                [ PoolRegistrationPoolId <-. pools ]
-                [  ]
-            repsertMany $ fmap
-                (\(Entity k val) -> (k, val {poolRegistrationFlag = Delisted}))
-                px
+        delistPools =
+            insertMany_ . fmap DelistedPool
```
hasufell: Same as above. We need to drop the table beforehand.

lib/core/src/Cardano/Pool/DB/Sqlite/TH.hs
```diff
@@ -120,11 +120,15 @@ PoolRegistration sql=pool_registration
     poolRegistrationPledge            Word64                        sql=pledge
     poolRegistrationMetadataUrl       W.StakePoolMetadataUrl  Maybe sql=metadata_url
     poolRegistrationMetadataHash      W.StakePoolMetadataHash Maybe sql=metadata_hash
-    poolRegistrationFlag              W.PoolFlag                    sql=flag
 
     Primary poolRegistrationPoolId poolRegistrationSlot poolRegistrationSlotInternalIndex
     deriving Show Generic
 
+DelistedPool sql=delisted_pool
```
hasufell: I'd propose to call the table `pool_delisted` or `pool_delistment` (is that even a word?). The other tables also start with `pool_`.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -2042,14 +2039,3 @@ instance FromJSON PoolMetadataSource where
 
 instance ToJSON PoolMetadataSource where
     toJSON = toJSON . toText
-
-data PoolFlag = NoPoolFlag | Delisted
```
hasufell, line 2046: One naming discussion less :)

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
```diff
@@ -1513,16 +1511,15 @@ prop_delistPoolsPersists DBLayer {..} cert =
         let poolid = view #poolId . snd $ cert
         -- delist pool
         run $ atomically $ delistPools [poolid]
-        delisted <- run . atomically . readPoolRegistration $ poolid
```
hasufell, line 1516: This whole test doesn't make sense anymore. Because the delisted flag doesn't exist in the table in question. So it doesn't test anything.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -560,6 +562,7 @@ readPoolDbData DBLayer {..} currentEpoch = atomically $ do
         retirementCertificates
         <$> readTotalProduction
         <*> readPoolMetadata
+        <*> (Set.fromList <$> readDelistedPools)
```
hasufell, line 564: Maybe we just pass it through as a Set? There's json instances for it anyway.

hasufell (changes requested):
Looks good, just a few remarks

lib/core/src/Cardano/Pool/DB/Model.hs
hasufell: And we should also test this property. `delistPools` overwrites, it doesn't add.

lib/core/src/Cardano/Pool/DB/Sqlite/TH.hs
hasufell: That might also fix the naming issue below

lib/core/src/Cardano/Pool/DB/Model.hs
jonathanknowles: We could. Though I agree we'd have to make it a wider discussion.

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
jonathanknowles: I agree that we could delete this test.

Though given your [earlier comment](https://github.com/input-output-hk/cardano-wallet/pull/2277#discussion_r513384085) about completely overwriting the set each time we do a `delistPools`, I think it would make sense to update the test so that it checks that the set is replaced, rather than augmented.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
jonathanknowles: By passing it through as a `Set`, do you mean writing a set to a single row?

I think that would make it harder to inspect the table using tools like `sqlitebrowser`.

In any case, having one row per pool ID means that we can more easily write queries on the table, especially if we need to perform a join elsewhere.

lib/core/src/Cardano/Pool/DB/Model.hs
jonathanknowles: Aha, I didn't realize that `delistPools` was supposed to completely replace the set each time.

In that case, I think the `delistPools` name is a bit confusing, as it sounds as though it ought to augment the existing set, rather than replacing it.

Perhaps call the functions `putDelistedPools` and `readDelistedPools`?

lib/core/src/Cardano/Pool/DB/Sqlite/TH.hs
jonathanknowles: Fixed in 35e29f854a2769a659f75903b56bc3ad4a2c7146.

lib/core/src/Cardano/Pool/DB/Model.hs
jonathanknowles: > That means we really want to overwrite here, not merge.

Fixed in b67c89fd8aed7b2349dd18ac4a0f8ed9e981823b.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
jonathanknowles: > Same as above. We need to drop the table beforehand.

Fixed in b67c89fd8aed7b2349dd18ac4a0f8ed9e981823b.

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
hasufell: Sure, but that's a different property. This test was about "persistence in light of certificate insertion".

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
hasufell: Nvm... it's fine this way

lib/core/src/Cardano/Pool/DB/Model.hs
jonathanknowles: > And we should also test this property. delistPools overwrites, it doesn't add.

Fixed in b3f3d7ed45cde2a603b9fe26a6a916c89fd97132.

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
jonathanknowles: > Sure, but that's a different property. This test was about "persistence in light of certificate insertion".

I've now deleted this test.

Fixed in c94d57890fd9ce1099e06731a0e66f2bce6bde63.

lib/core/src/Cardano/Pool/DB/Model.hs
```diff
@@ -453,6 +449,8 @@ mRemovePools poolsToRemove = do
         $ Map.filterWithKey $ \(_, p) _ -> retain p
     modify #retirements
         $ Map.filterWithKey $ \(_, p) _ -> retain p
+    modify #delisted
```
hasufell: Do we really want this? I'd say that the pool delisted set should just reflect what SMASH returns, without any local modifications?

lib/core/src/Cardano/Pool/DB/Sqlite.hs
```diff
@@ -515,6 +505,7 @@ newDBLayer trace fp timeInterpreter = do
             deleteWhere [ PoolRegistrationPoolId ==. pool ]
             deleteWhere [ PoolRetirementPoolId ==. pool ]
             deleteWhere [ StakeDistributionPoolId ==. pool ]
+            deleteWhere [ DelistedPoolId ==. pool ]
```
hasufell: Same as above.

lib/core/src/Cardano/Pool/DB/Model.hs
hasufell: Also note: the next query to SMASH will add the just deleted poolIds again.

jonathanknowles: Well, it all depends on whether you want the set of pool IDs in `pool_delistment` to always be a subset of the pools that we know about.

I'd say it's probably okay to remove them when we remove pools.

hasufell: But we're not yet satisfying that property, see

https://github.com/input-output-hk/cardano-wallet/pull/2277/files#diff-7221813f41d4c796791728852d1b5e3f2007929b3c613e3d07dda5484d11d486R494-R496

jonathanknowles: > Do we really want this? I'd say that the pool delisted set should just reflect what SMASH returns, without any local modifications?

Fair enough. I've fixed this in 12f5426917b675d1c9ee01fc2f48510d95ecf422.

lib/core/src/Cardano/Pool/DB/Sqlite.hs
jonathanknowles: Fixed in 12f5426917b675d1c9ee01fc2f48510d95ecf422.

hasufell (approved):

Issue #2279, opened 2020-10-29
Hydra being slow to be scheduled causing bors to timeout


Anviking:
# Context
 
 

Buildkite has a timeout. Hydra has timeouts. Bors waits on both of them to pass. Bors also has a timeout.

# Test Case

N/A

# Failure / Counter-example

https://github.com/input-output-hk/cardano-wallet/pull/2235#issuecomment-707697053

---

# Resolution 

---

# QA

Anviking:
Stale, closing

PR #2282, opened 2020-10-30
Increase bors timeout from 2h to 4h

Anviking:
# Issue Number

#2279 

# Overview

- [x] Increase bors timeout from 2h to 3h

# Comments

It seems hydra is often slow to schedule/run the jobs, causing ≈18% of bors r+ to fail.

It should be better to increase the timeout to 3h, than to have it fail.

This doesn't affect the timeout of buildkite or hydra themselves.

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -2,6 +2,6 @@ status = [
   "buildkite/cardano-wallet",
   "ci/hydra-build:required",
 ]
-timeout_sec = 7200
+timeout_sec = 14400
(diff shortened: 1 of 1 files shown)


rvl (approved):

Anviking:
I guess we'll do 4h instead 🤷‍♂️ 
bors r+

In the above build, two hydra jobs are still building after 3h and 20 min. And done after 3h and 35m

PR #2285, opened 2020-10-30
Reject SMASH URIs with paths

hasufell:
# Issue Number

ADP-502

# Overview

* The idea was to use the servant types of SMASH, but that's pending due to https://github.com/input-output-hk/smash/issues/110

# Comments

This was tested locally against testnet.

The sqlite serialisation is a little more permissive.


KtorZ (approved):
Looks right, although I believe is a breaking change with regard to the command-line arguments. cardano-launcher and the Daedalus config will need adjustments.

hasufell:
> Looks right, although I believe is a breaking change with regard to the command-line arguments. cardano-launcher and the Daedalus config will need adjustments.

Yep

PR #2286, opened 2020-10-30
Add health check

hasufell:
# Issue Number

ADP-496

## Details

- [x] Verify url and SMASH health on app start
- [x] Verify url and SMASH health on settings change
- [x] Expose SMASH health check proxy endpoints

### DIFF
--- a/lib/core/src/Cardano/Pool/Metadata.hs
+++ b/lib/core/src/Cardano/Pool/Metadata.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE Rank2Types #-}
+{-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE TypeApplications #-}
 
 -- |
@@ -19,6 +20,10 @@ module Cardano.Pool.Metadata
       fetchFromRemote
     , StakePoolMetadataFetchLog (..)
     , fetchDelistedPools
+    , healthCheck
+    , isHealthyStatus
+    , toHealthCheckSMASH
+    , HealthStatusSMASH (..)
 
     -- * Construct URLs
     , identityUrlBuilder
@@ -39,11 +44,11 @@ import Cardano.BM.Data.Severity
 import Cardano.BM.Data.Tracer
     ( HasPrivacyAnnotation (..), HasSeverityAnnotation (..) )
 import Cardano.Wallet.Api.Types
-    ( defaultRecordTypeOptions )
+    ( HealthCheckSMASH (..), HealthStatusSMASH (..), defaultRecordTypeOptions )
 import Cardano.Wallet.Primitive.AddressDerivation
     ( hex )
 import Cardano.Wallet.Primitive.Types
-    ( PoolId
+    ( PoolId (..)
     , StakePoolMetadata (..)
     , StakePoolMetadataHash (..)
     , StakePoolMetadataUrl (..)
@@ -62,12 +67,14 @@ import Control.Tracer
 import Crypto.Hash.Utils
     ( blake2b256 )
 import Data.Aeson
-    ( FromJSON (..)
-    , ToJSON (..)
+    ( FromJSON
+    , ToJSON
     , eitherDecodeStrict
     , fieldLabelModifier
     , genericParseJSON
     , genericToJSON
+    , parseJSON
+    , toJSON
     )
 import Data.Bifunctor
     ( first )
@@ -118,6 +125,13 @@ metadaFetchEp pid (StakePoolMetadataHash bytes)
   where
     hashStr = T.unpack $ T.decodeUtf8 $ convertToBase Base16 bytes
     pidStr  = T.unpack $ toText pid
+
+-- TODO: use SMASH servant types
+healthCheckEP :: String
+healthCheckEP = T.unpack $ T.intercalate "/" ["api", "v1", "status"]
+
+delistedEP :: String
+delistedEP = T.unpack $ T.intercalate "/" ["api", "v1", "delisted"]
 
 -- | TODO: import SMASH types
 newtype SMASHPoolId = SMASHPoolId
@@ -173,20 +187,18 @@ registryUrlBuilder baseUrl pid _ hash =
         { uriPath = "/" <> metadaFetchEp pid hash
         }
 
-fetchDelistedPools
+-- | A smash GET request that reads the result at once into memory.
+smashRequest
     :: Tracer IO StakePoolMetadataFetchLog
     -> URI
     -> Manager
-    -> IO (Maybe [PoolId])
-fetchDelistedPools tr uri manager = runExceptTLog $ do
-    pl <- getPoolsPayload
-    smashPids <- except $ eitherDecodeStrict @[SMASHPoolId] pl
-    forM smashPids $ except . first getTextDecodingError . toPoolId
+    -> ExceptT String IO ByteString
+smashRequest tr uri manager = getPayload
   where
-    getPoolsPayload :: ExceptT String IO ByteString
-    getPoolsPayload = do
+    getPayload :: ExceptT String IO ByteString
+    getPayload = do
         req <- withExceptT show $ except $ requestFromURI uri
-        liftIO $ traceWith tr $ MsgFetchDelistedPools uri
+        liftIO $ traceWith tr $ MsgFetchSMASH uri
         ExceptT
             $ handle fromIOException
             $ handle fromHttpException
@@ -202,6 +214,57 @@ fetchDelistedPools tr uri manager = runExceptTLog $ do
                 , show s
                 ]
 
+    fromHttpException :: Monad m => HttpException -> m (Either String a)
+    fromHttpException = return . Left . ("HTTP exception: " <>) . show
+
+-- | Gets the health status from the SMASH server. Returns
+-- @Nothing@ if the server is unreachable.
+healthCheck
+    :: Tracer IO StakePoolMetadataFetchLog
+    -> URI
+    -> Manager
+    -> IO (Maybe HealthStatusSMASH)
+healthCheck tr uri manager = runExceptTLog $ do
+    pl <- smashRequest tr
+        (uri { uriPath = "/" <> healthCheckEP , uriQuery = "", uriFragment = "" })
+        manager
+    except . eitherDecodeStrict @HealthStatusSMASH $ pl
+  where
+    runExceptTLog
+        :: ExceptT String IO HealthStatusSMASH
+        -> IO (Maybe HealthStatusSMASH)
+    runExceptTLog action = runExceptT action >>= \case
+        Left msg ->
+            Nothing <$ traceWith tr (MsgFetchHealthCheckFailure msg)
+        Right health -> do
+            traceWith tr (MsgFetchHealthCheckSuccess health)
+            pure $ Just health
+
+-- | Convert the result of @healthCheck@, which represents the
+-- server response to our own @HealthCheckSMASH@ type, which is a
+-- superset of it.
+toHealthCheckSMASH :: Maybe HealthStatusSMASH -> HealthCheckSMASH
+toHealthCheckSMASH = \case
+    (Just health)
+        | isHealthyStatus health -> Available
+        | otherwise -> Unavailable
+    _ -> Unreachable
+
+isHealthyStatus :: HealthStatusSMASH -> Bool
+isHealthyStatus (HealthStatusSMASH {..}) = T.toLower status == "ok"
+
+fetchDelistedPools
+    :: Tracer IO StakePoolMetadataFetchLog
+    -> URI
+    -> Manager
+    -> IO (Maybe [PoolId])
+fetchDelistedPools tr uri manager = runExceptTLog $ do
+    pl <- smashRequest tr
+        (uri { uriPath = "/" <> delistedEP , uriQuery = "", uriFragment = "" })
+        manager
+    smashPids <- except $ eitherDecodeStrict @[SMASHPoolId] pl
+    forM smashPids $ except . first getTextDecodingError . toPoolId
+  where
     runExceptTLog
         :: ExceptT String IO [PoolId]
         -> IO (Maybe [PoolId])
@@ -211,9 +274,6 @@ fetchDelistedPools tr uri manager = runExceptTLog $ do
 
         Right meta ->
             Just meta <$ traceWith tr (MsgFetchDelistedPoolsSuccess meta)
-
-    fromHttpException :: Monad m => HttpException -> m (Either String a)
-    fromHttpException = return . Left . ("HTTP exception: " <>) . show
 
 -- TODO: refactor/simplify this
 fetchFromRemote
@@ -301,14 +361,17 @@ fetchFromRemote tr builders manager pid url hash = runExceptTLog $ do
 
 fromIOException :: Monad m => IOException -> m (Either String a)
 fromIOException = return . Left . ("IO exception: " <>) . show
+
 data StakePoolMetadataFetchLog
     = MsgFetchPoolMetadata StakePoolMetadataHash URI
     | MsgFetchPoolMetadataSuccess StakePoolMetadataHash StakePoolMetadata
     | MsgFetchPoolMetadataFailure StakePoolMetadataHash String
     | MsgFetchPoolMetadataFallback URI Bool
-    | MsgFetchDelistedPools URI
+    | MsgFetchSMASH URI
     | MsgFetchDelistedPoolsFailure String
     | MsgFetchDelistedPoolsSuccess [PoolId]
+    | MsgFetchHealthCheckFailure String
+    | MsgFetchHealthCheckSuccess HealthStatusSMASH
     deriving (Show, Eq)
 
 instance HasPrivacyAnnotation StakePoolMetadataFetchLog
@@ -318,9 +381,11 @@ instance HasSeverityAnnotation StakePoolMetadataFetchLog where
         MsgFetchPoolMetadataSuccess{} -> Info
         MsgFetchPoolMetadataFailure{} -> Warning
         MsgFetchPoolMetadataFallback{} -> Warning
-        MsgFetchDelistedPools{} -> Info
+        MsgFetchSMASH{} -> Debug
         MsgFetchDelistedPoolsFailure{} -> Warning
         MsgFetchDelistedPoolsSuccess{} -> Info
+        MsgFetchHealthCheckFailure{} -> Warning
+        MsgFetchHealthCheckSuccess{} -> Info
 
 instance ToText StakePoolMetadataFetchLog where
     toText = \case
@@ -341,8 +406,8 @@ instance ToText StakePoolMetadataFetchLog where
                 then ""
                 else " Falling back using a different strategy."
             ]
-        MsgFetchDelistedPools uri -> mconcat
-            [ "Fetching delisted pools from ", T.pack (show uri)
+        MsgFetchSMASH uri -> mconcat
+            [ "Making a SMASH request to ", T.pack (show uri)
             ]
         MsgFetchDelistedPoolsSuccess poolIds -> mconcat
             [ "Successfully fetched delisted "
@@ -352,3 +417,11 @@ instance ToText StakePoolMetadataFetchLog where
         MsgFetchDelistedPoolsFailure err -> mconcat
             [ "Failed to fetch delisted pools: ", T.pack err
             ]
+        MsgFetchHealthCheckSuccess health -> mconcat
+            [ "Successfully checked health "
+            , T.pack (show health)
+            ]
+        MsgFetchHealthCheckFailure err -> mconcat
+            [ "Failed to check health: ", T.pack err
+            ]
+
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -150,8 +150,11 @@ module Cardano.Wallet.Api.Types
     , PostTransactionFeeDataT
     , ApiWalletMigrationPostDataT
 
-    -- * other
+    -- * Others
     , defaultRecordTypeOptions
+    , HealthStatusSMASH (..)
+    , HealthCheckSMASH (..)
+    , ApiHealthCheck (..)
     ) where
 
 import Prelude
@@ -205,6 +208,7 @@ import Cardano.Wallet.Primitive.Types
     , SlotLength (..)
     , SlotNo (..)
     , SlottingParameters (..)
+    , SmashServer (..)
     , StakePoolMetadata
     , StartTime (..)
     , WalletBalance (..)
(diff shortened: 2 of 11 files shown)


lib/shelley/src/Cardano/Wallet/Shelley.hs
```diff
@@ -351,6 +360,22 @@ serveWallet
                 $ \db@DBLayer{..} -> do
 
             forM_ settings $ atomically . putSettings
+            dbSettings <- atomically readSettings
+            -- initially verity smash server
+            case poolMetadataSource dbSettings of
```
hasufell: This is more or less on application start. We crash the app if the server isn't reachable.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -688,12 +693,24 @@ monitorMetadata
 monitorMetadata tr gp DBLayer{..} = do
     settings <- atomically readSettings
     manager <- newManager defaultManagerSettings
+
+    case poolMetadataSource settings of
```
hasufell: This usually triggers when something changes the smash setting after the server started. If the server isn't healthy, we crash the worker thread.

The problem is... the caller doesn't really know, but they should! I don't think it's enough to emit warnings here, because subsequent fetch attempts will just fail forever. We probably need an API to query the status of the metadata worker.

rvl (commented):
It's not clear from the PR description what this health check stuff is supposed to do. From all the discussion on that jira ticket, what did you end up deciding?

Can I suggest that a more user friendly approach would be to not crash when the status check fails? Perhaps log it at warning level and carry on. The smash server may be intermittently unreachable, but that's no reason to prevent starting the wallet.

If Daedalus want to prevent users from setting an incorrect smash url, and go back to the previous value if the new one was wrong, then perhaps provide an endpoint where they can do the status check _before_ changing the setting.

For example, a new endpoint `GET /v2/network/information/smash?url=https://smash.awstest.iohkdev.io`. (If the `url` query param is missing then it health checks the currently configured smash server.)

hasufell:
@rvl What about just requiring deadalus to do the health check on their end? It isn't complicated and they have more ways to react to incorrect input than we have. I think it's indeed better the wallet just logs this.

rvl:
It's possible. All Daedalus would need to know is whether a given smash url works - yes/no.
They may prefer better separation of concerns -- or may not care.
We can also basically proxy the smash health check, which is easy enough.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -740,6 +740,21 @@ monitorMetadata gcStatus tr sp db@(DBLayer{..}) = do
...
+        FetchSMASH uri -> do
+            let loop = do
+                    r <- healthCheck (Just trFetch) (unSmashServer uri) manager
+                    case r of
+                        (Just health)
+                            | isHealthy health -> pure ()
+                        _ -> do
+                            let ms = (* 1_000_000)
+                            let sleepTime = ms 60
```
hasufell: 60s is maybe a bit excessive?

hasufell:
Should be all set

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -2155,3 +2158,48 @@ type instance PostTransactionFeeDataT (n :: NetworkDiscriminant) =
...
+            "unavailable" -> pure Unavailable
+            "unreachable" -> pure Unreachable
+            "no_smash_configured" -> pure NoSMASH
+            e -> fail ("Unexpeced value: " <> e)
+
+instance ToJSON ApiHealthCheck where
+    toJSON Available = String "available"
+    toJSON Unavailable = String "unavailable"
+    toJSON Unreachable = String "unreachable"
+    toJSON NoSMASH = String "no_smash_configured"
```
KtorZ: Both instances would be better off derived using generics. We already have all the necessary plumbing for this. Please rewrite this using generic JSON. You'll need to change `NoSMASH` to `NoSmashConfigured`.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -2155,3 +2158,48 @@ type instance PostTransactionFeeDataT (n :: NetworkDiscriminant) =
...
+
+
+{-------------------------------------------------------------------------------
+                         SMASH interfacing types
+-------------------------------------------------------------------------------}
+
+-- | Parses the SMASH HealthCheck type from the SMASH API.
+data HealthCheck = HealthCheck
+    { status :: T.Text
+    , version :: T.Text
```
KtorZ: Do not use qualified import for `Text`, just `Text` is fine as per our coding style guide.

lib/shelley/src/Cardano/Wallet/Shelley.hs
```diff
@@ -427,13 +431,19 @@ instance ToText ApplicationLog where
...
+            , T.pack $ uriToString id uri ""
+            ]
 
 instance HasPrivacyAnnotation ApplicationLog
 instance HasSeverityAnnotation ApplicationLog where
     getSeverityAnnotation = \case
         MsgStarting _ -> Info
         MsgNetworkName _ -> Info
         MsgServerStartupError _ -> Alert
+        MsgFailedConnectSMASH _ -> Error
```
KtorZ: I don't think this deserve anything higher than `Warning`. A `Notice` will be fine. It's nothing wrong with the wallet application per se, and it won't prevent the wallet from operating other functionalities correctly.

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
```diff
@@ -427,6 +434,23 @@ server byron icarus shelley spl ntp =
...
             = Handler $ fmap ApiT $ liftIO $ getSettings spl
 
+    smash :: Server SMASH
+    smash = getCurrentSmashHealth :<|> getUriSmashHealth
+      where
+        getHealth smashServer = liftIO $ do
+            manager <- newManager defaultManagerSettings
+            health <- healthCheck Nothing (unSmashServer smashServer) manager
+            pure $ maybe Unreachable
+                (\h -> if isHealthy h then Available else Unavailable) health
```
KtorZ: Why not have `healthCheck` returns the sum-type instead of Boolean which require conversion at this stage :thinking:  ?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -740,40 +742,55 @@ monitorMetadata gcStatus tr sp db@(DBLayer{..}) = do
...
+                    | rsIterNumber < maxRetries = pure (not b)
+                    | otherwise = pure False
+
+                ms = (* 1_000_000)
+                baseSleepTime = ms 15
+
+            retrying (constantDelay baseSleepTime) retryCheck checkHealth
+
+        -- no smash configured, equivalent to healthy
+        _ -> pure True
```
KtorZ: That's quite error prone and one more reason to have this function deals directly with the `HealthCheck` sum-type at this level. This really is `NoSmashConfigured`.

specifications/api/swagger.yaml
```diff
@@ -1746,6 +1746,22 @@ components:
...
+
+    ApiHealthCheck: &ApiHealthCheck
+      type: string
+      enum: ["available", "unavailable", "unreachable", "no_smash_configured"]
+      description: |
+        The status of the SMASH server. Possible values are:
+          - available -> SMASH server is awaiting your requests
+          - unavailable -> SMASH server is running, but currently unavailable, try again in a short time
+          - unreachable -> SMASH server could not be reached or didn't return a health status
+          - no_smash_configured -> SMASH is currently not configured, adjust the Settings first
```
KtorZ: I'd suggest using a markdown table here, it reads better in the end documentation. See for instance the field `withdrawal` in: https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/postTransactionFee

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
hasufell: Because `Available` and `Unavailable` is only a subset of the sum-type. `isHealthy` has a smaller domain and only encodes the interpretation of the `status` values returned from SMASH.

lib/core/src/Cardano/Wallet/Api/Types.hs
hasufell: We have to use `UntaggedValue :: SumEncoding` for that and if someone adds a non-nullary constructor later, the json encoding is broken and whether our tests catch it depends on the Arbitrary instances.

```
  | UntaggedValue
    -- ^ Constructor names won't be encoded. Instead only the contents of the
    -- constructor will be encoded as if the type had a single constructor. JSON
    -- encodings have to be disjoint for decoding to work properly.
    --
    -- When decoding, constructors are tried in the order of definition. If some
    -- encodings overlap, the first one defined will succeed.
    --
    -- /Note:/ Nullary constructors are encoded as strings (using
    -- 'constructorTagModifier'). Having a nullary constructor alongside a
    -- single field constructor that encodes to a string leads to ambiguity.
    --
    -- /Note:/ Only the last error is kept when decoding, so in the case of
    -- malformed JSON, only an error for the last constructor will be reported.
```
I think this encoding is generally pretty fishy.

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
hasufell: I'll rework the types a bit to make it less ambiguous.

hasufell:
smash_health

lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -59,15 +63,17 @@ import Control.Tracer
...
     , eitherDecodeStrict
     , fieldLabelModifier
     , genericParseJSON
     , genericToJSON
+    , parseJSON
+    , toJSON
     )
 import Data.Bifunctor
-    ( first )
+    ( Bifunctor (first) )
```
rvl: Is this change necessary?

lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -116,6 +122,10 @@ metadaFetchEp pid (StakePoolMetadataHash bytes)
     hashStr = T.unpack $ T.decodeUtf8 $ convertToBase Base16 bytes
     pidStr  = T.unpack $ toText pid
 
+-- TODO: use SMASH servant types
```
rvl, line 129: Actually, we don't necessarily want to couple so tightly with the IOHK Haskell implementation of SMASH. It may happen that other SMASH servers are used instead.

There should be an API spec for it somewhere, and we should just work in reference to that.

Likewise, in `README.md`, I see a table containing SMASH software versions. That is not especially relevant to users, because they are not the ones operating the SMASH server(s).

lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -170,20 +180,24 @@ registryUrlBuilder baseUrl pid _ hash =
...
-    :: Tracer IO StakePoolMetadataFetchLog
+trMaybe
+    :: Maybe (Tracer IO StakePoolMetadataFetchLog)
+    -> StakePoolMetadataFetchLog
+    -> IO ()
+trMaybe mtr m = forM_ mtr $ \tr -> traceWith tr m
+
+-- | A smash GET request that reads the result at once into memory.
+smashRequest
+    :: Maybe (Tracer IO StakePoolMetadataFetchLog)
```
rvl: You will never need optional tracers like this.

There is `nullTracer :: Tracer m a` for that, or simply `mempty`.

lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -199,6 +213,53 @@ fetchDelistedPools tr uri manager = runExceptTLog $ do
...
+    -> URI
+    -> Manager
+    -> IO (Maybe HealthStatusSMASH)
+healthCheck mtr uri manager = runExceptTLog $ do
+    pl <- smashRequest mtr
+        (uri { uriPath = "/" <> healthCheckEP , uriQuery = "", uriFragment = "" })
+        manager
+    except . eitherDecodeStrict @HealthStatusSMASH $ pl
+  where
+    runExceptTLog
```
rvl, line 233: Here's a tip.
You do not need trace message constructors for both Success and Failure.
Merge the log message constructors to `MsgFetchHealthCheckResult (Either FailureType HealthCheckSuccessType)`.
All your logging code will become simpler like that.

lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -199,6 +213,53 @@ fetchDelistedPools tr uri manager = runExceptTLog $ do
...
+        :: ExceptT String IO HealthStatusSMASH
+        -> IO (Maybe HealthStatusSMASH)
+    runExceptTLog action = runExceptT action >>= \case
+        Left msg ->
+            Nothing <$ trMaybe mtr (MsgFetchHealthCheckFailure msg)
+
+        Right health
+            | isHealthyStatus health -> do
+                trMaybe mtr (MsgFetchHealthCheckFailure
+                    ("Server reports unhealthy status: " <> T.unpack (status health)))
```
rvl: It's better to do log message formatting in the ToText StakePoolMetadataFetchLog instance.
Then the program logic is not obscured by logging.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -740,40 +743,56 @@ monitorMetadata gcStatus tr sp db@(DBLayer{..}) = do
...
-                    fetchDelistedPools trFetch (toDelistedPoolsURI uri) manager
-            tid <- forkFinally
-                (gcDelistedPools gcStatus tr db getDelistedPools)
-                onExit
-            flip finally (killThread tid) $
-                loop (fetchThem $ fetcher [registryUrlBuilder uri])
+    health <- case poolMetadataSource settings of
+        FetchSMASH uri -> do
+            let checkHealth _ = do
+                    r <- healthCheck (Just trFetch) (unSmashServer uri) manager
```
rvl: Is it really necessary to run a health check before fetching data?
If the server is healthy it will return data. If it's not healthy it will return some kind of error from the fetch.

rvl (dismissed):

lib/core/src/Cardano/Pool/Metadata.hs
hasufell: I disagree. The entire point of servant is to be able to consume its types.

> It may happen that other SMASH servers are used instead.

That sounds too theoretical to base implementation decisions on it.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
hasufell: > Is it really necessary to run a health check before fetching data?

Yes, otherwise you're fetching data from an unknown server that may respond with whatever data. SMASH is configurable and we don't control where it points to.

That's the whole point of this ticket: https://jira.iohk.io/browse/ADP-496

lib/core/src/Cardano/Pool/Metadata.hs
hasufell: > Likewise, in README.md, I see a table containing SMASH software versions. That is not especially relevant to users, because they are not the ones operating the SMASH server(s).

Feel free to discuss that with @KtorZ who requested this change.

lib/core/src/Cardano/Pool/Metadata.hs
hasufell: Failure here is HTTP failure, success is just "we got an answer" and doesn't mean the server is healthy. I'd like to keep those separate, because they currently don't have the same severity either.

hasufell: But you are right in the sense that the response object shouldn't have information that diverges from the HTTP status code it returns.

lib/core/src/Cardano/Pool/Metadata.hs
KtorZ: > That sounds too theoretical to base implementation decisions on it.

It is already happening as we speak. Adapools has its own SMASH server. This is why I've also repeated in the past that we want to avoid re-using SMASH's internal types. We do integrate against the interface (a.k.a OpenAPI), not the implementation. 
Now, we'll probably end up using servant types _anyway_ on our end, so it could be interesting to define them in a library that is self contained, but this should not create any strong coupling with SMASH and SMASH development. 

> Likewise, in README.md, I see a table containing SMASH software versions. That is not especially relevant to users, because they are not the ones operating the SMASH server(s).

@rvl I've asked Julian to add that because it does actually concern users. Users are going to point cardano-wallet to SMASH servers and they want to know with which version of SMASH it works. And by version here, I mean the version of the corresponding OpenAPI specification, not the version of SMASH-the-haskell-executable (although arguably, major and minor should move in locksteps for these).

hasufell: > Now, we'll probably end up using servant types anyway on our end, so it could be interesting to define them in a library that is self contained, but this should not create any strong coupling with SMASH and SMASH development.

SMASH already exposes the servant types separately: https://github.com/input-output-hk/smash/tree/master/smash-servant-types

The servant types must be checked against the OpenAPI spec (or generated from it). Then anyone else can still use whatever language they want and we can still use the servant types, because these are isomorphic. The spec part is still TODO on SMASHs end afais, but once we have it, there's no reason not to use the types.

lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -199,6 +207,47 @@ fetchDelistedPools tr uri manager = runExceptTLog $ do
...
+    | T.toLower status /= "ok" = True
+    | otherwise = False
+
+fetchDelistedPools
+    :: Tracer IO StakePoolMetadataFetchLog
+    -> URI
+    -> Manager
+    -> IO (Maybe [PoolId])
+fetchDelistedPools tr uri manager = runExceptTLog $ do
+    pl <- smashRequest tr uri manager
```
hasufell: TODO: 

- [x] this is broken and needs to expand to the delisted endpoint

lib/core/src/Cardano/Pool/Metadata.hs
rvl: Whatever "result" type you choose, it can be pattern matched in the `getSeverityAnnotation` definition.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
rvl: Actually the ticket description says "Daedalus has no way to know whether such a [SMASH] url points to a valid server."
So this change should just provide a way for Daedalus to check that a SMASH URL is correct before applying the setting.

There is no real need to call the health check API before fetching data. Why does everything become more complicated when a health check function is added?

hasufell: > So this change should just provide a way for Daedalus to check that a SMASH URL is correct before applying the setting.

The ticket says that we want to know whether it points to a valid server, not whether the URL is valid. That's not the same thing. We are dealing with user input here. We don't want to start fetching stuff randomly and then show errors. We want to know ahead of time whether a URL points to an actual SMASH server and that server is ready to accept requests. And if it isn't, daedalus can reject the setting. If we just apply it and then error out when actual metadata is fetched, then that's poor user experience.

specifications/api/swagger.yaml
```diff
@@ -1700,6 +1728,25 @@ components:
...
+      enum: ["available", "unavailable", "unreachable", "no_smash_configured"]
+      description: |
+        The status of the SMASH server. Possible values are:
+
+        status                  | description
+        ---                     | ---
+        `"available"`           | server is awaiting your requests
+        `"unavailable"`         | server is running, but currently unavailable, try again in a short time
+        `"unreachable"`         | server could not be reached or didn't return a health status
+        `"no_smash_configured"` | SMASH is currently not configured, adjust the Settings first
```
KtorZ, line 1725: The health check status as a raw string feels a bit rough and prevent any further extension. What about wrapping this one in a `{ "health": ... }` or `{ "smash": ... }` or `{ "status": ... }` singleton..

specifications/api/swagger.yaml
```diff
@@ -389,6 +390,33 @@ x-ScriptValue: &ScriptValue
 
 x-script: &script
   <<: *ScriptValue
+  description: |
```
KtorZ: What exactly is this one doing here ? Sounds like a bad merge resolution.

specifications/api/swagger.yaml
```diff
@@ -341,6 +341,7 @@ x-ScriptValue: &ScriptValue
       type: string
       format: bech32
       pattern: "^(script_vkh)1[0-9a-z]*$"
+      example: script_vkh18srsxr3khll7vl3w9mqfu55n6wzxxlxj7qzr2mhnyreluzt36ms
```
KtorZ: ?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -740,40 +743,56 @@ monitorMetadata gcStatus tr sp db@(DBLayer{..}) = do
...
-                onExit
-            flip finally (killThread tid) $
-                loop (fetchThem $ fetcher [registryUrlBuilder uri])
+    health <- case poolMetadataSource settings of
+        FetchSMASH uri -> do
+            let checkHealth _ = do
+                    r <- healthCheck trFetch (unSmashServer uri) manager
+                    case r of
+                        (Just health)
+                            | isHealthyStatus health -> pure Available
```
KtorZ: We are doing this transformation in two places, which comfort me with the idea that `healthCheck` should really return a `HealthCheckSMASH` datatype in the first place instead of a `Maybe Bool`.

specifications/api/swagger.yaml
```diff
@@ -4068,3 +4133,25 @@ paths:
...
+      tags: ["Utils"]
+      summary: SMASH health
+      description: |
+        Get health status of given SMASH uri.
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema: *ApiSmashServer
+      responses: *responsesPostSmashHealth
```
KtorZ: A `POST` request is awkward here, there's no resources being created. May we stick to a more restful approach and offer instead one single endpoint `GET /v2/smash/health` which also accepts an optional query parameter `url` or type `ApiSmashServer`; when provided, we'll fetch health from this URL, when omitted, we'll fetch health from the currently configured smash server if any.

hasufell: I was just trying to avoid url encoding as query parameter, since it's pesky.

KtorZ: I'd favor pesky url-encoded query parameter over awkward non restful post request :blush:

specifications/api/swagger.yaml
hasufell: I'm also confused :sweat_smile: 
It's a nice example no?

KtorZ: It is, but what the heck has this addition to do with the PR :sweat_smile:  ... again, some merge resolution which didn't went well I assume.

lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
```diff
@@ -1151,6 +1151,13 @@ instance Malformed (BodyParam ApiMaintenanceActionPostData) where
           )
         ]
 
+instance Malformed (BodyParam (ApiT SmashServer)) where
+    malformed = first (BodyParam . Aeson.encode) <$>
```
KtorZ: Is it still relevant?

KtorZ (approved):

lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
hasufell: I guess we don't have query param specs yet:

```
    -- TODO
    -- Capture request query params as QueryParam
    --
    -- newtype QueryParam t = QueryParam Text
    --     deriving (Typeable)
    --
    -- type MkQueryRequest api :: *
    -- gEveryQueryParams :: Proxy api -> Request -> MkQueryRequest api
```

KtorZ: Indeed we do not, but the malformed instances are used only if the underlying type is used in the API. And now, there is no API call using `ApiT SmashServer` as a body parameter.

hasufell:
@KtorZ I rebased, maybe have a quick look again. Not the first time I messed up a rebase...

KtorZ (approved):
bors merge

PR #2289, opened 2020-11-03
Use openapi3 for schema tests

hasufell:
# Issue Number

ADP-417

# Overview

- [x] added openapi3 as dep, fixed stack.cabal
- [x] fixed types and import, worked around lack of `servant-openapi` package
- [x] fix test failures
- [x] fix jormungandr

# Comments


stack.yaml
```diff
@@ -34,6 +34,14 @@ extra-deps:
 - process-1.6.5.0
 - time-1.8.0.2
 
+- openapi3-3.0.0.1
+- optics-0.3
+- optics-core-0.3.0.1
+- optics-extra-0.3
+- optics-th-0.3.0.2
+- optics-vl-0.2.1
+- insert-ordered-containers-0.2.3.1
```
hasufell, line 44: There's a difficult to track down build failure with `0.2.3` (which is the lowest bound for openapi3). It seems it only fails on a certain combination of packages, so I can't provide an upstream PR without further investigation.

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -328,6 +323,41 @@ import qualified Data.Yaml as Yaml
...
 import qualified Test.Utils.Roundtrip as Utils
 
+import Servant.Swagger.Internal.TypeLevel.TMap
+    ( TMap )
+import Servant.Swagger.Internal.TypeLevel.Every
+    ( Every, EveryTF, tmapEvery )
+import Servant.Swagger.Internal.TypeLevel.API
+    ( BodyTypes )
+
+validateEveryToJSON
```
hasufell: This is stolen from `Servant.Swagger.Test`.

specifications/api/swagger.yaml
```diff
@@ -311,15 +311,9 @@ x-poolMetadataSource: &poolMetadataSource
...
-    - type: string
-      enum:
-        - none
-        - direct
-    - title: URI
-      type: string
-      format: uri
-      example: https://smash.cardano-mainnet.iohk.io/
+  type: string
+  pattern: '(null|direct|https?:\/\/[a-zA-Z0-9-_~.]+/?)'
```
hasufell: The oneOf had an overlap, because it doesn't check the format. So I think it's better to provide a lax javascript regex.

hasufell:
This broke all of nix

KtorZ:
@hasufell I've referenced the dependency (openapi3) from its github repository instead of its hackage package because the hackage index depends on which version of haskell.nix is used by the CI behind the scene. And the version we're currently using is too old so it doesn't know about openapi3.  

Bumping Haskell.nix is a possibility, but it sometimes can be quite some work which would be unnecessary _now_.

rvl:
Thanks for fixing the nix @KtorZ .

@hasufell, for the record, here is [the documentation about updating the pins](https://github.com/input-output-hk/cardano-wallet/wiki/Building#haskellnix-pin), and I just added a link to the Haskell.nix ChangeLog file.

hasufell:
> @hasufell, for the record, here is the documentation about updating the pins, and I just added a link to the Haskell.nix ChangeLog file.

I read that already and it didn't work:

```
$ niv update haskell.nix
Update haskell.nix
  INFO: new sources.nix available: 18 -> 23
    Please run 'niv init' or add the following line in the nix/sources.nix file:
    # niv: no_update
Done: Update haskell.nix
$ nix-shell
unpacking 'https://github.com/input-output-hk/haskell.nix/archive/23539aef9ea3591dc0d218c5f60b2a4f7d229a20.tar.gz'...
unpacking 'https://github.com/input-output-hk/hackage.nix/archive/7f67b08b9c6f5e9bd5128bf06de54f5de012ee60.tar.gz'...
warning: dumping very large path (> 256 MiB); this may run out of memory
trace: Using IOHK default nixpkgs
trace: gitSource.nix: /nix/store/dvp437wx7bmj4kfpvybifnj0aak27hmc-cardano-node-src does not seem to be a git repository,
assuming it is a clean checkout.
trace: Using index-state: 2020-07-15T00:00:00Z for cardano-node-src
trace: Using index-state: 2020-07-15T00:00:00Z for cardano-node-src
trace: Cleaning component source not supported for hpack package: cardano-addresses-2.1.0
error: The Haskell package set does not contain the package: indexed-profunctors (build dependency).
If you are using Stackage, make sure that you are using a snapshot that contains the package. Otherwise you may need to update the Hackage snapshot you are using, usually by updating haskell.nix.
```

rvl:
```
error: The Haskell package set does not contain the package: indexed-profunctors (build dependency).
If you are using Stackage, make sure that you are using a snapshot that contains the package.
```

What this means is that the Stackage snapshot doesn't contain indexed-profunctors. So it needed the change that @KtorZ added.

The Haskell.nix rev is now updated on master anyway. I have pushed a fix.

hasufell:
> What this means is that the Stackage snapshot doesn't contain indexed-profunctors. So it needed the change that @KtorZ added.

Can we add documentation about this? What are the options when this case happens and how to execute the different solutions.

rvl:
Sure. But note that the nix build is based on `stack.yaml`. So if the stack build fails due to dependencies not in stackage, then the nix build is also going to fail.

hasufell:
@rvl please don't do reverse merges back from master. It messes up the history and makes rebasing etc harder.

rvl:
What is the actual problem @hasufell? It should be easy for you to rebase, and then the merge commit disappears.
I wasn't going to rebase and force push your branch out of consideration for you, and also because it sometimes happens that the other person just force pushes the branch back to what it was before, and the changes are lost.

hasufell:
>  It should be easy for you to rebase, and then the merge commit disappears.

That's only true if you rebase against upstream master and that now causes lots of other conflicts. Squashing isn't easily possible anymore either (which reduces merge conflicts during rebases against upstream branches).

rvl:
It's easy. There are two merge conflicts which you would have needed to resolve anyway before merging into master.

hasufell:
> It's easy. There are two merge conflicts which you would have needed to resolve anyway before merging into master.

I'd prefer not to have reverse merges in my branches. Thanks :)

lib/test-utils/src/Test/Utils/OpenApi.hs
```diff
@@ -0,0 +1,185 @@
...
+        specs = tmapEvery (Proxy :: Proxy (Typeable ': Show ': Arbitrary ': cs)) aprop px
+
+        aprop :: forall p' a. (EveryTF cs a, Typeable a, Show a, Arbitrary a) => p' a -> Spec
+        aprop _ = prop (show (typeRep (Proxy :: Proxy a))) (f :: a -> Property)
+
+-- | Verify that every type used with @'JSON'@ content type in a servant API
+-- has compatible @'ToJSON'@ and @'ToSchema'@ instances using @'validateToJSONWithPatternChecker'@.
+--
+-- For validation without patterns see @'validateEveryToJSON'@.
+validateEveryToJSONWithPatternChecker :: forall proxy api. TMap (Every [Typeable, Show, Arbitrary, ToJSON, ToSchema]) (BodyTypes JSON api) =>
```
hasufell: This is a copy-paste. We might want to fork servant-swagger, but that will be more work.

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -1045,7 +1029,7 @@ instance Arbitrary Script where
             Positive m <- arbitrary
             let n' = n `div` (m + 1)
             scripts <- vectorOf m (scriptTree n')
-            atLeast <- choose (0, fromIntegral (m + 1))
+            atLeast <- choose (1, fromIntegral (m + 1))
```
hasufell, line 1032: https://github.com/input-output-hk/cardano-addresses/blob/350abfe36c3fc2949da567ca47a957a9d5789c92/core/lib/Cardano/Address/Script/Parser.hs#L61-L88

lib/core/cardano-wallet-core.cabal
```diff
@@ -232,6 +232,7 @@ test-suite unit
     , network
     , network-uri
     , persistent
+    , regex-pcre-builtin
```
hasufell, line 235: I don't know if it builds on windows, but it has the code bundled.

cabal.project
```diff
@@ -35,13 +35,13 @@ source-repository-package
...
     location: https://github.com/input-output-hk/cardano-addresses
-    tag: 9a64226fd749513ad0ed2e43e663f7132c92faa7
+    tag: 1399751501b03346265b2e429e5832bacd6df279
     subdir: command-line
 
 source-repository-package
     type: git
     location: https://github.com/input-output-hk/cardano-addresses
-    tag: 9a64226fd749513ad0ed2e43e663f7132c92faa7
+    tag: 1399751501b03346265b2e429e5832bacd6df279
```
KtorZ: release is coming soon.

lib/core/cardano-wallet-core.cabal
KtorZ: `bors try` will tell us this. There's a hydra job that builds the wallet for Windows, from Linux though.

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -2000,49 +1938,19 @@ instance ToSchema ApiPostRandomAddressData where
...
-    NamedSchema (Just "CredentialValue") $ mempty
-        & additionalProperties ?~ AdditionalPropertiesAllowed True
-
 -- | Utility function to provide an ad-hoc 'ToSchema' instance for a definition:
 -- we simply look it up within the Swagger specification.
 declareSchemaForDefinition :: Text -> Declare (Definitions Schema) NamedSchema
 declareSchemaForDefinition ref = do
     let json = foldl' unsafeLookupKey specification ["components","schemas",ref]
-    case Aeson.eitherDecode' $ replaceRefs  $ Aeson.encode json of
+    case Aeson.eitherDecode' $ Aeson.encode json of
```
KtorZ, line 1949: :+1:

specifications/api/swagger.yaml
```diff
@@ -308,18 +308,12 @@ x-poolMetadataSource: &poolMetadataSource
     Pool metadata source. This sets the metadata fetching strategy.
 
     Possible values are
-      * null -> no fetching
+      * none -> no fetching
```
KtorZ, line 311: :ok_hand:

KtorZ:
LGTM, modulo the revision on cardano-addresses which we need to address (release for this one is coming today hopefully).

hasufell:
bors try

KtorZ (approved):

PR #2291, opened 2020-11-03
bump version to 2020.11.3

paweljakubas:
# Issue Number

# Overview

- [ ] I have ...

# Comments

### DIFF
--- a/nix/migration-tests.nix
+++ b/nix/migration-tests.nix
@@ -56,7 +56,7 @@ let
       sha256 = "1nfzfdv03nz2bhin2jflxsmh597vkjifb76jsdg0384x89gi0a10"; }
     { rev = "v2020-09-30";
       sha256 = "19vmyq0m67ih295z1y8ng2rdn85mi0rvw26k711lbxsqri7fmyvz"; }
-    { rev = "v2020.11.3";
+    { rev = "v2020-11-03";
       sha256 = "0kc1ddidp1vip8rfnm7vp5avx46xr9nvyzn7q9r3yn50ylhx0i0j"; }
   ];
 
--- a/nix/migration-tests.nix
+++ b/nix/migration-tests.nix
@@ -35,8 +35,8 @@ with pkgs.lib;
 let
 
   # List of git revisions to test against.
-  # One can get sha256 for release via nix-prefetch-url, e.g. for v2020-10-13:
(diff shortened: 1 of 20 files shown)


scripts/make_release.sh
```diff
@@ -17,11 +17,12 @@ set -euo pipefail
 
 ################################################################################
 # Release-specific parameters (Change when you bump the version)
-OLD_GIT_TAG="v2020-09-30"
-OLD_CABAL_VERSION="2020.9.30"
+# Do not use trailing zeros, ie., not 2020.11.03 but 2020.11.3
```
Anviking, line 20: 👏

Anviking (approved):

nix/migration-tests.nix
```diff
@@ -56,7 +56,7 @@ let
       sha256 = "1nfzfdv03nz2bhin2jflxsmh597vkjifb76jsdg0384x89gi0a10"; }
     { rev = "v2020-09-30";
       sha256 = "19vmyq0m67ih295z1y8ng2rdn85mi0rvw26k711lbxsqri7fmyvz"; }
-    { rev = "v2020-10-13";
+    { rev = "v2020.11.3";
```
piotr-iohk: I think this should not change. Migration tests to be updated after release, and the format should be `v2020-11-03` here I suppose.

hasufell:
Should we wait for a patch release of cardano-addresses? Because it can't be built with cabal currently due to https://github.com/input-output-hk/cardano-addresses/commit/8ececd0d7c0ed6e0c654ff94dac8ad91a5ad596a#diff-3ff3d7b39854fd35883974ce0822dd5a599e98e093b8079419e9ecf21a7c6000
This problem propagates to the wallet.

nix/migration-tests.nix
rvl: Yes that's right

rvl (approved):
Don't forget the version number table in README.md.

nix/migration-tests.nix
paweljakubas: -> https://github.com/input-output-hk/cardano-wallet/pull/2291/commits/1408ac93cb7f3866144f67a9e01e696d4837e03e

paweljakubas:
@hasufell I think there is some more time for cardano-addresses release. There is one PR to be merged (should be merged around tomorrow), but also prefixes PR (HRPs compliant with CIP5) waits for be done <- @KtorZ

PR #2293, opened 2020-11-04
Revise API error codes

piotr-iohk:
# Issue Number

ADP-291

# Overview

- f533091070c2ccc7945ffff2143f3b6c4f06599c
  Revise error codes:  - 404 when wallet not found  - 403 when transaction is too big
  
- 430a31f88b5fa4d3632f4e5a53e4afa9a69bae32
  Update swagger
  
- e90b26430eaf90532966a92504af89f338fa89fc
  Update integration tests

# Comments

Whie reviewing https://github.com/input-output-hk/cardano-wallet/pull/2258 I noticed that some error codes are a bit inconsistent, e.g.:
 - 410 or 404 is returned when the wallet is not found.
 - 400 or 403 - when transaction is too big.
 
 This pr is about to revise it and:
 - return 404 on wallet not found (removing 410 altogether)
 - return 403 on transaction too big

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -432,7 +432,6 @@ import Servant
     , err403
     , err404
     , err409
-    , err410
     , err500
     , err503
     , serve
@@ -2431,7 +2430,7 @@ instance Buildable e => LiftHandler (ErrSelectForPayment e) where
                 , "transaction, then cancel the previous one first."
                 ]
         ErrSelectForPaymentTxTooLarge maxSize maxN  ->
-            apiError err400 TransactionIsTooBig $ mconcat
+            apiError err403 TransactionIsTooBig $ mconcat
                 [ "I am afraid that the transaction you're trying to submit is "
                 , "too large! The network allows transactions only as large as "
                 , pretty maxSize, "s! As it stands, the current transaction only "
@@ -2460,8 +2459,8 @@ instance LiftHandler ErrSignPayment where
     handler = \case
         ErrSignPaymentMkTx e -> handler e
         ErrSignPaymentNoSuchWallet e -> (handler e)
-            { errHTTPCode = 410
-            , errReasonPhrase = errReasonPhrase err410
+            { errHTTPCode = 404
+            , errReasonPhrase = errReasonPhrase err404
             }
         ErrSignPaymentWithRootKey e@ErrWithRootKeyNoRootKey{} -> (handler e)
             { errHTTPCode = 403
@@ -2550,8 +2549,8 @@ instance LiftHandler ErrSubmitTx where
     handler = \case
         ErrSubmitTxNetwork e -> handler e
         ErrSubmitTxNoSuchWallet e@ErrNoSuchWallet{} -> (handler e)
-            { errHTTPCode = 410
-            , errReasonPhrase = errReasonPhrase err410
+            { errHTTPCode = 404
+            , errReasonPhrase = errReasonPhrase err404
             }
 
 instance LiftHandler ErrNetworkUnavailable where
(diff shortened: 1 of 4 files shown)


lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2411,7 +2410,7 @@ instance Buildable e => LiftHandler (ErrSelectForPayment e) where
                 , "transaction, then cancel the previous one first."
                 ]
         ErrSelectForPaymentTxTooLarge maxSize maxN  ->
-            apiError err400 TransactionIsTooBig $ mconcat
+            apiError err403 TransactionIsTooBig $ mconcat
```
hasufell, line 2433: Not sure, I think this might actually be https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2440,8 +2439,8 @@ instance LiftHandler ErrSignPayment where
     handler = \case
         ErrSignPaymentMkTx e -> handler e
         ErrSignPaymentNoSuchWallet e -> (handler e)
-            { errHTTPCode = 410
```
hasufell, line 2463: Ok, so `Gone` would be a viable option if we know the wallet existed beforehand. But I guess we don't?

hasufell (commented):
Did you consider dropping 415 altogether?

lib/core/src/Cardano/Wallet/Api/Server.hs
hasufell: So, all uses of this type are as follows

```hs
    withRootKey @_ @s ctx wid pwd ErrSignPaymentWithRootKey $ \xprv scheme -> do
        let pwdP = preparePassphrase scheme pwd
        mapExceptT atomically $ do
            cp <- withExceptT ErrSignPaymentNoSuchWallet $ withNoSuchWallet wid $
                readCheckpoint (PrimaryKey wid)
            (cs', s') <- assignChangeAddressesForSelection
                argGenChange cs (getState cp)
            withExceptT ErrSignPaymentNoSuchWallet $
                putCheckpoint (PrimaryKey wid) (updateState s' cp)

            let keyFrom = isOwned (getState cp) (xprv, pwdP)
            let rewardAcnt = mkRewardAccount (xprv, pwdP)

            (tx, sealedTx) <- withExceptT ErrSignPaymentMkTx $ ExceptT $
                pure $ mkStdTx tl rewardAcnt keyFrom txExp md cs'

            (time, meta) <- liftIO $ mkTxMeta ti (currentTip cp) s' tx cs' txExp
            return (tx, meta, time, sealedTx)
```

That means we can only hit this code-path if we found a root key for the given wallet id. Does that imply the wallet existed?

KtorZ: Gone _made sense_ in the context of concurrent request. When one creates a transaction, there are multiple back-and-forth with the database in separate db transactions. In principle, the wallet could disappear between any of these db transaction might there be a concurrent request deleting the wallet at the same time.

Arguably, this is veeeeeeeery edgy, and from a user perspective doesn't matter much. So changing to 404 sounds like a good move.

lib/core/src/Cardano/Wallet/Api/Server.hs
KtorZ: Sounds relevant for this use case indeed.

KtorZ (approved):

lib/core/src/Cardano/Wallet/Api/Server.hs
piotr-iohk: Judging by the error message I'd say yes and no.

Yes, in case where the metadata is too big. As error message suggests:
```
Alternatively, try sending to less recipients or with smaller metadata.
```

Not necessarily, if the transaction turns out to be too big:
```
Note that I am selecting inputs randomly, so retrying *may work* 
```

For the sake of _keeping it simple_ I'd stay with `403`, which is more general.

piotr-iohk:
> Did you consider dropping 415 altogether?

I haven't actually. As far as I see there are some tests against `415` in the test suite, so maybe let's have it here.

PR #2300, opened 2020-11-05
Make nix-shell evaluation faster

rvl:
### Issue Number

ADP-382 / #1888 

### Overview

- Updates Haskell.nix rev to latest.
- Use the "materialization" feature so that evaluation of the nix-shell (done locally and on hydra) is cached as much as possible.

### Comments

```
$ GC_PRINT_STATS=1 command time nix-shell --option substitute false --run "cabal --version"

master branch:
  ...
  Grow heap to 2786488 KiB after 806218912 bytes allocated
  cabal-install version 3.2.0.0
  compiled using version 3.2.0.0 of the Cabal library 
  30.51user 2.34system 0:29.05elapsed 113%CPU (0avgtext+0avgdata 3300584maxresident)k
  0inputs+0outputs (0major+915357minor)pagefaults 0swaps

PR branch:
  ...
  Grow heap to 2393272 KiB after 442027296 bytes allocated
  cabal-install version 3.2.0.0
  compiled using version 3.2.0.0 of the Cabal library 
  26.63user 1.84system 0:24.78elapsed 114%CPU (0avgtext+0avgdata 2889004maxresident)k
  0inputs+0outputs (0major+812718minor)pagefaults 0swaps

memory usage: 2.3GiB  ( -14%)
startup time: 0:24.78 ( -15%)
```

PR #2301, opened 2020-11-06
Reduce nix-shell eval memory usage

rvl:
### Issue Number

ADP-382 / #1888

### Overview

1. Similar memory and speed gains as in #2300. cardano-node was previously imported as a niv source. It is also part of the haskell.nix package set, because it's a stack.yaml dependency. So there were haskell package sets duplicated basically.

2. Add a `platform` argument to the `release.nix` jobset. This will split the jobset into three. With support from Hydra (input-output-hk/ci-ops#26), this will make evaluation use about 1/3 of the memory, and allow jobs to start building quicker.

### Comments

- Removing cardano-wallet-jormungandr and nix build of jormungandr is done in #2313. This also reduces the weight to the nix-shell.
- It is best to merge this after #2245, because the iohk-monitoring build will be simpler.


nix/cardano-node-overlay.nix
```diff
@@ -0,0 +1,41 @@
...
+    } (mapAttrsToString (name: env: ''
+      cfg=$out/${name}
+      mkdir -p $cfg
+      jq . < ${mkTopology env} > $cfg/topology.json
+      jq . < ${__toFile "${name}-config.json" (__toJSON (updateConfig env))} > $cfg/configuration.json
+    '' + (if env.consensusProtocol == "Cardano" then ''
+      jq . < ${env.nodeConfig.ByronGenesisFile} > $cfg/genesis-byron.json
+      cp ${env.nodeConfig.ShelleyGenesisFile} $cfg/genesis-shelley.json
+    '' else ''
+      jq . < ${env.genesisFile} > $cfg/genesis.json
```
jbgi, line 39: this `else` clause will include `TPraos` (Shelley) protocol, thus not supporting formatting of genesis through `jq`.

jbgi: though I see now that this was already there before.

jbgi (approved):

PR #2303, opened 2020-11-09
Move `Hash` type and related functions into separate module.

jonathanknowles:
# Issue Number

Preparation for [ADP-347](https://jira.iohk.io/browse/ADP-347).

# Overview

This PR moves the `Hash` type and related functions to a separate module `Cardano.Wallet.Primitive.Types.Hash`. This allows code to use the `Hash` type without having to incur a dependency on the `Primitive.Types` module, which is already very large and slow to recompile.

# Comments

This PR doesn't re-export the `Hash` type from `Primitive.Types`.

Although it would be convenient to be able to import everything from `Primitive.Types` in one go (with a qualified import), it would also create a compilation bottleneck, since anything that imported `Hash` from the `Primitive.Types` module would have to depend on everything required to build `Primitive.Types`.

### DIFF
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -170,12 +170,13 @@ import Cardano.Wallet.Primitive.SyncProgress
 import Cardano.Wallet.Primitive.Types
     ( AddressState
     , Coin (..)
-    , Hash
     , PoolMetadataSource (..)
     , SortOrder
     , WalletId
     , WalletName
     )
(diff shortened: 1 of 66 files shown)


rvl (approved):
Seems reasonable to re-use the Hash type for monetary policy ID.

This will fix the circular dependency.

Moving more types to separate modules and not re-exporting from a single types module may help reduce compilation in future.

Looks fine, as long as it builds and passes stylish-haskell.

PR #2312, opened 2020-11-11
Update cabal.project

hasufell:


cabal.project
```diff
@@ -35,13 +35,13 @@ source-repository-package
 source-repository-package
     type: git
     location: https://github.com/input-output-hk/cardano-addresses
-    tag: 9a64226fd749513ad0ed2e43e663f7132c92faa7
+    tag: 1399751501b03346265b2e429e5832bacd6df279
```
hasufell, line 38: https://github.com/input-output-hk/cardano-wallet/blob/ddc445736e56bd94529ca05222ad93312604b270/stack.yaml#L49

rvl (commented):
Actually, I am not sure whether we should not use this rev of cardano-addresses.

The `stack.yaml` revision got bumped in PR #2253. When I did the final review, I thought this 
change had been reverted, but I guess not.
 
@KtorZ stated an objection here: https://github.com/input-output-hk/cardano-wallet/pull/2253#discussion_r510226169

I think we should put this PR on hold, make releases of both cardano-addresses and cardano-transactions (because the latter is also needed for #2245, and the latest version conflicts with cardano-addresses-2.1.0).

PR #2314, opened 2020-11-12
Testing maintenance actions

piotr-iohk:
# Issue Number

ADP-478

# Overview

- 2bb835971f0e8572206ac42a5da3e081d8cc1eb6
  Move updateMetadataSource and verifyMetadataSource to DSL.hs
  
- 26812b559714e69203876ea81d74c719951a9e64
  Make PostPoolMaintenance to be 204 No Content to comply with swagger
  
- afda59429c75da2c50565ddc035d0f406bd5a819
  Integration tests for triggering maintenance action while metadata source is none or direct
  

# Comments

First two test from the "proposal" https://jira.iohk.io/browse/ADP-478?focusedCommentId=27784&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-27784

And small fix for [postMaintenanceAction](https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/postMaintenanceAction) to return `204` and no content.

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
@@ -37,6 +37,8 @@ import Cardano.Wallet.Primitive.Types
     ( Coin (..)
     , Direction (..)
     , PoolId (..)
+    , PoolMetadataGCStatus (..)
+    , PoolMetadataSource (..)
     , StakePoolMetadata (..)
     , StakePoolTicker (..)
     , TxStatus (..)
@@ -89,6 +91,7 @@ import Test.Integration.Framework.DSL
     , delegationFee
     , emptyWallet
     , eventually
+    , eventuallyUsingDelay
     , expectErrorMessage
     , expectField
     , expectListField
@@ -110,10 +113,13 @@ import Test.Integration.Framework.DSL
     , quitStakePool
     , quitStakePoolUnsigned
     , request
+    , triggerMaintenanceAction
     , unsafeRequest
     , unsafeResponse
     , updateMetadataSource
     , verify
+    , verifyMaintenanceAction
+    , verifyMetadataSource
     , waitForNextEpoch
     , walletId
     , (.>)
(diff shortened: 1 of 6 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -144,6 +148,22 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
     let listPools ctx stake = request @[ApiStakePool] ctx
                 (Link.listStakePools stake) Default Empty
 
+    it "STAKE_POOLS_MAINTENANCE_01 - \
+        \trigger GC action when metadata source = direct" $ \ctx -> runResourceT $ do
+        updateMetadataSource ctx "direct"
+        verifyMetadataSource ctx FetchDirect
+        triggerMaintenanceAction ctx "gc_stake_pools"
+        eventually "GC Status shows as NotApplicable" $ do
```
hasufell: We should have an `eventually` that isn't 90+ seconds. Triggering this should be in the realm of <10s usually.

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -481,7 +482,7 @@ type DelegationFee = "wallets"
 type PostPoolMaintenance = "stake-pools"
     :> "maintenance-actions"
     :> ReqBody '[JSON] ApiMaintenanceActionPostData
-    :> Post '[JSON] ApiMaintenanceAction
+    :> PostNoContent
```
hasufell, line 485: Need to update `swagger.yaml`

hasufell (changes requested):
Looks good, but need to update swagger schema

lib/core/src/Cardano/Wallet/Api.hs
piotr-iohk: Swagger already shows:
Screenshot from 2020-11-12 14-41-13

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
piotr-iohk: ok, added eventually with 10s timeout.

hasufell (approved):

PR #2317, opened 2020-11-12
Update cardano-addresses to 3.1.0

hasufell:


KtorZ (approved):

hasufell:
Updated to 3.1.0

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -232,6 +230,7 @@ import Shelley.Spec.Ledger.BaseTypes
 import Type.Reflection
     ( Typeable, typeRep )
 
+import qualified Cardano.Address.Style.Shelley as SH
```
rvl, line 233: I would have preferred not changing it to use qualified imports. This module is called `...Shelley.Compatibility` so a prefix like `SH` (I assume it stands for Shelley?) is redundant.

rvl (approved):
:+1:

rvl:
bors merge

PR #2318, opened 2020-11-12
Additional tweaks with regards to static address construction via the API

KtorZ:
# Issue Number

ADP-461

# Overview

- 9b362af395375d30135b04961f1406e0e821337e
  :round_pushpin: **return non-extended public key (without chain code) as 'ApiVerificationKey'**
    The chain code is useless because there's no further derivation that can happen at this stage. Plus, it does make it harder to re-use the output from the 'getWalletKey' endpoint into others because one needs to decode the bech32 string and strip away the chain code.

- 3b005831530f01772ef4abd71dba6a5a8a59244a
  :round_pushpin: **revise API documentation regarding address construction from script**
    - give more details about how scripts are constructed
  - move description around so they show up correctly in ReDoc.
  - move address endpoints into the 'Address' group and remove the 'Utils' group

# Comments

TODO: use public key in script construction instead of hashes, and do the hashing internally.

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
@@ -51,7 +51,7 @@ import Cardano.Wallet.Primitive.SyncProgress
 import Cardano.Wallet.Primitive.Types
     ( walletNameMaxLength, walletNameMinLength )
 import Cardano.Wallet.Unsafe
-    ( unsafeFromHex )
+    ( unsafeFromHex, unsafeXPub )
 import Control.Monad
     ( forM, forM_ )
 import Control.Monad.IO.Class
@@ -134,6 +134,7 @@ import Test.Integration.Framework.TestData
 -- of cardano-crypto here.
 import qualified Cardano.Crypto.Wallet as CC
 import qualified Cardano.Wallet.Api.Link as Link
+import qualified Data.ByteString as BS
 import qualified Data.ByteString.Lazy as BL
 import qualified Data.Set as Set
 import qualified Data.Text as T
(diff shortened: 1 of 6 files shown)


lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1157,20 +1157,20 @@ instance FromJSON ApiVerificationKey where
...
-        parseXPub =
-            maybe (fail errXPub) pure . xpubFromBytes
-          where
-            errXPub =
-                "Not a valid extended public key."
+        parsePub bytes
+            | BS.length bytes == 32 =
+                pure bytes
+            | otherwise =
+                fail "Not a valid Ed25519 public key. Must be 32 bytes."
```
paweljakubas:
```suggestion
                fail "Not a valid Ed25519 public key. Must be 32 bytes, without chain code"
```

specifications/api/swagger.yaml
```diff
@@ -337,17 +337,18 @@ x-anyAddress: &anyAddress
...
       pattern: "^(script_vkh)1[0-9a-z]*$"
-      example: script_vkh18srsxr3khll7vl3w9mqfu55n6wzxxlxj7qzr2mhnyreluzt36ms
 
     - title: All
       type: object
       required:
         - all
       properties:
         all:
+          description: Script primitive for which  which all list elements' must be valid scripts.
```
paweljakubas, line 351:
```suggestion
          description: Script primitive for which all list elements' corresponding signing keys are expected to make script valid.
```

specifications/api/swagger.yaml
```diff
@@ -359,6 +360,7 @@ x-ScriptValue: &ScriptValue
         - any
       properties:
         any:
+          description: Script primitive for which  which any of the list elements' must be valid scripts. It is equivalent to `some` with `"at_least"=1`.
```
paweljakubas:
```suggestion
          description: Script primitive for which a signing key corresponding to any of the list elements' verification keys is expected to make the script valid. It is equivalent to `some` with `"at_least"=1`.
```

specifications/api/swagger.yaml
paweljakubas:
```suggestion
          description: Script primitive for which all signing keys corresponding to all list elements' verification keys are expected to make the script valid.
```

specifications/api/swagger.yaml
```diff
@@ -370,6 +372,7 @@ x-ScriptValue: &ScriptValue
         - some
       properties:
         some:
+          description: Script primitive for which at least a given number of list elements' must be valid scripts.
```
paweljakubas:
```suggestion
          description: Script primitive for which at least a given number of signing keys corresponding to the list elements' verification keys are expected to make the script valid.
```

paweljakubas (approved):
LGTM! Left some suggestions. Very fond of infographic you included!

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
```diff
@@ -998,23 +999,19 @@ spec = describe "SHELLEY_WALLETS" $ do
             matrix =
```
piotr-iohk, line 999: Idea for integration test:
 - create empty shelley wallet that has {walletId}
 - let paymentKey = `GET /wallets/{walletId}/keys/utxo_external/0`
 - let stakeKey = `GET /wallets/{walletId}/keys/mutable_account/0`
 - Construct address:
```
POST /addresses
{payment: ${paymentKey}, stake: ${stakeKey}}
```
- verify that this address is first address on {walletId}

KtorZ: Sounds like it would work :+1:

KtorZ:
bors retry

KtorZ:
bors retry

KtorZ:
bors retry

KtorZ:
bors retry 

please.

PR #2319, opened 2020-11-13
Add SMASH integration test

hasufell:
# Issue Number

https://jira.iohk.io/browse/ADP-478

* add integration test for SMASH via file stub server

### DIFF
--- a/lib/core-integration/src/Test/Integration/Framework/DSL.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/DSL.hs
@@ -133,6 +133,7 @@ module Test.Integration.Framework.DSL
     , oneSecond
     , getTTLSlots
     , updateMetadataSource
+    , bracketSettings
     , verifyMetadataSource
     , triggerMaintenanceAction
     , verifyMaintenanceAction
@@ -212,6 +213,7 @@ import Cardano.Wallet.Api.Types
     , DecodeStakeAddress (..)
     , EncodeAddress (..)
     , Iso8601Time (..)
+    , SettingsPutData (..)
     , WalletStyle (..)
     , insertedAt
     )
@@ -598,7 +600,7 @@ defaultTxTTL = 7200
 --
 updateMetadataSource :: (MonadIO m, MonadCatch m) => Context t -> Text -> m ()
 updateMetadataSource ctx t = do
-    r <- request @(ApiT Settings) ctx Link.putSettings Default payload
+    r <- request @SettingsPutData ctx Link.putSettings Default payload
     expectResponseCode HTTP.status204 r
  where
   payload = Json [aesonQQ| {
@@ -606,6 +608,18 @@ updateMetadataSource ctx t = do
            "pool_metadata_source": #{t}
             }
        } |]
+
+bracketSettings :: (MonadIO m, MonadCatch m) => Context t -> m () -> m ()
+bracketSettings ctx action = do
+    r@(_, response) <- request @(ApiT Settings) ctx Link.getSettings Default Empty
+    expectResponseCode HTTP.status200 r
+    case response of
+        Left e -> wantedSuccessButError e
+        Right s -> do
+            action
+            r' <- request @SettingsPutData ctx Link.putSettings Default
+                (Json $ Aeson.toJSON $ SettingsPutData s)
+            expectResponseCode HTTP.status204 r'
 
 verifyMetadataSource
     :: (MonadIO m, MonadCatch m)
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
@@ -1120,6 +1120,24 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
                                 "b45768c1a2da4bd13ebcaa1ea51408eda31dcc21765ccbd407cda9f2"]
                     ]
 
+            updateMetadataSource ctx "direct"
+            eventually "pools are not delisted anymore" $ do
+                r <- listPools ctx arbitraryStake
+                verify r
+                    [ expectListSize 3
+                    , expectField Prelude.id $ \pools' -> do
+                        let metadataActual = Set.fromList $
+                                mapMaybe (fmap getApiT . view #metadata) pools'
+                            delistedPools = filter (\pool -> Delisted `elem` flags pool)
+                                pools'
+                        metadataActual
+                            `shouldSatisfy` (`Set.isSubsetOf` metadataPossible)
+                        metadataActual
+                            `shouldSatisfy` (not . Set.null)
+                        (fmap (getApiT . view #id) delistedPools)
+                            `shouldBe` []
+                    ]
+
   where
     metadataPossible = Set.fromList
         [ StakePoolMetadata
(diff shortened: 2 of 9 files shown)


hasufell:
@piotr-iohk can you think of more scenarios?

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -1120,7 +1097,56 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
             let epochs = poolGarbageCollectionEpochNo <$> events
             (reverse epochs `zip` [1 ..]) `shouldSatisfy` all (uncurry (==))
 
+    it "STAKE_POOLS_SMASH_01 - fetching metadata from SMASH works with delisted pools" $
+        \ctx -> runResourceT $ do
+            smashUrl <- liftIO $ getEnv "CARDANO_WALLET_SMASH_URL"
```
KtorZ, line 1103: Running this one and STAKE_POOLS_LIST_01 in parallel will likely fail since they'll be both changing smash settings. Let's hope it doesn't happen _too often_.

lib/shelley/test/data/smash/api/v1/metadata/ec28f33dcbe6d6400a1e5e339bd0647c0973ca6c0cf9c2bbe6838dc6/af7789093aa1fbff0f6e7026ecdbbd470f04e9a0415e1e7e6b9f1d54cd353126
```diff
@@ -0,0 +1 @@
+{"homepage":"https://iohk.io","name":"Genesis Pool A","ticker":"GPA","description":null}
```
KtorZ: I'd be better to not duplicate test data here and rely on one single source of truth. Metadata for pools are declared in: 

https://github.com/input-output-hk/cardano-wallet/blob/0fb513921ff14cfc53aa3d508b170c4d2c24c878/lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs#L1402-L1513

Metadata are written to disk during the test cluster setup for each stake pool when the corresponding stake pool is registered, and a static server is started to serve them in the "direct" mode. 

https://github.com/input-output-hk/cardano-wallet/blob/0fb513921ff14cfc53aa3d508b170c4d2c24c878/lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs#L788-L820

We should use a similar strategy for SMASH. It's okay to read the entire MVar upfront and write all metadata in temporary directory for the SMASH stub.

KtorZ (changes requested):

lib/shelley/test/data/smash/api/v1/metadata/ec28f33dcbe6d6400a1e5e339bd0647c0973ca6c0cf9c2bbe6838dc6/af7789093aa1fbff0f6e7026ecdbbd470f04e9a0415e1e7e6b9f1d54cd353126
hasufell: I suggest that the files be the source of truth then, not the haskell code. That also avoids hide and seek with metadata hash, which currently depends on the aeson encoding.

KtorZ: > That also avoids hide and seek with metadata hash, which currently depends on the aeson encoding.

This is why we use code for this, so it is easily reproducible and does not depend on someone's editor behavior. Please stick to the current pattern and use the Haskell code as source to generate files.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
hasufell: Should be fixed

lib/shelley/test/integration/Main.hs
```diff
@@ -149,22 +150,25 @@ main = withUtf8Encoding $ withTracers $ \tracers -> do
...
+                    ByronMigrations.spec @n
+                    Transactions.spec @n
+                    Network.spec
+                    Network_.spec
+                    StakePools.spec @n
+                    ByronTransactions.spec @n
+                    ByronHWWallets.spec @n
+
+                -- possible conflict with StakePools
                 Settings.spec @n
```
KtorZ, line 172: :+1:

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -1120,7 +1097,56 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
...
+            updateMetadataSource ctx (T.pack smashUrl)
```
KtorZ, line 1104: We should run the test in a form of "bracket-style resource handler", which set the metadataSource back to what it was after the test. Otherwise it may have an influence on other tests depending on which one runs first.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -1120,7 +1098,56 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
...
+                                pools'
+                        metadataActual
+                            `shouldSatisfy` (`Set.isSubsetOf` metadataPossible)
+                        metadataActual
+                            `shouldSatisfy` (not . Set.null)
+                        (fmap (getApiT . view #id) delistedPools)
+                            `shouldBe` [PoolId . unsafeFromHex $
+                                "b45768c1a2da4bd13ebcaa1ea51408eda31dcc21765ccbd407cda9f2"]
+                    ]
+
```
piotr-iohk, line 1122: Maybe we could add additional check to make sure there are no delisted pools when we change to, say "direct":
```haskell
updateMetadataSource ctx "direct"
eventually "metadata is fetched - there are no delisted pools" $ do
                r <- listPools ctx arbitraryStake
                verify r [no pool is delisted]
```

piotr-iohk (approved):
LGTM

lib/core/test/unit/Cardano/Pool/DB/Properties.hs
```diff
@@ -1457,6 +1459,36 @@ prop_putLastMetadataGCReadLastMetadataGC DBLayer{..} posixTime = do
...
+    :: DBLayer IO
+    -> Set.Set PoolId
+    -> Property
+prop_removePoolMetadataDelistedPools DBLayer {..} pools =
+    monadicIO (setup >> prop)
+  where
+    setup = run $ atomically cleanDB
+
+    prop = do
+        run $ atomically $ putDelistedPools (Set.toList pools)
```
KtorZ, line 1473: Arguably, this :point_up: is `setup`, but I am neat-picking :)

KtorZ (approved):
Nice catch.

lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -607,6 +609,18 @@ updateMetadataSource ctx t = do
             }
        } |]
 
+bracketSettings :: (MonadIO m, MonadCatch m) => Context t -> m () -> m ()
+bracketSettings ctx action = do
```
KtorZ, line 613: We could have leveraged `resourcet` here for allocating / freeing settings.

PR #2321, opened 2020-11-16
rename data-types to reflect the 'Account' vs 'Wallet' duality

KtorZ:
# Issue Number

ADP-541

# Overview

  Renamed:

  Old Name                     | New Name
  ---                          | ---
  AccountPostData              | ApiAccountFromKey
  ApiByronWallet               | ApiByronAccount
  ApiByronWalletBalance        | ApiByronBalance
  ApiWallet                    | ApiAccount
  ApiWalletPassphraseInfo      | ApiEncryptionPassphraseInfo
  ApiWalletPassphrasea         | ApiEncryptionPassphrase
  ByronWalletFromXPrvPostData  | ApiByronAccountFromXPrv
  ByronWalletPostData          | ApiByronAccountFromPhrase
  ByronWalletPutPassphraseData | ApiByronAccountPutPassphraseData
  SomeByronWalletPostData      | ApiSomeByronAccountPostData
  WalletBalance                | Balance
  WalletId                     | AccountId
  WalletOrAccountPostData      | ApiAccountPostData
  WalletPostData               | ApiAccountFromPhrase
  WalletPutData                | ApiAccountPutData
  WalletPutPassphraseData      | AccountPutPassphraseData

  I've also renamed a few constructors to be more consistent with the
  new naming, where it made sense and where I could identify an obvious
  improvement.

# Comments


lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -428,13 +425,6 @@ spec = do
         validateEveryToJSONWithPatternChecker
             match
             (Proxy :: Proxy (Api ('Testnet 0) ApiStakePool))
-        -- NOTE See (ToSchema WalletOrAccountPostData)
```
Anviking, line 431: ?

Anviking (commented):
Looks good. I'm less concerned about the huge diff. Didn't see anything strange when I skimmed through the non-.json files, though I also assume find-replace works correctly. I'm more concerned with the DB Migration CI failure.

Also, wouldn't it be more sane to rename the golden json files than to re-generate them, to confirm that they still work?

KtorZ:
> Also, wouldn't it be more sane to rename the golden json files than to re-generate them, to confirm that they still work?

Actually, we had: 

- many generated golden json raising warnings because the content of the file was no longer matching the latest generated data (with the same seed) because of generators which changed. 

- many versioned files no longer matching any existing data-types (around 20 of them). 

- those I needed to rename according to the data-types renaming. 

So all in all, I went for purging them all and starting over on good foundation.

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
KtorZ: No longer useful since the upgrade to OpenAPI 3.0

PR #2327, opened 2020-11-17
Show deposits for fees and transactions

hasufell:
# Issue Number

ADP-470


lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -502,6 +502,8 @@ fromShelleyPParams bound pp = W.ProtocolParameters
         desiredNumberOfStakePoolsFromPParams pp
     , minimumUTxOvalue =
         minimumUTxOvalueFromPParams pp
+    , poolDeposit = poolDepositFromPParams pp
```
hasufell: Add deposits to ProtocolParameters...

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -746,6 +747,7 @@ data ApiTransaction (n :: NetworkDiscriminant) = ApiTransaction
     , withdrawals :: ![ApiWithdrawal n]
     , status :: !(ApiT TxStatus)
     , metadata :: !ApiTxMetadata
+    , deposits :: ![Quantity "lovelace" Natural]
```
hasufell: adjust API types

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2070,9 +2075,13 @@ joinStakePool ctx currentEpoch knownPools pid poolStatus wid =
...
         liftIO $ traceWith tr $ MsgIsStakeKeyRegistered isKeyReg
 
+        dep <- fmap keyDeposit $
+            withExceptT ErrJoinStakePoolNoSuchWallet
+                $ readWalletProtocolParameters @ctx @s @k ctx wid
+
         return $ if isKeyReg
-            then Join pid
-            else RegisterKeyAndJoin pid
+            then (Join pid, dep)
```
hasufell: I think this is a prior bug?

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1421,7 +1421,12 @@ estimateFeeForDelegation ctx wid = db & \DBLayer{..} -> do
     let action = if isKeyReg then Join pid else RegisterKeyAndJoin pid
     let selectCoins = selectCoinsForDelegationFromUTxO @_ @t @k
             ctx utxo txp minUtxo action
-    estimateFeeForCoinSelection $ Fee . feeBalance <$> selectCoins
+
+    dep <- fmap keyDeposit $
+        withExceptT ErrSelectForDelegationNoSuchWallet
+            $ readWalletProtocolParameters @ctx @s @k ctx wid
+    estimateFeeForCoinSelection (if isKeyReg then getCoin dep else 0)
```
hasufell: Interesting part: here we actually return some non-zero deposit

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2070,9 +2075,13 @@ joinStakePool ctx currentEpoch knownPools pid poolStatus wid =
...
 
+        dep <- fmap keyDeposit $
+            withExceptT ErrJoinStakePoolNoSuchWallet
+                $ readWalletProtocolParameters @ctx @s @k ctx wid
+
         return $ if isKeyReg
-            then Join pid
-            else RegisterKeyAndJoin pid
+            then (Join pid, 0)
+            else (RegisterKeyAndJoin pid, dep)
```
hasufell: Interesting part: here we actually return some non-zero deposit

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2070,9 +2075,13 @@ joinStakePool ctx currentEpoch knownPools pid poolStatus wid =
...
         liftIO $ traceWith tr $ MsgIsStakeKeyRegistered isKeyReg
 
+        dep <- fmap keyDeposit $
+            withExceptT ErrJoinStakePoolNoSuchWallet
+                $ readWalletProtocolParameters @ctx @s @k ctx wid
+
         return $ if isKeyReg
-            then Join pid
-            else RegisterKeyAndJoin pid
+            then (Join pid, 0)
```
hasufell: I think Join and RegisterKeyAndJoin need to be flipped here? Is that a bug?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1197,7 +1197,8 @@ selectCoinsForJoin ctx knownPools getPoolStatus pid wid = do
     curEpoch <- getCurrentEpoch ctx
 
     (utx, action, path) <- withWorkerCtx ctx wid liftE liftE $ \wrk -> do
-        action <- liftHandler
+        -- we never register stake key here, so deposit is irrelevant
+        (action, _) <- liftHandler
```
hasufell: this is a little fishy... but currently there's no way to trigger a register stake key via this endpoint, so adding deposit to CoinSelection types seems excessive

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2049,7 +2054,7 @@ joinStakePool
     -> PoolId
     -> PoolLifeCycleStatus
     -> WalletId
-    -> ExceptT ErrJoinStakePool IO DelegationAction
+    -> ExceptT ErrJoinStakePool IO (DelegationAction, Coin)
```
hasufell: Maybe a bit inelegant, but don't see a way to stuff it into DelegationAction and creating a separate type also seems excessive.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1457,11 +1459,12 @@ mkApiTransactionFromInfo
     -> TransactionInfo
     -> m (ApiTransaction n)
 mkApiTransactionFromInfo ti (TransactionInfo txid ins outs ws meta depth txtime txmeta) = do
-    apiTx <- mkApiTransaction ti txid (drop2nd <$> ins) outs ws (meta, txtime) txmeta $
-        case meta ^. #status of
+    apiTx <- mkApiTransaction ti txid (drop2nd <$> ins) outs ws (meta, txtime) txmeta
+        (case meta ^. #status of
```
hasufell: Funny thing... I couldn't move this stuff into a let or where binding without getting compile errors

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
```diff
@@ -173,6 +173,8 @@ ProtocolParameters
     protocolParametersDecentralizationLevel Percentage      sql=decentralization_level
     protocolParametersDesiredNumberOfPools  Word16          sql=desired_pool_number
     protocolParametersMinimumUtxoValue      W.Coin          sql=minimum_utxo_value
+    protocolParametersPoolDeposit           W.Coin          sql=pool_deposit
```
KtorZ: This is superfluous for cardano-wallet, we do never register stake pools and won't do in any near nor distant future.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -502,6 +502,8 @@ fromShelleyPParams bound pp = W.ProtocolParameters
...
+    , keyDeposit = keyDepositFromPParams pp
```
KtorZ: stakeKeyDeposit would be a slightly better name, there are many things that can qualify as "key" in the application.

specifications/api/swagger.yaml
```diff
@@ -1322,9 +1322,11 @@ components:
       required:
         - estimated_min
         - estimated_max
+        - deposit
       properties:
         estimated_min: *transactionAmount
         estimated_max: *transactionAmount
+        deposit: *transactionAmount
```
KtorZ, line 1328: For consistency, what about having a list of `deposits` here too ?

hasufell: Is it possible to have theoretically multiple for a fee estimation?

KtorZ: Theoretically yes. If we were to register multiple stake keys. Arguably, you could represent them as "one big deposit", but we're already taking the approach of showing them as multiple deposits in transactions.

piotr-iohk:
It seems that there are some DB migrations missing:
  - start cardano-wallet v2020-11-17 with some wallets
  - re-start wallet on the same wallet db with cardano-wallet from branch:
 ```
 [cardano-wallet.wallet-db:Notice:18] [2020-11-20 08:22:28.21 UTC] wallet table does not contain required field 'passphrase_scheme'. Adding this field with a default value of "EncryptWithPBKDF2".
[cardano-wallet.wallet-db:Notice:18] [2020-11-20 08:22:28.21 UTC] rnd_state_address table does not contain required field 'account_ix'. Adding this field with a default value of 2147483648.
[cardano-wallet.wallet-db:Error:18] [2020-11-20 08:22:28.22 UTC] Failed to migrate the database: : NOT NULL constraint failed: protocol_parameters_backup.key_deposit
[cardano-wallet.wallet-engine:Error:18] [2020-11-20 08:22:28.22 UTC] Worker has exited unexpectedly: MigrationError {getMigrationErrorMessage = ": NOT NULL constraint failed: protocol_parameters_backup.key_deposit"}
[cardano-wallet.wallet-db:Info:4] [2020-11-20 08:22:28.22 UTC] Found existing wallet: 1ceb45b37a94c7022837b5ca14045f11a5927c65
[cardano-wallet.wallet-db:Info:4] [2020-11-20 08:22:28.22 UTC] Found existing wallet: 99d0553519f56f32e54ed1404d421cb7eae3c541
[cardano-wallet.wallet-db:Notice:31] [2020-11-20 08:22:28.23 UTC] wallet table does not contain required field 'passphrase_scheme'. Adding this field with a default value of "EncryptWithPBKDF2".
[cardano-wallet.wallet-db:Notice:31] [2020-11-20 08:22:28.23 UTC] rnd_state_address table does not contain required field 'account_ix'. Adding this field with a default value of 2147483648.
[cardano-wallet.wallet-db:Error:31] [2020-11-20 08:22:28.23 UTC] Failed to migrate the database: : NOT NULL constraint failed: protocol_parameters_backup.key_deposit
[cardano-wallet.wallet-engine:Error:31] [2020-11-20 08:22:28.23 UTC] Worker has exited unexpectedly: MigrationError {getMigrationErrorMessage = ": NOT NULL constraint failed: protocol_parameters_backup.key_deposit"}
[cardano-wallet.wallet-db:Notice:32] [2020-11-20 08:22:28.24 UTC] wallet table does not contain required field 'passphrase_scheme'. Adding this field with a default value of "EncryptWithPBKDF2".
[cardano-wallet.wallet-db:Notice:32] [2020-11-20 08:22:28.24 UTC] rnd_state_address table does not contain required field 'account_ix'. Adding this field with a default value of 2147483648.
[cardano-wallet.wallet-db:Error:32] [2020-11-20 08:22:28.24 UTC] Failed to migrate the database: : NOT NULL constraint failed: protocol_parameters_backup.key_deposit
[cardano-wallet.network:Info:15] [2020-11-20 08:22:28.24 UTC] Querying the reward account balance for [] at 70617265<-[d39a9289-11491327#2057619]
[cardano-wallet.network:Info:27] [2020-11-20 08:22:28.24 UTC] Query GetEraStart took 0.000578931s
[cardano-wallet.network:Info:15] [2020-11-20 08:22:28.24 UTC] Query getAccountBalance took 0.000697062s
[cardano-wallet.network:Info:15] [2020-11-20 08:22:28.24 UTC]   delegations = fromList []
  rewards = fromList []

[cardano-wallet.wallet-engine:Error:32] [2020-11-20 08:22:28.24 UTC] Worker has exited unexpectedly: MigrationError {getMigrationErrorMessage = ": NOT NULL constraint failed: protocol_parameters_backup.key_deposit"}

 ```
 As a result there are no wallets reported although there were there before:
 ```
$ curl -X GET http://localhost:8090/v2/wallets
[]
 ```

piotr-iohk:
As far as I see on a wallet which is **not** delegating, the reported deposit is 0:
```
curl -X GET http://localhost:8090/v2/wallets/d36d116a31202fd3d10de6a629a293674a5ea400/delegation-fees

{"estimated_min":{"quantity":174257,"unit":"lovelace"},
  "deposits":[{"quantity":0,"unit":"lovelace"}],
  "estimated_max":{"quantity":174257,"unit":"lovelace"}}
```

While on the wallet that is **already** delegating the deposit is reported as 2000000 lovelace

```
curl -X GET http://localhost:8090/v2/wallets/1ceb45b37a94c7022837b5ca14045f11a5927c65/delegation-fees

{"estimated_min":{"quantity":172761,"unit":"lovelace"},
 "deposits":[{"quantity":2000000,"unit":"lovelace"}],
 "estimated_max":{"quantity":1000000,"unit":"lovelace"}}
```

Shouldn't it be the other way around?

piotr-iohk:
Perhaps this is not yet implemented, but deposit seems to be always empty on actual transaction (no matter if it is join/quit stake pool) and no matter if wallet is already delegating or not. 

Transaction always has:
```
"deposits":[]
```

Also question:
As far as I see we show deposit on:
 - delegation fee
 - transaction fee (for Byron and Shelley)
 - transaction (whether it is normal tx, or join/quit pool tx)

Do we want to consider showing deposit also on migration cost ([shelley migration](https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/getShelleyWalletMigrationInfo) and [byron migration](https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/getByronWalletMigrationInfo)) for consistency (essentially migration is a bunch of transactions), I imagine that for now it will be always 0, not sure if possible to be otherwise in future in any circumstances.

lib/core/src/Cardano/Wallet.hs
KtorZ: Why is that interesting :thinking: ?

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2118,6 +2127,8 @@ data FeeEstimation = FeeEstimation
     -- ^ Most coin selections will result in a fee higher than this.
     , estMaxFee :: Word64
     -- ^ Most coin selections will result in a fee lower than this.
+    , estDeposit :: Word64
+    -- ^ Deposit if stake key was registered
```
KtorZ: This data-type feels a little bit _abused_. There's nothing of an estimation when it comes to stake key deposits. The deposit value is fixed by the network.

lib/core/src/Cardano/Wallet/Api/Server.hs
KtorZ: > there's no way to trigger a register stake key via this endpoint

I challenge this. When call for the very first time so that this is the first delegation for this wallet, it will result in a stake key registration. I do not understand your rationale here?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1461,11 +1463,12 @@ mkApiTransactionFromInfo
...
             Pending  -> #pendingSince
             InLedger -> #insertedAt
-            Expired  -> #pendingSince
+            Expired  -> #pendingSince)
+        []
```
KtorZ: To help readability, consider making this a named argument:

```
    let deposits = []
    apiTx <- mkApiTransaction ... deposits
```

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: What do you mean ?

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
```diff
@@ -176,6 +176,7 @@ ProtocolParameters
     protocolParametersDecentralizationLevel Percentage      sql=decentralization_level
     protocolParametersDesiredNumberOfPools  Word16          sql=desired_pool_number
     protocolParametersMinimumUtxoValue      W.Coin          sql=minimum_utxo_value
+    protocolParametersKeyDeposit            W.Coin          sql=key_deposit
```
KtorZ: Sadly, this needs to be added as the very last. `persistent` does not "tolerate" adding new columns in the middle of existing declarations.

KtorZ (changes requested):
One thing important to notice is that, we are conflating the "key deposits" with the third fee component. The main reason for that was to accommodate both Jormungandr  and Cardano-node requirements: Jormungandr does not have a concept of "deposit", but instead, has an extra "certificate fee" which needs to be applied for each certificate. 

That is, we've defined transaction fees as: a + b*x + c*y where `x` and `y` depends on cardano-node / jormungandr. Now that jormungandr is gone, this is a little bit awkward because the key deposit is not exactly a fee. Plus, this PR now introduces yet-another-way to view the key deposit value (via the protocol parameters) which feels like the right way. We should consider revising the existing design to have one consistent view on key deposits. 

_I think_ the main change will be in the TransactionLayer initialization: 

```hs
    _initDelegationSelection
        :: FeePolicy
            -- Current fee policy
        -> DelegationAction
            -- What sort of action is going on
        -> CoinSelection
        -- ^ An initial selection where 'deposit' and/or 'reclaim' have been set
        -- accordingly.
    _initDelegationSelection (LinearFee _ _ (Quantity c)) = \case
        Quit{} -> mempty { reclaim = round c }
        Join{} -> mempty
        RegisterKeyAndJoin{} -> mempty { deposit = round c }
```

This `c` value would have to come from the protocol parameters more directly. Since the `c` component on the fee policy is now obsolete it makes sense to just remove it entirely and let the compiler guide us through places that are impacted.

lib/core/src/Cardano/Wallet/Api/Types.hs
hasufell: I meant this adjusted the API types :sweat_smile: ... these comments were meant as a walkthrough

lib/core/src/Cardano/Wallet.hs
hasufell: :sweat_smile: because here happens something. Most of the PR is types fixing. I wanted to point out where actual business logic is changed/added.

lib/core/src/Cardano/Wallet/Api/Server.hs
hasufell:
```hs
    coinSelections :: Server (CoinSelections n)
    coinSelections = (\wid ascd -> case ascd of
        (ApiSelectForPayment ascp) ->
            selectCoins shelley (delegationAddress @n) wid ascp
        (ApiSelectForDelegation (ApiSelectCoinsAction (ApiT action))) ->
            case action of
                Join pid -> selectCoinsForJoin @_ @()
                    shelley
                    (knownPools spl)
                    (getPoolLifeCycleStatus spl)
                    pid
                    (getApiT wid)
                -- if this would be enabled, we'd need to adjust selectCoinsForJoin
                -- to account for deposit
                RegisterKeyAndJoin _ -> throwError err400
                Quit -> selectCoinsForQuit @_ @() shelley wid
        )
```

`RegisterStakeKeyAndJoin` isn't allowed here.

KtorZ: Yes, and this is why re-using this type was a bit hacky :upside_down_face: ... This does not translate what's happening in the actual underlying logic. Calls from `selectCoinsForJoin` can totally generate stake key registration with implied key deposits. 

I'd suggest defining a proper api type to reflect this and prevent someone else to be bitten by this. The API action that are supported are either Join/Quit and they map internally to Join / RegisterKeyAndJoin / Quit.

hasufell: I've done:

1. Use a separate Api type for delegation action

```hs
data DelegationActionType = Join | Quit
    deriving (Eq, Generic, Show)
    deriving anyclass NFData

data ApiDelegationAction = ApiDelegationAction
    { action :: DelegationActionType
    , pool :: Maybe (ApiT PoolId)
    }
    deriving (Eq, Generic, Show)
    deriving anyclass NFData
```

2. add `deposits` field to coin selection

```hs
    , outputs :: ![ApiCoinSelectionOutput n]
    , change :: ![ApiCoinSelectionChange n]
    , certificates :: Maybe (NonEmpty ApiCertificate)
    , deposits :: ![Quantity "lovelace" Natural]
      deriving anyclass NFData
```

hasufell:
`_initDelegationSelection` has been revised

hasufell:
> While on the wallet that is already delegating the deposit is reported as 2000000 lovelace

Yep, that's because the code has a confusing `isKeyReg` function, which denotes whether a key is *already* registered, not whether one needs to be registered. I assumed the latter :sweat_smile:

hasufell:
@piotr-iohk can you try again?

piotr-iohk:
> @piotr-iohk can you try again?

As far as I see there is still the same issue wrt db migration -> https://github.com/input-output-hk/cardano-wallet/pull/2327#issuecomment-731026034

But even when starting fresh and creating a new wallet I get a lot of errors in the log (for both wallet and node errors in log rolling very fast) and wallets sync very slowly:

cardano-wallet log:
```
[cardano-wallet.wallet-engine:Error:74] [2020-11-24 11:07:37.88 UTC] 743f4884: Unexpected error following the chain: PersistMarshalError "Couldn't parse field `protocolParametersFeePolicy` from table `protocol_parameters`. not a valid value: PersistText \"155381.0 + 43.946x + \""
[cardano-wallet.network:Notice:74] [2020-11-24 11:07:37.88 UTC] Destroying cursor connection at ThreadId 109088

```

and cardano-node log:

```
[piotr-la:cardano.node.LocalErrorPolicy:Error:52] [2020-11-24 11:08:16.19 UTC] [String "ErrorPolicyUnhandledApplicationException (MuxError MuxBearerClosed \"<socket: 272> closed when reading data, waiting on next header True\")",String "ErrorPolicyTrace",String "LocalAddress {getFilePath = \"\"}"]

```

At some point I even got a crash on the node.

hasufell:
@piotr-iohk I believe I found the problem (and fixed it).

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1879,7 +1885,7 @@ mkTxMeta interpretTime blockHeader wState tx cs expiry =
         amtInps
             = sum (fromIntegral . getCoin . coin . snd <$> (inputs cs))
             + sum (mapMaybe ourWithdrawal $ Map.toList $ withdrawals tx)
-            + fromIntegral (reclaim cs)
+            + fromIntegral (CoinSelection.deposit cs)
```
KtorZ: This is wrong. Deposits and reclaims aren't the same thing. Deposits counts towards the `output` balance actually.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1199,8 +1199,9 @@ selectCoinsForJoin ctx knownPools getPoolStatus pid wid = do
     pools <- liftIO knownPools
     curEpoch <- getCurrentEpoch ctx
 
-    (utx, action, path) <- withWorkerCtx ctx wid liftE liftE $ \wrk -> do
-        action <- liftHandler
+    (utx, action, path, dep) <- withWorkerCtx ctx wid liftE liftE $ \wrk -> do
+        -- we never register stake key here, so deposit is irrelevant
```
KtorZ, line 1198: This comment seems no longer accurate?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1460,11 +1462,13 @@ mkApiTransactionFromInfo
...
-    apiTx <- mkApiTransaction ti txid (drop2nd <$> ins) outs ws (meta, txtime) txmeta $
-        case meta ^. #status of
+    let deposits = []
+    apiTx <- mkApiTransaction ti txid (drop2nd <$> ins) outs ws (meta, txtime) txmeta
+        (case meta ^. #status of
             Pending  -> #pendingSince
             InLedger -> #insertedAt
-            Expired  -> #pendingSince
+            Expired  -> #pendingSince)
+        deposits
```
KtorZ: Small stylistic note: I'd suggest to position the deposit argument before the last argument for better readability. Arguably, that function already has way too many arguments and a record would be better in the first place.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -481,11 +481,23 @@ data ApiCertificate
     deriving (Eq, Generic, Show)
     deriving anyclass NFData
 
+data DelegationActionType = Join | Quit
```
KtorZ: *ApiDelegationActionType

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -481,11 +481,23 @@ data ApiCertificate
...
+    deriving (Eq, Generic, Show)
+    deriving anyclass NFData
+
+data ApiDelegationAction = ApiDelegationAction
+    { action :: DelegationActionType
+    , pool :: Maybe (ApiT PoolId)
```
KtorZ: Why not: 

```hs
 data ApiDelegationActionType = Join PoolId | Quit 
```

to avoid situations where we have action = Quit and pool = Just {} , or reciprocally when we have action = Joint but pool = Nothing ... which wouldn't mean anything? Was this to work-around the JSON instance? Note that the same instance can be achieved with `Join PoolId | Quit` by using the following serialization options: 

```hs
apiDelegationActionTypeOptions :: Aeson.Options
apiDelegationActionTypeOptions = Aeson.defaultOptions
      { tagSingleConstructors = True
      , omitNothingFields = True
      , constructorTagModifier = camelTo2 '_'
      , fieldLabelModifier = camelTo2 '_' . dropWhile (== '_')
      , sumEncoding = TaggedObject
          { tagFieldName = "action"
          , contentsFieldName = "pool" 
          }
      }
```

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -497,6 +499,15 @@ migrateManually tr proxy defaultFieldValues =
...
             Just v -> T.pack $ show $ W.unEpochNo v
 
+    -- | Adds a 'key_deposit column to the 'protocol_parameters'
+    -- table if it is missing.
+    --
+    addKeyDepositIfMissing :: Sqlite.Connection -> IO ()
+    addKeyDepositIfMissing conn = do
+        addColumn_ conn True (DBField ProtocolParametersKeyDeposit) value
+      where
+        value = T.pack $ show $ W.getCoin $ defaultKeyDeposit defaultFieldValues
```
KtorZ:
```suggestion
        value = toText $ defaultKeyDeposit defaultFieldValues
```

lib/core/test/data/Cardano/Wallet/Api/ApiCoinSelectionTestnet0.json
```diff
@@ -1,1961 +1,4893 @@
 {
-    "seed": 1426841331592113606,
+    "seed": -4641725427390187379,
```
KtorZ: That file is way too large, but that's tangent to this PR.

lib/core/test/data/Cardano/Wallet/Api/ApiDelegationAction.json
```diff
@@ -0,0 +1,44 @@
...
+    "seed": -6826805226140653328,
+    "samples": [
+        {
+            "action": "join",
+            "pool": "pool1uhaqd3v0qn7zyxhvd3epu42tegsuvlu3h7f0s3av4csmgayxr95"
+        },
+        {
+            "action": "quit",
+            "pool": "pool17qkm5ns0r2zl4gwrr4ne50kypxcvcdcq7956xu7jhm6f28pg5c8"
+        },
```
KtorZ, line 9: The meaning of this one (quit + pool) is dubious.

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -1634,6 +1648,7 @@ instance Arbitrary (ApiTransaction t) where
             <*> genWithdrawals
             <*> pure txStatus
             <*> arbitrary
+            <*> arbitrary
```
KtorZ: Note: no need to generate infinitely large lists here.

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
```diff
@@ -236,16 +236,19 @@ server byron icarus shelley spl ntp =
...
                     (getPoolLifeCycleStatus spl)
-                    pid
+                    (getApiT pid)
                     (getApiT wid)
-                RegisterKeyAndJoin _ -> throwError err400
-                Quit -> selectCoinsForQuit @_ @() shelley wid
+                (Join, Nothing) -> Handler $ throwE $ apiError err400 BadRequest $
+                    mconcat
+                    [ "Missing PoolId for Join action."
+                    ]
```
KtorZ: :upside_down_face:

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -223,17 +222,8 @@ newTransactionLayer networkId = TransactionLayer
...
-            -- What sort of action is going on
-        -> CoinSelection
-        -- ^ An initial selection where 'deposit' and/or 'reclaim' have been set
-        -- accordingly.
-    _initDelegationSelection (LinearFee _ _ (Quantity c)) = \case
-        Quit{} -> mempty { reclaim = round c }
-        Join{} -> mempty
-        RegisterKeyAndJoin{} -> mempty { deposit = round c }
+        :: CoinSelection
+    _initDelegationSelection = mempty
```
KtorZ: I am not sure to understand this change? what happened to the reclaim and the deposit needed for the coin selection to run correctly?

KtorZ (changes requested):

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
hasufell: ?

KtorZ: https://github.com/input-output-hk/cardano-wallet/pull/2327#discussion_r532434912

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
hasufell: I'm also wondering why I did that. But what's even more interesting is why the tests pass

https://github.com/input-output-hk/cardano-wallet/pull/2327/files#diff-0a90babe9bf43651092fb58d3a37bd24cc2ca0a6d63f948c6c64c738a27f5392L1392

hasufell: Ah, now I remember... we removed the 3rd parameter for LinearFee https://github.com/input-output-hk/cardano-wallet/pull/2327/files#diff-45c6b9a78752ca17442ffe1ded898e5fd2f6931f95c9ae4ad21d30afcf0e2e1fL810 because that only existed because of jurmungandr as far as I remember the conversation

lib/core/src/Cardano/Wallet.hs
hasufell: Absolutely. I'm also not sure what was going on here.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1875,11 +1881,13 @@ mkTxMeta interpretTime blockHeader wState tx cs expiry =
     let
         amtOuts =
             sum (mapMaybe ourCoins (outputs cs))
+            + fromIntegral (CoinSelection.deposit cs)
```
hasufell: @KtorZ

KtorZ: Yes? What/why is this change?

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -1634,6 +1644,7 @@ instance Arbitrary (ApiTransaction t) where
             <*> genWithdrawals
             <*> pure txStatus
             <*> arbitrary
+            <*> resize 10 (listOf arbitrary)
```
KtorZ: > **resize:** _Overrides the size parameter. Returns a generator which uses the given size instead of the runtime-size parameter._

I am pretty sure that this is not what we want here :no_mouth:

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
KtorZ: Yes okay, the third parameter was removed from the linear fee, but it's because it now comes from the protocol parameters directly as the stake key deposit. The logic that applied here still applies however!

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
hasufell: Why?

This means that the *maximum* size of the list is going to be 10. You wanted the list to not be arbitrary length.

lib/core/src/Cardano/Wallet.hs
hasufell: I wanted to know if this is correct, because of your comment https://github.com/input-output-hk/cardano-wallet/pull/2327#discussion_r532431872

KtorZ: It is not. The current code is doing just fine :+1:

KtorZ: (This calculation happens on the transaction outputs directly, so the deposit is already _implicitly_  included in the difference between inputs and outputs).

KtorZ: Although, the point of this PR / Story is now to make it possible to identify whether the delta (in - out) is purely fee, or fee and deposits.

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
KtorZ: It'll effectively work because of how `listOf` behaves. I just found it a bit misleading because `resize 10` means that the size parameter will always be `10` whereas QuickCheck typically grow the size parameter to be larger and larger and produce more and more complex cases. `resize 10` circumvent that by forcing the size to always be the same (which acts as a max-size for lists and most other structures, true, but misleading).

piotr-iohk:
@hasufell doing some manual testing on testnet I encountered few problems wrt migration from previous versions of wallet and also delegation for the first time from a wallet, pls see details below.


- migrating to this branch from `v2020-11-17` (with some wallets) there is :x: :

```
[cardano-wallet.wallet-engine:Error:31] [2020-12-01 07:49:21.42 UTC] 94c0af10: Unexpected error following the chain: PersistMarshalError "Couldn't parse field `protocolParametersFeePolicy` from table `protocol_parameters`. not a valid value: PersistText \"155381.0 + 44.0x + 2000000.0y\""
```
Doing:
```
curl -X GET http://localhost:8090/v2/wallets/1ceb45b37a94c7022837b5ca14045f11a5927c65/delegation-fees
Something went wrong
```

 - migrating to this branch from `v2020-11-26` (with some wallets) there is no error in the logs, but doing :x: :

```
curl -X GET http://localhost:8090/v2/wallets/1ceb45b37a94c7022837b5ca14045f11a5927c65/delegation-fees
Something went wrong
```

and then in the log there is:

```
[GET] /v2/wallets/1ceb45b37a94c7022837b5ca14045f11a5927c65/delegation-fees
PersistMarshalError "Couldn't parse field `protocolParametersFeePolicy` from table `protocol_parameters`. not a valid value: PersistText \"155381.0 + 44.0x + 2000000.0y\""
```

- starting fresh:
  - normal tx fee (on shelley/icarus/byron) - :heavy_check_mark: (deposit empty)
  - normal tx (on shelley/icarus/byron) - :heavy_check_mark: (deposit empty)
  - delegation fees on wallet that does not delegate - :heavy_check_mark: (deposit 2 ada shown)
  - delegation fees on wallet that already delegates - :heavy_check_mark: (deposit empty)
  - redelegation - :heavy_check_mark: (no deposit on tx)
  - first delegation - :x: 
```
curl -X PUT http://localhost:8090/v2/stake-pools/pool1d03p2xfdcq09efx0hgy4jkr0tqdgvklues5cg3ud45t9wndafmm/wallets/1c36d04097d8300f4a1a4e2344a2281662b2b472 \
-d '{"passphrase":"Secure Passphrase"}' \


{
  "code": "created_invalid_transaction",
  "message": "That's embarrassing. It looks like I've created an invalid transaction that could not be parsed by the node. Here's an error message that may help with debugging: HardForkApplyTxErrFromEra S (Z (WrapApplyTxErr {unwrapApplyTxErr = ApplyTxError [LedgerFailure (DelegsFailure (DelplFailure (DelegFailure (StakeDelegationImpossibleDELEG (KeyHashObj (KeyHash \"10711397ce64528c54712fa1e242f730302c64b4a6df142c2f985e77\"))))))]}))"
}

```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1420,7 +1420,12 @@ estimateFeeForDelegation ctx wid = db & \DBLayer{..} -> do
     let action = if isKeyReg then Join pid else RegisterKeyAndJoin pid
     let selectCoins = selectCoinsForDelegationFromUTxO @_ @t @k
             ctx utxo txp minUtxo action
-    estimateFeeForCoinSelection $ Fee . feeBalance <$> selectCoins
+
+    dep <- fmap stakeKeyDeposit $
+        withExceptT ErrSelectForDelegationNoSuchWallet
+            $ readWalletProtocolParameters @ctx @s @k ctx wid
```
KtorZ, line 1430: Note: this needs to be fed into the `initDelegationSelection` function to correctly initialize a coin selection with appropriate deposits (or reclaim). The only real difference with _before_ is that the deposit amount now comes directly from the "stakeKeyDeposit" protocol parameter, and not from the fee policy. Logic-wise, the rest should not change.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1420,7 +1420,12 @@ estimateFeeForDelegation ctx wid = db & \DBLayer{..} -> do
...
+    estimateFeeForCoinSelection (if isKeyReg then Nothing else Just $ getCoin dep)
```
KtorZ: Note that passing this to the `estimateFeeForCoinSelection` function is a bit redundant with the coin selection itself. A coin selection is captured by:

```hs
data CoinSelection = CoinSelection
    { inputs  :: [(TxIn, TxOut)]
      -- ^ Picked inputs
    , withdrawal :: Word64
      -- ^ An available withdrawal amount, counting as an extra input
    , reclaim :: Word64
      -- ^ Claim back a deposit, counting as a an extra input
    , outputs :: [TxOut]
      -- ^ Picked outputs
    , change  :: [Coin]
      -- ^ Resulting changes
    , deposit :: Word64
      -- ^ A deposit counting as an extra output
    } deriving (Generic, Show, Eq)
```

So, the presence and full value of a deposit can be inferred from a `CoinSelection`. Therefore, a better approach which does not duplicate any logic with regards to deposits would be to change `feeBalance` to return not only a `Fee`, but also a deposit value and have that bubbling through `estimateFeeForCoinSelection`.

lib/core/src/Cardano/Wallet.hs
hasufell: It is:

```hs
    let selectCoins = selectCoinsForDelegationFromUTxO @_ @t @k
            ctx utxo txp minUtxo dep action
```

lib/core/src/Cardano/Wallet.hs
hasufell: Doing it in `feeBalance` is quite intrusive. I believe it's best to do it in a lambda. So:

```hs
    estimateFeeForCoinSelection isKeyReg
        $ (\cs -> (Fee $ feeBalance cs, CoinSelection.deposit cs)) <$> selectCoins
```


----

Edit: I think it's more annoying than useful. `estimateFeeForCoinSelection` repeats the action, so stuffing it into a tuple with fee is just confusing here.

hasufell:
> doing some manual testing on testnet I encountered few problems wrt migration from previous versions of wallet and also delegation for the first time from a wallet, pls see details below.

I think I found the culprit, we were missing an actual value migration, I added:

```hs
    updateFeeValue :: Sqlite.Connection -> IO ()
    updateFeeValue conn = do
        fee_policyInfo <- Sqlite.prepare conn $ T.unwords
            [ "SELECT fee_policy FROM protocol_parameters"
            , ";"
            ]
        row <- Sqlite.step fee_policyInfo
            >> Sqlite.columns fee_policyInfo
        Sqlite.finalize fee_policyInfo
        case row of
            [PersistText t] -> do
                (a:b:_) <- pure (T.splitOn " + " t)
                let newVal = a <> " + " <> b
                query <- Sqlite.prepare conn $ T.unwords
                    [ "UPDATE protocol_parameters"
                    , "SET fee_policy", "=", newVal
                    , ";"
                    ]
                Sqlite.step query *> Sqlite.finalize query
            _ -> pure ()
```
Do we need to set the new stakeKeyDeposit parameter here as well?
----

Why first delegation fails for you on fresh database, I have no clue.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -574,6 +587,27 @@ migrateManually tr proxy defaultFieldValues =
...
+            [PersistText t] -> do
+                (a:b:_) <- pure (T.splitOn " + " t)
+                let newVal = a <> " + " <> b
+                query <- Sqlite.prepare conn $ T.unwords
+                    [ "UPDATE protocol_parameters"
+                    , "SET fee_policy", "=", newVal
+                    , ";"
+                    ]
+                Sqlite.step query *> Sqlite.finalize query
+            _ -> pure ()
```
KtorZ: Please stick to the existing patterns with regards to database migrations: 

a) Avoid hard-coded column and table names, and use instead the `DBField` data-type for existing database types.

b) Define proper logging messages depending on the situation (migration needed vs migration not needed)

c) Make sure to add a unit test for testing the migration (see examples in https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs#L250-L272) 

d) Make sure to start your migration query with an `isFieldPresent` request. The database may be older than you think.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -223,17 +223,17 @@ newTransactionLayer networkId = TransactionLayer
...
         -> CoinSelection
         -- ^ An initial selection where 'deposit' and/or 'reclaim' have been set
         -- accordingly.
-    _initDelegationSelection (LinearFee _ _ (Quantity c)) = \case
-        Quit{} -> mempty { reclaim = round c }
+    _initDelegationSelection (Coin c) = \case
+        Quit{} -> mempty { reclaim = c }
         Join{} -> mempty
-        RegisterKeyAndJoin{} -> mempty { deposit = round c }
+        RegisterKeyAndJoin{} -> mempty { deposit = c }
```
KtorZ, line 284: :+1:

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -589,24 +589,77 @@ migrateManually tr proxy defaultFieldValues =
...
+                    , tableName field
+                    , ";"
+                    ]
+                row <- Sqlite.step fee_policyInfo
+                    >> Sqlite.columns fee_policyInfo
+                Sqlite.finalize fee_policyInfo
+                case row of
+                    [PersistText t] -> do
+                        traceWith tr $ MsgManualMigrationNeeded field t
+                        [a, b, c] <- pure (T.splitOn " + " t)
```
hasufell: Migrations don't need to be idempotent, do they? They're only ever run once, right?

Because if we would run this migration twice, this pattern match would fail.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -589,24 +589,77 @@ migrateManually tr proxy defaultFieldValues =
...
+                            ]
+                        Sqlite.step query *> Sqlite.finalize query
+
+                        -- update stake key val
+                        updateStakeKeyVal (DBField ProtocolParametersKeyDeposit) c
+
+                    _ -> fail "Couldn't query fee value!"
+      where
+        updateStakeKeyVal :: DBField -> Text -> IO ()
+        updateStakeKeyVal stakeField c = do
```
hasufell: So I decided to also extract and update the stake key value here, which is the old 3rd parameter of `FeePolicy` (which is now removed).

Please confirm this is what we want.

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
```diff
@@ -425,6 +431,43 @@ testMigrationPassphraseScheme = do
...
+        copyFile orig path
+        (logs, pp') <- captureLogging $ \tr -> do
+            withDBLayer @s @k tr defaultFieldValues (Just path) ti
+                $ \(_ctx, db) -> db & \DBLayer{..} -> atomically
+                $ do
+                    [wid] <- listWallets
+                    readProtocolParameters wid
+        let migrationMsg = filter isMsgManualMigration logs
+        length migrationMsg `shouldBe` 1
+        (getFeePolicy . txParameters <$> pp') `shouldBe`
```
hasufell: Here's the migration spec.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -589,24 +589,77 @@ migrateManually tr proxy defaultFieldValues =
...
+            isFieldPresent conn stakeField >>= \case
+                TableMissing ->
+                    -- this invariant must hold due to the parent context
+                    fail "ProtocolParameters table missing"
+                ColumnMissing -> do
+                    traceWith tr $ MsgManualMigrationNotNeeded stakeField
```
hasufell: Questionable, but well...

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -589,24 +589,77 @@ migrateManually tr proxy defaultFieldValues =
...
-            ]
-        row <- Sqlite.step fee_policyInfo
-            >> Sqlite.columns fee_policyInfo
-        Sqlite.finalize fee_policyInfo
-        case row of
-            [PersistText t] -> do
-                (a:b:_) <- pure (T.splitOn " + " t)
-                let newVal = a <> " + " <> b
+        let field = (DBField ProtocolParametersFeePolicy)
+            defaultValue = "0 + 0x"
```
hasufell: I don't see any other reasonable default value here.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
KtorZ: Actually, a migration is needed. And that migration is `addKeyDepositIfMissing`. Which is a bit unfortunate, because now, if we compose the two actions above (`addKeyDepositIfMissing` and `updateFeeValue`), we end up in a situation where the column is created with a possibly wrong value...

I'd suggest adding the `key_deposit` column here, instead of `addKeyDepositIfMissing`.

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
```diff
@@ -272,6 +274,10 @@ spec = do
             testMigrationAccountingStyle @ShelleyKey
                 "shelleyAccountingStyle-v2020-10-13.sqlite"
 
+        it "'migrate' db with new FeePolicy" $
+            testMigrationUpdateFeeValue @ShelleyKey
+                "shelleyAccountingStyle-v2020-10-13.sqlite"
```
KtorZ: Lazy or smart. I can't tell :grimacing:

KtorZ (approved):

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
hasufell: Haskellers know it can be both :grin:

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -639,17 +638,26 @@ migrateManually tr proxy defaultFieldValues =
...
+
         updateStakeKeyVal :: DBField -> Text -> IO ()
         updateStakeKeyVal stakeField c = do
             isFieldPresent conn stakeField >>= \case
                 TableMissing ->
                     -- this invariant must hold due to the parent context
                     fail "ProtocolParameters table missing"
                 ColumnMissing -> do
-                    traceWith tr $ MsgManualMigrationNotNeeded stakeField
+                    addColumn_ conn True stakeField c
```
hasufell: Here's the sweet juice :candy:

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1447,8 +1449,10 @@ mkApiTransactionFromInfo
     -> TransactionInfo
     -> m (ApiTransaction n)
 mkApiTransactionFromInfo ti (TransactionInfo txid ins outs ws meta depth txtime txmeta) = do
-    apiTx <- liftIO $ mkApiTransaction ti txid (drop2nd <$> ins) outs ws (meta, txtime) txmeta $
-        case meta ^. #status of
+    let deposits = []
+    apiTx <- liftIO $ mkApiTransaction ti txid (drop2nd <$> ins) outs ws (meta, txtime) txmeta
+        deposits
+        $ case meta ^. #status of
```
KtorZ: This is no more readable than before. It's actually worse. The 80-column max-length rule should apply, but not be detrimental to readability.  If it goes slightly beyond 80 chars, we can live with it.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -569,6 +571,102 @@ migrateManually tr proxy defaultFieldValues =
...
+        -- table if it is missing.
+        --
+        addKeyDepositIfMissing :: IO ()
+        addKeyDepositIfMissing = do
+            addColumn_ conn True (DBField ProtocolParametersKeyDeposit) value
+          where
+            value = toText $ defaultKeyDeposit defaultFieldValues
+
+        updateStakeKeyVal :: DBField -> Text -> IO ()
+        updateStakeKeyVal stakeField c = do
```
KtorZ: What about using a single function to cover all cases? Especially because they have quite different names although are acting on exactly the same table and column. `updateStakeKeyVal` should be enough in other cases too.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1328,26 +1427,29 @@ mkProtocolParametersEntity
     -> W.ProtocolParameters
     -> ProtocolParameters
 mkProtocolParametersEntity wid pp =
-    ProtocolParameters wid fp (getQuantity mx) dl desiredPoolNum minUTxO epochNo
+    ProtocolParameters wid fp (getQuantity mx) dl desiredPoolNum minUTxO
+        keyDep epochNo
```
KtorZ: This is no more readable than before.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
hasufell: I tried changing

```hs
    addColumn
        :: Sqlite.Connection
        -> Bool
        -> DBField
        -> Text
        -> Maybe (IO ()) -- ^ action to run when column is present
        -> IO SqlColumnStatus
    addColumn conn notNull field value updateCB = do
        isFieldPresent conn field >>= \st -> st <$ case st of
            TableMissing ->
                traceWith tr $ MsgManualMigrationNotNeeded field
            ColumnMissing -> do
                traceWith tr $ MsgManualMigrationNeeded field value
                query <- Sqlite.prepare conn $ T.unwords
                    [ "ALTER TABLE", tableName field
                    , "ADD COLUMN", fieldName field
                    , fieldType field, if notNull then "NOT NULL" else ""
                    , "DEFAULT", value
                    , ";"
                    ]
                _ <- Sqlite.step query
                Sqlite.finalize query
            ColumnPresent ->
                maybe (traceWith tr $ MsgManualMigrationNotNeeded field) id updateCB
```

but I think it makes it more convoluted. Although some code is shared here, it doesn't improve understanding imo.

hasufell: I improved the names a bit.

piotr-iohk:
@hasufell I think I may have encountered an issue:

 - doing first delegation - deposit is there:
```
curl -X PUT http://localhost:8090/v2/stake-pools/pool1ear0mr44np6zhkyf6uhcqut422jgpysn08mtsydjsf027fmqdvd/wallets/f0d444c7905fc7fa60ea19ff8867e024a14bf2c7 \
-d '{"passphrase":"Secure Passphrase"}' \
-H "Content-Type: application/json" | jq .deposits

...

 [
    {
      "quantity": 2000000,
      "unit": "lovelace"
    }
  ],
```

- but querying this transaction - deposit is empty:

```
curl -X GET http://localhost:8090/v2/wallets/f0d444c7905fc7fa60ea19ff8867e024a14bf2c7/transactions/b5b157eb0ef6600019a650cdb2dada07de9261d7a2ecd5057ed589adfa17eee1 | jq .deposits
...

[]
```

KtorZ:
:facepalm: good point @piotr-iohk, the deposit amount is never added to the tx meta database. We missed that very hard.

KtorZ (changes requested):
we must keep track of deposit as part of the transaction metadata so that the deposit can be reported correctly even once we are no longer aware of the initial coin selection.

piotr-iohk:
>  good point @piotr-iohk, the deposit amount is never added to the tx meta database. We missed that very hard.

I think integration test will be handy for that case.

hasufell:
@KtorZ I looked at the code and afais the deposit is:

1. never sent to the node
2. never makes it into the database

Instead, it's only ever directly used when calculating the fee. I'm suspecting the node/ledger keeps track of the deposit simply by looking at our delegation certificates, because that's currently the only way to have deposit/reclaim. If that's the case, I believe we don't need the deposit in the database, but can retrieve it based on the certificate and protocol parameters.

But I'm not sure... it also begs the question what happens if the protocol parameter is increased.

lib/core/src/Cardano/Wallet.hs
hasufell: I think we should try again to move it out.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1899,6 +1899,7 @@ mkTxMeta ti' blockHeader wState tx cs expiry =
                 , blockHeight = blockHeader ^. #blockHeight
                 , amount = Quantity $ distance amtInps amtOuts
                 , expiry = Just expiry
+                , deposit = Just . Quantity . Coin $ (cs ^. #deposit)
```
hasufell: This should do the trick... however... it doesn't. It still doesn't end up in the database. I couldn't figure out until now why. I followed the control flow and it should be hitting the database.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -215,11 +215,24 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
...
             (Link.listStakePools arbitraryStake) Empty
-        joinStakePool @n ctx pool (src, fixturePassphrase) >>= flip verify
+        rJoin <- joinStakePool @n ctx pool (src, fixturePassphrase)
+        verify rJoin
             [ expectResponseCode HTTP.status202
             , expectField (#status . #getApiT) (`shouldBe` Pending)
             , expectField (#direction . #getApiT) (`shouldBe` Outgoing)
+            , expectField #deposits (`shouldBe` [Quantity 1000000])
             ]
+        eventually "Wallet has joined pool and deposit info persists" $ do
```
hasufell, line 225: Here's the new test

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -215,11 +215,24 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
...
+            rJoin' <- request @(ApiTransaction n) ctx
+                (Link.getTransaction @'Shelley src
+                    (getFromResponse Prelude.id rJoin))
+                Default Empty
+            verify rJoin'
+                [ expectResponseCode HTTP.status200
+                , expectField (#status . #getApiT) (`shouldBe` InLedger)
+                , expectField (#direction . #getApiT) (`shouldBe` Outgoing)
+                , expectField #deposits (`shouldBe` [Quantity 1000000])
```
KtorZ: Please avoid the hard-coded value and use `depositAmt`

lib/core-integration/src/Test/Integration/Scenario/CLI/Shelley/Transactions.hs
```diff
@@ -125,7 +125,7 @@ spec = describe "SHELLEY_CLI_TRANSACTIONS" $ do
         let amt = fromIntegral minUTxOValue
         args <- postTxArgs ctx wSrc wDest amt Nothing Nothing
         Stdout feeOut <- postTransactionFeeViaCLI ctx args
-        ApiFee (Quantity feeMin) (Quantity feeMax) <- expectValidJSON Proxy feeOut
+        ApiFee (Quantity feeMin) (Quantity feeMax) [Quantity 0] <- expectValidJSON Proxy feeOut
```
KtorZ: That seems wrong. If there's no deposit, then there's no deposit.

lib/core/src/Cardano/Wallet.hs
KtorZ: This should be `Nothing` if `cs ^. #deposit == 0`

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
```diff
@@ -104,6 +104,7 @@ TxMeta
     txMetaAmount            Natural             sql=amount
     txMetaData              W.TxMetadata Maybe  sql=data
     txMetaSlotExpires       SlotNo Maybe        sql=slot_expires
+    txMetaDeposit           W.Coin Maybe        sql=deposit
```
KtorZ: For consistency and to avoid ambiguity, can we make it `sql=key_deposit`

lib/core/src/Cardano/Wallet.hs
hasufell: https://github.com/input-output-hk/cardano-wallet/blob/d4187a713364dca2f00a073338cdc65d37693d36/lib/core/src/Cardano/Wallet/Api/Server.hs#L1495-L1553
This is where the `txMeta` gets updated and the handler indeed returns a deposit. At the same time the handler also transmits this txMeta to the database in https://github.com/input-output-hk/cardano-wallet/blob/d4187a713364dca2f00a073338cdc65d37693d36/lib/core/src/Cardano/Wallet.hs#L1924-L1941

But it never shows up in the database.

KtorZ: What do you mean "it never shows up" ?

hasufell: > What do you mean "it never shows up" ?

The database field is `null`.

And I found the culprit: https://github.com/input-output-hk/cardano-wallet/blob/d4187a713364dca2f00a073338cdc65d37693d36/lib/core/src/Cardano/Wallet/Primitive/Model.hs#L383-L392

I don't see how to easily access `TransactionInfo` in that function without database access.

KtorZ: The model needs to be able to reconstruct the deposit's amount from the certificates present in the transaction and the active protocol parameters. 

The main issue here is to know what is the right value of the deposit _at the time the transaction was inserted_. We should be able to land back on our feet by checking the delta between inputs and outputs and the explicit transaction fees. With these, it becomes a simple equation with a single unknown.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -671,106 +678,81 @@ migrateManually tr proxy defaultFieldValues =
...
-                        -- update key deposit
-                        case rest of
-                            [c] -> updateKeyDeposit c
-                            _ -> pure ()
-
-                    [] -> addKeyDepositIfMissing
-                    xs -> fail ("Unexpected row result when querying fee value: " <> show xs)
+                case filter (/= PersistNull) row of
+                    [PersistText t] -> case T.splitOn " + " t of
+                        [a,b,c] -> do
```
hasufell, line 858: This is different than before and less permissive.

hasufell:
Wrt the new approach... we won't be able to distinguish different kinds of deposits I guess.

KtorZ:
> @hasufell Wrt the new approach... we won't be able to distinguish different kinds of deposits I guess 

Indeed, this won't be possible without also storing and tracking delegation certificates as part of our internal transaction model which is a bit too much of yake-shaving for this particular story. It'd have been nice to have, yet the original goal really is to be able to differentiate fees from deposit;

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
```diff
@@ -296,6 +305,65 @@ spec = parallel $ do
                     }
                 )
 
+        it "'migrate' db with new FeePolicy" $
+            testMigrationUpdateFeeValue @ShelleyKey
+                "shelleyRole-v2020-10-13.sqlite"
+                (LinearFee (Quantity 155381.0)  (Quantity 44.0))
+                (Coin 2000000)
+
+        it "'migrate' db to add fees to transactions" $
```
KtorZ, line 321: @piotr-iohk do you see any other interesting transaction from this database maybe that would be worth controlling?

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
```diff
@@ -313,6 +381,62 @@ sqliteSpecRnd = do
...
+                Incoming -> txInfoMeta ^. #fee `shouldSatisfy` isNothing
+                Outgoing -> txInfoMeta ^. #fee `shouldSatisfy` isJust
+
+        -- Also verify a few hand-picked transactions
+        forM_ caseByCase $ \(txid, expectedFee) -> do
+            case L.find ((== txid) . txInfoId) result of
+                Nothing ->
+                    fail $ "tx not found: " <> T.unpack (toText txid)
+                Just TransactionInfo{txInfoMeta} ->
+                    (txInfoMeta ^. #fee) `shouldBe` Just (Quantity expectedFee)
```
KtorZ: @piotr-iohk anything else you could think of that the migrated table should satisfy. In the end, the change is only operated on the `fee` column, so there isn't much more we can test for this particular column I believe...

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
piotr-iohk: Maybe two more with a bit more inputs / outputs:

`99907bf6ac73f6fe6fe25bd6b68bae6776425b9d15a7c46c7a49b85b8b03f291`:

10 inputs
10 outputs
withdrawal = 565251

`15940a7c1df8696279282046ebdb1ee890d4e9ac3c5d7213f360921648b36666`:

1 input
33 outputs
withdrawal = 561120

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
piotr-iohk: Perhaps validate also the calculated deposit? It will be 0 for all cases but one in current set of cases.

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
KtorZ: > Maybe two more with a bit more inputs / outputs:

Good idea!

KtorZ (approved):
Ok. @jonathanknowles it'd be good if you can do a quick sanity check before merging. Testing is pretty good on these areas and already caught many things (even a 1-year-old bug!) but another quick look won't hurt.

KtorZ:
bors try

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2035,6 +2049,61 @@ mkApiTransaction ti txid ins outs ws (meta, timestamp) txMeta setTimeReference =
...
 
+    depositIfAny :: Natural
+    depositIfAny
+        -- NOTE: totalIn will be zero for incoming transactions where inputs are
+        -- unknown, in which case, we also give no visibility on the deposit.
+        --
+        -- For outgoing transactions however, the total in is garanteed to be
+        -- greater or equal to totalOut; any reminder is actually a
+        -- deposit. Said differently, if totalIn > 0, then necessarily 'fee' on
+        -- metadata should be 'Just{}'
```
hasufell:
```suggestion
        -- For outgoing transactions however, the totalIn is guaranteed to be
        -- greater or equal to totalOut; any remainder is actually a
        -- deposit. Said differently, if totalIn > 0, then necessarily 'fee' on
        -- metadata should be 'Just{}'.
```

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -696,9 +709,171 @@ migrateManually tr proxy defaultFieldValues =
...
+    --
+    -- For outgoing transaction, it is possible to recalculate fees by
+    -- calculating the delta between the total input value minus the total
+    -- output value. The delta (inputs - output) is necessarily positive
+    -- (by definition of 'outgoing' transactions) and comprised of:
+    --
+    -- - Fees
+    -- - Total deposits if any
+    --
+    -- To substract deposit values from fees, we consider that any transaction
```
hasufell, line 741: Is this a reasonable assumption? I don't have an intuition for this, but I also don't see a hard justification for it.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -696,9 +709,171 @@ migrateManually tr proxy defaultFieldValues =
...
+    -- that has one or less output and fees greater than the key deposit (or min
+    -- utxo value) is a key registration transaction and the key deposit value
+    -- can be substracted from the delta to deduce the fees.
+    --
+    -- Note that ideally, we would do this in a single `UPDATE ... FROM` query
+    -- but the `FROM` syntax is only supported in SQLite >= 3.33 which is only
+    -- supported in the latest version of persistent-sqlite (2.11.0.0). So
```
hasufell, line 748: There are many reasons to abandon our fork of persistent-sqlite. This is another.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -696,9 +709,171 @@ migrateManually tr proxy defaultFieldValues =
     addScriptAddressGapIfMissing :: Sqlite.Connection -> IO ()
     addScriptAddressGapIfMissing conn =
         addColumn_ conn True (DBField SeqStateScriptGap) value
-      where
+     where
         value = T.pack $ show $ Seq.getAddressPoolGap Seq.defaultAddressPoolGap
 
+    -- This migration is rather delicate. Indeed, we need to introduce an
+    -- explicit 'fee' on known transactions, so only do we need to add the new
```
hasufell, line 723:
```suggestion
    -- explicit 'fee' on known transactions, so not only do we need to add the new
```

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -696,9 +709,171 @@ migrateManually tr proxy defaultFieldValues =
...
+    -- column (easy), but we also need to find the right value for that new
+    -- column (delicate).
+    --
+    -- Note that it is not possible to recover explicit fees on incoming
+    -- transactions without having access to the entire ledger (we do not know
+    -- the _amount_ from inputs of incoming transactions). Therefore, by
+    -- convention it has been decided that incoming transactions will have fee
+    -- equals to 0.
```
hasufell, line 731: Do we know if that could cause issues further down the line? Is it possible to use another magic value (such as `-1`) or make it nullable?

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -696,9 +709,171 @@ migrateManually tr proxy defaultFieldValues =
...
+    --
+    -- To substract deposit values from fees, we consider that any transaction
+    -- that has one or less output and fees greater than the key deposit (or min
+    -- utxo value) is a key registration transaction and the key deposit value
+    -- can be substracted from the delta to deduce the fees.
+    --
+    -- Note that ideally, we would do this in a single `UPDATE ... FROM` query
+    -- but the `FROM` syntax is only supported in SQLite >= 3.33 which is only
+    -- supported in the latest version of persistent-sqlite (2.11.0.0). So
+    -- instead, we queries all transactions which require an update in memory,
```
hasufell:
```suggestion
    -- instead, we query all transactions which require an update in memory,
```

hasufell (commented):
Generally, I'd prefer if we went the deposit route, but I cannot really judge how much work that would be. We definitely need to open a ticket for this though, since it's basically technical debt.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
KtorZ: No no. `0` is the value we want here for it conveys the right intuition: if I didn't make that transaction, from my perspective fees are 0. Since wallets are already user-centric (which is the reason why we can't know the amount any arbitrary inputs, but only those belonging to the wallet), it's a reasonable justification.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
KtorZ: Or to get away from persistent altogether :)

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
KtorZ: It is, on mainnet, testnet and staging networks at least. Surely, if one were to set very very small fee, it wouldn't be but such network doesn't exist yet and if it were to be created it wouldn't require any database migration in the first place. 

The rationale for this justification is that we know that key registration transaction coming from the wallet have by construction one single output and only need to cover fees. So they typically have a small number of inputs as well. A single output transaction is about 0.17 Ada, 0.20 Ada if you're unlucky and need several inputs. An input is ~40 bytes, so 1800 Lovelace on public networks. So you would need more than 1100 inputs for this assumption to become false (with 1100 inputs one could create a transaction that is more expensive than the key deposit, while still not be a key registration transaction). cardano-wallet would certainly not have constructed such transaction.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
hasufell: True. I think ORMs don't have much value, unless you assume inexperienced developers. Serialisation can be easily tested and optimizing for performance becomes less of a hassle. But that would probably be a massive patch.

KtorZ: > be a massive patch.

Not _much_ in the end. We used an indirection from the start, so it is a matter of re-implementing the [DBLayer](https://github.com/input-output-hk/cardano-wallet/blob/c955ca65309e1f74578e517792235816447ad19b/lib/core/src/Cardano/Wallet/DB.hs#L119) in terms of _something else_. We already have two implementations of it (SQLite with persistent & MVar)

KtorZ:
bors try

KtorZ:
bors merge

KtorZ:
bors retry

rvl:
With the unliftio PR merged, the `catcn` function no longer catches async exceptions.
That could have been the cause of the first bors failure.
Sorry for the churn.

So pushed a fix for `catch` + bonus changes.

The second bors failure looks like it could be general test flakiness, but I don't think we've seen those errors before.

bors try

rvl:
bors try

KtorZ:
> With the unliftio PR merged, the catcn function no longer catches async exceptions.

:facepalm: My bad. I rebased and did the change and pull off catch perfectly knowing I needed async catch. I need to get used to the "safe-exception" behavior, but I guess it's for the best. 

Thanks @rvl

lib/core/src/Cardano/Wallet/Registry.hs
```diff
@@ -195,10 +193,9 @@ defaultWorkerAfter
     -> IO ()
 defaultWorkerAfter tr = traceWith tr . \case
     Right _ -> MsgFinished
-    Left e -> case asyncExceptionFromException e of
-        Just ThreadKilled -> MsgThreadKilled
-        Just UserInterrupt -> MsgUserInterrupt
-        _ -> MsgUnhandledException $ pretty $ show e
+    Left e -> if isSyncException e
+        then MsgThreadCancelled
+        else MsgUnhandledException $ pretty $ show e
```
KtorZ: :+1: ok

KtorZ: Ah, no. Not okay. The `then` and `else` clauses are mixed up.

KtorZ: Fixed in: https://github.com/input-output-hk/cardano-wallet/compare/fe58426a3cd9609800c4eac9b5662f74bdb2a4dd..db3dc33736b7dd6424bb2afa32a88def19bacd20

KtorZ:
bors r+

Happy new year :tada:

hasufell:
May bors merge this before 2021 😂

KtorZ:
I am making an offering. It's been 10 years, but it's time to sacrifice the cat now.

PR #2329, opened 2020-11-18
Move selected primitive types to dedicated modules (Part 1).

jonathanknowles:
# Issue Number

Preparation for [ADP-347](https://jira.iohk.io/browse/ADP-347) (_Wallet should support UTxOs containing multiple token types_).

# Overview

This PR moves the following types (and related functions) to separate modules under the `Cardano.Wallet.Primitive.Types` hierarchy:

| Type | Destination Module |
| -- | -- |
| `Address` | `Cardano.Wallet.Primitive.Types.Address` |
| `ChimericAccount` | `Cardano.Wallet.Primitive.Types.ChimericAccount` |
| `Coin` | `Cardano.Wallet.Primitive.Types.Coin` |
| `Direction` | `Cardano.Wallet.Primitive.Types.Tx` |
| `Tx` | `Cardano.Wallet.Primitive.Types.Tx` |
| `Tx{Change,In,Meta,Metadata,Out,Status}` | `Cardano.Wallet.Primitive.Types.Tx` |
| `{Sealed,Unsigned}Tx` | `Cardano.Wallet.Primitive.Types.Tx` |

# Justification

This PR:
* makes it possible for code to use the above types without having to incur a dependency on the `Cardano.Wallet.Primitive.Types` module, which (together with the modules that depend on it) is already very large and slow to recompile.
* makes it possible to compile the above types in parallel, saving overall compilation time.

# Comments

This PR doesn't re-export the moved types from `Primitive.Types`.

Although it would be convenient to be able to import everything from `Primitive.Types` in one go (with a qualified import), it would also create a compilation bottleneck, since anything that imports one of the above types would have to depend on everything required to build `Primitive.Types`.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/Types/Address.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types/Address.hs
@@ -1,0 +1,126 @@
+{-# LANGUAGE DeriveGeneric #-}
+
+-- |
+-- Copyright: © 2018-2020 IOHK
+-- License: Apache-2.0
+--
+-- This module provides the 'Address' data type, which provides a human-friendly
+-- representation of a public key.
+--
+module Cardano.Wallet.Primitive.Types.Address
+    ( Address (..)
+    , AddressState (..)
+    ) where
+
+import Prelude
+
+import Control.DeepSeq
+    ( NFData (..) )
+import Data.Bifunctor
+    ( bimap )
+import Data.ByteArray.Encoding
+    ( Base (Base16), convertFromBase, convertToBase )
+import Data.ByteString
+    ( ByteString )
+import Data.Text.Class
+    ( CaseStyle (..)
+    , FromText (..)
+    , TextDecodingError (..)
+    , ToText (..)
+    , fromTextToBoundedEnum
+    , toTextFromBoundedEnum
+    )
+import Fmt
+    ( Buildable (..), prefixF, suffixF )
+import GHC.Generics
+    ( Generic )
+
+import qualified Data.Text.Encoding as T
+
+-- | Representation of Cardano addresses.
+--
+-- Addresses are basically a human-friendly representation of public keys.
+-- Historically in Cardano, there exist different sort of addresses, and new
+-- ones are to come. So far, we can distinguish between three types of
+-- address:
+--
+-- - Byron Random addresses, which holds a payload with derivation path details
+-- - Byron Sequential addresses, also known as Icarus'style addresses
+-- - Shelley base addresses, see also [implementation-decisions/address](https://github.com/input-output-hk/implementation-decisions/blob/master/text/0001-address.md)
+--
+-- For more details, see also [About Address Derivation](https://github.com/input-output-hk/cardano-wallet/wiki/About-Address-Derivation)
+--
+-- Shelley base addresses can be divided into two types:
+--
+-- - Single Addresses: which only hold a public spending key
+-- - Group Addresses: which hold both a spending and delegation keys
+--
+-- It'll therefore seem legitimate to represent addresses as:
+--
+-- @
+-- data Address
+--   = ByronAddress !ByteString
+--   | SingleAddress !XPub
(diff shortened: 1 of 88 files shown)


paweljakubas (approved):
not a small refactoring;-) But the funneling adopted Types -> (Tx,Coin,Address, ChimericAccount) seems like a nice move that will help in lowering compilation time. Maybe I would put `ChimericAccount` in `Address` and rename it to RewardAccount as we don't have Jormungandr in cardano-wallet - what do you think? BTW. Do we know how much we have gained due to those changes?

lib/core/src/Cardano/Wallet/Primitive/Types/Address.hs
```diff
@@ -0,0 +1,126 @@
+{-# LANGUAGE DeriveGeneric #-}
+
+-- |
+-- Copyright: © 2018-2020 IOHK
+-- License: Apache-2.0
+--
+-- This module provides the 'Address' data type, which provides a human-friendly
```
Anviking, line 7: How about `This module provides the main 'Address' data type used by the wallet`, or something to more show the relationship this module has with the rest of the code-base?

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
```diff
@@ -0,0 +1,340 @@
...
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE OverloadedLabels #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Cardano.Wallet.Primitive.Types.Tx
```
Anviking, line 10: You have doc comments for the other new modules, but not in this one

jonathanknowles:
> Maybe I would put ChimericAccount in Address and rename it to RewardAccount as we don't have Jormungandr in cardano-wallet - what do you think?

Just for the record: I agree with this suggestion. :+1: 

@paweljakubas and I agreed that since the surface area of this PR is rather large, that I will make another PR to do the renaming from `ChimericAccount` to `RewardAccount`. 

(It's important to merge this PR sooner rather than later, to avoid merge conflicts with other PRs.)

Anviking (approved):
(minor comments, that maybe could be considered in follow-up pr)

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
jonathanknowles: > You have doc comments for the other new modules, but not in this one

Great point @Anviking. I'll add this in a follow-up PR.

lib/core/src/Cardano/Wallet/Primitive/Types/Address.hs
jonathanknowles: > How about This module provides the main 'Address' data type used by the wallet, or something to more show the relationship this module has with the rest of the code-base?

Very good point! I'll add this in a follow-up PR, along with the other suggestions.

jonathanknowles:
@paweljakubas wrote:
> Do we know how much we have gained due to those changes?

It's likely that this PR on its own will not reduce the overall compilation time by much, if at all. To obtain significant reductions in compilation time, we'd need to:

- break up the remainder of `Primitive.Types`.
- break up the test case modules that depend on these types.
- break up other very large modules.

However, this PR is still an improvement for developers, as when developing interactively in `ghci` it's now possible to have a dependency on types like `TxOut` without having to depend on the entirety of `Primitive.Types`. So reloading changes should be much quicker in certain circumstances.

jonathanknowles:
bors retry

jonathanknowles:
bors retry

jonathanknowles:
bors retry

PR #2335, opened 2020-11-19
Move selected primitive types to dedicated modules (Part 2).

jonathanknowles:
# Issue Number

Preparation for [ADP-347](https://jira.iohk.io/browse/ADP-347) (_Wallet should support UTxOs containing multiple token types_).

# Overview

This PR cleans up a few loose ends left behind by PR https://github.com/input-output-hk/cardano-wallet/pull/2329:

- Renames `ChimericAccount` to `RewardAccount`.
  (see [related discussion](https://github.com/input-output-hk/cardano-wallet/pull/2329#pullrequestreview-533297295))
- Adds a module comment to `Primitive.Types.Tx`.
  (see [related discussion](https://github.com/input-output-hk/cardano-wallet/pull/2329#discussion_r525975667)) 
- Clarifies the module comment for `Primitive.Types.Address`.
  (see [related discussion](https://github.com/input-output-hk/cardano-wallet/pull/2329#discussion_r525973937))

It also moves the following UTxO-related types (and functions) to a new module:

| Type | Destination Module |
| -- | -- |
| `UTxO` | `Cardano.Wallet.Primitive.Types.UTxO` |
| `UTxOStatistics` | `Cardano.Wallet.Primitive.Types.UTxO` |
| `Dom` | `Cardano.Wallet.Primitive.Types.UTxO` |
| `HistogramBar` | `Cardano.Wallet.Primitive.Types.UTxO` |

# Motivation

The primary motivation for making these changes now, rather than later on, is to make development of multi-asset work more efficient.

So this PR limits itself to types that are likely to be affected by the MA work, rather than breaking up the entirety of `Primitive.Types`.

# Compilation Time Reductions

Here's a comparison of how long it takes to re-compile the whole of `cardano-wallet-core` (including tests) after making a tiny change to the `UTxO` type:

Before applying PR #2329 and PR #2335:
```
0m46.404s
0m45.124s
0m44.854s
```
After applying PR #2329 and PR #2335:
```
0m28.524s
0m28.364s
0m28.324s
```
That's approximately a 37% reduction in wall-clock time, with only slight changes to our module hierarchy.

**Machine used:**
```
vendor_id       : GenuineIntel
cpu family      : 6
model           : 85
model name      : Intel(R) Xeon(R) CPU
stepping        : 7
microcode       : 0x1
cpu MHz         : 2800.174
cache size      : 33792 KB
```

**Command:**
```
time stack build --ghc-options "-j +RTS -A256m -n4m -RTS" --fast --test --no-run-tests cardano-wallet-core
```

# Analysis

This PR:
* makes it possible for code to use the above types without having to incur a dependency on the `Cardano.Wallet.Primitive.Types` module, which (together with the modules that depend on it) is currently still large and quite slow to recompile.
* makes it possible to compile the above types in parallel, saving overall compilation time.
* makes module caching more effective: changes to one of the sub-modules will not necessarily invalidate the cache for the others, meaning that the compiler can often get away with compiling less code when changes are made.

# Comments

As before, this PR does not re-export the moved types from `Primitive.Types`, thereby forcing users of these types to import them directly from the module in which they're defined. This reduces the extent to which `Primitive.Types` is a compilation bottleneck.


lib/core/src/Cardano/Wallet/Primitive/Types/RewardAccount.hs
```diff
@@ -0,0 +1,48 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE TypeApplications #-}
+
+-- |
+-- Copyright: © 2018-2020 IOHK
+-- License: Apache-2.0
+--
+-- This module provides the 'RewardAccount' data type, used for staking
```
rvl: `s/staking/delegation/`  Staking is a taboo word.

Can the description be more specific than "purposes"?

lib/core/src/Cardano/Wallet/Primitive/Types/RewardAccount.hs
```diff
@@ -0,0 +1,48 @@
...
+import Data.ByteString
+    ( ByteString )
+import Data.Text.Class
+    ( FromText (..), ToText (..) )
+import Fmt
+    ( Buildable (..) )
+import GHC.Generics
+    ( Generic )
+
+-- | Also known as a staking key, a reward account is used in group-type
```
rvl: I don't think "staking key" is correct - this term does not appear in the delegation design spec.

rvl (commented):
I would say the primary purpose of this PR is to reduce recompilation time by splitting the Types module, rather than much to do with ADP-347.

Which is a commendable effort, thanks.

It looks more or less fine, though it's always hard to tell with renaming+code move type diffs.

Please do wait for reviews from those who made the review comments in the other PR.

lib/core/src/Cardano/Wallet/Primitive/Types/RewardAccount.hs
jonathanknowles: Hi @rvl 

I've pushed a fix in c21aabab7e0ee2a318e056a5298800f6c99ca795.

I've removed the "purposes" bit.

Does this work for you? (Feel free to suggest alternative wordings.)

lib/core/src/Cardano/Wallet/Primitive/Types/RewardAccount.hs
jonathanknowles: > I don't think "staking key" is correct - this term does not appear in the delegation design spec.

Removed in c21aabab7e0ee2a318e056a5298800f6c99ca795.

jonathanknowles:
@rvl wrote:

> I would say the primary purpose of this PR is to reduce recompilation time by splitting the Types module, rather than much to do with [ADP-347](https://jira.iohk.io/browse/ADP-347).

I think that's certainly a fair criticism.

It is true that my primary motivation for doing this now (rather than later) was to make life easier when developing for [ADP-347](https://jira.iohk.io/browse/ADP-347), but I think you're right that this could have been explicitly categorized under "reducing compilation time".

I do remember there being a technical debt ticket in JIRA for exactly this purpose (splitting up large modules to improve compilation time), but I can't seem to find it. :man_shrugging:

rvl:
It's not a criticism really. If you want to make good improvements like this, and have the spare capacity to do them, I don't think it should need to be hidden under the cover of a separate jira ticket.

paweljakubas (approved):
looks fine to me

Anviking (approved):
👍

PR #2336, opened 2020-11-19
Fix cardano-address version info and improve version info script

rvl:
### Issue Number

ADP-552

### Overview

- Add a script which can be used for manually inspecting release package versions.
- Bump cardano-addresses to fix its cabal package version.
- Adjust the nix build so that it can pass a git revision down to cardano-addresses.

### Comments

This contains an unreleased version of cardano-addresses and so needs a cardano-addresses 3.1.1 release.


piotr-iohk (approved):
Not sure about the nix stuff but script looks/works fine. Thx. 
Would be nice maybe to have a Windows script also at some point.

rvl:
I tested the nix stuff like this:
```
$(nix-build -A cardano-address --no-out-link)/bin/cardano-address --version

these paths will be fetched (1.34 MiB download, 8.75 MiB unpacked):
  /nix/store/nqrzp3n53ry7v1ksf2hg7qdqpxcbbmgj-cardano-addresses-cli-exe-cardano-address-3.1.1
copying path '/nix/store/nqrzp3n53ry7v1ksf2hg7qdqpxcbbmgj-cardano-addresses-cli-exe-cardano-address-3.1.1' from 'https://hydra.iohk.io'...
3.1.1 @ 817896ccf65cbd088eeff963f859ff4f6e8a9825
```

We can do a script for windows.

scripts/program-revs.sh
```diff
@@ -0,0 +1,35 @@
...
+
+set -euo pipefail
+
+for prog in cardano-wallet cardano-node cardano-cli cardano-address cardano-tx bech32; do
+  echo
+  if type -p $prog > /dev/null; then
+    echo "Your $prog is `type -p $prog`"
+    if [ $prog = cardano-wallet ]; then
+      $prog version
+    elif [ $prog = cardano-tx -o $prog = bech32 ]; then
```
KtorZ, line 27: A version switch was recently introduced on `bech32`. Though unreleased yet.

KtorZ (commented):
The script looks fine, I assume that the cardano-addresses bump was for testing it but should be remove before merging?

scripts/program-revs.sh
rvl: Yep, thanks @newhoggy. it would be nice if everything had a `--version` switch.

rvl:
I was going to wait for a cardano-addresses release before merging this, but it didn't happen as soon as I expected.
I will just remove the version bump and merge.

Issue #2189, opened 2020-09-29
Windows test fail: roundtrip JSON serialization & deserialization


piotr-iohk:
# Context
 
https://github.com/input-output-hk/cardano-wallet/runs/1179455249?check_suite_focus=true
https://github.com/input-output-hk/cardano-wallet/runs/1174007888?check_suite_focus=true
https://github.com/input-output-hk/cardano-wallet/runs/1171293981?check_suite_focus=true
etc.

# Test Case

```
can perform roundtrip JSON serialization & deserialization, and match existing golden files, JSON encoding of ApiAddressTestnet0, produces the same JSON as is found in test\data\Cardano\Wallet\Api\ApiAddressTestnet0.json
```

# Failure / Counter-example

```
Failures:

  src/Test/Aeson/Internal/GoldenSpecs.hs:74:5: 
  1) Cardano.Wallet.Api.Types, can perform roundtrip JSON serialization & deserialization, and match existing golden files, JSON encoding of ApiAddressTestnet0, produces the same JSON as is found in test\data\Cardano\Wallet\Api\ApiAddressTestnet0.json
       uncaught exception: ErrorCall
       Error in $.samples[0].id: invalid address

  To rerun use: --match "/Cardano.Wallet.Api.Types/can perform roundtrip JSON serialization & deserialization, and match existing golden files/JSON encoding of ApiAddressTestnet0/produces the same JSON as is found in test\\data\\Cardano\\Wallet\\Api\\ApiAddressTestnet0.json/"
```

---

# Resolution 

- More information in resolution PR #2339

---

# QA 

1. Windows nightly tests should pass.

Anviking:
Don't think windows tests are passing, but I think this problem was fixed. Think we can close.

PR #2339, opened 2020-11-20
Fix nightly windows tests

rvl:
### Issue Number

#2189

### Overview

During release v2020-11-17 I noticed that the windows tests had been failing for a while.

The unit tests broke somewhere between 4271b663bba2dfd353366a668cd43181d678c1c3 and d4d5cf3f1dbf478e49186f9e1742939af528bcaa.

This PR reverts the commit which added the JSON golden test file which fails.

I don't know what it is about this file which causes the error.


piotr-iohk (approved):
ok

Issue #2292, opened 2020-11-04
Flaky test - various DB properties causing timeout


Anviking:
# Context
 
 

Timeouts due to DB properties started occurring at the same time, and have been pretty frequent since. I now merged similar tickets all into this one.

https://github.com/input-output-hk/cardano-wallet/pull/2225#issuecomment-721630391

# Test Case

Several:

https://github.com/input-output-hk/cardano-wallet/blob/970bebbf10a0a984477e14f48033757f6726ce5b/lib/core/test/unit/Cardano/Wallet/DB/Properties.hs#L608-L636

https://github.com/input-output-hk/cardano-wallet/blob/970bebbf10a0a984477e14f48033757f6726ce5b/lib/core/test/unit/Cardano/Wallet/DB/Properties.hs#L824-L880

```haskell
-- | Trying to remove a same wallet twice should yield an error
prop_removeWalletTwice
    :: DBLayer IO s JormungandrKey
    -> ( PrimaryKey WalletId
       , Wallet s
       , WalletMetadata
       )
    -> Property
prop_removeWalletTwice db@DBLayer{..} (key@(PrimaryKey wid), cp, meta) =
    monadicIO (setup >> prop)
  where
    setup = liftIO $ do
        cleanDB db
        atomically $ unsafeRunExceptT $ initializeWallet key cp meta mempty pp
    prop = liftIO $ do
        let err = ErrNoSuchWallet wid
        atomically (runExceptT $ removeWallet key) `shouldReturn` Right ()
        atomically (runExceptT $ removeWallet key) `shouldReturn` Left err
```

# Failure / Counter-example

E.g. getting stuck on
```
Cardano.Wallet.DB.Sqlite.Types
  Values can be persisted and unpersisted successfully
    can persist and unpersist values of type 'SlotNo'
      +++ OK, passed 100 tests.
Cardano.Wallet.DB.Sqlite
  Sqlite
    Extra Properties about DB initialization
      createWallet . listWallets yields expected results
        +++ OK, passed 100 tests.
      creating same wallet twice yields an error
        +++ OK, passed 100 tests.
      removing 
```

---

# Resolution 

---

# QA

Anviking:
I wondered if this could be related to the parallelisation in some intricate way (by integration tests taking up more resources in CI).

But it seems the first failures occurred 2020-11-04 10:52 (on 7f745e74f), whereas the parallel (in buildkite) PR was merged 18:09 
[...]

PR #2341, opened 2020-11-23
Fix for Sqlite DB property tests timeout

rvl:
### Issue Number

Relates to #2292

### Overview

The sqlite database tests sometimes get stuck indefinitely.
The problem is an `Arbitrary` generator, because the error still occurs with `noShrinking`, and there is never a test failure.

Running the test with a small maximum heap size and profiling shows that the arbitrary function for `GenTxHistory` is responsible for the majority of execution time and memory usage.

I suspect it's due to the use of ``scale (`mod` 100)``. The "limit" of 100 is probably too high. It would anyway probably be better to reduce the size of GenTxHistory using either `div` or a fixed maximum size.

There is also a commit that reduces the size of generated TxMetadata for sqlite tests, because it doesn't need to be large.

### Comments

- Still need to do some experimentation with scaling.

- Looking at the profile below GenTxHistory, there is no single smoking gun in terms of execution time, so I think that means that the size of the GenTxHistory map is just too large.

- I tried sampling `arbitrary :: Gen GenTxHistory` by itself, but found no problems, so now I'm sttuck again.
  ```
  λ quickCheck $ withMaxSuccess 10000 (\h -> length (unGenTxHistory h) < 100)
  +++ OK, passed 10000 tests.
  ```

### DIFF
--- a/lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs
+++ b/lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs
@@ -42,7 +42,7 @@ import Cardano.Wallet.DB.Model
 import Cardano.Wallet.DummyTarget.Primitive.Types as DummyTarget
     ( block0, dummyGenesisParameters, mkTx, mockHash )
 import Cardano.Wallet.Gen
-    ( genMnemonic, genTxMetadata, shrinkSlotNo, shrinkTxMetadata )
+    ( genMnemonic, genSmallTxMetadata, shrinkSlotNo, shrinkTxMetadata )
 import Cardano.Wallet.Primitive.AddressDerivation
     ( Depth (..)
     , DerivationType (..)
@@ -429,7 +429,7 @@ instance Arbitrary TxStatus where
     arbitrary = elements [Pending, InLedger]
 
 instance Arbitrary TxMetadata where
-    arbitrary = genTxMetadata
+    arbitrary = genSmallTxMetadata
     shrink = shrinkTxMetadata
 
 instance Arbitrary Coin where
--- a/lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs
+++ b/lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs
@@ -257,7 +257,7 @@ instance Arbitrary GenTxHistory where
         -- We discard pending transaction from any 'GenTxHistory since,
         -- inserting a pending transaction actually has an effect on the
         -- checkpoint's pending transactions of the same wallet.
-        filter (not . isPending . snd) <$> scale (`mod` 100) arbitrary
+        filter (not . isPending . snd) <$> scale (min 25) arbitrary
       where
         sortTxHistory = filterTxHistory Nothing Descending wholeRange
 
--- a/lib/core/test/unit/Cardano/Wallet/Gen.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Gen.hs
@@ -20,6 +20,7 @@ module Cardano.Wallet.Gen
     , shrinkSlotNo
     , genTxMetadata
     , shrinkTxMetadata
+    , genSmallTxMetadata
     ) where
 
 import Prelude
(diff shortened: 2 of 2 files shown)


lib/core/test/unit/Cardano/Wallet/Gen.hs
```diff
@@ -222,6 +223,18 @@ genTxMetadata = do
         Left e -> fail $ show e <> ": " <> show (Aeson.encode json)
         Right metadata -> pure metadata
 
+-- | Generates a 'TxMetadata' containing only simple values.
+genSmallTxMetadata :: Gen TxMetadata
+genSmallTxMetadata = TxMetadata <$>
+    (Map.singleton <$> arbitrary <*> genSimpleTxMetadataValue)
```
jonathanknowles, line 229: Hi @rvl 

Just curious: will having only one entry in the map affect the coverage in any way?

(In other words, do we have test conditions that only trigger in the presence of many-valued metadata maps?)

lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs
```diff
@@ -257,7 +257,7 @@ instance Arbitrary GenTxHistory where
         -- We discard pending transaction from any 'GenTxHistory since,
         -- inserting a pending transaction actually has an effect on the
         -- checkpoint's pending transactions of the same wallet.
-        filter (not . isPending . snd) <$> scale (`mod` 100) arbitrary
+        filter (not . isPending . snd) <$> scale (min 25) arbitrary
```
jonathanknowles, line 260: Just curious about the choice to use `min` here, rather than `mod`.

If the generator frequently generates value lengths that are greater than 25 (according to some distribution), wouldn't this mean that we end up choosing 25 for much of the time?

jonathanknowles (commented):
Hi Rodney.

Thanks for doing this work! Your solution seems like a good idea.

I have just a couple of questions. (See comments!)

lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs
rvl: Important to realise: The size parameter is not the length of the generated list - it's the upper bound on list length.

QuickCheck gradually increases the size parameter as it runs tests.

The `scale` function adjusts the size according to the given (non-)linear function.
Using `mod` caps the size parameter at 100, but it doesn't make sense to me.
Better to clamp the size with the `min` function.

There is also the `resize` function which fixes a constant size. But I would like the size to grow - up to a point.

lib/core/test/unit/Cardano/Wallet/Gen.hs
rvl: Yes, thanks for the thoughts...

I thought it would be unlikely because the metadata is stored in the data as a json string.
That serialization is unit tested elsewhere.
Using a complex TxMetadata generator just makes things slower and less predictable.
A simple TxMetadata generator can still generate arbitrarily long strings with arbitrarily funny characters.

jonathanknowles (approved):

Anviking:
bors try

paweljakubas:
I tried `$ stack test --ta '-m "machine"' cardano-wallet-core:unit` on master 10 times and it works ok (last week it was not the case). It takes from 90 to 150 seconds. So I do not see the problem it was week/two weeks ago

Anviking:
bors try

Anviking:
bors try-

Anviking:
bors try

Anviking (approved):
Looks promising! Let's merge?

When trying your command:
```haskell
λ> import Test.QuickCheck
λ> import Cardano.Wallet.Gen
λ> :set -XTypeApplications
λ> :set +s
λ> quickCheck $ withMaxSuccess 10000 (\h -> length (unGenTxHistory h) < 100)
```

master:
```
+++ OK, passed 10000 tests.
(46.75 secs, 106,178,835,368 bytes)
```

this branch:
```
+++ OK, passed 10000 tests.
(17.41 secs, 42,873,792,512 bytes)
```

when having `pure Nothing` for the Tx metadata generator:
```
+++ OK, passed 10000 tests.
(16.01 secs, 40,785,673,640 bytes)
```

PR #2343, opened 2020-11-24
Test for mainnet value of ActiveSlotCoeff parameter

rvl:
### Issue Number

#2226 / ADP-554

### Overview

- Updates ticket number in TODO for ActiveSlotCoeff test.
- Adds check that the shelley mainnet value is hardcoded.


piotr-iohk (approved):
ack

PR #2347, opened 2020-11-25
[WIP] Calculate the minimum ada value for MA UTxO entries

hasufell:
# Issue Number

https://jira.iohk.io/browse/ADP-506

# Overview

- [x] Updated dependencies, similar to https://github.com/input-output-hk/cardano-wallet/pull/2245
- [x] Added a small module `Cardano.Wallet.Shelley.MultiAsset.Compatibility`
- [ ] Adjust `dustThreshold` and `adjustForFee` (this requires `TxOut` to be migrated to `TokenBundle`)

# Comments

The [Value](https://github.com/input-output-hk/cardano-ledger-specs/blob/7280ee5e71dbda7a6bc3610594a39e8ad534624a/shelley-ma/impl/src/Cardano/Ledger/Mary/Value.hs#L103) type has an `era` type parameter. Our token bundle currently doesn't have that. Converting between the two now begs the question on what to do about that.


lib/core/src/Cardano/Wallet/Shelley/MultiAsset/Compatibility.hs
```diff
@@ -0,0 +1,88 @@
...
+    ( ScriptHash (..) )
+
+import qualified Data.Map.Strict as Map
+import qualified Shelley.Spec.Ledger.Coin as SL
+
+-- | Convert our own @TokenBundle@ type to the corresponding
+-- @Value era@ ledger type. This should be very low in cost,
+-- because it just converts strict Map to strict Map.
+toLedgerSpecValue
+    :: forall era. (HashAlgorithm (ADDRHASH (Crypto era)))
```
hasufell: So here we need the `era` and we don't know what it is.

lib/core/src/Cardano/Wallet/Shelley/MultiAsset/Compatibility.hs
```diff
@@ -0,0 +1,88 @@
...
+    -> Value era
+toLedgerSpecValue = \(TokenBundle ada (TokenMap tb)) ->
+    Value (fromIntegral $ unCoin ada) . Map.foldrWithKey' outer Map.empty $ tb
+  where
+    outer (TokenPolicyId h) a m = case hashFromBytes $ getHash h of
+        Just sbs -> Map.insert
+            (PolicyID (ScriptHash sbs))
+            (Map.foldrWithKey inner mempty (toMap a))
+            m
+        Nothing -> m -- TODO: do we want to hard-fail here?
```
hasufell: I'm not sure if we want to short-circuit here. I'm leaning towards `ExceptT`.

lib/core/src/Cardano/Wallet/Shelley/MultiAsset/Compatibility.hs
```diff
@@ -0,0 +1,88 @@
...
+    inner (TokenName k) (TokenQuantity a) m = Map.insert (AssetName $ encodeUtf8 k) a m
+
+-- | Calculate the minimal UTxO value for a token bundle.
+--
+-- Returns @Nothing@ if Coin conversion fails (Integer to Word64).
+calculateMinCoin
+    :: forall era . (Era era)
+    => Coin        -- ^ the protocol minUTxOValue
+    -> TokenBundle
+    -> Maybe Coin
```
hasufell: Ledger Coin is `Integer`. Our coin is `Word64`. This is an awkward situation.

lib/core/src/Cardano/Wallet/Shelley/MultiAsset/Compatibility.hs
```diff
@@ -0,0 +1,88 @@
...
+calculateMinCoin = \c t -> fromLedgerCoin $ scaledMinDeposit
+    (toLedgerSpecValue @era t) (fromWalletCoin c)
+  where
+    fromLedgerCoin :: SL.Coin -> Maybe Coin
+    fromLedgerCoin (SL.Coin i) = Coin <$> (toIntegralSized i)
```
hasufell: TODO: also use `isValidCoin`

jonathanknowles:
Closing for now, as this has been implemented in #2461.

PR #2348, opened 2020-11-25
rework transaction size estimation with grosser approximation

KtorZ:
# Issue Number

ADP-567 / ADP-569

# Overview

- [ ] I have reworked transaction size estimation with grosser approximation

# Comments

  The function for calculting the size of a transaction in Shelley had
  grown out of hands and became a function really hard to reason about
  and debug. We kept running into either performance issues or simply
  incoherent fee calculations. This commit takes a simpler approach to
  the problem which is very straightforward to follow, but comes at the
  cost of slightly grosser approximations (in the sense that the
  calculated size will tend to be bigger than the actual size). Yet,
  there are several parts we could iterate on might it be necessary.

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
@@ -52,7 +52,7 @@ import Data.Quantity
 import Data.Text
     ( Text )
 import Data.Typeable
-    ( Typeable, typeOf )
+    ( Typeable )
 import Data.Word
     ( Word64 )
 import Test.Hspec
@@ -264,11 +264,6 @@ spec = describe "BYRON_MIGRATIONS" $ do
         addrs <- listAddresses @n ctx wNew
         let addr1 = (addrs !! 1) ^. #id
 
-        numOfTxs <- case (show (typeOf (_target ctx))) of
-                s | s == "Proxy * Jormungandr" -> pure 1
-                s | s == "Proxy * Shelley" -> pure 20
-                _ -> fail ("unknown target backend? (" <> show (typeOf (_target ctx)) <> ")")
-
         let payloadMigrate =
                 Json [json|
                     { passphrase: #{fixturePassphrase}
@@ -280,7 +275,7 @@ spec = describe "BYRON_MIGRATIONS" $ do
             payloadMigrate
         verify rm
             [ expectResponseCode HTTP.status202
-            , expectField id ( (`shouldBe` (numOfTxs)) . length )
+            , expectField id ( (`shouldBe` 19) . length )
             ]
 
         -- Check that funds become available in the target wallet:
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
@@ -40,15 +40,11 @@ import Prelude
 import Cardano.Address.Derivation
     ( XPrv, toXPub )
 import Cardano.Api.Typed
-    ( NetworkId, TxExtraContent (..) )
+    ( NetworkId, SerialiseAsCBOR (..), TxExtraContent (..) )
 import Cardano.Binary
     ( serialize' )
-import Cardano.Crypto.DSIGN
-    ( DSIGNAlgorithm (..), SignedDSIGN (..) )
 import Cardano.Crypto.Wallet
     ( XPub )
-import Cardano.Ledger.Crypto
-    ( Crypto (..) )
 import Cardano.Wallet.Primitive.AddressDerivation
     ( Depth (..), Passphrase (..), RewardAccount (..), WalletKey (..) )
 import Cardano.Wallet.Primitive.AddressDerivation.Byron
@@ -73,7 +69,6 @@ import Cardano.Wallet.Primitive.Types.Tx
     ( SealedTx (..), Tx (..), TxIn (..), TxMetadata, TxOut (..) )
 import Cardano.Wallet.Shelley.Compatibility
     ( Shelley
-    , StandardCrypto
     , sealShelleyTx
     , toCardanoLovelace
     , toCardanoStakeCredential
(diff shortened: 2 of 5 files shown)


lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -132,20 +132,12 @@ spec = do
...
-    estimateMaxInputsTests @IcarusKey Cardano.Mainnet
-        [(1,19),(10,14),(20,9),(30,4)]
-    estimateMaxInputsTests @IcarusKey (Cardano.Testnet (Cardano.NetworkMagic 0))
-        [(1,18),(10,13),(20,8),(30,2)]
+    estimateMaxInputsTests @ShelleyKey
+        [(1,27),(10,19),(20,10),(30,1)]
+    estimateMaxInputsTests @ByronKey
+        [(1,17),(10,11),(20,4),(30,0)]
+    estimateMaxInputsTests @IcarusKey
+        [(1,17),(10,11),(20,4),(30,0)]
```
KtorZ, line 140: NOTE 1: IcarusKey and ByronKey are now strictly equal since the estimation function doesn't make a difference between both. 

NOTE 2: Similarly, there's no longer a variation between mainnet and testnet (this only affected Byron addresses, and this new approach just consider the worst case scenario for Byron addresses all the time).

lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
```diff
@@ -280,7 +275,7 @@ spec = describe "BYRON_MIGRATIONS" $ do
             payloadMigrate
         verify rm
             [ expectResponseCode HTTP.status202
-            , expectField id ( (`shouldBe` (numOfTxs)) . length )
+            , expectField id ( (`shouldBe` 19) . length )
```
paweljakubas, line 278: why not `20` ? :thinking:

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -364,182 +350,322 @@ _decodeSignedTx bytes = do
...
+        = sizeOf_SmallArray
+        + sizeOf_SmallUInt
+        + sizeOf_Hash28
+
+    -- We carry addresses already serialized, so it's a matter of measuring.
+    sizeOf_Address addr
+        = 2 + toInteger (BS.length (unAddress addr))
+
+    -- For change address, we consider the worst-case scenario based on the
+    -- given wallet scheme. Byron addresses are larger.
```
paweljakubas, line 552: :+1:

lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
KtorZ: Because this PR changes the fee calculation a bit and it no longer generate 20 transactions but 19 :shrug:

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -364,182 +350,322 @@ _decodeSignedTx bytes = do
...
+        = case witTag of
+            TxWitnessByronUTxO{} -> 85
+            TxWitnessShelleyUTxO -> 59
+
+    -- Coins can really vary so it's very punishing to always assign them the
+    -- upper bound. They will typically be between 3 and 9 bytes (only 6 bytes
+    -- difference, but on 20+ outputs, one starts feeling it).
+    --
+    -- So, for outputs, since we have the values, we can compute it accurately.
+    sizeOf_Coin
```
paweljakubas, line 567: :+1:

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -364,182 +350,322 @@ _decodeSignedTx bytes = do
...
+        + sizeOf_BootstrapWitnesses
+      where
+        -- ?0 => [* vkeywitness ]
+        sizeOf_VKeyWitnesses
+            = (if numberOf_VkeyWitnesses > 0 then sizeOf_Array + sizeOf_SmallUInt else 0)
+            + sizeOf_VKeyWitness * numberOf_VkeyWitnesses
+
+        -- ?1 => [* multisig_script ]
+        sizeOf_MultisigScript
+            = 0
```
paweljakubas, line 594: I bet this one is to be updated very soon

KtorZ: Yes. I was thinking about adding a `TODO` tag with a ticket number maybe ?

paweljakubas (approved):
now all the ingredients decomposition is very clear. cddl is source of truth. Would be good at some in the future to have automatic translation of cddl from ledger-spec into size (dreaming loudly). Moreover, I do think this kind of function should be on cardano-ledger-spec side, and used in node, even exposed by it, and consumed by us.

piotr-iohk (approved):
Doing a manual regression on transactions and fees all works OK, including large multi-address transactions. It "feels" that fee estimation is now faster also. I have started benchmarks on the branch (https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/743)

KtorZ:
> It "feels" that fee estimation is now faster also

I would definitely expect that. The fee calculation and size estimation are now just a bunch of additions. No more bytestring being generated or fake signatures, or fake serializations... Just adding numbers.

lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
paweljakubas: yes, I see it now

KtorZ:
bors merge

KtorZ:
bors retry

KtorZ:
bors retry

KtorZ:
bors retry

KtorZ:
bors r

PR #2351, opened 2020-11-26
Clean up slotting functions

rvl:
### Overview

I am looking at the changes we need to make in #2245 to get the tests to pass.

The slotting module has a few functions that aren't needed any more.

This PR gets rid of them, so that it's easier to understand the code.

### Comments

This branch doesn't conflict with #2245.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -283,7 +283,6 @@ import Cardano.Wallet.Primitive.Slotting
     , TimeInterpreter
     , currentEpoch
     , endTimeOfEpoch
-    , epochSucc
     , firstSlotInEpoch
     , ongoingSlotAt
     , startTime
@@ -384,7 +383,7 @@ import Data.List.NonEmpty
 import Data.Map.Strict
     ( Map )
 import Data.Maybe
-    ( catMaybes, fromMaybe, isJust )
+    ( catMaybes, isJust )
 import Data.Proxy
     ( Proxy (..) )
 import Data.Quantity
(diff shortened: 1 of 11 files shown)


lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1811,12 +1810,6 @@ getNetworkInformation st nl = do
         handlePastHorizonException _ =
             MaybeT (pure Nothing)
 
-    -- Unsafe constructor for the next epoch. Chances to reach the last epoch
-    -- are quite unlikely in this context :)
-    unsafeEpochSucc :: HasCallStack => W.EpochNo -> W.EpochNo
```
Anviking, line 1816: 👍

lib/core/test/unit/Cardano/Wallet/Primitive/Slotting/Helpers.hs
```diff
@@ -0,0 +1,209 @@
...
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedLabels #-}
+{-# LANGUAGE TypeApplications #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+-- | Legacy slotting API functions.
+-- Used as a reference in tests.
+
+module Cardano.Wallet.Primitive.Slotting.Helpers
```
Anviking: Happy to see this moved to its own module, but is this not more `Legacy` or `Simple` than `Helpers`?

Anviking (approved):

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -692,9 +691,8 @@ monitorStakePools tr gp nl DBLayer{..} =
         -- #2196: We need the try. Arguably we shouldn't need it.
         liftIO (try @SomeException (timeInterpreter nl (epochOf currentSlot))) >>= \case
             Left _ -> return ()
-            Right currentEpoch -> do
-                let subtractTwoEpochs = epochPred <=< epochPred
-                forM_ (subtractTwoEpochs currentEpoch) $ \latestRetirementEpoch -> do
+            Right currentEpoch | currentEpoch >= 2 -> do
```
Anviking: Came to think about: isn't there an otherwise case missing now?

rvl: Yes - oops.

PR #2352, opened 2020-11-26
Rework slotting functions

rvl:
### Overview

I am looking at the changes in #2245 related to slotting. 

The way the error handling is, it seems like we are walking around with our shoelaces tied together.

#### Times relative to system start

Part of the problem is that converting a UTC time into a time relative to the system start can fail. (`UTCTimeToRel :: UTCTime -> Qry (Maybe Cardano.RelativeTime)`). But, really in most cases it won't. The best way to deal with that is to push relative time conversions outside of the queries. Then the caller of the slotting can make decisions about how to get a RelativeTime.

Things like converting a slot range to a time range become much simpler.

#### Shorten Slotting module.

We already chopped out a lot of old code in #2351.

But if we use unembellished types from ouroboros-consensus, this module can be even shorter. `Qry` is now just a  `ReaderT` wrapper around `HF.Qry`.

Running a query is to just apply the start time parameter to `runReaderT` and then call `Ouroboros.Consensus.HardFork.History.Qry.interpretQuery`.

#### Mixing with `IO`

Sometimes we want to know the chain-relative time for "now". A simplification is to assume that "now" is always after the start time. In the corner case, it suffices to use the relative time of 0. Now callers of this function don't need to worry about whether it can fail somehow.

#### Relative times are usually better - e.g. Sync progress

The calculation of sync progress with UTC is something like `(b - a - s) / (b - s)`. But with relative times it's just `a / b`.

#### Separate running a query from obtaining the time interpreter

Currently the TimeInterpreter is fetched from the network layer through an IORef (or similar). Then the query is run in IO, when most of the time it's a pure computation (due to the changes above). This just muddles things up.

#### Stake pool retirement time

Because stake pool retirement epochs are almost certainly past time safe zone, we need to use the unsafeExtend function to assume that the current epoch is the final epoch.

#### Transaction expiry (TTL)

Expiry times need to be converted fo/from slots. But these are rarely going to be outside of the safe zone. If they are, it is fine to refuse to convert, and show "Nothing" (for incoming tx) or reject the payment request (for outgoing tx).

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
@@ -2,7 +2,6 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE KindSignatures #-}
-{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedLabels #-}
 {-# LANGUAGE Rank2Types #-}
 
@@ -14,29 +13,41 @@
 -- @UTCTime@.
 
 module Cardano.Wallet.Primitive.Slotting
-    ( -- * New api using ouroboros-concensus
-      -- ** Queries
-      currentEpoch
-    , epochAt
+    ( -- ** Queries
+      Qry
+    , currentEpoch
     , epochOf
-    , startTime
+    , slotToUTCTime
+    , slotToRelTime
     , toSlotId
+    , slotRangeFromRelativeTimeRange
     , slotRangeFromTimeRange
     , firstSlotInEpoch
     , ongoingSlotAt
     , ceilingSlotAt
-    , endTimeOfEpoch
+    , timeOfEpoch
+    , getStartTime
     , querySlotLength
     , queryEpochLength
 
-    -- ** Running queries
-    , TimeInterpreter
-    , singleEraInterpreter
-    , mkTimeInterpreter
-    , HF.PastHorizonException (..)
-    , Qry
+      -- ** Blockchain-relative times
+    , RelativeTime
+    , toRelativeTime
+    , toRelativeTimeRange
+    , fromRelativeTime
+    , addRelTime
 
-    -- ** Helpers
+      -- ** What's the time?
+    , currentRelativeTime
+    , getCurrentTimeRelativeFromStart
+
+      -- ** Running queries
+    , TimeInterpreter
+    , mkSingleEraInterpreter
+    , mkTimeInterpreter
+    , PastHorizonException (..)
+
+      -- ** Helpers
     , unsafeEpochNo
     ) where
 
@@ -55,18 +66,23 @@ import Cardano.Wallet.Primitive.Types
     , SlottingParameters (..)
     , StartTime (..)
     , unsafeEpochNo
-    , wholeRange
     )
 import Control.Monad
-    ( ap, liftM, (<=<) )
+    ( join )
 import Control.Monad.IO.Class
     ( MonadIO, liftIO )
+import Control.Monad.Trans.Class
+    ( lift )
+import Control.Monad.Trans.Reader
+    ( ReaderT, ask, runReaderT )
 import Data.Coerce
     ( coerce )
 import Data.Functor.Identity
     ( Identity )
 import Data.Generics.Internal.VL.Lens
     ( (^.) )
+import Data.Maybe
+    ( fromMaybe )
 import Data.Time.Clock
     ( NominalDiffTime, UTCTime, addUTCTime, getCurrentTime )
 import Data.Word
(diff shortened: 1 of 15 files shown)


lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
```diff
@@ -55,123 +66,116 @@ import Cardano.Wallet.Primitive.Types
...
 --
--- This is preferable to asking for the start time of the /next/ epoch, because
--- the next epoch may be outside the forecast range, and result in
--- @PastHorizonException@.
-endTimeOfEpoch :: EpochNo -> Qry UTCTime
-endTimeOfEpoch epoch = do
+-- Querying the end time of /this/ epoch is preferable to querying the start
+-- time of the /next/ epoch, because the next epoch may be outside the forecast
+-- range, and result in 'PastHorizonException'.
+timeOfEpoch :: EpochNo -> Qry (UTCTime, UTCTime)
```
Anviking, line 156: So `timeOfEpoch` returns the _end_ time, and `slotToUTCTime` returns the _start_ time? I think this is confusing.

Shouldn't this function be keep it this function as `endTimeOfEpoch`?

Anviking: Ah, sorry nvm — just noticed the `(UTCTime, UTCTime)`

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
```diff
@@ -55,123 +66,116 @@ import Cardano.Wallet.Primitive.Types
...
---
+{-------------------------------------------------------------------------------
+                                    Queries
+-------------------------------------------------------------------------------}
 
-currentEpoch :: MonadIO m => TimeInterpreter m -> m (Maybe EpochNo)
-currentEpoch ti = ti . epochAt =<< liftIO getCurrentTime
+-- | This is 'Ouroboros.Consensus.HardFork.History.Qry.Qry' wrapped in a reader
+-- to provide the blockchain system start time as context.
+type Qry = ReaderT StartTime HF.Qry
```
Anviking, line 120: Ok, neat!

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
```diff
@@ -201,124 +202,140 @@ ongoingSlotAt x = do
...
+    toRelativeTimeOrZero start <$> liftIO getCurrentTime
+
+-- | The current system time, compared to the blockchain start time of the given
+-- 'TimeInterpreter'.
 --
--- We cannot manually specify when the fetching happens.
+-- If the current time is before the system start (this would only happen when
+-- launching testnets), the relative time is reported as 0.
+currentRelativeTime :: MonadIO m => TimeInterpreter m -> m RelativeTime
+currentRelativeTime ti = ti getStartTime >>= getCurrentTimeRelativeFromStart
```
Anviking: Rebasing #2245 I found that if `m ~ ExceptT PastHorizonException IO` then suddenly the `currentRelativeTime` can fail… with `PastHorizonException`, which doesn't make sense.

I changed it to
```haskell
currentRelativeTime :: MonadIO m => TimeInterpreter n -> m RelativeTime
```
but we can deal with that in #2245.

Anviking (approved):
Ok, separating the `RelativeTime` `Maybe` problem seems fair.

> Currently the TimeInterpreter is fetched from the network layer through an IORef (or similar). Then the query is run in IO, when most of the time it's a pure computation (due to the changes above). This just muddles things up.

Maybe. I think there's value in allowing the `TimeInterpreter` including the `IORef` to escape from the NetworkLayer and to be modified by combinators.

Anviking:
@rvl I rebased off master and added a hint fix commit. I'll leave it to you to squash and merge.

Although the tests are suddenly failing 🤔 That seems weird.

```
test/unit/Cardano/DB/Sqlite/DeleteSpec.hs:45:9:
--
  | 1) Cardano.DB.Sqlite.Delete.RefCount waitForFree waits for withRef to finish
  | expected: ((),False)
  | but got: ((),True)
  |  
  | To rerun use: --match "/Cardano.DB.Sqlite.Delete/RefCount/waitForFree waits for withRef to finish/"
  |  
  | Randomized with seed 2076715646
```

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
rvl: OK thanks.
Maybe we should change `MonadIO` to just `IO`, or `MonadTime` (from io-sim-classes package).

rvl:
Thanks.
I reverted the change I made to a `slotRangeFromTimeRange` test case, and fixed the code accordingly. And squashed everything.

I have never seen that test failure locally. It looks like a new one.

bors r+

PR #2355, opened 2020-11-27
Make unit test faster (from 360s down to 110s)

KtorZ:
# Issue Number

# Overview

- 2a545b7f885a974ff64261538e83e6a250facd35
  :round_pushpin: **sprinkle some 'parallel' in unit tests**
    Some visible gain in time from a baseline of 362s down to 277s with
  -j4 & +RTS -N4. Unfortunately, we can't parallelise the database tests
  as easily.

- 44c88ad1326c7f12412ee581ca26427ee055b5e2
  :round_pushpin: **make db properties tests parallelisable and parallelise them.**
  
- c3f48d92240097ff9e7b5ab197bcea0fb873b48d
  :round_pushpin: **make pool db unit tests run faster (80s -> 16s)**
    I've mainly relaxed some of the coverage constraints (which are really only there to make sure that tests are going through enough interesting cases, but there's no need to expect a _strict_ coverage). I've also made sure to not generate lists that are _too large_ if not needed to not burn CPU for no reason.

- 0d15e194955cf3e3f4229dbc69758a3db0e2ba20
  :round_pushpin: **reduce required number of successes for some long properties**
    - We've seen the rebalance selection property fail in 1400 and 1997 tests, there's no need to go to 10000.

  - The timing test for the logging middleware could really be a unit test, we are using QuickCheck here just
    for the reporting and because we were too lazy to think about good arbitrary values. Testing with 10 values
    is already more than sufficient.

  - depleteUTxO is long, and we are testing for 5 different values, although some have very similar results. I've
    picked a more sparse scale to still get to test various scenarios, but reduce the overall property time.

  - Various properties for 'depleteUTxO' were run with max success of 1000, that's unecessary large.

- 60a6b1993ccc1f6c8648c36e49dc5ff669473aa8
  :round_pushpin: **remove 'JormungandrKey' and surroundings.**
    Fortunately, 'ShelleyKey' is a drop-in replacement for everything that referred to 'JormungandrKey' in the database. It's good to finally do this because, most of the tests were still executed on the 'JormungandrKey' whereas most of what's used in production is actually 'ShelleyKey' ... we know that the underlying implementation is equal, but that's a bit dodgy :| ...

# Comments

Note: this is eating up to 40% of my available RAM though, so about 10 GB. Seems still okay-ish ?


lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -988,12 +987,16 @@ data ApiPoolId
...
             Left $ TextDecodingError $ unwords
             [ "Invalid account public key: expecting a hex-encoded value"
             , "that is 64 bytes in length."]
-        Right pubkey ->
+        Just pubkey ->
             Right $ ApiAccountPublicKey $ ApiT pubkey
+      where
+        xpubFromText :: ByteString -> Maybe XPub
+        xpubFromText = (either (const Nothing) Just <$> fromHex @ByteString)
+            >=> xpubFromBytes
```
KtorZ, line 999: Note: used to be in `Cardano.Wallet.Primitive.AddressDerivation.Jormungandr`

lib/core/test/unit/Cardano/Pool/DB/Arbitrary.hs
```diff
@@ -251,7 +255,7 @@ instance Arbitrary SinglePoolCertificateSequence where
             -- We must start with a registration certificate:
             certificates <- (:)
                 <$> (Registration <$> arbitrary)
-                <*> arbitrary
+                <*> scale (min 10) arbitrary
```
rvl: Maybe add a `limitScale = scale . min` if this is going to be used a lot.
But I'm not sure it's good to overuse this.
Normally you would want all sub-values of an arbitrary value to grow in proportion to the scale value.

lib/core/cardano-wallet-core.cabal
```diff
@@ -195,6 +194,7 @@ test-suite unit
       -threaded -rtsopts
       -Wall
       "-with-rtsopts=-M2G"
+      "-with-rtsopts=-N4"
```
rvl, line 197: You could use`-N` without a number to select the number of available cores on the system which is running the test.
I was surprised to find that hspec does not call `setNumCapabilities` when the `hspec --jobs` arg is used.

lib/core/test/unit/Cardano/Byron/Codec/CborSpec.hs
```diff
@@ -67,7 +67,7 @@ import qualified Data.ByteString.Lazy as BL
 {- HLINT ignore spec "Use head" -}
 
 spec :: Spec
-spec = do
+spec = parallel $ do
```
rvl, line 70: Is it possible to have global parallel setting for the unit test-suite?

rvl (approved):
Looks good.

Perhaps the high memory memory usage is due to a space leak.

lib/core/cardano-wallet-core.cabal
KtorZ: > I was surprised to find that hspec does not call setNumCapabilities when the hspec --jobs arg is used.

Me too. Hence the RTS option, although I've tried `-N` but it actually made the test slower for me (while also incapacitating my laptop for 5 minutes :sweat_smile: )... Maybe was due to a lack of RAM when trying to run on all CPU threads.

lib/core/test/unit/Cardano/Byron/Codec/CborSpec.hs
KtorZ: There a few tests we don't run in parallel, like those starting local servers and playing around with ports. Databases were also excluded initially but I've found a way to have them running in parallel too. 

Arguably, we could move the problematic tests in a another test suite like `integration`. I am also not sure at this stage whether we can have a single global `parallel`, maybe by tweaking:

https://github.com/input-output-hk/cardano-wallet/blob/4df2ecc5d5259ce333de6b2dfa39b6ac51b17652/lib/core/test/unit/Main.hs#L9-L10

lib/core/test/unit/Cardano/Pool/DB/Arbitrary.hs
KtorZ: >  But I'm not sure it's good to overuse this.

I agree. Although it's mainly use for database tests at this stage which kinda makes sense to me still. In quite many database property tests we've been generating huge lists of values which doesn't really bring anything more than slowness. 

Perhaps we could simply use a smaller scale factor instead of clamping the whole thing, a slightly modified version of `log` for instance or `sqrt`.

KtorZ: see: 80b0f8d8b79d6c2d56175d775a651ac2acc2217c

KtorZ:
bors merge

PR #2356, opened 2020-11-30
Run github windows tests on PR push

rvl:
### Issue Number

#2189 

### Overview

- Run windows tests when PR branches are pushed.
- Run windows tests when the master branch or bors branches are updated.
- Don't run windows tests on a nightly schedule any more.
- Some nightly test failures were because of a ssh issue on github actions. So added a fix for the branch update script.
- Integration tests are run on bors branches.

### Comments

- `cardano-wallet-core:unit` is failing due to #2526, so it has `continue-on-error: true`
- `cardano-wallet-launcher:unit` is failing, but meh, so it has `continue-on-error: true`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,21 +1,31 @@
+# On branch push, download the windows testing bundle build and run it.
+
 name: cardano-wallet Windows Tests
 
 on:
   push:
     branches:
       - master
+      - bors/staging
+      - bors/trying
+  pull_request:
 
 jobs:
   setup:
     runs-on: windows-2016
     name: Download windows testing bundle
     steps:
-      - name: Fetch Windows testing bundle
+      - name: 'Wait for Hydra build'
+        uses: rvl/hydra-build-products-action@master
+        id: hydra
+        with:
+          hydra: 'https://hydra.iohk.io'
+          jobs: 'cardano-wallet-tests-win64'
+      - name: 'Fetch Windows testing bundle'
         shell: powershell
         run: |
-          $url = "https://hydra.iohk.io/job/Cardano/cardano-wallet/cardano-wallet-tests-win64/latest/download/1"
           $output = "cardano-wallet-tests-win64.zip"
-          Invoke-WebRequest -Uri $url -OutFile $output
+          Invoke-WebRequest -Uri ${{ steps.hydra.outputs.buildProducts }} -OutFile $output
           Expand-Archive -LiteralPath $output -DestinationPath .
           Get-ChildItem
       - name: Save files
@@ -32,7 +42,7 @@ jobs:
       - uses: actions/download-artifact@v2
         with:
           name: cardano-wallet-tests-win64
-      - run: '.\\cardano-wallet-core-test-unit.exe --color'
+      - run: '.\\cardano-wallet-core-test-unit.exe --color +RTS -N2'
 
   cardano-wallet-test-unit:
     name: 'cardano-wallet:unit'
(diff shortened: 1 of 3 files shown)


rvl:
bors try

.github/workflows/ci.yml
```diff
@@ -1,21 +1,32 @@
...
 name: cardano-wallet Windows Tests
 
 on:
   push:
     branches:
       - master
+      - bors/staging
+      - bors/trying
+  pull_request:
+    types: [opened, synchronize, reopened]
```
KtorZ: | type | semantic |
| --- | --- | 
| opened | pull request is created |
| synchronize | commit(s) pushed to the pull request |
| reopened | closed pull request is reopened | 

(source: https://frontside.com/blog/2020-05-26-github-actions-pull_request/#how-do-workflows-trigger-on-pull_request)

Minor remark: maybe 'reopened' isn't needed here :thinking: ? If it's closed, it also means that tests were run previously already.

.github/workflows/ci.yml
```diff
@@ -32,7 +43,7 @@ jobs:
       - uses: actions/download-artifact@v2
         with:
           name: cardano-wallet-tests-win64
-      - run: '.\\cardano-wallet-core-test-unit.exe --color'
+      - run: '.\\cardano-wallet-core-test-unit.exe --color +RTS -N1'
```
KtorZ: :crossed_fingers: didn't help locally for me. Let's see here :man_shrugging:

KtorZ (approved):

.github/workflows/ci.yml
piotr-iohk: From the same source :point_up: :
_When a Github Actions workflow is configured to run on pull requests, by default it will be triggered on three different types of activities: opened, synchronize, and reopened._

```
name: Pull Request Workflow
on: pull_request
```
I think default is fine, and no need to specify types. :man_shrugging:

piotr-iohk (approved):

.github/workflows/ci.yml
rvl: OK https://github.com/input-output-hk/cardano-wallet/pull/2527#issuecomment-780439348 I believe you :+1:

I will change this to `-N2` to override the `ghc -with-rtsopts=-N4` setting from cabal. Then it should set the number of threads to match [the VM specs](https://docs.github.com/en/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources).

(edit: `-N` by itself did not seem to work)

.github/workflows/ci.yml
rvl: Yes you're right, I was just getting very confused by the docs.

It does seem like the 'reopened' event is not useful for this kind of workflow, but I will use the default to keep it simpler.

.github/workflows/ci.yml
KtorZ: Sounds like getting rid of parallel didn't help :( 
We may have some property tests that are a bit too much for Windows?

rvl: Something odd is going on with RTS opts. It's supposed to have `-M2G` set for both windows and linux.

rvl:
I have skipped the `Cardano.Wallet.DB.Sqlite` unit tests until we can debug the strangely high memory usage - ticket #2526.

bors r+

.github/workflows/ci.yml
KtorZ: Something I've experienced actually with specifying rts opts directly in the cabal file is that:

```
      "-with-rtsopts=-M2G"
      "-with-rtsopts=-N4"
```

This doesn't work. Only the last option gets taken into account. But this does:

```
      - '"-with-rtsopts=-M2G -N4"'
```

Issue #2207, opened 2020-10-05
Flaky SHELLEY_MIGRATE_01_big_wallet


Anviking:
# Context
 
 

Seen fail once: https://github.com/input-output-hk/cardano-wallet/pull/2201#issuecomment-702835170

# Test Case

https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs#L147-L231

# Failure / Counter-example

```
Build failed:

ci/hydra-build:required

Failures:

 src/Test/Integration/Scenario/API/Shelley/Migrations.hs:147:5:
 1) API Specifications, SHELLEY_MIGRATIONS, SHELLEY_MIGRATE_01_big_wallet - migrate a big wallet requiring more than one tx
 uncaught exception: IOException of type UserError
 user error (Waited longer than 90s (more than 2 epochs) for an action to resolve. Action: "wallet balance = expectedBalance". Error condition: Just (HUnitFailure (Just (SrcLoc {srcLocPackage = "cardano-wallet-core-integration-2020.9.30-HOEflGzxOTuDeHvKdxL76x", srcLocModule = "Test.Integration.Scenario.API.Shelley.Migrations", srcLocFile = "src/Test/Integration/Scenario/API/Shelley/Migrations.hs", srcLocStartLine = 218, srcLocStartCol = 27, srcLocEndLine = 218, srcLocEndCol = 62})) (ExpectedButGot Nothing "Quantity {getQuantity = 10000094246400}" "Quantity {getQuantity = 9000090649400}")))

 To rerun use: --match "/API Specifications/SHELLEY_MIGRATIONS/SHELLEY_MIGRATE_01_big_wallet - migrate a big wallet requiring more than one tx/"

```

---

# Resolution 

---

# QA

Anviking:
https://hydra.iohk.io/api/scmdiff?branch=&uri=https%3A%2F%2Fgithub.com%2Finput-output-hk%2Fcardano-wallet.git&type=git&rev2=ad8181b61698efc96cd9c44ebf46018c8c7b8c3d&rev1=a1e602a1077f7f6328bcc593beb4f63906cc250b

#2181 should have been included, so the error shouldn't be due to retrying with hard-coded mnemonic.

Anviking:
Could be due to fork + retry always failing because of hard-coded mnemonics.

Anviking:
[...]

Issue #2118, opened 2020-09-08
Property `prop_rebalanceSelection` occasionally fails.


jonathanknowles:
# Context

Instances of this failure:
 
1. https://hydra.iohk.io/build/4063979/nixlog/7
1. https://hydra.iohk.io/build/4208599/nixlog/1
1. https://hydra.iohk.io/build/4605661/nixlog/55
1. https://hydra.iohk.io/build/4865823/nixlog/3
1. https://hydra.iohk.io/build/4866583/nixlog/1

# Test Case

`prop_rebalanceSelection`

# Failure / Counter-example

```hs
      Falsifiable (after 5023 tests):
         SaveMoney
         Coin {getCoin = 105080}
         selection (before):
         inputs: - 2nd 409786d4 (~ 166350 @ 61646472...64722d33)
         withdrawal: 618680
         reclaim: -0
         outputs:
           - 74251 @ 61646472...64722d33
           - 76285 @ 61646472...64722d33
           - 163463 @ 61646472...64722d32
           - 198051 @ 61646472...64722d30
         change: [272980]
         deposit: -0

         selection (after):
         inputs: - 2nd 409786d4 (~ 166350 @ 61646472...64722d33)
         withdrawal: 618680
         reclaim: -0
         outputs:
           - 74251 @ 61646472...64722d33
           - 76285 @ 61646472...64722d33
           - 163463 @ 61646472...64722d32
           - 198051 @ 61646472...64722d30
         change: [105080]
         deposit: -0

         delta (before): 0
         delta (after):  167900
         total fee:      167900
         remaining fee:  0


  To rerun use: --match "/Cardano.Wallet.Primitive.Fee/prop_rebalanceSelection/The fee balancing algorithm converges for any coin selection./"

Randomized with seed 1973157653
```

# Resolution 

---

# QA

jonathanknowles:
Seen again here: https://hydra.iohk.io/build/4605661/nixlog/55/tail.

Extract:

```hs
  test/unit/Cardano/Wallet/Primitive/FeeSpec.hs:379:9: 
  1) Cardano.Wallet.Primitive.Fee.prop_rebalanceSelection The fee balancing algorithm converges for any coin selection.
       Falsifiable (after 1412 tests):
         SaveMoney
         Coin {getCoin = 75091}
         selection (before):
         inputs: - 1st 1b0e0d01 (~ 191127 @ 61646472...64722d33)
         withdrawal: 661092
[...]

PR #2357, opened 2020-11-30
Fix prop_rebalanceSelection wrt #2118

hasufell:
# Issue Number

#2118 

----

Wasn't able to write a reliable generator that always triggers this case... might be worth adding a unit test.

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
@@ -239,7 +239,7 @@ spec = describe "SHELLEY_MIGRATIONS" $ do
             Empty >>= flip verify
             [ expectField
                 #distribution
-                ((`shouldBe` (Just 99)) . Map.lookup 100_000_000_000)
+                ((`shouldBe` (Just 100)) . Map.lookup 100_000_000_000)
             ]
 
     it "SHELLEY_MIGRATE_02 - \
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
@@ -239,7 +239,7 @@ spec = describe "SHELLEY_MIGRATIONS" $ do
             Empty >>= flip verify
             [ expectField
                 #distribution
-                ((`shouldBe` (Just 100)) . Map.lookup 100_000_000_000)
+                ((`shouldBe` (Just 99)) . Map.lookup 100_000_000_000)
             ]
 
     it "SHELLEY_MIGRATE_02 - \
--- a/lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Migration.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Migration.hs
@@ -111,7 +111,7 @@ depleteUTxO feeOpts batchSize utxo =
       where
         noDust :: TxOut -> Maybe Coin
         noDust (TxOut _ c)
-            | c <= dustThreshold feeOpts = Nothing
+            | c < dustThreshold feeOpts = Nothing
             | otherwise = Just c
 
     -- | Attempt to balance the coin selection by reducing or increasing the
--- a/lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Migration.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Migration.hs
@@ -111,7 +111,7 @@ depleteUTxO feeOpts batchSize utxo =
       where
         noDust :: TxOut -> Maybe Coin
         noDust (TxOut _ c)
-            | c < dustThreshold feeOpts = Nothing
+            | c <= dustThreshold feeOpts = Nothing
             | otherwise = Just c
 
     -- | Attempt to balance the coin selection by reducing or increasing the
--- a/lib/core/src/Cardano/Wallet/Primitive/Fee.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Fee.hs
@@ -353,7 +353,7 @@ divvyFee (Fee f0) outs = go f0 [] outs
 -- return an empty list.
 coalesceDust :: Coin -> [Coin] -> [Coin]
 coalesceDust threshold coins
-    | balance coins < getCoin threshold =
+    | balance coins <= getCoin threshold =
         []
     | otherwise =
         let
--- a/lib/core/test/unit/Cardano/Wallet/Primitive/FeeSpec.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Primitive/FeeSpec.hs
@@ -350,7 +350,7 @@ spec = do
             }) (Right $ FeeOutput
             { csInps = [2,2]
             , csOuts = []
-            , csChngs = [1]
+            , csChngs = []
             })
 
     parallel $ describe "Fee Calculation: Generators" $ do
--- a/lib/core/test/unit/Cardano/Wallet/Primitive/FeeSpec.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Primitive/FeeSpec.hs
@@ -382,6 +382,35 @@ spec = do
             $ property
             $ withMaxSuccess 2000
             $ forAllBlind genSelection' prop_rebalanceSelection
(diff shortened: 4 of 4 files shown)


lib/core/src/Cardano/Wallet/Primitive/Fee.hs
```diff
@@ -357,7 +357,7 @@ coalesceDust threshold coins
         []
     | otherwise =
         let
-            filtered = L.filter (> threshold) coins
+            filtered = L.filter (\x -> x >= threshold && x > Coin 0) coins
```
hasufell: Dusty coins are anything that is **below** the threshold. So we remove dusty coins here (and anything that's `0`) and then split the dust value across the good coins.

lib/core/test/unit/Cardano/Wallet/Primitive/FeeSpec.hs
```diff
@@ -170,14 +170,14 @@ spec = do
...
 
-        -- Fee divvied, dust removed (dust = 1)
+        -- Fee divvied, dust removed (dust = 2)
         feeUnitTest id (FeeFixture
             { fInps = [20,20,20]
             , fOuts = [14,18,19]
             , fChngs = [6,2,1]
             , fUtxo = []
             , fFee = 3
-            , fDust = 1
```
hasufell: Without this fix, the error was

```
  test/unit/Cardano/Wallet/Primitive/FeeSpec.hs:622:5:
  1) Cardano.Wallet.Primitive.Fee, Fee calculation : unit tests, CoinSelection (inps=[20,20,20]outs=[14,18,19]chngs=[6,2,1]), UTxO=[]), fee=3 --> Right (FeeOutput {csInps = [20,20,20], csOuts = [14,18,19], csChngs = [6]})
       expected: Right (FeeOutput {csInps = [20,20,20], csOuts = [14,18,19], csChngs = [6]})
        but got: Right (FeeOutput {csInps = [20,20,20], csOuts = [14,18,19], csChngs = [4,1,1]})
```

which looking at it now makes sense. `1` in change output is allowed when dust threshold is `1`, so this should pass. Changing the threshold to `2` makes the unit test more exciting though.

hasufell:
Ran 100k tests:

```
Cardano.Wallet.Primitive.Fee
  prop_rebalanceSelection
    The fee balancing algorithm converges for any coin selection.
      +++ OK, passed 100000 tests; 619923 discarded:
      94.322% reserve > fee
      45.096% fee > delta
      41.094% reserve > 0 && fee > delta
      39.418% reserve > fee && fee > delta

Finished in 229.6284 seconds
1 example, 0 failures
```

lib/core/test/unit/Cardano/Wallet/Primitive/FeeSpec.hs
```diff
@@ -382,6 +382,35 @@ spec = do
             $ property
             $ withMaxSuccess 10000
             $ forAllBlind genSelection' prop_rebalanceSelection
+        it "If change is a coin equal the dust threshold, fee balancing still converges, wrt #2118" $
+            let dt = Coin {getCoin = 114754}
+                cs = CoinSelection {inputs = [
```
hasufell, line 387: This is copy pasted from the last test failure with seed `1510629119`.

lib/core/test/unit/Cardano/Wallet/Primitive/FeeSpec.hs
```diff
@@ -382,6 +382,35 @@ spec = do
...
+                    ,TxOut {address = Address {unAddress = "addr-2"}
+                        , coin = Coin {getCoin = 2336}}
+                    ,TxOut {address = Address {unAddress = "addr-3"}
+                        , coin = Coin {getCoin = 86104}}
+                    ,TxOut {address = Address {unAddress = "addr-0"}
+                        , coin = Coin {getCoin = 74851}}
+                ]
+                , change = [dt]
+                , deposit = 0}
+            in property $ prop_rebalanceSelection cs dt
```
hasufell, line 412: This is a nice trick to reuse the property test function as a unit test. It will only run once.

KtorZ (approved):

Anviking:
Failures might actually be proper ones — `SHELLEY_MIGRATE_01_big_wallet` might need to be adjusted if the fee calculation has changed.

hasufell:
I couldn't reproduce the failure locally:

```
API Specifications
  SHELLEY_MIGRATIONS
[cardano-wallet.integration:Notice:30] [2020-12-02 15:08:43.63 UTC] Configuring cluster in /tmp/test-790360365591b84b
[cardano-wallet.integration:Notice:39] [2020-12-02 15:08:45.85 UTC] Transitioning from Byron to Shelley... Please wait 20s...
[cardano-wallet.integration:Notice:39] [2020-12-02 15:09:07.90 UTC] Setting up faucet...
[cardano-wallet.integration:Notice:39] [2020-12-02 15:09:11.34 UTC] 4 stake pools are being registered on chain... Please wait 60s until active... Can be skipped using NO_POOLS=1.
[cardano-wallet.integration:Notice:588] [2020-12-02 15:10:28.27 UTC] http://127.0.0.1:45683/
    SHELLEY_MIGRATE_01_big_wallet -  migrate a big wallet requiring more than one tx

Finished in 145.4361 seconds
1 example, 0 failures
```

I'm suspecting there's *some* randomness to the values? Or is it because I'm only running this particular test and not all?

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Migration.hs
```diff
@@ -111,7 +111,7 @@ depleteUTxO feeOpts batchSize utxo =
       where
         noDust :: TxOut -> Maybe Coin
         noDust (TxOut _ c)
-            | c < dustThreshold feeOpts = Nothing
+            | c <= dustThreshold feeOpts = Nothing
```
hasufell: So this is the same issue in the Migration coin selection algorithm like in `coalesceDust` I believe. But I can't tell if this actually fixes #2207

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
```diff
@@ -239,7 +239,7 @@ spec = describe "SHELLEY_MIGRATIONS" $ do
             Empty >>= flip verify
             [ expectField
                 #distribution
-                ((`shouldBe` (Just 100)) . Map.lookup 100_000_000_000)
+                ((`shouldBe` (Just 99)) . Map.lookup 100_000_000_000)
```
hasufell: So there's one less... I'm suspecting because it got merged with the dust coins, which are now properly treated as dust?

hasufell:
bors try

hasufell:
bors try

hasufell:
bors r+

----

oops, this should have been a `try` :sweat_smile:

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
KtorZ: Just to make sure, could you control what the distribution looks like? (you know, if we've lost 100K Ada along the way, it is quite a problem ^.^)

hasufell: This is the distribution:

```
distribution = Just
  (fromList
    [ (10               , 0)
    , (100              , 0)
    , (1000             , 0)
    , (10000            , 0)
    , (100000           , 0)
    , (1000000          , 0)
    , (10000000         , 1)
    , (100000000        , 6)
    , (1000000000       , 0)
    , (10000000000      , 0)
    , (100000000000     , 99)
    , (1000000000000    , 1)
    , (10000000000000   , 0)
    , (100000000000000  , 0)
    , (1000000000000000 , 0)
    , (10000000000000000, 0)
    , (45000000000000000, 0)
    ]
  )
```

----

for comparison, the old distribution:

```
distribution = Just
  (fromList
    [ (10               , 0)
    , (100              , 0)
    , (1000             , 0)
    , (10000            , 0)
    , (100000           , 0)
    , (1000000          , 84)
    , (10000000         , 8)
    , (100000000        , 0)
    , (1000000000       , 0)
    , (10000000000      , 0)
    , (100000000000     , 100)
    , (1000000000000    , 0)
    , (10000000000000   , 0)
    , (100000000000000  , 0)
    , (1000000000000000 , 0)
    , (10000000000000000, 0)
    , (45000000000000000, 0)
    ]
  )
```

PR #2358, opened 2020-12-01
fix 'waitForFree' test in Sqlite.DeleteSpec (increase delay)

KtorZ:
# Issue Number

N/A

# Overview

- [ ] I have fix 'waitForFree' test in Sqlite.DeleteSpec (increase delay)

  The symptoms were that this test was sometimes failing. Looking at the
  scenario itself, it seems that it could simply be a race condition due
  to the machine being busy running other tests (now ran in parallel).

  The test is simple: it runs two actions concurrently:

  - One action which grabs a database reference and simulate some busy work
  - One action which wait for all db references to be "freed"

  The problem is that, the second action should only run after the
  database reference has been reserved already and this was "ensured" by
  adding a very small delay of 10 microseconds before the second action.
  Yet, now that tests are executed in parallel, we have many threads
  floating around and it could very much be that the second action
  sometimes get scheduled earlier than the first one, causing the second
  action to run and execute immediately since there's no database
  referenced that has been grabbed yet.

# Comments

### DIFF
--- a/lib/core/test/unit/Cardano/DB/Sqlite/DeleteSpec.hs
+++ b/lib/core/test/unit/Cardano/DB/Sqlite/DeleteSpec.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE NumericUnderscores #-}
+
 -- |
 -- Copyright: © 2018-2020 IOHK
 -- License: Apache-2.0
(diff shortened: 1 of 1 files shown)


Anviking (approved):
Ok

KtorZ:
bors merge

jonathanknowles:
Ah, just discovered this PR.

I expect that it probably fixes https://github.com/input-output-hk/cardano-wallet/issues/2361, which failed 10% of the time over 100 runs of the test suite.

PR #2359, opened 2020-12-01
Make the test cluster works through Shelley and Allegra / cardano-node-1.24.0 bump

KtorZ:
# Issue Number

# Overview

This seems to work node-wise (although I am using the latest revision from the node, which should be soon tagged as 1.24.0). A few notes:

  - We need to adjust the code for constructing transactions (in particular
    'mkUnsignedTx') since they are serialized in slightly different
    ways. For immediate testing purposes, we may "brute-force" this and
    simply use the Allegra era format all the time. But in practice, we
    may want to release a version of the wallet that can do both and
    cross the hard fork nicely. Maybe there's something we can use from the cardano-api 
    to make this easier :thinking: ?

  - The launcher code needs some cleanup so that all the commands drop
    the now-deprecated 'shelley' prefix and use top-level commands
    instead.

# Comments


lib/shelley/test/data/cardano-node-shelley/node.config
```diff
@@ -26,13 +26,13 @@ PBftSignatureThreshold: 1
...
-LastKnownBlockVersion-Major: 1
+LastKnownBlockVersion-Major: 2
 LastKnownBlockVersion-Minor: 0
 LastKnownBlockVersion-Alt: 0
 
 # Launcher code will submit an update proposal to trigger the hard-fork.
-TestShelleyHardForkAtVersion: 1
-
+TestShelleyHardForkAtEpoch: 1
+TestAllegraHardForkAtEpoch: 2
```
Anviking, line 35: The original intention of the `withCluster` was to allow the caller to run code on each era of the chain. Looking now, perhaps the existing doc comment doesn't make it clear, but:

could we still add a note about the fork(s) being triggered automatically in the documentation for `withCluster`?

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -1758,8 +1759,8 @@ instance ToText ClusterLog where
     toText = \case
         MsgStartingCluster dir ->
             "Configuring cluster in " <> T.pack dir
-        MsgWaitingForFork ->
-            "Transitioning from Byron to Shelley... Please wait 20s..."
+        MsgWaitingForFork fork ->
+            "Transitioning from " <> fork <> " ... Please wait 20s..."
```
Anviking, line 1774: 👍

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -747,14 +778,47 @@ fromShelleyTx (SL.Tx bod@(SL.TxBody ins outs certs wdrls _ _ _ _) _ mmd) =
...
+  where
+    SL.Tx bod@(SL.TxBody ins outs certs wdrls _ _ _ _) _ mmd = tx
+
+fromAllegraTx
+    :: ( SL.ShelleyBased era
+       , SL.Core.TxBody era ~ MA.TxBody era
+       , SL.Core.Value era ~ SL.Coin
+       , SL.Core.Metadata era ~ MA.Metadata era
+       , Ord (SL.Core.Script era)
+       )
```
rvl, line 808: Perhaps there is a single AllegraEra type that satifies all of these constraints?

lib/shelley/test/data/cardano-node-shelley/node.config
rvl: Yes, if we need the wallet to work in two eras then we ought to run some tests in both eras.
But perhaps while the wallet is effectively allegra-only it can be simpler and use the hard fork schedule.

rvl (approved):
Looks good to me, but sometimes it fails to register the stake pools. I am trying to see why that is.

rvl:
A few Cabal constraint fixes to the snapshot: input-output-hk/cardano-haskell#35.

rvl:
bors try

KtorZ:
bors merge

rvl:
bors merge

Anviking:
Trying one more time:
bors r+

Anviking:
Trying with -j 3
bors r+

PR #2362, opened 2020-12-02
Test how TimeInterpreter logging is affected by combinators

Anviking:
# Issue Number

ADP-550

# Overview

- [x] Added basic unit tests ensuring the TimeInterpreter logging is affected by combinators correctly

# Comments

```
Cardano.Wallet.Primitive.Slotting
  slotting
    TimeInterpreter conversions beyond the safe zone
      normally fails and logs failures as Notice
      (neverFails "because" ti) logs failures as Error
      (unsafeExtendSafeZone ti) doesn't fail nor log
      (expectAndThrowFailures ti) fails and logs as Notice
```

### DIFF
--- a/lib/core/test/unit/Cardano/Wallet/Primitive/SlottingSpec.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Primitive/SlottingSpec.hs
@@ -15,8 +15,6 @@ import Cardano.BM.Data.Severity
     ( Severity (..) )
 import Cardano.BM.Data.Tracer
     ( HasSeverityAnnotation (..) )
-import Cardano.BM.Trace
-    ( traceInTVarIO )
 import Cardano.Slotting.Slot
     ( SlotNo (..) )
 import Cardano.Wallet.Gen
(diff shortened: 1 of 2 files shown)


lib/core/test/unit/Cardano/Wallet/Primitive/SlottingSpec.hs
```diff
@@ -111,6 +145,72 @@ spec = do
...
+        before setupTestLogging $
+            describe "TimeInterpreter conversions beyone the safe zone" $ do
+
+            startTime <- runIO $ StartTime <$> getCurrentTime
+            let failingQry = slotToUTCTime (SlotNo 100000)
+
+            it "normally fails and logs failures as Notice" $ \(tr, tvar) -> do
+                let ti = mkTimeInterpreter tr startTime (pure forkInterpreter)
+                res <- runExceptT $ interpretQuery ti failingQry
+                logs <- readTVarIO tvar
```
Anviking: Extracting the common parts with a test helper was more tricky than I expected due to the different return-types, so leaving it like this with some duplication. Shouldn't matter that much.

rvl: I changed it to use `Test.Utils.Trace.captureLogging`.

rvl (approved):
:+1:

Anviking:
Thanks!
bors r+

PR #2363, opened 2020-12-02
Perhaps: Don't call unsafeExtendSafeZone when out of sync

Anviking:
# Issue Number

ADP-524

# Context

We currently use `unsafeExtendSafeZone` for:
1. getTxExpiry for setting TTL in the future
2. mkApiTransaction for showing TTL in the future
3. mkShelleyWallet for delegation status changes
4. combineDbAndLsqData for future pool retirement dates

# Overview

- [x] `unsafeExtendSafeZone` now does nothing if the current time is outside the horizon
- [ ] When we know what to do, some test would be nice, if possible

# Comments

- In a way I think it makes sense to first wait for the node to sync, before extrapolating.
- Will prevent extrapolating from Byron parameters on Mainnet, but not for custom chains which could still be in Byron.
- Pretty magical, stateful and weird though ( but could be solved by making the currentTimeInHorizon call explicitly on the call-sites)
- [Slack thread](https://input-output-rnd.slack.com/archives/GBT05825V/p1606818042019800)

## Alternatives
- Make `unsafeExtendSafeZone` extend the safe-zone based on the number of eras in the summary (a bit tricky, would probably add another version `interpretQuery` rather than a TimeInterpreter combinator)
- Use a combination of `currentTimeInHorizon :: IO Bool` and `unsafeExtendSafeZoneIf` on the call-sites to make this less magical
- Completely remove `unsafeExtendSafeZone`. 
    - Adjust the integration tests or the default TTL to work together
    - Write a custom function `forecastEpochStartFrom :: RelativeTime -> EpochNo -> Qry UTCTime` for delegation and pool retirement
        - would fail when the current time is outside the safe-zone
        - I think this the only function we need, but might have forgotten something

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -1626,7 +1626,7 @@ signPayment
     -> CoinSelection
     -> ExceptT ErrSignPayment IO (Tx, TxMeta, UTCTime, SealedTx)
 signPayment ctx wid argGenChange mkRewardAccount pwd md ttl cs = db & \DBLayer{..} -> do
-    txExp <- liftIO $ getTxExpiry ti ttl
+    txExp <- withExceptT ErrSignPaymentIncorrectTTL $ getTxExpiry ti ttl
     era <- liftIO $ currentNodeEra nl
     withRootKey @_ @s ctx wid pwd ErrSignPaymentWithRootKey $ \xprv scheme -> do
         let pwdP = preparePassphrase scheme pwd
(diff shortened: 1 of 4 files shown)


Anviking:
Alternative direction -> #2365

rvl:
> We currently use unsafeExtendSafeZone for:
> 1. getTxExpiry for setting TTL in the future
> 2. mkApiTransaction for showing TTL in the future
> 3. mkShelleyWallet for delegation status changes
> 4. combineDbAndLsqData for future pool retirement dates

1. If tx expiry can't be determined for a pending outgoing transaction, just don't show it.
2. If tx expiry time can't be converted to a slot when making a payment, refuse the request and return an error. (This will also make 1 much less likely). The user should either wait for their node to sync, or use a shorter TTL. Integration tests should use a shorter TTL because of their parameters.
3. mkShelleyWallet is only used by post wallet operations. It will never have delegation info containing an epoch and time.
4. This is the only one we really need to worry about. I think we should tweak the API spec and make the `ApiEpochInfo->epochStartTime :: UTCTime` field into a `Maybe`. So if we don't know the time, don't report anything.

Anviking:
I don’t think we should tweak the Api to remove the times of delegation change and pool retirement today.

Maybe #2363 can be an acceptable solution today, and then we can remove unsafeExtendSafeZone from TTL, and then see what we want to do for 3 and 4.

Given that the unlimited safe-zone of the final era might not be dependable, and times being useful for humans, I suspect that we always want to provide an estimate.

Anviking:
bors try

PR #2364, opened 2020-12-02
SMASH health check integration tests

piotr-iohk:
# Issue Number

ADP-496

# Overview

- 42118df8199526aa1df609a873371b020c33137f
  SMASH health check integration tests
  
- b0071346d98d9b14c8db48d32737557a7af9472a
  Add error codes to responsesGetSmashHealth in swagger file
  
- 98124561b7377c472b77af373f5c1c7e8cf11a44
  Add health check endpoint to SMASH stub

# Comments

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
@@ -17,6 +17,7 @@ import Prelude
 
 import Cardano.Wallet.Api.Types
     ( ApiCertificate (JoinPool, QuitPool, RegisterRewardAccount)
+    , ApiHealthCheck
     , ApiStakePool (flags)
     , ApiStakePoolFlag (..)
     , ApiT (..)
@@ -26,6 +27,7 @@ import Cardano.Wallet.Api.Types
     , DecodeAddress
     , DecodeStakeAddress
     , EncodeAddress
+    , HealthCheckSMASH (..)
     , WalletStyle (..)
     )
 import Cardano.Wallet.Primitive.AddressDerivation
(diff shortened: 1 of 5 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -1140,6 +1142,51 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
...
 
+    it "STAKE_POOLS_SMASH_HEALTH_01 - Can check SMASH health when configured" $
+        \ctx -> runResourceT $ bracketSettings ctx $ do
+            smashUrl <- liftIO $ getEnv "CARDANO_WALLET_SMASH_URL"
+            updateMetadataSource ctx (T.pack smashUrl)
+            r <- request @ApiHealthCheck
+                ctx Link.getCurrentSMASHHealth
+                Default Empty
+            expectResponseCode HTTP.status200 r
+            -- TODO: We may consider adding healthcheck API to SMASH stub
```
hasufell: Yes, I actually think we should!

Shouldn't be hard, needs to go here https://github.com/input-output-hk/cardano-wallet/blob/96b0b9f7d4d051035a33962408216572bc23926a/lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs#L901

Then we just do something like

```hs
        let bytes = Aeson.encode (HealthStatusSMASH "OK" "1.2.0")
        BL8.writeFile (baseDir </> "status") bytes
```

piotr-iohk: Thx, added in 9812456.

hasufell (approved):

PR #2308, opened 2020-11-09
Extend seq state

paweljakubas:
# Issue Number

https://jira.iohk.io/browse/ADP-537

# Overview

- [x] I have renamed AccountingStyle -> Role
- [x] I have dealt with migration of this move
- [x] I have added knownScripts to SeqState
- [x] I have realigned unit tests
- [x] I have implemented isShared
- [x] I have added property tests for isShared
- [x] I have added address pool gap for verification keys to SeqState, updated other parts of code, migration
- [x]  I have extended Sqlite schema to contain known scripts
 

# Comments

PR #2366, opened 2020-12-02
Demonstrating script address discovery and expose shared scripts in ApiWallet

paweljakubas:
# Issue Number

https://jira.iohk.io/browse/ADP-525

# Overview

- [x] Extend ApiWallet by sharedScripts
- [x] Updating testing layer of Api.Types
- [x] introducing `Metadata` having `TxMetadata` and `Script` and update code base
- [x] update tests wrt the above change
- [x] enable sending script via auxliaryScript in TxBody
- [x] discover scripts from SL.Tx in fromAllegra
- [x] use isShared in Model
- [x] add integration tests showing that the script with our verification keys is later discovered and exposed in sharedScripts in ApiWallet
- [ ] add more integration test with to wallets and script containing both verification keys
- [ ] deal with db migration

Prerequisite for using isShared:
- [x] PR merged -> https://github.com/input-output-hk/cardano-wallet/pull/2308 

# Comments


specifications/api/swagger.yaml
```diff
@@ -251,6 +251,13 @@ x-amount: &amount
       enum:
         - lovelace
 
+x-scriptStatus: &scriptStatus
+  type: string
+  enum:
+    - awaits_signing
+    - spent
+    - staked
```
KtorZ: I am not sure to understand the rationale behind those statuses? From our point of view, a script address exists on the blockchain or it doesn't. It'd be valuable however to show the balance available at that script address.

specifications/api/swagger.yaml
```diff
@@ -1409,12 +1416,37 @@ components:
...
+    ApiSharedScript: &ApiSharedScript
+      description: Discovered script that contains our keys
+      type: object
+      required:
+        - script_address
+        - our_keys
+        - funds_status
+        - funds_available
+      properties:
+        script_address: *AnyAddress
```
KtorZ: Seems like there's some excessive wrapping here. The resulting JSON will look like `{ "script_address": { "address": "..." } }`

What about simply showing it as `"address": ...` ?

specifications/api/swagger.yaml
```diff
@@ -1409,12 +1416,37 @@ components:
...
+        our_keys:
+          type: array
+          minItems: 1
+          items: *ApiVerificationKey
+        funds_status: *scriptStatus
+        funds_available: *amount
+
```
KtorZ, line 1440: I think it'll be valuable to show a JSON representation of the whole script no?

specifications/api/swagger.yaml
paweljakubas: I expect to have the following lifecycle of addresses with script:
1. some funds are sent to this address
2. even more money can be sent
3. tx with script with all required signatures is sent
 and depending on where script is embedded, we have money spent or money staked
1-2 -> script awaits signing but available balance can be different
3 -> funds are unlocked and are spent or staked

lib/core/test/data/Cardano/Wallet/Api/ApiWallet.json
```diff
@@ -1,476 +1,15262 @@
 {
-    "seed": 287889657345073705,
+    "seed": -7459132351643344646,
```
KtorZ: That file is _way_ too large :grimacing:

paweljakubas:
closing in favor of https://github.com/input-output-hk/cardano-wallet/pull/2675

PR #2367, opened 2020-12-03
Improve JSON errors in StakePoolMetadata

hasufell:
This was brought to my attention by Carlos Lopezdelara.
A pool didn't show up in daedalus, because the wallet
rejects the metadata, because the description is too long (exceeds 250 chars).

Logs were:

```
  [cardano-wallet.pools-engine:Info:38] [2020-12-02 03:08:00.64 UTC]
  Fetching metadata with hash 54e76fd5 from
  https://smash.cardano-mainnet.iohk.io/api/v1/metadata/c7013e0a9e0f04a0363998cac5db01f08aa466b755a1a91c42b4c4b3/54e76fd5a3eb80cbb7c0996d42fca355344e7a5614462a5b887fbdccf3fe0990
  [cardano-wallet.pools-engine:warning:38] [2020-12-02 03:08:00.83 UTC]
  Failed to fetch metadata with hash 54e76fd5: Error in $: empty
```
Payload was:

```
  {
   "name": "BraveHeartStakePool",
    "description": "At BraveHeartStakePool we are committed in providing
    you, the Cardano community with the ability to have financial
    'FREEDOM'. We will strive to provide quality reliably service 24/7 &
    aim to maximise your return on stake. BraveHeartStakePool 'BRAVE'. ",
     "ticker": "BRAVE",
      "homepage": "https://braveheartstakepool.com",
       "extended": "https://braveheartstakepool.com/brave-extended.json"
       }
  }
```
----

I think there are some things to discuss here:

1. where do these limitations come from?
2. why are these limitations not in SMASH?
3. how to improve this? The failure is way too silent. Even the updated json error isn't enough imo.

### DIFF
--- a/lib/core/src/Cardano/Pool/Metadata.hs
+++ b/lib/core/src/Cardano/Pool/Metadata.hs
@@ -335,9 +335,8 @@ fetchFromRemote tr builders manager pid url hash = runExceptTLog $ do
             -- Metadata are _supposed to_ be made of:
             --
             -- - A name (at most 50 UTF-8 bytes)
-            -- - An optional description (at most 250 UTF-8 bytes)
-            -- - A ticker (at most 5 UTF-8 bytes)
-            -- - A homepage (at most 100 UTF-8 bytes)
+            -- - An optional description (at most 255 UTF-8 bytes)
+            -- - A ticker (between 3 and 5 UTF-8 bytes)
             --
             -- So, the total, including a pretty JSON encoding with newlines ought
             -- to be less than 512 bytes. For security reasons, we only download the
--- a/lib/core/src/Cardano/Pool/Metadata.hs
+++ b/lib/core/src/Cardano/Pool/Metadata.hs
@@ -290,6 +290,8 @@ fetchFromRemote
     -> IO (Maybe StakePoolMetadata)
 fetchFromRemote tr builders manager pid url hash = runExceptTLog $ do
     chunk <- getChunk `fromFirst` builders
+    when (BS.length chunk > 512) $ throwE
+        "Metadata exceeds max length of 512 bytes"
     when (blake2b256 chunk /= coerce hash) $ throwE $ mconcat
         [ "Metadata hash mismatch. Saw: "
         , B8.unpack $ hex $ blake2b256 chunk
@@ -339,12 +341,13 @@ fetchFromRemote tr builders manager pid url hash = runExceptTLog $ do
             -- - A ticker (between 3 and 5 UTF-8 bytes)
             --
             -- So, the total, including a pretty JSON encoding with newlines ought
-            -- to be less than 512 bytes. For security reasons, we only download the
-            -- first 512 bytes.
+            -- to be less than or equal to 512 bytes. For security reasons, we only
+            -- download the first 513 bytes and check the length at the
+            -- call-site.
             case responseStatus res of
                 s | s == status200 -> do
                     let body = responseBody res
-                    Right . Just . BL.toStrict <$> brReadSome body 512
+                    Right . Just . BL.toStrict <$> brReadSome body 513
 
                 s | s == status404 -> do
                     pure $ Left "There's no known metadata for this pool."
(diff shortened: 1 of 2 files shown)


hasufell:
@KtorZ @ksaric

ksaric:
1. where do these limitations come from? 

From the spec of the metadata. See more info here - https://docs.cardano.org/projects/smash/en/latest/getting-started/how-to-install-smash.html#metadata

2. why are these limitations not in SMASH?

They are, but the length of the description is currently 255.

3. how to improve this? The failure is way too silent. Even the updated json error isn't enough imo.

You could reuse the validation from SMASH/node.

KtorZ:
@hasufell , also good to know cardano-wallet only reads the first `512` bytes when fetching metadata (as a security measure), and then tries to decode it as JSON. So if the JSON is longer than that, we end up with a chunk of garbage data that can't be decoded and a cryptic JSON error. See:

https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Pool/Metadata.hs#L326-L357

hasufell:
1. spec: https://hydra.iohk.io/build/790053/download/1/delegation_design_spec.pdf
2. smash: https://github.com/input-output-hk/smash/blob/142c6ab8d2fc6533fb65e1d5df9af445ca989622/smash-servant-types/src/Cardano/SMASH/Types.hs#L210
3. cardano-cli: https://github.com/input-output-hk/cardano-node/blob/2c1b9ade7306713a7e74a85c9d05c022c986c865/cardano-api/src/Cardano/Api/StakePoolMetadata.hs#L99

I don't see that the spec specifies the homepage length explicitly.

lib/core/src/Cardano/Pool/Metadata.hs
```diff
@@ -290,6 +290,8 @@ fetchFromRemote
     -> IO (Maybe StakePoolMetadata)
 fetchFromRemote tr builders manager pid url hash = runExceptTLog $ do
     chunk <- getChunk `fromFirst` builders
+    when (BS.length chunk >= 513) $ throwE
+        "Metadata exceeds max length of 512 bytes"
```
KtorZ, line 294: Why 513 ?

hasufell: Because 512 is the max allowed length. By peeking one byte further, we're able to determine whether this property holds.

If you mean to use `(BS.length chunk > 512)` instead, then sure, also can.

KtorZ: > If you mean to use (BS.length chunk > 512) instead, then sure, also can.

Well, it's not much this one that bothers me (although `> 512` reflects better the actual property than `>= 513`), but it's the `brReadSome body 513` below.

hasufell: Yes, that is on purpose. We no longer cut off the body, but fail for anything that's longer than `512` bytes, so we must *at least* read 513 bytes. If the body is 512 bytes, it will only be 512, there should be no padding.

KtorZ: > fail for anything that's longer than 512 bytes, so we must at least read 513 bytes. 

I do not understand how you draw this conclusion from this hypothesis? :no_mouth:

KtorZ: clarified in DM. My bad, we indeed want to know whether the body is longer than 512 bytes and can't do so if only reading up to 512 bytes :+1:

KtorZ: Note that I still think that this inequality would be better written as `BS.length chunk > 512` to reflect the actual predicate.

KtorZ (approved):

KtorZ:
:point_up: sorry, that's my bad. Currently fixing my mess in #2378  ...

rvl:
Sorry, I accidentally disabled the cardano-wallet bors jobset in Hydra.

bors r+

PR #2370, opened 2020-12-04
[ADP-500] Change addresses are not listed as soon as a transaction is no longer pending

hasufell:
# Issue Number

ADP-500

# Overview

- [x] created an integration test that checks the accepctance criteria and fails on the current branch
- [x] fixed the code in `knownAddresses` to also return all used change addresses

# Comments

## Questions

Do we need to communicate this to API users?

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE OverloadedLabels #-}
 {-# LANGUAGE QuasiQuotes #-}
@@ -32,6 +33,8 @@ import Cardano.Wallet.Primitive.Types.Address
     ( AddressState (..) )
 import Control.Monad
     ( forM, forM_ )
+import Control.Monad.IO.Class
+    ( liftIO )
 import Control.Monad.Trans.Resource
     ( runResourceT )
 import Data.Aeson
@@ -61,6 +64,7 @@ import Test.Integration.Framework.DSL
     , expectResponseCode
     , fixtureWallet
     , getFromResponse
+    , getWallet
     , json
     , listAddresses
     , minUTxOValue
@@ -77,6 +81,7 @@ import qualified Data.Aeson as Aeson
 import qualified Data.Aeson.Lens as Aeson
 import qualified Data.Text as T
 import qualified Network.HTTP.Types.Status as HTTP
+import qualified Test.Hspec as Hspec
 import qualified Test.Hspec.Expectations.Lifted as Expectations
 
 spec :: forall n.
@@ -252,6 +257,85 @@ spec = describe "SHELLEY_ADDRESSES" $ do
             , expectListField 0 (Aeson.key "id" . Aeson._String)
                 (`shouldNotSatisfy` T.isPrefixOf "addr_test")
             ]
+
+    Hspec.it "ADDRESS_LIST_06 - Used change addresses are listed after a transaction is no longer pending" $ \ctx -> runResourceT @IO $ do
+        let initPoolGap = 10
+            toDestVal = 10 * minUTxOValue
+            backToSrcVal = fmap (1 * minUTxOValue +)
+
+        let verifyAddrs i p addr = do
+                liftIO (length (filter ((== Used) . (^. (#state . #getApiT))) addr)
+                    `shouldBe` p)
+                liftIO (length addr `shouldBe` i)
+
+        -- 1. create shelley wallet
+        wSrc <- fixtureWallet ctx
+        wDest <- emptyWalletWith ctx ("Wallet", "cardano-wallet", initPoolGap)
+
+        -- 2. list addresses of that wallet (for later comparison)
+        listAddresses @n ctx wSrc >>= verifyAddrs 30 10
+
+        -- 3. send a transaction from that wallet
+        addrs <- listAddresses @n ctx wDest
+        verifyAddrs 10 0 addrs
+        let destination = (head addrs) ^. #id
+        let payload = Json [json|{
+                "payments": [{
+                    "address": #{destination},
+                    "amount": {
+                        "quantity": #{toDestVal},
+                        "unit": "lovelace"
+                    }
+                }],
+                "passphrase": "cardano-wallet"
+            }|]
+
+        request @(ApiTransaction n) ctx
+            (Link.createTransaction @'Shelley wSrc) Default payload
+            >>= flip verify
+                [ expectResponseCode HTTP.status202 ]
+
+        -- 4. wait for transaction to be inserted
+        eventually "Wallet balance = minUTxOValue" $ do
+            rb <- request @ApiWallet ctx
+                (Link.getWallet @'Shelley wDest) Default Empty
+            expectField
+                (#balance . #getApiT . #available)
+                (`shouldBe` Quantity toDestVal)
+                rb
+
+        -- 5. Check that change address is still there as unused
+        -- and the gap moved.
+        listAddresses @n ctx wSrc >>= verifyAddrs 31 11
+
+        -- 6. send another transaction from the destination wallet
+        listAddresses @n ctx wDest >>= verifyAddrs 11 1
+        addrs' <- listAddresses @n ctx wSrc
+        let destination' = (addrs' !! 5) ^. #id
+        let payload' = Json [json|{
+                "payments": [{
+                    "address": #{destination'},
+                    "amount": {
+                        "quantity": #{minUTxOValue},
+                        "unit": "lovelace"
+                    }
+                }],
+                "passphrase": "cardano-wallet"
+            }|]
+        currentSrcWallet <- getWallet ctx wSrc
+        let currentSrcWalletBalance = currentSrcWallet ^. (#balance . #getApiT . #available)
+        request @(ApiTransaction n) ctx
+            (Link.createTransaction @'Shelley wDest) Default payload'
+            >>= flip verify
+                [ expectResponseCode HTTP.status202 ]
+        eventually "Wallet balance = minUTxOValue + current" $ do
+            rb <- request @ApiWallet ctx
+                (Link.getWallet @'Shelley wSrc) Default Empty
+            expectField
+                (#balance . #getApiT . #available)
+                (`shouldBe` backToSrcVal currentSrcWalletBalance)
+                rb
+        listAddresses @n ctx wDest >>= verifyAddrs 12 2
 
     it "ADDRESS_INSPECT_01 - Address inspect OK" $ \ctx -> do
         let str = "Ae2tdPwUPEYz6ExfbWubiXPB6daUuhJxikMEb4eXRp5oKZBKZwrbJ2k7EZe"
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
@@ -17,6 +17,7 @@ import Prelude
 import Cardano.Wallet.Api.Types
     ( AnyAddress
     , ApiAddress
+    , ApiT (..)
     , ApiTransaction
     , ApiVerificationKey
     , ApiWallet
@@ -31,6 +32,8 @@ import Cardano.Wallet.Primitive.AddressDiscovery.Sequential
     ( defaultAddressPoolGap, getAddressPoolGap )
 import Cardano.Wallet.Primitive.Types.Address
     ( AddressState (..) )
+import Cardano.Wallet.Primitive.Types.Tx
+    ( TxStatus (..) )
 import Control.Monad
     ( forM, forM_ )
 import Control.Monad.IO.Class
@@ -40,7 +43,7 @@ import Control.Monad.Trans.Resource
 import Data.Aeson
     ( ToJSON (..), object, (.=) )
 import Data.Generics.Internal.VL.Lens
-    ( (^.) )
+    ( view, (^.) )
 import Data.Quantity
     ( Quantity (..) )
 import Data.Text
@@ -62,9 +65,9 @@ import Test.Integration.Framework.DSL
     , expectListField
     , expectListSize
     , expectResponseCode
+    , fixturePassphrase
     , fixtureWallet
     , getFromResponse
-    , getWallet
     , json
     , listAddresses
     , minUTxOValue
@@ -81,7 +84,6 @@ import qualified Data.Aeson as Aeson
 import qualified Data.Aeson.Lens as Aeson
 import qualified Data.Text as T
 import qualified Network.HTTP.Types.Status as HTTP
-import qualified Test.Hspec as Hspec
 import qualified Test.Hspec.Expectations.Lifted as Expectations
 
 spec :: forall n.
@@ -110,7 +112,7 @@ spec = describe "SHELLEY_ADDRESSES" $ do
 
     it "ADDRESS_LIST_01 - Can list addresses with non-default pool gap" $ \ctx -> runResourceT $ do
         let g = 15
-        w <- emptyWalletWith ctx ("Wallet", "cardano-wallet", g)
+        w <- emptyWalletWith ctx ("Wallet", fixturePassphrase, g)
         r <- request @[ApiAddress n] ctx
             (Link.listAddresses @'Shelley w) Default Empty
         expectResponseCode HTTP.status200 r
@@ -182,7 +184,7 @@ spec = describe "SHELLEY_ADDRESSES" $ do
     it "ADDRESS_LIST_03 - Generates new address pool gap" $ \ctx -> runResourceT $ do
         let initPoolGap = 10
         wSrc <- fixtureWallet ctx
-        wDest <- emptyWalletWith ctx ("Wallet", "cardano-wallet", initPoolGap)
+        wDest <- emptyWalletWith ctx ("Wallet", fixturePassphrase, initPoolGap)
 
         -- make sure all addresses in address_pool_gap are 'Unused'
         r <- request @[ApiAddress n] ctx
@@ -206,7 +208,7 @@ spec = describe "SHELLEY_ADDRESSES" $ do
                             "unit": "lovelace"
                         }
                     }],
-                    "passphrase": "cardano-wallet"
+                    "passphrase": #{fixturePassphrase}
                 }|]
 
             rTrans <- request @(ApiTransaction n) ctx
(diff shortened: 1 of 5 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
```diff
@@ -253,6 +255,59 @@ spec = describe "SHELLEY_ADDRESSES" $ do
...
+            }|]
+
+        _rTrans <- request @(ApiTransaction n) ctx
+            (Link.createTransaction @'Shelley wSrc) Default payload
+
+        -- 4. quickly, list addresses before the transaction is pending.
+        -- See that a new change address is listed there
+        addrsChange <- request @[ApiAddress n] ctx
+            (Link.listAddresses @'Shelley wSrc) Default Empty
+        verify addrsChange
```
paweljakubas: checking the number of addresses is one thing, another is to check address status of those addresses. We need to look at `Used/Unused` here and compare later

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
```diff
@@ -253,6 +255,59 @@ spec = describe "SHELLEY_ADDRESSES" $ do
...
+                (#balance . #getApiT . #available)
+                (`shouldBe` Quantity (1 * 1_000_000))
+                rb
+
+        -- 6. Check that change address is still there
+        addrsChange' <- request @[ApiAddress n] ctx
+            (Link.listAddresses @'Shelley wSrc) Default Empty
+        verify addrsChange'
+            [ expectResponseCode HTTP.status200
+            , expectListSize 31
```
paweljakubas: Remember, that when the address is used, address pool gap invariant kicks in. We need to have consecutive  `gap` number of Unused addresses. So it is better to observe the addresses engaged and make sure when Pending -> InLedger transition happens they address status evaluates as expected.

paweljakubas:
General idea: as the tx goes from Pending -> InLedger  we can probably investigate three situations:
(a)  we are receiving money and in both Pending/InLedger we should see our addresses (UTxOExternal flipped to Used)
(b) we are sending money to someone and in both Pending/InLedger we should see our addresses (UTxOInternal and UTxOExternal flipped to Used)
(c) hybrid case

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
```diff
@@ -253,6 +255,59 @@ spec = describe "SHELLEY_ADDRESSES" $ do
...
+        _rTrans <- request @(ApiTransaction n) ctx
+            (Link.createTransaction @'Shelley wSrc) Default payload
+
+        -- 4. quickly, list addresses before the transaction is pending.
+        -- See that a new change address is listed there
+        addrsChange <- request @[ApiAddress n] ctx
+            (Link.listAddresses @'Shelley wSrc) Default Empty
+        verify addrsChange
+            [ expectResponseCode HTTP.status200
+            , expectListSize 31
```
piotr-iohk: I'm guessing that 31 is because `fixtureWallet` has address pool gap = 20 and there are funds on first 10 addresses, so initially there was 30 addresses on the wallet, so the change address is generated as next one 31st. Perhaps a comment would be handy explaining this.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
```diff
@@ -253,6 +255,59 @@ spec = describe "SHELLEY_ADDRESSES" $ do
...
+            ]
+
+        -- 5. wait for transaction to be inserted
+        eventually "Wallet balance = initPoolGap * minUTxOValue" $ do
```
piotr-iohk: I think this should be:
```suggestion
        eventually "Wallet balance = minUTxOValue" $ do
```

initPoolGap has nothing to do here I guess.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
```diff
@@ -253,6 +255,59 @@ spec = describe "SHELLEY_ADDRESSES" $ do
...
+            rb <- request @ApiWallet ctx
+                (Link.getWallet @'Shelley wDest) Default Empty
+            expectField
+                (#balance . #getApiT . #available)
+                (`shouldBe` Quantity (1 * 1_000_000))
```
piotr-iohk:
```suggestion
                (`shouldBe` Quantity minUTxOValue)
```

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
```diff
@@ -794,19 +794,14 @@ instance
...
-                fromEnum . getAddressPoolGap . gap . internalPool $ s
-            discardUndiscoveredChange xs =
-                take (length ixs) $ drop (length xs - internalGap) xs
-            changeAddresses =
-                discardUndiscoveredChange $
-                    addresses (liftPaymentAddress @n @k) (internalPool s)
-            nonChangeAddresses =
+            -- instead of only showing as many unused change addresses as there
+            -- are pending transactions, we just show them all
+            visibleChangeAddresses = addresses (liftPaymentAddress @n @k) (internalPool s)
```
hasufell: The question is... does this break any assumptions further down the line? Is the length of `visibleNonChangeAddresses` fixed? Do we want to add info about internal vs external? Questions over questions.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
```diff
@@ -253,6 +256,54 @@ spec = describe "SHELLEY_ADDRESSES" $ do
...
+                (Link.getWallet @'Shelley wDest) Default Empty
+            expectField
+                (#balance . #getApiT . #available)
+                (`shouldBe` Quantity (1 * minUTxOValue))
+                rb
+
+        -- 5. Check that change address is still there as unused
+        -- and the gap moved.
+        addrsAfter <- listAddresses @n ctx wSrc
+        verifyAddrs addrsAfter 51 11
```
hasufell: Looks good. Prior to the transaction, this was `50` (total) and `10` (used).

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
```diff
@@ -253,6 +257,83 @@ spec = describe "SHELLEY_ADDRESSES" $ do
...
+            >>= flip verify
+                [ expectResponseCode HTTP.status202 ]
+        eventually "Wallet balance = minUTxOValue + current" $ do
+            rb <- request @ApiWallet ctx
+                (Link.getWallet @'Shelley wSrc) Default Empty
+            expectField
+                (#balance . #getApiT . #available)
+                (`shouldBe` fmap (1 * minUTxOValue +) currentSrcWalletBalance)
+                rb
+        listAddresses @n ctx wDest >>= verifyAddrs 22 2
```
hasufell: @paweljakubas this doesn't hold true, even after the patch to `knownAddresses`. We send another transaction from the destination wallet back to the source wallet. We should see one more unused address here, or is there simply no change?

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
```diff
@@ -253,6 +257,83 @@ spec = describe "SHELLEY_ADDRESSES" $ do
...
+        -- 1. create shelley wallet
+        wSrc <- fixtureWallet ctx
+        wDest <- emptyWalletWith ctx ("Wallet", "cardano-wallet", initPoolGap)
+
+        -- 2. list addresses of that wallet (for later comparison)
+        listAddresses @n ctx wSrc >>= verifyAddrs 50 10
+
+        -- 3. send a transaction from that wallet
+        addrs <- listAddresses @n ctx wDest
+        verifyAddrs 20 0 addrs
```
paweljakubas: that's what I expect given `initPoolGap` ie., 10 Unused change addresses and 10 Unused external addresses

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
```diff
@@ -253,6 +257,83 @@ spec = describe "SHELLEY_ADDRESSES" $ do
...
+                liftIO (length (filter ((== Used) . (^. (#state . #getApiT))) addr)
+                    `shouldBe` p)
+                liftIO (length addr `shouldBe` i)
+
+        -- 1. create shelley wallet
+        wSrc <- fixtureWallet ctx
+        wDest <- emptyWalletWith ctx ("Wallet", "cardano-wallet", initPoolGap)
+
+        -- 2. list addresses of that wallet (for later comparison)
+        listAddresses @n ctx wSrc >>= verifyAddrs 50 10
```
paweljakubas: I do not think that this one is needed

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
```diff
@@ -253,6 +257,83 @@ spec = describe "SHELLEY_ADDRESSES" $ do
...
+            rb <- request @ApiWallet ctx
+                (Link.getWallet @'Shelley wDest) Default Empty
+            expectField
+                (#balance . #getApiT . #available)
+                (`shouldBe` Quantity (2 * minUTxOValue))
+                rb
+
+        -- 5. Check that change address is still there as unused
+        -- and the gap moved.
+        listAddresses @n ctx wSrc >>= verifyAddrs 51 11
```
paweljakubas: this one is not needed - it just makes this test more sophisticated

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
```diff
@@ -253,6 +257,83 @@ spec = describe "SHELLEY_ADDRESSES" $ do
...
+
+        -- 6. send another transaction from the destination wallet
+        listAddresses @n ctx wDest >>= verifyAddrs 21 1
```
paweljakubas: this is what should be expected after money was sent to  address with index = 0. This should be address from external pool. After receiving funds this address should have flipped to Used. And because we need `initAddressGap=10` so one Unused address extended external address pool. internal (aka change) address pool stays intact, ie. 10 Unused addresses

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
paweljakubas: after sending from wDest we should have external address pool like before it, ie. Used address at index =0, and Unused from 1 up to 10.So 11 addresses in the pool. Now, one of internal addresses is Used. the question is which one. If it is the first one in the internal pool, then the above should be correct expectation, ie. we should have after two pools `<>` 
```
i=0 Used -- beginning of external pool
i=1 Unused
i=2 Unused
...
i=10 Unused -- end of external pool
i=11 Used -- beginning of internal pool
i=12 Unused 
...
i=21 Unused -- end of internal pool
```
It is not like that? maybe it is worth looking at seq state and make sure

paweljakubas:
@hasufell What are the values `sizeOfTwoPools`  and `numberOfUsedAddresses` in
```
listAddresses @n ctx wDest >>= verifyAddrs sizeOfTwoPools numberOfUsedAddresses
```
AFTER sending money back from wDest to wSrc 
(a) before applying your fix
(b) after applying your fix?

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
hasufell: Why not? I'd like to compare before the transaction and after the transaction.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
hasufell: I also don't understand why not. This was the original test case that proves the change address shows up. Test number 6. is an optional addition of sending funds back to the source wallet.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
hasufell: I think there's just no change generated... I changed the money that is initially sent from `wSrc` to `wDest` to `(3 * minUTxOValue)` and suddenly the test works.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
KtorZ: We aren't supposed to show change addresses until they've been used in a transaction... They are internal and not for users to use.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
paweljakubas: to me this is duplicated testing. Listing addresses in `wDest` already inspects this.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
```diff
@@ -253,6 +257,83 @@ spec = describe "SHELLEY_ADDRESSES" $ do
...
+
+        -- 3. send a transaction from that wallet
+        addrs <- listAddresses @n ctx wDest
+        verifyAddrs 20 0 addrs
+        let destination = (addrs !! 0) ^. #id
+        let payload = Json [json|{
+                "payments": [{
+                    "address": #{destination},
+                    "amount": {
+                        "quantity": #{2 * minUTxOValue},
```
paweljakubas: maybe it is good to send 10*minUTxOValue as we risk the situation when sending back from wDest that we can have out+fee+change where change is less than minUTxOValue and hence it is not visible in listAddresses. outputs and changes are required to be > minUTxOValue!

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
paweljakubas: you already show the case when listing addresses for wDest. So for me it is kinda duplication.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
paweljakubas: yes! I said in the comment above. both output and change NEEDS to be > minUTxOValue and to ensure that let's have bigger input in wDest. So that input - fee = change > minUTxOValue

paweljakubas: see https://github.com/input-output-hk/cardano-wallet/pull/2370#discussion_r539154549

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
hasufell: Done

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
hasufell: Done

hasufell:
> @hasufell What are the values `sizeOfTwoPools` and `numberOfUsedAddresses` in
> 
> ```
> listAddresses @n ctx wDest >>= verifyAddrs sizeOfTwoPools numberOfUsedAddresses
> ```
> 
> AFTER sending money back from wDest to wSrc
> (a) before applying your fix
> (b) after applying your fix?

```
Before fix:


=== src before transaction ===
expected sizeOfTwoPools: 30
expected numberOfUsedAddresses: 10
actual sizeOfTwoPools: 30
actual numberOfUsedAddresses: 10

=== dst before transaction ===
expected sizeOfTwoPools: 10
expected numberOfUsedAddresses: 0
actual sizeOfTwoPools: 10
actual numberOfUsedAddresses: 0

=== src after transaction ===
expected sizeOfTwoPools: 31
expected numberOfUsedAddresses: 11
actual sizeOfTwoPools: 30
actual numberOfUsedAddresses: 10

=== dst before second transaction ===
expected sizeOfTwoPools: 11
expected numberOfUsedAddresses: 1
actual sizeOfTwoPools: 11
actual numberOfUsedAddresses: 1

=== dst after second transaction ===
expected sizeOfTwoPools: 12
expected numberOfUsedAddresses: 2
actual sizeOfTwoPools: 11
actual numberOfUsedAddresses: 1

After fix:

=== src before transaction ===
expected sizeOfTwoPools: 30
expected numberOfUsedAddresses: 10
actual sizeOfTwoPools: 30
actual numberOfUsedAddresses: 10

=== dst before transaction ===
expected sizeOfTwoPools: 10
expected numberOfUsedAddresses: 0
actual sizeOfTwoPools: 10
actual numberOfUsedAddresses: 0

=== src after transaction ===
expected sizeOfTwoPools: 31
expected numberOfUsedAddresses: 11
actual sizeOfTwoPools: 31
actual numberOfUsedAddresses: 11

=== dst before second transaction ===
expected sizeOfTwoPools: 11
expected numberOfUsedAddresses: 1
actual sizeOfTwoPools: 11
actual numberOfUsedAddresses: 1

=== dst after second transaction ===
expected sizeOfTwoPools: 12
expected numberOfUsedAddresses: 2
actual sizeOfTwoPools: 12
actual numberOfUsedAddresses: 2

```

paweljakubas (approved):
The integration testing persuades me, applied fix also. Would be good also for @KtorZ to double check the fix. Also you need to soothe hlint and stylish

hasufell:
Note that the following property test currently fails:

https://github.com/input-output-hk/cardano-wallet/blob/1a2852fa9fe75e6e031f3069144f85d71e3c8e46/lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/SequentialSpec.hs#L473-L497

The reason being that we don't show *unused change addresses* anymore if there are *pending* transactions.

We have a few options here:

1. remove that part of the property test... because in order to see the change address after a `genChange`, we need to also send it to the chain and wait, afais. So this can only be covered by the integration test.
2. we re-add those unused change addresses to the output of `knownAddresses`... this depends on the acceptance criteria

KtorZ:
@hasufell https://github.com/input-output-hk/cardano-wallet/pull/2370#issuecomment-741875786

:point_right: (2)

We need to show those addresses, doing otherwise is actually a breaking change and quite confusing (we may otherwise end up with transactions that reference no known addresses -- from one client's perspective).

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/SequentialSpec.hs
```diff
@@ -488,9 +488,13 @@ prop_changeIsOnlyKnownAfterGeneration (intPool, extPool) =
         ]
   where
     prop_addrsNotInInternalPool addrs =
-        map (\x -> (ShowFmt x, isNothing $ fst $ lookupAddress id x intPool)) addrs
+        map (\(x, s) ->
```
hasufell, line 493: So now `prop_addrsNotInInternalPool` doesn't really hold, because we have *used known internal addresses* in the return of `knownAddresses`. That's expected afais (that was the point of this PR, no?), so I adjusted that expectation.

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/SequentialSpec.hs
```diff
@@ -488,9 +488,13 @@ prop_changeIsOnlyKnownAfterGeneration (intPool, extPool) =
...
+                let notInPool = isNothing $ fst $ lookupAddress id x intPool
+                    isUsed = s == Used
+                in (ShowFmt x, notInPool || isUsed))
```
KtorZ, line 496: :+1:

piotr-iohk:
I was thinking, are there potential performance implications of this change?

Like every new tx will "produce" new used address to be shown on a source wallet. I was wondering if this might affect wallets doing massive number of outgoing transactions (like exchanges) in the long run...

Maybe not though, listing addresses does not seem to be very costly... (http://cardano-wallet-benchmarks.herokuapp.com/nightbuilds/767)

@KtorZ @hasufell @paweljakubas anyway, maybe you have thoughts about it?

paweljakubas:
@piotr-iohk I *tentatively* incline to think that rather no significant performance cost is to be incurred by those changes. But not 100% sure with those things

piotr-iohk:
I have triggered benchmark on branch -> https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/768. I'm not sure if it'll show anything, maybe if there is a potential regression on transaction time itself.

piotr-iohk:
I have also made this observation. Doing transaction from the wallet (Wallet 1) to 10 external addresses (on Wallet 2), produced 6 change addresses on src Wallet 1.
Is this expected or is it rather every transaction should produce single change address? (As far as I see multi-adress txs not always produce many change addresses, sometimes only 1, not sure what's the factor for that, utxo maybe?)

Note: the tx was rather big -> 69 inputs and 16 outputs (so yeah, 10 outputs to external addresses and 6 change addresses).

Wallet 1 before transaction to 10 addresses:
```
Addresses:
  Total: 119
    Used: 59
    Unused: 60
    
{
  "distribution": {
    "10000000000000000": 0,
    "100000000000000": 0,
    "1000000000000": 0,
    "10000000000": 0,
    "100000000": 7,
    "1000000": 149,
    "10000": 0,
    "100": 0,
    "1000000000": 2,
    "100000000000": 0,
    "10": 0,
    "45000000000000000": 0,
    "1000": 0,
    "10000000000000": 0,
    "1000000000000000": 0,
    "100000": 0,
    "10000000": 18
  },
  "scale": "log10",
  "total": {
    "quantity": 766672855,
    "unit": "lovelace"
  }
}

```

Wallet 1 after transaction:

```
Addresses:
  Total: 125
    Used: 65
    Unused: 60
    
{
  "distribution": {
    "10000000000000000": 0,
    "100000000000000": 0,
    "1000000000000": 0,
    "10000000000": 0,
    "100000000": 9,
    "1000000": 88,
    "10000": 0,
    "100": 0,
    "1000000000": 2,
    "100000000000": 0,
    "10": 0,
    "45000000000000000": 0,
    "1000": 0,
    "10000000000000": 0,
    "1000000000000000": 0,
    "100000": 0,
    "10000000": 14
  },
  "scale": "log10",
  "total": {
    "quantity": 703595526,
    "unit": "lovelace"
  }
}

```

piotr-iohk:
Is this change intended only for Shelley wallets? Icarus and Byron don't seem to be affected atm.

piotr-iohk:
> I have triggered benchmark on branch -> https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/768. I'm not sure if it'll show anything, maybe if there is a potential regression on transaction time itself.

I have imported this run into http://cardano-wallet-benchmarks.herokuapp.com/nightbuilds/768 and analyzed a bit. 

I have noticed some _potential_ regression on listing tx time on mainnet restorations benchmark (build 768):

Screenshot from 2020-12-14 13-50-54
Screenshot from 2020-12-14 13-52-19

But not sure if it is conclusive. It is not present accross all benchmarks, just the two above, so might be just outstanding measurement. (see [other mainnet measurements](http://cardano-wallet-benchmarks.herokuapp.com/mainnet-restoration?type_of_wallet=all&type_of_measurement=list_transactions_time), [latency measurements](http://cardano-wallet-benchmarks.herokuapp.com/latency?latency_category=4&latency_benchmark=all&latency_measurement=listTransactions))

There is no post transaction benchmark at the moment, only [post tx fee](http://cardano-wallet-benchmarks.herokuapp.com/latency?latency_category=4&latency_benchmark=all&latency_measurement=postTransactionFee) where I don't see any regressions.

piotr-iohk:
> There is no post transaction benchmark at the moment, only post tx fee where I don't see any regressions.

I have added postTx benchmarks here -> https://github.com/input-output-hk/cardano-wallet/pull/2396. It'll be interesting to see the results on this branch as soon as the benchmark pr is merged.

piotr-iohk:
Actually I have already postTransaction and postTransactionTo5Addrs benchmark results for current master -> https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/776.

I can cherry-pick the commit with new benchmarks, run it on this branch and we can compare results.

Benchmark results with postTx for this branch should be here -> https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/779 :heavy_check_mark: 

EDIT:
I don't see any deterioration in postTx performance between master(776) and this branch(779):
[postTx](http://cardano-wallet-benchmarks.herokuapp.com/latency?latency_category=4&latency_benchmark=all&latency_measurement=postTransaction)
[postTxTo5Addrs](http://cardano-wallet-benchmarks.herokuapp.com/latency?latency_category=4&latency_benchmark=all&latency_measurement=postTransTo5Addrs)

hasufell:
Can this be merged @piotr-iohk @KtorZ

KtorZ (approved):

KtorZ:
bors merge

KtorZ:
bors retry

KtorZ:
bors retry

KtorZ:
bors retry

hasufell:
bors retry

hasufell:
bors retry

hasufell:
bors retry

PR #2374, opened 2020-12-07
Get rid of the 'target' parameterization

KtorZ:
# Issue Number

[ADP-548](https://jira.iohk.io/browse/ADP-548)

# Overview

- [x] I have gotten rid of the `target` type parameter 

  Since there's no longer a need to distinguish between several backend,
  this has become redundant and an annoying unnecessary level of
  indirection. The parameter was present on the NetworkLayer and the
  TransactionLayer to allow tweaking either, some behavior or some
  data-type specific to each backend. In particular, since the
  'validateSelection' function was only truly needed for Jormungandr
  which had some special rules about coin selections, I got rid of it
  entirely with the parameter for it is pointless with cardano-node. I
  did however keep the 'Cursor' data-instance for now as it provides a
  nice mechanism for decoupling the network layer implementation from
  the interface declaration. Plus, if we undertake some heavier changes
  in the network layer, this may be entirely refactored anyway so
  there's no need to bother too much about it _yet_.

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/Network.hs
+++ b/lib/core/src/Cardano/Wallet/Network.hs
@@ -19,7 +19,6 @@ module Cardano.Wallet.Network
     , follow
     , FollowAction (..)
     , FollowExit (..)
-    , GetStakeDistribution
     , getSlottingParametersForTip
 
     -- * Errors
@@ -60,7 +59,10 @@ import Cardano.Wallet.Primitive.Types
     , ProtocolParameters
     , SlotNo (..)
     , SlottingParameters (..)
+    , StakePoolsSummary
     )
+import Cardano.Wallet.Primitive.Types.Coin
+    ( Coin )
 import Cardano.Wallet.Primitive.Types.Hash
     ( Hash (..) )
 import Cardano.Wallet.Primitive.Types.RewardAccount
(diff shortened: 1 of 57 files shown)


paweljakubas (approved):
looks good and looks like very demanding in terms of editing task

lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -160,7 +162,11 @@ data NetworkLayer m target block = NetworkLayer
         -- ^ Broadcast a transaction to the chain producer
 
     , stakeDistribution
-        :: GetStakeDistribution target m
```
Anviking: 💯

lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -290,10 +294,10 @@ getSlottingParametersForTip nl = do
...
 -------------------------------------------------------------------------------}
 
--- | A cursor is local state kept by the chain consumer to use as the starting
--- position for 'nextBlocks'. The actual type is opaque and determined by the
--- backend @target@.
-data family Cursor target
+--- | A cursor is local state kept by the chain consumer to use as the starting
+--- position for 'nextBlocks'. The actual type is opaque and determined by the
+--- backend target.
+data family Cursor
```
Anviking: Could probably be simplified later to simply a opaque type rather than a type family, I take

rvl (approved):
Looks good - merge quick!

Anviking (approved):

lib/core/src/Cardano/Wallet/Network.hs
KtorZ: I initially tried as part of this PR ... BUT the decoupling it offers is kinda nice in the end. I wouldn't really like to pull in the actual definition of `Cursor` required from the cardano-node network layer:

```hs
data instance Cursor = Cursor
    (Async ())
    (Point (CardanoBlock StandardCrypto))
    (TQueue IO (ChainSyncCmd (CardanoBlock StandardCrypto) IO))
```

Also, I do think that one of our next step now that Jormungandr is gone is to review the architecture of the network layer entirely, in which case we may not even keep this concept of Cursor at all. So in the end, I left it as is.

Issue #2368, opened 2020-12-04
Flaky test: Network.Wai.Middleware.Logging, Logging Middleware, different request ids


jonathanknowles:
# Context
 
This failure was seen in the following overnight test runs of the entire `cardano-wallet-core:test:unit` test suite:

| Test<br>Date<br> | Wallet<br>Version<br> | Run<br>Count<br> | Failure<br>Count<br> | Failure<br> Rate<br> |
| -- | -- | --: | --: | --: | 
| 2020-12-04 | ec22c1ba | 48 | 1 | 2% |

# Test Case

`Network.Wai.Middleware.Logging, Logging Middleware, different request ids`

# Failure / Counter-example

```
test/unit/Network/Wai/Middleware/LoggingSpec.hs:225:13: 
1) Network.Wai.Middleware.Logging, Logging Middleware, different request ids
     Falsifiable (after 81 tests and 8 shrinks):
       NumberOfRequests 1
     expected: 1
      but got: 188

To rerun use: --match "/Network.Wai.Middleware.Logging/Logging Middleware/different request ids/"

Randomized with seed 1576720526
```

# Resolution 

---

# QA

Anviking:
Seen again https://github.com/input-output-hk/cardano-wallet/pull/2506#issuecomment-776931203

Anviking:
Test removed in  #2734

Issue #314, opened 2019-05-24
Bech32 encoder can be coerced into producing a string that can't be decoded


jonathanknowles:
## Context

When `encode` is applied to a `HumanReadablePart` containing one or more **upper-case** characters, it results in an **invalid** Bech32 string that cannot be decoded with the `decode` function.

## Steps to Reproduce
```hs
$ stack ghci bech32
> lower = fromRight undefined $ mkHumanReadablePart "test"
> upper = fromRight undefined $ mkHumanReadablePart "TEST"
```

## Expected behaviour
```hs
> encode lower mempty
Right "test12hrzfj"
> encode upper mempty
Right "test12hrzfj" -- SAME

# Should give the same output:
> decode $ fromRight undefined $ encode lower mempty
Right (HumanReadablePart "test","")
> decode $ fromRight undefined $ encode upper mempty
Right (HumanReadablePart "test","") -- SUCCESS
```

## Actual behaviour
```hs
# Should give the same output:
> encode lower mempty
Right "test12hrzfj"
> encode upper mempty
Right "test13jgcyw" -- DIFFERENT

# Should give the same output:
> decode $ fromRight undefined $ encode lower mempty
Right (HumanReadablePart "test","")
> decode $ fromRight undefined $ encode upper mempty
Left (StringToDecodeContainsInvalidChars []) -- FAILURE
```

## Cause

This appears to be caused by the [`encode`](https://github.com/input-output-hk/cardano-wallet/blob/master/lib/bech32/src/Codec/Binary/Bech32/Internal.hs#L158) function. Internally, it calculates a checksum based on the supplied `HumanReadablePart`, which can be upper or lower case (or mixed case). Unfortunately, the checksum calculation is case-sensitive. Later on, the human-readable part is converted to lower case before inclusion in the final output. If the original `HumanReadablePart` contained one or more upper case characters, the checksum will be inconsistent with the human readable part of the final output.

```hs
-- | Encode a human-readable string and data payload into a Bech32 string.      
encode :: HumanReadablePart -> ByteString -> Either EncodingError ByteString    
encode hrp@(HumanReadablePart hrpBytes) payload = do    
    let payload5 = toBase32 (BS.unpack payload)
    let payload' = payload5 ++ bech32CreateChecksum hrp payload5    
    let rest = map (word5ToChar Arr.!) payload'
    let output = B8.map toLower hrpBytes <> B8.pack "1" <> B8.pack rest   
[...]

PR #2377, opened 2020-12-08
Fix flaky WAI middleware logging test cases

rvl:
### Issue Number

Flaky test #2368

### Overview

- Adds a script for finding flaky tests.
- Fixes the test cases.

The "unique request ids" property was failing because it sometimes sees log messages from the previous iteration of the quickcheck property.

This happened much more on Windows, but I found I could also reproduce it with about 25-50 test runs on Linux.

I initially thought it was due to some buffering in iohk-monitoring. But that was wrong - `traceInTVarIO` is actually a really simple trace transformer (pretty cool actually!).

Then I thought perhaps what was happening is that the captured log list isn't fully evaluated before the test case finishes. But this was also wrong. We should probably make IOVars and log message types more strict, but this was not the bug.

The problem was in the test cases. They would complete one or more requests, then check what the server had logged. However, the server may not yet have completely finished its request handler at this time. Therefore, the "correct time measures" assertions sometimes failed due to missing LogResponse messages. Even worse, the "different request ids" property sometimes failed because logs from the previous property example were included in the logs of the current property example.

The solution is to judiciously place a sleep statement and filter logs from previous test runs.

### DIFF
--- a/lib/core/test/unit/Network/Wai/Middleware/LoggingSpec.hs
+++ b/lib/core/test/unit/Network/Wai/Middleware/LoggingSpec.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeOperators #-}
 
@@ -22,7 +23,7 @@ import Cardano.Wallet.Api.Server
 import Control.Concurrent
     ( threadDelay )
 import Control.Concurrent.Async
-    ( Async, async, cancel, mapConcurrently )
+    ( Async, async, cancel, mapConcurrently, replicateConcurrently_ )
 import Control.Concurrent.MVar
     ( newEmptyMVar, putMVar, readMVar )
 import Control.Concurrent.STM.TVar
@@ -102,11 +103,9 @@ import Servant.Server
 import Test.Hspec
     ( Spec, after, afterAll, beforeAll, describe, it, shouldBe, shouldContain )
 import Test.QuickCheck
-    ( Arbitrary (..), choose, property, withMaxSuccess )
+    ( Arbitrary (..), choose, counterexample, property, withMaxSuccess )
 import Test.QuickCheck.Monadic
-    ( monadicIO )
-import Test.Utils.Windows
-    ( pendingOnWindows, whenWindows )
+    ( assert, monadicIO, monitor )
 
 import qualified Data.Aeson as Aeson
 import qualified Data.List as L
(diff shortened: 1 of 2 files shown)


Anviking (approved):
> Adds an unsuccessful attempt to fix the test case.

So `clearLogs` didn't work?

I saw (locally) the failure once on master, and never so far on this branch.

scripts/fluke.sh seems like a great idea.

rvl:
`clearLogs` didn't work unfortunately @Anviking .

I suspect the failures could be due to unevaluated thunks.
So I made a change to reduce all ApiLogs to normal form before checking them.
This seems to have prevented test failures... so far (`fluke.sh` is up to repeat #314).
Reducing the ApiLog list to whnf was not enough to avoid test failures.

rvl:
I fixed it. It's a bit embarrassing because I went completely off track thinking that the failures were due to too-lazy evaluation. This was not the problem at all (although it won't hurt to tighten up evaluation a little).

lib/core/test/unit/Network/Wai/Middleware/LoggingSpec.hs
```diff
@@ -215,28 +214,39 @@ spec = describe "Logging Middleware"
...
+    it "correct time measures" $ \ctx -> withMaxSuccess 10 $
+        property $ \(NumberOfRequests n, RandomIndex i) -> monadicIO $ do
+            entries <- liftIO $ do
+                let reqs = mconcat
+                        [ replicate i (get ctx "/get")
+                        , [ get ctx "/long" ]
+                        , replicate (n - i) (get ctx "/get")
+                        ]
+                void $ mapConcurrently id reqs
+                waitForServerToComplete
```
KtorZ, line 241: Is this the only behavioral change which fixes the problem? (or am I missing something?)

lib/core/test/unit/Network/Wai/Middleware/LoggingSpec.hs
```diff
@@ -215,28 +214,39 @@ spec = describe "Logging Middleware"
...
 
-    it "different request ids" $ \ctx -> property $ \(NumberOfRequests n) ->
-        monadicIO $ liftIO $ do
-            void $ mapConcurrently (const (get ctx "/get")) (replicate n ())
-            entries <- readTVarIO (logs ctx)
+    it "different request ids" $ \ctx ->
+        property $ \(NumberOfRequests n) -> monadicIO $ do
+            entries <- liftIO $ do
+                replicateConcurrently_ n (get ctx "/get")
+                skipPrevLogs <$> takeLogs ctx
```
KtorZ, line 221: :+1:

KtorZ (approved):

lib/core/test/unit/Network/Wai/Middleware/LoggingSpec.hs
rvl: This sleep fixes "correct time measures", which failed occasionally. Since there are only 10 examples repeated, it's ok to sleep a bit.

I also inserted waitForServerToComplete into expectLogs, because it might also be susceptible to the race condition.

With the "different request ids" property, which was the one which failed most often, I could not really use sleeps because there are 100 examples run. So in that test I fix it by filtering out any logs from the previous example.

PR #2379, opened 2020-12-08
label update v2020-12-08 = 1.24.2

piotr-iohk:
# Issue Number

# Overview

Just a label update in readme.

# Comments


KtorZ (approved):

Anviking (approved):
Ah, I updated the link but not the label, good catch

We won't have to re-tag and wait again for CI, right?

**If** we care about this being correct if someone downloads the source of the tag and looks at the readme, maybe we can cherry pick it to the tag without re-triggering the builds.

piotr-iohk:
I think no re-tag.

Issue #2328, opened 2020-11-18
Flaky on Windows - BYRON_MIGRATE_01 -  migrate a big wallet requiring more than one tx


piotr-iohk:
# Context
 
https://github.com/input-output-hk/cardano-wallet/runs/1415410584?check_suite_focus=true

# Test Case

# Failure / Counter-example

```
Failures:

  src/Test/Integration/Scenario/API/Byron/Migrations.hs:295:27: 
  1) API Specifications, BYRON_MIGRATIONS, BYRON_MIGRATE_01 -  migrate a big wallet requiring more than one tx
       While verifying (Status {statusCode = 200, statusMessage = "OK"},Right (ApiWallet {id = ApiT {getApiT = WalletId {getWalletId = f2443cf608fdde7b504bd25765c188528e459673}}, addressPoolGap = ApiT {getApiT = AddressPoolGap {getAddressPoolGap = 20}}, balance = ApiT {getApiT = WalletBalance {available = Quantity {getQuantity = 949993300695}, total = Quantity {getQuantity = 949993300695}, reward = Quantity {getQuantity = 0}}}, delegation = ApiWalletDelegation {active = ApiWalletDelegationNext {status = NotDelegating, target = Nothing, changesAt = Nothing}, next = []}, name = ApiT {getApiT = WalletName {getWalletName = "Empty Wallet"}}, passphrase = Just (ApiWalletPassphraseInfo {lastUpdatedAt = 2020-11-18 00:37:00.8580276 UTC}), state = ApiT {getApiT = Ready}, tip = ApiBlockReference {absoluteSlotNumber = ApiT {getApiT = SlotNo 3370}, slotId = ApiSlotId {epochNumber = ApiT {getApiT = EpochNo {unEpochNo = 17}}, slotNumber = ApiT {getApiT = SlotInEpoch {unSlotInEpoch = 90}}}, time = 2020-11-18 00:38:55 UTC, block = ApiBlockInfo {height = Quantity {getQuantity = 1278}}}}))
       Waited longer than 90s to resolve action: "wallet balance = expectedBalance".
       expected: Quantity {getQuantity = 999992948100}
        but got: Quantity {getQuantity = 949993300695}

  To rerun use: --match "/API Specifications/BYRON_MIGRATIONS/BYRON_MIGRATE_01 -  migrate a big wallet requiring more than one tx/"

Randomized with seed 1261157913

Finished in 3169.0666 seconds
[...]

PR #2383, opened 2020-12-09
Attempt to fix flakyness in big wallet migrations scenarios

KtorZ:
# Issue Number

#2207 #2328 

# Overview

- [ ] I have attempted to fix flakyness in big wallet migrations scenarios by retrying a migration several time when transaction are rolled back along the way.

# Comments

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
@@ -20,6 +20,7 @@ import Cardano.Mnemonic
     ( entropyToMnemonic, genEntropy )
 import Cardano.Wallet.Api.Types
     ( ApiByronWallet
+    , ApiT (..)
     , ApiTransaction
     , ApiUtxoStatistics
     , ApiWallet
@@ -37,8 +38,12 @@ import Cardano.Wallet.Primitive.AddressDerivation.Icarus
     ( IcarusKey )
 import Cardano.Wallet.Primitive.AddressDerivation.Shelley
     ( ShelleyKey )
+import Cardano.Wallet.Primitive.Types.Address
+    ( Address )
+import Cardano.Wallet.Primitive.Types.Tx
+    ( TxStatus (..) )
 import Control.Monad
-    ( forM_ )
+    ( forM_, void, when )
 import Control.Monad.IO.Class
     ( liftIO )
 import Control.Monad.Trans.Resource
@@ -47,6 +52,8 @@ import Data.Generics.Internal.VL.Lens
     ( view, (^.) )
 import Data.Maybe
     ( mapMaybe )
+import Data.Proxy
+    ( Proxy (..) )
 import Data.Quantity
     ( Quantity (..) )
 import Data.Text
@@ -78,8 +85,10 @@ import Test.Integration.Framework.DSL
     , postByronWallet
     , randomAddresses
     , request
+    , unsafeRequest
     , unsafeResponse
     , verify
+    , waitForTxImmutability
     , walletId
     , (.>)
     )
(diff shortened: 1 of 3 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
```diff
@@ -469,6 +482,50 @@ spec = describe "BYRON_MIGRATIONS" $ do
...
+            -- 2/ Attempt to migrate
+            _ <- request @[ApiTransaction n] ctx endpointMigration Default payload
+
+            -- 3/ Wait "long-enough" for transactions to have been inserted.
+            -- What long enough is rather empirical here, but it must satisfies
+            -- one important criteria: it must be long-enough to assume that all
+            -- transactions are no longer in mempools AND, for those who've been
+            -- inserted, they are now deep enough to not be affected by
+            -- rollbacks.
+            threadDelay (5 * oneSecond)
```
KtorZ: Maybe that's actually not long enough. `k = 10` and `slot_length = 200ms`....

Anviking: It should take 2 seconds for a tx to become immutable once it gets included in a block, so 5 seconds — sounds — good enough at least
```
slotLength: 0.2
maxKESEvolutions: 90
securityParam: 10
```

KtorZ: That's my reasoning, 4s is basically `2k`, so 5 seems already conservative. But that's also assuming that the transaction are inserted _right away_. Anyway, I went for 15, just to hammer it down.

lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
```diff
@@ -261,34 +272,36 @@ spec = describe "BYRON_MIGRATIONS" $ do
...
                         (#balance . #getApiT . #available)
-                        ( `shouldBe` Quantity expectedBalance)
+                        (.> (Quantity expectedMinBalance))
                 , expectField
                         (#balance . #getApiT . #total)
-                        ( `shouldBe` Quantity expectedBalance)
+                        (.> (Quantity expectedMinBalance))
                 ]
+
         -- Analyze the target wallet UTxO distribution
```
Anviking, line 303: I get this from the `((`shouldBe` (Just 100)) . Map.lookup 10_000_000_000)` line below:
```
  src/Test/Integration/Scenario/API/Shelley/Migrations.hs:250:19:
  1) API Specifications, SHELLEY_MIGRATIONS, SHELLEY_MIGRATE_01_big_wallet -  migrate a big wallet requiring more than one tx
       While verifying (Status {statusCode = 200, statusMessage = "OK"},Right (ApiUtxoStatistics {total = Quantity {getQuantity = 10000094280000}, scale = ApiT {getApiT = Log10}, distribution = fromList [(10,0),(100,0),(1000,0),(10000,0),(100000,0),(1000000,84),(10000000,8),(100000000,0),(1000000000,0),(10000000000,0),(100000000000,100),(1000000000000,0),(10000000000000,0),(100000000000000,0),(1000000000000000,0),(10000000000000000,0),(45000000000000000,0)]}))
       expected: Just 100
        but got: Just 0

  To rerun use: --match "/API Specifications/SHELLEY_MIGRATIONS/SHELLEY_MIGRATE_01_big_wallet -  migrate a big wallet requiring more than one tx/"

Randomized with seed 151129959
```

lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
Anviking: _Might_ be an idea to
1. Make the reasoning explicit e.g. `threadDelay $ securityParamDelay + 3 * oneSecond`
2. Not go for 15s unless we know we need to.

such that we get a better understanding of exactly what kind of delay is needed, to apply to other tests.

lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
KtorZ: My bad, I did a wrong copy-paste thinking the assertions were the same for both shelley and byron but they aren't. That's been fixed shortly after but I assume you had the former branch :no_mouth:

Anviking (approved):
lgtm!

lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
piotr-iohk: Seems to work on my computer...

KtorZ:
bors merge

KtorZ:
It's been a while...

rvl:
Sorry, I accidentally disabled the cardano-wallet bors jobset in Hydra.

bors r+

PR #2384, opened 2020-12-10
Multi-Asset UTxO: Upgrade primitive types and DB schema

jonathanknowles:
# Issue Numbers

This PR is the first of several that implement the design described in [ADP-347](https://jira.iohk.io/browse/ADP-347) (_Wallet multi-asset UTxO_).

In particular, it implements the following subtasks:
  - [ADP-601](https://jira.iohk.io/browse/ADP-601) (_Upgrade primitive types_)
  - [ADP-602](https://jira.iohk.io/browse/ADP-602) (_Upgrade database schema_)

# User-facing changes

**None.**

# API changes

**None.**

# Overview

This section is intended to be a high-level summary.

For explanations of the design choices made by this PR, please see [ADP-347](https://jira.iohk.io/browse/ADP-347).

## Functionality that should be unaffected by this PR

<details>

### Coin Selection

The coin selection algorithms should operate in exactly the same way as they did before.

This PR modifies these algorithms in only a very small way: to look at only the `Coin` part of `TokenBundle`. Everything else should be the same.

</details>

## Types introduced by this PR

</details>

<details>
<br>

* `TokenPolicyId`

    Uniquely identifies a token policy based on the hash of its policy script.

    ```hs
    newtype TokenPolicyId = TokenPolicyId
        { unTokenPolicyId :: Hash "TokenPolicy" }
    ```

* `TokenName`

    Names a token within the scope of a policy.

    ```hs
    newtype TokenName = TokenName
        { unTokenName :: Text }
    ```

* `TokenQuantity`

    An integral quantity of tokens.

    ```hs
    newtype TokenQuantity = TokenQuantity
        { unTokenQuantity :: Integer }
    ```

* `AssetId`

    Combines a `TokenPolicyId` with a `TokenName` to uniquely identify an asset.

    ```hs
    data AssetId = AssetId
        { tokenPolicyId
            :: !TokenPolicyId
        , tokenName
            :: !TokenName
        }
    ```

* `TokenMap`

    A map of **non-ada** token quantities, scoped by `PolicyId` and `TokenName`.

    ```hs
    newtype TokenMap = TokenMap
        { unTokenMap
            :: Map TokenPolicyId (NonEmptyMap TokenName TokenQuantity)
        }
    ```

* `TokenBundle`

   Combines a non-ada `TokenMap` with an ada `Coin` quantity.

    ```hs
    data TokenBundle = TokenBundle
        { coin
            :: Coin
        , tokens
            :: TokenMap
        }
    ```

</details>

## Types modified by this PR

<details>
<br>

* `TxOut`

    ```patch
    data TxOut = TxOut
        { address
            :: !Address
    -   , coin
    -       :: !Coin
    -     -- ^ The ada amount associated with this output.
    +   , tokens
    +        :: !TokenBundle
    +      -- ^ The tokens associated with this output.
        }
    ```
</details>

## Database tables introduced by this PR

<details>
<br>

### `TxOutToken`

TxOutToken

([PDF](https://github.com/jonathanknowles/cardano-wallet/files/5664999/TxOutToken.pdf))

### `UTxOToken`

UTxOToken

([PDF](https://github.com/jonathanknowles/cardano-wallet/files/5665005/UTxOToken.pdf))

</details>

## Testing
<details>
<br>

### QuickCheck state machine and DB property tests

The existing property tests have been upgraded to generate `UTxO` and `TxOut` values that include multi-asset token bundles.

Care has been taken to ensure that we only generate token bundles of a reasonable size, to minimize our memory footprint.

### Soak testing

As part of testing this PR, the `cardano-wallet-core:test:unit` property test suite has been run hundreds of times to verify that the pass rate and total running time have not been adversely affected.

Some increase in run-time is to be expected, however, as in order to test DB serialization and deserialization, we must generate values of type `TokenBundle`, which occupy more space and take longer to generate than values of type `Coin`.

### Generators and Shrinkers

Generators and shrinkers for token-related types are located within `Gen` modules, according to the following general pattern:

- For type `T` located in module `M`, generators and shrinkers are located in module `M.Gen`.

- Generators and shinkers are defined in pairs, according to the naming pattern `{gen,shrink}TypeRange`, where `Type` is the name of the type, and `Range` refers to the range of values that are generated.

For example:

 - **`{gen,shrink}TokenBundleSmallRange`**
        Generate (or shrink) values of type `TokenBundle`, where elements of the data structure are chosen from small ranges, to promote collisions.

 - **`{gen,shrink}TokenQuantityLargePositive`**
        Generates(or shrink) large values of type `TokenQuantity` that are guaranteed to be strictly positive.

</details>


lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs
```diff
@@ -263,6 +267,33 @@ instance PathPiece TxId where
...
+instance PersistField TokenPolicyId where
+    toPersistValue = toPersistValue . toText
+    fromPersistValue = fromPersistValueFromText
+
+instance PersistFieldSql TokenPolicyId where
+    sqlType _ = sqlType (Proxy @Text)
+
+instance PersistField TokenQuantity where
+    toPersistValue = toPersistValue . toText
+    fromPersistValue = fromPersistValueFromText
```
jonathanknowles, line 295: Perhaps add a comment here to mention why we are serializing to text.

(Neither SQLite nor `persistent-sqlite` support a big integer data type.)

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -1655,7 +1656,10 @@ instance Arbitrary RewardAccount where
 
 instance Arbitrary Coin where
     -- No Shrinking
-    arbitrary = Coin <$> choose (0, 1_000_000_000_000_000)
+    arbitrary = genStrictlyPositiveCoin
+
+genStrictlyPositiveCoin :: Gen Coin
+genStrictlyPositiveCoin = Coin <$> choose (1, 1_000_000_000_000_000)
```
jonathanknowles: Perhaps we can replace this with calls to `Cardano.Wallet.Primitive.Types.Coin.Gen.genCoinLargePositive`.

lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs
```diff
@@ -411,7 +413,10 @@ instance Arbitrary TxMetadata where
     shrink = shrinkTxMetadata
 
 instance Arbitrary Coin where
-    arbitrary = Coin <$> choose (1, 100000)
+    arbitrary = genStrictlyPositiveCoin
+
+genStrictlyPositiveCoin :: Gen Coin
+genStrictlyPositiveCoin = Coin <$> choose (1, 100_000)
```
jonathanknowles: Again, we should consider replacing this with calls to `genCoinLargePositive` (or similar).

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelectionSpec.hs
```diff
@@ -375,7 +377,10 @@ instance Arbitrary Address where
 
 instance Arbitrary Coin where
     -- No Shrinking
-    arbitrary = Coin <$> choose (1, 100000)
+    arbitrary = genStrictlyPositiveCoin
+
+genStrictlyPositiveCoin :: Gen Coin
+genStrictlyPositiveCoin = Coin <$> choose (1, 100_000)
```
jonathanknowles: Again, replace this a call to a generator from `Coin.Gen`.

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -452,10 +456,11 @@ instance Arbitrary (WithPending WalletState) where
                 -- - Sometimes, we also add a withdrawal by creating another
                 --   change output and an explicit withdrawal in the
                 --   transaction.
+                let tokens = TB.fromCoin $ simulateFee $ txOutCoin out
                 let pending = withWithdrawal $ Tx
                         { txId = arbitraryHash
-                        , resolvedInputs = [(inp, coin out)]
-                        , outputs = [out { coin = simulateFee (coin out) }]
+                        , resolvedInputs = [(inp, txOutCoin out)]
+                        , outputs = [out {tokens}]
```
jonathanknowles, line 467: This needs checking. It looks like the call to `simulateFee` was missed out here.

lib/core/test/unit/Data/Map/NonEmpty/StrictSpec.hs
```diff
@@ -0,0 +1,199 @@
...
+                property prop_singleton_toList
+
+        parallel $ describe "Modification" $ do
+
+            it "prop_insert_invariant" $
+                property prop_insert_invariant
+            it "prop_insert" $
+                property prop_insert
+            it "prop_delete" $
+                property prop_delete
```
jonathanknowles: Looks like we're missing a `prop_delete_invariant` here.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs
```diff
@@ -0,0 +1,539 @@
...
+    acc c (asset, quantity) =
+        adjustQuantity c asset (`TQ.add` quantity)
+
+-- | Subtracts one token map from another.
+--
+subtract :: TokenMap -> TokenMap -> TokenMap
+subtract a b = F.foldl' acc a $ toFlatList b
+  where
+    acc c (asset, quantity) =
+        adjustQuantity c asset (`TQ.subtract` quantity)
```
jonathanknowles: After reviewing with @rvl, we decided to remove the `subtract` operation for the moment, as it's not clear that it gives us anything of value.

If we need to add this back in later, we can.

jonathanknowles: Fixed in 5bbd9b34a13a5b5d0b517fd9b2d19e3a1b9d73df.

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
jonathanknowles: Ah, turns out it wasn't missed out after all:

```hs
                let tokens = TB.fromCoin $ simulateFee $ txOutCoin out
                let pending = withWithdrawal $ Tx 
                        { txId = arbitraryHash
                        , resolvedInputs = [(inp, txOutCoin out)] 
                        , outputs = [out {tokens}]
                        , withdrawals = mempty
                        , metadata = Nothing  
                        }
```

lib/core/test/unit/Data/Map/NonEmpty/StrictSpec.hs
jonathanknowles: It turns out we don't need an invariant test here, as the type of `delete` is `k -> NonEmptyMap k a -> Map k a`. The result is an ordinary map.

lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs
jonathanknowles: Fixed in b76899d31d28b0d6342eea43e28f9e399b00c55b.

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
jonathanknowles: Fixed in bce20553d283b5a13c5eeccb76e19d6f88aa7e16.

lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs
jonathanknowles: Fixed in bce20553d283b5a13c5eeccb76e19d6f88aa7e16.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelectionSpec.hs
jonathanknowles: Fixed in bce20553d283b5a13c5eeccb76e19d6f88aa7e16.

lib/core/cardano-wallet-core.cabal
```diff
@@ -181,6 +199,19 @@ library
       Network.Wai.Middleware.ServerError
       Network.Wai.Middleware.Logging
       Ouroboros.Network.Client.Wallet
+        -- TODO:
```
rvl, line 202: Good.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -763,8 +767,9 @@ newDBLayer trace defaultFieldValues mDatabaseFile ti = do
                 insert_ (mkWalletEntity wid meta)
             when (isRight res) $ do
                 insertCheckpoint wid cp <* writeCache wid (Just cp)
-                let (metas, txins, txouts, ws) = mkTxHistory wid txs
-                putTxs metas txins txouts ws
+                let (metas, txins, txouts, txoutTokens, ws) =
```
rvl, line 921: What about renaming txoutTokens → txoutToks rather than breaking the line?

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1281,11 +1310,13 @@ txHistoryFromEntity ti tip metas ins outs ws =
             { W.txInfoId =
                 getTxId txid
             , W.txInfoInputs =
-                map mkTxIn $ filter ((== txid) . txInputTxId . fst) ins
+                mkTxIn <$> filter ((== txid) . txInputTxId . fst) ins
```
rvl: Was this a hlint suggestion? It doesn't seem particularly necessary - especially since the old form is still used a few lines down.

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
```diff
@@ -52,6 +52,8 @@ import qualified Cardano.Wallet.Primitive.AddressDiscovery.Sequential as W
 import qualified Cardano.Wallet.Primitive.Types as W
 import qualified Cardano.Wallet.Primitive.Types.Address as W
 import qualified Cardano.Wallet.Primitive.Types.Coin as W
+import qualified Cardano.Wallet.Primitive.Types.TokenPolicy as TP
+import qualified Cardano.Wallet.Primitive.Types.TokenQuantity as TQ
```
rvl: Just the `W` qualifier would be fine, like the rest.

lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs
```diff
@@ -263,6 +267,35 @@ instance PathPiece TxId where
...
+
+instance PersistField TokenPolicyId where
+    toPersistValue = toPersistValue . toText
+    fromPersistValue = fromPersistValueFromText
+
+instance PersistFieldSql TokenPolicyId where
+    sqlType _ = sqlType (Proxy @Text)
+
+instance PersistField TokenQuantity where
+    -- The `persistent-sqlite` package has no big number type, so we use a
```
rvl: It would be more correct to say that sqlite has no big integer type.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs
```diff
@@ -0,0 +1,285 @@
...
+{-# LANGUAGE UndecidableInstances #-}
+
+-- | Provides the 'TokenBundle' type, which combines a 'Coin' (lovelace) value
+--   with a map of named token quantities, scoped by token policy.
+--
+-- This module is meant to be imported qualified. For example:
+--
+-- >>> import qualified Cardano.Wallet.Primitive.Types.TokenBundle as TB
+--
+module Cardano.Wallet.Primitive.Types.TokenBundle
```
rvl, line 17: Please would you consider following my suggestion from yesterday and rolling TokenQuantity and TokenMap into this module - TokenBundle. Because we don't need so many little modules. Pretty much the only place where TokenQuantity and TokenMap are used is TokenBundle. Think about the person who needs to read this code. If you believe splitting it up will make compile times faster, then prove it.

lib/core/cardano-wallet-core.cabal
```diff
@@ -173,6 +189,8 @@ library
       Crypto.Hash.Utils
       Data.Binary.Get.Safe
       Data.Function.Utils
+      Data.Map.NonEmpty.Strict
+      Data.Map.NonEmpty.Strict.Internal
```
KtorZ: Have you considered existing hackage / stackage alternatives? There's https://hackage.haskell.org/package/nonempty-containers although it is lazy-only. 

Beside, may I consider putting this in a separate **package**. This is decoupled enough from the wallet to live out of it.

lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -494,7 +495,7 @@ expectWalletUTxO coins = \case
         let addr = Address "ARBITRARY"
         let constructUtxoEntry idx c =
                 ( TxIn (Hash "ARBITRARY") idx
-                , TxOut addr (Coin c)
+                , TxOut addr (TB.fromCoin $ Coin c)
```
KtorZ: Minor suggestion because `TB` feels rather cryptic first reading it; what about `Bundle` ? `Bundle.fromCoin` reads without ambiguity.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1458,40 +1510,67 @@ deleteDelegationCertificates wid filters = do
...
+        outputs <- do
+            outs <- fmap entityVal <$> selectList
+                [TxOutputTxId <-. txids]
+                [Asc TxOutputTxId, Asc TxOutputIndex]
+            forM outs $ \out ->
+                (out,) . fmap entityVal <$> selectList
+                    [ TxOutTokenTxId ==. txOutputTxId out
+                    , TxOutTokenTxIndex ==. txOutputIndex out
+                    ]
+                    []
```
KtorZ: The part fetching the output + token is identical (modulo the list of tx ids) here and above. I'd suggest factoring out the common part in a dedicated function to make avoid duplication and clearly show the commonality. 

On a different note, this number of requests that this generates seems rather _large_. I haven't yet checked the database benchmarks but we should make sure to measure this correctly and generate bench database with token bundles.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs
```diff
@@ -0,0 +1,529 @@
...
+    | (policy, tokenQuantities) <- toNestedList b
+    , (token, quantity) <- NE.toList tokenQuantities
+    ]
+
+-- | Converts a token map to a nested list.
+--
+toNestedList
+    :: TokenMap -> [(TokenPolicyId, NonEmpty (TokenName, TokenQuantity))]
+toNestedList =
+    fmap (fmap NEMap.toList) . Map.toList . unTokenMap
```
KtorZ, line 411: Note that, when it comes to database serialization functions, I think it is better to define functions such as fold and traverse to avoid effectively traversing maps twice (once for turning it into list, and then traversing the list to serialize to the database). Although, I imagine that `TokenMap` would remain pretty small (limited by the tx max size anyway) and thus it should be _okay_.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy.hs
```diff
@@ -0,0 +1,96 @@
...
+    toJSON = toJSON . toText
+
+instance ToText TokenPolicyId where
+    toText = toText . unTokenPolicyId
+
+instance FromText TokenPolicyId where
+    fromText = fmap TokenPolicyId . fromText
+
+mkTokenPolicyId :: ByteString -> TokenPolicyId
+mkTokenPolicyId = TokenPolicyId . Hash
```
KtorZ: This is arguably the unsafe way to do this. I'd expect the safe constructor to be `TokenPolicy -> TokenPolicyId`

lib/core/test/unit/Cardano/Wallet/Primitive/Types/TokenBundleSpec.hs
```diff
@@ -0,0 +1,43 @@
...
+
+spec :: Spec
+spec =
+    describe "Token bundle properties" $
+    modifyMaxSuccess (const 1000) $ do
+
+    describe "Class instances obey laws" $ do
+        testLawsMany @TokenBundle
+            [ eqLaws
+            , semigroupLaws
```
KtorZ, line 34: Interesting package / methods :+1:

KtorZ (commented):
I started (50/75) doing mainly a "surface" review to get acquainted with the changes. Nice work, I'll carry on with the review later today and early next week.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs
```diff
@@ -0,0 +1,529 @@
...
+
+--------------------------------------------------------------------------------
+-- Serialization
+--------------------------------------------------------------------------------
+
+-- | When used with the 'Buildable' or 'ToJSON' instances, provides a flat
+-- serialization style, where token quantities are paired with their asset
+-- identifiers.
+--
+newtype Flat a = Flat { getFlat :: a }
```
rvl, line 187: Are both of Flat and Nested _really_ needed?
I would expect the JSON serialisation to be defined only in the Api.Types module, and according to the swagger spec.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs
```diff
@@ -0,0 +1,529 @@
...
+getQuantity :: TokenMap -> AssetId -> TokenQuantity
+getQuantity (TokenMap m) (AssetId policy token) =
+    fromMaybe TQ.zero $ NEMap.lookup token =<< Map.lookup policy m
+
+-- | Updates the quantity associated with a given asset.
+--
+-- If the given quantity is zero, the resultant map will not have an entry for
+-- the given asset.
+--
+setQuantity :: TokenMap -> AssetId -> TokenQuantity -> TokenMap
```
rvl, line 457: I just grepped for this symbol, and couldn't find where it was used by the code. Is it used by the code?

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy/Gen.hs
```diff
@@ -0,0 +1,58 @@
...
+
+--------------------------------------------------------------------------------
+-- Utilities
+--------------------------------------------------------------------------------
+
+mkTokenPolicyId :: Char -> TokenPolicyId
+mkTokenPolicyId
+    = TP.mkTokenPolicyId
+    . unsafeFromHex
+    . B8.replicate tokenPolicyIdHexStringLength
```
rvl: It might be nicer for QC counterexamples if these hashes are just single characters in a..z, or something like that?

lib/core/src/Cardano/Wallet/Primitive/Types/TokenQuantity.hs
```diff
@@ -0,0 +1,121 @@
...
+zero = TokenQuantity 0
+
+--------------------------------------------------------------------------------
+-- Operations
+--------------------------------------------------------------------------------
+
+add :: TokenQuantity -> TokenQuantity -> TokenQuantity
+add (TokenQuantity x) (TokenQuantity y) = TokenQuantity $ x + y
+
+subtract :: TokenQuantity -> TokenQuantity -> TokenQuantity
```
rvl: Is subtract used anywhere?
And what happens if its result is negative?
I think this function should either be failable (i.e. Maybe TokenQuantity) or removed.

rvl (commented):
It looks nice, but there's a lot of fluff.
I am only half-way through this diff, and am getting the feeling there are many function defined which aren't really needed.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs
jonathanknowles: > I just grepped for this symbol, and couldn't find where it was used by the code. Is it used by the code?

**EDIT:** Yes. This function is used by `adjustQuantity`, which is used by `fromFlatList`, which is used by the DB deserialization functions.

~~You're right that this function isn't currently used.~~
> ~~@rvl if you prefer, I can remove it and re-add it in a later PR that addresses coin selection?~~

lib/core/src/Cardano/Wallet/Primitive/Types/TokenQuantity.hs
jonathanknowles: @rvl I've now removed this function.

I think you're right that if we need to compare or find the difference between two quantities, there are other, safer methods that we can use. 👍🏻 

Fixed in 9a5bc7ea71dd4e7ab46b4e02f6b04d12badd0686.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy/Gen.hs
jonathanknowles: @rvl I did consider using just a single character, but the `FromText` instance for `TokenPolicyId` is based on the `Hash` type, which expects to find a byte string of the correct length. If we change the existing generator to use just one character, then the textual serialization tests will fail.

The generator and shrinker currently produce values like this:
* `aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa`
* `bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb`

But I can see it would be nicer to have:
* `PolicyA`
* `PolicyB`
  ... and so on.

Of course, we **_could_** split this generator up into two different generators:

1. Generate valid-length hashes for serialization tests.
2. Generate short hashes for all other tests.

Or, we could use `ShowFmt` in the tests, and just display the first `n` characters in the event of a test failure?

lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs
jonathanknowles: @rvl wrote:
> It would be more correct to say that sqlite has no big integer type.

Good point!

I've updated the comment to reflect this.

Fixed in 9b79407ac74f5064242ffcef8dd5a40fd5e3e540.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
jonathanknowles: > Was this a hlint suggestion?

It _might_ have been 🤷🏻. (Though I can't reproduce it -- perhaps I implemented an `hlint` suggestion at some point and then made a later change which rendered the previous suggestion invalid...)

I must admit my natural instinct is to always use `fmap` (or `<$>`). Since we also use the `a <$> b` pattern quite a bit in the code, it seemed natural to write that.

Having said that, do you have a strong preference for `map` over `fmap` in this case? Is there a reason for preferring `map`? (Perhaps to make sure that it can't be used with some other `Functor`?)

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
jonathanknowles: > What about renaming txoutTokens → txoutToks rather than breaking the line?

As we already use `token` and `tokens` in other places, my preference would be to keep this name consistent with the usage in other places.

We _could_ rename `txoutTokens` to just `tokens`, but I think it's also nice for the name to indicate that these tokens are related to `txouts`.

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
jonathanknowles: @rvl Fixed in af1d54eab6d521ba0d7493688fb03dc152ece007.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs
jonathanknowles: @rvl wrote:
> rolling TokenQuantity and TokenMap into this module - TokenBundle
> If you believe splitting it up will make compile times faster, then prove it.

Compilation time is not the major motivation for having separate modules here.

`TokenMap` and `TokenBundle` are separate modules for the following reasons:

1. They fulfill different purposes:

    a. `TokenMap` defines a map of **_non-ada_** quantities, for situations where ada quantities are forbidden (such as the `mint` field of a transaction).

    b. `TokenBundle` adds an ada quantity and defines a set of extra operations for converting between bundles and `Coin` values.

2. `TokenMap` has an **invariant**. By encapsulating this in its own module and not exporting the public constructor:

    * we can limit the number of operations that potentially violate the invariant, requiring fewer tests.
    * we can avoid the complication of considering `Coin` values when writing tests for `TokenMap`, as it can't include them.<br><br>

3. `TokenMap` and `TokenBundle` have some common operations (for example, `getQuantity` and `adjustQuantity`) but they operate on different types.

    Moving them into the same module means we'd need a way to distinguish between these functions. We could use names that include the type they operate on, such as  `adjustTokenMapQuantity` and `adjustTokenBundleQuantity`, but this seems awkward. By putting them in different modules, they can have separate namespaces, removing the clash.

@rvl wrote:
> so many little modules

I'm not sure I agree that `TokenBundle` or `TokenMap` are small modules:

| Module | Size |
| -- | -- |
| `TokenMap` | 521 lines |
| `TokenBundle` | 285 lines |

I agree that `TokenQuantity` is relatively small, but it's not smaller than `Coin`, which is also separate. Because `TokenQuantity` is used by both `TokenMap` and `TokenBundle`, there doesn't seem to be a good reason to choose one or the other, unless we allow re-exports, which many people dislike due to code navigation issues.

> Think about the person who needs to read this code.

If a person is looking for the definition of type `TokenQuantity` (for example), isn't it easier to first look for a module called `TokenQuantity`, rather than trying to guess which of several other modules it could be in?

With editors that support go-to-definition, this becomes a moot point anyway.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs
jonathanknowles: > Are both of Flat and Nested really needed?

Good point.

We could easily get rid of one if we can show that only one is useful.

`Nested` is useful for debugging, as it groups tokens together by policy, saving space. It might also be useful for clients such as Daedalus, to provide a view that groups tokens by policy (since tokens can share the same name). But we should check this before deciding what to eliminate.

`Flat` might be preferable for clients that don't care about grouping, because it's simpler to parse.

> I would expect the JSON serialisation to be defined only in the Api.Types module

I had anticipated that these types (and their JSON instances) would be re-used by several types within the `Api.Types` module.

Though for the `TokenMap` `FromJSON` instance, we perform validation to guard against zero values, which is something that `TokenMap` itself is responsible for.

We could move these to the `Api.Types` module, but then the responsibility for guarding against zero values would be located in two modules, rather than one. It seems better to put all the validation for a type together with the type, as this way they can also be tested together.

It's actually not the first time we've done this. The following types all define JSON instances in their own modules, in some cases for similar reasons:

- `Quantity`
- `Percentage`
- `StakePoolTicker`
- `PoolOwner`
- `PoolMetadataSource`
- `SMASHPoolId`

lib/core/test/unit/Cardano/Wallet/Primitive/Types/TokenBundleSpec.hs
jonathanknowles: I agree, [`quickcheck-classes`](https://hackage.haskell.org/package/quickcheck-classes/docs/Test-QuickCheck-Classes.html) is certainly a useful package!

I was motivated to look for this functionality after having to define the following instance:

```hs
-- Since the 'TokenBundle' type deliberately does not provide an 'Ord' instance
-- (as that would lead to arithmetically invalid orderings), this means we can't
-- automatically derive an 'Ord' instance for the 'TxOut' type.
--
-- Instead, we define an 'Ord' instance that makes comparisons based on the list
-- representation of a 'TokenBundle'.
--
instance Ord TxOut where
    compare = comparing projection
      where
        projection :: TxOut ->
            ( Address
            , Coin
            , [(TokenPolicyId, NonEmpty (TokenName, TokenQuantity))]
            )
        projection out =
            ( out & view #address
            , out & view (#tokens . #coin)
            , out & view (#tokens . #tokens) & TM.toNestedList
            )
```

lib/core/cardano-wallet-core.cabal
jonathanknowles: @KtorZ wrote:
> Have you considered existing hackage / stackage alternatives? There's https://hackage.haskell.org/package/nonempty-containers although it is lazy-only.

I did indeed look through Hackage for alternatives. There are several non-empty maps, but sadly they all seem to be implemented in terms of lazy maps.

> Beside, may I consider putting this in a separate package. This is decoupled enough from the wallet to live out of it.

We could do that if you prefer. How about `strict-non-empty-containers`? (Unfortunately, `non-empty-containers` is already used.)

Having said that, we don't really have a need for it outside of `cardano-wallet-core`, so we probably wouldn't gain too much from moving it (yet).

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
rvl: OK.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
rvl: Either is fine by me, but if it doesn't need to be changed, there's no need to touch it.

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
rvl: Thanks

lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs
rvl: Thanks

lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs
rvl: Even with go-to-definition support in your editor (even github may support it soon too), having less files is easier to work with.

Yes TokenMap and TokenBundle are different things, but is TokenMap ever going to be used independently of TokenBundle?

It's just lots of lines of fluff.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs
rvl: Can we try to keep the code limited to what is essential for purpose, rather than adding functions which "might be needed"?
For a start, that would make less code to review, and therefore less possible objections from me.
But also it helps reduce confusion when it comes time to refactor.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs
rvl: Yes please - I am looking forward to reviewing this function in the _later_ PR.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy/Gen.hs
rvl: OK, that's too bad then. I have found that full-length hashes make counterexamples more difficult to understand.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenQuantity.hs
rvl: Thanks.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs
jonathanknowles: @rvl wrote:
> Yes TokenMap and TokenBundle are different things, but is TokenMap ever going to be used independently of TokenBundle?

**Yes.** I actually answered that in my [earlier comment](https://github.com/input-output-hk/cardano-wallet/pull/2384#discussion_r542176792):

> `TokenMap` defines a map of non-ada quantities, for situations where ada quantities are forbidden (**such as the `mint` field of a transaction**).

Users have a choice between:
* `TokenBundle`
   allows ada quantities (used by `TxOut.tokens`).
* `TokenMap`
   does not allow ada quantities (will be used by `Tx.mint`, once we support minting).

@rvl wrote:
> Even with go-to-definition support in your editor (even github may support it soon too), having less files is easier to work with.

I'm sorry, I'm not really persuaded by this argument. If that were true, why not just put everything in one large file?

Of course, there has to be a balance somewhere.

The fact that `TokenMap` has an invariant is a good reason to encapsulate this type in its own module. Not exporting the default constructor means that callers are forced to use smart constructors, which we test in `TokenMapSpec`. 

A side benefit is that tests for `TokenBundle` don't have to care about preserving the invariant of `TokenMap`, because the responsibility is encapsulated in that module (and its tests).

I could write more but I'd just be repeating my [earlier comment](https://github.com/input-output-hk/cardano-wallet/pull/2384#discussion_r542176792).

> It's just lots of lines of fluff.

I'm not sure how this comment really helps us move forward.

Can we concentrate on specific points? I'm happy to discuss any issues you find, and make changes, so long as there are valid reasons for doing so.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy/Gen.hs
jonathanknowles: > OK, that's too bad then. I have found that full-length hashes make counterexamples more difficult to understand.

I agree with this.

Though perhaps we can solve this by adding a `ShowFmt` to the tests? We could easily make the `Buildable` instances just display the first `n` bytes of each hash (or perhaps the prefix and suffix: `aaaa...aaaa`).

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
jonathanknowles: > The part fetching the output + token is identical (modulo the list of tx ids) here and above. I'd suggest factoring out the common part in a dedicated function to make avoid duplication and clearly show the commonality.

Well spotted! I didn't notice that.

Fixed in a726e950c6530adc8dbbb5c98a039447035d7c5a.

jonathanknowles: > On a different note, this number of requests that this generates seems rather large. I haven't yet checked the database benchmarks but we should make sure to measure this correctly and generate bench database with token bundles.

Great point.

The DB benchmark currently generates ada-only bundles.

I'll upgrade this to generate bundles that include non-ada tokens.

Anecdotally, when running the DB property tests, I do see a small lengthening in the total run-time (around 5% to 10%). However, this could be partly due to the fact that generating token bundles is inherently more time-consuming than just generating coins.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs
jonathanknowles: @rvl My apologies: I had forgotten that `setQuantity` is used by `adjustQuantity`.

That function is used by `add` and `fromFlatList`, which is used by the DB deserialization functions.

So it's not really possible to remove without removing functionality that we need.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
jonathanknowles: @rvl 

Fixed in d1e2ab04151ee25f907b64b9e174245ac0d895f7.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs
rvl: If you want to dig in and throw around "moving forward", then there's not much I can do, other than say that I'm trying to help you keep things simple.

Perhaps the data type for mint/burn ought to be different to the one for UTxO value? One of these is non-negative. You could consider removing the TokenQuantity module (fluff) and parameterise TokenMap by Integer or Natural. Up to you.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs
rvl: OK.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs
jonathanknowles: > If you want to dig in and throw around "moving forward", then there's not much I can do, other than say that I'm trying to help you keep things simple.

First of all, I appreciate that you're trying to help. Reviewing a large PR is not easy.

But surely you must realize that comments like "It's just lots of lines of fluff" are not helpful. This comes across as a bit condescending, especially when someone has put thought and effort into a design. If you have doubts about a particular design choice, why not just reach out to me and ask? I'm happy to answer questions about anything. I'm also happy to change things (within reason) if there's a clear advantage to doing things another way.

I'd really like to overcome this hurdle and work with you on this.

When you asked me to justify the separation of `TokenMap` and `TokenBundle` (which is mentioned in the design), I gave several reasons, in good faith:

1. Encapsulating the invariant of `TokenMap`.
2. Tests for `TokenMap` can focus on properties of the map (including the invariant), without the additional complication of ada `Coin` values.
3. Separation of concerns and responsibilities. `TokenMap` is responsible for the map structure, and `TokenBundle` is responsible for interoperability with `Coin` values.
4. Allowing the functions of each type to have their own separate namespaces to avoid name clashes (if you want an analogy, think of `Map` and `NonEmptyMap`, which share similar sets of operations, but operate on different types.).

But as far as I can tell, you've not answered these points. It's exhausting to participate in a review where the reviewer doesn't respond to points raised.

This whole thread started with:
> Please would you consider following my suggestion from yesterday and rolling TokenQuantity and TokenMap into this module - TokenBundle. Because we don't need so many little modules.

I get that you have a preference for larger modules, which I respect. But `TokenMap` and `TokenBundle` are not small modules.

After some to and fro, you mentioned that:
> having less files is easier to work with.

This seems very subjective. Where do you draw the line?

I think the following are far more important than module size:

- encapsulation of invariants.
- prevention of regressions.
- separation of concerns.
- clear name-spacing.

Should we sacrifice any of these just for the sake of dropping our module count by one?

> Perhaps the data type for mint/burn ought to be different to the one for UTxO value? One of these is non-negative. You could consider removing the TokenQuantity module (fluff) and parameterise TokenMap by Integer or Natural. Up to you.

This is actually a good point. Though I don't understand the need to use antagonistic language like "fluff". Why is that necessary? As for choice to put `TokenQuantity` in a `newtype`: this was in the design document.

**As for parameterization**: I agree we should at least consider this. I propose that the right time to discuss this will come when we design support for minting. For now, we're probably not in danger of creating negative quantities, as nothing in the wallet creates real token bundles (yet). Perhaps we can settle on `Integer` or `Natural` for now, and then parameterize later, if we need to?

lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs
jonathanknowles: > Can we try to keep the code limited to what is essential for purpose, rather than adding functions which "might be needed"?

As mentioned above, I'm  happy to remove one of these formats.

My original motivation for the flat/nested distinction was testing: visual feedback of grouping + flattening was useful while implementing these types.

I've asked for feedback from the Daedalus team if they have any strong preferences about which JSON format they'd prefer. Once we know the outcome, we can delete what we don't need.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs
rvl: It's not worth my time to rebut every dubious point, if I can see that you feel very strongly about this minor issue. Your PR - do as you wish.
For all this talk of invariants, the only invariant for TokenMap is that values are non-zero - but what happens if that is ever violated? Nothing probably! The wallet will very likely continue to function as normal. That invariant may even make implementing coin selection more difficult. And yet there is no invariant preventing negative token values!? This is a priority inversion due to overfluffing the fluff with too much fluff. Fluff is a technical term meaning low function to LOC ratio.

lib/core/cardano-wallet-core.cabal
KtorZ: I am not saying we should **publish** a new package, but simply organize our internal code differently. Since this is clearly not wallet-specific, there's no reason to keep it in `cardano-wallet-core`; it'll prevent the temptation from adding wallet-specific things to it.

jonathanknowles: > I am not saying we should publish a new package, but simply organize our internal code differently.

Oh definitely! I didn't think that was your suggestion. Sorry if my comment gave that impression. 😄 

I was just trying to come up with an internal package name. I can use something like `strict-non-empty-containers`, unless you feel strongly about an alternative name?

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
jonathanknowles: @jonathanknowles wrote:
> I'll upgrade this to generate bundles that include non-ada tokens.

Just for the record, we decided to handle the upgrade of the DB benchmarks in a separate ticket, which I have created here: https://jira.iohk.io/browse/ADP-627

lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs
jonathanknowles: > the only invariant for TokenMap is that values are non-zero - but what happens if that is ever violated?

The `Eq` instance will become incorrect, and it might even be the case that the minimum ada calculation (provided by the ledger) will give incorrect results (though I would need to double-check this).

Of course, we could rewrite our `Eq` instance to handle zero values, but that seems pointless given we have an auto-derived one that already works.

> The wallet will very likely continue to function as normal.

The wallet will also function as normal if the `TokenBundle` and `TokenMap` code is organized in 2 modules rather than 1, right? Merging modules (which was the reason for this thread) creates extra work here.

> That invariant may even make implementing coin selection more difficult. 

Can you say why the invariant would make coin selection more difficult?

The `getQuantity` function returns `0` in the case where there is no quantity for a given token, so callers shouldn't need to do anything special in the case where a bundle doesn't have a particular token.

> And yet there is no invariant preventing negative token values!?

AFAIK, this doesn't affect the integrity of the map, or the `Eq` instance.

Both [`cardano-ledger-specs`](https://github.com/input-output-hk/cardano-ledger-specs/blob/6e30ecf73c9296bc320d71f3a6b7bfad3ce117e9/shelley-ma/impl/src/Cardano/Ledger/Mary/Value.hs#L104) and [`cardano-node`](https://github.com/input-output-hk/cardano-node/blob/e20559ba6288d09bdc22a44cba7806ff85111274/cardano-api/src/Cardano/Api/Value.hs#L105) also use `Integer` here, presumably so that these types can be reused in the context of minting and burning.

But as you pointed out, we can parameterize this in a later PR, so that:
- we can use `Natural` in the case of a `TokenBundle` that's included in a transaction output.
- we can use `Integer` in the case of a non-ada `TokenMap` included in the `mint` field of a transaction.

lib/core/cardano-wallet-core.cabal
jonathanknowles: @KtorZ I've moved this type (and its tests) to a new internal package `strict-non-empty-containers`.

lib/strict-non-empty-containers/src/Data/Map/NonEmpty/Strict/Internal.hs
```diff
@@ -0,0 +1,190 @@
...
+        m {rest = Map.insert k v (rest m)}
+
+-- | Deletes a key and its value from the map.
+--
+-- When the key is not a member of the map, the original map is returned.
+--
+-- Since this function may reduce the number of elements to zero, it returns
+-- an ordinary map.
+--
+delete :: Ord k => k -> NonEmptyMap k a -> Map k a
```
rvl: This seemed to me like it would be difficult to use.
I thought that `Maybe (NonEmptyMap)` would be more usable in general.
Checking how it is actually used (one instance) - the result is immediately changed to Maybe.
So I suggest changing the type of this function to match.

lib/strict-non-empty-containers/src/Data/Map/NonEmpty/Strict/Internal.hs
```diff
@@ -0,0 +1,190 @@
...
+-- | A non-empty map from keys of type 'k' to values of type 'v'.
+--
+data NonEmptyMap k v = NonEmptyMap
+    { least
+        :: !(k, v)
+      -- ^ Invariant: this key must be less than than all the keys in 'rest'.
+    , rest
+        :: !(Map.Map k v)
+    }
+    deriving (Eq, Foldable, Functor, Generic, Show, Traversable)
```
rvl: This representation seems to make everything harder.
What about wrapping a Map in newtype, and preserve your invariants in the functions which operate on the NonEmptyMap.

rvl (approved):
Too fluffy for my liking, but otherwise good.
Just remember to address @KtorZ comments before merging.

lib/core-integration/src/Test/Integration/Framework/DSL.hs
jonathanknowles: > Minor suggestion because TB feels rather cryptic first reading it; what about Bundle? Bundle.fromCoin reads without ambiguity.

@KtorZ I also like your suggestion of `Bundle`, but it raises the question of how to import other modules like `TokenMap` in a consistent way.

The PR currently import token-related modules in the following standard ways:
```
import qualified Primitive.Types.TokenBundle   as TB
import qualified Primitive.Types.TokenMap      as TM
import qualified Primitive.Types.TokenQuantity as TQ
```
We could make this a bit more verbose:
```
import qualified Primitive.Types.TokenBundle   as TBundle
import qualified Primitive.Types.TokenMap      as TMap
import qualified Primitive.Types.TokenQuantity as TQuantity
```
Or even just use the whole name:
```
import qualified Primitive.Types.TokenBundle   as TokenBundle
import qualified Primitive.Types.TokenMap      as TokenMap
import qualified Primitive.Types.TokenQuantity as TokenQuantity
```

Would any of these options work for you?

lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs
jonathanknowles: > I think it is better to define functions such as fold and traverse to avoid effectively traversing maps twice (once for turning it into list, and then traversing the list to serialize to the database).

@KtorZ I like this idea very much. Though implementing it might require quite a few changes to plumbing in the DB layer.

Would you mind at all if I were to attempt this in a follow-up PR?

I'm also interested to test the effect on the DB benchmarks, if I can.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy.hs
jonathanknowles: > This is arguably the unsafe way to do this. I'd expect the safe constructor to be TokenPolicy -> TokenPolicyId

@KtorZ Thanks for spotting this. I had written this function very early on, and intended to make it safer later on.

I've now removed the function entirely.

Since we don't yet have a `TokenPolicy` type, I'll leave the creation of a `TokenPolicy -> TokenPolicyId` function to a future PR.

Fixed in f7d596eb3ab4920b4b9edb2c6aff4797be8b9182.

lib/strict-non-empty-containers/src/Data/Map/NonEmpty/Strict/Internal.hs
jonathanknowles: > What about wrapping a Map in newtype, and preserve your invariants in the functions which operate on the NonEmptyMap.

That's an interesting idea. Would you mind if I experiment with that in a future PR?

The only (minor) downside I can think of is that we might have some "impossible" cases within pattern matches. But it might be simpler overall.

The published non-empty (lazy) map types all use the "least" element technique, which I borrowed here. (For example: [nonempty-containers](https://hackage.haskell.org/package/nonempty-containers-0.3.4.1/docs/src/Data.Map.NonEmpty.Internal.html#NEMap).)

lib/strict-non-empty-containers/src/Data/Map/NonEmpty/Strict/Internal.hs
jonathanknowles: I did initially consider the type you suggested, but then later on changed it to match the existing (lazy) non-empty map implementations out there.

But I agree that your type makes more sense in the context that we use it, so I'm happy to change it back.

Fixed in: d5a54cd64c53f2f09c8c4703bbbab08a68d037ec.

lib/core-integration/src/Test/Integration/Framework/DSL.hs
KtorZ: I'd actually merge those modules together as `import qualified Primitive.Types.TokenBundle` and import it qualified as `Bundle` :no_mouth: 

The modules are very closely related and actually sharing data-structures from one another. A change in `TokenQuantity` will likely induce a change in `TokenMap` and then `TokenBundle`, so it makes sense to keep them close. Although it may not play well with the modules API which maybe overlap a bit. 

From the choices you've pointed out, the last one (whole name) is what reads clearer in my opinion. And in the end roughly the same to type. `To<tab>` will do in most cases :)

lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs
KtorZ: > Would you mind at all if I were to attempt this in a follow-up PR?

Nope, and it was more of a gratuitous remark / reminder. I don't think it'd matter much for token maps because maps would be rather small anyway (bounded by the transaction max size).

lib/core-integration/src/Test/Integration/Framework/DSL.hs
jonathanknowles: > `import qualified Primitive.Types.TokenBundle   as TokenBundle`
> `import qualified Primitive.Types.TokenMap      as TokenMap`
> `import qualified Primitive.Types.TokenQuantity as TokenQuantity`

@KtorZ wrote:
> From the choices you've pointed out, the last one (whole name) is what reads clearer in my opinion. And in the end roughly the same to type. `To<tab>` will do in most cases :)

Okay!

I've fixed this in ca948b4ed9046e9e6030c29a1f7055c596aab0e7.

PR #2387, opened 2020-12-10
Refresh and use era parameters using latest era genesis. 

KtorZ:
# Issue Number

ADP-609

# Overview

- 9a9f6a81d48b0fbfd6b29b0d15a9a0f5559299ea
  :round_pushpin: **rework management of genesis, protocol and slotting parameters**
    This commit introduces a few changes:

  1. It drops all the unused protocol parameters from the 'Checkpoint'
     table. This requires however a non-trivial database migration which
     is yet to be written.

  2. It moves the blockchain start time and genesis hash in the 'Wallet'
    table instead of the 'Checkpoint' table, since those don't change,
    they need not to be repeated in each checkpoint.

  3. It removes 'epochStability' from the 'GenesisParameters' and
     instead, adds 'securityParam' to the 'SlottingParameters'.
     'GenesisParameters' are now only referring to the 'Byron' genesis
     parameters used at the very beginning of the blockchain, whereas
     'SlottingParameters' are tied to each era.

  4. It renames 'getProtocolParameters' to 'currentProtocolParameters'
     in the network layer for consistency with other functions from the
     network layer.

  5. It moves and rename 'getSlottingParametersForTip' to the network
     layer as 'currentSlottingParameters' such that this can be
     implemented correctly in the network layer using the LSQ protocol
     and the 'GetGenesisConfig' query now available.

  6. It adjusts the block restoration loop to use the latest slotting
     parameters (instead of the Byron's genesis ones) and changes the
     checkpoint stability window to `3k` (the stability window is `3k/f`
     slots, but our checkpoint pruning operates on blocks, so `3k`).

- 6f4e9c37b020b38311d8b3925219e622033d0d9e
  :round_pushpin: **implement 'currentSlottingParameters' in the cardano-node's Network layer.**
  
- fb4b6d25345cc9a3d894223c314fd6bea9286cc8
  :round_pushpin: **add database migration for protocol parameters drop and associated unit test**
    This isn't an easy migration for there's no 'remove column' command available in SQLite. So instead, we have to re-create the table without the column we want and copy back all the previous checkpoints in the new table.

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -702,7 +702,7 @@ data ApiNetworkParameters = ApiNetworkParameters
     , blockchainStartTime :: !(ApiT StartTime)
     , slotLength :: !(Quantity "second" NominalDiffTime)
     , epochLength :: !(Quantity "slot" Word32)
-    , epochStability :: !(Quantity "block" Word32)
+    , securityParam :: !(Quantity "block" Word32)
     , activeSlotCoefficient :: !(Quantity "percent" Double)
     , decentralizationLevel :: !(Quantity "percent" Percentage)
     , desiredPoolNumber :: !Word16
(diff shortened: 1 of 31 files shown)


lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -702,7 +702,7 @@ data ApiNetworkParameters = ApiNetworkParameters
     , blockchainStartTime :: !(ApiT StartTime)
     , slotLength :: !(Quantity "second" NominalDiffTime)
     , epochLength :: !(Quantity "slot" Word32)
-    , epochStability :: !(Quantity "block" Word32)
+    , securityParam :: !(Quantity "block" Word32)
```
KtorZ: *securityParameter

rvl (approved):
Good that we can now query the genesis parameters for the current era, and therefore get the correct ActiveSlotsCoeff value.
The cleanup of the parameters from the Checkpoint database table was much needed.
Nice :+1:

KtorZ:
bors try

KtorZ:
bors merge

PR #2389, opened 2020-12-11
Switch to using UnliftIO.Exception bracket

rvl:
### Issue Number

Probably

### Overview

It turns out that the `Control.Exception.bracket` function isn't very safe to use with async code. The reason is the bracket cleanup handler may be interrupted by async exceptions, and if that happens, then cleanup action may not be done.

I am wondering whether this may be the cause of some test flakiness. Also I wonder whether cleanup handlers are the cause of deadlocking tests.

This change replaces `Control.Exception` with `UnliftIO.Exception` - the "async safe" variant.

The other benefit of `MonadUnliftIO` is that it provides "unlift" for functions of the form `withFoo :: (foo -> IO a) -> IO a` -- i.e. it lets you replace `IO` with other kinds of `m`.

Docs here: https://github.com/fpco/unliftio/tree/master/unliftio#readme

So it may fix flaky tests and/or help to avoid future concurrency bugs.

### Comments

- Most changes are replacing `Control.Exception` with `UnliftIO.Exception` and `Control.Concurrent.Async` with `UnliftIO.Async`.
- The `MonadCatch` class is replaced with `MonadUnliftIO`.
- As mentioned in the UnliftIO readme, we lose the `MonadCatch` instance for `Either` - not a bad thing really.
- The use of `MonadFail` was removed, and that was a good thing.

### DIFF
--- a/lib/core/src/Cardano/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/DB/Sqlite.hs
@@ -52,8 +52,6 @@ import Cardano.DB.Sqlite.Delete
     ( DeleteSqliteDatabaseLog )
 import Cardano.Wallet.Logging
     ( BracketLog, bracketTracer )
-import Control.Concurrent.MVar
-    ( newMVar, withMVarMasked )
 import Control.Monad
     ( join, mapM_, when )
 import Control.Monad.IO.Unlift
@@ -109,6 +107,8 @@ import UnliftIO.Compat
     ( handleIf, mkRetryHandler )
 import UnliftIO.Exception
     ( Exception, bracket_, handleJust, tryJust )
+import UnliftIO.MVar
+    ( newMVar, withMVarMasked )
 
 import qualified Data.Aeson as Aeson
 import qualified Data.ByteString.Char8 as B8
--- a/lib/core/src/Cardano/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/DB/Sqlite.hs
@@ -54,12 +54,10 @@ import Cardano.Wallet.Logging
     ( BracketLog, bracketTracer )
 import Control.Concurrent.MVar
     ( newMVar, withMVarMasked )
-import Control.Exception
-    ( Exception, bracket_, tryJust )
 import Control.Monad
     ( join, mapM_, when )
-import Control.Monad.Catch
-    ( Handler (..), MonadCatch (..), handleIf, handleJust )
+import Control.Monad.IO.Unlift
+    ( MonadUnliftIO (..) )
 import Control.Monad.Logger
     ( LogLevel (..) )
 import Control.Retry
@@ -107,6 +105,10 @@ import GHC.Generics
     ( Generic )
 import System.Log.FastLogger
     ( fromLogStr )
+import UnliftIO.Compat
+    ( handleIf, mkRetryHandler )
+import UnliftIO.Exception
+    ( Exception, bracket_, handleJust, tryJust )
 
 import qualified Data.Aeson as Aeson
 import qualified Data.ByteString.Char8 as B8
@@ -160,7 +162,7 @@ queryLogFunc tr _loc _source level str = traceWith tr (MsgQuery msg sev)
 
 -- | Run an action, and convert any Sqlite constraints exception into the given
 -- error result. No other exceptions are handled.
-handleConstraint :: MonadCatch m => e -> m a -> m (Either e a)
+handleConstraint :: MonadUnliftIO m => e -> m a -> m (Either e a)
 handleConstraint e = handleJust select handler . fmap Right
   where
       select (SqliteException ErrorConstraint _ _) = Just ()
@@ -178,7 +180,7 @@ handleConstraint e = handleJust select handler . fmap Right
 destroyDBLayer :: SqliteContext -> IO ()
 destroyDBLayer (SqliteContext {getSqlBackend, trace, dbFile}) = do
     traceWith trace (MsgClosing dbFile)
-    recovering pol [const $ Handler isBusy] (const $ close' getSqlBackend)
+    recovering pol (mkRetryHandler isBusy) (const $ close' getSqlBackend)
         & handleIf isAlreadyClosed
             (traceWith trace . MsgIsAlreadyClosed . showT)
         & handleIf statementAlreadyFinalized
--- a/lib/core/src/Cardano/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/DB/Sqlite.hs
@@ -222,6 +222,10 @@ startSqliteBackend manualMigration autoMigration tr fp = do
     lock <- newMVar unsafeBackend
     let observe :: IO a -> IO a
         observe = bracketTracer (contramap MsgRun tr)
+    -- runSqlConn is guarded with a lock because it's not threadsafe in general.
+    -- It is also masked, so that the SqlBackend state is not corrupted if a
+    -- thread gets cancelled while running a query.
+    -- See: https://github.com/yesodweb/persistent/issues/981
     let runQuery :: SqlPersistT IO a -> IO a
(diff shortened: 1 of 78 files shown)


KtorZ:
That's actually a very good point. I wonder sometimes if we should also not considering taking a more holistic approach and embrace something like https://github.com/commercialhaskell/rio.

rvl:
Yes we should look at using a better Prelude.
RIO seems decent enough, but really almost anything is better than base Prelude.

This bracket cleanup problem is probably the most serious issue that RIO fixes. I think it's quite likely that it currently affects cardano-wallet (sqlite perhaps), or has in the past (launcher). For the other safety improvements that RIO makes (e.g. partial functions, lazy IO), we are already aware of the pitfalls, and have managed to avoid them.

So I would like to use UnliftIO now, and then we choose a good Prelude later.

rvl:
bors try

KtorZ:
> So I would like to use UnliftIO now, and then we choose a good Prelude later.

Ho sorry I wasn't clear earlier. I wasn't suggesting to switch to RIO in this PR but more as a long-term goal. Your PR as it is make a lot of sense and should be merged A.S.A.P.

(thought I had approve it already!)

KtorZ (approved):

rvl:
Those are real integration test failures due to these changes. I shall need to investigate.

lib/core/src/UnliftIO/Compat.hs
```diff
@@ -39,3 +40,12 @@ mkRetryHandler
...
 mkRetryHandler shouldRetry = [const $ Exceptions.Handler shouldRetry]
+
+-- | A 'MonadUnliftIO' version of 'Control.Monad.Catch.handleIf'.
+handleIf
+    :: (MonadUnliftIO m, Exception e)
+    => (e -> Bool)
+    -> (e -> m a)
+    -> m a
+    -> m a
+handleIf f h = handle (\e -> if f e then h e else throwIO e)
```
KtorZ: :+1:

rvl:
bors try

lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -220,6 +222,10 @@ startSqliteBackend manualMigration autoMigration tr fp = do
     lock <- newMVar unsafeBackend
     let observe :: IO a -> IO a
         observe = bracketTracer (contramap MsgRun tr)
+    -- runSqlConn is guarded with a lock because it's not threadsafe in general.
+    -- It is also masked, so that the SqlBackend state is not corrupted if a
+    -- thread gets cancelled while running a query.
+    -- See: https://github.com/yesodweb/persistent/issues/981
```
KtorZ, line 228: :+1:

lib/core/src/Cardano/Wallet/Logging.hs
```diff
@@ -131,18 +135,26 @@ data BracketLog
...
 
 -- | Run a monadic action with 'BracketLog' traced around it.
-bracketTracer :: MonadCatch m => Tracer m BracketLog -> m a -> m a
+bracketTracer :: MonadUnliftIO m => Tracer m BracketLog -> m a -> m a
 bracketTracer tr action = do
     traceWith tr BracketStart
-    res <- action `onException` traceWith tr BracketException
-    traceWith tr BracketFinish
-    pure res
+    withException
```
KtorZ, line 158: Ok :+1:

lib/launcher/src/Cardano/Launcher.hs
```diff
@@ -148,59 +156,70 @@ instance Exception ProcessHasExited
...
+                (traceWith tr' MsgLauncherCleanup)
+                (action mstdin ph)
+        race waitForExit runAction
+
+    whenLeft res (traceWith tr . MsgLauncherFinish)
     pure res
   where
+    withCreateProcessMod inner =
+        fmap join $ tryJust spawnPredicate $
+        bracket (createProcess process) (liftIO . cleanupProcess) $
```
KtorZ: Why not stick with: https://hackage.haskell.org/package/unliftio-0.2.13/docs/UnliftIO-Process.html#v:withCreateProcess :thinking: ?

lib/test-utils/src/Test/Hspec/Extra.hs
```diff
@@ -97,27 +101,32 @@ aroundAll acquire =
         release  <- newEmptyMVar
         done     <- newEmptyMVar
 
-        pid <- async $ do
-            acquire $ \a -> do
-                putMVar resource a
-                await release
-            unlock done
+        pid <- async $ finally (unlock done) $ acquire $ \a -> do
```
KtorZ: :+1:

lib/test-utils/src/Test/Hspec/Extra.hs
```diff
@@ -97,27 +101,32 @@ aroundAll acquire =
...
 -- | A drop-in replacement for 'it' that'll automatically retry a scenario once
 -- if it fails, to cope with potentially flaky tests.
 --
 -- It also has a timeout of 10 minutes.
 it :: HasCallStack => String -> ActionWith ctx -> SpecWith ctx
-it = itWithCustomTimeout (10*minute)
+it = itWithCustomTimeout (60*minutes)
   where
-    minute = 60
+    minutes = 10
```
KtorZ, line 129: I get this change, but this confuses me. Doesn't matter, but it's interesting to see how brains pick those things up.

lib/test-utils/src/Test/Hspec/Extra.hs
```diff
@@ -126,28 +135,34 @@ itWithCustomTimeout
...
-        `catch` (\(_ :: SomeException)  -> throwIO e))
+itWithCustomTimeout sec title action = specify title $ \ctx -> do
+    e <- newEmptyMVar
+    race (timer >> tryTakeMVar e) (action' (void . tryPutMVar e) ctx) >>= \case
+       Left Nothing ->
+           assertFailure $ "timed out in " <> show sec <> " seconds"
+       Left (Just firstException) ->
+           throwIO firstException
+       Right () ->
+           pure ()
```
KtorZ, line 146: Hmmmm. Okay. We are loosing the "AsyncCancelled" handler and tests will report as failed when we CTRL-C them but It seems to be for the best.

KtorZ (approved):
Ok. Nice work @rvl

lib/launcher/src/Cardano/Launcher.hs
rvl: I need to revert this - it's a partial change.

lib/test-utils/src/Test/Hspec/Extra.hs
rvl: I needed to temporarily reduce the timeout to 1 minute - so in this form you can just edit the minutes variable to set the timeout.

lib/test-utils/src/Test/Hspec/Extra.hs
rvl: This is one of the nice things about unliftio - less things to think about when handling exceptions.
`UnliftIO.Exception.catch` will only catch synchronous exceptions.
Any `AsyncCancelled` or `ThreadKilled` exceptions can bubble up.
The disadvantage is that an MVar is needed here since AsyncCancelled (end of race) is never caught.

lib/test-utils/src/Test/Hspec/Extra.hs
rvl: Actually, there was a mistake here which broke all tests cleanup. It's so embarrassing, I have force pushed over it.

rvl:
Those test failures should be sorted now.

bors r+

rvl:
Weeder, doh.

bors r+

rvl:
Integration tests now seem to be leaking file descriptors, for the cardano-node socket.... (can see them growing with ``lsof +E -U -a -p `pidof integration` ``)

PR #2390, opened 2020-12-11
Move from old generic-lens to lens library.

mimi1vx:
# Issue Number

# Overview

- [x] I have removed bound dependency on old variant of generic-lens library

# Comments

### DIFF
--- a/lib/test-utils/cardano-wallet-test-utils.cabal
+++ b/lib/test-utils/cardano-wallet-test-utils.cabal
@@ -35,7 +35,7 @@ library
     , contra-tracer
     , filepath
     , file-embed
-    , generic-lens >=1.1.0.0 && < 1.2.0.0
+    , lens
     , hspec
     , hspec-core
     , hspec-expectations
--- a/lib/test-utils/cardano-wallet-test-utils.cabal
+++ b/lib/test-utils/cardano-wallet-test-utils.cabal
@@ -35,7 +35,6 @@ library
     , contra-tracer
     , filepath
     , file-embed
-    , lens
     , hspec
     , hspec-core
     , hspec-expectations
--- a/lib/test-utils/src/Test/Utils/Trace.hs
+++ b/lib/test-utils/src/Test/Utils/Trace.hs
@@ -10,7 +10,6 @@
 module Test.Utils.Trace
      ( withLogging
      , captureLogging
-     , countMsg
      ) where
 
 import Prelude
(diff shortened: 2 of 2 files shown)


lib/test-utils/cardano-wallet-test-utils.cabal
```diff
@@ -35,7 +35,7 @@ library
     , contra-tracer
     , filepath
     , file-embed
-    , generic-lens >=1.1.0.0 && < 1.2.0.0
+    , lens
```
KtorZ: I'd rather not depend on `lens` for this though. The `generic-lens` is a simpler alternative. If the problem regards the upper-bound on the library, then let's work that out.

mimi1vx: @KtorZ generic-lens 2.0.0.0 have pretty big incompatible changes., plus don't provide `(^?)`

KtorZ: (^?) is syntactic sugar, we can live without or re-implement it if needed. 

We really use a small part of the generic-lens API so I imagine the changes to be somewhat manageable. We could also consider `microlens` if it turns out to be too complicated.

mimi1vx: > (^?) is syntactic sugar, we can live without or re-implement it if needed.

+1 for this

> We really use a small part of the generic-lens API so I imagine the changes to be somewhat manageable. We could also consider `microlens` if it turns out to be too complicated.

microlens dont have Prism :(

KtorZ: It has _Left and _Right Traversals which if I recall correctly is the only thing we use Prism for in this package, so it's kinda equivalent.

KtorZ: Actually, it's only used in `countMsg` and it seems like this is now dead-code. The function isn't used anywhere anymore so we may as well get rid of it and the dependency altogether.

KtorZ: I wonder why this wasn't caught by `weeder` though :thinking:

mark-stopka: > Actually, it's only used in `countMsg` and it seems like this is now dead-code. The function isn't used anywhere anymore so we may as well get rid of it and the dependency altogether.

Best PR one can get :) Removing unnecessary code is always good! Glad you guys found that out!

KtorZ: I'll have to re-create the PR though, our CI pipeline does not run for PR made from external repositories.

KtorZ:
closing in favor of: https://github.com/input-output-hk/cardano-wallet/pull/2392

PR #2391, opened 2020-12-11
Tweak genesis params to make integration tests almost twice as fast

Anviking:
# Issue Number

Not sure. Not ADP-557 really. 

# Overview

- [x] Lower `k` from 10 to 5
- [x] Lower `epochLength` from 200 to 50
- [x] Lower byron slot length from 0.25 to 0.1 (exactly nothing happens here, so we just want to reach Shelley fast)
- [x] Make Allegra start at epoch 3 instead of 2. The cardano-cli cluster setup takes longer than an epoch, and fails if it enters Allegra.
- [x] Replace `waitForHardFork` with `TimeInterpreter`-based `waitForEpoch`

# Comments

- `stack test cardano-wallet:integration --ta '-j 8'` passes in 220s instead of 420s for me
- It is also possible to control the likelihood of rollback by adjusting f. I have not done so in this PR. I thought about adding a simple env var dial for controlling rollback frequency...

<details>
<summary>
Experimental results of adjusting `f`
</summary>
- f=0.5, sl=0.2, el=50 => 220s
    - 382 `ROLLBACK` printed (from debug print in wallet `follow` function)
- f=0.833333, sl=0.33333, el=30 => 234s
    - Failed the first time I ran it, on ADDRESS_IMPORT_05 - I can import 15000 of addresses
    - 1031 `ROLLBACK` printed
- f=0.3125, sl=0.125, el=80 => 209.7s
    - 143 `ROLLBACK` printed
</details>

### DIFF
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
@@ -117,13 +117,13 @@ import Cardano.Wallet.Primitive.Types
     ( Block (..)
     , EpochLength (unEpochLength)
     , EpochNo (..)
-    , GenesisParameters (getEpochStability)
     , NetworkParameters (..)
     , PoolId (..)
     , ProtocolMagic (..)
     , SlotLength (..)
     , SlottingParameters (..)
     , StartTime (..)
+    , stabilityWindowByron
     )
 import Cardano.Wallet.Primitive.Types.Coin
     ( Coin (..) )
@@ -1899,18 +1899,11 @@ timeInterpreterFromTestingConfig genesisParams dir = do
 
     let startTime = StartTime $ sgSystemStart shelleyGenesis
 
-    let byronSl =
-            unSlotLength $ getSlotLength $ slottingParameters genesisParams
-    let byronEl = getEpochLength $ slottingParameters genesisParams
-    let byronSlotLength =
-            unSlotLength $ getSlotLength $ slottingParameters genesisParams
-    let byronK =
-            SecurityParam
-            . fromIntegral
-            . getQuantity
-            . getEpochStability
-            . genesisParameters
-            $ genesisParams
+    let sp = slottingParameters genesisParams
+    let byronSl = unSlotLength $ getSlotLength sp
+    let byronEl = getEpochLength sp
+    let byronSlotLength = unSlotLength $ getSlotLength sp
+    let byronK = SecurityParam $ getQuantity $ stabilityWindowByron sp
     let byronParams = HF.defaultEraParams byronK (mkSlotLength byronSl)
 
     let shelleyEl = sgEpochLength shelleyGenesis
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
@@ -76,7 +76,7 @@ import Cardano.BM.Data.Output
 import Cardano.BM.Data.Severity
     ( Severity (..) )
 import Cardano.BM.Data.Tracer
-    ( HasPrivacyAnnotation (..), HasSeverityAnnotation (..) )
+    ( HasPrivacyAnnotation (..), HasSeverityAnnotation (..), nullTracer )
 import Cardano.Chain.Genesis
     ( GenesisData (..), readGenesisData )
 import Cardano.CLI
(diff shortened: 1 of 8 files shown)


Anviking:
bors try

Anviking:
[Custom build](https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/773#eb92761b-5471-46c6-9125-b85c38ba950d) running integration tests 7 times. Included this PR and tweaked `f` for more rollback. Failed only once on `ADDRESS_IMPORT_05`.

lib/shelley/test/data/cardano-node-shelley/shelley-genesis.yaml
```diff
@@ -47,12 +47,12 @@ maxLovelaceSupply: 1000000000000000000
 protocolMagicId: 764824073
 networkMagic: 764824073
 networkId: Mainnet
-epochLength: 200
+epochLength: 50
 staking:
 slotsPerKESPeriod: 86400
 slotLength: 0.2
 maxKESEvolutions: 90
-securityParam: 10
```
Anviking, line 55: Hm, it seemed we used k=10 for shelley and k=8 for byron... Weird, I wouldn't have thought that would work...

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -1841,3 +1861,64 @@ instance HasSeverityAnnotation ClusterLog where
...
+    let tFork =
+            fromRational (toRational $ maxRollbacks k) * 10 * byronSlotLength
+
+    let shelleyFork = HF.Bound
+            (RelativeTime tFork)
+            (W.SlotNo $ Cardano.unEpochSize el)
+            (Cardano.EpochNo 1)
+
+    let byronParams = HF.defaultEraParams k (mkSlotLength 0.1)
+    let shelleyParams = HF.defaultEraParams k (mkSlotLength 0.2)
```
KtorZ: This assumes that `k` is the same and would break if we change `k`. Since we can easily ready the byron-genesis from here, I'd not make this assumption in this function and simply read `k` and the slot length from both genesis files.

lib/shelley/test/data/cardano-node-shelley/shelley-genesis.yaml
KtorZ: Why not? `k` is really different between Byron and Shelley on Mainnet as well.

lib/shelley/test/integration/Main.hs
```diff
@@ -271,13 +271,13 @@ specWithServer (tr, tracers) = aroundAll withContext
     onByron _ = pure ()
     afterFork dir _ = do
         traceWith tr MsgSettingUpFaucet
+        let rewards = (,Coin $ fromIntegral oneMillionAda) <$>
+                concatMap genRewardAccounts mirMnemonics
+        moveInstantaneousRewardsTo tr' dir rewards
```
KtorZ, line 276: :+1:

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
Anviking: Sure, could do that. 

I didn't because I am thinking about turning things around anyway: have the code be the source of truth, and write to the genesis files, which I imagine would be useful for running integration tests on multiple eras.

lib/shelley/test/data/cardano-node-shelley/shelley-genesis.yaml
Anviking: > Was  securityParam :param_k: static across eras?

> It is and we (consensus) certainly hope it will remain so

[src](https://input-output-rnd.slack.com/archives/CCZSDE005/p1594041896132800?thread_ts=1594041863.132700&cid=CCZSDE005)

> k is really different between Byron and Shelley on Mainnet as well.

It is not? I guess the "maximum-rollback-length" differs (k vs 3k/f ?), but `k` is the same: 2160.

KtorZ: Ho, my bad, I was certain `k` was different by a factor 10.

KtorZ (approved):

Anviking:
bors try

Anviking:
bors try-
bors r+

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
Anviking: Done. Used the already-parsed genesis-parameters.

rvl (approved):
Good perfomance improvements.

I hope that reducing slot duration does not make tests more fragile.

My suspicion is that there are test cases which take many more slots than is necessary to test their thing, and fixing the worst of these may also reduce test run time.

rvl:
I rebased again and fixed the resulting breakage.

bors r+

PR #2392, opened 2020-12-11
Removed unused dependency to help the OpenSUSE folks packaging the wallet.

KtorZ:
# Issue Number

N/A

# Overview

- 4b2583818d049a8b23dec2dfd1a2f8fff62705b2
  :round_pushpin: **Move from old generic-lens to lens library.**
  
- 8ad680d469036f410cf96611ff75d2a0d8ef8203
  :round_pushpin: **actually remove now-dead code 'countMsg' function in test utils**
    as well as the dependency it requires.

# Comments

### DIFF
--- a/lib/test-utils/cardano-wallet-test-utils.cabal
+++ b/lib/test-utils/cardano-wallet-test-utils.cabal
@@ -35,7 +35,7 @@ library
     , contra-tracer
     , filepath
     , file-embed
-    , generic-lens >=1.1.0.0 && < 1.2.0.0
+    , lens
     , hspec
     , hspec-core
     , hspec-expectations
--- a/lib/test-utils/cardano-wallet-test-utils.cabal
+++ b/lib/test-utils/cardano-wallet-test-utils.cabal
@@ -35,7 +35,6 @@ library
     , contra-tracer
     , filepath
     , file-embed
-    , lens
     , hspec
     , hspec-core
     , hspec-expectations
--- a/lib/test-utils/src/Test/Utils/Trace.hs
+++ b/lib/test-utils/src/Test/Utils/Trace.hs
@@ -10,7 +10,6 @@
 module Test.Utils.Trace
      ( withLogging
      , captureLogging
-     , countMsg
      ) where
 
 import Prelude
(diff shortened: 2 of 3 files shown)


paweljakubas (approved):
ok, but why weeder has not detected this earlier :thinking:

KtorZ:
> ok, but why weeder has not detected this earlier thinking

I am wondering the same :thinking:

rvl:
I tried running `stack build && weeder` locally on the revision before this was merged and didn't get any weeder warnings. :confused:

rvl:
@mark-stopka - I am interested in how you get along packaging this for OpenSUSE. Suppose cardano-wallet requires 
`network==3.1.2.1` (it does) but some other installed package requires `network<3.1` - is it possible to install both versions of `ghc-network.rpm` at the same time?

mark-stopka:
@rvl, hard, we do have the ability to have 2 networks versions, but we rather work with upstream and remove the obsticle thru ongoing integration process, we do have an objective to have smooth experience of running ALL components and it's dependencies on RISC-V and IBM SeriesZ platforms, we are now going thru first update of the repository and writing some automation tools for compehensive CI/CD pipeline to [support dCloud needs](https://cardano.ideascale.com/a/dtd/Decentralized-Cloud-Platform/322909-48088), we also commited substancial financial resources towards development of [Cardano Haskell Networking Stack](https://github.com/2nd-Layer/rust-cardano-ouroboros-network) implementation in Rust.

We are also ramping up the number of Haskell developers available thru partnership with local university.

TL;DR we will see, but we prefer ustream first cooperation in the spirit of free and opensource software decentralized projects, but we have contingency in case it will be needed.

CC: @dcoutts, @erikd, @hasufell

rvl:
That sounds really good.

As developers of cardano-wallet, we are keen to see our software running on any OS or machine wanted by the community.

But I think it will be tricky to get cardano-wallet and cardano-node packaged "conventionally" on any distro for which parallel installation of multiple versions of Haskell libraries is hard (e.g. rpm or deb based distros). Maybe you get lucky and it works, but it will be fragile. We may accidentally break the OpenSUSE package by adding a dependency version constraint.

Something easier may be packaging a static binary in a RPM (that would be x86_64 only though).

mark-stopka:
@rvl unlike RHEL, SUSE produces the executables statically linked agains Haskell dependencies, so we would need multiple version support only in build environment, and we have that, we just want to avoid it if possible, that is a wall we are going to climb when we get there, now we focus on GHC-9.0.x development environment, we got 92 packages from Hackage to fix first, some are simple `contraint` overwrite, some require substantial changes, see our worklow

https://github.com/opensuse-haskell/configuration/pull/53

rvl:
Oh good - it's sensible for a distro to statically link Haskell libraries and dynamically link system libraries.

You are probably already aware that cardano-wallet is on ghc 8.6.5, but will soon move to ghc 8.10.x (not many changes required).

Good luck :+1: .

PR #2397, opened 2020-12-16
Fix bug preventing txs to be listed when filter spans multiple eras

Anviking:
# Issue Number

ADP-626

# Overview

- [x] Make rendering of MsgInterpreterPastHorizon more readable in logs
- [x] Add regression test for ADP-626
- [x] Fix ADP-626 by wrapping `HF.Qry` in a new monad to allow safe composition (like we used to before)
   - [x] TODO: Adjust doc comments

# Comments

- Tested manually that e.g. `cardano-wallet transaction list 94962db0b6ecda7b6d1af8e2afce61b166065545 --start 2019-10-17T10:15:00Z --end 2020-12-20T10:15:00Z` (Filter including Byron, Shelley and Allegra) works.
- We used to have this approach before the Allegra integration
- Alternative: Replace `slotRangeFromRelativeTimeRange :: Range RelativeTime -> Qry (Range SlotNo)` with something that takes a `TimeInterpreter m` and returns `m (Range SlotNo)`
    - To this seems to undermine having a `Qry` monad in the first place, if the calling wallet-code cannot compose queries, or has to be careful when doing so.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
@@ -95,6 +95,8 @@ import Data.Generics.Internal.VL.Lens
     ( (^.) )
 import Data.Maybe
     ( fromMaybe )
+import Data.Text
+    ( Text )
 import Data.Text.Class
     ( ToText (..) )
 import Data.Time.Clock
@@ -104,7 +106,7 @@ import Data.Word
 import Fmt
     ( blockListF', build, fmt, indentF )
 import GHC.Stack
-    ( HasCallStack, getCallStack, prettySrcLoc )
+    ( CallStack, HasCallStack, getCallStack, prettySrcLoc )
 import Ouroboros.Consensus.BlockchainTime.WallClock.Types
     ( RelativeTime (..), SystemStart (..), addRelTime )
 import Ouroboros.Consensus.HardFork.History.Qry
@@ -413,7 +415,7 @@ instance ToText TimeInterpreterLog where
       where
         renderPastHorizonException (PastHorizon callStack expr eras) t0 = mconcat
             [ "\nCalled from:\n"
-            , T.pack $ show $ prettySrcLoc $ snd $ last $ getCallStack callStack
+            , prettyCallStackTop callStack
             , "\nConverting expression:\n"
             , T.pack $ show expr
             , "\n\nWith knowledge about the following eras:\n"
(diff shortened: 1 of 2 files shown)


lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
```diff
@@ -334,29 +389,62 @@ data TimeInterpreter m = forall eras. TimeInterpreter
...
+            , renderEnd end
+            ]
+
+
+        renderEnd (HF.EraEnd b) = renderBound b
+        renderEnd (HF.EraUnbounded) = "<unbounded>"
+
+        renderBound (HF.Bound _time _slot epoch) = mconcat
+            [ build $ show epoch
+--            , "( "
```
Anviking: Oops, should be cleaned up.

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
```diff
@@ -128,19 +126,76 @@ import qualified Cardano.Slotting.Slot as Cardano
...
+runQuery
+     :: HasCallStack
+     => StartTime
+     -> Interpreter xs
+     -> Qry a
+     -> Either HF.PastHorizonException a
+runQuery startTime int = go
+  where
+     go :: Qry a -> Either HF.PastHorizonException a
+     go (EraContainedQry q) = HF.interpretQuery int q
```
Anviking, line 189: `q` will be run alone.

Anviking:
bors try

rvl (approved):
Oops and thanks for the fix.

I added a commit to make the regression test actually fail without the bugfix applied.

There should be a fix in ouroborus-consensus for this, but this works.

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
```diff
@@ -334,29 +389,52 @@ data TimeInterpreter m = forall eras. TimeInterpreter
...
             , "This should not happen because "
             , T.pack reason
-            , " Full error is: "
-            , T.pack (show e)
+            , renderPastHorizonException e t0
+            ]
+      where
+        renderPastHorizonException (PastHorizon callStack expr eras) t0 = mconcat
+            [ "\nCalled from:\n"
+            , T.pack $ show $ prettySrcLoc $ snd $ last $ getCallStack callStack
```
KtorZ: Note that `last` is partial. I hope there's always a stack :thinking:

KtorZ (approved):

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
rvl: I changed it to not use `last`.
It would be better to pretty print the entire callstack though. Because I think this line depends on the calling function to _not_ have HasCallStack.

PR #2400, opened 2020-12-16
Make wallet run on Mary

### LINKED ISSUE #2427: Flaky: ProcessHasExited "cluster didn't start correctly
# Context
 
 

https://hydra.iohk.io/build/5275648

# Test Case

N/A - error during local test cluster setup before tests.

# Failure / Counter-example

```
       ProcessHasExited "cluster didn't start correctly: 
           [ Left user error (Waited too long for: pool registration)
           , Left user error (Waited too long for: pool registration)
           , Left user error (Waited too long for: pool registration)
           , Left user error (Waited too long for: pool registration)
           ]" (ExitFailure 1)
```

Anviking:
# Issue Number

ADP-623

# Overview

- [x] Make the wallet run on Mary,
- [x] MA Value types are converted to/from TokenBundles.
- [x] Roundtrip tests for Value/TokenBundle
- [x] Fix disabled era error https://input-output-rnd.slack.com/archives/CFKLUH4R0/p1608146419455500
- [x] Test cluster does hard forks at start up - instantly.
- [x] Uses latest cardano-node master branch, with input-output-hk/cardano-node#2270 merged.

# Comments

- I imagine properly thinking about and removing the duplication would be pretty fun. (rvl: we can do that in the next PR)
- Local test cluster changes for testing in multiple eras ⇒ #2438


rvl:
I pushed some more WIP:
- Added a stack snapshot for input-output-hk/cardano-node#2216.
- Adjusted cluster launch code and config to use InstaFork™ technology.
- Renamed "shelley test cluster" to just "local cluster".
- Removed the commits from #2397 and #2391 and rebased on master.

KtorZ:
@Anviking why was it necessary to re-generate all the faucet keys :thinking: ?

Anviking:
@KtorZ I think @rvl renamed shelley-test-cluster to local-cluster, including the folder the keys where in

rvl:
I pulled out the local-cluster rename commits. We can do that another time.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy.hs
```diff
@@ -5,10 +5,10 @@
 module Cardano.Wallet.Primitive.Types.TokenPolicy
     (
       -- * Token Policies
-      TokenPolicyId
+      TokenPolicyId (..)
```
jonathanknowles, line 8: @Anviking

Just a thought.

Rather than exporting the default data constructors for `TokenPolicyId` and `TokenName`, perhaps we could define and export smart constructors instead?

It would be my preference to not export these default constructors, as it opens the door to people accidentally creating invalid token policy IDs (which should have a certain length), and this could cause subtle regressions in other places.

In the case of `TokenName`, there is actually already a `FromText` instance. Right now, this never fails, but if token names were to require validation, then calling `fromText` would be safer than using the default constructor.

Anviking: > It would be my preference to not export these default constructors, as it opens the door to people accidentally 

👍 Good catch, agree.

Though, I wonder what the best solution is 🤔 

> were to require validation, then calling fromText would be safer than using the default constructor.

If fromText fails in the node-to-wallet type-conversion, then the wallet and node have different validation rules, and surely the node's must be the correct ones. Do we actually want to validate in the conversion?

rvl: I exported those constructors, because it's impractical not to have them. PolicyId and AssetName values coming from the node are already validated.

jonathanknowles: > I exported those constructors, because it's impractical not to have them. PolicyId and AssetName values coming from the node are already validated.

Well, one alternative is to create functions `unsafePolicyId` and `unsafePolicyName` and export these instead of the default data constructors.

Advantages:

1. Callers of these functions are explicitly reminded that they're using unsafe functions. The onus is on them to check that the input is valid.

2. When people reviewing code see a new call to `unsafeX`, they can ask the question: "Has this input been validated?"

3. We can easily track usages of `unsafe` functions with a simple call to `grep` (or `ag`). This is a bit harder to do with the default data constructor.

rvl: The type constructor itself could be called UnsafePolicyId if you like.

jonathanknowles: > The type constructor itself could be called UnsafePolicyId if you like.

That's not a bad alternative (it passes the `grep` test), but it alters the output from `show`. So instead of:

```
PolicyId "deaddead....
```

We'd have:
```
UnsafePolicyId "deaddead....
```

It's a minor issue of course, but in a structure with lots of policy IDs, this increased length does tend to add up.

rvl:
I added a hard fork era config parameter to the test cluster.
It's almost possible to run tests on both allegra and mary, but for 2 problems with environment variables. One is fixed in #2419.

rvl:
Tests passed for me locally, except for one instance of the flaky test fixed by #2419.

bors try

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy.hs
```diff
@@ -57,14 +56,12 @@ instance ToText TokenPolicyId where
...
---------------------------------------------------------------------------------
--- Token names
---------------------------------------------------------------------------------
+    fromText = fmap UnsafeTokenPolicyId . fromText
 
-newtype TokenName = TokenName
-    { unTokenName :: Text }
+-- | Token names, defined by the monetary policy script.
+newtype TokenName =
+    -- | Construct a 'TokenTokenName' without any validation.
```
jonathanknowles: There's a typo here:

```suggestion
    -- | Construct a 'TokenName' without any validation.
```

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -504,7 +513,9 @@ withNetworkLayer tr np addrInfo (versionData, _) action = do
                     (Map.size stake)
                     (Map.size rewards)
                 return res
-            Left{} -> pure $ W.StakePoolsSummary 0 mempty mempty
+            Left e -> do
+                liftIO $ putStrLn $ "### " <> show e
```
rvl: Remove this

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -845,16 +873,84 @@ fromAllegraTx tx =
...
+    mkTokenName = W.UnsafeTokenName
+        . T.decodeUtf8With T.lenientDecode
+        . Cardano.serialiseToRawBytes
+
+    -- Lovelace to coin. Quantities from ledger should always fit in Word64.
+    mkCoin :: Cardano.Lovelace -> W.Coin
+    mkCoin = W.Coin . unsafeToWord64 . unQuantity . Cardano.lovelaceToQuantity
+
+    -- Do Integer to Natural conversion. Quantities from ledger TxOuts from
+    -- never be negative (but unminted values could be negative).
```
jonathanknowles: A small typo:

```suggestion
    -- Do Integer to Natural conversion. Quantities from ledger TxOuts can
    -- never be negative (but unminted values could be negative).
```

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -517,43 +528,96 @@ withNetworkLayer tr np addrInfo (versionData, _) action = do
...
                     (queue `send` cmd)
                 return $ fromPoolDistr <$> result
 
+            AnyCardanoEra MaryEra -> do
+                let cmd = CmdQueryLocalState pt (QueryIfCurrentMary Shelley.GetStakeDistribution)
+                result <- handleQueryFailure $ timeQryAndLog "GetStakeDistribution" tr
+                    (queue `send` cmd)
+                return $ fromPoolDistr <$> result
+            AnyCardanoEra ByronEra -> do
+                -- FIXME: Fail.
```
rvl: It will fail due to QueryIfCurrentShelley, but a shortcut could be used.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -517,43 +528,96 @@ withNetworkLayer tr np addrInfo (versionData, _) action = do
...
                 return $ optimumNumberOfPools <$> result
 
+            AnyCardanoEra MaryEra -> do
+                let cmd = CmdQueryLocalState pt (QueryIfCurrentMary Shelley.GetCurrentPParams)
+                result <- handleQueryFailure $ timeQryAndLog "GetCurrentPParams" tr
+                    (queue `send` cmd)
+                return $ optimumNumberOfPools <$> result
+
+            AnyCardanoEra ByronEra -> do
+                -- FIXME: Simply fail
```
rvl: Perhaps we can clean up this duplication in the next PR.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -751,14 +815,16 @@ mkTipSyncClient tr np localTxSubmissionQ onTipUpdate onPParamsUpdate onInterpret
         queryLocalState Nothing    __ = return ()
         queryLocalState (Just era) pt = do
             mb <- timeQryAndLog "GetEraStart" tr $ localStateQueryQ `send`
+                -- TODO: What should this code actually do? Return the most
+                -- recent hard-fork?
```
rvl, line 829: We need to ask Daedalus team and adjust the API spec possibly.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -1190,6 +1289,7 @@ handleMuxError tr onResourceVanished = pure . errorType >=> \case
         traceWith tr Nothing
         pure onResourceVanished
     MuxCleanShutdown -> pure False
+    MuxBlockedOnCompletionVar _ -> pure False -- TODO: Is this correct?
```
rvl, line 1281: Is this a new error code since the cardano-node version bump?

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy.hs
KtorZ: @jonathanknowles I think we do not care much about that last bit :blush: ... if that really bothers you: https://hackage.haskell.org/package/quiet-0.2/docs/Quiet.html

lib/shelley/test/integration/Main.hs
```diff
@@ -152,249 +154,262 @@ import qualified Test.Integration.Scenario.CLI.Shelley.Transactions as Transacti
...
+    parallelIf flag = if flag then parallel else sequential
+
+    testEras run = forM_ [MaryHardFork, AllegraHardFork] $
+      \era -> describe (show era) $ run era
 
 -- | Do all the program setup required for integration tests, create a temporary
 -- directory, and pass this info to the main hspec action.
 withTestsSetup :: (FilePath -> (Tracer IO TestsLog, Tracers IO) -> IO a) -> IO a
 withTestsSetup action = do
-    -- Handle SIGTERM properly
```
rvl: Oops I messed up the indenting

rvl (approved):
Have checked through it - seems more or less ok - except for the unfortunate amount of duplication which @Anviking is working on fixing.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
Anviking: Yes

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy.hs
jonathanknowles: > @jonathanknowles I think we do not care much about that last bit blush ...

Fair enough.

> if that really bothers you: https://hackage.haskell.org/package/quiet-0.2/docs/Quiet.html

We actually are _already_ using `Quiet` here:

```
newtype TokenPolicyId = TokenPolicyId
    { unTokenPolicyId :: Hash "TokenPolicy" }
    deriving stock (Eq, Ord, Generic)
    deriving (Read, Show) via (Quiet TokenPolicyId)
```

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy.hs
rvl: Thanks, fixed.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
rvl: Thanks, fixed.

lib/shelley/test/integration/Main.hs
rvl: Should be fixed now. I must have accidentally typed the key shortcut for really ugly automatic formatting.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -4,6 +4,7 @@
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE FunctionalDependencies #-}
```
KtorZ: :thinking: I haven't seen yet why this was introduced, but we can usually solve same problems with `TypeFamilies` which IMO gives more readable code and compilation errors.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -845,16 +873,84 @@ fromAllegraTx tx =
...
+
+fromCardanoValue :: Cardano.Value -> TokenBundle.TokenBundle
+fromCardanoValue = uncurry TokenBundle.fromFlatList . extract
+  where
+    extract value =
+        ( mkCoin $ Cardano.selectLovelace value
+        , convert $ Cardano.valueToList value )
+    convert = strip . map (bimap fromCardanoAssetId mkQuantity)
+
+    -- Remove ada from value.
```
KtorZ: ?

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -784,12 +847,38 @@ mkTipSyncClient tr np localTxSubmissionQ onTipUpdate onPParamsUpdate onInterpret
...
+        --
+        -- Perhaps both of these points towards @forAnyEra@ including `send`...
+        _forAnyEra
+            :: AnyCardanoEra
+            -> (Query Byron.ByronBlock res)
+            -> (forall shelleyEra. Query (Shelley.ShelleyBlock (shelleyEra StandardCrypto)) res)
+            -> Query
+                (CardanoBlock StandardCrypto)
+                (Either (MismatchEraInfo (CardanoEras StandardCrypto)) res)
+        _forAnyEra era inByron inShelleyBased = case era of
```
KtorZ: Where's that function being used?

lib/shelley/test/data/cardano-node-shelley/node.config
```diff
@@ -30,9 +30,12 @@ LastKnownBlockVersion-Major: 2
 LastKnownBlockVersion-Minor: 0
 LastKnownBlockVersion-Alt: 0
 
-# Launcher code will submit an update proposal to trigger the hard-fork.
-TestShelleyHardForkAtEpoch: 1
-TestAllegraHardForkAtEpoch: 3
+# These settings start the test cluster in the Mary era (a "virtual"
+# hard fork happens at the start of the first epoch).
+# They will be generated according to the local test cluster config.
```
KtorZ: Do we need that in this PR? Can't we stick to having this hard-coded in the configuration for now and adds this feature later?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -864,6 +871,48 @@ mkUnsignedTx era ttl cs md wdrls certs =
...
+            Cardano.TxAuxScriptsNone
+
+        , Cardano.txUpdateProposal =
+            Cardano.TxUpdateProposalNone
+
+        , Cardano.txMintValue =
+            Cardano.TxMintNone
+        }
+      where
+        --toErrMkTx :: Cardano.TxBodyError MaryEra -> ErrMkTx
```
KtorZ: Why is this commented out ?

lib/shelley/test/integration/Main.hs
```diff
@@ -152,49 +155,55 @@ import qualified Test.Integration.Scenario.CLI.Shelley.Transactions as Transacti
...
+    -- Because of this, don't run the CLI tests in parallel in hydra.
+    parallel' :: forall a. SpecWith a -> SpecWith a
+    parallel' spec = do
+        nix <- runIO inNixBuild
+        parallelIf (not nix) spec
+
+    parallelIf flag = if flag then parallel else sequential
+
+    testEras run = forM_ [MaryHardFork, AllegraHardFork] $
+        \era -> describe (show era) $ run era
```
KtorZ: We are running the test suite twice, in both Mary and Allegra ?

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
Anviking: Oops, I thought I had removed that.

Anviking:
>  which @Anviking is working on fixing.

Seemed to work out pretty nicely ( `anviking/ADP-623/simpler-mary` ) — but let's do the refactoring in a separate PR after this is merged.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
Anviking: Sounds good

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
Anviking: Redundant - dropping...

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
Anviking: If I try to import `MaryEra` like the other eras I get:

```
/Users/anviking/IOHK/cardano-wallet/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs:85:7: error:
    In module ‘Cardano.Wallet.Shelley.Compatibility’:
      ‘MaryEra’ is a data constructor of ‘CardanoEra’
    To import it use
      import Cardano.Wallet.Shelley.Compatibility( CardanoEra( MaryEra ) )
    or
      import Cardano.Wallet.Shelley.Compatibility( CardanoEra(..) )
   |
85 |     , MaryEra
   |       ^^^^^^^
```

despite other eras being importable just fine:
```
    ( AllegraEra
    , MaryEra
    , ShelleyEra
```
If I switch to importing `CardanoEra (MaryEra)`, then I get this strange error

```
/Users/anviking/IOHK/cardano-wallet/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs:911:42: error:
    • Expected a type, but
      ‘MaryEra’ has kind
      ‘CardanoEra Cardano.MaryEra’
    • In the first argument of ‘Cardano.TxBodyError’, namely ‘MaryEra’
      In the type signature:
        toErrMkTx :: Cardano.TxBodyError MaryEra -> ErrMkTx
      In an equation for ‘mkMaryTx’:
          mkMaryTx
            = left toErrMkTx
                $ Cardano.makeTransactionBody
                    $ Cardano.TxBodyContent
                        {Cardano.txIns = toCardanoTxIn . fst <$> CS.inputs cs,
                         Cardano.txOuts = toMaryTxOut <$> CS.outputs cs,
                         Cardano.txWithdrawals = Cardano.TxWithdrawals
                                                   Cardano.WithdrawalsInMaryEra wdrls,
                         Cardano.txCertificates = Cardano.TxCertificates
                                                    Cardano.CertificatesInMaryEra certs,
                         Cardano.txFee = Cardano.TxFeeExplicit
                                           Cardano.TxFeesExplicitInMaryEra fees,
                         Cardano.txValidityRange = (Cardano.TxValidityNoLowerBound,
                                                    Cardano.TxValidityUpperBound
                                                      Cardano.ValidityUpperBoundInMaryEra ttl),
                         Cardano.txMetadata = maybe
                                                Cardano.TxMetadataNone
                                                (Cardano.TxMetadataInEra
                                                   Cardano.TxMetadataInMaryEra)
                                                md,
                         Cardano.txAuxScripts = Cardano.TxAuxScriptsNone,
                         Cardano.txUpdateProposal = Cardano.TxUpdateProposalNone,
                         Cardano.txMintValue = Cardano.TxMintNone}
            where
                toErrMkTx :: Cardano.TxBodyError MaryEra -> ErrMkTx
                toErrMkTx = ErrConstructedInvalidTx . T.pack . Cardano.displayError
    |
911 |         toErrMkTx :: Cardano.TxBodyError MaryEra -> ErrMkTx
    |                                          ^^^^^^^
```

Can remove the comments completely for now.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
Anviking: @rvl ?

lib/shelley/test/data/cardano-node-shelley/node.config
Anviking: Started a related [slack thread](https://input-output-rnd.slack.com/archives/GBT05825V/p1610379419068700)

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy.hs
Anviking: I would have been slightly in favour of `unsafePolicyId` over `UnsafePolicyId` for the sake of:
- Less confusing `Show` output
- No unintended **de**-construction outside the module

but won't object, either.

Anviking:
bors try

lib/shelley/test/integration/Main.hs
rvl: Yes - slower - but I think it's better to be on the safe side.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
rvl: I clarified the function. When constructing the TokenBundle from an asset map, it needs separate ada and non-ada assets.

lib/shelley/test/data/cardano-node-shelley/node.config
rvl: I think we need it, because of all the AnyCardanoEra cases which are being changed. In this PR? Well, it's testing the new code from this PR.

But one thing led to another, and a small patch on the test cluster spread out a bit. So I have removed those test cluster era changes and put them into PR #2438.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy.hs
rvl: @Anviking we also need **de**-construction in Compatibility.hs.

The situation is a bit ridiculous with the multiple layers of wrapping around a 32 byte bytestring (maximum length not enforced by our code by the way). And Text is probably not the correct inner type anyway!

rvl:
bors try

KtorZ (approved):

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy.hs
Anviking: Ok, fair, perhaps.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -538,43 +545,88 @@ withNetworkLayerBase tr np addrInfo (versionData, _) action = do
...
                 return $ fromPoolDistr <$> result
 
+            AnyCardanoEra MaryEra -> do
+                let cmd = CmdQueryLocalState pt (QueryIfCurrentMary Shelley.GetStakeDistribution)
+                result <- handleQueryResult "GetStakeDistribution"
+                    (queue `send` cmd)
+                return $ fromPoolDistr <$> result
+            AnyCardanoEra ByronEra ->
+                -- see also: #2419
+                throwE $ ErrStakeDistributionQuery "Can't query stake distribution in byron era"
```
Anviking, line 561: 👍

Anviking (commented):
lgtm

Anviking:
Seems both buildkite and hydra passed in the `bors try`, but the bors status has failed to be updated
bors r+

Anviking:
🤞 
bors r+

PR #2401, opened 2020-12-16
Integrate with cardano-addresses 3.2.0 and add golden for script hashes for different simple script versions

paweljakubas:
# Issue Number

https://jira.iohk.io/browse/ADP-673

# Overview

- [x] I have added the first test for `RequireSignatureOf` of `Script` for both `SimpleScriptV1` and `SimpleScriptV2`
- [x] I have added more golden for `Script`
- [x] I have added golden for `Script` extended by `StartSlot` `ExpireSlot`
- [x] Rest code aligning when cardano-addresses 3.2.0 is released

# Comments

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
@@ -634,24 +634,6 @@ spec = describe "SHELLEY_ADDRESSES" $ do
                 let walAddr = fst (addrs !! idx ^. #id) ^. (#getApiT . #unAddress)
                 walAddr `Expectations.shouldBe` genAddr
 
-    it "ANY_ADDRESS_POST_13 - at_least must make sense" $ \ctx -> do
-        forM_ ([0, 4, 333] :: [Int]) $ \atLeast -> do
-            let payload = Json [json|{
-                    "payment": {
-                        "some": {
-                            "from" : [
-                                "script_vkh1yf07000d4ml3ywd3d439kmwp07xzgv6p35cwx8h605jfx0dtd4a",
-                                "script_vkh1mwlngj4fcwegw53tdmyemfupen2758xwvudmcz9ap8cnqk7jmh4",
-                                "script_vkh1qw4l62k4203dllrk3dk3sfjpnh3gufhtrtm4qvtrvn4xjp5x5rt"
-                                ],
-                             "at_least": #{atLeast}
-                             }
-                        }
-                }|]
-            r <- request @AnyAddress ctx Link.postAnyAddress Default payload
-            expectResponseCode HTTP.status400 r
-            expectErrorMessage "must have at least one credential" r
-
     it "POST_ACCOUNT_01 - Can retrieve account public keys" $ \ctx -> runResourceT $ do
         let initPoolGap = 10
         w <- emptyWalletWith ctx ("Wallet", fixturePassphrase, initPoolGap)
--- a/lib/core/test/unit/Cardano/Wallet/Gen.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Gen.hs
@@ -1,8 +1,11 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
+
+{-# OPTIONS_GHC -fno-warn-orphans #-}
 
 -- | Shared QuickCheck generators for wallet types.
 --
@@ -21,12 +24,15 @@ module Cardano.Wallet.Gen
     , genTxMetadata
     , shrinkTxMetadata
     , genSmallTxMetadata
+    , genScript
     ) where
 
 import Prelude
 
 import Cardano.Address.Derivation
     ( xpubFromBytes )
+import Cardano.Address.Script
+    ( KeyHash (..), Script (..) )
 import Cardano.Api.Typed
     ( TxMetadata (..)
     , TxMetadataJsonSchema (..)
@@ -67,10 +73,14 @@ import Data.Word
     ( Word, Word32 )
 import GHC.TypeLits
     ( natVal )
+import Numeric.Natural
+    ( Natural )
 import Test.QuickCheck
     ( Arbitrary (..)
     , Gen
+    , Positive (..)
     , PrintableString (..)
+    , arbitrarySizedNatural
     , choose
     , elements
     , listOf
@@ -78,7 +88,9 @@ import Test.QuickCheck
     , oneof
     , resize
     , scale
+    , shrinkIntegral
     , shrinkList
+    , sized
     , suchThat
     , vector
     , vectorOf
@@ -90,6 +102,7 @@ import qualified Data.Aeson as Aeson
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Char8 as B8
 import qualified Data.HashMap.Strict as HM
+import qualified Data.List as L
 import qualified Data.Map as Map
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
@@ -259,3 +272,48 @@ shrinkTxMetadataValue (TxMetaList xs) =
 shrinkTxMetadataValue (TxMetaNumber i) = TxMetaNumber <$> shrink i
 shrinkTxMetadataValue (TxMetaBytes b) = TxMetaBytes <$> shrinkByteString b
 shrinkTxMetadataValue (TxMetaText s) = TxMetaText <$> shrinkText s
+
+instance Arbitrary Natural where
+    shrink = shrinkIntegral
+    arbitrary = arbitrarySizedNatural
+
+genScript :: [KeyHash] -> Gen (Script KeyHash)
+genScript keyHashes = scale (`div` 3) $ sized scriptTree
+    where
+        scriptTree 0 = oneof
+            [ RequireSignatureOf <$> elements keyHashes
+            , ActiveFromSlot <$> arbitrary
+            , ActiveUntilSlot <$> arbitrary
+            ]
+        scriptTree n = do
+            Positive m <- arbitrary
+            let n' = n `div` (m + 1)
+            scripts' <- vectorOf m (scriptTree n')
+            let hasTimelocks = \case
+                    ActiveFromSlot _ -> True
+                    ActiveUntilSlot _ -> True
+                    _ -> False
+            let scriptsWithValidTimelocks = case L.partition hasTimelocks scripts' of
+                    ([], rest) -> rest
+                    ([ActiveFromSlot s1, ActiveUntilSlot s2], rest) ->
+                        if s2 <= s1 then
+                            rest ++ [ActiveFromSlot s2, ActiveUntilSlot s1]
+                        else
+                            scripts'
+                    ([ActiveUntilSlot s2, ActiveFromSlot s1], rest) ->
+                        if s2 <= s1 then
+                            rest ++ [ActiveFromSlot s2, ActiveUntilSlot s1]
+                        else
+                            scripts'
+                    ([ActiveFromSlot _], _) -> scripts'
+                    ([ActiveUntilSlot _], _) -> scripts'
+                    (_,rest) -> rest
+            case fromIntegral (L.length (filter (not . hasTimelocks) scriptsWithValidTimelocks)) of
+                0 -> scriptTree 0
+                num -> do
+                    atLeast <- choose (1, num)
+                    elements
+                        [ RequireAllOf scriptsWithValidTimelocks
+                        , RequireAnyOf scriptsWithValidTimelocks
+                        , RequireSomeOf atLeast scriptsWithValidTimelocks
+                        ]
--- a/lib/core/test/unit/Cardano/Wallet/Gen.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Gen.hs
@@ -5,8 +5,6 @@
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 
-{-# OPTIONS_GHC -fno-warn-orphans #-}
-
 -- | Shared QuickCheck generators for wallet types.
 --
 -- Our convention is to let each test module define its own @Arbitrary@ orphans.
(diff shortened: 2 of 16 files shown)


stack.yaml
```diff
@@ -13,6 +13,8 @@ packages:
 - lib/text-class
 - lib/test-utils
 - lib/shelley
+- ../cardano-addresses/command-line
```
paweljakubas: this is interim. when cardano-addresses 3.2.0 lands it should be removed. I am doing this for the sake of local testing

lib/shelley/test/unit/Cardano/Wallet/Shelley/CompatibilitySpec.hs
```diff
@@ -71,6 +73,8 @@ import Cardano.Wallet.Unsafe
     ( unsafeMkEntropy )
 import Codec.Binary.Bech32.TH
     ( humanReadablePart )
+import Codec.Binary.Encoding
```
rvl, line 82: It's confusing for me that this module is actually in `cardano-addresses`.

lib/core/test/unit/Cardano/Wallet/Primitive/ScriptsSpec.hs
```diff
@@ -186,8 +173,11 @@ prop_scriptDiscoveredByTwo (TwoAccountXPubsWithScript accXPub' accXPub'' script)
     let sciptKeyHashes = retrieveAllVerKeyHashes script
     let scriptKeyHashes' = scriptKeyHashesInMap script accXPub' seqState'
     let scriptKeyHashes'' = scriptKeyHashesInMap script accXPub'' seqState''
-    (scriptKeyHashes' <> scriptKeyHashes'')
-        === Just (Set.fromList (L.nub sciptKeyHashes))
+    let expected =
+            if L.length sciptKeyHashes == 0 then
+                Nothing
+            else Just (Set.fromList (L.nub sciptKeyHashes))
```
rvl: The nub isn't needed when making a Set.

lib/core/test/unit/Cardano/Wallet/Primitive/ScriptsSpec.hs
```diff
@@ -173,7 +158,9 @@ prop_scriptsDiscovered (AccountXPubWithScripts accXPub' scripts') = do
     let seqState0 = initializeState accXPub'
     let seqState = foldr (\script s -> snd $ isShared script s) seqState0 scripts'
     let scriptHashes = Set.fromList $ Map.keys $ getKnownScripts seqState
-    scriptHashes === Set.fromList (map toScriptHash scripts')
+    let scriptsWithKeyHashes =
+            L.filter (\s -> L.length (retrieveAllVerKeyHashes s) > 0) scripts'
```
rvl: Why all the `L.` prefixes? Most list functions are in Prelude.
```suggestion
    let scriptsWithKeyHashes = filter (not . null . retrieveAllVerKeyHashes) scripts'
```

lib/core/test/unit/Cardano/Wallet/Gen.hs
```diff
@@ -259,3 +272,48 @@ shrinkTxMetadataValue (TxMetaList xs) =
 shrinkTxMetadataValue (TxMetaNumber i) = TxMetaNumber <$> shrink i
 shrinkTxMetadataValue (TxMetaBytes b) = TxMetaBytes <$> shrinkByteString b
 shrinkTxMetadataValue (TxMetaText s) = TxMetaText <$> shrinkText s
+
+instance Arbitrary Natural where
```
rvl: Best not to define this orphan instance and use the generator function directly.

lib/core/test/unit/Cardano/Wallet/Gen.hs
```diff
@@ -259,3 +272,48 @@ shrinkTxMetadataValue (TxMetaList xs) =
...
+            ]
+        scriptTree n = do
+            Positive m <- arbitrary
+            let n' = n `div` (m + 1)
+            scripts' <- vectorOf m (scriptTree n')
+            let hasTimelocks = \case
+                    ActiveFromSlot _ -> True
+                    ActiveUntilSlot _ -> True
+                    _ -> False
+            let scriptsWithValidTimelocks = case L.partition hasTimelocks scripts' of
```
rvl: So the goal is to prune scripts with invalid time locks?
Will this work if there are multiple ActiveFromSlot/ActiveUntilSlot in the generated scripts?
I can think of some nonsensical ranges...
```haskell
let redundant = RequireAnyOf [ActiveFromSlot 2, ActiveFromSlot 5]
    impossible = RequireAllOf [ RequireAnyOf [ActiveFromSlot 5], RequireAnyOf [ActiveFromSlot 2] ]
```

lib/shelley/test/unit/Cardano/Wallet/Shelley/CompatibilitySpec.hs
```diff
@@ -309,6 +316,159 @@ spec = do
...
+    -> Script KeyHash
+    -> Cardano.Script lang
+    -> SpecWith ()
+checkScriptHashes title adrestiaScript nodeScript = it title $
+    (unScriptHash $ toScriptHash adrestiaScript) `shouldBe`
+    (Cardano.serialiseToRawBytes $ Cardano.hashScript nodeScript)
+
+testScriptsAllLangs
+    :: forall lang . Cardano.SimpleScriptVersion lang
+    -> SpecM () ()
```
rvl:
```suggestion
    -> Spec
```

lib/shelley/test/unit/Cardano/Wallet/Shelley/CompatibilitySpec.hs
```diff
@@ -309,6 +316,159 @@ spec = do
...
+testScriptsAllLangs version = do
```
rvl:
```suggestion
testScriptsAllLangs version = describe (show version) $ do
```

lib/shelley/test/unit/Cardano/Wallet/Shelley/CompatibilitySpec.hs
```diff
@@ -309,6 +316,159 @@ spec = do
...
+toPaymentHash txt =
+    case Cardano.deserialiseFromRawBytesHex (Cardano.AsHash Cardano.AsPaymentKey) (T.encodeUtf8 txt) of
+        Just payKeyHash -> Cardano.RequireSignature payKeyHash
+        Nothing -> error "Hash key not valid"
+
+checkScriptHashes
+    :: String
+    -> Script KeyHash
+    -> Cardano.Script lang
+    -> SpecWith ()
```
rvl, line 352:
```suggestion
    -> Spec
```

lib/shelley/test/unit/Cardano/Wallet/Shelley/CompatibilitySpec.hs
```diff
@@ -309,6 +316,159 @@ spec = do
...
+                                       , Cardano.RequireAnyOf [toPaymentHash hashKeyTxt2, toPaymentHash hashKeyTxt3, toPaymentHash hashKeyTxt4]
+                                       ]
+              )
+            , ( show version <> " nested 3"
+              , RequireSomeOf 1 [ toKeyHash hashKeyTxt1
+                                , RequireAllOf [ toKeyHash hashKeyTxt2
+                                               , RequireAnyOf [toKeyHash hashKeyTxt3, toKeyHash hashKeyTxt4 ]
+                                               ]
+                                ]
+              , toSimpleScript $
```
rvl: These unit tests look good. Could the hspec title be changed to say that they are golden tests?

lib/core/test/unit/Cardano/Wallet/Primitive/ScriptsSpec.hs
```diff
@@ -208,7 +198,8 @@ prop_markingDiscoveredVerKeys (AccountXPubWithScripts accXPub' scripts') = do
 prop_poolExtension
     :: AccountXPubWithScriptExtension
     -> Property
-prop_poolExtension (AccountXPubWithScriptExtension accXPub' scripts') = do
+prop_poolExtension (AccountXPubWithScriptExtension accXPub' scripts') =
+    all (\s -> L.length (retrieveAllVerKeyHashes s) > 0)  scripts' ==>
```
rvl:
```suggestion
    all (not . null . retrieveAllVerKeyHashes) scripts' ==>
```
What's the `==>` for?

specifications/api/swagger.yaml
```diff
@@ -387,6 +387,26 @@ x-ScriptValue: &ScriptValue
...
+
+    - title: Active until
+      type: object
+      required:
+        - active_until
+      properties:
+        active_until:
+          description: Transaction has to occur before a specified slot number.
+          type: integer
+          minimum: 0
```
rvl: In the ledger code, these are called `RequireTimeStart` and `RequireTimeExpire`. Any reason to use a different term?

```suggestion
    - title: Active from slot
      type: object
      required:
        - active_from
      properties:
        active_from:
          description: Transaction is only valid starting at the specified slot number (`≥ active_from`).
          type: integer
          minimum: 0

    - title: Active until slot
      type: object
      required:
        - active_until
      properties:
        active_until:
          description: Transaction is only valid before the specified slot number (`< active_until`).
          type: integer
          minimum: 0
```

rvl (commented):
:+1:

lib/shelley/test/unit/Cardano/Wallet/Shelley/CompatibilitySpec.hs
paweljakubas: agree here, what I was missing something in cardano-addresses was ToText, FromText...I think we should put all reusable building blocks, like Encoding, FromText, ToText, .... in separate repo like adrestia-base.... or cardano-basic

lib/core/test/unit/Cardano/Wallet/Gen.hs
paweljakubas: this is ok to construct but will not pass validation from cardano-addresses (there is required and recommneded one)

lib/core/test/unit/Cardano/Wallet/Primitive/ScriptsSpec.hs
paweljakubas: I require left side to be valid, ie., script with any ver kyes, otherwise this property does not make sense

specifications/api/swagger.yaml
paweljakubas: some time ago we thought active_from and active _until are somewhat better, easier ....

paweljakubas:
added also golden tests for `preimage` and worked on better reuse

specifications/api/swagger.yaml
rvl: OK yes they are better names.

lib/core/test/unit/Cardano/Wallet/Gen.hs
```diff
@@ -259,3 +270,47 @@ shrinkTxMetadataValue (TxMetaList xs) =
...
+
+genScript :: [KeyHash] -> Gen (Script KeyHash)
+genScript keyHashes = scale (`div` 3) $ sized scriptTree
+    where
+        scriptTree 0 = oneof
+            [ RequireSignatureOf <$> elements keyHashes
+            , ActiveFromSlot <$> genNatural
+            , ActiveUntilSlot <$> genNatural
+            ]
+        scriptTree n = do
```
KtorZ, line 283: Shouldn't the generator be more agnostic to the script underlying structure and simply generate truly arbitrary scripts. Some of them may contain invalid timelocks, which is fine. We can use quickcheck labels to measure coverage and make sure that we cover interesting scenarios. 

The risk of having a generator that is _too complex_ and too close to what the actual implementation expects is that, we may end up introducing bugs present in the implementation in the generator itself and, fail to catch wrong behaviors from the implementation. As far as I am concerned, the generator for a script could be `genericArbitrary` and would not need a lot of tweaking. This may generate scripts that can be considered "invalid" by our functions yet, such scripts are still plausible and should therefore be part of the generated set of inputs. 

If some properties only apply to wellformed scripts then it should be possible to discard wrong cases with a pre-condition `==>`

paweljakubas: thought some more about, experimented with things like : 
```
genScriptValidated :: [KeyHash] -> Gen (Script KeyHash)                                                                                                                                                        
genScriptValidated keyHashes =                                                                                                                                                                                 
    genScript keyHashes `suchThat` (isRight . validateScript RecommendedValidation)
```
which slowed generation considerably.
In the end I went with the simplest solution - maximum possible simplification

lib/shelley/test/unit/Cardano/Wallet/Shelley/CompatibilitySpec.hs
```diff
@@ -309,6 +322,190 @@ spec = do
...
+    -> SpecWith ()
+checkScriptPreimage title adrestiaScript nodeScript = it title $
+    (serializeScript adrestiaScript) `shouldBe`
+    BS.append "\00" (Cardano.serialiseToCBOR nodeScript)
+
+scriptMatrix
+    :: Cardano.SimpleScriptVersion lang
+    -> [(String, Script KeyHash, Cardano.Script lang)]
+scriptMatrix version =
+    [ ( show version <> " RequireSignatureOf"
```
rvl, line 371: We can probably drop the `show version` bit now because the `describe` already has it.

lib/core/test/unit/Cardano/Wallet/Gen.hs
rvl: Simple is good. It looks simpler now.
It's also possible to add some cover tests to ensure the generator makes some "interesting" valid test cases.

With any recursive type, I'm always a bit worried about the generator making things too big, or the shrinker being too slow... can you reassure me that this generator is going to be nice to us in future?

rvl (approved):
:tada:

rvl:
bors r+

Please

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
```diff
@@ -634,24 +634,6 @@ spec = describe "SHELLEY_ADDRESSES" $ do
                 let walAddr = fst (addrs !! idx ^. #id) ^. (#getApiT . #unAddress)
                 walAddr `Expectations.shouldBe` genAddr
 
-    it "ANY_ADDRESS_POST_13 - at_least must make sense" $ \ctx -> do
```
piotr-iohk, line 637: Why at_least can have no sense now? :sweat_smile: @paweljakubas

paweljakubas: ah, ok. so this test and more will be resurrected later as we will introduce required and recommended option in POST AnyAddress. Now we have JSON-parse validation and those are valid scripts. But they will not be of course when validation is available

PR #2405, opened 2020-12-18
Additional checks for input existence in transactions

piotr-iohk:
# Issue Number

It turns out there are no checks for that.
A bit related to #2327.

# Overview

- 2605a9f1071aad41bca21d8ef98210c117f7cfb9
  Additional checks for inputs in incoming and outgoing transactions in integration tests
  
- 1247f9a75598a3b3f44d6b7631645868fac272a4
  Additional checks for inputs in join/quit pool transactions
  
- bd2916b0423811b85328edb60ae427f2e4b4dee4
  remove redundant cases & fix fixture wallet sharing common funding UTxOs
      It's a bit fucked up. Since the scenario is using fixtureWallet, the
    UtxO from the wallet comes from one of the fixture transaction made
    when initializing the cluster. There are many pre-funded fixture
    wallets with 10 UTxO, but in order not to spend 10 days funding them
    all, the cluster generate pretty large transaction with ~100
    outputs. So a single fixture transaction funds about 10 fixture
    wallets.

    Now it means that, a particular fixture wallet will know the value
    of all inputs of its funding fixture transaction and therefore, when
    picking two fixtureWallet one after the other, it may likely happen
    that both wallets are able to resolve inputs of any UTxO coming from
    their original fixture transaction.

    I fixed it by simply using fixtureWalletWith [...] which generates
    completly new and uncorrelated UTxOs.

- cdf556ac20da1ca7afe12c831a0876633f704317
  Add input check for multi-address pending tx

# Comments

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
@@ -22,6 +22,8 @@ import Cardano.Wallet.Api.Types
     , ApiStakePoolFlag (..)
     , ApiT (..)
     , ApiTransaction
+    , ApiTxId (..)
+    , ApiTxInput (..)
     , ApiWallet
     , ApiWalletDelegationStatus (..)
     , DecodeAddress
@@ -221,6 +223,8 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
             , expectField (#status . #getApiT) (`shouldBe` Pending)
             , expectField (#direction . #getApiT) (`shouldBe` Outgoing)
             , expectField #deposit (`shouldBe` Quantity 1000000)
+            , expectField #inputs $ \inputs' -> do
+                inputs' `shouldSatisfy` all (isJust . source)
             ]
         eventually "Wallet has joined pool and deposit info persists" $ do
             rJoin' <- request @(ApiTransaction n) ctx
@@ -232,6 +236,21 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
                 , expectField (#status . #getApiT) (`shouldBe` InLedger)
                 , expectField (#direction . #getApiT) (`shouldBe` Outgoing)
                 , expectField #deposit (`shouldBe` Quantity 1000000)
+                ]
+
+        let txId = getFromResponse #id rJoin
+        let link = Link.getTransaction @'Shelley src (ApiTxId txId)
+        eventually "delegation transaction is in ledger" $ do
+            rSrc <- request @(ApiTransaction n) ctx link Default Empty
+            verify rSrc
+                [ expectResponseCode HTTP.status200
+                -- all tx inputs have address and amount
+                , expectField #inputs $ \inputs' -> do
+                    let addrAmt = (fmap source inputs')
+                    addrAmt `shouldSatisfy` (all isJust)
+                , expectField (#direction . #getApiT) (`shouldBe` Outgoing)
+                , expectField (#status . #getApiT) (`shouldBe` InLedger)
+                , expectField (#metadata . #getApiTxMetadata) (`shouldBe` Nothing)
                 ]
 
         -- Earn rewards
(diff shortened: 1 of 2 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -215,12 +217,32 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
...
-        joinStakePool @n ctx pool (src, fixturePassphrase) >>= flip verify
+        rJoin <- joinStakePool @n ctx pool (src, fixturePassphrase)
+        verify rJoin
             [ expectResponseCode HTTP.status202
             , expectField (#status . #getApiT) (`shouldBe` Pending)
             , expectField (#direction . #getApiT) (`shouldBe` Outgoing)
+            -- all tx inputs have address and amount
+            , expectField #inputs $ \inputs' -> do
+                let addrAmt = (fmap source inputs')
+                addrAmt `shouldSatisfy` (all isJust)
```
piotr-iohk: Test fails here... it turns out that for delegation inputs are not there in the output of the join/quit transaction.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -289,17 +290,55 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
...
+            verify rDst
+                [ expectSuccess
+                , expectResponseCode HTTP.status200
+                , expectField (#amount . #getQuantity) $
+                    (`shouldBe` amt)
+                -- all tx inputs have NO address and amount
+                -- as the tx is incoming
+                , expectField #inputs $ \inputs' -> do
+                    let addrAmt2 = (fmap source inputs')
+                    addrAmt2 `shouldSatisfy` (all isNothing)
```
piotr-iohk: Fails here :thinking: ... inputs are there on incoming transaction on `fixtureWallet`

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -357,15 +396,55 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
...
+            verify rDst
+                [ expectSuccess
+                , expectResponseCode HTTP.status200
+                , expectField (#amount . #getQuantity) $
+                    (`shouldBe` (2 * amt) )
+                -- all tx inputs have NO address and amount
+                -- as the tx is incoming
+                , expectField #inputs $ \inputs' -> do
+                    let addrAmt2 = (fmap source inputs')
+                    addrAmt2 `shouldSatisfy` (all isNothing)
```
piotr-iohk: Doesn't fail here... The difference is that the destination wallet is `emptyWallet`.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
piotr-iohk: I can confirm this also on `shelley-test-cluster`. So it seems that:
 -  in case of transaction `fixtureWalletSrc` -> `fixtureWalletDest` there are resolved inputs visible on `incoming` transaction on `fixtureWalletDest`.
 - in case of transaction `emptyWallet` -> `fixtureWallet` there are NO resolved inputs visible on `incoming` transaction on `fixtureWallet`

Could this be due to fixtureWallet share the same key or something, so they are able to resolve the inputs in case of transaction between them?

KtorZ: Yeah it's a bit fucked up. Since the scenario is using `fixtureWallet`, the UtxO from the wallet comes from one of the fixture transaction made when initializing the cluster. There are many pre-funded fixture wallets with 10 UTxO, but in order not to spend 10 days funding them all, the cluster generate pretty large transaction with ~100 outputs. So a single fixture transaction funds about 10 fixture wallets. 

Now it means that, a particular fixture wallet will know the value of all inputs of its funding fixture transaction and therefore, when picking two fixtureWallet one after the other, it _may_ likely happen that both wallets are able to resolve inputs of any UTxO coming from their original fixture transaction. 

I fixed it by simply using `fixtureWalletWith [...]` which generates completly new and uncorrelated UTxOs.

KtorZ (approved):
Reviewed it, fixed it and removed some redundant cases :+1:

KtorZ:
bors merge

KtorZ:
bors merge

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
piotr-iohk: @KtorZ as far as your simplifications to the checks, I see you have removed this one and also the one from multi-output transactions. I'd argue that these are redundant, especially that one above was failing before (not because of `fixtureWallet` thing). It is OK now (after https://github.com/input-output-hk/cardano-wallet/pull/2370 I suppose) but still worth checking.

piotr-iohk: Whoops, sorry, you didn't remove that. It was the commit diff that mislead me :sweat_smile:
Anyway... I have added small additional check to multi-output tx in cdf556a.

PR #2411, opened 2020-12-31
Let integration tests use the CARDANO_WALLET_PORT variable

rvl:
While updating the [Testing](https://github.com/input-output-hk/cardano-wallet/wiki/Testing) wiki page, I thought it would be better if environment variables worked for all tests and benchmarks.

### DIFF
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
@@ -35,23 +35,31 @@ module Cardano.Wallet.Shelley.Launch
     , poolConfigsFromEnv
     , RunningNode (..)
 
-      -- * Faucets
-    , sendFaucetFundsTo
-    , moveInstantaneousRewardsTo
-    , oneMillionAda
-
-    -- * Utils
-    , NetworkConfiguration (..)
-    , nodeSocketOption
-    , networkConfigurationOption
-    , parseGenesisData
-    , withSystemTempDir
+      -- * Configuration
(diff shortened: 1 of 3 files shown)


lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
```diff
@@ -448,18 +450,17 @@ testMinSeverityFromEnv =
...
-        Right port -> pure $ ListenOnPort port
-        Left e -> die (show e)
+    Just (Right port) -> pure $ ListenOnPort port
+    Just (Left e) -> die $ show e
 
 -- | Directory for extra logging. Buildkite will set this environment variable
 -- and upload logs in it automatically.
 testLogDirFromEnv :: IO (Maybe FilePath)
-testLogDirFromEnv = traverse makeAbsolute =<< lookupEnv "TESTS_LOGDIR"
+testLogDirFromEnv = traverse makeAbsolute =<< lookupEnvNonEmpty "TESTS_LOGDIR"
```
hasufell, line 463: `makeAbsolute ""` equals `pwd`. Isn't that acceptable?

hasufell (approved):

lib/shelley/src/Cardano/Wallet/Shelley/Launch.hs
rvl: Not really - the normal tests log directory should be used when that variable is empty or unset.

rvl:
Thanks

bors r+

rvl:
bors retry

Issue #2296, opened 2020-11-05
[BUG] - autogenerated genesis file can't be parsed by cardano-wallet. Error parsing `activeSlotsCoeff`


jbax:
# Context

Using command `cardano-node shelley genesis create ...` the resulting `genesis.json` file comes out with the following:

```

{
    "activeSlotsCoeff": 5.0e-2,
    "protocolParams": {
        "poolDeposit": 0,
        "protocolVersion": {
            "minor": 0,
            "major": 0
        },
 ...
```

When running the `cardano-wallet` process against that genesis.json file, the following exception is raised:

```
[cardano-wallet.main:Error:4] [2020-11-05 04:16:01.54 UTC] Failed to parse genesis configuration. You may want to check the filepath given via --genesis and make sure it points to a valid JSON genesis file. Here's (perhaps) some helpful hint: GenesisDataParseError "(line 2, column 26):\nunexpected \".\"\nexpecting digit, white space, \",\" or \"}\""
```

It seems like the cardano-wallet is unable to parse this double value `5.0e-2` however the JSON is valid. 

**Expected behavior**
The cardano wallet should not raise any errors parsing the genesis.json file.

**System info :**
| Information  | -                                                     |
| ---          | ---                                                   |
| Version      | 2020.10.13 (git revision: 77e04fc0f0c226f27ea2e293666bb8311ba0daa1)  |
| Platform     | Ubuntu Linux 20.02          |
| Installation | From Github Release            |

rvl:
Hi @jbax, what are you trying to do here?

When running `cardano-wallet serve --testnet`, you need to provide the _Byron-era_ genesis json file for the testnet. Shelley genesis parameters are fetched from the chain.

The byron genesis file doesn't have activeSlotsCoeff, so it probably doesn't matter whether or not we can parse numbers in that format.

jbax:
Yeah I noticed that few minutes ago. Sorry for wasting your time.

Is there any way to start the wallet without a Byron genesis file? I got a
cardano node running locally with shelley genesis only.

KtorZ:
@jbax No. Not anymore.

jbax:
thanks!

vkconsult:
Ran into the same issue. The error message could be more specific in this came imo.

PR #2413, opened 2020-12-31
Clarify 'MsgFailedToParseGenesis' with regards to which genesis file is expected.

KtorZ:
# Issue Number

#2296 

# Overview

- 21da8b6c0ce6e24648c0b32e2150de77a2a7705e
  :round_pushpin: **clarify 'MsgFailedToParseGenesis' with regards to which genesis file is expected**
    We wrote that message a while ago when there was only Byron and therefore a single genesis file. With Shelley (and later Goguen), there's now an ambiguity when speaking about 'a genesis file' so it is better called out explicitly.

# Comments


rvl (approved):

KtorZ:
bors merge

Issue #2218, opened 2020-10-06
Flaky BYRON_ADDRESSES, ADDRESS_IMPORT_05 - I can import 15000 of addresses


Anviking:
# Context

https://github.com/input-output-hk/cardano-wallet/pull/2161#issuecomment-698634128

The `ADDRESS_IMPORT_05 - I can import 15000 of addresses` test can probably be expected to be somewhat slow. 

# Test Case

https://github.com/input-output-hk/cardano-wallet/blob/4ad068a00187df54b52ba1c88ed68b2b264866bf/lib/core-integration/src/Test/Integration/Scenario/API/Byron/Addresses.hs#L401-L433

# Failure / Counter-example

```
ci/hydra-build:required

 src/Test/Integration/Scenario/API/Byron/Addresses.hs:410:46:
 1) API Specifications, BYRON_ADDRESSES, ADDRESS_IMPORT_05 - I can import 15000 of addresses
 uncaught exception: IOException of type UserError
 user error (Waited longer than 90s (more than 2 epochs) for an action to resolve. Action: "Addresses are imported". Error condition: Just (HUnitFailure (Just (SrcLoc {srcLocPackage = "cardano-wallet-core-integration-2020.9.22-6DPlnrdhceJLXA8CsrMoUY", srcLocModule = "Test.Integration.Scenario.API.Byron.Addresses", srcLocFile = "src/Test/Integration/Scenario/API/Byron/Addresses.hs", srcLocStartLine = 430, srcLocStartCol = 13, srcLocEndLine = 430, srcLocEndCol = 35})) (ExpectedButGot Nothing "15000" "0")))

 To rerun use: --match "/API Specifications/BYRON_ADDRESSES/ADDRESS_IMPORT_05 - I can import 15000 of addresses/"
```

---

# Resolution 

---

# QA

Anviking:
The integration tests looks pretty simple, I'd guess the problem isn't there.

This should be the responsible function:
```haskell
importRandomAddresses
    :: forall ctx s k.
        ( HasDBLayer s k ctx
        , RndStateLike s
        , k ~ ByronKey
        )
    => ctx
    -> WalletId
    -> [Address]
    -> ExceptT ErrImportRandomAddress IO ()
importRandomAddresses ctx wid addrs = db & \DBLayer{..} -> mapExceptT atomically $ do
    cp <- withExceptT ErrImportAddrNoSuchWallet
[...]

PR #2416, opened 2020-12-31
Use a single-striped connection pool for each database layer instead of a single shared connection

KtorZ:
# Issue Number

ADP-586

# Overview

- 73df15f60162b40b7dbd7166287d6154426f2312
  :round_pushpin: **use a single-striped connection pool for each database layer**

  It is a rather common practice to use a pool of database connection
  when dealing with databases. So far, we've been using a single shared
  connection per wallet worker with, in front of each connection a lock
  preventing concurrent access to the database. The lock is only
  necessary because of the way persistent handles query statements
  internally, in principle, SQLite handles concurrent database accesses
  just well.

  For basic wallets, this is a relatively useless change. But for larger
  wallets like those manipulated by exchanges, we've observed very slow
  response time due to concurrent access of the database lock. Indeed,
  some requests may grab the lock for 10 or 20 seconds, preventing any
  requests from going throug. However, most requests are read-only
  requests and could be executed in parallel, at the discretion of
  the SQLite engine. I hope that the introduction of a connection pool
  will improve the overall experience for large wallets by better
  serving concurrent requests on the database. Finger crossed.

# Comments


lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -345,19 +360,38 @@ instance MatchMigrationError SqliteException where
...
-    conn <- Sqlite.open connStr
-    executeManualMigration migration conn
-    backend <- wrapConnectionInfo (mkSqliteConnectionInfo connStr) conn logFunc
-    pure (backend, conn)
+    traceWith tr $ MsgConnStr connStr
+
+    let createConnection = do
+            let info = mkSqliteConnectionInfo connStr
+            conn <- Sqlite.open connStr
+            executeManualMigration migration conn
```
KtorZ: Note that I originally executed the migration outside of the `createConnection` with a bracket open / close / executeMigration ... and it was mostly fine except for database views in the DB.Pool. The database property tests would fail with a `no such table: <name of the view>` as if the migration had not been executed. 

I have looked a bit in the documentation but didn't find anything that suggested that views are scoped to a single connection but it seems to be the case. At least, it should be safe running migration on every connection anyway... but still strange.

KtorZ:
bors try

lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -219,32 +191,75 @@ startSqliteBackend
...
         observe = bracketTracer (contramap MsgRun tr)
     -- runSqlConn is guarded with a lock because it's not threadsafe in general.
     -- It is also masked, so that the SqlBackend state is not corrupted if a
     -- thread gets cancelled while running a query.
     -- See: https://github.com/yesodweb/persistent/issues/981
+    --
+    -- Note that `withResource` does already mask async exception but only for
+    -- dealing with the pool resource acquisition. The action is then ran
+    -- unmasked with the acquired resource. If an asynchronous exception occurs,
+    -- the resource is NOT placed back in the pool.
```
rvl: Docs say that an exception of _any type_ means that the resource is not placed back in the pool.
This will be quite helpful for error recovery I think.

lib/core/src/Cardano/DB/Sqlite.hs
rvl: Hmm. It seems preferable for the migrations to be executed before other pool connections are allowed. Perhaps there needs to be a transaction commit or something?

lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -345,19 +360,38 @@ instance MatchMigrationError SqliteException where
...
+            pure (backend, conn)
+
+    let destroyConnection = \(backend, _) -> do
+            destroySqliteBackend tr backend fp
+
+    createPool
+        createConnection
+        destroyConnection
+        numberOfStripes
+        timeToLive
```
rvl, line 392: If the pool thread is cancelled (i.e. wallet is deleted or server needs to exit), those lingering pool connections are going to get cleaned up promptly, right?

lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -345,19 +360,38 @@ instance MatchMigrationError SqliteException where
...
+        maximumConnections
+  where
+    numberOfStripes = 1
+    timeToLive = 600 :: NominalDiffTime
+    -- When running in :memory:, we want a single connection that does not get
+    -- cleaned up.
+    maximumConnections = maybe 1 (const 10) fp
```
rvl: Is there some difference with sqlite in-memory databases which mean only one concurrent connection is possible? Otherwise, maximumConnections may as well be the same for both.

The number 10 seems fine for each wallet database. Perhaps in future, to prevent scaling bottlenecks, the maximum for the stake pools database could be scaled by the number of wallets.

rvl (approved):
Great! This will help a lot. I don't know why we didn't do this earlier, since resource-pool is so easy to use. I actually assumed that the API server got to use a different DB connection from the restore thread - oops, wrong.

lib/core/src/Cardano/DB/Sqlite.hs
KtorZ: I hope.

lib/core/src/Cardano/DB/Sqlite.hs
KtorZ: Yes. I swapped the `destroyDBLayer` function with:

```hs
destroyDBLayer :: Tracer IO DBLog -> SqliteContext -> IO ()
destroyDBLayer tr SqliteContext{connectionPool,dbFile} = do
    traceWith tr (MsgDestroyConnectionPool dbFile)
    destroyAllResources connectionPool
```

So this is done as part of the `withDBLayer` bracket. The same quirks as before with regards to closing a single SQLite connection still applies but `destroyConnection` now does exactly what `destroyDBLayer` was doing, that is, retry for some cherry-picked exceptions and handle already closed connections as well.

lib/core/src/Cardano/DB/Sqlite.hs
KtorZ: I originally wanted to use [createSqlitePool](https://hackage.haskell.org/package/persistent-sqlite-2.10.6.2/docs/Database-Persist-Sqlite.html#v:createSqlitePool) from `persistent` but it doesn't give any ways to access the underlying connection (only the persistent abstraction `SqlBackend`) which we do need in a few places for "hardcore error handling" and also custom migrations. 

The `createSqlitePool` function rightfully mentions:

> Note that this should not be used with the :memory: connection string, as the pool will regularly remove connections, destroying your database. Instead, use withSqliteConn.

I stole the default `10` max connections and `600` seconds time to live from the default chosen in persistent. Using `1` for in the in-memory database seems to work-around the issue nicely as resource-pool always keep at least one connection. Since this is only used in testing, I found it to be a simple and non-invasive resolution.

KtorZ:
>  I don't know why we didn't do this earlier

Yes. Sometimes we forget the basics.

KtorZ:
bors merge

KtorZ:
:point_up: looks pretty serious, it's full chaos in the integration cluster. My guess is the database reference counter not counting properly since the mutable counter only count `DBLayer` but not individual connections so we are back with race-condition with regards to database deletion.

lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -127,14 +137,26 @@ import qualified Database.Sqlite as Sqlite
 
 -- | Context for the SQLite 'DBLayer'.
 data SqliteContext = SqliteContext
-    { connectionPool :: Pool (SqlBackend, Sqlite.Connection)
+    { connectionPool :: Pool SqlBackend
     -- ^ A handle to the Persistent SQL backend.
+    , isDatabaseActive :: TVar Bool
```
rvl: A link to https://www.sqlite.org/wal.html#busy might help.

KtorZ:
bors merge

KtorZ:
Forgot to push the latest change :|

KtorZ:
bors try

lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -167,9 +179,82 @@ queryLogFunc tr _loc _source level str = traceWith tr (MsgQuery msg sev)
...
+    return $ case migrationResult of
+        Left e  -> Left e
+        Right{} ->
+            let observe :: IO a -> IO a
+                observe = bracketTracer (contramap MsgRun tr)
+
+               -- runSqlConn is guarded with a lock because it's not threadsafe in
+               -- general.It is also masked, so that the SqlBackend state is not
+               -- corrupted if a thread gets cancelled while running a query.
+               -- See: https://github.com/yesodweb/persistent/issues/981
```
rvl: Perhaps in future we could drop the lock and mask_?
1. If we are confident that any pool connection will only be used by one thread at a time, then the lock isn't needed.
2. If the thread is cancelled during a database operation, then so be it. SQLite will rollback the uncommitted transaction. The connection won't be returned to the pool.

lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -202,49 +290,33 @@ destroyDBLayer (SqliteContext {getSqlBackend, trace, dbFile}) = do
...
+--     process B will get back an SQLITE_BUSY result because SQLite only supports
+--     one writer at a time. Process B will need to wait for process A to finish
+--     its transaction before starting a new transaction. The sqlite3_busy_timeout()
+--     and sqlite3_busy_handler() interfaces and the busy_timeout pragma are
+--     available to process B to help it deal with SQLITE_BUSY errors.
+--
+retryOnBusy :: Tracer IO DBLog -> IO a -> IO a
+retryOnBusy tr action =
+    recovering policy (mkRetryHandler isBusy) $ \RetryStatus{rsIterNumber} -> do
+        when (rsIterNumber > 0) $ traceWith tr (MsgRetryOnBusy rsIterNumber)
```
rvl: I like that there's logging. With the resource pool improvements. it will be interesting to see if this error ever happens in tests.

rvl (approved):
Excellent, thanks for this!

If I run the integration tests with `NO_CLEANUP=1`, then ctrl-c them, there are no longer `-shm` and `-wal` files left behind in the tests temporary directory. :tada:

Even better, wallets which were supposed to be deleted, actually seem to have been deleted.

rvl:
Fixed a weeder error and rebased.

bors r+

lib/core/src/Cardano/DB/Sqlite.hs
KtorZ: (1) is only true if we can start the database in full_mutex I guess, which persistent doesn't really allow us to set that flag (although, we've forked it so it's easily added). This will not prevent retries however in case two writing threads are trying to write something at the same time, but at least it should give more concurrency to readers :thinking: Also "if we are confident that any pool connection will only be used by one thread at a time" ==> I have no such confidence at this stage unfortunately :( 

(2) Hmmm. That's right. Sounds scary though :scream:

KtorZ:
Note that I've seen `1) API Specifications, BYRON_ADDRESSES, ADDRESS_IMPORT_05 - I can import 15000 of addresses` a few times while I was working on this. It was almost 100% reproducible when I used a full-jitter backoff on `retryOnBusy` instead of a constant delay. I am not sure how SQLite prioritize threads but it seems that long transactions may be "stuck" for a while if there are many short transactions happening concurrently (as it is happening in our test cluster with the 0.2s slot length...)

rvl:
Interesting. :thinking: 

I noticed this flaky test issue on the same test case: #2218
But it is probably a case of different causes for the same result.

KtorZ:
The above issue regarding the import 15.000 random addresses is now tackled by #2425

KtorZ:
bors merge

KtorZ:
Will wait for #2425 before re-attempting to merge since it seems that this PR makes it "worse".

rvl:
I tried to batch #2425 with this one, but I guess it didn't work due to a merge conflict. That's resolved now.

KtorZ:
bors merge

Anviking:
bors try

lib/shelley/80-wallets-to-0.01.json
```diff
@@ -0,0 +1 @@
+{"walletOverview":{"transactions":0,"addresses":20,"utxo":{"distribution":{"10000000000000000":0,"100000000000000":0,"1000000000000":0,"10000000000":0,"100000000":0,"1000000":0,"10000":0,"100":0,"1000000000":0,"100000000000":0,"10":0,"45000000000000000":0,"1000":0,"10000000000000":0,"1000000000000000":0,"100000":0,"10000000":0},"scale":"log10","total":{"quantity":0,"unit":"lovelace"}}},"listAddressesTime":"81.07 ms","restoreTime":"112.0 s","benchName":"80-wallets-to-0.01","readWalletTime":"30.13 ms","listTransactionsTime":"47.64 ms","estimateFeesTime":"155.9 ms"}
```
KtorZ: Whoops :point_up:

Anviking: My bad, fixed.

Anviking:
bors try

Anviking:
🤔 I don't see this locally

bors try

KtorZ:
@Anviking I've been seeing this one quite a lot on this PR and I think the PR makes it worse for some reason (hence also why I haven't really pushed for merging this yet). I am afraid that merging it as it is would introduce more flakyness in the test suite.

Anviking:
@KtorZ Yeah. I'm looking at this a bit because @rvl suggested it might fix the restoration bench.

And now there's some results: https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/823

There's quite a lot of
```
[31m[ci4-benc:cardano.node.LocalErrorPolicy:Error:58][0m [2021-01-25 13:22:52.86 UTC] IP LocalAddress "" ErrorPolicyUnhandledApplicationException (MuxError MuxBearerClosed "<socket: 429> closed when reading data, waiting on next header True")
[31m[ci4-benc:cardano.node.LocalErrorPolicy:Error:58][0m [2021-01-25 13:22:52.86 UTC] IP LocalAddress "" ErrorPolicyUnhandledApplicationException (MuxError MuxBearerClosed "<socket: 431> closed when reading data, waiting on next header True")
```
so sadly not it looks like. Or not on its own at least.

There's some new weird logs in the beginning too:
```
restore: SQLite3 returned ErrorError while attempting to perform prepare "SELECT \"wallet_id\", \"slot\", \"header_hash\", \"parent_header_hash\", \"block_height\" FROM \"checkpoint\" WHERE (\"wallet_id\"=?) ORDER BY \"slot\"": no such table: checkpoint
[31m[ci4-benc:cardano.node.LocalErrorPolicy:Error:58][0m [2021-01-25 12:56:03.27 UTC] IP LocalAddress "" ErrorPolicyUnhandledApplicationException (MuxError MuxBearerClosed "<socket: 216> closed when reading data, waiting on next header True")
```
maybe this could be related to the failing ADDRESS_IMPORT_05 (I think the fix for ADDRESS_IMPORT_05 changed that table)

rvl:
MuxBearerClosed means that the node client (i.e. wallet) has exited.

`no such table: checkpoint` means that the dblayer is not set up correctly in the benchmark.

I don't think either would explain the ADDRESS_IMPORT_05 failure.

I am optimistic that this change will help with StatementAlreadyFinalized "BEGIN" errors, once we figure out the above problems.

KtorZ:
`no such table: checkpoint` was symptomatic of a connection running queries before the database was fully migrated; I bumped into it a few times when I was developing the feature. Though, on the latest version, the connection pool should only be available _after_ migrations have been executed, that's true for code using the `DBFactory` and `withDBLayer`, although a few places were still creating database connection through different means... I've updated some unit tests already to use `withDBLayer` only. I couldn't do the same for the database benchmark for the bracket-style doesn't work well with criterion. For the restoration nightly, I don't remember how it is setup anymore; if it's not using `withDBLayer`, then it probably should.

Anviking:
I tried creating 50 wallets using this PR on mainnet, and it consistently crashes (when trying to start it again):
```
[cardano-wallet.wallet-engine:Error:651] [2021-01-25 15:59:27.25 UTC] Worker has exited unexpectedly: SQLite3 returned ErrorCan'tOpen while attempting to perform open "wallet-db4/she.0966a286934e3aa1d33f1931753fdb881e03c36b.sqlite".
[cardano-wallet.wallet-engine:Info:95] [2021-01-25 15:59:27.25 UTC] a63cb014: Applying blocks [90031 ... 91030]
[cardano-wallet.wallet-engine:Info:95] [2021-01-25 15:59:27.26 UTC] a63cb014: Creating checkpoint at 2da6f1c1<-[0ae9af1e-91030#91001]
[cardano-wallet.wallet-engine:Info:95] [2021-01-25 15:59:27.27 UTC] a63cb014: w27140, created at 2021-01-25 15:47:16 UTC, delegating to ∅
[cardano-wallet.wallet-engine:Info:95] [2021-01-25 15:59:27.27 UTC] a63cb014: syncProgress: still restoring (1.73%)
[cardano-wallet.wallet-engine:Info:95] [2021-01-25 15:59:27.27 UTC] a63cb014: discovered 0 new transaction(s)
[cardano-wallet.wallet-engine:Info:95] [2021-01-25 15:59:27.27 UTC] a63cb014: local tip: 2da6f1c1<-[0ae9af1e-91030#91001]
[cardano-wallet.main:Debug:4] [2021-01-25 15:59:27.27 UTC] Logging shutdown.
cardano-wallet-pool: SQLite3 returned ErrorCan'tOpen while attempting to perform open "wallet-db4/stake-pools.sqlite".
```
it works fine on `master`.

Anviking:
Hm… `ulimit -n 5000` made it work again. So might just be an increase in file-handle-usage.

rvl:
What's your default ulimit?
I would hope that a wallet restore only needs 1 active connection (i.e. 3 fds) in the connection pool for that wallet.

lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -423,13 +425,16 @@ instance ToText FollowLog where
...
+            let slot = pretty . slotNo
+                buildRange (x :| []) = x
+                buildRange xs = NE.head xs <> ".." <> NE.last xs
+                blockHeights = pretty . getQuantity . blockHeight <$> hdrs
             in mconcat
-                [ "Applying blocks [", pretty slFst, " ... ", pretty slLst, "]" ]
+                [ "Applying block numbers [", buildRange blockHeights, "]"
+                , "  Wallet/node slots: ", slot (NE.last hdrs)
+                , "/", slot tipHdr
+                ]
```
KtorZ, line 437: :grimacing: Good.

rvl:
I think checkpoint caching is related to the failures of ADDRESS_IMPORT_05 - but still looking into it.
Due to the connection pool a MVar lock on runQuery is no longer needed, but the cache is still using IORefs.

KtorZ:
Well yeah, the cache was using an IORef because of the guaranteed given by the global MVar lock on queries. If we drop that guarantee, the IORef becomes unsound and we should switch to a TVar instead.

rvl:
Switching IORef to TVar (or MVar) is not going to magically make it work. The problem is that it has readCache and writeCache instead of modifyCache.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1126,39 +1125,33 @@ newDBLayerWith cacheBehavior ti SqliteContext{runQuery} = do
...
-                    in Map.alter alter wid
-
-    let selectLatestCheckpoint
+        writeCache :: W.WalletId -> Maybe (W.Wallet s) -> SqlPersistT IO ()
+        writeCache wid = maybeUpdateCache . flip Map.alter wid . maybe (const Nothing) alterCache
+
+        alterCache :: W.Wallet s -> (Maybe (W.Wallet s) -> Maybe (W.Wallet s))
+        alterCache cp = \case
+            -- this seems suspicious
+            Just old | getHeight cp < getHeight old -> Just old
```
KtorZ, line 1308: This is mainly there because of the state-machine tests. In the application, we do in practice never have the case where we try to write a checkpoint that is older than the latest checkpoint written because we process them in sequence from genesis to the tip, and we drop checkpoints upon rolling back. 

However, the state-machine tests aren't really respecting any of this and may insert checkpoints in any arbitrary order. That little guard is here to cope with this and enforce the invariant that the cache always contains the _most recent checkpoint_, not in terms of insertion order in the database but in terms of blockheight on the chain. 

I realize that this deserves a proper comment note...

rvl:
I had another look at the "expected: 15000 addresses but got: 0" failure.
Basically there needs to be a mutex around the `DBLayer->atomically` function, otherwise a cache like this checkpoint cache cannot work.
Such a mutex would defeat the main purpose of a connection pool, which is to allow concurrent queries.
I think the best way to fix it would be to remove the cache and adjust the DBLayer so that wallet state can be updated incrementally.

rvl:
I put a mutex around runQuery, so the checkpoint caching will work now.
There will be a connection pool, but no query concurrency for the wallet database.
I hope we can merge this now @KtorZ , and work on removing the mutex later.

KtorZ:
>  There will be a connection pool, but no query concurrency for the wallet database.

What's the point of the connection pool then :sweat_smile: ? Beside maybe the various resource management improvement which also comes with the PR .. they are worth it in their own.

rvl:
Yes a query mutex will ensure that there is at most one connection in the pool for each wallet database - that's not much of a connection pool.
It may be a bit more effort to make it work without the mutex.
So let's merge this initial work now to avoid more merge conflicts.

rvl:
Rebased and fixed merge conflicts again....

bors r+

rvl:
Please

bors r+

PR #2417, opened 2021-01-02
Review worker error handling

rvl:
### Issue number

ADP-641

### Overview

With `MonadUnliftIO` it becomes easier to write polymorphic functions (`MonadIO` is not enough when there are callbacks).

So we can make the worker thread registry polymorphic, and run it in `IO` or `ExceptT e IO`.

That removes the need for using the `unsafeRunExceptT` function.

Although the register function can now work with checked exception types, it is also modified to be aware of IO exceptions which may be thrown during setup. These other exceptions are passed up to the caller.

The worker registry changes also fix a race condition where multiple worker threads could be started for the for the same key, before the worker resource is acquired.

### Comments

Three questions that I have:

1. What to do if the integrity check for only one of the wallets fails at startup. Carry on with just the working wallets? Or just fail?
   ⇒ A: Start the server anyway (previously it would fail at startup).
2. What to do if a wallet worker crashes. Should it be restarted? Should the server exit?
   ⇒ A: Restarts are handled elsewhere, although are probably not even necessary. Any other exceptions are truly unexpected. Maintain current behaviour.
3. I am not sure whether the `instance MonadUnliftIO (ExceptT e m)` is generally safe to use.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -93,6 +93,7 @@ module Cardano.Wallet.Api.Server
     -- * Server error responses
     , IsServerError(..)
     , liftHandler
+    , throwE
     , apiError
 
     -- * Internals
@@ -121,7 +122,11 @@ import Cardano.Address.Script
 import Cardano.Api
     ( AnyCardanoEra (..), CardanoEra (..), SerialiseAsCBOR (..) )
 import Cardano.BM.Tracing
-    ( HasPrivacyAnnotation (..), HasSeverityAnnotation (..) )
+    ( HasPrivacyAnnotation (..)
+    , HasSeverityAnnotation (..)
+    , PrivacyAnnotation (..)
+    , Severity (..)
+    )
 import Cardano.Mnemonic
     ( SomeMnemonic )
 import Cardano.Wallet
@@ -417,12 +422,7 @@ import Cardano.Wallet.Primitive.Types.Tx
     , txOutCoin
     )
 import Cardano.Wallet.Registry
-    ( HasWorkerCtx (..)
-    , MkWorker (..)
-    , WorkerLog (..)
-    , defaultWorkerAfter
-    , workerResource
-    )
+    ( HasWorkerCtx (..), MkWorker (..), WorkerLog (..), defaultWorkerAfter )
 import Cardano.Wallet.TokenMetadata
     ( TokenMetadataClient, fillMetadata )
 import Cardano.Wallet.Transaction
@@ -435,8 +435,6 @@ import Cardano.Wallet.Transaction
     , Withdrawal (..)
     , defaultTransactionCtx
     )
-import Cardano.Wallet.Unsafe
-    ( unsafeRunExceptT )
 import Control.Arrow
     ( second )
 import Control.DeepSeq
@@ -444,15 +442,17 @@ import Control.DeepSeq
 import Control.Error.Util
     ( failWith )
 import Control.Monad
-    ( forM, forever, join, void, when, (>=>) )
-import Control.Monad.IO.Class
-    ( MonadIO, liftIO )
+    ( forM, forM_, forever, join, unless, void, when, (>=>) )
+import Control.Monad.IO.Unlift
+    ( MonadIO, MonadUnliftIO (..), liftIO )
+import Control.Monad.Trans.Class
+    ( lift )
 import Control.Monad.Trans.Except
     ( ExceptT (..), runExceptT, throwE, withExceptT )
 import Control.Monad.Trans.Maybe
     ( MaybeT (..), exceptToMaybeT )
 import Control.Tracer
-    ( Tracer, contramap )
+    ( Tracer, contramap, traceWith )
 import Crypto.Hash.Utils
     ( blake2b224 )
 import Data.Aeson
@@ -463,12 +463,14 @@ import Data.ByteString
     ( ByteString )
 import Data.Coerce
     ( coerce )
+import Data.Either.Combinators
+    ( whenLeft )
 import Data.Either.Extra
     ( eitherToMaybe )
 import Data.Function
     ( (&) )
 import Data.Functor
-    ( (<&>) )
+    ( ($>), (<&>) )
 import Data.Functor.Identity
     ( Identity (..) )
 import Data.Generics.Internal.VL.Lens
@@ -552,10 +554,12 @@ import Type.Reflection
     ( Typeable, typeRep )
 import UnliftIO.Async
     ( race_ )
+import UnliftIO.Compat
+    ( unliftIOWith )
 import UnliftIO.Concurrent
     ( threadDelay )
 import UnliftIO.Exception
-    ( IOException, bracket, throwIO, tryAnyDeep, tryJust )
+    ( IOException, bracket, tryAnyDeep, tryJust )
 
 import qualified Cardano.Wallet as W
 import qualified Cardano.Wallet.Api.Types as Api
@@ -2992,13 +2996,20 @@ newApiLayer
     -> IO ctx
 newApiLayer tr g0 nw tl df tokenMeta coworker = do
     re <- Registry.empty
-    let trTx = contramap MsgSubmitSealedTx tr
-    let trW = contramap MsgWalletWorker tr
     let ctx = ApiLayer trTx trW g0 nw tl df re tokenMeta
-    listDatabases df >>= mapM_ (startWalletWorker ctx coworker)
-    return ctx
+    wallets <- listDatabases df
+    forM_ wallets $ \wid -> do
+        res <- runExceptT $ startWalletWorker ctx coworker wid
+        -- Workers which failed to start are logged, but the server will
+        -- continue regardless. API requests which use these wallets will fail.
+        whenLeft res $ traceWith tr . MsgWalletFailedToRegister
+    pure ctx
+  where
+    trTx = contramap MsgSubmitSealedTx tr
+    trW = contramap MsgWalletWorker tr
 
 -- | Register a wallet restoration thread with the worker registry.
+-- This is called at startup for each wallet.
 startWalletWorker
     :: forall ctx s k.
         ( ctx ~ ApiLayer s k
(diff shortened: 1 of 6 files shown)


rvl:
bors try

KtorZ:
@rvl 

> (1) What to do if the integrity check for only one of the wallets fails at startup. Carry on with just the working wallets? Or just fail?

So far, the strategy has been the former. Leave the dead wallet in a "dead" state. When queried, it should trigger a corresponding `WalletNotResponding` error.

> (2) What to do if a wallet worker crashes. Should it be restarted? Should the server exit?

Same as (1), that worker should die and be reported as "NotResponding". So far, we have been restarting worker when the `follow` function generated an exception. The main reason to do this is in case the connection with the node is lost and because of network exceptions (I recall this was only necessary for Jormungandr). The "connection lost" errors are already captured in the network layer which actually catch the exceptions and retries. So in principle it doesn't generate any exceptions anymore. 

I don't see any good reason for retrying anymore :thinking:, and even if there's one particular exception we would want to handle; we could bring it as a first-class value inside the `ExceptT` and let any exception "crash" the worker. Although arguably, most async exceptions would come from normal termination and shouldn't lead to an error but rather to a graceful shutdown.


> (3) I am not sure whether the instance MonadUnliftIO (ExceptT e m) is generally safe to use. Perhaps the instance should only be defined for the servant Handler type.

That's a good question. I imagine that it acts similarly as an instance on `MonadUnliftIO m` as it seems to be the strategy `ExceptT` uses for MonadThrow and MonadCatch already.

rvl:
Thanks @KtorZ - sounds good to me.
I have modified error handling in worker init (1).
With (2), my hope is that with the `UnliftIO.Exception` change, in the event of async exceptions, the clean up actions will never fail to be run completely.

hasufell:
What's the related ticket?

KtorZ:
@hasufell https://jira.iohk.io/browse/ADP-555. We agreed two iterations meetings ago to have Rodney, Johannes and myself address issues with regards to the wallet overall stability to help reducing CI pain.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2197,10 +2194,17 @@ newApiLayer
...
     re <- Registry.empty
     let ctx = ApiLayer tr g0 nw tl df re
-    listDatabases df >>= mapM_ (registerWorker ctx coworker)
+    wallets <- listDatabases df
+    forM_ wallets $ \wid -> do
+        res <- runExceptT $ registerWorker ctx coworker wid
+        -- Workers which failed to start are logged, but the server will
+        -- continue regardless. API requests which use these wallets will fail.
+        whenLeft res $
+            traceWith tr . MsgFromWorker wid . W.MsgIntegrityCheckFailed
```
KtorZ: :+1:

lib/core/src/Cardano/Wallet/Registry.hs
```diff
@@ -106,42 +112,75 @@ newtype WorkerRegistry key resource =
...
     => WorkerRegistry key resource
     -> key
     -> m (Maybe (Worker key resource))
 lookup (WorkerRegistry mvar) k =
-    liftIO (Map.lookup k <$> readMVar mvar)
+    Map.lookup k <$> readMVar mvar
+
+-- | Lookup the resource for a given worker. If the worker is registered, but
+-- not yet initialized, this will block until the resource is acquired and the
+-- 'workerBefore' action is complete.
```
KtorZ, line 133: :+1:

lib/core/src/Cardano/Wallet/Registry.hs
```diff
@@ -217,31 +261,29 @@ register
...
-    cleanup resourceVar result = do
+            workerBefore ctx' k
+            putMVar (workerResource worker) resource
+            putMVar setupVar Nothing
+            workerMain ctx' k
+    threadId <- work `forkFinally` \result -> do
         void $ registry `delete` k
-        void $ tryPutMVar resourceVar Nothing
-        after tr result
+        void $ tryPutMVar setupVar (either Just (const Nothing) result)
```
KtorZ, line 277: This means that we'll only rethrow exceptions that occurs _before_ `workerMain`, since after that the `setupVar` is no longer empty. Rethrowing such exception should be caught by Servant in the API handler and result in a 500 internal error which seems _valid_. The exception is then logged as part of `workerAfter`. 

I think this is okay :+1: , I'll play around with it a bit -- manually throwing exceptions here and there.

lib/core/src/UnliftIO/Compat.hs
```diff
@@ -61,3 +76,42 @@ handleIf
...
+    -> ((a -> m b) -> m b)
+unliftIOWith with action = do
+    u <- askRunInIO
+    liftIO (with (u . action))
+
+-- | Provides a Tracer in IO given a 'Tracer m'.
+unliftIOTracer :: MonadUnliftIO m => Tracer m a -> m (Tracer IO a)
+unliftIOTracer tr = do
+    u <- askRunInIO
+    pure $ Tracer $ \a -> u $ runTracer tr a
```
KtorZ, line 117:
```suggestion
    pure $ natTracer u tr
```

lib/core/src/UnliftIO/Compat.hs
```diff
@@ -61,3 +76,42 @@ handleIf
...
+        exceptToIO
+          (runInIO . (either (throwIO . InternalException) pure <=< runExceptT))
+
+-- | Wrapper for Left results of runExceptT. Used to deliver return values in
+-- the MonadUnliftIO instance of ExceptT.
+newtype InternalException e = InternalException { unInternalException :: e }
+    deriving Typeable
+instance Show (InternalException e) where
+    show _ = "MonadUnliftIO InternalException"
+instance Typeable e => Exception (InternalException e)
```
KtorZ, line 97: I wonder if we aren't going to shoot ourselves in the foot with this... We catch `SomeException` in a few places and this instances introduces a quite confusing behavior... I wonder if we couldn't leverage the fact that:

a) `m` will always be `IO` for us. 
b) Async exception aren't caught by unliftio functions. 

to write it as:

```hs
instance (MonadUnliftIO m, MonadIO m, Typeable e, Show e) => MonadUnliftIO (ExceptT e m) where
    withRunInIO exceptToIO =
        unSmuggleExceptT $ ExceptT $
            withRunInIO $ \runInIO -> Control.Exception.try $
                exceptToIO (runInIO . smuggleExceptT . runExceptT)
      where
        smuggleExceptT :: m (Either e a) -> m a
        smuggleExceptT = (=<<) $ either
            (liftIO . Control.Exception.throwIO . SomeAsyncException . InternalException)
            pure

        unSmuggleExceptT :: ExceptT SomeAsyncException m a -> ExceptT e m a
        unSmuggleExceptT = mapExceptT $ \inner -> inner >>= \case
            Right a -> return $ Right a
            Left asyncE@(SomeAsyncException e) ->
                case toInternalException e of
                    Just (InternalException e) -> return (Left e)
                    Nothing -> liftIO $ Control.Exception.throwIO asyncE

        toInternalException :: Exception any => any -> Maybe (InternalException e)
        toInternalException =
            Control.Exception.fromException . Control.Exception.toException
```

This way, we properly handle the ExceptT checked exception and other exceptions:

```hs
-- | Left MyException
case1 :: ExceptT MyException IO String
case1 = handle handler $ do
    throwE MyException
  where
    handler (_e :: MyException) = return "Caught"

-- | Left MyException
case2 :: ExceptT MyException IO String
case2 = handleAny handler $ do
    throwE MyException
  where
    handler (_e :: SomeException) = return "Caught"

-- | Right "Caught"
case3 :: ExceptT MyException IO String
case3 = handle handler $ do
    liftIO $ throwIO MyException
  where
    handler (_e :: MyException) = return "Caught"

-- | Right "Caught"
case4 :: ExceptT MyException IO String
case4 = handleAny handler $ do
    liftIO $ throwIO MyException
  where
    handler (_e :: SomeException) = return "Caught"
```

Otherwise, `case2` would return a `Right "Caught"` as well which is misleading.

lib/shelley/test/integration/Main.hs
```diff
@@ -371,7 +373,9 @@ instance HasSeverityAnnotation TestsLog where
         MsgBaseUrl {} -> Notice
         MsgCluster msg -> getSeverityAnnotation msg
         MsgPoolGarbageCollectionEvent _ -> Info
-        MsgServerError{} -> Critical
+        MsgServerError e
+            | isAsyncException e -> Info
+            | otherwise -> Critical
```
KtorZ: :+1:

lib/core/src/UnliftIO/Compat.hs
KtorZ: I don't think this instance is safe at all or behave as we would expect. For instance:

```hs
main :: IO ()
    print =<< runExceptT case1
    print =<< runExceptT (askUnliftIO >>= \u -> liftIO (unliftIO u case1))
```
  
This prints:

```
Left MyException
Main.hs: InternalException MyException
```

Where I'd expect the same behavior. Though, stepping back a bit, what would it even mean to run an ExceptT in IO? It doesn't make sense in the end because there's no way that we can capture the context regarding the checked exception in IO only. So, what would it mean to run `throwE MyException` in IO? At best it means "throw the exception", whereas the semantic given by ExceptT would be more "return a Left". But we can interpret "return a Left" is mere IO because we can only express `IO a`. 

We would need a generalized definition of `MonadUnliftIO` to make it work properly, something like:


```hs
class Monad m => MonadUnliftIO m where
    type F m :: * -> *
    
    withRunInIO :: ((forall a. m a -> IO (F m a)) -> IO (F m b)) -> m (F m b) 
```

And then define:

```hs
instance MonadUnliftIO m => MonadUnliftIO (ReaderT r m a) where
   type F (ReaderT r m) = Identity
   
instance MonadUnliftIO m => MonadUnliftIO (ExceptT e m a) where
   type F (ExceptT e m) = Either e
```

Without that, I don't see how we can write a sound instance for `ExceptT`.

rvl: Thanks for the analysis.
I will remove the `MonadUnliftIO` instances.
If we are going to define this instance, it should work generally.

HeinrichApfelmus:
The context for this pull request has been lost to time. Closing.

PR #2418, opened 2021-01-04
Update persistent to latest upstream version

hasufell:
# Issue Number

ADP-282

# Overview

1. fixed cabal build
2. updated persistent to latest upstream version, which should contain the cascade-delete patch


lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
```diff
@@ -107,10 +108,10 @@ TxMeta
     txMetaDirection         W.Direction         sql=direction
     txMetaSlot              SlotNo              sql=slot
     txMetaBlockHeight       Word32              sql=block_height
-    txMetaAmount            Natural             sql=amount
+    txMetaAmount            OverflowNatural     sql=amount
     txMetaData              W.TxMetadata Maybe  sql=data
     txMetaSlotExpires       SlotNo Maybe        sql=slot_expires
-    txMetaFee               Natural Maybe       sql=fee
+    txMetaFee               OverflowNatural Maybe sql=fee
```
hasufell, line 114: `Natural` isn't supported anymore and the `PersistField Natural` instance is an error instance :roll_eyes:, so we can't even overwrite it.

https://hackage.haskell.org/package/persistent-2.11.0.2/docs/Database-Persist-Types.html#t:OverflowNatural

https://github.com/yesodweb/persistent/blob/8e92f113a1624adfc1cc927911d66eb545e01400/persistent/Database/Persist/Class/PersistField.hs#L355-L368

PR #2423, opened 2021-01-06
Hardcode stake key deposit values to mainnet/testnet for database migrations

KtorZ:
# Issue Number

[ADP-470](https://jira.iohk.io/browse/ADP-470)

# Overview

- f5e8825142ecdf74f41b7e6380f3e15dc43defc6
  :round_pushpin: **hardcode stake key deposit values to mainnet/testnet for database migrations**
    Otherwise, we need to find a way to get an access to the Shelley genesis at this stage which is more work than
  necessary for this particular problem.

# Comments

Opened [ADP-643](https://jira.iohk.io/browse/ADP-643) as a consequence.


piotr-iohk (approved):
LGTM.
Migration now works as expected showing deposit amount correctly.

KtorZ:
bors merge

PR #2425, opened 2021-01-06
import and create random addresses in the 'pending' set

KtorZ:
# Issue Number

[ADP-619](https://jira.iohk.io/browse/ADP-619)

# Overview

- e427df198c3f1d271cf358cbb0a3945e4f7cdd9f
  :round_pushpin: **import and create random addresses in the 'pending' set**
    When we implemented these features, all addresses were placed in the discovered set. The problem with that is that it makes
  them affected by rollbacks. So a client may import 100K addresses, and the next minute they're all gone. We've seen some
  integration tests being 'flaky' actually just showing symptoms of that issue.

  In addition to moving imported and manually created addresses in the pending set, this commit also make the pending set
  returned as part of the total addresses set returned by the API to avoid breaking clients' code and because this is actually
  what most clients would want.

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Random.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Random.hs
@@ -31,7 +31,7 @@ module Cardano.Wallet.Primitive.AddressDiscovery.Random
     -- ** Low-level API
     , importAddress
     , ErrImportAddress(..)
-    , addDiscoveredAddress
+    , addPendingAddress
     , deriveRndStateAddress
     , findUnusedPath
     , unavailablePaths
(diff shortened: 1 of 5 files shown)


lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Random.hs
```diff
@@ -183,18 +181,19 @@ instance RndStateLike (RndState n) where
         case addressToPath addr (hdPassphrase s) of
             Nothing ->
                 Left (ErrAddrDoesNotBelong addr)
-            Just path | Map.member path (addresses s) ->
+            Just path | Map.member path (discoveredAddresses s) ->
+                Right s
```
rvl, line 185: :heavy_check_mark:

rvl (approved):
Good fix :+1:.

It seems like importing new addresses as "pending" is exactly the correct thing to do.

KtorZ:
bors retry

piotr-iohk (approved):
LGTM.

I've checked migration from the wallet version prior to this PR on rnd wallet with some unused addresses:
 - generated via `POST /byron-wallets/{walletId}/addresses`
 - imported via `PUT /byron-wallets/{walletId}/addresses/{addressId}`
 - imported via `PUT /byron-wallets/{walletId}/addresses`

After migration all addresses are present on the wallet.

PR #2431, opened 2021-01-08
Index the wallet's UTxO set by asset type

jonathanknowles:
# Issue Number

ADP-633

# Overview

This PR adds a `UTxOIndex` type, which:

* Provides a mapping from assets to sets of transaction inputs. A transaction input appears in the set if and only if its corresponding output contains a non-zero quantity of that asset.

* Makes it possible to efficiently compute the subset of a UTxO set containing a particular asset, or to select a UTxO entry containing that asset, without having to search linearly through the entire UTxO set.

* Keeps track of the current UTxO balance for all assets, making it possible to efficiently look up the quantity of a particular asset without having to sum across the entire UTxO set.

# Motivation

The wallet currently uses the **Random-Improve** algorithm to generate the majority of its coin selections. This algorithm works by choosing entries at random from the wallet's entire UTxO set until there are enough entries to cover the ada value of each desired output, and then attempts to improve upon each selection until its total value is around twice the value of each output (so as to generate useful change outputs that match the wallet owner's typical spending pattern).

However, in a multi-asset context where we're attempting to cover some amount of a given token, it's imperative that we have a way to efficiently search a subset of a UTxO for entries containing a particular asset without having to traverse the entire set.

# Design

## Indexing Modes

The `UTxOIndex` type supports two possible indexing modes:

1. **Incremental indexing**: entries are added over time via repeated applications of the `insert` operation.

1. **Just-in-time indexing**: the index is constructed from an existing UTxO set in one pass, using the `fromUTxO` operation.

We expect the first version of the MA coin selection algorithm to make use of _just-in-time_ indexing.

### Construction

```hs
fromUTxO
    :: UTxO -> UTxOIndex
```

### Insertion

```hs
insert
    :: TxIn -> TxOut -> UTxOIndex -> UTxOIndex
insertMany
    :: Foldable f => f (TxIn, TxOut) -> UTxOIndex -> UTxOIndex
```

### Deletion

```hs
delete
    :: TxIn -> UTxOIndex -> UTxOIndex
deleteMany
    :: Foldable f => f TxIn -> UTxOIndex -> UTxOIndex
```

### Random Selection

```hs
-- | Selects an entry at random from the index according to the given filter.
--
-- Returns the selected entry and an updated index with the entry removed.
--
-- Returns 'Nothing' if there were no matching entries.
--
selectRandom
    :: MonadRandom m
    => UTxOIndex
    -> SelectionFilter
    -> m (Maybe ((TxIn, TxOut), UTxOIndex))

-- | Specifies a filter for selecting UTxO entries.
--
data SelectionFilter
    = Any
        -- ^ Select any UTxO entry from the entire set.
    | WithAsset AssetId
        -- ^ Select any UTxO entry that has a non-zero quantity of the specified
        -- asset.
```


lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
```diff
@@ -0,0 +1,488 @@
...
+-- |
+-- Copyright: © 2018-2020 IOHK
+-- License: Apache-2.0
+--
+-- Provides internal functions for the 'UTxOIndex' type, which stores a UTxO
+-- set indexed by asset ID.
+--
+-- See the documentation for 'UTxOIndex' for more details.
+--
+module Cardano.Wallet.Primitive.Types.UTxOIndex.Internal
```
hasufell, line 18: Internal modules should expose everything. They're already labelled `Internal`, so there's no reason to e.g. not export the constructor (e.g. for testing purposes or whatnot).

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex.hs
```diff
@@ -0,0 +1,51 @@
+-- |
+-- Copyright: © 2018-2020 IOHK
+-- License: Apache-2.0
+--
+-- Provides the 'UTxOIndex' type, which stores a UTxO set indexed by asset ID.
```
hasufell: I find this a little hard to understand if I assume I didn't know anything about it yet.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex.hs
```diff
@@ -0,0 +1,51 @@
+-- |
+-- Copyright: © 2018-2020 IOHK
```
hasufell: 2021

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
```diff
@@ -0,0 +1,488 @@
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedLabels #-}
+
+-- |
+-- Copyright: © 2018-2020 IOHK
```
hasufell: 2021

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
```diff
@@ -0,0 +1,488 @@
...
+-- Type
+--------------------------------------------------------------------------------
+
+-- | A UTxO set that is indexed by asset identifier.
+--
+-- The index provides a mapping from assets to sets of transaction inputs.
+-- A transaction input appears in the set if and only if its corresponding
+-- output contains a non-zero quantity of that asset.
+--
+-- The index makes it possible to efficiently compute the subset of a UTxO
```
hasufell, line 130: Right, this is the motivation. It's more important then the rest of the comments. We should highlight this more.
Maybe also add that this is used for coin selection, which is the reason for all this.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
```diff
@@ -0,0 +1,488 @@
...
+-- The UTxO index data structure has an invariant that can be checked with
+-- the 'checkInvariant' function.
+--
+data UTxOIndex = UTxOIndex
+    { index
+        :: !(Map AssetId (NonEmptySet TxIn))
+    , balance
+        :: !TokenBundle
+    , utxo
+        :: !(Map TxIn TxOut)
```
hasufell, line 147: Why is this not using `newtype UTxO = UTxO { getUTxO :: Map TxIn TxOut }`?

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
```diff
@@ -0,0 +1,488 @@
...
+
+--------------------------------------------------------------------------------
+-- Construction
+--------------------------------------------------------------------------------
+
+-- | An index with no entries.
+--
+empty :: UTxOIndex
+empty = UTxOIndex
+    { index = mempty
```
hasufell: I'm generally not a fan of `mempty`. Using type-specific values is more clear, imo. I mean, there's a reason you provided `empty` and not just a Monoid instance.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
```diff
@@ -0,0 +1,488 @@
...
+    , balance = mempty
+    , utxo = mempty
+    }
+
+-- | Creates a singleton index from the specified input and output.
+--
```
hasufell, line 167: Why all these semi-empty lines?

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
```diff
@@ -0,0 +1,488 @@
...
+    | otherwise =
+        Just . flip Set.elemAt s <$> getRandomR (0, Set.size s - 1)
+
+--------------------------------------------------------------------------------
+-- Invariant
+--------------------------------------------------------------------------------
+
+-- | The result of checking the invariant with the 'checkInvariant' function.
+--
+data InvariantStatus
```
hasufell, line 398: I'd suggest to also explain the invariants right here in the data types. Then it's an easy overview of all of them, rather then searching for the functions.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs
```diff
@@ -157,6 +166,11 @@ instance TypeError ('Text "Ord not supported for token maps")
         => Ord TokenMap where
     compare = error "Ord not supported for token maps"
 
+instance PartialOrd TokenMap where
```
hasufell, line 203: We should explain why we need this `PartialOrd` and a simplistic example why `Ord` doesn't work.

hasufell (commented):
This is a partial review. Still have to look at the tests in more detail. But so far it looks good to me.

lib/core/test/unit/Cardano/Wallet/Primitive/Types/UTxOIndexSpec.hs
```diff
@@ -0,0 +1,531 @@
...
+prop_fromSequence_invariant = invariantHolds . UTxOIndex.fromSequence
+
+prop_insert_invariant :: TxIn -> TxOut -> UTxOIndex -> Property
+prop_insert_invariant i o u = invariantHolds $ UTxOIndex.insert i o u
+
+prop_delete_invariant :: TxIn -> UTxOIndex -> Property
+prop_delete_invariant i u = invariantHolds $ UTxOIndex.delete i u
+
+prop_selectRandom_invariant :: UTxOIndex -> SelectionFilter -> Property
+prop_selectRandom_invariant i f = monadicIO $ do
```
hasufell, line 188: Can we try avoid IO and use ` DRG gen => MonadRandom (MonadPseudoRandom gen)` instead? I also see this in many other places.

lib/core/test/unit/Cardano/Wallet/Primitive/Types/UTxOIndexSpec.hs
```diff
@@ -0,0 +1,531 @@
...
+prop_toList_fromSequence u =
+    UTxOIndex.fromSequence (UTxOIndex.toList u) === u
+
+--------------------------------------------------------------------------------
+-- Modification properties
+--------------------------------------------------------------------------------
+
+prop_delete_balance :: TxIn -> UTxOIndex -> Property
+prop_delete_balance i u =
+    UTxOIndex.balance (UTxOIndex.delete i u) === expected
```
hasufell, line 223: What's the chance `TxIn` appears in `UTxOIndex` here? I'm not sure we have a reasonable distribution.

We should probably use [cover](https://hackage.haskell.org/package/QuickCheck-2.14.2/docs/Test-QuickCheck.html#v:cover) here.

lib/core/test/unit/Cardano/Wallet/Primitive/Types/UTxOIndexSpec.hs
```diff
@@ -0,0 +1,531 @@
...
+prop_selectRandom_singleton :: SelectionFilter -> TxIn -> TxOut -> Property
+prop_selectRandom_singleton selectionFilter i o = monadicIO $ do
+    actual <- run $ UTxOIndex.selectRandom index selectionFilter
+    assert $ actual == expected
+  where
+    index = UTxOIndex.singleton i o
+    expected = case selectionFilter of
+        Any ->
+            Just ((i, o), UTxOIndex.empty)
+        WithAsset a | txOutHasAsset o a ->
```
hasufell, line 344: What's the distribution of this?

lib/core/test/unit/Cardano/Wallet/Primitive/Types/UTxOIndexSpec.hs
```diff
@@ -0,0 +1,531 @@
...
+--------------------------------------------------------------------------------
+-- Utilities
+--------------------------------------------------------------------------------
+
+-- | Selects all UTxO entries matching a particular filter.
+--
+-- Returns a list of all the entries that matched, and an updated index with
+-- the selected entries removed.
+--
+selectAll
```
hasufell, line 502: I'm seriously wondering why this is not in `Cardano.Wallet.Primitive.Types.UTxOIndex.Internal`.

lib/core/test/unit/Cardano/Wallet/Primitive/Types/UTxOIndexSpec.hs
```diff
@@ -0,0 +1,531 @@
...
+        WithAsset a | txOutHasAsset o a ->
+            Just ((i, o), UTxOIndex.empty)
+        WithAsset _ ->
+            Nothing
+
+-- | Attempt to select a random entry with any combination of assets.
+--
+-- This should always succeed, provided the index is not empty.
+--
+prop_selectRandom_one_any :: UTxOIndex -> Property
```
hasufell, line 353: Alright, this is the property test that actually tests that the implementation of the function doesn't misbehave (e.g. falsely returning Nothing).

But, it relies on other functions not being buggy (such as `delete` etc). There's a very low chance a bug may slip through. This is why I think we must have unit tests as well. Property tests are not a substitute for unit tests.

lib/strict-non-empty-containers/src/Data/Set/Strict/NonEmptySet.hs
```diff
@@ -0,0 +1,95 @@
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE TupleSections #-}
+
+-- |
+-- Copyright: © 2018-2020 IOHK
+-- License: Apache-2.0
+--
+-- Provides a strict implementation of a non-empty set.
```
hasufell, line 9: What was the motivation we added this module?

hasufell (commented):
Reviewed the tests. I think it looks pretty good, but I'd add a few unit tests to test the implementation of e.g. `selectRandom`. I explained my reasoning in one of [the comments](https://github.com/input-output-hk/cardano-wallet/pull/2431#discussion_r554739785).

lib/strict-non-empty-containers/src/Data/Set/Strict/NonEmptySet.hs
jonathanknowles: > What was the motivation we added this module?

It's used here:

```hs
data UTxOIndex = UTxOIndex
    { index
        :: !(Map AssetId (NonEmptySet TxIn))
    , balance
        :: !TokenBundle
    , utxo
        :: !(Map TxIn TxOut)
    }
    deriving (Eq, Generic, Read, Show)
```

Using a non-empty set has two advantages:

1. We don't have to manually enforce the invariant that `(asset, [])` is not allowed in the index.
2. The `Eq` instance can be auto-derived.

lib/core/test/unit/Cardano/Wallet/Primitive/Types/UTxOIndexSpec.hs
jonathanknowles: > I'm seriously wondering why this is not in
> Cardano.Wallet.Primitive.Types.UTxOIndex.Internal.

There's really no use for it outside of tests.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
jonathanknowles: > Why is this not using newtype UTxO = UTxO { getUTxO :: Map TxIn TxOut }?

We could, but it would add a layer of wrapping and unwrapping to every operation that manipulates the `utxo` field (which is internal).

The current implementation only needs to do this wrapping and unwrapping just once, if we actually need to convert to from an ordinary `UTxO`:
```
toUTxO :: UTxOIndex -> UTxO
toUTxO = UTxO . utxo

fromUTxO :: UTxO -> UTxOIndex
fromUTxO = Map.foldlWithKey' (\u i o -> insertUnsafe i o u) empty . getUTxO
```

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
jonathanknowles: Fixed in 00770944363b5c0fd2b9af5cd210a8c03f0e1a43.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
jonathanknowles: Fixed in 0ef36c8a6243c24fdd5ebcfce258c416748ddfab.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex.hs
jonathanknowles: Fixed in 0ef36c8a6243c24fdd5ebcfce258c416748ddfab.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs
jonathanknowles: > We should explain why we need this `PartialOrd` and a simplistic example why `Ord` doesn't work.

Fixed in 07116a8d489ff73d751c295089148df70ae8783b.

lib/core/test/unit/Cardano/Wallet/Primitive/Types/UTxOIndexSpec.hs
jonathanknowles: > What's the chance TxIn appears in UTxOIndex here?

I've added coverage checks in e99202270688d298d0a5d07f5dfb1b9dadb9bb94.

lib/core/test/unit/Cardano/Wallet/Primitive/Types/UTxOIndexSpec.hs
jonathanknowles: > What's the distribution of this?

It's around 1:1:

```
    Index Selection
      prop_SelectionFilter_coverage
        +++ OK, passed 200 tests:
        50.5% selectionFilter: WithAsset
        49.5% selectionFilter: Any
```

I've added a coverage check in ad53a80907287f00710a5816ee00dafe48494030.

lib/core/test/unit/Cardano/Wallet/Primitive/Types/UTxOIndexSpec.hs
```diff
@@ -0,0 +1,573 @@
...
+            , ("greatest", show greatest)
+            , ("midpoint", show midpoint)
+            , ("selected", show selected)
+            ]
+        assert $ fromIntegral smallest % 1 <= midpoint
+        assert $ fromIntegral greatest % 1 >= midpoint
+        assert $ selected `Set.member` set
+  where
+    confidence = stdConfidence {tolerance = 0.99}
+    elementCount = greatest - smallest
```
jonathanknowles: This is an off-by-one error.

jonathanknowles: > This is an off-by-one error.

Fixed in 370ae27087ef15548e38b0cd7b5416593cc4db41.

lib/core/test/unit/Cardano/Wallet/Primitive/Types/UTxOIndexSpec.hs
jonathanknowles: > Can we try avoid IO and use DRG gen => MonadRandom (MonadPseudoRandom gen) instead? I also see this in many other places.

I agree that it would be nicer to avoid `IO` here.

Just for the record: we held a review call and agreed that it would be acceptable to refactor this in a future PR.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
jonathanknowles: Just for the record: in a review call we agreed that leaving it as is would not be harmful.

If there is a need for someone to access the internals of this data type, it would be easy to create a patch that exports the default data constructor.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
jonathanknowles: Hi Julian. Thanks for this feedback!

I've tried to make this a bit clearer in commit 51f535da97bf9838e1e7397a3092a974aaeb763d.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex.hs
jonathanknowles: Hi Julian. Thanks for this feedback.

I've tried to make this a bit clearer in commit 51f535da97bf9838e1e7397a3092a974aaeb763d.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
jonathanknowles: Hopefully this is fixed by 368015f7af854d9af23bc8a07250a7cf4e615d09.

lib/core/src/Cardano/Wallet/Primitive/Types/Tx/Gen.hs
```diff
@@ -0,0 +1,109 @@
...
+
+--------------------------------------------------------------------------------
+-- Transaction hashes chosen from a small range (to allow collisions)
+--------------------------------------------------------------------------------
+
+genTxHashSmallRange :: Gen (Hash "Tx")
+genTxHashSmallRange = elements txHashes
+
+shrinkTxHashSmallRange :: Hash "Tx" -> [Hash "Tx"]
+shrinkTxHashSmallRange hash = filter (< hash) txHashes
```
KtorZ, line 46: Small remark: I am not exactly sure what this is shrinking. Unless I am missing something, there's no way to shrink a hash.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex.hs
```diff
@@ -0,0 +1,55 @@
+-- |
+-- Copyright: © 2018-2021 IOHK
+-- License: Apache-2.0
+--
+-- Provides the 'UTxOIndex' type, which indexes a UTxO set by asset ID.
+--
+-- The index makes it possible to efficiently compute the subset of a UTxO
+-- set containing a particular asset, or to select a UTxO entry containing
+-- that asset, without having to search through the entire UTxO set.
```
KtorZ, line 9: I have some reservations about this entire module. The claim seems to be that computing a subset of a UTxO containing a particular asset id would be too slow, yet.. is that really the case?

Partitioning a UTxO can be done in one pass, and is only needed before running a coin-selection. Even for large UTxO sets this should already be pretty fast to do, and it's a cost we would pay only for coin selections. Now, we'd be paying the cost of the UTxOIndex for every single block application. Although it may be small, it's also impacting a much larger part of the code for something that is only truly needed at one (infrequent) point in time 

Wouldn't it be easier to "flatten" the `TokenMap` instead to facilitate the partitioning, and do the partitioning when setting up the context for coin selections:

```hs
newtype TokenMap = TokenMap
    { unTokenMap
        :: Map AssetId TokenQuantity
    }
```

```hs
partitionUTxO :: UTxO -> NonEmpty AssetId -> Map AssetId UTxO
```

lib/core/src/Cardano/Wallet/Primitive/Types/Tx/Gen.hs
jonathanknowles: > Small remark: I am not exactly sure what this is shrinking. Unless I am missing something, there's no way to shrink a hash.

In the case of `genTxHashSmallRange`, hashes are initially chosen from a small fixed range, so as to allow collisions:

```hs
txHashes :: [Hash "Tx"]
txHashes = mkTxHash <$> ['0' .. '7']
```

The shrinker doesn't shrink the hashes themselves: it merely returns a small list of hashes that are _less than_ the given hash in the available set.

```hs
> generate genTxHashSmallRange
> a = it
> toText a
  "4444444444444444444444444444444444444444444444444444444444444444"

> toText <$> shrinkTxHashSmallRange a
[ "0000000000000000000000000000000000000000000000000000000000000000"
, "1111111111111111111111111111111111111111111111111111111111111111"
, "2222222222222222222222222222222222222222222222222222222222222222"
, "3333333333333333333333333333333333333333333333333333333333333333"
]
```

In some cases, this can lead to simpler counterexamples.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex.hs
jonathanknowles: @KtorZ wrote:
> I have some reservations about this entire module. The claim seems to be that computing a subset of a UTxO containing a particular asset id would be too slow, yet.. is that really the case?
> Partitioning a UTxO can be done in one pass, and is only needed before running a coin-selection. Even for large UTxO sets this should already be pretty fast to do, and it's a cost we would pay only for coin selections. Now, we'd be paying the cost of the UTxOIndex for every single block application. Although it may be small, it's also impacting a much larger part of the code for something that is only truly needed at one (infrequent) point in time

It seems we're comparing:

- **incremental indexing**: where we update the index in small chunks, as we parse blocks.
- **just-in-time indexing**: where we build the entire index in one go, before each coin selection.

If we believe that "just-in-time indexing" is acceptable for now, then we could simply remove the following commit from this PR, which integrates `UTxOIndex` into the `Wallet` type:

https://github.com/input-output-hk/cardano-wallet/pull/2431/commits/43bbeecd580dd0d6952223cbc815ec6c380884e6

This of course leaves us with the option of re-integrating it later on, if we discover (through benchmarking) that incremental indexing is preferable (perhaps for users of wallets with very large UTxO sets).

> Wouldn't it be easier to "flatten" the TokenMap instead to facilitate the partitioning, and do the partitioning when setting up the context for coin selections:
> ```
> newtype TokenMap = TokenMap
>    { unTokenMap
>        :: Map AssetId TokenQuantity
>    }
> partitionUTxO :: UTxO -> NonEmpty AssetId -> Map AssetId UTxO
> ```

This is exactly the type of feedback that would have been useful at the design stage.

Unfortunately, we're now very close to the deadline, and I suspect there is no time budget for making fundamental alterations to the design.

There are course many different ways to slice this problem.

When selecting a UTxO entry at random that contains a given asset, if we choose to select that entry, then we also need to remove it from consideration by future selections. In your proposed simplification, we'd need to remove the entry from all the values of the map returned by `partitionUTxO`.

But this is essentially what the `UTxOIndex` type (and `selectRandom` operation) already does. In addition, it also keeps track of the current balance, which is something we need to keep track of when running coin selection (both for the selected and remaining UTxO sets).

Given that we're under a tight deadline, and given that the existing coin selection code is built with the `UTxOIndex` type in mind, and given that it is fully tested and already been through the review process, I propose that we:

1. Keep the existing implementation of `UTxOIndex`.
2. Remove the commit integrating it into the `Wallet` type.
3. Create the index only when we need to run a coin selection.

Thoughts?

hasufell: > Thoughts?

Yes

hasufell (approved):

KtorZ:
(Please do not merge this one before today's release cut)

KtorZ (approved):

KtorZ:
bors merge

jonathanknowles:
bors merge

jonathanknowles:
bors merge

jonathanknowles:
bors merge

PR #2432, opened 2021-01-08
[ADP-634] Run SMASH metadata fetching in batches of 15 concurrently

hasufell:
# Issue Number

ADP-634

# Overview

All requests already share the same [Manager](https://hackage.haskell.org/package/http-client-0.7.3/docs/Network-HTTP-Client.html#t:Manager). I'm not sure if that implicitly leads to proper HTTP pipelining when used with `forConcurrently`, but it seems to work: the sync time on testnet goes from 3 minutes down to ~30 seconds.

# Comments

## Open questions

1. do we also want to add a way for daedalus to check the sync-status? If so, how to define it? Syncing is continuous.
2. is this safe? We don't want to DoS SMASH
3. do we want to make the batch size a settings? We already have a database table for settings...

### DIFF
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
@@ -114,7 +114,7 @@ import Cardano.Wallet.Unsafe
 import Control.Exception.Base
     ( AsyncException (..), asyncExceptionFromException )
 import Control.Monad
-    ( forM_, forever, void, when )
+    ( forM, forM_, forever, void, when )
 import Control.Monad.IO.Class
     ( liftIO )
 import Control.Monad.Trans.Except
@@ -131,6 +131,8 @@ import Data.Function
     ( (&) )
 import Data.Generics.Internal.VL.Lens
     ( view )
+import Data.List
+    ( nub, (\\) )
 import Data.List.NonEmpty
     ( NonEmpty (..) )
 import Data.Map
@@ -153,14 +155,14 @@ import Data.Time.Clock.POSIX
     ( getPOSIXTime, posixDayLength )
 import Data.Tuple.Extra
     ( dupe )
+import Data.Void
+    ( Void )
 import Data.Word
     ( Word64 )
 import Fmt
     ( fixedF, pretty )
 import GHC.Generics
     ( Generic )
-import Numeric.Natural
-    ( Natural )
 import Ouroboros.Consensus.Cardano.Block
     ( CardanoBlock, HardForkBlock (..) )
 import System.Random
@@ -740,7 +742,7 @@ monitorMetadata gcStatus tr sp db@(DBLayer{..}) = do
 
                 FetchDirect -> do
                     STM.atomically $ writeTVar gcStatus NotApplicable
-                    fetchMetadata manager [identityUrlBuilder]
+                    void $ fetchMetadata manager [identityUrlBuilder]
 
                 FetchSMASH (unSmashServer -> uri) -> do
                     STM.atomically $ writeTVar gcStatus NotStarted
@@ -749,7 +751,7 @@ monitorMetadata gcStatus tr sp db@(DBLayer{..}) = do
                     tid <- forkFinally
                         (gcDelistedPools gcStatus tr db getDelistedPools)
                         onExit
-                    fetchMetadata manager [registryUrlBuilder uri]
+                    void $ fetchMetadata manager [registryUrlBuilder uri]
                         `finally` killThread tid
 
         | otherwise ->
@@ -760,23 +762,29 @@ monitorMetadata gcStatus tr sp db@(DBLayer{..}) = do
     fetchMetadata
         :: Manager
         -> [UrlBuilder]
-        -> IO ()
+        -> IO Void
     fetchMetadata manager strategies = do
         inFlights <- STM.atomically $ newTBQueue maxInFlight
-        forever $ do
-            refs <- atomically (unfetchedPoolMetadataRefs $ fromIntegral maxInFlight)
-            forM_ refs $ \(pid, url, hash) -> withAvailableSeat inFlights $ do
+        endlessly [] $ \keys -> do
+            refs <- nub . (\\ keys) <$> atomically (unfetchedPoolMetadataRefs limit)
+            when (null refs) $ do
+                traceWith tr $ MsgFetchTakeBreak blockFrequency
+                threadDelay blockFrequency
+            forM refs $ \k@(pid, url, hash) -> k <$ withAvailableSeat inFlights (do
                 fetchFromRemote trFetch strategies manager pid url hash >>= \case
                     Nothing ->
                         atomically $ putFetchAttempt (url, hash)
                     Just meta -> do
                         atomically $ putPoolMetadata hash meta
-            when (null refs) $ do
-                traceWith tr $ MsgFetchTakeBreak blockFrequency
-                threadDelay blockFrequency
+                )
       where
-        maxInFlight :: Natural
-        maxInFlight = 20
+        -- Twice 'maxInFlight' so that, when removing keys currently in flight,
+        -- we are left with at least 'maxInFlight' keys.
+        limit = fromIntegral (2 * maxInFlight)
+        maxInFlight = 10
+
+        endlessly :: Monad m => a -> (a -> m a) -> m Void
+        endlessly zero action = action zero >>= (`endlessly` action)
 
         -- | Run an action asyncronously only when there's an available seat.
         -- Seats are materialized by a bounded queue. If the queue is full,
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
@@ -42,7 +42,9 @@ import Cardano.BM.Data.Tracer
 import Cardano.Pool.DB
     ( DBLayer (..), ErrPointAlreadyExists (..), readPoolLifeCycleStatus )
 import Cardano.Pool.Metadata
-    ( StakePoolMetadataFetchLog
+    ( Manager
+    , StakePoolMetadataFetchLog
+    , UrlBuilder
     , defaultManagerSettings
     , fetchDelistedPools
     , fetchFromRemote
@@ -112,7 +114,7 @@ import Cardano.Wallet.Unsafe
 import Control.Exception.Base
     ( AsyncException (..), asyncExceptionFromException )
 import Control.Monad
-    ( forM, forM_, forever, void, when )
+    ( forM_, forever, void, when )
 import Control.Monad.IO.Class
     ( liftIO )
 import Control.Monad.Trans.Except
@@ -157,6 +159,8 @@ import Fmt
     ( fixedF, pretty )
 import GHC.Generics
     ( Generic )
+import Numeric.Natural
+    ( Natural )
 import Ouroboros.Consensus.Cardano.Block
     ( CardanoBlock, HardForkBlock (..) )
 import System.Random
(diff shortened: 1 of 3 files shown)


hasufell:
@piotr-iohk maybe want to do some mainnet testing?

piotr-iohk:
I have done some testing against mainnet on Linux and Windows (on synced network).

**Linux (Ubuntu i7 6 core, 32GB RAM):**

Fetching time:
Not pipelined: ~8 minutes
Pipelined: ~1 minute


**Windows (VM 4 virtual CPU, 4GB RAM):**

Fetching time:
Not pipelined: ~9 min
Pipelined: ~2min


In both cases (and systems) `cardano-wallet` process memory consumption was steady at ~30-60MB.
Observable processor usage up to ~5% on Windows VM (via task manager).

I was wondering if it'd be worth adding such benchmark into benchmark suite for future reference?
I guess the problem would be to determine when the fetching is over?

hasufell:
So it seems no one is really sure what http-client does under the hood when sharing a `Manager`. I opened a ticket: https://github.com/snoyberg/http-client/issues/452

rvl:
Use wireshark to check what's happening (before and after). You don't need to decrypt the HTTPS, just look at how many TCP connections are being opened. Running HTTP requests in parallel is faster than in sequence, because the connection setups are concurrent.

rvl:
From looking at the source, I think the connection idle timeout for the http-client Manager connection pool is 30 seconds. I could be mistaken however.

hasufell:
Following the http-client ticket:

1. http-client doesn't do pipelining in any way and it isn't supported
2. the alternative HTTP2 multiplexing isn't possible, since http-client doesn't support HTTP2
3. There's only one library that seems to do HTTP2 and it has zero tests: https://github.com/lucasdicioccio/http2-client/issues/16

Given the improvement with `forConcurrently` I propose we go this way and make it a Setting that can be adjusted and call it a day.

rvl:
I just did some testing with wireshark and ghci - if you set the header `Connection: Keep-Alive` and the server supports it (SMASH server ought to), then the TCP connection will not be closed between requests. I can see the TCP FIN packet 30 seconds after the last request. It's not pipelining but this: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Keep-Alive

hasufell:
I think we have a pretty clear vision now on how to move this forward:

1. keep using async package to spawn threads and let http-client Manager sharing kick in
2. use a proper queue instead of batches, to ensure we're always waiting for 15 responses
3. add information to https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/listStakePools that indicates per pool whether it's fetched, still fetching or failed/retrying
4. maybe: add another endpoint to query a single pool metadata sync status, e.g. `/stake-pools/{poolId}/metadata/sync-status`

KtorZ:
@rvl implementation reviewed to use a bounded queue. Testing on mainnet, it took a bit less than a minute to fetch all metadata on my machine. I think that's "okay-ish" / sufficient at this stage. What could be useful as a follow-up work I imagine could be to provide some progress stats to Daedalus (e.g. number of pending attempts, number of known registrations...)

KtorZ:
There's one seemingly important yet subtle change in this PR with regards to the fetching strategy. 

From:
```hs
when (null refs || null successes) $ do
    traceWith tr $ MsgFetchTakeBreak blockFrequency
    threadDelay blockFrequency
``` 

To: 
```hs
when (null refs) $ do
    traceWith tr $ MsgFetchTakeBreak blockFrequency
    threadDelay blockFrequency
```

So, as of now, the wallet pauses for a whole 20s when all requests to fetching metadata failed. I think the rationale was originally to prevent the wallet from just eating up CPU by trying to fetch only invalid pools... What this create in practice is simply more unnecessary delays. There's about 8.5K pool registrations currently on mainnet, and after a full sync, there are still ~2254 fetch attempts still-to-proceed... so that's a lot of occasions to get a `null successes` on a particular loop. 

There's already a backoff which comes with the fetch attempts which prevent the wallet from doing needless work anyway. With this little change, we should massively improve the time needed for updating metadata upon new registrations, such as the problem described in: https://github.com/input-output-hk/daedalus/issues/2359

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -723,43 +738,79 @@ monitorMetadata gcStatus tr sp db@(DBLayer{..}) = do
...
+    -- Should be:
+    --     - <= queueSize / 2
+    numInFlight :: Int
+    numInFlight = 20
+
+    enqueueMetadata
+        :: TBQueue (PoolId, StakePoolMetadataUrl, StakePoolMetadataHash)
+        -> IO ()
+    enqueueMetadata queue = forever $ do
         refs <- atomically (unfetchedPoolMetadataRefs 100)
```
rvl: Perhaps set this number of refs to be smaller than queueSize - maybe the same as numInFlight.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -723,43 +738,79 @@ monitorMetadata gcStatus tr sp db@(DBLayer{..}) = do
...
+            -- First read is blocking, and others aren't. This ensures a cheap
+            -- pause of the thread when there's nothing to fetch, but it always
+            -- try to fetch at most 'numInFlight' at once without blocking if
+            -- there are less to fetch.
+            -- This could be the case when the queue has been emptied and
+            -- registrations are now coming one-by-one.
+            h <- readTBQueue queue
+            q <- catMaybes <$> replicateM (numInFlight - 1) (tryReadTBQueue queue)
+            pure (h:q)
+        forConcurrently_ refs $ \(pid, url, hash) -> do
```
rvl: This will still have uneven bursts of requests, won't it?
But if the fetch time is down to one minute, then it's not really a problem.

rvl (approved):
:+1: Nice.
The time taken to fetch all metadata is highly dependent on your ping to the smash server.
So a 1 minute fetch in Europe could be longer in Australia.
How long did it take before to fetch all metadata on your connection?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
KtorZ: Whoops, I should've removed the magic number here and use `queueSize` instead. Thing is, I don't want the queue to be empty if there's something to fetch. So ideally, this should be > numInFlight. The bounded queue will block on writes anyway if the queue is full.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
KtorZ: I see what you mean, we could actually make the consumer send queries one-by-one (asynchronously) and actually use the queue's size to limit the number of requests in flight :thinking: ...

KtorZ:
> @rvl: How long did it take before to fetch all metadata on your connection?

About 320s. 
50s now after the last change.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
KtorZ: Done :+1:

KtorZ: Whoops. Actually no, it's not that simple because now `unfetchedPoolMetadataRefs` gets called multiple times in a row before it's correctly updated.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -723,43 +734,58 @@ monitorMetadata gcStatus tr sp db@(DBLayer{..}) = do
...
+        -> IO ()
+    fetchMetadata manager strategies = do
+        inFlights <- STM.atomically $ newTBQueue maxInFlight
+        forever $ do
+            refs <- atomically (unfetchedPoolMetadataRefs $ fromIntegral maxInFlight)
+            forM_ refs $ \(pid, url, hash) -> withAvailableSeat inFlights $ do
+                fetchFromRemote trFetch strategies manager pid url hash >>= \case
+                    Nothing ->
+                        atomically $ putFetchAttempt (url, hash)
+                    Just meta -> do
```
hasufell, line 777: redundant `do`?

KtorZ: Who cares :sweat_smile: ? If the compiler judges it redundant, it'll remove it.

piotr-iohk:
For me on `mainnet`:
 - it was: **8m 5s**
 - with `jospald/ADP-634/improve-smash-sync-time` (a6c05d10fe5595a9d84c914fba7157d5415f64d4): **50s**

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
hasufell: I think hlint will bail?

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -723,43 +734,58 @@ monitorMetadata gcStatus tr sp db@(DBLayer{..}) = do
...
+
+    fetchMetadata
+        :: Manager
+        -> [UrlBuilder]
+        -> IO ()
+    fetchMetadata manager strategies = do
+        inFlights <- STM.atomically $ newTBQueue maxInFlight
+        forever $ do
+            refs <- atomically (unfetchedPoolMetadataRefs $ fromIntegral maxInFlight)
+            forM_ refs $ \(pid, url, hash) -> withAvailableSeat inFlights $ do
```
hasufell: This is a pretty neat pattern. Unrelatedly, I'm wondering if this is achievable with streamly or conduit.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -723,43 +734,58 @@ monitorMetadata gcStatus tr sp db@(DBLayer{..}) = do
...
+                    Nothing ->
+                        atomically $ putFetchAttempt (url, hash)
+                    Just meta -> do
+                        atomically $ putPoolMetadata hash meta
+            when (null refs) $ do
+                traceWith tr $ MsgFetchTakeBreak blockFrequency
+                threadDelay blockFrequency
+      where
+        maxInFlight :: Natural
+        maxInFlight = 20
```
hasufell: Probably doesn't matter for merging the PR, but did someone check other values? I'm not questioning that 20 is a reasonable default, but maybe there's more room to optimize? At which point will performance degrade or SMASH give up on us?

KtorZ: I've tried: 10, 20 and 50. There seem to be almost no difference and they all oscillate between 40s and 60s for a entire refresh :man_shrugging: .. I went for 10 in the end.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
KtorZ: I actually thought of using a conduit here, to stream things directly from the database and then, process them at the right path. Yet, that's much more work on the db side to do, and persistent API is quite unsatisfactory here (we use a custom raw SQL query behind the scene, so most of the persistent API is just unusable). Plus, we would need a "refreshable" stream, because after requests are processed, the set of metadata refs to fetch do actually change.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
KtorZ: Okay, now done properly. After a full run, there's no attempts with a retry_count greater than 3, which sounds right.

rvl:
Thanks :wave: 

bors r+

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
KtorZ: Not if asked gently: https://github.com/input-output-hk/cardano-wallet/blob/master/.hlint.yaml#L51

PR #2438, opened 2021-01-12
Make hard fork era configurable in local test cluster

rvl:
### Issue Number

ADP-623

### Overview

The program name `shelley-test-cluster` is renamed to `local-cluster` (era-independent).

The cluster module is moved so that application code is more clearly separated from code for setting up test environments.

This lets the final era of `local-cluster` be configured by an environment variable (`LOCAL_CLUSTER_ERA=mary`).

The integration tests now run on the current era (Allegra) and the next (Mary). This will help cover the several instances of era-specific cases in the code.

Once the hard fork is done, we can go back to testing on a single era.

### Comments

Two local clusters can't be run concurrently due to global variables used for test fixture data.


rvl:
bors try

Anviking:
> This sets the final era of the local test cluster be configured by an environment variable.

I take it you're planning to have the `testOnAllegra >> testOnMary` in the CI setup rather than in Haskell, or not at all?

Regardless, seems more flexible — I like it.

rvl:
@Anviking I clarified the PR description. The environment variable really only applies to the `shelley-test-cluster` exe. For integration tests, it's hard-coded in `Main.hs`.

Anviking:
bors try

Anviking:
bors try-

Anviking:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Launch/Cluster.hs
```diff
@@ -0,0 +1,1717 @@
...
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+
+-- |
+-- Copyright: © 2018-2020 IOHK
+-- License: Apache-2.0
+--
+-- Provides functions to launch cardano-nodes in a cluster for /testing/.
```
Anviking, line 22: 💯

lib/shelley/test/integration/Main.hs
```diff
@@ -276,34 +285,27 @@ specWithServer testDir (tr, tracers) = aroundAll withContext
...
+                onClusterStart (onReady $ T.pack smashUrl) dbDecorator
 
     tr' = contramap MsgCluster tr
-    setupFaucet = do
+    setupFaucet conn = do
         traceWith tr MsgSettingUpFaucet
         let rewards = (,Coin $ fromIntegral oneMillionAda) <$>
                 concatMap genRewardAccounts mirMnemonics
-        moveInstantaneousRewardsTo tr' testDir rewards
+        moveInstantaneousRewardsTo tr' conn testDir rewards
```
Anviking, line 293: Btw, I had been thinking one could probably factor out `setupFaucet` from integration:main and local-cluster to a simpler variant of withCluster (or even the same).

Anviking (approved):
Tried changing a `QueryIfCurrentAllegra` to `QueryIfCurrentMary` locally, and tests failed indeed.

@KtorZ's concern that the test flakiness would double (or more) might be very warranted though, with our current failure rates (60%). Maybe we should hold off for now.

Anviking:
bors try

Anviking:
Another con with this approach though is: when running
```
stack test cardano-wallet:integration --ta '-j 8'
```

locally, you probably only care about running the tests on any era.

But we'd now default to running them twice.

lib/shelley/test/integration/Main.hs
rvl: True - probably some other things could be factored out too.

rvl:
Thanks. I changed Main.hs to just run on mary, with the LOCAL_CLUSTER_ERA variable to override.

bors r+

lib/launcher/src/Cardano/Launcher/Node.hs
```diff
@@ -19,21 +25,59 @@ import Cardano.Launcher
...
+
+-- | Gets the socket filename or pipe name from 'CardanoNodeConn'. Whether it's
+-- a unix socket or named pipe depends on the value of 'isWindows'.
+nodeSocketFile :: CardanoNodeConn -> FilePath
+nodeSocketFile (CardanoNodeConn name) = name
+
+-- | Produces a 'CardanoNodeConn' if the socket path or pipe name (depending on
+-- 'isWindows') is valid.
+cardanoNodeConn :: FilePath -> Either String CardanoNodeConn
+cardanoNodeConn name
```
Anviking, line 59: I don't think these changes are needed now?, but can't hurt I take.

Issue #2320, opened 2020-11-16
Flaky test - The node backend is unreachable at the moment.


KtorZ:
# Context
 
**Bug report:** https://jira.iohk.io/browse/ADP-647

 

Saw only once, could just be us overloading the node due to parallel tests, especially in this region

- https://hydra.iohk.io/build/4800986
- https://hydra.iohk.io/build/5279297

# Test Case

- STAKE_POOLS_QUIT_02 from `src/Test/Integration/Scenario/API/Shelley/StakePools.hs:433`
- STAKE_POOLS_QUIT_01x

# Failure / Counter-example

```
  1) API Specifications, SHELLEY_STAKE_POOLS, STAKE_POOLS_QUIT_02 - Passphrase must be correct to quit
       uncaught exception: RequestException
       DecodeFailure "{\"code\":\"network_unreachable\",\"message\":\"The node backend is unreachable at the moment. Trying again in a bit might work.\"}"
```
```
  src/Test/Integration/Scenario/API/Shelley/StakePools.hs:850:9: 
  2) API Specifications, SHELLEY_STAKE_POOLS, STAKE_POOLS_QUIT_01x - Fee boundary values, STAKE_POOLS_QUIT_01xx - I can quit if I have enough to cover fee
       uncaught exception: RequestException
       DecodeFailure "{\"code\":\"network_unreachable\",\"message\":\"The node backend is unreachable at the moment. Trying again in a bit might work.\"}"
```

It seems that most recent failures are always preceded by:

```
  src/Test/Integration/Framework/DSL.hs:1471:16: 
  1) API Specifications, SHELLEY_WALLETS, WALLETS_LIST_01 - Wallets are listed from oldest to newest
       uncaught exception: ErrorCall
       getFromResponse failed to get item
       CallStack (from HasCallStack):
         error, called at src/Test/Integration/Framework/DSL.hs:1471:16 in cardano-wallet-core-integration-2020.12.21-AAYZFneZURcKNEp0gMVmAX:Test.Integration.Framework.DSL

  To rerun use: --match "/API Specifications/SHELLEY_WALLETS/WALLETS_LIST_01 - Wallets are listed from oldest to newest/"
```

So, it might be the one we ought to investigate. 

# Resolution

See https://jira.iohk.io/browse/ADP-647

rvl:
@KtorZ That error message is a big red herring.
There is no evidence that the node backend is unreachable.
[...]

PR #2442, opened 2021-01-12
Tweak bors-stats.sh to support jira links

Anviking:
# Issue Number

None

# Overview

- [x] Improve pattern excluding bors try (now removes two recent "try\nTimed out")
- [x] Support `ADP-` as tags for the breakdown

# Comments

### Before
Skärmavbild 2021-01-12 kl  11 47 06

### Now

Skärmavbild 2021-01-12 kl  11 46 06

### DIFF
--- a/scripts/bors-stats.sh
+++ b/scripts/bors-stats.sh
@@ -48,7 +48,7 @@ DATA=$(echo $QUERY \
       | sort_by (.createdAt)
       | map (
           select(.author.login == "iohk-bors")
-          | select(.bodyText | contains("try\nBuild") | not)
+          | select(.bodyText | startswith("try") | not)
           | select(.bodyText | contains("Canceled") | not)
           | select(.bodyText | contains("Merge conflict") | not)
           | select(.bodyText | contains("Rejected by too few approved reviews") | not)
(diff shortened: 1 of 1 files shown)


Anviking:
I reclassified the ADP-647 failures to #2320 for consistency, still. Maybe we should't allow ADP- then.

rvl (approved):
Yeah nice - I already did the same change locally yesterday. :-)

rvl:
If the test failure is due to a bug, then we need to be able to link to jira.

It's possible to get jira ticket titles with go-jira.

rvl:
I made some improvements to the script and ran shellcheck over it.

Anviking:
That breaks it for me (macOS) with:
```
ktemp: illegal option -- -
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix
```

Anviking:
You have just a warning, right? I reverted the commit.
bors r+

rvl:
Oops - I was assuming gnu coreutils. Turns on the macOS `mktemp` is [completely different](https://stackoverflow.com/a/31397073)! You need to install a better OS. :smile_cat:

PR #2444, opened 2021-01-13
Fix yq in travis

piotr-iohk:
# Issue Number

# Overview

- 54f7b714f16f405454012b1ba3e433f7a62f2135
  Fix yq command in travis.yml

# Comments

Tested here -> https://travis-ci.org/github/input-output-hk/cardano-wallet/builds/754259146.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,15 +49,12 @@ jobs:
     if: type != pull_request AND (tag =~ ^v OR commit_message =~ /TRAVIS_TRIGGER_RELEASE/)
     name: "Executables"
     script:
-     # # Upload swagger to bump.sh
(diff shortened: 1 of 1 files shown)


KtorZ (approved):
OK. :crossed_fingers:

rvl:
One thing I noticed about Travis is that it's really hard to test the CI steps locally.

It's a similar situation for Buildkite. My solution for Buildkite was to make every non-trivial pipeline step into a separate script, which are all kept in the [`.buildkite`](https://github.com/input-output-hk/cardano-wallet/tree/master/.buildkite) directory.

Most of the scripts can be run without nix, but nix can be used to automate installation of dependencies.
So with those two things you have a reproducible environment, and scripts can be tested locally. (Obviously environment variables `BUILDKITE_*` and `TRAVIS_*` don't exist locally, but you can set dummy values for testing.)

PR #2445, opened 2021-01-13
Don't readWalletMeta while restoring

Anviking:
# Issue Number

ADP-639

# Overview

- [x] Remove pointless readWalletMeta in restoreBlocks

# Comments

- Hope this fixes the restore bench (see commit message)

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -850,7 +850,6 @@ restoreBlocks
     -> ExceptT ErrNoSuchWallet IO ()
 restoreBlocks ctx wid blocks nodeTip = db & \DBLayer{..} -> mapExceptT atomically $ do
     cp   <- withNoSuchWallet wid (readCheckpoint $ PrimaryKey wid)
-    meta <- withNoSuchWallet wid (readWalletMeta $ PrimaryKey wid)
     sp   <- liftIO $ currentSlottingParameters nl
 
     unless (cp `isParentOf` NE.head blocks) $ fail $ T.unpack $ T.unwords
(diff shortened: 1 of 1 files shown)


KtorZ:
https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/809#e26d2361-279f-45fb-a473-7af1dcf17105

lib/core/src/Cardano/Wallet.hs
```diff
@@ -908,7 +907,6 @@ restoreBlocks ctx wid blocks nodeTip = db & \DBLayer{..} -> mapExceptT atomicall
 
     liftIO $ do
         progress <- walletSyncProgress @ctx @s ctx (NE.last cps)
-        traceWith tr $ MsgWalletMetadata meta
```
KtorZ, line 911: We could perhaps tweak the `MsgSyncProgress` to include a wallet name and possibly the active delegation status. We don't need the time interpreter for those, and it's still quite valuable to log those information when debugging logs.

rvl (commented):
Let's evict `TimeInterpreter` from the DB layer.

All `readWalletMeta` needs is the epoch number. It can be a function parameter.

The epoch number is needed for delegation status.
It would also be totally fine to cut the `WalletDelegation` field out of `WalletMetadata` and have them as separate entities.

Other DBLayer methods probably use the `TimeInterpreter` for various things. But they can be passed functions such as `SlotNo -> UTCTime` instead.

Anviking:
Closing as this didn't fix the restore bench.

> Let's evict TimeInterpreter from the DB layer.

Agree — but don't think it's critical

PR #2446, opened 2021-01-14
API spec: First revision of multi-asset changes

rvl:
### Issue Number

ADP-603

### Overview

Adjusts the OpenAPI specification to include multi-asset amounts in transactions and wallet balances.

Changes are designed to preserve API backwards compatibility, if users only care about Ada amounts.

A primary consideration was how multi-asset values would be presented and manipulated in user interfaces such as Daedalus.

- [Rendered](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/input-output-hk/cardano-wallet/rvl/adp-603/api-spec-ma/specifications/api/swagger.yaml#tag/Assets)

### Comments

- Targets PR #2447 branch, because API tests require the implementation to match the spec.

### DIFF
--- a/specifications/api/swagger.yaml
+++ b/specifications/api/swagger.yaml
@@ -3592,13 +3592,30 @@ paths:
       description: |
         <p align="right">status: <strong>under development</strong></p>
 
-        Fetch a single asset from its `policy_id` and `asset_name`, with its metadata if any.
+        Fetch a single asset from its `policy_id` and `asset_name`,
+        with its metadata if any.
 
         The asset must be associated with the wallet.
       parameters:
         - *parametersWalletId
         - *parametersPolicyId
         - *parametersAssetName
+      responses: *responsesGetAsset
+
+  /wallets/{walletId}/assets/{policyId}:
+    get:
+      operationId: getAssetDefault
+      tags: ["Assets"]
+      summary: Get Asset (empty name)
+      description: |
+        <p align="right">status: <strong>under development</strong></p>
+
+        Fetch the the asset from `policy_id` with an empty name.
+
+        The asset must be associated with the wallet.
+      parameters:
+        - *parametersWalletId
+        - *parametersPolicyId
       responses: *responsesGetAsset
 
   /wallets/{walletId}/statistics/utxos:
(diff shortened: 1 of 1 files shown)


specifications/api/swagger.yaml
```diff
@@ -433,6 +433,133 @@ x-settings: &settings
...
+
+    For policies with a single fungible asset item, this is typically
+    an empty string.
+
+    This value can be up to 32 bytes of arbitrary data (which is 64
+    hexadecimal digits).
+  # fixme: (remove this note) I changed it to hex from base64, the reason is below.
+  format: hex
+  maxLength: 64
+  # fixme: (remove this note) multiple examples seem not to validate
```
KtorZ: Unfortunately no, OpenAPI does not support multiple examples... I think there's a Redoc extension we could use if we really wanted that, but it's a vendor-specific extension so not a pure OpenAPI primitive.

KtorZ:
@rvl I've been capturing the metadata in a confluence document: https://input-output.atlassian.net/wiki/spaces/AD/pages/2137456733/Token+Identity+Metadata

I've added a few (sensible) constraints / validations not yet present in our swagger specs (e.g. maxLength for the token's name)

specifications/api/swagger.yaml
```diff
@@ -236,7 +236,7 @@ x-syncClockProgress: &syncClockProgress
     status: pending
 
 x-amount: &amount
-  description: Coins, in Lovelace
+  description: Coins, in Lovelace. Only relates to 'Ada'. Refer to `assets` for multi-assets wallets instead.
```
Anviking, line 239: Q: Can a user specify to send 0 ada, and 1 Apple asset?

I'm guessing: yes, but the coin selection will automatically send the minimum required ada along in the token bundle?

rvl:
Thanks @KtorZ I shall apply that json schema to the `swagger.yaml`.

I have one issue with the metadata schema.
Acronynm should not be a required field. For some assets it would not make sense.
We could say that if acronym is not present, then the name will be used, but UIs might truncate the text if there isn't enough space.

specifications/api/swagger.yaml
KtorZ: Indeed. And also assign the minimum ada value needed for the multi asset UTxO.

KtorZ:
@rvl good point. We can still change that.

KtorZ:
closing since #2447 has been merged and was already including this chunk of work.

PR #2447, opened 2021-01-14
Multi-asset API extensions

rvl:
### Issue Number

ADP-604

### Overview

Starting implementation of API spec in #2446.

- [x] listAssets endpoint
- [x] getAsset endpoint
- [x] Added `ApiWalletAssetsBalance` which provides the balance of non-ade assets.
- [x] Removed WalletBalance from Primitive.Types and replaced `ApiT WalletBalance` with `ApiWalletBalance`.
- [ ] Some kind of integration test for these endpoints.
- [x] Remove `Quantity "lovelace" a` where possible, except for the API.
- [x] TokenName inner type is now ByteString, add length validation.

### Comments
- This branch is based on top of the PR #2446 branch.

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -1905,7 +1905,7 @@ mkTxMeta ti' blockHeader wState tx cs expiry =
                 , direction = if amtInps > amtOuts then Outgoing else Incoming
                 , slotNo = blockHeader ^. #slotNo
                 , blockHeight = blockHeader ^. #blockHeight
-                , amount = Quantity $ distance amtInps amtOuts
+                , amount = Coin $ fromIntegral $ distance amtInps amtOuts
                 , expiry = Just expiry
                 }
             )
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -322,7 +322,7 @@ import Cardano.Wallet.Primitive.Types
 import Cardano.Wallet.Primitive.Types.Address
     ( Address (..), AddressState (..) )
 import Cardano.Wallet.Primitive.Types.Coin
-    ( Coin (..) )
+    ( Coin (..), addCoins, coinQuantity, sumCoins )
 import Cardano.Wallet.Primitive.Types.Hash
     ( Hash (..) )
 import Cardano.Wallet.Primitive.Types.RewardAccount
@@ -965,7 +965,7 @@ fetchRewardBalance
         )
     => ctx
     -> WalletId
-    -> IO (Quantity "lovelace" Word64)
+    -> IO Coin
 fetchRewardBalance ctx wid = db & \DBLayer{..} ->
     atomically $ readDelegationRewardBalance pk
   where
@@ -985,13 +985,13 @@ readNextWithdrawal
         )
     => ctx
     -> WalletId
-    -> Quantity "lovelace" Word64
-    -> IO (Quantity "lovelace" Word64)
-readNextWithdrawal ctx wid (Quantity withdrawal) = db & \DBLayer{..} -> do
+    -> Coin
+    -> IO Coin
+readNextWithdrawal ctx wid (Coin withdrawal) = db & \DBLayer{..} -> do
     liftIO (atomically $ readProtocolParameters $ PrimaryKey wid) <&> \case
         -- May happen if done very early, in which case, rewards are probably
         -- not woth considering anyway.
-        Nothing -> Quantity 0
+        Nothing -> Coin 0
         Just ProtocolParameters{txParameters} ->
             let policy = W.getFeePolicy txParameters
 
@@ -1002,8 +1002,8 @@ readNextWithdrawal ctx wid (Quantity withdrawal) = db & \DBLayer{..} -> do
 
             in
                 if toInteger withdrawal < 2 * costOfWithdrawal
-                then Quantity 0
-                else Quantity withdrawal
+                then Coin 0
+                else Coin withdrawal
   where
     db = ctx ^. dbLayer @s @k
     tl = ctx ^. transactionLayer @k
@@ -1048,7 +1048,7 @@ queryRewardBalance
         )
     => ctx
     -> RewardAccount
-    -> ExceptT ErrFetchRewards IO (Quantity "lovelace" Word64)
+    -> ExceptT ErrFetchRewards IO Coin
 queryRewardBalance ctx acct = do
     mapExceptT (fmap handleErr) $ getAccountBalance nw acct
   where
@@ -1056,7 +1056,7 @@ queryRewardBalance ctx acct = do
     handleErr = \case
         Right x -> Right x
         Left (ErrGetAccountBalanceAccountNotFound _) ->
-            Right $ Quantity 0
+            Right $ Coin 0
         Left (ErrGetAccountBalanceNetworkUnreachable e) ->
             Left $ ErrFetchRewardsNetworkUnreachable e
 
@@ -1292,7 +1292,7 @@ selectCoinsForPayment
     => ctx
     -> WalletId
     -> NonEmpty TxOut
-    -> Quantity "lovelace" Word64
+    -> Coin
     -> Maybe TxMetadata
     -> ExceptT ErrSelectForPayment IO CoinSelection
 selectCoinsForPayment ctx wid recipients withdrawal md = do
@@ -1301,7 +1301,7 @@ selectCoinsForPayment ctx wid recipients withdrawal md = do
         selectCoinsSetup @ctx @s @k ctx wid
 
     let pendingWithdrawal = Set.lookupMin $ Set.filter hasWithdrawal pending
-    when (withdrawal /= Quantity 0 && isJust pendingWithdrawal) $ throwE $
+    when (withdrawal /= Coin 0 && isJust pendingWithdrawal) $ throwE $
         ErrSelectForPaymentAlreadyWithdrawing (fromJust pendingWithdrawal)
 
     cs <- selectCoinsForPaymentFromUTxO @ctx @k
@@ -1339,14 +1339,14 @@ selectCoinsForPaymentFromUTxO
     -> W.TxParameters
     -> W.Coin
     -> NonEmpty TxOut
-    -> Quantity "lovelace" Word64
+    -> Coin
     -> Maybe TxMetadata
     -> ExceptT ErrSelectForPayment IO CoinSelection
 selectCoinsForPaymentFromUTxO ctx utxo txp minUtxo recipients withdrawal md = do
     lift . traceWith tr $ MsgPaymentCoinSelectionStart utxo txp recipients
     (sel, utxo') <- withExceptT handleCoinSelError $ do
         let opts = coinSelOpts tl (txp ^. #getTxMaxSize) md
-        CoinSelection.random opts recipients withdrawal utxo
+        CoinSelection.random opts recipients (coinQuantity withdrawal) utxo
 
     lift . traceWith tr $ MsgPaymentCoinSelection sel
     let feePolicy = feeOpts tl Nothing md txp minUtxo sel
@@ -1455,10 +1455,7 @@ selectCoinsForMigration
     => ctx
     -> WalletId
        -- ^ The source wallet ID.
-    -> ExceptT ErrSelectForMigration IO
-        ( [CoinSelection]
-        , Quantity "lovelace" Natural
-        )
+    -> ExceptT ErrSelectForMigration IO ([CoinSelection], Coin)
 selectCoinsForMigration ctx wid = do
     (utxo, _, txp, minUtxo) <- withExceptT ErrSelectForMigrationNoSuchWallet $
         selectCoinsSetup @ctx @s @k ctx wid
@@ -1476,10 +1473,7 @@ selectCoinsForMigrationFromUTxO
     -> W.Coin
     -> WalletId
        -- ^ The source wallet ID.
-    -> ExceptT ErrSelectForMigration IO
-        ( [CoinSelection]
-        , Quantity "lovelace" Natural
-        )
+    -> ExceptT ErrSelectForMigration IO ([CoinSelection], Coin)
 selectCoinsForMigrationFromUTxO ctx utxo txp minUtxo wid = do
     let feePolicy@(LinearFee (Quantity a) _) = txp ^. #getFeePolicy
     let feeOptions = (feeOpts tl Nothing Nothing txp minBound mempty)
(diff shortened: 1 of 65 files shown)


lib/core/src/Cardano/Wallet.hs
```diff
@@ -1905,7 +1905,10 @@ mkTxMeta ti' blockHeader wState tx cs expiry =
                 , direction = if amtInps > amtOuts then Outgoing else Incoming
                 , slotNo = blockHeader ^. #slotNo
                 , blockHeight = blockHeader ^. #blockHeight
-                , amount = Quantity $ distance amtInps amtOuts
+                -- fixme: ADP-347 calculate asset balance too
```
KtorZ: The `amount` should remain unchanged and ada-only. We will not be providing this information for assets. It is less useful for assets since fees can only be paid in Ada. What assets are being sent can be figured out from the resolved inputs and outputs fully.

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -376,7 +378,9 @@ prefilterBlock b u0 = runState $ do
         , direction = dir
         , slotNo = b ^. #header . #slotNo
         , blockHeight = b ^. #header . #blockHeight
-        , amount = Quantity amt
+        -- fixme: ADP-347
+        -- fixme: why on earth do we have both Coin and Quantity "lovelace" Natural?
```
KtorZ: At the very beginning, the `Quantity` type was only used in the Api as a way to generate pretty JSON. It has one day escaped its cage and spread over the rest of the codebase as a quick replacement for newtypes over numeric types.. So in some places we have `Coin`, in some `Word64`, in some `Quantity "lovelace" Natural` or `Quantity "lovelace" Word64` or also `Fee` which is very much the same thing as `Coin`.

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
```diff
@@ -268,12 +267,14 @@ instance Buildable (TxIn, TxOut) where
...
 -- application-specific data included with the transaction.
+--
+-- TODO: TxProperties or TxProps would be a good name for this type.
 data TxMeta = TxMeta
     { status :: !TxStatus
     , direction :: !Direction
     , slotNo :: !SlotNo
     , blockHeight :: !(Quantity "block" Word32)
-    , amount :: !(Quantity "lovelace" Natural)
+    , amount :: !TokenBundle
```
KtorZ: I think this should be `Coin` and only `Coin`.

lib/core/src/Cardano/Wallet.hs
rvl: OK then. Nonetheless I thought it would be good to calculate a total of assets moved in or out of the wallet by the transaction. In `mkTxMeta` it's easier to know which tx outputs are "ours". Resolved inputs and outputs show the address only.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1608,6 +1678,17 @@ instance FromJSON ApiByronWalletBalance where
...
+instance FromJSON ApiWalletAssetsBalance where
+    parseJSON = genericParseJSON defaultRecordTypeOptions
+instance ToJSON ApiWalletAssetsBalance where
+    toJSON = genericToJSON defaultRecordTypeOptions
+
+-- fixme: doesn't quite match the spec
+instance FromJSON (ApiT W.TokenMap) where
+    parseJSON = fmap (ApiT . W.getFlat) . parseJSON
+instance ToJSON (ApiT W.TokenMap) where
+    toJSON = toJSON . W.Flat . getApiT
```
jonathanknowles, line 1682: Hi @rvl 

I'm just curious about the reason for choosing `Flat` here. Is there a reason to not use `Nested`?

(From recollection, the Daedalus team indicated that they preferred a nested representation, where assets were grouped by token policy.)

Apologies if I'm missing some context here which would make the reason for this choice obvious.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy.hs
```diff
@@ -81,3 +83,11 @@ instance ToText TokenName where
 
 instance FromText TokenName where
     fromText = pure . UnsafeTokenName
```
jonathanknowles: Related question: are we planning to change the type of `TokenName` from `Text` to `ByteString` (or perhaps `ShortByteString`), given that it can be any sequence of bytes between 0 bytes and 32 bytes in length (and therefore might not be representable by `Text`).

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
```diff
@@ -268,12 +267,14 @@ instance Buildable (TxIn, TxOut) where
...
 -- application-specific data included with the transaction.
+--
+-- TODO: TxProperties or TxProps would be a good name for this type.
 data TxMeta = TxMeta
     { status :: !TxStatus
     , direction :: !Direction
     , slotNo :: !SlotNo
     , blockHeight :: !(Quantity "block" Word32)
-    , amount :: !(Quantity "lovelace" Natural)
+    , amount :: !Coin
```
jonathanknowles, line 282: 👍🏻

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -323,6 +331,30 @@ type SignMetadata = "wallets"
...
+    :> Capture "walletId" (ApiT WalletId)
+    :> "assets"
+    :> Get '[JSON] [ApiAsset]
+
+-- | https://input-output-hk.github.io/cardano-wallet/api/#operation/getAsset
+type GetAsset = "wallets"
+    :> Capture "walletId" (ApiT WalletId)
+    :> "assets"
+    :> Capture "policyId" (ApiT TokenPolicyId)
+    :> Capture "policyItem" (ApiT TokenName)
```
jonathanknowles: If we're using the names in the specification, perhaps should this be "assetName"?

```suggestion
    :> Capture "assetName" (ApiT TokenName)
```

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: Nesting makes it more complicated to process. The fact that there's a hierarchy policy_id -> asset_name is actually an implementation detail. End-users care about tokens, and tokens are uniquely identified by a pair (policy_id, asset_name). Tokens with different names are different which is why in the end, a flat views of assets makes makes more sense in the interface.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2044,6 +2084,7 @@ mkApiTransaction ti txid mfee ins outs ws (meta, timestamp) txMeta setTimeRefere
         , inputs = [ApiTxInput (fmap toAddressAmount o) (ApiT i) | (i, o) <- ins]
         , outputs = toAddressAmount <$> outs
         , withdrawals = mkApiWithdrawal @n <$> Map.toList ws
+        , forge = mempty  -- fixme: ADP-604
```
jonathanknowles: Question: Should this be `mint`? And do we need to support this particular field as part of ADP-604?

```suggestion
        , mint = mempty  -- fixme: ADP-604
```

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -427,6 +431,25 @@ newtype ApiMaintenanceAction = ApiMaintenanceAction
     { gcStakePools :: ApiT PoolMetadataGCStatus
     } deriving (Eq, Generic, Show)
 
+data ApiAsset = ApiAsset
+    { policyId :: ApiT W.TokenPolicyId
+    , policyItem :: ApiT W.TokenName
```
jonathanknowles: If we're using the names in the specification, perhaps should this be "assetName"?

(Apologies if I'm missing some context here.)

```suggestion
    , assetName :: ApiT W.TokenName
```

lib/core/src/Cardano/Wallet/Api.hs
rvl: Some previous drafts of the API spec used `policy_item` to refer to assets within a policy. I didn't yet update all the names in the code to use `asset_name`.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy.hs
rvl: Of course, because Text is the wrong type.

lib/core/src/Cardano/Wallet/Api/Types.hs
rvl: Best place to comment on the API would be #2446. But it's one line of javascript to convert between flat and nested. EIther way would probably be fine.

lib/core/src/Cardano/Wallet/Api/Server.hs
rvl: Ok thanks - i'll use "mint" to match the terminology from Cardano.Api.
The API field is a placeholder to help Daedalus team with integration.

The correct place to review these names is in #2446 - because ApiTransaction just mirrors what's in `swagger.yaml`.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy.hs
rvl: Fixed.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1245,6 +1266,45 @@ selectCoinsForQuit ctx (ApiT wid) = do
...
+listAssets
+    :: forall ctx s k.
+        ( ctx ~ ApiLayer s k
+        )
+    => ctx
+    -> ApiT WalletId
+    -> Handler [ApiAsset]
+listAssets ctx (ApiT wid) =
+    Handler $ ExceptT $ withDatabase df wid $ \db -> runHandler $ do
+        let wrk = hoistResource db (MsgFromWorker wid) ctx
```
KtorZ: Use `withWorkerCtx` instead:

```hs
withWorkerCtx @_ @s @k ctx wid liftE liftE $ \wrk -> liftHandler $
        W.readWallet @_ @s @k wrk wid 
```

KtorZ: `withWorkerCtx` comes with some error handling that is common to all wallets accesses.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -427,6 +437,32 @@ newtype ApiMaintenanceAction = ApiMaintenanceAction
...
+    , metadata = ApiT <$> metadata_
+    }
+
+makeDisplayName :: W.TokenPolicyId -> W.TokenName -> Maybe W.AssetMetadata -> Text
+makeDisplayName pid an = \case
+    Just (W.AssetMetadata md) -> md
+    Nothing -> case T.decodeUtf8' assetNameBytes of
+        Right "" -> hexText $ getHash $ W.unTokenPolicyId pid
+        Right text -> text
+        Left _ -> hexText assetNameBytes
```
KtorZ: I am not sure to get the rationale for the `Nothing` branch in this function. The `assetName` isn't something that is supposed to be user-facing or displayed to user. 

I'd suggest to leave it `null` if there's no metadata (that is, simply leave it as `metadata.name`) or, to go with your first idea in the cryptpad spec: a bech32 encoded string comprised of the policyId | assetName, possibly hashed. Though, it's not really anymore displayable than the policyId and assetName.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1748,12 +1834,16 @@ instance ToJSON (ApiT SlotNo) where
...
-            genericParseJSON defaultRecordTypeOptions bytes
-        if isValidCoin (Coin $ fromIntegral c)
-            then return v
-            else fail $
+    parseJSON = withObject "AddressAmount " $ \v ->
+        prependFailure "parsing AddressAmount failed, " $
+        AddressAmount
+            <$> v .: "address"
+            <*> (v .: "amount" >>= validateCoin)
+            <*> v .:? "assets" .!= mempty
```
KtorZ, line 1829: :+1:

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -370,13 +369,13 @@ prefilterBlock b u0 = runState $ do
...
-    mkTxMeta :: Natural -> Direction -> TxMeta
-    mkTxMeta amt dir = TxMeta
+    mkTxMeta :: Coin -> Direction -> TxMeta
+    mkTxMeta amount dir = TxMeta
         { status = InLedger
         , direction = dir
         , slotNo = b ^. #header . #slotNo
         , blockHeight = b ^. #header . #blockHeight
-        , amount = Quantity amt
+        , amount = amount
```
KtorZ, line 378: This should be okay-ish for Ada / Coin because we know that the max supply is 42B. But we should generally be careful when summing token quantities. Even if each token individually fit in a Word64, their sum might not. So for anything that regards folds or sums, we should stick to `Natural` IMO.

KtorZ: At the very least, if changing this to `Coin`, I'd like to leave a note in the code about exactly this.

rvl: I have added a note to `Coin.hs`.

In `Model.hs`, any sums are limited to a single transaction, so it's all fine for ada amounts, and token amounts use Natural.

I would not be against changing newtype Coin to use Natural or Integer, just to remove all doubt about overflow.

lib/core/src/Cardano/Wallet/Api/Server.hs
rvl: Thanks, that's better.

lib/core/src/Cardano/Wallet/Api/Types.hs
rvl: Ok I have removed displayName for the time being.

KtorZ (approved):

rvl:
Rebased and squashed...

bors r+

Anviking:
_rebased_

bors r+

KtorZ:
bors retry

lib/core/src/Cardano/Wallet.hs
rvl: This feature is tracked by ADP-683.

PR #2448, opened 2021-01-14
Add `WithAdaOnly` constructor to `SelectionFilter` for `UTxOIndex`.

jonathanknowles:
# Issue Number

ADP-605
ADP-633

# Overview

This PR adds the `WithAdaOnly` constructor to the `SelectionFilter` type of `UTxOIndex`.

It's now possible to select a UTxO entry containing **only ada** in the following way:

```hs
UTxOIndex.selectRandom index WithAdaOnly
```

If there is such an entry, the function will return `Just ((i, o), reducedIndex)`.
If there is no such entry, the function will return `Nothing`.

# Motivation

For MA coin selection, we need the ability to randomly select UTxO entries containing only ada, and no other asset quantities.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
@@ -307,6 +307,8 @@ size = Map.size . utxo
 data SelectionFilter
     = Any
         -- ^ Select any UTxO entry from the entire set.
+    | WithAdaOnly
+        -- ^ Select any UTxO entry that only has ada and no other assets.
     | WithAsset AssetId
         -- ^ Select any UTxO entry that has a non-zero quantity of the specified
         -- asset.
@@ -333,6 +335,7 @@ selectRandom u selectionFilter =
     selectionSet :: Set TxIn
     selectionSet = case selectionFilter of
         Any -> Map.keysSet $ utxo u
+        WithAdaOnly -> entriesWithAdaOnly u
         WithAsset a -> entriesWithAsset a u
 
 --------------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
```diff
@@ -343,6 +346,12 @@ selectRandom u selectionFilter =
 -- Utilities
 --------------------------------------------------------------------------------
 
+entriesWithAdaOnly :: UTxOIndex -> Set TxIn
+entriesWithAdaOnly u = Map.foldl'
+    (Set.difference)
+    (Map.keysSet $ utxo u)
+    (NonEmptySet.toSet <$> index u)
```
KtorZ, line 353: This works because the index does not contain any Ada since our `AssetId` data-type does not capture Ada but only other foreign assets.

KtorZ (approved):

KtorZ:
bors merge

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
jonathanknowles: > This works because the index does not contain any Ada since our `AssetId` data-type does not capture Ada but only other foreign assets.

Precisely.

Note that each set within the index contains entries that **include** a particular non-ada asset, but those entries might also include other non-ada assets too.

So if we specify `selectRandom (WithAsset a) index`, we get all the entries that include asset `a`, but these entries might also include other assets.

If we wanted to go further and optimize this `WithAdaOnly` filter, we could also cache the "ada-only" `TxIn` set as an extra field in the `UTxOIndex` type. But it's probably not worth it for our current use cases.

PR #2449, opened 2021-01-14
Re-write LocalStateQuery client logic to eliminate acquire failures

Anviking:
# Issue Number

ADP-647, #2320

# Overview

- [x] Allow `send`-ing a composition of queries against a single acquired point, not just one.
- [x] Makes acquire failures practically impossible
- [x] Re-add multi-era support with reduced boilerplate
- [x] Some polish still needed
- [x] Re-add tracing of query times (less granular than before, but done)

# Comments

Pretty sure this _is_
- eliminating acquire failures
- the right direction

But also
- Might introduce a new set of problems

## Failures

I have run tests locally a lot on this branch. One new failure I _occasionally_ see is
```
  src/Test/Integration/Framework/DSL.hs:1797:7:
  1) API Specifications, SHELLEY_STAKE_POOLS, STAKE_POOLS_JOIN_01rewards - Can join a pool, earn rewards and collect them
       expected a successful response but got an error: DecodeFailure "{\"code\":\"created_invalid_transaction\",\"message\":\"That's embarrassing. It looks like I've created an invalid transaction that could not be parsed by the node. Here's an error message that may help with debugging: HardForkApplyTxErrFromEra S (S (S (Z (WrapApplyTxErr {unwrapApplyTxErr = ApplyTxError [LedgerFailure (DelegsFailure (WithdrawalsNotInRewardsDELEGS (fromList [(RewardAcnt {getRwdNetwork = Mainnet, getRwdCred = KeyHashObj (KeyHash \\\"9c0ff007dd21bbf24960bd12ae4009efb8cad076228ef1a54c7b5dbe\\\")},Coin 7010064794)])))]}))))\"}"
       While verifying (Status {statusCode = 500, statusMessage = "Internal Server Error"},Left (DecodeFailure "{\"code\":\"created_invalid_transaction\",\"message\":\"That's embarrassing. It looks like I've created an invalid transaction that could not be parsed by the node. Here's an error message that may help with debugging: HardForkApplyTxErrFromEra S (S (S (Z (WrapApplyTxErr {unwrapApplyTxErr = ApplyTxError [LedgerFailure (DelegsFailure (WithdrawalsNotInRewardsDELEGS (fromList [(RewardAcnt {getRwdNetwork = Mainnet, getRwdCred = KeyHashObj (KeyHash \\\"9c0ff007dd21bbf24960bd12ae4009efb8cad076228ef1a54c7b5dbe\\\")},Coin 7010064794)])))]}))))\"}"))

  To rerun use: --match "/API Specifications/SHELLEY_STAKE_POOLS/STAKE_POOLS_JOIN_01rewards - Can join a pool, earn rewards and collect them/"
```
Important part is `WithdrawalsNotInRewardsDELEGS`. So seems we are not aware of the rewards having already been spent. Maybe this PR makes the rewards slower to update, somehow, not sure.

Edit: digging through my notes, I _have_ seen this failure on another branch — once. I think this PR makes it more likely to occur, but not be inherently related, then.

### DIFF
--- a/lib/core/src/Ouroboros/Network/Client/Wallet.hs
+++ b/lib/core/src/Ouroboros/Network/Client/Wallet.hs
@@ -524,7 +524,7 @@ localStateQuery getTip queue =
         LSQ.SendMsgRelease (respond res >> clientStIdle)
             -- We /could/ read all LocalStateQueryCmds from the TQueue, and run
             -- them against the same tip, if re-acquiring takes a long time. As
-            -- of Jan 2020, it seems like queries themselves take significantly
+            -- of Jan 2021, it seems like queries themselves take significantly
             -- longer than the acquiring.
       where
           go
--- a/lib/core/src/Ouroboros/Network/Client/Wallet.hs
+++ b/lib/core/src/Ouroboros/Network/Client/Wallet.hs
@@ -31,12 +31,15 @@ module Ouroboros.Network.Client.Wallet
     , localTxSubmission
 
       -- * LocalStateQuery
+    , LSQ (..)
     , LocalStateQueryCmd (..)
-    , LocalStateQueryResult
     , localStateQuery
+    , query
+    , currentEra
 
       -- * Helpers
     , send
+    , sendAsync
 
       -- * Logs
     , ChainSyncLog (..)
@@ -45,6 +48,8 @@ module Ouroboros.Network.Client.Wallet
 
 import Prelude
 
+import Cardano.Api
+    ( AnyCardanoEra )
 import Cardano.BM.Data.Severity
     ( Severity (..) )
 import Cardano.BM.Data.Tracer
@@ -53,8 +58,11 @@ import Cardano.Slotting.Slot
     ( WithOrigin (..) )
 import Cardano.Wallet.Network
     ( NextBlocksResult (..) )
+import Control.Monad
+    ( ap, guard, liftM )
 import Control.Monad.Class.MonadSTM
     ( MonadSTM
+    , STM
     , TQueue
     , atomically
     , isEmptyTQueue
@@ -67,6 +75,8 @@ import Control.Monad.Class.MonadSTM
     )
 import Control.Monad.Class.MonadThrow
     ( MonadThrow )
+import Control.Monad.IO.Class
+    ( MonadIO )
 import Control.Tracer
     ( Tracer, traceWith )
 import Data.Functor
@@ -104,12 +114,7 @@ import Ouroboros.Network.Protocol.ChainSync.Client
 import Ouroboros.Network.Protocol.ChainSync.ClientPipelined
     ( ChainSyncClientPipelined (..) )
 import Ouroboros.Network.Protocol.LocalStateQuery.Client
-    ( ClientStAcquiring (..)
-    , ClientStQuerying (..)
-    , LocalStateQueryClient (..)
-    )
-import Ouroboros.Network.Protocol.LocalStateQuery.Type
-    ( AcquireFailure )
+    ( ClientStAcquiring (..), LocalStateQueryClient (..) )
 import Ouroboros.Network.Protocol.LocalTxSubmission.Client
     ( LocalTxClientStIdle (..), LocalTxSubmissionClient (..) )
 import Ouroboros.Network.Protocol.LocalTxSubmission.Type
@@ -131,8 +136,10 @@ import qualified Ouroboros.Network.Protocol.LocalStateQuery.Client as LSQ
 chainSyncFollowTip
     :: forall m block era. (Monad m)
     => (block -> era)
-    -> (Maybe era -> Tip block -> m ())
+    -> (era -> Tip block -> m ())
     -- ^ Callback for when the tip changes.
+    --
+    -- Will not be called on rollbacks(!), but on the subsequent roll-forward.
     -> ChainSyncClient block (Point block) (Tip block) m Void
 chainSyncFollowTip toCardanoEra onTipUpdate =
     ChainSyncClient (clientStIdle False)
@@ -165,13 +172,15 @@ chainSyncFollowTip toCardanoEra onTipUpdate =
     -- era-agnostic (for now at least!) which isn't a big deal really because
     -- the era will simply be updated on the next RollForward which follows
     -- immediately after.
+    --
+    -- NOTE: Let's try not updating the tip on rollbacks?
     clientStNext True = ClientStNext
-            { recvMsgRollBackward = doUpdate . const Nothing
-            , recvMsgRollForward  = doUpdate . Just . toCardanoEra
+            { recvMsgRollBackward = \_ _ -> ChainSyncClient $ clientStIdle True
+            , recvMsgRollForward  = doUpdate . toCardanoEra
             }
       where
         doUpdate
-            :: Maybe era
+            :: era
             -> Tip block
             -> ChainSyncClient block (Point block) (Tip block) m Void
         doUpdate era tip = ChainSyncClient $ do
@@ -440,15 +449,10 @@ chainSyncWithBlocks tr fromTip queue responseBuffer =
 --
 -- LocalStateQuery
 
--- | Command to send to the localStateQuery client. See also 'ChainSyncCmd'.
-data LocalStateQueryCmd block (m :: * -> *)
-    = forall state. CmdQueryLocalState
-        (Point block)
-        (Query block state)
-        (LocalStateQueryResult state -> m ())
-
--- | Shorthand for the possible outcomes of acquiring local state parameters.
-type LocalStateQueryResult state = Either AcquireFailure state
+---- | Command to send to the localStateQuery client. See also 'ChainSyncCmd'.
+data LocalStateQueryCmd block m = forall a. SomeLSQ
+    (LSQ block m a)
+    (a -> m ())
 
 -- | Client for the 'Local State Query' mini-protocol.
 --
@@ -475,61 +479,104 @@ type LocalStateQueryResult state = Either AcquireFailure state
 --        │       ┌───────────┴───┐             ┌────────┴───────┐
 --        └───────┤   Acquired    │────────────>│   Querying     │
 --                └───────────────┘             └────────────────┘
+-- FIXME: I think the diagram is wrong. It is possible to query multiple times
+-- while we're acquired.
 --
+-- NOTE: Using AnyCardanoEra arguably goes against the grain of the abstract
+-- block type. We might be able to use the (Header block) as replacemet for
+-- (AnyCardanoEra, Tip block).
 localStateQuery
-    :: forall m block . (MonadThrow m, MonadSTM m)
-    => TQueue m (LocalStateQueryCmd block m)
+    :: forall m block . (MonadIO m, MonadSTM m, Eq (Point block))
+    => STM m (AnyCardanoEra, Point block)
+       -- ^ A way to fetch the current node tip
+    -> TQueue m (LocalStateQueryCmd block m)
         -- ^ We use a 'TQueue' as a communication channel to drive queries from
         -- outside of the network client to the client itself.
         -- Requests are pushed to the queue which are then transformed into
         -- messages to keep the state-machine moving.
     -> LocalStateQueryClient block (Point block) (Query block) m Void
-localStateQuery queue =
+localStateQuery getTip queue =
     LocalStateQueryClient clientStIdle
   where
     clientStIdle
         :: m (LSQ.ClientStIdle block (Point block) (Query block) m Void)
-    clientStIdle = awaitNextCmd <&> \case
-        CmdQueryLocalState pt query respond ->
-            LSQ.SendMsgAcquire (Just pt) (clientStAcquiring query respond)
+    clientStIdle = do
+        cmd <- awaitNextCmd
+        (era, pt) <- atomically getTip
+        pure $ LSQ.SendMsgAcquire (Just pt) (clientStAcquiring cmd pt era)
 
     clientStAcquiring
-        :: forall state. Query block state
-        -> (LocalStateQueryResult state -> m ())
+        :: LocalStateQueryCmd block m
+        -> Point block
+        -> AnyCardanoEra
         -> LSQ.ClientStAcquiring block (Point block) (Query block) m Void
-    clientStAcquiring query respond = LSQ.ClientStAcquiring
-        { recvMsgAcquired = clientStAcquired query respond
-        , recvMsgFailure = \failure -> do
-                respond (Left failure)
-                clientStIdle
+    clientStAcquiring qry oldPt era = LSQ.ClientStAcquiring
+        { recvMsgAcquired = clientStAcquired qry era
+        , recvMsgFailure = \_failure -> do
+            (newEra, newPt) <- atomically $ do
+                t <- getTip
+                guard (((snd t) /= oldPt))
+                return t
+            pure $ LSQ.SendMsgAcquire (Just newPt) (clientStAcquiring qry newPt newEra)
         }
 
     clientStAcquired
-        :: forall state. Query block state
-        -> (LocalStateQueryResult state -> m ())
-        -> LSQ.ClientStAcquired block (Point block) (Query block) m Void
-    clientStAcquired query respond =
-        LSQ.SendMsgQuery query (clientStQuerying respond)
-
-    -- By re-acquiring rather releasing the state with 'MsgRelease' it
-    -- enables optimisations on the server side.
-    clientStAcquiredAgain
-        :: m (LSQ.ClientStAcquired block (Point block) (Query block) m Void)
-    clientStAcquiredAgain = awaitNextCmd <&> \case
-        CmdQueryLocalState pt query respond ->
-            LSQ.SendMsgReAcquire (Just pt) (clientStAcquiring query respond)
-
-    clientStQuerying
-        :: forall state. (LocalStateQueryResult state -> m ())
-        -> LSQ.ClientStQuerying block (Point block) (Query block) m Void state
-    clientStQuerying respond = LSQ.ClientStQuerying
-        { recvMsgResult = \result -> do
-            respond (Right result)
-            clientStAcquiredAgain
-        }
+        :: LocalStateQueryCmd block m
+        -> AnyCardanoEra
+        -> (LSQ.ClientStAcquired block (Point block) (Query block) m Void)
+    clientStAcquired (SomeLSQ cmd respond) era = go cmd $ \res -> do
+        LSQ.SendMsgRelease (respond res >> clientStIdle)
+            -- We /could/ read all LocalStateQueryCmds from the TQueue, and run
+            -- them against the same tip, if re-acquiring takes a long time. As
+            -- of Jan 2020, it seems like queries themselves take significantly
+            -- longer than the acquiring.
+      where
+          go
+              :: forall a. LSQ block m a
+              -> (a -> (LSQ.ClientStAcquired block (Point block) (Query block) m Void))
+              -> (LSQ.ClientStAcquired block (Point block) (Query block) m Void)
+          go (LSQPure a) cont = cont a
+          go (LSQry qry) cont = LSQ.SendMsgQuery qry $ LSQ.ClientStQuerying $ \res -> do
+              pure $ cont res
+              -- TODO: It would be nice to trace the time it takes to run the
+              -- queries.
+          go (LSQBind ma f) cont = go ma $ \a -> do
+              go (f a) $ \b -> cont b
+          go (LSQEra) cont = cont era
 
     awaitNextCmd :: m (LocalStateQueryCmd block m)
     awaitNextCmd = atomically $ readTQueue queue
+
+-- | Monad for composing local state queries for the node /tip/.
+data LSQ block (m :: * -> *) a where
+    LSQPure :: a -> LSQ block m a
+    LSQBind :: LSQ block m a -> (a -> LSQ block m b) -> LSQ block m b
+
+    -- | A local state query.
+    LSQry :: (Query block res) -> LSQ block m res
+
+    -- | The era of the node tip the query is run against.
+    LSQEra :: LSQ block m AnyCardanoEra
+
+instance Functor (LSQ block m) where
+    fmap = liftM
+
+instance Applicative (LSQ block m) where
+    pure  = LSQPure
+    (<*>) = ap
+
+instance Monad (LSQ block m) where
+    return = pure
+    (>>=)  = LSQBind
+
+--
+-- Helpers
+
+query :: (Query block res) -> LSQ block m res
+query = LSQry
+
+currentEra :: LSQ block m AnyCardanoEra
+currentEra = LSQEra
 
 --------------------------------------------------------------------------------
 --
@@ -606,6 +653,15 @@ send queue cmd = do
     tvar <- newEmptyTMVarIO
     atomically $ writeTQueue queue (cmd (atomically . putTMVar tvar))
     atomically $ takeTMVar tvar
+
+
+sendAsync
+    :: MonadSTM m
+    => TQueue m (cmd m)
+    -> (cmd m)
+    -> m ()
+sendAsync queue cmd = do
+    atomically $ writeTQueue queue cmd
 
 -- Tracing
 
--- a/lib/core/src/Ouroboros/Network/Client/Wallet.hs
+++ b/lib/core/src/Ouroboros/Network/Client/Wallet.hs
@@ -136,10 +136,8 @@ import qualified Ouroboros.Network.Protocol.LocalStateQuery.Client as LSQ
 chainSyncFollowTip
     :: forall m block era. (Monad m)
     => (block -> era)
-    -> (era -> Tip block -> m ())
+    -> (Maybe era -> Tip block -> m ())
     -- ^ Callback for when the tip changes.
-    --
-    -- Will not be called on rollbacks(!), but on the subsequent roll-forward.
     -> ChainSyncClient block (Point block) (Tip block) m Void
 chainSyncFollowTip toCardanoEra onTipUpdate =
     ChainSyncClient (clientStIdle False)
@@ -172,15 +170,13 @@ chainSyncFollowTip toCardanoEra onTipUpdate =
     -- era-agnostic (for now at least!) which isn't a big deal really because
     -- the era will simply be updated on the next RollForward which follows
     -- immediately after.
-    --
-    -- NOTE: Let's try not updating the tip on rollbacks?
     clientStNext True = ClientStNext
-            { recvMsgRollBackward = \_ _ -> ChainSyncClient $ clientStIdle True
-            , recvMsgRollForward  = doUpdate . toCardanoEra
+            { recvMsgRollBackward = doUpdate . const Nothing
+            , recvMsgRollForward  = doUpdate . Just . toCardanoEra
             }
       where
         doUpdate
-            :: era
+            :: Maybe era
             -> Tip block
             -> ChainSyncClient block (Point block) (Tip block) m Void
         doUpdate era tip = ChainSyncClient $ do
--- a/lib/core/src/Ouroboros/Network/Client/Wallet.hs
+++ b/lib/core/src/Ouroboros/Network/Client/Wallet.hs
@@ -38,7 +38,6 @@ module Ouroboros.Network.Client.Wallet
 
       -- * Helpers
     , send
-    , sendAsync
 
       -- * Logs
     , ChainSyncLog (..)
@@ -631,15 +630,6 @@ send queue cmd = do
     tvar <- newEmptyTMVarIO
     atomically $ writeTQueue queue (cmd (atomically . putTMVar tvar))
     atomically $ takeTMVar tvar
-
-
-sendAsync
-    :: MonadSTM m
-    => TQueue m (cmd m)
-    -> (cmd m)
-    -> m ()
-sendAsync queue cmd = do
-    atomically $ writeTQueue queue cmd
 
 -- Tracing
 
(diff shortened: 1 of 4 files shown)


Anviking:
When running locally, I'm consistently running into this new failure:
```
  src/Test/Integration/Scenario/API/Shelley/StakePools.hs:541:29:
  5) API Specifications, SHELLEY_STAKE_POOLS, STAKE_POOLS_JOIN_01 - Can rejoin another stakepool
       While verifying (Status {statusCode = 200, statusMessage = "OK"},Right (ApiWallet {id = ApiT {getApiT = WalletId {getWalletId = 4139b85eb0c5b1408f4f84e8570818298efde7ea}}, addressPoolGap = ApiT {getApiT = AddressPoolGap {getAddressPoolGap = 20}}, balance = ApiT {getApiT = WalletBalance {available = Quantity {getQuantity = 999998856100}, total = Quantity {getQuantity = 999998856100}, reward = Quantity {getQuantity = 0}}}, delegation = ApiWalletDelegation {active = ApiWalletDelegationNext {status = NotDelegating, target = Nothing, changesAt = Nothing}, next = [ApiWalletDelegationNext {status = Delegating, target = Just (ApiT {getApiT = PoolId {getPoolId = "\187\DC1L\179}u\250\ENQ&\ETX(\194\&5\163\218\226\149\163=\v\166t\165\235\RS>V\142"}}), changesAt = Just (ApiEpochInfo {epochNumber = ApiT {getApiT = EpochNo {unEpochNo = 40}}, epochStartTime = 2021-01-14 14:28:09 UTC})}]}, name = ApiT {getApiT = WalletName {getWalletName = "Fauce
t Wallet"}}, passphrase = Just (ApiWalletPassphraseInfo {lastUpdatedAt = 2021-01-14 14:27:37.265958 UTC}), state = ApiT {getApiT = Ready}, tip = ApiBlockReference {absoluteSlotNumber = ApiT {getApiT = SlotNo 1930}, slotId = ApiSlotId {epochNumber = ApiT {getApiT = EpochNo {unEpochNo = 38}}, slotNumber = ApiT {getApiT = SlotInEpoch {unSlotInEpoch = 30}}}, time = 2021-01-14 14:27:55 UTC, block = ApiBlockInfo {height = Quantity {getQuantity = 980}}}}))
       expected: Just (ApiT {getApiT = EpochNo {unEpochNo = 39}})
        but got: Just (ApiT {getApiT = EpochNo {unEpochNo = 40}})
```
which I think might be due to queuing problems.

Edit: I saw it outside of this PR. Might not be new then.

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -475,62 +479,99 @@ type LocalStateQueryResult state = Either AcquireFailure state
...
 
+data LocalStateQueryCmd block m = forall a. SomeLSQ
+    (LSQ block m a)
+    (a -> m ())
+
+-- Monad for composing local state queries for the node /tip/.
+data LSQ block (m :: * -> *) a where
+    LSQPure :: a -> LSQ block m a
+    LSQBind :: LSQ block m a -> (a -> LSQ block m b) -> LSQ block m b
+    LSQry :: (Query block res) -> LSQ block m res
```
Anviking, line 536: 👆 The LSQ monad that makes this all possible

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -475,62 +479,99 @@ type LocalStateQueryResult state = Either AcquireFailure state
...
+                        if oldPt == newPt
+                        -- FIXME: Not the most elegant approach of waiting for a
+                        -- new tip.
+                        then threadDelay 0.01 >> go
+                        else do
+                            pure $ LSQ.SendMsgAcquire newPt (clientStAcquiring qry newPt (retries + 1))
+                go
+            else do
+                --respond (Left failure)
+                -- TODO: Fail
```
Anviking: I don't think clients should have to care about acquire failures. If we can't get a tip acquired with infinite retries, then something is truly wrong.

(Should in that case have infinite retries though)

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -538,82 +546,17 @@ withNetworkLayerBase tr np addrInfo (versionData, _) action = do
           where
             mkErr = ErrStakeDistributionQuery . T.pack . show
 
-        queryStakeDistribution pt = \case
-            AnyCardanoEra ShelleyEra -> do
-                let cmd = CmdQueryLocalState pt (QueryIfCurrentShelley Shelley.GetStakeDistribution)
-                result <- handleQueryResult "GetStakeDistribution"
```
Anviking, line 547: The tracing would need to be re-added when _interpreting_ the LSQ monad.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -820,98 +766,25 @@ mkTipSyncClient tr np localTxSubmissionQ onTipUpdate onPParamsUpdate onInterpret
...
-
-            (_, Left mismatch) ->
-                traceWith tr $ MsgLocalStateQueryEraMismatch mismatch
+        queryLocalState Nothing    = return ()
+        queryLocalState (Just _era) = do
+            let qry = do
+                    mb <- LSQry $ QueryAnytimeMary GetEraStart
+                    sp <- fmap (slottingParametersFromGenesis . getCompactGenesis)
+                        <$> LSQry (QueryIfCurrentMary Shelley.GetGenesisConfig)
+                    pp' <- LSQry $ QueryIfCurrentMary Shelley.GetCurrentPParams
```
Anviking: This is actually readable!

(but, ok some complexity needs to be re-added)

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -475,62 +479,99 @@ type LocalStateQueryResult state = Either AcquireFailure state
...
+        { recvMsgAcquired = clientStAcquired qry
         , recvMsgFailure = \failure -> do
-                respond (Left failure)
+            if retries < 1000
+            then do
+                let go = do
+                        newPt <- getTip
+                        if oldPt == newPt
+                        -- FIXME: Not the most elegant approach of waiting for a
+                        -- new tip.
```
Anviking: Should probably pass in a `TChan m ((Point block, AnyCardanoEra))`

Edit: absolutely not! Then we might get old values of the tip.

Anviking: Duncan's STM seminar would suggest that we should use a plan `TVar` and retry inside STM!

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
Anviking: added `shelleyBased` and `byronOrShelleyBased` helpers

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -131,8 +151,10 @@ import qualified Ouroboros.Network.Protocol.LocalStateQuery.Client as LSQ
 chainSyncFollowTip
     :: forall m block era. (Monad m)
     => (block -> era)
-    -> (Maybe era -> Tip block -> m ())
+    -> (era -> Tip block -> m ())
```
KtorZ: I am afraid this may lead to nasty errors. The reason why I initially used a `Maybe` here was to stil allow `Tip` updates. This is perhaps confusing and we may consider having 2 distinct callbacks then: one for the `Tip` and one for the `era`.

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -475,62 +494,97 @@ type LocalStateQueryResult state = Either AcquireFailure state
...
+    LSQBind :: LSQ block m a -> (a -> LSQ block m b) -> LSQ block m b
+    LSQry :: (Query block res) -> LSQ block m res
+    LSQEra :: LSQ block m AnyCardanoEra
+
+instance Functor (LSQ block m) where
+    fmap = liftM
+
+instance Applicative (LSQ block m) where
+    pure  = LSQPure
+    (<*>) = ap
```
KtorZ, line 543: The fact that `ap` is defined in terms of `Monad` is utterly confusing to me :smile: ... I guess this is because I started with PureScript and not Haskell, where everything comes from `Applicative` whereas Haskell usually defines things from `Monad` and downwards.

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -475,62 +494,97 @@ type LocalStateQueryResult state = Either AcquireFailure state
...
     LocalStateQueryClient clientStIdle
   where
     clientStIdle
         :: m (LSQ.ClientStIdle block (Point block) (Query block) m Void)
-    clientStIdle = awaitNextCmd <&> \case
-        CmdQueryLocalState pt query respond ->
-            LSQ.SendMsgAcquire pt (clientStAcquiring query respond)
+    clientStIdle = do
+        cmd <- awaitNextCmd
+        (era, pt) <- second getTipPoint <$> atomically (readTVar tipVar)
```
KtorZ: Seems like this is always using `Point block` instead of `Tip block`, why not store it as such then and avoid all the `getTipPoint` calls.

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -475,62 +494,97 @@ type LocalStateQueryResult state = Either AcquireFailure state
...
-        { recvMsgAcquired = clientStAcquired query respond
-        , recvMsgFailure = \failure -> do
-                respond (Left failure)
-                clientStIdle
+    clientStAcquiring qry oldPt era retries = LSQ.ClientStAcquiring
+        { recvMsgAcquired = clientStAcquired qry era
+        , recvMsgFailure = \_failure -> do
+            (newEra, newPt) <- second getTipPoint <$> (atomically $ do
+                t <- readTVar tipVar
+                guard (getTipPoint (snd t) /= oldPt)
```
KtorZ: So, in case the points previously known is different, the transaction is retried until it gets updated?

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -475,62 +494,97 @@ type LocalStateQueryResult state = Either AcquireFailure state
...
+                return t)
+            pure $ LSQ.SendMsgAcquire newPt (clientStAcquiring qry newPt newEra (retries + 1))
```
KtorZ: This is counting retries, what for? Is the goal to cap that at some point?

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -475,62 +494,97 @@ type LocalStateQueryResult state = Either AcquireFailure state
...
+          go
+              :: forall a. LSQ block m a
+              -> (a -> (LSQ.ClientStAcquired block (Point block) (Query block) m Void))
+              -> (LSQ.ClientStAcquired block (Point block) (Query block) m Void)
+          go (LSQPure a) cont = (cont a)
+          go (LSQry qry) cont = LSQ.SendMsgQuery qry $ LSQ.ClientStQuerying $ \res -> do
+              pure $ cont res
+          go (LSQBind ma f) cont = go ma $ \a -> do
+              go (f a) $ \b -> cont b
+          go (LSQEra) cont = cont era
```
KtorZ: Hmmm. The continuation above will `respond`, that is, push a message into the response queue. This means that it is possible for callers to create "ill-programs" that will generate many responses in the queue. For instance `LsqPure a >> LsqPure a` will yield two messages. That's pretty bad as we may both create a memory leak but also fill the queue with responses which may no longer correspond to queries. 

Unless I am missing something?

lib/core/src/Ouroboros/Network/Client/Wallet.hs
Anviking: What kind of nasty errors? It simplifies things, and I didn't see how it could be bad.

That said, now with `STM` `guard` it might be pretty ok to keep the `Maybe`… if we guard for `Just` values on the era.

lib/core/src/Ouroboros/Network/Client/Wallet.hs
Anviking: Initially I capped it, but now it is not capped and this should be removed. (unless we'd want to trace the retries, but 🤷‍♂️ )

lib/core/src/Ouroboros/Network/Client/Wallet.hs
Anviking: Yes, that's how I understood it from Duncan's talk, and it seems to work.

KtorZ: I see that the `Alternative` instance for `STM` is defined as `empty = retry` so ... yes, I guess it works :sweat_smile:

lib/core/src/Ouroboros/Network/Client/Wallet.hs
KtorZ: If we rollback, it means that for a while, we'll be using a tip that is in the future.

lib/core/src/Ouroboros/Network/Client/Wallet.hs
Anviking: And the retrying is efficient in some sense, because it only retries when a previously read value has changed

lib/core/src/Ouroboros/Network/Client/Wallet.hs
Anviking: Ah, now I get what you mean. No — I don't think so.

The final `respond` will only be called by the top-level `go`.

```
LSQPure a >> LSQPure a
== LSQBind (LSQPure a) (\_ -> LSQPure a)
```

which should evaluate something like this:
```
go ma $ \a -> go (f a) $ \b -> cont b
go (LSQPure a) $ \a -> go (f a) $ \b -> cont b
go (f a) $ \b -> cont b
go (LSQPure a) $ \b -> cont b
cont a
respond a
```

Anviking: Could perhaps abstract the run-function to the top level and test it — but not sure how, on the top off my head.

Anviking: Also: The danger I see is if the queries contain a partial function. If it throws, I think the entire state machine will come to a halt.

lib/core/src/Ouroboros/Network/Client/Wallet.hs
Anviking: Can't `fmap` over `TVar`s, but should switch to taking `STM (AnyCardanoEra, Point block)` anyway, such that the read-onlyness is enforced.

Anviking: Changed to `STM (AnyCardanoEra, Point block)` where possible.

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -475,62 +479,103 @@ type LocalStateQueryResult state = Either AcquireFailure state
...
+        -- acquired should be more efficient.
+      where
+          go
+              :: forall a. LSQ block m a
+              -> (a -> (LSQ.ClientStAcquired block (Point block) (Query block) m Void))
+              -> (LSQ.ClientStAcquired block (Point block) (Query block) m Void)
+          go (LSQPure a) cont = cont a
+          go (LSQry qry) cont = LSQ.SendMsgQuery qry $ LSQ.ClientStQuerying $ \res -> do
+              pure $ cont res
+              -- TODO: It would be nice to trace the time it takes to run the
```
Anviking: This does not appear to be straight forward to do. There is no opportunity to run `IO` _before_ the query, but _after_ the point is acquired.

Anviking: I re-added top-level `bracketQuery` in the network layer that time the full `send` action (includes acquiring, and all the included queries).

lib/core/src/Ouroboros/Network/Client/Wallet.hs
Anviking: Luckily the entire wallet crashes in that case:
```
[cardano-wallet.network:Notice:282] [2021-01-18 19:34:23.67 UTC] Query stakePoolsSummary took 19.62426s (too slow)
cardano-wallet: ExceptionInLinkedThread ThreadId 12 todo
  error, called at src/Cardano/Wallet/Shelley/Network.hs:502:18 in cardano-wallet-2021.1.12-3Kq63VUOZn92WgCVghwpjO:Cardano.Wallet.Shelley.Network
  withNetworkLayerBase, called at src/Cardano/Wallet/Shelley/Network.hs:294:5 in cardano-wallet-2021.1.12-3Kq63VUOZn92WgCVghwpjO:Cardano.Wallet.Shelley.Network
  withNetworkLayer, called at src/Cardano/Wallet/Shelley.hs:274:9 in cardano-wallet-2021.1.12-3Kq63VUOZn92WgCVghwpjO:Cardano.Wallet.Shelley
```
So should be fine.

Anviking:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
Anviking: Tricky, cf https://github.com/input-output-hk/cardano-wallet/pull/2449#discussion_r559741621

lib/core/src/Ouroboros/Network/Client/Wallet.hs
Anviking: Should be done -> 30c569d

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -985,75 +841,43 @@ newRewardBalanceFetcher
...
-                handleBalanceResult defaultValue res
+
+        let qry = byronOrShelleyBased
+                (pure defaultValue)
+                (fromBalanceResult <$>
+                    LSQry (Shelley.GetFilteredDelegationsAndRewardAccounts
+                        $ Set.map toStakeCredential accounts))
+
+        -- FIXME: Re-add trace
+        --liftIO $ traceWith tr $ MsgAccountDelegationAndRewards deleg rewardAccounts
```
Anviking: ☝️

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -317,43 +309,39 @@ withNetworkLayerBase tr np conn (versionData, _) action = do
...
 
-    queryRewardQ <- connectDelegationRewardsClient handlers
-
-    (nodeTipChan, networkParamsVar, interpreterVar, localTxSubmissionQ) <-
+    (readNodeEraAndTip, networkParamsVar, interpreterVar, localTxSubmissionQ) <-
         connectNodeTipClient handlers
 
+    -- Alternative read-action which waits for the era to be Just and uses Point
+    -- instead of Tip.
+    let readNodeEraAndTip' = readNodeEraAndTip >>= \case
```
Anviking: Could possibly converge on some neat api of `STM` actions for node-tip state, defined in the tip sync client, but leaving it like this for now.

Anviking:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
Anviking: Fixed in 8f00b12bef91b08cf007ab33d0deb06917c1dd19

Anviking:
bors try

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -475,62 +475,109 @@ type LocalStateQueryResult state = Either AcquireFailure state
 --        │       ┌───────────┴───┐             ┌────────┴───────┐
 --        └───────┤   Acquired    │────────────>│   Querying     │
 --                └───────────────┘             └────────────────┘
+-- FIXME: I think the diagram is wrong. It is possible to query multiple times
+-- while we're acquired.
 --
+-- NOTE: Using AnyCardanoEra arguably goes against the grain of the abstract
+-- block type. We might be able to use the (Header block) as replacemet for
+-- (AnyCardanoEra, Tip block).
```
KtorZ: I think this is a `FIXME`, but the one above is `NOTE` :)

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -475,62 +475,109 @@ type LocalStateQueryResult state = Either AcquireFailure state
...
-                respond (Left failure)
-                clientStIdle
+    clientStAcquiring qry oldPt era = LSQ.ClientStAcquiring
+        { recvMsgAcquired = clientStAcquired qry era
+        , recvMsgFailure = \_failure -> do
+            (newEra, newPt) <- atomically $ do
+                t <- getTip
+                guard (((snd t) /= oldPt))
+                return t
+            pure $ LSQ.SendMsgAcquire newPt (clientStAcquiring qry newPt newEra)
```
KtorZ: Did you check with the network team whether it is better to `ReAcquire` or `Acquire` in this situation? I am still not 100% sure about the semantic we should expect from the state query protocol in this regard.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
KtorZ: On a small stylistic note, this logic is quite specific to the `currentNodeEra` handler so I'd simply leave in there rather than defining a top-level let binding.

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -475,62 +475,109 @@ type LocalStateQueryResult state = Either AcquireFailure state
...
-            clientStAcquiredAgain
-        }
+        :: LocalStateQueryCmd block m
+        -> AnyCardanoEra
+        -> (LSQ.ClientStAcquired block (Point block) (Query block) m Void)
+    clientStAcquired (SomeLSQ cmd respond) era = go cmd $ \res -> do
+        LSQ.SendMsgRelease (respond res >> clientStIdle)
+            -- We /could/ read all LocalStateQueryCmds from the TQueue, and run
+            -- them against the same tip, if re-acquiring takes a long time. As
+            -- of Jan 2020, it seems like queries themselves take significantly
```
Anviking:
```suggestion
            -- of Jan 2021, it seems like queries themselves take significantly
```
Oops, wrong year!

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -632,13 +547,16 @@ withNetworkLayerBase tr np conn (versionData, _) action = do
...
-            bracketTracer (contramap (MsgWatcherUpdate header) tr) $
-                cb header
+        let go oldTip = do
+                tip <- atomically $ do
+                    tip <- readNodeEraAndTip
+                    guard (oldTip /= tip)
+                    return tip
+                cb (toBlockHeader . snd $ tip)
+                go tip
+        go (Just $ AnyCardanoEra ByronEra, TipGenesis )
```
KtorZ: we are loosing the `MsgWatcherUpdate` trace from this new approach.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -804,165 +723,102 @@ mkTipSyncClient
...
-                    traceWith tr $ MsgInterpreter interpreter
-                    onInterpreterUpdate interpreter
-
-        W.GenesisParameters
+    let queryParams = do
+            mb <- currentEra >>= \case
+                AnyCardanoEra ShelleyEra ->
+                    LSQry $ QueryAnytimeShelley GetEraStart
+                _ ->
+                    LSQry $ QueryAnytimeMary GetEraStart
```
KtorZ, line 720: Hmm. This seems wrong (not that this PR is changing anything but we should look into this).

Anviking: Yes. I know. I'm not sure what it's supposed to do. When we change `hardfork_at` to return all eras should make it right though.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -804,165 +723,102 @@ mkTipSyncClient
...
-                $ localTxSubmissionClientPeer
-                $ localTxSubmission localTxSubmissionQ
+        -- NOTE: If these queries are not sent with sendAsync, we'd cause
+        -- deadlock, as the local state query client might need to waiy for the
+        -- tip client to find a new tip.
+        --
+        -- NOTE 2: interpeter is updated every block. This is far more often than
+        -- necessary.
+        sendAsync localStateQueryQ $ SomeLSQ queryParams onPParamsUpdate'
+        sendAsync localStateQueryQ $ SomeLSQ queryInterpreter onInterpreterUpdate
```
KtorZ: Arguably, we only need those once per epoch.

KtorZ (approved):
Looks okay, although I admit being mildly scared by the number of changes in the network layer. Yet, I think it's going in the right direction and it makes the code simpler to reason about anyway; so new problems should also be easier to catch.

lib/core/src/Ouroboros/Network/Client/Wallet.hs
KtorZ: Living in the past :p

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
Anviking: It would be cool to add some logic for that. But while it would probably be easy, it would probably not  be completely trivial either.

Anviking: For instance, we can't observe the node tip `STM` and filter out the epoch-changes, since we're not guaranteed to see all intermediate states with `STM`.

In that case we'd probably need the tip-sync client to have specific logic for running actions on epoch-boundaries.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -985,75 +841,45 @@ newRewardBalanceFetcher
...
-          , (AnyCardanoEra, Tip (CardanoBlock StandardCrypto)) -> IO ()
+          , (Maybe AnyCardanoEra, Tip (CardanoBlock StandardCrypto)) -> IO ()
             -- Call on tip-change to refresh
           )
 newRewardBalanceFetcher tr gp queryRewardQ =
     newObserver (contramap MsgObserverLog tr) fetch
   where
     fetch
-        :: (AnyCardanoEra, Tip (CardanoBlock StandardCrypto))
+        :: (Maybe AnyCardanoEra, Tip (CardanoBlock StandardCrypto))
```
Anviking, line 809: Hm. This `Tip` is only used for logging. And the log message might lie, since it is not necessarily the tip we're going to use.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
Anviking: 👍 thanks

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
Anviking: I removed it from the trace.

lib/core/src/Ouroboros/Network/Client/Wallet.hs
rvl: I fixed the state diagram.

lib/core/src/Ouroboros/Network/Client/Wallet.hs
rvl: I was just looking at the LocalStateQuery types in ouroboros-network, and there is only one option after MsgFailure, and that's MsgReAcquire. In the Acquired state, the shortest way to move the state to another block is with MsgReAcquire.

rvl:
I like that this PR
1. has generalised the query types across eras.
2. has eliminated the possibility of acquire failures by removing the Point argument from CmdQueryLocalState (as suggested by @KTorz on ADP-647).

But otherwise I am finding the changes confusing, and have some suggestions to simplify:

1. To keep the LocalState current, have localStateQuery follow the tip variable via a skip queue, ReAcquiring whenever the tip changes.

2. You have noted the awkwardness of passing AnyCardanoEra around. I agree - let's remove it. We can just run the query for all eras, starting at Mary, and stopping when one succeeds. Then we don't need to use the result of the era query in subsequent queries.

3. Now, if queries can no longer fail, why do we need Monad? If queries no longer have a Point argument, then they can no longer change the state. They have no effects. The purpose of Monad is to do different things based on the results of previous actions - that's no longer required.

   But we do want to run multiple queries against the same LocalState in sequence and collect their results. So we need Applicative only.

Anviking:
> To keep the LocalState current, have localStateQuery follow the tip variable via a skip queue, ReAcquiring whenever the tip changes.

I briefly tried doing something like that. But we cannot run monadic actions in the acquired state, before deciding what query to send. So it seems that they really want us to idle only in the idle state — though not sure.

> We can just run the query for all eras, starting at Mary, and stopping when one succeeds. Then we don't need to use the result of the era query in subsequent queries.

There's another option: `QueryHardFork GetCurrentEra`. This seems much neater to me than trying all eras for every query. 

I am giving it a shot. I think the code will turn out cleaner, and also fix a on-startup-blocking problem I discovered yesterday with the current approach of getting the era from the tip sync client.

> if queries can no longer fail, why do we need Monad?

As you said (although you argued for its removal):

> [..] result of the era query in subsequent queries.

Besides the `currentEra >>= \case` use case, I hope the flexibility might be useful in the future as the differences between the eras grow larger.

I don't see how restricting ourselves to Applicative would make things better.

Anviking:
> before deciding what query to send. So it seems that they really want us to idle only in the idle state.

That said, we _could_ probably circumvent that by sending a Era query, and waiting for the next query in the response action… 🤔 

Edit: they will change it -> https://input-output-rnd.slack.com/archives/CFKLUH4R0/p1611326573008900

Anviking:
Slightly concerned about the rare `WithdrawalsNotInRewardsDELEGS` failures, but not sure it's directly caused by this PR, and think it's still worth merging.

bors r+

Anviking:
bors try

PR #2450, opened 2021-01-14
Multi-Asset Coin Selection

jonathanknowles:
# Issue Number

ADP-605

## Overview

This PR implements the **Random-Round-Robin** coin selection algorithm for multi-asset UTxO sets.

The **Random-Round-Robin** algorithm is inspired by the **Random-Improve** algorithm, but with some differences to accommodate UTxO sets with multiple assets.

## Top-Level Algorithm Description

The **Random-Round-Robin** algorithm considers the sum of all outputs **collectively**, selecting inputs to cover the total sum of all asset quantities, rather than running a separate selection for each input. It therefore drops the cardinality restriction of the original Random-Improve algorithm implementation.

### Steps

1. When selecting inputs, consider tokens in **round-robin fashion**, selecting one input per token before moving to the next token, to reduce the chance of over-selecting for any particular token.
2. For each token quantity under consideration, select enough inputs to cover at least 100% of that quantity, but aim to get as close to a target of 200% as possible. When we can make no further improvement for a given token, we eliminate that token from the round-robin selection process. An _improvement_ is defined as an additional selection that takes the total selected token quantity closer to 200% of the output token quantity, but not further away.
3. The round-robin selection phase terminates when we can make no further improvement for any token in the set under consideration.
4. After the selection phase is over, divide any excess token quantities (inputs − outputs) into change bundles, where:
    -  there is exactly one change bundle for each output.
    - the quantity of a given token in a change bundle is proportional to the quantity of that token in the corresponding output (modulo rounding).
    - the total quantity of a given token across all change bundles is equal to the total excess quantity of that token.
5. Redistribute additionally-selected tokens not present in the original outputs to the change bundles, where:
    - if there are fewer quantities for a given token than the number of change bundles, include these quantities without changing them.
    - if there are more quantities for a given token than the number of change bundles, repeatedly coalesce the smallest pair of quantities together until the total number of quantities is exactly equal to the number of change bundles.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
@@ -1,13 +1,22 @@
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedLabels #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
 module Cardano.Wallet.Primitive.CoinSelection.MA.RoundRobin
     (
+    -- * Running a selection
+      runSelection
+    , SelectionState (..)
+
+    -- * Running a selection step
+    , runSelectionStep
+    , SelectionLens (..)
+
     -- * Making change
-      makeChange
+    , makeChange
     , makeChangeForCoin
     , makeChangeForPaymentAssets
     , makeChangeForSurplusAssets
@@ -25,6 +34,9 @@ module Cardano.Wallet.Primitive.CoinSelection.MA.RoundRobin
     , runRoundRobin
     , runRoundRobinM
 
+    -- * Utility functions
+    , distance
+
     ) where
 
 import Prelude
@@ -41,6 +53,10 @@ import Cardano.Wallet.Primitive.Types.TokenMap
     ( AssetId, TokenMap )
 import Cardano.Wallet.Primitive.Types.TokenQuantity
     ( TokenQuantity (..) )
+import Cardano.Wallet.Primitive.Types.UTxOIndex
+    ( SelectionFilter (..), UTxOIndex (..) )
+import Control.Monad.Random.Class
+    ( MonadRandom (..) )
 import Data.Function
     ( (&) )
 import Data.Functor.Identity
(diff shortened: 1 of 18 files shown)


lib/numeric/src/Cardano/Numeric/Rounding.hs
```diff
@@ -0,0 +1,31 @@
...
+      -- ^ Round down to the nearest integral value.
+    deriving (Eq, Show)
+
+-- | Use the given rounding direction to round the given fractional value,
+--   producing an integral result.
+--
+round :: (RealFrac a, Integral b) => RoundingDirection -> a -> b
+round = \case
+    RoundUp -> ceiling
+    RoundDown -> floor
```
KtorZ: To be frank, I am very doubtful about this module :grimacing: ... I do think `ceiling` and `floor` are so common across all programming languages that there's no possible ambiguity about what those functions do.

lib/numeric/src/Cardano/Numeric/Util.hs
```diff
@@ -0,0 +1,208 @@
...
+-- [Sum 7, Sum 8]
+--
+-- >>> padCoalesce [Sum 8, Sum 4, Sum 2, Sum 1] (replicate 1 ())
+-- [Sum 15]
+--
+padCoalesce :: forall m a. (Monoid m, Ord m)
+    => NonEmpty m
+    -- ^ Source list
+    -> NonEmpty a
+    -- ^ Target list
```
KtorZ, line 73: Does this really have to be a list? It seems that we only really want a length here. The examples make it every more awkward for they all use `replicate n ()` :sweat_smile:

lib/numeric/src/Cardano/Numeric/Util.hs
```diff
@@ -0,0 +1,208 @@
...
+--
+--  2.  The sum of elements in the resulting list is equal to the original
+--      natural number:
+--
+--      >>> fmap sum (partitionNatural n weights) == Just n
+--
+--  3.  The size of each element in the resulting list is within unity of the
+--      ideal proportion.
+--
+partitionNatural
```
KtorZ, line 133: To my understanding, `partition` has the sense of `division`. It feels a bit weird to me in this context, wouldn't `distribute` better capture the behavior :thinking: ?

jonathanknowles: > To my understanding, partition has the sense of division. It feels a bit weird to me in this context, wouldn't distribute better capture the behavior thinking ?

In this case, the term "partition" is taken from number theory: 

https://en.wikipedia.org/wiki/Partition_(number_theory)

Quote:
> In number theory and combinatorics, a partition of a positive integer n, also called an integer partition, is a way of writing n as a sum of positive integers.

Admittedly, in our particular case, we are computing a special kind of partition, where the elements of the partition are proportional to the sizes of the specified weights (modulo rounding).

The term "distribute" would probably also work here, but IMO this name doesn't capture (as nicely) the property that the total sum is preserved (whereas the term "partition" implies this, if we assume the definition above).

lib/numeric/src/Cardano/Numeric/Rounding.hs
jonathanknowles: > To be frank, I am very doubtful about this module grimacing ... I do think ceiling and floor are so common across all programming languages that there's no possible ambiguity about what those functions do.

Good point. These functions could easily be moved into the `Util` module, as they're currently only used by the `partitionNatural` function.

lib/numeric/src/Cardano/Numeric/Util.hs
jonathanknowles: > Does this really have to be a list?

It definitely doesn't have to be a list!

However, accepting a non-empty list makes it nice to re-use from functions that **already** have a non-empty list of the target size.

For example:
> `changeForSurplusAssets` calls `makeChangeForSurplusAssets` with the already-existing `outputBundles`, which calls `partitionTokenQuantity`, which eventually calls `partitionNatural`. So we end up coalescing (or padding) the surplus asset quantities into exactly `n` chunks, where `n` is number of output bundles. 

As a bonus, using `NonEmpty` here allows us to make this function **total**, as the minimum number of elements in a non-empty list is always one.

If we accepted an `Int` (or something else), then presumably we'd need to choose from one of the following non-ideal options:

- make the function return `Maybe` if the target length is `< 1` (in which case the caller would need to deal with the awkward `Nothing` value that they know they shouldn't receive).
- make the function throw an error if the target length is `< 1`.
- do something else, like taking `max 1 targetLength`. But if the caller supplies a `targetLength` less than 1, that presumably represents a programming error that we'd want to know about.

lib/numeric/src/Cardano/Numeric/Rounding.hs
jonathanknowles: @KtorZ Fixed in [this change](https://github.com/input-output-hk/cardano-wallet/compare/0879623a8eaa3f3d3f9667d4c3ebd557a2fea138..b49a386a7347b21e4c9d9fc144e95e843b27a1ed).

KtorZ: Note that my main concern wasn't much about why this is in a separate module, but why this has to be defined altogether :). I think that any places that uses `round RoundUp` or `round RoundDown` could simply use `ceiling` and `floor`. That latter option is actually clearer and more idiomatic. There's no need to build an abstraction over it IMO.

lib/numeric/src/Cardano/Numeric/Util.hs
KtorZ: > As a bonus, using NonEmpty here allows us to make this function total, as the minimum number of elements in a non-empty list is always one.

Fair enough. That's a good argument.

lib/numeric/src/Cardano/Numeric/Rounding.hs
jonathanknowles: > I think that any places that uses round `RoundUp` or round `RoundDown` could simply use `ceiling` and `floor`. 

Very true.

However, my original goal here was simply to make this code more self-documenting: (taken from `partitionNatural`)

```hs
roundings :: NonEmpty RoundingDirection
roundings =
    applyN shortfall (NE.cons RoundUp) (NE.repeat RoundDown)
  where
    shortfall
        = fromIntegral target
        - fromIntegral @Integer
            (F.sum $ round RoundDown <$> portionsUnrounded)
```

IMO, the purpose of `roundings` is much clearer and more self-documenting with the `RoundingDirection` type. It's a non-empty list of rounding directions (either up or down).

One alternative would be to write something like this:
```hs
roundings :: (RealFrac a, Integral b) => NonEmpty (a -> b)
```

But this makes the intent much less obvious IMO, and would probably require an additional comment to explain what's going on.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -0,0 +1,575 @@
...
+    , utxoAvailable
+        :: !UTxOIndex
+    }
+    deriving (Eq, Show)
+
+-- | The result of performing a successful selection.
+--
+data SelectionResult = SelectionResult
+    { inputsSelected
+        :: !(NonEmpty (TxIn, TxOut))
```
KtorZ, line 140: We'll need to add (i.e. will do in my addition PR) the _modified outputs_ here too. That is, each output needs to also include a minimum Ada value. This is not typically set by users. One would say "I want to send 10 Apples to Alice", but the actual output would really be  "10 Apples & 1 Ada". In practice, it shouldn't affect users _too much_ because the extra Ada would be coming from selected inputs which contains that particular assets (and therefore, bring with them some existing ada deposit). The only user who pay the full cost of this is the one forging the initial asset supply. 

I think that the `performSelection` is the right moment to do this. As an input, one can provide the `outputsToCover` as requested by the user, but we would slightly alter each output to include a minimum coin value (unless a coin value is explicitly specified) and run the coin selection with this altered output set. I think we should also make it crystal clear in the API doc.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -0,0 +1,575 @@
...
+        pure $ Left $ BalanceInsufficient $ BalanceInsufficientError
+            {balanceAvailable, balanceRequired}
+    | otherwise =
+        Right . mkResult <$> runSelection utxoAvailable balanceRequired
+  where
+    balanceAvailable :: TokenBundle
+    balanceAvailable = view #balance utxoAvailable
+
+    balanceRequired :: TokenBundle
+    balanceRequired = F.foldMap (view #tokens) outputsToCover
```
KtorZ, line 217: Needs to add the extra cost coming from fees and deposits here.

KtorZ (approved):

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -170,32 +199,41 @@ performSelection
...
-        case NE.nonEmpty (UTxOIndex.toList selected) of
-            Nothing ->
-                unableToSelectAnyInputsError
-            Just inputsSelected ->
-                SelectionResult
+    mkResult :: SelectionState -> Either SelectionError SelectionResult
+    mkResult SelectionState {selected, leftover}
+        | not (balanceRequired `leq` balanceSelected) =
+            Left $ SelectionInsufficient $ SelectionInsufficientError
+                {inputsSelected = UTxOIndex.toList selected, balanceRequired}
```
KtorZ, line 223: I like this approach. We could have callers to performSelection possibly retry on `SelectionInsufficient` errors. If we're close to the required balance, the selection may still succeed with a different selection of inputs.

jonathanknowles: > We could have callers to performSelection possibly retry on SelectionInsufficient errors. If we're close to the required balance, the selection may still succeed with a different selection of inputs.

I think that's a great idea. 👍🏻

PR #2451, opened 2021-01-14
Multi-asset coin selection additions

KtorZ:
# Issue Number

ADP-605

# Overview

- 96c077d86c250322c116da06f8a90a203a9292d2
  :round_pushpin: **add few properties about filtering 'TokenMap'**
  
- 01969b450c0147d7771ae45118c327c85b41f7fd
  :round_pushpin: **allows for multiple filters when selecting matching quantities**
    This will be crucial for Hardware devices and other clients who won't support MA initially. It is a only a best effort solution as this would only prioritize ada-only UTxO but eventually reach out to multi-assets UTxO if there's not enough ada-only utxo to cover for a particular payment. It is also generally good for other purposes to only select Ada utxo if we can to avoid needlessly moving tokens around.

- bd2772651f24d192692130f0f148726990d097dd
  :round_pushpin: **Make 'prop_runSelection_UTxO_muchMoreThanEnough' blind about 'UTxOIndex'**
    Failures generate unreadable output which are pointless. For this test, it'll be better to rely on counterexample built for
  analyzing failures.

- 4ee5c71b6e7d4eeb98de482f160a1f52716d6ee4
  :round_pushpin: **allow 'runSelection' to take Ada from an extra source**
    Although, it only does so if at least one input has already been selected. I've adjusted properties accordingly, generating an extra source when it made sense, and leaving it as 'Nothing' for some others.

- c640a393a7970d3aefadce2a6e72393aa4bc2df6
  :round_pushpin: **revise 'makeChange' to take fees, deposits, withdrawals and minimum ada values into account**
    The strategy is as follows:

  1. We start by computing the surplus amount coming from the difference
     between the total requested outputs and the total incoming value.

  2. From there, we start by creating as many change outputs as there
     are outputs, and assign surplus tokens proportionally to the
     quantity present in each outputs. That is, to create change outputs
     that are following a similar distribution as the requested outputs.

  3. For tokens which are not present in any requested outputs (coming
     from selected inputs), we forward them into pre-existing change
     outputs, coalescing small values as necessary so that there the
     number of change outputs remains equal to the number of requested
     outputs.

  4. Then, we assign some coins from the surplus to each change output
     so that each can meet the minimum coin value required. At this
     point, we may return stop and return 'Nothing' if there's not
     enough surplus coins to assign all required minimum to all change
     outputs.

  5. Then, we split the remaining coin amount across all change outputs,
     proportionally to the size of each corresponding output.

- b6c442b8883ae4cbfbafb004d57d18df10d50db0
  :round_pushpin: **WIP: update 'performSelection' to work with fees, withdrawals, deposits and minimum coin value**
      This is still WIP for the various properties related to
    'performSelection' needs to be adjusted. New ones may also be added as
    necessary.

    It adds / tweaks quite a lot of the existing function but in essence it
    is doing pretty much the same thing, with retries. Let's give some
    details:

    - First of all, I've abstracted away as functions the functions that
      will need to calculate the minimum coin value needed for each
      output, and the one that would calculate the total cost (fee +
      deposits). These are trivially implemented using functions from
      the ledger and our current transaction layer.

    - Then, the main challenge is to construct a selection with change
      outputs in such way that the cost and the minimum coin value have
      been handled correctly.

    - As a starter, the algorithm augments every non-ada requested
      output with a minimum coin value. We expect users who want to
      transfer some asset A to only request a payment of asset A, and
      not worry about how much ada is necessary. So we take care of it.
      Requested outputs are also fully known, so we can do this step
      early on before running the round-robin selection.

    - Then, we run the round-robin selection which, in case of success,
      yields a set of inputs such that the sum of all inputs covers at
      least for all requested outputs, possibly more.

    - Then, need to create change outputs with the excess from the
      previous steps. However, we also need to make sure that we leave
      enough money out of the change to pay for the transaction cost
      (fees + deposits). Yet, the full cost of a selection is only
      known _after_ the selection has been made.

    - We start off by creating a pre-selection, that is, a selection
      with change outputs when the cost and minimum coin value are
      considered null. This is guaranteed to succeed because we have
      already selected enough inputs to cover to all requested output.
      The pre-selection gives us a baseline from which we can compute
      the final cost.

    - Then, we repeatedly attempt to construct a final selection, this
      time considering the actual cost (calculated from the
      pre-selection) and minimum coin value. If it fails, we select
      another ada-only input and recurse. The algorithm exits when we
      run out of ada-only input or, when a final selection could be
      successfully constructed. Note that for this to work, the cost
      function must not be affected by a change in the change outputs
      quantities. To enforce this, the cost function does not have to
      full 'TokenBundle' but only to '[AssetId]' and should be able to
      yield an approximate cost from this only. Adding more coins to a
      change output should therefore not change the cost of a selection,
      and ensure that the algorithm eventually converges.


lib/core/test/unit/Cardano/Wallet/Primitive/Types/TokenMapSpec.hs
```diff
@@ -302,6 +323,38 @@ prop_toNestedList_fromNestedList :: TokenMap -> Property
...
+
+-- | Verify that we can partition the token map using the predicate, and recover
+-- the original map by computing the union of both partitions.
+prop_filter_partition :: Fun AssetIdF Bool -> TokenMap -> Property
+prop_filter_partition f b =
+    let
+        l = TokenMap.filter (applyFun f . AssetIdF) b
+        r = TokenMap.filter (not . applyFun f . AssetIdF) b
+    in
+        (l <> r) === b
```
jonathanknowles, line 348: 👍🏻

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -130,23 +130,34 @@ runSelection available minimumBalance =
...
         }
 
     coinSelectionLens :: SelectionLens m SelectionState
     coinSelectionLens = SelectionLens
         { currentQuantity = coinQuantity . selected
         , minimumQuantity = fromIntegral $ unCoin minimumCoinQuantity
-        , selectQuantity = selectMatchingQuantity Any
+        , selectQuantity = selectMatchingQuantity
+            [ WithAdaOnly
+            , Any
```
jonathanknowles, line 538: 👍🏻 

I also thought of the same idea earlier on, but ran out of time before adding it.

I think it's a nice improvement, as it reduces the chance of "accidentally" selecting an output with some asset we're not actually interested in.

I suspect we could do the same for other assets: preferentially select inputs with only that asset (and no others), before moving on to selecting inputs that contain that asset (but may also contain other assets).

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -130,23 +130,34 @@ runSelection available minimumBalance =
...
+        -- others.
     -> SelectionState
     -> m (Maybe SelectionState)
-selectMatchingQuantity f s =
-    fmap updateState <$> UTxOIndex.selectRandom (leftover s) f
+selectMatchingQuantity []    _ = pure Nothing
+selectMatchingQuantity (h:q) s = do
+    UTxOIndex.selectRandom (leftover s) h >>= \case
+        Just s' -> pure $ Just $ updateState s'
+        Nothing -> selectMatchingQuantity q s
```

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -289,8 +290,8 @@ prop_runSelection_UTxO_moreThanEnough (Small index) = monadicIO $ do
         cutAssetSetSizeInHalf balanceAvailable
 
 prop_runSelection_UTxO_muchMoreThanEnough
-    :: Large UTxOIndex -> Property
-prop_runSelection_UTxO_muchMoreThanEnough (Large index) =
+    :: Blind (Large UTxOIndex) -> Property
```
jonathanknowles: Good idea. 👍🏻  I was not aware of the `Blind` wrapper type.

In the same spirit, even the "small" UTxO indices can be hard to read (when output through `show`). Perhaps it's worth putting `Blind` on all of them.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -116,6 +119,9 @@ runSelection available minimumBalance =
         , leftover = available
         }
 
+    extraSource :: Natural
+    extraSource = maybe 0 (fromIntegral . unCoin) mExtraSource
```
jonathanknowles: **Suggestion:** Perhaps this could be nested in a `where` clause under the definition of `coinSelectionLens`?

Alternatively, we could call this `extraCoinSource`, to make it clear that it's a source of `Coin`, rather than some other asset.

lib/core/src/Cardano/Wallet/Primitive/Types/Tx/Gen.hs
```diff
@@ -104,12 +105,12 @@ genTxInLargeRange = TxIn
 genTxOutSmallRange :: Gen TxOut
 genTxOutSmallRange = TxOut
     <$> genAddressSmallRange
-    <*> genTokenBundleSmallRange
+    <*> genTokenBundleSmallRange `suchThat` (/= TokenBundle.empty)
```
jonathanknowles: Nice addition, very concise and clear. 👍🏻

jonathanknowles (commented):
Hi @KtorZ 

This looks good so far. I left a small number of comments.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
KtorZ: > preferentially select inputs with only that asset

That's actually a good idea. And it shouldn't be too hard to implement.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
KtorZ: Generally speaking, the default `Show` instance is good for small data-types. When data-structures become complex, it is just unreadable (though sometimes useful to have a copy-pastable counter-example). As a rule of thumb, I prefer using few counter-examples with cherry-picked pieces of information that makes asserting the problem easier.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
KtorZ: Sounds good :+1:

lib/core/src/Cardano/Wallet/Primitive/Types/Tx/Gen.hs
KtorZ: I was puzzled by this one test failing, though i had run the test suite already a few time. Drilling a bit more into the problem, I found this. I don't think we should ever have an empty token bundle except as a starting point for folding, though there might be cases where it makes sense :thinking: ? The TxOut is definitely not one of them IMO, so I've only banished them there (and not in the token bundle generator itself).

jonathanknowles: > The TxOut is definitely not one of them IMO, so I've only banished them there (and not in the token bundle generator itself).

Yes, it's definitely the right decision to banish them here.

The token bundle generator itself still needs to generate `TokenBundle.empty` occasionally, since some of the compatibility properties rely on that to meet their coverage checks.

jonathanknowles: @KtorZ I've just realized that this is not **_quite_** correct.

I believe the correct filter condition should be that **the ada value is non-zero**. This is because it's possible to have a non-empty token bundle with an ada quantity of zero, which we don't want to see in `TxOut`.

I've added a commit to my PR that makes this change [here](https://github.com/input-output-hk/cardano-wallet/pull/2450/commits/912ea634fd671985f5578979c3bd55518d997957), as it's necessary to make the property tests for `performSelection` pass.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -257,30 +264,71 @@ prop_runSelection_UTxO_notEnough (Small index) = monadicIO $ do
...
+            assert True
+        Just balanceRequested | balanceRequested == TokenBundle.empty -> do
+            SelectionState {selected,leftover} <-
+                run $ runSelection extraSource index balanceRequested
+            let balanceSelected = view #balance selected
+            let balanceLeftover = view #balance leftover
+            assert $ balanceLeftover == view #balance index
+            assert $ balanceSelected == TokenBundle.empty
+        Just balanceRequested -> do
+            monitor (cover 80 True "sometimes there are Ada")
```
jonathanknowles: I suspect that if this case is never matched, the coverage check will not cause this property to fail, as it will never be reached.

Perhaps it's better to write a coverage check at the top-level (so it always fires) that looks something like this:

```hs
prop_runSelection_UTxO_extraSourceUsed extraSource (Small index) =
    checkCoverage $
    monitor $ cover 80 (isJust balanceRequested)
        "sometimes there are Ada"
    monadicIO $ case almostEverything of
    
```

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -0,0 +1,1015 @@
...
+        (assetsRequested `Set.isProperSubsetOf` assetsAvailable)
+        "assetsRequested ⊂ assetsAvailable"
+    monitor $ cover 50 (Set.size assetsRequested >= 4)
+        "size assetsRequested >= 4"
+    monitor $ counterexample $ unlines
+        [ "balance available: " <> show balanceAvailable
+        , "balance requested: " <> show balanceRequested
+        , "balance selected:  " <> show balanceSelected
+        , "balance leftover:  " <> show balanceLeftover
+        ]
```
jonathanknowles: Here we can take advantage of the `Buildable` instance for `TokenBundle`, which produces output that's easier to read than `Show`:

```suggestion
    monitor $ counterexample $ unlines
        [ "balance available:"
        , pretty (TokenBundle.Nested balanceAvailable)
        , "balance requested:"
        , pretty (TokenBundle.Nested balanceRequested)
        , "balance selected:"
        , pretty (TokenBundle.Nested balanceSelected)
        , "balance leftover:"
        , pretty (TokenBundle.Nested balanceLeftover)
        ]
```

The `Nested` instance produces output that's a bit more concise, as assets are grouped by token policy.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -0,0 +1,1015 @@
...
+            (assetsRequested `Set.isProperSubsetOf` assetsAvailable)
+            "assetsRequested ⊂ assetsAvailable"
+        monitor $ cover 50 (Set.size assetsRequested >= 4)
+            "size assetsRequested >= 4"
+        monitor $ counterexample $ unlines
+            [ "balance available: " <> show balanceAvailable
+            , "balance requested: " <> show balanceRequested
+            , "balance selected:  " <> show balanceSelected
+            , "balance leftover:  " <> show balanceLeftover
+            ]
```
jonathanknowles, line 723: LIkewise, we can use `pretty . TokenBundle.Nested` here to make the output of `counterexample` more readable.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
KtorZ: Indeed! Good point.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -114,52 +114,66 @@ import qualified Data.Set as Set
...
-data SelectionResult = SelectionResult
+data SelectionResult (change :: *) = SelectionResult
     { inputsSelected
         :: !(NonEmpty (TxIn, TxOut))
+        -- ^ A (non-empty) list of selected inputs from the wallet's UTxO.
+    , outputsConsidered
+        :: !(NonEmpty TxOut)
+        -- ^ Users aren't typically aware of the minimum coin value needed for
+        -- each output. So, we do alter any requested non-ada output to include
+        -- a minimum ada value.
```
jonathanknowles: **Proposal**: rename this to `outputsCovered`.

**Justification**:

* it more clearly indicates that these outputs have been **_covered_**, rather than just **_considered_** (which might give readers the impression that this is somehow a subset of the original set).
* it more clearly relates to `outputsToCover`.

```suggestion
    , outputsCovered
        :: !(NonEmpty TxOut)
        -- The result of adjusting 'outputsToCover' so that each output has been
        -- augmented with a minimum ada value.
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -114,52 +114,66 @@ import qualified Data.Set as Set
...
     , changeGenerated
-        :: !(NonEmpty TokenBundle)
-    , utxoRemaining
-        :: !UTxOIndex
```
jonathanknowles: **Proposal:** reinstate the `utxoRemaining` field.

**Justification:**

1. When we eventually update the wallet's UTxO state, we can use `utxoRemaining` as the source (as we do for the existing coin selection algorithms).
2. Our test suite uses this field to check that the UTxO has been appropriately depleted:
    ```hs
    assert $ utxoAvailable
        == UTxOIndex.insertMany inputsSelected utxoRemaining
    assert $ utxoRemaining
        == UTxOIndex.deleteMany (fst <$> inputsSelected) utxoAvailable
    ```
     (Strictly speaking, this is not essential, but it serves as an extra sanity check.)

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -168,36 +182,158 @@ data BalanceInsufficientError = BalanceInsufficientError
...
+    -- possible to make a particular output valid.
+    outputsConsidered :: NonEmpty TxOut
+    outputsConsidered = flip fmap outputsToCover $ \out ->
+        out { Tx.tokens = augmentBundle (Tx.tokens out) }
+      where
+        augmentBundle bundle =
+            if TokenBundle.getCoin bundle == Coin 0 then
+                bundle { coin = minCoinValueFor (view #tokens bundle) }
+            else
+                bundle
```
jonathanknowles: **Question**: What if the user has specified a non-zero coin value that is lower than the result of `minCoinValueFor`?

(Apologies if I have missed something elsewhere that would check for this.)

Would it not be safer to write:
```suggestion
        augmentBundle bundle = bundle
            {coin = max (coin bundle) (minCoinValueFor (view #tokens bundle))}
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -168,36 +182,158 @@ data BalanceInsufficientError = BalanceInsufficientError
 --
 performSelection
     :: forall m. (HasCallStack, MonadRandom m)
-    => SelectionCriteria
-    -> m (Either SelectionError SelectionResult)
-performSelection SelectionCriteria {outputsToCover, utxoAvailable}
+    => (TokenMap -> Coin)
+        -- ^ A function which computes the minimum required Ada coins for a
+        -- particular output.
```
jonathanknowles: Just a small fix to the wording:
```suggestion
        -- ^ A function which computes the minimum required ada quantity for a
        -- particular output.
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -168,36 +182,158 @@ data BalanceInsufficientError = BalanceInsufficientError
...
+    --
+    -- So, in order to not duplicate the logic from 'makeChange', we first
+    -- calculate a pre-selection considering the case where we have no fees to
+    -- pay, and no minimum value. This is guaranteed to succeed and will yield
+    -- the final shape of the change outputs (modulo amounts).
+    --
+    -- From there, we can successively re-iterate and construct a final
+    -- unbalanced selection that is leaving enough money out for fees. For this
+    -- method to work (efficiently), the function estimate the cost of a
+    -- selection MUST NOT depends on change token quantity. That is, adding more
```
jonathanknowles: Just some minor edits.
```suggestion
    -- method to work (efficiently), the function that estimates the cost of a
    -- selection MUST NOT depend on change token quantity. That is, adding more
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -168,36 +182,158 @@ data BalanceInsufficientError = BalanceInsufficientError
...
+    -- change to an output should not make it more expensive.
```
jonathanknowles: I'm still a bit confused as to the meaning of this sentence: 😅 

> "That is, adding more change to an output should not make it more expensive."

Does it mean: "increasing the token quantities of a change output should not make it more expensive"?

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -168,36 +182,158 @@ data BalanceInsufficientError = BalanceInsufficientError
 --
 performSelection
     :: forall m. (HasCallStack, MonadRandom m)
-    => SelectionCriteria
-    -> m (Either SelectionError SelectionResult)
-performSelection SelectionCriteria {outputsToCover, utxoAvailable}
+    => (TokenMap -> Coin)
+        -- ^ A function which computes the minimum required Ada coins for a
+        -- particular output.
+    -> (SelectionResult [AssetId] -> Coin)
```
jonathanknowles: **Proposal**: change this type to `SelectionResult (Set AssetId)`

**Justification**:

1. The type makes it clear that asset IDs cannot be duplicated in any given change output.
2. Fewer `Set a -> [a]` conversions needed.
3. Similarly, `transformChange` can be simplified to:
    ```hs
    transformChange :: NonEmpty TokenBundle -> NonEmpty (Set AssetId)
    transformChange = fmap TokenBundle.getAssets
    ```

jonathanknowles (commented):
Hi @KtorZ 

This is a partial review. I'm still working through and understanding the changes, but wanted to record my existing comments in case of browser malfunction. 😄

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -168,36 +182,158 @@ data BalanceInsufficientError = BalanceInsufficientError
...
+      where
+        noMinimumCoin :: TokenMap -> Coin
+        noMinimumCoin = const (Coin 0)
+
+        noCost :: Coin
+        noCost = Coin 0
+
+        transformChange :: NonEmpty TokenBundle -> NonEmpty [AssetId]
+        transformChange =
+            fmap (fmap fst . snd . TokenBundle.toFlatList)
```
jonathanknowles: **Proposal:** Narrow the range of this function from `SelectionResult [AssetId]` to `NonEmpty (Set AssetId)`, and give it a name that indicates it only generates initial change assets.

**Justification**: This function doesn't generate new values for `inputsSelected` or `outputsConsidered`, and the values of these fields in the `SelectionResult` can only ever be what was supplied by the caller. But this isn't made apparent by the type signature, which (arguably) increases the work required to understand this function.

By narrowing the function's range, we can make it clear to readers that this function is only responsible for generating an initial value for `changeGenerated`, and even simplify the definition:

```suggestion
    mkInitialChangeAssets :: NonEmpty (TxIn, TxOut) -> NonEmpty (Set AssetId)
    mkInitialChangeAssets inputsSelected = maybe
        (invariantResultWithNoCost inputsSelected)
        (fmap TokenBundle.getAssets)
        (makeChange noMinimumCoin noCost
            extraCoinSource
            (view #tokens . snd <$> inputsSelected)
            (view #tokens <$> outputsCovered)
      where
        noMinimumCoin :: TokenMap -> Coin
        noMinimumCoin = const (Coin 0)

        noCost :: Coin
        noCost = Coin 0
```

The call site then becomes:
```patch
+        let inputsSelected = mkInputsSelected selected
+        let changeGenerated = mkInitialChangeAssets inputsSelected
         let preSelection :: SelectionResult (Set AssetId)
-            preSelection = mkPreSelection (mkInputsSelected selected)
+            preSelection = SelectionResult
+                {inputsSelected, outputsCovered, changeGenerated}
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -168,36 +182,158 @@ data BalanceInsufficientError = BalanceInsufficientError
...
+
+    -- | This function starts from an initial pre-selection as a way to evaluate
+    -- the cost of a final selection, and then call 'makeChange' repeatedly until
+    -- it succeeds. Between each call, it selects an extra Ada-only input to
+    -- inject additional Ada to construct change address. Eventually it returns
+    -- Just a final selection, or nothing if it runs out of Ada-only inputs.
```
jonathanknowles: Just some minor corrections:

(I'm guessing that "change address" is meant to be "change output"?)

```suggestion
    -- | This function starts from an initial pre-selection as a way to evaluate
    -- the cost of a final selection, and then calls 'makeChange' repeatedly until
    -- it succeeds. Between each call, it selects an extra ada-only input to
    -- inject additional ada to construct change outputs. Eventually it returns
    -- just a final selection, or nothing if it runs out of ada-only inputs.
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
KtorZ: I did move it aside so that we could re-use the `SelectionResult` type for the fee calculation function. The fee calculation would do nothing with the leftover UTxO so it'd be misleading to pass it around. 

Though, I ended up having to tweak the `SelectionResult` to also introduce a type-parameter which maybe is a sign that in the end, we simply need a different type altogether.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -315,211 +483,206 @@ runSelectionStep lens s
...
+                excessAssets
+
+        let changeForUnknownAssets :: NonEmpty TokenMap
+            changeForUnknownAssets = F.foldr
+                (NE.zipWith (<>) . makeChangeForUnknownAsset outputTokens)
+                (TokenMap.empty <$ outputTokens)
+                unknownAssets
+
+        let change :: NonEmpty TokenMap
+            change = NE.zipWith (<>) changeForKnownAssets changeForUnknownAssets
```
jonathanknowles: **Proposal:** move these constants out of the `do` block and into the `where` clause.

**Justification:** their values are determined entirely by the arguments to `makeChange`.

By placing them in the `do` block, it invites the reader to suspect that their evaluation must occur in a monadic context, perhaps in some particular order. (Especially because there is already a `where` clause.)

After moving these to the where clause, the monadic part of the calculation is reduced to just what is essential:

```hs
    | otherwise = do
        (bundles, remainder) <-
            excessCoin `subtractCoin` requiredCost
            >>=
            runStateT (sequence (StateT . assignCoin minCoinValueFor <$> change))

        let changeForCoins :: NonEmpty TokenBundle
            changeForCoins = TokenBundle.fromCoin
                <$> makeChangeForCoins outputCoins remainder

        pure (NE.zipWith (<>) bundles changeForCoins)
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
KtorZ: > Question: What if the user has specified a non-zero coin value that is lower than the result of minCoinValueFor?

We ought to catch that and tell them they did wrong. I haven't added that check yet but it is indeed the right place to do it. However, instead of taking the max, we should error. We shouldn't change outputs that are manually specified by users, but tell them instead about their mistakes.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
KtorZ: :+1:

jonathanknowles (commented):
Hi @KtorZ

Another partial review. ⛅ 

Will post more comments as I work through the code. (I'm slowly getting there!)

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
KtorZ: > Does it mean

It means precisely that indeed! This may be confusing unless you have a deeper look / understanding about how fees are actually calculated. This has to do with the way objects are encoded to binary with CBOR. CBOR uses a variable-length encoding for numbers. That is, small numbers are encoded with 1 byte, bigger number with 3, or 5 or 9. 

Our fee calculation function now often consider the worse case to avoid the hassle coming from being _too precise_. Dealing with this has been a nightmare for a long time, everytime we would increase the quantity of a change output, fees would also increase, and as a result, we would need to reduce change outputs, and fees would decrease... and so forth ad infinitum.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
KtorZ: > Proposal: move these constants out of the do block and into the where clause.

I do disagree here. I find the code far easier to follow when these things are part of the body as let clauses. Then, the reader doesn't need to bump back-and-forth between somewhere in the where clause and the body. Generally speaking, I do prefer putting in where clauses utility functions, or transformation on arguments that are used through the entire function body. Though, these are local binding that are better understood with their surrounding context.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
KtorZ: Yes, I am not yet satisfied with this type signature and the many transformations happening here. I left a small note in the code about it and I intend to revise it. Re-using the `SelectionResult` type may not be the right path to follow at all.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
jonathanknowles: In that case, I'm not clear why the responsibility for assigning/checking minimum ada values of outputs should be in the `performSelection` function.

What if we were to instead:

1. Move the responsibility for assigning/checking the minimum ada values of outputs to the caller of the `performSelection` function.
2. Make it a pre-condition of `performSelection` that every output already satisfies `minCoinValueFor`.

Advantages:

1. Testing `performSelection` becomes more straightforward, as we don't have to worry about the augmenting-outputs-with-ada behaviour.
2. We wouldn't need the distinction between `outputsToCover` and `outputsCovered`, and this field could disappear from the `SelectionResult` record, making the documentation more straightforward.
3. The augmenting-outputs-with-ada functionality, if placed in its own function elsewhere, can be tested in isolation. (Right now, it's hard to test, as it's embedded within `performSelection`.)

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -315,211 +483,206 @@ runSelectionStep lens s
...
---------------------------------------------------------------------------------
-
-partitionCoin
+-- | Construct a list of coin change outputs based on a distribution given as
+-- first input. If the input distribution is filled with 0, this function throws
+-- a runtime error.
+--
+-- The output list has always the same size as the input list, and the sum of
+-- its values is always exactly equal to the 'Coin' value given as 2nd argument.
+makeChangeForCoins
```
jonathanknowles: **Suggestion**: rename this function to `makeChangeForCoin`

**Justification**: there is only one coin.

```suggestion
makeChangeForCoin
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -315,211 +483,206 @@ runSelectionStep lens s
...
+-- `runSelection`. The pre-condition (2) is a undirect consequence of assigning
+-- a minimum UTxO value to every output token bundle.
 makeChange
-    :: NonEmpty TokenBundle
+    :: (TokenMap -> Coin)
+        -- A function which computes the minimum required Ada coins for a
+        -- particular output.
+    -> Coin
+        -- ^ The minimal (and optimal) delta between the total Ada inputs and
+        -- total Ada outputs. This typically captures fees plus key deposits.
```
jonathanknowles: I reckon we could improve this comment by making it a bit more explicit:

```suggestion
        -- ^ The minimal (and optimal) delta between the total ada balance
        -- of all input bundles and the total ada balance of all output and
        -- change bundles, where:
        --
        --    delta = getCoin (fold inputBundles)
        --          - getCoin (fold outputBundles)
        --          - getCoin (fold changeBundles)
        --
        -- This typically captures fees plus key deposits.
        --
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -315,211 +483,206 @@ runSelectionStep lens s
...
+    -> Maybe Coin
+        -- ^ An extra source of Ada, if any.
+    -> NonEmpty TokenBundle
         -- ^ Token bundles of selected inputs
     -> NonEmpty TokenBundle
         -- ^ Token bundles of original outputs
-    -> NonEmpty TokenBundle
-        -- ^ Change bundles
```
jonathanknowles: This is just a suggestion, by why not use the existing `MakeChangeData` type here (or some variation of it)?

**Advantages**:

1. It's arguably easier to read the function documentation if it explicitly refers to named fields such as `requiredCost` and `inputBundles` and so on (assuming someone is reading the generated Haddock docs).
2. We can reuse this type from the test suite, instead of having to define it there.
3. Sites that call `makeChange` will become a bit more readable, as instead of a long list of arguments, each argument will be named:

```hs
            mChangeGenerated = makeChange MakeChangeData
                { minCoinValue = minCoinValueFor
                , requiredCost = costFor s
                , extraCoinSource
                , inputBundles = view #tokens . snd <$> inputsSelected
                , outputBundles = view #tokens <$> outputsCovered
                }
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
jonathanknowles: Fair enough! I can sympathize with your arguments here. Perhaps `let` versus `where` is a matter of personal style.

Though in this case, the reader is still having to jump back and forth between the `where` clause and the function body, as several constants used by `let` expressions are defined in the `where` clause (for example `totalInputValue`, `knownAssetIds`, and so on).

Perhaps we could consolidate them all to one section?

jonathanknowles (commented):
Hi @KtorZ 

Some more comments.

Still working through the rest of the PR!

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
jonathanknowles: > I did move it aside so that we could re-use the SelectionResult type for the fee calculation function. The fee calculation would do nothing with the leftover UTxO so it'd be misleading to pass it around.
> Though, I ended up having to tweak the SelectionResult to also introduce a type-parameter which maybe is a sign that in the end, we simply need a different type altogether.

I agree it's probably a good idea to use a separate type internally for the fee calculation, and then keep the `SelectionResult` as a simple record type without a type parameter. 👍🏻 

It certainly makes it easier for callers of `performSelection` to understand, I think!

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
jonathanknowles: Understood!

In that case, I think a simple tweak to the comment would clear up any confusion:

```patch
- adding more change to an output should not make it more expensive.
+ increasing the token quantities of a change output should not make it more expensive.
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
KtorZ: Indeed. I thought of it, and generally want to that for any function with more than 3 arguments. I wish records were easier to create on-the-fly in Haskell. I'll move the `MakeChangeData` from the test code to here.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -371,30 +453,72 @@ prop_runSelection_UTxO_notEnough (Small index) = monadicIO $ do
...
     assert $ balanceLeftover == TokenBundle.empty
   where
     balanceRequested = view #balance index
 
+prop_runSelection_UTxO_extraSourceUsed
+    :: Maybe Coin
+    -> Small UTxOIndex
+    -> Property
+prop_runSelection_UTxO_extraSourceUsed extraSource (Small index) = do
+    let hasSomeAda = maybe False (/= TokenBundle.empty) almostEverything
```
jonathanknowles: Asserting that a `TokenBundle` is non-empty is unfortunately not enough to guarantee that it has some ada. We must check the `Coin` value directly.

I made a small fix in: 781fc4809b892eb11f8b30751ab2ee09bc8acd0f. Feel free to cherry-pick it if it works for you (or just copy and paste if you prefer).

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -371,30 +453,72 @@ prop_runSelection_UTxO_notEnough (Small index) = monadicIO $ do
...
         (assetsRequested `Set.isProperSubsetOf` assetsAvailable)
         "assetsRequested ⊂ assetsAvailable"
     monitor $ cover 50 (Set.size assetsRequested >= 4)
         "size assetsRequested >= 4"
-    assert $ balanceRequested `leq` balanceSelected
+    monitor $ counterexample $ unlines
+        [ "balance available: " <> pretty (Flat balanceAvailable)
+        , "balance requested: " <> pretty (Flat balanceRequested)
+        , "balance selected:  " <> pretty (Flat balanceSelected)
+        , "balance leftover:  " <> pretty (Flat balanceLeftover)
```
jonathanknowles, line 689: `Flat` works perfectly fine here, though `Nested` would give more concise output (fewer lines), which could make failures slightly easier to comprehend.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -114,9 +124,27 @@ import qualified Data.Set as Set
...
+-- output. Increasing or decreasing a particular change quantity must not change
+-- the estimation.
+--
+data SelectionSkeleton = SelectionSkeleton
+    { inputsSkeleton
+        :: !UTxOIndex
+    , outputsSkeleton
+        :: !(NonEmpty TxOut)
+    , changeSkeleton
+        :: !(NonEmpty (Set AssetId))
```
jonathanknowles, line 163: I do like the `SelectionSkeleton` name.  😄 

All in all, I think the types are much clearer than before. 👍🏻

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -125,29 +153,70 @@ data SelectionCriteria = SelectionCriteria
...
+    } deriving (Generic, Eq, Show)
+
+-- | Indicates that some of the specified outputs aren't valid and do not
+-- contain the minimum coin value expected by the protocol.
+--
+-- See also: 'prepareOutputs'.
+data InsufficientMinCoinValueError = InsufficientMinCoinValueError
+    { insufficientlyCoveredOutput
+        :: !TxOut
+        -- ^ The invalid output which doesn't hold enough coin.
```
jonathanknowles: **Suggestion**: call this field `outputWithInsufficientAda` (or something similar)

**Reasoning**: this name more clearly relates to the fact that the ada amount is insufficient. (The output could still be **covered**, if we were to run the algorithm.)

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -125,29 +153,70 @@ data SelectionCriteria = SelectionCriteria
...
+    :: (TokenMap -> Coin)
+    -> NonEmpty TxOut
+    -> NonEmpty TxOut
+prepareOutputsWith minCoinValueFor = fmap $ \out ->
+    out { Tx.tokens = augmentBundle (Tx.tokens out) }
+  where
+    augmentBundle bundle =
+        if TokenBundle.getCoin bundle == Coin 0
+        then bundle { coin = minCoinValueFor (view #tokens bundle) }
+        else bundle
```
jonathanknowles, line 275: I think the code is much clearer now that the responsibility for preparing outputs has been moved out of `performSelection`. 👍🏻

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -125,29 +153,70 @@ data SelectionCriteria = SelectionCriteria
...
+
+-- | Prepare a set of outputs requested by users into valid Cardano outputs.
+-- That is, any output in Cardano needs to hold a minimum coin quantity (to
+-- prevent a certain kind of attack flooding the network with worthless UTxOs).
+--
+-- However, users do not typically specify a minimum ada value themselves.
+-- One would rather send '10 Apple' and not '10 Apple & 1.2 Ada'. Therefore,
+-- unless a coin value is explicitely specified, we do assign a coin value
+-- manually for each non-ada output. That value is the minimum value
+-- possible to make a particular output valid.
```
jonathanknowles: I made some small fixes to the documentation comments here: 4a97d1ad181f0fc35129e8ee71a85091a5c876d2

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -168,36 +237,133 @@ data BalanceInsufficientError = BalanceInsufficientError
...
 performSelection
     :: forall m. (HasCallStack, MonadRandom m)
-    => SelectionCriteria
+    => (TokenMap -> Coin)
+        -- ^ A function which computes the minimum required ada quantity for a
+        -- particular output.
+    -> (SelectionSkeleton -> Coin)
+        -- ^ A function which computes the extra cost corresponding to a given
+        -- selection. This function must not depends on the value of each change
+        -- output.
```
jonathanknowles: Ditto (regarding documentation comments). See 4a97d1ad181f0fc35129e8ee71a85091a5c876d2.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -168,36 +237,133 @@ data BalanceInsufficientError = BalanceInsufficientError
...
+    -- calculate a pre-selection considering the case where we have no fees to
+    -- pay, and no minimum value. This is guaranteed to succeed and will yield
+    -- the final shape of the change outputs (modulo amounts).
+    --
+    -- From there, we can successively re-iterate and construct a final
+    -- unbalanced selection that is leaving enough money out for fees. For this
+    -- method to work (efficiently), the function that estimates the cost of a
+    -- selection MUST NOT depend on change token quantity. That is, increasing
+    -- the token quantities of a change output must not make it more expensive.
+    predictChange
```
jonathanknowles, line 390: I really like the motivation section, but I think this comment is still missing just a short overview of what the `predictChange` function does.

For example, if I call `predictChange` with a `UTxOIndex`, what can I expect to get back?

I wonder, would it be possible to add:

- a short summary stating what the function returns (as a function of the input), along the lines of "This function creates ...",
- a statement of the post-conditions and properties?

jonathanknowles (commented):
Hi @KtorZ 

Another partial review.

With the latest changes, things look more understandable (and it should be easier to test).

Will leave more comments as I get further through!

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
KtorZ: I dislike `Nested` as I think it gives a wrong representation of the assets. I prefer to consider assets as a product between an AssetId and a Quantity. The PolicyId -> AssetName hierarchy is implementation detail which is irrelevant for reasoning about assets IMO.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
KtorZ: That's my 4th naming attempt I think :sweat_smile:. I went through `PreSelection`, `SelectionPartialResult` and `SelectionBootstrap` until I settled on `SelectionSkeleton` which I am pretty happy with as it convey the right idea.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
KtorZ: Fair enough

jonathanknowles: I fixed this in ca1d961b304fe0453a1ac7c9141adc60c0e6975e.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
jonathanknowles: Fair enough! I don't have a strong feeling either way.

I've converted all the counter examples to use the `Flat` representation in 781ca4d0a5eb544724bfb6a9fae117c6bede0962.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -363,211 +642,256 @@ runSelectionStep lens s
...
+        let (excessCoin, excessAssets) = TokenBundle.toFlatList excess
+
+        let unknownAssets =
+                Map.toList $ F.foldr discardKnownAssets mempty inputBundles
+
+        let changeForKnownAssets :: NonEmpty TokenMap
+            changeForKnownAssets = F.foldr
+                (NE.zipWith (<>) . makeChangeForKnownAsset outputTokens)
+                (TokenMap.empty <$ outputTokens)
+                excessAssets
```
jonathanknowles, line 719: I suspect that evaluating `changeForKnownAssets` is doing (some) extra work here unnecessarily, because it is folding over the entire set of excess asset quantities, rather than just the excess quantities of assets present in the outputs.

In the original implementation (#2450), we partitioned the assets into:
- **payment assets** (assets found in the original outputs).
- **surplus assets** (assets not found in the original assets, but that were nonetheless selected and therefore "surplus" to our requirements).

I think we could keep the original partitioning scheme here, and filter out the non-payment assets before calling `changeForKnownAssets`.

As for the terminology, I was a bit confused (at first) by the terms "known" and "unknown". I'm uncertain as to the sense in which a token can be "unknown". Perhaps I'm missing some context?

I'm not sure that the change in terminology is justified here. But assuming that we feel the term "surplus" is ambiguous, I think we could improve the understandability of the code by using the terms:

- `paymentAssets` (assets found in the original outputs)
- `nonPaymentAssets` (assets not found in the original outputs)

Thoughts?

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -363,211 +642,256 @@ runSelectionStep lens s
...
-        -- the total input value is greater than the total output value:
-        totalInputValue `TokenBundle.unsafeSubtract` totalOutputValue
-
-    change :: NonEmpty TokenBundle
-    change
-        = TokenBundle.fromCoin <$> changeForCoin
-        & NE.zipWith (<>) (TokenBundle.fromTokenMap <$> changeForPaymentAssets)
-        -- Here we sort in ascending order of coin value so that surplus assets
-        -- and coin values are in increasing order in the resulting bundles:
-        & NE.sortWith TokenBundle.getCoin
```
jonathanknowles, line 407: I'm not yet sure as to whether this property is maintained by the updated implementation, or whether we want to maintain it.

(Unfortunately, I did not have time to write a property test to check it.)

The original motivation was to avoid producing a UTxO distribution where either:
- massive token quantities are bundled with tiny coin quantities; or
- tiny token quantities are bundled with massive coin quantities.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -363,211 +642,256 @@ runSelectionStep lens s
...
-        surplusAssets :: [(AssetId, TokenQuantity)]
-        surplusAssets = filter
-            ((`Set.notMember` paymentAssetIds) . fst)
-            (TokenMap.toFlatList . view #tokens =<< NE.toList inputBundles)
-
     -- Identifiers of assets included in outputs.
-    paymentAssetIds :: Set AssetId
-    paymentAssetIds = TokenBundle.getAssets totalOutputValue
+    knownAssetIds :: Set AssetId
+    knownAssetIds = TokenBundle.getAssets totalOutputValue
```
jonathanknowles, line 801: I must admit I'm a bit curious as to the reason for the change in terminology here!

I think the term `paymentAssetIds` is actually very clear, as these assets correspond to assets found in the payments/outputs.

But it's not clear to me in what sense a token can be thought of as "unknown". I think this term is potentially a bit confusing.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -619,58 +878,348 @@ genMakeChangeData = flip suchThat isValidMakeChangeData $ do
...
+    else
+        totalChangeValue === TokenQuantity 0
+  where
+    changes = makeChangeForKnownAsset weights (asset, quantity)
+    totalChangeValue = F.fold ((`TokenMap.getQuantity` asset) <$> changes)
 
-prop_makeChangeForSurplusAssets_sort
-    :: Map AssetId (NonEmpty TokenQuantity) -> NonEmpty () -> Property
-prop_makeChangeForSurplusAssets_sort assetQuantities target = property $
-    inAscendingPartialOrder (makeChangeForSurplusAssets assetQuantities target)
```
jonathanknowles, line 663: Question: are we sure want to remove this property?

The original motivation was to:

- check that we are limiting the production of change bundles where large quantities of one asset are bundled with tiny quantities of some other asset, and
- to the greatest extent possible generate bundles where small quantities are bundled with other small quantities and large quantities with other large quantities.

The original comment for `makeChangeForSurplusAssets` did describe this goal, but alas was not clear enough at describing the motivation, which is my oversight!

jonathanknowles (commented):
Hi @KtorZ 

I think this is almost ready. Thanks for making all the changes. 👍🏻  I'm sorry it took a long time to review, as there were many changes to understand and verify.

I do however have a few comments and questions that I'd like us to clarify (see comments) if possible.

I took the liberty of adding a few commits, which I hope is okay:

- 6705fca42eab994e5918a336ade424c20f50bb9e: adds some missing documentation comments, and clarifies some of the existing comments.
- e4327d1d91302f75ce5aa7ff089c37677fa9d9c9: renames `insufficientlyCoveredOutput` to `outputWithInsufficientAda`, as discussed in a previous review comment (https://github.com/input-output-hk/cardano-wallet/pull/2451#discussion_r560856424).
- b3dbfffeb46eb91a3fddaffb6524fe1fec96e35e: changes all counterexamples to use the `Flat` representation, as discussed in a previous review comment. (The tests previously had a mixture of `Nested`, `Flat`, and the output of `show`.)

Thanks again for your perseverance!

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
KtorZ: It is doing a little extra work indeed. But partitioning the list upfront isn't necessarily less work. Here, we got the change for known and unknown assets in two list traversals. Though, when partitioning before, you end up with an extra first traversals for creating the partition (although, the next two are done on smaller lists)

In the end, I privileged simplicity over optimizations, because I find this last version easier to read. It took me some time to understand what exactly the previous version was doing, but more importantly, why it was still correct. It required more reasoning. Here, it's quite straightforward IMO.

Note: I've built the source with profiling yesterday evening and intended to run some profiling analysis on this entire code today. I've tried running `performSelection` on very large UTxO (> 50k) and it didn't went well. If this appears to be one of the bottleneck, I am happy to rework the implementation in favor of optimizations.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
KtorZ: I found `payment` equally confusing though, for the payment is actually the whole transaction. Perhaps, one better way to describe those are:

- userSpecifiedAssets
- changeAssets

What do you think?

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
KtorZ: I indeed failed to see the motivation for that property in the first place. Thanks for clearing that up. Sounds like a reasonable motivation, so I can see how to re-introduce that :+1:

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
jonathanknowles: I think it's important that the two terms are obvious antonyms of each other.

So this would definitely work:

- `userSpecifiedAssets`
- `nonUserSpecifiedAssets`

But perhaps this is even shorter and clearer:

- `outputAssets`
- `nonOutputAssets`

The last option makes it very obvious that the `outputAssets` were those assets actually found in the outputs. (And it would make it consistent with the comments, as those refer to "assets found in the outputs" and "assets not found in the original outputs")

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
jonathanknowles: Great! I think it should be fairly easy to re-introduce. (I suspect that the function still satisfies the property, as this behaviour is provided by `padCoalesce`.)

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
jonathanknowles: > I've tried running performSelection on very large UTxO (> 50k) and it didn't went well.

Just out of curiosity, how were you generating your very large UTxO sets? Were they taken from an existing wallet DB, or did you use a QC generator?

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
KtorZ: I don't like `outputAssets` for the same reasons I don't like `paymentAssets`, change outputs are outputs too :sweat_smile: ... So I think this settles it on `userSpecifiedAssets` / `nonUserSpecifiedAssets` 

I'll make the change :+1:

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
```diff
@@ -416,6 +420,9 @@ data InvariantStatus
     | InvariantAssetsInconsistent
       -- ^ Indicates that the 'index' and the cached 'balance' value disagree
       --   about which assets are included.
+    | InvariantIndexedCoinsAreIndeedCoins
+      -- ^ Indicates that all inputs indexed as 'coins' are indeed ada-only
+      -- outputs.
```
jonathanknowles: This type represents all the possible invariant **failures**, rather than successes.

So at least the comment is incorrect, as it actually indicates the reverse of what it says.

I reckon it would be better to rename this constructor (and change the comment) to avoid confusion. Here's a suggestion:

```suggestion
    | InvariantCoinIndexContainsNonCoins
      -- ^ Indicates that at least one entry in 'coins' is not ada-only.
```

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
```diff
@@ -473,8 +482,11 @@ indexIsComplete :: UTxOIndex -> Bool
 indexIsComplete u = F.all entryIndexed $ Map.toList $ utxo u
   where
     entryIndexed :: (TxIn, TxOut) -> Bool
-    entryIndexed (i, o) =
-        F.all (indexed i) (TokenBundle.getAssets $ view #tokens o)
+    entryIndexed (i, o)
+        | txOutIsCoin o =
+            i `Set.member` coins u
+        | otherwise =
+            F.all (indexed i) (TokenBundle.getAssets $ view #tokens o)
```
jonathanknowles, line 488: 👍🏻

jonathanknowles (commented):
Hi @KtorZ 

This is just a review of the optimization to `UTxOIndex`.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
jonathanknowles: I do prefer your implementation to [my prototype implementation](https://github.com/input-output-hk/cardano-wallet/compare/jonathanknowles/optimize-ada-only-selection?expand=1), as yours places the index completeness check within one function, which is more concise (and will probably run faster). 👍🏻

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
KtorZ: I used the QC generator. So arguably, the generator was already taking a lot of the time.

jonathanknowles: > I used the QC generator. So arguably, the generator was already taking a lot of the time.

That would also be my assumption.

I guess one way to check is to always start from the same very large UTxO set, and run multiple selections on that single set. Each selection should still be different, as the selection is still random (provided that we start with a different seed each time).

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
```diff
@@ -416,6 +420,9 @@ data InvariantStatus
     | InvariantAssetsInconsistent
       -- ^ Indicates that the 'index' and the cached 'balance' value disagree
       --   about which assets are included.
+    | InvariantIndexedCoinsAreNotCoins
+      -- ^ Indicates that all inputs indexed as 'coins' are indeed ada-only
+      -- outputs.
```
jonathanknowles: Hi @KtorZ

This comment is still wrong, I think, because it indicates the opposite of what it says.

Suggested wording (though feel free to pick a different wording):

```suggestion
    | InvariantIndexedCoinsAreNotCoins
      -- ^ Indicates that at least one member of 'coins' is not ada-only.
```

jonathanknowles (commented):
Hi @KtorZ

Just looked over again.

I think there's still an issue with the comment on `InvariantIndexedCoinsAreNotCoins`. But otherwise looking good.

Looking forward to seeing the final property addition!

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
KtorZ: On second thought, I fail a bit to see how the current implementation helps to satisfy the motivation you just highlighted and also, whether this is even something we do want in practice. The current implementation does the following:

```hs
    change :: NonEmpty TokenBundle
    change
        = TokenBundle.fromCoin <$> changeForCoin
        & NE.zipWith (<>) (TokenBundle.fromTokenMap <$> changeForPaymentAssets)
        -- Here we sort in ascending order of coin value so that surplus assets
        -- and coin values are in increasing order in the resulting bundles:
        & NE.sortWith TokenBundle.getCoin
        & NE.zipWith (<>) (TokenBundle.fromTokenMap <$> changeForSurplusAssets)
```

So, it starts off by assigning all surplus coins, and zip that with the change of user-defined assets. After this step, we have in principle change outputs which are (ideally) similar to user-specified outputs; or at least that's what the algorithm is trying to achieve. 

Then, we have a set of assets that aren't present in user-specified and have to distribute them somehow. So the most reasonable thing to do is to leave them untouched as much as possible and merge them with other token bundles. The only real area of maneuver we have is with which other change outputs to merge them? This new rework doesn't make any decision whereas the current did first sort existing tokens by coin quantity so that largest bundles will be merged with largest coin values. I don't quite see why we would generally want that. In practice, UTxOs with the largest coin values will likely be ada-only UTxOs, and it's probably best to keep them as ada-only. 

I do agree with the following motivation though:

> to the greatest extent possible generate bundles where small quantities are bundled with other small quantities and large quantities with other large quantities. 

Besides, with the new implementation, change coins are assigned only after the other changes have been calculated (for we need them to have been created _before_ so that the minimum coin value can be calculated for each). This makes it impossible to re-order the list before change coins are assigned (otherwise, we may deviate from the requested output distribution). 

So instead, since we know that: 

a) `padCoalesce` returns a list ordered in ascending order
b) We can re-order requested outputs for change calculation upfront, so that "bigger" outputs are first 

We can do a best effort at merging the largest non user-defined change assets with the largest user-defined change assets. Another issue is that we only have a partial order on token maps whereas we would need a total order here. It is maybe possible to define one for that purpose, such that:

- We consider the partial order for elements which can be partially ordered. 
- For others, we may consider their maximum quantity. 

e.g.

```
>>> sort [ [ (A,3) ], [ (A,1), (B,2) ], [ (C,4), (B,2) ], [ (B,1) ] ]
[ [ (B,1) ], [ (A,1), (B,2) ], [ (A,3) ], [ (C,4), (B,2) ]
```

Since the algorithm tends to lead to the elimination of small quantities, this should over time help generating UTxOs where large quantities are bundled together and small quantities together.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
KtorZ: Fixed :+1:

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
jonathanknowles: Hi @KtorZ 

I do agree with your reasoning here.

With regard to ordering maps according to their greatest token quantity: I think this is a very reasonable approach.

I was curious, however, as to whether the `totalOrder` function defined in f58841bfdf036f164fb5d82f643a3d48f5a9e9c6 (the one that gives priority to `PartialOrd`) actually produces a well-defined order. So just to make sure, I extracted the `totalOrder` function out, and tested it with `testLaws`. This revealed a few subtle issues with the ordering.

I did try to fix these issues, but was unable to produce an ordering that gives priority to `PartialOrd` and yet satisfies the `Ord` laws.

However, I do think that we could adopt a simpler approach here:

It could be enough to simply order token maps by their greatest quantity. In the case where there is a tie-break, we can fall back to lexicographic ordering, which provides fast termination and guarantees to produce a total order. (This is similar to the technique we use for ordering `TxOut`.)

To that end, I pushed a commit 185a712b41c91b2203685044b2f9d246efd2fcb4, which fixes the ordering, and also adds a property test to check that the ordering is well-defined, satisfying the `Ord` laws.

I hope this is okay. But if you still feel there is a better ordering, then perhaps we can adjust the `Ord` instance, or go for something else entirely. Since we now have a property test, it should be easy to check that whatever ordering we choose is valid.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
jonathanknowles: > So I think this settles it on userSpecifiedAssets / nonUserSpecifiedAssets

I agree that the combination of "user-specified" and "non-user-specified" definitely works here. I think it will also work well in any public facing documentation we write about the algorithms.

I also think it is much clearer than "user-defined" (as this could be interpreted as meaning that the user had defined their own assets, which is somewhat ambiguous).

I added some further clarifications to the documentation and comments in 5dfc0d17454059271a7e9ff8bd37b8acee7a3256.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -756,8 +756,29 @@ makeChange
...
+    --     - Otherwise, we order them using the maximum quantity of one of their
+    --       assets.
+    --
+    -- Over time, this should tend to bundle small (resp. large) bundles together.
     outputTokens :: NonEmpty TokenMap
-    outputTokens = view #tokens <$> outputBundles
+    outputTokens = NE.sortBy totalOrder (view #tokens <$> outputBundles)
+      where
+        totalOrder m1 m2
+            | m1 `leq` m2 = LT
```
jonathanknowles: I think this is not _quite_ right, as `leq` will also return `True` if two maps are equal.

To fix it, we'd need to introduce an additional guard:
```hs
             | m1 ==    m2 = EQ
             | m1 `leq` m2 = LT
```

However, this is unfortunately still not enough to generate a well-defined total ordering, as even the modified function fails when tested with `testLaws`.

However, we can make a total ordering where we first compare on the greatest token quantity, and then fall back to lexicographic ordering in the event of a tie-break. I made a patch that tests this here: 185a712b41c91b2203685044b2f9d246efd2fcb4

lib/core/test/unit/Cardano/Wallet/Primitive/Types/TokenMapSpec.hs
```diff
@@ -422,6 +424,13 @@ prop_adjustQuantity_hasQuantity b asset =
         | quantity > TokenQuantity.zero = TokenQuantity.pred quantity
         | otherwise = quantity
 
+prop_maximumQuantity_all
+    :: TokenMap -> Property
+prop_maximumQuantity_all b =
+    property $ all (<= maxQ) (snd <$> TokenMap.toFlatList b)
+  where
+    maxQ = TokenMap.maximumQuantity b
+
```
jonathanknowles, line 433: 👍🏻

jonathanknowles (approved):
LGTM.

@KtorZ thanks for all your work on this, I think the result is great.

Issue #2439, opened 2021-01-12
TRANS_LIST_02,03x - PatternMatchFail or ErrorCall


rvl:
# Context
 
This is a flaky test failure where the wallet response was not expected.

# Test Case

- API Specifications, SHELLEY_TRANSACTIONS, TRANS_LIST_02,03x - Can limit/order results with start, end and order
- API Specifications, SHELLEY_WALLETS, WALLETS_LIST_01 - Wallets are listed from oldest to newest

# Failure / Counter-example

Build failed:
  * [buildkite/cardano-wallet](https://buildkite.com/input-output-hk/cardano-wallet/builds/12828)

```
src/Test/Integration/Scenario/API/Shelley/Transactions.hs:1749:13:
--
  | 1) API Specifications, SHELLEY_TRANSACTIONS, TRANS_LIST_02,03x - Can limit/order results with start, end and order
  | uncaught exception: PatternMatchFail
  | src/Test/Integration/Scenario/API/Shelley/Transactions.hs:1749:13-74: Non-exhaustive patterns in [Just t2,
  | Just t1]
  |  
  | To rerun use: --match "/API Specifications/SHELLEY_TRANSACTIONS/TRANS_LIST_02,03x - Can limit/order results with start, end and order/"
  |  
  | Randomized with seed 260583187
```

Build failed:
  * [ci/hydra-build:required](https://hydra.iohk.io/build/5280905)

```
 src/Test/Integration/Framework/DSL.hs:1471:16: 
  1) API Specifications, SHELLEY_WALLETS, WALLETS_LIST_01 - Wallets are listed from oldest to newest
       uncaught exception: ErrorCall
       getFromResponse failed to get item
       CallStack (from HasCallStack):
         error, called at src/Test/Integration/Framework/DSL.hs:1471:16 in cardano-wallet-core-integration-2020.12.21-AAYZFneZURcKNEp0gMVmAX:Test.Integration.Framework.DSL

  To rerun use: --match "/API Specifications/SHELLEY_WALLETS/WALLETS_LIST_01 - Wallets are listed from oldest to newest/"
```

---

# Resolution 

---

# QA

Anviking:
Closing as stale

Issue #2430, opened 2021-01-07
Flaky TRANS_LIST_RANGE_03 / TRANS_LIST_RANGE_01


Anviking:
# Context
 

https://github.com/input-output-hk/cardano-wallet/pull/2370#issuecomment-753741824

# Test Case

https://github.com/input-output-hk/cardano-wallet/blob/c8ac57faea0e9d9a9bee21a498ee678c997f176e/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs#L2088-L2096

# Failure / Counter-example

```
 src/Test/Integration/Scenario/CLI/Shelley/Transactions.hs:641:5:
      2) CLI Specifications, SHELLEY_CLI_TRANSACTIONS, TRANS_LIST_RANGE_03 - Transaction at time t is NOT selected by range (..., t - 𝛿t]
           uncaught exception: StringException
           Control.Exception.Safe.throwString called with:
 
           Expected at least one transaction with a time.
           Called from:
             throwString (src/Test/Integration/Framework/DSL.hs:691:14 in cardano-wallet-core-integration-2020.12.21-2gb1H39TPSm79qdhbcZCTW:Test.Integration.Framework.DSL)
```

---

# Resolution 

---

# QA

Anviking:
Should have been fixed.

Issue #2426, opened 2021-01-07
Flaky: SHELLEY_TRANSACTIONS/TRANS_LIST_RANGE_xx


KtorZ:
# Context
 
 

https://hydra.iohk.io/build/5278704

# Test Case

`SHELLEY_TRANSACTIONS/TRANS_LIST_RANGE_02 `

# Failure / Counter-example

```
  1) API Specifications, SHELLEY_TRANSACTIONS, TRANS_LIST_RANGE_02 - Transaction at time t is NOT selected by range (t + 𝛿t, ...)
         uncaught exception: StringException
         Control.Exception.Safe.throwString called with:
       
         Expected at least one transaction with a time. 
         Called from:
         throwString (src/Test/Integration/Framework/DSL.hs:691:14 in cardano-wallet-core-integration-2020.12.21-cOKE6yQQBEAvS3rnChmPr:Test.Integration.Framework.DSL)

  To rerun use: --match "/API Specifications/SHELLEY_TRANSACTIONS/TRANS_LIST_RANGE_02 - Transaction at time t is NOT selected by range (t + \120575t, ...)/"
  ```

Anviking:
Should have been fixed by #2455

PR #2452, opened 2021-01-15
Fix some integration tests

piotr-iohk:
# Issue Number

#2439, #2430, #2426

# Overview

- ce4f29d0177aa40e51c9ceb0c7467d03c58d2545
  Make WALLETS_LIST_01 fail properly in case of unexpected response
  
- 61b5ff5309884856db94306dc3d0508b455529d2
  use request and check for status in DSL methods to make them fail properly in case of unexpected response (hopefully fixes TRANS_LIST_RANGE_)
  
- 852a446b524b3c541eaecf1f18810e5fa5d5e29f
  Make TRANS_LIST_02,03x more reliable (hopefully)
  
- e43b9302ebd999017bcfd70dbb7ba0cf540ea9cc
  update comment for eventually

# Comments

### DIFF
--- a/lib/core-integration/src/Test/Integration/Framework/DSL.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/DSL.hs
@@ -702,7 +702,9 @@ waitAllTxsInLedger
     -> m ()
 waitAllTxsInLedger ctx w = eventually "waitAllTxsInLedger: all txs in ledger" $ do
     let ep = Link.listTransactions @'Shelley w
-    (_, txs) <- unsafeRequest @[ApiTransaction n] ctx ep Empty
+    r <- request @[ApiTransaction n] ctx ep Default Empty
+    expectResponseCode HTTP.status200 r
+    let txs = getFromResponse id r
     view (#status . #getApiT) <$> txs `shouldSatisfy` all (== InLedger)
 
 waitForNextEpoch
@@ -1167,8 +1169,10 @@ fixtureWalletWithMnemonics ctx = snd <$> allocate create (free . fst)
                 "mnemonic_sentence": #{mnemonics},
                 "passphrase": #{fixturePassphrase}
                 } |]
-        (_, w) <- unsafeRequest @ApiWallet ctx
-            (Link.postWallet @'Shelley) payload
+        r <- request @ApiWallet ctx
+            (Link.postWallet @'Shelley) Default payload
+        expectResponseCode HTTP.status201 r
+        let w = getFromResponse id r
         race (threadDelay sixtySeconds) (checkBalance w) >>= \case
             Left _ -> expectationFailure'
                 "fixtureWallet: waited too long for initial transaction"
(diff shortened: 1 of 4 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -1739,13 +1739,27 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
     -- +---+----------+----------+------------+--------------+
     it "TRANS_LIST_02,03x - Can limit/order results with start, end and order"
         $ \ctx -> runResourceT $ do
-        let a1 = Quantity $ sum $ replicate 10 minUTxOValue
-        let a2 = Quantity $ sum $ replicate 10 (2 * minUTxOValue)
-        w <- fixtureWalletWith @n ctx $ mconcat
-                [ replicate 10 minUTxOValue
-                , replicate 10 (2 * minUTxOValue)
-                ]
-        txs <- listAllTransactions @n ctx w
```
Anviking, line 1748: Was the problem that the faucet txs would be rolled back? Do you think a `threadDelay` would have helped as well as your solution?

Anviking (approved):

lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -832,14 +834,14 @@ expectationFailure' msg = do
...
--- This __90s__ is mostly justified by the parameters in the shelley
+-- This __120s__ is mostly justified by the parameters in the shelley
 -- genesis. The longest action we have two wait for are about 2 epochs,
 -- which corresponds to 80s with the current parameters. Using something
 -- much longer than that isn't really useful (in particular, this doesn't
 -- depend on the host machine running the test, because the protocol moves
 -- forward at the same speed regardless...)
 eventually :: MonadIO m => String -> IO a -> m a
-eventually = eventuallyUsingDelay (500 * ms) 90
+eventually = eventuallyUsingDelay (500 * ms) 120
```
Anviking: With the current parameters 120s is 12 epochs, so a bit long… You found it necessary?

piotr-iohk: Is it? I thought it's ~40s based on the comment :point_up_2:

Anviking: https://github.com/input-output-hk/cardano-wallet/blob/922319c7d6e99abaf34e5dc36c7b3cbf8a5876e9/lib/shelley/test/data/cardano-node-shelley/shelley-genesis.yaml#L50-L53

Ah, comment got out of date with #2391. Was 40s before that.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
piotr-iohk: > Was the problem that the faucet txs would be rolled back?

I'm not sure, but I guess `threadDelay` would not be very reliable in such case?
TBH, I find this:

```
        w <- fixtureWalletWith @n ctx $ mconcat
                [ replicate 10 minUTxOValue
                , replicate 10 (2 * minUTxOValue)
                ]
```
a bit cryptic as opposed to this:

```
        _ <- postTx ctx linkTx w minUTxOValue
        verifyWalletBalance ctx w (Quantity minUTxOValue)

        _ <- postTx ctx linkTx w (2 * minUTxOValue)
        verifyWalletBalance ctx w (Quantity (3 * minUTxOValue))
```
which will additionally fail with an explicit/clear error if something is not right.

Anviking: 👌 👍

piotr-iohk:
I guess most of the failures I was targeting here are fixed by https://github.com/input-output-hk/cardano-wallet/pull/2455, but still there are some additional assertions and changes that make tests more clear.

PR #2459, opened 2021-01-21
Update cardano-node and libraries to 1.25.0

rvl:
### Issue Number

ADP-672

### Overview

- Updates cardano-node to [1.25.0 release](https://github.com/input-output-hk/cardano-node/releases/tag/1.25.0)
- Fixes some type errors resulting from update.
- Bumps hackage.nix revision, but leaves haskell.nix revision the same, due to a nix-shell problem.


lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -490,8 +490,9 @@ localStateQuery queue =
     clientStIdle
         :: m (LSQ.ClientStIdle block (Point block) (Query block) m Void)
     clientStIdle = awaitNextCmd <&> \case
-        CmdQueryLocalState pt query respond ->
-            LSQ.SendMsgAcquire pt (clientStAcquiring query respond)
+        CmdQueryLocalState _pt query respond ->
+            -- fixme: ADP-672 not sure why it's Maybe Point now
```
Anviking: I think `Nothing` now means "use the current tip"

Anviking: Reverted with rationale -> b2388bd

rvl: > Allowing the node to decide what the current tip is, would solve ADP-647, but it could also cause race conditions from queries potentially being run on different points.
> 
> We could still leverage Nothing in #2449.

OK good rationale.

Anviking (approved):

KtorZ (approved):

PR #2461, opened 2021-01-22
Add function `computeMinimumAdaQuantity`.

jonathanknowles:
# Issue Number

ADP-506

# Overview

This PR adds function `computeMinimumAdaQuantity`, which uses the **ledger specification** to calculate the minimum required ada quantity for a token bundle.

It also adds property tests to verify that `computeMinimumAdaQuantity` is:
- constant w.r.t. the ada quantity
- constant w.r.t. individual non-ada asset quantities


lib/shelley/src/Cardano/Wallet/Shelley/Compatibility/Ledger.hs
```diff
@@ -0,0 +1,173 @@
...
+-- Conversions between wallet types and ledger types
+--------------------------------------------------------------------------------
+
+class HasLedgerType wallet ledger | wallet -> ledger where
+    -- | Converts a value from a ledger type to the equivalent wallet type.
+    fromLedger
+        :: HasCallStack => ledger -> wallet
+    -- | Converts a value from a wallet type to the equivalent ledger type.
+    toLedger
+        :: HasCallStack => wallet -> ledger
```
KtorZ, line 120: This deviates quite a bit from the approach already taken in `.../Shelley/Compatibility.hs`. May we stick with the same approach and have these conversions in `.../Shelley/Compatibility.hs` following the existing naming conventions? 

We don't need a type-class for this, plain functions are more readable and easily passed around. There are already far enough constraints from integrating with cardano-api and the cardano-ledger-specs :+1:

jonathanknowles: Hi Matthias

Thanks for your feedback. 👍🏻 

> plain functions are more readable and easily passed around.

I've added https://github.com/input-output-hk/cardano-wallet/pull/2461/commits/0d96de1ebe8a39a4f1117a9b7aa7796b858b429e, which exports plain functions for each of the conversions, as requested:

```hs
      -- * Conversions from wallet types to ledger specification types
    , toLedgerCoin
    , toLedgerTokenBundle
    , toLedgerTokenPolicyId
    , toLedgerTokenName
    , toLedgerTokenQuantity

      -- * Conversions from ledger specification types to wallet types
    , toWalletCoin
    , toWalletTokenBundle
    , toWalletTokenPolicyId
    , toWalletTokenName
    , toWalletTokenQuantity
```

> We don't need a type-class for this

There's actually no need for callers of this module to use any type classes. They can just use the plain functions, which are now exported (see above).

My intention here was to encode the fact that certain pairs of types support round-trip conversion, which in turn allows concise round-trip testing:

```hs
describe "Roundtrip conversions between wallet and ledger specification types" $ do
    ledgerRoundtrip $ Proxy @Coin
    ledgerRoundtrip $ Proxy @TokenBundle
    ledgerRoundtrip $ Proxy @TokenName
    ledgerRoundtrip $ Proxy @TokenPolicyId
    ledgerRoundtrip $ Proxy @TokenQuantity
```

The expected properties can now be documented in _just one place_, rather than duplicating these properties in the comments of every single conversion function:
```hs
-- | A pairing between a wallet type and its equivalent ledger type.
--
-- Instances of this class should satisfy the following laws:
--
-- >>> toLedger . toWallet == id
-- >>> toWallet . toLedger == id
--
class Convert wallet ledger | wallet -> ledger
```

> This deviates quite a bit from the approach already taken in .../Shelley/Compatibility.hs.
> May we stick with the same approach and have these conversions in .../Shelley/Compatibility.hs following the existing naming conventions?

My reason for putting this in a separate module is:

- **`Compatibility`** seems to be mainly concerned with the **node**, and the types and functions it exports. (Though there might be some other stuff in there too.)
- **`Compatibility.Ledger`** is concerned with the **ledger specification**, which has completely different types and encodings for certain things (such as token bundles).

These two layers seem distinct enough to warrant having separate modules, wouldn't you say?

In general, assuming concepts can be cleanly separated, I'm not a big fan of very large modules that mix different concepts together. In addition, the `Compatibility` module is already huge: it's nearly 1500 lines!

KtorZ: Compatibility currently concerns cardano-api, ouroboros-network and cardano-ledger-specs. Basically, the cardano's ecosystem gravitating around cardano-node. I don't see any reason to put these in a separate module or to treat them differently than the rest. 

> I'm not a big fan of very large modules that mix different concepts together. In addition, the Compatibility module is already huge: it's nearly 1500 lines!

I know that. But if we want to split this module, we better do it consistently and with a more holistic approach; and not have _some_ functions in `Compatibility` and some others in `Compatibility.Ledger`. Now that just adds to the overall complexity. It also save us hardly any compilation time because both modules are needed around the same time to compile the final executable, and both comes already at the end of the compilation stack.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility/Ledger.hs
```diff
@@ -0,0 +1,243 @@
...
+--------------------------------------------------------------------------------
+
+-- | Uses the ledger specification to compute the minimum required ada quantity
+--   for a token bundle.
+--
+computeMinimumAdaQuantity
+    :: Coin
+    -- ^ The absolute minimum ada quantity specified by the protocol.
+    -> TokenBundle
+    -- ^ The token bundle to evaluate.
```
KtorZ: The function we've been using in the rest of the code works in principle from a `TokenMap` (since we do not yet know the coin value). I think we can adjust the rest of the codebase to work off `TokenBundle`, but if doable, I'd rather fix it here.

lib/shelley/test/unit/Cardano/Wallet/Shelley/Compatibility/LedgerSpec.hs
```diff
@@ -0,0 +1,351 @@
...
+
+prop_computeMinimumAdaQuantity_agnosticToAdaQuantity
+    :: Blind TokenBundle
+    -> ProtocolMinimum Coin
+    -> Property
+prop_computeMinimumAdaQuantity_agnosticToAdaQuantity
+    (Blind bundle) (ProtocolMinimum protocolMinimum) =
+        counterexample counterexampleText $ conjoin
+            [ compute bundle === compute bundleWithCoinMinimized
+            , compute bundle === compute bundleWithCoinMaximized
```
KtorZ, line 123: :+1:

lib/shelley/test/unit/Cardano/Wallet/Shelley/Compatibility/LedgerSpec.hs
```diff
@@ -0,0 +1,351 @@
...
+            , if assetCount == 0
+                then bundleMinimized === bundleMaximized
+                else bundleMinimized =/= bundleMaximized
+            ]
+  where
+    assetCount = Set.size $ TokenBundle.getAssets bundle
+    assetCountMinimized = Set.size $ TokenBundle.getAssets bundleMinimized
+    assetCountMaximized = Set.size $ TokenBundle.getAssets bundleMaximized
+    bundleMinimized = bundle `setAllQuantitiesTo` txOutMinTokenQuantity
+    bundleMaximized = bundle `setAllQuantitiesTo` txOutMaxTokenQuantity
```
KtorZ, line 166: :+1:

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility/Ledger.hs
jonathanknowles: Hi @KtorZ . I'll push a patch to this effect very shortly. 👍🏻

jonathanknowles: @KtorZ I've just pushed 39dfd7c1c6309a57010add65fa127630c656c4b1, which changes the type of the public function to:

```hs
-- | Uses the ledger specification to compute the minimum required ada quantity
--   for a token map.
--
computeMinimumAdaQuantity
    :: Coin
    -- ^ The absolute minimum ada quantity specified by the protocol.
    -> TokenMap
    -- ^ The token map to evaluate.
    -> Coin
    -- ^ The minimum ada quantity for the given token map.
```

The old function is kept around as an internal function. We do need to keep that around however, as:

- it's a simple wrapper around the ledger specification's `scaledMinDeposit` function, which expects to be supplied with an ada quantity
- we need to test that `scaledMinDeposit` remains constant w.r.t. the ada quantity.

KtorZ (approved):
:heavy_check_mark: , modulo the unnecessary module addition which I'd have preferred to avoid to keep the codebase more consistent.

PR #2462, opened 2021-01-22
RoundRobin multi-asset coin-selection integration in cardano-wallet

KtorZ:
# Issue Number

ADP-506 / ADP-605

# Overview

- 395b9542449c832178c51a1179440c6001510983
  :round_pushpin: **remove migration selection handlers until rework**
    Most of the code doesn't apply anymore because of the way we've changed how coin-selection works. So I've simply thrown everything away. It'll be possible to look at git later when re-implementing this to get some inspiration. It is also very likely that we may want to do things slightly differently, the problem and context being now different.

- bc3d0a600832499cd40347bd5a82fbaae8dbb987
  :round_pushpin: **rework 'TransactionLayer' to play well with RoundRobin MA selection algs.**
    The main change compared to before is the unification of all 'mkStdTx', 'mkDelegationJoinTx' and 'mkDelegationQuitTx' under one single interface 'mkTransaction'. We know for sure that this is possible because that's exactly how things are implemented behind the scene in the cardano-wallet package. This is a quite disruptive change, but should help really help simplifying the coin selection code which has grown large and complex over the past months with many pitfalls and logic duplication here and there.

  Another major change is the introduction of a 'TransactionCtx' type which should help unifying the caller interface in a more elegant way, instead of having to hand-craft every single argument by hand depending on the function to call. The main idea behind this 'TransactionCtx' is that it contains details about the transaction that are known prior to constructing the transaction. Typically, they'd come from the surrounding context (e.g. the current time) or, be user-provided.

- 84465c5c601544f55a60e240def3b3e722b5ed26
  :round_pushpin: **parameterize 'SelectionResult' over the change's inner type**
    The reason for this is to be able to re-use the 'SelectionResult' at various stages in the wallet layer, before and after having assigned change address to each change output.
  It opens room for writing nice functions such as: 'assignChangeAddress :: SelectionResult TokenBundle -> s -> (s, SelectionResult TxOut)' which, from its type signature informs
  nicely about what it is doing.

- fc7260badb5fc30bb005c0ccf16a475fde7a3b38
  :round_pushpin: **add 'extraCoinSource' to 'SelectionResult'**
    So that we can fully construct transaction metadata from a 'SelectionResult' and need not to re-do any extra computation or pass extra arguments about withdrawals and reclaims.

- 0a99a26ed61e76aaa6b85246d260c4b8f09f354a
  :round_pushpin: **add outputsCovered to 'performSelection', to make it possible to fully reconstruct a tx from a result**
    Adding it as a list to cope with delegation selections which have no outputs.

- 09b9b14ca2e3a63a98982319a81cf8419f521cdf
  :round_pushpin: **decouple deposits from 'estimateFee'**
    Fees and deposits are closely related but that aren't the same thing. So it's better to keep the fee estimation about fees, and leave deposits out of it.

- 8486593f9f35c52af1975ad3adc4c9945b331785
  :round_pushpin: **integrate RoundRobin Multi-Asset selection and new transaction layer in the wallet layer**
  
- 1a48de27d4c983ae0602e4770cb4c0afe6313ab9
  :round_pushpin: **upgrade the server handlers to use the upgraded multi-asset wallet layer.**
  
- b1a09c472e17210f5090f8cb245d2748a5a0a8b0
  :round_pushpin: **add simple conversion function to '../Types/Coin.hs' from Coin to integrals**
  
- 524d5b5df0fd5e91f752cfe4a2f4f7d74cc995ef
  :round_pushpin: **move 'calcSelectionDelta' to '.../MA/RoundRobin' and make it work for all 'SelectionResult' types**
    We need this in two places, on 'SeletionResult TxOut' and 'SelectionResult TokenBundle'.

- 7e49ff1d49dd6b187c649da992d2d22de7c6b5d2
  :round_pushpin: **Upgrade cardano-wallet package to work with new TransactionLayer**
  
- fda6581a16193ee3a2e83cf2549e90003d6ae6b3
  :round_pushpin: **Adjust error messages assertions in integration tests**
    To match new errors generated from the multi-assets selection.

- 2eb1e7b06d767bf25e362683bac2a449ed914d61
  :round_pushpin: **return and catch required cost with 'UnableToConstructChangeError'**
    So that we can still return a value when users want to estimate fees but there's not enough money to run a full selection. Though, what we really need is a complete rework of the how transaction are created which does not rely on fee estimation.

- 8fc5683a25beea31a7729465fa8816f0b72580b0
  :round_pushpin: **run 'coinSelector' last in the multi-asset selection.**
  
- aee4f87c299bfa5a8f7418d56b03d6e177da5890
  :round_pushpin: **implement pure version of 'selectionToUnsignedTx' for external coin selections**
    Did rework a bit how 'selectAssets' also work and allowed callers to pass an extra transformation function on the result as argument, using the state with which the selection was calculated. This allows for callers to run some extra computations on the exact same state. We have two use-cases for this particular argument:

   - Fee estimation functions, which transform the selection result into a 'Coin'
   - So-called external selections ran for read-only wallets, which transform the selection into an 'UnsignedTx' with fully qualified (i.e. with derivation paths) inputs and change.

- 61c48723812acd3b415bc22372b38a0d7661cd4f
  :round_pushpin: **Update types in `RoundRobinSpec` test suite.**
  This change updates various types within the `RoundRobinSpec` test suite in
light of recent changes to the `RoundRobin` module, allowing it to compile.

- 6c595e90f63e1a6fd82f83b7b290b76622b9cf5d
  :round_pushpin: **Test the value of `outputsCovered` returned by `performSelection`.**
  In particular, we wish to ensure the following property holds:

>>> outputsCovered == outputsToCover

- 2650cfff2d87dcfce912e393f8ba2ccbad4c35ba
  :round_pushpin: **remove now obsolete coin-selection modules.**
  
- b34a575a344d92fb9a7db3dc87fce02cdb72aaa3
  :round_pushpin: **rename 'missingCoins' field to 'shortfall' in the MA selection change error**
  
- 5264ba9da52f7af56bd75a5a3b064deca24739f9
  :round_pushpin: **introduce 'defaultTransactionCtx' to avoid repetition of empty contexts in various situation**
  
- e6319b176395ce832be2dbe5b38d10f4c90f7b99
  :round_pushpin: **update benchmarks & shelley transaction specs to work with new transaction layer.**
  
- 6080b021fd311e8744f141cb4ad36827bc99f863
  :round_pushpin: **do not count external withdrawals as 'ours' when constructing tx metadata**
  
- 64629400dd3415f267a453ebf6e2a51c7e7d61db
  :round_pushpin: **make sure to count user-specified outputs that are ours when constructing tx meta**
  
- 3ff070ffcdeb4bd98818e40a4bfb101b041fef03
  :round_pushpin: **finish implementation of 'calcMinimumCoinValue' to also account for multi-assets**
  
- ebde06041a3a0864a476c80484a1661b97a8231b
  :round_pushpin: **fix stake pool tests and transaction metadata reporting**
  
- 581309db1149e5172c434e9eb931744dd782ba30
  :round_pushpin: **remove now obsolete 'TransactionSpecShared' module.**
  
- a606ca9b03c20493626d4e7148c1d6ae3af8bb52
  :round_pushpin: **make 'estimateMaxNumberOfInputs' aware of multi-assets outputs.**
    And adjust unit tests accordingly.

- 9b9bd2f989f1c26f856b86e3ce3f63cefad71c05
  :round_pushpin: **use current mainnet value as tx max size in unit test**
    So that we get some order good of magnitudes when it comes to what can be expected for Mainnet.

- 971cd17a61a7adf087be680743eca445aaa18ffd
  :round_pushpin: **remove integration coin selection scenarios that are no-longer applicable.**
    Indeed, the wallet now _always_ create a change output. The only way
  to empty a wallet is to use the migration / swapping endpoints.

- 805e95210102d036f6f60e8697565fc878feb134
  :round_pushpin: **temporarily disable test integration migration scenarios**
    Until the endpoints handlers are re-implemented.

# Comments

Known issues / remaining work so far:

- [x] 1. Transactions with withdrawals from an external wallet are wrongly labeled as "Incoming". 

- [x] 2. `calcMinUTxOValue` needs to be properly implemented (see PR #2461)

- [x] 3. Some picky tests regarding stake pools which work from very precise fee amounts needs to be updated because expected fees are no longer matching. 

- [x] 4. `estimateMaxNumberOfInputs` only consider ada-only assets, resulting in estimation that can be way above the real one. Needs to account for the actual output shape. 

- [x] 5. The equivalent of `amount` for multi-assets on transaction is currently always blank, instead of being assigned with the delta between inputs and outputs. 
 
- [ ] 6. `assets` are missing from the external selections API, although the underlying code is already capable of running multi-asset external selections

- [ ] 7. `assets` are missing from the Byron API endpoints, whereas it is actually completely possible to have a Byron wallet with native assets.

- [ ] 8. Wallet migrations / swapping has been disabled


lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -510,8 +587,13 @@ runSelection limit mExtraCoinSource available minimumBalance =
         , leftover = available
         }
 
+    -- NOTE: We run the 'coinSelector' last, because we know that there are
+    -- necessarily coins in all inputs. Therefore, after having ran the other
+    -- selectors, we may already have covered for coins and need not to select
+    -- extra inputs.
```

Suggestion:

```suggestion
    -- NOTE: We run the 'coinSelector' last, because we know that every input
    -- necessarily has a non-zero ada amount. By running the other selectors
    -- first, we increase the probability that the coin selector will be able
    -- to terminate without needing to select an additional coin.
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -690,8 +772,7 @@ makeChange
         -- ^ Token bundles of original outputs.
     -> Either UnableToConstructChangeError (NonEmpty TokenBundle)
         -- ^ Generated change bundles.
-makeChange
-    minCoinValueFor requiredCost mExtraCoinSource inputBundles outputBundles
+makeChange minCoinValueFor requiredCost mExtraCoinSource inputBundles outputBundles
```
jonathanknowles, line 776: This line seems like an unnecessary or accidental change?

(With the change, it goes over the line limit, whereas before it was fine, I think.)

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -175,19 +191,74 @@ data SelectionLimit
...
+    :: (NonEmpty change -> Builder)
+    -> SelectionResult change
+    -> Builder
+buildSelectionResult changeF s@SelectionResult{inputsSelected,extraCoinSource} =
+    mconcat
+        [ nameF "inputs selected"  (inputsF inputsSelected)
+        , nameF "extra coin input" (build extraCoinSource)
+        , nameF "outputs covered"  (build $ outputsCovered s)
+        , nameF "change generated" (changeF $ changeGenerated s)
+        , nameF "size utxo remaining" (build $ UTxOIndex.size $ utxoRemaining s)
```
jonathanknowles: Suggestion: rather than trying to align these vertically, consider using line breaks.

Justification: the alignment doesn't buy us much here, and might actually lead to larger diffs if in future we need to add another entry that requires a longer name.

```suggestion
        [ nameF "inputs selected"
            (inputsF inputsSelected)
        , nameF "extra coin input"
            (build extraCoinSource)
        , nameF "outputs covered"
            (build $ outputsCovered s)
        , nameF "change generated"
            (changeF $ changeGenerated s)
        , nameF "size utxo remaining"
            (build $ UTxOIndex.size $ utxoRemaining s)
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -175,19 +191,74 @@ data SelectionLimit
 
 -- | The result of performing a successful selection.
 --
-data SelectionResult = SelectionResult
+data SelectionResult change = SelectionResult
```
jonathanknowles, line 194: I can see why a parameterized type is tempting here (as it's convenient), but I can't help wondering if there isn't a more straightforward solution.

If I've understood correctly, `SelectionResult` can be instantiated in two ways, for two different use cases:

1. **`SelectionResult TokenBundle`:**
      the result of running `performSelection`
      
2. **`SelectionResult TxOut`:**
      a balanced selection where change addresses have been assigned

Unless I'm mistaken, the second use case is not required by the `performSelection` function, and is only really used outside of the `RoundRobin` module.

## Proposal

What if we were to instead just make a new record type for the second use case, like this:

```hs
--- | A balanced coin selection where all change addresses have been assigned.
data Selection = Selection
    { inputs  :: ![(TxIn, TxOut)]
    , outputs :: ![TxOut]
    , change  :: ![TxOut]
    }
```

We'd produce a `Selection` by post-processing a `SelectionResult` and combining it with the original criteria. Something like this:
```hs
makeSelection
    :: GenChange s
    => SelectionCriteria
    -> SelectionResult
    -> s Selection
```

Then, unless I've misunderstood, we wouldn't have to include `outputsSelected` in the `SelectionResult` type, which removes the need to write a test that `performSelection` returns the correct value of `outputsSelected` (I think).

And then `mkTransaction` would have the following type:
```hs
        -> ProtocolParameters
            -- Current protocol parameters
        -> TransactionCtx
            -- An additional context about the transaction
        -> Selection
            -- A balanced coin selection where all change addresses have been
            -- assigned.
```

Thoughts?

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -242,9 +313,13 @@ data InsufficientMinCoinValueError = InsufficientMinCoinValueError
...
-newtype UnableToConstructChangeError = UnableToConstructChangeError
-    { missingCoins
-        :: Coin
+data UnableToConstructChangeError = UnableToConstructChangeError
+    { requiredCost
+        :: !Coin
+        -- ^ The minimal required cost needed for the transaction to be
+        -- considered valid. This does not include min Ada values.
+    , missingCoins
+        :: !Coin
```
jonathanknowles: Suggestion: use a singular here, as there's only one coin quantity.

```suggestion
    , shortfall
        :: !Coin
```

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -66,18 +64,27 @@ import Cardano.Wallet.Primitive.AddressDerivation.Icarus
     ( IcarusKey )
 import Cardano.Wallet.Primitive.AddressDerivation.Shelley
     ( ShelleyKey, toRewardAccountRaw )
-import Cardano.Wallet.Primitive.CoinSelection
-    ( CoinSelection (..), feeBalance )
+import Cardano.Wallet.Primitive.CoinSelection.MA.RoundRobin
+    ( SelectionCriteria (..)
+    , SelectionLimit (..)
+    , SelectionResult (changeGenerated, inputsSelected, outputsCovered)
```
jonathanknowles, line 71: Another sign that it might be simpler to make a dedicated `Selection` type, instead of parameterizing `SelectionResult`?

jonathanknowles (commented):
Hi @KtorZ 

This is just a partial review, but wanted to submit it before heading off (as it's late over here!).

 The only reservation I have so far relates to the parameterisation of `SelectionResult`. I suspect there might be a simpler solution (see comments).
 
 Looking forward to reviewing the rest!

Cheers!

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
KtorZ: No, it is done on purpose. There's no way one can argue that:

```hs
makeChange
    minCoinValueFor requiredCost mExtraCoinSource inputBundles outputBundles
    | not (totalOutputValue `leq` totalInputValue) =
```

is more readable than:

```hs
makeChange minCoinValueFor requiredCost mExtraCoinSource inputBundles outputBundles
    | not (totalOutputValue `leq` totalInputValue) =
```

So, despite lines going sometimes slightly over the 80 columns limit, readability and our judgement should have the final say so. In that case, I was confused when re-reading this bit of code and seeing the arguments at the same level as the function body, so I reverted the change.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
KtorZ: I agree is doesn't buy us much, but it also doesn't cost us much (literally, 2 white spaces). I think it make it still slightly more readable. Putting each one on a separate line seems excessive in that case.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
KtorZ: No, it is a sign that `DuplicateRecordFields` is a pain in the ass :upside_down_face:

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
KtorZ: We've got already quite many types all over the place when it comes to selections (and there are more in the transaction module like `UnsignedTx`). Thus, instead of introducing yet-a-new-type which has almost exactly the same semantic as an existing one but only differs on one particular field, I found it preferable to parameterize over the change type. 

Not only does it allow the reader to keep the same context when processing function using `SelectionResult TxOut` or `SelectionResult TokenBundle` (it's quite clear that it's roughly the same thing, it keeps the cognitive load low), but it also allow for re-using functions on that type quite easily. Plus, It makes type-signature quite clear about what's happening:

```hs
assignChangeAddresses
     :: SelectionResult TokenBundle
    -> SelectionResult TxOut
```

Neat.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
jonathanknowles: Hi @KtorZ I don't have a strong opinion on this.

However, I do think we would benefit from:

1. Adhering to our existing coding standards as much as possible.
2. If we want to deviate from the standard, have a discussion in the channel about how we want to revise it.

As I'm sure no-one likes to have these discussions in PRs! 😹 

In this case, our existing coding standard says to avoid variable length indentation:
https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#avoid-variable-length-indentation

I can completely understand the argument for aligning variable names in pattern matches, as it can aid with comprehension. But I'm not persuaded that vertical alignment is a good idea in the general case, for all the reasons pointed out in our standards document.

Perhaps we should discuss this in the channel, if we decide we want to revise our approach?

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
jonathanknowles: @KtorZ I can definitely sympathize with your argument.

But I also think there are drawbacks to having a function return part of its input within its output.

An unfamiliar reader of `performSelection` might very well wonder:
- Does `outputsCovered` really have the same contents as `outputsToCover`?
- Is it a subset, or perhaps something else?
- If I supply a non-empty list for `outputsToCover`, under what circumstances is it possible for `performSelection` to return an empty list?

Before this PR, there could be no such questions in the reader's mind, because that field didn't exist. But after applying this PR, the reader will have to examine the definition of `performSelection` to confirm the answer.

I think this is non-ideal, and would prefer to keep the range of `performSelection` as minimal as possible, but if we have to choose this option, I think we should:

1. Upgrade our property tests for `performSelection` to ensure that the intended relationship between `SelectionCriteria.outputsToCover` and `SelectionResult.outputsCovered` really does hold.

2. Describe the intended relationship between `SelectionCriteria.outputsToCover` and `SelectionResult.outputsCovered` in the documentation for `performSelection`, so that it's clear to the reader.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
jonathanknowles: > No, it is a sign that `DuplicateRecordFields` is a pain in the ass upside_down_face

Understood.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
jonathanknowles: @KtorZ I've created a PR (https://github.com/input-output-hk/cardano-wallet/pull/2469) which adds:
 
> 1. Upgrade our property tests for `performSelection` to ensure that the intended relationship between `SelectionCriteria.outputsToCover` and `SelectionResult.outputsCovered` really does hold.
> 2. Describe the intended relationship between `SelectionCriteria.outputsToCover` and `SelectionResult.outputsCovered` in the documentation for `performSelection`, so that it's clear to the reader.

I still believe this solution to be worse than simply adding a new concrete type (and leaving `performSelection` alone), but if we have to choose this option, I think that having tests would be beneficial.

KtorZ: > Before this PR, there could be no such questions in the reader's mind, because that field didn't exist. But after applying this PR, the reader will have to examine the definition of performSelection to confirm the answer.
> 
> I think this is non-ideal, and would prefer to keep the range of performSelection as minimal as possible

It seems like you're not objecting the parameterization so much but actually object the additions of the few other fields to the result. However, these are *necessary* for the rest of the code to work and to integrate with the new selection algorithms. Without that, we'd have to duplicate logic from `performSelection` from the same argument to re-construct the augmented result. 

This is generally not advised and has already caused us issues in the past. This PR actually resolves a lot of the logic duplication issues our current coin selection integration has and the best way to achieve this at this stage is to make sure that all information needed are returned by `performSelection`.

lib/core/src/Cardano/Wallet/Primitive/Types/Coin.hs
```diff
@@ -14,6 +14,8 @@ module Cardano.Wallet.Primitive.Types.Coin
     ( -- * Type
       Coin (..)
     , coinQuantity
+    , coinToInteger
+    , coinToNatural
```
rvl, line 18: Looks like coinToNatural isn't used in this PR?
coinToInteger is used for difference - could it be replaced with a difference function?

rvl (approved):
Looks good to me, nice clean ups with mkTransaction.

lib/core/src/Cardano/Wallet/Primitive/Types/Coin.hs
KtorZ: Yes indeed, it used to be used before your PR which changed many of the `Quantity` to `Coin`. The `coinToInteger` is only used in the readWithdrawal function; Integer are used here because the difference may actually underflow / be negative. I'll see if that can be written differently using only `distance` or `subtractCoin` :+1:

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
jonathanknowles: > It seems like you're not objecting the parameterization so much but actually object the additions of the few other fields to the result. 

That is indeed my main objection. I think the `performSelection` function is already quite complicated, and that the combination of inputs with outputs could be handled by another function that calls `performSelection`, returning some kind of record that combines `SelectionCriteria` with `SelectionResult`. This way, we wouldn't have to alter the definition of `performSelection`, or extend any of its property tests.

However, given our time constraints, I do think we can go with the solution in this PR, but with some extra small additions similar to:

- e628436fe88cd8905a557eebfcc921f987c82c6d
- 5e3d602ca4b8a0121a4b36b9e5ab3b766db7f497

(Feel free to cherry pick if you think these are useful.)

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1235,377 +1213,160 @@ normalizeDelegationAddress s addr = do
...
+    let dummyOutput  = TxOut dummyAddress (TokenBundle.fromCoin deposit)
+    let outs = dummyOutput :| []
+    selectAssets @ctx @s @k ctx wid tx outs $ \s sel ->
+        transform s (sel { outputsCovered = mempty })
+
+-- | Selects assets from the wallet's UTxO to satisfy the requested outputs in
+-- the given transaction context. In case of success, returns the selection
+-- and its associated cost. That is, the cost is equal to the difference between
+-- inputs and outputs.
+selectAssets
```
Anviking, line 1299: Wondered a bit about why `selectAssets` is not in `TransactionLayer` rather than `calcMinimumCoinValue`, `initSelectionCritereia`, `calcMinimumValue`.

It seems `readNextWithdrawal` needs separate access to `calcMinimumCost`. It is used from the Server.hs module, but the result is directly put back into `TransactionCtx`.

I suspect that logic could fit in `initSelectionCriteria`, such that Wallet.hs and Server.hs don't need to know about it.

KtorZ: > Wondered a bit about why selectAssets is not in TransactionLayer 

The main reason is because `selectAssets` contains the glue code for pulling information from the database and running the selection itself. The `TransactionLayer` is completely agnostic to the database. So, one could argue that it suffices to make the UTxO and protocol parameters arguments of the `selectAssets` function to make it work; but that is basically what the `performSelection` function is already. Moreover, we'll still need to put the glue / effectful code  somewhere to prepare the arguments. 

Beside, having separate functions as `calcMinimumCoinValue`, `initSelectionCriteria` and `calcMinimumValue` allows in principle for easier testing since these functions cover a very specific sub-part of the logic. 

> It seems readNextWithdrawal needs separate access to calcMinimumCost. It is used from the Server.hs module, but the result is directly put back into TransactionCtx. I suspect that logic could fit in initSelectionCriteria, such that Wallet.hs and Server.hs don't need to know about it.

I am afraid that this won't work because the users are actually driving the request for withdrawal, not the server. This is why the server needs to know about it, because it depends on what the user has chosen. If we were systematically withdrawing from the account without asking the user, I think we could indeed move that logic to `initSelectionCriteria`. Yet at the moment, withdrawals are only set in the `TransactionCtx` if:

a) They are large enough
b) They have been explicitly requested by users.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -376,23 +377,19 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
             -- because there is a fee for this tx
             [ expectResponseCode HTTP.status202
             , expectField (#status . #getApiT) (`shouldBe` Pending)
-            , expectField (#direction . #getApiT) (`shouldBe` Outgoing)
+            , expectField (#direction . #getApiT) (`shouldBe` Incoming)
             ]
         let txid = getFromResponse Prelude.id rq
-        let (Quantity quitFeeAmt) = getFromResponse #amount rq
-        let finalQuitAmt = Quantity (depositAmt ctx - quitFeeAmt)
+        let quitFeeAmt = getFromResponse #amount rq
```
KtorZ, line 383: cc @piotr-iohk This was also somewhat plain wrong. At least, now it's fixed.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -823,7 +820,7 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
     describe "STAKE_POOLS_JOIN_01x - Fee boundary values" $ do
         it "STAKE_POOLS_JOIN_01x - \
             \I can join if I have just the right amount" $ \ctx -> runResourceT $ do
-            w <- fixtureWalletWith @n ctx [costOfJoining ctx + depositAmt ctx]
+            w <- fixtureWalletWith @n ctx [costOfJoining ctx + depositAmt ctx + minUTxOValue]
```
KtorZ, line 823: Note: a little downside of the new unified approach for coin selection is that the algorithm must always be able to create a change output. Which means that, even is the wallet contains _just_ enough to pay for fees and deposit, one would still get an error if there's not enough to cover for the min utxo value of the change output. The actual cost of the transaction is unchanged however, because the change goes back to the wallet.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -879,7 +875,7 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
             verify rQuit
                 [ expectResponseCode HTTP.status202
                 , expectField (#status . #getApiT) (`shouldBe` Pending)
-                , expectField (#direction . #getApiT) (`shouldBe` Outgoing)
+                , expectField (#direction . #getApiT) (`shouldBe` Incoming)
```
KtorZ, line 878: cc @piotr-iohk

lib/core/test/unit/Cardano/WalletSpec.hs
```diff
@@ -576,76 +560,6 @@ walletListTransactionsSorted wallet@(wid, _, _) _order (_mstart, _mend) history
...
--- | Manufacture a coin selection that would result in the given fee.
-coinSelectionForFee :: FeeGen -> CoinSelection
-coinSelectionForFee (FeeGen (Coin fee)) = mempty
-    { CS.inputs =
-        [(TxIn (Hash "") 0, TxOut (Address "") (coinToBundle (1 + fee)))]
-    , CS.outputs =
-        [TxOut (Address "") (coinToBundle 1)]
-    }
-  where
-    coinToBundle = TokenBundle.fromCoin . Coin
```
KtorZ, line 647: Whoops. I was maybe a bit too fast in removing this one thinking it was something else. We likely still want something similar.

KtorZ:
bors try

KtorZ:
bors try

KtorZ:
bors try

ffs.

rvl:
What was happening with bors? The other day it seemed like the GitHub CI status and commit notifications APIs were really slow. Bors relies on notifications from github to trigger builds.

jonathanknowles:
~~bors r+~~

I just noticed that @KtorZ 's intention (above) was to [try](https://github.com/input-output-hk/cardano-wallet/pull/2462#issuecomment-768603827), rather than to [merge](https://github.com/input-output-hk/cardano-wallet/pull/2462#issuecomment-768602589). So I cancelled this action and reinitiated another `bors try` below.

Apologies for the noise!

jonathanknowles:
bors cancel

jonathanknowles:
bors try

rvl:
bors r-

(hydra build was user cancelled anyway)

lib/core-integration/src/Test/Integration/Framework/TestData.hs
```diff
@@ -225,19 +219,13 @@ versionLine = "Running as v" <> pack (showFullVersion version gitRevision)
...
-    , (show minUTxOValue)
-    , " Lovelace. In the current transaction the following pieces are not"
-    , " satisfying this condition"
-    ]
+errMsg403MinUTxOValue :: String
+errMsg403MinUTxOValue =
+    "Some outputs specifies an Ada value that is too small. Indeed, there's a \
+    \minimum Ada value specified by the protocol that each output must satisfy. \
+    \I'll handle that minimum value myself when you do not explicitly specify \
+    \an Ada value for outputs. Otherwise, you must specify enough Ada."
```
jonathanknowles:
```suggestion
    "Some outputs have ada values that are too small. There's a \
    \minimum ada value specified by the protocol that each output must satisfy. \
    \I'll handle that minimum value myself when you do not explicitly specify \
    \an ada value for an output. Otherwise, you must specify enough ada."
```

lib/core-integration/src/Test/Integration/Framework/TestData.hs
```diff
@@ -256,15 +244,9 @@ errMsg400StartTimeLaterThanEndTime startTime endTime = mconcat
...
-    \ more additional inputs, but I can't do that without increasing\
-    \ the size of the transaction beyond the acceptable limit."
-
-errMsg403DelegationFee :: Natural -> String
-errMsg403DelegationFee n =
-    "I'm unable to select enough coins to pay for a delegation certificate. \
-    \I need: " ++ show n ++ " Lovelace."
+errMsg403Fee =
+    "I am unable to finalize the transaction as there are not enough Ada I can \
+    \use to pay for either fees, or minimum Ada value in change outputs."
```
jonathanknowles:
```suggestion
    "I am unable to finalize the transaction, as there is not enough ada I can \
    \use to pay for fees, or to satisfy the minimum ada quantities of change \
    \outputs."
```

lib/core-integration/src/Test/Integration/Framework/TestData.hs
```diff
@@ -276,26 +258,16 @@ errMsg403NotAnIcarusWallet =
     "I cannot derive new address for this wallet type.\
     \ Make sure to use a sequential wallet style, like Icarus."
 
-errMsg403NotEnoughMoney_ :: String
-errMsg403NotEnoughMoney_ =
-    "I can't process this payment because there's not enough UTxO available in \
+errMsg403NotEnoughMoney :: String
+errMsg403NotEnoughMoney =
+    "I can't process this payment because there's not enough funds available in \
```
jonathanknowles:
```suggestion
    "I can't process this payment as there are not enough funds available in \
```

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -440,6 +438,11 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
                 ]
 
     it "TRANS_CREATE_03 - 0 balance after transaction" $ \ctx -> runResourceT $ do
+        liftIO $ pendingWith
+            "This test requires to know exactly how the underlying selection \
+            \implementation works. We may want to revise this test completely \
+            \without what we'll have to update it for every single change in \
+            \the fee calculation or selection algorithm."
```
jonathanknowles:
```suggestion
        liftIO $ pendingWith
            "This test relies on knowing exactly how the underlying selection \
            \implementation works. We may want to revise this test completely \
            \otherwise we'll have to update it for every single change in \
            \the fee calculation or selection algorithm."
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1235,377 +1212,182 @@ normalizeDelegationAddress s addr = do
...
--- coin selection for the given outputs. In order to construct (and
--- sign) an actual transaction, use 'signPayment'.
-selectCoinsForPayment
-    :: forall ctx s k.
-        ( HasTransactionLayer k ctx
-        , HasLogger WalletLog ctx
-        , HasDBLayer s k ctx
+-- | Augments the given outputs with new outputs. These new outputs corresponds
+-- to change outputs to which new addresses are being assigned to. This updates
+-- the wallet state as it needs to keep track of new pending change addresses.
```
```suggestion
-- | Augments the given outputs with new outputs. These new outputs correspond
-- to change outputs to which new addresses have been assigned. This updates
-- the wallet state as it needs to keep track of new pending change addresses.
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1235,377 +1212,182 @@ normalizeDelegationAddress s addr = do
...
-                    -
-                    W.balance' (concatMap inputs cs)
-            pure (cs, Coin leftovers)
-        _ -> throwE (ErrSelectForMigrationEmptyWallet wid)
+    liftIO $ traceWith tr $ MsgSelectionStart utxo outs
+    sel <- performSelection
+        (calcMinimumCoinValue tl pp)
+        (calcMinimumCost tl pp tx)
+        (initSelectionCriteria tl pp tx utxo outs)
+    liftIO $ traceWith tr $ MsgSelectionDone sel
```
jonathanknowles, line 1329: Will this produce very large output in the log?

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1892,27 +1488,19 @@ mkTxMeta ti' blockHeader wState tx cs expiry =
...
   where
     slotStartTime' = interpretQuery ti . slotToUTCTime
       where
-        ti = neverFails
-                "mkTxMeta slots should never be ahead of the node tip"
-                ti'
+        ti = neverFails "mkTxMeta slots should never be ahead of the node tip" ti'
 
-    ourCoins :: TxOut -> Maybe Natural
+    ourCoins :: TxOut -> Maybe Coin
```
```suggestion
    ourCoin :: TxOut -> Maybe Coin
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2163,14 +1770,28 @@ estimateFeeForCoinSelection deposit'
...
                 Left e
             (_, samples') ->
                 Right samples'
 
     repeats = 100 -- TODO: modify repeats based on data
 
+    -- | When estimating fee, it is rather cumbersome to return "cannot cover fee"
+    -- whereas clients are just asking for an estimation. Therefore, we convert
+    -- cannot cover errors into the necessary fee amount, even though there isn't
+    -- enough in the wallet to cover for these fees.
```
```suggestion
    -- if clients are just asking for an estimation. Therefore, we convert
    -- "cannot cover" errors into the necessary fee amount, even though there isn't
    -- enough in the wallet to cover for these fees.
```

lib/core/src/Cardano/Wallet/Transaction.hs
```diff
@@ -80,80 +85,35 @@ data TransactionLayer k = TransactionLayer
...
+    , calcMinimumCost
+        :: ProtocolParameters
+            -- Current protocol parameters
+        -> TransactionCtx
+            -- Additional information about the transaction
+        -> SelectionSkeleton
+            -- An intermediate representation of an ongoing selection
+        -> Coin
+        -- ^ Compute a minimal fee amount necessary to pay for a given selection
+        -- This also include necessary deposits.
```
```suggestion
        -- This also includes necessary deposits.
```

jonathanknowles (approved):
Hi @KtorZ 

This looks pretty good to me. I still have _some_ reservations about the types, but we can always evolve those later on. 👍🏻 

I've just left a few suggestions w.r.t. wording of user-facing errors.

lib/core/src/Cardano/Wallet.hs
KtorZ: Arf. Good point, because of the `utxoRemaining` in the `SelectionResult`, this can be quite large -- though only at debug severity. I'll tweak the message to text result to prevent this.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
Anviking: Isn't this weird?

I guess neither label make _perfect_ sense, but `Outgoing` more so?

Anviking: Or are we getting the deposit back now?

lib/core/src/Cardano/Wallet.hs
KtorZ: The `Buildable` instance was actually already sound:

```hs
buildSelectionResult
    :: (NonEmpty change -> Builder)
    -> SelectionResult change
    -> Builder
buildSelectionResult changeF s@SelectionResult{inputsSelected,extraCoinSource} =
    mconcat
        [ nameF "inputs selected" (inputsF inputsSelected)
        , nameF "extra coin input" (build extraCoinSource)
        , nameF "outputs covered" (build $ outputsCovered s)
        , nameF "change generated" (changeF $ changeGenerated s)
        , nameF "size utxo remaining" (build $ UTxOIndex.size $ utxoRemaining s)
        ]
  where
    inputsF :: NonEmpty (TxIn, TxOut) -> Builder
    inputsF = blockListF' "+" tupleF
```

so, all good.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
KtorZ: We do want Incoming because of the deposit indeed, there's more coming in than out, and that's how we define Incoming

KtorZ:
bors try

KtorZ:
Arg.

PR #2464, opened 2021-01-25
Remove unused ErrListPools

Anviking:
# Issue Number

ADP-647

# Overview

- [x] Remove `ErrStakeDistributionQuery` which is unused since #2449
- [x] Remove `ErrListPoolsPastHorizonException` which is also unused (we use `unsafeExtendSafeZone`, so it will never fail)
- [x] Remove the `ErrListPools` type entirely 

# Comments


rvl (approved):
Nice :ok_hand:

Anviking:
This is odd…
bors r+

PR #2466, opened 2021-01-25
Expose account public key in api

paweljakubas:
# Issue Number

adp-544

# Overview

- [x] I have added endpoint definition in swagger
- [x] I have added Api scaffolding and needed types, instances
- [x] I have added wallet impl
- [x] I have updated unit tests
- [x] I have added illustrative integration test

# Comments


specifications/api/swagger.yaml
```diff
@@ -3452,6 +3482,24 @@ paths:
         - *parametersAddressState
       responses: *responsesListAddresses
 
+  /wallets/{walletId}/keys:
```
rvl: It would be more coherent for API users if the endpoint for retrieving this key could be similar to the getWalletKey endpoint. So parameters related to derivation path and chain code are on the request path. Only the passphrase would be in the request body.

KtorZ: I agree, I'd be better to have the index as a path parameter here :+1:

paweljakubas: done- please refer HEAD

specifications/api/swagger.yaml
```diff
@@ -3789,6 +3828,25 @@ paths:
...
 
+  /wallets/{walletId}/keys/{index}:
+    post:
+      operationId: postAccountKey
+      tags: ["Keys"]
+      summary: Create
+      description: |
+        <p align="right">status: <strong>stable</strong></p>
+
+        Retrieve account public key from the wallet.
```
piotr-iohk, line 3865: Perhaps a note would be handy:

"Note: Only `Hard` indexes are supported by this endpoint."

(similar as it is on https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/getWalletKey)

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2351,13 +2353,50 @@ derivePublicKey ctx wid role_ ix = db & \DBLayer{..} -> do
...
+        , WalletKey k
+        , s ~ SeqState n k
+        )
+    => ctx
+    -> WalletId
+    -> Passphrase "raw"
+    -> DerivationIndex
+    -> ExceptT ErrReadAccountPublicKey IO (k 'AccountK XPub)
+readPublicAccountKey ctx wid pwd ix = db & \DBLayer{..} -> do
+    acctIx <- withExceptT ErrReadAccountPublicKeyInvalidIndex $ guardHardIndex ix
```
KtorZ, line 1962: I wonder if from a user-experience perspective, it isn't better to only accept indexes in the soft range, and manually translate them to the hard range. So, when using the API, one would simply post `"index": 0` without having to worry about it being hardened or soft. 

Not sure if this is confusing or better :thinking:

KtorZ: Ah, nevermind. Seeing the rest of the PR, I remembered that the JSON instance for `DerivationIndex` uses the `H` suffix. So account indexes are specified by users as `"index": "0H"` :+1:

KtorZ (approved):
LGTM :+1:

specifications/api/swagger.yaml
paweljakubas: thanks, done

KtorZ:
Will merge in batch with #2462

lib/shelley/test/unit/Cardano/Wallet/Shelley/.#CompatibilitySpec.hs
```diff
@@ -0,0 +1 @@
+pawel@pure.2559:1611843900
```
KtorZ: Arf.

PR #2468, opened 2021-01-26
Replace `hardfork_at` with `eras` object, and add node_era

Anviking:
# Issue Number

ADP-676, ADP-681

# Overview

- [x] Instead of storing ProtocolParameters in DB, just keep them in a `TVar`
- [x] Remove problem where the wallet might use genesis-derived protocol parameters and era on wallet start up, until the node first changed tip.
- [x] Replace `hardfork_at` with info of all eras
- [x] Add new `node_era` field in GET /network/information

- [ ] <s>See if we can order the eras correctly in the `era` object (for when the JSON is viewed by humans)</s> Seems impossible to do.

# Comments

- Could add absolute `slot` field also
- Arguably a bit sad that the eras get sorted seemingly alphabetically in the json

```json
$ cardano-wallet network parameters
{
    "slot_length": {
        "quantity": 1,
        "unit": "second"
    },
    "decentralization_level": {
        "quantity": 74,
        "unit": "percent"
    },
    "genesis_block_hash": "5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb",
    "blockchain_start_time": "2017-09-23T21:44:51Z",
    "desired_pool_number": 500,
    "epoch_length": {
        "quantity": 432000,
        "unit": "slot"
    },
    "eras": {
        "shelley": {
            "epoch_number": 208
        },
        "mary": null,
        "byron": {
            "epoch_start_time": "2017-09-23T21:44:51Z",
            "epoch_number": 0
        },
        "allegra": {
            "epoch_start_time": "2020-12-16T21:44:51Z",
            "epoch_number": 236
        }
    },
    "active_slot_coefficient": {
        "quantity": 5,
        "unit": "percent"
    },
    "security_parameter": {
        "quantity": 2160,
        "unit": "block"
    },
    "minimum_utxo_value": {
        "quantity": 1000000,
        "unit": "lovelace"
    }
}
```


lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -2190,6 +2190,11 @@ taggedSumTypeOptions base opts = base
     { sumEncoding = TaggedObject (_tagFieldName opts) (_contentsFieldName opts)
     }
 
+explicitNothingRecordTypeOptions :: Aeson.Options
+explicitNothingRecordTypeOptions = defaultRecordTypeOptions
+    { omitNothingFields = False
```
Anviking, line 2213: Not sure why this is enabled by default…

"Is the `mary` key not present because the wallet is not supporting `mary`, or because the `mary` fork has not yet occurred?"

Explicit `null` for the latter case seems much clearer.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Network.hs
```diff
@@ -51,7 +51,8 @@ spec = describe "SHELLEY_NETWORK" $ do
             [ expectField #decentralizationLevel (`shouldBe` d)
             , expectField #desiredPoolNumber (`shouldBe` nOpt)
             , expectField #minimumUtxoValue (`shouldBe` Quantity minUTxOValue)
-            , expectField #hardforkAt (`shouldNotBe` Nothing)
+            , expectField (#eras . #shelley) (`shouldNotBe` Nothing)
+            , expectField (#eras . #byron) (`shouldNotBe` Nothing)
```
KtorZ, line 55: :+1:

Anviking, line 55: Integration tests can in practice be run on any shelley-based era. So simply testing that _at least_ `byron` and `shelley` are not null, should be pretty good.

paweljakubas: but then you will not receive information about mary and allegra, right? :thinking:

specifications/api/swagger.yaml
```diff
@@ -1410,6 +1410,30 @@ components:
...
+        shelley: *nullableEpochInfo
+        allegra: *nullableEpochInfo
+        mary: *nullableEpochInfo
+      description: |
+
+        If and when each era started or will start.
+
+        The object is keyed by era names. The values either describe the epoch boundary
+        when the era starts (can be in the future or in the past), or are null if not yet
+        confirmed on-chain.
```
Anviking, line 1440: Because of
> (can be in the future or in the past)

wouldn't it make sense to provide a `era` field either here, or in `/network/information`?

A non-null `mary` does not necessarily mean that we're in Mary. We could also be just before Mary.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Network.hs
Anviking: Ok, fine, we can probably make the expectation dependent on the `LOCAL_CLUSTER_ERA` value, by storing it in the test `Context`

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -749,15 +751,30 @@ data ApiNetworkParameters = ApiNetworkParameters
     , decentralizationLevel :: !(Quantity "percent" Percentage)
     , desiredPoolNumber :: !Word16
     , minimumUtxoValue :: !(Quantity "lovelace" Natural)
-    , hardforkAt :: !(Maybe ApiEpochInfo)
+    , eras :: !ApiEraInfo
```
paweljakubas, line 755: why not to have 
```
eras: [ApiEraInfo]
...
data Eras = Byron | Shelley | Allegra | Mary deriving (Eq, Ord, Show)

data ApiEraInfo = ApiEraInfo {
   hardForkToEra :: !Era
   hardFromEra :: !(Maybe Era)
   timeOfTransition:: !(Maybe ApiEpochInfo)
}
```
?

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -953,6 +955,38 @@ newtype ActiveSlotCoefficient
...
 
+-- |
+--
+-- It is expected that there is an order, @byron, shelley, allegra, mary@, by
+-- which the @Maybe@ fields are filled in.
+--
+-- It might be cumbersome to work with this type. /But/ we don't need to. A
+-- product of @Maybe@ is both what we can query from the node, and
+-- what we need to provide in the wallet API.
+data EraInfo info = EraInfo
```
paweljakubas, line 966: If you opt for my suggestion of ApiEraInfo then this part should also comply

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: I think it is more robust in a sense that when new hard forks arrive they will just add new element to list. But maybe there is some rationale behind this way of storing era transitions?

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -709,19 +709,20 @@ mkTipSyncClient tr np localTxSubmissionQ onPParamsUpdate onInterpreterUpdate = d
             onPParamsUpdate pp sp
 
     let queryParams = do
-            mb <- currentEra >>= \case
-                AnyCardanoEra ShelleyEra ->
-                    LSQry $ QueryAnytimeShelley GetEraStart
-                _ ->
-                    LSQry $ QueryAnytimeMary GetEraStart
+            eraBounds <- W.EraInfo
+                <$> LSQry (QueryAnytimeByron GetEraStart)
```
paweljakubas, line 709: :+1:

lib/core/src/Cardano/Wallet/Api/Types.hs
Anviking: @KtorZ advocated for an object keyed by era-name since Daedalus wants to do stuff like `isAllegra` and `isShelley`, and already have to be highly aware of e.g. the ordering of them.

Although with https://github.com/input-output-hk/cardano-wallet/pull/2468#discussion_r565335344 in mind, I wonder if this changes 🤔

specifications/api/swagger.yaml
```diff
@@ -1410,6 +1410,30 @@ components:
     ApiNetworkClock: &ApiNetworkClock
       <<: *networkInformationNtpStatus
 
+    nullableEpochInfo: &nullableEpochInfo
+      <<: *epochInfo
+      nullable: True
+
+    ApiEraInfo: &ApiEraInfo
```
paweljakubas, line 1427: the same here in case ApiEraInfo is refactored

lib/core/src/Cardano/Wallet/Api/Types.hs
Anviking: Seems we should indeed add some kind of  `currentEra` field somewhere.

Maybe we should make this a list then, yes 🤔 

Off the top of my head I would probably just do
```
data ApiEraInfo = ApiEraInfo {
   eraName :: !Era
   start :: !(Maybe ApiEpochInfo)
}
```
since `hardFromEra` doesn't provide any extra information.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Network.hs
Anviking: Hm, seems unexpectedly tedious though… (Can't use a shelley type in cardano-wallet-core-integration) Will look more at it later.

lib/core/src/Cardano/Wallet/Api/Types.hs
KtorZ: I agree that the absence of a field doesn't have the same meaning as a `null` value. I think that in the context of our API indeed, we most likely always want fields corresponding to Maybe values to be present :thinking:

lib/core/test/data/Cardano/Wallet/Api/ApiEraInfo.json
```diff
@@ -0,0 +1,161 @@
...
+        {
+            "shelley": {
+                "epoch_start_time": "1859-05-28T14:00:00Z",
+                "epoch_number": 23155
+            },
+            "mary": {
+                "epoch_start_time": "1891-03-26T08:00:00Z",
+                "epoch_number": 53
+            },
+            "byron": null,
```
KtorZ, line 13: That would be problematic in practice :)

KtorZ (approved):
Sounds good, provided that we make sure of the behavior on start-up with regards to protocol parameters. I'd expect the server to block on a particular request until protocol parameters become available.

specifications/api/swagger.yaml
paweljakubas: I thought it would go like this : we know something like mary will happen (null), then we know when it happens (by specifying era it starts), and we also now current slot/era. So everything is fine;-)

lib/core/src/Cardano/Wallet/Api/Types.hs
Anviking: In-doubt I would stick to the current object format. I think it still has its qualities too. If you feel strongly about it, feel free to disagree.

But I imagine, if a need arises where a list would be great for Daedalus, we can change it then. Currently the need for this is unclear.

paweljakubas: you got me wrong Johannes. I like your proposal, just was thinking loudly and sharing some ideas ;-) no pun intended.great job!

paweljakubas (approved):

Issue #2393, opened 2020-12-11
Flaky test: Validate generators & shrinkers, Shrinker for CreateWallet


Anviking:
# Context
 

`Validate generators & shrinkers` was added recently in #2354, to catch problems with our shrinkers and generators. Seems the CreateWallet one is now failing.

https://github.com/input-output-hk/cardano-wallet/pull/2357#issuecomment-740124434
https://github.com/input-output-hk/cardano-wallet/pull/2382#issuecomment-742453481

Perhaps this could be the underlying cause for #2394 too.

# Test Case

https://github.com/input-output-hk/cardano-wallet/blob/60aee7da3932a4451f32b0897a27f1dbc31997fe/lib/core/test/unit/Cardano/Wallet/DB/StateMachine.hs#L1281-L1313

# Failure / Counter-example

```
Build failed:

ci/hydra-build:required

 test/unit/Cardano/Wallet/DB/StateMachine.hs:1298:28:
 1) Cardano.Wallet.DB.Sqlite, Validate generators & shrinkers, Shrinker for CreateWallet
 Timed out.

 To rerun use: --match "/Cardano.Wallet.DB.Sqlite/Validate generators & shrinkers/Shrinker for CreateWallet/"

Randomized with seed 809167721
```

---

# Resolution 

---

# QA

KtorZ:
Fixed in #2471

PR #2471, opened 2021-01-28
Fix flaky `validateGenerators`

KtorZ:
# Issue Number

#2393 

# Overview

- deb387a4f0389e035a3feebdbb5610361cc751c3
  :round_pushpin: **rework 'validateGenerators' for the db arbitrary instance to be more reliable**
    And also give more information on failure. This should reduce the flakyness and provide more metrics on failure.

- 977078cc5881770612dc217fed6e8d594df67771
  :round_pushpin: **rework shrinkers on PolicyId, AssetName and TxId to be more aggressive**
    There's really nothing simpler between 'TokenB' compared to 'TokenC', or between '1111111111111111111111111' compared to '22222222222222222222222'

  The goal of a shrinker is to provide smaller counter examples, if shrinking a value doesn't make it smaller, the shrinker is pointless. In this very case, the shrinkers are actually sucking of CPU time for nothing as noted by the 'validateGenerators' property sometimes failing to shrink large data structures in a reasonable time.

  However, allowing values to shrink down to one common denominator help to make a counter example more readable by reducing the amount of noise in an output. That is, it's easier if all values of a particular types are the same. Thus, this commit changes shrinkers to reach that common denominator much more aggressively (in one step).

# Comments

Occurred quite often lately, so I thought I'd give it a shot.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy/Gen.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy/Gen.hs
@@ -32,7 +32,11 @@ genTokenNameSmallRange :: Gen TokenName
 genTokenNameSmallRange = elements tokenNamesSmallRange
 
 shrinkTokenNameSmallRange :: TokenName -> [TokenName]
-shrinkTokenNameSmallRange name = filter (< name) tokenNamesSmallRange
+shrinkTokenNameSmallRange x
+    | x == simplest = []
+    | otherwise = [simplest]
+  where
+    simplest = head tokenNamesSmallRange
 
 tokenNamesSmallRange :: [TokenName]
 tokenNamesSmallRange = UnsafeTokenName . B8.snoc "Token" <$> ['A' .. 'D']
@@ -46,7 +50,11 @@ genTokenNameMediumRange :: Gen TokenName
 genTokenNameMediumRange = elements tokenNamesMediumRange
 
 shrinkTokenNameMediumRange :: TokenName -> [TokenName]
-shrinkTokenNameMediumRange name = filter (< name) tokenNamesMediumRange
+shrinkTokenNameMediumRange x
+    | x == simplest = []
+    | otherwise = [simplest]
+  where
+    simplest = head tokenNamesMediumRange
 
 tokenNamesMediumRange :: [TokenName]
 tokenNamesMediumRange = UnsafeTokenName . B8.snoc "Token" <$> ['A' .. 'Z']
@@ -59,7 +67,11 @@ genTokenPolicyIdSmallRange :: Gen TokenPolicyId
 genTokenPolicyIdSmallRange = elements tokenPolicies
 
 shrinkTokenPolicyIdSmallRange :: TokenPolicyId -> [TokenPolicyId]
-shrinkTokenPolicyIdSmallRange policy = filter (< policy) tokenPolicies
+shrinkTokenPolicyIdSmallRange x
+    | x == simplest = []
+    | otherwise = [simplest]
+  where
+    simplest = head tokenPolicies
 
 tokenPolicies :: [TokenPolicyId]
 tokenPolicies = mkTokenPolicyId <$> ['A' .. 'D']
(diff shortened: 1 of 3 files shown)


lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy/Gen.hs
```diff
@@ -32,7 +32,7 @@ genTokenNameSmallRange :: Gen TokenName
 genTokenNameSmallRange = elements tokenNamesSmallRange
 
 shrinkTokenNameSmallRange :: TokenName -> [TokenName]
-shrinkTokenNameSmallRange name = filter (< name) tokenNamesSmallRange
+shrinkTokenNameSmallRange _ = []
```
Anviking: Wouldn't it be better to have a shrinker like
```
if (x /= simplest) then [simplest] else []
```
to make counterexamples less noisy for human eyes?

KtorZ: Yes. But is `AAAAAA` simpler than `CCCCCC` ? I hardly think so :upside_down_face:

KtorZ: Right now, the generator produces names like `TokenA`, `TokenB`, `TokenC` etc ... And the only thing the shrinker does is to shrink lexicographically. So, `TokenC` is shrinked to [`TokenA`, `TokenB`] ... Same goes for the policy id and transaction ids.. This only takes needless CPU time and doesn't really make the output more readable. Hence their removal.

KtorZ: Arguably, yes, there could be a one-step shrinker trying to shrink all things down to one "blessed" value. That should still be efficient enough and make the output a bit more uniform in case where there are many `TokenName` or `PolicyId` involved in the output :+1:

KtorZ: Done :+1:

jonathanknowles: Hi @KtorZ 

I'm not totally convinced these shrinkers are pointless, or even problematic.

The point here is not to make the output of _individual_ `TokenName` values more readable, but to generate simpler counterexamples where there are _multiple_ `TokenName` values involved (or multiple `TokenPolicyId` values).

This is useful in the case of tests for `TokenBundle` (i.e., in `TokenBundleSpec`.)

Let's say we have the following fully-shrunk counterexamples:
1. a counterexample with two `TokenName` values `x` and `y`, where both are different.
2. a counterexample with two `TokenName` values `x` and `y`, where both are the same.

In case no. 1, we know it's probably important for the values to be different for the test to fail.
In case no. 2, we know that it might be important for the values to be the same for the test to fail.

## Proposal

Instead of deleting these shrinkers or amending them for all places that use them, simply remove calls to them in places where they're troublesome.

For example, in the QSM tests, we could just remove the reference to the shrinkers in the relevant `Arbitrary` instances.

So rather than writing:
```hs
instance Arbitrary TokenName where
    arbitrary = genTokenNameSmallRange
    shrink = shrinkTokenNameSmallRange
```

We could write the following instead:
```hs
instance Arbitrary TokenName where
    arbitrary = genTokenNameSmallRange
    -- No shrinking
```
 
Or if it's embedded in a larger structure, such as a `TokenMap`, then we could write:
```hs
instance Arbitrary TokenMap where
    arbitrary = genTokenMapSmallRange
    -- No shrinking
```

KtorZ: Please, see the updated commit :+1:

Anviking (approved):
Edit: some slight reservations about making the shrinkers worse, but 👌 / worth a shot too

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy/Gen.hs
Anviking: > We could write the following instead:

Rather than completely removing shrinking for the state machine tests, it would be better to keep `if (x /= simplest) then [simplest] else []`.

Otherwise, I agree with @jonathanknowles's sentiment above.

Anviking: > This is useful in the case of tests for TokenBundle (i.e., in TokenBundleSpec.)

Didn't think of this 👍

KtorZ:
bors retry

KtorZ:
Not sure why Hydra isn't restarted the jobs. I am almost tempted to merge this one manually because the changes are solely located in the unit tests which are passing just fine.

KtorZ:
Seems like there's a jobset running though: https://hydra.iohk.io/eval/1011601#tabs-unfinished

PR #2480, opened 2021-01-29
WIP: Remove core-integration package

Anviking:
# Issue Number

ADP-682

# Overview

- [x] Merge BenchShared and LatencyBenchShared with their respective benchmarks
- [x] Remove `core-integration` in favour for an… internal framework that can depend on `cardano-wallet`
    - Both bench:latency and cardano-wallet:integration depend on the test code
- [ ] Does CI like it?
- [ ] Fix weeder 

# Comments


Anviking:
Don't think this will work out.

rvl:
Thanks for having a try... When the time comes to do this for real, perhaps one way of making it easier would be to:

1. Merge everything into a single package (except a couple of packages which are obviously separate).
2. Then re-organise modules into separate packages. Some modules may need to be split if they address multiple "concerns".

What do you think?

PR #2482, opened 2021-02-01
Give priority to singleton asset bundles during coin selection

jonathanknowles:
# Issue Number

ADP-605

# Overview

This PR adjusts the coin selection algorithm to give priority to **_single-asset inputs_** when selecting for a given asset, similar to how we already give priority to coin-only inputs when selecting for coins.

By doing so, we:
- decrease the chance of selecting a **non-user-specified asset**.
- decrease the chance of **overselecting** a user-specified asset.
- potentially decrease the size and cost of a transaction in the case where the more relaxed filter would have selected a multi-asset input.

# Changes

- ac578218b32d265ba050f9c3336c2033c33b4582
  Extends the `SelectionFilter` type of `UTxOIndex`, adding a `WithAssetOnly` constructor. This allows callers of `selectRandom` to specify that they want to select an entry containing only the specified non-ada asset, and no other non-ada assets.
- 830e36f80e84946a0da89cb18ebb2bfe8167ad9d
  When selecting for UTxO entries containing a particular non-ada asset, give priority to entries that contain no other non-ada assets apart from the given asset.
- fc960182dc4877fe85ab6881f337a04af17635bd
  Test that selection lenses obey the correct priorities:
    - The `coinSelectionLens` should give priority to _coins_.
    - The `assetSelectionLens` should give priority to _singleton asset token bundles_: token bundles where there is only one non-ada asset.
- 1c8e0884ed06eb81ed0f5dab796a39c54dd89c3a
   Extracts out the responsibility for traversing through selection filters to function `selectRandomWithPriority`. As a consequence, the `limitReached` condition is now tested just once, regardless of how many filters are supplied, rather than repeatedly testing it on each recursive call to `selectMatchingQuantity`.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
@@ -85,7 +85,7 @@ import Cardano.Wallet.Primitive.Types.TokenBundle
 import Cardano.Wallet.Primitive.Types.TokenMap
     ( AssetId )
 import Cardano.Wallet.Primitive.Types.Tx
-    ( TxIn, TxOut, txOutIsCoin )
+    ( TxIn, TxOut )
 import Cardano.Wallet.Primitive.Types.UTxO
     ( UTxO (..) )
 import Control.DeepSeq
@@ -95,7 +95,7 @@ import Control.Monad.Random.Class
 import Data.Function
     ( (&) )
 import Data.Generics.Internal.VL.Lens
-    ( view )
+    ( over, view )
 import Data.Generics.Labels
     ()
 import Data.Map.Strict
@@ -141,14 +141,21 @@ import qualified Data.Set.Strict.NonEmptySet as NonEmptySet
 -- the 'checkInvariant' function.
 --
 data UTxOIndex = UTxOIndex
-    { index
+    { assetsAll
         :: !(Map AssetId (NonEmptySet TxIn))
+        -- An index of all entries that contain at least one non-ada asset.
+    , assetsSingleton
+        :: !(Map AssetId (NonEmptySet TxIn))
+        -- An index of all entries that contain exactly one non-ada asset.
     , coins
         :: !(Set TxIn)
+        -- An index of all entries that contain no non-ada assets.
     , balance
         :: !TokenBundle
+        -- The total balance of all entries.
     , utxo
         :: !(Map TxIn TxOut)
+        -- The complete set of all entries.
     }
     deriving (Eq, Generic, Read, Show)
 
@@ -162,7 +169,8 @@ instance NFData UTxOIndex
 --
 empty :: UTxOIndex
 empty = UTxOIndex
-    { index = Map.empty
+    { assetsAll = Map.empty
+    , assetsSingleton = Map.empty
     , coins = Set.empty
     , balance = TokenBundle.empty
     , utxo = Map.empty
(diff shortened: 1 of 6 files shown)


lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
```diff
@@ -245,20 +258,24 @@ insertMany = flip $ F.foldl' $ \u (i, o) -> insert i o u
...
+        -- entry is a member of the index, and therefore the balance must be
+        -- greater than or equal to the value of this output:
+        & over #balance (`TokenBundle.unsafeSubtract` view #tokens o)
+        & over #utxo (Map.delete i)
+        & case categorizeTxOut o of
+            IsCoin ->
+                over #coins (Set.delete i)
+            IsCoinWithSingletonAsset a -> id
+                . over #assetsSingleton (`deleteEntry` a)
+                . over #assetsAll (`deleteEntry` a)
```
KtorZ, line 277: Ok. So, `assetAll` also contain singleton assets (as stated in a comment above :+1:).

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
```diff
@@ -343,6 +363,37 @@ selectRandom u selectionFilter =
...
+    -> NonEmpty SelectionFilter
+    -- ^ A list of selection filters to be traversed in descending order of
+    -- priority, from left to right.
+    -> m (Maybe ((TxIn, TxOut), UTxOIndex))
+selectRandomWithPriority u filters =
+    selectRandom u (NE.head filters) >>= \case
+        Just en -> pure $ Just en
+        Nothing -> case NE.nonEmpty $ NE.tail filters of
+            Just fs -> selectRandomWithPriority u fs
+            Nothing -> pure Nothing
```
KtorZ: Debatable to have switched to `NonEmpty SelectionFilter` after all :upside_down_face:

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
```diff
@@ -368,16 +448,18 @@ entriesWithAsset a = maybe mempty NonEmptySet.toSet . Map.lookup a . index
...
+    & over #balance (`TokenBundle.add` view #tokens o)
+    & over #utxo (Map.insert i o)
+    & case categorizeTxOut o of
+        IsCoin ->
+            over #coins (Set.insert i)
+        IsCoinWithSingletonAsset a -> id
+            . over #assetsSingleton (`insertEntry` a)
+            . over #assetsAll (`insertEntry` a)
+        IsCoinWithMultipleAssets as ->
+            over #assetsAll (flip (F.foldl' insertEntry) as)
```
KtorZ, line 458: :+1:

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -843,6 +855,74 @@ prop_runSelectionStep_exceedsTargetAndGetsFurtherAway
...
+                -- This should never happen: we should always be able to select
+                -- *something* that matches.
+                monitor $ counterexample "Error: unable to select any entry"
+                assert False
+            Just SelectionState {selected} -> do
+                let output = head $ snd <$> UTxOIndex.toList selected
+                let bundle = view #tokens output
+                case F.toList $ TokenBundle.getAssets bundle of
+                    [a] -> assert $ a == asset
+                    _   -> assert $ not hasSingletonAsset
```
KtorZ, line 887: :+1:

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -843,6 +855,74 @@ prop_runSelectionStep_exceedsTargetAndGetsFurtherAway
...
+                -- This should never happen: we should always be able to select
+                -- *something* that matches.
+                monitor $ counterexample "Error: unable to select any entry"
+                assert False
+            Just SelectionState {selected} -> do
+                let output = head $ snd <$> UTxOIndex.toList selected
+                let bundle = view #tokens output
+                case F.toList $ TokenBundle.getAssets bundle of
+                    [] -> assert hasCoin
+                    _  -> assert $ not hasCoin
```
KtorZ, line 919: :+1: , essentially the same as above with a different selector.

KtorZ (approved):

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
jonathanknowles: > Ok. So, assetAll also contain singleton assets (as stated in a comment above +1).

I did have a debate with myself about making two disjoint sets: `assetsMixed` and `assetsSingleton`. But that might make it harder for callers that don't care if an asset is accompanied by other assets, since they'd have to specify two selection filters to get what they wanted. In the end, it seemed nicer to treat `assetsSingleton` as a subset of `assetsAll`, similar to how `WithAdaOnly` is a subset of `Any`.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
jonathanknowles: > Debatable to have switched to `NonEmpty SelectionFilter` after all upside_down_face

I did indeed hesitate before doing that. On balance, since there doesn't seem to be a use case for calling this function with an empty list of filters, I thought it would be prudent to make it clear in the type signature.

Though I was a bit surprised to find out that there doesn't seem to be a `tailSafe :: NonEmpty a -> Maybe (NonEmpty a)` defined anywhere. (Though maybe I haven't looked hard enough.)

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
jonathanknowles: Yes, indeed!

There might be a nice way to extract out the common functionality into a separate function. Though I guess we could leave that to a future PR. 😅

PR #2484, opened 2021-02-01
Allow constructing change output from _any_ input

KtorZ:
# Issue Number

ADP-345

# Overview

- 312b0b1dce9df7ad47bfb4763ce391c5125b24e1
  :round_pushpin: **allow constructing change output from _any_ input**
    In only a few days, this has proven to be a quite major issue users have been running into. The thing is that we can't really control how the shape of the UTxO evolves.  Quite easily, a user can end up with large quantity of Ada bound to a particular asset. Before this commit, when transacting on other assets, these Ada would've been locked and the user will be granted with a 'cannot_cover_fee' error, despite having plenty of Ada in his/her wallet. However, because of the way we construct change outputs, it is rather safe and easy to also select from any inputs and construct change outputs accordingly.

  As a matter of fact, the change construction is an iterative trial-and-error process. That is, it works in the following steps:

  - From a given set of input, try to construct valid change output to cover for the transaction need
  - If impossible, select another input and try again.

  Selecting an extra input which contains some potentially new assets will result in a completely different change output. However, since we are re-construct the entire change output sets on each iteration, it doesn't matter much. This processus is still imperfect in the sense that there may be some valid subset of the selection which can result in a valid transaction, whereas some other path may not. Yet, such cases should be rare and "edgy" in practice. 

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
@@ -394,9 +394,8 @@ performSelection minCoinValueFor costFor criteria
         state <- runSelection
             selectionLimit extraCoinSource utxoAvailable balanceRequired
         let balanceSelected = fullBalance (selected state) extraCoinSource
-        if balanceRequired `leq` balanceSelected then do
-            let predictedChange = NE.toList $ predictChange (selected state)
-            makeChangeRepeatedly predictedChange state
+        if balanceRequired `leq` balanceSelected then
+            makeChangeRepeatedly state
 
         else
             pure $ Left $ SelectionInsufficient $ SelectionInsufficientError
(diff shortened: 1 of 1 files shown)


lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -492,11 +491,11 @@ performSelection minCoinValueFor costFor criteria
...
     --
     makeChangeRepeatedly
-        :: [Set AssetId]
-        -> SelectionState
+        :: SelectionState
         -> m (Either SelectionError (SelectionResult TokenBundle))
-    makeChangeRepeatedly changeSkeleton s@SelectionState{selected,leftover} = do
+    makeChangeRepeatedly s@SelectionState{selected,leftover} = do
         let inputsSelected = mkInputsSelected selected
+        let changeSkeleton = NE.toList $ predictChange selected
```
KtorZ, line 500: I've moved this here inside the function body because now, it is possible for the predicted change to change on each iteration due to the selection of new multi-asset outputs.

piotr-iohk (approved):
LGTM. 

With the change I am successfully able to send MA tx which was giving me "cannot_cover_fee"  error before despite my balance was enough to cover tx/fee.

jonathanknowles (approved):
Looks fine to me!

A tiny amount of rebasing will be necessary after #2482 merges. (Since `[SelectionFilter]` has changed to `NonEmpty SelectionFilter`.)

KtorZ:
bors merge

please.

PR #2485, opened 2021-02-01
Do not fail with 'WithdrawalNotWorth' when estimating fees

KtorZ:
# Issue Number

ADP-345

# Overview

- a8bcc3cc82092c3c3b9269c2fb4bfefa2c9c9e08
  :round_pushpin: **Add extra check in existing scenario to capture regression regarding external withdrawals**
    This now fails with:

  ```
  403 Forbidden
  {
      "code": "withdrawal_not_worth",
      "message": "I've noticed that you're requesting a withdrawal from an account that is either empty or doesn't have a balance big enough to deserve being withdrawn. I won't proceed with that request."
  }
  ```

  which captures the regression.

- 23a698adf3387937a6f18bcc25ac67d08eb018d0
  :round_pushpin: **do not fail with 'WithdrawalNotWorth' when estimating fees**
    It used to be the case, and was changed when generalizing the management of the withdrawal in the API handlers. This commit re-introduces it.

# Comments

cc @nikolaglumac

NOTE: After discussion with the Daedalus team, we are no longer sure whether this is indeed a desired behavior. They'll discuss the matter and come back to us to know whether we should close or merge this PR.

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
@@ -2540,6 +2540,14 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
                     (`shouldBe` Quantity 0)
                 ]
 
+        -- Even though there are no rewards left, it should be possible to
+        -- estimate the cost of a withdrawal.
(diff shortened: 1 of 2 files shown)


KtorZ:
As discussed: we'll keep the new behavior.

Issue #2486, opened 2021-02-02
Flaky test: "Wallet with wrong wallet Id cannot be got"


jonathanknowles:
# Context
 
https://hydra.iohk.io/build/5479393/nixlog/1/tail

# Test Case

`/Cardano.Wallet/WalletLayer works as expected/Wallet with wrong wallet Id cannot be got/`

# Failure / Counter-example

```
Failures:

  test/unit/Cardano/WalletSpec.hs:408:5: 
  1) Cardano.Wallet, WalletLayer works as expected, Wallet with wrong wallet Id cannot be got
       Falsifiable (after 1 test):
         ((WalletId {getWalletId = b0e3b2a1101b2f476ca5f7a58c910c3d2947153c},WalletName {getWalletName = "My Wallet"},DummyState (fromList [])),WalletId {getWalletId = b0e3b2a1101b2f476ca5f7a58c910c3d2947153c})
       predicate failed on: Right (Wallet {utxo = UTxO {getUTxO = fromList []}, currentTip = BlockHeader {slotNo = SlotNo 0, blockHeight = Quantity {getQuantity = 0}, headerHash = Hash "\STX\197(;\170\rM\141D\\\226\&4\199IW\246\153\141\DLE-%\154\223\223\&2\226\229\185\GS\214\nh", parentHeaderHash = Hash "00000000000000000000000000000000"}, getState = DummyState (fromList [])},WalletMetadata {name = WalletName {getWalletName = "My Wallet"}, creationTime = 2021-02-02 07:35:42.642858 UTC, passphraseInfo = Nothing, delegation = WalletDelegation {active = NotDelegating, next = []}},fromList [])

  To rerun use: --match "/Cardano.Wallet/WalletLayer works as expected/Wallet with wrong wallet Id cannot be got/"

Randomized with seed 206660964
```

# Resolution 

Fixed by #2487

PR #2487, opened 2021-02-03
Fix flaky WalletSpec property

rvl:
# Issue Number

Flaky test #2486

# Overview

Enhances assertion to prevent random failures.

### DIFF
--- a/lib/core/test/unit/Cardano/WalletSpec.hs
+++ b/lib/core/test/unit/Cardano/WalletSpec.hs
@@ -402,10 +402,10 @@ walletGetProp newWallet = monadicIO $ liftIO $ do
 walletGetWrongIdProp
     :: ((WalletId, WalletName, DummyState), WalletId)
     -> Property
-walletGetWrongIdProp (newWallet, corruptedWalletId) = monadicIO $ liftIO $ do
+walletGetWrongIdProp (newWallet@(wid, _, _), walletId) = monadicIO $ liftIO $ do
     (WalletLayerFixture _db wl _walletIds _) <- liftIO $ setupFixture newWallet
-    attempt <- runExceptT $ W.readWallet wl corruptedWalletId
-    attempt `shouldSatisfy` isLeft
+    attempt <- runExceptT $ W.readWallet wl walletId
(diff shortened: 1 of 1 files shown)


jonathanknowles (approved):
LGTM.

I'm guessing that this would fail only very rarely, but good to fix nonetheless!

jonathanknowles:
Oops:

```hs
Module ‘Test.QuickCheck’ does not export ‘(/==)’
```
I guess this should be `=/=`.

lib/core/test/unit/Cardano/WalletSpec.hs
```diff
@@ -402,10 +403,10 @@ walletGetProp newWallet = monadicIO $ liftIO $ do
 walletGetWrongIdProp
     :: ((WalletId, WalletName, DummyState), WalletId)
     -> Property
-walletGetWrongIdProp (newWallet, corruptedWalletId) = monadicIO $ liftIO $ do
+walletGetWrongIdProp (newWallet@(wid, _, _), walletId) = monadicIO $ liftIO $ do
     (WalletLayerFixture _db wl _walletIds _) <- liftIO $ setupFixture newWallet
-    attempt <- runExceptT $ W.readWallet wl corruptedWalletId
-    attempt `shouldSatisfy` isLeft
+    attempt <- runExceptT $ W.readWallet wl walletId
+    attempt `shouldSatisfy` if wid /== walletId then isLeft else isRight
```
jonathanknowles: I'm guessing this won't work, as `if` is expecting a `Bool` here, but the result of evaluating `=/=` is a `Property`.

Perhaps:
```suggestion
    attempt `shouldSatisfy` if wid /= walletId then isLeft else isRight
```

rvl: oh yes, start compile & test, submit PR, then forget! fixed now.

rvl:
Thanks.

> I'm guessing that this would fail only very rarely, but good to fix nonetheless!

Well, since it happened in a bors merge, and you opened the issue, and the problem is pretty clear, then here we are...

bors r+

PR #2490, opened 2021-02-04
Return fewer change outputs when the minimum ada quantity cannot be satisfied.

jonathanknowles:
# Issue Number

https://jira.iohk.io/projects/ADP/issues/ADP-696

# Overview

By default, the coin selection algorithm attempts to make exactly one change output for each user-specified output.

But in situations where the supply of ada is limited, this strategy can lead to failure if it is not possible to assign the minimum ada quantities to all change outputs.

This PR relaxes this restriction, so that in situations where ada is limited, we allow the algorithm to return fewer change outputs. 

# Comments

- [x] Adjust coin selection algorithm to allow a reduced number of change outputs when the supply of ada is limited.
- [x] Update property tests to allow a slightly-increased fee in the case when change outputs have been dropped.
- [x] Revises the way we sort change maps within `makeChange` so that:
    - Empty change bundles are placed at the start.
    - Empty change bundles deriving from **_user-specified assets_** and empty change bundles deriving from **_non-user-specified assets_** are combined together, if possible.
- [x] Add property test to check that sorting token bundles by `AssetCount` always places coins before non-coins.
- [ ] Add property test to check that it is possible to send the entire balance to another wallet.
    - [ ] Generalize this property to work with non-zero minimum ada quantities.

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -1212,7 +1212,7 @@ selectionToUnsignedTx sel s =
     UnsignedTx
         (fullyQualifiedInputs $ inputsSelected sel)
         (outputsCovered sel)
-        (fullyQualifiedChange $ NE.toList $ changeGenerated sel)
+        (fullyQualifiedChange $ changeGenerated sel)
   where
     qualifyAddresses
         :: forall a t. (Traversable t)
@@ -1302,11 +1302,15 @@ selectAssetsNoOutputs ctx wid wal tx transform = do
                     (view #tokens (head $ outputsCovered sel))
                     (TokenBundle.fromCoin deposit)
             in
-                once (TokenBundle.add reclaim) (changeGenerated sel)
+                once
+                    (TokenBundle.empty)
+                    (TokenBundle.add reclaim)
+                    (changeGenerated sel)
         }
   where
-    once :: (a -> a) -> NonEmpty a -> NonEmpty a
-    once fn (a :| as) = fn a :| as
+    once :: a -> (a -> a) -> [a] -> [a]
+    once _ f (a : as) = f a : as
+    once a _ [] = [a]
 
 -- | Selects assets from the wallet's UTxO to satisfy the requested outputs in
 -- the given transaction context. In case of success, returns the selection
@@ -1440,7 +1444,7 @@ mkTxMeta
 mkTxMeta ti' blockHeader wState txCtx sel =
     let
         amtOuts = sumCoins $
-            (txOutCoin <$> NE.toList (changeGenerated sel))
+            (txOutCoin <$> changeGenerated sel)
             ++
             mapMaybe ourCoin (outputsCovered sel)
 
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -1287,30 +1287,61 @@ selectAssetsNoOutputs ctx wid wal tx transform = do
         { outputsCovered = mempty
         , changeGenerated =
             let
-                -- NOTE 1: This subtraction and head are safe because of the
+                -- NOTE 1: There are in principle 6 cases we may ran into, which
+                -- can be grouped in 3 groups of 2 cases:
+                --
+                -- (1) When registering a key and delegating
+                -- (2) When delegating
+                -- (3) When de-registering a key
+                --
+                -- For each case, there may be one or zero change output. For
+                -- all 3 cases, we'll treat the case where there's no change
+                -- output as an edge-case and also leave no change. This may be
+                -- in practice more costly than necessary because, by removing
+                -- the fake output, we'd in practice have some more Ada
+                -- available to create a change (and a less expensive
+                -- transaction). Yet, this would require quite some extra logic
+                -- here in addition to all the existing logic inside the
+                -- CoinSelection/MA/RoundRobin module already. If we were not
+                -- able to add a change output already, let's not try to do it
+                -- here. Worse that can be list is:
+                --
+                --     max (minUTxOValue, keyDepositValue)
+                --
+                -- which we'll deem acceptable under the circumstances (that can
+                -- only really happen if one is trying to delegate with already
+                -- a very small Ada balance, so that it's left with no Ada after
+                -- having paid for the delegation certificate. Why would one be
+                -- delegating almost nothing certainly is an edge-case not worth
+                -- considering for too long).
+                --
+                -- However, if a change output has been create, then we want to
+                -- transfer the surplus of value from the change output to that
+                -- change output (which is already safe). That surplus is
+                -- non-null if the `minUTxOValue` protocol parameter is
+                -- non-null, and comes from the fact that the selection
+                -- algorithm automatically assigns this value when presented
+                -- with a null output. In the case of (1), the output's value is
+                -- equal to the stake key deposit value, which may be in
+                -- practice greater than the `minUTxOValue`. In the case of (2)
+                -- and (3), the deposit is null. So it suffices to subtract
+                -- `deposit` to the value of the covered output to get the
+                -- surplus.
+                --
+                -- NOTE 2: This subtraction and head are safe because of the
                 -- invariants enforced by the asset selection algorithm. The
                 -- output list has the exact same length as the input list, and
                 -- outputs are at least as large as the specified outputs.
-                --
-                -- NOTE 2: When presented with 0 Ada outputs, the selection
-                -- algorithm will assign a minimum default value to the output.
-                -- So the output covered may be in practice bigger than the
-                -- output specified. This is the case when 'deposit' is null, in
-                -- which case, we want to make sure to add this extra minimum
-                -- value to the resulting change and not completely discard it.
-                reclaim = TokenBundle.unsafeSubtract
+                surplus = TokenBundle.unsafeSubtract
                     (view #tokens (head $ outputsCovered sel))
                     (TokenBundle.fromCoin deposit)
             in
-                once
-                    (TokenBundle.empty)
-                    (TokenBundle.add reclaim)
-                    (changeGenerated sel)
+                surplus `addToHead` changeGenerated sel
         }
   where
-    once :: a -> (a -> a) -> [a] -> [a]
-    once _ f (a : as) = f a : as
-    once a _ [] = [a]
+    addToHead :: TokenBundle -> [TokenBundle] -> [TokenBundle]
+    addToHead _    [] = []
+    addToHead x (h:q) = TokenBundle.add x h : q
 
 -- | Selects assets from the wallet's UTxO to satisfy the requested outputs in
 -- the given transaction context. In case of success, returns the selection
--- a/lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
@@ -800,7 +800,7 @@ makeChange minCoinValueFor requiredCost mExtraCoinSource inputBundles outputBund
         totalOutputCoinValueIsZero
     | otherwise = do
         (bundles, remainder) <-
-            maybe (Left $ changeError excessCoin change) Right $
+            maybe (Left changeError) Right $
                 excessCoin `subtractCoin` requiredCost
                 >>=
                 runStateT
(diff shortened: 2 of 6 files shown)


lib/core/src/Cardano/Wallet.hs
```diff
@@ -1288,11 +1288,15 @@ selectAssetsNoOutputs ctx wid tx transform = do
...
+                    (TokenBundle.empty)
+                    (TokenBundle.add reclaim)
+                    (changeGenerated sel)
         }
   where
-    once :: (a -> a) -> NonEmpty a -> NonEmpty a
-    once fn (a :| as) = fn a :| as
+    once :: a -> (a -> a) -> [a] -> [a]
+    once _ f (a : as) = f a : as
+    once a _ [] = [a]
```
jonathanknowles: This needs checking.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -508,41 +506,65 @@ performSelection minCoinValueFor costFor criteria
...
+            -- change outputs is fewer than optimal, because the supply of ada
+            -- was insufficient. Try again with more ada to see if it leads to
+            -- an improvement:
+            selectOneEntry s >>= \case
+                Just s' ->
+                    makeChangeRepeatedly s'
+                Nothing ->
+                    -- There is no more ada available. Terminate with a
+                    -- less-than-optimal number of change outputs.
+                    pure $ Right $ mkSelectionResult change
```
KtorZ, line 528: :+1:

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -508,41 +506,65 @@ performSelection minCoinValueFor costFor criteria
...
+        selectOneEntry = selectMatchingQuantity selectionLimit $
+            WithAdaOnly :| [Any]
+
+        SelectionState {selected, leftover} = s
+
+        cost = costFor SelectionSkeleton
+            { inputsSkeleton  = selected
+            , outputsSkeleton = NE.toList outputsToCover
+            , changeSkeleton
+            }
```
KtorZ, line 565: You really dislike let binding, don't you :smile:

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -791,105 +813,105 @@ makeChange
...
+        changeForUserSpecifiedAssets
+        changeForNonUserSpecifiedAssets
+
+    -- Change for all assets, with the minimum ada amounts added.
+    changeForAssetsWithMinimalCoins :: NonEmpty TokenBundle
+    changeForAssetsWithMinimalCoins =
+        assignMinimumCoin minCoinFor <$> changeForAssets
+
+    totalMinimumCoinForAssets =
+        F.foldMap TokenBundle.getCoin changeForAssetsWithMinimalCoins
```
KtorZ: I find this a little confusing now to have these two declarations in the top-level `where` because they are only used for error reporting. That's why I originally put `totalMinCoinValue` under `changeError`. IMO, the nicest feature about `where` is that it can be nested to be of limited scope and local to where it is used, which helps a lot with readability. Having a top-level flat list of local bindings reduces readability.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -914,6 +936,106 @@ makeChange minCoinValueFor requiredCost mExtraCoinSource inputBundles outputBund
...
+            -- Partition any remaining ada according to the weighted
+            -- distribution of output coins that remain in our list:
+            let outputCoinsRemaining = snd <$> (pair :| pairs) in
+            let bundlesForOutputCoins = TokenBundle.fromCoin <$>
+                    makeChangeForCoin outputCoinsRemaining adaRemaining in
+            -- Finally, combine the minimal coin asset bundles with the
+            -- bundles obtained by partitioning the remaining ada amount:
+            Just $ NE.toList $ NE.zipWith (<>)
+                bundlesForAssetsWithMinimumCoins
+                bundlesForOutputCoins
```
KtorZ, line 1013: This case is the "normal" case as we had before.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -914,6 +936,106 @@ makeChange minCoinValueFor requiredCost mExtraCoinSource inputBundles outputBund
...
+    -- ^ A list of pre-computed asset change maps paired with original output
+    -- coins, sorted into an order that ensures all empty token maps are at the
+    -- start of the list.
+    -> Maybe [TokenBundle]
+    -- ^ Resulting change bundles, or 'Nothing' if there was not enough ada
+    -- available to assign a minimum ada quantity to all non-empty token maps.
+assignCoinsToChangeMaps adaAvailable minCoinFor pairsAtStart =
+    loop adaRequiredAtStart pairsAtStart
+  where
+    loop !adaRequired !pairsNonEmpty = case pairsNonEmpty of
```
KtorZ, line 992: Isn't it a bit confusing that `adaRequired` comes as an argument here? I'd expect whatever branch needs it to re-calculate it in its context. That's the case for the first branch only.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -914,6 +936,106 @@ makeChange minCoinValueFor requiredCost mExtraCoinSource inputBundles outputBund
...
+                bundlesForAssetsWithMinimumCoins
+                bundlesForOutputCoins
+
+        (m, _) :| (p : ps) | TokenMap.isEmpty m && adaAvailable < adaRequired ->
+            -- We don't have enough ada available to pay for the minimum
+            -- required amount of every asset map, but we do have an empty
+            -- asset map that is safe to drop. This will reduce the amount of
+            -- ada required by a small amount:
+            let adaRequired' = adaRequired `Coin.distance` minCoinFor m in
+            loop adaRequired' (p :| ps)
```
KtorZ, line 1021: :+1:

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -914,6 +936,106 @@ makeChange minCoinValueFor requiredCost mExtraCoinSource inputBundles outputBund
...
+
+        (m, _) :| [] | TokenMap.isEmpty m && adaAvailable >= adaRequired ->
+            -- We didn't have any non-ada assets at all in our change. We can
+            -- simply return the available ada amount as a singleton change
+            -- bundle:
+            Just [TokenBundle.fromCoin adaAvailable]
```
KtorZ: Isn't that branch a special case of / redundant with the first branch? The first branch will always match before this one, and I'd indeed expect the result to be the same.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -914,6 +936,106 @@ makeChange minCoinValueFor requiredCost mExtraCoinSource inputBundles outputBund
...
+
+        (m, _) :| [] | TokenMap.isEmpty m && adaAvailable < adaRequired ->
+            -- We didn't have any non-ada assets at all in our change, but we
+            -- also don't have enough ada available to pay even for a single
+            -- change output. We just burn the available ada amount (which
+            -- will be small), returning no change.
+            Just []
```
KtorZ, line 1028: Yes. I like that is guarded by `adaAvailable < adaRequired`. Although I'd really prefer that the `adaRequired` is actually computed within the function loop itself, and not provided as argument. It makes reasoning and following the steps much easier and it becomes clear that `adaRequired` is necessarily small.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
KtorZ: (just a self-note)

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
jonathanknowles: Hi @KtorZ Good point!

I originally moved this up to a higher scope as I wanted to reuse it from another place. However, I eventually changed my implementation so that this was no longer necessary.

I'm happy to move this back under the `where` clause (where it was before).

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
jonathanknowles: > Isn't it a bit confusing that adaRequired comes as an argument here?

We could re-calculate it on every iteration, but presumably that would make it necessary to compute the minimum values across all remaining token maps in the list, which would lead to quadratic run-time in the length of the list, unless you're thinking of something else?

Here I've opted to compute the total requirement up front, and then just reduce it on each iteration by the correct amount (which should be equal to the minimum ada value of the map that we drop).

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
jonathanknowles: Fixed in 3ca8552436fd1ccc23e1256f27e34032d7adbae6.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
jonathanknowles: I think you're right. I've fixed this in 427e208bfddbe07794927f2e3855761b530196cf.

lib/core/src/Cardano/Wallet.hs
rvl: It seems a bit unnatural to have `changeGenerated = [mempty]`. Why not `[]`?

rvl (commented):
Looks pretty good so far!

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
jonathanknowles: Hi @KtorZ 

I've written a justification for this in https://github.com/input-output-hk/cardano-wallet/pull/2490#discussion_r570169560. Would be interested to know your thoughts on that!

rvl:
A regression test might be useful here.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -533,33 +579,52 @@ prop_performSelection minCoinValueFor costFor (Blind criteria) coverage =
...
+            , pretty absoluteMinCoinValue
+            , "actual coin delta:"
+            , pretty (TokenBundle.getCoin delta)
+            , "number of outputs:"
+            , pretty (length outputsCovered)
             ]
-        let delta = TokenBundle.unsafeSubtract
-                balanceSelected
-                (balanceRequired <> balanceChange)
+        assert True
```
KtorZ: :thinking: ?

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -533,33 +579,52 @@ prop_performSelection minCoinValueFor costFor (Blind criteria) coverage =
...
         assert $ balanceSufficient criteria
         assert $ on (==) (view #tokens)
             balanceSelected (balanceRequired <> balanceChange)
-        assert $ TokenBundle.getCoin delta == expectedCost
+        assert $ TokenBundle.getCoin delta >= expectedCost
+        assert $ TokenBundle.getCoin delta <= maximumExpectedDelta
```
KtorZ, line 604: :+1:

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -533,33 +579,52 @@ prop_performSelection minCoinValueFor costFor (Blind criteria) coverage =
...
                 assert True
       where
+        absoluteMinCoinValue = mkMinCoinValueFor minCoinValueFor TokenMap.empty
+        delta = TokenBundle.unsafeSubtract
+            balanceSelected
+            (balanceRequired <> balanceChange)
+        maximumExpectedDelta =
+            expectedCost `addCoin`
+            (absoluteMinCoinValue `multiplyCoin`
+                (length outputsCovered - length changeGenerated))
```
KtorZ, line 623: I'd expect the maximum delta to actually be: `expectedCost + absoluteMinCoinValue` no matter what. Did QC find counter-examples of that already :thinking: ?

KtorZ (approved):
Looks good. Thanks Jonathan :+1:

nikolaglumac:
@KtorZ @jonathanknowles please ping me once this is merged 🙏

KtorZ:
:+1: Jonathan is on holidays for the next two weeks, I'll wrap this up.

lib/core/src/Cardano/Wallet.hs
KtorZ: This is wrong indeed. The name of the function was maybe confusing. This is what this piece of code is intended to do:

```hs
    reclaim `addToHead` changeGenerated sel
  where
    addToHead :: TokenBundle -> [TokenBundle] -> [TokenBundle]
    addToHead x []    = [x]
    addToHead x (h:q) = TokenBundle.add x h : q
```

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
KtorZ: After checking and fiddling around with that, there are indeed cases where may not be able to create _any_ change output, in which case this property is indeed verified.

KtorZ:
bors merge

KtorZ:
bors merge

KtorZ:
@nikolaglumac :point_up:

PR #2489, opened 2021-02-04
Add missing APIs for multi-asset

rvl:
## Issue Number

ADP-698

## Overview

- Adds assets to the responses of the coin selections endpoints.
- Adds assets to the `/byron-wallets` endpoints.

PR #2492, opened 2021-02-04
API: get and list only assets associated with the wallet

rvl:
### Issue Number

ADP-603 / ADP-604

### Overview

- The list assets endpoint now returns all assets of the wallet, not just available assets.
- The get asset endpoint checks that the asset id really is associated with the wallet.

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
@@ -863,27 +863,18 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
 
     it "TRANS_ASSETS_GET_02 - Asset not present when isn't associated" $ \ctx -> runResourceT $ do
         wal <- fixtureMultiAssetWallet ctx
-        let polId = TokenPolicy.UnsafeTokenPolicyId $ Hash $ B8.replicate 56 '1'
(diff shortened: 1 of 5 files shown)


piotr-iohk (approved):
LGTM.

I wonder if we should also return `404` on list assets when there are no assets associated. Currently it returns just `[]`.

Note to self: check this on Byron wallets once #2489 is merged.

KtorZ:
> I wonder if we should also return 404 on list assets when there are no assets associated. Currently it returns just [].

It should return 404 when the wallet id isn't found. If the wallet is found but there are no assets, then `[]` is sound.

piotr-iohk:
> It should return 404 when the wallet id isn't found. If the wallet is found but there are no assets, then `[]` is sound.

Yeah, but get assets returns in `404` in such case:

```
{"code":"asset_not_present","message":"The requested asset is not associated with this wallet."}
```

So I was thinking it would be nice consistency if list wallet returned something similar... but maybe `[]` is fine indeed.

rvl:
To me, the 404 response is an error which means that the request path was referring to a resource that doesn't exist.
So, when trying to look up an AssetId which is not present, 404 is the correct response.
But when looking at the `/wallets/<wid>/assets` collection, that resource exists, even if it could be empty. Same goes with the transactions list.

rvl:
Thanks for the review and noticing the issue.

bors r+

PR #2496, opened 2021-02-05
Token Metadata Server integration

Anviking:
# Issue Number

ADP-688

# Overview

- Metadata client for input-output-hk/metadata-server#1
- New types to represent metadata according to CIP draft
- JSON parsers
- unit tests 
- plumbing of client into listAssets and getAssets
- mock metadata server

# Comments

I presume implementation details will change (e.g. metadata per asset vs metadata per policy).

### Todo

- [x] Start mock server for integration tests
- [x] add some metadata assertions to getAsset and listAsset.
- [x] more unit tests hmmm
- [x] manually try logged requests against deployed metadata server
     ```
     curl -i -H "Content-type: application/json" --data '{"subjects":["7f71940915ea5fe85e840f843c929eba467e6f050475bad1f10b9c274d1888c0"],"properties":["name","description", "asttd"]}' **https://metadata.cardano-testnet.iohkdev.io/metadata/query
     ```

### Next PR
- Unit test `fillMetadata`
- Manually test with testnet metadata server (fixme: paste details)
- More integration tests perhaps

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types.hs
@@ -143,7 +143,7 @@ module Cardano.Wallet.Primitive.Types
     , defaultSettings
     , unsafeToPMS
 
-    , TokenMetadataServerURI (..)
+    , TokenMetadataServer (..)
 
     -- * InternalState
     , InternalState (..)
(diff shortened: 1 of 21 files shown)


rvl:
I had a look at the metadata-server implementation and made a skeleton client accordingly, code pushed.

> I presume implementation details will change (e.g. metadata per asset vs metadata per policy)

For MVP, it will be metadata per asset. The subject is the hex-encoded asset id.

> Not sure how we'd integration test this down the line. We probably don't want to re-implement the server just for the sake of it.

A mock server would be quite doable. We only need to use one endpoint.

Anviking:
> A mock server would be quite doable. We only need to use one endpoint.

It wouldn't necessarily give us confidence that it matches the real-world implementation. Wouldn't just a static JSON file suffice?

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -1355,6 +1357,31 @@ distance a b =
...
+                    throwE (TextDecodingError
+                        "Not a valid URI scheme, only http/https is supported.")
+                | isNothing uriAuthority ->
+                    throwE
+                        (TextDecodingError "URI must contain a domain part.")
+                | not ((uriPath == "" || uriPath == "/")
+                && uriQuery == "" && uriFragment == "") ->
+                    throwE
+                        (TextDecodingError "URI must not contain a path/query/fragment.")
+            _ -> pure $ TokenMetadataServerURI uri'
```
Anviking: Could be de-duplicated

rvl: Done

Anviking:
bors try

rvl:
I put back metadata subject = AssetId.

AssetId ( = PolicyId + AssetName ) has to be the metadata subject.
Because AssetName is not the name of the asset, it's not necessarily self-describing.
Therefore we need the metadata subject to be AssetId, so that assets can have proper names (and descriptions, and abbreviations, and logos, etc.)
If your asset name is empty (the common case), then AssetId = PolicyId.

rvl (approved):

rvl:
bors try

Anviking:
lgtm

rvl:
Rebased and squashed...

bors r+

PR #2502, opened 2021-02-09
Rework checkpoint reading from database to avoid quadratic queries with number of assets

KtorZ:
# Issue Number

ADP-627

# Overview

Before this commit, every asset will trigger a database query. For, a UTxO of ~1000 entries, with 1 asset on each entry would generate an extra 1000 queries! This was noticeable from benchmarks with large query times on the read utxo benchmarks.

  | Bench                                                | Before   | After    |
  | ---                                                  | ---      | ---      |
  | UTxO (Read)/1 CP (ada-only) x 10000 UTxO             | 441.8 ms | 104.3 ms |
  | UTxO (Read)/1 CP (1 assets per output) x 10000 UTxO  | 12.36 s  | 183.5 ms |
  | UTxO (Read)/1 CP (2 assets per output) x 10000 UTxO  | 29.15 s  | 267.3 ms |
  | UTxO (Read)/1 CP (10 assets per output) x 10000 UTxO | 144.3 s  | 1.132 s  |
  | UTxO (Read)/1 CP (20 assets per output) x 10000 UTxO | /        | 2.242 s  |

# Comments

Normally, I would have solved this with a nice JOIN query... but joining with `persistent` is unpractical (especially because we don't have any intermediate model to unserialize to..). So, doing the merge in-memory since _anyway_, the whole thing is going to end up in memory eventually.

### DIFF
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
@@ -123,6 +123,8 @@ import Cardano.Wallet.Primitive.Slotting
     , interpretQuery
     , slotToUTCTime
     )
+import Cardano.Wallet.Primitive.Types.TokenBundle
+    ( TokenBundle )
 import Cardano.Wallet.Primitive.Types.TokenMap
     ( AssetId (..) )
 import Control.Monad
@@ -226,6 +228,7 @@ import qualified Cardano.Wallet.Primitive.Types.TokenBundle as TokenBundle
 import qualified Cardano.Wallet.Primitive.Types.Tx as W
 import qualified Cardano.Wallet.Primitive.Types.UTxO as W
 import qualified Data.List as L
+import qualified Data.Map.Merge.Strict as Map
 import qualified Data.Map.Strict as Map
 import qualified Data.Text as T
 import qualified Database.Sqlite as Sqlite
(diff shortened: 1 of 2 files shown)


KtorZ:
bors try

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1564,16 +1567,32 @@ checkpointFromEntity cp utxo =
...
-        | (UTxO _ _ (TxId input) ix addr coin, tokens) <- utxo
+    utxo = W.UTxO $ Map.merge
+        (Map.mapMissing (const mkFromCoin)) -- No assets, only coins
+        (Map.dropMissing) -- Only assets, impossible.
+        (Map.zipWithMatched (const mkFromBoth)) -- Both assets and coins
+        (Map.fromList
+            [ (W.TxIn input ix, (addr, coin))
+            | (UTxO _ _ (TxId input) ix addr coin) <- coins
+            ])
+        (Map.fromListWith TokenBundle.add
```
Anviking, line 1578: Why `Map.fromListWith TokenBundle.add`, but `Map.fromList` for ada?

Anviking: Ah —  DB `UTxO` is specific to one asset, so there can be multiple ones per input, which needs to be merged.

Anviking (approved):
Nice. lgtm

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
KtorZ: Precisely :+1:

PR #2503, opened 2021-02-10
Recover from partial cleanupCheckpointTable db migration

rvl:
### Issue Number

ADP-700

### Overview

- Test case to reproduce the migration error "Database migration: manual intervention required." (needed manual fiddling of the database - we could not reproduce it otherwise)
- Adjust the field checks in the checkpoint cleanup migration, so that it can recover from the situation (most of the time).


lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -430,16 +433,16 @@ migrateManually tr proxy defaultFieldValues =
...
-
-                -- 1. Add genesis_hash and genesis_start to the 'wallet' table.
                 [defaults] <- runSql conn $ select ["genesis_hash", "genesis_start"] orig
                 let [PersistText genesisHash, PersistText genesisStart] = defaults
                 addColumn_ conn True (DBField WalGenesisHash) (quotes genesisHash)
                 addColumn_ conn True (DBField WalGenesisStart) (quotes genesisStart)
 
-                -- 2. Drop columns from the 'checkpoint' table
+        -- 2. Drop columns from the 'checkpoint' table
+        isFieldPresentByName conn "checkpoint" "genesis_hash" >>= \case
```
Anviking, line 442: So if the checkpoint table contains the `genesis_hash`, we do the below migration...? Won't this always be the case? Particularly after the step above?

KtorZ: I got confused by that at first too. Yet, it should work fine because we are actually with two tables here:

- The wallet table
- The checkpoint table

That migration does two things:

1. It adds the genesis hash and genesis start to the _wallet_ table. 
2. It removes all the protocol parameters from the checkpoint table. 

Thus, once removed from the checkpoint table `isFieldPresentByName conn "checkpoint" "genesis_hash"` would prevent the migration from re-running again.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -448,12 +451,15 @@ migrateManually tr proxy defaultFieldValues =
                             , "slot_length", "epoch_length", "tx_max_size"
                             , "epoch_stability", "active_slot_coeff"
                             ]
+                _ <- runSql conn $ dropTable tmp
```
KtorZ, line 454: :+1:

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -430,16 +433,16 @@ migrateManually tr proxy defaultFieldValues =
                 traceWith tr $ MsgManualMigrationNotNeeded field
 
             ColumnMissing -> do
-                let orig = "checkpoint"
-                let tmp = orig <> "_tmp"
-
-                -- 1. Add genesis_hash and genesis_start to the 'wallet' table.
                 [defaults] <- runSql conn $ select ["genesis_hash", "genesis_start"] orig
```
KtorZ, line 436: Implicitly, this assumes that is the `genesis_hash` field does not exists in the `wallet` table, then it necessarily exists in the checkpoint table. This should be true for quite many releases back and indeed, works fine with the way the migration is written :+1:

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -479,7 +485,7 @@ migrateManually tr proxy defaultFieldValues =
             ]
 
         dropTable table = mconcat
-            [ "DROP TABLE " <> table <> ";"
+            [ "DROP TABLE IF EXISTS " <> table <> ";"
```
KtorZ, line 488: :+1:

KtorZ (approved):
Good catch.

KtorZ (commented):
bors merge

PR #2505, opened 2021-02-10
Add TOKEN_METADATA_SERVER variable for local-cluster and more properties

rvl:
# Issue Number

ADP-688

# Overview

- Adds TOKEN_METADATA_SERVER environment variable for the local test cluster
- ~Adds parsing of the other metadata properties - in such a way that parse failures of a single field should not cause a parse failure of the entire batch.~ ⇒ moved to #2507

### DIFF
--- a/lib/shelley/exe/local-cluster.hs
+++ b/lib/shelley/exe/local-cluster.hs
@@ -53,6 +53,7 @@ import Cardano.Wallet.Shelley.Launch.Cluster
     , sendFaucetAssetsTo
     , sendFaucetFundsTo
     , testMinSeverityFromEnv
+    , tokenMetadataServerFromEnv
     , walletListenFromEnv
     , walletMinSeverityFromEnv
     , withCluster
@@ -201,6 +202,9 @@ import qualified Data.Text as T
 --     "virtual hard forks" in the node config files.
 --     The final era can be changed with this variable.
 --
+-- - TOKEN_METADATA_SERVER  (default: none)
+--     Use this URL for the token metadata server.
+--
 -- - NO_POOLS  (default: stake pools nodes are started and registered)
 --     If set, the cluster will only start a BFT leader and a relay, no
 --     stake pools. This can be used for running test scenarios which do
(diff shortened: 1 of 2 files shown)


lib/shelley/exe/local-cluster.hs
```diff
@@ -201,6 +202,9 @@ import qualified Data.Text as T
 --     "virtual hard forks" in the node config files.
 --     The final era can be changed with this variable.
 --
+-- - TOKEN_METADATA_SERVER  (default: none)
+--     Use this URL for the token metadata server.
```
Anviking, line 206: 👍 

Not important now but:
- _Maybe_ worth including in the integration tests as well. IIRC all the other ENV vars are the same.
- One idea could be to allow file://path/to/000000000000.json as a special case, which would cause the wallet to simply read the json instead of sending a POST request. Maybe could be useful for testing locally.

lib/core/src/Cardano/Wallet/TokenMetadata.hs
```diff
@@ -182,21 +186,18 @@ newtype BatchResponse = BatchResponse
 data SubjectProperties = SubjectProperties
     { subject :: Subject
     , owner :: Maybe Signature
-    -- TODO: use Data.SOP.NP and parameterize type by property names
-    -- Name and description are required, both others may be missing the
-    -- response.
     , properties :: ( Property "name"
-                    -- , (PropertyValue "acronym", [Signature])
```
Anviking: Looking at https://input-output.atlassian.net/wiki/spaces/AD/pages/2137456733/Token+Metadata+Well-Known+Properties now this seems right, modulo:
1. "unit" is here disabled
2. It seems that spec fundamentally is different from the current server implementation — property values are not wrapped in { "value": ., "anSignatures": [] } there… 🤔

Anviking (approved):

KtorZ (approved):

lib/core/src/Cardano/Wallet/TokenMetadata.hs
rvl: Yes

lib/shelley/exe/local-cluster.hs
rvl: 1. Yes good idea - we can sort it later.
2. Also a good idea, but perhaps a better idea would be to see how possible it is to run the _real_ metadata-server with the test json file.

PR #2506, opened 2021-02-10
Allow individual JSON entries to fail parsing

Anviking:
# Issue Number

ADP-688

# Overview

- [x] Failing test case demonstrating problem
- [x] Fix using `<|>`

# Comments

- I haven't checked out #2505 yet.
- The mock server uses the same JSON instances when serving, which got a bit interesting when I tried including the faulty entry in the existing golden file, used by the other unit tests.

### DIFF
--- a/lib/core/src/Cardano/Wallet/TokenMetadata.hs
+++ b/lib/core/src/Cardano/Wallet/TokenMetadata.hs
@@ -68,8 +68,10 @@ import Cardano.Wallet.Primitive.Types.TokenMap
     ( AssetId (..) )
 import Cardano.Wallet.Primitive.Types.TokenPolicy
     ( AssetMetadata (..), TokenName (..), TokenPolicyId (..) )
+import Control.Applicative
+    ( (<|>) )
 import Control.Monad
-    ( when )
+    ( mapM, when )
 import Control.Tracer
     ( Tracer, contramap, traceWith )
 import Data.Aeson
@@ -97,7 +99,7 @@ import Data.Functor
 import Data.Hashable
     ( Hashable )
 import Data.Maybe
-    ( mapMaybe )
+    ( catMaybes, mapMaybe )
 import Data.String
     ( IsString (..) )
 import Data.Text
@@ -159,6 +161,7 @@ fillMetadata client assets f = do
         Left _e -> do
             -- TODO: Trace error?
             return $ fmap (f Nothing) assets
+
 
 {-------------------------------------------------------------------------------
                             Cardano Metadata Server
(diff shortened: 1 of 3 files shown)


lib/core/test/data/Cardano/Wallet/TokenMetadata/golden2.json
```diff
@@ -0,0 +1,83 @@
...
+              "publicKey": "8cc8963b65ddd0a49f7ce1acc2915d8baff505bbc4f8727a22bd1d28f8ad6632"
+            }
+          ]
+        }
+      },
+      {
+        "subject": "missing sigs",
+        "name": {
+          "value": "Token1",
+          "anSignatures": []
```
KtorZ, line 65: Is that supposed to be a bad entry ?

KtorZ:
```hs
        <*> o .:? "anSignatures" .!= []
```

Okay, so if the `anSignatures` is present, we expect it to be empty, but it can still be present. I think it's a bit of a harsh constraints and could bite us depending on how the final server implements it. We should allow empty signature lists as well.

lib/core/src/Cardano/Wallet/TokenMetadata.hs
```diff
@@ -450,8 +453,12 @@ instance FromJSON PropertyName where
...
 
 instance FromJSON BatchResponse where
-    parseJSON =  withObject "BatchResponse" $ \o ->
-        BatchResponse <$> o .: "subjects"
+    parseJSON =  withObject "BatchResponse" $ \o -> do
+        (a :: Value) <- o .: "subjects"
+        let maybeParseItem v = fmap Just (parseJSON v) <|> pure Nothing
+        BatchResponse <$> withArray "items"
+            (fmap catMaybes . mapM maybeParseItem . toList)
+            a
```
KtorZ: We could avoid the `toList` traversal here by parsing it as a list of Value in the first place:

```hs
parseJSON =  withObject "BatchResponse" $ \o -> do
    (xs :: [Value]) <- o .: "subjects"
    let maybeParseItem v = fmap Just (parseJSON v) <|> pure Nothing
    BatchResponse <$> (catMaybes <$> mapM maybeParseItem xs)
```

lib/core/test/data/Cardano/Wallet/TokenMetadata/golden2.json
KtorZ: My bad, actually `.!=` is used to provide a default value. So providing an empty list works just fine :+1:

KtorZ (approved):

KtorZ:
bors merge

PR #2508, opened 2021-02-10
Introduce separate shared state

paweljakubas:
# Issue Number

adp-685

# Overview

- [x] I have introduced SharedState
- [x] I have added impl of newSharedState and extendSharedState
- [x] I have added updating shared state functions 
- [x] I have added isShared
- [x] I have added property tests
- [x] I have removed script state from SeqState and cleaned unit testing, sqlite
- [x] Add design docs in specification dir
- [x] I have dealt with migration of the removal

# Comments


lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,217 @@
...
+    { accountKey :: k 'AccountK XPub
+        -- ^ Reward account public key associated with this wallet
+    , derivationPrefix :: DerivationPrefix
+        -- ^ Derivation path prefix from a root key up to the account key
+    , derivationKeyNumber :: KeyNumberScope
+        -- ^ Number of first keys that are used to produce candidate addresses
+    , paymentScriptTemplate :: !(Maybe ScriptTemplate)
+        -- ^ Script template together with a map of account keys and cosigners
+        -- for payment credential
+    , stakingScriptTemplate :: !(Maybe ScriptTemplate)
```
KtorZ: staking -> delegation 

:grimacing:

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,217 @@
...
+    ( Eq (k 'AccountK XPub)
+    ) => Eq (SharedState n k)
+
+instance
+    ( NFData (k 'AccountK XPub)
+    )
+    => NFData (SharedState n k)
+
+newtype KeyNumberScope =
+    KeyNumberScope { unKeyNumberScope :: Word8 }
```
KtorZ: I am not sure whether I understand the purpose of the `KeyNumberScope` fully yet, but a `Word8` seems small?

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,217 @@
...
+    => k 'AccountK XPub
+    -> Index 'Hardened 'AccountK
+    -> Maybe ScriptTemplate
+    -> Maybe ScriptTemplate
+    -> KeyNumberScope
+    -> Maybe (SharedState n k)
+mkSharedState accXPub accIx spendingTemplate stakingTemplate keyNum =
+    let
+        prefix =
+            DerivationPrefix ( purposeBIP44, coinTypeAda, accIx )
```
KtorZ: The purpose should be specific to this type of wallet. We have yet to define it with the corresponding CIP. But it won't be 44' nor 1852'

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,217 @@
...
+-- | A state to keep track of script templates, account public keys for cosigners,
+-- | and a combination of possible script addresses to look for in a ledger
+data SharedState (n :: NetworkDiscriminant) k = SharedState
+    { accountKey :: k 'AccountK XPub
+        -- ^ Reward account public key associated with this wallet
+    , derivationPrefix :: DerivationPrefix
+        -- ^ Derivation path prefix from a root key up to the account key
+    , derivationKeyNumber :: KeyNumberScope
+        -- ^ Number of first keys that are used to produce candidate addresses
+    , paymentScriptTemplate :: !(Maybe ScriptTemplate)
```
KtorZ: Why `Maybe` here? This ought to be obligatory right?

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,217 @@
...
+    -> Maybe ScriptTemplate
+    -> KeyNumberScope
+    -> Maybe (SharedState n k)
+mkSharedState accXPub accIx spendingTemplate stakingTemplate keyNum =
+    let
+        prefix =
+            DerivationPrefix ( purposeBIP44, coinTypeAda, accIx )
+        accXPub' = liftXPub $ getRawKey accXPub
+        rewardXPub =
+            getKey $ deriveAddressPublicKey accXPub' Stake minBound
```
KtorZ: There's actually no rewardXPub, the delegation part is a script and is given by the delegation script template!

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,217 @@
...
+                else
+                    []
+            _ -> undefined
+    in
+        if all isNothing [spendingTemplate, stakingTemplate] then
+            Nothing
+        else
+            Just $ SharedState accXPub prefix keyNum spendingTemplate stakingTemplate addressesToFollow
+
+replaceCosignersWithVerKeys
```
KtorZ: suggestion: `instantiateScript` ?

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,217 @@
...
+    :: ScriptTemplate
+    -> Map Cosigner Word8
```
KtorZ: I am not sure to get this type? The derivation index should be the same for all co-signers (by design).

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,217 @@
...
+    -> Maybe (Script KeyHash)
```
KtorZ: It'd be okay-ish to make this ` Script KeyHash` only. The only way to get `Nothing` here is if trying to derive an index that is out of bounds. This cannot happen in the lifetime of a wallet in principle, and if it ever happens, there's no way to recover from it. So this is better kept captured as an invariant / exception.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,217 @@
...
+    :: ScriptTemplate
+    -> XPub
+    -> KeyNumberScope
+    -> [Address]
+generateAddressCombination st@(ScriptTemplate xpubs _) stakeXPub (KeyNumberScope num) =
+    concatMap tryCreateAddress cosignerCombinations
+  where
+    cosigners' = Map.keys xpubs
+    formMaps = map Map.fromList . zipWith zip (cycle [cosigners'])
+    cosignerCombinations = case Map.size xpubs of
```
KtorZ: I am not sure to understand what this function does :thinking:  ?

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,313 @@
...
+{-------------------------------------------------------------------------------
+                                 State
+-------------------------------------------------------------------------------}
+
+-- | A state to keep track of script templates, account public keys of other cosigners,
+-- | verification keys used and unused fitting in the address pool gap,
+-- | and script addresses unused/used with the corresponding indices.
+data SharedState (n :: NetworkDiscriminant) k = SharedState
+    { shareStateAccountKey :: k 'AccountK XPub
+        -- ^ Reward account public key associated with this wallet
```
KtorZ: This shouldn't be there; the reward account will be defined by the delegation template.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,313 @@
...
+    , shareStateDerivationPrefix :: !DerivationPrefix
+        -- ^ Derivation path prefix from a root key up to the account key
+    , shareStateGap :: !AddressPoolGap
+        -- ^ Number of first keys that are used to produce candidate addresses
```
KtorZ: The comment is slightly misleading. The address gap represents the number of consecutive unused addresses available at all time.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,313 @@
...
+        -- ^ Script template together with a map of account keys and cosigners
+        -- for payment credential
+    , shareStateDelegationTemplate :: !(Maybe ScriptTemplate)
+        -- ^ Script template together with a map of account keys and cosigners
+        -- for staking credential. If not specified then the same template as for
+        -- payment is used
+    , shareStateOurAddresses :: !(Map Address (Index 'Soft 'ScriptK, AddressState))
+        -- ^ Our addresses meaning the addresses containing our verification key hashes
+        -- (if discovered) and represented here by corresponding indices of verification
+        -- keys present in the given address.
```
KtorZ: We'll likely need something more like we have in `AddressPool`:

```
        :: !(Map
                (KeyFingerprint "payment" key)
                (Index 'Soft 'AddressK, AddressState)
            )
```

because, an address belongs to the wallet only if the payment part is ours. The delegation part may be totally different. 

Shall we consider reviewing a bit `AddressPool` to allow for re-use here :thinking: ? The code exists and is battle-tested, so it may be good to not redo all that non-trivial logic.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,313 @@
...
+    -> ScriptTemplate
+    -> Maybe ScriptTemplate
+    -> SharedState n k
+newSharedState accXPub accIx g pTemplate dTemplate =
+    extendSharedState (toEnum 0) $
+    unsafeSharedState accXPub accIx g pTemplate dTemplate Map.empty
+
+replaceCosignersWithVerKeys
+    :: ScriptTemplate
+    -> Int
```
KtorZ: We can do better than `Int` here :) --> `Index 'Soft 'ScriptK` sounds like a good candidate.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,313 @@
...
+    -> Index 'Soft 'ScriptK
+    -> Address
+constructAddressFromIx tag pTemplate dTemplate ix =
+    let delegationCredential = DelegationFromScript . toScriptHash
+        paymentCredential = PaymentFromScript . toScriptHash
+        createAddress pScript' dScript' =
+            CA.unAddress $
+            delegationAddress tag
+            (paymentCredential pScript') (delegationCredential dScript')
+        dTemplate' = fromMaybe pTemplate dTemplate
```
KtorZ: This sounds incorrect. If there's no delegation template, then the address should not include any delegation part. Defaulting to using payment credentials as delegation credentials should be done in the API, not here.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,313 @@
...
+        Just dTemplate' ->
+            isRight (validateScriptTemplate RequiredValidation dTemplate')
+
+keyHashFromAccXPubIx
+    :: (SoftDerivation k, WalletKey k)
+    => k 'AccountK XPub
+    -> Index 'Soft 'ScriptK
+    -> KeyHash
+keyHashFromAccXPubIx accXPub ix =
+    hashVerificationKey $ deriveVerificationKey accXPub ix
```
KtorZ: Careful that keys on the payment template are derived with role=3, but keys on the delegation template are derive with role=4. I'd suggest to make the role explicit here to avoid shooting ourselves in the foot.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,313 @@
...
+retrieveAllCosigners :: Script Cosigner -> [Cosigner]
+retrieveAllCosigners = foldScript (:) []
+
+isShared
+    :: (SoftDerivation k, WalletKey k)
+    => Address
+    -> SharedState n k
+    -> (Maybe (Index 'Soft 'ScriptK, KeyHash), SharedState n k)
+isShared address shared@(SharedState !accXPub !prefix !g !pT !dT !ourAddresses) =
+    case Map.lookup address ourAddresses of
```
KtorZ: This won't work unfortunately :( ... For the reason stated in another comment. An address is ours if-and-only-if its payment part is ours. Please, let's strive to re-use code from `AddressPool`. 

We could do some pair-programming session to see how to abstract over the existing AddressPool if you want :+1:

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
paweljakubas: corrected this comment and added extended description about shared wallets -> https://github.com/input-output-hk/cardano-wallet/pull/2508/commits/9522d0c0a5b0fd39ffab1b03235f364c78a972f3

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
paweljakubas: corrected -> https://github.com/input-output-hk/cardano-wallet/pull/2508/commits/9522d0c0a5b0fd39ffab1b03235f364c78a972f3

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/2508/commits/9522d0c0a5b0fd39ffab1b03235f364c78a972f3

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
paweljakubas: ok, so this should be decided on client side - done https://github.com/input-output-hk/cardano-wallet/pull/2508/commits/9522d0c0a5b0fd39ffab1b03235f364c78a972f3

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
paweljakubas: corrected here https://github.com/input-output-hk/cardano-wallet/pull/2508/commits/9522d0c0a5b0fd39ffab1b03235f364c78a972f3

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
Anviking: I'm curious where you will go, because for #2521 I wonder if it couldn't be generalised to:

```haskell
:: !(Map
    (KeyFingerprint (s :: Symbol) key)
    (Index 'Soft 'AddressK, keyState)
)
```
to also work with stake keys

paweljakubas: yes, thanks, I am aware that `AddressPool` should be somehow reused here, the first step is here https://github.com/input-output-hk/cardano-wallet/pull/2508/commits/a69d4db860f8f949f588b0533b06260c06a8966a But maybe it will be possible to merry this with what is in Sequential.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
paweljakubas: some attempt here https://github.com/input-output-hk/cardano-wallet/pull/2508/commits/a69d4db860f8f949f588b0533b06260c06a8966a

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
```diff
@@ -260,23 +273,48 @@ data AddressPool
...
+    build (paymentTemplate, delegationTemplateM) =
+        mempty <> " payment script credential: " <> scriptPaymentF <>
+        " delegation script credential: " <> scriptDelegationF
+      where
+        scriptPaymentF = build $ toText paymentTemplate
+        scriptDelegationF = case delegationTemplateM of
+            Just s -> build $ toText s
+            Nothing -> "absent"
+
+instance ToText ScriptTemplate where
```
KtorZ: Shouldn't this be a `Buildable` instance? It doesn't look like an appropriate `ToText` instance.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
```diff
@@ -399,16 +355,24 @@ extendVerificationKeyPool ix pool
...
-    case typeRep (Proxy :: Proxy c) of
-        t | t == typeRep (Proxy :: Proxy 'UtxoInternal) ->
-            UtxoInternal
-          | t == typeRep (Proxy :: Proxy 'UtxoExternal) ->
-            UtxoExternal
-          | t == typeRep (Proxy :: Proxy 'MultisigScript) ->
-            MultisigScript
-        _ ->
-            MutableAccount
+role =  fromJust (tryUtxoExternal <|> tryUtxoInternal <|> tryMultisigScript <|> tryMutableAccount)
```
Anviking: `fromMaybe (error "role: unmatched type" <> show typeRep (Proxy :: Proxy c))`, for clearer error message _if_ were to go wrong?

(Or maybe `showsTypeRep` is the right function)

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -684,9 +715,9 @@ migrateManually tr proxy defaultFieldValues =
     -- | This table became @protocol_parameters@.
     removeOldTxParametersTable :: Sqlite.Connection -> IO ()
     removeOldTxParametersTable conn = do
-        dropTable <- Sqlite.prepare conn "DROP TABLE IF EXISTS tx_parameters;"
-        void $ Sqlite.stepConn conn dropTable
-        Sqlite.finalize dropTable
+        dropTable' <- Sqlite.prepare conn "DROP TABLE IF EXISTS tx_parameters;"
+        void $ Sqlite.stepConn conn dropTable'
+        Sqlite.finalize dropTable'
```
KtorZ, line 720: alternatively, I've started using something more generic for this intermediate variable in other migration like `statement` or `stmt`. 

```
        stmt <- Sqlite.prepare conn "DROP TABLE IF EXISTS tx_parameters;"
        void $ Sqlite.stepConn conn stmt
        Sqlite.finalize stmt
 ```

File: lib/core/src/Cardano/Wallet/DB/Sqlite.hs [hunk 20]
```diff
@@ -451,13 +538,7 @@ migrateManually tr proxy defaultFieldValues =
...
                             , "epoch_stability", "active_slot_coeff"
                             ]
-                _ <- runSql conn $ dropTable tmp
-                _ <- runSql conn $ createTable tmp filtered
-                _ <- runSql conn $ copyTable orig tmp filtered
-                _ <- runSql conn $ dropTable orig
-                _ <- runSql conn $ renameTable tmp orig
-
-                return ()
+                dropColumnOp conn orig tmp filtered
```
KtorZ, line 541: :+1:

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
```diff
@@ -236,6 +240,46 @@ defaultAddressPoolGap =
...
+        :: key 'AccountK XPub
+        -> ParentContext 'UtxoExternal key
+
+    ParentContextUtxoInternal
+        :: key 'AccountK XPub
+        -> ParentContext 'UtxoInternal key
+
+    ParentContextMutableAccount
+        :: key 'AccountK XPub
+        -> ParentContext 'MutableAccount key
```
KtorZ: I don't think there's any use-case for this one, is it?

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
```diff
@@ -564,35 +532,53 @@ extendAddressPool !ix !pool
...
-        . (Proxy @n,)
-        . deriveAddressPublicKey key cc
+
+    mkPaymentKey ix = \case
+        ParentContextUtxoExternal acct ->
+            mkPaymentKeyFromAccXPub acct
+        ParentContextUtxoInternal acct ->
+            mkPaymentKeyFromAccXPub acct
+        ParentContextMutableAccount acct ->
+            mkPaymentKeyFromAccXPub acct  -- think about whether it makes sense
```
KtorZ: Spoiler: I don't think it does :sweat_smile:

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
```diff
@@ -564,35 +532,53 @@ extendAddressPool !ix !pool
...
+        ParentContextMultisigScript _ payment delegation ->
+            mkPaymentKeyFromTemplates payment delegation
```
KtorZ, line 564: Now this seems suspicious to me. when I saw that the `ParentContextMultisigScript` definition above included the parent key, I thought the parent key had been factored out up there. But it isn't use here to construct payment keys, which suggests that `payment` and `delegation` both contains the parent key. Thus, there are two places which contain redundant information. I prefer the approach where the parent keys are held by each template; 

Unless I am missing something, there shouldn't be any parent account key in the `ParentContextMultisigScript` itself.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Script.hs
```diff
@@ -0,0 +1,176 @@
...
+
+replaceCosignersWithVerKeys
+    :: CA.Role
+    -> ScriptTemplate
+    -> Index 'Soft 'ScriptK
+    -> Script KeyHash
+replaceCosignersWithVerKeys role (ScriptTemplate xpubs scriptTemplate) ix =
+    replaceCosigner scriptTemplate
+  where
+    replaceCosigner :: Script Cosigner -> Script KeyHash
```
KtorZ, line 99: Side-note: this smells like `Script` would need a `Functor` instance.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Script.hs
```diff
@@ -0,0 +1,176 @@
...
+    replaceCosigner = \case
+        RequireSignatureOf c -> RequireSignatureOf $ toKeyHash c
+        RequireAllOf xs      -> RequireAllOf (map replaceCosigner xs)
+        RequireAnyOf xs      -> RequireAnyOf (map replaceCosigner xs)
+        RequireSomeOf m xs   -> RequireSomeOf m (map replaceCosigner xs)
+        ActiveFromSlot s     -> ActiveFromSlot s
+        ActiveUntilSlot s    -> ActiveUntilSlot s
+    toKeyHash :: Cosigner -> KeyHash
+    toKeyHash c =
+        let ix' = toEnum (fromEnum ix)
```
KtorZ: That `toEnum . fromEnum` transformation is a bit cryptic. The `Enum` interface is quite unsafe in practice, so if this is a safe transformation in this context, it is worth put in a separate function with a type signature and comment to indicate why it's safe. 

Alternatively, I suspect this is a merely a `coerce` ?

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Script.hs
```diff
@@ -0,0 +1,176 @@
...
+            (Just accXPub) = liftXPub <$> Map.lookup c xpubs
```
KtorZ: Is this pattern-match safe? If it is, the function should state why and possibly, make explicit any invariant that are expected of the ScriptTemplate.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Script.hs
```diff
@@ -0,0 +1,176 @@
...
+            verKey = deriveMultisigPublicKey accXPub ix'
+        in hashKey verKey
+    deriveMultisigPublicKey = case role of
+        CA.MultisigForPayment -> deriveMultisigForPaymentPublicKey
+        CA.MultisigForDelegation -> deriveMultisigForDelegationPublicKey
+        _ ->  error "replaceCosignersWithVerKeys is supported only for role=3 and role=4"
```
KtorZ, line 120: This could be enforced with a GADT and/or a type constraint.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Script.hs
```diff
@@ -0,0 +1,176 @@
...
+
+toNetworkTag
+    :: forall (n :: NetworkDiscriminant). Typeable n => CA.NetworkTag
+toNetworkTag =
+    fromMaybe (fromRight' $ mkNetworkDiscriminant 0) tryMainnet
+  where
+    tryMainnet =
+        case testEquality (typeRep @n) (typeRep @'Mainnet) of
+            Just Refl  -> rightToMaybe $ mkNetworkDiscriminant 1
+            Nothing -> Nothing
```
KtorZ, line 130: Ok :+1:

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Script.hs
```diff
@@ -0,0 +1,176 @@
...
+    -> Address
+liftDelegationAddress ix dTemplate (KeyFingerprint fingerprint) =
+    Address $ CA.unAddress $
+    delegationAddress (toNetworkTag @n)
+    (PaymentFromScript (ScriptHash fingerprint))
+    (delegationCredential dScript)
+  where
+    delegationCredential = DelegationFromScript . toScriptHash
+    dScript =
+        replaceCosignersWithVerKeys CA.MultisigForDelegation dTemplate ix
```
KtorZ, line 183: :+1:

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,311 @@
...
+                                Shared State
+-------------------------------------------------------------------------------}
+
+-- | Shared wallet is a new kind of wallet owned by one or more co-signers.
+-- | In this type of the wallet, addresses are defined by two monetary scripts
+-- | (one for the ownership of assets and the other one for the ownership of stake).
+-- | The two scripts can be the same, different, also the script for the ownership of stake
+-- | can be absent.
+--
+-- | The shared wallet can be in two states: pending and ready. In a pending state the wallet
```
KtorZ: "pending or ready" I imagine, unless we have Schrödinger states :p

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,311 @@
...
+
+trySharedState
+    :: forall (n :: NetworkDiscriminant) k.
+     ( MkKeyFingerprint k (Proxy n, k 'AddressK XPub)
+     , MkKeyFingerprint k Address
+     , Typeable n
+     , SoftDerivation k )
+    => SharedState n k
+    -> SharedState n k
+trySharedState = \case
```
KtorZ, line 230: I am not sure what this function does (turn a "pending" into an active state or identity if already active?)? Maybe the intent could be clarified with a better name?

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,311 @@
...
+                else
+                    sc
+            st = PendingSharedState
+                { pendingSharedStateDerivationPrefix = prefix
+                , pendingSharedStateAccountKey = accXPub'
+                , pendingSharedStatePaymentTemplate = updateScriptTemplate pT
+                , pendingSharedStateDelegationTemplate = updateScriptTemplate <$> dT
+                , pendingSharedStateAddressPoolGap = g
+                }
+        in trySharedState st
```
KtorZ, line 286: Ok :+1:

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -0,0 +1,311 @@
...
+    => Address
+    -> SharedState n k
+    -> (Maybe (Index 'Soft 'ScriptK, KeyHash), SharedState n k)
+isShared addr st = case st of
+    SharedState prefix pool ->
+        let (ixM, pool') = lookupAddress @n (const Used) addr pool
+            (ParentContextMultisigScript accXPub _ _) = context pool
+        in case ixM of
+            Just ix ->
+                (Just (coerce ix, keyHashFromAccXPubIx accXPub (coerce ix))
```
KtorZ, line 307: What does this hash corresponds to? (I mean, where do we intend to use it?)

KtorZ (approved):

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
paweljakubas: if we want to have a way to say which verification key was used in discovered script address, then probably we will need to drag along accXPub. It is used in `isShared` to construct keyHash with this accXPub and used index

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/2508/commits/c0c35b7f84b3e2863bb52d2fc70e1cfd7eb785d2

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
paweljakubas: yeah, removed in  https://github.com/input-output-hk/cardano-wallet/pull/2508/commits/c0c35b7f84b3e2863bb52d2fc70e1cfd7eb785d2

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
paweljakubas: :smile:  removed in https://github.com/input-output-hk/cardano-wallet/pull/2508/commits/c0c35b7f84b3e2863bb52d2fc70e1cfd7eb785d2

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Script.hs
paweljakubas: agree, will do it next time in cardano-addresses

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Script.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-wallet/pull/2508/commits/c0c35b7f84b3e2863bb52d2fc70e1cfd7eb785d2

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Script.hs
paweljakubas: yes, it is safe as we can only do it when each cosigner has corresponding accXPub collected. nevertheless, introduces invariant in https://github.com/input-output-hk/cardano-wallet/pull/2508/commits/c0c35b7f84b3e2863bb52d2fc70e1cfd7eb785d2

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
paweljakubas: this is the verification key corresponding to `ix` and `accXPub` that was found in discovered script address we co-share. Hence, I propose to include `accXPub` , besides script templates, to parent ctx

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
paweljakubas: ```turn a "pending" into an active state or identity if already active?``` yes, added comment

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
paweljakubas: corrected in https://github.com/input-output-hk/cardano-wallet/pull/2508/commits/c0c35b7f84b3e2863bb52d2fc70e1cfd7eb785d2

paweljakubas:
bors retry

PR #2509, opened 2021-02-11
Add detailed validation of metadata-server responses

rvl:
### Issue Number

ADP-413

### Overview

- Add detailed validation of metadata-server responses
- Adds more json parsing and validation tests
- Fixes the mock metadata server so that it supports invalid metadata, and only returns properties requested, rather than all properties available.


lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy.hs
```diff
@@ -157,3 +171,57 @@ newtype AssetLogo = AssetLogo
...
+validateMetadataUnit assetUnit@AssetUnit{name} =
+    (validateMinLength 1 name >>= validateMaxLength 30) $> assetUnit
+
+validateMetadataLogo :: AssetLogo -> Either String AssetLogo
+validateMetadataLogo logo
+    | len <= maxLen = Right logo
+    | otherwise = Left $ "Length must be no more than " ++ show maxLen ++ " bytes, got " ++ show len
+  where
+    len = BS.length $ unAssetLogo logo
+    maxLen = 87400
```
KtorZ: `maxLen` should be 65535 (64KB) here right? 87400 is the max-length of the base64-encoded data payload.

lib/core/src/Cardano/Wallet/TokenMetadata.hs
```diff
@@ -198,8 +221,8 @@ data SubjectProperties = SubjectProperties
     -- Name and description are required, both others may be missing the
     -- response.
     , properties ::
-        ( Property "name"
-        , Property "description"
+        ( Maybe (Property "name")
+        , Maybe (Property "description")
```
KtorZ, line 227: :thinking: Why `Maybe` here? These are two required properties.

rvl: This type is just for the response from the metadata-server, where possibly any field may be missing.
The function `metadataFromProperties :: SubjectProperties -> Maybe AssetMetadata` handles the case of missing fields required by cardano-wallet.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy.hs
rvl: OK

KtorZ (approved):
LGTM. Not sure what's missing from the draft PR ?

rvl:
It was still draft because there were test failures, which I eventually found were because I didn't update the mock metadata server after changing the types.

bors r+

PR #2521, opened 2021-02-15
Prototype multi-stake key wallet

Anviking:
# Issue Number

ADP-709

# Overview

- [x] Support creating wallets with `n` stake keys
- [x] When listing addresses we cycle through the available stake keys
    - Unsure if this is the behaviour we ultimately want though
- [x] Rewards are fetched for all accounts 
- [x] Joining a pool should do so with all keys
- [ ] Existing tests should pass
    - [x] Refine abstractions such that byron and icarus wallets are no longer broken (by unifying PaymentAddress and DelegationAddress vaguely in the style of #1702 ) 
    - [x] Integration tests passing after adding DelegationState
    - [ ] Integration tests passing after TransactionLayer changes
    - [ ] Unit tests passing / green CI
- [ ] We should be able to discover stake keys on chain — i.e. creating a 100 stake key wallet and restoring an existing one, should yield the same result. Probably modulo the 100 keys are actually delegated also.

# Comments

- State machine tests seemed to fail with other delegation states than the default one
- Should ensure property tests use arbitrary delegation states
- Should remove magic `mkDelegationState 1 0` with something more descriptive

### DIFF
--- a/lib/core-integration/src/Test/Integration/Framework/DSL.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/DSL.hs
@@ -228,9 +228,9 @@ import Cardano.Wallet.Api.Types
     )
 import Cardano.Wallet.Primitive.AddressDerivation
     ( HardDerivation (..)
+    , MkAddress (..)
     , NetworkDiscriminant (..)
     , Passphrase (..)
-    , PaymentAddress (..)
     , PersistPublicKey (..)
     , Role (..)
     , WalletKey (..)
@@ -1387,7 +1387,7 @@ fixtureRandomWallet = fmap fst . fixtureRandomWalletMws
 
 fixtureRandomWalletAddrs
     :: forall (n :: NetworkDiscriminant) m.
-        ( PaymentAddress n ByronKey
+        ( MkAddress n ByronKey
         , MonadIO m
         , MonadUnliftIO m
         )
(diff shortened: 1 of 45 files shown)


lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -1855,7 +1855,7 @@ shelleyAddresses mw =
         addrXPrv =
             deriveAddressPrivateKey pwd accXPrv UtxoExternal
     in
-        [ paymentAddress @n (publicKey $ addrXPrv ix)
+        [ mkAddress @n (publicKey $ addrXPrv ix) Nothing
```
Anviking, line 1858: I re-added `paymentAddress` as a helper, so this change can now be reverted.

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
```diff
@@ -201,6 +201,7 @@ ProtocolParameters
 -- Track whether the wallet's stake key is registered or not.
 StakeKeyCertificate
     stakeKeyCertWalletId             W.WalletId            sql=wallet_id
+    stakeKeyCertStakeKey             W.RewardAccount       sql=stake_key
```
Anviking, line 204: Now there are multiple stake keys per wallets, so we need an extra column here.

Anviking:
I think this PR can be re-purposed for the real implementation, but closing for now.

PR #2525, opened 2021-02-16
Add mock-token-metadata-server --port CLI option

rvl:
### Overview

Allows the mock server to use a fixed port - otherwise it's a bit annoying.

### Documentation

```
mock-token-metadata-server - for testing cardano-wallet native asset metadata

Usage: mock-token-metadata-server FILE [--port PORT]

Available options:
  FILE                     Path to the JSON file containing metadata.
  --port PORT              Port to listen on (default: random unused port)

This serves only the `POST /metadata/query' endpoint of Cardano metadata-server.
Other endpoints are unimplemented because they are not used by cardano-wallet. 

The OpenAPI specification of the real metadata-server is here: 

  https://github.com/input-output-hk/metadata-server/blob/master/specifications/api/openapi.yaml

To use this, create a `registry.json' file in the same format of that returned
by the metadata-server query endpoint. That is: 

  { "subjects": [entry1, entry2, ...] }

Each entry should match what is usually submitted to the metadata server. There
are some example json files in this registry: 

  https://github.com/input-output-hk/metadata-registry-testnet

To connect your wallet backend to the mock metadata server, use the URL which is
printed out when this program starts. 

Pass the URL to the wallet like so: 

  cardano-wallet serve --token-metadata-server URL ...

On every request, the server will read the file, filter and return the queried
metadata. So you don't need to restart the mock server after editing metadata. 

You can make test requests to the metadata server using curl: 

  curl -i -H "Content-type: application/json" \
    --data '{"subjects":["entry1", "entry2", ...],"properties":["name","description"]}' \
    http://localhost:PORT/metadata/query
```

### DIFF
--- a/lib/core/src/Cardano/Wallet/TokenMetadata/MockServer.hs
+++ b/lib/core/src/Cardano/Wallet/TokenMetadata/MockServer.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedLabels #-}
 {-# LANGUAGE TypeApplications #-}
@@ -18,6 +19,7 @@
 
 module Cardano.Wallet.TokenMetadata.MockServer
     ( withMetadataServer
+    , withMetadataServerOptions
     , queryServerStatic
     , queryServerReloading
 
@@ -57,9 +59,17 @@ import Cardano.Wallet.Unsafe
 import Control.Monad.IO.Class
     ( liftIO )
 import Data.Aeson
-    ( FromJSON (..), ToJSON (..), eitherDecodeFileStrict, object, (.=) )
+    ( FromJSON (..)
+    , ToJSON (..)
+    , Value (..)
+    , eitherDecodeFileStrict
+    , object
+    , (.=)
+    )
 import Data.ByteArray.Encoding
     ( Base (Base16, Base64), convertToBase )
+import Data.Function
+    ( (&) )
 import Data.Generics.Internal.VL.Lens
     ( view )
 import Data.HashSet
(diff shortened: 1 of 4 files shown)


lib/core/src/Cardano/Wallet/TokenMetadata/MockServer.hs
```diff
@@ -93,20 +111,41 @@ import qualified Data.Text.Encoding as T
...
 -- To be used with @queryServerStatic@.
 withMetadataServer
     :: IO (Server MetadataQueryApi)
     -> (TokenMetadataServer -> IO a)
     -> IO a
-withMetadataServer mkServer action = withApplication app (action . mkUrl)
+withMetadataServer = withMetadataServerExt id Nothing
+
+-- | Start a metadata server.
+withMetadataServerExt
```
Anviking: What does `Ext` mean? Seems a bit cryptic to me

lib/shelley/exe/mock-token-metadata-server.hs
```diff
@@ -54,3 +68,7 @@ opts = helper <*> cmd
             , "server. E.g.:\n"
             , "https://github.com/input-output-hk/metadata-registry-testnet/blob/7eb91951a2adf6f9deff9f3fbe990abc536657fa/789ef8ae89617f34c07f7f6a12e4d65146f958c0bc15a97b4ff169f16861707079636f696e.json"
             ])
+
+    portOpt = optional $ option auto $
+        long "port" <> metavar "PORT"
+        <> help "Port to listen on (default: random unused port)"
```
Anviking: 👍

Anviking (approved):
Haven't tried it but lgtm

lib/core/src/Cardano/Wallet/TokenMetadata/MockServer.hs
```diff
@@ -93,20 +111,41 @@ import qualified Data.Text.Encoding as T
 type MetadataQueryApi = "metadata" :> "query"
     :> ReqBody '[JSON] BatchRequest :> Post '[JSON] BatchResponse
 
--- | Start a metadata server.
+-- | Start a metadata server on any random port.
```
Anviking:
```suggestion
-- | Start a metadata server on any random port, and with no middleware.
```
(fully describing the difference to `withMetadataServerExt `)

rvl: Thanks, I renamed and clarified these.

lib/core/src/Cardano/Wallet/TokenMetadata/MockServer.hs
rvl: It was supposed to mean "extended". I have renamed it.

Issue #2526, opened 2021-02-16
Flaky: .\\cardano-wallet-core-test-unit.exe fails in Windows GH actions


piotr-iohk:
# Context
 
 

# Test Case

https://github.com/input-output-hk/cardano-wallet/runs/1906871545?check_suite_focus=true
https://github.com/input-output-hk/cardano-wallet/runs/1909329666?check_suite_focus=true

In the log there is:

```
cardano-wallet-core-test-unit.exe: getMBlocks: VirtualAlloc MEM_COMMIT failed: The paging file is too small for this operation to complete.
1145/100
Error: Process completed with exit code 1.
```

# Failure / Counter-example

---

# Resolution 

@KtorZ says that the high memory usage (well, too high for github actions) is probably due to the number of simultaneous tests that we run in parallel.

We have reduced the number of simultaneous tests (or attempted to), but it still fails.

The Sqlite tests are temporarily skipped on windows - for now.

---

# QA

piotr-iohk:
Seems to be addressed.

PR #2527, opened 2021-02-17
Disable parallel unit tests execution on Windows.

KtorZ:
# Issue Number

#2526

# Overview

- [ ] I have disabled parallel execution of unit tests on Windows, in hope it'd help with the OOM issues.

# Comments

### DIFF
--- a/lib/core/test/unit/Cardano/Byron/Codec/CborSpec.hs
+++ b/lib/core/test/unit/Cardano/Byron/Codec/CborSpec.hs
@@ -49,7 +49,9 @@ import Data.Text
 import Data.Word
     ( Word32, Word64 )
 import Test.Hspec
-    ( Expectation, Spec, describe, it, parallel, shouldBe )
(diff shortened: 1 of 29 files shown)


rvl (commented):
Oh, there's quite a lot of `parallel` in the unit tests isn't there.
In `cardano-wallet-core.cabal` we have `-with-rtsopts=-N4` preset for the unit tests.
We could instead make that conditional on platform, and/or provide `+RTS -N1` when running the tests under github actions.

KtorZ:
> @rvl We could instead make that conditional on platform, and/or provide +RTS -N1 when running the tests under github actions.

I've actually tried locally to also pass `-j1` to hspec (according to http://hspec.github.io/options.html) but the "parallelization" is still done on a single core :grimacing: and I am reaching 13GB of memory usage all-the-same. I wish there was an option in hspec to simply disable parallel testing on demand, but using less cores does not actually help :(

piotr-iohk (approved):
:crossed_fingers:

PR #2542, opened 2021-03-01
Fix calculation of missing balance in API error

Anviking:
# Issue Number

ADP-697

# Overview

- [x] Add new `difference` function for `TokenMap` and `TokenBundle`
- [x] Calculate the missing balance using `required - available` using `difference` rather than `subtract` 
- [x] TODO: More tests?
- [ ] TODO: Can we property test the API handler?
    - Since this logic is in the API rendering of the errors, it seems very tricky.

# Comments

Analysis of problem:

In Server.hs there's this calculation:

```
                        missing
                            = TokenBundle.Flat
                            $ fromMaybe TokenBundle.empty
                            $ TokenBundle.subtract
                                balanceRequired balanceAvailable
```

But because of this requirement of TokenBundle.subtract
```haskell
-- Returns 'Nothing' if the second bundle is not less than or equal to the first
-- bundle when compared with the `leq` function.
```
it should mean that if we own tokens that are not required, then we hit the nothing case, and we get a missing `coin: 0 tokens: []`.

Think we need something like this:
```haskell
difference :: TokenMap -> TokenMap -> TokenMap
difference = modifySimple $ Map.differenceWith TokenQuantity.subtract
  where
    modifySimple
        :: (Map AssetId TokenQuantity -> Map AssetId TokenQuantity -> Map AssetId TokenQuantity)
        -> TokenMap
        -> TokenMap
        -> TokenMap
    modifySimple f x y = fromFlatList . Map.toList
        $ f (Map.fromList . toFlatList $ x)
            (Map.fromList . toFlatList $ y)
```

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -281,10 +281,10 @@ import Cardano.Wallet.Primitive.AddressDiscovery.Sequential
     , purposeCIP1852
     )
 import Cardano.Wallet.Primitive.CoinSelection.MA.RoundRobin
-    ( BalanceInsufficientError (..)
-    , SelectionError (..)
+    ( SelectionError (..)
     , SelectionInsufficientError (..)
     , UnableToConstructChangeError (..)
+    , balanceMissing
     , selectionDelta
     )
 import Cardano.Wallet.Primitive.Model
@@ -2873,23 +2873,11 @@ instance LiftHandler ErrSelectAssets where
         ErrSelectAssetsSelectionError selectionError ->
             case selectionError of
                 BalanceInsufficient e ->
-                    let
-                        BalanceInsufficientError
-                            { balanceRequired
-                            , balanceAvailable
-                            } = e
-
-                        missing
-                            = TokenBundle.Flat
-                            $ fromMaybe TokenBundle.empty
-                            $ TokenBundle.subtract
-                                balanceRequired balanceAvailable
-                    in
-                        apiError err403 NotEnoughMoney $ mconcat
-                            [ "I can't process this payment as there are not "
-                            , "enough funds available in the wallet. I am only "
-                            , "missing: ", pretty missing
-                            ]
+                    apiError err403 NotEnoughMoney $ mconcat
+                        [ "I can't process this payment as there are not "
+                        , "enough funds available in the wallet. I am only "
+                        , "missing: ", pretty . Flat $ balanceMissing e
+                        ]
                 SelectionInsufficient e ->
                     apiError err403 TransactionIsTooBig $ mconcat
                         [ "I am not able to finalize the transaction "
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -2875,7 +2875,7 @@ instance LiftHandler ErrSelectAssets where
                 BalanceInsufficient e ->
                     apiError err403 NotEnoughMoney $ mconcat
                         [ "I can't process this payment as there are not "
-                        , "enough funds available in the wallet. I am only "
+                        , "enough funds available in the wallet. I am "
                         , "missing: ", pretty . Flat $ balanceMissing e
                         ]
                 SelectionInsufficient e ->
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -328,7 +328,7 @@ import Cardano.Wallet.Primitive.Types.Coin
 import Cardano.Wallet.Primitive.Types.Hash
     ( Hash (..) )
 import Cardano.Wallet.Primitive.Types.TokenBundle
-    ( TokenBundle (..) )
+    ( Flat (..), TokenBundle (..) )
 import Cardano.Wallet.Primitive.Types.TokenMap
     ( AssetId (..) )
 import Cardano.Wallet.Primitive.Types.TokenPolicy
--- a/lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs
@@ -44,6 +44,7 @@ module Cardano.Wallet.Primitive.Types.TokenBundle
     -- * Arithmetic
     , add
     , subtract
+    , difference
 
     -- * Quantities
     , getQuantity
@@ -96,6 +97,8 @@ import Data.Map.Strict
     ( Map )
 import Data.Map.Strict.NonEmptyMap
     ( NonEmptyMap )
+import Data.Maybe
+    ( fromMaybe )
 import Data.Set
     ( Set )
 import Fmt
@@ -105,6 +108,7 @@ import GHC.Generics
 import GHC.TypeLits
     ( ErrorMessage (..), TypeError )
 
+import qualified Cardano.Wallet.Primitive.Types.Coin as Coin
 import qualified Cardano.Wallet.Primitive.Types.TokenMap as TokenMap
 
 --------------------------------------------------------------------------------
@@ -292,6 +296,31 @@ add (TokenBundle (Coin c1) m1) (TokenBundle (Coin c2) m2) =
 --
 subtract :: TokenBundle -> TokenBundle -> Maybe TokenBundle
 subtract a b = guard (b `leq` a) $> unsafeSubtract a b
+
+
+-- | Analogous to @Set.difference@, return the difference between two token
+-- maps.
+--
+-- The following property holds:
+-- prop> x `leq` (x `difference` y) `add` y
+--
+-- Note that there's a `leq` rather than equality, which we'd expect if this was
+-- subtraction of integers. I.e.
+--
+-- >>> (0 - 1) + 1
+-- 0
+--
+-- whereas
+--
+-- >>> let oneToken = fromFlatList coin [(aid, TokenQuantity 1)]
+-- >>> (mempty `difference` oneToken) `add` oneToken
+-- oneToken
+difference :: TokenBundle -> TokenBundle -> TokenBundle
+difference (TokenBundle c1 m1) (TokenBundle c2 m2) =
+    TokenBundle
+        (fromMaybe (Coin 0) $ Coin.subtractCoin c1 c2)
+        (TokenMap.difference m1 m2)
+
 
 --------------------------------------------------------------------------------
 -- Quantities
--- a/lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs
@@ -297,7 +297,6 @@ add (TokenBundle (Coin c1) m1) (TokenBundle (Coin c2) m2) =
 subtract :: TokenBundle -> TokenBundle -> Maybe TokenBundle
 subtract a b = guard (b `leq` a) $> unsafeSubtract a b
 
-
 -- | Analogous to @Set.difference@, return the difference between two token
 -- maps.
 --
@@ -320,8 +319,6 @@ difference (TokenBundle c1 m1) (TokenBundle c2 m2) =
     TokenBundle
         (fromMaybe (Coin 0) $ Coin.subtractCoin c1 c2)
         (TokenMap.difference m1 m2)
-
-
 --------------------------------------------------------------------------------
 -- Quantities
 --------------------------------------------------------------------------------
--- a/lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs
@@ -531,20 +531,11 @@ subtract a b = guard (b `leq` a) $> unsafeSubtract a b
 -- >>> (mempty `difference` oneToken) `add` oneToken
 -- oneToken
 difference :: TokenMap -> TokenMap -> TokenMap
-difference = modifySimple $ Map.differenceWith TokenQuantity.subtract
+difference m1 m2 = L.foldl' reduce m1 (toFlatList m2)
   where
-    toFlatMap :: TokenMap -> Map AssetId TokenQuantity
-    toFlatMap  = Map.fromList . toFlatList
-
-    fromFlatMap :: Map AssetId TokenQuantity -> TokenMap
-    fromFlatMap = fromFlatList . Map.toList
-
-    modifySimple
-        :: (Map AssetId TokenQuantity -> Map AssetId TokenQuantity -> Map AssetId TokenQuantity)
-        -> TokenMap
-        -> TokenMap
-        -> TokenMap
-    modifySimple f x y = fromFlatMap (f (toFlatMap x) (toFlatMap y))
+    reduce :: TokenMap -> (AssetId, TokenQuantity) -> TokenMap
+    reduce m (a, q) = adjustQuantity m a
+        (fromMaybe TokenQuantity.zero . (`TokenQuantity.subtract` q))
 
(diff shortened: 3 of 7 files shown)


lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs
```diff
@@ -293,6 +294,13 @@ add (TokenBundle (Coin c1) m1) (TokenBundle (Coin c2) m2) =
 subtract :: TokenBundle -> TokenBundle -> Maybe TokenBundle
 subtract a b = guard (b `leq` a) $> unsafeSubtract a b
 
+difference :: TokenBundle -> TokenBundle -> TokenBundle
```
Anviking, line 317: Should describe the difference compared to `subtract` — which I'm, to be honest, not entirely sure of.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs
```diff
@@ -293,6 +294,13 @@ add (TokenBundle (Coin c1) m1) (TokenBundle (Coin c2) m2) =
...
+difference (TokenBundle (Coin c1) m1) (TokenBundle (Coin c2) m2) =
+    TokenBundle
+        (Coin $ c1 - c2)
```
Anviking: Should add the properties for TokenMap to TokenBundle too. This `-` seems dangerous.

lib/core/test/unit/Cardano/Wallet/Primitive/Types/TokenMapSpec.hs
```diff
@@ -255,6 +261,19 @@ prop_subtract_invariant m1 m2 = property $
   where
     Just result = TokenMap.subtract m1 m2
 
+prop_difference_zero :: TokenMap -> Property
+prop_difference_zero x =
+    x `difference` TokenMap.empty === x
+
+prop_difference_leq :: TokenMap -> TokenMap -> Property
+prop_difference_leq x y = property $
+    x `difference` y `leq` x
```
jonathanknowles: Should the reverse not also be true?

I.e.:

```haskell
x `difference` y `leq` y
```

And I'm guessing there could be a symmetric property:

```hs
x `difference` y = y `difference` x
```

lib/core/test/unit/Cardano/Wallet/Primitive/Types/TokenMapSpec.hs
```diff
@@ -255,6 +261,19 @@ prop_subtract_invariant m1 m2 = property $
...
+
+prop_difference_add :: TokenMap -> TokenMap -> Property
+prop_difference_add x y = property $
+    (x `difference` y) `TokenMap.add` y
+        === x
```
jonathanknowles: I think in general this isn't true, unless this is also true:

```hs
y `leq` x
```

Anviking: Yup, I just noticed.

Anviking: This seems to be true though:
```haskell
prop_difference_add :: TokenBundle -> TokenBundle -> Property
prop_difference_add x y = property $
    ((x `difference` y) `add` y)
        `geq` x
  where
    geq a b = not $ leq a b
```

jonathanknowles: The definition of `geq` should also handle the equality case:

```hs
geq a b = a == b || (not $ leq a b)
```

lib/core/test/unit/Cardano/Wallet/Primitive/Types/TokenMapSpec.hs
Anviking: If `x = {a -> 1}` and `y = {}`, then `x - y = {a -> 1}`, which is larger than `{}`

```haskell
       Falsifiable (after 8 tests and 8 shrinks):
         TokenMap (fromList [(UnsafeTokenPolicyId (Hash "\170\170\170\170\170\170\170\170\170\170\170\170\170\170\170\170\170\170\170\170\170\170\170\170\170\170\170\170"),NonEmptyMap {least = (UnsafeTokenName "TokenA",TokenQuantity 1), rest = fromList []})])
         TokenMap (fromList [])
```

lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs
Anviking: Fixed.

lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs
Anviking: Updated. Although, not with emphasis on the difference to `subtract`.

lib/core/test/unit/Cardano/Wallet/Primitive/Types/TokenBundleSpec.hs
```diff
@@ -39,6 +41,46 @@ spec =
             , semigroupMonoidLaws
             ]
 
+    describe "Arithmetic" $ do
+        it "prop_difference_zero (x - 0 = x)" $
+            property prop_difference_zero
+        it "prop_difference_zero2 (0 - x = 0)" $
```
Anviking, line 56: Didn't want to break the naming convention too much, but I _did_ want to provide more readable expressions.

Anviking: Also, prop_difference_zero{1,2,3} could be unified into a single property, but I think this might be simpler/more readable in some sense.

lib/core/test/unit/Cardano/Wallet/Primitive/Types/TokenMapSpec.hs
Anviking: I went with
```haskell
b `leq` a
```

lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs
```diff
@@ -293,6 +297,31 @@ add (TokenBundle (Coin c1) m1) (TokenBundle (Coin c2) m2) =
 subtract :: TokenBundle -> TokenBundle -> Maybe TokenBundle
 subtract a b = guard (b `leq` a) $> unsafeSubtract a b
 
+
```
jonathanknowles:
```suggestion
```

lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs
```diff
@@ -293,6 +297,31 @@ add (TokenBundle (Coin c1) m1) (TokenBundle (Coin c2) m2) =
...
+-- >>> let oneToken = fromFlatList coin [(aid, TokenQuantity 1)]
+-- >>> (mempty `difference` oneToken) `add` oneToken
+-- oneToken
+difference :: TokenBundle -> TokenBundle -> TokenBundle
+difference (TokenBundle c1 m1) (TokenBundle c2 m2) =
+    TokenBundle
+        (fromMaybe (Coin 0) $ Coin.subtractCoin c1 c2)
+        (TokenMap.difference m1 m2)
+
+
```
jonathanknowles:
```suggestion

```

lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs
```diff
@@ -293,6 +297,31 @@ add (TokenBundle (Coin c1) m1) (TokenBundle (Coin c2) m2) =
...
+-- whereas
+--
+-- >>> let oneToken = fromFlatList coin [(aid, TokenQuantity 1)]
+-- >>> (mempty `difference` oneToken) `add` oneToken
+-- oneToken
+difference :: TokenBundle -> TokenBundle -> TokenBundle
+difference (TokenBundle c1 m1) (TokenBundle c2 m2) =
+    TokenBundle
+        (fromMaybe (Coin 0) $ Coin.subtractCoin c1 c2)
+        (TokenMap.difference m1 m2)
```
jonathanknowles, line 321: 👍🏻

lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs
```diff
@@ -512,6 +513,39 @@ add a b = F.foldl' acc a $ toFlatList b
...
+
+    fromFlatMap :: Map AssetId TokenQuantity -> TokenMap
+    fromFlatMap = fromFlatList . Map.toList
+
+    modifySimple
+        :: (Map AssetId TokenQuantity -> Map AssetId TokenQuantity -> Map AssetId TokenQuantity)
+        -> TokenMap
+        -> TokenMap
+        -> TokenMap
+    modifySimple f x y = fromFlatMap (f (toFlatMap x) (toFlatMap y))
```
jonathanknowles: This function can actually be implemented in terms of `adjustQuantity`:

```suggestion
difference m1 m2 = L.foldl' reduce m1 (toFlatList m2)
  where
    reduce :: TokenMap -> (AssetId, TokenQuantity) -> TokenMap
    reduce m (a, q) = adjustQuantity m a
        (fromMaybe TokenQuantity.zero . (`TokenQuantity.subtract` q))
```

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -593,6 +600,18 @@ prop_performSelection_small minCoinValueFor costFor (Blind (Small criteria)) =
         . cover 10 (selectionLimited && selectionInsufficient result)
             "selection limited and insufficient"
   where
+    assetsInUTxO = not
+        . Set.null
+        . UTxOIndex.assets
+        $ utxoAvailable criteria
```
jonathanknowles: Suggestion: make this name a bit more explicit.

Justification: with the name `assetsInUTxO`, it might seem to the reader that this is the set of assets included in the UTxO.

```suggestion
    utxoHasAtLeastOneAsset = not
        . Set.null
        . UTxOIndex.assets
        $ utxoAvailable criteria
```

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -593,6 +600,18 @@ prop_performSelection_small minCoinValueFor costFor (Blind (Small criteria)) =
...
+
+    assetsInOutputs = not . Set.null $ TokenBundle.getAssets outputTokens
```
jonathanknowles: Suggestion: make this name a bit more explicit.

Justification: with the name `assetsInOutputs`, it might seem to the reader that this is the set of assets included in the outputs.

```suggestion
    outputsHaveAtLeastOneAsset =
        not . Set.null $ TokenBundle.getAssets outputTokens
```

jonathanknowles (commented):
Hi @Anviking 

Thanks for doing this!

I think this looks good so far. I have a few suggestions which I've left in comments.

I think there are a couple of things missing:

1. a property that tests **_what the result should be_**. In other words, we should test that the quantities produced by `difference a b` are equivalent to performing `TokenQuantity.subtract` on each pair of token quantities, with a default value of `0` for values that would be negative.
2. a property that checks that `difference` preserves the invariant. We currently do this for all of the public operations exported by `TokenMap`. Something like:
    ```hs
    prop_difference_invariant :: TokenMap -> TokenMap -> Property
    prop_difference_invariant m1 m2 =
        property $ invariantHolds $ TokenMap.difference m1 m2
    ```

Lastly, the properties for `difference` should probably be in their own section. (Currently, for `TokenMap`, they're lumped in with the `Invariant properties` section.)

lib/core/test/unit/Cardano/Wallet/Primitive/Types/TokenBundleSpec.hs
jonathanknowles: I think it's fine to have separate properties here. 👍🏻

Anviking:
Thanks for the suggestions @jonathanknowles 

About these:

> Lastly, the properties for difference should probably be in their own section. (Currently, for TokenMap, they’re lumped in with the Invariant properties section.)

Unless I’m missing something, they are in the Arithmetic section, lumped with the previously existing `add` and `subtract`.

> a property that tests what the result should be. In other words, we should test that the quantities produced by difference a b are equivalent to performing TokenQuantity.subtract on each pair of token quantities, with a default value of 0 for values that would be negative.

Wouldn't this be re-doing the re-implementation in the tests? Don't you think that is superfluous?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2873,23 +2873,11 @@ instance LiftHandler ErrSelectAssets where
...
-                                balanceRequired balanceAvailable
-                    in
-                        apiError err403 NotEnoughMoney $ mconcat
-                            [ "I can't process this payment as there are not "
-                            , "enough funds available in the wallet. I am only "
-                            , "missing: ", pretty missing
-                            ]
+                    apiError err403 NotEnoughMoney $ mconcat
+                        [ "I can't process this payment as there are not "
+                        , "enough funds available in the wallet. I am only "
```
jonathanknowles: Suggestion: reword this slightly.

Justification: it's possible the user is missing a large amount!

```suggestion
                        , "enough funds available in the wallet. I am "
```

lib/core/test/unit/Cardano/Wallet/Primitive/Types/TokenMapSpec.hs
```diff
@@ -255,6 +268,37 @@ prop_subtract_invariant m1 m2 = property $
   where
     Just result = TokenMap.subtract m1 m2
 
+prop_difference_zero :: TokenMap -> Property
+prop_difference_zero x =
+    x `difference` mempty === x
```
jonathanknowles: This property (and others) are currently in the "Invariant properties" section (starting at line 228).

They should probably be moved to the "Arithmetic properties" section.

See: https://github.com/input-output-hk/cardano-wallet/pull/2548/commits/a9c2f9a578b930e8c6100a386f09cbbadefc037d

lib/core/test/unit/Cardano/Wallet/Primitive/Types/TokenMapSpec.hs
```diff
@@ -255,6 +268,37 @@ prop_subtract_invariant m1 m2 = property $
...
+        delta = x `difference` y
+        yAndDelta = delta `TokenMap.add` y
+    in
+        counterexample ("x - y = " <> show delta) $
+        counterexample ("(x - y) + y = " <> show yAndDelta) $
+        property $ x `leq` yAndDelta
+
+prop_difference_invariant :: TokenMap -> TokenMap -> Property
+prop_difference_invariant m1 m2 =
+    property $ invariantHolds $ TokenMap.difference m1 m2
```
jonathanknowles, line 277: This property actually **_is_** an invariant property, so it can remain where it is!

jonathanknowles (commented):
Hi @Anviking 

Thanks for updating your PR!

In another PR (#2548) based on this one, I've added a couple of properties for the `difference` function (and added you as a reviewer).

lib/core/test/unit/Cardano/Wallet/Primitive/Types/TokenMapSpec.hs
Anviking: Ah! I was looking at the HSpec tree — thanks!

Anviking:
bors try

jonathanknowles (approved):
Hi @Anviking

LGTM!

Thanks again for making further changes.

PR #2543, opened 2021-03-01
E2E tests on testnet

piotr-iohk:
# Issue Number

ADP-607

# Overview

Moving E2E tests from https://github.com/piotr-iohk/cardano-wallet-rb to cardano-wallet.

# Comments

 - only e2e tests of cardano-wallet are moved, any tests related only to cardano-wallet-rb are not moved here.
 - the ruby wrapper over cardano-wallet stays for now in original repository https://github.com/piotr-iohk/cardano-wallet-rb


rvl (commented):
This is cool, thanks @piotr-iohk . I haven't tried running it locally yet, but I've had a browse of the code.

Any testnet wallets required for needed for these testcaes ought to be defined in an external file which is loaded as a github actions secret. Empty wallets for testing should be generated from new mnemonics each time.

It looks like many test cases are quite similar to what we already have in the hspec integration tests. I would be inclined to drop any rspec tests which are the same as a hspec test. (Perhaps in a subsequent PR though).

Using this, we should be able able to replace all the manual tests with rspec tests, right? (Except for test/manual/FlakyTest.md, but I don't think that one has much value anyway).

We could/should use the real testnet token metadata server in these tests right?

There is a file with a big list of addresses - is that needed?

piotr-iohk:
> Any testnet wallets required for needed for these testcaes ought to be defined in an external file which is loaded as a github actions secret. Empty wallets for testing should be generated from new mnemonics each time.

That's the case for empty wallets. Advantage of having mnemonics for other fixture wallets in code is that all tests can be easily run locally. 

> It looks like many test cases are quite similar to what we already have in the hspec integration tests. I would be inclined to drop any rspec tests which are the same as a hspec test. (Perhaps in a subsequent PR though).

Yes. I think some can be dropped.

> Using this, we should be able able to replace all the manual tests with rspec tests, right? (Except for test/manual/FlakyTest.md, but I don't think that one has much value anyway).

I think it mainly extends coverage by running e2e tests in conditions closer to mainnet. Those manual tests are mostly corner cases that are cumbersome to automate. Those e2e tests touch on some aspects of those manual tests (like SyncingInByronEra.md) but they do not replace them 1-1. 

> We could/should use the real testnet token metadata server in these tests right?

Yes, that's the goal.

> There is a file with a big list of addresses - is that needed?

That's used by one test importing 130k addresses to Byron wallet. I think it can be dropped, as there is [similar test](https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core-integration/src/Test/Integration/Scenario/API/Byron/Addresses.hs#L116) in local integration suite.

rvl:
If the fixture wallet mnemonics are checked in to git, then anyone could mess with them and make our tests fail.
It's possible to have the fixture wallets in a separate file, not checked in.

I am sure we can automate those manual tests. They are also cumbersome to run manually - especially on 3 platforms.

rvl (approved):
:tada:

rvl:
Is it possible to configure the directory where data is stored?
I usually prefer to keep chain dbs, etc outside of my git work tree.

piotr-iohk:
> If the fixture wallet mnemonics are checked in to git, then anyone could mess with them and make our tests fail.
It's possible to have the fixture wallets in a separate file, not checked in.

Ok, fair enough. :+1: I'll change this.

>Is it possible to configure the directory where data is stored?
I usually prefer to keep chain dbs, etc outside of my git work tree.

Yes, via env variables NODE_DB and WALLET_DB, there's this script checked in illustrates this -> https://github.com/input-output-hk/cardano-wallet/blob/da0349b67fd1fdb398a7fdbe06e7cbed72cdeb50/test/e2e/run_all_tests but I'll add a note to readme. 

> I am sure we can automate those manual tests. They are also cumbersome to run manually - especially on 3 platforms.

Yeah, probably they can be automated but I'm not sure about the cost vs. benefit, because they are a bit specific. For the release I'll still would like to run some smoke/sanity manual tests on each platform.

piotr-iohk:
@rvl it seems I don't have permissions to define GH actions secrets on cardano-wallet repo. Can you give me permissions?

rvl:
I ran this locally -- looks cool. The E2E shelley tests failed, probably because I wasn't able to decrypt the fixture wallets.

Would it be possible (now or later) to:
1. Make the `rake fixture_wallets_decode` command to skip if `fixture_wallets.json` already exists. So then we can decrypt once locally, then not need to set the environment variable.
2. Put the `*.gpg` file(s) in a `test/e2e/fixtures` subdirectory.
3. Have an option to skip the `rake get_latest_bins` task. Allow the `PATH` variable to be used ih the normal way for executing binaries.
4. Give [testing environment variables](https://github.com/input-output-hk/cardano-wallet/wiki/Testing#environment-variables) consistent names wherever possible. So prefix variable names with `TESTS_`. Other tests use `TESTS_LOGDIR` for the log directory, for example.
4. Rather than having multiple environment variables that all need configuration, have a `TESTS_STATEDIR` variable. Then the default value of other variables can be relative to that - e.g. `NODE_DB=$TESTS_STATEDIR/db WALLET_DB=$TESTS_STATEDIR/wallets`.

test/e2e/README.md
```diff
@@ -0,0 +1,131 @@
...
+$ rake spec
+$ rake stop_node_and_wallet
+```
+
+##### Running tests against local wallet
+
+One can also run tests against `cardano-wallet` and `cardano-node` which are specified on machine's `$PATH`:
+
+```
+$ rake run_on[testnet,local]
```
rvl, line 75: This way of setting the option seems odd.
Could it be done through an environment variable? Perhaps if `TESTS_E2E_BINDIR` is defined but empty, then use exes on the PATH.

rvl (approved):
This is a great start.

I pushed a few changes so that it's possible to use the existing nix-build to get cardano-wallet, cardano-node, and configs, rather than downloading the latest from hydra.

Please go ahead and merge when you're happy.

test/e2e/README.md
piotr-iohk: > Perhaps if TESTS_E2E_BINDIR is defined but empty, then use exes on the PATH.

I was thinking about this way too, but I thought that it's more convenient (i.e. less writing) if someone want's to switch between running tests against latest Hydra builds to testing against local installation and vice versa.

I understand that it might appear odd, but on the other hand it's sort of conventional way of providing parameters to `rake` task.

>I pushed a few changes so that it's possible to use the existing nix-build to get cardano-wallet, cardano-node, and configs, rather than downloading the latest from hydra.

Cool! how does one do that? Could you share some commands?
EDIT: Ok, pls ignore. I think found that part, it's just skipping the download when config pointing to a directory that's not writable (assuming that's a /nix/store).

piotr-iohk: > Please go ahead and merge when you're happy.

Ok, thanks. I'd like to merge today but I'd like to figure out the failures first, not to merge flaky tests.
I was thinking to run those tests on nightly basis for the time being (so not to be started on every merge) the reason is they take ~1h (and even more on Windows). Majority of the time is spent for syncing wallets. I think when we'll be able to improve restoration times (mainly ADP-690 and ADP-785) then the test execution time will benefit too.

piotr-iohk: >Perhaps if TESTS_E2E_BINDIR is defined but empty, then use exes on the PATH.

Ok, I've added this way too. It was just modifying one line in code.

```
$ TESTS_E2E_BINDIR="" rake run_on[testnet]
```
will also use wallet/node from $PATH.

piotr-iohk:
bors retry

PR #2546, opened 2021-03-03
Asset metadata JSON schema updates

rvl:
### Issue Number

ADP-770

### Overview

- Increase max length of `ticker` field to 5.
- Clarify OpenAPI documentation strings.
- Add a `ticker` sub-field to `unit`.
- Require `unit` field `decimals` to be positive. If it is 0, then `unit` wouldn't be needed.
- Rename `anSignatures` to `signatures`and add `sequenceNumber` in properties of the metadata-server JSON response. These are not used yet, but I may as well update it so we're ready to check signatures.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy.hs
@@ -217,6 +217,7 @@ instance NFData AssetMetadata
 data AssetUnit = AssetUnit
     { name :: Text -- ^ Name of the larger asset.
     , decimals :: Natural  -- ^ Number of zeroes to add to base unit.
+    , ticker :: Maybe Text -- ^ Optional abbreviation of larger asset name
     } deriving (Generic, Show, Eq, Ord)
 
 instance NFData AssetUnit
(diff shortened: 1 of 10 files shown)


lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy.hs
```diff
@@ -278,8 +279,17 @@ validateMetadataURL = fmap AssetURL .
...
+validateMetadataUnit = validateName >=> validateTicker >=> validateDecimals
+  where
+    validateName u@AssetUnit{name} =
+        (validateMinLength 1 name >>= validateMaxLength 30) $> u
+    validateTicker u@AssetUnit{ticker} =
+        case fmap validateMetadataTicker ticker of
+            Just (Left e) -> Left e
+            _ -> Right u
+    validateDecimals u@AssetUnit{decimals}
+        | decimals > 0 = Right u
```
KtorZ: Note that the submitter also enforces that decimals are < 20

KtorZ (approved):

lib/core/src/Cardano/Wallet/Primitive/Types/TokenPolicy.hs
rvl: Fixed, thanks.

PR #2548, opened 2021-03-03
Add extra `difference` properties to Token{Bundle,Map}

jonathanknowles:
# Issue Number

#2542 

# Overview

This PR adds the following properties that relate to the `difference` function:

- [x] `prop_difference_subtract`: relates the `difference` function to the `subtract` function.
- [x] `prop_difference_equality`: for bundles **x** and **y**, tests that reducing **x** by its excess over **y** gives the same result as reducing **y** by its excess over **x**.

### DIFF
--- a/lib/core/test/unit/Cardano/Wallet/Primitive/Types/TokenBundleSpec.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Primitive/Types/TokenBundleSpec.hs
@@ -6,12 +6,13 @@ module Cardano.Wallet.Primitive.Types.TokenBundleSpec
     ( spec
     ) where
 
-import Prelude
+import Prelude hiding
+    ( subtract )
 
 import Algebra.PartialOrd
     ( leq )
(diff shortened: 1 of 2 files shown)


lib/core/test/unit/Cardano/Wallet/Primitive/Types/TokenBundleSpec.hs
```diff
@@ -81,6 +98,23 @@ prop_difference_add x y =
...
+prop_difference_subtract x y =
+    y `leq` x ==> (===)
+        (x `subtract` y)
+        (Just $ x `difference` y)
+
+prop_difference_equality :: TokenBundle -> TokenBundle -> Property
+prop_difference_equality x y = checkCoverage $
+    cover 5 (not (isCoin xReduced))
+        "reduced bundles are not coins" $
+    xReduced === yReduced
```
Anviking, line 111: Hm, ok!

Anviking (approved):

PR #2554, opened 2021-03-05
Enhance go-jira scripts

rvl:
- Change auto-generated known issues list in release notes to use Severity <= 2 instead of Priority.
- Add a script to dump the issue list of the current sprint.
- Add a script to download all attachments from a particular ticket.
- Add a script to show tickets assigned to each team member.

### DIFF
--- a/.jira.d/config.yml
+++ b/.jira.d/config.yml
@@ -28,5 +28,17 @@ custom-commands:
 
   - name: release-notes-bugs
     help: Produce unresolved issues list for cardano-wallet release notes
+    options:
+      - name: minPriority
+        short: p
+        help: Minimum priority for query
+      - name: showPriority
+        type: bool
+        help: Also show priority and severity in the list
+        default: false
+      - name: minSeverity
+        help: Minimum priority for query
+        default: "(1, 2)"
+        hidden: true
     script: |-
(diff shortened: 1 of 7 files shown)


.jira.d/config.yml
```diff
@@ -13,20 +13,72 @@ project: ADP
...
+        required: true
+    script: |-
+      {{jira}} list --template sprint-{{ args.template}} --query "project=$JIRA_PROJECT AND sprint IN openSprints() AND type != epic ORDER BY rank ASC, created" --queryfields=issuetype,summary,status,assignee,priority,customfield_12502,description,customfield_10023,customfield_10021
+
+  - name: sprint-full
+    help: Markdown dump of issues in active sprint
+    script: |-
+      {{jira}} sprint_base full | pandoc -f jira -t markdown
+
   - name: sprint
```
Anviking, line 44: I thought the table-format was nicer than the list (but the new info is nice)

Skärmavbild 2021-04-14 kl  16 55 57

Skärmavbild 2021-04-14 kl  17 10 55

Anviking (approved):
Not familiar with this, but sounds good

.jira.d/config.yml
rvl: OK - changed to table format.

PR #2552, opened 2021-03-04
Split change bundles with excessive asset counts

jonathanknowles:
# Issue Number

ADP-727

# Overview

This PR has two main areas:

## Refactoring
- [x] Relocates all existing type-specific `equipartition` functions into the modules that define their types. We now have:
    | Module | Function |
    | -- | -- |
    | `Numeric.Util` | `equipartitionNatural` |
    | `Coin` | `equipartition` |
    | `TokenQuantity` | `equipartition` |
    | `TokenMap` | `equipartitionQuantities` |
    | `TokenMap` | `equipartitionQuantitiesWithUpperBound` |
    | `TokenBundle` | `equipartitionQuantitiesWithUpperBound` |
    
    **Justification:**
    - This PR will add **_more_** functions for splitting up bundles and maps, and the existing naming scheme was becoming too cumbersome to extend.
    - Equipartitioning a value of type `T` is something that relates to the definition of type `T`, so the natural home for such a function is within module `T`.
    
- [x] Adjusts `makeChange` to accept a record type, and unifies that type with `MakeChangeData` from the test suite.

    **Justification:**
    - Adding further arguments to this function is now less cumbersome, since the compiler will tell us that **_a field is missing_** in all the required call sites (both in implementation code and in test code), rather than producing a series of long type errors.
    - We cut down on some boilerplate from the test suite.

## New Functionality
- [x] Adds the following functions for equipartitioning the asset sets of maps and bundles:
    | Module | Function |
    | -- | -- |
    | `TokenMap` | `equipartitionAssets` |
    | `TokenBundle` | `equipartitionAssets` |
    
- [x] Adds an `assessBundleSize` parameter to `makeChange` and `performSelection`:
    ```hs
    assessBundleSize :: TokenBundle -> TokenBundleSizeAssessment
        
    data TokenBundleSizeAssessment
        = TokenBundleSizeWithinLimit
        -- ^ Indicates that the size of a token bundle does not exceed the maximum
        -- size that can be included in a transaction output.
        | TokenBundleSizeExceedsLimit
        -- ^ Indicates that the size of a token bundle exceeds the maximum size
        -- that can be included in a transaction output.
    ```
    
- [x] Adds function `splitBundlesWithExcessiveAssetCounts` to split change bundles with excessive asset counts, and calls it from `makeChange`.

- [x] Provides a definition of `assessTokenBundleSize` based on functionality exported by `Cardano.Api`.

## Testing

This PR adds:

[...]

PR #2555, opened 2021-03-05
Add an integration test for large asset counts

Anviking:
# Issue Number

Relates to #2552 

# Overview

- [x] Add integration test for covering the border in-between where we don't need to split token-bundle change, and where we do need to, to ensure we have the calculation correct

# Comments

```bash
stack ghci cardano-wallet-core-integration cardano-wallet:integration
```

```
λ> import System.Environment
λ> setEnv "NO_POOLS" "1"
λ> :main --match SeaHorse
```

### DIFF
--- a/lib/core-integration/src/Test/Integration/Faucet.hs
+++ b/lib/core-integration/src/Test/Integration/Faucet.hs
@@ -2242,15 +2242,11 @@ maryIntegrationTestAssets = maMnemonics >>= take 3
         ]
     combined p = simple p `TokenBundle.add` fruit p
 
--- Assets containing 120 different SeaHorses. Same policyId, but all different.
+-- | Create @n@ unique SeaHorse tokens for each provided @Address@.
 --
--- This is for testing with large token bundles / asset counts.
---
--- This is a separate definition from maryIntegrationTestAssets to allow minting
--- only 1 bundle per transaction without going over the maximum size
--- (as we do want them as large as possible).
+-- The result can be used for minting using the cli-based faucet.
 seaHorseTestAssets
-    :: Int
+    :: Int -- ^ Number of sea horses per address
     -> [Address]
     -> [(Address, (TokenBundle, [(String, String)]))]
 seaHorseTestAssets nPerAddr addrs = zip addrs $
--- a/lib/core-integration/src/Test/Integration/Faucet.hs
+++ b/lib/core-integration/src/Test/Integration/Faucet.hs
@@ -20,9 +20,14 @@ module Test.Integration.Faucet
     , mirMnemonics
     , maMnemonics
 
+    -- * Sea horses
+    , seaHorseTokenName
+    , seaHorsePolicyId
+
       -- * Integration test funds
     , shelleyIntegrationTestFunds
     , maryIntegrationTestAssets
+    , seaHorseTestAssets
 
       -- * Internals
     , genByronFaucets
@@ -93,6 +98,7 @@ import qualified Cardano.Wallet.Primitive.AddressDerivation.Byron as Byron
 import qualified Cardano.Wallet.Primitive.AddressDerivation.Icarus as Icarus
 import qualified Cardano.Wallet.Primitive.AddressDerivation.Shelley as Shelley
 import qualified Cardano.Wallet.Primitive.Types.TokenBundle as TokenBundle
+import qualified Data.ByteString.Char8 as B8
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
@@ -2128,7 +2134,7 @@ bigDustWallet = unsafeMkMnemonic
 
 shelleyIntegrationTestFunds :: [(Address, Coin)]
 shelleyIntegrationTestFunds = mconcat
-    [ seqMnemonics >>= (take 10 . map (, defaultAmt) . addresses . SomeMnemonic)
+    [ seqMnemonics >>= take 10 . map (, defaultAmt) . addresses . SomeMnemonic
 
     , zip
          (addresses $ SomeMnemonic onlyDustWallet)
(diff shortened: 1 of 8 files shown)


lib/core-integration/src/Test/Integration/Faucet.hs
```diff
@@ -2239,6 +2250,20 @@ maryIntegrationTestAssets = maMnemonics >>= take 3
         ]
     combined p = simple p `TokenBundle.add` fruit p
 
+    seaHorse is p = bundle p $ flip map is $ \i ->
+        (UnsafeTokenName ("SeaHorse" <> B8.pack (show i)), TokenQuantity 1)
+
+    specialAssets =
+        flip zip (map (\is -> mint (seaHorse is) (maryAssetScripts !! 4)) (chunks 2 [1 .. 200]))
```
Anviking: `maryAssetScripts !! 4` looked like the next un-used one, but should have some nicer way of choosing one.

chunks of 5 assets per address were to big (think we fund 10 or 20 addresses per tx), so went with 2

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -715,6 +718,41 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
                 , expectField (#assets . #total . #getApiT) (`shouldNotBe` TokenMap.empty)
                 ]
 
+    forM_ [64 .. 70] $ \n ->
```
Anviking: Would be nice to print / show some size metric too, as these numbers had to be stumbled upon by luck.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2404,7 +2404,7 @@ instance HasSeverityAnnotation WalletLog where
         MsgTip _ -> Info
         MsgBlocks _ -> Debug
         MsgSelectionStart{} -> Debug
-        MsgSelectionDone{} -> Debug
+        MsgSelectionDone{} -> Notice
```
Anviking: Was temporary

lib/shelley/test/data/cardano-node-shelley/shelley-genesis.yaml
```diff
@@ -16,9 +16,9 @@ protocolParams:
     major: 0
   minUTxOValue: 1000000
   decentralisationParam: 0.25 # means 75% decentralised
-  maxTxSize: 4096
+  maxTxSize: 40960
   minFeeA: 100
-  maxBlockBodySize: 239857
+  maxBlockBodySize: 2398570
```
Anviking: Should revert

lib/shelley/test/data/cardano-node-shelley/shelley-genesis.yaml
```diff
@@ -16,9 +16,9 @@ protocolParams:
     major: 0
   minUTxOValue: 1000000
   decentralisationParam: 0.25 # means 75% decentralised
-  maxTxSize: 4096
+  maxTxSize: 40960
```
Anviking: Pretty sure we need raise the maxTxSize, in order to hit the TxOutputTooLarge error, which also occurs at 4096.

jonathanknowles: > Pretty sure we need raise the maxTxSize, in order to hit the TxOutputTooLarge error, which also occurs at 4096.

It's my understanding that `maxTxSize` is currently `16384` bytes: https://github.com/input-output-hk/cardano-node/blob/99e678c008405426f4af774f16f4f132539622f5/configuration/cardano/mainnet-shelley-genesis.json#L13

maxTxSize

Anviking: 👍 let's use that then

lib/core-integration/src/Test/Integration/Faucet.hs
Anviking: Fixed - moved it to a separate definition

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -715,6 +719,54 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
...
+                                , "status code "
+                                , show s
+                                ]
+                        (_, Left e) -> Left $ show e
+
+                (nPerAddr,) . verifyRes <$> request @(ApiTransaction n) ctx
+                    (Link.createTransaction @'Shelley wSrc) Default payload
+
+            -- 3. They should all succeed
+            responses `shouldBe` (map ((, Right ()) . snd)  sourceWallets)
```
Anviking: With the incorrect (4096 bytes) max bundle size we get:

Skärmavbild 2021-03-09 kl  15 29 58

I.e. it fails, and the failure is still pretty readable.

Anviking:
bors try

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -715,6 +719,56 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
...
+            -- 2. Try spending from each wallet, and record the response.
+            responses <- forM sourceWallets $ \(wSrc, nPerAddr) -> do
+                let seaHorses = map $ \ix ->
+                        (( toText seaHorsePolicyId
+                        , toText $ seaHorseTokenName ix)
+                        , 1)
+                payload <- mkTxPayloadMA @n
+                    destAddr
+                    0
+                    (seaHorses [1, nPerAddr * 2])
```
Anviking, line 751: Magic `2` here. I did consider removing, but you'd need to do more than `let nAddrs = 2` to have it actually make sense. E.g. choose the relevant indexes in the setup phase. Wasn't sure if worth it.

rvl (approved):
Nice. In future when we have our own API for minting, asset tests will be better because we can get rid of the asset faucet, and easily mint what's needed in each test case.

Anviking:
bors try

Anviking:
Need to update the two `Transaction with too much metadata` tests after increasing the max tx size...

lib/core-integration/src/Test/Integration/Faucet.hs
```diff
@@ -2239,6 +2242,51 @@ maryIntegrationTestAssets = maMnemonics >>= take 3
         ]
     combined p = simple p `TokenBundle.add` fruit p
 
+-- Assets containing 120 different SeaHorses. Same policyId, but all different.
```
Anviking: Oops, this is out of date now

lib/core-integration/src/Test/Integration/Framework/Context.hs
```diff
@@ -59,6 +61,12 @@ data Context = Context
         -- era-specific assertions.
     , _smashUrl :: Text
         -- ^ Base URL of the mock smash server.
+
+    , _mintSeaHorseAssets :: Int -> [Address] -> IO ()
+        -- ^ TODO: Remove once we can unify cardano-wallet-core-integration and
+        -- cardano-wallet:integration, or when the wallet supports minting.
+        --
+        -- Cannot be used be serveral tests at a time. (!)
```
jonathanknowles:
```suggestion
        -- Cannot be used by several tests at a time. (!)
```

lib/core-integration/src/Test/Integration/Faucet.hs
```diff
@@ -2239,6 +2242,51 @@ maryIntegrationTestAssets = maMnemonics >>= take 3
...
+--
+-- This is for testing with large token bundles / asset counts.
+--
+-- This is a separate defenition from maryIntegrationTestAssets to allow minting
```
jonathanknowles:
```suggestion
-- This is a separate definition from maryIntegrationTestAssets to allow minting
```

jonathanknowles (approved):
Looks great.

Only spotted a couple of typos in comments (see comments.)

Anviking:
🤔 
bors r+

PR #2559, opened 2021-03-11
Change metavar from STRING to WALLET_NAME

Anviking:
# Issue Number

# Overview

- [x] Change metavar from STRING to WALLET_NAME, to make the help text more understandable
- [x] Can we add help text?
- [x] I guess some test need to be updated too

# Comments

### DIFF
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -1421,7 +1421,8 @@ transactionIdArgument = argumentT $ mempty
 -- | <name=STRING>
 walletNameArgument :: Parser WalletName
 walletNameArgument = argumentT $ mempty
-    <> metavar "STRING"
+    <> metavar "WALLET_NAME"
+    <> help "Name of the wallet."
 
 -- | <public-key=ACCOUNT_PUBLIC_KEY>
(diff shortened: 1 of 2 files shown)


Anviking:
bors try

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1421,7 +1421,8 @@ transactionIdArgument = argumentT $ mempty
 -- | <name=STRING>
 walletNameArgument :: Parser WalletName
 walletNameArgument = argumentT $ mempty
-    <> metavar "STRING"
+    <> metavar "WALLET_NAME"
+    <> help "The name of the new wallet."
```
KtorZ: "new" implies a certain context which may not fit the use-case (although we use it only for creating wallet _I think_).
I'd simply stick to "Name of the wallet" for description.

KtorZ (approved):

lib/cli/src/Cardano/CLI.hs
Anviking: I tried making it sound more like "here you get to choose the local name of the wallet", rather than "something that must match some existing value" — but fine, let's go for "Name of the wallet"

PR #2560, opened 2021-03-12
Validate user-specified output bundles when processing coin selection requests

jonathanknowles:
# Issue Number

ADP-727
ADP-782

# Overview

This PR adds validation to coin selection requests, checking all **user-specified outputs** meet the following pre-conditions:

- output bundles do not include token quantities in excess of the maximum bound:
    `maxBound :: Word64`
- output bundles, when serialized, are within the size limit specified by the protocol:
    `4000` bytes
    
If either of the above pre-conditions are violated, an API error is returned, with a message that informs the user how they can solve the problem (usually by breaking up bundles into smaller bundles):

- `ErrOutputTokenBundleSizeExceedsLimit`
- `ErrOutputTokenQuantityExceedsLimit`

# Testing

This PR adds integration tests for both of the above scenarios:

- `WALLETS_COIN_SELECTION_07`
- `WALLETS_COIN_SELECTION_08`

# Refactoring

This PR also performs a small refactoring in the same spirit as the refactorings included in PR #2552.

We extract out `partition` functions for `Coin` and `TokenQuantity` values into the modules that define their types, similarly to the `X.equipartition` functions.

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/CoinSelections.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/CoinSelections.hs
@@ -97,8 +97,8 @@ import Test.Integration.Framework.DSL
     )
 import Test.Integration.Framework.TestData
     ( errMsg400TxMetadataStringTooLong
-    , errMsg403TokenBundleSizeExceedsLimit
-    , errMsg403TokenQuantityExceedsMaxBound
+    , errMsg403OutputTokenBundleSizeExceedsLimit
+    , errMsg403OutputTokenQuantityExceedsLimit
     , errMsg404NoWallet
     , errMsg406
     , errMsg415
@@ -331,7 +331,7 @@ spec = describe "SHELLEY_COIN_SELECTION" $ do
                     @_ @'Shelley ctx sourceWallet (payment :| [])
             makeRequest >>= flip verify
                 [ expectResponseCode HTTP.status403
-                , expectErrorMessage $ errMsg403TokenQuantityExceedsMaxBound
+                , expectErrorMessage $ errMsg403OutputTokenQuantityExceedsLimit
                     (getApiT $ fst targetAddress)
                     (policyId)
                     (assetName)
@@ -364,7 +364,7 @@ spec = describe "SHELLEY_COIN_SELECTION" $ do
                     @_ @'Shelley ctx sourceWallet (payment :| [])
             makeRequest >>= flip verify
                 [ expectResponseCode HTTP.status403
-                , expectErrorMessage $ errMsg403TokenBundleSizeExceedsLimit
+                , expectErrorMessage $ errMsg403OutputTokenBundleSizeExceedsLimit
                     (getApiT $ fst targetAddress)
                     (assetCount)
                 ]
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/CoinSelections.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/CoinSelections.hs
@@ -9,6 +9,7 @@
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
+{- HLINT ignore "Use camelCase" -}
 
 module Test.Integration.Scenario.API.Shelley.CoinSelections
     ( spec
@@ -45,6 +46,14 @@ import Cardano.Wallet.Primitive.AddressDerivation.Shelley
     ( ShelleyKey )
 import Cardano.Wallet.Primitive.AddressDiscovery.Sequential
     ( coinTypeAda, purposeBIP44, purposeCIP1852 )
+import Cardano.Wallet.Primitive.Types.Hash
+    ( Hash (..) )
+import Cardano.Wallet.Primitive.Types.TokenMap
+    ( AssetId (..) )
+import Cardano.Wallet.Primitive.Types.TokenPolicy
+    ( TokenName (..), TokenPolicyId (..) )
+import Cardano.Wallet.Primitive.Types.Tx
+    ( txOutMaxTokenQuantity )
 import Control.Monad
     ( forM_ )
 import Data.Generics.Internal.VL.Lens
@@ -86,12 +95,15 @@ import Test.Integration.Framework.DSL
     )
 import Test.Integration.Framework.TestData
     ( errMsg400TxMetadataStringTooLong
+    , errMsg403TokenQuantityExceedsMaxBound
     , errMsg404NoWallet
     , errMsg406
     , errMsg415
     )
 
 import qualified Cardano.Wallet.Api.Link as Link
+import qualified Cardano.Wallet.Primitive.Types.TokenMap as TokenMap
+import qualified Cardano.Wallet.Primitive.Types.TokenQuantity as TokenQuantity
 import qualified Data.HashSet as Set
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Text as T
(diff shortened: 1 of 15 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Shelley/CoinSelections.hs
```diff
@@ -293,6 +309,74 @@ spec = describe "SHELLEY_COIN_SELECTION" $ do
...
+        wallets_coin_selection_07 @n
+
+    it "WALLETS_COIN_SELECTION_08 - \
+        \Single output with excessively high number of assets." $
+        wallets_coin_selection_08 @n
+
+-- Attempt to create a coin selection with an output that has an excessively
+-- high token quantity. (This should fail.)
+--
+wallets_coin_selection_07
```
Anviking: Let's stick to the style of having the scenarios inline in the Spec?

jonathanknowles: > Let's stick to the style of having the scenarios inline in the Spec?

I did it this way because including the scenarios inline involved a great deal of excessive wrapping (12 columns are already eaten up by the indentation).

But I don't feel strongly about it: I can re-inline them if you feel strongly about it. 👍🏻

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2866,8 +2869,45 @@ instance LiftHandler (ErrInvalidDerivationIndex 'Soft level) where
...
+        , ". Asset count: "
+        , pretty (view #assetCount e)
+        , "."
+        ]
+
+instance LiftHandler ErrTokenQuantityExceedsMaxBound where
+    handler e = apiError err403 TokenQuantityExceedsMaxBound $ mconcat
+        [ "One of the asset quantities you've specified is greater than the "
+        , "maximum quantity allowed in a single transaction output. Try "
+        , "splitting this quantity across two or more outputs. "
```
Anviking, line 2894: This is because the cap for a single output is `maxBound :: Word64`, but quantities in general are `Natural`?

Ah, yes.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2866,8 +2869,45 @@ instance LiftHandler (ErrInvalidDerivationIndex 'Soft level) where
...
+        , "Destination address: "
+        , pretty (view #address e)
+        , ". Token policy identifier: "
+        , pretty (view #tokenPolicyId $ asset e)
+        , ". Asset name: "
+        , pretty (view #tokenName $ asset e)
+        , ". Token quantity specified: "
+        , pretty (view #quantity e)
+        , ". Maximum allowable token quantity: "
+        , pretty (view #quantityMaxBound e)
```
Anviking, line 2904: Nice!

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/CoinSelections.hs
jonathanknowles: Fixed in 62944743aafa344b1b09935da53f900436818c13.

lib/core/src/Cardano/Wallet/Api/Server.hs
jonathanknowles: That's correct. But the cap only applies to single outputs.

It's quite possible to get **_more than_** `maxBound :: Word64` for a given asset if we're summing across a whole UTxO set to find a wallet's balance.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -1233,19 +1229,7 @@ makeChangeForCoin
...
-makeChangeForCoin targets excess =
-    -- The 'Natural -> Coin' conversion is safe, because 'partitionNatural'
-    -- guarantees to produce a list where every entry is less than or equal to
-    -- the target value.
-    maybe zeroWeightSum (fmap unsafeNaturalToCoin)
-        (partitionNatural (coinToNatural excess) weights)
-  where
-    zeroWeightSum :: HasCallStack => a
-    zeroWeightSum = error
-        "partitionValue: The specified weights must have a non-zero sum."
```
jonathanknowles, line 1245: Before this PR, this error message actually had the wrong function name. (Now fixed.)

lib/core/src/Cardano/Wallet/Transaction.hs
```diff
@@ -168,6 +175,37 @@ defaultTransactionCtx = TransactionCtx
...
+        ErrTokenBundleSizeExceedsLimit
+    | ErrSelectionCriteriaOutputTokenQuantityExceedsMaxBound
+        ErrTokenQuantityExceedsMaxBound
+    deriving (Eq, Generic, Show)
+
+data ErrTokenBundleSizeExceedsLimit = ErrTokenBundleSizeExceedsLimit
+    { address :: !Address
+      -- ^ The address to which this token bundle was to be sent.
+    , assetCount :: !Int
+      -- ^ The number of assets within the token bundle.
```
jonathanknowles, line 190: Note that it might be possible to be more helpful here, by giving a suggestion of **_how many bundles_** the user will need to split their oversized bundle into.

But this is harder than it sounds, as different partitionings will give rise to different serialized lengths. Consider this example:

- Bundle **A**: ❌ (_too large_)
    - `100` assets with names of length `1`.
    - `100` assets with names of length `32`.

We might determine that it's feasible to split this into 2 valid bundles, if the bundles are partitioned this way:

- Bundle **A1**: 🆗 
    - `50` assets with names of length `1`.
    - `50` assets with names of length `32`.
- Bundle **A2**: 🆗 
    - `50` assets with names of length `1`.
    - `50` assets with names of length `32`.

But if the user is not familiar with how bundles are serialized, they might end up partitioning the bundles like this:

- Bundle **A3**: 🆗 
    - `100` assets with names of length `1`.
- Bundle **A4**: ❌ (_too large_)
    - `100` assets with names of length `32`.

In the above example, bundle **A4** is still likely to be rejected.

jonathanknowles: Another option (for the future) would be to provide a suggested partitioning in the API error response, or to provide automated splitting of output bundles (should the user request this).

specifications/api/swagger.yaml
```diff
@@ -2693,6 +2693,33 @@ x-errInvalidCoinSelection: &errInvalidCoinSelection
...
+  title: token_quantity_exceeds_max_bound
+  properties:
+    message:
+      type: string
+      description: |
+        Returned when a user-specified transaction output contains a token
+        quantity that exceeds the maximum bound supported by the protocol.
+    code:
+      type: string
+      enum: ['token_quantity_exceeds_max_bound']
```
Anviking: How about calling the errors `too_many_tokens[_in_output]` and `too_large_token`?

It takes me a while to parse the difference between the the current terms...

token_bundle_size_exceeds_limit
token_quantity_exceeds_max_bound

They both start with `token`, end with `exceeds limit ≈ max bound`.

Although —`too_large_token` — maybe tokens can't have largeness 🤔

jonathanknowles: @Anviking thanks for this comment. I was also a bit unhappy with the naming.

I agree it's very hard to come up with names that are completely unambiguous, since `camelCase` and `snake_case` make it difficult to see the groupings. And using clarification words like "of" and "the" would probably make the names longer than we'd like.

The other confusing aspect is the distinction between **_asset_** and **_token_**.

Just to share with you my thinking: I tend to think of a "token quantity" as a **_discrete quantity_** of some **_asset_**, in the same way we've historically thought of a "coin quantity" as being a **_discrete quantity_** of ada (the asset).

I also think that:

- since these errors pertain to **_outputs_**, the name `Output` should probably appear in the name of each error.
- since both errors pertain to **_exceeding some limit_**, it's better to use consistent terminology. I've gone for "exceeds limit" in both cases.

I've revised the names to the following:

| Name | Meaning |
| -- | -- |
| `output_token_quantity_exceeds_limit` | A user-specified transaction **_output_** contains, for some asset, a **_token quantity_** that **_exceeds_** the **_limit_** supported by the protocol. |
| `output_token_bundle_size_exceeds_limit` | A user-specified transaction **_output_** contains a **_token bundle_** whose **_size_** (serialized length) **_exceeds_** the **_limit_** supported by the protocol. |

See: 7322723cbb9d7d44e020a8f8fb79fe62889da3ae

jonathanknowles:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -432,6 +424,82 @@ _estimateMaxNumberOfInputs txMaxSize ctx outs =
...
+        Left $
+            -- We encountered one or more excessively large token bundles.
+            -- Just report the first such bundle:
+            ErrSelectionCriteriaOutputTokenBundleSizeExceedsLimit $
+            ErrOutputTokenBundleSizeExceedsLimit {address, assetCount}
+    | (address, asset, quantity) : _ <- excessiveTokenQuantities =
+        Left $
+            -- We encountered one or more excessive token quantities.
+            -- Just report the first such quantity:
+            ErrSelectionCriteriaOutputTokenQuantityExceedsLimit $
```
Anviking, line 445: Just wondering: did you consider adding this logic to `performSelection` instead?

Anviking (approved):
lgtm!

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
jonathanknowles: > Just wondering: did you consider adding this logic to `performSelection` instead?

That's a good point. The `performSelection` function would also be a perfectly valid place to put this validation logic.

However, doing so would mean a little more work, as we'd have to adjust the property tests for `performSelection`. It might not be too hard though.

PR #2561, opened 2021-03-12
Print results to stdOut instead of stdErr in restoration benchmark

piotr-iohk:
# Issue Number

ADP-804

# Overview

- [ ] Print results to stdOut instead of stdErr in restoration benchmark

# Comments

It seems that results that should go to restoration-mainnet.txt file are printed to stdErr, which is for some reason not picked up. Trying to print them to stdOut and run benchmark on this branch to see if this helps...

### DIFF
--- a/lib/core-integration/src/Cardano/Wallet/BenchShared.hs
+++ b/lib/core-integration/src/Cardano/Wallet/BenchShared.hs
@@ -88,7 +88,7 @@ import Options.Applicative
     , value
     )
 import Say
-    ( sayErr )
+    ( say )
 import System.Directory
     ( createDirectoryIfMissing )
 import System.Environment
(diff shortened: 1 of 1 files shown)


piotr-iohk:
Started benchmark here -> https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/876

rvl:
Looks like the same result - `restore-mainnet.txt` is empty.

From inspecting `bench-restore.sh`, I see that it shouldn't matter if stdout/stderr is used, because both are collected with `benchCmd 2>&1 | tee restore.log`.

~I believe that the issue is the grep and awk filter on the next line - it eats everything.~ edit: nope

The issue seems to be that the node runs out of space in `/tmp` for its chain database and exits. This causes the benchmark to exit, without printing "All results:".

piotr-iohk:
Yes, indeed. I have updated the ticker ADP-804 and moved it to a "bug" type. Closing this PR.

PR #2563, opened 2021-03-15
Get logs from tests against docker

piotr-iohk:
# Issue Number

# Overview

- [ ] Get logs from tests against docker

# Comments

There were some failures on latest docker run:
https://github.com/input-output-hk/cardano-wallet/runs/2108457759?check_suite_focus=true

Unfortunately there are no logs uploaded from a docker runs currently (like they are from Win/Linux/Mac runs), so potential failures cannot be properly investigated. This change is to attach logs from docker also.


piotr-iohk:
Logs should be now attached for docker workflow as well -> https://github.com/input-output-hk/cardano-wallet/actions/runs/654440971

test/e2e/.rake_tasks~
```diff
@@ -3,6 +3,7 @@ clean_logs
 fixture_wallets_decode
 fixture_wallets_encode
 fixture_wallets_template
+get_docker_logs
```
rvl, line 6: What is this file?
I am used to the convention where files ending with `~` are backup files.
Is it hand-crafted, or auto-generated from `Rakefile`?
If the latter, then it should be removed and added to .gitignore.

test/e2e/Rakefile
```diff
@@ -250,6 +250,27 @@ task :get_latest_configs, [:env] do |task, args|
...
+  puts "\n  >> Getting docker logs into #{LOGS}"
+
+  mk_dir(LOGS)
+  node_logfile = `docker inspect --format='{{.LogPath}}' cardano-node`.gsub("\n",'')
+  wallet_logfile = `docker inspect --format='{{.LogPath}}' cardano-wallet`.gsub("\n",'')
+
+  node_cmd = "sudo cp #{node_logfile} #{LOGS}/node.log"
+  wallet_cmd = "sudo cp #{wallet_logfile} #{LOGS}/wallet.log"
+
+  puts node_cmd
```
rvl: Is there a way for rake to automatically print commands before they are run?

rvl (approved):
Looks helpful, thanks.

test/e2e/.rake_tasks~
piotr-iohk: It was autogenerated first but I've been adding new tasks to it. As far as I see rake uses it for autocompletion so it seemed useful to me to have it checked in, although it looks a bit awkward. There is another one .rakeTasks which seems to be autogenerated by Atom, and it's not useful. I'll remove that one.

piotr-iohk: It was autogenerated first but I've been adding new tasks to it. As far as I see rake uses it for autocompletion so it seemed useful to me to have it checked in, although it looks a bit awkward. There is another one .rakeTasks which seems to be autogenerated by Atom, and it's not useful. I'll remove that one.

test/e2e/Rakefile
piotr-iohk: I don't think so. I think I could wrap it in a `cmd` command or something like that.

piotr-iohk: I have added `cmd` method and wrapped all commands into it. (I've found there is a `sh` command that comes with rake, that does the same pretty much, however It was giving me strange failures on Windows for some reason)

PR #2566, opened 2021-03-17
Nightly restore bench: Make buildkite red if it fails

rvl:
### Issue Number

ADP-804

### Overview

If the nightly restore bench fails to produce a result, then make sure Buildkite has a red status.


piotr-iohk (approved):
:+1:

PR #2567, opened 2021-03-17
Guard pool metadata fetch result persistence with settings status

KtorZ:
# Issue Number

ADP-634

# Overview

- Guard pool metadata fetch result persistence with settings status

  Since #2432 (concurrent pool metadata fetching), the fetching of
  metadata is now done in new short-lived threads forked from the
  'fetchMetadata' thread worker. Killing the parent thread
  'fetchMetadata' does not actually kill child threads which may
  still complete afterwards.

  As a consequence, when changing settings from smash (or direct) to
  none, in-flight requests spawned in short-lived thread may still
  resolve even if the 'fetchMetadata' loop has been terminated.

  This commit makes sure to only write the result of requests if and
  only if the settings haven't changed since the parent thread was
  launched. This should prevent short-lived thread to update the
  database after the parent is killed when changing settings.


piotr-iohk (approved):
LGTM. I've turned on related e2e test which is now passing :tada:

rvl (approved):
Sweet

PR #2568, opened 2021-03-18
Better error handling in restore benchmark

rvl:
### Issue Number

ADP-804

### Overview

- Handle case where node exits before restore bench is finished.
- Fix disk space error.

### Comments

[Buildkite nightly builds](https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds?branch=rvl/adp-804/fix-restore-bench)

The restore benchmark no longer fails from the original error. It now seems to reach the 10 hour timeout on mainnet. So the benchmark wallets need to be made smaller... in another PR.

### DIFF
--- a/.buildkite/bench-restore.sh
+++ b/.buildkite/bench-restore.sh
@@ -14,20 +14,21 @@ log=restore.log
 results=restore-$network.txt
 total_time=restore-time.txt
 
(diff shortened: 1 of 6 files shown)


piotr-iohk (approved):
:crossed_fingers:

PR #2570, opened 2021-03-19
Database changes for resubmission of pending transactions

rvl:
### Issue Number

ADP-764

### Overview

Database layer support for transaction resubmission.

   - [x] Adds a new table with serialized transactions, and the latest slot that they have been submitted at.
   - [x] Add DBLayer method to list pending transactions for resubmissions
   - [x] Add DBLayer method to add a submitted transaction to the pool of transactions.
   - [x] Update db model and state machine tests with generators for new methods
   - [x] Swagger update

The wallet layer and server changes are in PR #2611.

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -187,6 +187,7 @@ import Cardano.Slotting.Slot
     ( SlotNo (..) )
 import Cardano.Wallet.DB
     ( DBLayer (..)
+    , ErrNoSuchTransaction (..)
     , ErrNoSuchWallet (..)
     , ErrRemoveTx (..)
     , ErrWalletAlreadyExists (..)
@@ -1695,7 +1696,7 @@ getTransaction ctx wid tid = db & \DBLayer{..} -> do
         Left err -> do
             throwE (ErrGetTransactionNoSuchWallet err)
         Right Nothing -> do
-            let err' = ErrNoSuchTransaction tid
+            let err' = ErrNoSuchTransaction wid tid
             throwE (ErrGetTransactionNoSuchTransaction err')
         Right (Just tx) ->
             pure tx
@@ -2222,10 +2223,6 @@ data ErrListTransactions
 data ErrGetTransaction
     = ErrGetTransactionNoSuchWallet ErrNoSuchWallet
     | ErrGetTransactionNoSuchTransaction ErrNoSuchTransaction
-    deriving (Show, Eq)
-
--- | Indicates that the specified transaction hash is not found.
-newtype ErrNoSuchTransaction = ErrNoSuchTransaction (Hash "Tx")
     deriving (Show, Eq)
 
 -- | Indicates that the specified start time is later than the specified end
--- a/lib/core/src/Cardano/Wallet/DB.hs
+++ b/lib/core/src/Cardano/Wallet/DB.hs
@@ -29,9 +29,11 @@ module Cardano.Wallet.DB
     , gapSize
 
       -- * Errors
-    , ErrRemoveTx (..)
     , ErrNoSuchWallet(..)
     , ErrWalletAlreadyExists(..)
+    , ErrNoSuchTransaction (..)
+    , ErrRemoveTx (..)
+    , ErrPutLocalTxSubmission (..)
     ) where
 
 import Prelude
@@ -259,7 +261,7 @@ data DBLayer m s k = forall stm. (MonadIO stm, MonadFail stm) => DBLayer
         -> Hash "Tx"
         -> SealedTx
         -> SlotNo
-        -> ExceptT ErrNoSuchWallet stm ()
+        -> ExceptT ErrPutLocalTxSubmission stm ()
         -- ^ Add or update a transaction in the local submission pool with the
         -- most recent submission slot.
 
@@ -334,16 +336,23 @@ newtype ErrNoSuchWallet
     = ErrNoSuchWallet WalletId -- Wallet is gone or doesn't exist yet
     deriving (Eq, Show)
 
+-- | Can't add a transaction to the local tx submission pool.
+data ErrPutLocalTxSubmission
+    = ErrPutLocalTxSubmissionNoSuchWallet ErrNoSuchWallet
+    | ErrPutLocalTxSubmissionNoSuchTransaction ErrNoSuchTransaction
+    deriving (Eq, Show)
+
 -- | Can't remove pending or expired transaction.
 data ErrRemoveTx
     = ErrRemoveTxNoSuchWallet ErrNoSuchWallet
-    | ErrRemoveTxNoSuchTransaction (Hash "Tx")
+    | ErrRemoveTxNoSuchTransaction ErrNoSuchTransaction
     | ErrRemoveTxAlreadyInLedger (Hash "Tx")
     deriving (Eq, Show)
 
--- | Can't perform given operation because there's no transaction
-newtype ErrNoSuchTransaction
-    = ErrNoSuchTransaction (Hash "Tx")
+-- | Indicates that the specified transaction hash is not found in the
+-- transaction history of the given wallet.
+data ErrNoSuchTransaction
+    = ErrNoSuchTransaction WalletId (Hash "Tx")
     deriving (Eq, Show)
 
 -- | Forbidden operation was executed on an already existing wallet
--- a/lib/core/src/Cardano/Wallet/DB.hs
+++ b/lib/core/src/Cardano/Wallet/DB.hs
@@ -1,4 +1,6 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE ExistentialQuantification #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs #-}
@@ -55,7 +57,15 @@ import Cardano.Wallet.Primitive.Types.Coin
 import Cardano.Wallet.Primitive.Types.Hash
     ( Hash )
 import Cardano.Wallet.Primitive.Types.Tx
-    ( TransactionInfo, Tx (..), TxMeta, TxStatus )
+    ( LocalTxSubmissionStatus
+    , SealedTx
+    , TransactionInfo
+    , Tx (..)
+    , TxMeta
+    , TxStatus
+    )
+import Control.DeepSeq
+    ( NFData )
 import Control.Monad.IO.Class
     ( MonadIO )
 import Control.Monad.Trans.Except
@@ -64,6 +74,8 @@ import Data.Quantity
     ( Quantity (..) )
 import Data.Word
     ( Word32, Word8 )
+import GHC.Generics
+    ( Generic )
 
 import qualified Data.List as L
 
@@ -242,6 +254,23 @@ data DBLayer m s k = forall stm. (MonadIO stm, MonadFail stm) => DBLayer
         --
         -- If the wallet doesn't exist, this operation returns an error.
 
+    , putLocalTxSubmission
+        :: PrimaryKey WalletId
+        -> Hash "Tx"
+        -> SealedTx
+        -> SlotNo
+        -> ExceptT ErrNoSuchWallet stm ()
+        -- ^ Add or update a transaction in the local submission pool with the
+        -- most recent submission slot.
+
+    , readLocalTxSubmissionPending
+        :: PrimaryKey WalletId
+        -> stm [LocalTxSubmissionStatus SealedTx]
+        -- ^ List all transactions from the local submission pool which are
+        -- still pending as of the latest checkpoint of the given wallet. The
+        -- slot numbers for first submission and most recent submission are
+        -- included.
+
     , updatePendingTxForExpiry
         :: PrimaryKey WalletId
         -> SlotNo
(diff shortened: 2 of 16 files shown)


lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
```diff
@@ -438,6 +439,17 @@ toTxHistory :: TransactionInfo -> (Tx, TxMeta)
...
 
+-- | Information about when a transaction was submitted to the local node.
+-- This is used for scheduling resubmissions.
+data LocalTxSubmissionStatus tx = LocalTxSubmissionStatus
+    { txId :: !(Hash "Tx")
+    , submittedTx :: !tx
+    , firstSubmission :: !SlotNo
+    -- ^ Time of first successful submission to the local node.
+    , latestSubmission :: !SlotNo
+    -- ^ Time of most recent resubmission attempt.
```
KtorZ, line 451: Would it be worth to add also the future schedule?

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1613,6 +1609,57 @@ forgetTx ctx wid tid = db & \DBLayer{..} -> do
...
+    => WalletId
+    -> ctx
+    -> IO a
+runLocalTxSubmissionPool wid ctx = db & \DBLayer{..} -> forever
+    watchNodeTip $ \bh -> bracketTracer trBracket $ do
+        -- TODO: ADP-764 limit pool check frequency to e.g. every 1000ms at most
+        pending <- atomically $ getLocalTxSubmissionPending (PrimaryKey wid)
+        -- re-submit those transactions, ignore errors
+        forM_ pending $ \st -> do
+            -- TODO: compare slots to schedule retries less often
```
KtorZ: Ah! That's exactly the bit I was looking for :no_mouth:; no schedule _yet_ that is.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -2240,21 +2254,52 @@ deletePendingOrExpiredTx wid tid = do
...
             ((TxMetaStatus <-. [W.Pending, W.Expired]):filt)
 
+listPendingLocalTxSubmissionQuery
+    :: W.WalletId
+    -> SqlPersistT IO [(W.SlotNo, LocalTxSubmission)]
+listPendingLocalTxSubmissionQuery wid = fmap unRaw <$> rawSql query params
+  where
+    query = "SELECT meta.slot,?? FROM (select tx_id, status, slot, MAX(slot) AS latest_slot FROM tx_meta WHERE wallet_id=? GROUP by tx_id) AS meta INNER JOIN local_tx_submission ON meta.tx_id=local_tx_submission.tx_id WHERE status=? AND slot=latest_slot"
+    params = [toPersistValue wid, toPersistValue W.Pending]
+    unRaw (Single sl, Entity _ tx) = (sl, tx)
```
KtorZ, line 2299: I'd love a comment which clarifies what is expected from this function. Looking at the query, I assume it's meant to return all most recent submissions for pending transactions (which is sort of hinted by the name already, but some clarifications would be much appreciated :pray: )

I actually went to see that function back again because I was getting a bit confused by the logic on the wallet side, thinking that we might be re-submitting transactions that are _in-ledger_ (which isn't a major issue per se, but is however useless). Now I see that this actually filters transactions by status, only considering _pending_ ones. Seems good.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -2240,21 +2254,52 @@ deletePendingOrExpiredTx wid tid = do
         Nothing -> fromIntegral <$> deleteWhereCount
             ((TxMetaStatus <-. [W.Pending, W.Expired]):filt)
 
+listPendingLocalTxSubmissionQuery
+    :: W.WalletId
+    -> SqlPersistT IO [(W.SlotNo, LocalTxSubmission)]
+listPendingLocalTxSubmissionQuery wid = fmap unRaw <$> rawSql query params
+  where
+    query = "SELECT meta.slot,?? FROM (select tx_id, status, slot, MAX(slot) AS latest_slot FROM tx_meta WHERE wallet_id=? GROUP by tx_id) AS meta INNER JOIN local_tx_submission ON meta.tx_id=local_tx_submission.tx_id WHERE status=? AND slot=latest_slot"
```
KtorZ: This `MAX(slot)` suggests that we are now storing multiple metadata per transaction? This raises an alarm in my head for some reasons I am not yet able to explain. I'd like to review the implication of this a bit more carefully (will do in a next review).

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
rvl: The schedule can be calculated pretty easily from this type.

lib/core/src/Cardano/Wallet.hs
rvl: We probably won't need anything more sophisticated that resend every _n_ slots.
(Maybe resending every slot is overkill, but perhaps not).

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
rvl: I'll add a comment to the query - it definitely needs explanation.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
rvl: Oops, this check isn't needed. There is only ever one TxMeta per txid per wallet. In the case of rollback, the status field is mutated.

Anviking:
I came across [this code](https://github.com/input-output-hk/cardano-wallet/blob/931c504eebf66e526c9999e139190adbf22044dc/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs#L436-L467), and wondered: how should tx retrying work across era-boundaries?

IIRC, the node has some way of converting past-era txs to new-era txs? If so I guess we should simply try to decode the `SealedTx` as any era tx, not necessarily just the current one.

lib/core/src/Cardano/Wallet/DB.hs
```diff
@@ -330,8 +358,8 @@ newtype ErrWalletAlreadyExists
 -- functions like 'enqueueCheckpoint' needs to be associated to a corresponding
 -- wallet. Some other may not because they are information valid for all wallets
 -- (like for instance, the last known network tip).
-newtype PrimaryKey key = PrimaryKey key
-    deriving (Show, Eq, Ord)
+newtype PrimaryKey key = PrimaryKey { unPrimaryKey :: key }
+    deriving (Show, Eq, Ord, Generic, NFData)
```
KtorZ, line 372: side-note: I am not sure this `PrimaryKey` type is any useful nowadays. Not the point of this PR, but we could perhaps leave a TODO / FIXME note about removing this altogether, it just creates noise in many places.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1517,6 +1534,7 @@ newDBLayerWith cacheBehavior tr ti SqliteContext{runQuery} = do
                     Nothing -> pure errNoSuchTransaction
                     Just _ -> do
                         count <- deletePendingOrExpiredTx wid tid
+                        deleteLocalTxSubmission wid [TxId tid]
```
KtorZ: Wouldn't be possible maybe to handle that using database constraints and have a cascade delete referencing the TxMeta table? There can't be a local submission without metadata, so if we delete the metadata, it makes sense to delete the tx submission in cascade and we don't have to worry about doing it manually :thinking: ?

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1623,6 +1632,63 @@ forgetTx ctx wid tid = db & \DBLayer{..} -> do
   where
     db = ctx ^. dbLayer @s @k
 
+-- | Given a LocalTxSubmission record, calculate the slot when it should be
+-- retried next.
+--
+-- The current implementation is really basic. Retry every 10 slots.
+scheduleLocalTxSubmission :: LocalTxSubmissionStatus tx -> SlotNo
+scheduleLocalTxSubmission st = (st ^. #latestSubmission) + 10
```
KtorZ: So that's about half the block frequency on mainnet, which would means effectively retrying on every block. Maybe a little _excessive_ for exchanges & al, but I see no harm doing this otherwise.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1623,6 +1632,63 @@ forgetTx ctx wid tid = db & \DBLayer{..} -> do
...
+        -- Re-submit transactions due, ignore errors
+        forM_ (filter (isScheduled sl) pending) $ \st -> do
+            res <- runExceptT $ postTx nw (st ^. #submittedTx)
+            traceWith tr (MsgRetryPostTxResult (st ^. #txId) res)
+            atomically $ putLocalTxSubmission
+                (PrimaryKey wid)
+                (st ^. #txId)
+                (st ^. #submittedTx)
+                sl
+    watchNodeTip submitPending
```

lib/core/src/Cardano/Wallet/DB.hs
rvl: Ah yes - that's what I thought too.

lib/core/src/Cardano/Wallet/DB/MVar.hs
```diff
@@ -218,14 +238,15 @@ alterDB
 alterDB convertErr db op = modifyMVar db (bubble . op)
   where
     bubble (Left e, db') = case convertErr e of
-        Just e' -> pure (db', Left e')
+        Just e' -> pure (force db', Left e')
         Nothing -> throwIO $ MVarDBError e
-    bubble (Right a, db') = pure (db', Right a)
+    bubble (Right a, db') = pure (force db', Right a)
```
KtorZ, line 243: :+1:

lib/core/src/Cardano/Wallet/DB/Model.hs
```diff
@@ -322,6 +346,23 @@ mRollbackTo wid requested db@(Database wallets txs) = case Map.lookup wid wallet
...
+        , submittedTxs = Map.filterWithKey (const . ((<= point) . txHeight)) $
+            submittedTxs wal
+        }
+      where
+        cp = mostRecentCheckpoint wal
+        tipHeight = maybe 0 (getQuantity . cpHeight) cp
+        point = Quantity (tipHeight - getQuantity epochStability)
+        cpHeight = view (#currentTip . #blockHeight)
+        txHeight txid = maybe (Quantity tipHeight) (view #blockHeight) $
+            Map.lookup txid $ txHistory wal
```
KtorZ: That isn't how the pruning works currently with SQLite. This isn't caught by the SM testing however because we never generate prune commands, but if we were I suspect that it'd fail :sweat_smile: ... 

We do prune the SQLite database using the `sparseCheckpoints` function, to reduce the number of past checkpoints we have archived. Pruning should likely not have any impact on submitted transactions however; and although interesting for testing & al, I think defining a model version for `prune` is out of the scope of this PR :thinking:  ?

lib/core/test/unit/Cardano/Wallet/DB/StateMachine.hs
```diff
@@ -650,10 +695,17 @@ generatorWithWid wids =
             <*> genSortOrder
             <*> genRange
             <*> arbitrary
-    , declareGenerator "RemovePendingTx" 4
-        $ RemovePendingTx <$> genId <*> arbitrary
+    -- TODO: Implement mGetTx
+    -- , declareGenerator "GetTx" 3
+    --     $ GetTx <$> genId <*> arbitrary
```
KtorZ: Old TODO which can now be done :thinking: ?

lib/core/src/Cardano/Wallet.hs
rvl: Fixed in #2611

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
rvl: Yes - good idea.

lib/core/src/Cardano/Wallet/DB/Model.hs
rvl: Ok yes, I have reverted that and intend to address it in another PR.

lib/core/test/unit/Cardano/Wallet/DB/StateMachine.hs
rvl: Still a TODO - we ought to have mGetTx in the qsm tests. I'll add it to the next PR.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
rvl: Done - it makes things simpler.

lib/core/src/Cardano/Wallet/DB.hs
rvl: I'll sort this in the next PR.

rvl:
> How should tx retrying work across era-boundaries?

In theory, I believe we don't need to do anything special, @Anviking, because we can submit to the node with type `GenTx (CardanoBlock crypto)`. In practice, we shall need to fix/remove the usage of `unsealShelleyTx` for this corner case. This is going to happen anyway.

rvl:
I believe all review comments so far have been addressed.

rvl:
bors try

KtorZ (approved):

PR #2571, opened 2021-03-19
Log to stderr for (>= Notice) and tweak severities

Anviking:
# Issue Number

ADP-630

# Overview

- [x] Log `>= Warning` to stderr and `< Warning` to stdout, such that they can be redirected to separate files
- [x] Tweaked the severities some messages

# Comments

- We could allow logging config via CLI, but do we really need it?
- I had a somewhat handwritten aggregation logic draft for the pool worker, but I didn't add include it in this PR. Maybe we should use iohk-monitoring for it instead.

### DIFF
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -1580,9 +1580,8 @@ data LogOutput
     = LogToStdStreams Severity
     -- ^ Log to console, with the given minimum 'Severity'.
     --
-    -- Logs of Notice or higher severity will be output to stderr. Info or lower
-    -- severity logs will be output to stdout.
-    --
+    -- Logs of Warning or higher severity will be output to stderr. Notice or
+    -- lower severity logs will be output to stdout.
     | LogToFile FilePath Severity
     deriving (Eq, Show)
 
@@ -1598,8 +1597,8 @@ mkScribe (LogToFile path sev) = pure $ ScribeDefinition
     , scRotation = Nothing
     }
 mkScribe (LogToStdStreams sev) =
-    [ mkScribe' (max Notice sev, maxBound, StderrSK)
-    , mkScribe' (sev, pred Notice, StdoutSK)
+    [ mkScribe' (max Warning sev, maxBound, StderrSK)
+    , mkScribe' (sev, pred Warning, StdoutSK)
     ]
   where
     mkScribe' (minSev, maxSev, kind) = ScribeDefinition
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -1577,22 +1577,18 @@ data Verbosity
     deriving (Eq, Show)
 
 data LogOutput
-    = LogToStdout Severity
+    = LogToStdStreams Severity
+    -- ^ Log to console, with the given minimum 'Severity'.
+    --
+    -- Logs of Notice or higher severity will be output to stderr. Info or lower
+    -- severity logs will be output to stdout.
+    --
     | LogToFile FilePath Severity
     deriving (Eq, Show)
 
 
-mkScribe :: LogOutput -> ScribeDefinition
-mkScribe (LogToStdout sev) = ScribeDefinition
-    { scName = "text"
-    , scFormat = ScText
-    , scKind = StdoutSK
-    , scMinSev = sev
-    , scMaxSev = Critical
-    , scPrivacy = ScPublic
-    , scRotation = Nothing
-    }
-mkScribe (LogToFile path sev) = ScribeDefinition
+mkScribe :: LogOutput -> [ScribeDefinition]
+mkScribe (LogToFile path sev) = pure $ ScribeDefinition
     { scName = T.pack path
     , scFormat = ScText
     , scKind = FileSK
@@ -1601,10 +1597,25 @@ mkScribe (LogToFile path sev) = ScribeDefinition
     , scPrivacy = ScPublic
     , scRotation = Nothing
     }
+mkScribe (LogToStdStreams sev) =
+    [ mkScribe' (max Notice sev, maxBound, StderrSK)
+    , mkScribe' (sev, pred Notice, StdoutSK)
+    ]
+  where
+    mkScribe' (minSev, maxSev, kind) = ScribeDefinition
+        { scName = "text"
+        , scFormat = ScText
+        , scKind = kind
+        , scMinSev = minSev
+        , scMaxSev = maxSev
+        , scPrivacy = ScPublic
+        , scRotation = Nothing
+        }
 
-mkScribeId :: LogOutput -> ScribeId
-mkScribeId (LogToStdout _) = "StdoutSK::text"
-mkScribeId (LogToFile file _) = T.pack $ "FileSK::" <> file
+
+mkScribeId :: LogOutput -> [ScribeId]
+mkScribeId (LogToStdStreams _) = ["StdoutSK::text", "StderrSK::text"]
+mkScribeId (LogToFile file _) = pure $ T.pack $ "FileSK::" <> file
 
 getPrometheusURL :: IO (Maybe (String, Port "Prometheus"))
 getPrometheusURL = do
@@ -1649,8 +1660,8 @@ initTracer loggerName outputs = do
         c <- defaultConfigStdout
         CM.setSetupBackends c [CM.KatipBK, CM.AggregationBK, CM.EKGViewBK, CM.EditorBK]
         CM.setDefaultBackends c [CM.KatipBK]
-        CM.setSetupScribes c $ map mkScribe outputs
-        CM.setDefaultScribes c $ map mkScribeId outputs
+        CM.setSetupScribes c $ outputs >>= mkScribe
+        CM.setDefaultScribes c $ outputs >>= mkScribeId
         CM.setBackends c "test-cluster.metrics" (Just [CM.EKGViewBK])
         CM.setBackends c "cardano-wallet.metrics" (Just [CM.EKGViewBK])
         forM_ ekgHP $ \(h, p) -> do
--- a/lib/cli/src/Cardano/CLI.hs
+++ b/lib/cli/src/Cardano/CLI.hs
@@ -1597,10 +1597,11 @@ mkScribe (LogToFile path sev) = pure $ ScribeDefinition
     , scRotation = Nothing
     }
 mkScribe (LogToStdStreams sev) =
-    [ mkScribe' (max Warning sev, maxBound, StderrSK)
-    , mkScribe' (sev, pred Warning, StdoutSK)
+    [ mkScribe' (max errMin sev, maxBound, StderrSK)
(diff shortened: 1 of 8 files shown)


Anviking:
bors try

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1577,22 +1577,15 @@ data Verbosity
     deriving (Eq, Show)
 
 data LogOutput
-    = LogToStdout Severity
+    = LogToStdouterr Severity
+    -- ^ Log [sev .. Critical] to either stderr or stdout.
+    -- To stderr for (>= Notice), and to stdout for (< Notice).
```
rvl:
```suggestion
    = LogToStdStreams Severity Severity
    -- ^ Log to console, with the given minimum 'Severity' for @stdout@ messages
    -- and minimum 'Severity' for @stderr@ messages.
```

Here's a better name, and we may as well make parameters for both severities - less explaining to do in comments.

rvl (commented):
I like it so far...

lib/cli/src/Cardano/CLI.hs
Anviking: Name — 👍 , but I don't think we should take parameters for both parameters. Then we'd need to add complexity to the call-sites (local-cluster, cardano-wallet, integration)

lib/core/src/Cardano/Wallet/TokenMetadata.hs
```diff
@@ -383,12 +383,12 @@ data TokenMetadataLog
 
 instance HasSeverityAnnotation TokenMetadataLog where
     getSeverityAnnotation = \case
-        MsgNotConfigured -> Notice
+        MsgNotConfigured -> Info
```
Anviking: This _would_ get drowned in the other Info logs. I wonder if there better ways of reminding the user that SMASH isn't configured...

- Print Notice log when listing pools (but maybe not)
- Provide some kind of wallet state in the HTTP headers of the response

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1577,22 +1577,18 @@ data Verbosity
     deriving (Eq, Show)
 
 data LogOutput
-    = LogToStdout Severity
+    = LogToStdStreams Severity
+    -- ^ Log to console, with the given minimum 'Severity'.
+    --
+    -- Logs of Notice or higher severity will be output to stderr. Info or lower
```
Anviking: If `Notice` is ["Normal but significant conditions"](https://github.com/input-output-hk/cardano-wallet/wiki/Logging-Guidelines), I guess one _could_ argue it should belong to stdout.

lib/core/src/Cardano/Wallet/TokenMetadata.hs
Anviking: Ah, this is the asset metadata. Anyway, I'll re-raise this to Notice.

lib/cli/src/Cardano/CLI.hs
Anviking: Ah, @rvl you said that too.

>  <= Notice on stdout 

Will change then.

lib/core/src/Network/Wai/Middleware/Logging.hs
```diff
@@ -257,8 +258,11 @@ instance ToText HandlerLog where
             path = T.decodeUtf8 $ rawPathInfo req
             query = T.decodeUtf8 $ rawQueryString req
         LogRequestBody ks body -> sanitize ks body
-        LogResponse time status -> mconcat [ code, " ", text, " in ", tsec ]
+        LogResponse time req status ->
+            mconcat [ method, " ", path, " ", code, " ", text, " in ", tsec ]
```
Anviking, line 262: Example in integration tests:
Skärmavbild 2021-03-22 kl  14 12 48

lib/cli/src/Cardano/CLI.hs
```diff
@@ -1601,10 +1596,25 @@ mkScribe (LogToFile path sev) = ScribeDefinition
     , scPrivacy = ScPublic
     , scRotation = Nothing
     }
+mkScribe (LogToStdStreams sev) =
+    [ mkScribe' (max Warning sev, maxBound, StderrSK)
+    , mkScribe' (sev, pred Warning, StdoutSK)
+    ]
+  where
```
rvl:
```suggestion
    [ mkScribe' (max errMin sev, maxBound, StderrSK)
    , mkScribe' (sev, pred errMin, StdoutSK)
    ]
  where
    errMin = Warning
```

lib/core/src/Network/Ntp.hs
```diff
@@ -139,20 +139,26 @@ instance ToText NtpTrace where
...
         NtpTraceLookupsFails -> Notice
-        NtpTraceClientStartQuery -> Info
+        NtpTraceClientStartQuery -> Debug
         NtpTraceNoLocalAddr -> Notice
-        NtpTraceResult _ -> Info
+        NtpTraceResult (NtpDrift micro)
+            | abs micro < (500*ms)  -> Debug   -- Not sure what limits actually
+            | abs micro < (1000*ms) -> Notice  -- matter, but these seem
+            | otherwise             -> Warning -- reasonable.
+        NtpTraceResult _ -> Debug
```
rvl, line 153: How much drift is tolerable depends on slot length. :smile_cat:
Anything like >0.5s would suggest that the user does not have automatic time updates enabled, which they ought to fix.

lib/core/src/Network/Wai/Middleware/Logging.hs
```diff
@@ -285,22 +289,29 @@ sanitize keys bytes = encode' $ case decode' bytes of
...
+        LogResponseBody{} -> Confidential
+        LogRequestFinish{} -> Public
 
 instance HasSeverityAnnotation HandlerLog where
     getSeverityAnnotation msg = case msg of
         LogRequestStart -> Debug
         LogRequest _ -> Info
         LogRequestBody _ _ -> Debug
-        LogResponse _ status ->
+        LogResponse t _ status -> max (severityFromRequestTime t) $
```
rvl, line 304: I like the idea of warning the user when values are too wrong.
But it may make the logs confusing if log messages change severity and appear or don't appear based on a time value.

My suggestion is make severity constant w.r.t. the metric and then add a trace transformer on top which monitors for abnormal values and then logs messages separately at a higher level. Should I push an example to the branch?

rvl (approved):
Great

lib/core/src/Cardano/Wallet/TokenMetadata.hs
rvl: This is actually a log message for Token Metadata.
It will be logged exactly once at startup.
So I think Notice is the correct level, right?

lib/core/src/Network/Wai/Middleware/Logging.hs
Anviking: Something like 1ab511e ?

Anviking: I'm not sure I like it though... Seems messier with more similar-looking logs.

Can we not use `max (severityFromRequestTime) $` ?

Anviking: I'll drop this (1ab511e), and merge with `max (severityFromRequestTime) $` in the meantime. Hope you don't mind.

```patch
diff --git a/lib/core/src/Network/Wai/Middleware/Logging.hs b/lib/core/src/Network/Wai/Middleware/Logging.hs
index cf98003eb..613804fa6 100644
--- a/lib/core/src/Network/Wai/Middleware/Logging.hs
+++ b/lib/core/src/Network/Wai/Middleware/Logging.hs
@@ -38,7 +38,7 @@ import Control.Applicative
 import Control.Arrow
     ( second )
 import Control.Tracer
-    ( Tracer, contramap, traceWith )
+    ( Tracer, contramap, contramapM, traceWith )
 import Data.Aeson
     ( FromJSON (..), ToJSON (..), Value (..) )
 import Data.ByteString
@@ -82,7 +82,7 @@ withApiLogger
     -> Middleware
 withApiLogger t0 settings app req0 sendResponse = do
     rid <- nextRequestId settings
-    let t = contramap (ApiLog rid) t0
+    let t = classifyReqTime $ contramap (ApiLog rid) t0
     traceWith t LogRequestStart
     start <- getCurrentTime
     (req, reqBody) <- getRequestBody req0
@@ -109,6 +109,18 @@ withApiLogger t0 settings app req0 sendResponse = do
         traceWith t (LogResponse time req status)
         traceWith t (LogResponseBody body)

+
+    transform :: Monad m => (a -> [a]) -> Tracer m a -> Tracer m a
+    transform multiply tr = flip contramapM tr $ \msg -> do
+        mapM_ (traceWith tr) (multiply msg)
+        return msg
+
+    classifyReqTime = transform f
+      where
+        f (LogResponse t req status) = [LogResponseSlow t req status]
+        f _ = []
+
+
 -- | API logger settings
 data ApiLoggerSettings = ApiLoggerSettings
     { _obfuscateKeys :: Request -> [Text]
@@ -245,6 +257,7 @@ data HandlerLog
     | LogRequestBody [Text] ByteString
     -- ^ Request content, with list of sensitive json keys.
     | LogResponse NominalDiffTime Request (Maybe Status)
+    | LogResponseSlow NominalDiffTime Request (Maybe Status)
     | LogResponseBody ByteString
     | LogRequestFinish
     deriving (Generic, Show)
@@ -258,6 +271,14 @@ instance ToText HandlerLog where
             path = T.decodeUtf8 $ rawPathInfo req
             query = T.decodeUtf8 $ rawQueryString req
         LogRequestBody ks body -> sanitize ks body
+        LogResponseSlow time req status ->
+            mconcat [ "Slow response ", method, " ", path, " ", code, " ", text, " in ", tsec ]
+          where
+            method = T.decodeUtf8 $ requestMethod req
+            path = T.decodeUtf8 $ rawPathInfo req
+            code = maybe "???" (toText . statusCode) status
+            text = maybe "Status Unknown" (T.decodeUtf8 . statusMessage) status
+            tsec = T.pack $ show time
         LogResponse time req status ->
             mconcat [ method, " ", path, " ", code, " ", text, " in ", tsec ]
           where
@@ -293,6 +314,7 @@ instance HasPrivacyAnnotation HandlerLog where
         LogRequest{} -> Public
         LogRequestBody{} -> Confidential
         LogResponse{} -> Public
+        LogResponseSlow{} -> Public
         LogResponseBody{} -> Confidential
         LogRequestFinish{} -> Public

@@ -301,11 +323,12 @@ instance HasSeverityAnnotation HandlerLog where
         LogRequestStart -> Debug
         LogRequest _ -> Info
         LogRequestBody _ _ -> Debug
-        LogResponse t _ status -> max (severityFromRequestTime t) $
+        LogResponse _ _ status ->
             case statusCode <$> status of
                 Just s | s == 503 -> Warning
                 Just s | s >= 500 -> Error
                 _ -> Info
+        LogResponseSlow t _ _ -> severityFromRequestTime t
         LogResponseBody _ -> Debug
         LogRequestFinish -> Debug

```

PR #2573, opened 2021-03-23
Create completely new TVars in LoggingSpec

Anviking:
# Issue Number

#2368 

# Overview

- Use different TVars in LoggingSpec. 
    - It seems the flakiness is caused by different test runs interfering with each other. Using a new TVar each run should hopefully fix it.

# Comments

### DIFF
--- a/lib/core/test/unit/Network/Wai/Middleware/LoggingSpec.hs
+++ b/lib/core/test/unit/Network/Wai/Middleware/LoggingSpec.hs
@@ -91,7 +91,7 @@ import Servant
 import Servant.Server
     ( Handler )
 import Test.Hspec
-    ( Spec, after, afterAll, beforeAll, describe, it, shouldBe, shouldContain )
+    ( Spec, after, before, describe, it, shouldBe, shouldContain )
 import Test.QuickCheck
     ( Arbitrary (..), choose, counterexample, property, withMaxSuccess )
 import Test.QuickCheck.Monadic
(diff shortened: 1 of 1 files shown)


KtorZ (approved):
I see nothing wrong with that, that's actually cheap and may help with the flakyness indeed. :+1: 
Those tests don't run in parallel though :thinking:

Anviking:
> Those tests don't run in parallel though

There is this comment talking about this:

```
-- | Remove any partial logs which have come from the previous property test
-- run. These can occur because once the requests have completed, the property
-- finishes, but the response side may still be finishing writing its log
-- messages.
skipPrevLogs :: [ApiLog] -> [ApiLog]
```

(and actually that function shouldn't be needed now - removing)

rvl (commented):
What this actually does is start a new HTTP server for each hspec test case. It's not only making a new empty TVar.
That will reduce the flakiness ... why didn't I think of that? :smile: 
There are also two hspec test cases using QuickCheck properties.
In these, the HTTP server and TVar will be re-used for all QuickCheck examples. So I expect to see flaky test failures happen there, for the same reason as before.
I think the best way to make these tests stable is to remove the use `property` / `monadicIO` in this spec.

Anviking:
> There are also two hspec test cases using QuickCheck properties.
In these, the HTTP server and TVar will be re-used for all QuickCheck examples. So I expect to see flaky test failures happen there, for the same reason as before.
I think the best way to make these tests stable is to remove the use property / monadicIO in this spec.

@rvl ah, good point.

I think this is a quick but fine workaround: https://github.com/input-output-hk/cardano-wallet/pull/2580

PR #2574, opened 2021-03-23
Decrease coverage requirement in UTxOIndexSpec

Anviking:
# Issue Number

#2575

# Overview

- [x] Decrease all `coverage 80` to `coverage 70` inside UTxOIndexSpec

# Comments


rvl (commented):
Do you happen to know why `cover 80` was chosen before, and why it's now too much?

Anviking:
I simply saw the #2575 failure, and thought it was an innocent slightly-to-strict `cover`-limit from when the tests were introduced. cc @jonathanknowles ?

jonathanknowles:
> Do you happen to know why `cover 80` was chosen before, and why it's now too much?

When we originally set these coverage requirements, the average real coverage was consistently higher. (I recall running these tests hundreds of times to make sure.)

However, if the underlying generators have recently been changed, then the average coverage could have dropped, causing the tests to sporadically fail.

Nevertheless, `70` should still be more than adequate here.

jonathanknowles (approved):
Hi @Anviking!

This looks fine to me.

See https://github.com/input-output-hk/cardano-wallet/pull/2574#issuecomment-809863535

Anviking:
bors r+
pls

PR #2533, opened 2021-02-22
Bump to GHC 8.10.4

Anviking:
# Issue Number

ADP-453

# Overview

- [x] Bump to GHC-8.10.4
- [x] Use OnDeleteCascade in new persistent version, instead of our workaround in the fork

##### Todo:

- [x] Remove `allow-newer`
- [x] core:unit tests pass locally
- [x] Look closer at what why we needed the persistent fork
   - Seems we can now simply use OnDeleteCascade
- [x] Integration tests pass locally
- [x] Investigate cardano-wallet:unit failures
- [x] Bump haskell.nix
- [x] Figure out how to use weeder -> seems green at least
- [x] Get CI green
- [x] Move to snapshot in cardano-haskell repo
- [x] Merge that first

# Comments

[Hydra jobset](https://hydra.iohk.io/jobset/Cardano/cardano-wallet-pr-2533)

PR #2576, opened 2021-03-24
Fix nightly benchmarks

rvl:
### Issue Number

ADP-804

### Overview

Nightly restore bench is still failing.

1. Restore bench: Increase cardano-node startup time allowance

   If the cardano-node database is shared between bench runs, it may need to rebuild the ledger, etc, at startup. So increase the time available for that.

2. Latency bench: Fix working directory

   The meaning of `$src` seems to have changed since the Haskell.nix bump in PR #2533.

3. DB Bench: I broke the criterion env setup - fixed now.

### DIFF
--- a/lib/shelley/bench/Restore.hs
+++ b/lib/shelley/bench/Restore.hs
@@ -764,7 +764,7 @@ waitForNodeSync
     => Tracer IO (BenchmarkLog n)
     -> NetworkLayer IO Block
     -> IO SlotNo
-waitForNodeSync tr nw = loop 10
+waitForNodeSync tr nw = loop 120 -- allow 30 minutes for first tip
   where
     loop :: Int -> IO SlotNo
     loop retries = do
@@ -788,7 +788,7 @@ waitForNodeSync tr nw = loop 10
                  traceWith tr $ MsgRetryShortly delay
                  threadDelay delay
                  loop (retries - 1)
-            else throwString "Gave up waitForNodeSync"
+            else throwString "Gave up in waitForNodeSync, waiting a tip"
 
 data BenchmarkLog (n :: NetworkDiscriminant)
     = MsgNodeTipTick BlockHeader SyncProgress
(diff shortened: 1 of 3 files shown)


lib/shelley/bench/Restore.hs
```diff
@@ -764,7 +764,7 @@ waitForNodeSync
     => Tracer IO (BenchmarkLog n)
     -> NetworkLayer IO Block
     -> IO SlotNo
-waitForNodeSync tr nw = loop 10
+waitForNodeSync tr nw = loop 120
```
KtorZ: I've seen it took up to 15 minutes locally when switching to newer version of cardano-node. Maybe good to be more conservative here?

rvl: That number works out to 30 minutes, which should be conservative enough, I hope. I added a comment to clarify. It would be clearer if this function used `Control.Retry.retrying` but it doesn't.

piotr-iohk: Hopefully 30 mins is enough (https://input-output-rnd.slack.com/archives/C7PJ29FME/p1616507642042100?thread_ts=1616507577.042000&cid=C7PJ29FME)

piotr-iohk (approved):

lib/shelley/bench/Restore.hs
rvl: I think on a dedicated linux benchmark machine it should be fine.

rvl:
The restore bench works a little better now.
There are some missing wallet "errors" logged after restoration finishes, due to the clean up code - these could be fixed later but are not harmful.

However the list transactions test after `0.4-percent-seq` fails:
```
restore: Heap exhausted;
restore: Current maximum heap size is 8589934592 bytes (8192 MB).
restore: Use `+RTS -M<size>' to increase it.
```
It's a lot of transactions to list, but 8GB seems excessive.

rvl:
bors r+

please

PR #2577, opened 2021-03-24
bump actions/cache to v2.1.4 in e2e workflows

piotr-iohk:
# Issue Number

# Overview

- [ ] I have bumped actions/cache to fix caching issue on Windows workflow

# Comments

Currently caching of node-db in Windows workflow [fails](https://github.com/input-output-hk/cardano-wallet/runs/2179176487?check_suite_focus=true) and as a result run takes ~3.5h instead of modest ~1h.
Apparently the issue(https://github.com/actions/cache/issues/468) is fixed in actions/cache@v2.1.4

### DIFF
--- a/.github/workflows/e2e-docker.yml
+++ b/.github/workflows/e2e-docker.yml
@@ -42,7 +42,7 @@ jobs:
 
     - name: Cache node db
       id: cache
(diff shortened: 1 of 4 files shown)


rvl (approved):
... OK ...

PR #2580, opened 2021-03-25
Further prevent flakiness of LoggingSpec properties

Anviking:
# Issue Number

#2368 

# Overview

- [x] Use `withMaxSuccess 1` to ensure each `ctx` is only used for one test.

# Comments

- Quick and I believe fine workaround for https://github.com/input-output-hk/cardano-wallet/pull/2573#pullrequestreview-619348957


rvl (approved):
Instead of limiting the property to 1 example (fine, but not nice from a flakiness perspective), we could (a) remove the use of QuickCheck here (see reverted commit), or (b) run a new server for each property example (see latest commit).

Anviking:
Fine, that works too 😸  — thanks
bors r+

Anviking:
bors r+

Pls

PR #2581, opened 2021-03-25
Cleanup unused NetworkLayer errors and logic

Anviking:
# Issue Number

ADP-630

# Overview

- [x] Remove unused `ErrNetworkUnavailable `, `ErrGetBlock`, `ErrGetTxParameters`
    - Does call into question what would happen if the node were to disconnect etc...
- [x] Remove `RetryImmediately` and `RetryLater` from `follow`, which were needed only for jormungandr 

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/Network.hs
+++ b/lib/core/src/Cardano/Wallet/Network.hs
@@ -224,26 +224,11 @@ mapCursor fn = \case
 -- | @FollowAction@ enables the callback of @follow@ to signal if the
 -- chain-following should @ExitWith@, @Continue@, or if the current callback
 -- should be forgotten and retried (@Retry@).
---
--- NOTE: @Retry@ is needed to handle data-races in
--- 'Cardano.Pool.Jormungandr.Metrics', where it is essensial that we fetch the
--- stake distribution while the node-tip
---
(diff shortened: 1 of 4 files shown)


lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -334,28 +288,24 @@ follow nl tr cps yield header =
...
 
-    step :: Int -> Bool -> Cursor -> IO FollowExit
-    step delay hasRolledForward cursor = runExceptT (nextBlocks nl cursor) >>= \case
-        Left e -> do
-            traceWith tr $ MsgNextBlockFailed e
-            sleep (retryDelay delay) hasRolledForward cursor
-
-        Right AwaitReply -> do
+    step :: Bool -> Cursor -> IO FollowExit
+    step hasRolledForward cursor = nextBlocks nl cursor >>= \case
```
Anviking, line 292: delay argument not needed anymore

rvl (approved):
Great, I like clean ups.

`FollowAction(ExitWith)` and `FollowExit(FollowFailure, FollowDone)` could also be removed, but that's a PR for another day...

rvl:
bors r+

Please

PR #2583, opened 2021-03-26
More concise Follow-logs using aggregation

Anviking:
# Issue Number

ADP-630

# Overview

**1. Invert control to empower follow** 

The idea is: the more we can concentrate information in one place (follow), the better we can produce good logs.

- [x] Use `FollowLog tr` where `tr` can be `StakePoolLog` or `WalletWorkerLog`, instead of like now, having them both wrap `FollowLog`.
- [x] Move rollback-logic from call-sites to `follow`
   - [x] Ensure this commit _perfectly_ preserves previous behaviour -> should be pretty close

**2. Tweak logging**
- [x] Lower common messages to `Debug`
- [x] Add some kind of aggregation (I think doing manually is easier than learning how to use iohk-monitoring-framework for it, and maybe more powerful)
    - [x] Improve 
    - [x] Move `syncProgress` calculation to `follow` 
    - [x] Make sure we are strict where we need to, and don't create massive thunks

# Comments

- Based on PR #2581,  #2586 and #2615 branches - merge them first.

## Lost network connection + 5 min wait:

Skärmavbild 2021-04-16 kl  16 58 45

## On reconnection

Skärmavbild 2021-04-20 kl  16 03 52

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -295,7 +295,7 @@ import Cardano.Wallet.Primitive.Slotting
     , unsafeExtendSafeZone
     )
 import Cardano.Wallet.Primitive.SyncProgress
-    ( SyncProgress, SyncTolerance (..), syncProgress )
+    ( SyncProgress, SyncTolerance (..) )
 import Cardano.Wallet.Primitive.Types
     ( ActiveSlotCoefficient (..)
     , Block (..)
@@ -442,8 +442,6 @@ import Fmt
     ( blockListF, pretty, (+|), (+||), (|+), (||+) )
 import GHC.Generics
     ( Generic )
-import GHC.Stack
-    ( HasCallStack )
 import Safe
     ( lastMay )
 import Statistics.Quantile
@@ -707,24 +705,15 @@ readWallet ctx wid = db & \DBLayer{..} -> mapExceptT atomically $ do
     db = ctx ^. dbLayer @IO @s @k
 
 walletSyncProgress
-    :: forall ctx s.
-        ( HasGenesisData ctx
-        , HasNetworkLayer IO ctx
-        , HasCallStack
-        )
+    :: forall ctx s. HasNetworkLayer IO ctx
     => ctx
     -> Wallet s
     -> IO SyncProgress
 walletSyncProgress ctx w = do
-    let tip = currentTip w
-    syncProgress st ti tip =<< currentRelativeTime ti
+    let tip = view #slotNo $ currentTip w
+    syncProgress nl tip
   where
-    (_, _, st) = ctx ^. genesisData
-
-    ti :: TimeInterpreter IO
-    ti = neverFails
-            "walletSyncProgress only converts times at the tip or before"
-            (timeInterpreter $ ctx ^. networkLayer)
+    nl = ctx ^. networkLayer
 
 -- | Update a wallet's metadata with the given update function.
 updateWallet
@@ -786,7 +775,6 @@ restoreWallet
         ( HasNetworkLayer IO ctx
         , HasDBLayer IO s k ctx
         , HasLogger WalletWorkerLog ctx
-        , HasGenesisData ctx
         , IsOurs s Address
         , IsOurs s RewardAccount
         )
@@ -834,9 +822,8 @@ rollbackBlocks ctx tr wid point = db & \DBLayer{..} -> do
 -- transaction history and corresponding metadata.
 restoreBlocks
     :: forall ctx s k.
-        ( HasDBLayer IO s k ctx
+        ( HasDBLayer s k ctx
         , HasNetworkLayer IO ctx
-        , HasGenesisData ctx
         , IsOurs s Address
         , IsOurs s RewardAccount
         )
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -210,7 +210,7 @@ import Cardano.Wallet.Network
     ( ErrGetAccountBalance (..)
     , ErrPostTx (..)
     , FollowAction (..)
-    , FollowExit (..)
+    , FollowExceptionRecovery (..)
     , FollowLog (..)
     , NetworkLayer (..)
     , follow
@@ -782,18 +782,11 @@ restoreWallet
     -> WalletId
     -> ExceptT ErrNoSuchWallet IO ()
 restoreWallet ctx wid = db & \DBLayer{..} -> do
-    cps <- liftIO $ atomically $ listCheckpoints wid
+    let readCps = liftIO $ atomically $ listCheckpoints wid
     let forward bs h innerTr = run $ do
             restoreBlocks @ctx @s @k ctx innerTr wid bs h
-    liftIO (follow nw tr cps forward (view #header)) >>= \(innerTr, act) -> do
-        case act of
-            FollowFailure ->
-                restoreWallet @ctx @s @k ctx wid
-            FollowRollback point -> do
-                rollbackBlocks @ctx @s @k ctx innerTr wid point
-                restoreWallet @ctx @s @k ctx wid
-            FollowDone ->
-                pure ()
+    let backward = runExceptT . rollbackBlocks @ctx @s @k ctx wid
+    liftIO $ follow nw tr readCps forward backward RetryOnExceptions (view #header)
   where
     db = ctx ^. dbLayer @IO @s @k
     nw = ctx ^. networkLayer
@@ -807,14 +800,11 @@ restoreWallet ctx wid = db & \DBLayer{..} -> do
 rollbackBlocks
     :: forall ctx s k. (HasDBLayer IO s k ctx)
     => ctx
-    -> Tracer IO WalletFollowLog
     -> WalletId
     -> SlotNo
-    -> ExceptT ErrNoSuchWallet IO ()
-rollbackBlocks ctx tr wid point = db & \DBLayer{..} -> do
-    lift $ traceWith tr $ MsgTryingRollback point
-    point' <- mapExceptT atomically $ rollbackTo wid point
-    lift $ traceWith tr $ MsgRolledBack point'
+    -> ExceptT ErrNoSuchWallet IO SlotNo
+rollbackBlocks ctx wid point = db & \DBLayer{..} -> do
+    mapExceptT atomically $ rollbackTo wid point
   where
     db = ctx ^. dbLayer @IO @s @k
 
@@ -822,7 +812,7 @@ rollbackBlocks ctx tr wid point = db & \DBLayer{..} -> do
 -- transaction history and corresponding metadata.
 restoreBlocks
     :: forall ctx s k.
-        ( HasDBLayer s k ctx
+        ( HasDBLayer IO s k ctx
         , HasNetworkLayer IO ctx
         , IsOurs s Address
         , IsOurs s RewardAccount
@@ -892,11 +882,8 @@ restoreBlocks ctx tr wid blocks nodeTip = db & \DBLayer{..} -> mapExceptT atomic
     prune wid epochStability
 
     liftIO $ do
-        progress <- walletSyncProgress @ctx @s ctx (NE.last cps)
         traceWith tr $ MsgWalletMetadata meta
-        traceWith tr $ MsgSyncProgress progress
         traceWith tr $ MsgDiscoveredTxs txs
-        traceWith tr $ MsgTip localTip
         traceWith tr $ MsgBlocks blocks
         traceWith tr $ MsgDiscoveredTxsContent txs
   where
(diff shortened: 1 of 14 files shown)


lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -251,25 +263,80 @@ data FollowExit
...
+    loop tr
+  where
+    loop tr = do
+        cursor <- readCursor
+        -- TODO: Pool worker would trace MsgStartMonitoring here
+        (follow' nl tr cursor forward header) >>= \act -> do
+            case act of
+                FollowFailure ->
+                    -- TODO: Log here, e.g. MsgCrashMonitoring
+                    case recovery of
```
rvl, line 326: I would remove FollowFailure and let exceptions bubble up to the caller of `follow`.
The caller can decide whether to retry, possibly using the `recovering` function (which has good logging abilities).

lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -378,19 +459,90 @@ instance ToText FollowLog where
...
+        MsgApplyBlocks _ _ -> Debug
+        MsgFollowLog msg -> getSeverityAnnotation msg
         MsgWillRollback _ -> Debug
+        MsgDidRollback _ _ -> Debug
+        MsgFailedRollingBack _ -> Error
         MsgWillIgnoreRollback _ _ -> Debug
+
+--
+-- Log aggregation
+-- Perhaps this could be done using iohk-monitoring as well.
```
rvl: It can all be done with contra-tracer and tracer-transformers.

lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -378,19 +459,90 @@ instance ToText FollowLog where
...
+-- - How many rollbacks we encounter
+-- - What the worker itself is finding on-chain
+-- - How long the worker operations are
+-- - When we are restored
+--
+-- If we are restored:
+-- - If the node goes out of sync with the network
+-- - If the wallet goes out of sync with the node
+monitorStatus :: Tracer IO (FollowLog tr) -> IO (Tracer IO (FollowLog tr))
+monitorStatus tr = do
```
rvl: Metrics averaged over time will be good. We can export them with ekg.
For this ticket though, we just want to log interesting events and state changes.

It _may_ work better if log interpretation and transformation is done at the app top-level, all together, rather than inside each component. I'm not sure yet though.

rvl (commented):
Good, on the right track. I also have some related logging fixes in #2570, so I'll pull them out into a separate PR.

lib/core/src/Cardano/Wallet/Network.hs
Anviking: > It may work better if log interpretation and transformation is done at the app top-level, all together, rather than inside each component. I'm not sure yet though.

Yeah, I was wondering about that too. Partly why I wondered if iohk-monitoring was the "right way" (despite its weirdness), as I think their backends and configuration setup served exactly as that (top-level, all together transformation).

I suspect we can get by for now doing the transformation here.

Anviking: > For this ticket though, we just want to log interesting events and state changes.

IMO, being able to tell 
1. If the wallet is restoring slowly
2. If the wallet is rolling back in a loop (I _think_ the stats would be able to tell this)
is an important thing we lose when we remove the "on-each-batch-logs"

lib/core/src/Cardano/Wallet.hs
```diff
@@ -750,73 +752,60 @@ listUtxoStatistics ctx wid = do
...
-        ( HasLogger WalletLog ctx
-        , HasDBLayer s k ctx
+        ( HasDBLayer s k ctx
         , HasNetworkLayer ctx
-        , HasGenesisData ctx
         , IsOurs s Address
         , IsOurs s RewardAccount
         )
     => ctx
+    -> Tracer IO WalletFollowLog
```
rvl: There's a tracer in `ctx` already, so no param needed.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -875,23 +864,21 @@ restoreBlocks ctx wid blocks nodeTip = db & \DBLayer{..} -> mapExceptT atomicall
...
     db = ctx ^. dbLayer @s @k
-    tr = ctx ^. logger @WalletLog
 
     logCheckpoint :: Wallet s -> IO ()
     logCheckpoint cp = traceWith tr $ MsgCheckpoint (currentTip cp)
 
     logDelegation :: (SlotNo, DelegationCertificate) -> IO ()
-    logDelegation (slotNo, cert) = traceWith tr $ MsgDelegation slotNo cert
+    logDelegation (slotNo, cert) = traceWith tr
+        $ MsgDiscoveredDelegationCert slotNo cert
```
rvl:
```suggestion
    logDelegation = traceWith tr . uncurry MsgDiscoveredDelegationCert
```

rvl: Delegations could be logged all together too.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -626,7 +626,8 @@ migrateManually tr proxy defaultFieldValues =
             ColumnPresent  -> do
                 value <- either (fail . show) (\x -> pure $ "\"" <> x <> "\"") $
                     fromPersistValueText (toPersistValue W.EncryptWithPBKDF2)
-                traceWith tr $ MsgManualMigrationNeeded passphraseScheme value
+                traceWith tr . MsgExpectedMigration
```
rvl: Slightly better now

rvl (commented):
Nice, I'll review more later.

Anviking:
_Not completely polished yet, but should be somewhat close... Or I haven't looked at it from a step back yet_ 

The logs are certainly less noisy:

Skärmavbild 2021-03-30 kl  21 07 49
Skärmavbild 2021-03-30 kl  20 26 10

_(testnet wallet, integration tests with info logs, both using this PR)_

But I hope they also provide enough information. In this regard this PR doesn't do anything magical. Aside from making some information much more apparent (changes to sync progress). We might be able to buffer some debug logs and provide on errors, but that's a whole separate beast.

lib/core/src/Cardano/Wallet.hs
Anviking: The idea was to have `follow` provide this tracer through the callback, such that it could intercept and aggregate on this "inner" trace too.

I didn't end up doing that, but I think the concept is sound and something that could be done in the future.

lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -381,19 +490,182 @@ instance ToText FollowLog where
...
+
+-- | Starts a new thread for monitoring health and statistics from
+-- the returned @FollowLog tr@.
+monitorFollowStats :: Tracer IO (FollowLog tr) -> IO (Tracer IO (FollowLog tr))
+monitorFollowStats tr = do
+    t0' <- getCurrentTime
+    let sp = NotResponding -- Hijacked as a default value. Should be fine.
+                           -- It appears to be otherwise unused in the codebase.
+    mvar <- newMVar $ emptyStats t0' sp
+    _ <- forkIO $ loop mvar (5 * second)
```
Anviking: `forkIO` — should presumable use something like `race` to get the monitoring thread to stop if `follow` stops.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
Anviking: Yeah, not sure what the proper way would be 🤷‍♂️

lib/core/src/Cardano/Wallet/Network.hs
Anviking: As said elsewhere, my goal was to move as much responsibility from the callers to `follow`.

Maybe `recovering` (inside follow) is a good idea, but we didn't use it before at least, so we could keep it separate from this PR.

Anviking:
bors try

lib/core/src/Cardano/Wallet/Network.hs
Anviking: Btw, the pool worker simply stops if you kill and restart the node... We should probably either recover from exceptions for both pool and wallet workers, or re-add an `ExceptT` (#2581)

lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -381,19 +490,190 @@ instance ToText FollowLog where
...
+    MsgApplyBlocks _tip blocks ->
+        s { blocksApplied = overCurrent (+ NE.length blocks) (blocksApplied s) }
+    MsgDidRollback _ _ ->
+        s { rollbacks = overCurrent (1 +) (rollbacks s) }
+    MsgSyncProgress p ->
+        s { prog = overCurrent (const p) (prog s) }
+    _ -> s
+
+instance ToText (FollowStats LogState) where
+    toText (FollowStats b r t sp) = syncStatus <> " " <> stats
```
rvl: This time-based summary is good for the stake pools thread, but for wallets I think it should be logging when there are incoming transactions, or when restoration is in progress, and complete silence otherwise.

I'll move this statistics code into the Cardano.Wallet.Logging module too.

rvl (commented):
Looks good. I'll apply a couple clean ups then merge.

lib/core/src/Cardano/Wallet/Network.hs
Anviking: > but for wallets I think it should be logging when there are incoming transactions

Just to be clear: the logs collected for aggregation are still shown or not depending on severity. It is still logging discoveries.

> or when restoration is in progress

Yes, ideally the `SyncProgress` `LogState` transitions should probably be logged immediately when they occur, not only from a periodic monitor-check. My hunch is still that regular status updates wallet workers are worth it too, though.  A summary every 5 minutes should be rare compared to other Info-logs (like request logs from polling), and alert us if we see slow or no restoration progress.

Anviking: Or I guess we could  _monitor_ / poll the progress regularly without _logging_ it (+ stats), if we are still in-sync.

If the chain-following were to halt, we'd then notice, without being _required_ to log every-five-minutes status updates.

Anviking:
bors try

Anviking:
bors try

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -3019,21 +3018,21 @@ instance IsServerError (Request, ServerError) where
 -- | The type of log messages coming from the server 'ApiLayer', which may or
 -- may not be associated with a particular worker thread.
 data WalletEngineLog
-    = MsgWallet (WorkerLog WalletId WalletLog)
+    = MsgWalletWorker (WorkerLog WalletId W.WalletWorkerLog)
```
Anviking: Not sure if this diff is from a rebase or not 🤔

lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -377,21 +490,216 @@ instance ToText tr => ToText (FollowLog tr) where
...
+-- warn there.
+explainedSeverityAnnotation :: FollowStats LogState -> (Severity, Maybe Text)
+explainedSeverityAnnotation s
+        | progressMovedBackwards = (Warning, Just "progress decreased")
+        | noBlocks && notRestored = (Warning, Just "not applying blocks")
+        | otherwise = (Info, Nothing)
+  where
+    progressMovedBackwards = current (prog s) < prev (prog s)
+    notRestored = current (prog s) /= Ready
+    noBlocks = (current (blocksApplied s) - prev (blocksApplied s)) <= 0
```
Anviking, line 660: I did this as a workaround for the problem described in the `prog :: f SyncProgress` doc  above.

But having the sync progress reported wrong (not re-calculated as the current time changes), could probably be really confusing. I should have a look at this again.

Anviking: This is now fixed by making it easy to calculate syncProgress without node-connection. Now it returns `NotResponding` rather than blocking.

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: No, I think it was intentional.

lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -377,21 +480,223 @@ instance ToText tr => ToText (FollowLog tr) where
...
+    -> (SlotNo -> IO SyncProgress)
+    -> ((Tracer IO (FollowLog tr)) -> IO ())
+    -> IO ()
+withFollowStatsMonitoring tr calcSyncProgress act = do
+    t0' <- getCurrentTime
+    var <- newTMVarIO $ emptyStats t0'
+    let tr' = flip contramapM tr $ \msg -> do
+            atomically $ do
+                s <- takeTMVar var
+                putTMVar var (updateStats msg s)
```
rvl: This is not going to leave unevaluated thunks, is it? I suppose all the results are consumed by the other thread.

rvl (approved):
Good improvement :+1:

rvl:
Oops I reviewed in the wrong order, and buildkite picked up the non-master merge.

lib/core/src/Cardano/Wallet/Network.hs
Anviking: Good that you brought it up again. I did add a `force` in `updateStats` long time ago, but:
1. I'm not sure if that was effective or not because the outer lazy `putTMVar`
2. After asking in #haskell it seems it seems `force` is not the right way (but ok, also sounds familiar)

Anviking: I replaced `TMVar` with `StrictTMVar` for now (could have used `$!` on this line also I think).

Will have a look at how to replace the `force` with something better in a follow-up PR.

PR #2586, opened 2021-03-29
Add logging of tx submission

rvl:
### Issue Number

ADP-630 / ADP-764

### Overview

These are some logging/worker improvements I made in #2570. I'll put them in a separate PR because they overlap with #2583.

- Adds logging of transaction submission results in the wallet layer.
- Fixes the logging types in the tracer bundle.
- Removes duplication between `initWorker` and `registerWorker`.

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -162,6 +162,7 @@ module Cardano.Wallet
 
     -- * Logging
     , WalletLog (..)
+    , TxSubmitLog (..)
     ) where
 
 import Prelude hiding
@@ -187,6 +188,8 @@ import Cardano.Wallet.DB
     , defaultSparseCheckpointsConfig
     , sparseCheckpoints
     )
+import Cardano.Wallet.Logging
+    ( traceWithExceptT )
 import Cardano.Wallet.Network
     ( ErrGetAccountBalance (..)
     , ErrPostTx (..)
@@ -342,7 +345,9 @@ import Control.DeepSeq
     ( NFData )
 import Control.Monad
     ( forM, forM_, replicateM, unless, when )
-import Control.Monad.IO.Class
+import Control.Monad.Class.MonadTime
+    ( getCurrentTime )
+import Control.Monad.IO.Unlift
     ( liftIO )
 import Control.Monad.Trans.Class
     ( lift )
@@ -398,7 +403,7 @@ import Data.Set
 import Data.Text.Class
     ( ToText (..) )
 import Data.Time.Clock
-    ( NominalDiffTime, UTCTime, getCurrentTime )
+    ( NominalDiffTime, UTCTime )
 import Data.Type.Equality
     ( (:~:) (..), testEquality )
 import Data.Word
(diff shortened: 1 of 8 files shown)


lib/core/src/Cardano/Wallet/Registry.hs
```diff
@@ -250,28 +246,50 @@ register registry ctx k (MkWorker before main after acquire) = do
...
             | otherwise -> T.take 8 (toText key) <> ": " <> toText msg
 
 instance HasPrivacyAnnotation (WorkerLog key msg)
 instance HasSeverityAnnotation msg => HasSeverityAnnotation (WorkerLog key msg) where
     getSeverityAnnotation = \case
-        MsgFinished -> Notice
+        MsgThreadAfter msg -> getSeverityAnnotation msg
+        MsgFromWorker _ msg -> getSeverityAnnotation msg
+
+data AfterThreadLog
```
Anviking, line 269: 👍 (short doc comment could be nice though)

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2370,26 +2321,74 @@ registerWorker
...
+  where
+    before ctx' _ = void $ unsafeRunExceptT $ createWallet ctx'
+    re = ctx ^. workerRegistry @s @k
+
+-- | Something to pass as the coworker action to 'newApiLayer', which does
+-- nothing, and never exits.
+idleWorker :: ctx -> wid -> IO ()
+idleWorker _ _ = forever $ threadDelay maxBound
+
+startWorker
```
Anviking: Some short doc-comments connecting startWorker, registerWorker and initWorker would be useful

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -3009,3 +3008,29 @@ instance LiftHandler (Request, ServerError) where
...
                 ) : headers
             }
+
+{-------------------------------------------------------------------------------
+                               Logging
+-------------------------------------------------------------------------------}
+
+-- | The type of log messages coming from the server 'ApiLayer', which may or
+-- may not be associated with a particular worker thread.
+data WalletEngineLog
```
Anviking, line 3017: 👌
But you have `MsgTxSubmit` both in `WalletEngineLog` _and_ in `WalletLog`? 🤔

Is this intentional? For txs when first submitted from a wallet, and for when they are retried?

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1557,38 +1562,45 @@ submitTx
...
+        , HasLogger WalletLog ctx
         )
     => ctx
     -> WalletId
     -> (Tx, TxMeta, SealedTx)
     -> ExceptT ErrSubmitTx IO ()
 submitTx ctx wid (tx, meta, binary) = db & \DBLayer{..} -> do
-    withExceptT ErrSubmitTxNetwork  $
+    withExceptT ErrSubmitTxNetwork $ traceWithExceptT tr $
         postTx nw binary
```
Anviking, line 1573: I think tracing here makes sense, but `postTx` also has a debug-level `MsgPostTx`, which is arguably a bit weird

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -896,10 +883,7 @@ instance HasSeverityAnnotation StakePoolLog where
         MsgFetchPoolMetadata e -> getSeverityAnnotation e
         MsgFetchTakeBreak{} -> Debug
         MsgGCTakeBreak{} -> Debug
-        MsgGCFinished{} -> Debug
-        MsgGCThreadKilled{} -> Debug
-        MsgGCUserInterrupt{} -> Debug
-        MsgGCUnhandledException{} -> Debug
```
Anviking, line 902: nice

Anviking (approved):
👌 think it makes sense

lib/core/src/Cardano/Wallet/Api/Server.hs
rvl: Yes - one is for the raw transaction submit endpoint - which is not associated with any wallet.
I'll rename the constructor though, to make this less confusing.

lib/core/src/Cardano/Wallet.hs
rvl: Yes - that logging is good for debug, but it's too low-level, and doesn't log the result.
The transaction submission event is something that the user should be told about with nice log messages.

lib/core/src/Cardano/Wallet/Registry.hs
rvl: OK done.

lib/core/src/Cardano/Wallet/Api/Server.hs
rvl: Added a doc comment, and renamed these functions to be more descriptive.

rvl:
Thanks for the review comments, they should be addressed. Let's try our luck...

bors r+

PR #2587, opened 2021-03-29
Mount middleware for automatically sanitizing JSON Big Ints upon explicit client request

KtorZ:
# Issue Number

ADP-837

# Overview

- d5f2044821d10baa9282b40999e60b3fc8f0093c
  :round_pushpin: **allow 'mintSeaHorseAssets' to take an extra quantity parameter**
  
- 687f6335d92c0f2821540dc1a2c0364d8aef4840
  :round_pushpin: **define integration test scenario illustrating the use of 'X-Max-Safe-Integer'**
  
- cde44c146528a8fc0901fc5a80299b29348496dc
  :round_pushpin: **write Wai middleware for serializing BigInt as String on demand**
    This is done via a middleware for two reasons:
  - a) It is low overhead when disabled. So that the default behavior of the server remains unchanged.
  - b) Doing so makes the whole approach quite 'holistic' and we risk not to forget turning a potentially
       big int into a string later when writing JSON instances.

- 92d852cd23d6b0ff9eb2bb93808c304d22e394b6
  :round_pushpin: **sketch out simple property for 'traverseJSON'**
    As a sanity check to convince myself that the function does indeed preserve the JSON structure.

- e92568e25db462a611d35b3e4eedb99041fd8d16
  :round_pushpin: **modify 'AddressAmount' JSON parser to also accept amounts as string by default.**
    No need for a specific header for that one since this is simply an additive fallback behavior. Thus, it is only triggered
  for clients which seek for it.

# Comments

TODO: 

- [ ] Document this in the API documentation.

### DIFF
--- a/lib/core-integration/cardano-wallet-core-integration.cabal
+++ b/lib/core-integration/cardano-wallet-core-integration.cabal
@@ -91,6 +91,7 @@ library
       Test.Integration.Scenario.API.Byron.Transactions
       Test.Integration.Scenario.API.Byron.Migrations
       Test.Integration.Scenario.API.Byron.Network
+      Test.Integration.Scenario.API.Miscellaneous
       Test.Integration.Scenario.API.Shelley.Addresses
       Test.Integration.Scenario.API.Shelley.CoinSelections
       Test.Integration.Scenario.API.Shelley.HWWallets
--- a/lib/core-integration/src/Test/Integration/Faucet.hs
+++ b/lib/core-integration/src/Test/Integration/Faucet.hs
@@ -2252,21 +2252,22 @@ maryIntegrationTestAssets tips = maMnemonics >>= take 3
         ]
     combined p = simple p `TokenBundle.add` fruit p
 
--- | Create @n@ unique SeaHorse tokens for each provided @Address@.
+-- | Create @q@ tokens of @n@ SeaHorse assets for each provided @Address@.
 --
 -- The result can be used for minting using the cli-based faucet.
 seaHorseTestAssets
-    :: Int -- ^ Number of sea horses per address
+    :: TokenQuantity -- ^ Quantity of tokens per asset, per address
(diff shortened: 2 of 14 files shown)


KtorZ:
@piotr-iohk putting this one one your radar, if you want to have a quick look at it.

piotr-iohk:
Thanks @KtorZ. I've updated ruby client (https://github.com/piotr-iohk/cardano-wallet-rb/pull/33). I might add some more tests to e2e suite, although don't want to just replay the same/similar tests there.

So if I'm not mistaken setting `X-Max-Safe-Integer` for any request should result in any integer returned by the API to be reported as string, correct?

I believe it needs to be documented in APIDoc, not sure where would be good place though, since it is global for all requests.

piotr-iohk:
@KtorZ I've pushed commit 85b3849 documenting the header in swagger.yaml. Added it to all endpoints except wallet/byron wallet detlete and forget transaction endpoints, since they don't return anything anyway...

Screenshot from 2021-03-30 15-25-27

KtorZ:
Nikola said he wanted to put that one on hold until he got time to play around the `json-bigint` node library which could likely solve their problem without requiring changes from this PR.

piotr-iohk:
@nikolaglumac it seems https://github.com/input-output-hk/daedalus/pull/2472 was fixed without the need of this PR?

nikolaglumac:
> @nikolaglumac it seems [input-output-hk/daedalus#2472](https://github.com/input-output-hk/daedalus/pull/2472) was fixed without the need of this PR?

@piotr-iohk please stand by. We need to confirm few things related to HW first.

rvl:
Now that Daedalus can parse numbers > 2<sup>53</sup> from JSON (as can any javascript app using the [json-bigint](https://www.npmjs.com/package/json-bigint) module), we don't need this feature.

PR #2589, opened 2021-03-31
Hash tx metadata before signing

rvl:
### Issue Number

ADP-800

### Overview

When signing tx metadata using the `signMetadata` endpoint, hash the message first before signing.

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
@@ -1115,9 +1115,9 @@ spec = describe "SHELLEY_WALLETS" $ do
         let dummyChainCode = BS.replicate 32 0
         let sig = CC.xsignature $ BL.toStrict $ getFromResponse id rSig
         let key = unsafeXPub $ fst (getFromResponse #getApiVerificationKey rKey) <> dummyChainCode
-        let msg = unsafeFromHex "A10071706C65617365207369676E20746869732E"
+        let msgHash = unsafeFromHex "1228cd0fea46f9a091172829f0c492c0516dceff67de08f585a4e048a28a6c9f"
 
-        liftIO $ CC.verify key msg <$> sig `shouldBe` Right True
+        liftIO $ CC.verify key msgHash <$> sig `shouldBe` Right True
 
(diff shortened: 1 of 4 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
```diff
@@ -1115,9 +1115,10 @@ spec = describe "SHELLEY_WALLETS" $ do
         let dummyChainCode = BS.replicate 32 0
         let sig = CC.xsignature $ BL.toStrict $ getFromResponse id rSig
         let key = unsafeXPub $ fst (getFromResponse #getApiVerificationKey rKey) <> dummyChainCode
-        let msg = unsafeFromHex "A10071706C65617365207369676E20746869732E"
+        let _msg = unsafeFromHex "A10071706C65617365207369676E20746869732E"
```
paweljakubas: this one probably we want to delete, right?

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2001,8 +2003,11 @@ signMetadataWith ctx wid pwd (role_, ix) metadata = db & \DBLayer{..} -> do
             let DerivationPrefix (_, _, acctIx) = derivationPrefix (getState cp)
             let acctK = deriveAccountPrivateKey encPwd rootK acctIx
             let addrK = deriveAddressPrivateKey encPwd acctK role_ addrIx
-            let msg   = serialiseToCBOR metadata
-            pure $ Signature $ BA.convert $ CC.sign encPwd (getRawKey addrK) msg
+            pure $
+                Signature $ BA.convert $
+                CC.sign encPwd (getRawKey addrK) $
+                hash @ByteString @Blake2b_256 $
```
paweljakubas, line 2009: :+1:

paweljakubas (approved):
lgtm! wonder if we can add some golden testing basing on some already existent test in cardano-node?

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Wallets.hs
rvl: Yeah, the string is already shown in the comment above. I deleted it.

rvl:
I added a cheap "golden" signature check to the same integration test.

PR #2590, opened 2021-03-31
API specification for multi-asset wallet migration

jonathanknowles:
# Issue Number

ADP-840

# Overview

This PR revises the **_API specification_** to support multi-asset wallet migrations.

The **_server implementation_** will be added in a subsequent PR. (Connecting the API with the algorithm.)

# Details

This PR:

- [x] Adds a `ApiWalletMigrationBalance` type:

    ```yaml
    ApiWalletMigrationBalance: &ApiWalletMigrationBalance
      type: object
      required:
        - ada
        - assets
      properties:
        ada: *amount
        assets: *walletAssets
    ```

- [x] Revises the `ApiWalletMigrationInfo` type:

    ```yaml
    ApiWalletMigrationInfo: &ApiWalletMigrationInfo
      type: object
      required:
        - selections
        - total_fee
        - balance_leftover
        - balance_selected
      properties:
        selections:
          type: array
          items: *ApiCoinSelection
        total_fee:
          <<: *amount
        balance_leftover:
          <<: *ApiWalletMigrationBalance
        balance_selected:
          <<: *ApiWalletMigrationBalance

    ```


lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -993,6 +995,18 @@ data ApiPostRandomAddressData = ApiPostRandomAddressData
...
+    } deriving (Eq, Generic, Show)
+
+data ApiWalletMigrationBalance = ApiWalletMigrationBalance
+    { ada :: !(Quantity "lovelace" Natural)
+    , assets :: !(ApiT W.TokenMap)
+    } deriving (Eq, Generic, Show)
+
 data ApiWalletMigrationPostData (n :: NetworkDiscriminant) (s :: Symbol) =
     ApiWalletMigrationPostData
     { passphrase :: !(ApiT (Passphrase s))
```
piotr-iohk: passphrase wouldn't be needed here, but in the next step (i.e. confirming migration) if I'm not mistaken?
Screenshot from 2021-03-31 13-21-24

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -993,6 +995,18 @@ data ApiPostRandomAddressData = ApiPostRandomAddressData
     } deriving (Eq, Generic, Show)
       deriving anyclass NFData
 
+data ApiWalletMigration (n :: NetworkDiscriminant) = ApiWalletMigration
+    { selections :: !(NonEmpty (ApiCoinSelection n))
+    , cost :: !(Quantity "lovelace" Natural)
```
Anviking: No strong opinion, but if this was named `fee` it would be consistent with `ApiTransaction`.

specifications/api/swagger.yaml
```diff
@@ -1628,6 +1628,30 @@ components:
...
+        - cost
+        - balance_included
+        - balance_leftover
+      properties:
+        selections:
+          type: array
+          items: *ApiCoinSelection
+        cost: *amount
+        balance_included: *ApiWalletMigrationBalance
+        balance_leftover: *ApiWalletMigrationBalance
```
Anviking: I guess we should explain why there might be leftovers, now or later, such that users know if they need to care.

Assets could be leftover in some cases, I recall. Could dust ada end up here too?

Edit: Looking at the previous description:

> Leftovers dust coins which won't be migrated nor spent as fees.

I'd guess so too? 

And I see you're on it:

> TODO: Updates the API endpoint documentation.

Anviking (commented):
Looks reasonable, I think

lib/core/src/Cardano/Wallet/Api/Types.hs
Anviking: Also, is POST the correct verb now when the selection is only returned and not submitted?

I guess that depends on whether we mark the tx as pending immediately or not, which was a point of uncertainty in https://jira.iohk.io/browse/ADP-342

specifications/api/swagger.yaml
jonathanknowles: > I guess we should explain why there might be leftovers, now or later, such that users know if they need to care.

Definitely!

> Assets could be leftover in some cases, I recall.

That's right. The requirements allow us to omit assets from a migration if the algorithm cannot find a viable strategy to include them. An example of when this might happen is if the remaining UTxO contains a single entry with the minimum ada amount and a single non-ada asset. In this situation, it's impossible to create a transaction, because there's not enough ada to pay for the fee (the only source of ada is the minimum ada amount, which can't be reduced). The only other option is to burn the assets, which the requirements do not allow.

> Could dust ada end up here too?

Ada quantities that are less than or equal to `input_marginal_fee` will be omitted by the algorithm automatically, and not even considered for inclusion.

It is also possible for ada quantities greater than `input_marginal_fee` to be omitted: if the remaining UTxO set contains only **_n_** small values whose total value is less than (`minimum_ada_quantity` + `base_transaction_fee` + **_n_** × `input_marginal_fee`), then it won't be possible to create a transaction from them.

lib/core/src/Cardano/Wallet/Api/Types.hs
jonathanknowles: > No strong opinion, but if this was named fee it would be consistent with ApiTransaction.

Good point! Perhaps `totalFee` would be even better, because each individual selection has its own fee, and this is the sum of all fees. What do you think?

lib/core/src/Cardano/Wallet/Api/Types.hs
jonathanknowles: > I guess that depends on whether we mark the tx as pending immediately or not.

The intention is for the user to:

1. Call the `migrateWallet` endpoint, which supplies a list of selections (= unsigned transactions).
2. Inspect the results, check that the total fee and the list of selections are acceptable.
3. Call another endpoint (yet-to-be-added), which accepts a list of selections. In the case of HW wallets, the user would sign them first.

The yet-to-be-added endpoint would check that all the supplied selections are valid, by checking that all the UTxO entries still exist, before submitting them as transactions to the network.

So I think there's no need for the `migrateWallet` endpoint to mark transactions as pending here.

If the user calls the `migrateWallet` endpoint and then makes another transaction that spends some of the UTxO entries in the meantime, they can easily get themselves unstuck by calling `migrateWallet` again to generate a new set of selections that don't include those UTxO entries.

jonathanknowles: > Also, is POST the correct verb now when the selection is only returned and not submitted?

I suspect POST is still the correct verb, as the user still has to post a list of target addresses so that we can generate real coin selections with `(Address, TokenBundle)` pairs as the outputs.

jonathanknowles: > passphrase wouldn't be needed here, but in the next step (i.e. confirming migration) if I'm not mistaken?

Ah yes, well-spotted!

You're right: only the final step (calling the yet-to-be-added submission endpoint) should require a passphrase.

jonathanknowles: > passphrase wouldn't be needed here, but in the next step (i.e. confirming migration) if I'm not mistaken?

Fixed in ed63066f2360baaaef90fb15337e636207885e74.

lib/core/src/Cardano/Wallet/Api/Types.hs
jonathanknowles: Changed this to `totalFee` in commit 03696719acfd86b225d6fca8f11766f911a0b1c0.

CuneytMutlu:
3431ed03a0278a81041660fa1684f579

lib/core/src/Cardano/Wallet/Api/Types.hs
jonathanknowles: @piotr-iohk @Anviking 

This has changed: we now keep the existing API endpoints (without changing them), but just modify the `ApiWalletMigrationInfo` type. (See updated PR description.)

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -1279,7 +1295,13 @@ instance Arbitrary ApiByronWalletBalance where
...
 
-instance Arbitrary ApiWalletMigrationInfo where
+instance Arbitrary ApiWalletMigrationBalance where
+    arbitrary = ApiWalletMigrationBalance
+        <$> reasonablySized arbitrary
+        <*> reasonablySized arbitrary
+    shrink = genericShrink
+
+instance Arbitrary (ApiWalletMigrationInfo n) where
     arbitrary = genericArbitrary
```
KtorZ: Screenshot from 2021-04-29 07-25-30

Be careful with genericArbitrary! This generates data-set that are way too large!

jonathanknowles: > Be careful with genericArbitrary! This generates data-set that are way too large!

Thanks for spotting this! I'll revise the `Arbitrary` instance to generate a smaller data set. 👍🏻

specifications/api/swagger.yaml
```diff
@@ -2097,18 +2097,50 @@ components:
...
+        - selections
+        - total_fee
+        - balance_leftover
+        - balance_selected
       properties:
-        migration_cost:
-          <<: *amount
-          description: Total amount which will be paid as fees for the migration.
-        leftovers:
+        selections:
```
rvl, line 2117: This field could have a description too.

rvl (commented):
There should also be updates to the `getShelleyWalletMigrationInfo` and `migrateShelleyWallet` specs, right?

specifications/api/swagger.yaml
jonathanknowles: > This field could have a description too.

That's a good point!

I've added a description in this updated commit: 202e0240bfbfcdb88bbdb3ef780dce36a13cb890.

jonathanknowles:
Hi @rvl

> There should also be updates to the `getShelleyWalletMigrationInfo`

I plan to update the **description** of this endpoint in the next PR, when the algorithm is connected up to the API. (We'll have to update the description then anyway, in order to remove the  "⚠️ IMPORTANT: this endpoint has been temporarily disabled!" warning.)

As for their **return types**, these currently both have the same return types:
- `getShelleyWalletMigrationInfo`
- `getByronWalletMigrationInfo`

These both return a `responsesGetWalletMigrationInfo`, which returns a `ApiWalletMigrationInfo`. I think it's okay for both to return the same type. In the Byron case, the `assets` part of each balance field will be empty.

> and `migrateShelleyWallet` specs, right?

I wasn't planning to change the type of this endpoint in any way, as it currently returns a list of transactions, which seems to be exactly what we need.

But perhaps you've thought of something that I've missed?

rvl:
In `migrateShelleyWallet` the description mentions "shape", but this no longer applies. It could also be revised to say that it transfers as much funds as possible.

The description of `getShelleyWalletMigrationInfo` could also be revised. Its summary is "Calculate Cost", which is not quite right any more.

jonathanknowles:
> In `migrateShelleyWallet` the description mentions "shape", but this no longer applies. It could also be revised to say that it transfers as much funds as possible.
> 
> The description of `getShelleyWalletMigrationInfo` could also be revised. Its summary is "Calculate Cost", which is not quite right any more.

Hi @rvl 

I've updated the descriptions in this pair of commits:
* 7cdff84587f5cf3286f7258db1343c07ecccaba0: `migrate{Byron,Shelley}Wallet`
* 76c08675f29c965f02658bacf37675a918ac5c7e: `get{Byron,Shelley}WalletMigrationInfo`

rvl (approved):
Nice thanks - better now.

rvl:
Right, that bors failure was correct. This was the error message:
```
$ ././.buildkite/check-bors.sh
master commit is 182e86427dad393fc4f97037d4a9faebc5d95743
bors/staging parent commits are e687560265a0a49d0f3d5056969ec5b0c0288afa 3e14fb4abc2da149242656f51006d12a51f4b479
 
Refusing to merge because the pull request does not target master.
You should only use Bors to merge into the master branch.
Either change the PR base branch to master, or merge manually.
```

jonathanknowles:
> Right, that bors failure was correct. This was the error message:

Perhaps it's because of a race condition?

The branch of this PR was originally based on `jonathanknowles/multi-asset-migration-transaction-layer`, which just merged to `master` (https://github.com/input-output-hk/cardano-wallet/pull/2620).

If branch `A` is based on branch `B` which is based on `master`, but then branch `B` is merged to `master`, then I've observed that GitHub will automatically update the base branch of `A` to be `master`.

But if bors attempts to build before this happens, then it will fetch something that's not yet based on `master`, which seems to be what happened here.

When I did `bors r+` for the second time, I didn't have to do any manual rebasing onto `master` -- it seems that GitHub did this automatically.

rvl:
Yes - a race, sort of. Bors would have created the merge commit for the next batch before it got notification from GitHub that the PR base branch changed.

PR #2592, opened 2021-03-31
Add derivation paths in addresses listing

paweljakubas:
# Issue Number

adp-766

# Overview

- [x] Introducing `ApiAddressInfo` in swagger
- [x] Adding `ApiAddressInfo` in Api of core
- [x] changing `knownAddresses` to return also two derivation indices
- [x] aligning `listAddresses` of Api.Server and Wallet
- [x] aligning core unit tests
- [x] aligning integration testing 

Example `ApiAddressInfo` for payment addr (role=0) and index=17537 (for byron accIx and addrIx are returned, ie. its derivation path as is there understood):
```
    {
      "address": {
        "state": "used",
        "id": "<addr>"
      },
      "derivation_path": [
        "0",
        "17537"
      ]
    }
```

# Comments


lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -681,7 +682,7 @@ postTx
     -> m (HTTP.Status, Either RequestException (ApiTransaction n))
 postTx ctx (wSrc, postTxEndp, pass) wDest amt = do
     addrs <- listAddresses @n ctx wDest
-    let destination = (addrs !! 1) ^. #id
+    let destination = (addrs !! 1) ^. (#address . #id)
```
KtorZ, line 685: This nesting seems _wrong_ :thinking: 
(I'll comment on the swagger schema modification directly).

specifications/api/swagger.yaml
```diff
@@ -1427,6 +1427,15 @@ components:
...
         state: *addressState
 
+    ApiAddressInfo: &ApiAddressInfo
+      type: object
+      required:
+        - address
+        - derivation_path
+      properties:
+        address: *ApiAddress
+        derivation_path: *derivationPath
```
KtorZ, line 1437: This breaks the current API in an unfortunate and unnecessary way. We can introduce the extra `derivation_path` in the top-level object, without indirection. Is there any reason why this indirection was introduced? 

It seems that it would suffice to extend `ApiAddress` with a new, required, `derivation_path` field.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery.hs
```diff
@@ -127,15 +127,15 @@ class CompareDiscovery s where
...
+-- | Extract the list of all known addresses. Also specify Role/Index.
 --
 -- NOTE: Change addresses aren't considered "known" until they've been used. The
 -- rationale is that, we don't want users or consumers of the wallet to be using
 -- change addresses prematurely.
 class KnownAddresses s where
     knownAddresses
         :: s
-        -> [(Address, AddressState)]
+        -> [(Address, AddressState, DerivationIndex, DerivationIndex)]
```
KtorZ, line 138: Shouldn't the result be a `NonEmpty DerivationIndex` like it's done for `IsOurs`? Derivation paths actually have different length depending on the type of wallets. Byron wallets have 2 indexes, Icarus and Shelley have 5.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Random.hs
```diff
@@ -334,9 +334,17 @@ instance CompareDiscovery (RndState n) where
...
+        , retrieve ((,Unused) <$> pendingAddresses s)
         ]
+     where
+       construct (derPath, (addr, state)) =
+           ( addr
+           , state
+           , DerivationIndex $ getIndex $ fst derPath
+           , DerivationIndex $ getIndex $ snd derPath
+           )
+       retrieve = map construct . Map.toList
```
KtorZ, line 347: A fold of the `Map` would be better here to avoid traversing the list twice.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
```diff
@@ -371,19 +371,36 @@ role = fromMaybe (error $ "role: unmatched type" <> show (typeRep @c))
...
+            Just Refl  -> Just (DerivationIndex 0)
+            Nothing -> Nothing
+    tryUtxoInternal =
+        case testEquality (typeRep @c) (typeRep @'UtxoInternal) of
+            Just Refl  -> Just (DerivationIndex 1)
+            Nothing -> Nothing
+    tryMutableAccount =
+        case testEquality (typeRep @c) (typeRep @'MutableAccount) of
+            Just Refl  -> Just (DerivationIndex 2)
+            Nothing -> Nothing
```
KtorZ, line 389: Seems like this is duplicating logic with the `Enum` instance for `Role` already. I'd suggest instead to define a functon to reify a type-level role into a `Role` value, and then piggy-back on the Enum instance to get the role index to avoid duplicating bits of logic.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
```diff
@@ -371,19 +371,36 @@ role = fromMaybe (error $ "role: unmatched type" <> show (typeRep @c))
...
 addresses
-    :: forall c k. ()
+    :: forall c k. Typeable c
     => (KeyFingerprint "payment" k -> Address)
     -> AddressPool c k
-    -> [(Address, AddressState)]
+    -> [(Address, AddressState, DerivationIndex, DerivationIndex)]
 addresses mkAddress =
-    map (\(k, (_, st)) -> (mkAddress k, st))
+    map (\(k, (ix, st)) -> (mkAddress k, st, roleIndex @c, DerivationIndex $ getIndex ix))
```
KtorZ, line 403: This is missing the derivation path prefix.

KtorZ (changes requested):

paweljakubas:
closing in favor of https://github.com/input-output-hk/cardano-wallet/pull/2598

PR #2597, opened 2021-04-02
API: Add content-type to all structured error responses

### LINKED ISSUE #2596: Missing content-type application/json;charset=utf-8 for status code 404 on /wallets
# Context

I'm building a library for Elixir that talks to the cardano-wallet and I think I found a bug.

The content-type header seems to be missing when calling http://localhost:8090/v2/wallets/<wallet_id> with an non existing id which returns with status code 404. 

This is the headers I get back:
    {"date", "Thu, 01 Apr 2021 21:12:48 GMT"},
    {"server", "Warp/3.3.5"},
    {"content-length", "122"}

If I call the same endpoint but with a badly formatted id which returns a status code 400 I get these headers:
    {"date", "Thu, 01 Apr 2021 21:12:48 GMT"},
    {"server", "Warp/3.3.5"},
    {"content-length", "92"},
    {"content-type", "application/json;charset=utf-8"}

The problem with this is that the content type is used by many http clients to automatically decode the json response to a struct so the implementation will be cluttered with workarounds.

## Essential Information

 - **Version**:      master
 - **Platform**:     Linux
 - **Installation**: Built from source
 - **Network**: testnet

rvl:
### Issue Number

#2596 / ADP-853

### Overview

- [x] Make the `LiftHandler` type class simpler.
- [x] Add content-type header for all uses of `apiError`
- [x] Add regression test.
- [x] Add content-type to generic api error handler tests.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -528,7 +528,7 @@ start
 start settings tr tlsConfig socket application = do
     logSettings <- newApiLoggerSettings <&> obfuscateKeys (const sensitive)
     runSocket
-        $ handleRawError (curry handler)
+        $ handleRawError (curry toServerError)
         $ withApiLogger tr logSettings
         application
   where
@@ -2440,12 +2440,12 @@ withWorkerCtx ctx wid onMissing onNotResponding action =
 -- | Maps types to servant error responses.
 class IsServerError e where
     -- | A structured human-readable error code to return to API clients.
-    handler :: e -> ServerError
+    toServerError :: e -> ServerError
 
 -- | Lift our wallet layer into servant 'Handler', by mapping each error to a
 -- corresponding servant error.
 liftHandler :: IsServerError e => ExceptT e IO a -> Handler a
-liftHandler action = Handler (withExceptT handler action)
+liftHandler action = Handler (withExceptT toServerError action)
 
 liftE :: IsServerError e => e -> Handler a
 liftE = liftHandler . throwE
@@ -2456,6 +2456,9 @@ apiError err code message = err
         [ "code" .= code
         , "message" .= T.replace "\n" " " message
         ]
+    , errHeaders =
+        (hContentType, renderHeader $ contentType $ Proxy @JSON)
+        : errHeaders err
     }
 
 data ErrUnexpectedPoolIdPlaceholder = ErrUnexpectedPoolIdPlaceholder
@@ -2476,7 +2479,7 @@ showT :: Show a => a -> Text
 showT = T.pack . show
 
 instance IsServerError ErrTemporarilyDisabled where
-    handler = \case
+    toServerError = \case
         ErrTemporarilyDisabled ->
             apiError err501 NotImplemented $ mconcat
                 [ "This endpoint is temporarily disabled. It'll be made "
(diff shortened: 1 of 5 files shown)


lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2434,23 +2437,28 @@ withWorkerCtx ctx wid onMissing onNotResponding action =
...
 
 apiError :: ServerError -> ApiErrorCode -> Text -> ServerError
 apiError err code message = err
     { errBody = Aeson.encode $ Aeson.object
         [ "code" .= code
         , "message" .= T.replace "\n" " " message
         ]
+    , errHeaders =
+        (hContentType, renderHeader $ contentType $ Proxy @JSON)
+        : errHeaders err
```
KtorZ, line 2461: :+1:

lib/core/test/unit/Cardano/Wallet/Api/ServerSpec.hs
```diff
@@ -214,3 +218,60 @@ spec = describe "API Server" $ do
...
+    it "ErrNoSuchWallet" $ do
+        let handler :: ExceptT ErrNoSuchWallet IO ()
+            handler = throwE $ ErrNoSuchWallet wid
+            wid = unsafeFromText "0000000000000000000000000000000000000000"
+        let expectedErr = ServerError
+                { errHTTPCode = 404
+                , errReasonPhrase = "Not Found"
+                , errBody = mconcat
+                    [ "{\"code\":\"no_such_wallet\",\"message\":\"I couldn't "
+                    , "find a wallet with the given id: 0000000000000000000000000000000000000000\"}" ]
```
KtorZ: We could avoid the message here, this is tested in other layers and will only be painful to maintain over time.

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
```diff
@@ -169,41 +173,40 @@ spec = parallel $ do
...
+assertErrorResponse
+    :: HasCallStack
+    => Int -- ^ Expected status
+    -> Text -- ^ Expected error code string
+    -> ExpectedError
+    -> SResponse
+    -> Session ()
+assertErrorResponse status code (ExpectedError msg) response = do
+    response & assertStatus status
+    response & assertHeader "Content-Type" "application/json;charset=utf-8"
```
KtorZ, line 185: :clap:

KtorZ (approved):

lib/core/test/unit/Cardano/Wallet/Api/ServerSpec.hs
rvl: OK I changed the assertions to be more lenient and maintainable.
As far as I can see, the error message is tested by integration tests but nowhere else.

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
rvl: These tests don't actually cover the `no_such_wallet` error reported in issue #2596.

rvl:
Thanks for the review comments.

bors r+

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
KtorZ: Yes indeed, the `404` are sort of cumbersome to trigger with this generic approach which only focuses on what exists.

PR #2604, opened 2021-04-06
Correct purpose differentiation between Icarus and Shelley

paweljakubas:
# Issue Number

adp-766

# Overview

- [x] correcting account -> hardened derivation
- [x] differentiate between Icarus and Shelley 

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
@@ -70,6 +70,8 @@ module Cardano.Wallet.Primitive.AddressDiscovery.Sequential
     -- ** Benchmarking
     , SeqAnyState (..)
     , mkSeqAnyState
+
+    , GetPurpose (..)
     ) where
 
 import Prelude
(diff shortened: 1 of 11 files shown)


lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
```diff
@@ -389,13 +391,17 @@ addresses mkAddress =
     . indexedKeys
   where
     toDerivationPath ix = NE.fromList $ map DerivationIndex
-        [ getIndex purposeBIP44
-        , getIndex purposeCIP1852
-        , 0   -- corresponds to account number
+        [ getIndex $ getPurpose @k
+        , getIndex coinTypeAda
+        , getIndex $ minBound @(Index 'Hardened 'AccountK)
```
KtorZ, line 396: Awe. Indeed. Sorry I didn't catch that earlier :facepalm:

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
```diff
@@ -389,13 +391,17 @@ addresses mkAddress =
...
         , fromIntegral $ fromEnum $ role @c
         , getIndex ix
         ]
 
+-- It is used for geting purpose for a given key.
+class GetPurpose key where
+    getPurpose :: Index 'Hardened 'PurposeK
```
KtorZ, line 403: :+1:

KtorZ (approved):

KtorZ:
Note: approving now, but our test coverage should've been / should be sufficient to cover this (thus, please make sure to upgrade the integration or unit test suite to cover this, they should fail before this change and pass after :pray:)

piotr-iohk:
I have added additional assertions to check if derivation path is valid for Shelley / Icarus and Byron addresses in the integration tests.

piotr-iohk (approved):
lgtm.

PR #2608, opened 2021-04-08
fix expect / include assertions in e2e tests & post release

piotr-iohk:
# Issue Number

n/a

# Overview

- 66dd5f1e48c9ec3dbea4c7023febe6690a4da07f
  fix expect / include assertions
  
- 606ef89d53c517625c9ffe5fe5844ba6dedba0a9
  update migration-tests.nix

# Comments

They suddenly started to fail, on checking if response includes some text. Probably due to some silent dependency update (perhaps httparty?). The fix is to explicitly cast response to string...

### DIFF
--- a/nix/migration-tests.nix
+++ b/nix/migration-tests.nix
@@ -38,10 +38,6 @@ let
   # One can get sha256 for release via nix-prefetch-url, e.g. for v2021.4.8:
   # nix-prefetch-url --unpack https://github.com/input-output-hk/cardano-wallet/archive/v2021-02-15.zip
   releases = [
-    { rev = "v2020-09-30";
-      sha256 = "19vmyq0m67ih295z1y8ng2rdn85mi0rvw26k711lbxsqri7fmyvz"; }
-    { rev = "v2020-10-13";
-      sha256 = "0kc1ddidp1vip8rfnm7vp5avx46xr9nvyzn7q9r3yn50ylhx0i0j"; }
     { rev = "v2020-11-03";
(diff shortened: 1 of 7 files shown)


rvl (commented):
OK, how mysterious?!

Could we keep a ruby gem lock file in git so that the dependencies don't randomly change?

piotr-iohk:
> OK, how mysterious?!

It is... I haven't expected this at all :sweat_smile:.

> Could we keep a ruby gem lock file in git so that the dependencies don't randomly change?

We do have [Gemfile.lock](https://github.com/input-output-hk/cardano-wallet/blob/master/test/e2e/Gemfile.lock) checked.
I suppose that's a downside of using [pessimistic](https://thoughtbot.com/blog/a-healthy-bundle#pessimistic-version) operator (`~>`) which tells "Bundler that it’s fine to upgrade a gem, as long as the gem version doesn’t change too much". On one hand that's good for automatically fixing/updating minor changes between dependencies as opposed to doing it manually. On the other hand I guess there is a risk of some break like this. I suppose that, in this particular case, it's also a downside of duck typing in ruby... Probably some change in the child dependency prevented the HTTPParty response object from being automatically casted to string when inisde `expect` assertion... That's my guess (but tbh I haven't spend time digging.)

I haven't ever experience any break like this before, let's see if this happens more and then maybe we can try to make dependencies more strict? Nevertheless it is a common practice, i.e. our dependencies most probably use `~>` for their dependencies etc.

rvl:
I think it would be best if dependency versions were locked to exact versions.

piotr-iohk:
Locked dependencies in eeb694a.

rvl (approved):
Good, thanks

PR #2609, opened 2021-04-09
e2e tests: additional checks for application/json response headers

piotr-iohk:
# Issue Number

ADP-853

# Overview

- [ ] I have added additional checks in e2e tests checking that responses return `'content-type' => ['application/json;charset=utf-8']` header

# Comments

Only `204` reponses do not return `'content-type' => ['application/json;charset=utf-8']` header. I assume that it is OK, since `204` returns no content?

I've temporarily cherry-picked c832eb50644e3810d5d67ea5a6c3c2449e1ae40a from #2608, to be able to run all tests.

### DIFF
--- a/test/e2e/Rakefile
+++ b/test/e2e/Rakefile
@@ -44,25 +44,25 @@ end
 task :fixture_wallets_template do
   puts "\n  >> Creating #{FIXTURES_FILE}"
 
-  fixture_wallets = { :linux => {:shelley => mnemonic_sentence(24),
+  fixture_wallets = { :linux => { :shelley => mnemonic_sentence(24),
                                  :icarus => mnemonic_sentence(15),
                                  :byron => mnemonic_sentence(12) },
-                      :macos => {:shelley => mnemonic_sentence(24),
+                      :macos => { :shelley => mnemonic_sentence(24),
                                  :icarus => mnemonic_sentence(15),
                                  :byron => mnemonic_sentence(12) },
-                      :windows => {:shelley => mnemonic_sentence(24),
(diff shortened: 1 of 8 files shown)


rvl (changes requested):
Could we refactor the response assertions to keep it maintainable?
Almost every request will have application/json content-type, except for a handful.

piotr-iohk:
> Could we refactor the response assertions to keep it maintainable?

@rvl you mean something like this -> https://github.com/input-output-hk/cardano-wallet/pull/2609/commits/9e65a6cfe183c5dfc8869ccdf44c5cd4783c934f

rvl:
Yes - a little. But it would be better if the response content type were checked automatically, so there doesn't need to be this assertion in every single test case. It could be tested when parsing the response (i.e. only decode as json if that's the content-type), or as part of the http status code check.

piotr-iohk:
Ok, gotcha. Currently it is like this:

```ruby
expect(response).to have_http 201
expect(response).to have_expected_headers
```
which I could change to something like:

```ruby
expect(response).to be_correct 201
```
where I could check both assertions, with additional logic that when response code is `204` the `application/json` header is not expected.

I guess the pros is that it is more concise, cons that two assertions combined in one... which is sort of anti-pattern I suppose. :thinking:

piotr-iohk:
>Yes - a little. But it would be better if the response content type were checked automatically, so there doesn't need to be this assertion in every single test case.

Done in 27df4ce.

rvl (approved):
:+1:

rvl:
I don't think it's an anti-pattern. I think it's better to explicitly test more high level things.

PR #2615, opened 2021-04-15
More misc logging tweaks and preparation for aggregation

Anviking:
# Issue Number

# Overview

**Split off from #2583** 

See commit messages for overview.

# Comments

Should be the more straight-forward and auxiliary parts of #2583. 

Maybe af6498b98343bc51b791bb1613d2f5b9fd63b93d shouldn't be here though, as it's a slightly weird intermediate state — but I think it should work.


lib/core/src/Cardano/Wallet.hs
```diff
@@ -776,61 +778,58 @@ restoreWallet
...
-        , HasDBLayer s k ctx
-        )
+    :: forall ctx s k. (HasDBLayer s k ctx)
     => ctx
+    -> Tracer IO WalletFollowLog
     -> WalletId
     -> SlotNo
     -> ExceptT ErrNoSuchWallet IO ()
-rollbackBlocks ctx wid point = db & \DBLayer{..} -> do
+rollbackBlocks ctx tr wid point = db & \DBLayer{..} -> do
```
rvl, line 826: Just wondering out loud ... Perhaps we could even just remove the Tracer from `ctx`, and explicitly pass contramapped tracers around.

lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -254,25 +251,27 @@ data FollowExit
 --
 -- Exits with 'Nothing' in case of error.
 follow
-    :: forall block e. (Show e)
+    :: forall block e tr. (Show e)
```
rvl, line 254: Perhaps `tr` -> `msg` ?

rvl (approved):
Nice :+1:

rvl:
I have rebased and resolved merge conflicts, though it may not fully compile now.

Anviking:
Ah, thanks!

lib/core/src/Cardano/Wallet.hs
Anviking: 🤔 I guess it would make it more difficult for functions in the Server module to call the functions here

lib/core/src/Cardano/Wallet/Network.hs
Anviking: Kind of liked the `tr`, but I guess that makes sense as `Tracer IO tr` would appear like a tracer of tracer 👍 

Adding a commit in the other PR.

PR #2618, opened 2021-04-19
Multi-Asset Migration Algorithm

jonathanknowles:
# Issue Number

ADP-839

# Overview

This PR adds an algorithm for performing migrations of multi-asset UTxO sets from one wallet to another.

It consists of four main parts:

1. Module `Migration`, which provides a top-level public API, defined in terms of concrete wallet types.
2. Module `Migration.Planning`, which contains an algorithm for planning migrations at a high
level. It determines how to partition a UTxO set into entries of different types, and in which order to add UTxO entries to selections.
3. Module `Migration.Selection`, which provides functions for incrementally constructing a
selection to be included in a migration plan. (A selection is the basis for an individual transaction.)
4. The `TxConstraints` type, which provides an abstract cost model for transactions, allowing parts of a transaction can be costed individually. This allows the incremental cost of extending a selection to be calculated without having to recalculate the cost of the entire selection.

# Performance

For pure ada wallets, we can process around 30,000 UTxO entries per second. 

We can process around 3,000 UTxO entries per second with wallets of the following composition:

| Percentage | UTxO Entry Type |
| -- | -- |
| 10% | Pure ada entries below the minimum ada quantity |
| 40% | Pure ada entries with at least the minimum ada quantity |
| 40% | Multi-asset token bundles with precisely the minimum ada quantity |
| 10% | Multi-asset token bundles with more than the minimum ada quantity |

# Design

## UTxO partitioning

The algorithm starts by partitioning the UTxO set into three types of UTxO entry:

| Type | Definition |
| -- | -- |
| **_supporter_** | an entry that is capable of paying for itself |
| **_freerider_** | an entry that is **not** capable of paying for itself |
| **_ignorable_** | a dust ada entry that should not be included because its value is less than the marginal cost of an input. |

## Adding entries to transactions

The algorithm uses a **supporter** entry to initialize each transaction, but then gives priority to adding **freerider** entries until the transaction is full. It adds a **supporter** entry only when there's not enough ada available to pay for an additional **freerider** entry.

extending a selection

## Creating outputs within transactions

The algorithm minimizes the number of outputs for each transaction (and thus minimizes the fee required) by merging together all value provided by the inputs. Due to the nature of our transaction constraints (which limit the size of an output), the algorithm usually produces transactions with between 1 and 4 outputs.

# Requirements

The algorithm in this PR is designed to satisfy the following requirements:

-  **REQ-1**
  The algorithm _must_ succeed at selecting all pure ada entries greater than `txBaseFee` + `minAdaQuantity` + `marginalFeeForInputAndOutput`.
- **REQ-2**
  The algorithm _must_ succeed at selecting all multi-asset entries whose ada quantities are greater than or equal to `txBaseFee` + `minAdaQuantity` + `marginalFeeForInputAndOutput`.
- **REQ-3**
  The algorithm _should_ make a best effort to select entries whose ada quantities are less than `txBaseFee` + `minAdaQuantity` + `marginalFeeForInputAndOutput`, but is permitted to omit such entries if it cannot find a viable strategy to select them.
- **REQ-4**
  The algorithm _must_ produce **T**: a set of transactions for which fees are balanced, and **R**: the remainder of the UTxO set that could not be selected, such that the inputs of **T** and the entries in **R** are disjoint, and such that the union of inputs in **T** and entries in **R** is equal to the original UTxO set.
- **REQ-5**
  The algorithm _must_ preserve the balance of every non-ada asset, so that for any given asset **a**, the total quantity of asset **a** in the selected inputs is exactly equal to the total quantity of asset **a** in the generated outputs.
- **REQ-6**
  The algorithm _must not_ produce transactions with dependencies between them; it should be possible to broadcast the generated transactions in any order.
- **REQ-7**
  The algorithm _must not_ produce outputs with token quantities that exceed the maximum allowed in an output `(2^64 − 1)`.
- **REQ-8**
  The algorithm _must not_ produce outputs whose serialized length exceeds the maximum allowed length (`4000` bytes).
- **REQ-9**
  The algorithm _must not_ produce transactions that exceed the transaction size limit (`16384` bytes).
- **REQ-10**
  The algorithm _must_ move the reward balance in its entirety.
- **REQ-11**
  The algorithm _must_ be deterministic: if run twice on the same UTxO set, it must generate exactly the same migration plan. (This is necessary to support an API where a non-hardware wallet user can generate a plan, inspect the plan, and then initiate a migration without having to resubmit the plan.)

# Non-Requirements

- **NON-REQ-1**
  Preserving the UTxO distribution.


jonathanknowles:
Had a review call with @Anviking, where we identified a number of things to add.

**To add to this PR:**

- [x] Write a comment for `Selection.balance`, to explain its purpose, what it expects and what it guarantees.
    (Fixed in c6eb01a1d0c6f5e1c8e2cae46571f29831aa4eac)
- [x] Write a comment for `Selection.minimizeFee` to explain the motivation for coalescing most of the ada into one of the outputs (rather than spreading it evenly across all outputs).
    (Fixed in c6eb01a1d0c6f5e1c8e2cae46571f29831aa4eac)
- [x] Write a comment for `Selection.addValueToOutputs` to explain its purpose, what it expects, what it guarantees.
    (Fixed in c6eb01a1d0c6f5e1c8e2cae46571f29831aa4eac)
- [x] In the property test for `Planning.createPlan`, use `QuickCheck.tabulate` instead of `QuickCheck.label` for whenever there is a range of values, but a shared title.
    (Fixed by b2186bdae2d4d3c95559099fef1a8320bb9ea4a8)
- [x] In the property test for `Planning.createPlan`, instead of reporting how many of each category were **_not_** selected, report how many **_were_** selected, thus making the results easier to interpret (avoiding a double negative).
    (Fixed by b2186bdae2d4d3c95559099fef1a8320bb9ea4a8)
- [x] Add a property test to show that `uncategorizeUTxOEntries . categorizeUTxOEntries == id`.
    (Fixed by cd78eae976fcd70d896c5bdef75d07b9b23a9cac)
- [x] Add a property test for `Migration.createPlan` which checks that `utxoProvided == union utxoNotSelected utxoSelected`.
    (Fixed by 152d9579c2d19f1b58979571d84bc0d7609f1702)
- [x] Add a property test for `Migration.createPlan` which checks that `Set.empty == intersect utxoNotSelected utxoSelected`.
    (Fixed by eee3f2eaa181d4c7759acd775acceb0bda9c9d51)
- [x] Add a property test for `Migration.createPlan` to verify that it gives us the same result as calling the inner `Planning.createPlan`.
    (Fixed by eee3f2eaa181d4c7759acd775acceb0bda9c9d51)
- [x] In the property test for `Selection.create`, check that the fee has been minimized.
(Fixed in adf7dd0ad75b153f86c6e18b4e8340a302107221)
- [x] In the property test for `Selection.extend`, check that the fee has been minimized.
(Fixed in adf7dd0ad75b153f86c6e18b4e8340a302107221)

**To add to the subsequent PR (integration with the Shelley transaction layer):**

- Test all the properties again, but with real Shelley transaction constraints rather than mock constraints. This will require us to make the properties reusable from the `cardano-wallet-shelley` test suite. (We can use the same approach that we have for the `Gen` modules.)
- Add an API integration test to check that calling `getMigrationInfo` gives the same result when run twice.
- In the Shelley transaction layer tests, we should check that the fee of each selection is not significantly different from calling "computeFee . estimateSize" (a pre-existing function in the `Shelley.Transaction` module).
- Think of a way to build a fail-safe sanity check for fee minimization.

lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
```diff
@@ -0,0 +1,787 @@
...
+        -- outputs into ascending order according to the number of assets that
+        -- would need to be added to each output.
+        --
+        -- In the absolute ideal case, where an existing output's assets are a
+        -- superset of the output value to be added, merging with that output
+        -- will not increase its asset count.
+        --
+        -- As a tie-breaker, we give priority to outputs with smaller numbers
+        -- of assets. Merging with a smaller output is more likely to succeed,
+        -- because merging with a larger ouput is more likely to fall foul of
```
jonathanknowles: Small typo:
```suggestion
        -- because merging with a larger output is more likely to fall foul of
```

lib/core/src/Cardano/Wallet/Primitive/Migration/Planning.hs
```diff
@@ -0,0 +1,271 @@
...
+    }
+    deriving (Eq, Show)
+
+categorizeUTxO
+    :: TxSize s
+    => TxConstraints s
+    -> UTxO
+    -> CategorizedUTxO (TxIn, TxOut)
+categorizeUTxO constraints (UTxO u) = categorizeUTxOEntries constraints $
+    (\(i, o) -> ((i, o), view #tokens o)) <$> Map.toList u
```
Anviking, line 258: It seems `categorizeUTxO` and `uncategorizeUTxO` are both needed only in `createPlan`. 

How about moving the `entry <-> UTxO` conversion there as part of a `where`-clause? It might be a bit clearer division of responsibility then.

lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
```diff
@@ -0,0 +1,787 @@
...
+     (SelectionSizeIncorrectError s)
+    deriving (Eq, Show)
+
+verify
+    :: TxSize s
+    => TxConstraints s
+    -> Selection i s
+    -> SelectionCorrectness s
+verify constraints selection
+    | Just e <- checkAssetBalance selection =
```
Anviking: I hadn't seen this style before. Wouldn't 
```haskell
checkAssetBalance :: Selection i s -> Either SelectionAssetBalanceIncorrectError ()
```
and
```haskell
verify = do
    left SelectionAssetBalanceIncorrect $ checkAssetBalance selection
    left SelectionFeeIncorrect $ ...
    ...
```
be more plain?

lib/core/test/unit/Cardano/Wallet/Primitive/Migration/PlanningSpec.hs
```diff
@@ -0,0 +1,367 @@
...
+        [ pure (Coin 0)
+        , genCoinRange (Coin 1) (Coin 1_000_000)
+        ]
+    pure ArgsForCreatePlan
+        { mockConstraints
+        , mockInputs
+        , mockRewardWithdrawal
+        }
+
+prop_createPlan :: ArgsForCreatePlan -> Property
```
Anviking: I wonder: isn't this `ArgsX` pattern similar to `forAll` 🤔 

Haven't done this myself, but I think you could have 
```haskell
prop_createPlan :: MockTxConstraints -> MockWithdrawal -> Property
prop_createPlan constraints (MockWithdrawal wdrl) = forAll (genMockInputs constraints) $ \inputs -> 
```
such that the generators are only _as coupled_ as necessary but no more, and you can more easily have shrinkers for the constraints and the withdrawal.

lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
jonathanknowles: I agree that the control flow could be a little clearer.

What do you think about this solution: d31d1ec5a72202561fb4f15012ec6ba1dbf1d229?

lib/core/src/Cardano/Wallet/Primitive/Migration/Planning.hs
jonathanknowles: > It seems categorizeUTxO and uncategorizeUTxO are both needed only in createPlan.
> How about moving the entry <-> UTxO conversion there as part of a where-clause? It might be a bit clearer division of responsibility then.


Good idea! I've fixed this in d4819813bff15429724003d79aa489a77327a1fa.

lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
jonathanknowles: Fixed in 9519bdd9d2aaa404821866dec171bd8557a2c05c.

lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
```diff
@@ -0,0 +1,796 @@
...
+import qualified Data.Foldable as F
+import qualified Data.List as L
+import qualified Data.List.NonEmpty as NE
+import qualified Data.Set as Set
+
+--------------------------------------------------------------------------------
+-- Selections
+--------------------------------------------------------------------------------
+
+data Selection i s = Selection
```
jonathanknowles: Perhaps use more descriptive type parameters here.

Instead of `i`, use `inputId`.
Instead of `s`, use `size`.

lib/core/test/unit/Cardano/Wallet/Primitive/Migration/SelectionSpec.hs
```diff
@@ -0,0 +1,1036 @@
...
+        "Success" $
+    cover 10 (resultHasZeroFeeExcess result)
+        "Success with zero fee excess" $
+    cover 0.1 (resultHasInsufficientAda result)
+        "Failure due to insufficient ada" $
+    cover 0.1 (resultIsFull result)
+        "Failure due to oversized selection" $
+    case result of
+        Left SelectionAdaInsufficient ->
+            property True
```
jonathanknowles, line 296: Suggestion: instead of returning `True` here, call `Selection.create` to verify that this function will also produce the same error.

lib/core/test/unit/Cardano/Wallet/Primitive/Migration/SelectionSpec.hs
```diff
@@ -0,0 +1,1036 @@
...
+        Left (SelectionFull e) ->
+            property (selectionSizeMaximum e < selectionSizeRequired e)
```
jonathanknowles, line 298: Suggestion: also call `Selection.create` here to verify that this function will also produce the same error.

lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
Anviking: I think `size` makes sense if keeping it 👍 

How about just `input` instead of `inputId`? I don't get why this distinction is preferable.

It seems `i ~ (TxIn, TxOut)` for the real algorithm also... Doesn't seem like an "id"?

Anviking:
> Test all the properties again, but with real Shelley transaction constraints rather than mock constraints. This will require us to make the properties reusable from the cardano-wallet-shelley test suite. (We can use the same approach that we have for the Gen modules.)

Just a through: You said you didn't just want to use the real constraints, I believe for the sake of coverage, shrinking etc. Couldn't you simply have generators which 50% use real constraints and 50% use mock constraints?

Although I guess then you would never have coverage reports using the real parameters alone, which might be useful.

lib/core/test/unit/Cardano/Wallet/Primitive/MigrationSpec.hs
```diff
@@ -0,0 +1,174 @@
...
+
+import Cardano.Wallet.Primitive.Migration
+    ( RewardWithdrawal (..), createPlan )
+import Cardano.Wallet.Primitive.Migration.Planning
+    ( categorizeUTxO, uncategorizeUTxO )
+import Cardano.Wallet.Primitive.Migration.SelectionSpec
+    ( MockTxConstraints
+    , genCoinRange
+    , genMockTxConstraints
+    , genTokenBundleMixed
```
Anviking, line 16: I think this is technically against our coding standards:

> [**Test files are separated and self-contained**](https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#test-files-are-separated-and-self-contained)
> It is really easy to make the testing code more complex than the actual code it's initially testing. Limiting the interaction between test modules helps keeping a good maintainability and a rather low overhead when it comes to extend, modify, read or comprehend some tests. Also, in many cases, we do actually want to have different arbitrary generators for different test cases so sharing instances is risky and cumbersome.

lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
jonathanknowles: Okay, I've gone for `input` and `size`, as suggested.

Fixed in 2a2f17dc8384a68ff2e968f39a125d0a02f1e926.

> It seems i ~ (TxIn, TxOut) for the real algorithm also... Doesn't seem like an "id"?

Thinking about this, `createPlan` actually only needs to return a `MigrationPlan` where `i ~ TxIn`, since the caller of `createPlan` can figure out which `TxOut` each `TxIn` refers to, if they really need to do so.

Perhaps we can change this to be `i ~ TxIn`.

lib/core/test/unit/Cardano/Wallet/Primitive/MigrationSpec.hs
jonathanknowles: > Limiting the interaction between test modules helps keeping a good maintainability and a rather low overhead when it comes to extend, modify, read or comprehend some tests.

I think there's a clear tension between **inter-module interaction** (which we usually want to minimize) and **code reuse** (which might be desirable).

I agree we **definitely** want to avoid the problem of "spooky action at a distance"! (Where we change something in one module that has a nefarious effect on another module.)

In the past, we had many `Arbitrary` instances that were shared between test modules. Of course, shared `Arbitrary` instances have the following disadvantages:
 - When looking at a given call to `arbitrary`, it's often hard to tell where the instance is defined. You first need to know the type, and then need to check the module dependency graph.
 - When changing the definition of an instance, it's hard to tell which other modules and tests will be affected.
 - Instances may be defined in terms of more calls to `arbitrary`, which makes solving the instance resolution graph even harder.
  
In this PR, we've not shared any `Arbitrary` instances between modules. (Except, perhaps, for the `Arbitrary (NonEmpty a)` instance, which is relatively innocuous.)

In every case of code sharing, we've explicitly imported just the generator functions we need. So it should be easy for a reader to figure out where something is coming from, and there should be no "spooky action at a distance".

In my mind, this is a good balance: we get the code reuse we want, and the reader can see explicitly where each piece of code comes from.

Having said this, I think we might end up (in the next PR), moving some of these generators to a `Gen` module, so that we can reuse them from the Shelley test suite.

jonathanknowles:
> Just a through: You said you didn't just want to use the real constraints, I believe for the sake of coverage, shrinking etc. Couldn't you simply have generators which 50% use real constraints and 50% use mock constraints?
> Although I guess then you would never have coverage reports using the real parameters alone, which might be useful.

I agree with you that we really need to get coverage with the real parameters alone.

I think it's clear that we need the following:

1. A way to get fast iteration and feedback when adjusting something in `Primitive.Migration`.
2. A way to test with the real Shelley parameters.

But the separation between `cardano-wallet-core` and `cardano-wallet` (Shelley) means that our primary test suite for `Primitive.Migration` cannot depend on the real parameters, unless we make our mock parameters identical. But making them identical might not be desirable: we sometimes want to use parameters that are smaller (for example), to enable tests to run faster, or to test edge cases.

I think there's a good compromise solution here:

1. Define reusable generators and properties in terms of `TxConstraints`.
2. Call these properties with `genMockTxConstraints` in the `cardano-wallet-core` test suite (enabling rapid feedback.)
3. Call these properties with `shelleyTxConstraints` in the `cardano-wallet` test suite (giving confidence that everything works with the real constraints.)

From the above, we already have parts **1** and **2**. The next PR can add part **3** (though it might need us to move some of the generators and properties to a shared `Gen` module).

lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
```diff
@@ -0,0 +1,801 @@
...
+    -> (Coin, NonEmpty TokenBundle)
+    -- ^ Fee excess and output bundles after optimization.
+minimizeFee constraints (currentFeeExcess, outputs) =
+    NE.fromList <$> run currentFeeExcess (NE.toList outputs) []
+  where
+    run :: Coin -> [TokenBundle] -> [TokenBundle] -> (Coin, [TokenBundle])
+    run (Coin 0) remaining processed =
+        (Coin 0, L.reverse processed <> remaining)
+    run feeExcessRemaining [] processed =
+        (feeExcessRemaining, L.reverse processed)
```
jonathanknowles: Suggestion: remove these reversals, as they are almost certainly unnecessary.

Their existence is a hangover from an earlier iteration of the design, where maintaining the order of outputs was important.

jonathanknowles: Fixed in f9de1b4dad3c642b350217c0afad318ae058e9ec.

lib/core/test/unit/Cardano/Wallet/Primitive/Migration/PlanningSpec.hs
```diff
@@ -0,0 +1,418 @@
...
+    arbitrary = Small <$> genArgsForCreatePlan
+        (0, 100) genMockInput
+
+instance Arbitrary (Large ArgsForCreatePlan) where
+    arbitrary = Large <$> genArgsForCreatePlan
+        (1_000, 1_000) genMockInput
+
+instance Arbitrary (Giant ArgsForCreatePlan) where
+    arbitrary = Giant <$> genArgsForCreatePlan
+        (100_000, 100_000) genMockInputAdaOnly
```
jonathanknowles: With recent optimizations, I think we can relax the restriction of using ada-only inputs for the "giant" test, provided that we reduce the "giant" size to something more manageable, such as 10,000 entries. (Which should take around 3 seconds.)

jonathanknowles: Fixed in 522a895a11384b9faebc6868997f08f836ef0fde.

lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
```diff
@@ -0,0 +1,956 @@
...
+
+--------------------------------------------------------------------------------
+-- Selection correctness
+--------------------------------------------------------------------------------
+
+-- | Indicates whether or not a selection is correct.
+--
+data SelectionCorrectness size
+    = SelectionCorrect
+    | SelectionIncorrect (SelectionCorrectnessError size)
```
Anviking: Could we replace this with just `Either (SelectionCorrectnessError size) ()`? To me it doesn't seem worth it's own type.

lib/core/test/unit/Cardano/Wallet/Primitive/Migration/PlanningSpec.hs
jonathanknowles: Hi @Anviking 

As requested, I've used `forAll` and `forAllShrink` to replace all the `ArgsX` types.

Thanks for the suggestion, I think it has really simplified the test code. (Much less boilerplate!)

lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
```diff
@@ -0,0 +1,956 @@
...
+-- License: Apache-2.0
+--
+-- This module contains functions for incrementally constructing a selection
+-- to be included in a migration plan.
+--
+-- A selection is the basis for a single transaction.
+--
+-- Use 'create' to create a selection with one or more inputs.
+-- Use 'extend' to extend a selection with an additional input.
+-- Use 'verify' to verify the correctness of a selection.
```
Anviking:
```suggestion
```

This makes it sound like you should be using `verify` when using this module, when in reality it is used only for testing. 

I think if you simply drop this line, the module doc becomes really neat

lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
```diff
@@ -0,0 +1,956 @@
...
+      SelectionOutputBelowMinimumAdaQuantityError
+    | SelectionOutputSizeExceedsLimit
+      SelectionOutputSizeExceedsLimitError
+    | SelectionSizeExceedsLimit
+     (SelectionSizeExceedsLimitError size)
+    | SelectionSizeIncorrect
+     (SelectionSizeIncorrectError size)
+    deriving (Eq, Show)
+
+-- | Verifies a selection for correctness.
```
Anviking: How about just adding:
```suggestion
-- | Verifies a selection for correctness. For testing, as selections from @create@ and @extend@ are guaranteed to be correct.
```
to make it clear that you're not normally supposed to care about it.

Also I forgot exactly the rationale for having it in the library code, but _maybe_ you could add a quick word about that.

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
```diff
@@ -493,3 +503,66 @@ txOutMinTokenQuantity = TokenQuantity 1
...
+--------------------------------------------------------------------------------
+-- Constraints
+--------------------------------------------------------------------------------
+
+-- | Provides an abstract cost model for transactions.
+--
+-- This allows parts of a transaction to be costed individually, without
+-- having to compute the cost of an entire transaction.
+--
+data TxConstraints s = TxConstraints
```
Anviking, line 524: This type will never be used when say listing Txs from DB.

It's currently used from Migration.Selection, Migration.Planning and Migration.

Is not Migration.Selection a better module for it?

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
```diff
@@ -493,3 +503,66 @@ txOutMinTokenQuantity = TokenQuantity 1
...
+    { txBaseCost :: Coin
+      -- ^ The constant cost of an empty transaction.
+    , txBaseSize :: s
```
Anviking, line 527: Suggestion: to get rid of some boilerplate, you _could_ do:

```haskell
data SizeAndCost s = SizeAndCost
    { size :: s
    , cost :: Coin
    }

data TxConstraints s = TxConstraints
    { txBase :: SizeAndCost s
      -- ^ The constant size and cost of an empty transaction.
    , txInput :: SizeAndCost s
      -- ^ The size and cost of including an extra input to a transaction.
    , txOutput :: TokenBundle -> SizeAndCost s
      -- ^ The size and cost of including an extra output to a transaction.
    , txRewardWithdrawal :: Coin -> SizeAndCost s
      -- ^ The variable size and cost of a reward withdrawal.
    , txOutputMaximumSize :: s
      -- ^ The maximum size allowed
    , txOutputMaximumTokenQuantity :: TokenQuantity
      -- ^ The maximum token quantity that can appear in a transaction output.
    , txOutputMinimumAdaQuantity :: TokenMap -> Coin
      -- ^ The variable minimum ada quantity of a transaction output.
    , txMaximumSize :: s
      -- ^ The maximum size of a transaction.
    }
```

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
Anviking: Although having it next to the similar `TokenBundleSizeAssessor` makes sense, so maybe keep it.

lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
Anviking: Also, looking _now_, I wonder if having the correctness checks in this module actually makes sense. Maybe could be moved later, if an opportunity arises.

lib/core/src/Cardano/Wallet/Primitive/Migration/Planning.hs
```diff
@@ -0,0 +1,307 @@
...
+--
+-- This function terminates when the selection cannot be extended further
+-- (because doing so would cause it to exceed the size limit of a transaction),
+-- or when there are no more UTxO entries available for selection.
+--
+-- Priority is given to selecting "freerider" entries: entries that cannot pay
+-- for themselves. A "supporter" entry is only added to the selection if there
+-- is not enough ada to pay for a "freerider" entry.
+--
+extendSelection
```
Anviking:
```suggestion
fillSelection
```
?
Otherwise I would have thought it only performed one step.

Anviking (commented):
Some stylistic comments; will review more

lib/core/test/unit/Cardano/Wallet/Primitive/Migration/PlanningSpec.hs
Anviking: Nice! 🎉

lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
Anviking: I think it reads better, but I don't see why we need `SelectionCorrectness` https://github.com/input-output-hk/cardano-wallet/pull/2618#discussion_r617352652

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
```diff
@@ -493,3 +503,66 @@ txOutMinTokenQuantity = TokenQuantity 1
...
+      -- ^ The constant size of an empty transaction.
+    , txInputCost :: Coin
+      -- ^ The constant cost of a transation input.
```
Anviking:
```suggestion
      -- ^ The constant cost of a transation input.
      -- 
      -- NOTE: For UTxOs sharing the same payment key this function and the current wallet in general will slightly overestimate the cost / fee, as it assumes one witness per input. (https://github.com/input-output-hk/cardano-wallet/blob/2acc0305e51d5b725701897b96cbc1ed8a61a477/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs#L582-L584)
```

I think this deserves mentioning

lib/core/src/Cardano/Wallet/Primitive/Migration/Planning.hs
Anviking: d4819813bff15429724003d79aa489a77327a1fa seems to have been dropped. Was this a mistake?

lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
```diff
@@ -0,0 +1,956 @@
...
+--------------------------------------------------------------------------------
+-- Selection correctness: minimum ada quantities
+--------------------------------------------------------------------------------
+
+data SelectionOutputBelowMinimumAdaQuantityError =
+    SelectionOutputBelowMinimumAdaQuantityError
+        { outputBundle :: TokenBundle
+          -- ^ The output that is below the expected minimum ada quantity.
+        , expectedMinimumAdaQuantity :: Coin
+          -- ^ The expected minimum ada quantity.
```
Anviking: Tbh, wondering if all of this error-boilerplate is worth it. Couldn't `verify` simply be an `Either String ()` in the spec, or something like that?

(Similar to https://github.com/input-output-hk/cardano-wallet/pull/2618/files#r617676467)

lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
jonathanknowles: > Could we replace this with just Either (SelectionCorrectnessError size) ()? To me it doesn't seem worth it's own type.

Which of these do you think is clearer for the reader: (extract from test code)

```hs
Selection.verify constraints selection === SelectionCorrect
```

```hs
Selection.verify constraints selection === Right ()
```

I think the first example is much clearer, as it doesn't require the reader to understand what `Right ()` means in this particular context.

> Also, looking now, I wonder if having the correctness checks in this module actually makes sense. Maybe could be moved later, if an opportunity arises.

The `Selection.create`, `Selection.extend` and `Selection.balance` functions all refer to `Selection.verify` in their documentation:

```hs
-- | Creates a selection with the given inputs.
--
-- Guarantees the following property for a returned selection 's':
--
-- >>> verify s == SelectionCorrect
```
This gives a simple, concise statement of what these functions guarantee, without having to list all the properties that the resultant selection should satisfy. The reader can navigate to `verify`, and see for themselves.

If we were to move the `verify` function into the test suite, we'd presumably have two options:

1. Within the comments for `create`, `extend` and `balance`, list all the properties that they satisfy. But there are many such properties. And we'd have to write them out three times, or refer to a shared comment. Comments are not compiler checked, and it's likely that they'd bit rot over time.

2. Within the comments for `create`, `extend` and `balance`, tell the reader to look at the properties in `SelectionSpec`. But this arguably increases the burden of understanding for the reader, as they now have to navigate to a different module in a different subproject just to understand the properties that these functions guarantee.

lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
jonathanknowles: I agree that removing this line wouldn't be harmful.

In any case, if the user navigates to `create` or `extend`, they'll soon find out that the comments for these functions refer to `verify` as a way to describe what they guarantee.

jonathanknowles: Fixed in 7bd44365061e0dfc9d435a0ffcb34c12d1dc8fa5.

lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
jonathanknowles: @Anviking We actually already use a similar technique within `UTxOIndex`:

https://github.com/input-output-hk/cardano-wallet/blob/2acc0305e51d5b725701897b96cbc1ed8a61a477/lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs#L490-L501

https://github.com/input-output-hk/cardano-wallet/blob/2acc0305e51d5b725701897b96cbc1ed8a61a477/lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs#L504-L506

lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
jonathanknowles: > to make it clear that you're not normally supposed to care about it.

Thanks for this suggestion. I've added a comment in ec6970e420525f13d223a70b33eb3f81a3fb6f46.

> Also I forgot exactly the rationale for having it in the library code, but maybe you could add a quick word about that.

See: https://github.com/input-output-hk/cardano-wallet/pull/2618#discussion_r618171707

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
jonathanknowles: I agree this would reduce the number of fields, but I'm not entirely convinced the benefit would exceed the cost!

In particular, I find the revised field names to be less clear. For example, at first glance, `txInput` seems like it ought to return an input, but actually it's the cost and size **_of_** an input. IMO, `txInputSize` and `txInputCost` seem more straightforward and easy to understand.

lib/core/src/Cardano/Wallet/Primitive/Migration/Planning.hs
jonathanknowles: Thanks for the suggestion to rename this. I agree it could be a little clearer.

I've renamed this to `extendSelectionUntilFull`, as it does indeed repeatedly extend a selection until it's full.

Fixed in 97f5c261b6d23b65e14dcb67c72101a62db8b58d.

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
jonathanknowles: Good point!

I've added a comment in a95d51bb3726359c971ae62b92379522ff38c05e.

Anviking (approved):
LGTM.

PR #2619, opened 2021-04-21
Shared wallets testing

piotr-iohk:
# Issue Number

ADP-660

# Overview

- [ ] I have added e2e tests using also cardano-addresses from the bundle
- [ ] marked shared eps as experimental for the time being

# Comments

### DIFF
--- a/specifications/api/swagger.yaml
+++ b/specifications/api/swagger.yaml
@@ -5137,7 +5137,7 @@ paths:
       tags: ["Shared Wallets"]
       summary: Create
       description: |
-        <p align="right">status: <strong>experimental</strong></p>
+        <p align="right">status: <strong>⚠ under development</strong></p>
 
         Create a shared wallet from an account public key, its index and script templates.
       requestBody:
@@ -5153,7 +5153,7 @@ paths:
       tags: ["Shared Wallets"]
       summary: Get
       description: |
-        <p align="right">status: <strong>experimental</strong></p>
+        <p align="right">status: <strong>⚠ under development</strong></p>
       parameters:
         - *parametersWalletId
       responses: *responsesGetSharedWallet
@@ -5174,7 +5174,7 @@ paths:
       tags: ["Shared Wallets"]
       summary: Update Payment
       description: |
-        <p align="right">status: <strong>experimental</strong></p>
+        <p align="right">status: <strong>⚠ under development</strong></p>
       parameters:
         - *parametersWalletId
       requestBody:
@@ -5190,7 +5190,7 @@ paths:
       tags: ["Shared Wallets"]
       summary: Update Delegation
       description: |
-        <p align="right">status: <strong>experimental</strong></p>
+        <p align="right">status: <strong>⚠ under development</strong></p>
       parameters:
         - *parametersWalletId
       requestBody:
--- a/specifications/api/swagger.yaml
+++ b/specifications/api/swagger.yaml
@@ -5137,7 +5137,7 @@ paths:
       tags: ["Shared Wallets"]
       summary: Create
       description: |
-        <p align="right">status: <strong>stable</strong></p>
+        <p align="right">status: <strong>experimental</strong></p>
 
         Create a shared wallet from an account public key, its index and script templates.
       requestBody:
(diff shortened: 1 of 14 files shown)


test/e2e/Gemfile
```diff
@@ -1,6 +1,7 @@
 source 'https://rubygems.org'
 
-gem 'cardano_wallet', '0.3.2'
+gem 'cardano_wallet', '0.3.4'
```
rvl: I like pinning of all the other deps, but this one could have pessimistic versioning, because we control it.

test/e2e/Gemfile
```diff
@@ -1,6 +1,7 @@
...
+# gem 'cardano_wallet', path: "~/wb/cardano_wallet"
```
rvl, line 4: Does `gem` provide a command to link in a local package for development?
I remember that python pip had the `pip install --editable ~/project` option..

test/e2e/helpers/utils.rb
```diff
@@ -5,8 +5,21 @@
...
+      puts cmd.gsub(/\s+/, ' ')
+      res = `#{cmd}`
+      puts res
+      res
+    end
+
+    def cardano_address_get_acc_xpub(mnemonics, derivation_path)
+      bins = absolute_path ENV['TESTS_E2E_BINDIR']
+      bins = (bins == '' ? bins : "#{bins}/")
+      bins = '' if is_win?
```
rvl: The standard way of locating executables is by using the `PATH` environment variable.
We should almost never need to add paths when spawning commands.
The test framework could add `TESTS_E2E_BINDIR` onto the front of `PATH` at startup.

test/e2e/spec/shelley_spec.rb
```diff
@@ -2,7 +2,7 @@
 
   describe CardanoWallet::Shelley::Wallets do
 
-    before(:each) do
+    after(:each) do
```
rvl, line 5: :slightly_smiling_face:

test/e2e/spec/spec_helper.rb
```diff
@@ -75,6 +76,31 @@
...
+                        { "cosigner#0": "#{acc_xpub}" },
+                      "template":
+                          { "all":
+                             [ "cosigner#0",
+                               "cosigner#1"
+                             ]
+                          }
+                      })
+  pscript = JSON.parse(script_template)
+  dscript = JSON.parse(script_template)
```
rvl: The json conversion step isn't needed, is it?
The script template could stay as a ruby dict.

specifications/api/swagger.yaml
```diff
@@ -5137,7 +5137,7 @@ paths:
       tags: ["Shared Wallets"]
       summary: Create
       description: |
-        <p align="right">status: <strong>stable</strong></p>
+        <p align="right">status: <strong>experimental</strong></p>
```
rvl: We should probably upgrade this to "under development", just to leave no doubt.

rvl (commented):
Great

test/e2e/Gemfile
piotr-iohk: It seems there are other options https://stackoverflow.com/questions/26876000/ruby-gem-equivalent-of-pip-install-e, this being one of them.

test/e2e/helpers/utils.rb
piotr-iohk: Done.

test/e2e/spec/spec_helper.rb
piotr-iohk: Done.

test/e2e/spec/shared_spec.rb
```diff
@@ -0,0 +1,257 @@
...
+    end
+
+
+    describe "Create wallets" do
+      it "I can create, get and delete wallet from mnemonics getting acc_xpub from cardano-address" do
+        w = SHARED.wallets
+
+        m24 = mnemonic_sentence(24)
+        acc_ix = '0H'
+        acc_xpub = cardano_address_get_acc_xpub(m24, "1852H/1815H/#{acc_ix}")
```
paweljakubas, line 16: it is going to be 1854H very soon - just indicating

paweljakubas (approved):
sweat

test/e2e/spec/shared_spec.rb
piotr-iohk: Yep. :+1:

PR #2620, opened 2021-04-23
Add `constraints` field to `TransactionLayer`

jonathanknowles:
# Issue Number

ADP-839

# Overview

This PR:

- [x] adds field `constraints` of type `TxConstraints` to the `TransactionLayer` record type.
- [x] adds an implementation of `constraints` to `Shelley.Transaction.TransactionLayer`.
- [x] adds property tests to demonstrate the correctness of `Shelley.Transaction.TransactionLayer.constraints`.

# Refactoring

This PR also:

- [x] makes `TxSize` into a concrete type exported by `Primitive.Types.Tx`.
- [x] removes `calcMinimumCoinValue` from `TransactionLayer`, as `TxConstraints.txOutputMinimumAdaQuantity` makes this redundant.

# Future work

This PR will make it possible to:

- remove `tokenBundleSizeAssessor` from `TransactionLayer`, as `TxConstraints.txOutputMaximumSize` makes this redundant.

### DIFF
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
@@ -40,6 +40,7 @@ module Cardano.Wallet.Shelley.Transaction
     , mkTx
     , mkTxSkeleton
     , mkUnsignedTx
+    , txConstraints
     ) where
 
 import Prelude
@@ -85,22 +86,26 @@ import Cardano.Wallet.Primitive.Types.Address
     ( Address (..) )
 import Cardano.Wallet.Primitive.Types.Coin
     ( Coin (..), addCoin, subtractCoin )
+import Cardano.Wallet.Primitive.Types.TokenBundle
+    ( TokenBundle )
 import Cardano.Wallet.Primitive.Types.TokenMap
     ( AssetId (..), TokenMap )
 import Cardano.Wallet.Primitive.Types.TokenPolicy
     ( TokenName (..) )
 import Cardano.Wallet.Primitive.Types.TokenQuantity
-    ( TokenQuantity )
+    ( TokenQuantity (..) )
 import Cardano.Wallet.Primitive.Types.Tx
     ( SealedTx (..)
     , TokenBundleSizeAssessment (..)
     , TokenBundleSizeAssessor (..)
     , Tx (..)
+    , TxConstraints (..)
     , TxMetadata (..)
     , TxOut (..)
     , TxSize (..)
     , txOutCoin
     , txOutMaxTokenQuantity
+    , txSizeDistance
     )
 import Cardano.Wallet.Shelley.Compatibility
     ( AllegraEra
@@ -108,6 +113,7 @@ import Cardano.Wallet.Shelley.Compatibility
     , fromAllegraTx
     , fromMaryTx
     , fromShelleyTx
+    , maxTokenBundleSerializedLengthBytes
     , sealShelleyTx
     , toAllegraTxOut
     , toCardanoLovelace
@@ -139,6 +145,8 @@ import Control.Monad
     ( forM )
 import Data.ByteString
     ( ByteString )
+import Data.Function
+    ( (&) )
 import Data.Generics.Internal.VL.Lens
     ( view )
 import Data.Generics.Labels
@@ -152,7 +160,7 @@ import Data.Set
 import Data.Type.Equality
     ( type (==) )
 import Data.Word
-    ( Word16 )
+    ( Word16, Word64, Word8 )
 import Fmt
     ( Buildable, pretty )
 import GHC.Stack
@@ -169,6 +177,7 @@ import qualified Cardano.Crypto.DSIGN as DSIGN
 import qualified Cardano.Crypto.Hash.Class as Crypto
 import qualified Cardano.Crypto.Wallet as Crypto.HD
 import qualified Cardano.Ledger.Core as SL
+import qualified Cardano.Wallet.Primitive.Types.Coin as Coin
 import qualified Cardano.Wallet.Primitive.Types.TokenBundle as TokenBundle
 import qualified Cardano.Wallet.Primitive.Types.TokenMap as TokenMap
 import qualified Cardano.Wallet.Primitive.Types.UTxOIndex as UTxOIndex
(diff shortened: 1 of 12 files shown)


lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -799,9 +797,8 @@ estimateTxSize witnessTag ctx (SelectionSkeleton inputCount outs chgs) =
     sizeOf_Output TxOut {address, tokens}
         = sizeOf_SmallArray
         + sizeOf_Address address
-        + sizeOf_SmallArray
-        + sizeOf_Coin (TokenBundle.getCoin tokens)
-        + F.foldl' (\t -> (t +) . sizeOf_NativeAsset) 0 (TokenBundle.getAssets tokens)
```
Anviking, line 702: Sounds good, but does it not risk being slower?

jonathanknowles: > Sounds good, but does it not risk being slower?

I eventually decided to get rid of this. I'm pretty sure that our estimate is quite a bit higher than the real serialized length (for Shelley), but I'd be happier changing this function when we have a test to check that `estimateTxSize` gives answers that are greater than or equal to the results of serializing a transaction with `cardano-node` functions.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -540,45 +550,82 @@ _decodeSignedTx era bytes = do
         _ ->
             Left ErrDecodeSignedTxNotSupported
 
+data TxSkeleton = TxSkeleton
```
Anviking, line 651: Might as well put the
> his record type includes the minimal set of fields that are absolutely
necessary for the `estimateTxSize` function to perform its calculations,
and nothing else.
> 
> As such, the fields of `TxSkeleton` contain a subset of the data included
in the union of `SelectionSkeleton` and `TransactionCtx`.

From the commit message as a doc comment here.

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -704,3 +742,201 @@ dummyProtocolParameters = ProtocolParameters
...
+    deriving (Eq, Show)
+
+instance Arbitrary (Large TokenBundle) where
+    arbitrary = fmap Large . genFixedSizeTokenBundle =<< choose (1, 128)
+
+-- Tests that if a bundle is oversized (when serialized), then a comparison
+-- between 'txOutputSize' and 'txOutputMaximumSize' should also indicate that
+-- the bundle is oversized.
+--
+prop_txConstraints_txOutputMaximumSize :: Large TokenBundle -> Property
```
Anviking: 👍

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -704,3 +742,201 @@ dummyProtocolParameters = ProtocolParameters
...
+prop_txConstraints_txBaseSize :: Property
+prop_txConstraints_txBaseSize =
+    txBaseSize mockTxConstraints
+        === estimateTxSize emptyTxSkeleton
+
+-- Tests that using 'txConstraints' to estimate the cost of a non-empty
+-- selection produces a result that is consistent with the result of using
+-- 'estimateTxCost'.
+--
+prop_txConstraints_txCost :: MockSelection -> Property
```
Anviking, line 833: 👍

Anviking (approved):
lgtm!

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
jonathanknowles: Ah good point! I'll add some comments in there. 👍🏻

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -546,45 +551,180 @@ _decodeSignedTx era bytes = do
...
+    , txDelegationAction = view #txDelegationAction context
+    , txRewardWithdrawal = withdrawalToCoin $ view #txWithdrawal context
+    , txWitnessTag = witness
+    , txInputCount = view #skeletonInputCount skeleton
+    , txOutputs = view #skeletonOutputs skeleton
+    , txChange = view #skeletonChange skeleton
+    }
+
+estimateTxCost :: ProtocolParameters -> TxSkeleton -> Coin
+estimateTxCost pp args =
```
jonathanknowles:
```suggestion
estimateTxCost pp skeleton =
```

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -704,3 +742,201 @@ dummyProtocolParameters = ProtocolParameters
...
+
+instance Arbitrary (Large TokenBundle) where
+    arbitrary = fmap Large . genFixedSizeTokenBundle =<< choose (1, 128)
+
+-- Tests that if a bundle is oversized (when serialized), then a comparison
+-- between 'txOutputSize' and 'txOutputMaximumSize' should also indicate that
+-- the bundle is oversized.
+--
+prop_txConstraints_txOutputMaximumSize :: Large TokenBundle -> Property
+prop_txConstraints_txOutputMaximumSize (Large bundle) =
```
jonathanknowles: The `bundle` argument probably be blinded to avoid spamming the logs with the output of `show` for huge token bundles. (We already have the data we need in the counterexamples.)

jonathanknowles: Fixed in 2099f94ebad72a8bdf1bd04b7e42120694075d32.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
jonathanknowles: Fixed in 59e10c52431f22d2998e5967b7a0595e7110f771.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
jonathanknowles: Fixed in d7ff662ccfd5c79531b047cbd37e7bc9f52080d0.

rvl (approved):

rvl:
bors r-

Let's do the release first, then try to merge this.

PR #2621, opened 2021-04-23
Replace force with strict fields and nothunk test

Anviking:
# Issue Number

ADP-630

# Overview

- [x] remove `force`
- [x] Use strict record fields in `LogState` and `FollowStats`
- [x] Add a nothunks test for sanity 

# Comments

`force` didn't help. It might have made things worse.

Testing in ghci with `:set +s` and `iterate (updateStats msg1) s0 !! 100000` we get:
- With force -> (0.42 secs, 297,818,512 bytes)
- No force -> (0.16 secs, 86,713,928 bytes)
- No force but with strict record fields -> (0.09 secs, 74,874,224 bytes)


paweljakubas (approved):
persuasive change!

PR #2624, opened 2021-04-26
Address derivation for shared wallets

paweljakubas:
# Issue Number

ADP-866

# Overview

- [x] Adding Shared style
- [x] Accommodate new cardano-addresses, align code, change MutableAccount -> Stake
- [x] Deal with prefixes of shared wallet vs shelley wallet by adding SharedWalletKeys
- [x] Align all unit tests
- [x] Align all integration tests plus add test checking newly added validation in cardano-addresses
- [x] Add integration testing 
- [x] Enable receiving hashed verification keys, now only verification keys are present which needs to be hashed to be used in script
- [x] Sneak SharedKey in shared wallet rather than ShelleyKey  
- [x] Removed MultisigScript from Role and refactored ParentContext

# Comments

[Rendered API docs](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/input-output-hk/cardano-wallet/paweljakubas/adp-866/add-shared-style/specifications/api/swagger.yaml)


lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -184,7 +185,7 @@ data Depth
 data Role
     = UtxoExternal
     | UtxoInternal
-    | MutableAccount
+    | Stake
```
KtorZ: This will cause a breaking change in the API which generically derive path parameters from these constructors. I agree with the change here, but the JSON instance of the API should be modified to also still accept `mutable_account` if provided to not break existing integrations.

lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs
```diff
@@ -497,13 +498,23 @@ instance PersistFieldSql ScriptHash where
...
-        . T.encodeUtf8
+    fromText txt = do
+        (keyRole, txt') <- case T.uncons txt of
+            Just (firstChar, rest) ->case firstChar of
+                '0' -> pure (CA.Payment, rest)
+                '2' -> pure (CA.Delegation, rest)
+                _ -> Left $ TextDecodingError "KeyHash should begin with either '0' or '1'."
+            Nothing -> Left $ TextDecodingError "KeyHash should not be empty."
+        bimap textDecodingError (KeyHash keyRole)
+            $ convertFromBase Base16 $ T.encodeUtf8 txt'
```
KtorZ: This is a red-herring to me :thinking: ... 

Why would you need to differentiate them in the database? Surely, the context when deserializing from the database is enough to assign the right role to the key hash. If it's not, then it should likely go in a separate column, instead of being serialized all at once.

Regardless, this will require a database migration and I fail to see how existing databases can be migrated :thinking: ? (or said differently, why is this distinction only occurs now?)

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
paweljakubas: after some thinking I reverted this change

lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs
paweljakubas: I checked that KeyHash is not stored. These instances are not needed in the end. so removed them. They are some remnants of past version of shared wallet I forgot to erase.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Shared.hs
```diff
@@ -0,0 +1,178 @@
...
+        hash (unXPub pub)
+
+    getRawKey =
+        getKey
+
+    liftRawKey =
+        SharedKey
+
+    keyTypeDescriptor _ =
+        "shared"
```
paweljakubas: I have doubts what should be here to be honest

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2139,13 +2140,38 @@ derivePublicKey ctx wid role_ ix = db & \DBLayer{..} -> do
...
+        $ withExceptT ErrDerivePublicKeyNoSuchWallet
+        $ withNoSuchWallet wid
+        $ readCheckpoint wid
+
+    let acctK = accountPublicKey $ getState cp
+    let addrK = deriveAddressPublicKey acctK role_ addrIx
+
+    return addrK
+  where
+    db = ctx ^. dbLayer @IO @s @k
```
KtorZ: This code looks _very_ similar to `derivePublicKeyShelley`. Both function could be unified rather easily by providing the account key lookup as a function:

```hs
derivePublicKey
    :: forall ctx s k n.
        ( HasDBLayer IO s k ctx
        , SoftDerivation k
        )
    => ctx
    -> WalletId
    -> (s -> k 'AccountK XPub)
    -> Role
    -> DerivationIndex
    -> ExceptT ErrDerivePublicKey IO (k 'AddressK XPub)
derivePublicKey ctx wid getAccountKey role_ ix = ...
```

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -880,6 +886,31 @@ type DeleteSharedWallet = "shared-wallets"
...
+    :> Capture "index" (ApiT DerivationIndex)
+    :> Get '[JSON] ApiVerificationKeyShared
+
+-- | https://input-output-hk.github.io/cardano-wallet/api/#operation/postAccountKeyShared
+type PostAccountKeyShared = "shared-wallets"
+    :> Capture "walletId" (ApiT WalletId)
+    :> "keys"
+    :> Capture "index" (ApiT DerivationIndex)
+    :> ReqBody '[JSON] ApiPostAccountKeyData
+    :> PostAccepted '[JSON] ApiAccountKeyShared
```
KtorZ, line 918: I don't think these are necessary; Surely the structure and outer interface is the same as for standard 1852' keys, right?

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Shared.hs
KtorZ: This is the name then used to prefix database's names. In other wallet scheme, we typically use a 3-letter identifier:

- she
- ica
- rnd

I'd say that for consistency, we probably want to also have a 3-letter name here but I am perhaps being too maniac :grimacing:

paweljakubas: went for `sha`

lib/core/src/Cardano/Wallet.hs

lib/core/src/Cardano/Wallet/Api.hs
paweljakubas: IMHO it is small code bloating that serves well in improving code clarity. We have
```
1827     ApiVerificationKeyShelley: &ApiVerificationKeyShelley                                                                                                                                                     
1828       type: string                                                                                                                                                                                            
1829       format: bech32                                                                                                                                                                                          
1830       pattern: "^((addr_vk)|(stake_vk))1[0-9a-z]*$"                                                                                                                                                           
1831                                                                                                                                                                                                               
1832     ApiVerificationKeyShared: &ApiVerificationKeyShared                                                                                                                                                       
1833       type: string                                                                                                                                                                                            
1834       format: bech32                                                                                                                                                                                          
1835       pattern: "^((addr_shared_vk)|(stake_shared_vk))1[0-9a-z]*$"
```
And the same on Api.Types level, here ToJSON instances capturing those prefixes, and also enpoints consuming either `ApiVerificationKeyShelley` or `ApiVerificationKeyShared`. I think it is dead simply and ... that's good.

paweljakubas:
@KtorZ  @rvl  I added PaymentAddress instances to make it pass - at this level (we are not discovering, sending txs) it is not needed. But this is to be changed.. This bits should be used https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Script.hs#L189 I need to think how to approach it. Would prefer to do it in the next PR addressing discovery ..

specifications/api/swagger.yaml
```diff
@@ -5199,3 +5231,40 @@ paths:
           application/json:
             schema: *ApiSharedWalletPatchData
       responses: *responsesPatchSharedWallet
+
+  /shared-wallets/{walletId}/keys/{role}/{index}?hash=false:
```
paweljakubas: I am not sure if query params should be exposed like this.Nevertheless, 
```
openapi-spec-validator --schema 3.0.0 specifications/api/swagger.yaml
```
passes

lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -1433,6 +1438,42 @@ getSharedWallet ctx (ApiSharedWallet (Right w)) = do
...
+    -> Role
+    -> DerivationIndex
+    -> Maybe Bool
+    -> m (HTTP.Status, Either RequestException ApiVerificationKeyShared)
+getSharedWalletKey ctx (ApiSharedWallet (Left w)) role ix hashed = do
+    let link = Link.getSharedWalletKey w role ix hashed
+    request @ApiVerificationKeyShared ctx link Default Empty
+getSharedWalletKey ctx (ApiSharedWallet (Right w)) role ix hashed = do
+    let link = Link.getSharedWalletKey w role ix hashed
+    request @ApiVerificationKeyShared ctx link Default Empty
```
KtorZ: Minor note: this implicitly suggests to me that both implementations are different, whereas they aren't. I would favour a prismatic accessor to extract the wallet from the `ApiSharedWallet` instead, or simply pattern match on that field only inside the function body, e.g.

```hs
let w = case apiSharedWalletOf ...
let link = ...
request @...
```

lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -1433,6 +1438,42 @@ getSharedWallet ctx (ApiSharedWallet (Right w)) = do
...
+    -> DerivationIndex
+    -> Headers
+    -> Payload
+    -> m (HTTP.Status, Either RequestException ApiAccountKeyShared)
+postAccountKeyShared ctx (ApiSharedWallet (Left w)) ix headers payload = do
+    let link = Link.postAccountKeyShared w ix
+    request @ApiAccountKeyShared ctx link headers payload
+postAccountKeyShared ctx (ApiSharedWallet (Right w)) ix headers payload = do
+    let link = Link.postAccountKeyShared w ix
+    request @ApiAccountKeyShared ctx link headers payload
```
KtorZ: Same here :point_up:

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
```diff
@@ -891,6 +887,24 @@ spec = describe "SHELLEY_ADDRESSES" $ do
...
         expectErrorMessage msg r
 
+    it "ANY_ADDRESS_POST_17d - script with mixed payment/delegation verification keys is invalid" $ \ctx -> do
+        let payload = Json [json|{
+                "payment": {
+                    "some": {
+                        "from" : [
+                            "addr_shared_vkh1zxt0uvrza94h3hv4jpv0ttddgnwkvdgeyq8jf9w30mcs6y8w3nq",
+                            "stake_shared_vkh1nac0awgfa4zjsh4elnjmsscz0huhss8q2g0x3n7m539mwaa5m7s"
+                            ],
```
KtorZ, line 899: :+1:

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
```diff
@@ -891,6 +887,24 @@ spec = describe "SHELLEY_ADDRESSES" $ do
...
+                         "at_least": 1
+                         }
+                    },
+                "validation": "required"
+            }|]
+        r <- request @AnyAddress ctx Link.postAnyAddress Default payload
+        expectResponseCode HTTP.status400 r
+        let msg = "All keys of a script must have the same role: either payment or delegation."
```
KtorZ: This should not be hard-coded here but rather defined in the `Framework/TestData` module for lack of a better mechanism.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/SharedWallets.hs
```diff
@@ -119,8 +127,8 @@ spec = describe "SHARED_WALLETS" $ do
                           }
                     }
                 } |]
```
KtorZ, line 127: Note that, I think this doesn't generally reflect a _real_ use-case. Users don't generally know how to get their account public key from their recovery phrase. A typical scenario when creating from a mnemonic would be to:

1. Create a shared wallet with an empty mapping of cosigners 
2. Query the API to get the account key associated with the wallet 
3. Complete the script template by providing the mapping for my own key

Likely, the step 2 & 3 would automatically be done by Daedalus on the behalf of the end-user; still I'd love to see our scenarios reflecting this and not resort on custom `accPubKeyFromMnemonics` ;) ... In the end, the goal of the API scenarios is to described capabilities doable only via the API. 
 
---

I imagine that this may be conflicting with the invariant we are trying to enforce, right? So it may be necessary to either:

- Provide a way to easily derive keys from a given mnemonic phrase via the API (less preferred approach)
- Allow maybe for assigning a special string `"self"` to cosigners, such that one could submit:
    ```json
    { "cosigner#0": "self" },
    ```
   And let the server replace it with the correct account key internally.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2123,18 +2124,55 @@ derivePublicKey
...
+        ( HasDBLayer IO s k ctx
+        , SoftDerivation k
+        )
+    => ctx
+    -> (s -> k 'AccountK XPub)
+    -> WalletId
+    -> Role
+    -> DerivationIndex
+    -> ExceptT ErrDerivePublicKey IO (k 'AddressK XPub)
+derivePublicKey ctx toAccXPub wid role_ ix = db & \DBLayer{..} -> do
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2123,18 +2124,55 @@ derivePublicKey
...
     addrIx <- withExceptT ErrDerivePublicKeyInvalidIndex $ guardSoftIndex ix
 
     cp <- mapExceptT atomically
         $ withExceptT ErrDerivePublicKeyNoSuchWallet
         $ withNoSuchWallet wid
         $ readCheckpoint wid
 
-    -- NOTE: Alternatively, we could use 'internalPool', they share the same
-    --       account public key.
```
KtorZ, line 2125: That comment got lost along the way! It would be nice to preserve it in `toAccXPubShelley` :+1:

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1267,6 +1277,35 @@ withLegacyLayer' (ApiT wid)
...
+    tryShared liftE deadShared
+  where
+    tryShared onMissing onNotResponding = withWorkerCtx @_
+        @(SharedState n SharedKey)
+        @SharedKey
+        shared
+        wid
+        onMissing
+        onNotResponding
+        (const withShared)
```
KtorZ: This will not be needed if the API uses a different set of endpoints for shared wallets and standard shelley wallets. 
It seems like the API now defines a completely different section for `shared-wallets`, in which case, there can only be one choice of API layer for the corresponding endpoint.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1043,11 +1046,21 @@ data ApiWalletSignData = ApiWalletSignData
     } deriving (Eq, Generic, Show)
       deriving anyclass NFData
 
-newtype ApiVerificationKey = ApiVerificationKey
+newtype ApiVerificationKeyShelley = ApiVerificationKeyShelley
     { getApiVerificationKey :: (ByteString, Role)
     } deriving (Eq, Generic, Show)
       deriving anyclass NFData
 
+data VerificationKeyHashing = HashingApplied | WithoutHashing
```
KtorZ: `WithHashing | WithoutHashing` perhaps ?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1043,11 +1046,21 @@ data ApiWalletSignData = ApiWalletSignData
...
+    deriving (Eq, Generic, Show)
+    deriving anyclass NFData
+
+data ApiVerificationKeyShared = ApiVerificationKeyShared
+    { getApiVerificationKey :: (ByteString, Role)
+    , hashed :: VerificationKeyHashing
```
KtorZ, line 1073: I am not quite sure about what this field is for :thinking: ?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1436,50 +1455,97 @@ instance ToJSON (ApiT DerivationIndex) where
...
+instance ToJSON ApiVerificationKeyShelley where
+    toJSON (ApiVerificationKeyShelley (pub, role_)) =
         toJSON $ Bech32.encodeLenient hrp $ dataPartFromBytes pub
       where
         hrp = case role_ of
             UtxoInternal -> [humanReadablePart|addr_vk|]
             UtxoExternal -> [humanReadablePart|addr_vk|]
             MutableAccount -> [humanReadablePart|stake_vk|]
-            MultisigScript -> [humanReadablePart|script_vk|]
+            MultisigScript -> error "only role=0,1,2 is supported for ApiVerificationKeyShelley"
```
KtorZ: Shouldn't `MultisigScript` actually be removed altogether from the corresponding data-type?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1436,50 +1455,97 @@ instance ToJSON (ApiT DerivationIndex) where
...
+        parseRoleHashing = \case
+            hrp | hrp == [humanReadablePart|addr_shared_vk|] -> pure (UtxoExternal, WithoutHashing)
+            hrp | hrp == [humanReadablePart|stake_shared_vk|] -> pure (MutableAccount, WithoutHashing)
+            hrp | hrp == [humanReadablePart|addr_shared_vkh|] -> pure (UtxoExternal, HashingApplied)
+            hrp | hrp == [humanReadablePart|stake_shared_vkh|] -> pure (MutableAccount, HashingApplied)
+            _ -> fail errRole
+          where
+            errRole =
+                "Unrecognized human-readable part. Expected one of:\
+                \ \"addr_shared_vk\" or \"stake_shared_vk\"."
```
KtorZ: or `_vkh` ?

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -499,17 +500,24 @@ deriveRewardAccount pwd rootPrv =
...
     -> KeyHash
-hashVerificationKey =
-    KeyHash . blake2b224 . xpubPublicKey . getRawKey
+hashVerificationKey role' =
+    KeyHash keyRole . blake2b224 . xpubPublicKey . getRawKey
+  where
+    keyRole = case role' of
+        UtxoExternal -> CA.Payment
+        MutableAccount -> CA.Delegation
+        _ -> error "verification keys make sense only for payment (role=0) and delegation (role=2)"
```
KtorZ: And `role=1`; The type signature is not restricted to `purpose=1854` so `role=1` should be possible here, or the type signature should be further restricted.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Shared.hs
```diff
@@ -0,0 +1,248 @@
...
+
+instance MkKeyFingerprint SharedKey Address where
+    paymentKeyFingerprint (Address bytes) =
+        Right $ KeyFingerprint $ BS.take hashSize $ BS.drop 1 bytes
+
+instance MkKeyFingerprint SharedKey (Proxy (n :: NetworkDiscriminant), SharedKey 'AddressK XPub) where
+    paymentKeyFingerprint (_, paymentK) =
+        Right $ KeyFingerprint $ blake2b224 $ xpubPublicKey $ getKey paymentK
+
+instance PaymentAddress 'Mainnet SharedKey where
```
KtorZ: I don't think this instance make sense for `SharedKey`. An address (payment or delegation) can in principle only be constructed from the state using the script template!

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1043,11 +1046,21 @@ data ApiWalletSignData = ApiWalletSignData
...
+
+data ApiVerificationKeyShared = ApiVerificationKeyShared
+    { getApiVerificationKey :: (ByteString, Role)
+    , hashed :: VerificationKeyHashing
+    } deriving (Eq, Generic, Show)
+      deriving anyclass NFData
+
 data ApiPostAccountKeyData = ApiPostAccountKeyData
     { passphrase :: ApiT (Passphrase "raw")
     , extended :: Bool
```
KtorZ: Only seeing this now, and I think we can do better than `Bool` here.

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
```diff
@@ -489,11 +498,28 @@ server byron icarus shelley multisig spl ntp =
 
     sharedWallets :: Server SharedWallets
     sharedWallets =
-             postSharedWallet multisig generateKeyFromSeed ShelleyKey
-        :<|> (fmap fst . getWallet multisig mkSharedWallet)
-        :<|> patchSharedWallet multisig ShelleyKey Payment
-        :<|> patchSharedWallet multisig ShelleyKey Delegation
-        :<|> deleteWallet multisig
+             postSharedWallet @_ @_ @SharedKey multisig Shared.generateKeyFromSeed SharedKey
+        :<|> (\wid -> withMultisigLayer wid
```
KtorZ: Cf earlier comment, I don't think this indirection is necessary.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -247,6 +249,9 @@ instance TxWitnessTagFor IcarusKey where
 instance TxWitnessTagFor ByronKey where
     txWitnessTagFor = TxWitnessByronUTxO Byron
 
+instance TxWitnessTagFor SharedKey where
```
KtorZ: This instance looks *very* suspicious to me. This should not be needed for Shared wallets since they are signed independently and using a different process. Plus, `TxWitnessTagFor` is used for calculating the size of the resulting transaction; in the case of shared scripts this also needs to acknowledge for the script in the witness which require more changes to the code down the line.

specifications/api/swagger.yaml
```diff
@@ -5283,3 +5315,40 @@ paths:
...
+    get:
+      operationId: getSharedWalletKey
+      tags: ["Keys"]
```
piotr-iohk: I think those two new endpoints should go to separate category "Shared Keys" grouped within Shared wallets as currently it renders:

Screenshot from 2021-05-05 11-25-37

piotr-iohk:
**Question:**
For shared wallet created from acc pub key one cannot get acc pub key:
```
curl -X POST http://localhost:8090/v2/shared-wallets/f70d591d06a83517e7e46a69a3da58cb52baf60e/keys/0H \
-d '{"passphrase":"Secure Passphrase","extended":true}' \

{"code":"no_root_key","message":"I couldn't find a root private key for the given wallet: f70d591d06a83517e7e46a69a3da58cb52baf60e. However, this operation requires that I do have such a key. Either there's no such wallet, or I don't fully own it."}
```
The same behavior is actually for shelley wallet from acc pub key and it's corresponding [endpoint](https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/postAccountKey), however I'm wondering if it is in fact desired behavior? :thinking:

specifications/api/swagger.yaml
piotr-iohk: fixed in c68a6bc.

lib/core-integration/src/Test/Integration/Framework/DSL.hs
paweljakubas: done

lib/core-integration/src/Test/Integration/Framework/DSL.hs
paweljakubas: done, also for other Shared-related helpers

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Addresses.hs
paweljakubas: done, also for other errors connected with scripts

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/SharedWallets.hs
paweljakubas: yes, it is conflicting with invariant we introduced in the previous shared wallet PR. I will try with `self` though

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
paweljakubas: removed newTransactionLayer from apiLayer for the time being. Will return to it when txs are to be supported for shared wallets

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Shared.hs
paweljakubas: there was unnecessary `Payment n SharedKey`constraint in Sqlite's PersistentState. Now it is removed

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/SharedWallets.hs
paweljakubas: done, added illustrative integration tests, aligned swagger and core unit tests

specifications/api/swagger.yaml
rvl: I fixed it for you.

rvl:
> For shared/shelley wallet created from acc pub key one cannot get acc pub key.

That seems like a bug to me @piotr-iohk, although the impact is low given that the user would presumably already know their acc pub key.

lib/core/src/Cardano/Wallet.hs
paweljakubas: done

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: added

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: introduced ```data KeyFormat = Extended | NonExtended ``` as a replacement

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: indeed, fixed

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
paweljakubas: done

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: removed from Role and refactored ParentContext

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
paweljakubas: corrected

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/SharedWallets.hs
```diff
@@ -237,6 +276,118 @@ spec = describe "SHARED_WALLETS" $ do
...
+                    }
+                } |]
+        rPostWithSelf <- postSharedWallet ctx Default payloadPostWithSelf
+        verify rPostWithSelf
+            [ expectResponseCode HTTP.status201
+            ]
+
+        let walWithSelf = getFromResponse id rPostWithSelf
+
+        getWalletIdFromSharedWallet walWithSelf `Expectations.shouldBe` getWalletIdFromSharedWallet wal
```
KtorZ, line 338: :+1:

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2113,18 +2116,57 @@ derivePublicKey
...
+toAccXPubShared = accountPublicKey
+
+toAccXPubShelley
+    :: SeqState n k
+    -> k 'AccountK XPub
+toAccXPubShelley s =
+    -- NOTE: Alternatively, we could use 'internalPool', they share the same
+    --       account public key.
+    let (Seq.ParentContextUtxo acctK) = Seq.context $ Seq.externalPool s
+    in acctK
```
KtorZ: The two functions above sound like implementation of a potential type class `HasAccount`, which would allow writing `derivePublicKey` with a single implementation by leveraging this capability.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -668,6 +680,7 @@ postWallet
         , IsOurs s RewardAccount
         , Typeable s
         , Typeable n
+        , (k == SharedKey) ~ 'False
```
KtorZ, line 679: Is this constraint needed :thinking: ?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1444,98 +1483,168 @@ instance ToJSON (ApiT DerivationIndex) where
...
+parsePub :: MonadFail f => ByteString -> KeyFormat -> f ByteString
+parsePub bytes extd
+    | BS.length bytes == bytesExpectedLength =
+          pure bytes
+    | otherwise =
+          fail $ parsePubErr extd
+  where
+    bytesExpectedLength = case extd of
+        Extended -> 64
+        NonExtended -> 32
```
KtorZ, line 1615: I see that these values are more and more hard-coded everywhere and this makes me _uncomfortable_. Plus, there seems to be about 12x different ways of parsing a key, with different format and whatnot. Would it be possible to try unifying those and find a better, more consistent way of handling all of that? For example, the approach taken in cardano-addresses provides a thin but powerful module api  for parsing and managing keys we should strive to re-use this, even if it is by re-wrapping it into some module of cardano-wallet.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1444,98 +1483,168 @@ instance ToJSON (ApiT DerivationIndex) where
...
-                pure bytes
-            | otherwise =
-                fail $ parsePubErr extd
+instance FromJSON KeyFormat where
+    parseJSON = withBool "KeyFormat" $ \case
+        True -> pure Extended
+        False -> pure NonExtended
+instance ToJSON KeyFormat where
+    toJSON Extended = toJSON True
+    toJSON NonExtended = toJSON False
```
KtorZ: Avoid Boolean like the plague in the API and prefer named constant over it. Especially since we already have a sum-type for it! The JSON instance could / should be derived generically from the sum-type.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -2331,6 +2441,50 @@ instance {-# OVERLAPS #-} EncodeStakeAddress n
...
+
+instance FromJSON XPubOrSelf where
+    parseJSON t = parseXPub t <|> parseSelf t
+      where
+        parseXPub = withText "XPub" $ \txt ->
+            case fromBase16 (T.encodeUtf8 txt) of
+                Left err -> fail err
+                Right hex' -> case xpubFromBytes hex' of
+                    Nothing -> fail "Extended public key cannot be retrieved from a given hex bytestring"
+                    Just validXPub -> pure $ OtherCosigner validXPub
```
KtorZ: Small note: the type now suggests that this `OtherCosigner` is not us whereas in practice, it can be. So I'd suggest renaming this to something more generic like `SomeAccountKey ...` to not be misleading.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/SharedKey.hs
```diff
@@ -0,0 +1,57 @@
...
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-- |
+-- Copyright: © 2018-2021 IOHK
+-- License: Apache-2.0
+--
+-- Definition of 'Shared' Keys.
+
+module Cardano.Wallet.Primitive.AddressDerivation.SharedKey
```
KtorZ, line 21: Note that there currently is a module: `Cardano.Wallet.Primitive.AddressDiscovery.Script` with a lot of things which should probably be in this module here? What do you think? Seems to me that the `...AddressDiscovery.Script` module is an artifact from the previous work which only adds confusion now.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
```diff
@@ -244,37 +246,36 @@ defaultAddressPoolGap =
...
+        => key 'AccountK XPub
+        -> ParentContext chain key
+
+    ParentContextShared
+        :: (key ~ SharedKey)
+        => key 'AccountK XPub
         -> ScriptTemplate
         -> Maybe ScriptTemplate
-        -> ParentContext 'MultisigScript key
+        -> ParentContext chain key
```
KtorZ: Suggestion: no need for type parameter here, can be written as:

```hs
    ParentContextShared
        :: SharedKey 'AccountK XPub
        -> ScriptTemplate
        -> Maybe ScriptTemplate
        -> ParentContext chain SharedKey
```

Also, the `chain` parameter seems redundant now?

lib/core/src/Cardano/Wallet/Api/Types.hs

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: unfortunately yes, otherwise it does not compile

lib/core/src/Cardano/Wallet.hs
paweljakubas: introduced`GetAccount` as we have `GetPurpose` already. That was good suggestion, I managed to marry a lot of code ie., simplify and cut a lot of code :blush:

lib/core/src/Cardano/Wallet/Api/Types.hs

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
paweljakubas: changed to `ParentContext 'UtxoExternal key`

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/SharedKey.hs
paweljakubas: yes, it is very reasonable to do it. I planned to do it in the next PR. Nevertheless, did it and sprinkled bits between SharedState and SharedKey to avoid cyclic dependency.

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: I agree. If you don't mind I would like to do it in the next PR. I would like to think more about it. :pray:

lib/core/src/Cardano/Wallet/Api/Server.hs
KtorZ: Interesting. I thought GHC would figure it out :shrug:

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery.hs
```diff
@@ -151,3 +153,11 @@ class KnownAddresses s where
 -- Hardened derivation is used at this level.
 coinTypeAda :: Index 'Hardened 'CoinTypeK
 coinTypeAda = toEnum 0x80000717
+
+-- It is used for geting purpose for a given key.
+class GetPurpose (key :: Depth -> Type -> Type)  where
+    getPurpose :: Index 'Hardened 'PurposeK
+
+-- It is used for geting account public key for a given state.
+class GetAccount s (key :: Depth -> Type -> Type) | s -> key  where
```
KtorZ, line 162: Why the functional dependency here :thinking: ?

lib/shelley/src/Cardano/Wallet/Shelley.hs
```diff
@@ -288,7 +290,7 @@ serveWallet
                     Server.idleWorker
                 shelleyApi <- apiLayer (newTransactionLayer net) nl
                     (Server.manageRewardBalance proxy)
-                multisigApi <- apiLayer (newTransactionLayer net) nl
+                multisigApi <- apiLayer undefined nl
```
KtorZ: Whoops? I understand you want to do this in a second PR and there's currently nothing wired into the transaction layer but may we use a more descriptive error than `undefined` here :grimacing: ?

specifications/api/swagger.yaml
```diff
@@ -4109,6 +4186,7 @@ x-tagGroups:
   - name: Shelley (Shared Wallets)
     tags:
     - Shared Wallets
+    - Shared Keys
```
KtorZ: Small note: I'd put the `Shelley (Shared Wallets)` section before `Byron (Random)`. Byron is legacy so it makes more sense to have it presented last :+1:

KtorZ (approved):
Not a small PR :upside_down_face: , it would be good to try to limit the number of changes which goes in a single PR in the future to make iterations less painful. For examples:

- Tackle only one endpoint at a time
- Work on core changes first in isolation (e.g. the SharedKey module) before wiring it up into all layers of the code
- Do database changes separately

Modulo some final minor remarks, I think the PR is _okay_ now. Though quite large also and therefore, it is possible that I have missed something. Before we completely close the shared story and call it done I'd like to have some time to play with the API from a user perspective as a form of acceptance test and I invite @piotr-iohk to do the same. It may be a good opportunity to also write an extensive user guide about that!

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2186,6 +2186,7 @@ updateCosigner
         , Typeable n
         , WalletKey k
         , HasDBLayer IO s k ctx
+        , k ~ SharedKey
```
rvl, line 2189: Just wondering aloud, could the constraints in these functions be made shorter because the types are basically fixed?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -668,6 +679,7 @@ postWallet
...
```
rvl, line 679: I am a bit confused by this.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2166,11 +2217,32 @@ postAccountPublicKey
...
+        , ctx ~ ApiLayer s k
+        , HardDerivation k
+        , WalletKey k
+        )
+    => ctx
+    -> ApiT WalletId
+    -> ApiT DerivationIndex
+    -> ApiPostAccountKeyData
+    -> Handler ApiAccountKeyShared
+postAccountPublicKeyShared ctx (ApiT wid) (ApiT ix) (ApiPostAccountKeyData (ApiT pwd) extd) = do
```
rvl: This function looks almost idential to `postAccountPublicKeyShelley`. Likewise for the JSON instances of `ApiAccountKeyShared`.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2166,11 +2217,32 @@ postAccountPublicKey
...
     withWorkerCtx @_ @s @k ctx wid liftE liftE $ \wrk -> do
```
rvl, line 2218: I suggest combining them and adding a hrp prefix string as a type parameter or something like that.

lib/shelley/src/Cardano/Wallet/Shelley.hs
rvl: I'm guessing there is some tricky type error?

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
```diff
@@ -487,13 +493,22 @@ server byron icarus shelley multisig spl ntp =
...
 
-    sharedWallets :: Server SharedWallets
-    sharedWallets =
-             postSharedWallet multisig generateKeyFromSeed ShelleyKey
-        :<|> (fmap fst . getWallet multisig mkSharedWallet)
-        :<|> patchSharedWallet multisig ShelleyKey Payment
-        :<|> patchSharedWallet multisig ShelleyKey Delegation
-        :<|> deleteWallet multisig
+    sharedWallets
+        :: ApiLayer (SharedState n SharedKey) SharedKey
```
rvl, line 498: The other server functions get their `ctx` from the function scope.
These ones could also, right?

specifications/api/swagger.yaml
```diff
@@ -5244,7 +5322,10 @@ paths:
       description: |
         <p align="right">status: <strong>⚠ under development</strong></p>
 
-        Create a shared wallet from an account public key, its index and script templates.
+        Create a shared wallet from an account public key and script
+        template index.
+
+        **TODO:** ADP-866 is it also restored? If not, when?
```
rvl: Would it be possible to describe what happens when the shared wallet is created? Is it restored immediately, or must further API requests be made, etc, etc.

specifications/api/swagger.yaml
```diff
@@ -5259,6 +5340,8 @@ paths:
       summary: Get
       description: |
         <p align="right">status: <strong>⚠ under development</strong></p>
+
+        **TODO:** ADP-866 document this endpoint.
```
rvl: I put these TODO messages in for you ;-)

rvl (commented):
It's looking good - almost there!

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery.hs
paweljakubas: we have s ~ SomethingKey n key, and return key ~Account XPub, and I wanted to enforce that key is from s...

lib/shelley/src/Cardano/Wallet/Shelley.hs
paweljakubas: done

specifications/api/swagger.yaml
paweljakubas: done

lib/core/src/Cardano/Wallet.hs
paweljakubas: we can think of introducing constriant groups probably but I am not sure and do not know if they will be more clear. Personally, I am accustomed to explicit constrains. But I was one day looking at ledger repo and they introduced those composite constraints and I think they serve well, but they have situation that one the same composite constraints is in many places...

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: so we use this here https://hackage.haskell.org/package/base-4.15.0.0/docs/src/Data-Type-Equality.html#%3D%3D and we needed as we introduced different ParentContext for SharedKey and the rest (ShelleyKey/IcarusKey..). So we want here to say that k must not be SharedKey. for SharedKey we just specify `k ~ SharedKey`

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: indeed, good point. adopted `postAccountPublicKey` in both Shelley and Shared cases

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: done

lib/shelley/src/Cardano/Wallet/Shelley.hs
paweljakubas: no no...there is required TxWintnessTag instance and I want to leave it for ADP-686. And for the current endpoints tx later is not engaged

lib/shelley/src/Cardano/Wallet/Shelley/Api/Server.hs
paweljakubas: the problem is that it takes ctx type from the above context and we have ApiLayer (SharedStaten ShelleyKey) ShelleyKey ... so I needed to do this trick to make it tick...Nevertheless, I will relook if I can do something in coming PR....

specifications/api/swagger.yaml
paweljakubas: done

specifications/api/swagger.yaml
paweljakubas: done

paweljakubas:
bors retry

paweljakubas:
bors retry

Issue #2630, opened 2021-04-30
Flaky test: `Migration.SelectionSpec.prop_extend`


jonathanknowles:
# Context

Seen in local testing.

Fails around 2% of the time (20 failures seen over 1000 tests).

# Test Case

`Migration.SelectionSpec, Extending selections, prop_extend`

# Failure / Counter-example

```hs
test/unit/Cardano/Wallet/Primitive/Migration/SelectionSpec.hs:133:9:
 1) Cardano.Wallet.Primitive.Migration.Selection.Cardano.Wallet.Primitive.Migration.SelectionSpec, Extending selections, prop_extend
    Falsified (after 3227 tests):
    (*)
    "mockConstraints":
        MockTxConstraints
            { mockTxCostFunction = MockTxCostFunction
                { baseCost = Coin 765
                , sizeCost = Coin 3
                }
            , mockTxBaseSize = 426
            , mockTxInputSize = 4
            , mockTxOutputMaximumSize = 400
            , mockTxOutputMaximumTokenQuantity = 1939
            , mockTxOutputMinimumAdaQuantity = MockTxOutputMinimumAdaQuantity
                { perOutput = Coin 7
                , perOutputAsset = Coin 1
                }
            , mockTxMaximumSize = 2042
            }
    "correctness":
        SelectionCorrect
    "feeExcessExpected":
        Coin 3
    Condition violated:
        "Rebalancing the selection leaves it unchanged"

  To rerun use: --match "/Cardano.Wallet.Primitive.Migration.Selection/Cardano.Wallet.Primitive.Migration.SelectionSpec/Extending selections/prop_extend/"
  Randomized with seed 1170712698
```

Seeds known to provoke failure:
- `779325056`
- `301898547`
- `943930477`
- `880641516`
- `1000314715`
- `830602854`
- `14079222`
- `883834686`
- `171480765`
- `1820043751`
- `1562591007`
- `229965591`
- `1491192159`

# Resolution 

---

# QA

jonathanknowles:
My working hypothesis is that `minimizeFee` is not fulfilling its obligation to be idempotent. I'd like to add a test for idempotency to this function.

I've added a set of seeds that cause this failure to the issue description.

PR #2632, opened 2021-04-30
Fix `prop_{create,extend}` in `SelectionSpec`.

jonathanknowles:
# Issue Number

https://github.com/input-output-hk/cardano-wallet/issues/2630 / ADP-898

# Summary

This PR fixes the following flaky tests:
- `Migration.SelectionSpec.prop_create`
- `Migration.SelectionSpec.prop_extend`

These tests would both fail around 2% of the time, for the same underlying reason.

# Testing

This PR was tested by running the entire `SelectionSpec` **1000** times. No failures were observed.

# Analysis

Both `Selection.create` and `Selection.extend` are required to produce a balanced `Selection` where the fee (and fee excess) is minimized.

Both of these functions rely on `minimizeFee`: this function is required to be **idempotent**, so that:

>  `∀ s. minimizeFee (minimizeFee s) == minimizeFee s`.

To check that `create` and `extend` produce selections with minimal fees, it's sufficient to verify that the selections they return have fees that _cannot be minimized further_. To do this, our pre-existing property tests _already_ call `minimizeFee` and verify that the result does not change:
```hs
    verify (feeExcess selection == feeExcessExpected)
  where
    (feeExcessExpected, _) =
        minimizeFee constraints (feeExcess selection, outputs selection)  
```

However, in addition to this check, we were needlessly asserting that `balance` should be idempotent. The `balance` function is not actually designed for this: it's designed to always be called with outputs that are minimal in their ada quantities.

# Changes

This PR:

- removes the unnecessary and invalid assertion for the idempotency of `balance`.
- makes the pre-condition for `balance` explicit with a documentation comment.
- adds an explicit idempotency test for `minimizeFee`.
- improves error reporting in case of test failures, by displaying the actual fee excess.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Migration/Selection.hs
@@ -230,6 +230,8 @@ extend constraints selection (inputId, inputBundle) =
 -- The ada quantities of the outputs are maximized in order to minimize the fee
 -- excess.
 --
+-- Pre-condition: outputs have minimal ada quantities.
+--
 -- Guarantees the following property for a returned selection 's':
 --
 -- >>> verify s == SelectionCorrect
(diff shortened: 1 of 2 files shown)


paweljakubas (approved):
:+1:

PR #2636, opened 2021-04-30
Model state for tracking stake keys

Anviking:
# Issue Number

ADP-724

# Overview

- [x] Add concepts of
    - `usableKeys` - keys to be used as part of addresses
    - `activeKeys` - keys that are registered and delegating
    - `presentableKeys` - keys that can be displayed to the user (superset of the other two)
- [x] Some simple golden tests for understanding
- [x] Property tests using mock user-actions and chain data
- [x] Add pointer UTxO
    - [x] Add notion of mock ledger to test resistance to rollbacks
    - [x] Implement the pointer UTxO
    - [x] Ensure it works
- [x] Look to test interactions with old wallets which don't know about the pointer, nor other stake keys
    - [x] There's some failure at least
    - [x] Improve counterexample
    - [x] Fix 
- [ ] Look to test concurrent wallet-actions (might be slightly different from txs that could be dropped or re-ordered)
    - Think we're ok without it
- [ ] We need a lookup-map for delegation certificates
    - Shouldn't affect the testing, so not crucial.

# Comments

```
Goldens
  initialDelegationState
    presentableKeys is [0]
    usableKeys is [0]
  registering and delegating key 0
    presentableKeys == [0, 1]
    usableKeys is still [0]
  registering and delegating key 1
    presentableKeys == [0, 1, 2]
    usableKeys == [0, 1]
  Impossible gaps in stake keys (shouldn't happen unless someone manually constructs txs to mess with the on-chain state)
    presentableKeys == [0, 1, 2] (doesn't find 5)
    usableKeys == [0, 1]
(usableKeys s) are consequtive
  +++ OK, passed 100 tests.
(presentableKeys s) are consequtive
  +++ OK, passed 100 tests.
(activeKeys s) are consequtive
  +++ OK, passed 100 tests.
pointer is only created and destroyed in specific cases
  +++ OK, passed 100 tests.
adversaries can't affect usableKeys
  +++ OK, passed 100 tests.
adversaries can't affect pointer ix
  +++ OK, passed 100 tests.
(apply (cmds <> CmdSetPortfolioOf 0) s0) === s0
  +++ OK, passed 100 tests.
no rejected txs, normally
  +++ OK, passed 100 tests.
can recover from dropped transactions
  +++ OK, passed 2000 tests.
```

### DIFF
--- a/lib/core/cardano-wallet-core.cabal
+++ b/lib/core/cardano-wallet-core.cabal
@@ -174,6 +174,7 @@ library
       Cardano.Wallet.Primitive.AddressDiscovery
       Cardano.Wallet.Primitive.Slotting
       Cardano.Wallet.Primitive.AddressDiscovery.Random
+      Cardano.Wallet.Primitive.Delegation.State
       Cardano.Wallet.Primitive.AddressDiscovery.Sequential
       Cardano.Wallet.Primitive.AddressDiscovery.Shared
       Cardano.Wallet.Primitive.SyncProgress
@@ -298,6 +299,7 @@ test-suite unit
     , QuickCheck
     , quickcheck-classes
     , quickcheck-state-machine >= 0.6.0
+    , quiet
     , random
     , retry
     , safe
@@ -363,6 +365,7 @@ test-suite unit
       Cardano.Wallet.Primitive.AddressDiscovery.RandomSpec
       Cardano.Wallet.Primitive.AddressDiscovery.SequentialSpec
       Cardano.Wallet.Primitive.AddressDiscovery.SharedSpec
+      Cardano.Wallet.Primitive.Delegation.StateSpec
       Cardano.Wallet.Primitive.AddressDiscoverySpec
       Cardano.Wallet.Primitive.CoinSelection.MA.RoundRobinSpec
       Cardano.Wallet.Primitive.MigrationSpec
--- a/lib/core/src/Cardano/Wallet/Primitive/Delegation/State.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Delegation/State.hs
@@ -1,0 +1,498 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-- | Module for `DelegationState`.
+module Cardano.Wallet.Primitive.Delegation.State
+    ( -- * Creation
+      DelegationState (..)
+    , initialDelegationState
+
+      -- * Operations
+    , presentableKeys
+    , usableKeys
+    , activeKeys
+
+    -- * For Testing
+    , keyAtIx
+    , lastActiveIx
+    , PointerUTxO (..)
+    , State (..)
+    , Key0Status (..)
+
+      -- * Chain following model
+    , Tx (..)
+    , Cert (..)
+    , applyTx
+    , setPortfolioOf
+    )
+    where
+
+import Prelude
+
+import Cardano.Crypto.Wallet
+    ( XPub )
+import Cardano.Wallet.Primitive.AddressDerivation
+    ( Depth (..)
+    , DerivationType (..)
+    , Index (..)
+    , MkKeyFingerprint (paymentKeyFingerprint)
+    , Role (..)
+    , SoftDerivation (..)
+    , ToRewardAccount (..)
+    )
+import Cardano.Wallet.Primitive.Types.Address
+    ( Address )
+import Cardano.Wallet.Primitive.Types.Coin
+    ( Coin (..) )
+import Cardano.Wallet.Primitive.Types.Hash
+    ( Hash (..) )
+import Cardano.Wallet.Primitive.Types.RewardAccount
+    ( RewardAccount )
+import Cardano.Wallet.Primitive.Types.Tx
+    ( TxIn (..), TxOut (..) )
+import Control.DeepSeq
+    ( NFData )
+import Data.Maybe
+    ( maybeToList )
+import GHC.Generics
+    ( Generic )
+import Quiet
+    ( Quiet (..) )
+
+import qualified Cardano.Wallet.Primitive.Types.TokenBundle as TB
+
+--------------------------------------------------------------------------------
+-- Delegation State
+--------------------------------------------------------------------------------
+
+-- | Delegation state
+--
+-- === Goals
+-- 1. Allow a wallet to have an arbitrary number of stake keys
+-- 2. Ensure those stake keys can be automatically discovered on-chain
+-- 3. Ensure the wallet is always aware of all stake keys it registers, even in
+-- the case of concurrent user actions on multiple wallet instances, and old
+-- wallet software.
+--
+--
+-- === How
+--
+-- We track a consecutive range of keys that is extended with delegation of the
+-- next unused key, and shrunk with key de-registration. We use a `PointerUTxO`
+-- to ensure that transactions changing the state can't be accepted to the
+-- ledger in any other order than intended. We also need some special care
+-- regarding stake key 0, which old wallet software could try to de-register.
+--
+-- === Diagram
+-- Diagram of states, where the list denotes active (registered /and/ delegating)
+-- keys.
+--
+-- Here we assume the minUTxOValue is 1 ada.
+--
+-- Note that intermediate steps for the `PointerUTxO` should be skipped within a
+-- transaction.
+-- E.g. to transition from [] to [0,1,2] we should deposit 1 ada to key 3,
+-- skipping key 2.
+--
+-- See the implementation of `setPortfolioOf` and `applyTx` for more details.
+--
+-- @
+-- ┌────────────────────┐           ┌────────────────────┐                     ┌────────────────────┐            ┌─────────────────────┐
+-- │                    │           │                    │                     │                    │            │                     │
+-- │                    │           │                    │       Pointer       │                    │            │                     │
+-- │                    │──────────▶│                    │ ──────deposit──────▶│                    │ ─────────▶ │                     │ ─────────▶
+-- │                    │           │                    │                     │       [0,1]        │            │       [0,1,2]       │
+-- │         []         │           │        [0]         │                     │1 ada held by key 2 │            │ 1 ada held by key 3 │
+-- │                    │           │                    │                     │                    │            │                     │
+-- │                    │           │                    │       Pointer       │                    │            │                     │
+-- │                    │◀──────────│                    │ ◀─────deposit ──────│                    │◀────────── │                     │◀──────────
+-- │                    │           │                    │       returned      │                    │            │                     │
+-- └────────────────────┘◀──┐       └────────────────────┘                     └────────────────────▲            ▲─────────────────────▲            ▲
+--                          └───┐                                                     │       ▲     └─┐         ┌┘      │       ▲      └─┐         ┌┘
+-- Normal states                └───┐                                                 │       │       └─┐     ┌─┘       │       │        └─┐     ┌─┘
+--                                  └───┐                                             │       │         └─┐ ┌─┘         │       │          └─┐ ┌─┘
+-- ╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳└───┐╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳╳│╳╳╳╳╳╳╳│╳╳╳╳╳╳╳╳╳╳╳│ │╳╳╳╳╳╳╳╳╳╳╳│╳╳╳╳╳╳╳│╳╳╳╳╳╳▶     ├─┤
+--                                          └───┐                                     │       │         ┌─┘ └─┐         │       │          ┌─┘ └─┐
+-- States caused by                             └────┐Pointer                         ▼       │       ┌─┘     └─┐       ▼       │        ┌─┘     └─┐
+-- old wallet                                        └deposit                  ┌────────────────────┐─┘         └┬─────────────────────┐─┘         └─
+-- de-registering                                     returned─┐               │                    │            │                     │
+-- stake-key 0                                                 └────┐          │                    │ ─────────▶ │                     │ ─────────▶
+-- of multi-stake                                                   └────┐     │                    │            │                     │
+-- key wallet                                                            └──── │        [1]         │            │        [1,2]        │
+--                                                                             │1 ada held by key 2 │            │ 1 ada held by key 3 │
+--                                                                             │                    │            │                     │
+--                                                                             │                    │◀────────── │                     │◀──────────
+--                                                                             │                    │            │                     │
+--                                                                             │                    │            │                     │
+--                                                                             └────────────────────┘            └─────────────────────┘
+-- @
+data DelegationState k = DelegationState
+    { -- | The account public key from which the stake keys should be derived.
+      rewardAccountKey :: k 'AccountK XPub
+    , state :: State
+    } deriving (Generic)
+
+-- | Construct the initial delegation state.
+initialDelegationState
+    :: k 'AccountK XPub
+    -> DelegationState k
+initialDelegationState accK = DelegationState accK Zero
+
+-- | The internal state, without the account key.
+--
+-- TODO: Perhaps we should model this as
+-- @S = S1 * S2, where S1 = Bool, S2 = ix * UTxOPointer@ - having two
+-- "concurrent" states tracking stake keys, where the first one is identical to
+-- legacy single-stake key wallets.
+--
+-- Maybe that would help simplify `applyTx` and `setPortfolioOf`...
+data State
+    -- | No active stake keys. The initial state of a new wallet.
+    = Zero
+
+    -- | The first stake-key (index 0) is registered and either delegating or
+    -- about to be delegating.
+    | One
+
+    -- | There is more than one active stake keys. Can only be reached using
+    -- wallets with support for multiple stake keys.
+    | More
+        !(Index 'Soft 'AddressK)
+          -- nextKeyIx - the ix of the next unused key
+        PointerUTxO
+          -- ^ pointer utxo that need to be spent when changing state.
+        !Key0Status
+    deriving (Eq, Show, Generic)
+
+instance NFData State
+
+-- | Is key 0 still registered? For compatibility with
+-- single-stake-key wallets, we need to track this.
+--
+-- >>> activeKeys (More (toEnum 3) p ValidKey0)
+-- [0, 1, 2]
+--
+-- >>> activeKeys (More (toEnum 3) p MissingKey0)
+-- [1, 2]
+--
+-- (pseudocode; requires a bit more boilerplate to compile)
+--
+-- See the implementaiton of @applyTx@ for how it is used.
+data Key0Status = ValidKey0 | MissingKey0
+    deriving (Eq, Show, Generic)
+
+instance NFData Key0Status
+
+instance (NFData (k 'AccountK XPub), NFData (k 'AddressK XPub))
+    => NFData (DelegationState k)
+
+deriving instance
+    ( Show (k 'AccountK XPub)
+    , Show (k 'AddressK XPub)
+    ) => Show (DelegationState k)
+
+deriving instance
+    ( Eq (k 'AccountK XPub)
+    , Eq (k 'AddressK XPub)
+    ) => Eq (DelegationState k)
+
+keyAtIx
+    :: (SoftDerivation k)
+    => DelegationState k
+    -> Index 'Soft 'AddressK
+    -> k 'AddressK XPub
+keyAtIx s = deriveAddressPublicKey (rewardAccountKey s) MutableAccount
+
+nextKeyIx
+    :: DelegationState k
+    -> Index 'Soft 'AddressK
+nextKeyIx s = case state s of
+    Zero -> minBound
+    One -> succ minBound
+    More ix _ _ -> ix
+
+lastActiveIx
+    :: DelegationState k
+    -> Maybe (Index 'Soft 'AddressK)
+lastActiveIx s
+    | nextKeyIx s == minBound = Nothing
+    | otherwise               = Just $ pred $ nextKeyIx s
+
+data PointerUTxO = PointerUTxO { pTxIn :: TxIn, pCoin :: Coin }
+    deriving (Generic, Eq, Show)
+    deriving anyclass NFData
+
+-- | Returns the index corresponding to the payment key the `PointerUTxO`
+-- should be locked with for a portfolio of a given size @n@.
+--
+-- In our current implementation we require the `PointerUTxO` to be created in
+-- the @[0] -> [0,1] transition@, i.e. @nextKeyIx 1 -> nextKeyIx 2@.
+pointerIx
+    :: Int
+    -> Maybe (Index 'Soft 'AddressK)
+pointerIx 0 = Nothing
+pointerIx 1 = Nothing
+pointerIx n = Just $ toEnum n
+
+-- | Retrieve the `PointerUTxO` from a `DelegationState` if it has one.
+pointer :: DelegationState k -> Maybe PointerUTxO
+pointer (DelegationState _ (More _ p _)) = Just p
+pointer _ = Nothing
+
+--------------------------------------------------------------------------------
+-- Chain
+--------------------------------------------------------------------------------
+
+-- | A transaction type specific to `DelegationState`.
+--
+-- Intented to be converted both from and to a more real transaction type.
+--
+-- When constructing a real transaction from `Tx`, these `outputs`
+-- should appear before other outputs. In the theoretical event that there's
+-- also a user-specified output with the same payment key as the pointer output,
+-- `applyTx` will track the first one as the new pointer.
+data Tx = Tx
+    { certs :: [Cert]
+    , inputs :: [(TxIn, Coin)]
+    , outputs :: [TxOut]
+    }
+    deriving (Eq, Generic)
+    deriving Show via (Quiet Tx)
+
+instance Semigroup Tx where
+    (Tx cs1 is1 os1) <> (Tx cs2 is2 os2) =
+        Tx (cs1 <> cs2) (is1 <> is2) (os1 <> os2)
+
+data Cert
+    = RegisterKey RewardAccount
+    | Delegate RewardAccount
+      -- ^ Which pool we're delegating to is here (and for now) irrelevant.
+      -- The main thing is that there exists a witness on-chain for this stake
+      -- key (registration certs don't require witnesses)
+      --
+      -- TODO: We may also want to add the PoolId here.
+    | DeRegisterKey RewardAccount
+    deriving (Eq, Show, Generic)
+
+-- | Given a `DelegationState`, produce a `Tx` registering or de-registering
+-- stake-keys, in order to have @n@ stake-keys.
+--
+-- E.g. @setPortfolioOf s0 _ _ 2@ creates a tx which after application causes
+-- the state to have @activeKeys == [0,1]@
+--
+-- Returns @Nothing@ if the target @n@ is already reached.
+setPortfolioOf
+    :: (SoftDerivation k, ToRewardAccount k)
+    => DelegationState k
+    -> Coin
+        -- ^ minUTxOVal
+    -> (k 'AddressK XPub -> Address)
+        -- ^ A way to construct an Address
+    -> (RewardAccount -> Bool)
+        -- ^ Whether or not the key is registered.
+        --
+        -- TODO: Need a Set or Map for the real implementation with LSQ.
+    -> Int
+        -- ^ Target number of stake keys.
+    -> Maybe Tx
+setPortfolioOf ds minUTxOVal mkAddress isReg n =
+    repairKey0IfNeededTx <> changeStateTx
+  where
+    -- The `changeStateTx` calculation assumes that key 0 is registered. If it
+    -- is not, we fix it here.
+    --
+    -- At some point we will need to take a `Set PoolId` instead of `n`. If
+    -- we're in the state of key 1 and 2 delegating to pools B and C
+    -- respectively, we likely want:
+    -- - `setPortfolio [A,B,C]` to delegate key 0 to A, instead of key 3.
+    -- - `setPortfolio [B,C]` to replace key 2 with key 0
+    repairKey0IfNeededTx :: Maybe Tx
+    repairKey0IfNeededTx = case repairKey0 $ state ds of
+        [] -> Nothing
+        cs -> Just $ Tx cs [] []
+      where
+        repairKey0 (More _ _ MissingKey0) | n > 0 = deleg [minBound]
+        repairKey0 _ = []
+
+    changeStateTx :: Maybe Tx
+    changeStateTx = txWithCerts $ case compare (toEnum n) (nextKeyIx ds) of
+        GT -> deleg [nextKeyIx ds .. toEnum (n - 1)]
+        EQ -> []
+        LT -> dereg $ reverse [toEnum n .. (pred $ nextKeyIx ds)]
+
+      where
+        txWithCerts [] = Nothing
+        txWithCerts cs = Just $ Tx
+            { certs = cs
+            , inputs = maybeToList (mkTxIn <$> pointer ds)
+            , outputs = maybeToList $
+                -- Note that this is the only place where we move the pointer.
+                -- I.e. repairKey0IfNeededTx won't do it on its own.
+                (\i -> TxOut
+                    (mkAddress $ keyAtIx ds i)
+                    (TB.fromCoin minUTxOVal)
+                ) <$> pointerIx n
+            }
+          where
+            mkTxIn (PointerUTxO txIx coin) = (txIx, coin)
+            -- Note: If c > minUTxOVal we need to rely on the wallet to return the
+            -- difference to the user as change.
+
+    deleg :: [Index 'Soft 'AddressK] -> [Cert]
+    deleg = (>>= \ix ->
+        if isReg (acct ix)
+        then [Delegate (acct ix)]
+        else [RegisterKey (acct ix),  Delegate (acct ix)]
+        )
+
+
+    dereg :: [Index 'Soft 'AddressK] -> [Cert]
+    dereg ixs =
+        [ DeRegisterKey (acct ix)
+        | ix <- ixs
+        , isReg . acct $ ix
+        -- We need to /at least/ check @isReg (key 0)@, because the user could
+        -- have deregistered it using old wallet software.
+        ]
+
+    acct = toRewardAccount . keyAtIx ds
+
+-- | Apply a `Tx` to a `DelegationState`.
+--
+-- Expects the @PointerUTxO@ to be correctly managed, and will panic otherwise.
+applyTx
+    :: forall k. ( SoftDerivation k
+        , ToRewardAccount k
+        , MkKeyFingerprint k Address
+        , MkKeyFingerprint k (k 'AddressK XPub))
+    => Tx
+    -> Hash "Tx"
+    -> DelegationState k
+    -> DelegationState k
+applyTx (Tx cs _ins outs) h ds0 = foldl applyCert ds0 cs
+  where
+    applyCert ds cert = flip modifyState ds $ case cert of
+            RegisterKey _                   -> id
+            Delegate k
+               | k == nextKey ds            -> inc
+               | otherwise                  -> modifyKey0 cert
+            DeRegisterKey k
+               | Just k == lastActiveKey ds -> dec
+               | otherwise                  -> modifyKey0 cert
+      where
+        inc s = case s of
+            Zero -> One
+            One -> More (toEnum 2) (findOut $ toEnum 2) ValidKey0
+            More ix _ is0Reg -> let ix' = succ ix in More ix' (findOut ix') is0Reg
+        dec s = case s of
+            Zero -> error "impossible: can't decrement beyond zero"
+            One -> Zero
+            More ix _ is0Reg
+                | fromEnum ix > 2 -> let ix' = pred ix in More ix' (findOut ix') is0Reg
+                | otherwise -> case is0Reg of
+                    ValidKey0   -> One
+                    MissingKey0 -> Zero
+
+        findOut ix = case map mkPointer pointerOuts of
+            (x:_) -> x
+            _     -> error $ mconcat
+                [ "couldn't find pointer output for ix "
+                , show ix
+                , " with state "
+                , show $ state ds
+                ]
+          where
+            isOurOut (TxOut addr _b) =
+                case (paymentKeyFingerprint @k $ keyAtIx ds ix, paymentKeyFingerprint addr) of
+                (Right fp, Right fp')
+                    | fp == fp' -> True
+                    | otherwise -> False
+                _ -> False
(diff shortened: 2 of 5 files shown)


lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/DelegationSpec.hs
```diff
@@ -0,0 +1,350 @@
...
+                        "\nAccepted txs:\n" <> unlines (map show txs) <> "\n"
+                forAllShow (sublistOf chain) showTxs $ \subChain -> do
+                    let s = apply subChain s0
+                    case (applyLedger subChain initialLedger) of
+                        Right (Ledger regs) -> counterexample
+                            "valid chain => expecting the wallet's activeKeys \
+                            \to match all registered keys in the ledger" $
+                            Set.fromList (fmap toRewardAccount (activeKeys s))
+                                === regs
+                        Left _ -> label "invalid chain" True
```
Anviking: Can now show that rollbacks are a problem!
```
 1) Mock ledger, Dropped or re-ordered transactions, chain valid to ledger => activeKeys == regs ledger
       Falsified (after 13 tests and 6 shrinks):
         [CmdSetPortfolioOf 1,CmdSetPortfolioOf 2]

         Accepted txs:
         Tx [RegisterKey (RewardAccount "1"),Delegate (RewardAccount "1")]


         valid chain => expecting the wallet's activeKeys to match all registered keys in the ledger
         fromList [] /= fromList [RewardAccount "1"]
```

Anviking: Seems `sublistOf` cannot re-order elements. Not sure if this matters. Maybe 🤔 Should either fix or add a comment at least.

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/DelegationSpec.hs
```diff
@@ -0,0 +1,329 @@
...
+    map (CmdSetPortfolioOf . length . activeKeys)
+    . scanl (flip applyTx) (initialDelegationState accK)
+
+chainFromCmds :: [Cmd] -> [Tx]
+chainFromCmds goals = map Tx . filter (not . null) $ reverse $ go 0 [] [] goals
+  where
+    -- NOTE: For convenience we treat ix=-1 as no keys, as ix=0 means the first
+    -- key (at index 0) is registered and delegating.
+    go i chain tx (cmd:rest) = case cmd of
+        CmdSetPortfolioOf n
```
Anviking: Just realized that this will never generate delegation certificates on their own, which should be pretty important to test.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
```diff
@@ -0,0 +1,361 @@
...
+      -- having a length of @nextKeyIx@.
+      --
+      -- E.g. a state with nextKeyIx=0 has no keys, and nextKeyIx=2 has the keys
+      -- [0,1].
+    , nextKeyIx :: !(Index 'Soft 'AddressK)
+
+      -- | By threading a long a "pointer UTxO", we leverage the ledger rules to
+      -- ensure transactions cannot be re-ordered in a way that breaks the
+      -- rules/assumtions for how stake-keys are managed.
+    , pointer :: !(Maybe PointerUTxO)
```
Anviking: This `PointerUTxO` now also contains the `nextKeyIx`, so we _should_ be able to remove one of them. I think.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
```diff
@@ -0,0 +1,361 @@
...
+        -> DelegationState k
+    modifyIx f s = s { nextKeyIx = f (nextKeyIx s) }
+
+    applyCert c s = modifyKey0 $ flip modifyIx s $ case c of
+        RegisterKey _ -> id
+        Delegate k
+            | k == nextKey s            -> succ
+            | otherwise                 -> id
+        DeRegisterKey k
+            | Just k == lastActiveKey s && hitGap k -> pred . pred -- i.e. to 0
```
Anviking: If we 
1. have the active keys [0,1]
2. Use an old wallet to de-register key 0
3. See a de-registration certificate for key 1

Then our active keys should be [] (and not [0])

If our new wallet _wants_ to go to [0], it should use `[DeReg 1, Reg 0, Delegate 0]`, i.e. re-registering the first key.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
```diff
@@ -0,0 +1,361 @@
...
+import Quiet
+    ( Quiet (..) )
+
+import qualified Cardano.Wallet.Primitive.Types.TokenBundle as TB
+
+--------------------------------------------------------------------------------
+-- Stake Key State
+--------------------------------------------------------------------------------
+
+data DelegationState k = DelegationState
```
Anviking: Should probably rename to something like `StakeKeyState k`.

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/DelegationSpec.hs
```diff
@@ -0,0 +1,423 @@
...
+
+
+applyCmds :: Env -> [Cmd] -> Env
+applyCmds = foldl (flip stepCmd)
+
+applyTxs :: Env -> [Tx] -> Env
+applyTxs = foldl (flip tryApplyTx)
+
+txid :: Tx -> Hash "Tx"
+txid = Hash . BS.take 4 . blake2b224 . B8.pack . show
```
Anviking: At one point I wanted a lookup-map to assign txs with sequentially incrementing ids — like tx0, tx1, tx2 — but didn't end up doing it. As long as they are formatted nicely, short hashes are fairly readable.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
Anviking: Removed in cc1f2bc

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
```diff
@@ -0,0 +1,361 @@
...
+    ( Eq (k 'AccountK XPub)
+    , Eq (k 'AddressK XPub)
+    ) => Eq (DelegationState k)
+
+accountAtIx
+    :: (SoftDerivation k, ToRewardAccount k)
+    => DelegationState k
+    -> Index 'Soft 'AddressK
+    -> RewardAccount
+accountAtIx s = toRewardAccount . deriveAddressPublicKey (rewardAccountKey s) MutableAccount
```
KtorZ: Minor: `account` only is confusing, because this is still at the 5th level. I'd go for `rewardAccountAt`

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
```diff
@@ -0,0 +1,361 @@
...
+        -- changes. Not sure if this is the case.
+        txWithCerts [] = Nothing
+        txWithCerts cs = Just $ Tx
+            { certs = cs
+            , inputs = maybeToList (mkTxIn <$> pointer s)
+            , outputs =
+                [ TxOut
+                    (mkAddress $ keyAtIx s $ nextKeyIx s')
+                    (TB.fromCoin minUTxOVal)
+                ]
```
KtorZ: I was thinking whether or not we want to create the pointer utxo if de-registering the key at index=1. The rationale for not creating it would be to keep the process symmetric: there's no pointer if only the key at index=0 is registered. 

To keep this compatibility somewhat possible, I have in the proposal only required that the UTxO is created for key registration beyond index=0 (non inclusive) and for key de-registration up to index=2 (inclusive) as I _think_ this would help the implementation to have only one well-defined behavior regarding the first key.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
```diff
@@ -0,0 +1,361 @@
...
+        [(ix,TxOut _addr tb)]
+            -> s { pointer = Just $ PointerUTxO (TxIn h ix) (TB.getCoin tb) }
+        (_x:_) -> error "todo"
+            -- There shouldn't be more than one pointer, but theoretically
+            -- possible. If a user sends funds to an address corresponding to
+            -- the stake key (why would they do this though?), we could mistake
+            -- it for the pointer.
+            --
+            -- TODO: Is this actually bad though? Or is it good because it will
+            -- then be sent back as change? Any other problems?
```
KtorZ: That's also something I stumble upon when updating the proposal which I dismissed by requiring that the pointer can only be a UTxO that is part of a transaction which also contains delegation certificate or key de-registration certificates. There may be other UTxO at the reward address, but they should not count as pointer (and the wallet should be free to ignore them or redeem them, whatever is the simplest). 

It however forces the wallet to "follow the pointer" in a way, and simply querying the LSQ isn't sufficient to figure out whether a pointer exist. I think that following the pointer was however already necessary in order to spend it right (we need to have the exact TxIn for that) ?

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
```diff
@@ -0,0 +1,361 @@
...
+        s = foldl (flip applyCert) s0 cs
+        isOurOut i (TxOut addr _b) =
+            case (paymentKeyFingerprint @k $ keyAtIx s i, paymentKeyFingerprint addr) of
+            (Right fp, Right fp')
+                | fp == fp' -> True
+                | otherwise -> False
+            _ -> False
+        -- To workaround rollback + old wallets, we also look for the pointer at
+        -- 0 and 1, aside from nextKeyIx.
+        f o = any (`isOurOut` o) [ nextKeyIx s, minBound, succ minBound ]
```
KtorZ: Can do better than `f` :wink:

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
```diff
@@ -0,0 +1,361 @@
...
+        , MkKeyFingerprint k Address
+        , MkKeyFingerprint k (k 'AddressK XPub))
+    => Tx
+    -> Hash "Tx"
+    -> DelegationState k
+    -> DelegationState k
+applyTx (Tx cs _ins outs) h s0 =
+    let
+        s = foldl (flip applyCert) s0 cs
+        isOurOut i (TxOut addr _b) =
```
KtorZ: To identify if a transaction is relevant to the `DelegationState`, I think it would suffice to check for delegation or key de-registration certificate containing either the next key index, or the last active key index. We don't really need to worry about output. The CIP explicitly requires that the first output of a such a transaction is the pointer output; 

Thus, I would do it the other way around (also because it may be _much_ faster that way): lookup for certificates relevant to us, figure out from the certificate what is the last active key index and then, have a guard / invariant checking that the first output does indeed match. 

In case it doesn't, that is an invariant violation and the wallet behavior is undefined. Crashing seems excessive although., beyond this point we really do not know what the delegation state of the wallet is :thinking:

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
Anviking: Initially I was worried that could lead to multiple interfering histories/pointers, but maybe it's not actually the case if the pointer can only be created in one transition 🤔 Can sanity check in tests…

Anviking: Tests still pass -> cee3a604eb6f279565d7f42fe7ecb2a90df3b3a1 

Good idea!

Anviking: Also, because of this, we could maybe re-work `DelegationState` to eliminate impossible states a bit:

I.e. rather than
```haskell
    , nextKeyIx :: !(Index 'Soft 'AddressK)

      -- | By threading a long a "pointer UTxO", we leverage the ledger rules to
      -- ensure transactions cannot be re-ordered in a way that breaks the
      -- rules/assumtions for how stake-keys are managed.
    , pointer :: !(Maybe PointerUTxO)

      -- | For compatibility with single-stake-key wallets, we track whether
      -- stake key 0 is registered or not separately.
      --
      -- See the implementaiton of @applyTx@ for how it is used.
    , isKey0Reg :: Bool
```
something more like:
```haskell
data State
    -- | No active stake keys. The initial state of a new wallet.
    = Zero

    -- | The first stake-key (index 0) is registered and either delegating or
    -- about to be delegating.
    | One

    -- | There is more than one active stake keys
    | More
        !(Index 'Soft 'AddressK) -- nextKeyIx - the ix of the next unused key
        !PointerUTxO -- pointer utxo that need to be spent when changing state.
        !Bool -- Is key 0  still registered?
```
👀

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
Anviking: Now removed 😛

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
Anviking: I'm sticking with `DelegationState` for now.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
Anviking: I ended up dropping the entire function instead.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
Anviking: 👍 I changed it to assume the pointer _only_ changes based on the delegation certificates.

> The CIP explicitly requires that the first output of a such a transaction is the pointer output

Wouldn't that be an unnecessary loss of flexibility? 🤔 (Not that I know what we'd use it for, maybe changing portfolio of multiple accounts at the same time 🤷‍♂️ )

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
Anviking: Done

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
Anviking: > by requiring that the pointer can only be a UTxO that is part of a transaction which also contains delegation certificate or key de-registration certificates

Now done.

> It however forces the wallet to "follow the pointer" in a way, and simply querying the LSQ isn't sufficient to figure out whether a pointer exist. I think that following the pointer was however already necessary in order to spend it right (we need to have the exact TxIn for that) ?

I haven't put much thought about using LSQ here, but I don't see why it would help. Now we do need to store a `TxIn`, but it's not that complicated.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
```diff
@@ -0,0 +1,464 @@
...
+-- should be locked with for a portfolio of a given size @n@.
+--
+-- In our current implementation we require the `PointerUTxO` to be created in
+-- the @[0] -> [0,1] transition@, i.e. @nextKeyIx 1 -> nextKeyIx 2@.
+pointerIx
+    :: Int
+    -> Maybe (Index 'Soft 'AddressK)
+pointerIx 0 = Nothing
+pointerIx 1 = Nothing
+pointerIx n = Just $ toEnum n
```
Anviking: Now we have
`[0,1], 1 ada in held by key 2`
because I previously allowed
`[], 1 ada in held by key 0`

Now the latter state is not possible. So we _could_ have the pointer correspond to to the `lastActive` rather than the `nextUnused` ix. I.e.
`[0,1], 1 ada in held by key 1`

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
KtorZ: > Wouldn't that be an unnecessary loss of flexibility? thinking (Not that I know what we'd use it for, maybe changing portfolio of multiple accounts at the same time man_shrugging )

More of a convention (perhaps a useless one?). In the end, since the wallet tracks the UTxO it's quite easy to locate it in the input set.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
```diff
@@ -0,0 +1,464 @@
...
+
+    -- | There is more than one active stake keys. Can only be reached using
+    -- wallets with support for multiple stake keys.
+    | More
+        !(Index 'Soft 'AddressK)
+          -- nextKeyIx - the ix of the next unused key
+        PointerUTxO
+          -- ^ pointer utxo that need to be spent when changing state.
+        !Bool
+          -- ^ Is key 0 still registered? For compatibility with
```
KtorZ: A little newtype or sum-type would be better :pray:

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
KtorZ: I am a little bit confused by this function at the moment, maybe it will get clearer in the rest of the review. However, I would expect the pointer to be created during the transition [0] -> [0,1] (as you state in comment) and that, `pointerIx n = Just n`; It feels more intuitive to me for the pointer to track the _highest active key_.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
```diff
@@ -0,0 +1,464 @@
...
+
+--------------------------------------------------------------------------------
+-- Constructors
+--------------------------------------------------------------------------------
+
+-- | Construct the initial delegation state.
+initialDelegationState
+    :: k 'AccountK XPub
+    -> DelegationState k
+initialDelegationState accK = DelegationState accK Zero
```
KtorZ: Minor remark: this is quite far from the `DelegationState` definition. Since this is pretty much a smart constructor for that data-type I'd expect to see it close to the definition. (module exports also suggests that).

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
```diff
@@ -0,0 +1,464 @@
...
+-- should appear before other outputs. In the theoretical event that there's
+-- also a user-specified output with the same payment key as the pointer output,
+-- `applyTx` will track the first one as the new pointer.
+data Tx = Tx
+    { certs :: [Cert]
+    , inputs :: [(TxIn, Coin)]
+    , outputs :: [TxOut]
+    }
+    deriving (Eq, Generic)
+    deriving Show via (Quiet Tx)
```
KtorZ: You could potentially use a type-class for that to be more like a DSL for which the primitive Tx is one possible implementation. No need to perform conversions back and forth with this; only an instance of the class is needed which is likely trivial to write. 

```hs
class DelegationStateTx tx of
    certs :: tx -> [Cert]
    inputs :: tx -> [(TxIn, Coin)]
    outputs :: tx -> [TxOut]
```

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
```diff
@@ -0,0 +1,464 @@
...
+    repairKey0IfNeededTx = case repairKey0 $ state s of
+        [] -> Nothing
+        cs -> Just $ Tx cs [] []
+      where
+        repairKey0 (More _ _ False) | n > 0 = deleg [minBound]
+        repairKey0 _ = []
+
+    changeStateTx :: Maybe Tx
+    changeStateTx = txWithCerts $ case compare (toEnum n) (nextKeyIx s) of
+        GT -> deleg [nextKeyIx s .. toEnum (n - 1)]
```
KtorZ: Took me some time to convince myself that the next index was inded `toEnum (n - 1)` partly because it feels to me that the logic of `nextKeyIx` is sort of "leaking out". That's the kind of moment where I think functions like `scanl / iterate` fit nicely to avoid silly mistake, imagine a function `nextState :: DelegationState -> DelegationState`, such that it can then be written as:

```hs
take n $ iterate nextState s 
```

This reads much better to me as in "give me the next `n` states"; and the same approach can be done when going in the other direction using a `predState` analogous function.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
```diff
@@ -0,0 +1,464 @@
...
+                -- I.e. repairKey0IfNeededTx won't do it on its own.
+                (\i -> TxOut
+                    (mkAddress $ keyAtIx s i)
+                    (TB.fromCoin minUTxOVal)
+                ) <$> pointerIx n
+            }
+          where
+            mkTxIn (PointerUTxO i c) = (i, c)
+            -- Note: If c > minUTxOVal we need to rely on the wallet to return the
+            -- difference to the user as change.
```
KtorZ: That will be interesting to incorporate in the coin selection and is similar to some of the needs of the Plutus application backend. The wallet will need to be able to "continue" the construction of a transaction from a pre-filled one.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
```diff
@@ -0,0 +1,464 @@
...
+    :: (SoftDerivation k, ToRewardAccount k)
+    => DelegationState k
+    -> Coin
+        -- ^ minUTxOVal
+    -> (k 'AddressK XPub -> Address)
+        -- ^ A way to construct an Address
+    -> (RewardAccount -> Bool)
+        -- ^ Whether or not the key is registered.
+        --
+        -- TODO: Need a Set or Map for the real implementation with LSQ.
```
KtorZ: How are you thinking to approach this function without any monadic effects though :thinking:  ? The caller of `setPortfolioOf` won't know the accounts upfront, so there's little chance that it'll be able to resolve that predicate without having to query the node? Or am I missing something?

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
```diff
@@ -0,0 +1,464 @@
...
+    -> Int
+        -- ^ Target number of stake keys.
+    -> Maybe Tx
+setPortfolioOf s minUTxOVal mkAddress isReg n =
+    repairKey0IfNeededTx <> changeStateTx
+  where
+    repairKey0IfNeededTx :: Maybe Tx
+    repairKey0IfNeededTx = case repairKey0 $ state s of
```
KtorZ: A comment to clarify what this does would be nice. If I understand correctly, this re-register the initial key in case where it has been de-registered by some other mean? Although I do think the code is correct, I am also wondering why this is _necessary_ ? I am generally quite against doing anything on the behalf of users without explicit consent. Imagine the scenario where a user has restored the same wallet in Daedalus and Yoroi. She then de-registers her first stake key in Yoroi which allows it because it doesn't implement this proposal yet. Then, some weeks later while she re-delegate to new pools, the first stake key gets re-registered without her asking anything. 

I don't think the first key should actually be handled so specially. Rather, the multi-stake key state should really only care about keys beyond the first one. If the first key has been de-registered, then so be it :shrug: ... Checking this before trying to de-register it (as you do below) should be sufficient.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
Anviking: I'm thinking something like:
```haskell
let accounts = Set.fromList $ accoutAtIx [0 .. n]
(registrationStatus :: Map RewardAccount Bool) <- checkIfRegistered nl
setPortfolioOf n registrationStatus
```
Or taking a function of `Set RewardAccount -> m (Map RewardAccount Bool)` (and making `setPortfolioOf n` return `m` too) if it turns out to be neater.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
Anviking: I think it's neat to return the same `Tx` type from `cmdSetPortfolioOf`, as what we take in `applyCmd`. And I like defaulting to no type classes.

Otherwise you'd have some kind of `mkTx :: certs -> ins -> outs -> tx` (which might not make sense), or some  "`TxRecipe`" in addition.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
Anviking: > It feels more intuitive to me for the pointer to track the highest active key

Yeah, I was starting to feel the same. Might try that. Here it would be `pointerIx n = Just $ pred $ toEnum n` though, as `n` is the number of stake keys you want.

That said, there should be some _small_ fee-tradeoffs from this choice (if I understand correctly)

With the pointer tracking `nextKeyIx`:
- Adding stake keys: we need to spend the pointer at `nextKeyIx`, _and_ delegate with `nextKeyIx`, so we save a witness.
- Removing stake keys: no savings

With the pointer tracking `lastActiveIx`:
- Adding stake keys: no savings
- Removing stake keys: we need to spend the pointer at `lastActiveIx`, _and_ deregister with `lastActiveIx`, so we save a witness.

But I guess the difference is so small it's not even worth thinking about with 44 lovelace/byte.

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/DelegationSpec.hs
```diff
@@ -0,0 +1,556 @@
...
+                usableKeys s2 `shouldBe` [toEnum 0, toEnum 1]
+
+        let s3 = regAndDeleg 5 s2
+        describe "Impossible gaps in stake keys (shouldn't happen unless\
+            \ someone manually constructs txs to mess with the on-chain state)" $ do
+            it "presentableKeys == [0, 1, 2] (doesn't find 5)" $ do
+                (presentableKeys s3) `shouldBe`
+                    [toEnum 0, toEnum 1, toEnum 2]
+            it "usableKeys == [0, 1]" $ do
+                usableKeys s3 `shouldBe` [toEnum 0, toEnum 1]
```
KtorZ: :+1:

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/DelegationSpec.hs
```diff
@@ -0,0 +1,556 @@
...
+
+    it "(usableKeys s) are consequtive" $ property
```
KtorZ: ~~consequtive~~ -> consecutive ?

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/DelegationSpec.hs
```diff
@@ -0,0 +1,556 @@
...
+    -- Whatever happens, we should be able to recover using a single
+    -- @CmdSetPortfolioOf n@, and be concistent with the ledger.
+    it "can recover from dropped transactions" $ withMaxSuccess 2000
+        prop_rollbacks
+
+--
+-- Properties
+--
+
+prop_keysConsequtive :: (DelegationState StakeKey' -> [StakeKey]) -> [Cmd] -> Property
```
KtorZ: :+1:

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/DelegationSpec.hs
```diff
@@ -0,0 +1,556 @@
...
+-- | PointerUTxO should only be created/destroyed in very specific cases.
+prop_pointerRules :: [Cmd] -> Property
+prop_pointerRules cmds = do
+    let env = applyCmds env0 cmds
+    forAllTxs env $ \case
+        Tx [] []    [_o] -> True -- Not the actual pointer
+        Tx cs []    [_o] -> Delegate (acct 1) `elem` cs
+        Tx cs [_i]  []   -> DeRegisterKey (acct 1) `elem` cs
+        Tx cs [_i]  [_o] -> any ((> 1) . ixFromCert) cs
+        Tx cs []    []   -> all ((< 1) . ixFromCert) $ filter notReg cs
```
KtorZ: :+1:

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/DelegationSpec.hs
```diff
@@ -0,0 +1,556 @@
...
+        $ \tx -> counterexample (pretty tx) $ prop tx
+
+-- | Tests that `isAdversarial` commands have no effect on the usableKeys of the
+-- state.
+prop_usableKeysAdversaries :: [Cmd] -> Property
+prop_usableKeysAdversaries cmds = do
+    counterexample "\nstate /= state without adversarial cmds" $ do
+        let usableKeys' = usableKeys . wallet . applyCmds env0
+        usableKeys' cmds
+            === usableKeys' (filter (not . isAdversarial) cmds)
```
KtorZ: :+1:

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/DelegationSpec.hs
```diff
@@ -0,0 +1,556 @@
...
+
+-- | Tests that `isAdversarial` commands have no effect on the `PointerUTxO`.
+prop_pointerAdversaries :: NonNegative Int -> [RewardAccount] -> Property
+prop_pointerAdversaries (NonNegative n) keys = do
+    let env = applyCmds env0 [CmdSetPortfolioOf n]
+    let env' = applyCmds env (map CmdMimicPointerOutput keys)
+    state (wallet env') === state (wallet env)
```
KtorZ: :+1:

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/DelegationSpec.hs
```diff
@@ -0,0 +1,556 @@
...
+
+-- | CmdSetPortfolioOf 0 returns us to the initial state.
+prop_canAlwaysGoTo0 :: [Cmd] -> Property
+prop_canAlwaysGoTo0 cmds = do
+    let env = applyCmds env0 (cmds ++ [CmdSetPortfolioOf 0])
+    counterexample (pretty env) $
+        activeKeys (wallet env) === activeKeys (wallet env0)
```
KtorZ: :+1:

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/DelegationSpec.hs
```diff
@@ -0,0 +1,556 @@
...
+        let isSubsetOf a b = counterexample (show a <> " ⊄ " <> show b)
+                $ a `Set.isSubsetOf` b
+
+        let allActiveKeysRegistered e =
+                Set.map toRewardAccount (Set.fromList (activeKeys $ wallet e))
+                    `isSubsetOf` regs (ledger e)
+
+        counterexample (pretty env) $
+            length (activeKeys $ wallet env) === n
+                .&&. allActiveKeysRegistered env
```
KtorZ: :+1:

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/DelegationSpec.hs
```diff
@@ -0,0 +1,556 @@
...
+      -- depending on whether it is registered or not.
+      --
+      -- Having a "toggle"-command instead of two separate commands, makes
+      -- generating valid arbitrary values easier.
+    | CmdAdversarialReg RewardAccount
+      -- ^ Someone could pay 2 ada to (re-)register your stake key. Your wallet
+      -- shouldn't be affected negatively from it.
+    | CmdMimicPointerOutput RewardAccount
+      -- ^ Someone could send funds to the same UTxO
+    deriving (Generic, Eq)
```
KtorZ: Nice :ok_hand:

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/DelegationSpec.hs
```diff
@@ -0,0 +1,556 @@
...
+
+isAdversarial :: Cmd -> Bool
+isAdversarial (CmdSetPortfolioOf _) = False
+isAdversarial (CmdAdversarialReg _) = True
+isAdversarial CmdOldWalletToggleFirstKey = False
+isAdversarial (CmdMimicPointerOutput _) = True -- Kind of
```
KtorZ: Don't be fooled by appearances; because they send us money, doesn't mean they are friendly :smile:, that's definitely adversarial.

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/DelegationSpec.hs
```diff
@@ -0,0 +1,556 @@
...
+
+instance Arbitrary Cmd where
+    -- We don't want to generate too many adversarial registrations (we don't
+    -- expect them to happen in reality), but at least some, and enough to cause
+    -- consistent failures if something is wrong.
+    arbitrary = frequency
+        [ (85, CmdSetPortfolioOf . getNonNegative <$> arbitrary)
+        , (5, CmdAdversarialReg <$> arbitrary)
+        , (10, pure CmdOldWalletToggleFirstKey)
+        , (10, CmdMimicPointerOutput <$> arbitrary)
```
KtorZ: 85+5+10+10 = 110. Not sure if you wanted to split that across 100 :sweat_smile: ...

KtorZ (commented):
That's some very nice testing. I like the approach and it gives a lot of confidence about what's being done. 

Some general remarks: the code is sometimes hard to follow. Many short identifiers where they could be a bit more explicit, and some functions / local bindings would benefits from type signatures to guide the reader.  It'd be very nice to iterate a bit on the code itself to try to make it crystal clear since the problem that it solves is already rather intricate.

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/DelegationSpec.hs
Anviking: Ah! Yes, I was suspicious about my spelling, looked it up with the macOS dictionary / force click, but missed that it said "No entries for X. Did you mean Y?" in addition to the definition, so I interpreted it as correct 😅

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
```diff
@@ -0,0 +1,464 @@
...
+-- [0, 1]
+--
+-- Note that for @s0@, we have no active stake keys, but we still want to use
+-- key 0 as part of addresses.
+usableKeys :: SoftDerivation k => DelegationState k -> [k 'AddressK XPub]
+usableKeys s = case lastActiveIx s of
+    Just i -> map (keyAtIx s) [minBound .. i]
+    Nothing -> [keyAtIx s minBound]
+
+-- | For testing. Returns all registered and delegating stake keys.
```
Anviking: Related to the topic of how key 0 should be handled:

This doc-comment is technically a lie. Key 0 might not be delegating.

1. This is misleading
2. I do hope the tests still pass if this might skip key 0 accordingly

Anviking: Added comments both here and for `usableKeys`, and the tests still pass if `activeKeys` skips key 0 accordingly.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
Anviking: > Then, some weeks later while she re-delegate to new pools, the first stake key gets re-registered without her asking anything.

`setPortfolioOf n` currently doesn't care about the pool-choice of each key. It couldn't just silently delegate key 0 without knowing what to delegate to (hadn't completely realised this though, so good that you brought it up). A more concrete scenario:

A user delegates to pools A, B, C with stake keys 0, 1 and 2 respectively.

Yoroi shows her wallet is delegating to A. She undelegates.

According to Future Daedalus she is… in some broken state?… delegating to B, C.

I imagine
- `setPortfolio [A,B,C]` would repair the state by re-registering and delegating key 0
- `setPortfolio [B,C]` would shift `{1 -> B, 2 -> C}` to `{ 0 -> C, 1 -> B }`, by de-registering key 2, and re-registering key 0
    - De-registering keys is a problem we need to solve in general. If she just recently de-registered key 0, the rewards she loses from key 2 might be offset by re-covering the rewards in key 0 though.

To me it seems much easier if we always use key 0 in addresses, some percent of the time, and treat a missing key 0, as a invalid, temporary state. 

When the user un-delegated using the old wallet, the change addresses used key 0. Same for all txs using the old wallet.

If both wallets can agree that key 0 should be used for _some_ addresses, there's less rebalancing/fighting to be made.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
Anviking: I tried adding
```haskell
nextState :: (Index 'Soft 'AddressK -> pointer) -> State pointer -> State pointer
prevState :: (Index 'Soft 'AddressK -> pointer) -> State pointer -> State pointer
```
to be able to advance the `State` with a `TxOut` as a pointer, rather than a `TxIn`, and also re-use them in `applyTx`.

But there's also the problem of:
1. Do we call `nextState` or `prevState`, and when do we stop?
2. How to yield certificates, which might need to be modified based on `isReg`.

It seemed like a `iterate`-approach would end up very complicated, potentially with many `foldl`, `scanl` and `take` steps. And boundary-cases of `takeWhile` seems just as complicated as `.. toEnum (n - 1)`.

So without an "aha" idea… I think being somewhat imperative here is fine… The tests should cover a lot of things, and give us confidence.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
Anviking: I went with

```haskell
-- | Is key 0 still registered? For compatibility with
-- single-stake-key wallets, we need to track this.
--
-- >>> activeKeys (More (toEnum 3) p ValidKey0)
-- [0, 1, 2]
--
-- >>> activeKeys (More (toEnum 3) p MissingKey0)
-- [1, 2]
--
-- (pseudocode; requires a bit more boilerplate to compile)
--
-- See the implementaiton of @applyTx@ for how it is used.
data Key0Status = ValidKey0 | MissingKey0
    deriving (Eq, Show, Generic)
```
ce5a3718e9607885b84c370badd75a0d93d4a6bc

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
Anviking: (added a comment)

KtorZ (approved):

Anviking:
I created https://jira.iohk.io/browse/ADP-938 for a future opportunity to revise the design and align with the CIP if needed.

Before merging, will
- [x] Look through it once more
- [x] Change the module name (not AddressDiscovery)
- [x] Squash

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Delegation.hs
Anviking: For the record, a argument in the other direction I didn't think of until recently:

Assuming
- re-delegation of existing stake-keys requires spending and re-creating the pointer
- the pointer uses `lastActiveIx`

Then we save a witness when changing the delegation at `lastActiveIx`.

PR #2637, opened 2021-05-03
Make `createMigrationPlan` accept a list of addresses

jonathanknowles:
# Issue Number

ADP-840

# Overview

This PR adjusts the following API operations to accept a non-empty list of addresses:

- `createByronWalletMigrationPlan`
- `createShelleyWalletMigrationPlan`

These operations need addresses in order to construct the coin selections required for a migration plan, as every coin selection output must have an address, and these can only be provided by the caller.

If the caller supplies **_n_** addresses, but the migration planner needs to create **_m_** outputs where **_m_** > **_n_**, then the server implementations of the above endpoints will reuse the given addresses in a cyclic fashion. (This was already the case in the old migration implementation.)


Anviking (approved):
👌

PR #2642, opened 2021-05-06
Add transaction signing API endpoint with stub impl

rvl:
### Issue Number

ADP-902

### Overview

- [x] Swagger spec updates for new transaction signing API.
- [x] New API types and servant stuff.
- [x] Rename some definitions to old so as to mark them as soft-deprecated.
- [x] Add stub implementation to server handler with `TODO` comments.

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -115,6 +115,7 @@ module Cardano.Wallet
     , assignChangeAddressesAndUpdateDb
     , assignChangeAddressesWithoutDbUpdate
     , selectionToUnsignedTx
+    , buildAndSignTransaction
     , signTransaction
     , constructTransaction
     , ErrSelectAssets(..)
@@ -363,6 +364,7 @@ import Cardano.Wallet.Primitive.Types.Tx
     ( Direction (..)
     , LocalTxSubmissionStatus
     , SealedTx (..)
+    , SerialisedTxParts (..)
     , TransactionInfo (..)
     , Tx
     , TxChange (..)
@@ -1538,13 +1540,49 @@ selectAssets ctx (utxo, cp, pending) tx outs transform = do
         hasWithdrawal :: Tx -> Bool
         hasWithdrawal = not . null . withdrawals
 
+signTransaction
+    :: forall ctx s k.
+        ( HasTransactionLayer k ctx
+        , HasDBLayer IO s k ctx
+        , HasNetworkLayer IO ctx
+        , IsOwned s k
+        )
+    => ctx
+    -> WalletId
+    -> ((k 'RootK XPrv, Passphrase "encryption") -> (XPrv, Passphrase "encryption"))
+       -- ^ Reward account derived from the root key (or somewhere else).
+    -> Passphrase "raw"
+    -> ByteString
+    -> ExceptT ErrSignPayment IO SerialisedTxParts
+signTransaction ctx wid mkRwdAcct pwd txBody = db & \DBLayer{..} -> do
+    era <- liftIO $ currentNodeEra nl
+    let _decoded = decodeSignedTx tl era txBody
+    withRootKey @_ @s ctx wid pwd ErrSignPaymentWithRootKey $ \xprv scheme -> do
+        let pwdP = preparePassphrase scheme pwd
+        mapExceptT atomically $ do
+            cp <- withExceptT ErrSignPaymentNoSuchWallet $ withNoSuchWallet wid $
+                readCheckpoint wid
+
+            -- TODO: ADP-919 implement this
+            let _keyFrom = isOwned (getState cp) (xprv, pwdP)
+            let _rewardAcnt = mkRwdAcct (xprv, pwdP)
+            -- withExceptT ErrSignPaymentMkTx $ ExceptT $ pure $
+            --     witnessTransaction tl rewardAcnt keyFrom txBody
+            let tx = mempty
+            pure $ SerialisedTxParts tx txBody []
+
+  where
+    db = ctx ^. dbLayer @IO @s @k
+    tl = ctx ^. transactionLayer @k
+    nl = ctx ^. networkLayer
+
 -- | Produce witnesses and construct a transaction from a given selection.
 --
 -- Requires the encryption passphrase in order to decrypt the root private key.
 -- Note that this doesn't broadcast the transaction to the network. In order to
 -- do so, use 'submitTx'.
 --
-signTransaction
+buildAndSignTransaction
     :: forall ctx s k.
         ( HasTransactionLayer k ctx
         , HasDBLayer IO s k ctx
@@ -1561,8 +1599,7 @@ signTransaction
     -> TransactionCtx
     -> SelectionResult TxOut
     -> ExceptT ErrSignPayment IO (Tx, TxMeta, UTCTime, SealedTx)
-signTransaction ctx wid mkRwdAcct pwd txCtx sel =
-    db & \DBLayer{..} -> do
+buildAndSignTransaction ctx wid mkRwdAcct pwd txCtx sel = db & \DBLayer{..} -> do
     era <- liftIO $ currentNodeEra nl
     withRootKey @_ @s ctx wid pwd ErrSignPaymentWithRootKey $ \xprv scheme -> do
         let pwdP = preparePassphrase scheme pwd
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -364,6 +364,7 @@ import Cardano.Wallet.Primitive.Types.Tx
     ( Direction (..)
     , LocalTxSubmissionStatus
     , SealedTx (..)
+    , SerialisedTx (..)
     , SerialisedTxParts (..)
     , TransactionInfo (..)
     , Tx
@@ -1635,7 +1636,7 @@ constructTransaction
     -> WalletId
     -> TransactionCtx
     -> SelectionResult TxOut
-    -> ExceptT ErrConstructTx IO ByteString
+    -> ExceptT ErrConstructTx IO SerialisedTx
 constructTransaction ctx wid txCtx sel =
     db & \DBLayer{..} -> do
     era <- liftIO $ currentNodeEra nl
--- a/lib/core/src/Cardano/Wallet/Api.hs
+++ b/lib/core/src/Cardano/Wallet/Api.hs
@@ -514,7 +514,7 @@ type SignTransaction n = "wallets"
     :> Capture "walletId" (ApiT WalletId)
     :> "transactions-sign"
     :> ReqBody '[JSON] PostSignTransactionData
-    :> PostAccepted '[JSON, OctetStream] ApiSignedTransaction
+    :> PostAccepted '[JSON] ApiSignedTransaction
 
 -- | https://input-output-hk.github.io/cardano-wallet/api/#operation/postTransaction
 type CreateTransactionOld n = "wallets"
@@ -828,7 +828,7 @@ type SignByronTransaction n = "byron-wallets"
     :> Capture "walletId" (ApiT WalletId)
     :> "transactions-sign"
     :> ReqBody '[JSON] PostSignTransactionData
-    :> PostAccepted '[JSON, OctetStream] ApiSignedTransaction
+    :> PostAccepted '[JSON] ApiSignedTransaction
 
 -- | https://input-output-hk.github.io/cardano-wallet/api/#operation/postByronTransaction
 type CreateByronTransactionOld n = "byron-wallets"
--- a/lib/core/src/Cardano/Wallet/Api.hs
+++ b/lib/core/src/Cardano/Wallet/Api.hs
@@ -49,7 +49,7 @@ module Cardano.Wallet.Api
     , CoinSelections
         , SelectCoins
 
-    , Transactions
+    , ShelleyTransactions
         , SignTransaction
         , CreateTransactionOld
         , PostTransactionFeeOld
@@ -185,6 +185,7 @@ import Cardano.Wallet.Api.Types
     , ApiSharedWallet
     , ApiSharedWalletPatchData
     , ApiSharedWalletPostData
+    , ApiSignTransactionPostData
     , ApiSignedTransaction
     , ApiStakeKeysT
     , ApiT
@@ -205,7 +206,6 @@ import Cardano.Wallet.Api.Types
     , Iso8601Time
     , KeyFormat
     , MinWithdrawal
-    , PostSignTransactionData
     , PostTransactionFeeOldDataT
     , PostTransactionOldDataT
     , SettingsPutData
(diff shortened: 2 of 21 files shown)


lib/core/src/Cardano/Wallet.hs
```diff
@@ -1437,11 +1438,46 @@ selectAssets ctx (utxo, cp, pending) tx outs transform = do
...
+        , HasDBLayer IO s k ctx
+        , HasNetworkLayer IO ctx
+        , IsOwned s k
+        )
+    => ctx
+    -> WalletId
+    -> ((k 'RootK XPrv, Passphrase "encryption") -> (XPrv, Passphrase "encryption"))
+       -- ^ Reward account derived from the root key (or somewhere else).
+    -> Passphrase "raw"
+    -> ByteString
```
paweljakubas, line 1556: Maybe introduce `newtype TxBody` rather than `ByteString` ?

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -453,17 +455,26 @@ type SelectCoins n = "wallets"
...
     :<|> DeleteTransaction
     :<|> GetTransaction n
 
+-- | https://input-output-hk.github.io/cardano-wallet/api/#operation/postSignTransaction
+type SignTransaction n = "wallets"
+    :> Capture "walletId" (ApiT WalletId)
+    :> "transactions"
+    :> "sign"
+    :> ReqBody '[JSON] PostSignTransactionData
+    :> PostAccepted '[JSON, OctetStream] ApiSerialisedTransaction
```
paweljakubas: we want both json and octet, not just octet?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -778,22 +779,27 @@ data ByronWalletPutPassphraseData = ByronWalletPutPassphraseData
     , newPassphrase :: !(ApiT (Passphrase "raw"))
     } deriving (Eq, Generic, Show)
 
-data PostTransactionData (n :: NetworkDiscriminant) = PostTransactionData
+data PostSignTransactionData = PostSignTransactionData
+    { txBody :: !ApiSerialisedTransaction -- TODO: ADP-902 or tx
+    , passphrase :: !(ApiT (Passphrase "lenient"))
```
paweljakubas, line 911: "lenient" here :thinking:

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -2072,9 +2078,19 @@ instance ToJSON (ApiT BoundType) where
 instance FromJSON (ApiT BoundType) where
     parseJSON = fmap ApiT . genericParseJSON defaultSumTypeOptions
 
-instance DecodeAddress t => FromJSON (PostTransactionData t) where
+instance FromJSON ApiSerialisedTransaction where
+    parseJSON = fmap ApiSerialisedTransaction . fromBaseText Base16
+instance ToJSON ApiSerialisedTransaction where
+    toJSON = toBaseText Base16 . view #payload
```
paweljakubas: very elegant I must say:-)

lib/core/src/Cardano/Wallet/Api.hs
paweljakubas: Is it connected with the fact that we want both binary (serviced by octet) and hex string (serviced by json)?

specifications/api/swagger.yaml
```diff
@@ -1879,6 +1895,28 @@ components:
...
+        - <<: *serializedHex
+          title: Serialized transaction body
+
+    x-txWits: &txWits
+      type: array
+      items: *serializedHex
+      title: Serialized transaction witnesses
+
+    ApiSerialisedTransaction: &ApiSerialisedTransaction
+      oneOf:
```
paweljakubas: ah, I got it now, why `[JSON, OctetStream]`

specifications/api/swagger.yaml
```diff
@@ -4324,13 +4407,36 @@ paths:
...
+        content:
+          application/json:
+            schema: *ApiPostSignTransactionData
+      responses: *responsesPostSignedTransaction
+
   /wallets/{walletId}/transactions:
     post:
       operationId: postTransaction
-      tags: ["Transactions"]
+      tags: ["Legacy Transactions"]
```
paweljakubas: what about waiting with `Transactions -> Legacy Transactions` and `stable -> deprecated` until the last ticket in this epic?

specifications/api/swagger.yaml
```diff
@@ -1879,6 +1895,28 @@ components:
...
+
+    ApiSerialisedTransaction: &ApiSerialisedTransaction
+      oneOf:
+        - <<: *serializedHex
+          title: Serialized transaction
+        - type: object
+          required:
+            - body
+            - wits
+          properties:
```
paweljakubas: How should I understand this? Either we get serialized hex which means all needed witnesses were constructed and added and the signed txs can be submitted OR I have partially signed tx, ie., not enough witnesses collected yet to treat it as fully signed tx?

paweljakubas:
@rvl we have now
1. postSignTransaction in which we have serialized tx body and password needed to sign. As a result we have blob that can be sent.
2. postSignedTransaction which submits the result of 1.

What about creating serialized tx body from outputs, delegations, ... It is in another ticket, right?

lib/core/src/Cardano/Wallet/Api/Types.hs
rvl: It means that there is no minimum length for the given passphrase.
We only want a minimum length when creating a new wallet or changing its password.

lib/core/src/Cardano/Wallet/Api.hs
rvl: The user can decide whether they want raw transaction bytes, or hex-encoded bytes inside a JSON object.

lib/core/src/Cardano/Wallet.hs
rvl: Yes it needs some type, but that will be added in the next PR. This function is a placeholder stub, so I hope that ByteString is good enough for the moment.

specifications/api/swagger.yaml
rvl: OK fair enough.

specifications/api/swagger.yaml
rvl: I added some more swagger doc strings.
The idea is that the user can submit an encoded tx, or an encoded tx body with zero or more witnesses.
The signing endpoint will add new witnesses to the set.

rvl:
Thanks for the review comments @paweljakubas .

You were probably slightly confused because I made an error with the return types.
PostSignedTransaction should return either a TxId or error, nothing else.

Constructing the TxBody is for another PR, ticket ADP-909.

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -453,17 +455,26 @@ type SelectCoins n = "wallets"
...
-    :<|> PostTransactionFee n
+    :<|> PostTransactionFeeOld n
     :<|> DeleteTransaction
     :<|> GetTransaction n
 
+-- | https://input-output-hk.github.io/cardano-wallet/api/#operation/postSignTransaction
+type SignTransaction n = "wallets"
+    :> Capture "walletId" (ApiT WalletId)
+    :> "transactions"
+    :> "sign"
```
KtorZ: What about `signatures` to be a bit more rest-ish. Verbs / action should ideally be captured by the HTTP method.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -778,22 +779,27 @@ data ByronWalletPutPassphraseData = ByronWalletPutPassphraseData
     , newPassphrase :: !(ApiT (Passphrase "raw"))
     } deriving (Eq, Generic, Show)
 
-data PostTransactionData (n :: NetworkDiscriminant) = PostTransactionData
+data PostSignTransactionData = PostSignTransactionData
+    { tx :: !ApiSerialisedTransaction -- TODO: ADP-902 or tx
```
KtorZ: We've avoided shorten names and acronym in the API so far. I'd suggest to go for "transaction" or "body" since this is really just the transaction body without witnesses in principle.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -2072,9 +2078,19 @@ instance ToJSON (ApiT BoundType) where
 instance FromJSON (ApiT BoundType) where
     parseJSON = fmap ApiT . genericParseJSON defaultSumTypeOptions
 
-instance DecodeAddress t => FromJSON (PostTransactionData t) where
+instance FromJSON ApiSerialisedTransaction where
+    parseJSON = fmap ApiSerialisedTransaction . fromBaseText Base16
```
KtorZ: Wouldn't Base64 be a better fit here? We typically use Base16 for short-ish strings and base64 for longer payloads.

lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
```diff
@@ -1156,12 +1157,37 @@ instance Malformed (BodyParam (ApiSelectCoinsData ('Testnet pm))) where
...
+            [ ("1020344", "Error in $: parsing Cardano.Wallet.Api.Types.PostSignTransactionData failed")
+            , ("\"hello\"", "Error in $: parsing Cardano.Wallet.Api.Types.PostSignTransactionData failed, expected Object, but encountered String")
+            , ("{\"txBody: \"\", \"random\"}", msgJsonInvalid)
+            , ("{\"txBody: \"lah\", \"passphase\": \"Secure Passphrase\"}", "fixme")
+            , ("{\"txBody: 1020344, \"passphase\": \"Secure Passphrase\"}", "fixme")
+            ]
+         jsonValid = first (BodyParam . Aeson.encode) <$> paymentCases ++
+            [ -- passphrase
+              ( [aesonQQ|
+                { "tx_body": "cafecafe"
```
KtorZ: I didn't catch where the code from Api/Types could generate a `tx_body` field :thinking:  ?

lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
```diff
@@ -451,34 +451,52 @@ instance
...
+    type MkHeaderRequest (Verb m s '[] a) = [Request]
+    gEveryHeader _ req =
+        [req { requestMethod = reflectMethod (Proxy @m) }]
+
 instance
     ( ReflectMethod m
     , Accept ct
-    ) => GEveryEndpoints (Verb (m :: StdMethod) s '[ct] a)
+    , GEveryEndpoints (Verb m s cts a)
+    ) => GEveryEndpoints (Verb (m :: StdMethod) s (ct ': cts) a)
```
KtorZ: :+1:

lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
rvl: Yes I changed the types and forgot to update these tests.

lib/core/src/Cardano/Wallet/Api/Types.hs
rvl: I don't think it matters that much. I thought hex might be slightly easier to use. Base64 is 1.33× raw size and hex is 2× raw size - both acceptable for transactions given their max size. Anyway I have changed it to Base64 because the sign metadata endpoint uses Base64, as you say. I also changed the encoding accepted by the `cardano-wallet transaction submit` CLI from hex to base64 to align with this.

lib/core/src/Cardano/Wallet/Api/Types.hs
rvl: Yep good idea, fixed.

lib/core/src/Cardano/Wallet/Api.hs
rvl: I'm afraid that REST is a poor fit for this endpoint (perhaps in general too, but that's a separate discussion).

The resource is the wallet (specifically the keys contained in the wallet) and action is to sign a blob of data (a TxBody), given a spending password, and return the blob of data plus its signatures (TxBody + TxWits).

None of the HTTPs methods adequately describe this signing operation. And no request path name can make a verb look like a noun.

So I think the least worst name is something short and simple: `POST /v2/wallet/<id>/transactions/sign`.

Slightly related, I think the corresponding servant handler function name I have chosen is thoroughly confusing: `postSignTransaction`. What do you think of renaming that to `signTransactionHandler` ?

KtorZ: > (perhaps in general too, but that's a separate discussion).

I couldn't agree more. 

> None of the HTTPs methods adequately describe this signing operation. And no request path name can make a verb look like a noun.

I see it differently :thinking: , to me `POST /wallets/{walletId}/witnesses`  for instance would read just fine. This creates a `witness` resource from a given payload on the given `wallet`. So it's kinda rest-ish.  

> Slightly related, I think the corresponding servant handler function name I have chosen is thoroughly confusing: postSignTransaction. What do you think of renaming that to signTransactionHandler ?

Yeah... I always have mixed feelings when naming handlers. One the one hand, keeping some form of consistent naming from the endpoint structure makes it "brain dead" and easy to name / figure out. But on the other hand, it's often ugly and, we have deviated so many times from it that we have lost any benefits from consistent naming. This being said, I like your suggestion :blush:

paweljakubas (approved):
lgtm, the rest bits can come in next PRs

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
```diff
@@ -398,12 +400,27 @@ instance Buildable a => Buildable (WithDirection a) where
...
+-- | A serialised transaction that may be only partially signed, or even
+-- invalid.
+newtype SerialisedTx = SerialisedTx { payload :: ByteString }
+    deriving stock (Show, Eq, Generic, Ord)
+    deriving newtype (Semigroup, Monoid, ByteArray, ByteArrayAccess, NFData)
+
+-- | @SerialisedTxParts@ is a serialised transaction body, and a possibly
+-- incomplete set of serialised witnesses, along with an encoding of the
+-- combined transaction.
+data SerialisedTxParts = SerialisedTxParts
```
Anviking, line 418: Why is `SerialisedTxParts` needed when we have `SerialisedTx`?

Isn't the structure it exposes already unambiguously exposed in the raw CBOR?

rvl: Yes, but I think it will be helpful for API consumers to get both.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1538,13 +1541,49 @@ selectAssets ctx (utxo, cp, pending) tx outs transform = do
...
+signTransaction
+    :: forall ctx s k.
+        ( HasTransactionLayer k ctx
+        , HasDBLayer IO s k ctx
+        , HasNetworkLayer IO ctx
+        , IsOwned s k
+        )
+    => ctx
+    -> WalletId
+    -> ((k 'RootK XPrv, Passphrase "encryption") -> (XPrv, Passphrase "encryption"))
```
paweljakubas, line 1553: well, as I did in construct tx PR, one can argue we can read public reward account key from wallet state, we need this structure only when private reward account key is needed...which probably is the case here

paweljakubas (approved):
much work is needed to implement signing and prove it works which is to be done in coming PRs. API seems reasonable, some solid/beautifully-conveying-idea-under-the-hood  data structures were introduced

PR #2643, opened 2021-05-07
Remove icarus genesis gap hack

Anviking:
# Issue Number

ADP-828

# Overview

- [x] Use the standard gap size, even for the genesis block.
- [x] Fund icarus fixture wallets using txs, not as part of the byron genesis
    - Not entirely sure why this helped though 🤔 - the wallets only had 10 addresses, so the order shouldn't have mattered anyway...

# Comments


Anviking:
bors try

KtorZ (approved):
Yes! I am pretty sure there's also a special Icarus wallet with addresses starting only at a high index (like 500 or so) which would now fail because of this. 

I think it's fine to get rid of this scenario now, since it was mainly for the ITN snapshot. @piotr-iohk, thoughts?

Anviking:
bors try

Anviking:
bors try-

Anviking:
> I am pretty sure there's also a special Icarus wallet with addresses starting only at a high index (like 500 or so) which would now fail because of this.

👍 

> @piotr-iohk, thoughts?

I think Piotr is off until something like Wednesday. But I can't either see why the test would matter now, and he'll have the opportunity to object in the QA phase of the ticket too, so merging.

bors r+

paweljakubas (approved):
lgmt

PR #2644, opened 2021-05-10
Connect API with migration algorithm

jonathanknowles:
# Issue Number

ADP-840

# Overview

This PR:

- [x] Implements `Api.Server.createWalletMigrationPlan`.
- [x] Implements `Api.Server.migrateWallet`.
- [x] Adjusts `migrateWallet` to require a non-empty list of addresses.
- [x] Adjusts the success response types for all migration endpoints to be `202` `ACCEPTED`.
- [x] Resurrects all previously-disabled integration tests (for ada-specific wallets).
- [x] Removes the "disabled" warnings on migration endpoints.
- [x] Adds integration test coverage for MA wallets.

# QA Due Diligence

After rewriting the integration test suite, I ran the entire integration test suite 500 times in an effort to increase confidence that there are no flaky tests. The test suite passed 500 times, with no failures or errors.

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
@@ -208,6 +208,30 @@ spec = describe "BYRON_MIGRATIONS" $ do
                     (errMsg403NothingToMigrate $ sourceWallet ^. walletId)
                 ]
 
+    it "BYRON_CREATE_MIGRATION_PLAN_05 - \
+        \Creating a plan is deterministic."
+        $ \ctx -> forM_ [fixtureRandomWallet, fixtureIcarusWallet]
+        $ \fixtureByronWallet -> runResourceT $ do
+            sourceWallet <- fixtureByronWallet ctx
+            targetWallet <- emptyWallet ctx
+            targetAddresses <- listAddresses @n ctx targetWallet
+            let targetAddressIds = targetAddresses <&>
+                    (\(ApiTypes.ApiAddress addrId _ _) -> addrId)
+            let ep = Link.createMigrationPlan @'Byron sourceWallet
+            response1 <- request @(ApiWalletMigrationPlan n) ctx ep Default
+                (Json [json|{addresses: #{targetAddressIds}}|])
+            response2 <- request @(ApiWalletMigrationPlan n) ctx ep Default
+                (Json [json|{addresses: #{targetAddressIds}}|])
+            expectResponseCode HTTP.status202 response1
+            expectResponseCode HTTP.status202 response2
+            expectField (#selections) ((.> 0) . length) response1
+            expectField (#selections) ((.> 0) . length) response2
+            case (snd response1, snd response2) of
+                (Right plan1, Right plan2) ->
+                    liftIO $ plan1 `shouldBe` plan2
+                _ ->
+                    error "Unable to compare plans."
+
     describe "BYRON_MIGRATE_05 - I could migrate to any valid address" $ do
         forM_ [ ("Byron", emptyRandomWallet)
               , ("Icarus", emptyIcarusWallet)
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
@@ -142,17 +142,22 @@ spec = describe "BYRON_MIGRATIONS" $ do
                     (`shouldBe` 0)
                 ]
 
-    it "BYRON_CALCULATE_02 - \
-        \Cannot calculate fee for empty wallet."
+    it "BYRON_CREATE_MIGRATION_PLAN_02 - \
+        \Cannot create plan for empty wallet."
         $ \ctx -> forM_ [emptyRandomWallet, emptyIcarusWallet]
         $ \emptyByronWallet -> runResourceT $ do
-            liftIO $ pendingWith "Migration endpoints temporarily disabled."
-            w <- emptyByronWallet ctx
-            let ep = Link.createMigrationPlan @'Byron w
-            r <- request @(ApiWalletMigrationPlan n) ctx ep Default Empty
-            verify r
+            sourceWallet <- emptyByronWallet ctx
+            targetWallet <- emptyWallet ctx
+            targetAddresses <- listAddresses @n ctx targetWallet
+            let targetAddressIds = targetAddresses <&>
+                    (\(ApiTypes.ApiAddress addrId _ _) -> addrId)
+            let ep = Link.createMigrationPlan @'Byron sourceWallet
+            response <- request @(ApiWalletMigrationPlan n) ctx ep Default
+                (Json [json|{addresses: #{targetAddressIds}}|])
+            verify response
                 [ expectResponseCode HTTP.status403
-                , expectErrorMessage (errMsg403NothingToMigrate $ w ^. walletId)
+                , expectErrorMessage
+                    (errMsg403NothingToMigrate $ sourceWallet ^. walletId)
                 ]
 
     it "BYRON_CALCULATE_02 - \
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
@@ -232,6 +232,19 @@ spec = describe "BYRON_MIGRATIONS" $ do
                 _ ->
                     error "Unable to compare plans."
 
+    describe "BYRON_MIGRATE_01 - \
+        \After a migration operation successfully completes, the correct \
+        \amounts eventually become available in the target wallet for an \
+        \arbitrary number of specified addresses, and the balance of the \
+        \source wallet is completely depleted."
+        $ do
+            testAddressCycling "Random" fixtureRandomWallet  1
+            testAddressCycling "Random" fixtureRandomWallet  3
+            testAddressCycling "Random" fixtureRandomWallet 10
+            testAddressCycling "Icarus" fixtureIcarusWallet  1
+            testAddressCycling "Icarus" fixtureIcarusWallet  3
+            testAddressCycling "Icarus" fixtureIcarusWallet 10
+
     describe "BYRON_MIGRATE_05 - I could migrate to any valid address" $ do
         forM_ [ ("Byron", emptyRandomWallet)
               , ("Icarus", emptyIcarusWallet)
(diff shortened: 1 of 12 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
```diff
@@ -193,7 +220,7 @@ spec = describe "SHELLEY_MIGRATIONS" $ do
               testAddressCycling 3
               testAddressCycling 10
 
-    Hspec.it "SHELLEY_MIGRATE_01_big_wallet - \
```
Anviking, line 167: 🙅‍♂️ I would not. `Hspec.it` doesn't retry on failures. `it` does retry, but the funds of the hard-coded mnemonic can only be spent once, most likely making it pointless.

Anviking (commented):
Will look closer later, but looks good 👍

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
Anviking: _hoping we can remove the retrying `it` soon_

jonathanknowles: That's interesting! I had no idea that this difference even existed. Thanks for pointing it out.

I'm just thinking how we could make this difference clearer, so that the name of the function makes it obvious whether or not the contents will be retried on failure.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1427,46 +1458,45 @@ selectAssets ctx (utxo, cp, pending) tx outs transform = do
...
+    db & \DBLayer{..} -> do
     era <- liftIO $ currentNodeEra nl
     withRootKey @_ @s ctx wid pwd ErrSignPaymentWithRootKey $ \xprv scheme -> do
         let pwdP = preparePassphrase scheme pwd
         mapExceptT atomically $ do
-            cp <- withExceptT ErrSignPaymentNoSuchWallet $ withNoSuchWallet wid $
-                readCheckpoint wid
+            cp <- withExceptT ErrSignPaymentNoSuchWallet
+                $ withNoSuchWallet wid
+                $ readCheckpoint wid
```
rvl, line 1497: Er, what's wrong with the other indentation?

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2310,6 +2413,11 @@ data ErrStartTimeLaterThanEndTime = ErrStartTimeLaterThanEndTime
     , errEndTime :: UTCTime
     } deriving (Show, Eq)
 
+data ErrCreateMigrationPlan
+    = ErrCreateMigrationPlanEmpty
```
rvl, line 2419: Couldn't it just return and empty list of transactions? That seems fairly obvious and easy to handle for users.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1992,26 +2007,141 @@ quitStakePool ctx (ApiT wid) body = do
...
+
+mkApiWalletMigrationPlan
+    :: forall n s. IsOurs s Address
+    => s
+    -> NonEmpty (ApiT Address, Proxy n)
+    -> Withdrawal
+    -> MigrationPlan
+    -> Maybe (ApiWalletMigrationPlan n)
+mkApiWalletMigrationPlan s addresses rewardWithdrawal plan
+    | Just selections <- maybeSelections =
```
rvl: ```case maybeSelections of```

rvl (commented):
This seems pretty good so far :+1:. You could also update the migration endpoint status info in `swagger.yaml`. Unnecessary whitespace changes make it harder to provide a good review.

lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
```diff
@@ -115,66 +117,259 @@ spec :: forall n.
...
+
+        -- Compute the expected migration plan:
+        responsePlan <- request @(ApiWalletMigrationPlan n) ctx
+            (Link.createMigrationPlan @'Byron sourceWallet) Default
+            (Json [json|{addresses: #{targetAddressIds}}|])
+        verify responsePlan
+            [ expectResponseCode HTTP.status202
+            , expectField
+                (#totalFee . #getQuantity)
+                (`shouldBe` 2_460_400)
```
rvl, line 299: In integration tests, exact values like this may cause us pain in future.
It might be possible to make the test case more forgiving of changes, without reducing its effectiveness.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
jonathanknowles: Fixed in d4f6fd647456f10c81b993635d71c597a2ea440b.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -3207,6 +3336,13 @@ instance IsServerError ErrOutputTokenQuantityExceedsLimit where
         , "."
         ]
 
+instance IsServerError ErrCreateMigrationPlan where
+    toServerError = \case
+        ErrCreateMigrationPlanEmpty ->
+            -- TODO: Provide a more useful error message:
+            apiError err403 NothingToMigrate "Nothing to migrate"
```
jonathanknowles: We should provide a more detailed error message here.

At the very least, it needs to explain the types of circumstances that might lead to this error, so the user can take action.

lib/core/src/Cardano/Wallet/Api/Server.hs
jonathanknowles: Ah yes, the pattern guard is a bit unnecessary here!

(I think it might be a hangover from a previous version of the function.)

I've fixed this in a671d6eb69202f89285f87cad0c034df9a9bf7bf.

lib/core/src/Cardano/Wallet.hs
jonathanknowles: You're right, it could.

However, the `NothingToMigrate` condition was already a specific error in the old `migrateWallet` API. This PR just reuses the same approach for `createMigrationPlan`.

It covers two cases:
- The wallet is genuinely empty (therefore there really is nothing available to migrate).
- The wallet isn't empty, but the migration algorithm wasn't able to pay for any of the UTxO entries to be migrated, and therefore the plan is empty.

For the second case, I think it's useful to have a specific error here, as we can provide a useful hint to the user:

> I wasn't able to construct a migration plan. This could be because your wallet is empty, or it could be because the amount of ada in your wallet is insufficient to pay for any of the funds to be migrated. Try adding some ada to your wallet before trying again."

For comparison, the **_original_** message was:

>  I can't migrate the wallet with the given ID, because it's either empty or full of small coins which wouldn't be worth migrating."

lib/core/src/Cardano/Wallet/Api/Server.hs
jonathanknowles: Fixed in 713990db02026c286437a7061b275ba41fc72a3e.

jonathanknowles:
bors try

lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
jonathanknowles: Hi @rvl. I think I understand your concerns here.

However, I think there are good arguments in favour of using exact values:

- The new migration algorithm is completely deterministic: the plan is fully determined by the current UTxO, with no randomness. So if we start with the same fixture wallet and transaction constraints, we should always end up with precisely the same plan.

- To increase confidence that there are no flaky tests, I've run the entire migration integration test suite 500 times and confirmed that it always passes.

- Stating exact values means we can increase our confidence that the API is producing plans with reasonable fees, since fees are explicitly encoded as literals in the test code. I think it's good to be a bit paranoid here, as the potential cost to a user of miscalculating a fee is that they spend more ada than required (potentially without noticing).

- If someone alters the fixture wallet definitions or the fee calculation algorithm, then it will cause the tests to fail. However, they're likely to fail in an extremely obvious way, and it's likely that we can fix them very quickly just by entering new numbers. It also gives a fresh chance for a human pair of eyes to confirm that fees produced by the API are reasonable.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
```diff
@@ -114,330 +128,595 @@ spec :: forall n.
...
+                [ expectResponseCode HTTP.status202
                 , expectField (#totalFee . #getQuantity)
-                    (.> 0)
+                    (`shouldBe` 255_200)
+                , expectField (#selections)
+                    ((`shouldBe` 1) . length)
+                , expectField (#balanceSelected . #ada . #getQuantity)
+                    (`shouldBe` 1_000_000_000_000)
+                , expectField (#balanceLeftover . #ada . #getQuantity)
+                    (`shouldBe` 0)
```
Anviking, line 152: 👍

lib/core/src/Cardano/Wallet.hs
```diff
@@ -17,6 +17,7 @@
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
+{-# OPTIONS_GHC -fno-warn-redundant-constraints #-}
```
Anviking: ☝️ Guessing your HLS quick-fix added this by mistake

lib/core/src/Cardano/Wallet.hs
Anviking: Arguably it might want to be worth including the case-specific `balanceLeftover` in the 403 case. 

Could we have a err403 *but* returning `ApiWalletMigrationPlan` with `selections = []`? It couldn't include an error message though, and perhaps a bit weird, so probably not.

Anviking (approved):
👍

lib/core/src/Cardano/Wallet.hs
Anviking: Or it was intentional?
Skärmavbild 2021-05-14 kl  18 12 18

Could we get by with a `type` alias instead, not to have to disable the warning for the entire module?

jonathanknowles: @Anviking Thanks, this is a good suggestion. 👍🏻 

I also had reservations about disabling this warning. It's a shame that it's not possible to disable these warnings for just a single function. (Or is it?) I notice that we disable this warning in other modules -- perhaps we shouldn't do that.

I agree with you that it's just simpler to use a type synonym here.

I've fixed this in b9136673f1f69610b331a37f9dd50004d85571c7.

lib/core/src/Cardano/Wallet.hs
jonathanknowles: > Could we have a err403 but returning ApiWalletMigrationPlan with selections = []? It couldn't include an error message though, and perhaps a bit weird, so probably not.

In that specific case, the `balanceLeftover` would be precisely the same as the wallet's balance. So perhaps it wouldn't be providing any more info than the user already has. (Presumably, they already know their balance, or can easily check it.)

PR #2646, opened 2021-05-10
Fix slow wallet deletion and db issues

Anviking:
# Issue Number

ADP-827

# Overview

- [x] Don't retry on SQLITE_BUSY when trying to close

# Comments

- The retrying would timeout anyway, leaving the connection open and the file un-deleted.

## Alternative attempts / notes
- `.deleted` lock and removing `waitForFree` lead to err500 on wallet creation.
- (On master) it seemed to consistently (100% of the time) get stuck in destroySqliteBackend trying to close, _when_ a db query was cancelled.

## Effects on tests

With `-j 8`
- 240s
- 0 >10s DELETE or POST /wallet requests

### DIFF
--- a/lib/core/src/Cardano/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/DB/Sqlite.hs
@@ -236,11 +236,13 @@ destroySqliteBackend
     -> IO ()
 destroySqliteBackend tr sqlBackend dbFile = do
     traceWith tr (MsgCloseSingleConnection dbFile)
-    retryOnBusy tr (close' sqlBackend)
+    (close' sqlBackend)
         & handleIf isAlreadyClosed
             (traceWith tr . MsgIsAlreadyClosed . showT)
         & handleIf statementAlreadyFinalized
             (traceWith tr . MsgStatementAlreadyFinalized . showT)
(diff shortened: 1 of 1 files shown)


lib/core/src/Cardano/DB/Sqlite.hs
```diff
@@ -236,25 +238,6 @@ destroySqliteBackend
     -> IO ()
 destroySqliteBackend tr sqlBackend dbFile = do
```
Anviking, line 237: Function is now misleading. Should remove if we go with this solution.

rvl (commented):
Wouldn't this prevent sqlite db connections from ever being closed, unless there is an error with a query? That would be a resource leak.

Anviking:
Ah! Yes. 😐

Anviking:
New approach — just remove the retries to `close'` on SQLITE_BUSY:

I compared it with `master`, and they seem similar in the amount of wallets not being cleaned up:

[ADP-827.zip](https://github.com/input-output-hk/cardano-wallet/files/6464903/ADP-827.zip)

Anviking:
Closing for now

PR #2647, opened 2021-05-11
Add offchain-metadata-tools as dependency

sevanspowell:
# Issue Number

ADP-916

# Overview

**DRAFT** for CI purposes only.

- [ ] I have added offchain-metadata-tools as dependency to get access to metadata types, ultimately to prevent duplication.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -15,6 +15,12 @@ packages:
 
 source-repository-package
     type: git
+    location: https://github.com/input-output-hk/offchain-metadata-tools
+    tag: 65f2d445d51819bce6537b85e964081b146fd1e3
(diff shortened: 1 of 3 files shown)


cabal.project
```diff
@@ -137,6 +143,7 @@ allow-newer:
   , hjsonpointer:*
   , cardano-sl-x509:ip
   , cardano-addresses:aeson
+  , github:base16-bytestring
```
rvl, line 146: This is an example of a dependency that we don't want dragged in.

rvl (commented):
Cool. Don't forget `stack.yaml`, which is our primary source for build dependencies.

PR #2650, opened 2021-05-13
List addresses and wallets for shared wallet

paweljakubas:
# Issue Number

adp-686

# Overview

- [x] Extension to WalletStyle and Discrimination in Api.Link
- [x] Adding listSharedWallets with integration tests
- [x] Resign from getSharedWallet in Api.Link in favour of getWallet
- [x] Resign from deleteSharedWallet in Api.Link in favour of deleteWallet
- [x] Resign from postSharedWallet in Api.Link in favour of postWallet
- [x] add `normalizeSharedAddress`
- [x] add `KnownAddresses`  instance for `SharedState`
- [x] add `CompareDiscovery`  instance for `SharedState`
- [x] add listAddresses to shared wallet
- [x] add integration testing for listSharedAddresses
- [x] change `AddressDiscovery.SharedState` to `AddressDiscovery.Shared` 
- [x] enable UtxOInternal for ApiVerificationKeyShared

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -93,6 +93,7 @@ module Cardano.Wallet
     , updateCosigner
     , ErrAddCosignerKey (..)
     , ErrConstructSharedWallet (..)
+    , normalizeSharedAddress
 
     -- ** Address
     , createChangeAddress
@@ -262,17 +263,19 @@ import Cardano.Wallet.Primitive.AddressDiscovery
 import Cardano.Wallet.Primitive.AddressDiscovery.Random
     ( ErrImportAddress (..), RndStateLike )
 import Cardano.Wallet.Primitive.AddressDiscovery.Sequential
-    ( SeqState
+    ( ParentContext (..)
+    , SeqState
     , defaultAddressPoolGap
-    , derivationPrefix
     , mkSeqStateFromRootXPrv
     , purposeBIP44
     )
 import Cardano.Wallet.Primitive.AddressDiscovery.SharedState
     ( CredentialType (..)
     , ErrAddCosigner (..)
-    , SharedState
+    , SharedState (..)
+    , SharedStateFields (..)
     , addCosignerAccXPub
+    , isShared
     )
 import Cardano.Wallet.Primitive.CoinSelection.MA.RoundRobin
     ( SelectionError (..)
@@ -433,7 +436,7 @@ import Data.List
 import Data.List.NonEmpty
     ( NonEmpty (..) )
 import Data.Maybe
-    ( fromMaybe, mapMaybe )
+    ( fromJust, fromMaybe, mapMaybe )
 import Data.Proxy
     ( Proxy )
 import Data.Quantity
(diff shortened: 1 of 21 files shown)


lib/core/src/Cardano/Wallet.hs
```diff
@@ -2311,6 +2315,34 @@ updateCosigner ctx wid accXPub cosigner cred = db & \DBLayer{..} -> do
...
+    -> Maybe Address
+normalizeSharedAddress s@(SharedState _ state') addr = case state' of
+    PendingFields _ -> Nothing
+    ReadyFields pool -> do
+        let (ParentContextShared _ _ dTM) = Seq.context pool
+        fingerprint <- eitherToMaybe (paymentKeyFingerprint @k addr)
+        let (ixM, _) = isShared addr s
+        case dTM of
+            Just dT ->
+                pure $ Shared.liftDelegationAddress @n (fromJust ixM) dT fingerprint
```
KtorZ: `fromJust` :grimacing: ? Can we do something a bit better here? There's no guarantee from the type signature that the given address belongs to the wallet and that `isShared` would resolve.

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -448,6 +455,36 @@ instance GetAccount (SharedState n k) k where
...
         in accXPub
 
+instance
+    ( MkKeyFingerprint k (Proxy n, k 'AddressK XPub)
+    , MkKeyFingerprint k Address
+    , SoftDerivation k
+    , Typeable n
+    ) => CompareDiscovery (SharedState n k) where
+    compareDiscovery (SharedState _ state) a1 a2 = case state of
+        PendingFields _ -> error "comparing addresses in pending shared state does not make sense"
```
KtorZ, line 465: What about only defining the instance on `AddressPool 'UtxoExternal SharedKey` then to force the caller to provide the right item?

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -448,6 +455,36 @@ instance GetAccount (SharedState n k) k where
...
+        ix :: Address -> AddressPool 'UtxoExternal k -> Maybe (Index 'Soft 'AddressK)
+        ix a = fst . lookupAddress @n id a
+
+instance
+    ( Typeable n
+    , GetPurpose k
+    ) => KnownAddresses (SharedState n k) where
+    knownAddresses (SharedState _ s) = nonChangeAddresses
+      where
+          -- TO_DO - After enabling txs for shared wallets we will need to expand this
```
KtorZ: ~~TO_DO~~ -> TODO (the TODO and FIXME tags are actually picked up by the CI!)

KtorZ (approved):
:+1: modulo some changes regarding safety and wild runtime errors :grimacing:

lib/core/src/Cardano/Wallet.hs
paweljakubas: indeed, corrected

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
paweljakubas: experimented with it a little bit and IMHO it adds to code complexity in the end. But, I added integration test for listing addresses in pending state to show it is not crashing. it just returns `[]` which is reasonable in my opinion

PR #2652, opened 2021-05-14
Test migration of SeaHorses failing due to insufficient ada

Anviking:
# Issue Number

#2644

# Overview

- [x] Remove `pendingWith` to enable `SHELLEY_CREATE_MIGRATION_PLAN_04 - Cannot create a plan for a wallet that only contains dust.`
- [x] Fund the source wallet with SeaHorses and as small amount of ada to make the migration fail 

# Comments

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
@@ -68,7 +68,7 @@ import Data.Word
 import Numeric.Natural
     ( Natural )
 import Test.Hspec
-    ( SpecWith, describe, pendingWith )
+    ( SpecWith, describe )
 import Test.Hspec.Expectations.Lifted
     ( shouldBe, shouldSatisfy )
 import Test.Hspec.Extra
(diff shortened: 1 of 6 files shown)


Anviking:
@jonathanknowles I saw the `pendingWith` in #2644 and got curious if this would work. Feel free to take over, or close if you were to see more problems than value.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
```diff
@@ -67,6 +67,7 @@ import Data.Word
     ( Word64 )
 import Numeric.Natural
     ( Natural )
+import System.Environment
```
Anviking:
```suggestion
```

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
```diff
@@ -196,18 +197,21 @@ spec = describe "SHELLEY_MIGRATIONS" $ do
                         (errMsg404NoWallet $ sourceWallet ^. walletId)
                     ]
 
+
     it "SHELLEY_CREATE_MIGRATION_PLAN_04 - \
         \Cannot create a plan for a wallet that only contains dust."
```
Anviking: I guess with this PR this might technically not be a dust wallet?, but it should at least be a wallet of `Freerider`s

Anviking:
Also: maybe you don't even need the tokens for a test like this? Just the right amount of ada?

jonathanknowles:
> @jonathanknowles I saw the `pendingWith` in #2644 and got curious if this would work. Feel free to take over, or close if you were to see more problems than value.

Hi @Anviking. Thanks for creating this!

I agree with you that it would be ideal to have a test with a real dust wallet, or a wallet that cannot be fully migrated because it's dominated by freerider entries.

An interesting case, perhaps worth testing, would be where the UTxO has too many entries to fit into a single transaction, but the migration algorithm cannot construct a second transaction because all the non-freerider entries are used up in the first transaction. That would allow us to have coverage for a partial migration.

Perhaps we can have a call about it on Monday?

By the way, have a nice weekend!

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
jonathanknowles: For a wallet of just ada values, we **_could_** define "dust" as an ada value that can't be an input to a singleton transaction.

Each value would need to be less than **_output minimum ada quantity_** + **_base cost of tx_** + **_marginal cost of tx input_**.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
```diff
@@ -196,18 +197,21 @@ spec = describe "SHELLEY_MIGRATIONS" $ do
...
-                    "style": "random"
-                    } |]
-            sourceWallet <- unsafeResponse <$> postWallet ctx payloadRestore
+            sourceWallet <- emptyWallet ctx
+            srcAddrs <- map (getApiT . fst . view #id)
+                <$> listAddresses @n ctx sourceWallet
+
+            -- Big enough to mint and sent to sourceWallet, but small enough
+            -- that the migration to targetWallet then fails.
+            let edgeCase = Coin 3_300_000
```
jonathanknowles: Note to self: this is the amount of ada per input.

Perhaps mention that this is above the minimum ada amount.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
```diff
@@ -196,18 +197,21 @@ spec = describe "SHELLEY_MIGRATIONS" $ do
...
+
+            liftIO $ _mintSeaHorseAssets ctx 10 edgeCase srcAddrs
+            waitForTxImmutability ctx
+
```
jonathanknowles, line 217: Suggestion: add a check to see that the source balance is as we expect it (i.e., that it is greater than zero, and the set of assets in the balance is not the empty set).

jonathanknowles (approved):
LGTM!

If you have time, it would be great to add checks that we have a non-zero balance in:

- `SHELLEY_CREATE_MIGRATION_PLAN_04`
- `BYRON_CREATE_MIGRATION_PLAN_04`

Of course, these are slightly different tests, but they are designed to drive the same condition, which is a non-empty wallet cannot be migrated, even partly.

Anviking:
@jonathanknowles I tried adding a similar balance check to the byron version, but it failed 🤔 
```patch
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Byron/Migrations.hs
@@ -202,6 +202,15 @@ spec = describe "BYRON_MIGRATIONS" $ do
             let ep = Link.createMigrationPlan @'Byron sourceWallet
             response <- request @(ApiWalletMigrationPlan n) ctx ep Default
                 (Json [json|{addresses: #{targetAddressIds}}|])
+
+            eventually "dust wallet has funds" $
+                request @ApiWallet ctx
+                    (Link.getWallet @'Shelley sourceWallet) Default Empty
+                    >>= flip verify
+                    [ expectField (#balance . #available . #getQuantity)
+                        (.> 0)
+                    ]
+
             verify response
                 [ expectResponseCode HTTP.status403
                 , expectErrorMessage
```
🤦‍♂️ -> wait I see why

PR #2654, opened 2021-05-17
Add property tests for `migrationPlanToSelectionWithdrawals`.

jonathanknowles:
# Issue Number

ADP-840

# Overview

This PR:
- [x] provides missing test coverage for the `migrationPlanToSelectionWithdrawals` function.
- [x] fixes the order in which target addresses are assigned to outputs.

# Background

Before this PR, the order in which target addresses were assigned to outputs was not cyclical, as expected:

target-address-cycling

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -1853,9 +1853,10 @@ migrationPlanToSelectionWithdrawals
     -> Maybe (NonEmpty (SelectionResultWithoutChange, Withdrawal))
 migrationPlanToSelectionWithdrawals plan rewardWithdrawal outputAddressesToCycle
     = NE.nonEmpty
+    $ L.reverse
     $ fst
-    $ L.foldr
-        (accumulate)
+    $ L.foldl'
+        (flip accumulate)
         ([], NE.toList $ NE.cycle outputAddressesToCycle)
         (view #selections plan)
   where
@@ -1888,7 +1889,7 @@ migrationPlanToSelectionWithdrawals plan rewardWithdrawal outputAddressesToCycle
 
         outputAddressesRemaining :: [Address]
         outputAddressesRemaining =
-            drop (length $ view #outputs migrationSelection) outputAddresses
+            drop (length outputsCovered) outputAddresses
 
 {-------------------------------------------------------------------------------
                                   Delegation
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -121,6 +121,7 @@ module Cardano.Wallet
     -- ** Migration
     , createMigrationPlan
     , migrationPlanToSelectionWithdrawals
+    , SelectionResultWithoutChange
     , ErrCreateMigrationPlan (..)
 
     -- ** Delegation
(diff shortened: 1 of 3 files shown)


lib/core/test/unit/Cardano/WalletSpec.hs
```diff
@@ -364,12 +387,20 @@ spec = parallel $ do
...
+                property prop_migrationPlanToSelectionWithdrawals_addresses
+            it "Inputs and outputs are preserved in the correct order." $
+                property prop_migrationPlanToSelectionWithdrawals_io
+
+  where
+    pidA = PoolId "A"
+    pidB = PoolId "B"
+    pidUnknown = PoolId "unknown"
+    knownPools = Set.fromList [pidA, pidB]
+    next = WalletDelegationNext
```
jonathanknowles, line 403: This `where` clause is a bit annoying: it should belong to the pool delegation tests, but its scope covers the entire property test suite for this module.

lib/core/src/Cardano/Wallet/Primitive/Types/Address.hs
```diff
@@ -88,8 +90,10 @@ import qualified Data.Text.Encoding as T
 --
 newtype Address = Address
     { unAddress :: ByteString
-    } deriving (Read, Show, Generic, Eq, Ord)
-      deriving anyclass (NFData, Hashable)
+    }
+    deriving (Generic, Eq, Ord)
+    deriving anyclass (NFData, Hashable)
+    deriving (Read, Show) via (Quiet Address)
```
jonathanknowles, line 96: This makes the output of `show` much more concise for values of type `Address`.

Instead of:
```hs
Address { unAddress = "XXXX" }
```

We now have:
```hs
Address "XXXX"
```

lib/core/test/unit/Cardano/WalletSpec.hs
```diff
@@ -1019,6 +1050,170 @@ prop_throttle tc@(ThrottleTest interval diffTimes) = monadicIO $ do
...
+
+genMigrationTargetAddresses :: Gen (NonEmpty Address)
+genMigrationTargetAddresses = do
+    addressCount <- choose (1, 8)
+    pure $ (:|)
+        (mkAddress 'A')
+        (mkAddress <$> take (addressCount - 1) ['B' ..])
+  where
+    mkAddress :: Char -> Address
+    mkAddress c = Address $ "" `B8.snoc` c
```
jonathanknowles: Suggestion: use `B8.singleton` here.

jonathanknowles: Fixed in 054ab640f5db64707ef9e66dd8f736669a1905db.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1887,7 +1889,7 @@ migrationPlanToSelectionWithdrawals plan rewardWithdrawal outputAddressesToCycle
 
         outputAddressesRemaining :: [Address]
         outputAddressesRemaining =
-            drop (length $ view #outputs migrationSelection) outputAddresses
+            drop (length outputsCovered) outputAddresses
```
Anviking, line 1892: Is this a fix, or is it equivalent?

Anviking (approved):

lib/core/src/Cardano/Wallet.hs
jonathanknowles: > Is this a fix, or is it equivalent?

It's equivalent.

I guess, strictly speaking, this should be in a separate commit, as it isn't part of the fix. But it seems too minor to create a separate commit just for that.

PR #2658, opened 2021-05-19
Integration test coverage for migration of freeriders

jonathanknowles:
# Issue Number

ADP-840

# Overview

This PR adds integration test coverage for migrations of wallets with freerider UTxO entries.

In particular, we add two test cases:

- [x] `CREATE_MIGRATION_PLAN_07`
    This test verifies that we can create a **_complete_** migration plan for a wallet with a large number of freerider UTxO entries, but with **_just enough_** non-freerider entries to enable the entire UTxO set to be migrated.
- [x] `CREATE_MIGRATION_PLAN_08`
    This test verifies that we can create a **_partial_** migration plan for a wallet with a large number of freerider UTxO entries, but with **_not quite enough_** non-freerider entries to enable the entire UTxO set to be migrated.

In addition, this PR:

- [x] Adds a synchronization lock around the `mintSeaHorseAssets` operation, as it is not safe to be called by more than one thread concurrently.

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
@@ -435,6 +435,135 @@ spec = describe "SHELLEY_MIGRATIONS" $ do
                         . TokenMap.getAssets)
                 ]
 
+    it "SHELLEY_CREATE_MIGRATION_PLAN_08 - \
+        \Can create a partial migration plan for a wallet with a large number \
+        \of freerider UTxO entries, but with not quite enough non-freerider \
+        \entries to enable the entire UTxO set to be migrated."
+        $ \ctx -> runResourceT $ do
+
+            -- Create a source wallet with just one pure ada entry that is
+            -- large enough to create a singleton transaction:
+            sourceWallet <- fixtureWalletWith @n ctx [ 100_000_000 ]
+
+            -- Check that the source wallet has the expected balance and UTxO
+            -- distribution:
+            request @ApiWallet ctx
+                (Link.getWallet @'Shelley sourceWallet) Default Empty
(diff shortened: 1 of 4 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
```diff
@@ -513,9 +518,10 @@ spec = describe "SHELLEY_MIGRATIONS" $ do
                     <$> listUnusedAddresses @n ctx sourceWallet
                 liftIO $ _mintSeaHorseAssets ctx
                     perEntryAssetCount
+                    batchSize
                     perEntryAdaQuantity
                     sourceAddresses
-                waitForTxImmutability ctx
+            waitForTxImmutability ctx
```
jonathanknowles, line 524: @Anviking Is this safe? If we don't wait for tx immutability between each batch, then how can we be sure we're getting a batch of unused addresses when we call `listUnusedAddresses`?

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Migrations.hs
```diff
@@ -302,6 +321,262 @@ spec = describe "SHELLEY_MIGRATIONS" $ do
...
+                , expectField (#balanceLeftover . #ada . #getQuantity)
+                    (`shouldBe`  29_687_500)
+                , expectField (#balanceSelected . #assets . #getApiT)
+                    ((.> 0)
+                        . Set.size
+                        . TokenMap.getAssets)
+                , expectField (#balanceLeftover . #assets . #getApiT)
+                    ((.> 0)
+                        . Set.size
+                        . TokenMap.getAssets)
```
Anviking, line 577: nice!

Anviking (approved):

PR #2662, opened 2021-05-20
Basic migration plans e2e tests

piotr-iohk:
# Issue Number

ADP-680

# Overview

- [ ] e2e tests on testnet for creating migration plans for Byron/Icarus/Shelley wallets.

# Comments

### DIFF
--- a/test/e2e/spec/byron_spec.rb
+++ b/test/e2e/spec/byron_spec.rb
@@ -272,7 +272,8 @@ RSpec.describe CardanoWallet::Byron do
       addrs = SHELLEY.addresses.list(target_id).map { |a| a['id'] }
 
       plan = BYRON.migrations.plan(id, addrs)
-      expect(plan).to be_correct_and_respond 501
+      expect(plan).to be_correct_and_respond 403
+      expect(plan.to_s).to include "nothing_to_migrate"
     end
 
     it "I could create migration plan - random" do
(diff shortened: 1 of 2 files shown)


test/e2e/spec/byron_spec.rb
```diff
@@ -272,7 +272,8 @@
       addrs = SHELLEY.addresses.list(target_id).map { |a| a['id'] }
 
       plan = BYRON.migrations.plan(id, addrs)
-      expect(plan).to be_correct_and_respond 501
+      expect(plan).to be_correct_and_respond 403
+      expect(plan.to_s).to include "nothing_to_migrate"
```
jonathanknowles, line 276: This line reads just like an English sentence! 👍🏻

test/e2e/spec/e2e_spec.rb
```diff
@@ -129,6 +129,19 @@
 
     end
 
+    describe "Shelley Migrations" do
+      it "I can create migration plan shelley -> shelley" do
+        target_id = create_shelley_wallet
+        addrs = SHELLEY.addresses.list(target_id).map { |a| a['id'] }
+
+        plan = SHELLEY.migrations.plan(@wid, addrs)
```
jonathanknowles, line 137: Does the wallet referred to with `wid` have non-ada assets?

piotr-iohk: Yes, it has some non-ADA assets. It is fixture Shelley wallet used in e2e testnet tests.

jonathanknowles: In that case, I wonder if it would be possible to add an assertion about `balance_selected`, to assert that the set of unique assets is _not_ equal to `[]`?

piotr-iohk: done in 091e063.

jonathanknowles (approved):

PR #2663, opened 2021-05-21
Improve shared wallets

paweljakubas:
# Issue Number

adp-934

# Overview

- [x] Adding "incomplete" to status for pending shared wallets
- [x] comprehensive validation of script templates when posting them
- [x] integration testing of validation of script templates
- [x] add GET /shared-wallets/wid/keys and whole infrastructure
- [x] add GET /wallets/wid/keys
- [x] unite ApiVerificationKeyShelley and ApiVerificationKeyShared and hence enable with/without hashing capability for shelley style
- [x] better reuse of code in Api.Link
- [x] test new endpoints

# Comments
Prerequisite : https://github.com/input-output-hk/cardano-addresses/pull/131

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shared/Wallets.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shared/Wallets.hs
@@ -32,7 +32,7 @@ import Cardano.Wallet.Api.Types
     , KeyFormat (..)
     )
 import Cardano.Wallet.Primitive.AddressDerivation
-    ( DerivationIndex (..), Passphrase (..), Role (..), hex )
+    ( DerivationIndex (..), Passphrase (..), Role (..) )
 import Cardano.Wallet.Primitive.AddressDiscovery.Shared
     ( CredentialType (..) )
 import Cardano.Wallet.Primitive.SyncProgress
@@ -56,7 +56,9 @@ import Data.Text
 import Data.Text.Class
     ( ToText (..) )
 import Test.Hspec
-    ( SpecWith, describe, shouldBe, shouldNotBe )
+    ( SpecWith, describe )
+import Test.Hspec.Expectations.Lifted
+    ( shouldBe, shouldNotBe )
 import Test.Hspec.Extra
     ( it )
 import Test.Integration.Framework.DSL
@@ -75,10 +77,12 @@ import Test.Integration.Framework.DSL
     , fixturePassphrase
     , genMnemonics
     , genXPubs
+    , getAccountKeyShared
     , getFromResponse
     , getSharedWallet
     , getSharedWalletKey
     , getWalletIdFromSharedWallet
+    , hexText
     , json
     , listFilteredSharedWallets
     , notDelegating
@@ -94,6 +98,10 @@ import Test.Integration.Framework.TestData
     , errMsg403KeyAlreadyPresent
     , errMsg403NoDelegationTemplate
     , errMsg403NoSuchCosigner
+    , errMsg403TemplateInvalidDuplicateXPub
+    , errMsg403TemplateInvalidNoCosignerInScript
+    , errMsg403TemplateInvalidScript
+    , errMsg403TemplateInvalidUnknownCosigner
     , errMsg403WalletAlreadyActive
     )
 
@@ -103,7 +111,6 @@ import qualified Data.Set as Set
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
 import qualified Network.HTTP.Types as HTTP
-import qualified Test.Hspec.Expectations.Lifted as Expectations
 
 spec :: forall n.
     ( DecodeAddress n
@@ -167,8 +174,17 @@ spec = describe "SHARED_WALLETS" $ do
             [ expectResponseCode HTTP.status202
             , expectField #format (`shouldBe` Extended)
             ]
-        let (ApiAccountKeyShared bytes _) = getFromResponse id rKey
-        T.decodeUtf8 (hex bytes) `Expectations.shouldBe` accXPubDerived
+        let ApiAccountKeyShared bytes _ = getFromResponse id rKey
+        hexText bytes `shouldBe` accXPubDerived
+
+        aKey <- getAccountKeyShared ctx wal (Just Extended)
+
+        verify aKey
+            [ expectResponseCode HTTP.status200
+            , expectField #format (`shouldBe` Extended)
+            ]
+        let ApiAccountKeyShared bytes' _ = getFromResponse id aKey
+        hexText bytes' `shouldBe` accXPubDerived
 
     it "SHARED_WALLETS_CREATE_02 - Create a pending shared wallet from root xprv" $ \ctx -> runResourceT $ do
         m15txt <- liftIO $ genMnemonics M15
@@ -220,8 +236,8 @@ spec = describe "SHARED_WALLETS" $ do
             [ expectResponseCode HTTP.status202
             , expectField #format (`shouldBe` Extended)
             ]
-        let (ApiAccountKeyShared bytes _) = getFromResponse id rKey
-        T.decodeUtf8 (hex bytes) `Expectations.shouldBe` accXPubDerived
+        let ApiAccountKeyShared bytes _ = getFromResponse id rKey
+        hexText bytes `shouldBe` accXPubDerived
 
     it "SHARED_WALLETS_CREATE_03 - Create an active shared wallet from account xpub" $ \ctx -> runResourceT $ do
         (_, accXPubTxt):_ <- liftIO $ genXPubs 1
@@ -240,8 +256,8 @@ spec = describe "SHARED_WALLETS" $ do
                           }
                     }
                 } |]
-        r <- postSharedWallet ctx Default payload
-        verify (second (\(Right (ApiSharedWallet (Right res))) -> Right res) r)
+        rPost <- postSharedWallet ctx Default payload
+        verify (second (\(Right (ApiSharedWallet (Right res))) -> Right res) rPost)
             [ expectResponseCode HTTP.status201
             , expectField
                     (#name . #getApiT . #getWalletName) (`shouldBe` "Shared Wallet")
@@ -257,6 +273,16 @@ spec = describe "SHARED_WALLETS" $ do
             , expectField #delegationScriptTemplate (`shouldBe` Nothing)
             , expectField (#accountIndex . #getApiT) (`shouldBe` DerivationIndex 2147483658)
             ]
+
+        let wal = getFromResponse id rPost
+        aKey <- getAccountKeyShared ctx wal (Just Extended)
+
+        verify aKey
+            [ expectResponseCode HTTP.status200
+            , expectField #format (`shouldBe` Extended)
+            ]
+        let ApiAccountKeyShared bytes' _ = getFromResponse id aKey
+        hexText bytes' `shouldBe` accXPubTxt
 
     it "SHARED_WALLETS_CREATE_04 - Create a pending shared wallet from account xpub" $ \ctx -> runResourceT $ do
         (_, accXPubTxt):_ <- liftIO $ genXPubs 1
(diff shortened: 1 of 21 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Shared/Wallets.hs
```diff
@@ -258,6 +273,17 @@ spec = describe "SHARED_WALLETS" $ do
...
+        let wal = getFromResponse id rPost
+        aKey <-
+            getAccountKeyShared ctx wal (Just True)
+
+        verify aKey
+            [ expectResponseCode HTTP.status200
+            , expectField #format (`shouldBe` Extended)
+            ]
+        let (ApiAccountKeyShared bytes' _) = getFromResponse id aKey
+        T.decodeUtf8 (hex bytes') `Expectations.shouldBe` accXPubTxt
```
rvl: The lifted expectations module is meant to replace Test.Hspec - so no need to import it qualified.

Also a util function for hexText would be nice.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2391,8 +2412,8 @@ data ErrAddCosignerKey
     deriving (Eq, Show)
 
 data ErrConstructSharedWallet
-    = ErrConstructSharedWalletMissingKey
-        -- ^ The shared wallet' script template doesn't have the wallet's account public key
+    = ErrConstructSharedWalletWrongScriptTemplate CredentialType Text
```
rvl: Can we put this error type into the `AddressDiscovery.Shared` module with a `ToText` instance, and `Text` replaced with `ErrValidateScriptTemplate`?

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -382,6 +386,13 @@ type PostAccountKey = "wallets"
     :> ReqBody '[JSON] ApiPostAccountKeyData
     :> PostAccepted '[JSON] ApiAccountKey
 
+-- | https://input-output-hk.github.io/cardano-wallet/api/#operation/getAccountKey
+type GetAccountKey = "wallets"
+    :> Capture "walletId" (ApiT WalletId)
+    :> "keys"
+    :> QueryParam "extended" Bool
```
rvl: It's probably OK to have `hash` as a bool query param, but maybe not `extended`. Didn't we use a sum type elsewhere?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2353,13 +2344,33 @@ postAccountPublicKey
     -> Handler account
 postAccountPublicKey ctx mkAccount (ApiT wid) (ApiT ix) (ApiPostAccountKeyData (ApiT pwd) extd) = do
     withWorkerCtx @_ @s @k ctx wid liftE liftE $ \wrk -> do
-        k <- liftHandler $ W.readPublicAccountKey @_ @s @k wrk wid pwd ix
+        k <- liftHandler $ W.getAccountPublicKeyAtIndex @_ @s @k wrk wid pwd ix
+        pure $ mkAccount (xPubtoBytes extd $ getRawKey k) extd
+
+xPubtoBytes :: KeyFormat -> XPub -> ByteString
+xPubtoBytes = \case
```
rvl: That was confusing with the letter case differences. How about:
```suggestion
publicKeyToBytes' :: WalletKey k => KeyFormat -> k -> ByteString
publickKeyToBytes' f k = case f of
    Extended -> xpubToBytes $ getRawKey k
    NonExtended -> xpubPublicKey $ getRawKey k
```

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1484,29 +1487,39 @@ instance FromJSON (ApiT DerivationIndex) where
...
+        (role, hashing) <- parseRoleHashing hrp
+        payload <- case hashing of
+            WithoutHashing -> parsePubVer bytes
+            WithHashing -> parsePubVerHash bytes
+        pure $ ApiVerificationKeyShelley (payload,role) hashing
       where
-        parseRole = \case
-            hrp | hrp == [humanReadablePart|addr_vk|] -> pure UtxoExternal
-            hrp | hrp == [humanReadablePart|stake_vk|] -> pure MutableAccount
+        parseRoleHashing = \case
```
rvl, line 1532: How would it look if we get the hrp as text, split on `"_"` and pattern match on that?

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
```diff
@@ -305,17 +314,37 @@ accountXPubCondition
...
+validateScriptTemplates
     :: WalletKey k
     => k 'AccountK XPub
+    -> ValidationLevel
     -> ScriptTemplate
     -> Maybe ScriptTemplate
-    -> Bool
-walletCreationInvariant accXPub pTemplate dTemplate =
-    isValid pTemplate && maybe True isValid dTemplate
+    -> Either (CredentialType, Text) ()
```
rvl: It would be nice to wrap this return code up in something like
```haskell
data ErrScriptTemplateInvalid = ErrScriptTemplateInvalid !CredentialType !ErrValidateScriptTemplate
instance ToText ErrScriptTemplateInvalid where
    -- ...
```

specifications/api/swagger.yaml
```diff
@@ -267,6 +267,11 @@ x-addressState: &addressState
     - used
     - unused
 
+x-statusIncomplete: &statusIncomplete
+  type: string
+  enum:
```
rvl: Can we add a description field to document the possible status values?

specifications/api/swagger.yaml
```diff
@@ -3095,18 +3104,18 @@ x-errSharedWalletCannotUpdateKey: &errSharedWalletCannotUpdateKey
...
-  title: shared_wallet_create_not_allowed
+  title: shared_wallet_script_template_not_validated
   properties:
     message:
       type: string
       description: |
-        Returned when a user tries to create a shared wallet with missing wallet's
-        account key in all the script templates.
+        Returned when a user tries to create a shared wallet with script template
+        that is not validated.
```
```suggestion
        Returned when a user tries to create a shared wallet with script template
        that does not pass validation.
```

specifications/api/swagger.yaml
```diff
@@ -3095,18 +3104,18 @@ x-errSharedWalletCannotUpdateKey: &errSharedWalletCannotUpdateKey
...
     code:
       type: string
-      enum: ['shared_wallet_create_not_allowed']
+      enum: ['shared_wallet_script_template_not_validated']
```
rvl: Not validated = Invalid?

```suggestion
      enum: ['shared_wallet_script_template_invalid']
```

specifications/api/swagger.yaml
```diff
@@ -4505,7 +4534,7 @@ paths:
       summary: Create
       description: |
         <p align="right">status: <strong>stable</strong></p>
-        Retrieve account public key from the wallet.
+        Retrieve any account public key from the wallet provided it was created from mnemonics.
```
```suggestion
        Derive an account public key for any account index. For this key derivation to be possible, the wallet must have been created from mnemonic.
```

specifications/api/swagger.yaml
```diff
@@ -5441,7 +5489,7 @@ paths:
       summary: Create
       description: |
         <p align="right">status: <strong>stable</strong></p>
-        Retrieve account public key from the shared wallet.
+        Derive any account public key from the shared wallet provided the wallet was created from mnemonics.
```
rvl: Comment as above

specifications/api/swagger.yaml
```diff
@@ -5454,6 +5502,22 @@ paths:
...
       responses: *responsesPostAccountKeyShared
 
+  /shared-wallets/{walletId}/keys:
+    get:
+      operationId: getAccountKeyShared
+      tags: ["Shared Keys"]
+      summary: Get
+      description: |
+        <p align="right">status: <strong>stable</strong></p>
+        Retrieve account public key of the shared wallet.
```
```suggestion
        Retrieve the account public key of this shared wallet.
```

rvl (approved):
:+1: Nice improvements. I have some suggestions though.

piotr-iohk:
@paweljakubas perhaps that's a silly question, but why do we need two separate endpoints for getting account public key? Couldn't this be realized by already existing:
 - https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/postAccountKey
 - https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/postAccountKeyShared

I assume this was done to address:

2. cannot return account public key for the wallet that's created from acc xpub using https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/postAccountKeyShared
Following error is returned:

```
 {"code":"no_root_key","message":"I couldn't find a root private key for the given wallet: 0eb51c11fc2eda7b2b34aa99ceb89bd6044e4015. However, this operation requires that I do have such a key. Either there's no such wallet, or I don't fully own it."}
```
The same behavior is for shelley wallets. Both have been considered as a bug as per comment -> https://github.com/input-output-hk/cardano-wallet/pull/2624#issuecomment-834135538

specifications/api/swagger.yaml
```diff
@@ -2076,13 +2081,15 @@ components:
...
         - payment_script_template
+        - status
       properties:
         id: *walletId
         name: *walletName
         account_index: *derivationSegment
         address_pool_gap: *walletAddressPoolGap
         payment_script_template: *scriptTemplate
         delegation_script_template: *scriptTemplate
+        status: *statusIncomplete
```
piotr-iohk: As far as I see the status of complete/active wallet is displayed as:
```
  "state": {
    "status": "syncing",
   ...
  }

or

  "state": {
    "status": "ready"
  }
```

Whereas incomplete wallet's status is not wrapped into `state` and displays only:

```
"status": "incomplete"
```

Maybe it would be nice to keep this convention and have:

```
  "state": {
    "status": "incomplete"
  }
```
(That's something I'd expect as an API user at least)

paweljakubas: done

lib/core/src/Cardano/Wallet/Api.hs
paweljakubas: introduced "format=extended" and "format=non_extended" and used `Maybe KeyFormat` rather than `Maybe Bool`

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Shared.hs
paweljakubas: done

paweljakubas:
@piotr-iohk I was thinking long how to fix "posting" account public keys for account based wallets. And get to the conclusion that we should have POST which stands for "deriving" and it means we need mnemonics to do it, so the wallet have to be root key based. Also we can GET the account public key for a given wallet, and it should work for each wallet, account and root based, and also for shared wallets and shelley wallets. I added support for both shelley and shared wallets, also checked in integration testing that it works for root and account based wallets

rvl:
@piotr-iohk Thanks for the question about multiple API endpoints which do similar things. This is not a silly question at all and it's good to take a step back and assess the API as a whole, like you have done.

To add to @paweljakubas answer, yes they do slightly different things. The POST one can derive an account public key for any index. To do this, it must be supplied with the password to decrypt the root key. The GET one can only return the "default" or "current" account public key for this wallet. That key is already cached in the wallet state, so no need to derive anything, and therefore no need for a password.

Then we get on to the mechanics of HTTP. It is generally discouraged to put passwords in GET query parameters. The preferred way is to use a HTTP header or put the password into the request body of a POST request (the latter often being easier for API users). However, these key derivation endpoints are really GETs, semantically speaking.

Anyway, your comment convinced me that it is less confusing to have only one account key endpoint, with optional parameters. It has to be POST. After this PR, I would like to spec out a streamlined API for deriving, which will save us work in future.

rvl:
I squashed the git history a bit, rebased, made a minor stylistic change to `swagger.yaml`, and fixed the failing integration tests.

bors r+

piotr-iohk:
@rvl @paweljakubas thanks for your responses.

> Anyway, your comment convinced me that it is less confusing to have only one account key endpoint, with optional parameters. It has to be POST. After this PR, I would like to spec out a streamlined API for deriving, which will save us work in future.

@rvl is this something we'd like to tackle as part of ADP-934 or later?

rvl:
I think it can be a separate ticket, so we don't hold up the address derivation part of multisig.

PR #2664, opened 2021-05-24
Fix slow ApiSharedWallet unit tests

rvl:
### Issue Number

Found during ADP-902

### Overview

The `Cardano.Wallet.Api.Types` tests are running very slowly. Also the unit tests were sometimes timing out in CI.

Profiling showed that the `Arbitrary` instance of `ApiSharedWallet` was the slowest part.
This change improves the speed of this generator.
Now it is the OpenAPI validation which is the slowest part.

### Comments

We may also be able to reduce the size of arbitrary `ApiSharedWallet` values, so that validation gets faster. Because it's still not especially quick. Something like the `instance Arbitrary WalletName` change in this PR could probably be applied.

### DIFF
--- a/lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
@@ -327,6 +327,7 @@ import Test.QuickCheck
     , arbitraryPrintableChar
     , arbitrarySizedBoundedIntegral
     , choose
+    , chooseInt
     , counterexample
     , elements
     , frequency
@@ -334,6 +335,7 @@ import Test.QuickCheck
     , property
     , scale
     , shrinkIntegral
+    , sized
     , vector
     , vectorOf
     , (.&&.)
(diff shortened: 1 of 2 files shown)


lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -1568,8 +1570,9 @@ instance Arbitrary WalletId where
 
 instance Arbitrary WalletName where
     arbitrary = do
-        nameLength <- choose (walletNameMinLength, walletNameMaxLength)
-        WalletName . T.pack <$> replicateM nameLength arbitraryPrintableChar
+        len <- Test.QuickCheck.scale (min walletNameMaxLength) $ sized $ \n ->
+            chooseInt (walletNameMinLength, walletNameMinLength `max` n)
```
jonathanknowles, line 1574: 👍🏻

lib/core/test/unit/Cardano/Wallet/Gen.hs
```diff
@@ -338,27 +328,15 @@ genScriptTemplateComplete :: Gen ScriptTemplate
...
-    return $ Shelley.unsafeGenerateKeyFromSeed (s, g) encryptionPass
-
-genXPub :: Gen XPub
-genXPub = do
-    getRawKey . publicKey <$> (genRootKeySeqWithPass =<< genPassphrase (0, 16))
+genMockXPub :: Gen XPub
+genMockXPub = fromMaybe impossible . xpubFromBytes . BS.pack <$> genBytes
+  where
+    genBytes = vectorOf 64 arbitrary
+    impossible = error "incorrect length in genMockXPub"
```
jonathanknowles, line 338: 👍🏻

jonathanknowles (approved):

lib/core/test/unit/Cardano/Wallet/Gen.hs
paweljakubas: yeah, we can generate here random bytes :100:

paweljakubas (approved):
great catches!

PR #2665, opened 2021-05-24
Add basic latency benchmarks for migrations

piotr-iohk:
# Issue Number

ADP-680

# Overview

- 2c31861c18d72c5ca8ce46fe39a1f9e87e59ba68
  Add basic latency benchmarks for migrations

# Comments

[Migration Plan](http://cardano-wallet-benchmarks.herokuapp.com/latency?latency_category=4&latency_benchmark=all&latency_measurement=postMigrationPlan)
[Migration](http://cardano-wallet-benchmarks.herokuapp.com/latency?latency_category=4&latency_benchmark=all&latency_measurement=postMigration)

### DIFF
--- a/lib/shelley/bench/Latency.hs
+++ b/lib/shelley/bench/Latency.hs
@@ -37,6 +37,7 @@ import Cardano.Wallet.Api.Types
     , ApiTxId (..)
     , ApiUtxoStatistics
     , ApiWallet
+    , ApiWalletMigrationPlan (..)
     , EncodeAddress (..)
     , WalletStyle (..)
     )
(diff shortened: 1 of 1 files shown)


lib/shelley/bench/Latency.hs
```diff
@@ -384,6 +386,23 @@ walletApiBench capture ctx = do
...
+
+        -- Perform a migration:
+        let endpointMigrate = Link.migrateWallet @'Shelley walMAForMigration
+        t12b <- measureApiLogs capture $ request @[ApiTransaction n]
+            ctx endpointMigrate Default $
+            Json [json|
+                { passphrase: #{fixturePassphrase}
+                , addresses: #{addresses}
+                }|]
+        fmtResult "postMigration      " t12b
```
jonathanknowles, line 404: 👍🏻

lib/shelley/bench/Latency.hs
```diff
@@ -203,15 +204,16 @@ walletApiBench capture ctx = do
     nFixtureWallet n = do
         wal1 : wal2 : _ <- replicateM n (fixtureWallet ctx)
         walMA <- fixtureMultiAssetWallet ctx
-        pure (wal1, wal2, walMA)
+        walMAForMigration <- fixtureMultiAssetWallet ctx
+        pure (wal1, wal2, walMA, walMAForMigration)
```
jonathanknowles: Hi @piotr-iohk! This is just a suggestion, but perhaps we could use a name that makes it obvious that it's a source wallet to be migrated:

```suggestion
        pure (wal1, wal2, walMA, maWalletToMigrate)
```

What do you think about this name?

("For migration" could imply that it's a target wallet.)

jonathanknowles (approved):
LGTM! 👍🏻 

I only have one suggestion about the name of the wallet.

PR #2666, opened 2021-05-25
Add GET /wallets/{walletId}/stake-keys endpoint

Anviking:
# Issue Number

ADP-721

# Overview

- [x] Add `GET /wallets/{wid}/stake-keys` endpoint
- [x] Provide information about stake-key 0
- [x] Show the distribution of ada across different stake keys

# Comments

Skärmavbild 2021-05-25 kl  13 01 30

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
@@ -18,6 +18,7 @@ import Prelude
 import Cardano.Wallet.Api.Types
     ( ApiCertificate (JoinPool, QuitPool, RegisterRewardAccount)
     , ApiHealthCheck
+    , ApiStakeKeys
     , ApiStakePool (flags)
     , ApiStakePoolFlag (..)
     , ApiT (..)
@@ -84,7 +85,7 @@ import Numeric.Natural
 import Test.Hspec
     ( SpecWith, describe, pendingWith )
 import Test.Hspec.Expectations.Lifted
-    ( shouldBe, shouldSatisfy )
+    ( expectationFailure, shouldBe, shouldSatisfy )
 import Test.Hspec.Extra
     ( flakyBecauseOf, it )
 import Test.Integration.Framework.Context
@@ -118,6 +119,7 @@ import Test.Integration.Framework.DSL
     , postWallet
     , quitStakePool
     , quitStakePoolUnsigned
+    , replaceStakeKey
     , request
     , triggerMaintenanceAction
     , unsafeRequest
@@ -127,6 +129,7 @@ import Test.Integration.Framework.DSL
     , verifyMaintenanceAction
     , verifyMetadataSource
     , waitForNextEpoch
+    , waitForTxImmutability
     , walletId
     , (.<)
     , (.>)
@@ -303,6 +306,23 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
                   [ expectField
                       (#balance . #reward) (`shouldBe` walletRewards)
                   ]
+
+
+        -- Listing stake keys shows
+        request @(ApiStakeKeys n) ctx (Link.listStakeKeys src) Default Empty
+            >>= flip verify
+            [ expectField (#_foreign) (`shouldBe` [])
+            , expectField (#_ours) (\case
+                [acc] -> do
+                    (acc ^. #_stake) .> Quantity 0
+                    acc ^. (#_delegation . #active . #status)
+                        `shouldBe` Delegating
+                    acc ^. (#_delegation . #active . #target)
+                        `shouldBe` (Just pool)
+                _ -> expectationFailure "wrong number of accounts in \"ours\""
+                )
+            , expectField (#_none . #_stake) (.> Quantity 0)
+            ]
 
         -- there's currently no withdrawals in the wallet
         rw1 <- request @[ApiTransaction n] ctx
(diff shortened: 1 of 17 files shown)


lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1426,6 +1431,42 @@ newtype ApiMnemonicT (sizes :: [Nat]) =
     ApiMnemonicT { getApiMnemonicT :: SomeMnemonic }
     deriving (Generic, Show, Eq)
 
+
+data ApiOurStakeKey n = ApiOurStakeKey
+     { _index :: ApiT DerivationIndex
```
Anviking: Seems `DerivationIndex` is encoded as a JSON string, which seems unnecessary. Going with `Natural` then, I guess. Could maybe have some `ApiSoftIndex` newtype if desired…

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -1234,6 +1254,111 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
...
+                        `shouldBe` NotDelegating
+                _ -> expectationFailure "wrong number of accounts in \"ours\""
+                )
+            , expectField (#_null . #_stake) (`shouldBe` balance)
+            ]
+
+        -- By sending funds to ourselves, we associate funds with our stake key.
+        -- Both from the payment output itself and the change.
+        addrs <- listAddresses @n ctx w
+        let delegatingBalance = 950000000000 :: Int
```
Anviking: Misleading variable name; should inline

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1426,6 +1431,42 @@ newtype ApiMnemonicT (sizes :: [Nat]) =
...
+
+newtype ApiNullStakeKey = ApiNullStakeKey
+    { _stake :: Quantity "lovelace" Natural
+      -- ^ The total stake of the wallet UTxO that is not associated with a
+      -- stake key, because it's part of an enterprise address.
+    } deriving (Generic, Eq, Show)
+
+data ApiStakeKeys n = ApiStakeKeys
+    { _ours :: ![ApiOurStakeKey n]
+    , _foreign :: ![ApiForeignStakeKey n]
```
Anviking, line 1506: Should be consistent on strictness

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1426,6 +1431,46 @@ newtype ApiMnemonicT (sizes :: [Nat]) =
...
+    { _stake :: Quantity "lovelace" Natural
+      -- ^ The total stake of the wallet UTxO that is not associated with a
+      -- stake key, because it's part of an enterprise address.
+    } deriving (Generic, Eq, Show)
+
+-- | Collection of stake keys associated with a wallet.
+data ApiStakeKeys n = ApiStakeKeys
+    { _ours :: ![ApiOurStakeKey n]
+    , _foreign :: ![ApiForeignStakeKey n]
+    , _null :: !ApiNullStakeKey
```
KtorZ: `null` being a special keyword in JavaScript and JSON, I'd highly recommend to NOT using `null` as a field name! Perhaps `none` could work a bit better or `inactive` (though stake can be associated with a stake key and still be inactive.. so perhaps it is a poor choice).

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2023,6 +2036,83 @@ quitStakePool ctx (ApiT wid) body = do
...
+        , s ~ SeqState n k
+        , HasNetworkLayer IO ctx
+        , Typeable n
+        , Typeable s
+        )
+    => (Address -> Maybe RewardAccount)
+    -> ctx
+    -> ApiT WalletId
+    -> Handler (ApiStakeKeys n)
+listStakeKeys stakeRef ctx (ApiT wid) = do
```
rvl: This is quite a big function. Is it at all possible to break it up a little? Then it might be possible to unit test some of it.

rvl (approved):
Looks pretty sensible :+1:

lib/core/src/Cardano/Wallet/Api/Types.hs
Anviking: Ok, fair! Probably `none` then

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: 👍 The thought occurred to me, but wasn't sure if worth it; thanks for suggesting, will give it a shot

lib/core/src/Cardano/Wallet/Api/Types.hs
Anviking: Another thing: I think currently the balance of say script and pointer addresses get included into this `none`, which could be really confusing if it happens...

`none` should probably only correspond to enterprise addresses. 

Then we could have another bucket for "other/unknown", or leave it out for now.

Anviking: > Then we could have another bucket for "other/unknown", or leave it out for now.

For now I just mentioned that `none` _could_ include more things than enterprise addresses in the swagger API. There is no clear use-case besides testing/debugging, so I think this is fine as long as it is mentioned.

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: I split off a helper that could run in `Identity`. But to actually unit-test it, I can't get a way with only defining Api types like I do here. Will do in a separate PR.

Edit: although those attempts didn't go great, so maybe not.

PR #2667, opened 2021-05-25
Add property test for `UTxOIndex.selectRandomWithPriority`.

jonathanknowles:
# Issue Number

ADP-890

# Overview

This PR adds a property test for `UTxOIndex.selectRandomWithPriority`.

The `selectRandomWithPriority`  function is designed to:
- select an entry at random from a UTxO index according to a specified list of filter conditions;
- traverse the specified list of filter conditions in order of priority **_from left to right_**.

The test added in this PR provides a basic sanity check to verify that priority order is respected.

# Sample Output

```hs
Cardano.Wallet.Primitive.Types.UTxOIndex
  Indexed UTxO set properties
    Index Selection
      prop_selectRandomWithPriority
        +++ OK, passed 1600 tests:
        59.69% have match for neither asset 1 nor asset 2
        17.12% have match for asset 1 but not for asset 2
        16.31% have match for asset 2 but not for asset 1
         6.88% have match for both asset 1 and asset 2

Finished in 1.0870 seconds
1 example, 0 failures
```

# QA Due Diligence

I ran this test 500 times to increase confidence that it will not fail spuriously. No failures were encountered.

### DIFF
--- a/lib/core/test/unit/Cardano/Wallet/Primitive/Types/UTxOIndexSpec.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Primitive/Types/UTxOIndexSpec.hs
@@ -57,12 +57,13 @@ import Test.QuickCheck
     , conjoin
     , counterexample
     , cover
+    , forAll
     , oneof
     , property
     , stdConfidence
+    , suchThat
     , withMaxSuccess
     , (===)
-    , (==>)
     )
 import Test.QuickCheck.Classes
     ( eqLaws )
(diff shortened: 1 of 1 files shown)


lib/core/test/unit/Cardano/Wallet/Primitive/Types/UTxOIndexSpec.hs
```diff
@@ -531,6 +537,38 @@ prop_selectRandom_all_withAssetOnly u a = checkCoverage $ monadicIO $ do
     assert $ UTxOIndex.deleteMany (fst <$> selectedEntries) u == u'
     assert $ UTxOIndex.insertMany selectedEntries u' == u
 
+-- | Verify that priority order is respected when selecting with more than
+--   one filter.
+--
+prop_selectRandomWithPriority :: UTxOIndex -> Property
+prop_selectRandomWithPriority u =
+    forAll (genAssetIdSmallRange) $ \a1 ->
```
rvl, line 545: Just wondering, are the two levels of forAll essential, or can one AssetId be a function parameter?

rvl (approved):
:ok_hand:

lib/core/test/unit/Cardano/Wallet/Primitive/Types/UTxOIndexSpec.hs
jonathanknowles: Indeed, the first level could be replaced with a function parameter.

However, by using `forAll` for both, it's easy to see (at a glance) that both asset ids come from the same generator.

PR #2670, opened 2021-05-26
API: Pass TokenMetadataError through to API

rvl:
### Issue Number

ADP-925

### Overview

If there was an error fetching metadata in the listMetadata or getMetadata endpoints, tell the user, rather than just returning "no metadata".

### Comments

Could someone please pick this up and merge for v2021-05-26.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -266,7 +266,7 @@ import Cardano.Wallet.Primitive.Types.Tx
 import Cardano.Wallet.Primitive.Types.UTxO
     ( BoundType, HistogramBar (..), UTxOStatistics (..) )
 import Cardano.Wallet.TokenMetadata
-    ()
+    ( TokenMetadataError (..) )
 import Codec.Binary.Bech32
     ( dataPartFromBytes, dataPartToBytes )
 import Codec.Binary.Bech32.TH
@@ -480,7 +480,8 @@ data ApiAsset = ApiAsset
     , assetName :: ApiT W.TokenName
     , fingerprint :: ApiT W.TokenFingerprint
     , metadata :: Maybe ApiAssetMetadata
-    } deriving (Eq, Generic, Ord, Show)
+    , metadataError :: Maybe Text
+    } deriving (Eq, Generic, Show)
       deriving anyclass NFData
 
 data ApiAssetMetadata = ApiAssetMetadata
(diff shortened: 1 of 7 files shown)


lib/core/src/Cardano/Wallet/TokenMetadata.hs
```diff
@@ -356,6 +355,9 @@ data TokenMetadataError
         -- ^ Error from aeson decoding of JSON
     deriving (Generic, Show, Eq)
 
+instance NFData TokenMetadataError where
+    rnf = rnf . show
```
Anviking, line 359: Is this `show` right? 🤔

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -480,7 +480,8 @@ data ApiAsset = ApiAsset
     , assetName :: ApiT W.TokenName
     , fingerprint :: ApiT W.TokenFingerprint
     , metadata :: Maybe ApiAssetMetadata
-    } deriving (Eq, Generic, Ord, Show)
+    , metadataError :: Maybe Text
```
Anviking: Replaced with a proper sum-type

Anviking (approved):
lgtm

Testing this seems a bit cumbersome, as I think our mock metadata server would refuse to serve invalid JSON.

Anviking:
Oops - having hlint passing is probably a good idea before borsing 🙈

lib/core/src/Cardano/Wallet/TokenMetadata.hs
rvl: Yes it's a cheap trick to make `NFData` instances, where some of the fields don't have `NFData`.
In this case the problem was `SomeException`.
`show` usually works well enough for the purpose of deeply evaluating some data.

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -1791,8 +1799,37 @@ instance Arbitrary Word31 where
     arbitrary = arbitrarySizedBoundedIntegral
     shrink = shrinkIntegral
 
+instance Arbitrary AssetDecimals where
+    arbitrary = AssetDecimals <$> choose (1, 19)
+
+instance Arbitrary AssetMetadata where
+    -- TODO: We should add a proper arbitrary instance
```
rvl, line 1806: Thanks, yes we should.

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -1791,8 +1799,37 @@ instance Arbitrary Word31 where
...
+    arbitrary = AssetMetadata
+        <$> (pure "asset")
+        <*> (pure "An asset")
+        <*> (oneof [pure Nothing, pure $ Just "AST"])
+        <*> genMaybe (pure $ AssetURL $ fromJust $
+                parseURI "https://asset.url")
```
Anviking, line 1812: FYI - I first tried using an existing `Aribitrary URI` instance, but it failed, because it generated `http` urls despite the swagger api promising it to be `https`. 

I'm not sure what the metadata server promises, but if it allows `http` urls, then our swagger spec promises too much...

PR #2672, opened 2021-05-26
remove redundant volumes from docker-compose

piotr-iohk:
# Issue Number

# Overview

- [ ] I have removed unnecessary volumes from docker-compose that used to be used for testing some mainnet_candidate networks that no longer exist.

# Comments

Thanks @james-iohk for spotting this!


rvl (approved):
Thanks

PR #2679, opened 2021-05-28
Add UTxO snapshot API endpoint and CLI command.

jonathanknowles:
# Issue Number

Related to ADP-890.

# Overview

This PR adds the `getWalletUtxoSnapshot` API endpoint and associated CLI command.

The intended use is for debugging and development.

The endpoint generates a complete snapshot of a wallet's UTxO set, as an unordered list of UTxO entries.

It produces JSON output similar to the following:
```json
"entries": [
  { "ada":         {"quantity": 8, "unit": "lovelace"}
  , "ada_minimum": {"quantity": 1, "unit": "lovelace"}
  , "assets": []
  },
  { "ada":         {"quantity": 4, "unit": "lovelace"}
  , "ada_minimum": {"quantity": 2, "unit": "lovelace"}
  , "assets": [
      { "policy_id": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
      , "asset_name": "APPLE"
      , "quantity": 10
      }
    ]
  },
  { "ada":         {"quantity": 4, "unit": "lovelace"}
  , "ada_minimum": {"quantity": 2, "unit": "lovelace"}
  , "assets": [
      { "policy_id": "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"
      , "asset_name": "BANANA"
      , "quantity": 20
      }
    ]
  }
]
```

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Link.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Link.hs
@@ -44,6 +44,7 @@ module Cardano.Wallet.Api.Link
     , postWallet
     , putWallet
     , putWalletPassphrase
+    , getUTxOsSnapshot
     , getUTxOsStatistics
     , createMigrationPlan
     , migrateWallet
@@ -233,6 +234,21 @@ getUTxOsStatistics
 getUTxOsStatistics w = discriminate @style
     (endpoint @Api.GetUTxOsStatistics (wid &))
     (endpoint @Api.GetByronUTxOsStatistics (wid &))
+    (notSupported "Shared")
+  where
+    wid = w ^. typed @(ApiT WalletId)
+
+getUTxOsSnapshot
+    :: forall (style :: WalletStyle) w.
+        ( HasCallStack
+        , Discriminate style
+        , HasType (ApiT WalletId) w
+        )
+    => w
+    -> (Method, Text)
+getUTxOsSnapshot w = discriminate @style
+    (endpoint @Api.GetWalletUtxoSnapshot (wid &))
+    (endpoint @Api.GetByronWalletUTxOSnapshot (wid &))
     (notSupported "Shared")
   where
     wid = w ^. typed @(ApiT WalletId)
--- a/lib/core/src/Cardano/Wallet/Api/Link.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Link.hs
@@ -44,7 +44,7 @@ module Cardano.Wallet.Api.Link
     , postWallet
     , putWallet
     , putWalletPassphrase
-    , getUTxOsSnapshot
+    , getWalletUtxoSnapshot
     , getUTxOsStatistics
     , createMigrationPlan
     , migrateWallet
@@ -238,7 +238,7 @@ getUTxOsStatistics w = discriminate @style
   where
     wid = w ^. typed @(ApiT WalletId)
 
-getUTxOsSnapshot
+getWalletUtxoSnapshot
     :: forall (style :: WalletStyle) w.
         ( HasCallStack
         , Discriminate style
@@ -246,9 +246,9 @@ getUTxOsSnapshot
         )
     => w
     -> (Method, Text)
-getUTxOsSnapshot w = discriminate @style
+getWalletUtxoSnapshot w = discriminate @style
     (endpoint @Api.GetWalletUtxoSnapshot (wid &))
-    (endpoint @Api.GetByronWalletUTxOSnapshot (wid &))
+    (endpoint @Api.GetByronWalletUtxoSnapshot (wid &))
     (notSupported "Shared")
   where
     wid = w ^. typed @(ApiT WalletId)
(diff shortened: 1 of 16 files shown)


piotr-iohk:
@jonathanknowles I've added some basic integration tests for API and CLI. Perhaps assertions might be improved for checking the entries though...

piotr-iohk (approved):
lgtm

paweljakubas (approved):
lgtm!

I would suggest more integration test(s) - can be in some future PR:
1. setup fixture wallet and verify snapshot. Then realize tx and when, both pending/in_ledger, verify utxo snapshots for both source and target wallet 
2. Do some funds flow testing and compare expectations of utxo snapshot and utxo statistics
3. Do something interesting with non-ada things and check like in point 1. And like point 2 if we have already utxo statistics for native assets

lib/core/src/Cardano/Wallet/Api/Link.hs
```diff
@@ -237,6 +238,21 @@ getUTxOsStatistics w = discriminate @style
...
+        ( HasCallStack
+        , Discriminate style
+        , HasType (ApiT WalletId) w
+        )
+    => w
+    -> (Method, Text)
+getWalletUtxoSnapshot w = discriminate @style
+    (endpoint @Api.GetWalletUtxoSnapshot (wid &))
+    (endpoint @Api.GetByronWalletUtxoSnapshot (wid &))
+    (notSupported "Shared")
```
rvl, line 252: Why not supported for shared wallets? They are have UTxO, once active.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -697,6 +699,20 @@ newtype ApiWalletPassphrase = ApiWalletPassphrase
     } deriving (Eq, Generic, Show)
       deriving anyclass NFData
 
+newtype ApiWalletUtxoSnapshot = ApiWalletUtxoSnapshot
```
rvl, line 702: I think it's possible to avoid the extra type by just using `ApiT [ApiWalletUtxoSnapshotEntry]`.

specifications/api/swagger.yaml
```diff
@@ -2255,6 +2255,39 @@ components:
...
+          <<: *walletAssets
+          description: |
+            The set of non-ada assets associated with this UTxO entry.
+
+    ApiWalletUtxoSnapshot: &ApiWalletUtxoSnapshot
+      type: object
+      required:
+        - entries
+      properties:
+        entries:
```
rvl, line 2284: The extra level of wrapping in an object isn't necessary, is it?

specifications/api/swagger.yaml
```diff
@@ -4437,6 +4479,19 @@ paths:
         - *parametersWalletId
       responses: *responsesGetUTxOsStatistics
 
+  /wallets/{walletId}/utxo:
+    get:
+      operationId: getWalletUtxoSnapshot
+      tags: ["Wallets"]
+      summary: A snapshot of the wallet's UTxO set
+      description: |
+        Generate a snapshot of the wallet's UTxO set.
```
rvl, line 4488: Most endpoints have a `status: stable` note.
Since this is intended to be used for debugging, better to not imply stability - perhaps add something like `status: unstable`.

rvl (commented):
Nice addition, thanks :+1:

lib/core/src/Cardano/Wallet/Api/Link.hs
jonathanknowles: > Why not supported for shared wallets? They are have UTxO, once active.

Good point.

@piotr-iohk, do you know of a good reason to not support shared wallets?

specifications/api/swagger.yaml
jonathanknowles: > The extra level of wrapping in an object isn't necessary, is it?

I think this is arguable either way.

I agree with you that using a list would be enough to capture the current use case.

But on the other hand, using an object gives us the flexibility to add extra fields relating to a snapshot at a later time, without having to break compatibility with clients of the API (including ourselves).

From an API aesthetics point of view, I feel it's clearer to be able to point to a single type and say "this single type represents a complete UTxO snapshot", similarly to how we use `UTxO` to represent a complete UTxO set.

specifications/api/swagger.yaml
jonathanknowles: > Most endpoints have a status: stable note.
> Since this is intended to be used for debugging, better to not imply stability - perhaps add
> something like status: unstable.

Good point!

lib/core/src/Cardano/Wallet/Api/Link.hs
piotr-iohk: Nope, my guess was that we will support them eventually at some point (once they are active). :)
If it is possible to wire them up right now then, yes, why not.

PR #2681, opened 2021-05-31
Wait longer for first tip in restore-bench

Anviking:
# Issue Number

ADP-846

# Overview

- [x] Wait longer for first node tip in restore bench 

# Comments

Attempt to fix this timeout:

https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/964#fb8228aa-158e-4517-b5bb-883e2be30b24

```
[bench-restore:Notice:15] [2021-05-31 01:01:17.09 UTC] [cardano-node.66880] withBackend action done. Terminating child process
restore: UnliftIO.Exception.throwString called with:

Gave up in waitForNodeSync, waiting a tip
Called from:
  throwString (bench/Restore.hs:777:18 in main:Main)
```

It might just be the node re-validating its entire state, taking a lot
of time. In this case, waiting longer should help.

### Update

 https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/966#32b592d5-567d-41b9-91bb-d99d89209506

After a while we start seeing `Initial node synchronization: still restoring (97.04%)`, so it seems to work.

### DIFF
--- a/lib/shelley/bench/Restore.hs
+++ b/lib/shelley/bench/Restore.hs
@@ -754,7 +754,7 @@ waitForNodeSync
     => Tracer IO (BenchmarkLog n)
     -> NetworkLayer IO Block
     -> IO SlotNo
-waitForNodeSync tr nw = loop 120 -- allow 30 minutes for first tip
+waitForNodeSync tr nw = loop 480 -- allow 120 minutes for first tip
   where
     loop :: Int -> IO SlotNo
     loop retries = do


piotr-iohk (approved):
:crossed_fingers:

ghost:
Bro you are my home by!;I don't know how that wallets on me something to do
with, it MEW cx does that me I can re claim them?.

Anviking:
Fixed this error, but the previous heap exhausted error remains.

rvl (commented):
Thanks!

Two hours is quite a while. :grimacing: 

But usually the chain should be cached on the benchmark machine between runs.

And we don't want to start the benchmark before cardano-node is fully synced, because it will influence the measurements.

Anviking:
> Two hours is quite a while.
>
> But usually the chain should be cached on the benchmark machine between runs

Yes. We should mostly wait for _either_ the ledger-validation, or full synchronisation, but never both.

PR #2683, opened 2021-05-31
Correct error method when hardened index wrong

paweljakubas:
# Issue Number

ADP-927

# Overview

- [x] Correcting hardened index error

# Comments


piotr-iohk (approved):
lgtm

PR #2684, opened 2021-05-31
Add properties for listing stake keys

Anviking:
# Issue Number

ADP-721

# Overview

- [x] Test `no rewards => sum of all keys' stake == utxo balance"`
- [x] Test `ours and foreign keys are disjoint`
- [x] ~Rename `getAccountBalance` to `getCachedAccountBalance` and add un-cached `fetchAccountBalance`~ -> done in #2694 instead 
- [x] ~Make `listStakeKeys` call `fetchAccountBalance` instead of `getCachedAccountBalance`~ -> done in #2694 instead
- [x] Improve integration testing of the rewardBalance field
- [x] Tweak API docs   

# Comments


lib/core/test/unit/Cardano/Wallet/Api/ServerSpec.hs
```diff
@@ -166,7 +218,7 @@ networkInfoSpec = describe "getNetworkInformation" $ do
     it "doesn't return 500 when the time interpreter horizon is behind\
        \ the current time" $ property $ \(gap' ::(NonNegative Int)) ->
         monadicIO $ do
-        let gap = fromRational $ toRational $ getNonNegative gap'
+        let (gap :: NominalDiffTime) = fromRational $ toRational $ getNonNegative gap'
```
Anviking, line 222: Should be because of my added `NoMonomorphismRestriction `

lib/core/test/unit/Cardano/Wallet/Api/ServerSpec.hs
```diff
@@ -265,3 +317,116 @@ errorHandlingSpec = describe "liftHandler and toServerError" $ do
...
+    in
+        -- Leveraging the fact that addresses from genUTxOIndexSmall are
+        -- "ADDR{0 .. 7}"
+        if char == '7'
+        then Nothing
+        else Just $ RewardAccount $ B8.pack [char]
+
+instance Arbitrary UTxO where
+    arbitrary = toUTxO <$> genUTxOIndexSmall
+    shrink = map toUTxO . shrinkUTxOIndexSmall . fromUTxO
```
jonathanknowles: Hi @Anviking 

I'm guessing that there's quite a bit of unnecessary repeated re-indexing going on here:
- `shrinkUTxOIndexSmall` first unwraps the index, shrinks the entries, and then recreates the index.
- `shrink` (defined above) again first calls `fromUTxO` to recreate the index, then calls `shrinkUTxOIndexSmall` (which reindexes), only to throw away the index again in the final step.

But the properties defined in this PR don't seem to make use of UTxO indexing functionality.

Perhaps we could avoid the indexing steps entirely?

I'm guessing that one way to do this would be to extract out the UTxO-generation part of `genUTxOIndexSmall` to a new function `UTxO.Gen.genUTxOSmall`, which `genUTxOIndexSmall` could reuse. And of course, the same for shrinkers.

What do you think?

lib/core/test/unit/Cardano/Wallet/Api/ServerSpec.hs
```diff
@@ -265,3 +317,116 @@ errorHandlingSpec = describe "liftHandler and toServerError" $ do
...
+    labeledCheck notF f a b = counterexample
+        (show a <> " " <> notF <> " " <> show b)
+        (property $ f a b)
+
+    disjoint = labeledCheck "is not disjoint from" Set.disjoint
+
+    keys = Set.fromList . map (view #_key)
+
+prop_listStakeKeysBalance
+    :: UTxO
```
jonathanknowles:
```suggestion
    :: Blind UTxO
```
I recall that the output of `show` for values of type `UTxO` can be quite verbose and hard-to-read, particularly if there are multiple UTxO entries.

Perhaps you could use something like [`pShow`](https://hackage.haskell.org/package/pretty-simple-4.0.0.0/docs/Text-Pretty-Simple.html#v:pShow) to render them in a clearer way on failure (perhaps as a counterexample string).

lib/core/test/unit/Cardano/Wallet/Api/ServerSpec.hs
```diff
@@ -265,3 +317,116 @@ errorHandlingSpec = describe "liftHandler and toServerError" $ do
...
+    in
+       cover 35 isNonTrivial "non-trivial" $
+            ((totalStake _ours) +
+                (totalStake _foreign) +
+                (totalStake [_none]))
+                ===
+            getQuantity (coinToQuantity (TokenBundle.getCoin (balance utxo)))
+  where
+    keys = Set.fromList . map (view #_key)
+    totalStake = sum . map (getQuantity . view (#_stake))
```
jonathanknowles: I'm just guessing, but perhaps it's possible to use `F.foldMap` here, if the elements are monoidal. (Though I haven't tried it.)

```suggestion
    totalStake = F.foldMap (getQuantity . view (#_stake))
```

lib/core/test/unit/Cardano/Wallet/Api/ServerSpec.hs
```diff
@@ -1,9 +1,17 @@
...
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE NoMonomorphismRestriction #-}
 {-# LANGUAGE NumericUnderscores #-}
+{-# LANGUAGE OverloadedLabels #-}
 {-# LANGUAGE Rank2Types #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
jonathanknowles, line 15: I'm looking forward to the day that all these commonly-used language extensions get rolled into a new Haskell standard. 😆

jonathanknowles (commented):
Hi @Anviking.

I had a quick look, and left some comments. I can have another look later if you need a full review.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -57,6 +57,7 @@ module Cardano.Wallet.Api.Server
     , getTransaction
     , listWallets
     , listStakeKeys
+    , listStakeKeys'
```
rvl, line 61: I think some of the difficulties might be caused by how and where `listStakeKeys` and `listStakeKeys'` are split.
If it's possible to pass values, do that, rather than passing functions.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2043,68 +2043,66 @@ quitStakePool ctx (ApiT wid) body = do
...
-                , _key = (ApiT acc, Proxy)
-                , _rewardBalance = coinToQuantity $
-                    rewards acc
-                , _delegation = deleg
-                , _stake = coinToQuantity $
-                    stake (Just acc) <> rewards acc
-                }
+    -> [(RewardAccount, ApiWalletDelegation)]
+        -- ^ Delegation status
+    -> [(RewardAccount, Coin)]
```
Anviking: Is taking 
```
-> [(RewardAccount, Natural)]
-> [(RewardAccount, ApiWalletDelegation)]
-> [(RewardAccount, Coin)]
```
which all have to match really simpler?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2043,68 +2043,66 @@ quitStakePool ctx (ApiT wid) body = do
...
 listStakeKeys'
-    :: forall (n :: NetworkDiscriminant) m. Monad m
-    => UTxO.UTxO
-        -- ^ The wallet's UTxO
-    -> (Address -> Maybe RewardAccount)
+    :: forall (n :: NetworkDiscriminant)
+     . (Address -> Maybe RewardAccount)
         -- ^ Lookup reward account of addr
-    -> (Set RewardAccount -> m (Map RewardAccount (Maybe Coin)))
-        -- ^ Batch fetch of rewards
```
Anviking: While `[(RewardAccount, Coin)]` is slightly less verbose than this, it does move responsibility of "which accounts should we fetch rewards for" to the caller. I wanted to avoid this, as `listStakeKeys'` might discover new "foreign" stake keys which might as well fetch rewards for.

(Optimal behaviour here could be discussed though)

lib/core/test/unit/Cardano/Wallet/Api/ServerSpec.hs
Anviking: I recall trying it, but the problem was that they weren't `Monoidal`. Maybe if one converted back to `Coin` rather than `Quantity` though...

jonathanknowles: I guess you could use `Semigroup.Sum` (which is an instance of `Semigroup`).

So:
```hs
    totalStake = getSum . F.foldMap (Sum . getQuantity . view (#_stake))
```

This would have the small advantage of only traversing the list once.

But it's probably not worth it for this small traversal 😄.

lib/core/test/unit/Cardano/Wallet/Api/ServerSpec.hs
Anviking: Yeah, good idea, thanks for the exact link.

A `PrettyShow` wrapper seems better to me though -> 580b801c601765522f1cdd04adb2159c41897935

lib/core/test/unit/Cardano/Wallet/Api/ServerSpec.hs
Anviking: 2d087558c27a2004c65b1cef7426ec3dd2255785 

I was worried there might be a performance hit to the UTxOIndex generators (Instead of a `[(in,out)] -> UTxOIndex` transformation, we now have `[(in, out)] -> UTxO -> UTxOIndex`), but from trying RoundRobinSpec in ghci very briefly, I think we might be fine.

lib/core/src/Cardano/Wallet/Api/Server.hs
rvl: This was just a draft that I sketched and accidentally pushed to the wrong branch.
It would be better to have a record type or n-tuple.

lib/core/src/Cardano/Wallet/Api/Server.hs
rvl: I suggest to start simple, then add complexity later as needed.
It will make things easier for both yourself and the people who have to review the work.
If a monadic callback will be needed in the future to support foreign stake keys, then add it future.
Until then, make it as simple as possible - Value parameters instead of monadic function parameters.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2127,14 +2133,26 @@ listStakeKeys lookupStakeRef ctx (ApiT wid) = do
...
+
+            -- With ADP-808 we will be able to have many keys. Currently there's
+            -- only one.
             let ourKeys = case mourAccount of
                     Just acc -> [(acc, 0, ourApiDelegation)]
                     Nothing -> []
 
+            -- This uses `getAccountBalance` under the hood. This should give us
+            -- automatic batching of requests, but at the expense of some
+            -- results potentially being zero until listing stake keys once
```
Anviking: I had forgotten `getAccountBalance` worked like this just before. (will only return cached values and never block while fetching)

It seems like something that should be fixed, but we have the same behaviour for calculating the wallet balance — we only include the reward balance if we have fetched them before, and afterwards we refetch on every tip-change.

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: @rvl thanks a lot for having a look at this. I made some attempts to improve the in-code comments.

To my eyes, the monadic callback is not something needed for the future. Foreign stake keys in the UTxO with rewards could exist at this very moment (though always unlikely). And they are not known until the `listStakeKeys'` helper calculate them.

I also think it makes some conceptual sense to separate _how to fetch_ from _what to fetch_.

What I said in DM was that _ourKeys_ is currently just this:
```haskell
          let ourKeys = case mourAccount of
                   Just acc -> [(acc, 0, ourApiDelegation)]
                   Nothing -> []
```
and will have more entries in the future. 

That said, I want to have a closer look at https://github.com/input-output-hk/cardano-wallet/pull/2684#discussion_r643069025 (or maybe you have immediate opinion?), which might affect things slightly, (in particular the need of the `lookupUsing`/ `Set -> Map` indirection), if we want to guarantee up-to-date account reward balances.

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: Should be fixed with 18b36d071eab781397ae2b2a82a6bd25b4df602d

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -355,8 +355,10 @@ withNetworkLayerBase tr np conn (versionData, _) tol action = do
             _postTx localTxSubmissionQ era sealed
         , stakeDistribution =
             _stakeDistribution queryRewardQ
-        , getAccountBalance =
+        , getCachedAccountBalance =
             _getAccountBalance rewardsObserver
+        , fetchRewards =
+            fetchRewardAccounts tr queryRewardQ
```
Anviking: I guess it should be `fetchAccountBalance` for consistency…

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: @rvl on the topic of making things as simple as possible "until then" — I switched the function to actually use an un-cached, but batched `fetchAccountBalances`, which I think makes things look more logical, and further justifies the use of `(Set RewardAccount -> m (Map RewardAccount Coin))`:

Skärmavbild 2021-06-03 kl  15 58 30

lib/core/test/unit/Cardano/Wallet/Api/ServerSpec.hs
Anviking: I ended up using `Coin` now.

Anviking:
@rvl @jonathanknowles could any of you you have another look? 🙏 

Happy to have a quick call about it if appropriate.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxO/Gen.hs
```diff
@@ -0,0 +1,79 @@
...
+--------------------------------------------------------------------------------
+-- Small indices
+--------------------------------------------------------------------------------
+
+genUTxOSmall :: Gen UTxO
+genUTxOSmall = do
+    entryCount <- frequency
+        [ (1, pure 0)
+        , (1, pure 1)
+        , (32, choose (2, 64))
```
rvl, line 39: Yes, this is what the QuickCheck `size` parameter is for.

lib/core/test/unit/Cardano/Wallet/Api/ServerSpec.hs
```diff
@@ -166,7 +225,7 @@ networkInfoSpec = describe "getNetworkInformation" $ do
     it "doesn't return 500 when the time interpreter horizon is behind\
        \ the current time" $ property $ \(gap' ::(NonNegative Int)) ->
         monadicIO $ do
-        let gap = fromRational $ toRational $ getNonNegative gap'
+        let (gap :: NominalDiffTime) = fromRational $ toRational $ getNonNegative gap'
```
rvl, line 222: OK, but isn't this the more normal way to write it?
```suggestion
        let gap = fromRational $ toRational $ getNonNegative gap' :: NominalDiffTime
```

lib/core/test/unit/Cardano/Wallet/Api/ServerSpec.hs
```diff
@@ -265,3 +324,140 @@ errorHandlingSpec = describe "liftHandler and toServerError" $ do
...
+        it "ours and foreign keys are disjoint" $ do
+            property prop_listStakeKeysDisjoint
+
+prop_listStakeKeysDisjoint
+    :: PrettyShow UTxO
+    -> SaneOurStakeKeysInput
+    -> (Map RewardAccount Coin)
+    -> Property
+prop_listStakeKeysDisjoint (PrettyShow utxo) (SaneOurStakeKeysInput ours) m =
+    let
```
rvl, line 336: Since there's already a `where`, you don't need `let`.

lib/core/test/unit/Cardano/Wallet/Api/ServerSpec.hs
```diff
@@ -265,3 +324,140 @@ errorHandlingSpec = describe "liftHandler and toServerError" $ do
...
+                , totalStake [_none]
+                ]
+                ===
+            TokenBundle.getCoin (balance utxo)
+
+-- | Type for where the reward accounts and indices are unique.
+--
+-- This is required for some properties, and does not seem worthwhile to protect
+-- against.
+newtype SaneOurStakeKeysInput
```
rvl, line 372: If unique accounts and indices are a precondition of `listStakeKeys'`, then perhaps this wrapper type should be used for the `ourKeys` parameter of that function?

But on the other hand, the `ourKeys` list is always of length 1, so unique by definition...

lib/core/test/unit/Cardano/Wallet/Api/ServerSpec.hs
```diff
@@ -265,3 +324,140 @@ errorHandlingSpec = describe "liftHandler and toServerError" $ do
...
+
+accountOfAddress :: Address -> Maybe RewardAccount
+accountOfAddress (Address bytes) =
+    let
+        str = B8.unpack bytes
+        char = last str
+    in
+        -- Leveraging the fact that addresses from genUTxOIndexSmall are
+        -- "ADDR{0 .. 7}"
+        if char == '7'
```
rvl, line 423: If you're looking for something less fragile, then the address→account mapping could be stored in the generated test case.

lib/core/test/unit/Cardano/Wallet/Api/ServerSpec.hs
```diff
@@ -265,3 +324,140 @@ errorHandlingSpec = describe "liftHandler and toServerError" $ do
...
+--
+-- The UTxO will contain stake keys ['0' .. '6'], so we want this instance to
+-- generate both elements from and outside this set.
+instance Arbitrary RewardAccount where
+    arbitrary = RewardAccount . B8.singleton
+        <$> elements (['0' .. '7'] ++ ['a' .. 'z'])
+
+-- Helpers
+
+newtype PrettyShow a = PrettyShow a
```
rvl: I put this into the `Test.Utils.Pretty` module in #2708.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2152,24 +2157,27 @@ listStakeKeys
...
+        -- With ADP-808 we will be able to have many keys. Currently there's
+        -- only one.
+        let ourKeys = case mourAccount of
+                Just acc -> [(acc, 0, ourApiDelegation)]
+                Nothing -> []
+
+        liftIO $ listStakeKeys' @n
+            utxo
+            lookupStakeRef
+            (fetchRewardAccountBalances nl)
```
rvl, line 2179: I still don't like this way of splitting it because it means you still have a big function that mixes pure and monadic code.

lib/core/test/unit/Cardano/Wallet/Api/ServerSpec.hs
Anviking: > but isn't this the more normal way to write it?

I have no idea, is it?

But to my eyes:
1. I think it reads more natural / I don't have a sense for the precedence of the `::`
2. When using a monadic bind, you'd also have to specify the type of the monad, which you might not want to:
```haskell
a <- getEnv "CI" :: IO String

:set -XScopedTypeVariables
a :: String <- getEnv "CI" 
```

lib/core/test/unit/Cardano/Wallet/Api/ServerSpec.hs
Anviking: For readability I think it makes sense to have `let .. in ..` be meant to be read sequentially, and keep auxiliary helpers in `where` clauses. The `res` is very important, but `disjoint` is not.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxO/Gen.hs
Anviking: Yup, felt like that (I don't want to change anything now though, as this is merely mirroring `UTxOIndex.Gen`)

rvl: Sure. It's something we should fix though, so I tried fixing it... but found that there is a long chain of small/large variants which need to be changed. And so I gave up. :frowning_face:

lib/core/test/unit/Cardano/Wallet/Api/ServerSpec.hs
rvl: Well, that syntax requires the `ScopedTypedVariables` extension, so it's less normal than the normal way.
All you need to remember about `::` type annotations is that no operator has higher precedence.
The extension syntax is useful for `<-` but this here is a let expression.
There's nothing incorrect about it, but it just looked strange to me.

Anviking:
Thanks for the comments, closing temporarily while the task is on hold.

PR #2687, opened 2021-06-02
Fix link to list stake keys in swagger

piotr-iohk:
# Issue Number

# Overview

- 6e5bab93ac9fe161216489e3c2235caf3dad6f36
  Fix link to list stake keys in swagger

# Comments

### DIFF
--- a/specifications/api/swagger.yaml
+++ b/specifications/api/swagger.yaml
@@ -4717,9 +4717,9 @@ paths:
 
   /wallets/{walletId}/stake-keys:
     get:
-      operationId: listStakePools
+      operationId: listStakeKeys
       tags: ["Stake Pools"]
-      summary: List stake keys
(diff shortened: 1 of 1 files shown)


Anviking (approved):
Ah, thanks

Anviking:
Looking now, maybe "list stake keys" should be in the "Shelley wallets" section? 🤔 

Instead of here

Skärmavbild 2021-06-02 kl  11 17 51

piotr-iohk:
> Looking now, maybe "list stake keys" should be in the "Shelley wallets" section?

Could be... I kind of like it in Stake Pools cause it is related to delegation in the end. Similarly delegation fees could be also in the "Shelley wallets" section... Perhaps in future it'll deserve a separate section, like Multi-delegation when more eps arrive.

PR #2688, opened 2021-06-03
Restoration bench: increase heap and only bench {0%,0.2%} x {seq,rnd}

Anviking:
# Issue Number

ADP-846

# Overview

Let the nightly benchmark finish successfully within the alotted time.

- [x] Increase max heap from 8GB to 16GB
- [x] Replace 0.1% and 0.4% rnd and seq benchmarks with 0.2% to prevent timeout.

# Comments

[Build status](https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds?branch=anviking%2FADP-846%2Fmore-heap)

### DIFF
--- a/lib/shelley/bench/Restore.hs
+++ b/lib/shelley/bench/Restore.hs
@@ -313,7 +313,8 @@ cardanoRestoreBench tr c socketFile = do
         , benchRestoreSeqWithOwnership (Proxy @4)
 
         , benchRestoreRndWithOwnership (Proxy @0)
-        , benchRestoreRndWithOwnership (Proxy @1)
+        -- benchRestoreRndWithOwnership (Proxy @1)
+        -- Removed to prevent timeout in CI.
         , benchRestoreRndWithOwnership (Proxy @4)
         ]
   where
--- a/lib/shelley/bench/Restore.hs
+++ b/lib/shelley/bench/Restore.hs
@@ -309,13 +309,10 @@ cardanoRestoreBench tr c socketFile = do
         , benchRestoreMultipleWallets 100 (unsafeMkPercentage 0.01)
 
         , benchRestoreSeqWithOwnership (Proxy @0)
-        , benchRestoreSeqWithOwnership (Proxy @1)
-        , benchRestoreSeqWithOwnership (Proxy @4)
+        , benchRestoreSeqWithOwnership (Proxy @2)
 
         , benchRestoreRndWithOwnership (Proxy @0)
-        -- benchRestoreRndWithOwnership (Proxy @1)
-        -- Removed to prevent timeout in CI.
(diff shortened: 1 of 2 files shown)


lib/shelley/bench/Restore.hs
```diff
@@ -313,7 +313,8 @@ cardanoRestoreBench tr c socketFile = do
         , benchRestoreSeqWithOwnership (Proxy @4)
 
         , benchRestoreRndWithOwnership (Proxy @0)
-        , benchRestoreRndWithOwnership (Proxy @1)
+        -- benchRestoreRndWithOwnership (Proxy @1)
+        -- Removed to prevent timeout in CI.
```
Anviking: Keeping it commented to prevent someone from thinking the absence of it is a mistake.

And I kept 0.4% rnd, because I recall it being similar to some exchanges (but I don't really know).

Anviking:
Looking at the current progress, removing 0.1%-rnd will most certainly not be enough.

[Before removing anything](https://buildkite.com/organizations/input-output-hk/pipelines/cardano-wallet-nightly/builds/969/jobs/6d0e4f13-5797-439e-a081-498218447136/raw_log), it started 0.4%-rnd 30 minutes before the timeout. So even just removing 0.4%-rnd will be tight. 

I suspect we might need to remove both 0.4%-seq and 0.4%-rnd…

Anviking:
```
list transactions: 2946 s
```
From 0.4%-seq [src](https://buildkite.com/organizations/input-output-hk/pipelines/cardano-wallet-nightly/builds/969/jobs/6d0e4f13-5797-439e-a081-498218447136/raw_log)

😳

lib/shelley/bench/Restore.hs
```diff
@@ -309,12 +309,10 @@ cardanoRestoreBench tr c socketFile = do
...
 
         , benchRestoreSeqWithOwnership (Proxy @0)
-        , benchRestoreSeqWithOwnership (Proxy @1)
-        , benchRestoreSeqWithOwnership (Proxy @4)
+        , benchRestoreSeqWithOwnership (Proxy @2)
 
         , benchRestoreRndWithOwnership (Proxy @0)
-        , benchRestoreRndWithOwnership (Proxy @1)
-        , benchRestoreRndWithOwnership (Proxy @4)
+        , benchRestoreRndWithOwnership (Proxy @2)
```
rvl, line 315: It's a little bit confusing to have the address discovery ratio as a unitless type parameter, but now I see this means 0% and 0.2%.

Probably just one or two percentage benchmarks is enough anyway.

Next thing we should do is improve the performance so we can increase these numbers. :smile:.

And also to benchmark the sequential address pool code in isolation.

rvl (approved):
Thanks for fixing it - looks good to me.

PR #2693, opened 2021-06-07
Allow specifying purpose for acc x pub

paweljakubas:
# Issue Number

adp-950

# Overview

- [x] updated swagger
- [x] enable passing purpose
- [x] adjust core unit tests
- [x] add integration test
- [x] guard purpose with integration test  

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -449,7 +449,7 @@ import Data.List
 import Data.List.NonEmpty
     ( NonEmpty (..) )
 import Data.Maybe
-    ( fromJust, fromMaybe, isJust, mapMaybe )
+    ( fromMaybe, mapMaybe )
 import Data.Proxy
     ( Proxy )
 import Data.Quantity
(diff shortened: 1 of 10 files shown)


lib/core/src/Cardano/Wallet.hs
```diff
@@ -2332,16 +2335,23 @@ readAccountPublicKey ctx wid = db & \DBLayer{..} -> do
...
-getAccountPublicKeyAtIndex ctx wid pwd ix = db & \DBLayer{..} -> do
-    acctIx <- withExceptT ErrReadAccountPublicKeyInvalidIndex $ guardHardIndex ix
+getAccountPublicKeyAtIndex ctx wid pwd ix purposeM = db & \DBLayer{..} -> do
+    acctIx <- withExceptT ErrReadAccountPublicKeyInvalidAccountIndex $ guardHardIndex ix
+
+    when (isJust purposeM) $ do
+        purposeGuarded <- runExceptT $ guardHardIndex (fromJust purposeM)
+        case purposeGuarded of
+            Left err -> throwE $ ErrReadAccountPublicKeyInvalidPurposeIndex err
+            Right _ -> pure ()
```
rvl: Something like this can avoid the need for isJust/fromJust/runExceptT/throwE.
```suggestion
    maybe (pure ()) (withExceptT ErrReadAccountPublicKeyInvalidPurposeIndex . guardHardIndex) purpose
```

specifications/api/swagger.yaml
```diff
@@ -4666,7 +4676,8 @@ paths:
       description: |
         <p align="right">status: <strong>stable</strong></p>
         Derive an account public key for any account index. For this key derivation to be possible,
-        the wallet must have been created from mnemonic.
+        the wallet must have been created from mnemonic. In request body, arbitrary purpose can be specified.
+        Omitting the purpose segment means default purpose, ie., 1852H, will be used.
```
```suggestion
        the wallet must have been created from mnemonic.
        
        It is possible to use the optional `purpose` field to override that branch of the derivation path
        with different hardened derivation index. If that field is omitted, the default purpose
        for Cardano wallets (`1852H`) will be used.
```

rvl (approved):
Looks good - with 2 requested changes. Thanks

lib/core/src/Cardano/Wallet.hs
paweljakubas: indeed!

PR #2694, opened 2021-06-08
Factor `getAccountBalance` into `getCachedAccountBalance` and `fetchAccountBalances`

Anviking:
# Issue Number

Split off from #2684 

# Overview

- [x] Rename `getAccountBalance` to `getCachedAccountBalance` for clarity
- [x] Add `fetchAccountBalances` function for un-cached behaviour
- [x] `fetchAccountBalances` from `listStakeKeys`  

# Comments

### DIFF
--- a/lib/core/src/Cardano/Wallet/Network.hs
+++ b/lib/core/src/Cardano/Wallet/Network.hs
@@ -175,7 +175,7 @@ data NetworkLayer m block = NetworkLayer
         :: Coin -- Stake to consider for rewards
         -> m StakePoolsSummary
 
-    , getCachedAccountBalance
+    , getCachedRewardAccountBalance
         :: RewardAccount
         -> m Coin
         -- ^ Return the cached reward balance of an account.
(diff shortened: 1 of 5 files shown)


lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -172,9 +175,20 @@ data NetworkLayer m block = NetworkLayer
...
         :: RewardAccount
-        -> ExceptT ErrGetAccountBalance m Coin
+        -> m Coin
+        -- ^ Return the cached reward balance of an account.
+        --
+        -- If there is no cached value, it will return `Coin 0`, and add the
+        -- account to the internal set of observed account, such that it will be
+        -- fetched later.
+
+    , fetchAccountBalances
```
paweljakubas: occasionally I think that  `fetchRewardAccountBalances` and `getCachedRewardAccountBalance` would be better here  in spite of being long...

paweljakubas (approved):
I like this decomposition. If we have a proof that it helps with flaky integration testing then it is great

Anviking:
Thanks for having a look.

> If we have a proof that it helps with flaky integration testing then it is great

This wasn't intended to fix flaky tests, although that's a good point, this caching-behaviour could be involved 🤔 

(This PR only changes the behaviour of listStakeKeys. Wallet reward balances still use the cached version.)

lib/core/src/Cardano/Wallet/Network.hs
Anviking: 👍 fair -> d81089b8bb205af2f082d0704eac7410aba1c541

PR #2696, opened 2021-06-09
Timeout earlier on SQLITE_BUSY when deleting wallets in tests

Anviking:
# Issue Number

ADP-827

# Overview

- [x] Set `CARDANO_WALLET_INTEGRATION` env var in integration tests
- [x] Change timeout to close all connections from `60s` to `2s` if `CARDANO_WALLET_INTEGRATION` is set.

# Comments

- Workaround that should make CI faster (and perhaps a bit more stable), until we have time to investigate and address the underlying issue and its effects in production. The problem in production is likely much less serious, as users likely don't create and delete wallets as often as in our integration tests.
- Locally, running `stack test cardano-wallet:integration --ta '-j 6'` I get:
    - 2s -> 335s
    - 5s -> 380s
    - 60s (without the `setEnv`) -> 1190s

So that's why I went with 2s rather than 5s.

### DIFF
--- a/lib/core/src/Cardano/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/DB/Sqlite.hs
@@ -78,6 +78,8 @@ import Data.Aeson
     ( ToJSON (..) )
 import Data.Function
     ( (&) )
+import Data.Functor
+    ( (<&>) )
 import Data.List
     ( isInfixOf )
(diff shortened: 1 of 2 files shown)


rvl (approved):
Thanks @Anviking - this should hopefully improve the integration tests.

Issue #2699, opened 2021-06-10
Failing migration tests (e.g. SHELLEY_MIGRATE_08)


Anviking:
# Context
 
 

# Test Case

```haskell
        liftIO $ migrateWallet ctx sourceWallet targetAddressIds

        -- Check that funds become available in the target wallet.
        let expectedTargetBalance =
                sourceBalance - balanceLeftover - expectedFee
        eventually "Target wallet balance reaches expected balance" $ do
            response <- request @ApiWallet ctx
                (Link.getWallet @'Shelley targetWallet) Default Empty
            verify response
                [ expectField
                    (#balance . #available . #getQuantity)
                    (`shouldBe` expectedTargetBalance)
                , expectField
                    (#balance . #total . #getQuantity)
                    (`shouldBe` expectedTargetBalance)
                ]

```

# Failure / Counter-example

```
Failures:
  src/Test/Integration/Scenario/API/Shelley/Migrations.hs:926:23:
  1) API Specifications, SHELLEY_MIGRATIONS, SHELLEY_MIGRATE_08 - It's possible to migrate a wallet with many small ada quantities, provided that the total balance is significantly greater than the minimum ada quantity for an output.
       While verifying (Status {statusCode = 200, statusMessage = "OK"},Right (ApiUtxoStatistics {total = Quantity {getQuantity = 0}, scale = ApiT {getApiT = Log10}, distribution = fromList [(10,0),(100,0),(1000,0),(10000,0),(100000,0),(1000000,0),(10000000,0),(100000000,0),(1000000000,0),(10000000000,0),(100000000000,0),(1000000000000,0),(10000000000000,0),(100000000000000,0),(1000000000000000,0),(10000000000000000,0),(45000000000000000,0)]}))
       expected: [(100000000,1)]
        but got: []
```

---

# Resolution 

We use `eventually` and the normal, nonretrying `Hspec.it` here. 

This means that the condition of `eventually` can be satisfied, but then rolled back. 

We should either use:
1. `waitForTxImmutability`
2. A cluster without forks
3. Retry test on failures

---

# QA

rvl:
What if we created an API endpoint that could pause/resume syncing of a wallet?

Once the `eventually` condition is met, pause chain updates, check the condition a second time, run the block of assertions, then resume the wallet.

Anviking:
Should be fixed.

PR #2700, opened 2021-06-10
Prefer waiting over eventually in migration tests

Anviking:
# Issue Number

ADP-970, #2699 

# Overview

- [x] `waitForTxImmutability` should now only need to be 10s instead of 16s
- [x]  Use `waitForTxImmutability` instead of `eventually` to protect against rollbacks in migration tests

# Comments

- An alternative solution would be to use the retrying `it` with `eventually`, but we should moving away from the retrying it, not re-adding it.

### DIFF
--- a/lib/core-integration/src/Test/Integration/Framework/DSL.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/DSL.hs
@@ -945,7 +945,7 @@ waitForTxImmutability _ctx = liftIO $ do
     --
     -- let stabilityDelay   = round (sl * 3 * k / f)
     -- let txInsertionDelay = round (sl * 10 / f)
-    let stabilityDelay   = 12 * oneSecond
+    let stabilityDelay   = 6 * oneSecond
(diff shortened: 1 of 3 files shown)


Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

rvl (commented):
So the reason for this change is that between checking one wallet condition and the next, the chain rolls back and invalidates the first condition, such that the second condition fails?

Before making a solution, do we have logging which proves conclusively that this is the case?

And what about other ways to solve this? We need some methods to make our tests less fragile - otherwise this flakiness will keep on happening.

My two ideas so far are:
1. Provide API endpoints to pause and resume syncing for a wallet - preventing rollback while assertions are being checked.
2. Provide a way for assertions to be contingent on a certain TxId having status=InLedger. If an assertion fails, and the Tx has gone back to Pending, then restart the `eventually` block from the beginning.

rvl (approved):
Why not

PR #2702, opened 2021-06-10
Payment construction endpoint

paweljakubas:
# Issue Number

adp-909

# Overview

- [x] POST transactions/construct in swagger
- [x] Api.Types reflecting things added in swagger
- [x] add Api skeleton
- [x] unit tests for Api.Types, Malformed
- [x] extend Transaction interface by adding mkUnsignedTransaction
- [x] fill in implementation of Cardano.Api.Server and Cardano-Wallet function implementing 'payments' functionality
- [x] prepare first integration testing

# Comments


lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -823,6 +835,44 @@ data ByronWalletPutPassphraseData = ByronWalletPutPassphraseData
...
+data ApiMultiDelegationAction = Joining (ApiT PoolId) Natural | Leaving (ApiT PoolId)
+    deriving (Eq, Generic, Show)
+    deriving anyclass NFData
+
+data ApiConstructTransactionData (n :: NetworkDiscriminant) = ApiConstructTransactionData
+    { payments :: !(Maybe (NonEmpty (AddressAmount (ApiT Address, Proxy n))))
+    , withdrawal :: !(Maybe ApiWithdrawalPostData)
+    , metadata :: !(Maybe (ApiT TxMetadata))
+    , mint :: !(Maybe (ApiT W.TokenMap))
+    , delegation :: ![ApiMultiDelegationAction]
```
paweljakubas: @Anviking does it make sense? see `ApiMultiDelegationAction` - Natural will stand for soft index of stake key. Thanks to that we could in one tx realize several join/leave pool actions. Maybe it could help with portfolio recreation, reduce needed roundtrips, etc.

Anviking: I think `[ApiMultiDelegationAction]` as a raw/manual/low-level kind of multi-delegation api would be really nice. `Joining (ApiT PoolId) Natural` makes sense to me.

But instead of `Leaving (ApiT PoolId)`, we should presumably have `Leaving Natural`. We don't have to specify a pool id when we de-register a stake key, but we do need to know which stake key we want to deregister.

And I think we currently use the terminology `Join` and `Quit`, so why not stick to it?

Anviking: (I just realised I didn't plan for the `DelegationState` to _validate_ (several) user-provided multi-delegation actions, but it would be nice to support, and I don't see how it could be problematic.)

But it does mean that, as said, the delegation actions need to be validated by the `DelegationState` eventually.

And for now, I guess you would have to start with only supporting key `0` or maybe even a single action, when implementing it.

paweljakubas: thanks for the look. Now I have : 
```
data ApiMultiDelegationAction = Joining (ApiT PoolId) Natural | Leaving Natural
```

lib/core/src/Cardano/Wallet/Api/Link.hs
```diff
@@ -633,6 +634,21 @@ getTransaction w t = discriminate @style
...
+        ( HasCallStack
+        , HasType (ApiT WalletId) w
+        , Discriminate style
+        )
+    => w
+    -> (Method, Text)
+createUnsignedTransaction w = discriminate @style
+    (endpoint @(Api.ConstructTransaction Net) (wid &))
+    (notSupported "Byron")
+    (notSupported "Shared")
```
rvl: Perhaps add a `TODO: ADP-909` comment or something - These wallet types should be supported in the final version.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1918,6 +1927,83 @@ postTransactionFee ctx (ApiT wid) body = do
...
+        )
+    => ctx
+    -> ArgGenChange s
+    -> ApiT WalletId
+    -> ApiConstructTransactionData n
+    -> Handler (ApiConstructTransaction n)
+constructTransaction ctx genChange (ApiT wid) body = do
+    let toAddressAmount (ApiPaymentAddresses content) =
+            addressAmountToTxOut <$> content
+        toAddressAmount (ApiPaymentAll _) = undefined -- this will be tackled when migration is supported
```
rvl: Let's make sure our app doesn't implode when the user finds an implemented feature.
Can we use HTTP 501 instead of `undefined` or `error`, wherever possible?

```suggestion
    let toAddressAmount (ApiPaymentAddresses content) = pure (addressAmountToTxOut <$> content)
         toAddressAmount (ApiPaymentAll _) = liftE ErrTemporarilyDisabled -- TODO: this will be tackled when migration is supported
```

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1918,6 +1927,83 @@ postTransactionFee ctx (ApiT wid) body = do
...
+
+    (wdrl, mkRwdAcct) <-
+        mkRewardAccountBuilder @_ @s @_ @n ctx wid (body ^. #withdrawal)
+
+    ttl <- liftIO $ W.getTxExpiry ti mTTL
+    let txCtx = defaultTransactionCtx
+            { txWithdrawal = wdrl
+            , txMetadata = md
+            , txTimeToLive = ttl
+            --, txDelegationAction -- this will be tackled when delegations are supported
```
rvl:
```suggestion
            , txDelegationAction = Nothing -- TODO: ADP-909 this will be tackled when delegations are supported
```

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1918,6 +1927,83 @@ postTransactionFee ctx (ApiT wid) body = do
...
+                    $ W.selectAssetsNoOutputs @_ @s @k wrk wid w txCtx transform
+                pure (sel, sel', estMin)
+
+        blob <- liftHandler
+            $ W.constructUnsignedTransaction @_ @s @k wrk wid mkRwdAcct pwd txCtx sel
+
+        pure (mkApiCoinSelection [] Nothing md sel', fee, blob)
+
+    let txSerialized = ApiSerialisedTransaction (ApiBytesT $ convertToBase Base64 blob)
+    pure $ ApiConstructTransaction txSerialized apiSelection (Quantity $ fromIntegral fee)
```
rvl: I think it would be nicer if there were a separate:
```haskell
mkApiConstructTransaction :: sel -> fee -> Either ErrMkTx ApiConstructTransaction
```

Currently `W.constructUnsignedTransaction` is getting the reward account XPrv, but only the XPub is necessary for the TxBody, right?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -823,6 +846,90 @@ data ByronWalletPutPassphraseData = ByronWalletPutPassphraseData
...
+    deriving (Eq, Generic, Show)
+    deriving anyclass NFData
+
+-- | Input parameters for transaction construction.
+data ApiConstructTransactionData (n :: NetworkDiscriminant) = ApiConstructTransactionData
+    { payments :: !(Maybe (ApiPaymentDestination n))
+    , withdrawal :: !(Maybe ApiWithdrawalPostData)
+    , metadata :: !(Maybe (ApiT TxMetadata))
+    , mint :: !(Maybe (ApiT W.TokenMap))
+    , delegations :: !(Maybe [ApiMultiDelegationAction])
```
rvl: I forget, why did we have `Maybe [ApiMultiDelegationAction]` here?
Was there a semantic difference between `Nothing` and `Just []`?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -823,6 +846,90 @@ data ByronWalletPutPassphraseData = ByronWalletPutPassphraseData
...
+-- | New transaction API version of 'AddressAmount'.
+data AddressValue addr = AddressValue
+    { address :: !addr
+    , value :: !(ApiT W.TokenBundle)
+    } deriving (Eq, Generic, Show)
+      deriving anyclass NFData
+--}
+
+data ApiPaymentDestination (n :: NetworkDiscriminant)
+    = ApiPaymentAddresses !(NonEmpty (AddressAmount (ApiAddressIdT n)))
```
rvl, line 895: If using `NonEmpty` for `ApiPaymentAddresses`, then I suggest adding back the "None" constructor here, and removing the `Maybe` from `ApiConstructTransactionData`.
```suggestion
    = ApiPaymentNone
    --  ^ The only output is a change address.
    | ApiPaymentAddresses !(NonEmpty (AddressAmount (ApiAddressIdT n)))
```

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -823,6 +846,90 @@ data ByronWalletPutPassphraseData = ByronWalletPutPassphraseData
...
+
+data ApiConstructTransaction (n :: NetworkDiscriminant) = ApiConstructTransaction
+    { serializedTransaction :: !ApiSerialisedTransaction
+    , coinSelection :: !(ApiCoinSelection n)
+    , fee :: !(Quantity "lovelace" Natural)
+    } deriving (Eq, Generic, Show)
+      deriving anyclass NFData
+
+-- | Index of the stake key.
+newtype StakeKeyIndex = StakeKeyIndex (ApiT DerivationIndex)
```
rvl:
```suggestion
newtype ApiStakeKeyIndex = ApiStakeKeyIndex (ApiT DerivationIndex)
```

lib/core/src/Cardano/Wallet/Transaction.hs
```diff
@@ -100,6 +100,26 @@ data TransactionLayer k = TransactionLayer
         -- This expects as a first argument a mean to compute or lookup private
         -- key corresponding to a particular address.
 
+    , mkUnsignedTransaction
+        :: AnyCardanoEra
+            -- Era for which the transaction should be created.
+        -> XPrv
```
rvl: I think only an XPub is needed here, right?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -247,6 +247,32 @@ instance TxWitnessTagFor IcarusKey where
...
 
+constructUnsignedTx
+    :: forall era.
+        ( EraConstraints era
+        )
+    => Cardano.NetworkId
+    -> (Maybe Cardano.TxMetadata, [Cardano.Certificate])
+    -> SlotNo
+    -- ^ Slot at which the transaction will expire.
+    -> XPrv
```
rvl:
```suggestion
    -> RewardAccount
```

specifications/api/swagger.yaml
```diff
@@ -1006,6 +1022,14 @@ x-transactionOutputs: &transactionOutputs
       amount: *amount
       assets: *walletAssets
 
+x-transactionNoAmountOutputs: &transactionNoAmountOutputs
```
rvl, line 1025: I don't this extra declaration should be necessary.
It's just an array of addresses.

specifications/api/swagger.yaml
```diff
@@ -1088,6 +1116,35 @@ x-transactionWithdrawals: &transactionWithdrawals
...
+  type: object
+  required:
+    - action
+    - stake_key_index
+  properties:
+    action:
+      type: string
+      enum: ["quit", "join"]
+    pool: *stakePoolId
+    stake_key_index: *derivationSegment
```
rvl: What do you think of this way? I like it more.

```suggestion
  oneOf:
    - type: object
      required:
        - join
      properties:
        join:
          type: object
          required:
            - pool
            - stake_key_index
          properties:
            pool: *stakePoolId
            stake_key_index: *derivationSegment
    - type: object
      required:
        - quit
      properties:
        quit:
          type: object
          required:
            - stake_key_index
          properties:
            stake_key_index: *derivationSegment
```

specifications/api/swagger.yaml
```diff
@@ -2601,6 +2674,37 @@ components:
         passphrase: *lenientPassphrase
         address_index: *addressIndex
 
+    ApiValidityInterval: &ApiValidityInterval
+      type: object
+      required:
+        - invalid_before
+        - invalid_hereafter
```
rvl, line 2694: We would want the Haskell validity bound type to stay the same, but in JSON, the bounds should be optional, with a missing key meaning "unspecified".

rvl (commented):
Looks great @paweljakubas

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: indeed, in the end XPub is needed.

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: It is reasonable to use `delegations: Maybe ...` as it can be optional on Api level, one can omit "delegations" field.. I will use `Maybe (NonEmpty ApiMultiDelegationAction)` though

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
paweljakubas: done

lib/core/src/Cardano/Wallet/Transaction.hs
paweljakubas: indeed, used XPub rather than XPrv

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: well, we have `Maybe ApiPaymentDestination` used so optionality is captured in that way. I am slightly in favor of what it is now, but will not argue a lot;-)

specifications/api/swagger.yaml
paweljakubas: so it is used here : 
```
2686     ApiPaymentDestination: &ApiPaymentDestination                                                                                                                                                             
2687       oneOf:                                                                                                                                                                                                  
2688         - title: outputs with specified amounts                                                                                                                                                               
2689           <<: *transactionOutputs                                                                                                                                                                             
2690         - title: outputs without amounts                                                                                                                                                                      
2691           <<: *transactionNoAmountOutputs
```
and in contrast to `transactionOutputs` it does not include `amount` and `assets` as we want to use it in migration. So how to capture migration output here?

specifications/api/swagger.yaml
paweljakubas: so we have at this moment : 
```
2693     ApiConstructTransactionData: &ApiConstructTransactionData                                                                                                                                                 
2694       type: object                                                                                                                                                                                            
2695       required:                                                                                                                                                                                               
2696         - passphrase                                                                                                                                                                                          
2697       properties:                                                                                                                                                                                             
2698         passphrase:                                                                                                                                                                                           
2699           <<: *lenientPassphrase                                                                                                                                                                              
2700           description: The wallet's master passphrase.                                                                                                                                                        
2701         payments: *ApiPaymentDestination                                                                                                                                                                      
2702         withdrawal: *transactionWithdrawalRequestSelf                                                                                                                                                         
2703         metadata: *transactionMetadata                                                                                                                                                                        
2704         mint: *transactionMint                                                                                                                                                                                
2705         delegations: *transactionDelegations                                                                                                                                                                  
2706         validity_interval: *ApiValidityInterval
```
so the idea was to sneak optionality by making `validity_interval` optional, but make `ApiValidityInterval` well defined. kinda the same as for `ApiPaymentDestination` - the same idea...Once again I do not have strong opinion about this, could be either way, but maybe it would be ok to be consistent..

specifications/api/swagger.yaml
paweljakubas: I agree it is better. Adopted

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -0,0 +1,129 @@
...
+    ( DecodeAddress n
+    , DecodeStakeAddress n
+    , EncodeAddress n
+    , PaymentAddress n IcarusKey
+    ) => SpecWith Context
+spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
+    it "TRANS_NEW_CREATE_01x - Single Output Transaction" $ \ctx -> runResourceT $ do
+        let initialAmt = 3*minUTxOValue
+        wa <- fixtureWalletWith @n ctx [initialAmt]
+        wb <- fixtureWalletWith @n ctx [initialAmt]
```
piotr-iohk: This could be `emptyWallet` I guess.

specifications/api/swagger.yaml
```diff
@@ -4654,6 +4793,24 @@ paths:
         - *parametersTransactionId
       responses: *responsesDeleteTransaction
 
+  /wallets/{walletId}/transactions/construct:
+    post:
+      operationId: postTransactionConstruct
+      tags: ["Transactions"]
+      summary: Create
```
piotr-iohk:
```suggestion
      summary: Construct
```
?

piotr-iohk:
Currently in the API we have only `passphrase` being required payload parameter.
I guess we want also one of `payments` or `delegations` to be required?

Screenshot from 2021-06-28 10-42-23


EDIT: Do we need passphrase after all at the payment construction step as signing will be done in subsequent step?

lib/core/src/Cardano/Wallet/Api/Types.hs
Anviking: Aside from arguably looking a bit strange, wouldn't `Maybe (NonEmpty ApiMultiDelegationAction)` forbid `"delegations" = []`, potentially making it more difficult for folks to write API clients?

Could we not have `delegations :: [ApiMultiDelegationAction]`, but have a custom JSON parser defaulting to `[]` if the key is not present?

... if we even need to allow it to be omitted?

lib/core/src/Cardano/Wallet.hs
```diff
@@ -113,12 +113,15 @@ module Cardano.Wallet
     , selectAssetsNoOutputs
     , assignChangeAddresses
     , assignChangeAddressesAndUpdateDb
+    , assignChangeAddressesWithoutDbUpdate
     , selectionToUnsignedTx
     , signTransaction
+    , constructUnsignedTransaction
```
rvl: I think we can just call this `constructTransaction`

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1560,6 +1583,39 @@ signTransaction ctx wid mkRwdAcct pwd txCtx sel =
...
+       )
+       -- ^ Reward account derived from the root key (or somewhere else).
+    -> Passphrase "raw"
+    -> TransactionCtx
+    -> SelectionResult TxOut
+    -> ExceptT ErrConstructTx IO ByteString
+constructUnsignedTransaction ctx wid mkRwdAcct pwd txCtx sel =
+    db & \DBLayer{..} -> do
+    era <- liftIO $ currentNodeEra nl
+    withRootKey @_ @s ctx wid pwd ErrConstructTxWithRootKey $ \xprv scheme -> do
```
rvl: If we aren't signing anything then we shouldn't need the wallet root key.
We should be able to get `rewardAcct` from the database using `readRewardAccount`.

lib/core/src/Cardano/Wallet/Api/Client.hs
```diff
@@ -323,6 +331,7 @@ byronTransactionClient =
         , postExternalTransaction = _postExternalTransaction
         , deleteTransaction = _deleteTransaction
         , getTransaction = _getTransaction
+        , constructTransaction = error "TODO should be supported in the final version of Transaction Workflow."
```
rvl:
```suggestion
        , constructTransaction = error "TODO ADP-909 implement new transaction API for all wallet types."
```

lib/core/src/Cardano/Wallet/Api/Link.hs
```diff
@@ -633,6 +634,21 @@ getTransaction w t = discriminate @style
     tid = ApiTxId (t ^. typed @(ApiT (Hash "Tx")))
     mkURL mk = mk wid tid
 
+createUnsignedTransaction
```
rvl, line 637:
```suggestion
createTransaction
```

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -2342,6 +2483,52 @@ instance DecodeAddress t => FromJSON (PostTransactionData t) where
...
+    toJSON (ApiPaymentAddresses addrs) = toJSON addrs
+    toJSON (ApiPaymentAll addrs) = toJSON addrs
+
+instance DecodeAddress t => FromJSON (ApiConstructTransactionData t) where
+    parseJSON = genericParseJSON defaultRecordTypeOptions
+instance EncodeAddress t => ToJSON (ApiConstructTransactionData t) where
+    toJSON = genericToJSON defaultRecordTypeOptions
+
+instance ToJSON ApiValidityBound where
+    toJSON ApiValidityBoundUnspecified = String "unspecified"
```
rvl: Better to use `Null` I think.

lib/core/src/Cardano/Wallet/Api/Types.hs
rvl: Yeah actually it seems like the JSON instances are a bit easier when using Maybe.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1874,6 +1986,35 @@ instance FromJSON ApiDelegationAction where
...
+        object [ "quit" .= object [ "stake_key_index" .= toJSON stakeKey ] ]
+instance FromJSON ApiMultiDelegationAction where
+    parseJSON = withObject "ApiMultiDelegationAction" $ \o -> do
+        actionJoin <- o .:? "join"
+        actionQuit <- o .:? "quit"
+        case (actionJoin, actionQuit) of
+            (Just joinObj, Nothing) -> do
+                poolId <- joinObj .: "pool"
+                ix <- joinObj .: "stake_key_index"
+                pure $ Joining poolId ix
```
rvl:
```suggestion
            (Just o, Nothing) -> Joining <$> o .: "pool" <*> o .: "stake_key_index"
```

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1874,6 +1986,35 @@ instance FromJSON ApiDelegationAction where
...
+            (Nothing, Just quitObj) -> do
+                ix <- quitObj .: "stake_key_index"
+                pure $ Leaving ix
```
rvl:
```suggestion
            (Nothing, Just o) -> Leaving <$> o .: "stake_key_index"
```

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1874,6 +1986,35 @@ instance FromJSON ApiDelegationAction where
...
+            _ -> fail "ApiMultiDelegationAction needs 'join' or 'quit' field"
```
rvl:
```suggestion
            _ -> fail "ApiMultiDelegationAction needs either 'join' or 'quit', but not both"
```

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -2342,6 +2483,52 @@ instance DecodeAddress t => FromJSON (PostTransactionData t) where
...
+    toJSON (ApiValidityBoundAsSlot sl) = toJSON sl
+instance FromJSON ApiValidityBound where
+    parseJSON obj = processString obj <|> processObject obj
+      where
+        processString = withText "ApiValidityBound string" $ \str ->
+            if str == "unspecified" then
+                pure ApiValidityBoundUnspecified
+            else
+                fail "invalid string of ApiValidityBound"
+        processObject = withObject "ApiValidityBound object" $ \o -> do
```
rvl, line 2497: It should be possible try parsing the Quantities directly without testing the "unit" field.

specifications/api/swagger.yaml
rvl: For the JSON side, we want an omitted field to mean unspecified bound.

rvl (approved):
Nice work. I still have a few comments though, and have also pushed some minor changes.

piotr-iohk:
Currently in swagger only `passphrase` is marked as `required` parameter (afaiu it is required for getting reward account), however I think that we need also at least one of `payments`, `delegations` or `mint` to be there as well. It would be nice to express this in swagger somehow (not sure how as at least one of the mentioned fields is required also).

Anyway, one can currently set just `passphrase` in the request and it seems that in such case the payment is constructed, which seems rather bad:

```
$ curl -X POST http://localhost:8090/v2/wallets/23b218fd80907f03f4ce4bc138eb3131e08dc0f8/transactions-construct \
-d '{"passphrase":"Secure Passphrase"}' \
-H "Content-Type: application/json" | jq

{
  "fee": {
    "quantity": 2166293,
    "unit": "lovelace"
  },
  "serialized_transaction": "ZzZRQWdZSllJQnowL2pTajFYNmhLbzdDaXNDNEdrTWhhTEZBRC9NSDlrd1BQUDlGUVFFMEFBR0JnbGc1QUluL3h1TUJFRS81UkNGdHhIeTQxb1lZTno0aDhjVmd1azRSQUsxTTRFajFSMHN4bTlFVlpPZENoaGFycXFmU3BVN0UzUEVLWDBiOEdnQjNpSEVDR2dBaERoVURHZ0hVam5pZi8vWT0=",
  "coin_selection": {
    "withdrawals": [],
    "inputs": [
      {
        "amount": {
          "quantity": 10000006,
          "unit": "lovelace"
        },
        "address": "addr_test1qqph473hczdyrk9c9te8qh3xd6gfug604800r4rxt82a2njvupy0236txxdaz9tyuapgv94t42na9f2wcnw0zzjlgm7qw40nap",
        "id": "8b59d852cbf65fad4897eeffaa32ff09ae8c9b382dc042b436425b6e91cc77ee",
        "derivation_path": [
          "1852H",
          "1815H",
          "0H",
          "0",
          "22"
        ],
        "assets": [],
        "index": 0
      }
    ],
    "deposits": [],
    "change": [
      {
        "amount": {
          "quantity": 7833713,
          "unit": "lovelace"
        },
        "address": "addr_test1qzyll3hrqygyl72yy9kugl9c66rpsde7y8cu2c96fcgspt2vupy0236txxdaz9tyuapgv94t42na9f2wcnw0zzjlgm7qp0g974",
        "derivation_path": [
          "1852H",
          "1815H",
          "0H",
          "1",
          "0"
        ],
        "assets": []
      }
    ],
    "outputs": []
  }
}
```

It seems that one full utxo is selected and "spent" to one of the wallet's own addresses...
Checked on rev 2df0c15f79dfde1fa3e9d00e777b36c7d4becfc4.

rvl:
In the case where the user didn't provide any of payments, delegations or mint, I'm not sure whether we need to disallow it - as long as the result is a valid transaction. Perhaps a user would like to post a transaction with nothing but metadata?

piotr-iohk:
> In the case where the user didn't provide any of payments, delegations or mint, I'm not sure whether we need to disallow it - as long as the result is a valid transaction. Perhaps a user would like to post a transaction with nothing but metadata?

That would also incur fees + minUtxoValue, so in fact (at least) a payment, right? 
I guess it's possible to send metadata with mint or delegation too, but at least one of the payment/delegation/mint would be needed.

lib/core/src/Cardano/Wallet.hs

lib/core/src/Cardano/Wallet/Api/Link.hs
paweljakubas: unfortunately, there is already `createTransaction`

paweljakubas:
bors retry

paweljakubas:
bors retry

PR #2705, opened 2021-06-11
Miscellaneous release scripts enhancements

rvl:
### Overview

- Improves commit messages on the `gh-pages` branch - now easier to see how the changes correspond to `master` branch and tags.
- Adds a script to reformat the Bump.sh API changes as markdown suitable for pasting into the release notes.
- Adds the handy `jira find` command.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -81,12 +81,15 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: 2.7
-      - name: 'Install gems'
-        run: 'gem install bump-cli'
       - name: 'Install dependencies'
-        run: 'sudo snap install yq'
+        run: |
+          gem install bump-cli
+          sudo snap install yq
+          sudo apt-get -yu install html-xml-utils
       - name: 'Update Release in Bump.sh'
         run: './scripts/gh/update-bump.sh'
         env:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - rvl/misc-scripts  # temporary testing
```
Anviking: ℹ️ (don't forget)

scripts/gh/show-bump.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
Anviking, line 1: no nix?

Anviking (approved):
👌 sounds good to me

scripts/gh/show-bump.sh
rvl: Yeah often I would do:
```bash
#! /usr/bin/env nix-shell
#! nix-shell -i bash -p gnused html-xml-utils
```
but since this script is usually running on GitHub actions then we don't have `nix-shell` installed, and so we would need to execute it with `bash ./scripts/gh/show-bump.sh` ...

.github/workflows/release.yml
rvl: Since it seems to work, I just did a `git reset --hard 'HEAD^'` and force push.

rvl:
Thanks

bors r+

PR #2707, opened 2021-06-11
Add new improved ./scripts/bors-stats.rb 

Anviking:
# Issue Number

ADP-988 / ADP-970

# Overview

- [x] Add new `./bors-stats.rb` to replace `./bors-stats.sh` 

# Comments

```bash
Commands:
  bors-stats.rb cause           # List failures and merged PRs to determine causes
  bors-stats.rb help [COMMAND]  # Describe available commands or one specific command
  bors-stats.rb list            # list all failures with optional filter (e.g. list 2292)
```

```bash
Usage:
  bors-stats.rb list

Options:
  [--count=N]                                          # How many PRs to fetch
                                                       # Default: 100
  [--details=DETAILS]                                  # Show details (true/false/auto)
                                                       # Default: auto
  [--search=SEARCH]                                    # Only show failures containing this string
  [--annotate=one two three]                           # Convert matches into tags
  [--auto-annotate], [--no-auto-annotate]              # Look for hspec failures and add as tags
  [--auto-annotate-fuzzy], [--no-auto-annotate-fuzzy]  # Look for hspec failures and add as tags (fuzzily)
  [--fetch-system], [--no-fetch-system]                # Scrape and add mac or linux as tag
  [--force-refetch], [--no-force-refetch]              # Fetch new data even if recent cache exists.

list all failures with optional filter (e.g. list 2292)
```

Skärmavbild 2021-06-11 kl  17 49 10

- Not sure how easy this will run on other systems than mine yet, so I'm leaving the original bors-stats.sh with a notice.
- I compared the outputs of the two yesterday, and they matched (but by default the new script fetches more data)


rvl (commented):
This looks helpful but I had problems running it.
First it said the `ansi` package was missing, so I removed the require, and all the ANSI calls.
Then it said `thor` was missing...
Couldn't this work with just the basic packages that come with ruby 2.6?

Anviking:
Thanks for trying!

I switched to using nix:
```
#! /usr/bin/env nix-shell
#! nix-shell -i ruby -p "ruby.withPackages (ps: with ps; [ thor ])"
```

I didn't find the `ansi` package using nix, but it was easy to remove.

`thor` would take a bit more time to remove (I've never written ruby CLI tools), so I didn't. (But now it uses nix)

Anviking:
Idea for future: Maybe one could add some more general and flexible way of filtering like:
```bash
./scripts/bors-stats.rb list | jq '<some jq filter>' | ./scripts/bors-stats.rb view
```

Or:
```bash
./scripts/bors-stats.rb list --jq-filter "<some jq filter>"
```

rvl:
I had another try...
Looks like the ruby package brotli isn't packaged in nixpkgs, so we would need a Gemfile or something.
But we can just use `curl` for fetching, and it handles redirects, compression, etc.
I pushed a change to do that, but I'm not sure if it works fully.
Ruby is hard.

piotr-iohk:
It seems that logs fetched from hydra are in binary format :thinking: :
```
$ ./bors-stats.rb download-logs --dir logs2 --tag '#timeout'
https://hydra.iohk.io/build/6463832
https://hydra.iohk.io/build/6467483
https://hydra.iohk.io/build/6468421
https://hydra.iohk.io/build/6599026

$ cat logs2/6599026.log 
QT�bt���'�1�9�*U��3��n�Ӥ��S�E��4����ˤ�x�8e��0��(������;������U�����*S����pun�8~6��!�"+z3]I`
```

rvl:
Fixing now... stand by. I forgot to add the `curl --compressed` which enables decompression.

rvl:
Try now. Check `curl -V` for brotli support - most distros should have it by now. Otherwise use:
```console
$ nix-shell --run "./scripts/bors-stats.rb download-logs ..."
```

piotr-iohk:
> ```
> nix-shell --run "./scripts/bors-stats.rb download-logs ..."
> ```

Ok, well. It doesn't work for me from nix-shell:
```
$ nix-shell --run "./scripts/bors-stats.rb download-logs --dir logs_b"
warning: dumping very large path (> 256 MiB); this may run out of memory
curl: (61) Unrecognized content encoding type. libcurl understands deflate, gzip content encodings.
curl exited with status pid 113191 exit 61

```

But works fine while running as script:

```
$ ./scripts/bors-stats.rb download-logs --dir logs_b

https://hydra.iohk.io/build/5925989
https://buildkite.com/input-output-hk/cardano-wallet/builds/14064
https://hydra.iohk.io/build/5925989
https://buildkite.com/input-output-hk/cardano-wallet/builds/14064
https://buildkite.com/input-output-hk/cardano-wallet/builds/14075
...

```
(I'm wondering why it downloads `5925989` and `14064`) twice... anyway, both: logs from hydra and buildkite are in the text format now.

rvl:
@piotr-iohk Make sure you're on revision 4e51c50f102eda1bf7f22f2654455e464ce8d72a 
```
git fetch && git checkout anviking/ruby-bors-stats && git stash && git reset --hard origin/anviking/ruby-bors-stats
```

rvl:
It probably downloads the same hydra build logs twice because it's a cached failure.

rvl (approved):
I made more improvements
- Nicer log messages
- Added some source comments
- Password command
- Reordered functions in approximate order of most interesting to least interesting
- Added caching (and --force-refetch override) to most fetches
- Tweaked log filenames and cache path
- Squashed and rebased

I retested the bits that I understand how to use, so hopefully haven't broken anything again.

Issue #2428, opened 2021-01-07
(Instance of #2855) Flaky: HW_WALLETS_01x - Restoration from account public key preserves funds


KtorZ:
# Context
 
 

https://hydra.iohk.io/build/5280393

# Test Case

SHELLEY_CLI_HW_WALLETS/HW_WALLETS_01x

# Failure / Counter-example

```
  src/Test/Integration/Scenario/CLI/Shelley/HWWallets.hs:130:18: 
  3) CLI Specifications, SHELLEY_CLI_HW_WALLETS, HW_WALLETS_01x - Restoration from account public key preserves funds
       Waited longer than 90s to resolve action: "Wallet balance is as expected".
       expected: Quantity {getQuantity = 1000000}
        but got: Quantity {getQuantity = 0}
```

Anviking:
Tracking as #2855

PR #2708, opened 2021-06-11
Add more waitForTxImmutability

Anviking:
# Issue Number

ADP-970 / #2428 

# Overview

- [x] Replace `eventually` with `waitForTxImmutability` in exactly the locations where we've seen failures (#2428 )

# Comments


Anviking:
bors try

rvl:
The test that failed above is one of the tests which got the `waitForTxImmutability` treatment.

It raises the question of whether this fix is an effective way of fixing the flaky integration tests.

I believe that `waitForTxImmutability` _will_ prevent failed assertions due to rollback, and this failure is due to _another_ problem: as you can see, there are `Connection lost with the node` messages. I don't think we know the cause of these yet. Nonetheless, such errors will cause some rollback by design, and the wallet might not yet be resynced when the wait timeout passes.

The pause/resume solution should be more robust, but these changes are probably still helpful in the meantime.

lib/core-integration/src/Test/Integration/Scenario/CLI/Shelley/HWWallets.hs
```diff
@@ -122,13 +124,16 @@ spec = describe "SHELLEY_CLI_HW_WALLETS" $ do
         _ <- expectValidJSON (Proxy @(ApiTransaction n)) op
         cp `shouldBe` ExitSuccess
 
-        eventually "Wallet balance is as expected" $ do
```
rvl, line 125: I was thinking that it would be a shame to lose these high-level functional descriptions of that conditions are being tested.

So I have added a `verifyMsg` function which includes such a description.

rvl (approved):
Thanks - this should help, even if we still get failures for different reasons.

I modified the DSL `verify` function to include a descriptive message, and also to pretty-show the value being checked.

lib/test-utils/src/Test/Utils/Pretty.hs
```diff
@@ -0,0 +1,41 @@
+-- | A convenience wrapper type for pretty-showing test values.
+
+module Test.Utils.Pretty
+    ( Pretty (..)
```
Anviking, line 4: Good idea 👍 I didn't see that it existed before in SelectionSpec

lib/text-class/src/Data/Text/Class.hs
```diff
@@ -80,6 +83,9 @@ class ToText a where
     -- | Encode the specified value as text.
     toText :: a -> Text
 
+    default toText :: Buildable a => a -> Text
```
Anviking, line 86: ✔️

Anviking:
Awesome, thanks @rvl! 

bors r+

Anviking:
> I believe that waitForTxImmutability will prevent failed assertions due to rollback, and this failure is due to another problem: 

I thought so too, and was pretty confident, but with this immediate second failure I'm starting to wonder 🤔

Anviking:
> was pretty confident

I do recall a case where it was successful in some TX_EXTERNAL tests, but don't recall the details / care to find link now

But this might be something new...

Anviking:
bors try

Anviking:
bors try

Anviking:
The fact that it is failing seemingly consistently is interesting...

PR #2711, opened 2021-06-15
Reject unknown record fields in `FromJSON` instances.

jonathanknowles:
# Issue Number

⚠️ Experimental ⚠️ 

ADP-989

# Overview

This PR uses [`rejectUnknownFields`](https://hackage.haskell.org/package/aeson-1.5.6.0/docs/Data-Aeson.html#v:rejectUnknownFields) to reject unknown record fields in JSON objects sent via the API.

It makes no changes to the API specification itself.

# Remaining Work

- [ ] Adjust manually-defined `FromJSON` instances to fail if unknown fields are present with the `onlyAllowObjectFields` function.

#  Problematic Types

The following types all exhibited JSON property test failures when `rejectUnknownFields` was enabled:

- [x] `ApiActiveSharedWallet`
    (includes `ApiBlockReference` which includes `ApiSlotId`)
- [x] `ApiBlockReference`
    (includes `ApiSlotId`)
- [x] `ApiByronWallet`
    (includes `ApiBlockReference` which includes `ApiSlotId`)
- [x] `ApiNetworkInformation`
    (includes `ApiBlockReference` which includes `ApiSlotId`)
- [x] `ApiSharedWallet`
    (includes `ApiBlockReference` which includes `ApiSlotId`)
- [x] `ApiSlotReference`
    (includes `ApiSlotId`)
- [x] `ApiTransaction`
    (includes `ApiBlockReference` which includes `ApiSlotId`)
- [x] `ApiWallet`
    (includes `ApiBlockReference` which includes `ApiSlotId`)
- [x] `SomeByronWalletPostData`
    (caused by the `style` field)
- [x] `StakePoolMetadata`
    (fixed in e1f735e6f98505cee10eb2a904c9747404e3a673)

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -288,7 +288,7 @@ import Control.Arrow
 import Control.DeepSeq
     ( NFData )
 import Control.Monad
-    ( guard, when, (>=>) )
+    ( forM_, guard, when, (>=>) )
 import Data.Aeson.Types
     ( FromJSON (..)
     , SumEncoding (..)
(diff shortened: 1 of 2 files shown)


rvl:
This is a good idea I think.
We do need to be careful when enabling this however, because we may break users' applications.

PR #2712, opened 2021-06-15
 Mint burn API only, stubbed implementation

sevanspowell:
# Issue Number

ADP-346
ADP-862

# Overview

This PR modifies the cardano-wallet API to support minting and burning. Minting and burning are not actually implemented, just stubbed. This was done to minimize the amount of context the reviewer needs to keep in mind when reviewing - we are just reviewing the form of the new minting and burning endpoints.

- Add a new "mint" endpoint under the "assets" API group.
- Add a number of new API types to represent the data the user must submit with a mint/burn request, and the data they receive back.
- Adjust swagger specification to match new API types.
- Adds Arbitrary instances for API types for property tests.
- Adds a stub for the minting and burning implementation.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api.hs
+++ b/lib/core/src/Cardano/Wallet/Api.hs
@@ -40,6 +40,7 @@ module Cardano.Wallet.Api
         , ListAssets
         , GetAsset
         , GetAssetDefault
+        , MintBurnAssets
 
     , Addresses
         , ListAddresses
@@ -173,6 +174,7 @@ import Cardano.Wallet.Api.Types
     , ApiHealthCheck
     , ApiMaintenanceAction
     , ApiMaintenanceActionPostData
+    , ApiMintedBurnedTransactionT
     , ApiNetworkClock
     , ApiNetworkInformation
     , ApiNetworkParameters
@@ -206,6 +208,7 @@ import Cardano.Wallet.Api.Types
     , Iso8601Time
     , KeyFormat
     , MinWithdrawal
+    , PostMintBurnAssetDataT
     , PostTransactionFeeOldDataT
     , PostTransactionOldDataT
     , SettingsPutData
@@ -291,7 +294,7 @@ type ApiV2 n apiPool = "v2" :> Api n apiPool
 type Api n apiPool =
          Wallets
     :<|> WalletKeys
-    :<|> Assets
+    :<|> Assets n
     :<|> Addresses n
     :<|> CoinSelections n
     :<|> ShelleyTransactions n
@@ -422,8 +425,9 @@ type GetAccountKey = "wallets"
   See also: https://input-output-hk.github.io/cardano-wallet/api/#tag/Assets
 -------------------------------------------------------------------------------}
 
-type Assets =
-    ListAssets
+type Assets n =
+    MintBurnAssets n
+    :<|> ListAssets
     :<|> GetAsset
     :<|> GetAssetDefault
 
@@ -447,6 +451,13 @@ type GetAssetDefault = "wallets"
     :> "assets"
     :> Capture "policyId" (ApiT TokenPolicyId)
     :> Get '[JSON] ApiAsset
+
+-- | https://input-output-hk.github.io/cardano-wallet/api/#operation/mintBurnAssets
+type MintBurnAssets n = "wallets"
+    :> Capture "walletId" (ApiT WalletId)
+    :> "assets"
+    :> ReqBody '[JSON] (PostMintBurnAssetDataT n)
+    :> PostAccepted '[JSON] (ApiMintedBurnedTransactionT n)
 
 {-------------------------------------------------------------------------------
                                   Addresses
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -66,11 +66,15 @@ module Cardano.Wallet.Api.Types
     , ApiSelectCoinsPayments (..)
     , ApiSelectCoinsAction (..)
     , ApiCoinSelection (..)
+    , ApiMintBurnOperation (..)
+    , ApiMintData(..)
+    , ApiBurnData(..)
     , ApiCoinSelectionChange (..)
     , ApiCoinSelectionInput (..)
     , ApiCoinSelectionOutput (..)
     , ApiCoinSelectionWithdrawal (..)
     , ApiBase64
+    , ApiMintBurnData (..)
     , ApiStakePool (..)
     , ApiStakePoolMetrics (..)
     , ApiStakePoolFlag (..)
@@ -93,6 +97,8 @@ module Cardano.Wallet.Api.Types
     , ApiSerialisedTransaction (..)
     , ApiSignedTransaction (..)
     , ApiTransaction (..)
+    , ApiMintedBurnedTransaction (..)
+    , ApiMintedBurnedInfo (..)
     , ApiWithdrawalPostData (..)
     , ApiMaintenanceAction (..)
     , ApiMaintenanceActionPostData (..)
@@ -150,6 +156,7 @@ module Cardano.Wallet.Api.Types
     , ApiPaymentDestination (..)
     , ApiValidityInterval (..)
     , ApiValidityBound
+    , PostMintBurnAssetData(..)
 
     -- * API Types (Byron)
     , ApiByronWallet (..)
@@ -200,8 +207,10 @@ module Cardano.Wallet.Api.Types
     , ApiConstructTransactionDataT
     , PostTransactionOldDataT
     , PostTransactionFeeOldDataT
+    , ApiMintedBurnedTransactionT
     , ApiWalletMigrationPlanPostDataT
     , ApiWalletMigrationPostDataT
+    , PostMintBurnAssetDataT
 
     -- * API Type Conversions
     , coinToQuantity
@@ -1050,6 +1059,33 @@ data ApiTransaction (n :: NetworkDiscriminant) = ApiTransaction
     , mint :: !(ApiT W.TokenMap)
     , status :: !(ApiT TxStatus)
     , metadata :: !ApiTxMetadata
+    } deriving (Eq, Generic, Show)
+      deriving anyclass NFData
+
+-- | The response cardano-wallet returns upon successful submission of a
+-- mint/burn transaction.
+data ApiMintedBurnedTransaction (n :: NetworkDiscriminant) = ApiMintedBurnedTransaction
+    { transaction :: !(ApiTransaction n)
+    -- ^ Information about the mint/burn transaction itself.
+    , mintedBurned :: !(NonEmpty (ApiT ApiMintedBurnedInfo))
+    -- ^ Helpful information about each unique asset minted or burned (where the
+    -- identity is the policyId + asset name of the asset).
+    }
+    deriving (Eq, Generic, Show)
+    deriving anyclass NFData
+
+data ApiMintedBurnedInfo = ApiMintedBurnedInfo
+    { monetaryPolicyIndex :: !(ApiT DerivationIndex)
+    -- ^ The monetary policy index the asset was minted/burnt under.
+    , policyId            :: !(ApiT W.TokenPolicyId)
+    -- ^ The policy ID the asset was minted/burnt under.
+    , assetName           :: !(ApiT W.TokenName)
+    -- ^ The name of the asset minted/burnt.
+    , subject             :: !(ApiT W.TokenFingerprint)
+    -- ^ The subject of the asset minted/burnt. This is useful to users wishing
+    -- to attach metadata to their asset.
+    , script              :: !(ApiT (Script KeyHash))
+    -- ^ The script which this asset was minted and/or burned under
     } deriving (Eq, Generic, Show)
       deriving anyclass NFData
 
@@ -2606,6 +2642,18 @@ instance (DecodeAddress t, DecodeStakeAddress t) => FromJSON (ApiConstructTransa
 instance (EncodeAddress t, EncodeStakeAddress t) => ToJSON (ApiConstructTransaction t) where
     toJSON = genericToJSON defaultRecordTypeOptions
 
+instance
+    ( DecodeAddress n
+    , DecodeStakeAddress n
+    ) => FromJSON (ApiMintedBurnedTransaction n) where
+    parseJSON = genericParseJSON defaultRecordTypeOptions
+
+instance
+    ( EncodeAddress n
+    , EncodeStakeAddress n
+    ) => ToJSON (ApiMintedBurnedTransaction n) where
+    toJSON = genericToJSON defaultRecordTypeOptions
+
 instance FromJSON ApiWithdrawalPostData where
     parseJSON obj =
         parseSelfWithdrawal <|> fmap ExternalWithdrawal (parseJSON obj)
@@ -3305,6 +3353,8 @@ type family ApiConstructTransactionT (n :: k) :: Type
 type family ApiConstructTransactionDataT (n :: k) :: Type
 type family PostTransactionOldDataT (n :: k) :: Type
 type family PostTransactionFeeOldDataT (n :: k) :: Type
+type family ApiMintedBurnedTransactionT (n :: k) :: Type
+type family PostMintBurnAssetDataT (n :: k) :: Type
 type family ApiWalletMigrationPlanPostDataT (n :: k) :: Type
 type family ApiWalletMigrationPostDataT (n :: k1) (s :: k2) :: Type
 type family ApiPutAddressesDataT (n :: k) :: Type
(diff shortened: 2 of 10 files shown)


sevanspowell:
Believe test failures are flaky tests.

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -439,6 +443,16 @@ type GetAssetDefault = "wallets"
     :> Capture "policyId" (ApiT TokenPolicyId)
     :> Get '[JSON] ApiAsset
 
+-- | https://input-output-hk.github.io/cardano-wallet/api/#operation/mintBurnToken
+type MintBurnAsset n = "wallets"
+    :> Capture "walletId" (ApiT WalletId)
+    -- :> "assets"
+    -- :> "mint"
+    -- This has been changed to prevent overlapping routes:
+    :> "assets-mint"
```
sevanspowell: I'm not happy with this, does anyone know of a way to prevent servant from overlapping the routes?

The other routes in this namespace are of the form:

```
/assets/$policyId/...
```

So when it sees:
```
/assets/mint
```
It tries to treat "mint" as a policy ID and fails - even though this is a POST endpoint and the rest are GET.

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -439,6 +443,16 @@ type GetAssetDefault = "wallets"
     :> Capture "policyId" (ApiT TokenPolicyId)
     :> Get '[JSON] ApiAsset
 
+-- | https://input-output-hk.github.io/cardano-wallet/api/#operation/mintBurnToken
+type MintBurnAsset n = "wallets"
```
jonathanknowles: Suggestion: make this plural (to capture the general case):

```suggestion
-- | https://input-output-hk.github.io/cardano-wallet/api/#operation/mintBurnAssets
type MintBurnAssets n = "wallets"
```

lib/core/src/Cardano/Wallet/Api.hs
jonathanknowles: Hi @sevanspowell 

> It tries to treat "mint" as a policy ID and fails - even though this is a POST endpoint and the rest are GET.

Does the order in which the endpoints appear in the Servant type definition make a difference? What happens if you try the following order:
```hs
type Assets =
    ListAssets
    :<|> MintBurnAssets
    :<|> GetAsset
    :<|> GetAssetDefault
``

**Review by jonathanknowles — COMMENTED** [review]
Hi @sevanspowell 

Thanks for making this PR. 👍🏻 

I have a question: just for my own enlightenment, is there an overview of the use cases and requirements that the API in this PR is designed to satisfy? (One that I can read through and get a general idea of how the user is intended to use this API.)

For example: I noticed that the `ApiMintBurnData` can take a series of mint-burn operations. I'm just wondering what the intended use case is for this. (So I can educate myself and get a better idea of how to evaluate the API in PR.)

Thanks again!

File: lib/core/src/Cardano/Wallet/Api.hs (hunk 1 above)
Comment (sevanspowell) [reply lib/core/src/Cardano/Wallet/Api.hs]: Hey @jonathanknowles, the order in which the endpoints appear does not seem to matter in my experiments. I'll try and generate a failing test to better demonstrate what's going on.

**Comment by sevanspowell** [comment]
I think we originally saw two primary use cases for this feature:

1. To add minting and burning to our cardano-wallet test suite
2. To allow users to mint and burn using the cardano-wallet abstraction

The first use case is largely undefined, so for now all we will say is that it is beneficial to have minting and burning present in our test suites.

Regarding the second use case, minting and burning is currently achieved using the cardano-cli. A detailed example of minting an asset using cardano-cli can be found [here](https://input-output-hk.github.io/offchain-metadata-tools/).

Simplified, the important points are:
  - Assets are minted and burned under a "policy"
    - For example, the simplest policy is: "only transactions signed by this
      key" may mint or burn tokens.
  - The typical process for minting/burning is:
    1. Generate keys for the minting policy.
    2. Create a policy script that allows only those keys to mint/burn assets.
    3. Build a transaction that mints an asset under that policy script, with an
       asset name.
    4. Sign the transaction with the policy script and the minting keys.
    5. Submit the transaction.

A user of cardano-wallet typically wants to make use of [HD wallets](https://input-output-hk.github.io/adrestia/docs/key-concepts/hierarchical-deterministic-wallets/) and so doesn't want to manually create keys for minting and burning, we have formed this API to meet that expectation.

The user doesn't refer to an asset via it's policy script, instead we ask the user for an index into a key derivation path (the monetary policy index), from this index we derive the keys for the policy, create a policy script from those keys, then build and submit the minting or burning transaction.

We do not try and support more complex policy scripts at this time (for example, the user could create a policy script that's can only mint after a given slot), which simplifies the requirements of this API.

I imagine a typical use-case of this API would be to mint a token to sell as an NFT:

```
export ASSET_NAME=$(echo my-nft | basenc --base16)
export RECV_ADDR="one of my addresses"

POST wallets/{walletID}/assets/mint 
  { "mint_burn": {
      "asset_name": "$ASSET_NAME",
      "monetary_policy_index": "0",
      "operations": [
        {
          "mint": {
            "amount": {
              "quantity": 1,
              "unit": "assets"
            }
            , "receiving_address": "$RECV_ADDR"
          }
        }
      ],
      "passphrase": "..."
    }
 }
```

One of the first things a user would want to do with their NFT is add metadata for it using offchain-metadata-tools:

```
export SUBJECT=$(resp from prev post | jq .subject)
export POLICY=$(resp from prev post | jq .policy)
token-metadata-creator entry $SUBJECT \
  --name "Quid" \
  --description "The un-offical currency of Great Britain." \
jq '.policy = "$POLICY"' $SUBJECT.json.draft
```

Next the user will need to attest to the metadata properties:

```
token-metadata-creator entry $SUBJECT -a policy/policy.skey
```

Unfortunately, this use case will fail here, we don't have access to the policy keys! They were abstracted away by cardano-wallet. So at this stage the usefulness of this API to the user is questionable.

Perhaps this could be handled in some way by modifying offchain-metadata-tools and/or cardano-wallet.

File: lib/core/src/Cardano/Wallet/Api.hs (hunk 1 above)
Comment (sevanspowell) [reply lib/core/src/Cardano/Wallet/Api.hs]: I've got a bit stuck trying to demonstrate this with a test, but I've been able to reproduce it by CURLing the running wallet.

The order of the Servant type definition makes no difference to the outcome.

Comment (sevanspowell) [reply lib/core/src/Cardano/Wallet/Api.hs]: What I've done (suggested by @rvl), is to just make the mint/burn endpoint:

```
POST /wallets/.../assets
```

Comment (piotr-iohk) [reply lib/core/src/Cardano/Wallet/Api.hs]: The rdoc was complaining about duplicate keys in swagger.yaml. I've fixed it in 6d17d33.

**Comment by piotr-iohk** [comment]
>I imagine a typical use-case of this API would be to mint a token to sell as an NFT:
...
>Unfortunately, this use case will fail here, we don't have access to the policy keys! They were abstracted away by cardano-wallet. So at this stage the usefulness of this API to the user is questionable.

I believe that despite not being able to put token metadata in the metadata server, it is still possible to use this feature as to mint NFTs. AFAIK NFT creators are adding some metadata on-chain to a minting transaction. There is a standard that's being discussed -> https://github.com/Berry-Pool/CIPs/blob/master/CIP-NFTMetadataStandard.md. In any case I think it would be good to state that limitation into release notes and swagger. I've added a note into swagger in f2bef0c.

File: specifications/api/swagger.yaml [hunk 3]
```diff
@@ -2748,6 +2748,86 @@ components:
...
+        - amount
+      properties:
+        receiving_address: *addressId
+        amount: *ApiAssetQuantity
+
+    ApiBurnData: &ApiBurnData
+      allOf:
+        - *ApiAssetQuantity
+
+    ApiMintBurnOperation: &ApiMintBurnOperation
```
piotr-iohk, line 2971: If I'm not mistaken operation is not quite correctly formatted in API doc:
Screenshot from 2021-06-17 21-24-19

It is:
```
   "operations": [
      {
        "mint": {
          "receiving_address": "addr1sjck9mdmfyhzvjhydcjllgj9vjvl522w0573ncustrrr2rg7h9azg4cyqd36yyd48t5ut72hgld0fg2xfvz82xgwh7wal6g2xt8n996s3xvu5g",
          "amount": {
            "quantity": 14,
            "unit": "assets"
          }
        },
        "burn": {
          "quantity": 14,
          "unit": "assets"
        }
      }
    ]
```
but shouldn't it be:
```
 [
      {
        "mint": {
          "receiving_address": "addr_test1qq223dyrt96yx79rz3yda2vs6xjx28ts2ax39hp8cxq2p0shpgwhzykzha0z6w27gxsps40gurvxjt2vxh26009q5wqqxpj2ey",
          "amount": {
            "quantity": 14,
            "unit": "assets"
          }
        } 
       },
       {
        "burn": {
          "quantity": 14,
          "unit": "assets"
        }
      }
]
```
i.e. "mint" and "burn" in separate `{}` ?

sevanspowell: Good eye @piotr-iohk, I've fixed this in a6b9895.

Thanks for adding the warning about metadata too!

sevanspowell: See new redoc output:

mint
burn

piotr-iohk: Nice, thanks!

Question: Are multiple `mint` or `burn` going to be allowed in single transaction?

`"operations": []` is an Array object, which implicates that it will be the case, just want to make sure.
On the other hand there is `Array one of` in the redoc, so it might be slightly confusing, what do you think? Maybe if there is only one mint or burn allowed they should be inside `"operations": {}` not `"operations": []`?

sevanspowell: We planned for multiple mint/burn operations in a single transaction. It's something supported by cardano-node and it wasn't hard to allow.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1026,6 +1034,24 @@ data ApiTransaction (n :: NetworkDiscriminant) = ApiTransaction
...
+    , monetaryPolicyIndex :: !(ApiT DerivationIndex)
+    -- ^ The monetary policy index the asset was minted/burnt under.
+    , policyId            :: !(ApiT W.TokenPolicyId)
+    -- ^ The policy ID the asset was minted/burnt under.
+    , assetName           :: !(ApiT W.TokenName)
+    -- ^ The name of the asset minted/burnt.
+    , subject             :: !(ApiT W.TokenFingerprint)
+    -- ^ The subject of the asset minted/burnt. This is useful to users wishing
+    -- to attach metadata to their asset.
+    }
```
rvl: How about including the script in the response? Is that feasible?

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -3334,3 +3380,114 @@ instance FromJSON (ApiT SmashServer) where
...
+-- the user, we ask the user not for a specific signature, but rather for a key
+-- derivation index, which we use to derive the signature to construct the
+-- policy with.
+data ApiMintBurnData (n :: NetworkDiscriminant) = ApiMintBurnData
+    { monetaryPolicyIndex :: !(Maybe (ApiT DerivationIndex))
+    -- ^ The key derivation index to use to construct the policy.
+    , assetName           :: !(ApiT W.TokenName)
+    -- ^ The name of the asset to mint/burn.
+    , operations          :: !(NonEmpty (ApiMintBurnOperation n))
+    -- ^ The set of minting and burning operations to perform.
```
rvl: This doesn't seem quite right to me. There could be contradictory elements in the operations list.

What about this?
```suggestion
    , assetNames :: !(Map (ApiT W.TokenName) (ApiMintBurnOperation n))
    -- ^ A mint/burn amount for assets of the policy script
```

specifications/api/swagger.yaml
```diff
@@ -4660,6 +4766,33 @@ paths:
...
+      tags: ["Assets"]
+      summary: Mint/Burn
+      description: |
+        <p align="right">status: <strong>stable</strong></p>
+
+        Mint and burn assets from the wallet.
+
+        **⚠️ WARNING ⚠️**
+
+        Please note that due to the fact that there is no physical access to policy keys
```
rvl: Kinda ... but the user can use cardano-address to derive the policy key themselves according to the forthcoming CIP.

We could also add an endpoint which returns policy keys ... in future.

specifications/api/swagger.yaml
```diff
@@ -4660,6 +4766,33 @@ paths:
...
       responses: *responsesListAssets
 
+    post:
+      operationId: mintBurnAssets
+      tags: ["Assets"]
+      summary: Mint/Burn
+      description: |
+        <p align="right">status: <strong>stable</strong></p>
+
+        Mint and burn assets from the wallet.
```
rvl, line 4797: It would be worth explaining a little bit about how the script is made -- i.e. how policy keys are chosen.

specifications/api/swagger.yaml
```diff
@@ -4660,6 +4766,33 @@ paths:
         - *parametersWalletId
       responses: *responsesListAssets
 
+    post:
+      operationId: mintBurnAssets
+      tags: ["Assets"]
+      summary: Mint/Burn
+      description: |
+        <p align="right">status: <strong>stable</strong></p>
```
```suggestion
        <p align="right">status: <strong>under development</strong></p>
```

lib/core/src/Cardano/Wallet/Api.hs
```diff
@@ -448,6 +452,13 @@ type GetAssetDefault = "wallets"
     :> Capture "policyId" (ApiT TokenPolicyId)
     :> Get '[JSON] ApiAsset
 
+-- | https://input-output-hk.github.io/cardano-wallet/api/#operation/mintBurnAssets
+type MintBurnAssets n = "wallets"
+    :> Capture "walletId" (ApiT WalletId)
+    :> "assets"
```
rvl, line 458: Did you resolve the overlapping API path issue?
If not, don't fret too much about ugly URLs - because we will soon be rolling `PostMintBurnAssetDataT` into the transaction creation endpoint.

rvl (changes requested):
Sweet

lib/core/src/Cardano/Wallet/Api/Types.hs
sevanspowell: It's not quite what it seems. The operations contain no reference to what they are operating on (it's either `mint 3` or `burn 4`, there is no reference to the asset).

The caller can choose one monetary policy index, one asset name, then any number of mint/burn operations for that asset.

I should probably allow the user to operate on multiple assets:

```
 data PostMintBurnAssetData (n :: NetworkDiscriminant) = PostMintBurnAssetData
-    { mintBurn   :: !(ApiMintBurnData n)
+    { mintBurn   :: !(NonEmpty (ApiMintBurnData n)) 
     -- ^ Minting and burning request.
```

And then offer a singular operation for each asset, sorta like it's done in Tx construction:

```
 data ApiMintBurnData (n :: NetworkDiscriminant) = ApiMintBurnData
     { monetaryPolicyIndex :: !(Maybe (ApiT DerivationIndex))
     -- ^ The key derivation index to use to construct the policy.
     , assetName           :: !(ApiT W.TokenName)
     -- ^ The name of the asset to mint/burn.
-    , operations          :: !(NonEmpty (ApiMintBurnOperation n))
+    , operation           :: !(ApiMintBurnOperation n)
```

sevanspowell: I've made this change in e7adcde7f.

specifications/api/swagger.yaml
sevanspowell: Addressed in 5780469e3.

specifications/api/swagger.yaml
sevanspowell: Slightly improved this with b245fe4f9.

lib/core/src/Cardano/Wallet/Api.hs
sevanspowell: Yep, resolved overlapping API path issue.

lib/core/src/Cardano/Wallet/Api/Types.hs
sevanspowell: I think so. We could return the JSON representation of the script, or the CBOR, but I think JSON is more appropriate?

lib/core/src/Cardano/Wallet/Api/Types.hs
sevanspowell: Oh... But if the user is allowed to mint multiple assets in one API call I'll have to rework the response... a sec.

lib/core/src/Cardano/Wallet/Api/Types.hs
sevanspowell: Addressed in 619ba4806.

lib/core/src/Cardano/Wallet/Api/Types.hs
sevanspowell: Addressed in 619ba48.

rvl (approved):
Good stuff, thanks.

rvl:
I added some minor rework, squashed, and rebased.

bors r+

PR #2716, opened 2021-06-15
Tweak cluster params to reduce likelyhood of rollbacks

Anviking:
# Issue Number

ADP-970

# Overview

- [x] Tweak cluster slotting parameters to reduce likelyhood of rollbacks 

# Comments

- [ ] Haven't run this locally yet, should compare with master

```
I have notes from an experiment on 11 December 2020:

- f=0.5, sl=0.2, el=50 => 220s
    - 382 `ROLLBACK` printed
- f=0.833333, sl=0.33333, el=30 => 234s
    - Failed once the first time I ran it, on ADDRESS_IMPORT_05 - I can import 15000 of addresses
    - 1031 `ROLLBACK` printed
- f=0.3125, sl=0.125, el=80 => 209.7s
    - 143 `ROLLBACK` printed

The more slots are empty (using lowering f), the less collissions (slot
battles) we have, keeping the time of an epoch in seconds fixed.

So f=0.3125 should hopefully alleviate some rollback-related flakiness
in CI.

The downside is that we are less likely to catch real problems, but
there's always a tradeoff, and flaky tests are currently hurting us a
lot.

```

### DIFF
--- a/lib/core-integration/src/Test/Integration/Framework/DSL.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/DSL.hs
@@ -56,6 +56,7 @@ module Test.Integration.Framework.DSL
     , minUTxOValue
     , slotLengthValue
     , securityParameterValue
+    , activeSlotCoeff
     , epochLengthValue
     , defaultTxTTL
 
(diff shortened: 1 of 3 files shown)


Anviking:
bors try

Anviking:
bors try

Anviking:
Ok, fine, let's not, as they hide the problems. But we could consider this as an option.

PR #2709, opened 2021-06-11
Mark LoggingSpec pending on macOS to alleviate timeouts

Anviking:
# Issue Number

ADP-970

# Overview

- [x] Mark LoggingSpec pending on macOS

PR #2717, opened 2021-06-16
Reduce the number of concurrent request threads in middleware tests.

jonathanknowles:
# Issue Number

ADP-978

# Overview

This PR reduces the maximum number of concurrent requests created in middleware tests.

Experimental testing shows that these tests fail if virtual memory is constricted to 4 GiB, but pass if the number of concurrent requests is restricted to a smaller number.


jonathanknowles:
Closing for the moment, as we have already disabled these tests on mac platforms in CI. (See #2709)

PR #2718, opened 2021-06-16
scripts/bors-stats.rb: fix --force-refresh

Anviking:
# Issue Number

ADP-970

# Overview

- [x] Fix ` ./scripts/bors-stats.rb list --fetch-system`

# Comments


Anviking:
Actually, I don't think `--force-refetch` should affect `--fetch-system`.

`--force-refetch` for GitHub data is often useful to fetch recent changes. Re-fetching the os is not useful unless your cache is somehow corrupted.

rvl (approved):
Fair enough :+1:

Issue #2472, opened 2021-01-28
Unit tests sometimes time out. 


KtorZ:
# Context
 
 

- https://hydra.iohk.io/build/5448328
- https://hydra.iohk.io/build/5470171
- https://hydra.iohk.io/build/5757989/nixlog/1/tail

# Test Case

?

# Failure / Counter-example

```hs
Build failed:

    ci/hydra-build:required

        Valid ISO 8601 extended UTC times
          2008-09-15T15:53:00Z
            +++ OK, passed 1 test.
          2008-09-15T15:53:00.1Z
            +++ OK, passed 1 test.
          2008-09-15T15:53:00.12Z
            +++ OK, passed 1 test.
        Non-time examples
          Data.Time.TextSpec[56:21]
            +++ OK, passed 1 test.
          w
            +++ OK, passed 1 test.
          wibble
            +++ OK, passed 1 test.
        ISO 8601 basic format without times
          2008
            +++ OK, passed 1 test.
          200809
            +++ OK, passed 1 test.
          20080915
            +++ OK, passed 1 test.
        ISO 8601 basic format without timezones
          20080915T155300
            +++ OK, passed 1 test.
          20080915T155300.1
            +++ OK, passed 1 test.
          20080915T155300.12
            +++ OK, passed 1 test.
building of '/nix/store/zy6scpy44agjf03q800lfc9mrk9zlyq1-cardano-wallet-core-test-unit-2021.3.4-check' timed out after 900 seconds of silence
```

rvl:
This time out is likely to be the same one as #2393 and #2436.
They all stop at pretty much the same place.

Anviking:
Could _perhaps_ be related:

Here, locally, it seems it blows up when trying to display the counterexample:

<img width="1536" alt="Skärmavbild 2021-02-16 kl  14 48 44" src="https://user-images.githubusercontent.com/304423/108071542-22ef1100-7066-11eb-89c1-fdf839839d23.png">

The error was specific to #2521, but maybe the size of the counterexample could be related to this issue.

rvl:
Yeah nice hypothesis @Anviking - plausible because Arbitrary checkpoints will probably include Arbitrary TokenBundles now.

Anviking:
[...]

PR #2723, opened 2021-06-17
Use LineBuffering in unit tests

Anviking:
# Issue Number

ADP-970 / #2472 

# Overview

- [x] Try using line buffering in unit tests, like we do for integration tests

# Comments


rvl (approved):
Oh yes, I was going to mention this but forgot... With this change we will see exactly where it gets stuck.

rvl (approved):
Good thanks - I added a `Test.Utils.Startup.withLineBuffering` function for this, and slightly reorganised the `Test.Utils.` modules.

PR #2725, opened 2021-06-18
Modify coin selection algorithm to support minting and burning

sevanspowell:
# Issue Number

ADP-997
ADP-346

# Pre-merge checks

- [x] Perform soak test of test suite to ensure no flakiness has been introduced.
    (✅ 200 test runs with `--match "RoundRobin"` passed with no failures.)

# Overview

This PR modifies the coin selection algorithm to account for minting and burning.

- Modifies "performSelection" to account for minted and burnt tokens. On a superficial level, mints provide extra tokens to use, where burns increase the amount of tokens we need to select.
- Modifies the "makeChange" algorithm to account for minted and burnt tokens. When tokens are minted they are added to the largest change bundle. When tokens are burnt they are removed from the smallest change bundles until the burn has been satisfied. See the code for more detailed commentary.
- Adds property tests that ensure the "addMintChange" and "removeBurnChange" family of functions:
  - maintain number of change bundles,
  - maintain partial ascending ordering of change bundles
  - modify the total value of the change bundle by their respective mint/burn amounts.
- Adds a few unit tests to ensure the change generation occurs as we expect it to.
- Adds some debugging information to the coin selection tests to make debugging in the presence of minting and burning easier.


lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -1313,6 +1313,7 @@ data ApiErrorCode
     | SharedWalletNoSuchCosigner
     | SharedWalletCannotUpdateKey
     | SharedWalletScriptTemplateInvalid
+    | MintedNotSpent
```
sevanspowell: This does not necessarily have to be part of the public API.

The API structure does not allow a user to specify a mint that is not sent to an explicit address, so under the current conditions, this error should never be thrown. Hence it probably doesn't need to be in the public API.

Saying that, a token which is minted but not spent or burnt is something that could happen with some other call to performSelection, so I've included this error as part of the public API.

sevanspowell: But I'm happy to go either way. We could also consider sending minted values that don't have an explicit receiving address to a change bundle. The "makeChange" function already handles that case, we'd just have to update "performSelection".

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -588,7 +650,7 @@ prop_performSelection_small minCoinValueFor costFor (Blind (Small criteria)) =
     checkCoverage $
     cover 30 (balanceSufficient criteria)
         "balance sufficient" $
-    cover 30 (not $ balanceSufficient criteria)
+    cover 25 (not $ balanceSufficient criteria)
```
sevanspowell, line 703: These coverage values were reduced because it seems my generator changes have reduced the occurrence of these paths.

I"m not sure how to fix this.

piotr-iohk:
:thinking: Interesting, sending POST on this branch to the ep gives me `405`:
```
$ curl -X POST http://localhost:8090/v2/wallets/1b0aa24994b4181e79116c131510f2abf6cdaa4f/assets \
-d '{"mint_burn":{"monetary_policy_index":"11","asset_name":"484148414841","operations":[{"mint":{"receiving_address":"addr_test1qq223dyrt96yx79rz3yda2vs6xjx28ts2ax39hp8cxq2p0shpgwhzykzha0z6w27gxsps40gurvxjt2vxh26009q5wqqxpj2ey","amount":{"quantity":14,"unit":"assets"}}},{"burn":{"quantity":14,"unit":"assets"}}]},"passphrase":"Secure Passphrase","metadata":{"0":{"string":"cardano"},"1":{"int":14},"2":{"bytes":"2512a00e9653fe49a44a5886202e24d77eeb998f"},"3":{"list":[{"int":14},{"int":42},{"string":"1337"}]},"4":{"map":[{"k":{"string":"key"},"v":{"string":"value"}},{"k":{"int":14},"v":{"int":42}}]}},"time_to_live":{"quantity":10,"unit":"second"}}' -H "Content-Type: application/json" | jq


{
  "code": "method_not_allowed",
  "message": "You've reached a known endpoint but I don't know how to handle the HTTP method specified. Please double-check both the endpoint and the method: one of them is likely to be incorrect (for example: POST instead of PUT, or GET instead of POST...)."
}

```

sevanspowell:
> Interesting, sending POST on this branch to the ep gives me `405`:
> 
> ```
> $ curl -X POST http://localhost:8090/v2/wallets/1b0aa24994b4181e79116c131510f2abf6cdaa4f/assets \
> -d '{"mint_burn":{"monetary_policy_index":"11","asset_name":"484148414841","operations":[{"mint":{"receiving_address":"addr_test1qq223dyrt96yx79rz3yda2vs6xjx28ts2ax39hp8cxq2p0shpgwhzykzha0z6w27gxsps40gurvxjt2vxh26009q5wqqxpj2ey","amount":{"quantity":14,"unit":"assets"}}},{"burn":{"quantity":14,"unit":"assets"}}]},"passphrase":"Secure Passphrase","metadata":{"0":{"string":"cardano"},"1":{"int":14},"2":{"bytes":"2512a00e9653fe49a44a5886202e24d77eeb998f"},"3":{"list":[{"int":14},{"int":42},{"string":"1337"}]},"4":{"map":[{"k":{"string":"key"},"v":{"string":"value"}},{"k":{"int":14},"v":{"int":42}}]}},"time_to_live":{"quantity":10,"unit":"second"}}' -H "Content-Type: application/json" | jq
> 
> 
> {
>   "code": "method_not_allowed",
>   "message": "You've reached a known endpoint but I don't know how to handle the HTTP method specified. Please double-check both the endpoint and the method: one of them is likely to be incorrect (for example: POST instead of PUT, or GET instead of POST...)."
> }
> ```

Ah, interesting @piotr-iohk. My stub definition was malformed. Should be fixed now. Of course by "fixed" I mean the stub will still return:

```
< HTTP/1.1 500 Internal Server Error
< Transfer-Encoding: chunked
< Date: Fri, 18 Jun 2021 09:04:34 GMT
< Server: Warp/3.3.14
< Content-Type: text/plain; charset=utf-8
< 
Something went wrong
```

Because the API endpoint is still stubbed, I've only changed coin selection.

It should correctly verify the form of the request data though!

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -274,10 +287,44 @@ data SelectionError
...
+        :: !TokenMap
+      -- ^ The values to burn
+    , requestedOutputs
+        :: !TokenBundle
+      -- ^ The outputs the requester asked to be covered
+    } deriving (Generic, Eq, Show)
+
+-- | Calculate the minted values not present in the burns or outputs
+-- from a @OutputsInsufficientError@.
+outputsMissing :: OutputsInsufficientError -> TokenBundle
```
jonathanknowles: Suggestions:

1. Change the return type of this function to `TokenMap`, as the `Coin` (ada) portion can only ever be zero.
2. Indicate in the function name that it refers to output **_assets_** that are missing, rather than just outputs. (We already use the "outputs" terminology to mean **_actual_** outputs (with addresses), rather than their aggregated value.)

```suggestion
missingOutputAssets :: OutputsInsufficientError -> TokenMap
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -370,18 +417,32 @@ prepareOutputsWith minCoinValueFor = fmap $ \out ->
 -- Returns 'BalanceInsufficient' if the total balance of 'utxoAvailable' is not
 -- strictly greater than or equal to the total balance of 'outputsToCover'.
 --
+-- Returns 'OutputsInsufficientError' if the 'minted' values are not a
+-- subset of the 'outputsToCover' plus the 'burned' values. That is,
+-- the minted values are not spent or burned.
```
jonathanknowles: Nice comment. 👍🏻

lib/core/src/Cardano/Wallet/Api/Types.hs
jonathanknowles: Suggestion: make this even more descriptive:

```suggestion
    | TokensMintedButNotSpentOrBurned
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -274,10 +287,44 @@ data SelectionError
...
+        OutputsInsufficientError
     | UnableToConstructChange
         UnableToConstructChangeError
     deriving (Generic, Eq, Show)
 
+-- | Indicates that the minted value is not spent or burned. That is, the minted
+-- value is not a subset (`leq`) of the value of the requested outputs plus the
+-- burned value.
+--
+-- We introuduced this error because it does not make sense for a user to mint a
```
jonathanknowles: Small typo:

```suggestion
-- We introduced this error because it does not make sense for a user to mint a
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -274,10 +287,44 @@ data SelectionError
...
         (NonEmpty InsufficientMinCoinValueError)
+    | OutputsInsufficient
+        OutputsInsufficientError
     | UnableToConstructChange
         UnableToConstructChangeError
     deriving (Generic, Eq, Show)
 
+-- | Indicates that the minted value is not spent or burned. That is, the minted
+-- value is not a subset (`leq`) of the value of the requested outputs plus the
+-- burned value.
```
jonathanknowles: Nice comment again! 👍🏻

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -274,10 +287,44 @@ data SelectionError
...
+data OutputsInsufficientError = OutputsInsufficientError
+    { assetsToMint
+        :: !TokenMap
+      -- ^ The values to mint
+    , assetsToBurn
+        :: !TokenMap
+      -- ^ The values to burn
+    , requestedOutputs
+        :: !TokenBundle
+      -- ^ The outputs the requester asked to be covered
```
jonathanknowles: Suggestions:

1. Use `TokenMap` here, as the `Coin` (ada) value is not relevant to the error.
2. Use a field name that clarifies it's the **_combined value_** of all non-ada assets in the requested outputs, rather than the outputs themselves. (We already use the "outputs" terminology to mean **_actual_** outputs (with addresses), rather than their aggregated value.)

```suggestion
    , requestedOutputAssets
        :: !TokenMap
      -- ^ The output assets the requester asked to be covered
```

lib/core/src/Cardano/Wallet/Api/Types.hs
jonathanknowles: In the absence of more concrete requirements to do otherwise, I think the solution you have gone for is great for now.

Ideally, product management would be able to give an opinion on the following questions:

1. What should happen when someone mints assets but doesn't spend them all in the user-specified outputs?
2. Which of the following options is preferable:
    a. require users to spend or burn everything they mint (returning an error if they don't); OR
    b. require the wallet to pay minted-but-not-spent-or-burned assets back as change.
    
Perhaps we can discuss this with PM in the adrestia slack channel?

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -274,10 +287,44 @@ data SelectionError
...
+-- from a @OutputsInsufficientError@.
+outputsMissing :: OutputsInsufficientError -> TokenBundle
+outputsMissing (OutputsInsufficientError mint burn out) =
+    -- We use difference, which will show us the quantities in "mint"
+    -- that are not in "out + burn". Any amount present in "out + burn",
+    -- but not present in "mint" will simply be zeroed out, which is
+    -- the behaviour we want for this error report.
+    TokenBundle.fromTokenMap mint
+        `TokenBundle.difference`
+            (out `TokenBundle.add` TokenBundle.fromTokenMap burn)
```
jonathanknowles: Suggestion: express this calculation in terms of `TokenMap` as the `Coin` (ada) part is not relevant.
```suggestion
    mint `TokenMap.difference` (burn `TokenMap.add` out)
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -400,10 +461,19 @@ performSelection
         -- ^ The selection goal to satisfy.
     -> m (Either SelectionError (SelectionResult TokenBundle))
 performSelection minCoinFor costFor bundleSizeAssessor criteria
+    -- Is the minted value all spent or burnt?
+    | not (TokenBundle.fromTokenMap assetsToMint
+        `leq` (requestedOutputs
+             `TokenBundle.add` TokenBundle.fromTokenMap assetsToBurn)) =
+        pure $ Left $ OutputsInsufficient $ OutputsInsufficientError
+            { assetsToMint, assetsToBurn, requestedOutputs }
```
jonathanknowles: Suggestion: rewrite this guard in terms of `TokenMap`, as the `Coin` (ada) value is not relevant.

Since we're using `leq` here, we can also use the `<>` function to add token maps together, which is more concise:

```suggestion
    | not (assetsToMint `leq` (assetsToBurn <> requestedOutputAssets)) =
        pure $ Left $ OutputsInsufficient $ OutputsInsufficientError
            { assetsToMint, assetsToBurn, requestedOutputAssets }
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -400,10 +461,19 @@ performSelection
...
+
+    -- Do we have enough available balance?
     | not (balanceRequired `leq` balanceAvailable) =
         pure $ Left $ BalanceInsufficient $ BalanceInsufficientError
             { balanceAvailable, balanceRequired }
 
+    -- Are the values of the outputs too small?
```
jonathanknowles: Suggestion: Make this comment specific to **minimum ada quantities**:

(Thanks for adding these comments by the way!)

```suggestion
    -- Are the minimum ada quantities of the outputs too small?
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -426,8 +495,12 @@ performSelection minCoinFor costFor bundleSizeAssessor criteria
         , utxoAvailable
         , selectionLimit
         , extraCoinSource
+        , assetsToMint
+        , assetsToBurn
         } = criteria
 
+    requestedOutputs = F.foldMap (view #tokens) outputsToCover
```
jonathanknowles, line 519: Suggestion: split this up into two definitions, and avoid using the term "outputs" here, as this value is actually an aggregate balance value.

The second definition can be used in contexts where we only need the non-ada assets:

```suggestion
    requestedOutputBalance :: TokenBundle
    requestedOutputBalance = F.foldMap (view #tokens) outputsToCover

    requestedOutputAssets :: TokenMap
    requestedOutputAssets = view #tokens requestedOutputBalance
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -517,6 +600,9 @@ performSelection minCoinFor costFor bundleSizeAssessor criteria
     -- Eventually it returns just a final selection, or 'Nothing' if no more
     -- ada-only inputs are available.
     --
+    -- This function also takes a list of tokens that are to be
+    -- burned, and hence although an input will be consumed for them,
+    -- this function won't make an associated output for them.
```
jonathanknowles: Some suggested rewordings:

```suggestion
    -- This function also takes a set of tokens that are to be burned,
    -- and hence although one or more inputs will be consumed for
    -- them, this function won't make associated outputs for them.
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -942,6 +1030,10 @@ data MakeChangeCriteria minCoinFor bundleSizeAssessor = MakeChangeCriteria
         -- ^ Token bundles of selected inputs.
     , outputBundles :: NonEmpty TokenBundle
         -- ^ Token bundles of original outputs.
+    , assetsToMint :: TokenMap
+        -- ^ Minted tokens provide an additional source of tokens.
+    , assetsToBurn :: TokenMap
+        -- ^ Burned tokens consume an input but produce no output.
```
jonathanknowles: This second comment feels a bit wrong somehow, since we might consume **_more than one input_** in order to cover the amount to be burned.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -1375,6 +1492,205 @@ makeChangeForCoin
     -> NonEmpty Coin
 makeChangeForCoin = flip Coin.unsafePartition
 
+-- Once we know how much change to give, grouping the change into bundles is a
+-- somewhat complicated topic. We want to bundle the change such that it
+-- provides, as much as possible, useful sets of funds that are likely to be
+-- consumed by the user in future, where larger values are considered more
+-- "useful". A key property is that the number of change bundles should reflect
```
jonathanknowles: Suggestion: make the definition of "useful" a bit more precise:

```suggestion
-- consumed by the user in future, where values that more closely approximate
-- the user-specified outputs are considered more "useful". A key property is
-- that the number of change bundles should reflect ...
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -1375,6 +1492,205 @@ makeChangeForCoin
...
+-- the number of outputs the user requested. If the user sends value to five
+-- different outputs, we should output five change bundles.
+--
+-- However, we also want to mint and burn tokens. We consider minting and
+-- burning values "non-user-specified outputs" because they aren't
+-- explicitly included in the outputs the user specified ("outputs to
+-- cover"). Minted tokens add to the amount of change we need to give and
```
jonathanknowles: > We consider minting and burning values "non-user-specified outputs" because they aren't explicitly included in the outputs the user specified ("outputs to cover")

I suspect this statement isn't quite right.

For example, it's quite possible for a user to mint `x` tokens of asset `A`, and **_also_** include `y` tokens of asset `A` in the outputs they specify.

So in this sense, asset `A` is indeed a user-specified output asset, and it should probably appear in the _change for user-specified outputs_.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -1375,6 +1492,205 @@ makeChangeForCoin
...
+  :: TokenMap
+  -> NonEmpty TokenMap
+  -> NonEmpty TokenMap
+removeBurnValuesFromChangeMaps burns =
+    F.foldr
+        (\burn -> (removeBurnValueFromChangeMaps burn .))
+        id
+        (TokenMap.toFlatList burns)
+
+
```
jonathanknowles: https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#use-only-a-single-blank-line-between-top-level-definitions

```suggestion

```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -1375,6 +1492,205 @@ makeChangeForCoin
...
+                          -- Burnt all the gas, don't burn any more
+                          _qty ->
+                              xs
+                  in
+                      burnt `NE.cons` burnMore
+
+removeBurnValuesFromChangeMaps
+  :: TokenMap
+  -> NonEmpty TokenMap
+  -> NonEmpty TokenMap
```
jonathanknowles: Perhaps add a comment to this function, and comments for each of the arguments?

(Though the name pretty obviously indicates that it's a pluralization of `removeBurnValueFromChangeMaps`.)
```suggestion
    :: TokenMap
    -> NonEmpty TokenMap
    -> NonEmpty TokenMap
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -1375,6 +1492,205 @@ makeChangeForCoin
...
+-- function is the same as removing the burnt value from the change maps:
+--   F.fold changeMaps `TokenMap.difference` TokenMap.singleton assetId qty
+--          == F.fold (f (assetId, qty) changeMaps)
+-- 2. length is conserved
+-- 3. order is preserved (i.e., the result is in ascending partial order).
+removeBurnValueFromChangeMaps
+    :: (AssetId, TokenQuantity)
+    -> NonEmpty TokenMap
+    -> NonEmpty TokenMap
+removeBurnValueFromChangeMaps (assetId, assetQty) =
```
jonathanknowles: Suggestion: consider simplifying this function with the use of `TokenMap.adjustQuantity`.

I tested this implementation against your property tests, and it appears to work (and I think it executes more quickly, _possibly_ because we're not constructing a singleton map for each iteration, and also because we're no longer relying on `TokenMap.difference`):

```hs
removeBurnValueFromChangeMaps (assetId, assetQty) maps =
    NE.fromList $ burn assetQty (NE.toList maps) []
  where
    burn :: TokenQuantity -> [TokenMap] -> [TokenMap] -> [TokenMap] 
    burn _ [] alreadyProcessed =
        reverse alreadyProcessed
    burn gas (this : rest) alreadyProcessed 
        | TokenMap.getQuantity this assetId >= gas =
            reverse alreadyProcessed <> (thisReduced : rest)
        | otherwise =   
            burn gasReduced rest (thisReduced : alreadyProcessed)   
      where 
        gasReduced = gas
            `TokenQuantity.difference` TokenMap.getQuantity this assetId
        thisReduced = TokenMap.adjustQuantity this assetId  
            (`TokenQuantity.difference` gas)
```

Note this relies on the following function which is not yet defined in the `TokenQuantity` module:

```hs
difference :: TokenQuantity -> TokenQuantity -> TokenQuantity
difference a b = fromMaybe (TokenQuantity 0) $ subtract a b
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -1375,6 +1492,205 @@ makeChangeForCoin
...
+    burnWithGas assetQty
+  where
+      -- Subtract two token quantities, clamping negative values to zero.
+      subtract' :: TokenQuantity -> TokenQuantity -> TokenQuantity
+      subtract' x y =
+          fromMaybe TokenQuantity.zero (x `TokenQuantity.subtract` y)
```
jonathanknowles: Suggestion: extract this into a `difference` function in the `TokenQuantity` module, and add a short property test for it.

For other types, it's common to use `difference` to capture a subtraction that's bounded by some type-specific concept of "zero". For example:

- [`TokenMap.difference`](https://github.com/input-output-hk/cardano-wallet/blob/fe50727c01b4b07f98bcfdd36be81a67e4b2d40b/lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs#L544)
- [`TokenBundle.difference`](https://github.com/input-output-hk/cardano-wallet/blob/fe50727c01b4b07f98bcfdd36be81a67e4b2d40b/lib/core/src/Cardano/Wallet/Primitive/Types/TokenBundle.hs#L322)
- [`Map.difference`](https://hackage.haskell.org/package/containers-0.6.4.1/docs/Data-Map-Strict.html#v:difference)
- [`Set.difference`](https://hackage.haskell.org/package/containers-0.6.4.1/docs/Data-Set.html#v:difference)

Bonus points: use this new function to simplify the definition of [`TokenMap.difference`](https://github.com/input-output-hk/cardano-wallet/blob/fe50727c01b4b07f98bcfdd36be81a67e4b2d40b/lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap.hs#L544).

jonathanknowles (commented):
Hi @sevanspowell 

Many thanks for creating this PR.


I've not yet finished looking through it, but what I've seen so far looks good.

I've added a few comments and suggestions. Will continue looking tomorrow!

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
jonathanknowles: It's not necessarily a big problem. I'd be concerned if we had to reduce coverage to a very low percentage, but `25` seems fine here. 👍🏻

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -3558,6 +3559,13 @@ instance IsServerError ErrSelectAssets where
                         , "must specify enough ada. Here are the problematic "
                         , "outputs:\n" <> pretty (indentF 2 $ blockListF xs)
                         ]
+                OutputsInsufficient e ->
+                    apiError err403 MintedNotSpent $ mconcat
```
KtorZ: Isn't that a server error or do we expect clients to specify all the necessary outputs? I am asking because of the PR description which mentions a modification of the change generation to include minted value and I see arguments for both.

jonathanknowles: > do we expect clients to specify all the necessary outputs?

Yes. The current design expects all minted tokens to be spent in user-specified outputs.

But you're quite right, we could adjust the design to return minted-but-not-spent tokens back as change.

In my opinion, this is a bit of a gap in the requirements. See: https://github.com/input-output-hk/cardano-wallet/pull/2725#discussion_r655894252

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
sevanspowell: Yes sorry, that comment doesn't even reflect the implementation. The implementation treats minted/burnt tokens as user-specified if they are in the user-specified output set - as you said. I'll update the comment.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
sevanspowell: Addressed in df9d84252

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
sevanspowell: Addressed in 402615ace

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
sevanspowell: Addressed in dc69c81ef

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
sevanspowell: Addressed in dc69c81ef

lib/core/src/Cardano/Wallet/Api/Types.hs
sevanspowell: Addressed in 6902c9bf3

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
sevanspowell: Addressed in 059779d22

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
jonathanknowles: It's also possible to split up these functions like so:
```hs
removeBurnValueFromChangeMaps       
    :: (AssetId, TokenQuantity) -> NonEmpty TokenMap -> NonEmpty TokenMap   
removeBurnValueFromChangeMaps (assetId, assetQty) maps = maps       
    & fmap (`TokenMap.getQuantity` assetId) 
    & reduceTokenQuantities assetQty
    & NE.zipWith (`TokenMap.setQuantity` assetId) maps      
    
reduceTokenQuantities       
    :: TokenQuantity -> NonEmpty TokenQuantity -> NonEmpty TokenQuantity    
reduceTokenQuantities toBurn quantities =   
    NE.fromList $ burn toBurn (NE.toList quantities) []     
  where     
    burn _ [      ] ys = reverse ys 
    burn b (x : xs) ys      
        | x >= b = reverse ys <> (x' : xs)  
        | otherwise = burn b' xs (x' : ys)  
      where 
        b' = b `TokenQuantity.difference` x
        x' = x `TokenQuantity.difference` b    
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
sevanspowell: Addressed in 89f100c40, bonus points in 5f839c7bf

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
sevanspowell: Addressed in af3622d43 / 89f100c40

sevanspowell: Addressed follow-up comment in c6b3146

lib/core/src/Cardano/Wallet/Api/Server.hs
sevanspowell: Yeah, I'll have to chase up the requirements for this.

sevanspowell: For now, I think it's best to leave this as-is: require the user to explicitly send the tokens to an address. The requirements are a bit unclear on this point, and to allow users to not specify an address would require more work. Some parts of the code don't care at all (change bundle generation), but the selection algorithm does.

The API rejects requests to mint that don't also include an address to send them (so these requests are already stopped before the selection algorithm), but I've added an error the server could return if the API were to allow it, just in-case the behavior of the API is changed in future.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -3558,6 +3559,13 @@ instance IsServerError ErrSelectAssets where
                         , "must specify enough ada. Here are the problematic "
                         , "outputs:\n" <> pretty (indentF 2 $ blockListF xs)
                         ]
+                OutputsInsufficient e ->
+                    apiError err403 TokensMintedButNotSpentOrBurned $ mconcat
+                        [ "I can't process this transaction because some "
+                        , "minted values were not spent or burned. These "
+                        , "are the values that should be spent or burned: "
+                        , pretty . Flat $ missingOutputAssets e
+                        ]
```
jonathanknowles, line 3700: 👍🏻

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -175,6 +182,13 @@ data SelectionCriteria = SelectionCriteria
     , extraCoinSource
         :: !(Maybe Coin)
         -- ^ An optional extra source of ada.
+    , assetsToMint
+        :: !TokenMap
+        -- ^ When a token is minted, it provides an extra source of tokens.
+    , assetsToBurn
+        :: !TokenMap
+        -- ^ When a token is burned, the selection algorithm must find
+        -- inputs to satisfy the burn.
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -274,10 +288,44 @@ data SelectionError
...
+data OutputsInsufficientError = OutputsInsufficientError
+    { assetsToMint
+        :: !TokenMap
+      -- ^ The values to mint
+    , assetsToBurn
+        :: !TokenMap
+      -- ^ The values to burn
+    , requestedAssetOutputs
+        :: !TokenMap
+      -- ^ The outputs the requester asked to be covered
```
jonathanknowles: Hi @sevanspowell 

I think this should probably be `requestedOutputAssets`, rather than `requestedAssetOutputs`, as these are "the assets of the outputs", rather than "the outputs of the assets".


```suggestion
    , requestedOutputAssets
        :: !TokenMap
      -- ^ The balance of assets contained in the user-specified outputs
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -274,10 +288,44 @@ data SelectionError
...
+-- from a @OutputsInsufficientError@.
+missingOutputAssets :: OutputsInsufficientError -> TokenMap
+missingOutputAssets (OutputsInsufficientError mint burn out) =
+    -- We use difference, which will show us the quantities in "mint"
+    -- that are not in "out + burn". Any amount present in "out + burn",
+    -- but not present in "mint" will simply be zeroed out, which is
+    -- the behaviour we want for this error report.
+    mint
+        `TokenMap.difference`
+            (out `TokenMap.add` burn)
```
jonathanknowles: Nice! This is much simpler than before.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -1117,17 +1221,21 @@ makeChange criteria
...
 
     totalInputValue :: TokenBundle
-    totalInputValue = TokenBundle.add
-        (F.fold inputBundles)
-        (maybe TokenBundle.empty TokenBundle.fromCoin extraCoinSource)
+    totalInputValue =
+      F.fold inputBundles
+          `TokenBundle.add` F.foldMap TokenBundle.fromCoin extraCoinSource
+          -- Mints represent extra inputs from "the void"
+          `TokenBundle.add` TokenBundle.fromTokenMap assetsToMint
```
jonathanknowles:
```suggestion
    totalInputValue =
        F.fold inputBundles
            `TokenBundle.add` F.foldMap TokenBundle.fromCoin extraCoinSource
            -- Mints represent extra inputs from "the void"
            `TokenBundle.add` TokenBundle.fromTokenMap assetsToMint
```

Also, you could consider using `<>` here. It's already a common idiom within our codebase to use `<>` for adding together token bundles and token maps.

```suggestion
    totalInputValue =
        F.fold inputBundles
            <> F.foldMap TokenBundle.fromCoin extraCoinSource
            -- Mints represent extra inputs from "the void"
            <> TokenBundle.fromTokenMap assetsToMint
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -1375,6 +1495,196 @@ makeChangeForCoin
...
+addMintValueToChangeMaps (assetId, assetQty) = \case
+    -- This is the largest element, add mint value here
+    (x :| []) ->
+        TokenMap.add (TokenMap.singleton assetId assetQty) x :| []
+    -- There is a larget element, recurse to end of list
+    (x :| (y : rest)) ->
+        let
+            xs = y :| rest
+        in
+          x `NE.cons` addMintValueToChangeMaps (assetId, assetQty) xs
```
jonathanknowles: Suggestion: use [`over`](https://hackage.haskell.org/package/lens-5.0.1/docs/Control-Lens-Setter.html#v:over), [`last1`](https://hackage.haskell.org/package/lens-5.0.1/docs/Control-Lens-Combinators.html#v:last1), and `TokenMap.adjustQuantity` here:

```hs
 addMintValueToChangeMaps (assetId, assetQty) =
     Lens.over last1 $ \m -> TokenMap.adjustQuantity m assetId (<> assetQty)
```

Or, alternatively, if you'd prefer to avoid using lens:

```hs
addMintValueToChangeMaps (assetId, assetQty) =
    modifyLast $ \m -> TokenMap.adjustQuantity m assetId (<> assetQty)
  where
    modifyLast f xs = case NE.reverse xs of
        (y :| ys) -> NE.reverse (f y :| ys)
```

jonathanknowles (commented):
Hi Sam.

Some more comments. I'm about to look at the property tests. Will leave more comments later!

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
sevanspowell: Addressed in aada295

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
sevanspowell: Addressed in 400c7c316. I've gone and used the second solution, only because lens isn't already a dependency of cardano-wallet-core.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -2926,6 +3248,184 @@ prop_mapMaybe_oracle xs fn =
...
+    -> Property
+prop_addMintValueToChangeMaps_order mint numChangeMaps nonUserQtys = 
+    let
+      -- We can only maintain ascending partial order if there already is
+      -- ascending partial order. Random data provided by an arbitrary instance
+      -- isn't guaranteed to be in ascending partial order, and we can't easily
+      -- "sort it". One quick way to get ascending partial order is to make the
+      -- change maps ourself. Presuming of course
+      -- "makeChangeForNonUserSpecifiedAssets" is correctly implemented, which
+      -- other tests should confirm.
```
jonathanknowles: **Suggestion**: consider using [`NE.scanl`](https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-List-NonEmpty.html#v:scanl) to generate values in ascending partial order.

You're right that an arbitrary list of `TokenMap` values `[m1, m2, m3, ..., mi]` is not ordered.

However, if we treat the generated list as a list of **_differences_**, then we can transform those differences into a list that is **_guaranteed_** to be in ascending partial order:

```hs
[ m1
, m1 <> m2
, m1 <> m2 <> m3
, ...
, mi <> m2 <> m3 <> ... <> mi
]
```

**Full implementation:**
```hs
prop_addMintValueToChangeMaps_order
    :: (AssetId, TokenQuantity)
    -> NonEmpty TokenMap
    -> Property
prop_addMintValueToChangeMaps_order mint changeMapDiffs =
    property
        $ inAscendingPartialOrder
        $ addMintValueToChangeMaps mint changeMaps
  where
    -- A list of change maps already in ascending partial order.
    changeMaps = NE.scanl (<>) TokenMap.empty changeMapDiffs
```

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -2926,6 +3248,184 @@ prop_mapMaybe_oracle xs fn =
...
+      changeMaps =
+          makeChangeForNonUserSpecifiedAssets
+          numChangeMaps
+          nonUserQtys
+          TokenMap.empty
+          TokenMap.empty
+    in
+      inAscendingPartialOrder (removeBurnValueFromChangeMaps burn changeMaps)
+      ===
+      True
```
jonathanknowles: Similarly, we can also simplify this property with [`NE.scanl`](https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-List-NonEmpty.html#v:scanl):

```hs
prop_removeBurnValueFromChangeMaps_order
    :: (AssetId, TokenQuantity)
    -> NonEmpty TokenMap
    -> Property
prop_removeBurnValueFromChangeMaps_order burn changeMapDiffs =
    property
        $ inAscendingPartialOrder
        $ removeBurnValueFromChangeMaps burn changeMaps
  where
    changeMaps = NE.scanl (<>) TokenMap.empty changeMapDiffs
```

lib/core/src/Cardano/Wallet/Primitive/Types/TokenQuantity.hs
```diff
@@ -135,6 +136,10 @@ pred (TokenQuantity q) = TokenQuantity $ Prelude.pred q
 succ :: TokenQuantity -> TokenQuantity
 succ (TokenQuantity q) = TokenQuantity $ Prelude.succ q
 
+-- | Substract two token quantities, clamping negative values to zero.
+difference :: TokenQuantity -> TokenQuantity -> TokenQuantity
+difference x y = fromMaybe zero $ subtract x y
+
```
jonathanknowles, line 145: 👍🏻

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -1610,29 +1747,59 @@ isValidMakeChangeData p = (&&)
...
-    totalInputValue = TokenBundle.add
-        (F.fold $ inputBundles p)
-        (maybe TokenBundle.empty TokenBundle.fromCoin (view #extraCoinSource p))
-    totalOutputValue = F.fold $ outputBundles p
+    totalInputValue =
+        F.fold (inputBundles p)
+        `TokenBundle.add` (F.foldMap
+                           TokenBundle.fromCoin
+                           (view #extraCoinSource p))
+        `TokenBundle.add` TokenBundle.fromTokenMap (view #assetsToMint p)
```
jonathanknowles: Suggestion: use `<>` here:

```suggestion
        F.fold (inputBundles p)
            <> F.foldMap TokenBundle.fromCoin (view #extraCoinSource p)
            <> TokenBundle.fromTokenMap (view #assetsToMint p)
```

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -1873,11 +2071,18 @@ prop_makeChange_fail_minValueTooBig p =
...
-        (maybe TokenBundle.empty TokenBundle.fromCoin (view #extraCoinSource p))
+    totalInputValue = 
+        F.fold (inputBundles p)
+            `TokenBundle.add`
+                maybe
+                    TokenBundle.empty
+                    TokenBundle.fromCoin
+                    (view #extraCoinSource p)
+            `TokenBundle.add`
+                TokenBundle.fromTokenMap (view #assetsToMint p)
```
jonathanknowles: Suggestion: use `<>` and [`F.foldMap`](https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-Foldable.html#v:foldMap) here.

This relies on the `Foldable` instance for `Maybe`, and the `Monoid` instance for `TokenBundle`.

```suggestion
    totalInputValue = 
        F.fold (inputBundles p)
            <> F.foldMap TokenBundle.fromCoin (view #extraCoinSource p)
            <> TokenBundle.fromTokenMap (view #assetsToMint p)
```

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -2926,6 +3248,184 @@ prop_mapMaybe_oracle xs fn =
...
+-- after calling this function is equivalent of just removing the burnt value
+-- from the change maps.
+prop_removeBurnValueFromChangeMaps_value
+    :: (AssetId, TokenQuantity)
+    -> NonEmpty TokenMap
+    -> Property
+prop_removeBurnValueFromChangeMaps_value (assetId, qty) changeMaps =
+    F.fold changeMaps `TokenMap.difference` TokenMap.singleton assetId qty
+    ===
+    F.fold (removeBurnValueFromChangeMaps (assetId, qty) changeMaps)
```
jonathanknowles, line 3353: 👍🏻

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -2926,6 +3248,184 @@ prop_mapMaybe_oracle xs fn =
...
+-- the amount it was asked to.
+prop_reduceTokenQuantities_value
+    :: TokenQuantity -> NonEmpty TokenQuantity -> Property
+prop_reduceTokenQuantities_value reduceQty qtys =
+    qtyListSum qtys `TokenQuantity.difference` reduceQty
+    ===
+    qtyListSum (reduceTokenQuantities reduceQty qtys)
+    where
+        qtyListSum :: Foldable t => t TokenQuantity -> TokenQuantity
+        qtyListSum = TokenQuantity . getSum . F.foldMap (Sum . unTokenQuantity)
```
jonathanknowles: Suggestion: use [`F.fold`](https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-Foldable.html#v:fold) here to replace `qtyListSum`.

This works because `TokenQuantity` has a `Monoid` instance with additive semantics. (Hence we don't actually need to wrap it in `Sum`.)

```hs
prop_reduceTokenQuantities_value reduceQty qtys =
    F.fold qtys `TokenQuantity.difference` reduceQty
    ===
    F.fold (reduceTokenQuantities reduceQty qtys)
```

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -2926,6 +3248,184 @@ prop_mapMaybe_oracle xs fn =
...
+      tokenQtys = fmap (`TokenMap.getQuantity` assetId) changeMaps
+
+      inOrder :: Ord a => NonEmpty a -> Bool
+      inOrder (_ :| [])       = True
+      inOrder (x :| (y : xs)) = x <= y && inOrder (y :| xs)
+
+    in
+      inOrder (reduceTokenQuantities reduceQty tokenQtys)
+      ===
+      True
```
jonathanknowles: Suggestion: use [`NE.scanl`](https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-List-NonEmpty.html#v:scanl) to generate a list of quantities in ascending order:

```hs
prop_reduceTokenQuantities_order
    :: TokenQuantity
    -> NonEmpty TokenQuantity
    -> Property
prop_reduceTokenQuantities_order reduceQty qtyDiffs =
    property
        $ inAscendingOrder
        $ reduceTokenQuantities reduceQty qtys
  where
    -- Returns 'True' if (and only if) the given list is in ascending order.
    inAscendingOrder xs = NE.sort xs == xs

    -- A list of quantities already in ascending order.
    qtys = NE.scanl (<>) TokenQuantity.zero qtyDiffs
```

jonathanknowles (commented):
Hi @sevanspowell . Here's a first review of the property tests.

Will add more comments later!

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
sevanspowell: Addressed in e77bcc4cf.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
sevanspowell: Addressed in e77bcc4cf.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
sevanspowell: Where can I download your smarts? :) Addressed in e77bcc4cf.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
sevanspowell: Addressed in ec539305f.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
sevanspowell: Addressed in ec539305f.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
sevanspowell: Addressed in 455d96490.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -1610,29 +1741,57 @@ isValidMakeChangeData p = (&&)
...
 genMakeChangeData :: Gen MakeChangeData
 genMakeChangeData = flip suchThat isValidMakeChangeData $ do
     outputBundleCount <- choose (0, 15)
     let inputBundleCount = outputBundleCount * 4
+
+    inputBundles <- genTokenBundles inputBundleCount
+    outputBundles <- genTokenBundles outputBundleCount
+
+    -- Tokens in the input but not the output can be burned
+    assetsToBurn <- TokenMap.fromFlatList <$>
```
sevanspowell: As discussed with @jonathanknowles , I do not think these generators sufficiently cover the potential "universe" of MakeChangeData. They unnecessarily make the set of user-specified assets and the set of assets to burn disjoint.

I made a hacky PR to see how the tests would behave if I altered the generator: https://github.com/input-output-hk/cardano-wallet/pull/2740

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
```diff
@@ -567,17 +602,68 @@ genSelectionCriteria genUTxOIndex = do
           )
         ]
     extraCoinSource <- oneof [ pure Nothing, Just <$> genCoinSmall ]
+    assetsToBurn <- TokenMap.fromFlatList <$>
```
sevanspowell: As discussed with @jonathanknowles , I do not think these generators sufficiently cover the potential "universe" of SelectionCriteria. They unnecessarily make the set of user-specified assets and the set of assets to burn disjoint.

I made a hacky PR to see how the tests would behave if I altered the generator: https://github.com/input-output-hk/cardano-wallet/pull/2740

lib/core/src/Cardano/Wallet/Primitive/Types/TokenQuantity.hs
```diff
@@ -135,6 +136,10 @@ pred (TokenQuantity q) = TokenQuantity $ Prelude.pred q
 succ :: TokenQuantity -> TokenQuantity
 succ (TokenQuantity q) = TokenQuantity $ Prelude.succ q
 
+-- | Substract two token quantities, clamping negative values to zero.
```
jonathanknowles:
```suggestion
-- | Subtract the second token quantity from the first, clamping negative values to zero.
```

jonathanknowles: Fixed in 7456073f2d0ce6a90067c1c7319f79057229c70f.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
jonathanknowles: I've added a set of coverage checks to cover several possibilities in f7a93280fa4b3cbe0d070846351720fa02fc4435. I think this covers all the possibilities we're interested in, including that the user-specified assets and assets to burn are not always disjoint.

Out of these coverage checks, only one is not satisfied (namely, that we occasionally generate minted values that are not spent or burned). I've fixed this in f8d177742fabe84bfbe4966886581042dbd87f2f.

However, this revealed a faulty assertion in `onOutputsInsufficient`, which I have fixed in 36b2be9432f87e89a26995e3985677120e7cf266.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobinSpec.hs
jonathanknowles: Hi @sevanspowell 

I've added some extra coverage checks in 96b6a756f6217e5801ff3c647d8f36d1c30dd222.

It's not as bad as we expected: we do get some overlap between the sets of user specified assets and the sets of assets to burn.

Though we were missing coverage for the case where there's an overlap between the sets of assets to mint and the assets to burn. So I've tweaked the generator a bit to allow for this case in 378d35f5b37097e8814ced6801e5988213654a96.

jonathanknowles:
bors try

jonathanknowles (approved):
I've given this a fairly exhaustive review, and added some missing coverage to the test suite. I think it's good to go. 👍🏻

PR #2726, opened 2021-06-18
Require more recent version of `hspec` (>= 2.8.2).

jonathanknowles:
# Issue Number

ADP-978

# Overview

This PR upgrades `hspec` to a more recent version that supports the following options:

* `--times` prints an execution time for each individual spec.
* `--print-slow-items[=n]` prints the `n` slowest spec items at the end of a test run.


rvl (approved):
Neat

rvl:
That version of hspec will require an index-state update.
We may as well bump Haskell.nix for that.

PR #2724, opened 2021-06-18
Add Test.Utils.Pretty and verifyMsg

rvl:
### Issue Number

ADP-970

### Overview

- Add a more descriptive `verifyMsg` function to the integration tests.
- Move the `Pretty` helper newtype into `Test.Utils.Pretty`.

### Comments

This code was rescued from #2708.

PR #2727, opened 2021-06-18
Set -j 1 in macOS nix unit tests

Anviking:
# Issue Number

#2472 / ADP-970

# Overview

- [x] Set `-j 1` for hydra Mac builds, in hope this alleviates unit timeouts

# Comments

A hypothesis is that #2472 is caused by heavy load and unfocused
resources from running the tests concurrently, risking that the slowest
hspec runner - and thererefore the stdout - being silent for 900s causing
hydra to timeout.

Setting -j 1 should hopefully focus the resource we have in one place. It
should go silent less often, at the expense of the full run getting slower.

## Testing locally

Tested locally with `nix-build -A checks.cardano-wallet-core`. Seems to work.

It now takes 399s instead of 299.7877 on my machine (Not that dramatic slowdown).

We have a `-with-rtsopts=-N4` in the .cabal file. I'm not sure if that's good or not, but I suspect it doesn't matter as much as the `-j` option.

### DIFF
--- a/nix/haskell.nix
+++ b/nix/haskell.nix
@@ -86,6 +86,16 @@ let
       in {
         packages.cardano-wallet-core.components.tests = {
           unit.preCheck = noCacheOnBorsCookie;
+          # Attempt to ensure visible progress in the macOS hydra job.
+          #
+          # An hypothesis is that #2472 is caused by heavy load and unfocused
+          # resources from running the tests concurrently, risking that the slowest
+          # hspec runner - and thererefore the stdout - being silent for 900s causing
+          # hydra to timeout.
+          #
+          # Setting -j 1 should hopefully focus the resource we have in one place. It
+          # should go silent less often, at the expense of the full run getting slower.
+          unit.testFlags = lib.optionals pkgs.stdenv.hostPlatform.isDarwin ["-j" "1"];
         };
 
         packages.cardano-wallet.components.tests = {


nix/haskell.nix
```diff
@@ -104,6 +104,17 @@ let
...
+          # Attempt to ensure visible progress in the macOS hydra job.
+          #
+          # An hypothesis is that #2472 is caused by heavy load and unfocused
+          # resources from running the tests concurrently, risking that the slowest
+          # hspec runner - and thererefore the stdout - being silent for 900s causing
+          # hydra to timeout.
+          #
+          # Setting -j 1 should hopefully focus the resource we have in one place. It
+          # should go silent less often, at the expense of the full run getting slower.
+          unit.testFlags = if pkgs.stdenv.hostPlatform.isDarwin then ["-j" "1"] else [];
```
rvl: This is setting the `unit` test-suite flags for the `cardano-wallet` package, but I think you want the `cardano-wallet-core` package, right?

Try moving this block to line 88.

Anviking:
bors try

rvl:
I believe that the main reason why tests were getting stuck on macos more is the problem fixed by #2734.
That problem is a simple case of readMVar waiting forever for a putMVar that never happens.
The exception condition (port in use, address unavailable, etc) happened more with macos.

Anviking:
> I believe that the main reason why tests were getting stuck on macos more is the problem fixed by #2734.
That problem is a simple case of readMVar waiting forever for a putMVar that never happens.

The entire `LoggingSpec` was marked pending on macOS in #2709, after which we have observed one proper bors timeout in https://github.com/input-output-hk/cardano-wallet/pull/2724#issuecomment-864116545, and IIRC a couple more on normal PR non-bors builds. 

To me it seems like marking it pending had little to no effect.

> That problem is a simple case of readMVar waiting forever for a putMVar that never happens.

From `./scripts/bors-stats.rb list --tag "#2472" --details true` there are no failures looking like direct LoggingSpec timeouts. If the LoggingSpecs spawned a lot of threads, and lots of work, perhaps this could contribute to the first hspec thread being slow. "readMVar waiting forever" doesn't sound like a lot of work though.

I think`-j 1` is very much worth a shot.

nix/haskell.nix
```diff
@@ -86,6 +86,16 @@ let
...
```
rvl: This is very very minor, but just so you know, there's a handy [`lib.optionals`](https://nixos.org/manual/nixpkgs/stable/#function-library-lib.lists.optionals) function which does the same thing.
```suggestion
          unit.testFlags = lib.optionals pkgs.stdenv.hostPlatform.isDarwin ["-j" "1"];
```
Either way is perfectly fine though.

rvl (approved):
OK Thanks @Anviking

rvl:
> I think `-j 1` is very much worth a shot.

Fair enough.

bors r+

PR #2741, opened 2021-06-25
Adjust fee estimation to account for minting and burning.

sevanspowell:
# Issue Number

ADP-998
ADP-955

# Overview

This PR:

- [x] adjusts fee estimation to account for minted assets and scripts.
- [x] adds property and unit tests to cover changes to `estimateTxSize`.
- [x] makes small adjustments to integration test expectations for transaction fees.


lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -183,7 +197,7 @@ spec = do
         prop "roundtrip for Byron witnesses" prop_decodeSignedByronTxRoundtrip
 
     estimateMaxInputsTests @ShelleyKey
-        [(1,115),(5,106),(10,101),(20,85),(50,32)]
+        [(1,114),(5,106),(10,101),(20,85),(50,32)]
```
sevanspowell, line 201: I've changed the expected number of inputs, because my change to the fee estimation caused this test to fail, but I'll admit I don't quite understand the test. 

A small difference such as this in the expected result doesn't seem to be a cause for concern, but if anyone has more knowledge on this I'd love to hear it.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -48,6 +49,10 @@ import Prelude
 
 import Cardano.Address.Derivation
     ( XPrv, toXPub )
+import Cardano.Address.Script
+    ( KeyHash
+    , Script (ActiveFromSlot, ActiveUntilSlot, RequireAllOf, RequireAnyOf, RequireSignatureOf, RequireSomeOf)
```
rvl: Importing all constructors is usually fine. A list of specific constructors just gets messy and tedious to update.
```suggestion
    , Script (..)
```

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -94,7 +99,7 @@ import Cardano.Wallet.Primitive.Types.TokenMap
 import Cardano.Wallet.Primitive.Types.TokenPolicy
     ( TokenName (..) )
 import Cardano.Wallet.Primitive.Types.TokenQuantity
-    ( TokenQuantity (..) )
+    ( TokenQuantity (TokenQuantity) )
```
rvl:
```suggestion
    ( TokenQuantity (..) )
```

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -896,6 +928,21 @@ estimateTxSize skeleton =
         sizeOf_MetadataHash
             = maybe 0 (const (sizeOf_SmallUInt + sizeOf_Hash32)) txMetadata
 
+        -- ?8 => uint ; validity interval start
+        sizeOf_ValidityIntervalStart
+          = sizeOf_UInt
```
rvl: About this one - is it for the validity interval of the transaction as a whole or, is it an interval which is part of a script?
Are the validity interval bounds optional? What about the size of the validity interval end?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -896,6 +928,21 @@ estimateTxSize skeleton =
...
+
+        -- ?9 => mint = multiasset<int64>
+        -- multiasset<a> = { * policy_id => { * asset_name => a } }
+        -- policy_id = scripthash
+        -- asset_name = bytes .size (0..32)
+        sizeOf_Mint AssetId{tokenName}
```
rvl, line 950: Could `sizeOf_Mint` and `sizeOf_NativeAsset` be refactored and parameterised by the inner type?
Like how it's done in the `shelley-ma.cddl` file with `multiasset<a>`.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -827,6 +855,8 @@ estimateTxSize skeleton =
     --   , ? 5 : withdrawals
     --   , ? 6 : update
     --   , ? 7 : metadata_hash
+    --   , ? 8 : uint ; validity interval start
+    --   , ? 9 : mint
```
rvl, line 872: These are things added in shelley-ma.
There is a link to the source `shelley.cddl` file above.
Can we also add a link to `shelley-ma.cddl`?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -1107,6 +1178,11 @@ estimateTxSize skeleton =
     sizeOf_SmallArray = 1
     sizeOf_Array = 3
 
+-- Small helper function for summing values. Given a list of values, get the sum
+-- of the values, after the given function has been applied to each value.
+sumVia :: (Foldable t, Num m) => (a -> m) -> t a -> m
```
rvl, line 1212: Nice

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -1063,6 +1109,26 @@ estimateTxSize skeleton =
...
+    --   // script_any        = (2, [ * native_script ])
+    --   // script_n_of_k     = (3, n: uint, [ * native_script ])
+    --   // invalid_before    = (4, uint)
+    --      ; Timelock validity intervals are half-open intervals [a, b).
+    --      ; This field specifies the left (included) endpoint a.
+    --   // invalid_hereafter = (5, uint)
+    --      ; Timelock validity intervals are half-open intervals [a, b).
+    --      ; This field specifies the right (excluded) endpoint b.
+    --   ]
+    sizeOf_NativeScript = \case
```
rvl, line 1142: Good to add this into the size estimation, because I think it would also be needed for spending from a multi-sig wallet.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -1094,6 +1160,11 @@ estimateTxSize skeleton =
     sizeOf_UInt = 5
     sizeOf_LargeUInt = 9
 
+    -- A CBOR Int which is less than 23 in value fits on a single byte. Beyond,
+    -- the first byte is used to encode the number of bytes necessary to encode
+    -- the number, followed by the number itself.
+    sizeOf_LargeInt = 9
```
rvl: Would it be better to call this `sizeOf_Int64`? Because that's the max size for minting.

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
rvl: I guess that the max inputs should decrease if you have caused all fee estimations to increase.
I wonder why only the first test case needed adjustment though.

rvl (approved):
Nice work - tests look good.
I have some change requests, so please merge once they're done.

P.S. Squash commits together where it make sense

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -714,6 +719,8 @@ data TxSkeleton = TxSkeleton
     , txInputCount :: !Int
     , txOutputs :: ![TxOut]
     , txChange :: ![Set AssetId]
+    , txScripts :: [Script KeyHash]
+    , txMintAssets :: [AssetId]
```
jonathanknowles: Hi @sevanspowell.

Does this field refer only to minted assets, or can it also contain burned assets?

If it only refers to minted assets, then will you need to add another field for burned assets?

If it also contains burned assets, then what happens if there are assets that are both minted and burned? (This is allowed by coin selection, but I'm not sure if it's planned to allow this in the API, so asking just in case!)

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -794,13 +808,27 @@ estimateTxSize skeleton =
     numberOf_Withdrawals
         = if txRewardWithdrawal > Coin 0 then 1 else 0
 
+    -- Total number of signatures the scripts require
+    numberOf_ScriptVkeyWitnesses
+      = sumVia scriptRequiredKeySigs txScripts
```
jonathanknowles:
```suggestion
    numberOf_ScriptVkeyWitnesses
        = sumVia scriptRequiredKeySigs txScripts
```

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -794,13 +808,27 @@ estimateTxSize skeleton =
...
+
+    scriptRequiredKeySigs :: Num num => Script KeyHash -> num
+    scriptRequiredKeySigs (RequireSignatureOf _) = 1
+    scriptRequiredKeySigs (RequireAllOf ss)      = sumVia scriptRequiredKeySigs ss
+    scriptRequiredKeySigs (RequireAnyOf ss)      = sumVia scriptRequiredKeySigs ss
+    scriptRequiredKeySigs (ActiveFromSlot _)     = 0
+    scriptRequiredKeySigs (ActiveUntilSlot _)    = 0
+    -- We don't know how many the user will sign with, so we just assume the worst case of "signs with all".
+    scriptRequiredKeySigs (RequireSomeOf _ ss)   = sumVia scriptRequiredKeySigs ss
```
jonathanknowles: Request: follow the coding standards here (line length + avoid vertical alignment).

Suggestion: use `LambdaCase` to resolve this:
```hs
    scriptRequiredKeySigs :: Num num => Script KeyHash -> num
    scriptRequiredKeySigs = \case
        RequireSignatureOf _ ->
            1
        RequireAllOf ss ->
            sumVia scriptRequiredKeySigs ss
        RequireAnyOf ss ->
            sumVia scriptRequiredKeySigs ss
        ActiveFromSlot _ ->
            0
        ActiveUntilSlot _ ->
            0
        RequireSomeOf _ ss ->
            -- We don't know how many the user will sign with, so we just assume
            -- the worst case of "signs with all".
            sumVia scriptRequiredKeySigs ss
```

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
jonathanknowles:
```suggestion
        sizeOf_ValidityIntervalStart
            = sizeOf_UInt
```

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
jonathanknowles, line 950: @sevanspowell  is there a `sizeOf_Burn` function to accompany `sizeOf_Mint`?

Or is it the intention to put everything into the same field?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -1027,9 +1073,9 @@ estimateTxSize skeleton =
                 then sizeOf_Array + sizeOf_SmallUInt else 0)
             + sizeOf_VKeyWitness * numberOf_VkeyWitnesses
 
-        -- ?1 => [* multisig_script ]
-        sizeOf_MultisigScript
-            = 0
+        -- ?1 => [* native_script ]
+        sizeOf_NativeScripts [] = 0
+        sizeOf_NativeScripts ss = sizeOf_Array + sizeOf_SmallUInt + sumVia sizeOf_NativeScript ss
```
jonathanknowles: Request: break up this long line.

Suggestion: match the style used elsewhere in the `estimateTxSize` function:

```suggestion
        sizeOf_NativeScripts ss
            = sizeOf_Array
            + sizeOf_SmallUInt
            + sumVia sizeOf_NativeScript ss
```

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -1063,6 +1109,26 @@ estimateTxSize skeleton =
...
+    --      ; Timelock validity intervals are half-open intervals [a, b).
+    --      ; This field specifies the right (excluded) endpoint b.
+    --   ]
+    sizeOf_NativeScript = \case
+        RequireSignatureOf _ -> sizeOf_SmallUInt + sizeOf_Hash28
+        RequireAllOf ss      -> sizeOf_SmallUInt + sizeOf_Array + sumVia sizeOf_NativeScript ss
+        RequireAnyOf ss      -> sizeOf_SmallUInt + sizeOf_Array + sumVia sizeOf_NativeScript ss
+        RequireSomeOf _ ss   -> sizeOf_SmallUInt + sizeOf_UInt + sizeOf_Array + sumVia sizeOf_NativeScript ss
+        ActiveFromSlot _     -> sizeOf_SmallUInt + sizeOf_UInt
+        ActiveUntilSlot _    -> sizeOf_SmallUInt + sizeOf_UInt
```
jonathanknowles: Request: follow the coding standards here (line length + avoid vertical alignment).

Suggestion:
```hs
        RequireSignatureOf _ ->
            sizeOf_SmallUInt + sizeOf_Hash28
        RequireAllOf ss ->
            sizeOf_SmallUInt + sizeOf_Array + sumVia sizeOf_NativeScript ss
        RequireAnyOf ss ->
            sizeOf_SmallUInt + sizeOf_Array + sumVia sizeOf_NativeScript ss
        RequireSomeOf _ ss ->
            sizeOf_SmallUInt
                + sizeOf_UInt
                + sizeOf_Array
                + sumVia sizeOf_NativeScript ss
        ActiveFromSlot _ ->
            sizeOf_SmallUInt + sizeOf_UInt
        ActiveUntilSlot _ ->
            sizeOf_SmallUInt + sizeOf_UInt
```

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -239,6 +259,110 @@ spec = do
...
+            in
+                (if null scripts
+                 then property $ marginalCost == 0
+                 else property $ marginalCost > 0
+                ) & classify (null scripts) "null scripts"
+                & counterexample ("marginal cost: " <> show marginalCost)
+                & counterexample ("cost with: " <> show costWith)
+                & counterexample ("cost without: " <> show costWithout)
+
+        it "increasing mint increases tx size at least proportionate to asset names"
```
jonathanknowles:
```suggestion
        it "increasing mint increases tx size at least proportionally to asset names"
```

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -239,6 +259,110 @@ spec = do
...
+                    minFeeSkeleton emptyTxSkeleton
+
+                marginalCost :: Integer
+                marginalCost = costWith - costWithout
+            in
+                (if null assets
+                 then property $ marginalCost == 0
+                 else property $ marginalCost > 0
+                ) & classify (null assets) "null minting assets"
+                & counterexample ("marginal cost: " <> show marginalCost)
```
jonathanknowles:
```suggestion
                (if null assets
                    then property $ marginalCost == 0
                    else property $ marginalCost > 0
                )
                & classify (null assets) "null minting assets"
                & counterexample ("marginal cost: " <> show marginalCost)
```

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -239,6 +259,110 @@ spec = do
...
+                    minFeeSkeleton emptyTxSkeleton
+
+                marginalCost :: Integer
+                marginalCost = costWith - costWithout
+            in
+                (if null scripts
+                 then property $ marginalCost == 0
+                 else property $ marginalCost > 0
+                ) & classify (null scripts) "null scripts"
+                & counterexample ("marginal cost: " <> show marginalCost)
```
jonathanknowles:
```suggestion
                (if null scripts
                    then property $ marginalCost == 0
                    else property $ marginalCost > 0
                )
                & classify (null scripts) "null scripts"
                & counterexample ("marginal cost: " <> show marginalCost)
```

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -239,6 +259,110 @@ spec = do
...
+                marginalSize = sizeWith - sizeWithout
+            in
+                -- Larger asset names means more bytes in the tx which should
+                -- mean a more expensive tx. Adding the mints should increase
+                -- the marginal size at least as much as the size of the asset
+                -- names.
+                property (marginalSize >= lengthAssetNames)
+                & classify (null mints) "null minting assets"
+                & counterexample ("asset names length: "
+                                  <> show lengthAssetNames)
```
jonathanknowles:
```suggestion
                & counterexample
                    ("asset names length: " <> show lengthAssetNames)
```

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -239,6 +259,110 @@ spec = do
...
+                -- the tx will need to be witnessed by those vkeys (in the worst
+                -- case).
+                property
+                  (marginalSize >= scriptLengths + numWitnesses * sizeWitness)
+                & classify (null scripts) "no scripts"
+                & classify (scriptLengths == 0) "zero script lengths"
+                & classify (numWitnesses == 0) "no witnesses"
+                & counterexample ("script lengths: " <> show scriptLengths)
+                & counterexample ("witness size: "
+                                  <> show (numWitnesses * sizeWitness))
```
jonathanknowles:
```suggestion
                & counterexample
                    ("witness size: " <> show (numWitnesses * sizeWitness))
```

jonathanknowles (changes requested):
Hi @sevanspowell 

Thanks for making this PR!

My main question relates to burning: is support for burned assets included in this PR?

I've done an initial review. I still have a bit of work to do to understand the test suite additions. I'll have another look tomorrow morning!

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: 1. This is the validity interval of the transaction as a whole, separate validity intervals exist for the scripts (see [here](https://github.com/input-output-hk/cardano-wallet/blob/5517e3f699aaeeb78fe3abd5bf7cc41042dc83fb/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs#L1117)).

2. Shelley and later eras support an upper bound on the validity range, this is known as the TTL, so it is already part of the fee estimation (https://github.com/input-output-hk/cardano-wallet/blob/5517e3f699aaeeb78fe3abd5bf7cc41042dc83fb/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs#L866). Allegra and later eras support a lower bound on the validity range, and that's what I'm adding here.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: Thanks :) Addressed in b8cbd8a82

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: Addressed in 5dbf25e6a.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: Addressed in 9c64ab8be

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: Addressed in 73caf085c.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: Addressed in 51d5f9371.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: So both ends of the validity bound are present, just one end is aliased as "TTL".

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: Addressed in 73caf085c.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: Addressed in 73caf085c.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: Addressed in 73caf085c.

sevanspowell: Addressed in 73caf085c.

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
sevanspowell: Addressed in 73caf085c.

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
sevanspowell: Addressed in 73caf085c.

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
sevanspowell: Addressed in 73caf085c.

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
sevanspowell: Addressed in 73caf085c.

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
sevanspowell: Addressed in 73caf085c.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: There is some somewhat confusing terminology here. `cardano-node` just refers to the minting and burning field as "mint". So this refers to both minted and burned assets.

When constructing a transaction manually, mints look something like `+3 deadbeef.aaaaaa`, whereas burns look something like `-3 deadbeef.aaaaaa`. If an asset is both minted and burned, it will just be added twice to the `txMintAssets` list. This is what we want because the Tx construction will do that same, add some data for the mint, add some data for the burn. From the perspective of tx size estimation, negative (burned) and positive (minted) quantities are no different, they are both represented with an `int64`. We don't have to do any fancy logic here, mints and burns are both treated as (qty, policyId, assetName), as they will be in the final Tx.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: See https://github.com/input-output-hk/cardano-wallet/pull/2741#discussion_r669255695

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: I've improved the terminology and added a comment to explain the usage of the field: c0de116.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
rvl: This suggestion makes a small function consume a quite unnecessary amount of vertical space. :stuck_out_tongue: 

You could also keep it under 80 characters, and vertically compact too, just by renaming the `ss` variable to `s`, and/or by removing the column alignment on `=`.

rvl (approved):
Thanks @sevanspowell

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: Easier to scroll down than sideways? :P

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -714,6 +717,12 @@ data TxSkeleton = TxSkeleton
     , txInputCount :: !Int
     , txOutputs :: ![TxOut]
     , txChange :: ![Set AssetId]
+    , txScripts :: [Script KeyHash]
+    , txMintBurnAssets :: [AssetId]
+    -- ^ Assets that have been both minted and burned, or minted or burned
+    -- multiple times, will appear multiple times in this list, once for each
+    -- mint or burn. For example if the caller "mints 3" of asset id "A", and
+    -- "burns 3" of asset id "A", "A" will appear twice in the list.
```
jonathanknowles, line 727: Thanks, this is a really nice and clear explanation. 👍🏻

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
jonathanknowles: > Addressed in 73caf08.

Hi Sam, I think this commit may have been dropped, perhaps accidentally?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: https://github.com/input-output-hk/cardano-wallet/pull/2741/files#diff-6218e625cd8c0f10e842ef1c10eb04bebf307ccecad42038e3b12b287a763690R818

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: https://github.com/input-output-hk/cardano-wallet/commit/73caf085cb7144f29e933a18e72dd60d955d041d#diff-6218e625cd8c0f10e842ef1c10eb04bebf307ccecad42038e3b12b287a763690R941

sevanspowell: https://github.com/input-output-hk/cardano-wallet/commit/73caf085cb7144f29e933a18e72dd60d955d041d#diff-6218e625cd8c0f10e842ef1c10eb04bebf307ccecad42038e3b12b287a763690R941

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: https://github.com/input-output-hk/cardano-wallet/pull/2741/files#diff-6218e625cd8c0f10e842ef1c10eb04bebf307ccecad42038e3b12b287a763690R948

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: https://github.com/input-output-hk/cardano-wallet/pull/2741/files#diff-6218e625cd8c0f10e842ef1c10eb04bebf307ccecad42038e3b12b287a763690R787

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: https://github.com/input-output-hk/cardano-wallet/pull/2741/files#diff-6218e625cd8c0f10e842ef1c10eb04bebf307ccecad42038e3b12b287a763690R1184

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: https://github.com/input-output-hk/cardano-wallet/pull/2741/files#diff-6218e625cd8c0f10e842ef1c10eb04bebf307ccecad42038e3b12b287a763690R721

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: https://github.com/input-output-hk/cardano-wallet/pull/2741/files#diff-6218e625cd8c0f10e842ef1c10eb04bebf307ccecad42038e3b12b287a763690R816

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
sevanspowell: https://github.com/input-output-hk/cardano-wallet/pull/2741/files#diff-6218e625cd8c0f10e842ef1c10eb04bebf307ccecad42038e3b12b287a763690R944

jonathanknowles (changes requested):

jonathanknowles (approved):

jonathanknowles (changes requested):
Hi @sevanspowell 

I think we might need to make further changes to this PR.

According to the integration test failure log ([example](https://hydra.iohk.io/build/6925587/nixlog/1)), it seems that the fee estimates for ordinary transactions have increased, even for those that don't involve minting or burning.

I'm not sure if this is intentional or not.

Perhaps we need to look at the `estimateTxSize` function again, to see why estimates are increasing even when there are no minted or burned tokens.

OTOH, if this is intentional, then we'll probably need to adjust the expectations of the failing integration tests. This isn't necessarily a bad thing, as it gives us an opportunity to check (with a human pair of eyes) that the fees still remain reasonable.

Let me know if I can be of any assistance!

sevanspowell:
> Hi @sevanspowell
> 
> I think we might need to make further changes to this PR.
> 
> According to the integration test failure log ([example](https://hydra.iohk.io/build/6925587/nixlog/1)), it seems that the fee estimates for ordinary transactions have increased, even for those that don't involve minting or burning.
> 
> I'm not sure if this is intentional or not.
> 
> Perhaps we need to look at the `estimateTxSize` function again, to see why estimates are increasing even when there are no minted or burned tokens.
> 
> OTOH, if this is intentional, then we'll probably need to adjust the expectations of the failing integration tests. This isn't necessarily a bad thing, as it gives us an opportunity to check (with a human pair of eyes) that the fees still remain reasonable.
> 
> Let me know if I can be of any assistance!

Yes, the base fee estimation will increase somewhat, due to the presence of the new "validity interval start" field. The size of the field is 5 bytes. I'm verifying now that this is the only cause for the increase in fees.

EDIT: I've verified that the cardano-wallet integration test passes when the validity interval is commented out:

```
    sizeOf_TransactionBody
        = sizeOf_SmallMap
        + sizeOf_Inputs
        + sizeOf_Outputs
        + sizeOf_Fee
        + sizeOf_Ttl
        + sizeOf_Certificates
        + sizeOf_Withdrawals
        + sizeOf_Update
        + sizeOf_MetadataHash
        -- + sizeOf_ValidityIntervalStart
        + sumVia sizeOf_Mint txMintBurnAssets
```

I think this verifies that the cause of the increase in fees is solely due to the presence of the new validity interval start, for which we'd expect an increase in fees. I'll modify the test expectations accordingly.

Anviking:
I added `#expected` annotations to the last bors failures as it sounded like they were caused by the PR itself. Feel free to change if incorrect.

jonathanknowles (approved):
LGTM!

jonathanknowles (changes requested):
Hi @sevanspowell!

It looks as though these tests are still failing:

https://github.com/input-output-hk/cardano-wallet/pull/2741#issuecomment-881154662

I can also reliably reproduce these failures locally, with:
```
stack build --fast --test cardano-wallet:test:integration --test-arguments '--match "STAKE_POOLS"'
```

From a quick look, these tests appear to probe the boundaries of fee calculations. So it's somewhat expected that these might fail if our fee calculations change.

Let me know if I can provide any assistance!

jonathanknowles (approved):
👍🏻

Issue #2730, opened 2021-06-21
Cardano wallet's threads are dying while syncing


Fell-x27:
# Context
There is a cardano wallet instance with few hundreds wallets connected to.
But you can't syncronize it, because threads are unstable.

## Essential Information

 - **Version**:      v2021-06-11 (git revision: 1532352c75cfe02015873c4c82b0d76eb798dfd8)
 - **Platform**:      LInux
 - **Installation**:  Binaries from release
 - **Network**:      mainnet 

logs piece with an error:

```
[cardano-wallet.wallet-db:Info:1282] [2021-06-21 18:44:56.56 UTC] Starting connection pool for ../../mainnet-wallets/ica.###somewalletid#####.sqlite
[cardano-wallet.wallet-engine:Info:1285] [2021-06-21 18:44:56.57 UTC] 1436aa99: Chain following starting. Requesting intersection using 12 checkpoints, the latest being 6c3f040d<-[bd1b1e9c-32734691#5881200]
[cardano-wallet.wallet-db:Info:1282] [2021-06-21 18:44:56.61 UTC] Closing single database connection (../../mainnet-wallets/ica.###somewalletid#####.sqlite)

[cardano-wallet.wallet-engine:Error:1282] [2021-06-21 18:44:56.61 UTC] Worker has exited: Unhandled exception: user error (Pattern match failure in do expression at src/Cardano/Wallet/DB/Sqlite.hs:596:17-26)

[cardano-wallet.wallet-db:Info:1282] [2021-06-21 18:44:56.61 UTC] Stopping database connection pool ../../mainnet-wallets/ica.###somewalletid#####.sqlite
[cardano-wallet.wallet-db:Info:1298] [2021-06-21 18:44:56.61 UTC] Starting connection pool for ../../mainnet-wallets/ica.###somewalletid#####.sqlite

```

# Steps to Reproduce

1. launch cardano wallet over the node with required version (1.27 for this case)
2. It has to have many wallets attached to. About 400 is enough.

Start command: 
```
cardano-wallet serve --listen-address 127.0.0.1 --port 8100 --node-socket ../../cardano.socket --database ../../mainnet-wallets/ --sync-tolerance 300s --pool-metadata-fetching https://smash.cardano-mainnet.iohk.io --mainnet 

```

## Expected behavior

Wallets are syncing.

## Actual behavior

Wallets are going to bloody pit wit message:
```
[...]

PR #2745, opened 2021-06-30
Handle node disconnections

Anviking:
# Issue Number

ADP-871

# Overview

- [x] Ensure follow catches asyncronous exceptions from connectClient,
such that it can restart with a new connection and cursor.

- [x] Keep Local State Queries and to-be-submitted Txs queued until their
requests finish, not just when they start. If a query is interrupted by
the node being disconnected, it will block until a connection is
re-established, and then retry.

- [ ] Remove need to catch async exceptions
    - Think this is a good opportunity to refactor / simplify all the chain-following code.
- [ ] Verify ouroboros-network errors are logged

# Comments

Integration tests could probably be written, but it would:
- take some time better spend on other tasks
- add complexity to the already complex cluster setup

so I'd lean towards it not being worth it at this moment.

I tested (see below) manually though, which could be added as a documented .md test if desired.

### Interrupted chain-sync & wallet listing
1. start node and wallet
2. list wallets => [wallet1]
3. kill node
4. list wallets => [wallet1] (previously [], or crash)
5. start node 
6. wallet resumes syncing
7. list wallets => [wallet1]

### Interrupted Local State Query
1. Start node and wallet
2. `cardano-wallet stake-pool list --stake 0`
3. kill node
4. observe that nothing happens
5. re-start node
6. stake-pool list request finishes after a while

### DIFF
--- a/lib/core/src/Cardano/Wallet/Network.hs
+++ b/lib/core/src/Cardano/Wallet/Network.hs
@@ -108,7 +108,7 @@ import UnliftIO.Async
 import UnliftIO.Concurrent
     ( threadDelay )
 import UnliftIO.Exception
-    ( SomeException, bracket, handle )
+    ( SomeException, bracket, handleSyncOrAsync )
 
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Text as T
@@ -378,26 +378,33 @@ follow'
     -- ^ Getter on the abstract 'block' type
     -> IO FollowExit
 follow' nl tr cps yield header =
-    bracket (initCursor nl cps) (destroyCursor nl) (sleep 0 False)
+    bracket
+        (initCursor nl cps)
+        (destroyCursor nl)
+        (handleExceptions . sleep 0 False)
   where
     innerTr = contramap MsgFollowLog tr
     delay0 :: Int
     delay0 = 500*1000 -- 500ms
 
+    handleExceptions :: IO FollowExit -> IO FollowExit
+    handleExceptions =
+        -- Node disconnections are seen as async exceptions from here. By
+        -- catching them, `follow` will try to establish a new connection
+        -- depending on the `FollowExceptionRecovery`.
+        handleSyncOrAsync (traceException *> const (pure FollowFailure))
+      where
+        traceException :: SomeException -> IO ()
+        traceException e = do
+            traceWith tr $ MsgUnhandledException $ T.pack $ show e
+
     -- | Wait a short delay before querying for blocks again. We also take this
     -- opportunity to refresh the chain tip as it has probably increased in
     -- order to refine our syncing status.
     sleep :: Int -> Bool -> Cursor -> IO FollowExit
-    sleep delay hasRolledForward cursor = handle exitOnAnyException $ do
-        when (delay > 0) (threadDelay delay)
-        step hasRolledForward cursor
-      where
-        -- Any unhandled synchronous exception should be logged and cause the
-        -- chain follower to exit.
-        exitOnAnyException :: SomeException -> IO FollowExit
-        exitOnAnyException e = do
-            traceWith tr $ MsgUnhandledException $ T.pack $ show e
-            pure FollowFailure
+    sleep delay hasRolledForward cursor = do
+            when (delay > 0) (threadDelay delay)
+            step hasRolledForward cursor
 
     step :: Bool -> Cursor -> IO FollowExit
     step hasRolledForward cursor = nextBlocks nl cursor >>= \case
--- a/lib/core/src/Ouroboros/Network/Client/Wallet.hs
+++ b/lib/core/src/Ouroboros/Network/Client/Wallet.hs
@@ -62,6 +62,7 @@ import Control.Monad.Class.MonadSTM
     , atomically
     , isEmptyTQueue
     , newEmptyTMVarIO
+    , peekTQueue
     , putTMVar
     , readTQueue
     , takeTMVar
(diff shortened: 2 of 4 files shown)


lib/core/src/Cardano/Wallet/Network.hs
```diff
@@ -378,26 +378,33 @@ follow'
...
     innerTr = contramap MsgFollowLog tr
     delay0 :: Int
     delay0 = 500*1000 -- 500ms
 
+    handleExceptions :: IO FollowExit -> IO FollowExit
+    handleExceptions =
+        -- Node disconnections are seen as async exceptions from here. By
+        -- catching them, `follow` will try to establish a new connection
+        -- depending on the `FollowExceptionRecovery`.
+        handleSyncOrAsync (traceException *> const (pure FollowFailure))
```
rvl, line 395: I really doubt that we need to handle asynchronous exceptions.
Just `UnliftIO.Exception.handleAny` (any synchronous exception) should be enough.

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -595,9 +609,14 @@ localTxSubmission queue =
   where
     clientStIdle
         :: m (LocalTxClientStIdle tx err m Void)
-    clientStIdle = atomically (readTQueue queue) <&> \case
+    clientStIdle = atomically (peekTQueue queue) <&> \case
```
rvl, line 612: It might be easier to keep using `readTQueue` but then put an `unGetTQueue` in an `onException` handler.

rvl (commented):
Looks promising...

lib/core/src/Cardano/Wallet/Network.hs
Anviking: We already _used to_ catch synchronous exceptions.

Note that chain-sync is run on a separate thread from `follow`, connected via [`link`](https://github.com/input-output-hk/cardano-wallet/blob/2738c71c44466d99804acb455a15ae8a5cb51cd7/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs#L422), which seems to indeed cause async exceptions to be thrown:

```haskell
λ> f = async (threadDelay 1000000 >> throwIO (userError "hi")) >>= link >> threadDelay 3000000

λ> f `catchSyncOrAsync` (\(e::SomeException) -> putStrLn "Caught")
Caught

λ>  f `catchAny` (\_e -> putStrLn "Caught")
*** Exception: ExceptionInLinkedThread (ThreadId 988) user error (hi)
```

rvl: OK, then we should either be catching the exceptions in the chain sync thread, or not using `link`. Trying to catch exceptions from other threads is a recipe for disaster.

Anviking: Ok 👍

piotr-iohk:
This looks to be improving also ADP-1013 / https://github.com/input-output-hk/cardano-wallet/issues/2730  (Wallet workers die while syncing when there are many wallets).

Having 240+ wallets I don't see workers dying. Although there are in the log `progress decreased` entires:
```
bed93eb5: Syncing (1.64% percent) Applied 0 blocks, 0 rollbacks in the last 30.213246312s. Currently at slot 50031. (progress decreased)
```
Perhaps it is due to the fact there are a lot of wallets syncing at the same time and therefore syncing is much slower.
There are also entries saying wallet looses connection with the node:
```
[cardano-wallet.network:Warning:9858] [2021-07-01 11:17:33.39 UTC] Connection lost with the node.
```
However wallets seem to be operational.

Checked on rev: v2021-06-11 (git revision: 0b5a22840dd049e021a7cd916bc354febbcf5739).

piotr-iohk:
However, one thing I've noticed is deleting a wallet (even when there's only one wallet) results in DB "lock" pretty much every time:
```
[cardano-wallet.wallet-db:Info:174] [2021-07-01 13:29:42.12 UTC] Waiting for 1 withDatabase 075fde25...7f11b717 call(s) to finish
[cardano-wallet.wallet-db:Info:174] [2021-07-01 13:29:42.17 UTC] Waiting for 1 withDatabase 075fde25...7f11b717 call(s) to finish
[cardano-wallet.wallet-db:Info:174] [2021-07-01 13:29:42.22 UTC] Waiting for 1 withDatabase 075fde25...7f11b717 call(s) to finish
...
[cardano-wallet.wallet-db:Notice:174] [2021-07-01 13:31:38.75 UTC] Timed out waiting for 1 withDatabase 075fde25...7f11b717 call(s) to finish. Attempting to remove the database anyway.
[cardano-wallet.wallet-db:Info:174] [2021-07-01 13:31:38.75 UTC] Removing wallet's database. Wallet id was 075fde25...7f11b717
[cardano-wallet.api-server:Error:174] [2021-07-01 13:31:38.75 UTC] [RequestId 8] DELETE /v2/wallets/075fde25bfe36db533c3f91ad1ecbc657f11b717 204 No Content in 116.628722893s

```

Anviking:
> However, one thing I've noticed is deleting a wallet (even when there's only one wallet) results in DB "lock" pretty much every time:

In what context? Disconnected node or lots of wallets?

piotr-iohk:
> > However, one thing I've noticed is deleting a wallet (even when there's only one wallet) results in DB "lock" pretty much every time:
> 
> In what context? Disconnected node or lots of wallets?

Nope, if you just:
 - create single wallet
 - delete wallet

and wallet deletion takes very long time with above entires in the log.

Anviking:
Ah, might be the concrete problem of https://github.com/input-output-hk/cardano-wallet/pull/2745#discussion_r661412735 ! I had tried killing cardano-wallet (which worked fine), but not deleting a single wallet!

Anviking:
> There are also entries saying wallet looses connection with the node:

Yeah, even if re-opening connections help, it's interesting why they are closed to begin with... 

There should be a more detailed error message from ouroboros-network (but it might be "lost" / not shown)

_can try to reproduce / look over logging if needed after sorting out the need-to-catch-async problem_

Anviking:
Closing in favour of #2750, at least for now.

PR #2746, opened 2021-06-30
Update windows badge to point to proper workflow

piotr-iohk:
# Issue Number

# Overview

- [ ] Update windows badge to point to proper workflow

# Comments

It seems it was pointing to some obsolete workflow.


rvl (approved):
Thanks

PR #2750, opened 2021-07-01
Refactor chain following

Anviking:
### Issue Number

Based on #2745 
ADP-871

### Motivation

Simplify and clean up the chain following code, in the hope of making the wallet more robust against node disconnects. In particular, reduce the number of threads used to run the ChainSync protocol.

### Overview

* The `chainSync` function now takes a record of callbacks, `ChainFollower`, as an argument. These callbacks are used to request the current intersection, as well as react to roll forward and roll backwards messages.
* Remove the old `Cursor` type.
* Reduce the number of threads used for implementing the above.

### Progress

- [x] Integration tests run without errors
- [ ] Clean up / review more closely
    - [x] Corner cases
    - [x] Logging 
        - [x] Aggregation of sync progress works again
        - [x] Ensure that all `ChainFollowLog` messages are traced or removed if redundant.
        - [x] `withFollowStatsMonitoring` runs in a separate thread so that it can compute statistics in regular time intervals.
    - [ ] Node disconnect errors
        - [x] … are thrown as exceptions in `connectTo` and are handled by `recoveringNodeConnection`
        - [ ] Check again at the end that the PR solves the bug
    - [x] Can we property-test this? (Probably no benefit.)
    - [x] Check for performance regressions (running the chain-sync and db operations on the same thread might make it slightly slower)
        - Actually seemed 10% faster running integration tests locally with -j 8.
            - ef3fdc31e4ae5bcd784552d95cec40a74a22662a -> 274s (n=2)
            - master -> 303s (n=1) <- but I only ran once
            - Speedup _could be_ related to chain-following now being able to rollback without necessarily re-negotiating the intersection.


lib/core/src/Cardano/Wallet/DB.hs
```diff
@@ -329,6 +331,8 @@ newtype ErrNoSuchWallet
     = ErrNoSuchWallet WalletId -- Wallet is gone or doesn't exist yet
     deriving (Eq, Show)
 
+instance Exception ErrNoSuchWallet
```
rvl: The possibility of both checked and unchecked exceptions - this is something we want to avoid.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -331,21 +326,30 @@ withNetworkLayerBase tr np conn (versionData, _) tol action = do
...
+            --
+            -- TODO: Recover on connection lost exceptions!
+            connectClient tr handlers client versionData conn
+
+        , currentNodeTip =
             fromTip getGenesisBlockHash <$> atomically readNodeTip
         , currentNodeEra =
             -- NOTE: Is not guaranteed to be consistent with @currentNodeTip@
             readCurrentNodeEra
         , watchNodeTip = do
```
rvl, line 366: With the reformulation of chainSync, this function will probably be redundant.

rvl (commented):
It will be good to simplify chain following.

lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
Anviking: I don't see how `watchNodeTip` would be related to this PR.

We use it for
1. Caching reward balances in DB
2. Tx resubmission

It would be good to unify the DB and in-memory caching of rewards, but that's a separate concern. (I think this might contribute to values being slower to be updated, and perhaps to some integration test flakiness).

Anviking: I guess you mean we could implement `watchNodeTip` using `chainSync`?

But I think 
1) having a single tip-sync protocol and multiple `STM` observers (allow callbacks to be skipped under load)
2) not having to deserialise block bodies

are good qualities with the current approach, plus I don't see how `ChainFollower` could allow fast-forwarding the intersection to the tip.

Anviking:
Closing for now.

rvl:
Let's keep the branch and re-open this PR later, because this refactor is something that we definitely need.

HeinrichApfelmus:
bors try

HeinrichApfelmus:
The integration test failures seem to be unrelated to the new chain following code. On my local machine, these failures do pass.

Instead, the failures seem to be related to:

* File deletion on windows (when cleaning up old wallets)
* Problems with DB concurrency (SQLITE_BUSY)

It appears that these problems will have to be addressed in the database layer.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -876,31 +876,121 @@ restoreWallet
...
+        }
   where
     db = ctx ^. dbLayer @IO @s @k
-    nw = ctx ^. networkLayer
-    tr = contramap MsgFollow (ctx ^. logger @WalletWorkerLog)
+    nw = ctx ^. networkLayer @IO
+    tr = ctx ^. logger @WalletWorkerLog
+    (block0, _, _) = ctx ^. genesisData
+
+    -- See Note [CheckedExceptionsAndCallbacks]
```
Anviking, line 915: 👍

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2769,28 +2858,30 @@ guardQuit WalletDelegation{active,next} rewards = do
 -- | Log messages for actions running within a wallet worker context.
 data WalletWorkerLog
     = MsgWallet WalletLog
-    | MsgFollow (FollowLog WalletFollowLog)
+    | MsgWalletFollow WalletFollowLog
+    | MsgChainFollow ChainFollowLog
```
Anviking, line 3086: 👌 fair.

lib/core/src/Cardano/Wallet/Primitive/SyncProgress.hs
```diff
@@ -99,40 +99,29 @@ instance FromText SyncTolerance where
...
---
--- Initially, `X` gives a relatively poor estimation of the network height, as
--- it assumes that every next slot will be a block. But, as we ingest blocks,
--- `h` becomes bigger and `X` becomes smaller making the progress estimation
--- better and better. At some point, `X` is null, and we have `p = h / h`
+-- In the Cardano consensus protocol, only a fraction of slots contains blocks.
+-- Hence, the progress percentage will often be < 100%,
+-- as the slot corresponding to the current wall-clock time
+-- may not be filled with a block.
+-- The sync tolerance should be large enough to accomodate this issue.
```
Anviking, line 111: 👍

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -306,51 +286,65 @@ chainSyncWithBlocks tr fromTip queue responseBuffer =
...
+    -- Explictly negotiate the genesis point
+    clientStNegotiateGenesis
+        :: m (P.ClientPipelinedStIdle 'Z block (Point block) (Tip block) m Void)
+    clientStNegotiateGenesis = do
+        let genesis = [Point Origin]
+        traceWith tr $ MsgChainFindIntersect genesis
+        pure $ P.SendMsgFindIntersect genesis $
+            clientStIntersect
+                { P.recvMsgIntersectNotFound = \_tip ->
+                    throwIO ErrChainSyncNoIntersectGenesis
```
Anviking, line 343: I guess this would be caught by `catchFromIO` in `restoreWallet` rather than causing `chainSync` to go into a retry-loop, which seems good.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1444,7 +1444,9 @@ newDBLayerWith cacheBehavior tr ti SqliteContext{runQuery} = do
                         [ StakeKeyCertSlot >. nearestPoint
                         ]
                     refreshCache wid
-                    pure (Right nearestPoint)
+                    selectLatestCheckpointCached wid >>= \case
```
Anviking, line 1447: What is this check for?

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -505,7 +520,7 @@ localStateQuery queue =
         :: LocalStateQueryCmd block m
         -> m (LSQ.ClientStAcquired block (Point block) (Query block) m Void)
     clientStAcquired (SomeLSQ cmd respond) = pure $ go cmd $ \res -> do
-        LSQ.SendMsgRelease (respond res >> clientStIdle)
+        LSQ.SendMsgRelease (respond res >> finalizeCmd >> clientStIdle)
```
Anviking: I wonder if not technically a lost connection between `respond res` and `finaliseCmd` can lead to the callback being called twice. 

But this should be exceedingly rare, and it shouldn't affect queries using our `send` helper:
```
send
    :: MonadSTM m
    => TQueue m (cmd m)
    -> ((a -> m ()) -> cmd m)
    -> m a
send queue cmd = do
    tvar <- newEmptyTMVarIO
    atomically $ writeTQueue queue (cmd (atomically . putTMVar tvar))
    atomically $ takeTMVar tvar
```
and I don't imagine it would affect anything else either.

If you concur, maybe we should add a note about it though.

Anviking (commented):
From an incomplete first pass now as well as some commits I saw in the past, lgtm!, left some comments, but will have another look.

Maybe we could regardless have a ~20 minute call about it tomorrow some time after the futurespective?

piotr-iohk:
The branch survives the test from ADP-871 which is great, but each time, after node or wallet server is restarted, all the wallets begin to sync from scratch, that is:
 - after restarting the node (with wallet server working in bg) 
 - or after restarting the wallet (with node working in bg)
 - also after restarting both... :)

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
HeinrichApfelmus: The reason is simply that I changed `rollbackTo` to return full `BlockHeader`s instead of just `SlotNo`s. That's why we can't just return `nearestPoint`, but have to fetch the corresponding block header from the database.

lib/core/src/Ouroboros/Network/Client/Wallet.hs
HeinrichApfelmus: Yes, exactly. This is an infinite loop that we can detect, so let's throw an error instead.

lib/core/src/Ouroboros/Network/Client/Wallet.hs
HeinrichApfelmus: Hm, I do not believe that there is a race condition here. 🤔 As far as I can tell, losing the connection to a node is a synchronous exception: The next read or write to the underlying socket will fail. As long as we do not read or write to the socket, there will be no exception related to it.

If this code path receives an asynchronous exception, then this will not be caught by `recoveringNodeConnection`, and it doesn't matter whether we put anything back into the queue or not.

But I agree that it's best for `respond` to avoid throwing a synchronous exception (e.g. from trying to read the node) and will add a note to that effect.

HeinrichApfelmus:
I rewrote the commit history (though perhaps a bit too aggressively). 🤓

piotr-iohk:
Would you mind rebasing over master :pray:

HeinrichApfelmus:
> Would you mind rebasing over master 🙏

Sure, no problem. I have fixed the issue with the `HasLogger` class.

lib/core/src/Ouroboros/Network/Client/Wallet.hs
```diff
@@ -306,51 +292,67 @@ chainSyncWithBlocks tr fromTip queue responseBuffer =
...
+    rollbackBuffer pt = filter (\b -> At (blockSlot b) <= pointSlot pt)
+
+    clientStNegotiateIntersection
+        :: m (P.ClientPipelinedStIdle 'Z block (Point block) (Tip block) m Void)
+    clientStNegotiateIntersection = do
+        points <- readLocalTip chainFollower
+        -- Cave: An empty list is interpreted as requesting the genesis point.
+        let points' = if null points
+                then [Point Origin]
+                else sortBy (flip compareSlot) points -- older points last
```
Anviking, line 305: Ah wow — does this mean on `master` we'd always rollback to the oldest common checkpoint? 🤔 

I see there's a `Asc CheckpointSlot` in the wallet DB `listCheckpoints`
https://github.com/input-output-hk/cardano-wallet/blob/c8cbdb8e40763f13bc58ecc8af9b39f34b0d0314/lib/core/src/Cardano/Wallet/DB/Sqlite.hs#L1416

Is there a case for dropping the `sortBy` and changing the direction of the sort the DB returns (both wallet db and pool db)? 

Pros:
- Perhaps negligible performance increase
- avoids the cognitive overhead of sorting it in different directions in two different places

Con:
-  easier for chainSync call-sites to mess up

lib/core/src/Cardano/Wallet.hs
```diff
@@ -917,23 +919,32 @@ restoreWallet ctx wid = db & \DBLayer{..} -> do
...
-pseudoPointSlot ChainPointAtGenesis = W.SlotNo 0
-pseudoPointSlot (ChainPoint slot _) = slot
+    -- See NOTE [PointSlotNo]
+    pseudoPointSlot :: ChainPoint -> SlotNo
+    pseudoPointSlot ChainPointAtGenesis = W.SlotNo 0
+    pseudoPointSlot (ChainPoint slot _) = slot
 
-toChainPoint :: W.BlockHeader -> ChainPoint
-toChainPoint (BlockHeader slot _ h _) = ChainPoint slot h
+toChainPoint :: W.Block -> W.BlockHeader -> ChainPoint
```
Anviking, line 982: An alternative to `W.Block` would be take a `W.Hash "Genesis"` directly, which is a type we use in a few places:

```
lib/core/src/Cardano/Wallet.hs:    | ErrCheckIntegrityDifferentGenesis (Hash "Genesis") (Hash "Genesis")
lib/core/src/Cardano/Wallet/Api/Types.hs:    { genesisBlockHash :: !(ApiT (Hash "Genesis"))
lib/core/src/Cardano/Wallet/Api/Types.hs:instance FromJSON (ApiT (Hash "Genesis")) where
lib/core/src/Cardano/Wallet/Api/Types.hs:instance ToJSON (ApiT (Hash "Genesis")) where
lib/core/src/Cardano/Wallet/Primitive/Types.hs:    { getGenesisBlockHash :: Hash "Genesis"
lib/core/src/Cardano/Wallet/Primitive/Types/Hash.hs:instance FromText (Hash "Genesis")         where fromText = hashFromText 32
lib/core/test-common/Cardano/Wallet/DummyTarget/Primitive/Types.hs:genesisHash :: Hash "Genesis"
lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs:            jsonRoundtripAndGolden $ Proxy @(ApiT (Hash "Genesis"))
lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs:            |] `shouldBe` (Left @String @(ApiT (Hash "Genesis")) msg)
lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs:instance Arbitrary (Hash "Genesis") where
lib/core/test/unit/Cardano/Wallet/Primitive/SlottingSpec.hs:instance Arbitrary (Hash "Genesis") where
lib/core/test/unit/Cardano/Wallet/Primitive/SlottingSpec.hs:    arbitrary = return $ Hash "Genesis Hash"
lib/core/test/unit/Cardano/Wallet/Primitive/Types/HashSpec.hs:        textRoundtrip $ Proxy @(Hash "Genesis")
lib/core/test/unit/Cardano/Wallet/Primitive/Types/HashSpec.hs:            fromText @(Hash "Genesis") "----" ===
lib/core/test/unit/Cardano/Wallet/Primitive/Types/HashSpec.hs:instance Arbitrary (Hash "Genesis") where
lib/shelley/src/Cardano/Wallet/Byron/Compatibility.hs:fromChainHash :: W.Hash "Genesis" -> ChainHash ByronBlock -> W.Hash "BlockHeader"
lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs:    => W.Hash "Genesis"
lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs:    :: W.Hash "Genesis"
lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs:    :: W.Hash "Genesis"
lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs:toTip :: W.Hash "Genesis" -> W.BlockHeader -> Tip (CardanoBlock sc)
lib/shelley/test/unit/Cardano/Wallet/Shelley/CompatibilitySpec.hs:instance Arbitrary (Hash "Genesis") where
```
and should prevent callers passing in just any block.

Not that important though, so feel free to keep it as is.

Anviking (commented):
LGTM! Thanks for picking this up!

I cannot officially approve since it is was my PR, so feel free to do so yourself.

lib/core/src/Ouroboros/Network/Client/Wallet.hs
Anviking: >  As far as I can tell, losing the connection to a node is a synchronous exception: The next read or write to the underlying socket will fail. As long as we do not read or write to the socket, there will be no exception related to it.

Ok, good point, thanks

lib/core/src/Cardano/Wallet.hs
HeinrichApfelmus: I would indeed prefer to keep it as it is — my main motivation for using the block as opposed to the hash was to not duplicate the line

```
genesisHash = genesisBlock ^. (#header . #headerHash)
```

More importantly, I expect the `toChainPoint` function to be thrown out of `Cardano.Wallet` module again, and incorporated into the DB layer.

lib/core/src/Ouroboros/Network/Client/Wallet.hs
HeinrichApfelmus: > Ah wow — does this mean on master we'd always rollback to the oldest common checkpoint? 🤔

Fortunately not, as that would have meant rolling back to genesis (this is the issue that Piotr encountered). There was a call to `reverse` somewhere in the old `follow` function on master. 😅

Actually, the specification of the ChainSync mini-protocol stipulates that the node should ignore the sort order and returns the youngest point on the chain — but `cardano-node` behaves differently. See https://github.com/input-output-hk/ouroboros-network/issues/3443

> Is there a case for dropping the sortBy and changing the direction of the sort the DB returns (both wallet db and pool db)?

In light of the ChainSync spec, I would make the case for keeping the `sortBy` in the `chainSync` function and instead dropping the sort direction from the DB layer.

HeinrichApfelmus (approved):

lib/core/src/Ouroboros/Network/Client/Wallet.hs
HeinrichApfelmus: > In light of the ChainSync spec, I would make the case for keeping the sortBy in the chainSync function and instead dropping the sort direction from the DB layer.

(But do nothing for now, as this would become obsolete in the DB layer redesign anyway.)

piotr-iohk (approved):
LGTM! Awesome improvement!

Playing around with the branch a little and It looks that it solves multiple stability issues a.k.a. "vanishing wallets", addressed in:
 - ADP-871
 - ADP-1013
 - ADP-798

I've also run this branch against e2e tests few times and no "wallet_not_responding" observed, which was rather common before, now passing e.g.:

https://github.com/input-output-hk/cardano-wallet/actions/runs/1415608203
https://github.com/input-output-hk/cardano-wallet/actions/runs/1412956182
https://github.com/input-output-hk/cardano-wallet/actions/runs/1415762485
https://github.com/input-output-hk/cardano-wallet/actions/runs/1412958475

:heart:

PR #2751, opened 2021-07-05
More integration tests for construct tx endpoint

piotr-iohk:
# Issue Number

ADP-909

# Overview

Integration tests covering:

- [x] More malformed examples
- [x] Only validate_interval
- [x] Tx with only metadata
- [x] Tx with only withdrawal from self and external
- [x] Only single output tx :warning: `pending` - outputs are empty 
- [x] Only multi-output tx :warning: `pending` - outputs are empty 
- [x] Multi asset transactions :warning: `pending` - outputs are empty 
- [x] Join and quit pool
  - join :warning: `pending` - deposits are empty
  - quit :warning: `pending` - can quit when you didn't join
  - errMsg404NoSuchPool :warning: `pending` - error not returned when should be
- [ ] ~Minting~
- [x] validate_interval on different transactions (payments, delegations, mint)
  - invalid_before / invalid_hereafter :warning: `pending` -  can be < 0 but shouldn't
- [x] Transaction with all (payments, delegations, mint, metadata, withdrawal, validate_interval)

# Comments / Issues

 - Should construct ep return `200` instead of `202` (coin selection returns `200`)
 - `amount` field would be probably handy to have in the response  
 - Output field on response is empty (`"output": []`) on payment transactions, which seems like a bug (to be addressed in ADP-985)... for now tests marked as `pendingWith`
 - Deposit is empty on construct tx response on joining pool and shouldn't
 - You can quit pool before even joining
 - errMsg404NoSuchPool is not returned when trying to join absent pool

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
@@ -25,6 +25,7 @@ import Cardano.Wallet.Api.Types
     ( ApiCoinSelectionInput (..)
     , ApiConstructTransaction
     , ApiFee (..)
+    , ApiStakePool
     , ApiT (..)
     , ApiWallet
     , DecodeAddress
@@ -43,7 +44,7 @@ import Control.Monad.IO.Unlift
 import Control.Monad.Trans.Resource
     ( runResourceT )
 import Data.Generics.Internal.VL.Lens
-    ( (^.) )
+    ( view, (^.) )
 import Data.Maybe
     ( isJust )
 import Data.Proxy
@@ -64,6 +65,8 @@ import Test.Integration.Framework.DSL
     ( Context
     , Headers (..)
     , Payload (..)
+    , arbitraryStake
+    , arbitraryStake
     , emptyWallet
     , expectErrorMessage
     , expectField
@@ -86,11 +89,14 @@ import Test.Integration.Framework.TestData
     ( errMsg403Fee
     , errMsg403InvalidConstructTx
     , errMsg403MinUTxOValue
+    , errMsg403NotDelegating
     , errMsg403NotEnoughMoney
+    , errMsg404NoSuchPool
     )
 
 import qualified Cardano.Wallet.Api.Link as Link
 import qualified Data.List.NonEmpty as NE
+import qualified Data.Text as T
 import qualified Network.HTTP.Types.Status as HTTP
 
 spec :: forall n.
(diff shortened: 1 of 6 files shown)


paweljakubas (approved):
lgtm

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -78,19 +106,105 @@ spec :: forall n.
...
             (Link.createUnsignedTransaction @'Shelley wa) Default emptyPayload
         verify rTx
             [ expectResponseCode HTTP.status403
+            , expectErrorMessage errMsg403InvalidConstructTx
+            ]
+
+    it "TRANS_NEW_CREATE_01b - Validity interval only is not allowed" $ \ctx -> runResourceT $ do
+        wa <- fixtureWallet ctx
+        let validityInterval = Json [json|{
+                "validity_interval": {
```
rvl, line 123: @paweljakubas We should also add a simple validation (invalid before <= invalid hereafter) of the validity interval to the API JSON parser.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -114,7 +231,512 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
...
+          \parsing ApiValidityBound object failed, \
+          \expected Object, but encountered String \
+          \- to be fixed in ADP-985"
+
+        wa <- fixtureWallet ctx
+
+        let payload = Json [json|{
+                "withdrawal": "self",
+                "validity_interval": {
+                    "invalid_before": "unspecified",
```
rvl, line 662: It should be that omission of the interval bound, or just setting it to null means "unspecified" -- no need to use a special string.

specifications/api/swagger.yaml
```diff
@@ -1363,7 +1363,7 @@ x-transactionTTL: &transactionTTL
 x-unspecified: &unspecified
   type: string
   enum:
-    - uspecified
+    - unspecified
```
rvl, line 1366: @paweljakubas we should remove this "unspecified" and use null/undefined/missing to denote unspecified.

rvl (approved):
LGTM - thanks.

specifications/api/swagger.yaml
piotr-iohk: Yes, we've discussed this with Paweł and for now just added some tests (also covering other "failing" scenarios) and make them "pending"), such that then can be enabled later once the issues are fixed. Findings also listed here -> https://jira.iohk.io/browse/ADP-909.

PR #2754, opened 2021-07-07
Implement signTransaction

paweljakubas:
### Overview

- [x] fix output when constructing tx and show it works in integration tests (adp-909 artefact)
- [x] extend Transaction layer by adding `mkSignedTransaction`
- [x] add impl of signing in Shelley.Transaction
- [x] handle deserializing of value of SerializedTx
- [x] add signTx as link and extend integration test 
- [x] handle extracting inputs from TxBody
- [x] handle extracting withdrawals from TxBody
- [x] do Shelley.Transaction refactoring 
- [x] align Wallet.signTransaction accordingly
- [x] align Api.Server.signTransaction accordingly

### Comments

Based on the branch of PR #2877. Merge that first.

### Issue Number

ADP-919


lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -202,45 +206,55 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
...
-        verify rTx
+
+            (Link.getTransactionFee @'Shelley wa) constrPayload
+        rConstrTx <- request @(ApiConstructTransaction n) ctx
+            (Link.createUnsignedTransaction @'Shelley wa) Default constrPayload
+        verify rConstrTx
             [ expectSuccess
             , expectResponseCode HTTP.status202
-            , expectField (#coinSelection . #inputs) (`shouldSatisfy` (not . null))
-            , expectField (#coinSelection . #outputs) (`shouldSatisfy` (not . null))
```
piotr-iohk, line 223: Why removing this assertion, especially that the test was failing before due to missing outputs? (pls see `pendingWith` message)

paweljakubas: just check it later in the same test, but deeper:
```
        let coinSelOutputs = getFromResponse (#coinSelection . #outputs) rConstrTx
        coinSelOutputs `shouldBe`
            [ApiCoinSelectionOutput destAddr (Quantity amt) (ApiT W.empty)]
```

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -3299,6 +3301,40 @@ instance IsServerError ErrSignPayment where
...
+                , "Retrying may work, but something really went wrong..."
+                ]
+        ErrSignTxInvalidSerializedTx hint ->
+            apiError err500 CreatedInvalidTransaction hint
+        ErrSignTxInvalidEra ->
+            apiError err500 CreatedInvalidTransaction $ mconcat
+                [ "Whoops, it seems like I just experienced a hard-fork in the "
+                , "middle of other tasks. This is a pretty rare situation but "
+                , "as a result, I must throw-away what I was doing. Please "
+                , "retry whatever you were doing in a short delay."
```
```suggestion
                , "as a result, I must throw away what I was doing. Please "
                , "retry your request."
```

Hi @paweljakubas: I'm just wondering, how long should the user wait here? Is it a case of a few seconds, or a few minutes?

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
```diff
@@ -392,11 +416,91 @@ instance Buildable a => Buildable (WithDirection a) where
...
+withinEra (AnyCardanoEra MaryEra) (AnyCardanoEra _) = True
+withinEra (AnyCardanoEra AllegraEra) (AnyCardanoEra AlonzoEra) = False
+withinEra (AnyCardanoEra AllegraEra) (AnyCardanoEra MaryEra) = False
+withinEra (AnyCardanoEra AllegraEra) (AnyCardanoEra _) = True
+withinEra (AnyCardanoEra ShelleyEra) (AnyCardanoEra AlonzoEra) = False
+withinEra (AnyCardanoEra ShelleyEra) (AnyCardanoEra MaryEra) = False
+withinEra (AnyCardanoEra ShelleyEra) (AnyCardanoEra AllegraEra) = False
+withinEra (AnyCardanoEra ShelleyEra) (AnyCardanoEra _) = True
+withinEra (AnyCardanoEra ByronEra) (AnyCardanoEra ByronEra) = True
+withinEra (AnyCardanoEra ByronEra) _ = False
```
jonathanknowles: Hi @paweljakubas. Just wondering, could this be expressed in terms of ordering for eras, with `<`?

Alternatively, this seems a bit shorter:

```hs
withinEra :: AnyCardanoEra -> AnyCardanoEra -> Bool
withinEra (AnyCardanoEra e1) (AnyCardanoEra e2) = case (e1, e2) of
    (AlonzoEra , _         ) -> True
    (MaryEra   , AlonzoEra ) -> False
    (MaryEra   , _         ) -> True
    (AllegraEra, AlonzoEra ) -> False
    (AllegraEra, MaryEra   ) -> False
    (AllegraEra, _         ) -> True
    (ShelleyEra, AlonzoEra ) -> False
    (ShelleyEra, MaryEra   ) -> False
    (ShelleyEra, AllegraEra) -> False
    (ShelleyEra, _         ) -> True
    (ByronEra  , ByronEra  ) -> True
    (ByronEra  , _         ) -> False
```

(just thinking out loud: maybe it's also nice to avoid `_` here, to be more explicit about which cases match, and catch compile errors in case this sum type gets changed)

paweljakubas:
bors try

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
```diff
@@ -465,18 +465,19 @@ cardanoTxFromBytes maxEra bs = asum $ map snd $ filter (withinEra maxEra . fst)
...
+    (MaryEra   , _         ) -> True
+    (AllegraEra, AlonzoEra ) -> False
+    (AllegraEra, MaryEra   ) -> False
+    (AllegraEra, _         ) -> True
+    (ShelleyEra, AlonzoEra ) -> False
+    (ShelleyEra, MaryEra   ) -> False
+    (ShelleyEra, AllegraEra) -> False
+    (ShelleyEra, _         ) -> True
+    (ByronEra  , ByronEra  ) -> True
+    (ByronEra  , _         ) -> False
```
jonathanknowles: Hey @paweljakubas. I wonder if this is possible to express with `>=`.

Is there an `Ord` instance for the type of value within `AnyCardanoEra`?

If so, it seems like the case expression could be replaced with something like `x >= y`.

But perhaps I've misunderstood something here. 😄

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
```diff
@@ -392,11 +416,92 @@ instance Buildable a => Buildable (WithDirection a) where
...
+    -> ByteString -- ^ Serialised transaction
+    -> Either DecoderError (InAnyCardanoEra Cardano.Tx)
+cardanoTxFromBytes maxEra bs = asum $ map snd $ filter (withinEra maxEra . fst)
+    [ (anyCardanoEra AlonzoEra, InAnyCardanoEra AlonzoEra <$> deserialiseFromCBOR (Cardano.AsTx Cardano.AsAlonzoEra) bs)
+    , (anyCardanoEra MaryEra, InAnyCardanoEra MaryEra <$> deserialiseFromCBOR (Cardano.AsTx Cardano.AsMaryEra) bs)
+    , (anyCardanoEra AllegraEra, InAnyCardanoEra AllegraEra <$> deserialiseFromCBOR (Cardano.AsTx Cardano.AsAllegraEra) bs)
+    , (anyCardanoEra ShelleyEra, InAnyCardanoEra ShelleyEra <$> deserialiseFromCBOR (Cardano.AsTx Cardano.AsShelleyEra) bs)
+    , (anyCardanoEra ByronEra, InAnyCardanoEra ByronEra <$> deserialiseFromCBOR (Cardano.AsTx Cardano.AsByronEra) bs)
+    ]
+  where
```
jonathanknowles: Suggestion: extract out the shared logic into a separate function:

```suggestion
cardanoTxFromBytes maxEra bs = asum $ map snd $ filter (withinEra maxEra . fst)
    [ deserialise AlonzoEra  Cardano.AsAlonzoEra
    , deserialise MaryEra    Cardano.AsMaryEra
    , deserialise AllegraEra Cardano.AsAllegraEra
    , deserialise ShelleyEra Cardano.AsShelleyEra
    , deserialise ByronEra   Cardano.AsByronEra
    ]
  where
    deserialise
        :: forall era. Cardano.IsCardanoEra era
        => CardanoEra era
        -> Cardano.AsType era
        -> (AnyCardanoEra, Either DecoderError (InAnyCardanoEra Cardano.Tx))
    deserialise era asEra =
        ( anyCardanoEra era
        , InAnyCardanoEra era <$> deserialiseFromCBOR (Cardano.AsTx asEra) bs
        )
```

rvl: Yep good idea, thanks.

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
rvl: There is no `Ord` on `AnyCardanoEra`, so that's what we're defining here.
The wildcard are because no era is higher that AlonzoEra, as an example.
Anyway I think it can be made shorter just by assigning a natural to each era.

HeinrichApfelmus: Assigning naturals would be morally equivalent to treating `AnyCardanoera` as an `Enum`, I suppose?

rvl (approved):

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Balance.hs
```diff
@@ -510,15 +515,6 @@ performSelection minCoinFor costFor bundleSizeAssessor criteria
...
                 }
   where
-    SelectionCriteria
-        { outputsToCover
-        , utxoAvailable
-        , selectionLimit
-        , extraCoinSource
-        , assetsToMint
-        , assetsToBurn
-        } = criteria
```
jonathanknowles: Hi @rvl.

I'd like to ask a favour: would it be possible to avoid changing these to use `RecordWildCards`, at least for the moment? 

There are two things here:
- it makes it harder to figure out where fields have come from, as some field names are shared by multiple records. If we use record wildcards, then go-to-definition no longer works.
- I'm currently working on this module, so it would help me greatly if we could keep any changes to a minimum!

Thanks!

lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
```diff
@@ -192,20 +305,20 @@ data SelectionParams = SelectionParams
...
 
 -- | Prepares the given user-specified outputs, ensuring that they are valid.
 --
-prepareOutputs
+prepareOutputsForMinUTxO
     :: SelectionConstraints
     -> NonEmpty TxOut
     -> Either ErrPrepareOutputs (NonEmpty TxOut)
-prepareOutputs constraints outputsUnprepared
+prepareOutputsForMinUTxO constraints outputsUnprepared
```
jonathanknowles: Hi @rvl. I think this name change is actually unnecessary, as this function is not just about the minimum UTxO: it also:
- checks the serialized sizes of outputs
- checks that outputs don't have token quantities in excess of the maximum (`maxBound :: Word64`).

I reckon it would be better to leave the name alone here.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -202,45 +206,54 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
...
+              "passphrase": #{fixturePassphrase}
+          }|]
+        rSignTx <- request @ApiSignedTransaction ctx
+            (Link.signTransaction @'Shelley wa) Default signPayload
+        verify rSignTx
+            [ expectSuccess
+            , expectResponseCode HTTP.status202
+            ]
+
+        -- and send it in two steps
```
piotr-iohk, line 249: So it seems this part is missing. Can this be added at this stage already?

KtorZ:
Closing in favor of: https://github.com/input-output-hk/cardano-wallet/pull/2943

PR #2756, opened 2021-07-09
Add coverage checks to property tests for `TokenMap.intersection` 

jonathanknowles:
# Issue Number

ADP-997

# Pre-merge checks

- [x] Perform soak test of test suite to ensure no flakiness has been introduced.
    (✅ 1000 test runs with `--match "/Token map properties/Arithmetic/prop_intersection"` passed with no failures.)

# Overview

PR #2725 added a `TokenMap.intersection` function and property tests.

This PR:
- [x] adds coverage checks to the `prop_intersection_*` series of properties.
- [x] where necessary, adjusts property inputs in order to attain the required level of coverage.

### DIFF
--- a/lib/test-utils/src/Test/QuickCheck/Extra.hs
+++ b/lib/test-utils/src/Test/QuickCheck/Extra.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 
 -- |
@@ -10,12 +11,17 @@
 module Test.QuickCheck.Extra
     ( reasonablySized
     , shrinkInterleaved
+
+    -- * Combinators
+    , ManyFolded (..)
     ) where
 
 import Prelude
 
 import Test.QuickCheck
-    ( Gen, scale )
+    ( Arbitrary (..), Gen, scale )
+
+import qualified Data.Foldable as F
 
 -- | Resize a generator to grow with the size parameter, but remains reasonably
 -- sized. That is handy when testing on data-structures that can be arbitrarily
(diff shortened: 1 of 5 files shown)


sevanspowell (approved):
Looks good to me!

lib/test-utils/src/Test/QuickCheck/Extra.hs
```diff
@@ -51,3 +57,22 @@ shrinkInterleaved (a, shrinkA) (b, shrinkB) = interleave
...
+--------------------------------------------------------------------------------
+
+-- | A combinator for generating larger arbitrary values, produced by
+--   generating many smaller arbitrary values and folding them together.
+--
+-- Applying this combinator to a type 'A' with an instance of 'Monoid' will
+-- generate larger values of 'A', based on simple monoidal concatenation of
+-- multiple smaller values.
+--
+newtype ManyFolded a = ManyFolded
```
sevanspowell: This is required because the available Arbitrary instance for TokenMap is something like this:

```
instance Arbitrary TokenMap where
    arbitrary = genTokenMapSmallRange
```

right?

I think I prefer Hedgehog's use of functions to Quickcheck's typeclass instances, but that's not the question here.

rvl: I found this confusing @jonathanknowles.

Adding in `Blind` didn't help (couldn't `Show` be defined with `show . mconcat`?

Why can't we just use the QuickCheck `size` parameter? See [Coding Standards - QuickCheck](https://github.com/input-output-hk/cardano-wallet/wiki/Coding-Standards#use-the-size-parameter).

rvl (commented):
Thanks @jonathanknowles .
Coverage checks - good.
ManyFolded - unsure.

lib/test-utils/src/Test/QuickCheck/Extra.hs
jonathanknowles: Thanks @rvl. I agree with you that using the `size` parameter would be a better solution here.

I've added a commit which does this here: 86f48565e9822e3cb1d76721236e621c152e72e5

PR #2760, opened 2021-07-12
Byron construct tx endpoint - initial integration tests

piotr-iohk:
# Issue Number

ADP-909

# Overview

- [x] Adding integration tests for Byron construct tx ep.
   **NOTE:** only `payments` request parameter is tested, as according to https://input-output-rnd.slack.com/archives/GBT05825V/p1626077556008700?thread_ts=1625814178.001100&cid=GBT05825V it should be the only one supported in byront construct tx ep.

# Comments

 Byron construct tx endpoint returns:
 - For Byron wallets: `500 - Something went wrong`
 - For Icarus wallets: `403`:
```
"code":"invalid_wallet_type",
"message":"It is regrettable but you've just attempted an operation that is invalid for this type of wallet. Only new 'Shelley' wallets can do something with rewards and this one isn't."}
```
Therefore tests have been mostly marked as `pendingWith`.

### DIFF
--- a/lib/core-integration/cardano-wallet-core-integration.cabal
+++ b/lib/core-integration/cardano-wallet-core-integration.cabal
@@ -89,6 +89,7 @@ library
       Test.Integration.Scenario.API.Byron.Addresses
       Test.Integration.Scenario.API.Byron.CoinSelections
       Test.Integration.Scenario.API.Byron.Transactions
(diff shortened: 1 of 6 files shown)


paweljakubas (approved):
yes, all should pass after transaction rewrite

PR #2761, opened 2021-07-14
Add specification for ADP-957.

jonathanknowles:
# Issue Number

ADP-957

# Overview

This PR adds a high-level specification for ADP-957.


docs/requirements/ADP-957.md
```diff
@@ -0,0 +1,156 @@
...
+    cover. The collateral inputs must not themselves be locked by a script.
+
+*   The protocol parameter `maxCollateralInputs` is used to limit,
+    additionally, the total number of collateral inputs, and thus the total
+    number of additional signatures that must be checked during validation.
+
+# Proposed requirements
+
+1.  The wallet, when requested to create a transaction requiring collateral,
+    should reserve one or more pure-ada UTxO entries to serve as collateral
```
piotr-iohk: why more than one? I mean is it possible to have more than 1 collateral utxo required for transaction or 1 should be enough?

docs/requirements/ADP-957.md
```diff
@@ -0,0 +1,156 @@
...
+        parameter, then return an ordered list with two selections:
+
+        1. Selection 1: a transaction that creates appropriate collateral.
+        2. Selection 2: a transaction corresponding to the user request.
+
+    In each of the above cases, the user can inspect the list of selections
+    returned, and decide whether or not they want to go ahead (by signing the
+    selections, and/or submitting them).
+
+4.  Provide an explicit endpoint that creates a collateral rearrangement
```
piotr-iohk: I think that is relatively simple(st) solution which also provides pretty good user experience. It is simple as in: _when your tx fails you are advised to trigger this collateral ep first_ (which is far better than manually creating tx by themselves).

docs/requirements/ADP-957.md
```diff
@@ -0,0 +1,156 @@
...
+3.  The wallet's create transaction operation should accept a parameter
+    `collateralPercentMaximum`, with a default value that can be overridden
+    by the caller. The wallet, when requested to create a transaction requiring
+    collateral, should fail with an error if it **is** possible to meet the
+    minimum amount, but **not** possible to find a set of pure-ada entries
+    whose total balance is less than or equal to
+    `collateralPercentMaximum × transactionFee`.
+
+4.  The wallet should ensure that any UTxO entries used for collateral are
+    **locked** until the transactions that include them are validated.
```
piotr-iohk: Is locking needed as a requirement here? I mean, I guess, locking is something that we'd like to eventually have, but it is not necessarily part of req for collateral (Perhaps I'm wrong).

docs/requirements/ADP-957.md
```diff
@@ -0,0 +1,156 @@
...
+
+5.  The wallet should report the balance of posted collateral entries as part
+    of the total balance, but not as part of the available balance.
```
Anviking: If the tx is valid, will the collateral be spent like normal inputs? If so, is requirement not just how the wallet currently works?

And additionally if so, it seems misleading: _Inputs_ of pending txs are excluded from both the available and total balance. It's the corresponding _change outputs_ that gets added to the total balance only.

docs/requirements/ADP-957.md
jonathanknowles: @piotr-iohk wrote:
> why more than one? I mean is it possible to have more than 1 collateral utxo required for transaction or 1 should be enough?

A transaction body allows `[0 .. n]` collateral inputs. See https://input-output-hk.github.io/cardano-node/cardano-api/Cardano-Api-TxBody.html#t:TxInsCollateral

In practice, the upper bound is limited by the `maxCollInputs` protocol parameter. See
https://github.com/input-output-hk/cardano-ledger-specs/blob/bdc953e0b5615c2f86593630235174fd7c0f8c38/alonzo/formal-spec/protocol-parameters.tex#L105

jonathanknowles: In some cases, one collateral input might be enough.

But for transactions with higher fees, we might need to select more than one collateral input.

docs/requirements/ADP-957.md
jonathanknowles: @piotr-iohk wrote:
> I think that is relatively simple(st) solution which also provides pretty good user experience. It is simple as in: when your tx fails you are advised to trigger this collateral ep first (which is far better than manually creating tx by themselves).

I completely agree with you: it would be nice to provide the user with a convenient way to unblock themselves. Either via the explicit endpoint proposal (4), or the just-in-time rebalancing transaction proposal (3). 

Though I believe the priority at this stage is to deliver an MVP.

If we adopt proposal 1 (do nothing in addition to satisfying the basic requirements), then we can release something sooner that will at least allow advanced users to start working with collateral.

jonathanknowles:
A point that is relevant to these requirements:

> It is permitted to use the same inputs as both collateral and a regular inputs, as exactly one set of inputs is ever collected : collateral ones in the case of script failure, and regular inputs in the case when all scripts validate.

Source: https://hydra.iohk.io/build/6752480/download/1/alonzo-changes.pdf (Section 3.2)

docs/requirements/ADP-957.md
jonathanknowles: > If the tx is valid, will the collateral be spent like normal inputs? If so, is requirement not just how the wallet currently works?

My understanding is that:

If a transaction in a block is marked as having **_passed validation_**, then we must apply the following state transitions:

```
UTxO       := UTxO ⋪ inputs tx
pendingTxs := pendingTxs − tx
```

If a transaction in a block is marked as having **_failed validation_**, then we must apply the following state transitions:

```
UTxO       := UTxO ⋪ collateralInputs tx
pendingTxs := pendingTxs − tx
```

So in other words, when we process a transaction from a block, we either remove the collateral inputs from the UTxO, or we remove the ordinary inputs from the UTxO, but not both.

docs/requirements/ADP-957.md
jonathanknowles: I agree this requirement could be worded a bit better. I'll see if I can reword it.

My understanding: transactions that include collateral input will be marked as part of the pending set, as normal.

However, we'll have to be a bit careful here, as:
* for any given transaction `tx`:
  `inputs tx ∩ collateralInputs tx` can be non-empty.
* for any given pair of pending transactions `tx1` and `tx2`:
 `collateralInputs tx1 ∩ collateralInputs tx2` must be empty.

jonathanknowles: @piotr-iohk I've now reworded the requirements to be:

  6.  The wallet application must ensure that UTxO entries used for collateral inputs in pending transactions are not available for reuse as collateral inputs in future transactions.  
      
  7.  The wallet application should consider a UTxO entry used for collateral in a pending transaction to be part of the UTxO set that is available for ordinary input selection in future transactions, unless that UTxO entry is also used as an ordinary input in a pending transaction.
  
Hopefully this is clearer, but let me know if you think anything is unclear.

docs/requirements/ADP-957.md
jonathanknowles: @Anviking the treatment of transactions in blocks is now covered in the specification. Hopefully this should be clear, but let me know if you think anything needs to be clarified or improved further!

docs/requirements/ADP-957.md
```diff
@@ -0,0 +1,201 @@
...
+    percentage of the total transaction fee its collateral must (at a minimum)
+    cover. The collateral inputs must not themselves be locked by a script.
+
+*   The protocol parameter `maxCollateralInputs` is used to limit,
+    additionally, the total number of collateral inputs, and thus the total
+    number of additional signatures that must be checked during validation.
+
+# Requirements
+
+1.  The wallet application, when requested to create a transaction requiring
```
rvl: It's not fully defined which transactions need collateral.
Transactions with plutus script-locked outputs need collateral.
The collateral needed is a percentage of the total of plutus script-locked outputs.

docs/requirements/ADP-957.md
```diff
@@ -0,0 +1,201 @@
...
+
+*   **Collateral inputs**, in a transaction, are used to cover transaction fees
+    in the case that a non-native script fails. The term **collateral** refers
+    to the total ada contained in the UTxOs referenced by collateral inputs.
+    Collateral inputs can **only** contain ada: they may **not** contain
+    non-ada assets.
+
+*   The protocol parameter `collateralPercent` is used to specify the
+    percentage of the total transaction fee its collateral must (at a minimum)
+    cover. The collateral inputs must not themselves be locked by a script.
```
rvl: > The collateral inputs must not themselves be locked by a script.

This is a requirement, and one that we are interested in too.
Coin selection must only pick VK inputs for collateral.

docs/requirements/ADP-957.md
```diff
@@ -0,0 +1,201 @@
...
+    protocol parameter and the transaction fee:
+
+        collateralAmount ≥ txFee × collateralPercent
+
+    If it is not possible to satisfy the minimum bound, the wallet application
+    must return an error to the caller.
+
+4.  The wallet application, when requested to create a transaction requiring
+    collateral, should select an amount of ada that is less than or equal to a
+    maximum bound that is determined by a user-specified `maxCollateralFactor`
```
rvl: This isn't a requirement - it's a solution for a requirement.
The requirement is that we shouldn't put too much of the user's ada "at risk" of being lost by an invalid transaction.

docs/requirements/ADP-957.md
```diff
@@ -0,0 +1,201 @@
...
+
+        maxCollateralFactor ≥ 1
+
+    If it is not possible to satisfy the maximum bound, the wallet application
+    must return an error to the caller.
+
+5.  The wallet application may provide a default value for the
+    `maxCollateralFactor` parameter.
+
+6.  The wallet application, when requested to create a transaction requiring
```
rvl: From here onwards, the "requirements" are very much a specification of how the wallet should work with collateral.
Could we just put them under a "Design specification" section?

docs/requirements/ADP-957.md
```diff
@@ -0,0 +1,201 @@
...
+
+    If it is not possible to satisfy the minimum bound, the wallet application
+    must return an error to the caller.
+
+4.  The wallet application, when requested to create a transaction requiring
+    collateral, should select an amount of ada that is less than or equal to a
+    maximum bound that is determined by a user-specified `maxCollateralFactor`
+    parameter:
+
+        collateralAmount ≤ txFee × collateralPercent × maxCollateralFactor
```
rvl: This seems like a reasonable scheme to use.

It's not strictly necessary to implement any solution however.
The user will be shown how much of their ada is at stake, before they are asked to sign and submit the transaction.
Could we shift the user-defined max collateral solution into a new section?

rvl (commented):
Nice, thanks @jonathanknowles .

Some of these requirements are just restating what's in the alonzo spec.
They could be summarised as a single requirement, something like, "wallet should select collateral inputs where needed, as per the alonzo spec." However the details of _how_ we shall make valid alonzo era transactions are good to specify.

docs/requirements/ADP-957.md
jonathanknowles: > The requirement is that we shouldn't put too much of the user's ada "at risk" of being lost by an invalid transaction.

Sure. Though my motivation for writing this document is to clarify exactly what terms like "too much" mean.

If we only say that the wallet "shouldn't put too much ada at risk", then how much is too much? It's not clear how we would measure whether the requirement has been satisfied.

I've tried to apply the [SMART principles](https://erivanramos.medium.com/there-is-a-smart-way-to-write-software-requirements-48add56d3972) where it makes sense. In particular, to make each requirement specific and measurable: i.e., something that we can write a test for.

docs/requirements/ADP-957.md
```diff
@@ -0,0 +1,201 @@
...
+    of ada that satisfies the minimum and maximum bounds, but is permitted to
+    select a higher, non-optimal amount (still within bounds) in cases where
+    searching for the optimal solution would entail an unreasonably high
+    performance cost.
+
+7.  The wallet application must ensure that UTxO entries used for collateral
+    inputs in pending transactions are not available for reuse as collateral
+    inputs in future transactions.
+
+8.  The wallet application should consider a UTxO entry used for collateral in
```
sevanspowell: I know we were talking about this yesterday, but I didn't quite catch the conclusion. Surely if a UTxO is put up for collateral, it shouldn't be made available to selection? Unless we enforce some other property:

Given,
1. some transaction with a collateral input
2. some transaction whose input is the collateral of Tx#1

Tx#2 remains pending until Tx#1 is approved?

Or does the node layer handle this? Like "first-in wins".

sevanspowell: Thanks for this document btw, I think it's great :+1:

docs/requirements/ADP-957.md
jonathanknowles: @piotr-iohk In light of our recent discussions, I've once again adjusted the wording, to read:

5.  The wallet application must ensure that **_UTxO entries used as collateral
    inputs_** in pending transactions are **not** available for reuse either as
    collateral inputs or as ordinary inputs in future transactions.

6.  The wallet application must ensure that **_UTxO entries used as ordinary
    inputs_** in pending transactions are **not** available for reuse either as
    collateral inputs or as ordinary inputs in future transactions.
    

7.  The wallet application, when requested to create a transaction requiring
    collateral, is permitted to select a set of collateral inputs that
    intersects with the set of ordinary inputs for that transaction:

        inputs t ≠ ∅ ∧ collateralInputs t ≠ ∅

        =/=> (does not imply)

        inputs t     ∩ collateralInputs t = ∅

docs/requirements/ADP-957.md
jonathanknowles: > Coin selection must only pick VK inputs for collateral.

@rvl 
Thanks, I've updated the specification to make this clear.

specifications/wallet/ADP-957-collateral.md
```diff
@@ -0,0 +1,201 @@
...
+
+    b.  If the wallet detects that the UTxO set has **insufficient** collateral,
+        then return an ordered list with two selections:
+
+        1. Selection 1: a transaction that creates appropriate collateral.
+        2. Selection 2: a transaction corresponding to the user request.
+
+    c.  If the wallet detects that the UTxO set has **sufficient** collateral,
+        but that it's not possible to satisfy the `collateralPercentMaximum`
+        parameter, then return an ordered list with two selections:
```
jonathanknowles, line 161: This section should be removed, as we no longer have a `collateralPercentMaximum` parameter.

specifications/wallet/ADP-957-collateral.md
```diff
@@ -0,0 +1,201 @@
...
+
+*   A **rebalancing transaction**: a transaction whose output addresses are all
+    owned by the wallet, whose sole purpose is to rebalance the UTxO
+    distribution in some way.
+
+*   A **collateral rebalancing transaction**: a transaction that rebalances the
+    UTxO distribution so that it has outputs suitable for inclusion as
+    collateral inputs in a later transaction. Such a transaction would ensure
+    the presence of one or more pure-ada UTxO entries, and could be tailored to
+    producing collateral entries of a certain size.
```
sevanspowell, line 201:
```suggestion
    producing collateral entries of a certain size.
    
*   A **VK input**: At a high-level, a VK input is an input not locked
    by a script. From a practical point of view, consider the types of
    addresses:

| Address type       | Payment Credential | Stake Credential |
|--------------------|--------------------|------------------|
| Base address       | keyhash            | keyhash          |
|                    | scripthash         | keyhash          |
|                    | keyhash            | scripthash       |
|                    | scripthash         | scripthash       |
| Pointer address    | keyhash            | ptr              |
|                    | scripthash         | ptr              |
| Enterprise address | keyhash            | -                |
|                    | scripthash         | -                |
| Bootstrap address  | keyhash            | -                |
| Stake address      | -                  | keyhash          |
|                    | -                  | scripthash       |
| Future formats     | ?                  | ?                |

    From this point of view, A VK input is any UTxO whose output
    address has a payment credential type of "keyhash". This means
    that the UTxO is not locked by a script, nor has an output address
    that is a stake address. For an up-to-date table of address types,
    please see cardano-ledger-specs.
```

docs/requirements/ADP-957.md
jonathanknowles: This section has now been deleted. (Which simplifies things a great deal.)

PR #2765, opened 2021-07-19
Fix weeder 2.x config and remove weeds

### LINKED ISSUE #2695: Weeder: Command exited with code -9
# Context
 
# Test Case

Weeder check

# Failure / Counter-example

```
--- Weeder
weeder 
error: Command exited with code -9!
Continuing...
./buildkite-script-15be0fbf-9d48-4997-b874-643c35bebbf3: line 5: 30308 Killed                  ./sr/bin/rebuild --build-dir=$BUILD_DIR --cache-dir=$CACHE_DIR
```

# Resolution 

---

# QA

rvl:
### Issue Number

This could fix #2695, I really hope it does.

### Overview

I was looking into #2695, running `weeder` locally to see what it could be doing.

But it turns out that Weeder has been doing nothing at all since upgrading `weeder` to 2.1.3, which happened when we upgraded to ghc-8.10.4.

It needed to have a ghc option added to generate `.hie` files, which I didn't set before. If Weeder can't find `.hie` files, it just reports zero weeds (ocharles/weeder#19).

So I have:
1. Fixed the Weeder config for Stack and Cabal.
2. Removed weeds which would have been introduced while weeder wasn't working.
3. Removed weeds found by Weeder 2.x, which were never found by Weeder 1.0.9.

### Comments

```
 49 files changed, 109 insertions(+), 1681 deletions(-)
```


lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
```diff
@@ -1213,9 +1213,9 @@ instance Malformed (BodyParam ApiSignTransactionPostData) where
...
             , ("\"hello\"", "Error in $: parsing Cardano.Wallet.Api.Types.ApiSignTransactionPostData(ApiSignTransactionPostData) failed, expected Object, but encountered String")
             , ("{\"transaction\": \"\", \"random\"}", msgJsonInvalid)
-            , ("{\"transaction\": \"lah\", \"passphase\": \"Secure Passphrase\"}", "Error in $.transaction: Parse error. Expecting Base64-encoded format.")
-            , ("{\"transaction\": 1020344, \"passphase\": \"Secure Passphrase\"}", "Error in $.transaction: parsing 'Base64 ByteString failed, expected String, but encountered Number")
-            , ("{\"transaction\": { \"body\": 1020344 }, \"passphase\": \"Secure Passphrase\"}", "Error in $.transaction: parsing 'Base64 ByteString failed, expected String, but encountered Object")
+            , ("{\"transaction\": \"lah\", \"passphrase\": \"Secure Passphrase\"}", "Error in $.transaction: Parse error. Expecting Base64-encoded format.")
+            , ("{\"transaction\": 1020344, \"passphrase\": \"Secure Passphrase\"}", "Error in $.transaction: parsing 'Base64 ByteString failed, expected String, but encountered Number")
```
Anviking, line 1217: This mis-spelling wasn't intentional?

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -573,15 +540,7 @@ fromTip genesisHash tip = case getPoint (getTipPoint tip) of
...
-        -- TODO
-        -- We only use the parentHeaderHash in the
-        -- 'Cardano.Wallet.Network.BlockHeaders' chain follower only required for
-        -- Jörmungandr, this is therefore useless to have in 'normal' BlockHeader
-        --
-        -- Yet, since we also serialize these to the database, this requires
-        -- some non-trivial changes. Not fixing this right now is also a
-        -- possibility.
-        , parentHeaderHash = W.Hash "parentHeaderHash - unused in Shelley"
+        , parentHeaderHash = W.Hash ""
```
Anviking: Removing `parentHeaderHash` would be nice, but is this `"parentHeaderHash - unused in Shelley"` to `""` change safe? It couldn't affect compatibility with existing DBs somehow?

Anviking: No, I think we only do logic with the BlockHeader after converting it to a `Point CardanoBlock`, so this should be fine.

Anviking (approved):
👍 My, that's a lot of changes (skimmed them)

lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
rvl: It was a mistake - found when changing to strictRecordTypeOptions.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
rvl: I tried removing `parentHeaderHash` but then found that it is sort of used in `Cardano.Wallet.restoreBlocks` !

Anviking: Ah!

rvl: Perhaps you could try removing parentHeaderHash? To keep DB compatibility, just leave the field there, but don't use it.

Anviking: I suspect it might make sense to do with/after #2750 if we can replace `BlockHeader` with `Point CardanoBlock` and be able to persist the Origin point in a nicer way. Currently I _think_ rollbacks to `Origin` might break:

```
G 0 1 <-- Wallet has checkpoints at genesis, slot 0 and slot 1
Node asks wallet to rollback to G
Wallet believes genesis has slotNo 0:
G 0
Node asks wallet to re-apply block in slot 0:
G 0 0 <- Wrong!
```

Although...

1. This invariant is nice to have, particularly as we lack great chain-following testing
https://github.com/input-output-hk/cardano-wallet/blob/8d1b692fec0c77e488a724dba97c9fa51ab9f4e5/lib/core/src/Cardano/Wallet.hs#L888-L893
2. I am a bit concerned about how to handle backward compatibility regarding how `Origin` is encoded https://github.com/input-output-hk/cardano-wallet/blob/914668d4e40db6a0b999869c4c4f5222fa6884d0/lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs#L392-L398

jonathanknowles (commented):
Hi @rvl. 

It's good to see all these weeds removed!

However, I think this PR also does a couple of other things:

1. Makes some behavioural changes to tests
2. Renames some functions

These behavioural changes are bundled together into a single commit with the weeding changes.

**Request:** Would it be possible to move the behavioural changes (and function renamings) into a separate PR?

This would make it easier to give these non-weeding changes the scrutiny they deserve, I think.

Example of a function renaming:
```diff
- shrinkCoinSmall :: Coin -> [Coin]
- shrinkCoinSmall (Coin c) = Coin <$> shrink c
+ shrinkCoin :: Coin -> [Coin]
+ shrinkCoin (Coin c) = Coin <$> shrink c
```

Example of a behavioural change:
```diff
 instance Arbitrary Coin where
     arbitrary = genCoinSmallPositive
-    shrink = shrinkCoinSmallPositive
+    shrink = shrinkCoin
```
- **Before**: The shrinker was limited to positive, non-zero values.
- **After**: The shrinker can now produce zero values.

It **_might_** be the case that generating zero values is fine here. But I think we should make sure, just in case. And there are other similar changes like this, within a relatively large commit that otherwise just deletes weeds.

It would be great if we could move these changes into a separate PR and give them a good review of their own!

rvl:
OK done @jonathanknowles - here is your PR:  #2768.

lib/core/test/unit/Cardano/Wallet/Primitive/Types/TokenMapSpec.hs
```diff
@@ -1051,7 +1050,7 @@ instance Arbitrary TokenQuantity where
     -- verify that the token map invariant (that a map contains no
     -- zero-valued tokens) is maintained.
     arbitrary = genTokenQuantitySmall
-    shrink = shrinkTokenQuantitySmall
+    shrink = shrinkTokenQuantitySmallPositive
```
jonathanknowles: Hi @rvl 

I think this still makes a change to the shrinking behaviour. If possible, can this also be moved to the other PR?

lib/core/src/Cardano/Wallet/Primitive/Types/UTxO.hs
```diff
@@ -107,18 +99,6 @@ instance Buildable UTxO where
...
--- Otherwise, returns the selected entry and, the UTxO minus the selected one.
-pickRandom
-    :: UTxO
-    -> IO (Maybe (TxIn, TxOut), UTxO)
-pickRandom (UTxO utxo)
-    | Map.null utxo =
-        return (Nothing, UTxO utxo)
-    | otherwise = do
-        ix <- randomRIO (0, toEnum (Map.size utxo - 1))
-        return (Just $ Map.elemAt ix utxo, UTxO $ Map.deleteAt ix utxo)
```
jonathanknowles, line 120: Ah yes! This is a vestige of the old coin selection algorithm. :)

jonathanknowles (commented):
Hi @rvl 

This looks mostly good, though I think there is still one behavioural change w.r.t. shrinking that got left behind (see review comment).

Could you also move this to the other PR (#2768), so it can be considered together with the other changes?

Many thanks!

lib/core/test/unit/Cardano/Wallet/Primitive/Types/TokenMapSpec.hs
jonathanknowles: Original shrinking behaviour restored in cb2c01deacec22597133627feabcde6b7c335952.

jonathanknowles (approved):

lib/core/test/unit/Cardano/Wallet/Primitive/Types/TokenMapSpec.hs
rvl: OK thanks. :+1:

PR #2766, opened 2021-07-19
Misc dev env fixes

rvl:
### Issue Number

Found while working on #2765.

### Overview

This is a mixed bag of small tweaks.

- ~Adds `.dir-locals.el` with conservative presets for Emacs users.~
- Add tracing of job names to Hydra evaluation (thanks for the tip @hamishmack)
- Delete some unneeded code in `shell.nix` (thanks for the tip @jbgi @nrdxp)
- Change `scripts/gen-hie.sh` to make symlinks if `hie.yaml` and `.ghci` don't exist.
- `.gitignore` and `.editorconfig` tweaks.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -28,7 +28,7 @@ let
   mkShell = name: project: project.shellFor rec {
     inherit name;
     packages = ps: lib.attrValues (selectProjectPackages ps);
-    buildInputs = (with walletPackages; [
+    nativeBuildInputs = (with walletPackages; [
         cardano-node
         cardano-cli
         cardano-address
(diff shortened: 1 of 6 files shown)


.dir-locals.el
```diff
@@ -0,0 +1,16 @@
+;; Emacs settings for cardano-wallet
```
sevanspowell: Suggestion: I think I'd prefer if this file was renamed to `.dir-locals.el.example` or something similar. It's super useful to mine for ideas, but quickly diverges between programmers, and I can't imagine dealing with the merge conflicts being fun, you'd have to stash your local changes every time you pull then pop them afterwards?

That way we could contribute ideas to the example file without having to worry about breaking other's configs.

shell.nix
```diff
@@ -59,15 +59,6 @@ let
 
     CARDANO_NODE_CONFIGS = pkgs.cardano-node-deployments;
 
-    # If any build input has bash completions, add it to the search
```
sevanspowell, line 62: Ooh, what makes this not necessary anymore?

sevanspowell (approved):

shell.nix
rvl: I put a link in the commit message to this: NixOS/nixpkgs#103501

.dir-locals.el
rvl: Yes you're right.
I will remove it, and use its text as an example in our documentation.
It would be nice to have editors working properly "out of the box."
With VSCode, you can add stuff like recommended plugins and settings into the `.vscode` directory.
But perhaps this is too difficult to do for emacs.

shell.nix
sevanspowell: Thank you :+1:

PR #2769, opened 2021-07-20
Mint/burn script witnesses groundwork

sevanspowell:
# Issue Number

ADP-955

# Overview

- Add new Alonzo-era code to mkUnsignedTx in order to mint and burn tokens.
- Add new AddressDerivation module implementing policy key derivation described in CIP-1855 (for minting and burning).
- Add cardano script and script witness conversion functions to Shelley.Compatibility module.
- Expose some internal functions in the Shelley.Compatibility module so they can be re-used in the above.
- Add PolicyK option to the "Depth" enum (for mint/burn policy keys).
- Add minimal tests to test the key derivation.

# Comments

I have done some work to hook the API up to the underlying `mkUnsignedTx` function but have shied away from committing that code, because I know the new transaction workflow is coming soon.

I've also not implemented any of the acceptance criteria for ADP-955, due to reliance on the new transaction workflow, this PR is simply laying some groundwork.

The new version of cardano-node requires the user to provide script witnesses in transaction construction, not transaction signing. The payment keys associated with that script must still sign the transaction in the signing step.

### DIFF
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
@@ -84,6 +84,8 @@ import Cardano.Wallet.Primitive.CoinSelection.MA.RoundRobin
     , prepareOutputsWith
     , selectionDelta
     )
+import Cardano.Wallet.Primitive.MintBurn
+    ( TxMintBurn )
 import Cardano.Wallet.Primitive.Types
     ( FeePolicy (..), ProtocolParameters (..), TxParameters (..) )
 import Cardano.Wallet.Primitive.Types.Address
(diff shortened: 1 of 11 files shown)


sevanspowell:
This is probably best viewed commit-by-commit.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -1249,72 +1256,119 @@ mkUnsignedTx
...
+        Nothing -> Right Cardano.TxMintNone
+        Just (val, scripts) -> tryConvertToTxMintValue era val scripts
 
-    , Cardano.txUpdateProposal =
-        Cardano.TxUpdateProposalNone
-
-    , Cardano.txMintValue =
-        Cardano.TxMintNone
-    }
+    left toErrMkTx $ Cardano.makeTransactionBody $ Cardano.TxBodyContent
```
sevanspowell, line 1269: Sorry, I had to change all the indentation here or it wouldn't compile :(

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -1249,72 +1256,119 @@ mkUnsignedTx
     -> [(Cardano.StakeAddress, Cardano.Lovelace)]
     -> [Cardano.Certificate]
     -> Cardano.Lovelace
+    -> Maybe (TokenMap, [Script KeyHash])
```
sevanspowell: Just realised this means I can only support minting, not burning. I'll need to fix this up.

sevanspowell: Fixed!

sevanspowell:
Closed in favour of smaller PRs, see the first one here: https://github.com/input-output-hk/cardano-wallet/pull/2774

PR #2774, opened 2021-07-22
Derive token policy forging keys according to CIP-1855

sevanspowell:
# Issue Number

ADP-1044

## Overview

- Add new AddressDerivation module implementing policy key derivation described in CIP-1855 (for minting and burning).
- Add PolicyK option to the "Depth" enum (for mint/burn policy keys).
- Add minimal tests to test the key derivation.


lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -167,7 +170,14 @@ import qualified Data.Text.Encoding as T
...
 data Depth
-    = RootK | PurposeK | CoinTypeK | AccountK | RoleK | AddressK | ScriptK
+    = RootK
+    | PurposeK
+    | CoinTypeK
+    | AccountK
+    | RoleK
+    | AddressK
+    | ScriptK
+    | PolicyK
```
rvl, line 190: Better update all the comments about `Depth`, starting with the one just above.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
```diff
@@ -492,6 +502,14 @@ deriveRewardAccount pwd rootPrv =
     let accPrv = deriveAccountPrivateKey pwd rootPrv minBound
     in deriveAddressPrivateKey pwd accPrv MutableAccount minBound
 
+hashVerificationKey
+    :: WalletKey k
+    => KeyRole
+    -> k d XPub
```
rvl:
```suggestion
    -> key depth XPub
```

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/MintBurn.hs
```diff
@@ -0,0 +1,99 @@
...
+import Cardano.Address.Derivation
+    ( XPrv )
+import Cardano.Address.Script
+    ( KeyHash )
+import Cardano.Crypto.Wallet
+    ( deriveXPrv )
+import Cardano.Crypto.Wallet.Types
+    ( DerivationScheme (DerivationScheme2) )
+import Cardano.Wallet.Primitive.AddressDerivation
+    ( Depth (PolicyK, PurposeK, RootK)
```
rvl: Please just use wildcards for constructor and record imports - then it's less painful to modify the code later.
```suggestion
    ( Depth (..)
```

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/MintBurn.hs
```diff
@@ -0,0 +1,99 @@
...
+  -- ^ Root private key to derive policy private key from
+  -> Index 'Hardened 'PolicyK
+  -- ^ Index of policy script
+  -> (key 'PolicyK XPrv, KeyHash)
+  -- ^ Policy private key
+derivePolicyKey pwd rootPrv policyIx = (policyK, vkeyHash)
+  where
+    policyK = liftRawKey policyPrv
+    policyPrv = derivePolicyPrivateKey pwd (getRawKey rootPrv) policyIx
+    vkeyHash = hashVerificationKey CA.Payment (publicKey policyK)
```
rvl, line 100: Purpose field here doesn't look right ... right?

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/MintBurn.hs
```diff
@@ -0,0 +1,99 @@
...
+  :: WalletKey key
+  => Passphrase "encryption"
+  -- ^ Passphrase for wallet
+  -> key 'RootK XPrv
+  -- ^ Root private key to derive policy private key from
+  -> Index 'Hardened 'PolicyK
+  -- ^ Index of policy script
+  -> (key 'PolicyK XPrv, KeyHash)
+  -- ^ Policy private key
+derivePolicyKey pwd rootPrv policyIx = (policyK, vkeyHash)
```
rvl: Looks good - perhaps you can think of a more different name for this function?

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivation/MintBurnSpec.hs
```diff
@@ -0,0 +1,95 @@
...
+import qualified Cardano.Address.Script as CA
+
+spec :: Spec
+spec = do
+    parallel $ describe "Mint/Burn Policy key Address Derivation Properties" $ do
+        it "Policy key derivation from master key works for various indexes" $
+            property prop_keyDerivationFromXPrv
+        it "Policy public key hash matches private key" $
+            property prop_keyHashMatchesXPrv
+
```
rvl, line 88: Perhaps one or two golden tests that you have generated manually with cardano-cli?

rvl (approved):
Looks good :+1:

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
sevanspowell: Mmmm... not ideal. Kinda brutalizing the abstraction here, but 'ScriptK did it first and it's the path of least resistance...

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
sevanspowell: Funnily enough I wanted to do this originally but LSP was giving me errors. Seems to be fine now :+1:

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/MintBurn.hs
sevanspowell: :thinking: I wonder if there's an LSP setting for this. These imports are the way they are because I clicked "LSP: import this for me".

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/MintBurn.hs
sevanspowell: You mean the `CA.Payment` value? I stripped out the `Role -> KeyRole` indirection because I found that confusing when initially using the function. As someone who wasn't familiar with the wallet, `Payment | Delegation :: KeyRole` made a lot more sense than `UtxoExternal | UtxoInternal | MutableAccount :: Role`, and I figured since I was the only one using it, I might as well use the simpler option. Also it doesn't feel like there's much need for abstraction here, we're literally reaching in and manipulating bytes. But I'm not strongly opposed to putting the `Role -> KeyRole` indirection back in.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
sevanspowell: Updated doco.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
sevanspowell: Fixed.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/MintBurn.hs
sevanspowell: Fixed.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/MintBurn.hs
sevanspowell: Done.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation.hs
rvl: Yeah if it weren't for ScriptK I would have asked you to try and find a better way of representing depth.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/MintBurn.hs
rvl: Yes LSP does the same for me, but it's so excessive ... and if you have LSP it shows where a symbol comes from in the tooltip.

lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/MintBurn.hs
rvl: Yeah that's all fine, but `KeyHash { role = Payment }` ? Actually I suppose Payment is suitable for this.

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivation/MintBurnSpec.hs
```diff
@@ -0,0 +1,145 @@
...
+    :: Passphrase "encryption"
+    -> XPrv
+    -> Index 'Hardened 'PolicyK
+    -> Index 'Hardened 'PolicyK
+    -> Property
+prop_keyDerivationDiffIndexDiffKey pwd masterkey policyIx1 policyIx2 =
+    key1 =/= key2
+  where
+    key1 :: XPrv
+    key2 :: XPrv
```
rvl, line 148:
```suggestion
    key1, key2 :: XPrv
```
... or just leave out the type annotations.

rvl (approved):

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDerivation/MintBurnSpec.hs
```diff
@@ -0,0 +1,237 @@
...
+            unit_comparePolicyKeys goldenTestMnemonic (Index 0x80000000)
+                "acct_xsk1tqqqnvtppk994fxm05wtlvp6uuu58srue9st8ew29vr5mxxf89tj2ze05pm8qjkyfetpzl58jkgx6dd3s96szhyxfajpc2gxv22xef0ems2cnvh5d5xjemgghg9m8489v6c9rvnt9za2pruyrtkp5y77l5ujxeug"
+                -- Child key 1855H/1815H/0H generated by cardano-address CLI
+                -- from test mnemonic
+            unit_comparePolicyKeyHashes goldenTestMnemonic (Index 0x80000000)
+                "6adb501348b99cd38172f355615d7c1d0b9d1e3fc69b565b85a98127"
+                -- Hash of child key 1855H/1815H/0H generated by cardano-address
+                -- CLI from test mnemonic
+            unit_comparePolicyKeys goldenTestMnemonic (Index 0x80000010)
+                "acct_xsk1rprds8krzhspy8xhkupk2270hzkexmkdga98zv5ns4x9d9kf89t4qp438a26sn62wsfldthkrvnwe6vnetuehgxyxt4hm46zsw03mknuwef0afcljmultpjjwx8zqm0ftnfkqucwvf5qr4gved6h3gnyeqcsur4p"
```
rvl, line 81: cardano-address did not produce the correct bech32 prefix for policy keys, but never mind.

rvl (approved):
Excellent, thanks @sevanspowell !

PR #2781, opened 2021-07-29
Provide stake pool desirability scores and pledge status

HeinrichApfelmus:
### Issue number

ADP-996, ADP-507

### Overview

In this PR, we add additional metrics to the stake pool listing in order to be able to display fairer pool rankings. Specifically, we add a *stake pool desirability score* and the *pledge status* to the `GET /v2/stake-pools` endpoint. This information is intended to supplement the ranking of pools whose non-myopic member rewards ranking is very low.

The *desirability score* reflects how cost-effective and reliable the pool operator runs their pool, and how much of their own stake they are willing to pledge, but does not take the current stake delegated to the pool into account. In the long run, only the first `k` pools with highest desirability scores are expected to attract delegators, where `k` is a protocol parameter (`k = 500` as of 13th Jan 2022). However, in the short term, rewards for delegators can be much lower than the desirability score, as the pool has not attracted enough stake to reach saturation; also, good estimates of the pool's block production performance may not be available yet.

The *pledge status* indicates whether a pool owner has delegated enough stake to their own pool in order to meet the pledge that they made when registering their pool. A pool where the `owner_stake_lower_than_pledge` flag is set will earn *no rewards*, which is a good reason to delegate elsewhere.

See also the [REST API documentation](https://input-output-hk.github.io/cardano-wallet/api/edge/#tag/Stake-Pools) for more information (available once this PR is merged).

### Details

* Example endpoint response with new fields:
```graphql
{
  …,
  "metrics": {
    …,
    "desirability_score": 0,
    "owner_stake": {
      "quantity": 0,
      "unit": "lovelace"
    },
  },
  "flags": [
    "owner_stake_lower_than_pledge"
  ],
}
```
* Computing the pool metrics takes a bit of time. In order to speed up subsequent queries, the wallet process caches the computation. This behavior can be turned off with the `--no-cache-listpools` command line flag. The cache time-to-live (TTL) can be modified with the `--cache-listpools-ttl DURATION` flag (default: 1 hour). See `cardano-wallet --help` for more information.

### Comments

* At the moment, the non-myopic rewards are used to rank stake pools in Daedalus, and the desirability score is intended to supplement this information. Research at IOG is currently looking into pool metrics that are easier to grasp and better reflect rewards in both the short and the long term.


HeinrichApfelmus:
Hrm.  I am not entirely happy with the `invariantApiStakePool` function.

With the addition of the `OwnerStakeLowerThanPledge` flag, the information contained in the API response is slightly redundant, but I think that the clarity of the flag to wallet frontends is worth it.

Ideally, one would now use a smart constructor in order to construct `ApiStakePool` values. However, the data type contains many fields, which we cannot refer to by name if we use a smart constructor. This is one of those moments where anonymous records would be very useful.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -737,6 +740,28 @@ fromNonMyopicMemberRewards =
...
+getDesirabilities
+    :: forall crypto. ()
+    => SL.RewardProvenance crypto
+    -> Map W.PoolId W.StakePoolDesirability
+getDesirabilities =
+    Map.map fromDesirability
+    . Map.mapKeys fromPoolId
+    . SL.desirabilities
+    where
+    fromDesirability (SL.Desirability x y) = W.StakePoolDesirability x y
```
jonathanknowles: Hi @HeinrichApfelmus 

Just a tiny point of style! We usually deindent `where` by two spaces, to make it stand out visually. So like this:

```suggestion
getDesirabilities =
    Map.map fromDesirability
    . Map.mapKeys fromPoolId
    . SL.desirabilities
  where
    fromDesirability (SL.Desirability x y) = W.StakePoolDesirability x y
```

darko-mijic:
@HeinrichApfelmus, can we already get results from the API in this draft PR? Uncached obviously.

If so, we could start testing this to see if we have everything we need @daniloprates

HeinrichApfelmus:
> @HeinrichApfelmus, can we already get results from the API in this draft PR? Uncached obviously.
> 
> If so, we could start testing this to see if we have everything we need @daniloprates

@darko-mijic Yes, you can get real, uncached results, please feel free to start testing. I created a branch [HeinrichApfelmus/ADP-996/daedalus-testing](https://github.com/input-output-hk/cardano-wallet/tree/HeinrichApfelmus/ADP-996/daedalus-testing) which you can reference. (The pull request here will see couple of force pushs before it is merged into master). Let me know if there is any trouble.

piotr-iohk:
@HeinrichApfelmus some ad-hoc tests I made against **mainnet** on this branch + node (bc225ae3085ba6f4f4007c50c4877bc4cebcd7de) vs cardano-wallet v2021-06-11 + node 1.27.0

This branch:

```
[cardano-wallet.network:Notice:7145] [2021-07-30 11:57:42.56 UTC] Query stakePoolsSummary took 359.389211548s (too slow)
[cardano-wallet.network:Info:7145] [2021-07-30 11:57:42.56 UTC] Fetched pool data from node tip using LSQ. Got 2860 pools in the stake distribution, and 2907 pools in the non-myopic member reward map.
[cardano-wallet.api-server:Error:7145] [2021-07-30 11:57:44.39 UTC] [RequestId 9] GET /v2/stake-pools 200 OK in 391.886787718s
```

v2021-06-11:

```
[cardano-wallet.network:Notice:69] [2021-07-30 12:12:45.79 UTC] Query stakePoolsSummary took 85.979565801s (too slow)
[cardano-wallet.network:Info:69] [2021-07-30 12:12:45.79 UTC] Fetched pool data from node tip using LSQ. Got 2860 pools in the stake distribution, and 2907 pools in the non-myopic member reward map.
[cardano-wallet.api-server:Error:69] [2021-07-30 12:12:48.18 UTC] [RequestId 7] GET /v2/stake-pools 200 OK in 113.490084737s

```

HeinrichApfelmus:
I'm trying to test commit 2bb8130bc92ebc7d0907e73cf3b04cfeeef76894 on a local cluster, but it turns out that the `totalStake` provided by `GetRewardProvenance` equals `0` there, and computations of relative stakes become divisions by zero. Ouch.

Should work on mainnet, though.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -228,17 +229,18 @@ newStakePoolLayer
     -> NetworkLayer IO (CardanoBlock sc)
     -> DBLayer IO
     -> IO ()
-    -> IO StakePoolLayer
+    -> IO (CacheWorker, StakePoolLayer)
 newStakePoolLayer gcStatus nl db@DBLayer {..} restartSyncThread = do
-    pure $ StakePoolLayer
+    (worker, _stakeDistribution) <- nooooCacheWorker hour (stakeDistribution nl) 
+    pure $ (worker, StakePoolLayer
```
rvl: HLint will probably whinge about this line.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -277,48 +279,24 @@ newStakePoolLayer gcStatus nl db@DBLayer {..} restartSyncThread = do
...
-        liftIO $ sortByReward seed
-            . map snd
-            . Map.toList
-            <$> combineDbAndLsqData
-                (timeInterpreter nl)
-                (nOpt rawLsqData)
-                lsqData
-                dbData
+        mnlData <- stakeDistribution nl
+        case mnlData of
```
rvl:
```suggestion
        stakeDistribution nl >>= \case
```

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -332,144 +310,252 @@ newStakePoolLayer gcStatus nl db@DBLayer {..} restartSyncThread = do
...
+-- runs a function periodically and caches the result.
+newtype CacheWorker = CacheWorker { runCacheWorker :: IO () }
+
+-- | For testing: Don't actually cache anything.
+nooooCacheWorker :: Int -> IO a -> IO (CacheWorker, IO a)
+nooooCacheWorker _ action = pure (CacheWorker $ pure (), action)
+
+-- | Create a new cache and a worker to fill it.
+newCacheWorker
+    :: Int -- ^ time period in seconds
```
rvl:
```suggestion
    :: NominalDiffTime -- ^ cache TTL
```

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -332,144 +310,252 @@ newStakePoolLayer gcStatus nl db@DBLayer {..} restartSyncThread = do
...
+newCacheWorker seconds action = do
+    cache <- newTVarIO Nothing
+    let worker :: IO ()
+        worker = forever $ do
+            threadDelay (3 * 1_000_000) -- 3 seconds grace period
+            a <- action
+            STM.atomically $ writeCache cache a
+            threadDelay (max 0 (seconds-3) * 1_000_000)
+    return (CacheWorker $ worker, STM.atomically $ readCache cache)
+    -- TODO: make cache worker more intelligent
```
rvl: Does it need to be much more intelligent?
Perhaps the cache period could be defined in terms of slots/epochs.
This looks like it pre-fills the cache shortly after startup. I think that's good.
It should probably recover from exceptions in the action though (perhaps by retrying a few times).
And some debug-level tracing would be good in case anything goes wrong with the queries.

rvl (commented):
Looks great!!

Before PR approval, we will just need some testing that the non-myopic rankings calculated by the wallet are the same as what would be calculated by the ledger. Currently I'm not sure what the best way to do this might be. Using the local cluster integration tests may be the easiest way.

I notice that the TVar caching is present but disabled.
Without caching, it would still be just as fast as before, right?

I tried testing it out locally and got some PastHorizonExceptions - not sure yet whether this is due to your changes, or because of the Alonzo update in master branch.

HeinrichApfelmus:
Thanks Rodney!

> Before PR approval, we will just need some testing that the non-myopic rankings calculated by the wallet are the same as what would be calculated by the ledger. Currently I'm not sure what the best way to do this might be. Using the local cluster integration tests may be the easiest way.

Well, I am very confident that I copied Eq.(2) of Section 5.5.3 in SL-D1 correctly, so if there is a discrepancy, then it is the ledger that is wrong. 😉 At least, I don't think that it's worth testing this particular equality. However, I do agree that some testing should be done. In fact, I tried the `GET /v2/stake-pools` endpoint on the local cluster, and I get a division by zero error. 😳 It turns out that the `GetRewardProvenance` query returns `totalStake = 0`; this is extraordinarily weird. It seems that the query returns bad data, I made a bug report in Jira about it. I'll make inquiries with the ledger team, but this appears to be a blocker for now. 🙈 

> I notice that the TVar caching is present but disabled.
> Without caching, it would still be just as fast as before, right?

Yes. I kept it disabled it for the purpose of testing.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
HeinrichApfelmus: > It should probably recover from exceptions in the action though (perhaps by retrying a few times).
> And some debug-level tracing would be good in case anything goes wrong with the queries.

This is the intelligence I was looking for but couldn't think of myself. Thanks! 😊

rvl:
Could we conditionally disable the GRP LSQ caching by environment variable?
We already have a few of these for testing - see [the wiki](https://github.com/input-output-hk/cardano-wallet/wiki/Testing#environment-variables).

nikolaglumac:
@HeinrichApfelmus @rvl can you please merge in the latest master branch into this PR?
AFAIK this PR doesn't use the latest available node version (1.29.0) and we need it to continue testing 🙏

rvl:
OK I have rebased, however I'm not sure whether cardano-node 1.29.0 contains a full fix for GetRewardProvenance total stake.

nikolaglumac:
Many thanks @rvl!

> OK I have rebased, however I'm not sure whether cardano-node 1.29.0 contains a full fix for GetRewardProvenance total stake.

Who could check this?

HeinrichApfelmus:
Thanks Rodney! It appears that there as been some work on the ledger side to fix the GetRewardProvenance issue, I tracked down input-output-hk/cardano-ledger-specs#2433 which has been merged. Not sure whether this has propagated to cardano-node yet.

HeinrichApfelmus:
@nikolaglumac I just learned that the fix is going to be in cardano-node 1.30.0.

nikolaglumac:
Thanks for letting us know @HeinrichApfelmus 🙏 
Please inform us once the new version of the node lands into this PR!

HeinrichApfelmus:
bors try

HeinrichApfelmus:
bors try

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
```diff
@@ -1070,42 +1081,11 @@ spec = describe "SHELLEY_STAKE_POOLS" $ do
                 -- Make sure the rewards are not all equal:
                 rewards pool1 .> rewards pool3
 
-        it "non-myopic-rewards are based on stake" $ \ctx -> runResourceT $ do
-            eventually "rewards are smaller for smaller stakes" $ do
```
piotr-iohk, line 1073: I'm wondering if it is not due to the caching perhaps? As far as I see on mainnet, both `desirability_score` and `non_myopic_member_rewards` do not change no matter what value of `stake` parameter is provided to the endpoint.

piotr-iohk:
Few notes from playing on the branch on testnet and discussing with @HeinrichApfelmus:

- Initial `list stake` pool request takes now around 300-400s on mainnet. That is significant deterioration from around 100s before. This is due to `LocalStateQuery` taking much longer in order to compute desirability score so I believe not much can be done atm.
- There is caching of LocalStateQuery data implemented, so all the subsequent requests to list stake pool are instantaneous. But I'm not sure if it does not influence returned data (https://github.com/input-output-hk/cardano-wallet/pull/2781/files#r713893894 ) :thinking: 
- The LocalStateQuery data is cached for 1h by default. There is an env variable `CACHE_LOCALSTATEQUERY_TTL` and `NO_CACHE_LOCALSTATEQUERY` documented in https://github.com/input-output-hk/cardano-wallet/wiki/Testing, however if doesn't seem to work as `CACHE_LOCALSTATEQUERY_TTL=5 cardano-wallet server...` as the request seems to be cached anyway.
- "meets pledge" status of some stake pools on mainnet does not match the result reported by e.g. pooltool.io, which is under investigation.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/StakePools.hs
HeinrichApfelmus: Can you try really big stakes on mainnet? 🤔 The `non_myopic_member_rewards` should change, especially on pools which have a low `non_myopic_member_rewards` score, when you supply a stake of `10000000000000000` (10 billion Ada).

piotr-iohk: No, I don't see them changing even for 45 billion... e.g.

```
$ curl -X GET http://localhost:8091/v2/stake-pools?stake=1 | jq | grep -B 50 AGG1

[
  {
    "flags": [],
    "metrics": {
      "saturation": 0.0003302915840810417,
      "desirability_score": 2153289257678,
      "non_myopic_member_rewards": {
        "quantity": 2153289257678, <---------------------------------
        "unit": "lovelace"
      },
      "produced_blocks": {
        "quantity": 1,
        "unit": "block"
      },
      "relative_stake": {
        "quantity": 0,
        "unit": "percent"
      },
      "owner_stake": {
        "quantity": 15206293156,
        "unit": "lovelace"
      }
    },
    "cost": {
      "quantity": 340000000,
      "unit": "lovelace"
    },
    "margin": {
      "quantity": 0,
      "unit": "percent"
    },
    "pledge": {
      "quantity": 15200000000,
      "unit": "lovelace"
    },
    "metadata": {
      "homepage": "https://www.adagogo1.com",
      "name": "ADAgogo1",
      "ticker": "AGG1",

$ curl -X GET http://localhost:8091/v2/stake-pools?stake=45000000000000000 | jq | grep -B 50 AGG1

[
  {
    "flags": [],
    "metrics": {
      "saturation": 0.0003302915840810417,
      "desirability_score": 2153289257678,
      "non_myopic_member_rewards": {
        "quantity": 2153289257678, <---------------------------------
        "unit": "lovelace"
      },
      "produced_blocks": {
        "quantity": 1,
        "unit": "block"
      },
      "relative_stake": {
        "quantity": 0,
        "unit": "percent"
      },
      "owner_stake": {
        "quantity": 15206293156,
        "unit": "lovelace"
      }
    },
    "cost": {
      "quantity": 340000000,
      "unit": "lovelace"
    },
    "margin": {
      "quantity": 0,
      "unit": "percent"
    },
    "pledge": {
      "quantity": 15200000000,
      "unit": "lovelace"
    },
    "metadata": {
      "homepage": "https://www.adagogo1.com",
      "name": "ADAgogo1",
      "ticker": "AGG1",
```

piotr-iohk: Ok, perhaps for smaller `non_myopic_member_rewards` they do change indeed:
```
$ curl -X GET http://localhost:8091/v2/stake-pools?stake=1 | jq | grep -B 35 GOAT4

    "metrics": {
      "saturation": 0.016319055947771045,
      "desirability_score": 42371242529,
      "non_myopic_member_rewards": {
        "quantity": 357659983, <---------------------------------
        "unit": "lovelace"
      },
      "produced_blocks": {
        "quantity": 44,
        "unit": "block"
      },
      "relative_stake": {
        "quantity": 0,
        "unit": "percent"
      },
      "owner_stake": {
        "quantity": 1075490098497,
        "unit": "lovelace"
      }
    },
    "cost": {
      "quantity": 340000000,
      "unit": "lovelace"
    },
    "margin": {
      "quantity": 2.5,
      "unit": "percent"
    },
    "pledge": {
      "quantity": 1000000000000,
      "unit": "lovelace"
    },
    "metadata": {
      "homepage": "https://goatstake.com",
      "name": "Goat Stake",
      "ticker": "GOAT4",

$ curl -X GET http://localhost:8091/v2/stake-pools?stake=45000000000000000 | jq | grep -B 35 GOAT4

    "metrics": {
      "saturation": 0.016319055947771045,
      "desirability_score": 42371242529,
      "non_myopic_member_rewards": {
        "quantity": 42371242529, <---------------------------------
        "unit": "lovelace"
      },
      "produced_blocks": {
        "quantity": 44,
        "unit": "block"
      },
      "relative_stake": {
        "quantity": 0,
        "unit": "percent"
      },
      "owner_stake": {
        "quantity": 1075490098497,
        "unit": "lovelace"
      }
    },
    "cost": {
      "quantity": 340000000,
      "unit": "lovelace"
    },
    "margin": {
      "quantity": 2.5,
      "unit": "percent"
    },
    "pledge": {
      "quantity": 1000000000000,
      "unit": "lovelace"
    },
    "metadata": {
      "homepage": "https://goatstake.com",
      "name": "Goat Stake",
      "ticker": "GOAT4",
```

HeinrichApfelmus: Yup. In fact, the definition of `sigma'` in the computation of `optimalRewards`,

https://github.com/input-output-hk/cardano-wallet/blob/85be73804bb6a243a8a8f73e13d144320c17dc46/lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs#L364-L384

implies that we should actually *expect* the non-myopic rewards to *not* change with the user-supplied stake for the `nOpt` pools that rank best on desirability score — the non-myopic reward prediction assumes that they will each attract an equal share `z0 = 1/nOpt` of the stake, and adding more stake to them will *not* increase the rewards at all, as `sigma'` will not increase beyond `z0`.

If we want to replicate the concept of "smaller pool" in the integration test, I think we should set the optimum number of pools to `2` in the genesis file (or any number that is small than the total number of pools that stays active in the local cluster.)

piotr-iohk: Ok, I understand that this change was introduced in this PR to comply with delegation design spec? 

>If we want to replicate the concept of "smaller pool" in the integration test, I think we should set the optimum number of pools to 2 in the genesis file (or any number that is small than the total number of pools that stays active in the local cluster.)

I think that's a good idea + changing this test to verify this. There are a lot of "smaller pools" on mainnet. So if we had nOpt = 2 (currently it is 3 I believe) then we would have, at least 1 "smaller pool" (as there are 3 pools in the integration tests). So, for this one pool the change of non-myopic rewards score should be visible in case of changing stake.

HeinrichApfelmus:
@piotr-iohk

* I have implemented new command line flags (`--no-cache-listpools`, `--cache-listpools-ttl`) for the `cardano-wallet` executable, which still ignores the environment variables. Is this behavior acceptable and/or documented well enough for your taste?
* Unfortunately, it turns out that the "meet pledge" status cannot be fixed without better data from the node. We will have to wait for a new node version.

piotr-iohk:
> @piotr-iohk
> 
> * I have implemented new command line flags (`--no-cache-listpools`, `--cache-listpools-ttl`) for the `cardano-wallet` executable, which still ignores the environment variables. Is this behavior acceptable and/or documented well enough for your taste?
> * Unfortunately, it turns out that the "meet pledge" status cannot be fixed without better data from the node. We will have to wait for a new node version.

Looks good! I've included command line flags in last revision of https://github.com/input-output-hk/cardano-wallet/wiki/Wallet-command-line-interface

nikolaglumac:
Hi @HeinrichApfelmus 👋 
Any update/ETA on this PR?

HeinrichApfelmus:
Hi @nikolaglumac 👋 
This PR is waiting for the merge of https://github.com/input-output-hk/ouroboros-network/pull/3423 (and subsequent node release).

nikolaglumac:
Please keep us posted 🙏 
Many thanks!

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -1294,8 +1310,10 @@ fromShelleyAddress = W.Address
 fromShelleyCoin :: SL.Coin -> W.Coin
 fromShelleyCoin (SL.Coin c) = Coin.unsafeFromIntegral c
 
+{-
```
rvl: Just delete it

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
```diff
@@ -328,144 +309,280 @@ newStakePoolLayer gcStatus nl db@DBLayer {..} restartSyncThread = do
...
+-- in order to rank them by desirability,
+-- and we need to know the stake that the user wants to delegate.
+scorePools
+    :: Ord poolId
+    => RewardParams
+    -> Map poolId (RewardInfoPool, a)
+    -> Coin -- ^ Stake that the user wants to delegate
+    -> Map poolId (PoolScore, RewardInfoPool, a)
+scorePools params pools t
+    = Map.fromList $ zipWith doScore sortedByDesirability areTop
```
rvl, line 436: We should first sort by non-myopic rewards, then sort by desirability, and then sort by seeded random order.

lib/shelley/test/data/cardano-node-shelley/shelley-genesis.yaml
```diff
@@ -31,7 +31,7 @@ protocolParams:
   maxBlockHeaderSize: 217569
   keyDeposit: 1000000
   keyDecayRate: 0
-  nOpt: 3
+  nOpt: 2
```
rvl, line 34: I think it's pretty hard to use the local cluster and integration tests for testing this change.

Testing the new pool listing on existing mainnet pools and comparing with the old implementation will be more fruitful.

specifications/api/swagger.yaml
```diff
@@ -1764,42 +1764,74 @@ x-stakePoolSaturation: &stakePoolSaturation
   type: number
   minimum: 0
   description: |
-    Saturation-level of the pool based on the desired number of pools aimed by the network.
-    A value above `1` indicates that the pool is saturated.
+    The saturation of a pool is the ratio of the current `relative_stake`
```
rvl, line 1767: Nice updates to the docs.

rvl (commented):
Looks great, but I just have a query about the sorting of the pools list.

Would you also be able to write some explanation in the PR description of what will change for Daedalus users in the stake pools list? Because this is something we will want to highlight in the release notes of the next version.

lib/shelley/src/Cardano/Wallet/Shelley/Pools.hs
HeinrichApfelmus: Ok, I will adjust the value to sort by. (The actual sorting is done in the `sortByReward` function, which currently sorts according to non-myopic rewards. The code snippet above is internal to the `scorePools` function.)

HeinrichApfelmus: Actually, it turns out that the non-myopic member rewards will never be equal to exactly zero, as they are essentially equal to the rewards for pool stake = owner stake + user stake. Hence, there will probably never be any ties that are resolved by the additional sorting by desirability score. Nonetheless, I will add that.

HeinrichApfelmus:
Thanks for the review, Rodney! 😊 

> Would you also be able to write some explanation in the PR description of what will change for Daedalus users in the stake pools list?

Sure, I have extended the PR description. As I do not know how exactly Daedalus will display the pool metrics to users, I have opted to give a more general description of the how the metrics work, leaving a more user-centric description to our front-end team.

HeinrichApfelmus:
bors try

rvl (approved):
:+1:

HeinrichApfelmus:
bors try

HeinrichApfelmus:
Status update for @nikolaglumac : The pledge status and desirability status have been implemented and tested, but unfortunately, it unexpectedly turns out that the query is too slow to be useable on mainnet in node 1.33 (but this is likely to improve in newer node versions). If it's really urgent, we could reverse-engineer at least the pledge status with better performance; if so, please chat about this with Teddy, our product owner.

nikolaglumac:
I leave this up to @daniloprates.

thedanheller:
> If it's really urgent, we could reverse-engineer at least the pledge status with better performance; if so, please chat about this with Teddy, our product owner.

@HeinrichApfelmus it would be great to have at least pledge status, if possible, although I wouldn't say it's more urgent than releasing support for Cardano Node 1.33.

abailly:
This PR is 3 years old and would probably take longer to rebase than recreate. Closing.

PR #2782, opened 2021-07-29
Balance endpoint - to be retired 

paweljakubas:
# Issue Number

ADP-656

# Overview

- [x] added endpoint, all Api types, swagger
- [x] Api types tested in unit testing and passing
- [x] added Server's balanceTransaction function capturing the logic
- [x] extended Transaction interface : updateTx
- [x] added deserialisation tests in shelley unit tests
- [x] added error handling
- [x] did some very first version coin selection with external inputs, to be replaced with ADP-1070
- [x] tested in integration testing several errors (already covered tx, wrong format, cannot deserialise) plus successful rebalance - with wallet's generated CBORs 
- [x] implement updateSealedTx in shelley being impl of interface updateTx
- [x] add testing for deserialisation/coin selection for plutus examples

# Comments

Based on the PR #2896 branch.


lib/core/src/Cardano/Wallet/Transaction.hs
```diff
@@ -187,20 +236,103 @@ withdrawalToCoin = \case
...
+    first f (SignTransactionResult t a w c) =
+        SignTransactionResult t (fmap (fmap f) a) (fmap (fmap f) w) (fmap (fmap f) c)
+    second f (SignTransactionResult t a w c) =
+        SignTransactionResult (f t) a w c
+
+-- | In the wallet, signing keys are symmetrically encrypted at rest using a key
+-- derived from the user's spending passphrasee. This data type pairs an
+-- encrypted signing key with the passphrase needed to decrypt it.
+--
+-- We call this type "decrypted", because although the key is still encrypted, the passphrase is right there, so it may as well be decrypted.
```
```suggestion
-- We call this type "decrypted", because although the key is still encrypted,
-- the passphrase is right there, so it may as well be decrypted.
```

lib/test-utils/src/Test/Utils/Pretty.hs
```diff
@@ -39,3 +40,8 @@ instance Show a => ToText (Pretty a)
 -- the "Fmt" module.
 pShowBuilder :: Show a => a -> Builder
 pShowBuilder = fromLazyText . pShow
+
+-- | Like '===, but prettier.
```
jonathanknowles:
```suggestion
-- | Like '===', but prettier.
```

specifications/api/swagger.yaml
```diff
@@ -2755,6 +2858,47 @@ components:
...
+      type: object
+      required:
+        - txIn
+        - txOut
+      properties:
+        txIn: *txInNode
+        txOut: *txOutNode
+
+    ApiExternalInputs: &ApiExternalInputs
+      description: A flat list of node compliant inputs.
```
sevanspowell, line 2920: I think I'm missing something, can't we just get the inputs from the serialised transaction itself?

specifications/api/swagger.yaml
```diff
@@ -2755,6 +2858,47 @@ components:
             - <<: *transactionWithdrawalRequestSelf
             - <<: *transactionRedemptionRequest
 
+    ApiTransactionExtended: &ApiTransactionExtended
```
sevanspowell, line 2899: Could we perhaps describe what we expect this transaction to look like? I can tell from this description that it's CBOR encoded, but what might the transaction look like before it's CBOR encoded? Do we expect it to be missing inputs and/or outputs?

sevanspowell: Just as someone with no context, I'm not sure how I might generate a transaction to call this endpoint.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2016,6 +2023,85 @@ postTransactionFeeOld ctx (ApiT wid) body = do
...
+
+    toTxIn (ApiCoinSelectionInput (ApiT txid) ix _ _ _ _) =
+        TxIn txid ix
+    toTxOut (ApiCoinSelectionChange (ApiT addr,_) (Quantity amt) (ApiT assets) _) =
+        TxOut addr (TokenBundle (Coin $ fromIntegral amt) assets)
+    getInpsChange cs =
+        ( map toTxIn $ NE.toList $ cs ^. #inputs
+        , map toTxOut $ cs ^. #change)
+
+    areOutputsCovered (Tx _ feeM _ outs _ _) = -- TODO in ADP-656 is it correct- double check
```
sevanspowell, line 2057: I think there's value in moving meaningful logic outside the handler. For example, this concept of "Is this transaction balanced" could be useful to other modules, and is more easily tested outside the handler. Inside the handler, it's very difficult to test this logic.

lib/core-integration/src/Test/Integration/Framework/TestData.hs
```diff
@@ -621,6 +622,9 @@ errMsg400ScriptNotUniformRoles :: String
 errMsg400ScriptNotUniformRoles =
     "All keys of a script must have the same role: either payment or delegation."
 
+errMsg403OutputsAlreadyCovered :: String
+errMsg403OutputsAlreadyCovered =
+    "The transaction to be balanced has already all outputs covered. Please send a transaction that requires more inputs to be picked to be balanced."
```
Anviking, line 627: If PAB _happens_ to construct a balanced tx, shouldn't the balance endpoint be a no-op rather than failing?

> Please send a transaction that requires more inputs to be picked to be balanced.

"Please find a way to make the tx more expensive" sounds like an odd recommendation...

paweljakubas: yes, message is unfortunate for two reasons: first of all, unbalanced tx is the one that lacks outputs OR inputs, Secondly, we should rather write that sent tx is balanced and does not need further balancing but signing and then submitting.

paweljakubas: I resigned from this error naming in https://github.com/input-output-hk/cardano-wallet/pull/2900

Anviking: I was more wondering: why not remove the error?

If the endpoint succeeds in the — I take rare event of PAB submitting an already balanced tx, then they wouldn't have to add special handling for:
> we should rather write that sent tx is balanced and does not need further balancing but signing and then submitting.

Anviking: Also seemed to me that it _might_ be easier for _us too_, to not care about defining `areOutputsCovered`. I would presume the code responsible for balancing the tx would already be able to notice when the tx was balanced, and return without any changes… but could be wrong.

paweljakubas: The problem is that I am reusing ApiCoinSelection in the response of endpoint and it cannot be empty.So in case of already balance tx I tentatively choose to use some error to circumvent this problem. As a first step.

rvl:
Some of this PR has moved into #2900 and #2906. @paweljakubas, would you please be able to update the title and description accordingly?

Anviking:
@paweljakubas this PR can be closed, right?

paweljakubas:
Closed in favor of https://github.com/input-output-hk/cardano-wallet/pull/2906

PR #2787, opened 2021-07-30
Initial Cardano.Wallet.Prelude and local package re-org

rvl:
### Description

Basically we want to solve these problems:

- Massive numbers of imports in every single source file which makes editing and refactoring especially quite laborious.

- Tangled dependencies between cardano-wallet-core and cardano-wallet-shelley packages which cause tangled code.

- Testing code mixed up with the application code.

- Misleading module and package names where cardano-wallet-specific code has overly general names.

#### This PR

This is the first cut of Cardano.Wallet.Prelude.

- Adds Cardano.Wallet.Prelude to the cardano-wallet-core package.
- Adds a cardano-wallet-base package containing shared util code.
- Decouples launcher, core, text-class, and test-utils packages from each other.
  - cardano-wallet-test-utils is now a standalone package, not depending on other wallet packages such as text-class.
  - Removes cardano-wallet-launcher dependency on cardano-wallet-text-class.

- Sets all cabal files to spec version 3.0 (3.4 is not supported by stack).
- Adds some Cabal sub-libs for testing code. QuickCheck generators are no longer in the application library and instead have their own separate sub-lib.
- Adds common stanzas to the cabal files to reduce duplication.
- Resolves a great number of unused-packages and redundant-constraints warnings.
- Renames and moves a few modules. For example, the miniprotocols client code is now under Cardano.Wallet.Network.
- Renames wallet-specific packages to cardano-wallet-*.

#### Next PRs:

- Disentangle the cardano-wallet and cardano-wallet-core packages.
- Get rid of cardano-wallet-core-integration.

### Comments

Tips for reviewing: This diff is huge but it's pretty much all deletions, or renaming, or fiddling whitespace (sorry). Concentrate on 
what's exported from `Cardano.Wallet.Prelude`, and what's changed in the Cabal files. The rest is pretty boring. With Cabal common stanzas, some modules got warnings enabled that they previously didn't have, so there are also some tweaks to resolve these warnings.

The contents of the Cardano.Wallet.Prelude will probably need adjustment. Unless it's really wrong, I would propose refining it in future PRs, because this PR is horrible to rebase.

#### Package dependencies before
packages-master-20220115 dot

#### Package dependencies after
packages-pr-20220115 dot

### Issue

ADP-1010, ADP-1322


lib/core-integration/src/Test/Integration/Scenario/API/Byron/Transactions.hs
```diff
@@ -160,9 +147,9 @@ spec = describe "BYRON_TRANSACTIONS" $ do
                 (Link.getWallet @'Shelley wDest) Default Empty
             verify rb
                 [ expectField (#assets . #available . #getApiT)
-                    (`shouldNotBe` TokenMap.empty)
+                    (`shouldNotBe` mempty)
                 , expectField (#assets . #total . #getApiT)
-                    (`shouldNotBe` TokenMap.empty)
+                    (`shouldNotBe` mempty)
```
jonathanknowles: I'm not too fond of using `mempty` in places like these, because:

1. The reader has to do extra work to determine what type of `mempty` is being used. (harder to review in a web browser.)
2. By explicitly stating the type, we get extra protection against accidentally testing something else that also happens to have a `Monoid` instance.

Given that it's also fairly idiomatic for data structure types to have an explicit `empty` value, I think it makes our code more idiomatic if we do the same. For example:

- [Data.Map.Strict.empty](https://hackage.haskell.org/package/containers-0.6.5.1/docs/Data-Map-Strict.html#v:empty)
- [Data.Set.empty](https://hackage.haskell.org/package/containers-0.6.5.1/docs/Data-Set.html#v:empty)
- [Data.ByteString.empty](https://hackage.haskell.org/package/bytestring-0.11.1.0/docs/Data-ByteString.html#v:empty)
- [Data.Text.empty](https://hackage.haskell.org/package/text-1.2.4.1/docs/Data-Text.html#v:empty)
- [Data.IntMap.Strict.empty](https://hackage.haskell.org/package/containers-0.6.5.1/docs/Data-IntMap-Strict.html#v:empty)

I actually quite like seeing terms like `Map.empty` and `Set.empty` in places where the context doesn't make it clear what type is being used.

I think `mempty` is okay in contexts where the type is obvious, for example, in a `fold` whose result has the type of the function.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -1349,7 +1328,7 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
                     Nothing
             rl <- request @([ApiTransaction n]) ctx linkList Default Empty
             verify rl [expectListSize 2]
-            pure (getFromResponse Prelude.id rl)
+            pure (getFromResponse idFunc rl)
```
jonathanknowles, line 1329: Hi @rvl. Could you say a little bit about the motivation for renaming `id`?

If it's because we want to use `id` for record fields, then in future we can use the [`NoFieldSelectors`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/field_selectors.html#extension-FieldSelectors) language extension, which means code like the following would work just fine:

```hs
{-# LANGUAGE NoFieldSelectors #-}

data Wibble = Wibble
    { id :: WibbleId
    , ...
    }
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs
```diff
@@ -871,7 +871,7 @@ runSelection limit mExtraCoinSource available minimumBalance =
   where
     initialState :: SelectionState
     initialState = SelectionState
-        { selected = UTxOIndex.empty
+        { selected = mempty
```
jonathanknowles: Similarly to my earlier comment, I think the use of `mempty` here makes the code harder to read, especially in a web-browser.

If someone is using an editor with type feedback, then of course, they can look it up, but it's an extra step that isn't currently necessary.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
```diff
@@ -167,25 +166,26 @@ data UTxOIndex = UTxOIndex
...
+instance Semigroup UTxOIndex where
+    a <> b = fromSequence (toList a <> toList b)
+
+instance Monoid UTxOIndex where
+    mempty = UTxOIndex
+        { assetsAll = Map.empty
+        , assetsSingleton = Map.empty
+        , coins = Set.empty
+        , balance = mempty
+        , utxo = Map.empty
```
jonathanknowles: In the above expression, the type of each of the field is obvious to someone casually reading through, except for `balance`!

I reckon this would be clearer:
```hs
    mempty = UTxOIndex
        { assetsAll = Map.empty
        , assetsSingleton = Map.empty
        , coins = Set.empty
        , balance = TokenBundle.empty
        , utxo = Map.empty
        }
 ```

File: lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs [hunk 5]
```diff
@@ -197,15 +197,15 @@ singleton i o = insertUnsafe i o empty
 -- and all others will be ignored.
 --
 fromSequence :: Foldable f => f (TxIn, TxOut) -> UTxOIndex
-fromSequence = flip insertMany empty
+fromSequence = flip insertMany mempty
```
jonathanknowles, line 197: OTOH, I think it's perfectly fine to use `mempty` here, because the context makes it obvious what the type is.

Just to emphasise that I'm not against `mempty` per-se: 😄

jonathanknowles (commented):
Hi @rvl 

I like the direction this is going in 👍🏻, though I have a few reservations about the use of `mempty`. (See comments.)

I haven't looked through everything yet, so will come and take another look tomorrow!

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
rvl: Yes, `id` is a common and useful variable name, and not just for record fields.

You still have the `id` function in scope, and `idFunc` is an additional name which is shorter than `Cardano.Wallet.Prelude.id`.

Enabling `NoFieldSelectors` would cause rather more inconvience than simply aliasing a function, and it's not available in the GHC version we are using anyway.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
rvl: I changed them all to `mempty`. So the zero UTxOIndex is the product of 5 things which are all zero. Nobody should be particularly concerned about the types here.

lib/core-integration/src/Test/Integration/Scenario/API/Byron/Transactions.hs
rvl: I can't agree with this, sorry.

Using `Monoid` we can reduce this:
```haskell
expectField (#assets . #total . #getApiT) (`shouldNotBe` TokenMap.empty)
```
to this:
```haskell
expectField (#assets . #total) (`shouldNotBe` mempty)
```
It reads nicely as a test case assertion, and fits better in 80 columns too. :wink:

The type of empty isn't particularly relevant. We just want to know that total assets are not zero.

Not really sure about `Data.Text.empty` and `Data.ByteString.empty` being idiomatic. They are superfluous in my opinion.

lib/base/src/Cardano/Wallet/Base.hs
```diff
@@ -0,0 +1,201 @@
...
+    , Tracer (..)
+    , contramap
+    , traceWith
+    , nullTracer
+    , say
+    , sayErr
+
+    -- * Debugging
+    , traceShowId
+    , trace
```
Anviking, line 97: Practical, but might be possible to confuse with `traceWith`?

Anviking (approved):

lib/core-integration/src/Test/Integration/Scenario/API/Byron/Transactions.hs
```diff
@@ -159,10 +146,8 @@ spec = describe "BYRON_TRANSACTIONS" $ do
             rb <- request @ApiWallet ctx
                 (Link.getWallet @'Shelley wDest) Default Empty
             verify rb
-                [ expectField (#assets . #available . #getApiT)
-                    (`shouldNotBe` TokenMap.empty)
-                , expectField (#assets . #total . #getApiT)
-                    (`shouldNotBe` TokenMap.empty)
+                [ expectField (#assets . #available) (`shouldNotBe` mempty)
+                , expectField (#assets . #total) (`shouldNotBe` mempty)
```
jonathanknowles, line 150: Suggestion: give these `mempty` instances a type annotation.

```suggestion
                [ expectField (#assets . #available)
                    (`shouldNotBe` (mempty :: TokenMap))
                , expectField (#assets . #total)
                    (`shouldNotBe` (mempty :: TokenMap))
```

lib/core-integration/src/Test/Integration/Scenario/API/Byron/Transactions.hs
```diff
@@ -213,10 +198,8 @@ spec = describe "BYRON_TRANSACTIONS" $ do
...
```
jonathanknowles, line 202: Suggestion: give these `mempty` instances a type annotation.

```suggestion
                [ expectField (#assets . #available)
                    (`shouldNotBe` (mempty :: TokenMap))
                , expectField (#assets . #total)
                    (`shouldNotBe` (mempty :: TokenMap))
```

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -611,8 +590,8 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
         w <- fixtureMultiAssetWallet ctx
         r <- request @ApiWallet ctx (Link.getWallet @'Shelley w) Default Empty
         verify r
-            [ expectField (#assets . #available . #getApiT) (`shouldNotBe` TokenMap.empty)
-            , expectField (#assets . #total . #getApiT) (`shouldNotBe` TokenMap.empty)
+            [ expectField (#assets . #available) (`shouldNotBe` mempty)
+            , expectField (#assets . #total) (`shouldNotBe` mempty)
```
jonathanknowles, line 594: Suggestion: give these `mempty` instances a type annotation.

```suggestion
            [ expectField (#assets . #available)
                (`shouldNotBe` (mempty :: TokenMap))
            , expectField (#assets . #total)
                (`shouldNotBe` (mempty :: TokenMap))
```

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -651,12 +630,9 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
...
             verify rb
-                [ expectField (#assets . #available . #getApiT)
-                    (`shouldNotBe` TokenMap.empty)
-                , expectField (#assets . #total . #getApiT)
-                    (`shouldNotBe` TokenMap.empty)
-                , expectField (#balance . #total . #getQuantity)
-                    (`shouldBe` minCoin)
+                [ expectField (#assets . #available) (`shouldNotBe` mempty)
+                , expectField (#assets . #total) (`shouldNotBe` mempty)
+                , expectField (#balance . #total) (`shouldBe` Quantity minCoin)
```
jonathanknowles, line 635: Suggestion: give these `mempty` instances a type annotation.

```suggestion
                [ expectField (#assets . #available)
                    (`shouldNotBe` (mempty :: TokenMap))
                , expectField (#assets . #total)
                    (`shouldNotBe` (mempty :: TokenMap))
                , expectField (#balance . #total)
                    (`shouldBe` Quantity minCoin)
```

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -709,8 +685,8 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
             rb <- request @ApiWallet ctx
                 (Link.getWallet @'Shelley wDest) Default Empty
             verify rb
-                [ expectField (#assets . #available . #getApiT) (`shouldNotBe` TokenMap.empty)
-                , expectField (#assets . #total . #getApiT) (`shouldNotBe` TokenMap.empty)
+                [ expectField (#assets . #available) (`shouldNotBe` mempty)
+                , expectField (#assets . #total) (`shouldNotBe` mempty)
```
jonathanknowles, line 689: Suggestion: give these mempty instances a type annotation.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -821,8 +797,8 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
...
```
jonathanknowles, line 801: Suggestion: give these mempty instances a type annotation.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -2423,14 +2399,10 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
...
-                , expectField
-                        (#balance . #available) (`shouldBe` amt)
-                , expectField
-                        (#assets . #total) (`shouldBe` mempty)
-                , expectField
-                        (#assets . #available) (`shouldBe` mempty)
+                [ expectField (#balance . #total) (`shouldBe` amt)
+                , expectField (#balance . #available) (`shouldBe` amt)
+                , expectField (#assets . #total) (`shouldBe` mempty)
+                , expectField (#assets . #available) (`shouldBe` mempty)
```
jonathanknowles, line 2405: Suggestion: give these mempty instances a type annotation.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -508,7 +492,7 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
         let expectedTxOutTarget = WalletOutput $ ApiWalletOutput
                 { address = addrDest
                 , amount = Quantity amt
-                , assets = ApiT TokenMap.empty
+                , assets = ApiT mempty
```
jonathanknowles, line 529: Suggestion: give this `mempty` instance a type annotation.

```suggestion
                , assets = ApiT (mempty :: TokenMap)
```

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -579,7 +563,7 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
         let expectedTxOutTarget' = WalletOutput $ ApiWalletOutput
                 { address = addrDest
                 , amount = Quantity amt
-                , assets = ApiT TokenMap.empty
+                , assets = ApiT mempty
```
jonathanknowles, line 604: Suggestion: give this `mempty` instance a type annotation.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -603,7 +587,7 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
         let expectedTxOutSource = WalletOutput $ ApiWalletOutput
                 { address = addrSrc
                 , amount = Quantity $ initialAmt - (amt + fromIntegral expectedFee)
-                , assets = ApiT TokenMap.empty
+                , assets = ApiT mempty
```
jonathanknowles, line 628: Suggestion: give this `mempty` instance a type annotation.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -810,8 +794,8 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
                 , expectField
                         (#balance . #available . #getQuantity)
                         (`shouldBe` amt)
-                , expectField (#assets . #available . #getApiT) (`shouldNotBe` TokenMap.empty)
-                , expectField (#assets . #total . #getApiT) (`shouldNotBe` TokenMap.empty)
+                , expectField (#assets . #available . #getApiT) (`shouldNotBe` mempty)
+                , expectField (#assets . #total . #getApiT) (`shouldNotBe` mempty)
```
jonathanknowles: Suggestion: give these `mempty` instances a type annotation.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -904,8 +888,8 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
                 , expectField
                         (#balance . #available . #getQuantity)
                         (`shouldBe` inTxAmt)
-                , expectField (#assets . #available . #getApiT) (`shouldNotBe` TokenMap.empty)
-                , expectField (#assets . #total . #getApiT) (`shouldNotBe` TokenMap.empty)
+                , expectField (#assets . #available . #getApiT) (`shouldNotBe` mempty)
+                , expectField (#assets . #total . #getApiT) (`shouldNotBe` mempty)
```
jonathanknowles, line 1011: Suggestion: give these `mempty` instances a type annotation.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/BalanceSpec.hs
```diff
@@ -679,13 +658,13 @@ prop_performSelection_small mockConstraints (Blind (Small params)) =
         "not nonZeroExtraCoinSource && not nonZeroExtraCoinSink" $
 
     -- Inspect the sets of minted and burned assets:
-    cover 20 (view #assetsToMint params /= TokenMap.empty)
+    cover 20 (view #assetsToMint params /= mempty)
```
jonathanknowles: I feel this is one of those cases where indicating the type helps with readability.

In particular, it's useful to distinguish between `TokenMap` and `TokenBundle`, as only the latter type allows ada quantities.

Perhaps we can use a type annotation? (`mempty :: TokenMap`).

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOIndex/Internal.hs
```diff
@@ -178,25 +163,26 @@ data UTxOIndex = UTxOIndex
 
 instance NFData UTxOIndex
 
+instance Semigroup UTxOIndex where
+    a <> b = fromSequence (toList a <> toList b)
```
jonathanknowles, line 167: @rvl Is there any reason to add this instance of `Semigroup` for `UTxOIndex`?

If so, perhaps we can add a check to `UTxOIndexSpec` for lawfulness?

Like so:

```patch
     describe "Cardano.Wallet.Primitive.Types.UTxOIndexSpec" $ do
    
     parallel $ describe "Class instances obey laws" $ do
         testLawsMany @UTxOIndex
         [ eqLaws
+        , monoidLaws
+        , semigroupLaws
         ]  
```

jonathanknowles (commented):
Hi @rvl 

Overall I do appreciate the large reduction in import noise. Thanks!

Though I feel we've lost a bit of readability and automatic compiler checking by converting all of the `empty` functions to `mempty`.

I find that in test code, it's useful for both readability and correctness to specify that we really are expecting a `TokenMap`, rather than something else (like `TokenBundle` or a `Coin`).

Would it be possible to add type annotations to these places, especially in the integration tests? I've highlighted **_some_** (see comments), but there are way too many for me to leave comments for all of them! Some lens traversals make it hard to know exactly what type is expected.

As an aside, it seems to be quite idiomatic in the Haskell library world to define an `empty` value when defining a container type. For instance:
- [Data.Map.Strict.empty](https://hackage.haskell.org/package/containers-0.6.5.1/docs/Data-Map-Strict.html#v:empty)
- [Data.HashMap.Strict.empty](https://hackage.haskell.org/package/unordered-containers-0.2.16.0/docs/Data-HashMap-Strict.html#v:empty)
- [Data.ByteString.empty](https://hackage.haskell.org/package/bytestring-0.11.2.0/docs/Data-ByteString.html#v:empty)
- [Data.ByteString.Short.empty](https://hackage.haskell.org/package/bytestring-0.11.2.0/docs/Data-ByteString-Short.html#v:empty)
- [Data.ByteString.Builder.empty](https://hackage.haskell.org/package/bytestring-0.11.2.0/docs/Data-ByteString-Builder-Internal.html#v:empty)
- [Data.Sequence.empty](https://hackage.haskell.org/package/containers-0.6.5.1/docs/Data-Sequence.html#v:empty)
- [Data.Text.empty](https://hackage.haskell.org/package/text-2.0/docs/Data-Text-Internal-Lazy.html#v:empty)
- ...

One advantage of this idiom is that we don't always need or want to define a `Semigroup` instance. (The `Semigroup` instance for `UTxOIndex` isn't really needed, for instance.)

HeinrichApfelmus (approved):
Thanks! 😊 Looks like a great step forward to me.

I like that we stick to the standard `Prelude` and export some often used additional functions in `Cardano.Wallet.Base`.

I'm wondering whether we may want to switch the two names `Cardano.Wallet.Prelude` and `Cardano.Wallet.Base`, though? 🤔 The idea being that `Cardano.Wallet.Prelude` would be "`Prelude` + the stuff that was 'clearly' missing from the `Prelude`", whereas `Cardano.Wallet.Base` would be "the fully stuffed `Prelude` + often used functions that are more specific to the wallet".

rvl:
Looks like we will need PR #3115 merged before this. Stack doesn't seem to be working - even for the private internal libraries case.

HeinrichApfelmus:
This pull request tried to do too many things at once and grew stale too fast. I have saved the module dependency graphs. Closing.

PR #2788, opened 2021-07-30
Buildkite: Clean up .hie files before stack build

### LINKED ISSUE #2785: Weeder picks up stale `.hie` files from build cache and fails
# Context

When making a stack build, there are `.hie` files generated in `.stack-work` that weeder users.
The `.stack-work` directories are saved in cache between builds.

If a source file is removed, the `.hie` file still remains in the `.stack-work` directory.

This causes false positives from weeder.

# Test Case

# Failure / Counter-example

---

# Resolution 

---

# QA

rvl:
### Issue Number

Resolves #2785

### Overview

Wipes out all the `.hie` files in `.stack-work` before starting the build.

### DIFF
--- a/.buildkite/rebuild.hs
+++ b/.buildkite/rebuild.hs
@@ -233,7 +233,7 @@ setupBuildDirectory dryRun buildDir = do
 -- Remove certain files which get cached but could cause problems for subsequent
 -- builds.
 cleanBuildDirectory :: FilePath -> IO ()
-cleanBuildDirectory buildDir = findTix buildDir >>= mapM_ rm
+cleanBuildDirectory buildDir = mapM_ rm =<< findTix buildDir <> findHie buildDir
 
 -- | Check for the presence of new or modified test data files which someone
 -- forgot to check in.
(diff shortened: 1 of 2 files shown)


.buildkite/rebuild.hs
```diff
@@ -233,7 +233,9 @@ setupBuildDirectory dryRun buildDir = do
 -- Remove certain files which get cached but could cause problems for subsequent
 -- builds.
 cleanBuildDirectory :: FilePath -> IO ()
-cleanBuildDirectory buildDir = findTix buildDir >>= mapM_ rm
+cleanBuildDirectory buildDir = mapM_ rm =<< do
+    findTix buildDir
+    findHie buildDir
```
Anviking: I'm a bit confused by this do-notation: this is in `IO` and isn't the results from `findTix` simply dropped?

Anviking (approved):
Hope it works! One comment: https://github.com/input-output-hk/cardano-wallet/pull/2788/files#r679886779

.buildkite/rebuild.hs
rvl: Err duh yes you're right - fixed with the monoid instance of `IO a`.

PR #2763, opened 2021-07-15
Basic Alonzo support

Anviking:
# Issue Number

ADP-1025

# Overview

- [x] Ensure the wallet can run in Alonzo
- [x] Get all integration tests green
- [x] Fix warnings
- [x] Cleanup
    - <s>[ ]Test that addresses don't affect the min ada value</s> Existing integration tests should be enough
    - [x] <s>Does it still work on testnet? -> The ledger function in question cannot know the current network, so would be highly surprising if it didn't work on testnet </s>
    - [x] Update MinUTxOValue properties to generate both ShelleyMA and Alonzo cases
    - [x] Go through everything from start to finish 
- [x] Run integration tests on both Mary and alonzo
    - [x] Check that the Mary era is actually picked up, and it's not just running Alonzo twice
- <s>[ ] Using cardano-node 1.28.0 tag</s> -> #2778
- <s>[x] Cabal build also works</s> -> #2776

# Comments

- [Hydra jobset](https://hydra.iohk.io/jobset/Cardano/cardano-wallet-pr-2763)
- [Cabal Nightly build](https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds?branch=anviking%2FADP-952%2Falonzo) (use environment variable `step=cabal` when creating a new build)

PR #2791, opened 2021-07-31
Negotiate NodeToClientVersion for non-breaking mainnet support

Anviking:
# Issue Number

ADP-1025

# Background

#2763 made the wallet demand NodeToClientV_9. With cardano-node 1.28.0, V_9 is only enabled if  `TestEnableDevelopmentNetworkProtocols` is set in the node config.

# Overview

- [x] Make the wallet support both `V_8` and `V_9`.
- [x] Ensure integration tests on Mary are run without `TestEnableDevelopmentNetworkProtocols: True`

# Comments

### DIFF
--- a/lib/shelley/bench/Restore.hs
+++ b/lib/shelley/bench/Restore.hs
@@ -133,7 +133,11 @@ import Cardano.Wallet.Primitive.Types.UTxO
 import Cardano.Wallet.Shelley
     ( SomeNetworkDiscriminant (..) )
 import Cardano.Wallet.Shelley.Compatibility
-    ( HasNetworkId (..), NodeVersionData, emptyGenesis, fromCardanoBlock )
+    ( HasNetworkId (..)
+    , NodeToClientVersionData
+    , emptyGenesis
+    , fromCardanoBlock
+    )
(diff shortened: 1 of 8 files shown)


rvl (approved):
Version negotiation - Excellent, thanks. :+1:

It's only the codecs which need V_9, right? We don't make any queries yet that specifically require V_9? And I suppose that `QueryIfCurrentAlonzo` works under V_8 by never ever running the query.

Anviking:
> We don't make any queries yet that specifically require V_9? And I suppose that QueryIfCurrentAlonzo works under V_8

On Mary, nothing we do should require V_9. Even `QueryAnytimeAlonzo` works (double-checked manually, see commit message).

Frankly, I'm not sure exactly what needs `V_9` on Alonzo. Only that:
- `NodeToClientV_9` has a comment `enabled @CardanoNodeToClientVersion7@, i.e., Alonzo`
- When we start the _integration tests_ in Alonzo, we get `HardForkEncoderDisabledEra (SingleEraInfo {singleEraName = "Alonzo"})` in the node logs (seemingly caused by wallet requests)

So I'd suspect `QueryIfCurrentAlonzo` doesn't work with V_8.

PR #2790, opened 2021-07-30
Temporarily add TestEnableDevelopmentNetworkProtocols to e2e tests

piotr-iohk:
# Issue Number

ADP-1059

# Overview

Temporarily add TestEnableDevelopmentNetworkProtocols to e2e tests

# Comments

To be removed when no longer needed. For now adding for the sake of nightly e2e tests passing...

Running Win/Docker/Mac tests against this branch -> https://github.com/input-output-hk/cardano-wallet/actions
Linux passed locally.

PR #2793, opened 2021-08-01
Fix PatternMatchFail error in shared wallet integration tests

rvl:
### Issue Number

Found during ADP-1059.

### Overview

This error was seen when trying to merge https://github.com/input-output-hk/cardano-wallet/pull/2790#issuecomment-890312237.

The `postSharedWallet` request was failing for some reason, but instead of saying that the request failed, it said:

```
uncaught exception: PatternMatchFail
src/Test/Integration/Scenario/API/Shared/Wallets.hs:1029:25-75: Non-exhaustive patterns in lambda
```

This change lets it print the correct error message depending on whether the request failed or the shared wallet was pending.

### Comments

I'm not sure why the `postSharedWallets` request failed in the first place. Any ideas @paweljakubas ?

### DIFF
--- a/lib/core-integration/cardano-wallet-core-integration.cabal
+++ b/lib/core-integration/cardano-wallet-core-integration.cabal
@@ -49,6 +49,7 @@ library
     , cryptonite
     , deepseq
     , directory
+    , either
     , extra
     , filepath
     , fmt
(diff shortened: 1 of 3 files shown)


paweljakubas (approved):
lgtm! very nice useof view and traverse :+1:

rvl:
The STAKE_POOLS_JOIN_01 flakiness seems to be back, although it never really left.

bors r+

PR #2796, opened 2021-08-02
Revert "Temporarily add TestEnableDevelopmentNetworkProtocols to e2e tests"

Anviking:
ADP-1025

Reverts input-output-hk/cardano-wallet#2790

With #2791, it shouldn't be needed.


piotr-iohk (approved):
thx

PR #2799, opened 2021-08-04
Provide a pure function to select collateral inputs from a UTxO set

jonathanknowles:
# Issue Number

ADP-1035

Co-pilot: @sevanspowell 

# Overview

This PR provides a function `selectCollateral`, that when provided with the following parameters:

 - `coinsAvailable`
    _a set of pure-ada UTxO entries_
 - `maximumSelectionSize`
    _an upper bound on the number of entries that may be selected_
 - `minimumSelectionAmount`
    _a lower bound on the sum of entries selected_
 - `searchSpaceLimit`
    _an upper bound on the search space size_
 
returns a subset of `coinsAvailable` to use as collateral.

This function tries **_two strategies_** in the following order, picking the first strategy that succeeds:

   1. Attempt to select an amount of collateral that is as small as possible.
   2. Attempt to select collateral from the largest coins available.

**The first strategy**, given unlimited computation time, will always produce an optimal result: the smallest possible amount of collateral. However, if the required search space is large, and if the `searchSpaceLimit` parameter is set to a value that's smaller than the required search space size, then this strategy will return without computing a result.

**The second strategy** sacrifices optimality and always produces a result if one is available, by looking only at the very largest coins available. This result can be computed very quickly, without using much search space.

The combination of these two strategies means that we can satisfy the following properties:

If the attempt to select collateral **_succeeds_**, then the function returns a value `coinsSelected` that satisfies the following properties:

```hs
>>> sum  coinsSelected ≥ minimumSelectionAmount
>>> size coinsSelected ≤ maximumSelectionSize
>>>      coinsSelected ⊆ coinsAvailable
```

If the attempt to select collateral **_fails_**, then the function returns a value `largestCombinationAvailable` that satisfies the following properties:

```hs
>>> sum  largestCombinationAvailable < minimumSelectionAmount
>>> size largestCombinationAvailable ≤ maximumSelectionSize
>>>      largestCombinationAvailable ⊆ coinsAvailable              
```


sevanspowell:
I have created a PR to add even more commentary to the collateral tests.

The commentary is already very good, so this PR is really more to clarify my understanding. 

https://github.com/input-output-hk/cardano-wallet/pull/2800

sevanspowell (approved):

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/CollateralSpec.hs
```diff
@@ -0,0 +1,1125 @@
...
+        $ \maximumSelectionSize ->
+    let params = SelectCollateralParams
+            { coinsAvailable
+            , maximumSelectionSize
+            , minimumSelectionAmount
+            , searchSpaceLimit = SearchSpaceLimit 1_000_000
+            } in
+    prop_selectCollateral_common params $ selectCollateralSmallest params
+
+-- In this test, we only consider sets of available coins that when sorted into
```
sevanspowell, line 277: This is great :)

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/CollateralSpec.hs
```diff
@@ -0,0 +1,1125 @@
...
+            property prop_firstRight
+
+    parallel $ describe "takeUntil" $ do
+
+        it "prop_takeUntil_identity" $
+            property prop_takeUntil_identity
+        it "prop_takeUntil_head" $
+            property prop_takeUntil_identity
+        it "prop_takeUntil_takeWhile" $
+            property prop_takeUntil_identity
```
jonathanknowles: This looks like a copy-and-paste error:

```suggestion
        it "prop_takeUntil_identity" $
            property prop_takeUntil_identity
        it "prop_takeUntil_head" $
            property prop_takeUntil_head
        it "prop_takeUntil_takeWhile" $
            property prop_takeUntil_takeWhile
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Collateral.hs
```diff
@@ -0,0 +1,404 @@
...
+--    >>> size coinsSelected ≤ maximumSelectionSize
+--    >>>      coinsSelected ⊆ coinsAvailable
+--
+-- If the attempt to select collateral fails:
+--
+--    >>> sum  largestCombinationAvailable < minimumSelectionAmount
+--    >>> size largestCombinationAvailable ≤ maximumSelectionSize
+--    >>>      largestCombinationAvailable ⊆ coinsAvailable
+--
+selectCollateral :: forall inputId. Ord inputId => SelectCollateral inputId
```
jonathanknowles, line 167: This is marked by `weeder` as a weed. We could solve this by adding a basic test for it.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/CollateralSpec.hs
jonathanknowles: Fixed in f68bf670489512c78dcb69dde731bf86c26177b2.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Collateral.hs
jonathanknowles: Fixed in d88cddfdeb1f4dd88b1cb64a2b7c59303aab0dce.

PR #2800, opened 2021-08-04
Sevanspowell/select collateral commentary

sevanspowell:
# Issue Number

ADP-1035

# Overview

Attempts to add some clarifying commentary to @jonathanknowles selectCollateral work. The commentary is already excellent however, so this PR is really not required.


rvl:
Is this still needed @sevanspowell?

PR #2801, opened 2021-08-04
Bump to cardano-node alonzo-white-1.4

Anviking:
# Issue Number

ADP-1062

# Overview

- [x] Bump to alonzo-white-1.4
- [x] Figure out how to fix the CLI tests
- [x] Fix any newly appearing CI errors 

# Comments

https://github.com/input-output-hk/cardano-haskell/pull/53
https://github.com/input-output-hk/cardano-haskell/pull/54


rvl:
Don't forget to make a PR from your branch on cardano-haskell.
The CI there should hopefully find errors if any.

Anviking:
Hm, hydra was fully green before I bumped optparse-applicative again... 👀 

_Reverting_

lib/test-utils/src/Test/Hspec/Goldens.hs
```diff
@@ -47,8 +62,12 @@ textGolden settings title value = do
             writeGoldenAndFail f value "No existing golden file found"
 
   where
+    -- If running on windows, will try to read the windows-specific golden
+    -- first.
     readGolden :: FilePath -> IO (Either IOError Text)
-    readGolden f = try (TIO.readFile f)
+    readGolden f
+        | os == "mingw32" = try $ TIO.readFile (f <> ".win") <|> TIO.readFile f
```
rvl: There is a helper function `Test.Utils.Platform.isWindows` that you can use here.

The `Alternative` instance of `IO` won't work as we need here.

For example:
```
λ import UnliftIO.Exception 
λ try (throwString "bad" <|> putStrLn "good") :: IO (Either SomeException ())
Left UnliftIO.Exception.throwString called with:

bad
Called from:
 throwString
    ( <interactive>: 8 : 6 in interactive:Ghci4 )
λ 
```

I have pushed a fix.

lib/cli/test/data/Cardano/CLISpec/key from-recovery-phrase --help.win
```diff
@@ -0,0 +1,13 @@
...
+
+Available options:
+  -h,--help                Show this help text
+  STYLE                    Byron | Icarus | Shelley | Shared
+
+The recovery phrase is read from stdin.
+
+Example:
+  [0;1m$ unit.exe recovery-phrase generate \[0m
+  [0;1m| unit.exe key from-recovery-phrase Icarus[0m
```
rvl: Shouldn't these golden files have newlines at the end? Our CLI does print a final newline.

rvl (approved):
LGTM, thanks!

lib/cli/test/data/Cardano/CLISpec/key from-recovery-phrase --help.win
Anviking: I don't think they used to have: https://github.com/input-output-hk/cardano-wallet/blob/e4503c84811713cc66d857a32763e29e2994127a/lib/cli/test/unit/Cardano/CLISpec.hs#L131-L139

Perhaps the _CLI_ would _print_ an extra newline, but the string we get from optparse-applicative in `CLISpec` does not.

lib/test-utils/src/Test/Hspec/Goldens.hs
Anviking: 🤔 
```haskell
λ> try (throwString "bad" <|> putStrLn "good") :: IO (Either SomeException ())
Left UnliftIO.Exception.throwString called with:
bad
Called from:
  throwString (<interactive>:21:6 in interactive:Ghci4)

λ> try (throwIO (userError "bad") <|> putStrLn "good") :: IO (Either SomeException ())
good
Right ()
```

Anviking:
The commit ordering of
1. Re-write CLI tests without changing them (with passing tests)
2. Bump, causing CLI goldens to change
was very intentional (modulo a windows fixup)

I've now rebased.

rvl:
OK fair enough. By ordering commits like that you have shown that the layout from this forked optparse-applicative is not as good as the previous formatting.

I added a fix to make the cli unit tests on windows pass.

lib/test-utils/src/Test/Hspec/Goldens.hs
Anviking: > The Alternative instance of IO won't work as we need here.

I doubt that was a problem though I don't understand the above behaviour, but the other changes in your fix look fair enough.

Anviking:
Thanks
bors r+

rvl:
The cli unit tests were failing on Windows. I have added a fix for that.

lib/cli/test/data/Cardano/CLISpec/key from-recovery-phrase --help.win
rvl: OK. I like newlines at the end of text files, so I added a golden post-processing step to do that.

PR #2804, opened 2021-08-05
jira: fix api token auth

rvl:
### Issue Number

None

### Overview

This change updates the go-jira config and adds a patch to support API keys on iohk.jira.io.

Thanks to @sevanspowell for debugging why API token auth wasn't working.

### Comments

The upstream PR is go-jira/jira#438


sevanspowell (approved):
Oh wow nice!

PR #2814, opened 2021-08-10
Further proposed simplifications for: Identify UTxOs suitable for use as collateral

jonathanknowles:
# Issue Number

#2807 

# Overview

This PR proposes a simplification for the `Primitive.Collateral` module. It:
- introduces `addressSuitableForCollateral`, which is a simple composition of `addressType` and `addressTypeSuitableForCollateral`.
- removes the `classifyCollateralAddress` function, which performed a similar role to `addressSuitableForCollateral`.
- updates property and unit tests appropriately.

PR #2815, opened 2021-08-10
Addendum to Identify UTxOs suitable simplifications

sevanspowell:
Made a few tweaks around @jonathanknowles suggested improvements:
- Took Jonathan's changed property and made a new property, restoring the old one, as they are slightly different.
- Cleared up and updated some of the commentary.

### Issue Number

#2814


jonathanknowles (approved):
LGTM!

PR #2811, opened 2021-08-09
Bump to cardano-node alonzo-purple-1.0.1

Anviking:
# Issue Number

ADP-1066

# Overview

- [x] I have bumped to cardano-node alonzo-purple-1.0.1

# Comments

https://github.com/input-output-hk/cardano-haskell/pull/55

PR #2816, opened 2021-08-10
Bump version to v2021-08-11

Anviking:
- [x] Bump wallet version to v2021-08-10 in preparation for release

### Issue Number

Release.

### Comments

Dependent on #2811

### DIFF
--- a/scripts/make_release.sh
+++ b/scripts/make_release.sh
@@ -19,10 +19,9 @@ set -euo pipefail
 ################################################################################
 # Release-specific parameters (Change when you bump the version)
 # Release tags must follow format vYYYY-MM-DD.
-GIT_TAG="v2021-07-30"
-OLD_GIT_TAG="v2021-06-11"
-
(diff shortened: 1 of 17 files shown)


scripts/make_release.sh
```diff
@@ -19,10 +19,9 @@ set -euo pipefail
 ################################################################################
 # Release-specific parameters (Change when you bump the version)
 # Release tags must follow format vYYYY-MM-DD.
-GIT_TAG="v2021-07-30"
-OLD_GIT_TAG="v2021-06-11"
-
-CARDANO_NODE_TAG="1.28.0"
+GIT_TAG="v2021-08-10"
```
rvl: Could we make it v2021-08-11, because we will probably only fully finish it tomorrow.

rvl (approved):

scripts/make_release.sh
Anviking: 👍

PR #2817, opened 2021-08-11
Add collateral inputs to wallet types, DB, and API.

jonathanknowles:
### Issue Number

ADP-967 / ADP-1069 / ADP-1039

### Overview

This PR adds all the **_boilerplate_** required for collateral inputs to:
- the wallet database;
- the wallet primitive types;
- the wallet API.

This PR makes **_no attempt_** to manipulate collateral inputs in any way, other than to marshall data around from one layer to another. It also makes **_no attempt_** to parse collateral inputs from network blocks. Where necessary, future work has been marked with `TODO` comments. (With a link to ADP-957).

Currently, all API calls returning collateral inputs will return the **_empty list_**.

## API

This PR adds API entries and types for collateral inputs, with the following differences from ordinary inputs:
- while a transaction must have at least one ordinary input, it's possible for a transaction to have **no** collateral inputs.
- while an ordinary input may be associated with one or more non-ada assets, a collateral input **may not** have non-ada assets.

This is reflected in the following differences between types:
```diff
- x-transactionInputs: &transactionInputs
+ x-transactionCollateral: &transactionCollateral
    description: |
-     A list of transaction inputs.
+     A list of transaction inputs used for collateral.
    type: array
    minItems: 0
    items:
      type: object
      required:
        - id
        - index
      properties:
        address: *addressId
        amount: *amount
-       assets: *walletAssets
        id: *transactionId
        index:
          type: integer
          minimum: 0
```

```diff
- x-transactionResolvedInputs: &transactionResolvedInputs
+ x-transactionResolvedCollateral: &transactionResolvedCollateral
-   description: A list of transaction inputs.
+   description: A list of transaction inputs used for collateral.
    type: array
-   minItems: 1
+   minItems: 0
    items:
      type: object
      required:
        - id
        - index
        - address
        - amount
        - derivation_path
      properties:
        address: *addressId
        amount: *amount
-       assets: *walletAssets
        id: *transactionId
        derivation_path: *derivationPath
        index:
          type: integer
          minimum: 0
```

### Database

This PR adds a `TxCollateral` table, with analagous fields to the `TxIn` table.

### Primitive Types

This PR adds fields for collateral inputs to the following primitive types:
  - `UnsignedTx`
  - `TransactionInfo`
  - `Tx`


lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1920,18 +1924,38 @@ mkApiTransactionFromInfo
...
             Expired  -> #pendingSince
-    return $ case meta ^. #status of
+    return $ case txInfoMeta ^. #status of
         Pending  -> apiTx
-        InLedger -> apiTx { depth = Just depth  }
+        InLedger -> apiTx {depth = Just txInfoDepth}
         Expired  -> apiTx
   where
-      drop2nd (a,_,c) = (a,c)
+    TransactionInfo
```
jonathanknowles: Perhaps we could just use `#view` to refer to each field instead of record wild-card syntax.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -2828,6 +2867,16 @@ instance EncodeAddress n => ToJSON (ApiTxInput n) where
...
 
+instance DecodeAddress n => FromJSON (ApiTxCollateral n) where
+    parseJSON v = ApiTxCollateral <$> optional (parseJSON v) <*> parseJSON v
+
+instance EncodeAddress n => ToJSON (ApiTxCollateral n) where
+    toJSON (ApiTxCollateral s i) =
+        Object (maybe mempty (fromValue . toJSON) s <> fromValue (toJSON i))
+      where
+        fromValue (Object o) = o
+        fromValue _ = mempty
```
jonathanknowles, line 2878: These instances are basically copies of the `ApiTxInput` instances. Perhaps we can simplify them, and use a more generic deriving style.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -2165,13 +2219,24 @@ selectTxs = fmap concatUnzip . mapM select . chunksOf chunkSize
...
         resolvedInputs <- fmap toOutputMap $
             combineChunked inputs $ \inputsChunk ->
                 traverse readTxOutTokens . fmap entityVal =<<
                     selectList
                         [TxOutputTxId <-. (txInputSourceTxId <$> inputsChunk)]
                         [Asc TxOutputTxId, Asc TxOutputIndex]
 
+        resolvedCollateralInputs <- fmap toOutputMap $
+            combineChunked collateralInputs $ \collateralInputsChunk ->
+                traverse readTxOutTokens . fmap entityVal =<< selectList
```
jonathanknowles, line 2235: Note that we don't actually expect to read any tokens for collateral inputs.

We should consider simplifying this logic.

specifications/api/swagger.yaml
```diff
@@ -1212,6 +1230,27 @@ x-transactionResolvedInputs: &transactionResolvedInputs
         type: integer
         minimum: 0
 
+x-transactionResolvedCollateral: &transactionResolvedCollateral
+  description: A list of transaction inputs used for collateral
```
sevanspowell, line 1234: What does "resolved" here mean?

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -2182,7 +2247,10 @@ selectTxs = fmap concatUnzip . mapM select . chunksOf chunkSize
             []
 
         pure
-            ( inputs `resolveWith` resolvedInputs
+            ( inputs
+                `resolveInputWith`resolvedInputs
```
sevanspowell:
```suggestion
                `resolveInputWith` resolvedInputs
```

jonathanknowles: Fixed in fd0c08234f222349de62cedcf8a66f841af46c60.

lib/core/src/Cardano/Wallet/Api/Server.hs
jonathanknowles: Fixed in b91d00ecde1bbc4bcbeedecf3c0534cf0b0ebf0d.

sevanspowell:
The subtleties are lost on me, but I've read through this carefully and it looks logical and consistent.

By the subtleties, I just mean I'm not sure I can answer the questions of:
- Is the API in the schema we want?
- Is the DB in the schema we want?

specifications/api/swagger.yaml
jonathanknowles: > What does "resolved" here mean?

Good question. I'm following the existing terminology for inputs, as much as possible.

For "resolved inputs": in addition to including the `txId` and `txIndex` (which is enough to uniquely identify an input), we also include the following information, which shows what this input "resolves", or "maps" to:
- the `address`
- the `derivationPath`
- the `amount`

jonathanknowles:
> The subtleties are lost on me, but I've read through this carefully and it looks logical and consistent.

Thanks for looking through!

> By the subtleties, I just mean I'm not sure I can answer the questions of:
> 
>     * Is the API in the schema we want?
>     * Is the DB in the schema we want?

I think the main principle here is that "collateral inputs" should have a similar representation to ordinary inputs, in both the API and the DB.

The primary differences (at the type-level) are:
1. Collateral inputs cannot have non-ada asset quantities.
2. In general, a transaction must have at least one normal input, but it doesn't have to have collateral inputs.

So in the API, we have the following differences:
```diff
- x-transactionResolvedInputs: &transactionResolvedInputs
+ x-transactionResolvedCollateral: &transactionResolvedCollateral
-   description: A list of transaction inputs
+   description: A list of transaction inputs used for collateral
    type: array
-   minItems: 1
+   minItems: 0
    items:
      type: object
      required:
        - id
        - index
        - address
        - amount
        - derivation_path
      properties:
        address: *addressId
        amount: *amount
-       assets: *walletAssets
        id: *transactionId
        derivation_path: *derivationPath
        index:
          type: integer
          minimum: 0
```
```diff
- x-transactionInputs: &transactionInputs
+ x-transactionCollateral: &transactionCollateral
    description: |
-     A list of transaction inputs.
+     A list of transaction inputs used for collateral.
    type: array
    minItems: 0
    items:
      type: object
      required:
        - id
        - index
      properties:
        address: *addressId
        amount: *amount
-       assets: *walletAssets
        id: *transactionId
        index:
          type: integer
          minimum: 0
```

jonathanknowles:
> Is the DB in the schema we want?

The schema for `TxCollateral` should be the same as the schema for `TxIn`, just with different field name prefixes.

You might have wondered whether `TxIn` would include an extra `assets` field, but assets are actually stored separately (in another table), because they're in a many-to-one relationship with each UTxO entry. So as a result, the `TxCollateral` table is the same shape (has the same fields) as the `TxIn` table.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2856,7 +2856,7 @@ mkApiTransaction
     -> Lens' (ApiTransaction n) (Maybe ApiBlockReference)
     -> IO (ApiTransaction n)
 mkApiTransaction
-    ti txid mfee ins cins outs ws (meta, timestamp) txMeta setTimeReference = do
+    ti txid mfee cins ins outs ws (meta, timestamp) txMeta setTimeReference = do
```
sevanspowell: Ooh, this looks like a very likely place to make a mistake, is there anyway we can distinguish between the two types of [(TxIn, Maybe TxOut)] that makes sense in this context?

sevanspowell (approved):

sevanspowell:
> > Is the DB in the schema we want?
> 
> The schema for `TxCollateral` should be the same as the schema for `TxIn`, just with different field name prefixes.
> 
> You might have wondered whether `TxIn` would include an extra `assets` field, but assets are actually stored separately (in another table), because they're in a many-to-one relationship with each UTxO entry. So as a result, the `TxCollateral` table is the same shape (has the same fields) as the `TxIn` table.

Thanks for this! I've re-reviewed and added my :heavy_check_mark:

sevanspowell (approved):

lib/core/src/Cardano/Wallet/Api/Server.hs
jonathanknowles: @sevanspowell I agree that it would be easy for a caller to get these arguments muddled up, especially given that the types are identical.

I've created a record type `MkApiTransactionParams`, and altered `mkApiTransaction` to take this record instead of all the positional parameters. Hopefully this should make it easier at call sites to determine which parameter is which.

Commit: c9bc8e687068eb4bd0e556596266d22644ae0833

PR #2820, opened 2021-08-12
Make scripts/gh/show-bump.sh run in nix

Anviking:
- [x] Make scripts/gh/show-bump.sh run in nix-shell, such that you don't need to worry about installing dependencies manually.

### Issue Number

None / Release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -85,11 +85,13 @@ jobs:
         run: |
           gem install bump-cli
           sudo snap install yq
-          sudo apt-get -yu install html-xml-utils
       - name: 'Update Release in Bump.sh'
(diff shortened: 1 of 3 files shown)


Anviking:
Not sure why it fails with
```
In ./scripts/gh/show-bump.sh line 1:
--
  | #! /usr/bin/env nix-shell
  | ^-- SC1008: This shebang was unrecognized. ShellCheck only supports sh/bash/dash/ksh. Add a 'shell' directive to specify.
```
while nix-shell is fine in `make_release.sh`

rvl:
If not using a standard shell script interpreter like `#!/bin/sh`, then shellcheck needs to be told which kind of shell with a directive. I fixed it.

rvl (approved):
Thanks

PR #2825, opened 2021-08-16
Adjust chain-following code to account for collateral inputs

sevanspowell:
## Overview 

- Adjusted chain-following code to account for collateral inputs

### Issue Number

ADP-1036


sevanspowell:
I think I want to change the type of `isValidScript` from `Maybe Bool` to a more descriptive type, but I want to do that after I establish some better tests.

PR #2819, opened 2021-08-11
Adjust available UTxO to account for collateral inputs

jonathanknowles:
### Issue Number

ADP-1038

### Comments

This PR:
- [x] adjusts the `availableUTxO` function to account for collateral inputs.
- [x] adds property tests for `availableUTxO`.
- [x] adds a property test relating `availableUTxO` to `availableBalance`.

PR #2768, opened 2021-07-20
Simplify QC generators for commonly-used types.

rvl:
# Issue Number

#2695

# Overview

This PR simplifies the generators (and shrinkers) for the following types:
- [x] `Coin`
- [x] `TokenQuantity`
- [x] `TokenPolicyId`
- [x] `TokenName`
- [x] `AssetId`
    
Each type now has a single generator and shrinker that relies on the **_QC size parameter_**. These have the following form:
- `genType`
- `shrinkType`

Where **_strictly-positive_** values are required, we also have:
- `genTypePositive`
- `shrinkTypePositive`

Where values across the **_full range_** of a type are required, we also have:
- `genTypeFullRange`
- `shrinkTypeFullRange`

This PR also adjusts coverage checks for some properties.

# Future work

In a future PR, we can simplify generators for `TokenMap`, `TokenBundle`, `UTxOIndex`, `TxIn`, `TxOut`.

PR #2829, opened 2021-08-17
Rewrite generators to use the QC size parameter

jonathanknowles:
### Issue Number

Follow-up from #2768
Supports work for #2819 

### Comments

This PR adjusts generators for the following types to use the QC size parameter:
- `Address`
- `Hash "Tx"`
- `TxIndex` (`Word32`)
- `TxIn`
- `TxOut`
- `UTxO`
- `UTxOIndex`

Some minor adjustments to coverage conditions were necessary (but surprisingly few).

This PR also:
- Adds the module `UTxO.Gen`, so that we can generate values of `UTxO` without having to generate indices (which add extra overhead), and redefines the `UTxOIndex` generators in terms of these generators.
- Adds the functions `genSized2` and `genSized2With`, which restore size linearity to generators of compound values (those defined in terms of other generators).

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap/Gen.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap/Gen.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE NamedFieldPuns #-}
-{-# LANGUAGE TypeApplications #-}
 
 module Cardano.Wallet.Primitive.Types.TokenMap.Gen
     ( genAssetId
--- a/lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap/Gen.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap/Gen.hs
@@ -43,13 +43,12 @@ import Test.QuickCheck
     , choose
     , functionMap
     , oneof
-    , resize
     , shrinkList
     , sized
     , variant
     )
 import Test.QuickCheck.Extra
-    ( shrinkInterleaved )
+    ( genSized2With, shrinkInterleaved )
 
 import qualified Cardano.Wallet.Primitive.Types.TokenMap as TokenMap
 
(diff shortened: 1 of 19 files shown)


lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap/Gen.hs
```diff
@@ -58,22 +57,7 @@ import qualified Cardano.Wallet.Primitive.Types.TokenMap as TokenMap
...
-    -- /quadratically/ with the size parameter, which is /not/ what we want.
-    --
-    -- Therefore, we pass each individual generator a size parameter that
-    -- is the square root of the original.
-    --
-    let sizeSquareRoot = max 1 $ ceiling $ sqrt $ fromIntegral @Int @Double size
-    AssetId
-        <$> resize sizeSquareRoot genTokenPolicyId
-        <*> resize sizeSquareRoot genTokenName
+genAssetId = genSized2With AssetId genTokenPolicyId genTokenName
```
sevanspowell, line 59: Ha nice!

sevanspowell (approved):

lib/core/src/Cardano/Wallet/Primitive/Types/TokenMap/Gen.hs
jonathanknowles: Thanks!

If we ever have to make a generator for values composed of more than two parts, we can easily make a similar `genSized3`/`genSizedWith3` function.

Though I'm hoping we won't need such a thing. 😄

lib/test-utils/src/Test/QuickCheck/Extra.hs
```diff
@@ -38,6 +40,44 @@ import Test.QuickCheck
...
+-- If the current value of the size parameter is 1000, then to generate a range
+-- of up to 1000 different composite values, we can resize each individual
+-- generator so that it generates up to 10 different values:
+--
+-- >>> genComposite = Composite
+-- >>>     <$> scaleToRoot 3 genA
+-- >>>     <*> scaleToRoot 3 genB
+-- >>>     <*> scaleToRoot 3 genC
+--
+scaleToRoot :: Int -> Gen a -> Gen a
```
rvl, line 63: This is a nice idea which might be needed _very occasionally_, but in general, I don't believe we need to fret too much over the `size` parameter. It's not meant to be so precise - for example, `instance Arbitrary (a,b,c,d,e)` in QuickCheck doesn't do any such scaling. From the developer perspective, it's fairly obvious that if you have a big ADT then its total size and range will be bigger than that of a small ADT.

PR #2810, opened 2021-08-09
Buildkite fixes

rvl:
### Issue Number

Found during ADP-1062.

### Overview

- Improve display of section headers in logs
- Don't run integration tests on every PR (caused by #2789)

PR #2832, opened 2021-08-18
Fix usage of --skip in buildkite builds

Anviking:
- [x] Call `--skip "integration"` instead of `--skip cardano-wallet:integration`. The latter is ineffective.
- [x] Confirm that the behaviour in CI indeed looks ok.
    - [x] Integration tests are skipped in normal builds
    - [x] Integration tests in Mary are run from bors
    - [x] Integration tests in Alonzo are run from bors

### Issue Number

None. Follow-up to #2810 .


Anviking:
bors try

rvl (approved):
Thanks for fixing!

PR #2835, opened 2021-08-20
update e2e test assertions so it is in-line with updated entries in testnet metadata registry

piotr-iohk:
- 7b41900fe712bf65916bd83d77471e9036e4fa4b
  update e2e test assertions so it is in-line with updated entries in testnet metadata registry

### Issue Number

n/a


rvl (approved):
Thanks

PR #2836, opened 2021-08-23
Provide property tests for `changeUTxO`

jonathanknowles:
### Issue Number

ADP-1084

### Comments

This PR provides property tests for the `changeUTxO` function.

Since this function is re-used by `totalUTxO` (which we want to modify and then verify), it's useful to first verify that `changeUTxO` is correct.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/Types/UTxO.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types/UTxO.hs
@@ -29,12 +29,15 @@ module Cardano.Wallet.Primitive.Types.UTxO
     , computeUtxoStatistics
     , excluding
     , isSubsetOf
+    , null
     , log10
     , restrictedBy
     , restrictedTo
+    , size
     ) where
 
-import Prelude
+import Prelude hiding
+    ( null )
 
 import Cardano.Wallet.Primitive.Types.Coin
     ( Coin (..) )
@@ -126,6 +129,12 @@ restrictedBy (UTxO utxo) =
 restrictedTo :: UTxO -> Set TxOut -> UTxO
 restrictedTo (UTxO utxo) outs =
     UTxO $ Map.filter (`Set.member` outs) utxo
+
+null :: UTxO -> Bool
+null (UTxO u) = Map.null u
+
+size :: UTxO -> Int
+size (UTxO u) = Map.size u
 
 data UTxOStatistics = UTxOStatistics
     { histogram :: ![HistogramBar]
--- a/lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
@@ -453,24 +453,23 @@ prop_availableUTxO makeProperty =
 --    - odd-parity addresses:
 --      addresses with a pop count (Hamming weight) that is odd.
 --
--- With the above definitions in mind, this test uses the following filter
--- conditions to categorize output addresses:
+-- The choice to categorize addresses by parity has two advantages:
 --
---    - 'Even' : matches addresses whose parity is even
---    - 'Odd'  : matches addresses whose parity is odd
---    - 'All'  : matches all addresses, regardless of their parity.
+--    - The parity of an address can be determined by a pure function.
 --
--- Each of the above filter conditions is represented with a custom 'IsOurs'
--- instance that we provide to the 'changeUTxO' function.
+--    - Addresses with even parity and odd parity are equally frequent.
+--
+-- A custom 'IsOurs' instance is used to categorize output addresses according
+-- to their parity.
 --
 -- Since a given address can either be even, or odd, but not both, it follows
 -- that the following results must be disjoint:
 --
---    - 'changeUTxO pendingTxs Even'
---    - 'changeUTxO pendingTxs Odd'
+--    - 'changeUTxO pendingTxs (IsOursIf (== Even) . addressParity)'
+--    - 'changeUTxO pendingTxs (IsOursIf (==  Odd) . addressParity)'
 --
--- This test applies each of the above filter conditions to 'changeUTxO' and
--- verifies that all entries match the condition that was provided.
+-- This test applies the above filter conditions to 'changeUTxO' and verifies
+-- that all entries match the condition that was provided.
 --
 prop_changeUTxO :: Property
 prop_changeUTxO =
@@ -500,15 +499,15 @@ prop_changeUTxO =
 
         -- The UTxO set that contains all available output addresses.
         utxoAll :: UTxO
-        utxoAll = changeUTxO txSet All
+        utxoAll = changeUTxO txSet $ IsOursIf @Address (const True)
 
         -- The UTxO set that contains only even-parity output addresses.
         utxoEven :: UTxO
-        utxoEven = changeUTxO txSet Even
+        utxoEven = changeUTxO txSet $ IsOursIf ((== Even) . addressParity)
 
         -- The UTxO set that contains only odd-parity output addresses.
         utxoOdd :: UTxO
-        utxoOdd  = changeUTxO txSet Odd
+        utxoOdd  = changeUTxO txSet $ IsOursIf ((== Odd) . addressParity)
 
         txSet :: Set Tx
         txSet = Set.fromList $ uncurry makeTx <$> (zip txIds txOutputs)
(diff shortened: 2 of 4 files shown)


lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -425,6 +438,202 @@ prop_availableUTxO makeProperty =
...
+--
+-- This test applies each of the above filter conditions to 'changeUTxO' and
+-- verifies that all entries match the condition that was provided.
+--
+prop_changeUTxO :: Property
+prop_changeUTxO =
+    forAllShrink (listOf genTxOutputs) (shrinkList shrinkTxOutputs) inner
+  where
+    inner :: [TxOutputs] -> Property
+    inner txOutputs =
```
rvl:
```suggestion
newtype Classify a = Classify (a -> Bool)

prop_changeUTxO :: Classify Address -> [TxOut] -> Property
prop_changeUTxO (Classify condition) txOutputs =
```

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -425,6 +438,202 @@ prop_availableUTxO makeProperty =
...
+        , withdrawals = Map.empty
+        }
+
+    -- Creates a dummy transaction identifier from a single word.
+    makeTxId :: Word8 -> Hash "Tx"
+    makeTxId i = Hash $ BS.pack [i]
+
+-- | Represents all the outputs of a transaction (both change and non-change).
+--
+newtype TxOutputs = TxOutputs
```
rvl, line 540: I don't get why we go to the trouble of making a `newtype`, but don't add an `Arbitrary` instance, and then use `shrinkAll`, when a simple `[TxOut]` parameter on the property would do the same thing.

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -425,6 +438,202 @@ prop_availableUTxO makeProperty =
...
+genTxOutputs :: Gen TxOutputs
+genTxOutputs = TxOutputs <$> listOf genTxOut
+
+shrinkTxOutputs :: TxOutputs -> [TxOutputs]
+shrinkTxOutputs (TxOutputs outs) = TxOutputs <$> shrinkList shrinkTxOut outs
+
+-- | An implementation of 'IsOurs' that matches all addresses.
+--
+instance IsOurs All Address where
+    isOurs = isOursIf (const True)
```
jonathanknowles: **Suggestion**: replace each of the type class instances with a single instance that encapsulates a match condition.

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -425,6 +438,202 @@ prop_availableUTxO makeProperty =
...
+--    - 0b00000001 : odd  (Hamming weight = 1)
+--    - 0b00000010 : odd  (Hamming weight = 1)
+--    - 0b00000011 : even (Hamming weight = 2)
+--    - 0b00000100 : odd  (Hamming weight = 1)
+--    - ...
+--    - 0b11111110 : odd  (Hamming weight = 7)
+--    - 0b11111111 : even (Hamming weight = 8)
+--
+addressParity :: Address -> Parity
+addressParity = parity . addressPopCount
```
rvl, line 583: Any other `Address -> Bool` function would be just as good, as long it splits the set of generated addresses into roughly equal parts.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxO.hs
```diff
@@ -127,6 +130,12 @@ restrictedTo :: UTxO -> Set TxOut -> UTxO
 restrictedTo (UTxO utxo) outs =
     UTxO $ Map.filter (`Set.member` outs) utxo
 
+null :: UTxO -> Bool
+null (UTxO u) = Map.null u
```
rvl, line 134: Really? The only place this is used it's with `(not (UTxO.null x)`. But `x /= mempty` would be fine.

rvl (approved):

sevanspowell:
I really love the way you've set out the comments, makes it easy to understand your logic and reasoning :+1: I'm trying to emulate your style in my own work.

My first thought was that `changeUTxO` is such a simple function and I wasn't sure how your new tests actually tested it, but now I understand that though the function appears simple, the existence of `State s`, and the way the behaviour of the function could arbitrarily change based on `s` and the `IsOurs` instance, makes more involved testing necessary.

I find this interesting because I'm working in a very similar place at the moment and I've made some modifications, let me know what you think:

- Using a type class is equivalent to implicitly passing the typeclass members to the function.

- Therefore we can re-write `changeUTxO` as:
    
    ```
    changeUTxO
    :: (Address -> s -> (Maybe (NonEmpty DerivationIndex), s))
    -> Set Tx
    -> s
    -> UTxO
    changeUTxO isOurs pending = evalState $
        mconcat <$> mapM (state . utxoOurs isOurs) (Set.toList pending)
    ```

   (also making the relevant change to utxoOurs and so on...)

- But `utxoOurs`, and by extension `changeUTxO` uses `isOurs :: Address -> s -> (Maybe (NonEmpty DerivationIndex), s)` purely as a `isOurs :: Address -> s -> (Bool, s)` function. You confirm this with your `isOursIf` function, which essentially lifts `Address -> Bool` to a "proper" `isOurs` function (in a non-total manner). i.e. we don't care about the result
- A function can easily be written to demote `isOurs :: Address -> s -> (Maybe DerivationIndex, s)` to `isOurs :: Address -> s -> (Bool, s)`.
- and `entity -> s -> (result, s)` is a special case of `entity -> f result`, where `f` is `s -> (s,)` (a functor)
- So we can further re-write `changeUTxO` as:

    ```
    changeUTxO
    :: (Address -> f Bool)
    -> Set Tx
    -> f UTxO
    changeUTxO isOurs pending = ...
    ```
    
    while retaining the original behaviour, although how we call changeUTxO will change (i.e. `evalState $ changeUTxO isOursBool`)
    
- For the purposes of testing, we can now do a couple of things:
    1. Offload the burden-of-proof that the `Address -> f Bool`, and the `f` functor, are working correctly off to the writer of those functions.
        - That's still "us", but maybe testing our demoted `isOursBool` function in isolation is a little bit easier.
    2. Re-write the tests of this function to not use State at all (e.g. use Identity functor for `f`). We're making it more explicit that this function can only rely on specific properties of `f` (e.g. Functor, Applicative, Monad).

- Maybe we can even write a property that asserts the equivalence between `changeUTxO` and some other similar exisiting function like `filterM`:
      ∀f set. toList (changeUTxO f set) = filterM f (toList set)

I haven't tested this out so your mileage may vary. One thing to note is that we can "not use typeclasses" in `changeUTxO`, but continue using typeclasses everywhere else, it's an incremental change rather than a "don't ever use IsOurs", although "some" people "out there" do have an aversion to using typeclases that don't have laws (i.e. Functor typeclass has laws, but not IsOurs). That's a bit irrelevant to all this though.

TLDR; the type signature of `changeUTxO` can be changed to more accurately reflect it's usage, and so shift responsibility of some of the more complicated parts off to another module's tests, which /might/ be worth it.

Otherwise looks really good! This is just food for thought and I'm happy to approve and merge as-is.

sevanspowell (approved):

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
jonathanknowles: @rvl Nice suggestion!

I've actually made a similar change in 0ffa43da16b9232aafcf613453709a279545cc74, which replaces the various types with a single `IsOursIf {condition :: a -> Bool}` newtype (with a single instance).

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
jonathanknowles: Yes, definitely.

It's an arbitrary choice (excuse the pun!) to use this particular function. It has two nice advantages:

1. Addresses are "even" or "odd" with equal frequency (meaning we don't have to do anything special with generation).
2. Address parity can be determined with a pure function (meaning we don't have to keep any state around in our `IsOurs` instance).

I've added a comment to `prop_changeUTxO` with these reasons.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxO.hs
jonathanknowles: > Really? The only place this is used it's with `(not (UTxO.null x)`. But `x /= mempty` would be fine.

I agree with you that `x /= mempty` would work fine!

Though I find that `UTxO.null` is a little more self-documenting (in this case), as:

- A reader doesn't have to navigate to the definition to figure out the type of thing being tested, because it's stated right at the point of use; and
- The `null` function is idiomatic across different container types, so it's familiar.

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
jonathanknowles: Fixed in 0ffa43da16b9232aafcf613453709a279545cc74.

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
jonathanknowles: @rvl Fair enough!

I've added an `Arbitrary` instance for `TxOutputs` in 761ecc21a740e95ef5367a9a41919edfbbf7763a.

PR #2838, opened 2021-08-23
Adjust the `totalUTxO` function to not exclude collateral inputs.

jonathanknowles:
### Issue Number

ADP-1084

### Comments

This PR adjusts the `totalUTxO` function so that collateral inputs are not excluded.


sevanspowell (approved):
Looks good to me!

PR #2841, opened 2021-08-24
Library `dbvar` for delta encodings and mutable `DBVar` variable

HeinrichApfelmus:
### Issue number

ADP-1100, ADP-1102, ADP-1111, ADP-1112, ADP-1134

### Overview

This is the initial implementation of a library `dbvar` about delta encodings.

The Cardano Wallet backend becomes faster if we store only the deltas between checkpoints of the wallet state in the database. The wallet state is essentially a `List` of known addresses and a `Set` of UTxOs. The library implemented here can handle this use case. In addition, the use of side effects in the wallet can be significantly reduced with the new `DBVar` abstraction.

The library provides

* A mutable variable type `DBVar` which represents values that are kept in memory, but which are written to the hard drive on every update.
* Delta encodings and combinators on them (`Delta` type class, `Embedding` type).
* An in-memory type for database tables and delta encodings for it.
* A `Chain` data type which can store checkpoints of the wallet state when combined with delta encodings

### Progress

* The API and implementation are now stable enough to be merged into the repository and be a dependency of other pull requests.
* The API is still preliminary and will see some streamlining when we use it in anger.
* The modules `Database.Schema` and `Demo.Database` are still experimental.
* Property tests are not provided yet. We will rely on existing unit tests of the wallet while integrating, up until the point where the database format is changed.

### Comments

* As this library is of general utility, I would like to release it on hackage eventually.
    * Before that, I would like to reorganize it slightly and rename it to `delta-encodings`.


HeinrichApfelmus:
As discussed today, I would like to start adding this library into our codebase, so that the database layer can be refactored and eventually switched to the delta encoding of the wallet state checkpoints. The API abstractions are stable, but I intend to postpone property tests until just before the switch in order to accommodate minor API changes that arise during refactoring. With that in mind, reviews welcome! 😊

lib/dbvar/README.md
```diff
@@ -0,0 +1,41 @@
...
+The main modules are:
+
+* [Data.Delta](src/Data/Delta.hs) — delta encodings
+    * Contains the `Delta` class. Each member type `da` represents
+      a delta encoding for a base type `Base da`.
+      A few basic delta encodings are implemented.
+    * Contains the `Embedding` type which represents an embedding
+      of one type and its delta encoding into another type.
+      `Embedding` can be composed and form a category, but for technical
+      reasons, they only implement the `Semigroupoid` class.
```
paweljakubas: could you specify why "only" semigroupoid?

lib/dbvar/README.md
```diff
@@ -0,0 +1,41 @@
...
+* [Data.DBVar](src/Data/DBVar.hs) — delta encodings
+    * Contains the `DBVar` mutable variable abstraction.
+      The intention is that these variables are stored in-memory
```
paweljakubas: yeah, having distinction between in-memory and drive persistence is absolutely ok for now.
But what about introducing some time in the future Memory abstraction. And here could be anything: in-memory, redis, kafka, ....., disk. And DBVar and Storage could assume to be attached any instance of Memory...

lib/dbvar/src/Data/Chain.hs
```diff
@@ -0,0 +1,301 @@
...
+import Data.List
+    ( unfoldr )
+import Data.Map.Strict
+    ( Map )
+import Data.Semigroupoid
+    ( o )
+import Data.Table
+    ( DeltaTable (..), Pile (..), Table )
+
+import qualified Data.Map as Map
```
paweljakubas:
```suggestion
import qualified Data.Map.Strict as Map
```

lib/dbvar/src/Data/DBVar.hs
```diff
@@ -0,0 +1,331 @@
...
+        > updateS s a da  ≠  writeS s (apply a da)
+
+* __Exceptions__:
+    It is expected that the functions 'loadS', 'updateS', 'writeS'
+    do not throw synchronous exceptions. In the worst case,
+    'loadS' should return 'Nothing' after reading or writing
+    to the store was unsuccesful.
+-}
+data Store m da = Store
+    { loadS   :: m (Maybe (Base da))
```
paweljakubas: I think it would be better to have `Either` here. If someone wants Maybe then one can use for example https://hackage.haskell.org/package/either-5.0.1.1/docs/Data-Either-Combinators.html#v:leftToMaybe

lib/dbvar/src/Demo/Database.hs
```diff
@@ -0,0 +1,180 @@
...
+
+addressSqlIso :: Iso' (Edge Node AddressInPool) AddressRow
+addressSqlIso = iso ab ba
+  where
+    ab Edge{from,to,via=AddressInPool{address,index}} =
+        Table :. Col from :. Col to :. Col 0 :. Col address :. Col index
+    ba (Table :. Col from :. Col to :. Col _ :. Col address :. Col index) =
+        Edge{from,to,via=AddressInPool{address,index}}
+
+type AddressRow = Table "addresses"
```
paweljakubas, line 187: very fond of this way of expressing "location/path/..." ala  `servant` way

lib/dbvar/src/Demo/Database.hs
```diff
@@ -0,0 +1,180 @@
...
+    embedStore embed =<< newSqlStore
+  where
+    embed = embedIso addressSqlIso
+        `o` (tableIntoDatabase `o` chainIntoTable Pile getPile)
+
+addressSqlIso :: Iso' (Edge Node AddressInPool) AddressRow
+addressSqlIso = iso ab ba
+  where
+    ab Edge{from,to,via=AddressInPool{address,index}} =
+        Table :. Col from :. Col to :. Col 0 :. Col address :. Col index
```
paweljakubas, line 183: :+1:

lib/dbvar/src/Database/Schema.hs
```diff
@@ -0,0 +1,293 @@
...
+runSql :: MonadIO m => Query () -> SqlPersistT m ()
+runSql Query{stmt,params} = Persist.rawExecute stmt params
+
+{-------------------------------------------------------------------------------
+    SQL queries
+-------------------------------------------------------------------------------}
+-- | An SQL query that returns a list of values of type @row@.
+data Query row = Query
+    { stmt :: Text
+    -- ^ SQL statement containing placeholders \"?\" which are
```
paweljakubas, line 192: :+1:

lib/dbvar/src/Data/Table.hs
```diff
@@ -0,0 +1,255 @@
...
+
+{-------------------------------------------------------------------------------
+    Pile
+-------------------------------------------------------------------------------}
+-- | A 'Pile' is a set of values.
+-- Unlike 'Set', it is represented as a list, and avoids the 'Ord' constraint.
+--
+-- This type is useful for highlighting that a collection of values
+-- has no specific order, even though it is not represented as a 'Set'.
+newtype Pile a = Pile { getPile :: [a] }
```
paweljakubas, line 189: why Set is not ok?

lib/dbvar/src/Demo/Database.hs
HeinrichApfelmus: Me too. 😊

lib/dbvar/src/Data/Table.hs
HeinrichApfelmus: The main motivation for the `Pile` type is to slightly improve the types of database results. In many database libraries, a `SELECT` operation returns a list of rows, but this list is *unordered*, i.e. we cannot rely on any particular order of its elements. Nobody seems to complain that e.g. `Database.Persist.Class.selectList` does not return a `Set`, but a list… 😅  I figured that a `SELECT` operation should at least return a `Pile`.

The two advantages of `Pile` over `Set` are

* It does not require an `Ord` constraint on the elements as `Set` does.
* Conversion to and from a list is faster than `Set`, because there are no balancing operations.

There is an argument to be made that `Pile` provides a good representation of the notion of a "set" for types which do not have good normal forms, and therefore no good implementation for `Ord`. For example, you can put infinite lists into a `Pile`, but you may not want to put them into a `Set`. 🤔

lib/dbvar/README.md
HeinrichApfelmus: I have updated the description. The issue is that

```
id :: Embedding da da
```

is difficult to write. The notion of "embedding" only applies to delta encodings of types, so we need a class constraint 

```
id :: Delta da => Embedding da da
```

An `id` function with this class constraint is no longer part of the `Category` class.
There are ways around that, but they have other drawbacks.

lib/dbvar/README.md
HeinrichApfelmus: I have clarified the wording a little bit, so that it becomes (hopefully) more clear that `DBVar` is intended to store a Haskell value in the sense that this value is available in a representation that is native to the runtime / language.

I think the `Store` type itself might already be a good memory abstraction — we can `loadS` to a native Haskell value (including appropriate marshaling) and write back. 🤔

More general stores may require a slightly more general abstraction. In the most general case, the way to modify values would become

`modifyVar :: Var da -> (Fetcher a (da,b) -> IO b)`

Here, `Fetcher a` is similar to a reader monad that can retrieve "portions" of a value of type `a`. This `Fetcher` monad-think would behave very similar to a lens. Anyway, I decided to exclude partial loading of values for this project.

lib/dbvar/src/Data/DBVar.hs
HeinrichApfelmus: Hm. This seemingly minor change has larger ramifications. In particular, if `Store` has an additional type parameter `e` that keeps track of potential error messages, then `Embedding` should have one, too.

But then — which of the following type signatures should we give `embedStore`?

```
embedStore_1 :: (…) => Embedding da db -> Store m db -> Store m da
embedStore_2 :: (…) => Embedding da db -> Store e m db -> Store e m da
embedStore_3 :: (…) => Embedding e da db -> Store e m db -> Store e m da
embedStore_4 :: (…) => Embedding e1 da db -> Store e2 m db -> Store (e1 :+: e2) m da
```

Taking the union of the error messages, solution No. 4, seems like the most general choice, but this solution would require tools for managing unions of error types, which we currently do not employ.

Another solution would be to use the most general union of error types, `SomeException`, and use the type signature

```
loadS5 :: Store m da -> m (Either SomeException (Base da))
```

In this way, the error is not visible in the type of `Store`, but at least the error message is not lost. I think that I prefer this approach, actually.

Yet another option would be to use the underlying monad `m` to signal a failure case. But that would be unsuitable for `Embedding`.

On that note, I should go back and make sure that the `DBVar` are capable of handling asynchronous exceptions. This may not be the case as I originally hoped for. 🤔

lib/dbvar/README.md
paweljakubas: ok, fair point

lib/dbvar/README.md
paweljakubas: thanks for the clarification

lib/dbvar/src/Data/Table.hs
paweljakubas: sounds reasonable :+1:

paweljakubas (approved):
Very nice abstraction introduced here! Especially mesmerized by `servant` like adoption to get resources. I am in favor of merging this as it is, and then add in separate PRs testing for this (an interesting instances of concepts, checking properties of many algebra laws that should be obeyed, etc). I bet some abstractions could be polished, comments improved as those tests and practical realizations at work (in tests) are tried. Also, some cleaning of commit history, squashing would do no harm!

HeinrichApfelmus:
Thanks for taking the time to review this, Pawel! 😊 I'm also in favor of merging this as it is, as I agree that writing tests and actually using this library will help the most with polishing the abstractions and comments.

I'll wrap this up by

* fixing robustness in the presence of asynchronous exceptions
* replacing `Maybe` with `Either SomeException` in the `load`* functions
* squashing the commit history a bit

and then merging the pull request. Thanks!

I'm also fond of the `servant`-like database row types, though they will probably the last feature to be integrated into the wallet code, as Persistent does a good-enough job for the time being.

HeinrichApfelmus:
bors merge

PR #2843, opened 2021-08-25
Add functions `{gen,shrink}Tx` to `Primitive.Tx.Gen`

jonathanknowles:
### Issue Number

ADP-1084

### Comments

This PR adds a reusable generator and shrinker for transactions (of type `Tx`) to the `Primitive.Tx.Gen` module.

It uses these functions to simplify or eliminate `Arbitrary` instances in the following modules:
- `Cardano.Wallet.Primitive.ModelSpec`
- `Cardano.Wallet.Primitive.TypesSpec`
- `Cardano.WalletSpec`

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/Types/Tx/Gen.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types/Tx/Gen.hs
@@ -1,11 +1,17 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE PartialTypeSignatures #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# OPTIONS_GHC -fno-warn-partial-type-signatures #-}
 
 module Cardano.Wallet.Primitive.Types.Tx.Gen
-    ( genTxHash
+    ( genTx
+    , genTxHash
     , genTxIndex
     , genTxIn
     , genTxInLargeRange
     , genTxOut
+    , shrinkTx
     , shrinkTxHash
     , shrinkTxIndex
     , shrinkTxIn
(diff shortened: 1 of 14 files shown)


lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -390,41 +366,22 @@ prop_availableUTxO
 prop_availableUTxO makeProperty =
     forAllShrink (scale (* 4) genUTxO) shrinkUTxO
         $ \utxo ->
-    forAllShrink (listOf genTxInputs) (shrinkList shrinkTxInputs)
-        $ \pendingTxInputs ->
-    inner utxo pendingTxInputs
+    forAllShrink (scale (`div` 2) $ listOf genTx) (shrinkList shrinkTx)
```
sevanspowell, line 369: Why do we `scale (div 2)` here?

sevanspowell (approved):

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
sevanspowell: Like I think I understand the size paramater, but I'm not sure why we need to divide it by two here.

jonathanknowles: By using `scale (div 2)`, we halve the value of the size parameter passed to all the inner generators of `genTx`, meaning that average size of transactions will be halved (even though the actual size may vary).

In testing, I found that this didn't affect the coverage conditions we care about, while making the tests run faster.

jonathanknowles: In general, users of `genTx` can choose the size of transactions to generate by manipulating the size parameter in this way.

It's the same principle for most of the other generators that we provide in `Gen` modules. (Since most of these now generate "sized" values, according to the size parameter.)

If you find that you can scale down the size parameter **_and_** keep the coverage you want (which should be explicitly tested), then I think it makes sense to do so, as we can reduce the overall test execution time.

jonathanknowles: We could also have used `scale (mod n)` here, for some value of `n`, which would place a **_fixed_** upper limit on the size of transactions (while still allowing the size to vary).

lib/core/src/Cardano/Wallet/Primitive/Types/Tx/Gen.hs
```diff
@@ -15,35 +21,117 @@ module Cardano.Wallet.Primitive.Types.Tx.Gen
...
+        (shrinkList (liftShrink2 shrinkTxIn shrinkCoinPositive))
+        (shrinkList (liftShrink2 shrinkTxIn shrinkCoinPositive))
+        (shrinkList shrinkTxOut)
+        (liftShrink shrinkTxMetadata)
+        (shrinkMapWith shrinkRewardAccount shrinkCoinPositive)
+
+txWithoutIdToTx :: TxWithoutId -> Tx
+txWithoutIdToTx t@TxWithoutId {..} = Tx {..}
+  where
+    txId = mockHash $ txWithoutIdToTuple t
```
jonathanknowles, line 122: We probably don't need to convert to a tuple here, as `TxWithoutId` already has a `Show` instance:
```suggestion
    txId = mockHash t
```

Will fix this in follow-up PR.

jonathanknowles:
bors cancel

jonathanknowles:
bors cancel

PR #2848, opened 2021-08-26
Move important UTxO state transition functions to the top-level and test them

sevanspowell:
This work is intended to allow future modifications to the UTxO state transition function (i.e. collateral) to be tested in isolation, at the unit level.

My changes are primarily to the nested function "applyTx". I have tried to keep my changes minimal:
  - Use more verbose variable names in that function to make the purpose of each variable clear.
  - More clearly delineate between transaction outputs that are unspent and transaction outputs "we know about" but may possibly be spent.
  - Make some effort to separate the ideas of "knowledge of a UTxO" from "ownership of a UTxO" to make testing easier.
  - Extract out the core state transition function from the existing "do lots" "applyTx" function and move it to the top-level.
  - Move other useful functions to the top-level for the same reason.
  - Test important functions in wallet model spec.
  - Re-implement hasKnownInput/hasKnownOutput to make their implementations a little more clear and symmetric.

### Comments

These changes make testing collateral much easier, we can just test the "applyTxToUTxO" function at the unit level, and adjust existing model tests as necessary.

### Issue Number

ADP-1092

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/Model.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Model.hs
@@ -482,14 +482,9 @@ prefilterBlock b u0 = runState $ do
         -> Tx
         -> State s ([(Tx, TxMeta)], UTxO)
     applyTx (!txs, !prevUTxO) tx = do
-        -- All transaction outputs we know about, that belong to us, and that
-        -- haven't been validated as "unspent" yet. It is important to not
-        -- remove spent transactions, because if we do, we lose information
-        -- about transactions inputs and outputs we know about.
-        knownTxO <- (prevUTxO <>) <$> filterOurUTxOs isOurs' (utxoFromTx tx)
         -- The next UTxO state (apply a state transition) (e.g. remove
         -- transaction outputs we've spent).
-        ourNextUTxO <- filterOurUTxOs isOurs' (applyTxToUTxO tx knownTxO)
+        ourNextUTxO <- filterOurUTxOs isOurs' (applyTxToUTxO tx prevUTxO)
 
         ourWithdrawals <- Coin . sum . fmap (unCoin . snd) <$>
             mapMaybeM ourWithdrawal (Map.toList $ withdrawals tx)
@@ -499,6 +494,11 @@ prefilterBlock b u0 = runState $ do
                 balance (prevUTxO `difference` ourNextUTxO)
                 `TB.add` TB.fromCoin ourWithdrawals
 
+        -- All transaction outputs we know about, that belong to us, and that
+        -- haven't been validated as "unspent" yet. It is important to not
+        -- remove spent transactions, because if we do, we lose information
+        -- about transactions inputs and outputs we know about.
+        knownTxO <- (prevUTxO <>) <$> filterOurUTxOs isOurs' (utxoFromTx tx)
         -- A transaction has a known input if one of the transaction inputs
         -- matches a transaction input we know about.
         let hasKnownInput =
--- a/lib/core/src/Cardano/Wallet/Primitive/Model.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Model.hs
@@ -39,6 +39,10 @@ module Cardano.Wallet.Primitive.Model
     , applyBlock
     , applyBlocks
     , unsafeInitWallet
+    , applyTxToUTxO
+    , utxoFromTx
+    , filterOurUTxOs
+    , difference
 
     -- * Accessors
     , currentTip
@@ -80,15 +84,15 @@ import Cardano.Wallet.Primitive.Types.Tx
     , txOutCoin
     )
 import Cardano.Wallet.Primitive.Types.UTxO
-    ( Dom (..), UTxO (..), balance, excluding, restrictedBy )
+    ( Dom (..), UTxO (..), balance, excluding, restrictedTo )
 import Control.DeepSeq
     ( NFData (..), deepseq )
 import Control.Monad
-    ( foldM, forM )
+    ( foldM )
 import Control.Monad.Extra
     ( mapMaybeM )
 import Control.Monad.Trans.State.Strict
-    ( State, evalState, runState, state )
+    ( State, StateT, evalState, runState, state )
 import Data.Functor
     ( (<&>) )
 import Data.Generics.Internal.VL.Lens
@@ -98,7 +102,7 @@ import Data.Generics.Labels
 import Data.List.NonEmpty
     ( NonEmpty (..) )
 import Data.Maybe
-    ( catMaybes, isJust )
+    ( isJust )
 import Data.Set
     ( Set )
 import Fmt
@@ -320,6 +324,97 @@ totalUTxO
 totalUTxO pending wallet@(Wallet _ _ s) =
     availableUTxO pending wallet <> changeUTxO pending s
 
+-- | Applies a transaction to a UTxO, moving it from one state from another.
+-- When applying a transaction to a UTxO:
+--   1. We need to remove any unspents that have been spent in the transaction.
+--   2. Add any unspents that we've received via the transaction.
+--
+-- We don't consider "ownership" here (is this address ours?), only "do we
+-- know about this address" (i.e. is it present in our UTxO?).
+--
+-- balance (applyTxToUTxO tx u) = balance u
+--                              + balance (utxoFromTx tx)
+--                              - balance (u `restrictedBy` inputs tx)
+-- unUTxO (applyTxToUTxO tx u) = unUTxO u
+--   `Map.union` unUTxO (utxoFromTx tx)
+--   `Map.difference` unUTxO (u `restrictedBy` inputs tx)
+applyTxToUTxO
+    :: Tx
+    -> UTxO
+    -> UTxO
+applyTxToUTxO tx !u = do
+    let
+        -- When a transaction comes in, we now know about a set of new unspents.
+        newKnown = u <> utxoFromTx tx
+        -- But we also know that some of the new unspents we know about, as well
+        -- as (potentially) some old ones, may now be spent.
+        spentKnown =
+            Set.fromList (inputs tx) `Set.intersection` dom newKnown
+
+    newKnown `excluding` spentKnown
+
+-- | Limit a UTxO set to just the UTxOs that are ours, according to some
+-- given function.
+--
+-- filterOurUTxOs (const $ pure True) u = u
+-- filterOurUTxOs (const $ pure False) u = mempty
+-- filterOurUTxOs f mempty = mempty
+-- balance (filterOurUTxOs f (applyTxToUTxO tx mempty)) =
+--   foldMap (\o -> do
+--               ours <- f (address o)
+--               if ours then tokens o else mempty
+--            ) (outputs tx)
+filterOurUTxOs :: forall f. Monad f => (Address -> f Bool) -> UTxO -> f UTxO
+filterOurUTxOs isOursF (UTxO m) =
+    UTxO <$> Map.traverseMaybeWithKey filterFunc m
+    where
+        filterFunc :: TxIn -> TxOut -> f (Maybe TxOut)
+        filterFunc _txin txout = do
+            ours <- isOursF $ txout ^. #address
+            pure $ if ours then Just txout else Nothing
+
+-- | Construct a UTxO corresponding to a given transaction. It is important for
+-- the transaction outputs to be ordered correctly, since they become available
+-- inputs for the subsequent blocks.
+--
+-- balance (utxoFromTx tx) = foldMap tokens (outputs tx)
+utxoFromTx :: Tx -> UTxO
+utxoFromTx Tx {txId, outputs} =
+    UTxO $ Map.fromList $ zip (TxIn txId <$> [0..]) outputs
+
+isOurs' :: forall s m. (Monad m, IsOurs s Address) => Address -> StateT s m Bool
+isOurs' addr = do
+    mDerivationIndex <- state $ isOurs addr
+    case mDerivationIndex of
+        Nothing -> pure False
+        Just _  -> pure True
+
+-- | Get the elements in u1 that are not in u2. In the case that elements are in
+-- both, get the difference of the value of the TxOut (the TokenBundle value) in
+-- both entries, removing any entries that are fully spent.
+--
+-- u `difference` mempty = u
+-- mempty `difference` u = mempty
+-- u `difference` u = mempty
+difference :: UTxO -> UTxO -> UTxO
+difference u1 u2 =
+    let
+        u1' = unUTxO u1
+        u2' = unUTxO u2
+
+        diffFunc :: TxOut -> TxOut -> Maybe TxOut
+        diffFunc a b =
+            let
+                tokens1 = tokens a
+                tokens2 = tokens b
+                diff = tokens1 `TB.difference` tokens2
+            in
+                if diff == mempty
+                then Nothing
+                else Just $ TxOut (address a) diff
+    in
+        UTxO $ Map.differenceWith diffFunc u1' u2'
+
 {-------------------------------------------------------------------------------
                                Internals
 -------------------------------------------------------------------------------}
@@ -386,16 +481,34 @@ prefilterBlock b u0 = runState $ do
         => ([(Tx, TxMeta)], UTxO)
         -> Tx
         -> State s ([(Tx, TxMeta)], UTxO)
-    applyTx (!txs, !u) tx = do
-        ourU <- state $ utxoOurs tx
-        let ourIns = Set.fromList (inputs tx) `Set.intersection` dom (u <> ourU)
-        let u' = (u <> ourU) `excluding` ourIns
+    applyTx (!txs, !prevUTxO) tx = do
+        -- All transaction outputs we know about, that belong to us, and that
+        -- haven't been validated as "unspent" yet. It is important to not
+        -- remove spent transactions, because if we do, we lose information
+        -- about transactions inputs and outputs we know about.
+        knownTxO <- (prevUTxO <>) <$> filterOurUTxOs isOurs' (utxoFromTx tx)
+        -- The next UTxO state (apply a state transition) (e.g. remove
+        -- transaction outputs we've spent).
+        ourNextUTxO <- filterOurUTxOs isOurs' (applyTxToUTxO tx knownTxO)
+
         ourWithdrawals <- Coin . sum . fmap (unCoin . snd) <$>
             mapMaybeM ourWithdrawal (Map.toList $ withdrawals tx)
-        let received = balance ourU
-        let spent = balance (u `restrictedBy` ourIns) `TB.add` TB.fromCoin ourWithdrawals
-        let hasKnownInput = ourIns /= mempty
-        let hasKnownOutput = ourU /= mempty
+
+        let received = balance (ourNextUTxO `difference` prevUTxO)
+        let spent =
+                balance (prevUTxO `difference` ourNextUTxO)
+                `TB.add` TB.fromCoin ourWithdrawals
+
+        -- A transaction has a known input if one of the transaction inputs
+        -- matches a transaction input we know about.
+        let hasKnownInput =
+                Set.fromList (inputs tx) `Set.intersection` dom knownTxO
+                /= mempty
+        -- A transaction has a known output if one of the outputs exists in the
+        -- set of unspent transaction outputs we know about.
+        let hasKnownOutput =
+                knownTxO `restrictedTo` Set.fromList (outputs tx)
+                /= mempty
         let hasKnownWithdrawal = ourWithdrawals /= mempty
 
         -- NOTE 1: The only case where fees can be 'Nothing' is when dealing with
@@ -423,7 +536,7 @@ prefilterBlock b u0 = runState $ do
         return $ if hasKnownOutput && not hasKnownInput then
             let dir = Incoming in
             ( (tx { fee = actualFee dir }, mkTxMeta (TB.getCoin received) dir) : txs
-            , u'
+            , ourNextUTxO
             )
         else if hasKnownInput || hasKnownWithdrawal then
             let
@@ -433,10 +546,10 @@ prefilterBlock b u0 = runState $ do
                 amount = distance adaSpent adaReceived
             in
                 ( (tx { fee = actualFee dir }, mkTxMeta amount dir) : txs
-                , u'
+                , ourNextUTxO
                 )
         else
-            (txs, u)
+            (txs, prevUTxO)
 
 -- | Get the change UTxO
 --
@@ -451,21 +564,4 @@ changeUTxO
     -> s
     -> UTxO
 changeUTxO pending = evalState $
-    mconcat <$> mapM (state . utxoOurs) (Set.toList pending)
-
--- | Construct our _next_ UTxO (possible empty) from a transaction by selecting
--- outputs that are ours. It is important for the transaction outputs to be
--- ordered correctly, since they become available inputs for the subsequent
--- blocks.
-utxoOurs
-    :: IsOurs s Address
-    => Tx
-    -> s
-    -> (UTxO, s)
-utxoOurs tx = runState $ toUtxo <$> forM (zip [0..] (outputs tx)) filterOut
-  where
-    toUtxo = UTxO . Map.fromList . catMaybes
-    filterOut (ix, out) = do
-        state (isOurs $ address out) <&> \case
-            Just{}  -> Just (TxIn (txId tx) ix, out)
-            Nothing -> Nothing
+    mconcat <$> mapM (filterOurUTxOs isOurs' . utxoFromTx) (Set.toList pending)
--- a/lib/core/src/Cardano/Wallet/Primitive/Model.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Model.hs
@@ -84,7 +84,7 @@ import Cardano.Wallet.Primitive.Types.Tx
     , txOutCoin
     )
 import Cardano.Wallet.Primitive.Types.UTxO
-    ( Dom (..), UTxO (..), balance, excluding, restrictedTo )
+    ( Dom (..), UTxO (..), balance, excluding )
 import Control.DeepSeq
     ( NFData (..), deepseq )
 import Control.Monad
@@ -494,22 +494,30 @@ prefilterBlock b u0 = runState $ do
                 balance (prevUTxO `difference` ourNextUTxO)
                 `TB.add` TB.fromCoin ourWithdrawals
 
-        -- All transaction outputs we know about, that belong to us, and that
-        -- haven't been validated as "unspent" yet. It is important to not
-        -- remove spent transactions, because if we do, we lose information
-        -- about transactions inputs and outputs we know about.
         (ownedAndKnownTxIns, ownedAndKnownTxOuts) <- do
-            u <- (prevUTxO <>) <$> filterOurUTxOs isOurs' (utxoFromTx tx)
-            let map = unUTxO u
-            pure (Map.keys map, Map.elems map)
+            -- A new transaction expands the set of transaction inputs/outputs
+            -- we know about:
+            let known = prevUTxO <> utxoFromTx tx
+            -- But not all those transaction inputs/outputs belong to us:
+            ownedAndKnown <- filterOurUTxOs isOurs' known
+            -- Also, the new transaction may spend some transaction
+            -- inputs/outputs. But we don't want to apply that logic yet. If we
+            -- do, any spent transaction input/output will be removed from our
+            -- knowledge base.
+            -- Therefore, because this is not technically an "Unspent TxO" set,
+            -- let's just return the TxIns and TxOuts, as the type "UTxO" will
+            -- create expectations which we explicitly aren't fulfilling:
+            let m = unUTxO ownedAndKnown
+            pure (Map.keys m, Map.elems m)
+
         -- A transaction has a known input if one of the transaction inputs
         -- matches a transaction input we know about.
         let hasKnownInput =
                 Set.fromList (inputs tx)
                     `Set.intersection` (Set.fromList ownedAndKnownTxIns)
                 /= mempty
-        -- A transaction has a known output if one of the outputs exists in the
-        -- set of unspent transaction outputs we know about.
+        -- A transaction has a known output if one of the transaction outputs
+        -- matches a transaction output we know about.
         let hasKnownOutput =
                 Set.fromList (outputs tx)
                     `Set.intersection` (Set.fromList ownedAndKnownTxOuts)
--- a/lib/core/src/Cardano/Wallet/Primitive/Model.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Model.hs
@@ -41,6 +41,7 @@ module Cardano.Wallet.Primitive.Model
     , unsafeInitWallet
     , applyTxToUTxO
     , utxoFromTx
+    , spendTx
 
     -- * Accessors
     , currentTip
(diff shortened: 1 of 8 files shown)


lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -320,6 +324,97 @@ totalUTxO
...
+
+-- | Get the elements in u1 that are not in u2. In the case that elements are in
+-- both, get the difference of the value of the TxOut (the TokenBundle value) in
+-- both entries, removing any entries that are fully spent.
+--
+-- u `difference` mempty = u
+-- mempty `difference` u = mempty
+-- u `difference` u = mempty
+difference :: UTxO -> UTxO -> UTxO
+difference u1 u2 =
```
jonathanknowles: **Suggestion**: move this `difference` function to the `UTxO` module.

**Justification**: it's a function that operates purely on `UTxO` values, and could viewed as analogous to these difference functions:

* `TokenBundle.difference`
* `TokenMap.difference`
* `Set.difference`
* `Map.difference`
* ...

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -320,6 +324,97 @@ totalUTxO
...
+--               if ours then tokens o else mempty
+--            ) (outputs tx)
+filterOurUTxOs :: forall f. Monad f => (Address -> f Bool) -> UTxO -> f UTxO
+filterOurUTxOs isOursF (UTxO m) =
+    UTxO <$> Map.traverseMaybeWithKey filterFunc m
+    where
+        filterFunc :: TxIn -> TxOut -> f (Maybe TxOut)
+        filterFunc _txin txout = do
+            ours <- isOursF $ txout ^. #address
+            pure $ if ours then Just txout else Nothing
```
jonathanknowles: **Suggestion**: Move this function to the `UTxO` module, and perhaps rename it to `filterByAddress` or `filterByAddressM`.

**Justification**: Although this is used in a context where we're filtering for "our" addresses, in principle it could operate with **_any_** monadic indicator function on addresses. Giving it a more general name will match the nice generalization that you've made in the type signature.

jonathanknowles: You could even have a pair of functions:
- `filterByAddressM` (your monadic function above)
- `filterByAddress` (a wrapper around `filterAddressM` that uses the `Identity` monad).

(See here for another example of this pattern: https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/Primitive/CoinSelection/MA/RoundRobin.hs#L1847-L1859)

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -320,6 +324,97 @@ totalUTxO
...
+-- filterOurUTxOs (const $ pure False) u = mempty
+-- filterOurUTxOs f mempty = mempty
+-- balance (filterOurUTxOs f (applyTxToUTxO tx mempty)) =
+--   foldMap (\o -> do
+--               ours <- f (address o)
+--               if ours then tokens o else mempty
+--            ) (outputs tx)
+filterOurUTxOs :: forall f. Monad f => (Address -> f Bool) -> UTxO -> f UTxO
+filterOurUTxOs isOursF (UTxO m) =
+    UTxO <$> Map.traverseMaybeWithKey filterFunc m
```
jonathanknowles: 👍🏻

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -386,16 +481,34 @@ prefilterBlock b u0 = runState $ do
...
-    applyTx (!txs, !u) tx = do
-        ourU <- state $ utxoOurs tx
-        let ourIns = Set.fromList (inputs tx) `Set.intersection` dom (u <> ourU)
-        let u' = (u <> ourU) `excluding` ourIns
+    applyTx (!txs, !prevUTxO) tx = do
+        -- All transaction outputs we know about, that belong to us, and that
+        -- haven't been validated as "unspent" yet. It is important to not
+        -- remove spent transactions, because if we do, we lose information
+        -- about transactions inputs and outputs we know about.
+        knownTxO <- (prevUTxO <>) <$> filterOurUTxOs isOurs' (utxoFromTx tx)
```
jonathanknowles: Perhaps a typo?

```suggestion
        knownUTxO <- (prevUTxO <>) <$> filterOurUTxOs isOurs' (utxoFromTx tx)
```

lib/core/src/Cardano/Wallet/Primitive/Model.hs
sevanspowell: Absolutely :+1:

lib/core/src/Cardano/Wallet/Primitive/Model.hs
sevanspowell: Yeah, I think that's a good idea. Makes testing easier too, as more often than not the tester will just write the test in terms of some Identity-like monad.

lib/core/src/Cardano/Wallet/Primitive/Model.hs
sevanspowell: Not quite. Perhaps you can give me some guidance on this?

My original thought process was that UTxO = **unspent** transaction output, but (something I found quite confusing about this function) is that in this case, `knownTxO` represents transaction outputs we know about and *does not* exclude spent transactions, so it is **not** a UTxO.

Why don't we exclude spent transactions? Because if we do, we lose information about which transactions we "know" about, and so can't calculate "hasKnownInput" and "hasKnownOutput" accurately.

I have created a commit to take a step towards clearing this up.

I have some ideas to clear this up further.

sevanspowell: 740c71e

sevanspowell: 05d18c6 & 46abf39

lib/core/src/Cardano/Wallet/Primitive/Model.hs
jonathanknowles: @sevanspowell and I had a discussion about the `difference` function.

We believe it might make more sense if this function were to have the following signature:
```patch
- difference :: UTxO -> UTxO     -> UTxO
+ difference :: UTxO -> Set TxIn -> UTxO
```
Reasons:
- The current implementation relies on finding the difference between two `TxOut` values in the event that a key appears in both UTxO sets.
- But it's not clear what useful answer we can obtain from differencing a pair of `TxOut` values. For example, if the pair of `TxOut` values have different addresses, then which address should we pick, and why?
- In all situations that we currently deal with, there should only be one possible `TxOut` for a given `TxIn`. (We shouldn't ever have a state where there are two possible `TxOut` values for a given `TxIn`.)
- If we assume that there is only one unique mapping for any given `TxIn`, it makes sense to constrain the type so that we're only subtracting a set of `TxIn` values (so the function does not need to deal with the odd case where `TxOut` values differ) from the key set of our `UTxO` parameter.

However, after making this simplification, we can see there is already a pre-existing function with these semantics:
```hs
excluding :: UTxO -> Set TxIn ->  UTxO
excluding (UTxO utxo) =
    UTxO . Map.withoutKeys utxo
```

Hence, in all places that currently use `UTxO.difference`, it might make more sense to call `UTxO.excluding` with just the set of `TxIn` values from the second map.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxO.hs
```diff
@@ -136,6 +144,29 @@ null (UTxO u) = Map.null u
 size :: UTxO -> Int
 size (UTxO u) = Map.size u
 
+-- | Limit a UTxO set to just the UTxOs that are ours, according to some
+-- given function.
```
jonathanknowles: Suggestion: make the comment more general, to match the generality of the function:

```suggestion
-- | Filter a 'UTxO' set with an indicator function on 'Address' values.
--
-- Returns the subset of UTxO entries that have addresses for which the given
-- indicator function returns 'True'.
```

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -386,16 +432,47 @@ prefilterBlock b u0 = runState $ do
...
         -> Tx
         -> State s ([(Tx, TxMeta)], UTxO)
-    applyTx (!txs, !u) tx = do
-        ourU <- state $ utxoOurs tx
-        let ourIns = Set.fromList (inputs tx) `Set.intersection` dom (u <> ourU)
-        let u' = (u <> ourU) `excluding` ourIns
+    applyTx (!txs, !prevUTxO) tx = do
+        -- The next UTxO state (apply a state transition) (e.g. remove
+        -- transaction outputs we've spent).
+        ourNextUTxO <- filterByAddressM isOurs' (applyTxToUTxO tx prevUTxO)
```
jonathanknowles: I suspect this change might alter the performance characteristics of the `applyTx` function.

Before this change:
- We only apply the `isOurs` filter to the outputs of the transaction we're considering.
- We assume that all entries in the current UTxO set already belong to the wallet, so they don't need to be re-filtered.

After this change:
- We first add all transaction outputs to the UTxO set, regardless of whether they belong to us.
- We then have no option but to apply the `isOurs` filter to all entries in the resulting UTxO set, possibly re-filtering entries that have already been filtered before.

I'm wondering, is there a way to rewrite this section so that we don't need to filter the whole UTxO set on each call to `applyTx`?

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -1407,3 +1431,66 @@ blockchain =
...
+        === unUTxO u
+          `Map.union`
+              unUTxO (utxoFromTx tx)
+          `Map.difference`
+              unUTxO (u `UTxO.restrictedBy` Set.fromList (inputs tx))
+    
+prop_filterByAddress_allOurs :: Property
+prop_filterByAddress_allOurs =
+    forAllShrink genUTxO shrinkUTxO $ \u ->
+        filterByAddress (const True) u === u
```
jonathanknowles: 👍🏻

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -1407,3 +1431,66 @@ blockchain =
...
+
+prop_filterByAddress_noneOurs :: Property
+prop_filterByAddress_noneOurs =
+    forAllShrink genUTxO shrinkUTxO $ \u ->
+        filterByAddress (const False) u === mempty
```
jonathanknowles: 👍🏻

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -1407,3 +1431,66 @@ blockchain =
...
+                    if f (address o) then tokens o else mempty
+                ) (outputs tx)
+
+prop_filterByAddress_filterByAddressM :: Bool -> Property
+prop_filterByAddress_filterByAddressM b =
+    let
+        f = const b
+    in
+        forAllShrink genUTxO shrinkUTxO $ \u ->
+            filterByAddress f u === runIdentity (filterByAddressM (pure . f) u)
```
jonathanknowles: Is it possible to add a property that uses a real filtering function?

There's already a convenient helper function available we can use here: `addressParity`.

You could choose either of the following:
```hs
(== Even) . addressParity
(==  Odd) . addressParity
```
Either of the above functions will select around half of all addresses.

Then you could verify that all addresses in the resulting UTxO set really do match what you expect.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxO.hs
jonathanknowles: Addressed in https://github.com/input-output-hk/cardano-wallet/pull/2860/commits/7ce039d642e309f206084b012a8b39ee351796db (#2860).

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
jonathanknowles: Addressed in https://github.com/input-output-hk/cardano-wallet/pull/2860/commits/d275095bc273a15bf21f117564e8ef80adc8d0bb (#2860).

lib/core/src/Cardano/Wallet/Primitive/Model.hs
sevanspowell: Given that `applyTxToUTxO tx u = (u <> utxoFromTx tx) "excluding" inputs tx`, we can work algebraically and re-arrange to: `u "excluding" inputs tx <> utxoFromTx tx`, provided that the law `utxoFromTx tx "excluding" inputs tx === utxoFromTx tx` holds. We definitely want to test it holds, because if it doesn't, it's saying that the UTxO set we produce from the transaction isn't in-fact "unspent" at all!

Re-arranging then helps us to redefine `applyTx` as: `u "excluding" inputs tx <> filterByAdressM isOurs' (utxoFromTx tx)`, which is equivalent to the current implementation, under the assumption that `u` only contains transactions that belong to us.

So if my logic is correct, yes I can re-define it to not filter the whole UTxO set.

sevanspowell: `applyTxToUTxO` is perhaps too course a primitive. I could define a finer primitive `spendTx`, and assert that:
`applyTxToUTxO = spendTx tx u <> utxoFromTx tx = spendTx tx (u <> utxoFromTx tx)`
Then `filterByAddressM isOurs' (applyTxToUTxO tx prevUTxO)`, could be reduced:
```
filterByAddressM isOurs' (applyTxToUTxO tx u)
filterByAddressM isOurs' (spendTx tx u <> utxoFromTx tx) [equivalence mentioned above]
spendTx tx u <> filterByAddressM isOurs' (utxoFromTx tx) [we'd need a new law...]

New law: spendTx tx (filterByAddress f u) = filterByAddress f (spendTx tx u)
```

In my mind, that new law is solid. Spending a tx doesn't change any address properties which could affect a filter. Regardless, the law will constrain the possible implementations.

sevanspowell: This is something Sandy talks about in "Algebra-Driven Design". Defining a program in terms of it's laws and algebraic properties helps us change performance characteristics while maintaining equivalence.

sevanspowell: 7717363e7b27c0181e94c183720e0d709b522451

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -320,6 +322,53 @@ totalUTxO
...
+-- inputs for the subsequent blocks.
+--
+-- balance (utxoFromTx tx) = foldMap tokens (outputs tx)
+-- utxoFromTx tx `excluding` Set.fromList (inputs tx) = utxoFrom tx
+utxoFromTx :: Tx -> UTxO
+utxoFromTx Tx {txId, outputs} =
+    UTxO $ Map.fromList $ zip (TxIn txId <$> [0..]) outputs
+
+isOurs' :: forall s m. (Monad m, IsOurs s Address) => Address -> StateT s m Bool
+isOurs' addr = do
```
jonathanknowles: Minor suggestion: we could consider calling this `isOurAddress`, as it's morally equivalent to a specialization of `isOurs` to the `Address` type. (This would make the name a bit more self-documenting, at call sites.)

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -320,6 +322,53 @@ totalUTxO
...
+    mDerivationIndex <- state $ isOurs addr
+    case mDerivationIndex of
+        Nothing -> pure False
+        Just _  -> pure True
```
jonathanknowles:
```suggestion
isOurs' = fmap isJust . state . isOurs
```

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -386,16 +435,48 @@ prefilterBlock b u0 = runState $ do
...
+        let spent =
+                balance (prevUTxO `excluding` dom ourNextUTxO)
+                `TB.add` TB.fromCoin ourWithdrawals
+
+        (ownedAndKnownTxIns, ownedAndKnownTxOuts) <- do
+            -- A new transaction expands the set of transaction inputs/outputs
+            -- we know about:
+            let known = prevUTxO <> utxoFromTx tx
+            -- But not all those transaction inputs/outputs belong to us:
+            ownedAndKnown <- filterByAddressM isOurs' known
```
jonathanknowles: Hi Sam. I suspect that evaluating this expression will still re-filter the wallet's whole UTxO, since:
```hs
   filterByAddressM isOurs' known
== filterByAddressM isOurs' (prevUTxO <> utxoFromTx tx)
```
I'm wondering, is there any way we can avoid refiltering parts of the UTxO set that have been filtered before?

Perhaps:
```suggestion
            ownedAndKnown <- (prevUTxO <>)
                <$> filterByAddressM isOurs' (utxoFromTx tx)
```

(Apologies if I've misunderstood something!)

lib/core/src/Cardano/Wallet/Primitive/Model.hs
sevanspowell: Done (locally).

lib/core/src/Cardano/Wallet/Primitive/Model.hs
sevanspowell: True! Yep, we can just do what you suggested, I will fix this locally.

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -386,16 +435,48 @@ prefilterBlock b u0 = runState $ do
...
+            -- create expectations which we explicitly aren't fulfilling:
+            let m = unUTxO ownedAndKnown
+            pure (Map.keys m, Map.elems m)
+
+        -- A transaction has a known input if one of the transaction inputs
+        -- matches a transaction input we know about.
+        let hasKnownInput =
+                Set.fromList (inputs tx)
+                    `Set.intersection` (Set.fromList ownedAndKnownTxIns)
+                /= mempty
```
jonathanknowles: Suggestion: use `Set.disjoint`:

```suggestion
        let hasKnownInput = not $ Set.disjoint
                (Set.fromList $ inputs tx)
                (Set.fromList ownedAndKnownTxIns)
```

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -386,16 +435,48 @@ prefilterBlock b u0 = runState $ do
...
+        -- A transaction has a known output if one of the transaction outputs
+        -- matches a transaction output we know about.
+        let hasKnownOutput =
+                Set.fromList (outputs tx)
+                    `Set.intersection` (Set.fromList ownedAndKnownTxOuts)
+                /= mempty
```
jonathanknowles: Suggestion: use `Set.disjoint`:

```suggestion
        let hasKnownOutput = not $ Set.disjoint
                (Set.fromList $ outputs tx)
                (Set.fromList ownedAndKnownTxOuts)
```

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -451,21 +532,4 @@ changeUTxO
...
-    -> s
-    -> (UTxO, s)
-utxoOurs tx = runState $ toUtxo <$> forM (zip [0..] (outputs tx)) filterOut
-  where
-    toUtxo = UTxO . Map.fromList . catMaybes
-    filterOut (ix, out) = do
-        state (isOurs $ address out) <&> \case
-            Just{}  -> Just (TxIn (txId tx) ix, out)
-            Nothing -> Nothing
+    mconcat <$> mapM (filterByAddressM isOurs' . utxoFromTx) (Set.toList pending)
```
jonathanknowles: Nice! 👍🏻

lib/core/src/Cardano/Wallet/Primitive/Types/UTxO.hs
```diff
@@ -136,6 +144,32 @@ null (UTxO u) = Map.null u
...
+-- Returns the subset of UTxO entries that have addresses for which the given
+-- indicator function returns 'True'.
+filterByAddressM :: forall f. Monad f => (Address -> f Bool) -> UTxO -> f UTxO
+filterByAddressM isOursF (UTxO m) =
+    UTxO <$> Map.traverseMaybeWithKey filterFunc m
+    where
+        filterFunc :: TxIn -> TxOut -> f (Maybe TxOut)
+        filterFunc _txin txout = do
+            ours <- isOursF $ view #address txout
+            pure $ if ours then Just txout else Nothing
```
jonathanknowles: Formatting:

```suggestion
filterByAddressM isOursF (UTxO m) =
    UTxO <$> Map.traverseMaybeWithKey filterFunc m
  where
    filterFunc :: TxIn -> TxOut -> f (Maybe TxOut)
    filterFunc _txin txout = do
        ours <- isOursF $ view #address txout
        pure $ if ours then Just txout else Nothing
```

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -182,6 +184,27 @@ spec = do
...
+            it "has expected entries"
+                (property prop_applyTxToUTxO_entries)
+            it "applyTxToUTxO then filterByAddress"
+                (property prop_filterByAddress_balance_applyTxToUTxO)
+            it "spendTx/utxoFromTx"
+                (property prop_applyTxToUTxO_spendTx_utxoFromTx)
+
+        describe "utxoFromTx" $ do
+            it "has expected balance" (property prop_utxoFromTx_balance)
+            it "is unspent" (property prop_utxoFromTx_is_unspent)
```
jonathanknowles: Suggestion: keep to the prevailing style:
```suggestion
        describe "utxoFromTx" $ do
            it "has expected balance"
                (property prop_utxoFromTx_balance)
            it "is unspent"
                (property prop_utxoFromTx_is_unspent)
```

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -182,6 +184,27 @@ spec = do
...
+
+        describe "spendTx" $ do
+            it "is subset of UTxO" (property prop_spendTx_isSubset)
+            it "balance is <= balance of UTxO" (property prop_spendTx_balance_inequality)
+            it "has expected balance" (property prop_spendTx_balance)
+            it "definition" (property prop_spendTx)
+            it "commutative with filterByAddress" (property prop_spendTx_filterByAddress)
```
jonathanknowles: Suggestion: keep to the prevailing style:

```suggestion
        describe "spendTx" $ do
            it "is subset of UTxO"
                (property prop_spendTx_isSubset)
            it "balance is <= balance of UTxO"
                (property prop_spendTx_balance_inequality)
            it "has expected balance"
                (property prop_spendTx_balance)
            it "definition"
                (property prop_spendTx)
            it "commutative with filterByAddress"
                (property prop_spendTx_filterByAddress)
```

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -1407,3 +1372,105 @@ blockchain =
...
+prop_applyTxToUTxO_balance :: Property
+prop_applyTxToUTxO_balance =
+    forAllShrink genTx shrinkTx $ \tx ->
+    forAllShrink genUTxO shrinkUTxO $ \u ->
+        balance (applyTxToUTxO tx u)
+        === balance u
+          `TokenBundle.add`
+              balance (utxoFromTx tx)
+          `TokenBundle.difference`
+              balance (u `UTxO.restrictedBy` Set.fromList (inputs tx))
```
jonathanknowles: Since we don't modify the generators and shrinkers in any way, perhaps we could define `Arbitrary` instances for `Tx` and `UTxO`, and simplify these properties?

For example:

```suggestion
prop_applyTxToUTxO_balance :: Tx -> UTxO -> Property
prop_applyTxToUTxO_balance tx u =   
    balance (applyTxToUTxO tx u) ===
    balance u
        `TokenBundle.add`
            balance (utxoFromTx tx)
        `TokenBundle.difference`
            balance (u `UTxO.restrictedBy` Set.fromList (inputs tx))
```

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -1407,3 +1372,105 @@ blockchain =
...
+prop_applyTxToUTxO_entries :: Property
+prop_applyTxToUTxO_entries =
+    forAllShrink genTx shrinkTx $ \tx ->
+    forAllShrink genUTxO shrinkUTxO $ \u ->
+        unUTxO (applyTxToUTxO tx u)
+        === unUTxO u
+          `Map.union`
+              unUTxO (utxoFromTx tx)
+          `Map.difference`
+              unUTxO (u `UTxO.restrictedBy` Set.fromList (inputs tx))
```
jonathanknowles: Suggestion: use `Arbitrary` instances here.

See https://github.com/input-output-hk/cardano-wallet/pull/2848/files#r700738862

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -1407,3 +1372,105 @@ blockchain =
...
+prop_filterByAddress_balance_applyTxToUTxO :: Bool -> Property
+prop_filterByAddress_balance_applyTxToUTxO b =
+    let
+        f = const b
+    in
+        forAllShrink genTx shrinkTx $ \tx ->
+            balance (filterByAddress f (applyTxToUTxO tx mempty))
+            === foldMap (\o -> do
+                    if f (address o) then tokens o else mempty
+                ) (outputs tx)
```
jonathanknowles: Suggestion: use `Arbitrary` instances here.

See https://github.com/input-output-hk/cardano-wallet/pull/2848/files#r700738862

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -1407,3 +1372,105 @@ blockchain =
...
+
+prop_utxoFromTx_balance :: Property
+prop_utxoFromTx_balance =
+    forAllShrink genTx shrinkTx $ \tx ->
+        balance (utxoFromTx tx) === foldMap tokens (outputs tx)
```
jonathanknowles: Suggestion: use `Arbitrary` instances here.

See https://github.com/input-output-hk/cardano-wallet/pull/2848/files#r700738862

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -1407,3 +1372,105 @@ blockchain =
...
+
+prop_utxoFromTx_is_unspent :: Property
+prop_utxoFromTx_is_unspent =
+    forAllShrink genTx shrinkTx $ \tx ->
+        utxoFromTx tx `excluding` Set.fromList (inputs tx)
+        === utxoFromTx tx
```
jonathanknowles: Suggestion: use `Arbitrary` instances here.

See https://github.com/input-output-hk/cardano-wallet/pull/2848/files#r700738862

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -1407,3 +1372,105 @@ blockchain =
...
+
+-- spendTx tx u `isSubsetOf` u
+prop_spendTx_isSubset :: Property
+prop_spendTx_isSubset =
+    forAllShrink genTx shrinkTx $ \tx ->
+    forAllShrink genUTxO shrinkUTxO $ \u ->
+        spendTx tx u `UTxO.isSubsetOf` u
```
jonathanknowles: Suggestion: use `Arbitrary` instances here.

See https://github.com/input-output-hk/cardano-wallet/pull/2848/files#r700738862

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -1407,3 +1372,105 @@ blockchain =
...
+
+-- balance (spendTx tx u) <= balance u
+prop_spendTx_balance_inequality :: Property
+prop_spendTx_balance_inequality =
+    forAllShrink genTx shrinkTx $ \tx ->
+    forAllShrink genUTxO shrinkUTxO $ \u ->
```
jonathanknowles: Suggestion: use `Arbitrary` instances here.

See https://github.com/input-output-hk/cardano-wallet/pull/2848/files#r700738862

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -1407,3 +1372,105 @@ blockchain =
...
+            rhs = balance u
+        in
+            isJust (rhs `TokenBundle.subtract` lhs)
+            & counterexample ("balance (spendTx tx u) = " <> show lhs)
+            & counterexample ("balance u = " <> show rhs)
+
+prop_spendTx_balance :: Property
+prop_spendTx_balance =
+    forAllShrink genTx shrinkTx $ \tx ->
+    forAllShrink genUTxO shrinkUTxO $ \u ->
```
jonathanknowles: Suggestion: use `Arbitrary` instances here.

See https://github.com/input-output-hk/cardano-wallet/pull/2848/files#r700738862

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -1407,3 +1372,105 @@ blockchain =
...
+                balance u
+                    `TokenBundle.unsafeSubtract`
+                        balance (u `UTxO.restrictedBy` Set.fromList (inputs tx))
+        in
+            lhs === rhs
+
+prop_spendTx :: Property
+prop_spendTx =
+    forAllShrink genTx shrinkTx $ \tx ->
+    forAllShrink genUTxO shrinkUTxO $ \u ->
```
jonathanknowles: Suggestion: use `Arbitrary` instances here.

See https://github.com/input-output-hk/cardano-wallet/pull/2848/files#r700738862

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -1407,3 +1372,105 @@ blockchain =
...
+        spendTx tx u === u `excluding` Set.fromList (inputs tx)
+
+prop_applyTxToUTxO_spendTx_utxoFromTx :: Property
+prop_applyTxToUTxO_spendTx_utxoFromTx =
+    forAllShrink genTx shrinkTx $ \tx ->
+    forAllShrink genUTxO shrinkUTxO $ \u ->
```
jonathanknowles: Suggestion: use `Arbitrary` instances here.

See https://github.com/input-output-hk/cardano-wallet/pull/2848/files#r700738862

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -1407,3 +1372,105 @@ blockchain =
...
+        , applyTxToUTxO tx u === spendTx tx (u <> utxoFromTx tx)
+        ]
+
+prop_spendTx_filterByAddress :: Bool -> Property
+prop_spendTx_filterByAddress b =
+    let
+        f = const b
+    in
+        forAllShrink genTx shrinkTx $ \tx ->
+        forAllShrink genUTxO shrinkUTxO $ \u ->
```
jonathanknowles: Suggestion: use `Arbitrary` instances here.

See https://github.com/input-output-hk/cardano-wallet/pull/2848/files#r700738862

jonathanknowles (commented):
Hi Sam

I've already made a few comments. But to add to those:

1. I reckon it would be useful to add some coverage checks to the following properties:
      - [x] `prop_applyTxToUTxO_balance`
      - [x] `prop_applyTxToUTxO_entries`
      - [x] `prop_applyTxToUTxO_spendTx_utxoFromTx`
      - [x] `prop_filterByAddress_balance_applyTxToUTxO`
      - [x] `prop_spendTx_balance_inequality`
      - [x] `prop_spendTx_balance`
      - [x] `prop_spendTx_filterByAddress`
      - [x] `prop_spendTx_isSubset`
      - [x] `prop_spendTx`
      - [x] `prop_utxoFromTx_balance`
      - [x] `prop_utxoFromTx_is_unspent`

1. There are several properties where we use `forAll` with `genX` and `shrinkX` for some type `X`, where we don't modify the generator in any way.

    I think we **_could_** simplify the code a bit by introducing `Arbitrary` instances (within `ModelSpec`) for `UTxO`, `Tx`, and so on. For example: https://github.com/input-output-hk/cardano-wallet/pull/2848/files#r700738862. Though I reckon it's a good idea to do this only after adding coverage checks, in case we need to alter the size parameter for any of the generators.

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -320,6 +322,56 @@ totalUTxO
...
+--     `Map.difference` unUTxO (u `restrictedBy` inputs tx)
+-- applyTxToUTxO tx u = spend tx u <> utxoFromTx tx
+-- applyTxToUTxO tx u = spend tx (u <> utxoFromTx tx)
+applyTxToUTxO
+    :: Tx
+    -> UTxO
+    -> UTxO
+applyTxToUTxO tx !u = spendTx tx u <> utxoFromTx tx
+
+-- | Remove unspents that have been consumed by the giventransaction.
```
jonathanknowles: 👾

```suggestion
-- | Remove unspents that have been consumed by the given transaction.
```

sevanspowell: ffedc07

jonathanknowles (approved):
LGTM!

I've run the test suite with `--qc-max-success 10000`, and all the tests pass.

sevanspowell:
bors retry

jonathanknowles:
bors retry

PR #2851, opened 2021-08-26
Buildkite: Run weeder as early as possible, after build, before test

rvl:
1. Moves the Weeder command earlier in the Stack Rebuild step.
2. Updates Weeder to 2.2.0, which fixes the problem of false weeds due to stale `.hie` files.

### DIFF
--- a/.buildkite/rebuild.hs
+++ b/.buildkite/rebuild.hs
@@ -259,6 +259,9 @@ listLocalPackages =
     isBuild = contains "./.stack-work" <|> contains "./dist"
     textLines shell = fold (format fp <$> shell) Fold.list
 
+findStackWork :: FilePath -> Shell FilePath
+findStackWork = find (ends "/.stack-work")
+
 ----------------------------------------------------------------------------
 -- Buildkite
 -- https://buildkite.com/docs/pipelines/environment-variables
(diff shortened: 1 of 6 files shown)


rvl:
We have a problem with `.stack-work` caching and `.hie` files.

If there is a build where a module was removed or renamed, but where the `.hie` file remains in `.stack-work`, Weeder claims that every symbol in that dangling `.hie` file is a weed. Which is true, but beside the point.

I first attempted to fix this by removing all `.hie` files from `.stack-work` before building.
But then, if `stack build` is run, any modules where there is already an up-to-date `.o` file will not be rebuilt (which is good), and thus the `.hie` file is not rebuilt (which isn't good).

Weeder only looks in `.hie` files. If they aren't present, then it says "no weeds".

So I believe the solution is to selectively clean just the `.hie` files for which the corresponding `.hs` module doesn't exist. But this might be kinda tricky.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2941,3 +2942,7 @@ instance HasSeverityAnnotation TxSubmitLog where
         MsgRetryPostTxResult _ (Right _) -> Info
         MsgRetryPostTxResult _ (Left _) -> Debug
         MsgProcessPendingPool msg -> getSeverityAnnotation msg
+
+-- | Empty fluff.
+nullFluff :: Monoid a => a
+nullFluff = mempty
```
jonathanknowles: Hi @rvl 

It seems this doesn't cause weeder to fail, at least not in [this build](https://buildkite.com/input-output-hk/cardano-wallet/builds/16364#24482922-1720-469b-9589-5a4f096483df).

Perhaps I've misunderstood something?

rvl: Yes - https://github.com/input-output-hk/cardano-wallet/pull/2851#issuecomment-906917880

rvl:
It looks like the [2.2.0 Weeder release](https://github.com/ocharles/weeder/blob/master/CHANGELOG.md) from a few days ago fixes the issue with reading `.hie` files that don't have a corresponding `.hs` file. :tada:

HeinrichApfelmus:
We no longer use Stack. The context for this pull request has been lost to time. Closing.

PR #2856, opened 2021-08-27
Adjust chain-following code to account for collateral inputs

sevanspowell:
The purpose of this PR is to modify the the code that follows the blockchain to adjust the Wallet's UTxO when collateral inputs are lost due to a transaction that fails phase-2 script validation (Plutus scripts).

I have:
- Add valid script field to `MkApiTransactionParams`, `ApiTransaction`, `TxMeta`, `Tx`, and `TransactionInfo`.
- Modify state transition logic (when a Tx is applied to a UTxO) to account for collateral inputs.
- Add integration test for collateral inputs.
- Add property tests for state transition logic.
- Add is_valid_script field to API specification.

### Issue Number

ADP-1036

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -2892,7 +2892,7 @@ data MkApiTransactionParams = MkApiTransactionParams
     , txMeta :: W.TxMeta
     , txMetadata :: Maybe W.TxMetadata
     , txTime :: UTCTime
-    , txIsValidScript :: Maybe Bool
+    , txIsValidScript :: W.ScriptValidation
     }
     deriving (Eq, Generic, Show)
 
@@ -2942,7 +2942,7 @@ mkApiTransaction timeInterpreter setTimeReference tx = do
         , mint = mempty  -- TODO: ADP-xxx
         , status = ApiT (tx ^. (#txMeta . #status))
         , metadata = ApiTxMetadata $ ApiT <$> (tx ^. #txMetadata)
-        , isValidScript = tx ^. #txIsValidScript
+        , isValidScript = ApiT $ tx ^. #txIsValidScript
         }
 
     depositIfAny :: Natural
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -1883,6 +1883,7 @@ postTransactionOld ctx genChange (ApiT wid) body = do
             , txMeta
             , txMetadata = tx ^. #metadata
             , txTime
+            , txIsValidScript = (tx ^. #isValidScript)
             }
   where
     ti :: TimeInterpreter (ExceptT PastHorizonException IO)
@@ -1950,6 +1951,7 @@ mkApiTransactionFromInfo ti info = do
             , txMeta = info ^. #txInfoMeta
             , txMetadata = info ^. #txInfoMetadata
             , txTime = info ^. #txInfoTime
+            , txIsValidScript = info ^. #txInfoIsValidScript
             }
     return $ case info ^. (#txInfoMeta . #status) of
         Pending  -> apiTx
@@ -2145,6 +2147,7 @@ joinStakePool ctx knownPools getPoolStatus apiPoolId (ApiT wid) body = do
             , txMeta
             , txMetadata = Nothing
             , txTime
+            , txIsValidScript = tx ^. #isValidScript
             }
   where
     ti :: TimeInterpreter (ExceptT PastHorizonException IO)
@@ -2235,6 +2238,7 @@ quitStakePool ctx (ApiT wid) body = do
             , txMeta
             , txMetadata = Nothing
             , txTime
+            , txIsValidScript = tx ^. #isValidScript
             }
   where
     ti :: TimeInterpreter (ExceptT PastHorizonException IO)
@@ -2489,6 +2493,7 @@ migrateWallet ctx withdrawalType (ApiT wid) postData = do
                     , txMeta
                     , txMetadata = Nothing
                     , txTime
+                    , txIsValidScript = tx ^. #isValidScript
                     }
   where
     addresses = getApiT . fst <$> view #addresses postData
@@ -2887,6 +2892,7 @@ data MkApiTransactionParams = MkApiTransactionParams
     , txMeta :: W.TxMeta
     , txMetadata :: Maybe W.TxMetadata
     , txTime :: UTCTime
+    , txIsValidScript :: Maybe Bool
     }
     deriving (Eq, Generic, Show)
 
@@ -2936,6 +2942,7 @@ mkApiTransaction timeInterpreter setTimeReference tx = do
         , mint = mempty  -- TODO: ADP-xxx
         , status = ApiT (tx ^. (#txMeta . #status))
         , metadata = ApiTxMetadata $ ApiT <$> (tx ^. #txMetadata)
+        , isValidScript = tx ^. #txIsValidScript
         }
 
     depositIfAny :: Natural
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -2892,7 +2892,7 @@ data MkApiTransactionParams = MkApiTransactionParams
     , txMeta :: W.TxMeta
     , txMetadata :: Maybe W.TxMetadata
     , txTime :: UTCTime
-    , txIsValidScript :: W.ScriptValidation
+    , txIsValidScript :: W.TxScriptValidity
     }
     deriving (Eq, Generic, Show)
 
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -1883,7 +1883,7 @@ postTransactionOld ctx genChange (ApiT wid) body = do
             , txMeta
             , txMetadata = tx ^. #metadata
             , txTime
-            , txIsValidScript = (tx ^. #isValidScript)
+            , txScriptValidity = tx ^. #scriptValidity
             }
   where
     ti :: TimeInterpreter (ExceptT PastHorizonException IO)
@@ -1951,7 +1951,7 @@ mkApiTransactionFromInfo ti info = do
             , txMeta = info ^. #txInfoMeta
             , txMetadata = info ^. #txInfoMetadata
             , txTime = info ^. #txInfoTime
-            , txIsValidScript = info ^. #txInfoIsValidScript
+            , txScriptValidity = info ^. #txInfoScriptValidity
             }
     return $ case info ^. (#txInfoMeta . #status) of
         Pending  -> apiTx
@@ -2147,7 +2147,7 @@ joinStakePool ctx knownPools getPoolStatus apiPoolId (ApiT wid) body = do
             , txMeta
             , txMetadata = Nothing
             , txTime
-            , txIsValidScript = tx ^. #isValidScript
+            , txScriptValidity = tx ^. #scriptValidity
             }
   where
     ti :: TimeInterpreter (ExceptT PastHorizonException IO)
@@ -2238,7 +2238,7 @@ quitStakePool ctx (ApiT wid) body = do
             , txMeta
             , txMetadata = Nothing
             , txTime
-            , txIsValidScript = tx ^. #isValidScript
+            , txScriptValidity = tx ^. #scriptValidity
             }
   where
     ti :: TimeInterpreter (ExceptT PastHorizonException IO)
@@ -2493,7 +2493,7 @@ migrateWallet ctx withdrawalType (ApiT wid) postData = do
                     , txMeta
                     , txMetadata = Nothing
                     , txTime
-                    , txIsValidScript = tx ^. #isValidScript
+                    , txScriptValidity = tx ^. #scriptValidity
                     }
   where
     addresses = getApiT . fst <$> view #addresses postData
@@ -2892,7 +2892,7 @@ data MkApiTransactionParams = MkApiTransactionParams
     , txMeta :: W.TxMeta
     , txMetadata :: Maybe W.TxMetadata
     , txTime :: UTCTime
-    , txIsValidScript :: W.TxScriptValidity
+    , txScriptValidity :: W.TxScriptValidity
     }
     deriving (Eq, Generic, Show)
 
@@ -2942,7 +2942,7 @@ mkApiTransaction timeInterpreter setTimeReference tx = do
         , mint = mempty  -- TODO: ADP-xxx
         , status = ApiT (tx ^. (#txMeta . #status))
         , metadata = ApiTxMetadata $ ApiT <$> (tx ^. #txMetadata)
-        , isValidScript = ApiT $ tx ^. #txIsValidScript
+        , scriptValidity = ApiT $ tx ^. #txScriptValidity
         }
 
     depositIfAny :: Natural
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -2892,7 +2892,7 @@ data MkApiTransactionParams = MkApiTransactionParams
     , txMeta :: W.TxMeta
     , txMetadata :: Maybe W.TxMetadata
     , txTime :: UTCTime
-    , txScriptValidity :: W.TxScriptValidity
+    , txScriptValidity :: Maybe W.TxScriptValidity
     }
     deriving (Eq, Generic, Show)
 
@@ -2942,7 +2942,7 @@ mkApiTransaction timeInterpreter setTimeReference tx = do
         , mint = mempty  -- TODO: ADP-xxx
         , status = ApiT (tx ^. (#txMeta . #status))
         , metadata = ApiTxMetadata $ ApiT <$> (tx ^. #txMetadata)
-        , scriptValidity = ApiT $ tx ^. #txScriptValidity
+        , scriptValidity = ApiT <$> tx ^. #txScriptValidity
         }
 
     depositIfAny :: Natural
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -1094,7 +1094,7 @@ data ApiTransaction (n :: NetworkDiscriminant) = ApiTransaction
     , mint :: !(ApiT W.TokenMap)
     , status :: !(ApiT TxStatus)
     , metadata :: !ApiTxMetadata
-    , scriptValidity :: !(ApiT TxScriptValidity)
+    , scriptValidity :: !(Maybe (ApiT TxScriptValidity))
     } deriving (Eq, Generic, Show, Typeable)
       deriving anyclass NFData
 
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -3641,9 +3641,9 @@ instance ToJSON (ApiT (Script KeyHash)) where
     toJSON = toJSON . getApiT
 
 instance FromJSON (ApiT TxScriptValidity) where
-    parseJSON = fmap ApiT . genericParseJSON (Aeson.defaultOptions
-        { constructorTagModifier = camelTo2 '_' . drop 8 })
+    parseJSON = fmap ApiT . genericParseJSON Aeson.defaultOptions
+        { constructorTagModifier = camelTo2 '_' . drop 8 }
 
 instance ToJSON (ApiT TxScriptValidity) where
-    toJSON = genericToJSON (Aeson.defaultOptions
-        { constructorTagModifier = camelTo2 '_' . drop 8 }) . getApiT
+    toJSON = genericToJSON Aeson.defaultOptions
+        { constructorTagModifier = camelTo2 '_' . drop 8 } . getApiT
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -304,11 +304,11 @@ import Cardano.Wallet.Primitive.Types.Hash
     ( Hash (..) )
 import Cardano.Wallet.Primitive.Types.Tx
     ( Direction (..)
-    , ScriptValidation (..)
     , SerialisedTx (..)
     , TxConstraints (..)
     , TxIn (..)
     , TxMetadata
+    , TxScriptValidity (..)
     , TxStatus (..)
     , txMetadataIsNull
     )
@@ -346,6 +346,7 @@ import Data.Aeson.Types
     , rejectUnknownFields
     , sumEncoding
     , tagSingleConstructors
+    , typeMismatch
     , withObject
     , withText
     , (.!=)
@@ -1094,7 +1095,7 @@ data ApiTransaction (n :: NetworkDiscriminant) = ApiTransaction
     , mint :: !(ApiT W.TokenMap)
     , status :: !(ApiT TxStatus)
     , metadata :: !ApiTxMetadata
-    , isValidScript :: !(ApiT ScriptValidation)
+    , isValidScript :: !(ApiT TxScriptValidity)
     } deriving (Eq, Generic, Show, Typeable)
       deriving anyclass NFData
 
@@ -3640,13 +3641,15 @@ instance FromJSON (ApiT (Script KeyHash)) where
 instance ToJSON (ApiT (Script KeyHash)) where
     toJSON = toJSON . getApiT
 
-instance FromJSON (ApiT ScriptValidation) where
-    parseJSON (Aeson.Null) = pure $ ApiT ScriptsNotSupported
-    parseJSON (Aeson.Bool True) = pure $ ApiT ScriptValidationPassed
-    parseJSON (Aeson.Bool False) = pure $ ApiT ScriptValidationFailed
-    parseJSON _ = pure $ ApiT ScriptValidationFailed
+instance FromJSON (ApiT TxScriptValidity) where
+    parseJSON (Aeson.Null) = pure $ ApiT TxScriptsUnsupported
+    parseJSON (Aeson.Bool True) = pure $ ApiT TxScriptValid
+    parseJSON (Aeson.Bool False) = pure $ ApiT TxScriptInvalid
+    parseJSON invalid =
+        prependFailure "parsing TxScriptValidity failed, "
+            (typeMismatch "Maybe Bool" invalid)
 
-instance ToJSON (ApiT ScriptValidation) where
-    toJSON (ApiT ScriptsNotSupported) = Aeson.Null
-    toJSON (ApiT ScriptValidationPassed) = Aeson.Bool True
-    toJSON (ApiT ScriptValidationFailed) = Aeson.Bool False
+instance ToJSON (ApiT TxScriptValidity) where
+    toJSON (ApiT TxScriptsUnsupported) = Aeson.Null
+    toJSON (ApiT TxScriptValid) = Aeson.Bool True
+    toJSON (ApiT TxScriptInvalid) = Aeson.Bool False
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -346,7 +346,6 @@ import Data.Aeson.Types
     , rejectUnknownFields
     , sumEncoding
     , tagSingleConstructors
-    , typeMismatch
     , withObject
     , withText
     , (.!=)
@@ -3642,14 +3641,9 @@ instance ToJSON (ApiT (Script KeyHash)) where
     toJSON = toJSON . getApiT
 
 instance FromJSON (ApiT TxScriptValidity) where
-    parseJSON (Aeson.Null) = pure $ ApiT TxScriptsUnsupported
-    parseJSON (Aeson.Bool True) = pure $ ApiT TxScriptValid
-    parseJSON (Aeson.Bool False) = pure $ ApiT TxScriptInvalid
-    parseJSON invalid =
-        prependFailure "parsing TxScriptValidity failed, "
-            (typeMismatch "Maybe Bool" invalid)
+    parseJSON = fmap ApiT . genericParseJSON (Aeson.defaultOptions
+        { constructorTagModifier = camelTo2 '_' . drop 8 })
 
 instance ToJSON (ApiT TxScriptValidity) where
-    toJSON (ApiT TxScriptsUnsupported) = Aeson.Null
-    toJSON (ApiT TxScriptValid) = Aeson.Bool True
-    toJSON (ApiT TxScriptInvalid) = Aeson.Bool False
+    toJSON = genericToJSON (Aeson.defaultOptions
+        { constructorTagModifier = camelTo2 '_' . drop 8 }) . getApiT
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -1095,7 +1095,7 @@ data ApiTransaction (n :: NetworkDiscriminant) = ApiTransaction
     , mint :: !(ApiT W.TokenMap)
     , status :: !(ApiT TxStatus)
     , metadata :: !ApiTxMetadata
-    , isValidScript :: !(ApiT TxScriptValidity)
+    , scriptValidity :: !(ApiT TxScriptValidity)
     } deriving (Eq, Generic, Show, Typeable)
       deriving anyclass NFData
 
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -1093,6 +1093,7 @@ data ApiTransaction (n :: NetworkDiscriminant) = ApiTransaction
     , mint :: !(ApiT W.TokenMap)
     , status :: !(ApiT TxStatus)
     , metadata :: !ApiTxMetadata
+    , isValidScript :: !(Maybe Bool)
     } deriving (Eq, Generic, Show, Typeable)
       deriving anyclass NFData
 
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -304,6 +304,7 @@ import Cardano.Wallet.Primitive.Types.Hash
     ( Hash (..) )
 import Cardano.Wallet.Primitive.Types.Tx
     ( Direction (..)
+    , ScriptValidation (..)
     , SerialisedTx (..)
     , TxConstraints (..)
     , TxIn (..)
@@ -1093,7 +1094,7 @@ data ApiTransaction (n :: NetworkDiscriminant) = ApiTransaction
     , mint :: !(ApiT W.TokenMap)
     , status :: !(ApiT TxStatus)
     , metadata :: !ApiTxMetadata
-    , isValidScript :: !(Maybe Bool)
+    , isValidScript :: !(ApiT ScriptValidation)
     } deriving (Eq, Generic, Show, Typeable)
       deriving anyclass NFData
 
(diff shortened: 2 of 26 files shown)


lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2880,6 +2885,7 @@ data MkApiTransactionParams = MkApiTransactionParams
     , txMeta :: W.TxMeta
     , txMetadata :: Maybe W.TxMetadata
     , txTime :: UTCTime
+    , txIsValidScript :: Maybe Bool
```
sevanspowell: I need to choose a more descriptive type for this:

Nothing -> Not an era that cares about validation
Just False -> Cares about validation, but validation failed
Just True -> Cares about validation and validation succeeded

sevanspowell: 06493c0

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -503,7 +512,9 @@ prefilterBlock b u0 = runState $ do
 
         return $ if hasKnownOutput && not hasKnownInput then
```
sevanspowell, line 543: A Tx that fails will have no known output nor any known input, thanks to the implementation of `utxoFromTx`. This means it will hit the last `else` in this `if-else`, which is to just use the `prevUTxO`, but we also want to remove collateral from the UTxO...

I need to write a failing test to demonstrate what I mean.

sevanspowell: I think this will suffice as an implementation:

```
        let hasKnownInput =
                (if failedScriptValidation (tx ^. #isValidScript)
                 then Set.fromList (collateralInputs tx)
                 else Set.fromList (inputs tx)
                ) `Set.intersection` (Set.fromList ownedAndKnownTxIns)
                /= mempty
```

sevanspowell: Actually not quite sufficient. We can’t presume that we already know about our inputs that could be used for collateral. The transaction may introduce collateral inputs that we own, so we need to expand the right hand side of the Set intersection too.

sevanspowell: It is surprising this logic is not tested at a unit level, it’s very particular.

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
```diff
@@ -459,10 +478,31 @@ data TransactionInfo = TransactionInfo
...
     } deriving (Generic, Show, Eq)
 
 instance NFData TransactionInfo
 
+-- | Indicates whether the script associated with a transaction has passed or
+-- failed validation. Pre-Alonzo era, scripts were not supported.
+data ScriptValidation = ScriptsNotSupported
+                      | ScriptValidationPassed
+                      | ScriptValidationFailed
+  deriving (Generic, Show, Eq, Ord)
```
jonathanknowles: Suggestions:
1. Use the `Tx` prefix, since this relates to a transaction and appears in the `Tx` module.
2. Use `Validity`, as we can then say "what is the validity of the script for this transaction"?

As for the second point, we could use "validation status".

```suggestion
-- | Indicates whether the script associated with a transaction has passed or
-- failed validation. Pre-Alonzo era, scripts were not supported.
data TxScriptValidity
    = TxScriptsUnsupported
    -- ^ Indicates that scripts were not supported in the era in which this
    -- transaction was created.
    | TxScriptValid
    -- ^ Indicates that the script passed validation.
    | TxScriptInvalid
    -- ^ Indicates that the script failed validation.
  deriving (Generic, Show, Eq, Ord)
```

lib/core/src/Cardano/Wallet/Primitive/Types/Tx/Gen.hs
```diff
@@ -125,12 +132,18 @@ txToTxWithoutId :: Tx -> TxWithoutId
...
+    (a1, a2, a3, a4, a5, a6, a7)
 
 tupleToTxWithoutId :: _ -> TxWithoutId
-tupleToTxWithoutId (a1, a2, a3, a4, a5, a6) =
-    (TxWithoutId a1 a2 a3 a4 a5 a6)
+tupleToTxWithoutId (a1, a2, a3, a4, a5, a6, a7) =
+    (TxWithoutId a1 a2 a3 a4 a5 a6 a7)
+
+genScriptValidation :: Gen ScriptValidation
+genScriptValidation = genericArbitrary
```
jonathanknowles: Here you could also use `arbitraryBoundedEnum`, which will generate a uniform distribution across all possibilities.

lib/core/test/unit/Cardano/WalletSpec.hs
```diff
@@ -1284,7 +1285,14 @@ dummyTransactionLayer = TransactionLayer
...
         let tid = mkTxId inps' (outputsCovered cs) mempty Nothing
-        let tx = Tx tid Nothing inps' cinps' (outputsCovered cs) mempty Nothing
+        let tx = Tx tid
+                    Nothing
+                    inps'
+                    cinps'
+                    (outputsCovered cs)
+                    mempty
+                    Nothing
+                    ScriptsNotSupported
```
jonathanknowles: Suggestion: use record construction syntax here. It's easier to guess what `Nothing` refers to if there's a field name provided!

specifications/api/swagger.yaml
```diff
@@ -2097,6 +2105,7 @@ components:
         mint: *transactionMint
         status: *transactionStatus
         metadata: *transactionMetadata
+        is_valid_script: *transactionScriptValidity
```
jonathanknowles: Suggestion: use `script_validity` here:
```suggestion
        script_validity: *transactionScriptValidity
```

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2880,6 +2885,7 @@ data MkApiTransactionParams = MkApiTransactionParams
     , txMeta :: W.TxMeta
     , txMetadata :: Maybe W.TxMetadata
     , txTime :: UTCTime
+    , txIsValidScript :: W.ScriptValidation
```
```suggestion
    , txScriptValidity :: W.TxScriptValidity
```

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
```diff
@@ -111,6 +111,7 @@ TxMeta
     txMetadata              W.TxMetadata Maybe  sql=data
     txMetaSlotExpires       SlotNo Maybe        sql=slot_expires
     txMetaFee               Word64 Maybe        sql=fee
+    txMetaIsValid           Bool Maybe          sql=is_valid
```
jonathanknowles:
```suggestion
    txMetaScriptValidity    Bool Maybe          sql=script_validity
```

lib/core/test/unit/Cardano/WalletSpec.hs
sevanspowell: I'm glad you said that, because the order of inputs and collateralInputs was incorrect!

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
sevanspowell: a6a26a6

lib/core/test/unit/Cardano/WalletSpec.hs
sevanspowell: e2dff11

sevanspowell: 95b37cf

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
sevanspowell: a6a26a6

lib/core/src/Cardano/Wallet/Api/Server.hs
sevanspowell: a6a26a6

specifications/api/swagger.yaml
sevanspowell: f21ba22

specifications/api/swagger.yaml
```diff
@@ -1674,6 +1674,14 @@ x-maximumTokenBundleSize: &maximumTokenBundleSize
...
         - byte
 
+x-transactionScriptValidity: &transactionScriptValidity
+  description: |
+    For Alonzo-era and later transactions, indicates whether the phase-2
+    monetary policy script (e.g. Plutus script) used in the transaction
+    validated or not. Previous to Alonzo-era, this is null.
+  nullable: true
+  type: boolean
+  default: null
```
jonathanknowles: Hi @sevanspowell

I think it might be better if this was an _enumeration_, rather than a Boolean.

For example:
```yaml
  x-txScriptValidity: &txScriptValidity
    type: string
    enum:
      - unsupported
      - valid
      - invalid
```

I think this has the following advantages:

1. We can generate the `ToJSON` and `FromJSON` instances generically, rather than having to write them manually (which requires maintenance), as it's now just a 1:1 mapping from the API type to the primitive type.
2. We don't need to deal with the missing case, as it's represented explicitly.
3. The individual cases are more self-documenting.

We also use enumerations elsewhere in the API to represent this kind of multi-option value. So it would be consistent with what we've done before. For example:

```hs
data ApiEra
    = ApiByron
    | ApiShelley
    | ApiAllegra
    | ApiMary
    | ApiAlonzo
    deriving (Show, Eq, Generic, Enum, Ord, Bounded)
    deriving anyclass NFData

instance FromJSON ApiEra where
    parseJSON = genericParseJSON $ Aeson.defaultOptions
        { constructorTagModifier = drop 4 . camelTo2 '_' }
instance ToJSON ApiEra where
    toJSON = genericToJSON $ Aeson.defaultOptions
        { constructorTagModifier = drop 4 . camelTo2 '_' }  
```

Thoughts?

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1979,6 +1990,11 @@ txHistoryFromEntity ti tip metas ins cins outs ws =
                 Quantity $ fromIntegral $ if tipH > txH then tipH - txH else 0
             , W.txInfoTime =
                 t
+            , W.txInfoScriptValidity =
+                    case isValid of
+                        Nothing -> W.TxScriptsUnsupported
+                        Just False -> W.TxScriptInvalid
+                        Just True -> W.TxScriptValid
```
jonathanknowles: Formatting: use 4-space indents:

```suggestion
            , W.txInfoScriptValidity =
                case isValid of
                    Nothing -> W.TxScriptsUnsupported
                    Just False -> W.TxScriptInvalid
                    Just True -> W.TxScriptValid
```

lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs
```diff
@@ -111,6 +111,7 @@ TxMeta
     txMetadata              W.TxMetadata Maybe  sql=data
     txMetaSlotExpires       SlotNo Maybe        sql=slot_expires
     txMetaFee               Word64 Maybe        sql=fee
+    txMetaScriptValidity    Bool Maybe          sql=script_validity
```
jonathanknowles, line 114: In some ways it would be nice to use an enumeration type here that matches the primitive type, as this would make it easier to add additional constructors in future, should we need to.

But that would presumably require a DB migration to get right. (Because the column would be missing for existing transactions.)

So I think it's the right choice to use a `Bool Maybe` here.

lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs
```diff
@@ -798,6 +807,29 @@ instance PersistField DerivationPrefix where
 instance PersistFieldSql DerivationPrefix where
     sqlType _ = sqlType (Proxy @Text)
 
+----------------------------------------------------------------------------
+-- ScriptValidation
+
+instance PersistField TxScriptValidity where
+    toPersistValue TxScriptsUnsupported = PersistNull
+    toPersistValue TxScriptValid = PersistBool True
+    toPersistValue TxScriptInvalid = PersistBool False
```
jonathanknowles: Suggestion: use `LambdaCase` here for brevity:

```suggestion
    toPersistValue = \case
        TxScriptsUnsupported -> PersistNull
        TxScriptValid -> PersistBool True
        TxScriptInvalid -> PersistBool False
```

lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs
```diff
@@ -798,6 +807,29 @@ instance PersistField DerivationPrefix where
...
+    fromPersistValue PersistNull = Right TxScriptsUnsupported
+    fromPersistValue (PersistBool True) = Right TxScriptValid
+    fromPersistValue (PersistBool False) = Right TxScriptInvalid
+    fromPersistValue x =
+        Left $ T.concat [
+          "Failed to parse Haskell type `ScriptValidation`;"
+          , " expected null or boolean"
+          , " from database, but received: "
+          , T.pack (show x)
+          ]
```
jonathanknowles: Suggestion: use `LambdaCase` here for brevity.

Also, using `T.unwords` will insert spaces for you.

```suggestion
    fromPersistValue = \case
        PersistNull       -> Right TxScriptsUnsupported
        PersistBool True  -> Right TxScriptValid
        PersistBool False -> Right TxScriptInvalid
        x -> Left $ T.unwords
            [ "Failed to parse Haskell type `ScriptValidation`;"
            , "expected null or boolean"
            , "from database, but received:"
            , T.pack (show x)
            ]
```

lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs
```diff
@@ -798,6 +807,29 @@ instance PersistField DerivationPrefix where
...
+
+instance PersistFieldSql TxScriptValidity where
+    -- sqlType _ = sqlType (Proxy @ScriptValidation)
+    sqlType _ = SqlBool
```
jonathanknowles: I think you can write this as:
```suggestion
    sqlType _ = sqlType (Proxy @Bool)
```

But maybe it should be:
```suggestion
    sqlType _ = sqlType (Proxy @(Maybe Bool))
```

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -373,7 +375,12 @@ applyTxToUTxO tx !u = spendTx tx u <> utxoFromTx tx
 -- spendTx tx (filterByAddress f u) = filterByAddress f (spendTx tx u)
 -- spendTx tx (u <> utxoFromTx tx) = spendTx tx u <> utxoFromTx tx
 spendTx :: Tx -> UTxO -> UTxO
-spendTx tx !u = u `excluding` Set.fromList (inputs tx)
+spendTx tx !u =
+    u `excluding` (
+        if failedScriptValidation (tx ^. #scriptValidity)
+        then Set.fromList (collateralInputs tx)
+        else Set.fromList (inputs tx)
+    )
```
jonathanknowles: Suggestion: extract out an `inputsToExclude` definition.

Advantage: avoid the multi-line parentheses, and avoid repeating the `Set.fromList`.

```suggestion
spendTx tx !u =
    u `excluding` Set.fromList inputsToExclude
  where
    inputsToExclude =
        if failedScriptValidation (tx ^. #scriptValidity)
        then collateralInputs tx
        else inputs tx
```

specifications/api/swagger.yaml
sevanspowell: Yeah, I think that's great. Implementing now...

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -556,4 +567,11 @@ changeUTxO
     -> UTxO
 changeUTxO pending = evalState $
     mconcat
-    <$> mapM (filterByAddressM isOurAddress . utxoFromTx) (Set.toList pending)
+    <$> mapM (filterByAddressM isOurAddress . mkUTxOFromTx) (Set.toList pending)
+    where
+        -- Generate a UTxO from an transaction, assuming that it passes phase-2
+        -- script validation. Crucially, utxoFromTx will exclude failed
+        -- transactions, hence we define our own function.
```
jonathanknowles: This leaves me with a question:

For transactions in the pending set, what meaning can we ascribe to the `txScriptValidity` field? It seems that at this stage, we wouldn't have enough information to assign a meaningful value. (Unless I'm mistaken?)

The fact that we must ignore the `txScriptValidity` field in the `changeUTxO` function indicates to me that there might be a better way to represent the concept of "not-yet-known" or "unknown".

For example, we could give the `TxScriptValidity` type just two options (`TxScriptValid` and `TxScriptInvalid`), and then give `txScriptValidity` a type of `Maybe TxScriptValidity`. The `Nothing` option would represent the case where we don't yet know. But this also seems non-ideal.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -1019,8 +1028,12 @@ fromAlonzoTx
        , [W.DelegationCertificate]
        , [W.PoolCertificate]
        )
-fromAlonzoTx (Alonzo.ValidatedTx bod _wits _isValidating aux) =
-    fromAlonzoTxBodyAndAux bod aux
+fromAlonzoTx (Alonzo.ValidatedTx bod _wits (Alonzo.IsValid isValid) aux) =
+    (\(tx, d, p) -> (tx { W.scriptValidity = if isValid
+                                            then W.TxScriptValid
+                                            else W.TxScriptInvalid
+                        }, d, p))
```
jonathanknowles: Request: try to avoid using variable-length indentation. (See https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#avoid-variable-length-indentation)

Suggestion:
```suggestion
    (\(tx, d, p) -> (tx {W.scriptValidity = validity}, d, p))
    $ fromAlonzoTxBodyAndAux bod aux
  where
    validity =
        if isValid
        then W.TxScriptValid
        else W.TxScriptInvalid
```

jonathanknowles (commented):
@sevanspowell and I had a discussion about the representation of a transaction's script validity (currently represented by the `Tx.txScriptValidity :: TxScriptValidity` field).

After consideration, we both feel that it doesn't make sense to talk about the script validation status of a transaction that is still in the pending set: after all, transactions in this state have not yet been subjected to phase-2 validation, so we can't yet know what the outcome will be.

Therefore, we propose to:

- [x] Change the type of `Tx.txScriptValidity` to `Maybe TxScriptValidity`.
    (This makes it possible to assign pending transactions with a script validity of `Nothing`.)
- [x] Change the type of `TxInfo.txInfoScriptValidity` to `Maybe TxScriptValidity`.
- [x] Change the definition of `TxScriptValidity` to:
    ```hs
    data TxScriptValidity
        = TxScriptValid
        | TxScriptInvalid
    ```
- [x] Change the API definition of `txScriptValidity` to:
    ```yaml
    x-txScriptValidity: &txScriptValidity
      type: string
      enum:
        - valid
        - invalid
    ```
The absence of a validity means that either:
- script validity is not supported for this type of transaction (for example, if the transaction was created in an era prior to Alonzo); or
- this transaction has not yet been subjected to script validation.

We both feel that this representation is still not completely satisfactory. Even under the revised representation, the `Tx` type still make it possible to represent a pending transaction that has a script validity, which of course doesn't make sense.

Ideally, we'd want to represent pending transactions with a type that **_doesn't_** have a script validity field, so it would not be possible to represent this invalid state.

However, the `Tx` type is deeply embedded in multiple places, and replacing this with a `PendingTx` type would likely be a lot of work, and almost certainly cause us to exceed our time budget for this task. So we propose to:

- [x] Create an item of technical debt: represent pending transactions with a type that doesn't include script validity.

sevanspowell:
https://input-output.atlassian.net/browse/ADP-1116 

Technical debt ticket created :+1:

sevanspowell:
> @sevanspowell and I had a discussion about the representation of a transaction's script validity (currently represented by the `Tx.txScriptValidity :: TxScriptValidity` field).
>
> ...

3997bf9

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1952,8 +1958,13 @@ txHistoryFromEntity ti tip metas ins cins outs ws =
...
   where
     startTime' = interpretQuery ti . slotToUTCTime
-    mkItem m = mkTxWith (txMetaTxId m) (txMetaFee m) (txMetadata m) (mkTxDerived m)
-    mkTxWith txid mfee meta derived = do
+    mkItem m = mkTxWith
+               (txMetaTxId m)
+               (txMetaFee m)
+               (txMetadata m)
+               (mkTxDerived m)
+               (txMetaScriptValidity m)
```
jonathanknowles:
```suggestion
    mkItem m = mkTxWith
        (txMetaTxId m)
        (txMetaFee m)
        (txMetadata m)
        (mkTxDerived m)
        (txMetaScriptValidity m)
```

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1934,6 +1935,11 @@ mkTxMetaEntity wid txid mfee meta derived = TxMeta
     , txMetaFee = fromIntegral . W.unCoin <$> mfee
     , txMetaSlotExpires = derived ^. #expiry
     , txMetadata = meta
+    , txMetaScriptValidity =
+            case scriptValidity of
+                Nothing -> Nothing
+                (Just W.TxScriptValid) -> Just True
+                (Just W.TxScriptInvalid) -> Just False
```
jonathanknowles:
```suggestion
    , txMetaScriptValidity = scriptValidity <&> \case
          W.TxScriptValid -> True
          W.TxScriptInvalid -> False
```

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1979,6 +1990,11 @@ txHistoryFromEntity ti tip metas ins cins outs ws =
                 Quantity $ fromIntegral $ if tipH > txH then tipH - txH else 0
             , W.txInfoTime =
                 t
+            , W.txInfoScriptValidity =
+                case isValid of
+                    Nothing -> Nothing
+                    Just False -> Just W.TxScriptInvalid
+                    Just True -> Just W.TxScriptValid
```
jonathanknowles:
```suggestion
            , W.txInfoScriptValidity = isValid <&> \case
                True -> W.TxScriptValid
                False -> W.TxScriptInvalid
```

lib/core/test/shared/Cardano/Wallet/DummyTarget/Primitive/Types.hs
```diff
@@ -130,8 +136,9 @@ mkTx
     -> [TxOut]
     -> Map RewardAccount Coin
     -> Maybe TxMetadata
+    -> Maybe TxScriptValidity
     -> Tx
-mkTx fees ins cins outs wdrls md =
+mkTx fees ins cins outs wdrls md isValid =
```
jonathanknowles:
```suggestion
mkTx fees ins cins outs wdrls md validity =
```

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -2187,6 +2192,7 @@ instance Arbitrary (ApiTransaction n) where
             <*> arbitrary
             <*> pure txStatus
             <*> arbitrary
+            <*> arbitrary
```
jonathanknowles: Especially with these long record types, I think it's easier to figure out which generator is being used if we call it directly. (Otherwise, it requires the reader to work out which `Arbitrary` instance is being used.)
```suggestion
            <*> genTxScriptValidity
```

lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs
```diff
@@ -376,10 +379,10 @@ arbitraryChainLength = 10
...
 
 instance Arbitrary Tx where
-    shrink (Tx _tid fees ins cins outs wdrls md) =
-        [ mkTx fees ins' cins' outs' wdrls' md'
-        | (ins', cins', outs', wdrls', md') <-
-            shrink (ins, cins, outs, wdrls, md)
+    shrink (Tx _tid fees ins cins outs wdrls md isValid) =
+        [ mkTx fees ins' cins' outs' wdrls' md' isValid'
+        | (ins', cins', outs', wdrls', md', isValid') <-
+            shrink (ins, cins, outs, wdrls, md, isValid)
```
jonathanknowles: Suggestion: use a name that's consistent with the name of the type:

```suggestion
    shrink (Tx _tid fees ins cins outs wdrls md validity) =
        [ mkTx fees ins' cins' outs' wdrls' md' validity'
        | (ins', cins', outs', wdrls', md', validity') <-
            shrink (ins, cins, outs, wdrls, md, validity)
```

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
sevanspowell: 3 spaces?

lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs
```diff
@@ -388,7 +391,7 @@ instance Arbitrary Tx where
         outs <- fmap (L.take 5 . getNonEmpty) arbitrary
         wdrls <- fmap (Map.fromList . L.take 5) arbitrary
         fees <- arbitrary
-        mkTx fees ins cins outs wdrls <$> arbitrary
+        mkTx fees ins cins outs wdrls <$> arbitrary <*> arbitrary
```
jonathanknowles: I tend to think with these long record types, the calls to `arbitrary` make it harder than necessary to figure out exactly which generator is being used.

If we use the name of the generator directly, then it's easy for a reader to navigate to the source:

Suggestion:
```suggestion
        mkTx fees ins cins outs wdrls <$> arbitrary <*> genTxScriptValidity
```

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -496,6 +503,8 @@ prop_changeUTxO_inner pendingTxs =
           -- No outputs are omitted when we select everything:
         , UTxO.size utxoAll == F.sum (F.length . view #outputs <$> pendingTxs)
         ]
+    & counterexample ("size utxoAll = " <> show (UTxO.size utxoAll))
+    & counterexample ("sum outputs = " <> show (F.sum (F.length . view #outputs <$> pendingTxs)))
```
jonathanknowles: https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#limit-line-length-to-80-characters

Suggestion: use the `report` combinator, as this handles pretty-printing for you (no need to call `show`).

```suggestion
    & report                                                                                                                                                                        
        (UTxO.size utxoAll)                                                                                                                                                         
        "UTxO.size utxoAll"                                                                                                                                                         
    & report                                                                                                                                                                        
        (F.sum (F.length . view #outputs <$> pendingTxs))                                                                                                                           
        "F.sum (F.length . view #outputs <$> pendingTxs)"    
```

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
jonathanknowles: Apologies! I've edited my comment. 👍🏻

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
jonathanknowles: PR https://github.com/input-output-hk/cardano-wallet/pull/2876 moves `report` to a common location, making it easy to reuse here.

specifications/api/swagger.yaml
sevanspowell: 1531fc7

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
sevanspowell: 6c327fadfe2f1ed417ba2e05f3d200ad87ae00fa

lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs
sevanspowell: 383227a

lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs
sevanspowell: 383227a

lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs
sevanspowell: 2f5cbb5

lib/core/src/Cardano/Wallet/Primitive/Model.hs
sevanspowell: de801657f95c2097e387a83565de213271260ab6

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
sevanspowell: 3997bf9

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
sevanspowell: 047085a247c0a21339dd41a6fc98a38dfe1a8b82

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
sevanspowell: 047085a247c0a21339dd41a6fc98a38dfe1a8b82

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
sevanspowell: 047085a247c0a21339dd41a6fc98a38dfe1a8b82

lib/core/test/shared/Cardano/Wallet/DummyTarget/Primitive/Types.hs
sevanspowell: 4199c7f0925a78c92558361cad85135ee9bb85ea

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
sevanspowell: 1105933ed6bcdd7d26ca9fb5d4ea4e7e7b143459

lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs
sevanspowell: 4199c7f

lib/core/test/unit/Cardano/Wallet/DB/Arbitrary.hs
sevanspowell: 1105933ed6bcdd7d26ca9fb5d4ea4e7e7b143459

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
sevanspowell: 1105933ed6bcdd7d26ca9fb5d4ea4e7e7b143459

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -3637,3 +3639,11 @@ instance FromJSON (ApiT (Script KeyHash)) where
     parseJSON = fmap ApiT . parseJSON
 instance ToJSON (ApiT (Script KeyHash)) where
     toJSON = toJSON . getApiT
+
+instance FromJSON (ApiT TxScriptValidity) where
+    parseJSON = fmap ApiT . genericParseJSON (Aeson.defaultOptions
+        { constructorTagModifier = camelTo2 '_' . drop 8 })
```
jonathanknowles: The parentheses can be removed (as this is record update syntax):
```suggestion
    parseJSON = fmap ApiT . genericParseJSON Aeson.defaultOptions
        { constructorTagModifier = camelTo2 '_' . drop 8 }
```

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -3637,3 +3639,11 @@ instance FromJSON (ApiT (Script KeyHash)) where
...
+
+instance ToJSON (ApiT TxScriptValidity) where
+    toJSON = genericToJSON (Aeson.defaultOptions
+        { constructorTagModifier = camelTo2 '_' . drop 8 }) . getApiT
```
jonathanknowles: Likewise:

```suggestion
    toJSON = genericToJSON Aeson.defaultOptions
        { constructorTagModifier = camelTo2 '_' . drop 8 } . getApiT
```

lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs
```diff
@@ -798,6 +803,27 @@ instance PersistField DerivationPrefix where
...
+instance PersistField TxScriptValidity where
+    toPersistValue = \case
+        TxScriptValid -> PersistBool True
+        TxScriptInvalid -> PersistBool False
+
+    fromPersistValue = \case
+        PersistBool True -> Right TxScriptValid
+        PersistBool False -> Right TxScriptInvalid
+        x -> Left $ T.unwords
+            [ "Failed to parse Haskell type `ScriptValidation`;"
```
jonathanknowles:
```suggestion
            [ "Failed to parse Haskell type `TxScriptValidity`;"
```

lib/core/test/unit/Cardano/Wallet/DB/Sqlite/TypesSpec.hs
```diff
@@ -72,6 +74,7 @@ spec = do
         persistRoundtrip $ Proxy @StdGen
         persistRoundtrip $ Proxy @(Nested TxMetadata)
         persistRoundtrip $ Proxy @(Simple TxMetadata)
+        persistRoundtrip $ Proxy @TxScriptValidity
```
jonathanknowles, line 77: 👍🏻

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
```diff
@@ -459,10 +478,34 @@ data TransactionInfo = TransactionInfo
     -- ^ Creation time of the block including this transaction.
     , txInfoMetadata :: !(Maybe TxMetadata)
     -- ^ Application-specific extension data.
+    , txInfoScriptValidity :: !(Maybe TxScriptValidity)
+    -- ^ Tag indicating whether non-native scripts in this transaction are
+    -- expected to validate. This is added by the block creator when
+    -- constructing the block. May be Nothing for pre-Alonzo and pending
+    -- transactions.
```
jonathanknowles:
```suggestion
    -- ^ Tag indicating whether non-native scripts in this transaction passed
    -- validation. This is added by the block creator when constructing the
    -- block. May be 'Nothing' for pre-Alonzo and pending transactions.
```

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
```diff
@@ -522,8 +524,101 @@ fileModeSpec =  do
...
+
+                    -- Slot 300
+                    mockApply db (dummyHash "block3a") []
+                    getTxsInLedger db `shouldReturn`
+                        -- We lost 8 to collateral
+                        [ (Outgoing, 8)
+                        -- We got 12 from the faucet
+                        , (Incoming, 12)
+                        ]
+                    getAvailableBalance db `shouldReturn` 4
```
jonathanknowles, line 621: 👍🏻

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
sevanspowell: 388c812

jonathanknowles (approved):
LGTM! 👍🏻 

@sevanspowell thanks again for all your hard work on this!

Issue #2861, opened 2021-08-30
Flaky test: `/Cardano.Wallet.DB.Sqlite/State machine test (SharedState 'Mainnet SharedKey)/Sequential/`


jonathanknowles:
### Please ensure:

- [X] This is actually a flaky test already present in the code and not caused by your PR.

### Context

Run tests on `master` (9e71c1c9f96720975278cce7714cd32d59ab545f) with the following command:

```
stack build --fast --test cardano-wallet-core:test:unit --test-arguments '--match "Sequential/" --seed 1376776602'
```

### Job name

N/A

### Test case name(s)

`/Cardano.Wallet.DB.Sqlite/State machine test (SharedState 'Mainnet SharedKey)/Sequential/`

### Error message

```shell
ExceptionThrown "PersistMarshalError \"selectSourceRes: Couldn't parse field `txMetadata` from table `tx_meta`. Value out of range within the metadata item 153907: {\\\"stri
ng\\\":\\\"\\240\\144\\171\\161\\&8\\240\\147\\128\\134\\240\\173\\183\\129W\\240\\171\\190\\137L\\240\\159\\165\\169\\\\\\\\`\\240\\171\\179\\169hF\\240\\166\\151\\134\\240\\172\\18
4\\134\\240\\170\\158\\183\\240\\165\\167\\153\\240\\152\\131\\174\\240\\168\\149\\138\\240\\163\\183\\139\\&4\\240\\159\\147\\138m\\239\\165\\165a\\\"}\\nText string metadata value 
must consist of at most 64 UTF8 bytes, but it consists of 69 bytes., vals: [PersistText \\\"0059f2d6efdbc67d28fbfb96c27064654d49a4c74f781d1db516d2bc08a24d75\\\",PersistText \\\"6245b
53521f8eaf97e3d6404385b1835da0bb135\\\",PersistText \\\"in_ledger\\\",PersistInt64 0,PersistInt64 529,PersistInt64 947,PersistInt64 22861241273355126,PersistText \\\"{\\\\\\\"153907\
\\\\\\":{\\\\\\\"string\\\\\\\":\\\\\\\"\\\\68321\\\\&8\\\\77830\\\\187841W\\\\180105L\\\\129385\\\\\\\\\\\\\\\\`\\\\179433hF\\\\157126\\\\183814\\\\174007\\\\154073\\\\98542\\\\1651
[...]

PR #2862, opened 2021-08-31
Fix `TxMetaText` generators.

jonathanknowles:
### Issue Number

#2861

### Comments

This PR:

- [x] fixes generators for `TxMetadata` so that they only produce `TxMetaText` values where the UTF-8-serialized length does not exceed 64 bytes.
- [x] renames generators to `genSimpleTxMetadata` and `genNestedTxMetadata` to emphasise the difference in generated values (simple vs nested).
- [x] uses `UnicodeString` to generate `TxMetaText` content.
- [x] adds round-trip persistence tests for `genSimpleTxMetadata` and `genNestedTxMetadata`.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Gen.hs
+++ b/lib/core/src/Cardano/Wallet/Gen.hs
@@ -87,7 +87,7 @@ import Test.QuickCheck
     ( Arbitrary (..)
     , Gen
     , Positive (..)
-    , PrintableString (..)
+    , UnicodeString (..)
     , arbitrarySizedNatural
     , choose
     , elements
@@ -97,6 +97,7 @@ import Test.QuickCheck
     , resize
     , scale
     , shrinkList
+    , shrinkMap
     , sized
     , sublistOf
     , suchThat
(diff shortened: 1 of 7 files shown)


lib/core/src/Cardano/Wallet/Gen.hs
```diff
@@ -225,19 +225,32 @@ shrinkByteString bs
   where
     n = BS.length bs
 
-genText :: Gen Text
-genText =
-    (T.pack . take 32 . getPrintableString <$> arbitrary) `suchThat` guardText
+genTxMetaText :: Gen Text
+genTxMetaText =
+    (T.pack . take 32 . getPrintableString <$> arbitrary)
```
rvl: We could make this variable-length, with a hard maximum of 64 unicode characters, which is then filtered by `guardTxMetaText`.

```suggestion
    (T.pack . getUnicodeString <$> scale (min 64) arbitrary)
```

Also I believe it's fine to use `UnicodeString` instead of `PrintableString`.

lib/core/src/Cardano/Wallet/Gen.hs
```diff
@@ -225,19 +225,32 @@ shrinkByteString bs
...
+    `suchThat` guardTxMetaText
 
-shrinkText :: Text -> [Text]
-shrinkText t
+shrinkTxMetaText :: Text -> [Text]
+shrinkTxMetaText t
     | n <= 1    = []
-    | otherwise = filter guardText [ T.take (n `div` 2) t, T.drop (n `div` 2) t ]
+    | otherwise = filter guardTxMetaTextPrefix
+        [ T.take (n `div` 2) t, T.drop (n `div` 2) t ]
```
rvl: I'm pretty sure `guardText` isn't needed anymore - a relic of a previous JSON encoding of TxMetadata. So we could just re-use the `UnicodeString` shrinker.

```haskell
shrinkTxMetaText t = [T.pack s | u <- shrink (UnicodeString $ T.unpack t), let s = getUnicodeString u, length s > 0]
```

lib/core/test/unit/Cardano/Wallet/DB/Sqlite/TypesSpec.hs
```diff
@@ -136,5 +146,24 @@ instance Arbitrary TokenQuantity where
     arbitrary = genTokenQuantityFullRange
     shrink = shrinkTokenQuantityFullRange
 
+newtype Small a = Small { unSmall :: a } deriving (Eq, Show)
+newtype Large a = Large { unLarge :: a } deriving (Eq, Show)
+
+instance Arbitrary (Small TxMetadata) where
+    arbitrary = Small <$> genSmallTxMetadata
```
rvl: I looked at the definitions of `genSmallTxMetadata` and `genTxMetadata` to try and see why they couldn't be replaced with `Test.QuickCheck.scale`.

The distinction seems to be that "small" generates non-recursive metadata values from the 3 primitive metadata types, and non-small generates metadata values of any depth.

So perhaps we could fix the naming. So remove `Large` and rename Small to something like `Primitive` or `Simple`.

rvl (approved):
Thanks

lib/core/src/Cardano/Wallet/Gen.hs
jonathanknowles: Fixed in 58cc2af86cf7d3029d4676963acf065d7b63e18b.

lib/core/src/Cardano/Wallet/Gen.hs
jonathanknowles: Fixed in ab6f53b4639fa56cf6c4bc714948d7d0b31e2ebe and 58cc2af86cf7d3029d4676963acf065d7b63e18b.

lib/core/test/unit/Cardano/Wallet/DB/Sqlite/TypesSpec.hs
jonathanknowles: Fixed (hopefully) in dc709002de61c1f67f5074288341f33312d1e035.

PR #2863, opened 2021-08-31
scripts/make_release.sh: handle errors when jira command fails

rvl:
Adds a little more error handling to the release script.


Anviking (approved):
👌

PR #2866, opened 2021-09-02
Expose execution unit calculation and adjust fee estimation accordingly 

paweljakubas:
Ticket: adp-1024

We want to adjust fee calculation by adding cost stemming from plutus scripts. They have memory imprint and number of steps to execute. This is encapsulated in `ExecutionUnit` and is standarized. This was introduced to Primitives.Types. Moreover, we can extract pricing for those two resources. It is available in ledger's PParams and in wallet it is represented in `ExecutionUnitPrices`data type. Wallet's `ProtocolParameters` were extended by this. `executionUnitPrices` field is optional as only available for Alonzo era. Whole extracting from network layer, exposing in `ApiNetworkParameters` was done, along with needed changes in swagger.

There is test to be added in integration tests to show it works.

Also to be done:
1. implement `_calcScriptExecutionCost` that takes `TxBody` from `SelaedTx` and `ProtocolParameters` - after sign tx is merged *DONE*
2. Add golden unit testing for `_calcScriptExecutionCost`  *DONE*
3. Extend `TransactionCtx` and `TransactionSkeleton`  and add script execution cost field *DONE*
4. Adopt additional execution cost from script in overall fee *DONE*

How we use it?
1. get ProtocolParameters from network where now we have current `executionUnitPrices`
2. invoke from Transaction layer:
```
calcScriptExecutionCost :: ProtocolParameters -> SealedTx -> Coin 
```
It deserialise sealed tx, look at redeemers inside script data and get all execution units. then it takes execution units prices from protocol parameters and calculate cost.
3. invoke TransactionCtx with value calculated in point 2 

### Comments
Still we need to fill in (I bet Sam or Jonathan is working on that) the following details of tx sizes in `TransactionSkeleton`
```
    -- Until we actually support minting and burning, leave these as empty.
    , txScripts = []
    , txMintBurnAssets = []
```
We should probably use serialiseToCBOR on them separately, invoke ByteString.lenght and get tx sizes as a result


lib/core/src/Cardano/Wallet/Transaction.hs
```diff
@@ -123,6 +123,14 @@ data TransactionLayer k = TransactionLayer
         -- ^ Compute a minimal fee amount necessary to pay for a given selection
         -- This also includes necessary deposits.
 
+    , calcScriptExecutionCost
+        :: ProtocolParameters
+            -- Current protocol parameters
+        -> SealedTx
+            -- The constructed transaction that could contain
```
jonathanknowles: What could it contain? 😄

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -1049,8 +1067,49 @@ instance Buildable ProtocolParameters where
...
+instance ToJSON ExecutionUnitPrices where
+    toJSON ExecutionUnitPrices{priceExecutionSteps, priceExecutionMemory} =
+        object [ "step_price"  .= toRationalJSON priceExecutionSteps
+               , "memory_unit_price" .= toRationalJSON priceExecutionMemory
+               ]
+     where
+         toRationalJSON :: Rational -> Value
+         toRationalJSON r = case fromRationalRepetendLimited 20 r of
+             Right (s, Nothing) -> toJSON s
+             _                  -> toJSON r
```
Anviking, line 1105: Are we sure this dual representation is a good idea?

As a user looking at the value, having a simple scientific representation does seem nice. 

As an application integrating with the wallet API you always need to care about the rational case anyway, so the dual representations makes the JSON parsing more difficult.

But perhaps this aids downstream applications to deal with rationals correctly, if there is no easy `fromRationalRepetendLimited` in their language? E.g. when Daedalus wants to display this to the user.

If so, down the line, maybe we could have a HTTP header for controlling the precision of all rationals in the API.

lib/core/src/Cardano/Wallet/Transaction.hs
paweljakubas: done

lib/core/src/Cardano/Wallet/Primitive/Types.hs
paweljakubas: so from protocol parameters we got
```
executionUnitPrices = Just (ExecutionUnitPrices {priceExecutionSteps = 1 % 1, priceExecutionMemory = 1 % 1 )
```
and I differentiate between two cases when we have remainder = 0 and otherwise when ratio is evaluated. And we have
1. when rational is integer then number
2. when rational is not integer then denominator/nominator

Examples:
```
0.2 -> 1/5 -> denominator:5, nominator:1
2 -> 2
```

And I would slightly;-) argue it is acceptable solution. But I am open to other things. Not very strong opinion (as always from me when it comes to user facing info)

paweljakubas:
bors try

paweljakubas:
bors try

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
```diff
@@ -1949,6 +1953,13 @@ instance Arbitrary ApiNetworkParameters where
     arbitrary = genericArbitrary
     shrink = genericShrink
 
+instance Arbitrary ExecutionUnitPrices where
+    shrink = genericShrink
+    arbitrary = do
+        step <- arbitrary `suchThat` (>= 0)
```
Anviking: Can we not use
```suggestion
        step <- getNonNegative <$> arbitrary
```
to avoid `suchThat`?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -481,6 +491,32 @@ dummySkeleton inputCount outputs = SelectionSkeleton
...
+        executeUnitsM
+  where
+    prices = view #executionUnitPrices pp
+    costOfExecutiveUnits perStep perMem (W.ExecutionUnits steps mem) =
+        Coin $ ceiling $ perStep * (fromIntegral steps) + perMem * (fromIntegral mem)
+    executeUnitsM =
+        case Cardano.deserialiseFromCBOR (Cardano.AsTx Cardano.AsAlonzoEra) (getSealedTx sealedTx) of
+            Right txValid ->
+                let getScriptData (Cardano.ShelleyTxBody _ _ _ scriptdata _ _) = scriptdata
+                    getScriptData _ = error "we should not expect Cardano.ByronTxBody here"
```
Anviking, line 531:
```suggestion
                    getScriptData _ = error "_calcScriptExecutionCost: Cardano.ByronTxBody is unexpected"
```

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Network.hs
```diff
@@ -68,6 +69,13 @@ spec = describe "SHELLEY_NETWORK" $ do
         let knownEras = [minBound .. _mainEra ctx]
         let unknownEras = [minBound .. maxBound] \\ knownEras
 
+        let checkExecutionUnitPricesPresence
+                :: ApiEra
+                -> ((HTTP.Status, Either RequestException ApiNetworkParameters) -> IO ())
+            checkExecutionUnitPricesPresence = \case
+                ApiAlonzo -> expectField #executionUnitPrices (`shouldNotBe` Nothing)
+                _ -> expectField #executionUnitPrices (`shouldBe` Nothing)
```
Anviking, line 77: 👍

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -481,6 +491,32 @@ dummySkeleton inputCount outputs = SelectionSkeleton
...
+        maybe
+        (Coin 0)
+        (Coin.sumCoins . map (costOfExecutiveUnits perStep perMem))
+        executeUnitsM
+  where
+    prices = view #executionUnitPrices pp
+    costOfExecutiveUnits perStep perMem (W.ExecutionUnits steps mem) =
+        Coin $ ceiling $ perStep * (fromIntegral steps) + perMem * (fromIntegral mem)
+    executeUnitsM =
+        case Cardano.deserialiseFromCBOR (Cardano.AsTx Cardano.AsAlonzoEra) (getSealedTx sealedTx) of
```
Anviking: This will break next fork 🤔 

Should our `TransactionLayer` have a `SealedTx ->  Era -> Cardano.Tx` function, for simple re-use here?

If we don't address now, a comment / TODO would be good.

lib/core/test/unit/Cardano/Wallet/Api/TypesSpec.hs
paweljakubas: done

lib/core/src/Cardano/Wallet/Transaction.hs
```diff
@@ -159,6 +167,9 @@ data TransactionCtx = TransactionCtx
     -- ^ Transaction expiry (TTL) slot.
     , txDelegationAction :: Maybe DelegationAction
     -- ^ An additional delegation to take.
+    , txPlutusScriptExecutionCost :: Coin
+    -- ^ An execution cost of plutus script determined by execution units
+    -- and prices obtained from network
```
Anviking: There could be multiple scripts in a tx? So this would be the total cost of all the scripts?

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -481,6 +491,32 @@ dummySkeleton inputCount outputs = SelectionSkeleton
...
+    Nothing -> Coin 0
+    Just (W.ExecutionUnitPrices perStep perMem) ->
+        maybe
+        (Coin 0)
+        (Coin.sumCoins . map (costOfExecutiveUnits perStep perMem))
+        executeUnitsM
+  where
+    prices = view #executionUnitPrices pp
+    costOfExecutiveUnits perStep perMem (W.ExecutionUnits steps mem) =
+        Coin $ ceiling $ perStep * (fromIntegral steps) + perMem * (fromIntegral mem)
```
Anviking: Is it intentional to have `sum . map (ceiling . _)` rather than `ceiling . sum . map _`?

Because if the ledger is happy with `ceiling(0.1+0.1+0.1) lovelace =  1 lovelace`, rather than `3 lovelace` we should go for the cheaper option.

Anviking: I think there's risk for overflow from the use of the `Coin` constructor here:
```
λ> ceiling (18446744073709551615 % 1) :: Word64
18446744073709551615
λ> ceiling (18446744073709551616 % 1) :: Word64
0
```
I think we should at the least `error` on over/underflow and if out of bounds of `Coin`.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -481,6 +491,32 @@ dummySkeleton inputCount outputs = SelectionSkeleton
         TokenBundle.getAssets . view #tokens <$> outputs
     }
 
+_calcScriptExecutionCost
+    :: ProtocolParameters
+    -> SealedTx
+    -> Coin
+_calcScriptExecutionCost pp sealedTx = case prices of
+    Nothing -> Coin 0
```
Anviking: I tried changing the style of this logic:
```haskell
    (Just exPrices, Just units) ->
        Coin.unsafeNaturalToCoin
        . ceiling
        . sum
        $ map (costOfExecutionUnits exPrices) units
    (Nothing, Just units) ->
        -- This case probably means that the node tip is not yet in Alonzo.
        --
        -- TODO: Not sure what the calling code would look like. Should we
        -- return @Nothing@, or continue throwing? We probably don't want to
        -- return `Coin 0`, as the cost would then be underestimated.
        error "_calcScriptExecutionCost: ExecutionUnitPrices is not available"
    (_, Nothing) -> Coin 0
```
which made it easier to spot: what if there are scripts but no prices? As per the comment, I suspect `Coin 0` is the wrong fallback here.

Anviking (changes requested):
Looks sound to me overall 👍, but probably good to have another set of eyes too.

I left a few comments, but most importantly I think:
1. Theoretical Coin overflow https://github.com/input-output-hk/cardano-wallet/pull/2866#discussion_r707545052
2. Returning `Coin 0` if there is no cost model https://github.com/input-output-hk/cardano-wallet/pull/2866#discussion_r707547292

Anviking:
Q: Were the execution units from the golden tests retrieved/checked with cardano-cli?

lib/core/src/Cardano/Wallet/Primitive/Types.hs
Anviking: May I just plainly ask though: what was your motivation for
```
0.2 -> 1/5 -> denominator:5, nominator:1
2 -> 2
```
rather than
```
0.2 -> 1/5 -> denominator:5, nominator:1
2 -> denominator:1, nominator:2
```
?

lib/core/src/Cardano/Wallet/Transaction.hs
Anviking: Perhaps just:
```
Total execution cost of plutus scripts, determined by their execution units
and prices obtained from network.
```

Anviking:
_Had a passing thought about property testing this and similar work. It would be nice if there was a cheap and more general way. Is there some unexplored middle-ground between integration and unit tests? Might think more about it tomorrow_...

Anviking:
PR with a few suggestions https://github.com/input-output-hk/cardano-wallet/pull/2893

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
paweljakubas: indeed, when sign tx will be in master we will use primitives introduced there as a replacement. For the time being the above one does a trick I believe

Anviking: Ah, ok 👍

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
paweljakubas: I believe it is fine to have `sum . map (ceiling)` -> https://github.com/input-output-hk/cardano-ledger-specs/blob/3d6edf3981a0f1b762e4dfeaecd7eecb755ea928/alonzo/impl/src/Cardano/Ledger/Alonzo/Scripts.hs#L252

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
paweljakubas: adopted

lib/core/src/Cardano/Wallet/Transaction.hs
paweljakubas: yes, there can be more than one redeeemer which is the case in some examples shared with us and introduced in shelley core unit test

lib/core/src/Cardano/Wallet/Primitive/Types.hs
paweljakubas: just thought when we have integer it is redundant to have denominator. That was my thinking...also I see node team uses the same approach for presenting rational so maybe it is good to consistent with them (with balance tx endpoint we start to accept some cardano-api's types in our api as a move for better interoperability)

Anviking (approved):
Worst things that theoretically could happen should be:
1. the fee calculated for Alonzo txs is slightly wrong leading to failed txs
2. `error "_calcScriptExecutionCost: ExecutionUnitPrices is not available"` being hit

Doubt (1) is the case, but we'd notice later on if so. (2) is still better than `Coin 0` I think, and if needed we can fix later when we have more of the full picture.

So think we can merge.

PR #2869, opened 2021-09-06
jira: update endpoint URL

rvl:
We have a new jira server. Update the go-jira config and links.

### Comments

To make go-jira work again, you will need to get a new API token and update your go-jira config.

PR #2871, opened 2021-09-07
Add pass to go-jira setup instructions

sevanspowell:
Completely optional additions to https://github.com/input-output-hk/cardano-wallet/pull/2869 PR, because users can just go view the go-jira documentation.

However, I didn't realize how easy it would be to hook into my existing GPG setup, so I'm adding these instructions to make it clear that if you're already using a YubiKey & a GPG agent, it's super easy to make go-jira use this setup.

### DIFF
--- a/.jira.d/config.yml
+++ b/.jira.d/config.yml
@@ -18,6 +18,30 @@
 #    password-source: keyring
 #    password-name: input-output.atlassian.net/your.name@iohk.io
 #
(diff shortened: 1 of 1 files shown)


rvl (approved):
Yes `pass` is pretty convenient. There is also `gopass`, but I've not tried. I like the "keyring" source because keepassxc integrates with it.

PR #2877, opened 2021-09-08
Add stake key deposit parameters to runWalletCoinSelection

rvl:
This is the coin selection part extracted from PR #2754.

The goal is to remove certificate deposit / withdrawal hacks from `Cardano.Wallet.selectAssets` and move the logic towards the coin selection module.

There are some Cardano.Wallet.Util and logging improvements also brought across, which are used by `selectAssets` and `runWalletCoinSelection`.

### Comments

There is a new `CoinSelectionSpec` module with a property for `runWalletCoinSelection`, but it's not able to handle all cases yet.

Based on the PR #2887 branch - merge that first.

### Issue Number

ADP-919

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
@@ -52,7 +52,7 @@ import Cardano.Wallet.Primitive.CoinSelection.Balance
 import Cardano.Wallet.Primitive.Types.Address
     ( Address (..) )
 import Cardano.Wallet.Primitive.Types.Coin
-    ( Coin (..) )
+    ( Coin (..), addCoin, scaleCoin, subtractCoin )
 import Cardano.Wallet.Primitive.Types.TokenBundle
     ( TokenBundle )
 import Cardano.Wallet.Primitive.Types.TokenMap
@@ -74,11 +74,15 @@ import Control.Monad.Random.Class
 import Control.Monad.Trans.Except
     ( ExceptT (..), withExceptT )
 import Data.Generics.Internal.VL.Lens
-    ( view )
+    ( over, set, view )
 import Data.Generics.Labels
     ()
+import Data.List
+    ( foldl', partition )
 import Data.List.NonEmpty
     ( NonEmpty (..) )
+import Data.Maybe
+    ( fromMaybe )
 import Data.Word
     ( Word16 )
 import GHC.Generics
@@ -117,11 +121,12 @@ runWalletCoinSelection sc sp = do
 type SelectionReturn = Either SelectionError (SelectionResult TokenBundle)
 
 makeWrapper
-    :: SelectionConstraints
+    :: HasCallStack
+    => SelectionConstraints
     -> SelectionParams
     -> Either ErrPrepareOutputs (SelectionReturn -> SelectionReturn, PerformSelection)
 makeWrapper sc@SelectionConstraints{..} SelectionParams{..} =
-    (id,) . getArgs <$> prepareOutputs outputsToCover
+    (fmap repairOutputs,) . getArgs <$> prepareOutputs outputsToCover
   where
     -- TODO: [ADP-1037] Adjust coin selection and fee estimation to handle
     -- collateral inputs.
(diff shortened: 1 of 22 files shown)


lib/core/src/Cardano/Wallet.hs
```diff
@@ -2836,10 +2835,8 @@ data WalletFollowLog
 
 -- | Log messages from API server actions running in a wallet worker context.
 data WalletLog
-    = MsgSelectionStart UTxOIndex (NonEmpty TxOut)
-    | MsgSelectionError SelectionError
-    | MsgSelectionReportSummarized SelectionReportSummarized
-    | MsgSelectionReportDetailed SelectionReportDetailed
+    = MsgSelectAssets UTxOIndex TransactionCtx [TxOut] (BracketLog' (Either ErrSelectAssets (SelectionResult TokenBundle)))
```
jonathanknowles: Suggestion: extract this collection of fields out as a separate record type.

Justification: the types alone don't make it obvious what each field is for.

So something like:
```suggestion
    = MsgSelectAssets SelectAssetsInfo
```
and:
```hs
data SelectAssetsInfo = SelectAssetsInfo
    { utxoAvailable
        :: UTxOIndex
    , outputsRequested
        :: [TxOut]
    , transactionCtx
        :: TransactionCtx
    , logResult
        :: BracketLog (Either ErrSelectAssets (SelectionResult TokenBundle))
    }

File: lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs [hunk 2]
```diff
@@ -89,51 +102,82 @@ import qualified Data.Set as Set
...
+    -- list, so validate for this precondition.
+    ensureNonEmptyOutputs =
+        maybe (Left ErrPrepareOutputsTxOutMissing) Right . NE.nonEmpty
+
+    addExtraValueSinkToOutputs os
+        | null os && extraValueSink /= mempty =
+            (TxOut dummyAddress extraValueSink : os)
+        | null os =
+            (TxOut dummyAddress dummyValue : os)
+        | otherwise = os
```
jonathanknowles, line 174: Hi @rvl. It seems impossible for this function to return an empty list, since:

- The first two cases construct a non-empty list;
- The last case must also be non-empty, as the second pattern failed to match.

If this is intended, then this transformation is possible:

f83c68fe6a0928cb52da53d3598b3f6a63f1a9b1

This removes the need for `ensureNonEmptyOutputs` and `ErrPrepareOutputsTxOutMissing`.

But perhaps I've misunderstood something?

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Balance.hs
```diff
@@ -510,15 +515,6 @@ performSelection minCoinFor costFor bundleSizeAssessor criteria
...
                 }
   where
-    SelectionCriteria
-        { outputsToCover
-        , utxoAvailable
-        , selectionLimit
-        , extraCoinSource
-        , assetsToMint
-        , assetsToBurn
-        } = criteria
```
jonathanknowles: Hi @rvl!

I wonder if we can avoid using `RecordWildCards` here?

My personal experience is that with this extension, I find it much harder to determine where symbols are introduced and consumed, especially in long functions such as this one. I suspect I'm not alone in this! With `NamedFieldPuns`, it's a simple matter to find out where things are introduced and consumed, even if it is a bit annoying to have to list out the fields you want.

Haskell Language Server might make this easier, but that's not available in all contexts (especially in a code review context).

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -1373,61 +1363,10 @@ newtype Signature (what :: Type) = Signature { getSignature :: ByteString }
...
-    :: String
-        -- ^ The message
-    -> a
-        -- ^ The value to test
-    -> (a -> Bool)
-        -- ^ The predicate
-    -> a
-invariant msg a predicate =
-    if predicate a then a else error msg
-
```
jonathanknowles, line 1407: Nice idea to move these utility functions to a dedicated module.

(The `Types` module is already bloated enough, so moving these out is good IMO.)

jonathanknowles:
I think we should be okay to close this PR for the moment, since:

- stake key deposit parameters have already been merged to `master` in previous PRs.
- the workaround for the "no outputs" case has already been removed. (See https://github.com/input-output-hk/cardano-wallet/pull/2917)

We can reopen if we find that we've missed anything.

PR #2883, opened 2021-09-10
Allow coin selection with pre-selected inputs

jonathanknowles:
## Issue Number

ADP-1118

## Summary and Scope

This PR makes it possible to call `CoinSelection.performSelection` with an available UTXO set that has some entries pre-selected. The pre-selected entries will be included in the result.

In addition, this PR:
- renames `SelectionInsufficientError` to `SelectionLimitReachedError` to clarify that this error condition relates to the selection limit.

## Limitations

This PR only adjusts the API provided by `CoinSelection` (and `CoinSelection.Balance`). It does not adjust `Wallet.selectAssets` to accept a `UTxOSelection`. This is left to future PRs.

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
@@ -161,7 +161,6 @@ import Test.Integration.Framework.TestData
     , errMsg403MinUTxOValue
     , errMsg403NotAShelleyWallet
     , errMsg403NotEnoughMoney
-    , errMsg403TxTooBig
     , errMsg403WithdrawalNotWorth
     , errMsg403WrongPass
     , errMsg404CannotFindTx
@@ -1112,7 +1111,7 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
             (Link.createTransactionOld @'Shelley wa) Default payload
 
         expectResponseCode HTTP.status403 r
-        expectErrorMessage errMsg403TxTooBig r
+        expectErrorMessage errMsg403EmptyUTxO r
 
     it "TRANSMETA_ESTIMATE_01 - fee estimation includes metadata" $ \ctx -> runResourceT $ do
         (wa, wb) <- (,) <$> fixtureWallet ctx <*> emptyWallet ctx
@@ -1175,7 +1174,7 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
 
         verify r
             [ expectResponseCode HTTP.status403
-            , expectErrorMessage errMsg403TxTooBig
+            , expectErrorMessage errMsg403EmptyUTxO
             ]
 
     describe "TRANS_ESTIMATE_08 - Bad payload" $ do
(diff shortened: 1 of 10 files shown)


lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
```diff
@@ -199,13 +243,98 @@ data SelectionError
...
+
+        modifyExistingInputs :: UTxO -> UTxO
+        modifyExistingInputs = const UTxO.empty
+
+        modifyRewardWithdrawal :: Maybe Coin -> Maybe Coin
+        modifyRewardWithdrawal =
+            Just . (view #coin existingInputValue <>) . fromMaybe mempty
+
+        modifyUTxOAvailable :: UTxOIndex -> UTxOIndex
+        modifyUTxOAvailable = UTxOIndex.deleteMany $ UTxO.dom existingInputs
```
sevanspowell: Oh, didn't think about that, an existing input means that we've effectively lost that input from our UTxO set.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
```diff
@@ -199,13 +243,98 @@ data SelectionError
...
+            modifyAssetsToMint
+        . over #existingInputs
+            modifyExistingInputs
+        . over #rewardWithdrawal
+            modifyRewardWithdrawal
+        . over #utxoAvailable
+            modifyUTxOAvailable
+      where
+        modifyAssetsToMint :: TokenMap -> TokenMap
+        modifyAssetsToMint = TokenMap.add (view #tokens existingInputValue)
```
sevanspowell: Why are the existing input values added to assets to mint?

lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
jonathanknowles: > Oh, didn't think about that, an existing input means that we've effectively lost that input from our UTxO set.

Yes that's right! We definitely shouldn't reselect it.

By removing it here, it means that we don't have to modify the internal algorithm: the internal algorithm can assume that everything in `availableUTxO` really is available for selection.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
jonathanknowles: Good question.

When we have pre-existing inputs, these are effectively "supplying extra value" to a coin selection.

The `assetsToMint` field currently provides the only well-tested way of supplying extra value.

By adding to this field here, we're effectively getting the ability to "add the value from several extra inputs" for free. We can also take advantage of the fact that the `assetsToMint` and `assetsToBurn` fields are already well covered with property tests. In a sense, we're treating the inner algorithm as a black box that we trust will do the right thing.

Of course, it would be nicer if the `assetsToMint` and `assetsToBurn` field names could reflect their generality. So we have a ticket for precisely that: ADP-1117

sevanspowell (approved):

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Transactions.hs
```diff
@@ -1112,7 +1111,7 @@ spec = describe "SHELLEY_TRANSACTIONS" $ do
             (Link.createTransactionOld @'Shelley wa) Default payload
 
         expectResponseCode HTTP.status403 r
-        expectErrorMessage errMsg403TxTooBig r
+        expectErrorMessage errMsg403EmptyUTxO r
```
piotr-iohk, line 1114: This doesn't seem right... `fixtureWallet` has UTxOs cc: @jonathanknowles.

PR #2884, opened 2021-09-10
Don't require all minted assets to be spent or burned

jonathanknowles:
### Issue Number

ADP-1113

### Summary

This PR removes the restriction that all minted assets must be spent or burned.

Minted assets that are not spent or burned are now returned in the change outputs.


sevanspowell:
Yay! This is great :) Very cathartic.

sevanspowell (approved):

PR #2885, opened 2021-09-10
Add generators and shrinkers for `SelectionLimit` and `SelectionSkeleton`.

jonathanknowles:
### Comments

This PR adds generators and shrinkers for `SelectionLimit` and `SelectionSkeleton`.

### Issue Number

ADP-1118


lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Balance.hs
```diff
@@ -233,12 +235,21 @@ emptySkeleton = SelectionSkeleton
...
+data SelectionLimitOf a
     = NoLimit
       -- ^ Indicates that there is no limit.
-    | MaximumInputLimit Int
+    | MaximumInputLimit a
       -- ^ Indicates a maximum limit on the number of inputs to select.
-    deriving (Eq, Show)
+    deriving (Eq, Functor, Show)
+
+instance Ord a => Ord (SelectionLimitOf a) where
```
sevanspowell, line 247: Just to confirm, the Ord instance of SelectionLimitOf treats NoLimit is treated like "infinity"? If so, this instance makes sense.

sevanspowell (approved):

sevanspowell:
Looks good to me :+1:

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Balance.hs
jonathanknowles: Yes, that's correct!

For completeness (and to satisfy the `Ord` laws), we must also state that `NoLimit == NoLimit`.

jonathanknowles:
bors retry

PR #2887, opened 2021-09-10
Split runWalletCoinSelection wrapper into a separate function

rvl:
The `runWalletCoinSelection` function calls the generic `performSelection` function.

This PR splits the preparation of `performSelection` parameters into a new wrapper function.

### Issue Number

ADP-919

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -223,7 +223,9 @@ import Cardano.Wallet.Logging
     ( BracketLog
     , BracketLog' (..)
     , bracketTracer
-    , resultTracer
+    , formatResultMsgWith
+    , resultSeverity
+    , traceResult
     , traceWithExceptT
     , unliftIOTracer
     )
@@ -294,15 +296,13 @@ import Cardano.Wallet.Primitive.AddressDiscovery.Shared
     , isShared
     )
 import Cardano.Wallet.Primitive.CoinSelection
-    ( ErrPrepareOutputs (..)
+    ( ErrWalletSelection (..)
     , SelectionConstraints (..)
-    , SelectionError (..)
     , SelectionParams (..)
-    , performSelection
+    , runWalletCoinSelection
     )
 import Cardano.Wallet.Primitive.CoinSelection.Balance
-    ( SelectionReportDetailed
-    , SelectionReportSummarized
+    ( SelectionError (..)
     , SelectionResult (..)
     , UnableToConstructChangeError (..)
     , emptySkeleton
@@ -482,7 +482,7 @@ import Data.Void
 import Data.Word
     ( Word64 )
 import Fmt
-    ( blockListF, pretty, (+|), (+||), (|+), (||+) )
+    ( blockListF, build, fmt, nameF, pretty, (+|), (+||), (|+), (||+) )
 import GHC.Generics
     ( Generic )
 import Safe
@@ -500,7 +500,6 @@ import qualified Cardano.Crypto.Wallet as CC
 import qualified Cardano.Wallet.Primitive.AddressDiscovery.Random as Rnd
 import qualified Cardano.Wallet.Primitive.AddressDiscovery.Sequential as Seq
 import qualified Cardano.Wallet.Primitive.AddressDiscovery.Shared as Shared
-import qualified Cardano.Wallet.Primitive.CoinSelection.Balance as Balance
 import qualified Cardano.Wallet.Primitive.Migration as Migration
 import qualified Cardano.Wallet.Primitive.Types as W
 import qualified Cardano.Wallet.Primitive.Types.Coin as Coin
@@ -1394,7 +1393,7 @@ selectAssetsNoOutputs
     -> ExceptT ErrSelectAssets IO result
 selectAssetsNoOutputs ctx wid wal tx transform = do
     -- NOTE:
-    -- Could be made nicer by allowing 'performSelection' to run with no target
+    -- Could be made nicer by allowing 'runWalletCoinSelection' to run with no target
     -- outputs, but to satisfy a minimum Ada target.
     --
     -- To work-around this immediately, I am simply creating a dummy output of
@@ -1402,7 +1401,8 @@ selectAssetsNoOutputs ctx wid wal tx transform = do
     -- result. The resulting selection will therefore have a delta that is at
     -- least the size of the deposit (in practice, slightly bigger because this
     -- extra outputs also increases the apparent minimum fee).
-    deposit <- calcMinimumDeposit @_ @s @k ctx wid
+    deposit <- withExceptT ErrSelectAssetsNoSuchWallet $
+        calcMinimumDeposit @_ @s @k ctx wid
     let dummyAddress = Address ""
     let dummyOutput  = TxOut dummyAddress (TokenBundle.fromCoin deposit)
     let outs = dummyOutput :| []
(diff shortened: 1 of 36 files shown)


lib/core/src/Cardano/Wallet.hs
```diff
@@ -2842,10 +2835,8 @@ data WalletFollowLog
 
 -- | Log messages from API server actions running in a wallet worker context.
 data WalletLog
-    = MsgSelectionStart UTxOIndex (NonEmpty TxOut)
-    | MsgSelectionError SelectionError
-    | MsgSelectionReportSummarized SelectionReportSummarized
-    | MsgSelectionReportDetailed SelectionReportDetailed
+    = MsgSelectAssets UTxOIndex TransactionCtx [TxOut] (BracketLog' (Either ErrSelectAssets (SelectionResult TokenBundle)))
```
jonathanknowles, line 2838: My comment from the previous review applies here:

https://github.com/input-output-hk/cardano-wallet/pull/2877#discussion_r705265581

jonathanknowles (commented):
Hi @rvl 

The following commits seem totally fine, no issues:
- ✅  f4f4840271795a923d7b7fcd43c12523f8f1d9a2
- ✅ 3b35cc4050cfd84033889c2b7d7faa0541409753

**Regarding ba2ad3220fd9db724283fd894e42a3250cd00b60:**

I've read through the wrapper function, and I believe I understand how it works. I'm okay to give it ✅ on the basis that:
- we have tested the case where it succeeds;
- test coverage for the failure modes can be improved in further PRs;
- merging this PR will unblock other urgent work.

**Regarding ba2ad3220fd9db724283fd894e42a3250cd00b60:**

Since I'm at the end of the time I have available this evening, I'm not really able to check this. Though I'd be happy to look at this again when I'm back. But if you can get someone to check over it, then I'm fine with it.

lib/core/src/Cardano/Wallet.hs
rvl: For normal data types it is reasonable to use record fields, but these types are just for logging. The message constructors are always used in the vicinity of the type definition. We want the definitions to be compact and unobtrusive, and we want it to be easy to add new log message constructors.

lib/core/src/Cardano/Wallet/Logging.hs
```diff
@@ -148,86 +160,187 @@ stdoutTextTracer = Tracer $ liftIO . B8.putStrLn . T.encodeUtf8 . toText
...
+    ]
+
+-- | A good default mapping of message severities for 'traceResult'.
+resultSeverity :: Severity -> BracketLog' (Either e r) -> Severity
+resultSeverity base = \case
+    BracketStart -> base
+    BracketFinish (Left _) -> Error
+    BracketFinish (Right _) -> base
+    BracketException _ -> Error
+    BracketAsyncException _ -> base
```
sevanspowell, line 218: Hmm, why does this return `base` instead of error? What about the asynchronous nature of the exception encourages this distinction?

lib/core/src/Cardano/Wallet.hs
sevanspowell: In general, I agree with Jonathan here. I have no idea how to produce and consume a value of this type, and I have no helpers or commentary to guide me, and adding helpers or commentary is hindered by the fact that it's all embedded. But if you truly feel the benefits outweigh the costs, and that this will only be used in a limited context, then I don't mind*.

*until I have to touch it

sevanspowell:
I'm a bit foggy on the "why". What motivated this change?

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Balance.hs
```diff
@@ -281,6 +283,10 @@ data SelectionResult change = SelectionResult
     }
     deriving (Generic, Eq, Show)
 
+-- | Make a transaction input resolver from the coin selection.
+inputFromSelection :: SelectionResult change -> TxIn -> Maybe TxOut
+inputFromSelection cs = flip lookup $ NE.toList $ view #inputsSelected cs
```
jonathanknowles, line 288: This function seems a bit specific to include in the `Balance` module. Perhaps it could be defined locally, where needed?

(I suspect it's currently a weed though -- I couldn't find a place where it's used in this PR.)

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Balance.hs
```diff
@@ -445,6 +451,23 @@ prepareOutputsWith minCoinValueFor = fmap $ \out ->
         then bundle { coin = minCoinValueFor (view #tokens bundle) }
         else bundle
 
+-- | Parameters of the 'performSelection' function.
+data PerformSelection = PerformSelection
```
jonathanknowles, line 455:
```suggestion
data PerformSelectionParams = PerformSelectionParams
```

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Balance.hs
```diff
@@ -1115,17 +1120,6 @@ makeChange criteria
...
-    MakeChangeCriteria
-        { minCoinFor
-        , bundleSizeAssessor
-        , requiredCost
-        , extraCoinSource
-        , inputBundles
-        , outputBundles
-        , assetsToMint
-        , assetsToBurn
-        } = criteria
```
jonathanknowles, line 1127: This change seems superfluous to this PR.

Could it be removed?

The risk of superfluous changes is that we increase the risk of merge conflicts later on.

I'm not convinced that the use of `RecordWildCards` helps here, for the same reasons stated in previous PRs. There are so many identifiers in use within this module that using record wild cards makes it harder for readers to figure out where bindings have been introduced.

lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
```diff
@@ -192,20 +217,19 @@ data SelectionParams = SelectionParams
...
+-- failed.
+data ErrWalletSelection
+    = ErrWalletSelectionOutputs ErrPrepareOutputs
+    | ErrWalletSelectionBalance SelectionError
     deriving (Eq, Show)
 
 -- | Prepares the given user-specified outputs, ensuring that they are valid.
---
-prepareOutputs
+validateTxOutsForSelection
```
jonathanknowles, line 228: Suggestion: use `prepareTxOutsForSelection`.
Justification:
1. This function doesn't just perform validation: it also adjusts the contents of the outputs.
2. We've used "prepare" everywhere else. (`ErrPrepareOutputs` for example.)
```suggestion
prepareTxOutsForSelection
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1496,50 +1502,48 @@ selectAssets
...
+                , computeSelectionLimit =
+                    view #computeSelectionLimit tl pp txCtx
+                , maximumCollateralInputCount =
+                    view #maximumCollateralInputCount pp
+                , depositAmount =
+                    view #stakeKeyDeposit pp
+                }
+            SelectionParams
+                { assetsToBurn = TokenMap.empty -- TODO: [ADP-861]
+                , assetsToMint = TokenMap.empty -- TODO: [ADP-861]
```
jonathanknowles, line 1527: I think this should probably be ADP-346.
```suggestion
                { assetsToBurn = TokenMap.empty -- TODO: [ADP-346]
                , assetsToMint = TokenMap.empty -- TODO: [ADP-346]
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1388,15 +1393,16 @@ selectAssetsNoOutputs
...
     -- outputs, but to satisfy a minimum Ada target.
     --
     -- To work-around this immediately, I am simply creating a dummy output of
     -- exactly the required deposit amount, only to discard it on the final
     -- result. The resulting selection will therefore have a delta that is at
     -- least the size of the deposit (in practice, slightly bigger because this
     -- extra outputs also increases the apparent minimum fee).
-    deposit <- calcMinimumDeposit @_ @s @k ctx wid
+    deposit <- withExceptT ErrSelectAssetsNoSuchWallet $
+        calcMinimumDeposit @_ @s @k ctx wid
```
jonathanknowles, line 1405: Hi @rvl, are you planning to move the `dummyOutput` logic into the `CoinSelection` module as part of this PR?

jonathanknowles (commented):
Hi @rvl 

I've had another look, and left some more comments.

I think that this PR might become easier to finish once we've made progress with exposing a `extraCoinSink` parameter. (Which will be tackled as part of a revised ADP-1117.)

jonathanknowles:
I think we should be able to close this PR for the moment, since:

- the only commit that hasn't been merged is https://github.com/input-output-hk/cardano-wallet/pull/2887/commits/47461b60a860abfbdd23acded0875f30dc1a7e2a;
- the functionality provided by this commit has already been provided by other PRs that are already merged to `master`.

PR #2889, opened 2021-09-10
tests: Add --env NAME=VALUE option to hspec runners

rvl:
Adds extra option handling to the integration test suite.
Environment variables can be set from the command line, which can be helpful.


Anviking (approved):
👌  If you really want `pure: true` — sure.

A bit torn on `hspecMain`-style functions hiding custom magic without being explicit about it.

rvl:
Perhaps I could put our magic `hspecMain` into a different module? It's nice to have the common setup code for all test suites in one place.

bors r+

PR #2891, opened 2021-09-12
Simplify handling of `extraCoinSource` within `Balance` module.

jonathanknowles:
## Issue Number

ADP-1113

## Overview

This PR:
- simplifies the handling of `extraCoinSource` within the `Balance` module.
- clarifies the handling of the case where we produce an empty selection (one where we have selected no inputs).
- provides a set of functions to calculate UTXO balance sufficiency, and uses them to reduce code duplication in `Balance` and `BalanceSpec`.

## Details

Many parts of the `Balance` module give special consideration to the "extra coin source", by performing various calculations to add and subtract it from the balance. It's considered in all of the following places:

- `performSelection`
- `runSelection`
- `makeChange`

There's quite a bit of code duplication. This makes introducing further changes riskier than necessary.

In reality, a finished selection must **_always_** include at least one input. So it's safe to assume that the extra coin source can always be completely spent.

Therefore, it's unnecessary to even consider the extra coin source when selecting UTxO entries within `runSelection`, provided that `runSelection` is called with the correct balance to cover, and provided we ensure that `runSelection` always produces at least one input.

### DIFF
--- a/lib/core-integration/src/Test/Integration/Framework/TestData.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/TestData.hs
@@ -45,6 +45,7 @@ module Test.Integration.Framework.TestData
     , errMsg403NotAByronWallet
     , errMsg403NotAnIcarusWallet
     , errMsg403NotEnoughMoney
+    , errMsg403EmptyUTxO
     , errMsg403WrongPass
     , errMsg403AlreadyInLedger
     , errMsg404NoSuchPool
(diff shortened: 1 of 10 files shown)


sevanspowell:
From what I've seen so far, this looks good. I appreciate the reduction in cognitive overhead by removing the special handling around extraCoinSource.

jonathanknowles:
@sevanspowell and I had a review call, where we walked through all the commits in this PR.

A couple of suggestions were made:
- [x] We should add commentary to each of the cases in `prop_runSelectionNonEmpty_inner`, to explain the reasoning for each of the expectations. (Added in 730d4a224e453868f1b664c2afb3313014983781.)
- [x] We should extract out a function for `selectionDelta` (computing the difference between total input value and total output value) and reuse it from the test suite. We decided that this could be handled in a future PR. (Added to the scope of ADP-1117.)

sevanspowell:
Thank you for breaking this PR up into logical commits! I'm very convinced that this PR is correct :heavy_check_mark:

sevanspowell (approved):

PR #2894, opened 2021-09-14
Add docker-compose smoke test into GH actions for Linux and MacOS

piotr-iohk:
- [ ] Trying to reproduce the issue and as a result adding GH actions workflows for testing docker-compose.yml on Linux and MacOS

### Comments

Cannot add workflow for Windows as Linux based containers are not supported on Windows VMs in GH actions (https://github.com/actions/virtual-environments/issues/252)

### Issue Number

ADP-1115

### DIFF
--- a/.github/workflows/docker_linux.yml
+++ b/.github/workflows/docker_linux.yml
@@ -1,0 +1,30 @@
+name: Docker-compose Linux
+on:
+  schedule:
+  - cron:  "0 23 * * *"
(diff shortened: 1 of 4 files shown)


rvl (approved):
Good idea.

piotr-iohk:
Workflows are scheduled nightly which should be enough, it's possible also to run by hand which can be useful on releases.

PR #2895, opened 2021-09-14
Logging and test improvements

rvl:
A few miscellaneous improvements to logging and tests, used for the signTx feature.

### Comments

The individual commit messages should explain things well enough.

### Issue Number

ADP-919

PR #2896, opened 2021-09-14
Change SealedTx to have dual representations, stub signTx endpoint

rvl:
This improves the `SealedTx` type so that it can be used in either bytestring or Cardano.Tx form, whichever is more convenient.

### Comments

A couple tests needed to be disabled for this change to be separated from PR #2754, but they will be put back in a later PR.

Based on the PR #2895 branch - merge that first.

### Issue Number

ADP-919


paweljakubas (approved):
lgtm! Target `SealedTx` type and multitude of transformations introduced. A number of things stubbed, but it is ok. In the finalizing PR there will be a chance to add more changes. For the time being this PR should unlock many other PRs.

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
```diff
@@ -445,8 +605,7 @@ newtype SerialisedTx = SerialisedTx { payload :: ByteString }
 -- incomplete set of serialised witnesses, along with an encoding of the
```
sevanspowell:
```suggestion
-- incomplete set of serialised witnesses.
```

sevanspowell (approved):
It's very difficult for me to say "this is correct and I'm convinced it hasn't broken anything", but at the very least this looks sane :heavy_check_mark:

lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
```diff
@@ -1487,6 +1512,9 @@ instance Malformed (BodyParam (ApiBytesT 'Base64 SerialisedTx))
 -- no cases here as all bad requests are served by ErrDecodeSignedTxWrongPayload
 -- in Server.hs. Tested by integration tests.
 
+instance Malformed (BodyParam (ApiT SealedTx)) where
+    malformed = [] -- fixme: add a couple things
```
sevanspowell, line 1516: Can this be filled in at this stage?

lib/core/test/unit/Cardano/WalletSpec.hs
```diff
@@ -613,40 +611,7 @@ walletKeyIsReencrypted
...
-        , outputsCovered =
-            [ TxOut (Address "destination") (TokenBundle.fromCoin $ Coin 14) ]
-        , changeGenerated =
-            [ (TokenBundle.fromCoin $ Coin 1) ]
-        , utxoRemaining =
-            UTxOIndex.empty
-        }
-
-    ctx = defaultTransactionCtx
+walletKeyIsReencrypted (_wid, _wname) (_xprv, _pwd) _newPwd = property True
```
sevanspowell, line 613: Is there a JIRA ticket logged for re-implementing this?

rvl: There is now: ADP-1132.

sevanspowell: Thanks!

lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
sevanspowell: I think this might be more difficult that it seems, only an OctetStream is accepted by the endpoint. I'm not sure BodyParam has been designed around this use-case. Investigating further.

sevanspowell: The default request body for this machinery is to use JSON as the content type:

```
defaultApiRequest :: Request
defaultApiRequest = defaultRequest
    { requestHeaders =
        [ (hContentType, "application/json")
        , (hAccept, "*/*")
        ]
    }
```

I think I need to modify:

```
    type MkBodyRequest (ReqBody '[ct] a :> sub) = BodyParam a -> IO (MkBodyRequest sub)
    gEveryBodyParam _ req b =
        gEveryBodyParam (Proxy @sub) <$> (setRequestBody b req)
```

To pattern match on `OctetStream` on the type-level and set the content-type accordingly.

Not sure on how to do that yet.

sevanspowell: This seems to be working, so glad I did some research on Servant a little while ago:

```
instance {-# OVERLAPPING #-}
    ( GEveryEndpoints sub
    ) => GEveryEndpoints (ReqBody '[OctetStream] a :> sub)
  where
    gEveryEndpoint _ =
        gEveryEndpoint (Proxy @sub)

    type MkPathRequest (ReqBody '[OctetStream] a :> sub) = MkPathRequest sub
    gEveryPathParam _ =
        gEveryPathParam (Proxy @sub)

    type MkBodyRequest (ReqBody '[OctetStream] a :> sub) = BodyParam a -> IO (MkBodyRequest sub)
    gEveryBodyParam _ req b =
        gEveryBodyParam (Proxy @sub) <$> (setRequestBody b req {
            requestHeaders = [ (hContentType, "application/octet-stream")
                             , (hAccept, "*/*")
                             ]
        })

    type MkHeaderRequest (ReqBody '[OctetStream] a :> sub) = Header "Content-Type" OctetStream -> MkHeaderRequest sub
    gEveryHeader _ req (Header h) =
        gEveryHeader (Proxy @sub) $ req
            { requestHeaders = requestHeaders req ++ [(hContentType, h)] }

```

rvl: Great, so if we put a few examples like this:
```haskell
malformed = [("", _todo1), ("hello", _todo2), ("cafecafe", _todo3)]
```
then the tests pass??

lib/core/src/Cardano/Wallet/Transaction.hs
```diff
@@ -42,33 +43,39 @@ import Cardano.Wallet.Primitive.AddressDerivation
...
     ( NonEmpty )
 import Data.Text
     ( Text )
+import Fmt
+    ( Buildable (..), genericF )
 import GHC.Generics
     ( Generic )
 
-data TransactionLayer k = TransactionLayer
+data TransactionLayer k tx = TransactionLayer
```
Anviking, line 78: I am intrigued about the direction of this `tx` parameter, but it seems you are not really leveraging it in this pr, right? There is no `mapTx :: (tx1 -> tx2) -> TransactionLayer k tx1 -> TransactionLayer k tx2` or similar, and you only use it with `SealedTx`?

rvl: Yes, you are right - this type parameter isn't used by this PR. I am hoping to be able to:
1. Swap the SealedTx type for a mock type in unit testing.
2. Remove the dependency on cardano-api in cardano-wallet-core.

rvl:
Added weeder roots for temporarily unused functions.

bors r+

PR #2897, opened 2021-09-15
Unify `selectionDelta` calculations.

jonathanknowles:
## Issue Number

ADP-1070

## Summary

This PR:
- introduces the `SelectionDelta` type, which encodes the possibility of:
  - a `SelectionSurplus`
    (where all output assets are paid for by the inputs)
  - a `SelectionDeficit`
    (where some output assets are not paid for by the inputs)
- introduces the `selectionDelta{AllAssets,Coin}` functions, which compute the selection delta for all assets, or just the ada asset, respectively.
- uses the `selectionDelta{AllAssets,Coin}` functions to unify several repeated calculations relating to deltas, including within the test suite.

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -2052,6 +2052,8 @@ migrationPlanToSelectionWithdrawals plan rewardWithdrawal outputAddressesToCycle
             , utxoRemaining = UTxOIndex.empty
             , extraCoinSource
             , changeGenerated = []
+            , assetsToMint = TokenMap.empty
+            , assetsToBurn = TokenMap.empty
             }
 
         -- NOTE:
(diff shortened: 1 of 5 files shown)


jonathanknowles:
@sevanspowell and I had a call to discuss this PR.

We identified the following tasks:
- [x] Add TODOs to all `assetsToMint` and `assetsToBurn` fields that we will need to update when ADP-347 is eventually tackled.
    (Fixed in 0d37b6f77182d354a8f5a06f29636328950eb0ac)
- [x] Make the `selectionDelta` function a synonym of `selectionSurplusCoin`, and add a `TODO` on `selectionDelta` that advises callers to switch to `selectionSurplusCoin`. (Note that we can't actually use GHC's `DEPRECATED` annotation, since this will cause the build to fail.
    (Fixed in efac20e4bd32103aceaad515f82ecc3dceea4ff7)

In a future PR:
- Find a way to unify all the other delta calculations. The obstacle preventing us from doing this now is that the other delta calculations do not operated on a `SelectionResult`. So further work will be required to unify them.

sevanspowell (approved):

PR #2900, opened 2021-09-15
Balance tx endpoint skeleton

paweljakubas:
adp-656

Sign tx Pr is prerequisite for this. 

This PR is rebased after sign tx PR landed into master and golden json are regenerated. 
The following test suites pass:
```
stack test cardano-wallet:unit
stack test cardano-wallet-core:unit
```


lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -3429,6 +3441,19 @@ instance IsServerError ErrConstructTx where
             apiError err501 NotImplemented
                 "This feature is not yet implemented."
 
+instance IsServerError ErrBalanceTx where
+    toServerError = \case
+        ErrBalanceTxTxAlreadyBalanced ->
+            apiError err403 TransactionAlreadyBalanced $ mconcat
+                [ "The transaction is already balanced. "
+                , "Please send a transaction that requires more inputs/outputs to be picked to be balanced."
```
Anviking, line 3473: As said in
https://github.com/input-output-hk/cardano-wallet/pull/2782#discussion_r709126379, I was wondering if we really need `ErrBalanceTxTxAlreadyBalanced` at all, but I saw your response.

lib/core/src/Cardano/Wallet/Transaction.hs
```diff
@@ -152,6 +155,12 @@ data TransactionLayer k = TransactionLayer
         -> ByteString
         -> Either ErrDecodeSignedTx (Tx, SealedTx)
         -- ^ Decode an externally-signed transaction to the chain producer
+
+    , updateTx
+        :: SealedTx
+        -> ( [TxIn], [TxOut] )
+        -> Either ErrUpdateSealedTx (SealedTx, Word64)
+        -- ^ Update tx by adding additional inputs and outputs and give recalculated fee
```
Anviking, line 167: 👌

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -411,6 +411,8 @@ newTransactionLayer networkId = TransactionLayer
 
     , decodeSignedTx =
         _decodeSignedTx
+
+    , updateTx = \_sealedTx _inpsOuts -> undefined
```
Anviking:
```suggestion
    , updateTx = \_sealedTx _inpsOuts -> error "updateTx: not yet implemented"
```
good practice for readability, and for easier tracking if we were to hit it at runtime

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -2679,6 +2718,120 @@ instance DecodeAddress t => FromJSON (ApiConstructTransactionData t) where
...
+        , "txOut" .= toJSON outs ]
+
+instance DecodeAddress n => FromJSON (ApiBalanceTransactionPostData n) where
+    parseJSON = withObject "ApiBalanceTransactionPostData" $ \o -> do
+        cbor <- o .: "transaction" >>= (\trObj -> trObj .: "cborHex")
+        cosigners <- o .: "signatories"
+        bs <- fmap getApiBytesT $ parseJSON @(ApiBytesT 'Base16 ByteString) cbor
+        let sealedTxFromBytes _x = undefined
+        case sealedTxFromBytes bs of
+            Left err -> fail $ "cborHex seems to be not deserializing correctly due to " -- <> show err
```
Anviking: Commented code ☝️ (`- <> show err`)

Anviking (commented):
Thanks for splitting into a smaller PR!

Will have a final look soon

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: yes, I think when we have balanced tx we should react with error (if we want to use the same response as when constructTransaction endpoint). ApiCoinSelection cannot be empty to my best knowledge

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
paweljakubas: done

lib/core/src/Cardano/Wallet/Api/Types.hs
paweljakubas: I know, I commented/undefined several places to compile (it was before sign tx PR merged). Now, after sign tx PR it is not the case anymore

sevanspowell:
Thanks for splitting this up @paweljakubas, it really does help a lot.

sevanspowell (approved):
I've had a read through today and it seems sane. Like I said in my Slack message, if there is a specification for this work we can review against, I'll re-review in the morning when I'm fresh.

lib/shelley/test/unit/Cardano/Wallet/Shelley/CompatibilitySpec.hs
```diff
@@ -381,6 +393,74 @@ spec = do
...
+                , "uniswap-2.json"
+                , "uniswap-7.json"
+                --, "uniswap-9.json" -- Error in $[0]: there should be one 'lovelace' in 'value'
+                , "vesting-2.json"
+                ]
+        forM_ matrix $ \json -> do
+            let testFile = testPlutusDir </> json
+            it json $ property $ \(_thereWillBeWalletsHere :: Int) -> monadicIO $ do
+                bs <- run $ BL.readFile testFile
+                let decodeResult = eitherDecode @(ApiBalanceTransactionPostData 'Mainnet) bs
```
Anviking, line 461:
```suggestion
                let decodeResult = eitherDecode @(ApiBalanceTransactionPostData 'Mainnet) bs
                monitor $ counterexample ("decodeResult = " <> show decodeResult)

```

specifications/api/swagger.yaml
```diff
@@ -3137,7 +3250,46 @@ components:
...
+    ApiTransactionExtended: &ApiTransactionExtended
+      type: object
+      required:
+        - cborHex
+        - description
+        - type
+      properties:
+        cborHex: *serialisedTransactionHex
+        description: *txDescription
+        type: *txType
```
Anviking, line 3262: Is this precision needed rather than decoding as the latest possible tx?

When submitting txs through the node, we don't have to specify the era I believe? Why here?

lib/shelley/test/unit/Cardano/Wallet/Shelley/CompatibilitySpec.hs
```diff
@@ -381,6 +393,74 @@ spec = do
...
+                , "stablecoin_1-3.json"
+                , "stablecoin_1-4.json"
+                --, "stablecoin_2-2.json" --Error in $[0]: Value should not be empty
+                , "stablecoin_2-3.json"
+                , "token-account-2.json"
+                , "token-account-5.json"
+                , "uniswap-10.json"
+                , "uniswap-2.json"
+                , "uniswap-7.json"
+                --, "uniswap-9.json" -- Error in $[0]: there should be one 'lovelace' in 'value'
```
Anviking, line 454: Is the plan not to support these commented tests? Or to be addressed later?

Anviking (approved):

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: I would suspect/hope either 1) allowing `ApiCoinSelection` to be empty or 2) having both `ApiCoinSelection` and `ApiNonEmptyCoinSelection` would be saner than the workarounds both we and PAB would need for this error. But we'll see when implementing it.

lib/shelley/test/unit/Cardano/Wallet/Shelley/CompatibilitySpec.hs
paweljakubas: they are malformed IMHO. plutus team is expected to deliver corrected jsons - but wanted to document how it is

specifications/api/swagger.yaml
paweljakubas: In this place I only complied with what plutus scripts have inside. See examples in shelley/test/data/plutus directory of this PR

lib/shelley/test/unit/Cardano/Wallet/Shelley/CompatibilitySpec.hs
paweljakubas: will adopt this in coming PR or this one if is not merged in this attempt

PR #2901, opened 2021-09-16
Add Api.TypesSpec test cases for SealedTx request body

sevanspowell:
### Description

This PR contains some suggestions/extra work for #2896.

### Ticket

ADP-919

### DIFF
--- a/lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
@@ -1543,7 +1543,10 @@ instance Malformed (BodyParam (ApiBytesT 'Base64 SerialisedTx))
 -- in Server.hs. Tested by integration tests.
 
 instance Malformed (BodyParam (ApiT SealedTx)) where
-    malformed = [] -- fixme: add a couple things
+    malformed = first BodyParam <$>
+        [ ("", "DecoderErrorDeserialiseFailure 'Shelley Tx' (DeserialiseFailure 0 'end of input')")
+        , ("cafecafe", "DecoderErrorDeserialiseFailure 'Shelley Tx' (DeserialiseFailure 0 'expected list len or indef')")
+        ]
 
 instance Malformed (BodyParam ApiPostRandomAddressData) where
     malformed = first (BodyParam . Aeson.encode) <$>
--- a/lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
+++ b/lib/core/test/unit/Cardano/Wallet/ApiSpec.hs
@@ -123,7 +123,7 @@ import Network.Wai.Test
 import Servant
     ( Accept (..), Application, ReqBody, Server, StdMethod (..), Verb, serve )
 import Servant.API
-    ( (:<|>) (..), (:>), Capture )
+    ( (:<|>) (..), (:>), Capture, OctetStream )
 import Servant.API.Verbs
     ( NoContentVerb, ReflectMethod (..) )
 import Test.Hspec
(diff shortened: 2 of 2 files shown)


lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
```diff
@@ -1513,7 +1515,12 @@ instance Malformed (BodyParam (ApiBytesT 'Base64 SerialisedTx))
 -- in Server.hs. Tested by integration tests.
 
 instance Malformed (BodyParam (ApiT SealedTx)) where
-    malformed = [] -- fixme: add a couple things
+    malformed = first BodyParam <$>
+        [ ("", "DecoderErrorDeserialiseFailure 'Shelley Tx' (DeserialiseFailure 0 'end of input')")
+        , ("hello", "DecoderErrorDeserialiseFailure 'Shelley Tx' (DeserialiseFailure 0 'expected list len or indef')")
+        , ("cafecafe", "DecoderErrorDeserialiseFailure 'Shelley Tx' (DeserialiseFailure 0 'expected list len or indef')")
+        , (BSL.fromStrict . T.encodeUtf8 $ validSealedTxBase64, "")
```
sevanspowell: I would expect the case of using a "validSealedTxBase64" to succeed, but it does not. Fails with:

```
DecoderErrorDeserialiseFailure 'Shelley Tx' (DeserialiseFailure 0 'expected list len or indef')
```

lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
```diff
@@ -1513,7 +1515,12 @@ instance Malformed (BodyParam (ApiBytesT 'Base64 SerialisedTx))
 -- in Server.hs. Tested by integration tests.
 
 instance Malformed (BodyParam (ApiT SealedTx)) where
-    malformed = [] -- fixme: add a couple things
+    malformed = first BodyParam <$>
+        [ ("", "DecoderErrorDeserialiseFailure 'Shelley Tx' (DeserialiseFailure 0 'end of input')")
```
sevanspowell, line 1547: Are these the error messages you would expect @rvl?

lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
sevanspowell: Perhaps I'm converting it to ByteString incorrectly.

rvl: Oops, this endpoint accepts plain bytes, not hex-encoded bytes.

lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
```diff
@@ -113,7 +113,9 @@ import Servant
 
 import qualified Data.Aeson as Aeson
 import qualified Data.ByteString as BS
+import qualified Data.ByteString.Lazy as BSL
```
rvl:
```suggestion
import qualified Data.ByteString.Lazy as BL
```

rvl (approved):
Thanks.

lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
sevanspowell: Ah yes, you're right. That works!

sevanspowell:
bors try

lib/core/test/unit/Cardano/Wallet/Api/Malformed.hs
```diff
@@ -1513,7 +1513,11 @@ instance Malformed (BodyParam (ApiBytesT 'Base64 SerialisedTx))
 -- in Server.hs. Tested by integration tests.
 
 instance Malformed (BodyParam (ApiT SealedTx)) where
-    malformed = [] -- fixme: add a couple things
+    malformed = first BodyParam <$>
+        [ ("", "DecoderErrorDeserialiseFailure 'Shelley Tx' (DeserialiseFailure 0 'end of input')")
+        , ("hello", "DecoderErrorDeserialiseFailure 'Shelley Tx' (DeserialiseFailure 0 'expected list len or indef')")
+        , ("cafecafe", "DecoderErrorDeserialiseFailure 'Shelley Tx' (DeserialiseFailure 0 'expected list len or indef')")
```
rvl: We probably only need one of these examples. They fail for the same reason.

rvl (approved):
Thanks, please squash before merging.

sevanspowell:
bors retry

PR #2906, opened 2021-09-16
Balance endpoint integration

paweljakubas:
adp-656

We have preliminary impl of balanceTransaction that needs to work:
1. coin selection with external inputs (@jonathanknowles ) DONE
2. updateTx impl in shelley (@Anviking ) DONE

running tests:
```
stack test --ta '-m "TRANS_NEW_BALANCE_"' cardano-wallet:integration
```

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
@@ -853,7 +853,7 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
 
     it "TRANS_NEW_BALANCE_01e - plutus with missing covering inputs wallet enough funds" $ \ctx -> runResourceT $ do
 
-        liftIO $ pendingWith "Needs all prerequisites integrated with balanceTransaction - to be fixed in ADP-656"
+        --liftIO $ pendingWith "Needs all prerequisites integrated with balanceTransaction - to be fixed in ADP-656"
 
         -- constructing source wallet
         let initialAmt = 110_000_000_000
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
@@ -95,7 +95,6 @@ import Test.Integration.Framework.TestData
     )
 
 import qualified Cardano.Wallet.Api.Link as Link
-import qualified Data.List.NonEmpty as NE
 import qualified Data.Text as T
 import qualified Network.HTTP.Types.Status as HTTP
 
@@ -227,7 +226,7 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
 
         let filterInitialAmt =
                 filter (\(ApiCoinSelectionInput _ _ _ _ amt' _) -> amt' == Quantity initialAmt)
-        let coinSelInputs = filterInitialAmt $ NE.toList $
+        let coinSelInputs = filterInitialAmt $
                 getFromResponse (#coinSelection . #inputs) rTx
         length coinSelInputs `shouldBe` 1
 
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
@@ -818,7 +818,6 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
 
     it "TRANS_NEW_BALANCE_01d - single-output transaction with missing covering inputs" $ \ctx -> runResourceT $ do
 
-        liftIO $ pendingWith "Coin selection not integrated with balanceTransaction - to be fixed in ADP-656"
         -- constructing source wallet
         let initialAmt = 110_000_000_000
         let inpAmt = minUTxOValue (_mainEra ctx)
@@ -851,8 +850,6 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
             ]
 
     it "TRANS_NEW_BALANCE_01e - plutus with missing covering inputs wallet enough funds" $ \ctx -> runResourceT $ do
-
-        --liftIO $ pendingWith "Needs all prerequisites integrated with balanceTransaction - to be fixed in ADP-656"
 
         -- constructing source wallet
         let initialAmt = 110_000_000_000
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -1335,7 +1335,7 @@ selectionToUnsignedTx
 selectionToUnsignedTx wdrl sel s =
     UnsignedTx
         { unsignedInputs =
-            fullyQualifiedInputs $ view #inputs sel
+            fullyQualifiedInputs $ NE.toList $ view #inputs sel
         , unsignedOutputs =
             view #outputs sel
         , unsignedChange =
@@ -1346,28 +1346,24 @@ selectionToUnsignedTx wdrl sel s =
             fullyQualifiedWithdrawal wdrl
         }
   where
+    -- NOTE: External addresses, not known to the wallet, will be filtered out.
     qualifyAddresses
-        :: forall a t. (Traversable t)
-        => (a -> Address)
-        -> t a
-        -> t (a, NonEmpty DerivationIndex)
+        :: (a -> Address)
+        -> [a]
+        -> [(a, NonEmpty DerivationIndex)]
     qualifyAddresses getAddress hasAddresses =
-        fromMaybe
-        (error
-         "selectionToUnsignedTx: unable to find derivation path of a known \
-         \input or change address. This is impossible.")
-        (traverse withDerivationPath hasAddresses)
+        mapMaybe withDerivationPath hasAddresses
       where
         withDerivationPath hasAddress =
             (hasAddress,) <$> fst (isOurs (getAddress hasAddress) s)
 
-    fullyQualifiedInputs :: Traversable t => t (TxIn, TxOut) -> t input
+    fullyQualifiedInputs :: [(TxIn, TxOut)] -> [input]
     fullyQualifiedInputs =
         fmap mkInput . qualifyAddresses (view #address . snd)
       where
         mkInput ((txin, txout), path) = (txin, txout, path)
 
-    fullyQualifiedChange :: Traversable t => t TxOut -> t change
+    fullyQualifiedChange :: [TxOut] -> [change]
     fullyQualifiedChange =
         fmap mkChange . qualifyAddresses (view #address)
       where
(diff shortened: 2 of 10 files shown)


lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2193,12 +2197,116 @@ constructTransaction ctx genChange (ApiT wid) body = do
...
+                }
+
+        (internalUtxoAvailable, wallet, pendingTxs) <-
+            liftHandler $ W.readWalletUTxOIndex @_ @s @k wrk wid
+
+        let externalSelectedUtxo =
+                UTxOIndex.fromSequence externalInputs
+
+        let utxoAvailable = UTxOSelection.fromIndexPair
+                (internalUtxoAvailable, externalSelectedUtxo)
```
jonathanknowles, line 2245: This looks good. 👍🏻

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2193,12 +2197,116 @@ constructTransaction ctx genChange (ApiT wid) body = do
...
+            , wallet
+            }
+            transform
+
+        fee <- liftHandler $ W.estimateFee runSelection
+        return (fee, unsignedtx)
+
+    let UnsignedTx colls inps _outs change _wdrl = unsignedtx
+    let extraBody = ExtraTxBodyContent
+            { extraInputs = (\(txin, _, _) -> txin) <$> F.toList inps
```
Anviking, line 2277: Won't `inps` already include the pre-selected inputs from `sealedTxIncoming`?

`updateTx` would then try to append the whole `inps` to the pre-selected inputs.

Although as the ledger represents the inputs as a `Set`, it might not actually matter in the end.

paweljakubas: yes, we may filter out preselected inputs. although ledger stores inputs as set so it should not affect txbody or fee.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2193,12 +2197,116 @@ constructTransaction ctx genChange (ApiT wid) body = do
...
+    let (Tx _id _fee _coll _inps outs wdrlMap mdM _validity) = txIncoming
+
+    ((FeeEstimation feeMin _), unsignedtx) <- withWorkerCtx ctx wid liftE liftE $ \wrk -> do
+
+        (acct, _, _) <- liftHandler $ W.readRewardAccount @_ @s @k @n wrk wid
+        (wdrl, _) <-
+            if Map.member acct wdrlMap then
+                mkRewardAccountBuilder @_ @s @_ @n ctx wid (Just SelfWithdrawal)
+            else
+                mkRewardAccountBuilder @_ @s @_ @n ctx wid Nothing
```
```suggestion
        (wdrl, _) <- mkRewardAccountBuilder @_ @s @_ @n ctx wid $
            if Map.member acct wdrlMap
            then Just SelfWithdrawal
            else Nothing
```

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
```diff
@@ -426,11 +426,14 @@ data UnsignedTx input output change withdrawal = UnsignedTx
...
-        :: NonEmpty input
-        -- Inputs are *necessarily* non-empty because Cardano requires at least
+        :: [input]
+        -- ^ Inputs are *necessarily* non-empty because Cardano requires at least
         -- one UTxO input per transaction to prevent replayable transactions.
         -- (each UTxO being unique, including at least one UTxO in the
         -- transaction body makes it seemingly unique).
+        --
+        -- *However* when used to respresent the inputs known by the wallet, in
+        -- contrast to all inputs, it can be empty.
```
jonathanknowles:
```suggestion
        -- *However* when used to represent the inputs known by the wallet, in
        -- contrast to all inputs, it can be empty.
```

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2193,12 +2197,116 @@ constructTransaction ctx genChange (ApiT wid) body = do
...
+                , utxoAvailableForCollateral =
+                    UTxOIndex.toUTxO internalUtxoAvailable
+                , wallet
+                } getFee
+              where getFee = const (selectionDelta TokenBundle.getCoin)
+
+        let transform s sel' =
+                W.assignChangeAddresses genChange sel' s
+                & uncurry (W.selectionToUnsignedTx (txWithdrawal txCtx))
+        unsignedtx <- liftHandler $ W.selectAssets @_ @s @k wrk W.SelectAssetsParams
```
Anviking, line 2261: Because `selectAssets` only knows about the things we tell it, and not the entire `sealedTxIncoming`, I think exotic things like submitting update proposals will result in imbalanced txs.

In a re-thought approach, related to ADP-891, and I think which you've mentioned, we could calculate `minfee sealedTxIncoming` instead and go from there (either by overestimating for the TxExtraBodyContent, or trying to find some fix point)

But I guess the similar and more pressing issue is datums, which everyone was aware of on today's call.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1346,28 +1346,24 @@ selectionToUnsignedTx wdrl sel s =
             fullyQualifiedWithdrawal wdrl
         }
   where
+    -- NOTE: External addresses, not known to the wallet, will be filtered out.
```
Anviking, line 1349: When talking briefly with Nikola and hearing / being reminded that they will present the result of balanceTransaction in full detail to the user, this seems wrong.

We probably _should_:
1. Make the `DerivationPath` optional on `ApiCoinSelection`
2. Don't filter out pre-selected inputs/outputs
3. Ensure all other possible tx data is exposed. (We're probably missing glue code for showing datums in api here)

lib/core/src/Cardano/Wallet/Api/Server.hs

lib/core/src/Cardano/Wallet.hs
Anviking: (Not critical for merging this PR and I could have a look)

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: agree here, it is reasonable to start with minFee, and then increment knowing the cost per inp/out/coll. But we need to remember that evaluation of min fee requires number of witnesses, so it would influence selecting inputs/collaterals in feedback loop. It is not staightforward for me how, in details, it should work

lib/core/src/Cardano/Wallet.hs
paweljakubas: Daedalus will invoke balance tx, so it will know details of external inputs already - see "inputs" in https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/balanceTransaction And in `ApiCoinSelection` we will present ADDED inps/outs/colls (always being the resource of the wallet). So the client will know everything having "inputs" and response. 
Some argument in favor of what is now is that it is in line what constructTx responds, and as both construct and balance are first step before sign+submit. It makes sense to be compatible....

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2193,12 +2197,115 @@ constructTransaction ctx genChange (ApiT wid) body = do
...
+    ti :: TimeInterpreter (ExceptT PastHorizonException IO)
+    ti = timeInterpreter nl
+
+    apiExternalInps = body ^. #inputs
+    toTxInTxOut
+        (ApiExternalInput (ApiTxIn (ApiT hashTx) ix)
+        (ApiTxOut (ApiT addr, _) _ (Quantity amt) (ApiT assets))) =
+        (TxIn hashTx ix, TxOut addr (TokenBundle (Coin $ fromIntegral amt) assets))
+    externalInputs = toTxInTxOut <$> apiExternalInps
+    sealedTxIncoming = body ^. #transaction . #getApiT
```
Anviking, line 2305: Just some stylistic thoughts:
I think I have more of a taste for either:
1. Defining important values as `let` bindings at the top
2. Making the `where` definitions all functions rather than values

such that you can read it from top to bottom and easily see where `txIncoming` is coming from.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2193,12 +2197,115 @@ constructTransaction ctx genChange (ApiT wid) body = do
...
+    -> ArgGenChange s
     -> ApiT WalletId
     -> ApiBalanceTransactionPostData n
     -> Handler (ApiConstructTransaction n)
-balanceTransaction _ctx (ApiT _wid) _body =
-    liftHandler $ throwE ErrBalanceTxNotImplemented
+balanceTransaction ctx genChange (ApiT wid) body = do
+    pp <- liftIO $ NW.currentProtocolParameters nl
+
+    --TODO  deal with coll and validity and delegations
```
Anviking, line 2219: 👍 Yes.

Collateral we could simply forbid [as discussed](https://input-output-rnd.slack.com/archives/GBT05825V/p1633089400064100).

Delegations and other stuff could be present in the sealedTxIncoming, but are probably not _crucial_. (unused by PAB)

Validity, however, does sound crucial to support. (used by PAB)

Related to https://github.com/input-output-hk/cardano-wallet/pull/2906#discussion_r721160180

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -857,8 +855,6 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
 
     it "TRANS_NEW_BALANCE_01e - plutus with missing covering inputs wallet enough funds" $ \ctx -> runResourceT $ do
 
-        liftIO $ pendingWith "Needs all prerequisites integrated with balanceTransaction - to be fixed in ADP-656"
```
Anviking, line 856: _Could_ probably have more assertions in the `verify rTx` below, but as you say, the most crucial thing is the full e2e scenarios.

Anviking (commented):
It is great to see this start coming together concretely!

My take is that remaining problems are:
1. Critical: Ensure fees are not underestimated because of `datums`
2. Critical: Ensure e.g. ProtocolParameter Hash, or other stuff won't lead to underestimated fees
3. Moderate: Is the API response for CoinSelection appropriate for Daedalus?
4. Low/Moderate: Certificates, update proposals leading to underestimated fees

and that we shouldn't _rush_ to merge this before seeing how the critical things will pan out.

I suspect there are 3 potential solutions for (1):
a. Tweak the coin selection in a minimal / hacky way to account for datums
b. Call ledger/node's `minfee` in a loop - first on the partial input tx, and then after adding inputs and outputs until you converge.
c. Call ledger/node's `minfee` once, and then use the existing `TxConstraints` overestimations to add additional inputs and outputs.

## Edit 

I must have gotten confused in the original message. Got things a bit wrong. Now edited. We _are_ adjusting the original binary. Ensuring tx body content is preserved is not the problem. The problem is ensuring the fee estimation also is aware of all that content.

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: 🤔 might have gotten this slightly wrong. Will revisit tomorrow.

Anviking (approved):

PR #2908, opened 2021-09-17
Provide `MockSelectionConstraints` type and generator.

jonathanknowles:
## Issue Number

ADP-1118
ADP-1136

## Summary

This PR:
- [x] Introduces the `MockSelectionConstraints` type with an associated generator and shrinker.
    The goal of this type is to:
    - generate arbitrary selection constraints to test `performSelection`;
    - provide good shrinking of these constraints.
- [x] Uses the `MockSelectionConstraints` type to simplify the `prop_performSelection` family of properties.
- [x] Moves `computeSelectionLimit` out of the `SelectionParams` record and into the `SelectionConstraints` record.
    This location is more appropriate, since:
    - `SelectionConstraints` is designed to collect constraints that are _common to all selections in the current era_, whereas `SelectionParams` is designed to collect parameters _specific to a given selection_.
    - `SelectionConstraints` is now a subset of `SelectionConstraints` in the parent module.

Additional work (to be used by a future PR):
- [x] Adds the `Coin.difference` function with property tests.
- [x] Adds the `TokenMap.size` function with property tests.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Balance.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Balance.hs
@@ -411,10 +411,6 @@ data SelectionResult change = SelectionResult
     , changeGenerated
         :: ![change]
         -- ^ A list of generated change outputs.
-    , utxoRemaining
-        :: !UTxOIndex
-        -- ^ The subset of 'utxoAvailable' that remains after performing
-        -- the selection.
     , assetsToMint
         :: !TokenMap
         -- ^ The assets to mint.
@@ -874,14 +870,13 @@ performSelection constraints params
             , extraCoinSink
             , changeGenerated = changeGenerated
             , outputsCovered = NE.toList outputsToCover
-            , utxoRemaining = leftover
             , assetsToMint
             , assetsToBurn
             }
 
         selectOneEntry = selectCoinQuantity selectionLimit
 
-        SelectionState {selected, leftover} = s
+        SelectionState {selected} = s
 
         requiredCost = computeMinimumCost SelectionSkeleton
             { skeletonInputCount = UTxOIndex.size selected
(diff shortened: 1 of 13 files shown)


lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/BalanceSpec.hs
```diff
@@ -1939,60 +1999,136 @@ mkMinCoinValueFor = \case
...
       -- token bundle.
     deriving (Eq, Show)
 
-mkBundleSizeAssessor
-    :: MockTokenBundleSizeAssessor -> TokenBundleSizeAssessor
-mkBundleSizeAssessor m = TokenBundleSizeAssessor $ case m of
-    NoBundleSizeLimit ->
+genMockAssessTokenBundleSize :: Gen MockAssessTokenBundleSize
+genMockAssessTokenBundleSize =
+    -- TODO:
```
sevanspowell, line 2102: Is there a reason why we don't?

jonathanknowles: > Is there a reason why we don't?

If I recall correctly, at the time we wrote `prop_performSelection` we believed it would complicate the property test too much to impose a token bundle size limit, so we always generated criteria without them.

I'm actually planning to improve this generator in the next PR: https://github.com/input-output-hk/cardano-wallet/pull/2909

lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Balance.hs
```diff
@@ -411,10 +411,6 @@ data SelectionResult change = SelectionResult
     , changeGenerated
         :: ![change]
         -- ^ A list of generated change outputs.
-    , utxoRemaining
```
sevanspowell, line 413: :clap:

sevanspowell (approved):

PR #2909, opened 2021-09-18
Support coin selection with no outputs.

jonathanknowles:
### Issue Number

ADP-1136

### Summary

This PR makes it possible to perform coin selection when there are no user-specified outputs.

### DIFF
--- a/lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/BalanceSpec.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/BalanceSpec.hs
@@ -29,7 +29,6 @@ import Cardano.Wallet.Primitive.CoinSelection.Balance
     , MakeChangeCriteria (..)
     , RunSelectionParams (..)
     , SelectionConstraints (..)
-    , SelectionDelta (..)
     , SelectionError (..)
     , SelectionInsufficientError (..)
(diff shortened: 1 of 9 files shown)


lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/BalanceSpec.hs
```diff
@@ -1095,6 +1062,163 @@ prop_performSelection mockConstraints (Blind params) coverage =
...
+            -- ^ The transformed result.
+        }
+    deriving Generic
+
+-- | A functor instance on 'TransformationReport' that allows the final result
+--   value to be transformed by a function transformer.
+--
+instance Functor (TransformationReport paramsTransformed result) where
+    fmap f (TransformationReport paramsTransformed result resultUntransformed)
+        = TransformationReport paramsTransformed result (f resultUntransformed)
```
jonathanknowles: **Suggestion: derive this instance automatically.**

This instance can actually be derived automatically, since the result to be transformed is the final type parameter.

jonathanknowles: Fixed in faf943eda6b181917f0cce8ee3dcad440a290cb2.

sevanspowell (approved):

PR #2912, opened 2021-09-20
Balance tx integration tests

paweljakubas:
adp-1142

Three tests are pending:
1. On will be unpended when adp-1143 lands
2. The second will be unpended when coin selection with external inputs is integrated
3. The last one is proof the whole integration works


lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -740,6 +740,148 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
   -- minting
   -- update with sign / submit tx where applicable
   -- end to end join pool and get rewards
+
+    it "TRANS_NEW_BALANCE_01a - multiple-output transaction with all covering inputs present" $ \ctx -> runResourceT $ do
+
+        liftIO $ pendingWith "No errMsg403transactionAlreadyBalanced is returned - to be fixed in ADP-656"
```
piotr-iohk, line 746: If it is already balanced can we just sign and submit?
If so then good to add sign and submit steps and perhaps reflect this information in error message.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -740,6 +740,148 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
...
+                \280ec283eb7e12146726bfe68dcd69c7a85123ce2f7a10e7afa062de6e\
+                \c013404d4f9909877d452fc57dfe4f8b67f94e0ea1e8a0ba1a000f422a\
+                \825839011a2f2f103b895dbe7388acc9cc10f90dc4ada53f46c841d2ac\
+                \44630789fc61d21ddfcbd4d43652bf05c40c346fa794871423b65052d7\
+                \614c1b0000000ba42b176a82583901c59701fee28ad31559870ecd6ea9\
+                \2b143b1ce1b68ccb62f8e8437b3089fc61d21ddfcbd4d43652bf05c40c\
+                \346fa794871423b65052d7614c1b0000000ba42b176a021a000234d803\
+                \198ceb0e80a0f5f6" :: Text
+        let balancePayload = Json [json|{
+              "transaction": { "cborHex" : #{serializedTxHex}, "description": "", "type": "Tx AlonzoEra" },
```
piotr-iohk, line 776: The type is always `"type": "Tx AlonzoEra"` it seems which also is also the only allowed according to API doc (https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/balanceTransaction). As far as I see (testing manually) there does not seem to be any checking of that value and you can basically put anything there...

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -740,6 +740,148 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
...
+              "signatories": [],
```
piotr-iohk, line 777: signatories are always empty for all scenarios. (I believe this is the case also for provided examples from https://github.com/input-output-hk/cardano-wallet/tree/master/lib/shelley/test/data/plutus), I think we should have some test with signatories too.

piotr-iohk (commented):
That's a good start.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
paweljakubas: let's do it in separate test suite and PR. Here I would like to focus entirely on balancing

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
paweljakubas: we don't use this information in api machinery as everything (information about compatibility with Alonzo Era or other era) is inside cborHex and we get this from deserialisation. The schema was imposed on us - please remember. Maybe we will validate the description, but now now.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
paweljakubas: signatories is not used now, it is some next stage and at this stage only informatory (although it needs to be extended account public key). What is important now is `cborHex` and `inputs`

piotr-iohk (approved):

PR #2874, opened 2021-09-07
Fix `decodeUnsignedTx` to correctly try previous eras

paweljakubas:
### Issue Number

ADP-1110

### Comments

This PR adjusts `decodeSignedTx` so that when decoding in a given era, it also correctly tries all previous eras.

PR #2914, opened 2021-09-20
Add updateSealedTx and ExtraTxBodyContent

Anviking:
- [x] Refactoring: Replace `to{Shelley, Allegra, Mary, Alonzo}TxOut` with `toCardanoTxOut` for easier re-use.
- [x] Add `updateSealedTx :: SealedTx -> ExtraTxBodyContent -> Either ErrUpdateSealedTx SealedTx`
- [x] Test that `updateSealedTx noExtraBodyContent` works with all PAB goldens
- [x] Test that `updateSealedTx noExtraBodyContent` fails when there are existing key witnesses
- [x] Test that inputs and outputs are combined correctly. 

### Comments

Not tested:
- That it works for all previous eras.

### Issue Number

ADP-1140

Relates to / should be needed by #2906


lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -3472,8 +3472,14 @@ instance IsServerError ErrBalanceTx where
                 [ "The transaction is already balanced. "
                 , "Please send a transaction that requires more inputs/outputs to be picked to be balanced."
                 ]
-        ErrBalanceTxUpdateError (ErrUpdateSealedTxBodyError hint) ->
-            apiError err500 CreatedInvalidTransaction hint
+        ErrBalanceTxUpdateError (ErrExistingKeyWitnesses n) ->
+            apiError err403 CreatedInvalidTransaction $ mconcat
+                [ "The transaction could not be balanced, because it contains "
+                , T.pack (show n), " "
+                , "existing key-witnesses which would be invalid after " -- TODO: I /think/ key-witnesses is an accurate term here, but haven't checked.
```
Anviking: In the ledger it seems the witnesses are all equal:
https://github.com/input-output-hk/cardano-ledger-specs/blob/0eb9b38bb125ec8eedf0f7aabf9e09f9b7bf79f7/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWitness.hs#L181-L188

But it seems we only get key witnesses from cardano-api:
https://github.com/input-output-hk/cardano-node/blob/0ed04ab8d0a205691219f04623a3d880545a3f34/cardano-api/src/Cardano/Api/Tx.hs#L493-L507

This might actually be a separate problem: any existing non-key witnesses will be lost by the balancing…

Anviking: It turned out that script witnesses, datums, etc were stored in the cardano-api version of the `TxBody`. So all is good.

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -1600,3 +1617,111 @@ matrixNormalTxExamples =
...
+                        --, "stablecoin_2-2.json" --Error in $[0]: Value should not be empty
+                        , "stablecoin_2-3.json"
+                        , "token-account-2.json"
+                        , "token-account-5.json"
+                        , "uniswap-10.json"
+                        , "uniswap-2.json"
+                        , "uniswap-7.json"
+                        --, "uniswap-9.json" -- Error in $[0]: there should be one 'lovelace' in 'value'
+                        , "vesting-2.json"
+                        ]
```
Anviking, line 1694: Duplicated with CompatibilitySpec… Could de-duplicate via some shared `TestData` or `Util` module if we want to

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -3478,8 +3478,16 @@ instance IsServerError ErrBalanceTx where
...
-        ErrBalanceTxUpdateError (ErrUpdateSealedTxBodyError hint) ->
-            apiError err500 CreatedInvalidTransaction hint
+        ErrBalanceTxUpdateError ErrByronTxNotSupported ->
+            apiError err403 CreatedInvalidTransaction
+                "Balancing Byron transactions is not supported."
+        ErrBalanceTxUpdateError (ErrExistingKeyWitnesses n) ->
+            apiError err403 CreatedInvalidTransaction $ mconcat
+                [ "The transaction could not be balanced, because it contains "
+                , T.pack (show n), " "
+                , "existing key-witnesses which would be invalid after "
```
paweljakubas, line 3486: I would even say that at this stage we should not have any witnesses collected. signing should be done only after tx is balanced (hence I slightly incline to introduce AlreadyBalancedTx error in balance tx PR)

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -1251,70 +1248,77 @@ toCardanoLovelace (W.Coin c) = Cardano.Lovelace $ safeCast c
...
-  where
-    datumHash = Cardano.TxOutDatumHashNone
-    addrInEra = tina "toCardanoTxOut: malformed address"
-        [ Cardano.AddressInEra (Cardano.ShelleyAddressInEra Cardano.ShelleyBasedEraAlonzo)
-            <$> deserialiseFromRawBytes AsShelleyAddress addr
-
-        , Cardano.AddressInEra Cardano.ByronAddressInAnyEra
-            <$> deserialiseFromRawBytes AsByronAddress addr
-        ]
+    toShelleyTxOut :: HasCallStack => W.TxOut -> Cardano.TxOut ShelleyEra
```
paweljakubas, line 1258: I wonder if we could not reuse here better the functions `toEraTxOut`. And apply the trick was applied here in `toGenTx` https://github.com/input-output-hk/cardano-wallet/pull/2874/files

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -441,6 +453,142 @@ mkDelegationCertificates da accXPub =
...
+        -- cases could perhaps be joined. It is not however. And we still need
+        -- to treat Alonzo and Shelley differently.
+        adjust
+            :: ExtraTxBodyContent
+            -> ShelleyBasedEra era
+            -> Ledger.TxBody (Cardano.ShelleyLedgerEra era)
+            -> Ledger.TxBody (Cardano.ShelleyLedgerEra era)
+        adjust (ExtraTxBodyContent extraInputs extraOutputs modifyFee) era body = case era of
+            ShelleyBasedEraAlonzo -> body
+                    { Alonzo.outputs = Alonzo.outputs body
```
paweljakubas, line 498: no adding extra outputs here? it seems wrong at first glance

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
paweljakubas: I know. In the first version I wanted to dump the plutus script. Definitely something could be done to reuse those jsons name as they repeat in 3 places already. Feel free to de-duplicate if you have any idea and is eager to to do it withing this PR

paweljakubas (approved):
good work! We can probably simplify error and say there are witnesses without specifying the number. Also please double check the possible missing output update for alonzo - I think suggested below property test in point 2 would catch that!

Also suggestion for another property tests:
1. we have extra input(s) and update -> after updating we always find after decostruction those inputs
2. the same for extra outputs
3. the same when we update fee - the new value will be there
4. also combinations of 1-3
5. maybe check additionally after point 4 that binary payload (ie. BS.length) is always bigger after updating  
And do those for different eras. Maybe we could reuse generators from cardano-api - there are some exposed for txOut and txIn I believe (if possible)

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
Anviking: Not sure what you're suggesting here. `toCardanoTxOut` is basically a `toEraTxOut`, no?

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
Anviking: Wasn't super-eager, but for the sake of making the property tests neater, I might give it a shot in a follow-up PR.

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: I added a `Please sign the transaction after it is balanced instead.` line afterwards.

However, if for some reason a user want to add key-witness, I'm not sure if the balance-endpoint can take the size of the witness into account for the coin selection (if it is not known to the wallet). Maybe it does…

Anviking:
Thanks @paweljakubas, good catch!

I added this property test:
```haskell
prop_updateSealedTx :: PartialTx -> [TxIn] -> [TxOut] -> Coin -> Property
prop_updateSealedTx (PartialTx tx) extraIns extraOuts newFee = do
    let extra = ExtraTxBodyContent extraIns extraOuts (const newFee)
    let tx' = either (error . show) id
            $ updateSealedTx extra tx

    conjoin
        [ ins tx' === ins tx <> Set.fromList extraIns
        , outs tx' === outs tx <> Set.fromList extraOuts
        , fee tx' === Just newFee
        ]
  where
    ins = Set.fromList . map fst . view #resolvedInputs . _decodeSealedTx
    outs = Set.fromList . view #outputs . _decodeSealedTx
    fee = view #fee . _decodeSealedTx
```
which caught the problem, and satisfies your point 1-4.

I'll merge this now, but will try to simplify the other updateSealedTx tests a bit and e.g. address your point (5) in a follow-up PR. It should be somewhat covered already by the fact that `updateSealedTx noExtraTxContent` produces the exact same binary.

Anviking:
> there are witnesses without specifying the number

I thought it provided some transparency without showing the full witnesses themselves. 

If someone is unsure what counts as a "key-witness", or is doubting the wallet implementation, they can cross-check the number with the data they provide.

Anviking:
I'm marking the property test pending. It passed when changing the dummyAddress size, but doing so breaks other tests. I think it just needs to be done, and should be harmless, but it's a big enough change to warrant a new PR.

PR #2917, opened 2021-09-22
Remove `selectAssetsNoOutputs` workaround

jonathanknowles:
## Issue Number

ADP-1136 (and others)

## Summary

This PR removes the `selectAssetsNoOutputs` workaround.


jonathanknowles:
bors try

jonathanknowles:
bors try

sevanspowell (approved):

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -186,7 +186,7 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
         (_, mw) <- rewardWallet ctx
         wa <- fixtureWallet ctx
         let withdrawal = Json [json|{ "withdrawal": #{mnemonicToText mw} }|]
-        let expectedFee = 1139500
+        let expectedFee = 139500
```
piotr-iohk, line 189: I don't know why those fees dropped at exactly 1 ADA. This is not something I can observe on testnet. E.g. constructing tx with only metadata or only withdrawals produces the same fee for `master` and `v2021-09-09` (which doesn't include this change). The fee is reasonable in both cases, ~0,16 ADA. cc @jonathanknowles

PR #2920, opened 2021-09-23
Expose min fee from node

paweljakubas:
adp-1143

`evaluateTransactionFee` from cardano-api uses cardano-api's ProtocolParamters. Hence,
1. cardano-api's ProtocolParameters were exposed via new network method DONE
2. evaluateMinimumFee was added to transaction interface DONE
3. impl in shelley using cardano-api function was used using cardano-api's ProtocolParameters and SealedTx DONE
4. function determining the number of witnesses was implemented  DONE 
5. try to address calculation of number of certificate wits (using ledger things) DONE

Limitations:
- not for ByronEra and not for Byron witnesses

### DIFF
--- a/lib/core/src/Cardano/Wallet/Transaction.hs
+++ b/lib/core/src/Cardano/Wallet/Transaction.hs
@@ -151,9 +151,14 @@ data TransactionLayer k tx = TransactionLayer
             -- Current protocol parameters
         -> tx
             -- The sealed transaction
-        -> Coin
+        -> Maybe Coin
         -- ^ Evaluate a minimal fee amount necessary to pay for a given tx
         -- using ledger's functionality
+        --
+        -- Will estimate how many witnesses there /should be/, so it works even
+        -- for unsigned transactions.
+        --
+        -- Returns `Nothing` for ByronEra transactions.
 
     , computeSelectionLimit
         :: ProtocolParameters
--- a/lib/core/src/Cardano/Wallet/Transaction.hs
+++ b/lib/core/src/Cardano/Wallet/Transaction.hs
@@ -80,6 +80,9 @@ import Fmt
 import GHC.Generics
     ( Generic )
 
+import qualified Cardano.Api.Shelley as Node
+
+
 data TransactionLayer k tx = TransactionLayer
     { mkTransaction
         :: AnyCardanoEra
@@ -142,6 +145,15 @@ data TransactionLayer k tx = TransactionLayer
             -- The constructed transaction that could contain plutus scripts
         -> Coin
         -- ^ Compute an execution costs of scripts in a given transaction.
+
+    , evaluateMinimumFee
+        :: Node.ProtocolParameters
+            -- Current protocol parameters
+        -> tx
+            -- The sealed transaction
+        -> Coin
+        -- ^ Evaluate a minimal fee amount necessary to pay for a given tx
+        -- using ledger's functionality
 
     , computeSelectionLimit
         :: ProtocolParameters
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
@@ -74,6 +74,7 @@ module Cardano.Wallet.Shelley.Compatibility
     , toCardanoStakeCredential
     , toCardanoValue
     , fromCardanoValue
+    , fromCardanoLovelace
     , rewardAccountFromAddress
     , fromShelleyPParams
     , fromAlonzoPParams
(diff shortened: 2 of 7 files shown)


lib/shelley/src/Cardano/Wallet/Shelley/Network.hs
```diff
@@ -766,7 +769,19 @@ mkTipSyncClient tr np onPParamsUpdate onInterpreterUpdate onEraUpdate = do
                     <$> LSQry Shelley.GetCurrentPParams)
                 (fromAlonzoPParams eraBounds
                     <$> LSQry Shelley.GetCurrentPParams)
-            return (pp, sp)
+
+            ppNode <- onAnyEra
+                (error "not sure at this moment how to handle that")
```
Anviking: Can we not make this `Nothing`?

paweljakubas: we can. good idea

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -622,6 +626,62 @@ dummySkeleton inputCount outputs = SelectionSkeleton
...
 
+_evaluateMinimumFee
+    :: Cardano.ProtocolParameters
+    -> SealedTx
+    -> Coin
+_evaluateMinimumFee pp tx =
+    fromCardanoLovelace minFee
+  where
+    minFee = case getSealedTxBody tx of
+        InAnyCardanoEra ShelleyEra txbody ->
```
Anviking, line 646: I'm looking into this, but I don't think we need to expose the ledger txbodies here... The `Cardano.Tx body wits` pattern should be enough...

Anviking: Or is it intentional for some reason?

Anviking: The required `IsShelleyBasedEra era`, I think we should be able to conjure up somehow...

paweljakubas: unfortunately,in the case of tx certs I cannot extract from cardano-node due to how it sneaks BuildTx there. And it turn out that doing this through ledger tx is a way...

paweljakubas: feel free to try without ledger bypass if you have a while

lib/core/src/Cardano/Wallet/Transaction.hs
```diff
@@ -143,6 +146,15 @@ data TransactionLayer k tx = TransactionLayer
         -> Coin
         -- ^ Compute an execution costs of scripts in a given transaction.
 
+    , evaluateMinimumFee
+        :: Node.ProtocolParameters
+            -- Current protocol parameters
+        -> tx
+            -- The sealed transaction
+        -> Coin
```
Anviking: Perhaps we should return `Maybe` instead of erroring in case of byron, but I think leaving it as-is is fine while still trying to fit the pieces together.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -716,6 +722,166 @@ txParametersFromPParams maxBundleSize pp = W.TxParameters
     naturalToDouble :: Natural -> Double
     naturalToDouble = fromIntegral
 
+--------------------------------------------------------------------------------
+-- Copied from cardano-api
+-- To be removed when once again exposed.
```
Anviking, line 727: I added this note for future readers

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
Anviking: For the record, what I had missed until our quick call was:

- We need to predict how many witnesses there will be, as the transaction is not signed.
- We cannot use [estimateTransactionKeyWitnessCount](https://github.com/input-output-hk/cardano-node/blob/0ed04ab8d0a205691219f04623a3d880545a3f34/cardano-api/src/Cardano/Api/Fees.hs#L284-L285)

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -622,6 +626,62 @@ dummySkeleton inputCount outputs = SelectionSkeleton
...
+            let (Cardano.ShelleyTxBody _ ledgertxbody _ _ _ _) = txbody
+                certNum = length $ ShelleyMA.certs' ledgertxbody
+            in Cardano.evaluateTransactionFee @Cardano.AllegraEra pp txbody (witsNum txbody certNum) 0
+        InAnyCardanoEra MaryEra txbody ->
+            let (Cardano.ShelleyTxBody _ ledgertxbody _ _ _ _) = txbody
+                certNum = length $ ShelleyMA.certs' ledgertxbody
+            in Cardano.evaluateTransactionFee @Cardano.MaryEra pp txbody (witsNum txbody certNum) 0
+        InAnyCardanoEra AlonzoEra txbody ->
+            let (Cardano.ShelleyTxBody _ ledgertxbody _ _ _ _) = txbody
+                certNum = length $ Alonzo.txcerts ledgertxbody
```
Anviking, line 660: I think these `certNum` could be a slight overestimate, as not all certs require witnesses (KeyReg being the exception). This should affect the result of `_evaluateMinimumFee` when Daedalus users delegate for the first time.

Another, more theoretical case would also be where all the certs require just a single key witness.

Not a big deal, but could add a comment about it.

lib/core/src/Cardano/Wallet/Transaction.hs
Anviking: I ended up adding a `Maybe` here.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
Anviking: 6faea90

Anviking (approved):
LGTM, will be useful to have `currentNodeProtocolParameters`

PR #2922, opened 2021-09-27
Extract out type `UTxOSelection`

jonathanknowles:
## Issue Number

ADP-1118

## Summary

This PR introduces the `UTxOSelection` and `UTxOSelectionNonEmpty` types, which represent in-progress selections of entries from a UTxO set.

These types have the following characteristics:
- `UTxOSelection`
  a selection where the selected set may be empty
- `UTxOSelectionNonEmpty`
  a selection where the selected set cannot be empty

Both types share the same internal state:

```hs
data State = State
    { leftover :: !UTxOIndex
      -- ^ UTxOs that have not yet been selected.
    , selected :: !UTxOIndex
      -- ^ UTxOs that have already been selected.
    }
```

The non-emptiness of a UTxO selection is an important post-condition of `Balance.runSelection`, so this PR uses this type to remove some run-time assertions from the `Balance` module. (These assertions are moved to the `UTxOSelection` module, where they are tested with property tests.)

Additionally, this PR:
- uses the `UTxOSelection` and `UTxOSelectionNonEmpty` types to replace the use of `SelectionState` within `CoinSelection.Balance`.
- removes the now-redundant run-time assertions for a non-empty selection from `Balance`, as the `UTxOSelectionNonEmpty` type handles this internally.

## Follow-on work

Adjust `Balance.performSelection` to accept a `UTxOSelection` as part of the `SelectionParams` instead of the existing `utxoAvailable :: UTxOIndex` field. With a few adjustments, this will allow `performSelection` to be called with existing inputs.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/Types/UTxOSelection.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types/UTxOSelection.hs
@@ -1,0 +1,378 @@
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedLabels #-}
+{-# LANGUAGE TypeFamilies #-}
+
+-- |
+-- Copyright: © 2018-2021 IOHK
+-- License: Apache-2.0
+--
+-- Provides the 'UTxOSelection' type, which represents a selection of UTxO
+-- entries from a UTxO set.
+--
+-- It consists of a pair of UTxO sets:
+--
+--    - the selected set: UTxOs that have already been selected;
+--    - the leftover set: UTxOs that have not yet been selected.
+--
+-- To construct a 'UTxOSelection' where none of the UTxOs are selected, use
+-- the 'fromIndex' function.
+--
+-- To construct a 'UTxOSelection' where some of the UTxOs are selected, use
+-- either the 'fromIndexFiltered' or the 'fromIndexPair' functions.
+--
+-- To select an element (and move it from the leftover set to the selected
+-- set), use the 'select' function.
+--
+-- A 'UTxOSelection' can be promoted to a 'UTxOSelectionNonEmpty', indicating
+-- that the selected set contains at least one UTxO. To promote a selection,
+-- either use the 'toNonEmpty' function to assert that it is non-empty, or use
+-- the 'select' function to select a single entry.
+--
+module Cardano.Wallet.Primitive.Types.UTxOSelection
+    (
+      -- * Classes
+      IsUTxOSelection
+
+      -- * Types
+    , UTxOSelection
+    , UTxOSelectionNonEmpty
+
+      -- * Construction and deconstruction
+    , empty
+    , fromIndex
+    , fromIndexFiltered
+    , fromIndexPair
+    , toIndexPair
+
+      -- * Promotion and demotion
+    , fromNonEmpty
+    , toNonEmpty
+
+      -- * Indicator functions
+    , isEmpty
+    , isNonEmpty
+    , isMember
+    , isLeftover
+    , isSelected
+    , isSubSelectionOf
+    , isProperSubSelectionOf
+
+      -- * Accessor functions
+    , leftoverBalance
+    , leftoverSize
+    , leftoverIndex
+    , leftoverList
+    , leftoverUTxO
+    , selectedBalance
+    , selectedSize
+    , selectedIndex
+    , selectedList
+    , selectedUTxO
+
+      -- * Modification
+    , select
+    , selectMany
+
+    ) where
+
+import Prelude
+
+import Cardano.Wallet.Primitive.Types.TokenBundle
+    ( TokenBundle )
+import Cardano.Wallet.Primitive.Types.Tx
+    ( TxIn, TxOut )
+import Cardano.Wallet.Primitive.Types.UTxO
+    ( UTxO )
+import Cardano.Wallet.Primitive.Types.UTxOIndex
+    ( UTxOIndex )
+import Control.Monad
+    ( ap, (<=<) )
+import Data.Bool
+    ( bool )
+import Data.Function
+    ( (&) )
+import Data.Generics.Internal.VL.Lens
+    ( over )
+import Data.Generics.Labels
+    ()
+import Data.List.NonEmpty
+    ( NonEmpty )
+import Data.Maybe
+    ( fromMaybe )
+import Data.Tuple
+    ( swap )
+import GHC.Generics
+    ( Generic )
+
+import qualified Cardano.Wallet.Primitive.Types.UTxO as UTxO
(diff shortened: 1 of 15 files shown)


lib/core/test/unit/Cardano/Wallet/Primitive/Types/UTxOSpec.hs
```diff
@@ -54,6 +77,61 @@ spec =
...
+        (UTxO.filter (not . f) u `isNonEmptyProperSubsetOf` u)
+        "UTxO.filter (not . f) u `isNonEmptyProperSubsetOf` u"
+    . cover 10
+        (UTxO.size (UTxO.filter f u) > UTxO.size (UTxO.filter (not . f) u))
+        "UTxO.size (UTxO.filter f u) > UTxO.size (UTxO.filter (not . f) u)"
+    . cover 10
+        (UTxO.size (UTxO.filter f u) < UTxO.size (UTxO.filter (not . f) u))
+        "UTxO.size (UTxO.filter f u) < UTxO.size (UTxO.filter (not . f) u)"
+  where
+    u1 `isNonEmptyProperSubsetOf` u2 = True
```
sevanspowell: Is this True necessary?

jonathanknowles: > Is this True necessary?

It's definitely not necessary!

It's just a stylistic choice so that we can place each condition on a line beginning with `&&`.

jonathanknowles: @sevanspowell I've simplified these expressions in 61f1dc8c4078e70d4dddc7f59c13af2e80ff11a0.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOSelection.hs
```diff
@@ -0,0 +1,379 @@
...
+import qualified Cardano.Wallet.Primitive.Types.UTxO as UTxO
+import qualified Cardano.Wallet.Primitive.Types.UTxOIndex as UTxOIndex
+import qualified Data.Foldable as F
+import qualified Data.List.NonEmpty as NE
+
+--------------------------------------------------------------------------------
+-- Classes
+--------------------------------------------------------------------------------
+
+class HasUTxOSelectionState u where
```
sevanspowell, line 119: Sorry, I have a knee-jerk reaction to typeclasses. I feel it's un-idiomatic Haskell to use typeclasses in this fashion.

In regards to `HasUTxOSelectionState`, using typeclasses purely for overloading is OK, but I don't feel the benefit outweighs the cost. Being a little wordy is OK too:

```
  state :: UTxOSelection -> State
  stateNonEmpty :: UTxOSelectionNonEmpty -> State

  -- or, demote to a common type, define function on common type:
  selection :: UTxOSelectionNonEmpty -> UTxOSelection
  state :: UTxOSelection -> State
```

In regards to `IsUTxOSelection`, maybe you could use higher-order functions instead:

```
newtype UTxOSelection f = UTxOSelection (State f)

selectedList :: UTxOSelection f -> f (TxIn, TxOut)
```

Annoyingly, you're restricted by UTxOIndex, it's defined as follows:

```
data UTxOIndex = UTxOIndex
    { assetsAll
        :: !(Map AssetId (NonEmptySet TxIn))
        -- An index of all entries that contain at least one non-ada asset.
    , assetsSingleton
        :: !(Map AssetId (NonEmptySet TxIn))
        -- An index of all entries that contain exactly one non-ada asset.
    , coins
        :: !(Set TxIn)
        -- An index of all entries that contain no non-ada assets.
    , balance
        :: !TokenBundle
        -- The total balance of all entries.
    , utxo
        :: !(Map TxIn TxOut)
        -- The complete set of all entries.
    }
```

But in it's truest form `type UTxOIndex = Map TxIn TxOut`, the other fields are just implementation details, caching.

I'd suggest redefining UTxOIndex as:

```
-- Note, not exporting constructor
module UTxOIndex (UTxOIndex) where

data UTxOIndex f = UTxOIndex (f (TxIn, TxOut))
-- But really, the exact type doesn't matter, this could have all the other fields too, so long as they're not exposed

-- Observations
assetsAll :: Foldable f => UTxOIndex f -> Map AssetId (NonEmptySet TxIn)

-- Denotation (what this abstraction actually "is")
utxo :: UTxOIndex -> f (TxIn, TxOut)

-- We've removed the "Map" type for convenience, but `f (TxIn, TxOut)` should have all the properties of a `Map` (test this in property tests/laws).
```

Then you can redefine UTxOSelection as:

```
data State f = State
    { leftOver :: !(UTxOIndex List)
    , selected :: !(UTxOIndex f)
    }

selectedList :: UTxOSelection f -> f (TxIn, TxOut)
selectedList (UTxOSelection s) = utxo $ selectedIndex s
```

tldr; use higher-order functions instead of typeclasses.

However, because I can't justify my dislike of typeclasses for things other than `Functor, Applicative, Monad`, and because the work to re-factor is non-trivial, and because I'm hand-waving a lot of detail (which could possibly complicate or make impossible what I've suggested), I'm happy to approve this as-is.

sevanspowell (approved):

lib/core/src/Cardano/Wallet/Primitive/Types/UTxOSelection.hs
sevanspowell: The second point seems like a natural fit because you're trying to define two functions:

```
selectedIndex :: UTxOSelectionPossiblyEmpty -> [(TxIn, TxOut)]
selectedIndex :: UTxOSelectionNonEmpty -> NonEmpty (TxIn, TxOut)
```

With some better names:

```
selectedIndex :: UTxOSelectionList -> List (TxIn, TxOut)
selectedIndex :: UTxOSelectionNonEmpty -> NonEmpty (TxIn, TxOut)
```

And some creative spacing:
```
selectedIndex :: UTxOSelection List -> List (TxIn, TxOut)
selectedIndex :: UTxOSelection NonEmpty -> NonEmpty (TxIn, TxOut)
```

Then some abstraction, we get:

```
selectedIndex :: UTxOSelection f -> f (TxIn, TxOut)
selectedIndex :: UTxOSelection f -> f (TxIn, TxOut)
```

That is, the type of the selection (PossiblyEmpty/NonEmpty) is also the type of the selected list (PossiblyEmpty/NonEmpty), they're in lockstep.

jonathanknowles: @sevanspowell Thank-you for this feedback, it's much appreciated!

It's given me some food for thought.

With the current definition of `UTxOIndex` as it is, I suspect that the modification you suggest will indeed require quite a bit of refactoring.

I'm always searching for ways to simplify / improve these types, so I'll definitely give this some more thought.

One (small) advantage of using type classes here is that the user can write:

- `UTxOSelection`
- `UTxOSelectionNonEmpty`

Without having to specify any extra type parameters. Though admittedly, this could also be accomplished with type synonyms.

PR #2923, opened 2021-09-27
Add `collateral` field to result of `CoinSelection.performSelection`.

jonathanknowles:
## Issue Number

ADP-1037

## Summary

This PR:
- Adds a `collateral` field to the record returned by `CoinSelection.performSelection`.
- Adjusts `SelectionReport{Detailed,Summarized}` to include a report of collateral inputs.
- Removes a now-redundant type parameter from the `CoinSelection.Balance.SelectionResult` type.
- Fixes a small bug with the `computedFee` field of `SelectionReportSummarized`.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
@@ -27,11 +27,16 @@ module Cardano.Wallet.Primitive.CoinSelection
     , SelectionConstraints (..)
     , SelectionParams (..)
     , SelectionError (..)
+    , Selection
+    , SelectionOf (..)
 
     , prepareOutputs
     , ErrPrepareOutputs (..)
     , ErrOutputTokenBundleSizeExceedsLimit (..)
     , ErrOutputTokenQuantityExceedsLimit (..)
+
+    -- * Queries
+    , selectionDelta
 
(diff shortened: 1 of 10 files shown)


lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
```diff
@@ -318,3 +419,125 @@ data ErrOutputTokenQuantityExceedsLimit = ErrOutputTokenQuantityExceedsLimit
...
+makeSelectionReport s = SelectionReport
+    { summary = makeSelectionReportSummarized s
+    , detail = makeSelectionReportDetailed s
+    }
+
+makeSelectionReportSummarized :: Selection -> SelectionReportSummarized
+makeSelectionReportSummarized s = SelectionReportSummarized {..}
+  where
+    computedFee
+        = Coin.distance totalAdaBalanceIn totalAdaBalanceOut
```
jonathanknowles: This is now incorrect. We should use the `selectionDelta` function for an accurate result.

jonathanknowles: Fixed in d66bd9bb50fed1c61614386aeea84a43166c99be.

sevanspowell (approved):

PR #2924, opened 2021-09-27
Add disclaimer about address_pool_gap address discovery in API doc

piotr-iohk:
- [ ] Add disclaimer about address_pool_gap address discovery in API doc

### Comments

Adding this note on recommendation from Jide, who's saying that many partners are trying to generate addresses with cardano-address and than looking into importing them into Shelley wallets. We also had some questions about this on GH recently.

Screenshot from 2021-09-27 18-48-19

### DIFF
--- a/specifications/api/swagger.yaml
+++ b/specifications/api/swagger.yaml
@@ -898,6 +898,11 @@ x-walletAddressPoolGap: &walletAddressPoolGap
 
     Beside, using large gaps is **not recommended** as it may induce important performance degradations. Use at your own risks.
 
+    **IMPORTANT DISCLAIMER 2:** There is no way to `import` addresses generated outside (e.g. using cardano-addresses) into the wallet.
+    Wallet only discovers transactions on its used and unused addresses that are within its currently seen `address_pool_gap`.
+    Transactions on addresses that "belong" to the wallet but happen to be beyond its `address_pool_gap` will not be visible to the wallet.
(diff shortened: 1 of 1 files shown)


specifications/api/swagger.yaml
```diff
@@ -898,6 +898,10 @@ x-walletAddressPoolGap: &walletAddressPoolGap
 
     Beside, using large gaps is **not recommended** as it may induce important performance degradations. Use at your own risks.
 
+    **IMPORTANT DISCLAIMER 2:** There is no way to `import` addresses generated outside (e.g. using cardano-addresses) into the wallet.
+    Wallet only discovers transactions on its used and unused addresses that are within its currently seen `address_pool_gap`.
+    Transactions on addresses that "belong" to the wallet but happen to be beyond its `address_pool_gap` will not be visible on the wallet. See: [CIP-1852](https://github.com/cardano-foundation/CIPs/blob/master/CIP-1852/CIP-1852.md).
```
HeinrichApfelmus:
```suggestion
    Transactions on addresses that "belong" to the wallet but happen to be beyond its `address_pool_gap` will not be visible to the wallet.
    This is a technical limitation of the industry standard [BIP-44][]. See also [CIP-1852][]
      [cip-1852]: https://github.com/cardano-foundation/CIPs/blob/master/CIP-1852/CIP-1852.md
      [bip-44]: https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki
```

I suggest that we mention that it's "not our fault" by referring to an "industry standard". 😅

HeinrichApfelmus (approved):

specifications/api/swagger.yaml
piotr-iohk: Corrected.

PR #2931, opened 2021-09-29
Give callers of `selectAssets` more control over UTxO selection

jonathanknowles:
## Issue Numbers

ADP-1037
ADP-1118

## Summary

This PR splits the `utxoAvailable` parameter of `selectAssets` into two:
- `utxoAvailableForInputs`.
- `utxoAvailableForCollateral`.

Note that these two UTxO sets **_are_** allowed to intersect: the ledger does **_not_** require that they are disjoint. So it's perfectly acceptable to supply the same set in both cases.

However, some callers of `selectAssets` may prefer to **pre-select** a subset of inputs within `utxoAvailableForInputs`. This is possible with the following constructors for `UTxOSelection`:

- [`fromIndexFiltered`](https://github.com/input-output-hk/cardano-wallet/blob/6626489542ce442a8451fb46fb5339346e62b4ee/lib/core/src/Cardano/Wallet/Primitive/Types/UTxOSelection.hs#L196)
- [`fromIndexPair`](https://github.com/input-output-hk/cardano-wallet/blob/6626489542ce442a8451fb46fb5339346e62b4ee/lib/core/src/Cardano/Wallet/Primitive/Types/UTxOSelection.hs#L205)

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -107,6 +107,7 @@ module Cardano.Wallet
 
     -- ** Payment
     , getTxExpiry
+    , SelectAssetsParams (..)
     , selectAssets
     , readWalletUTxOIndex
     , assignChangeAddresses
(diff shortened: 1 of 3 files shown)


lib/core/src/Cardano/Wallet.hs
```diff
@@ -1411,6 +1415,18 @@ calcMinimumCoinValues ctx outs = do
     nl = ctx ^. networkLayer
     tl = ctx ^. transactionLayer @k
 
+-- | Parameters for the 'selectAssets' function.
+--
+data SelectAssetsParams s result = SelectAssetsParams
```
paweljakubas, line 1420: :+1:

paweljakubas (approved):
lgtm

PR #2933, opened 2021-09-29
remove unused node-config volume

piotr-iohk:
- [ ] remove unused node-config volume

### Comments
There was unused node-config volume on the node docker, leftover from https://github.com/input-output-hk/cardano-wallet/pull/2830, which seems to cause problems on starting docker-compose with newer version of the node (1.29.0 -> 1.30.1) on top of older db.

### Issue Number

ADP-1165


Anviking (approved):
🆗

PR #2937, opened 2021-09-30
Update to ghc 8.10.7 and latest haskell.nix, nixpkgs-unstable and iohk-nix

hamishmack:
We are trying out an `aarch64-darwin` build of `daedalus` on https://ci.zw3rk.com/project/daedalus.  It needs these updates.

See https://github.com/input-output-hk/daedalus/pull/2684 and https://github.com/input-output-hk/daedalus/pull/2683


hamishmack:
@rvl, I am not sure how useful this is to you.  I see you already have PRs for these updates.  I did run into an issue with `cardano-config` needing `git` that required addition of the `build-tools` override (that might have only been when I was using the master branch of `cardano-node`).  `haskell-language-server` also seemed to need a bump to work with ghc 8.10.7.

hamishmack:
We are using master again now that #2969 has been merged.

PR #2942, opened 2021-10-01
Refactor DBLayer to use in-memory Checkpoints

HeinrichApfelmus:
### Issue number

ADP-1169

### Overview

In this step, we introduce a type Checkpoints a which stores all checkpoints of the wallet state in memory. The current ad-hoc cache of the latest checkpoints is replaced by a DBVar that stores these checkpoints instead.

### Details

* Keep the `Checkpoints` type stupid and simple for now. It will be replaced later.
* This pull request will (temporarily) increase the memory requirements for the wallet, as the collection of checkpoints is kept in-memory. However, memory usage will go back to normal (and better) once delta encodings are integrated.

### Comments

Merge PR #2841 before this one, because this pull request is based on the branch of the former.

### DIFF
--- a/lib/core/src/Cardano/Wallet/DB.hs
+++ b/lib/core/src/Cardano/Wallet/DB.hs
@@ -42,10 +42,11 @@ import Cardano.Wallet.Primitive.AddressDerivation
 import Cardano.Wallet.Primitive.Model
     ( Wallet )
 import Cardano.Wallet.Primitive.Types
-    ( BlockHeader
+    ( ChainPoint
     , DelegationCertificate
     , GenesisParameters
     , Range (..)
+    , Slot
     , SlotNo (..)
     , SortOrder (..)
     , WalletId
@@ -161,7 +162,7 @@ data DBLayer m s k = forall stm. (MonadIO stm, MonadFail stm) => DBLayer
 
     , listCheckpoints
         :: WalletId
-        -> stm [BlockHeader]
+        -> stm [ChainPoint]
         -- ^ List all known checkpoint tips, ordered by slot ids from the oldest
         -- to the newest.
 
@@ -301,12 +302,13 @@ data DBLayer m s k = forall stm. (MonadIO stm, MonadFail stm) => DBLayer
 
     , rollbackTo
         :: WalletId
-        -> SlotNo
-        -> ExceptT ErrNoSuchWallet stm BlockHeader
-        -- ^ Drops all checkpoints and transaction data after the given slot.
+        -> Slot
+        -> ExceptT ErrNoSuchWallet stm ChainPoint
+        -- ^ Drops all checkpoints and transaction data which
+        -- have appeared after the given 'ChainPoint'.
         --
-        -- Returns the actual slot to which the database has rolled back. This
-        -- slot is guaranteed to be earlier than (or identical to) the given
+        -- Returns the actual 'ChainPoint' to which the database has rolled back.
+        -- Its slot is guaranteed to be earlier than (or identical to) the given
         -- point of rollback but can't be guaranteed to be exactly the same
         -- because the database may only keep sparse checkpoints.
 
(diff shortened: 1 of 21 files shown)


paweljakubas (commented):
I run through logic and it seems to proceed as I would expect. If all tests pass it is giving crucial confidence that replacing SlotNo with SlotPoint and BlockHeader with ChainPoint is not doing any harm. What about db migrations? No way they are going to be affected? Maybe it would be good if @piotr-iohk plays with migrations using this branch.

I will give another look, even more detailed and style related, tomorrow (until now I was mainly occupied with how it works and why)

HeinrichApfelmus:
> I run through logic and it seems to proceed as I would expect. If all tests pass it is giving crucial confidence that replacing SlotNo with SlotPoint and BlockHeader with ChainPoint is not doing any harm. What about db migrations? No way they are going to be affected? Maybe it would be good if @piotr-iohk plays with migrations using this branch.
> 
> I will give another look, even more detailed and style related, tomorrow (until now I was mainly occupied with how it works and why)

Thanks Pawel! 😊

The DB migrations should not be affected — this PR does not change the database format in any way. The database is actually a little buggy in that it stores a checkpoint for the first slot after genesis, `At (SlotNo 0)`, but no checkpoint for genesis, `Origin`. This does not affect testnet or mainnet, but it has affected local clusters that start a new network, and where there is a rollback to genesis.

I don't think that it makes to fix this in the database file format right now, because a future PR will radically change the format anyway, adding incremental updates. However, I do think that it makes sense to at least use the correct representation for the in-memory type. Unit tests will pass as long as they never try to roll back to `Origin` rather than to `At (SlotNo 0)`. In fact, before the change in types, the unit tests were not actually able to rollback further than `At (SlotNo 0)`.

paweljakubas (approved):
lgtm

lib/core/src/Cardano/Wallet/DB.hs
```diff
@@ -301,12 +302,13 @@ data DBLayer m s k = forall stm. (MonadIO stm, MonadFail stm) => DBLayer
 
     , rollbackTo
         :: WalletId
-        -> SlotNo
-        -> ExceptT ErrNoSuchWallet stm BlockHeader
-        -- ^ Drops all checkpoints and transaction data after the given slot.
+        -> Slot
```
rvl, line 305: Given that we don't require the database to be rolled back further than a few thousand blocks, a value of `Origin` should never be passed for this argument of `rollbackTo`.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -2050,27 +2048,191 @@ genesisParametersFromEntity (Wallet _ _ _ _ _ hash startTime) =
...
+    apply (Insert key a) = Map.insert key a
+    apply (Delete key) = Map.delete key
+    apply (Adjust key da) = Map.adjust (apply da) key
+
+{- NOTE [PointSlotNo]
+
+'SlotNo' cannot represent the genesis point.
+
+Historical hack. The DB layer can't represent 'Origin' in the database,
+instead we have mapped it to 'SlotNo 0', which is wrong.
```
rvl, line 2069: It shouldn't matter; we don't need rollback to genesis.

rvl (approved):
Great :+1:

HeinrichApfelmus:
bors merge

lib/core/src/Cardano/Wallet/DB.hs
HeinrichApfelmus: I agree that this will not happen in practice on testnet and mainnet. However, the ouroboros-network library does represent points on the blockchain with an explicit origin, and a rollback to genesis can happen in principle (and does happen on local test clusters). By using a representation that is more closely aligned with ouroboros-network (i.e. `Slot` instead of `SlotNo`), the code actually becomes simpler, as we can remove conversion functions such as

```haskell
    pseudoPointSlot :: ChainPoint -> SlotNo
    pseudoPointSlot ChainPointAtGenesis = W.SlotNo 0
    pseudoPointSlot (ChainPoint slot _) = slot
```

PR #2943, opened 2021-10-01
ADP-919: Sign Transactions Endpoint

KtorZ:
- :round_pushpin: **Remove withdrawal from the sign-transaction API.**
    Withdrawals are 'needed' in this context when they refer to an external wallet (i.e. via a mnemonic phrase) so that, the required signature for such withdrawal can be produced (derived from the root key associated with the mnemonic). However:

  (a) The use-case for such external withdrawals stem from the ITN rewards, which were distributed more than 14 months ago to Shelley wallets that had participated in the incentivized testnet. It is reasonable to think that most concerned users have already redeemed their rewards. Those who haven't can still do it via the old transaction endpoint.

  (b) Having withdrawal as part of this API endpoint is awkward and confusing. Instead, it would be more logical / practical to perform the signature during the construction of the transaction; that is, when constructing a transaction with an external withdrawal, the result would already contain a signature from the associated mnemonic.  Such transaction would then still need to be passed to the sign-transaction endpoint to gather signature from the wallet itself.

  Alternatively, not supporting this feature through the new transaction construction endpoint is also a very sensible choice. Regardless of the option, worrying about it _now_ is unnecessary / a waste of time.

- :round_pushpin: **Extract vk signing from 'mkTx'**
    Note that, doing so I have slightly altered the semantic of the function which can no longer return a 'ErrKeyNotFoundForAddress'.

  Not finding the relevant key for a given input will now simply skip that input. Ultimately, the code used to yield to a 500 server errors BUT, that case was truly unreachable. Proof is that there's no end-to-end test which happen to be looking for that error case since it cannot (must not) occur under any circumstances. It really is a deep server error.

  This change does not however makes the error 'silent', but merely defers it to the submission. If a necessary signature is missing, the transaction will eventually be rejected by the local node which will inform the user about the missing key.

  The rationale for changing the semantic is that, in the context of signing 'external' transactions, we want to allow _other_ inputs than ours. This is crucial in the context of Alonzo since transactions may contain and consume script inputs. So it's important for this operation to skip such inputs and simply care about signing inputs relevant of the wallet.

  Note that, as a TODO, we should also now check for extra required signatures which can be specified in the transaction body.

- :round_pushpin: **Extend transaction layer with 'addVkWitnesses' and use it.**
  
- :round_pushpin: **Move withdrawal signing into 'signTransaction'**
    So that we avoid too many repetition in the upcoming implementation of addVkWitnesses, and also, centralize the creation of signatures into one place. Note that this does not change the semantic of the function and it works regardless of where the reward account comes from (external or internal).

- :round_pushpin: **Implement 'addVkWitnesses' for the shelley target.**
    A bit of repetition here because of the encapsulated era which can't leave its scope. All-in-all, it's not a big deal and someone has to deal with the multiple eras into some place, might it be via a type-class or via similar branches in a case like here. Note also that in case of Byron transactions, this does not work, which is likely a fair limitation.

- :round_pushpin: **Add basic end-to-end test for signing**
    The test does a full cycle: construction > signing > submission.

    However, some elements of the signing response are currently stubbed out. Obtaining them (the serialized body and witnesses) is _doable_ but also look like extra / needless work :thinking_face: ? I'd argue that returning both the full serialized tx AND the serialized body AND the serialized witnesses is redundant. It seems like the interface currently is the serialized transaction (also with the construction endpoint) so it'd make more sense to only return this. To be discussed.

- :round_pushpin: **Fix unit test after changing semantic of transaction construction**
    This no longer fails on unknown inputs, but skip the input.

- :round_pushpin: **Add test rejecting unsigned Tx**
  
- :round_pushpin: **Add new test for signing withdrawals**
  
- :round_pushpin: **Fix compilation error on withdrawal test.**
  
- :round_pushpin: **Add counter-example serialized transaction.**
  
- :round_pushpin: **Add unit test for failed CBOR decoding of tx w/ withdrawal**
  This unit test comes from failed integration test for signing transactions.

- :round_pushpin: **Decode transaction bytes from base16 before deserialising them :facepalm:.**
  
- :round_pushpin: **Rely on 'SealedTx' JSON instance for serialization (Base64)**
    And do not provide raw bytes which are automatically serialized as Base16

- :round_pushpin: **Factor out submission code from all new-tx sign tests.**
  
- :round_pushpin: **Remove redundant response elements from sign transaction result.**
    The rationale is that, these elements are already present in the serialized transaction and, we assume that a client able to consume parts of the serialized transaction and put them together manually into a full serialized transaction would know anyway how to handle CBOR in the first place. Thus, splitting the full serialized transactions into smaller part if left as a client concern.

- :round_pushpin: **Add e2e test scenario illustrating requiredSigners for transactions**
    This currently fails, as expected, since the feature isn't yet implemented in the API handler.

- :round_pushpin: **Include signatures for 'extra-key-witnesses' in transactions**
    Note however that the wallet will only include signatures for keys it knows of, that is, for keys that are within the current visible window of the sequential state. That is okay because we do not expect the wallet to actually 'have' keys outside of the tracked window.


lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
```diff
@@ -7,6 +7,7 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedLabels #-}
+{-# LANGUAGE RankNTypes #-}
```
KtorZ: Whoops. Added for implementing a function I didn't end up using. This got included by mistake.

KtorZ:
TODO: 

- [x] More test scenarios to cover additional use-cases (see end of e2e spec file, some bullet points)
- [x] Support for extra required signatures which is a special field introduced in Alonzo that may contain extra signatures. Those keys need to be resolved somehow but it's unclear how (the wallet doesn't keep track of keys per se, but of addresses...) 
- [x] Challenge / Discuss the current choice of response for the sign-transaction API. I believe that the current response is redundant and adds extra work for little benefits? (see commit notes)

cc @sevanspowell 

(NB: the second point may be a known limitation we accept for now in order to get things unlocked a bit).

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -882,6 +889,102 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
...
+            [ expectResponseCode HTTP.status202
+            ]
+
+    -- TODO: Moar tests scenarios to cover in the context of sign-transactions
+    --
+    -- - Signing with non-own / scripts inputs
+    -- - Signing with pre-existing witnesses
+    -- - Signing with collaterals
+    -- - Signing with extra required signatures
+    -- - ideas?
```
piotr-iohk: There are actually several cases in this file that were created only up to "Construct" that, I guess, can be now filled with sign and submit steps. They all have relevant comment:
```
TODO: now we should sign it and send it in two steps
```
or
```
TODO: sign
```

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -882,6 +889,102 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
...
+
+        -- Submit tx
+        let submitEndpoint = Link.postExternalTransaction
+        let headers = Headers
+                [ ("Content-Type", "application/octet-stream")
+                , ("Accept", "application/json")
+                ]
+        r <- request @ApiTxId ctx submitEndpoint headers (NonJson $ BL.fromStrict sealedTx)
+        verify r
+            [ expectResponseCode HTTP.status500
```
piotr-iohk: Perhaps it would be nice to "catch" `500` and have some `40x` here?

KtorZ: We thought the same, though not really relevant to this PR but I think the submitEndpoint should not return 500 errors for invalid transactions, since this is precisely a client error.

lib/core/test/unit/Cardano/WalletSpec.hs
```diff
@@ -1258,17 +1258,18 @@ dummyTransactionLayer = TransactionLayer
...
-            (xprv, Passphrase pwd) <- withEither
-                (ErrKeyNotFoundForAddress addr) $ keystore addr
-            let sigData = tx ^. #txId . #getHash
-            let sig = CC.unXSignature $ CC.sign pwd (getKey xprv) sigData
-            return $ xpubToBytes (getKey $ publicKey xprv) <> sig
+        let wit = forMaybe (NE.toList $ view #inputs cs) $ \(_, TxOut addr _) -> do
+                (xprv, Passphrase pwd) <- keystore addr
+                let sigData = tx ^. #txId . #getHash
+                let sig = CC.unXSignature $ CC.sign pwd (getKey xprv) sigData
+                return $ xpubToBytes (getKey $ publicKey xprv) <> sig
```
KtorZ, line 1265: Not exactly relevant to this PR, but necessary because of the change of semantic when signing / the removal of the `ErrKeyNotFoundForAddress`

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -306,6 +309,11 @@ instance Arbitrary AnyShelleyBasedEra where
 
 decodeSealedTxSpec :: Spec
 decodeSealedTxSpec = describe "SealedTx serialisation/deserialisation" $ do
+    it "tx with withdrawal" $ do
+        let bytes = unsafeFromHex "84a70081825820410a9cd4af08b3abe25c2d3b87af4c23d0bb2fb7577b639d5cfbdfe13a4a696c0c0d80018182583901059f0c7b9899793d2c9afaeff4fd09bedd9df3b8cb1b9c301ab8e0f7fb3c13a29d3798f1b77b47f2ddb31c19326b87ed6f71fb9a27133ad51b000001001d19d714021a000220ec03198d0f05a1581de1fb3c13a29d3798f1b77b47f2ddb31c19326b87ed6f71fb9a27133ad51b000000e8d4a510000e80a0f5f6"
+        let sealedTx = sealedTxFromBytes bytes
+        sealedTx `shouldSatisfy` isRight
+
```
KtorZ, line 316: We stumbled upon this while testing, and got surprised by the result so we made it a golden test. In the end, the `roundtrip` test below is actually pretty weak as it doesn't test most constituents of transactions. In particular, withdrawals aren't tested at all by this property :grimacing:

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
piotr-iohk: I've issued bug ticket for this ADP-1172.

sevanspowell (approved):

piotr-iohk (approved):

KtorZ:
bors +

_pretty please :pray:_

KtorZ:
I didn't miss you.

PR #2947, opened 2021-10-05
Add Nix expressions required to use newest fmt library

sevanspowell:
- We want to update to the newest fmt library (fmt-6.3.0) to make use of new Buildable instances.
- Ran `niv update haskell.nix` to get latest hackage.nix, which provides us with the Nix expressions for fmt-6.3.0.
- There is some build issue with haskell-language-server-1.2.0.0 and the version of haskell.nix we've pinned, so update haskell-language-server to 1.4.0.0.
- Update the index-state so we can build it.
- `./nix/regenerate.sh`: regenerate the materialized files


nix/materialized/haskell-language-server/.plan.nix/haskell-language-server.nix
```diff
@@ -101,7 +103,7 @@
         };
       exes = {
         "haskell-language-server" = {
-          depends = ((((((((((((((([
+          depends = (((((((((((((((((([
```
jonathanknowles, line 106: 😄

jonathanknowles (approved):

HeinrichApfelmus:
The "Stack Rebuild" item in buildkite fails with

```
bash: /nix/store/wx1vk75bpdr65g6xwxbj4rw0pk04v5j3-glibc-2.27/lib/libc.so.6: version `GLIBC_2.33' not found (required by /nix/store/yn7nd41grg2mzvzliphalk70hs0rfpdr-ncurses-6.2/lib/libncursesw.so.6)
```

This prevents merging, I suppose?

No idea what's going on, but @rvl mentioned an issue similar to this in our internal nix-best-practices channel.

This "GLIBC_2.33 issue" also prevents the update to GHC 8.10.7 in #2937

PR #2951, opened 2021-10-06
Balance endpoint errors

paweljakubas:
adp-1179

- [x] I have added ErrAlreadyBalanced error
- [x] I have added ErrTransactionWithCollateral error

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -2220,12 +2220,13 @@ balanceTransaction ctx genChange (ApiT wid) body = do
     --redeemers will be updated when evaluated, so for already balanced tx they are
     --going to be nonzero.
     let executionFee = calcScriptExecutionCost tl pp sealedTxIncoming
+    let (Tx _id _fee collaterals _inps outs wdrlMap mdM _validity) = txIncoming
 
     if areOutputsCovered txIncoming executionFee then
         liftHandler $ throwE ErrBalanceTxTxAlreadyBalanced
+    else if not (null collaterals) then
+        liftHandler $ throwE ErrBalanceTxCollateralsPresent
     else do
-        let (Tx _id _fee _coll _inps outs wdrlMap mdM _validity) = txIncoming
-
         ((FeeEstimation feeMin _), unsignedtx) <- withWorkerCtx ctx wid liftE liftE $ \wrk -> do
 
             (acct, _, _) <- liftHandler $ W.readRewardAccount @_ @s @k @n wrk wid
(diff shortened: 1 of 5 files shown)


lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2214,81 +2215,88 @@ balanceTransaction
...
+    --redeemers will be updated when evaluated, so for already balanced tx they are
+    --going to be nonzero.
+    let executionFee = calcScriptExecutionCost tl pp sealedTxIncoming
+    let (Tx _id _fee collaterals _inps outs wdrlMap mdM _validity) = txIncoming
+
+    if areOutputsCovered txIncoming executionFee then
+        liftHandler $ throwE ErrBalanceTxTxAlreadyBalanced
+    else if not (null collaterals) then
+        liftHandler $ throwE ErrBalanceTxCollateralsPresent
+    else do
```
KtorZ, line 2229: May I suggest to avoid this if / then / do / else do pattern. We are in a `Handler` Monad here, so 

```hs
         liftHandler $ throwE ErrBalanceTxTxAlreadyBalanced
```

will short-circuit the execution of the rest of the handler. Not only this is cleaner / less imperative style, but it will also avoid to indent further the rest of the code, causing needless conflicts and making the diff a bit harder to read :blush:

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2305,6 +2313,12 @@ balanceTransaction ctx genChange (ApiT wid) body = do
     txIncoming = decodeTx tl sealedTxIncoming
     toTxOut (TxChange addr amt assets _) =
         TxOut addr (TokenBundle amt assets)
+    getExternalInps (ApiExternalInput (ApiTxIn (ApiT hashTx) ix) (ApiTxOut _ _ (Quantity amt) _)) =
+        (TxIn hashTx ix, Coin $ fromIntegral amt)
+    externalIns = getExternalInps <$> apiExternalInps
+    areOutputsCovered (Tx _ feeM _ _ outs _ _ _) execUnitsCost =
+        sumCoins (map snd $ L.nubBy (\x y -> fst x == fst y) externalIns) ==
```
paweljakubas, line 2320: maybe we need here `externalIns ++ colls ++ inps` the last two from `(Tx _ feeM colls inps outs _ _ _)` :thinking:  but then the unpended example does not work (and this is just complete normal tx)

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2305,6 +2313,12 @@ balanceTransaction ctx genChange (ApiT wid) body = do
...
+        sumCoins [fromMaybe (Coin 0) feeM, sumCoins $ txOutCoin <$> outs, execUnitsCost]
```
KtorZ, line 2321: Hmmm. knowing if the transaction is balanced is more complicated. There are many parts to the equation, including withdrawals and deposit / refund. I also don't understand the rationale? Why would it be an error to try to balance an already balanced transaction? It should actually work ok as a no-op, and not error out :thinking:

paweljakubas: I wonder if coin selection survives already balanced tx, there are NonEmpty's used there

paweljakubas: I will find out

KtorZ: Well, it should :)

PR #2952, opened 2021-10-06
ADP-1183 (Part 1): Balancing execution units 

KtorZ:
- :round_pushpin: **Add tx max script execution units to ProtocolParameters**
  
- :round_pushpin: **Calculate max script execution cost of transactions instead of relying on redeemers' ex units.**
    In practice, we want to actually execute a script to set the right value for each redeemer. And not the other way around. So, unless the transaction is fully constructed, the ExUnits value on redeemers is bonkers.

- :round_pushpin: **Allow changing ExUnits of redeemers on sealed tx**
    Doing so, I've also renamed ExtraTxBodyContent -> TxUpdate since (a) it's not only about extra content, (b) it does not only concern the transaction body, but also witnesses now.

- :round_pushpin: **Rework balance-tx to always use max execution units.**
    This is because, we can't actually predict script execution cost
  before the full final transaction is constructed. So, as a first step,
  we use the biggest approximation we can use which is to always
  consider the max execution steps and memory. While this is NOT okay
  for production code, it is sufficient to kick off preliminary testing
  and integration work.

  The goal is to now, in further commits, adjust the ex units of each
  redeemers after balancing, changing the change output consequently.
  Note that this should not in practice change the cost of execution of
  a script, if it does, then we push back on this by arguing that the
  underlying script validator is bonkers. Developers ought not to do
  this.

- :round_pushpin: **Fix typos in comments**
  Co-authored-by: paweljakubas <jakubas.pawel@gmail.com>
Co-authored-by: Jonathan Knowles <jonathan.knowles@iohk.io>
- :round_pushpin: **Calculate script integrity hash when updating pre-constructed tx**
    This is required for any transaction which includes scripts, redeemers or datums as witnesses.

- :round_pushpin: **Write e2e scenario using balance+submit on transaction containing Plutus script(s).**
  
- :round_pushpin: **Rework tests using pre-constructed Shelley transactions**
    - Avoid hard-coding names in the test file, but instead, list files in the test directory.
  - Use pre-constructed files for all tests, instead of having some inlined hex-binary in the test module.
  - Provide more coverage and fix tests actually wrong now that the script integrity hash changes on calls to updateSealedTx

- :round_pushpin: **replace plutus pre-generated transactions with newly generated, cbor-only binaries.**
  
- :round_pushpin: **Remove redundant (with integration scenarios) compatibility JSON golden spec for scripts.**
  
- :round_pushpin: **Cleanup balance endpoint and reduce noise.**
    There were a lot of things in this handler which were not necessary, as well as many intermediary names which would just add to the noise when reading it. In particular, there's no need to lookup the reward account of the wallet (and it's actually even wrong, because withdrawals may be completely external). Wanting to re-use existing functions is quite flawed in this context because many of the pre-conditions for those functions don't actually hold in the generic balance-tx handler.

- :round_pushpin: **Wire collateral requirements into balance endpoint handler**
  
- :round_pushpin: **Add feePadding to TransactionCtx, used by calcMinimumCost**
  
- :round_pushpin: **Add fee padding to transaction context for balancing**
  Co-authored-by: Johannes Lund <johannes.lund@iohk.io>

- :round_pushpin: **Fix ordering of argument for 'subtract' currently causing underflow.**
  
- :round_pushpin: **Rework balance transaction request body to enable passing redeemers.**
    Because coin-selection will change the order of inputs, we can't construct transactions with redeemers already in place. Redeemers need to be constructed after the facts. Thus, we pass them explicitely in a form that is isomorphic to the ledger's (ScriptPurpose, Data). This way, it's easier to construct transactions once balanced.

### Comments

To be done in a next PR:

- Calculate the actual execution cost for all redeemers after balancing (using [evaluateTransactionExecutionUnits](https://input-output-hk.github.io/cardano-node/cardano-api/lib/Cardano-Api-Fees.html#v:evaluateTransactionExecutionUnits) from the `cardano-api`, or its equivalent from the `cardano-ledger-specs`).

- Replace the execution units of each redeemers with their correspond value. 

- Adjust the change output's value accordingly. 

It should actually not be necessary to re-run the selection at all if we indeed assume that changing the change output's value should not change the execution cost. If it does, then the local node will eventually reject the transaction, and that's none of our business.  

### Issue Number

ADP-1183


lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2213,23 +2213,37 @@ balanceTransaction
     -> ApiBalanceTransactionPostData n
     -> Handler (ApiConstructTransaction n)
 balanceTransaction ctx genChange (ApiT wid) body = do
+    -- TODO: Most of the body of this function should really belong to
+    -- Cardano.Wallet to keep the Api.Server moduel free of business logic!
```
paweljakubas:
```suggestion
    -- Cardano.Wallet to keep the Api.Server module free of business logic!
```

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2268,17 +2280,22 @@ balanceTransaction ctx genChange (ApiT wid) body = do
             }
             transform
 
-        fee <- liftHandler $ W.estimateFee runSelection
-        return (fee, unsignedtx)
-
     let UnsignedTx colls inps _outs change _wdrl = unsignedtx
-    let extraBody = ExtraTxBodyContent
+    -- TODO: At this stage, we set all execution units for all redeemers to the
+    -- max cost, which is guarantee to succeed (given the coin selection above
```
paweljakubas:
```suggestion
    -- max cost, which is guaranteed to succeed (given the coin selection above
```

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2268,17 +2280,22 @@ balanceTransaction ctx genChange (ApiT wid) body = do
...
+    -- was done with the same assumption) but also terribly inneffective when it
```
paweljakubas:
```suggestion
    -- was done with the same assumption) but also terribly ineffective when it
```

lib/core/src/Cardano/Wallet/Transaction.hs
```diff
@@ -199,25 +201,26 @@ data TransactionLayer k tx = TransactionLayer
 
     , updateTx
         :: tx
-        -> ExtraTxBodyContent
+        -> TxUpdate
```
paweljakubas, line 205: yeah, TxUpdate is definitely better name :sweat_smile:

lib/core/src/Cardano/Wallet/Transaction.hs
```diff
@@ -199,25 +201,26 @@ data TransactionLayer k tx = TransactionLayer
...
--- `updateTx`.
-data ExtraTxBodyContent = ExtraTxBodyContent
+-- | Describes modifications that can be made to a `Tx` using `updateTx`.
+data TxUpdate = TxUpdate
     { extraInputs :: [TxIn]
     , extraCollateral :: [TxIn]
        -- ^ Only used in the Alonzo era and later. Will be silently ignored in
        -- previous eras.
     , extraOutputs :: [TxOut]
+    , newExUnits :: ScriptWitnessIndex -> ExecutionUnits -> ExecutionUnits
```
paweljakubas, line 218: :+1:

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -555,38 +556,58 @@ noExtraTxBodyContent = ExtraTxBodyContent [] [] [] id
...
-            Right $ Cardano.ShelleyTxBody shelleyEra (adjust ebc shelleyEra bod) scripts scriptData aux val
+    modifyLedgerTx ebc (Cardano.ShelleyTxBody shelleyEra bod scripts scriptData aux val) =
+            Right $ Cardano.ShelleyTxBody shelleyEra
+                (adjustBody ebc shelleyEra bod)
+                scripts
+                (modifyRedeemers scriptData)
+                aux
+                val
       where
+        modifyRedeemers :: Cardano.TxBodyScriptData era -> Cardano.TxBodyScriptData era
```
paweljakubas, line 625: :+1:

paweljakubas (approved):
LGTM! Big shout for the comments in Cardano.Server module! What about adding property test that starts with some SealedTx makes updating of redeemer execution units and as a check we lookup redeemers and  see if the execution cost were really properly updated? Redundant or worth having? 

Also as now we have `TxUpdate` we need to change accordingly here:
1. https://github.com/input-output-hk/cardano-wallet/blob/master/lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs#L1654
2. https://github.com/input-output-hk/cardano-wallet/blob/master/lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs#L1728

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
paweljakubas: maybe it is absolutely obvious `modifyRedeemers` is correct, but feel some property test showing the case would not be harmful;-)

paweljakubas: What about updating this property https://github.com/input-output-hk/cardano-wallet/blob/master/lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs#L1749 ?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2213,23 +2213,37 @@ balanceTransaction
...
+
+        -- NOTE: It is not possible to know the script execution cost in
+        -- advance because it actually depends on the final transaction. Inputs
+        -- selected as part of the fee balancing might have an influence on the
+        -- execution cost.
+        -- However, they are bounded so it is possible to balance the
+        -- transaction considering only the maximum cost, and only after, try to
+        -- adjust the change and ExUnits of each redeemer to something more
+        -- sensible than the max execution cost.
+        let maxCost = maxScriptExecutionCost tl pp sealedTxIncoming
```
Anviking: It just occurred to me: `actualCost <= maxCost` only holds if `pp` is the protocol parameters active when the tx is accepted, i.e. if `pp` and the protocol parameter hash match.

Not sure if we, or the client will set the protocol parameter hash, but we should make sure it always matches with `pp`.

lib/core/src/Cardano/Wallet/Transaction.hs
```diff
@@ -199,25 +201,26 @@ data TransactionLayer k tx = TransactionLayer
...
+       -- ^ Adjust execution units on existing redeemers.
     , newFee :: Coin -> Coin
```
Anviking, line 220: I thought I went with `modifyFee` here, which would make more sense when it's a `a -> a`. If you feel like renaming to say `modifyExUnits` / `modifyFee`, feel free.

Anviking (commented):
Sounds sane! Will have another look tomorrow with fresh eyes, but feel free to merge before that

lib/core/src/Cardano/Wallet/Transaction.hs
```diff
@@ -156,13 +158,13 @@ data TransactionLayer k tx = TransactionLayer
...
 
-    , calcScriptExecutionCost
+    , maxScriptExecutionCost
         :: ProtocolParameters
             -- Current protocol parameters
         -> tx
             -- The constructed transaction that could contain plutus scripts
         -> Coin
-        -- ^ Compute an execution costs of scripts in a given transaction.
+        -- ^ Compute the maximum execution costs of scripts in a given transaction.
```
```suggestion
        -- ^ Compute the maximum execution cost of scripts in a given transaction.
```

paweljakubas:
@KtorZ maybe it would be good also to have a check like this for max execution unit prices in network's integration test -> https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Network.hs#L78

lib/core/src/Cardano/Wallet/Api/Server.hs
KtorZ: The protocol parameter hash is only needed for validators who depends on it. I'd expect the PAB (or really, anyone constructing the transaction) to set it to what is expected.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -555,38 +556,58 @@ noExtraTxBodyContent = ExtraTxBodyContent [] [] [] id
...
-    modifyLedgerTxBody
-        :: ExtraTxBodyContent
+    modifyLedgerTx
+        :: forall era. ()
+        => TxUpdate
         -> Cardano.TxBody era
         -> Either ErrUpdateSealedTx (Cardano.TxBody era)
-    modifyLedgerTxBody ebc (Cardano.ShelleyTxBody shelleyEra bod scripts scriptData aux val) =
-            Right $ Cardano.ShelleyTxBody shelleyEra (adjust ebc shelleyEra bod) scripts scriptData aux val
+    modifyLedgerTx ebc (Cardano.ShelleyTxBody shelleyEra bod scripts scriptData aux val) =
```
Anviking, line 587: Re the name change… It still takes a `TxBody` 🤷‍♂️

Anviking (approved):
lgtm! (modulo some test would be nice)

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
KtorZ: It does indeed, but that's because the cardano-api has made the choice of including script witnesses inside a type called `TxBody` :grimacing: ... which is actually not at all reflecting the `TxBody`. Script witnesses are, witnesses.

KtorZ: The function is actually called `modifyLedgerTx` but it's not a "ledger tx" (if reading this as a tx from the cardano-ledger-specs), it's a cardano-api's tx.

KtorZ:
> (modulo some test would be nice)

I am working on adding the full end-to-end tests. I thought there was already more tests for the balancing endpoint but I discovered that, not so much :grimacing: ..

KtorZ:
> @KtorZ maybe it would be good also to have a check like this for max execution unit prices in network's integration test -> https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/Network.hs#L78

Yes and.. not :smile: ... I haven't added the max execution units to the API yet because that's extra side-work which is really not related to this feature. So I would rather not do it _now_ to not add needless noise.

lib/shelley/test/data/plutus-integration/ping-pong_1-1.json
```diff
@@ -0,0 +1,8 @@
+{
```
Anviking: Did you mean to have this in the `lib/core-integration/test/` instead?

I couldn't run the ping-pong test without copying this from `lib/shelley/test/data/plutus-integration` to `lib/core-integration/test/data/plutus-integration`.

KtorZ: Yes. I moved it before pushing the PR, but didn't commit :|

KtorZ:
bors retry

KtorZ:
bors retry

PR #2953, opened 2021-10-07
Test fundamental properties of `CoinSelection.performSelection`.

jonathanknowles:
## Issue number

ADP-1037

## Summary

This PR tests fundamental properties of the [`CoinSelection.performSelection`](https://github.com/input-output-hk/cardano-wallet/blob/ba3d8d5c8cde7b7aba81e596d4eea425022c49fd/lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs#L122) function.

## Details

In the case where [`performSelection`](https://github.com/input-output-hk/cardano-wallet/blob/ba3d8d5c8cde7b7aba81e596d4eea425022c49fd/lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs#L122) successfully produces a [`Selection`](https://github.com/input-output-hk/cardano-wallet/blob/ba3d8d5c8cde7b7aba81e596d4eea425022c49fd/lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs#L458) `s` for a given set of [`SelectionConstraints`](https://github.com/input-output-hk/cardano-wallet/blob/ba3d8d5c8cde7b7aba81e596d4eea425022c49fd/lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs#L343) `cs` and [`SelectionParams`](https://github.com/input-output-hk/cardano-wallet/blob/ba3d8d5c8cde7b7aba81e596d4eea425022c49fd/lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs#L382) `ps`, we verify that the following properties hold:

* `selectionHasSufficientCollateral cs ps s`
* `selectionHasValidSurplus cs ps s`

## Not included in this PR

This PR doesn't include:
- tests that verify the failure cases of `performSelection`
- tests that verify the minimum collateral amount calculation.

These tests will be included in future PRs.

### DIFF
--- a/lib/test-utils/src/Test/QuickCheck/Extra.hs
+++ b/lib/test-utils/src/Test/QuickCheck/Extra.hs
@@ -27,6 +27,10 @@ module Test.QuickCheck.Extra
     , shrinkInterleaved
     , shrinkMapWith
 
+      -- * Generating and shrinking natural numbers
+    , chooseNatural
+    , shrinkNatural
+
       -- * Counterexamples
     , report
     , verify
@@ -41,15 +45,22 @@ module Test.QuickCheck.Extra
 
 import Prelude
 
+import Data.IntCast
+    ( intCast, intCastMaybe )
 import Data.Map.Strict
     ( Map )
+import Data.Maybe
+    ( mapMaybe )
 import Fmt
     ( indentF, (+|), (|+) )
+import Numeric.Natural
+    ( Natural )
 import Test.QuickCheck
     ( Arbitrary (..)
     , Gen
     , Property
     , Testable
+    , choose
     , counterexample
     , liftArbitrary2
     , liftShrink2
@@ -59,6 +70,7 @@ import Test.QuickCheck
     , shrinkList
     , shrinkMapBy
     , suchThat
+    , suchThatMap
     , (.&&.)
     )
 import Test.QuickCheck.Gen.Unsafe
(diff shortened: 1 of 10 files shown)


lib/test-utils/src/Test/QuickCheck/Extra.hs
```diff
@@ -302,6 +314,22 @@ shrinkInterleaved (a, shrinkA) (b, shrinkB) = interleave
     interleave xs [] = xs
     interleave [] ys = ys
 
+--------------------------------------------------------------------------------
+-- Generating and shrinking natural numbers
+--------------------------------------------------------------------------------
+
+chooseNatural :: (Natural, Natural) -> Gen Natural
+chooseNatural (lo, hi) =
+    choose (intCast lo, intCast hi)
```
jonathanknowles: Perhaps we could use [`chooseInteger`](https://hackage.haskell.org/package/QuickCheck-2.14.2/docs/Test-QuickCheck.html#v:chooseInteger) here, as the QC documentation claims that this is faster.

lib/test-utils/src/Test/QuickCheck/Extra.hs
```diff
@@ -302,6 +314,22 @@ shrinkInterleaved (a, shrinkA) (b, shrinkB) = interleave
...
+    `suchThatMap`
+    intCastMaybe @Integer @Natural
+
+shrinkNatural :: Natural -> [Natural]
+shrinkNatural n
+    = mapMaybe (intCastMaybe @Integer @Natural)
+    $ shrink
```
jonathanknowles: Investigate whether `shrinkIntegral` would be faster / more appropriate here.

lib/test-utils/src/Test/QuickCheck/Extra.hs
jonathanknowles: Fixed in 36fe56c3c4adb531c68c9bde4f7b1d5fa1681438.

lib/test-utils/src/Test/QuickCheck/Extra.hs
jonathanknowles: Fixed in 36fe56c3c4adb531c68c9bde4f7b1d5fa1681438.

sevanspowell (approved):

PR #2954, opened 2021-10-07
Sevanspowell/adp 1042/tx needs collateral

sevanspowell:


lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Balance.hs
```diff
@@ -146,6 +148,8 @@ import Cardano.Wallet.Primitive.Types.UTxOSelection
     ( IsUTxOSelection, UTxOSelection, UTxOSelectionNonEmpty )
 import Control.Monad.Extra
     ( andM )
+import Control.DeepSeq
+    ( NFData )
```
jonathanknowles, line 152: This one looks leftover.

jonathanknowles:
Closing for the moment, since we've switched to the approach started in #2956.

PR #2957, opened 2021-10-07
Add feePadding for use in balanceTransaction

Anviking:
- [x] Replace manually calculated padding with updating the ExUnits to max before calling `evaluateMinimumFee`.

### Issue Number

ADP-1182

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -401,7 +401,6 @@ import Cardano.Wallet.Primitive.SyncProgress
 import Cardano.Wallet.Primitive.Types
     ( Block
     , BlockHeader (..)
-    , FeePolicy (..)
     , NetworkParameters (..)
     , PassphraseScheme (..)
     , PoolId
@@ -452,6 +451,7 @@ import Cardano.Wallet.Transaction
     , TxUpdate (..)
     , Withdrawal (..)
     , defaultTransactionCtx
+    , noTxUpdate
     )
 import Cardano.Wallet.Unsafe
     ( unsafeRunExceptT )
@@ -2250,7 +2250,11 @@ balanceTransaction ctx genChange (ApiT wid) body = do
         -- adjust the change and ExUnits of each redeemer to something more
         -- sensible than the max execution cost.
         let txPlutusScriptExecutionCost = maxScriptExecutionCost tl pp partialTx
-        let txContext = defaultTransactionCtx
+
+        -- NOTE: The `ErrUpdateSealedTx` thrown from here will hide the ones
+        -- thrown from the final `updateTx`. This doesn't matter.
+        txContext <- liftUpdateTxErr $ padFeeEstimation partialTx pp nodePParams $
+            defaultTransactionCtx
                 { txPlutusScriptExecutionCost
                 , txMetadata
                 , txWithdrawal
@@ -2259,7 +2263,7 @@ balanceTransaction ctx genChange (ApiT wid) body = do
                         SelectionCollateralRequired
                     else
                         SelectionCollateralNotRequired
-                } & padFeeEstimation partialTx pp nodePParams
+                }
 
         -- FIXME: The coin selection and reported fees will likely be wrong in
         -- the presence of certificates (and deposits / refunds). An immediate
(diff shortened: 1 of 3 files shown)


lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2226,14 +2235,31 @@ balanceTransaction ctx genChange (ApiT wid) body = do
...
-        let txCtx = defaultTransactionCtx
+
+        let txCtx' = defaultTransactionCtx
                 { txPlutusScriptExecutionCost = executionFee
                 , txMetadata = mdM
                 , txWithdrawal = wdrl
                 , txTimeToLive = ttl
                 }
 
+        let realFee = evaluateMinimumFee tl pp' sealedTxIncoming
```
KtorZ: `realFee` is a bit confusing here, because they're still not the real fee but an estimation -- just a better one. The real fee we can obtain at the very end, by looking at the size of the serialised tx.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2214,9 +2217,15 @@ balanceTransaction
     -> Handler (ApiConstructTransaction n)
 balanceTransaction ctx genChange (ApiT wid) body = do
     pp <- liftIO $ NW.currentProtocolParameters nl
+    pp' <- (error "TODO: balanceTransaction in byron")
+        <$> liftIO (NW.currentNodeProtocolParameters nl)
+
+    -- TODO: Check that sealedTxIncoming contains a PParam hash matching both pp
+    -- and pp'. Without this, collateral could be forfeited!
```
KtorZ: I am not sure to understand this one? Note that I took care of adding the script integrity hash when necessary as part of #2952

Anviking: Ah, thanks, indeed. If we always set the script integrity hash, then it should be nothing to worry about. I thought it looked like in [cardano-api](https://github.com/input-output-hk/cardano-node/blob/0ed04ab8d0a205691219f04623a3d880545a3f34/cardano-api/src/Cardano/Api/TxBody.hs#L1192) (but I now see it's only used to derive the script integrity hash), and that you said PAB would set it (https://github.com/input-output-hk/cardano-wallet/pull/2952#discussion_r723914685)

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2308,17 +2312,20 @@ balanceTransaction ctx genChange (ApiT wid) body = do
...
-            { coinSelection
-            , transaction
-            , fee = coinToQuantity (newFee oldFee)
-            }
+    updatedTx <- liftUpdateTxErr (updateTx tl nodePParams partialTx txUpdate)
+    return $ ApiConstructTransaction
+        { coinSelection
+        , transaction = ApiT updatedTx
+        , fee = coinToQuantity (newFee oldFee)
+        }
```
KtorZ, line 2320: Minor stylistic suggestion:

```suggestion
    let fee = coinToQuantity (newFee oldFee)
    transaction <- ApiT <$> liftUpdateTxErr (updateTx tl nodePParams partialTx txUpdate)
    return $ ApiConstructTransaction { coinSelection, transaction, fee }
```

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2352,25 +2359,29 @@ balanceTransaction ctx genChange (ApiT wid) body = do
...
-            walletTx = decodeTx tl sealedTx
-            worseEstimate = calcMinimumCost tl pp txCtx skeleton
-            skeleton = SelectionSkeleton
+        -> Either ErrUpdateSealedTx TransactionCtx
+    padFeeEstimation sealedTx' pp pp' txCtx = do
+        -- Ensure `evaluateMinimumFee` always will overestimate even though the
+        -- execution units are not known yet. Calling `updateTx` like this will:
+        -- 1. make the CBOR encoding of the execution units as large as possible
+        -- 2. add the script integrity hash
+        let maxUnits = pp ^. (#txParameters . #getMaxExecutionUnits)
```
KtorZ, line 2368: Note that the max units are already part of the context as `txPlutusScriptExecutionCost`, so it'd be better to pull from there and avoid duplicating this.

Anviking: `txPlutusScriptExecutionCost` has type `Coin`, not `ExecutionUnits`, however…

Anviking: I was going to say:
> I think there's some oddness to how `padFeeEstimation`, `selectAssets`, the rest of `balanceTransaction`, `TransactionCtx` all work together, but I think it's "fine"… for now / not something to be tackled by this PR.

But I suppose we could set the `newExUnits` already in `balanceTransaction` and keep the result…  not to have to do it twice… 🤔

KtorZ: Actually, newExUnits is gone :s ... We'll have to discuss that.

Anviking:
Closing as it's out of date, will open a new version eventually as part of ADP-1214

PR #2959, opened 2021-10-07
Add submit transaction endpoint

paweljakubas:
adp-1171

- [x] Adding submit transaction endpoint
- [x] Add mkTxMetaWithoutSel Cardano.Wallet
- [x] Impl submitTransaction in Cardano.Api.Server
- [x] understand/get `deposit` and `refunds` from `SealedTx`
- [x] Add integration tests showing Pending -> InLedger trnasformation

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
@@ -382,7 +382,7 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
 
     it "TRANS_NEW_CREATE_04a - Single Output Transaction" $ \ctx -> runResourceT $ do
 
-        liftIO $ pendingWith "Missing outputs on response - to be fixed in ADP-985"
+        --liftIO $ pendingWith "Missing outputs on response - to be fixed in ADP-985"
 
         let initialAmt = 3 * minUTxOValue (_mainEra ctx)
         wa <- fixtureWalletWith @n ctx [initialAmt]
@@ -403,6 +403,7 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
             , expectField (#coinSelection . #change) (`shouldSatisfy` (not . null))
             , expectField (#fee . #getQuantity) (`shouldBe` feeMin)
             ]
+        let expectedFee = getFromResponse (#fee . #getQuantity) rTx
 
         let filterInitialAmt =
                 filter (\(ApiCoinSelectionInput _ _ _ _ amt' _) -> amt' == Quantity initialAmt)
@@ -410,9 +411,31 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
                 getFromResponse (#coinSelection . #inputs) rTx
         length coinSelInputs `shouldBe` 1
 
-        -- TODO: now we should sign it and send it in two steps
-        --       make sure it is delivered
-        --       make sure balance is updated accordingly on src and dst wallets
+        apiTx <- unsafeGetTx rTx
+
+        signedTx <- signTx ctx wa apiTx
+
+        void $ submitTx ctx signedTx [ expectResponseCode HTTP.status202 ]
+
+        eventually "Source wallet balance is decreased by amt + fee" $ do
+            rWa <- request @ApiWallet ctx
+                (Link.getWallet @'Shelley wa) Default Empty
+            verify rWa
+                [ expectSuccess
+                , expectField
+                        (#balance . #available . #getQuantity)
+                        (`shouldBe` initialAmt - (amt + fromIntegral expectedFee))
+                ]
+
+        eventually "Target wallet balance is amt" $ do
+            rWr <- request @ApiWallet ctx
+                (Link.getWallet @'Shelley wb) Default Empty
+            verify rWr
+                [ expectSuccess
+                , expectField
+                        (#balance . #available . #getQuantity)
+                        (`shouldBe` amt)
+                ]
 
     it "TRANS_NEW_CREATE_04b - Cannot spend less than minUTxOValue" $ \ctx -> runResourceT $ do
         wa <- fixtureWallet ctx
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
@@ -32,8 +32,7 @@ import Cardano.Wallet.Api.Types
     , ApiCoinSelectionInput (..)
     , ApiConstructTransaction (..)
     , ApiFee (..)
-    , ApiSerialisedTransaction
-    , ApiSignedTransaction (..)
+    , ApiSerialisedTransaction (..)
     , ApiStakePool
     , ApiT (..)
     , ApiTransaction
@@ -1034,8 +1033,7 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
                                , "passphrase": #{fixturePassphrase}
                                }|]
         let signEndpoint = Link.signTransaction @'Shelley wa
-        signedTx <- getFromResponse #transaction <$>
-            request @ApiSerialisedTransaction ctx signEndpoint Default toSign
+        (_, Right signedTx) <- request @ApiSerialisedTransaction ctx signEndpoint Default toSign
 
         -- Submit tx
         txId <- submitTx ctx signedTx [ expectResponseCode HTTP.status202 ]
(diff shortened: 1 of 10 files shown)


lib/core/src/Cardano/Wallet.hs
```diff
@@ -1657,6 +1659,66 @@ getTxExpiry ti maybeTTL = do
...
+
+        amtInps
+            = sumCoins (map snd colls ++ map snd inps ) -- here we need to remove overlapping txins between collateral and normal inputs
+            & case txWithdrawal txCtx of
+                w@WithdrawalSelf{} -> addCoin (withdrawalToCoin w)
+                WithdrawalExternal{} -> Prelude.id
+                NoWithdrawal -> Prelude.id
+    in return TxMeta
+       { status = Pending
+       , direction = if amtInps > amtOuts then Outgoing else Incoming
```
KtorZ: I'd argue that we should only add to our pending set transactions that are _outgoing_. So this check is potentially refundant.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1657,6 +1659,66 @@ getTxExpiry ti maybeTTL = do
...
+    -> [(TxIn, Coin)]
+    -> [(TxIn, Coin)]
+    -> [TxOut]
+    -> IO TxMeta
+mkTxMetaWithoutSel blockHeader wState txCtx colls inps outs =
+    let
+        amtOuts = sumCoins $ mapMaybe (flip ourCoin wState) outs
+
+        amtInps
+            = sumCoins (map snd colls ++ map snd inps ) -- here we need to remove overlapping txins between collateral and normal inputs
```
KtorZ: Collaterals are not inputs! They are only spent in case where the transaction is failing, they don't count towards the input side of the balance. However, certificate refunds do.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2306,6 +2308,46 @@ balanceTransaction ctx genChange (ApiT wid) body = do
...
+submitTransaction ctx (ApiT wid) (ApiT sealedTx) = do
+    ttl <- liftIO $ W.getTxExpiry ti Nothing
+    let (Tx txId _ colls inps outs wdrlMap _ _) = tx
+
+    _ <- withWorkerCtx ctx wid liftE liftE $ \wrk -> do
+        (acct, _, _) <- liftHandler $ W.readRewardAccount @_ @s @k @n wrk wid
+        (wdrl, _) <- mkRewardAccountBuilder @_ @s @_ @n ctx wid $
+            if Map.member acct wdrlMap
+            then Just SelfWithdrawal
+            else Nothing
```
KtorZ, line 2445: I really feel uneasy about these conversions... Now that transactions are constructed outside the boundaries of the wallet, there are many invariants / pre-conditions the wallet core functions were holding on that are no longer valid. For example, we use to calculate the _amount_ by diffing inputs+refunds+withdrawals - outputs-deposit, but because we knew that transactions could only ever have one withdrawal or one deposit. With this new API, this is no longer true and may lead to situation where the wallet reports an amount that is completely off. 

The `TransactionContext` and internals of the wallet need to be reworked to work with multi withdrawals and multi deposits and... etc, before we can support such an API I believe.

lib/core/src/Cardano/Wallet.hs
paweljakubas: indeed, I set it to Outgoing

lib/core/src/Cardano/Wallet.hs
paweljakubas: removed collaterals. thanks for the remark, I thought their lifecycle is different. How to get certificate refunds though? I need it to add to https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs#L199 and get when `decodeTx` from https://github.com/input-output-hk/cardano-node/blob/master/cardano-api/src/Cardano/Api/TxBody.hs#L1151 located here https://github.com/input-output-hk/cardano-node/blob/master/cardano-api/src/Cardano/Api/TxBody.hs#L1212 ?

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: tx could include withdrawals also belonging to different wallets, but within a given wallet we can have one, right? I agree that this kind of calculations are in general incorrect for fee if we base on tx size. But maybe, here, in the context of a given wallet they have a chance to be correct?

lib/core/src/Cardano/Wallet.hs
paweljakubas: the same with deposits probably - will need to add them to `Tx`

paweljakubas: I researched the topic. So we take deposit taken and returned in coin selection. Within `selectAssets`, we set the following fields within the `SelectionParams` record:
```
certificateDepositsTaken    :: Natural
certificateDepositsReturned :: Natural 
```
And the following field within `SelectionConstraints`
```
certificateDepositAmount :: Coin
```
Within the `CoinSelection` module, these are converted internally to `Balance.SelectionParams.{extraCoin{Source,Sink}`, by multiplying each count with `certificateDepositAmount`.

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: According to https://github.com/input-output-hk/cardano-wallet/pull/2959#discussion_r724909999
I would argue we have here not:
```
inputs+refunds+withdrawals - outputs-deposit
```
but rather 
```
inputs+withdrawals - outputs
```
as deposit paying/returning should be already accounted in inputs/outputs

specifications/api/swagger.yaml
```diff
@@ -5733,6 +5733,23 @@ paths:
...
+      summary: Submit
+      description: |
+        <p align="right">status: <strong>under development</strong></p>
+        Submit a transaction that was already created and signed.
+      parameters:
+        - *parametersWalletId
+      requestBody:
+        required: true
+        content:
+          application/octet-stream:
```
piotr-iohk: If we're going for new endpoint I'd be for considering using `application/json` and accepting CBOR in base64 encoding, which would be consistent with the workflow, i.e.:
 - balance/construct -> returns `transaction` in base64
 - sign (takes `transaction` in base64) -> returns `transaction` in base64
 - submitTransaction (takes `transaction` base64 instead of binary)

paweljakubas: good idea. adopted your proposal

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -390,16 +390,109 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
...
+            rWr <- request @ApiWallet ctx
+                (Link.getWallet @'Shelley wb) Default Empty
+            verify rWr
+                [ expectSuccess
+                , expectField
+                        (#balance . #available . #getQuantity)
+                        (`shouldBe` amt)
+                ]
+
+    it "TRANS_NEW_CREATE_04a - Single Output Transaction with submitWithWid" $ \ctx -> runResourceT $ do
```
piotr-iohk, line 441: I think we should use submitWithWid in all tests and have one or two with only submitTx, since it is the new, intended workflow. Then we'll have better coverage.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -1149,7 +1242,7 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
             request @(ApiConstructTransaction n) ctx constructEndpoint Default payload
 
         -- Submit tx
-        void $ submitTx ctx sealedTx [ expectResponseCode HTTP.status500 ]
+        void $ submitTx ctx (ApiSignedTransaction sealedTx) [ expectResponseCode HTTP.status500 ]
```
piotr-iohk: It would be nice to fix this on new submit endpoint. (There's issue for it ADP-1172, linked to ADP-1171)

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2306,6 +2308,46 @@ balanceTransaction ctx genChange (ApiT wid) body = do
...
+        , HardDerivation k
+        , Typeable s
+        , Typeable n
+        , Bounded (Index (AddressIndexDerivationType k) 'AddressK)
+        )
+    => ctx
+    -> ApiT WalletId
+    -> ApiSignedTransaction
+    -> Handler ApiTxId
+submitTransaction ctx (ApiT wid) (ApiSignedTransaction (ApiT sealedTx)) = do
```
piotr-iohk: Is there "re-submission" mechanism on new submitTransaction?

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
paweljakubas: I will do it in separate PR solving exact this problem ADP-1172. After this PR lands into master

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: hmm, I am not sure it is good idea to take responsibility here and deal with ledger limitations (I bet this is consequence when mempool is full and cannot accept more txs). Maybe @KtorZ  has strong opinion on that, though.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
paweljakubas: If there is decision we go with this endpoint full scale, I am happy to propagate using `submitWithWid` everywhere in TransactionNew module

paweljakubas:
closing in favor of https://github.com/input-output-hk/cardano-wallet/pull/3047

PR #2962, opened 2021-10-11
Add coverage for `toBalanceConstraintsParams`.

jonathanknowles:
## Issue Number

ADP-1037

## Summary

This PR:

- [x] replaces an erroneous call to `subtract` with a dedicated function `reduceSelectionLimitBy`.
- [x] provides property tests to verify that `reduceSelectionLimitBy` has the correct behaviour for all selection limits.
- [x] provides property tests to verify that `toBalanceConstraintsParams` performs the correct adjustment to the `computeMinimumCost` function. 
- [x] provides property tests to verify that `toBalanceConstraintsParams` performs the correct adjustment to the `computeSelectionLimit` function. 

## Background

The `subtract` function (from `Prelude`) is equivalent to `flip (-)`, which means that `subtract a b` is equivalent to `b - a`, rather than the other way round.

Thanks to @KtorZ for discovering this.

### DIFF
--- a/lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/BalanceSpec.hs
+++ b/lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/BalanceSpec.hs
@@ -79,6 +79,7 @@ import Cardano.Wallet.Primitive.CoinSelection.Balance
     , mapMaybe
     , performSelection
     , performSelectionEmpty
+    , reduceSelectionLimitBy
     , reduceTokenQuantities
     , removeBurnValueFromChangeMaps
     , removeBurnValuesFromChangeMaps
@@ -192,6 +193,7 @@ import Test.QuickCheck
     , Blind (..)
     , Fun
     , Gen
+    , Negative (..)
     , Positive (..)
     , Property
     , applyFun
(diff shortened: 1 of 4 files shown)


lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelection/BalanceSpec.hs
```diff
@@ -3678,6 +3691,60 @@ prop_runRoundRobin_generationOrder initialState = property $
...
+    :: SelectionLimit -> Property
+prop_reduceSelectionLimitBy_reductionZero limit =
+    prop_reduceSelectionLimitBy_coverage_limit limit .&&.
+    limit `reduceSelectionLimitBy` 0 == limit
+
+prop_reduceSelectionLimitBy_reductionPositive
+    :: SelectionLimit -> Positive Int -> Property
+prop_reduceSelectionLimitBy_reductionPositive limit (Positive reduction) =
+    prop_reduceSelectionLimitBy_coverage_limit limit .&&.
+    limit == fmap (+ reduction) (limit `reduceSelectionLimitBy` reduction)
```
Anviking, line 3746: 👍

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelectionSpec.hs
```diff
@@ -254,6 +270,120 @@ prop_performSelection_onSuccess_hasSuitableCollateral cs _ps selection =
...
+    computeMinimumCostOriginal :: SelectionSkeleton -> Coin
+    computeMinimumCostOriginal = constraints ^. #computeMinimumCost
+
+    computeMinimumCostAdjusted :: SelectionSkeleton -> Coin
+    computeMinimumCostAdjusted =
+        toBalanceConstraintsParams (constraints, params)
+            & fst & view #computeMinimumCost
+
+    costOriginal :: Coin
+    costOriginal = computeMinimumCostOriginal skeleton
```
Anviking, line 329: Looking at this again it occurred to me: isn't it strange that the outer constraints contain the unadjusted and wrong minimum cost? If someone were to call the outer `computeMinimumCost` they would get the wrong results?

Anviking (approved):
LGTM! Thanks for the walkthrough call. Left one comment / thought.

lib/core/test/unit/Cardano/Wallet/Primitive/CoinSelectionSpec.hs
jonathanknowles: > If someone were to call the outer `computeMinimumCost` they would get the wrong results?

The only way a caller could call this function is if they prepared a `SelectionSkeleton` externally.

If the caller chose to do this, then the onus would be on the caller to provide a correct `SelectionSkeleton` that accurately represented the transaction for which they wanted to compute a minimum cost. But assuming they did prepare an accurate `SelectionSkeleton`, then this function would still give them the correct result, since the contract required of `computeMinimumCost` is that:

```computeMinimumCost skeleton >= actualCostOfSerializedTx skeleton```

PR #2964, opened 2021-10-11
Remove 'coin_selection' field from the balance API's response

KtorZ:
- :round_pushpin: **Remove 'coin_selection' field from the Balance API's response**
    (a) the implementation of this field was wrong, and getting it right is quite involved as it requires revising several parts of the core wallet. It is perhaps not desirable / out-of-scope for this endpoint actually.

  (b) the coin_selection result is only truly useful when constructing a transaction which is under control of the wallet. Balancing may be done on any arbitrary foreign transaction and as such, breaks several invariants of the wallet (only one withdrawal, only one delegation certificate at a time etc..).

### Comments

Based off [`KtorZ/ADP-1183/balancing-execution-units`](https://github.com/input-output-hk/cardano-wallet/pull/2952)

### Issue Number

ADP-1183

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -95,7 +95,7 @@ module Cardano.Wallet.Api.Types
     , ApiSignTransactionPostData (..)
     , PostTransactionOldData (..)
     , PostTransactionFeeOldData (..)
-    , ApiSignedTransaction (..)
+    , ApiSerialisedTransaction (..)
     , ApiTransaction (..)
     , ApiMintedBurnedTransaction (..)
     , ApiMintedBurnedInfo (..)
@@ -996,7 +996,7 @@ data PostTransactionFeeOldData (n :: NetworkDiscriminant) = PostTransactionFeeOl
 
 type ApiBase64 = ApiBytesT 'Base64 ByteString
 
-newtype ApiSignedTransaction = ApiSignedTransaction
+newtype ApiSerialisedTransaction = ApiSerialisedTransaction
     { transaction :: ApiT SealedTx
     } deriving stock (Eq, Generic, Show)
       deriving anyclass (NFData)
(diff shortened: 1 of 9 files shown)


lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -996,7 +996,7 @@ data PostTransactionFeeOldData (n :: NetworkDiscriminant) = PostTransactionFeeOl
 
 type ApiBase64 = ApiBytesT 'Base64 ByteString
 
-newtype ApiSignedTransaction = ApiSignedTransaction
+newtype ApiSerializedTransaction = ApiSerializedTransaction
```
KtorZ: Renamed since it's no longer only used for the sign endpoint, but also as a result of the balance endpoint.

paweljakubas: maybe it is worth to align this with https://github.com/input-output-hk/cardano-wallet/blob/master/specifications/api/swagger.yaml#L2217

specifications/api/swagger.yaml
```diff
@@ -2285,15 +2285,18 @@ components:
       items: *serialisedTransactionBase64
       title: Serialized transaction witnesses
 
-    ApiSignedTransaction: &ApiSignedTransaction
-      description: |
-        The result of signing a transaction (serialized and encoded).
+    ApiSerializedTransaction: &ApiSerializedTransaction
```
paweljakubas: we have also this https://github.com/input-output-hk/cardano-wallet/blob/master/specifications/api/swagger.yaml#L2217
One letter difference 's' -> 'z'. Would be good to reconcile

paweljakubas: ah, they are the same. Both use serialisedTransactionBase64

KtorZ: Arg! Indeed. UK vs US. Seems like the API uses the UK's style more, so I'll go with that.

lib/core/src/Cardano/Wallet/Api/Types.hs
```diff
@@ -996,7 +996,7 @@ data PostTransactionFeeOldData (n :: NetworkDiscriminant) = PostTransactionFeeOl
 
 type ApiBase64 = ApiBytesT 'Base64 ByteString
 
-newtype ApiSignedTransaction = ApiSignedTransaction
+newtype ApiSerialisedTransaction = ApiSerialisedTransaction
```
paweljakubas, line 999: so the idea is to use in response of balanceTx,  and in response of signTx `ApiSerialisedTransaction`? and they are going to be base64 encoded bytestring (signTx will also accept hex coming from constructTx), the same as in case of request of submitTx. Right?

KtorZ: Correct :+1:

KtorZ: Note that this doesn't change anything compared to what was already there w.r.t to the serialised transaction returned by the API, it however removes the `coin_selection` field from the balance's result. For the rest, it's identical.

paweljakubas (approved):

PR #2967, opened 2021-10-11
ADP-1183 (Part 2): Balancing Plutus transactions 

KtorZ:
- :round_pushpin: **Add core type 'Redeemer' and compat functions with Cardano ledger.**
    Certificates have been left out currently, to be done in a later commit / iteration.

- :round_pushpin: **Add util 'modifyM' function for state-monad.**
  
- :round_pushpin: **Expose constructor for TimeInterpreter**
    Will be needed in order to evaluate Plutus scripts and assign execution units to redeemers.
  I originally wanted to expose only record accessors, but this isn't working for the 'interpreter' field because of the existential; eras would escape their scope. Hence the full constructor being exposed. Alternatively, we could expose the 'EpochInfo' from this module instead to keep the module's API somewhat sane.

- :round_pushpin: **Extend 'TransactionLayer' with 'assignScriptRedeemers'**
    The goal is to move and replace the bunch of logic recently added into 'updateTx' to keep the latter only about adding inputs/outputs, and the former about handling redeemers. This makes it easier to follow and test down the line.

- :round_pushpin: **Implement 'assignScriptRedeemers' for Cardano.**
  
- :round_pushpin: **Remove newExUnits from TxUpdate, moved as separate function.**
    This was a quick-n-dirty hack to get started with the integration and testing. Adding witnesses is actually quite more involved in the end and it is therefore better done in a, far from trivial, separate function.

- :round_pushpin: **Remove 'certifying' type of redeemers in the API**
    I've mistakenly messed up when introducing this one originally an the type is plain wrong (it's Byron's certificate where it should be Shelley's). For now, the focus is on getting the balancing endpoint to work, and the case of redeemers for certificates is quite an edge-case, so it can be addressed in subsequent PRs.

- :round_pushpin: **Wire updateTx and assignScriptRedeemers errors into the API.**
  
- :round_pushpin: **Move 'mapFirst' local function to Cardano.Wallet.Util**
    For re-use across other modules.

- :round_pushpin: **Add txOutAddCoin function to primitives for transaction.**
  
- :round_pushpin: **Finalize implementation of 'balanceTransaction'**
    This should now work properly for any redeemers and, with proper execution units & fees. Not really heavily tested at this stage, but it's coming.

- :round_pushpin: **Make fee padding more robust to number (and content) of redeemers.**
  
- :round_pushpin: **Also provide datum with external inputs.**
    It is actually needed to lookup redeemer pointers of spending redeemers.

- :round_pushpin: **Resolve additional issues with execution units and clean up code.**
  
- :round_pushpin: **Improve error reporting for redeemers error in the balancing endpoint.**
  
- :round_pushpin: **Add new plutus scenario: GameStateMachine.**
    Currently fails, because this scenario requires the wallet to handle minting/burning accordingly.

- :round_pushpin: **Support balancing minted and burnt tokens in pre-constructed transactions.**
  
- :round_pushpin: **Enable 3rd step of the GameStateMachine contract. Currently fails.**
  
- :round_pushpin: **Also provide input resolution for own inputs when evaluating scripts.**
    Doh. This was a tough one. Turns out that, the 'UTXO' given to 'evaluateTransactionExecutionUnits' MUST contain ALL inputs referenced in the transaction. If some are missing, it silently ignore them when constructing the script context for the underlying validator causing the script to either fail, or evaluate with a wrong context. I raised the issue with the ledger team as I firmly believe the 'evaluateTransactionExecutionUnits' ought to fail loudly when the UTXO given to it is incomplete (easily verifiable).

- :round_pushpin: **Move epochInfo creation to TimeInterpreter module**
    To create the 'EpochInfo' needed to evaluate scripts, I had to expose the constructor of `TimeInterpreter`, making the abstraction leaky. That's not good, especially when it is quite easy to keep the abstraction boundary closed by providing ways of doing the conversions from the module itself.

- :round_pushpin: **Move back extra outputs at the _end_**
    For rapid testing, I originally change the way extra outputs (i.e. the change output) were added to the transaction to match the same convention used by the PAB. That is, the PAB have them by default as the first output of every transaction, and contract traces are built with this assumption in mind.
  However, having them at the beginning is really not handy, since there may be in practice more than one change output! It is thus easier to append them to existing outputs, and have contracts works from a 0-based index instead.

- :round_pushpin: **Remove script evaluation scaling**
    I believe the issue was due to the missing internal inputs not present in the UTXO given to the evaluator. So, fixed in 1530e19f4

- :round_pushpin: **Add a 3rd Plutus end-to-end scenario**
    This 3rd scenarios exercises two components of the transaction workflow: minting and burning of tokens (automatically detected from the transaction context), as well as, required signatures from a transaction (field 14), that are necessary here to execute the minting / burning policy. The policy itself is a template which requires a verification key hash and validates if the minting/burning transaction is signed by the sk corresponding to that key hash. When executing the scenario, the vk is pulled from the wallet known keys, and added as required signers to the transaction so that the wallet produces a signature for that key regardless of whether it is needed for an input.

- :round_pushpin: **Add some extra notes after self-review.**
  
- :round_pushpin: **Add a new function to integration tests' context to inject rewards into a script address.**
  
- :round_pushpin: **Use 'StakeAddress' instead of 'RewardAccount' for withdrawal redeemers**
    RewardAccount are associated with stake keys in all boundary interfaces; that is, they are serialised and parsed as type 08 addresses, whereas they could in principle be type 09 (and this is precisely what withdrawal e2e plutus scenario is testing. The wallet codebase does not have a good abstraction for representing stake addresses in general, but cardano-api does, and it's quite handy since it's merely a wrapper around the cardano-ledger-specs' types, so easily re-used in 'assignNullRedeemers'. It also has existing conversions from and to bech32, so it is actually a low hanging fruit.

- :round_pushpin: **Add 4th e2e Plutus scenario for trying out withdrawal from script address.**
  
- :round_pushpin: **Regenerate nix**
  
### Comments

N/A

### Issue Number

ADP-1183


lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -871,6 +855,165 @@ _maxScriptExecutionCost pp tx = case view #executionUnitPrices pp of
...
+        -> ExceptT ErrAssignRedeemers m (Map Alonzo.RdmrPtr (Either ErrAssignRedeemers Alonzo.ExUnits))
+    evaluateExecutionUnits alonzoTx = withExceptT ErrAssignRedeemersPastHorizon $ do
+        let utxo = utxoFromAlonzoTx alonzoTx
+        let costs = toCostModelsAsArray (Alonzo._costmdls pparams)
+
+        (systemStart, epochInfo) <- case ti of
+            TimeInterpreter{interpreter,blockchainStartTime} -> do
+                let systemStart = toSystemStart blockchainStartTime
+                epochInfo <- lift (interpreterToEpochInfo <$> interpreter)
+                pure (systemStart, epochInfo)
```
KtorZ: NOTE: This could be moved to the Slotting module, to avoid leaking `TimeInterpreter` abstraction's internals.

KtorZ:
bors try

KtorZ:
bors try

KtorZ:
bors try

KtorZ:
bors try

PR #2970, opened 2021-10-12
Factor out pure balanceTransaction logic

Anviking:
### Motivation

Make property testing of the `balanceTransaction`  logic possible.

### Overview

- [x] Add `m` to `HasLogger` constraint rather than to assume IO
- [x] Make selectAssets run in `MonadRandom m` rather than IO
- [x] Let selectAssets take PParams directly, rather than fetching from NetworkLayer
- [x] Remove IO from assignScriptRedeemers
- [x] Factor out W.balanceTransaction running in `MonadRandom m`

### Issue Number

ADP-1182

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -1404,7 +1404,7 @@ balanceTransaction
     let candidateMinFee = fromMaybe (Coin 0) $
             evaluateMinimumFee tl nodePParams candidateTx
 
-    let surplus = delta `Coin.distance` candidateMinFee
+    let surplus = delta `Coin.difference` candidateMinFee
     assembleTransaction $ TxUpdate
         { extraInputs
         , extraCollateral
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -1424,9 +1424,11 @@ balanceTransaction
       where
         resolveInput :: TxIn -> Maybe (TxOut, Maybe (Hash "Datum"))
         resolveInput i =
-            (\(_,o,d) -> (o,d)) <$> L.find (\(i',_,_) -> i == i') externalInputs
+            (\(_,o,d) -> (o,d))
+                <$> L.find (\(i',_,_) -> i == i') externalInputs
             <|>
-            (\(_,o) -> (o, Nothing)) <$> L.find (\(i',_) -> i == i') (extraInputs update)
+            (\(_,o) -> (o, Nothing))
+                <$> L.find (\(i',_) -> i == i') (extraInputs update)
 
     extractFromTx tx =
         let (Tx _id _fee _coll _inps outs wdrlMap meta _vldt, toMint, toBurn)
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -1365,7 +1365,7 @@ balanceTransaction
                     , inputs
                     , fst <$> (sel' ^. #collateral)
                     , sel' ^. #change
-                )
+                    )
         withExceptT ErrBalanceTxSelectAssets $
             selectAssets @_ @m @s @k ctx pp SelectAssetsParams
                 { outputs
@@ -1381,18 +1381,18 @@ balanceTransaction
     -- Once the coin-selection is done, we need to
     --
     -- (a) Add selected inputs, collateral and change outputs to the transaction
-    -- (b) Assign correct execution units to every redeemers
+    -- (b) Assign correct execution units to every redeemer
     -- (c) Correctly reference redeemed entities with redeemer pointers
     -- (d) Adjust fees and change output(s) to the new fees.
     --
     -- There's a strong assumption that modifying the fee value AND increasing
-    -- the coin value of change outputs does not modify transaction fees; or
+    -- the coin values of change outputs does not modify transaction fees; or
     -- more exactly, does not modify the execution units of scripts. This is in
-    -- principle a fair assumption because scripts validators ought to be
-    -- unaware of change outputs. If their execution cost increases when change
-    -- output increases, then it becomes impossible to guarantee that the fee
+    -- principle a fair assumption because script validators ought to be
+    -- unaware of change outputs. If their execution costs increase when change
+    -- output values increase, then it becomes impossible to guarantee that fee
     -- balancing will ever converge towards a fixed point. A script validator
-    -- doing such thing is considered bonkers and this is not a behavior we
+    -- doing such a thing is considered bonkers and this is not a behavior we
     -- ought to support.
 
     candidateTx <- assembleTransaction $ TxUpdate
@@ -1429,24 +1429,29 @@ balanceTransaction
             (\(_,o) -> (o, Nothing)) <$> L.find (\(i',_) -> i == i') (extraInputs update)
 
     extractFromTx tx =
-        let (Tx _id _fee _coll _inps outs wdrlMap meta _vldt, toMint, toBurn) = decodeTx tl tx
+        let (Tx _id _fee _coll _inps outs wdrlMap meta _vldt, toMint, toBurn)
+                = decodeTx tl tx
             -- TODO: Find a better abstraction that can cover this case.
             wdrl = WithdrawalSelf
-                (error $ "WithdrawalSelf: reward-account should never been use "
-                      <> "when balancing transactions but it was!"
+                (error $ unwords
+                    [ "WithdrawalSelf: reward account should never have been used"
+                    , "when balancing a transaction, but it was!"
+                    ]
                 )
-                (error $ "WithdrawalSelf: derivation path should never been use "
-                      <> "when balancing transactions but it was!"
+                (error $ unwords
+                    [ "WithdrawalSelf: derivation path should never have been used"
+                    , "when balancing a transaction, but it was!"
+                    ]
                 )
                 (sumCoins wdrlMap)
          in (outs, wdrl, meta, toMint, toBurn)
 
     -- | Wallet coin selection is unaware of many kinds of transaction content
     -- (e.g. datums, redeemers), which could be included in the input to
-    -- `balanceTransaction`. As a workaround we add some padding using
-    -- `evaluateMinimumFee`.
+    -- 'balanceTransaction'. As a workaround we add some padding using
+    -- 'evaluateMinimumFee'.
     --
-    -- TODO: This logic needs to be consistent with how we call `selectAssets`,
+    -- TODO: This logic needs to be consistent with how we call 'selectAssets',
     -- so it would be good to join them into some single helper.
     padFeeEstimation
         :: SealedTx
@@ -1466,8 +1471,10 @@ balanceTransaction
             -- redeemers ex units increased from 0 to their actual values.
             extraMargin = Coin $ ceiling $ (*) b $ fromIntegral
                 $ sizeOfScriptIntegrityHash
-                + sum ((+sizeOfRedeemerCommon) . BS.length . redeemerData <$> redeemers)
+                + sum (map sizeOfRedeemer redeemers)
               where
+                sizeOfRedeemer
+                    = (+ sizeOfRedeemerCommon) . BS.length . redeemerData
                 sizeOfScriptIntegrityHash = 35
                 sizeOfRedeemerCommon = 17
 
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -624,7 +624,7 @@ type HasDBLayer m s k = HasType (DBLayer m s k)
 
 type HasGenesisData = HasType (Block, NetworkParameters, SyncTolerance)
 
-type HasLogger msg = HasType (Tracer IO msg)
+type HasLogger m msg = HasType (Tracer m msg)
 
 -- | This module is only interested in one block-, and tx-type. This constraint
 -- hides that choice, for some ease of use.
@@ -645,10 +645,10 @@ genesisData =
     typed @(Block, NetworkParameters, SyncTolerance)
 
 logger
-    :: forall msg ctx. HasLogger msg ctx
-    => Lens' ctx (Tracer IO msg)
+    :: forall m msg ctx. HasLogger m msg ctx
+    => Lens' ctx (Tracer m msg)
 logger =
-    typed @(Tracer IO msg)
+    typed @(Tracer m msg)
 
 networkLayer
     :: forall m ctx. (HasNetworkLayer m ctx)
@@ -868,7 +868,7 @@ restoreWallet
     :: forall ctx s k.
         ( HasNetworkLayer IO ctx
         , HasDBLayer IO s k ctx
-        , HasLogger WalletWorkerLog ctx
+        , HasLogger IO WalletWorkerLog ctx
         , IsOurs s Address
         , IsOurs s RewardAccount
         )
@@ -884,7 +884,7 @@ restoreWallet ctx wid = db & \DBLayer{..} -> do
   where
     db = ctx ^. dbLayer @IO @s @k
     nw = ctx ^. networkLayer
-    tr = contramap MsgFollow (ctx ^. logger @WalletWorkerLog)
+    tr = contramap MsgFollow (ctx ^. logger @_ @WalletWorkerLog)
 
     run :: ExceptT ErrNoSuchWallet IO () -> IO (FollowAction ErrNoSuchWallet)
     run = fmap (either ExitWith (const Continue)) . runExceptT
@@ -1106,7 +1106,7 @@ queryRewardBalance ctx acct = do
 
 manageRewardBalance
     :: forall ctx s k (n :: NetworkDiscriminant).
-        ( HasLogger WalletWorkerLog ctx
+        ( HasLogger IO WalletWorkerLog ctx
         , HasNetworkLayer IO ctx
         , HasDBLayer IO s k ctx
         , Typeable s
@@ -1143,7 +1143,7 @@ manageRewardBalance _ ctx wid = db & \DBLayer{..} -> do
   where
     db = ctx ^. dbLayer @IO @s @k
     NetworkLayer{watchNodeTip} = ctx ^. networkLayer
-    tr = contramap MsgWallet $ ctx ^. logger @WalletWorkerLog
+    tr = contramap MsgWallet $ ctx ^. logger @_ @WalletWorkerLog
 
 {-------------------------------------------------------------------------------
                                     Address
@@ -1433,8 +1433,8 @@ data SelectAssetsParams s result = SelectAssetsParams
 selectAssets
     :: forall ctx s k result.
         ( HasTransactionLayer k ctx
-        , HasLogger WalletWorkerLog ctx
         , HasNetworkLayer IO ctx
+        , HasLogger IO WalletWorkerLog ctx
         )
     => ctx
     -> SelectAssetsParams s result
@@ -1718,7 +1718,7 @@ submitTx
     :: forall ctx s k.
         ( HasNetworkLayer IO ctx
         , HasDBLayer IO s k ctx
-        , HasLogger WalletWorkerLog ctx
+        , HasLogger IO WalletWorkerLog ctx
         )
     => ctx
     -> WalletId
@@ -1751,7 +1751,7 @@ submitExternalTx
     :: forall ctx k.
         ( HasNetworkLayer IO ctx
         , HasTransactionLayer k ctx
-        , HasLogger TxSubmitLog ctx
+        , HasLogger IO TxSubmitLog ctx
         )
     => ctx
     -> SealedTx
@@ -1824,7 +1824,7 @@ runLocalTxSubmissionPool
     :: forall ctx s k m.
         ( MonadUnliftIO m
         , MonadMonotonicTime m
-        , HasLogger WalletWorkerLog ctx
+        , HasLogger IO WalletWorkerLog ctx
         , HasNetworkLayer m ctx
         , HasDBLayer m s k ctx
         )
@@ -1857,7 +1857,7 @@ runLocalTxSubmissionPool cfg ctx wid = db & \DBLayer{..} -> do
     rateLimited = throttle (rateLimit cfg) . const
 
     tr = unliftIOTracer $ contramap (MsgWallet . MsgTxSubmit) $
-        ctx ^. logger @WalletWorkerLog
+        ctx ^. logger @_ @WalletWorkerLog
     trBracket = contramap MsgProcessPendingPool tr
     trRetry i = contramap (MsgRetryPostTx i) tr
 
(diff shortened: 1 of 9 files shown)


lib/core/src/Cardano/Wallet.hs
```diff
@@ -1264,6 +1272,139 @@ normalizeDelegationAddress s addr = do
...
+            } & padFeeEstimation partialTx pp nodePParams
+
+        -- FIXME: The coin selection and reported fees will likely be wrong in
+        -- the presence of certificates (and deposits / refunds). An immediate
+        -- "fix" is to return a proper error from the handler when any key or
+        -- pool registration (resp. deregistration) certificate is found in the
+        -- transaction. A long-term fix is to handle this case properly during
+        -- balancing.
+    let transform s sel =
+            let (sel', _) = assignChangeAddresses @s undefined sel s
```
Anviking: Oops. Will fix.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -297,6 +298,7 @@ import Cardano.Wallet.Primitive.AddressDiscovery.Shared
     )
 import Cardano.Wallet.Primitive.CoinSelection
     ( Selection
+    , SelectionCollateralRequirement (SelectionCollateralNotRequired, SelectionCollateralRequired)
```
jonathanknowles: Did you mean to do:
```suggestion
    , SelectionCollateralRequirement (..)
```

Or do you prefer to import these explicitly?

```suggestion
    , SelectionCollateralRequirement
        ( SelectionCollateralNotRequired
        , SelectionCollateralRequired
        )
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1264,6 +1272,139 @@ normalizeDelegationAddress s addr = do
...
+    -> ExceptT ErrBalanceTx m SealedTx
+balanceTransaction
+    tr
+    generateChange
+    (pp, nodePParams)
+    (internalUtxoAvailable, wallet, pendingTxs)
+    tl
+    resolvedInputs
+    partialTx
+    = do
```
jonathanknowles, line 1319: There are quite a lot of parameters. Would it be helpful to define a record type to hold them all?

This might also make it clearer at call sites which parameter is which (because each parameter is named). The errors when you accidentally omit a parameter can be more concise, too, rather than a long list of type errors for each parameter that has the wrong type.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1441,9 +1582,25 @@ selectAssets
...
-    guardPendingWithdrawal
     pp <- liftIO $ currentProtocolParameters nl
-    liftIO $ traceWith tr $ MsgSelectionStart
+    selectAssets' tr tl pp params transform
+  where
+    nl = ctx ^. networkLayer
+    tl = ctx ^. transactionLayer @k
+    tr = contramap MsgWallet $ ctx ^. logger
+
+-- | Less convenient and restrictive version of `selectAssets`.
```
jonathanknowles: I suspect Haddock only adds hyperlinks if you use "straight" single quotes.

```suggestion
-- | Less convenient and restrictive version of 'selectAssets'.
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1441,9 +1582,25 @@ selectAssets
...
+selectAssets'
+    :: forall m s k result. MonadRandom m
+    => Tracer m WalletLog
+    -> TransactionLayer k SealedTx
+    -> ProtocolParameters
+    -> SelectAssetsParams s result
```
jonathanknowles, line 1663: Would it be helpful to add these other parameters into the `SelectAssetsParams` record type?

Anviking:
Thanks @jonathanknowles, grouping arguments in records probably makes sense. My strategy so far was to make minimal changes and not do any premature abstractions, but I'll have a look. But I might try experimenting with writing actual property tests on top of this first, in case I learn something.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1264,6 +1272,139 @@ normalizeDelegationAddress s addr = do
...
+    -> [(TxIn, TxOut)] -- ^ Resolved inputs
+    -> SealedTx
+    -> ExceptT ErrBalanceTx m SealedTx
+balanceTransaction
+    tr
+    generateChange
+    (pp, nodePParams)
+    (internalUtxoAvailable, wallet, pendingTxs)
+    tl
+    resolvedInputs
```
KtorZ: `externalInputs` would be a better name I think. They are resolved inputs yes, but only the external ones unknown of the wallet.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1441,9 +1582,25 @@ selectAssets
...
+selectAssets'
+    :: forall m s k result. MonadRandom m
+    => Tracer m WalletLog
+    -> TransactionLayer k SealedTx
+    -> ProtocolParameters
+    -> SelectAssetsParams s result
+    -> (s -> Selection -> result)
+    -> ExceptT ErrSelectAssets m result
+selectAssets' tr tl pp params transform = do
+    guardPendingWithdrawal
```
KtorZ, line 1667: Why this function? What's the problem with pending withdrawal and this PR?

Anviking: I didn't change anything,  just moved it from `selectAssets` to here.

KtorZ: I know, but I don't get the rationale for creating `selectAssets'` and `selectAsset`. Why is this distinction needed?

Anviking: Fixed. Keeping a single version of the function but with the following tweaks:

commit c76a655732f5abdabb00ec0faab00345deb73c9b
Author: Johannes Lund <johannes.lund@iohk.io>
Date:   Wed Oct 20 19:52:28 2021 +0200

    Let selectAssets take PParams directly, rather than fetching from NetworkLayer

    Passing the full NetworkLayer seems overkill, and would have required
    less elegant mocking to property test balanceTransaction
    (calling selectAssets).

commit 1c1d8abdd559156612ffde2330d8c739e0484596
Author: Johannes Lund <johannes.lund@iohk.io>
Date:   Wed Oct 20 19:47:11 2021 +0200

    Make selectAssets run in `MonadRandom m` rather than IO

commit 014c539ad765934f8cd01510d0ee0dc6dbc12971
Author: Johannes Lund <johannes.lund@iohk.io>
Date:   Wed Oct 20 19:44:48 2021 +0200

    Add `m` to `HasLogger` constraint rather than to assume IO

lib/core/src/Cardano/Wallet.hs
Anviking: Fixed, they ended up from me using LSP code actions to auto-import.

lib/core/src/Cardano/Wallet.hs
Anviking: Mis-remembered; thanks!

lib/core/src/Cardano/Wallet.hs
Anviking: I added the type
```
data PartialTx = PartialTx
    { sealedTx :: SealedTx
    , inputs :: [(TxIn, TxOut, Maybe (Hash "Datum"))]
    , redeemers :: [Redeemer]
    } deriving (Show, Generic, Eq)
```

(which I used in my prototype property test anyway), making the signature read

```
balanceTransaction
    :: forall m s k ctx.
        ( HasTransactionLayer k ctx
        , HasLogger m WalletWorkerLog ctx
        , GenChange s
        , MonadRandom m
        )
    => ctx
    -> ArgGenChange s
    -> (W.ProtocolParameters, Cardano.ProtocolParameters)
    -> TimeInterpreter (Either PastHorizonException)
    -> (UTxOIndex, Wallet s, Set Tx)
    -> PartialTx
    -> ExceptT ErrBalanceTx m SealedTx
```

Possibly there are more things to group, but I see nothing more as obvious as `PartialTx`, which I think conceptually makes sense.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1264,6 +1284,193 @@ normalizeDelegationAddress s addr = do
                                   Transaction
 -------------------------------------------------------------------------------}
 
+data PartialTx = PartialTx
```
Anviking, line 1292: Should document this, I realised.

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
```diff
@@ -549,6 +550,22 @@ expectAndThrowFailures = hoistTimeInterpreter (runExceptT >=> eitherToIO)
     eitherToIO (Right x) = pure x
     eitherToIO (Left e) = throwIO e
 
+-- | Pre-fetches a snapshot of the epoch history from the node, such that the
+-- resulting 'TimeInterpreter' doesn't require 'IO'.
+--
+-- Please concider /not/ using this function, as it disables all logging.
```
Anviking:
```suggestion
-- Please consider /not/ using this function, as it disables all logging.
```

lib/core/src/Cardano/Wallet.hs
Anviking: Diff outdated, marking as resolved.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1264,6 +1284,199 @@ normalizeDelegationAddress s addr = do
...
+                    else
+                        SelectionCollateralNotRequired
+                } & padFeeEstimation partialTx
+
+        -- FIXME: The coin selection and reported fees will likely be wrong in
+        -- the presence of certificates (and deposits / refunds). An immediate
+        -- "fix" is to return a proper error from the handler when any key or
+        -- pool registration (resp. deregistration) certificate is found in the
+        -- transaction. A long-term fix is to handle this case properly during
+        -- balancing.
```
jonathanknowles, line 1360: **Suggestion**: remove this comment.

**Justification**: this comment is now out of date, because coin selection already properly accounts for withdrawals, deposits and refunds.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1264,6 +1284,199 @@ normalizeDelegationAddress s addr = do
...
+        let transform s sel =
+                let (sel', _) = assignChangeAddresses generateChange sel s
+                    inputs = F.toList (sel' ^. #inputs)
+                 in ( selectionDelta txOutCoin sel'
+                    , inputs
+                    , fst <$> (sel' ^. #collateral)
+                    , sel' ^. #change
+                )
```
```suggestion
                    , sel' ^. #change
                    )
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1264,6 +1284,199 @@ normalizeDelegationAddress s addr = do
...
+                , utxoAvailableForCollateral
+                , wallet
+                }
+                transform
+
+    -- NOTE:
+    -- Once the coin-selection is done, we need to
+    --
+    -- (a) Add selected inputs, collateral and change outputs to the transaction
+    -- (b) Assign correct execution units to every redeemers
```
```suggestion
    -- (b) Assign correct execution units to every redeemer
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1264,6 +1284,199 @@ normalizeDelegationAddress s addr = do
...
+    -- (c) Correctly reference redeemed entities with redeemer pointers
+    -- (d) Adjust fees and change output(s) to the new fees.
+    --
+    -- There's a strong assumption that modifying the fee value AND increasing
+    -- the coin value of change outputs does not modify transaction fees; or
```
```suggestion
    -- the coin values of change outputs does not modify transaction fees; or
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1264,6 +1284,199 @@ normalizeDelegationAddress s addr = do
...
+    -- more exactly, does not modify the execution units of scripts. This is in
+    -- principle a fair assumption because scripts validators ought to be
```
```suggestion
    -- principle a fair assumption because script validators ought to be
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1264,6 +1284,199 @@ normalizeDelegationAddress s addr = do
...
+    -- unaware of change outputs. If their execution cost increases when change
```
```suggestion
    -- unaware of change outputs. If their execution costs increase when change
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1264,6 +1284,199 @@ normalizeDelegationAddress s addr = do
...
+    -- output increases, then it becomes impossible to guarantee that the fee
```
```suggestion
    -- output values increase, then it becomes impossible to guarantee that fee
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1264,6 +1284,199 @@ normalizeDelegationAddress s addr = do
...
+    -- balancing will ever converge towards a fixed point. A script validator
+    -- doing such thing is considered bonkers and this is not a behavior we
```
```suggestion
    -- doing such a thing is considered bonkers and this is not a behavior we
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1264,6 +1284,199 @@ normalizeDelegationAddress s addr = do
...
+    candidateTx <- assembleTransaction $ TxUpdate
+        { extraInputs
+        , extraCollateral
+        , extraOutputs
+        , newFee = const delta
+        }
+    let candidateMinFee = fromMaybe (Coin 0) $
+            evaluateMinimumFee tl nodePParams candidateTx
+
+    let surplus = delta `Coin.distance` candidateMinFee
```
jonathanknowles: **Suggestion**: consider using `Coin.difference` here.

**Justification**: If this code expects that `delta >= candidateMinFee`, then `Coin.distance` will give a nonsensical answer if this expectation is violated. It's probably better to compute a `surplus` of `0` in this case.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1264,6 +1284,199 @@ normalizeDelegationAddress s addr = do
...
+            tl nodePParams ti resolveInput redeemers tx'
+      where
+        resolveInput :: TxIn -> Maybe (TxOut, Maybe (Hash "Datum"))
+        resolveInput i =
+            (\(_,o,d) -> (o,d)) <$> L.find (\(i',_,_) -> i == i') externalInputs
+            <|>
+            (\(_,o) -> (o, Nothing)) <$> L.find (\(i',_) -> i == i') (extraInputs update)
+
+    extractFromTx tx =
+        let (Tx _id _fee _coll _inps outs wdrlMap meta _vldt, toMint, toBurn) = decodeTx tl tx
```
jonathanknowles: **Suggestion**: Consider using `NamedFieldPuns` here, or otherwise use label accessors (`view #outputs`).

**Justification**: This code only cares about a subset of the record fields. Pattern matching on everything also makes the line much longer than our line length limit.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1264,6 +1284,199 @@ normalizeDelegationAddress s addr = do
...
+            -- TODO: Find a better abstraction that can cover this case.
+            wdrl = WithdrawalSelf
+                (error $ "WithdrawalSelf: reward-account should never been use "
+                      <> "when balancing transactions but it was!"
+                )
```
jonathanknowles: Suggestion: remove variable alignment and fix the broken error message.

```suggestion
                (error $ unwords
                    [ "WithdrawalSelf: reward account should never have been used"
                    , "when balancing a transaction, but it was!"
                    ]
                )
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1264,6 +1284,199 @@ normalizeDelegationAddress s addr = do
...
+                (error $ "WithdrawalSelf: derivation path should never been use "
+                      <> "when balancing transactions but it was!"
+                )
```
jonathanknowles: Suggestion: remove variable alignment and fix the broken error message.

```suggestion
                (error $ unwords
                    [ "WithdrawalSelf: derivation path should never have been used"
                    , "when balancing a transaction, but it was!"
                    ]
                )
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1264,6 +1284,199 @@ normalizeDelegationAddress s addr = do
...
+                (sumCoins wdrlMap)
+         in (outs, wdrl, meta, toMint, toBurn)
+
+    -- | Wallet coin selection is unaware of many kinds of transaction content
+    -- (e.g. datums, redeemers), which could be included in the input to
+    -- `balanceTransaction`. As a workaround we add some padding using
+    -- `evaluateMinimumFee`.
+    --
+    -- TODO: This logic needs to be consistent with how we call `selectAssets`,
+    -- so it would be good to join them into some single helper.
```
jonathanknowles: Suggestion: use Haddock-friendly quotation marks.

```suggestion
    -- 'balanceTransaction'. As a workaround we add some padding using
    -- 'evaluateMinimumFee'.
    --
    -- TODO: This logic needs to be consistent with how we call 'selectAssets',
    -- so it would be good to join them into some single helper.
```

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1264,6 +1284,199 @@ normalizeDelegationAddress s addr = do
...
+                , skeletonOutputs = view #outputs walletTx
+                , skeletonChange = mempty
+                }
+            LinearFee _ (Quantity b) = pp ^. #txParameters . #getFeePolicy
+            -- NOTE: Coping with the later additions of script integrity hash and
+            -- redeemers ex units increased from 0 to their actual values.
+            extraMargin = Coin $ ceiling $ (*) b $ fromIntegral
+                $ sizeOfScriptIntegrityHash
+                + sum ((+sizeOfRedeemerCommon) . BS.length . redeemerData <$> redeemers)
+              where
```
jonathanknowles: Suggestion: use `F.foldMap` and extract out a function `sizeOfRedeemer`.

```suggestion
                + F.foldMap sizeOfReedemer redeemers
              where
                sizeOfRedeeder
                    = (+ sizeOfRedeemerCommon) . BS.length . redeemerData
```

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2517,14 +2347,15 @@ delegationFee ctx (ApiT wid) = do
     withWorkerCtx ctx wid liftE liftE $ \wrk -> liftHandler $ do
         w <- withExceptT ErrSelectAssetsNoSuchWallet $
             W.readWalletUTxOIndex @_ @s @k wrk wid
+        pp <- liftIO $ NW.currentProtocolParameters (wrk ^. networkLayer)
         deposit <- W.calcMinimumDeposit @_ @s @k wrk wid
-        mkApiFee (Just deposit) [] <$> W.estimateFee (runSelection wrk deposit w)
+        mkApiFee (Just deposit) [] <$> W.estimateFee (runSelection wrk pp deposit w)
```
jonathanknowles: Suggestion: Wrap line to fit within line limit.

```suggestion
        mkApiFee (Just deposit) [] <$>
            W.estimateFee (runSelection wrk pp deposit w)
```

lib/core/src/Cardano/Wallet/Primitive/Slotting.hs
```diff
@@ -549,6 +550,22 @@ expectAndThrowFailures = hoistTimeInterpreter (runExceptT >=> eitherToIO)
     eitherToIO (Right x) = pure x
     eitherToIO (Left e) = throwIO e
 
+-- | Pre-fetches a snapshot of the epoch history from the node, such that the
+-- resulting 'TimeInterpreter' doesn't require 'IO'.
+--
+-- Please consider /not/ using this function, as it disables all logging.
```
jonathanknowles, line 556: 👍🏻

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -932,15 +931,14 @@ _assignScriptRedeemers (toAlonzoPParams -> pparams) ti resolveInput redeemers tx
     evaluateExecutionUnits
         :: Map Alonzo.RdmrPtr Redeemer
         -> AlonzoTx
-        -> ExceptT ErrAssignRedeemers m (Map Alonzo.RdmrPtr (Either ErrAssignRedeemers Alonzo.ExUnits))
-    evaluateExecutionUnits indexedRedeemers alonzoTx = withExceptT ErrAssignRedeemersPastHorizon $ do
+        -> Either ErrAssignRedeemers (Map Alonzo.RdmrPtr (Either ErrAssignRedeemers Alonzo.ExUnits))
+    evaluateExecutionUnits indexedRedeemers alonzoTx = left ErrAssignRedeemersPastHorizon $ do
```
jonathanknowles: Request: fit these lines within the line length limit.

```suggestion
        -> Either ErrAssignRedeemers
            (Map Alonzo.RdmrPtr (Either ErrAssignRedeemers Alonzo.ExUnits))
    evaluateExecutionUnits indexedRedeemers alonzoTx =
        left ErrAssignRedeemersPastHorizon $ do
```

jonathanknowles (commented):
Hi @Anviking 

This PR seems like a very positive step towards making `balanceTransaction` more testable. Thank-you for making it! 👍🏻 

I only have a few suggestions and requests: some of these apply to code that was moved (and therefore not "changed" by this PR per se), but I think it would still be good to clean up the highlighted areas before merging.

lib/core/src/Cardano/Wallet.hs
Anviking: No, it is still valid.

I'm sure coin-selection is self-consistent; it accounts for the [txDelegationAction](https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/Transaction.hs#L253) just fine. However the real world is bigger. In `balanceTransaction` we have no guarantee to be able to map the raw `[ledgerCert]` to `Maybe DelegationAction`.

That's (at least one reason) why I think https://input-output-rnd.slack.com/archives/GBT05825V/p1634585642350800?thread_ts=1634549806.343500&cid=GBT05825V would make sense.

lib/core/src/Cardano/Wallet.hs
Anviking: Possibly, but I think there might be usefulness to showing which fields are ignored. In-doubt, I'd prefer to keep it as is for the time being. I will move `decodeTx` to a new line, however.

lib/core/src/Cardano/Wallet.hs
Anviking: Sounds very sane 👍

Anviking:
Thanks @jonathanknowles — I believe I have addressed all suggestions except for https://github.com/input-output-hk/cardano-wallet/pull/2970#discussion_r737333734 which I think is invalid and https://github.com/input-output-hk/cardano-wallet/pull/2970#discussion_r737108283 which I'd be inclined not to do.

sevanspowell:
Given it a good read over. This seems sensible :+1: I'm a bit dubious on the padding and estimation, but I guess we'll catch anything off in our property tests.

Anviking:
> I'm a bit dubious on the padding and estimation

I'm very dubious too 😅 I wouldn't want to start changing it before having property tests, however.

jonathanknowles (approved):
Hi @Anviking 

Many thanks again for making this PR. It looks good to me!

PR #2971, opened 2021-10-12
Add witnesses to Tx in order check to submit

paweljakubas:
adp-1172

- [x] I have extended Tx to account for witnesses
- [x] I have make sure decodeTx gets wits from SealedTx
- [x] Use decodeTx in submitTx for checking if there are enough wits included 
- [x] adjust swagger, core unit and integration testing


KtorZ:
This is adding more logic to the wallet, to catch something that is already returned to us from the local tx monitor. I would **strongly** recommend not adding any extra validation logic to the wallet, but instead, analyze the error obtained from the transaction submission and, format it into a more readable error for the wallet's end users. 

The problem underlined by ADP-1172 isn't really about witnesses, but is more general: the wallet error reporting for transaction submission error is quite poor.

paweljakubas:
@KtorZ  fair argument. Thanks! I am closing this in favor of https://github.com/input-output-hk/cardano-wallet/pull/2976

PR #2976, opened 2021-10-14
Expose node errors in submitExternalTx

paweljakubas:
- [ ] I have ...

### DIFF
--- a/lib/core-integration/src/Test/Integration/Framework/TestData.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/TestData.hs
@@ -98,6 +98,7 @@ module Test.Integration.Framework.TestData
     , errMsg403TemplateInvalidScript
     , errMsg403InvalidConstructTx
     , errMsg403transactionAlreadyBalanced
+    , errMsg403TransactionInvalid
     ) where
 
 import Prelude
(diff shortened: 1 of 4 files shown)


piotr-iohk:
There is one case also where the error looks quite unfriendly. That is when tx is not serialized correctly:

```
$ cat tx.bin 
aaa

$ curl  -X POST --data-binary "@tx.bin" http://localhost:8090/v2/proxy/transactions -H "Content-Type: application/octet-stream"

{"code":"bad_request",
"message":"DecoderErrorDeserialiseFailure 'Shelley Tx' (DeserialiseFailure 0 'expected list len or indef')"}
```

The error in such a case used to be:
```
{"code":"malformed_tx_payload",
"message":"I couldn't verify that the payload has the correct binary format. Therefore I couldn't send it to the node. Please check the format and try again."}
```
This error has been logged as ADP-1145, but perhaps it is easy enough to address here?
It's covered by currently pending [test](https://github.com/input-output-hk/cardano-wallet/blob/master/test/e2e/spec/misc_spec.rb#L270).

paweljakubas:
closing in favor of https://github.com/input-output-hk/cardano-wallet/pull/3047

PR #2977, opened 2021-10-15
Display serialised tx endpoint

paweljakubas:
adp-974

- [x] swagger update
- [x] adding types
- [x] adding `decodeTransaction` in Server
- [x] add `lookupTxIns` in Wallet
- [x] add `lookupTxOuts` in Wallet
- [x] adjust unit testing
- [x] add integration tests for inputs/outputs/withdrawals/metadata

The endpoint acts in the context of a given wallet and is able to detect the wallet's resources. In detail, the enpdoint detects:
- id of tx
- fee
- inputs, belonging to other wallets and the wallets - in that case it lookups utxo and give amount and derivation path
- outputs, belonging to other wallets or to the wallets - in that case it either see its output or change (here it looks ahead using gap) and enrich info by current amount and derivation path of address
- withdrawals, belonging to other wallet or the wallet
- metadata
- script validity


lib/core/src/Cardano/Wallet.hs
```diff
@@ -1149,6 +1154,128 @@ manageRewardBalance _ ctx wid = db & \DBLayer{..} -> do
...
+        case UTxO.lookup txin (totalUTxO mempty cp) of
+            Nothing -> (txin, Nothing)
+            Just (txout@(TxOut addr _)) ->
+                let path = map (\(_, _,p) -> p) $
+                        filter (\(addr', _,_) -> addr' == addr) walletAddrs
+                in case path of
+                    [] -> (txin, Nothing) -- will probably use error here
+                    path':_ -> (txin, Just (txout, path'))
+
+extAddr
```
Anviking: 1. We could probably find a more descriptive name than `extAddr`. Seems similar to the existing `normalizeDelegationAddress`, but perhaps different enough not to combine.
2. **But**... Why are we modifying the address in the first place? 🤔 That seems very off to me. We should be looking up addresses based on their payment key hashes only, right?

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1149,6 +1154,128 @@ manageRewardBalance _ ctx wid = db & \DBLayer{..} -> do
...
+    :: forall k (n :: NetworkDiscriminant).
+        ( WalletKey k
+        , DelegationAddress n k
+        )
+    => Address
+    -> XPub
+    -> Address
+extAddr addr xpub =
+    liftDelegationAddress @n @k ((\(Right finger) -> finger) $
```
Anviking: Some `either (error "extAddr: impossible: Address cannot be malformed")` would be nicer than just ignoring the `Left` case here.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1149,6 +1154,128 @@ manageRewardBalance _ ctx wid = db & \DBLayer{..} -> do
...
+    let f (addr, addrState, ix) =
+            ( extAddr @k @n addr xpub
+            , addrState
+            , ix )
+    let walletAddrs = map f $ knownAddresses (getState cp)
+    pure $ map (tryGetTxOutPath cp walletAddrs) txins
+  where
+    db = ctx ^. dbLayer @IO @s @k
+    tryGetTxOutPath cp walletAddrs txin =
+        case UTxO.lookup txin (totalUTxO mempty cp) of
```
Anviking: Isn't `totalUTxO mempty cp` the same thing as `utxo cp`? But seems good that we don't remove the inputs from the pending txs — I take the pending txs wouldn't be decodable/displayable then.

Also, not sure we _need to_ nor if we _should_, but we _could_:
- Use `changeUTxO pending s <> utxo` (but not excluding pending inputs), to be able to resolve chains of dependent txs (I don't think this would work currently)
- Use LSQ to (potentially) `lookup` the `txin` in the node's full `UTxO`

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1149,6 +1154,128 @@ manageRewardBalance _ ctx wid = db & \DBLayer{..} -> do
...
+            Nothing -> (txin, Nothing)
+            Just (txout@(TxOut addr _)) ->
+                let path = map (\(_, _,p) -> p) $
+                        filter (\(addr', _,_) -> addr' == addr) walletAddrs
+                in case path of
+                    [] -> (txin, Nothing) -- will probably use error here
```
Anviking: > will probably use error here

Sounds sane. Interesting case 😄 — yes, if the TxIn can be looked up from the utxo, it should exist in `knownAddresses`, I would presume. Unless we were to use some different notion of `utxo` (see the other comment).

Anviking: > Unless we were to use some different notion of utxo

Or maybe for this reason keep it as is, just in case 🤷‍♂️

lib/core/src/Cardano/Wallet.hs
paweljakubas: ad 1. yes, `normaliseDelegationAddress` does almost the same as `extAddr`
ad 2. Why we needed? Because in the cbor there is delegation address as represented in the chain. and we store fingerprints really, so we need to adjust them in order to compare.

lib/core/src/Cardano/Wallet.hs
paweljakubas: I will rely on `normaliseDelegationAddress` rather than this

lib/core/src/Cardano/Wallet.hs
paweljakubas: I was thinking about pending txs. I would for now use `totalUTxO mempty cp` and think about pending tx in the next move

lib/core/src/Cardano/Wallet.hs
paweljakubas: yes, maybe for now let's keep it as it is

Anviking:
@paweljakubas Sorry, but I didn't see the point of most of the complicated (and seemingly a bit out-of-place) logic in lookupTx{Ins,Outs} 😅 I tried replacing most of it with just `isOurs` in 32480e1, and tests seemed to pass, except for in one place where I believe the test expected the wrong value (c0c9b73).

Could you let me know if it looks good to you or whether I have misunderstood something?

paweljakubas:
maybe it is ok to reuse `isOurs` but you erased the code that adds change addresses to "candidate" addresses. Why it is needed? Because this endpoint should work after construction, signing and discovery. And after construction we can omit these change addresses, right? so we need to sneak them (here assumption is that pool gap restriction is used)

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1171,103 +1165,37 @@ lookupTxIns ctx wid txins = db & \DBLayer{..} -> do
...
           $ withExceptT ErrDecodeTxNoSuchWallet
           $ withNoSuchWallet wid
           $ readCheckpoint wid
-    let f (addr, addrState, ix) =
-            ( extendAddr cp addr
-            , addrState
-            , ix )
-    let walletAddrs = map f $ knownAddresses (getState cp)
-
-    --We are analyzing the next gapPool change addresses
```
paweljakubas: so we need this changeAddrs, otherwise source wallet will not see change address in outputs UNTIL the wallet discovers tx (so after submit). That is consequence of decoupling construction/signing/submitting and persisting pending things only during submitting

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -479,7 +479,7 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
         let addrDest = (addrs !! addrIx) ^. #id
         let expectedTxOutTarget = WalletOutput $ ApiWalletOutput
                 { address = addrDest
-                , amount = Quantity 0
```
paweljakubas: no it should be 0. This is what is currently in this address. So after construction, source wallet does not see this as WalletOutput but ExternalOutput. Target wallet see this as WalletOuptut, and see `amountIncoming=Quantity amt`, but `amount=Quantity 0` as there is nothing at THIS stage in this wallet. Right? It is going to be `amount=amountIncoming` after discovery. I did this on purpose to have a way to see whole cycle if this resource is ours :-)

paweljakubas: see line 548 and 
```
        let expectedTxOutTarget' = WalletOutput $ ApiWalletOutput
                { address = addrDest
                , amount = Quantity amt -- now we have this
                , assets = ApiT TokenMap.empty
                , derivationPath = NE.fromList
                    [ ApiT (DerivationIndex 2147485500)
                    , ApiT (DerivationIndex 2147485463)
                    , ApiT (DerivationIndex 2147483648)
                    , ApiT (DerivationIndex 0)
                    , ApiT (DerivationIndex $ fromIntegral addrIx)
                    ]
                , amountIncoming = Quantity amt
                , assetsIncoming = ApiT TokenMap.empty
                }
```

paweljakubas: this is after submitting and discovering

Anviking (approved):
lgtm! Hoping we at some point later on can get CBOR <-> JSON roundtrip properties, which I believe would be the ultimate test that we display all information!

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -395,8 +456,62 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
...
+
+        rDecodedTxSource <- request @(ApiDecodedTransaction n) ctx
+            (Link.decodeTransaction @'Shelley wa) Default decodePayload
+        verify rDecodedTxSource $
+            sharedExpectationsBetweenWallets ++
+            [ expectField #inputs (`shouldSatisfy` areOurs)
+            , expectField #outputs (`shouldNotContain` [expectedTxOutTarget])
+
+            -- Check that the change output is there:
+            , expectField (#outputs) ((`shouldBe` 1) . length . filter isOutOurs)
```
Anviking, line 504: I added a check for the change output in the last commit along with a slight tweak, using `should{Not}Contain [txOut]` rather than `shouldSatisfy` and `elem`, which should produce nicer error messages and be slightly simpler.

paweljakubas:
@Anviking thanks for the thorough review and commits! I will create following ticket in which we property test using genTx and expose mint/burn and other things in this endpoint

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
paweljakubas: good change

PR #2981, opened 2021-10-20
flake.nix: Add aarch64-darwin to hydraJobs

hamishmack:
Adds `aarch64-darwin` to release.nix.  Something like this will be handy when hydra has `aarch64-darwin` builders.  We could test it on `ci.zw3rk.com`.


rvl:
Thanks @hamishmack - I rebased this on latest master which now uses the `hydraJobs` nix flake attribute.

rvl (approved):

PR #2983, opened 2021-10-20
Enable more tests for construct ep

piotr-iohk:
- bcf78bf4592f291fe8bf91debf90ae837d15becc
  Enable remaining construct ep integration tests (where applicable)
  
- df75842f6da952ac2a4dfea6840b0106c27f5f9c
  Make signTx DSL useful also in balance -> sign -> submit workflow
  
- 4e26f5d2c1423a4c2530ab953396d9fc0f00bcad
  Update swagger to reflect current state of affairs, in particular removing Byron Transactions New altogether from displaying on rendered api doc
  
- d80d6b675465d3a0b3fc2b8ffacd6a505592a368
  Use expectedFee returned by construct ep rather than explicit value

### Comments

Wanted to add more tests for balance ep, but noticed first that few tests for construct can be enabled. 
Also updated swagger so it reflects current state more accurately.

### Issue Number

sort of ADP-1223

### DIFF
--- a/lib/core-integration/src/Test/Integration/Framework/DSL.hs
+++ b/lib/core-integration/src/Test/Integration/Framework/DSL.hs
@@ -235,7 +235,6 @@ import Cardano.Wallet.Api.Types
     , ApiBlockReference (..)
     , ApiByronWallet
     , ApiCoinSelection
-    , ApiConstructTransaction (transaction)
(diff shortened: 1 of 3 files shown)


paweljakubas (approved):
lgtm

PR #2993, opened 2021-10-27
Verify coin selection balancing errors.

jonathanknowles:
## Issue Number

ADP-1037

## Summary

This PR finalizes the verification of selection balancing errors returned from `CoinSelection.performSelection`.

## Details

This PR:

- adds verification for errors of the following types:
    - `EmptyUTxO`.
    - `BalanceInsufficientError`.
    - `InsufficientMinCoinValueError`.
    - `UnableToConstructChangeError`.
- adds helper functions to simplify verification functions.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
@@ -121,7 +121,7 @@ import Data.Maybe
 import Data.Ratio
     ( (%) )
 import Data.Semigroup
-    ( mtimesDefault )
+    ( All (..), mtimesDefault )
 import Fmt
     ( Buildable (..), genericF )
 import GHC.Generics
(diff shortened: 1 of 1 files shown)


lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
```diff
@@ -718,6 +821,73 @@ verifySelectionLimitReachedError cs ps e
...
+  where
+    -- The result of attempting to re-run the selection process with minimal
+    -- constraints, where we have:
+    --
+    --   - a minimum cost function that always returns zero.
+    --   - a minimum ada quantity function that always returns zero.
+    --
+    resultWithMinimalConstraints :: Either SelectionError Selection
+    resultWithMinimalConstraints =
+        runIdentity $ runExceptT $ performSelection cs' ps
```
sevanspowell, line 1058:
```suggestion
        -- By using our Identity instance of MonadRandom, we always generate the
        -- same number. This means we'll always pick the same numbered entry in
        -- our UTxO set. This isn't an issue however, because when we select
        -- elements from the UTxO set, we remove them, so we won't always select
        -- the same input.
        runIdentity $ runExceptT $ performSelection cs' ps
```

What do you think of this?

sevanspowell (approved):

lib/core/src/Cardano/Wallet/Primitive/CoinSelection.hs
jonathanknowles: It's an excellent suggestion: I'll add it in!

jonathanknowles: Fixed in 7e8aced4087d2425432016106b29f67b398e42a9.

PR #2995, opened 2021-11-01
Simplifications to decode tx endpoint

Anviking:
- [x] Remove distinction between amount and amount_incoming
- [x] Simplify lookupTx{Ins,Outs} using isOurs 

### Comments

In response to #2977 

### Issue Number

ADP-974


Anviking:
bors try

PR #2996, opened 2021-11-01
Add prop_balanceTransactionBalanced

Anviking:
## Overview

- Add preconditions to `balanceTransaction` such that we fail instead of producing imbalanced results...
    - when there are withdrawals with different network tags
    - when there are deposits or refunds (pool reg, stake key reg/dereg) (apparently not pool retirement)
    - when collateral is pre-selected
    - when we underestimate the fee (likely related to variable length coin encoding and the padding)
- Tweak generators in `Cardano.Api.Gen`
- Add (pending) `prop_balanceTransactionBalanced`
    - Decent shrinking and failure-presentation
    - Used to discover the preconditions above
    - Generator coverages of some corner cases are missing
        - empty txIns
        - empty txOuts
        - gens never cause balanceTransaction to select collateral for some reason (!)
        - PParams are currently static
    - In-the-end marked pending because of a new variable-length-coin-encoding corner cases

## Comments

The problems we theorised during the coarse first implementation are indeed caught by the property (collateral, fee encoding corner cases, deposits), which is nice. Discovering additional problems was nice too.

While the coverage is still not ideal, I believe it's good enough to give re-writing the implementation a go. 

### Issue Number

ADP-1214


lib/core/src/Cardano/Wallet/Primitive/Types/Coin/Gen.hs
```diff
@@ -19,7 +20,11 @@ import Test.QuickCheck
 --------------------------------------------------------------------------------
 
 genCoin :: Gen Coin
-genCoin = sized $ \n -> Coin . fromIntegral <$> choose (0, n)
+genCoin = frequency
+    [ (1, pure (Coin 0))
+    , (5, Coin <$> choose (1_000_000, 1_000_000_000_000))
+    ]
```
jonathanknowles: Hi @Anviking 

We recently pushed a series of PRs to convert many generators (such as this one) to use the QC size parameter.

I think using the size parameter is a sensible default, since:
- for many tests of pure algorithms that don't care about the full range of a coin (coin selection, migration), we just need small-ish values.
- QC will adjust this parameter automatically over several runs to provide a reasonable distribution.
- the size parameter is adjustable with [`scale`](https://hackage.haskell.org/package/QuickCheck-2.14.2/docs/Test-QuickCheck.html#v:scale) and [`resize`](https://hackage.haskell.org/package/QuickCheck-2.14.2/docs/Test-QuickCheck.html#v:resize), so it's easy for people writing properties that need different scales to modify the generators on which they rely to get the coverage they need.

For tests that specifically need very large values, or special ranges, we usually provide dedicated generators.

By removing the connection to the `size` parameter, we make it more difficult/impossible for properties that consume these generators to use [`scale`](https://hackage.haskell.org/package/QuickCheck-2.14.2/docs/Test-QuickCheck.html#v:scale), and more difficult for people writing tests to get the coverage they need.

I don't mind changing strategy, though if we're going to change strategy (and change generators such as this one so that they no longer use the size parameter), then perhaps we should discuss our strategy together, so that we can gain consensus about how we wish to proceed?

Perhaps the development meeting would be a good opportunity?

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -1216,16 +1282,18 @@ emptyTxSkeleton = mkTxSkeleton
...
 mockProtocolParameters = dummyProtocolParameters
-    { txParameters = TxParameters
+    { executionUnitPrices = Just (ExecutionUnitPrices 1 1)
+    , txParameters = TxParameters
         { getFeePolicy = mockFeePolicy
         , getTxMaxSize = Quantity 16384
         , getTokenBundleMaxSize = TokenBundleMaxSize $ TxSize 4000
         , getMaxExecutionUnits = ExecutionUnits 10000000 10000000000
         }
+    , minimumUTxOvalue = MinimumUTxOValue $ Coin 1000000
```
jonathanknowles, line 1311: Suggestion: use `NumericUnderscores` for large values such as these.

lib/core/test/unit/Cardano/Api/GenSpec.hs
```diff
@@ -433,7 +432,7 @@ instance Arbitrary SlotNo where
     arbitrary = genSlotNo
 
 genLovelaceCoverage :: Lovelace -> Property
-genLovelaceCoverage = unsignedCoverage (maxBound @Word16) "lovelace"
+genLovelaceCoverage = unsignedCoverage (maxBound @Word32) "lovelace"
```
Anviking, line 437: `Lovelace` is unbounded — is having to pass in a Word32 bound weird?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -3721,6 +3722,30 @@ instance IsServerError ErrBalanceTx where
                 ]
         ErrBalanceTxSelectAssets err -> toServerError err
         ErrBalanceTxAssignRedeemers err -> toServerError err
+        ErrBalanceTxNotYetSupported ConflictingNetworks ->
+            apiError err403 CreatedInvalidTransaction $ mconcat
+                [ "There are withdrawals for multiple networks (e.g. both "
+                , "mainnet and testnet) in the provided transaction. This "
+                , "makes no sense, and I'm confused."
```
jonathanknowles, line 3777:
```suggestion
                , "mainnet and testnet) in the provided transaction. This "
                , "is not supported."
```

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -1438,6 +1520,389 @@ instance Arbitrary KeyHash where
...
+                (utxo, wal, pending)
+                partialTx) (show . Pretty) $ \case
+            Right (sealedTx ) -> do
+                label "success"
+                    $ classify (originalBalance == Cardano.Lovelace 0)
+                        "already balanced"
+                    $ classify (txFee sealedTx > Cardano.Lovelace 1_000_000)
+                        "fee above 1 ada"
+                    (txBalance sealedTx combinedUTxO === 0)
+                    .&&. (abs (txFee sealedTx) .<= 4_000_000)
```
Anviking, line 1807: TODO: Test `fee >= minFee` but also not too much of an over-estimate. Perhaps another property (and definitely another PR/ticket)

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: I thought I'd be consistent with the other anthropomorphised error-messages 😂

Anviking:
bors try

Anviking:
bors try

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1457,11 +1476,80 @@ balanceTransaction
...
 
+    -- Coin selection does not support pre-defining collateral. In Sep 2021
+    -- consensus was that we /could/ allow for it with just a day's work or so,
+    -- but that the need for it was unclear enough that it was not in any way
+    -- a priority.
+    case Cardano.txInsCollateral bod of
+        Cardano.TxInsCollateralNone -> return ()
+        Cardano.TxInsCollateral _ [] -> return ()
+        Cardano.TxInsCollateral _ _ ->
+            throwE $ ErrBalanceTxNotYetSupported ExistingCollateral
```
Anviking: This is not expected to change, so the wrapping `ErrBalanceTxNotYetSupported ` might be inappropriate. It should deserve a top-level constructor.

Anviking:
bors try

Anviking:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -1245,6 +1246,16 @@ fromCardanoTxIn
 fromCardanoTxIn (Cardano.TxIn txid (Cardano.TxIx ix)) =
     W.TxIn (fromShelleyTxId $ Cardano.toShelleyTxId txid) (fromIntegral ix)
 
+fromCardanoTxOut :: IsCardanoEra era => Cardano.TxOut era -> W.TxOut
```
Anviking:
```suggestion
- | WARNING: Datum hashes are lost in the conversion!
fromCardanoTxOut :: IsCardanoEra era => Cardano.TxOut era -> W.TxOut
```

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -3721,6 +3722,30 @@ instance IsServerError ErrBalanceTx where
...
+                ]
+        ErrBalanceTxExistingCollateral ->
+            apiError err403 CreatedInvalidTransaction $ mconcat
+                [ "I cannot balance transactions with pre-defined collateral."
+                ]
+        ErrBalanceTxNotYetSupported ZeroAdaOutput ->
+            apiError err500 CreatedInvalidTransaction $ mconcat
+                [ "The transaction contains one or more outputs with "
```
sevanspowell:
```suggestion
                [ "The transaction contains one or more outputs with zero ada outputs."
```

lib/core/test/unit/Cardano/Api/GenSpec.hs
sevanspowell: This is just bad UX on my part. You don't need to pass in the maxBound of the unsigned you are testing the coverage of, but you do need to pass in some value that the test can use to delineate between "large" and "small" values (with things above the value being considered large).

I'm not really happy with this code.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1414,11 +1433,83 @@ balanceTransaction
...
+    let zeroAdaOutputs =
+            filter (\o -> view (#tokens . #coin) o == Coin 0 ) outputs
+    unless (null zeroAdaOutputs) $
+        throwE $ ErrBalanceTxNotYetSupported ZeroAdaOutput
+
+    case Cardano.txCertificates bod of
+        Cardano.TxCertificatesNone -> return ()
+        Cardano.TxCertificates _ certs _
+            | any isReg certs -> throwE $ ErrBalanceTxNotYetSupported Deposits
+            | otherwise -> return ()
```
Anviking: Should move — this check is slightly in the wrong place

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1431,11 +1433,83 @@ balanceTransaction
...
+    --
+    -- For instance the partial tx might contain two withdrawals with the same
+    -- key but different networks:
+    -- [ (Mainnet, pkh1, coin1)
+    -- , (Testnet, pkh1, coin2)
+    -- ]
+    --
+    -- Even though this is absurd, the node/ledger @evaluateTransactionBalance@
+    -- will count @coin1+coin2@ towards the total balance. Because the wallet
+    -- does not concider the network tag, it will drop one of the two, leading
```
sevanspowell:
```suggestion
    -- does not consider the network tag, it will drop one of the two, leading
```

sevanspowell:
Thanks for your work on this Johannes :tada: I think I'm ok with this, as long as my understanding of it is correct: This is a first step in fixing bugs with the current balanceTransaction implementation.

I'm not sure how I feel about the coverage of the generators (which is my fault too!), but that's unimportant I think. The important thing is that the generators and the test have found issues, and we should use these going forward to fix the implementation of balance transaction.

I have one unimportant suggestion here: https://github.com/input-output-hk/cardano-wallet/pull/3052, but otherwise, looks good!

sevanspowell (approved):

lib/core/src/Cardano/Wallet/Primitive/Types/Coin/Gen.hs
Anviking: I ended up not having to modify this Arbitrary instance.

Anviking:
Thanks! Yes, this is a first step!

bors r+

PR #2997, opened 2021-11-02
Flakify nix build

jbgi:
Adds a `flake.nix` file.

## Checklist

- [x] Flake build with `nix build .#cardano-wallet` succeeds.
- [x] (optional) Legacy build with `nix-build -A cardano-wallet` works
- [x] Release build with `nix-build release.nix -A cardano-wallet-linux64` works.
- [x] Cabal build inside `nix develop` works.
- [x] Stack build inside `nix develop` works.
- [x] Direnv config uses flake.
- [x] Flake is built in a Hydra jobset. All jobs pass in this jobset and there are no evaluation errors. => https://hydra.iohk.io/jobset/Cardano/cardano-wallet-flake
- [x] PR opened on `ci-ops` to update Hydra jobsets for cardano-wallet master branch, bors staging/trying, and PRs, so that they build with nix flakes. => https://github.com/input-output-hk/ci-ops/pull/85
- [x] Development environment (`nix develop`) is cached on Hydra - for Linux at least.
- [x] Release build is cached on Hydra.
- [x] Nix closure size of cardano-wallet derivation has not increased due to unnecessary added dependencies.
- [x] Documentation (docs directory, nix code comments, etc) has been updated to use nix flake commands instead of nix-build/nix-shell.
- [x] Instructions for bumping nix dependencies have been updated to use `nix flake` commands.
- [x] Buildkite and GitHub actions CI passes.

## Issue number

ADP-983


rvl (commented):
This is amazing, thanks @jbgi.

rvl:
I tried `nix develop` on this branch and it seems to be building quite a few haskell library dependencies.
But really it should just be getting cached builds from Hydra, right?

jbgi:
Unfortunately I think native flake is downloading inputs dependencies using a different mechanism (`builtins.fetchGit` or `builtins.fetchTarball` I believe) than what our flake-compat is using (`pkgs.fetchzip`), in order to work in restricted mode.
This in turn modify derivation inputs and requires some rebuild... :(
I guess this should be solved with native flake support in hydra and/or allowing required third-party repo in restricted mode.

rvl:
We would need caching from Hydra. The result of this should be cached at least, right?
```
nix-build -A cardano-wallet
```

nix/nixos/cardano-wallet-service.nix
```diff
@@ -0,0 +1,193 @@
...
+
+    package = mkOption {
+      type = types.package;
+      default = cfg.project.hsPkgs.cardano-wallet.components.exes.cardano-wallet;
+      description = "Package for the cardano wallet executable.";
+    };
+
+    listenAddress = mkOption {
+      type = types.str;
+      default = "*";
```
rvl: Better default to `127.0.0.1` I think.

nix/nixos/cardano-wallet-service.nix
```diff
@@ -0,0 +1,193 @@
...
+          "--database \"${cfg.database}\""
+        ++ lib.mapAttrsToList
+          (name: level: ''
+            --trace-${name}=${level} '')
+          cfg.trace
+        ++ cfg.rtsArgs)
+      ;
+    };
+
+    project = mkOption {
```
rvl: Usually most services just have the `package` option. I don't think we need the haskell.nix `project` to also be configurable.

nix/nixos/cardano-wallet-service.nix
```diff
@@ -0,0 +1,193 @@
...
+  inherit (lib) optionalString mkIf mkEnableOption mkOption types;
+  logLevels = types.enum [ "debug" "info" "notice" "warning" "error" "critical" "alert" "emergency" "off" ];
+  cardanoEnvAttrs = cfg.environments.${cfg.environment};
+in
+{
+  options.services.cardano-wallet = {
+
+    enable = mkEnableOption "Cardano Wallet service";
+
+    script = mkOption {
```
rvl: Can we change the `script` option into a list of command-line arguments? This seems to be more usual in nixos:

```nix
serverArgs = mkOption {
  description = "Command-line to launch cardano-wallet server.";
  type = types.separatedString " ";
  default = lib.concatStringsSep " " [ "--listen-address" (lib.escapeShellArg cfg.listenAddress) /* ... */ ];
};
```

nix/nixos/cardano-wallet-service.nix
```diff
@@ -0,0 +1,193 @@
...
+    listenAddress = mkOption {
+      type = types.str;
+      default = "*";
+      description = "Which host to bind the API server to.";
+    };
+
+    port = mkOption {
+      type = types.port;
+      default = 8090;
+      description = "The default port cardano wallet will listen on.";
```
rvl: "The port on which the cardano-wallet HTTP API server will listen."

nix/nixos/cardano-wallet-service.nix
```diff
@@ -0,0 +1,193 @@
...
+        iohk-nix/cardano-lib/default.nix, or, it can be a custom
+        name, for example, if launching a test cluster that does
+        not have an environment yet defined in iohk-nix.  If custom,
+        the environments option must contain an entry for this environment,
+        of the same form as in cardano-lib environments found
+        in iohk-nix.
+      '';
+    };
+
+    environments = mkOption {
```
rvl: I'm not too keen on `environments` style configuration directly in `services.cardano-wallet`.
It would be better as a layer on top.

This higher-level layer would configure the services to work together according to the chosen environment. E.g:
```
{ config, lib, ... }: {
services.cardano = {
  environment = lib.cardano.mainnet;
  wallet.enable = true;
  node.enable = true;
  faucet = {
    enable = true;
    keyFile = config.deployment.keys.faucet-key.path;
  };
  dbsync.enable = false;
};
}
```

nix/nixos/cardano-wallet-service.nix
```diff
@@ -0,0 +1,193 @@
...
+    };
+
+    syncTolerance = mkOption {
+      type = types.ints.positive;
+      default = 300;
+      description = "Time duration within which we consider being synced with the network. Expressed in seconds.";
+    };
+
+    poolMetadataFetching = mkOption {
+      type = types.enum [ "none" "direct" "smash" ];
```
rvl: Could the type please be like this?
```
type = types.nullOr (types.submodule {
  options = {
    enable = mkEnableOption "Stake pool metadata fetching.";
    smashURL = mkOption {
      description = ''
        URL of SMASH metadata proxy to use.
        If null, metadata will be fetched directly from the
        stake pool's URL.
      '';
      type = types.nullOr types.str;
      default = null;
    };
  };
});
```

nix/nixos/cardano-wallet-service.nix
```diff
@@ -0,0 +1,193 @@
...
+      type = types.attrsOf logLevels;
+      default = { };
+      description = ''
+        For each tracer, minimum severity for a message to be logged, or
+        "off" to disable the tracer".
+      '';
+    };
+
+    rtsArgs = mkOption {
+      type = types.listOf types.str;
```
rvl: This seems to be more commonly used in nixos:
```
rtsOpts = mkOption {
  type = types.separatedString " ";
  default = "-N2";
  example = "-M2G -N4";
  description = ''
    GHC runtime-system options for the cardano-wallet process.
    See https://downloads.haskell.org/ghc/8.10.7/docs/html/users_guide/runtime_control.html#setting-rts-options-on-the-command-line for documentation.
  '';
};
```

nix/nixos/cardano-wallet-service.nix
```diff
@@ -0,0 +1,193 @@
...
+  config = mkIf cfg.enable {
+
+    systemd.services.cardano-wallet = {
+      description = "cardano-wallet daemon";
+      wantedBy = [ "multi-user.target" ];
+      inherit (cfg) script;
+
+      serviceConfig = {
+        DynamicUser = true;
+        StateDirectory = "cardano-wallet/${cfg.environment}";
```
rvl: Since it's not possible to run multiple wallet servers with different environments under the same nixos configuration, I don't think it's necessary to add the environment name into the state directory path.

nix/nixos/cardano-wallet-service.nix
```diff
@@ -0,0 +1,193 @@
...
+
+  config = mkIf cfg.enable {
+
+    systemd.services.cardano-wallet = {
+      description = "cardano-wallet daemon";
+      wantedBy = [ "multi-user.target" ];
+      inherit (cfg) script;
+
+      serviceConfig = {
+        DynamicUser = true;
```
rvl: Will this user have permission to open the cardano-node socket file?
We really want it so that the socket file is protected by permissions, and only members of the `cardano-node` group (for example) can open it.
But how would someone add the cardano-wallet service's user to the correct group?

nix/nixos/cardano-wallet-service.nix
```diff
@@ -0,0 +1,193 @@
...
+    };
+
+  };
+
+  config = mkIf cfg.enable {
+
+    systemd.services.cardano-wallet = {
+      description = "cardano-wallet daemon";
+      wantedBy = [ "multi-user.target" ];
+      inherit (cfg) script;
```
rvl: Can we do it like this?
```
serviceConfig.ExecStart = "${cfg.package}/bin/cardano-wallet serve ${cfg.rtsOpts} ${cfg.serverArgs}";
```

rvl (changes requested):
A NixOS service, great thanks @jbgi!
Would it be possible to get this in a separate PR so that we can merge before the big flake refactor?
And a nixos test for the service would be awesome. It doesn't need to be part of CI though, just something we could run locally.

nix/nixos/cardano-wallet-service.nix
jbgi: `serviceConfig.SupplementaryGroups = "cardano-node";` should work.

rvl (commented):
I tested this branch again.
1. A cabal build inside `nix develop` seems to work, which is good. I did not check whether the devShell was fully cached by Hydra.
2. When I try `nix --print-build-logs build .#cardano-wallet` in a completely fresh git worktree it proceeds to build cardano-wallet locally - we need this build to be cached by Hydra.

rvl (approved):
Thanks @jbgi - fantastic work. :+1:
I split off the NixOS service into #3083.
We will merge this immediately after the upcoming release.

PR #2999, opened 2021-11-03
Add generic n-ary product shrinking

sevanspowell:
- Translate the round-robin shrinking algorithm to work on n-ary products (types
  with one constructor [product types], with n constructor arguments), in the
  Generics.SOP universe.
- Write a property to assert that the old selectionParams shrinker and the new
  one produce the same result.
- Replace all instances of liftShrinkN with a generic alternative.

See the following resource for more info on generics-sop: https://raw.githubusercontent.com/kosmikus/cufp-tutorial-2016/master/LectureNotes.pdf


jonathanknowles (approved):
Thanks for making this PR @sevanspowell!

LGTM.

PR #3000, opened 2021-11-03
Workaround for bech32 on Windows e2e tests

piotr-iohk:
- 21f613e617a64e086990e820d9631ed83498e4d5
  Workaround for bech32 on Windows

### Issue Number

n/a

### DIFF
--- a/test/e2e/helpers/utils.rb
+++ b/test/e2e/helpers/utils.rb
@@ -1,6 +1,7 @@
 require 'bip_mnemonic'
 require 'httparty'
 require 'fileutils'
+require 'tmpdir'
 
 module Helpers
   module Utils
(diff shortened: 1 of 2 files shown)


HeinrichApfelmus (approved):
LGTM!

One could use `popen` in order to write to the stdinput of the `bech32` process directly, but I don't know how one would do that in Ruby — and it might still not work as expected on Windows. Using a temporary file works.

Anviking:
It seems `bech32` strips `\n` here: https://github.com/input-output-hk/bech32/blob/7d3c9fb61a014069ad5dcae643dd12c9ebbeb6f3/bech32/app/Main.hs#L125

Maybe a non-workaround solution would be also removing `\r\n`, which I guess might be involved, e.g. using [T.strip](https://hackage.haskell.org/package/text-1.2.5.0/docs/Data-Text.html#v:strip)

PR #2990, opened 2021-10-22
Improve scripts for haskell-language-server and ghci

rvl:
It now generates a custom HLS config on the fly. Instructions are in `hie-direnv.yaml`.

### Issue Number

ADP-1229

PR #3001, opened 2021-11-04
Fix haskell-language-server / direnv config

rvl:
Last minute shell script change in #2990 broke it - oops.

This PR also adds a CI step to check the hls config.

### Issue Number

ADP-1229


Anviking (approved):
Thanks! Seems to work.

PR #3002, opened 2021-11-04
Manual test covering ADP-871

piotr-iohk:
- [x] I have added manual test for ADP-871 case to be executed as part of release checklist

### Comments
Although the issue was also apparent in automated e2e tests, I think there is no harm in adding additional check.

### Issue Number

ADP-871


HeinrichApfelmus (approved):
LGTM! 😊

PR #3004, opened 2021-11-08
Operators for generic n-ary product shrinking

jonathanknowles:
## Issue Number

#2999

## Summary

This PR provides operators to make it easier to use generic n-ary product shrinking.
   
These operators allow us to compose shrinkers with a syntax that is very
similar to the applicative syntax for composing generators. This makes it easier to compare the definitions of generators and shrinkers that are related to one another.

For example:

```haskell

data MyRecord = MyRecord
    { foo :: Foo
    , bar :: Bar
    , baz :: Baz
    }
    deriving (Eq, Generic, Show)

genMyRecord :: Gen MyRecord
genMyRecord = MyRecord
    <$> genFoo
    <*> genBar
    <*> genBaz

shrinkMyRecord :: MyRecord -> [MyRecord]
shrinkMyRecord = genericRoundRobinShrink
    <@> shrinkFoo
    <:> shrinkBar
    <:> shrinkBaz
    <:> Nil

```

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Balance/Gen.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/CoinSelection/Balance/Gen.hs
@@ -30,7 +30,7 @@ import Test.QuickCheck
     , shrinkMapBy
     )
 import Test.QuickCheck.Extra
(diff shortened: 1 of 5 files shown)


sevanspowell (approved):
Nice :+1:

PR #3006, opened 2021-11-09
Add GitHub action to auto-approve docs PRs

rvl:
Adds a CI step which looks at what changed in a PR. If it's "only" docs which changed, then the PR gets an automatic review approval.

### Comments

Reviews aren't automatically dismissed if non-docs changes are added to the PR. That would require some modifications to the https://github.com/hmarr/auto-approve-action code.

### Issue Number

ADP-1254

### DIFF
--- a/.github/workflows/approve-docs.yml
+++ b/.github/workflows/approve-docs.yml
@@ -1,0 +1,21 @@
+name: Auto-Approve Docs
+on: pull_request_target
+
+jobs:
+  review:
+    if: '${{ github.ref_protected }}'
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: 'Check scope of PR'
+        run: './scripts/gh/check-scope.sh scope'
+        id: check
+
+      # Note that GitHub actions do not run from forked repos, so this
+      # will only approve PRs from branches in the main repo.
+      - name: 'Approve Docs'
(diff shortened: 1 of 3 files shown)


sevanspowell (approved):
Looks good :+1: 

Checked the check-scope.sh script locally and it seems to work.

.github/workflows/approve-docs.yml
```diff
@@ -0,0 +1,31 @@
...
+          dirs=$(./scripts/check-scope.sh "remotes/origin/$base" "remotes/origin/$branch")
+          echo
+          echo "Set scope output"
+          echo "::set-output name=scope::$dirs"
+        id: check
+
+      # Note that GitHub actions do not run from forked repos, so this
+      # will only approve PRs from branches in the main repo.
+      - name: 'Approve Docs'
+        if: "${{ steps.check.outputs.scope == 'docs' }}"
```
sevanspowell, line 18: The output of check-scope.sh is an array, which is then assigned to `steps.check.outputs.scope`. So you're checking that an array is equal to a single element? What happens in that case?

rvl: It's not an array, just a normal string.
So I want to check that there was exactly one top-level directory changed, and it was `docs`. This can be done with string comparison.

Your point stands, however. I realise that I was not handling multi-line strings properly when setting the output variable. I just pushed an amended commit which reworks newline handling.

PR #3012, opened 2021-11-10
Replace `ApiCoinSelectionInput` type with `ApiWalletInput`.

jonathanknowles:
These types are identical, and their serialized forms are also identical.


paweljakubas (approved):
Agree here `ApiCoinSelectionInput` and `ApiWalletInput` are the same

PR #3019, opened 2021-11-15
Tighten criteria for accepting Shelley or Byron addresses in REST API

HeinrichApfelmus:
### Summary

It has become established usage that Byron addresses are associated with the Base58 encoding while Shelley addresses are associated with the Bech32 encoding.

To catch more user errors, we now enforce this usage more strictly.

Specifically,

* Base58 encoded addresses are assumed to be Byron addresses and are additionally deserialized to CBOR to check whether they comply with the  Byron address binary format.
* We no longer accept Base16 encoded addresses. The swagger documentation has already highlighted that only Base58 and Bech32 are accepted.

To avoid loss of funds, we strongly recommend to use Shelley addresses and the Bech32 encoding, as this encoding includes error detection and is more robust against misspellings and truncations.

Relevant standards: [CIP 19 — Cardano addresses](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0019).

### Comments

(EDIT: correction)

* Ideally, lay users would only see a single canonical version of every address. Unfortunately, we are not quite there yet, and this PR is only a partial improvement. For example,
    * Bech32-encoded addresses can represent Byron addresses instead of Shelley addresses. These would be shown in Base58 encoding in the blockchain explorer.
    * A sequence of bytes which is too long to be a valid Shelley address, but has a header-byte that indicates a Shelley address, may be truncated to the proper size. Put differently, seemingly invalid addresses are mapped to valid ones. Fortunately, the error correction inherent to Bech32 and reserving the Base58 encoding for Byron addresses (implemented in this PR) make this less likely to occur through user mistake.
* A general clean up of our address data types would be nice, but is out of scope for this pull request.

### Issue Number

ADP-1033, ADP-1280

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/CLI/Shelley/Transactions.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/CLI/Shelley/Transactions.hs
@@ -105,6 +105,7 @@ import Test.Integration.Framework.TestData
     , errMsg404CannotFindTx
     , errMsg404NoWallet
     , falseWalletIds
+    , invalidByronBase58
     , kanjiWalletName
     , polishWalletName
     , wildcardsWalletName
@@ -935,24 +936,19 @@ spec = describe "SHELLEY_CLI_TRANSACTIONS" $ do
           ]
 
       encodeErr = "Unrecognized address encoding"
-      encodeErr2 = "Unable to decode address"
       parseErr = "Parse error. Expecting format \"<amount>@<address>\""
       matrixInvalidAddrs =
--- TODO: For the haskell node, hex is valid. For jormungandr it is not.
---  longAddr = replicate 10000 '1'
---  We should optimally find a way to test this.
---          [ ( "long hex", longAddr, encodeErr )
---          , ( "short hex", "1", encodeErr )
           [ ( "-1000", "-1000", encodeErr )
-          , ( "q", "q", encodeErr2 )
-          , ( "empty", "", encodeErr2 )
+          , ( "q", "q", encodeErr )
+          , ( "empty", "", encodeErr )
           , ( "wildcards", T.unpack wildcardsWalletName, parseErr )
           , ( "arabic", T.unpack arabicWalletName, encodeErr )
           , ( "kanji", T.unpack kanjiWalletName, encodeErr )
           , ( "polish", T.unpack polishWalletName, encodeErr )
           , ( "[]", "[]", encodeErr )
-          , ( "no address", "", encodeErr2 )
+          , ( "no address", "", encodeErr )
           , ( "address is space", " ", encodeErr )
+          , ( "invalid Byron", T.unpack invalidByronBase58, encodeErr)
           ]
       errNum = "Expecting natural number"
       matrixInvalidAmt =
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
@@ -2043,10 +2043,10 @@ decodeBytes t =
         Just bytes ->
             Right bytes
         _ ->
-            Left $ TextDecodingError $ mconcat
+            Left $ TextDecodingError $ unwords
                 [ "Unrecognized address encoding: must be either bech32 or base58."
-                , " Perhaps your address is not entirely right?"
-                , " Every character has to be correct!"
+                , "Perhaps your address is not entirely correct?"
+                , "Please double-check each character within the address and try again."
                 ]
 
 -- | Attempt to decode a Shelley 'Address' using a Bech32 encoding.
--- a/lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
+++ b/lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
@@ -232,12 +232,10 @@ import Data.Binary.Put
     ( putByteString, putWord8, runPut )
 import Data.Bits
     ( (.&.), (.|.) )
-import Data.ByteArray.Encoding
-    ( Base (Base16), convertFromBase )
 import Data.ByteString
     ( ByteString )
 import Data.ByteString.Base58
-    ( bitcoinAlphabet, decodeBase58, encodeBase58 )
+    ( bitcoinAlphabet, encodeBase58 )
 import Data.ByteString.Short
     ( fromShort, toShort )
 import Data.Coerce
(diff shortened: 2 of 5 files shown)


lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -2040,28 +2039,36 @@ instance DecodeAddress ('Testnet pm) where
...
         Just bytes ->
             Right bytes
         _ ->
-            Left $ TextDecodingError
-                "Unrecognized address encoding: must be either bech32, base58 or base16"
+            Left $ TextDecodingError $ mconcat
+                [ "Unrecognized address encoding: must be either bech32 or base58."
+                , " Perhaps your address is not entirely right?"
+                , " Every character has to be correct!"
+                ]
```
jonathanknowles: Suggestion: by using `unwords` here, rather than `mconcat`, we can avoid having to manually insert spaces.

```suggestion
            Left $ TextDecodingError $ unwords
                [ "Unrecognized address encoding: must be either bech32 or base58."
                , "Perhaps your address is not entirely right?"
                , "Every character has to be correct!"
                ]
```

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -2040,28 +2039,36 @@ instance DecodeAddress ('Testnet pm) where
...
+    case tryBech32 t <|> tryBase58 t of
         Just bytes ->
             Right bytes
         _ ->
-            Left $ TextDecodingError
-                "Unrecognized address encoding: must be either bech32, base58 or base16"
+            Left $ TextDecodingError $ mconcat
+                [ "Unrecognized address encoding: must be either bech32 or base58."
+                , " Perhaps your address is not entirely right?"
+                , " Every character has to be correct!"
```
jonathanknowles: Perhaps consider the following wording:
> Perhaps your address is not entirely correct?
> Please double-check each character within the address and try again.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -2040,28 +2039,36 @@ instance DecodeAddress ('Testnet pm) where
...
+                ]
 
--- | Attempt decoding an 'Address' using a Bech32 encoding.
+-- | Attempt to decode a Shelley 'Address' using a Bech32 encoding.
 tryBech32 :: Text -> Maybe ByteString
-tryBech32 text = do
-    (_, dp) <- either (const Nothing) Just (Bech32.decodeLenient text)
-    dataPartToBytes dp
+tryBech32 = fmap CA.unAddress . CA.fromBech32
```
jonathanknowles, line 2054: By following `CA.fromBech32` to its definition, I think I have confirmed that `fromBech32` does indeed also use `Bech32.decodeLenient`. However, I just want to check: do we have regression test coverage to warn us if this ever changes?

(I ask, because it's just inside the realm of possibility that someone might change the definition of `CA.fromBech32` to use `Bech32.decode`, rather than its more lenient variant.)

jonathanknowles (approved):
Hi @HeinrichApfelmus 

Thank-you for making this PR!

I've had a look through, and it seems very reasonable. Though I have a question and a couple of small suggestions. (See comments.)

Thanks!

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
HeinrichApfelmus: Hm. I'm not sure if we should check this, for two reasons:

1. The address format is standardized in [CIP-0019][1] which lists the `cardano-addresses` package as a reference implementation. So, regardless of how `Cardano.Address.fromBech32` is implemented, it must be correct — by virtue of being a reference. 😅
2. Using `Bech32.decode` instead of `Bech32.decodeLenient` would actually not be a regression — as far as the definition and error correcting properties of the Bech32 encoding are concerned. 🙈  ([BIP-0173][bech32]: "A Bech32 string is at most 90 characters long and consists of […]"). Shelley addresses are typically 57 bytes in length (1 byte header + 28 bytes payment part + 28 bytes delegation part), so that would work out. The big exception are pointer addresses which can be arbitrarily long, but of which there are only a handful of instances on mainnet. 🙈 

  [1]: https://github.com/cardano-foundation/CIPs/tree/master/CIP-0019#reference-implementations
  [bech32]: https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki

HeinrichApfelmus: Oh wait. Even the regular Shelley addresses would not quite fit into the 90 character limit, as 57 bytes = 456 bits = 92.5 * (5 bits), and one character encodes 5 bits.

HeinrichApfelmus: I have added a test that covers Shelly addresses with delegation part; these will start to fail if the 90 character limit is enforced. In this way, we are testing functionality while covering the formalities.

HeinrichApfelmus:
Thanks for your review, Jonathan! 😊 

Will merge unless there are further objections.

rvl (commented):
Would it be possible to add a note to the swagger API doc, just to make clear the decoding behaviour to users?

lib/core-integration/src/Test/Integration/Scenario/CLI/Shelley/Transactions.hs
```diff
@@ -935,23 +935,17 @@ spec = describe "SHELLEY_CLI_TRANSACTIONS" $ do
...
-          , ( "empty", "", encodeErr2 )
+          , ( "q", "q", encodeErr )
+          , ( "empty", "", encodeErr )
           , ( "wildcards", T.unpack wildcardsWalletName, parseErr )
           , ( "arabic", T.unpack arabicWalletName, encodeErr )
           , ( "kanji", T.unpack kanjiWalletName, encodeErr )
           , ( "polish", T.unpack polishWalletName, encodeErr )
           , ( "[]", "[]", encodeErr )
-          , ( "no address", "", encodeErr2 )
+          , ( "no address", "", encodeErr )
```
piotr-iohk, line 949: Maybe it would be actually good to add addresses from the issue to that list?

HeinrichApfelmus: For reasons of privacy, I suppose that we do not want to add the addresses from the issue. However, I have added a randomly generated Base58 string that is not a valid Byron address.

PR #3022, opened 2021-11-16
Make sign tx testable

sevanspowell:
This PR is primarily about improving the testability of signTransaction, and providing a number of property tests.

- [x] In the signTransaction code path, pushed IO further to the edge of the program, enabling us to unit test more of the logic of signTransaction.
- [x] Provided a number of property tests to prove key points of signTransaction (it's witness adding behaviour)
    - [x] Also, specifically added a test to assert that signTransaction maintains the script integrity hash of a transaction, for ADP-1047.
- [x]  Expanded the transaction generator slightly to possibly generate ScriptWitnesses for TxIns.

### Issue Number

ADP-1047 / ADP-1139


rvl:
@sevanspowell Could we please have a title and description for this PR?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1878,8 +1881,31 @@ signTransaction ctx (ApiT wid) body = do
     let pwd = coerce $ body ^. #passphrase . #getApiT
     let sealedTx = body ^. #transaction . #getApiT
 
-    sealedTx' <- withWorkerCtx ctx wid liftE liftE $ \wrk ->
-        liftHandler $ W.signTransaction wrk wid pwd sealedTx
+    sealedTx' <- withWorkerCtx ctx wid liftE liftE $ \wrk -> liftHandler $ do
+        let
+            db = wrk ^. W.dbLayer @IO @s @k
```
paweljakubas, line 1886: We have those parts in almost all places in `where`. Shouldn't we have here also (just for consistency)?

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -281,43 +337,18 @@ import Test.Utils.Paths
...
-    ( posixSecondsToUTCTime )
-import GHC.Generics
-    ( Generic )
-import Test.QuickCheck.Property
-    ( label )
-
 import qualified Cardano.Api as Cardano
+import qualified Cardano.Api.Gen as Cardano
 import qualified Cardano.Api.Shelley as Cardano
+import qualified Cardano.Crypto.Hash.Blake2b as Blake2b
```
paweljakubas: what about using here `as Crypto` ?

paweljakubas: or just import without qualification like in other places in codebase?

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1968,43 +1969,39 @@ selectAssets ctx pp params transform = do
...
+  -- ^ The total UTxO set of the wallet (i.e. if pending transactions all
+  -- applied).
+  -> SealedTx
+  -- ^ The transaction to sign
+  -> SealedTx
+  -- ^ The original transaction, with additional signatures added where
+  -- necessary
+signTransaction tl era keyLookup (rootKey, rootPwd) utxo =
+    let
+        rewardAcnt :: (XPrv, Passphrase "encryption")
```
paweljakubas, line 1995: what about exposing this function? It is used also in tests. Maybe it would be good to have code reuse

paweljakubas (approved):
very nice properties added. It would be great to continue this path of testing when adding test coverage of decodeTx (in some forthcoming PRs). 
1. We see that signing can add some bits and this is tested.What about properties showing that signing will not change inps, outs, validity, ...., some things' ordering already present in the body?  
2. I see common patterns in properties. For example,  we have the same structure:
```
case Cardano.THINGSupportedInEra era of
    Nothing -> 
    Just _ -> 
```  
I wonder if there is a way to extract that common structure and reuse it in all property tests?

lib/core/src/Cardano/Wallet/Api/Server.hs
sevanspowell: It won't work in this case, because `wrk` is not an argument to the outer function, `wrk` won't be in scope in the `where` block.

sevanspowell:
> very nice properties added. It would be great to continue this path of testing when adding test coverage of decodeTx (in some forthcoming PRs).
> 
>     1. We see that signing can add some bits and this is tested.What about properties showing that signing will not change inps, outs, validity, ...., some things' ordering already present in the body?
> 
>     2. I see common patterns in properties. For example,  we have the same structure:
> 
> 
> ```
> case Cardano.THINGSupportedInEra era of
>     Nothing -> 
>     Just _ -> 
> ```
> 
> I wonder if there is a way to extract that common structure and reuse it in all property tests?

Thanks for the feedback @paweljakubas! I've addressed a number of your points in subsequent commits.

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -345,6 +376,434 @@ spec = do
...
+    -> (XPrv, Passphrase "encryption")
+    -- ^ Root key of wallet
+    -> UTxO
+    -- ^ UTxO of wallet
+    -> [(XPrv, Passphrase "encryption")]
+    -- ^ Extra keys to form basis of address -> key lookup function
+    -> Property
+prop_signTransaction_neverChangesTxBody (AnyCardanoEra era) rootK utxo extraKeys =
+    forAll (genTxInEra era) $ \tx -> do
+        let
```
rvl, line 760: May as well combine the let and where bindings at the deeper level...
```suggestion
    forAll (genTxInEra era) $ \tx -> bodyContentBefore == bodyContentAfter
        where
            tl = testTxLayer
            
```
(same goes for the other `prop_` functions)

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -345,6 +376,434 @@ spec = do
...
+       )
+    => [a]
+    -> [a]
+    -> Property
+checkSubsetOf elems xs = do
+    conjoin $ (flip foldMap) elems $ \x ->
+        [ x `elem` xs
+        & counterexample ("actual set: " <> show xs)
+        & counterexample ("expected elem: " <> show x)
+        ]
```
rvl: I don't think the foldMap is needed?
```suggestion
    counterexample ("actual set: " <> show xs) $ conjoin
        [ x `elem` xs & counterexample ("expected elem: " <> show x)
        | x <- elems ]
```

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -345,6 +376,434 @@ spec = do
     transactionConstraintsSpec
     updateSealedTxSpec
     balanceTransactionSpec
+    describe "Sign transaction" $ do
+        forAllEras (\era -> it ("signTransaction adds reward account witness when necessary (" <> show era <> ")") $
+            property (prop_signTransaction_addsRewardAccountKey era))
```
rvl, line 381: I was thinking that it would be nice to have the property description/right beside the property definition, but still exposing the `Property` as a standalone function. Something like this...

```suggestion
        describe desc_signTransaction_addsRewardAccountKey $
            forAllEras $ \era -> it (show era) $
                property (prop_signTransaction_addsRewardAccountKey era))
            
        -- ... --
        
desc_signTransaction_addsRewardAccountKey = "signTransaction adds reward account witness when necessary" :: String
```

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -345,6 +376,434 @@ spec = do
...
+    -> Property
+prop_signTransaction_addsRewardAccountKey (AnyCardanoEra era) rootXPrv utxo wdrlAmt = do
+    whenSupportedInEra Cardano.withdrawalsSupportedInEra era $
+        \(supported :: Cardano.WithdrawalsSupportedInEra era) -> do
+        let
+            rootK :: (ShelleyKey 'RootK XPrv, Passphrase "encryption")
+            rootK = first liftRawKey rootXPrv
+
+            rawRewardK :: (XPrv, Passphrase "encryption")
+            rawRewardK = (getRawKey $ deriveRewardAccount (snd rootK) (fst rootK), snd rootK)
```
rvl, line 465: If we find that tests are taking too long to run due to key generation, we could shift these things up into a precooked record of key inputs.

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -345,6 +376,434 @@ spec = do
...
+                expectedWits =
+                    InAnyCardanoEra era <$>
+                        case Cardano.cardanoEraStyle era of
+                            LegacyByronEra ->
+                                error "Withdrawal witnesses are not supported in the Byron era."
+                            ShelleyBasedEra _ ->
+                                [mkShelleyWitness txBody rawRewardK]
+
+            checkCoverage $
+                expectedWits `checkSubsetOf` (getSealedTxWitnesses sealedTx')
```
rvl, line 501: Cool. I was wondering whether the reward account xpub could be used to verify new witnesses, but it seems easier and more effective to just generate the witness directly, as you have done.

rvl (approved):
Excellent work - these are really good tests to have! :star:

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
sevanspowell: Agree, could be cool. But I wouldn't want these to be the only tests that do it.

sevanspowell:
bors try

sevanspowell:
bors cancel

PR #3026, opened 2021-11-18
Lower the wanted constraints on constructTransaction and selectCoins

dfordivam:
This change simply lowers the wanted constraints on a couple of APIs, thereby ensuring that future changes to these APIs keep them general enough to ensure their usage in a wider context.


jonathanknowles (approved):
Hi @dfordivam 

Many thanks for making this PR. This change looks good to me.

I suspect that there are also other functions to which we could apply the same treatment!

PR #3027, opened 2021-11-18
Less wallets in e2e tests

piotr-iohk:
- c3c42888c71c34ee8bec07b4b9b6d4c84c78ddbb
  Limit nb of wallets in e2e tests
  
- 637f4ff3e2071f86bd82866b7dbd0c9d1da0d652
  Migrate all funds back to fixture wallet back from target wallet that's used across the test suite

### Comments
Using single target wallet for transactions in e2e tests and extending/refactoring assertions for checking balances on target and source wallets before and after transactions taking in account amounts, assets and fees. At the end of the suite all funds are migrated back to the fixture wallet which is a good test in itself.

### Issue Number

ADP-1196

### DIFF
--- a/.github/workflows/e2e-docker.yml
+++ b/.github/workflows/e2e-docker.yml
@@ -2,7 +2,7 @@ name: E2E Docker
 
 on:
   schedule:
-  - cron:  "0 23 * * *"
+  - cron:  "0 0 * * *"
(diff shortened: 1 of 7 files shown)


paweljakubas (approved):
lgtm

PR #3029, opened 2021-11-19
Verify minted and burned assets are decoded in mint-burn contract

piotr-iohk:
- ad6e0d6437f2ec5f9d8adb792ed0d442a2fc4ddf
  Verify minted and burned assets are decoded in mint-burn contract

### Issue Number

ADP-1245

PR #3033, opened 2021-11-23
Fix mint-burn assets decode assertion

piotr-iohk:
- [x] I have fixed mint-burn assets decode assertion introduced in https://github.com/input-output-hk/cardano-wallet/pull/3029. 

### Comments
Policy id of minted/burned asset in this test depends on fixture wallet's pub key so it is different for each fixture wallet.

### Issue Number

ADP-1245


rvl (approved):
OK

PR #3034, opened 2021-11-23
Redefine `Coin` in terms of `Natural` and remove the `Bounded` instance.

jonathanknowles:
## Issue Number

ADP-1283

## Summary

This PR:
- Redefines `Coin` in terms of `Natural` (rather than `Word64`).
- Removes the `Bounded` instance for `Coin`.
- Adds `txOut{Min,Max}Coin` constants to `Primitive.Types.Tx`, in the same style as the existing `txOut{Min,Max}TokenQuantity` constants.
- Pushes validation checks for coins that must be within a certain range to the places where those checks are actually required.

## Implementation Notes

Where possible, this PR uses `intCast` and `intCastMaybe` (instead of `fromIntegral`) to perform statically-checked integral conversations.

## Motivation

The wallet uses the `Coin` type to represent an amount of _lovelace_.

It's currently defined in terms of `Word64`, and has a `Bounded` instance that limits the range to `[0, 45_000_000_000_000_000]`.

However, this approach is problematic, for several reasons:

- Boundary checks for `Coin` values make sense in contexts where there are boundaries.  However, for pure `Coin` values in the absence of any particular context, there isn't one particular upper bound that makes sense.
- The current choice of `Bounded` specifically defines the limits of what can be included in a _transaction output_. However, encoding values that appear in transaction outputs is not the only use of the `Coin` type.
- For example, it's quite reasonable to use `Coin` for other purposes, like finding the total volume of ada transacted over some period of time. For such usages, there is no obvious choice of upper limit that we could consider “ideal”.
- Right now, we often use awkward workarounds such as casting `Coin` values to `Natural` values before computing summations, thus losing the descriptive value of using the `Coin` type to mark that “this is a quantity of lovelace”.
- Given that the default data constructor for `Coin` is exported, it’s possible to directly construct “invalid” Coin values from outside the `Coin` module.
- Using the existing `Semigroup` instance, we can combine valid `Coin` values into values that are **_invalid_**, which can lead to unexpected run-time errors when those “invalid” values are passed to functions that we believe to be safe.  💣 

For an example of the last problem, consider this:

```haskell
> :set -XNumericUnderscores
> :set -XOverloadedLists
> import qualified Cardano.Wallet.Primitive.Types.Coin as Coin
> coinA = Coin 45_000_000_000_000_000
> coinB = coinA <> coinA
> coinB
Coin 90000000000000000
> Coin.equipartition coinB [()]
*** Exception: unsafeNaturalToCoin: overflow
  error, called at /home/jsk/projects/input-output-hk/cardano-wallet-master/lib/core/src/Cardano/Wallet/Primitive/Types/Coin.hs:131:37 in main:Cardano.Wallet.Primitive.Types.Coin
```

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -420,7 +420,7 @@ import Cardano.Wallet.Primitive.Types
 import Cardano.Wallet.Primitive.Types.Address
     ( Address (..), AddressState (..) )
 import Cardano.Wallet.Primitive.Types.Coin
-    ( Coin (..), coinQuantity )
+    ( Coin (..) )
 import Cardano.Wallet.Primitive.Types.Hash
     ( Hash (..) )
 import Cardano.Wallet.Primitive.Types.Redeemer
@@ -596,6 +596,7 @@ import qualified Cardano.Wallet.Primitive.AddressDerivation.Icarus as Icarus
 import qualified Cardano.Wallet.Primitive.CoinSelection.Balance as Balance
 import qualified Cardano.Wallet.Primitive.CoinSelection.Collateral as Collateral
 import qualified Cardano.Wallet.Primitive.Types as W
+import qualified Cardano.Wallet.Primitive.Types.Coin as Coin
 import qualified Cardano.Wallet.Primitive.Types.TokenBundle as TokenBundle
 import qualified Cardano.Wallet.Primitive.Types.TokenMap as TokenMap
 import qualified Cardano.Wallet.Primitive.Types.Tx as W
@@ -3241,8 +3242,8 @@ mkApiFee :: Maybe Coin -> [Coin] -> FeeEstimation -> ApiFee
 mkApiFee mDeposit minCoins (FeeEstimation estMin estMax) = ApiFee
     { estimatedMin = qty estMin
     , estimatedMax = qty estMax
-    , minimumCoins = coinQuantity <$> minCoins
-    , deposit = coinQuantity $ fromMaybe (Coin 0) mDeposit
+    , minimumCoins = Coin.unsafeToQuantity <$> minCoins
+    , deposit = Coin.unsafeToQuantity $ fromMaybe (Coin 0) mDeposit
     }
   where
     qty = Quantity . fromIntegral
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -3831,7 +3831,7 @@ instance IsServerError ErrCannotQuit where
                 , "although you're not even delegating, nor won't be in an "
                 , "immediate future."
                 ]
-        ErrNonNullRewards (Coin rewards) ->
+        ErrNonNullRewards rewards ->
             apiError err403 NonNullRewards $ mconcat
                 [ "It seems that you're trying to retire from delegation "
                 , "although you've unspoiled rewards in your rewards "
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -3242,8 +3242,8 @@ mkApiFee :: Maybe Coin -> [Coin] -> FeeEstimation -> ApiFee
 mkApiFee mDeposit minCoins (FeeEstimation estMin estMax) = ApiFee
     { estimatedMin = qty estMin
     , estimatedMax = qty estMax
-    , minimumCoins = Coin.unsafeToQuantity <$> minCoins
-    , deposit = Coin.unsafeToQuantity $ fromMaybe (Coin 0) mDeposit
+    , minimumCoins = Quantity . Coin.toNatural <$> minCoins
+    , deposit = Quantity . Coin.toNatural $ fromMaybe (Coin 0) mDeposit
     }
   where
     qty = Quantity . fromIntegral
--- a/lib/core/src/Cardano/Wallet/Primitive/Model.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Model.hs
@@ -67,7 +67,7 @@ import Cardano.Wallet.Primitive.Types
 import Cardano.Wallet.Primitive.Types.Address
     ( Address (..) )
 import Cardano.Wallet.Primitive.Types.Coin
-    ( Coin (..), distance, sumCoins )
+    ( Coin (..), distance )
 import Cardano.Wallet.Primitive.Types.RewardAccount
     ( RewardAccount (..) )
 import Cardano.Wallet.Primitive.Types.TokenBundle
@@ -531,7 +531,7 @@ prefilterBlock b u0 = runState $ do
 
                 (Nothing, Outgoing) -> -- Byron
                     let
-                        totalOut = sumCoins (txOutCoin <$> outputs tx)
+                        totalOut = F.fold (txOutCoin <$> outputs tx)
 
                         totalIn = TB.getCoin spent
                     in
--- a/lib/core/src/Cardano/Wallet/Primitive/Types/Coin.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types/Coin.hs
@@ -82,12 +82,17 @@ import qualified Prelude
 --
 -- Reminder: 1 ada = 1,000,000 lovelace.
 --
+-- The 'Coin' type has 'Semigroup' and 'Monoid' instances that correspond
+-- to ordinary addition and summation.
+--
 newtype Coin = Coin
     { unCoin :: Natural
     }
     deriving stock (Ord, Eq, Generic)
     deriving (Read, Show) via (Quiet Coin)
 
+-- | The 'Semigroup' instance for 'Coin' corresponds to ordinary addition.
+--
 instance Semigroup Coin where
     -- Natural doesn't have a default Semigroup instance.
     (<>) = add
--- a/lib/core/src/Cardano/Wallet/Primitive/Types/Coin.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Types/Coin.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE ExplicitForAll #-}
 {-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE TypeApplications #-}
 
@@ -13,6 +14,17 @@
 module Cardano.Wallet.Primitive.Types.Coin
     ( -- * Type
       Coin (..)
+
+      -- * Conversions (Safe)
+    , fromIntegral
+    , fromWord64
+    , toWord64
+
+      -- * Conversions (Unsafe)
+    , unsafeFromIntegral
+    , unsafeToWord64
+
+      -- * Compatibility
     , coinQuantity
     , coinToInteger
     , coinToNatural
@@ -35,7 +47,8 @@ module Cardano.Wallet.Primitive.Types.Coin
 
     ) where
 
-import Prelude
+import Prelude hiding
+    ( fromIntegral )
 
 import Cardano.Numeric.Util
     ( equipartitionNatural, partitionNatural )
@@ -43,10 +56,14 @@ import Control.DeepSeq
     ( NFData (..) )
 import Control.Monad
     ( (<=<) )
+import Data.Bits
+    ( Bits )
 import Data.Foldable
     ( foldl' )
 import Data.Hashable
     ( Hashable )
+import Data.IntCast
+    ( intCast, intCastMaybe )
 import Data.List.NonEmpty
     ( NonEmpty (..) )
 import Data.Maybe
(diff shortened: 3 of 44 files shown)


lib/core/src/Cardano/Wallet/Primitive/Types/UTxO.hs
```diff
@@ -297,8 +297,10 @@ log10 = Log10
 
 -- | Compute UtxoStatistics from UTxOs
 computeUtxoStatistics :: BoundType -> UTxO -> UTxOStatistics
-computeUtxoStatistics btype =
-    computeStatistics (pure . unCoin . txOutCoin) btype . Map.elems . unUTxO
+computeUtxoStatistics btype
+    = computeStatistics (pure . fromIntegral . unCoin . txOutCoin) btype
```
Anviking:
```suggestion
    = computeStatistics (pure . Coin.unsafeToWord64 . txOutCoin) btype
```

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -3241,8 +3242,8 @@ mkApiFee :: Maybe Coin -> [Coin] -> FeeEstimation -> ApiFee
 mkApiFee mDeposit minCoins (FeeEstimation estMin estMax) = ApiFee
     { estimatedMin = qty estMin
     , estimatedMax = qty estMax
-    , minimumCoins = coinQuantity <$> minCoins
-    , deposit = coinQuantity $ fromMaybe (Coin 0) mDeposit
+    , minimumCoins = Coin.unsafeToQuantity <$> minCoins
+    , deposit = Coin.unsafeToQuantity $ fromMaybe (Coin 0) mDeposit
```
Anviking: We're converting to a `Natural`, so this ends up being safe. Can't we use `Coin.toNatural` instead?

Same for minimumCoins.

lib/core/src/Cardano/Wallet/Primitive/Types/Coin.hs
```diff
@@ -302,9 +298,7 @@ equipartition
     -> NonEmpty Coin
     -- ^ The partitioned coins.
 equipartition c =
-    -- Note: the natural-to-coin conversion is safe, as partitioning guarantees
-    -- to produce values that are less than or equal to the original value.
-    fmap unsafeNaturalToCoin . equipartitionNatural (toNatural c)
```
Anviking: 👍

lib/core/src/Cardano/Wallet/Primitive/Types/Coin.hs
```diff
@@ -108,13 +106,7 @@ instance ToText Coin where
...
 
 instance FromText Coin where
-    fromText = validate <=< (fmap Coin . fromText @Natural)
-      where
-        validate x
-            | isValidCoin x =
-                return x
-            | otherwise =
-                Left $ TextDecodingError "Coin value is out of bounds"
+    fromText = fmap Coin . fromText @Natural
```
Anviking, line 107: This seems like a _slight_ change of behaviour in the CLI https://github.com/input-output-hk/cardano-wallet/blob/a1838496ba697208daf2889ffcd420bea4d1b0a5/lib/cli/src/Cardano/CLI.hs#L1408-L1409

but I guess it makes sense — `stake` is not related to a single TxOut ada value.

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
```diff
@@ -953,6 +956,15 @@ mockSealedTx :: HasCallStack => ByteString -> SealedTx
...
     (internalError "mockSealedTx: attempted to decode gibberish")
 
+{-------------------------------------------------------------------------------
+                          Checks
+-------------------------------------------------------------------------------}
+
+coinIsValidForTxOut :: Coin -> Bool
+coinIsValidForTxOut c = (&&)
+    (c >= txOutMinCoin)
+    (c <= txOutMaxCoin)
```
Anviking, line 966: Are we sure we couldn't have e.g. a 46B ada TxOut on a custom testnet?

I find this hard-coded value in the ledger specs
https://github.com/input-output-hk/cardano-ledger/blob/4347e54b2827090e898702ccfb93cc731b4a1909/eras/byron/ledger/impl/src/Cardano/Chain/Common/Lovelace.hs#L176, but it is for _Byron_.

So I _guess_ you could setup a single-era cluster post-Byron with a different max supply. I presume this is what would calculate it: https://github.com/input-output-hk/cardano-ledger/blob/4347e54b2827090e898702ccfb93cc731b4a1909/eras/shelley/impl/src/Cardano/Ledger/Shelley/Genesis.hs#L434

For the proper `Cardano` mode (all eras), I suspect 45B still holds, though. So this might not be the most important thing to care about even if true.

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -531,7 +531,7 @@ prefilterBlock b u0 = runState $ do
 
                 (Nothing, Outgoing) -> -- Byron
                     let
-                        totalOut = sumCoins (txOutCoin <$> outputs tx)
+                        totalOut = F.fold (txOutCoin <$> outputs tx)
```
Anviking, line 534: I appreciate the clarity of the name `sumCoins` over `F.fold`, but I won't per-se object with this change either.

lib/core/src/Cardano/Wallet/Primitive/Types/Tx/Gen.hs
```diff
@@ -235,6 +259,67 @@ shrinkTxOut (TxOut a b) = uncurry TxOut <$> shrinkInterleaved
...
+--
+genTxOutCoin :: Gen Coin
+genTxOutCoin = frequency
+    [ (1, pure txOutMinCoin)
+    , (1, pure txOutMaxCoin)
+    , (8, Coin.fromNatural <$> chooseNatural
+        ( Coin.toNatural txOutMinCoin + 1
+        , Coin.toNatural txOutMaxCoin - 1
+        )
+      )
```
Anviking, line 282: Would it be an idea to do something like this instead?
```haskell
n <- fromIntegral . getNonNegative <$> arbitrary @Int :: Natural
frequency
  [ (1, Coin.fromNatural $ txOutMinCoin + n)
  , (1, Coin.fromNatural $ txOutMaxCoin - n)
  , (8, Coin.fromNatural <$> chooseNatural (min, max))
  ] 
```
Motivation:
- More coverage _near_ boundaries
- Manually excluding the boundary cases from the full-range `chooseNatural` seems pointless

lib/shelley/test/unit/Cardano/Wallet/Shelley/CompatibilitySpec.hs
```diff
@@ -409,7 +411,7 @@ prop_assessTokenBundleSize_enlarge b1' b2' =
     assess b1 == TokenBundleSizeExceedsLimit ==> conjoin
         [ assess (b1 `TokenBundle.add` b2)
             === TokenBundleSizeExceedsLimit
-        , assess (b1 `TokenBundle.setCoin` maxBound)
+        , assess (b1 `TokenBundle.setCoin` txOutMaxCoin)
```
Anviking, line 414: 👌

Anviking (approved):
LGTM, I think the direction makes sense. (And thanks for tackling this!) I left a few comments from going through the commits start to finish, once.

lib/core/src/Cardano/Wallet/Primitive/Types/Coin.hs
```diff
@@ -69,26 +76,21 @@ import Quiet
...
--- However be careful when summing coins, for example, if calculating historical
--- volumes traded, because this may overflow.
 newtype Coin = Coin
-    { unCoin :: Word64
+    { unCoin :: Natural
     }
     deriving stock (Ord, Eq, Generic)
     deriving (Read, Show) via (Quiet Coin)
 
 instance Semigroup Coin where
```
HeinrichApfelmus, line 96: One of the most important things about the `Coin` abstraction is that it has a `Semigroup` and a `Monoid` instance which correspond to *summation*. Could you add corresponding comments in the Haddock documentation?

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
HeinrichApfelmus: I agree that we may want to avoid hard-coding the minimum and maximum value for `Coin` — they are actually configured in the Shelley genesis. (In fact, our local cluster has a higher maximum Ada quantity than 45B.)

In fact, I believe that we can remove `txOutMaxCoin` entirely — checking the maximum bound will never catch any user mistakes that will not manifest in a different way (e.g. inability to find inputs with sufficient funds).

When discussing this, Jonathan suggested to make a second PR in order to keep the scope of this PR to a refactoring (as opposed to a change in functionality).

lib/shelley/cardano-wallet.cabal
```diff
@@ -61,6 +61,7 @@ library
     , filepath
     , fmt
     , generic-lens
+    , int-cast
```
HeinrichApfelmus, line 64: We use a bewildering variety of packages. At some point, we may want to document and consolidate our choices, so that writing mundane code becomes more streamlined (e.g. Text vs ByteString vs lazy ByteString; converting to/from hex). 😅 But I do not believe that the variety really hurts readability, so the status quo is fine as well.

HeinrichApfelmus (approved):

lib/core/src/Cardano/Wallet/Primitive/Types/UTxO.hs
jonathanknowles: Ah, well spotted! I'll fix this.

lib/core/src/Cardano/Wallet/Primitive/Types/Coin.hs
jonathanknowles: Yes. I agree this should be acceptable. As you mention, this is unrelated to a transaction output.

If `stake` needs to be constrained further, then we could perhaps make a wrapper type that constrains the value further. Thoughts?

lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs
jonathanknowles: @Anviking I completely agree with your observation. This value could be different in custom testnets, so hard-coding it within the wallet primitive layer is of questionable value.

Though as @HeinrichApfelmus mentioned, it was my aim to avoid changing any behaviour in this PR.

I think we could/should consider making a ticket to capture this issue. I'll try to do that today.

lib/core/src/Cardano/Wallet/Primitive/Model.hs
jonathanknowles: > I appreciate the clarity of the name sumCoins over F.fold, but I won't per-se object with this change either.

I think my (not very strongly felt) objection to this stems from the fact that `sumCoins` seems specialized, but in reality it folds over the coins in a very similar way to how `F.fold` adds with the `Semigroup` and `Monoid` instances, so it _seems_ a bit superfluous.

But my objection is not a strong one! I would prefer the name `sum` though, if we were to retain it, since the current API of `Coin` is intended to be imported in qualified fashion.

lib/core/src/Cardano/Wallet/Primitive/Types/Tx/Gen.hs
jonathanknowles: I think we could definitely consider that!

Though perhaps in a separate PR?

(My aim with this PR was to keep it as a pure refactoring, as much as possible.)

lib/core/src/Cardano/Wallet/Primitive/Types/Coin.hs
jonathanknowles: Good point!

I've added some comments in 535d0dd85454b3d2ea1206c04ef13169c801c826.

lib/shelley/cardano-wallet.cabal
jonathanknowles: I do agree we use a very large variety of packages!

Though to some extent, perhaps this reflects the fragmentation and/or modularity within the Haskell package ecosystem?

In this case, I believe the `int-cast` package is rather well-designed and very useful for catching errors. We actually already do use it within `cardano-wallet`: this just adds it to some of the other build targets.

lib/core/src/Cardano/Wallet/Api/Server.hs
jonathanknowles: Good point!

I've fixed this in 1937bf7212ef86e67dc7957c5105a62aad5107ef.

lib/core/src/Cardano/Wallet/Primitive/Types/UTxO.hs
jonathanknowles: Fixed in 0b95b179480e2d7766eebf369b4616805ef0746d.

lib/shelley/cardano-wallet.cabal
jonathanknowles: > ByteString vs lazy ByteString

To some extent, I think our hand is forced here by the packages we rely on.

Historically, Aeson has used lazy `ByteString` by default. Though I've noticed that strict variants have recently been added for the most common functions.

jonathanknowles:
Due to problems with hydra, I plan to merge this PR **_manually_**. 🔧 

I have confirmed that the following test suites pass when run locally (for commit a8d68a3371287a44843799e0828fdbb0e3e0427a):
- `cardano-wallet-core:test:unit`
- `cardano-wallet:test:unit`
- `cardano-wallet:test:integration`

I have run each test suite **_twice_** to guard against spurious test failures.

PR #3036, opened 2021-11-24
Add integration test for Plutus.Contracts.Currency

HeinrichApfelmus:
This pull request adds an integration test for the smart contract [Plutus.Contract.Currency](https://github.com/input-output-hk/plutus-apps/blob/main/plutus-use-cases/src/Plutus/Contracts/Currency.hs) from the [plutus-use-cases](https://github.com/input-output-hk/plutus-apps/tree/main/plutus-use-cases) repository.

This test essentially constructs a transaction in binary format (CBOR) that is suitable for the `transactions-balance` endpoint. The Plutus script is included in pre-complied form. This approach is somewhat low-level, but follows the style of the integration tests that came before me.

### Comments

* For completeness, the code required to obtain the pre-compiled contract from the Plutus source code is recorded in the file `lib/core-integration/aux/Plutus/CurrencyContract.hs`. I have not made any attempt to include the necessary Plutus dependencies in the `core-integration` package. One should be able to run it any REPL that loads the `plutus-use-cases` package.
* The pre-compiled Plutus contract needs to be edited in order to include a specific TxOutRef. However, it's not as simple as replacing a few bytes, because the binary encoding of Plutus contract is a *bit*-wise encoding, defined in the [flat][] package. For reference, I have reimplemented the flat encoding of integers in the file `lib/core-integration/aux/Plutus/FlatInteger.hs`.
* To make the binary format (CBOR) of the transaction easier to understand and learn, I have relied on the Haskell representation `Codec.CBOR.Term.Term` to construct the transaction. The format itself is defined in [alonzo.cddl](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/alonzo/test-suite/cddl-files/alonzo.cddl).

  [flat]: https://hackage.haskell.org/package/flat

### Issue Number

ADP-1221, ADP-1234


piotr-iohk:
Got this running locally:
```
Failures:

  src/Test/Integration/Plutus.hs:423:29: 
  1) API Specifications, NEW_SHELLEY_TRANSACTIONS, Plutus scenarios, currency
       uncaught exception: ErrorCall
       Integration test for currency contract can only handle TxInput indices 0-31. Sorry.
         error, called at src/Test/Integration/Plutus.hs:423:29 in cardano-wallet-core-integration-2021.11.11-GK96ACHnYb8G230fohhqiN:Test.Integration.Plutus

  To rerun use: --match "/API Specifications/NEW_SHELLEY_TRANSACTIONS/Plutus scenarios/currency/"

Randomized with seed 350012249

```

HeinrichApfelmus:
>   1) API Specifications, NEW_SHELLEY_TRANSACTIONS, Plutus scenarios, currency
>        uncaught exception: ErrorCall
>        Integration test for currency contract can only handle TxInput indices 0-31. Sorry.

🙈 Ouch. I was hoping that in practice, unspent transaction outputs would be limited to have an index in the range 0–31. 😅 The reason for this restriction is that I have to bake the index into the Plutus contract, but this is not as simple as replacing a couple of hex bytes with new bytes — the index is stored in individual bits, and might even be variable-length encoded. 🙈 I'll try to increase the range.

lib/core-integration/src/Test/Integration/Plutus.hs
```diff
@@ -296,9 +315,122 @@ withdrawScript_1 =
...
+--  @/transactions-balance@ endpoint of the wallet in order
+-- to add change outputs and further inputs to cover fees if necessary.
+-- This endpoint also needs to make sure that the minted assets
+-- are assigned to a change output.
+currencyTx
+    :: forall n. EncodeAddress n
+    => ApiWalletInput n
+    -- ^ UTxO that is hard-wired into the smart contract.
+    -> Aeson.Value
+currencyTx input = Aeson.object
```
piotr-iohk, line 342: This shows all the bits of the transaction very nicely!

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -1802,6 +1806,24 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
...
+                            (_, result) <- selectCoins @_ @'Shelley ctx w $
+                                singleton $ AddressAmount
+                                    { address = (addr, proxy)
+                                    , amount  = Quantity 10_000_000
+                                    , assets  = ApiT TokenMap.empty
+                                    }
+                            pure $ head . view #inputs <$> result
+                    txOutRef <- fromEither =<< getFreshUTxO
+                    let mint = PlutusScenario.currencyTx txOutRef
+                    pure (mint, [])
```
piotr-iohk, line 1825: Curretly all Plutus tx tests are wired to the same "Plutus scenarios" sub-suite which is neat but doesn't give opportunity for more sophisticated assertions. Like, for _Currency_ test, we could check that "apfel" and "banana" are really minted and on wallet balance after the transaction. What do you think?

I am trying to add more assertions though when adjusting contract for e2e tests (e.g. [here](https://github.com/input-output-hk/cardano-wallet/blob/master/test/e2e/spec/e2e_spec.rb#L122-L159)) so maybe that's enough.

HeinrichApfelmus: > Like, for Currency test, we could check that "apfel" and "banana" are really minted and on wallet balance after the transaction. What do you think?

I was wondering the same thing. To be honest, I think that your e2e tests look a lot cleaner than the Haskell version. 😅 I'm not sure exactly why, but I believe the reason is that the Haskell version has a hard time constructing and deconstructing JSON values.

If you ask me, I would rather spend the time on improving the integration test DSL as a whole rather than trying to improve this individual test.

In this case, the smart contract checks that the assets are minted; thus, as long as the smart contract is touched by the node, we can be certain that these values are minted, as the node would reject the transaction otherwise.

piotr-iohk (approved):
:+1:

HeinrichApfelmus:
bors merge

HeinrichApfelmus:
bors retry

PR #3037, opened 2021-11-25
Fix wallet state transition functions to properly account for script validation failures.

jonathanknowles:
## Issue Number

ADP-1039

## Summary

This PR:

- breaks up wallet state transition functions into smaller functions for easier testing.
- adds function `isOurTx`, which returns `True` if and only if a transaction is relevant to the wallet.
- adjusts the withdrawal sum computation within `applyOurTxToUTxO` to correctly account for script validation failure.
- adds a property test for `applyOurTxToUTxO` that relates it to `isOurTx` and `applyTxToUTxO`.
- fixes the termination logic of `applyOurTxToUTxO` so that it returns a result only when appropriate.


sevanspowell (approved):
Great! I'm glad to have more coverage of this. I hope we can have more still.

PR #3023, opened 2021-11-16
Fix plutus

ghost:
Honestly I struggle to properly name the PR. It brings the plutus related updates from @j-mueller's fork. I guess important for plutus-pab and cardano-wallet integration.

PR #3038, opened 2021-11-25
Update dependencies

kk-hainq:
This is more or less a proof-of-concept to bump the dependencies of this repository. There have been several breaking changes, mainly from `cardano-ledger` and `cardano-node` so this is non-trivial with several undecided `fixme`s here and there. I've also been able to clean many Nix files by removing unused Plutus subdirs. I suspect we can still trim a lot more across the board.

Context: `cardano-wallet` is now a dependency of `plutus-apps`, making these bumps useful for dApp developers who aggressively bump `plutus-apps` themselves (for new PAB commits, node fixes, etc). While this is not critical and we have no preferences on how to handle it, both #3023 and this would be very useful to dApp developers. We are willing to contribute in any way we can as well.

### Jira Issue

ADP-1266


rvl:
Thanks @kk-hainq - we were going to wait until [cardano-node 1.32.0](https://github.com/input-output-hk/cardano-node/releases) before updating. But please let us know if it's ready to review.

kk-hainq:
@rvl Thank you for the confirmation. This is just a sample draft from our fork to proceed for now. I think the actual update would need to be several more modular PRs. I guess something like #3023 would be a better start.

In the meantime, we could also open a no-op PR that removes unused dependencies/sub dirs/Nix files if you would accept one?

kk-hainq:
Closed in favour of #3050 (thank you!!).

PR #3043, opened 2021-12-01
Change aux dir to extra to fix git checkout on Windows

piotr-iohk:
- [x] Changed dir name `aux` to `extra` as `aux` is not ok for git on Windows

### Comments

https://confluence.atlassian.com/bitbucketserverkb/error-invalid-path-during-git-clone-to-windows-client-1085186345.html

Apparently `aux` is one of reserved file names on Windows and git is chocking on it. This is problem for e2e tests on Windows, like [here](https://github.com/input-output-hk/cardano-wallet/runs/4372713205?check_suite_focus=true).

Changing this dir name to `extra` seems to fix the problem -> [here](https://github.com/input-output-hk/cardano-wallet/actions/runs/1524776586).

### Issue Number

n/a


HeinrichApfelmus (approved):
😳🤣 Windows. 🪟🙄

PR #2008, opened 2020-08-06
Provide PoolId to the SMASH url builder

KtorZ:
# Issue Number

# Overview

- e91e6e04bf96d9fd3af5b2077097127c87e2a761
  :round_pushpin: **Provide PoolId to the SMASH url builder**
    SMASH just undergone a breaking change in its API, going from:

  ```
  GET api/v1/metadata/{hash}
  ```

  to

  ```
  GET api/v1/metadata/{poolId}/{hash}
  ```

# Comments

Manually tested on the mainnet candidate and a local instance of SMASH latest master. Providing a wrong URL:

```
[cardano-wallet.pools-engine:Info:135] [2020-08-06 15:03:01.05 UTC] Fetching metadata with hash 67b52f96 from http://localhost:8080/api/v1/7b6dbe3b5c78af9efba3c283d20dfb075c6a6d61d919d2c7d6610576/67b52f96eb40fcc18fa9cd8b9d7dd620811756f1029411d105de2eaf79655fa4
[cardano-wallet.pools-engine:Warning:135] [2020-08-06 15:03:01.05 UTC] Failed to fetch metadata with hash 67b52f96: There's no known metadata for this pool.
[cardano-wallet.pools-engine:Info:135] [2020-08-06 15:03:01.05 UTC] Fetching metadata with hash 311522eb from http://localhost:8080/api/v1/5393eb32b547a34f4946927fbda172b6908cc75f592b2f6151ea31dc/311522eb2b41ada16f33dab31b9138546d0e2cc134cfb73a67207f9e5709bf3c
[cardano-wallet.pools-engine:Warning:135] [2020-08-06 15:03:01.05 UTC] Failed to fetch metadata with hash 311522eb: There's no known metadata for this pool.
[cardano-wallet.pools-engine:Info:135] [2020-08-06 15:03:01.05 UTC] Fetching metadata with hash 81405d32 from http://localhost:8080/api/v1/3feb43c4f05ab53ca5426bb66d38a76a1b5d878d8a8d3e14726a01ce/81405d326af6d1b6f06d18bebf1189e66b5e7312a1d4e0e4bbb2fbb160328388
[cardano-wallet.pools-engine:Warning:135] [2020-08-06 15:03:01.05 UTC] Failed to fetch metadata with hash 81405d32: There's no known metadata for this pool.
```

And providing a correct URL:

```
[...]

PR #3046, opened 2021-12-02
Reorganize modules in Cardano.Wallet.DB.*

HeinrichApfelmus:
### Issue number

ADP-1288

### Overview

In this step, we reorganize the modules in the `Cardano.Wallet.DB.*` hierarchy; the aim is to be able to quickly swap out the implementation of the `mkStoreWalletsCheckpoints` function by changing one import. We want to keep both the previous implementation and a subsequent new implementation around for some time.

### Details

* The migration function `migrateManually` is moved to a separate module `Cardano.Wallet.DB.Sqlite.Migration`. This has the beneficial side effect that this migration function cannot be accidentally be affected by refactoring in modules that are not imported by this module.
* The creation of the checkpoints store, `mkStoreWalletsCheckpoints`, is moved to a separate module. We can swap out its implementation by importing a different module in the future.

### Comments

Merge PR #2942 before this one, because this pull request is based on the branch of the former.

### DIFF
--- a/lib/core/src/Cardano/Wallet/DB/Checkpoints.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Checkpoints.hs
@@ -1,0 +1,141 @@
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE OverloadedLabels #-}
+{-# LANGUAGE TypeFamilies #-}
+
+-- |
+-- Copyright: © 2021 IOHK
+-- License: Apache-2.0
+--
+-- Data type that represents a collection of checkpoints.
+-- Each checkpoints is associated with a 'Slot'.
+
+module Cardano.Wallet.DB.Checkpoints
+    ( getPoint
+
+    -- * Checkpoints  
+    , Checkpoints (..)
+    , singleton
+    , getLatest
+    , findNearestPoint
+    
+    -- * Delta types
+    , DeltaCheckpoints (..)
+    , DeltaMap (..)
+    ) where
+
+import Prelude
+
+import Data.Delta
+    ( Delta (..) )
+import Data.Generics.Internal.VL.Lens
+    ( over, view )
+import Data.Map.Strict
+    ( Map )
+import Data.Maybe
+    ( fromMaybe )
+import GHC.Generics
+    ( Generic )
+
+import qualified Cardano.Wallet.Primitive.Model as W
+import qualified Cardano.Wallet.Primitive.Types as W
+import qualified Data.Map.Strict as Map
+import qualified Data.Set as Set
+
+{- NOTE [PointSlotNo]
+
+'SlotNo' cannot represent the genesis point.
+
+Historical hack. The DB layer can't represent 'Origin' in the database,
+instead we have mapped it to 'SlotNo 0', which is wrong.
+
+Rolling back to SlotNo 0 instead of Origin is fine for followers starting
+from genesis (which should be the majority of cases). Other, non-trivial
+rollbacks to genesis cannot occur on mainnet (genesis is years within
+stable part, and there were no rollbacks in byron).
+
+Could possibly be problematic in the beginning of a testnet without a
+byron era. /Perhaps/ this is what is happening in the
+>>> [cardano-wallet.pools-engine:Error:1293] [2020-11-24 10:02:04.00 UTC]
+>>> Couldn't store production for given block before it conflicts with
+>>> another block. Conflicting block header is:
+>>> 5bde7e7b<-[f1b35b98-4290#2008]
+errors observed in the integration tests.
+
+The issue has been partially fixed in that 'rollbackTo' now takes
+a 'Slot' as argument, which can represent the 'Origin'.
+However, the database itself mostly stores slot numbers.
+
+FIXME LATER during ADP-1043: As we move towards in-memory data,
+all slot numbers in the DB file will either be replaced by
+the 'Slot' type, or handled slightly differently when it
+is clear that the data cannot exist at the genesis point
+(e.g. for TxHistory).
+
+-}
+
+-- | Helper function: Get the 'Point' of a wallet state.
+getPoint :: W.Wallet s -> W.Slot
+getPoint =
+    W.toSlot . W.chainPointFromBlockHeader . view #currentTip
+
+{-------------------------------------------------------------------------------
+    Checkpoints
+-------------------------------------------------------------------------------}
+{- HLINT ignore Checkpoints "Use newtype instead of data" -}
+-- | Collection of checkpoints indexed by 'Slot'.
+data Checkpoints a = Checkpoints
+    { checkpoints :: Map W.Slot a
+    } deriving (Eq,Show,Generic)
+-- FIXME LATER during ADP-1043:
+--  Use a more sophisticated 'Checkpoints' type that stores deltas.
+
+-- | Make a single checkpoint.
+singleton :: W.Slot -> a -> Checkpoints a
+singleton key a = Checkpoints $ Map.singleton key a
+
+-- | Get the checkpoint with the largest 'SlotNo'.
+getLatest :: Checkpoints a -> (W.Slot, a)
+getLatest = from . Map.lookupMax . view #checkpoints 
+  where
+    from = fromMaybe (error "getLatest: Genesis checkpoint is missing!")
+    -- FIXME LATER during ADP-1043:
+    --   Make sure that 'Checkpoints' always has a genesis checkpoint
+
+-- | Find the nearest 'Checkpoint' that is either at the given point or before.
+findNearestPoint :: Checkpoints a -> W.Slot -> Maybe W.Slot
+findNearestPoint m key = fst <$> Map.lookupLE key (view #checkpoints m)
+
+{-------------------------------------------------------------------------------
+    Delta type for Checkpoints
+-------------------------------------------------------------------------------}
+data DeltaCheckpoints a
+    = PutCheckpoint W.Slot a
+    | RollbackTo W.Slot
+        -- Rolls back to the latest checkpoint at or before this slot.
+    | RestrictTo [W.Slot]
+        -- ^ Restrict to the intersection of this list with
+        -- the checkpoints that are already present.
+        -- The genesis checkpoint will always be present.
+
+instance Delta (DeltaCheckpoints a) where
+    type Base (DeltaCheckpoints a) = Checkpoints a
+    apply (PutCheckpoint pt a) = over #checkpoints $ Map.insert pt a
+    apply (RollbackTo pt) = over #checkpoints $
+        Map.filterWithKey (\k _ -> k <= pt)
+    apply (RestrictTo pts) = over #checkpoints $ \m ->
+        Map.restrictKeys m $ Set.fromList pts
+
+{-------------------------------------------------------------------------------
+    A Delta type for Maps
+-------------------------------------------------------------------------------}
+-- | Delta type for 'Map'.
+data DeltaMap key da
+    = Insert key (Base da)
+    | Delete key
+    | Adjust key da
+
+instance (Ord key, Delta da) => Delta (DeltaMap key da) where
+    type Base (DeltaMap key da) = Map key (Base da)
+    apply (Insert key a) = Map.insert key a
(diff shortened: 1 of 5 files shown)


lib/core/src/Cardano/Wallet/DB/Checkpoints.hs
```diff
@@ -0,0 +1,141 @@
...
+-- FIXME LATER during ADP-1043:
+--  Use a more sophisticated 'Checkpoints' type that stores deltas.
+
+-- | Make a single checkpoint.
+singleton :: W.SlotPoint -> a -> Checkpoints a
+singleton key a = Checkpoints $ Map.singleton key a
+
+-- | Get the checkpoint with the largest 'SlotNo'.
+getLatest :: Checkpoints a -> (W.SlotPoint, a)
+getLatest = from . Map.lookupMax . view #checkpoints
```
paweljakubas, line 98: why not to use Either here with explicit Error. use could then also use `maybeToLeft` https://hackage.haskell.org/package/either-5.0.1.1/docs/src/Data.Either.Combinators.html#maybeToLeft

lib/core/src/Cardano/Wallet/DB/Checkpoints.hs
```diff
@@ -0,0 +1,141 @@
...
+the 'SlotPoint' type, or handled slightly differently when it
+is clear that the data cannot exist at the genesis point
+(e.g. for TxHistory).
+
+-}
+
+-- | Helper function: Get the 'Point' of a wallet state.
+getPoint :: W.Wallet s -> W.SlotPoint
+getPoint =
+    W.toSlotPoint . W.chainPointFromBlockHeader . view #currentTip
```
paweljakubas: what is returned when not even genesis block header? :thinking:

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1289,87 +521,63 @@ newDBLayerWith
...
+newDBLayerWith _cacheBehavior tr ti SqliteContext{runQuery} = do
+    -- FIXME LATER during ADP-1043:
+    --   Remove the 'NoCache' behavior, we cannot get it back.
+    --   This will affect read benchmarks, they will need to benchmark
+    --   'loadDBVar' instead.
 
-    cacheVar <- newMVar Map.empty :: IO (MVar (Map W.WalletId (W.Wallet s)))
+    -- FIXME LATER during ADP-1043:
+    --   Handle the case where loading the database fails.
+    checkpointsVar <- runQuery $ loadDBVar mkStoreWalletsCheckpoints
```
paweljakubas: :+1:

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -778,27 +790,67 @@ instance Buildable (Block) where
...
+        :: Quantity "block" Word32
+    , headerHash
+        :: !(Hash "BlockHeader")
+    , parentHeaderHash
+        :: !(Hash "BlockHeader")
+    } deriving (Show, Eq, Ord, Generic)
+
+-- | Magic value denoting the hash of a block that \"does not exist\".
+-- Specifically, this is the hash of \"the parent block\" of the genesis block.
+hashOfNoParent :: Hash "BlockHeader"
```
paweljakubas: another approach could be to use introduce something like "parent" with  Maybe...and in that case no need for magic values...it would be just Nothing

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -778,27 +790,67 @@ instance Buildable (Block) where
...
         <> if null txs then " ∅" else "\n" <> indentF 4 (blockListF txs)
 
+data BlockHeader = BlockHeader
+    { slotNo
+        :: SlotNo
+    , blockHeight
+        :: Quantity "block" Word32
+    , headerHash
+        :: !(Hash "BlockHeader")
+    , parentHeaderHash
```
paweljakubas: why not to have here `!(Maybe (Hash "blockHeader"))` ?

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -778,27 +790,67 @@ instance Buildable (Block) where
...
+    } deriving (Show, Eq, Ord, Generic)
+
+-- | Magic value denoting the hash of a block that \"does not exist\".
+-- Specifically, this is the hash of \"the parent block\" of the genesis block.
+hashOfNoParent :: Hash "BlockHeader"
+hashOfNoParent = Hash . BS.pack $ replicate 32 0
+
+-- | Check whether a block with a given 'BlockHeader' is the genesis block.
+isGenesisBlockHeader :: BlockHeader -> Bool
+isGenesisBlockHeader = (== hashOfNoParent) . view #parentHeaderHash
```
paweljakubas: then here we would have `isNothing` rather than `hashOfNoParent` ;-)

lib/core/src/Cardano/Wallet/Primitive/Types.hs
```diff
@@ -811,31 +863,30 @@ instance Buildable ChainPoint where
...
+-- to a point on an actual chain with valid block hashes,
+-- but merely to a timeslot which can hold a single block.
+-- This implies:
+--
+-- * 'SlotPoint' has a linear ordering implemented in the 'Ord' class
+--   (where @Origin < At slot@).
+-- * Using 'SlotPoint' in QuickCheck testing requires less context
+-- (such as an actual simulated chain.)
+--
+-- TODO: Better name, perhaps simply 'Slot'?
```
paweljakubas: yeah, `Slot` is exactly what it is ;-)

paweljakubas (approved):
I like reorganizations adopted. It is seen what is new, what is old, and also that we don't affect with Checkpoints migration code at all now.

I like Checkpoint step 1 refactoring. It makes things easier to understand. To my best knowledge logic of checkpoint lifecycle is maintained (which is further evidenced by all current tests passing). I would think about introducing Maybe in parentBlockHeader and hence have Nothing rather than magic value for first block header. 

Good job! Waiting for another iteration!

Also would be good if other team member would review this PR (for both double check and educational/knowledge sharing purposes)

lib/core/src/Cardano/Wallet/DB/Checkpoints.hs
HeinrichApfelmus: Hm. This error case would be a programming error, as we assume that a checkpoint for genesis is always stored, so I thought it would be fine to use a partial function here. 🤔

But I can see that if we consider `Checkpoints` as an abstraction independent of the database, then the error might in principle occur, e.g. we have

```
getLatest (apply (RollbackTo pt2) (singleton pt1 a)) = _|_
```

I think that the appropriate solution is actually to change `singleton` to only accept the genesis slot (and rename the function to `fromGenesis`) and to comment that `RestricTo` will never remove the genesis checkpoint. Then, the `Checkpoints` data type will satisfy the invariant that the map is always nonempty.

I will add a change to this effect.

lib/core/src/Cardano/Wallet/DB/Checkpoints.hs
HeinrichApfelmus: I don't understand, what do you mean? If the current tip is not the genesis block header, then `At (Slot slot)` will be returned, where `slot` is the purported slot number recorded in the `BlockHeader`.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
HeinrichApfelmus: Hm. 🤔  I agree — but unfortunately, the database does serialize the `parentHeaderHash` in the table:

https://github.com/input-output-hk/cardano-wallet/blob/3a24c14b3208da7a6e890de1c2f3f261ab8d5b01/lib/core/src/Cardano/Wallet/DB/Sqlite/TH.hs#L210

and then we would have the magic value in the serialization. But I suppose that is an improvement over having the magic value in the type.

lib/core/src/Cardano/Wallet/Primitive/Types.hs
HeinrichApfelmus: 👍 I will rename it to `Slot`. 😄

lib/core/src/Cardano/Wallet/DB/Checkpoints.hs
paweljakubas: yes, yes

lib/core/src/Cardano/Wallet/Primitive/Types.hs
paweljakubas: yes, we could probably deal with this magic value at the level of `PersistField` instance or so. and everywhere above we could have Maybe. just suggestion

paweljakubas: so have magic value defined somewhere here https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/DB/Sqlite/Types.hs and `toPersistValue Nothing = hashOfNoParent`

HeinrichApfelmus: I have added the `hashOfNoParent` constant to this module, but stopped short of changing the `BlockId` type. Instead, I have created two conversion functions.

lib/core/src/Cardano/Wallet/DB/Checkpoints.hs
HeinrichApfelmus: I'm beginning to confuse myself by doing a lot rebasing, so I would like to postpone the introduction of `fromGenesis` to the next PR in this series.

lib/core/src/Cardano/Wallet/DB/Checkpoints.hs
```diff
@@ -0,0 +1,141 @@
...
+import qualified Cardano.Wallet.Primitive.Types as W
+import qualified Data.Map.Strict as Map
+import qualified Data.Set as Set
+
+{- NOTE [PointSlotNo]
+
+'SlotNo' cannot represent the genesis point.
+
+Historical hack. The DB layer can't represent 'Origin' in the database,
+instead we have mapped it to 'SlotNo 0', which is wrong.
```
rvl, line 49: It shouldn't matter - as you write below, we don't need rollback to genesis. Only the unstable blocks can be rolled back.

lib/core/src/Cardano/Wallet/DB/Checkpoints.hs
```diff
@@ -0,0 +1,141 @@
...
+from genesis (which should be the majority of cases). Other, non-trivial
+rollbacks to genesis cannot occur on mainnet (genesis is years within
+stable part, and there were no rollbacks in byron).
+
+Could possibly be problematic in the beginning of a testnet without a
+byron era. /Perhaps/ this is what is happening in the
+>>> [cardano-wallet.pools-engine:Error:1293] [2020-11-24 10:02:04.00 UTC]
+>>> Couldn't store production for given block before it conflicts with
+>>> another block. Conflicting block header is:
+>>> 5bde7e7b<-[f1b35b98-4290#2008]
```
rvl, line 61: To me, this looks more like rollback isn't quite working correctly in the pools db.
The `#2008` means block height 2008, which is a long way from genesis.

Anviking: Hm. Good point.

HeinrichApfelmus:
bors merge

lib/core/src/Cardano/Wallet/DB/Checkpoints.hs
HeinrichApfelmus: We did receive reports were #0 is affected:

> Unexpected error following the chain:user error (restoreBlocks: given chain isn't a valid continuation. Wallet is at: `891fcd4b<-[9c9290a4-0#0]` but the given chain continues starting from: `891fcd4b<-[db619f76-14#0]`)

Full log: https://gist.github.com/runeksvendsen/c143926249e32b22ee2d184581c9776c

Anviking: Yes. But the pool db problem could be unrelated then.

PR #3047, opened 2021-12-02
New submit tx endpoint

paweljakubas:
adp-1171

- [x] swagger extended
- [x] added impl scaffolding
- [x] added submitTransaction link
- [x] add first impl for submitTransaction
- [x] add integration testing  
- [x] add impl for submitTransaction using decodeTransaction

### Comments

This PR depends on https://github.com/input-output-hk/cardano-wallet/pull/3028 in a sense that should be merged after


lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -2275,16 +2276,16 @@ signTx ctx w sealedTx expectations = do
...
+    pure $ getFromResponse Prelude.id r
 
 submitTx
     :: MonadUnliftIO m
     => Context
-    -> ApiT SealedTx
+    -> ApiSerialisedTransaction
     -> [(HTTP.Status, Either RequestException ApiTxId) -> m ()]
     -> m ApiTxId
 submitTx ctx tx expectations = do
```
piotr-iohk, line 2287:
```suggestion
submitExternalTx ctx tx expectations = do
```

lib/core-integration/src/Test/Integration/Framework/DSL.hs
```diff
@@ -2294,6 +2295,21 @@ submitTx ctx tx expectations = do
     verify r expectations
     pure $ getFromResponse Prelude.id  r
 
+submitTxWithWid
```
piotr-iohk, line 2298:
```suggestion
submitTx
```

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -275,14 +277,14 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
...
-        case getMetadata (cardanoTx $ getApiT signedTx) of
+        case getMetadata (cardanoTx $ getApiT (signedTx ^. #transaction)) of
             Nothing -> error "Tx doesn't include metadata"
             Just m  -> case Map.lookup 1 m of
                 Nothing -> error "Tx doesn't include metadata"
                 Just (Cardano.TxMetaText "hello") -> pure ()
                 Just _ -> error "Tx metadata incorrect"
 
-        let txCbor = getFromResponse #transaction (HTTP.status202, Right $ ApiSerialisedTransaction signedTx)
+        let txCbor = getFromResponse #transaction (HTTP.status202, Right signedTx)
```
piotr-iohk, line 290:
```suggestion
        let txCbor = signedTx ^. #transaction 
```

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -766,6 +768,76 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
...
+
+        let queryTx = Link.getTransaction @'Shelley wa (ApiTxId txid)
+        rGetTx <- request @(ApiTransaction n) ctx queryTx Default Empty
+        verify rGetTx
+            [ expectResponseCode HTTP.status200
+            , expectField
+                (#amount . #getQuantity)
+                (`shouldBe` initialAmt - (amt + fromIntegral expectedFee))
+            , expectField (#direction . #getApiT) (`shouldBe` Outgoing)
+            , expectField (#status . #getApiT) (`shouldBe` Pending)
```
piotr-iohk, line 837: I wonder if this won't cause race conditions in local cluster :thinking: 
Once the feature is merged I will update e2e tests as well. Asserting for `pending` there will be more "stable" I guess.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -324,7 +326,7 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
 
         signedTx <- signTx ctx wa apiTx [ expectResponseCode HTTP.status202 ]
 
-        let txCbor = getFromResponse #transaction (HTTP.status202, Right $ ApiSerialisedTransaction signedTx)
+        let txCbor = getFromResponse #transaction (HTTP.status202, Right signedTx)
```
piotr-iohk, line 336:
```suggestion
        let txCbor = signedTx ^. #transaction 
```

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -370,7 +372,7 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
...
```
piotr-iohk, line 386:
```suggestion
        let txCbor = signedTx ^. #transaction 
```

piotr-iohk:
@paweljakubas I found 1 issue/corner case playing on the branch. (Planning to add an integration test for that in https://github.com/input-output-hk/cardano-wallet/pull/3079).

To reproduce:
 1. construct and sign transaction using wallet A
 2. submit it from wallet B

Expected:
I'd expect an error message from the API saying that the transaction does not belong to this wallet

Actual:
Transaction is submitted successfully but it is constantly in `pending` until it goes to `expired`.

paweljakubas:
@piotr-iohk  Well, this behavior is expected in a sense that it is possible for any wallet to send any tx. And wallet B should have here pending->expired as it does not affect it, wallet A should experience balance change after the tx is sent via wallet B. 
Nevertheless, we can add check and disallow it. So it will be not possible if in tx there is no inputs/outputs/withdrawals belonging to a given wallet. We have this information as we use decodeTransaction inside. @Anviking are you in favor of this check? I am inclined to agree we should have it as Piotr suggested.

piotr-iohk:
Yes, this may be misleading/confusing if someone makes a mistake submitting tx from another wallet than it originates from, therefore I'm for forbidding it (with proper error message).

Anviking:
>  @Anviking are you in favor of this check? I am inclined to agree we should have it as Piotr suggested.

> Transaction is submitted successfully but it is constantly in pending until it goes to expired.

I'd agree. So If I understand correctly, a transaction can be submitted using a wallet to which it is foreign, and the tx may make it on chain, **but** the wallet will never notice? This seems like a big potential source of confusion, so seems better to forbid it.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -1938,7 +2070,10 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
             request @(ApiConstructTransaction n) ctx constructEndpoint Default payload
 
         -- Submit tx
-        void $ submitTx ctx sealedTx [ expectResponseCode HTTP.status500 ]
+        submittedTx <- submitTxWithWid ctx w (ApiSerialisedTransaction sealedTx)
+        verify submittedTx
+            [ expectResponseCode HTTP.status500
```
piotr-iohk: I'd advocate also for fixing this here. Submitting unsigned transaction should present proper error from wallet API, not `500` from the node.

Anviking:
> We have this information as we use decodeTransaction inside.

I came to think about, if guaranteeing things like `all txs that can be submitted to be pending, can also transition to in_ledger` are desirable, perhaps we could re-use the discovery logic from `Primitive.Model`. 

https://github.com/input-output-hk/cardano-wallet/blob/25ef435c6e6a30d9d8ce3e454ee20845bdc4ae10/lib/core/src/Cardano/Wallet/Primitive/Model.hs#L598

However I don't feel strongly about it. Possibly in the realm of multiple wallet instances and inter-dependent txs, your approach might be more reliable (it would discover txs which only spend inputs which are not _yet_ in the wallet utxo) 🤷‍♂️ 

@paweljakubas what do you think about a property test along the lines of the above? (`all txs that can be submitted to be pending, can also transition to in_ledger`) Possibly a bit tricky, but also possibly very good to have.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
paweljakubas: sounds reasonable

Anviking:
Maybe something like this (just scaffold code)
```haskell
-- | 'submitTransaction' should accept exactly those 
-- transactions which are discoverable on-chain.
-- 
-- If it were to accept a tx which cannot be discovered, we would not
-- find out when it transitions from `pending` to `in_ledger`.
-- 
-- If it were to reject a tx which can be discovered, we 
-- probably should have accepted it.
prop_submitTransactionAcceptsDiscoverableTxs tx w0 = do
    let (res, w1) = submitTransaction tx
    case res of
      Right _ -> label "success" $ do
        listTransactions w1 `containsPending` tx
        let w2 = applyBlock (mockBlock [tx]) w1
        listTransactions w2 `containsInLedger` tx
      Left  _ -> label "failure (tx unrelated to wallet)" $ do
        listTransactions w1 === []
        let w2 = applyBlock (mockBlock [tx]) w1
        listTransactions w2 === []
```

The above assumes pure `submitTransaction` and `listTransaction` which is unrealistic atm. But with `monadicIO`, a no-op dummy `NetworkLayer` and either Sqlite or MVar DBLayer... maybe something like that could work...

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2330,6 +2335,88 @@ decodeTransaction ctx (ApiT wid) (ApiSerialisedTransaction (ApiT sealed)) = do
...
+    when (isForeign apiDecoded) $
+        liftHandler $ throwE ErrSubmitTransactionForeignWallet
+    let ourOuts = getOurOuts apiDecoded
+    let ourInps = getOurInps apiDecoded
+
+    let allInpsNum = length $ apiDecoded ^. #inputs
+    let witsNum = length $ getSealedTxWitnesses sealedTx
+    when (allInpsNum > witsNum) $
+        liftHandler $ throwE $
+        ErrSubmitTransactionPartiallySignedOrNoSignedTx allInpsNum witsNum
```
Anviking, line 2363: I think this risks accidentally blocking txs with script inputs from being submitted.

`getSealedTxWitnesses` returns the key-witnesses. But what if we have 1 script inputs, 1 wallet input, and one key witness? Then the tx should be valid, but will fail here?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2330,6 +2335,88 @@ decodeTransaction ctx (ApiT wid) (ApiSerialisedTransaction (ApiT sealed)) = do
...
+        let generalOuts = apiDecodedTx ^. #outputs
+        in map toTxOut $ filter isOutOurs generalOuts
+
+    getOurWdrl rewardAcct path apiDecodedTx =
+        let generalWdrls = apiDecodedTx ^. #withdrawals
+            isWdrlOurs (ApiWithdrawalGeneral _ _ context) = context == Our
+        in case filter isWdrlOurs generalWdrls of
+            [ApiWithdrawalGeneral (ApiT acct, _) (Quantity amt) _] ->
+                let acct' = invariant "reward account should be the same" acct (rewardAcct ==)
+                in WithdrawalSelf acct' path (Coin $ fromIntegral amt)
```
```suggestion
                in WithdrawalSelf acct' path (Coin amt)
```
`Natural -> Natural` conversion unneeded

specifications/api/swagger.yaml
```diff
@@ -5888,6 +5923,23 @@ paths:
...
       responses: *responsesDecodedTransaction
 
+  /wallets/{walletId}/transactions-submit:
+    post:
+      operationId: submitTransaction
+      tags: ["Transactions New"]
+      summary: Submit
+      description: |
+        <p align="right">status: <strong>stable</strong></p>
+        Submit a transaction that was already created and signed.
```
Anviking, line 5933:
```suggestion
        Submit a transaction that was already created and signed. Fails for foreign transactions.
        
        If the submission to the local node was successful, the wallet will store the pending transaction and retry until confirmed or expired.
```

Maybe something like this to mention:
1. this endpoint will track the transaction as pending
2. that it will fail if the transaction is foreign (without duplication the definition of foreign)

specifications/api/swagger.yaml
```diff
@@ -4213,6 +4218,28 @@ x-errCreatedInvalidTransaction: &errCreatedInvalidTransaction
...
+x-errMissingWitnessesInTransaction: &errMissingWitnessesInTransaction
+  <<: *responsesErr
+  title: missing_witnesses_in_transaction
+  properties:
+    message:
+      type: string
+      description: Occurs when a transaction to be submitted is not fully signed.
+    code:
+      type: string
+      enum: ['missing_witnesses_in_transaction']
```
Anviking, line 4241: Explicitly defining the relevant errors in swagger sound like a good idea.

But I'd vote for
1. Removing this now. If the node finds anything wrong with the tx, it will be included in the error response (just not in a swagger defined format).
2. As a new U/S, map [Cardano.TxValidationErrorInMode CardanoMode](https://github.com/input-output-hk/cardano-wallet/blob/987693fc2eeb3b71ea20bbc2568050d5c295fc31/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs#L424) to wallet-types with swagger schemas.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2330,6 +2335,88 @@ decodeTransaction ctx (ApiT wid) (ApiSerialisedTransaction (ApiT sealed)) = do
...
+        , IsOwned s k
+        , Typeable s
+        , Typeable n
+        )
+    => ctx
+    -> ApiT WalletId
+    -> ApiSerialisedTransaction
+    -> Handler ApiTxId
+submitTransaction ctx apiw@(ApiT wid) apitx@(ApiSerialisedTransaction (ApiT sealedTx)) = do
+    ttl <- liftIO $ W.getTxExpiry ti Nothing
```
Anviking, line 2352: This is the wrong `ttl`. It will be derived from the current time rather than the tx.

Anviking (commented):
👍 

I think I found two issues that should be fixed before merging (see comments). If you agree I think the solutions would be to:  
1. Drop the witness check for now
2. Extract the actual `ttl` from the tx in the TransactionLayer`decodeTx`, to use in the TxMeta when submitting

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: Probably we need this:

> Extract the actual ttl from the tx in the TransactionLayerdecodeTx, to use in the TxMeta when submitting

Anviking: Cardano.TxBodyContent contains:
```
    txValidityRange  :: (TxValidityLowerBound era, TxValidityUpperBound era),
```
<s>But I'm not sure what we do with the lower bound, or if the upper bound doesn't exist 🙃 </s>

<s>We might need DB changes to — at least — allow `TxMeta` to have no TTL…</s>

We _can_ have a `Nothing` `expiry` https://github.com/input-output-hk/cardano-wallet/blob/987693fc2eeb3b71ea20bbc2568050d5c295fc31/lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs#L403

so converting the upper validity bound should go fine. The lower bound we _could_ choose to not support/show at the moment.

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: As we talked yesterday, at this moment we will recommend using submit-without-wid endpoint for anything outside integration tests (so "normal" txs at this moment). We will address all PAB / multisig cases as they appear. Right now all PAB integration tests pass with new endpoint. Here, we are just trying to say that when we have a specified number of "normal" inputs then sending tx that is partially signed or not signed at all does not make sense. I think it is reasonable FIRST approximation. The next moves will be:(a) deal with collaterals, (b) deal with script inputs (here sometimes as you say several wits will be required -> we will need to have a way to decode that first then enhance condition). Please notice that we are proceeded step by step and until we have not whole new tx workflow finished old endpoints and proxy submit is recommended to use.

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: nice catch -> corrected

specifications/api/swagger.yaml
paweljakubas: corrected

specifications/api/swagger.yaml
paweljakubas: expressed my opinion above

lib/core/src/Cardano/Wallet/Api/Server.hs
paweljakubas: yes, it should be as you write. BUT this is the subject of ADP-1193 - I will add TODO to remember that though

lib/core/src/Cardano/Wallet/Api/Server.hs
piotr-iohk: >As we talked yesterday, at this moment we will recommend using submit-without-wid endpoint for anything outside integration tests (so "normal" txs at this moment). 

My understanding was that we rather want to recommend submit-with-wid moving forward, especially in the construct-sign-submit flow? However if this particular issue is too big to address here, then I'd suggest capturing it in the Jira ticket? I understand that it would only affect balance-sign-submit flow?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2330,6 +2335,89 @@ decodeTransaction ctx (ApiT wid) (ApiSerialisedTransaction (ApiT sealed)) = do
...
+                NoWithdrawal
+
+    isInpOurs (WalletInput _) = True
+    isInpOurs _ = False
+    toTxInp (WalletInput (ApiWalletInput (ApiT txid) ix _ _ _ _)) =
+        (TxIn txid ix, Coin 0)
+    toTxInp _ = error "we should have only our inputs at this point"
+    getOurInps apiDecodedTx =
+        let generalInps = apiDecodedTx ^. #inputs
+        in map toTxInp $ filter isInpOurs generalInps
```
Anviking, line 2409: Not that important, but I think you could combine `toTxInp` and `isInpOurs` to avoid the `error` case.

Anviking:
bors try

lib/core/src/Cardano/Wallet/Api/Server.hs
Anviking: Ok fair, it shouldn't be a problem with the construct-sign-submit flow.

> I think it is reasonable FIRST approximation. The next moves will be:(a) deal with collaterals, (b) deal with script inputs (here sometimes as you say several wits will be required -> we will need to have a way to decode that first then enhance condition)

Just note that I'm highly sceptical of defining the the condition ourselves, rather than looking at the error we get from the node.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
Anviking: Good point, I would think so too.

While part of me would be reluctant merging this check, I guess it might also end up being fine in practice. (Only time will tell)
 
Some property test (or any test with mock node) should be more stable as well.

Anviking (approved):
Ok, if the focus is `construct-sign-submit` this seems like a fair first step. Just that the pending txs will show the wrong TTL.

PR #3048, opened 2021-12-08
Restrict `Checkpoints` to be constructible from genesis only

HeinrichApfelmus:
### Issue number

ADP-1043

### Overview

In this step, we implement a suggestion that came up during review where we remove the `singleton` function in favor of `fromGenesis`, making sure that the `Checkpoints` structure always contains a checkpoint for the genesis block.

### Details

* The `initializeWallet` function will now throw an exception if it is incorrectly called with a checkpoint that is not at the genesis block.
* Many unit tests are adapted to use the `InitialCheckpoint` generator in order to comply with the above restriction.

### Comments

Merge PR #3046 before this one, because this pull request is based on the branch of the former.

### DIFF
--- a/lib/core/src/Cardano/Wallet/DB/Checkpoints.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Checkpoints.hs
@@ -13,8 +13,10 @@ module Cardano.Wallet.DB.Checkpoints
     ( getPoint
 
     -- * Checkpoints  
-    , Checkpoints (..)
-    , singleton
+    , Checkpoints
+    , checkpoints
+    , loadCheckpoints
+    , fromGenesis
     , getLatest
     , findNearestPoint
     
@@ -85,21 +87,30 @@ getPoint =
 -- | Collection of checkpoints indexed by 'Slot'.
 data Checkpoints a = Checkpoints
     { checkpoints :: Map W.Slot a
+    -- ^ Map of checkpoints. Always contains the genesis checkpoint.
     } deriving (Eq,Show,Generic)
 -- FIXME LATER during ADP-1043:
 --  Use a more sophisticated 'Checkpoints' type that stores deltas.
 
--- | Make a single checkpoint.
-singleton :: W.Slot -> a -> Checkpoints a
-singleton key a = Checkpoints $ Map.singleton key a
+-- | Turn the list of checkpoints into a map of checkpoints.
+--
+-- FIXME LATER during ADP-1043:
+--   The database actually does not store the checkpoint at genesis,
+--   but the checkpoint after that.
+--   Hence, this function does not check whether the genesis checkpoint
+--   is in the list of checkpoints.
+loadCheckpoints :: [(W.Slot, a)] -> Checkpoints a
+loadCheckpoints = Checkpoints . Map.fromList
+
+-- | Begin with the genesis checkpoint.
+fromGenesis :: a -> Checkpoints a
+fromGenesis a = Checkpoints $ Map.singleton W.Origin a
 
 -- | Get the checkpoint with the largest 'SlotNo'.
 getLatest :: Checkpoints a -> (W.Slot, a)
 getLatest = from . Map.lookupMax . view #checkpoints 
   where
-    from = fromMaybe (error "getLatest: Genesis checkpoint is missing!")
-    -- FIXME LATER during ADP-1043:
-    --   Make sure that 'Checkpoints' always has a genesis checkpoint
+    from = fromMaybe (error "getLatest: there should always be at least a genesis checkpoint")
 
 -- | Find the nearest 'Checkpoint' that is either at the given point or before.
 findNearestPoint :: Checkpoints a -> W.Slot -> Maybe W.Slot
@@ -123,7 +134,7 @@ instance Delta (DeltaCheckpoints a) where
     apply (RollbackTo pt) = over #checkpoints $
         Map.filterWithKey (\k _ -> k <= pt)
     apply (RestrictTo pts) = over #checkpoints $ \m ->
-        Map.restrictKeys m $ Set.fromList pts
+        Map.restrictKeys m $ Set.fromList (W.Origin:pts)
 
 {-------------------------------------------------------------------------------
     A Delta type for Maps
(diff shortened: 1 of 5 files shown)


lib/core/src/Cardano/Wallet/DB/Checkpoints.hs
```diff
@@ -85,21 +87,30 @@ getPoint =
...
     { checkpoints :: Map W.Slot a
+    -- ^ Map of checkpoints. Always contains the genesis checkpoint.
     } deriving (Eq,Show,Generic)
 -- FIXME LATER during ADP-1043:
 --  Use a more sophisticated 'Checkpoints' type that stores deltas.
 
--- | Make a single checkpoint.
-singleton :: W.Slot -> a -> Checkpoints a
-singleton key a = Checkpoints $ Map.singleton key a
+-- | Turn a list of checkpoints.
```
paweljakubas:
```suggestion
-- | Turn the list of checkpoints into a map of checkpoints.
```

lib/core/src/Cardano/Wallet/DB/Checkpoints.hs
```diff
@@ -85,21 +87,30 @@ getPoint =
...
+--   The database actually does not store the checkpoint at genesis,
+--   but the checkpoint after that.
+--   Hence, this function does not check whether the genesis checkpoint
+--   is in the list of checkpoints.
+loadCheckpoints :: [(W.Slot, a)] -> Checkpoints a
+loadCheckpoints = Checkpoints . Map.fromList
+
+-- | Begin with the genesis checkpoint.
+fromGenesis :: a -> Checkpoints a
+fromGenesis a = Checkpoints $ Map.singleton W.Origin a
```
paweljakubas, line 107: :+1:

lib/core/src/Cardano/Wallet/DB/Checkpoints.hs
```diff
@@ -85,21 +87,30 @@ getPoint =
...
 
 -- | Get the checkpoint with the largest 'SlotNo'.
 getLatest :: Checkpoints a -> (W.Slot, a)
 getLatest = from . Map.lookupMax . view #checkpoints 
   where
-    from = fromMaybe (error "getLatest: Genesis checkpoint is missing!")
-    -- FIXME LATER during ADP-1043:
-    --   Make sure that 'Checkpoints' always has a genesis checkpoint
+    from = fromMaybe (error "getLatest: impossible: no genesis checkpoint?!")
```
paweljakubas:
```suggestion
    from = fromMaybe (error "getLatest: there should always be at least a genesis checkpoint")
```

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -535,15 +535,26 @@ newDBLayerWith _cacheBehavior tr ti SqliteContext{runQuery} = do
...
+                    Nothing  -> (Nothing, ())
                     Just _   -> (Just $ Adjust wid $ PutCheckpoint point cp, ())
 
+    -- Insert genesis checkpoint into the DBVar.
+    -- Throws an internal error if the checkpoint is not actually at genesis.
+    let insertCheckpointGenesis wid cp
+            | W.isGenesisBlockHeader header =
+                updateDBVar checkpointsVar $ Insert wid $ fromGenesis cp
+            | otherwise =
+                throwIO $ ErrInitializeNotGenesis wid header
```
paweljakubas, line 554: thinking if exposing `ErrInitializeNotGenesis` as `pure $ Left ...`. we could have, then no `ErrNoSuchWallet` here
```
    , putCheckpoint
        :: WalletId
        -> Wallet s
        -> ExceptT ErrNoSuchWallet stm ()
```
but ErrPutCheckpoint which is sum type witth error no wallet and initialize not genesis?

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -1547,3 +1561,9 @@ selectGenesisParameters wid = do
 -- violated.
 data ErrRollbackTo = ErrNoOlderCheckpoint W.WalletId W.Slot deriving (Show)
 instance Exception ErrRollbackTo
+
+-- | Can't initialize a wallet because the given 'BlockHeader' is not genesis.
+data ErrInitializeNotGenesis
```
paweljakubas, line 1566: If we do what I suggested above then this would go here https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/DB.hs

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
```diff
@@ -717,10 +717,15 @@ prop_randomOpChunks (KeyValPairs pairs) =
             unsafeRunExceptT $ putCheckpoint k cp
             unsafeRunExceptT $ putWalletMeta k meta
         else do
-            atomically $ unsafeRunExceptT $ initializeWallet k cp meta mempty gp
+            let cp0 = imposeGenesisState cp
+            atomically $ unsafeRunExceptT $ initializeWallet k cp0 meta mempty gp
             Set.fromList <$> atomically listWallets
                 `shouldReturn` Set.fromList (k:keys)
 
+    imposeGenesisState :: Wallet s -> Wallet s
```
paweljakubas, line 725: or maybe `enforce` ? ;-) `impose` is also good and conveys idea here

paweljakubas (approved):
LGTM! Left several suggestions. Maybe you will find defining new sum error in putCheckpoint method of DBLayer interface reasonable. I feel we should not differentiate no wallet and no genesis from each other in a sense they are both very serious/critical and if so why not to treat them on the same level ;-)

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
HeinrichApfelmus: I opted to go for an exception here, because `ErrInitializeNotGenesis` is a programming error (it's the responsibility of the caller of `initializeWallet` to make sure that only the genesis checkpoint is passed), whereas `ErrNoSuchWallet` may be a user error (the user specified a wallet ID that is not in the database).

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
HeinrichApfelmus: May the `impose` be with you! :D

HeinrichApfelmus:
bors merge

HeinrichApfelmus:
bors retry

PR #3050, opened 2021-12-09
Bump cardano-node to 1.32.1

rvl:
Updates to [cardano-node 1.32.1](https://github.com/input-output-hk/cardano-node/releases/tag/1.32.1)

Release notes for [1.31.0](https://github.com/input-output-hk/cardano-node/releases/tag/1.31.0).

- [x] Updates cardano-node version in `stack.yaml`.
- [x] Updates cardano-node version in `cabal.project`.
- [x] Fix new build errors.
- [x] Fix any test failures.
- [x] Update readme.

### Comments

- [Hydra jobset](https://hydra.iohk.io/jobset/Cardano/cardano-wallet-pr-3050)

See also: #3023 #3038

### Issue Number

ADP-1266


rvl:
bors try

rvl:
bors try

lib/core/src/Cardano/Api/Gen.hs
```diff
@@ -506,9 +504,9 @@ genScriptData =
 
 genExecutionUnits :: Gen ExecutionUnits
 genExecutionUnits = do
-    (Large steps) <- arbitrary
-    (Large mem) <- arbitrary
-    pure $ ExecutionUnits steps mem
+    (Large (steps :: Int)) <- arbitrary
+    (Large (mem :: Int)) <- arbitrary
```
Anviking: Using `Word64` might be the safer choice to preserve the previous behaviour. (`ExecuionUnits` had Word64 fields on 1.30.1)

lib/core/src/Cardano/Api/Gen.hs
```diff
@@ -708,18 +706,18 @@ genTxOutValue era =
...
-genTxOutDatumHash :: CardanoEra era -> Gen (TxOutDatumHash era)
+genTxOutDatumHash :: CardanoEra era -> Gen (TxOutDatum ctx era)
 genTxOutDatumHash era =
     case scriptDataSupportedInEra era of
-        Nothing -> pure TxOutDatumHashNone
+        Nothing -> pure TxOutDatumNone
         Just supported -> oneof
-            [ pure TxOutDatumHashNone
+            [ pure TxOutDatumNone
             , TxOutDatumHash supported <$> genHashScriptData
```
Anviking: Interesting, it seems there's both `TxOutDatum'` and `TxOutDatumHash` now.

Could we rename `genTxOutDatumHash` to `genTxOutDatum` and also generate `TxOutDatum'`?

Or at the least add a note about it.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -738,17 +739,17 @@ fromLedgerExUnits
     -> W.ExecutionUnits
 fromLedgerExUnits (Alonzo.ExUnits mem steps) =
     W.ExecutionUnits
-    { executionSteps = steps
-    , executionMemory = mem
+    { executionSteps = fromIntegral steps
+    , executionMemory = fromIntegral mem
```
Anviking: This looks like an unsafe `Natural` to `Word64` conversion.

I believe this is something we _really_ don't want to overflow (even if it's impossible or has no effect at the moment).

Can we change `W.ExecutionUnits` to also use `Natural` instead of `Word64`?

Anviking (approved):
Great! lgtm modulo three comments which I think should be addressed.

Anviking:
Also don't forget to update the readme compatibility table!

Anviking:
I rebased, some build error remains though

rvl:
Thanks ... but you may have accidentally force pushed over some commits :grin:. I have pushed them back now, and will address your review comments shortly.

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
rvl: Yes, definitely.

lib/core/src/Cardano/Api/Gen.hs
rvl: I renamed `genTxOutDatumHash` to `genTxOutDatum`, but where is this `TxOutDatum'` ?

lib/core/src/Cardano/Api/Gen.hs
rvl: Yes - fixed.

lib/core/src/Cardano/Api/Gen.hs
Anviking: https://github.com/input-output-hk/cardano-node/blob/4f65fb9a27aa7e3a1873ab4211e412af780a3648/cardano-api/src/Cardano/Api/TxBody.hs#L1136-L1143

I think a note would be good

Anviking:
> but you may have accidentally force pushed over some commits 😁. I have pushed them back now

🙈 👍

PR #3056, opened 2021-12-13
Introduce `Prologue` and `Discoveries` of an address book

HeinrichApfelmus:
### Issue number

ADP-1043

### Overview

In this step, we begin refactoring the address discovery state (the `s` in `Wallet s`). First, I would like to update our nomenclature and call this state an *address book*. Second, we introduce two new data families `Prologue s` and `Discoveries s`, with the intention that any address book is isomorphic to a pair of these two types, that is `s ~ (Prologue s, Discoveries s)`. The idea is that the Prologue of the address book contains "static" information such as public keys or the address gap (default 20), while the Discoveries of the address book contains information that was gathered during on-chain address discovery.

### Details

* We introduce a type class `AddressBookIso` which provides the isomorphism `s ~ (Prologue s, Discoveries s)`.
* For this pull request, we avoid making any changes to the `AddressDiscovery.*` hierarchy. Instead, the isomorphisms are implemented using newly defined data families whose representation is close to the types in the database.
* The implementation for the sequential states uses `liftPaymentAddress` to convert the payment part to and from the `Address` type. This is ok when convering to/from the database layer, but does not make sense for in-memory data structures. Therefore, further work in this epic will introduce changes in the `AddressDiscovery.*` hierarchy.

### Comments

* The overall goal is that, in a future PR, we can represent the wallet state along the lines of `WalletState s ~ (Prologue s, Checkpoints (BlockHeader, UTxO, Discoveries s) `.
* Merge PR #3048 before this one, because this pull request is based on the branch of the former.

### DIFF
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite/AddressBook.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite/AddressBook.hs
@@ -1,0 +1,228 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+{-# LANGUAGE TypeFamilies #-}
+-- |
+-- Copyright: © 2021 IOHK
+-- License: Apache-2.0
+--
+-- Address books consist of a 'Prologue'
+-- and discovered addresses ('Discoveries').
+module Cardano.Wallet.DB.Sqlite.AddressBook
+    ( AddressBookIso (..)
+    , Prologue (..)
+    , Discoveries (..)
+    , SeqAddressList (..)
+    )
+  where
+
+import Prelude
+
+import Cardano.Address.Derivation
+    ( XPub )
+import Cardano.Wallet.Primitive.AddressDerivation
+    ( Depth (..)
+    , MkKeyFingerprint (..)
+    , NetworkDiscriminant (..)
+    , PaymentAddress (..)
+    , Role (..)
+    , SoftDerivation (..)
+    )
+import Cardano.Wallet.Primitive.AddressDerivation.SharedKey
+    ( SharedKey (..) )
+import Cardano.Wallet.Primitive.AddressDiscovery
+    ( GetPurpose )
+import Cardano.Wallet.Primitive.Types.Address
+    ( Address (..), AddressState (..) )
+import Data.Generics.Internal.VL
+    ( Iso', iso, withIso )
+import Data.Kind
+    ( Type )
+import Data.Map.Strict
+    ( Map )
+import Data.Proxy
+    ( Proxy (..) )
+import Data.Type.Equality
+    ( type (==) )
+import Data.Typeable
+    ( Typeable )
+
+import qualified Cardano.Wallet.Primitive.AddressDiscovery.Random as Rnd
+import qualified Cardano.Wallet.Primitive.AddressDiscovery.Sequential as Seq
+import qualified Cardano.Wallet.Primitive.AddressDiscovery.Shared as Shared
+import qualified Cardano.Wallet.Primitive.Types.Address as W
+import qualified Data.Map.Strict as Map
+
+{-------------------------------------------------------------------------------
+    AddressBook isomorphism
+-------------------------------------------------------------------------------}
+-- | FIXME LATER during ADP-1043:
+-- Move 'Prologue' and 'Discoveries' closer into address types.
+class AddressBookIso s where
+    -- | Address information contained in the prologue of the address book,
+    -- such as public keys or the address gap.
+    data Prologue s :: Type
+    -- | Addresses that were collected during discovery on the blockchain.
+    data Discoveries s :: Type
+
+    -- | Isomorphism between the address book type 's'
+    -- and its two components.
+    addressIso :: Iso' s (Prologue s, Discoveries s)
+
+{-------------------------------------------------------------------------------
+    Sequential address book
+-------------------------------------------------------------------------------}
+-- | Sequential list of addresses
+newtype SeqAddressList (c :: Role) = SeqAddressList [(W.Address, W.AddressState)]
+
+-- piggy-back on SeqState existing instance, to simulate the same behavior.
+instance AddressBookIso (Seq.SeqState n k)
+    => AddressBookIso (Seq.SeqAnyState n k p)
+  where
+    data Prologue (Seq.SeqAnyState n k p) = PS (Prologue (Seq.SeqState n k))
+    data Discoveries (Seq.SeqAnyState n k p) = DS (Discoveries (Seq.SeqState n k))
+
+    addressIso = withIso addressIso $ \from to ->
+        let from2 st = let (a,b) = from $ Seq.innerState st in (PS a, DS b)
+            to2 (PS a, DS b) = Seq.SeqAnyState $ to (a,b)
+        in  iso from2 to2
+
+-- | Isomorphism for sequential address book.
+instance 
+    ( MkKeyFingerprint key (Proxy n, key 'AddressK XPub)
+    , GetPurpose key
+    , SoftDerivation key
+    , PaymentAddress n key
+    , Typeable n
+    , (key == SharedKey) ~ 'False
+    ) => AddressBookIso (Seq.SeqState n key)
+  where
+    data Prologue (Seq.SeqState n key)
+        = SeqPrologue (Seq.SeqState n key)
+        -- Trick: We keep the type, but we empty the discovered addresses
+    data Discoveries (Seq.SeqState n key)
+        = SeqDiscoveries
+            (SeqAddressList 'UtxoInternal)
+            (SeqAddressList 'UtxoExternal)
+
+    addressIso = iso from to
+      where
+        from (Seq.SeqState int ext a b c) =
+            ( SeqPrologue $ Seq.SeqState (emptyPool int) (emptyPool ext) a b c
+            , SeqDiscoveries (toDiscoveries @n int) (toDiscoveries @n ext)
+            )
+        to (SeqPrologue (Seq.SeqState int ext a b c), SeqDiscoveries ints exts)
+          = Seq.SeqState (fromDiscoveries @n int ints) (fromDiscoveries @n ext exts) a b c
+
+-- | Extract the discovered addresses from an address pool.
+toDiscoveries
+    :: forall (n :: NetworkDiscriminant) (c :: Role) key.
+    ( GetPurpose key
+    , PaymentAddress n key
+    , Typeable c
+    ) => Seq.AddressPool c key -> SeqAddressList c
+toDiscoveries pool = SeqAddressList
+    [ (a,st) | (a,st,_) <- Seq.addresses (liftPaymentAddress @n) pool ]
+
+-- | Variant of 'toDiscoveries' that can be used with 'SharedKey'
+toDiscoveriesShared
+    :: forall (n :: NetworkDiscriminant) (c :: Role) key.
+    ( GetPurpose key
+    , Typeable c
+    , Typeable n
+    , key ~ SharedKey
+    ) => Seq.AddressPool c key -> SeqAddressList c
+toDiscoveriesShared pool = SeqAddressList
+    [ (a,st) | (a,st,_) <- Seq.addresses (Shared.liftPaymentAddress @n) pool ]
+
+-- | Fill an empty address pool with addresses.
+fromDiscoveries
+    :: forall (n :: NetworkDiscriminant) (c :: Role) key.
+    ( MkKeyFingerprint key (Proxy n, key 'AddressK XPub)
+    , MkKeyFingerprint key Address
+    , SoftDerivation key
+    , Typeable n
+    , Typeable c
+    ) => Seq.AddressPool c key -> SeqAddressList c -> Seq.AddressPool c key
+fromDiscoveries ctx (SeqAddressList addrs) =
+    Seq.mkAddressPool @n (Seq.context ctx) (Seq.gap ctx) addrs
+
+-- | Remove all discovered addresses from an address pool,
+-- but keep context.
+emptyPool :: Seq.AddressPool c key -> Seq.AddressPool c key
+emptyPool pool = pool{ Seq.indexedKeys = Map.empty }
+
+{-------------------------------------------------------------------------------
+    Shared key address book
+-------------------------------------------------------------------------------}
+-- | Isomorphism for multi-sig address book.
+instance
+    ( MkKeyFingerprint key (Proxy n, key 'AddressK XPub)
+    , MkKeyFingerprint key Address
+    , GetPurpose key
+    , SoftDerivation key
+    , Typeable n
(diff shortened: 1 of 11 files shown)


lib/core/src/Cardano/Wallet/DB/Sqlite/AddressBook.hs
```diff
@@ -0,0 +1,228 @@
...
+class AddressBookIso s where
+    -- | Address information contained in the prologue of the address book,
+    -- such as public keys or the address gap.
+    data Prologue s :: Type
+    -- | Addresses that were collected during discovery on the blockchain.
+    data Discoveries s :: Type
+
+    -- | Isomorphism between the address book type 's'
+    -- and its two components.
+    addressIso :: Iso' s (Prologue s, Discoveries s)
```
paweljakubas, line 76: :+1:

lib/core/src/Cardano/Wallet/DB/Sqlite/AddressBook.hs
```diff
@@ -0,0 +1,228 @@
...
+
+{-------------------------------------------------------------------------------
+    Sequential address book
+-------------------------------------------------------------------------------}
+-- | Sequential list of addresses
+newtype SeqAddressList (c :: Role) = SeqAddressList [(W.Address, W.AddressState)]
```
paweljakubas, line 82: Why we have here list? Because we are implying that the sequence of addresses correspond to address indices?So we have first entry corresponding to soft index ix=0, etc.? Why not to have a map with soft index as key? As we plan to enlarge this list (via `(++) [new address]`) or rollback usually with operations in the end of it...it seems the list could be quite expensive - naively thinking. Maybe strict map would be better here and we would not loose the capabilities here? What do you think about this argument?

HeinrichApfelmus: I completely agree that we will need a more efficient data structure than lists here — which is why I started refactoring the address pool logic, and was able to implement it for `SharedStates` in the upcoming PR #3068 😄. Step by step.

For the present PR, the list is sufficient: At the moment, the list is created when loading from the database and then immediately converted into an address pool through the `addressIso`. Similar procedure when storing to the database. As long as the conversion with `addressIso` happens just at these points, immediately before or after the database is accessed, we do pretty much exactly the same thing that the previous version of our code did, so we do not pay a performance penalty for now. Of course, once the isomorphism is used in the context of `Checkpoints (Discoveries s, UTxO,…)`, we need it to be faster. But that is for another PR.

paweljakubas: ok. fair enough

lib/core/src/Cardano/Wallet/DB/Sqlite/AddressBook.hs
```diff
@@ -0,0 +1,228 @@
...
+
+-- | Extract the discovered addresses from an address pool.
+toDiscoveries
+    :: forall (n :: NetworkDiscriminant) (c :: Role) key.
+    ( GetPurpose key
+    , PaymentAddress n key
+    , Typeable c
+    ) => Seq.AddressPool c key -> SeqAddressList c
+toDiscoveries pool = SeqAddressList
+    [ (a,st) | (a,st,_) <- Seq.addresses (liftPaymentAddress @n) pool ]
```
paweljakubas, line 131: We do like this because SeqAddressList is directly mapped to DB, right?

paweljakubas: if yes, then for the future reference we need to remember about some probable future developments here:
1. multi accountsupport - in that case liftPaymentAddress will not be enough as payment credential is not enough. We would need to also differentiate against staking credential so staking key hash (for Sequential mode) or staking script hash (for Shared mode)
2. we have (at least in Sequential mode) several fingerprints of address to support: the current one (aka base with key hash in payment credential), enterprise address (key hash payment credenital without staking credential) and pointer address (key hash payment with pointer stake credential)
We need (probably later) to discuss if new abstraction are able to seamlessly support them

lib/core/src/Cardano/Wallet/DB/Sqlite/AddressBook.hs
```diff
@@ -0,0 +1,228 @@
...
+-- | Variant of 'toDiscoveries' that can be used with 'SharedKey'
+toDiscoveriesShared
+    :: forall (n :: NetworkDiscriminant) (c :: Role) key.
+    ( GetPurpose key
+    , Typeable c
+    , Typeable n
+    , key ~ SharedKey
+    ) => Seq.AddressPool c key -> SeqAddressList c
+toDiscoveriesShared pool = SeqAddressList
+    [ (a,st) | (a,st,_) <- Seq.addresses (Shared.liftPaymentAddress @n) pool ]
```
paweljakubas, line 142: the same comment as before... if we want to cover all possible addresses having script hashes in credentials this could not be 1-to-1 mapping anymore also here

lib/core/src/Cardano/Wallet/DB/Sqlite/AddressBook.hs
```diff
@@ -0,0 +1,228 @@
...
+                    , SharedDiscoveries (toDiscoveriesShared @n pool)
+                    )
+        to  ( SharedPrologue st@(Shared.SharedState a b0)
+            , SharedDiscoveries addrs
+            )
+          = case b0 of
+            Shared.PendingFields{} -> st
+            Shared.ReadyFields pool0
+                -> Shared.SharedState a $ Shared.ReadyFields
+                $ fromDiscoveries @n pool0 addrs
```
paweljakubas, line 199: ok, should be fine :+1:

lib/core/src/Cardano/Wallet/DB/Sqlite/AddressBook.hs
```diff
@@ -0,0 +1,228 @@
...
+    data Discoveries (Rnd.RndAnyState n p) = DR (Discoveries (Rnd.RndState n))
+
+    addressIso = withIso addressIso $ \from to ->
+        let from2 st = let (a,b) = from $ Rnd.innerState st in (PR a, DR b)
+            to2 (PR a, DR b) = Rnd.RndAnyState $ to (a,b)
+        in  iso from2 to2
+
+-- | Isomorphism for HD random address book.
+instance AddressBookIso (Rnd.RndState n) where
+    data Prologue (Rnd.RndState n)
```
paweljakubas, line 217: will look at this during the second look tomorrow. Looks correct at first sight

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
```diff
@@ -225,19 +234,20 @@ mkStoreCheckpoints wid =
...
-    :: forall s. PersistState s
+    :: forall s. PersistAddressBook s
     => W.WalletId
     -> SqlPersistT IO [(W.Slot, W.Wallet s)]
 selectAllCheckpoints wid = do
     cps <- fmap entityVal <$> selectList
         [ CheckpointWalletId ==. wid ]
         [ Desc CheckpointSlot ]
-    fmap catMaybes $ forM cps $ \cp -> do
+    -- FIXME LATER during ADP-1043: Presence of these tables?
```
paweljakubas, line 246: would be good to have check on if the tables exist and only after they are proceed :+1:

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
```diff
@@ -225,19 +234,20 @@ mkStoreCheckpoints wid =
...
+    prologue <- fromJust <$> loadPrologue wid
```
paweljakubas: so here we have assumption that static part (aka Prologue) never changes ... we will need to be careful here when adding support for multi accounts... list of acctXPub, so a given accXPub, could stop to be static...just writing for future ourselves;-)

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
```diff
@@ -225,19 +234,20 @@ mkStoreCheckpoints wid =
...
+    forM cps $ \cp -> do
         utxo <- selectUTxO cp
-        st <- selectState (checkpointId cp)
-        pure $
-            (\s -> let c = checkpointFromEntity @s cp utxo s in (getPoint c, c))
-            <$> st
+        discoveries <- loadDiscoveries wid (checkpointSlot cp)
+        let st = withIso addressIso $ \_ from -> from (prologue, discoveries)
+        pure $ let c = checkpointFromEntity @s cp utxo st in (getPoint c, c)
```
paweljakubas: would be in favor of extracting let in the above line. just for the sake of clarity

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
```diff
@@ -349,44 +362,45 @@ checkpointFromEntity cp (coins, tokens) =
...
-
--- | Functions for saving/loading the wallet's address discovery state into
--- SQLite.
-class PersistState s where
-    -- | Store the state for a checkpoint.
-    insertState :: (W.WalletId, W.SlotNo) -> s -> SqlPersistT IO ()
-    -- | Load the state for a checkpoint.
-    selectState :: (W.WalletId, W.SlotNo) -> SqlPersistT IO (Maybe s)
+-- | Functions for saving / loading the wallet's address book to / from SQLite
+class AddressBookIso s => PersistAddressBook s where
```
paweljakubas, line 375: I am fond of this abstraction :100:

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
```diff
@@ -475,21 +464,33 @@ selectSeqStatePendingIxs wid =
   where
     fromRes = fmap (W.Index . seqStatePendingIxIndex . entityVal)
 
+selectSeqAddressList
+    :: forall c. Typeable c
+    => W.WalletId -> W.SlotNo -> SqlPersistT IO (SeqAddressList c)
```
paweljakubas, line 476: ok, so I have answer here about relation of SeqAddressList and DB ;-)

lib/core/test/unit/Cardano/WalletSpec.hs
```diff
@@ -1330,9 +1331,19 @@ newtype DummyState
     = DummyState (Map Address (Index 'Soft 'AddressK))
     deriving (Generic, Show, Eq)
 
-instance Sqlite.PersistState DummyState where
-    insertState _ _ = error "DummyState.insertState: not implemented"
-    selectState _ = error "DummyState.selectState: not implemented"
+instance Sqlite.AddressBookIso DummyState where
```
paweljakubas, line 1334: At some point, at other PR,  maybe we could have more useful DummyState instances here and get additional confidence by adding tests? Do you see something potentially useful here?

HeinrichApfelmus: As far as I understand, the purpose of the `WalletSpec` module is to test that the different layers work well together; however, the individual layers such as the DB layer are tested in other modules. So, I wouldn't make `DummyState` more fancy here, but rather ensure that we have sufficient testing of the individual layers.

In this case, I was wondering whether it was worth testing that the instances of `AddressBookIso` are indeed isomorphisms, but then decided against it, as the will-test-that code is too close to the to-be-tested code, and the current `SqliteSpec` state machine tests would highlight any issues here as well.

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
HeinrichApfelmus: Yes. Perhaps a good way to think about is that the "static part" is only "static" in the sense that it does not depend on information that discovered on the blockchain, but not "static" in the sense that it can never change. That's one of the reasons why I have called it "prologue" instead of "static". 😄

lib/core/src/Cardano/Wallet/DB/Sqlite/AddressBook.hs
HeinrichApfelmus: > We do like this because SeqAddressList is directly mapped to DB, right?

Yes, indeed. Will be changed again in the next steps.

> […] staking credential […]

That's a very good point. From looking at the code, I got the impression that the Sequential state only looks at the payment key hashes. In other words, whenever the state sees an address on chain (which may have a stake key hash), it will extract the payment key hash, add it to the address pool as appropriate, convert it to an enterprise address and store that address in the database. Put differently, the database stores discovered Sequential payment key hashes, but formatted as enterprise addresses.

The situation for Shared states was not entirely clear to me, but by chasing the various type classes, I also got the impression that only the payment key hash is subject to discovery. In other words, if an address appears on the chain, and the stake credential of this address does not match the delegation template, but the payment credential does match the payment template, then it will still be discovered and added to the database. Makes sense: If you own the payment credential, you can always reassign the delegation part by making a transaction; the access control on the payment credential partially overrides the access control on the delegation part. I suppose we will need to think about this corner case again. 🤔 

In any case, for the time being, this PR only attempts to do what the code did before, no changes in functionality intended. 😅

paweljakubas: so for shared wallets we store payment credential only which is script hash. shared wallet's credential and sequential wallet credential happen to have the same byte imprint as the same hashes (blake224) in both cases is used.

What is important here is that maybe to avoid complications and choose simplicity it would be better to store both credentials, whatever they are represented (keyhash, scripthash or three numbers for pointer addresses). It would allow support of any address type and also seamlessly support multiaccount (on the level of discovery) 

I feel we will need to revisit this soon or later (maybe even during shared wallet work soon). I also see challenges (db migration issues....)

lib/core/src/Cardano/Wallet/DB/Sqlite/AddressBook.hs
paweljakubas: looks fine to my best knwoledge

lib/core/test/unit/Cardano/WalletSpec.hs
paweljakubas: ok, fair enough

paweljakubas (approved):
great PR! very fond of introducing of decoupling between nonchanging and changing parts of the state. And treat them explicitely as separate.

lib/core/src/Cardano/Wallet/DB/Sqlite/AddressBook.hs
HeinrichApfelmus: > I feel we will need to revisit this soon or later (maybe even during shared wallet work soon). I also see challenges (db migration issues....)

Agreed. I suggest that we leave things as they are for now and revisit this soon during the shared wallet work.

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
HeinrichApfelmus: Good point. I have added a commit where a new error type `ErrBadFormat` is introduced, which I intend to eventually cover all errors related to malformed database files. As a first step, it absorbs the `fromJust` case here.

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
HeinrichApfelmus: Added in the additional commit.

HeinrichApfelmus:
bors merge

PR #3063, opened 2021-12-15
Bump cardano-addresses to 3.7.0

rvl:
Updates cardano-addresses to [3.7.0](), which adds an `address_type` field to the inspectAddress output.

### Issue Number

[DDW-742](https://input-output.atlassian.net/browse/DDW-742)

### DIFF
--- a/specifications/api/swagger.yaml
+++ b/specifications/api/swagger.yaml
@@ -2085,6 +2085,11 @@ components:
             description: Only for 'Byron' style.
             type: string
             format: base16
+          address_type:
+            description: The raw type field of the address.
+            type: integer
+            minimum: 0
+            maximum: 15
 
     ApiNetworkTip: &ApiNetworkTip
       <<: *slotReference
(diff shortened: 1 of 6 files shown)


specifications/api/swagger.yaml
```diff
@@ -2085,6 +2085,11 @@ components:
             description: Only for 'Byron' style.
             type: string
             format: base16
+          address_type:
+            description: The raw type field of the address.
```
paweljakubas: I wonder if we should not somewhere here also add the table as in cardano-addresses describing codes?

paweljakubas (approved):
should be fine - we will see if all tests pass. We can maybe add some assertions in integration testing? it should be "address_type":0 as we are dealing mainly with base addressess having in both payment and staking credential key hashes

piotr-iohk:
bors try

specifications/api/swagger.yaml
piotr-iohk: Added table and some tests.

PR #3065, opened 2021-12-17
Add integration test for ADP-1286 corner case

piotr-iohk:
- [ ] I have added integration test for issue with tx balancing from ADP-1286 (currently failing)

### Comments

@paweljakubas we can either merge as `pending` and enable once the bug is fixed or... add fix on top of this.

### Issue Number

ADP-1286


paweljakubas (approved):
let's make it pending and merge it

PR #3067, opened 2021-12-20
Improve asset assertion in e2e tests

piotr-iohk:
- [ ] I have improved assertion checking asset balance after tx so it is possible to assert specific asset balance

### Comments

This is to get rid of potential false positives when there are some more assets on fixture wallets.


paweljakubas (approved):
lgtm

PR #3072, opened 2022-01-04
Fix regenerate-byron.sh to comply with current cardano-cli

piotr-iohk:
- [ ] Fix regenerate-byron.sh to comply with current cardano-cli

### DIFF
--- a/lib/shelley/test/data/cardano-node-shelley/regenerate-byron.sh
+++ b/lib/shelley/test/data/cardano-node-shelley/regenerate-byron.sh
@@ -42,7 +42,7 @@ cardano-cli byron genesis genesis \
    --protocol-parameters-file byron.genesis.spec.json \
    --genesis-output-dir tmp
 
-cat tmp/genesis.json | yq e -P - > byron-genesis-init.yaml
+yq -y . < tmp/genesis.json > byron-genesis-init.yaml
 mv -vf tmp/delegate-keys.*.key .
 mv -vf tmp/delegation-cert.*.json .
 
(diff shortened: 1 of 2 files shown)


lib/shelley/test/data/cardano-node-shelley/regenerate-byron.sh
```diff
@@ -29,21 +29,20 @@ cat > byron.genesis.spec.json <<EOF
...
    --total-balance 50 \
-   --byron-formats \
    --delegate-share 1 \
    --avvm-entry-count 0 \
    --avvm-entry-balance 0 \
    --protocol-parameters-file byron.genesis.spec.json \
    --genesis-output-dir tmp
 
-yq -y . < tmp/genesis.json > byron-genesis-init.yaml
+cat tmp/genesis.json | yq e -P - > byron-genesis-init.yaml
```
rvl: This is a very different `yq` from the one I have... maybe let's just use json.

```suggestion
jq . tmp/genesis.json > byron-genesis-init.yaml
```

rvl (approved):

lib/shelley/test/data/cardano-node-shelley/regenerate-byron.sh
piotr-iohk: Oh, it seems there are two:
 - https://kislyuk.github.io/yq/
 - https://mikefarah.gitbook.io/yq/

I had the latter, going back to the first one... (as it seems it is used here and there across different scripts)

rvl (approved):
Thanks

PR #3073, opened 2022-01-04
Refactor and simplify `SeqState n k`

HeinrichApfelmus:
### Issue number

ADP-1308

### Overview

In this step, we continue refactoring the address discovery state. Here, we refactor and simplify the `SeqState n k` type to use the new abstract data type `Pool addr ix`, which aids with BIP-44 style address discovery.

### Details

* The testing module `PoolSpec` now also provides a shrinker `shrinkPool` for use in the old testing module `Cardano.Wallet.Primitive.AddressDiscovery.SequentialSpec`.
* The property tests pertaining to the address discovery aspects of `SeqState` are superseded by the more general unit tests in `Cardano.Wallet.Address.PoolSpec`.
 
### Comments

* Merge PR ##3068 before this one, because this pull request is based on the branch of the former.

### DIFF
--- a/lib/core/bench/db-bench.hs
+++ b/lib/core/bench/db-bench.hs
@@ -87,7 +87,6 @@ import Cardano.Wallet.Primitive.AddressDerivation
     , Passphrase (..)
     , PaymentAddress (..)
     , PersistPrivateKey
-    , Role (..)
     , WalletKey (..)
     )
 import Cardano.Wallet.Primitive.AddressDerivation.Byron
@@ -97,14 +96,12 @@ import Cardano.Wallet.Primitive.AddressDerivation.Shelley
 import Cardano.Wallet.Primitive.AddressDiscovery.Random
     ( DerivationPath, RndState (..), mkRndState )
 import Cardano.Wallet.Primitive.AddressDiscovery.Sequential
-    ( AddressPool
-    , DerivationPrefix (..)
-    , ParentContext (..)
+    ( DerivationPrefix (..)
+    , SeqAddressPool (..)
     , SeqState (..)
     , coinTypeAda
     , defaultAddressPoolGap
-    , emptyPendingIxs
-    , mkAddressPool
+    , mkSeqStateFromAccountXPub
     , mkSeqStateFromRootXPrv
     , purposeCIP1852
     )
@@ -179,6 +176,8 @@ import Data.Functor
     ( ($>) )
 import Data.Functor.Identity
     ( Identity (..) )
+import Data.List
+    ( foldl' )
 import Data.Map.Strict
     ( Map )
 import Data.Maybe
@@ -216,6 +215,7 @@ import UnliftIO.Temporary
 
 import qualified Cardano.BM.Configuration.Model as CM
 import qualified Cardano.BM.Data.BackendKind as CM
+import qualified Cardano.Wallet.Address.Pool as AddressPool
 import qualified Cardano.Wallet.Primitive.AddressDerivation.Byron as Byron
 import qualified Cardano.Wallet.Primitive.Types.TokenBundle as TokenBundle
 import qualified Cardano.Wallet.Primitive.Types.TokenMap as TokenMap
@@ -359,13 +359,7 @@ bgroupWriteSeqState db = bgroup "SeqState"
             pure cps
         cps :: [WalletBench]
         cps =
-            [ snd $ initWallet (withMovingSlot i block0) $
-                SeqState
-                    (mkIntPool a i)
-                    (mkExtPool a i)
-                    emptyPendingIxs
-                    rewardAccount
-                    defaultPrefix
+            [ snd $ initWallet (withMovingSlot i block0) $ mkSeqState a i
             | i <- [1..n]
             ]
 
@@ -373,17 +367,18 @@ benchPutSeqState :: DBLayerBench -> [WalletBench] -> IO ()
 benchPutSeqState DBLayer{..} cps = do
     unsafeRunExceptT $ mapExceptT atomically $ mapM_ (putCheckpoint testWid) cps
 
-mkExtPool :: Int -> Int -> AddressPool  'UtxoExternal ShelleyKey
-mkExtPool numAddrs i =
-    mkAddressPool @'Mainnet (ParentContextUtxo ourAccount) defaultAddressPoolGap addrs
+mkSeqState :: Int -> Int -> SeqState 'Mainnet ShelleyKey
+mkSeqState numAddrs _ = s 
+    { internalPool = fillPool (internalPool s)
+    , externalPool = fillPool (externalPool s)
+    }
   where
-    addrs = [ force (mkAddress i j, Unused) | j <- [1..numAddrs] ]
-
-mkIntPool :: Int -> Int -> AddressPool 'UtxoInternal ShelleyKey
-mkIntPool numAddrs i =
-    mkAddressPool @'Mainnet (ParentContextUtxo ourAccount) defaultAddressPoolGap addrs
-  where
-    addrs = [ force (mkAddress i j, Unused) | j <- [1..numAddrs] ]
+    s = mkSeqStateFromAccountXPub @'Mainnet
+        ourAccount purposeCIP1852 defaultAddressPoolGap
+    fillPool (SeqAddressPool pool0) = SeqAddressPool $
+        foldl' (\p ix -> AddressPool.update (gen ix) p) pool0 [0 .. numAddrs-1]
+      where
+        gen ix = AddressPool.generator pool0 $ toEnum ix
 
 ----------------------------------------------------------------------------
 -- Wallet State (Random Scheme) Benchmarks
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -352,6 +352,7 @@ import Cardano.Wallet.Primitive.AddressDiscovery.Sequential
     ( DerivationPrefix (..)
     , SeqState (..)
     , defaultAddressPoolGap
+    , getGap
     , mkSeqStateFromAccountXPub
     , mkSeqStateFromRootXPrv
     , purposeCIP1852
@@ -748,7 +749,6 @@ postWallet
         , MkKeyFingerprint k (Proxy n, k 'AddressK XPub)
         , MkKeyFingerprint k Address
         , WalletKey k
-        , Bounded (Index (AddressIndexDerivationType k) 'AddressK)
         , HasDBFactory s k ctx
         , HasWorkerRegistry s k ctx
         , IsOurs s RewardAccount
@@ -776,7 +776,6 @@ postShelleyWallet
         , MkKeyFingerprint k (Proxy n, k 'AddressK XPub)
         , MkKeyFingerprint k Address
         , WalletKey k
-        , Bounded (Index (AddressIndexDerivationType k) 'AddressK)
         , HasDBFactory s k ctx
         , HasWorkerRegistry s k ctx
         , IsOurs s RewardAccount
@@ -816,6 +815,7 @@ postAccountWallet
         , HasWorkerRegistry s k ctx
         , IsOurs s RewardAccount
         , (k == SharedKey) ~ 'False
+        , Typeable n
         )
     => ctx
     -> MkApiWallet ctx s w
@@ -873,7 +873,7 @@ mkShelleyWallet ctx wid cp meta pending progress = do
     let available = availableBalance pending cp
     let total = totalBalance pending reward cp
     pure ApiWallet
-        { addressPoolGap = ApiT $ getState cp ^. #externalPool . #gap
+        { addressPoolGap = ApiT $ getGap $ getState cp ^. #externalPool
         , balance = ApiWalletBalance
             { available = coinToQuantity (available ^. #coin)
             , total = coinToQuantity (total ^. #coin)
@@ -1269,6 +1269,7 @@ postIcarusWallet
         , k ~ IcarusKey
         , HasWorkerRegistry s k ctx
         , PaymentAddress n IcarusKey
+        , Typeable n
         )
     => ctx
     -> ByronWalletPostData '[12,15,18,21,24]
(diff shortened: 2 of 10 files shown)


lib/core/bench/db-bench.hs
```diff
@@ -359,31 +359,26 @@ bgroupWriteSeqState db = bgroup "SeqState"
...
+mkSeqState numAddrs _ = s 
+    { internalPool = fillPool (internalPool s)
+    , externalPool = fillPool (externalPool s)
+    }
   where
-    addrs = [ force (mkAddress i j, Unused) | j <- [1..numAddrs] ]
+    s = mkSeqStateFromAccountXPub @'Mainnet
+        ourAccount purposeCIP1852 defaultAddressPoolGap
+    fillPool (SeqAddressPool pool0) = SeqAddressPool $
+        foldl' (\p ix -> AddressPool.update (gen ix) p) pool0 [1..numAddrs]
```
paweljakubas: shouldn't be [1..numAddrs - 1] ? we have the first I bet in pool0 ? :thinking:

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -873,7 +873,7 @@ mkShelleyWallet ctx wid cp meta pending progress = do
     let available = availableBalance pending cp
     let total = totalBalance pending reward cp
     pure ApiWallet
-        { addressPoolGap = ApiT $ getState cp ^. #externalPool . #gap
+        { addressPoolGap = ApiT $ getGap $ getState cp ^. #externalPool
```
paweljakubas, line 876: someone doesn't like nested lens ? ;-)  ---> kidding here, I see you want to reuse `getGap`

lib/core/src/Cardano/Wallet/DB/Sqlite/AddressBook.hs
```diff
@@ -99,56 +81,47 @@ instance AddressBookIso (Seq.SeqState n k)
...
-    ) => Seq.AddressPool c key -> SeqAddressList c -> Seq.AddressPool c key
-fromDiscoveries ctx (SeqAddressList addrs) =
-    Seq.mkAddressPool @n (Seq.context ctx) (Seq.gap ctx) addrs
-
--- | Remove all discovered addresses from an address pool,
--- but keep context.
-emptyPool :: Seq.AddressPool c key -> Seq.AddressPool c key
-emptyPool pool = pool{ Seq.indexedKeys = Map.empty }
+        from (Seq.SeqState int ext a b c d) =
+            let int0 = clear int
```
paweljakubas, line 97: what about introducing "AddressPool.emptyPool" instead of clearing?

lib/core/bench/db-bench.hs
paweljakubas: a ok, pool0 is an empty pool if I understand well. If yes, then everything's fine

lib/core/src/Cardano/Wallet/DB/Sqlite/AddressBook.hs
```diff
@@ -99,56 +81,47 @@ instance AddressBookIso (Seq.SeqState n k)
...
+            in  ( SeqPrologue $ Seq.SeqState int0 ext0 a b c d
+                , SeqDiscoveries (addresses int) (addresses ext)
+                )
+        to (SeqPrologue (Seq.SeqState int0 ext0 a b c d), SeqDiscoveries ints exts)
+          = Seq.SeqState (loadUnsafe int0 ints) (loadUnsafe ext0 exts) a b c d
+
+-- | Address data from sequential address pool.
+-- The phantom type parameter @c@ prevents mixing up
+-- the internal with the external pool.
+newtype SeqAddressMap (c :: Role) (key :: Depth -> Type -> Type) = SeqAddressMap
```
paweljakubas, line 108: :+1:

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
```diff
@@ -396,67 +392,48 @@ instance PersistAddressBook (Seq.SeqState n k)
...
-        void $ dbChunked insertMany_
-            [ SeqStateAddress wid sl addr ix UtxoInternal state
-            | (ix, (addr, state)) <- zip [0..] ints
-            ]
-        void $ dbChunked insertMany_
-            [ SeqStateAddress wid sl addr ix UtxoExternal state
-            | (ix, (addr, state)) <- zip [0..] exts
-            ]
+    insertDiscoveries wid sl (SeqDiscoveries ints exts) = do
+        insertSeqAddressMap @n wid sl ints
```
paweljakubas, line 438: indeed, the erased two structures begged for reuse. UtxoInternal and UntixExternal was the only difference between them

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
```diff
@@ -470,17 +447,34 @@ selectSeqStatePendingIxs wid =
...
-    :: forall c. Typeable c
-    => W.WalletId -> W.SlotNo -> SqlPersistT IO (SeqAddressList c)
-selectSeqAddressList wid sl = do
-    SeqAddressList . map (toPair . entityVal) <$> selectList
+insertSeqAddressMap
+    :: forall n c key. (PaymentAddress n key, Typeable c)
+    =>  W.WalletId -> W.SlotNo -> SeqAddressMap c key -> SqlPersistT IO ()
+insertSeqAddressMap wid sl (SeqAddressMap pool) = void $
+    dbChunked insertMany_
+        [ SeqStateAddress wid sl (liftPaymentAddress @n addr)
```
paweljakubas, line 474: correct here for payment key hash credential provided we have ONE staking credential (also key hash).

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
```diff
@@ -470,17 +447,34 @@ selectSeqStatePendingIxs wid =
...
+    ) => W.WalletId -> W.SlotNo -> SqlPersistT IO (SeqAddressMap c key)
+selectSeqAddressMap wid sl = do
+    SeqAddressMap . Map.fromList . map (toTriple . entityVal) <$> selectList
         [ SeqStateAddressWalletId ==. wid
         , SeqStateAddressSlot ==. sl
         , SeqStateAddressRole ==. Seq.role @c
         ] [Asc SeqStateAddressIndex]
   where
-    toPair x = (seqStateAddressAddress x, seqStateAddressStatus x)
+    toTriple x =
```
paweljakubas, line 491: :+1:

lib/core/src/Cardano/Wallet/Primitive/AddressDiscovery/Sequential.hs
```diff
@@ -163,12 +156,21 @@ import GHC.TypeLits
...
     ( Typeable, typeRep )
 
+import qualified Cardano.Wallet.Address.Pool as AddressPool
 import qualified Data.List as L
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Map.Strict as Map
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
 
+-- | Convenient alias for commonly used class contexts on keys.
```
paweljakubas:
```suggestion
-- | Convenient constraint alias for commonly used class contexts on keys.
```

lib/core/test/unit/Cardano/Wallet/Address/PoolSpec.hs
```diff
@@ -23,26 +25,26 @@ spec = do
...
-            forAll genUsageWithGap $ \usage ->
-                let pool = fromUsage usage
+            forAll (genUsageForGap $ AddressPool.gap testPool) $ \usage ->
+                let pool = fromUsage testPool usage
                 in  all ($ pool) [prop_gap, prop_fresh]
 
-        it "sequence of updates preserves invariants"
-            prop_updates
+        it "sequence of updates preserves invariants" $
+            prop_updates testPool
```
paweljakubas, line 38: what about property like 
1. get to some pool state s1
2. clear pool -> no addresses here in the pool
3. update to state s1 -> state here the same as in point 1

etc? (or it is automatically tested somewhere)
also properties using removeN  interleaved with ^^^

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/SequentialSpec.hs
```diff
@@ -160,35 +149,6 @@ spec = do
     parallel $ describe "DerivationPrefix" $ do
         textRoundtrip (Proxy @DerivationPrefix)
 
-    let styles =
-            [ Style (Proxy @'UtxoExternal)
-            , Style (Proxy @'UtxoInternal)
-            ]
-
-    parallel $ describe "AddressPool (Shelley)" $ do
```
paweljakubas, line 168: I am assuming all properties here, both for internal/external pools, and for both Shelley/Icarus keys are reproduced, ie. no property specified here is omitted?

lib/core/bench/db-bench.hs
HeinrichApfelmus: I think that you're right — `mkSeqStateFromAccountXPub` creates two pools with `defaultAddressPoolGap` many addresses already in them, and they start from index `0`, not `1`. So, `[0…(numAddrs-1)]` would be correct (in the case where `numAddrs` is larger than the pool gap). 🤔 Then again, this is the benchmarking module, which is somewhat obsolete due to caching being on by default now; we will need to revisit this properly anyway. But there is no harm in keeping the module reasonably tidy. 🙂

lib/core/src/Cardano/Wallet/DB/Sqlite/AddressBook.hs
HeinrichApfelmus: Well, `AddressPool.new` constructs an empty pool, but it needs a `generator` and a `gap` as arguments. The function `AddressPool.clear` essentially just calls `AddressPool.new` with these arguments taken from an existing pool. Unfolding the definition of `clear` here would yield

```haskell
let int0 = AddressPool.new (AddressPool.generator int) (AddressPool.gap int) 
```

I also find it unfortunate that the pool parameters (`gap`, `generator`), which belong to the `Prologue` part, are baked in the abstract data type together with the pool addresses, which belong to the `Discoveries` part, but I think that it's a necessary compromise to keep the invariants intact. At some point in the code, things will look a bit weird when trying to "unbake" the two, and, well, this is that point in the code.

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
HeinrichApfelmus: I didn't mind the reuse; the thing I like most about the new formulation (which was there before at some point) is that it ensures that e.g. `int` is paired with `UtxoInternal`. In contrast, the structures removed here would, in principle, permit switching `ints` and `exts` without type error.

Of course, trying to protect against every switch of variables via the type system would be overkill, but I feel that it has merit here.

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
HeinrichApfelmus: Hopefully. I just followed the previous code, which used `liftPaymentAddress @n` as argument of type `(KeyFingerprint "payment" k -> Address)` for the now-removed `addresses` function. So, if the previous code was correct, then so is the new one. 😄

lib/core/test/unit/Cardano/Wallet/Primitive/AddressDiscovery/SequentialSpec.hs
HeinrichApfelmus: Yes, these property tests have all been superseded by the tests in `Cardano.Wallet.Address.PoolSpec`.

Thanks to the fact that the new `Pool` type is polymorphic in the `addr`, we no longer have to specialize to different key types or chain types — we get this as a free theorem: If the pool works for `generator = id`, it works for every other address type.

lib/core/test/unit/Cardano/Wallet/Address/PoolSpec.hs
HeinrichApfelmus: Hm, not sure. In order to test that `update` respects the internal invariants (`prop_gap`, `prop_fresh`), I figured that I could pick any reasonably complex sequence of updates (here `addrs1` and `addrs2` in `prop_updates`). If the code contained a bug, then it would show up in this sequence already; I'm not sure if a more systematic construction of update sequences would yield more information. 🤔 I also figured that apart from the internal invariants, the code is more or less a copy of `Data.Map`, and that testing that the `Map` is tracked correctly would be overkill. 😅

I would prefer to keep `removeN` for shrinking only, as it uses `loadUnsafe` and would have to be tested separately. In this way, we do not have to test this function: A buggy shrinker affects our ability to find the root cause of bugs, but it does not give rise to false positives or negatives.

lib/core/src/Cardano/Wallet/DB/Sqlite/AddressBook.hs
paweljakubas: ok. thanks for the clarification. sounds reasonable

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
paweljakubas: yes, I just want to underline that we have fingerprint properly handled for enterprise and base addresses (with key hash credentials), and only for the situation of one staking credential;-)

lib/core/test/unit/Cardano/Wallet/Address/PoolSpec.hs
paweljakubas: indeed, testing `Map` is overkill, but maybe having roundtrip : state -> clear -> state no. Up to you. was just thinking loudly

paweljakubas (approved):
very nice simplification adopted. it is great that the PR resulted in -400 net loc. There are some merge conflicts to resolve before merging though.

lib/core/test/unit/Cardano/Wallet/Address/PoolSpec.hs
HeinrichApfelmus: > having roundtrip : state -> clear -> state 

You have convinced me — I have added a property `prop_updates_order` which creates a pool from a random sequence of updates and then compares this to a pool built from the `Used` addresses, but updated in the order of their indices. 🤓

HeinrichApfelmus:
bors merge

PR #3074, opened 2022-01-07
Fix gitRevFromGit with newer process libraries

angerman:
This code was initially lifted likely from cardano-node, and now lives in cardano-config, but is replicated here...
It was fixed in cardano config recently by @newhoggy in cardano-config#4. Now we need to fix it here as well.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -122,7 +122,7 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-config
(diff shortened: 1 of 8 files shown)


rvl (approved):
Thanks @angerman. I changed the whitespacing to match, and simplified the error handling path.

rvl:
Ok - bumped the cardano-config dep to a revision which includes the same fix.

PR #3076, opened 2022-01-11
Successfully depend on Plutus

sevanspowell:
We might include Plutus in our stack.yaml, but we don't actually use Plutus atm, so it hasn't been added to our build plan. If we try to add it, it fails. This PR attempts to fix that.


sevanspowell:
Already done

PR #3077, opened 2022-01-11
Split address discovery and UTxO updates in wallet model

HeinrichApfelmus:
### Issue number

ADP-1323

### Overview

In this step, we prepare the introduction of delta encodings in the "business logic" of the wallet, i.e. the `applyBlock` function. To facilitate keeping track of state changes, we split this function into two parts:

* `discoverAddresses` changes only the address discovery state `s` by looking at all transactions in the block.
* `applyBlockToUTxO` changes only the `UTxO` set, using the addresses discovered in the previous step.

By disentangling the changes to the wallet state, it also becomes more clear in which order the addresses within the block are discovered. Due to the nature of address discovery with an address gap, this order does matter!

### Details

* I have taken the liberty to rewrite and update the module commentary.
    * For example, the commentary to `applyOurTxToUTxO` was slightly misleading in that the function did *not* expect a precondition that the transaction `tx` is guaranteed to be relevant to the wallet. The main reason for this lack of precondition is that a `TxIn` only references an output, but does not record the address associated with that output -- thus, we cannot know whether a `TxIn` belongs to us unless we retrieve its address. In this case, we retrieve the address via the `UTxO` set that we have to carry around anyway.

### Comments

* The property tests for the wallet business logic are currently based on a small initial fragment of Mainnet. Since this is a core component of the wallet relating to asset balances, we may want to look into more elaborate generation of valid chains. Then again, combining `UTxO` sets is not rocket science, and in addition to careful code reading, the existing tests should be sufficient.
    * One idea for generating new chain fragments would be to randomize addresses within a given chain fragment.

### DIFF
--- a/lib/core/src/Cardano/Wallet/Primitive/Model.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Model.hs
@@ -54,7 +54,6 @@ module Cardano.Wallet.Primitive.Model
     , applyOurTxToUTxO
     , changeUTxO
     , discoverAddresses
-    , isOurTx
     ) where
 
 import Prelude
@@ -474,44 +473,6 @@ discoverAddresses block s0 = s2
 -- address discovery state.
 ours :: IsOurs s addr => s -> addr -> Bool
 ours s x = isJust . fst $ isOurs x s
-
--- | Performs address discovery and indicates whether a given transaction is
--- relevant to the wallet.
---
--- This function is only used for unit tests -- wallet model code uses the
--- 'ours' and 'updateOurs' variants.
-isOurTx
-    :: (IsOurs s Address, IsOurs s RewardAccount)
-    => Tx
-    -> UTxO
-    -> State s Bool
-isOurTx tx u
-    -- If a transaction has failed script validation, then the ledger rules
-    -- require that applying the transaction shall have no effect other than
-    -- to fully spend the collateral inputs included within that transaction.
-    --
-    -- Therefore, such a transaction is only relevant to the wallet if it has
-    -- one more collateral inputs that belong to the wallet.
-    --
-    | failedScriptValidation tx =
-        txHasRelevantCollateral
-    | otherwise =
-        F.or <$> sequence
-            [ txHasRelevantInput
-            , txHasRelevantOutput
-            , txHasRelevantWithdrawal
-            ]
-  where
-    txHasRelevantCollateral =
-        pure . not . UTxO.null $
-        u `UTxO.restrictedBy` Set.fromList (fst <$> tx ^. #resolvedCollateral)
-    txHasRelevantInput =
-        pure . not . UTxO.null $
-        u `UTxO.restrictedBy` Set.fromList (fst <$> tx ^. #resolvedInputs)
-    txHasRelevantOutput =
-        F.or <$> sequence (isOursState . (^. #address) <$> tx ^. #outputs)
-    txHasRelevantWithdrawal =
-        F.or <$> sequence (isOursState . fst <$> Map.toList (tx ^. #withdrawals))
 
 -- | Add an address to the address discovery state, iff it belongs to us.
 updateOurs :: IsOurs s addr => s -> addr -> s
--- a/lib/core/src/Cardano/Wallet/Primitive/Model.hs
+++ b/lib/core/src/Cardano/Wallet/Primitive/Model.hs
@@ -1,33 +1,29 @@
 {-# LANGUAGE BangPatterns #-}
-{-# LANGUAGE BlockArguments #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE FlexibleInstances #-}
-{-# LANGUAGE GADTs #-}
-{-# LANGUAGE LambdaCase #-}
-{-# LANGUAGE MultiWayIf #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedLabels #-}
-{-# LANGUAGE RankNTypes #-}
-{-# LANGUAGE ScopedTypeVariables #-}
 
 -- |
 -- Copyright: © 2018-2020 IOHK
 -- License: Apache-2.0
 --
--- Here we find the "business logic" to manage a Cardano wallet. This is a
--- direct implementation of the model from the [Formal Specification for a Cardano Wallet](https://github.com/input-output-hk/cardano-wallet/blob/master/specifications/wallet/formal-specification-for-a-cardano-wallet.pdf)
--- Note that, this module is purposedly agnostic to how blocks are retrieved or
--- how various types are serialized.
+-- This module implements the "business logic" to manage a Cardano wallet.
+-- It is a direct implementation of the model from the
+-- [Formal Specification for a Cardano Wallet](https://github.com/input-output-hk/cardano-wallet/blob/master/specifications/wallet/formal-specification-for-a-cardano-wallet.pdf).
 --
--- This is really about how the wallet keep track of its internal state, and its
--- UTxO (where the relationship is defined via the 'IsOurs' abstraction to allow
--- this core code to be used with any sort of derivation scheme).
+-- in other words, this module is about how the wallet keeps
+-- track of its internal state, specifically the 'UTxO' set and
+-- the address discovery state.
+-- This module is intentionally agnostic to specific address formats,
+-- and instead relies on the 'IsOurs' abstraction.
+-- It is also agnostic to issues such as how blocks are retrieved from the network,
+-- or how the state is serialized and cached in the local database.
 --
 -- All those functions are pure and there's no reason to shove in any sort of
--- side-effects in here :)
+-- side-effects in here. :)
 
 module Cardano.Wallet.Primitive.Model
     (
@@ -35,17 +31,11 @@ module Cardano.Wallet.Primitive.Model
       Wallet
 
     -- * Construction & Modification
-    , FilteredBlock (..)
     , initWallet
     , updateState
+    , FilteredBlock (..)
     , applyBlock
     , applyBlocks
-    , unsafeInitWallet
-    , applyTxToUTxO
-    , applyOurTxToUTxO
-    , utxoFromTx
-    , spendTx
-    , isOurTx
 
     -- * Accessors
     , currentTip
@@ -54,8 +44,20 @@ module Cardano.Wallet.Primitive.Model
     , totalBalance
     , totalUTxO
     , availableUTxO
-    , changeUTxO
     , utxo
+
+    -- * Internal
+    , unsafeInitWallet
+
+    -- * Testing
+    -- ** UTxO
+    , spendTx
+    , utxoFromTx
+    , applyTxToUTxO
+    , applyOurTxToUTxO
+    , changeUTxO
+    , discoverAddresses
+    , isOurTx
     ) where
 
 import Prelude
@@ -89,17 +91,11 @@ import Cardano.Wallet.Primitive.Types.Tx
     , txOutCoin
     )
 import Cardano.Wallet.Primitive.Types.UTxO
-    ( Dom (..), UTxO (..), balance, excluding, filterByAddressM )
+    ( Dom (..), UTxO (..), balance, excluding )
 import Control.DeepSeq
     ( NFData (..), deepseq )
-import Control.Monad
-    ( foldM )
-import Control.Monad.Extra
-    ( mapMaybeM )
 import Control.Monad.Trans.State.Strict
-    ( State, evalState, runState, state )
-import Data.Functor
-    ( (<&>) )
+    ( State, evalState, state )
 import Data.Generics.Internal.VL.Lens
     ( (^.) )
 import Data.Generics.Labels
@@ -122,29 +118,31 @@ import GHC.Generics
 import qualified Cardano.Wallet.Primitive.Types.TokenBundle as TB
 import qualified Cardano.Wallet.Primitive.Types.UTxO as UTxO
 import qualified Data.Foldable as F
+import qualified Data.List as L
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Map as Map
 import qualified Data.Set as Set
 
 {-------------------------------------------------------------------------------
-                                     Type
+    Type
 -------------------------------------------------------------------------------}
 
--- | An opaque wallet type, see 'initWallet', 'updateState', 'applyBlock', and
--- 'applyBlocks' to construct and update wallets.
+-- | Abstract data type representing a wallet state.
 --
--- Internally, this keeps track or a few things including:
+-- A 'Wallet' keeps track of transaction outputs and associated addresses
+-- that belong to /us/ -- we are able to spend these outputs because
+-- we know the secret keys belonging to the addresses.
+-- This information is associated to a particular point on the blockchain.
+--
+-- Internally, a 'Wallet' keeps track of
 --
 --  - UTxOs
---  - Transaction history
 --  - Known & used addresses, via address discovery state
---  - Blockchain parameters
+--  - The associated 'BlockHeader' indicating the point on the chain.
 --
 -- The 'Wallet' is parameterized over a single type:
 --
--- - @s@ is a /state/ used to keep track of known addresses. The business logic
---   doesn't know how to answer the question \"Is this address ours?\", so we
---   expect this state to be able to answer that for us.
+-- - @s@ is a /state/ used to keep track of known addresses.
 --   Typically, this state will be an instance of the 'IsOurs' class,
 --   e.g. @'IsOurs' s 'Address'@.
 --
@@ -179,10 +177,10 @@ instance Buildable s => Buildable (Wallet s) where
         <> indentF 4 (build s)
 
 {-------------------------------------------------------------------------------
-                          Construction & Modification
+    Construction & Modification
 -------------------------------------------------------------------------------}
 
--- | Create a an empty wallet and apply the given genesis block
+-- | Create a an empty wallet and apply the given genesis block.
 --
 -- The wallet tip will be set to the header of the applied genesis block.
 initWallet
@@ -192,13 +190,14 @@ initWallet
     -> s
         -- ^ Initial address discovery state
     -> ([(Tx, TxMeta)], Wallet s)
-initWallet block s =
-    let
-        ((FilteredBlock _ txs, u), s') = applyBlockToUTxO block mempty s
-    in
-        (txs, Wallet u (header block) s')
+initWallet block0 s = (transactions, w1)
+  where
+    w0 = Wallet mempty undefined s
+    (FilteredBlock{transactions}, w1) = applyBlock block0 w0
 
--- | Constructs a wallet from the exact given state. Using this function instead
+-- | Construct a wallet from the exact given state.
+--
+-- Using this function instead
 -- of 'initWallet' and 'applyBlock' allows the wallet invariants to be
 -- broken. Therefore it should only be used in the special case of loading
 -- wallet checkpoints from the database (where it is assumed a valid wallet was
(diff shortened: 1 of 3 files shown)


HeinrichApfelmus:
bors try

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -1,51 +1,41 @@
...
 -- |
 -- Copyright: © 2018-2020 IOHK
 -- License: Apache-2.0
 --
--- Here we find the "business logic" to manage a Cardano wallet. This is a
--- direct implementation of the model from the [Formal Specification for a Cardano Wallet](https://github.com/input-output-hk/cardano-wallet/blob/master/specifications/wallet/formal-specification-for-a-cardano-wallet.pdf)
--- Note that, this module is purposedly agnostic to how blocks are retrieved or
--- how various types are serialized.
+-- This module implements the "business logic" to manage a Cardano wallet.
+-- It is a direct implementation of the model from the
```
paweljakubas:
```suggestion
-- It is a direct implementation of the model, with extensions, from the
```

paweljakubas: you know, in the paper ^^^ there was not notion of for example, withdrawal, defined or even mentioned ;-)

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -1,51 +1,41 @@
...
+-- [Formal Specification for a Cardano Wallet](https://github.com/input-output-hk/cardano-wallet/blob/master/specifications/wallet/formal-specification-for-a-cardano-wallet.pdf).
 --
--- This is really about how the wallet keep track of its internal state, and its
--- UTxO (where the relationship is defined via the 'IsOurs' abstraction to allow
--- this core code to be used with any sort of derivation scheme).
+-- in other words, this module is about how the wallet keeps
```
paweljakubas:
```suggestion
-- In other words, this module is about how the wallet keeps
```

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -122,29 +118,31 @@ import GHC.Generics
...
+    Type
 -------------------------------------------------------------------------------}
 
--- | An opaque wallet type, see 'initWallet', 'updateState', 'applyBlock', and
--- 'applyBlocks' to construct and update wallets.
+-- | Abstract data type representing a wallet state.
+--
+-- A 'Wallet' keeps track of transaction outputs and associated addresses
+-- that belong to /us/ -- we are able to spend these outputs because
+-- we know the secret keys belonging to the addresses.
```
paweljakubas:
```suggestion
-- we know the corresponding signing key belonging to the output. Hence, we are able to produce witness engaging those outputs as they become inputs in forthcoming transactions according to UTxO model. 
```

paweljakubas: maybe something like this is more precise

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -179,10 +177,10 @@ instance Buildable s => Buildable (Wallet s) where
         <> indentF 4 (build s)
 
 {-------------------------------------------------------------------------------
-                          Construction & Modification
+    Construction & Modification
 -------------------------------------------------------------------------------}
 
--- | Create a an empty wallet and apply the given genesis block
+-- | Create a an empty wallet and apply the given genesis block.
```
paweljakubas:
```suggestion
-- | Create an empty wallet and apply a given genesis block.
```

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -223,30 +222,33 @@ updateState s (Wallet u tip _) = Wallet u tip s
 -- | Represents the subset of data from a single block that are relevant to a
 --   particular wallet, discovered when applying a block to that wallet.
 data FilteredBlock = FilteredBlock
-    { delegations :: ![DelegationCertificate]
-        -- ^ Stake delegations made on behalf of the wallet, listed in order of
-        -- discovery. If the list contains more than element, those that appear
+    { transactions :: ![(Tx, TxMeta)]
+        -- ^ The set of transactions that affect the wallet,
+        -- list in the order in which they appears in the block.
```
paweljakubas:
```suggestion
        -- list in the order in which they appear in the block.
```

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -223,30 +222,33 @@ updateState s (Wallet u tip _) = Wallet u tip s
...
+    , delegations :: ![DelegationCertificate]
+        -- ^ Stake delegations made on behalf of the wallet,
+        -- listed in the order in which they appear on chain.
```
paweljakubas:
```suggestion
        -- listed in the order in which they appear on the chain.
```

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -354,14 +353,44 @@ totalUTxO pending (Wallet u _ s) =
...
 
+-- | Retrieve the change 'UTxO' contained in a set of pending transactions.
+--
+-- We perform /some/ address discovery within the list of pending addresses,
+-- but we do not store the result.
+-- Instead, we essentially assume that the address discovery state @s@ contains
+-- enough information to collect the change addresses in the pending
+-- transactions.
+--
+-- Cave:
```
paweljakubas: what do you mean by 'Cave' - limitations?

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -393,12 +425,13 @@ spendTx tx !u =
         then collateralInputs tx
         else inputs tx
 
--- | Construct a UTxO corresponding to a given transaction. It is important for
--- the transaction outputs to be ordered correctly, since they become available
--- inputs for the subsequent blocks.
+-- | Construct a 'UTxO' corresponding to a given transaction.
+--
+-- It is important for the transaction outputs to be ordered correctly,
+-- as they become available inputs for subsequent blocks.
```
paweljakubas: but we have TxIn basically having TxId and index and it is enough to point correctly to any input, right?

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -411,58 +444,43 @@ utxoFromUnvalidatedTx :: Tx -> UTxO
...
+-- | Perform address discovery by going through all transactions
+-- and delegation certificates in the block.
+discoverAddresses
+    :: (IsOurs s Address, IsOurs s RewardAccount)
+    => Block -> s -> s
+discoverAddresses block s0 = s2
+  where
+    -- NOTE: Order in which we perform discovery is important.
+    s1 = L.foldl' discoverCert s0 (block ^. #delegations)
+    s2 = L.foldl' discoverTx   s1 (block ^. #transactions)
```
paweljakubas, line 459: indeed, left fold is correct here :+1:

paweljakubas:
bors try

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -411,58 +444,43 @@ utxoFromUnvalidatedTx :: Tx -> UTxO
...
+
+    discoverCert s cert = updateOurs s (dlgCertAccount cert)
+
+    -- NOTE: Only outputs and withdrawals can introduce new addresses.
```
paweljakubas:
```suggestion
    -- NOTE: Only outputs and withdrawals can potentially result in extension of pool gap, so knowledge of new addresses.
```

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -411,58 +444,43 @@ utxoFromUnvalidatedTx :: Tx -> UTxO
...
+    -- Inputs and collateral are forced to use existing addresses.
+    discoverTx s tx = discoverWithdrawals (discoverOutputs s tx) tx
+    discoverOutputs s tx =
+        L.foldl' (\s_ out -> updateOurs s_ (out ^. #address)) s (tx ^. #outputs)
+    discoverWithdrawals s tx =
+        L.foldl' updateOurs s $ Map.keys (tx ^. #withdrawals)
+
+-- | Helper based on 'isOurs'.
+updateOurs :: IsOurs s entity => s -> entity -> s
```
paweljakubas: is updateOurs used elsewhere? If not then maybe we can put it  where of discoverAddresses

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -490,99 +508,118 @@ isOurTx tx u
...
+-- that the address discovery state @s@ identifies all our output addresses
+-- in the given 'Block'.
+--
+-- A 'FilteredBlock' is returned in addition to the new 'UTxO' set.
+-- This 'FilteredBlock' includes those transactions and delegations
+-- that are in the given 'Block' and that are also relevant to the wallet,
+-- i.e. they have
+-- 
+-- * Outputs with known addresses
+-- * Inputs referrring to known outputs.
```
paweljakubas, line 534:
```suggestion
-- * Inputs referrring to known outputs of previous transactions.
```

paweljakubas:
```suggestion
-- * Inputs referring to known outputs of previous transactions.
```

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -490,99 +508,118 @@ isOurTx tx u
...
+-- 
+-- In practice, most transactions that are relevant have a change output to the
```
paweljakubas: is it true? If the transaction has our inputs then indeed we have change also, but what if we are recepient wallet, then change is not ours almost always, right? Exception would be transactions to our own wallet

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -490,99 +508,118 @@ isOurTx tx u
...
-    -> UTxO
     -> s
-    -> ((FilteredBlock, UTxO), s)
-applyBlockToUTxO b u0 = runState $ do
-    delegations <- mapMaybeM ourDelegation (b ^. #delegations)
-    (transactions, ourU) <- foldM applyOurTx (mempty, u0) (b ^. #transactions)
-    return (FilteredBlock {delegations, transactions}, ourU)
+    -> UTxO
+    -> (FilteredBlock, UTxO)
+applyBlockToUTxO block s u0 =
```
paweljakubas: don't we want to use bangs here?

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -594,40 +631,24 @@ applyOurTxToUTxO !slotNo !blockHeight !tx !prevUTxO = do
...
+    in if hasKnownWithdrawal || prevUTxO /= ourNextUTxO
+        then Just ((tx {fee = actualFee dir}, txmeta), ourNextUTxO)
+        else Nothing
+
+ourWithdrawalSumFromTx
+    :: IsOurs s RewardAccount
+    => s -> Tx -> Coin
+ourWithdrawalSumFromTx s tx
+    -- If a transaction has failed script validation, then the ledger rules
+    -- require that applying the transaction shall have no effect other than
```
paweljakubas, line 619:
```suggestion
    -- require that applying the transaction shall have an affect only on collateral inputs specified in that transaction.
```

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -594,40 +631,24 @@ applyOurTxToUTxO !slotNo !blockHeight !tx !prevUTxO = do
...
+    -- to fully spend the collateral inputs included within that transaction.
+    --
+    -- Therefore, any reward withdrawals included in such a transaction should
+    -- also have no effect.
+    --
+    | failedScriptValidation tx =
+        Coin 0
+    | otherwise =
+        F.foldMap snd
+        . filter (ours s . fst)
```
paweljakubas: I wonder if we could be more performant here :thinking:

paweljakubas (approved):
Looks like valid and correct splitting of concerns between updating addresses discovery and utxo set upon applying of a transaction. To my best knowledge it should behave as before this refactoring. now it is more clear what happens. 

Would be good if @Anviking or @rvl or @jonathanknowles also look at this before merging

lib/core/src/Cardano/Wallet/Primitive/Model.hs
HeinrichApfelmus: You mean like `foldMap` on the map directly, without `toList` and `filter` in order to avoid building intermediate Maps?

paweljakubas: possibly, Map has Foldable instance ... we have Map.filterWithKey. just food for thought

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -299,14 +301,11 @@ totalBalance pending rewards wallet@(Wallet _ _ s) =
...
-        | hasPendingWithdrawals pending = mempty
+        | hasPendingWithdrawals = mempty
         | otherwise = TB.fromCoin rewards
 
     hasPendingWithdrawals =
-        anyS (anyM (\acct _ -> isJust $ fst (isOurs acct s)) . withdrawals)
-      where
-        anyS predicate = not . Set.null . Set.filter predicate
-        anyM predicate = not . Map.null . Map.filterWithKey predicate
+        any (any (ours s) . Map.keys . withdrawals) pending
```
rvl, line 304: Much nicer

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -54,8 +43,19 @@ module Cardano.Wallet.Primitive.Model
     , totalBalance
     , totalUTxO
     , availableUTxO
-    , changeUTxO
     , utxo
+
+    -- * Internal
+    , unsafeInitWallet
+
+    -- * Testing
```
rvl: These functions aren't only for testing - some are used by `applyBlock`. So they ought to go under the "Internal" heading I think.

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -411,178 +444,152 @@ utxoFromUnvalidatedTx :: Tx -> UTxO
...
-        pure . not . UTxO.null $
-        u `UTxO.restrictedBy` Set.fromList (fst <$> tx ^. #resolvedCollateral)
-    txHasRelevantInput =
-        pure . not . UTxO.null $
-        u `UTxO.restrictedBy` Set.fromList (fst <$> tx ^. #resolvedInputs)
-    txHasRelevantOutput =
-        F.or <$> sequence (isOurAddress . (^. #address) <$> tx ^. #outputs)
-    txHasRelevantWithdrawal =
-        F.or <$> sequence (isOurWithdrawal <$> Map.toList (tx ^. #withdrawals))
+    -- NOTE: Order in which we perform discovery is important.
```
rvl, line 457: Because order of state updates is important, perhaps using `do` notation would be clearer?

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -814,12 +816,64 @@ prop_applyOurTxToUTxO_someOurs ourState slotNo blockHeight tx utxo =
...
+-- relevant to the wallet.
+--
+-- This function is only used for unit tests -- wallet model code uses the
+-- 'ours' and 'updateOurs' variants.
+isOurTx
+    :: (IsOurs s Address, IsOurs s RewardAccount)
+    => Tx
+    -> UTxO
+    -> State s Bool
+isOurTx tx u
```
rvl, line 846: I moved this into ModelSpec, since it's only used to check a property.

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -341,27 +339,58 @@ totalUTxO
...
+-- enough information to collect the change addresses in the pending
+-- transactions.
+--
+-- Caveats:
+-- * Rollbacks can invalidate this assumption. 🙈
+-- * The order of pending transactions is based on transaction hashes,
+--   and typically does not agree with the order in which we have submitted
+--   them onto the chain. Hence, the address discovery phase is not really
+--   very effective.
+--   TODO: Add slot to 'Tx' and sort the pending set by slot.
```
rvl, line 367: I added this TODO.

rvl (approved):
Looks good to me, thanks for improving the Haddocks too.

I also took some liberties with your branch wrt Haddocks, some variable names, etc.

lib/core/src/Cardano/Wallet/Primitive/Model.hs
rvl: I was thinking the same, but `updateOurs` is sort of the sibling of `ours` which needs to be at the top level.

HeinrichApfelmus: > `updateOurs` is sort of the sibling of `ours` which needs to be at the top level.

Yes. I actually intend to move these functions into the `IsOurs` class in some future PR.

lib/core/src/Cardano/Wallet/Primitive/Model.hs
HeinrichApfelmus: I think that in the sentence, "change" is intended to refer to any output that goes to our wallet, not just the "proper" change outputs (with /1 derivation). I will reword this for clarity.

lib/core/src/Cardano/Wallet/Primitive/Model.hs
HeinrichApfelmus: Unfortunately, the shotgun approach to using bangs usually does not improve performance. (The best place to put bangs is on record fields of data types anyway.)

In this case, it's the responsibility of `applyBlockToUTxO` to make sure that the accumulator is evaluated to WHNF; this is indeed the case.

That said, I soon intend to take another look at whether any of our long-lasting state is in normal form.

lib/core/src/Cardano/Wallet/Primitive/Model.hs
HeinrichApfelmus: I'll make "Exported for testing" a subheading of "Internal".

lib/core/src/Cardano/Wallet/Primitive/Model.hs
HeinrichApfelmus: We not only perform operations in sequence, but also need to collect the results with `F.or`. In order to collect results with `do`, we would need to bring new identifiers into scope, but they would be boilerplate. The present form is more succinct.

Anyway, since `isOurTx` is used for testing only, I would leave it as it is.

lib/core/src/Cardano/Wallet/Primitive/Model.hs
HeinrichApfelmus: Yes, I meant the latin "beware!". Rodney changed it to "Caveats", which is better.

lib/core/src/Cardano/Wallet/Primitive/Model.hs
HeinrichApfelmus: Yes, it's just that this index depends on the order of outputs in the list of outputs — that's all what this comment intends to say.

lib/core/src/Cardano/Wallet/Primitive/Model.hs
HeinrichApfelmus: I prefer the original wording (which is from before my time, I only moved the function around), as it strongly emphasizes that a transaction which has failed validation is mostly void.

HeinrichApfelmus:
bors try

HeinrichApfelmus:
bors merge

HeinrichApfelmus:
bors merge

HeinrichApfelmus:
bors retry

PR #3078, opened 2022-01-11
New submit tx ep testing

piotr-iohk:
- [x] Small fix to swagger
- [x] I have updated existing e2e tests with new submit ep
- [x] Add some more tests submit ep specific

### Issue Number

ADP-1171


paweljakubas (approved):
let's merge to my branch

PR #3080, opened 2022-01-13
cabal.project: Add upper bound on hashable package version

rvl:
This fixes a cabal build error due to an API change on the Hashable
class.

It also adds some more constraints found in `cardano-node/cabal.project`.

### Issue Number

ADP-1263


piotr-iohk (approved):
OK

PR #3083, opened 2022-01-17
Add nixos service and scripts

rvl:
- [x] Adds a NixOS service module for cardano-wallet.
- [x] Adds a NixOS test for the service.
- [x] Documentation on how to use the NixOS service.

### Comments

This PR is @jbgi's work and was split off from #2997.

### Issue Number

ADP-1357

### DIFF
--- a/nix/scripts.nix
+++ b/nix/scripts.nix
@@ -1,3 +1,13 @@
+################################################################################
+# Wallet service wrapper script
+#
+# This script runs the cardano-wallet server preconfigured with
+# a specific network.
+#
+# It is used by devops to run the wallet in Nomad clusters.
+#
+################################################################################
+
 { evalService
 , project
 , customConfigs
--- a/nix/scripts.nix
+++ b/nix/scripts.nix
@@ -1,0 +1,76 @@
+{ evalService
+, project
+, customConfigs
+}:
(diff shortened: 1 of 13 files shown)


nix/scripts.nix
```diff
@@ -0,0 +1,76 @@
+{ evalService
```
rvl, line 11: I'm not so sure yet about these scripts.
Our users would like to have convenience scripts for connecting to mainnet/testtnet, but possibly in a way where nix is optional.
Can we just do the NixOS service in this PR, and review these convenience scripts in a new PR?

rvl (commented):
Looks great @jbgi.
I can't :heavy_check_mark: because I opened this PR.

nix/scripts.nix
jbgi: I think those scripts can/are also to be used as entry-point for bitte clusters containers, (cc @blaggacao for confirmation) so would be needed as part of the flake.

disassembler: Yes, these scripts are required to run the wallet in nomad clusters, such as ERC20 converter.

rvl: OK, thanks @jbgi, @disassembler.

PR #3084, opened 2022-01-18
docs: Add info about how WalletId is calculated

rvl:
@blaggacao was asking about this, so here's a doc.

### Issue Number

[ERC-34](https://input-output.atlassian.net/browse/ERC-34)

### DIFF
--- a/docs/design/WalletId.md
+++ b/docs/design/WalletId.md
@@ -1,0 +1,53 @@
+---
+title: WalletId
+page:
+  headHtml: |
+    <snippet var="js.prism" />
+    <snippet var="js.mathjax" />
+---
+
+The `WalletId` is a hexadecimal string derived from the wallet's mnemonic.
+
+It is used by the `cardano-wallet` server to refer to specific wallets.
+
+For all wallet types, the `WalletId` is a _blake2b-160_ hash of
+something. This hash function produces a 20-byte digest, which becomes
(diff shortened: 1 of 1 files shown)


docs/design/WalletId.md
```diff
@@ -0,0 +1,53 @@
...
+
+```bash
+#!/usr/bin/env bash
+
+xargs \
+  | cardano-address key from-recovery-phrase Shelley \
+  | cardano-address key public --with-chain-code \
+  | bech32 \
+  | python3 -c 'import hashlib,sys; print(hashlib.blake2b(bytes.fromhex(sys.stdin.read()), digest_size=20).hexdigest())'
+```
```
blaggacao: I'd personally prefer a plain CLI command since it's (a little bit) easier to "pin" in a nix shell.

```suggestion
xargs \
  | cardano-address key from-recovery-phrase Shelley \
  | cardano-address key public --with-chain-code \
  | bech32 \
  | xxd -r -p \
  | b2sum -a blake2b -l 160
```
```

rvl: Yes fair enough. I needed to tweak the `b2sum` command to work in my environment (b2sum (GNU coreutils) 8.32).

blaggacao: Oh yeah. That slipped my attention, but I would have had to, too. Thx!

PR #3087, opened 2022-01-20
docs: Fix links on CLI manual

rvl:
- It needed some tricky HTML coding to fix the links.
- The github actions workflow for publishing docs also needs to be updated.

### Issue Number

ADP-1358

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -11,18 +11,21 @@ jobs:
   docs:
     runs-on: ubuntu-20.04
     steps:
-      - name: '📥 Checkout repository'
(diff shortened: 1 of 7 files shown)


piotr-iohk:
Thanks! do we have rendered version somewhere with the fix?

piotr-iohk (approved):
Ok, it would be nice maybe also to have a sort of "go back" link at the each command description to ease the navigation even more.

rvl:
Thanks @piotr-iohk - I put back the "top" navigation links.

Adding a rendered version of docs for PRs is a good idea.

PR #3092, opened 2022-01-21
Factor out guards into helpers in balanceTransaction

Anviking:
- [x] Factor out guards in balanceTransaction into helpers for readability

### Comments

Preparation for other work.

### Issue Number

ADP-1372

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -1435,75 +1435,10 @@ balanceTransaction
     let (outputs, txWithdrawal, txMetadata, txAssetsToMint, txAssetsToBurn)
             = extractFromTx partialTx
 
-    -- Coin selection does not support pre-defining collateral. In Sep 2021
-    -- consensus was that we /could/ allow for it with just a day's work or so,
-    -- but that the need for it was unclear enough that it was not in any way
-    -- a priority.
-    case Cardano.txInsCollateral bod of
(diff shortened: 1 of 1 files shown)


lib/core/src/Cardano/Wallet.hs
```diff
@@ -1684,6 +1619,80 @@ balanceTransaction
...
+                Cardano.TxWithdrawals _ wdrls -> Set.size
+                    (Set.fromList $ map networkOfWdrl wdrls) > 1
+        when conflictingWdrlNetworks $
+            throwE $ ErrBalanceTxNotYetSupported ConflictingNetworks
+
+    guardExistingCollateral = do
+        -- Coin selection does not support pre-defining collateral. In Sep 2021
+        -- consensus was that we /could/ allow for it with just a day's work or so,
+        -- but that the need for it was unclear enough that it was not in any way
+        -- a priority.
```
jonathanknowles, line 1689: I might be wrong, but suspect it would not just be "a day's work" 😄 .

jonathanknowles (approved):

lib/core/src/Cardano/Wallet.hs
Anviking: 😄 I thought that was what you said

Anviking:
Thanks!
bors r+

Anviking:
Bors r+

PR #3096, opened 2022-01-24
Add delegation to construct tx

paweljakubas:
ADP-1190

- [x] Handling joining in constructTransaction
- [x] Enable joining integration tests
- [x] Support rejoining stake pool
- [x] Support quitting stake pool
- [x] Add multidelegation error
- [x] Tackle technical debt concerning certs when adding witness, for both new and old tx workflow
- [x] Add multiaccount error


lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -494,26 +495,33 @@ newTransactionLayer networkId = TransactionLayer
                     mkTx networkId payload ttl stakeCreds keystore wdrl
                         selection delta
 
-    , addVkWitnesses = \_era stakeCreds addressResolver inputResolver sealedTx -> do
+    , addVkWitnesses = \_era usingStakeCreds stakeCreds addressResolver inputResolver sealedTx -> do
```
Anviking: Is it not possible to use the existing `resolveRewardAcct` instead of adding `usingStakeCreds`?

I.e. add a `mapMaybe mkCertWitness certs` in the same style as:

https://github.com/input-output-hk/cardano-wallet/blob/5ad348d26313363bacec061ee963c89fd9d467b8/lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs#L411-L412

I'm worried `usingStakeCreds` makes the responsibility of the function unclear, entangles it with server-handler logic making it difficult to test in the style of Sam's https://github.com/input-output-hk/cardano-wallet/blob/master/lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs#L380-L394

paweljakubas: ok, so I need to sneak information whether during signing there are OUR delegations in tx. If yes then I need to add additional witness. If not then I don't need to add them, otherwise I will have bloated payload and coin selection give rise to too small fees. Because we have right now, construction and signing decoupled then I need to provide this info either at the server level (and I did it as it was straightforward using decodeTransaction) or I can do it probably at shelley's transaction level. At this stage I just wanted to solve this issue of picking additional witness and make all tests green. But we can try to improve it :+1:  Let me think for a while

paweljakubas: so that would mean to add those certs during  construcTransaction and then only capture them during signing? :thinking:

paweljakubas: it seems as redundant argument to signTransaction to have (body,wits). It should be just body and wits lookup in body under certificates

paweljakubas: In the end I left `(body, wits)` as we will need it for partial signing. `wits` could be wits from other parties...So we need to leave it. Nevertheless, now old tx workflow and new one works, we discover reward account from selaedTx, and translate them into XPrv and pwd when ours. No more artificial sneaking of delegs wits

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -437,6 +439,8 @@ signTransaction networkId resolveRewardAcct resolveAddress resolveInput (body, w
             | (addr, _) <- fromCardanoWdrls $ Cardano.txWithdrawals bodyContent
             ]
 
+        certs = take 1 $ fromCardanoCerts $ Cardano.txCertificates bodyContent
```
paweljakubas: `take 1` is needed as we can have stake credential registration and joining stake pool certs together and they contribute to the same witness! So we need only one

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -487,10 +491,7 @@ newTransactionLayer networkId = TransactionLayer
                 withShelleyBasedEra era $ do
                     let stakeXPub = toXPub $ fst stakeCreds
                     let certs = mkDelegationCertificates action stakeXPub
-                    let mkWits unsigned =
```
paweljakubas, line 490: no need anymore for old tx workflow

paweljakubas:
bors try

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
Anviking: 😬 `signTransaction` should work with any transaction. Not just the ones where the cert that needs signing is the first one.

You shouldn't need to worry about duplicate key witnesses as they are represented as a set in the [ledger](https://github.com/input-output-hk/cardano-ledger/blob/ba20b1f7546b029750b52d091a1234ea4922bf90/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxWitness.hs#L167-L169). If you'd like to be pedantic and the `Eq` instance is available, you could `nub` the result of `mapMaybe mkWdrlCertWitness certs`.

Anviking: Besides, a stake credential registration shouldn't require a witness… which actually, would need special handling 🤔  

From the [shelley ledger spec](https://hydra.iohk.io/job/Cardano/cardano-ledger-specs/shelleyLedgerSpec/latest/download-by-type/doc-pdf/ledger-spec):
Skärmavbild 2022-01-27 kl  11 04 31





So we should probably change
```
        Cardano.StakeAddressRegistrationCertificate cred ->
            toRewardAccount cred
```
to return `Nothing`.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2140,20 +2142,45 @@ constructTransaction ctx genChange (ApiT wid) body = do
...
+                        poolStatus <- liftIO (getPoolStatus pid)
+                        pools <- liftIO knownPools
+                        curEpoch <- getCurrentEpoch ctx
+                        (action, _) <- liftHandler
+                            $ W.joinStakePool @_ @s @k wrk curEpoch pools pid poolStatus wid
+                        pure action
+                    [(Leaving _)] ->
+                        liftHandler $ W.quitStakePool @_ @s @k wrk wid
+                    _ ->
+                        liftHandler $ throwE ErrConstructTxMultidelegationNotSupported
```
Anviking, line 2182: 👍

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
paweljakubas: ok, you are right. Great catch! I forgot that stake credential does NOT need witness! Also `take 1` on this level is wrong. We could have in tx delegation certs that do not belong to the wallet, and are "inserted" before the wallet's cert. I will push soon. Need to test locally first

Anviking: There's also some [other certs](https://github.com/input-output-hk/cardano-node/blob/4f65fb9a27aa7e3a1873ab4211e412af780a3648/cardano-api/src/Cardano/Api/Certificate.hs#L116-L119), including pool reg certs which might be very interesting to support, but seems fine to leave out for now.

paweljakubas: yes, we handled them in decodeTx. But here we are interested into extracting stake credentials, and then answer the question whether they are our reward accounts, and if yes then add witness

lib/shelley/src/Cardano/Wallet/Shelley/Compatibility.hs
```diff
@@ -1277,6 +1278,25 @@ fromCardanoWdrls = \case
             , fromCardanoLovelace coin
             )
 
+fromCardanoCerts
+    :: Cardano.TxCertificates build era
+    -> [W.RewardAccount]
+fromCardanoCerts = \case
```
Anviking: Because this function is not really just converting a "Cardano cert" to a "wallet cert", could we find a more descriptive name? Perhaps `cardanoCertKeysForWitnesses` or `cardanoCertsRequiringWitnesses`… or if you can think of something better...

paweljakubas: well, these not cert keys requiring wits but CANDIDATES for that. And THE KEYS when they happen to include the wallet's reward account as stake credentials. We can have in tx delegation certs belonging to other wallets. I renamed to `cardanoCertKeysForWitnesses`

paweljakubas:
@Anviking I run locally:
1. stack test cardano-wallet-core:unit
2. stack test cardano-wallet:integration
and they passes

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -1053,30 +1063,250 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
...
             @[ApiStakePool]
             ctx (Link.listStakePools arbitraryStake) Empty
 
-        let delegation = Json [json|{
+        let delegationJoin = Json [json|{
                 "delegations": [{
                     "join": {
-                        "pool": #{pool'},
+                        "pool": #{pool1},
                         "stake_key_index": "0H"
```
Anviking: Unrelated: I didn't know `stake_key_index` was a field. Is this the account index in `1852'/1815'/0'/2/0`?

Is it really a good idea to both have it required and to ignore it? Users could pass in the wrong value today, and then suddenly it breaks when the wallet starts respecting `stake_key_index`.

Although I know the endpoint is `status: under development`.

lib/shelley/src/Cardano/Wallet/Shelley/Transaction.hs
```diff
@@ -437,6 +439,8 @@ signTransaction networkId resolveRewardAcct resolveAddress resolveInput (body, w
             | (addr, _) <- fromCardanoWdrls $ Cardano.txWithdrawals bodyContent
             ]
 
+        certs = cardanoCertKeysForWitnesses $ Cardano.txCertificates bodyContent
```
Anviking, line 442: 👍

Anviking (approved):
Great; thanks! lgtm.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
paweljakubas: so the idea was to have multiaccount delegation supported. Of course now, we only have "0H" derivation segment supported, as the account is "0H". I will add error and integration test for that. Also add comment in swagger. Nevertheless, having this capability with non-changing API as multiaccount comes might be reasonable.

paweljakubas:
I had two times:
```
Failures:

  src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs:2590:54: 
  1) API Specifications, NEW_SHELLEY_TRANSACTIONS, Plutus scenarios, currency
       uncaught exception: RequestException
       ClientError (Object (fromList [("code",String "redeemer_script_failure"),("message",String "I was unable to assign execution units to one of your redeemers: minting(6a14cb153ab369cb78fcb6d4c2219d8998c01215dc0152479595fe18); Its execution is failing with the following error: ValidationFailedV1 (CodecError (DeserialiseFailure 3280 \"NotEnoughSpace (0x0000004202a1dce0,S {currPtr = 0x0000004202a1dcdf, usedBits = 6})\")) [].")]))
```

Looks like a lack of resources ... 
When run locally
```
$ stack test --ta '-m "Plutus scenarios"' cardano-wallet:integration
...
      ping-pong (27255ms)
      game state-machine (22185ms)
      mint-burn (12331ms)
      withdrawal (19871ms)
      currency (1592ms)

Finished in 83.2763 seconds, used 10.6713 seconds of CPU time
5 examples, 0 failures
```

@Anviking @rvl @sevanspowell  Do you understand?

paweljakubas:
when I run locally ALL tests  
```
stack test cardano-wallet:integration
```
I encounter also this error, ALSO 
```
TRANS_NEW_JOIN_01a - Can join stakepool, rejoin another and quit FAILED [1]
```
missing pool :thinking:

paweljakubas:
bors retry

paweljakubas:
bors try

piotr-iohk:
@paweljakubas, As far as I see `deposit` is empty when making delegation for the first time, and it should show 2 ADA.

```
curl -X POST http://localhost:8090/v2/wallets/98ae2ce26143a126a19a2d47e60b539a596b4d2b/transactions-construct \
-d '{"delegations":[{"join":{"pool":"pool1mgjlw24rg8sp4vrzctqxtf2nn29rjhtkq2kdzvf4tcjd5pl547k","stake_key_index":"0H"}}]}' \
-H "Content-Type: application/json" | jq .coin_selection.deposits

[]
```

Also deposit is not included in `decode` endpoint.

```
curl -X POST http://localhost:8090/v2/wallets/98ae2ce26143a126a19a2d47e60b539a596b4d2b/transactions-decode \
-d '{"transaction":"hKcAgYJYIBUKRF9NHy5pJ5Ha7JwJ8y1sjCWj+cpsfPFP+AhTdaqgAA2AAYGCWDkAGuA1W/LXhsGyBGci8K+M3tstVDIVQtlX2HDkcDKluI66ZOWc6L9hWCUA8EPtncDVb/uJ7B15fioaAYUDqAIaAAKs/QMaAu/2JwSCggCCAFgcMqW4jrpk5Zzov2FYJQDwQ+2dwNVv+4nsHXl+KoMCggBYHDKluI66ZOWc6L9hWCUA8EPtncDVb/uJ7B15fipYHNol9yqjQeAasGLCwGWlU5qKOV12AqzRMTVeJNoOgKD19g=="}' \

...
{
  "withdrawals": [],
  "inputs": [
...
... but no info about deposit ...
```
It would be good to actually include `deposit` there too.
I think this should be addressed as part of ADP-1190, either in this PR or separate one.

paweljakubas:
@piotr-iohk let's do it in separate PR when this one is finally merged. Withing adp-1190 ticket

PR #3097, opened 2022-01-24
Use Integer instead of Word128 for ghcjs

hamishmack:
The `wide-word` package does not support 32bit platforms like ghcjs.

@rvl:
> We really didn't need wide-word in this situation.
> Execution speed or memory usage is not relevant.
> Integer would work just as well.


jonathanknowles (approved):
Hi @hamishmack 

Thanks for making this. It looks fine to me!

PR #3098, opened 2022-01-26
docs: Add first version of Eras note

rvl:
This adds some general information about hard forking for cardano-wallet developers and others interested in the topic.

### DIFF
--- a/docs/design/Eras.md
+++ b/docs/design/Eras.md
@@ -1,0 +1,56 @@
+# Cardano Eras
+
+A blockchain "hard fork" is a change to the block producing software such that any new blocks produced would be invalid according to the old unchanged software. Therefore, the unchanged software would not accept these blocks into its chain. If nodes running the new and modified software continue to produce blocks on top of their chain, a "fork" will result, starting at the parent block of the first incompatible block.
+
+Cardano manages hard forks by gating the new block generation code behind a feature flag. The network agrees upon a slot at which all nodes will enable the new block generation code. After that slot, the network is said to be operating in a new "era."
+
+Agreement on the slot number (epoch really) to switch to the new era happens by an on-chain voting protocol. If the voting proposal is carried, operators of nodes must ensure that their software version will be compatible with the new era, before the hard fork occurs.
+
+## The Hard Fork Combinator
+
+The mechanism whereby the Cardano Node operates in different modes - depending on which slot it is up to - is called the "Hard Fork Combinator".
+
+The following [seminars](https://input-output.atlassian.net/wiki/spaces/EN/pages/718962750/IOHK+Research+Engineering+Seminar) (company internal use only) are a good introduction:
+
+| Date | Title | Description |
+| --- | --- | --- |
+| 2020/05/15 | [Hardfork Combinator I: Reconciling The Eras](https://drive.google.com/file/d/1m_jKQM_gxBm0ctLqIq9NGj5_nXPI66Su/view) | This session will explain the technology behind the fabled hardfork combinator for Cardano. As this is quite a bit of machinery, and central to Shelley, we will take the time to talk about it in two sessions, to give enough background that everyone can follow along. This first session will focus on the combinator itself, and how advanced Haskell features allow us to travel from one typed world to another. In a future session, which yet needs to be scheduled, he will present on the subtleties of treating time in the presence of a hardfork. |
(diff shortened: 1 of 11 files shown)


docs/design/Eras.md
```diff
@@ -0,0 +1,56 @@
...
+
+Both `cardano-node` and `cardano-wallet` have command-line parameters and/or configuration options for the hard-fork combinator.
+
+### Genesis
+
+Nodes must be configured with genesis blocks. The genesis block of Cardano is the Byron genesis block. Some subsequent eras (Shelley and Alonzo, at the time of writing) also have their own genesis block.
+
+Only the header hash of a genesis block is stored on-chain.
+
+Note that (annoyingly) there are two blocks with epoch 0 and slot 0. One of these is the genesis block, and the other is the first real block, whose parent header hash is the genesis block header hash.
```
HeinrichApfelmus: Technically, the genesis block does not have a slot number and epoch. I think it's not even a block, but I will ask around to pin this down.

```suggestion
Note that the genesis block is not really a block, but a specification of the initial ledger state; it is not associated with a slot or epoch number. The first real block begins at slot 0 and ushers in epoch 0, and it's parent header hash is the hash of the genesis block.
```

docs/design/Eras.md
```diff
@@ -0,0 +1,56 @@
...
+
+When `cardano-wallet` connects to its local `cardano-node`, it must provide a header hash for the byron genesis block. Header hashes for subsequent eras are contained on-chain in the voting system update proposals.
+
+For _mainnet_, we have hardcoded the genesis block hash. It will never change. For _staging_ and _testnet_, users must supply a genesis block to the [[cli]], so that its hash may be used when connecting to the local node.
+
+### _Instafork_™ Technology
+
+Hard forking to Alonzo (the latest era, at the time of writing) by natural methods requires several epochs' worth of network operation time to take effect. For disposable local testnets, this would delay startup and make our integration tests slower. Therefore, `cardano-node` can be configured to hard fork itself at the start of any given epoch number, without an update proposal. It's also possible to hard fork directly into all the eras at epoch 0. This is what we use for the integration tests `local-cluster`.
```
HeinrichApfelmus:
```suggestion
Hard forking to Alonzo (the latest era, at the time of writing) by natural methods requires several epochs' worth of network operation time to take effect. For disposable local testnets, this would delay startup and make our integration tests slower. Therefore, `cardano-node` can be configured to hard fork itself at the start of any given epoch number, without an update proposal. It's also possible to hard fork directly into all the eras at epoch 0. This is what we use for the integration tests `local-cluster`. (The configuration options are named `TestShelleyHardForkAtEpoch` and similar.)
```

HeinrichApfelmus (approved):
Thanks, Rodney!

The story about the genesis block / state is somewhat confusing, and I'm no longer sure myself. I'll try to get "the truth" by asking around.

docs/design/Eras.md
HeinrichApfelmus: Learned the answer:

* Genesis is indeed *not* a block. The Cardano explorer shows it as "block" [5f20df9335…](https://explorer.cardano.org/en/block?id=5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb) anyway.
* The first block on mainnet is an Epoch Boundary Block (EBB). It's a Byron-only concept. This block actually *shares* a slot number with the next block. So, it seems that there are other blocks which can have the same slot number. 😳 Not sure what to make of it. Anyway, Cardano Explorer shows it here: [89d9b5a5…](https://explorer.cardano.org/en/block?id=89d9b5a5b8ddc8d7e5a6795e9774d97faf1efea59b2caf7eaf9f8c5b32059df4).

docs/design/Eras.md
Unisay: I am confused: its written here that Alonzo is the latest **era** while https://roadmap.cardano.org/en/ only mentions   

- Byron
- Shelley
- Goguen
- Basho
- Voltaire

[other random pages](https://www.coindesk.com/markets/2021/08/23/cardano-alonzo-hard-fork-what-you-need-to-know/) say "Alonzo hard fork" (not era). 
What am I missing?

Unisay (commented):
Thanks for the useful links: I learned about seminars records and ledger specs.

docs/design/Eras.md
rvl: Yeah, I'm not quite sure of the exact story, but roadmap eras are quite broad. Goguen Era needed to be split into multiple hard forks. In the source code however, a type with an `IsCardanoEra` instance represents a hard fork, e.g. [Cardano.Api.AlonzoEra](https://input-output-hk.github.io/cardano-node/cardano-api/lib/Cardano-Api.html#t:AlonzoEra). So a roadmap era is different to a `Cardano.Api` era. We shall need to ask/dig around to clarify this.

docs/design/Eras.md
rvl: Thanks, @HeinrichApfelmus! I shall correct the text on these two points. I was indeed getting mixed up with the EBB.

docs/design/Eras.md
HeinrichApfelmus: > So a roadmap era is different to a Cardano.Api era

I also believe that it's unfortunate overloading of the word "era" that has occurred.

PR #3099, opened 2022-01-27
Add `WalletState`, a pure model for the entire wallet

HeinrichApfelmus:
### Issue number

ADP-1375

### Overview

In this step, we introduce a data type `WalletState` which represents the entire wallet state — not just the most recent checkpoint, but _all_ checkpoints.
```
data WalletState s = WalletState
    { prologue    :: Prologue s
    , checkpoints :: Checkpoints (BlockHeader, UTxO, Discoveries s)
    } 
```
The states for the different wallets currently stored in the `walletsVar` DBVar. Eventually, the data type will become the purely functional in-memory representation of all the data associated with a wallet (though perhaps with a different name). The `DBLayer` type will eventually be replaced by a `Store` for values of this type.

The introduction of this type has become possible thanks to the previous separation of the address discovery state `s` into `Prologue s` and `Discoveries s` (PRs #3056, #3068, #3073).

### Details

* As the queries in the `DBLayer` will more and more become queries on the in-memory cache `walletsVar` instead of queries on the database table, I have begun to add unit tests for the database tables. Here, I have added a property test for `loadPrologue` and `insertPrologue`. Eventually, these separate tests will become a single generic test for a `Store` of `Table`.

### Comments

* One of the next steps will be to replace `UTxO` in the above type by its delta encoding `DeltaUTxO`. This will reduce the memory footprint of the in-memory representation.
* The `Cardano.Wallet.DB.Model` module actually implements a pure model for the entire wallet state, including TxHistory etc. When extending the type, we can scavenge from there.

### DIFF
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite.hs
@@ -87,18 +87,9 @@ import Cardano.Wallet.DB
     , sparseCheckpoints
     )
 import Cardano.Wallet.DB.Checkpoints
-    ( DeltaCheckpoints (..)
-    , DeltaMap (..)
-    , findNearestPoint
-    , fromGenesis
-    , getLatest
-    , getPoint
-    )
+    ( DeltaCheckpoints (..) )
 import Cardano.Wallet.DB.Sqlite.CheckpointsOld
-    ( PersistAddressBook (..)
-    , blockHeaderFromEntity
-    , mkStoreWalletsCheckpoints
-    )
+    ( PersistAddressBook (..), blockHeaderFromEntity, mkStoreWallets )
 import Cardano.Wallet.DB.Sqlite.Migration
     ( DefaultFieldValues (..), migrateManually )
 import Cardano.Wallet.DB.Sqlite.TH
@@ -121,6 +112,16 @@ import Cardano.Wallet.DB.Sqlite.TH
     )
 import Cardano.Wallet.DB.Sqlite.Types
     ( BlockId (..), TxId (..) )
+import Cardano.Wallet.DB.WalletState
+    ( DeltaMap (..)
+    , DeltaWalletState1 (..)
+    , findNearestPoint
+    , fromGenesis
+    , fromWallet
+    , getBlockHeight
+    , getLatest
+    , getSlot
+    )
 import Cardano.Wallet.Primitive.AddressDerivation
     ( Depth (..), PersistPrivateKey (..), WalletKey (..) )
 import Cardano.Wallet.Primitive.Slotting
@@ -531,7 +532,7 @@ newDBLayerWith _cacheBehavior tr ti SqliteContext{runQuery} = do
 
     -- FIXME LATER during ADP-1043:
     --   Handle the case where loading the database fails.
-    checkpointsVar <- runQuery $ loadDBVar mkStoreWalletsCheckpoints
+    walletsVar <- runQuery $ loadDBVar mkStoreWallets
 
     -- NOTE
     -- The cache will not work properly unless 'atomically' is protected by a
@@ -542,19 +543,25 @@ newDBLayerWith _cacheBehavior tr ti SqliteContext{runQuery} = do
     -- assuming that the wallet already exists.
     let insertCheckpointCached wid cp = do
             liftIO $ traceWith tr $ MsgCheckpointCache wid MsgPutCheckpoint
-            modifyDBMaybe checkpointsVar $ \ws ->
-                let point = getPoint cp
-                in  case Map.lookup wid ws of
-                    Nothing  -> (Nothing, ())
-                    Just _   -> (Just $ Adjust wid $ PutCheckpoint point cp, ())
+            modifyDBMaybe walletsVar $ \ws ->
+                case Map.lookup wid ws of
+                    Nothing -> (Nothing, ())
+                    Just _  ->
+                        let (prologue, wcp) = fromWallet cp
+                            slot = getSlot wcp
+                            delta = Just $ Adjust wid
+                                [ UpdateCheckpoints $ PutCheckpoint slot wcp
+                                , ReplacePrologue prologue
+                                ]
+                        in  (delta, ())
 
     -- Insert genesis checkpoint into the DBVar.
     -- Throws an internal error if the checkpoint is not actually at genesis.
-    let insertCheckpointGenesis wid cp
-            | W.isGenesisBlockHeader header =
-                updateDBVar checkpointsVar $ Insert wid $ fromGenesis cp
-            | otherwise =
-                throwIO $ ErrInitializeNotGenesis wid header
+    let insertCheckpointGenesis wid cp =
+            case fromGenesis cp of
+                Nothing -> throwIO $ ErrInitializeGenesisAbsent wid header
+                Just wallet ->
+                    updateDBVar walletsVar $ Insert wid wallet
           where
             header = cp ^. #currentTip
 
@@ -563,8 +570,8 @@ newDBLayerWith _cacheBehavior tr ti SqliteContext{runQuery} = do
             :: W.WalletId
             -> SqlPersistT IO (Maybe (W.Wallet s))
         selectLatestCheckpointCached wid = do
-            let get = fmap (snd . getLatest) . Map.lookup wid
-            cp <- get <$> readDBVar checkpointsVar
+            let get = fmap getLatest . Map.lookup wid
+            cp <- get <$> readDBVar walletsVar
             liftIO $ traceWith tr $ MsgCheckpointCache wid $ MsgGetCheckpoint $ isJust cp
             pure cp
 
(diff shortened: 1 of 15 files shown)


lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -542,19 +543,25 @@ newDBLayerWith _cacheBehavior tr ti SqliteContext{runQuery} = do
...
     -- Insert genesis checkpoint into the DBVar.
     -- Throws an internal error if the checkpoint is not actually at genesis.
-    let insertCheckpointGenesis wid cp
-            | W.isGenesisBlockHeader header =
-                updateDBVar checkpointsVar $ Insert wid $ fromGenesis cp
-            | otherwise =
-                throwIO $ ErrInitializeNotGenesis wid header
+    let insertCheckpointGenesis wid cp =
+            case fromGenesis cp of
+                Nothing -> throwIO $ ErrInitializeNotGenesis wid header
```
paweljakubas: what do you think to rename it to `ErrInitializeGenesisAbsent` ?

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -542,19 +543,25 @@ newDBLayerWith _cacheBehavior tr ti SqliteContext{runQuery} = do
...
-                    Nothing  -> (Nothing, ())
-                    Just _   -> (Just $ Adjust wid $ PutCheckpoint point cp, ())
+            modifyDBMaybe walletsVar $ \ws ->
+                case Map.lookup wid ws of
+                    Nothing -> (Nothing, ())
+                    Just _  ->
+                        let (prologue, wcp) = fromWallet cp
+                            slot = getSlot wcp
+                            delta = Just $ Adjust wid
+                                [ UpdateCheckpoints $ PutCheckpoint slot wcp
```
paweljakubas, line 553: so basically this is a moment we do persistence, ie. store what is in `cp` and write to the DB, right? And thanks to construction of `Adjust` we will only write "difference" data, right? not all data from scratch?

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -576,21 +583,21 @@ newDBLayerWith _cacheBehavior tr ti SqliteContext{runQuery} = do
...
-                    Just cps ->
-                        let willKeep cp =
-                                (cp ^. #currentTip ^. #blockHeight ^. #getQuantity)
-                                `Set.member` heights
-                            slots = Map.filter willKeep (cps ^. #checkpoints)
-                        in  (Just $ Adjust wid $ RestrictTo $ Map.keys slots, ())
+                    Just wal ->
+                        let willKeep cp = getBlockHeight cp `Set.member` heights
+                            slots = Map.filter willKeep (wal ^. #checkpoints ^. #checkpoints)
+                            delta = Just $ Adjust wid $ (:[]) $ UpdateCheckpoints $
```
paweljakubas: (minor stylistic remarks) What about `Just . Adjust wid . (:[]) . UpdateCheckpoints . RestrictTo $ Map.keys slots` ?

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
```diff
@@ -639,24 +646,26 @@ newDBLayerWith _cacheBehavior tr ti SqliteContext{runQuery} = do
...
-                    Just cps -> case findNearestPoint cps requestedPoint of
+                    Just wal -> case findNearestPoint wal requestedPoint of
                         Nothing ->
                             ( Nothing
                             , throwIO $ ErrNoOlderCheckpoint wid requestedPoint
                             )
                         Just nearestPoint ->
-                            ( Just $ Adjust wid $ RollbackTo nearestPoint
-                            , pure $ Map.lookup nearestPoint $ cps ^. #checkpoints
+                            ( Just $ Adjust wid $ (:[]) $ UpdateCheckpoints $
```
paweljakubas: if you are ok with stylist the above remark then we can change also here

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
```diff
@@ -175,27 +178,27 @@ import qualified Data.Map.Merge.Strict as Map
...
+        writeS (mkStoreWallet wid) a
     update _ (Delete wid) = do
         -- FIXME LATER during ADP-1043:
         --  Deleting an entry in the Checkpoint table
         --  will trigger a delete cascade. We want this cascade
         --  to be explicit in our code.
         deleteWhere [CheckpointWalletId ==. wid]
     update _ (Adjust wid da) =
-        updateS (mkStoreCheckpoints wid) undefined da
+        updateS (mkStoreWallet wid) undefined da
```
paweljakubas, line 201: not a big fun of using `undefined` .

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
```diff
@@ -175,27 +178,27 @@ import qualified Data.Map.Merge.Strict as Map
...
         -- FIXME LATER during ADP-1043:
```
paweljakubas, line 202: aha, there is plan to replace it properly :+1:

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
```diff
@@ -204,23 +207,52 @@ mkStoreWalletsCheckpoints = Store{loadS=load,writeS=write,updateS=update}
...
         selectAll :: SqlPersistT IO [Entity Wallet]
         selectAll = selectList [] []
 
--- | Store for 'Checkpoints' of a single wallet.
+-- | Store for 'WalletState' of a single wallet.
+mkStoreWallet
+    :: forall s. PersistAddressBook s
+    => W.WalletId
+    -> Store (SqlPersistT IO) (DeltaWalletState s)
+mkStoreWallet wid =
```
paweljakubas, line 221: I wonder if we can introduce reasonable `invariant` to make sure `DeltaWalletState s` is legal/not corrupt (especially in `updateS`). It could be not the case now (as we have prologue and checkpoints only), but when new things like  tx history, pending txs, etc. will be added maybe it is going to be crucial this is the case. Do you have opinion on that?

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
HeinrichApfelmus: Essentially yes. More specifically, at this point in the code, we specify the delta that should be applied to the `Store`.

The actual update is performed by the `updateS` function of the `Store` here:

https://github.com/input-output-hk/cardano-wallet/blob/c649f5730632fa4498750dc6b111a43b380e42c2/lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs#L258-L259

Currently, this function still writes the entire `UTxO` and discovered addresses for the given checkpoint. In order to change that, we need to a) change the constructor `PutCheckpoint` to something like `PutDifferenceToPreviousCheckpoint` and b) change the database schema to accommodate this. One of the next few pull requests will do a), and one of the last commits in this series will be about b).

That said, it is true that thanks to `Adjust wid`, only the checkpoint for one wallet will be written — not all of them.

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
HeinrichApfelmus: Yes. I'll remove `undefined` soon.

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
HeinrichApfelmus: That's a good question: As deltas are now first-class values, how can we ensure that deltas `da` are always paired with the `Base da` value from which they were derived? 🤔

Unfortunately, I don't have a good answer to that yet, but a couple of thoughts:

1. The status quo has the same problem, so at least we are not doing worse. For example, random calls to `putTxHistory` could also mess up the database, it's the responsibility of the caller of `putTxHistory` to make sure that the call is appropriate for the current state of the database.
2. Maybe first-class deltas that can be applied to different values is something we actually want?

    * For example, the `DeltaSet1` type from `Data.Delta` satisfies interesting algebraic properties such as `apply [Insert a, Delete a] = apply (Insert a)`.
    * In this light, it is sensible to design the delta type so that applying it will never fail. For example, `Data.Set.delete` never fails to return a result, even if the element that is to be deleted is not in the set. An alternative design would have been `Data.Set.delete :: Set a -> Maybe (Set a)`, but that would have been more cumbersome to use.
    * Given a partial function `changeP :: b -> Maybe b`, we can always turn it into a total function `changeT :: b -> b` by defining `changeT b = fromMaybe b $ changeP b`. So, the idea would be to always use the second definition when designing a delta type and implementing `apply`.

3. In the other direction, perhaps it is possible to use the type system to ensure that each delta is paired with the value? In a dependently typed language, one could make the value part of the delta type. That would be very cumbersome, though. But perhaps it is possible to simulate this with some type "tags" that keep track of which delta belongs to which value? 

    ```
    newtype Delta da x y
    newtype Base da x

    apply :: Delta da x y -> Base da x -> Base da y
    ```
    Here, the tag `x` ensures that the base value and the delta belong together. Chances are that managing these tags will involve existential quantification and be somewhat cumbersome as well, but maybe there is some low-hanging fruit here.

For now, I'm leaning towards 1 and 2 — making sure that deltas can always be applied and use property tests to make that sequences of deltas are correctly generated.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
HeinrichApfelmus: I agree. Actually, I decided that I do not like `(:[])` and added a commit where I use `[ x ]` instead.

lib/core/src/Cardano/Wallet/DB/Sqlite.hs
paweljakubas: ok, thanks for clarification

lib/core/src/Cardano/Wallet/DB/Sqlite/CheckpointsOld.hs
paweljakubas: indeed, 1 and 2 sounds reasonable

paweljakubas (approved):
lgtm! Another step progressing towards new DB layer.  It would be good if @rvl also give it another look before merging. This PR seems to have a tiny risk of destroying anything, we still write whole utxos and discovered addresses as before, schema is intact. I am very fond of how new storage abstractions eat out bite by bite the legacy solution:-)

HeinrichApfelmus:
bors try

piotr-iohk:
@HeinrichApfelmus I've run e2e suite also on top of this branch and all passed. So lgtm.

HeinrichApfelmus:
bors merge

HeinrichApfelmus:
bors retry

HeinrichApfelmus:
bors retry

HeinrichApfelmus:
bors retry

PR #3100, opened 2022-01-27
Re-write balanceTransaction using Node.evaluateTransactionBalance

Anviking:
- [x] Rewrite balanceTransaction using `evaluateTransactionBalance`
     - Effects on goldens can be viewed in the diff of this PR
- [x] Enable `prop_balanceTransactionBalanced`
- [x] Polish
- [x] Even more polish
    - [x] Will look through from top to bottom before merging

### Future work
- [x] Pay surplus as excess fee when we cannot afford a change-output. (Would like to do while comparing constructTx with balanceTx)
- [x] Ensure tx size limit is respected (Requires fixing ADP-1484 _and_ [this](https://github.com/input-output-hk/cardano-wallet/blob/ae8a2206b6f659b879d3bf14fd4ffdad7e781992/lib/core/src/Cardano/Wallet.hs#L1788-L1818) )
 
### Comments

First part introducing goldens and the evaluateTransactionBalance function: #3150

### Issue Number

ADP-1372

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
@@ -1700,8 +1700,8 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
         \Cannot balance when I cannot afford collateral" $
         \ctx -> runResourceT $ do
         wa <- fixtureWalletWith @n ctx
-            [ 2_000_000
-            , 2_000_000
+            [ 2_500_000
+            , 2_500_000
             ]
         let toBalance = Json PlutusScenario.pingPong_1
         rTx <- request @ApiSerialisedTransaction ctx
@@ -1731,11 +1731,11 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
             , expectErrorMessage errMsg403Collateral
             , expectErrorMessage $ unwords
                 [ "I need an ada amount of at least:"
-                , "2.779500"
+                , "4.278900"
                 ]
             , expectErrorMessage $ unwords
                 [ "The largest combination of pure ada UTxOs I could find is:"
-                , "[1.853000]"
+                , "[2.852600]"
                 ]
             ]
 
@@ -1847,9 +1847,6 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
     it "TRANS_NEW_BALANCE_05/ADP-1286 - \
         \I can balance correctly in case I need to spend my remaining ADA for fee" $
         \ctx -> runResourceT $ do
-        liftIO $ pendingWith
-            "ADP-1286 - ValueNotConservedUTxO: Transaction seems balanced incorrectly \
-            \in case when less than minUtxOValue is left on the wallet"
         wa <- fixtureWalletWith @n ctx [3_000_000]
         -- PlutusScenario.pingPong_1 is sending out 2₳ therefore tx fee
         -- needs to be 1₳ to comply with minUTxOValue constraint
--- a/lib/core/src/Cardano/Api/Gen.hs
+++ b/lib/core/src/Cardano/Api/Gen.hs
@@ -256,8 +256,14 @@ genLovelace = Lovelace <$> frequency
             , (8, pure 65536)
             , (90, pure 4294967296)
             ]
-        x <- choose (-1_000_000, 1_000_000)
-        pure $ max 0 $ boundary + x
+
+        offset <- frequency
+            [ (1, choose (-10, 10))
+            , (1, choose (-220_000, -150_000))
+            , (1, choose (150_000, 220_000))
+            , (1, choose (-1_000_000, 1_000_000))
+            ]
+        pure $ max 0 $ boundary + offset
 
 
 genTxFee :: CardanoEra era -> Gen (TxFee era)
--- a/lib/core/src/Cardano/Api/Gen.hs
+++ b/lib/core/src/Cardano/Api/Gen.hs
@@ -259,8 +259,13 @@ genLovelace = Lovelace <$> frequency
 
         offset <- frequency
             [ (1, choose (-10, 10))
+
+            -- Offset by values close to common fee values, in both the positive
+            -- and negative direction, with the hope that this helps find
+            -- corner-cases.
             , (1, choose (-220_000, -150_000))
             , (1, choose (150_000, 220_000))
+
             , (1, choose (-1_000_000, 1_000_000))
             ]
         pure $ max 0 $ boundary + offset
--- a/lib/core/src/Cardano/Api/Gen.hs
+++ b/lib/core/src/Cardano/Api/Gen.hs
@@ -39,6 +39,7 @@ module Cardano.Api.Gen
   , genValueForMinting
   , genSignedQuantity
   , genTxMintValue
+  , genSignedValue
   , genNetworkMagic
   , genNetworkId
   , genStakeCredential
@@ -458,6 +459,24 @@ genValueForTxOut = do
         ]
     assetQuantities <- infiniteListOf genUnsignedQuantity
     ada <- fromInteger . unLovelace <$> genLovelace
+    return $ valueFromList $ (AdaAssetId, ada) : zip assetIds assetQuantities
+  where
+    unLovelace (Lovelace l) = l
+
+-- | Generate a 'Value' which could represent the balance of a partial
+-- transaction, where both ada and other assets can be included, and quantities
+-- can be both positive and negative.
+genSignedValue :: Gen Value
+genSignedValue = do
+    assetIds <- oneof
+        [ nub <$> scale (`div` 4) (listOf genAssetIdNoAda)
+        , pure []
+        ]
+    assetQuantities <- infiniteListOf genSignedQuantity
+    ada <- fromInteger . unLovelace <$> oneof
+        [ genLovelace
+        , negate <$> genLovelace
+        ]
     return $ valueFromList $ (AdaAssetId, ada) : zip assetIds assetQuantities
   where
     unLovelace (Lovelace l) = l
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -197,6 +197,7 @@ module Cardano.Wallet
     , throttle
     , guardHardIndex
     , withNoSuchWallet
+    , posAndNegFromCardanoValue
 
     -- * Logging
     , WalletWorkerLog (..)
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -1682,6 +1682,11 @@ balanceTransactionWithSelectionStrategy
         , extraCollateral
         , extraOutputs
         , feeUpdate = UseNewTxFee $ unsafeFromLovelace minfee0
+         -- TODO [ADP-1514] Ensure the choice of fee here doesn't cause the fee
+         -- minimization to fail.
+         --
+         -- Assuming fees are between 0.065536 and 4.294967296 ada, it
+         -- shouldn't, but it would be better not to rely on this.
         }
 
     (balance, candidateMinFee) <- balanceAfterSettingMinFee candidateTx
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -1545,12 +1545,38 @@ balanceTransaction
     -> (UTxOIndex WalletUTxO, Wallet s, Set Tx)
     -> PartialTx
     -> ExceptT ErrBalanceTx m SealedTx
-balanceTransaction
+balanceTransaction ctx change pp ti wallet ptx = do
+    let balanceWith strategy = balanceTransactionWithSelectionStrategy @m @s @k
+            ctx change pp ti wallet strategy ptx
+    balanceWith SelectionStrategyOptimal
+        `catchE` \case
+            ErrBalanceTxMaxSizeLimitExceeded
+                -> balanceWith SelectionStrategyMinimal
+            otherErr
+                -> throwE otherErr
+
+balanceTransactionWithSelectionStrategy
+    :: forall m s k ctx.
+        ( HasTransactionLayer k ctx
+        , GenChange s
+        , MonadRandom m
+        , HasLogger m WalletWorkerLog ctx
+        )
+    => ctx
+    -> ArgGenChange s
+    -> (W.ProtocolParameters, Cardano.ProtocolParameters)
+    -> TimeInterpreter (Either PastHorizonException)
+    -> (UTxOIndex WalletUTxO, Wallet s, Set Tx)
+    -> SelectionStrategy
+    -> PartialTx
+    -> ExceptT ErrBalanceTx m SealedTx
+balanceTransactionWithSelectionStrategy
     ctx
     generateChange
     (pp, nodePParams)
     ti
     (internalUtxoAvailable, wallet, _pendingTxs)
+    selectionStrategy
     ptx@(PartialTx partialTx@(cardanoTx -> Cardano.InAnyCardanoEra _ (Cardano.Tx (Cardano.TxBody bod) _)) externalInputs redeemers)
     = do
     guardExistingCollateral
@@ -1953,7 +1979,7 @@ balanceTransaction
                 , utxoAvailableForCollateral =
                       UTxOSelection.availableMap utxoSelection
                 , utxoAvailableForInputs = utxoSelection
-                , selectionStrategy = SelectionStrategyOptimal
+                , selectionStrategy = selectionStrategy
 
                 }
             in
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -1583,15 +1583,34 @@ balanceTransactionWithSelectionStrategy
     guardZeroAdaOutputs (extractOutputsFromTx partialTx)
     guardConflictingWithdrawalNetworks partialTx
 
-    -- TODO: Nothing guarantees consistency between externalInputs and the
-    -- actual txins!
-
     (balance0, minfee0) <- balanceAfterSettingMinFee partialTx
 
     (extraInputs, extraCollateral, extraOutputs) <- do
         let externalSelectedUtxo = UTxOIndex.fromSequence $
                 map (\(i, TxOut a b,_datumHash) -> (WalletUTxO i a, b))
                 externalInputs
+        -- TODO [ADP-1544] 'externalInputs' could conflict with the actual
+        -- inputs in 'partialTx'.
+        --
+        -- Values could be ommitted, added or conflict with the wallet and
+        -- ledger UTxO sets.
+        --
+        -- It appears 'prop_balanceTransactionUnresolvedInputs'
+        --
+        -- Ideally this would at-worst result in the resulting tx being rejected
+        -- by the node. However, /perhaps/:
+        -- 1. Any inconsistent input resolution inside 'balanceTransaction'
+        --    could lead to unexpected results like fees being burnt.
+        -- 2. Incorrect resolution could could lead to phase 2 validation
+        --    failures, if e.g. different coin values lead to a different amount
+        --    of execution units.
+        --
+        -- 'prop_balanceTransactionUnresolvedInputs' show that
+        -- 'balanceTransaction' fails if outputs are omitted because of
+        -- 'assignScriptRedeemers'. This should make the problem (1) unlikely,
+        -- but not necessarily impossible.
+        --
+        -- We should investigate and make sure to handle this better.
 
         -- NOTE: It is not possible to know the script execution cost in
         -- advance because it actually depends on the final transaction. Inputs
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -3392,7 +3392,6 @@ data ErrBalanceTx
     | ErrBalanceTxAssignRedeemers ErrAssignRedeemers
     | ErrBalanceTxNotYetSupported BalanceTxNotSupportedReason
     | ErrBalanceTxFailedBalancing Cardano.Value
-    | ErrBalanceTxMaxSizeLimitExceeded
 
 -- TODO: Remove once problems are fixed.
 data BalanceTxNotSupportedReason
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -1848,16 +1848,16 @@ balanceTransactionWithSelectionStrategy
         (Cardano.InAnyCardanoEra _ (Cardano.Tx (Cardano.TxBody body) _))) = do
         -- Use of withdrawals with different networks breaks balancing.
         --
-        -- For instance the partial tx might contain two withdrawals with the same
-        -- key but different networks:
+        -- For instance the partial tx might contain two withdrawals with the
+        -- same key but different networks:
         -- [ (Mainnet, pkh1, coin1)
         -- , (Testnet, pkh1, coin2)
         -- ]
         --
-        -- Even though this is absurd, the node/ledger @evaluateTransactionBalance@
-        -- will count @coin1+coin2@ towards the total balance. Because the wallet
-        -- does not consider the network tag, it will drop one of the two, leading
-        -- to a discrepancy.
+        -- Even though this is absurd, the node/ledger
+        -- @evaluateTransactionBalance@ will count @coin1+coin2@ towards the
+        -- total balance. Because the wallet does not consider the network tag,
+        -- it will drop one of the two, leading to a discrepancy.
         let networkOfWdrl ((Cardano.StakeAddress nw _), _, _) = nw
         let conflictingWdrlNetworks = case Cardano.txWithdrawals body of
                 Cardano.TxWithdrawalsNone -> False
@@ -1869,9 +1869,9 @@ balanceTransactionWithSelectionStrategy
     guardExistingCollateral (cardanoTx ->
         (Cardano.InAnyCardanoEra _ (Cardano.Tx (Cardano.TxBody body) _))) = do
         -- Coin selection does not support pre-defining collateral. In Sep 2021
-        -- consensus was that we /could/ allow for it with just a day's work or so,
-        -- but that the need for it was unclear enough that it was not in any way
-        -- a priority.
+        -- consensus was that we /could/ allow for it with just a day's work or
+        -- so, but that the need for it was unclear enough that it was not in
+        -- any way a priority.
         case Cardano.txInsCollateral body of
             Cardano.TxInsCollateralNone -> return ()
             Cardano.TxInsCollateral _ [] -> return ()
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -1577,18 +1577,11 @@ balanceTransactionWithSelectionStrategy
     ti
     (internalUtxoAvailable, wallet, _pendingTxs)
     selectionStrategy
-    ptx@(
-        PartialTx partialTx@(
-            cardanoTx ->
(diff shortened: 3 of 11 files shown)


Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

Anviking:
bors try

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1752,6 +1701,180 @@ balanceTransaction
...
+                        tweak (Quantity maxSize) = Quantity
+                            . fromMaybe 0
+                            . intCastMaybe
+                            $ (intCast maxSize) - fixedSize
+                        pp' = over (#txParameters . #getTxMaxSize) tweak pp
+                    in
+                        (view #computeSelectionLimit tl)
+                            pp'
+                            defaultTransactionCtx
+                            outputs
```
Anviking: This is the part I'm uncertain about... Can this be merged as is?

Testing that `balanceTransaction` doesn't create too large txs would require fixing [ADP-1484](https://input-output.atlassian.net/browse/ADP-1484). I would not want to make this PR dependent on that.

Testing that `balanceTransaction` doesn't fail needlessly with `SelectionLimitReached` might be possible as outlined by the comment, but it might take a few days.

Perhaps the
```
let constraints = txConstraints pparams tag
                     --     nWits = estimateNumberOfWitnesses tx
                     --     budget = getTxMaxSize
                     --          - serializedSize tx
                     --          - nWits * (sizeOfWitness constraints)
                     --  in
                     --      budget `div` (sizeOfInput constraints)
```
approach suggested above, but without testing is good enough (for now). Misjudging what we can fit in a tx by a few bytes shouldn't be the end of the world, given that we already may misjudge because of change (ADP-1484)

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1503,45 +1506,26 @@ balanceTransaction
...
-                    else
-                        SelectionCollateralNotRequired
-                } & padFeeEstimation partialTx
-
-        -- FIXME: The coin selection and reported fees will likely be wrong in
-        -- the presence of certificates (and deposits / refunds). An immediate
-        -- "fix" is to return a proper error from the handler when any key or
-        -- pool registration (resp. deregistration) certificate is found in the
-        -- transaction. A long-term fix is to handle this case properly during
-        -- balancing.
```
Anviking, line 1592: 🎉

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1458,7 +1467,6 @@ instance Buildable PartialTx where
 balanceTransaction
     :: forall m s k ctx.
         ( HasTransactionLayer k ctx
-        , HasLogger m WalletWorkerLog ctx
```
Anviking, line 1527: Should probably preserve the logging in the new `selectAssets'`

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -1803,17 +1803,13 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
...
-        wa <- fixtureWalletWith @n ctx [3_000_000]
-        -- PlutusScenario.pingPong_1 is sending out 2₳ therefore tx fee
-        -- needs to be 1₳ to comply with minUTxOValue constraint
-        let expectedFee = 1_000_000
+        let initialBalance = 3_180_000
+        wa <- fixtureWalletWith @n ctx [initialBalance]
+        -- PlutusScenario.pingPong_1 is sending out 2₳
+        let amt = 2_000_000
         let balancePayload = Json PlutusScenario.pingPong_1
-        let hasExpectedFee = [expectField (#fee . #getQuantity) (`shouldBe` expectedFee)]
```
Anviking: Could revisit as part of this PR 🤔 

When I was talking to Piotr I didn't know paying higher fees when we cannot afford a change output was already the normal behaviour of the old tx workflow.

lib/core/src/Cardano/Wallet.hs
Anviking: Should be easily doable to preserve: https://github.com/input-output-hk/cardano-wallet/blob/70cc0619e7c23def2747ab602ef5602741a61076/lib/core/src/Cardano/Wallet.hs#L2144-L2149

Possibly some new log message could be added as well… Should look at it…

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
Anviking: Supporting higher fees seemed to fix the pending tests without many changes 👍 

However, I'd now like to be clearer about:
- Why I needed to tweak the wallet balance in `TRANS_NEW_BALANCE_02c`
- Why there is a slight regression for balancing pingPong_2: wallet now requires 3 ada rather than 2.75 ada

Maybe improving logging will give some insight.

lib/shelley/test/data/balanceTx/pingPong_2/golden
```diff
@@ -53,12 +53,12 @@
...
+ 2.900000,ErrBalanceTxSelectAssets (ErrSelectAssetsSelectionError (SelectionCollateralErrorOf (SelectionCollateralError {largestCombinationAvailable = fromList [((TxIn {inputId = Hash "00000000000000000000000000000000", inputIx = 0},Address "`\177\229\224\251t\200l\128\USdhA\224|\219B\223\139\130\239<\228\229|\181A.w"),Coin 2900000)], minimumSelectionAmount = Coin 4350000})))
+ 2.950000,ErrBalanceTxSelectAssets (ErrSelectAssetsSelectionError (SelectionCollateralErrorOf (SelectionCollateralError {largestCombinationAvailable = fromList [((TxIn {inputId = Hash "00000000000000000000000000000000", inputIx = 0},Address "`\177\229\224\251t\200l\128\USdhA\224|\219B\223\139\130\239<\228\229|\181A.w"),Coin 2950000)], minimumSelectionAmount = Coin 4425000})))
+ 3.000000,ErrBalanceTxSelectAssets (ErrSelectAssetsSelectionError (SelectionCollateralErrorOf (SelectionCollateralError {largestCombinationAvailable = fromList [((TxIn {inputId = Hash "00000000000000000000000000000000", inputIx = 0},Address "`\177\229\224\251t\200l\128\USdhA\224|\219B\223\139\130\239<\228\229|\181A.w"),Coin 3000000)], minimumSelectionAmount = Coin 4500000})))
```
Anviking: This may appear odd. When we have 3 ada we get an error saying that we need at least 4.5 ada. But with 3.05 ada, balancing succeeds.

The reason is:
1. The minimum fee is around 2 ada (≈0.5 from the size, and ≈1.5 for the _maximum_ ex unit costs)
2. With an excess of _just_ 1 ada, this is right at the limit of where coin-selection can't create a change output. Instead it returns a higher "selection delta" / "fee".
3. The collateral requirement is based on the actual `fee`, not the `minfee`. With `collateralPercentage: 150%`, this results in a 3 ada *1.5 = 4.5 ada collateral requirement.

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
Anviking: > Why I needed to tweak the wallet balance in TRANS_NEW_BALANCE_02c
> Why there is a slight regression for balancing pingPong_2: wallet now requires 3 ada rather than 2.75 ada

After looking at it, the cases for 3 and 3.05 ada make sense to me: https://github.com/input-output-hk/cardano-wallet/pull/3100#discussion_r818642803

I'm not sure what changed, but it might be that the new behaviour is more "correct" as for being conservative, and it's such a small change in behaviour that I wouldn't worry about it.

lib/core/src/Cardano/Wallet.hs
Anviking: I added a `MsgSelectionForBalancingStart UTxO PartialTx` analogous to `MsgSelectionStart UTxO [TxOut]`, and kept the SelectionError/Report ones the same.

Anviking:
bors try

lib/core/src/Cardano/Wallet.hs
Anviking: N/A thanks to new approach supported by #3164

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -1805,9 +1805,6 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
     it "TRANS_NEW_BALANCE_05/ADP-1286 - \
         \I can balance correctly in case I need to spend my remaining ADA for fee" $
         \ctx -> runResourceT $ do
-        liftIO $ pendingWith
-            "ADP-1286 - ValueNotConservedUTxO: Transaction seems balanced incorrectly \
-            \in case when less than minUtxOValue is left on the wallet"
```
Anviking, line 1852: PR will fix https://input-output.atlassian.net/browse/ADP-1286

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1780,6 +1814,140 @@ balanceTransaction
...
+                , utxoAvailableForCollateral =
+                      UTxOSelection.availableMap utxoSelection
+                , utxoAvailableForInputs = utxoSelection
+                , selectionStrategy = selectionStrategy
+
+                }
+            in
+                flip evalRand (stdGenFromSeed seed)
+                    $ runExceptT
+                    $ performSelection selectionConstraints selectionParams
```
Anviking: This is the idea discussed in #3093

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1780,6 +1814,140 @@ balanceTransaction
...
+        -> Either (SelectionError WalletSelectionContext) Selection
+    selectAssets' tx utxoSelection balance fee0 seed =
+        let
+            txPlutusScriptExecutionCost = maxScriptExecutionCost tl pp redeemers
+            colReq =
+                    if txPlutusScriptExecutionCost > Coin 0 then
+                        SelectionCollateralRequired
+                    else
+                        SelectionCollateralNotRequired
+
```
Anviking, line 1904:
```suggestion
```

lib/core/src/Cardano/Api/Gen.hs
```diff
@@ -256,8 +256,14 @@ genLovelace = Lovelace <$> frequency
             , (8, pure 65536)
             , (90, pure 4294967296)
             ]
-        x <- choose (-1_000_000, 1_000_000)
-        pure $ max 0 $ boundary + x
+
+        offset <- frequency
+            [ (1, choose (-10, 10))
+            , (1, choose (-220_000, -150_000))
+            , (1, choose (150_000, 220_000))
```
Anviking: Tried to find corner-cases more consistently by adding coin values close to `boundary ± typicalFee`.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1634,6 +1634,48 @@ balanceTransaction
...
+    balanceAfterSettingMinFee tx = ExceptT . pure $ do
+        -- This fee padding works around the problem of the size of a change
+        -- output increasing when minimizing the fee.
+        --
+        -- For this quick workaround, we assume all change outputs will contain
+        -- 0.065536 ada or more. This is certainly the case on mainnet with a
+        -- minUTxOValue of ≈1 ada. Then we only need to care about the following
+        -- two sizes of CBOR lovelace values:
+        -- - 0.065536 ada is encoded as 5 bytes
+        -- - 4.294967296 ada is encoded as 9 bytes
```
Anviking: clarify that this is the lower bound

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1500,28 +1508,23 @@ balanceTransaction
...
-
     guardExistingCollateral
-    guardZeroAdaOutputs outputs
-    guardDeposits
+    guardZeroAdaOutputs (extractOutputsFromTx partialTx)
     guardConflictingWithdrawalNetworks
 
-    (delta, extraInputs, extraCollateral, extraOutputs) <- do
+    -- TODO: Nothing guarantees consistency between externalInputs and the
+    -- actual txins!
```
Anviking: Create a task for this

Anviking: Think through

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1589,32 +1572,49 @@ balanceTransaction
     -- doing such a thing is considered bonkers and this is not a behavior we
     -- ought to support.
 
+    let unsafeFromLovelace (Cardano.Lovelace l) = Coin.unsafeFromIntegral l
     candidateTx <- assembleTransaction $ TxUpdate
         { extraInputs
         , extraCollateral
         , extraOutputs
-        , feeUpdate = UseNewTxFee delta
+        , feeUpdate = UseNewTxFee $ unsafeFromLovelace minfee0
```
Anviking, line 1684: Could think about the effect of the size of the fee field

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1589,32 +1572,49 @@ balanceTransaction
...
-             candidateTx)
-        Right
-        (delta `Coin.subtract` candidateMinFee)
+    -- If there are no change outputs, "burn" the surplus as extra fee.
+    --
+    -- This should only happen when coin-selection cannot afford to construct a
+    -- change output with the appropriate minUTxOValue.
+    let extraFeeToBurn =
+            case extraOutputs of
+                [] | surplus > Coin 20_000_000 -> error $ unwords
```
Anviking: Define in terms of `minUTxOValue`

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -2541,8 +2534,8 @@ prop_balanceTransactionBalanced
     -> ShowBuildable PartialTx
     -> StdGenSeed
     -> Property
-prop_balanceTransactionBalanced (Wallet' utxo wal pending) (ShowBuildable partialTx) seed
-    = withMaxSuccess 200 $ do
```
Anviking, line 2545: Could rename `Balanced` -> `Valid`

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1780,6 +1813,140 @@ balanceTransaction
...
+            selectionParams = SelectionParams
+                -- The following fields are essensially adjusting the coin
+                -- selections notion of balance by @balance0 - sum inputs + sum
+                -- outputs + fee0@ where @balance0@ is the balance of the
+                -- partial tx.
+                --
+                -- We need to compensate for inputs and outputs affecting both
+                -- 1) the @balance@ itself and coin-selection and 2) the coin
+                -- selections notion of balance. Hence we add @-sum inputs + sum
+                -- outputs@ to cancel the effect of (2).
```
Anviking:
```suggestion
```

Redundant with comment just above

lib/core/src/Cardano/Wallet.hs
```diff
@@ -3594,3 +3767,38 @@ instance HasSeverityAnnotation TxSubmitLog where
             BracketException _ -> Error
             _ -> Debug
         MsgProcessPendingPool msg -> getSeverityAnnotation msg
+
+-- | Convert a 'Cardano.Value' into a positive and negative component. Useful
+-- to convert the potentially negative balance of a partial tx into
+-- TokenBundles.
+posAndNegFromCardanoValue
```
Anviking, line 3838: This function is essential to the function of `balanceTransaction` and is covered indirectly by `prop_balanceTransactionBalanced` and similar.

But I see there's an existing `toCardanoValue :: TokenBundle -> Cardano.Value` function. So a test like
```haskell
forAll \(v :: Cardano.Value) -> do
    let (pos, neg) = posAndNegFromCardanoValue 
    (toCardanoValue pos <> toCardanoValue neg) === v
```
might be straight-forward.

I don't think it's essential at _this stage_,  but I could give it a shot when I have time.

Anviking: Also: I'm imagining we'd replace the `(TokenBundle.TokenBundle, TokenBundle.TokenBundle)` in the future with something that can represent negative quantities (like `Cardano.Value`) (in coin-selection)

Anviking:
bors try

lib/core/src/Cardano/Api/Gen.hs
```diff
@@ -256,8 +256,14 @@ genLovelace = Lovelace <$> frequency
...
+            , (1, choose (-1_000_000, 1_000_000))
```
jonathanknowles, line 270: @Anviking I think these constants would benefit from a comment explaining why they were chosen.

What do you think?

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1484,44 +1494,65 @@ instance Buildable PartialTx where
...
+balanceTransactionWithSelectionStrategy
     ctx
     generateChange
     (pp, nodePParams)
     ti
-    (internalUtxoAvailable, wallet, pendingTxs)
-    (PartialTx partialTx@(cardanoTx -> Cardano.InAnyCardanoEra _ (Cardano.Tx (Cardano.TxBody bod) _)) externalInputs redeemers)
+    (internalUtxoAvailable, wallet, _pendingTxs)
+    selectionStrategy
+    ptx@(PartialTx partialTx@(cardanoTx -> Cardano.InAnyCardanoEra _ (Cardano.Tx (Cardano.TxBody bod) _)) externalInputs redeemers)
```
jonathanknowles: This could benefit from some wrapping, as:
- it's hard to see where the nesting is
- it's a very long line

I've made a commit here which fixes the issue:
fb01b2bb95e146e0d6f20b3a558e3fd4085edf3f

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1484,44 +1494,65 @@ instance Buildable PartialTx where
...
+    -> ExceptT ErrBalanceTx m SealedTx
+balanceTransaction ctx change pp ti wallet ptx = do
+    let balanceWith strategy = balanceTransactionWithSelectionStrategy @m @s @k
+            ctx change pp ti wallet strategy ptx
+    balanceWith SelectionStrategyOptimal
+        `catchE` \case
+            ErrBalanceTxMaxSizeLimitExceeded
+                -> balanceWith SelectionStrategyMinimal
+            otherErr
+                -> throwE otherErr
```
jonathanknowles, line 1557: 👍🏻

lib/core/src/Cardano/Wallet.hs
jonathanknowles: @Anviking I agree it's a good idea to make a task for this.

Do you think it qualifies as "technical debt"?

Regardless, perhaps after creating a task, we can make a "TODO" item in the code that references the task.

What do you think?

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2007,10 +2174,10 @@ selectAssets ctx pp params transform = do
     let selectionParams = SelectionParams
             { assetsToMint =
                 params ^. (#txContext . #txAssetsToMint)
-            , assetsToBurn =
-                params ^. (#txContext . #txAssetsToBurn)
             , extraCoinIn = Coin 0
             , extraCoinOut = Coin 0
+            , assetsToBurn =
+                params ^. (#txContext . #txAssetsToBurn)
```
jonathanknowles: Any reason why this one gets moved, but not `assetsToMint`? 😄

lib/core/src/Cardano/Wallet.hs
jonathanknowles: I'm guessing the round-trip test would need to include a negation:

```hs
forAll \(v :: Cardano.Value) -> do
    let (pos, neg) = posAndNegFromCardanoValue 
    (toCardanoValue pos <> (negate <$> toCardanoValue neg)) === v
```

jonathanknowles: > Also: I'm imagining we'd replace the (TokenBundle.TokenBundle, TokenBundle.TokenBundle) in the future with something that can represent negative quantities (like Cardano.Value) (in coin-selection)

Once coin selection no longer depends on `TokenBundle` and `TokenMap`, then we can potentially remove these from the wallet in favour of using Cardano API types.

This is part of the scope of https://input-output.atlassian.net/browse/ADP-1449.

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -2535,33 +2553,33 @@ prop_balanceTransactionBalanced
...
                         "balanced tx has collateral"
-                    (txBalance sealedTx combinedUTxO === 0)
-                    .&&. (abs (txFee sealedTx) .<= 4_000_000)
-                    -- Fee limit chosen at a hunch for the sake of sanity. As
-                    -- long as the property uses mainnet PParams, this is
-                    -- useful.
+                    $ conjoin
+                        [ txBalance sealedTx combinedUTxO === mempty
+                        , (prop_expectFeeExcessSmallerThan
+                            (Cardano.Lovelace 1_200_000) sealedTx)
```
jonathanknowles: Perhaps it would be worth adding a comment here to describe the reasoning behind the 1.2 ada value?

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -2608,7 +2626,8 @@ prop_balanceTransactionBalanced (Wallet' utxo wal pending) (ShowBuildable partia
                 (ErrSelectAssetsSelectionError
                 (SelectionBalanceErrorOf
                 (SelectionLimitReached _)))) ->
-                label "selection limit reached" $ property True
+                    let msg = "selection limit reached should never be returned"
+                    in counterexample msg $ property False
```
jonathanknowles, line 2654: 👍🏻

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -2617,7 +2636,75 @@ prop_balanceTransactionBalanced (Wallet' utxo wal pending) (ShowBuildable partia
...
+        PartialTx s ins r = partialTx'
+        ins' = flip map ins $ \(i,o,dh) ->
+            case UTxO.lookup i walletUTxO of
+                Just o' -> (i,o',Nothing)
+                Nothing -> (i,o,dh)
+
+    walletUTxO :: UTxO
+    walletUTxO =
+          let Wallet' _ w _ = wallet
+          in view #utxo w
```
jonathanknowles:
```suggestion
      let Wallet' _ w _ = wallet
      in view #utxo w
```

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
Anviking: 2c533e2cd312ffc9e505a9be79ee5a7cdd55a49e

Anviking: 2c533e2cd312ffc9e505a9be79ee5a7cdd55a49e

lib/core/src/Cardano/Wallet.hs
Anviking: I remember now. What makes this work is the assumption that the fee with the current mainnet parameters always is encoded with 5 bytes.

Relates to https://input-output.atlassian.net/browse/ADP-1514 anyhow: c2a3eddc3

We _could_ revert back to using `delta`... but I'd rather keep it like this until tackling ADP-1514

lib/core/src/Cardano/Wallet.hs
Anviking: Oops. Might have happened during a rebase. Reverted to the original location.

lib/core/src/Cardano/Api/Gen.hs
Anviking: Done

lib/core/src/Cardano/Wallet.hs
Anviking: :+1: 2ed7982740998eae1aca0a25701043389441daef and https://input-output.atlassian.net/browse/ADP-1544

lib/shelley/test/unit/Cardano/Wallet/Shelley/TransactionSpec.hs
```diff
@@ -2546,21 +2544,20 @@ balanceTransactionGoldenSpec = describe "balance goldens" $ do
...
-prop_balanceTransactionBalanced
+prop_balanceTransactionValid
     :: Wallet'
     -> ShowBuildable PartialTx
     -> StdGenSeed
     -> Property
-prop_balanceTransactionBalanced wallet (ShowBuildable partialTx') seed
-    = withMaxSuccess 1000 $ do
+prop_balanceTransactionValid wallet (ShowBuildable partialTx') seed
+    = withMaxSuccess 20000 $ do
```
Anviking: Oops. Didn't mean to commit the change to `20000`. Should revert

lib/core/src/Cardano/Wallet.hs
Anviking: The property passes, reassuringly: 7de15e6566bfb216f0860e954418f52999061c03

lib/core/src/Cardano/Wallet.hs
Anviking: done

lib/core/src/Cardano/Wallet.hs
Anviking: This is not entirely straight-forward. The `minUTxOValue` isn't a protocol parameter anymore in Alonzo. 

One could perhaps use this function https://github.com/input-output-hk/cardano-wallet/blob/8106b842c80c164664030da879308ef018c33f9a/lib/core/src/Cardano/Wallet/Primitive/Types/Tx.hs#L891 if we pass in a `TokenMap.empty`, but without knowing for sure that it's 100% reliable and works in this context, the 20 ada safety limit seems safer.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1524,43 +1535,83 @@ instance Buildable PartialTx where
...
+        -- 2. Incorrect resolution could could lead to phase 2 validation
+        --    failures, if e.g. different coin values lead to a different amount
+        --    of execution units.
+        --
+        -- 'prop_balanceTransactionUnresolvedInputs' show that
+        -- 'balanceTransaction' fails if outputs are omitted because of
+        -- 'assignScriptRedeemers'. This should make the problem (1) unlikely,
+        -- but not necessarily impossible.
+        --
+        -- We should investigate and make sure to handle this better.
```
jonathanknowles, line 1614: 👍🏻

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1570,49 +1621,45 @@ balanceTransaction
...
-                , txContext
-                , utxoAvailableForInputs
-                , utxoAvailableForCollateral
-                , wallet
-                , selectionStrategy = SelectionStrategyOptimal
-                }
-                transform
+
+        lift $ traceWith tr $ MsgSelectionForBalancingStart
+            (CS.toExternalUTxOMap $ UTxOIndex.toMap internalUtxoAvailable)
```
jonathanknowles: I suspect this conversion can be avoided if we don't require ourselves to include a `UTxO` value in the `MsgSelectionForBalancingStart` value, since we only ever care about the size, which is the same before and after applying `toExternalUTxOMap`.

I'll push a commit that removes this conversion.

jonathanknowles: Fixed in 80f346ce232504e4cf01094193e660e5646793ad.

jonathanknowles (approved):
This looks reasonable to me. I see that you've highlighted all areas for further work with `TODO` comments and links to tickets, which is great.

Thanks for making this PR!

👍🏻

PR #3103, opened 2022-02-01
Handle deposit and refund in constructTransaction and decodeTransaction

paweljakubas:
adp-1190

- [x] I have added support for deposits in decodeTransaction plus testing (core/integration) plus swagger 
- [x] I have added support for deposits in constructTransaction plus integration testing
- [x] I have added support fo refunds in constructTransaction/decodeTransaction/ApiCoinSelection plus integration testing


piotr-iohk (approved):
LGTM.

Enabled also `TRANS_NEW_CREATE_MULTI_TX` integration test and added relevant e2e test.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2301,6 +2302,10 @@ decodeTransaction ctx (ApiT wid) (ApiSerialisedTransaction (ApiT sealed)) = do
         , assetsMinted = ApiT toMint
         , assetsBurned = ApiT toBurn
         , certificates = map (toApiAnyCert acct acctPath) allCerts
+        , deposits =
+                map (const (Quantity . fromIntegral . unCoin . W.stakeKeyDeposit $ pp)) $
+                filter ourRewardAccountRegistration $
+                map (toApiAnyCert acct acctPath) allCerts
```
Anviking: So it seems we don't show <s>refunds</s> reclaims when quitting neither here nor in the `ApiCoinSelection` type?

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2301,6 +2302,10 @@ decodeTransaction ctx (ApiT wid) (ApiSerialisedTransaction (ApiT sealed)) = do
         , assetsMinted = ApiT toMint
         , assetsBurned = ApiT toBurn
         , certificates = map (toApiAnyCert acct acctPath) allCerts
+        , deposits =
+                map (const (Quantity . fromIntegral . unCoin . W.stakeKeyDeposit $ pp)) $
```
Anviking, line 2309: An interesting, but presumably unavoidable thing is that the deposit value could in theory be changed to 3 ada next epoch, and then when you call decodeTx on your old, first delegation tx, it shows as having 3 ada in deposit even though you just paid 2 ada.

I think this is the case at least.

Anviking (approved):
lgtm at a glance. If we show deposits, maybe we should show refunds as well? (or perhaps reclaim is the terminology...) A option could also be to allow negative values in the existing `deposits` list.

paweljakubas:
@Anviking @piotr-iohk I have added `deposits_taken` and `deposits_returned` to both `ApiCoinSelection` and ApiDecodedTransaction`, updated swagger, all unit tests, also extended integration tests due to those changes. I you are happy now, then let's merge it and close adp-1190

lib/core/src/Cardano/Wallet/Api/Server.hs

Anviking (approved):

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -2511,6 +2513,8 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
         verify rDecodedTx4
             [ expectResponseCode HTTP.status202
             , expectField #certificates (`shouldBe` [quittingCert])
+            , expectField #depositsReturned (`shouldBe` [Quantity 1000000])
```
Anviking, line 2402: 👍

piotr-iohk:
> @Anviking @piotr-iohk I have added `deposits_taken` and `deposits_returned` to both `ApiCoinSelection` and ApiDecodedTransaction`, updated swagger, all unit tests, also extended integration tests due to those changes. I you are happy now, then let's merge it and close [adp-1190](https://input-output.atlassian.net/browse/ADP-1190)

@paweljakubas I've noticed that deposit is shown also on `GET /wallets/{walletId}/transactions/{transactionId}`. I believe it should be also updated with `deposit_taken` and `deposit_returned` for consistency.
E.g. Tx for quitting stake pool:
```
curl -X GET http://localhost:8090/v2/wallets/1b0aa24994b4181e79116c131510f2abf6cdaa4f/transactions/a17b0498a8ad7c944098888be4ebe770a20020503e176f78157d66ced59ff08f | jq

# shows
....
  "deposit": {
    "quantity": 0,
    "unit": "lovelace"
  },
....

# should be showing:
....
  "deposit_taken": {
    "quantity": 0,
    "unit": "lovelace"
  },
  "deposit_returned": {
    "quantity": 2000000,
    "unit": "lovelace"
  },
....
....
```

piotr-iohk:
https://github.com/input-output-hk/cardano-wallet/pull/3103#issuecomment-1028774563
TBD in separate PR under ADP-1190.

PR #3105, opened 2022-02-04
`database_schema_version` table stores current database schema version.

Unisay:
### Issue Number

[ADP-1135](https://input-output.atlassian.net/browse/ADP-1135)

This PR adds a database table `database_metadata` with 1 row and 1 column in it: `version` (from here DV).
The idea is that wallet library knows (hardcodes) version of the database file format that it expects to work with (EV), and then:
- if `DV > EV` then its a "database from the future" and no changes should be made to it. Following migrations are aborted.
- if `DV < EV` then its a database from the past, and it needs to be migrated (the new mechanism to apply migrations isn't implemented yet)
- if `DV = EV` then its all good.

### DIFF
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite/Migration.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite/Migration.hs
@@ -144,14 +144,10 @@ migrateManually tr proxy defaultFieldValues =
         return ()
 
     dropTable :: Text -> Text
-    dropTable table = mconcat
-        [ "DROP TABLE IF EXISTS " <> table <> ";"
-        ]
+    dropTable table = "DROP TABLE IF EXISTS " <> table <> ";"
 
     getTableInfo :: Text -> Text
-    getTableInfo table = mconcat
-        [ "PRAGMA table_info(", table, ");"
-        ]
+    getTableInfo table = "PRAGMA table_info(" <> table <> ");"
 
     filterColumn :: [Text] -> [PersistValue] -> Maybe [PersistValue]
     filterColumn excluding = \case
--- a/lib/core/src/Cardano/Wallet/DB/Sqlite/Migration.hs
+++ b/lib/core/src/Cardano/Wallet/DB/Sqlite/Migration.hs
@@ -1,8 +1,9 @@
 {-# LANGUAGE BlockArguments #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase #-}
-{-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE TypeApplications #-}
 
 -- |
@@ -16,6 +17,8 @@
 module Cardano.Wallet.DB.Sqlite.Migration
     ( DefaultFieldValues (..)
     , migrateManually
+    , SchemaVersion (..)
+    , currentSchemaVersion
     , InvalidDatabaseSchemaVersion (..)
     )
     where
@@ -25,7 +28,6 @@ import Prelude
 import Cardano.DB.Sqlite
     ( DBField (..)
     , DBLog (..)
-    , DatabaseMetadataLog (..)
     , ManualMigration (..)
     , fieldName
     , fieldType
@@ -33,8 +35,6 @@ import Cardano.DB.Sqlite
     )
 import Cardano.Wallet.DB.Sqlite.TH
     ( EntityField (..) )
-import Cardano.Wallet.DB.Sqlite.Types
-    ( DatabaseFileFormatVersion (..) )
 import Cardano.Wallet.Primitive.AddressDerivation.Icarus
     ( IcarusKey )
 import Cardano.Wallet.Primitive.AddressDerivation.Shelley
@@ -59,6 +59,8 @@ import Database.Persist.Class
     ( toPersistValue )
 import Database.Persist.Types
     ( PersistValue (..), fromPersistValueText )
+import Numeric.Natural
+    ( Natural )
 import UnliftIO.Exception
     ( Exception, throwIO, throwString )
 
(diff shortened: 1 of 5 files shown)


lib/core/src/Cardano/Wallet/DB/Sqlite/Migration.hs
```diff
@@ -116,6 +131,67 @@ migrateManually tr proxy defaultFieldValues =
...
+        where trace = traceWith tr . MsgMetadata
+
+    createMetadataTableIfMissing :: Sqlite.Connection -> IO TableCreationResult
+    createMetadataTableIfMissing conn = do
+        res <- runSql conn
+            "SELECT name FROM sqlite_master \
+            \WHERE type='table' AND name='database_metadata'"
+        case res of
+           [] -> TableCreated <$ runSql conn
+            "CREATE TABLE database_metadata\
```
rvl: > I forgot to mention a detail about how we use peristent-sqlite.
> The automatic migration system (`mkMigrate "migrateAll"`, intended for development use only) is enabled, so it will create any tables that don't exist, as long as you have an entity definition for it in the TH.
> But the idea is that we shall soon stop using `migrateAll`, so this `CREATE TABLE` statement is probably necessary after all. :slightly_smiling_face: 

My first attempt was exactly this: I defined table/entity in TH and `migrateAll` created the table. However, the problem with this approach is that it only creates a table definition and doesn't allow to populate it with data. In this case we need to insert a single row in the newly created table, and I don't know how to do both with TH/persistent.
I wrote an insert statement in a manual migration step, but it turned out that manual migrations are ran _before_ `migrateAll` 🤷🏻‍♂️ 

The second attempt is presented in this PR: `CREATE TABLE` and `INSERT` statements in one single manual migration.


> 1. The table name: the word metadata is already used twice in our DB schema for different things.

Yeah, this is unfortunate. My line of reasoning was the following:

The definition of **metadata** is **data that describes other data**. 
(Similarly **metainformation** is an **information about other information**).

I see the purpose of this new table is to hold information (data) about the database (other information):
* schema version
* how to handle it (`kill_me` etc)
* potentially other attributes.

To me the `database_metadata` seems to be a better name than `file_format` because: 
- `schema` is commonly used word to describe a database structure: tables, indexes, etc. I understand `format` differently: same schema could be stored in different formats (e.g. in MySql there are `myisam` and `innodb` formats which can store same schema). 
- it replaces `file_` (implementation detail) with `database_` (role):  `file` doesn't apply when this database is created in `:memory:`.

It turns out that metadata is already used in other places... Do you think that `database_metadata` will clash with them and create confusion?

Alternatively, I can rename `database_metadata` to `database_schema_version` table to avoid connotations with other `metadata` concepts that are in use.  We'll want to rename it later when (if) we also store `kill_me` flag in it I reckon...

> 2. I was thinking that it may be helpful for debugging migration problems to record when particular migrations were applied. In that spirit, perhaps include the software version string as extra information.

I agree. I was thinking of addressing this task in smaller iterations. First iteration without history of applied migrations, second one with them. Do you think it doesn't make sense to ship this version number without reworking migrations mechanism?

> 3. Unconditionally executing `CREATE TABLE IF NOT EXISTS` should be fine - I know existing code logs whether tables were created or not - unnecessary IMO.

If the only thing needed is a table creation - such conditional `CREATE` statement is enough. However, migration also needs to `INSERT` a singleton row that contains version value, and this task isn't idempotent, it should only be executed for the first time when table is created.

> 4. 
> ```
> sqlite> CREATE TABLE IF NOT EXISTS migration ( number INTEGER NOT NULL, applied TIMESTAMP DEFAULT CURRENT_TIMESTAMP );
> sqlite> INSERT INTO migration (number) VALUES (0000), (0005), (0042), (0001);
> sqlite> SELECT MAX(number) FROM migration;
> 42
> sqlite> .headers on
> sqlite> SELECT * FROM migration ORDER BY number DESC LIMIT 1;
> number|applied
> 42|2022-02-05 03:46:34
> ```

This is a migration history example that makes sense to implement on top of the `version` field:
we discussed it with @HeinrichApfelmus that `version` that we store in the table actually addresses two concerns:
1. Version of the migration approach (persistent based, future one based on sql files, etc)
1. Version of migration within that migration approach (1.sql, 2.sql, 3.sql, etc)

@HeinrichApfelmus @rvl do you guys think it makes sense for us to discuss it via call in order to get on the same page?

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
```diff
@@ -1244,6 +1255,17 @@ testMigrationPassphraseScheme = do
     Right walOldScheme     = fromText "4a6279cd71d5993a288b2c5879daa7c42cebb73d"
     Right walNoPassphrase  = fromText "ba74a7d2c1157ea7f32a93f255dac30e9ebca62b"
 
+testCreateMetadataTable ::
+    forall s k. (k ~ ShelleyKey, s ~ SeqState 'Mainnet k) => IO ()
+testCreateMetadataTable = do
```
rvl: Nice

rvl (commented):
:+1:

lib/core/test/unit/Cardano/Wallet/DB/SqliteSpec.hs
Unisay: I'll also test for the case when database version is newer than the one supported by the wallet library.

lib/core/src/Cardano/Wallet/DB/Sqlite/Migration.hs
HeinrichApfelmus: > To me the `database_metadata seems` to be a better name than `file_format` because:

My view is actually that the `database` is the implementation detail and that `file` is the role — the main purpose of this thing is to persist the wallet state onto disk, and the file format just happens to be an .sqlite file for better interoperability. (Consequently, I consider the :memory: database to be useful for testing only.) That said, I don't mind the name of the table, though I would prefer `database_schema_version` over `database_metadata` indeed because we also have "transaction metadata".

>  Do you think it doesn't make sense to ship this version number without reworking migrations mechanism?

I would say yes — this version number implementation solves the issue that old wallet software needs to be able to reject new file formats (and not write to the file at all).

Unisay:
I made the following changes to this PR:
- database table renamed to `database_schema_version` to avoid clashes with other usages of the word `metadata`.
- removed traces from migration as it seems to have a complexity costs not worth the value. Verification in the unit test is based on the SQL query instead of traces now.

@HeinrichApfelmus do you think this PR is ready to be merged or is there anything else you'd like to add? Thanks.

HeinrichApfelmus (approved):
😊

Unisay:
🎆

PR #3108, opened 2022-02-04
Use `modifyDBVar` instead of `putCheckpoint` in `Cardano.Wallet`

HeinrichApfelmus:
### Issue number

ADP-1406

### Overview

In this step, we finally remove `putCheckpoint` in favor of `modifyDBVar`. In the `Cardano.Wallet` module, deep within every monadic action, there is a pure function that wants to be freed — and `modifyDBVar` (or rather its friend `modifyDBMaybe`) finally allows us to free those pure functions! For example, we can now implement `updateCosigner` as

```
updateCosigner … = ExceptT $ atomically $ modifyDBMaybe walletsDB $
        adjustNoSuchWallet wid ErrAddCosignerKeyNoSuchWallet
            updateCosigner'
```

where

```
updateCosigner'
    :: WalletState s
    → Either ErrAddCosignerKey (DeltaWalletState s, ())
```

is a pure function that describes how the wallet state is changed.

### Details

* The `DBLayer` now exports the `walletsDB :: DBVar (WalletState s)`. 🥳
* Previously, the `DBLayer` was tested using a `StateMachine` approach. As more and more data is cached by the `DBVar`, we need to start testing `Store` explicitly. This pull request finally adds property tests for the `mkStoreWallet`. These tests are very similar to the state machine tests, in that random sequences of `DeltaWalletState` are generated, but they are simpler as they do not need to worry about results of applying a delta.

### Comments

* Removing the use of `readCheckpoint` will the subject of a later pull request.
* The functions in `Cardano.Wallet` still use `atomically` to ensure the consistency of the database operations. Once the entire wallet state has been migrated into the in-memory `WalletState s` type, we can drop `atomically` and instead rely on the atomicity of `DBVar`. This will increase parallelism, as functions that only want to read the variable, but do not want to write to it, do not need to obtain a write-lock.
* One of the next steps will be to replace the constructor `PutCheckpoint` for the delta type of wallet states with an actual, smaller delta encoding. This will reduce the memory footprint of the in-memory representation.

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -248,6 +248,12 @@ import Cardano.Wallet.DB
     , defaultSparseCheckpointsConfig
     , sparseCheckpoints
     )
+import Cardano.Wallet.DB.Checkpoints
+    ( DeltaCheckpoints (..) )
+import Cardano.Wallet.DB.Sqlite.AddressBook
+    ( AddressBookIso, getPrologue )
+import Cardano.Wallet.DB.WalletState
+    ( DeltaMap (..), DeltaWalletState1 (..), fromWallet, getLatest, getSlot )
 import Cardano.Wallet.Logging
     ( BracketLog
     , BracketLog' (..)
@@ -473,6 +479,8 @@ import Data.ByteString
     ( ByteString )
 import Data.Coerce
     ( coerce )
+import Data.DBVar
+    ( modifyDBMaybe )
 import Data.Either
     ( partitionEithers )
 import Data.Either.Extra
@@ -494,9 +502,11 @@ import Data.IntCast
 import Data.Kind
     ( Type )
 import Data.List
-    ( scanl' )
+    ( foldl' )
 import Data.List.NonEmpty
     ( NonEmpty (..) )
+import Data.Map
+    ( Map )
 import Data.Maybe
     ( fromMaybe, mapMaybe )
 import Data.Proxy
@@ -565,6 +575,7 @@ import qualified Data.ByteString as BS
 import qualified Data.Foldable as F
 import qualified Data.List as L
 import qualified Data.List.NonEmpty as NE
+import qualified Data.Map.Strict as Map
 import qualified Data.Set as Set
 import qualified Data.Text as T
 import qualified Data.Vector as V
@@ -758,6 +769,8 @@ createIcarusWallet ctx wid wname credentials = db & \DBLayer{..} -> do
             , delegation = WalletDelegation NotDelegating []
             }
     mapExceptT atomically $
+        -- FIXME: Why `updateState s cp` and not `cp`?
+        -- The genesis block could very well update the address discovery state.
         initializeWallet wid (updateState s cp) meta hist gp $> wid
   where
     db = ctx ^. dbLayer @IO @s @k
@@ -902,6 +915,7 @@ restoreWallet
         , HasLogger IO WalletWorkerLog ctx
         , IsOurs s Address
         , IsOurs s RewardAccount
+        , AddressBookIso s
         )
     => ctx
     -> WalletId
@@ -987,6 +1001,7 @@ restoreBlocks
         , HasNetworkLayer IO ctx
         , IsOurs s Address
         , IsOurs s RewardAccount
+        , AddressBookIso s
         )
     => ctx
     -> Tracer IO WalletFollowLog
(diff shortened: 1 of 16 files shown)


HeinrichApfelmus:
bors try

lib/core/src/Cardano/Wallet.hs
```diff
@@ -1040,14 +1055,29 @@ restoreBlocks ctx tr wid blocks nodeTip = db & \DBLayer{..} -> mapExceptT atomic
...
-            putCheckpoint wid cp'
+        getBlockHeight cp = fromIntegral $
+            cp ^. #currentTip . #blockHeight . #getQuantity
+        willKeep cp = getBlockHeight cp `Set.member` unstable
+        cpsKeep = filter willKeep (NE.init cps) <> [NE.last cps]
 
-    liftIO $ logCheckpoint (NE.last cps)
-    putCheckpoint wid (NE.last cps)
+        -- NOTE: We have to update the 'Prologue' as well,
+        -- as it can contain addresses for pending transactions,
```
paweljakubas, line 1064: well, the question is whether Prologue needs to have addresses of pending txs. It does not sounds like static data. But when we have then we need to take the burden of having them ;-) Nevertheless, pending txs needs o be taken care when rollbacks happen

lib/core/src/Cardano/Wallet/DB.hs
```diff
@@ -146,6 +150,15 @@ data DBLayer m s k = forall stm. (MonadIO stm, MonadFail stm) => DBLayer
         :: stm [WalletId]
         -- ^ Get the list of all known wallets in the DB, possibly empty.
 
+    , walletsDB
```
paweljakubas, line 153: :+1:

lib/core/src/Cardano/Wallet/DB.hs
```diff
@@ -146,6 +150,15 @@ data DBLayer m s k = forall stm. (MonadIO stm, MonadFail stm) => DBLayer
...
+        :: DBVar stm (DeltaMap WalletId (DeltaWalletState s))
+        -- ^ 'DBVar' containing the 'WalletState' of each wallet in the database.
```
paweljakubas, line 155:
```suggestion
        -- ^ 'DBVar' contains the 'WalletState' of each wallet in the database that in turn
```

lib/core/src/Cardano/Wallet/DB.hs
```diff
@@ -146,6 +150,15 @@ data DBLayer m s k = forall stm. (MonadIO stm, MonadFail stm) => DBLayer
...
+        -- Currently contains all 'Checkpoints' of the 'UTxO' and the
```
paweljakubas, line 156:
```suggestion
        -- contains all 'Checkpoints' of the 'UTxO' and the
```

lib/core/src/Cardano/Wallet/DB/Sqlite/AddressBook.hs
```diff
@@ -70,23 +72,36 @@ class AddressBookIso s where
...
-    data Discoveries (Seq.SeqAnyState n k p) = DS (Discoveries (Seq.SeqState n k))
+    data Prologue (Seq.SeqAnyState n key p) = PS (Prologue (Seq.SeqState n key))
+    data Discoveries (Seq.SeqAnyState n key p) = DS (Discoveries (Seq.SeqState n key))
 
     addressIso = withIso addressIso $ \from to ->
         let from2 st = let (a,b) = from $ Seq.innerState st in (PS a, DS b)
             to2 (PS a, DS b) = Seq.SeqAnyState $ to (a,b)
         in  iso from2 to2
 
+instance Eq (Seq.SeqState n k) => Eq (Prologue (Seq.SeqAnyState n k p)) where
```
paweljakubas, line 96: I wonder if there is no way to kinda automate those equality instances in such a way that it could be derived? There is quite a number of them here

lib/core/test/unit/Cardano/Wallet/DB/Sqlite/StoresSpec.hs
```diff
@@ -88,56 +123,142 @@ prop_prologue_load_write
...
--- | Like 'assert', but allow giving a label / title before running a assertion
-assertWith :: String -> Bool -> PropertyM IO ()
-assertWith lbl condition = do
-    let flag = if condition then "✓" else "✗"
-    monitor (counterexample $ lbl <> " " <> flag)
-    assert condition
+genDeltaWalletState
+    :: (GenState s, AddressBookIso s)
+    => GenDelta (DeltaWalletState s)
+genDeltaWalletState wallet = frequency . map (second updateCheckpoints) $
```
paweljakubas, line 177: :+1:

lib/core/test/unit/Cardano/Wallet/DB/Sqlite/StoresSpec.hs
```diff
@@ -185,3 +295,13 @@ dummyUTCTime = posixSecondsToUTCTime 1506203091
 dummyHash :: Hash "BlockHeader"
 dummyHash = Hash $ unsafeFromHex
     "5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb"
+
+{-------------------------------------------------------------------------------
```
paweljakubas, line 299: what about relocating this function for example here https://github.com/input-output-hk/cardano-wallet/blob/master/lib/test-utils/src/Test/QuickCheck/Extra.hs It seems to be very useful

paweljakubas (approved):
lgtm! very nice that functions start to use `UpdateCheckpoints` and `ReplacePrologue` commands and we don't have commands impl scattered but rather centralized. It would be more maintanable to adopt performance measures on them. Was that also your intention?
Also benchmarks not deteriorated?

lib/core/src/Cardano/Wallet/DB/Sqlite/AddressBook.hs
HeinrichApfelmus: Maybe. In this case, at least I managed to avoid a proliferation of type class contexts by using FlexibleContexts .

At some point, I would like to move all address-related stuff into a `Cardano.Wallet.Address.*` hierarchy and work on removing most of the clutter.

piotr-iohk:
Running benchmark against this branch: https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/1258

piotr-iohk:
> Running benchmark against this branch: https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/1258

The restoration benchmark above timed out because this branch had old timeout values for them.
I've rebased this branch on `master` to get updated timeout values and re-launched benchmarks: https://buildkite.com/input-output-hk/cardano-wallet-nightly/builds/1262.

HeinrichApfelmus:
Thanks Piotr! The benchmarks are becoming slower and slower, but at least they do not seem to be negatively affected by this particular pull request.

HeinrichApfelmus:
bors merge

PR #3110, opened 2022-02-07
select witnesses only for unique inps

paweljakubas:
adp-1190

- [x] I have laxed condition when checking if tx is partially signed. We take into consideration only unique inputs now
- [x] Integration test

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -2415,7 +2415,7 @@ submitTransaction ctx apiw@(ApiT wid) apitx@(ApiSerialisedTransaction (ApiT seal
     let ourOuts = getOurOuts apiDecoded
     let ourInps = getOurInps apiDecoded
 
-    let allInpsNum = length $ apiDecoded ^. #inputs
+    let allInpsNum = length $ L.nubBy sameInput $ apiDecoded ^. #inputs
     let witsNum = length $ getSealedTxWitnesses sealedTx
     when (allInpsNum > witsNum) $
         liftHandler $ throwE $
@@ -2476,6 +2476,13 @@ submitTransaction ctx apiw@(ApiT wid) apitx@(ApiSerialisedTransaction (ApiT seal
         in
             all isInpForeign generalInps &&
             all isWdrlForeign generalWdrls
+
+    sameInput inp1 inp2 = case (inp1, inp2) of
+        (WalletInput (ApiWalletInput id1 ix1 addr1 _ _ _), WalletInput (ApiWalletInput id2 ix2 addr2 _ _ _) ) ->
+               id1 == id2 && ix1 == ix2 && addr1 == addr2
+        (ExternalInput txin1, ExternalInput txin2) ->
+               txin1 == txin2
+        _ -> False
 
 joinStakePool
     :: forall ctx s n k.
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -2415,7 +2415,7 @@ submitTransaction ctx apiw@(ApiT wid) apitx@(ApiSerialisedTransaction (ApiT seal
     let ourOuts = getOurOuts apiDecoded
     let ourInps = getOurInps apiDecoded
 
-    let inpsWits = length $ L.nubBy sameInput $
+    let inpsWits = length $ L.nubBy samePaymentKey $
             (apiDecoded ^. #inputs) ++ (apiDecoded ^. #collateral)
     let wdrlWits = length $ apiDecoded ^. #withdrawals
     let certWits = length $ filter certWitNeeded $ apiDecoded ^. #certificates
@@ -2481,7 +2481,7 @@ submitTransaction ctx apiw@(ApiT wid) apitx@(ApiSerialisedTransaction (ApiT seal
             all isInpForeign generalInps &&
             all isWdrlForeign generalWdrls
 
-    sameInput inp1 inp2 = case (inp1, inp2) of
+    samePaymentKey inp1 inp2 = case (inp1, inp2) of
         (WalletInput (ApiWalletInput _ _ _ derPath1 _ _), WalletInput (ApiWalletInput _ _ _ derPath2 _ _) ) ->
                derPath1 == derPath2
         (ExternalInput txin1, ExternalInput txin2) ->
--- a/lib/core/src/Cardano/Wallet/Api/Server.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Server.hs
@@ -2415,15 +2415,12 @@ submitTransaction ctx apiw@(ApiT wid) apitx@(ApiSerialisedTransaction (ApiT seal
     let ourOuts = getOurOuts apiDecoded
     let ourInps = getOurInps apiDecoded
 
-    let inpsWits = length $ L.nubBy samePaymentKey $
+    let witsRequiredForInputs = length $ L.nubBy samePaymentKey $
             (apiDecoded ^. #inputs) ++ (apiDecoded ^. #collateral)
-    let wdrlWits = length $ apiDecoded ^. #withdrawals
-    let certWits = length $ filter certWitNeeded $ apiDecoded ^. #certificates
-    let witsNeeded = inpsWits + wdrlWits + certWits
-    let witsNum = length $ getSealedTxWitnesses sealedTx
-    when (witsNeeded > witsNum) $
+    let totalNumberOfWits = length $ getSealedTxWitnesses sealedTx
+    when (witsRequiredForInputs > totalNumberOfWits) $
         liftHandler $ throwE $
-        ErrSubmitTransactionPartiallySignedOrNoSignedTx witsNeeded witsNum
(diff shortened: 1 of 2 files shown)


piotr-iohk:
I've pushed a test illustrating the issue in 34c5fa523bb125e3f88f6af63b78d3822529fda1.

Unfortunately it is still failing:
```
   TRANS_NEW_JOIN_02 - Can join stakepool in case I have many UTxOs on 1 address FAILED [1] (20247ms)

Failures:

  src/Test/Integration/Framework/DSL.hs:2446:7: 
  1) API Specifications, NEW_SHELLEY_TRANSACTIONS, TRANS_NEW_JOIN_02 - Can join stakepool in case I have many UTxOs on 1 address
       expected a successful response but got an error: ClientError (Object (fromList [("code",String "missing_witnesses_in_transaction"),("message",String "The transaction has 3 inputs and 2 witnesses included. Submit fully-signed transaction.")]))
       While verifying value:
         ( Status
             { statusCode = 403
             , statusMessage = "Forbidden"
             }
         , Left
             ( ClientError
                 ( Object
                     ( fromList
                         [
                             ( "code"
                             , String "missing_witnesses_in_transaction"
                             )
                         ,
                             ( "message"
                             , String "The transaction has 3 inputs and 2 witnesses included. Submit fully-signed transaction."
                             )
                         ]
                     )
                 )
             )
         )

  To rerun use: --match "/API Specifications/NEW_SHELLEY_TRANSACTIONS/TRANS_NEW_JOIN_02 - Can join stakepool in case I have many UTxOs on 1 address/"

Randomized with seed 1324821070

Finished in 20.2742 seconds, used 6.9438 seconds of CPU time
1 example, 1 failure

```

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2415,7 +2415,7 @@ submitTransaction ctx apiw@(ApiT wid) apitx@(ApiSerialisedTransaction (ApiT seal
     let ourOuts = getOurOuts apiDecoded
     let ourInps = getOurInps apiDecoded
 
-    let allInpsNum = length $ apiDecoded ^. #inputs
+    let allInpsNum = length $ L.nubBy sameInput $ apiDecoded ^. #inputs
```
```suggestion
    let allInpsNum = length $ L.nubBy samePaymentKey $ apiDecoded ^. #inputs
```

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2477,6 +2481,20 @@ submitTransaction ctx apiw@(ApiT wid) apitx@(ApiSerialisedTransaction (ApiT seal
             all isInpForeign generalInps &&
             all isWdrlForeign generalWdrls
 
+    sameInput inp1 inp2 = case (inp1, inp2) of
+        (WalletInput (ApiWalletInput _ _ _ derPath1 _ _), WalletInput (ApiWalletInput _ _ _ derPath2 _ _) ) ->
+               derPath1 == derPath2
+        (ExternalInput txin1, ExternalInput txin2) ->
```
paweljakubas: for external inputs we really cannot do better than that. we don't have knowledge of derivation paths. So in essence, we are not exact here. We provide reasonable upper bound. I will think about it when adding support for partial signing though

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2477,6 +2481,20 @@ submitTransaction ctx apiw@(ApiT wid) apitx@(ApiSerialisedTransaction (ApiT seal
...
+               txin1 == txin2
+        _ -> False
+
+    certWitNeeded = \case
+        WalletDelegationCertificate (JoinPool _ _) -> True
+        WalletDelegationCertificate (QuitPool _) -> True
+        DelegationCertificate (JoinPoolExternal _ _) -> True
+        DelegationCertificate (QuitPoolExternal _) -> True
+        _ -> False
```
paweljakubas, line 2492: registration of stake credential does not need witness. Also I omitted witnesses from pool ops, mirs, etc.

lib/core/src/Cardano/Wallet/Api/Server.hs

piotr-iohk (approved):
Lgtm. Integration test passes, so does e2e tests that used to fail.

lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -2415,11 +2415,15 @@ submitTransaction ctx apiw@(ApiT wid) apitx@(ApiSerialisedTransaction (ApiT seal
     let ourOuts = getOurOuts apiDecoded
     let ourInps = getOurInps apiDecoded
 
-    let allInpsNum = length $ apiDecoded ^. #inputs
+    let inpsWits = length $ L.nubBy samePaymentKey $
+            (apiDecoded ^. #inputs) ++ (apiDecoded ^. #collateral)
+    let wdrlWits = length $ apiDecoded ^. #withdrawals
+    let certWits = length $ filter certWitNeeded $ apiDecoded ^. #certificates
```
Anviking: This appears to ignore any potential for overlap of required witnesses between all of `inpsWits`, `wdrlWits` `certWits` (aside from among the inputs themselves).

So I believe this will overestimate the required number of witnesses if either:
1. A withdrawal key is identical to a input payment key
2. A certificate key is identical to a input payment key
3. There are multiple certs with the same key
4. There's a cert and withdrawal with the same key
5. Multiple withdrawals with the same key (absurd, not sure if technically possible)

I think (4) could be a common occurrence: users re-delegating and the wallet automatically withdrawing their rewards at the same time.

Anviking: <s>Locally when I run `cabal v2-test cardano-wallet:integration --test-option=-j --test-option="10" --test-option="--match" --test-option "TRANS_NEW"` I do get a failure 
Skärmavbild 2022-02-08 kl  13 20 05

because the new tx workflow test for joining does [wait for rewards](https://github.com/input-output-hk/cardano-wallet/blob/baf862039942c05ab0b5e78874f66df51b742a2e/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs#L2258-L2282) before rejoining.

In CI it appears to have passed because of the retrying `it` 😰 </s>

Anviking: Hm… never mind, I spoke too quickly. There are some Plutus related failures in CI however, and I guess the `retry` error logs are related to them.

Anviking: Looking into why then [this waiting for rewards](https://github.com/input-output-hk/cardano-wallet/blob/baf862039942c05ab0b5e78874f66df51b742a2e/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs#L2258-L2282) still works with this PR, plus trying to change the wallet type to a `rewardWallet` for extra rewards in the beginning, it seems that the `expectField #withdrawals (`shouldNotBe` [])` fails for the first two joins… I.e. there are no actual withdrawals 🤔 (I thought there would be)

paweljakubas: it looks like there are needed 2 pools (as in hte test we are rejoining), and sometimes the second pool is not available

Anviking: I pushed an integration test demonstrating (4)

```
 src/Test/Integration/Framework/DSL.hs:2446:7:
  1) API Specifications, NEW_SHELLEY_TRANSACTIONS, TRANS_NEW_JOIN_01e - Can re-join and withdraw at once
       expected a successful response but got an error: ClientError (Object (fromList [("code",String "missing_witnesses_in_transaction"),("message",String "The transaction has 3 inputs and 2 witnesses included. Submit fully-signed transaction.")]))
       While verifying value:
         ( Status
             { statusCode = 403
             , statusMessage = "Forbidden"
             }
         , Left
             ( ClientError
                 ( Object
                     ( fromList
                         [
                             ( "code"
                             , String "missing_witnesses_in_transaction"
                             )
                         ,
                             ( "message"
                             , String "The transaction has 3 inputs and 2 witnesses included. Submit fully-signed transaction."
                             )
                         ]
                     )
                 )
             )
         )
```
and a simple fix: let's not count certs and withdrawals towards the number of required witnesses for now.

Anviking (approved):
lgtm after https://github.com/input-output-hk/cardano-wallet/pull/3110#discussion_r801639726

Anviking:
Looks like there was a haddock error

```
cardano-wallet-core-integration> src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs:2574:9: error:
  | cardano-wallet-core-integration>     parse error on input ‘-- \| Send ada to single wallet address to have [1A, 1A, 1A]’
  | cardano-wallet-core-integration>      \|
  | cardano-wallet-core-integration> 2574 \|         -- \| Send ada to single wallet address to have [1A, 1A, 1A]
  | cardano-wallet-core-integration>      \|         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | cardano-wallet-core-integration>     errMsg400ScriptIllFormed (src/Test/Integration/Framework/TestData.hs:628)
```
Let's see if it works now.

paweljakubas:
bors retry

paweljakubas:
bors retry

paweljakubas:
bors retry

paweljakubas:
bors retry

PR #3111, opened 2022-02-08
Fork the `Selection` type.

jonathanknowles:
## Issue Number

ADP-1410

## Background

The original `Selection` type was **_polymorphic_** in the type of **_change outputs_**. The intention behind this was to allow the wallet to reuse the `Selection` type in different contexts, _both_ **before** _and_ **after** assigning addresses to change outputs.

However, the internal coin selection library is **not** concerned with assigning addresses to change outputs. Assigning change addresses is a function of the **wallet**, rather than the coin selection library, and the coin selection library shouldn't have to be concerned with this detail.

Moreover, requiring the internal coin selection library's `Selection` type to be polymorphic in the type of change outputs makes the internal library functions and types more complicated than necessary.

## Primary Changes

This PR:

- [x] Forks the `Selection` type into two separate types:
    1. A wallet-specific type,
        located in and exported from `Cardano.Wallet.CoinSelection`
        where the type of `change` is kept polymorphic (as is currently the case).
    2. An internal type,
        located in and exported from `Cardano.Wallet.CoinSelection.Internal`
        where the type of `change` is just `[TokenBundle]`.
- [x] Provides a pair of functions `to{External,Internal}Selection` within `Cardano.Wallet.CoinSelection` to handle conversion between these two types.

## Additional Changes

This PR also:
- [x] Moves the `SelectionReport` type (and related functions) to the wallet-specific `Cardano.Wallet.CoinSelection` module, as this is only required by the wallet, and not required by the internal coin selection library.

## Motivation

- Forking the `Selection` type will make it easier to make further simplifications to the internal type, as we can restrict any breakage to the `to{External,Internal}Selection` functions.
- Retaining the original wallet-specific type means that we don’t have to make any changes to wallet code (which relies on `change` being polymorphic) as part of the library extraction process. If we eventually want to change the wallet-specific type to something else, we can change it later on.

### DIFF
--- a/lib/core/src/Cardano/Wallet/CoinSelection.hs
+++ b/lib/core/src/Cardano/Wallet/CoinSelection.hs
@@ -2,7 +2,6 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE OverloadedLabels #-}
 {-# LANGUAGE RecordWildCards #-}
-{-# OPTIONS_GHC -fno-warn-orphans #-}
 
 -- |
 -- Copyright: © 2022 IOHK
--- a/lib/core/src/Cardano/Wallet/CoinSelection.hs
+++ b/lib/core/src/Cardano/Wallet/CoinSelection.hs
@@ -58,17 +58,13 @@ module Cardano.Wallet.CoinSelection
     where
 
 import Cardano.Wallet.CoinSelection.Internal
-    ( Selection
-    , SelectionCollateralRequirement (..)
+    ( SelectionCollateralRequirement (..)
     , SelectionConstraints (..)
     , SelectionError (..)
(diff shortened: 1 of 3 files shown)


HeinrichApfelmus (approved):
Looks good to me! The monomorphic version of `Selection` is indeed no less general than the polymorphic one.

The specific form of the address may be relevant when estimating fees, though. But it's currently overestimated anyway. 🤔

jonathanknowles:
> The specific form of the address may be relevant when estimating fees, though. But it's currently overestimated anyway.

Yes, it's currently overestimated!

If it turns out to be necessary in future, we could introduce an abstraction that gives the "cost" of a particular input, or the cost of a particular "address", hopefully without the library needing to know about the `Address` type.

PR #3114, opened 2022-02-08
Export `ApiValidityBound` constructors

### LINKED ISSUE #3112: Export `ApiValidityBound` constructors from `Cardano.Wallet.Api.Types`
### The problem that you wish to solve

When using the `constructTransaction` endpoint from Haskell, one may need to construct a value for the `validityInterval` field of the `ApiConstructTransactionData` type which has the type `ApiValidityBound`. However, the constructors of `ApiValidityBound` are not exported from `Cardano.Wallet.Api.Types`, which means there's no way to construct such a value — except by parsing it from JSON, which is prone to errors and, thus, runtime failure.

### Description

By exporting the constructors for `ApiValidityBound`, we can be sure at compile-time that we can construct a value of this type.

### Implementation suggestions

Export the constructors for `ApiValidityBound` from `Cardano.Wallet.Api.Types`

runeksvendsen:
Fixes #3112 (please refer to that issue for the rationale behind this tiny change)

### DIFF
--- a/lib/core/src/Cardano/Wallet/Api/Types.hs
+++ b/lib/core/src/Cardano/Wallet/Api/Types.hs
@@ -156,7 +156,7 @@ module Cardano.Wallet.Api.Types
     , ApiStakeKeyIndex (..)
     , ApiPaymentDestination (..)
     , ApiValidityInterval (..)
-    , ApiValidityBound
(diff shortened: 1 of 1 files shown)


rvl (approved):
Thanks :+1:

PR #3116, opened 2022-02-09
Make wallet responsible for determining collateral suitability

jonathanknowles:
## Issue Number

ADP-1411

## Background

The coin selection library is currently responsible for filtering the available UTxO set for entries that are suitable for use as collateral. However, this responsibility is actually delegated back to the wallet, as the **_wallet_** must provide a definition of the following function:
```hs
utxoSuitableForCollateral
    :: (TxIn, TxOut) -> Maybe Coin
    -- ^ Indicates whether an individual UTxO entry is suitable for use as
    -- a collateral input. This function should return a 'Coin' value if
    -- (and only if) the given UTxO is suitable for use as collateral.
```
The coin selection library merely **_applies_** this function as a **_filter_** on the set of available UTxO entries.

## Why this is a problem

The requirement to provide the `utxoSuitableForCollateral` function means that the coin selection library must depend on the `TxOut` type (as collateral suitability is a function of `Address`). Our goal is to evolve the coin selection library so that it does not depend on wallet-specific types like `TxOut`, `Address`, and `UTxO`.

## Changes made by this PR

This PR:

- [x] Removes the `utxoSuitableForCollateral` function from `SelectionConstraints`.
- [x] Makes the wallet fully responsible for identifying which UTxO entries are suitable for collateral, by requiring it to pre-filter the `utxoAvailableForCollateral` set.
- [x] Changes the type of `SelectionParams.utxoAvailableForCollateral` from `UTxO` to Map `TxIn Coin`, as collateral entries can never have non-ada assets.
- [x] Changes the type of `Selection.collateral` from `[(TxIn, TxOut)]` to `[(TxIn, Coin)]`.

## Notes

We will eventually replace all usages of `TxIn` in the coin selection library with a type parameter and `Ord` constraints (where appropriate). This PR is a step towards this goal.

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -316,7 +316,7 @@ import Cardano.Wallet.Primitive.AddressDiscovery.Shared
     , addCosignerAccXPub
     )
 import Cardano.Wallet.Primitive.Collateral
-    ( asCollateral )
+    ( addressSuitableForCollateral )
 import Cardano.Wallet.Primitive.Migration
     ( MigrationPlan (..) )
 import Cardano.Wallet.Primitive.Model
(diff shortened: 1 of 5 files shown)


rvl (approved):
Nice :ok_hand:

rvl:
Is this change compatible with your balance transaction work @Anviking?

Anviking (approved):

PR #3102, opened 2022-01-31
Prototype: "light mode" for cardano-wallet

HeinrichApfelmus:
### Issue number

ADP-1370

### Overview

This prototype explores whether we can implement a "light mode" for cardano-wallet which aims to make synchronisation faster by using a less trusted source for blockchain data.

Result: Yes, the implementation of a "light mode" is feasible, but restricted to Shelley wallets.

For the full report, see the [README](https://github.com/input-output-hk/cardano-wallet/blob/HeinrichApfelmus/ADP-1379/prototypes/light-mode-test/README.md).

The prototype is a Cabal project contained in the folder `/prototypes/light-mode-test/`.

### Comments

* This prototype uses https://blockfrost.io as external data source. In order to run the actions in the `Demo` module, you have to register with Blockfrost and put your API token in `secrets/blockfrost.mainnet.token`.
* How do we get a rendered version of the report for easier reading? 🤔

PR #3118, opened 2022-02-09
Specification of light-mode

HeinrichApfelmus:
### Issue number

ADP-1370

### Overview

This pull request adds a [specification (DRAFT) for the cardano-wallet light-mode](https://github.com/input-output-hk/cardano-wallet/blob/master/docs/design/specs/light-mode.md).

See also PR https://github.com/input-output-hk/cardano-wallet/pull/3102 for the prototype that explores the feasibility of this specification and implements some parts of it for demonstration purposes.


specifications/wallet/spec-light-mode.md
```diff
@@ -0,0 +1,123 @@
...
+
+* `CRED` specifies how to connect to the less trusted blockchain data source. (MVP: Use [Blockfrost][]; `CRED` is a filepath to the secret token.)
+* The `--light` argument and the `--node-socket` arguments are mutally exclusive with each other.
+
+### REST API
+
+When the executable was started in light-mode:
+
+* The endpoints in the hierarchy `/v2/byron-wallets/*` MUST return an error.
+* The endpoints in the hierarchy `/v2/shared-wallets/*` MAY return an error.
```
rvl: I think we should state that we intend to support light shared wallets, but it would not be implemented in the very first iteration.

specifications/wallet/spec-light-mode.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+* `CRED` specifies how to connect to the less trusted blockchain data source. (MVP: Use [Blockfrost][]; `CRED` is a filepath to the secret token.)
+* The `--light` argument and the `--node-socket` arguments are mutally exclusive with each other.
+
+### REST API
+
+When the executable was started in light-mode:
+
+* The endpoints in the hierarchy `/v2/byron-wallets/*` MUST return an error.
```
rvl: There should be an explanation somewhere of why this is.
"... Because byron wallets use random derivation indices, they are significantly more challenging to make a light wallet for."

However we do have options should a byron light wallet really be desired:
1. Develop a third-party byron address discovery service which takes a byron root XPub and sends back a list of addresses belonging to that wallet. (May not be a great idea security-wise).
2. For existing byron wallets in Daedalus, their wallet database contains addresses generated/discovered so far. Without the node running, we can use these addresses for light wallet transaction searching. So the light wallet balance would be up-to-date, as long as no new addresses have been generated/used since that wallet was last fully synced.

specifications/wallet/spec-light-mode.md
```diff
@@ -0,0 +1,123 @@
...
+        ```
+        
+        By using this type as argument to `applyBlocks`, we are guaranteed that `BlockSummary` is only used in conjunction with sequential state. Caveat: It would be better if we could avoid parametrizing the `NetworkLayer` type with the address discovery state `s`.
+
+
+## Quality assurance
+
+### Benchmarks
+
+* Compare restoration times for an empty wallet in node-mode and in light-mode.
```
rvl:
```suggestion
* Compare restoration times for an empty wallet in node-mode and in light-mode.
```
```suggestion
* Compare restoration times for an empty wallet in node-mode and in light-mode.
  The existing nightly wallet restoration benchmark can be updated for this purpose.
```

specifications/wallet/spec-light-mode.md
```diff
@@ -0,0 +1,123 @@
...
+* Number of requests that we make to the data source needs to be monitored / kept in check.
```
rvl: Perhaps add: "Of interest is whether the number of requests scales linearly with the number of transactions × number of addresses."

specifications/wallet/spec-light-mode.md
```diff
@@ -0,0 +1,123 @@
+# Specification: Light mode for cardano-wallet
+
+9th Feb 2022
+
+Status: DRAFT
+
+# Synopsis
+
+This document specifies a light-mode for `cardano-wallet`.
+This mode aims to make synchronisation to the blockchain faster by using a less trusted source for blockchain data.
```
rvl:
```suggestion
This mode aims to make synchronisation to the blockchain faster by trusting an off-chain source of aggregated blockchain data.
```

specifications/wallet/spec-light-mode.md
```diff
@@ -0,0 +1,123 @@
...
+
+In addition, the "trust versus speed" trade-off will be partially obsoleted by [Mithril][] technology, which aims to give us "trust *and* speed" by providing verified ledger state data as opposed to block data. The act of implementing light-mode in `cardano-wallet` does not only offer immediate benefits, but will, in fact, partially prepare the codebase for Mithril technology.
+
+  [daedalus]: https://daedaluswallet.io
+  [nami]: https://namiwallet.io
+  [blockfrost]: http://blockfrost.io
+  [mithril]: https://iohk.io/en/research/library/papers/mithrilstake-based-threshold-multisignatures/
+
+# Motivation
+
```
rvl: Perhaps something like these:

1. Allow users to operate their wallet without needing to wait for the node to download and validate the entire blockchain.
2. Allow users to run cardano-wallet and Daedalus on systems with 8GB of memory and <= 10GB of disk space.
3. Allow users to control where they sit on the trust vs convenience spectrum, depending upon their current circumstances.

specifications/wallet/spec-light-mode.md
```diff
@@ -0,0 +1,123 @@
...
+## Limitations
+
+* *Trust*. The external data source does *not* provide the same level of protection against *omission* of transactions as the Proof-of-Stake consensus protocol.
+* *Privacy*. In addition to the reduction of trust in the blockchain data, we now also reveal private information about addresses belonging to a single wallet.
+* *Address schemes*. Only *Shelley* wallets with *sequential address discovery* can be supported.
+
+However, these limitations are shared by all existing light wallets. In fact, some light wallets only provide *single-address* wallets, which is an additional reduction of privacy.
+
+# Specification
+
```
rvl: This would be a good place to describe the address+transaction discovery loop.

rvl (commented):
It's great so far, thanks @HeinrichApfelmus

specifications/wallet/spec-light-mode.md
```diff
@@ -0,0 +1,143 @@
...
+## User-visible benefits
+
+* Allow users to operate their wallet without needing to wait for the node to download and validate the entire blockchain.
+* Allow users to run cardano-wallet and Daedalus on systems with significantly less than 8GB of memory and less than 10GB of disk space.
+* Allow users to control where they sit on the trust vs convenience spectrum, depending upon their current circumstances.
+
+## Technical benefits
+
+* *Speed*. With light-mode, we expect synchronisation times of < 1 minute for a wallet with 1'000 transactions. In contrast, synchronisation of an empty wallet currently takes ~ 50 minutes by connecting to a node.
```
rvl: It takes 50 minutes to sync an empty wallet -- assuming that the node has itself synced to the chain tip and built its ledger state, which itself takes hours.

specifications/wallet/spec-light-mode.md
```diff
@@ -0,0 +1,143 @@
...
+* [Cardano-launcher][] will have to be modified to not launch the `cardano-node` process when the cardano-wallet is launched in light-mode.
+* Provision of the data source is outside the scope of light-mode. We use [Blockfrost][] for demonstration purposes in the MVP.
+
+  [cardano-launcher]: https://github.com/input-output-hk/cardano-launcher
+
+# Design alternatives
+
+## Byron wallets
+
+Byron wallets cannot be supported by the light-mode as described above, as it is based on an efficient query `Address -> Transactions` which Byron wallets cannot take advantage of. For these wallets, alternative designs would have to be explored Possibilities include:
```
rvl:
```suggestion
Byron wallets cannot be supported by the light-mode as described above, as it is based on an efficient query `Address -> Transactions` which Byron wallets cannot take advantage of. For these wallets, alternative designs would have to be explored. Possibilities include:
```

specifications/wallet/spec-light-mode.md
```diff
@@ -0,0 +1,143 @@
...
+* *Address schemes*. Only *Shelley* wallets with *sequential address discovery* can be supported.
+
+However, these limitations are shared by all existing light wallets. In fact, some light wallets only provide *single-address* wallets, which is an additional reduction of privacy.
+
+# Specification
+
+## Overview
+
+The implementation of light-mode is based on an efficient query `Address -> Transactions` which the blockchain data source provides. This query is useful to wallets that use *sequential address discovery* (Shelley wallets). These wallets work with a sequence of potential addresses `addr_0`, `addr_1`, `addr_2`, …. For each integer `n`, there is a deterministic procedure for generating the corresponding address `addr_n`. The wallet generates the first few addresses in this sequence and uses the above query to retrieve the corresponding transactions. When no transactions are found for `g` ("address gap") many consecutive addresses, the procedure stops — as the wallet never puts addresses with higher indices on the chain. In other words, this iterative loop yields all transactions that belong to the wallet.
+
```
rvl: I wouldn't normally indulge in unnecessary pictures, but for this case, could we please have a simple flow chart showing the address+tx discovery loop?

specifications/wallet/spec-light-mode.md
```diff
@@ -0,0 +1,143 @@
...
+This procedure is implemented and demonstrated in the `light-mode-test` prototype.
+
+## Functionality
+
```
rvl: And here is another gratuitous picture opportunity.
Could we please have two block diagrams - full-node and light-mode?
Like a cut-down version of the diagram here: https://input-output-hk.github.io/cardano-wallet/developers/Architecture
In the "full-node" version, wallet links to node in 4 ways: chainsync, localblockfetch, localstatequery, localtxsubmission.
In the "light-mode" version, wallet requests a list of transactions by address from the block service.

specifications/wallet/spec-light-mode.md
```diff
@@ -0,0 +1,143 @@
...
+                    :: NonEmpty Block
+                    -> BlockData m s
+                Summary
+                    :: BlockSummary m
+                    -> BlockData m (SeqState n k)
+        ```
+        
+        By using this type as argument to `applyBlocks`, we are guaranteed that `BlockSummary` is only used in conjunction with sequential state. Caveat: It would be better if we could avoid parametrizing the `NetworkLayer` type with the address discovery state `s`.
+
+
```
rvl: One more thing that we should mention.
Rather than using LocalTxSubmission on cardano-node, we will be asking a web service to submit our transactions.
An example of such a thing is `cardano-submit-api`.
We assume this web service will be operated and paid for by someone.

rvl (approved):
Great!

I've suggested a couple diagrams. If you don't have time, I could do them.

Could we move the document into `docs/design/specs/light-mode.md`? Then it will be published. It doesn't matter if it's still a DRAFT - as long as the status is shown.

specifications/wallet/spec-light-mode.md
HeinrichApfelmus: Ok!

Note: We do not use `BlockFetch`, but `ChainSync` to receive the blocks from the node. This is line with the network specification, which recommends that

> The (full)-node to client protocol uses an instance of the chain synchronisation protocol that transfers full blocks, while the node-to-node instance only transfers block headers. In the node-to-node case, the block fetch protocol (Section 3.8) is used to transfer full blocks.

PR #3129, opened 2022-02-11
Remove assertions for pending tx after doing submit tx (it's flaky)

piotr-iohk:
- 6ffe085b35d838b89a166268ce94d4d5dd7aeeef
  remove Pending tx assertion as it causes flakiness
  
- 0d9870fd879957f536113ae2ca89389448b08eaf
  remove Pending tx assertion from e2e tests too as they cause false-positives sometimes as well

### DIFF
--- a/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
+++ b/lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
@@ -843,7 +843,6 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
         verify rGetTx
             [ expectResponseCode HTTP.status200
             , expectField (#direction . #getApiT) (`shouldBe` Outgoing)
-            , expectField (#status . #getApiT) (`shouldBe` Pending)
             ]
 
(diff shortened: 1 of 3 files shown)


lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
```diff
@@ -843,7 +843,6 @@ spec = describe "NEW_SHELLEY_TRANSACTIONS" $ do
         verify rGetTx
             [ expectResponseCode HTTP.status200
             , expectField (#direction . #getApiT) (`shouldBe` Outgoing)
-            , expectField (#status . #getApiT) (`shouldBe` Pending)
```
Anviking, line 846: 👍 But we haven't seen this being flaky yet?

Anviking (approved):
👌

lib/core-integration/src/Test/Integration/Scenario/API/Shelley/TransactionsNew.hs
piotr-iohk: I had a failure locally... so it'd be just a matter of time I believe.

PR #3130, opened 2022-02-11
nixpkgs: 21.05 -> 21.11

jbgi:
buildkite agents have been update to 21.11.


rvl (approved):
Great, thanks JB!

When I made this same change in PR #3115, I noticed that the nixosTests failed to build. Same thing has happened in here.
One of the python dependencies of the nixos tests framework seems to fail its unit tests.
I am not sure why those packages need to be rebuilt anyway - surely they would have been built and cached by NixOS hydra?

PR #3132, opened 2022-02-11
Fix for the `listAssets` API which shows assets unrelated to a wallet

Unisay:
### Issue Number

ADP-710

#### Problematic current state:

list of wallet assets `WAs`  derived like this:

> for each transaction `T` in blockchain:  
>   if `T` has at least one output associated with the current wallet:  
>     add `T` to the list of Wallet transactions `WTs`;  

> for each transaction `T` in `WTs`:
>   for each output `O` of `T`: 
>     for each asset `A` in token bundle of `O`:
>       add `A` to the list of wallet assets `WAs`;

<details><summary><strong><s>Solution №1</s></strong> (unfoldable, superseded by the Solution №2, see below)</summary><p>

When restoring wallet state from blocks of transactions all transaction outputs that are not related to wallet address are omitted;

> for each transaction `T` in blockchain:   
> if `T` has at least one output associated with the current wallet:
> let `T'` be `T` with all unrelated* outputs removed;
> add `T'` to the list of wallet transactions `WTs`;

> for each transaction `T` in `WTs`:
> for each output `O` of `T`:
> for each asset `A` in token bundle of `O`:
> add `A` to the list of wallet assets `WAs`;

* output is considered to be unrelated if its address if not the one belonging to the wallet;

#### Consequences

* Wallet states collected before the fix contain transactions with all outputs (including unrelated) In order for the unrelated assets (those found in unrelated outputs) to disappear wallet state needs to be restored from scratch.

* Modified transactions `T'` are no longer valid as some of their outputs were omitted;
</p></details>

<details><summary><strong><s>Solution №2</s></strong> (unfoldable, superseded by the Solution №3, see below)</summary><p>

#### Solution №2

> for each transaction `T` in blockchain:   
> if `T` has at least one output associated with the current wallet:
> let `T'` be `T` with all outputs tagged with their belonging (ours/theirs) - `b(O)`;
> add `T'` to the list of wallet transactions `WTs`;

> for each transaction `T` in `WTs`:
> for each output `O` of `T`:
> if `b(O)` is not `Ours` then skip this `O`; 
> for each asset `A` in token bundle of `O`:
> add `A` to the list of wallet assets `WAs`;
</p></details>

#### Solution №3

> for each transaction `T` in `WTs`:
> for each output `O` of `T`:
> if `O` is not `Ours` then skip this `O`; 
> for each asset `A` in token bundle of `O`:
> add `A` to the list of wallet assets `WAs`;

### Ad-hoc improvements

- [x] Move asset extraction logic to the `Cardano.Wallet` module
- [x] Add an extra property test to cover transaction application.

### DIFF
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -158,6 +158,7 @@ module Cardano.Wallet
     -- ** Transaction
     , forgetTx
     , listTransactions
+    , extractWalletAssetsFromTxs
     , getTransaction
     , submitExternalTx
     , submitTx
@@ -2489,6 +2490,14 @@ listTransactions ctx wid mMinWithdrawal mStart mEnd order = db & \DBLayer{..} ->
                 $ slotRangeFromTimeRange
                 $ Range mStart mEnd
 
+-- | Extract assets associated with a given wallet from its transaction history.
+extractWalletAssetsFromTxs
+    :: WalletId -> [TransactionInfo] -> [TokenMap.AssetId]
+extractWalletAssetsFromTxs _walletId = Set.toList . Set.unions . map txAssets
+  where
+    txAssets = Set.unions
+        . map (TokenBundle.getAssets . view #tokens)
+        . txInfoOutputs
 
 -- | Get transaction and metadata from history for a given wallet.
 getTransaction
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -2507,7 +2507,7 @@ listAssets ctx wid = db & \DBLayer{..} -> do
             . map (TokenBundle.getAssets . view #tokens)
             . filter ourOut
             . txInfoOutputs
-        ourOut TxOut {address} = ourAddress address
+        ourOut TxOut{address} = ourAddress address
         ourAddress addr = isJust . fst . isOurs addr $ getState cp
     pure $ Set.unions $ map txAssets txs
   where
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -158,7 +158,7 @@ module Cardano.Wallet
     -- ** Transaction
     , forgetTx
     , listTransactions
-    , extractWalletAssetsFromTxs
+    , listAssets
     , getTransaction
     , submitExternalTx
     , submitTx
@@ -2491,14 +2491,17 @@ listTransactions ctx wid mMinWithdrawal mStart mEnd order = db & \DBLayer{..} ->
                 $ Range mStart mEnd
 
 -- | Extract assets associated with a given wallet from its transaction history.
-extractWalletAssetsFromTxs
+listAssets
     :: forall s k ctx. (HasDBLayer IO s k ctx, IsOurs s Address)
     => ctx
     -> WalletId
-    -> [TransactionInfo]
-    -> ExceptT ErrNoSuchWallet IO [TokenMap.AssetId]
-extractWalletAssetsFromTxs ctx wid txInfos = db & \DBLayer{..} -> do
+    -> ExceptT ErrNoSuchWallet IO (Set TokenMap.AssetId)
+listAssets ctx wid = db & \DBLayer{..} -> do
     cp <- mapExceptT atomically $ withNoSuchWallet wid $ readCheckpoint wid
+    txs <- lift . atomically $
+        let noMinWithdrawal = Nothing
+            allTxStatuses = Nothing
+        in readTxHistory wid noMinWithdrawal Ascending wholeRange allTxStatuses
     let txAssets :: TransactionInfo -> Set TokenMap.AssetId
         txAssets = Set.unions
             . map (TokenBundle.getAssets . view #tokens)
@@ -2506,7 +2509,7 @@ extractWalletAssetsFromTxs ctx wid txInfos = db & \DBLayer{..} -> do
             . txInfoOutputs
         ourOut TxOut {address} = ourAddress address
         ourAddress addr = isJust . fst . isOurs addr $ getState cp
-    pure $ Set.toList $ Set.unions $ map txAssets txInfos
+    pure $ Set.unions $ map txAssets txs
   where
     db = ctx ^. dbLayer @IO @s @k
 
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -509,7 +509,7 @@ import Data.List.NonEmpty
 import Data.Map.Strict
     ( Map )
 import Data.Maybe
-    ( fromMaybe, mapMaybe )
+    ( fromMaybe, mapMaybe, isJust )
 import Data.Proxy
     ( Proxy )
 import Data.Quantity
@@ -2492,12 +2492,23 @@ listTransactions ctx wid mMinWithdrawal mStart mEnd order = db & \DBLayer{..} ->
 
 -- | Extract assets associated with a given wallet from its transaction history.
 extractWalletAssetsFromTxs
-    :: WalletId -> [TransactionInfo] -> [TokenMap.AssetId]
-extractWalletAssetsFromTxs _walletId = Set.toList . Set.unions . map txAssets
+    :: forall s k ctx. (HasDBLayer IO s k ctx, IsOurs s Address)
+    => ctx
+    -> WalletId
+    -> [TransactionInfo]
+    -> ExceptT ErrNoSuchWallet IO [TokenMap.AssetId]
+extractWalletAssetsFromTxs ctx wid txInfos = db & \DBLayer{..} -> do
+    cp <- mapExceptT atomically $ withNoSuchWallet wid $ readCheckpoint wid
+    let txAssets :: TransactionInfo -> Set TokenMap.AssetId
+        txAssets = Set.unions
+            . map (TokenBundle.getAssets . view #tokens)
+            . filter ourOut
+            . txInfoOutputs
+        ourOut TxOut {address} = ourAddress address
+        ourAddress addr = isJust . fst . isOurs addr $ getState cp
+    pure $ Set.toList $ Set.unions $ map txAssets txInfos
   where
-    txAssets = Set.unions
-        . map (TokenBundle.getAssets . view #tokens)
-        . txInfoOutputs
+    db = ctx ^. dbLayer @IO @s @k
 
 -- | Get transaction and metadata from history for a given wallet.
 getTransaction
--- a/lib/core/src/Cardano/Wallet.hs
+++ b/lib/core/src/Cardano/Wallet.hs
@@ -509,7 +509,7 @@ import Data.List.NonEmpty
 import Data.Map.Strict
     ( Map )
 import Data.Maybe
-    ( fromMaybe, mapMaybe, isJust )
+    ( fromMaybe, isJust, mapMaybe )
 import Data.Proxy
     ( Proxy )
 import Data.Quantity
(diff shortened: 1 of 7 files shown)


lib/core/src/Cardano/Wallet/Api/Server.hs
```diff
@@ -1734,20 +1734,11 @@ listAssets ctx wid = do
...
-    W.listTransactions @_ @_ @_ wrk wid Nothing Nothing Nothing Descending
-  where
-    allTxAssets = Set.toList . Set.unions . map txAssets
-    txAssets = Set.unions
-        . map (TokenBundle.getAssets . view #tokens)
-        . W.txInfoOutputs
+    :: forall s k. ApiLayer s k -> ApiT WalletId -> Handler [AssetId]
+listAssetsBase ctx (ApiT walletId) =
+    withWorkerCtx ctx walletId liftE liftE $ \wrk ->
+        liftHandler $ W.extractWalletAssetsFromTxs walletId <$>
```
rvl: Ah yes - that's better.

Unisay:
@rvl 

I implemented what we discussed today: instead of filtering "foreign" assets at "query" time we instead filter them at wallet restore time, omitting "foreign" TX outputs. 

Please take a look 👀

lib/core/src/Cardano/Wallet/Primitive/Model.hs
```diff
@@ -554,8 +555,13 @@ applyOurTxToUTxO
     -> UTxO
     -> Maybe ((Tx, TxMeta), UTxO)
 applyOurTxToUTxO !slotNo !blockHeight !s !tx !prevUTxO =
-    if hasKnownWithdrawal || prevUTxO /= ourNextUTxO
-        then Just ((tx {fee = actualFee dir}, txmeta), ourNextUTxO)
+    if ourWithdrawalSum /= mempty || prevUTxO /= ourNextUTxO
+        then
+            let updatedTx = tx
+                    { fee = actualFee dir
+                    , outputs = filter (ours s . address) (outputs tx)
```
rvl: Yes this looks like it will work for `listAssets`.
If everything breaks due to this change, then perhaps we could annotate the TxOut address about whether it belongs to the wallet.

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -819,6 +831,38 @@ prop_applyOurTxToUTxO_someOurs ourState slotNo blockHeight tx utxo =
...
+    -> SlotNo
+    -> Quantity "block" Word32
+    -> Tx
+    -> UTxO
+    -> Property
+prop_applyOurTxToUTxO_omitUnrelatedOuputs st ourAddress slot blocks tx utxo =
+    appliedTx & maybe discard ( \tx' ->
+        report (tx' ^. #outputs) "tx' outputs"  $
+        property $
+            all (`Set.member` ourAddresses st') (address <$> tx' ^. #outputs)
```
rvl: I think we also want an assertion which says that any output addresses which are in `tx` but not `tx'` must not be members of `ourAddresses`.

rvl (commented):
Great

lib/core/src/Cardano/Wallet/Primitive/Model.hs
HeinrichApfelmus: Uhh, this does indeed look like it could break everything. 😅 I think the original idea for the `Tx` type was to represent a transaction exactly as it appears on the chain — no modifications. (I believe that the adjustment to the `fee` field is a hack, as this field did not exist in Byron, but is added for clarity.) Changing this assumption has several implications:

* The `/v2/wallets/{walletId}/transactions` endpoint changes in a backwards-incompatible way and no longer presents transactions as they appear on-chain. For example, `cardano-node` will return a different result than this endpoint if you give it the same transaction hash.
* `Tx` are stored in the database via `putTxHistory`. If we change the contents of `Tx`, we would have to add a migration for the database as well.
* The `inputIx` field of the `TxIn` refers to an index in the list `outputs`. If we filter the list, the indices get rearranged. 😱 We are safe in that `ourNextUTxO` uses the original `tx`, not the `updatedTx` to compute the `UTxO` set. Users of the `/v2/…/transactions` endpoint are not so lucky. If we really want to filter outputs from the `Tx`, we should at least record their indices as well.

In light of these implications, my preference would be to keep the original idea of `Tx` representing a complete transaction, and do the filtering in `listAssets` a posteriori.

rvl: The `Tx` type in cardano-wallet has never been a transaction exactly as it appears on the chain. Multiple fields are missing!
It can be whatever we want or need it to be. We can compute info about the transaction once during restoration, then cache it in the db. An example of such info is the transaction amount.

I ask, do users really need to see TxOuts which don't concern their wallet?
If they do, perhaps the wallet should store the original serialized transaction in the database for this purpose.

Although, if our code is making assumptions all over the place that the TxOuts aren't filtered -- you raise a very good point about numbering -- then it's probably easiest to _not_ filter, but instead annotate the TxOut address with a flag indicating whether the address belongs to this wallet.

Yes this will probably need database migrations; we can do migrations.
Also users will not see a propertly filtered listAssets unless they restore from genesis. This is also fine.

Unisay: @HeinrichApfelmus appreciate your input! 
I've added *Solution #2* to the issue description that takes your input into account as described by @rvl

Unisay: @HeinrichApfelmus finally I finished covering discussed implementation with property tests. Please take another look and if there are no obstacles - I'll merge it.

lib/core/src/Cardano/Wallet.hs
```diff
@@ -2411,6 +2412,28 @@ listTransactions ctx wid mMinWithdrawal mStart mEnd order = db & \DBLayer{..} ->
...
+    txs <- lift . atomically $
+        let noMinWithdrawal = Nothing
+            allTxStatuses = Nothing
+        in readTxHistory wid noMinWithdrawal Ascending wholeRange allTxStatuses
+    let txAssets :: TransactionInfo -> Set TokenMap.AssetId
+        txAssets = Set.unions
+            . map (TokenBundle.getAssets . view #tokens)
+            . filter ourOut
+            . txInfoOutputs
+        ourOut TxOut {address} = ourAddress address
```
HeinrichApfelmus: Minor style comment: I like to avoid whitespace between a record constructor and its curly braces so that I can count the number of function arguments by counting whitespace.

```suggestion
        ourOut TxOut{address} = ourAddress address
```

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -770,6 +773,24 @@ prop_applyOurTxToUTxO_allOurs slotNo blockHeight tx utxo =
     shouldHaveResult :: Bool
     shouldHaveResult = evalState (isOurTx tx utxo) AllOurs
 
+-- Verifies that UTxO set can't grow when there are no "our" transactions.
+-- Hint: UTxO set shrinks when foreign transaction uses it as collateral.
```
HeinrichApfelmus: 🤔? I don't think that the hint is correct. "Collateral" has a very specific meaning and the UTxO can shrink even when no "collateral" inputs are used. Also, a transaction cannot be "foreign" if it spends an output belonging to the wallet!

I think the underlying issue is that there are two ways to recognize whether a transaction creates or spends an output belonging to our wallet: a) checking whether the address of a transaction output belongs to us and b) looking up a transaction input in our `UTxO`. (`isOurTx`gives a complete list of ways). In particular, transaction inputs cannot be recognized by their address — transaction inputs are stored as references and are not resolved. The only way to recognize a `TxIn` is to have a current `UTxO` set as reference — this is a subtle, but important point. In turn, this implies that the `utxo` arguments from all the tests are, in fact, transaction outputs that have once been recognized as "ours".

In that light, I recommend to remove the `prop_applyOurTxToUTxO_noneOurs` test — the assumption that no outputs ever belong to the test wallet logically implies that the argument `utxo` should empty. The logic that the test was trying to cover is probably already covered by `prop_applyOurTxToUTxO_someOurs`.

lib/core/test/unit/Cardano/WalletSpec.hs
```diff
@@ -642,6 +655,87 @@ walletListTransactionsSorted wallet@(wid, _, _) _order (_mstart, _mend) history
...
+                    , outputs = [out1, out2]
+                    , metadata = mempty
+                    , withdrawals = mempty
+                    , scriptValidity = Nothing
+                    }
+        assets <- listHistoricalAssets [ (tx, txMeta) ]
+        monitor $ report out1 "Output with related address"
+        monitor $ report out2 "Output with unrelated address"
+        monitor $ report assets "Discovered assets"
+        assert $ assets == getAssets (out1 ^. #tokens)
```
HeinrichApfelmus, line 688: By comparing the assets in `out1`, this test subsumes `property walletListAssets`? 🤔 I'm actually not entirely sure about our test strategy myself, personally I like to err on the "do not add tests that make me feel good, but only marginally improve coverage" side. Up to you.

specifications/api/swagger.yaml
```diff
@@ -5633,8 +5633,8 @@ paths:
         List all assets associated with the wallet, and their metadata
         if known.
 
-        An asset is _associated_ with a wallet if it is involved in a
-        transaction of the wallet.
+        An asset is _associated_ with a wallet if, at one point in history,
+        it was spendable by the wallet.
```
HeinrichApfelmus, line 5637: 👍

HeinrichApfelmus (changes requested):

lib/core/test/unit/Cardano/WalletSpec.hs
Unisay: You're right. I'll remove first property as it is also checked by the second one. In other words I can see how 1st property can fail without 2nd property failing too.

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -788,8 +796,8 @@ prop_applyOurTxToUTxO_noneOurs slotNo blockHeight tx utxo =
     report (utxoFromTx tx) "utxoFromTx tx" $
     property $ utxo' `isSubsetOf` utxo
   where
-    utxo' =
-        maybe utxo snd $ applyOurTxToUTxO slotNo blockHeight NoneOurs tx utxo
+    utxo' = maybe utxo snd $
+        applyOurTxToUTxO slotNo blockHeight (OurUtxo utxo) tx utxo
```
HeinrichApfelmus: Thanks, I like this new type much better. 😊 Though now the property could, in principle, be violated, because the transaction `tx` may contain outputs with addresses that are also in the `utxo`.

lib/core/test/unit/Cardano/Wallet/Primitive/ModelSpec.hs
```diff
@@ -577,11 +579,17 @@ instance IsOurs AllOurs a where
...
--- | A simplified wallet state that never marks entities as "ours".
+-- | A simplified wallet state that tracks UTxOs as "ours".
 --
-data NoneOurs = NoneOurs
+newtype OurUtxo = OurUtxo UTxO
 
-instance IsOurs NoneOurs a where
+instance IsOurs OurUtxo Address where
+    isOurs addr s@(OurUtxo utxo) =
+        if any (address >>> (== addr)) (Map.elems (unUTxO utxo))
```
HeinrichApfelmus:
```suggestion
        if any ((== addr) . address) (Map.elems (unUTxO utxo))
```
🥺

HeinrichApfelmus (approved):

PR #3144, opened 2022-02-18
docs: Improve address discovery diagram for the light mode spec

Unisay:
- [x] I have added gap counter to the address discovery diagram

### DIFF
--- a/docs/design/specs/light-mode.md
+++ b/docs/design/specs/light-mode.md
@@ -62,11 +62,13 @@ This procedure can be visualized in a flow chart:
 :::{.mermaid-container}
 ```mermaid
 flowchart TB
-  start([begin]) --> init[j := 0]
+  start([begin]) --> init[j := 0\ng := 0\ngap := 20]
   init --> query[query addr_j]
   query --> tx{transactions?}
-  tx -- no --> g{happened g times?}
-  tx -- yes --> add[j := j+1]
+  tx -- no --> gapAdd[g := g+1] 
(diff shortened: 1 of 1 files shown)


docs/design/specs/light-mode.md
```diff
@@ -62,11 +62,13 @@ This procedure can be visualized in a flow chart:
...
 ```mermaid
 flowchart TB
-  start([begin]) --> init[j := 0]
+  start([begin]) --> init[j := 0\ng := 0\ngap := 20]
   init --> query[query addr_j]
   query --> tx{transactions?}
-  tx -- no --> g{happened g times?}
-  tx -- yes --> add[j := j+1]
+  tx -- no --> gapAdd[g := g+1] 
+  gapAdd --> g{g > gap?}
```
HeinrichApfelmus, line 69: `>=` or `>`? 😅 The rule is "Between any two used addresses, there are at most `gap` many consecutive unused addresses. (An address is used if the query returns a nonempty list of transactions, unused otherwise.)".

Unisay: This seems correct to me: `g` counts unused addresses, if `g` is greater than `gap` then search is over, thus allowing it to have at most `g = gap` unused addresses.

Unisay: Please merge this PR if you agree with my logic.

HeinrichApfelmus: > `g` counts unused addresses

That's a very good argument / invariant.

PR #3145, opened 2022-02-18
Add mock-up of CoinSelection module

rvl:
This is an attempt to come up with the simplest possible interface for coin selection that would work, without using concrete types.

It will hopefully help us define a common understanding of what the coin selection module will look like after the refactors are done.

### Comments

Please check and see whether it makes sense to you, and try to think of necessary operations that it can't support.

### Issue Number

ADP-1449


sevanspowell:
I think we could distill the coin selection function down further:

These are overloaded terms, so they should be considered in the context of "coin selection".

```
type UTxO = (TxIn, TxOut)
type Wallet = Set UTxO
-- An abstract type, think of this more generally as "Goal" or "Target". It can
-- include minting and burning amounts, etc. I've left this abstract here, but I
-- imagine we'd decide on a common type.
data Value
```

`TxIn`, `TxOut`, and `Value`, should be considered "simple types with the minimum possible information required for coin selection". E.g. `TxOut` could just be a `TokenBundle` or some other purely monetary type.

```
selectCoins
  :: Wallet
  -- ^ From this wallet (i.e. utxo set)
  -> Value
  -- ^ select stuff to cover this abstract value
  -> Either e (Wallet, Set UTxO)
  -- ^ and optionally return a new wallet (utxo set), and the selected inputs

-- optionally (recognizing State pattern):
selectCoins :: Value -> Wallet -> Either e (Wallet, Set UTxO)
-- optionally (recognizing State pattern):
selectCoins :: Value -> Either e (State Wallet (Set UTxO))
```

These are functions that produce coin selection functions:

```
randomImprove
  :: RandomImproveParameters
  -> (Wallet -> Value -> Either RandomImproveFail (Wallet, Set UTxO))

sweep
  :: SweepParameters
  -> (Wallet -> Value -> Either SweepFail (Wallet, Set UTxO))
```

`RandomImproveParameters` might include cost functions or characteristics of the blockchain. If parameters need to be shared, we could rewrite as:

```
randomImprove
  :: CommonParameters
  -> RandomImproveParameters
  -> (Wallet -> Value -> Either RandomImproveFail (Wallet, Set UTxO))

sweep
  :: CommonParameters
  -> SweepParameters
  -> (Wallet -> Value -> Either SweepFail (Wallet, Set UTxO))
```

There is at least one coin selection function which ignores all characteristics of the block chain. So I think it's important to not include transaction constraints and cost functions in every coin selection function:

```
mempty -- Just returns original wallet and an empty selection
  :: (Wallet -> Value -> Either e (Wallet, Set UTxO))
```

These coin selection functions could be arbitrarily decorated to recover whatever functionality we need from our current coin selection interface.

I wonder that if there's the concept of "mempty" w.r.t coin selection (above), perhaps there's also the concept of "compose", perhaps via some Applicative or Alternative interface. Not that I think we have much use for that atm.

```
data CoinSelectionFn a = Value -> State Wallet a

pure :: a -> CoinSelectionFn a
pure = const . pure

<*> :: CoinSelectionFn (a -> b) -> CoinSelectionFn a -> CoinSelectoinFn b
<*> ff fg value = (ff value) <*> (fg value)
```

I think distilling the concept of "coin selection" down to it's base elements (from a conceptual sense, not a Cardano-specific sense) will have a far-reaching simplifying and clarifying effect. Seeing potential composition opportunities at this stage is good evidence of that.

I think a good Cardano coin-selection library would present a key interface (i.e. `Wallet -> Value -> Either e (Wallet, Set UTxO`)) consistent with any UTxO-based blockchain - obviously we would hard-code it to Cardano for convenience sake, but the concept should be transferable.

prototypes/coin-selection/CoinSelection.hs
```diff
@@ -0,0 +1,278 @@
...
+--
+
+-- | The transaction type is polymorphic and defined by the caller. Coin
+-- selection however needs to get certain information about the transaction
+-- under construction.
+--
+-- This class defines the types of transaction components and therefore
+-- represents transactions which can have the coin selection algorithms of this
+-- module applied to them.
+class TxBody tx where
```
HeinrichApfelmus: One of the main questions is whether we want concrete types (like `Coin`) or polymorphic types (e.g. `Coin tx`).

* The advantage of concrete types is that the type signatures become simpler — without any loss of generality. For example, the canonical choice for `Amount` is `newtype Amount = Amount Natural`. A coin selection algorithm does not become more general if we use a polymorphic type like `Amount tx` — the type `Natural` embodies exactly the information that we need of a monetary amount.
* However, the advantage of polymorphic types (at least in this context) is that we can plug in existing types without *coercion cost*. If we were to use `newtype Amount = Amount Natural`, then we would have to convert to our `Coin` type, or the `Cardano.Api` equivalent, depending on where we use it. By using `newtype`, we can avoid a *runtime* coercion cost, but we probably still have to pay a *code* coercion cost, as we need to write code that coerces the types.

Our codebase actually contains cautionary tales for both approaches:
* The plethora of type class contexts in `Cardano.Wallet.Primitive.Address*` highlights the complexity of overly polymorphic types.
* `Cardano.Wallet.Shelley.Compatibility` highlights the coercion cost of concrete types.

That said, I lean slightly in favor of not paying the coercing cost, provided that we keep the class `TxBody` in check. For example, that would involve adding constraints such as `Monoid (Amount tx)` to the class definition (so that they are brought into scope when the constraint `TxBody tx => …` is imposed), and making the type families injective.

prototypes/coin-selection/CoinSelection.hs
```diff
@@ -0,0 +1,278 @@
...
+    -> Selector
+    -- ^ How to select inputs from the UTxO.
+    -> Change
+    -- ^ How to assign excess value to change outputs.
+    -> UTxO tx
+    -- ^ Transaction outputs available to use as inputs of a this transaction.
+    -> tx
+    -- ^ Partial transaction body which is the starting point for coin selection.
+    -> Either SelectionError (NonEmpty tx)
+    -- ^ One or more funded and balanced transaction bodies, or an error message.
```
HeinrichApfelmus, line 284: Which of the transaction bodies are completions of the partial transaction body that was given as argument? 🤔 I assume that only one body can follow the template, while the other transactions are preparatory transactions that rearrange the UTxO? Unfortunately, in the latter case, the `TxBody` constructor does not contain enough functions to construct a transaction out of the blue — something like `mempty` is needed. 😅

prototypes/coin-selection/CoinSelection.hs
```diff
@@ -0,0 +1,278 @@
...
+    deriving Generic
+-}
+
+-- | A list of assets and their quantity.
+type Value tx = Map (AssetId tx) (Amount tx)
+
+-- | Unspent transaction outputs. These are the things which fund transactions.
+--
+-- Note that this type is a bit different from the usual map from TxIn -> TxOut.
+type UTxO tx = [(TxIn tx, Value tx)]
```
HeinrichApfelmus: We probably want to be able to handle the case of wallets with large `UTxO` sets, hence we probably need something with faster indexing, i.e. a `Map (TxIn tx) (Value tx)`.

prototypes/coin-selection/CoinSelection.hs
```diff
@@ -0,0 +1,278 @@
...
+--   withdrawals. Put this 'TxIn' either in the 'UTxO' list or in the initial
+--   @tx@.
+-- [Coalesce UTxO]: Add a transaction output with an address belonging to the
+--   wallet and pass the v'Sweep' option.
+-- [Split UTxO]: Add several transaction outputs with addresses belonging to the
+--   wallet. Use the v'Balanced' option to distribute funds evenly.
+-- [Spending particular UTxO]: To ensure that a particular UTxO(s) is/are spent,
+--   add them to the 'TxIn' set of the initial @tx@.
+-- [Get fee]: Caller should calculate the difference of inputs and outputs
+--   in the returned transaction.
```
HeinrichApfelmus, line 268: I'm not sure if we should leave the computation or recording of the fee to the caller? 🤔 The `xyzzy` algorithm has to deal with fees anyway, when it applies `marginalCostOfOutput` and (in the unlucky case) attempts to cover the cost with an additional input. In other words, the algorithm (in this form at least), has to deal with fees in a recursive fashion anyway, and it makes sense to take all the responsibility away from the caller. Then again, this would probably involve adding an additional `cost :: tx -> Cost tx` to the `Params` type. 🤔

prototypes/coin-selection/CoinSelection.hs
```diff
@@ -0,0 +1,278 @@
...
+--   and possibly also a stake key registration certificate. These cost money.
+--   However 'xyzzy', is not necessarily Cardano-specific or even POS-specific.
+--   Therefore, the 'Address' type for @tx@ should include a constructor for
+--   certificates. The implementation of 'appendTxOutput' should put these
+--   "certificate" address outputs in the correct place within the concrete
+--   'TxBody' type.
+-- [Withdraw from reward account]: Reward account withdrawals add value to the
+--   transaction. The 'TxIn' type for @tx@ should include a constructor for
+--   withdrawals. Put this 'TxIn' either in the 'UTxO' list or in the initial
+--   @tx@.
```
HeinrichApfelmus, line 260: > The 'TxIn' type for @tx@ should include a constructor for withdrawals.

👍 

I think the main reason for no making withdrawals equal to `TxIn` on the main chain is that withdrawals could be double-spent, while `TxIn` cannot be double-spent. We may want to consider using a slightly different name, e.g. `FundsIn`. 🤔

prototypes/coin-selection/CoinSelection.hs
```diff
@@ -0,0 +1,278 @@
...
+--
+-- [Payment]: Add a 'TxOut' to the initial @tx@ with the recipient
+--   'Address' and 'Value'.
+-- [Change Output]: Add a 'TxOut' to the initial @tx@ such that the 'Address'
+--   satisfies 'isOurAddress'. Its 'Value' would usually be zero.
+--   If the maximum output size is reached, another 'TxOut' with the same
+--   address will be added to the transaction.
+-- [No new change addresses]: If there is no transaction output that satisfies
+--   v'IsOurAddress', then change will be directed back to the addresses
+--   belonging to the wallet which funded the transaction.
```
HeinrichApfelmus, line 249: I like this idea of specifying a skeleton for transaction outputs, and filling it up. However, I feel that the simplicity of the scheme is lost a bit if we allow the algorithm to add new outputs — perhaps the function should even return an error in this case? 🤔 That said, I like Sam's observation that the funds can always be funneled through a total `Amount`, and only then be redistributed to transaction outputs. Then again, this "total amount" funnel does not consider fees, which *do* depend on the number of transaction outputs. 🤔

prototypes/coin-selection/CoinSelection.hs
```diff
@@ -0,0 +1,278 @@
...
+-- [Join stake pool]: This transaction will involve a delegation certificate,
+--   and possibly also a stake key registration certificate. These cost money.
+--   However 'xyzzy', is not necessarily Cardano-specific or even POS-specific.
+--   Therefore, the 'Address' type for @tx@ should include a constructor for
+--   certificates. The implementation of 'appendTxOutput' should put these
+--   "certificate" address outputs in the correct place within the concrete
+--   'TxBody' type.
```
Anviking, line 256: The abstract coin-selection shouldn't need to know about certificates. They can be treated as something which affects the size (and therefore fee) as well as balance (w.r.t deposits) of a transaction via some `evaluateTransaction{Balance,MinFee}` functions passed to `xyzzy`.

prototypes/coin-selection/CoinSelection.hs
Anviking: I think @rvl's idea is to 1) make coin-selection unaware of withdrawals, while 2) still allowing coin-selection to still choose when they should be made.

This is an interesting idea. But:
a. The distinction is important: txs must contain at least one actual `TxIn` [[shelley ledger spec, p22]](https://hydra.iohk.io/build/12728100/download/1/ledger-spec.pdf)
b. Do we really need coin-selection to choose when to make withdrawals?

I think leaving the choice of "should I make a withdrawal or not?" to the caller is fine. This is what we currently [do](https://github.com/input-output-hk/cardano-wallet/blob/ea9f07ce2806d5132b95723136c18c63b63da9f8/lib/core/src/Cardano/Wallet/Transaction.hs#L253). The caller would then add it to the partial `tx`, and coin-selection would not need to be aware of the withdrawal, only its effect on the balance through `evaluateTransactionBalance`.

prototypes/coin-selection/CoinSelection.hs
rvl: Yes, probably.
This type alias would be better named "AvailableInputs." There are no map lookups needed for coin selection.
Depending on the coin selection algorithm, `xyzzy` will either select randomly from these pairs, or sort them by value and take largest first.
So I renamed the element type, and changed the list to `Foldable`.

prototypes/coin-selection/CoinSelection.hs
rvl: Oops, of course.
The next transaction might be a preparatory transaction, or simply a continuation of the previous transaction which reached a size limit.

prototypes/coin-selection/CoinSelection.hs
rvl: I don't care much about performance.
But I think polymorphic is better here because then the caller can control the representation, and the unit tests and benchmarks can look simple because they use simple types directly.
`Amount` and `Cost` could reasonably be concreted to `Natural`, but the other types could be anything.

prototypes/coin-selection/CoinSelection.hs
rvl: These are good points.

jonathanknowles (commented):
Hi @rvl

Thanks for creating this PR!

I have a few comments and questions:

### Questions about scope

1. What user story do we intend for this design to fulfil? It seems to me that we might be interested in the following user story:
    > As a Cardano application developer, I want to have a Haskell library I can link to that can balance partially-completed transactions, where the resulting transactions are fully-balanced, have minimal fees, and are guaranteed to be accepted by the Cardano ledger. I want to be able to supply partial transactions that can use the full range of functionality allowed by Cardano API types, including withdrawals, minting, burning, execution units, scripts, protocol parameter updates, and so on. I want to know that this transaction balancing functionality is at least equivalent in power to that provided by the `balanceTransaction` endpoint of `cardano-wallet`, and that it is tested to at least the same level of quality.
    
    Do we have a user story like this already? If so, then we could link to it. If not, then we could create one.

2.  Am I right in saying that this PR seems to be about designing a more general interface for our existing [`balanceTransaction` function](https://github.com/input-output-hk/cardano-wallet/blob/ea9f07ce2806d5132b95723136c18c63b63da9f8/lib/core/src/Cardano/Wallet.hs#L1413), rather than coin selection per se? As I see it, coin selection (or UTxO selection) is just one of the many tasks performed by `balanceTransaction`. But in the vision presented in this PR, `balanceTransaction` has other responsibilities, including:
    - planning how to create multiple transactions, when one transaction isn't enough.
    - transforming domain-specific concepts such as minting and burning into lower-level concepts that coin selection can understand.
    - final post-hoc fee minimization

3. The PR links to ADP-1449, but that particular ticket is strictly limited in scope to updating our existing `CoinSelection` modules to remove dependencies on these types: the `TokenMap`, `TokenBundle`, and `Coin` types. Should we create a new ticket with a story similar to the one above? (I'm guessing that such a ticket does not yet exist. 😄)

4. I like the thought experiment approach used in this PR, as it allows us to think about how we might redesign `balanceTransaction` without the shackles of having to refactor our current code base. This approach seems complementary to the approach that @Anviking  is taking, which is to take the [existing `balanceTransaction` function](https://github.com/input-output-hk/cardano-wallet/blob/ea9f07ce2806d5132b95723136c18c63b63da9f8/lib/core/src/Cardano/Wallet.hs#L1413) and simplify the type signature until it cannot be made simpler. Perhaps the final design would be somewhere in the middle?

### Questions about this prototype

1. Is the API intended to be used with Cardano API types?
1. How should the API respond if the partial transaction supplied is already too large? Should the API return an error?
1. How should the API respond if any of the existing outputs violate the maximum output size limit?
1. How should the API respond if any of the existing outputs do not satisfy the minimum ada quantity? (I'm assuming that we will want to generalize the "minimum ada quantity" to be a function that's agnostic to different assets, so that `balanceTransaction` does not need to know about ada in particular.)
1. How should the API respond if any of the existing outputs have asset quantities in excess of the upper limit?
1. How is collateral represented? What behaviour should the API implementation have if insufficient collateral is presented in the partial transaction? If the implementation is responsible for selecting more collateral, then what behaviour should the API implementation when the remaining UTxO set does not have enough entries suitable for use as collateral? (There have been some ideas presented elsewhere, such as the creation of an initial "re-balancing" transaction that liberates one or more pure ada entries for use as collateral. See ([here](https://github.com/cardano-foundation/CIPs/pull/104#issuecomment-894044142)).) 
1. How does this API support the ability to compute a minimum ada quantity? We could imagine a function `computeMinimumValue` of type `Value -> Value` that increases the relevant components of a `Value` until it satisfies `satisfiesMinimumValue :: Value -> Bool`. The internal coin selection algorithm needs to be able to call a function similar to this one in order to produce valid change outputs. I think that a Boolean test is probably not powerful enough.
1. In the design presented, is it possible for the caller to specific an additional source of input value, or an additional sink for output value? Justification: the internal coin selection algorithm needs to distinguish between UTxO inputs and other forms of input value, since the distribution of UTxO inputs are also a useful source of information on how to construct change values. 
1. Under what circumstances should the API return just one transaction. Under what circumstances should it return more?

Anviking:
> This approach seems complementary to the approach that @Anviking is taking, which is to take the [existing balanceTransaction function](https://github.com/input-output-hk/cardano-wallet/blob/ea9f07ce2806d5132b95723136c18c63b63da9f8/lib/core/src/Cardano/Wallet.hs#L1413) and simplify the type signature until it cannot be made simpler. Perhaps the final design would be somewhere in the middle?

FYI, @rvl, here is one (quick) sketch for that: https://hackmd.io/I_ChHOHBQvSGbo52lS7Jng

rvl:
The ticket number shouldn't matter too much @jonathanknowles.
ADP-1449 seems close enough.
The point is that we come up with a common idea and terminology for where the API is going, that it be as simple as possible, and that there is actual code to reference.

